PR #12954, opened 2022-05-11
docs: add `sysbatch` to scheduling internals

schmichael:
We could definitely write more about it here, but I just wanted to get the smallest possible correction out first.


DerekStrickland (approved):
:shipit:

PR #12955, opened 2022-05-11
core: emit node evals only for sys jobs in dc

schmichael:
Whenever a node joins the cluster, either for the first time or after
being `down`, we emit a evaluation for every system job to ensure all
applicable system jobs are running on the node.

This patch adds an optimization to skip creating evaluations for system
jobs not in the current node's DC. While the scheduler performs the same
feasability check, skipping the creation of the evaluation altogether
saves disk, network, and memory.

### DIFF
--- a/.changelog/12955.txt
+++ b/.changelog/12955.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+core: On node updates skip creating evaluations for jobs not in the node's datacenter.
+```
--- a/nomad/node_endpoint.go
+++ b/nomad/node_endpoint.go
@@ -171,7 +171,7 @@ func (n *Node) Register(args *structs.NodeRegisterRequest, reply *structs.NodeUp
 
 	// Check if we should trigger evaluations
 	if shouldCreateNodeEval(originalNode, args.Node) {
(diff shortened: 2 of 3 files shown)


tgross (approved):
LGTM! This feels like a good place to look for additional optimizations in reducing evals too!

schmichael:
@tgross I was hoping to just get a PR up, but an issue is probably a good idea: https://github.com/hashicorp/nomad/issues/12981

I'm tempted to base further work off this PR and get them all merged in a group (hence the Draft status)... but that might just be complicating things for no reason. Let me know if you have a preference for how to handle a batch of small eval improvements.

tgross:
> I'm tempted to base further work off this PR and get them all merged in a group (hence the Draft status)... but that might just be complicating things for no reason. Let me know if you have a preference for how to handle a batch of small eval improvements.

Yeah sorry I may have jumped the gun in reviewing here knowing that you've got an active investigation underway. But batching them up seems fine. Ideally each fix is a discrete commit within the PR to help reviewability?

robloxrob:
Hooray! Christmas in July!

PR #12959, opened 2022-05-12
Add nomad env to envoy bootstrap

jorgemarey:
Just saw #12543 .We are doing things a bit different in a fork we have. This PR takes a different aproach and adds enviroment variables to the envoy boostrap hook. This along with https://github.com/hashicorp/consul/pull/13049 allow us to configure things like the following in the consul configuration:

```
kind = "proxy-defaults"
name = "global"

Config {
      envoy_dogstatsd_url = "udp://172.26.64.1:9125"
      envoy_stats_tags = [
        "alloc_id=${NOMAD_ALLOC_ID}",
        "nomad_group=${NOMAD_GROUP_NAME}",
        "nomad_job=${NOMAD_JOB_NAME}",
        "nomad_namespace=${NOMAD_NAMESPACE}",
      ]
}
```
This can also be set on the nomad sidecar configuration.

### DIFF
--- a/.changelog/12959.txt
+++ b/.changelog/12959.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+connect: add nomad environment variables to envoy bootstrap
+```
--- a/client/allocrunner/taskrunner/envoy_bootstrap_hook.go
+++ b/client/allocrunner/taskrunner/envoy_bootstrap_hook.go
@@ -292,6 +292,8 @@ func (h *envoyBootstrapHook) Prestart(ctx context.Context, req *ifs.TaskPrestart
 
 	// Create environment
 	bootstrapEnv := bootstrap.env(os.Environ())
+	// append nomad environment variables to the bootstrap environment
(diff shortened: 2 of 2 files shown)


Amier3:
We might want to put a hold on this pending the changes in https://github.com/hashicorp/consul/pull/13049

shoenig:
Unless I'm mistaken, closed by https://github.com/hashicorp/nomad/pull/14311

jorgemarey:
Hi @shoenig. I guess that this specific case presented in the issue (adding the stats tags to envoy) is solved by https://github.com/hashicorp/nomad/pull/14311, but I don't think it's the only use case for passing the NOMAD_ environment variables to the envoy task. I think that providing those values to envoy could allow users to better modify their escape-hatchs if necesary. What do you think?

shoenig:
Sounds reasonable enough to me ~

shoenig (commented):
LGTM

Issue #10941, opened 2021-07-26
Cannot update a dispatched job


sirkrypt0:
### Nomad version

`Nomad v1.1.2 (60638a086ef9630e2a9ba1e237e8426192a44244)`

### Operating system and Environment details

Ubuntu 20.04

### Issue

Jobs created via `dispatch` cannot be updated.

### Reproduction steps

1. Start a Nomad dev aget
    ```
    $ nomad agent -dev
    ```
1. Run the reproduction parameterized batch job
    ```
    $ nomad run batch.nomad
    ```
1. Dispatch an instance of the parameterized job
    ```
    $ nomad job dispatch batch
    ```
1. Open the Nomad UI and navigate into the batch job
1. Navigate into the dispatched job
1. Edit the jobs definition, e.g., change available resources
1. Plan and submit the job

#### Expected Result

The job is updated correctly.

#### Actual Result

The following error is shown:

```
cannot update non-parameterized job to being parameterized
```

### Job file (if appropriate)

```hcl
job "batch" {
  datacenters = ["dc1"]
  type        = "batch"

  parameterized {}

  group "batch" {
    task "sleep" {
      driver = "docker"

      config {
        image   = "alpine:latest"
        command = "sleep"
        args = ["infinity"]
      }

      resources {
        cpu    = 40
        memory = 40
      }
    }
  }
}
```

tgross:
Hi @sirkrypt0! That's the expected behavior, but not well documented and definitely a poor error message. Once a job has been dispatched, the dispatch payload has been set and it's no longer "parameterized". But trying to update it uses the parameterized job definition of its parent job, and that can't be done.

Generally speaking what you'll want to do is to update the parent job and then re-dispatch it.

sirkrypt0:
Alright, thanks for the clarification :)

Then this issue can be closed, unless you want to keep it open for a better documentation/error message.

waquidvp:
@tgross [nomad-pipeline](https://github.com/HyperBadger/nomad-pipeline) is a small tool I am working on. It allows you do create "pipeline" jobs in Nomad. It does this by using lifecycle hooks that updates the task groups `count`. It was inspired by this blog post: https://www.cgbaker.net/2020/12/hacking-nomad-job-dependencies/.

Since it relies on having to update the job to work, the behavior described in this bug means that it doesn't work with parameterized (dispatched) job. I know you have a valid reason for closing this bug however I though I show a possible use-case for needing the ability to update a dispatched job.

[...]

PR #12973, opened 2022-05-12
docs: note that already-dispatched jobs cannot be updated

tgross:
Closes https://github.com/hashicorp/nomad/issues/10941


shoenig (approved):
LGTM

PR #12932, opened 2022-05-10
secure variables: initial state store

tgross:
This changeset is a mostly-empty implementation of the memdb schema and state store operations we'll need for https://github.com/hashicorp/nomad/issues/12802. 

Nothing here has gone through RFC review yet, so it'll likely need to be revised, but this can serve as a skeleton for us to do little spikes as part of the design process. Implementations are largely incomplete.

PR #12980, opened 2022-05-13
Provide mock secure variables implementation

angrycub:
This PR provides a minimal backend to test the secure variable API. 

 ⚠️ **Warning** ⚠️ 
Data in this mock backend will become irrational with leadership transitions, so it can not be used in any way other than to perform early work on the UI in a single node case.

- secure variables: initial state store (#12932)
- Add SecureVariable mock
- Add SecureVariableStub
- Add SecureVariable Copy and AsStub funcs
- Wire up secure variable mock interface
- Fix panic when not RPC endpoint not found

### DIFF
--- a/command/agent/http.go
+++ b/command/agent/http.go
@@ -57,13 +57,16 @@ var (
 	// tag isn't enabled
 	stubHTML = "<html><p>Nomad UI is disabled</p></html>"
 
-	// allowCORS sets permissive CORS headers for a handler
-	allowCORS = cors.New(cors.Options{
-		AllowedOrigins:   []string{"*"},
-		AllowedMethods:   []string{"HEAD", "GET"},
-		AllowedHeaders:   []string{"*"},
-		AllowCredentials: true,
-	})
+	// allowCORSWithMethods sets permissive CORS headers for a handler, used by
+	// wrapCORS and wrapCORSWithMethods
+	allowCORSWithMethods = func(methods ...string) *cors.Cors {
+		return cors.New(cors.Options{
+			AllowedOrigins:   []string{"*"},
+			AllowedMethods:   methods,
+			AllowedHeaders:   []string{"*"},
+			AllowCredentials: true,
+		})
+	}
 )
 
 type handlerFn func(resp http.ResponseWriter, req *http.Request) (interface{}, error)
@@ -407,9 +410,13 @@ func (s HTTPServer) registerHandlers(enableDebug bool) {
 	s.mux.HandleFunc("/v1/operator/scheduler/configuration", s.wrap(s.OperatorSchedulerConfiguration))
 
 	s.mux.HandleFunc("/v1/event/stream", s.wrap(s.EventStream))
+
 	s.mux.HandleFunc("/v1/namespaces", s.wrap(s.NamespacesRequest))
 	s.mux.HandleFunc("/v1/namespace", s.wrap(s.NamespaceCreateRequest))
 	s.mux.HandleFunc("/v1/namespace/", s.wrap(s.NamespaceSpecificRequest))
+
+	s.mux.Handle("/v1/vars", wrapCORS(s.wrap(s.SecureVariablesListRequest)))
+	s.mux.Handle("/v1/var/", wrapCORSWithAllowedMethods(s.wrap(s.SecureVariableSpecificRequest), "HEAD", "GET", "PUT", "DELETE"))
 
 	uiConfigEnabled := s.agent.config.UI != nil && s.agent.config.UI.Enabled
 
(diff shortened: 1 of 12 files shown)


nomad/structs/secure_variables.go
```diff
@@ -28,6 +28,65 @@ type SecureVariableData struct {
...
+			out.UnencryptedData[k] = v
+		}
+	}
+	if sv.EncryptedData != nil {
+		out.EncryptedData = sv.EncryptedData.Copy()
+	}
+	return out
+}
+
+func (sv SecureVariable) AsStub() SecureVariableStub {
```
tgross, line 61: Super nitpicky for consistency: this method is always called `Stub()` in the rest of the code base.

nomad/structs/secure_variables.go
```diff
@@ -28,6 +28,65 @@ type SecureVariableData struct {
...
+
+func (sv SecureVariable) Copy() SecureVariable {
+	out := SecureVariable{
+		Namespace:   sv.Namespace,
+		Path:        sv.Path,
+		CreateIndex: sv.CreateIndex,
+		CreateTime:  sv.CreateTime,
+		ModifyIndex: sv.ModifyIndex,
+		ModifyTime:  sv.ModifyTime,
+	}
```
tgross, line 48: Primitive fields are copy-on-assignment. So typically we do this:

```suggestion
func (sv *SecureVariable) Copy() *SecureVariable {
	if sv == nil {
		return nil
	}
	out := *sv
```

It looks slightly "clever" but it avoids having to remember to add every new field to the `Copy` method except for the collections (maps, pointers, and slices).

command/agent/secure_variable_endpoint.go
```diff
@@ -0,0 +1,109 @@
...
+	return out.Data, nil
+}
+
+func (s *HTTPServer) SecureVariableSpecificRequest(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
+	path := strings.TrimPrefix(req.URL.Path, "/v1/var/")
+	if len(path) == 0 {
+		return nil, CodedError(400, "Missing secure variable path")
+	}
+	switch req.Method {
+	case "GET":
```
tgross, line 39: There are consts for these like `http.MethodGet` in `net/http`: https://pkg.go.dev/net/http#pkg-constants but we're admittedly not very good about remembering to use them 😀

command/agent/http.go
```diff
@@ -407,10 +410,14 @@ func (s HTTPServer) registerHandlers(enableDebug bool) {
 	s.mux.HandleFunc("/v1/operator/scheduler/configuration", s.wrap(s.OperatorSchedulerConfiguration))
 
 	s.mux.HandleFunc("/v1/event/stream", s.wrap(s.EventStream))
+
 	s.mux.HandleFunc("/v1/namespaces", s.wrap(s.NamespacesRequest))
 	s.mux.HandleFunc("/v1/namespace", s.wrap(s.NamespaceCreateRequest))
 	s.mux.HandleFunc("/v1/namespace/", s.wrap(s.NamespaceSpecificRequest))
 
+	s.mux.Handle("/v1/vars", wrapCORS(s.wrap(s.SecureVariablesListRequest)))
+	s.mux.Handle("/v1/var/", wrapCORSWithAllowedMethods(s.wrap(s.SecureVariableSpecificRequest), "HEAD", "GET", "PUT", "DELETE"))
```
tgross, line 419: Do we want to leave a TODO here about cleaning this up when we're done with some of the initial implementation? I think there's value in supporting the OpenAPI tooling you demonstrated but we probably want to do that more holistically across the API.

helper/codec/inmem.go
```diff
@@ -20,6 +20,9 @@ func (i *InmemCodec) ReadRequestHeader(req *rpc.Request) error {
 }
 
 func (i *InmemCodec) ReadRequestBody(args interface{}) error {
+	if args == nil {
+		return nil
+	}
```
tgross, line 25: Interesting... is this an existing panic bug? Or is it the kind of thing we can only ever hit in testing?

tgross (approved):
LGTM @angrycub! I've left some comments and suggestions but this is great. I spun up the PR to check it out myself:

```
$ curl -XPOST -d '{"Namespace": "default", "Items": {"k":"v"}}' http://127.0.0.1:4646/v1/var/my/weird/thing

$ nomad operator api '/v1/var/my/weird/thing'  | jq .
{
  "CreateIndex": 0,
  "CreateTime": null,
  "Items": {
    "k": "v"
  },
  "ModifyIndex": 0,
  "ModifyTime": null,
  "Namespace": "default",
  "Path": "my/weird/thing"
}
```

nomad/secure_variable_mock.go
```diff
@@ -0,0 +1,107 @@
...
+// Delete removes a key from the store. Removing a non-existent key is a no-op
+func (mvs *MockVariableStore) Reset() {
+	fmt.Println("***** Reset *****")
+	mvs.m.Lock()
+	mvs.m.Unlock()
+	mvs.backingMap = make(map[string]*structs.SecureVariable)
+}
+
+func init() {
+	fmt.Println("***** Initializing mock variables backend *****")
```
tgross, line 73: This is breaking the integration test in `command`: it's printing `****` at the start of the `nomad job inspect` command, and the test is expecting _only_ the JSON body to come back. Let's remove this (or move it to a logger).

helper/codec/inmem.go
angrycub: If code makes it out if the shop with an unwired RPC, it would panic.  It's unlikely because of testing, but the panic hides the error from being output when the test fails; so I lost a lot of time step-debugging what would have been a comparatively simple error fix.

tgross: Good catch then, that makes life easier for everyone going forward :+1:

Issue #12963, opened 2022-05-12
Adding CSI volume in-place Job update has no effect


rbastiaans-tc:
### Nomad version
```
Nomad v1.3.0-rc.1 (31b0a1841cb6d3c7aa64bb2162f50798ba3b3d6e)
```

### Operating system and Environment details

Debian 10 Buster

### Issue

Submitting an updated job with the change of adding a CSI volume has no effect.

No volume is added to the allocation and the allocation is not restarted.

Note: stopping and then starting the job does result in a volume on the allocation, but in-place update does not.

### Reproduction steps

* Submit and run a job with no (CSI) volume
* Modify jobfile and add a (CSI) volume
* Submit modified job

```
# nomad run test_no_volume.hcl 
==> 2022-05-12T08:23:48Z: Monitoring evaluation "be1e9cdc"
    2022-05-12T08:23:48Z: Evaluation triggered by job "csitest"
==> 2022-05-12T08:23:49Z: Monitoring evaluation "be1e9cdc"
    2022-05-12T08:23:49Z: Evaluation within deployment: "dd37c4ab"
    2022-05-12T08:23:49Z: Allocation "3867e54b" created: node "12aa7338", group "csitest"
    2022-05-12T08:23:49Z: Allocation "5a5cd642" created: node "33aed506", group "csitest"
    2022-05-12T08:23:49Z: Allocation "8fc577f9" created: node "c9e801c0", group "csitest"
    2022-05-12T08:23:49Z: Evaluation status changed: "pending" -> "complete"
==> 2022-05-12T08:23:49Z: Evaluation "be1e9cdc" finished with status "complete"
==> 2022-05-12T08:23:49Z: Monitoring deployment "dd37c4ab"
  ✓ Deployment "dd37c4ab" successful
    
    2022-05-12T08:24:00Z
    ID          = dd37c4ab
    Job ID      = csitest
    Job Version = 6
    Status      = successful
    Description = Deployment completed successfully
    
    Deployed
    Task Group  Desired  Placed  Healthy  Unhealthy  Progress Deadline
    csitest     3        3       3        0          2022-05-12T08:33:59Z

# nomad volume status
Container Storage Interface
ID                         Name                       Plugin ID          Schedulable  Access Mode
csitest[0]                 csitest[0]                 cinder_csi_plugin  true         <none>
csitest[1]                 csitest[1]                 cinder_csi_plugin  true         <none>
[...]

PR #13008, opened 2022-05-13
scheduler: volume updates should always be destructive

tgross:
Fixes https://github.com/hashicorp/nomad/issues/12963

No task driver with file system isolation can support updating mounts in-place. Host volume and CSI volume or volume mount additions or updates should therefore always be destructive.

---

This changeset includes some unit testing but I've manually tested it on real CSI volumes as well. Adding a new volume:

```
$ nomad job plan ./demo/csi/hostpath/redis.nomad
+/- Job: "example"
+/- Task Group: "cache" (1 create/destroy update)
  +   Volume {
      + AccessMode:     "single-node-reader-only"
      + AttachmentMode: "file-system"
      + Name:           "volume0"
      + PerAlloc:       "true"
      + ReadOnly:       "true"
      + Source:         "test-volume"
      + Type:           "csi"
      }
  +/- Task: "redis" (forces in-place update)
```

Altering an existing volume:

```
$ nomad job plan ./demo/csi/hostpath/redis.nomad
+/- Job: "example"
+/- Task Group: "cache" (1 create/destroy update)
  +/- Volume {
    +/- AccessMode:     "single-node-reader-only" => "single-node-writer"
        AttachmentMode: "file-system"
        Name:           "volume0"
        PerAlloc:       "true"
    +/- ReadOnly:       "true" => "false"
        Source:         "test-volume"
        Type:           "csi"
      }
      Task: "redis"
```

### DIFF
--- a/.changelog/13008.txt
+++ b/.changelog/13008.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+volumes: Fixed a bug where additions, updates, or removals of host volumes or CSI volumes were not treated as destructive updates
+```
--- a/scheduler/util.go
+++ b/scheduler/util.go
@@ -524,6 +524,12 @@ func tasksUpdated(jobA, jobB *structs.Job, taskGroup string) bool {
 		return true
 	}
 
+	// Check if volumes are updated (no task driver can support
+	// altering mounts in-place)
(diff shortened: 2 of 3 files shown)


DerekStrickland (approved):
:shipit:

PR #13012, opened 2022-05-13
12986 UI fails to load job when there is an "@" in job name in nomad 130

### LINKED ISSUE #12986: UI fails to load job when there is @ in job name in nomad 1.3.0
### Nomad version

```
Nomad v1.3.0 (52e95d64113e01be05d585d8b4c07f6f19efebbc)
```

```
NAME="Ubuntu"
VERSION="20.04.2 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.2 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

### Issue

Nomad UI returns 404 when there is `@` sign in job name

image

### Reproduction steps

Run nomad job with `@` in name

#### Expected Result

UI should show normal job status

#### Actual Result

404

### Job file (if appropriate)

```
job "@hello_world" {

  datacenters = ["dc1"]
  type = "service"

  group "app" {
    count = 2

    network {
      port "http" {
        to = 8000
      }
    }


    service {
      name = "webapp"
      tags = ["urlprefix-/", "traefik.enable=true", "traefik.http.routers.http.rule=Path(`/`)"]
      port = "http"

      check {
        name     = "alive"
        type     = "http"
        path     = "/"
        interval = "10s"
        timeout  = "2s"
      }
    }


    restart {
      attempts = 2
      interval = "30m"
      delay = "15s"
      mode = "fail"
    }

    task "server" {
      driver = "docker"

      config {
        image = "mnomitch/hello_world_server"
        ports = ["http"]
      }

      env {
        MESSAGE = "Hello World!"
      }
    }
  }
}

```

UI sends request with namespace set to job name. I think there is some code that extracts namespace from URL by spliting url by `@` sign

```
Request URL: https://[redacted]/v1/job/?namespace=hello_world
```

philrenaud:
Resolves a bug where we were mis-identifying the namespace of jobs with the "@" character in their names.

Now, we explicitly link to @default for any namespace-less links to jobs, and specifically look for the last "@" in a job name to determine namespace.

Resolves #12986

### DIFF
--- a/ui/app/models/job.js
+++ b/ui/app/models/job.js
@@ -37,13 +37,7 @@ export default class Job extends Model {
 
   @computed('plainId')
   get idWithNamespace() {
-    const namespaceId = this.belongsTo('namespace').id();
-
-    if (!namespaceId) {
-      return this.plainId;
-    } else {
-      return `${this.plainId}@${namespaceId}`;
-    }
+    return `${this.plainId}@${this.belongsTo('namespace').id()}`;
   }
 
   @computed('periodic', 'parameterized', 'dispatched')
--- a/ui/app/models/job.js
+++ b/ui/app/models/job.js
@@ -39,7 +39,7 @@ export default class Job extends Model {
   get idWithNamespace() {
     const namespaceId = this.belongsTo('namespace').id();
 
-    if (!namespaceId || namespaceId === 'default') {
+    if (!namespaceId) {
       return this.plainId;
     } else {
       return `${this.plainId}@${namespaceId}`;
--- a/ui/app/models/job.js
+++ b/ui/app/models/job.js
@@ -37,7 +37,7 @@ export default class Job extends Model {
 
   @computed('plainId')
   get idWithNamespace() {
-    return `${this.plainId}@${this.belongsTo('namespace').id()}`;
+    return `${this.plainId}@${this.belongsTo('namespace').id() ?? 'default'}`;
   }
 
   @computed('periodic', 'parameterized', 'dispatched')
--- a/ui/app/routes/jobs/job.js
+++ b/ui/app/routes/jobs/job.js
@@ -9,6 +9,7 @@ export default class JobRoute extends Route {
   @service can;
   @service store;
   @service token;
+  @service router;
 
   serialize(model) {
     return { job_name: model.get('idWithNamespace') };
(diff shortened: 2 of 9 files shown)


ui/app/routes/jobs/job.js
```diff
@@ -15,7 +15,17 @@ export default class JobRoute extends Route {
...
+    let name,
+      namespace = 'default';
+    const { job_name } = params;
+    const delimiter = job_name.lastIndexOf('@');
+    if (delimiter !== -1) {
+      name = job_name.slice(0, delimiter);
+      namespace = job_name.slice(delimiter + 1);
+    } else {
+      name = job_name;
+      namespace = 'default';
```
DerekStrickland, line 27: Not a blocker, but do you need this again?

ui/app/models/job.js
```diff
@@ -37,13 +37,7 @@ export default class Job extends Model {
...
   @computed('plainId')
   get idWithNamespace() {
-    const namespaceId = this.belongsTo('namespace').id();
-
-    if (!namespaceId || namespaceId === 'default') {
-      return this.plainId;
-    } else {
-      return `${this.plainId}@${namespaceId}`;
-    }
+    return `${this.plainId}@${this.belongsTo('namespace').id()}`;
```
DingoEatingFuzz, line 40: Relying on `this.belongsTo('namespace').id()` to have a value scares me a little bit. I think it always should but it might be nice to still be defensive here and slip a `?? 'default'` in.

DerekStrickland (approved):
:shipit:

ui/app/routes/jobs/job.js
```diff
@@ -15,7 +15,17 @@ export default class JobRoute extends Route {
...
+    }
```
DingoEatingFuzz, line 28: Maybe we can also add a redirect for when `job_name` doesn't have the delimiter? So `/jobs/foo` redirects to `/jobs/foo@default`.

ui/app/routes/jobs/job.js
philrenaud: Good catch!

ui/app/models/job.js
philrenaud: Good call

ui/app/routes/jobs/job.js
philrenaud: Good suggestion; adding the following:

```
      this.router.transitionTo('jobs.job.index', `${name}@${namespace}`);
```

PR #13016, opened 2022-05-13
keyring state store operations

tgross:
Implement the basic upsert, list, and delete operations for `RootKeyMeta` needed by the Keyring RPCs.

This changeset also implements two convenience methods `RootKeyMetaByID` and `GetActiveRootKeyMeta` which are useful for testing but also will be needed to implement the rest of the RPCs.

### DIFF
--- a/nomad/state/state_store.go
+++ b/nomad/state/state_store.go
@@ -6627,14 +6627,92 @@ func (s *StateStore) SecureVariablesQuotas(ws memdb.WatchSet) (memdb.ResultItera
 	return iter, nil
 }
 
+// UpsertRootKeyMeta saves root key meta or updates it in-place.
 func (s *StateStore) UpsertRootKeyMeta(index uint64, rootKeyMeta *structs.RootKeyMeta) error {
-	return nil
+	txn := s.db.WriteTxn(index)
+	defer txn.Abort()
+
+	// get any existing key for updating
+	raw, err := txn.First(TableRootKeyMeta, indexID, rootKeyMeta.KeyID)
+	if err != nil {
+		return fmt.Errorf("root key metadata lookup failed: %v", err)
+	}
+
+	isRotation := false
+
+	if raw != nil {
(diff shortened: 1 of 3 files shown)


nomad/state/state_store.go
```diff
@@ -6627,14 +6627,92 @@ func (s *StateStore) SecureVariablesQuotas(ws memdb.WatchSet) (memdb.ResultItera
...
 func (s *StateStore) UpsertRootKeyMeta(index uint64, rootKeyMeta *structs.RootKeyMeta) error {
-	return nil
+	txn := s.db.WriteTxn(index)
+	defer txn.Abort()
+
+	// get any existing key for updating
+	raw, err := txn.First(TableRootKeyMeta, indexID, rootKeyMeta.KeyID)
+	if err != nil {
+		return fmt.Errorf("root key metadata lookup failed: %v", err)
+	}
```
angrycub, line 6639: If this is an Upsert, should it fail on missing key?

tgross: The API for memdb is a little surprising here, but `txn.First` returns 3 possible states:
* a valid pointer and no error, if the key exists. Think of it as the "200 OK" case.
* a nil pointer and no error, if the key doesn't exist. The "404" case.
* a nil pointer and an error, if the lookup throws some internal error. The "500" case. (Ex. there _is_ no `TableRootKeyMeta` table, or a bad key, or whatever. In Nomad's usage these are basically always programmer errors that we catch in testing.)

You can see the difference in handling in the `Delete` method below:

```go
	existing, err := txn.First(TableRootKeyMeta, indexID, keyID)
	if err != nil {
		return fmt.Errorf("root key metadata lookup failed: %v", err)
	}
	if existing == nil {
		return fmt.Errorf("root key metadata not found")
	}
```

For `Upsert`, we want to return an error only if there's a real error, and just continue if we didn't get any value back because we'll just use the one the caller provided.

PR #13029, opened 2022-05-16
docs: API package tests need a binary with your changes

tgross:
Add a note to the contributing guide pointing out that if you're
writing `api` package tests, you need to build a binary with any of
your changes.


lgfa29 (approved):
It gets me every time 😬

PR #13041, opened 2022-05-17
[WIP] consul-template: Add fault tolerant defaults

DerekStrickland:
** DO NOT MERGE **

This PR adds fault-tolerant defaults to the default agent config.

- `BlockQueryWaitTime`: defaults to 5 min, matching the [default](https://www.consul.io/api-docs/features/blocking#blocking-queries) value Consul uses
- `MaxStale`: defaults to 10 years matching the [default](https://www.consul.io/docs/agent/config/config-files#max_stale) value Consul uses
- `Wait.Min`: defaults to 5 seconds overriding the [default](https://github.com/hashicorp/consul-template/blob/9b0db8d7e76ee01ecde4db53fbd5c11f7eb9a6a8/config/wait.go#L90) of 0 that Consul Template uses
- `Wait.Max`: defaults to 4 minutes matching the [default](https://github.com/hashicorp/consul-template/blob/9b0db8d7e76ee01ecde4db53fbd5c11f7eb9a6a8/config/wait.go#L90) Consul Template uses
- `ConsulRetry.Attempts`: defaults to 0 which is the `unlimited` value - overrides the Consul Template [default](https://github.com/hashicorp/consul-template/blob/9b0db8d7e76ee01ecde4db53fbd5c11f7eb9a6a8/config/retry.go#L11) of 12
- `ConsulRetry.Backoff`: defaults to 250ms matching the [default](https://github.com/hashicorp/consul-template/blob/9b0db8d7e76ee01ecde4db53fbd5c11f7eb9a6a8/config/retry.go#L15) that Consul Template uses
- `ConsulRetry.MaxBackoff`: defaults to 1 minute matching [default](https://github.com/hashicorp/consul-template/blob/9b0db8d7e76ee01ecde4db53fbd5c11f7eb9a6a8/config/retry.go#L18) that Consul Template uses
- `VaultRetry.Attempts`: defaults to 0 which is the `unlimited` value - overrides the Consul Template [default](https://github.com/hashicorp/consul-template/blob/9b0db8d7e76ee01ecde4db53fbd5c11f7eb9a6a8/config/retry.go#L11) of 12
- `VaultRetry.Backoff`: defaults to 250ms matching the [default](https://github.com/hashicorp/consul-template/blob/9b0db8d7e76ee01ecde4db53fbd5c11f7eb9a6a8/config/retry.go#L15) that Consul Template uses
- `VaultRetry.MaxBackoff`: defaults to 1 minute matching [default](https://github.com/hashicorp/consul-template/blob/9b0db8d7e76ee01ecde4db53fbd5c11f7eb9a6a8/config/retry.go#L18) that Consul Template uses

### DIFF
--- a/client/config/config.go
+++ b/client/config/config.go
@@ -717,8 +717,24 @@ func DefaultConfig() *Config {
 		NoHostUUID:              true,
 		DisableRemoteExec:       false,
 		TemplateConfig: &ClientTemplateConfig{
-			FunctionDenylist: DefaultTemplateFunctionDenylist,
-			DisableSandbox:   false,
+			FunctionDenylist:   DefaultTemplateFunctionDenylist,
+			DisableSandbox:     false,
+			BlockQueryWaitTime: helper.TimeToPtr(5 * time.Minute),   // match Consul default
+			MaxStale:           helper.TimeToPtr(87600 * time.Hour), // match Consul default
+			Wait: &WaitConfig{
+				Min: helper.TimeToPtr(5 * time.Second),
+				Max: helper.TimeToPtr(1 * time.Hour),
+			},
+			ConsulRetry: &RetryConfig{
+				Attempts:   helper.IntToPtr(0), // unlimited
+				Backoff:    helper.TimeToPtr(250 * time.Millisecond),
+				MaxBackoff: helper.TimeToPtr(1 * time.Minute),
+			},
+			VaultRetry: &RetryConfig{
+				Attempts:   helper.IntToPtr(0), // unlimited
+				Backoff:    helper.TimeToPtr(250 * time.Millisecond),
+				MaxBackoff: helper.TimeToPtr(1 * time.Minute),
+			},
 		},
 		RPCHoldTimeout:     5 * time.Second,
 		CNIPath:            "/opt/cni/bin",


mikenomitch:
Since this will probably involve some docs changes, can we sneak in an explanation of the unit of "max_stale" here https://www.nomadproject.io/docs/configuration/client#max_stale (maybe the default value being there will make this obvious?)

A user had mentioned that this was a bit confusing, and they didnt know if it was a time value or index diff or something like that.

DerekStrickland:
> Since this will probably involve some docs changes, can we sneak in an explanation of the unit of "max_stale" here https://www.nomadproject.io/docs/configuration/client#max_stale (maybe the default value being there will make this obvious?)
> 
> A user had mentioned that this was a bit confusing, and they didn't know if it was a time value or index diff or something like that.

I've updated the default value in that last commit, but IIRC this text is lifted directly from the Consul docs. Any specific feedback about what is unclear?

client/config/config.go
```diff
@@ -717,8 +717,24 @@ func DefaultConfig() *Config {
...
+			BlockQueryWaitTime: helper.TimeToPtr(5 * time.Minute),         // match Consul default
+			MaxStale:           helper.TimeToPtr(DefaultTemplateMaxStale), // match Consul default
+			Wait: &WaitConfig{
+				Min: helper.TimeToPtr(5 * time.Second),
+				Max: helper.TimeToPtr(4 * time.Minute),
+			},
+			ConsulRetry: &RetryConfig{
+				Attempts:   helper.IntToPtr(0), // unlimited
+				Backoff:    helper.TimeToPtr(250 * time.Millisecond),
+				MaxBackoff: helper.TimeToPtr(1 * time.Minute),
```
tgross, line 731: If we weren't setting these, they'd use the CT default (right?), so setting them here to the default is really more for our own documentation. Do we want to comment that these defaults were lifted directly from CT so that in the future we don't think of them as having been specifically designed?

tgross (approved):
The code change here LGTM. 

I've left some suggestions on cleaning up the documentation while we're in here, but maybe that should be in a separate PR for backporting concerns. We'll also need a changelog entry and a section of the version-specific upgrade guide detailing the changes we're making here.

client/config/config.go
DerekStrickland: That's a good point. I think I'll exclude `Backoff` and `MaxBackoff` and just rely on the `Finalize` in CT if the user doesn't specify. The real improvement is the unlimited attempts.

client/config/config.go
```diff
@@ -717,8 +717,24 @@ func DefaultConfig() *Config {
 		NoHostUUID:              true,
 		DisableRemoteExec:       false,
 		TemplateConfig: &ClientTemplateConfig{
```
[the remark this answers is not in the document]
DerekStrickland: I believe that this is for the client. In case you meant `jobspec`, the only overlap with the `jobspec` template stanza is `wait`. The `jobpsec` value will override the client config as long as it is within the `wait_bounds` if specified. Otherwise it will take the bounded value (one or both depending on if both bounds are crossed).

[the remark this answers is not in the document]
tgross: Oh right, sorry, wrong part of the code 😊

website/content/docs/configuration/client.mdx
```diff
@@ -250,23 +250,23 @@ chroot as doing so would cause infinite recursion.
...
   for the results of a blocking query. Many endpoints in Consul support a feature known as
   "blocking queries". A blocking query is used to wait for a potential change
   using long polling.
 
-- `consul_retry` `(Code: nil)` - This controls the retry behavior when an error is
-  returned from Consul. Consul Template is highly fault tolerant, meaning it does
-  not exit in the face of failure. Instead, it uses exponential back-off and retry
-  functions to wait for the cluster to become available, as is customary in distributed
-  systems.
+- `consul_retry` `(Code: { attempts = 0 backoff = "250ms" max_backoff = "1m" })`-
```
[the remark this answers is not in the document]
DerekStrickland: I like map. I'd prefer that to a separate code block, but I can go either way. Updating to map now for next submission.

website/content/docs/configuration/client.mdx
```diff
@@ -217,14 +217,14 @@ chroot as doing so would cause infinite recursion.
...
+- `max_stale` `(string: "87600h")` - # This is the maximum interval to allow "stale"
   data. By default, only the Consul leader will respond to queries. Requests to
   a follower will forward to the leader. In large clusters with many requests,
   this is not as scalable. This option allows any follower to respond to a query,
   so long as the last-replicated data is within this bound. Higher values result
   in less cluster load, but are more likely to have outdated data.
 
-- `wait` `(Code: nil)` - Defines the minimum and maximum amount of time to wait
+- `wait` `(Code: { min = "5s" max = "4m" })` - Defines the minimum and maximum amount of time to wait
   for the Consul cluster to reach a consistent state before rendering a template.
```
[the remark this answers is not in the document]
DerekStrickland: I reviewed the consul-template code and updated it based on my understanding of what that code is doing. It's a bit different than what I feel the previous doc stated. Wouldn't mind reviewing this with another set of eyes to make sure it's accurate.  Maybe at our co-working session tomorrow afternoon?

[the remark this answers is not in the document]
tgross: Sorry, I may be misunderstanding... it looks like your change was to remove "for the Consul cluster to reach a consistent state"? That describes the behavior less, doesn't it?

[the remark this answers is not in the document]
DerekStrickland: @tgross  Conferred with CT engineer and got his approval on the current text.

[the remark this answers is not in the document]
tgross: ok, works for me

website/content/docs/configuration/client.mdx
```diff
@@ -217,27 +217,35 @@ chroot as doing so would cause infinite recursion.
   files on the client host via the `file` function. By default, templates can
   access files only within the [task working directory].
 
-- `max_stale` `(string: "")` - # This is the maximum interval to allow "stale"
+- `max_stale` `(string: "87600h")` - This is the maximum interval to allow "stale"
   data. By default, only the Consul leader will respond to queries. Requests to
```
schmichael, line 221: > By default, only the Consul leader will respond to queries.

I don't think this is true anymore right? Looks like 87.6k hours is the new default.

website/content/docs/configuration/client.mdx
```diff
@@ -217,27 +217,35 @@ chroot as doing so would cause infinite recursion.
...
   in less cluster load, but are more likely to have outdated data.
 
-- `wait` `(Code: nil)` - Defines the minimum and maximum amount of time to wait
-  for the Consul cluster to reach a consistent state before rendering a template.
-  This is useful to enable in systems where network connectivity to Consul is degraded,
-  because it will reduce the number of times a template is rendered. This configuration is
-  also exposed in the _task template stanza_ to allow overrides per task.
+- `wait` `(map: { min = "5s" max = "4m" })` - Defines the minimum and maximum amount
+  of time to wait before attempting to re-render a template. By default, Consul Template
+  will loop continually and re-render. If this value is set, Nomad will configure Consul
```
schmichael, line 229: > By default, Consul Template

Aren't 5s/4m the defaults?

website/content/docs/configuration/client.mdx
DerekStrickland: Nice catch!

website/content/docs/configuration/client.mdx
DerekStrickland: They are. I thought our convention was to put the defaults in these code snippet sections. Is that not the case?

schmichael: Oh hm, maybe I just misread this sentence. I think when I see:

> By default X. If this value is set, Y.

I assume that the value in question is *not* set by default. The phrasing definitely isn't explicitly saying that, so maybe it's just my poor reading comprehension at work. :sweat_smile: 

The preceding and following sentences clear it all up, but maybe these 2 could be rephrased a bit:

> Consul Template re-renders templates whenever rendered variables from Consul, Nomad, or Vault change. However in order to minimize how often tasks are restarted or reloaded, Nomad will configure <rest of the second sentence here>.

My main point being is that I think the `By default...` in the first sentence makes users think _this parameter changes that behavior when it doesn't._ Re-rendering in a loop is just how Consul Template works and these parameters just tweak some timings.

DerekStrickland: Ah ok. That makes sense. I'll stage a new PR with these doc changes.

valodzka: > Aren't 5s/4m the defaults?

They are not. Default max in consul-template 4 times of **Min**imum value not 4 **Min**utes

valodzka:
@DerekStrickland I noticed that you incorrectly read [consul-template code](https://github.com/hashicorp/consul-template/blob/9b0db8d7e76ee01ecde4db53fbd5c11f7eb9a6a8/config/wait.go#L100).  `c.Max = TimeDuration(4 * *c.Min)` mean 4 times of **Min**imum value, not 4 **Min**utes.

DerekStrickland:
@valodzka Thanks for catching this!

PR #13044, opened 2022-05-17
cli: update default redis and use nomad service discovery

### LINKED ISSUE #12927: docs: update redis example version
We make use of `redis:3.2` in a lot of examples in docs and in `nomad job init`. 

https://hub.docker.com/_/redis doesn't even list redis 3 as a supported version anymore. 

Update the docs and reference job(s) to `redis:7`, which should be a supported version for quite some time.

### LINKED ISSUE #12958: Example Redis deployment doesn't work on 1.3.0 without consul
### Nomad version
Output from `nomad version`

```
Nomad v1.3.0 (52e95d64113e01be05d585d8b4c07f6f19efebbc)
```

```
uname -a                         
Linux pop-os 5.17.5-76051705-generic #202204271406~1651504840~22.04~63e51bd SMP PREEMPT Mon May 2 15: x86_64 x86_64 x86_64 GNU/Linux

lsb_release -a
No LSB modules are available.
Distributor ID:	Pop
Description:	Pop!_OS 22.04 LTS
Release:	22.04
Codename:	jammy
```

### Issue

The example deployment from `nomad init` doesn't seem to run anymore on a single node Nomad acting as server+client (and no consul).

### Reproduction steps

1. Save the file as `nomad.hcl`

```hcl
datacenter = "dc1"
data_dir   = "/opt/nomad/data"

log_level = "DEBUG"

bind_addr = "0.0.0.0"

server {
  enabled          = true
  bootstrap_expect = 1
}

client {
  enabled = true
  meta {
    stack       = "enigma"
    component   = "rms"
    rms_node_id = "1"
  }
}

plugin "docker" {
  config {
    allow_privileged = true
    volumes {
      enabled = true
    }
    extra_labels = ["job_name", "job_id", "task_group_name", "task_name", "namespace", "node_name", "node_id"]
  }
}
```

2. Run nomad with:

```bash
nomad agent -dev -config=nomad.hcl
```

3. Run the example file

```bash
nomad init
nomad run example.nomad
```

4. Visit the UI for failed alloc

image

#### Expected Result

As of 1.2.6 until yesterday this worked normally, so it should also continue to work on 1.3.0?

#### Actual Result

It seems to check whether consul is present on the client or not and errors out with `Constraint ${attr.consul.version} semver >= 1.7.0 filtered 1 node`

When I go to the client attributes list (at http://localhost:4646/ui/clients/c8927149-1161-f27c-7079-1adce6d56697), I don't see any `consul` block. Which makes sense because this `consul` block doesn't exist in the agent config file. The same issue happens even if I run `nomad agent -dev` so the agent config is not the issue here, I suppose.

### Job file (if appropriate)

Same as `nomad init`

shoenig:
This PR updates the version of redis used in our examples from 3.2 to 7.
The old version is very not supported anymore, and we should be setting
a good example by using a supported version.

The long-form example job is now fixed so that the service stanza uses
nomad as the service discovery provider, and so now the job runs without
a requirement of having Consul running and configured.

Closes #12927
Closes #12958

### DIFF
--- a/api/jobs_test.go
+++ b/api/jobs_test.go
@@ -576,7 +576,7 @@ func TestJobs_Canonicalize(t *testing.T) {
 								Name:   "redis",
 								Driver: "docker",
 								Config: map[string]interface{}{
-									"image": "redis:3.2",
+									"image": "redis:7",
 									"port_map": []map[string]int{{
 										"db": 6379,
 									}},
@@ -708,7 +708,7 @@ func TestJobs_Canonicalize(t *testing.T) {
 								Name:   "redis",
 								Driver: "docker",
 								Config: map[string]interface{}{
-									"image": "redis:3.2",
+									"image": "redis:7",
 									"port_map": []map[string]int{{
 										"db": 6379,
 									}},
--- a/client/allocrunner/taskrunner/task_runner_test.go
+++ b/client/allocrunner/taskrunner/task_runner_test.go
@@ -681,7 +681,7 @@ func TestTaskRunner_TaskEnv_Image(t *testing.T) {
 	task := alloc.Job.TaskGroups[0].Tasks[0]
 	task.Driver = "docker"
 	task.Config = map[string]interface{}{
-		"image":        "redis:3.2-alpine",
+		"image":        "redis:7-alpine",
 		"network_mode": "none",
 		"command":      "sh",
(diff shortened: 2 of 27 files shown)


jrasell (approved):
LGTM, thanks @shoenig!

Issue #11638, opened 2021-12-07
Evaluation broker control API


tgross:
The Nomad engineering team would love to get some community feedback on a proposed set of features!

During incident response for Nomad or for Nomad workloads, operators may find that the scheduler can compound the ongoing incident by pushing forward with evaluations as best it can. We've recently been working on a bunch of improvements for incident response like bypassing `shutdown_delay` (https://github.com/hashicorp/nomad/issues/11448), making `num_schedulers` tunable via API (https://github.com/hashicorp/nomad/issues/11449), overriding evaluation priority on job registration (https://github.com/hashicorp/nomad/issues/11434), and an API to halt job registration (https://github.com/hashicorp/nomad/issues/11450).

Operators may want to have some sort of top-level "maintenance mode" where evaluation processing is paused or otherwise controlled while a problem is being debugged.

### Eval Broker Pause/Resume

Evaluations are the unit of work for the Nomad scheduler. They are created...
* When you register, dispatch, or scale a job.
* When clients update the server with information about failed allocations.
* When the scheduler can't place all allocations in a job and creates a "blocked" eval to try again later.
* When the leader schedules a periodic job or garbage collection internal job.

The eval broker is the component on the leader that takes evaluations that have been written to raft and queues them for scheduling on one of the scheduler workers. Pausing the eval broker would prevent the scheduler queues from receiving new work and let the workers catch up so they can reduce CPU, memory, and disk resources. But note that this also includes any scheduling work necessary to reschedule failed allocations and periodic tasks! This isn't something to be done lightly, but as an emergency measure by operators.

For implementing Pause/Resume, there's an existing `enabled` flag which gets toggled whenever a server steps up/down from leadership (the eval broker only runs on the leader). We check this flag in a few places:

* When Evals get upserted, they're enqueued from the FSM on the leader, that is when `enabled = true` (ref [`fsm.go#L746`](https://github.com/hashicorp/nomad/blob/v1.2.2/nomad/fsm.go#L746))
* Likewise when an Eval is dequeued we check that `enabled` flag (ref [`eval_broker.go#L374-L377`](https://github.com/hashicorp/nomad/blob/v1.2.2/nomad/eval_broker.go#L374-L377))
* The leader has a eval restorer that enqueues all the pending evals (ref [`leader.go#L491-L517`](https://github.com/hashicorp/nomad/blob/v1.2.2/nomad/leader.go#L491-L517))
[...]

PR #13045, opened 2022-05-17
core: allow pausing and un-pausing of leader broker routine

jrasell:
Test failure: percy-ui and unrelated
Related to https://github.com/hashicorp/nomad/issues/11638
Commits: split to make review easier

This changes allows operators to control whether or not the eval broker is running on the leader server. This is useful in outage situations where administrators wish to stop any work being available to the scheduler workers. It is also requisite work for future work which will allow deletion of evaluations which have not been processed.

Alongside pausing the eval broker, the blocked evals process is also disabled at the same time. This sub-process pushes evaluations into the eval broker and is also restored from the state store using the same process as the eval broker. It therefore makes sense, considering the end goal, to disable both processes for state consistency.

To ensure operators pausing/un-pausing the broker doesn't conflict with leadership transitions a mutex is used to control access to the eval broker and blocked evals processes. This is used along with a leadership check when changing the broker status which requires taking into account operator configuration.

The issue mentions disabling the `reapFailedEvaluations` process, however, I don't believe this is required for this change as disabling the eval broker also flushes all stored evaluation data. It therefore seemed safer to leave this alone and not require additional coordination.

The new `operator scheduler` commands allow inspecting and modifying the scheduler config. This is useful as it doesn't require you to supply the full payload object or remember the exact curl command to run.

### DIFF
--- a/website/content/docs/commands/operator/scheduler-set-config.mdx
+++ b/website/content/docs/commands/operator/scheduler-set-config.mdx
@@ -1,0 +1,82 @@
+---
+layout: docs
+page_title: 'Commands: operator scheduler set-config'
+description: |
+  Modify the scheduler configuration.
+---
+
+# Command: operator scheduler set-config
+
+The scheduler operator set-config command is used to modify the scheduler
+configuration.
+
+## Usage
+
+```plaintext
+nomad operator scheduler set-config [options]
+```
+
+If ACLs are enabled, this command requires a token with the `operator:write`
+capability.
+
+## General Options
+
+@include 'general_options_no_namespace.mdx'
+
+## Set Config Options
+
+- `-check-index` - If set, the scheduler config is only updated if the passed
+  modify index matches the current server side version. If a non-zero value is
+  passed, it ensures that the scheduler config is being updated from a known
+  state.
+
+- `-scheduler-algorithm` - Specifies whether scheduler binpacks or spreads
+  allocations on available nodes. Must be one of `["binpack"|"spread"]`.
+
+- `-memory-oversubscription` - When true, tasks may exceed their reserved memory
+  limit, if the client has excess memory capacity. Tasks must specify [`memory_max`]
+  to take advantage of memory oversubscription. Must be one of `[true|false]`.
+
+- `-reject-job-registration` - When true, the server will return permission denied
+  errors for job registration, job dispatch, and job scale APIs, unless the ACL
+  token for the request is a management token. If ACLs are disabled, no user
+  will be able to register jobs. This allows operators to shed load from automated
+  processes during incident response. Must be one of `[true|false]`.
+
+- `-pause-eval-broker` - When set to true, the eval broker which usually runs on
+  the leader will be disabled. This will prevent the scheduler workers from
+  receiving new work. Must be one of `[true|false]`.
+
+- `-preempt-batch-scheduler` - Specifies whether preemption for batch jobs
+  is enabled. Note that if this is set to true, then batch jobs can preempt any
+  other jobs. Must be one of `[true|false]`.
+
+- `-preempt-service-scheduler` - Specifies whether preemption for service jobs
+  is enabled. Note that if this is set to true, then service jobs can preempt any
+  other jobs. Must be one of `[true|false]`.
+
+- `-preempt-sysbatch-scheduler` - Specifies whether preemption for system batch
+  jobs is enabled. Note that if this is set to true, then system batch jobs can
+  preempt any other jobs. Must be one of `[true|false]`.
+
+- `-preempt-system-scheduler` - Specifies whether preemption for system jobs
+  is enabled. Note that if this is set to true, then system jobs can preempt any
+  other jobs. Must be one of `[true|false]`.
+
+## Examples
+
+Modify the scheduler algorithm to spread:
+
+```shell-session
+$ nomad operator scheduler set-config -scheduler-algorithm=spread
+Scheduler configuration updated!
+```
+
+Modify the scheduler algorithm to spread using the check index flag:
+
+```shell-session
+$ nomad operator scheduler set-config -scheduler-algorithm=spread -check-index=5
+Scheduler configuration updated!
+```
+
+[`memory_max`]: /docs/job-specification/resources#memory_max
--- a/.changelog/13045.txt
+++ b/.changelog/13045.txt
@@ -1,0 +1,7 @@
+```release-note:improvement
+cli: Added `scheduler get-config` and `scheduler set-config` commands to the operator CLI
+```
+
+```release-note:improvement
+core: Added the ability to pause and un-pause the eval broker and blocked eval broker
+```
--- a/website/content/api-docs/operator/scheduler.mdx
+++ b/website/content/api-docs/operator/scheduler.mdx
@@ -40,18 +40,20 @@ $ curl \
   "Index": 5,
   "KnownLeader": true,
   "LastContact": 0,
+  "NextToken": "",
   "SchedulerConfig": {
     "CreateIndex": 5,
+    "MemoryOversubscriptionEnabled": false,
     "ModifyIndex": 5,
-    "SchedulerAlgorithm": "spread",
-    "MemoryOversubscriptionEnabled": true,
-    "RejectJobRegistration": false,
+    "PauseEvalBroker": false,
     "PreemptionConfig": {
-      "SystemSchedulerEnabled": true,
-      "SysBatchSchedulerEnabled": false,
       "BatchSchedulerEnabled": false,
-      "ServiceSchedulerEnabled": false
-    }
+      "ServiceSchedulerEnabled": false,
+      "SysBatchSchedulerEnabled": false,
+      "SystemSchedulerEnabled": true
+    },
+    "RejectJobRegistration": false,
+    "SchedulerAlgorithm": "binpack"
   }
 }
 ```
@@ -64,9 +66,23 @@ $ curl \
 - `SchedulerConfig` `(SchedulerConfig)` - The returned `SchedulerConfig` object has configuration
   settings mentioned below.
 
-  - `SchedulerAlgorithm` `(string: "binpack")` - Specifies whether scheduler binpacks or spreads allocations on available nodes.
+  - `SchedulerAlgorithm` `(string: "binpack")` - Specifies whether scheduler
+    binpacks or spreads allocations on available nodes.
 
-  - `MemoryOversubscriptionEnabled` `(bool: false)` <sup>1.1 Beta</sup> - When `true`, tasks may exceed their reserved memory limit, if the client has excess memory capacity. Tasks must specify [`memory_max`](/docs/job-specification/resources#memory_max) to take advantage of memory oversubscription.
+  - `MemoryOversubscriptionEnabled` `(bool: false)` <sup>1.1 Beta</sup> - When
+    `true`, tasks may exceed their reserved memory limit, if the client has excess
+    memory capacity. Tasks must specify [`memory_max`](/docs/job-specification/resources#memory_max)
+    to take advantage of memory oversubscription.
+
+  - `RejectJobRegistration` `(bool: false)` - When `true`, the server will return
+    permission denied errors for job registration, job dispatch, and job scale APIs,
+    unless the ACL token for the request is a management token. If ACLs are disabled,
+    no user will be able to register jobs. This allows operators to shed load from
+    automated processes during incident response.
+
+  - `PauseEvalBroker` `(bool: false)` - When set to `true`, the eval broker which
+    usually runs on the leader will be disabled. This will prevent the scheduler
+    workers from receiving new work.
 
   - `PreemptionConfig` `(PreemptionConfig)` - Options to enable preemption for various schedulers.
 
@@ -120,6 +136,7 @@ server state is authoritative.
   "SchedulerAlgorithm": "spread",
   "MemoryOversubscriptionEnabled": false,
   "RejectJobRegistration": false,
+  "PauseEvalBroker": false,
   "PreemptionConfig": {
     "SystemSchedulerEnabled": true,
     "SysBatchSchedulerEnabled": false,
@@ -133,9 +150,20 @@ server state is authoritative.
   binpacks or spreads allocations on available nodes. Possible values are
   `"binpack"` and `"spread"`
 
-- `MemoryOversubscriptionEnabled` `(bool: false)` <sup>1.1 Beta</sup> - When `true`, tasks may exceed their reserved memory limit, if the client has excess memory capacity. Tasks must specify [`memory_max`](/docs/job-specification/resources#memory_max) to take advantage of memory oversubscription.
+- `MemoryOversubscriptionEnabled` `(bool: false)` <sup>1.1 Beta</sup> - When
+  `true`, tasks may exceed their reserved memory limit, if the client has excess
+  memory capacity. Tasks must specify [`memory_max`](/docs/job-specification/resources#memory_max)
+  to take advantage of memory oversubscription.
 
-- `RejectJobRegistration` `(bool: false)` - When `true`, the server will return permission denied errors for job registration, job dispatch, and job scale APIs, unless the ACL token for the request is a management token. If ACLs are disabled, no user will be able to register jobs. This allows operators to shed load from automated proceses during incident response.
+- `RejectJobRegistration` `(bool: false)` - When `true`, the server will return
+  permission denied errors for job registration, job dispatch, and job scale APIs,
+  unless the ACL token for the request is a management token. If ACLs are disabled,
+  no user will be able to register jobs. This allows operators to shed load from
+  automated processes during incident response.
+
+- `PauseEvalBroker` `(bool: false)` - When set to `true`, the eval broker which
+  usually runs on the leader will be disabled. This will prevent the scheduler
+  workers from receiving new work.
 
 - `PreemptionConfig` `(PreemptionConfig)` - Options to enable preemption for
   various schedulers.
@@ -143,9 +171,9 @@ server state is authoritative.
   - `SystemSchedulerEnabled` `(bool: true)` - Specifies whether preemption for
     system jobs is enabled. Note that if this is set to true, then system jobs
     can preempt any other jobs.
-    
-  - `SysBatchSchedulerEnabled` `(bool: false)` <sup>1.2</sup> - Specifies 
-    whether preemption for system batch jobs is enabled. Note that if this is 
+
+  - `SysBatchSchedulerEnabled` `(bool: false)` <sup>1.2</sup> - Specifies
+    whether preemption for system batch jobs is enabled. Note that if this is
     set to true, then system batch jobs can preempt any other jobs.
 
   - `BatchSchedulerEnabled` `(bool: false)` - Specifies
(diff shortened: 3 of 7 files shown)


command/operator_scheduler_set_config.go
```diff
@@ -0,0 +1,170 @@
...
+	if err != nil {
+		o.Ui.Error(fmt.Sprintf("Error initializing client: %s", err))
+		return 1
+	}
+
+	// Fetch the current configuration. This will be used as a base to merge
+	// user configuration onto.
+	resp, _, err := client.Operator().SchedulerGetConfiguration(nil)
+	if err != nil {
+		o.Ui.Error(fmt.Sprintf("Error querying for Autopilot configuration: %s", err))
```
schmichael, line 83:
```suggestion
		o.Ui.Error(fmt.Sprintf("Error querying for scheduler configuration: %s", err))
```

nomad/eval_endpoint.go
```diff
@@ -126,6 +126,13 @@ func (e *Eval) Dequeue(args *structs.EvalDequeueRequest,
 			args.SchedulerVersion, scheduler.SchedulerVersion)
 	}
 
+	// If the eval broker is not enabled, but we are the leader, we have
+	// nothing to do. Not checking this will return errors to the workers which
+	// does nothing but clog up the logs.
+	if !e.srv.evalBroker.Enabled() {
+		return nil
+	}
```
schmichael, line 134: Will this cause the workers to try to Dequeue in a tight loop that does nothing? Logging errors is good to avoid, but we also have 0 visibility into when this is hit.

When there are no evals to be processed the eval broker blocks the Dequeue RPC until there is an eval to be processed. Perhaps we need a similar mechanism here so that instead of returning when the eval broker is disabled, we block until it is enabled or there is a leadership transition?

jrasell: I didn't realise quite how tight that loop was; thanks for pointing this out.

Yeah that sort of functionality makes sense and I will look into how this could be implemented over the coming days.

jrasell:
I am also wondering if we need evalBroker enabled checks on other RPC handlers such as `Ack` and `Nack`?

command/operator_scheduler_set_config.go
```diff
@@ -0,0 +1,170 @@
...
+    job dispatch, and job scale APIs, unless the ACL token for the request is a
+    management token. If ACLs are disabled, no user will be able to register jobs.
+    This allows operators to shed load from automated processes during incident
+    response.
+
+  -pause-eval-broker=[true|false]
+    When set to true, the eval broker which usually runs on the leader will be
+    disabled. This will prevent the scheduler workers from receiving new work.
+
+  -preemption-batch-scheduler=[true|false]
```
DerekStrickland, line 153: Out of curiosity, what happens if they set more than one of these to true?

DerekStrickland (approved):
Great job on a difficult task! I left some non-blocking comments and questions. I'd personally wait on schmichael's review to merge, but you may have already been through that and it's just not updated.

e2e/operator_scheduler/input/basic.nomad
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  group "operator_scheduler" {
+
+    task "test" {
+      driver = "raw_exec"
+
+      config {
+        command = "bash"
+        args    = ["-c", "sleep 15000"]
```
tgross, line 17: It might be worth dropping this down to 1s and making it a batch job so that we can guarantee a quick exit once we've turned the broker back on; that'll reduce the risk of test flakiness on the `WaitForJobStop` (which definitely _should_ be enough time, but...)

lgfa29 (approved):
Nice work!

Lots of minor comments, but one thing that may be worth looking into is to disable deployment monitoring for the `nomad job run` command if the eval broker is disabled, otherwise it will get stuck:

```console
$ nomad run example.nomad
==> 2022-06-10T11:11:17-04:00: Monitoring evaluation "076115a0"
    2022-06-10T11:11:17-04:00: Evaluation triggered by job "countdash"
==> 2022-06-10T11:11:18-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:19-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:20-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:21-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:22-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:23-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:24-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:25-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:26-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:27-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:28-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:29-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:30-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:31-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:32-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:33-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:34-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:35-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:36-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:37-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:38-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:39-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:40-04:00: Monitoring evaluation "076115a0"
==> 2022-06-10T11:11:41-04:00: Monitoring evaluation "076115a0"
```

helper/broker/notify.go
```diff
@@ -0,0 +1,108 @@
...
+
+	// Create a timeout timer and use the helper to ensure this routine doesn't
+	// panic and making the stop call clear.
+	timeoutTimer, timeoutStop := helper.NewSafeTimer(timeout)
+
+	// Defer a function which performs all the required cleanup of the
+	// subscriber once it has been notified of a change, or reached its wait
+	// timeout.
+	defer func() {
+		g.unsubscribeCh <- updateCh
```
tgross, line 93: I spent a half hour writing up a whole thing about how I think we can potentially block here and cause updates to get dropped but then re-read and realized that it's not an issue because this function just returns and isn't responsible for new updates. So I'm leaving this comment here to spare another reviewer the same effort. 😊

website/content/docs/commands/operator/scheduler-set-config.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+@include 'general_options_no_namespace.mdx'
+
+## Set Config Options
+
+- `-scheduler-algorithm` - Specifies whether scheduler binpacks or spreads
+  allocations on available nodes. Must be one of `["binpack"|"spread"]`.
+
+- `-memory-oversubscription` - When true, tasks may exceed their reserved memory
+  limit, if the client has excess memory capacity. Tasks must specify memory_max
```
tgross, line 32:
```suggestion
  limit, if the client has excess memory capacity. Tasks must specify `memory_max`
```

(maybe link it too?)

website/content/docs/commands/operator/scheduler-set-config.mdx
```diff
@@ -0,0 +1,68 @@
...
+  to take advantage of memory oversubscription. Must be one of `[true|false]`.
+
+- `-reject-job-registration`- When true, the server will return permission denied
```
tgross, line 35:
```suggestion
- `-reject-job-registration` - When true, the server will return permission denied
```

nomad/eval_endpoint.go
```diff
@@ -131,6 +131,23 @@ func (e *Eval) Dequeue(args *structs.EvalDequeueRequest,
...
+	// SetEnabled function was called to modify its state. It is possible this
+	// is because of leadership transition, therefore the RPC should exit to
+	// allow all safety checks and RPC forwarding to occur again.
+	//
+	// The log line is trace, because the default worker timeout is 500ms which
+	// produces a large amount of logging.
+	if !e.srv.evalBroker.Enabled() {
+		message := e.srv.evalBroker.enabledNotifier.WaitForChange(args.Timeout)
+		e.logger.Trace("eval broker wait for un-pause", "message", message)
+		return nil
```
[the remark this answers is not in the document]
jrasell: The server workers are the only processes which call this endpoint and will only utilise the meta if the [response eval object is not nil](https://github.com/hashicorp/nomad/blob/main/nomad/worker.go#L484-L487). Interestingly, the worker uses the meta that is set [here](https://github.com/hashicorp/nomad/blob/main/nomad/eval_endpoint.go#L157-L159) and noting which is set with the `setQueryMeta` function call. Therefore skipping the `setQueryMeta` call is fine here.

command/operator_scheduler_set_config.go
DerekStrickland: That answers it.

command/operator_scheduler_set_config.go
```diff
@@ -0,0 +1,170 @@
...
+	return mergeAutocompleteFlags(o.Meta.AutocompleteFlags(FlagSetClient),
+		complete.Flags{
+			"-scheduler-algorithm":           complete.PredictAnything,
+			"-memory-oversubscription":       complete.PredictAnything,
+			"-reject-job-registration":       complete.PredictAnything,
+			"-pause-eval-broker":             complete.PredictAnything,
+			"-preemption-batch-scheduler":    complete.PredictAnything,
+			"-preemption-service-scheduler":  complete.PredictAnything,
+			"-preemption-sysbatch-scheduler": complete.PredictAnything,
+			"-preemption-system-scheduler":   complete.PredictAnything,
```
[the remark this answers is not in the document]
jrasell: I went with `preempt-*-scheduler` to keep it inline somewhat with the API object. I also changes the prediction like you pointed out.

jrasell:
> Nice work!
> 
> Lots of minor comments, but one thing that may be worth looking into is to disable deployment monitoring for the `nomad job run` command if the eval broker is disabled, otherwise it will get stuck:
> 
> ```
> $ nomad run example.nomad
> ==> 2022-06-10T11:11:17-04:00: Monitoring evaluation "076115a0"
>     2022-06-10T11:11:17-04:00: Evaluation triggered by job "countdash"
> ==> 2022-06-10T11:11:18-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:19-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:20-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:21-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:22-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:23-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:24-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:25-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:26-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:27-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:28-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:29-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:30-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:31-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:32-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:33-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:34-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:35-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:36-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:37-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:38-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:39-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:40-04:00: Monitoring evaluation "076115a0"
> ==> 2022-06-10T11:11:41-04:00: Monitoring evaluation "076115a0"
> ```
> 
> Also missing a `CHANGELOG` entry.

Changelog has been added.

I quite like the UX of the job submission in situation when the eval broker is paused. It shows the evaluation is making no progress which is exactly correct and will unblock if the eval broker is enabled and the eval being watched changes state. In order to avoid this we would need to make an extra call to the API for another that monitors evals such as deployments promotion, job stop, and others. That commands also have the detach flag which can be used to avoid this.

tgross:
That the "Monitoring evaluation" message falls inside the loop instead of at the top feels almost accidental; in most cases we'd only ever see it once. If we move it up so that it's outside the loop we still the effect of "hey we're waiting here" without making a bunch of noise in the terminal?

lgfa29:
> That the "Monitoring evaluation" message falls inside the loop instead of at the top feels almost accidental; in most cases we'd only ever see it once. If we move it up so that it's outside the loop we still the effect of "hey we're waiting here" without making a bunch of noise in the terminal?

Good point, I think that would work as well 👍

command/operator_scheduler_get_config.go
```diff
@@ -0,0 +1,117 @@
...
+	// Output the information.
+	o.Ui.Output(formatKV([]string{
+		fmt.Sprintf("Scheduler Algorithm|%s", schedConfig.SchedulerAlgorithm),
+		fmt.Sprintf("Memory Oversubscription|%v", schedConfig.MemoryOversubscriptionEnabled),
+		fmt.Sprintf("Reject Job Registration|%v", schedConfig.RejectJobRegistration),
+		fmt.Sprintf("Pause Eval Broker|%v", schedConfig.PauseEvalBroker),
+		fmt.Sprintf("Preemption System Scheduler|%v", schedConfig.PreemptionConfig.SystemSchedulerEnabled),
+		fmt.Sprintf("Preemption Service Scheduler|%v", schedConfig.PreemptionConfig.ServiceSchedulerEnabled),
+		fmt.Sprintf("Preemption Batch Scheduler|%v", schedConfig.PreemptionConfig.BatchSchedulerEnabled),
+		fmt.Sprintf("Preemption SysBatch Scheduler|%v", schedConfig.PreemptionConfig.SysBatchSchedulerEnabled),
```
schmichael, line 85: Can we include `"Modify Index|%v", resp.ModifyIndex` to encourage use of CAS with these operations? Really don't want operators overwriting one another's configs unintentionally.

schmichael:
```suggestion
		fmt.Sprintf("Preemption SysBatch Scheduler|%v", schedConfig.PreemptionConfig.SysBatchSchedulerEnabled),
		fmt.Sprintf("Modify Index|%v", resp.ModifyIndex),
```

command/operator_scheduler_set_config.go
```diff
@@ -0,0 +1,172 @@
...
+	flags.Usage = func() { o.Ui.Output(o.Help()) }
+
+	flags.StringVar(&o.schedulerAlgorithm, "scheduler-algorithm", "", "")
+	flags.Var(&o.memoryOversubscription, "memory-oversubscription", "")
+	flags.Var(&o.rejectJobRegistration, "reject-job-registration", "")
+	flags.Var(&o.pauseEvalBroker, "pause-eval-broker", "")
+	flags.Var(&o.preemptBatchScheduler, "preempt-batch-scheduler", "")
+	flags.Var(&o.preemptServiceScheduler, "preempt-service-scheduler", "")
+	flags.Var(&o.preemptSysBatchScheduler, "preempt-sysbatch-scheduler", "")
+	flags.Var(&o.preemptSystemScheduler, "preempt-system-scheduler", "")
```
schmichael, line 68: Can we add `check-index` like `nomad job run` to allow and even encourage use of the existing CAS operator?

command/operator_scheduler_set_config.go
```diff
@@ -0,0 +1,172 @@
...
+	// Set up a client.
+	client, err := o.Meta.Client()
+	if err != nil {
+		o.Ui.Error(fmt.Sprintf("Error initializing client: %s", err))
+		return 1
+	}
+
+	// Fetch the current configuration. This will be used as a base to merge
+	// user configuration onto.
+	resp, _, err := client.Operator().SchedulerGetConfiguration(nil)
```
schmichael, line 83: If we add `-check-index` we should fail here if `resp.ModifyIndex != check-index` because that means someone else updated the scheduler config and the user could end up with unintentional results. Imagine during an outage:

1. Operator Alice runs `nomad operator scheduler set-config -reject-job-registration=true`
2. Operator Bob runs `nomad operator scheduler set-config -pause-eval-broker=true`

Without CAS there's no way to know which of the following will be the final state of the scheduler config:

1. RejectJobRegistration=true; PauseEvalBroker=false
2. RejectJobRegistration=false; PauseEvalBroker=true
3. RejectJobRegistration=true; PauseEvalBroker=true

Because the HTTP GETs and PUTs may interleave in any order.

Both Alice and Bob must _manually_ do a Get-before-Set and put the ModifyIndex form the result into the `-check-index` flag to ensure the resulting state is what they intend.

command/operator_scheduler_set_config.go
```diff
@@ -0,0 +1,172 @@
...
+	o.memoryOversubscription.Merge(&schedulerConfig.MemoryOversubscriptionEnabled)
+	o.rejectJobRegistration.Merge(&schedulerConfig.RejectJobRegistration)
+	o.pauseEvalBroker.Merge(&schedulerConfig.PauseEvalBroker)
+	o.preemptBatchScheduler.Merge(&schedulerConfig.PreemptionConfig.BatchSchedulerEnabled)
+	o.preemptServiceScheduler.Merge(&schedulerConfig.PreemptionConfig.ServiceSchedulerEnabled)
+	o.preemptSysBatchScheduler.Merge(&schedulerConfig.PreemptionConfig.SysBatchSchedulerEnabled)
+	o.preemptSystemScheduler.Merge(&schedulerConfig.PreemptionConfig.SystemSchedulerEnabled)
+
+	// Check-and-set the new configuration.
+	result, _, err := client.Operator().SchedulerCASConfiguration(schedulerConfig, nil)
```
schmichael, line 103: We should always include a CAS index here:

1. Prefer the user's from `-check-index` if it is set (> 0)
2. If it is not set use `resp.ModifyIndex` to ensure this set does not interleave with another user's set in an unintended way.

This is not a concern for most of our other APIs like job registration because they're doing _atomic_ sets and will therefore either successfully set the object to exactly what they intend or receive an error. The flags on this command introduce PATCH-like semantics so we have to ensure we're not interleaving updates.

schmichael (approved):

Issue #13031, opened 2022-05-16
1.3.0 container memory constraints not in effect leading to OOMs


djenriquez:
### Nomad version
Nomad v1.2.4 (9f21b724acf115a58d9ecdc97d281aef01d36387)
Nomad v1.3.0 (52e95d64113e01be05d585d8b4c07f6f19efebbc)

### Operating system and Environment details
Amazon Linux 2
### Issue
We recently upgraded our servers and nomad clients to 1.3.0 and witnessed a substantial increase in OOMs reported for java containerized applications in our system. Investigation shows that the OOMs are all being reported on Nomad 1.3.0 clients. 

When comparing the `docker inspect` of two different allocations for the same job, I noticed the following difference:
v1.3.0: `"CgroupParent": "cpuset",`
v.1.2.4: `"CgroupParent": "",`

And in the environment variables, this new env var pops up for 1.3.0: `"NOMAD_PARENT_CGROUP=/nomad",`

Also, an early clue was that only Java applications which used the `-XX:MaxRAMPercentage`, and related configuration, were affected. Java applications that used a hard defined `-Xmx` and `-Xms` were not affected. This lead us to thinking that cgroup limits were not being respected, which ended up being the case.

We believe this issue is related to https://github.com/hashicorp/nomad/pull/12274. When looking at our client properties, we do see a cgroup version, and see that we are `v1`, which leads us to believe this was supposed to be a no-op, but unfortunately, was not.
<img width="718" alt="Screen Shot 2022-05-16 at 10 00 08 AM" src="https://user-images.githubusercontent.com/4636740/168648391-ff94ad85-69b8-4e4d-b11a-703968b38b44.png">

djenriquez:
To add additional context:

In 1.2.4, we had a job defined with 4000MB of memory and a `XX:MaxRAMPercentage` of 50. When looking at the java heap, we see it is allocated ~2000MB, which is what we expect.

On 1.3.0, that same job was scheduled an allocation that reported a heap of ~8000MB, completely ignoring the 4000MB limit we had set.

tgross:
1.3.1 was a security and panic fix, so it didn't include the rest of the work merged into the 1.3.x series. We have a Nomad 1.3.2 planned soonish that'll include this.

binelson:
Great, thanks @tgross

PR #13058, opened 2022-05-18
drivers/docker: do not set cgroup parent in v1 mode

shoenig:
This PR fixes a bug where the `CgroupParent` on the docker
`HostConfig` struct was accidentally being set when running in
cgroups v1 mode.

Fixes https://github.com/hashicorp/nomad/issues/13031


tgross (approved):
LGTM!

djenriquez:
Nice, thanks!

PR #13060, opened 2022-05-18
Update who-uses-nomad.mdx

vmptk:
Add Fly.io (@superfly)


apollo13:
Question is how long they plan to stay on it https://community.fly.io/t/declarative-configuration/679/23 ;)

tgross:
@vmptk do you work for Fly? We generally only want to take PRs like this from folks who can "officially" attest to their own org using Nomad.

tgross:
Going to close this out

Issue #13067, opened 2022-05-19
Nomad Post-install script doesn't run


twunderlich-grapl:
### Nomad version
Output from `nomad version`
Nomad v1.3.0
### Operating system and Environment details
Amazon Linux 2
### Issue
As of Nomad 1.3.0, installing nomad does not create the /opt/nomad/data directory automatically
In essence it appears that .release/linux/postinst is not run.
My suspicion is that this broke as part of the move to hashicorps common release tooling in https://github.com/hashicorp/nomad/pull/12276

### Reproduction steps
Install Nomad 1.3.0 on Amazon Linux.
Check to see if /opt/nomad/data exists

#### Expected Result
/opt/nomad/data should exist
#### Actual Result
/opt/nomad/data  does not exist

PR #12276, opened 2022-03-14
[Main] Onboard to CRT

claire-labry:
Onboard Nomad OSS [`main`] to Common Release Tooling

There are a few important changes to call out in this PR:

GitHub Actions is being used as the CI provider to build the go binary artifacts and linux packages. This logic lives in .github/workflows/build.yml and workflow runs are available under https://github.com/hashicorp/nomad/actions.
Artifact builds will continue to run on PR's and merges to main. This behavior can be changed by updating the on: logic in the .github/workflows/build.yml file.

The .release/ci.hcl file contains some metadata for our release service. The only part you'll likely ever need to change is the release_branches array -- when you want to release off of a new release branch, update this value to the release branch name and be sure to update the branches array in .github/workflows/build.yml to the same value. This should be set to either main, the release branch that we'll be releasing off of (once release branches are adopted), or a test branch that you are working off of if you need to validate every part of the pipeline before merging.

PR to update `backport-assistant` https://github.com/hashicorp/nomad/pull/12311

Still TODO:

- [x] Open a separate PR to update and fix backport-assistant (Will tag that PR to this one for ease of visibility)
- [x] Rebase my commits to clean up this PR. 
- [x] Clean up this branch before merging [slack ID and branch naming from onboarding tests]

PR #13068, opened 2022-05-19
Use the postinstall script

twunderlich-grapl:
## Issue
Fixes https://github.com/hashicorp/nomad/issues/13067

## Summary
The linux postinstall script was created as part of https://github.com/hashicorp/nomad/pull/12276 but wasn't actually used. This sets the reference so the script will actually be used.
The primary benefit is that the following lines are run automatically as part of the install instead of requiring users to create it themselves
```
mkdir -p /opt/nomad/data
chown nomad:nomad /opt/nomad/data
chown -R nomad:nomad /etc/nomad.d
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -233,6 +233,7 @@ jobs:
           rpm_depends: "openssl"
           config_dir: ".release/linux/package/"
           preinstall: ".release/linux/preinst"
+          postinstall: ".release/linux/postinst"
           postremove: ".release/linux/postrm"
 
       - name: Set Package Names


mdeggies:
Great catch! Thanks for your contribution.

claire-labry:
Thank you @twunderlich-grapl for this catch 🙏🏻

claire-labry:
I think the tests are re-running -- thank you @mdeggies for adding the labels. I'll merge this in once the tests are finished

claire-labry (approved):

twunderlich-grapl:
Thank you all! And thank you @claire-labry for doing the original work of getting the release process standardized!

PR #13069, opened 2022-05-19
Secure Variables UI: /variables/new and /variables/*path

philrenaud:
- Adds routing and basic templates for clicking into a variable path, and creating new variables.
- Adds a new `<SecureVariablesForm>` component to do high-level operations on `/new`, but can be reused on `/*path/edit` later.
- Adds basic CRUD and model loading on variables routes
  - Currently not very efficient; we're loading /vars on ever /var/* in order to make sure we don't fall out of sync. Can probably reduce number of fetches here.
- Adds Ember Data transforms to modify createRecord / .save() behaviour

## Screenshots

Variables index:
image

A terminal variable path page:
image

A functional `<SecureVariablesForm>` 
image


ui/app/abilities/variable.js
```diff
@@ -10,10 +10,22 @@ export default class extends AbstractAbility {
...
   @computed('rulesForNamespace.@each.capabilities')
   get policiesSupportVariableView() {
     return this.rulesForNamespace.some((rules) => {
       return get(rules, 'SecureVariables');
     });
   }
+
+  @computed('rulesForNamespace.@each.capabilities')
+  get policiesSupportVariableCreation() {
+    return true; // TODO: check SecureVariables.<path>.capabilities[]
```
philrenaud, line 29: https://github.com/hashicorp/nomad/issues/13092

ui/app/adapters/var.js
```diff
@@ -0,0 +1,31 @@
...
+  // PUT instead of POST on create;
+  // /v1/var instead of /v1/vars on create (urlForFindRecord)
+  createRecord(store, type, snapshot) {
+    let data = this.serialize(snapshot);
+    return this.ajax(
+      this.urlForFindRecord(snapshot.id, snapshot.modelName),
+      'PUT',
+      { data }
+    );
+  }
```
philrenaud, line 30: This intercepts any createRecord() calls that Ember Data generates for new secure variables (any time a new model instantiation of `var` happens, and then is `.save()`ed), replaces POST with PUT, and singularizes the path noun.

ui/app/components/secure-variable-form.hbs
```diff
@@ -0,0 +1,42 @@
+  <form
+    class="new-secure-variables"
+    {{on "submit" this.saveNewVariable}}
+  >
+    {{!-- TODO: {{if this.parseError 'is-danger'}} on inputs --}}
```
philrenaud, line 5: Issue: https://github.com/hashicorp/nomad/issues/13093

ui/app/controllers/variables/new.js
```diff
@@ -0,0 +1,23 @@
+import Controller from '@ember/controller';
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+export default class VariablesNewController extends Controller {
+  @service router;
+
+  @action
+  saveNewVariable({ path, keyValues }) {
+    // TODO: validation
```
philrenaud, line 9: https://github.com/hashicorp/nomad/issues/13093

ui/app/templates/variables/index.hbs
```diff
@@ -3,13 +3,72 @@
...
+          <tr {{on "click" (fn this.goToVariable row.model)}}>
+            <td>
+              {{row.model.path}}
+            </td>
+            <td>
+              {{row.model.namespace}}
+            </td>
+          </tr>
+        </t.body>
+      </ListTable>
```
philrenaud, line 62: Currently a flat list of terminal paths, but plans to make this more conventional are in issue https://github.com/hashicorp/nomad/issues/13094

ui/app/templates/variables/new.hbs
```diff
@@ -0,0 +1,5 @@
+{{page-title "New Secure Variable"}}
+<Breadcrumb @crumb={{hash label="New" args=(array "variables.new")}} />
+<section class="section">
+  <SecureVariableForm @save={{action this.saveNewVariable}} />
+</section>
```
tgross, line 5: Missing trailing newline, which results in stuff like this:

```
$ wc -l ui/app/templates/variables/new.hbs
       4 ui/app/templates/variables/new.hbs
```

ui/app/templates/variables/variable.hbs
```diff
@@ -0,0 +1,12 @@
+{{page-title "Secure Variables: " this.model.path}}
+<Breadcrumb @crumb={{this.breadcrumb}} />
```
tgross, line 2: While testing this manually I noticed the path is all displayed as a single string. Are the paths supposed be presented as breadcrumbs or is that just not wired up yet? (Maybe that's part of https://github.com/hashicorp/nomad/issues/13094?)

tgross (approved):
LGTM once the tests are passing again. 

I've pulled this down locally and played with it agains a dev server with ACLs enabled and everything behaves as I'd expect. I've left a couple non-blocking comments.

ui/app/templates/variables/variable.hbs
philrenaud: Yep, in the above issue I'll be splitting paths into trees for better navigation. Like Github and other tree displayers, I'll be handling those non-terminal paths (not linked, single string) different than terminal ones (linked, separate from rest of string)

ui/app/adapters/variable.js
```diff
@@ -0,0 +1,46 @@
...
+  pathForType() {
+    return 'var';
+  }
+  handleResponse(_status, _headers, response) {
+    const successful = ('' + _status).startsWith(2);
+    if (successful) {
+      if (response) {
+        if (isArray(response)) {
+          response.forEach((item) => (item.ID = item.Path));
+          // TODO: do this in the serializer. Map path to id.
```
philrenaud, line 17: https://github.com/hashicorp/nomad/issues/13108

ui/app/adapters/variable.js
```diff
@@ -0,0 +1,30 @@
+import ApplicationAdapter from './application';
+import { pluralize } from 'ember-inflector';
+import classic from 'ember-classic-decorator';
+
+@classic
```
ChaiWithJai, line 5: `question` `non-blocking`: Do we need the `classic` decorator here?

https://github.com/emberjs/ember-classic-decorator

ui/app/components/secure-variable-form.js
```diff
@@ -0,0 +1,20 @@
...
+      key: '',
+      value: '',
+    });
+  }
+
+  @action
+  saveNewVariable(e) {
+    e.preventDefault();
+    const { path, keyValues } = this;
+    this.args.save({ path, keyValues });
```
ChaiWithJai, line 18: `question` `thought` `non-blocking`: This is eventually going to be an asynchronous action, correct?

We have a `Trigger` component that you might want to check out when that time comes.

ui/mirage/factories/variable.js
```diff
@@ -0,0 +1,21 @@
...
+  },
+  namespace: 'default',
+  items() {
+    return (
+      this.Items || {
+        [faker.database.column()]: faker.database.collation(),
+        [faker.database.column()]: faker.database.collation(),
+        [faker.database.column()]: faker.database.collation(),
+        [faker.database.column()]: faker.database.collation(),
+        [faker.database.column()]: faker.database.collation(),
```
ChaiWithJai, line 17: `question` `non-blocking`: Should we wrap this into a utility to create a mock object on the fly?

`question`: The property name is `items` which infers a list, but we're generating a single object from faker in the second part here.

ui/app/routes/variables/variable.js
```diff
@@ -0,0 +1,12 @@
+import Route from '@ember/routing/route';
+import withForbiddenState from 'nomad-ui/mixins/with-forbidden-state';
+import WithModelErrorHandling from 'nomad-ui/mixins/with-model-error-handling';
+
+export default class VariablesVariableRoute extends Route.extend(
+  withForbiddenState,
+  WithModelErrorHandling
+) {
+  model(params) {
+    return this.store.findRecord('variable', params.path, { reload: true });
```
ChaiWithJai, line 10: Are we missing injecting the `store` service here? Do we need to declare that this is an `async` model?

https://guides.emberjs.com/release/routing/specifying-a-routes-model/

ui/app/routes/variables/variable.js
```diff
@@ -0,0 +1,12 @@
+import Route from '@ember/routing/route';
+import withForbiddenState from 'nomad-ui/mixins/with-forbidden-state';
+import WithModelErrorHandling from 'nomad-ui/mixins/with-model-error-handling';
+
+export default class VariablesVariableRoute extends Route.extend(
+  withForbiddenState,
```
ChaiWithJai, line 6: `nit`: I'm noticing this is. `camelCase` instead of  `PascalCase`.

ui/app/routes/variables/new.js
```diff
@@ -0,0 +1,17 @@
+import Route from '@ember/routing/route';
+
+export default class VariablesNewRoute extends Route {
+  model() {
```
ChaiWithJai, line 4: `question`: This model should be `async`, no?

ui/app/routes/variables/new.js
```diff
@@ -0,0 +1,17 @@
...
+    return this.store.findAll('namespace').then(() => {
+      return this.store.createRecord('variable');
```
ChaiWithJai, line 6: `question`: Do we need to `.then` our `createRecord`? We're not reliant on whether the first promise resolves or rejects according to this logic?

It appears that we want to have all namespaces available when the route loads and then create a new variable upon entering this route.

Also, what happens if either promise returns a non `2XX` code?

Maybe this should be:

```
try {
 // findAll namespaces
 // createRecord
} catch (e) {
  // error logic
}
```

ui/app/routes/variables.js
```diff
@@ -12,7 +14,8 @@ export default class VariablesRoute extends Route.extend(withForbiddenState) {
     }
   }
   model() {
-    // TODO: Populate model from /variables
-    return {};
+    return RSVP.hash({
```
ChaiWithJai, line 17: `question` Do we want to use `RSVP.hash` here? I think the library is when we want to trigger an asynchronous callback after a promise has resolved or rejected and reference the result.

I think we might be able to get away with:

```
async model() {
 try {
   this.store.findAll('variable');
 } catch(e) {
   // error logic
 }
}
```

I think that the Promise constructor and `try/catch` syntactic sugar were introduced after ES7 which is why we were previously relying on this pattern.

ChaiWithJai (approved):
LGTM:  made some small comments.

ui/app/adapters/variable.js
philrenaud: Good question, and it appears that we do need it here (and in our equivalent serializer). This keeps convention with our other adapters, at the very least. I think it would be a good idea to get away from @classic altogether soon, but not in this PR.

ui/mirage/factories/variable.js
philrenaud: `items` is the desired API convention here, and I think it's warranted even though it's an object not an array. Another fine name might have been "entries", but either way: it's a key/value object

ui/app/routes/variables/new.js
philrenaud: Good call, and I think this is possibly not needed. This was taking a page from the `jobs/run` playbook, but by definition we'll already have namespaces on this route. Simplifying now.

ui/app/routes/variables/variable.js
philrenaud: Good catch; resolved.

PR #13076, opened 2022-05-19
Don't buffer json logs on agent startup

### LINKED ISSUE #13015: JSON agent logs buffered on startup
### Issue

I recently discovered that the Nomad agent buffers all of its logs on startup until it reaches a certain point in the initialization process: https://github.com/hashicorp/nomad/blob/bab219a8ba38d84cc4641b810339372172ccf00a/command/agent/command.go#L775-L779

This behavior made a recent attempt to debug issues with a slow-loading custom task driver plugin confusing and difficult. I wasn't able to view the logs until after the initialization was completed, and log lines were tagged with incorrect timestamps (reflecting when the buffer was flushed instead of when the log lines were originally recorded).

I couldn't find any mention of this behavior in the documentation, and only figured out what was happening by reading through the source code. (In hindsight, it seems that the `Log data will stream in below` message hints at the buffer being flushed, but this meaning not clear enough from the text alone.)

The original commit introducing this `logGate` / `GatedWriter` (21655765085fad463b25a00fcc6994116b61f3ce) was very early in the project, and there is no commit message or discussion around this behavior, so I still don't understand why this log buffering was implemented in the first place or what purpose it continues to serve. I can't think of any scenario where this kind of buffering behavior would actually be helpful.

Could this buffering behavior be simply removed, or at least made configurable, so that agent logs (including logs generated by task-driver plugins) can be flushed immediately? If there is a reason why this buffering can't or shouldn't be removed, could some additional documentation / comments detailing the reason for this behavior be added to the project?

wjordan:
Fixes #13015.

There's no reason to buffer json logs on agent startup, since logs in this format already aren't reordered (see https://github.com/hashicorp/nomad/issues/13015#issuecomment-1129157784).
This PR adds an early `logGate.Flush()` where the UI implementation is swapped out for JSON logs, so they are no longer buffered.

### DIFF
--- a/.changelog/13076.txt
+++ b/.changelog/13076.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+agent: logs are no longer buffered at startup when logging in JSON format
+```
--- a/command/agent/command.go
+++ b/command/agent/command.go
@@ -703,6 +703,8 @@ func (c *Command) Run(args []string) int {
 	// Swap out UI implementation if json logging is enabled
 	if config.LogJson {
 		c.Ui = &logging.HcLogUI{Log: logger}
+		// Don't buffer json logs because they aren't reordered anyway.
+		logGate.Flush()
 	}
 
 	// Log config files


tgross (approved):
LGTM! I've tested this out and everything works as expected for both the JSON and non-JSON cases. This will ship in Nomad 1.3.2.

PR #13077, opened 2022-05-19
keyring HTTP API

tgross:
Implements the secure variables keyring HTTP API.

### DIFF
--- a/command/agent/keyring_endpoint.go
+++ b/command/agent/keyring_endpoint.go
@@ -21,18 +21,13 @@ func (s *HTTPServer) KeyringRequest(resp http.ResponseWriter, req *http.Request)
 		case http.MethodGet:
 			return s.keyringListRequest(resp, req)
 		case http.MethodPost, http.MethodPut:
-			return s.keyringUpsertRequest(resp, req, "")
+			return s.keyringUpsertRequest(resp, req)
 		default:
 			return nil, CodedError(405, ErrInvalidMethod)
 		}
 	case strings.HasPrefix(path, "key"):
 		keyID := strings.TrimPrefix(req.URL.Path, "/v1/operator/keyring/key/")
 		switch req.Method {
-		case http.MethodPost, http.MethodPut:
-			// the request body includes the key ID, so the POST keys/ path
-			// works fine, but this gives a little flexibility to HTTP API
-			// callers and makes delete feel more consistent
-			return s.keyringUpsertRequest(resp, req, keyID)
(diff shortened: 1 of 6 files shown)


command/agent/keyring_endpoint.go
```diff
@@ -0,0 +1,140 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/hashicorp/nomad/api"
+	"github.com/hashicorp/nomad/nomad/structs"
+)
+
+// KeyringRequest is used route operator/raft API requests to the implementing
+// functions.
+func (s *HTTPServer) KeyringRequest(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
```
tgross, line 15: Note that the RFC was missing (until just now) the URL paths for this HTTP API, so I'm totally open to changing those.

command/agent/keyring_endpoint.go
```diff
@@ -0,0 +1,140 @@
...
+
+	path := strings.TrimPrefix(req.URL.Path, "/v1/operator/keyring/")
+	switch {
+	case strings.HasPrefix(path, "keys"):
+		switch req.Method {
+		case http.MethodGet:
+			return s.keyringListRequest(resp, req)
+		case http.MethodPost, http.MethodPut:
```
angrycub, line 23: My only concern about doubling the ways to access the same endpoint is that it creates another API route that doesn't provide any end-user value, could actually create confusion, and prevents us from using the verb later should we want to.

command/agent/keyring_endpoint.go
```diff
@@ -0,0 +1,140 @@
...
+			return s.keyringUpsertRequest(resp, req, "")
+		default:
+			return nil, CodedError(405, ErrInvalidMethod)
+		}
+	case strings.HasPrefix(path, "key"):
+		keyID := strings.TrimPrefix(req.URL.Path, "/v1/operator/keyring/key/")
+		switch req.Method {
+		case http.MethodPost, http.MethodPut:
```
angrycub, line 31: See above comment re:route doubling

angrycub (approved):
Everything looked good to me.  Just had the one piece of feedback about tightening up our API routing in general.

command/agent/keyring_endpoint.go
tgross: Yeah I felt conflicted about that. I'll remove the `/v1/operator/keyring/key/` version.

PR #13080, opened 2022-05-19
E2E: upgrade playwright package and container image

tgross:
The nightly playwright tests are currently failing because of a
mismatch between the expected version of Chromium and what's in the
container image. Unfortunately the previous specific tag we were using
for the container image is no longer tagged on the registry. With some
testing, I was able to find an image tag that results in a good run.

---

```
$ ./run.sh

up to date, audited 3 packages in 1s

found 0 vulnerabilities

Running 2 tests using 1 worker

  ✓  [chromium] › tests/example.spec.js:3:1 › authenticated users can see their policies (1s)
  ✓  [webkit] › tests/example.spec.js:3:1 › authenticated users can see their policies (2s)


  2 passed (6s)
```


jrasell (approved):
LGTM!

PR #13082, opened 2022-05-19
add default_ineligible configuration option

### LINKED ISSUE #10329: When making a client join a cluster, I’d like the client to be ineligible to accept job allocations until I intervene manually
### Proposal
When making a client join a cluster, I’d like the client to be ineligible to accept job allocations until I intervene manually.

### Use-cases
I'm building a user interface that will be part of a commercial product that uses Nomad to run docker jobs. It has a web interface to provide an overview of servers in the cluster. Part of that interface includes adding nodes to the cluster, which will provide a command to run to install and join a server to the cluster.

When the server joins, I'd like it to be ineligible to accept new jobs until the user intervenes to promote the server up to eligible as they may wish to perform additional configuration from within the web UI, such as making the server run specific jobs using constraints.

Thus the process should be:
1. Login to my products web UI to copy a command to run
2. Run the command on the new server
3. Wait for the server to finish running the command and for the new server to appear in my products web UI
4. View the newly added server in my products web UI, making optional configuration changes via the UI
5. Once ready, promote the server to an eligible server for running jobs.
6. Allow nomad to rebalance jobs based on its scheduling config.

I am using Consul to perform cluster discovery, so I think this would have to be a flag added to the configuration files to prevent eligibility upon initial join of the cluster. Once joined, the node would need to ignore the flag on client restarts, instead using the status recognised from within the cluster.

### Attempted Solutions
I've looked at how to use the existing join process. This would require a multi step, non-atomic process to join, mark as ineligible and potentially drain any allocated jobs. The chance of jobs being allocated are hopefully low but still possible causing unnecessary delay/overhead to bring the new node online.

I have considered the theoretical option of how I might be able to use constraints in jobs to limit to servers that I consider eligible. This would need me to track an application level eligibility outside of Nomad (e.g. a status of each node in Consul KV), and every job would need to be re-planned once the server were marked eligible. A process I don't think is going to be scalable.

### LINKED ISSUE #13081: Configure new client nodes to be ineligible by default
### Proposal

Add a new `default_ineligible` client configuration that allows new client nodes to be registered as ineligible by default.

### Use-cases

After provisioning servers meant to act as new Nomad Client nodes, it can be helpful to run a few manual checks (occasionally finding bugs in the provisioning process) before allowing new tasks to be scheduled on the new node, and making the newly-created node ineligible by default would simplify this process by preventing any allocations before the node is marked as fully ready.

### Attempted Solutions

There are a couple potential workarounds, each with some slight drawbacks:

- **Don't start the Nomad service at all until the client should be eligible.** This is okay, but removes the Nomad service initialization from the automated provisioning process and prevents us from marking the node eligible through the Nomad API (or CLI/UI) directly, which provides a bit of operational convenience.
- **Use the API (or CLI/UI) to mark the server as ineligible immediately after startup.** Also okay, but leaves open a brief scheduling window so pending tasks might still be immediately scheduled on the node before it gets marked ineligible.

wjordan:
Specifies if scheduling eligibility should be disabled by default for new client nodes.

Fixes #13081, see this issue for discussion of the use-case.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -1750,6 +1750,12 @@ func (c *Client) setupNode() error {
 		return fmt.Errorf("error syncing dynamic node metadata: %w", err)
 	}
 
+	if c.config.DefaultIneligible {
+		node.SchedulingEligibility = structs.NodeSchedulingIneligible
+	} else {
+		node.SchedulingEligibility = structs.NodeSchedulingEligible
+	}
+
 	c.config = newConfig
 	return nil
 }
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -2304,6 +2304,7 @@ func TestClient_DefaultIneligible(t *testing.T) {
 	testutil.WaitForLeader(t, s1.RPC)
 
 	c1, cleanupC1 := TestClient(t, func(c *config.Config) {
+		c.DevMode = false
 		c.DefaultIneligible = true
 		c.RPCHandler = s1
 	})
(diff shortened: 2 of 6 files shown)


client/client.go
```diff
@@ -1384,6 +1385,9 @@ func (c *Client) setupNode() error {
 	node := c.config.Node
 	if node == nil {
 		node = &structs.Node{}
+		if c.config.DefaultIneligible {
+			node.SchedulingEligibility = api.NodeSchedulingIneligible
+		}
```
jrasell, line 1390: This nil check is a safety check to avoid a panic in the event the node config object is not correctly set. If we chase the function calls from this [`client.setupNode`](https://github.com/hashicorp/nomad/blob/24eb703e740ba5e2859312979e7b23e871507781/client/client.go#L420) call we see:

- it is called by `client.NewClient` which comes from [`agent.setupClient`](https://github.com/hashicorp/nomad/blob/24eb703e740ba5e2859312979e7b23e871507781/command/agent/agent.go#L924)
- the `agent.setupClient` function includes a call to [`agent.clientConfig`](https://github.com/hashicorp/nomad/blob/24eb703e740ba5e2859312979e7b23e871507781/command/agent/agent.go#L489)
- `agent.clientConfig` calls [`convertClientConfig`](https://github.com/hashicorp/nomad/blob/24eb703e740ba5e2859312979e7b23e871507781/command/agent/agent.go#L555) which performs a number of [node struct setup](https://github.com/hashicorp/nomad/blob/24eb703e740ba5e2859312979e7b23e871507781/command/agent/agent.go#L617-L628) items

Therefore during normal and hopefully every client setup this condition is never hit and therefore the eligibility toggle is not used.

client/client_test.go
```diff
@@ -1795,3 +1796,18 @@ func TestClient_ReconnectAllocs(t *testing.T) {
 	require.False(t, invalid, "expected alloc to not be marked invalid")
 	require.Equal(t, unknownAlloc.AllocModifyIndex, finalAlloc.AllocModifyIndex)
 }
+
+func TestClient_DefaultIneligible(t *testing.T) {
```
jrasell, line 1800: This test seems like it is attempting to match the exact conditions seen within the code which as previously noted is a bit of a red herring.

jrasell (changes requested):
Hi @wjordan and thanks for taking an initial pass at this. I have added some inline comments around the functionality.

I am also curious of how this approach would work. If the value is set via a config file value, would the operator be expected to modify this value and then restart the agent?

The Nomad API also includes a method for updating [node eligibility](https://www.nomadproject.io/api-docs/nodes#toggle-node-eligibility) which operators could use. If this was the case, and the agent restarted (crash or otherwise) would the config value override what the API call did?

wjordan:
To answer your questions:

> I am also curious of how this approach would work. If the value is set via a config file value, would the operator be expected to modify this value and then restart the agent?

The intended behavior is to only apply the config value as an initial default for newly-registered nodes, so the operator would be expected to set this via a config file value _before_ starting the agent the first time.

> The Nomad API also includes a method for updating [node eligibility](https://www.nomadproject.io/api-docs/nodes#toggle-node-eligibility) which operators could use. If this was the case, and the agent restarted (crash or otherwise) would the config value override what the API call did?

No, the config value is only intended to allow configuring the default for newly-registered nodes, so it wouldn't override anything once the node has been registered.

wjordan:
Third attempt at this in abc5188bf25935026da8515808cf3a792a2914b2, just set the client node's `SchedulingEligibility` value based on `DefaultIneligible` in `setupNode`: https://github.com/hashicorp/nomad/blob/31e6a62381ba8e9b89f831bc570799c03affd118/client/client.go#L1475-L1479

This value will be sent to the server on the initial registration. Re-registering an existing node (e.g., on agent restart) should already correctly preserve the existing value, thanks to existing logic in `upsertNodeTxn` that retains `SchedulingEligibility`: https://github.com/hashicorp/nomad/blob/5b5c8535979de63abb52dc90b8ee0f08f5927dba/nomad/state/state_store.go#L896

Tyrael:
any update on this? was surprised that this isn't supported yet.

miguelwhite:
Are there any updates on this?

luizrojo:
Any updates on this change?

josegonzalez:
@wjordan would you be able to rebase the PR? If not, mind if I reissue it to get this feature merged by the hashicorp folks?

4W5ZeZPMc:
@tgross @jrasell is it possible to have a look again at this PR, this feature would be awesome

tehut:
@wjordan I did not intend to push to your fork (I assumed I wouldn't have permissions)  but if you are still following this PR and would like to  add these [test fixes](https://github.com/hashicorp/nomad/pull/27965/changes/6d6a060f8722b423e938a8a484b5e8d4fa3195d7) from the PR I opened. I'd be happy to merge your PR instead.

tehut:
@wjordan I'm going to close this as merged, thank you for the contribution!!

PR #13083, opened 2022-05-20
Update service.check.task definition to match code

josegonzalez:
Nomad errors out when attempting to specify a task for a service that uses consul connect but does not have script or gRPC checks. See https://github.com/hashicorp/nomad/blob/304d0cf5958065d14ab8b704055b1bb11d915876/nomad/structs/structs.go#L6643 for details.


Amier3 (approved):
@josegonzalez 

Thanks for raising this PR! These changes LGTM!

PR #13087, opened 2022-05-20
docs: `mount_flags` takes a slice of strings

PinkLolicorn:
The description of `mount_flags` provides incorrect example
of the accepted value format.

This fixes the issue by changing the example from a string
`ro,noatime` to a slice of strings `["ro", "noatime"]`.


tgross (approved):
Thanks @PinkLolicorn!

PR #13098, opened 2022-05-23
Secure Variables: Variables - State store, FSM, RPC

angrycub:
- Improve SecureVariables Mocks
- Add Equals() method for SecureVariables
- Split out secure variables state store impl
- SV state store implementation; tests
- Secure Variables: State Store
- Secure Variables: FSM
- Secure Variables: RPC
- Secure Variables: HTTP API


nomad/mock/mock.go
```diff
@@ -2331,3 +2338,48 @@ func SecureVariable() *structs.SecureVariable {
 	}
 	return sv
 }
+
+// SecureVariables returns a random number of secure variables between min
+// and max inclusive.
+func SecureVariables(minU, maxU uint8) MockSecureVariables {
```
tgross, line 2344: Does randomness here get us a lot for unit testing? If we do run into a case where a test fails randomly because of this, it feels like it's tricky to reproduce the test case that caused the problem.

nomad/secure_variables_endpoint.go
```diff
@@ -1,133 +1,470 @@
...
+// callable via the SecureVariables RPCs and externally via the "/v1/var{s}"
+// HTTP API.
 type SecureVariables struct {
 	srv       *Server
 	logger    hclog.Logger
 	encrypter *Encrypter
+
+	// ctx provides context regarding the underlying connection, so we can
+	// perform TLS certificate validation on internal only endpoints.
+	ctx *RPCContext
```
tgross, line 29: I don't think we have any internal-only endpoints for this RPC handler.

nomad/secure_variables_endpoint.go
```diff
@@ -1,133 +1,470 @@
...
+	reply.Index = index
 	return nil
 }
 
-func (sv *SecureVariables) List(args *structs.SecureVariablesListRequest, reply *structs.SecureVariablesListResponse) error {
-	if done, err := sv.srv.forward("SecureVariables.List", args, args, reply); done {
+// Delete removes a single secure variable, as specified by its namespace and
+// path from Nomad. This is typically called by Nomad nodes, however, in extreme
+// situations can be used via the CLI and API by operators.
+func (sv *SecureVariables) Delete(
```
tgross, line 117: I'm not so sure about this comment and associated ACL workflow... when is `SecureVariables.Delete` going to be called by Nomad nodes? Aren't all the `SecureVariables.*` RPCs driven by either humans or by the template implementation (which is indirectly through humans too)?

Same applies for `SecureVariables.Upsert` and `handleMixedAuthEndpoint`

nomad/secure_variables_endpoint.go
```diff
@@ -1,133 +1,470 @@
...
+
+	switch err {
+	case nil:
+		// If ACLs are enabled, ensure the caller has the submit-job namespace
+		// capability.
+		if aclObj != nil {
+			hasSubmitJob := aclObj.AllowNsOp(args.RequestNamespace(), acl.NamespaceCapabilitySubmitJob)
+			if !hasSubmitJob {
+				return structs.ErrPermissionDenied
+			}
```
tgross, line 137: Obviously we don't have the new ACL policy set up yet but we should leave a TODO here just for ourselves. 😀 

Same applies to `SecureVariables.Upsert`

nomad/secure_variables_endpoint.go
```diff
@@ -1,133 +1,470 @@
...
 
-	// TODO: placeholder for serialization and encryption
-	var buf bytes.Buffer
-	enc := gob.NewEncoder(&buf)
-	err := enc.Encode(args.Data.UnencryptedData)
-	if err != nil {
+	// Iterate the secure variables and validate them. Any error results in the
+	// call failing.
+	for _, i := range args.Data {
+		if err := i.Validate(); err != nil {
```
tgross, line 86: If we've failed validation we should `continue` so that we don't try to encrypt an invalid variable.

nomad/secure_variables_endpoint.go
```diff
@@ -1,133 +1,470 @@
...
+				if err != nil {
+					return err
+				}
+
+				// Ensure we never set the index to zero, otherwise a blocking query cannot be used.
+				// We floor the index at one, since realistically the first write must have a higher index.
+				if index == 0 {
+					index = 1
+				}
+				reply.Index = index
```
tgross, line 223: We've got a helper method for this block:

```suggestion
				 sv.srv.replySetIndex(state.TableSecureVariables, &reply.QueryMeta)
```

nomad/secure_variables_endpoint.go
```diff
@@ -1,133 +1,470 @@
...
+					WithNamespace: true,
+					WithID:        true,
+				},
+			)
+
+			// Set up our output after we have checked the error.
+			var svs []*structs.SecureVariableStub
+
+			// Build the paginator. This includes the function that is
+			// responsible for appending a registration to the secure variables
```
tgross, line 319:
```suggestion
			// responsible for appending a variable to the secure variables
```

nomad/secure_variables_endpoint.go
```diff
@@ -1,133 +1,470 @@
...
+			paginatorImpl, err := paginator.NewPaginator(iter, tokenizer, nil, args.QueryOptions,
+				func(raw interface{}) error {
+					sv := raw.(*structs.SecureVariable)
+					svStub := sv.Stub()
+					svs = append(svs, &svStub)
+					return nil
+				})
+			if err != nil {
+				return structs.NewErrRPCCodedf(
+					http.StatusBadRequest, "failed to create result paginator: %v", err)
```
tgross, line 330: I don't know that we want to add the HTTP status codes in the RPC handler? But I think this (and the one below) are internal errors and not bad requests if we do.

tgross (commented):
This is looking pretty good @angrycub. I've left a bunch of comments but most of it is small potatoes. The only major item that I'm unsure about is the ACLs handling with ACLs vs node secrets.

nomad/state/state_store_secure_variables.go
```diff
@@ -0,0 +1,217 @@
...
+		sv.CreateIndex = index
+		sv.CreateTime = now
+		sv.ModifyIndex = index
+		sv.ModifyTime = now
+	}
+
+	// Insert the secure variable
+	if err := txn.Insert(TableSecureVariables, sv); err != nil {
+		return fmt.Errorf("secure variable insert failed: %v", err)
+	}
```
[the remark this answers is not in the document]
tgross: Actually, I'm having a second thought on that... maybe that work should be in follow-up PRs. The `EncryptionCount` work will be more sensible once we have the encrypter to wire up, and the `SecureVariablesQuota` work can come later in general b/c we need to have ENT-side work done on that as well.

command/agent/secure_variable_endpoint.go
```diff
@@ -18,7 +18,7 @@ func (s *HTTPServer) SecureVariablesListRequest(resp http.ResponseWriter, req *h
 	}
 
 	var out structs.SecureVariablesListResponse
-	if err := s.agent.RPC("SecureVariables.List", &args, &out); err != nil {
+	if err := s.agent.RPC(structs.SecureVariablesListRPCMethod, &args, &out); err != nil {
```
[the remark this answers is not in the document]
angrycub: I lifted this from Service Registrations. I'm totally game to use the string or keep the constants, but we should probably plant a flag as a team on this pattern one way or the other moving forward.

nomad/fsm_test.go
```diff
@@ -3368,6 +3368,77 @@ func TestFSM_DeleteServiceRegistrationsByNodeID(t *testing.T) {
...
+	ws := memdb.NewWatchSet()
+	out, err := fsm.State().GetSecureVariable(ws, svs[0].Namespace, svs[0].Path)
+	assert.Nil(t, err)
+	assert.NotNil(t, out)
+
+	out, err = fsm.State().GetSecureVariable(ws, svs[1].Namespace, svs[1].Path)
+	assert.Nil(t, err)
+	assert.NotNil(t, out)
+}
+
```
angrycub, line 3441:
```suggestion
```

angrycub: @tgross  strike this whole test?

command/agent/secure_variable_endpoint.go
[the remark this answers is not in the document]
tgross: Oh interesting. Ok, I'm game to keep it then and do that going forward.

nomad/fsm_test.go
tgross: Yeah go for it

nomad/mock/mock.go
angrycub: Fair point.  I guess the thing that I appreciate about randomized mocking is that I didn't have to add extra testing code to make the values different enough to ensure that I'm not getting false successes. This mock generator could also be used to make hundreds of values for testing pagination and the UI at scale. Perhaps there is use for both approaches—make some static mocks for units and then have the random one for fuzz testing the units or just generating a large amount of faux data? I will defer to your call on this one; just wanted to add the extra context.

nomad/state/state_store_secure_variables.go
```diff
@@ -0,0 +1,217 @@
...
+	// Check if the secure variable already exists
+	existing, err := txn.First(TableSecureVariables, indexID, sv.Namespace, sv.Path)
+	if err != nil {
+		return fmt.Errorf("secure variable lookup failed: %v", err)
+	}
+
+	sv.Canonicalize()
+
+	// Setup the indexes correctly
+	now := time.Now().Round(0)
```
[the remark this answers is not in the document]
angrycub: To eliminate the realtime clock information. I was having challenges with equality checking otherwise.  It could have been a red herring though.  Let me do some testing without it.

nomad/secure_variables_endpoint.go
angrycub: I agree.  I think that we should be coding it with internal server error, but this particular code is in all of the other list endpoints that have been paginated in this way.  Should we maintain consistency with a TODO to move them all to internal errors or start doing the new (more correct) thing here and PR the other endpoints to move them to Internal Server Errors?

tgross: I took a second look at where the error is coming from: [`paginator.go#L37-L64`](https://github.com/hashicorp/nomad/blob/v1.3.0/nomad/state/paginator/paginator.go#L37-L64). The only way we can get an error from `NewPaginator` is if the filter expression is bad, which _is_ a Bad Request! The only way we can get an error from `Page()` is if the append func throws an error, which we never actually do. 

So I think what you've got here is actually fine.

nomad/state/state_store_secure_variables.go
```diff
@@ -0,0 +1,217 @@
...
+	if existing != nil {
+		exist := existing.(*structs.SecureVariable)
+		if !shouldWrite(sv, exist) {
+			*updated = false
+			return nil
+		}
```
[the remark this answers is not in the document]
tgross: Per our sidebar discussion, we'll defer this to a follow-up PR.

tgross (approved):
LGTM! Let's merge it!

PR #13103, opened 2022-05-24
warn destructive update only when count > 1

yaroot:
The warning doesn't make sense when count is 0 or 1 (destructive anyway). It's distractive and may cause users to ignore useful warnings.

### DIFF
--- a/.changelog/13103.txt
+++ b/.changelog/13103.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+cli: warn destructive update only when count is greater than 1
+```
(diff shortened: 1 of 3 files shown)


yaroot:
@Amier3 I've opened #13116 for this

lgfa29 (changes requested):
Thanks for the PR @yaroot!

Would you mind adding some test cases to [`TestJob_Warnings`](https://github.com/hashicorp/nomad/blob/3d45dcb64a7dbb0c67b3bddd7bcf3bdeee9e49c1/nomad/structs/structs_test.go#L175) to check this condition?

And also add a [`CHANGELOG` entry](https://github.com/hashicorp/nomad/blob/main/contributing/CHANGELOG.md) for this change.

yaroot:
@lgfa29 can you take a look

lgfa29 (approved):
Looks good, thank you 😄

PR #13137, opened 2022-05-26
Secure Variables: show/hide functionality when adding new values

### LINKED ISSUE #13128: Secure Variables: /new: Allow toggling value show/hide

philrenaud:
(Twig branch of #13069; resolves #13128)

- Hides values in the values field by default (shows them as stars, default password field behaviour)
- Allows the ability to toggle show/hide of those fields

### Side-effects
- Introduces HashiCorp's [Flight Icons](https://flight-hashicorp.vercel.app/) library. We'll probably move to this over x-icon across the app with time.
- Adds a target to our jsconfig to allow Glimmer type-checking to work. Otherwise, `//@ts-check` on ember files with `@action` or `get()` will throw type errors.

### Screenshots
Default
image

Toggled
image

### DIFF
--- a/ui/app/components/secure-variable-form.hbs
+++ b/ui/app/components/secure-variable-form.hbs
@@ -29,7 +29,7 @@
           />
       </label>
 
-      <label>
+      <label class="value-label">
         <span>Value
         </span>
         <Input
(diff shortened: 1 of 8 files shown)


ui/app/components/secure-variable-form.hbs
```diff
@@ -15,23 +16,42 @@
...
+
+      <button
+        class="show-hide-values button is-borderless"
+        type="button"
+        tabindex="-1"
+        {{on "click" this.toggleShowHide}}
+      >
+        <FlightIcon
+          @name="{{if this.hideValues "eye-off" "eye"}}"
+          @title="{{if this.hideValues "Show Values" "Hide Values"}}"
```
ChaiWithJai, line 48: `question` `non-blocking`: Shouldn't we remove the quotations since these are dynamic values?

ui/app/components/secure-variable-form.js
```diff
@@ -1,8 +1,25 @@
...
+import { tracked } from '@glimmer/tracking';
+import { A } from '@ember/array';
 
 export default class SecureVariableFormComponent extends Component {
-  keyValues = [{ key: '', value: '' }];
+  path = '';
+  keyValues = A([{ key: '', value: '' }]);
+
+  @tracked
+  hideValues = true;
```
ChaiWithJai, line 13: `nit` `non-blocking`: When naming boolean parameters I normally like to prefix with `is` `should` and `are`.

e.g.

`hideValues` --> `areValuesHidden`

ui/app/components/secure-variable-form.hbs
```diff
@@ -15,23 +16,42 @@
...
+          autocomplete="new-password" {{!-- prevent auto-fill --}}
+          />
       </label>
+
+      <button
+        class="show-hide-values button is-borderless"
+        type="button"
+        tabindex="-1"
+        {{on "click" this.toggleShowHide}}
+      >
```
ChaiWithJai, line 45: `thought` `non-blocking`: I noticed that we `each` over every key/value pair in the form. However, this `button` mutates state for all key/value pairs not just the associated `label` and `input` where the click occurs.

Should we convert this `button` into a `toggle` that lives on the top of the form to signify this?

A good example is the dark mode toggle on the top right of this blog:  https://overreacted.io/

ChaiWithJai (approved):
LGTM - just some minor comments.

ui/app/components/secure-variable-form.hbs
philrenaud: It's a good thought, and @juliezzhou I defer to your design direction on this one.

My thinking with placing them within the context of the form is they're visually closer to the thing they're showing/hiding, and thus require less explanation. If the eye icon were just at the top of the form alone, I'd be worried that users wouldn't grok what it's for before clicking.

PR #13167, opened 2022-05-27
keyring replication

tgross:
Replication for the secure variables keyring. Because only key
metadata is stored in raft, we need to distribute key material
out-of-band from raft replication. A goroutine runs on each server and
watches for changes to the `RootKeyMeta`. When a new key is received,
attempt to fetch the key from the leader. If the leader doesn't have
the key (which may happen if a key is rotated right before a leader
transition), try to get the key from any peer.

### DIFF
--- a/nomad/encrypter.go
+++ b/nomad/encrypter.go
@@ -2,6 +2,7 @@ package nomad
 
 import (
 	"bytes"
+	"context"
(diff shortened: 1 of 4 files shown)


angrycub (approved):
LGTM

PR #13168, opened 2022-05-27
==> amazon-ebs: E: There were unauthenticated packages and -y was use…

jaydeland:
The default process fails with this error:

==> amazon-ebs: E: There were unauthenticated packages and -y was used without --allow-unauthenticated


lgfa29 (approved):
Thanks @jaydeland!

I think it's been a while since we don't update these files so a lot may be outdated...let us know how it goes 🙂

PR #13172, opened 2022-05-30
docs: add nomad-pipeline to community tools page

waquidvp:
Nomad Pipeline is a tool I have been working on and using at work. It is a bit hacky but I thought it would be cool to have it in this list for others that might be looking for "pipelines on Nomad".


jrasell (approved):
LGTM, thanks @waquidvp!

Issue #13181, opened 2022-05-31
allow fingerprinting of classic cpu architecture names


shantanugadgil:
### Proposal
Currently the cpu is fingerprinted using the GOARCH (`amd64`, `arm64`).

Sometimes we need the "traditional" name of the CPU architecture; `i386`, `x86_64`, etc.

This proposal is to allow the Nomad agent to fingerprint the "classic" name for the CPU architecture.

### Use-cases
Software which ship using the classic CPU architecture in their filenames cannot be downloaded using the "inbuilt" variable `${attr.cpu.arch}`.

We need to write custom script (?) code to use `aarch64`, `x86_64` in the download file name depending on `amd64`, `arm64`.

### Attempted Solutions
Currently making do using a custom script and leveraging if/else using the output of `uname -m`

Please note: the name "classic" is what popped into my mind, I am open to any other named being used for this as well.

lgfa29:
Closed by #13182

PR #13182, opened 2022-05-31
fingerprint kernel architecture name

shantanugadgil:
first attempt at classic cpu architecture names while fingerprinting the cpu.

ref: #13181 

Note as mentioned in the issue, I am fine if folks can come up with a name different that "classic" :)

### DIFF
--- a/.changelog/13182.txt
+++ b/.changelog/13182.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-fingerprint: add support for detecting 'classic' names of CPU architectures (`x86_64`, `aarch64`, etc.)
+fingerprint: add support for detecting 'uname -m' for Linux (attribute `cpu.machine`)
 ```
--- a/.changelog/13182.txt
+++ b/.changelog/13182.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-fingerprint: add support for detecting 'uname -m' on Linux clients. (attribute `cpu.machine`)
+fingerprint: add support for detecting kernel architecture of clients. (attribute: `kernel.arch`)
 ```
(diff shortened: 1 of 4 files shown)


shantanugadgil:
@angrycub @lgfa29 @jrasell 

Any inputs here? I am not sure what/why the tests are failing!

lgfa29:
Vercel requires an explicit approval when deploying code from PRs, not a test failure 😄 

I will double check with team with regards to the naming here, in the meantime, would mind adding a [`CHANGELOG` entry](https://github.com/hashicorp/nomad/blob/main/contributing/CHANGELOG.md)?

Thanks for the PR!

lgfa29:
@shantanugadgil I spoke with the team and they brought up a good point, which is that these naming are really standardized and the same ISA can have even more than two common representations. For example, `386` can be `i386` but also `x86` in some places, so it would not be feasible for us to support all the different options.

One suggestion is to instead set node metadata values for the architectures that you need. How does that sound for you?

shantanugadgil:
after thinking about this a bit more ... what I essentially need/want is (pseudo-code):

```sh

if (linux) {
    attr.cpu.machine=$(uname -m)
}

```
`attr.cpu.machine` can be `attr.os.machine` as well.

lgfa29:
Ah yeah, that's a great point. To make the implementation even more robust and support other operating systems, you can read this value from the `gopsutil`, which is the library we use to fingerprint other system properties. 

More specifically, you can do this in the `host.go` fingerprinter and read the value of `KernelArch` from [`host.InfoStat`](https://pkg.go.dev/github.com/shirou/gopsutil/v3@v3.22.5/host#InfoStat), just like we do for [`kernel.version`](https://github.com/hashicorp/nomad/blob/v1.3.1/client/fingerprint/host.go#L33). So this would be a new attribute called `kernel.arch`.

And then, just one last thing missing in the PR would to document this new attribute in this page:
https://github.com/hashicorp/nomad/blob/main/website/content/docs/runtime/interpolation.mdx#node-variables-interpreted_node_vars-node-variables-

lgfa29 (approved):
Thanks @shantanugadgil!

I just pushed a commit to keep the attribute table in alphabetical order.

PR #13188, opened 2022-05-31
ui:  add service count to `StatBox`

ChaiWithJai:
This PR adds the total count of services from the `job` model to the `StatBox` in the Job Detail Overview page.


philrenaud (commented):
Should we be checking for `{{#if @job.services.length}}`  before adding this span?

I see the following and it might be confusing  to an end-user without a number:

image

Should it be named "Services" instead of "Service"?

PR #13190, opened 2022-05-31
remove end-user algorithm selection

tgross:
After internal design review, we decided to remove exposing algorithm
choice to the end-user for the initial release. We'll solve nonce
rotation by forcing rotations automatically on key GC (in a core job,
not included in this changeset). Default to AES-256 GCM for the
following criteria:

* faster implementation when hardware acceleration is available
* FIPS compliant
* implementation in pure go
* post-quantum resistance

### DIFF
--- a/api/keyring.go
+++ b/api/keyring.go
@@ -21,7 +21,6 @@ func (c *Client) Keyring() *Keyring {
 type EncryptionAlgorithm string
 
 const (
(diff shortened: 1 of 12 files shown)


tgross:
@angrycub I've added a second commit onto this to fix a bug I discovered in how we were decoding keys. The keys ended up being the right size but truncated (just padded out with null bytes). We would've discovered this pretty quickly once we wired up the encryption, but no reason to have it sitting around buggy in the meantime.

PR #13195, opened 2022-06-01
provide state store query for variables by key ID

tgross:
The core jobs to garbage collect unused keys and perform full key
rotations will need to be able to query secure variables by key ID for
efficiency. Add an index to the state store and associated query
function and test.

### DIFF
--- a/nomad/state/schema.go
+++ b/nomad/state/schema.go
@@ -1230,8 +1230,69 @@ func secureVariablesTableSchema() *memdb.TableSchema {
 					},
 				},
 			},
+			"key_id": {
+				Name:         "key_id",
+				AllowMissing: false,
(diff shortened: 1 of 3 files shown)


nomad/state/schema.go
```diff
@@ -1230,10 +1230,71 @@ func secureVariablesTableSchema() *memdb.TableSchema {
 					},
 				},
 			},
+			"key_id": {
```
angrycub, line 1233: Is there any benefit in making this an exported constant?

angrycub (approved):
LGTM, one non-blocking question, but good to go.

nomad/state/schema.go
tgross: Done in 2fa6a473f9b772c62f13f67ec473f4db197d330f

PR #13196, opened 2022-06-01
refactor index threshold calculation for core GC jobs

tgross:
Almost all GC jobs check the index of the objects being GC'd to see if
they're older than a configured threshold. This code was repeated six
times in `CoreScheduler` with only logging changes, so it seems safe
to extract it as its own method.

Ran into this while working on 2 new core jobs for Secure Variables, so
this changeset will eliminate the need for 2 more repetitions of this function.

### DIFF
--- a/nomad/core_sched.go
+++ b/nomad/core_sched.go
@@ -99,20 +99,8 @@ func (c *CoreScheduler) jobGC(eval *structs.Evaluation) error {
 		return err
 	}
 
-	var oldThreshold uint64
-	if eval.JobID == structs.CoreJobForceGC {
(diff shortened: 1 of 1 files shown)


tgross:
> One thing I might add are functions like `c.getJobThreshold`, `c.getNodeThreshold`, `c.getEvalThreshold` that bundle up the constant parameters so that someone doesn't accidentally goof them up in a call

Yeah I definitely considered that, but those functions would only ever have one caller each, and any new GC job would have its own helper function bundling the constants. So we'd be exposed to goofing it up on each new GC job anyways.

PR #13202, opened 2022-06-01
Secure Variables: Build a path tree and traverse it at /variables/*path

### LINKED ISSUE #13094: UI: Treat Secure Variables index page list like a branching tree
- Terminal parts of paths (the part after the last `/`) should be links and be indented appropriately
- Non-terminal multi-child parts of the path should show up as links to their subdirectory
- Non-terminal parts of the path should show up as plain text

philrenaud:
Resolves #13094 

- Allows for easier traversal within the `/variables` route 
- Compacts "empty" directories to provide a structure like the following:

Given a variables list of:
```
["a/b/c/foo0",
"a/b/c/bar1",
"a/b/c/d/e/foo2",
"a/b/c/d/e/bar3",
"a/b/c/d/e/f/foo4",
"a/b/c/d/e/f/g/foo5",
"a/b/c/x/y/z/foo6",
"a/b/c/x/y/z/bar7",
"a/b/c/x/y/z/baz8",
"w/x/y/foo9",
"w/x/y/z/foo10",
"w/x/y/z/bar11"]
```

render a path structure, showing one level of depth at a time, like:
```
DIR a/b/c/
  VAR foo0
  VAR bar1
  DIR d/e/
    VAR foo2
    VAR bar3
    DIR f/
      VAR foo4
      VAR bar5
      DIR g/
        VAR foo6
        VAR bar7
        VAR baz8
DIR w/x/y/
  VAR foo9
  DIR z/
    VAR foo10
    VAR bar11

```
Note: since the `/a` and `/a/b` directories have no variables, they are compacted down to a single layer of depth, as far as the tree is concerned. This helps us eliminate user clicks in situations where all variables are structured deeply.

image

## TODO
- [ ] Tests for /path and utils/path-tree
- [ ] Table styling

#TO-MAYBE-DO
- [ ] Messaging on compacted-but-otherwise-empty paths (like `/a` in the `/a/b/c` chain mentioned above) currently has a "There are no variables here" messaging. Could un-compact the tree for scenarios like this instead.
- [ ] Make the trie algorithm in path-tree.js more efficient / use an insertion pattern rather than deletion


ui/app/routes/variables.js
```diff
@@ -13,9 +14,11 @@ export default class VariablesRoute extends Route.extend(WithForbiddenState) {
       this.router.transitionTo('/jobs');
     }
   }
-  model() {
+  async model() {
+    const variables = await this.store.findAll('variable');
     return RSVP.hash({
-      variables: this.store.findAll('variable'),
+      variables,
+      pathTree: new PathTree(variables),
```
philrenaud, line 21: the /variables route now returns both a PathTree (traversal structure) and the variables themselves (with all metadata and info, like time created). 

There is a design decision that could be made to combine these here, but it seems worthwhile to keep them separate.

ui/app/routes/variables/new.js
```diff
@@ -1,10 +1,11 @@
 import Route from '@ember/routing/route';
 
 export default class VariablesNewRoute extends Route {
-  model() {
-    return this.store.createRecord('variable');
+  model(params) {
+    return this.store.createRecord('variable', { path: params.path });
```
philrenaud, line 5: Nice little shortcut for when you are creating a new variable from within an existing directory; it pre-fills the path field and focuses your cursor after the last slash.

ui/app/routes/variables/path.js
```diff
@@ -0,0 +1,12 @@
+import Route from '@ember/routing/route';
+
+export default class VariablesPathRoute extends Route {
+  model({ absolutePath }) {
+    let treeAtPath = this.modelFor('variables').pathTree.findPath(absolutePath);
```
philrenaud, line 5: Originally written to let us do recursive tree traversal and can still be expanded to allow for that (we could have `<PathTree>` call itself with its children).

ui/app/routes/variables/variable.js
```diff
@@ -7,6 +7,6 @@ export default class VariablesVariableRoute extends Route.extend(
   WithModelErrorHandling
 ) {
   model(params) {
-    return this.store.findRecord('variable', params.path);
+    return this.store.findRecord('variable', decodeURIComponent(params.path));
```
philrenaud, line 10: Handle the possibility of a user pasting both `my/path/to` and `my%2fpath%2fto` correctly

ui/app/styles/components/secure-variables.scss
```diff
@@ -28,3 +28,13 @@
...
+
+.path-tree {
+  margin-left: 5px;
+  padding-left: 5px;
+  border-left: 1px dotted grey;
+  a {
+    font-weight: bold;
+    margin-left: 11px;
+  }
+}
```
philrenaud, line 40: Temporary / still unstyled

ui/app/utils/path-tree.js
```diff
@@ -0,0 +1,123 @@
+// @ts-check
```
philrenaud, line 1: The create-a-tree-from-paths-array class. 

Introduces a few JSDoc annotations to the codebase that I hope we can expand as well.

ui/tests/unit/utils/path-tree-test.js
```diff
@@ -0,0 +1,10 @@
+import pathTree from 'nomad-ui/utils/path-tree';
+import { module, test } from 'qunit';
+
+module('Unit | Utility | path-tree', function () {
+  // TODO: Replace this with your real tests.
```
philrenaud, line 5: Big TODO here!

ui/app/routes/variables/variable.js
ChaiWithJai: `nit`:  Should we extract this functionality into a function that creates a derived state variable.

Example:

```
const sanitizedPath = decodeURIComponent(path);

async model({path}) {
  return this.store.findRecord('variable', path);
}
```

ui/app/templates/variables/path.hbs
```diff
@@ -0,0 +1,48 @@
...
+{{/each}}
+  <section class="section">
+    <div class="toolbar">
+      <div class="toolbar-item is-right-aligned is-mobile-full-width">
+        <div class="button-bar">
+        {{!-- TODO: make sure qpNamespace persists to here --}}
+        {{#if (can "create variable" namespace=this.qpNamespace)}}
+          <LinkTo
+            @route="variables.new"
+            @query={{hash namespace=this.qpNamespace path=(concat this.model.absolutePath "/")}}
```
ChaiWithJai, line 13: `blocking`:  This breaks from an architectural pattern that I was hoping to persist from [RFC-130:  Namespaces as Identifiers](https://docs.google.com/document/d/1ssx1q0k1c3wYhSBS0DIY08njk3jfUdy2cODhgkC7c24/edit).

Is it out of the question to serialize namespaces into the identifier to avoid this pattern?

I could see this causing lurking bugs and confusion on how we handle namespaces and potentially preventing us from setting namespaces in the configuration as mentioned in the RFC comments.

ui/tests/unit/utils/path-tree-test.js
ChaiWithJai: Yup! This is going to be super helpful for documenting this feature.

ui/app/utils/path-tree.js
```diff
@@ -0,0 +1,123 @@
...
+                });
+            acc.root.children[cur.path].absolutePath = cur.path;
+          } else {
+            acc.root.files
+              ? acc.root.files.push({ name, absoluteFilePath })
+              : (acc.root.files = [{ name, absoluteFilePath }]);
+          }
+          return acc;
+        },
+        { root: { files: [], children: {}, absolutePath: '' } }
```
ChaiWithJai, line 98: `question`:  Should we extract this logic out so its more readable/testable?

ui/app/utils/path-tree.js
```diff
@@ -0,0 +1,123 @@
+// @ts-check
+
+// eslint-disable-next-line no-unused-vars
+import VariableModel from '../models/variable';
+// eslint-disable-next-line no-unused-vars
+import MutableArray from '@ember/array/mutable';
```
ChaiWithJai, line 6: `question`:  Are we actually relying on this array being mutable/reacting to updates?

ChaiWithJai (approved):
LGTM. My only blocking feedback is regarding namespaces being used as query parameters. Everything else is good to go IMO.

ui/app/utils/path-tree.js
philrenaud: We're relying on the **type** specifically; it's the param for our constructor. This allows us to do `<eventualArrName>.pushObject()` etc. without our typechecker saying that `pushObject doesn't exist on type Array`

ui/app/templates/variables/path.hbs
philrenaud: After a conversation: namespaces don't apply in the context of a .new page any more than they do for a .index page, so keeping this as is for now.

ui/app/utils/path-tree.js
ChaiWithJai: `question`: Where do we call `pushObject`? I just checked the diff and missed it.

`thought` `non-blocking`: Is there any concern that this would accidentally mutate? When invoking `PathTree` would we want to want to pass this in as `new PathTree(variables.removeArrayObserver)` to prevent a re-computation that would accidentally cause a re-render? My days with the `useEffect` hook have left me totally scarred so I'm just curious here.

PR #13204, opened 2022-06-02
Update hashistack.tf

jaydeland:
Replace the template_file with the templatefile() funtion. This is needed for arm64.


tgross (changes requested):
Hi @jaydeland! The changes you've made here don't follow from the PR description (or at least not in a way that's obvious to me 😀 ). Can you clarify the reason for these changes?

tgross (approved):
LGTM

PR #13218, opened 2022-06-03
core: reschedule evicted batch job when resources become available

shoenig:
This is second attempt at the cherry-pick of https://github.com/hashicorp/nomad/pull/13209 onto `release/1.2.x` (where I botched the rebase and got confused)

### DIFF
--- a/scheduler/reconcile_util_test.go
+++ b/scheduler/reconcile_util_test.go
@@ -2,7 +2,7 @@ package scheduler
 
 import (
 	"testing"
-
+	
 	"github.com/hashicorp/nomad/ci"
 	"github.com/hashicorp/nomad/helper"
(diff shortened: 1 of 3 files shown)


scheduler/reconcile_util_test.go
```diff
@@ -2,7 +2,8 @@ package scheduler
 
 import (
 	"testing"
-
+	"time"
```
lgfa29, line 5: CI is complaining that this import is not used.

lgfa29 (approved):
Just need to fix that `time` import 👍

scheduler/reconcile_util_test.go
shoenig: thanks :man_facepalming:

PR #13222, opened 2022-06-03
ui:  create dummy template for `jobs/:jobId/services` route

ChaiWithJai:
This PR creates a dummy template. Notably, missing from the table is the `Number of Allocations` column. We're currently unsure if this column is tenable given the API schema for services lists `AllocID:  string` instead of `Allocations: string[]`. The API schema per NMD-130 suggests there's a one-to-one relationship which will require some clarity.

Expected:
image

Actual:
image

### DIFF
--- a/ui/app/templates/jobs/job/services.hbs
+++ b/ui/app/templates/jobs/job/services.hbs
@@ -1,0 +1,54 @@
+{{page-title "Job " this.model.name " services"}}
+<JobSubnav @job={{this.model}} />
+<section class="section">
+  {{#if this.model.services.length}}
+    <ListTable
+      @source={{this.model.services}}
+      @sortProperty={{this.sortProperty}}
+      @sortDescending={{this.sortDescending}}
+      @class="with-foot" as |t|
+    >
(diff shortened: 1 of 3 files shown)


ui/app/templates/jobs/job/services.hbs
```diff
@@ -0,0 +1,54 @@
+{{page-title "Job " this.model.name " services"}}
```
philrenaud, line 1: In the interest of simplifying things, you could consider a template with aliases to your model properties here. See for example how `controllers/jobs/job/clients` handles this: https://github.com/hashicorp/nomad/blob/main/ui/app/controllers/jobs/job/clients.js#L64-L65

You would have `<JobSubnav @job={{this.job}} />` and a page title of `"Job" this.job.name` which is a little more clear to future coders.

But the biggest benefit, means that if we ever wanted to use this template to list the services of a job from outside of that job context (say, a "Services meta page" or something) we could make a change to this file's controller instead of remapping the route model.

philrenaud (approved):

PR #13223, opened 2022-06-03
workload identity

tgross:
In order to support implicit ACL policies for tasks to get their own
secrets, each task would need to have its own ACL token. This would
add extra raft overhead as well as new garbage collection jobs for
cleaning up task-specific ACL tokens. Instead, Nomad will create a
workload Identity Claim for each task.

An Identity Claim is a JSON Web Token (JWT) signed by the server’s
private key and attached to an Allocation at the time a plan is
applied. The encoded JWT can be submitted as the X-Nomad-Token header
to replace ACL token secret IDs for the RPCs that support identity
claims.

Whenever a key is is added to a server’s keyring, it will use the key
as the seed for a Ed25519 public-private private keypair. That keypair
will be used for signing the JWT and for verifying the JWT.

This implementation is a ruthlessly minimal approach to support the
secure variables feature. When a JWT is verified, the allocation ID
will be checked against the Nomad state store, and non-existent or
terminal allocation IDs will cause the validation to be rejected. This
is sufficient to support the secure variables feature at launch
without requiring implementation of a background process to renew
soon-to-expire tokens.

---

Best reviewed commit-by-commit. Note this is targeting the `secure-variables` 
feature branch so it's relying on work already reviewed there.

### DIFF
--- a/nomad/encrypter.go
+++ b/nomad/encrypter.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/aes"
 	"crypto/cipher"
+	"crypto/ed25519"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
@@ -15,6 +16,7 @@ import (
 	"sync"
 	"time"
 
+	jwt "github.com/golang-jwt/jwt/v4"
 	log "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-msgpack/codec"
 	"golang.org/x/time/rate"
@@ -27,15 +29,22 @@ const nomadKeystoreExtension = ".nks.json"
 
 // Encrypter is the keyring for secure variables.
 type Encrypter struct {
-	lock         sync.RWMutex
-	keys         map[string]*structs.RootKey // map of key IDs to key material
-	ciphers      map[string]cipher.AEAD      // map of key IDs to ciphers
+	srv          *Server
 	keystorePath string
+
+	keyring map[string]*keyset
+	lock    sync.RWMutex
+}
+
+type keyset struct {
+	rootKey    *structs.RootKey
+	cipher     cipher.AEAD
+	privateKey ed25519.PrivateKey
 }
 
 // NewEncrypter loads or creates a new local keystore and returns an
 // encryption keyring with the keys it finds.
-func NewEncrypter(keystorePath string) (*Encrypter, error) {
+func NewEncrypter(srv *Server, keystorePath string) (*Encrypter, error) {
 	err := os.MkdirAll(keystorePath, 0700)
 	if err != nil {
 		return nil, err
@@ -44,14 +53,14 @@ func NewEncrypter(keystorePath string) (*Encrypter, error) {
 	if err != nil {
 		return nil, err
 	}
+	encrypter.srv = srv
 	return encrypter, nil
 }
 
 func encrypterFromKeystore(keystoreDirectory string) (*Encrypter, error) {
 
 	encrypter := &Encrypter{
-		ciphers:      make(map[string]cipher.AEAD),
-		keys:         make(map[string]*structs.RootKey),
+		keyring:      make(map[string]*keyset),
 		keystorePath: keystoreDirectory,
 	}
 
@@ -107,6 +116,57 @@ func (e *Encrypter) Encrypt(unencryptedData []byte, keyID string) []byte {
 	return unencryptedData
 }
 
+// keyIDHeader is the JWT header for the Nomad Key ID used to sign the
+// claim. This name matches the common industry practice for this
+// header name.
+const keyIDHeader = "kid"
+
+// SignClaims signs the identity claim for the task and returns an
+// encoded JWT with both the claim and its signature
+func (e *Encrypter) SignClaims(claim *structs.IdentityClaims) (string, error) {
+
+	keyset, err := e.activeKeySet()
+	if err != nil {
+		return "", err
+	}
+
+	token := jwt.NewWithClaims(&jwt.SigningMethodEd25519{}, claim)
+	token.Header[keyIDHeader] = keyset.rootKey.Meta.KeyID
+
+	tokenString, err := token.SignedString(keyset.privateKey)
+	if err != nil {
+		return "", err
+	}
+
+	return tokenString, nil
+}
+
+// VerifyClaim accepts a previously-signed encoded claim and validates
+// it before returning the claim
+func (e *Encrypter) VerifyClaim(tokenString string) (*structs.IdentityClaims, error) {
+
+	token, err := jwt.ParseWithClaims(tokenString, &structs.IdentityClaims{}, func(token *jwt.Token) (interface{}, error) {
+		if _, ok := token.Method.(*jwt.SigningMethodEd25519); !ok {
+			return nil, fmt.Errorf("unexpected signing method: %v", token.Header["alg"])
+		}
+		raw := token.Header[keyIDHeader]
+		if raw == nil {
+			return nil, fmt.Errorf("missing key ID header")
+		}
+		keyID := raw.(string)
+		keyset, err := e.keysetByID(keyID)
+		if err != nil {
+			return nil, err
+		}
+		return keyset.privateKey.Public(), nil
+	})
+
+	if claims, ok := token.Claims.(*structs.IdentityClaims); ok && token.Valid {
+		return claims, nil
+	}
+	return nil, err
+}
+
 // Decrypt takes an encrypted buffer and then root key ID. It extracts
 // the nonce, decrypts the content, and returns the cleartext data.
 func (e *Encrypter) Decrypt(encryptedData []byte, keyID string) ([]byte, error) {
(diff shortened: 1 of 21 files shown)


nomad/secure_variables_endpoint.go
```diff
@@ -367,3 +367,56 @@ func (sv *SecureVariables) decrypt(v *structs.SecureVariable) error {
...
+		// Perform our ACL validation. If the object is nil, this means ACLs
+		// are not enabled, otherwise trigger the allowed namespace function.
+		if aclObj != nil {
+			if !aclObj.AllowNsOp(args.RequestNamespace(), cap) {
+				return structs.ErrPermissionDenied
+			}
+		}
+		return nil
+	}
+
```
schmichael, line 387: We may want to only continue when `err == ErrTokenNotFound`? We should at least log when `err != ErrTokenNotFound` as otherwise we'd be silently ignoring some error conditions that indicate pretty weird states (such as being unable to compile policies). I also don't want to assume any future error return values in the `Server.ResolveToken` path are safe to silently drop.

nomad/secure_variables_endpoint.go
```diff
@@ -367,3 +367,56 @@ func (sv *SecureVariables) decrypt(v *structs.SecureVariable) error {
...
+	// Attempt to verify the token as a JWT with a workload
+	// identity claim
+	claim, err := sv.srv.ResolveClaim(args.AuthToken)
+	if err != nil {
+		return structs.ErrPermissionDenied
+	}
+
+	store := sv.srv.fsm.State()
```
schmichael, line 395:
```suggestion
	store := sv.srv.fsm.State().Snapshot()
```

I think we want to always use a snapshot for readonly cases to (a) ensure a consistent view of state and (b) prohibit changes.

`Server.ResolveToken` _also_ takes a snapshot, so I wonder if they could/should be combined? Memdb snapshots are awfully cheap, so it's probably not worth worrying about. This does mean token resolution may observe multiple states in the course of authenticating a single token, but I can't think of any inconsistencies that could cause.

nomad/secure_variables_endpoint.go
```diff
@@ -367,3 +367,56 @@ func (sv *SecureVariables) decrypt(v *structs.SecureVariable) error {
...
+	alloc, err := store.AllocByID(nil, claim.AllocationID)
+	if err != nil || alloc == nil || alloc.Job == nil {
+		return structs.ErrPermissionDenied
```
schmichael, line 398: Trace log and/or metric perhaps? It would be nice to have _some_ way to observe if this is occurring.

Same comment for all of the other errors below.

nomad/secure_variables_endpoint.go
```diff
@@ -367,3 +367,56 @@ func (sv *SecureVariables) decrypt(v *structs.SecureVariable) error {
...
+	expect := []string{"jobs", alloc.Job.ID, alloc.TaskGroup, claim.TaskName}
+	if len(parts) > len(expect) {
+		return structs.ErrPermissionDenied
+	}
+
+	for idx, part := range parts {
+		if part != expect[idx] {
+			return structs.ErrPermissionDenied
+		}
+	}
```
schmichael, line 420: Some comments around the purpose of pathOrPrefix either here and/or in the func definition would be helpful. It looks like the purpose is to lock token access down to just secure variable paths for this task?

Also I think this logic is equivalent to `if pathOrPrefix != strings.Join(expect) { return structs.ErrPermissionDenied }` although I think you manage to dodge making any memory allocations.

nomad/structs/structs.go
```diff
@@ -10273,6 +10279,40 @@ func (a *Allocation) Reconnected() (bool, bool) {
...
+func (a *Allocation) ToIdentityClaims() *IdentityClaims {
+	now := jwt.NewNumericDate(time.Now().UTC())
+	return &IdentityClaims{
+		Namespace:    a.Namespace,
+		AllocationID: a.ID,
+		RegisteredClaims: jwt.RegisteredClaims{
+			// TODO: in Nomad 1.5.0 we'll have a refresh loop to
+			// prevent allocation identities from expiring before the
+			// allocation is terminal. Once that's implemented, add an
+			// ExpiresAt here ExpiresAt: &jwt.NumericDate{},
```
schmichael, line 10291: This is still a maybe to me as JWTs can be fully stateless without expiration or revocation:

If the signature is signed by a valid key and the alloc is alive, the jwt is valid!

If a jwt gets leaked, kill the alloc, you can't trust its identity!

Although without expiration no one can do offline attenuation: you have to hit a Nomad API to validate the JWT. If we're ok with tokens living _past_ an allocation's termination, then expiration means any valid Nomad JWT is valid without having to ask Nomad about it.

I don't know though. Definitely a critical thing to think through.

nomad/structs/structs.go
```diff
@@ -10273,6 +10279,40 @@ func (a *Allocation) Reconnected() (bool, bool) {
...
+	}
+	return claims
+}
+
+// IdentityClaims are the input to a JWT identifying a workload. It
+// should never be serialized to msgpack unsigned.
+type IdentityClaims struct {
+	Namespace    string `json:"nomad_namespace"`
+	AllocationID string `json:"nomad_allocation_id"`
+	TaskName     string `json:"nomad_task"`
```
schmichael, line 10311: Hm, I wonder how much we need to get these right on the first try since they're going to live as long as their allocation and allocation's might live a long time. (An argument for expiration I guess!)

If we want to be safe I think `job` will be necessary for user services to find them useful to do auth like:

> `Task api of Job user-api` is allowed to write to `Task db of Job user-db`

Where `write` is a capability encoded in the user's code doing the jwt-based auth.

I suspect region might also be useful:

> `Task api of Job user-api in Region Z` is allowed to write to `Task db of Job user-db in Region Z`

Unsure about that one though.

At one point I was thinking ClusterID could even be included to ensure claims are universally unique, but then I realized that's the job of the signing key. I don't think we want people trying to validate the signing key as well as the ClusterID (which we don't generally expose anyway) to ensure the JWT is minted by their trusted Nomad cluster. I don't think ClusterID would add anything useful.

schmichael (commented):
:tada: :tada: :tada: 

Very excited about this and it's looking really good! I think the general approach of tacking the tokens onto allocations at schedule-time is good. Makes the sensitive nature a bit trickier to handle as we use the same Allocation struct all over the place, but the json tag covers every case of exposure I can think of.

Being able to commit them atomically with the Allocation makes their statefulness way easier to handle. Very little overhead from a raft/storage perspective, no concerns about consistency/skew with the Allocation object itself. Allocation objects also already have multiple writers (eg servers writes DesiredStatus, clients write ClientStatus), so if we end up mutating these keys on the server, that's already a pattern we follow elsewhere.

nomad/structs/structs.go
tgross: > Although without expiration no one can do offline attenuation: you have to hit a Nomad API to validate the JWT. If we're ok with tokens living past an allocation's termination, then expiration means any valid Nomad JWT is valid without having to ask Nomad about it.

Yeah my thinking here is that one of our big internally-proposed use cases is for Vault, and being able to get just the public key from Nomad and use it for many token validations would cut down on cross-traffic.

nomad/plan_apply.go
```diff
@@ -237,6 +237,11 @@ func (p *planner) applyPlan(plan *structs.Plan, result *structs.PlanResult, snap
 		// to approximate the scheduling time.
 		updateAllocTimestamps(req.AllocsUpdated, now)
 
+		err := p.signAllocIdentities(plan.Job, req.AllocsUpdated)
+		if err != nil {
+			return nil, err
+		}
+
```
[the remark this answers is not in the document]
tgross: I like where you're going with that, but there's a complexity/reliability tradeoff: the leader is the only instance guaranteed to have the key material. We use raft to replicate the key metadata but then followers need to pull the key material separately to write to their on-disk keystore and load into the keyring (ref https://github.com/hashicorp/nomad/pull/13167). So during a key rotation, all the schedulers on the followers would hang while waiting for the key material to get replicated.

nomad/secure_variables_endpoint.go
tgross: > `Server.ResolveToken` also takes a snapshot, so I wonder if they could/should be combined?

I didn't want to tie allocation lifecycle, which we currently only care about for this RPC endpoint, in with the `ResolveToken` used everywhere else. At least not yet: if we end up making workload identity more broadly useful across RPCs, I think I'd agree but I didn't want to make that call prematurely.

nomad/secure_variables_endpoint.go
tgross: Good catch! Will fix.

nomad/secure_variables_endpoint.go
```diff
@@ -180,13 +179,14 @@ func (sv *SecureVariables) List(
 		return sv.listAllSecureVariables(args, reply)
 	}
 
-	if aclObj, err := sv.srv.ResolveToken(args.AuthToken); err != nil {
+	// FIXME: Temporary ACL Test policy. Update once implementation complete
+	err := sv.handleMixedAuthEndpoint(args.QueryOptions,
+		acl.NamespaceCapabilitySubmitJob, args.Prefix)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
tgross: In the changes I had to make to the service discovery RPC's version of this same code, I checked specifically whether the token was UUID shaped before trying to verify it as a claim so as not to change the response. We can do that here as well.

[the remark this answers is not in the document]
tgross: Resolved in 23af20e29b2398437d86c45487d7bccf7f639cb3

nomad/secure_variables_endpoint.go
tgross: Resolved in 23af20e29b2398437d86c45487d7bccf7f639cb3

nomad/secure_variables_endpoint.go
tgross: Resolved in 23af20e29b2398437d86c45487d7bccf7f639cb3

nomad/secure_variables_endpoint.go
tgross: Resolved in 23af20e29b2398437d86c45487d7bccf7f639cb3

nomad/structs/structs.go
tgross: Added JobID in 23af20e29b2398437d86c45487d7bccf7f639cb3

nomad/encrypter.go
```diff
@@ -150,22 +215,46 @@ func (e *Encrypter) addCipher(rootKey *structs.RootKey) error {
...
+		privateKey: privateKey,
+	}
 	return nil
 }
 
 // GetKey retrieves the key material by ID from the keyring
 func (e *Encrypter) GetKey(keyID string) ([]byte, error) {
+	keyset, err := e.keysetByID(keyID)
+	if err != nil {
+		return []byte{}, err
```
schmichael, line 234: Just a nitpick. If there's a reason it needs to be an initialized slice best to comment it.

```suggestion
		return nil, err
```

nomad/plan_apply.go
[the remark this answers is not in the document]
schmichael: > the leader is the only instance guaranteed to have the key material

I didn't realize this! Thanks.

nomad/secure_variables_endpoint.go
```diff
@@ -367,3 +368,95 @@ func (sv *SecureVariables) decrypt(v *structs.SecureVariable) error {
...
+	if err != nil {
+		metrics.IncrCounter([]string{
+			"nomad", "secure_variables", "invalid_allocation_identity"}, 1)
+		sv.logger.Trace("allocation identity was not valid", "error", err)
+		return structs.ErrPermissionDenied
+	}
+
+	err = sv.authValidatePrefix(alloc, claim.TaskName, pathOrPrefix)
+	if err != nil {
+		sv.logger.Trace("allocation identity did not have permission for path")
```
schmichael, line 411:
```suggestion
		sv.logger.Trace("allocation identity did not have permission for path", "error", err)
```

Right?

schmichael (approved):

nomad/secure_variables_endpoint.go
tgross: The error is _always_ `ErrPermissionDenied`, which is a little redundant. But given this is a trace log, it can't hurt to make the logger a little future-proof for new errors.

PR #13227, opened 2022-06-06
state: only update index on change when deleting evals.

jrasell:
When deleting evaluations and allocations during a reap event, the
index table entries for evals and allocs was updated irregardless
of whether changes were made.

This change modifies the state logic so that the index table is
only modified when the corresponding table has actually been
modified. Along with matching expected behaviour, this change has
the potential to reduce the number of times blocking queries will
return without any real state change.


tgross (approved):
LGTM!

shoenig (approved):

PR #13257, opened 2022-06-06
Implement HTTP search API for Variables

angrycub:
- Add Path only index for SecureVariables
- Add GetSecureVariablesByPrefix; refactor tests
- Add search for SecureVariables

### DIFF
--- a/nomad/search_endpoint.go
+++ b/nomad/search_endpoint.go
@@ -35,6 +35,7 @@ var (
 		structs.Plugins,
 		structs.Volumes,
 		structs.ScalingPolicies,
+		structs.SecureVariables,
 		structs.Namespaces,
 	}
 )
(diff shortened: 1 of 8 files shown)


nomad/search_endpoint.go
```diff
@@ -382,6 +385,15 @@ func getResourceIter(context structs.Context, aclObj *acl.ACL, namespace, prefix
...
 		return memdb.NewFilterIterator(iter, nsCapFilter(aclObj)), nil
+	case structs.SecureVariables:
+		iter, err := store.GetSecureVariablesByPrefix(ws, prefix)
+		if err != nil {
+			return nil, err
+		}
+		if aclObj == nil {
+			return iter, nil
+		}
+		return memdb.NewFilterIterator(iter, nsCapFilter(aclObj)), nil
```
tgross, line 396: I think this check will always currently fail because we don't have a `nsCapFilter` case for our type. Maybe we should drop that in with the job ACL check, just as we've done for the other Storage API RPCs? That way we can at least have a working build that we can bench test until we've created the ACL policy.

tgross (approved):
LGTM!

angrycub:
Rebased on sv-split-enc-dec so going to hold it for that merge first.

PR #13258, opened 2022-06-06
SV HTTP API: skip empty variables

angrycub:
If an inbound payload doesn't contain `Items` collection, we should skip writing it and return an error.


tgross (approved):
LGTM!

PR #13265, opened 2022-06-07
driver/docker: Don't pull InfraImage if it exists

### LINKED ISSUE #10318: Allow using a local infra_image
### Proposal
Allow specifying a local infra_image under docker config in order to avoid pulling the image.

### Use-cases
For my current usecase, I would like to avoid having nomad reach out to pull images at all if possible, so I was hoping to use a local version of the infra image instead of Nomad pulling it.

I can see from the code that right now its always trying to pull the infra_image when setting up the network, and you cant specify a local image like you can for task images.

### LINKED ISSUE #11014: network bridge mode,Can I use a local mirror for pause-amd64?
my config

```
    network {
      mode = "bridge"
      port "http" {
        to = 80
      }
      port "https" {
        to=443
      }
    }
```

my local image

```
gcr.io/google_containers/pause-amd64                              3.1         da86e6ba6ca1   3 years ago    742kB
```

but i get error,Because i can't access gcr.io

```
failed to setup alloc: pre-run hook "network" failed: failed to create network for alloc: Failed to pull `gcr.io/google_containers/pause-amd64:3.1`: API error (500): Get https://gcr.io/v2/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

tbehling:
This PR tweaks the behavior of Nomad's Docker driver during network setup.

When Nomad sets up a new Docker network, it establishes the network namespace by running a "no-op" container, internally called `InfraImage`.  By default the container's image is `gcr.io/google_containers/pause-amd64:3.1` (where `amd64` is the GOARCH).

Every time a network is set up, that container image is pulled from the Google Container Registry (or alternate image, if configured).  In a cluster with a high rate of Nomad job creation, the Docker pull is seen to intermittently fail (about 0.4%), perhaps due to rate limiting or normal transient network problems.

Because this image is versioned (tagged `3.1`, currently), it is not necessary to run a Docker pull on every use.  This PR proposes a refinement to the Docker driver's `CreateNetwork` function to first check if the image is present locally before pulling it.  The expectation is to improve resiliency of Nomad job launches, and to remove a bit of excess load from GCR (or other configured registry).

Fixes #11014
Fixes #11380
Fixes #10318

### DIFF
--- a/drivers/docker/network.go
+++ b/drivers/docker/network.go
@@ -92,10 +92,28 @@ func (d *Driver) DestroyNetwork(allocID string, spec *drivers.NetworkIsolationSp
 		return fmt.Errorf("failed to connect to docker daemon: %s", err)
 	}
 
-	return client.RemoveContainer(docker.RemoveContainerOptions{
+	if err := client.RemoveContainer(docker.RemoveContainerOptions{
 		Force: true,
 		ID:    spec.Labels[dockerNetSpecLabelKey],
-	})
+	}); err != nil {
+		return err
+	}
+
+	if d.config.GC.Image {
+
+		// The Docker image ID is needed in order to correctly update the image
+		// reference count. Any error finding this, however, should not result
+		// in an error shutting down the allocrunner.
+		dockerImage, err := client.InspectImage(d.config.InfraImage)
+		if err != nil {
+			d.logger.Warn("InspectImage failed for infra_image container destroy",
+				"image", d.config.InfraImage, "error", err)
+			return nil
+		}
+		d.coordinator.RemoveImage(dockerImage.ID, allocID)
+	}
+
+	return nil
 }
 
 // createSandboxContainerConfig creates a docker container configuration which
@@ -124,8 +142,8 @@ func (d *Driver) pullInfraImage(allocID string) error {
 	if tag != "latest" {
 		dockerImage, err := client.InspectImage(d.config.InfraImage)
 		if err != nil {
-			d.logger.Debug("InspectImage failed for infra_image container pull", "image", d.config.InfraImage, "error", err)
-			return err
+			d.logger.Debug("InspectImage failed for infra_image container pull",
+				"image", d.config.InfraImage, "error", err)
 		} else if dockerImage != nil {
 			// Image exists, so no pull is attempted; just increment its reference count
 			d.coordinator.IncrementImageReference(dockerImage.ID, d.config.InfraImage, allocID)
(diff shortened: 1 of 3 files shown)


drivers/docker/network.go
```diff
@@ -28,17 +28,24 @@ func (d *Driver) CreateNetwork(allocID string, createSpec *drivers.NetworkCreate
...
+		if err != nil {
+			d.logger.Debug("auth failed for infra container image pull", "image", d.config.InfraImage, "error", err)
+		}
+		_, err = d.coordinator.PullImage(d.config.InfraImage, authOptions, allocID, noopLogEventFn, d.config.infraImagePullTimeoutDuration, d.config.pullActivityTimeoutDuration)
+		if err != nil {
+			return nil, false, err
+		}
+	} else {
+		// Image exists, so just increment its reference count
+		d.coordinator.IncrementImageReference(dockerImage.ID, d.config.InfraImage, allocID)
```
tbehling, line 48: Review question: Is it appropriate to invoke `IncrementImageReference` here?  I adapted this logic from `createImage`, but I don't know if there is maybe a reason to not use `IncrementImageReference` in this context. https://github.com/hashicorp/nomad/blob/9c582eaf6c0a68c1e32be45057161c8a0405b0bf/drivers/docker/driver.go#L537-L545

Also, for the `callerID` string (3rd argument) I used the `allocID`, because there's no task ID available like in `driver.go`.  I'm not sure if that's the most appropriate value though?

tbehling:
I just realized this is likely a dupe of https://github.com/hashicorp/nomad/pull/11380.

jrasell (changes requested):
Thanks for this PR @tbehling it's looking great!

It would be nice to change a bit of the logic around the tag checking before merging, please let me know if you have the time to take this change on, otherwise I am happy to push this.

This PR also needs a changelog entry: https://github.com/hashicorp/nomad/blob/main/contributing/CHANGELOG.md#developer-guide

TurtleIntercept:
Would someone be able to review this again please :) Would solve a big issue we're having!

jrasell (approved):
Hi @tbehling and thanks for working on the changes for this PR. I have pushed commit https://github.com/hashicorp/nomad/pull/13265/commits/e3a60b431fa514e968db9349388a38633237bf65 in order to resolve a couple of issue I found locally.

The first is that my previous comment regarding checking the error returned from `client.InspectImage` was a half truth. I checked the library and when the image is not found an error is returned. This therefore meant the infra-image could never be downloaded because the driver would always error and exit without trying to download it. I have reverted this behaviour as seen elsewhere in the Docker driver. I plan to raise a followup PR to wrap this client call in a helper so other committers do not need to be aware of this and we can handle the error type returned to give better errors and logs to operators.

The second was that although we were incrementing the infra-image reference count, we were not decrementing the count when the network was destroyed. I have therefore modified the `DestroyNetwork` function call to perform a best-effort image reference decrement.

@tgross seeing as I have made a couple of significant changes to this PR, i'd appreciate another set of eyes on this PR.

tgross (approved):

PR #13271, opened 2022-06-07
e2e: use longer wait in template update triggers to avoid flake.

jrasell:
The test has failed on/off recently and therefore this extends to wait period in hopes to fix this.


DerekStrickland (approved):
:shipit:

tgross (approved):

PR #13273, opened 2022-06-07
Disable path input when model is not new

### LINKED ISSUE #13178: UI: Forbid editing the path of a secure variable
Currently, will throw a "You can't update the ID of <var>" error.

philrenaud:
Resolves #13178 

- Still allows you to edit the path when editing at /new
- Disallows path editing on existing variables

### DIFF
--- a/ui/app/components/secure-variable-form.hbs
+++ b/ui/app/components/secure-variable-form.hbs
@@ -12,7 +12,8 @@
           @type="text"
           @value={{@model.path}}
           placeholder="/path/to/variable"
-          class="input"
+          class="input path-input"
+          disabled={{not (@model.isNew)}}
           {{autofocus}}
           />
       </label>
--- a/ui/app/components/secure-variable-form.hbs
+++ b/ui/app/components/secure-variable-form.hbs
@@ -13,7 +13,7 @@
           @value={{@model.path}}
           placeholder="/path/to/variable"
           class="input path-input"
-          disabled={{not (@model.isNew)}}
+          disabled={{not @model.isNew}}
           {{autofocus}}
           />
       </label>
--- a/ui/tests/integration/components/secure-variable-form-test.js
+++ b/ui/tests/integration/components/secure-variable-form-test.js
@@ -144,4 +144,30 @@ module('Integration | Component | secure-variable-form', function (hooks) {
       );
     });
   });
+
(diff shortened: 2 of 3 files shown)


ui/tests/integration/components/secure-variable-form-test.js
```diff
@@ -144,4 +144,30 @@ module('Integration | Component | secure-variable-form', function (hooks) {
...
+  test('Prevent editing path input on existing variables', async function (assert) {
+    assert.expect(3);
+
+    const variable = await this.server.create('variable', {
+      name: 'foo',
+      namespace: 'bar',
+      path: '/baz/bat',
+      keyValues: [{ key: '', value: '' }],
+    });
+    variable.isNew = false;
```
ChaiWithJai, line 157: `question`:  Are we unable to write (below) because we haven't added this property to the model yet?

```
    const variable = await this.server.create('variable', {
      name: 'foo',
      namespace: 'bar',
      path: '/baz/bat',
      keyValues: [{ key: '', value: '' }],
      isNew: false,
    });
```

ChaiWithJai (approved):
LGTM - just wondering if this diff should update the model, as well.

ui/app/components/secure-variable-form.hbs
```diff
@@ -12,7 +12,8 @@
           @type="text"
           @value={{@model.path}}
           placeholder="/path/to/variable"
-          class="input"
+          class="input path-input"
+          disabled={{not @model.isNew}}
```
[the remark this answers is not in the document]
philrenaud: isNew is an ember object concept that returns true if the object in question has never had a `.save()` call performed: https://api.emberjs.com/ember-data/3.9/classes/DS.Model/properties/isNew?anchor=isNew

ui/tests/integration/components/secure-variable-form-test.js
philrenaud: Turns out it's slightly more nuanced than this; Ember Mirage's implementation of isNew returns as a function (https://www.ember-cli-mirage.com/versions/v0.3.x/models/) on this.server.create(), so we set it afterwards to override the function property.

philrenaud: (it's necessary to do this at all because mirage models are not themselves ember models)

PR #13278, opened 2022-06-07
Notify a user if they try to make a variable with an existing path

### LINKED ISSUE #13127: Secure Variables: /new: Handle when the path to a variable already exists

philrenaud:
Resolves #13127

Adds a check on path input when creating a new secure variable. Checks for pre and post `/` character matches as well.

If there's an error, prompts the user to either change their path, or edit the existing variable.


ui/app/components/secure-variable-form.js
```diff
@@ -16,6 +21,20 @@ export default class SecureVariableFormComponent extends Component {
     return !this.args.model?.path;
   }
 
+  @action
+  validatePath(e) {
+    const value = trimPath([e.target.value]);
+    let existingVariable = this.store
+      .peekAll('variable')
+      .without(this.args.model)
+      .find((v) => v.path === value);
```
tgross, line 30: Is `this.store` isolated by namespace? If not, I think this might collide unintentionally if we have the same path in two different namespaces.

ChaiWithJai: Agreed, `this.store` is not isolated by namespace. We create a special property on jobs to help us avoid this by taking the name of the job and its namespace and storing that as the `id` (e.g. `"['app', 'default']"`).

ui/app/components/secure-variable-form.js
```diff
@@ -2,12 +2,17 @@ import Component from '@glimmer/component';
...
+import { trimPath } from '../helpers/trim-path';
+
 export default class SecureVariableFormComponent extends Component {
   @service router;
+  @service store;
 
   @tracked
   shouldHideValues = true;
 
+  @tracked duplicatePathError = false;
```
ChaiWithJai, line 14: `nit` `non-blocking`:  This has a union type of `boolean` and `interface {variable: Promise<Array<EmberDataObject>>}`.

Slightly confusing if I'm scanning this. But not a blocker by any means at all. I'm usually just documenting my learning when I do review.

ui/app/components/secure-variable-form.js
ChaiWithJai: `praise`:  great job avoiding a network request with `peekAll`.

ChaiWithJai (approved):
LGTM, maybe worth considering the `namespace` as an identifier here that Tim mentioned.

ui/app/components/secure-variable-form.js
```diff
@@ -16,6 +21,20 @@ export default class SecureVariableFormComponent extends Component {
     return !this.args.model?.path;
   }
 
+  @action
+  validatePath(e) {
+    const value = trimPath([e.target.value]);
+    let existingVariable = this.store
+      .peekAll('variable')
```
ChaiWithJai, line 28: `question` `thought` `non-blocking`:  Can you humor me?

What happens if there are 2 co-workers and they're both creating a variable within minutes of one another? There's no fetch request here to pick up on this.

Should we consider this scenario and use `findAll` instead?

ui/app/components/secure-variable-form.js
philrenaud: Yep, you're both absolutely right.

I'll be taking a "namespacifying" pass at all Secure Variables routes soon. Figured I'd get the base logic in first.

ui/app/components/secure-variable-form.js
philrenaud: Good call; let me set default to null here and name it better.

ui/app/components/secure-variable-form.js
philrenaud: As it stands, an error will happen from the API, and get passed up as a "The variable you are trying to create already exists" to the client.

I'll need to pass that on to the user as part of form validation. But one's work wouldn't override the other's.

Issue #12465, opened 2022-04-05
Vault token renewals race with task shutdown


tgross:
@cullenmcdermott reported spurious Vault client errors when a batch job that required Vault had exited.

Searching through their logs, we find the following series of events. Note: these log messages are reconstructed from structured log outputs, so they may not match standard Nomad stdout logging you'd find in the journal:

> T40:58.352672Z  [client.alloc_runner.task_runner] not restarting task alloc_id=cb3d5b0f-3a08-23f0-a8bc-46590e2e5af4
> T40:58.631671Z [client.driver_mgr.docker] stopped container
> T40:58.650978Z [client.alloc_runner] killing task alloc_id=cb3d5b0f-3a08-23f0-a8bc-46590e2e5af4
> T40:58.651019Z [client.gc] marking allocation for GC alloc_id=cb3d5b0f-3a08-23f0-a8bc-46590e2e5af4
> T40:58.785934Z [client.vault] error during renewal of lease or token failed due to a non-fatal error; retrying

That last log line came with the following multi-line log:

```
failed to renew the vault token: Error making API request.

URL: PUT https://active.vault.service.consul:8200/v1/auth/token/renew-self
Code: 400. Errors:

* invalid lease ID
```

And then finally:

> T41:13.787658Z [client.vault] error renewing token

```
failed to renew the vault token: Error making API request.

URL: PUT https://active.vault.service.consul:8200/v1/auth/token/renew-self
Code: 403. Errors:

* permission denied
```

The final set of errors are coming from [`vaultclient.go#L360-L363`](https://github.com/hashicorp/nomad/blob/v1.2.6/client/vaultclient/vaultclient.go#L360-L363) which gets called from [`vault_hook.go#L229`](https://github.com/hashicorp/nomad/blob/v1.2.6/client/allocrunner/taskrunner/vault_hook.go#L229) in the task runner. At a quick glance it looks like the renew step does not include the task runner's context, so we don't have a way to interrupt that operation. The fix will probably need to be threading the task runner context into the vault client renewal somehow.

cc @lgfa29 and @mikenomitch as a heads up for this internally-reported issue that we should probably get onto the near-term backlog.

tgross:
Cross-linking to https://github.com/hashicorp/nomad/issues/12261 because although it's not the same error message or symptoms, they smell potentially related.

mikenomitch:
[...]

PR #13279, opened 2022-06-08
vault_hook: build context from task_runner shutdown context

DerekStrickland:
Closes: #12465

This PR addresses a defect where the Vault token renewal process may race with a task shutting down or being killed. The code has been modified in the following ways.

- `newTaskHook` now requires `context.Context` and a `context.CancelFunc` parameters and uses these passed values instead of `context.Background()`.
- `taskRunner.initHooks` calls `joincontext.Join` on `shutdownCtx` and `killCtx` and passes the result to `newTaskHook`

Since the passed context is a joined version of `shutdownCtx` and `killCtx`, either being cancelled will cause the `vaultHook.run` loop to stop the renewal and exit.

### DIFF
--- a/client/allocrunner/taskrunner/task_runner_hooks.go
+++ b/client/allocrunner/taskrunner/task_runner_hooks.go
@@ -84,7 +84,8 @@ func (tr *TaskRunner) initHooks() {
 
 	// If Vault is enabled, add the hook
 	if task.Vault != nil {
-		tr.runnerHooks = append(tr.runnerHooks, newVaultHook(&vaultHookConfig{
+		joinedCtx, joinedCancel := joincontext.Join(tr.killCtx, tr.shutdownCtx)
+		tr.runnerHooks = append(tr.runnerHooks, newVaultHook(joinedCtx, joinedCancel, &vaultHookConfig{
 			vaultStanza: task.Vault,
 			client:      tr.vaultClient,
 			events:      tr,
@@ -93,7 +94,6 @@ func (tr *TaskRunner) initHooks() {
 			logger:      hookLogger,
 			alloc:       tr.Alloc(),
 			task:        tr.taskName,
-			parentCtx:   tr.shutdownCtx,
 		}))
 	}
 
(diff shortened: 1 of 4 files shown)


lgfa29:
After looking deeper into this part of the code I think there may be something else going on, and I'm not sure this would fix the problem.

The Vault hook is also a `Shutdown` and a `Stop` hook. `Shutdown` hooks [are called right after `shutdownCtxCancel` is cancelled](https://github.com/hashicorp/nomad/blob/v1.3.1/client/allocrunner/taskrunner/task_runner.go#L1377-L1378). `Stop` hooks are called [when the task runner `Run()` loop exists](https://github.com/hashicorp/nomad/blob/v1.3.1/client/allocrunner/taskrunner/task_runner.go#L669-L672), _except_ when it's a shutdown and the loop will return immediately, but the Vault renewal would be stopped by that as well since the Vault hook is a `Shutdown` hook. When `killCtxCancel` is called, the `Run()` loop exits calling the `Stop` hooks.

Another indication of this is that, as you mentioned earlier, the test passes even without the code change. 

So, in theory, the current code looks right, which leaves a few other options:

- Everything I said is wrong, and that's not how things are supposed to work 😅 
- The task runner is not calling its hooks the way it's supposed to do
- The Vault client `stopRenew` is not working properly (it actually doesn't do much)
- Some weird timing issue (the renewal loop runs between the task being killed and the hooks actually been called)

Possible lines of investigations:

- Add more hooks to the Vault hook. Maybe setting it as a [`PreKill`](https://github.com/hashicorp/nomad/blob/v1.3.1/client/allocrunner/interfaces/task_lifecycle.go#L131-L138) for example could cancel the loop before the token attempts a renewal.
- Investigate when the token is actually revoked in Vault. I haven't quite found where that happens yet. I suspect it may happen in the server.
- Look for code paths where the `Shutdown` and/or `Stop` hooks may not be called.
- Investigate if the problem is caused by [recovered tokens](https://github.com/hashicorp/nomad/blob/v1.3.1/client/allocrunner/taskrunner/vault_hook.go#L141-L142) that is no longer valid. Maybe unlikely as we would've seen this [`failed to start renewal of Vault token`](https://github.com/hashicorp/nomad/blob/v1.3.1/client/allocrunner/taskrunner/vault_hook.go#L233).
- Maybe this is expected, due to weird timing, but eventually the token is revoked and everything is good. In this case we may reconsider emitting an error message while there are still retried left.

I will take another look on Monday, hopefully this will all still make sense then 😅

lgfa29:
Closing in favour of https://github.com/hashicorp/nomad/pull/18985

Issue #10911, opened 2021-07-20
CPU metrics not tracked on AWS Graviton instances


tvdu29:
### Nomad version
`Nomad v1.1.2 (60638a086ef9630e2a9ba1e237e8426192a44244)`

### Operating system and Environment details
Ubuntu 20.04 on aws c6g.xlarge

### Issue
The CPU metrics reported in Nomad UI of our clients running on AWS Graviton are all locked at 0Mhz.
We observe this behaviour in both the client and task view.
![2021-07-20 11 35 45 nomad prod adotmob com 1d61aca46f9b](https://user-images.githubusercontent.com/11725577/126300214-6930efb7-344e-4f48-8796-288843d7dc05.png)
![2021-07-20 11 36 00 nomad prod adotmob com cab8bf1168fb](https://user-images.githubusercontent.com/11725577/126300276-4233cacb-df22-4eb1-8f28-00c036790057.png)

We believe that it is wrongly reported because our telegraf agent correctly reports CPU usage.

### Reproduction steps
Launch a nomad client on AWS Graviton instance & deploy jobs to it.

#### Expected Result
CPU usage reported in client & task views

#### Actual Result
CPU usage locked at 0Mhz in client & task views

bfqrst:
...good catch @tvdu29 ! All of my Nomads and Consuls running in Gravitons but never paid attention to this! Confirming...

zhiguangwang:
Same here, the latest c7g instances with Graviton 3 are also affected by this issue, which is a bummer.

tgross:
Hi folks! Unfortunately the AWS fingerprinter has to rely on a hardcoded table of CPU data because the library we use for fingerprinting can't get the right data out of AWS. I've updated the table in https://github.com/hashicorp/nomad/pull/13280 and that'll ship in the next release of Nomad, with backports to 1.2.x and 1.1.x.

PR #13280, opened 2022-06-08
update AWS cpu info for fingerprinter

tgross:
Fixes https://github.com/hashicorp/nomad/issues/10911


DerekStrickland (approved):
Double checked all these and they look goos.

Issue #13264, opened 2022-06-06
CSI: discard claims on GC'd nodes


tgross:
From https://github.com/hashicorp/nomad/issues/12346#issuecomment-1135910108: 

> There's no way for the server to send a node unpublish command to the node plugin that's running on a down node. What's becoming clear here is that the CSI specification as written _cannot_ guarantee correct behavior in the case of downed nodes, so what K8s apparently does is just gives up and lets you potentially corrupt your data in the multi-writer volume use case. User expectations seem to agree, so we may need to rework the design here to make it possible to discard claims for lost nodes and just assume that they're never coming back. This is why we built [stop_after_client_disconnect](https://www.nomadproject.io/docs/job-specification/group%23stop_after_client_disconnect) but maybe we'll need to enforce that setting for CSI-claiming tasks unless specifically disabled by the user.

* We'll alter the claim Unpublish workflow to skip the Node Unpublish RPCs for any node that is marked as `lost`, `disconnected`, or nil (GC'd). Just log the fault and send the controller RPCs, if any. 
* If users want correct behavior on the clients, they'll include `stop_after_client_disconnect`. We should consider making this the default behavior for allocations that include CSI volume claims, but this might cause us backwards compatibility grief.
* In later work as a performance optimization, we should also update the `volumewatcher` to watch for updates to the Node table. When a node is GC'd or marked as `lost`, emit a volume reap on any volumes that have claims on that node.

PR #13301, opened 2022-06-08
CSI: skip node unpublish on GC'd or down nodes

tgross:
Fixes https://github.com/hashicorp/nomad/issues/13264

If the node has been GC'd or is down, we can't send it a node
unpublish. The CSI spec requires that we don't send the controller
unpublish before the node unpublish, but in the case where a node is
gone we can't know the final fate of the node unpublish step.

The `csi_hook` on the client will unpublish if the allocation has
stopped and if the host is terminated there's no mount for the volume
anyways. So we'll now assume that the node has unpublished at its
end. If it hasn't, any controller unpublish will potentially hang or
error and need to be retried.

(Note that while this behavior isn't ideal, it appears to match user
expectations and the behavior reported by k8s users.)

### DIFF
--- a/.changelog/13301.txt
+++ b/.changelog/13301.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+csi: Fixed a bug where volume claims on lost or garbage collected nodes could not be freed
+```
--- a/nomad/csi_endpoint.go
+++ b/nomad/csi_endpoint.go
@@ -686,6 +686,25 @@ RELEASE_CLAIM:
 
 func (v *CSIVolume) nodeUnpublishVolume(vol *structs.CSIVolume, claim *structs.CSIVolumeClaim) error {
 	v.logger.Trace("node unpublish", "vol", vol.ID)
+
+	store := v.srv.fsm.State()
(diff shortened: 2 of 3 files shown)


.changelog/13301.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+csi: Fixed a bug where volume claims on lost or garbage collected nodes could not be freed
```
tgross, line 2: Note for reviewers: I'm torn on whether to call this a bug or improvement but calling it a bug makes it something we can backport so I'm leaning that way.

lgfa29 (approved):
I was thinking about other possible node status (like initializing or draining) but those would be fine since the node is still around to handle the request right?

tgross:
> I was thinking about other possible node status (like initializing or draining) but those would be fine since the node is still around to handle the request right?

Exactly! The `disconnected` state is a little weird as well, but in that case we're keeping the claims if the node can be marked disconnected, because the allocations will still be up and running, just temporarily unavailable.

PR #13308, opened 2022-06-09
helpers: provide a few generic helper functions

shoenig:
This PR deprecates some functions in favor of generic alternatives.

The new functions are compatible only with Nomad v1.4+.

The old functions (and their use) should not be removed until Nomad v1.6+.

emphasis on DO NOT BACKPORT


tgross (approved):
LGTM.

Not backporting means that backports that use these functions will break, but I think that's exactly what we want here.

PR #13312, opened 2022-06-09
remove unbackportable test

tgross:
https://github.com/hashicorp/nomad/pull/13301 was backported to 1.2.x and 1.1.x but included a test that would only run under 1.3.0+. Remove this test.


shoenig (approved):
Should the terraform lock file be in here?

tgross:
> Should the terraform lock file be in here?

Whoops, no! Fixed.

PR #13314, opened 2022-06-09
ci: remove step that generates GO_LDFLAGS

lgfa29:
These flags were not being used because GNUmakefile overwrites them with
another value. We also don't want to set `-s -w` since they remove
information that is important for production debug.

In other projects this variable is used to override the default `-dev`
prerelease that is set even if `VersionPrerelease` is empty, but in
Nomad this check is never actually done because this conditional in
`version/version.go` is always false:

```go
func GetVersion() *VersionInfo {
  // ...
  rel := VersionPrerelease
  // ...
  if GitDescribe == "" && rel == "" && VersionPrerelease != "" {
    rel = "dev"
  }
  // ...
}
```

This seems like some leftover from a previous release process, but I
decided the leave the code as is.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,25 +72,9 @@ jobs:
         with:
           name: metadata.json
           path: ${{ steps.generate-metadata-file.outputs.filepath }}
-  generate-ld-flags:
-    needs: get-product-version
-    runs-on: ubuntu-latest
-    outputs:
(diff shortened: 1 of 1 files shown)


tgross (approved):
LGTM. Thanks for picking up the `-s -w` stuff -- that'll make a huge difference in a SEV1 support call some terrible late night 😀

shoenig (approved):
LGTM!

modrake (approved):

PR #13330, opened 2022-06-10
build: rearrange go path order on windows

shoenig:
If PATH comes first, an older version of Go is used that cannot install
dependencies that use features of newer versions of Go, which we just
installed.


tgross (approved):
LGTM!

PR #13334, opened 2022-06-10
template: improve default language for max_stale and wait

DerekStrickland:
This PR addresses two issues:

- The language for the `max_stale` setting contained out of date "by default" language relative to the new default behavior.
- The `wait` language was clarified to remove a confusing "by default" statement when explaining how Consul Template works.

### DIFF
--- a/website/content/docs/configuration/client.mdx
+++ b/website/content/docs/configuration/client.mdx
@@ -247,15 +247,17 @@ chroot as doing so would cause infinite recursion.
   access files only within the [task working directory].
 
 - `max_stale` `(string: "87600h")` - This is the maximum interval to allow "stale"
-  data. By default, only the Consul leader will respond to queries. Requests to
-  a follower will forward to the leader. In large clusters with many requests,
-  this is not as scalable. This option allows any follower to respond to a query,
-  so long as the last-replicated data is within this bound. Higher values result
-  in less cluster load, but are more likely to have outdated data.
+  data. If `max_stale` is not set, only the Consul leader will respond to queries, and
+  requests that reach a follower will forward to the leader. In large clusters with
+  many requests, this is not as scalable. This option allows any follower to respond
+  to a query, so long as the last-replicated data is within this bound. Higher values
+  result in less cluster load, but are more likely to have outdated data. This default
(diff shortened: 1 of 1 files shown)


website/content/docs/configuration/client.mdx
```diff
@@ -247,15 +247,17 @@ chroot as doing so would cause infinite recursion.
   access files only within the [task working directory].
 
 - `max_stale` `(string: "87600h")` - This is the maximum interval to allow "stale"
-  data. By default, only the Consul leader will respond to queries. Requests to
-  a follower will forward to the leader. In large clusters with many requests,
-  this is not as scalable. This option allows any follower to respond to a query,
-  so long as the last-replicated data is within this bound. Higher values result
-  in less cluster load, but are more likely to have outdated data.
+  data. If `max_stale` is not set, only the Consul leader will respond to queries, and
```
schmichael, line 250:
```suggestion
  data. If `max_stale` is set to `0`, only the Consul leader will respond to queries, and
```

Since the default is 87.6k hours, users have to set it to 0 explicitly to get this behavior. It may be "not set" from Nomad's perspective of 0 values, but it **is set** from the user's perspective of having to write `max_stale = "0"` in their configuration file.

website/content/docs/configuration/client.mdx
```diff
@@ -247,15 +247,17 @@ chroot as doing so would cause infinite recursion.
...
+  requests that reach a follower will forward to the leader. In large clusters with
+  many requests, this is not as scalable. This option allows any follower to respond
+  to a query, so long as the last-replicated data is within this bound. Higher values
+  result in less cluster load, but are more likely to have outdated data. This default
+  of 10 years (`87600h`) matches the default Consul configuration.
```
schmichael, line 255: Ha I hadn't done the math before. I guess that's a valid way to implement "indefinitely" :sweat_smile:

schmichael (approved):
Thanks Derek!

website/content/docs/configuration/client.mdx
DerekStrickland: Nice distinction. Thank you.

PR #13340, opened 2022-06-10
CSI: make plugin health_timeout configurable in csi_plugin stanza

### LINKED ISSUE #13179: CSI: make readiness probe timeout configurable to support faulty plugins
### Nomad version
Output from `nomad version`
Ubuntu Linux 20.04
### Issue
CSI Plugin on Nomad 1.3 have a very short timeout. Affects Portworx which can take 15 minutes to be installed.
### Reproduction steps
Try to install Portworx on Nomad as a job following this guide:
https://docs.portworx.com/install-with-other/nomad/installation/install-as-a-nomad-job/

#### Expected Result
Portworx running on Nomad
#### Actual Result
CSI plugin timeout

ggriffiths:
Fixes #13179 

Test results: Configured health_timeout as `5m` and introduced an artificial error to make sure Portworx was killed after 5m:
```
Recent Events:
Time                  Type                     Description
2022-06-10T20:05:40Z  Killed                   Task successfully killed
2022-06-10T20:05:40Z  Terminated               Exit Code: 2, Exit Message: "Docker container exited with non-zero exit code: 2"
2022-06-10T20:05:40Z  Killing                  CSI plugin did not become healthy before configured 5m0s health timeout
2022-06-10T20:05:40Z  Plugin became unhealthy  Error: CSI plugin failed probe: timeout while connecting to gRPC socket: failed to stat socket: stat /etc/nomad.d/client/csi/plugins/62bdddd8-3955-92d2-cc23-02b6f6204e62/csi.sock: no such file or directory
2022-06-10T19:58:15Z  Started                  Task started by client
2022-06-10T19:58:13Z  Task Setup               Building Task Directory
2022-06-10T19:58:13Z  Received                 Task received by client
```

Signed-off-by: Grant Griffiths <ggriffiths@purestorage.com>

### DIFF
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -7255,6 +7255,10 @@ func (t *Task) Validate(ephemeralDisk *EphemeralDisk, jobType string, tgServices
 			mErr.Errors = append(mErr.Errors, fmt.Errorf("CSIPluginConfig PluginType must be one of 'node', 'controller', or 'monolith', got: \"%s\"", t.CSIPluginConfig.Type))
 		}
 
+		if t.CSIPluginConfig.HealthTimeout == 0 {
+			t.CSIPluginConfig.HealthTimeout = 30 * time.Second
+		}
+
 		// TODO: Investigate validation of the PluginMountDir. Not much we can do apart from check IsAbs until after we understand its execution environment though :(
 	}
 
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -7255,10 +7255,6 @@ func (t *Task) Validate(ephemeralDisk *EphemeralDisk, jobType string, tgServices
 			mErr.Errors = append(mErr.Errors, fmt.Errorf("CSIPluginConfig PluginType must be one of 'node', 'controller', or 'monolith', got: \"%s\"", t.CSIPluginConfig.Type))
 		}
 
-		if t.CSIPluginConfig.HealthTimeout == 0 {
-			t.CSIPluginConfig.HealthTimeout = 30 * time.Second
-		}
-
 		// TODO: Investigate validation of the PluginMountDir. Not much we can do apart from check IsAbs until after we understand its execution environment though :(
 	}
 
--- a/.changelog/13340.txt
+++ b/.changelog/13340.txt
@@ -1,0 +1,3 @@
+```release-note:improvements
+csi: Made the CSI Plugin supervisor health check configurable with a new CSI Stanza health_timeout field
+```
--- a/api/tasks.go
+++ b/api/tasks.go
@@ -1039,6 +1039,10 @@ type TaskCSIPluginConfig struct {
 	//
 	// Default is /csi.
 	MountDir string `mapstructure:"mount_dir" hcl:"mount_dir,optional"`
+
+	// HealthTimeout is the time after which the CSI plugin tasks will be killed
+	// if the CSI Plugin is not healthy.
+	HealthTimeout time.Duration `mapstructure:"health_timeout" hcl:"health_timeout,optional"`
 }
 
 func (t *TaskCSIPluginConfig) Canonicalize() {
--- a/api/tasks.go
+++ b/api/tasks.go
@@ -1049,4 +1049,8 @@ func (t *TaskCSIPluginConfig) Canonicalize() {
 	if t.MountDir == "" {
 		t.MountDir = "/csi"
 	}
+
+	if t.HealthTimeout == 0 {
+		t.HealthTimeout = 30 * time.Second
+	}
 }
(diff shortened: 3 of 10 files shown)


nomad/structs/structs.go
```diff
@@ -7255,6 +7255,10 @@ func (t *Task) Validate(ephemeralDisk *EphemeralDisk, jobType string, tgServices
 			mErr.Errors = append(mErr.Errors, fmt.Errorf("CSIPluginConfig PluginType must be one of 'node', 'controller', or 'monolith', got: \"%s\"", t.CSIPluginConfig.Type))
 		}
 
+		if t.CSIPluginConfig.HealthTimeout == 0 {
+			t.CSIPluginConfig.HealthTimeout = 30 * time.Second
```
ggriffiths, line 7259: Added the default value as 30 seconds for backward compatibility. Let me know if this is the right spot for this @tgross

tgross: There are two issues with this spot:

* This will help for ensuring backwards compatibility during upgrades, where the server might be newer than the client (although we don't recommend upgrading in that order). But we also need to account for cases where the job is already in the state store and lives across an upgrade. Those plugins won't hit `Validate()` again and so when the plugin probe reattaches we'll get 0.

* Also, although this technically works because we pass a pointer into the `Validate()` methods, we don't want to mutate the job during validation. If you look at `nomad/job_endpoint.go` you'll see a reference to both "admission mutators" and "admission validators". (This is mostly for architectural clarity in job registration.)

So let's move this down into the `csiPluginSupervisorHook` constructor. That'll solve both the problems above.

tgross (changes requested):
This looks great @ggriffiths! 

There's a [jobspec checklist](https://github.com/hashicorp/nomad/blob/main/contributing/checklist-jobspec.md) which I probably should have pointed out in the original issue, but it looks like you've already hit almost everything needed there except for the implementation for [`structs/diff.go`](https://github.com/hashicorp/nomad/blob/main/nomad/structs/diff.go).

We'll also want this new field added to the [CSI plugin parameters](https://github.com/hashicorp/nomad/blob/main/website/content/docs/job-specification/csi_plugin.mdx#csi_plugin-parameters) documentation.

Thanks for the PR!

nomad/structs/structs.go
ggriffiths: Sounds good to me :+1:

ggriffiths:
> There's a [jobspec checklist](https://github.com/hashicorp/nomad/blob/main/contributing/checklist-jobspec.md) which I probably should have pointed out in the original issue, but it looks like you've already hit almost everything needed there except for the implementation for [`structs/diff.go`](https://github.com/hashicorp/nomad/blob/main/nomad/structs/diff.go).

@tgross, thanks for the pointer! For `structs/diff.go`, should I add a new function for the CSIPluginConfig? I'm not seeing any existing logic in `diff.go` for this struct.

tgross:
> @tgross, thanks for the pointer! For `structs/diff.go`, should I add a new function for the `CSIPluginConfig`? I'm not seeing any existing logic in `diff.go` for this struct.

Oh. You've just discovered a bug in `job plan` output. 😊 We detect that the `csi_plugin` block has changed but aren't producing a diff of what fields have changed. This failing test shows the problem:

<details><summary>failing test</summary>

```diff
diff --git a/nomad/structs/diff_test.go b/nomad/structs/diff_test.go
index 1a5751a8c..153c5bb94 100644
--- a/nomad/structs/diff_test.go
+++ b/nomad/structs/diff_test.go
@@ -7377,6 +7377,40 @@ func TestTaskDiff(t *testing.T) {
                                },
                        },
                },
+               {
+                       Name: "CSIPlugin edited",
+                       Old: &Task{
+                               CSIPluginConfig: &TaskCSIPluginConfig{
+                                       ID:       "id0",
+                                       Type:     "monolith",
+                                       MountDir: "/csi",
+                               },
+                       },
+                       New: &Task{
+                               CSIPluginConfig: &TaskCSIPluginConfig{
+                                       ID:       "id0",
+                                       Type:     "monolith",
+                                       MountDir: "/whatever",
+                               },
+                       },
+                       Expected: &TaskDiff{
+                               Type: DiffTypeEdited,
+                               Objects: []*ObjectDiff{
+                                       {
+                                               Type: DiffTypeEdited,
+                                               Name: "CSIPluginConfig",
+                                               Fields: []*FieldDiff{
+                                                       {
+                                                               Type: DiffTypeEdited,
+                                                               Name: "MountDir",
+                                                               Old:  "/csi",
+                                                               New:  "/whatever",
+                                                       },
+                                               },
+                                       },
+                               },
+                       },
+               },
        }

        for _, c := range cases {
```

</details>

Because that's a bug and this PR is for a feature improvement, we'll need to handle those as separate PRs so we can backport the plan fix. So don't worry about diff for now and I'll come back for that later once this has been merged.

PR #13341, opened 2022-06-10
Toast-style global notification state

philrenaud:
Global notification state that persists across routing change (you can fire a notification and redirect the user and the notification will stay)

- Progress bar toggleable on/off
- can either have a duration or be "sticky"
- Currently allow for only error and success states but extensible

image

### DIFF
--- a/ui/app/components/secure-variable-form.js
+++ b/ui/app/components/secure-variable-form.js
@@ -10,6 +10,7 @@ import EmberObject from '@ember/object';
 
 export default class SecureVariableFormComponent extends Component {
   @service router;
+  @service flashMessages;
 
   @tracked
   shouldHideValues = true;
@@ -61,9 +62,19 @@ export default class SecureVariableFormComponent extends Component {
     const value = e.target.value;
     if (value.includes('.')) {
       entry.warnings.set('dottedKeyError', 'Key should not contain a period.');
+      this.flashMessages.danger('Key should not contain a period.');
     } else {
       delete entry.warnings.dottedKeyError;
       entry.warnings.notifyPropertyChange('dottedKeyError');
+      this.flashMessages.add({
+        title: 'woohoo',
+        message: value,
+        type: 'success',
+        destroyOnClick: false,
+        timeout: 4000,
+        // sticky: Math.random() > 0.5 ? true : false,
+        showProgress: true,
+      });
     }
   }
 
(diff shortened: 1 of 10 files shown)


ui/tests/helpers/flash-message.js
```diff
@@ -0,0 +1,3 @@
+import FlashObject from 'ember-cli-flash/flash/object';
+
+FlashObject.reopen({ init() {} });
```
ChaiWithJai, line 3: `thought` `non-blocking`: I've noticed a few issues arise from calling `reopen` on an `EmberObject`. Thankfully, this is officially an adopted Ember add-on.

https://www.pzuraq.com/blog/emberjs-native-class-update-2019-edition#avoiding-reopen-and-reopenclass

ui/app/components/secure-variable-form.js
```diff
@@ -87,9 +88,27 @@ export default class SecureVariableFormComponent extends Component {
...
+    try {
+      this.args.model.set('keyValues', this.keyValues);
+      this.args.model.setAndTrimPath();
+      await this.args.model.save();
+      this.flashMessages.add({
+        title: 'Secure Variable saved',
+        message: `${this.args.model.path} successfully saved`,
+        type: 'success',
+        destroyOnClick: false,
+        timeout: 5000,
```
ChaiWithJai, line 100: `question` `performance (testing)`: Will this block the thread of execution in our test suites? I could imagine our test suite being slowed down waiting for this flash message to disappear. Should we consider `timeout` to read from our `config` file that can determine if we're in `dev`, `prod` or `testing` and appropriately set the timer?

ChaiWithJai (approved):
Great job! Excited to see where else we add these toast notifications to the application.

ui/package.json
```diff
@@ -71,6 +71,7 @@
     "ember-cli-clipboard": "^0.13.0",
     "ember-cli-dependency-checker": "^3.2.0",
     "ember-cli-deprecation-workflow": "^2.1.0",
+    "ember-cli-flash": "^3.0.0",
```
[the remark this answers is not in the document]
philrenaud: Hm interesting -- I took to ember-clli-flash after asking about it in our frontend slack channel and being told flash messages were the norm here. I'm open to either, especially since notifications seem more recently updated.

ui/app/components/secure-variable-form.js
philrenaud: No, these are considered non-blocking; they happen asynchronously and allow transitions and anything else to happen aside from them.

PR #13342, opened 2022-06-10
docs: explain behavior of system gc command

schmichael:
Thanks to @tbehling for pointing out how incomplete these docs were. I had forgotten `system gc` ignores the server's `threshold` configuration!

Rendered: 
image


jrasell (approved):
LGTM!

Issue #11900, opened 2022-01-21
vault_token is written to /secrets world-readable


grembo:
*Update 2022-05-31: As things evolved and to clarify the mission: The best solution and my goal is to store the token in a separate directory that is not accessible from inside the container. #11905 accomplishes this by writing the vault token to a private directory with strict file permissions (also protecting the token from other unprivileged processes on the host running nomad). It then adds a parameter called `file` which allows writing a copy of vault token to `secrets` like before*

Nomad writes a job's vault token to `/secrets/vault_token` using world-readable permissions, as default perms (0666) are used (and then umask is applied):

https://github.com/hashicorp/nomad/blob/eee5d90e8bc105201d5a2b98d480af2cd76df833/client/allocrunner/taskrunner/vault_hook.go#L344-L351

In practice (using a default umask `0022`) this means _/secrets/vault_token_ is world readable and writeable only by root (0644).

As changing this might break existing setups, an alternative would be to add a `perms` parameter to the `vault` stanza like a job's [template stanza already supports](https://www.nomadproject.io/docs/job-specification/template#perms).

If there's a chance for adoption, I'm happy to implement/open a pull request for the feature.

_Update: Misread how perms work in writefile, so I updated the issue title accordingly. This makes writing a patch for this more complicated (I already prepared one), as - in theory - existing users could have altered umask. It also means that the issue could be worked around by altering umask for the nomad process, even though this might have unexpected side-effects._

jrasell:
Hi @grembo and thanks for raising this along with the PR. I'll raise this internally for discussion and hope to come back with feedback on the PR in a timely manner.

schmichael:
The implementation of #11905 looks good, but I'm not sure it adds significant protections.

To be clear the Vault token is protected from arbitrary readers by two other layers:

1. It is protected from reading by processes outside Nomad by the permissions on [`alloc_dir`](https://www.nomadproject.io/docs/configuration/client#alloc_dir) being only readable by the Nomad agent's user (preferably root).
2. It is protected from reading by other allocations and tasks by the chroot/filesystem-isolation of the task driver used. (This doesn't apply to raw_exec, but raw_exec runs as root so there's no hiding anything from it!).

[...]

PR #13343, opened 2022-06-12
Add `file` parameter to job's `vault` stanza

grembo:
This complements the `env` parameter, so that the operator can author
tasks that don't share their Vault token with the payload. As a result,
more powerful tokens can be used in a job definition, allowing it to
use template stanzas to issue all kinds of secrets (database secrets,
Vault tokens with very specific policies, etc.), without sharing that
issuing power with the task itself as long as a driver with `image`
isolation is used.

This is accomplished by creating a directory called `private` within
the task's working directory, which shares many properties of
the `secrets` directory (tmpfs where possible, not accessible by
`nomad alloc fs` or Nomad's web UI), but isn't mounted into/bound to the
container.

If the `file` parameter is set to `true` (its default), the Vault token
is also written to the NOMAD_SECRETS_DIR, so the default behavior is
backwards compatible. Even if the operator never changes the default,
they will still benefit from the improved behavior of Nomad never reading
the token back in from that - potentially altered - location.

See #11900


grembo:
authrequired

Can someone please authorize this (if it is a blocker to proceed with the review)? Thanks!

tgross:
Hi @grembo we weren't waiting on CI for a review here, I think there was just some confusion as to which of the several PRs we have open for the original issue was the one we want. I'm going to assign the PR review to @lgfa29 as it looks like it he was the one who spent a bunch of time going back and forth with you on the design.

website/content/docs/job-specification/vault.mdx
```diff
@@ -109,6 +113,35 @@ vault {
...
+  env = false
+  file = false
+}
+
+template {
+  data = <<-EOH
+  {{with secret "auth/token/create/nomad-job" "policies=examplepolicy"}}{{.Auth.ClientToken}}{{ end }}
+  EOH
+  destination = "${NOMAD_SECRETS_DIR}/examplepolicy.token"
+  change_mode = "noop"
```
tgross, line 139: It's not clear from this example why we want `noop` here?

api/tasks.go
```diff
@@ -857,6 +857,7 @@ type Vault struct {
 	Env          *bool    `hcl:"env,optional"`
 	ChangeMode   *string  `mapstructure:"change_mode" hcl:"change_mode,optional"`
 	ChangeSignal *string  `mapstructure:"change_signal" hcl:"change_signal,optional"`
+	File         *bool    `mapstructure:"file" hcl:"file,optional"`
```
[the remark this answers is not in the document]
grembo: For my application this would work (I don't want the token either on disk nor in the environment). I could imagine that there are users though that need the file on disk, but want to avoid it in the environment (e.g., to exclude it from logging), those people would be in trouble if we change behavior. I have no metrics, if that is a thing, so this is theoretic.

grembo:
> Hi @grembo we weren't waiting on CI for a review here, I think there was just some confusion as to which of the several PRs we have open for the original issue was the one we want. I'm going to assign the PR review to @lgfa29 as it looks like it he was the one who spent a bunch of time going back and forth with you on the design.

Apologies, I though my comment on #11900 (and turning the previous review into a Draft) would make it clear. Let me abandon the previous review to reduce the noise.

website/content/docs/job-specification/vault.mdx
tgross: Keeping it in the example but providing a little comment as to why you might want it seems fine.

I kind of like the notion of having it default to `noop` if both `env = false` and `file = false`, because all the client's internal "update" hooks would still get fired without touching the task ("noop" is unfortunately named, because it's not strictly no-op, just no-op for the task driver). Let's check that this wouldn't unexpectedly change the behavior of existing allocations after the client gets upgraded.

api/tasks.go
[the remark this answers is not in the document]
tgross: Good news, another Nomad user came to us asking for this same sort of thing and they definitely want to have `env` without `file` for their use case. So that's at least one vote for keeping them separate and none against 😁.

website/content/docs/job-specification/vault.mdx
grembo: @tgross 
> Let's check that this wouldn't unexpectedly change the behavior of existing allocations after the client gets upgraded.

I would assume that it's safe, since the `file` parameter
1. couldn't be specified beforehand (nomad doesn't accept unknown params in the vault stanza)
2. is new and defaults to true

so upgrading the client can never result in a situation where `env == true && file == true`.

grembo: @tgross I extended the example to show how to use mix `change_mode`s and added a few words explaining what this means.

grembo:
@tgross Hi, any news on this? Thanks!

lgfa29:
Hi @grembo 👋 

Apologies for delay on reviewing this, it unfortunately fell out of my radar...I will give it another pass in the coming days!

marknl:
Hi all! We're looking forward to see this PR merged into Nomad.
@lgfa29 Any update?

lgfa29:
Hi everyone 👋 

Apologies for the delay on moving this forward, this is an area of the code we've been iterating over and it's also a sneaky complex part as well, so it took me a while to get the time to dedicate to a thorough review.

I have been able to give it another pass and found some critical issues on the upgrade path.The biggest one is that [this line](https://github.com/hashicorp/nomad/pull/13343/files#diff-fdbf72fb40a559f0592c927e632f7f8a7178915920dd439de74e8798bc4acb88R351) will cause an existing alloc to fail when the Nomad agent is upgraded because the `/private` directory doesn't exist in that allocation.

The other big problem is the lesson we learned the hard way in https://github.com/hashicorp/nomad/pull/17087, which is that we can't default to a non-zero value for a new attribute. What this means is that I had to change the attribute name from `file` to `disable_file` so that existing jobs are loaded in a backwards compatible way after an upgrade. Without this, if an old alloc fails it would be recreated with `file: false` even though we need it to be `true`.

Since this PR has been stale for a while because of us, I don't think it would fair to ask @grembo to do all of these changes, so I implemented the fixes in https://github.com/hashicorp/nomad/compare/vault-token-file-rebased (ignore that `tmp` commit if you see it, I just want to run our test suite without opening a PR 😅).

Due to the merge conflicts I also had to rebase the branch, which means I would need to force push to your branch @grembo to get the changes to this PR. Would that be OK?

grembo:
Hi @lgfa29,

Thanks for picking this up again.

> I have been able to give it another pass and found some critical issues on the upgrade path.The biggest one is that [this line](https://github.com/hashicorp/nomad/pull/13343/files#diff-fdbf72fb40a559f0592c927e632f7f8a7178915920dd439de74e8798bc4acb88R351) will cause an existing alloc to fail when the Nomad agent is upgraded because the `/private` directory doesn't exist in that allocation.
> 

That's a good find, how are you planning to work around it (create the directory in case it is missing)?

> The other big problem is the lesson we learned the hard way in #17087, which is that we can't default to a non-zero value for a new attribute. What this means is that I had to change the attribute name from `file` to `disable_file` so that existing jobs are loaded in a backwards compatible way after an upgrade. Without this, if an old alloc fails it would be recreated with `file: false` even though we need it to be `true`.

My intention was to have file set to "false" by default in the long run (following the philosophy of safe defaults and opt-in into less secure/more open ones). I understand the backwards compatibility argument though - I have some ideas how such a change could be managed (e.g., have a "use_defaults_<version>" knob, that allows to opt-into a new set of secure defaults), but all of these add extra complications that would delay making progress. So if that's the way to move forward, so be it ;)

> 
> Since this PR has been stale for a while because of us, I don't think it would fair to ask @grembo to do all of these changes, so I implemented the fixes in https://github.com/hashicorp/nomad/compare/vault-token-file-rebased (ignore that `tmp` commit if you see it, I just want to run our test suite without opening a PR 😅).
> 
> Due to the merge conflicts I also had to rebase the branch, which means I would need to force push to your branch @grembo to get the changes to this PR. Would that be OK?

Yes, please go ahead

lgfa29:
For reference, this is how I tested the upgrade path.

1. Start dev Vault agent and login with root token.
    ```
    $ vault server -dev
    $ vault login
    ```
1. Create Nomad server role,  jobs policy, and secret.
    ```json
    {
      "disallowed_policies": "nomad-server",
      "token_explicit_max_ttl": 0,
      "name": "nomad-cluster",
      "orphan": true,
      "token_period": 259200,
      "renewable": true
    }
    ```
    ```hcl
    path "secret/data/*" {
      capabilities = ["read"]
    }
    ```
    ```
    $ vault policy write nomad-job vault-job-policy.hcl
    $ vault write /auth/token/roles/nomad-cluster @nomad-cluster-role.json
    $ vault kv put -mount=secret test test=1
    ```
1. Start Nomad agent version 1.5.6 with this config.
    ```hcl
    data_dir = "/tmp/nomad-vault/data"
    
    server {
      enabled          = 1
      bootstrap_expect = 1
    }
    
    client {
      enabled = true
    }
    
    vault {
      enabled          = true
      address          = "http://localhost:8200"
      task_token_ttl   = "5m"
      create_from_role = "nomad-cluster"
      token            = "<VAULT ROOT TOKEN>"
    }
    ```
    ```
    $ /opt/hashicorp/nomad/1.5.6/nomad agent -config ./config.hcl
    ```
1. Run job.
    ```hcl
    job "example" {
      group "cache" {
        task "redis" {
          driver = "docker"
    
          config {
            image = "redis:7"
          }
    
          vault {
            policies = ["nomad-job"]
          }
    
          template {
            data        = <<EOF
    {{with secret "secret/test"}}{{.Data.data.test}}{{end}}
    EOF
            destination = "config.txt"
          }
        }
      }
    }
    ```
    ```
    $ nomad run example.hcl
    ```
1. Verify Vault token is written to disk.
    ```console
    $ tree /tmp/nomad-vault/data/alloc
    /tmp/nomad-vault/data/alloc
    └── 3e1318f6-0ee0-8725-ac72-ed1a928d6401
        ├── alloc
        │   ├── data
        │   ├── logs
        │   │   ├── redis.stderr.0
        │   │   └── redis.stdout.0
        │   └── tmp
        └── redis
            ├── config.txt
            ├── local
            ├── secrets
            │   ├── api.sock
            │   └── vault_token
            └── tmp
    
    9 directories, 5 files
    ```
1. Stop Nomad agent and start the custom binary.
    ```
    $ ~/go/bin/nomad agent -config ./config.hcl
    ```
1. Verify allocation is restored without changes (no new task event) and file system stays the same.
1. Update job to set `vault.disable_file = true`.
2. Run updated job and verify a new allocation is created with the right folder structure.
    ```console
    $ tree /tmp/nomad-vault/data/alloc
    /tmp/nomad-vault/data/alloc
    ├── 3e1318f6-0ee0-8725-ac72-ed1a928d6401
    │   ├── alloc
    │   │   ├── data
    │   │   ├── logs
    │   │   │   ├── redis.stderr.0
    │   │   │   └── redis.stdout.0
    │   │   └── tmp
    │   └── redis
    │       ├── config.txt
    │       ├── local
    │       ├── secrets
    │       │   └── vault_token
    │       └── tmp
    └── 7ea55a0d-c12c-5590-4263-2fe3e222cb18
        ├── alloc
        │   ├── data
        │   ├── logs
        │   │   ├── redis.stderr.0
        │   │   └── redis.stdout.0
        │   └── tmp
        └── redis
            ├── config.txt
            ├── local
            ├── private
            │   └── vault_token
            ├── secrets
            │   └── api.sock
            └── tmp
    
    19 directories, 9 files
    ```

lgfa29:
> That's a good find, how are you planning to work around it (create the directory in case it is missing)?

The fix was two-fold:
1. On `Prestart` I'm also looking for the token in `/secrets` path.
2. On `writeToken` I'm checking if the `/private` dir is missing, which indicates this is an old task, and then keep writing to `/secrets`.

You can check the diff here: https://github.com/hashicorp/nomad/commit/2f881b7664fcc0e4b8cbf9e20be6cde6b430aaef#diff-fdbf72fb40a559f0592c927e632f7f8a7178915920dd439de74e8798bc4acb88

I renamed the variables because I was getting confused which was which 😅 

Another thing that was missing that I forgot to mention before is that we need a destructive update when this flag changes because the alloc runner may need to recreate the task dir to create `/private` if it doesn't exist. This was a quick fix: https://github.com/hashicorp/nomad/commit/83eeb46eaf5ef53744465a6d3afe53a8bd13818b

> My intention was to have file set to "false" by default in the long run

Yeah, we can start discussions about changing the default behaviour, but that is a bit more complicated. We can open an issue once this is released to discuss it further.

> Yes, please go ahead

Awesome, will do, thanks!

client/allocrunner/taskrunner/task_runner_test.go
```diff
@@ -1672,6 +1677,119 @@ func TestTaskRunner_BlockForVaultToken(t *testing.T) {
...
+	go tr.Run()
+
+	testWaitForTaskToDie(t, tr)
+
+	// Verify task exited successfully.
+	finalState := tr.TaskState()
+	must.Eq(t, structs.TaskStateDead, finalState.State)
+	must.False(t, finalState.Failed)
+
+	// Verfiry token is in the private dir.
```
tgross, line 1719:
```suggestion
	// Verify token is in the private dir.
```

client/allocrunner/taskrunner/task_runner_test.go
```diff
@@ -1672,6 +1677,119 @@ func TestTaskRunner_BlockForVaultToken(t *testing.T) {
...
+
+	conf, cleanup := testTaskRunnerConfig(t, alloc, task.Name)
+	defer cleanup()
+
+	// Remove private dir and write the Vault token to the secrets dir to
+	// simulate an old task.
+	err := conf.TaskDir.Build(false, nil)
+	must.NoError(t, err)
+	err = os.Remove(conf.TaskDir.PrivateDir)
+	must.NoError(t, err)
```
tgross, line 1752: As discussed in Slack, this is failing but isn't going to work because it's not doing the unlink we need (ref [`fs_linux.go#L81-L92`](https://github.com/hashicorp/nomad/blob/main/client/allocdir/fs_linux.go#L81-L92)). We could maybe just lift that function into this test but the test is going to be pointless immediately after 1.6.0 ships so I'd say let's just remove this test. We've seen it work E2E.

tgross (approved):
LGTM code-wise, once comments are addressed.

website/content/docs/concepts/filesystem.mdx
```diff
@@ -68,6 +70,11 @@ allocation directory like the one below.
     `NOMAD_TASK_DIR`. Note this is not the same as the "task working
     directory". This directory is private to the task.
 
+  - **«taskname»/private/**: This directory is used by Nomad to store private files
+    related to the allocation, such as Vault tokens, that are not always shared with tasks
+    when using `image` isolation. The contents of files in this directory cannot be read
+    by the `nomad alloc fs` command.
```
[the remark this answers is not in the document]
lgfa29: Yeah, it's kind of a tricky situation. I think that, ideally, when `disable_file`is `true` the token would written to a path outside of the alloc dir, but within Nomad's `data_dir`, but that would require plumbing the token all the way to the template runner.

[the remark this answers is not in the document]
lgfa29: Made it extra scary 😄 
image
image

client/allocrunner/taskrunner/task_runner_test.go
lgfa29: We came up with a better solution in https://github.com/hashicorp/nomad/pull/13343/commits/371105a15c0b7046aaf7138b40793ad3c4314a82: restrict the test to Linux environments and unmount the path before deleting it.

lgfa29:
UI test failure is a known flake. I thought I had fixed it in https://github.com/hashicorp/nomad/pull/17676 but it seems like it's not 100% yet.

lgfa29 (approved):

lgfa29:
This should go out in Nomad 1.6.0. Thank you very much @grembo for the contribution!

PR #13354, opened 2022-06-13
website: fix redirects with fragments

lgfa29:
Vercel redirects don't support fragments in relative destination paths,
so an absolute URL must be specified instead.


thiskevinwang (approved):
This looks reasonable to me
Looping in @BRKalow / @hashicorp/web-platform  for additional eyes on this!

PR #13364, opened 2022-06-14
Return evalID if `-detach` flag is passed to job revert

hugespoon:
Make `nomad job revert -detach` do what the docs say it does.

### DIFF
--- a/command/job_revert.go
+++ b/command/job_revert.go
@@ -164,7 +164,13 @@ func (c *JobRevertCommand) Run(args []string) int {
 
 	// Nothing to do
 	evalCreated := resp.EvalID != ""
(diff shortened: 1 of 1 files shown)


tgross (approved):
LGTM, thanks @hugespoon. Can you add a changelog entry as well? To do so, add a file at `.changelog/13364.txt` with contents like:

~~~
```release-note:bug
cli: Fixed a bug where the evaluation ID was not returned from `job revert` when the `-detach` flag was used
```
~~~

@DerekStrickland this is a bug fix so we should mark it for backport.

hugespoon:
@tgross @DerekStrickland Thanks for the review. Added the changelog entry.

PR #13370, opened 2022-06-14
fixup changelog entry for backported regression fix

tgross:
The changelog entry for #13340 indicated it was an improvement. But on
discussion, it was determined that this was a workaround for a
regression. Update the changelog to make this clear.


DerekStrickland (approved):
LGTM

PR #13375, opened 2022-06-14
implement Encrypt/Decrypt methods of encrypter

tgross:
The `Encrypt` method generates an appropriately-sized nonce and uses
that buffer as the prefix for the ciphertext. This keeps the
ciphertext and nonce together for decryption, and reuses the buffer as
much as possible without presenting the temptation to reuse the
cleartext buffer owned by the caller.

We include the key ID as the "additional data" field that's used as an
extra input to the authentication signature, to provide additional
protection that a ciphertext originated with that key.

Refactors the locking for the keyring so that the public methods are
generally (with one commented exception) responsible for taking the
lock and then inner methods are assumed locked.

Also fixes a bug where we forgot to check the bytes read from cryptorand
when generating our encryption key.

### DIFF
--- a/nomad/structs/secure_variables.go
+++ b/nomad/structs/secure_variables.go
@@ -227,12 +227,13 @@ func NewRootKey(algorithm EncryptionAlgorithm) (*RootKey, error) {
 
 	switch algorithm {
 	case EncryptionAlgorithmAES256GCM:
-		key := make([]byte, 32)
+		const keyBytes = 32
+		key := make([]byte, keyBytes)
 		n, err := cryptorand.Read(key)
 		if err != nil {
 			return nil, err
 		}
-		if n < 32 {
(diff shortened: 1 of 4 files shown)


nomad/structs/secure_variables.go
```diff
@@ -228,9 +228,13 @@ func NewRootKey(algorithm EncryptionAlgorithm) (*RootKey, error) {
 	switch algorithm {
 	case EncryptionAlgorithmAES256GCM:
 		key := make([]byte, 32)
```
angrycub, line 230: Maybe make the byte length shared between the length and the check. 1/2
```suggestion
		const keyBytes = 32
		key := make([]byte, keyBytes)
```

angrycub (approved):
Small edits. Nothing blocking. 👍

Issue #13382, opened 2022-06-15
chore: detach not always respected by CLI


DerekStrickland:
### Nomad version
1.3.x

### Issue

There are multiple CLI commands that have copy pasted code related to the `detach` flag that may or may not be behaving properly.  This was discovered when reviewing this community [PR](https://github.com/hashicorp/nomad/pull/13364). We should review and correct any remaining issues.

PR #13405, opened 2022-06-16
cli: fix detach handling

DerekStrickland:
Closes: #13382

This PR includes fixes to several CLI commands that are not handling the detach flag correctly. In each case, the command should be printing an Eval ID before exiting. The affected commands are:

- `deployment fail`
- `deployment promote`
- `deployment resume`
- `deployment unblock`
- `job promote`

### DIFF
--- a/.changelog/13405.txt
+++ b/.changelog/13405.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cli: Fixed remaining bugs where the evaluation ID was not printed when the `-detach` flag was passed
+``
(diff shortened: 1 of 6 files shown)


lgfa29 (approved):
LGTM.

Just missing a CHANGELOG entry. It may be good to have some tests for these as well if possible.

DerekStrickland:
> LGTM.
> 
> Just missing a CHANGELOG entry. It may be good to have some tests for these as well if possible.

Oops. Yep. Will add changelog entry. 

Regarding tests: I'm not sure how that can be achieved. I looked around for other examples but didn't find any. I'll take another pass on that.

PR #13406, opened 2022-06-16
ui:  make eye toggles only toggle on/off for their respective row

ChaiWithJai:
Enable `SecureVariableForm` to allow users to specify if the `value` in a key-value pair input group is obscured instead of obscuring all key-value pairs in the form.

image


philrenaud (approved):
Looks good to me! My only suggestion would be to rename `input-group` to something more specific; `secure-variable-value-group` or something similar. Otherwise we'll be popping into `input-group` on accident when looking for generalized `input` stuff in the future, I bet.

Issue #10289, opened 2021-04-02
Nomad scheduling halting


jonathanrcross:
### Nomad version
Nomad v1.0.4 (9294f35f9aa8dbb4acb6e85fa88e3e2534a3e41a)

### Operating system and Environment details
Centos 7
Linux <hostname> 3.10.0-1160.15.2.el7.x86_64 #1 SMP Wed Feb 3 15:06:38 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

### Issue
We are seeing the Nomad brokers suddenly stop or have blocked evaluations (we think), although we have capacity capable of fulfilling the required constraints asked. Attempting to restart already running services seems to also be blocked. As soon as we restart the leader, scheduling resumes (and allocates) as normal with no changes in the available clients. So this has us thinking since nomad is optimistically concurrent that is certainly a leader related issue but either with the evaluation broker or plan broker, maybe some mutex /lockingissue or the leader is somehow unresponsive but responsive (lots of guesses)? Below are some of the observations we've tracked and will be sending an email with pprofs, leader log and operator debug archive.

### Observations
![Screen Shot 2021-04-02 at 8 45 08 AM](https://user-images.githubusercontent.com/16668441/113416903-44932600-9390-11eb-99b1-ea6bc2baac13.png)
![Screen Shot 2021-04-02 at 8 45 35 AM](https://user-images.githubusercontent.com/16668441/113416906-465ce980-9390-11eb-8ac4-3d59a62ab822.png)
![Screen Shot 2021-04-02 at 8 45 50 AM](https://user-images.githubusercontent.com/16668441/113416913-48bf4380-9390-11eb-8f3b-834a0201e8cf.png)

### Sever configuration
```hcl

autopilot {
  cleanup_dead_servers = true
  last_contact_threshold = "4s"
}

server {
  enabled = true
  bootstrap_expect = 3
  encrypt = "..."
  server_join {
    ...
  }

  node_gc_threshold = "15s"
  job_gc_interval = "1m"
  job_gc_threshold = "30s"

  heartbeat_grace = "15s"
  max_heartbeats_per_second = 45.0
  raft_multiplier = 7
}
```

### Reproduction steps
Unable to reproduce

#### Actual Result
All Jobs are stuck in pending, restarting a job also does not take affect.

drewbailey:
@jonathanrcross Sorry that you are running into this and thank you for the detailed report and logs. 

[...]

PR #13407, opened 2022-06-16
fix deadlock in plan_apply

tgross:
The plan applier has to get a snapshot with a minimum index for the
plan it's working on in order to ensure consistency. Under heavy raft
loads, we can exceed the timeout. When this happens, we hit a bug
where the plan applier blocks waiting on the `indexCh` forever, and
all schedulers will block in `Plan.Submit`.

Closing the `indexCh` when the `asyncPlanWait` is done with it will
prevent the deadlock without impacting correctness of the previous
snapshot index.

---

Note for reviewers: the test code here can't actually exercise this bug,
because there's no interface for us to introduce this timeout. But this
code path was not covered by unit testing, so with these tests we at least
exercise the normal state.

6fab937 is how we verified the behavior and that the fix works, but we'll
want to circle back later to build out some fault injection into the state
store code at some point.

The deadlocked goroutine stack looks like this:

```
goroutine 625541538 [chan receive, 514 minutes]:
github.com/hashicorp/nomad/nomad.(*planner).planApply(0xc000a410a0)
	github.com/hashicorp/nomad/nomad/plan_apply.go:156 +0x35e
created by github.com/hashicorp/nomad/nomad.(*Server).establishLeadership
	github.com/hashicorp/nomad/nomad/leader.go:250 +0x21e
```

cc @jazzyfresh @lgfa29 @mikenomitch 
Fixes https://github.com/hashicorp/nomad/issues/10289

### DIFF
--- a/nomad/plan_apply.go
+++ b/nomad/plan_apply.go
@@ -154,12 +154,12 @@ func (p *planner) planApply() {
 		// This also limits how out of date our snapshot can be.
 		if planIndexCh != nil {
 			idx := <-planIndexCh
+			planIndexCh = nil
 			prevPlanResultIndex = max(prevPlanResultIndex, idx)
 			snap, err = p.snapshotMinIndex(prevPlanResultIndex, pending.plan.SnapshotIndex)
 			if err != nil {
 				p.logger.Error("failed to update snapshot state", "error", err)
 				pending.respond(nil, err)
-				planIndexCh = nil
 				continue
 			}
 		}
--- a/nomad/plan_apply.go
+++ b/nomad/plan_apply.go
@@ -153,12 +153,15 @@ func (p *planner) planApply() {
 		// Ensure any parallel apply is complete before starting the next one.
 		// This also limits how out of date our snapshot can be.
 		if planIndexCh != nil {
+			fmt.Println("waiting for idx...") // DEBUG
 			idx := <-planIndexCh
+			fmt.Println("got index", idx) // DEBUG
 			prevPlanResultIndex = max(prevPlanResultIndex, idx)
 			snap, err = p.snapshotMinIndex(prevPlanResultIndex, pending.plan.SnapshotIndex)
 			if err != nil {
 				p.logger.Error("failed to update snapshot state", "error", err)
 				pending.respond(nil, err)
+				planIndexCh = nil
 				continue
 			}
 		}
@@ -368,14 +371,12 @@ func updateAllocTimestamps(allocations []*structs.Allocation, timestamp int64) {
 func (p *planner) asyncPlanWait(indexCh chan<- uint64, future raft.ApplyFuture,
 	result *structs.PlanResult, pending *pendingPlan) {
 	defer metrics.MeasureSince([]string{"nomad", "plan", "apply"}, time.Now())
+	defer close(indexCh)
 
 	// Wait for the plan to apply
 	if err := future.Error(); err != nil {
 		p.logger.Error("failed to apply plan", "error", err)
 		pending.respond(nil, err)
-
-		// Close indexCh on error
-		close(indexCh)
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


nomad/state/state_store.go
```diff
@@ -222,6 +225,13 @@ func (s *StateStore) SnapshotMinIndex(ctx context.Context, index uint64) (*State
 	var retries uint
 	var retryTimer *time.Timer
 
+	// DEBUG: this is to introduce a one-time timeout
+	if index == 7 && stop {
+		stop = false
+		time.Sleep(6000 * time.Millisecond)
+		return nil, ctx.Err()
+	}
```
tgross, line 233: TODO: remove this once we have a test we can run that'll exercise this code without it.

nomad/plan_apply.go
```diff
@@ -368,14 +373,12 @@ func updateAllocTimestamps(allocations []*structs.Allocation, timestamp int64) {
 func (p *planner) asyncPlanWait(indexCh chan<- uint64, future raft.ApplyFuture,
 	result *structs.PlanResult, pending *pendingPlan) {
 	defer metrics.MeasureSince([]string{"nomad", "plan", "apply"}, time.Now())
+	defer close(indexCh)
```
schmichael, line 376: Would this alone have fixed the deadlock? I think so and since we always use this index inside a `max(...)` it seems like an always safe operation.

Oof, good lesson about writing defensive code the first time round.

schmichael (approved):

nomad/plan_apply.go
robloxrob: You never know if you will have time to come back and make it better in a second pass 🤣

tgross: > Would this alone have fixed the deadlock?

Yes! That's how we did the initial test and fix. Nil'ing the channel gets us two additional things:
* It's belt-and-suspenders so that we know there's no possible way we're still polling on the channel.
* It locally communicates to the reader that the channel is being discarded, as opposed to having to go read the `asyncPlanWait` code further down (and doing so is super cheap).

nomad/plan_apply.go
```diff
@@ -159,6 +159,7 @@ func (p *planner) planApply() {
 			if err != nil {
 				p.logger.Error("failed to update snapshot state", "error", err)
 				pending.respond(nil, err)
+				planIndexCh = nil
```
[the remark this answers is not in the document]
tgross: > Should this be on L157 right after we receive from the chan?

Yup, good catch. I was focused on the error branch because there's where the bug showed up, but there's no reason to keep it once we've received a value.

> To put another way: I think `planIndexCh` is always a one-shot chan. It will always receive exactly one index value and never get reused.

Yes, agreed!

> If you think that's accurate I wonder if we should wrap planIndexCh in a little struct with a method like:
> ...
> We also use `planIndexCh == nil` to signal "no outstanding plan application", so our struct would probably need a helper method for that too. Unsure if the extra struct would simplify or complicate reading this code... just an idea.

Hm... I like this idea in principle. But the logic here leans heavily on the specific semantics of channels (ex. a closed channel isn't nil) and wrapping it in a struct just seems like it'll obscure that by moving the logic away from where we're using it. The fallthrough `select` on line 104-113 makes this especially gross because it relies on the fact that we get 0 on error, so we can't even call `recv()` and check for 0 there so it needs it's own `index, ok := recvOrDefault()` function.

<details><summary>some example code</summary>

```go

type asyncPlanIndex struct {
	planIndexCh chan uint64
}

func (a *asyncPlanIndex) hasOutstandingPlan() bool {
	return a.planIndexCh == nil
}

func (a *asyncPlanIndex) recv() uint64 {
	v := <-a.planIndexCh
	a.planIndexCh = nil
	return v
}

func (a *asyncPlanIndex) recvOrDefault() (uint64, bool) {
	select {
	case v := <-a.planIndexCh:
		return v, true
	default:
		return 0, false
	}
}

func (a *asyncPlanIndex) send(v uint64) {
	a.planIndexCh <- v
	close(a.planIndexCh)
}

```

</details>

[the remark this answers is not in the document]
schmichael: > But the logic here leans heavily on the specific semantics of channels

Oof yeah, this is a pattern to watch out for in the future. Not always a problem but clearly made this code more fragile than if a little effort had been put into encapsulating the logic in a little purpose-built struct.

PR #13414, opened 2022-06-17
ui:  update  component api to accept callback fns to be fired on cancel/prompt

ChaiWithJai:
This PR removes the `Edit` `LinkTo` component from the template when the `Delete` button is clicked.

image

Side Effect:  This PR updates the `TwoStepButton` component API to accept an `onPrompt` callback function to be fired with the `promptForConfirmation` action and adds logic for the `onCancel` callback function to be fired when the `TwoStepButton` registers on outside click event.

### DIFF
--- a/ui/app/components/two-step-button.js
+++ b/ui/app/components/two-step-button.js
@@ -24,6 +24,7 @@ export default class TwoStepButton extends Component {
   inlineText = false;
   onConfirm() {}
   onCancel() {}
+  onPrompt() {}
 
   state = 'idle';
   @equal('state', 'idle') isIdle;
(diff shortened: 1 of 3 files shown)


ui/app/controllers/variables/variable/index.js
```diff
@@ -10,6 +11,18 @@ export default class VariablesVariableIndexController extends Controller {
...
 
+  @tracked isDeleting = false;
+
+  @action
+  onDeletePrompt() {
+    this.isDeleting = true;
+  }
+
+  @action
+  onCancel() {
```
philrenaud, line 22: Suggestion: rename this to something more specific. We may have more functionality than just "delete" in the future, and `onCancel` would be ambiguous then.

Maybe `onDeleteCancel()`

ui/app/components/two-step-button.js
```diff
@@ -33,6 +34,9 @@ export default class TwoStepButton extends Component {
     while (true) {
       let ev = yield waitForEvent(document.body, 'click');
       if (!this.element.contains(ev.target) && !this.awaitingConfirmation) {
+        if (this.onCancel) {
+          this.onCancel();
+        }
```
philrenaud, line 39: Surprised this is needed here! onCancel() isn't a new convention for two-step buttons. Has it just never been used in practice before this?

philrenaud (approved):

PR #13415, opened 2022-06-17
Path Tree compaction refactor

philrenaud:
- Moves folder compaction (listing folder "foo" in a directory as "foo/bar" when it only has one child descendent named "bar") to render time instead of compute time
- Removes secondary functions from the Path Tree util (compact paths, pathToObject) and standardizes its types

In practice, this lets you traverse the directory tree structure for variables at a compacted path without issue.

So if you have a single variable at `a/b/c/d/e/myVar`, you previously would only be able to see anything at `a/b/c/d/e`; now you could traverse to `a/b/c` if you so chose, and `d/e/` would be listed therein.

### DIFF
--- a/ui/app/utils/compact-path.js
+++ b/ui/app/utils/compact-path.js
@@ -5,8 +5,6 @@
  * @returns
  */
 export default function compactPath(branch, name = '') {
-  // console.log('lup', branch);
-  // name = `${name}/${branch.name}`;
   let { children, files } = branch;
   if (children && Object.keys(children).length === 1 && !files.length) {
     const [key] = Object.keys(children);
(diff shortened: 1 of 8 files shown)


ui/app/utils/compact-path.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Takes a branch created by our path-tree, and if it has only a single directory as descendent and no files, compacts it down to its terminal folder (the first descendent folder with either files or branching directories)
+ * Uses tail recursion
+ * @param {import("./path-tree").NestedPathTreeNode} branch
+ * @returns
+ */
+export default function compactPath(branch, name = '') {
```
mikenomitch, line 7: 👨‍🍳 💋 👌

mikenomitch (commented):
LGTM (went over w/ Phil on Zoom & briefly reviewed code)

Note: I can't officially +1 stuff

PR #13421, opened 2022-06-17
Track plan rejection history and automatically mark clients as ineligible

### LINKED ISSUE #12920: Automatically down-rank or quarantine nodes with failing placements
### Proposal

When the scheduler attempts to place a task on a node, and the allocation fails, that node should be ranked lower (or quarantined) the next time the scheduler is looking for a node to place.

### Use-cases

In some instances, the Nomad scheduler repeatedly selects the same node to schedue on, but for some reason deployments on that node fail repeatedly. This can be seen in [this issue](https://github.com/hashicorp/nomad/issues/9506). In these cases, the entire cluster can become blocked due to several bad nodes. This feature would make Nomad far more resilient in when there are bad nodes that the scheduler does not know are bad.

### Attempted Solutions

A person can monitor for failed placements and/or blocked evaluations and intervene. This takes a lot of effort and knowledge for the Nomad operator to do, and shouldn't be necessary in the first place.

### LINKED ISSUE #13017: Automatically mark repeatedly rejected nodes in plans as ineligible
The plan applier should track nodes rejected from plans as repeatedly rejected nodes are a sign of the pathological #9506 `plan for node rejected` bug.

### Bug

Docs: https://www.nomadproject.io/docs/operations/monitoring-nomad#progress
Related issue: #9506 

Since plans for nodes may be rejected for a number of legitimate and incorrect reasons, we should be defensive about handling the pathological condition of workers and the plan applier disagreeing about the validity of a node. Nomad does optimistic concurrent scheduling: multiple concurrent scheduling workers may submit conflicting plans from which the serializing plan applier must choose a "winner" (eg select both would exhaust the cpu resources on the node). The worker whose plan "loses" then refreshes their cluster state and attempts to schedule the workload again.

Unfortunately this means the workers and plan applier have different logic for assessing the feasibility of nodes. Any slight divergence in their node feasibility assessment may lead to a pathological loop:

1. Worker submits a Plan containing placements for Node A
2. Plan Applier validates Plan and disagrees that Node A is a valid placement
3. Worker sees the last plan was only partially applied so retries, still thinking Node A is valid
4. Goto Step 2

While there are various backoffs and limits to cause evaluations to eventually clear, this can take *minutes* due to backoffs and retries. In a busy cluster, especially a cluster recovering from a partition or other event that causes a flurry of activity, minutes-per-evaluation can mean it takes days, weeks, or even months to clear. Even an eval queue that's effectively stuck on these pathological evals for hours is enough to cause severe service disruption.

### Proposal

Since the existing eval backoff and retry mechanism is insufficient to deal with "bad" nodes (nodes for whom the workers and plan applier disagree about their feasibility), a new mechanism for detecting this pathological condition must be added. Since it is impossible to differentiate between a correctness issue (worker/applier disagreement) and intentional conflict resolution due to concurrent schedulers, *we must rely on heuristics to make progress in the face of bad nodes.*

The Plan Applier is the ideal place to implement the heuristic bad node detection mechanism for 2 reasons:

1. Since the Plan Applier is single threaded and sees all plans, no coordination around the heuristic's state is necessary across goroutines or processes. The Plan Applier can track bad nodes and invalidate them without having to coordinate such tracking with Workers. Workers can remain unchanged.
2. The Plan Applier commits Plan Results to Raft while Workers avoid making Raft commits until an Evaluation is fully processed to allow them to have as little coordination and overhead as possible. Therefore whatever decisions the heuristic mechanism makes can be committed to Raft by the Plan Applier and be *globally visible not only to the workers and plan applier, but to users who need to observe issues in their cluster.*

#### Heuristic Mechanism

The intentional reason placements are rejected from nodes is to account for the following race condition possible due to Nomad's optimistically concurrent scheduling:

1. Worker `W1` dequeues Eval `E1` for a low priority batch job
2. Worker `W2` dequeues Eval `E2` for a high priority system job
3. `W1` submits a plan with placement on Node `N1` that uses all remaining cpu on `N1`
4. `W2` working concurrently with `W1` submits a plan with a placement on `N1` that _also_ uses all remaining cpu on `N1`
5. The Plan Applier dequeues `W1`'s plan first, accepts it.

There are actually a number of properties of Nomad and its common use cases that *make these sort of conflicts more likely:*

1. New nodes coming up create evals for *all* system jobs
2. Treating Nomad as a work queue by submitting more low priority batch jobs than a cluster has capacity to process concurrently is a common and suggested use case
3. On modern servers with many cores it is likely when a node is added all of the system evals and unblocked batch evals may be processed by concurrent workers more or less simultaneously
4. Nomad's preemption feature is a great solution for the above case of a new node receiving low priority batch placements before high priority system jobs. However since preemption is costly to calculate Nomad puts it in the concurrent workers for optimal scalability. While this makes preemption and the plan applier's node rejection work well together, it does complicate the heuristic for detecting pathological node rejections as _some_ level of rejections is expected for preemption to work optimally.
5. Nomad's default bin packing algorithm means concurrent workers are likely to select the same nearly-at-capacity nodes for new placements, increasing the likelihood that some will be rejected by the plan applier.

All of this to say: *the plan applier is expected to reject some node placements for correctness, so how do we detect pathological node rejections?*

The general heuristic should be:

> If a node is rejected N times in D second, mark it as `ineligible`.

The heuristic will have the following parameters:

1. Number of nodes to track rejected stats for to cap the amount of memory used.
2. Time Window (D) to track rejections within.
3. Threshold (N) for rejections per Window before we mark it as ineligible.

Time Window (D) and Threshold (N) should be configurable.

The maximum number of nodes tracked (`#1`) does *not* need to be configurable and instead hardcoded to something like 50 nodes. Even in very large clusters (thousands of nodes) the maximum number of bad nodes ever observed at once was 20-30. Even if the cap is too low to track all bad nodes simultaneously, as bad nodes are marked ineligible they will cease being tracked and make space for more nodes to be tracked.

The `PlanResult` struct can encode the nodes to be marked ineligible:

```go
type PlanResult struct {
	// ...

	// IneligibleNodes are nodes the plan applier has repeatedly rejected
	// placements for and should therefore be considered ineligible by
	// workers to avoid retrying them repeatedly.
	IneligibleNodes []string
}
```

The FSM handler for the `ApplyPlanResultsRequestType` Raft log message will need to be updated to mark nodes ineligible.

When retrying evaluations workers are sent the `PlanResult`, but it is **not** necessary for the worker to explicitly read IneligibleNodes because the RefreshIndex will cause the worker to use a snapshot of cluster state that *includes the nodes as being marked ineligible.*

#### Visibility

A metric and log line should be emitted every time a new node is marked ineligible by the heuristic.

The existing `nomad.plan.node_rejected` metric and associated log line should be sufficient to guide users on how best to tune the Time Window and Threshold parameters, although opportunities to improve visibility should be taken.

lgfa29:
Plan rejections occur when the scheduler work and the leader plan
applier disagree on the feasibility of a plan. This may happen for valid
reasons: since Nomad does parallel scheduling, it is expected that
different workers will have a different state when computing placements.

As the final plan reaches the leader plan applier, it may no longer be
valid due to a concurrent scheduling taking up intended resources. In
these situations the plan applier will notify the worker that the plan
was rejected and that they should refresh their state before trying
again.

In some rare and unexpected circumstances it has been observed that
workers will repeatedly submit the same plan, even if they are always
rejected.

While the root cause is still unknown this mitigation has been put in
place. The plan applier will now track the history of plan rejections
per client and include in the plan result a list of node IDs that should
be set as ineligible if the number of rejections in a given time window
crosses a certain threshold. The window size and threshold value can be
adjusted in the server configuration.

Closes #13017
Closes #12920

-----------------------

_Note for reviewers:_ since we can't yet reliably reproduce this bug, the way I tested this was by applying [this patch](https://gist.github.com/lgfa29/44a3420fa588248fe39c942ee87131c0) that causes a plan to be rejected if it is evaluated by a server running the env var `CRASH` set it's for a client with a name that starts with `crash`.

So, after applying the patch, start a 3 server cluster with one of them having the `CRASH` env var set and make sure this server becomes the leader. Start a client with the name starting with `crash` and run a job.

Monitoring the log you should see the plan rejection messages and, after a few minutes, the client will become ineligible. You can then start a client without `crash` in the name to verify that the job scheduling will proceed to the new client.

### DIFF
--- a/command/agent/config.go
+++ b/command/agent/config.go
@@ -556,7 +556,7 @@ type RaftBoltConfig struct {
 // tracker.
 type PlanRejectionTracker struct {
 	// Enabled controls if the plan rejection tracker is active or not.
-	Enabled bool `hcl:"enabled"`
+	Enabled *bool `hcl:"enabled"`
 
 	// NodeThreshold is the number of times a node can have plan rejections
 	// before it is marked as ineligible.
@@ -582,8 +582,8 @@ func (p *PlanRejectionTracker) Merge(b *PlanRejectionTracker) *PlanRejectionTrac
 		return &result
 	}
 
-	if b.Enabled {
-		result.Enabled = true
+	if b.Enabled != nil {
+		result.Enabled = b.Enabled
 	}
 
 	if b.NodeThreshold != 0 {
@@ -1037,6 +1037,7 @@ func DefaultConfig() *Config {
 			RaftProtocol:      3,
 			StartJoin:         []string{},
 			PlanRejectionTracker: &PlanRejectionTracker{
+				Enabled:       helper.BoolToPtr(false),
 				NodeThreshold: 100,
 				NodeWindow:    5 * time.Minute,
 			},
--- a/command/agent/config.go
+++ b/command/agent/config.go
@@ -503,6 +503,10 @@ type ServerConfig struct {
 	// This value is ignored.
 	DefaultSchedulerConfig *structs.SchedulerConfiguration `hcl:"default_scheduler_config"`
 
+	// PlanRejectionTracker configures the node plan rejection tracker that
+	// detects potentially bad nodes.
+	PlanRejectionTracker *PlanRejectionTracker `hcl:"plan_rejection_tracker"`
+
 	// EnableEventBroker configures whether this server's state store
 	// will generate events for its event stream.
 	EnableEventBroker *bool `hcl:"enable_event_broker"`
@@ -546,6 +550,46 @@ type RaftBoltConfig struct {
 	//
 	// Default: false.
 	NoFreelistSync bool `hcl:"no_freelist_sync"`
+}
+
+// PlanRejectionTracker is used in servers to configure the plan rejection
+// tracker.
+type PlanRejectionTracker struct {
+	// NodeThreshold is the number of times a node can have plan rejections
+	// before it is marked as ineligible.
+	NodeThreshold int `hcl:"node_threshold"`
+
+	// NodeWindow is the time window used to track active plan rejections for
+	// nodes.
+	NodeWindow    time.Duration
+	NodeWindowHCL string `hcl:"node_window" json:"-"`
+
+	// ExtraKeysHCL is used by hcl to surface unexpected keys
+	ExtraKeysHCL []string `hcl:",unusedKeys" json:"-"`
+}
+
+func (p *PlanRejectionTracker) Merge(b *PlanRejectionTracker) *PlanRejectionTracker {
+	if p == nil {
+		return b
+	}
+
+	result := *p
+
+	if b == nil {
+		return &result
+	}
+
+	if b.NodeThreshold != 0 {
+		result.NodeThreshold = b.NodeThreshold
+	}
+
+	if b.NodeWindow != 0 {
+		result.NodeWindow = b.NodeWindow
+	}
+	if b.NodeWindowHCL != "" {
+		result.NodeWindowHCL = b.NodeWindowHCL
+	}
+	return &result
 }
 
 // Search is used in servers to configure search API options.
@@ -985,6 +1029,10 @@ func DefaultConfig() *Config {
 			EventBufferSize:   helper.IntToPtr(100),
 			RaftProtocol:      3,
 			StartJoin:         []string{},
+			PlanRejectionTracker: &PlanRejectionTracker{
+				NodeThreshold: 15,
+				NodeWindow:    10 * time.Minute,
+			},
 			ServerJoin: &ServerJoin{
 				RetryJoin:        []string{},
 				RetryInterval:    30 * time.Second,
@@ -1584,6 +1632,10 @@ func (s *ServerConfig) Merge(b *ServerConfig) *ServerConfig {
 
 	if b.EventBufferSize != nil {
 		result.EventBufferSize = b.EventBufferSize
+	}
+
+	if b.PlanRejectionTracker != nil {
+		result.PlanRejectionTracker = result.PlanRejectionTracker.Merge(b.PlanRejectionTracker)
 	}
 
 	if b.DefaultSchedulerConfig != nil {
--- a/command/agent/config.go
+++ b/command/agent/config.go
@@ -555,6 +555,9 @@ type RaftBoltConfig struct {
 // PlanRejectionTracker is used in servers to configure the plan rejection
 // tracker.
 type PlanRejectionTracker struct {
+	// Enabled controls if the plan rejection tracker is active or not.
+	Enabled bool `hcl:"enabled"`
+
 	// NodeThreshold is the number of times a node can have plan rejections
 	// before it is marked as ineligible.
 	NodeThreshold int `hcl:"node_threshold"`
@@ -577,6 +580,10 @@ func (p *PlanRejectionTracker) Merge(b *PlanRejectionTracker) *PlanRejectionTrac
 
 	if b == nil {
 		return &result
+	}
+
+	if b.Enabled {
+		result.Enabled = true
 	}
 
 	if b.NodeThreshold != 0 {
@@ -1030,8 +1037,8 @@ func DefaultConfig() *Config {
 			RaftProtocol:      3,
 			StartJoin:         []string{},
 			PlanRejectionTracker: &PlanRejectionTracker{
-				NodeThreshold: 15,
-				NodeWindow:    10 * time.Minute,
+				NodeThreshold: 100,
+				NodeWindow:    5 * time.Minute,
 			},
 			ServerJoin: &ServerJoin{
 				RetryJoin:        []string{},
--- a/nomad/plan_apply_node_tracker.go
+++ b/nomad/plan_apply_node_tracker.go
@@ -12,8 +12,7 @@ import (
 )
 
 type BadNodeTracker interface {
-	IsBad(string) bool
-	Add(string)
+	Add(string) bool
 	EmitStats(time.Duration, <-chan struct{})
 }
 
@@ -21,9 +20,8 @@ type BadNodeTracker interface {
 // used when tracking is disabled.
 type NoopBadNodeTracker struct{}
 
-func (n *NoopBadNodeTracker) Add(string)                               {}
 func (n *NoopBadNodeTracker) EmitStats(time.Duration, <-chan struct{}) {}
-func (n *NoopBadNodeTracker) IsBad(string) bool {
+func (n *NoopBadNodeTracker) Add(string) bool {
 	return false
 }
 
@@ -88,50 +86,24 @@ func NewCachedBadNodeTracker(logger hclog.Logger, config CachedBadNodeTrackerCon
 	}, nil
 }
 
-// IsBad returns true if the node has more rejections than the threshold within
-// the time window.
-func (c *CachedBadNodeTracker) IsBad(nodeID string) bool {
-	logger := c.logger.With("node_id", nodeID)
-	logger.Debug("checking if node is bad")
-
+// Add records a new rejection for a node and returns true if the number of
+// rejections reaches the threshold.
+//
+// If it's the first time the node is added it will be included in the internal
+// cache. If the cache is full the least recently updated or accessed node is
+// evicted.
+func (c *CachedBadNodeTracker) Add(nodeID string) bool {
 	value, ok := c.cache.Get(nodeID)
 	if !ok {
-		logger.Debug("node not in cache")
-		return false
-	}
-
-	// Limit the number of nodes we report as bad to avoid mass assigning nodes
-	// as ineligible, but do it after Get to keep the cache entry fresh.
-	if !c.limiter.Allow() {
-		logger.Info("returning false due to rate limiting")
-		return false
-	}
-
-	stats := value.(*badNodeStats)
-	score := stats.score()
-
-	if score >= c.threshold {
-		logger.Debug("node is bad", "score", score)
-		return true
-	}
-
-	logger.Debug("node is not bad", "score", score)
-	return false
-}
-
-// Add records a new rejection for node. If it's the first time a node is added
-// it will be included in the internal cache. If the cache is full the least
-// recently updated or accessed node is evicted.
-func (c *CachedBadNodeTracker) Add(nodeID string) {
-	value, ok := c.cache.Get(nodeID)
-	if !ok {
-		value = newBadNodeStats(c.window)
+		value = newBadNodeStats(nodeID, c.window)
 		c.cache.Add(nodeID, value)
 	}
-
 	stats := value.(*badNodeStats)
-	score := stats.record()
-	c.logger.Debug("adding node plan rejection", "node_id", nodeID, "score", score)
+
+	now := time.Now()
+	stats.record(now)
+
+	return c.isBad(now, stats)
 }
 
 // EmitStats generates metrics for the bad nodes being currently tracked. Must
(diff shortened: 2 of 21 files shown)


command/agent/config.go
```diff
@@ -985,6 +1029,10 @@ func DefaultConfig() *Config {
 			EventBufferSize:   helper.IntToPtr(100),
 			RaftProtocol:      3,
 			StartJoin:         []string{},
+			PlanRejectionTracker: &PlanRejectionTracker{
+				NodeThreshold: 15,
+				NodeWindow:    10 * time.Minute,
+			},
```
lgfa29, line 1035: I'm also not sure about these defaults. The workers will retry the plan with exponential back-off, so depending on the cluster load and eval queue size it may take a while for the node to reach the threshold.

So I picked these values based on a threshold number that is likely to indicate a problem (it could maybe even be a little higher?), not just normal plan rejections, and a time window that encompasses a significant part of the rejection history but that will probably have expired by the time operators detected, drained, and restarted the client. 

A short time window may risk the threshold never being hit and a large one risks a normal plan rejection triggering the ineligibility since the node score can still be high.

nomad/structs/structs.go
```diff
@@ -11403,8 +11419,9 @@ type PlanResult struct {
 
 // IsNoOp checks if this plan result would do nothing
 func (p *PlanResult) IsNoOp() bool {
-	return len(p.NodeUpdate) == 0 && len(p.NodeAllocation) == 0 &&
-		len(p.DeploymentUpdates) == 0 && p.Deployment == nil
+	return len(p.IneligibleNodes) == 0 && len(p.NodeUpdate) == 0 &&
```
lgfa29, line 11422: This new check will cause the [no-op fast-path](https://github.com/hashicorp/nomad/blob/v1.3.1/nomad/plan_apply.go#L147-L151) to be skipped in cases there it previously wouldn't, so it may be worth double checking if nothing will break.

website/content/docs/configuration/server.mdx
```diff
@@ -238,6 +242,28 @@ server {
   section for more information on the format of the string. This field is
   deprecated in favor of the [server_join stanza][server-join].
 
+### `plan_rejection_tracker` Parameters
+
+The leader plan rejection tracker can be adjusted to prevent evaluations from
+getting stuck due to always being scheduled to a client that may have an
+unexpected and undetected issues. Refer to [Monitoring
```
schmichael, line 249:
```suggestion
unexpected issue. Refer to [Monitoring
```

schmichael (commented):
Reviewed all but the core algorithm. Will get that done ASAP.

command/agent/config.go
schmichael: From our out of band discussion: let's leave this in the config file. If you're fiddling with it live, you can just mark bad nodes ineligible with existing commands and fiddle with these settings after your cluster is healthy.

Hopefully we can find good enough defaults that no one needs to worry about this anyway.

nomad/plan_apply_node_tracker.go
```diff
@@ -0,0 +1,203 @@
...
+	value, ok := c.cache.Get(nodeID)
+	if !ok {
+		logger.Debug("node not in cache")
+		return false
+	}
+
+	// Limit the number of nodes we report as bad to avoid mass assigning nodes
+	// as ineligible, but do it after Get to keep the cache entry fresh.
+	if !c.limiter.Allow() {
+		logger.Info("returning false due to rate limiting")
```
tgross, line 106: Nitpick: info is kind of a weird log level because it's not clear how actionable it is. I try to reserve it for the top-level HTTP API and the agent startup. This strikes me as something that should be a warning. 

I also wonder if we want to put the rate limit here or after we score? Checking the score seems ok to do (it's all cheap and in-memory and doesn't mutate state).

nomad/plan_apply_node_tracker.go
```diff
@@ -0,0 +1,203 @@
...
+		window:    config.Window,
+		threshold: config.Threshold,
+	}, nil
+}
+
+// IsBad returns true if the node has more rejections than the threshold within
+// the time window.
+func (c *CachedBadNodeTracker) IsBad(nodeID string) bool {
+	logger := c.logger.With("node_id", nodeID)
+	logger.Debug("checking if node is bad")
```
tgross, line 95: Having a debug log line for every branch of this method seems really noisy, and on a busy cluster they'll get interleaved with other logs. Maybe one log line at the caller would reduce the noise?

nomad/plan_apply_node_tracker.go
```diff
@@ -0,0 +1,203 @@
...
+
+	value, ok := c.cache.Get(nodeID)
+	if !ok {
+		logger.Debug("node not in cache")
+		return false
```
tgross, line 100: It looks like we should never hit this because calling `Add` first is an invariant of the design. We could return an error here instead, or even better change the construction of `Add` and `IsBad` so that we both add and check in the same call in `plan_apply.go`

nomad/plan_apply_node_tracker.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+	return count
+}
+
+// record adds a new entry to the stats history and returns the new score.
+func (s *badNodeStats) record() int {
+	now := time.Now()
+	s.history = append(s.history, now)
+	return s.score()
+}
```
tgross, line 203: On large clusters I'd expect the LRU would age-out most of the tracked nodes eventually. But I think there's a pathological case here where a stable cluster with 50 or fewer nodes will continue to grow the size of the `badNodeStats.history` forever, so long as none of the nodes get plan rejections more frequently than the window allows. 

It's a small bit of data (24 bytes each rejection), but we could eliminate it entirely if we dropped the old entries either when we called `record()` or when we called `score()`.

nomad/plan_apply_node_tracker_test.go
```diff
@@ -0,0 +1,155 @@
...
+	time.Sleep(window / 2)
+	stats.record()
+	require.Equal(t, 4, stats.score())
+
+	time.Sleep(window / 2)
+	require.Equal(t, 1, stats.score())
+
+	time.Sleep(window / 2)
+	require.Equal(t, 0, stats.score())
+}
```
tgross, line 155: Something that can make tests faster and more reliable is to have the `score()` and `record()` methods take a timestamp as their argument. The caller in `Add/IsBad` can pass it `time.Now()` so it works the same way, but it's testable by constructing timestamps in the test without needing to sleep.

tgross (changes requested):
This is looking great @lgfa29! I've left a few suggestions that could tighten it up.

website/content/docs/operations/monitoring-nomad.mdx
```diff
@@ -149,10 +149,29 @@ While it is possible for these log lines to occur infrequently due to normal
 cluster conditions, they should not appear repeatedly and prevent the job from
 eventually running (look up the evaluation ID logged to find the job).
 
-If this log *does* appear repeatedly with the same `node_id` referenced, try
+Nomad tracks the history of plan rejections per client and will mark it as
```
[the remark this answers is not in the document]
lgfa29: Ah yes, the `enabled = false` was added later and I forgot to update the docs. Thanks!

nomad/plan_apply_node_tracker.go
```diff
@@ -0,0 +1,203 @@
...
+
+// NoopBadNodeTracker is a no-op implementation of bad node tracker that is
+// used when tracking is disabled.
+type NoopBadNodeTracker struct{}
+
+func (n *NoopBadNodeTracker) Add(string)                               {}
+func (n *NoopBadNodeTracker) EmitStats(time.Duration, <-chan struct{}) {}
+func (n *NoopBadNodeTracker) IsBad(string) bool {
+	return false
+}
```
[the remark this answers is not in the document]
lgfa29: For the first implementation I was just setting the rate limiter to zero, but there was still quite a bit of code running in a critical path, so I chose this no-op approach instead. Can't have bugs if there is no code to run 😄

nomad/plan_apply_node_tracker.go
lgfa29: Yeah, good point. I think I will actually downgrade to `Debug` along with the others, there really isn't much value knowing this as it should be a temporary state. With `nomad monitor` is easy to consume the leader log in a lower log level if needed.

And nice catch with moving the rate limit. Calling `Allow` here will consume a token even if the score is below the threshold, which ultimately end up waste it.

nomad/plan_apply_node_tracker.go
lgfa29: Yeah that makes sense. I modified the interface so it only exposes `Add` and `EmitMetrics`. `isBad` now takes the object in the cache that `Add` created or retrieved, so there's no need for this check anymore 👍

nomad/plan_apply_node_tracker.go
lgfa29: Makes sense. I reduce the number of log messages and dropped their level to `Trace` since this should just be doing its job 😅 

If something was supposed to be happening but it's not we can `monitor` the leader.

nomad/plan_apply_node_tracker.go
lgfa29: Good catch! I made it so `score` will remove entries that happened before the window start, so they will never be considered any more.

nomad/plan_apply_node_tracker_test.go
lgfa29: Nice! I updated what I could, but I didn't find a way to do something like this with the rate limiter, so it still sleeps for 1s.

tgross (approved):
LGTM :+1:

command/agent/config.go
```diff
@@ -548,6 +552,53 @@ type RaftBoltConfig struct {
 	NoFreelistSync bool `hcl:"no_freelist_sync"`
 }
 
+// PlanRejectionTracker is used in servers to configure the plan rejection
+// tracker.
+type PlanRejectionTracker struct {
+	// Enabled controls if the plan rejection tracker is active or not.
+	Enabled bool `hcl:"enabled"`
```
schmichael, line 559: If we ever want to default this to `true` we'll have to change the type to a pointer (`*bool`) to differentiate unset vs explicitly-false. That shouldn't be a problem though since unlike objects sent over RPCs we never have to worry about an agent on the new version sending a `nil` to an agent still expecting a true or false.

nomad/state/state_store.go
```diff
@@ -359,6 +370,22 @@ func (s *StateStore) UpsertPlanResults(msgType structs.MessageType, index uint64
 	txn := s.db.WriteTxnMsgT(msgType, index)
 	defer txn.Abort()
 
+	// Mark nodes as ineligible.
+	now := time.Now().Unix()
```
schmichael, line 374: Since UpsertPlanResults is called by the FSM we cannot check wallclock time here as it will differ on each server creating skew in their statestores. Since it's only informational there shouldn't be a negative functional impact, but it's still preferable to keep our FSMs deterministic on every server.

schmichael (approved):

nomad/state/state_store.go
lgfa29: Fixed in https://github.com/hashicorp/nomad/pull/13421/commits/fb2e761cda83d847960df140a743537db0827d2d

command/agent/config.go
lgfa29: Good point, I may as well do the work now 😄 
https://github.com/hashicorp/nomad/pull/13421/commits/f5936f0830e1b5a490a8a4f5f7d39f138cd628a3

PR #13424, opened 2022-06-17
Throw an error if you try to save a Secure Variable with no items

### LINKED ISSUE #13393: Secure Variables: compact/remove empty Items[] on save
Currently, an added-but-empty row will append to the Items array, but we should clear these out on save.

image
image

philrenaud:
Resolves #13393

1. Compacts items (key/values) by filtering out those with no key/value
2. Throws an error message to the user if they attempt to save a Secure Variable with no items

### DIFF
--- a/ui/app/components/secure-variable-form.js
+++ b/ui/app/components/secure-variable-form.js
@@ -77,9 +77,19 @@ export default class SecureVariableFormComponent extends Component {
   async save(e) {
     e.preventDefault();
     try {
+      const nonEmptyItems = this.keyValues.filter(
+        (item) => item.key && item.value
+      );
+      if (!nonEmptyItems.length) {
+        throw new Error('Please provide at least one key/value pair.');
+      } else {
+        this.keyValues = nonEmptyItems;
+      }
+
       this.args.model.set('keyValues', this.keyValues);
       this.args.model.setAndTrimPath();
       await this.args.model.save();
+
       this.flashMessages.add({
         title: 'Secure Variable saved',
         message: `${this.args.model.path} successfully saved`,
@@ -88,6 +98,8 @@ export default class SecureVariableFormComponent extends Component {
         timeout: 5000,
         showProgress: true,
       });
+
+      this.router.transitionTo('variables.variable', this.args.model.path);
     } catch (error) {
       this.flashMessages.add({
         title: `Error saving ${this.args.model.path}`,
@@ -97,6 +109,5 @@ export default class SecureVariableFormComponent extends Component {
         sticky: true,
       });
     }
-    this.router.transitionTo('variables.variable', this.args.model.path);
   }
 }
--- a/ui/app/components/secure-variable-form.js
+++ b/ui/app/components/secure-variable-form.js
@@ -98,7 +98,6 @@ export default class SecureVariableFormComponent extends Component {
         timeout: 5000,
         showProgress: true,
       });
-
       this.router.transitionTo('variables.variable', this.args.model.path);
     } catch (error) {
       this.flashMessages.add({
--- a/ui/app/components/secure-variable-form.js
+++ b/ui/app/components/secure-variable-form.js
@@ -78,7 +78,7 @@ export default class SecureVariableFormComponent extends Component {
     e.preventDefault();
     try {
       const nonEmptyItems = this.keyValues.filter(
-        (item) => item.key && item.value
+        (item) => item.key.trim() && item.value
       );
       if (!nonEmptyItems.length) {
         throw new Error('Please provide at least one key/value pair.');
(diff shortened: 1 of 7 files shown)


ui/tests/integration/components/variable-paths-test.js
```diff
@@ -77,21 +77,21 @@ module('Integration | Component | variable-paths', function (hooks) {
       .dom('tbody tr:first-child td:first-child a')
       .hasAttribute(
         'href',
-        '/ui/variables/foo/bar/baz',
+        '/ui/variables/var/foo/bar/baz',
```
philrenaud, line 80: Side-effects: correcting for a previous change to add /var to disambiguate variable path URLs

ui/app/components/secure-variable-form.js
```diff
@@ -77,9 +77,19 @@ export default class SecureVariableFormComponent extends Component {
   async save(e) {
     e.preventDefault();
     try {
+      const nonEmptyItems = this.keyValues.filter(
+        (item) => item.key && item.value
```
ChaiWithJai, line 81: `nit`: This doesn't factor for whitespace. I set the key to be ` ` and the value to be ` `. And this is still valid. Maybe add `trim`?

ui/app/models/variable.js
```diff
@@ -67,6 +67,6 @@ export default class VariableModel extends Model {
    */
   setAndTrimPath() {
     this.set('path', trimPath([this.path]));
-    this.id = this.path;
+    this.set('id', this.get('path'));
```
ChaiWithJai, line 70: `nit` `non-blocking`:  you could have this be `setProperties`.

https://api.emberjs.com/ember/4.4/functions/@ember%2Fobject/setProperties

ui/mirage/config.js
```diff
@@ -848,13 +848,13 @@ export default function () {
 
   this.put('/var/:id', function (schema, request) {
     const { Path, Namespace, Items } = JSON.parse(request.requestBody);
-    server.create('variable', {
+    const newVar = server.create('variable', {
```
ChaiWithJai, line 851: `nit` `non-blocking`: you could just `return` here if you wanna play code golf.

ui/tests/acceptance/secure-variables-test.js
```diff
@@ -119,6 +119,25 @@ module('Acceptance | secure variables', function (hooks) {
     assert.notOk(fooLink, 'foo0 file is no longer present');
   });
 
+  test('it does not allow you to save if you lack Items', async function (assert) {
+    defaultScenario(server);
```
ChaiWithJai, line 123: `question`: Should this move into a `beforeEach` hook?

ui/tests/acceptance/secure-variables-test.js
```diff
@@ -119,6 +119,25 @@ module('Acceptance | secure variables', function (hooks) {
...
+    window.localStorage.nomadTokenSecret = server.db.tokens[0].secretId;
```
ChaiWithJai, line 124: `question`: Should we clean this up at the end of the test?

ChaiWithJai (approved):
LGTM some small comments. I do think that logic for actions in `secure-variable-form` is starting to get exceptionally complex.

ui/app/components/secure-variable-form.js
```diff
@@ -77,9 +77,19 @@ export default class SecureVariableFormComponent extends Component {
...
+      );
+      if (!nonEmptyItems.length) {
+        throw new Error('Please provide at least one key/value pair.');
```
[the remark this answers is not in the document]
philrenaud: no timeout on error message is deliberate; we want the user to acknowledge that a mistake was made.

ui/app/components/secure-variable-form.js
philrenaud: Hm, I think this is probably safe to do for the key, but I'm going to avoid it for value.

I think there's a world where someone wants a value of a whitespace character that would otherwise be trimmed for their secure variable, and we shouldn't necessarily nuke that. 

cc @angrycub  / @tgross in case there's a good reason to not trim keys as well.

ui/app/models/variable.js
philrenaud: I think that would cause an error; I want to set the `id` only after I set the `path` above, and editing both in a single `setProperties` would not give me a modified path.

ui/tests/acceptance/secure-variables-test.js
philrenaud: Hm not yet; I imagine there will be scenarios for this form where I don't want defaultScenario(), but set up some something custom / edge-casey.

ui/tests/acceptance/secure-variables-test.js
philrenaud: I think this probably isn't harmful / no need to clean it up yet. The precedent for this is found in another dozen+ test files throughout.

ui/app/models/variable.js
ChaiWithJai: Gotcha! We should make sure `setAndTrimPath` doesn't run on `edit` and only on `create` since `path` shouldn't be a mutable property.

PR #13427, opened 2022-06-19
drivers/docker: add context timeouts to docker client calls

### LINKED ISSUE #9503: review adding timeouts to Docker API calls
From https://github.com/hashicorp/nomad/pull/9502:

> This changeset is a minimal intervention to add the timeout to the Docker API
> call only when we have the `kill_timeout` set. Future work should examine
> whether we should be threading contexts through other `go-dockerclient` API
> calls.

danishprakash:
Closes #9503

Signed-off-by: danishprakash <grafitykoncept@gmail.com>

### DIFF
--- a/drivers/docker/driver.go
+++ b/drivers/docker/driver.go
@@ -428,6 +428,10 @@ func (d *Driver) createContainer(client createContainerClient, config docker.Cre
 	image string) (*docker.Container, error) {
 	// Create a container
 	attempted := 0
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute) // TODO: update arbitrary timeout
+	defer cancel()
+	config.Context = ctx
+
 CREATE:
 	container, createErr := client.CreateContainer(config)
 	if createErr == nil {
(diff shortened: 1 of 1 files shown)


shoenig:
@danishprakash did you intend for this PR to still be marked as a Draft? If not, do you mind rebasing with main and marking it Ready for Reveiw?

danishprakash:
@shoenig yeah, I marked it as a draft since I wanted some inputs from the team and I hadn't gone through the whole codebase with this change. But I think you can go ahead and review it if you can and I can then address the comments.

shoenig:
Sorry for the super slow responses @danishprakash :grimacing: 

One key change I think we should make is making the timeout value configurable in the docker plugin configuration.

Similar to how we already have an option for `pull_activity_timeout`, let's add something like `endpoint_timeout` which is general for all other docker endpoint interactions. 

So we'll want add that, then apply it to all the places we can, and update the docs.

tgross:
Hi @danishprakash! We'd love to have this PR if you have time to revive it. It's the sort of thing that's been hard to make bubble up to the top of our priority list.

tgross:
@danishprakash we've decided to go ahead with migrating to the official SDK client over here: https://github.com/hashicorp/nomad/pull/23966. The work of making sure all the contexts used in the official SDK have reasonable timeout behaviors will still need to get done, but you'd basically be starting from scratch. So I'm going to close this PR out. If you're still interested in helping out with this once 1.9.0 goes out, we'd be happy to review a new PR.

PR #13429, opened 2022-06-20
Implement Check and Set for Delete and Upsert

angrycub:
Added check and set semantics for secure variables. Ended up implementing them in the HTTP handler because of the way that the RPC functions return back flattened errors. Plus wouldn't want the secrets being leaked out in an error logging path. So do a little double duty in the handler to fetch the current value.  In cases that there is a check and set error, we will return 409 and the Items that are currently stored in the secure variable to perhaps make the UI work a little more spiffy


nomad/secure_variables_endpoint.go
```diff
@@ -468,3 +477,32 @@ func (sv *SecureVariables) authValidatePrefix(alloc *structs.Allocation, taskNam
...
+	var reply = new(structs.SecureVariablesReadResponse)
+	var rArgs = &structs.SecureVariablesReadRequest{
+		Path: path,
+		QueryOptions: structs.QueryOptions{
+			Region:    wo.Region,
+			AuthToken: wo.AuthToken,
+			Namespace: wo.Namespace,
+		},
+	}
+	sv.Read(rArgs, reply)
```
tgross, line 495: Sending this back thru the top-level `Read` method means we end up having to recheck all the ACLs, set up a blocking query, and emit metrics for the `Read` RPC. We should have this read from the state store snapshot directly.

command/agent/secure_variable_endpoint.go
```diff
@@ -78,20 +80,30 @@ func (s *HTTPServer) secureVariableUpsert(resp http.ResponseWriter, req *http.Re
...
 	var out structs.SecureVariablesUpsertResponse
 	if err := s.agent.RPC(structs.SecureVariablesUpsertRPCMethod, &args, &out); err != nil {
+		if strings.Contains(err.Error(), "check-and-set conflict") {
+			q, _ := s.secureVariableQuery(resp, req, path)
+			sv := q.(*structs.SecureVariableDecrypted)
+			out.Conflict = make([]*structs.SecureVariableDecrypted, 1)
+			out.Conflict[0] = sv
+			resp.WriteHeader(http.StatusConflict)
+			return sv, nil
+		}
```
tgross, line 103: `*structs.SecureVariableDecrypted` is available to the RPC. I think we can populate this `Conflict` field in the RPC handler and then let `len(out.Conflict)` determine whether we have to set the header. That'll let us avoid having to parse the error string or re-parse the original request.

nomad/structs/secure_variables.go
```diff
@@ -167,16 +167,22 @@ type SecureVariablesQuota struct {
...
+	svur.CheckIndex = &ci
+}
+
 type SecureVariablesEncryptedUpsertRequest struct {
 	Data []*SecureVariableEncrypted
 	WriteRequest
 }
 
 type SecureVariablesUpsertResponse struct {
+	Conflict []*SecureVariableDecrypted
```
tgross, line 185: Maybe pluralize this because it's a slice? Or does it really need to be a slice to begin with?

```suggestion
	Conflicts []*SecureVariableDecrypted
```

tgross (commented):
Looking pretty good @angrycub!

nomad/secure_variables_endpoint.go
```diff
@@ -468,3 +477,32 @@ func (sv *SecureVariables) authValidatePrefix(alloc *structs.Allocation, taskNam
 	}
 	return nil
 }
+
+func (sv *SecureVariables) handleCheckAndSet(path string, idx *uint64, wo structs.WriteRequest) error {
```
tgross, line 481: This has lifted the implementation from `Job.Register`'s `-check-index`, which wasn't a bad approach b/c jobs changes infrequently and the `-check-index` is really about making sure a `Job.Register` has the same state as a previous `Job.Plan`. But for secure variables, users are going to want to use CAS as a primitive for locks and other concurrent access approaches, so we should try to be more strict with concurrent updates. We should push this logic all the way down to the state store, similar to how it was done in Consul: [ex. `kvs.go#L363-L388`](https://github.com/hashicorp/consul/blob/main/agent/consul/state/kvs.go#L363-L388)

tgross (changes requested):

command/agent/secure_variable_endpoint_test.go
```diff
@@ -268,15 +278,76 @@ func TestHTTP_SecureVariables(t *testing.T) {
...
-			// Check the variable was created
-			out, err := rpcReadSV(s, sv1.Namespace, sv1.Path)
+			{
+				out, err := rpcReadSV(s, sv.Namespace, sv.Path)
+				require.NoError(t, err)
+				require.NotNil(t, out)
+				require.NotEqual(t, svU.SecureVariableMetadata, out.SecureVariableMetadata)
+
+				svU.CreateIndex, svU.ModifyIndex = out.CreateIndex, out.ModifyIndex
+				svU.CreateTime, svU.ModifyTime = out.CreateTime, out.ModifyTime
```
tgross, line 288: Maybe there's something weird going on with the scope here that I'm missing, but what do these two lines get us? (Same for the `update-cas` test below.)

nomad/secure_variables_endpoint.go
```diff
@@ -464,3 +497,53 @@ func (sv *SecureVariables) authValidatePrefix(claims *structs.IdentityClaims, ns
...
+func (s *SecureVariables) validateCASUpdate(cidx uint64, sv *structs.SecureVariableDecrypted, conflict **structs.SecureVariableDecrypted) error {
+	return s.validateCAS(cidx, sv.Namespace, sv.Path, conflict)
+}
+
+func (s *SecureVariables) validateCASDelete(cidx uint64, namespace, path string, conflict **structs.SecureVariableDecrypted) error {
+	return s.validateCAS(cidx, namespace, path, conflict)
+}
+
+func (s *SecureVariables) validateCAS(cidx uint64, namespace, path string, conflictOut **structs.SecureVariableDecrypted) error {
+	casConflict := errors.New("conflict")
```
tgross, line 510: For later: maybe we can move this error to `nomad/structs/errors.go`. Then we can use nicer constructs like `errors.Is` within the RPC handler instead of `if err.Error() == "conflict"` (which we can't avoid with the HTTP handler unfortunately).

command/agent/secure_variable_endpoint.go
```diff
@@ -75,23 +77,49 @@ func (s *HTTPServer) secureVariableUpsert(resp http.ResponseWriter, req *http.Re
...
 	var out structs.SecureVariablesUpsertResponse
 	if err := s.agent.RPC(structs.SecureVariablesUpsertRPCMethod, &args, &out); err != nil {
+
+		// This handles the cases where there is an error in the CAS checking
+		// function that renders it unable to return the conflicting variable
+		// so it returns a text error. We can at least consider these unknown
+		// moments to be CAS violations
+		if strings.Contains(err.Error(), "cas error:") {
+			resp.WriteHeader(http.StatusConflict)
+		}
```
tgross, line 105: :+1:  I like the way you've handled a mildly annoying aspect of the state store API here and how you've provided a rationale for the behavior in the comment.

tgross (approved):
LGTM!

command/agent/secure_variable_endpoint_test.go
```diff
@@ -192,7 +194,7 @@ func TestHTTP_SecureVariables(t *testing.T) {
 			require.NoError(t, err)
 			respW := httptest.NewRecorder()
 			obj, err := s.Server.SecureVariableSpecificRequest(respW, req)
-			require.EqualError(t, err, "Secure variable missing required Items object.")
+			require.EqualError(t, err, "secure variable missing required Items object")
```
tgross, line 197: It looks like we missed a few of these: https://app.circleci.com/pipelines/github/hashicorp/nomad/26598/workflows/37b8c8e2-7ffb-4484-a67a-b68082cecb18/jobs/290337

command/agent/secure_variable_endpoint_test.go
angrycub: They follow-on after a NotEqual check above that ensures that I didn't mutate `svU` in the update, then updating the time and indexes lets me do a simple equality check to make sure the remainder of the metadata is correct.  Probably overkill at the moment, but I'm imagining a time when metadata has more in it than namespace, path.

PR #13436, opened 2022-06-20
Append an empty item when editing an existing Secure Variable

### LINKED ISSUE #13394: Secure Variables: When rendering /edit, append a blank Items[] key/value so the other last row is deletable
Currently, loads like this, with the last row not deletable: 
image

Instead, let's make the initial state like this:
image

philrenaud:
Resolves #13394

Appends an empty row to the end of your Items array when you modify an existing Secure Variable.

This ensures that all previous rows are deletable, and with https://github.com/hashicorp/nomad/pull/13424, will compact/remove empty items on save.

### DIFF
--- a/ui/app/components/secure-variable-form.hbs
+++ b/ui/app/components/secure-variable-form.hbs
@@ -1,4 +1,6 @@
-<form class="new-secure-variables" autocomplete="off" {{on "submit" this.save}}>
+<form class="new-secure-variables" autocomplete="off" {{on "submit" this.save}}
+  {{did-insert this.appendItemIfEditing}}
+>
   {{!-- TODO: {{if this.parseError 'is-danger'}} on inputs --}}
   <div>
     <label>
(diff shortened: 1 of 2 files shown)


ChaiWithJai (approved):
I'm noticing some lurking errors and warning messages as a result of the `save` logic. I have a feeling that `setAndTrimPath` shouldn't be running when the form is in `edit` mode. We disable the `input` field on the client, however, we still run `setAndTrimPath` which resets the `id` of the model we're currently working on.

This can lead to some nasty side-effects and I get this warning:

`WARNING: The 'id' for a RecordIdentifier should not be updated once it has been set. Attempted to set id for 'variable:blue/berkshire/lime (@ember-data:lid-variable-blue/berkshire/lime)' to 'null'.`

Now that we're adding a callback that's fired when the form element is appended onto the DOM, we should maybe consider explicitly modeling state.

PR #13439, opened 2022-06-20
Secure Variables: JSON view

philrenaud:
image

### DIFF
--- a/ui/app/controllers/variables/variable/index.js
+++ b/ui/app/controllers/variables/variable/index.js
@@ -45,4 +45,14 @@ export default class VariablesVariableIndexController extends Controller {
   onDismissError() {
     this.error = null;
   }
+
+  //#region Code View
+  @tracked
+  isCodeView = false;
+
+  toggleCodeView() {
+    this.isCodeView = !this.isCodeView;
+  }
+
+  //#endregion Code View
 }
--- a/ui/app/helpers/stringify-object.js
+++ b/ui/app/helpers/stringify-object.js
@@ -1,0 +1,9 @@
+/**
+ * Changes a JSON object into a string
+ */
+
+import { helper } from '@ember/component/helper';
+
+export default helper(function stringifyObject(positional /*, named*/) {
+  return JSON.stringify(positional[0], null, 2);
+});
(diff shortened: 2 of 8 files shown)


ui/app/controllers/variables/variable/index.js
```diff
@@ -6,6 +6,12 @@ import { inject as service } from '@ember/service';
 import { tracked } from '@glimmer/tracking';
 
 export default class VariablesVariableIndexController extends Controller {
+  queryParams = [
+    {
+      view: 'view',
+    },
+  ];
```
ChaiWithJai, line 13: `nit`: This doesn't need an alias. When we have a query parameter `namespace` we need to alias that because `namespace` is a reserved keyword in Ember.

you good with:

```
queryParams = ['view'];
@tracked view = 'table';
```

ui/app/helpers/stringify-object.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Changes a JSON object into a string
+ */
+
+import { helper } from '@ember/component/helper';
+
+export default helper(function stringifyObject(positional /*, named*/) {
+  return JSON.stringify(positional[0], null, 2);
```
ChaiWithJai, line 8: `question`:  Should this get some tests around it to explain the whitespace argument?
`nit`: Should we opt for named arguments? Or refactor `positional` to be something like:

```
function stringifyObject([obj]) {
  return JSON.stringify(obj, null, 2)
}
```

Alternatively, we could write this API

```
function stringifyObject([obj], { replacer = null, whiteSpace = 2 }) {
  return JSON.stringify(obj, replacer, whiteSpace);
}
```

ui/tests/unit/models/variable-test.js
```diff
@@ -30,4 +30,22 @@ module('Unit | Model | variable', function (hooks) {
     });
     assert.equal(model.keyValues.length, 1);
   });
+
+  test('it correctly moves between keyValues and itmes', function (assert) {
```
ChaiWithJai, line 34: `nit`:  typo "itmes" should "items"

ChaiWithJai (approved):
LGTM, but I think the helper API could be cleaned up and tested.

https://guides.emberjs.com/release/components/helper-functions/

PR #13440, opened 2022-06-20
core job for secure variables re-key

tgross:
When the `Full` flag is passed for key rotation, we kick off a core
job to decrypt and re-encrypt all the secure variables so that they
use the new key.

### DIFF
--- a/nomad/core_sched.go
+++ b/nomad/core_sched.go
@@ -827,7 +827,7 @@ func (c *CoreScheduler) rootKeyRotateOrGC(eval *structs.Evaluation) error {
 			break
 		}
 		keyMeta := raw.(*structs.RootKeyMeta)
-		if keyMeta.Active {
+		if keyMeta.Active() {
 			continue // never GC the active key
 		}
 		if keyMeta.CreateIndex > oldThreshold {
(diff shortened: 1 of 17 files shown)


nomad/core_sched.go
```diff
@@ -892,6 +895,94 @@ func (c *CoreScheduler) rootKeyRotation(eval *structs.Evaluation) (bool, error)
...
+		}
+		args := &structs.SecureVariablesUpsertRequest{
+			Data: variables,
+			WriteRequest: structs.WriteRequest{
+				Region:    c.srv.config.Region,
+				AuthToken: eval.LeaderACL, // TODO: will this work?
+			},
+		}
+		reply := &structs.SecureVariablesUpsertResponse{}
+		return c.srv.RPC("SecureVariables.Upsert", args, reply)
```
angrycub, line 950: Using Upsert will change the ModifyIndex while rekeying the variable. This will cause CAS conflicts on items that haven't haven't been modified from the user's perspective. How do we feel about that?

angrycub (dismissed):
I had one question, but if it's not a concern then I'm good to go on this

nomad/core_sched.go
tgross: I _think_ that's actually what we want. If this job is running you probably _don't_ want to be able to accidentally overwrite anything until it's done with that batch. I don't love that it ends up resolving any open blocking queries though.

angrycub (approved):
I think we could workshop the state names a little bit, but definitely not going to block the merge based on that.

Issue #13310, opened 2022-06-09
Secure Variables: Acceptance tests for Create and Edit flows


philrenaud:
Mimic the "it correctly traverses to and deletes a variable" flow in acceptance/secure-variables-test.js

ChaiWithJai:
### Create Flow

Happy Path (create new variable)

- Click Create
- Navigates to `/variables/new`
- Fill the form out
- Save the form
- Navigate to `/variables/$VARIABLE_NAME`

Happy Path (start creating new variable, and quit in the middle of the process)

- Click Create
- Navigates to `/variables/new`
- Fill the form out
- Navigate away from page (assert no variable was created)
- Click Create, again (assert draft state was NOT saved)

Sad Path (no permissions)

- Click Create (without permissions)
- Error message
- Navigate to `/variables/new`
- Route is guarded and sent back to `/variables` route with toast notification

Sad Path (invalid path name)
Sad Path (invalid key/value pair)
Sad Path (network error)
Sad Path (slow internet connection versus timed toast notification)
Sad Path (variable already exists)

ChaiWithJai:
### Edit Flow

Happy Path (edit variable)

- Click variable
- Click edit
- Save
- Navigate to `variables/VARIABLE_ID`

Happy Path (quit while editing)
Sad Path (lacking permissions)
Sad Path (resource was deleted by someone else)
Sad Path (invalid path name)
Sad Path (invalid key/value pair)
Sad Path (network error)
Sad Path (duplicate key name)

PR #13447, opened 2022-06-21
ui:  create variable permission logic

ChaiWithJai:
This PR creates logic for the `canCreate` permission, updates templates to properly check for this ability.

Closes [#13435](https://github.com/hashicorp/nomad/issues/13435).

Side Effect:  Update `Token` factory to enable `blue/berkshire` path to have `create` ability by default.

Enabled:
image

Disabled:
image

### DIFF
--- a/ui/app/abilities/variable.js
+++ b/ui/app/abilities/variable.js
@@ -1,10 +1,16 @@
 import { computed, get } from '@ember/object';
 import { or } from '@ember/object/computed';
-import { inject as service } from '@ember/service';
 import AbstractAbility from './abstract';
 
 export default class Variable extends AbstractAbility {
-  @service router;
+  // Pass in a namespace to `can` or `cannot` calls to override
+  // https://github.com/minutebase/ember-can#additional-attributes
+  path = '*';
+
+  get _path() {
+    if (!this.path) return '*';
+    return this.path;
+  }
 
   @or(
     'bypassAuthorization',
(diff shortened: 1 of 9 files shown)


ui/tests/unit/abilities/variable-test.js
```diff
@@ -15,87 +15,151 @@ module('Unit | Ability | variable', function (hooks) {
...
+        selfToken: { type: 'client' },
+      });
+
+      this.owner.register('service:token', mockToken);
+
+      assert.ok(this.ability.canCreate);
     });
 
-    this.owner.register('service:token', mockToken);
+    test('it permits creating variables when token has SecureVariables with create capabilities in its rules', function (assert) {
```
tgross, line 137: Something to consider on these tests is that we don't test many of the "it doesn't permit" cases. The one that jumps out at me is that we don't test that having a capability in one namespace doesn't give access to that same capability in another namespace.

ChaiWithJai: Yup. Namespaces are a behemoth of an issue and one of the most important testing stories for the UI. This is ticketed in [13310](https://github.com/hashicorp/nomad/issues/13310#issuecomment-1159178019) which this PR is a prerequisite to.

We should totally collaborate on creating a list of plain English specifications around namespaces before we get too deep into implementation. Which is what I had in mind [yesterday](https://hashicorp.slack.com/archives/C033DJD6EJJ/p1655743562030679?thread_ts=1655742096.654959&cid=C033DJD6EJJ).

Great job staying active on these PRs by the way!! Looking forward to collaborating with you more.

ui/app/abilities/variable.js
```diff
@@ -24,10 +27,14 @@ export default class extends AbstractAbility {
     });
   }
 
-  @computed('rulesForNamespace.@each.capabilities') // TODO:  edit computed property to be SecureVariables.Path "DYNAMIC PATH"
+  @computed(
+    'rulesForNamespace.@each.capabilities',
+    'router.currentRoute.params.absolutePath'
```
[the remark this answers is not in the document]
ChaiWithJai: Wanna write a code sample?

philrenaud (approved):

PR #13453, opened 2022-06-21
implement quota tracking for secure variablees

tgross:
We need to track per-namespace storage usage for secure variables even
in Nomad OSS so that a cluster can be seamlessly upgraded from OSS to
ENT without having to re-calculate quota usage.

Provide a hook in the upsert RPC for enforcement of quotas in
ENT. This will be a no-op in Nomad OSS.


angrycub (approved):
lgtm

PR #13460, opened 2022-06-22
docs: clarify total_escaped is just an optimization

schmichael:
I'm not attached to this PR, so we can throw it out if you think trying to cram in more details about a subtle metrics only confuses things further.

Also I didn't see how best to format long content in mdx tables, so we can throw it out based purely on ugliness of source. :sweat_smile: A `Metrics Codex` where we could actually write more than one sentence for a metric would be nice as I feel we end up relying on engineer's memory and code spelunking to get any nuanced meaning for metrics.


tgross (approved):
LGTM

PR #13461, opened 2022-06-22
Edit Secure Variables as JSON

philrenaud:
- Allows users to create and modify Secure Variables in JSON mode using Codemirror
- Instead of using the existing (and now deprecated) IvyCodemirror, adds a `{{code-mirror}}` modifier with more robust error states (more in line with Vault)
- Checks for JSON errors / lints on update, and only updates saveable values when there are no errors present.

### TODO
- [x] Check for non-standard SV shape before saving. For example, a user can wrap their object in an array (`[]`) and it's still valid JSON, but we should warn them / prevent save.
- [x] Clean up code-mirror.js a bit; remove ruby mode, etc.
- [x] Find out why existing Ivy plugins are not inheriting color scheme
- [x] JSON on /new

image


ui/app/components/secure-variable-form.js
```diff
@@ -75,10 +111,16 @@ export default class SecureVariableFormComponent extends Component {
 
   @action
   async save(e) {
-    e.preventDefault();
+    if (e.type === 'submit') {
```
ChaiWithJai, line 114: `thought`:  This does highlight that this function is now being used to: autosave, save on click and update on click. The function API and logic might benefit from being explicit about this because a refactoring for one use case to this function could break the other 2 down the line.

ui/app/components/secure-variable-form.js
```diff
@@ -75,10 +111,16 @@ export default class SecureVariableFormComponent extends Component {
...
+    }
+    // TODO: temp, hacky way to force translation to tabular keyValues
+    if (this.view === 'json') {
+      this.translateAndValidateItems('table');
+    }
     try {
-      const nonEmptyItems = this.keyValues.filter(
-        (item) => item.key.trim() && item.value
+      const nonEmptyItems = A(
+        this.keyValues.filter((item) => item.key.trim() && item.value)
```
ChaiWithJai, line 123: `thought` `non-blocking`:  the `filter` method and associated length check in the conditional aren't asynchronous callbacks and technically don't belong in the `try/catch` block. I can totally understand that we're throwing an error that should be swallowed by the `catch` and then show a flash method.

But, since that check is synchronous in nature and we're writing a form, shouldn't the submit button just be disabled with validation text? IMO, flash messages are here to help us display information after an asynchronous call is made and communicate status of the async request.

ui/app/components/secure-variable-form.js
```diff
@@ -110,13 +152,75 @@ export default class SecureVariableFormComponent extends Component {
...
+        )
+      );
+    }
+
+    // Reset any error state, since the errorring json will not persist
+    set(this, 'JSONError', null);
+  }
+
+  get stringifiedItems() {
+    return JSON.stringify(this.args.model.items, null, 2);
```
ChaiWithJai, line 204: `question`:  Didn't you write a helper for this? We could move this to the template if that's the case.

ui/app/components/secure-variable-form.hbs
```diff
@@ -1,7 +1,6 @@
-<form class="new-secure-variables" autocomplete="off" {{on "submit" this.save}}
-  {{did-insert this.appendItemIfEditing}}
->
-  {{!-- TODO: {{if this.parseError 'is-danger'}} on inputs --}}
+{{did-update this.onViewChange @view}}
+{{did-insert this.establishKeyValues}}
```
ChaiWithJai, line 2: `praise`:  good use of render modifiers.

ui/app/modifiers/code-mirror.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  @action
+  _onChange(editor) {
+    this.args.named.onUpdate(editor.getValue(), this._editor);
+  }
+
+  _setup() {
+    if (!this.element) {
+      throw new Error('CodeMirror modifier has no element');
```
ChaiWithJai, line 37: `question`: Who handles this error? Does the package?

ui/app/components/secure-variable-form.js
```diff
@@ -110,13 +152,75 @@ export default class SecureVariableFormComponent extends Component {
...
   }
 
+  //#region JSON Editing
+
+  view = this.args.view;
+  // Prevent duplicate onUpdate events when @view is set to its already-existing value,
+  // which happens because parent's queryParams and toggle button both resolve independently.
+  @action onViewChange([view]) {
+    if (view !== this.view) {
+      set(this, 'view', view);
```
ChaiWithJai, line 162: `nit` `non-blocking`: `view` is an enum that can be `table` or `json`, right? Why are we calling `set` to update its value? Can't we just mark `view` as a `tracked` property and re-assign via `this.view = view`;

I immediately (and incorrectly) inferred that view was an object, not a string when I read this.

https://guides.emberjs.com/release/in-depth-topics/autotracking-in-depth/#toc_plain-old-javascript-objects-pojos

ChaiWithJai: `blocking`:  Should updating this value here, update the query parameter in the controller? This component is being passed this prop from the parent template which has a tracked `view` property as a query parameter. This `view` argument will reactively update to the controller, if we set a change here, shouldn't the controller update?

ChaiWithJai (approved):
There's some potential bugs I see here that I've outlined. Also, because we're following atomic PRs, I haven't been able to give my input on API design, but I do make advocate for refactoring the `save` method.

I totally understand that adding tests could bloat this PR past 1,000 lines of code and you already know this work inside and out, but seeing tests would've helped me grok this more quickly and provide more meaningful feedback.

ui/app/modifiers/code-mirror.js
```diff
@@ -0,0 +1,60 @@
...
+import Modifier from 'ember-modifier';
+
+import 'codemirror/addon/edit/matchbrackets';
+import 'codemirror/addon/selection/active-line';
+import 'codemirror/addon/lint/lint.js';
+import 'codemirror/addon/lint/json-lint.js';
+// right now we only use the ruby and javascript, if you use another mode you'll need to import it.
+// https://codemirror.net/mode/
+import 'codemirror/mode/ruby/ruby';
+import 'codemirror/mode/javascript/javascript';
```
tgross, line 13: Given that this isn't actually Javascript either, but JSON, is there a JSON-specific mode so that we don't carry along anything needed for Javascript editing too? (Which I imagine is non-trivial? Not sure how this works under the hood)

philrenaud: I gave this a shot and it turns out that codemirror's javascript mode/parser is actually really good at color-highlighting HCL!

image

it's what IvyCodemirror was using behind the scenes. So while the file is non-trivially large, the language mode file should still be included.

Dropping Ruby, though.

ui/app/components/secure-variable-form.js
philrenaud: I think that is a good pattern so long as this was just a form; but since we're also including JSON editing capabilities, I worry about having a submit button that flickers in/out of disabled state as a user types. Let me think about this one a bit more; if I can make it appear reasonably to the end user, I'll make this a `formHasNoNonEmptyItems` property or something similar.

ChaiWithJai: **Chants uncontrollably** "Statecharts, Statecharts, Statecharts!"

mikenomitch:
Non-blocking issue when editing json: If I have two of the same key and then press delete on the comma at the end, the whole like will disappear.

This happened to me when I was copy/pasting to eventually change the dupe key, but I wanted to fix the json formatting before changing the key, but then the line disappeared.

https://user-images.githubusercontent.com/2732204/176784696-84005d39-f057-4e56-b3eb-feafda219bca.mov

mikenomitch:
Also saving dupe keys with different values just made one disappear in non-JSON editing. Feels like this should be an error that blocks save rather than anything else.

philrenaud:
@mikenomitch duplicate keys and snap-json-to-top / delete block errors are both resolved.

The  duplicate error currently shows up as a warning to the user, but doesn't prohibit save. I think this is a good first step.

PR #13464, opened 2022-06-22
additional ACL Policy tests

tgross:
This changeset includes some additional unit tests for secure
variables ACL policies, so that we have explicit coverage of edge
cases we're discussing with the UI folks.


ChaiWithJai (approved):
Great job!

PR #13467, opened 2022-06-23
core: remove support for raft protocol version 2

shoenig:
This PR checks server config for raft_protocol, which must now
be set to 3 or unset (0). When unset, version 3 is used as the
default.


tgross (approved):
LGTM

PR #13472, opened 2022-06-23
query for leader in `operator debug` command

tgross:
The `operator debug` command doesn't output the leader anywhere in the
output, which adds extra burden to offline debugging (away from an
ongoing incident where you can simply check manually). Query the
`/v1/status/leader` API but degrade gracefully.


jrasell (approved):
Just needs a changelog entry, otherwise LGTM!

PR #13474, opened 2022-06-23
ui:  handle path matching

ChaiWithJai:
This PR is one of two PRs that will unblock  [#13447](https://github.com/hashicorp/nomad/pull/13447).

This PR specifically handle path matching when there is no exact match. However, this PR is not in use because there will be a follow-up PR for integrating namespaces as part of the path matching process.

### DIFF
--- a/ui/app/abilities/variable.js
+++ b/ui/app/abilities/variable.js
@@ -41,21 +41,26 @@ export default class Variable extends AbstractAbility {
 
   @computed('rulesForNamespace.@each.capabilities', 'path')
   get policiesSupportVariableCreation() {
+    const matchingPath = this._nearestMatchingPath(this.path);
     return this.rulesForNamespace.some((rules) => {
-      const keyName = `SecureVariables.Path "${this.path}".Capabilities`;
+      const keyName = `SecureVariables.Path "${matchingPath}".Capabilities`;
       const capabilities = get(rules, keyName) || [];
       return capabilities.includes('create');
     });
   }
 
-  @computed('token.selfTokenPolicies.[]')
+  @computed('token.selfTokenPolicies.[]', '_namespace')
   get allPaths() {
-    return get(this, 'token.selfTokenPolicies')
+    return (get(this, 'token.selfTokenPolicies') || [])
       .toArray()
       .reduce((paths, policy) => {
-        const [variables] = get(policy, 'rulesJSON.Namespaces').mapBy(
-          'SecureVariables'
+        const matchingNamespace = this._findMatchingNamespace(
+          get(policy, 'rulesJSON.Namespaces') || [],
+          this._namespace
         );
+        const variables = (get(policy, 'rulesJSON.Namespaces') || []).find(
+          (namespace) => namespace.Name === matchingNamespace
+        )?.SecureVariables;
         paths = { ...paths, ...variables };
         return paths;
       }, {});
--- a/ui/app/abilities/variable.js
+++ b/ui/app/abilities/variable.js
@@ -41,4 +41,24 @@ export default class Variable extends AbstractAbility {
       return capabilities.includes('create');
     });
   }
+
+  _nearestMatchingPath(path) {
+    const formattedPathKey = `Path "${path}"`;
+    const pathNames = Object.keys(this.allPaths);
+
+    if (pathNames.includes(formattedPathKey)) return path;
+  }
+
+  @computed('token.selfTokenPolicies.[]')
+  get allPaths() {
+    return get(this, 'token.selfTokenPolicies')
+      .toArray()
+      .reduce((paths, policy) => {
+        const [variables] = get(policy, 'rulesJSON.Namespaces').mapBy(
+          'SecureVariables'
+        );
+        paths = { ...paths, ...variables };
+        return paths;
+      }, {});
+  }
 }
--- a/ui/app/abilities/variable.js
+++ b/ui/app/abilities/variable.js
@@ -134,4 +134,40 @@ export default class Variable extends AbstractAbility {
       return matchingPath;
     });
   }
+
+  _doesMatchPattern(pattern, path) {
+    const parts = pattern?.split(WILDCARD_GLOB);
+    const hasLeadingGlob = pattern?.startsWith(WILDCARD_GLOB);
+    const hasTrailingGlob = pattern?.endsWith(WILDCARD_GLOB);
+    const lastPartOfPattern = parts[parts.length - 1];
+    const isPatternWithoutGlob = parts.length === 1 && !hasLeadingGlob;
+
+    if (!pattern || !path || isPatternWithoutGlob) {
+      return pattern === path;
+    }
+
+    if (pattern === WILDCARD_GLOB) {
+      return true;
+    }
+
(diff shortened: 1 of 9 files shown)


ui/tests/unit/abilities/variable-test.js
```diff
@@ -162,4 +162,228 @@ module('Unit | Ability | variable', function (hooks) {
...
+      const nearestMatchingPath = this.ability._nearestMatchingPath(path);
+
+      assert.equal(
+        nearestMatchingPath,
+        'foo',
+        'It should return the exact path match.'
+      );
+    });
+
+    test('returns capabilities for the nearest ancestor if no exact match', function (assert) {
```
tgross, line 202: Referring to this as "ancestor" is harmless in terms of the code but might be confusing for future readers of this code, as we're not really checking ancestors as though they were filesystem paths but doing closest string matching.

ui/app/abilities/variable.js
```diff
@@ -41,4 +47,86 @@ export default class Variable extends AbstractAbility {
...
+      }
+      const count = matchingPath.match(WILDCARD_PATTERN_REGEX)?.length || 0;
+      if (currentPath.match(WILDCARD_PATTERN_REGEX)?.length > count) {
+        matchingPath = currentPath;
+      } else if (currentPath.match(WILDCARD_PATTERN_REGEX)?.length === count) {
+        // Chose suffix over prefix
+        if (currentPath.endsWith(WILDCARD_GLOB)) {
+          matchingPath = currentPath;
+        }
+      }
```
tgross, line 127: This algorithm works for the tests that we have but I'm worried that not using the same algorithm we have in [`findAllMatchingWildcards`](https://github.com/hashicorp/nomad/blob/6fb29066d9e24b5921061a22f99303cf07d4523c/acl/acl.go#L421-L445) is going to run into subtle discrepancies between the behavior of the API and the UI. 

The algorithm described there is:
* candidates are sorted
* a path is a candidate if it glob matches
* each candidate has a difference calculated as `len(path) - len(pattern) + (number of globs in pattern)`
* the smallest difference is used, using the first candidate

Trying to implement a separate algorithm seems like a recipe for trying to encode behaviors that are emergent from the original algorithm (like the prefix before suffix thing, which is really as a result of sorting).

ChaiWithJai: Alright, I can refactor this. I was trying to follow your and @philrenaud's guidance re:  "Allow by default" behavior. This was the loosest algorithm that I could design to patch that pattern.

But, I could totally use some help turning the 4 parts of the algorithm into specifications that I can test against. Specifically:

- what the sorting pattern is (e.g. ascending by first character of path)
- how to compute `len(pattern`
- number of globs in pattern??? as in `foo/*/bar/*/baz` is a possibility?

ui/tests/unit/abilities/variable-test.js
ChaiWithJai: Ah! Can we double-click on this? My mental model comes from web development and we have elements that appear in the browser. So if we have a table, it might look like this?

```
<table>
  <th>
    <td>Words</td>
  </th>
</table>
```

In this case, if we're using `td` as a point of reference, we'd call `th` and `table` its ancestors. My hallucination was that file system paths are also hierarchical. Like a folder that has all of your jobs, and then you click on a job in your folder and that is made up of task groups, and that folder is made up of tasks.

```
Jobs --> `JobId@Namespace` --> Task Groups --> `TaskId`
```

But, you're saying we're not really checking the ancestor here, because we could match on `*/someName` as a pathway. So this is some type of "fuzzyStringMatchingAlgorithm" and I should be saying:

```
test('returns closest fuzzy matching string')
```

Or am I still wrong?

ui/app/abilities/variable.js
tgross: * The sorting comes out of the [recursive walk](https://github.com/hashicorp/go-immutable-radix/blob/49d1d02c49a783de548d1ba8ae8fde466a20b9e6/node.go#L300-L315) of the [immutable radix tree](https://github.com/hashicorp/go-immutable-radix) data structure, which gets you a lexicographical sort order. See https://play.golang.com/p/Rtzyi8OfnaT for an example (press **Run** on that page).
* `len(pattern)` is the length in bytes of the policy's path
* number of globs in the pattern is yes just a count of `*` characters. More than one is a possibility (although incredibly ill-advised on the part of the operator 😀 )

philrenaud (approved):
Impressive and looks pretty exhaustive to the edge cases! Per Tim's comment, your best bet is to directly mimic what's happening with the backend implementation here.

We'll need to be vigilant about updating this in the future, should the backend glob matching ever change, too. I'm usually wary about having redundant checks like this, but it seems fine here.

ui/app/abilities/variable.js
ChaiWithJai: OK, I think writing this algorithm is going to be pretty challenging.

ChaiWithJai: Hmmm... so do you think I should just translate this into JavaScript to get the sorting done?

```
func recursiveWalk(n *Node, fn WalkFn) bool {
	// Visit the leaf values if any
	if n.leaf != nil && fn(n.leaf.key, n.leaf.val) {
		return true
	}

	// Recurse on the children
	for _, e := range n.edges {
		if recursiveWalk(e.node, fn) {
			return true
		}
	}
	return false
}
```

philrenaud (changes requested):
Running into a few issues here in testing, the most notable of which is a JSON error when trying to load the variables index route with the following ACL:


```
namespace "default" {
  policy = "write"
  capabilities = ["alloc-node-exec", "list-jobs"]
  secure_variables {

    # full access to secrets in all project paths
    path "project/*" {
    }

    # full access to secrets in all project paths
    path "*" {
      capabilities = ["list"]
    }

    # read/list access within a "system" path belonging to administrators
    path "system/*" {
      capabilities = ["read", "list"]
    }
  }

}
```

I think there's an issue with the abstract ability's `rulesForNamespace` not inheriting `secure_variables` rules or something similar.  We should dig in a bit (provided my ACL is formatted correctly, which I'm not sure about)

ChaiWithJai:
> Running into a few issues here in testing, the most notable of which is a JSON error when trying to load the variables index route with the following ACL:
> 
> ```
> namespace "default" {
>   policy = "write"
>   capabilities = ["alloc-node-exec", "list-jobs"]
>   secure_variables {
> 
>     # full access to secrets in all project paths
>     path "project/*" {
>       capabilities = ["write", "read", "destroy", "list"]
>     }
> 
>     # full access to secrets in all project paths
>     path "*" {
>       capabilities = ["list"]
>     }
> 
>     # read/list access within a "system" path belonging to administrators
>     path "system/*" {
>       capabilities = ["read", "list"]
>     }
>   }
> 
> }
> ```
> 
> I think there's an issue with the abstract ability's `rulesForNamespace` not inheriting `secure_variables` rules or something similar. We should dig in a bit (provided my ACL is formatted correctly, which I'm not sure about)

OK, this should be resolved by [13605](https://github.com/hashicorp/nomad/pull/13605).

philrenaud (approved):

PR #13475, opened 2022-06-24
ui:  handle namespace filtering for capabilities check

ChaiWithJai:
This PR is one of two PRs that will unblock [13447](https://github.com/hashicorp/nomad/pull/13447).

This PR specifically handles namespace filtering when handling path matching.

### DIFF
--- a/ui/app/abilities/variable.js
+++ b/ui/app/abilities/variable.js
@@ -41,21 +41,26 @@ export default class Variable extends AbstractAbility {
 
   @computed('rulesForNamespace.@each.capabilities', 'path')
   get policiesSupportVariableCreation() {
+    const matchingPath = this._nearestMatchingPath(this.path);
     return this.rulesForNamespace.some((rules) => {
-      const keyName = `SecureVariables.Path "${this.path}".Capabilities`;
+      const keyName = `SecureVariables.Path "${matchingPath}".Capabilities`;
       const capabilities = get(rules, keyName) || [];
       return capabilities.includes('create');
     });
   }
 
-  @computed('token.selfTokenPolicies.[]')
+  @computed('token.selfTokenPolicies.[]', '_namespace')
   get allPaths() {
-    return get(this, 'token.selfTokenPolicies')
+    return (get(this, 'token.selfTokenPolicies') || [])
       .toArray()
       .reduce((paths, policy) => {
-        const [variables] = get(policy, 'rulesJSON.Namespaces').mapBy(
-          'SecureVariables'
+        const matchingNamespace = this._findMatchingNamespace(
+          get(policy, 'rulesJSON.Namespaces'),
+          this._namespace
         );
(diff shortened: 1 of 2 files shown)


ui/app/abilities/variable.js
```diff
@@ -41,21 +41,26 @@ export default class Variable extends AbstractAbility {
...
+  @computed('token.selfTokenPolicies.[]', '_namespace')
   get allPaths() {
-    return get(this, 'token.selfTokenPolicies')
+    return (get(this, 'token.selfTokenPolicies') || [])
       .toArray()
       .reduce((paths, policy) => {
-        const [variables] = get(policy, 'rulesJSON.Namespaces').mapBy(
-          'SecureVariables'
+        const matchingNamespace = this._findMatchingNamespace(
+          get(policy, 'rulesJSON.Namespaces'),
```
philrenaud, line 58: I'm running into a fatal JS error in trying to test this ( @ https://nomad-storybook-and-pxvy5gkcu-hashicorp.vercel.app/ui/variables ) where this line resolves to `undefined`.

The policy itself has rulesJSON that looks like the following at this point:
image

philrenaud (changes requested):
This currently blocks access to the /variables route in the UI; there probably needs to be an update to the token and/or policy factories in mirage, plus some logic in case there is no Namespaces property present.

ui/app/abilities/variable.js
ChaiWithJai: When `_findMatchingNamespaces` is invoked, the first argument `policyNamespaces` is `undefined` at runtime. This is fairly strange given that `get(policy, 'rulesJSON.Namespaces')` is also used in the `variables` computation.

That's because at runtime in our development environment, `rulesJSON` returns:

```
{
    "Node": {
        "Policy": "read"
    }
}
```

This is clearly out of order with what we're setting in the Token Mirage factory.

Can you help me spin up a Nomad agent with appropriate ACL tokens for Secure Variables? I'm curious to see if this is a quirk of our Mirage set-up or an actual bug.

At runtime, per our Token Mirage fixture we should expect `rulesJSON` to return:

```
        rulesJSON: {
          Namespaces: [
            {
              Name: 'default',
              Capabilities: ['list-jobs', 'alloc-exec', 'read-logs'],
              SecureVariables: {
                'Path "*"': {
                  Capabilities: ['list', 'create'],
                },
                'Path "blue/berkshire"': {
                  Capabilities: ['list', 'create', 'edit', 'delete'],
                },
              },
            },
          ],
        },
```

philrenaud: My read of [Nomad's ACL policies](https://learn.hashicorp.com/tutorials/nomad/access-control-policies#namespace-rules) is that Namespaces as a top level rules object is optional; there are a few tests within the UI that read `rulesJSON.Node` directly.

I _think_ that's probably still a valid ACL rule

philrenaud (approved):

PR #13482, opened 2022-06-24
Dynamic numeric keynav for multiple tables

philrenaud:
Binds keyboard shortcuts to iterable/enumerable items, such as rows in a table that link to tasks, allocations, etc.

- This moves the existing `cleanPattern()` logic out of the modifier and into the keyboard service, so it can keep track of route state ("what is the full list of enumerable commands on this page?")
- adds an `enumerated = true` param to the `keyboard-command` modifier. If present, it auto-generates the number of the shortcut's pattern within `addCommands()`
- Simplifies the `cleanPattern()` logic a bit since it no longer gets invoked from a template directly

image


ui/app/modifiers/keyboard-shortcut.js
```diff
@@ -1,44 +1,31 @@
 import { inject as service } from '@ember/service';
 import Modifier from 'ember-modifier';
 import { registerDestructor } from '@ember/destroyable';
-import { assert } from '@ember/debug';
 
 export default class KeyboardShortcutModifier extends Modifier {
   @service keyboard;
+  @service router;
```
ChaiWithJai, line 7: `suggestion` `question`:  I'm noticing we're using the `router` service alongside `registerDestructor`. But, I'm wondering if we need all that code because its going to fire specifically after page load.

`element-modifiers` have their own lifecycle that ties to the lifecycle of a DOM element.

Take this example code:

```
export default modifier((element, [callback]) => {
  function handleClick(event) {
    if (!element.contains(event.target)) {
      callback();
    }
  }

  document.addEventListener("click", handleClick);

  return () => {
    document.removeEventListener("click", handleClick);
  };
});
```

The function tears down once the element is removed from the DOM (e.g. changing conditional variable or transitioning to a new route. I think the Keyboard Service could delete some code and be simplified if we follow this pattern.

But, I might be missing something, is there a reason why we're not using the element's lifecycle?

ui/app/components/keyboard-shortcuts-modal.js
```diff
@@ -28,7 +27,7 @@ export default class KeyboardShortcutsModalComponent extends Component {
    * hints: filter keyCommands to those that have an element property,
    * and then compute a position on screen to place the hint.
    */
-  @computed('keyboard.keyCommands.length', 'keyboard.displayHints')
+  @computed('keyboard.{keyCommands.length,displayHints}')
```
ChaiWithJai, line 30: `question`: Why not just declare `length` and `displayHints` as tracked properties here? Are we relying on `computed` properties for a cache?

ui/app/modifiers/keyboard-shortcut.js
```diff
@@ -1,44 +1,31 @@
...
-      pattern = [`Shift+${('0' + patternNumber).slice(-2)}`]; // Shift+01, not Shift+1
+  modify(
+    element,
+    _positional,
+    {
+      label,
+      pattern = '',
+      action = () => {},
+      menuLevel = false,
+      enumerated = false,
```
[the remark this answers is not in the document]
philrenaud: In this PR that's true, but the `keyboard-shortcut` modifier is used elsewhere with enumerated false in many places (main nav for example)

Since it adds behaviour, I'd prefer to have the default state default to false here

ui/app/modifiers/keyboard-shortcut.js
philrenaud: (I would have loved to not have to do this; it was in fact my first attempt. But destroyables in Ember have timings that don't align with practical use for this at all)

ui/app/components/keyboard-shortcuts-modal.js
philrenaud: Could you expand on this a bit? What is the benefit to having these as tracked properties of the modal vs global state in the service?

ChaiWithJai: This is a Glimmer component and the long-term plan is moving to a pull-based reactivity model (tracked properties) and fully away ([deprecating](https://github.com/emberjs/rfcs/issues/536)) from a push-based reactivity model (computeds, observables, two-way bindings).

PR #13485, opened 2022-06-25
api: use testing.T.Setenv to set env vars in tests

shoenig:
Just minor cleanup to use `T.Setenv` instead of `os.Setenv` + `defer`.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -18,8 +18,8 @@ import (
 	"time"
 
 	"github.com/gorilla/websocket"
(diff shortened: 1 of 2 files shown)


lgfa29 (approved):
TIL. Neat!

lgfa29:
Skipped backporting to `release/1.1.x` because `T.SetEnv` was added in Go 1.17 and `release/1.1.x` is on Go 1.16.

PR #13487, opened 2022-06-26
feat: docker/docker/pkg/term has been deprecated in favor of moby/term

greut:
See https://github.com/moby/moby/pull/40825


greut:
semgrep is missing that `git config --global --add safe.directory "${GITHUB_WORKSPACE}"` as Git 2.35.2 kicked in.

PR #13492, opened 2022-06-27
core: allow deleting of evaluations

jrasell:
This PR adds the ability to delete evals and adds to the work performed to allow the pausing of the eval and blocked eval brokers.

There is a change made to the signature of the eval reap RPC which we will want to backport in order to make future backport releases easier. Once this PR has been reviewed and approved, I will open this change as a PR separately.

### DIFF
--- a/command/agent/eval_endpoint_test.go
+++ b/command/agent/eval_endpoint_test.go
@@ -6,7 +6,9 @@ import (
 	"net/http/httptest"
 	"testing"
 
+	"github.com/hashicorp/nomad/api"
 	"github.com/hashicorp/nomad/ci"
+	"github.com/hashicorp/nomad/helper/uuid"
 	"github.com/hashicorp/nomad/nomad/mock"
 	"github.com/hashicorp/nomad/nomad/structs"
 	"github.com/stretchr/testify/require"
@@ -108,6 +110,135 @@ func TestHTTP_EvalPrefixList(t *testing.T) {
 			t.Fatalf("expected eval ID: %v, Actual: %v", eval2.ID, e[0].ID)
 		}
 	})
+}
+
+func TestHTTP_EvalsDelete(t *testing.T) {
+	ci.Parallel(t)
+
+	testCases := []struct {
+		testFn func()
+		name   string
+	}{
+		{
+			testFn: func() {
+				httpTest(t, nil, func(s *TestAgent) {
+
+					// Create an empty request object which doesn't contain any
+					// eval IDs.
+					deleteReq := api.EvalDeleteRequest{}
+					buf := encodeReq(&deleteReq)
+
+					// Generate the HTTP request.
+					req, err := http.NewRequest(http.MethodDelete, "/v1/evaluations", buf)
+					require.NoError(t, err)
+					respW := httptest.NewRecorder()
+
+					// Make the request and check the response.
+					obj, err := s.Server.EvalsRequest(respW, req)
+					require.Equal(t,
+						CodedError(http.StatusBadRequest, "request does not include any evaluation IDs"), err)
+					require.Nil(t, obj)
+				})
+			},
+			name: "too few eval IDs",
+		},
+		{
+			testFn: func() {
+				httpTest(t, nil, func(s *TestAgent) {
+
+					deleteReq := api.EvalDeleteRequest{EvalIDs: make([]string, 8000)}
+
+					// Generate a UUID and add it 8000 times to the eval ID
+					// request array.
+					evalID := uuid.Generate()
+
+					for i := 0; i < 8000; i++ {
+						deleteReq.EvalIDs[i] = evalID
+					}
+
+					buf := encodeReq(&deleteReq)
+
+					// Generate the HTTP request.
+					req, err := http.NewRequest(http.MethodDelete, "/v1/evaluations", buf)
+					require.NoError(t, err)
+					respW := httptest.NewRecorder()
+
+					// Make the request and check the response.
+					obj, err := s.Server.EvalsRequest(respW, req)
+					require.Equal(t,
+						CodedError(http.StatusBadRequest,
+							"request includes 8000 evaluations IDs, must be 7281 or fewer"), err)
+					require.Nil(t, obj)
+				})
+			},
+			name: "too many eval IDs",
+		},
+		{
+			testFn: func() {
+				httpTest(t, func(c *Config) {
+					c.NomadConfig.DefaultSchedulerConfig.PauseEvalBroker = true
+				}, func(s *TestAgent) {
+
+					// Generate a request with an eval ID that doesn't exist
+					// within state.
+					deleteReq := api.EvalDeleteRequest{EvalIDs: []string{uuid.Generate()}}
+					buf := encodeReq(&deleteReq)
+
+					// Generate the HTTP request.
+					req, err := http.NewRequest(http.MethodDelete, "/v1/evaluations", buf)
+					require.NoError(t, err)
+					respW := httptest.NewRecorder()
+
+					// Make the request and check the response.
+					obj, err := s.Server.EvalsRequest(respW, req)
+					require.Contains(t, err.Error(), "eval not found")
+					require.Nil(t, obj)
+				})
+			},
+			name: "eval doesn't exist",
+		},
+		{
+			testFn: func() {
+				httpTest(t, func(c *Config) {
+					c.NomadConfig.DefaultSchedulerConfig.PauseEvalBroker = true
+				}, func(s *TestAgent) {
+
+					// Upsert an eval into state.
+					mockEval := mock.Eval()
+
+					err := s.Agent.server.State().UpsertEvals(
+						structs.MsgTypeTestSetup, 10, []*structs.Evaluation{mockEval})
+					require.NoError(t, err)
+
+					// Generate a request with the ID of the eval previously upserted.
+					deleteReq := api.EvalDeleteRequest{EvalIDs: []string{mockEval.ID}}
+					buf := encodeReq(&deleteReq)
+
+					// Generate the HTTP request.
+					req, err := http.NewRequest(http.MethodDelete, "/v1/evaluations", buf)
+					require.NoError(t, err)
+					respW := httptest.NewRecorder()
+
+					// Make the request and check the response.
+					obj, err := s.Server.EvalsRequest(respW, req)
+					require.Nil(t, err)
+					require.Nil(t, obj)
+
+					// Ensure the eval is not found.
+					readEval, err := s.Agent.server.State().EvalByID(nil, mockEval.ID)
+					require.NoError(t, err)
+					require.Nil(t, readEval)
+				})
+			},
+			name: "successfully delete eval",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			tc.testFn()
+		})
+	}
 }
 
 func TestHTTP_EvalAllocations(t *testing.T) {
--- a/command/eval_delete.go
+++ b/command/eval_delete.go
@@ -1,0 +1,406 @@
+package command
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/hashicorp/nomad/api"
+	"github.com/hashicorp/nomad/api/contexts"
+	"github.com/posener/complete"
+)
+
+type EvalDeleteCommand struct {
+	Meta
+
+	filter string
+	yes    bool
+
+	// deleteByArg is set when the command is deleting an evaluation that has
+	// been passed as an argument. This avoids need for confirmation.
+	deleteByArg bool
+
+	// numDeleted tracks the total evaluations deleted in a single run of this
+	// command. It provides a way to output this information to the user at the
+	// command completion.
+	numDeleted int
+
+	// client is the lazy-loaded API client and is stored here, so we don't
+	// need to pass it to multiple functions.
+	client *api.Client
+}
+
+func (e *EvalDeleteCommand) Help() string {
+	helpText := `
+Usage: nomad eval delete [options] <evaluation>
+
+  Delete an evaluation by ID. If the evaluation ID is omitted, this command
+  will use the filter flag to identify and delete a set of evaluations. If ACLs
+  are enabled, this command requires a management ACL token.
+
+  This command should be used cautiously and only in outage situations where
+  there is a large backlog of evaluations not being processed. During most
+  normal and outage scenarios, Nomads reconciliation and state management will
+  handle evaluations as needed.
+
+  The eval broker is expected to be paused prior to running this command and
+  un-paused after. This can be done using the following two commands:
+    - nomad operator scheduler set-config -pause-eval-broker=true
+    - nomad operator scheduler set-config -pause-eval-broker=false
+
+General Options:
+
+  ` + generalOptionsUsage(usageOptsNoNamespace) + `
+
+Eval Delete Options:
+
+  -filter
+    Specifies an expression used to filter evaluations by for deletion. When
+    using this flag, it is advisable to ensure the syntax is correct using the
+    eval list command first.
+
+  -yes
+    Bypass the confirmation prompt if an evaluation ID was not provided.
+`
+
+	return strings.TrimSpace(helpText)
+}
+
+func (e *EvalDeleteCommand) Synopsis() string {
+	return "Delete evaluations by ID or using a filter"
+}
+
+func (e *EvalDeleteCommand) AutocompleteFlags() complete.Flags {
+	return mergeAutocompleteFlags(e.Meta.AutocompleteFlags(FlagSetClient),
+		complete.Flags{
+			"-filter": complete.PredictAnything,
+			"-yes":    complete.PredictNothing,
+		})
+}
+
+func (e *EvalDeleteCommand) AutocompleteArgs() complete.Predictor {
+	return complete.PredictFunc(func(a complete.Args) []string {
+		client, err := e.Meta.Client()
+		if err != nil {
+			return nil
+		}
+
+		resp, _, err := client.Search().PrefixSearch(a.Last, contexts.Evals, nil)
+		if err != nil {
+			return []string{}
+		}
+		return resp.Matches[contexts.Evals]
+	})
+}
+
+func (e *EvalDeleteCommand) Name() string { return "eval delete" }
+
+func (e *EvalDeleteCommand) Run(args []string) int {
+
+	flags := e.Meta.FlagSet(e.Name(), FlagSetClient)
+	flags.Usage = func() { e.Ui.Output(e.Help()) }
+	flags.StringVar(&e.filter, "filter", "", "")
+	flags.BoolVar(&e.yes, "yes", false, "")
+	if err := flags.Parse(args); err != nil {
+		return 1
+	}
+
+	args = flags.Args()
+
+	if err := e.verifyArgsAndFlags(args); err != nil {
+		e.Ui.Error(fmt.Sprintf("Error validating command args and flags: %v", err))
+		return 1
+	}
+
+	// Get the HTTP client and store this for use across multiple functions.
+	client, err := e.Meta.Client()
+	if err != nil {
+		e.Ui.Error(fmt.Sprintf("Error initializing client: %s", err))
+		return 1
+	}
+	e.client = client
+
+	// Ensure the eval broker is paused. This check happens multiple times on
+	// the leader, but this check means we can provide quick and actionable
+	// feedback.
+	schedulerConfig, _, err := e.client.Operator().SchedulerGetConfiguration(nil)
+	if err != nil {
+		e.Ui.Error(fmt.Sprintf("Error querying scheduler configuration: %s", err))
+		return 1
+	}
+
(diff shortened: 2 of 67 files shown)


nomad/structs/structs.go
```diff
@@ -822,8 +822,10 @@ type EvalUpdateRequest struct {
 	WriteRequest
 }
 
-// EvalDeleteRequest is used for deleting an evaluation.
-type EvalDeleteRequest struct {
+// EvalReapRequest is used for reaping evaluations and allocation. This action
+// is performed by the core scheduler which is periodic, or manually by
+// operators using the eval delete RPC.
```
shoenig, line 827: Is it the case an `EvalReapRequest` can be manually triggered?

nomad/structs/eval.go
```diff
@@ -0,0 +1,24 @@
...
+	// EvalDeleteRPCMethod is the RPC method for batch deleting evaluations
+	// using their IDs.
+	//
+	// Args: EvalDeleteRequest
+	// Reply: EvalDeleteResponse
+	EvalDeleteRPCMethod = "Eval.Delete"
+)
+
+// EvalDeleteRequest is the request object used when operators are manually
+// deleting evaluations. The number of evaluations IDs within the request must
```
shoenig, line 13:
```suggestion
// deleting evaluations. The number of evaluation IDs within the request must
```

command/eval_delete.go
```diff
@@ -0,0 +1,529 @@
...
+	// per page value is less than the total number we can include in a single
+	// delete request. This allows us to keep the maximum return object of the
+	// eval list call to fairly small.
+	//
+	// The JSON serialized evaluation API object is 350-380B in size.
+	// 2426 * 380B (3.8e-4 MB) = 0.92MB. We may want to make this configurable
+	// in the future, but this is counteracted by the CLI logic which will loop
+	// until the user tells it to exit, or all evals matching the filter are
+	// deleted. 2426 * 3 falls below the maximum limit for eval IDs in a single
+	// delete request.
```
shoenig, line 341:
```suggestion
	// deleted. 2426 * 3 falls below the maximum limit for eval IDs in a single
	// delete request (set by MaxEvalIDsPerDeleteRequest).
```

command/eval_delete.go
```diff
@@ -0,0 +1,529 @@
...
+			return 0, nil
+		} else {
+			return 1, errors.New("failed to find any evals that matched filter criteria")
+		}
+	}
+
+	if code, err := e.batchDelete(evalsToDelete); err != nil {
+		return code, err
+	}
+
```
shoenig, line 385: would it make sense to log progress here?

website/content/api-docs/evaluations.mdx
```diff
@@ -204,6 +204,55 @@ $ curl \
...
+## Delete Evaluations
+
+This endpoint deletes evaluations. In order to utilise this endpoint the
+eval broker should be paused via the
+[update_scheduler_configuration][operator scheduler update configuration] API
+endpoint.
+
+This API endpoint should be used cautiously and only in outage situations where
+there is a large backlog of evaluations not being processed. During most normal
+and outage scenarios, Nomads reconciliation and state management will handle
```
shoenig, line 216:
```suggestion
and outage scenarios, Nomad's reconciliation and state management will handle
```

website/content/api-docs/evaluations.mdx
```diff
@@ -204,6 +204,55 @@ $ curl \
...
+[required ACLs](/api-docs#acls).
+
+| Blocking Queries | ACL Required |
+| ---------------- | ------------ |
+| `NO`             | `management` |
+
+### Parameters
+
+- `:EvalIDs` `(array<string>: <required>)`- An array of evaluation UUIDs to
+  delete. This must be the full UUID, not the short 8-character one.
```
shoenig, line 234:
```suggestion
  delete. This must be a full length UUID and not a prefix.
```

website/content/docs/commands/eval/delete.mdx
```diff
@@ -0,0 +1,82 @@
...
+description: |
+The eval delete command is used to delete evaluations.
+---
+
+# Command: eval delete
+
+The `eval delete` command is used to delete evaluations. It should be used
+cautiously and only in outage situations where there is a large backlog of
+evaluations not being processed. During most normal and outage scenarios,
+Nomads reconciliation and state management will handle evaluations as needed.
```
shoenig, line 13:
```suggestion
Nomad's reconciliation and state management will handle evaluations as needed.
```

nomad/eval_endpoint.go
```diff
@@ -416,6 +417,149 @@ func (e *Eval) Reap(args *structs.EvalDeleteRequest,
...
+	if err != nil {
+		return err
+	}
+
+	// Update the index and return.
+	reply.Index = index
+	return nil
+}
+
+// evalDeleteSafe ensures an evaluation is safe to delete based on its related
```
shoenig, line 501: Would be helpful to have this behavior described in the docs

shoenig (approved):
LGTM, this is great @jrasell! Just the usual nits/suggestions

nomad/structs/eval.go
```diff
@@ -0,0 +1,24 @@
+package structs
```
schmichael, line 1: As much as I love splitting this package up into more files, I'm not sure about just moving EvalDeleteRequest off on its own.

Should all Requests/Responses go together? Should we group by object type?

Do we even need a distinct type? I think we could just use the existing EvalDeleteRequest for everything and let `Allocs` be empty.

Not a big deal. Just a bit worried about splitting such similar types up might cause confusion.

website/content/docs/commands/eval/delete.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+@include 'general_options.mdx'
+
+## Delete Options
+
+- `-scheduler`: Filter by scheduler type (one of "batch", "service", "system",
+  or "sysbatch"). This option may be specified multiple times.
+
+- `-status`: Filter by eval status (one of "blocked", "pending", "complete",
+  "failed", or "canceled"). This option may be specified multiple times.
```
lgfa29, line 37:
```suggestion
- `-status`: Filter by eval status (one of `blocked`, `pending`, `complete`,
  `failed`, or `canceled`). This option may be specified multiple times.
```

website/content/docs/commands/eval/delete.mdx
```diff
@@ -0,0 +1,82 @@
...
+  times.
+
+- `-deployment`: Filter by an associated deployment ID. This option may be
+  specified multiple times.
+
+- `-node`: Filter by an associated node ID. This option may be specified multiple
+  times.
+
+- `-filter-operator`: The filter operator to use when multiple filter options are
+  specified. This defaults to "and" but also supports "or".
```
lgfa29, line 49:
```suggestion
  specified. This defaults to `and` but also supports `or`.
```

website/content/api-docs/evaluations.mdx
```diff
@@ -204,6 +204,55 @@ $ curl \
...
+[blocking queries](/api-docs#blocking-queries) and
+[required ACLs](/api-docs#acls).
+
+| Blocking Queries | ACL Required |
+| ---------------- | ------------ |
+| `NO`             | `management` |
+
+### Parameters
+
+- `:EvalIDs` `(array<string>: <required>)`- An array of evaluation UUIDs to
```
lgfa29, line 233:
```suggestion
- `EvalIDs` `(array<string>: <required>)`- An array of evaluation UUIDs to
```

lgfa29 (commented):
Some early comments, but looking good so far! I will give it a try more thoroughly tomorrow to see how unhappy the scheduler gets with vanishing evals 😄

nomad/eval_endpoint.go
```diff
@@ -416,6 +417,149 @@ func (e *Eval) Reap(args *structs.EvalDeleteRequest,
...
+// allocation and job information. This follows similar, but different rules to
+// the eval reap checking, to ensure evaluations for running allocs or allocs
+// which need the evaluation detail are not deleted.
+func evalDeleteSafe(allocs []*structs.Allocation, job *structs.Job) bool {
```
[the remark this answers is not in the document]
schmichael: No, but we only allow deleting when the eval broker is paused and the queue is re-populated after being unpaused.

However, you made me realize we don't enforce the eval broker is paused in the *fsm* which is required to ensure deletes don't interleave with queue pauses/unpauses.

In the Periodic GC case we're only deleting terminal evals which cannot be in the eval broker(s) anyway, so there's no need to worry about mutating the state store in a way that makes it inconsistent with the queues.

However in this delete path I don't think there's anything stopping someone from unpausing the queue during an eval delete.

Luckily the queue is paused/resumed from Raft so we can serialize these operations! I think the FSM handler for deleting evals needs to check `SchedulerConfig.PauseBroker` whenever being told to delete an eval manually and refuse to if the broker is running.

command/agent/eval_endpoint_test.go
```diff
@@ -110,6 +112,135 @@ func TestHTTP_EvalPrefixList(t *testing.T) {
...
+					respW := httptest.NewRecorder()
+
+					// Make the request and check the response.
+					obj, err := s.Server.EvalsRequest(respW, req)
+					require.Equal(t,
+						CodedError(http.StatusBadRequest, "request does not include any evaluation IDs"), err)
+					require.Nil(t, obj)
+				})
+			},
+			name: "too few eval IDs",
```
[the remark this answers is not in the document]
jrasell: Noted. I think I do it this way because of the process in which I build the tests which starts with what the inputs are. I have no doubt this is done in many places where I have written tests, so something I can look into changing moving forward.

nomad/structs/structs.go
jrasell: The comment is because the reap and delete RPCs ultimately use the same encapsulation struct to delete evals via Raft. This avoids the need for new messages. I'll try and clear up the comment to make this clearer.

nomad/eval_endpoint.go
jrasell: Yes but I want to do that for both eval reap (GC) and eval delete (operator) so will tackle this separately.

command/eval_delete.go
```diff
@@ -0,0 +1,529 @@
...
+	// Always attempt to revert the state of the eval broker. In the event this
+	// succeeds, but the eval deletion failed, maintain the non-zero exit code
+	// so failures are clear.
+	return func(exitCode int) int {
+		if err := e.revertEvalBroker(); err != nil {
+			e.Ui.Error(fmt.Sprintf("Error reverting eval broker state: %s", err))
+			exitCode = 1
+		}
+		return exitCode
+	}(exitCode)
```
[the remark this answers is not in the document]
jrasell: Yeah you're correct. This is leftover from previous iterations where I needed the defer, but this can be cleaned up now, thanks.

nomad/eval_endpoint.go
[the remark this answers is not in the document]
tgross: A wrench in this: the scheduler config isn't in raft, right? So if we check it in the FSM we could have inconsistent views between the leader and the followers when we apply the entry and then the followers wouldn't apply the replicated logs.

command/eval_delete.go
```diff
@@ -0,0 +1,529 @@
...
+				api.JobTypeService,
+				api.JobTypeBatch,
+				api.JobTypeSystem,
+				api.JobTypeSysBatch,
+			),
+			"-status":          complete.PredictAnything,
+			"-job":             complete.PredictAnything,
+			"-deployment":      complete.PredictAnything,
+			"-node":            complete.PredictAnything,
+			"-filter-operator": complete.PredictSet("and", "or"),
```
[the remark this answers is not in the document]
jrasell: > Rather than expose operators and HashiCorp field teams to an unsafely sharp tool, the initial implementation of this command will require exercising fine-grained control over which evaluations to delete

I went down this road according to what was written and approved within the RFC document. If we now prefer to go down the route which keeps consistency with other commands I can go ahead and pull this out and implement the single filter flag. This at least had an added benefit that operator can test the filter via the list command first, before running it on the delete.

command/eval_delete.go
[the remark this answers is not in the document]
jrasell: I'll add some additional logging around the changing of the scheduler config state to help this problem. The alternative is to remove this state change from the command altogether, and require operators to do this before running this command. I don't feel this is the correct option though, as it adds additional overhead to operators when it's likely this will be used in stressful situations.

It's also worth noting that by default, the command when using filters go through two confirmation steps before each batch of deletions. This should help reduce the regret it moments caused.

command/eval_delete.go
```diff
@@ -0,0 +1,529 @@
...
+		return 1
+	}
+	e.client = client
+
+	// Pause the eval broker to ensure no more work can be dequeued and
+	// processed by the schedulers.
+	if err := e.pauseEvalBroker(); err != nil {
+		e.Ui.Error(fmt.Sprintf("Error pausing eval broker: %s", err))
+		return 1
+	}
```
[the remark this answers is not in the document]
jrasell: Moving to single thread with other comments in this area.

nomad/eval_endpoint.go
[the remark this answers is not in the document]
jrasell: Good points.

I think in order to achieve this without the need for a new Raft message I can add a new boolean field on the request that gets sent via Raft which is used to indicate whether the operation was user initiated. This will default to false mimicking the reap behaviour, and the new RPC eval delete endpoint will set this to true which will then force the state store to perform a lookup within a single transaction to confirm the broker is paused.

command/eval_delete.go
jrasell: Yeah good shout, i'll add a line per iteration so things are very clear.

nomad/structs/eval.go
jrasell: I have a preference that all RPC's have their own req and reply arguments. I find it makes reading easier, any changes in the future simpler as they only affect a single RPC, and avoids the requirement for understanding that certain fields are unused when using certain RPCs.

Arguably how definitions are grouped depends on what exists and factors such as import cycles. Nomad's structure involves a single pkg (structs) which contains all the RPC definitions, object structures, and helper functions. I therefore prefer to have individual files pertaining to each RPC endpoint object such as `Eval` or `ServiceRegistration`.

command/eval_delete.go
```diff
@@ -0,0 +1,525 @@
...
+
+	// If there is another page of evaluations matching the filter, call this
+	// function again and delete the next batch of evals. We pause for a 500ms
+	// rather than just run as fast as the code and machine possibly can. This
+	// means deleting 13million evals will take roughly 13-15 mins, which seems
+	// reasonable.
+	if opts.NextToken != "" {
+		time.Sleep(500 * time.Millisecond)
+		return e.handleFlagFilterDelete(opts.NextToken, filter)
```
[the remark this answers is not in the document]
lgfa29: Measuring this will be super interesting! I think the compiler may be able to optimize this into a tail recursion?

command/eval_delete.go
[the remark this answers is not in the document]
jrasell: This change has been made within commit https://github.com/hashicorp/nomad/pull/13492/commits/1ef732e2e96a690cda3ed14737abbaa88836ddbd

command/eval_delete.go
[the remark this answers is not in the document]
jrasell: This has now been resolved within https://github.com/hashicorp/nomad/pull/13492/commits/432226048f5d0e83c357a68408e51da726502068

command/eval_delete.go
jrasell: This is resolved within https://github.com/hashicorp/nomad/pull/13492/commits/26e1e28dfa5290563af8a20748d865c2118da0ce along with another problem I found.

command/eval_delete.go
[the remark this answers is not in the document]
jrasell: I wasn't able to find any detail regarding tail recursion optimisation and for the avoidance of unknowns I have altered the behaviour within https://github.com/hashicorp/nomad/pull/13492/commits/4a7cd5f20c305b724ec8f37d09c4c21a0293df11. Although I wouldn't expect any operator to delete 13 millions evals in a single filter query, this command is designed for those types of situations so it's possible.

> we could ensure the evals weren't on the stack

I am a little lost on this point and would appreciate some tips to full understand how this could be achieved or pointers to documentation. Is this the idea of using pointers and such which are known to escape to the heap?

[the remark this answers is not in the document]
tgross: (Forgive me if I'm saying anything you already know here, but just a quick brain-dump on the subject.)

Normally locals end up on the stack, which is where we want them for performance. Freeing the stuff that's on the stack is a matter of decrementing the stack pointer and there's no garbage collection required. But sometimes the compiler can't tell if something is safe to have on the stack b/c it might be reused outside the function call (ex. in a caller, on another goroutine, etc), and that's "escape analysis". The entirety of [Go's documentation on this]((https://github.com/golang/go/wiki/CompilerOptimizations#escape-analysis)
) is somewhat laconic:

> Gc compiler does global escape analysis across function and package boundaries. However, there are lots of cases where it gives up. For example, anything assigned to any kind of indirection (*p = ...) is considered escaped. Other things that can inhibit analysis are: function calls, package boundaries, slice literals, subslicing and indexing, etc. Full rules are too complex to describe, so check the -m output.

For recursive functions with potentially large numbers of executions like this, we _want_ large objects to escape to the heap, because that'll reduce the risk we use up the available stack space. And fortunately in Go's case it gives up pretty easily. We can compile with `-gcflags -m` to see whether the compiler escapes any particular line. Running this on Nomad would require digging around a bit in the output. But for a much smaller example, I whipped up this program that queries a dev agent:

<details><summary>silly program</summary>

```go
package main

import (
	"encoding/json"
	"fmt"
	"net/http"
)

type NomadAgentHealthResponse struct {
	Client Inner `json:"client"`
	Server Inner `json:"server"`
}
type Inner struct {
	Message string `json:"message"`
	Ok      bool   `json:"ok"`
}

const healthURL = "http://localhost:4646/v1/agent/health"

var client = &http.Client{}

func recurse(i int) error {
	if i < 1 {
		return nil
	}
	raw, err := client.Get(healthURL)
	if err != nil {
		return err
	}
	resp := &NomadAgentHealthResponse{}
	defer raw.Body.Close()

	err = json.NewDecoder(raw.Body).Decode(resp)
	if err != nil {
		return err
	}
	fmt.Println(resp)
	i--
	return recurse(i)
}

func noRecurse(i int) error {
	for i := i; i > 0; i-- {
		raw, err := client.Get(healthURL)
		if err != nil {
			return err
		}
		resp := &NomadAgentHealthResponse{}
		defer raw.Body.Close()

		err = json.NewDecoder(raw.Body).Decode(resp)
		if err != nil {
			return err
		}
		fmt.Println(resp)
	}
	return nil
}

func main() {

	err := recurse(5)
	if err != nil {
		fmt.Println(err)
	}
	err = noRecurse(5)
	if err != nil {
		fmt.Println(err)
	}

}
```

</details>

When we run this, we can see the objects we created do escape to the heap in both the recursive and non-recursive cases. You could use the same tooling (plus a bunch of grepping for the appropriate line 😀 ) to determine if your objects escaped to the heap here as well.

```
$ go run -gcflags -m .
# escape
./main.go:33:23: inlining call to json.NewDecoder
./main.go:37:13: inlining call to fmt.Println
./main.go:51:24: inlining call to json.NewDecoder
./main.go:55:14: inlining call to fmt.Println
./main.go:64:14: inlining call to fmt.Println
./main.go:68:14: inlining call to fmt.Println
./main.go:30:10: &NomadAgentHealthResponse{} escapes to heap
./main.go:33:23: &json.Decoder{...} escapes to heap
./main.go:37:13: ... argument does not escape
./main.go:48:11: &NomadAgentHealthResponse{} escapes to heap
./main.go:51:24: &json.Decoder{...} escapes to heap
./main.go:55:14: ... argument does not escape
./main.go:64:14: ... argument does not escape
./main.go:68:14: ... argument does not escape
&{{ok true} {ok true}}
&{{ok true} {ok true}}
&{{ok true} {ok true}}
&{{ok true} {ok true}}
&{{ok true} {ok true}}
&{{ok true} {ok true}}
&{{ok true} {ok true}}
&{{ok true} {ok true}}
&{{ok true} {ok true}}
&{{ok true} {ok true}}
```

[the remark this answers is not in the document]
jrasell: Thanks @tgross, of all the documentation i've found online today and previously on this subject this is certainly some of the best conscience detail.

It's worth noting I did compile the command pkg using the `-m`  ldflag earlier and couldn't see any relevant lines to the recursion being escaped (although the output was hard to grok) so I think this change was certainly worth it.

nomad/structs/eval.go
tgross: I think we're in agreement that splitting up the giant `structs.go` is good. So I think there's just a follow-up refactoring PR to be done here after this gets merged, which is to move the _other_ eval-related stuff in `structs.go` into this file too.

command/eval_delete.go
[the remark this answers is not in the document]
schmichael: Ugh sorry for missing this in RFC review James. :(

schmichael (approved):
Amazing work. Sorry for the last second changes and late reviews from me.

Issue #13483, opened 2022-06-24
Nomad's native service registration fails if raw_exec and docker drivers are combined in the same task group


groggemans:
### Nomad version
Nomad v1.3.1 (2b054e38e91af964d1235faa98c286ca3f527e56)

### Operating system and Environment details
Linux + docker

### Issue
Using native service registration fails if a raw_exec and docker tasks are combined.
(using two docker tasks in the same task group isn't an issue)

### Reproduction steps
Below example jobs runs two counting loops in a sh script, one in a container, the other directly on the host.
There is a dummy port and service registration that's not actually used, but that does not make a difference. (The initial discovery was with a job that actually used the port, but it's not needed to replicate the behavior)
I discovered it with a pretask hook as the raw_exec task, but using an actually long living task seems to have the same effect.

#### Expected Result
The service is registered and can be checked with `nomad service list`.

#### Actual Result
The service fails to register with below error;

```
Task hook failed | task_services: rpc error: service registration insert failed: object missing primary index
```

### Job file (if appropriate)

```hcl
job "example" {
  datacenters = ["dc1"]

  group "counter" {
    network {
      port "random" {
      }
    }

    task "raw" {
      driver = "raw_exec"
      config {
            command = "${NOMAD_TASK_DIR}/my-script.sh"
      }
      template {
        destination = "${NOMAD_TASK_DIR}/my-script.sh"
        perms = "755"
        data = <<EOF
#!/bin/sh
i=0

while true; do
  echo "$i"
  i=$((i+1))
  sleep 2
done

EOF
      }
    }

    task "docker" {
      driver = "docker"
      config {
        image = "alpine:3.13.4"
        command = "${NOMAD_TASK_DIR}/my-script.sh"
        ports = ["random"]
      }
      service {
        provider = "nomad"
        name = "counter"
        port = "random"
      }

      template {
        destination = "${NOMAD_TASK_DIR}/my-script.sh"
        perms = "755"
        data = <<EOF
#!/bin/sh
i=0

while true; do
  echo "$i"
  i=$((i+1))
  sleep 2
done

EOF
      }
    }
  }
}
```

attachmentgenie:
@jrasell is this supposed to be part of 1.3.2? in any case i am still observing it there too

lgfa29:
> is this supposed to be part of 1.3.2?

[...]

PR #13493, opened 2022-06-27
client: fixed a problem calculating a service namespace.

jrasell:
When calculating a services namespace for registration, the code
assumed the first task within the task array would include a
service block. This is incorrect as it is possible only a latter
task within the array contains a service definition.

This change fixes the logic, so we correctly search for a service
definition before identifying the namespace.

closes https://github.com/hashicorp/nomad/issues/13483

### DIFF
--- a/.changelog/13493.txt
+++ b/.changelog/13493.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+client: Fixed a problem calculating a services namespace
+```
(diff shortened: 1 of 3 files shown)


DerekStrickland (approved):
:shipit:

shoenig (approved):
LGTM!

PR #13511, opened 2022-06-28
docs: fixup HCL2 index collection function documentation.

### LINKED ISSUE #13494: Wrong documentation for hcl index function
### Nomad version
v1.3.1 (2b054e38e91af964d1235faa98c286ca3f527e56)

### Issue
Documentation for the hcl index function (https://github.com/hashicorp/nomad/blob/main/website/content/docs/job-specification/hcl2/functions/collection/index-fn.mdx) is incorrect. It claims to return the index of an element, but instead returns the element at a index. E.g. `index(["a", "b", "c"], 1)` returns `b` instead of `index(["a", "b", "c"], "b")` returning `1`. See https://pkg.go.dev/github.com/zclconf/go-cty/cty/function/stdlib?utm_source=godoc#Index for the function that is being used.

That being said, a function that does what the documentation currently claims it does would be very useful.

### Reproduction steps
Use `${index(["a", "b", "c"], "b")}` in a job file and run it.

#### Expected Result
Replaces `${index(["a", "b", "c"], "b")}` with `1` in the job.

#### Actual Result
Error in function call; Call to function "index" failed: key for tuple must be number., and 1 other diagnostic(s)

jrasell:
closes #13494

### DIFF
--- a/website/content/docs/job-specification/hcl2/functions/collection/index-fn.mdx
+++ b/website/content/docs/job-specification/hcl2/functions/collection/index-fn.mdx
@@ -1,25 +1,30 @@
 ---
 layout: docs
 page_title: index - Functions - Configuration Language
-description: The index function finds the element index for a given value in a list.
+description: |-
+  The index function returns an element from the given list or map
+  using the given index or key.
 ---
 
 # `index` Function
 
-`index` finds the element index for a given value in a list.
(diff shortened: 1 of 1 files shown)


DerekStrickland (approved):
Nice clarification

PR #9716, opened 2021-01-03
Version update for demo Vagrant: Ubuntu 20.04, Nomad 1.0.1, Consul 1.9.0

nileshcs:
Vagrant was using Ubuntu 18.04 as base; update to use 20.04

Also updated Nomad and Consul versions to the latest

PR #13518, opened 2022-06-28
Demo Vagrant version updates (nomad, consul, ubuntu)

TheConner:
- Use latest nomad 1.3.1 and consul 1.12.2
- Use latest Ubuntu LTS (easier to get podman plugin working, 18.04 does not have podman in main repositories).
- Use generic Ubuntu image instead of bento, generic supports libvirt KVM provider in addition to all other providers bento supports
- Add increased memory for libvirt to match other VM providers

If the issues with snap brought up in #9716 are still present, I also validated Ubuntu 18.04 still works (not with podman plugin, of course)

### DIFF
--- a/demo/vagrant/Vagrantfile
+++ b/demo/vagrant/Vagrantfile
@@ -25,7 +25,7 @@ sudo docker --version
 sudo apt-get install unzip curl vim -y
 
 echo "Installing Nomad..."
-NOMAD_VERSION=1.0.1
+NOMAD_VERSION=1.3.1
 cd /tmp/
 curl -sSL https://releases.hashicorp.com/nomad/${NOMAD_VERSION}/nomad_${NOMAD_VERSION}_linux_amd64.zip -o nomad.zip
 unzip nomad.zip
@@ -35,7 +35,7 @@ sudo chmod a+w /etc/nomad.d
 
 
 echo "Installing Consul..."
-CONSUL_VERSION=1.9.0
+CONSUL_VERSION=1.12.2
 curl -sSL https://releases.hashicorp.com/consul/${CONSUL_VERSION}/consul_${CONSUL_VERSION}_linux_amd64.zip > consul.zip
 unzip /tmp/consul.zip
 sudo install consul /usr/bin/consul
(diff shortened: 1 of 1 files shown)


demo/vagrant/Vagrantfile
```diff
@@ -69,13 +69,18 @@ nomad -autocomplete-install
 SCRIPT
 
 Vagrant.configure(2) do |config|
-  config.vm.box = "bento/ubuntu-18.04" # 18.04 LTS
+  config.vm.box = "generic/ubuntu2204" # 22.04 LTS, Jammy
```
tgross, line 72: The unfortunately-named [generic](https://app.vagrantup.com/generic) org is not some "official" repo of common boxes, but one maintained by a company Lavabit LLC. We've been quite happy with `bento`, so we should keep it even if we decide to bump to the newer LTS.

TheConner: Ah yes, I did seem to miss that detail. I assumed generic was indeed some kind of generic provider and not ones maintained by another company. Looking into the Bento project, I also noticed that Bento is maintained by Chef which is owned by Progress Software.

For making the libvirt provider work, there are a few options I can think of:
1. Use something like [vagrant-mutate](https://github.com/sciurus/vagrant-mutate) to convert the image type, a bit of a hack
2. Wait for / help the Bento project with getting libvirt builds set up, some [context here](https://github.com/chef/bento/issues/1405). It seems like libvirt support is blocked by some infra issues on their end.
3. Use Ubuntu's official [cloud images](http://cloud-images.ubuntu.com/daily/server/jammy/current/) for Vagrant, although these unfortunately support only the [Virtualbox provider](https://app.vagrantup.com/ubuntu/boxes/jammy64)

Given all these solutions aren't ideal, I'll revert to the Bento version of 22.04 LTS.

tgross: > Looking into the Bento project, I noticed that Bento is maintained by Chef which is owned by Progress Software.

Yes, it's also third party but we have a history of success with them (and use them in our own development environments, so they're well-tested).

tgross (approved):
LGTM.

demo/vagrant/Vagrantfile
tgross: I'm fine to merge as-is and when upstream gets their libvirt situation sorted we'll get that for free.

PR #13528, opened 2022-06-29
Replaced other instances of IvyCodeMirror throughout the app

philrenaud:


ui/app/templates/components/job-editor.hbs
```diff
@@ -51,17 +51,14 @@
...
+      <div
         data-test-editor
-        aria-label="Job definition"
-        @value={{or this.job._newDefinition this.jobSpec}}
-        @valueUpdated={{this.updateCode}}
-        @options={{hash
-          mode="javascript"
+        {{code-mirror
+          screenReaderLabel="Job definition"
+          content=(or this.job._newDefinition this.jobSpec)
```
ChaiWithJai, line 58: `blocking` `question`: `JobEditor` doesn't have a `jobSpec` property and neither of these properties are reactive, however, we call `set` on `_newDefintion` which might do the trick.

Do we know where `this.jobSpec` is coming from? Searching the codebase, this is the only use of `jobSpec`?

ui/app/templates/components/json-viewer.hbs
```diff
@@ -1,11 +1,9 @@
-<IvyCodemirror
+<div
   data-test-json-viewer
-  @value={{this.jsonStr}}
-  @options={{hash
-    mode="javascript"
+  {{code-mirror
+    content=this.jsonStr
```
ChaiWithJai, line 4: `suggestion`:  If you invoke the `stringifyObject` helper here and pass `this.json` instead, we can delete the backing component class.

ui/tests/helpers/codemirror.js
```diff
@@ -1,18 +1,6 @@
...
-    const element = document.querySelector(selector);
-    invariant(element, `Selector ${selector} matched no elements`);
-
-    const cm = cmService.instanceFor(element.id);
-    invariant(cm, `No registered CodeMirror instance for ${selector}`);
-
-    return cm;
+export function getCodeMirrorInstance() {
+  return function () {
+    return document.querySelector('.CodeMirror').CodeMirror;
```
ChaiWithJai, line 3: `suggestion`: We can deprecate this helper entirely now this its just selecting a DOM element without tapping into the container registry or throwing an error.

Its used in 3 files. And we should be able to write:

```
const expected = document.querySelector('.CodeMirror').CodeMirror.value;
const actual = JSON.stringify(result, null, 2);

assert.equals(expected, actual);
```

ui/app/components/secure-variable-form.js
```diff
@@ -201,7 +201,7 @@ export default class SecureVariableFormComponent extends Component {
   }
 
   get stringifiedItems() {
-    return JSON.stringify(this.args.model.items, null, 2);
+    return stringifyObject([this.args.model.items]);
```
ChaiWithJai, line 204: `nit`:  We could just call this directly in the template and avoid adding another variable to the template.

ChaiWithJai (approved):
There's some more deprecations and cleaning up we can do while we're here.

ui/app/templates/components/job-editor.hbs
ChaiWithJai: This could introduce a potential breaking change and thankfully it'll be shipped in a major release. But, I'd love to chat with @lgfa29 re:  UI release process.

@lgfa29 - we have some serious re-designs for the Job Detail Overview page (this PR is unrelated to that) coming down the pipe. But, this PR is a good example of potential breaking API changes in our UI code. Should we start coordinating with the release process for these types of things?

lgfa29: I don't think we need any special coordination. Since this is all set to go out with 1.4.0 all we need is for the changes to be in `main` before the release.

One thing we can do is to update `bindata_assetfs.go` in `main` so that anyone running `make dev` will see the latest UI changes before the release. This can help get more eyes and test major changes before the beta.

PR #13529, opened 2022-06-29
test:  acceptance test for editing a variable

ChaiWithJai:
This PR does a basic test for editing a variable and not having permissions to edit a variable. The second test should fail because we have not implemented getting editing capabilities yet.


philrenaud (approved):
LGTM!

Issue #12918, opened 2022-05-06
Eval status command does not respect namespace wildcard


mikenomitch:
### Nomad version
v1.3.0-rc1

### Operating system and Environment details

MacOS

### Issue

When passing in the -namespace="*" flag into the "eval status" command, it returns a 404 instead of returning the proper evaluation.

### Reproduction steps

* Get an eval ID
* Get its info using the status command "nomad eval status -namespace="default" baa193b9"
* Note it returns the info
* Try to get its info using a namespace wildcard "nomad eval status -namespace="*" baa193b9"
* Note it says it is not found

#### Expected Result

The wildcard value should allow getting eval info from any namespace.

#### Actual Result

The eval info is not returned.

![Namespace Wildcard](https://user-images.githubusercontent.com/2732204/167219502-918fd62d-c82c-493d-a88a-674fedfe7937.jpg)

PR #13530, opened 2022-06-29
Failing test and TODO for wildcard

philrenaud:


ui/app/controllers/evaluations/index.js
```diff
@@ -35,7 +35,7 @@ export default class EvaluationsController extends Controller {
     'currentEval',
     'pageSize',
     'status',
-    'qpNamespace',
+    { qpNamespace: 'namespace' },
```
ChaiWithJai, line 38: `blocking`: I don't think this resolves the issue. This sets the alias of the query parameter in the URL. But, this shouldn't impact formatting the request in the route.

image

If we need to make any edits, I would suspect we'd need to check the formatting logic in `generateFilterExpression`. However, the [tests](tests/acceptance/evaluations-test.js) assert on how user actions impact the request we're sending and the default request we send when we enter the view. I'd love to learn if we wrote those tests and the formatting logic incorrectly.

ChaiWithJai (changes requested):
If we need to make any edits client-side, I would suspect we'd need to check the formatting logic in generateFilterExpression. However, the [tests](https://github.com/hashicorp/nomad/pull/13530/tests/acceptance/evaluations-test.js) assert on how user actions impact the request we're sending and the default request we send when we enter the view. I'd love to learn if we wrote those tests and the formatting logic incorrectly.

I highly recommend reading that filtering logic and the tests to see if we got something wrong. But, I believe @mikenomitch pointed the real issue out [12918](https://github.com/hashicorp/nomad/issues/12918) and we should ask @lgfa29 to prioritize this issue in the backlog.

Thankfully, @lgfa29 is also an expert web developer and can provide insights if I'm wrong here.

ui/app/controllers/evaluations/index.js
ChaiWithJai: That's your call to make. We can totally adopt that pattern, but that's an undocumented behavior we'd have to load into our mental models. We specifically designed this to set namespace to `*` when entering this view. I could see the mental baggage of keeping this type of knowledge in our heads with an Ember Routing API change.

It's good UX to have. So totally your call, I'll support whatever you decide. I'll update my review on this PR to approved and let you handle this how you think is best.

ChaiWithJai (approved):

lgfa29:
> If we need to make any edits client-side, I would suspect we'd need to check the formatting logic in generateFilterExpression. However, the [tests](https://github.com/hashicorp/nomad/pull/13530/tests/acceptance/evaluations-test.js) assert on how user actions impact the request we're sending and the default request we send when we enter the view. I'd love to learn if we wrote those tests and the formatting logic incorrectly.
> 
> I highly recommend reading that filtering logic and the tests to see if we got something wrong. But, I believe @mikenomitch pointed the real issue out [12918](https://github.com/hashicorp/nomad/issues/12918) and we should ask @lgfa29 to prioritize this issue in the backlog.
> 
> Thankfully, @lgfa29 is also an expert web developer and can provide insights if I'm wrong here.

#12918 could have contributed to this error if the UI requests a specific eval using `*` as the namespace, but I'm not sure if it does. If you still run into problems you can rebase onto https://github.com/hashicorp/nomad/pull/13551 or wait for it be merged and rebase `main`.

tgross (approved):
LGTM. I left some comments on the API bits in https://github.com/hashicorp/nomad/pull/13552 but nothing blocking.

lgfa29 (approved):

PR #13533, opened 2022-06-30
ui:  glob matching algorithm

ChaiWithJai:
This PR creates a private method on the Variable Ability class to mirror the special glob matching algorithm used by the Nomad API (a standalone [Go module](https://github.com/ryanuber/go-glob)).

This PR is the second to last step in the UI being able to match the way the Nomad API computes a path's capabilities.


tgross (approved):
LGTM! Nice work!

PR #13534, opened 2022-06-30
docs: fix Plan{,Result}.NodeUpdate comment

schmichael:
It appears way back when this was first implemented in
9a917281af9c0a97a6c59575eaa52c5c86ffc60d, it was renamed from
NodeEvict (with a correct comment) to NodeUpdate. The comment was
changed from referring to only evictions to referring to "all allocs" in
the first sentence and "stop or evict" in the second.

This confuses every time I see it because I read the name (NodeUpdate)
and first sentence ("all the allocs") and assume this represents *all*
allocations... which isn't true.

I'm going to assume I'm the only one who doesn't read the 2nd sentence
and that's why this suboptimal wording has lasted 7 years, but can we
change it for my sake?

### DIFF
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -11145,8 +11145,8 @@ type Plan struct {
 	// of the plan by only including it once.
 	Job *Job
 
-	// NodeUpdate contains all the allocations for each node. For each node,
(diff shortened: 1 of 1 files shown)


tgross (approved):
LGTM!

PR #13545, opened 2022-07-01
ui:  add can edit to variable capabilities

ChaiWithJai:
This PR creates a `can-edit` property in the Variable Ability class and adds an `ember-can` check in the `variables.variable.index` template to access the `Edit` button.

### DIFF
--- a/ui/tests/unit/abilities/variable-test.js
+++ b/ui/tests/unit/abilities/variable-test.js
@@ -202,6 +202,107 @@ module('Unit | Ability | variable', function (hooks) {
     });
   });
 
+  module('#edit', function () {
+    test('it does not permit editing variables by default', function (assert) {
+      const mockToken = Service.extend({
+        aclEnabled: true,
+      });
+
+      this.owner.register('service:token', mockToken);
+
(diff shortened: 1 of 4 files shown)


ui/tests/unit/abilities/variable-test.js
```diff
@@ -202,6 +202,107 @@ module('Unit | Ability | variable', function (hooks) {
...
+                      Capabilities: ['list'],
+                    },
+                  },
+                },
+                {
+                  Name: 'pablo',
+                  Capabilities: [],
+                  SecureVariables: {
+                    'Path "foo/bar"': {
+                      Capabilities: ['edit'],
```
philrenaud, line 288: I'm not sure that "edit" is a capability that we support here; see [the list of policy capabilities](https://github.com/hashicorp/nomad/pull/13000/files#diff-f8c1a4d6b1e8929cdfaad3339b0da9fd941befa72da499b6bbd326c32e3f9d3fR74-R77) we support here.

I think if this is changed to "write", the code here is otherwise in great shape to support editing.

PR #13547, opened 2022-07-02
docs: Add 'serial' attribute to usb driver

CarbonCollins:
:wave: Ive published a new version of the USB driver which adds a new fingerprinted attribute. This PR is to update the documentation to reflect this change


CarbonCollins:
The driver release in question: https://gitlab.com/CarbonCollins/nomad-usb-device-plugin/-/releases/0.3.0

PR #13552, opened 2022-07-04
Backport #13530 to 1.2.x

lgfa29:
Backport the API portion of #13530 to `release/1.2.x`. Wildcard namespace list support for the eval endpoint was added in 1.2.4, so this doesn't have to be backported to 1.1.x.

### DIFF
--- a/nomad/eval_endpoint.go
+++ b/nomad/eval_endpoint.go
@@ -389,11 +389,28 @@ func (e *Eval) List(args *structs.EvalListRequest,
 	}
 	defer metrics.MeasureSince([]string{"nomad", "eval", "list"}, time.Now())
 
+	namespace := args.RequestNamespace()
+	var allow func(string) bool
+
 	// Check for read-job permissions
-	if aclObj, err := e.srv.ResolveToken(args.AuthToken); err != nil {
+	aclObj, err := e.srv.ResolveToken(args.AuthToken)
+	switch {
+	case err != nil:
 		return err
-	} else if aclObj != nil && !aclObj.AllowNsOp(args.RequestNamespace(), acl.NamespaceCapabilityReadJob) {
+	case aclObj == nil:
+		allow = func(string) bool {
+			return true
+		}
+	case namespace == structs.AllNamespacesSentinel:
+		allow = func(ns string) bool {
+			return aclObj.AllowNsOp(ns, acl.NamespaceCapabilityReadJob)
+		}
+	case !aclObj.AllowNsOp(namespace, acl.NamespaceCapabilityReadJob):
 		return structs.ErrPermissionDenied
+	default:
+		allow = func(string) bool {
+			return true
+		}
 	}
 
 	// Setup the blocking query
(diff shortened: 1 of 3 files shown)


nomad/eval_endpoint.go
```diff
@@ -404,34 +421,48 @@ func (e *Eval) List(args *structs.EvalListRequest,
...
-			var evals []*structs.Evaluation
-			paginator := state.NewPaginator(iter, args.QueryOptions,
-				func(raw interface{}) {
-					eval := raw.(*structs.Evaluation)
-					evals = append(evals, eval)
+
+				iter = memdb.NewFilterIterator(iter, func(raw interface{}) bool {
+					if eval := raw.(*structs.Evaluation); eval != nil {
+						nsAllowed := allowableNamespaces == nil ||
+							allowableNamespaces[eval.Namespace]
```
lgfa29, line 448: In 1.3.x this check is done using the [paginator.NamespaceFilter](https://github.com/hashicorp/nomad/blob/v1.3.1/nomad/state/paginator/filter.go#L20-L41), which doesn't exist in 1.2.x.

Instead of modifying `args.ShouldBeFiltered` I thought it would be better to make the verification here to keep `ShouldBeFiltered` the same in `release/1.2.x` and `release/1.3.x`.

nomad/eval_endpoint.go
```diff
@@ -404,34 +421,48 @@ func (e *Eval) List(args *structs.EvalListRequest,
...
-				iter, err = store.EvalsByNamespace(ws, args.RequestNamespace())
-			}
-			if err != nil {
-				return err
-			}
 
-			iter = memdb.NewFilterIterator(iter, func(raw interface{}) bool {
-				if eval := raw.(*structs.Evaluation); eval != nil {
-					return args.ShouldBeFiltered(eval)
+			// check if user has permission to all namespaces
```
tgross, line 425: Nitpick: I think this comment is being copied from elsewhere, but "has permission to all namespaces" seems misleading when what we're really getting is the set of namespace we have permission to.

nomad/eval_endpoint.go
```diff
@@ -389,11 +389,28 @@ func (e *Eval) List(args *structs.EvalListRequest,
...
+	case namespace == structs.AllNamespacesSentinel:
+		allow = func(ns string) bool {
+			return aclObj.AllowNsOp(ns, acl.NamespaceCapabilityReadJob)
+		}
+	case !aclObj.AllowNsOp(namespace, acl.NamespaceCapabilityReadJob):
 		return structs.ErrPermissionDenied
+	default:
+		allow = func(string) bool {
+			return true
+		}
```
tgross, line 413: This seems like an example of "the right code looks wrong", because this means we've already passed the ACL check so we don't need to filter below. Maybe leave a comment for it?

```suggestion
		allow = func(string) bool {
			return true // we've already passed the only ACL check
		}
```

tgross (approved):
LGTM

shoenig (approved):
LGTM

nomad/eval_endpoint.go
lgfa29: In rewriting this I ended up doing a larger refactoring to push most of this logic into the `acl` package. 

https://github.com/hashicorp/nomad/pull/13606 is the main refactoring.
https://github.com/hashicorp/nomad/pull/13608 applies the changes to the existing job and alloc list endpoints.
https://github.com/hashicorp/nomad/pull/13530 applies the changes to the evals list endpoint to fix a bug.

I will redo this eval list bug backport once https://github.com/hashicorp/nomad/pull/13606 is merged into `releases/1.2.x`.

PR #13580, opened 2022-07-04
metrics: add node_ip label to host metrics

### LINKED ISSUE #12921: Add IP Address to host metrics
### Proposal

It would be nice to have the client IP Address in any [host metrics](https://www.nomadproject.io/docs/operations/metrics-reference#host-metrics) emitted for that client.

### Use-cases

Quickly jumping to specific nodes given a specific log line.

lgfa29:
Closes #12921


client/client.go
```diff
@@ -771,6 +771,12 @@ func (c *Client) NodeID() string {
 	return c.config.Node.ID
 }
 
+// NodeIP returns the IP address the node is listening
+func (c *Client) NodeIP() string {
+	addr, _, _ := net.SplitHostPort(c.config.Node.HTTPAddr)
```
tgross, line 776: For clients, the RPC address would be the address seen by the servers that shows up in `node status` outputs and the like. Maybe we should use that instead of the HTTP address, which is only going to be used by external consumers?

Either way, should we call this method and the metrics label for the specific IP (RPC or HTTP) we're using, so that if we come back later to add another IP we can do so?

lgfa29: Sounds good. I will use the RPC address and rename the method (`NodeRPCIP`) and metric `node_rpc_ip`. Thanks!

client/client.go
```diff
@@ -771,6 +771,11 @@ func (c *Client) NodeID() string {
 	return c.config.Node.ID
 }
 
+// NodeIP returns the IP address the node is listening
```
tgross, line 774: Oh, right... this comment points out a problem with trying to expose the address of the client... the client doesn't bind on the RPC address (which is why you had to thread it down from the agent config). So this might be the right IP address from the configuration but isn't necessarily meaningful to the operator if they've got multiple interfaces on the host. 

I suspect a lot of clients are just going to show 127.0.0.1 here, which isn't terribly useful as a metric. But I think that's _also_ going to be a problem for the HTTP address. Before I send you off on another wild goose chase here, maybe we should ping @mikenomitch and ask for what he was originally thinking with https://github.com/hashicorp/nomad/issues/12921?

lgfa29: Sounds good, I will follow-up on that, thanks for the guidance.

Another option that I thought about was to use to first fingerprinted IP, which also what's set in `unique.network.ip-address`, but I don't know if that's a good value either. 

I will leave this open until we get more clarification 👍

lgfa29:
Marking this as a draft for now since we don't have a good source for client IP address.

PR #13581, opened 2022-07-05
cli: improve output of eval commands

lgfa29:
Use the same output format when listing multiple evals in the `eval
list` command and when `eval status <prefix>` matches more than one
eval.

Include the eval namespace in all output formats and always include the
job ID in `eval status` since, even `node-update` evals are related to a
job.

Add Node ID to the evals table output to help differentiate
`node-update` evals.

### DIFF
--- a/.changelog/13581.txt
+++ b/.changelog/13581.txt
@@ -1,0 +1,7 @@
+```release-note:improvement
+cli: display namespace and node ID in the `eval list` command and when `eval status` matches multiple evals
+```
+
+```release-note:improvement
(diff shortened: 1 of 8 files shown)


jrasell (approved):
LGTM!

I pushed two additional commits to this PR which should be checked before merging. The first was merging main into the branch in order to fix a merge conflict. The second was to move `eval delete` to use this listing helper and update the docs for this accordingly.

lgfa29:
The changes look good, thanks!

PR #13589, opened 2022-07-05
Fix Testutil for delve debugging API tests

angrycub:
When running an API test in a debugger, the tests immediately "PASS" rather than run. This is because the discover code attempts to run the debug test executable _**as**_ the Nomad binary, which errors and then ends in the Skip pathway.


tgross (approved):
LGTM 👍

PR #13596, opened 2022-07-05
refact:  update token factory

ChaiWithJai:
This PR updates the Token factory in our Ember Mirage set-up to update with `rules` and `rulesJSON` properties to reflect the API changes made.

### DIFF
--- a/ui/mirage/factories/token.js
+++ b/ui/mirage/factories/token.js
@@ -38,8 +38,19 @@ namespace "default" {
   policy = "read"
   capabilities = ["list-jobs", "alloc-exec", "read-logs"]
   secure_variables {
+    # full access to secrets in all project paths
+    path "project/*" {
+      capabilities = ["write", "read", "destroy", "list"]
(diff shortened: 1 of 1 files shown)


ui/mirage/factories/token.js
```diff
@@ -38,8 +38,19 @@ namespace "default" {
...
   secure_variables {
+    # full access to secrets in all project paths
+    path "project/*" {
+      capabilities = ["write", "read", "destroy", "list"]
+    }
+
+    # full access to secrets in all project paths
     path "*" {
-      capabilities = ["list"]
+      capabilities = ["write", "read", "destroy", "list"]
```
philrenaud, line 48: thought: maybe we should leave the `*` capabilities for our demo token to just `read` and `list`? That way testers can experience what a limited experience looks like (and anything under `project/*` could be used to demonstrate higher-capability stuff like editing)

philrenaud (approved):

PR #13606, opened 2022-07-06
api: refactor ACL check for namespace wildcard

lgfa29:
Improve how the all namespaces wildcard (`*`) is handled when checking
ACL permissions. When using the wildcard namespace the `AllowNsOp` would
return false since it looks for a namespace called `*` to match.

This commit changes this behavior to return `true` when the queried
namespace is `*` and the token allows the operation in _any_ namespace.

Actual permission must be checked per object. The helper function
`AllowNsOpFunc` returns a function that can be used to make this
verification.


nomad/alloc_endpoint_test.go
```diff
@@ -1284,8 +1284,9 @@ func TestAllocEndpoint_List_AllNamespaces_ACL_OSS(t *testing.T) {
 		{
 			Label:     "all namespaces with insufficient token",
 			Namespace: "*",
-			Allocs:    []*structs.Allocation{},
 			Token:     ns1tokenInsufficient.SecretID,
+			Error:     true,
+			Message:   structs.ErrPermissionDenied.Error(),
```
lgfa29, line 1289: This was the only test failure after this refactoring, but I think this is the right answer? The token used doesn't have the required `read-job` permission, so it should return a `403` rather than an empty list.

acl/acl.go
```diff
@@ -307,6 +340,34 @@ func (a *ACL) matchingNamespaceCapabilitySet(ns string) (capabilitySet, bool) {
 	return a.findClosestMatchingGlob(a.wildcardNamespaces, ns)
 }
 
+// anyNamespaceAllows returns true if any namespace in the ACL object allows
+// the given operation.
+// If op is an empty string it returns true if any namespace allows any
+// operation.
```
tgross, line 346: It feels like passing an empty op is a programmer error, not a condition we intentionally support. Should this return `ErrPermissionDenied` so that we can catch that in testing?

.changelog/13606.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+api: refactor ACL check when using the all namespaces wildcard
+```
```
tgross, line 3: If the external behavior doesn't change, we shouldn't provide a changelog. If the external behavior changes, we should describe that change here instead.

acl/acl.go
lgfa29: Yeah, I didn't like this very much but figured it was OK given it's a private method. The key is that it needs to differentiate between two scenarios:

- `AllowNamespaceOperation` needs to know if any namespace allows a specific operation
- `AllowNamespace` needs to know if any namespace allows any operation (i.e., all don't deny)

Having a "magic" input (empty string in this case) serves as that switch between the two cases.

But looking at the code I found a better way to do it, so thanks for raising it!

tgross (approved):
LGTM 👍

acl/acl.go
```diff
@@ -307,6 +340,42 @@ func (a *ACL) matchingNamespaceCapabilitySet(ns string) (capabilitySet, bool) {
...
+	})
+}
+
+// anyNamespaceAllows returns true if the callback cb returns true for any
+// namespace operation of the ACL object.
+func (a *ACL) anyNamespaceAllows(cb func(capabilitySet) bool) bool {
+	allow := false
+
+	checkFn := func(_ []byte, iv interface{}) bool {
+		v := iv.(capabilitySet)
```
shoenig, line 365: do we not need to check this is okay? if we can statically assert, shouldn't the checkFn parameter be a `capabilitySet`?

lgfa29: `checkFn` needs to be a [`WalkFn`](https://pkg.go.dev/github.com/hashicorp/go-immutable-radix#WalkFn), so the signature is pretty set.

This iterates over the internal tree of namespaces, if there's data that is not a `capabilitySet` then something pretty wrong happened 😅 

[`findAllMatchingWildcards`](https://github.com/hashicorp/nomad/blob/v1.3.1/acl/acl.go#L361-L363) doesn't have the check either, so I think it's safe to assume this would always work?

shoenig: Ahh I see, yup just ignore me!

shoenig (approved):

PR #13608, opened 2022-07-06
api: apply new ACL check for wildcard namespace

lgfa29:
In #13606 the ACL check was refactored to better support the all
namespaces wildcard (`*`). This commit applies the changes to the jobs
and alloc list endpoints.

### DIFF
--- a/nomad/alloc_endpoint_test.go
+++ b/nomad/alloc_endpoint_test.go
@@ -1284,8 +1284,9 @@ func TestAllocEndpoint_List_AllNamespaces_ACL_OSS(t *testing.T) {
 		{
 			Label:     "all namespaces with insufficient token",
 			Namespace: "*",
-			Allocs:    []*structs.Allocation{},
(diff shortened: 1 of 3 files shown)


nomad/alloc_endpoint_test.go
```diff
@@ -1284,8 +1284,9 @@ func TestAllocEndpoint_List_AllNamespaces_ACL_OSS(t *testing.T) {
 		{
 			Label:     "all namespaces with insufficient token",
 			Namespace: "*",
-			Allocs:    []*structs.Allocation{},
 			Token:     ns1tokenInsufficient.SecretID,
+			Error:     true,
+			Message:   structs.ErrPermissionDenied.Error(),
```
lgfa29, line 1289: This was the only test the failed after the refactoring, but I think returning a `403` is the correct result in this case? The test token doesn't have the `read-job` capability.

tgross (approved):
LGTM

shoenig (approved):

PR #13618, opened 2022-07-06
ui:  add namespace filter to `variables.index`

ChaiWithJai:
This PR conditionally adds a namespace filter to `variables.index` based on the number of namespaces a user has access to.

Side Effect:  This PR cleans up improper template variables in `variables.index` and removes the `SearchBox` because there is no associated functionality or state required to allow the feature to work. Handlebars is failure silent, meaning if you call a variable that doesn't exist (e.g. `this.foo`), then handlebars will just not render that markup without erroring.

Figma Mock:
image

Actual Result:
image


philrenaud (changes requested):
Looks like it's off to a good start; missing a few things:
- Some variables from the mirage factory set to have different namespaces, for easier smoke-testing in mirage
- Can't seem to get the filter to work on the table at root/index:
image
- Filter needs to also be added to the /path route (for when you're in a folder)

Because of this, I might consider moving the filter logic up to the variables controller itself, rather than the index. That way paths and index alike can inherit from it.

ui/app/components/variable-paths.hbs
```diff
@@ -31,6 +31,7 @@
           <LinkTo
             @route="variables.variable"
             @model={{file.absoluteFilePath}}
+            @query={{hash namespace="*"}}
```
ChaiWithJai, line 34: Since we're in the same route (and the parent route) is still activated, `LinkTo` will automatically forward the current query parameter. To override this behavior, we need to explicitly reset this parameter back to its original value `*`.

ui/app/components/variable-paths.hbs
```diff
@@ -16,7 +16,7 @@
         <td colspan="3">
           <span>
             <FlightIcon @name="folder" />
-            <LinkTo @route="variables.path" @model={{folder.data.absolutePath}}>
+            <LinkTo @route="variables.path" @model={{folder.data.absolutePath}} @query={{hash namespace="*"}}>
```
ChaiWithJai, line 19: Since we're in the same route (and the parent route) is still activated, `LinkTo` will automatically forward the current query parameter. To override this behavior, we need to explicitly reset this parameter back to its original value `*`.

ui/app/routes/variables.js
```diff
@@ -9,14 +9,23 @@ export default class VariablesRoute extends Route.extend(WithForbiddenState) {
...
     }
   }
-  async model() {
+
+  async model({ qpNamespace }) {
+    const namespace = qpNamespace ?? '*';
     try {
-      const variables = await this.store.findAll('variable');
+      await this.store.findAll('namespace');
+      const variables = await this.store.query('variable', { namespace });
```
ChaiWithJai, line 28: This was the uncommitted code that I was referring to.

ui/app/adapters/variable.js
```diff
@@ -21,6 +21,12 @@ export default class VariableAdapter extends ApplicationAdapter {
     let baseUrl = this.buildURL(modelName);
     return pluralize(baseUrl);
   }
+
+  urlForQuery(_query, modelName) {
+    let baseUrl = this.buildURL(modelName);
+    return pluralize(baseUrl);
+  }
```
philrenaud, line 28: Is this method used anywhere?

It seems like perhaps this needs to be implemented further; as it stands, if I save a variable on namespace A, my request for that variable immediately upon creation fails. In practice, this throws a 404 immediately upon creating a new variable with a namespace on this branch. 
image

ui/app/components/secure-variable-form.js
```diff
@@ -137,6 +149,15 @@ export default class SecureVariableFormComponent extends Component {
         this.keyValues = nonEmptyItems;
       }
 
+      if (this.args.model?.isNew) {
+        if (this.namespaceOptions) {
+          this.args.model.set('namespace', this.variableNamespace);
+        } else {
+          const [namespace] = this.store.peekAll('namespace');
```
philrenaud, line 156: I don't think this works; `store` is not defined in this file / injected as a service, and I get the following on save:

image

philrenaud (changes requested):
Ran into a couple problems in testing, and a few stylistic issues as well. The save -> redirect logic is the most significant hurdle here IMO.

ui/app/components/secure-variable-form/namespace-filter.js
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  @action
+  formatAndSetNamespaces() {
+    // Triggered on the promise in fetchNamespaces resolving
+    const namespaces = this.store
+      .peekAll('namespace')
+      .map(({ name }) => ({ key: name, label: name }));
+
+    if (namespaces.length <= 1) return null;
```
philrenaud, line 22: Just want to check on this: are we sure the logic here is okay?

What if I have access to a single namespace (`my-namespace`), but not `*` / `default` ?

Will this mean that, when trying to create a new variable, it doesn't automatically set my namespace to `my-namespace`, and instead falls back to `*`, which would fail on POST?

ui/app/components/secure-variable-form.hbs
```diff
@@ -39,17 +39,31 @@
...
+      @data={{hash 
+        disabled=(not @model.isNew)
+        selection=this.variableNamespace
+        namespaceOptions=this.namespaceOptions
+      }}
+      @fns={{hash
+        onSelect=this.setNamespace
+        setNamespaceOptions=this.setNamespaceOptions
+      }}
+    />
```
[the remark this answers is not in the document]
ChaiWithJai: Should be resolved now.

ui/app/components/secure-variable-form.js
ChaiWithJai: We were missing store injection. Should be resolved now.

ui/app/components/secure-variable-form/namespace-filter.js
ChaiWithJai: `formatAndSetNamespaces` is for generating the namespaces that will appear in the dropdown list. First, we peek in the store to grab a list of the namespaces. So, if you only have access to `my-namespace` then we would presume that the API returns only `my-namespace`.

The `/namespaces` endpoint is just going to return a list of namespaces -- we're not responsible for knowing if the API is returning results that the user should not have access to.

Side Note:  A user should not be able to set `*` as a namespace. We use `*` as a wildcard for searching and filtering namespaces in our `get` requests.

tgross: Right! The `Namespace.List` RPC that populates that response returns only namespaces you have access to (ref  [`namespace_endpoint.go#L252-L255`](https://github.com/hashicorp/nomad/blob/v1.3.2/nomad/namespace_endpoint.go#L252-L255)), even if you pass `*` wildcard in the request.

ui/app/adapters/variable.js
ChaiWithJai: `urlForQuery` is invoked when we use the `query` method on the `store` service.

ChaiWithJai: Specifically, we're handling how we structure the resource in the request URL.

Sometimes we must be plural `/v1/vars/` and other times we must be singular `/v1/var/:resourceId`.

ChaiWithJai: This behavior is correct here. However, it might have a collision with [13627](https://github.com/hashicorp/nomad/pull/13627).

ui/app/components/secure-variable-form.hbs
philrenaud, line 52: Noticing a style / positioning error when receiving a "path already exists" error
image

ui/app/components/secure-variable-form.js
```diff
@@ -31,6 +32,23 @@ export default class SecureVariableFormComponent extends Component {
    * @type {DuplicatePathWarning}
    */
   @tracked duplicatePathWarning = null;
```
philrenaud, line 34: There is a `validatePath()` action that should be updated here (I think! @tgross  / @angrycub please correct me if this is an incorrect assumption): If I have a variable at `my-vars/a` on the `default` namespace, but try to create a new one at `my-vars/a` on the `my-other-namespace` namespace, we currently throw a duplicate error. I think this is probably supposed to be allowed, as confusing as the circumstance might be to the end-user.

ui/app/components/secure-variable-form.js
```diff
@@ -31,6 +32,23 @@ export default class SecureVariableFormComponent extends Component {
...
+  @tracked variableNamespace = null;
+  @tracked namespaceOptions = null;
+
+  @action
+  setNamespace(namespace) {
+    this.variableNamespace = namespace;
+  }
```
philrenaud, line 41: Suggestion: this should inherit from the `?namespace=...` query param that gets set by selecting a namespace on the path/index page before creation. That query param persists to the /new page, so we can use it to help save the user a click.

ui/app/adapters/variable.js
philrenaud: This seems totally resolved now! 🎉

philrenaud (approved):
Looks good and works well! I've noted a styling bug, but otherwise any issues can be handled separately; I've added 2 of them: https://github.com/hashicorp/nomad/issues/13864 (which seems kind of fringe and tricky to deal with) + https://github.com/hashicorp/nomad/issues/13863 (which seems straightforward enough)

Great work!

ui/app/components/secure-variable-form.js
ChaiWithJai: Yes, good point. But, this action is about setting a property on a form, this isn't being used to filter, sort and be passed to a request and its really about form state so I wouldn't have it be a query parameter because then it would imply one of the cases above.

The query param shouldn't persist to the `/new` route.

ui/app/components/secure-variable-form.hbs
ChaiWithJai: The issue here is that, we started this work before this change was merged in and now because we're expecting a new child, we need to rethink the design of the whole grid.

I also believe that the `relative` position attribute will need to be rethought for this.

PR #13620, opened 2022-07-06
Handling createTime and modifyTime through serialization now that they contain nanoseconds

philrenaud:
Otherwise, we'd get an unmarshall error on save, and/or an invalid date on tables in the UI.

### DIFF
--- a/ui/app/models/variable.js
+++ b/ui/app/models/variable.js
@@ -48,10 +48,10 @@ export default class VariableModel extends Model {
   @attr('number') createIndex;
   /** @type {number} */
   @attr('number') modifyIndex;
-  /** @type {string} */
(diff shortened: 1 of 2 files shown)


ChaiWithJai (approved):
LGTM. 

`suggestion:` Whenever, I'm trying assert that the request I'm sending to the server is formatted correctly, I'll write a test case like this (below) to ensure this doesn't turn into a bug later on:

```
server.get('/someEndPoint', function (_server, fakeRequest) {
        assert.deepEqual(
          fakeRequest.queryParams,
          {
            someProperty: 'someValue',
          },
          'It makes another server request using the options selected by the user'
        );
        return [];
      });
```

Issue #12273, opened 2022-03-14
Nomad API "Unexpected EOF" Errors


evandam:
### Nomad version
`Nomad v1.2.6 (a6c6b475db5073e33885377b4a5c733e1161020c)`

### Operating system and Environment details
Ubuntu 18.04

### Issue
We sporadically see API calls on Nomad clients start to fail with "Unexpected EOF" errors.

### Reproduction steps
Unclear, seemingly random on Nomad clients.

#### Expected Result
Nomad API handles requests correctly

#### Actual Result
```bash
$ nomad server members
Error querying servers: Get "http://127.0.0.1:4646/v1/agent/members": EOF
$ curl localhost:4646/v1/status/leader
curl: (56) Recv failure: Connection reset by peer
```

### Job file (if appropriate)
N/A

### Nomad Server logs (if appropriate)

There's nothing that jumps out in server logs, but here was the last hour before an instance of this error. Note that the "permission denied" errors can be safely ignored, it's just checks we have to verify ACL permissions in place.

https://gist.github.com/evandam/1d1f2dd2a5032426736107f414959e2d

### Nomad Client logs (if appropriate)
N/A

I know it's not much to go on, but let me know if there's any additional detail I can provide.

DerekStrickland:
Hi @evandam  

I'm sorry you are having issues. I'll take a look a the log and see if anything stands out. I'm curious about something.  When this happens have you ever looked at the servers/clients to see how they are doing at the OS level? Specifically, I am wondering if they are experiencing any memory, CPU, or network pressure.

DerekStrickland:
Also, do you have any client logs showing the error, or are you only experiencing it from the command line?

evandam:
Hey @DerekStrickland, adding a couple screenshots of our monitoring - doesn't look like anything unusual in terms of system activity. There was a small spike from ~20% to ~45% CPU, but I believe that was a minute or so after the first EOF error so probably not related. Memory and network pressure seems fine too.

There aren't any client logs, we're really just seeing it on the command line/curl. For what it's worth, I've seen this happening on our Nomad client nodes but never on a server (could be coincidence, could be related to different system loads, etc.) but maybe worth pointing out.

![Screen Shot 2022-03-14 at 10 23 30 AM](https://user-images.githubusercontent.com/3985383/158226985-f71d63f2-e467-4a0d-b888-867032eccca1.png)
[...]

Issue #8718, opened 2020-08-23
Connection reset by peer when querying /v1/agent/self endpoint


scalp42:
Hi folks,

We're seeing errors every day when trying to query the `/v1/agent/self` endpoint.

<details>
  <summary markdown="span">Here's an example on the actual clients:</summary>

root@nomad-compute-i-0316e38d5be06a803 [dev-usw2-dev1] ~ # date ; curl localhost:4646/v1/agent/self
Sun Aug 23 21:02:19 UTC 2020
curl: (56) Recv failure: Connection reset by peer
root@nomad-compute-i-0316e38d5be06a803 [dev-usw2-dev1] ~ # date ; curl localhost:4646/v1/agent/self
Sun Aug 23 21:02:19 UTC 2020
curl: (56) Recv failure: Connection reset by peer
root@nomad-compute-i-0316e38d5be06a803 [dev-usw2-dev1] ~ # date ; curl localhost:4646/v1/agent/self
Sun Aug 23 21:02:20 UTC 2020
curl: (56) Recv failure: Connection reset by peer
root@nomad-compute-i-0316e38d5be06a803 [dev-usw2-dev1] ~ # date ; curl localhost:4646/v1/agent/self
Sun Aug 23 21:02:20 UTC 2020
curl: (56) Recv failure: Connection reset by peer
root@nomad-compute-i-0316e38d5be06a803 [dev-usw2-dev1] ~ # date ; curl localhost:4646/v1/agent/self
Sun Aug 23 21:02:21 UTC 2020
[...]

PR #13480, opened 2022-06-24
implement RPC rate limiting

tgross:
early proof-of-concept for providing pre-auth rate limiting at the Nomad
servers, with the ability to observe which users are creating the load.

_This is very early draft to assist internal design discussions and doesn't have any configuration done. It's only wired up to a single RPC endpoint at the moment. It'll also need to implement support for workload identities once #13000 has been merged to `main`._

PR #9608, opened 2020-12-10
Use go-connlimit to ratelimit with 429 responses

shoenig:
Expands on https://github.com/hashicorp/nomad/pull/9359 from @benbuzbee 

In addition this PR fixes the original tests, which produced the correct result for the wrong reason. In TLS mode, the clients were creating HTTP connections which would be shunted by the server in the same way being ratelimited would (i.e. `io.EOF`). Instead, use a tls `net.Conn` for tls tests.

Also rebased the original commit to remove extraneous file. 

Closes #9359

PR #13621, opened 2022-07-06
Return 429 response on HTTP max connection limit

wjordan:
The `limits.http_max_conns_per_client` configuration option sets a limit on the max number of concurrent TCP connections per client IP address:

https://github.com/hashicorp/nomad/blob/0b3ba5c77cc3ecc62bd380bcd2d52d272cf49c8c/website/content/docs/configuration/index.mdx#L197-L200

The current implementation behavior silently closes the TCP connection, resulting in opaque 'unexpected EOF' or 'connection reset' errors which many users fail to correctly diagnose, and is likely the root cause of various reported Nomad issues.

Instead of silently closing the connection, this PR returns a `429 Too Many Requests` HTTP response with a helpful error message to aid debugging when the connection limit is unintentionally reached. This should help diagnose/fix some instances of #12273 and #8718, where the fix is simply to increase the connection limit from the default of 100.

The default of 100 is easily hit when doing anything with blocking queries, since each blocking query requires a separate HTTP client connection. One example from my use-case: the Nomad Autoscaler monitors each job status using a [blocking query](https://github.com/hashicorp/nomad-autoscaler/blob/562c3192c10c8361d6dd6d5db152feaad8e25f07/plugins/builtin/target/nomad/plugin/state.go#L158-L162), which means that any autoscaler agent monitoring more than 100 jobs will exceed this default limit in normal operation and will need to be increased.

One tradeoff in this PR is that the connection is kept open slightly longer as the agent writes out the 429 response. The `HTTPConnStateFuncWithDefault429Handler` function accepts a write-deadline argument, which I've set as ~0 (no deadline)~ (updated: `10 * time.millisecond` to match [existing behavior](https://github.com/hashicorp/consul/pull/8221) in Consul) in this PR, this could alternatively be made configurable through an additional config option, if desired.

### DIFF
--- a/.changelog/13621.txt
+++ b/.changelog/13621.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+api: HTTP server now returns a 429 error code when hitting the connection limit
+```
--- a/command/agent/http.go
+++ b/command/agent/http.go
@@ -216,7 +216,7 @@ func makeConnState(isTLS bool, handshakeTimeout time.Duration, connLimit int) fu
 			// Still return the connection limiter
 			return connlimit.NewLimiter(connlimit.Config{
 				MaxConnsPerClientIP: connLimit,
-			}).HTTPConnStateFunc()
+			}).HTTPConnStateFuncWithDefault429Handler(0)
 		}
 		return nil
 	}
@@ -226,7 +226,7 @@ func makeConnState(isTLS bool, handshakeTimeout time.Duration, connLimit int) fu
 		// handshake timeout.
 		connLimiter := connlimit.NewLimiter(connlimit.Config{
 			MaxConnsPerClientIP: connLimit,
-		}).HTTPConnStateFunc()
+		}).HTTPConnStateFuncWithDefault429Handler(0)
 
 		return func(conn net.Conn, state http.ConnState) {
 			switch state {
--- a/command/agent/http.go
+++ b/command/agent/http.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/armon/go-metrics"
 	"golang.org/x/time/rate"
 	"net"
 	"net/http"
@@ -278,6 +279,7 @@ func connLimiter(connLimit int, logger log.Logger) func(conn net.Conn, state htt
 		MaxConnsPerClientIP: connLimit,
 	}).HTTPConnStateFuncWithErrorHandler(func(err error, conn net.Conn) {
 		if err == connlimit.ErrPerClientIPLimitReached {
+			metrics.IncrCounter([]string{"nomad", "agent", "http", "exceeded"}, 1)
 			if n := limiter.Reserve(); n.Delay() == 0 {
 				logger.Warn("Too many concurrent connections", "address", conn.RemoteAddr().String(), "limit", connLimit)
 				conn.SetDeadline(time.Now().Add(10 * time.Millisecond))
--- a/command/agent/http.go
+++ b/command/agent/http.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"golang.org/x/time/rate"
 	"net"
 	"net/http"
 	"net/http/pprof"
@@ -153,7 +154,7 @@ func NewHTTPServers(agent *Agent, config *Config) ([]*HTTPServer, error) {
 		httpServer := http.Server{
 			Addr:      srv.Addr,
 			Handler:   handlers.CompressHandler(srv.mux),
-			ConnState: makeConnState(config.TLSConfig.EnableHTTP, handshakeTimeout, maxConns),
+			ConnState: makeConnState(config.TLSConfig.EnableHTTP, handshakeTimeout, maxConns, srv.logger),
 			ErrorLog:  newHTTPServerLogger(srv.logger),
 		}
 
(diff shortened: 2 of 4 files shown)


wjordan:
I found that Consul returns a 429 response on max-connection limit (hashicorp/consul#8221) and uses `10 * time.Millisecond` for its write timeout, so I added a commit with a matching 10ms timeout for consistency.

tgross (commented):
Hi @wjordan! This PR effectively reverts a PR made by my colleague @shoenig in https://github.com/hashicorp/nomad/commit/f0dff3fadab35e2bd692f52eb75700fdce151b52, where we weren't happy with how the default 429 handler interacted with our handshake.

I'm working on rate limiting in https://github.com/hashicorp/nomad/pull/13480 which will send a 429 code as you'd expect. I can make sure we're introducing that where we can for connection tracking as well.

I'm going to close this PR, but thanks so much for contributing!

wjordan:
Thanks for pointing me to the previous discussion. The concern in https://github.com/hashicorp/nomad/pull/9608#pullrequestreview-549435793 seems to be that writing 429 responses to the `net.Conn` seems like a DoS, since the write calls could block and prevent the main thread from handling connections from other addresses.

It seems the DoS concern is mostly theoretical (I was unable to reproduce any DoS in my own local testing), and I think it's based on some unclear assumptions about what might cause a `conn.Write` call to block- I believe each TCP socket has its own send buffer which should be empty for any `StateNew` connection, so blocking writes should never typically happen. In any case, either reducing the timeout to `1 * time.microsecond` and/or wrapping the write calls in their own goroutines could be a reasonable mitigation if this is a remaining concern. (I don't see how #13480 is related, as that seems focused on rate-limiting API calls and not TCP-connection rates, at least in its current form.)

Beyond this, rejecting this PR in Nomad raises an issue about Consul's current behavior- if this behavior was considered a DoS issue in Nomad, does that mean that it's currently a DoS issue for Consul? At the very least I'd like to see the various Hashicorp products agree on _some_ consistent behavior across them, whatever it may be.

tgross:
Hey @wjordan I suspect you're right that blocking sends aren't an issue, but I'm also not sure that's actually the problem.

If we simply drop the connection we don't need to do the TLS handshake (but we also can't send the 429). But if we accept the connection and then immediately close it after sending the 429 we've already done an awful lot of work for the TLS handshake and that's the primary thing the connection rate limiting is intended to protect against. Once you've successfully connected it doesn't really matter how many connections you have; you can DoS the cluster by sending a ton of work to it (which is what #13480 is for).

I'm going to ping my colleagues @shoenig and @schmichael here, as they both worked on the original implementation and the PR this PR reverts. They might have more thoughts.

> Beyond this, rejecting this PR in Nomad raises an issue about Consul's current behavior- if this behavior was considered a DoS issue in Nomad, does that mean that it's currently a DoS issue for Consul? At the very least I'd like to see the various Hashicorp products agree on _some_ consistent behavior across them, whatever it may be.

Agreed.

wjordan:
Please see [wjordan/nomad#http_max_conn_limit_429](https://github.com/hashicorp/nomad/compare/main...wjordan:nomad:http_max_conn_limit_429?expand=1) for followup commits (3372c070ffa8e2a63bd5eeab1a2a512f4f134cb1, 9768e429ea457a5f3b39cb3d330884de5a094adb) that should address the concerns raised, by rate-limiting the 429 responses and adding a (not rate-limited) counter for tracking the total number of limit-exceeding connection attempts. Please let me know if this headed in a helpful direction and I will open a new PR.

schmichael:
## This PR

tl;dr -> :+1: :shipit: 

Hi @wjordan! Thanks for sticking with this! Security-sensitive issues are always tricky, and obviously we've gone back and forth on it ourselves.

In the absence of unique factors, Nomad follows Consul's lead on such matters, so I think we can accept this PR and match their behavior. It obviously increases risk somewhat, but the current behavior obviously causes significant pain for legitimate users and is very difficult to debug. I'm willing to take the risk for the sake of the common case.

## Followups

As far as the followup commits are concerned, I think the metric would be great. :+1: 

I like your approach to logging as it limits the rate to 10/s. AFAICT from reading https://pkg.go.dev/golang.org/x/time/rate#NewLimiter the 100 burst limit will not take effect because we only call `Reserve()` and never the `...N()` variant.

I think we can just set `burst=1` and leave the rest be. I suspect bursts would only contain the same remote IP more times, so it seems like we can have a fairly low rate limit on this log line. Perhaps even lower it to 1/s? I'm fine with 1 or 10.

tgross (approved):
Thanks @schmichael! 

@wjordan I've kicked CI to re-run the test flakes but it looked like lint-go was still failing for `goimport -d`. I've done that and added a changelog entry. Once CI is green again we'll get this merged. Thanks for the PR and being patient with us figuring this out! 😀

tgross:
This has been merged and will ship in the next version of Nomad!

wjordan:
:tada: Thanks for seeing this through! I understand the extra scrutiny required for security-sensitive issues and appreciate the extra attention.

Note that I had pushed the followup commits (adding rate-limiting and logging/metric) to this PR, so they were included in the merge-commit. It sounded like @schmichael had requested some slight changes (or at least some further discussion) on the followups, so just wanted to make sure the full merge was intended.

---
> I like your approach to logging as it limits the rate to 10/s. AFAICT from reading https://pkg.go.dev/golang.org/x/time/rate#NewLimiter the 100 burst limit will not take effect because we only call Reserve() and never the ...N() variant.

`Reserve` is just shorthand for `ReserveN(time.now(), 1)`, so the burst limit should apply correctly. (It seemed fine in my local ad-hoc testing)

> I think we can just set burst=1 and leave the rest be. I suspect bursts would only contain the same remote IP more times, so it seems like we can have a fairly low rate limit on this log line. Perhaps even lower it to 1/s? I'm fine with 1 or 10.

That would make sense for the log entries, but aa rate-limit too low could make 429 responses inconsistent if multiple processes are trying to establish connections. Maybe adding a separate 1/s rate-limit for log entries would make sense, even if it adds a bit of extra complexity?

Issue #13553, opened 2022-07-04
UI: show Jobs, Task Groups, and Tasks that a variable (being edited or read) can be accessed by


philrenaud:
If the path is prefixed with `/job`, parse out the path string to identify the job, group, and tasks that have access.

### From job/group/task pages, in the "JOB DETAILS" bar:
- [x] Link to variables/var/job/jobname file from jobs/job (if such a file exists)
- [x] Link to variables/var/job/jobname/groupname from jobs/job/group
- [x] Link to variables/path/job/jobname/groupname/taskname from allocations/allocation/task

### From the variables to their job/group/task entities:
- [x] Link from a variable that you're editing to any entities that have access to it by default
- [x] Link from a variable at rest to any entities that have access to it by default (just viewing the var page)
- [x] When editing, inform the user that they can prefix their path with job, group or task name to automatically expose the secure variable to a job/group/task.

PR #13627, opened 2022-07-06
Link Variables to Jobs

philrenaud:
Adds a hint on edit, and a link on view, to secure variables.
image
image

--- 
Related: [13553](https://github.com/hashicorp/nomad/issues/13553)


ChaiWithJai (changes requested):
`blocking`:  Can this get a set of tests?

ui/app/models/variable.js
```diff
@@ -80,4 +80,33 @@ export default class VariableModel extends Model {
...
+    if (this.path.startsWith('jobs/') && this.path.split('/').length <= 4) {
+      return this.path
+        .split('/')
+        .slice(1, 4)
+        .reduce((acc, pathPart, index) => {
+          acc[entityTypes[index]] = pathPart;
+          return acc;
+        }, emptyEntities);
+    } else {
+      return emptyEntities;
```
ChaiWithJai, line 109: `blocking`: I think this needs feature and programmer tests otherwise, I think this will break in the future.

ui/app/templates/variables/variable/index.hbs
```diff
@@ -66,6 +66,11 @@
     {{/if}}
   </div>
 </h1>
+
+{{#if (or this.model.pathLinkedEntities.job this.model.pathLinkedEntities.group this.model.pathLinkedEntities.task)}}
+  <SecureVariableForm::RelatedEntities @job={{this.model.pathLinkedEntities.job}} @group={{this.model.pathLinkedEntities.group}} @task={{this.model.pathLinkedEntities.task}} />
+{{/if}}
```
ChaiWithJai, line 72: `blocking`: Can this get an acceptance? I'm pretty sure I'll break this in the future.

ui/app/components/secure-variable-form.hbs
```diff
@@ -16,6 +16,13 @@
         {{autofocus}}
         data-test-path-input
       />
+      {{#if @model.isNew}}
```
ChaiWithJai, line 19: `thought` `suggestion` `non-blocking`:  This conditional makes me want to rethink the component API. The component is used for new and existing variables and I think we can clean this up to reflect those differences.

ui/tests/unit/models/variable-test.js
```diff
@@ -48,4 +48,68 @@ module('Unit | Model | variable', function (hooks) {
     assert.equal(Object.entries(model.items)[1][0], 'myVar');
     assert.equal(Object.entries(model.items)[1][1], 'myValue');
   });
+
+  test('it computes linked entities', function (assert) {
+    let store = this.owner.lookup('service:store');
+
+    let model = store.createRecord('variable');
+    model.setProperties({
+      path: 'jobs/my-job-name/my-group-name/my-task-name',
```
ChaiWithJai, line 57: `question`:  How will this handle parameterized and periodic jobs?

ChaiWithJai (approved):
There might be some lurking bugs here, but we can wait until its time for smoke testing to see if they're actually an issue worth addressing.

ui/tests/unit/models/variable-test.js
philrenaud: linking should work the same way; jobs.job and jobs.job.task-group. I won’t be linking to the dispatched instance of it, but that would be less useful anyhow. The nature of Secure Vars means that any dispatched job that matches the glob pattern of the variable has access (not a specific dispatch)

PR #13629, opened 2022-07-06
ui:  allow namespaces to be set in forms

ChaiWithJai:
This PR handles setting a namespace in the create secure variables form.

Actual:
image

Figma:
image

### DIFF
--- a/ui/app/components/secure-variable-form.hbs
+++ b/ui/app/components/secure-variable-form.hbs
@@ -1,7 +1,7 @@
 {{did-update this.onViewChange @view}}
 {{did-insert this.establishKeyValues}}
 <form class="new-secure-variables" autocomplete="off" {{on "submit" this.save}}>
-  <div>
+  <div class="path-namespace">
     <label>
       <span>
         Path
@@ -39,9 +39,23 @@
         .
       </p>
     {{/if}}
+    <SecureVariableForm::NamespaceFilter 
+      @data={{hash 
+        disabled=(not @model.isNew)
+        selection=this.variableNamespace
+        namespaceOptions=this.namespaceOptions
+      }}
+      @fns={{hash
+        onSelect=this.setNamespace
+        setNamespaceOptions=this.setNamespaceOptions
+      }}
+    />
   </div>
   {{#if (eq this.view "json")}}
-    <div class="editor-wrapper boxed-section-body is-full-bleed {{if this.JSONError "error"}}">
+    <div
+      class="editor-wrapper boxed-section-body is-full-bleed
+        {{if this.JSONError 'error'}}"
+    >
       <div
         data-test-json-editor
         {{code-mirror
@@ -49,7 +63,7 @@
           onUpdate=this.updateCode
           extraKeys=(hash Cmd-Enter=(action "save"))
         }}
-      />
+      ></div>
       {{#if this.JSONError}}
         <p class="help is-danger">
           {{this.JSONError}}
--- a/ui/app/components/secure-variable-form.js
+++ b/ui/app/components/secure-variable-form.js
@@ -31,6 +31,18 @@ export default class SecureVariableFormComponent extends Component {
    * @type {DuplicatePathWarning}
    */
   @tracked duplicatePathWarning = null;
+  @tracked variableNamespace = 'default';
+  @tracked namespaceOptions = null;
+
+  @action
+  setNamespace(namespace) {
+    this.variableNamespace = namespace;
+  }
+
+  @action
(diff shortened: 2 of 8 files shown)


ui/app/components/secure-variable-form.hbs
```diff
@@ -1,8 +1,8 @@
 {{did-update this.onViewChange @view}}
 {{did-insert this.establishKeyValues}}
 <form class="new-secure-variables" autocomplete="off" {{on "submit" this.save}}>
-  <div>
-    <label>
+  <div style="display: flex">
+    <label style="flex-grow: 5">
```
ChaiWithJai, line 5: @philrenaud I have some bigger priorities than doing guesswork on CSS, can we just ticket this and come back to it later? I noticed you're drifting from our normal CSS pattern in favor of a grid format for the row below this, but I don't have the bandwidth to learn CSS Grid.

I care more about addressing the `403` error that I reported in Slack along with the saving collision issue and testing/refactoring.

philrenaud: Regardless of what we use here, assigning the styles inline is an approach that doesn't scale. Let's move these to the rest of the secure variables scss file

ui/mirage/factories/variable.js
```diff
@@ -22,4 +22,19 @@ export default Factory.extend({
...
+    if (!variable.namespace) {
+      const namespace = server.db.namespaces.length
+        ? pickOne(server.db.namespaces).id
+        : null;
+      variable.update({
+        namespace,
+      });
+    } else {
+      variable.update({
+        namespace: variable.namespace,
```
philrenaud, line 36: Noticing a bug here: the provided namespace and the returned namespace don't match in mirage
image

ChaiWithJai: That's a good catch, we probably have a similar bug in the `jobs` factory because that's where I copied this code from.

ui/app/components/secure-variable-form.js
```diff
@@ -31,6 +32,22 @@ export default class SecureVariableFormComponent extends Component {
...
+  @tracked variableNamespace = 'default';
+
+  @action
+  setNamespace(namespace) {
+    this.variableNamespace = namespace;
+  }
+
+  get namespaceOptions() {
+    const namespaces = this.store
+      .peekAll('namespace')
```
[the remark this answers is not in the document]
philrenaud: Hm; any idea why this might be not showing namespaces after setting acl? Are you able to reproduce?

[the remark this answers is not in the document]
ChaiWithJai: Maybe the `this.store.unloadAll();` logic that's in `routes/application`? There's tons of hidden gotchas throughout the entire UI. I'll handle data loading directly in the component using the `Trigger` component we have.

ui/app/components/secure-variable-form.hbs
```diff
@@ -33,17 +33,31 @@
         .
       </p>
     {{/if}}
+    <SecureVariableForm::NamespaceFilter
```
ChaiWithJai, line 36: @philrenaud moved data loading logic into this component. Here's the [commit](https://github.com/hashicorp/nomad/pull/13629/commits/cef152780fcfe4b255a360ef16cbe765bb3fa716) so you can review this piece in isolation.

ui/app/components/secure-variable-form/namespace-filter.hbs
```diff
@@ -0,0 +1,24 @@
+<Trigger @do={{this.fetchNamespaces}} @onSuccess={{this.formatAndSetNamespaces}} as |trigger|>
```
philrenaud, line 1: Q: I'm not generally familiar with our `Trigger` component, but is it comparable to a `did-insert` render modifier?

philrenaud (approved):

ui/app/components/secure-variable-form/namespace-filter.hbs
ChaiWithJai: Not quite. The `Trigger` component is how we handle asynchronous behavior in a component. The `Trigger` component has a state machine, where we trigger the `onTrigger` callback function to run when the component mounts, and then we also allow subsequent callback functions to run based to the result of the Promise and that's why the component API has `onSuccess` and `onError` and it resets its state back to `idle` using `triggerTask` so we don't get stuck in a failed state.

We use `did-insert` to trigger a callback function to run when a component mounts. And I think in the future, we might be able to a template helper to trigger a fetch request directly in the template.

PR #13630, opened 2022-07-06
boltdd: fixup typo and use unreachable assertion 

shoenig:
- helpers: fix critical typo 
  - we'll just pretend this never happened 
- helpers: use unreachable assertion


schmichael (approved):
:see_no_evil:

PR #13631, opened 2022-07-06
docs: add docs for simple load balancing nomad services

shoenig:
This PR adds a section to template docs for simple load balancing with nomad servicse.

### DIFF
--- a/website/content/docs/job-specification/template.mdx
+++ b/website/content/docs/job-specification/template.mdx
@@ -312,6 +312,33 @@ EOF
   }
 ```
 
+### Simple Load Balancing with Nomad Services
+
+~> Simple load balancing with Nomad Services is new in Nomad 1.3.2.
+
+The `nomadService` function now supports simple load balancing by selecting
+instances of a service via [rendezvous hashing][rhash].
+To enable simple load balancing, the `nomadService` function requires 3 arguments.
+
+- The number of services to select
+- The hashing key (should be unique, but consistent per requester)
+- The service name
+
+By using `NOMAD_ALLOC_ID` as the hashing key, the selected instances will remain mostly
+stable for the allocation, even as instances of the service are added or removed.
+
+```hcl
+template {
+  data        = <<EOH
+{{$allocID := env "NOMAD_ALLOC_ID" -}}
+# Configuration for 3 redis instances, as assigned via rendezvous hashing.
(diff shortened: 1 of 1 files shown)


website/content/docs/job-specification/template.mdx
```diff
@@ -312,6 +312,33 @@ EOF
   }
 ```
 
+### Simple Load Balancing with Nomad Services
+
+~> Simple load balancing with Nomad Services is new in Nomad 1.3.2.
+
+The `nomadService` function now supports simple load balancing by selecting
```
mikenomitch, line 319: Does `nomadServices` support it too?

website/content/docs/job-specification/template.mdx
```diff
@@ -312,6 +312,33 @@ EOF
...
+instances of a service via [rendezvous hashing][rhash].
+To enable simple load balancing, the `nomadService` function requires 3 arguments.
+
+- The number of services to select
+- The hashing key (should be unique, but consistent per requester)
+- The service name
+
+By using `NOMAD_ALLOC_ID` as the hashing key, the selected instances will remain mostly
+stable for the allocation, even as instances of the service are added or removed.
```
mikenomitch, line 328: Maybe add a bit more detail to "mostly stable"?

Thoughts on this: "Each time the template code is run, if the service instance remain the same, `nomadService` will return the same instances for each allocation. If service instances are added or removed, such that there are now N service instances, there is a 1/N chance of a selected instance being replaced. This helps maintain a more consistent output when writing configuration files and allows for fewer restarts and signaling of Nomad Tasks."

website/content/docs/job-specification/template.mdx
```diff
@@ -312,6 +312,33 @@ EOF
...
+template {
+  data        = <<EOH
+{{$allocID := env "NOMAD_ALLOC_ID" -}}
+# Configuration for 3 redis instances, as assigned via rendezvous hashing.
+{{range nomadService 3 $allocID "redis"}}
+  server {{ .Address }}:{{ .Port }};{{- end }}
+{{- end}}
+EOH
+}
+```
```
mikenomitch, line 340: Can we also show selecting a single one? - I think that'll be the majority of the use

mikenomitch (approved):
Pre-emptively approving but see comments

PR #13651, opened 2022-07-08
core: merge reserved_ports into host_networks

### LINKED ISSUE #13505: Reserved port duplicates/conflicts block all allocations
### Nomad version

Linux

### Issue
Any duplicate port reservations or conflicts within the agent config will result in all job allocations failing with a port conflict. Even jobs that don't have any port specification will fail.

It can be debated if this is a bug, as there are only very specific use cases where you have multiple aliases for the same network or have reserved port ranges defined on a global and host_network specific level. You could argue that in this case it would be ok if people can only define the port reservation on one of the network aliases or on the global level and only have network specific ports in the alias config. But it feels rather unnatural to me and in that case should have some giant warnings in the documentation about this.

### Reproduction steps
Either have a global port reservation overlap a host_network reservation or have multiple aliases for the same address range reserve the same ports.

#### Expected Result
In the past the port reservations merged and didn't cause any conflicts.

#### Actual Result
Allocations fail because of port conflicts, even if the job isn't reserving any ports

### Job file (if appropriate)
Anny job will do, I've tested locally with the example job (`nomad job init`)

agent config snipped
```
    host_network "LAN1" {
        cidr = "192.168.0.0/24"
        reserved_ports = "22"
    }
    host_network "default" { 
        cidr = "192.168.0.0/24"
        reserved_ports = "22"
    }
```
The conflict also happens if there's a `reserved` block including a port which is also declared in one of the host_network definitions.

  Task Group "cache" (failed to place 1 allocation):
    * Class "gp_compute" exhausted on 1 nodes
    * Dimension "network: port collision" exhausted on 1 nodes

schmichael:
Fixes #13505

This fixes #13505 by treating reserved_ports like we treat a lot of jobspec settings: merging settings from more global stanzas (`client.reserved.reserved_ports`) "down" into more specific stanzas (`client.host_networks[].reserved_ports`).

As discussed in #13505 there are other options, and since it's totally broken right now we have some flexibility:

1. Treat overlapping `reserved_ports` on addresses as invalid and refuse to start agents. However, I'm not sure there's a cohesive model we want to publish right now since so much 0.9-0.12 compat code still exists! We would have to explain to folks that if their `-network-interface` and `host_network` addresses overlapped, they could only specify reserved_ports in one place or the other?! It gets ugly.
2. Use the global client.reserved.reserved_ports value as the default and treat host_network[].reserverd_ports as overrides. My first suggestion in the issue, but @groggemans made me realize the addresses on the agent's interface (as configured by -network-interface) may overlap with host_networks, so you'd need to _remove_ the global reserved_ports from addresses shared with a shared network?! This seemed really confusing and subtle for users to me.

So I think "merging down" creates the most expressive yet understandable approach. I've played around with it a bit, and it doesn't seem too surprising. The only frustrating part is how difficult it is to _observe_ the available addresses and ports on a node! However that's a job for another PR.

Sorry I snuck in a couple other refactorings. I really want to make this code more maintainable, so I tried to move it in that direction where I didn't think it would be a huge distraction (eg the `interface{} -> string` switch). I can back out any of that if you think it's best to keep this tight and focused.

### DIFF
--- a/command/agent/agent.go
+++ b/command/agent/agent.go
@@ -617,6 +617,20 @@ func convertClientConfig(agentConfig *Config) (*clientconfig.Config, error) {
 	}
 	conf.HostVolumes = hvMap
 
+	// Ensure only one of reserved.reserved_ports or
+	// host_network.reserved_ports is set.
+	/*
+		if globalRes := agentConfig.Client.Reserved.ReservedPorts; globalRes != "" {
+			for _, hostnet := range agentConfig.Client.HostNetworks {
+				if hostRes := hostnet.ReservedPorts; hostRes != "" {
+					// Global and network-specific reserved ports
+					// aren't allowed.
+					return nil, fmt.Errorf("Cannot specify reserved.reserved_ports (%q) and host_network[%q].reserved port.", globalRes, hostRes)
+				}
+			}
+		}
+	*/
+
 	// Setup the node
 	conf.Node = new(structs.Node)
 	conf.Node.Datacenter = agentConfig.Datacenter
--- a/command/agent/agent.go
+++ b/command/agent/agent.go
@@ -617,20 +617,6 @@ func convertClientConfig(agentConfig *Config) (*clientconfig.Config, error) {
 	}
 	conf.HostVolumes = hvMap
 
-	// Ensure only one of reserved.reserved_ports or
-	// host_network.reserved_ports is set.
-	/*
-		if globalRes := agentConfig.Client.Reserved.ReservedPorts; globalRes != "" {
-			for _, hostnet := range agentConfig.Client.HostNetworks {
-				if hostRes := hostnet.ReservedPorts; hostRes != "" {
-					// Global and network-specific reserved ports
-					// aren't allowed.
-					return nil, fmt.Errorf("Cannot specify reserved.reserved_ports (%q) and host_network[%q].reserved port.", globalRes, hostRes)
-				}
-			}
-		}
-	*/
-
 	// Setup the node
 	conf.Node = new(structs.Node)
 	conf.Node.Datacenter = agentConfig.Datacenter
--- a/nomad/structs/network.go
+++ b/nomad/structs/network.go
@@ -36,13 +36,18 @@ var (
 
 // NetworkIndex is used to index the available network resources
 // and the used network resources on a machine given allocations
+//
+// Fields are exported so they may be JSON serialized for debugging.
+// Fields are *not* intended to be used directly.
 type NetworkIndex struct {
-	AvailNetworks  []*NetworkResource              // List of available networks
-	NodeNetworks   []*NodeNetworkResource          // List of available node networks
-	AvailAddresses map[string][]NodeNetworkAddress // Map of host network aliases to list of addresses
-	AvailBandwidth map[string]int                  // Bandwidth by device
-	UsedPorts      map[string]Bitmap               // Ports by IP
-	UsedBandwidth  map[string]int                  // Bandwidth by device
+	TaskNetworks  []*NetworkResource              // List of available networks
+	GroupNetworks []*NodeNetworkResource          // List of available node networks
+	HostNetworks  map[string][]NodeNetworkAddress // Map of host network aliases to list of addresses
+	UsedPorts     map[string]Bitmap               // Ports by IP
+
+	// Deprecated bandwidth fields
+	AvailBandwidth map[string]int // Bandwidth by device
+	UsedBandwidth  map[string]int // Bandwidth by device
 
 	MinDynamicPort int // The smallest dynamic port generated
 	MaxDynamicPort int // The largest dynamic port generated
@@ -51,9 +56,9 @@ type NetworkIndex struct {
 // NewNetworkIndex is used to construct a new network index
 func NewNetworkIndex() *NetworkIndex {
 	return &NetworkIndex{
-		AvailAddresses: make(map[string][]NodeNetworkAddress),
-		AvailBandwidth: make(map[string]int),
+		HostNetworks:   make(map[string][]NodeNetworkAddress),
 		UsedPorts:      make(map[string]Bitmap),
+		AvailBandwidth: make(map[string]int),
 		UsedBandwidth:  make(map[string]int),
 		MinDynamicPort: DefaultMinDynamicPort,
 		MaxDynamicPort: DefaultMaxDynamicPort,
@@ -84,9 +89,9 @@ func (idx *NetworkIndex) Copy() *NetworkIndex {
 	c := new(NetworkIndex)
 	*c = *idx
 
-	c.AvailNetworks = copyNetworkResources(idx.AvailNetworks)
-	c.NodeNetworks = copyNodeNetworks(idx.NodeNetworks)
-	c.AvailAddresses = copyAvailAddresses(idx.AvailAddresses)
+	c.TaskNetworks = copyNetworkResources(idx.TaskNetworks)
+	c.GroupNetworks = copyNodeNetworks(idx.GroupNetworks)
+	c.HostNetworks = copyAvailAddresses(idx.HostNetworks)
 	if idx.AvailBandwidth != nil && len(idx.AvailBandwidth) == 0 {
 		c.AvailBandwidth = make(map[string]int)
 	} else {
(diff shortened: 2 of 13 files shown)


nomad/structs/network.go
```diff
@@ -171,61 +176,131 @@ func (idx *NetworkIndex) Overcommitted() bool {
...
+// An error is returned if the Node cannot produce a consistent NetworkIndex
+// such as if reserved_ports are unparseable. Any such Node is effectively
+// unschedulable for workloads which require networks.
+//
+// Any errors returned by SetNode indicate a bug! The bug may lie in client
+// code not properly validating its configuration or it may lie in improper
+// Node object handling by servers. Users should not be able to cause SetNode
+// to error. Data that cause SetNode to error should be caught upstream such as
+// a client agent refusing to start with an invalid configuration.
+func (idx *NetworkIndex) SetNode(node *Node) error {
```
schmichael, line 196: It's easy to miss in the diff, so I want to call this refactoring out:

SetNode used to return `(collide bool, reason string)` just like AddAllocs. Functionally returning an error really isn't any different, but I wanted to *signal* something different to developers: SetNode should never have a "collision!"

AddAllocs can have collisions. It's a normal part of trying to make placements and preemptions and failing. There's nothing "erroneous" about that.

SetNode on the the other hand should *never return an error at runtime.* A better way to put it is: anything that SetNode considers an error could have been caught upstream (like on the client agent) through validation. I thought the old call signature really sent the wrong message that SetNode collisions could Just Happen as a normal part of an optimistically concurrent scheduler.

nomad/structs/network.go
```diff
@@ -51,9 +56,9 @@ type NetworkIndex struct {
 // NewNetworkIndex is used to construct a new network index
 func NewNetworkIndex() *NetworkIndex {
 	return &NetworkIndex{
-		AvailAddresses: make(map[string][]NodeNetworkAddress),
-		AvailBandwidth: make(map[string]int),
+		HostNetworks:   make(map[string][]NodeNetworkAddress),
```
tgross, line 59: At first I was thinking it would be safer to create the `TaskNetworks` and `GroupNetworks` slices here too, but I'm realizing by leaving them out we effectively enforce that `SetNode` has been called before use (on pain of panic). So :+1: here

scheduler/propertyset.go
```diff
@@ -346,10 +346,6 @@ func getProperty(n *structs.Node, property string) (string, bool) {
 	if !ok {
 		return "", false
 	}
-	nodeValue, ok := val.(string)
-	if !ok {
-		return "", false
-	}
 
-	return nodeValue, true
+	return val, true
```
tgross, line 350: With the signature of `resolveTarget` refactored, I think we can reduce this even more to `return resolveTarget(property, n)`?

scheduler/rank.go
```diff
@@ -211,13 +211,13 @@ OUTER:
...
-		if collide, reason := netIdx.SetNode(option.Node); collide {
+		if err := netIdx.SetNode(option.Node); err != nil {
 			iter.ctx.SendEvent(&PortCollisionEvent{
-				Reason:   reason,
+				Reason:   err.Error(),
 				NetIndex: netIdx.Copy(),
 				Node:     option.Node,
 			})
-			iter.ctx.Metrics().ExhaustedNode(option.Node, "network: port collision")
+			iter.ctx.Metrics().ExhaustedNode(option.Node, "network: invalid node")
```
tgross, line 220: This is great: the refactoring makes the reasoning more obvious to us as developers, but changing the message here will also make this case stand out loudly in plan metrics when we try to debug it.

nomad/structs/network.go
```diff
@@ -401,29 +441,40 @@ func (idx *NetworkIndex) AddReservedPortsForIP(ports string, ip string) (collide
...
+		if ip[j] > 0 {
+			break
+		}
+	}
+}
+
+// AssignPorts based on an ask from the scheduler processing a group.network
+// stanza. Supports multi-interfaces through node configured host_networks.
+//
+// AssignNetwork supports the deprecated task.network stanza.
```
tgross, line 477:
```suggestion
// AssignTaskNetwork supports the deprecated task.network stanza.
```

nomad/structs/network.go
```diff
@@ -171,61 +176,131 @@ func (idx *NetworkIndex) Overcommitted() bool {
...
 	} else if node.Resources != nil {
-		networks = node.Resources.Networks
+		taskNetworks = node.Resources.Networks
 	}
 
+	// nodeNetworks are used for group.network asks.
 	var nodeNetworks []*NodeNetworkResource
 	if node.NodeResources != nil && len(node.NodeResources.NodeNetworks) != 0 {
 		nodeNetworks = node.NodeResources.NodeNetworks
 	}
```
tgross, line 214: Nitpick: so it doesn't look like we reference or mutate `node.NodeResources` or `nodeNetworks` at all in the two blocks below, so it might be better to move this block down to right above where it's first used (right above the `// TODO: upgrade path?` comment).

schmichael: Good call. Also removing that TODO because I don't even know what it's suggesting.

schmichael:
I spent a long time playing with this today and wanted to share some findings:

A key aspect of Nomad's networking is that ports are reserved by IPs, not "networks" whether defined with `network_interface` or `host_networks`. This makes sense as `(IP:Port)` tuples are what matter at the end of the day. A network device can't have a port reservation but IP addresses on that device can.

So I realized what probably would be the most correct approach to `reserved_port` handling is:

1. `client.reserved.reserved_ports` reserves ports on the *IP addresses associated with* the `default` network (this is the network defined by `network_interface`)
2. `client.host_networks[].reserved_ports` reserve ports on the *IP addresses defined by their CIDR or device.*

This means port reservations may or may not overlap. For example:

- The `default` network is 192.168.1.1 and reserves port 22
- Host network `eth0` is 192.168.1.0/24 and reserves port 80

The end result would be:

- 192.168.1.1 has ports 22 *and* 80 reserved!
- The rest of the 192.168.1.0/24 subnet would have only port 80 reserved.

But this isn't how existing code works, and all of the comments and docs seem to treat `client.reserved.reserved_ports` the way this PR does: **as the global port reservations.** This is conceptually simpler but does mean you can't reserve a port on the default network and *not* on a `host_network`. "Fixing" that behavior would be a backward compatibility issue.

That being said the output of `nomad node status -verbose` lists `Host Networks` *without* their global port reservations! So if the intended behavior is that they inherit reserved ports, we probably need to fix that as well.

nomad/structs/network.go
```diff
@@ -36,13 +36,18 @@ var (
...
+// Fields are *not* intended to be used directly.
 type NetworkIndex struct {
-	AvailNetworks  []*NetworkResource              // List of available networks
-	NodeNetworks   []*NodeNetworkResource          // List of available node networks
-	AvailAddresses map[string][]NodeNetworkAddress // Map of host network aliases to list of addresses
-	AvailBandwidth map[string]int                  // Bandwidth by device
-	UsedPorts      map[string]Bitmap               // Ports by IP
-	UsedBandwidth  map[string]int                  // Bandwidth by device
+	TaskNetworks  []*NetworkResource              // List of available networks
+	GroupNetworks []*NodeNetworkResource          // List of available node networks
```
lgfa29, line 44: Are these values deprecated as well? `TaskNetworks` seems to be related to the network at `task -> resource` level and `GroupNetworks` I'm not even sure where they are used 😅 

If they are not, maybe update the comments as well?

```suggestion
	TaskNetworks  []*NetworkResource              // List of networks defined at the task level
	GroupNetworks []*NodeNetworkResource          // List of networks defined at the group level (???)
```

nomad/structs/network.go
lgfa29: This is great. I wonder now what kind of actions we can take when an error is returned here. Maybe force a leadership transition to see if a new leader is able to handle the node? Or have a way to flush node state and request new fingerprint data from the node?

nomad/structs/network.go
```diff
@@ -171,61 +176,142 @@ func (idx *NetworkIndex) Overcommitted() bool {
...
+	// client.host_network["eth1"] = {reserved_ports = "1-1000"}
+	//
+	// Addresses on taskNetworks reserve port 22
+	// Addresses on eth0 reserve 22,80,443 (note 22 is also reserved!)
+	// Addresses on eth1 reserve 1-1000
+	globalResPorts := []uint{}
+
+	// COMPAT(0.11): Remove in 0.11
+	// Handle reserving ports, handling both new and old client Node
+	// layouts
```
lgfa29, line 223: Is the `else if node.Reserved != nil` code path that is deprecated?

nomad/structs/network_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	"github.com/hashicorp/nomad/ci"
+	"github.com/shoenig/test/must"
```
lgfa29, line 10: If the plan is to backport the bug fix we unfortunately can't use this yet 😢

website/content/docs/configuration/client.mdx
```diff
@@ -206,9 +206,12 @@ chroot as doing so would cause infinite recursion.
 
 - `disk` `(int: 0)` - Specifies the amount of disk to reserve, in MB.
 
-- `reserved_ports` `(string: "")` - Specifies a comma-separated list of ports to
-  reserve on all fingerprinted network devices. Ranges can be specified by using
-  a hyphen separated the two inclusive ends.
+- `reserved_ports` `(string: "")` - Specifies a comma-separated list of ports
+  to reserve on all fingerprinted network devices. Ranges can be specified by
+  using a hyphen separated the two inclusive ends. See also
```
tgross, line 211:
```suggestion
  using a hyphen separating the two inclusive ends. See also
```

nomad/structs/network.go
tgross: This implication of "SetNode on the the other hand should _never return an error at runtime_" is that the node itself is giving us bad data because of a programmer error, so I'd expect the node to give us the same bad fingerprint again. At that point our invariants are broken so I'm not sure it's a good idea to try to recover rather than throw an error that fails scheduling -- we want this to be noisy.

tgross (approved):
Once the `must` in tests are fixed up this LGTM

command/agent/agent.go
```diff
@@ -617,6 +617,20 @@ func convertClientConfig(agentConfig *Config) (*clientconfig.Config, error) {
...
+		if globalRes := agentConfig.Client.Reserved.ReservedPorts; globalRes != "" {
+			for _, hostnet := range agentConfig.Client.HostNetworks {
+				if hostRes := hostnet.ReservedPorts; hostRes != "" {
+					// Global and network-specific reserved ports
+					// aren't allowed.
+					return nil, fmt.Errorf("Cannot specify reserved.reserved_ports (%q) and host_network[%q].reserved port.", globalRes, hostRes)
+				}
+			}
+		}
+	*/
```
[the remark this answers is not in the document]
schmichael: Yes it is for option 1 but no we shouldn't keep it! Bad git hygiene on my part. Removing.

nomad/structs/network_test.go
schmichael: Doh, of course. I forgot `main` isn' 1.3.2. Thanks!

nomad/structs/network.go
schmichael: Good call. Wrote some more books-in-comments. :grimacing:

nomad/structs/network.go
schmichael: Correct, clarifying.

lgfa29 (approved):
Nice changes! The added comments in the structs are very helpful.

I added the `release/1.1.x` and `release/1.2.x` labels for backport.

PR #13658, opened 2022-07-08
snapshot restore-from-archive streaming and filtering

tgross:
This changeset implements two improvements to restoring FSM snapshots from archives:
* The existing implementation decompresses the archive to a temporary file before reading it in to the FSM. For large snapshots this performs a lot of disk IO. Stream decompress the snapshot as we read it, without first writing to a temporary file. This also moves some of the work to a second core.
* Add bexpr filters to the `RestoreFromArchive` helper. The operator can pass these as `-filter` arguments to `nomad operator snapshot state` (and other commands in the future) to include only desired data when reading the snapshot.

Deferred for this PR: the `nomad operator snapshot state` command still has to load everything that's been filtered into the FSM before writing it out to a large JSON blob. We should provide a tool that streams the decoded objects directly to an encoder without loading into the FSM, so that we can emit NDJSON, write out to a sqlite DB, etc.

---

Example:

Starting with a 439MB snapshot (~13GiB uncompressed), I want to filter for all objects associated with 3 different jobs and 3 different nodes:

```
time nomad operator snapshot state -filter '
    JobID == "job1" or
    JobID == "job2" or
    JobID == "job3" or
    NodeID == "3b3471d7-c519-8e3c-d7fd-dc692ca44744" or
    NodeID == "455775de-b4b4-0cb6-75eb-6c534618a005" or
    NodeID == "0d8e2a62-2712-cb4c-fb15-9831fdac57fe" or
    ID == "job1" or
    ID == "job2" or
    ID == "job3" or
    ID == "3b3471d7-c519-8e3c-d7fd-dc692ca44744" or
    ID == "455775de-b4b4-0cb6-75eb-6c534618a005" or
    ID == "0d8e2a62-2712-cb4c-fb15-9831fdac57fe"
' \
      ./nomad_operator_snapshot_save_2022_05_12_1543-0700.snap \
      > filtered-state.json

real    24m15.805s
user    29m55.036s
sys     7m50.618s

$ cat filtered-state.json| jq '.Allocs | length'
5490
$ cat filtered-state.json| jq '.Evals | length'
667
```

Previously this would write ~13GiB to disk, read 14GiB from disk, and saturate 1 core for over an hour before running out of memory on my machine (16GiB) and crashing.

With this change, the command reads ~450MiB from disk, only writes the 197MiB JSON blob to disk, and uses about 150% CPU, maxing out memory usage around 330MB.

### DIFF
--- a/.changelog/13658.txt
+++ b/.changelog/13658.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+cli: `operator snapshot state` supports `-filter` expressions and stream decompresses snapshots to avoid writing large temporary files
+```
--- a/.changelog/13658.txt
+++ b/.changelog/13658.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-cli: `operator snapshot state` supports `-filter` expressions and stream decompresses snapshots to avoid writing large temporary files
+cli: `operator snapshot state` supports `-filter` expressions and avoids writing large temporary files
 ```
--- a/helper/raftutil/snapshot.go
+++ b/helper/raftutil/snapshot.go
@@ -8,10 +8,11 @@ import (
 	"github.com/hashicorp/raft"
 
 	"github.com/hashicorp/nomad/helper/snapshot"
+	"github.com/hashicorp/nomad/nomad"
 	"github.com/hashicorp/nomad/nomad/state"
 )
 
-func RestoreFromArchive(archive io.Reader) (*state.StateStore, *raft.SnapshotMeta, error) {
+func RestoreFromArchive(archive io.Reader, filter *nomad.FSMFilter) (*state.StateStore, *raft.SnapshotMeta, error) {
 	logger := hclog.L()
 
 	fsm, err := dummyFSM(logger)
@@ -19,7 +20,7 @@ func RestoreFromArchive(archive io.Reader) (*state.StateStore, *raft.SnapshotMet
 		return nil, nil, fmt.Errorf("failed to create FSM: %w", err)
 	}
 
-	// r is closed by Restore, w is closed by CopySnapshot
+	// r is closed by RestoreFiltered, w is closed by CopySnapshot
 	r, w := io.Pipe()
 
 	errCh := make(chan error)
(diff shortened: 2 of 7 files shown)


.changelog/13658.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+cli: `operator snapshot state` supports `-filter` expressions and stream decompresses snapshots to avoid writing large temporary files
```
shoenig, line 2:
```suggestion
cli: `operator snapshot state` supports `-filter` expressions and avoids writing large temporary files
```

shoenig: feel free to ignore; just more to the point

helper/raftutil/snapshot.go
```diff
@@ -3,44 +3,48 @@ package raftutil
...
-		return nil, nil, fmt.Errorf("failed to seek: %w", err)
-	}
+	go func() {
+		meta, err := snapshot.CopySnapshot(archive, w)
+		if err != nil {
+			errCh <- fmt.Errorf("failed to read snapshot: %w", err)
+		}
+		if meta != nil {
+			metaCh <- meta
+		}
```
shoenig, line 36: the `select` statement below assumes the invariant that one of these will send, but using the double if here suggests its possible neither could; could this be an if/else instead?

nomad/fsm.go
```diff
@@ -1459,54 +1474,53 @@ func (n *nomadFSM) Restore(old io.ReadCloser) error {
...
-			 */
-			job.Canonicalize()
-
-			if err := restore.JobRestore(job); err != nil {
-				return err
+			if filter.Include(job) {
+				/* Handle upgrade paths:
+				 * - Empty maps and slices should be treated as nil to avoid
+				 *   un-intended destructive updates in scheduler since we use
+				 *   reflect.DeepEqual. Starting Nomad 0.4.1, job submission sanitizes
```
shoenig, line 1493:
```suggestion
				 *   reflect.DeepEqual
```

I've noticed `slices.Equal` and `maps.Equal` do the Right Thing; will be nice to purge `DeepEqual` soon

PR #10893, opened 2021-07-13
edit ember-can to add additional attribute for namespace

JBhagat841:
We wanted the ability to get our namespace from query params
in order to do this, we're using additional attributes via
ember-can to set a bound property directly from our
handlebar file. This sets us up better in the event that
the namespace filter changes on the UI because our handlebar
file will be aware of the change, whereas our ability may not
update as the namespace filter updates.

PR #13659, opened 2022-07-08
[bugfix, ui] Allow running jobs from a namespace-limited token

philrenaud:
Currently, there are a few things preventing a user from running a new job via the UI, in an environment where their ACL token has limited namespace abilities:

1. We had previously [tied the "Run job" button to an ability that depended on the current namespace](https://github.com/hashicorp/nomad/pull/10893#discussion_r673418277). However, this means that your ability to run a job is dependent on filtering your jobs index table, which is pretty confusing to an end-user. Worse, if you don't have any jobs currently running, you don't get any option to set this namespace filter at all.
    - The abilities/job `can run` ability now looks at all abilities across all namespaces within your policy. This means there's a chance you submit a job for which you don't have write permissions, but good news: it gets caught at several steps along the way upon submission. It's better to not restrict access to this editor and this change reflects that.
2. If you did somehow manage to make it to the /run page, ever since 1.2.6, [Nomad job parsing requires an ACL token](https://www.nomadproject.io/docs/upgrade/upgrade-specific#acl-requirement-for-the-job-parse-endpoint) but namespaces have never been sent along with that request. Thus, if you had `* { read }` and `myNamespace { write }`, it would try to parse your job without a namespace and return a 403.
    - This is a pretty circular issue: the thing that let us determine the namespace of a job was the `/parse` request. But now that that parse request is conditional upon the namespace... you get where this is going.
    - So, we could try to move the parse() functionality into the browser, which comes with a lot of its own chance for error (Job HCL parsing within the Nomad UI doesn't sound like something we should be doing). Instead, this PR opts to include a Namespaces dropdown on the job/run UI. It uses this when passing the job to `/parse`.
^--- **great news update**: Turns out we don't need to parse the namespace, we just need _a namespace_. All other things being equal, a `POST` to `/parse` will 403 but `/parse?namespace=*` will 20x.

Side-effect: Includes better error messaging for ACL permission errors upon job submission.

image

### DIFF
--- a/ui/app/controllers/jobs/run.js
+++ b/ui/app/controllers/jobs/run.js
@@ -3,7 +3,6 @@ import { inject as service } from '@ember/service';
 
 export default class RunController extends Controller {
   @service router;
-
   onSubmit(id, namespace) {
     this.router.transitionTo('jobs.job', `${id}@${namespace || 'default'}`);
   }
--- a/ui/app/controllers/jobs/run.js
+++ b/ui/app/controllers/jobs/run.js
@@ -1,56 +1,10 @@
 import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
-import { computed } from '@ember/object';
-import { scheduleOnce } from '@ember/runloop';
-import { serialize } from 'nomad-ui/utils/qp-serialize';
-import { get, set } from '@ember/object';
 
 export default class RunController extends Controller {
   @service router;
-  @service system;
-  @service store;
-
-  queryParams = [
-    {
-      qpNamespace: 'namespace',
-    },
-  ];
 
   onSubmit(id, namespace) {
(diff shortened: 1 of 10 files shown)


ChaiWithJai (approved):

ui/app/controllers/jobs/run.js
```diff
@@ -1,9 +1,56 @@
...
+      scheduleOnce('actions', () => {
+        // eslint-disable-next-line ember/no-side-effects
+        set(this, 'qpNamespace', '*');
+      });
+    }
+
+    return availableNamespaces;
+  }
+
+  setFacetQueryParam(queryParam, selection) {
```
philrenaud, line 47: Forgive the naming conventions (qpNamespace, queryParam, etc.) - will revise. Was pullig from jobs/volumes routes.

PR #13677, opened 2022-07-11
docs: move operator subcommands under their own trees

tgross:
The sidebar navigation tree for the `operator` sub-sub commands is
getting cluttered and we have a new set of commands coming to support
secure variables keyring as well. Move these all under their own
subtrees.


DerekStrickland (approved):
:shipit:

PR #13680, opened 2022-07-11
Adds sort, show/hide, and copy functionality to Variable pages

### LINKED ISSUE #13637: UI: Variable page copy-to-clipboard + show/hide
Note: doing away with k/v deletion from this page

image

philrenaud:
Resolves #13637


tgross (approved):
LGTM. 

One odd thing I noticed is that the current sort order is preserved if you switch to the JSON edit view. I double-checked and this is handled safely by the the API no problem, but I wonder if the order of the JSON document changing would give a user hesitation.

PR #13696, opened 2022-07-11
docs: rename Internals to Concepts

tgross:
Rename the Internals section to Concepts as a first step towards making this body of knowledge a little more approachable to operators. I've checked Learn and it looks like there are no existing links from Learn to this section (which probably says a lot), but I've added the redirects anyways.

@schmichael said, in an internal chat:
> I would totally +1 `Internals/Deployments` and/or renaming `Internals` to `Concepts` 

😀


website/redirects.js
```diff
@@ -1386,6 +1386,11 @@ module.exports = [
     destination: '/api-docs/:path*',
     permanent: true,
   },
+  {
+    source: '/docs/internals/:path*',
+    destination: '/docs/concepts/:path*',
+    permanent: true,
+  },
```
schmichael, line 1393: Well this is pleasant.

website/content/docs/concepts/index.mdx
```diff
@@ -1,15 +1,16 @@
...
-# Nomad Internals
+# Nomad Concepts
 
-This section covers the internals of Nomad and explains the technical
-details of how Nomad functions, its architecture, and sub-systems.
+This section covers the core concepts of Nomad and explains the
+technical details of how Nomad functions, its architecture, and
+sub-systems.
 
 -> **Note:** Knowledge of Nomad internals is not
```
schmichael, line 15: I wonder if we should just remove this. The last sentence "if you're operating Nomad you should know this" does a real 180 from the first sentence! The individual `Advanced` pages all have a warning at the top already, and I don't think anybody comes to nomadproject.io thinking they have to read every page to use Nomad? I hope not!

schmichael (approved):
:tada: :bow:

shoenig (approved):
LGTM!

PR #13705, opened 2022-07-11
SV: fixes for namespace handling

angrycub:
**ACL check namespace value in SecureVariable**
Since the RPC endpoint can handle multiple variables in a single Upsert and they could target different namespaces, we need to check the ACL permissions per update.

**Error on wildcard namespace**
Guards against someone having NOMAD_NAMESPACE or the namespace flag set to `*` on a write.


tgross (approved):
LGTM, nice catch on this.

PR #13707, opened 2022-07-11
SV: CLI: var list command

angrycub:
Includes the switch from the ApplyFunc to the FilterIterator for managing allowed namespaces and prefix.

### DIFF
--- a/command/var_list.go
+++ b/command/var_list.go
@@ -24,10 +24,11 @@ func (c *VarListCommand) Help() string {
 	helpText := `
 Usage: nomad var list [options] <prefix>
 
-  List is used to list available secure variables.
+  List is used to list available secure variables. Supplying an optional prefix,
+  filters the list to variables having a path starting with the prefix.
 
   If ACLs are enabled, this command will return only secure variables stored at
-  namespaced paths where the token has the `read` capability.
+  namespaced paths where the token has the ` + "`read`" + ` capability.
 
 General Options:
 
@@ -141,7 +142,7 @@ func (c *VarListCommand) Run(args []string) int {
 		items = vars
 
 		if quiet {
-			items = itemsToQuietJSONReadySlice(vars, c.Meta.namespace)
+			items = dataToQuietJSONReadySlice(vars, c.Meta.namespace)
 			obj = items
 		}
 
@@ -177,7 +178,7 @@ func (c *VarListCommand) Run(args []string) int {
 	case quiet:
 		c.Ui.Output(
 			formatList(
-				itemsToQuietStringSlice(vars, c.Meta.namespace)))
+				dataToQuietStringSlice(vars, c.Meta.namespace)))
 
 	case len(tmpl) > 0:
 		out, err := Format(json, tmpl, vars)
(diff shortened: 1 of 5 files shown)


command/var_list.go
```diff
@@ -0,0 +1,276 @@
...
+		},
+	)
+}
+
+func (c *VarListCommand) AutocompleteArgs() complete.Predictor {
+	return complete.PredictNothing
+}
+
+func (c *VarListCommand) Synopsis() string {
+	return "List secure variable specifications"
```
tgross, line 77: Maybe?

```suggestion
	return "List secure variable metadata"
```

command/var_list.go
```diff
@@ -0,0 +1,276 @@
...
+)
+
+type VarListCommand struct {
+	Prefix string
+	Meta
+}
+
+func (c *VarListCommand) Help() string {
+	helpText := `
+Usage: nomad var list [options] <prefix>
```
tgross, line 25: The doc here is missing a description of the `prefix` arg

command/var_list.go
```diff
@@ -0,0 +1,276 @@
...
+		rows[i+1] = fmt.Sprintf("%s|%s|%s",
+			sv.Namespace,
+			sv.Path,
+			time.Unix(0, sv.ModifyTime),
+		)
+	}
+	return formatList(rows)
+}
+
+func itemsToQuietStringSlice(vars []*api.SecureVariableMetadata, ns string) []string {
```
tgross, line 231: Nitpick: calling these functions "itemsTo..." overloads the usage of the field name `Items`, and this would be a little easier to read if we avoided it.

command/var_list.go
```diff
@@ -0,0 +1,276 @@
...
+	// If ns is the wildcard namespace, we have to provide namespace
+	// as part of the quiet output, otherwise it can be a simple list
+	// of paths.
+	toPathStr := func(v *api.SecureVariableMetadata) string {
+		if ns == "*" {
+			return fmt.Sprintf("%s|%s", v.Namespace, v.Path)
+		}
+		return v.Path
+	}
```
tgross, line 240: Q: Is the primary envisioned use case of `-q` for scripting? If so, does it make scripting harder if we only include the namespace conditionally?

command/var_list_test.go
```diff
@@ -0,0 +1,511 @@
...
+// 	for _, p := range ps {
+// 		if field == "namespace" && p.Namespace == val {
+// 			out = append(out, p)
+// 		}
+// 		if field == "prefix" && strings.HasPrefix(p.Path, val) {
+// 			out = append(out, p)
+// 		}
+// 	}
+// 	return &out
+// }
```
tgross, line 428: looks like some leftovers?

command/var_list_test.go
```diff
@@ -0,0 +1,511 @@
...
+		},
+		{
+			name:         "plaintext/quiet",
+			args:         []string{"-q"},
+			expectStdOut: strings.Join(variables.HavingNamespace(api.DefaultNamespace).Strings(), "\n"),
+		},
+		{
+			name:         "plaintext/quiet/prefix",
+			args:         []string{"-q", "a/b/c"},
+			expectStdOut: strings.Join(variables.HavingNSPrefix(api.DefaultNamespace, "a/b/c").Strings(), "\n"),
```
tgross, line 177: I don't think this is a blocker but maybe more for future reference: I feel like having logic spread across some new test-only structs and the return value of the variable setup code, etc. muddies the tests. The setup code isn't concurrent and all the paths are known in advance. What stops us from having the raw expected output here, maybe with test helper to strip off the timestamps? Then the test case reads fairly straightforwardly without having to refer to code at both the top and bottom of the page to know what we're doing.

tgross (approved):
LGTM 👍 

I've left some minor comments and suggestions but nothing that's really a blocker here.

command/var_list_test.go
angrycub, line 428:
```suggestion
```

PR #13708, opened 2022-07-12
13553 secure vars linked from jobs

philrenaud:
Adds a relationship between a job and variables, defined by any variables that are prefixed with that job's name.

Further adds a model property of `pathLinkedVariable` for all jobs, task groups, and tasks, that look for an exact match against their (or their parent job's) variables. Displays this in the details bar:

image
image
image

### DIFF
--- a/ui/app/models/job.js
+++ b/ui/app/models/job.js
@@ -328,4 +328,9 @@ export default class Job extends Model {
   // spec first. In order to preserve both the original HCL and the parsed response
   // that will be submitted to the create job endpoint, another prop is necessary.
   @attr('string') _newDefinitionJSON;
+
+  @computed('variables')
+  get pathLinkedVariable() {
+    return this.variables?.objectAt(0);
+  }
 }
--- a/ui/app/models/job.js
+++ b/ui/app/models/job.js
@@ -138,6 +138,7 @@ export default class Job extends Model {
   @hasMany('allocations') allocations;
   @hasMany('deployments') deployments;
   @hasMany('evaluations') evaluations;
+  @hasMany('variables') variables;
   @belongsTo('namespace') namespace;
   @belongsTo('job-scale') scaleState;
 
--- a/ui/app/models/job.js
+++ b/ui/app/models/job.js
@@ -329,8 +329,15 @@ export default class Job extends Model {
   // that will be submitted to the create job endpoint, another prop is necessary.
   @attr('string') _newDefinitionJSON;
 
-  @computed('variables')
+  @computed('variables', 'parent', 'plainId')
   get pathLinkedVariable() {
-    return this.variables?.objectAt(0);
+    if (this.parent.get('id')) {
+      return this.variables?.findBy(
+        'path',
+        `jobs/${JSON.parse(this.parent.get('id'))[0]}`
+      );
+    } else {
+      return this.variables?.findBy('path', `jobs/${this.plainId}`);
+    }
   }
 }
--- a/ui/app/models/task.js
+++ b/ui/app/models/task.js
@@ -50,4 +50,27 @@ export default class Task extends Fragment {
   @attr('number') reservedEphemeralDisk;
 
   @fragmentArray('volume-mount', { defaultValue: () => [] }) volumeMounts;
+
+  async _fetchParentJob() {
+    let job = await this.store.findRecord('job', this.taskGroup.job.id, {
+      reload: true,
+    });
+    this._job = job;
+  }
+
+  get pathLinkedVariable() {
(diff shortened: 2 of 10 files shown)


ui/app/routes/variables.js
```diff
@@ -16,7 +16,7 @@ export default class VariablesRoute extends Route.extend(WithForbiddenState) {
   }
   async model() {
     try {
-      const variables = await this.store.findAll('variable');
+      const variables = await this.store.findAll('variable', { reload: true });
```
philrenaud, line 19: Because we can now load a partial version of a variable (the filtered relationships of a job), we force reload upon entering the variables route.

ui/app/serializers/job.js
```diff
@@ -71,6 +71,10 @@ export default class JobSerializer extends ApplicationSerializer {
       .buildURL(modelName, hash.ID, hash, 'findRecord')
       .split('?');
 
+    const variableLookup = hash.ParentID
+      ? JSON.parse(hash.ParentID)[0]
+      : hash.PlainId;
```
philrenaud, line 76: variableLookup: parameterized and periodic jobs don't have knowable IDs, and therefore our variables won't assign to them directly, but instead to their parents.

ui/app/models/job.js
```diff
@@ -327,4 +328,16 @@ export default class Job extends Model {
   // spec first. In order to preserve both the original HCL and the parsed response
   // that will be submitted to the create job endpoint, another prop is necessary.
   @attr('string') _newDefinitionJSON;
+
+  @computed('variables', 'parent', 'plainId')
+  get pathLinkedVariable() {
+    if (this.parent.get('id')) {
```
ChaiWithJai, line 334: `thought` `suggestion`:  The `get` method is a little dangerous because we don't know if the request is synchronous or asynchronous. A work around that we use for getting the `id` is using `this.belongsTo('$ModelName').id()`. And a slightly hackier workaround to escape the behavior of a promise is using `model.content`.

ui/app/models/job.js
```diff
@@ -327,4 +328,16 @@ export default class Job extends Model {
   // spec first. In order to preserve both the original HCL and the parsed response
   // that will be submitted to the create job endpoint, another prop is necessary.
   @attr('string') _newDefinitionJSON;
+
+  @computed('variables', 'parent', 'plainId')
+  get pathLinkedVariable() {
```
ChaiWithJai, line 333: `praise`:  Great job handling the behavior of parameterized and periodic child jobs.

ui/app/models/task.js
```diff
@@ -50,4 +50,27 @@ export default class Task extends Fragment {
   @attr('number') reservedEphemeralDisk;
 
   @fragmentArray('volume-mount', { defaultValue: () => [] }) volumeMounts;
+
+  async _fetchParentJob() {
+    let job = await this.store.findRecord('job', this.taskGroup.job.id, {
+      reload: true,
+    });
+    this._job = job;
```
ChaiWithJai, line 58: `question`: I'm a little confused by why we're setting a private variable on the `Task` fragment. The `Task` fragment is owned by the `TaskGroup` fragment and that is stored (and owned) on our `Job` model. As a result of this relationship, shouldn't we already have access to the owner (the job)?

ui/app/models/task.js
```diff
@@ -50,4 +50,27 @@ export default class Task extends Fragment {
...
+      let jobID = this._job.plainId;
+      if (this._job.parent.get('plainId')) {
+        jobID = this._job.parent.get('plainId');
+      }
+      return this._job.variables?.findBy(
+        'path',
+        `jobs/${jobID}/${this.taskGroup.name}/${this.name}`
+      );
+    }
+  }
```
ChaiWithJai, line 75: `question`:  How will we handle this behavior in our Mirage factory?

ui/app/routes/variables.js
ChaiWithJai: `praise`:  Great catch! If in the future, we want to optimize for page load times (some pages trigger a ton of waterfall requests), we can use the `shouldBackgroundReloadAll` method in the adapter layer to render what we currently have and fetch the rest of the information in the background.

ChaiWithJai (approved):
Great job.

ui/app/models/task.js
philrenaud: Mirage factory is pretty simple right now: it just returns an unfiltered list of our vars at /vars. We don't attempt to mimic the filter behaviour because of the potential for complexity here, and because the `pathLinkedVariable` equality check works with a filtered or unfiltered variables list.

Issue #13120, opened 2022-05-25
☂️  SSO / OIDC / Token Expiry


philrenaud:
## ACL Token Expiration
The ability to set optional expiration times on ACL tokens. A prerequisite of Nomad SSO.

### RFC
  - [ ] (L) Draft RFC
  - [ ] (S) Circulate within Nomad team
  - [ ] (S) Circulate within RFC mailing list
### Implementation
  - [ ] (S) Update Server configuration with new options
  - [ ] (S) Modify ACL token API and CLI
  - [ ] (S) Modify ACL token state store schema
  - [ ] (S) Modify ACL token resolve RPC funcs
  - [ ] (M) Add new ACL token expiry garbage collector
### Docs and Test
  - [ ] (S) Update server config website documentation
  - [ ] (S) Update API ACL token website documentation
  - [ ] (S) Update CLI ACL token website documentation
  - [ ] (S) Identify and update internal docs around ACL usage
  - [ ] (M) Add token expiry E2E tests
  - [ ] (L) New Learn guide

## ACL Token Roles
The ability to group ACL policies within ACL roles. A prerequisite of Nomad SSO.

### RFC
  - [ ] (L) Draft RFC
  - [ ] (S) Circulate within Nomad team
  - [ ] (S) Circulate within RFC mailing list
### Implementation
  - [ ] (L) Implement ACL roles schema and store funcs
  - [ ] (M) Implement ACL roles structs and diff funcs
  - [ ] (L) implement ACL roles RPC endpoints
  - [ ] (L) implement ACL roles API endpoints
  - [ ] (M) implement ACL roles CLI
  - [ ] (S) Update ACL token create CLI
  - [ ] (L) Modify ACL token resolve functionality
  - [ ] (M) Modify replication functionality to include ACL roles
### Docs and Test
  - [ ] (M) Add ACL role API website documentation
  - [ ] (M) Add ACL role CLI website documentation
  - [ ] (M) Identify and update ACL authentication website documentation
  - [ ] (L) Add ACL role E2E tests
  - [ ] (L) New Learn guide

## ACL SSO (OIDC)
The SSO feature itself. There are still a number of unknown from my side as previous work focussed on expiry and roles.

### RFC
  - [ ] (L) Draft RFC
  - [ ] (S) Circulate within Nomad team
  - [ ] (S) Circulate within RFC mailing list
### Implementation
  - [ ] (L) Implement ACL auth method schema and store funcs
[...]

PR #13718, opened 2022-07-12
core: add ACL token expiry state, struct, and RPC handling.

jrasell:
The ACL token state schema has been updated to utilise two new
indexes which track expiration of tokens that are configured with
an expiration TTL or time. A new state function allows listing
ACL expired tokens which will be used by internal garbage
collection.

The ACL endpoint has been modified so that all validation happens
within a single function call. This is easier to understand and
see at a glance. The ACL token validation now also includes logic
for expiry TTL and times. The ACL endpoint upsert tests have been
condensed into a single, table driven test.

There is a new token canonicalize which provides a single place
for token canonicalization, rather than logic spread in the RPC
handler.

targets: feature branch

### DIFF
--- a/helper/funcs.go
+++ b/helper/funcs.go
@@ -126,6 +126,8 @@ func StringToPtr(str string) *string {
 }
 
 // TimeToPtr returns the pointer to a time.Duration.
+//
+// Deprecated; use pointer.Of instead.
 func TimeToPtr(t time.Duration) *time.Duration {
 	return &t
 }
--- a/nomad/acl_endpoint_test.go
+++ b/nomad/acl_endpoint_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/hashicorp/nomad/nomad/mock"
 	"github.com/hashicorp/nomad/nomad/structs"
 	"github.com/hashicorp/nomad/testutil"
+	"github.com/shoenig/test/must"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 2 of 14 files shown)


helper/funcs.go
```diff
@@ -130,6 +130,12 @@ func TimeToPtr(t time.Duration) *time.Duration {
 	return &t
 }
 
+// TimeTimeToPtr returns the pointer to a time.Time. Its name is unfortunate,
+// but needed due to the TimeToPtr function.
+func TimeTimeToPtr(t time.Time) *time.Time {
+	return &t
+}
```
shoenig, line 137: there is now `helper/pointer.Of`

nomad/acl_endpoint_test.go
```diff
@@ -1453,85 +1453,175 @@ func TestACLEndpoint_Bootstrap_Reset(t *testing.T) {
...
+				req := &structs.ACLTokenUpsertRequest{
+					Tokens: []*structs.ACLToken{p1},
+					WriteRequest: structs.WriteRequest{
+						Region:    DefaultRegion,
+						AuthToken: aclToken.SecretID,
+					},
+				}
+				var resp structs.ACLTokenUpsertResponse
+				require.NoError(t, msgpackrpc.CallWithCodec(codec, structs.ACLUpsertTokensRPCMethod, req, &resp))
+				require.Greater(t, resp.Index, uint64(0))
```
shoenig, line 1487: leave the type casting in the dust
https://github.com/shoenig/test/blob/main/must/must.go#L235

nomad/structs/acl.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// IsExpired compares the ACLToken.ExpirationTime against the passed t to
+// identify whether the token is considered expired. The function can be called
+// without checking whether the ACL token has an expiry time.
+func (a *ACLToken) IsExpired(t time.Time) bool {
+
+	// This is the fastest exit point and means we don't need to perform any
+	// expiration checks.
+	if t.IsZero() || !a.HasExpirationTime() {
```
shoenig, line 136: Is there a use case for checking `IsExpired` against a 0 time, and expecting the result to be false? Maybe it'd be safer to assume the token is expired in that case?

shoenig (approved):
LGTM; just the one question about expiration, and a couple suggestions 

Also feel free to `s/interface{}/any/g`

PR #13725, opened 2022-07-12
Add semgrep rule to catch non-determinism in FSM

schmichael:
See `message:` in rule for details.

This catches a few uses that should be cleaned up:

```
$ semgrep scan -c .semgrep/fsm_time.yml nomad
... 

nomad/fsm.go
     semgrep.no-time-in-fsm
        time.Now() should not be called from within the Server's FSM. Apply Raft log messages to the
        State Store must be deterministic so that each server contains exactly the same state. Since
        time drifts between nodes, it must be set before the Raft log message is applied so that all
        Raft members see the same timestamp.

        568┆ Launch:    time.Now(),


  nomad/state/state_store.go
     semgrep.no-time-in-fsm
        time.Now() should not be called from within the Server's FSM. Apply Raft log messages to the
        State Store must be deterministic so that each server contains exactly the same state. Since
        time drifts between nodes, it must be set before the Raft log message is applied so that all
        Raft members see the same timestamp.

       4450┆ status.RequireProgressBy = time.Now().Add(status.ProgressDeadline)
       5866┆ iter = memdb.NewFilterIterator(iter, expiredOneTimeTokenFilter(time.Now()))
       6651┆ rootKeyMeta.CreateTime = time.Now()
```

### DIFF
--- a/.semgrep/fsm_time.yml
+++ b/.semgrep/fsm_time.yml
@@ -1,0 +1,26 @@
+rules:
+  - id: "no-time-in-fsm"
+    patterns:
+      - pattern: time.Now()
+
+      # Metric state is local to the server and therefore must use time.
+      - pattern-not-inside: |
+          defer metrics.MeasureSince(..., time.Now())
+
+      # The timetable's whole point is to roughly track timestamps for Raft log
+      # indexes, so it must use time.
+      - pattern-not-inside: |
+          $N.timetable.Witness(log.Index, time.Now().UTC())
+    message: |
+      time.Now() should not be called from within the Server's FSM. Apply Raft
+      log messages to the State Store must be deterministic so that each server
+      contains exactly the same state. Since time drifts between nodes, it must
+      be set before the Raft log message is applied so that all Raft members
+      see the same timestamp.
+    languages:
+      - "go"
+    severity: "WARNING"
+    paths:
+      include:
+        - "nomad/fsm.*"
+        - "nomad/state/state_store.*"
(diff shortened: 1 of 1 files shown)


.semgrep/fsm_time.yml
```diff
@@ -0,0 +1,26 @@
...
+      - pattern: time.Now()
+
+      # Metric state is local to the server and therefore must use time.
+      - pattern-not-inside: |
+          defer metrics.MeasureSince(..., time.Now())
+
+      # The timetable's whole point is to roughly track timestamps for Raft log
+      # indexes, so it must use time.
+      - pattern-not-inside: |
+          $N.timetable.Witness(log.Index, time.Now().UTC())
```
lgfa29, line 13:
```suggestion
      # Metric state is local to the server and therefore must use time.
      - pattern-not-inside: |
          defer metrics.MeasureSince(...)

      # The timetable's whole point is to roughly track timestamps for Raft log
      # indexes, so it must use time.
      - pattern-not-inside: |
          $N.timetable.Witness(...)
```

I think it doesn't really matter what parameters are passed to these functions?

lgfa29 (approved):
Minor suggestion, but LGTM as-is as well.

Currently semgrep only runs on code changes in a PR, so the existing cases would not be a problem 👍

.semgrep/fsm_time.yml
schmichael: Perfect

tgross:
I've fixed the keyring and OTT cases in #13737 and #13733

PR #11167, opened 2021-09-10
Support configurable dynamic port range

a-zagaevskiy:
This changes makes it possible to configure the range that is used for assigning dynamic ports (#8186).

PR #13726, opened 2022-07-12
Backport of core: merge reserved_ports into host_networks into release/1.1.x

### LINKED ISSUE #13505: Reserved port duplicates/conflicts block all allocations

hc-github-team-nomad-core:
## Backport

This PR is auto-generated from #13651 to be assessed for backporting due to the inclusion of the label backport/1.1.x.

---

Fixes #13505

This fixes #13505 by treating reserved_ports like we treat a lot of jobspec settings: merging settings from more global stanzas (`client.reserved.reserved_ports`) "down" into more specific stanzas (`client.host_networks[].reserved_ports`).

As discussed in #13505 there are other options, and since it's totally broken right now we have some flexibility:

1. Treat overlapping `reserved_ports` on addresses as invalid and refuse to start agents. However, I'm not sure there's a cohesive model we want to publish right now since so much 0.9-0.12 compat code still exists! We would have to explain to folks that if their `-network-interface` and `host_network` addresses overlapped, they could only specify reserved_ports in one place or the other?! It gets ugly.
2. Use the global client.reserved.reserved_ports value as the default and treat host_network[].reserverd_ports as overrides. My first suggestion in the issue, but @groggemans made me realize the addresses on the agent's interface (as configured by -network-interface) may overlap with host_networks, so you'd need to _remove_ the global reserved_ports from addresses shared with a shared network?! This seemed really confusing and subtle for users to me.

So I think "merging down" creates the most expressive yet understandable approach. I've played around with it a bit, and it doesn't seem too surprising. The only frustrating part is how difficult it is to _observe_ the available addresses and ports on a node! However that's a job for another PR.

Sorry I snuck in a couple other refactorings. I really want to make this code more maintainable, so I tried to move it in that direction where I didn't think it would be a huge distraction (eg the `interface{} -> string` switch). I can back out any of that if you think it's best to keep this tight and focused.


lgfa29:
@schmichael I had to use the private version of `maxValidPort` because it was only exposed as public in https://github.com/hashicorp/nomad/pull/11167/commits/fc89835dafec78a1369292ebb597548e4e206408.

PR #13733, opened 2022-07-13
keyring: upserting key metadata in FSM must be deterministic

tgross:
Fixes https://github.com/hashicorp/nomad/pull/13725 for the secure variables keyring. This hasn't shipped in a release of Nomad, so that's why it's in its own PR with no changelog.


schmichael (approved):
Looks like NewRootKey was already setting this via leader.go:initializeKeyring and the Rotate RPC? If so :+1:

PR #13737, opened 2022-07-13
fsm: one-time token expiration should be deterministic

tgross:
Fixes https://github.com/hashicorp/nomad/pull/13725 for one-time tokens.

When applying a raft log to expire ACL tokens, we need to use a
timestamp provided by the leader so that the result is deterministic
across servers.


nomad/structs/structs.go
```diff
@@ -12077,6 +12077,7 @@ type OneTimeTokenDeleteRequest struct {
 
 // OneTimeTokenExpireRequest is a request to delete all expired one-time tokens
 type OneTimeTokenExpireRequest struct {
+	Timestamp time.Time
```
schmichael, line 12080: Doesn't this need to get set in `CoreScheduler.expiredOneTimeTokenGC`? I don't see where this gets set, and `core_sched.go` is where the other GC operations compute their threshold based on the timetable/local clock before submitting the deletions via raft.

tgross: It should be set on the `ACL.ExpireOneTimeTokens` RPC, which I missed b/c the tests passed. 

Most of the core jobs use the timetable block specifically because they _don't_ submit the deletions via raft, but via the RPC. While the leader creates the GC eval, it can be processed on any worker (ref [`worker.go#L565-L569`](https://github.com/hashicorp/nomad/blob/main/nomad/worker.go#L565-L569)).

So when I wrote the expire OTT job originally, I hit the same "quiet clusters will have wonky timestamps" problem @jrasell discovered, and so I wanted to use the wall clock on the leader to get a more accurate expiration. But I messed that up by not having the timestamp set in the leader's RPC, but in the FSM instead. So we should be setting that on the leader's handler for the RPC and we'll be all set.

tgross: Fixed in 9d8138cc6c106b8615b750948c10992e8a2fa7c8

schmichael (approved):

PR #13742, opened 2022-07-13
secure vars: updates should reduce quota tracking if smaller

tgross:
When secure variables are updated, we were adding the update to the
existing quota tracking without first checking whether it was an
update to an existing variable. In that case we need to add/subtract
only the difference between the new and existing quota usage.


angrycub (approved):
Looks good

PR #13743, opened 2022-07-13
secure vars: limit maximum size of variable data

tgross:
To discourage accidentally DoS'ing the cluster with secure variables
data, we're providing a very low limit to the maximum size of a given
secure variable. This currently matches the limit for dispatch
payloads.

In future versions, we may increase this limit or make it
configurable, once we have better metrics from real-world operators.


DerekStrickland (approved):
:shipit:

PR #13748, opened 2022-07-13
workload identity: use parent ID for dispatch/periodic jobs

tgross:
Workload identities grant implicit access to policies, and operators
will not want to craft separate policies for each invocation of a
periodic or dispatch job. Use the parent job's ID as the JobID claim.

Tagging @philrenaud just as an FYI, as he noticed this problem.


shoenig (approved):
LGTM!

tgross:
Rebased to fix conflict from https://github.com/hashicorp/nomad/pull/13848

PR #13750, opened 2022-07-13
tests: add a space between node name and timestamp

tgross:
A minor improvement to make server logs in testing outputs more readable, which I ran into while debugging https://github.com/hashicorp/nomad/pull/13749

Before:
> node-2832022-07-13T12:57:04.242Z [INFO]  raft@v1.3.5/raft.go:369: nomad.raft: entering leader state: leader="Node at 127.0.0.1:9268 [Leader]"

After:
> node-283 2022-07-13T12:57:04.242Z [INFO]  raft@v1.3.5/raft.go:369: nomad.raft: entering leader state: leader="Node at 127.0.0.1:9268 [Leader]"


shoenig (approved):
:space_invader:

PR #13754, opened 2022-07-14
[ui] Keyboard nav: "u" key to go up a level

### LINKED ISSUE #13753: `u` to go up a level

philrenaud:
Resolves #13753

Adds a conditional `keyboard-shortcut` modifier to breadcrumbs; always the second-to-last one, should it exist.

Using prior art for conditional modifiers courtesy of https://v5.chriskrycho.com/journal/conditional-modifiers-and-helpers-in-emberjs/

### DIFF
--- a/ui/app/templates/components/app-breadcrumbs.hbs
+++ b/ui/app/templates/components/app-breadcrumbs.hbs
@@ -1,7 +1,7 @@
 <Breadcrumbs as |breadcrumbs|>
-  {{#each breadcrumbs as |crumb|}}
+  {{#each breadcrumbs as |crumb iter|}}
     {{#let crumb.args.crumb as |c|}}
-      {{component (concat "breadcrumbs/" (or c.type "default")) crumb=c}}
+      {{component (concat "breadcrumbs/" (or c.type "default")) crumb=c oneCrumbUp=(eq iter (sub breadcrumbs.length 2))}}
     {{/let}}
   {{/each}}
 </Breadcrumbs>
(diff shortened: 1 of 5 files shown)


ui/.template-lintrc.js
```diff
@@ -12,4 +12,7 @@ module.exports = {
     },
     'no-implicit-this': { allow: ['keyboard-commands'] },
   },
+  ignore: [
+    'app/components/breadcrumbs/*' // using {{(modifier)}} syntax
+  ]
```
philrenaud, line 17: It seems like `{{(modifier ...)}}` is valid conditional syntax, and the absence of `()` in fact breaks this functionality, but the template linter doesn't like it, so ignoring it in the breadcrumb files.

ui/app/templates/components/app-breadcrumbs.hbs
```diff
@@ -1,7 +1,7 @@
 <Breadcrumbs as |breadcrumbs|>
-  {{#each breadcrumbs as |crumb|}}
+  {{#each breadcrumbs as |crumb iter|}}
     {{#let crumb.args.crumb as |c|}}
-      {{component (concat "breadcrumbs/" (or c.type "default")) crumb=c}}
+      {{component (concat "breadcrumbs/" (or c.type "default")) crumb=c isOneCrumbUp=(eq iter (sub breadcrumbs.length 2))}}
```
philrenaud, line 4: Since the active page is always the last breadcrumb, we want the penultimate one. Hence, `(eq iter (sub breadcrumbs.length 2)`

ui/.template-lintrc.js
ChaiWithJai: `blocking`: I don't think the linter was able to pick up on this because the test suite failed to run.

philrenaud:
> Failing tests because its failing prettier. I think .template-lintrc.js isn't ignoring the correct file. Can you take a look at the error message?

@ChaiWithJai Looks like the right files were ignored; I was just missing a trailing comma.  Fixed @ https://github.com/hashicorp/nomad/pull/13754/commits/155514d934e936f7ea5f9471cb1d73d7deeab4ad

PR #13756, opened 2022-07-14
acl: add token expiry checking to ACL token resolution.

jrasell:
This commit adds basic expiry checking when performing ACL token
resolution. This expiry checking is local to each server and does
not at this time take into account potential time skew on server
hosts.

A new error message has been created so clients whose token has
expired get a clear message, rather than a generic token not
found.

The ACL resolution tests have been refactored into table driven
tests, so additions are easier in the future.

This is the simpler of two potential approaches, the second of
which I will raise an issue to track as a potential future
enhancement. This followup includes a new leader loop that watches
for ACL token table changes and maintains a heap of expiring tokens
so that the leader can set expiry authoritatively. There is significant
overhead in this approach that needs to be researched and
considered.

targets: feature branch

### DIFF
--- a/nomad/acl.go
+++ b/nomad/acl.go
@@ -54,6 +54,9 @@ func resolveTokenFromSnapshotCache(snap *state.StateSnapshot, cache *lru.TwoQueu
 		if token == nil {
 			return nil, structs.ErrTokenNotFound
 		}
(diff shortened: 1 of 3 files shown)


lgfa29 (approved):

shoenig (approved):
LGTM

PR #13762, opened 2022-07-14
ui:  fix placement of filter in `SecureVariablesForm`

ChaiWithJai:
This PR resolves an issue where the namespace filter:

1. Always rendered a 2 column grid, regardless of where the filter is rendered.
2. Placement of the namespace filter when the form is in create mode.

Side Effect:  remove unused `related-entities` CSS property. There is no `.related-entities` property that is a standalone CSS class. Maybe it was confused with `.notification.related-entities`?

### DIFF
--- a/ui/app/styles/components/secure-variables.scss
+++ b/ui/app/styles/components/secure-variables.scss
@@ -27,6 +27,8 @@
   }
 
   .path-input {
+    height: 2.25em;
+
     &:disabled {
       background-color: #f5f5f5;
     }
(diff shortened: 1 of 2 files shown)


ui/app/styles/components/secure-variables.scss
```diff
@@ -27,6 +27,8 @@
   }
 
   .path-input {
+    height: 2.25em;
```
ChaiWithJai, line 30: `app/styles/core/forms.scss` sets the default height of `input` to be `auto` which overrides the normal `2.25 em` expected in the `Input` component which is also how `SingleSelectDropdown` has its height set to.

philrenaud (approved):
Works like a charm. Nice work!

PR #13776, opened 2022-07-14
Replace ArrowLeft etc. with an ascii arrow

philrenaud:
image

Replaces "ArrowUp" with `↑` in the modal, and so forth.

### DIFF
--- a/ui/app/helpers/clean-keycommand.js
+++ b/ui/app/helpers/clean-keycommand.js
@@ -1,0 +1,23 @@
+// @ts-check
+import { helper } from '@ember/component/helper';
+
+const KEY_ALIAS_MAP = {
+  ArrowRight: '→',
+  ArrowLeft: '←',
+  ArrowUp: '↑',
+  ArrowDown: '↓',
+  '+': ' + ',
+  // Enter: '⏎',
+  // Tab: '⇥',
+  // Space: '␣',
+  // Shift: '⇧',
+};
+
+export default helper(function cleanKeycommand([part] /*, named*/) {
+  console.log('cleaning', part);
+  Object.keys(KEY_ALIAS_MAP).forEach((key) => {
+    part = part.replace(key, KEY_ALIAS_MAP[key]);
+  });
+
+  return [part];
+});
--- a/ui/app/helpers/clean-keycommand.js
+++ b/ui/app/helpers/clean-keycommand.js
@@ -13,11 +13,10 @@ const KEY_ALIAS_MAP = {
   // Shift: '⇧',
 };
 
-export default helper(function cleanKeycommand([part] /*, named*/) {
-  console.log('cleaning', part);
-  Object.keys(KEY_ALIAS_MAP).forEach((key) => {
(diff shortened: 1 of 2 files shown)


ui/app/helpers/clean-keycommand.js
```diff
@@ -0,0 +1,23 @@
...
+  ArrowDown: '↓',
+  '+': ' + ',
+  // Enter: '⏎',
+  // Tab: '⇥',
+  // Space: '␣',
+  // Shift: '⇧',
+};
+
+export default helper(function cleanKeycommand([part] /*, named*/) {
+  console.log('cleaning', part);
```
ChaiWithJai, line 17: `blocking`:  remove the `log` statement.

ui/app/helpers/clean-keycommand.js
```diff
@@ -0,0 +1,23 @@
...
+  Object.keys(KEY_ALIAS_MAP).forEach((key) => {
+    part = part.replace(key, KEY_ALIAS_MAP[key]);
```
ChaiWithJai, line 19: `suggestion` `non-blocking`:  Generally I'm not a fan of mutating an argument passed a function.

ui/app/helpers/clean-keycommand.js
```diff
@@ -0,0 +1,23 @@
...
+  ArrowUp: '↑',
+  ArrowDown: '↓',
+  '+': ' + ',
+  // Enter: '⏎',
+  // Tab: '⇥',
+  // Space: '␣',
+  // Shift: '⇧',
+};
+
+export default helper(function cleanKeycommand([part] /*, named*/) {
```
ChaiWithJai, line 16: `question`:  Why do we pass in a `key` in the template, but we name the argument `part`?

ui/app/helpers/clean-keycommand.js
```diff
@@ -0,0 +1,23 @@
...
+  console.log('cleaning', part);
+  Object.keys(KEY_ALIAS_MAP).forEach((key) => {
+    part = part.replace(key, KEY_ALIAS_MAP[key]);
+  });
+
+  return [part];
```
ChaiWithJai, line 22: `question`:  This returns a list that will be rendered inside of a `span`. If we're returning a list, shouldn't invoke this template helper and assign the return value to a variable and then `each` over that?

ChaiWithJai (approved):

ui/app/helpers/clean-keycommand.js
philrenaud: The key might be a sequence, like `["g", "v"]` for `go to variables`. `g` and `v` would be individual `part`s.

philrenaud: Hm you know what? You bring up a good point here. I'm using too many words to mean roughly the same thing. Will bring them all down to `part`.

ui/app/helpers/clean-keycommand.js
ChaiWithJai: Wow! So checkout how fail-safe Handlebars is.

image

PR #13781, opened 2022-07-15
Keyboard Nav: let users rebind their shortcuts

philrenaud:
Saves shortcuts to localStorage, and gives a small interface for rebinding keys.

image

Side-effects:
- adds autofocus and click-outside-dismiss to the keyboard modal
- Modifies the Keynav modal header styles
- Adds general enable/disable for keynav

### DIFF
--- a/ui/app/components/keyboard-shortcuts-modal.js
+++ b/ui/app/components/keyboard-shortcuts-modal.js
@@ -21,7 +21,12 @@ export default class KeyboardShortcutsModalComponent extends Component {
    */
   @computed('keyboard.keyCommands.[]')
   get commands() {
-    return this.keyboard.keyCommands.filter((c) => c.label && c.action);
+    return this.keyboard.keyCommands.reduce((memo, c) => {
+      if (c.label && c.action && !memo.find((m) => m.label === c.label)) {
+        memo.push(c);
+      }
+      return memo;
+    }, []);
   }
 
   /**
--- a/ui/app/services/keyboard.js
+++ b/ui/app/services/keyboard.js
@@ -41,6 +41,19 @@ const DIGIT_MAP = {
   ')': 0,
 };
 
+const DISALLOWED_KEYS = [
+  'Shift',
+  'Backspace',
+  'Delete',
+  'Meta',
+  'Alt',
+  'Control',
+  'Tab',
+  'CapsLock',
+  'Clear',
+  'ScrollLock',
+];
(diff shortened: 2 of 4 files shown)


ui/app/services/keyboard.js
```diff
@@ -42,122 +66,121 @@ export default class KeyboardService extends Service {
...
+      {
+        label: 'Show Keyboard Shortcuts',
+        action: () => {
+          this.shortcutsVisible = true;
+        },
+      },
+    ].map((command) => {
+      const persistedValue = window.localStorage.getItem(
+        `keyboard.command.${command.label}`
+      );
```
philrenaud, line 174: Dealing with localStorage directly in the service is the result of trying to use the utils/local-storage computed for awhile to no avail. This gives more control.

ui/app/components/keyboard-shortcuts-modal.js
```diff
@@ -49,4 +55,8 @@ export default class KeyboardShortcutsModalComponent extends Component {
   untetherFromElement(self, _, { hint }) {
     hint.binder.destroy();
   }
+
+  @action toggleListener() {
+    this.keyboard.enabled = !this.keyboard.enabled;
+  }
```
ChaiWithJai, line 61: `question`: Are we using the `localStorageProperty` API correctly here? I notice we're using this action instead of the `set` method on the API. Does this save to `localStorage` correctly?

ui/app/services/keyboard.js
ChaiWithJai: Yup, I noticed that in the `action` for `enable` you're using, as well. I think you might want to read how the `utils/local-storage` works.

I think because `localStorageProperty` is a `computed` property -- which always makes me think that the behavior isn't synchronous which is a problem inside of a callback in a `map` function.

ChaiWithJai (approved):
Looks like we weren't able to use our `localStorage` utility method in most of these places.

Maybe we can see if we can use some [prior art](https://github.com/jelhan/ember-local-storage-decorator/blob/master/addon/index.js) to upgrade our implementation at some point.

ui/app/components/keyboard-shortcuts-modal.js
philrenaud: Yep! Setting a property in this way is an implicit `set` for computed macros. I think they probably use JS proxies under the hood.

PR #13784, opened 2022-07-15
Update README.md

TLC-10:
Updating to use a list for readability


tgross:
Hi! Thanks, but I think we're pretty happy with the format!

PR #13786, opened 2022-07-15
metrics: even classless blocked evals get metrics

### LINKED ISSUE #13759: When there are blocked evaluations, nomad.nomad.blocked_evals.[cpu,memory] are always 0
### Nomad version
Nomad v1.3.0

Nomad running on Debian on AWS

### Issue
I'm trying to use `nomad.nomad.blocked_evals.memory` as information to pass to the Nomad autoscaler to scale out the cluster, but the metric is _always_ unset.

Ex.

```
$ curl localhost:4646/v1/metrics?format=prometheus 2>/dev/null | grep "blocked_evals"
# HELP nomad_nomad_blocked_evals_job_cpu nomad_nomad_blocked_evals_job_cpu
# TYPE nomad_nomad_blocked_evals_job_cpu gauge
nomad_nomad_blocked_evals_job_cpu{job="example",namespace="default"} 0
# HELP nomad_nomad_blocked_evals_job_memory nomad_nomad_blocked_evals_job_memory
# TYPE nomad_nomad_blocked_evals_job_memory gauge
nomad_nomad_blocked_evals_job_memory{job="example",namespace="default"} 3000
# HELP nomad_nomad_blocked_evals_total_blocked nomad_nomad_blocked_evals_total_blocked
# TYPE nomad_nomad_blocked_evals_total_blocked gauge
nomad_nomad_blocked_evals_total_blocked 1
# HELP nomad_nomad_blocked_evals_total_escaped nomad_nomad_blocked_evals_total_escaped
# TYPE nomad_nomad_blocked_evals_total_escaped gauge
nomad_nomad_blocked_evals_total_escaped 0
# HELP nomad_nomad_blocked_evals_total_quota_limit nomad_nomad_blocked_evals_total_quota_limit
# TYPE nomad_nomad_blocked_evals_total_quota_limit gauge
nomad_nomad_blocked_evals_total_quota_limit 0
```

It looks like the per-job stats are being populated, but the per-node stats are not being populated. Is there something I'm missing here, since per-node stats are enabled in the `telemetry` block and I'd expect these values to be non-zero if there are blocked allocations.

### Reproduction steps

Try and deploy a Nomad job with a memory resource request that exceeds what is available in a cluster.

#### Expected Result
Non-zero metrics

#### Actual Result
Empty metrics

shoenig:
This PR fixes a bug where blocked evaluations with no class set would
not have metrics exported at the dc:class scope.

Fixes #13759


nomad/blocked_evals_stats.go
```diff
@@ -89,6 +89,10 @@ func generateResourceStats(eval *structs.Evaluation) *BlockedResourcesStats {
 		for class := range allocMetrics.ClassExhausted {
 			classes[class] = struct{}{}
 		}
+		if len(allocMetrics.ClassExhausted) == 0 {
+			// some evaluations have no class
```
lgfa29, line 93: Rude 😂

lgfa29 (approved):

PR #13805, opened 2022-07-18
core: add expired token garbage collection periodic jobs.

jrasell:
Two new periodic core jobs have been added which handle removing
expired local and global tokens from state. The local core job is
run on every leader; the global core job is only run on the leader
within the authoritative region.

targets: feature branch

### DIFF
--- a/nomad/core_sched.go
+++ b/nomad/core_sched.go
@@ -51,6 +51,10 @@ func (c *CoreScheduler) Process(eval *structs.Evaluation) error {
 		return c.csiPluginGC(eval)
 	case structs.CoreJobOneTimeTokenGC:
 		return c.expiredOneTimeTokenGC(eval)
+	case structs.CoreJobLocalTokenExpiredGC:
+		return c.expiredACLTokenGC(eval, false)
+	case structs.CoreJobGlobalTokenExpiredGC:
+		return c.expiredACLTokenGC(eval, true)
 	case structs.CoreJobForceGC:
 		return c.forceGC(eval)
 	default:
@@ -78,6 +82,13 @@ func (c *CoreScheduler) forceGC(eval *structs.Evaluation) error {
 	if err := c.expiredOneTimeTokenGC(eval); err != nil {
 		return err
 	}
+	if err := c.expiredACLTokenGC(eval, false); err != nil {
+		return err
+	}
+	if err := c.expiredACLTokenGC(eval, true); err != nil {
+		return err
+	}
+
 	// Node GC must occur after the others to ensure the allocations are
 	// cleared.
 	return c.nodeGC(eval)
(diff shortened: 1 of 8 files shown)


nomad/core_sched.go
```diff
@@ -773,6 +791,82 @@ func (c *CoreScheduler) expiredOneTimeTokenGC(eval *structs.Evaluation) error {
 	return c.srv.RPC("ACL.ExpireOneTimeTokens", req, &structs.GenericResponse{})
 }
 
+// expiredACLTokenGC handles running the garbage collector for expired ACL
+// tokens. It can be used for both local and global tokens and includes
+// behaviour to account for periodic and user actioned garbage collection
+// invocations.
+func (c *CoreScheduler) expiredACLTokenGC(eval *structs.Evaluation, global bool) error {
+
+	expiryThreshold := c.getThreshold(eval, "expired_acl_token",
```
shoenig, line 800: maybe `expiryThresholdIdx` - I thought this was a duration at first

nomad/core_sched.go
```diff
@@ -773,6 +791,82 @@ func (c *CoreScheduler) expiredOneTimeTokenGC(eval *structs.Evaluation) error {
...
+	// however, as the caller, we do not know at which point in the array the
+	// tokens are no longer expired. This time therefore forms the basis at
+	// which we draw the line in the iteration loop and find the final expired
+	// token that is eligible for deletion.
+	now := time.Now().UTC()
+
+	for raw := expiredIter.Next(); raw != nil; raw = expiredIter.Next() {
+		token := raw.(*structs.ACLToken)
+
+		// The indexes mean if we come across an unexpired token, we can exit
```
shoenig, line 823:
```suggestion
		// The iteration order of the indexes mean if we come across an unexpired token, we can exit
```

nomad/core_sched.go
```diff
@@ -773,6 +791,82 @@ func (c *CoreScheduler) expiredOneTimeTokenGC(eval *structs.Evaluation) error {
...
+		// as we have found all currently expired tokens.
+		if !token.IsExpired(now) {
+			break
+		}
+
+		// Check whether the token is old enough to be deleted and skip if it
+		// is not.
```
shoenig, line 830: The phrasing of the comment is opposite of the code - would rather read something like, "Check if the token is recent enough to skip, otherwise we'll delete it"

nomad/leader.go
```diff
@@ -762,59 +773,89 @@ func (s *Server) schedulePeriodic(stopCh chan struct{}) {
...
 
+// schedulePeriodicAuthoritative is a long-lived routine intended for use on
+// the leader within the authoritative region only. It periodically queues work
+// onto the _core scheduler for ACL based activities such as removing expired
+// global ACL tokens.
+func (s *Server) schedulePeriodicAuthoritative(stopCh chan struct{}) {
+
+	// Set up the expired ACL global token garbage collection ticker.
+	globalTokenExpiredGC := time.NewTicker(s.config.ACLTokenExpirationGCInterval)
+	defer globalTokenExpiredGC.Stop()
```
shoenig, line 833: this is fine but I like helper.NewSafeTimer better

- better enforces the caller to handle the Stop
- using `.Reset(interval)` means the actual sleep time between invocations is the specified `interval`, rather than what is effectively `interval - enqueue time`.

nomad/leader_test.go
```diff
@@ -3,6 +3,7 @@ package nomad
 import (
 	"errors"
 	"fmt"
+	"github.com/shoenig/test/must"
 	"sort"
 	"strconv"
 	"testing"
```
shoenig, line 9: Are you using GoLand? I swear recently stopped automatically correctly sorting imports by stdlib vs. 3rdparty

shoenig (approved):
LGTM!; just the usual ignore-able suggestions

shoenig (approved):
LGTM!

PR #13813, opened 2022-07-18
docs: move checks into own page

shoenig:
This PR creates a top-level `check` page for job-specification docs.

The content for checks is about half the content of the service page, and
is about to increase in size when we add docs about Nomad service checks.
Seemed like a good idea to just split the checks section out into its own
thing (e.g. check_restart is already a topic).

Not adding Nomad service check stuff yet lets us backport this change

Mostly just a lift-and-shift but with some tweaked examples to de-emphasize
the use of script checks.

### DIFF
--- a/website/content/docs/job-specification/check.mdx
+++ b/website/content/docs/job-specification/check.mdx
@@ -1,0 +1,363 @@
+---
+layout: docs
+page_title: check Block - Job Specification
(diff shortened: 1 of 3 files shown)


schmichael (approved):
Thank you for doing this! Long overdue!

PR #13820, opened 2022-07-18
SV CLI: var init

angrycub:
Command to create a skeleton secure variable specification

- Nomad dep: add museli/reflow
- SV CLI: var init

### DIFF
--- a/command/var_init.go
+++ b/command/var_init.go
@@ -1,0 +1,212 @@
+package command
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"regexp"
+	"strings"
+
+	"github.com/muesli/reflow/wordwrap"
+	"github.com/posener/complete"
+)
+
+const (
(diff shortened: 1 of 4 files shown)


command/var_init.go
```diff
@@ -0,0 +1,212 @@
...
+	var out strings.Builder
+	out.WriteString(ss[0])
+
+	// reassemble and split the wraps taking the indent into account
+	input = strings.Join(ss[1:], "\n")
+	ss = strings.Split(wordwrap.String(input, lineLen-indent), "\n")
+	prefixStringList(ss, strings.Repeat(" ", indent))
+
+	return strings.Join(ss, "\n")
+}
```
tgross, line 196: This function is unused. Is this to support an incoming PR? (In which case, that's fine to leave)

tgross (approved):
LGTM!

command/var_init.go
angrycub: Pulling it for now.

PR #13825, opened 2022-07-18
patch namespace related issue to saving and querying

ChaiWithJai:


ui/app/templates/variables/variable/index.hbs
```diff
@@ -36,7 +36,7 @@
   </div>
   <div>
     {{#unless this.isDeleting}}
-      {{#if (can "write variable" path=this.model.absolutePath)}}
+      {{#if (can "write variable" path=this.model.path)}}
```
philrenaud, line 39: Out of curiosity, why `absolutePath` to `path` ? absolutePath includes everything from root on forward, no?

philrenaud (approved):

ui/app/templates/variables/variable/index.hbs
ChaiWithJai: `absolutePath` does not exist on the model. The `variables.variable.index` route does not return multiple models by creating the `PathTree` presumably because `variables.variable.index` is displaying information about a single variable and not listing folder paths (where I believe that `absolutePath` logic is required).

Issue #13352, opened 2022-06-13
[CNI] network mode bridge doesn't allow hairpin


ygersie:
When spinning up a job that uses CNI to setup the forwarding the container can't reach itself on the host port. This probably isn't a very common use case but when deploying a container that needs to discover itself and its peers through a Consul endpoint we get back the host IP + Port including itself. The connection towards the endpoint that references itself will then not work and gives timeouts. You can reproduce with the following job:
```hcl
job "ygersie" {
  datacenters = ["dc1"]
  namespace   = "default"

  group "example" {
    count = 1

    network {
      mode = "bridge"
      port "foo" {
        to = 1337
      }
    }

    task "example" {
      driver = "docker"
      config {
        image = "alpine"
        args  = ["nc", "-lk", "-p", "${NOMAD_PORT_foo}", "-e", "cat"]
      }

      resources {
        cpu    = 100
        memory = 64
      }
    }
  }
}
```
And then from the container a netcat times out:
```sh
/ # nc -v ${NOMAD_HOST_IP_foo} ${NOMAD_HOST_PORT_foo}
```
I compiled a version of Nomad with [hairpinMode](https://www.cni.dev/plugins/v1.1/main/bridge/#network-configuration-reference) enabled in the [nomadCNIConfigTemplate](https://github.com/hashicorp/nomad/blob/main/client/allocrunner/networking_bridge_linux.go#L141) which resolves the issue.

Can this be made either configureable or enabled by default or is there any particular reason why I wouldn't want this?

DerekStrickland:
Hi @ygersie 

Thanks for reporting this issue, and for providing a potential solution!  I'll take a look at what you've got here and then discuss where it might fit in the roadmap. Also, feel free to submit a PR to add the configuration. You might get to it more quickly than we do, and community PRs are always welcome!

DerekStrickland:
I don't know that we could set it as the default. Some quick research seems to indicate that not all CNI implementations support it. I think the PR would need to default to false but allow user configuration to enable hairpin mode.

ygersie:
@DerekStrickland it may not be supported by all CNI implementations but this is specifically the one used to setup:
```
network {
  mode = "bridge"
}
```
which is used to setup port forwarding using the CNI plugins and also required when using Consul Connect. Afaik it won't affect any other (user supplied) configurations. But if there might be other issues then it definitely needs to be configurable. Please let me know what you think.

jrasell:
[...]

PR #13834, opened 2022-07-19
allow configuring hairpinning on default nomad bridge

A-Helberg:
[Hairpinning](https://en.wikipedia.org/wiki/Network_address_translation#NAT_hairpinning) is required by a lot of clustering applications (eg. Grafana Loki).
These applications connect to all members of their cluster, including themselves.
Currently they cannot connect to themselves as hairpin-mode is not enabled on the default nomad bridge.

As suggested in https://github.com/hashicorp/nomad/issues/13352 I've added a configuration option that allows a user to enable hairpinning on the default nomad cni bridge.

### DIFF
--- a/website/content/docs/configuration/client.mdx
+++ b/website/content/docs/configuration/client.mdx
@@ -142,6 +142,9 @@ client {
 - `bridge_network_subnet` `(string: "172.26.64.0/20")` - Specifies the subnet
   which the client will use to allocate IP addresses from.
 
+- `bridge_network_hairpin` `(bool: false)` - Specifies wether the bridge should
+  allow hairpinning.
+
 - `artifact` <code>([Artifact](#artifact-parameters): varied)</code> -
   Specifies controls on the behavior of task
   [`artifact`](/docs/job-specification/artifact) stanzas.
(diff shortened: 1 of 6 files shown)


A-Helberg:
I've just finished testing in a staging cluster I have access to and it seems to work as intended.

Please let me know if there's anything I should change or update.

website/content/docs/configuration/client.mdx
```diff
@@ -142,6 +142,9 @@ client {
 - `bridge_network_subnet` `(string: "172.26.64.0/20")` - Specifies the subnet
   which the client will use to allocate IP addresses from.
 
+- `bridge_network_hairpin` `(bool: false)` - Specifies wether the bridge should
```
tgross, line 145:
```suggestion
- `bridge_network_hairpin` `(bool: false)` - Specifies whether the bridge should
```

A-Helberg:
Thanks for the feedback!

I'll address both comments asap.

The templating idea sounds like a better long term solution. It would decouple nomad from any future changes in the cni.

On the other hand if you decide to go with this pr, I'm not familiar with fingerprinting for nomad. I would appreciate a file to start working from.

krundru (approved):

lgfa29:
Hi @A-Helberg 👋 

We ended up missing this PR and re-implemented this feature in https://github.com/hashicorp/nomad/pull/15961, so I'm going to close this. Apologies for that, but thank you so much for all the work!

Issue #12792, opened 2022-04-27
Batch Job Priority is ignored


Fuco1:
### Nomad version
Nomad v1.2.6 (a6c6b475db5073e33885377b4a5c733e1161020c)

### Operating system and Environment details

Linux ubuntu in Azure.

### Issue

With batch jobs, I would expect the allocations from higher priority job to be placed first.

In fact, I've just rewrote a bunch of our infra related to batch job submission only to find out this is not the case :man_facepalming:  Is there any way to make sure the batch job allocations with higher priority are placed first? 

We're running 100k+ task groups doing time series forecasting but I really need to be able to just run it and forget it and be sure that what goes in first comes out first, otherwise I run 10 jobs and instead of waiting 1 hours for the first to finish, I wait 10 hours for all of them to finish.

See also https://github.com/hashicorp/nomad/issues/9845

### Reproduction steps

Submit two batch jobs with different priorities.

#### Expected Result

Allocations from higher priority job are placed first.

#### Actual Result

Allocations are placed at random.

lgfa29:
Hi @Fuco1 👋 

The `priority` value defines two things:

1. the order in which evaluations are dequeued for scheduling 
2. the order of importance of jobs when the cluster is in resource contention

The doc page you linked refer to `1`. The Nomad leader has an internal queue of evaluations that need to be processed in order to, among other things, create and schedule allocations. A job with a high priority will "skip" this queue and jump to the front to be processed before lower priority jobs.

What this means is that this evaluation will be _processed_ first, but that doesn't mean it will _start_ first. Different jobs, clients, allocations etc. have different characteristics, so the allocation will take different time to start.

If your cluster is in a relatively healthy state, this queue will be pretty small, so the priority value won't have much effect. You can monitor the `nomad.nomad.plan.queue_depth` [metric](https://www.nomadproject.io/docs/operations/metrics-reference#key-metrics) to get a sense of what's going on in the queue.

For `2`, that's the what's described in the [comment in the issue you linked](https://github.com/hashicorp/nomad/issues/9845#issuecomment-762857009). When the Nomad cluster is low on resources and you try to schedule something that won't fit, Nomad will use the job priority to determine which allocations can be stopped to make room for the new ones. This is what's called [preemption](https://www.nomadproject.io/docs/internals/scheduling/preemption).

[...]

Issue #9845, opened 2021-01-18
[Bug] Batch Job Priority is ignored


jonatasfreitasv:
## Nomad Version: 1.0.2

Job.Priority attr not affect a true Priority...when I put 70 for one and 100 for other, the schedule is randomize, 
and don't put the biggest priority to run first.

``` json
      {
        "Job": {
          "ID": "${(j \ "execution_uuid").as[String].toLowerCase.substring(0, 8)}.${(j \ "job_uuid").as[String]}",
          "Name": "${(j \ "execution_uuid").as[String].toLowerCase.substring(0, 8)}.${(j \ "job_uuid").as[String]}",
          "Type": "batch",
          "Priority": ${(j \ "metadata" \ "strategy_source" \ "nomad_job_priority").as[Int]},
          "Meta": {
            "execution_uuid": "${(j \ "execution_uuid").as[String]}",
            "sku_parent_color_key": "${(j \ "sku_key").as[String]}"
          },
          "Datacenters": [
            "a",
            "b",
            "c"
          ],
          "TaskGroups": [
            {
              "Name": "${(j \ "execution_uuid").as[String]}",
              "Meta": {
                "execution_uuid": "${(j \ "execution_uuid").as[String]}",
                "job_uuid": "${(j \ "job_uuid").as[String]}",
                "external_uuid": "${(j \ "external_uuid").as[String]}"
              },
              "Count": 1,
              "RestartPolicy":{
                    "Attempts": 3,
                    "Delay": 30000000000,
                    "Mode": "fail"
              },
              "Tasks": [
                {
                  "Name": "${(j \ "execution_uuid").as[String].toLowerCase.substring(0, 8)}.${(j \ "job_uuid").as[String]}",
                  "Meta": {
                      "execution_uuid": "${(j \ "execution_uuid").as[String]}",
                      "job_uuid": "${(j \ "job_uuid").as[String]}",
                      "external_uuid": "${(j \ "external_uuid").as[String]}"
                  },
                  "Driver": "docker",
                  "Config": {
                    "dns_servers": [
                      "8.8.8.8",
                      "8.8.4.4"
                    ],
                    "args": ["main.py"],
                    "command": "python",
                    "image": "${(j \ "metadata" \ "strategy_source" \ "model_docker_image").as[String]}",
                    "cpu_hard_limit": false
                  },
                  "env": {
                    "SOURCE_EVENT": "${sourceEvent}"
                  },
                  "kill_timeout": "600",
[...]

PR #13835, opened 2022-07-19
docs: job-specification: Explain that priority has no effect on run order

nh2:
PRing as suggested in https://github.com/hashicorp/nomad/issues/9845#issuecomment-1187324137 (CC @tgross).

Makes the issues from #9845 and #12792 less surprising to the user.

I think it would make Nomad much better to start batch jobs in prority order (see #12792), but at least this documents that it currently doesn't.

### DIFF
--- a/website/content/docs/job-specification/job.mdx
+++ b/website/content/docs/job-specification/job.mdx
@@ -105,6 +105,8 @@ job "docs" {
 - `priority` `(int: 50)` - Specifies the job priority which is used to
   prioritize scheduling and access to resources. Must be between 1 and 100
   inclusively, with a larger value corresponding to a higher priority.
+  Priority currently only has an effect when job preemption is enabled.
+  It does not have an effect on which of multiple pending jobs is run first.
 
 - `region` `(string: "global")` - The region in which to execute the job.
 


website/content/docs/job-specification/job.mdx
```diff
@@ -105,6 +105,8 @@ job "docs" {
 - `priority` `(int: 50)` - Specifies the job priority which is used to
   prioritize scheduling and access to resources. Must be between 1 and 100
   inclusively, with a larger value corresponding to a higher priority.
+  Priority currently only has an effect when job preemption is enabled.
+  It does not have an effect on which of multiple pending jobs is run first.
```
tgross, line 109: Where possible, we should avoid language in the docs that hints at a roadmap, because then we end up needing to keep the language in sync. (Ex. if we decide we're definitely never going to support it)

```suggestion
  Priority only has an effect when job preemption is enabled.
  It does not have an effect on which of multiple pending jobs is run first.
```

nh2: Makes sense, thanks!

tgross (approved):

tgross:
@nh2 I've merged this and that will trigger a "backport of docs" workflow to the various versioned docs. If you get a ping from any of those PRs b/c the backport failed, don't worry about it... the fix is really annoying so I'll pick that up for you. Thanks again for the PR!

Issue #13844, opened 2022-07-19
`nomad job run -json` panics after successful registration of a periodic job when `Job.Periodic.SpecType` is not defined


tristanpemble:
### Nomad version
Nomad v1.3.1

### Operating system and Environment details
Darwin 21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:37 PDT 2022; root:xnu-8020.121.3~4/RELEASE_ARM64_T6000 arm64 arm unknown Darwin

### Issue
When running an API formatted job file (JSON job) via `nomad job run -json`, and the job has a periodic spec, job registration succeeds, and then the command panics with a SIGSEGV.

### Reproduction steps
Create an API job with a `Periodic.Spec`, but do not define `Periodic.SpecType`. Run it with `nomad job run -json`.

#### Expected Result
The command completes successfully.

#### Actual Result
The command panics with a SIGSEGV

### Job file (if appropriate)

```
{
  "Datacenters": [
    "dc1"
  ],
  "ID": "foo",
  "Name": "foo",
  "Periodic": {
    "Spec": "* * * * * *"
  },
  "TaskGroups": [
    {
      "Name": "foo",
      "Tasks": [
        {
          "Config": {
            "args": [
              "hello world"
            ],
            "command": "echo"
          },
          "Driver": "raw_exec",
          "Name": "foo"
        }
      ]
    }
  ],
  "Type": "batch"
}
```

### Nomad Server logs (if appropriate)
```
    2022-07-19T08:06:51.937-0700 [DEBUG] http: request complete: method=PUT path=/v1/jobs duration="353µs"
```

### Nomad Client logs (if appropriate)
```
Job registration successful
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x101208e98]

goroutine 1 [running]:
github.com/hashicorp/nomad/api.(*PeriodicConfig).Next(0x140007e14a0, {0x37ea8788, 0xeda68bf8b, 0x0})
	github.com/hashicorp/nomad/api@v0.0.0-20220407202126-2eba643965c4/jobs.go:788 +0x28
github.com/hashicorp/nomad/command.(*JobRunCommand).Run(0x14000724a00, {0x1400004e0a0, 0x2, 0x2})
	github.com/hashicorp/nomad/command/job_run.go:355 +0x1044
github.com/mitchellh/cli.(*CLI).Run(0x140003d7680)
[...]

PR #13845, opened 2022-07-19
fix(#13844): canonicalize job to avoid nil pointer deference

tristanpemble:
I am new to this codebase, so unsure I have the best approach, but this fixes my issue in #13844. I traced the issue down to the nil pointer [here](https://github.com/hashicorp/nomad/blob/208b68221184fe5712272c13f4f307373fb9117e/api/jobs.go#L792), which is called [from this path](https://github.com/hashicorp/nomad/blob/208b68221184fe5712272c13f4f307373fb9117e/command/job_run.go#L349-L355).

This code is called after the job registration succeeds on the server. I believe the root of the issue is that when submitting an API job via the CLI, it is canonicalized server-side, but not client-side. Registration succeeds on the server, and then this code path is called. It references paths that have not been canonicalized with their default values, and fails.

Another option would be to canonicalize the job before registration (maybe in the `JobGetter` when loading the job?), but being new to the codebase, I am wary of the unintended change in behavior that might create for endusers.

### DIFF
--- a/command/job_run.go
+++ b/command/job_run.go
@@ -349,6 +349,9 @@ func (c *JobRunCommand) Run(args []string) int {
 
 	evalID := resp.EvalID
 
+	// #13844: canonicalize the job in case it was a partial API definition
+	job.Canonicalize()
+
 	// Check if we should enter monitor mode
 	if detach || periodic || paramjob || multiregion {
 		c.Ui.Output("Job registration successful")
--- a/command/job_run.go
+++ b/command/job_run.go
@@ -349,9 +349,6 @@ func (c *JobRunCommand) Run(args []string) int {
 
 	evalID := resp.EvalID
 
-	// #13844: canonicalize the job in case it was a partial API definition
-	job.Canonicalize()
-
 	// Check if we should enter monitor mode
 	if detach || periodic || paramjob || multiregion {
 		c.Ui.Output("Job registration successful")
--- a/.changelog/13845.txt
+++ b/.changelog/13845.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+api: Fixed a nil pointer dereference when periodic jobs are missing their periodic spec
+```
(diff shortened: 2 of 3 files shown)


Amier3:
Hey @tristanpemble 

Thanks for making this PR and congrats on the first contribution. We'll take a look at this soon 😄

command/job_run.go
```diff
@@ -345,6 +345,9 @@ func (c *JobRunCommand) Run(args []string) int {
 
 	evalID := resp.EvalID
 
+	// #13844: canonicalize the job in case it was a partial API definition
```
tgross, line 348: Just FYI, there's no need to have the issue number in the comments. That's what `git blame` is for 😀

tgross (changes requested):
Hi @tristanpemble! Thanks for this PR!

The fix you've got here will totally work for the case of the CLI, but it'll still be broken for other consumers of the HTTP API. The panic you've hit in https://github.com/hashicorp/nomad/issues/13844 is at [`api/jobs.go#L788`](https://github.com/hashicorp/nomad/blob/v1.3.1/api/jobs.go#L788). The `p` pointer is nil, and that causes the panic when we try to access one of its fields. But in this case, we have a safe fallback behavior later in the `Next` method. So a fix like the diff below should fix both the CLI panic you've hit and prevent anyone else from hitting the same bug if they don't call `Canonicalize()`:

```diff
-if *p.SpecType == PeriodicSpecCron {
+if p != nil && *p.SpecType == PeriodicSpecCron {
```

If you can make this change I'd be happy to get this PR merged. You can also add a changelog entry file at `.changelog/13845.txt` with the following text:

    ```improvement
    api: Fixed a bug where accessing the next periodic invocation for a job could panic if not set
    ```

tgross:
Hi! :wave: I want to get this PR landed in the upcoming Nomad 1.5.0, so I've pushed a commit with the requested fix and a changelog entry. Once that's green I'll get this merged.

tgross (approved):

PR #13849, opened 2022-07-19
keyring: use nanos for `CreateTime` in key metadata

tgross:
Most of our objects use int64 timestamps derived from `UnixNano()` instead of
`time.Time` objects. Switch the keyring metadata to use `UnixNano()` for
consistency across the API.


DerekStrickland (approved):
LGTM

PR #13857, opened 2022-07-20
[CI-only] Use pattern matching for release_branches

mdeggies:
### Description
Pattern matching was [recently added](https://github.com/hashicorp/crt-orchestrator/pull/51) so that teams no longer have to explicitly list every branch that should trigger the CRT pipeline. This simplifies release preparation- anytime a new release branch is created, it will produce releasable artifacts and exercise the full pipeline.

This has been tested in multiple projects since being rolled out. There are no nomad-specific tests that need to be done.

### Links
PR where this functionality was added: https://github.com/hashicorp/crt-orchestrator/pull/51

### PR Checklist

* [ ] updated test coverage
* [ ] external facing docs updated
* [X] not a security concern

### DIFF
--- a/.release/ci.hcl
+++ b/.release/ci.hcl
@@ -13,10 +13,7 @@ project "nomad" {
 
     release_branches = [
       "main",
-      "release/1.0.x",
-      "release/1.1.x",
-      "release/1.2.x",
-      "release/1.3.x",
+      "release/**",
     ]
   }
 }


.release/ci.hcl
```diff
@@ -13,10 +13,7 @@ project "nomad" {
 
     release_branches = [
       "main",
-      "release/1.0.x",
-      "release/1.1.x",
-      "release/1.2.x",
-      "release/1.3.x",
+      "release/**",
```
tgross, line 16: Won't this trigger release workflows on dead branches? Ex. https://github.com/hashicorp/nomad/branches/all?query=release%2F

mdeggies: The pipeline won't be exercised until a new commit is made to the branch. When I say pipeline, I mean all of the common workflows that run after a commit is made on a release branch (not including the promotions to staging/production, which still require a manual trigger using `bob`).

mdeggies:
Can I get an approval please? Thanks!

tgross (approved):
LGTM!

kpenfound (approved):
🚀

PR #13859, opened 2022-07-20
cleanup: example refactoring out map[string]struct{} using set.Set

shoenig:
This PR is a little experiment / demo of using [go-set](https://github.com/hashicorp/go-set) to replace the use of `map[T]struct{}` as a make-shift set.


tgross (approved):
LGTM

PR #13869, opened 2022-07-20
servicedisco: ensure service uniqueness in job validation 

### LINKED ISSUE #13814: servicedisco: unable to register services of the same name but different tags in the same group
I suspect we might be storing services of a group in a `map[service.name]` somewhere, when that isn't enough to ensure uniqueness (i.e. instead use `service.Hash()`).

E.g.

```hcl
job "simple" {
  datacenters = ["dc1"]
  type        = "service"

  group "database" {

    service {
      name     = "db"
      tags     = ["one"]
      provider = "nomad"
    }

    service {
      name     = "db"
      tags     = ["two"]
      provider = "nomad"
    }

    task "db" {
      driver = "raw_exec"
      config {
        command = "bash"
        args    = ["-c", "sleep 15000"]
      }
    }
  }
}
```

only one instance of the service shows up 
```
$ nomad service info db
Job ID     Address  Tags   Node ID   Alloc ID
simple_lb  <none>   [two]  1a9a9fe9  37058c6a
```

and 
```
$ nomad service list
Service Name  Tags
db            [two]
```

shoenig:
87ef5178d1111b092ffa86be46b0caaff8638027: some cleanup to use `set` in the `validateServices` function 

0cc18331058f7ff8d14907f2f71d0870d4515e87: track uniqueness of services in `validateServices` function 

Fixes #13814

Closes #13836 - This PR went in the other direction - to _support_ groups with services that had the same <name, task|group, port_label>. The change is gnarly - basically we would need to add something like a hash of the tags of a service to the unique ID generated by `MakeAllocationServiceID` - which is then used as an index in the state store. All services with tags would be given new IDs, causing unwanted churn in large clusters. The benefit doesn't come close to justifying the means, so I'm going to abandon the feature and just validate the intrinsic invariant we already have (see: this PR).

Spot check:

example validation on identical services with port label set
```
Error submitting job: Unexpected response code: 500 (1 error occurred:
	* Task group database validation failed: 1 error occurred:
	* Services are not unique: [group->db:p1]
```

same but without port label set
```
Error submitting job: Unexpected response code: 500 (1 error occurred:
	* Task group database validation failed: 1 error occurred:
	* Services are not unique: [group->db]
```

```hcl
job "simple" {
  datacenters = ["dc1"]
  type        = "service"

  group "database" {

    service {
      name     = "db"
      tags     = ["two"]
      # port = "p1"
      provider = "nomad"
    }

    service {
      name     = "db"
      tags     = ["one"]
      # port = "p1"
      provider = "nomad"
    }

    task "db" {
      driver = "raw_exec"
      config {
        command = "bash"
        args    = ["-c", "sleep 15000"]
      }
    }
  }
}
```

</details>

### DIFF
--- a/.changelog/13869.txt
+++ b/.changelog/13869.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+servicedisco: Fixed a bug where non-unique services would escape job validation
+```
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -6628,6 +6628,19 @@ func (tg *TaskGroup) validateServices() error {
 	// Accumulate task names in this group
 	taskSet := set.New[string](len(tg.Tasks))
 
+	// each service in a group must be unique (i.e. used in MakeAllocServiceID)
+	type unique struct {
+		name string
+		task string
+		port string
+	}
+
+	// Accumulate service IDs in this group
+	idSet := set.New[unique](0)
+
+	// Accumulate IDs that are duplicates
+	idDuplicateSet := set.New[unique](0)
+
 	// Accumulate the providers used for this task group. Currently, Nomad only
 	// allows the use of a single service provider within a task group.
 	providerSet := set.New[string](1)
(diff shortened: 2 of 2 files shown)


jrasell (approved):
LGTM!

PR #13880, opened 2022-07-21
block deleting namespaces if the namespace contains a volume

tgross:
When we delete a namespace, we check to ensure that there are no non-terminal
jobs, which effectively covers evals, allocs, etc. CSI volumes are also
namespaced, so extend this check to cover CSI volumes.

We'll also need to add this check for Secure Variables, but this is in a separate
PR because it'll need to be backported.


shoenig (approved):
LGTM!

PR #13888, opened 2022-07-21
block deleting namespace if it contains a secure variable

tgross:
When we delete a namespace, we check to ensure that there are no non-terminal
jobs or CSI volume, which also covers evals, allocs, etc. Secure variables are
also namespaces, so extend this check to them as well.


shoenig (approved):
LGTM!

PR #13893, opened 2022-07-22
Keyboard Navigation acceptance tests

philrenaud:
Adds tests for general Keyboard nav usage:
- Dynamic tables
- Dynamic keynav (arrow keys)
- Shortcuts modal
- Hints
- Shortcuts enable/disable
- Shortcuts re-binding
- Shortcuts from localStorage

### DIFF
--- a/ui/app/components/keyboard-shortcuts-modal.hbs
+++ b/ui/app/components/keyboard-shortcuts-modal.hbs
@@ -31,7 +31,7 @@
                 {{/if}}
               {{/if}}
 
-              <button class="re-bind" disabled={{or (not command.rebindable) command.recording}} type="button" {{on "click" (action this.keyboard.rebindCommand command)}}>
+              <button data-test-rebinder disabled={{or (not command.rebindable) command.recording}} type="button" {{on "click" (action this.keyboard.rebindCommand command)}}>
                 {{#each command.pattern as |key|}}
                   <span>{{clean-keycommand key}}</span>
                 {{/each}}
--- a/ui/app/components/keyboard-shortcuts-modal.hbs
+++ b/ui/app/components/keyboard-shortcuts-modal.hbs
@@ -42,7 +42,7 @@
     <footer>
       <strong>Keyboard shortcuts {{#if this.keyboard.enabled}}enabled{{else}}disabled{{/if}}</strong>
       <Toggle
-        data-test-cpu-toggle
+        data-test-enable-shortcuts-toggle
         @isActive={{this.keyboard.enabled}}
         @onToggle={{this.toggleListener}}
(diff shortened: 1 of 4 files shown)


ui/app/components/keyboard-shortcuts-modal.hbs
```diff
@@ -30,7 +31,7 @@
                 {{/if}}
               {{/if}}
 
-              <button disabled={{or (not command.rebindable) command.recording}} type="button" {{on "click" (action this.keyboard.rebindCommand command)}}>
+              <button class="re-bind" disabled={{or (not command.rebindable) command.recording}} type="button" {{on "click" (action this.keyboard.rebindCommand command)}}>
```
ChaiWithJai, line 34: `nit`: This class is misleading. The class is being added to this button to be used for a test assertion. Personally, I prefer using a `data-test` attribute for this. Even if you're inside of an `#each` you can set the value of the attribute so you can query the correct DOM node. But, if that doesn't work, you can apply an element id which is a better querying pattern.

Disregard this comment if this is actually an applied style, but I checked the `12762-ui-general-keyboard-navigation` and didn't see it there.

ChaiWithJai (approved):

ui/tests/acceptance/keyboard-test.js
```diff
@@ -0,0 +1,208 @@
+// @ts-check
+import { module, test } from 'qunit';
+import { setupApplicationTest } from 'ember-qunit';
+import {
+  click,
+  currentURL,
+  visit,
+  triggerEvent,
+  findAll,
```
ChaiWithJai, line 9: `blocking`: This is unused.

ui/app/components/keyboard-shortcuts-modal.hbs
philrenaud: Got it; modified to a `data-test-rebinder` attribute.

ui/tests/acceptance/keyboard-test.js
philrenaud: now used

ui/app/components/keyboard-shortcuts-modal.js
```diff
@@ -42,18 +43,25 @@ export default class KeyboardShortcutsModalComponent extends Component {
...
-      element: self,
-      target: element,
-      attachment: 'top left',
-      targetAttachment: 'top left',
-      targetModifier: 'visible',
-    });
-    hint.binder = binder;
+  @action
+  tetherToElement(element, hint, self) {
+    if (!this.config.isTest) {
```
philrenaud, line 48: Tried very hard to avoid using a config.isTest call, but Tether doesn't play nice with Ember tests at all. A simple check here solves a world of headaches.

philrenaud:
Resolves #13890

PR #13894, opened 2022-07-22
cleanup: remove unneeded darwin lib

shoenig:
Pretty sure this is dead code as of some recent-ish version of go-psutil.

See the rename notice in
https://github.com/hashicorp/nomad/commits/main/lib/darwin/include/smc.c


tgross (approved):
LGTM

PR #13907, opened 2022-07-23
Add Nomad RetryConfig to agent template config

### LINKED ISSUE #13019: Add config option to avoid task failure on template failure
### Proposal

Currently, when a Nomad task is using a template stanza and that template fails to render because it cannot communicate with an external service, the Nomad agent will retry several times but eventually fail the underlying task. This makes Nomad tasks brittle and can couple Nomad task health to the health of external services like Consul, Vault, or the Nomad Server cluster.

For instance, if Vault goes down for 10 minutes, if Nomad jobs are using secrets in the template stanza, and they retry for 5 minutes, then all of the Nomad tasks for this job will fail. This is particularly bad as the Vault servers might be subject to a thundering herd of Nomad-based requests once they come back up.

Another example is if there are network issues between Nomad clients and servers while Nomad is using native service discovery. If two "edge" clients temporarily lose connection to the Nomad servers, they won't be able to refresh their nomadService information in templates. This should be okay and the clients should continue to run with the data they had on hand. Instead, these edge workloads will be killed. This makes native service discovery very hard to use for edge workloads.

Instead of going down, the Nomad tasks should log warnings, but use stale data. This new configuration value should be added to the [template stanza](https://www.nomadproject.io/docs/job-specification/template) for each task, and configured globally [here](https://www.nomadproject.io/docs/configuration/client#template-parameters) as well.

In the 1.3.X cycle, we should opt in to the new behavior, but we should strongly suggest in the docs & changelog to do so. In 1.4.0, we should switch the default to opt out.

### Use-cases

* Nomad cluster resiliency with Vault
* Nomad cluster resiliency with Consul
* Nomad native service discovery use on flakey/edge connections

DerekStrickland:
Closes #13019 

This PR introduces a retry configuration option for Nomad dependencies in Consul Template. For example, Nomad native service discovery and secure variables will make calls to the Nomad API. If connectivity is lost between the client and the servers, these calls will fail and the task may fail or run with stale data forever depending on the type of error. Operators can now specify a `nomad_retry` stanza in the same way they can for Vault or Consul.  The default config `attempts` is set to `0` which allows for infinite retries.


lgfa29 (approved):
:shipit:

PR #13932, opened 2022-07-28
deps: remove deprecated net/context

tgross:
The `golang.org/x/net/context` package was merged into the stdlib as of go
1.7. Update the imports to use the identical stdlib version. Clean up import
blocks for the impacted files to remove unnecessary package aliasing.

(Note we still have a transitive dependency, but removing our copy will help.)

### DIFF
--- a/client/logmon/server.go
+++ b/client/logmon/server.go
@@ -1,9 +1,9 @@
 package logmon
 
 import (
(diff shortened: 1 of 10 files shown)


lgfa29 (approved):
Thanks for the cleanup!

PR #13938, opened 2022-08-01
cli: do not import structs, use API package only.

jrasell:


command/alloc_status.go
```diff
@@ -498,7 +496,7 @@ func buildDisplayMessage(event *api.TaskEvent) string {
 		desc = strings.Join(parts, ", ")
 	case api.TaskRestarting:
 		in := fmt.Sprintf("Task restarting in %v", time.Duration(event.StartDelay))
-		if event.RestartReason != "" && event.RestartReason != restarts.ReasonWithinPolicy {
+		if event.RestartReason != "" && event.RestartReason != "Restart within policy" {
```
tgross, line 499: Can we copy this into `api` too, so that we don't have a raw string floating around?

command/operator_debug.go
```diff
@@ -913,7 +912,7 @@ func (c *OperatorDebugCommand) collectAgentHost(path, id string, client *api.Cli
 	if err != nil {
 		c.Ui.Error(fmt.Sprintf("%s/%s: Failed to retrieve agent host data, err: %v", path, id, err))
 
-		if strings.Contains(err.Error(), structs.ErrPermissionDenied.Error()) {
+		if strings.Contains(err.Error(), "Permission denied") {
```
tgross, line 915: Let's copy this into `api` too?

tgross (approved):
LGTM!

If we want to keep this state we should probably include an import check like we do for `api` (ref [`GNUmakefile#L179-L180`](https://github.com/hashicorp/nomad/blob/main/GNUmakefile#L179-L180)). We'll need to make sure that check doesn't descend into the `command/agent` package directory, which is always going to be dependent on structs b/c we make RPCs from there.

shoenig:
Nice! :+1:

PR #13941, opened 2022-08-01
deps: update test package

shoenig:
Upgrading to v0.3.0 gives us the ability to add more context to test case failures.

examples in 
https://github.com/shoenig/test/pull/23


tgross (approved):
LGTM

PR #13948, opened 2022-08-01
add namespace to path matching algorithm

ChaiWithJai:
This PR adds namespaces to the path matching algorithm. Previously, we were expecting a set namespace to look up paths based upon like we do for jobs. However, that same logic does not apply for secure variables and because we're not setting a namespace like we do with jobs. Instead, we're doing a dictionary lookup based upon each individual namespace on a secure variable and secure variables can belong to the same hierarchy but have different namespaces.


tgross (approved):
LGTM

PR #13951, opened 2022-08-01
secure vars: enforce ENT quotas (OSS work)

tgross:
This PR implements the OSS side of the remaining quota enforcement work for secure variables, which will be merged in the ENT repo.

* Move the secure variables quota enforcement calls into the state store to ensure quota checks are atomic with quota updates (in the same transaction).
* Switch to a machine-size int instead of a uint64 for quota tracking. The ENT-side quota spec is described as int, and negative values have a meaning as "not permitted at all". Using the same type for tracking will make it easier to do the math around checks, and uint64 is infeasibly large anyways.
* Implements the required HTTP API and CLI changes
* Update the API docs

(ref https://github.com/hashicorp/nomad-enterprise/pull/788)


shoenig (approved):
LGTM

PR #13952, opened 2022-08-01
docs: fix path for quota/usage API

tgross:
The example path in the quota API docs doesn't match the path in the actual API (or the table above it).


DerekStrickland (approved):
:shipit:

Issue #13945, opened 2022-08-01
checks: support for more nomad http check configurations


shoenig:
Currently `http` checks on services using nomad as a provider can only make use of GET requests. 

Make it so that 
- [ ] other `method` types can be specified
- [ ] a `body` request payload can be specified
- [ ] a `header` block can be specified

elieassi:
Good morning @shoenig !

Any clues to around what timeline version 1.4.0 will be out? i see most of these and smilar ones are assigned to that milestone


Thank you!

shoenig:
Hi @elieassi! We don't have a set date, but we're aiming for end of August to release a v1.4-beta, and then the final release would come about 2 weeks after that.

elieassi:
Thats great news,  thanks @shoenig !

PR #13953, opened 2022-08-01
nsd: add support for specifying check.method in nomad service checks

shoenig:
Unblock `check.method` in service validation. Add tests around making sure this value gets plumbed through.

Part of https://github.com/hashicorp/nomad/issues/13945

### DIFF
--- a/helper/funcs.go
+++ b/helper/funcs.go
@@ -3,6 +3,7 @@ package helper
 import (
 	"crypto/sha512"
 	"fmt"
+	"net/http"
 	"path/filepath"
 	"reflect"
 	"regexp"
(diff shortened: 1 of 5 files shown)


helper/funcs.go
```diff
@@ -716,3 +717,21 @@ func NewSafeTimer(duration time.Duration) (*time.Timer, StopFunc) {
...
+	case http.MethodPut:
+	case http.MethodPatch:
+	case http.MethodDelete:
+	case http.MethodConnect:
+	case http.MethodOptions:
+	case http.MethodTrace:
+	default:
+		return false
+	}
+	return true
```
jrasell, line 736: This is just a note rather than a suggestion for other readers as this made me look into the difference between this style and having a single case to catch all the methods like below.

```
func IsMethodHTTP_ALT(s string) bool {
	switch strings.ToUpper(s) {
	case http.MethodGet, http.MethodHead, http.MethodPost, http.MethodPut, http.MethodPatch,
		http.MethodDelete, http.MethodConnect, http.MethodOptions, http.MethodTrace:
		return true
	default:
		return false
	}
}
```

I ran a benchmark as I was curious if there was a difference and wonder if this will change with the changes made in go 1.19 to the compiler using a jump table for string switch statements.
```
Benchmark_IsMethodHTTP-16        	134591882	         9.000 ns/op
Benchmark_IsMethodHTTP_ALT-16    	163068037	         7.306 ns/op
```

Anyway, please ignore, I was just curious and the current implementation is slightly easier to read and 2ns is nothing. :D

jrasell (commented):
LGTM!

jrasell (approved):
LGTM!

helper/funcs.go
shoenig: Always love a good benchmark! FWIW I tend to code more vertically because that's the only direction my mousewheel goes :disappointed:

PR #13966, opened 2022-08-02
Backport of Add stage_publish_dir field to csi_plugin stanza of a job into release/1.3.x

### LINKED ISSUE #13263: Make the base of CSI staging_target_paths and target_paths configurable
In the linked PR, @tgross effectively ensured that containers running CSI plugins would receive staging_target_paths like `/local/csi/staging`. The work was aimed at making giving multiple CSI plugins unique locations outside their containers to stage mounts at (while inside their containers, each plugin would see an isolated `/local/csi/staging directory`). The assumption (I think) was that CSI plugins should not care what path a mount is staged at.

In the below comment, @tgross mentioned he would eventually make this /local/csi path configurable in the csi_plugin jobspec. Though there was some discussion on whether or not there was even a use case for this, it was implied that the work would be done.

> Note to reviewers: this PR is already huge and adding to this field to the jobspec is both a minor change and a lot of code to review. I'm going to do it in a follow-on PR.

_Originally posted by @tgross in https://github.com/hashicorp/nomad/pull/12078#discussion_r810125222_

For better or worse, the BeeGFS CSI driver does need this change to work with the GA Nomad CSI implementation. (We were originally able to run in the beta implementation.) For a variety of reasons, the BeeGFS CSI driver container image does not ship with the mount utility or certain other required userspace utilities (e.g. beegfs-ctl). Instead, we use a bind mount and a chroot to execute the versions of these utilities that are already installed on a running node. This works fine in Kubernetes, where staging_target_paths are reasonably predictable and provided as absolute paths that resolve in the node's mount namespace (e.g. `/var/lib/kubelet/plugins/kubernetes.io/csi/pv/`). A simple bind mount like `/var/lib/kubelet/plugins/kubernetes.io/csi/pv/ : /var/lib/kubelet/plugins/kubernetes.io/csi/pv/` ensures the plugin code and the chrooted utilities have the same view of the mount location.

Because Nomad provides staging_target_paths like `/local/csi/staging/...`, a similar solution is not possible. We would like to be able to configure Nomad as @tgross suggested. For example, if Nomad would (by default) create the bind mount `/opt/nomad/opt/nomad/client/csi/monolith/beegfs-plugin : /local/csi` (as it does today), we would instead like it to create the bind mount `/opt/nomad/opt/nomad/client/csi/monolith/beegfs-plugin : /opt/nomad/opt/nomad/client/csi/monolith/beegfs-plugin` and provide staging_target_paths like `/opt/nomad/opt/nomad/client/csi/monolith/beegfs-plugin/staging/...`

Is the referenced jobspec changed still planned?

hc-github-team-nomad-core:
## Backport

This PR is auto-generated from #13919 to be assessed for backporting due to the inclusion of the label backport/1.3.x.

---

Resolves #13263.

This PR adds a new field to the csi_plugin stanza of a job that determines the base (inside a CSI plugin's container) of CSI staging_target_path and target_path directories.

This is my first PR to Nomad. I have attempted to go through the bullets in the checklist below. Some of the bullets are (I think) not applicable. For the docs, I tried to find all locations within /docs where CSI is mentioned and add the field as appropriate.

I updated comments on a couple of fields that I found to be confusing while trying to figure out where and how to get this implemented. I'm obviously open to reverting anything that the maintainers think is not an improvement.

## Code

* [x] Consider similar features in Consul, Kubernetes, and other tools. Is there prior art we should match? Terminology, structure, etc?
* [x] Add structs/fields to `api/` package
  * `api/` structs usually have Canonicalize and Copy methods
  * New fields should be added to existing Canonicalize, Copy methods
  * Test the structs/fields via methods mentioned above
* [x] Add structs/fields to `nomad/structs` package
  * `structs/` structs usually have Copy, Equals, and Validate methods
  * Validation happens in this package and _must_ be implemented
  * Note that analogous struct field names should match with `api/` package
  * Test the structs/fields via methods mentioned above
  * Implement and test other logical methods
* [x] Add conversion between `api/` and `nomad/structs/` in `command/agent/job_endpoint.go`
  * Add test for conversion
* [x] Determine JSON encoding strategy for responses from RPC (see "JSON Encoding" below)
  * [x] Write `nomad/structs/` to `api/` conversions if necessary and write tests
* [x] Implement diff logic for new structs/fields in `nomad/structs/diff.go`
  * Note that fields must be listed in alphabetical order in `FieldDiff` slices in `nomad/structs/diff_test.go`
  * Add test for diff of new structs/fields
* [x] Add change detection for new structs/fields in `scheduler/util.go/tasksUpdated`
  * Might be covered by `.Equals` but might not be, check.
  * Should return true if the task must be replaced as a result of the change.

## Docs

* [ ] Changelog
* [x] Jobspec entry https://www.nomadproject.io/docs/job-specification/index.html
* [ ] Jobspec sidebar entry https://github.com/hashicorp/nomad/blob/main/website/data/docs-navigation.js
* [ ] Job JSON API entry https://www.nomadproject.io/api/json-jobs.html
* [ ] Sample Response output in API https://www.nomadproject.io/api/jobs.html
* [ ] Consider if it needs a guide https://www.nomadproject.io/guides/index.html


tgross:
Already backported in https://github.com/hashicorp/nomad/pull/13965

PR #13968, opened 2022-08-02
checks: better goroutine handling for test tcp server

shoenig:
This PR hopefully fixes a race condition of this little test tcp server used by the check observer is making connections for test cases.

I believe the tcp listener would sometimes not listen fast enough, and the checker expecting a successful connection would fail to connect. Instead, we now wait for Listen to return before kicking off a goroutine to handle the Accept

### DIFF
--- a/client/serviceregistration/checks/client_test.go
+++ b/client/serviceregistration/checks/client_test.go
@@ -311,7 +311,7 @@ func TestChecker_Do_TCP(t *testing.T) {
 			switch tc.tcpMode {
 			case "ok":
 				// simulate tcp server by listening
-				go tcpServer(ctx, tc.tcpPort)
+				tcpServer(t, ctx, tc.tcpPort)
 			case "hang":
 				// simulate tcp hang by setting an already expired context
 				timeout, stop := context.WithDeadline(ctx, now.Add(-1*time.Second))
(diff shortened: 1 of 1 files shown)


client/serviceregistration/checks/client_test.go
```diff
@@ -327,16 +327,26 @@ func TestChecker_Do_TCP(t *testing.T) {
...
+	})
+
+	go func() {
+		// caller can stop us by cancelling ctx
+		for {
+			con, conErr := l.Accept()
+			if conErr != nil {
+				return
+			}
+			_ = con.Close()
```
tgross, line 349: We're closing this connection immediately after accepting. Will the code under test be ok with this if it gets the ACK and then an immediate RST? Alternatively, can we lean on the context to close out the listener and all the connections so that we know they'll stick around until we're done with the test?

tgross (approved):
LGTM. I've left one comment/question but given this is definitely an improvement over the existing flaky code it's not a blocker.

client/serviceregistration/checks/client_test.go
shoenig: Oh good call, let's do that

jrasell (approved):
LGTM!

PR #13971, opened 2022-08-02
qemu: reduce monitor socket path

lgfa29:
The QEMU driver can take an optional `graceful_shutdown` configuration
which will create a Unix socket to send ACPI shutdown signal to the VM.

Unix sockets have a hard length limit and the driver implementation
assumed that QEMU versions 2.10.1 were able to handle longer paths. This
is not correct, the linked QEMU fix only changed the behaviour from
silently truncating longer socket paths to throwing an error.

By validating the socket path before starting the QEMU machine we can
provide users a more actionable and meaningful error message, and by
using a shorter socket file name we leave a bit more room for
user-defined values in the path, such as the task name.

The maximum length allowed is also platform-dependant, so validation
needs to be different for each OS.

### DIFF
--- a/.changelog/13971.txt
+++ b/.changelog/13971.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+qemu: use shorter socket file names to reduce the chance of hitting the max path length
+```
--- a/drivers/qemu/driver.go
+++ b/drivers/qemu/driver.go
@@ -13,7 +13,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/coreos/go-semver/semver"
 	hclog "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/nomad/client/taskenv"
 	"github.com/hashicorp/nomad/drivers/shared/eventer"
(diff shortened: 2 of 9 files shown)


.changelog/13971.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:improvement
+qemu: use shorter socket file names to reduce the chance of hitting the max path length
+```
+
+```release-note:improvement
+qemu: restore the monitor socket path when restoring a QEMU task.
+```
```
shoenig, line 7: Should we note this as a bug fix? Technically the behavior is different now even for existing tasks.

shoenig (approved):
LGTM; just the one thought

.changelog/13971.txt
lgfa29: Hum...good question. I think then I will split this into two PRs: one for the bug fix where the socket path was not being restored (and will be backported) and another for reducing the socket name and improving the error message (not backported).

lgfa29: https://github.com/hashicorp/nomad/pull/14000 handles the bug fix. I will rebase this branch with socket renaming once that one is merged.

PR #13978, opened 2022-08-03
nsd: add support for setting headers on nomad service http checks

shoenig:
This PR enables setting of the headers block on services registered
into Nomad's service provider. Works just like the existing support
in Consul checks.

Part 2 of #13945


jrasell (approved):
LGTM!

PR #13996, opened 2022-08-03
[ui] "can list variables" capability refactor

philrenaud:
In order to determine if you should list variables / be able to access the variables route at all, check against all your policies' namespaces' secvars' paths' capabilties to see if any of them contain `write`, `read`, `list`, or `destroy` capabilities (in other words, any of the possible capabilities).

If any of your paths, from within any of your namespaces' SecureVariables blocks, contains any of those abilities, you get a link to the route in your nav.

Created a generic lookup function to pave the way for future lookups along the lines of `can view variable` etc.

### DIFF
--- a/.changelog/13996.txt
+++ b/.changelog/13996.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: when determining whether to let a user see Secure Variables, check against all paths in all namespaces.
+```
--- a/.changelog/13996.txt
+++ b/.changelog/13996.txt
@@ -1,3 +1,0 @@
-```release-note:improvement
-ui: when determining whether to let a user see Secure Variables, check against all paths in all namespaces.
-```
--- a/ui/app/abilities/variable.js
+++ b/ui/app/abilities/variable.js
@@ -1,3 +1,4 @@
+// @ts-check
 import { computed, get } from '@ember/object';
 import { or } from '@ember/object/computed';
 import AbstractAbility from './abstract';
@@ -40,8 +41,46 @@ export default class Variable extends AbstractAbility {
 
   @computed('rulesForNamespace.@each.capabilities')
   get policiesSupportVariableView() {
-    return this.rulesForNamespace.some((rules) => {
-      return get(rules, 'SecureVariables');
+    return this.policyNamespacesIncludeSecureVariablesCapabilities(
+      this.token.selfTokenPolicies,
+      ['list', 'read', 'write', 'destroy']
+    );
+  }
+
+  /**
+   *
+   * Map to your policy's namespaces,
+   * and each of their SecureVariables blocks' paths,
+   * and each of their capabilities.
+   * Then, check to see if any of the permissions you're looking for
+   * are contained within at least one of them.
+   *
+   * @param {Object} policies
+   * @param {string[]} capabilities
+   * @returns {boolean}
+   */
+  policyNamespacesIncludeSecureVariablesCapabilities(
+    policies = [],
+    capabilities = []
+  ) {
+    const namespacesWithSecureVariableCapabilities = policies
+      .toArray()
+      .map((policy) => get(policy, 'rulesJSON.Namespaces'))
+      .flat()
+      .map((namespace = {}) => {
+        return namespace.SecureVariables?.Paths;
+      })
+      .flat()
+      .compact()
+      .map((secVarsBlock = {}) => {
+        return secVarsBlock.Capabilities;
+      })
+      .flat()
+      .compact();
+
+    // Check for requested permissions
+    return namespacesWithSecureVariableCapabilities.some((abilityList) => {
+      return capabilities.includes(abilityList);
     });
   }
 
--- a/ui/app/abilities/variable.js
+++ b/ui/app/abilities/variable.js
@@ -39,7 +39,7 @@ export default class Variable extends AbstractAbility {
   )
   canDestroy;
 
-  @computed('rulesForNamespace.@each.capabilities', 'token.selfTokenPolicies')
+  @computed('token.selfTokenPolicies')
   get policiesSupportVariableList() {
     return this.policyNamespacesIncludeSecureVariablesCapabilities(
       this.token.selfTokenPolicies,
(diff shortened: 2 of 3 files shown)


ChaiWithJai (changes requested):
This works for the base case of viewing variables at the root level. However, it doesn't handle the case of a variable stored at `foo/bar/baz/qud` where we need to check view permissions of `foo/bar/`.

Also, we now no longer need to extend `AbstractAbility` because we're not relying on inheriting any functionality or state from that class.

Maybe this would be better suited to be:

`can view at root` which would be implemented in the `gutter-menu` and then we'll still need `can view variable` as an ability throughout the application.

ui/app/abilities/variable.js
```diff
@@ -40,8 +41,46 @@ export default class Variable extends AbstractAbility {
 
   @computed('rulesForNamespace.@each.capabilities')
```
ChaiWithJai, line 42: `nit`:  This is now unused.

philrenaud:
> Also, we now no longer need to extend AbstractAbility because we're not relying on inheriting any functionality or state from that class.

@ChaiWithJai are you sure? We're still using at least `selfTokenIsManagement` from that class, that I can tell. What am I missing here?

philrenaud:
> can view at root which would be implemented in the gutter-menu and then we'll still need can view variable as an ability throughout the application.

As implemented right now, all "can list variable" throughout the app has to do with being able to give you access to the root list*, as the name implies. If we wanted to include an ability check at every path level, we could make this accept another parameter, but I don't see that as being a requirement — the path route contains a link back to root if there are no variables returned there, anyway.

_* or depends on `pathLinkedVariable`, which implies specific variable list ability anyway._

It seems to make more sense to reserve `can view variable` for the case like you mentioned above; linking directly to a variable file itself. That ability should certainly take in the desired path as param.

.changelog/13996.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+ui: when determining whether to let a user see Secure Variables, check against all paths in all namespaces.
+```
```
tgross, line 3: Secure Variables hasn't been released yet, so we shouldn't include a changelog entry.

tgross (changes requested):
LGTM from a technical standpoint, my only blocking comment is on the changelog entry. Some additional non-blocking thoughts:

* I do wonder if it would be helpful to have a second token in the Mirage fixtures that has a much more limited set of permissions (ex. list for some namespaces/paths, read for others, no write); that might make it easier to see the behaviors are correct in the Vercel preview.

* I'm also going to throw a late design wrench into things here: I'm pretty sure the requirement to disable the top-level Secure Variables sidebar came from @mikenomitch under the assumption that administrators would want to disable the whole feature. But that also makes the feature opt-in and less discoverable in the UI. Would it hurt us to have the sidebar button visible and then only check the permissions when it's clicked? (Also: I'm 100% ok with saying "it is what it is for now and we'll punt this discussion to a post-1.4.0 release 😀 )

.changelog/13996.txt
philrenaud: remove; thanks!

philrenaud:
@tgross The suggestion to add a reduced-capabilities mirage fixture is great; I'll add that post-PR.

As for hiding the sidebar nav, I'm going to hold off on that right now per brief slack thread.

tgross (approved):

ui/app/abilities/variable.js
```diff
@@ -38,10 +39,48 @@ export default class Variable extends AbstractAbility {
...
-  @computed('rulesForNamespace.@each.capabilities')
-  get policiesSupportVariableView() {
-    return this.rulesForNamespace.some((rules) => {
-      return get(rules, 'SecureVariables');
+  @computed('token.selfTokenPolicies')
+  get policiesSupportVariableList() {
+    return this.policyNamespacesIncludeSecureVariablesCapabilities(
+      this.token.selfTokenPolicies,
+      ['list', 'read', 'write', 'destroy']
+    );
```
ChaiWithJai, line 47: `nit` `suggestion`: The first parameter is guaranteed to always be this value and the second parameter is a hard coded array. The function API for `policyNamespacesIncludeSecureVariablesCapabilities` should just be a computed getter that returns a boolean instead.

ui/app/abilities/variable.js
```diff
@@ -38,10 +39,48 @@ export default class Variable extends AbstractAbility {
   )
   canDestroy;
 
-  @computed('rulesForNamespace.@each.capabilities')
-  get policiesSupportVariableView() {
-    return this.rulesForNamespace.some((rules) => {
-      return get(rules, 'SecureVariables');
+  @computed('token.selfTokenPolicies')
+  get policiesSupportVariableList() {
```
ChaiWithJai, line 43: Why doesn't this support accepting a `path` as a parameter like the other policy checks in place?

ChaiWithJai (approved):

ui/app/abilities/variable.js
philrenaud: My feeling here is the best way for this to operate is to let the server (200 or 403/404) tell us what we need to know. As such, the `list variables` capability is only meant to be used for "do they have access to variables, generally" as far as the UI is concerned. No path needed for that.

ui/app/abilities/variable.js
philrenaud: A getter won't let us pass parameters; there are plenty of other situations where we would want a generic "check to see if my policy support this thing, generally" for variables — the `can create` capability could leverage this to determine if it should even bother showing the "Create New Variable" button by passing in only `['write']` for example.

ChaiWithJai: path is a bound property of this ability and accessible via `this.path` just like namespace.

ui/app/abilities/variable.js
```diff
@@ -38,10 +39,48 @@ export default class Variable extends AbstractAbility {
...
+    policies = [],
+    capabilities = []
+  ) {
+    const namespacesWithSecureVariableCapabilities = policies
+      .toArray()
+      .map((policy) => get(policy, 'rulesJSON.Namespaces'))
+      .flat()
+      .map((namespace = {}) => {
+        return namespace.SecureVariables?.Paths;
+      })
```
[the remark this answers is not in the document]
philrenaud: Agreed there's some overlap; leaving separate rather than consolidating now, as allPaths is specific to the namespace you're in. Should we rename `allPaths` to `currentNamespacePaths` ?

PR #13998, opened 2022-08-03
template: set default UID/GID to -1

lgfa29:
UID/GID 0 is usually reserved for the root user/group. While Nomad
clients are expected to run as root it may not always be the case.

Setting these values as -1 if not defined will fallback to the pervious
behaviour of not attempting to set file ownership and use whatever
UID/GID the Nomad agent is running as. It will also keep backwards
compatibility, which is specially important for platforms where this
feature is not supported, like Windows.


pkazmierczak (approved):
LGTM 👍

DerekStrickland (approved):
:shipit:

PR #14000, opened 2022-08-03
qemu: restore monitor socket path

lgfa29:
When a QEMU task is recovered the monitor socket path was not being
restored into the task handler, so the `graceful_shutdown` configuration
was effectively ignored if the client restarted.


shoenig (approved):
LGTM!

PR #14001, opened 2022-08-03
allow unhealthy canaries without blocking autopromote

### LINKED ISSUE #8150: Autopromote fails when an allocation goes unhealthy but is properly replaced
### Nomad version
```
nomad -v
Nomad v0.11.2 (807cfebe90d56f9e5beec3e72936ebe86acc8ce3)
```

Amazon Linux 2

### Issue
A job set to autopromote ended up requiring manual promotion to complete the deployment. The only thing different in this case was one of the allocations were unhealthy, so Nomad replaced that allocation. That new allocation became healthy and the healthy threshold was met. However, rather than autopromoting, the deployment stalled.

The deployment status was reporting `Deployment is running pending automatic promotion` when queried, but it was only when we clicked "Promote" in the UI did the deployment complete successfully.

Screen Shot 2020-06-10 at 1 16 49 PM

In this deployment, I clicked "promote" at 1:07 PM PT, which is when the deployment completed. Notice the `progress_deadline` values were about half an hour before this. So this tells me the canary allocations satisfied the healthy threshold as to not fail the deployment, but for some reason, required a manual promotion.

Unfortunately, I'm not sure how to replicate this as we're seeing this issue a low percentage of the time, but given as many deployments as we do, we see it a few times a week.

chuckyz:
fixes #8150 

One thing I'm not 100% about this is if the test is correct.  If I'm reading everything from copy/pasting the other one and understanding the data model this _should_ "just work," because we're never setting `c3` to healthy, so that should test this exact bit of logic (where we want 2 healthy canaries, have 3 placed, and only 2 are ever promoted to healthy).

### DIFF
--- a/.changelog/14001.txt
+++ b/.changelog/14001.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+deployments: Fixed a bug that prevented auto-approval if canaries were marked as unhealthy during deployment
+```
--- a/nomad/deploymentwatcher/deployment_watcher.go
+++ b/nomad/deploymentwatcher/deployment_watcher.go
@@ -293,17 +293,21 @@ func (w *deploymentWatcher) autoPromoteDeployment(allocs []*structs.AllocListStu
 			continue
 		}
 
-		if !dstate.AutoPromote || dstate.DesiredCanaries != len(dstate.PlacedCanaries) {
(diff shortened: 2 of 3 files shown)


lgfa29 (approved):
Thanks for the PR @chuckyz!

The tests look great, I just pushed a commit a simplify it a little. This bug didn't seem to happen with rolling updates (only canary), so I removed them from the tests.

I also pushed a changelog entry for this change.

PR #10785, opened 2021-06-18
lifecycle: unit test for lifecycle task behavior on restarts

tgross:
Work-in-progress PR for discussion with @isabeldepapel while we work on the collection of work in #9464 #9841 #9840 #10017 #10578 (cc @jazzyfresh as a heads up)

* This test is incomplete but the goal is that it sets up a job with all the various lifecycle tasks and then will trigger each of the possible behaviors that @isabeldepapel identifies in the flowchart she's working on, and asserts the resulting restart behavior. 
* I've done a bit of refactoring to reuse this mock generator for some other `LifecycleAlloc` tests, without trying to go all-in on it.

PR #14009, opened 2022-08-04
allocrunner: refactor task coordinator

lgfa29:
The current implementation for the task coordinator unblocks tasks by
performing destructive operations over its internal state (like closing
channels and deleting maps from keys).

This presents a problem in situations where we would like to revert the
state of a task, such as when restarting an allocation with tasks that
have already exited.

With this new implementation the task coordinator behaves more like a
finite state machine where task may be blocked/unblocked multiple times
by performing a state transition.

This initial part of the work only refactors the task coordinator and
is functionally equivalent to the previous implementation. Future work
will build upon this to provide bug fixes and enhancements.

------

_Note to reviewers_: the implementation ended up a little different from the internal RFC description. During tests I noticed that there were more state transitions than I first thought about and the mechanism of closing channels became brittle. 

It was very easy to accidentally close a channel twice, which meant that we had to be very careful during state transitions. This resulted in an implicit dependency between a state and the path taken to reach it (e.g., if coming from the `X` state don't close channel `Y`).

So I opted for an earlier idea which is to have channels with a producer that can be disabled on demand. This makes the operation idempotent, so we don't have to worry about previous states during state transition, but it adds a bit more code and could have some other problems I may have missed.


client/allocrunner/task_coordinator_test.go
```diff
@@ -0,0 +1,449 @@
...
+}
+
+func isChannelBlocking(ch <-chan struct{}) bool {
+	select {
+	case <-ch:
+		return false
+	default:
+		return true
+	}
+}
```
lgfa29, line 449: Since this is just a refactoring of existing code, I tried to keep the behaviour the same, including the tests, but the file rename created a new diff. Here's the old vs. new test diff: https://gist.github.com/lgfa29/6c7d7b6590a9ea377ad12e886c3bfd4c

client/allocrunner/alloc_runner_test.go
```diff
@@ -1811,3 +1836,90 @@ func TestAllocRunner_Lifecycle_Shutdown_Order(t *testing.T) {
...
+
+func TestHasSidecarTasks(t *testing.T) {
+	ci.Parallel(t)
+
+	falseV, trueV := false, true
+
+	cases := []struct {
+		name string
+		// nil if main task, false if non-sidecar hook, true if sidecar hook
+		indicators []*bool
```
shoenig, line 1848: Seems like we could make this more obvious by using a little 3 state enum

client/allocrunner/task_coordinator.go
```diff
@@ -0,0 +1,386 @@
...
+
+// taskCoordinatorState represents a state of the task coordinatator FSM.
+type taskCoordinatorState string
+
+const (
+	taskCoordinatorStateInit      taskCoordinatorState = "init"
+	taskCoordinatorStatePrestart                       = "prestart"
+	taskCoordinatorStateMain                           = "main"
+	taskCoordinatorStatePoststart                      = "poststart"
+	taskCoordinatorStatePoststop                       = "poststop"
```
shoenig, line 19: the "only set the type once" trick only works with `iota`, you have to set each of these as `taskCoordinatorState` to actually be that type

https://go.dev/play/p/gIyL3sfdiAw

client/allocrunner/task_coordinator.go
```diff
@@ -0,0 +1,386 @@
...
+	currentState     taskCoordinatorState
+	currentStateLock sync.RWMutex
+
+	// controllers store a controller for each task lifecycle.
+	controllers map[taskLifecycle]*taskCoordinatorController
+}
+
+// newTaskCoordinator returns a new taskCoordinator in the "init" state and
+// with all tasks blocked.
+func newTaskCoordinator(logger hclog.Logger, tasks []*structs.Task, shutdownCh <-chan struct{}) *taskCoordinator {
```
shoenig, line 70: **_feel free to ignore_**

using channels to coordinate shutdowns is a practice from the early days of Go - nowadays it's trendy to use a Context with a Cancel func, is that something that would make sense here?

shoenig:
@lgfa29 any chance you've run this under the Go race detector?

client/allocrunner/alloc_runner_test.go
lgfa29: Ah good point. I just copied this test from another file and didn't pay much attention to it 😅

client/allocrunner/task_coordinator.go
lgfa29: :old_man_yells_at_gohper:

lgfa29:
> @lgfa29 any chance you've run this under the Go race detector?

I have not, but that's a great idea. I have a janky random task state generator that I used to find some subtle bugs, I will run it with the race detector activated.

client/allocrunner/task_coordinator.go
```diff
@@ -0,0 +1,386 @@
...
+		// Assume task is "main" by default.
+		lifecycle = taskLifecycleMain
+	}
+
+	return c.controllers[lifecycle].waitCh()
+}
+
+// taskStateUpdated notifies that a task state has changed. This may cause the
+// taskCoordinator FSM to progresses to another state.
+func (c *taskCoordinator) taskStateUpdated(states map[string]*structs.TaskState) {
```
tgross, line 125: A note to any other reviewers: these `*structs.TaskState` pointers are to copies of the task state created in allocrunner

client/allocrunner/task_coordinator_test.go
```diff
@@ -0,0 +1,449 @@
...
+	shutdownCh := make(chan struct{})
+	defer close(shutdownCh)
+	coord := newTaskCoordinator(logger, tasks, shutdownCh)
+
+	// All tasks start blocked.
+	requireTaskBlocked(t, coord, init1Task)
+	requireTaskBlocked(t, coord, init2Task)
+	requireTaskBlocked(t, coord, mainTask)
+
+	// Set initial state, prestart tasks are allowed to run, mais is blocked.
```
tgross, line 240: (there's a few of these in this file copy-and-pasted)

```suggestion
	// Set initial state, prestart tasks are allowed to run, main is blocked.
```

client/allocrunner/task_coordinator_test.go
```diff
@@ -0,0 +1,449 @@
...
+	})
+}
+
+func requireChannelBlocking(t *testing.T, ch <-chan struct{}, name string) {
+	testutil.WaitForResult(func() (bool, error) {
+		return isChannelBlocking(ch), nil
+	}, func(_ error) {
+		t.Fatalf("%s channel was passing, should be blocking", name)
+	})
+}
```
tgross, line 440: I love all these helper functions: they make the tests super clear!

client/allocrunner/task_coordinator_test.go
```diff
@@ -0,0 +1,449 @@
...
+			Failed: false,
+		},
+	}
+	coord.taskStateUpdated(states)
+	requireTaskAllowed(t, coord, initTask)
+	requireTaskAllowed(t, coord, sideTask)
+	requireTaskBlocked(t, coord, mainTask)
+}
+
+func TestTaskCoordinator_PoststartStartsAfterMain(t *testing.T) {
```
tgross, line 348: These tests are great: can they be meaningfully adapted to exercise the expected state transitions in the case of a task restart (of all the various lifecycles), or do you think that's out of scope for these tests?

tgross (commented):
@lgfa29 this is a great PR and will make a huge improvement in comprehensibility of this tricky area of the client. I've left some comments for discussion.

client/allocrunner/task_coordinator_controller.go
```diff
@@ -0,0 +1,79 @@
...
+		ch:        make(chan struct{}),
+		updatedCh: make(chan struct{}),
+	}
+}
+
+// run starts the taskCoordinatorController and feeds the channel if the
+// controller is not blocked.
+func (c *taskCoordinatorController) run(shutdownCh <-chan struct{}) {
+	for {
+		if c.blocking {
```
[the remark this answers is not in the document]
lgfa29: Yup, good catch! It turns out that this `blocking` field was completely unnecessary and I was able to rewrite this using only channels, which turned out to be much cleaner (and safer 🙂), so thanks!

client/allocrunner/task_coordinator_controller.go
```diff
@@ -0,0 +1,79 @@
...
+func newTaskCoordinatorController() *taskCoordinatorController {
+	return &taskCoordinatorController{
+		blocking:  true,
+		ch:        make(chan struct{}),
+		updatedCh: make(chan struct{}),
+	}
+}
+
+// run starts the taskCoordinatorController and feeds the channel if the
+// controller is not blocked.
```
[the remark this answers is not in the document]
lgfa29: Great observation. https://github.com/hashicorp/nomad/pull/14009/commits/c5be26beeebdc7aea332ccfaa8dbbd9d06440d37 tries to make the code more clear by moving these new structs into their own package, which allows for better interfaces by actually using public and private methods, and for shorter names without the repetitive `task*` prefix. It also has more extensive documentation in the `doc.go` file.

Is it more clear now?

client/allocrunner/task_coordinator_controller.go
```diff
@@ -0,0 +1,79 @@
...
+				return
+			case <-c.updatedCh:
+				continue
+			}
+		}
+	}
+}
+
+// waitCh returns a channel that can be used to determine if the caller should
+// block and wait.
```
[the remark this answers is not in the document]
lgfa29: Yup, that's right. Thanks!

client/allocrunner/task_coordinator_test.go
lgfa29: Fun fact: "mais" means "more" is Portuguese 😬

client/allocrunner/task_coordinator_test.go
lgfa29: I've been trying to think how to incorporate restarts here (along with #10785), but I couldn't find a good way yet. Since the next step will be to fix the restart flow I think I will postpone additional tests until then.

client/allocrunner/task_coordinator.go
```diff
@@ -0,0 +1,386 @@
...
+	// taskLifecyclePrestartSidecar are tasks with the "prestart" hook and
+	// sidecar set to "true".
+	taskLifecyclePrestartSidecar = "prestart_sidecar"
+
+	// taskLifecycleMain are tasks without a lifecycle or a lifecycle with an
+	// empty hook value.
+	taskLifecycleMain = "main"
+
+	// taskLifecyclePoststart are tasks with the "poststart" hook.
+	taskLifecyclePoststart = "poststart"
```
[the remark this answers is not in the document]
lgfa29: I've been thinking about this, and initially I thought it wouldn't matter. 

The only way a task can run again is with a restart and currently the restart only affects task that are running. So in your example, the DB initialization task will always run only once, unless you happen to send a restart command while the task is still running.

The work I'm doing will allow for a restart command to affect all tasks, so the DB initialization task will run again, but in the expected order. Future work could allow restarting a `dead` task, so this would be relevante for that.

But then I thought about task recovery, and _maybe_ this could be a problem? I wasn't able to fully repro, but just in case this seems like the correct assertion. It also made me re-think how the next state is calculated, so it was worth another look 😄

lgfa29:
@shoenig @tgross @schmichael I'm done fiddling with this PR now 😄

shoenig (approved):
LGTM! The improvement in readability is :100: 

One last thought - would it be worth holding off on backporting until it has time to bake in the 1.4 beta? We could still backport it to 1.3.x on the 1.4 release; this just seems a bit much to go into a bugfix backport directly.

client/allocrunner/task_coordinator_controller.go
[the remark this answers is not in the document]
tgross: Yes, awesome!

client/allocrunner/task_coordinator.go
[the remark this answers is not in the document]
tgross: > The only way a task can run again is with a restart and currently the restart only affects task that are running. 

Ah right, that's a good point. I think there's probably a use case for allowing a task like this to be started again but that's totally out of scope for this PR.

client/allocrunner/task_coordinator_test.go
tgross: Makes sense.

client/allocrunner/tasklifecycle/gate.go
```diff
@@ -0,0 +1,80 @@
...
+// Gate is used by the Coordinator to block or allow tasks from running.
+//
+// It provides a channel that taskRunners listens on to determine when they are
+// allowed to run. The Gate has an infinite loop that is either feeding this
+// channel (therefore allowing listeners to proceed) or not doing anything
+// (causing listeners to block an wait).
+//
+// The Coordinator uses the Gate Open() and Close() methods to control this
+// producer loop.
+type Gate struct {
```
tgross, line 17: Calling this `Gate` is a great metaphor and the use of the API is really clear as a result 👍

tgross (approved):
Nice work here @lgfa29!

client/allocrunner/tasklifecycle/coordinator.go
```diff
@@ -0,0 +1,415 @@
...
+	return c
+}
+
+// Restore is used to set the Coordinator FSM to the correct state when an
+// alloc is restored. Must be called before the allocrunner is running.
+func (c *Coordinator) Restore(states map[string]*structs.TaskState) {
+	// Skip the "init" state when restoring since the tasks were likely already
+	// running, causing the Coordinator to be stuck waiting for them to be
+	// "pending".
+	c.currentStateLock.Lock()
```
schmichael, line 118: Not a big deal, but if the AR must not be running, is there any reason to lock?

I ask because if there *is* a reason to lock, then this locking strategy may be unsafe as mutations to the coordinator state and task states may interleave like:

1. Restore Goroutine: enterStateLocked(Prestart)
2. Other Goroutine: enterStateLocked(`<other>`)
3. Other Goroutine: TaskStateUpdated(`<other>`)
4. Restore Goroutine: TaskStateUpdated(states)

So you would end up with incoherent results: the restored task states but some `<other>` coordinator state.

All that to say I'm pretty sure there is no other goroutine manipulating task lifecycles concurrently with Restore, so this isn't a bug, just slightly confusing code.

client/allocrunner/tasklifecycle/coordinator.go
```diff
@@ -0,0 +1,415 @@
...
+		}
+		return coordinatorStatePoststop
+
+	case coordinatorStatePoststop:
+		return coordinatorStatePoststop
+	}
+
+	// If the code reaches here it's a programming error, since the switch
+	// statement should cover all possible states and return the next state.
+	c.logger.Warn(fmt.Sprintf("unexpected state %s", c.currentState))
```
schmichael, line 194:
```suggestion
	c.logger.Warn("unexpected state", "state", c.currentState)
```

client/allocrunner/tasklifecycle/gate.go
```diff
@@ -0,0 +1,80 @@
...
+func (g *Gate) Close() {
+	select {
+	case <-g.shutdownCh:
+	case g.updateCh <- gateClosed:
+	}
+}
+
+// run starts the infinite loop that feeds the channel if the Gate is opened.
+func (g *Gate) run(initState bool) {
+	isOpen := initState
```
schmichael, line 60: I think this needs to close sendCh when it exits so that waiters on `WaitCh()` don't block indefinitely?
```suggestion
func (g *Gate) run(initState bool) {
	defer close(g.sendCh)

	isOpen := initState
```

client/allocrunner/tasklifecycle/gate.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+}
+
+// Close is used to block listeners from proceeding.
+func (g *Gate) Close() {
+	select {
+	case <-g.shutdownCh:
+	case g.updateCh <- gateClosed:
+	}
+}
```
schmichael, line 56: Perhaps include in the comment what the shutdown behavior is as if the callers cares about whether they were unblocked due to Open/Close vs Shutdown, they'll need to check for the shutdown condition themselves.

A quick peek didn't reveal any callers that would care about Open/Close vs Shutdown, but I think a comment would still be nice in case this code is used more broadly in the future.

client/allocrunner/tasklifecycle/testing.go
```diff
@@ -0,0 +1,56 @@
+package tasklifecycle
+
+import (
+	"testing"
```
schmichael, line 4: We do not want to pull the `testing` package into our core binary. Use https://github.com/mitchellh/go-testing-interface and its `testing.TB` instead of `testing` and `*testing.T`

schmichael (approved):
`testing` package comment is the only blocker I think. Great work!

client/allocrunner/tasklifecycle/testing.go
shoenig: in theory we should be able to set  `//go:build test` and make sure of that 

haven't tried it though :sweat_smile:

schmichael: Wait did that get added? I thought it had been rejected?! https://github.com/golang/go/issues/21360

shoenig: Oh wow, I totally just assumed that would work :man_facepalming:

client/allocrunner/task_coordinator.go
lgfa29: I tried to do this in #14127, but it was a bigger change than I expected since the channels in the `TaskRunner` and `AllocRunner` are used in several places, so I had to drop it.

client/allocrunner/tasklifecycle/doc.go
```diff
@@ -0,0 +1,90 @@
...
+            │     ╷             │     │       │
+            │     ╷             │     │       │ ┌─────────────┐
+            │   Prestart       Main   │       │ │ TASK RUNNER │
+            └─────┼─┬───────────┬─┬───┘       └─┤   (Main)    │
+                  ╷ │Open/      │ │Open/        └─────────────┘
+                  ╷ │Close      │ │Close
+               ┌──┴─▼─┐      ┌──┴─▼─┐
+               │ GATE │      │ GATE │
+               └──────┘      └──────┘
+*/
```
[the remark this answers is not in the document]
lgfa29: I used [ASCIIFlow](https://asciiflow.com/) since it was the first result when I searched for ASCII diagram tool 😬 

I had to edit the result a bit to make the dashed like more distinguishable, but I will link to what I have.

client/allocrunner/tasklifecycle/coordinator.go
lgfa29: Ah good point. Yeah, there should not be anything else running here, as mentioned in the method godoc so I will remove the lock. Another option would be to have to a `taskStateUpdatedLocked` method that implements the logic of `TaskStateUpdated`, but that seems unnecessary at this point.

client/allocrunner/tasklifecycle/coordinator.go
lgfa29: Would this be a good candidate for a `panic` as well, like in https://github.com/hashicorp/nomad/pull/14009#discussion_r943760369? 🤔

client/allocrunner/tasklifecycle/gate.go
lgfa29: Good question, I'm not sure what's the best behaviour here. Closing the channel will leave the gate open on exit, which may result in tasks unexpectedly running? 

So I thought that preserving the gate state would be the safer option, but this could cause goroutines to be blocked forever and leak resources?

client/allocrunner/tasklifecycle/testing.go
schmichael: I think we all have. Kind of boggles my mind that even Rob Pike asked for it and they're like `nah`. Test helpers right inside the package they're intended to be used with seems like a cool idea to me. :shrug:

client/allocrunner/tasklifecycle/coordinator.go
schmichael: Oh! Good call! I think so: The intention seems like the other place: if the method in question doesn't have a case for every possible value, then we goofed.

`<insert unhelpful grumbling about type systems that let you enforce such constraints here>`

client/allocrunner/tasklifecycle/gate.go
lgfa29: Not sure if this is what you were thinking about, but I added these lines:

```go
// if the gate shutdownch channel is closed, this method is a no-op so callers
// should check its state.
```

client/allocrunner/tasklifecycle/testing.go
lgfa29: Oh big TIL. Smells like a new semgrep rule 😄

lgfa29: The [version we're currently using](https://github.com/hashicorp/nomad/blob/v1.3.3/go.mod#L95) doesn't have the [`TB` interface](https://github.com/mitchellh/go-testing-interface/blob/v1.14.1/testing.go), so I'm going to use just `T`.

client/allocrunner/tasklifecycle/gate.go
schmichael: Yeah, definitely want to avoid starting tasks up due to an agent shutdown signal!

Regardless of the behavior here we need to ensure _callers do the right thing:_ callers must always check for an shutdown condition to determine whether to transition states or just escape from the runner ASAP because it's shutting down.

So I think it's actually best to make it explicitly the *callers* problem to figure out what's happening:

1. If WaitCh() is closed: transition states
2. If shutdownCh is closed: shutdown

But the comment on WaitCh should probably make it clear that handling shutdowns are the callers concern so no one ends up blocking forever.

lgfa29: Ahh OK. I have this in the `NewGate` docs, which is `shutdownCh` is passed:

> The Gate should not be used after the shutdownCh is closed.

But I will add a similar not to the `WaitCh` as well 👍

lgfa29: Added the following comment:

```go
// Callers must also check the state of the shutdownCh used to create the Gate
// to avoid blocking indefinitely.
```

PR #14020, opened 2022-08-04
[ui] "Can Read" checks on individual Secure Variables

### LINKED ISSUE #13744: Secure Variables:  Handle `read` capability.

philrenaud:
Resolves #13744 
---
- Adds a path/namespace-aware check for whether individual variables can be accessed via the ui: links become unclickable and not-link-coloured.
- Adds an Unauthorized banner if you do end up on a variable page for which you do not have access (403). Say, you copied it from your machine where you have management access, and you sent it to your network-connected, touchscreen-enabled fridge and clicked it via the tiny touchpad. In this scenario you now have the ability to scowl at your kitchen appliance and read instructions for contacting your administrator.

### Side-effects:
- Made clickable links, both folders and files, blue and underlined to emphasize the difference. (cc @juliezzhou who suggested it)
- Breadcrumb creation in the variables/variable controller now depends on `params` rather than `model`, since we will sometimes still want to render the route without resolving the model (403), and we'd still want a breadcrumb in that scenario.

### TODOs:
- [x] Mirage fixture for a user with fewer SecVars privileges to makes testing easier
- [x] Unit tests for "can read"
- [x] Acceptance tests for paths page

image

### DIFF
--- a/.changelog/13996.txt
+++ b/.changelog/13996.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: when determining whether to let a user see Secure Variables, check against all paths in all namespaces.
+```
--- a/.changelog/13996.txt
+++ b/.changelog/13996.txt
@@ -1,3 +1,0 @@
-```release-note:improvement
-ui: when determining whether to let a user see Secure Variables, check against all paths in all namespaces.
-```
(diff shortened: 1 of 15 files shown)


tgross:
> Made clickable links, both folders and files, blue and underlined to emphasize the difference. (cc @juliezzhou who suggested it)

👍 I really love this!

ChaiWithJai (approved):

PR #14023, opened 2022-08-04
nsd: add support for setting request body in http checks

### LINKED ISSUE #13945: checks: support for more nomad http check configurations

shoenig:
This PR adds support for settings check.body in checks of services
making use of Nomad's service provider.

Closes #13945


jrasell (approved):
LGTM!

PR #14033, opened 2022-08-05
core: automatically plumb task name into task-level services and checks

shoenig:
This PR causes the `task.Name` to be plumbed automatically to task-level services and task-level service checks' `.TaskName` field. The jobspec validation needs to loosen up a bit - such that we accept setting the task name but only if it actually matches the task in which that service/check exists in.


jrasell (approved):
LGTM!

PR #14036, opened 2022-08-05
secure vars: filter by path in List RPCs

tgross:
The List RPCs only checked the ACL for the Prefix argument of the request. Add
an ACL filter to the paginator for the List RPC.

Extend test coverage of ACLs in the List RPC and in the `acl` package, and add a
"deny" capability so that operators can deny specific paths or prefixes below an
allowed path.

---

Note this PR is targeting https://github.com/hashicorp/nomad/pull/13922 because otherwise I'd have to rewrite all the tests once that's merged. If we merge #13922 this will retarget main and should merge cleanly.


nomad/secure_variables_endpoint.go
```diff
@@ -355,7 +365,7 @@ func (s *SecureVariables) listAllSecureVariables(
 			// Identify which namespaces the caller has access to. If they do
 			// not have access to any, send them an empty response. Otherwise,
 			// handle any error in a traditional manner.
-			allowedNSes, err := allowedNSes(aclObj, stateStore, allowFunc)
+			_, err := allowedNSes(aclObj, stateStore, allowFunc)
 			switch err {
 			case structs.ErrPermissionDenied:
 				reply.Data = make([]*structs.SecureVariableMetadata, 0)
```
tgross, line 371: @angrycub an open question here is whether we want to just return `[]` if you have no permissions to the prefix on any namespace rather than `structs.ErrPermissionDenied`. We probably want to keep this outer check because then we don't have to iterate over everything in the state store, but that'd be for performance and not changing the output. Thoughts?

angrycub: Currently we return empty values rather than permission errors in other endpoints. I'm not 100% sure as to the whys though. I don't have strong feels for one approach versus the other, the only advantage that I see being that we don't have to special case it downstream.

This would be an interesting thread to pull and see if there is some reason it exists in the prior art.

angrycub (approved):
LGTM

nomad/secure_variables_endpoint.go
tgross: Ok, that's how it was implemented in this PR, so we should be good-to-go.

Issue #14034, opened 2022-08-05
abilities/variable: Wildcard namespace name is not applied in allPaths()


philrenaud:
allPaths() looks for an exact namespace match when determining if a user should be able to create, list, read, or destroy at a given path.

However, our ACLs can contain wildcards: so both the following would need to be matched against:

```
namespace "*" {
  policy = "read"
  secure_variables { ...
```

```
namespace "namespaces/project123/*" {
  policy = "read"
  secure_variables {...
```

ChaiWithJai:
This example might need to be re-created:

```
namespace "*" {
  policy = "read"
  secure_variables { ...
```

```
namespace "namespaces/project123/*" {
  policy = "read"
  secure_variables {...
```

These namespaces read like paths.

philrenaud:
@ChaiWithJai `_nearestMatchingPath` is used for testing against wildcards in paths; this is specifically about wildcards in namespaces: https://learn.hashicorp.com/tutorials/nomad/access-control-policies

ChaiWithJai:
Gotcha. So when we take a look at `allPaths` lets note this specific line of code:

```
const matchingNamespace = this.namespace ?? 'default';
```

Every time we compute matching paths, we first do a look-up based on the bound namespace property passed into ability. If no namespace is provided, we set a default namespace to `default`. The same way that we set the default namespace to `default` in the `Variable` model.

So now, we need to implement a similar matching algorithm on namespaces that are being passed in. I worked fairly closely with @tgross on the path matching algorithm, and everything was designed to be small re-usable utility functions so we should be in luck here.

I'd recommend a refactor that should look something like:

```
const allPaths () = > {
  const matchingNamespace = nearestMatchingNamespace(this.namespace);
  ...restOfLogic
}

const nearestMatchingNamespace(namespace) {
   if (!namespace) return '*' // or return `default` -- I'll defer this to Tim

   let policyNamespaces = get(policy, 'rulesJSON.Namespaces') || [];

   let matchingNamespace = this._findMatchingNamespace(
          policyNamespaces,
          namespace
        );
}
```

We have a reason to continue extending the `Abstract` ability to use `_findMatchingNamespace` (this handle glob matching) to handle this glob matching behavior.

This should be a simple patch and we can throw a few integration test cases around this. Overall, it should be a pretty simple patch.

PR #14037, opened 2022-08-05
refact:  namespace glob matching

ChaiWithJai:
This PR addresses matching namespaces when a namespace glob is used.

Closes [14034](https://github.com/hashicorp/nomad/issues/14034).

Reference:  https://learn.hashicorp.com/tutorials/nomad/access-control-policies

### DIFF
--- a/ui/tests/unit/abilities/variable-test.js
+++ b/ui/tests/unit/abilities/variable-test.js
@@ -1069,5 +1069,76 @@ module('Unit | Ability | variable', function (hooks) {
         'It should return the exact path match.'
       );
     });
+
+    test('it handles globs in namespaces', function (assert) {
+      const mockToken = Service.extend({
+        aclEnabled: true,
+        selfToken: { type: 'client' },
+        selfTokenPolicies: [
+          {
+            rulesJSON: {
+              Namespaces: [
+                {
+                  Name: '*',
+                  Capabilities: ['list-jobs', 'alloc-exec', 'read-logs'],
+                  SecureVariables: {
+                    Paths: [
+                      {
+                        Capabilities: ['list'],
+                        PathSpec: '*',
+                      },
+                    ],
+                  },
+                },
+                {
+                  Name: 'namespace-1',
+                  Capabilities: ['list-jobs', 'alloc-exec', 'read-logs'],
+                  SecureVariables: {
+                    Paths: [
+                      {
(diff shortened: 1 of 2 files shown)


ui/tests/unit/abilities/variable-test.js
```diff
@@ -1069,5 +1069,76 @@ module('Unit | Ability | variable', function (hooks) {
...
+    test('it handles globs in namespaces', function (assert) {
+      const mockToken = Service.extend({
+        aclEnabled: true,
+        selfToken: { type: 'client' },
+        selfTokenPolicies: [
+          {
+            rulesJSON: {
+              Namespaces: [
+                {
+                  Name: '*',
```
tgross, line 1082: Given the example in https://github.com/hashicorp/nomad/issues/14034 should we test against a partial match example? ex. include a rule for `pab*` namespace as well, so that we verify we're getting the nearest-match behavior we'd expect.

tgross (approved):
LGTM

ui/tests/unit/abilities/variable-test.js
ChaiWithJai: I agree with that sentiment and this is where I wonder how far to take testing.

That logic is handled by another [function](https://github.com/hashicorp/nomad/blob/main/ui/app/abilities/abstract.js#L83-L126) that is used throughout the application - but not tested.

So should we write unit tests for that other function because that functionality is encapsulated within that separate function we're invoking.

No one ever taught me functional programming and I feel like I need more clarity on cases like this.

tgross: The way I look at it is that you want to figure out the right balance of coverage vs cost (both runtime and cost for you to crank out the tests) . 

The nice thing about testing the `_findMatchingNamespace` directly is that there's less setup involved in the test so you can test a larger number of combinations more cheaply to get full coverage of that function, and then you know any callers can rely on it being correct. Then you can potentially cover the callers (and their callers, etc.) at a higher level of abstraction. In the case of this test, the only behavior you're exercising is that of `_findMatchingNamespace`, so that suggests that might be a better place to put the test.

philrenaud (approved):
This works great! Smoke tests for partial matching went well, too.

If we wanted to refactor these to be a little more functionally pure, per comment discussion, I think that's a very worthy refactor. Not something we need to think about pre-1.4 I think, though.

PR #14040, opened 2022-08-05
Add notes about DAS being prometheus only

mikenomitch:
Add note that DAS is prometheus only?

Question: Can it also use raw Nomad? If so, I must clarify this!


jrasell (approved):
This is accurate; LTGM!

PR #14057, opened 2022-08-09
ACL Role: add RPC, HTTP API, and API SDK functionality.

jrasell:
New ACL Role RPC endpoints have been created to allow the creation,
update, read, and deletion of ACL roles. All endpoints require a
management token; in the future readers will also be allowed to
view roles associated to their ACL token.

The create endpoint in particular is responsible for deduplicating
ACL policy links and ensuring named policies are found within
state. This is done within the RPC handler so we perform a single
loop through the links for slight efficiency.

These new HTTP endpoints are exposed under the /v1/acl/roles
and /v1/acl/role endpoints.

targets: feature branch
notes: best viewed by commit which each represents a different
Nomad system.

### DIFF
--- a/api/acl.go
+++ b/api/acl.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"errors"
 	"fmt"
 	"time"
 )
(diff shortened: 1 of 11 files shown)


api/acl.go
```diff
@@ -202,6 +203,90 @@ func (a *ACLTokens) ExchangeOneTimeToken(secret string, q *WriteOptions) (*ACLTo
...
+	if err != nil {
+		return nil, nil, err
+	}
+	return &resp, wm, nil
+}
+
+// Update is used to update an existing ACL role.
+func (a *ACLRoles) Update(role *ACLRole, w *WriteOptions) (*ACLRole, *WriteMeta, error) {
+	if role.ID == "" {
+		return nil, nil, errors.New("missing ACL role ID")
```
shoenig, line 242: we create this same error a number of times, maybe extract it out?

command/agent/acl_endpoint.go
```diff
@@ -322,3 +322,200 @@ func (s *HTTPServer) ExchangeOneTimeToken(resp http.ResponseWriter, req *http.Re
...
+		return s.aCLRoleRequest(resp, req, suffixParts[0])
+	case 2:
+		// This endpoint only supports GET.
+		if req.Method != http.MethodGet {
+			return nil, CodedError(http.StatusMethodNotAllowed, ErrInvalidMethod)
+		}
+
+		// Ensure the role name is not an empty string which is possible if the
+		// caller requested "/v1/acl/role/name/"
+		if suffixParts[1] == "" {
```
shoenig, line 398: could we also assert `suffixParts[0] == "name"`?

command/agent/acl_endpoint.go
```diff
@@ -322,3 +322,200 @@ func (s *HTTPServer) ExchangeOneTimeToken(resp http.ResponseWriter, req *http.Re
...
+		}
+
+		return s.aclRoleGetByNameRequest(resp, req, suffixParts[1])
+
+	default:
+		return nil, CodedError(http.StatusBadRequest, "invalid URI")
+	}
+}
+
+func (s *HTTPServer) aCLRoleRequest(
```
shoenig, line 409:
```suggestion
func (s *HTTPServer) aclRoleRequest(
```

shoenig (approved):
LGTM! just a couple suggestions

Issue #13940, opened 2022-08-01
cli commands not forwarding to leader node for VAULT_TOKEN validation 


shoenig:
https://github.com/hashicorp/nomad/issues/13062#issuecomment-1199413980

> Additionally, it seems that if I run a validate on any system other than a nomad leader node, i get this error:

```
Job validation errors:
1 error occurred:
	* failed to lookup Vault token: Vault client not active
```

PR #14065, opened 2022-08-09
cli: forward request for job validation to nomad leader

shoenig:
This PR changes the behavior of 'nomad job validate' to forward the
request to the nomad leader, rather than responding from any server.

This is because we need the leader when validating Vault tokens, since
the leader is the only server with an active vault client.

Closes https://github.com/hashicorp/nomad/issues/13940

### DIFF
--- a/nomad/job_endpoint.go
+++ b/nomad/job_endpoint.go
@@ -464,10 +464,8 @@ func getSignalConstraint(signals []string) *structs.Constraint {
 	}
 }
 
-// Summary retrieves the summary of a job
-func (j *Job) Summary(args *structs.JobSummaryRequest,
-	reply *structs.JobSummaryResponse) error {
-
+// Summary retrieves the summary of a job.
+func (j *Job) Summary(args *structs.JobSummaryRequest, reply *structs.JobSummaryResponse) error {
 	if done, err := j.srv.forward("Job.Summary", args, args, reply); done {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


nomad/job_endpoint.go
```diff
@@ -511,8 +509,14 @@ func (j *Job) Summary(args *structs.JobSummaryRequest,
...
 }
 
-// Validate validates a job
+// Validate validates a job.
+//
+// Must forward to the leader, because only the leader will have a live Vault
+// client with which to validate vault tokens.
 func (j *Job) Validate(args *structs.JobValidateRequest, reply *structs.JobValidateResponse) error {
+	if done, err := j.srv.forward("Job.Validate", args, args, reply); done {
+		return err
```
lgfa29, line 518: I don't think request allows for stale right? But if we would were to allow validation without forwarding (like a local check) we could downgrade Vault validation errors to warnings. But since (I think?) stale requests are not allowed here we're probably good as it is.

lgfa29 (approved):
LGTM! I think the issue number in the PR message is wrong, it seems like it was supposed to be #13940.

This [paragraph in docs](https://github.com/hashicorp/nomad/blob/main/website/content/api-docs/validate.mdx?plain=1#L16-L18) could use some updating as well, I don't think that is even true anymore? 😅

nomad/job_endpoint.go
shoenig: Nope! The `JobValidateRequest` rpc embeds a WriteRequest, which has no notion of stale

Issue #14061, opened 2022-08-09
[ui, bug] Can Write Variable check is not permissive enough


philrenaud:
Two bugs:

- The link to create a new variable in the "no variables found" form is active even when the button is disabled 
![image](https://user-images.githubusercontent.com/713991/183720170-f911ac6e-340a-4926-bcf9-f259fad65066.png)

- A policy with `write` in one of the paths should have access to create a new variable from root; because if there are no yet-created variables, how would the user even be able to create one?

```
    # full access to secrets in all project paths
    path "project/*" {
      capabilities = ["write", "read", "destroy", "list"]
    }

    # full access to secrets in all project paths
    path "*" {
      capabilities = ["list"]
    }
```

ChaiWithJai:
> A policy with write in one of the paths should have access to create a new variable from root; because if there are no yet-created variables, how would the user even be able to create one?

I don't think this should block `1.4.0`. This user journey should get a design and specifically address how we'll address error state before we pick up work on this IMO.

philrenaud:
@ChaiWithJai we don't block users from root variable creation at the API or CLI levels and we shouldn't block them at the UI level either

philrenaud:
Resolved by #14067

PR #14067, opened 2022-08-09
[ui] Fixes a bug for first-time SecVars users on namespaces

### LINKED ISSUE #13949: Bug:  We need to figure out which namespace to check on the `/variables` view to check if a user has access to create a variable at the root level.

philrenaud:
Resolves #13949 and #14061

Makes the empty-variables-list link adhere to the same rules as the create button
Adds a more permissive check when you're on * namespace and * path

Added a TODO for handling * namespace on a specific path. We're being too restricted about that currently and should consider modifying allPaths to account for checking against multiple namespaces at once.

### DIFF
--- a/ui/app/abilities/variable.js
+++ b/ui/app/abilities/variable.js
@@ -107,7 +107,7 @@ export default class Variable extends AbstractAbility {
     });
   }
 
-  @computed('path', 'allPaths')
+  @computed('path', 'namespace', 'allPaths')
   get policiesSupportVariableWriting() {
     if (this.namespace === WILDCARD_GLOB && this.path === WILDCARD_GLOB) {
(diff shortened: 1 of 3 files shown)


ChaiWithJai (approved):
This has TODO comments and more lines of comments than code for the core logic. I'm just "approving" to make sure I'm not blocking you from doing things your way. But IMHO this should remain in draft.

ui/app/abilities/variable.js
```diff
@@ -101,10 +109,23 @@ export default class Variable extends AbstractAbility {
 
   @computed('path', 'allPaths')
```
ChaiWithJai, line 110: `blocking` you're missing declaring dependencies here.

PR #13670, opened 2022-07-11
ui: fix zero memory utilization bug on systems using cgroups v2

shoenig:
A quick band-aid fix for https://github.com/hashicorp/nomad/issues/13023. On systems using cgroups v2, the `MemoryStats` struct no longer reports `.RSS`, but rather `.Usage`. Kind of just assuming they mean the same thing.

The fix here is to just check if `.Usage` exists and if so use that, otherwise fallback to the `.RSS` field.

```
"MemoryStats": {
	"RSS": 0,
	"Cache": 0,
	"Swap": 0,
	"MappedFile": 0,
	"Usage": 98656256,
	"MaxUsage": 0,
	"KernelUsage": 0,
	"KernelMaxUsage": 0,
	"Measured": [
		"Cache",
		"Swap",
		"Usage"
	]
},
```

@philrenaud @ChaiWithJai 

Fixes #13023

PR #14069, opened 2022-08-09
cli: for systems with cgroups v2, fix alloation resource utilization showing 0 memory used

brian-athinkingape:
Follow up to https://github.com/hashicorp/nomad/pull/13670 which fixes the bug for the UI, but not using the CLI (`nomad node status`).


shoenig (approved):
LGTM; thanks for the fix @brian-athinkingape!

PR #14070, opened 2022-08-09
ci: delete generated files after build

lgfa29:
The generated files are created as part of the final build process but
should not present in the repository afterwards.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -186,10 +186,18 @@ jobs:
             echo "Version ${{ github.event.inputs.version }} is a prerelease, skipping update of LAST_RELEASE"
           fi
 
+      - name: Remove generated files
+        run: |
+          # These generated files are only needed when building the final
+          # binary and should be not be present in the repository afterwards.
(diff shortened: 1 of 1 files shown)


DerekStrickland (approved):
Nice!

.github/workflows/release.yml
```diff
@@ -186,10 +186,18 @@ jobs:
             echo "Version ${{ github.event.inputs.version }} is a prerelease, skipping update of LAST_RELEASE"
           fi
 
+      - name: Remove generated files
+        run: |
+          # These generated files are only needed when building the final
+          # binary and should be not be present in the repository afterwards.
+          find . -name '*.generated.go' | xargs git rm
```
apollo13, line 193: Wouldn't a `-delete` argument to find achieve the same? Is `git rm` really necessary or would a `rm` do as well

lgfa29: There's a small advantage in using `git rm` because it's like a `rm && git add`, so the removed file is automatically placed in staging area. Not a big difference though 😄

PR #14071, opened 2022-08-09
[ui] Show a different message when there are no tasks in a job

### LINKED ISSUE #10545: ui: Exec shouldn’t suggest choosing a task when there are none
Right?

image

Possible message:

> Run a job and then select a task to start your session.

Context: running the example job on macOS without having Docker started.

geovannyAvelar:
Show a different message in the terminal when there are no tasks in the job.

Closes #10545

### DIFF
--- a/.changelog/14071.txt
+++ b/.changelog/14071.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: Display different message when trying to exec into a job with no task running.
+```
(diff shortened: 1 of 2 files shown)


lgfa29 (approved):
Thank you for the contribution @geovannyAvelar!

I think this situation can happen for a few different reasons (some expected, others maybe not?) so the initial suggestion was a bit inaccurate, so I replaced it with a generic status message.

Issue #13950, opened 2022-08-01
Bug:  Inject namespace bound property in `/jobs/job/task-group`, `allocations/allocations/task` view and `gutter-menu` component.


ChaiWithJai:

ChaiWithJai:
Should be blocked by [14063](https://github.com/hashicorp/nomad/issues/14063) which is blocked by [14066](https://github.com/hashicorp/nomad/issues/14066).

PR #14072, opened 2022-08-10
UI variables made to be unique by namespace and path

### LINKED ISSUE #14066: [ui] Variables should be unique by `namespace` and `path`
A user cannot create a variable using a path that is already used even if its under a different namespace.

Image

philrenaud:
resolves #14066, #13950 and #14066

Adds namespaces to the stored identity of variables.

This means you can have multiple variables at `a/b/c/foo0` if their namespaces are different.

IDs are now stored in the browser as `<pathname>@<namespace>`, or `a/b/c/foo0@default` in the above case, and URLs reflect this:
image

Note that we now display same-pathed, different-namespaced variables alongside one another:
image

### DIFF
--- a/ui/app/components/secure-variable-form.js
+++ b/ui/app/components/secure-variable-form.js
@@ -6,7 +6,7 @@ import { tracked } from '@glimmer/tracking';
 import { inject as service } from '@ember/service';
 import { trimPath } from '../helpers/trim-path';
 import { copy } from 'ember-copy';
-import EmberObject, { computed, set } from '@ember/object';
+import EmberObject, { set, computed } from '@ember/object';
 // eslint-disable-next-line no-unused-vars
 import MutableArray from '@ember/array/mutable';
 import { A } from '@ember/array';
--- a/ui/app/components/secure-variable-form.js
+++ b/ui/app/components/secure-variable-form.js
@@ -180,6 +180,7 @@ export default class SecureVariableFormComponent extends Component {
       });
       this.router.transitionTo('variables.variable', this.args.model.path);
     } catch (error) {
+      console.log('errrrrrrrr', error);
       this.flashMessages.add({
         title: `Error saving ${this.args.model.path}`,
         message: error,
--- a/ui/app/components/secure-variable-form.js
+++ b/ui/app/components/secure-variable-form.js
@@ -178,7 +178,7 @@ export default class SecureVariableFormComponent extends Component {
         destroyOnClick: false,
         timeout: 5000,
       });
-      this.router.transitionTo('variables.variable', this.args.model.path);
+      this.router.transitionTo('variables.variable', this.args.model.id);
     } catch (error) {
       this.flashMessages.add({
         title: `Error saving ${this.args.model.path}`,
--- a/ui/app/components/secure-variable-form.js
+++ b/ui/app/components/secure-variable-form.js
@@ -49,9 +49,8 @@ export default class SecureVariableFormComponent extends Component {
 
   get shouldDisableSave() {
     const disallowedPath =
-      this.args.model?.path?.startsWith('nomad/') &&
-      !this.args.model?.path?.startsWith('nomad/jobs');
-    return !!this.JSONError || !this.args.model?.path || disallowedPath;
+      this.path?.startsWith('nomad/') && !this.path?.startsWith('nomad/jobs');
+    return !!this.JSONError || !this.path || disallowedPath;
   }
 
   /**
@@ -99,7 +98,6 @@ export default class SecureVariableFormComponent extends Component {
   /**
    * @type {DuplicatePathWarning}
    */
-  // @computed('args.{model.path,existingVariables}', 'variableNamespace')
   get duplicatePathWarning() {
     const existingVariables = this.args.existingVariables || [];
     const pathValue = trimPath([this.path]);
@@ -175,12 +173,13 @@ export default class SecureVariableFormComponent extends Component {
       }
 
       this.args.model.set('keyValues', this.keyValues);
+      this.args.model.set('path', this.path);
       this.args.model.setAndTrimPath();
       await this.args.model.save();
 
       this.flashMessages.add({
         title: 'Secure Variable saved',
-        message: `${this.args.model.path} successfully saved`,
+        message: `${this.path} successfully saved`,
         type: 'success',
         destroyOnClick: false,
         timeout: 5000,
@@ -188,7 +187,7 @@ export default class SecureVariableFormComponent extends Component {
       this.router.transitionTo('variables.variable', this.args.model.id);
     } catch (error) {
       this.flashMessages.add({
-        title: `Error saving ${this.args.model.path}`,
+        title: `Error saving ${this.path}`,
         message: error,
         type: 'error',
         destroyOnClick: false,
--- a/ui/app/components/secure-variable-form.js
+++ b/ui/app/components/secure-variable-form.js
@@ -145,7 +145,7 @@ export default class SecureVariableFormComponent extends Component {
     if (e.type === 'submit') {
       e.preventDefault();
     }
-    // TODO: temp, hacky way to force translation to tabular keyValues
+
     if (this.view === 'json') {
       this.translateAndValidateItems('table');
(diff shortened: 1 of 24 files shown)


ui/app/components/secure-variable-form.js
```diff
@@ -94,19 +85,29 @@ export default class SecureVariableFormComponent extends Component {
...
-    const value = trimPath([e.target.value]);
+  /**
+   * @typedef {Object} DuplicatePathWarning
+   * @property {string} path
+   */
+
+  /**
+   * @type {DuplicatePathWarning}
+   */
+  @computed('args.{model.path,existingVariables}', 'variableNamespace')
```
ChaiWithJai, line 96: `blocking`: This doesn't need to be a computed. `variableNamespace` is a `tracked` property and the remaining 3 arguments are arguments that are passed to the component that can be a reactive getter.

```
get path() {
   return this.args.model.path;
}

get duplicatePathWarning() {
  // existing logic
}
```

ChaiWithJai: `existingVariables` is coming from the controller which peeks the store when the controller is invoked but isn't updating.

ui/app/components/secure-variable-form.js
```diff
@@ -94,19 +85,29 @@ export default class SecureVariableFormComponent extends Component {
...
+  get duplicatePathWarning() {
     const existingVariables = this.args.existingVariables || [];
+    const pathValue = trimPath([this.args.model.path]);
     let existingVariable = existingVariables
       .without(this.args.model)
-      .find((v) => v.path === value);
+      .find(
+        (v) => v.path === pathValue && v.namespace === this.variableNamespace
+      );
```
ChaiWithJai, line 104: `suggestion`: the `filter` method takes a callback, too, if you want to play code golf.

ui/app/controllers/variables/variable.js
```diff
@@ -3,12 +3,14 @@ import Controller from '@ember/controller';
 export default class VariablesVariableController extends Controller {
   get breadcrumbs() {
     let crumbs = [];
-    this.params.path.split('/').reduce((m, n) => {
+    let id = decodeURI(this.params.id.split('@').slice(0, -1).join('@')); // remove namespace
+    let namespace = this.params.id.split('@').slice(-1)[0];
+    id.split('/').reduce((m, n) => {
```
ChaiWithJai, line 8: `question`: does this handle jobs that have `/` in their name?

ui/app/routes/variables/variable.js
```diff
@@ -8,8 +8,9 @@ export default class VariablesVariableRoute extends Route.extend(
 ) {
   @service store;
   model(params) {
+    // console.log('paramas', params, this.store.findAll('variable'));
```
ChaiWithJai, line 11: `blocking`: remove

ui/app/serializers/variable.js
```diff
@@ -3,12 +3,17 @@ import ApplicationSerializer from './application';
 
 @classic
 export default class VariableSerializer extends ApplicationSerializer {
-  primaryKey = 'Path';
+  // primaryKey = 'Path';
```
ChaiWithJai, line 6: should this stay here?

ui/mirage/scenarios/default.js
```diff
@@ -207,17 +207,17 @@ function variableTestCluster(server) {
 
   server.create('variable', {
     id: `nomad/jobs/${variableLinkedJob.id}/${variableLinkedGroup.name}/${variableLinkedTask.name}`,
```
ChaiWithJai, line 209: `question`: why are these `id` properties being updated to include `namespace?

ui/tests/acceptance/secure-variables-test.js
```diff
@@ -103,9 +103,8 @@ module('Acceptance | secure variables', function (hooks) {
     await percySnapshot(assert);
 
     await click(fooLink);
-    assert.equal(
-      currentURL(),
-      '/variables/var/a/b/c/foo0',
+    assert.ok(
+      currentURL().includes('/variables/var/a/b/c/foo0'),
```
ChaiWithJai, line 107: `blocking`: The assertion should include that the `namespace` is included in the URL otherwise the page will `404` or return the wrong variable.

ChaiWithJai: I think this test is still failing.

ui/tests/integration/components/secure-variable-form-test.js
```diff
@@ -219,6 +223,8 @@ module('Integration | Component | secure-variable-form', function (hooks) {
 
   module('Validation', function () {
```
ChaiWithJai, line 224: One of these tests is still failing.

ui/app/components/variable-paths.js
```diff
@@ -26,9 +26,9 @@ export default class VariablePathsComponent extends Component {
   }
 
   @action
-  async handleFileClick({ path, variable: { namespace } }) {
+  async handleFileClick({ path, variable: { id, namespace } }) {
     if (this.can.can('read variable', null, { path, namespace })) {
```
ChaiWithJai, line 30: Our abilities no longer need 2 bound properties now that `id` includes `path` and `namespace`. We can update that API accordingly and reduce the number of arguments we're trying to pass around.

ChaiWithJai (changes requested):
2 tests are failing. But good job so far.

ui/app/controllers/variables/variable.js
philrenaud: Hi @ChaiWithJai — that's specifically what this handles (guessing you mean variables).

A `/` in a variable name is the thing that's being split on here, and implies a folder-like hierarchy.

ui/app/components/secure-variable-form.js
philrenaud: Good call! Fixed

ui/mirage/scenarios/default.js
philrenaud: @ChaiWithJai For demo purposes, we want the variables that link to jobs, to also have the same namespace as the jobs/groups/tasks they're linking to. Otherwise we'd have a 404ing link situation (which can arise! We don't stop you from making a variable that starts with `nomad/jobs/` against a job that does not exist). We just want our mirage demo flow to be on the happy path.

ui/app/components/variable-paths.js
philrenaud: This seems like a fine improvement to make, but I'd like to handle it outside of the context of this PR for fear of inadvertent side-effects.

ui/app/components/secure-variable-form.js
ChaiWithJai, line 96: This `computed` here notifies me of a lurking bug. The reason we're relying on this `computed` instead of reactive getter because there's a breaking test is because of how we're handling `this.args.model.path` in the `Input` in our form.

The `@value` property of `Input` creates a bidirectional binding. In our case, passing `@model.path` is causing the user action to directly manipulate this record in the store. And the `mockedModel` value we pass as an argument in the store isn't a reference to the record in the Ember Data store which causes that test to fail, as it should.

But, if we correct that bug we'd just need to do something like:

```
class SecureVariableForm extends Component {
  @tracked path;

  constructor() {
    this.path = path;
  }

  get duplicatePath() {
    ...remainingLogic();
  }
}

---secure-variable-form.hbs---

<Input @value={{this.path}} /> // instead of @model.path which directly mutates the store

```

ui/app/routes/variables/variable.js
ChaiWithJai: Still here.

ui/tests/acceptance/secure-variables-test.js
ChaiWithJai: Hey! Noticed this was marked as resolved, but I didn't see the assertion change.

ChaiWithJai (approved):

ui/app/components/secure-variable-form.js
philrenaud: Resolved, thank you!

PR #14087, opened 2022-08-11
cli: add new acl role subcommands for CRUD role actions.

jrasell:
targets: feature branch

### DIFF
--- a/command/acl_role_create.go
+++ b/command/acl_role_create.go
@@ -1,0 +1,148 @@
+package command
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/nomad/api"
+	"github.com/mitchellh/cli"
+	"github.com/posener/complete"
+)
+
+// Ensure ACLRoleCreateCommand satisfies the cli.Command interface.
+var _ cli.Command = &ACLRoleCreateCommand{}
+
+// ACLRoleCreateCommand implements cli.Command.
+type ACLRoleCreateCommand struct {
+	Meta
+
+	name        string
+	description string
+	policyNames []string
+	json        bool
+	tmpl        string
+}
+
+// Help satisfies the cli.Command Help function.
+func (a *ACLRoleCreateCommand) Help() string {
+	helpText := `
+Usage: nomad acl token create [options]
+
+  Create is used to create new ACL roles. Use requires a management token.
+
+General Options:
+
+  ` + generalOptionsUsage(usageOptsDefault|usageOptsNoNamespace) + `
+
+ACL Create Options:
+
+  -name
+    Sets the human readable name for the ACL role. The name must be between
+    1-128 characters and is a required parameter.
+
+  -description
+    A free form text description of the role that must not exceed 256
+    characters.
+
+  -policy-name
+    Specifies a policy to associate with the role identified by their name. This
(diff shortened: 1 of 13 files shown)


command/acl_role_create.go
```diff
@@ -0,0 +1,148 @@
...
+    Sets the human readable name for the ACL role. The name must be between
+    1-128 characters and is a required parameter.
+
+  -description
+    A free form text description of the role that must not exceed 256
+    characters.
+
+  -policy-name
+    Specifies a policy to associate with the role identified by their name. This
+    flag can be specified multiple times and should be specified at least once.
```
DerekStrickland, line 49:
```suggestion
    flag can be specified multiple times and must be specified at least once.
```

command/acl_role_info_test.go
```diff
@@ -0,0 +1,95 @@
...
+	s := ui.OutputWriter.String()
+	require.Contains(t, s, fmt.Sprintf("ID           = %s", aclRole.ID))
+	require.Contains(t, s, fmt.Sprintf("Name         = %s", aclRole.Name))
+	require.Contains(t, s, "Description  = <none>")
+	require.Contains(t, s, fmt.Sprintf("Policies     = %s", aclPolicy.Name))
+
+	ui.OutputWriter.Reset()
+	ui.ErrorWriter.Reset()
+
+	// Look up the ACL role using its ID.
```
DerekStrickland, line 85:
```suggestion
	// Look up the ACL role using its Name.
```

command/acl_role_update.go
```diff
@@ -0,0 +1,217 @@
...
+	var updatedRole api.ACLRole
+
+	// Depending on whether we are merging or not, we need to take a different
+	// approach.
+	switch a.noMerge {
+	case true:
+
+		// Perform some basic validation on the submitted role information to
+		// avoid sending API and RPC requests which will fail basic validation.
+		if a.name == "" {
```
DerekStrickland, line 138: Should we validate max length. or are we relying on the API/RPC to bubble that up? I guess that applies to other commands as well.

command/acl_role_update_test.go
```diff
@@ -0,0 +1,124 @@
...
+	code = cmd.Run([]string{"-address=" + url, "-token=" + rootACLToken.SecretID, aclRole.ID})
+	require.Equal(t, 1, code)
+	require.Contains(t, ui.ErrorWriter.String(), "Please provide at least one flag to update the ACL role")
+
+	ui.OutputWriter.Reset()
+	ui.ErrorWriter.Reset()
+
+	// Update the description using the merge method.
+	code = cmd.Run([]string{
+		"-address=" + url, "-token=" + rootACLToken.SecretID, "-description=badger-badger-badger", aclRole.ID})
```
DerekStrickland, line 87: 🤣

command/acl_role_update.go
jrasell: I am following what we do for policies for the most part, however, where we do validation is something I always question because it seems there is a little bit here and there. Generally though, I think it would need to be in the HTTP API rather than the CLI so we would cover both CLI and API users in the validation.

I'll resolve this for now, but we should chat internally and come up with a plan around how this should look at how we can incrementally implement it.

Issue #13939, opened 2022-08-01
nomad plan does not recognize VAULT_TOKEN


shoenig:
https://github.com/hashicorp/nomad/issues/13062#issuecomment-1199413980

> there is no -vault-token option for nomad plan

michalgm:
thanks for opening this - just in case it is not clear, it would be ideal if nomad plan read VAULT_TOKEN from the environment if vault-token is not specified, the same as other commands

Issue #13062, opened 2022-05-19
nomad validate does not recognize VAULT_TOKEN


EtienneBruines:
### Nomad version
```
Nomad v1.3.0 (52e95d64113e01be05d585d8b4c07f6f19efebbc)
```

Things worked fine on Nomad 1.2.6, but ever since upgrading to 1.3.0 we've had this issue. ~~Intermittently~~ (The Intermittently was due to a data-race issue where the Nomad tokens issued by Vault weren't valid yet and the `nomad validate` therefore skipped checking completely. The `nomad validate` command ignores my `VAULT_TOKEN` values 100% of the time)

### Operating system and Environment details
Ubuntu 21.10 Impish Indri on amd64

### Issue
Unable to validate jobs with a `vault.policies` value. 

Reproduced on multiple different machines.

### Reproduction steps
Use the example job file below. 

```shell
VAULT_TOKEN="enter-root-token-here" nomad validate job.nomad
```

The problem is not limited to using the vault root token, but also occurs with other valid tokens. Rubbish tokens (e.g. typing `abc` as a token) also give the same error. 

It is later picked up on by `nomad run` - but somehow the `nomad validate` seems to completely ignore the value I provide. 

#### Expected Result
Everything OK. 

#### Actual Result
```
Job validation errors:
1 error occurred:
        * Vault used in the job but missing Vault token
```

### Job file (if appropriate)
```hcl
job "test1" {
  datacenters = ["dc1"]

  group "test2" {
    task "test3" {
      driver = "docker"

      config {
        image = "alpine:latest"
      }

      vault {
        policies = ["my-policy"]
      }
    }
  }
}
```

### Nomad Server logs (if appropriate)
```
2022-05-19T08:03:50.906Z [TRACE] nomad.job: job mutate results: mutator=canonicalize warnings=[] error=<nil>
2022-05-19T08:03:50.907Z [TRACE] nomad.job: job mutate results: mutator=connect warnings=[] error=<nil>
2022-05-19T08:03:50.907Z [TRACE] nomad.job: job mutate results: mutator=expose-check warnings=[] error=<nil>
2022-05-19T08:03:50.907Z [TRACE] nomad.job: job mutate results: mutator=constraints warnings=[] error=<nil>
2022-05-19T08:03:50.907Z [TRACE] nomad.job: job validate results: validator=connect warnings=[] error=<nil>
[...]

PR #14088, opened 2022-08-11
cli: support vault token in plan command

shoenig:
This PR fixes a regression where the 'job plan' command would not respect
a Vault token if set via --vault-token or $VAULT_TOKEN.

Basically the same bug/fix as for the validate command in https://github.com/hashicorp/nomad/issues/13062

Fixes https://github.com/hashicorp/nomad/issues/13939


pkazmierczak (approved):
looks good! 👍

PR #14098, opened 2022-08-12
Solves a harsh bug where you have namespace access but no secvars access

philrenaud:
Twig branch: Solves an issue where you have a policy like this:

```
namespace "*" {
  policy = "read"
  secure_variables {
    # full access to secrets in all project paths
    path "*" {
      capabilities = ["list"]
    }
  }
}

namespace "mad*" {
  policy = "read"
  secure_variables {
    path "*" {
      capabilities = []
    }
  }
}

namespace "nonsense" {
  policy = "write"
  secure_variables {
    # full access to secrets in all project paths
    path "*" {
    }
  }
}
```

This makes it so we correctly return the forbidden state error to the user, rather than throw a fatal exception like we were doing before, when they switch to the `madness` namespace. Applies at both index and path levels.

image

### DIFF
--- a/ui/app/routes/variables.js
+++ b/ui/app/routes/variables.js
@@ -2,6 +2,7 @@ import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
 import WithForbiddenState from 'nomad-ui/mixins/with-forbidden-state';
 import notifyError from 'nomad-ui/utils/notify-error';
+import notifyForbidden from 'nomad-ui/utils/notify-forbidden';
 import PathTree from 'nomad-ui/utils/path-tree';
 
 export default class VariablesRoute extends Route.extend(WithForbiddenState) {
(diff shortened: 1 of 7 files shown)


ui/app/routes/variables.js
```diff
@@ -30,13 +30,13 @@ export default class VariablesRoute extends Route.extend(WithForbiddenState) {
...
       );
-
       return {
         variables,
         pathTree: new PathTree(variables),
       };
     } catch (e) {
-      notifyError(this)(e);
+      notifyForbidden(this)(e);
+      return e;
```
philrenaud, line 39: Passing the error state down to child (path, index) routes so they can run WithForbiddenState against it.

ChaiWithJai: This seems like a lot of extra work for capturing this state. If the error is caught on the `variables` route why can't we just edit the `variables.hbs` template to render the `ForbiddenMessage` else render the `outlet`?

philrenaud: @ChaiWithJai because the ability to filter down / change your namespace exists at /index and /path, but not at the parent level. If we pass the forbidden message up here, the user has no way of escaping it. (say they're on default namespace but only have Secure Vars access on namespace-2)0

ChaiWithJai: OK, no worries.

ChaiWithJai (approved):

PR #14100, opened 2022-08-12
Handle conflict swith a cas qp on save and create

### LINKED ISSUE #13420: Secure Variables: Handle conflicts on save
Secure variables have a `ModifyIndex` that we expose to the UI. We can use this to inform a user that they're saving a variable that, unbeknownst to them, may have been saved by another user in the time they've had their page open.

There are a few ways this could be tackled:

### UI solution:
- Before attempting to PUT a variable, perform a GET first to see if the modifyIndex of the variable you're about to save matches the one from the server.
- If not, present the user with a notification that there's a conflict, and prompting them to either refresh their page or overwrite anyway

### Server-led solution:
- When a PUT comes in for a variable, and it has a modifyIndex that differs from the payload of the PUT, send back a [409](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409) in order to trigger the UI to notify the user / ask if they want to overwrite
- Allow for a `?force=true` header or parameter on PUT that overwrites conflicting variables

cc @angrycub @juliezzhou @mikenomitch @tgross for any suggestions or thoughts on the above ways to handle variable save conflict.

philrenaud:
Resolves #13420

Handles situations in which you try to save a variable, but somebody elsewhere has saved a variable underneath you while you had your page open.

image

## Progress: 
- [x] Apply check-and-set query params
- [x] Let the user know there was a conflict
- [x] Show conflicting version
- [X] Integration tests
- [x] Mirage fixture

### DIFF
--- a/ui/tests/acceptance/secure-variables-test.js
+++ b/ui/tests/acceptance/secure-variables-test.js
@@ -663,8 +663,8 @@ module('Acceptance | secure variables', function (hooks) {
       assert
         .dom('[data-test-file-row]:not(.inaccessible)')
         .exists(
-          { count: 3 },
-          'Shows 3 variable files, none of which are inaccessible'
+          { count: 4 },
+          'Shows 4 variable files, none of which are inaccessible'
         );
 
       await click('[data-test-file-row]');
@@ -683,8 +683,8 @@ module('Acceptance | secure variables', function (hooks) {
       assert
         .dom('[data-test-file-row].inaccessible')
         .exists(
-          { count: 3 },
-          'Shows 3 variable files, all of which are inaccessible'
+          { count: 4 },
+          'Shows 4 variable files, all of which are inaccessible'
         );
 
       // Reset Token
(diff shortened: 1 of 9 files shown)


ChaiWithJai (approved):

ui/tests/acceptance/secure-variables-test.js
```diff
@@ -553,6 +553,43 @@ module('Acceptance | secure variables', function (hooks) {
...
+      assert
+        .dom('.notification.conflict')
+        .exists('Notification alerting user of conflict is present');
+
+      document.querySelector('[data-test-var-key]').value = ''; // clear current input
+      await typeIn('[data-test-var-key]', 'buddy');
+      await typeIn('[data-test-var-value]', 'pal');
+      await click('[data-test-submit-var]');
+
+      await click('button[data-test-overwrite-button');
```
ChaiWithJai, line 575: Great job! Just close this brace out and you should be good.

`button[data-test-overwrite-button` to `button[data-test-overwrite-button]`

PR #14119, opened 2022-08-15
client: fix race in heartbeat tracker

schmichael:
Part 1/n of my effort to make agent tests race free.

PR #14120, opened 2022-08-15
nomad: fix race in Bootstrapped access

schmichael:
Part 2/n of my effort to make agent tests race free. (Prev: #14119)

I gave up on an initial effort of refactoring serf interactions out of Server because it was just going to be too big of a distraction at the moment. I think that's the eventual direction we should head (if we don't just remove serf entirely).


schmichael:
Rebased to get Go 1.19 from #14132

shoenig (approved):
LGTM!

PR #14121, opened 2022-08-15
deployments: fix data races

schmichael:
Part 3/n of my effort to make agent tests race free. (Prev: #14120)

Both priority and state related fields may be mutated concurrently and need to be accessed with the lock acquired.

Our whole approach to flush()/setenabled in these daemon goroutines could be refactored to avoid these locking concerns (eg creating whole new structs instead of re-initializing all fields on "live" structs), but it was out of scope for my refactoring effort.

### DIFF
--- a/nomad/deploymentwatcher/deployments_watcher.go
+++ b/nomad/deploymentwatcher/deployments_watcher.go
@@ -195,10 +195,10 @@ func (w *Watcher) watchDeployments(ctx context.Context) {
 func (w *Watcher) getDeploys(ctx context.Context, minIndex uint64) ([]*structs.Deployment, uint64, error) {
 	// state can be updated concurrently
 	w.l.Lock()
-	state := w.state
+	stateStore := w.state
 	w.l.Unlock()
 
-	resp, index, err := state.BlockingQuery(w.getDeploysImpl, minIndex, ctx)
+	resp, index, err := stateStore.BlockingQuery(w.getDeploysImpl, minIndex, ctx)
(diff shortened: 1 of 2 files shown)


nomad/deploymentwatcher/deployments_watcher.go
```diff
@@ -193,7 +193,12 @@ func (w *Watcher) watchDeployments(ctx context.Context) {
 
 // getDeploys retrieves all deployments blocking at the given index.
 func (w *Watcher) getDeploys(ctx context.Context, minIndex uint64) ([]*structs.Deployment, uint64, error) {
-	resp, index, err := w.state.BlockingQuery(w.getDeploysImpl, minIndex, ctx)
+	// state can be updated concurrently
+	w.l.Lock()
+	state := w.state
+	w.l.Unlock()
+
+	resp, index, err := state.BlockingQuery(w.getDeploysImpl, minIndex, ctx)
```
jrasell, line 201: The `state` variable collides with an import name so maybe we could rename the variable:

```suggestion
	stateStore := w.state
	w.l.Unlock()

	resp, index, err := stateStore.BlockingQuery(w.getDeploysImpl, minIndex, ctx)
```

jrasell (approved):
LGTM with a minor comment about variable name collision.

nomad/deploymentwatcher/deployments_watcher.go
schmichael: Good call, thanks!

Issue #13800, opened 2022-07-18
Nomad Logs, Cat, ReadAt API calls timeout silently and appear unnecessarily slow as a result


robinovitch61:
### Nomad version
v1.3.1

### Operating system and Environment details
MacOS local machine querying Nomad running on Ubuntu in AWS

### Issue
I am moving from directly crafting http queries to the nomad api to using the [provided nomad go api client](https://github.com/hashicorp/nomad/tree/cd047cdc03cdc61c7ebf1ea60615eff96509b606/api), which has all the configuration boilerplate I could want and I am very happy exists!

But, I ran into a hard to track down issue when porting my application to this client. There is an action taken in [fs.queryClientNode](https://github.com/hashicorp/nomad/blob/cd047cdc03cdc61c7ebf1ea60615eff96509b606/api/fs.go#L178-L210) that silently slows down the [Logs](https://github.com/hashicorp/nomad/blob/cd047cdc03cdc61c7ebf1ea60615eff96509b606/api/fs.go#L233), [Cat](https://github.com/hashicorp/nomad/blob/cd047cdc03cdc61c7ebf1ea60615eff96509b606/api/fs.go#L109), and [ReadAt](https://github.com/hashicorp/nomad/blob/cd047cdc03cdc61c7ebf1ea60615eff96509b606/api/fs.go#L97) api calls in some situations. It tries to query from the node first, with a default timeout of 1s. If that times out, it falls back to querying the server, which returns very quickly. In my case, it was timing out every time silently, then falling back to the server call and returning quickly, but it took awhile of looking at the code and realizing that's what was occurring before I was able to grok everything.

After realizing this was the case, I had to add the following in my application before my call to `Logs`
```
// This is currently very important and strange. The logs api attempts to go through the node directly
// by default. The default timeout for this is 1 second. If it fails, it falls silently to going through
// the server. Since it always fails, at least in my Nomad setup, make it timeout immediately by setting
// the timeout to something tiny.
api.ClientConnTimeout = 1 * time.Microsecond
```

### Reproduction steps
On a nomad server where node queries consistently fail/timeout, hit the logs API, and note that it could never be faster than 1s roundtrip time unless you manually set `api.ClientConnTimeout` to something very small before calling it.

#### Expected Result
Logs api is as fast as possible without a silent 1s timeout.

#### Actual Result
[...]

PR #14122, opened 2022-08-15
api: document warnings for setting `api.ClientConnTimeout`

tgross:
HTTP API consumers that have network line-of-sight to client nodes can connect
directly for a small number of APIs. But in environments where the consumer
doesn't have line-of-sight, there's a long pause waiting for the
`api.ClientConnTimeout` to expire. Warn about this in the API docs so that
authors can avoid the extra timeout.

Closes https://github.com/hashicorp/nomad/issues/13800
No changelog for this, because it's entirely golang API docs.


DerekStrickland (approved):
LGTM

Issue #14038, opened 2022-08-05
secure vars ACL policy parsing allows invalid policies


tgross:
The following invalid policy will be parsed successfully, but not allow the expected access to all paths in the `dev` namespace:

```hcl
namespace "dev" {
  secure_variables "*" {
      capabilities = ["read", "write"]
  }
}
```

The intended policy is:

```hcl
namespace "dev" {
  secure_variables {
    path  "*" {
      capabilities = ["read", "write"]
    }
  }
}
```

This is particularly problematic when combined with a valid wildcard policy with wider permissions, like the following:

```hcl
namespace "*" {
  secure_variables {
    path  "*" {
      capabilities = ["read", "list"]
    }
  }
}

namespace "prod" {
  secure_variables "*" {
      capabilities = ["deny"]
  }
}
```

In this case, the `prod` denial will be broken. We should tighten up the parser to prevent this kind user error from happening silently.

PR #14123, opened 2022-08-15
ACL: disallow missing `path` in secure variable policy

tgross:
Fixes https://github.com/hashicorp/nomad/issues/14038

The HCL parser allows for labels that aren't needed, which makes it easy to
accidentally write a `secure_variable` block that has the intended path as the
label for that block instead of the innner `path` block. This can result in
silent failure to lock down variables if an incorrectly specified block was used
to reduce the scope of capabilities (for example, if another correctly-written
rule allows access to `*`).

We can't detect the extraneous label in the HCL API, but we can detect if we're
missing `path` blocks entirely. Use this to block obvious user errors.

### DIFF
--- a/acl/policy.go
+++ b/acl/policy.go
@@ -328,14 +328,17 @@ func Parse(rules string) (*Policy, error) {
 		}
 
 		if ns.SecureVariables != nil {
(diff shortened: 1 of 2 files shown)


DerekStrickland (approved):
👍

Issue #10578, opened 2021-05-13
Restarting allocations does not seems to respect lifecycle and shudown_delay constraints


scyd:
### Nomad version
Nomad 1.0.2

### Operating system and Environment details
CentOS 8

### Issue
when restarting running allocation via GUI or CLI:

  1. shudown_delay for the task is not applicable when it is killed 
  2. logic seems to be stopping and starting for each task  without applying lifecycle rules (pre-start tasks first...etc) nor leader flag.

### Reproduction steps

1.  have one task groups with prestarts tasks, leader tasks and shutdown_delay tasks.
2. restarts the allocation

#### Expected Result
Expecting Nomad to :

1. Stop all the tasks applying shutdown_delay if specified (like a standard allocation stop)
2. Once all tasks stopped/dead , start the tasks applying lifecycle rules and leader tag. 

#### Actual Result
no order in restarting tasks.

### Job file (if appropriate)
![Screen Shot 2021-05-12 at 9 50 04 PM](https://user-images.githubusercontent.com/7786856/118081714-c8b5d180-b370-11eb-80a3-37c45c91d639.png)

If possible please post relevant logs in the issue.
```
2021-05-13T04:53:29.705Z [INFO]  client.alloc_runner.task_runner: restarting task: alloc_id=fa89cf36-e47b-c9ea-a9c8-c1d4ffcef1d0 task=main_task2 reason= delay=0s
	client.alloc_runner.task_runner: running exited hook: alloc_id=fa89cf36-e47b-c9ea-a9c8-c1d4ffcef1d0 task=prestart_with_shutdown_delay_30_leader name=stats_hook start="2021-05-13 04:53:26.707639186 +0000 UTC m=+35.630485729"
	client.alloc_runner.task_runner: finished exited hooks: alloc_id=fa89cf36-e47b-c9ea-a9c8-c1d4ffcef1d0 task=prestart_with_shutdown_delay_30_leader name=stats_hook end="2021-05-13 04:53:26.7076722 +0000 UTC m=+35.630518742" duration=33.013µs
	client.alloc_runner.task_runner: running exited hook: alloc_id=fa89cf36-e47b-c9ea-a9c8-c1d4ffcef1d0 task=prestart_with_shutdown_delay_30_leader name=consul_services start="2021-05-13 04:53:26.707691538 +0000 UTC m=+35.630538077"
[...]

PR #14127, opened 2022-08-16
Task lifecycle restart

### LINKED ISSUE #9464: Alloc restart does not restart one-shot lifecycle tasks
### Nomad version
```
Nomad v1.0.0-beta3 (fcb32ef7ba037b2b83e7a0fd1c53c7410a2990db)
```

### Issue
The alloc restart command optimistically attempts to restart all tasks in the allocation; it collects any task-level errors and returns them:
https://github.com/hashicorp/nomad/blob/v1.0.0-beta3/client/allocrunner/alloc_runner.go#L1173-L1178

For an allocation with a dead `poststart` task, the main task will be restarted, but the `poststart` task will fail to restart with `Task not running` (because it's dead). 

There are a few changes we could make here:
* There is an argument that `poststart` tasks should be re-run when the main tasks are restarted. 
* Otherwise, I think there's an argument that `RestartAll` should not return an error; for an alloc-level restart, we should not attempt to restart dead `poststart` tasks
* Even if we keep the behavior the same, the error message could be improved to indicate *which task* failed to restart. Currently, if there are are multiple dead `poststart` tasks, each will return an error on restart, resulting in the increasingly unhelpful error message: 
```
$ nomad alloc restart bab9
Failed to restart allocation:

Unexpected response code: 500 (2 errors occurred:
	* Task not running
	* Task not running

)
```

### Reproduction steps
1. Run the job below
2. After the poststart allocation has stopped, run `nomad alloc restart <alloc-id>`
3. Note the error message

### Job file (if appropriate)
```
job "repro7875" {
  type = "service"
  datacenters = ["dc1"]
  group "repro" {
    task "main" {
      driver = "exec"
      config {
        command = "sleep"
        args = ["3600"]
      }  
    }
    task "poststart" {
      driver = "exec"
      config {
        command = "env"
      }
      lifecycle {
        hook = "poststart"
      }
    }
  }
}
```

If possible please post relevant logs in the issue.

Logs and other artifacts may also be sent to: nomad-oss-debug@hashicorp.com

Please link to your Github issue in the email and reference it in the subject
line:

> To: nomad-oss-debug@hashicorp.com
>
> Subject: GH-1234: Errors garbage collecting allocs

Emails sent to that address are readable by all HashiCorp employees but are *not* publicly visible.

### LINKED ISSUE #9688: poststart task completion does not impede poststop tasks
### Nomad version
1.0.1

### Issue
I was curious about whether tasks with lifecycle hook `poststop` should wait for tasks with lifecycle hook `poststart` to be dead? Is there a plan to support this "hard-gating" functionality?

### LINKED ISSUE #9841: Nomad prestart task breaks "alloc restart"
### Nomad version
Nomad v0.12.0 (8f7fbc8e7b5a4ed0d0209968faf41b238e6d5817)

CentOs 7.5 VM env on Virtualbox 6.1

### Issue
Nomad alloc restart not working with `prestart` lifecycle task

Expected behaviour, alloc restart the `prestart` task and then the `main` task. 

### Reproduction steps
- Run the nomad task with the following task file
```
 nomad job status test
ID            = test
Name          = test
Submit Date   = 2021-01-18T11:47:41Z
Type          = service
Priority      = 50
Datacenters   = dc1
Namespace     = default
Status        = running
Periodic      = false
Parameterized = false

Summary
Task Group  Queued  Starting  Running  Failed  Complete  Lost
test        0       0         1        0       0         0

Latest Deployment
ID          = ff4a14b1
Status      = running

Deployed
test        1        1       0        0          2021-01-18T11:57:41Z

Allocations
ID        Node ID   Task Group  Version  Desired  Status   Created  Modified
df20b14a  1d358dc0  test        0        run      running  15s ago  7s ago

```
- Run `nomad alloc restart`
```
nomad alloc restart df20b14a
Failed to restart allocation:

Unexpected response code: 500 (rpc error: 2 errors occurred:
	* Task not running
	* Task not running

)
```

### Job file (if appropriate)
```
job "test" {
  datacenters = ["dc1"]
  type = "service"

  group "test" {
    restart {
      interval = "6m"
      attempts = 10
      delay    = "10s"
      mode     = "delay"
    }

    # add prestart task
    task "test-pre" {
      driver = "docker"
      lifecycle {
        hook = "prestart"
        sidecar = false
      }

      config {
        image = "alpine:3.8"
        command = "sh"

        args = ["-c", "echo test > /alloc/test_file"]
      }
    }

    task "test" {
      driver = "docker"

      config {
        image = "alpine:3.8"
        command = "sh"

        args = ["-c", "if [ ! -s /alloc/test_file ]; then sleep 5; exit 1; else while sleep 3600; do :; done; fi"]
      }
    }
  }
}
```

lgfa29:
Following-up on the work done in #14009, this PR implements a new restart mode that allows for all tasks of an allocation to be restarted, even those that have already run, such as non-sidecar prestart or poststop tasks.

It also solves some related issues, such as the restart command failing with `Task not running` due to dead tasks in the allocation. These errors are now ignore when restarting the allocation (restarting a dead task with `-task` will still result in this error).

Closes #9464
Closes #9688
Closes #9841

-----------

_Note to reviewers_: The internal RFC proposed a new task state (`complete`) to differentiate between a task that is really `dead` (and would never run again) from that a task that finished running, but is waiting for a restart. 

But during implementation I noticed that there are quite a few places where the `dead` state was being checked, but then I also noticed that it didn't really matter if the task was `dead` or `complete` so I was able to implement this functionally without the need for the extra state.

### DIFF
--- a/.changelog/14127.txt
+++ b/.changelog/14127.txt
@@ -1,0 +1,7 @@
+```release-note:improvement
+client: add option to restart all tasks of an allocation, even if the task already run, such as non-sidecar prestart and poststart tasks.
+```
+
+```release-note:improvement
+client: only start poststop tasks after poststart tasks are done.
+```
--- a/.changelog/14127.txt
+++ b/.changelog/14127.txt
@@ -1,5 +1,5 @@
 ```release-note:improvement
-client: add option to restart all tasks of an allocation, even if the task already run, such as non-sidecar prestart and poststart tasks.
+client: add option to restart all tasks of an allocation, regardless of lifecycle type or state.
 ```
 
 ```release-note:improvement
--- a/api/allocations.go
+++ b/api/allocations.go
@@ -156,6 +156,16 @@ func (a *Allocations) Restart(alloc *Allocation, taskName string, q *QueryOption
 	return err
 }
 
+func (a *Allocations) RestartAllTasks(alloc *Allocation, q *QueryOptions) error {
+	req := AllocationRestartRequest{
+		AllTasks: true,
+	}
+
+	var resp struct{}
+	_, err := a.client.putQuery("/v1/client/allocation/"+alloc.ID+"/restart", &req, &resp, q)
+	return err
+}
+
 // Stop stops an allocation.
 //
 // Note: for cluster topologies where API consumers don't have network access to
@@ -447,6 +457,7 @@ func (a Allocation) RescheduleInfo(t time.Time) (int, int) {
 
 type AllocationRestartRequest struct {
 	TaskName string
+	AllTasks bool
 }
 
 type AllocSignalRequest struct {
--- a/api/allocations.go
+++ b/api/allocations.go
@@ -141,7 +141,9 @@ func (a *Allocations) GC(alloc *Allocation, q *QueryOptions) error {
 	return err
 }
 
-// Restart restarts an allocation.
+// Restart restarts the tasks that are currently running or a specific task if
+// taskName is provided. An error is returned if the task to be restarted is
+// not running.
 //
 // Note: for cluster topologies where API consumers don't have network access to
 // Nomad clients, set api.ClientConnTimeout to a small value (ex 1ms) to avoid
@@ -156,6 +158,12 @@ func (a *Allocations) Restart(alloc *Allocation, taskName string, q *QueryOption
 	return err
 }
 
+// RestartAllTasks restarts all tasks in the allocation, regardless of
+// lifecycle type or state. Tasks will restart following their lifecycle order.
+//
+// Note: for cluster topologies where API consumers don't have network access to
+// Nomad clients, set api.ClientConnTimeout to a small value (ex 1ms) to avoid
+// long pauses on this API call.
 func (a *Allocations) RestartAllTasks(alloc *Allocation, q *QueryOptions) error {
 	req := AllocationRestartRequest{
 		AllTasks: true,
--- a/client/alloc_endpoint_test.go
+++ b/client/alloc_endpoint_test.go
@@ -68,6 +68,45 @@ func TestAllocations_Restart(t *testing.T) {
 	})
 }
 
+func TestAllocations_RestartAllTasks(t *testing.T) {
+	ci.Parallel(t)
+
+	require := require.New(t)
+	client, cleanup := TestClient(t, nil)
+	defer cleanup()
+
+	alloc := mock.LifecycleAlloc()
+	require.Nil(client.addAlloc(alloc, ""))
+
+	// Can't restart all tasks while specifying a task name.
+	req := &nstructs.AllocRestartRequest{
+		AllocID:  alloc.ID,
+		AllTasks: true,
+		TaskName: "web",
+	}
+	var resp nstructs.GenericResponse
+	err := client.ClientRPC("Allocations.Restart", &req, &resp)
+	require.Error(err)
+
+	// Good request.
+	req = &nstructs.AllocRestartRequest{
+		AllocID:  alloc.ID,
+		AllTasks: true,
+	}
+
+	testutil.WaitForResult(func() (bool, error) {
+		var resp2 nstructs.GenericResponse
+		err := client.ClientRPC("Allocations.Restart", &req, &resp2)
+		if err != nil && strings.Contains(err.Error(), "not running") {
+			return false, err
+		}
+
+		return true, nil
+	}, func(err error) {
+		t.Fatalf("err: %v", err)
+	})
+}
+
 func TestAllocations_Restart_ACL(t *testing.T) {
 	ci.Parallel(t)
 	require := require.New(t)
--- a/client/allocrunner/alloc_runner.go
+++ b/client/allocrunner/alloc_runner.go
@@ -546,40 +546,63 @@ func (ar *allocRunner) handleTaskStateUpdates() {
 			}
 		}
 
-		// if all live runners are sidecars - kill alloc
-		if killEvent == nil && hasSidecars && !hasNonSidecarTasks(liveRunners) {
-			killEvent = structs.NewTaskEvent(structs.TaskMainDead)
-		}
-
-		// If there's a kill event set and live runners, kill them
-		if killEvent != nil && len(liveRunners) > 0 {
-
-			// Log kill reason
-			switch killEvent.Type {
-			case structs.TaskLeaderDead:
-				ar.logger.Debug("leader task dead, destroying all tasks", "leader_task", killTask)
-			case structs.TaskMainDead:
-				ar.logger.Debug("main tasks dead, destroying all sidecar tasks")
-			default:
-				ar.logger.Debug("task failure, destroying all tasks", "failed_task", killTask)
+		if len(liveRunners) > 0 {
+			// if all live runners are sidecars - kill alloc
+			if killEvent == nil && hasSidecars && !hasNonSidecarTasks(liveRunners) {
+				killEvent = structs.NewTaskEvent(structs.TaskMainDead)
 			}
 
-			// Emit kill event for live runners
-			for _, tr := range liveRunners {
-				tr.EmitEvent(killEvent)
+			// If there's a kill event set and live runners, kill them
+			if killEvent != nil {
+
+				// Log kill reason
+				switch killEvent.Type {
+				case structs.TaskLeaderDead:
+					ar.logger.Debug("leader task dead, destroying all tasks", "leader_task", killTask)
+				case structs.TaskMainDead:
+					ar.logger.Debug("main tasks dead, destroying all sidecar tasks")
+				default:
+					ar.logger.Debug("task failure, destroying all tasks", "failed_task", killTask)
+				}
+
+				// Emit kill event for live runners
+				for _, tr := range liveRunners {
+					tr.EmitEvent(killEvent)
+				}
+
+				// Kill 'em all
+				states = ar.killTasks()
+
+				// Wait for TaskRunners to exit before continuing to
+				// prevent looping before TaskRunners have transitioned
+				// to Dead.
+				for _, tr := range liveRunners {
+					ar.logger.Info("killing task", "task", tr.Task().Name)
+					select {
+					case <-tr.WaitCh():
+					case <-ar.waitCh:
+					}
+				}
 			}
+		} else {
+			// If there are no live runners left kill all non-poststop task
+			// runners to unblock them from the alloc restart loop.
+			for _, tr := range ar.tasks {
+				if tr.IsPoststopTask() {
+					continue
+				}
 
-			// Kill 'em all
-			states = ar.killTasks()
-
-			// Wait for TaskRunners to exit before continuing to
-			// prevent looping before TaskRunners have transitioned
-			// to Dead.
-			for _, tr := range liveRunners {
-				ar.logger.Info("killing task", "task", tr.Task().Name)
 				select {
 				case <-tr.WaitCh():
 				case <-ar.waitCh:
+				default:
+					// Kill task runner without setting an event because the
+					// task is already dead, it's just waiting in the alloc
+					// restart loop.
+					err := tr.Kill(context.TODO(), nil)
+					if err != nil {
+						ar.logger.Warn("failed to kill task", "task", tr.Task().Name, "error", err)
+					}
 				}
 			}
 		}
--- a/client/allocrunner/alloc_runner.go
+++ b/client/allocrunner/alloc_runner.go
@@ -548,7 +548,8 @@ func (ar *allocRunner) handleTaskStateUpdates() {
 
 		if len(liveRunners) > 0 {
 			// if all live runners are sidecars - kill alloc
-			if killEvent == nil && hasSidecars && !hasNonSidecarTasks(liveRunners) {
+			hasLiveSidecars := hasSidecars && !hasNonSidecarTasks(liveRunners)
+			if killEvent == nil && hasLiveSidecars {
 				killEvent = structs.NewTaskEvent(structs.TaskMainDead)
 			}
 
@@ -573,11 +574,11 @@ func (ar *allocRunner) handleTaskStateUpdates() {
 				// Kill 'em all
 				states = ar.killTasks()
 
-				// Wait for TaskRunners to exit before continuing to
-				// prevent looping before TaskRunners have transitioned
-				// to Dead.
+				// Wait for TaskRunners to exit before continuing. This will
+				// prevent looping before TaskRunners have transitioned to
+				// Dead.
 				for _, tr := range liveRunners {
-					ar.logger.Info("killing task", "task", tr.Task().Name)
+					ar.logger.Info("waiting for task to exit", "task", tr.Task().Name)
 					select {
 					case <-tr.WaitCh():
 					case <-ar.waitCh:
@@ -1228,20 +1229,24 @@ func (ar *allocRunner) GetTaskEventHandler(taskName string) drivermanager.EventH
 }
 
 // Restart satisfies the WorkloadRestarter interface and restarts all tasks
-// that are currently running. Only the TaskRestartRunningSignal event type may
-// be used.
+// that are currently running.
+//
+// The event type will be set to TaskRestartRunningSignal to comply with
+// internal restart logic requirements.
 func (ar *allocRunner) Restart(ctx context.Context, event *structs.TaskEvent, failure bool) error {
 	if event.Type != structs.TaskRestartRunningSignal {
-		return fmt.Errorf("Invalid event %s for alloc restart request", event.Type)
+		event.Type = structs.TaskRestartRunningSignal
 	}
 	return ar.restartTasks(ctx, event, failure)
 }
 
-// RestartTask restarts the provided task. Only TaskRestartSignal event type
-// may be used.
+// RestartTask restarts the provided task.
+//
+// The event type will be set to TaskRestartSignal to comply with internal
+// restart logic requirements.
 func (ar *allocRunner) RestartTask(taskName string, event *structs.TaskEvent) error {
 	if event.Type != structs.TaskRestartSignal {
-		return fmt.Errorf("Invalid event %s for task restart request", event.Type)
+		event.Type = structs.TaskRestartSignal
 	}
 
 	tr, ok := ar.tasks[taskName]
@@ -1252,11 +1257,13 @@ func (ar *allocRunner) RestartTask(taskName string, event *structs.TaskEvent) er
 	return tr.Restart(context.TODO(), event, false)
 }
 
-// RestartRunning restarts all tasks that are currently running. Only the
-// TaskRestartRunningSignal event type may be used.
+// RestartRunning restarts all tasks that are currently running.
+//
+// The event type will be set to TaskRestartRunningSignal to comply with
+// internal restart logic requirements.
 func (ar *allocRunner) RestartRunning(event *structs.TaskEvent) error {
 	if event.Type != structs.TaskRestartRunningSignal {
-		return fmt.Errorf("Invalid event %s for running tasks restart request", event.Type)
+		event.Type = structs.TaskRestartRunningSignal
 	}
 	return ar.restartTasks(context.TODO(), event, false)
 }
--- a/client/allocrunner/alloc_runner.go
+++ b/client/allocrunner/alloc_runner.go
@@ -1230,25 +1230,12 @@ func (ar *allocRunner) GetTaskEventHandler(taskName string) drivermanager.EventH
 
 // Restart satisfies the WorkloadRestarter interface and restarts all tasks
 // that are currently running.
-//
-// The event type will be set to TaskRestartRunningSignal to comply with
-// internal restart logic requirements.
 func (ar *allocRunner) Restart(ctx context.Context, event *structs.TaskEvent, failure bool) error {
-	if event.Type != structs.TaskRestartRunningSignal {
-		event.Type = structs.TaskRestartRunningSignal
-	}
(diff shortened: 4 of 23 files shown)


api/allocations.go
```diff
@@ -127,6 +127,16 @@ func (a *Allocations) Restart(alloc *Allocation, taskName string, q *QueryOption
 	return err
 }
 
+func (a *Allocations) RestartAllTasks(alloc *Allocation, q *QueryOptions) error {
```
lgfa29, line 130: I created a new method here to avoid an `api` package breaking change in the `Restart` method above.

client/allocrunner/alloc_runner.go
```diff
@@ -1205,19 +1227,55 @@ func (ar *allocRunner) GetTaskEventHandler(taskName string) drivermanager.EventH
...
 
-// RestartTask signalls the task runner for the  provided task to restart.
-func (ar *allocRunner) RestartTask(taskName string, taskEvent *structs.TaskEvent) error {
+// Restart satisfies the WorkloadRestarter interface and restarts all tasks
+// that are currently running. Only the TaskRestartRunningSignal event type may
+// be used.
+func (ar *allocRunner) Restart(ctx context.Context, event *structs.TaskEvent, failure bool) error {
+	if event.Type != structs.TaskRestartRunningSignal {
+		return fmt.Errorf("Invalid event %s for alloc restart request", event.Type)
+	}
```
lgfa29, line 1236: I was wondering if it would be better to just overwrite `event.Type` here since I don't think it _really_ matters for callers, but I may be missing something so it would be good to hear from others.

client/allocrunner/alloc_runner_test.go
```diff
@@ -483,6 +484,468 @@ func TestAllocRunner_Lifecycle_Poststop(t *testing.T) {
...
+				"prestart-oneshot":  structs.TaskState{State: "dead", Restarts: 0},
+				"prestart-sidecar":  structs.TaskState{State: "running", Restarts: 0},
+				"poststart-oneshot": structs.TaskState{State: "dead", Restarts: 0},
+				"poststart-sidecar": structs.TaskState{State: "running", Restarts: 0},
+				"poststop":          structs.TaskState{State: "pending", Restarts: 0},
+			},
+		},
+		{
+			name:      "restart leader main task",
+			hasLeader: true,
```
lgfa29, line 644: These tests were pulled from #10785 where, in these leader tests, it [mentions](https://github.com/hashicorp/nomad/pull/10785/files#diff-95deb9e4e5232c48e5c186b01e002fd8689a2d38bcee018c08548401b25daa25R1674) that: 

> // TODO: currently fails because sidecars aren't restarted

I think this was to test a fix for https://github.com/hashicorp/nomad/issues/10578 which would synchronize the lifecycle of sidecar tasks with the leader, but this was not implemented here, so I adjusted the test to match the current behaviour (restarting the leader will _not_ restart sidecars). Once a fix for #10578 is done these tests will fail and will need to be adjusted.

client/allocrunner/alloc_runner_unix_test.go
```diff
@@ -154,9 +154,10 @@ func TestAllocRunner_Restore_CompletedBatch(t *testing.T) {
 	alloc := mock.Alloc()
 	alloc.Job.Type = structs.JobTypeBatch
 	task := alloc.Job.TaskGroups[0].Tasks[0]
-	task.Driver = "mock_driver"
+	task.Driver = "raw_exec"
 	task.Config = map[string]interface{}{
-		"run_for": "2ms",
+		"command": "sleep",
+		"args":    []string{"2"},
```
lgfa29, line 160: I had to use the `raw_exec` driver instead of `mock_driver` here and in the other restore test (`TestTaskRunner_Restore_Dead`) because, for some reason, the `mock_driver` behaved differently from others.

For some reason the task handle was not `nil` when restoring a `dead` task, so the `Run` method would get stuck here, with `handleKill` never exiting: https://github.com/hashicorp/nomad/blob/4ac9de1aa619c852b31e02c809f5a4b30299a230/client/allocrunner/taskrunner/task_runner.go#L673-L674

I'm not sure if it's a problem with the `mock_driver` or my task restore changes.

client/allocrunner/taskrunner/sids_hook_test.go
```diff
@@ -297,11 +296,7 @@ func TestTaskRunner_DeriveSIToken_UnWritableTokenFile(t *testing.T) {
 	go tr.Run()
 
 	// wait for task runner to finish running
-	select {
-	case <-tr.WaitCh():
-	case <-time.After(time.Duration(testutil.TestMultiplier()*15) * time.Second):
-		r.Fail("timed out waiting for task runner")
-	}
+	testWaitForTaskToDie(t, tr)
```
lgfa29, line 299: Changes like this had to be done because the task runner `Run()` method doesn't return anymore, it gets blocked waiting for a possible restart, so `WaitCh` is not closed until the task reaches a point where it will not restart anymore.

client/allocrunner/taskrunner/task_runner.go
```diff
@@ -505,12 +514,14 @@ func (tr *TaskRunner) Run() {
 	var result *drivers.ExitResult
 
 	tr.stateLock.RLock()
-	dead := tr.state.State == structs.TaskStateDead
+	restoredDead := tr.state.State == structs.TaskStateDead
```
lgfa29, line 517: As I mentioned in https://github.com/hashicorp/nomad/pull/14127#discussion_r948582301, I'm not sure if the restore logic is 100% correct. It would be good to investigate this more.

client/allocrunner/taskrunner/task_runner.go
```diff
@@ -505,12 +514,14 @@ func (tr *TaskRunner) Run() {
...
 	tr.stateLock.RUnlock()
 
-	// if restoring a dead task, ensure that task is cleared and all post hooks
-	// are called without additional state updates
-	if dead {
+	// If restoring a dead task, ensure that task is cleared and all post hooks
+	// are called without additional state updates.
+	// If the alloc is not terminal we must proceed until the ALLOC_RESTART
+	// loop to allow the task to run again in case the alloc is restarted.
+	if restoredDead && tr.Alloc().TerminalStatus() {
```
tgross, line 524: This includes both the server intent and the current client state, which means we're relying on one of them to get the updated allocation. I think if we hit this for a task before the client status is updated we'll end up falling past the driver handle cleanup, which might explain the results you're seeing with the mock driver.

client/allocrunner/taskrunner/task_runner.go
tgross: Commented a bit further down on the `restoredDead` check.

Also, I'm not sure I understand the changed name. Calling it `restoredDead` reads to me like it's been restored in the past from the dead state. But I think the code change means "it's _currently_ dead but eligible to be restored"? `deadAndCanBeRestored` is kind of wordy but maybe there's something better that's not coming to mind?

client/allocrunner/alloc_runner.go
tgross: Normally we'd probably validate this at the RPC handler but in this case the same handler gets used for all three methods. That being said, we switch on the method based on `Event.Type`, right? So if it's wrong down here in the alloc runner that's a programmer's error where I'd expect it's fine to assert the right value via overwriting `event.Type`.

tgross (changes requested):
This is looking really good @lgfa29! I think it's probably worth hunting down that `rawexec` vs `mock` behavior before merging though.

api/allocations.go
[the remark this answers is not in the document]
lgfa29: Good point, I will add it to both methods 👍

command/alloc_restart.go
```diff
@@ -31,9 +34,14 @@ General Options:
 
 Restart Specific Options:
 
+  -all-tasks
+    If set, all tasks in the allocation will be restarted, even the ones that
+    already ran. This option cannot be used if a task is defined.
```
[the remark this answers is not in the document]
lgfa29: I avoided `-task` here because the task name can also be passed as an argument (`nomad alloc restart 1234 web`) but I can mention both.

.changelog/14127.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:improvement
+client: add option to restart all tasks of an allocation, even if the task already run, such as non-sidecar prestart and poststart tasks.
```
shoenig, line 2:
```suggestion
client: add option to restart all tasks of an allocation, regardless of lifecycle type or state.
```

i dunno, seems shorter

.changelog/14127.txt
```diff
@@ -0,0 +1,7 @@
...
+```
+
+```release-note:improvement
+client: only start poststop tasks after poststart tasks are done.
+```
```
shoenig, line 7: bug or feature?

api/allocations.go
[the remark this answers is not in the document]
shoenig: :headstone: godoc.org

client/alloc_endpoint_test.go
```diff
@@ -68,6 +68,45 @@ func TestAllocations_Restart(t *testing.T) {
...
+		var resp2 nstructs.GenericResponse
+		err := client.ClientRPC("Allocations.Restart", &req, &resp2)
+		if err != nil && strings.Contains(err.Error(), "not running") {
+			return false, err
+		}
+
+		return true, nil
+	}, func(err error) {
+		t.Fatalf("err: %v", err)
+	})
```
shoenig, line 107: looks like this is copied form elsewhere, can it be extracted into a helper function?

client/allocrunner/alloc_runner.go
```diff
@@ -547,40 +546,63 @@ func (ar *allocRunner) handleTaskStateUpdates() {
...
-			switch killEvent.Type {
-			case structs.TaskLeaderDead:
-				ar.logger.Debug("leader task dead, destroying all tasks", "leader_task", killTask)
-			case structs.TaskMainDead:
-				ar.logger.Debug("main tasks dead, destroying all sidecar tasks")
-			default:
-				ar.logger.Debug("task failure, destroying all tasks", "failed_task", killTask)
+		if len(liveRunners) > 0 {
+			// if all live runners are sidecars - kill alloc
+			if killEvent == nil && hasSidecars && !hasNonSidecarTasks(liveRunners) {
```
shoenig, line 551: could the double negative here be wrapped in a `hasLiveSidecars(...)` helper?

client/allocrunner/alloc_runner.go
```diff
@@ -547,40 +546,63 @@ func (ar *allocRunner) handleTaskStateUpdates() {
...
+				}
 
-			// Kill 'em all
-			states = ar.killTasks()
+				// Emit kill event for live runners
+				for _, tr := range liveRunners {
+					tr.EmitEvent(killEvent)
+				}
+
+				// Kill 'em all
```
shoenig, line 573: :scream:

client/allocrunner/alloc_runner.go
```diff
@@ -547,40 +546,63 @@ func (ar *allocRunner) handleTaskStateUpdates() {
...
+				states = ar.killTasks()
+
+				// Wait for TaskRunners to exit before continuing to
+				// prevent looping before TaskRunners have transitioned
+				// to Dead.
```
shoenig, line 578:
```suggestion
				// Wait for TaskRunners to exit before continuing. This will
				// prevent looping before TaskRunners have transitioned
				// to Dead.
```

client/allocrunner/alloc_runner.go
```diff
@@ -547,40 +546,63 @@ func (ar *allocRunner) handleTaskStateUpdates() {
...
+				for _, tr := range liveRunners {
+					ar.logger.Info("killing task", "task", tr.Task().Name)
```
shoenig, line 580: more clear if we say like "waiting for task to exit ..."

client/allocrunner/alloc_runner_test.go
```diff
@@ -1213,7 +1676,11 @@ func TestAllocRunner_MoveAllocDir(t *testing.T) {
 	ar.Run()
 	defer destroy(ar)
 
-	require.Equal(t, structs.AllocClientStatusComplete, ar.AllocState().ClientStatus)
+	testutil.WaitForResult(func() (bool, error) {
+		return ar.AllocState().ClientStatus == structs.AllocClientStatusComplete, nil
+	}, func(_ error) {
+		t.Fatalf("expected alloc to be complete")
+	})
```
shoenig, line 1683: copied below; make a helper?

client/allocrunner/alloc_runner_unix_test.go
shoenig: Probably worth getting to the bottom of this; just debug logging the plugin interface entrypoints and the parameters being passed in should shed light on what's happening

lgfa29: Yup, thanks for the push back, my implementation was wrong 😅

client/allocrunner/taskrunner/task_runner.go
lgfa29: Yeah, good point. What got me confused initially was that this variable was never updated, so it was only `true` when restoring a dead task. 

Since my initial implementation was incorrect there's now another value that needs to be checked here, so I reverted this one to `dead`.

client/allocrunner/taskrunner/task_runner.go
lgfa29: Nice find. I'm not 100% if this was the problem with the tests, but it made me realize that important information here is actually whether the `Run()` method exited or not. Using the local client state fixed the problems with the `mock` driver 🎉

.changelog/14127.txt
lgfa29: Thanks!

.changelog/14127.txt
lgfa29: We never specified the expected behaviour here, so it's not like it was doing something wrong. The code change will now always follow this order, so I think it' an `improvement`.

client/alloc_endpoint_test.go
lgfa29: Ah yeah, it came from [here](https://github.com/hashicorp/nomad/blob/main/client/alloc_endpoint_test.go#L58-L68). But I'm not sure if it's generic enough to be extracted as a helper? The two tests happen to call the `Allocations.Restart` RPC, but interpretation of the result may change if the code changes.

client/allocrunner/alloc_runner.go
lgfa29: Hum...I don't think it can be extracted into a method because the live or not condition is determined by the method input: here it's passing a map of live runners, so the result will be live runners that are not sidecars.

I could move it to a intermediate variable is more descriptive: 

```go
hasLiveSidecars := hasSidecars && !hasNonSidecarTasks(liveRunners)
if killEvent == nil && hasLiveSidecars {
	killEvent = structs.NewTaskEvent(structs.TaskMainDead)
}
```

Does this look better?

lgfa29:
@tgross @shoenig I think https://github.com/hashicorp/nomad/pull/14127/commits/e66625876fbc9c1dfc3f550d889f165e270bbb2b fixed the problems with task restore and I was able to use the `mock` driver in the tests.

https://github.com/hashicorp/nomad/pull/14127/commits/8b813fdae2b88d19035312903d1c957e2d73147f address the rest of the code review changes.

client/allocrunner/taskrunner/task_runner_test.go
```diff
@@ -349,6 +349,87 @@ func TestTaskRunner_Restore_Running(t *testing.T) {
 	assert.Equal(t, 1, started)
 }
 
+// TestTaskRunner_Restore_Dead asserts that restoring a dead task will place it
+// back in the correct state. If the task was waiting for an alloc restart it
+// must be able to be restarted after restore, otherwise a restart must fail.
```
tgross, line 354: Great test!

tgross (approved):
LGTM!

DerekStrickland (approved):
Amazing work! Left some questions/suggestions, but not blocking.

client/allocrunner/alloc_runner.go
```diff
@@ -1205,19 +1228,61 @@ func (ar *allocRunner) GetTaskEventHandler(taskName string) drivermanager.EventH
...
 
-// RestartTask signalls the task runner for the  provided task to restart.
-func (ar *allocRunner) RestartTask(taskName string, taskEvent *structs.TaskEvent) error {
+// Restart satisfies the WorkloadRestarter interface and restarts all tasks
+// that are currently running.
+//
+// The event type will be set to TaskRestartRunningSignal to comply with
+// internal restart logic requirements.
+func (ar *allocRunner) Restart(ctx context.Context, event *structs.TaskEvent, failure bool) error {
+	if event.Type != structs.TaskRestartRunningSignal {
```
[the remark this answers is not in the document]
lgfa29: Hum..not sure if I understood, do you mean something like this?

```go
func (ar *allocRunner) Restart(ctx context.Context, event *structs.TaskEvent, failure bool) error {
	ev := event.Copy()
	ev.Type = structs.TaskRestartRunningSignal
	return ar.restartTasks(ctx, ev, failure)
}
```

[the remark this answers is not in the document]
tgross: I thought the original reason we were overwriting was a form of assertion? If it's not the right task restart signal, it's a bug?

client/allocrunner/taskrunner/lifecycle.go
```diff
@@ -9,25 +9,63 @@ import (
...
+	switch taskState.State {
+	case structs.TaskStatePending:
+		// Tasks that are "pending" are never allowed to restart.
+		return ErrTaskNotRunning
+	case structs.TaskStateDead:
+		// Tasks that are "dead" are only allowed to restart when restarting
+		// all tasks in the alloc, otherwise the taskCoordinator will prevent
+		// it from running again, and if their Run method is still running.
+		if event.Type != structs.TaskRestartAllSignal || localState.RunComplete {
+			return ErrTaskNotRunning
```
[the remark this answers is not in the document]
lgfa29: Hum...I think the error is the same, that task tried to restart but it wasn't running

client/allocrunner/taskrunner/task_runner.go
```diff
@@ -674,6 +685,38 @@ MAIN:
...
+			case <-tr.killCtx.Done():
+				break ALLOC_RESTART
+			case <-tr.shutdownCtx.Done():
+				return
+			case <-tr.restartCh:
+				// Restart without delay since the task is not running anymore.
+				restart, _ := tr.shouldRestart()
+				if restart {
+					// Set runner as not dead to allow the MAIN loop to run.
+					dead = false
```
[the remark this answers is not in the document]
lgfa29: Nope, `dead` is a local variable, so not shared with anything else. It's initialized from state, and that _is_ protected with lock.

client/allocrunner/alloc_runner.go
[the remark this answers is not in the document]
lgfa29: Yeah, my understanding is that, for the caller of the `allocRunner.Restart*` functions, the event type doesn't matter, so overwriting it was a way to "correct" the call if, for some reason, it was done using the wrong type.

There is no valid scenario where you would call, for example, `RestartRunning` with anything other than the `TaskRestartRunningSignal` event type, so instead of returning an error we correct the call.

The `Restart` method specifically is an interesting one. It is used to implement the `WorkloadRestarter` interface (defined [here](https://github.com/hashicorp/nomad/blob/v1.3.3/command/agent/consul/check_watcher.go#L24-L27) and [here](https://github.com/hashicorp/nomad/blob/v1.3.3/client/serviceregistration/service_registration.go#L48-L52)) so, in theory, it could be used to perform different types of restart, but I chose to keep it as-is (restart only tasks that are currently running).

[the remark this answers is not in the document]
DerekStrickland: That's almost what I meant. I was suggesting we make the copy and emit both events. I was thinking that with this implementation you lose the info on which code path triggered the restart.

client/allocrunner/taskrunner/task_runner.go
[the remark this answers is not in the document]
DerekStrickland: Perfect. That was my concern.

client/allocrunner/alloc_runner.go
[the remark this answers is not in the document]
lgfa29: A few things to note here:

- task events are used to surface information to users, they are not something like code tracing.
- event types have a fairly restrict [set of possible values](https://github.com/hashicorp/nomad/blob/v1.3.3/nomad/structs/structs.go#L8005-L8098), so they are not the best to help trace code path.
- events have other, more meaningful fields, like `DisplayMessage`, `Details`, `ReastartReason` etc. and those will be kept from the input.
- task events are stored in state, and we only keep the last 10, so emitting (almost) duplicate events will take up a slot and provide little value to users.

That being said, my implementation does abuse the task event mechanism a bit. I tried to think of an alternative, but couldn't think of one yet. I will take another look to see if I can think of something else.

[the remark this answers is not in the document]
lgfa29: https://github.com/hashicorp/nomad/pull/14127/commits/e620cd187452b45ab04bf65d56aa2f90fabfc9eb removes all of this task event nonsense 😅 

I also got ride of the new task event types to avoid any confusion on the developer side as to which type to use. I think it's easy enough for users to distinguish between them given their [event description](https://github.com/hashicorp/nomad/blob/e620cd187452b45ab04bf65d56aa2f90fabfc9eb/client/client.go#L939-L953):
image

PR #12831, opened 2022-04-29
[ui] General keyboard navigation in the UI

philrenaud:
Resolves #12762 

## A general Keyboard / Keynav service for the Nomad UI
- This allows Nomad users to navigate and operate their UI by use of keyboard alone and provides powerful defaults to move around:
  - absolute routes (such as `g j` to go to jobs index),
  - relative routes (such as `shift + right-arrow` to move to the next subnav section), and
  - dynamic routes (such as `shift + 03` to go to the third link in a given list)
- Creates a new global modal, accessible by pressing `?` anyplace in the app
- Gives users the choice to enable/disable shortcuts, or more advanced, to rebind individual commands to the pattern of their choosing.

## Implementation and extension
- Establishes a `keyboard` service that emits a consumable `buffer` array and `matchedCommand` object
- Listens for keystrokes in any non-input elements
- On each keystroke, if the buffer matches a pattern/sequence in `keyboard.keyCommands`, that command's `action` is run. These actions can be any function, and are commonly `router.transitionTo()` objects.
- Uses [Tether](https://www.npmjs.com/package/tether) to place hints on-screen for when the user holds `Shift`
- Shortcuts can be added by use of modifier (such as [the Go to ACLs link](https://github.com/hashicorp/nomad/pull/12831/files#diff-1ec2d76a07ab84748c05e126dc5a4acd5cc4ff74274d8e7c2232bb46274ea887R63)), or by use of component-level, conditional helper `keyboard-commands` (for example, how we [conditionally add the escape key to Evaluations modal](https://github.com/hashicorp/nomad/pull/12831/files#diff-bc34608303f7472092dae6e0956c0751561489b44098aae229b22be72fb6b94dR2-R4))
- Buffer clears after 0.75sec or when a successful keyboard command is executed

## Side-effects
- In order to facilitate a `u` shortcut to "Go up a level" from nested routes, reworked many otherwise-inline subnavs into their own components, such as the [plugin subnav](https://github.com/hashicorp/nomad/pull/12831/files#diff-ed282c6370801ee51d43fbae8b5f449ee01b3c71e31a6ac1ac3c333c1c587674R1-R6)
- In order to work around an Ember Router bug around LinkTos with query properties when deeply nested, invoked [the classic linkTo component](https://github.com/hashicorp/nomad/pull/12831/files#diff-e59da3ad52b5b03e7a6fdb11e2402d729cd46a998234a63bc409166522c8655bR1-R13) sparingly

[...]

PR #14138, opened 2022-08-16
[ui] general keyboard navigation: 1.3.4 release

philrenaud:
(Port of #12831 for pre-1.4 release)

## A general Keyboard / Keynav service for the Nomad UI
- This allows Nomad users to navigate and operate their UI by use of keyboard alone and provides powerful defaults to move around:
  - absolute routes (such as `g j` to go to jobs index),
  - relative routes (such as `shift + right-arrow` to move to the next subnav section), and
  - dynamic routes (such as `shift + 03` to go to the third link in a given list)
- Creates a new global modal, accessible by pressing `?` anyplace in the app
- Gives users the choice to enable/disable shortcuts, or more advanced, to rebind individual commands to the pattern of their choosing.

## Implementation and extension
- Establishes a `keyboard` service that emits a consumable `buffer` array and `matchedCommand` object
- Listens for keystrokes in any non-input elements
- On each keystroke, if the buffer matches a pattern/sequence in `keyboard.keyCommands`, that command's `action` is run. These actions can be any function, and are commonly `router.transitionTo()` objects.
- Uses [Tether](https://www.npmjs.com/package/tether) to place hints on-screen for when the user holds `Shift`
- Shortcuts can be added by use of modifier (such as [the Go to ACLs link](https://github.com/hashicorp/nomad/pull/12831/files#diff-1ec2d76a07ab84748c05e126dc5a4acd5cc4ff74274d8e7c2232bb46274ea887R63)), or by use of component-level, conditional helper `keyboard-commands` (for example, how we [conditionally add the escape key to Evaluations modal](https://github.com/hashicorp/nomad/pull/12831/files#diff-bc34608303f7472092dae6e0956c0751561489b44098aae229b22be72fb6b94dR2-R4))
- Buffer clears after 0.75sec or when a successful keyboard command is executed

## Side-effects
- In order to facilitate a `u` shortcut to "Go up a level" from nested routes, reworked many otherwise-inline subnavs into their own components, such as the [plugin subnav](https://github.com/hashicorp/nomad/pull/12831/files#diff-ed282c6370801ee51d43fbae8b5f449ee01b3c71e31a6ac1ac3c333c1c587674R1-R6)
- In order to work around an Ember Router bug around LinkTos with query properties when deeply nested, invoked [the classic linkTo component](https://github.com/hashicorp/nomad/pull/12831/files#diff-e59da3ad52b5b03e7a6fdb11e2402d729cd46a998234a63bc409166522c8655bR1-R13) sparingly

*Showing the keyboard modal (user pressed "?"):*
image

*Showing hints active (user holding "Shift"):*
image

### DIFF
--- a/ui/app/components/breadcrumbs/default.js
+++ b/ui/app/components/breadcrumbs/default.js
@@ -1,0 +1,18 @@
+import { action } from '@ember/object';
+import Component from '@glimmer/component';
+import KeyboardShortcutModifier from 'nomad-ui/modifiers/keyboard-shortcut';
+import { inject as service } from '@ember/service';
+
+export default class BreadcrumbsTemplate extends Component {
+  @service router;
+
+  @action
+  traverseUpALevel(args) {
+    const [path, ...rest] = args;
+    this.router.transitionTo(path, ...rest);
+  }
+
+  get maybeKeyboardShortcut() {
+    return this.args.isOneCrumbUp() ? KeyboardShortcutModifier : null;
+  }
+}
--- a/ui/app/components/client-subnav.js
+++ b/ui/app/components/client-subnav.js
@@ -1,5 +1,8 @@
 import Component from '@ember/component';
 import { tagName } from '@ember-decorators/component';
+import { inject as service } from '@ember/service';
 
 @tagName('')
-export default class ClientSubnav extends Component {}
+export default class ClientSubnav extends Component {
+  @service keyboard;
+}
--- a/ui/app/services/keyboard.js
+++ b/ui/app/services/keyboard.js
@@ -1,0 +1,480 @@
+// @ts-check
+import Service from '@ember/service';
+import { inject as service } from '@ember/service';
+import { timeout, restartableTask } from 'ember-concurrency';
+import { tracked } from '@glimmer/tracking';
+import { compare } from '@ember/utils';
+import { A } from '@ember/array';
+// eslint-disable-next-line no-unused-vars
+import EmberRouter from '@ember/routing/router';
+import { schedule } from '@ember/runloop';
+import { action, set } from '@ember/object';
+import { guidFor } from '@ember/object/internals';
+import { assert } from '@ember/debug';
+// eslint-disable-next-line no-unused-vars
+import MutableArray from '@ember/array/mutable';
+import localStorageProperty from 'nomad-ui/utils/properties/local-storage';
+
+/**
+ * @typedef {Object} KeyCommand
+ * @property {string} label
+ * @property {string[]} pattern
+ * @property {any} action
+ * @property {boolean} [requireModifier]
+ * @property {boolean} [enumerated]
+ * @property {boolean} [recording]
+ * @property {boolean} [custom]
+ * @property {boolean} [exclusive]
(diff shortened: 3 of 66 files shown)


ui/app/components/breadcrumbs/default.js
```diff
@@ -0,0 +1,18 @@
...
+
+  @action
+  traverseUpALevel(args) {
+    const [path, ...rest] = args;
+    this.router.transitionTo(path, ...rest);
+  }
+
+  get maybeKeyboardShortcut() {
+    return this.args.isOneCrumbUp() ? KeyboardShortcutModifier : null;
+  }
```
ChaiWithJai, line 17: `nit` `suggestion`: This has a fixed value.

The function definition is:

```
  isOneCrumbUp(iter = 0, totalNum = 0) {
    return iter === totalNum - 2;
  }
```

No arguments are passed to this function, thus it will always be `false`.

ui/app/components/client-subnav.js
```diff
@@ -1,5 +1,8 @@
 import Component from '@ember/component';
 import { tagName } from '@ember-decorators/component';
+import { inject as service } from '@ember/service';
 
 @tagName('')
-export default class ClientSubnav extends Component {}
+export default class ClientSubnav extends Component {
+  @service keyboard;
```
ChaiWithJai, line 7: `suggestion` `non-blocking`: I think Ember supports [injecting services directly into templates](https://github.com/emberjs/rfcs/issues/543) so we can remove backing Component classes when they're not needed like here.

ui/app/services/keyboard.js
```diff
@@ -0,0 +1,480 @@
...
+            command.action();
+
+            // TODO: Temporary dev log
+            if (this.config.isDev) {
+              this.matchedCommands.forEach((command) =>
+                console.log('command run', command, command.action.toString())
+              );
+            }
+          }
+        });
```
ChaiWithJai, line 418: `suggestion`: Do you want to remove this before it goes into prod?

ui/app/services/keyboard.js
```diff
@@ -0,0 +1,480 @@
...
+    yield timeout(DEBOUNCE_MS);
+    if (recorder) {
+      this.endRebind(recorder);
+    }
+    this.clearBuffer();
+  }
+
+  get matchedCommands() {
+    // Ember Compare: returns 0 if there's no diff between arrays.
+    // TODO: do we think this is faster than a pure JS .join("") comparison?
```
ChaiWithJai, line 431: `suggestion`: Do you want to remove this before it goes into prod?

ChaiWithJai (approved):

ui/app/components/breadcrumbs/default.js
philrenaud: This is perhaps made more complicated by virtue of the curried function having the same name as the applied function, but this refers to the isOneCrumbUp defined in the component here: https://github.com/hashicorp/nomad/blob/963cc40f0807c0409f2f45ca82332b2ee1a93222/ui/app/templates/components/app-breadcrumbs.hbs#L4

Issue #11564, opened 2021-11-24
`helper` package functions to copy maps/slices encourage unsafe usage


tgross:
The various generic copy helper functions in the [`helper`](https://github.com/hashicorp/nomad/blob/v1.2.0/helper/funcs.go#L301-L403) package return `nil` if the input is zero-length. This was likely done to save memory by avoiding heap-allocating empty maps and slices (a few pointers each). But this encourages usage where `nil` maps are being passed around and need to be checked _everywhere_ after a copy, or we can hit a panic. See https://github.com/hashicorp/nomad/issues/11563#issuecomment-977929300 for an example.

The functions have a lot of callers, so one approach we might want to consider here is having the `Copy` methods in the `nomad/structs` package use new variants that allocate the empty map.

schmichael:
Fixed by #14139

PR #14139, opened 2022-08-16
client: fix data races in config handling

schmichael:
Part 4/n of my effort to make agent tests race free. (Prev: #14121)

This is the big ugly one as it tackles one of the oldest sources of data races in Nomad: client config handling.

Before this change, Client had 2 copies of the config object: config and configCopy. There was no guidance around which to use where (other than configCopy's comment to pass it to alloc runners), both are shared among goroutines and mutated in data racy ways. At least at one point I think the idea was to have `config` be mutable and then grab a lock to overwrite `configCopy`'s pointer atomically. This would have allowed alloc runners to read their config copies in data race safe ways, but this isn't how the current implementation worked.

This change takes the following approach to safely handling configs in the client:

1. `Client.config` is the only copy of the config and all access must go through the `Client.configLock` mutex
2. Since the mutex *only protects the config pointer itself and not fields inside the Config struct:* all config mutation must be done on a *copy* of the config, and then Client's config pointer is overwritten while the mutex is acquired. Alloc runners and other goroutines with the old config pointer will not see config updates.
3. Deep copying is implemented on the Config struct to satisfy the previous approach. The TLS Keyloader is an exception because it has its own internal locking to support mutating in place. An unfortunate complication but one I couldn't find a way to untangle in a timely fashion.
4. To facilitate deep copying I made an *internally backward incompatible API change:* our `helper/funcs` used to turn containers (slices and maps) with 0 elements into nils. This probably saves a few memory allocations but makes it very easy to cause panics. Since my new config handling approach uses more copying, it became very difficult to ensure all code that used containers on configs could handle nils properly. Since this code has caused panics in the past, I fixed it: nil containers are copied as nil, but 0-element containers properly return a new 0-element container. No more "downgrading to nil!"

### DIFF
--- a/client/acl.go
+++ b/client/acl.go
@@ -81,7 +81,7 @@ func (c *Client) ResolveSecretToken(secretID string) (*structs.ACLToken, error) 
 
 func (c *Client) resolveTokenAndACL(secretID string) (*acl.ACL, *structs.ACLToken, error) {
 	// Fast-path if ACLs are disabled
-	if !c.config.ACLEnabled {
+	if !c.GetConfig().ACLEnabled {
 		return nil, nil, nil
 	}
 	defer metrics.MeasureSince([]string{"client", "acl", "resolve_token"}, time.Now())
@@ -127,7 +127,7 @@ func (c *Client) resolveTokenValue(secretID string) (*structs.ACLToken, error) {
 	raw, ok := c.tokenCache.Get(secretID)
 	if ok {
 		cached := raw.(*cachedACLValue)
-		if cached.Age() <= c.config.ACLTokenTTL {
+		if cached.Age() <= c.GetConfig().ACLTokenTTL {
 			return cached.Token, nil
 		}
 	}
@@ -179,7 +179,7 @@ func (c *Client) resolvePolicies(secretID string, policies []string) ([]*structs
 
 		// Check if the cached value is valid or expired
 		cached := raw.(*cachedACLValue)
-		if cached.Age() <= c.config.ACLPolicyTTL {
+		if cached.Age() <= c.GetConfig().ACLPolicyTTL {
 			out = append(out, cached.Policy)
 		} else {
 			expired = append(expired, cached.Policy)
--- a/client/client.go
+++ b/client/client.go
@@ -734,9 +734,9 @@ func (c *Client) reloadTLSConnections(newConfig *nconfig.TLSConfig) error {
 
 	// Keep the client configuration up to date as we use configuration values to
 	// decide on what type of connections to accept
-	c.configLock.Lock()
-	c.config.TLSConfig = newConfig
-	c.configLock.Unlock()
+	c.UpdateConfig(func(c *config.Config) {
+		c.TLSConfig = newConfig
+	})
 
 	c.connPool.ReloadTLS(tlsWrap)
 
--- a/client/client.go
+++ b/client/client.go
@@ -772,21 +772,29 @@ func (c *Client) GetConfig() *config.Config {
 
 // Datacenter returns the datacenter for the given client
 func (c *Client) Datacenter() string {
+	c.configLock.RLock()
+	defer c.configLock.RUnlock()
 	return c.config.Node.Datacenter
 }
 
 // Region returns the region for the given client
 func (c *Client) Region() string {
+	c.configLock.RLock()
+	defer c.configLock.RUnlock()
 	return c.config.Region
 }
 
 // NodeID returns the node ID for the given client
 func (c *Client) NodeID() string {
+	c.configLock.RLock()
+	defer c.configLock.RUnlock()
 	return c.config.Node.ID
 }
 
 // secretNodeID returns the secret node ID for the given client
 func (c *Client) secretNodeID() string {
+	c.configLock.RLock()
+	defer c.configLock.RUnlock()
 	return c.config.Node.SecretID
 }
 
@@ -1865,9 +1873,8 @@ func (c *Client) retryRegisterNode() {
 
 // registerNode is used to register the node or update the registration
 func (c *Client) registerNode() error {
-	node := c.Node()
 	req := structs.NodeRegisterRequest{
-		Node:         node,
+		Node:         c.Node(),
 		WriteRequest: structs.WriteRequest{Region: c.Region()},
 	}
 	var resp structs.NodeUpdateResponse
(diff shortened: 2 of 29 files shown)


tgross:
> To facilitate deep copying I made an internally backward incompatible API change: our helper/funcs used to turn containers (slices and maps) with 0 elements into nils. This probably saves a few memory allocations but makes it very easy to cause panics. Since my new config handling approach uses more copying, it became very difficult to ensure all code that used containers on configs could handle nils properly. Since this code has caused panics in the past, I fixed it: nil containers are copied as nil, but 0-element containers properly return a new 0-element container. No more "downgrading to nil!"

I've had https://github.com/hashicorp/nomad/issues/11564 open for a while because of this, so that's a 👍 from me.

shoenig:
Just FYI there's already a `helpers/pointer` [package](https://github.com/hashicorp/nomad/blob/main/helper/pointer/pointer.go) for pointer helpers, where `CopyPtr` could exist as something like `pointer.Copy`

schmichael:
> Just FYI there's already a `helpers/pointer` [package](https://github.com/hashicorp/nomad/blob/main/helper/pointer/pointer.go) for pointer helpers, where `CopyPtr` could exist as something like `pointer.Copy`

TIL! That's fantastic. Moving it.

client/acl.go
```diff
@@ -81,7 +81,7 @@ func (c *Client) ResolveSecretToken(secretID string) (*structs.ACLToken, error)
 
 func (c *Client) resolveTokenAndACL(secretID string) (*acl.ACL, *structs.ACLToken, error) {
 	// Fast-path if ACLs are disabled
-	if !c.config.ACLEnabled {
+	if !c.GetConfig().ACLEnabled {
```
schmichael, line 84: This is the first pattern you'll notice: config access moved to always use a getter that properly locks.

The risk here is if it is used in one of the few places that manually manages config locking you can deadlock. Tests caught the couple times that happened to me immediately, so I don't think it's a big risk.

client/client.go
```diff
@@ -733,9 +737,9 @@ func (c *Client) reloadTLSConnections(newConfig *nconfig.TLSConfig) error {
 
 	// Keep the client configuration up to date as we use configuration values to
 	// decide on what type of connections to accept
-	c.configLock.Lock()
-	c.config.TLSConfig = newConfig
-	c.configLock.Unlock()
+	c.UpdateConfig(func(c *config.Config) {
+		c.TLSConfig = newConfig
+	})
```
schmichael, line 742: Here's another pattern you should see a lot (and I call out in the configLock comment): using the Client.UpdateConfig helper when mutating the config. It handles the locking and copying aspects of safely manipulating the config at runtime.

client/client.go
```diff
@@ -1403,13 +1425,18 @@ func (c *Client) nodeID() (id, secret string, err error) {
 
 // setupNode is used to setup the initial node
 func (c *Client) setupNode() error {
-	node := c.config.Node
+	c.configLock.Lock()
+	defer c.configLock.Unlock()
+
+	newConfig := c.config.Copy()
```
shoenig, line 1431: Are we holding the lock but also writing a copy of the config because we can't ensure nothing else is modifying the original config?

helper/funcs.go
```diff
@@ -33,6 +33,12 @@ var invalidFilenameNonASCII = regexp.MustCompile(`[[:^ascii:]/\\<>:"|?*]`)
 // invalidFilenameStrict = invalidFilename plus additional punctuation
 var invalidFilenameStrict = regexp.MustCompile(`[/\\<>:"|?*$()+=[\];#@~,&']`)
 
+type Copyable[T any] interface {
+	any
```
shoenig, line 37: Never seen this; what does requiring `any` do here?

shoenig (approved):
LGTM! Couple random thoughts but this is awesome!

client/client.go
schmichael: Yes and more! The lock is doing two things in this case:

1. Go Memory Model Data Race Protection - Ensuring `c.config.Copy()` doesn't interleave with mutations in other threads (either to the c.config pointer following or the reading of the config struct itself).
2. Application Logic Race Protection - Ensuring that the mutations to `config` in this thread don't interleave with mutations in other threads (although in `setupNode()` this should never happen... although tests do strange things and if they can't be written race free, we can't use `go test -race` to detect races in non-test-code!).

We could satisfy (1) with fine grained locking like:

```go
func (c *Client) setupNode() error {
  c.configLock.Lock()
  newConfig := c.config.Copy()
  c.configLocik.Unlock()

  // ... Mutate newConfig safely here ...

  c.configLock.Lock()
  c.config = newConfig
  c.configLock.Unlock()
  return nil
}
```

Way smaller critical area and Go's race detector would still be happy! However (2) would be broken if there was another goroutine calling `UpdateConfig(func(c *config.Config) { c.Foo = bar })` during the `// ... Mutate ...` comment's work, its changes would be overwritten by the final setting of `c.config = newConfig`!

So we end up with big critical sections to ensure both Go's memory model and our application's logic are coherent and happy.

(Which brings us to your other comment... surely there's a better way?!)

client/client.go
```diff
@@ -171,15 +171,22 @@ type AllocRunner interface {
...
+	// Client.UpdateConfig() helper. If you need more fine grained control use
+	// the following pattern:
+	//
+	// 	c.configLock.Lock()
+	// 	newConfig := c.config.Copy()
+	// 	// <mutate newConfig>
+	// 	c.config = newConfig
+	// 	c.configLock.Unlock()
+	config     *config.Config
+	configLock sync.Mutex
```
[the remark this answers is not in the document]
schmichael: Yes! I only very briefly explored this route because the changes it entails get huge very quickly.

I think there's a middle ground that's the best of both worlds though: there may be so few mutable fields on `config.Config` that we could extract only them into getters/interfaces and let `config.Config` be immutable with no locking on access! `TLSConfig` already does this around key loading, although it's unfortunately jarring because it's the _only_ example of this pattern and the surrounding code can be difficult to trace.

Without checking `Node` and `TLSConfig` might be the only two mutable fields! I'm hoping to explore this with my dynamic node metadata work.

helper/funcs.go
schmichael: Whoops! Nothing! This was my first generic interface rodeo and it took me a few tries. Fixed. :sweat_smile:

Issue #13995, opened 2022-08-03
secure vars implicit ACL policy expects invalid name


tgross:
The ACL policy parsing we added for secure variables (scheduled to ship in Nomad 1.4.0) allows workload identities to gain new privileges when an appropriately named policy is created. As noted in a conversation with @apollo13 https://github.com/hashicorp/nomad/commit/9384ba19ad7b7c2794c71d08c318b2ad5290d40b#r80061480, the name template for this policy is the rather ugly `_:$job_id/$group/$task`. 

As it turns out, this isn't just ugly it's actually an invalid ACL policy name! We use [this regex](https://github.com/hashicorp/nomad/blob/e16ec6ec7a87cdfa50a1b5a2e14e2086f76263ca/nomad/structs/structs.go#L51) to check the policy name. I'm going to have to circle back to fix this later in the week but wanted to write it down so we don't forget to fix it. A proposed naming convention: `nomadjob-$job-$group-$task`

tgross:
@schmichael I'm going to revisit this when I get back after next week but do you know if there's a hidden reason for this regex other than maybe CLI/UI ergonomics? It seems like it could be safe to allow `/` without breaking anything, but I'll need to dig in further.

schmichael:
tl;dr - Naming things is impossible. Lifting the restriction seems ok with caveats.

## Context

> if there's a hidden reason for this regex other than maybe CLI/UI ergonomics?

The digital record seems to be lacking in explanation, and my fuzzy memory isn't dredging up anything...

...however I'm guessing this is around the time we were feeling regret in Nomad and Consul about our lack of identifier validation early on. Consul obviously needs DNS-friendly identifiers in a number of places. Nomad had also just run face first into the issue of `/`s being valid in Job IDs *and* now used in APIs like [`/v1/job/:job_id/dispatch`](https://www.nomadproject.io/api-docs/jobs#dispatch-job). `nomad status <anything>/dispatch` always returns a 405 because we can't differentiate between a sub-path of `/dispatch` and a job name ending in `/dispatch`.

So it seems likely this was an attempt to start with the minimum-viable-allowable-characters as it's far easier to expand allowed characters over time than restrict them further.

## Problems with proposals

Both proposed naming schemes run afoul of the `/`-is-allowed-in-job-IDs problem. If you have two jobs:
1. Job: `fooSEPbar`, Group: `eggs`, Task: `spam`
2. Job: `foo`, Group: `barSEPeggs`, Task: `spam`

[...]

PR #14140, opened 2022-08-16
allow ACL policies to be associated with workload identity

tgross:
The original design for workload identities and ACLs allows for operators to
extend the automatic capabilities of a workload by using a specially-named
policy. This has shown to be potentially unsafe because of naming collisions, so
instead we'll allow operators to explicitly attach a policy to a workload
identity.

This changeset adds workload identity fields to ACL policy objects and threads
that all the way down to the command line. It also a new secondary index to the
ACL policy table on namespace and job so that claim resolution can efficiently
query for related policies.

---

Fixes https://github.com/hashicorp/nomad/issues/13995

cc @schmichael @apollo13 @angrycub ~~I've got this in draft while I do some end-to-end verification~~ but the overall design seems sound enough for a first look if you've got comments.

### DIFF
--- a/.changelog/14140.txt
+++ b/.changelog/14140.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+cli: `acl policy info` output format has changed to improve readability with large policy documents
+```
--- a/api/acl.go
+++ b/api/acl.go
@@ -212,11 +212,15 @@ type ACLPolicyListStub struct {
 
 // ACLPolicy is used to represent an ACL policy
 type ACLPolicy struct {
-	Name        string
-	Description string
-	Rules       string
-	CreateIndex uint64
-	ModifyIndex uint64
+	Name         string
+	Description  string
+	Rules        string
+	JobNamespace string
+	JobID        string
+	Group        string
+	Task         string
+	CreateIndex  uint64
+	ModifyIndex  uint64
 }
 
 // ACLToken represents a client token which is used to Authenticate
--- a/api/acl.go
+++ b/api/acl.go
@@ -212,15 +212,21 @@ type ACLPolicyListStub struct {
 
 // ACLPolicy is used to represent an ACL policy
 type ACLPolicy struct {
-	Name         string
-	Description  string
-	Rules        string
-	JobNamespace string
-	JobID        string
-	Group        string
-	Task         string
-	CreateIndex  uint64
-	ModifyIndex  uint64
+	Name        string
+	Description string
+	Rules       string
+	JobACL      *JobACL
+
+	CreateIndex uint64
+	ModifyIndex uint64
+}
+
+// JobACL represents an ACL policy's attachment to a job, group, or task.
+type JobACL struct {
+	Namespace string
+	JobID     string
+	Group     string
+	Task      string
 }
 
 // ACLToken represents a client token which is used to Authenticate
--- a/command/acl_policy_apply.go
+++ b/command/acl_policy_apply.go
@@ -129,13 +129,17 @@ func (c *ACLPolicyApplyCommand) Run(args []string) int {
 
 	// Construct the policy
 	ap := &api.ACLPolicy{
-		Name:         policyName,
-		Description:  description,
-		Rules:        string(rawPolicy),
-		JobNamespace: namespace,
-		JobID:        jobID,
-		Group:        group,
-		Task:         task,
+		Name:        policyName,
+		Description: description,
+		Rules:       string(rawPolicy),
+	}
+	if namespace != "" {
+		ap.JobACL = &api.JobACL{
+			Namespace: namespace,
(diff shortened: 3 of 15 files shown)


.changelog/14140.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+cli: `acl policy info` output format has changed to improve readability with large policy documents
```
tgross, line 2: This is the only bit of this that's changing already-shipped code. If we want to be fussy we can pull this out to its own PR but it's a tiny change that won't get backported so seemed safe to keep here.

tgross:
Some end-to-end testing:

```
$ WI=<token extracted from a running job=httpd>
$ NOMAD_TOKEN=$WI nomad operator api '/v1/vars'
Permission denied%

$ echo '{"Items": {"name": "Tim"}}' | nomad operator api -X PUT "/v1/var/example"
$ nomad operator api '/v1/var/example?namespace=default'
{"CreateIndex":34,"CreateTime":1660675347090671505,"Items":{"name":"Tim"},"ModifyIndex":34,"ModifyTime":1660675347090671505,"Namespace":"default","Path":"example"}%

# not allowed, as expected
$ NOMAD_TOKEN=$WI nomad operator api '/v1/var/example?namespace=default'
Permission denied%

# attach the operator policy to the job
$ nomad acl policy apply -namespace default -job httpd operator ./operator.hcl
Successfully wrote "operator" ACL policy!

$ NOMAD_TOKEN=$WI nomad operator api '/v1/var/example' | jq .
{
  "CreateIndex": 34,
  "CreateTime": 1660675347090671600,
  "Items": {
    "name": "Tim"
  },
  "ModifyIndex": 34,
  "ModifyTime": 1660675347090671600,
  "Namespace": "default",
  "Path": "example"
}

# add a job-specific secret
$ echo '{"Items": {"name": "Tim"}}' | nomad operator api -X PUT "/v1/var/nomad/jobs/httpd"

# read it as expected
$ NOMAD_TOKEN=$WI nomad operator api '/v1/var/nomad/jobs/httpd' | jq .
{
  "CreateIndex": 40,
  "CreateTime": 1660675566715335400,
  "Items": {
    "name": "Tim"
  },
  "ModifyIndex": 40,
  "ModifyTime": 1660675566715335400,
  "Namespace": "default",
  "Path": "nomad/jobs/httpd"
}
```

api/acl.go
```diff
@@ -212,11 +212,15 @@ type ACLPolicyListStub struct {
...
-	Description string
-	Rules       string
-	CreateIndex uint64
-	ModifyIndex uint64
+	Name         string
+	Description  string
+	Rules        string
+	JobNamespace string
+	JobID        string
+	Group        string
```
apollo13, line 220: Nitpicking but having `Namespace` and `ID` prefixed with `Job` and `Group` and `Task` not feels kinda weird.

nomad/structs/structs.go
```diff
@@ -11841,6 +11851,19 @@ func (a *ACLPolicy) Validate() error {
...
+		mErr.Errors = append(mErr.Errors, err)
+	}
+	if a.JobID == "" && a.Group != "" {
+		err := fmt.Errorf("job ID must be set to set group")
+		mErr.Errors = append(mErr.Errors, err)
+	}
+	if a.Group == "" && a.Task != "" {
+		err := fmt.Errorf("group must be set to set task")
+		mErr.Errors = append(mErr.Errors, err)
+	}
```
shoenig, line 11865: same if-condition ordering nit from above

shoenig (approved):
LGTM! just the usual suggestions

api/acl.go
tgross: I like that, it makes it obvious that this is a special use (and has a few words fewer memory allocations for the common case).

command/acl_policy_apply.go
```diff
@@ -89,11 +111,31 @@ func (c *ACLPolicyApplyCommand) Run(args []string) int {
...
+		return 1
+	}
+	if jobID == "" && group != "" {
+		c.Ui.Error("-job is required if -group is set")
+		return 1
+	}
+	if group == "" && task != "" {
+		c.Ui.Error("-group is required if -task is set")
+		return 1
+	}
```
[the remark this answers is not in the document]
tgross: Ok, will fix.

api/acl.go
tgross: Now I'm remembering why I didn't do this originally, which is that the state store index schema was super weird. Lemme spend a little bit to see if I can figure out how to do it. Indexing on `"JobACL.Namespace"` actually does the right thing, but only if `JobACL` is non-nil so it'll break everywhere else.

tgross: Just needed a custom indexer. Done.

PR #14141, opened 2022-08-16
cleanup: helper func for comparing slices of .Equals types

shoenig:
This PR adds a `helper.EqualsFunc` interface (types that implement `.Equals`) and a `ElementsEquals` helper function which returns true if two slices contain the same elements (any order) - as determined by the type's `.Equals` function.


schmichael (approved):
This generics thing is pretty neat.

PR #14143, opened 2022-08-16
cleanup: more cleanup of slices that are really sets

shoenig:
This PR makes use of `go-set` and `ElementsEquals` helpers to cleanup more of these set-like equality checks on slices. 

The wrapper functions are left as-is - there are tests on them that prove the changes here are safe.

### DIFF
--- a/nomad/structs/services.go
+++ b/nomad/structs/services.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/go-multierror"
+	"github.com/hashicorp/go-set"
 	"github.com/hashicorp/nomad/helper"
 	"github.com/hashicorp/nomad/helper/args"
 	"github.com/hashicorp/nomad/helper/pointer"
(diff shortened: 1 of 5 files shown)


nomad/structs/services.go
```diff
@@ -1492,61 +1477,16 @@ type ConsulUpstream struct {
...
-	}
-}
-
 // Equals returns true if the structs are recursively equal.
 func (u *ConsulUpstream) Equals(o *ConsulUpstream) bool {
 	if u == nil || o == nil {
 		return u == o
 	}
+	return *u == *o
+}
```
schmichael, line 1486: This could be replaced by `pointer.Eq`

nomad/structs/services.go
```diff
@@ -1562,21 +1502,8 @@ type ConsulExposePath struct {
...
-		for _, pathB := range pathsB {
-			if pathA == pathB {
-				continue LOOP
-			}
-		}
-		return false
-	}
-	return true
+func exposePathsEqual(a, b []ConsulExposePath) bool {
+	return set.From(a).Equal(set.From(b))
```
schmichael, line 1506: This copies both `a` and `b`. Not a big deal in this case, but we may not want to make this our default pattern. A generic verison of the loop below called `SliceSetEq` or `ElementsEq` or something might be a better pattern to use in places like this where we don't have an existing `go-set.Set` type to work with. I would imagine our "set" slices usually have so few elements that even the pathological case of iterating over all of `b` for every element of `a` is faster than creating a new map.

schmichael (approved):

schmichael (approved):

PR #14144, opened 2022-08-16
ui: fix ability checking when anon policy is missing namespace

tgross:
ACL Policies aren't required to have any `namespace` blocks, and this is
particularly common with the anonymous policy. If a user visits the web UI
without a token already in their local storage and the anonymous policy has no
`namespace` blocks, the UI will hit unhandled exceptions when rendering the
sidebar or jobs page.

Filter for the case where there's no `namespace` block.

### DIFF
--- a/ui/app/abilities/job.js
+++ b/ui/app/abilities/job.js
@@ -31,6 +31,7 @@ export default class Job extends AbstractAbility {
     // For each policy record, extract all policies of all namespaces
     const allNamespacePolicies = policies
       .toArray()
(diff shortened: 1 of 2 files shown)


philrenaud (approved):

ChaiWithJai (approved):

ChaiWithJai (approved):
LGTM. Great job! I think this is your second PR on the UI(?).

Welcome to JavaScript-land!!

I assigned you as the "assignee" to allow you to decide when you want to merge.

PR #14151, opened 2022-08-17
cleanup: replace TypeToPtr helper methods with pointer.Of

pkazmierczak:
Bumping compile time requirement to go 1.18 allows us to simplify our pointer helper methods.


tgross (approved):
LGTM. 👍 

This forces API consumers to bump to go1.18 but seeing as how go1.17 is out of support I don't think that's actually a problem? We should almost certainly backport this to all supported versions, because (a) it'll make for a lot of merge conflicts on backport if we don't and (b) we should get the minimum toolchain bump into those older supported versions as well.

pkazmierczak:
> that `checks` action does look suspicious though
> 
> ```
> ==> Linting hclog statements...
> hclogvet: Unexpected package creation during export data loading
> make: *** [GNUmakefile:162: check] Error 1
> Error: Process completed with exit code 2.
> ```

Yeah... so it's because `hclogvet` uses an outdated go tools version. They [recently fixed](https://github.com/golang/tools/commit/f4515dde7950dcca340dbca4670ff90b06e3d1d2) this issue, I guess I'm gonna have to open a PR against `go-hclog`...

I tested on my laptop with `hclogvet` built against latest go-tools and it worked fine.

shoenig:
@pkazmierczak yeah go ahead and make the PR to update `go-hclog`; I had to do something similar during the 1.18 upgrade too

PR #14165, opened 2022-08-17
Variables-specific keyboard commands

philrenaud:
Adds variables-specific keyboard nav:
- `g v` to go to variables, globally
- `v` to show/hide values on a variable page
- `j` to switch to/from JSON mode when viewing or editing a variable
- `Shift + NUM` to enter a folder/file

image

( Port of https://github.com/hashicorp/nomad/pull/12831/commits/68e5999665041e267c5440ac5c1af81847eb098d and https://github.com/hashicorp/nomad/pull/12831/commits/357b3d73417497beabb76a921a704f56b0c3177f )

### DIFF
--- a/ui/app/components/variable-paths.hbs
+++ b/ui/app/components/variable-paths.hbs
@@ -13,7 +13,12 @@
   <tbody>
     {{#each this.folders as |folder|}}
       <tr data-test-folder-row {{on "click" (fn this.handleFolderClick folder.data.absolutePath)}}>
-        <td colspan="3">
+        <td colspan="3"
+          {{keyboard-shortcut 
+            enumerated=true
+            action=(fn this.handleFolderClick folder.data.absolutePath)
+          }}
+        >
           <span>
             <FlightIcon @name="folder" />
             <LinkTo @route="variables.path" @model={{folder.data.absolutePath}} @query={{hash namespace="*"}}>
@@ -29,6 +34,10 @@
         data-test-file-row
         {{on "click" (fn this.handleFileClick file)}}
         class={{if (can "read variable" path=file.absoluteFilePath namespace=file.variable.namespace) "" "inaccessible"}}
+        {{keyboard-shortcut 
+          enumerated=true
+          action=(fn this.handleFileClick file)
+        }}
       >
         <td>
           <FlightIcon @name="file-text" />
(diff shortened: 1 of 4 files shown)


ui/app/components/variable-paths.hbs
```diff
@@ -13,7 +13,12 @@
   <tbody>
     {{#each this.folders as |folder|}}
       <tr data-test-folder-row {{on "click" (fn this.handleFolderClick folder.data.absolutePath)}}>
-        <td colspan="3">
+        <td colspan="3"
+          {{keyboard-shortcut 
+            enumerated=true
+            action=(fn this.handleFolderClick folder.data.absolutePath)
```
ChaiWithJai, line 19: `blocking`: This should accept a namespace.

ui/app/components/variable-paths.hbs
```diff
@@ -29,6 +34,10 @@
         data-test-file-row
         {{on "click" (fn this.handleFileClick file)}}
         class={{if (can "read variable" path=file.absoluteFilePath namespace=file.variable.namespace) "" "inaccessible"}}
+        {{keyboard-shortcut 
+          enumerated=true
+          action=(fn this.handleFileClick file)
```
ChaiWithJai, line 39: `blocking`: This should accept a namespace.

ChaiWithJai (approved):
Hey buddy, great job! But, I think this should will into errors with namespaces.

ui/app/components/variable-paths.hbs
philrenaud: Hi @ChaiWithJai — the `handleFileClick` action reads namespace from the variable (passed here as `file`) itself. As such , namespaces do not need to be passed in. This is the same logic we use a couple lines above this in the on "click".

ui/app/components/variable-paths.hbs
philrenaud: clicking into a folder doesn't do any variable `read`, only variable `list` (like clicking into a folder in a file directory). As such, if the folder is present at all it means the user has access to list at least one variable within it and it should remain clickable. Thus, no namespace passing.

In addition: the concept of "folders" are generated in the ui and do not themselves even *have* namespaces. They can contain many variables from many different namespaces. So nothing to pass in even if it was a good idea to do so.

Issue #14166, opened 2022-08-17
[ui] Secure Variables create form leads a `404` error message after creating a new variable.


ChaiWithJai:
Creating a new secure variable using the form in `/variables/new` template returns a `404` error message. However, the value is saved.

Replay:
https://app.replay.io/recording/secure-variables-404-error--06bb1914-f05f-4a77-b1f4-624c90e3e2d8

This is likely a result of structure the `PUT` request to create variable having the following query parameter:  `&cas=0` which is causing that to concatenate to the variable's path.

Issue #14167, opened 2022-08-17
[ui] Secure Variables are being saved with the wrong path.


ChaiWithJai:
Save a new secure variable in the UI and notice that the variable is stored with an incorrect path.

![Image](https://user-images.githubusercontent.com/41024828/185209384-cafbb12a-bcb0-46db-a98c-a38f615b162d.png)

PR #14168, opened 2022-08-17
[ui] Secure Vars: ? instead & for check-and-set param

philrenaud:
Resolves https://github.com/hashicorp/nomad/issues/14166 and https://github.com/hashicorp/nomad/issues/14167

### DIFF
--- a/ui/app/adapters/variable.js
+++ b/ui/app/adapters/variable.js
@@ -13,7 +13,7 @@ export default class VariableAdapter extends ApplicationAdapter {
     let data = this.serialize(snapshot);
     let baseUrl = this.buildURL(type.modelName, data.ID);
     const checkAndSetValue = snapshot?.attr('modifyIndex') || 0;
-    return this.ajax(`${baseUrl}&cas=${checkAndSetValue}`, 'PUT', { data });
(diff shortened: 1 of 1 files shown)


ui/app/adapters/variable.js
```diff
@@ -13,7 +13,7 @@ export default class VariableAdapter extends ApplicationAdapter {
     let data = this.serialize(snapshot);
     let baseUrl = this.buildURL(type.modelName, data.ID);
     const checkAndSetValue = snapshot?.attr('modifyIndex') || 0;
-    return this.ajax(`${baseUrl}&cas=${checkAndSetValue}`, 'PUT', { data });
+    return this.ajax(`${baseUrl}?cas=${checkAndSetValue}`, 'PUT', { data });
```
ChaiWithJai, line 16: `praise`: simple solution!

ChaiWithJai (approved):
Didn't pull this down, but it makes sense intuitively.

ui/app/adapters/variable.js
ChaiWithJai: `suggestion`: This probably should've gotten tested when it was originally implemented. Missing this logic in our Mirage set-up caused our tests to pass when they shouldn't have. I think another reason why using Playwright tests would've caught this error.

PR #14173, opened 2022-08-17
rpc: fix race in conn last used tracking

schmichael:
Part 5/n of my effort to make agent tests race free. (Prev: #14139)


shoenig (approved):
LGTM!

PR #14176, opened 2022-08-18
acl: add replication to ACL Roles from authoritative region.

jrasell:
ACL Roles along with policies and global token will be replicated
from the authoritative region to all federated regions. This
involves a new replication loop running on the federated leader.

Policies and roles may be replicated at different times, meaning
the policies and role references may not be present within the
local state upon replication upsert. In order to bypass the RPC
and state check, a new RPC request parameter has been added. This
is used by the replication process; all other callers will trigger
the ACL role policy validation check.

There is a new ACL RPC endpoint to allow the reading of a set of
ACL Roles which is required by the replication process and matches
ACL Policies and Tokens. A bug within the ACL Role listing RPC has
also been fixed which returned incorrect data during blocking
queries where a deletion had occurred.

targets: feature branch

### DIFF
--- a/nomad/acl_endpoint_test.go
+++ b/nomad/acl_endpoint_test.go
@@ -1657,7 +1657,7 @@ func TestACLEndpoint_UpsertTokens(t *testing.T) {
 				aclRole1.Policies = []*structs.ACLRolePolicyLink{{Name: policy1.Name}}
 
 				require.NoError(t, testServer.fsm.State().UpsertACLRoles(
-					structs.MsgTypeTestSetup, 20, []*structs.ACLRole{aclRole1}))
+					structs.MsgTypeTestSetup, 20, []*structs.ACLRole{aclRole1}, false))
 
 				// Create a token which references the created ACL role. This
 				// role reference is duplicated to ensure the handler
@@ -1710,7 +1710,7 @@ func TestACLEndpoint_UpsertTokens(t *testing.T) {
 				aclRole1.Policies = []*structs.ACLRolePolicyLink{{Name: policy1.Name}}
 
 				require.NoError(t, testServer.fsm.State().UpsertACLRoles(
-					structs.MsgTypeTestSetup, 20, []*structs.ACLRole{aclRole1}))
+					structs.MsgTypeTestSetup, 20, []*structs.ACLRole{aclRole1}, false))
 
 				// Create an ACL token with both ACL role and policy links.
 				tokenReq1 := &structs.ACLTokenUpsertRequest{
@@ -2019,7 +2019,7 @@ func TestACL_DeleteRolesByID(t *testing.T) {
 
 	// Create two ACL roles and put these directly into state.
 	aclRoles := []*structs.ACLRole{mock.ACLRole(), mock.ACLRole()}
-	require.NoError(t, testServer.State().UpsertACLRoles(structs.MsgTypeTestSetup, 10, aclRoles))
+	require.NoError(t, testServer.State().UpsertACLRoles(structs.MsgTypeTestSetup, 10, aclRoles, false))
 
 	// Attempt to delete an ACL role without setting an auth token. This should
 	// fail.
@@ -2094,7 +2094,7 @@ func TestACL_ListRoles(t *testing.T) {
 	aclRoles := []*structs.ACLRole{mock.ACLRole(), mock.ACLRole()}
 	aclRoles[0].ID = "prefix-" + uuid.Generate()
 	aclRoles[1].ID = "prefix-" + uuid.Generate()
-	require.NoError(t, testServer.State().UpsertACLRoles(structs.MsgTypeTestSetup, 10, aclRoles))
+	require.NoError(t, testServer.State().UpsertACLRoles(structs.MsgTypeTestSetup, 10, aclRoles, false))
 
 	// Try listing roles without a valid ACL token.
 	aclRoleReq1 := &structs.ACLRolesListRequest{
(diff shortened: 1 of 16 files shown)


nomad/acl_endpoint_test.go
```diff
@@ -2145,6 +2145,128 @@ func TestACL_ListRoles(t *testing.T) {
...
+				Region:        DefaultRegion,
+				AuthToken:     aclRootToken.SecretID,
+				MinQueryIndex: aclRoleResp4.Index,
+				MaxQueryTime:  10 * time.Second,
+			},
+		}
+		var aclRoleResp5 structs.ACLRolesListResponse
+		err = msgpackrpc.CallWithCodec(codec, structs.ACLListRolesRPCMethod, aclRoleReq5, &aclRoleResp5)
+		require.NoError(t, err)
+		resultCh <- &aclRoleResp5
```
tgross, line 2166: If we fail `require.NoError` here we won't ever send on the `resultCh` and I'm pretty sure that'll block the main test goroutine? We might want to either use `assert` here (and then make sure the main goroutine can handle a nil `resultCh`) or send the error back as well and have the main goroutine `select` over either an error channel or the `resultCh`.

(Same for `TestACL_GetRolesByID` below.)

nomad/structs/acl.go
```diff
@@ -360,6 +368,12 @@ func (a *ACLRole) Copy() *ACLRole {
 // roles.
 type ACLRolesUpsertRequest struct {
 	ACLRoles []*ACLRole
+
+	// AllowMissingPolicies skips the ACL Role policy link verification and is
+	// used by the replication process. The replication cannot ensure policies
+	// are present before ACL Roles are replicated.
+	AllowMissingPolicies bool
```
tgross, line 375: Yeah I can't see any way to avoid this because blocking queries can't be made across a multiple tables. But there's nothing in this PR that changes how queries of `ACLRoles` work. How do we handle a role with a missing policy when the role gets checked?

nomad/structs/acl.go
```diff
@@ -395,6 +409,20 @@ type ACLRolesListResponse struct {
 	QueryMeta
 }
 
+// ACLRolesByIDRequest is the request object when performing a lookup of
+// multiple roles by the ID.
+type ACLRolesByIDRequest struct {
+	ACLRoleIDS []string
```
tgross, line 415:
```suggestion
	ACLRoleIDs []string
```

tgross (approved):
LGTM @jrasell!

nomad/leader.go
```diff
@@ -1689,6 +1690,215 @@ func diffACLTokens(store *state.StateStore, minIndex uint64, remoteList []*struc
...
+// process should continue.
+//
+// Typical use:
+//
+//	  if s.replicationBackoffContinue(stopCh) {
+//		   continue
+//		 } else {
+//	    return
+//	  }
+func (s *Server) replicationBackoffContinue(stopCh chan struct{}) bool {
```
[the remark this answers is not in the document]
jrasell: Absolutely; I was surprised quite how boiler plate the ACL policies, tokens, and roles was. This was certainly down to Seth's comments about `goto` and I couldn't bare to put more in the codebase.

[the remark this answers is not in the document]
shoenig: :speak_no_evil:

nomad/leader.go
```diff
@@ -1689,6 +1690,215 @@ func diffACLTokens(store *state.StateStore, minIndex uint64, remoteList []*struc
...
+	// The remote ACL role tracking is keyed by the role ID; the value is an
+	// empty struct as we already have the full object.
+	remote := make(map[string]struct{})
+
+	// Read all the ACL role currently held within our local state. This panic
+	// will only happen as a developer making a mistake with naming the index
+	// to use.
+	iter, err := store.GetACLRoles(nil)
+	if err != nil {
+		panic("failed to iterate local ACL roles")
```
shoenig, line 1867: `panic(err)` would be more informative, but yeah none of this really matters

shoenig (approved):

nomad/leader.go
```diff
@@ -1689,6 +1690,215 @@ func diffACLTokens(store *state.StateStore, minIndex uint64, remoteList []*struc
...
+	for {
+		select {
+		case <-stopCh:
+			return
+		default:
+
+			// Rate limit how often we attempt replication. It is OK to ignore
+			// the error as the context will never be cancelled and the limit
+			// parameters are controlled internally.
+			_ = limiter.Wait(context.Background())
```
schmichael, line 1730: You could [change stopCh](https://github.com/hashicorp/nomad/blob/v1.3.3/nomad/leader.go#L67) to a context and use it here.

No need to hold this up for a refactor though.

nomad/leader.go
```diff
@@ -1689,6 +1690,215 @@ func diffACLTokens(store *state.StateStore, minIndex uint64, remoteList []*struc
...
+				// role are present. We must set allow missing to true.
+				args := structs.ACLRolesUpsertRequest{
+					ACLRoles:             fetched,
+					AllowMissingPolicies: true,
+				}
+
+				// Perform the upsert directly via Raft.
+				_, _, err := s.raftApply(structs.ACLRolesUpsertRequestType, &args)
+				if err != nil {
+					s.logger.Error("failed to update ACL roles", "error", err)
```
schmichael, line 1807: I *think* this will be the first place we'll exit if this server has lost leadership since the `select` including `stopCh` above. Since there's a lot of work being done between there and here (over a potentially slow network no less!), we may want to either check `stopCh` before attempting the raft apply, or skip the error log line if the error is `ErrLeadershipLost`? :thinking: 

Not a blocker. I just know leader elections often come under scrutiny during incidents and this seems like a likely error log to see and not really need to worry about if leadership is just being transferred normally at an unlucky time.

schmichael (approved):

nomad/structs/acl.go
jrasell: When you say checked, is this referring to a token that links to an ACL role which is being used and resolved for API authorisation? If so, this will work in the same way as an ACL token which has policy linkings where the token has been replicated but the policy has not yet. The role policy links are unpacked and the policy is read from state; if the policy is not found within state, then the permissions are not applied to the token and we continue onto the next policy in the list: https://github.com/hashicorp/nomad/blob/f-gh-13120-sso-umbrella/nomad/acl.go#L152-L166

In a case where a role has been replicated, but the single policy referenced has not, use of the token will result in a permission denied error, until the policy is also replicated. Generally this time gap should be very small as all the replication functions use blocking queries.

nomad/leader.go
```diff
@@ -1689,6 +1690,215 @@ func diffACLTokens(store *state.StateStore, minIndex uint64, remoteList []*struc
...
+	store *state.StateStore, minIndex uint64, remoteList []*structs.ACLRole) (
+	delete []string, update []string) {
+
+	// The local ACL role tracking is keyed by the role ID and the value is the
+	// hash of the role.
+	local := make(map[string][]byte)
+
+	// The remote ACL role tracking is keyed by the role ID; the value is an
+	// empty struct as we already have the full object.
+	remote := make(map[string]struct{})
```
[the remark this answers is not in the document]
jrasell: I don't believe I can use this here as the mapping is used to perform checks on which ACL roles need to be deleted from our local state: https://github.com/hashicorp/nomad/blob/f-gh-13120-acl-role-replication/nomad/leader.go#L1894-L1898

nomad/leader.go
jrasell: Agreed; this was just copied to match the other replication loops. I'll update this to be a formatted message that includes this description along with the actual error.

nomad/leader.go
jrasell: This makes a lot of sense and I think having both additions is well worth it.

nomad/leader.go
jrasell: I thought about this initially when checking the limiter implementation and how we use it. I think there are a number improvements we can make to the replication functions as a followup, this being one of them.

nomad/leader.go
jrasell: This change can be seen here: https://github.com/hashicorp/nomad/blob/f-gh-13120-acl-role-replication/nomad/leader.go#L1879-L1881

nomad/leader.go
jrasell: This change can be seen here: https://github.com/hashicorp/nomad/blob/f-gh-13120-acl-role-replication/nomad/leader.go#L1753-L1779

nomad/acl_endpoint_test.go
jrasell: This change can be seen here: https://github.com/hashicorp/nomad/blob/f-gh-13120-acl-role-replication/nomad/acl_endpoint_test.go#L2151-L2179

nomad/structs/acl.go
tgross: Aye, that all makes sense. Most of the time that'll mean briefly not getting access to something you should in a non-authoritative region, but that's expected given the replication is async. 

I wasn't able to figure out any scenario where this can cause "unexpected open" behavior expect in cases where policies were updated in separate transactions in the authoritative region. But in that case (a) the "unexpectedly open" behavior will _also_ happen on the authoritative region and it's a user error, so replication just widens the window, and (b) that scenario is also what we already have today with tokens. So :+1:

nomad/acl_endpoint_test.go
tgross: Nice!

nomad/leader.go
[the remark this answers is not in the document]
tgross: Gotcha.

PR #14181, opened 2022-08-18
cleanup: fixing warnings and refactoring of command package, part 2

shoenig:
This PR continues the cleanup of the command package, removing linter
warnings, refactoring to use helpers, making tests easier to read, etc.

### DIFF
--- a/command/agent_info_test.go
+++ b/command/agent_info_test.go
@@ -1,11 +1,11 @@
 package command
 
 import (
-	"strings"
 	"testing"
 
 	"github.com/hashicorp/nomad/ci"
 	"github.com/mitchellh/cli"
+	"github.com/shoenig/test/must"
 )
 
 func TestAgentInfoCommand_Implements(t *testing.T) {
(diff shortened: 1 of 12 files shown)


command/agent_info_test.go
```diff
@@ -16,49 +16,43 @@ func TestAgentInfoCommand_Implements(t *testing.T) {
...
+	defer stopTestAgent(srv)
 
 	ui := cli.NewMockUi()
 	cmd := &AgentInfoCommand{Meta: Meta{Ui: ui}}
 
 	code := cmd.Run([]string{"-address=" + url})
-	if code != 0 {
-		t.Fatalf("expected exit 0, got: %d", code)
-	}
+	must.Zero(t, code)
```
pkazmierczak, line 25: I won't block because of this, but don't we want these explanatory messages on failures?

pkazmierczak (approved):

command/agent_info_test.go
shoenig: not when the context is obvious - this is like the test output equivalent of 
```
i++ // increment i 
```

PR #14187, opened 2022-08-18
build: print installed go version in cricle on windows

shoenig:
This PR fixes the circle workflow step on windows where we print
the go version. Like the other commands that use Go, we must
inject the install path into PATH first.


pkazmierczak (approved):
nice 👍

PR #14188, opened 2022-08-18
events: fix race in acl event handling

schmichael:
Part 6/n of my effort to make agent tests race free. (Prev: #14173)

Doing a shallow copy seems sufficient here. Note that this is on the Event (singular) object and that sharing the envelope Events (plural) object in the publishCh appears fine. If it's not fine future `go test -race` runs will find it at least. :sweat_smile:


tgross (approved):
LGTM!

PR #14192, opened 2022-08-18
core: fix race mutating jobs in scaling api

schmichael:
Part 8/n of my effort to make agent tests race free. (Prev: #14188)

Since the state store returns a pointer to the shared job structs in
memdb we must always copy it before mutating it and applying the new
version via raft. Otherwise if the rpc fails before the mutated job is
committed to raft (either due to validation, bug, crash, or other exit
condition), the leader server will have an updated copy of the job that
other servers will not have.


jrasell (approved):
LGTM!

PR #14199, opened 2022-08-19
[ui] Allocation route services table: show task-level services

### LINKED ISSUE #14198: Show Task-level services on the alloc page
Currently, the Services table on Allocations routes only shows those services that are defined at the `group` level in a job spec.

However, we also allow defining services at the `task` level, and while our new `services` endpoints see these, our UI doesn't check for them.

philrenaud:
(resolves #14198)

Adds service fragments to allocations and union taskGroup and task services. Works with both Consul and Nomad service providers.

### Before (Nomad)
image

### After (Nomad)
image

### After (Consul)
image

### DIFF
--- a/ui/app/controllers/allocations/allocation/index.js
+++ b/ui/app/controllers/allocations/allocation/index.js
@@ -50,7 +50,7 @@ export default class IndexController extends Controller.extend(Sortable) {
   @computed('model.states.@each.task')
   get tasks() {
     return this.get('model.states').mapBy('task') || [];
-  };
+  }
 
   @computed('tasks.@each.services')
   get taskServices() {
--- a/ui/app/controllers/allocations/allocation/index.js
+++ b/ui/app/controllers/allocations/allocation/index.js
@@ -47,19 +47,22 @@ export default class IndexController extends Controller.extend(Sortable) {
     return (this.get('model.allocatedResources.ports') || []).sortBy('label');
   }
 
+  @computed('model.states.@each.task')
+  get tasks() {
+    return this.get('model.states').mapBy('task') || [];
+  };
+
+  @computed('tasks.@each.services')
+  get taskServices() {
+    return this.get('tasks')
+      .map((t) => (t.get('services') || []).toArray())
+      .flat()
+      .compact();
+  }
+
   @computed('model.taskGroup.services.@each.name')
   get groupServices() {
     return (this.get('model.taskGroup.services') || []).sortBy('name');
-  }
-
-  @computed('model.states.@each.services')
-  get taskServices() {
-    const allTaskServicesFragments =
-      this.get('model.states').mapBy('task.services').compact() || [];
-    const allTaskServices = allTaskServicesFragments
-      .map((frag) => frag.toArray())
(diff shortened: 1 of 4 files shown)


ui/app/controllers/allocations/allocation/index.js
```diff
@@ -47,10 +48,22 @@ export default class IndexController extends Controller.extend(Sortable) {
...
+  get taskServices() {
+    const allTaskServicesFragments =
+      this.get('model.states').mapBy('task.services').compact() || [];
+    const allTaskServices = allTaskServicesFragments
+      .map((frag) => frag.toArray())
+      .flat();
+    return allTaskServices.sortBy('name');
+  }
+
+  @union('taskServices', 'groupServices') services;
```
ChaiWithJai, line 65: `question`: We're going to use this entity to map over and visualize the status of our services, right? Currently, this won't re-compute the value of `services` because although we're marking these as computed properties, they're not updating in real-time.

We should fetch the data for `taskServices` and `groupServices` in the route and then mark those entities as `watchRelationship`. I don't think we'll be able to pick up registrations, updates, and de-registrations for this view right now. 

When I read the first paragraph of the Client State section of this [RFC](https://docs.google.com/document/d/1CBtnO1x6jAYcq_g3GUVEwhEDrMfV_i-kjPmxRpa5sm0/edit?usp=sharing), I think we're gonna need to consider being real-time from the very beginning of this project.

ui/app/controllers/allocations/allocation/index.js
```diff
@@ -47,10 +48,22 @@ export default class IndexController extends Controller.extend(Sortable) {
   }
 
   @computed('model.taskGroup.services.@each.name')
-  get services() {
+  get groupServices() {
     return (this.get('model.taskGroup.services') || []).sortBy('name');
```
ChaiWithJai, line 52: `suggestion:` This is an asynchronous request. `allocations.allocation` does not load the `TaskGroup` and this should trigger a fetch request. `TaskGroup` can fail as a request here and we won't notify the user.

ui/app/controllers/allocations/allocation/index.js
```diff
@@ -47,10 +48,22 @@ export default class IndexController extends Controller.extend(Sortable) {
...
   }
 
+  @computed('model.states.@each.services')
```
ChaiWithJai, line 55: `question`: Is this the correct dependent key?

ChaiWithJai (approved):
I think that Service Discovery is supposed to be a real-time feature. Giving this an approval so you're not blocked on me, however.

philrenaud:
> I think that Service Discovery is supposed to be a real-time feature. Giving this an approval so you're not blocked on me, however.

Hey @ChaiWithJai — you're correct that the Service Discovery UI feature will be realtime. However, you might call this pre-work: the existing services table, currently on the allocations page in the UI, does not show any task-level services. All this PR seeks to do is to extend that already-existing functionality to include task-level services, too.

shoenig (commented):
neat!

PR #14203, opened 2022-08-19
template: use pointer values for gid and uid

### LINKED ISSUE #14085: Windows templating failure with error "failed looking up user: managing file ownership is not supported on Windows"
### Nomad version
1.3.3

Agents running Windows Server 2022 10.0.20348 Build 20348 on AWS EC2

### Issue

Unable to render template with error 
```
Template failed: error rendering "(dynamic)" => "C:\\nomad\\data\\alloc\\239e8d2c-894a-b1a5-d59e-b7eb09819f2e\\testtask\\local\\test.txt": failed looking up user: managing file ownership is not supported on Windows
```

It seems that this happens on Windows if `user` is anything but "" as seen here https://github.com/hashicorp/consul-template/blob/583050a85eea46c3faa947b3247ff31a7ec90234/renderer/file_ownership_windows.go?q=lookupUser#L26

But I'm not explicitly setting user, so it's unclear where this value is coming from.

### Reproduction steps

Run a simple job with a single template of any value on a Windows agent.

#### Expected Result

The job runs successfully, and the template renders correctly.

### Job file (if appropriate)

```
job "testjob" {
  datacenters = ["dc1"]
  type = "batch"

  group "testgroup" {
    count = 1

    constraint {
      attribute = "${attr.kernel.name}"
      value = "windows"
    }

    task "testtask" {
      driver = "raw_exec"

      template {
        data = "Test"
        destination = "local/test.txt"
      }
    }
  }
}
```

With agent config of:
```

datacenter = "dc1"
data_dir = "C:\\nomad\\data"

client {
  enabled = true
  servers = ["${NomadUrl}:4647"]
}

plugin "raw_exec" {
  config {
    enabled = true
  }
}
```

lgfa29:
When a Nomad agent starts and loads jobs that already existed in the
cluster, the default template `uid` and `gid` was being set to `0`, since this
is the zero value for `int`. This caused these jobs to fail in
environments where it was not possible to use `0`, such as in Windows
clients.

In order to differentiate between an explicit `0` and a `template` where
these properties were not set we need to use a pointer.

Closes #14085


pkazmierczak (approved):
LGTM 👍

website/content/docs/job-specification/template.mdx
```diff
@@ -84,15 +84,15 @@ refer to the [Learn Go Template Syntax][gt_learn] Learn guide.
 - `perms` `(string: "644")` - Specifies the rendered template's permissions.
   File permissions are given as octal of the Unix file permissions `rwxrwxrwx`.
 
-- `uid` `(int: -1)` - Specifies the rendered template owner's user ID. Negative
+- `uid` `(int: nil)` - Specifies the rendered template owner's user ID. Negative
   values will use the ID of the Nomad agent user.
```
tgross, line 88: If we're changing this default to `nil` we should document what `nil` does too.

.changelog/14203.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+template: Fixed a bug where job templates would use `uid` and `gid` 0 after upgrading to Nomad 1.3.3
```
tgross, line 2: I feel like we should probably also document the Windows-specific error behavior here?

tgross (approved):
LGTM!

website/content/docs/job-specification/template.mdx
lgfa29: Ah good point

PR #14204, opened 2022-08-19
client: stats need latest allocdir

schmichael:
In #14139 this code was changed to use the original copy of the config,
but Config.AllocDir is updated in the `Client.init()` method for dev
agents.

This uses the latest version of the alloc dir (which cannot change
further at runtime without a client restart which would reinitialize
the stats collector as well).


shoenig (approved):
LGTM!

PR #14205, opened 2022-08-19
SV: Make GetSecureVariable a blocking query

angrycub:
Looks like I missed the Watch on the GetSecureVariable function in the state store causing blocking queries against variables to block until the timeout.  Added a test to validate that `GetSecureVariable` blocks as expected based on the test for `Node`

### DIFF
--- a/nomad/secure_variables_endpoint_test.go
+++ b/nomad/secure_variables_endpoint_test.go
@@ -1,10 +1,12 @@
 package nomad
 
 import (
+	"encoding/json"
 	"fmt"
 	"math/rand"
 	"strings"
 	"testing"
+	"time"
 
 	msgpackrpc "github.com/hashicorp/net-rpc-msgpackrpc"
 	"github.com/shoenig/test"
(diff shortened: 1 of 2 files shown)


nomad/secure_variables_endpoint_test.go
```diff
@@ -560,3 +562,136 @@ namespace "*" {}
...
+	delay := 100 * time.Millisecond
+	time.AfterFunc(delay, func() {
+		writeVar(t, s1, 100, "default", "aaa")
+	})
+
+	// Upsert the variable we are watching later
+	delay = 200 * time.Millisecond
+	time.AfterFunc(delay, func() {
+		writeVar(t, s1, 200, "default", "bbb")
+	})
```
tgross, line 585: I think I understand what you're trying do to here, but `time.AfterFunc` blocks until the time elapses:

> AfterFunc waits for the duration to elapse and then calls f in its own goroutine.

So we wait doing nothing and then the `writeVar` is going to race with the read request below. The best way to do this kind of test is to set up the blocking query in its own goroutine that returns either a result or an error in a channel. Then have the caller `select` on that channel and a `context.WithTimeout` (to make sure we don't block forever). In pseudocode something like:

```go
var blockingErr <-chan error
go func() {
  blockingErr <- msgpackrpc.CallWithCodec(...)
}()

ctx, cancel := context.WithTimeout(context.Background(), time.Duration(...)
defer cancel()
writeVar()

select {
  case err := <- blockingErr:
    // do stuff with the error
  case <- ctx.Done()
    t.Fatal(...)
}
```

tgross (changes requested):
Bug fix LGTM, just need to fix up the test and this will be good-to-go.

nomad/secure_variables_endpoint_test.go
angrycub: I think perhaps the description of what it's doing is bad.  I reproduced a simple version of this code [here](https://goplay.tools/snippet/DzPpq1ydSMK) and it seems to work as expected as long as you wait a sufficient amount at the end.  Since the writes are the part happening out of process and the blocking queries are happening in the flow, I would expect that this test performs as expected.

angrycub: I also used debug log lines to validate it when I borrowed it from node_endpoint_test.go.

tgross: Huh, yeah I think you're right there: https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/time/sleep.go;l=163  Easy docstring PR to make to get your name in the Go contributors list 😉

tgross (approved):

PR #14221, opened 2022-08-22
build: go.mod should require go1.19

shoenig:
Since we started using `atomic.Pointer`, we should specify the `go1.19`
requirement in our `go.mod` files.


schmichael (approved):
Ah, good catch. Thanks!

angrycub (approved):
✅

PR #14223, opened 2022-08-22
ui: task lifecycle restart all tasks

lgfa29:
This PR adds the option to restart all tasks in an allocation and fixes a CSS rule that was defined by a task `finishedAt` value. Since tasks that finished running can now be restarted, this check caused the lifecycle element to be rendered with the wrong class.

After/Before:
image

### DIFF
--- a/ui/app/adapters/allocation.js
+++ b/ui/app/adapters/allocation.js
@@ -6,12 +6,17 @@ import classic from 'ember-classic-decorator';
 export default class AllocationAdapter extends Watchable {
   stop = adapterAction('/stop');
 
-  restart(allocation, taskName) {
+  restart(allocation, taskName, allTasks) {
     const prefix = `${this.host || '/'}${this.urlPrefix()}`;
     const url = `${prefix}/client/allocation/${allocation.id}/restart`;
-    return this.ajax(url, 'PUT', {
-      data: taskName && { TaskName: taskName },
-    });
+    let data = {};
+    if (taskName) {
+      data.TaskName = taskName;
+    }
+    if (allTasks) {
+      data.AllTasks = !!allTasks;
+    }
+    return this.ajax(url, 'PUT', { data });
   }
 
   ls(model, path) {
--- a/ui/app/adapters/allocation.js
+++ b/ui/app/adapters/allocation.js
@@ -6,17 +6,18 @@ import classic from 'ember-classic-decorator';
 export default class AllocationAdapter extends Watchable {
   stop = adapterAction('/stop');
 
-  restart(allocation, taskName, allTasks) {
+  restart(allocation, taskName) {
     const prefix = `${this.host || '/'}${this.urlPrefix()}`;
     const url = `${prefix}/client/allocation/${allocation.id}/restart`;
-    let data = {};
-    if (taskName) {
-      data.TaskName = taskName;
-    }
-    if (allTasks) {
-      data.AllTasks = !!allTasks;
-    }
-    return this.ajax(url, 'PUT', { data });
+    return this.ajax(url, 'PUT', {
+      data: taskName && { TaskName: taskName },
+    });
+  }
+
+  restartAll(allocation) {
+    const prefix = `${this.host || '/'}${this.urlPrefix()}`;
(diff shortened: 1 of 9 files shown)


ui/app/adapters/allocation.js
```diff
@@ -6,12 +6,17 @@ import classic from 'ember-classic-decorator';
...
     const url = `${prefix}/client/allocation/${allocation.id}/restart`;
-    return this.ajax(url, 'PUT', {
-      data: taskName && { TaskName: taskName },
-    });
+    let data = {};
+    if (taskName) {
+      data.TaskName = taskName;
+    }
+    if (allTasks) {
+      data.AllTasks = !!allTasks;
```
ChaiWithJai, line 17: `question`: Why is this a boolean?

ui/app/controllers/allocations/allocation/index.js
```diff
@@ -97,9 +97,9 @@ export default class IndexController extends Controller.extend(Sortable) {
   })
   stopAllocation;
 
-  @task(function* () {
+  @task(function* (allTasks) {
     try {
-      yield this.model.restart();
+      yield this.model.restart('', allTasks);
```
ChaiWithJai, line 102: `question` `thought`: You're passing in an empty string as a first argument here because we're looking to restart all tasks instead of only selecting one. That strikes me as a code smell for an API that should be refactored. But, I'm betting this is a backported change and we don't want to be risky like that.

Do you have any commentary about this so I can learn more?

ChaiWithJai (approved):
LGTM. I think I understand why you decided to edit to `restart` API to have a 3rd argument instead of refactoring the `taskName` logic. However, I wonder if there was another option out there that would be more maintainable.

Great job nonetheless!

ui/app/adapters/allocation.js
```diff
@@ -6,12 +6,17 @@ import classic from 'ember-classic-decorator';
 export default class AllocationAdapter extends Watchable {
   stop = adapterAction('/stop');
 
-  restart(allocation, taskName) {
+  restart(allocation, taskName, allTasks) {
```
philrenaud, line 9: I was going to suggest using `taskName` as the param you could optionally pass `"all"` to, to prevent you having to use a third parameter here, but I'm not sure if taskName can be arbitrary / one might call their task "all" in some scenario.

Either way: this seems like something worth breaking out into its own function like `restartAll()` rather than adding into the restart function.

philrenaud (approved):

ui/app/adapters/allocation.js
lgfa29: This API expects a boolean: https://github.com/hashicorp/nomad/pull/14127/files#diff-b4fa614bde4cb3604e2b5a1b95d00e4381a06e6e1cb1a29f2c34fe44a0444d01R468

Is there a better way to make sure a param is a boolean? That's how I used to do it 😅

lgfa29:
Ahh good point on the `restart` function changes @ChaiWithJai and @philrenaud. You both made the same comment so I'm combining the discussion here 😄 

I didn't give too much thought about it, so a new param seemed like the easier option. But your comments make sense, I think I was sill with Go in mind where things are strongly typed, but in JS this is indeed handled differently. 

A new `restartAll()` function sounds good. Using a special string `"all"` seems dangerous since people could have a task called `all` and then 💥.

Another option would be to overload the second argument and then do a type check in the function, so something like:

```javascript
restart(allocation, taskNameOrAll) {
  data = {}
  if (typeof taskNameOrAll === 'string') {
    data.TaskName = taskNameOrAll
  } else {
    data.AllTasks = !!taskNameOrAll
  }
  // ...
}
```

I've seen this used in JS a few times, but I don't particularly like it 😅 

I think I will go with the `restartAll`, unless there's a better option?

ui/app/controllers/allocations/allocation/index.js
lgfa29: The API takes a body with either 

```json
{
  "Task": "my-task"
}
```

Or 

```json
{
  "AllTasks": true
}
```

If you try to set both it will result in an error. I didn't think too much about the changes to the function signatures but you're right, it doesn't look good with a random empty string 😅

PR #14224, opened 2022-08-22
ui: use task state to determine if task is active

lgfa29:
The current implementation uses the task's finishedAt field to determine
if a task is active of not, but this check is not accurate. A task in
the "pending" state will not have finishedAt value but it's also not
active.

This discrepancy results in some components, like the inline stats chart
of the task row component, to be displayed even whey they shouldn't.

### DIFF
--- a/ui/app/models/task-state.js
+++ b/ui/app/models/task-state.js
@@ -1,5 +1,5 @@
 import { computed } from '@ember/object';
-import { alias, none, and } from '@ember/object/computed';
+import { alias, and } from '@ember/object/computed';
 import Fragment from 'ember-data-model-fragments/fragment';
 import { attr } from '@ember-data/model';
 import {
@@ -19,7 +19,6 @@ export default class TaskState extends Fragment {
   @attr('date') finishedAt;
   @attr('boolean') failed;
 
-  @none('finishedAt') isActive;
   @and('isActive', 'allocation.isRunning') isRunning;
 
   @computed('task.kind')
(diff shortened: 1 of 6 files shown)


ui/app/models/task-state.js
```diff
@@ -58,6 +57,11 @@ export default class TaskState extends Fragment {
     return classMap[this.state] || 'is-dark';
   }
 
+  @computed('state')
+  get isActive() {
+    return this.state === 'running';
+  }
```
ChaiWithJai, line 63: `praise`: Very smart to map `running` to `active`. This will age well!

ui/tests/acceptance/allocation-detail-test.js
```diff
@@ -158,24 +158,47 @@ module('Acceptance | allocation detail', function (hooks) {
...
+        server.db.taskStates.update(task.id, { state: states[i] });
+      });
+
+    await Allocation.visit({ id: allocation.id });
+
+    Allocation.tasks.forEach((taskRow, i) => {
+      const task = server.db.taskStates
+        .where({ allocationId: allocation.id })
+        .sortBy('name')[i];
+      const events = server.db.taskEvents.where({ taskStateId: task.id });
```
ChaiWithJai, line 193: `praise`: excellent use of those Mirage APIs!

ChaiWithJai (approved):
Awesome, awesome job. LGTM.

philrenaud (approved):
lgtm!

lgfa29:
Seems like I broke something. I will fix the tests tomorrow before merge 👍

Issue #14213, opened 2022-08-22
job/services: Aggregate allocations by Alloc ID


philrenaud:
![image](https://user-images.githubusercontent.com/713991/185930121-8f807067-6691-4a40-9131-2d53c10e7cce.png)

Issue #14202, opened 2022-08-19
Scaffold new job/services parent route


philrenaud:
Add a new route at `ui/job/:jobID/services`

PR #14226, opened 2022-08-23
Nomad Services: job routes, model, and serializer updates

philrenaud:
Adds a jobs/:jobid/services route and child routes

image
image

### TODO:
- [ ] Acceptance tests
- [ ] Mirage fixtures

### Side-effects:
- former `service` model renamed to `service-fragment`; still in use on allocations page
- Adds a `replace` flag to watchRelationships method, that when true, will unload all instances of a type from the store. 
  - Previous instances of watchable routes (like deployments, job/clients, etc.) are additive: any new fetches of those resources can safely be added on in the UI.
  - In our case, however, we don't want to show older / invalid services. by adding an `unloadAll` when the replace flag is set, we only see those services that were returned in the latest refresh.
- Adds a relationship between jobs and services, now that jobs have a /services endpoint

---
Resolves https://github.com/hashicorp/nomad/issues/14202
Resolves https://github.com/hashicorp/nomad/issues/14213


ui/tests/unit/serializers/service-test.js
```diff
@@ -0,0 +1,23 @@
...
+    assert.ok(serializer);
+  });
+
+  test('it serializes records', function (assert) {
+    let store = this.owner.lookup('service:store');
+    let record = store.createRecord('service', {});
+
+    let serializedRecord = record.serialize();
+
+    assert.ok(serializedRecord);
```
ChaiWithJai, line 21: `question`: Is this generated via `ember-cli` or are you trying to design this like a kata?

ChaiWithJai (approved):
There's about 8 test files. That can/should be removed.

ui/tests/unit/serializers/service-test.js
philrenaud: this is auto-generated as a default ember blueprint

PR #14227, opened 2022-08-23
acl: allow tokens to lookup linked roles.

jrasell:
When listing or reading an ACL role, roles linked to the ACL token
used for authentication can be returned to the caller.

I looked into using go-set for the allowed logic within the handler,
the requirement for creating the set, then transforming it meant it
was not suitable.

targets: feature branch

### DIFF
--- a/nomad/acl_endpoint.go
+++ b/nomad/acl_endpoint.go
@@ -12,6 +12,7 @@ import (
 	metrics "github.com/armon/go-metrics"
 	log "github.com/hashicorp/go-hclog"
 	memdb "github.com/hashicorp/go-memdb"
+	"github.com/hashicorp/go-set"
 	policy "github.com/hashicorp/nomad/acl"
 	"github.com/hashicorp/nomad/helper"
 	"github.com/hashicorp/nomad/helper/uuid"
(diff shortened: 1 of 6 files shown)


nomad/acl_endpoint.go
```diff
@@ -1264,13 +1264,38 @@ func (a *ACL) ListRoles(
...
+		}
+		if token == nil {
+			return structs.ErrTokenNotFound
+		}
+
+		roles = make(map[string]struct{}, len(token.Roles))
+
+		for _, roleLink := range token.Roles {
+			roles[roleLink.ID] = struct{}{}
+		}
```
shoenig, line 1296: just tagged v0.1.3 of go-set to include FromFunc, to turn all this into one line

nomad/acl_endpoint.go
```diff
@@ -1300,9 +1325,16 @@ func (a *ACL) ListRoles(
...
-			// Iterate all the results and add these to our reply object.
+			// Iterate all the results and add these to our reply object. Check
+			// before appending to the reply that the caller is allowed to view
+			// the role.
 			for raw := iter.Next(); raw != nil; raw = iter.Next() {
-				reply.ACLRoles = append(reply.ACLRoles, raw.(*structs.ACLRole).Stub())
+
+				role := raw.(*structs.ACLRole)
+
+				if _, ok := roles[role.ID]; ok || isManagement {
```
shoenig, line 1335: with `roles` being a set this would be
```
if isManagement || roles.Contains(role.ID) {
```

shoenig (approved):
LGTM! just optional code golfing

PR #14231, opened 2022-08-23
docs: fix an anchor link in secure vars docs

tgross:
This link was missed in https://github.com/hashicorp/nomad/pull/14140


DerekStrickland (approved):
LGTM

tgross:
I checked all the links into the WI and SV docs by hand and it looks like I missed one more. Fixed that here too.

PR #14234, opened 2022-08-23
e2e: add `globalTeardown` script to not leak into other tests

ChaiWithJai:
Adding a `globalTeardown` script to call `DELETE` on all `namespaces`, `policies` and `tokens` created by the E2E test suite.

Note: We may need to add an extra step in case one of these REST API calls fails to ensure that we bubble this back up to the test runner for the E2E suite. But that should be its own PR.

### DIFF
--- a/e2e/ui/global-teardown.js
+++ b/e2e/ui/global-teardown.js
@@ -6,10 +6,8 @@ async function globalTeardown() {
       await client(`/namespace/dev`, {method: 'delete'});
       await client(`/acl/policy/operator`, {method: 'delete'});
       await client(`/acl/policy/dev`, {method: 'delete'});
-      await client(`/acl/policy/anon`, {method: 'delete'});
       await client(`/acl/token/operator`, {method: 'delete'});
       await client(`/acl/token/dev`, {method: 'delete'});
-      await client(`/acl/token/anon`, {method: 'delete'});
     } catch (e) {
       console.error('ERROR:  ', e)
     }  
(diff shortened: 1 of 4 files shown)


e2e/ui/global-teardown.js
```diff
@@ -0,0 +1,18 @@
+import { client } from './api-client.js';
+
+async function globalTeardown() {
+  try {
+      await client(`/namespace/prod`, {method: 'delete'});
+      await client(`/namespace/dev`, {method: 'delete'});
+      await client(`/acl/policy/operator`, {method: 'delete'});
+      await client(`/acl/policy/dev`, {method: 'delete'});
+      await client(`/acl/policy/anon`, {method: 'delete'});
```
tgross, line 9: Something I missed on the other PR: you've got a policy named `anon`. This won't get treated by Nomad the policy `anonymous`. That's implied by your needing to have a token with that policy; if it were really an anonymous policy you wouldn't need a token for it.

That being said, we already have an anonymous policy on the cluster from [`bootstrap-nomad.sh#L22-L25`](https://github.com/hashicorp/nomad/blob/main/e2e/terraform/scripts/bootstrap-nomad.sh#L22-L25) and we probably shouldn't touch it in these tests.

ChaiWithJai: Ahhh! Good call out.

ChaiWithJai: Removed it from these tests.

philrenaud (approved):

Issue #14235, opened 2022-08-23
nomad fails to place second allocation in basic job


shoenig:
This job reliably fails to run one of the two allocations. Using dev mode and machine has plenty of resources. The only interesting thing I can think of is it's running CentOS 9.

```
➜ uname -a
Linux thinkcentre 5.14.0-142.el9.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Aug 4 18:15:17 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

The job
```hcl
job "py" {
  datacenters = ["dc1"]

  group "web1" {
    task "py1" {
      driver = "raw_exec"
      config {
        command = "python3"
        args    = ["-m", "http.server", "8081"]
      }
      resources {
        memory = 128
	cpu = 100
      }
    }
  }

  group "web2" {
    task "py2" {
      driver = "raw_exec"
      config {
        command = "python3"
        args    = ["-m", "http.server", "8082"]
      }
      resources {
        memory = 128
	cpu = 100
      }
    }
  }  
}
```

Running the job
```
➜ nomad job run py.nomad
==> 2022-08-23T10:52:09-05:00: Monitoring evaluation "bcf3f9a8"
    2022-08-23T10:52:09-05:00: Evaluation triggered by job "py"
    2022-08-23T10:52:10-05:00: Evaluation within deployment: "2733f7be"
    2022-08-23T10:52:10-05:00: Allocation "91ddb242" created: node "d1b86c81", group "web1"
    2022-08-23T10:52:10-05:00: Allocation "adeb4be7" created: node "d1b86c81", group "web2"
    2022-08-23T10:52:10-05:00: Evaluation status changed: "pending" -> "complete"
==> 2022-08-23T10:52:10-05:00: Evaluation "bcf3f9a8" finished with status "complete"
==> 2022-08-23T10:52:10-05:00: Monitoring deployment "2733f7be"
  ⠏ Deployment "2733f7be" in progress...
    
    2022-08-23T10:52:55-05:00
    ID          = 2733f7be
    Job ID      = py
    Job Version = 0
    Status      = running
    Description = Deployment is running
    
    Deployed
    Task Group  Desired  Placed  Healthy  Unhealthy  Progress Deadline
    web1        1        1       0        0          2022-08-23T11:02:09-05:00
    web2        1        1       1        0          2022-08-23T11:02:19-05:00^C
```

One of the allocs will never become healthy 

```
Allocations
[...]

Issue #14217, opened 2022-08-22
panic from dropDirPermissions


shoenig:
While working on an unrelated e2e test, hit this panic

```
nomad: pthread_mutex_lock.c:90: ___pthread_mutex_lock: Assertion `mutex->__data.__owner == 0' failed.
SIGABRT: abort
PC=0x7f15ab71f54c m=15 sigcode=18446744073709551610
signal arrived during cgo execution

goroutine 601 [syscall]:
runtime.cgocall(0x1e0c770, 0xc000c519b0)
	runtime/cgocall.go:158 +0x5c fp=0xc000c51988 sp=0xc000c51950 pc=0x40761c
os/user._Cfunc_mygetpwnam_r(0xc000c2c140, 0xc000d12570, 0x7f1548000b60, 0x400, 0xc000464620)
	_cgo_gotypes.go:159 +0x4c fp=0xc000c519b0 sp=0xc000c51988 pc=0xaa62ec
os/user.lookupUser.func1.1({0xc000c2c140, 0xc000760f20?, 0x0?}, 0xc000500800?, 0xc000c51a58?, 0x411547?)
	os/user/cgo_lookup_unix.go:69 +0xbb fp=0xc000c51a10 sp=0xc000c519b0 pc=0xaa6fdb
os/user.lookupUser.func1()
	os/user/cgo_lookup_unix.go:69 +0x31 fp=0xc000c51a50 sp=0xc000c51a10 pc=0xaa6ef1
os/user.retryWithBuffer(0xc000760f20, 0xc000c51b40)
	os/user/cgo_lookup_unix.go:244 +0x39 fp=0xc000c51a98 sp=0xc000c51a50 pc=0xaa8619
os/user.lookupUser({0x26d7c32, 0x6})
	os/user/cgo_lookup_unix.go:60 +0x168 fp=0xc000c51b90 sp=0xc000c51a98 pc=0xaa6cc8
os/user.Lookup({0x26d7c32, 0x6})
	os/user/lookup.go:36 +0x91 fp=0xc000c51bd0 sp=0xc000c51b90 pc=0xaa5c11
github.com/hashicorp/nomad/client/allocdir.dropDirPermissions({0xc0006e0370, 0x49}, 0xc51d30?)
	github.com/hashicorp/nomad/client/allocdir/fs_unix.go:42 +0x125 fp=0xc000c51ca8 sp=0xc000c51bd0 pc=0xdff805
[...]

PR #14248, opened 2022-08-23
build: set osusergo build tag by default

shoenig:
This PR activates the osuergo build tag in GNUMakefile. This forces the os/user
package to be compiled without CGO. Doing so seems to resolve a race condition
in `getpwnam_r` that causes alloc creation to hang or panic on `user.Lookup("nobody")`.

Fixes: https://github.com/hashicorp/nomad/issues/14235
Fixes: https://github.com/hashicorp/nomad/issues/14217

### DIFF
--- a/.changelog/14248.txt
+++ b/.changelog/14248.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+client: Fixed a bug where user lookups would hang or panic
+```
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -21,7 +21,7 @@ ifndef BIN
 BIN := $(GOPATH)/bin
 endif
 
-GO_TAGS ?=
+GO_TAGS ?= osusergo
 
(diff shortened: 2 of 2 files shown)


lgfa29 (approved):
TIL. Thanks!

lgfa29:
Is this an issue in previous versions as well? How far back should we backport it? 🤔

DerekStrickland (approved):
+1 TIL

tgross (approved):
LGTM

tgross:
> Is this an issue in previous versions as well? How far back should we backport it? 

We saw it in 1.2.6 (ref https://github.com/hashicorp/nomad/issues/12342) but didn't have a repro at the time. As far as I can tell the problem is inherent to the design (concurrent lookups b/c we start tasks concurrently), not to a specific version. We should probably backport to all supported versions.

PR #14249, opened 2022-08-23
Variables CLI documentation

angrycub:
In the Nomad Secure Variables CLI, there are some new flags that I am proposing we adopt for managing input and output formatting. This PR presents these flags as they would appear in the documentation so that we can discuss these as potential updates.

Of specific note, are the `-if` and `-of` flags which are meant to manage the input and output format respectively. Invocations of the `nomad var put` command, might need to set both of these values since secure variable data can come from stdin and put has the option of returning the value once set.

Another option that seemed to have value as I was writing up the commands was `-exit-code-not-found`, which would allow users to override the default exit code of 1 in cases where the variable was not found.

[Preview Website Link](https://nomad-itc2rd315-hashicorp.vercel.app/docs/commands/var)

### DIFF
--- a/website/content/docs/commands/var/get.mdx
+++ b/website/content/docs/commands/var/get.mdx
@@ -16,7 +16,7 @@ The `var get` command is used to get the contents of an existing variable.
 nomad var get [options] <path>
 ```
 
-If ACLs are enabled, this command requires a token with the 'variables:read'
+If ACLs are enabled, this command requires a token with the `variables:read`
 capability for the target variable's namespace.
 
 ## General Options
--- a/website/content/docs/commands/var/get.mdx
+++ b/website/content/docs/commands/var/get.mdx
@@ -1,0 +1,69 @@
+---
+layout: docs
+page_title: "Command: var get"
+description: |-
+  The "var get" command retrieves the secure variable at the given path from
+  Nomad. If no variable exists at that path, an error is returned.
+---
+
+# Command: var get
+
+The `var get` command retrieves the secure variable from Nomad. If the path
+does not exist the command will return a configurable exit code.
+
+## Usage
+
+```plaintext
+nomad var get [options] <path> [<item key>]
+```
+
+When ACLs are enabled, this command requires a token with the
+`secure-variable:read` capability for the variable's namespace and path.
+
+
+## General Options
+
+@include 'general_options.mdx'
+
+## Output Options
+
+- `-field` `(string: "")`: Print only the field with the given name. Specifying
+  this option will take precedence over other formatting directives. The result
+  will not have a trailing newline making it ideal for piping to other processes.
+
+- `-of` `(enum: table|json|hcl|go-template)`: Output format to render the secure
+  variable data in. When this flag is set to `go-template`, provide the template
+  content using the `-template` option. Defaults to `table` when the output is
+  sent to standard output and `json` when the output is redirected.
+
+- `-template` `(string: "")`: Template to render output with. Valid only when
+  the output format is set to `go-template`.
+
+- `-q`: Silence all non-error output
+
+## Command Options
+
+- `-exit-code-not-found` `(int: 1)`: Exit code to set if the secure variable is
+   not found. Defaults to 1.
+
+## Examples
+
+Retrieve the variable stored at path "secret/creds":
+
+```shell-session
+$ nomad var get secret/creds
+Namespace   = default
+Path        = secret/creds
+Create Time = 2022-08-23 11:14:37.189939 -0400 EDT
+Check Index = 116
+
+Items
+passcode = my-long-passcode
+```
+
+Return only the "passcode" item from the variable stored at "secret/creds":
+
+```shell-session
+$ nomad var get -field=passcode secret/creds
+my-long-passcode
+```
--- a/website/content/docs/commands/var/get.mdx
+++ b/website/content/docs/commands/var/get.mdx
@@ -46,7 +46,7 @@ Retrieve the variable stored at path "secret/creds":
 $ nomad var get secret/creds
 Namespace   = default
 Path        = secret/creds
-Create Time = 2022-08-23 11:14:37.189939 -0400 EDT
+Create Time = 2022-08-23T11:14:37-04:00
 Check Index = 116
 
 Items
--- a/website/content/docs/commands/var/get.mdx
+++ b/website/content/docs/commands/var/get.mdx
@@ -2,24 +2,22 @@
 layout: docs
 page_title: "Command: var get"
 description: |-
-  The "var get" command retrieves the secure variable at the given path from
+  The "var get" command retrieves the variable at the given path from
   Nomad. If no variable exists at that path, an error is returned.
 ---
 
 # Command: var get
 
-The `var get` command retrieves the secure variable from Nomad. If the path
-does not exist the command will return a configurable exit code.
+The `var get` command is used to get the contents of an existing variable.
 
 ## Usage
 
 ```plaintext
-nomad var get [options] <path> [<item key>]
+nomad var get [options] <path>
 ```
 
-When ACLs are enabled, this command requires a token with the
-`secure-variable:read` capability for the variable's namespace and path.
-
+If ACLs are enabled, this command requires a token with the 'variables:read'
+capability for the target variable's namespace.
 
 ## General Options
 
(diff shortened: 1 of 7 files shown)


website/content/docs/commands/var/purge.mdx
```diff
@@ -0,0 +1,63 @@
+---
+layout: docs
+page_title: "Command: var purge"
+description: |-
+  The "var purge" command permanently removes the specified secure variable
+  from Nomad.
+---
+
+# Command: var purge
```
DerekStrickland, line 9: My one bikeshed is that `delete` feels more natural here and matches other product lines. I get that it matches `job stop purge` but that is maybe something we'd want to alias in a V2 so that we can have more RESTy nomenclature.

angrycub: We anticipate that there will be versions and soft-delete coming in a future release, so I selected `purge` for the current hard-delete behavior.

tgross:
Working in parallel with @angrycub on docs: I've updated this PR based on feedback and internal discussions, did the Great Renaming, updated the flags based on the CLI code that's been merged and some code that @angrycub has in-flight for `var list` that should be merged today, and rebased on main.

Preview link: https://nomad-penb7gqgq-hashicorp.vercel.app/docs/commands/var (plus all the subcommands)

tgross:
Current preview: https://nomad-q6v6kh1lr-hashicorp.vercel.app/docs/commands/var

website/content/docs/commands/var/init.mdx
```diff
@@ -0,0 +1,34 @@
...
+The `var init` creates an example variable specification file that can be used
+as a starting point to customize further.
+
+## Usage
+
+```plaintext
+nomad var init <filename>
+```
+
+When no filename is supplied, a default filename of "spec.nsv.hcl" or
```
DerekStrickland, line 19: Should these still be "nsv" after the name change?

website/content/docs/commands/var/init.mdx
```diff
@@ -0,0 +1,34 @@
...
+- `-out` `(enum: hcl | json)`: Format of generated variable
+  specification. Defaults to `hcl`.
+
+## Examples
+
+Create an example variable specification:
+
+```shell-session
+$ nomad var init
+Example variable specification written to spec.nsv.hcl
```
DerekStrickland, line 33: Same "nsv" question. I'll tag all the occurrences I see.

website/content/docs/commands/var/list.mdx
```diff
@@ -0,0 +1,137 @@
...
+the command. When the output format is JSON, the returned variable list is
+wrapped with additional metadata that contains the next page token. For non-JSON
+output, the next page token is printed as a message to standard error and
+standard output contains the variables from the specifically requested page.
+
+
+## Usage
+
+```plaintext
+nomad var list [options] [<prefix>]
```
DerekStrickland, line 26: Why brackets and alligators?

website/content/docs/commands/var/list.mdx
```diff
@@ -0,0 +1,137 @@
...
+        "KnownLeader": true,
+        "LastContact": 0,
+        "LastIndex": 43,
+        "NextToken": "default.nomad/jobs/variable/www/nginx",
+        "RequestTime": 875792
+    }
+}
+```
+
+As with the tabular version, provide the `QueryMeta.NextToken` value as the
```
DerekStrickland, line 136:
```suggestion
As with the tabular version, provide the
```

I don't think you need this part of the sentence, but ignore if you do.

website/content/docs/commands/var/put.mdx
```diff
@@ -0,0 +1,87 @@
...
+nomad var put [options] <path to store variable> [<variable spec file reference>] [<key>=<value>]...
+```
+
+Variable metadata and items can be supplied using a variable specification, by
+using command arguments, or by a combination of the two techniques. An entire
+variable specification can be provided to the command via standard input (stdin)
+by setting the first argument to "-" or from a file by using an @-prefixed path
+to a variable specification file. When providing variable data via stdin, you
+must provide the `-in` flag with the format of the specification, either "hcl"
+or "json"
```
DerekStrickland, line 25:
```suggestion
or "json".
```

website/content/docs/commands/var/put.mdx
```diff
@@ -0,0 +1,87 @@
...
+nomad var put [options] <variable spec file reference> [<key>=<value>]...
+nomad var put [options] <path to store variable> [<variable spec file reference>] [<key>=<value>]...
+```
+
+Variable metadata and items can be supplied using a variable specification, by
+using command arguments, or by a combination of the two techniques. An entire
+variable specification can be provided to the command via standard input (stdin)
+by setting the first argument to "-" or from a file by using an @-prefixed path
+to a variable specification file. When providing variable data via stdin, you
+must provide the `-in` flag with the format of the specification, either "hcl"
```
DerekStrickland, line 24:
```suggestion
must provide the `-in` flag with the format of the specification, which must be either "hcl"
```

website/content/docs/commands/var/put.mdx
```diff
@@ -0,0 +1,87 @@
...
+or "json"
+
+Items to be stored in the variable can be supplied using the specification, as a
+series of key-value pairs, or both. The value for a key-value pair can be a
+string, an @-prefixed file reference, or a '-' to get the value from stdin. Item
+values provided from file references or stdin are consumed as-is with no
+additional processing and do not require the input format to be specified.
+
+Values supplied as command line arguments supersede values provided in the any
```
DerekStrickland, line 33: I feel like this sentence is missing some words or pivoted in it's implementation.  "In the any" doesn't make sense I don't think.

website/content/docs/commands/var/put.mdx
```diff
@@ -0,0 +1,87 @@
...
+destination namespace.
+
+## General Options
+
+@include 'general_options.mdx'
+
+## Put Options
+
+- `-check-index` `(int: <unset>)`: If set, the variable is only acted upon if
+  the server-side version's index matches the provided value. When a variable
```
DerekStrickland, line 45:
```suggestion
  the server-side version index matches the provided value. When a variable
```

website/content/docs/commands/var/put.mdx
```diff
@@ -0,0 +1,87 @@
...
+```shell-session
+$ nomad var put secret/creds passcode=my-long-passcode
+```
+
+The data can also be consumed from a file on disk by prefixing with the "@"
+symbol. For example, you can store a variable using a specification created with
+the `nomad var init` command.
+
+```shell-session
+$ nomad var put secret/foo @spec.nsv.json
```
DerekStrickland, line 80: NSV

DerekStrickland (commented):
Left some suggestions and called out what appear to be left over artifacts not caught during the rename.

website/content/docs/commands/var/put.mdx
tgross: Oops, right. This got copy-pasted from the CLI docs so I'll need to fix it there too 😁

website/content/docs/commands/var/put.mdx
tgross: This is a possessive case: we're comparing the index that belongs to the version that belongs to the server with the index that belongs to the version to that belongs to the caller.

website/content/docs/commands/var/list.mdx
tgross: Yeah, it's telling you where to get the value that you pass into `-page-token`.

website/content/docs/commands/var/list.mdx
tgross: Optional argument with a value. A fairly standardized style (used in parsers like [docopt](http://docopt.org/)), where `[options]` is a reference to the list of options below but `[<prefix>]` is an optional argument (vs `<prefix>` for a required argument)

website/content/docs/commands/var/put.mdx
tgross: Fix the CLI here: https://github.com/hashicorp/nomad/pull/14552

website/content/docs/commands/var/init.mdx
tgross: Good catch! Fixed and fixed over here too: https://github.com/hashicorp/nomad/pull/14552

tgross:
Current preview: https://nomad-cbaf3oftd-hashicorp.vercel.app/docs/commands/var

website/content/docs/commands/var/get.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+The `var get` command is used to get the contents of an existing variable.
+
+## Usage
+
+```plaintext
+nomad var get [options] <path>
+```
+
+If ACLs are enabled, this command requires a token with the 'variables:read'
```
shoenig, line 19:
```suggestion
If ACLs are enabled, this command requires a token with the `variables:read`
```

shoenig (approved):
LGTM! just the one thing

PR #14295, opened 2022-08-24
service-health-bar

ChaiWithJai:
This PR creates a component to use when its time to visualize Service Health on the Allocation Detail Page. The component handles accepting a `RecordArray` and performs a `filterBy` method and return the `length` property of:  failing, success and pending services.

image

### DIFF
--- a/ui/app/components/service-status-bar.js
+++ b/ui/app/components/service-status-bar.js
@@ -1,0 +1,45 @@
+import { computed } from '@ember/object';
+import DistributionBar from './distribution-bar';
+import { attributeBindings } from '@ember-decorators/component';
+import classic from 'ember-classic-decorator';
+
+@classic
+@attributeBindings('data-test-service-status-bar')
+export default class ServiceStatusBar extends DistributionBar {
+  layoutName = 'components/distribution-bar';
+
+  services = null;
+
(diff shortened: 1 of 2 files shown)


ui/app/components/service-status-bar.js
```diff
@@ -0,0 +1,45 @@
+import { computed } from '@ember/object';
+import DistributionBar from './distribution-bar';
+import { attributeBindings } from '@ember-decorators/component';
+import classic from 'ember-classic-decorator';
+
+@classic
+@attributeBindings('data-test-service-status-bar')
```
philrenaud, line 7: What are the main reasons to keep this a classic component? Whether or not we do so, would the ergonomics be better for us to keep test attribute bindings in template logic, maybe using `tagname = ""` and treating the template like a more modern component?

philrenaud (approved):

ui/app/components/service-status-bar.js
ChaiWithJai: There's no benefit beyond speed. This component is derived from `DistrbutionStatusBar` and specification on the design requested to have the exact pattern as the `JobStatusBar`. I figured it would be unforgiveable if I made a brand new component instead of using what's already available.

PR #14296, opened 2022-08-24
Add command "nomad tls"

lhaig:
This PR adds the same functionality that is currently available in Consul.

It adds the following commands 
'nomad tls ca create '
'nomad tls cert create -server / -client / - cli '
'nomad tls cert -info '

It also adds the docs


command/tls_cert_create.go
```diff
@@ -0,0 +1,289 @@
+package command
+
+import (
+	"crypto/x509"
+	"fmt"
+	"io/ioutil"
```
pkazmierczak, line 6: `ioutil` is being deprecated so we're trying not to import it in new code. Take a look at the documentation of functions you call and replace them with their equivalents from the `os` package please.

command/tls_cert_create.go
```diff
@@ -0,0 +1,289 @@
...
+		c.Ui.Error("Please provide the ca")
+		return 1
+	}
+	if key == "" {
+		c.Ui.Error("Please provide the key")
+		return 1
+	}
+	if !((server && !client && !cli) ||
+		(!server && client && !cli) ||
+		(!server && !client && cli)) {
```
pkazmierczak, line 162: This could be simplified I think? 
```suggestion
	if (!server || client || cli) && (server || !client || cli) && (!server || client || !cli) {
```
Or is that simpler? I'm not sure. Just a suggestion.

pkazmierczak (commented):
Hey Lance, nice work!

I left some comments but generally looks good, I also did not notice any errors running tests against this or using it. 

My general question is, though: where does this feature come from? Was it requested by someone? Are we sure we want it implemented this way? I'd prefer someone with more seniority within the project to weigh in. 

Also, where would the target implementation live? Your PR is against main, are we targeting 1.4 for including this? Should it be backported? Would be great to get some more context.

command/tls.go
```diff
@@ -0,0 +1,62 @@
...
+	"strings"
+
+	"github.com/mitchellh/cli"
+)
+
+type TLSCommand struct {
+	Meta
+}
+
+func FileDoesNotExist(file string) bool {
```
jrasell, line 14: This function is only used within the command package, so I think we can make this private.

command/tls.go
```diff
@@ -0,0 +1,62 @@
...
+	return strings.TrimSpace(helpText)
+}
+
+func (c *TLSCommand) Synopsis() string {
+	return "Generate Self Signed TLS Certificates for Nomad"
+}
+
+func (c *TLSCommand) Name() string { return "tls" }
+
+func (c *TLSCommand) Run(args []string) int {
```
jrasell, line 60: The `args` argument is not used, so we can explicitly ignore this:
```suggestion
func (c *TLSCommand) Run(_ []string) int {
```

command/tls_ca.go
```diff
@@ -0,0 +1,51 @@
...
+	return complete.PredictNothing
+}
+
+func (c *TLSCACommand) Synopsis() string {
+	return "Helpers for managing CAs"
+}
+
+func (c *TLSCACommand) Name() string { return "tls ca" }
+
+func (c *TLSCACommand) Run(args []string) int {
```
jrasell, line 49: The `args` argument is not used, so we can explicitly ignore this:
```suggestion
func (c *TLSCACommand) Run(_ []string) int {
```

command/tls_ca.go
```diff
@@ -0,0 +1,51 @@
...
+`
+	return strings.TrimSpace(helpText)
+}
+
+func (c *TLSCACommand) AutocompleteArgs() complete.Predictor {
+	return complete.PredictNothing
+}
+
+func (c *TLSCACommand) Synopsis() string {
+	return "Helpers for managing CAs"
```
jrasell, line 44: nit: technically we only provide help creating CA certificates.

command/tls_ca_create.go
```diff
@@ -0,0 +1,151 @@
...
+			"-name-constraint":            complete.PredictAnything,
+		})
+}
+
+func (c *TLSCACreateCommand) AutocompleteArgs() complete.Predictor {
+	return complete.PredictNothing
+}
+
+func (c *TLSCACreateCommand) Synopsis() string {
+	return "Create Certificate Authority for Nomad"
```
jrasell, line 81:
```suggestion
	return "Create a Certificate Authority for Nomad"
```

command/tls_ca_create.go
```diff
@@ -0,0 +1,151 @@
...
+
+
+CA Create Options:
+  -additional-name-constraint
+    Add name constraints for the CA. Results in rejecting certificates
+    for other DNS than specified. Can be used multiple times. Only used 
+    in combination with -name-constraint.
+
+  -common-name
+    Common Name of CA. Defaults to Nomad Agent CA..
```
jrasell, line 44:
```suggestion
    Common Name of CA. Defaults to Nomad Agent CA.
```

command/tls_ca_create.go
```diff
@@ -0,0 +1,151 @@
...
+
+  -days
+    Provide number of days the CA is valid for from now on.
+    Defaults to 5 years or 1825 days..
+
+  -domain
+    Domain of nomad cluster. Only used in combination with -name-constraint. 
+    Defaults to nomad..
```
jrasell, line 52:
```suggestion
    Defaults to nomad.
```

command/tls_ca_create.go
```diff
@@ -0,0 +1,151 @@
...
+    Add name constraints for the CA. Results in rejecting certificates
+    for other DNS than specified. Can be used multiple times. Only used 
+    in combination with -name-constraint.
+
+  -common-name
+    Common Name of CA. Defaults to Nomad Agent CA..
+
+  -days
+    Provide number of days the CA is valid for from now on.
+    Defaults to 5 years or 1825 days..
```
jrasell, line 48:
```suggestion
    Defaults to 5 years or 1825 days.
```

command/tls_ca_create.go
```diff
@@ -0,0 +1,151 @@
...
+  -domain
+    Domain of nomad cluster. Only used in combination with -name-constraint. 
+    Defaults to nomad..
+
+  -name-constraint
+    Add name constraints for the CA. Results in rejecting
+    certificates for other DNS than specified. If turned on localhost and 
+    -domain will be added to the allowed DNS. If the Ui is going to be served 
+    over HTTPS its DNS has to be added with -additional-constraint. It is not
+    possible to add that after the fact! Defaults to false..
```
jrasell, line 59:
```suggestion
    possible to add that after the fact! Defaults to false.
```

command/tls_ca_create.go
```diff
@@ -0,0 +1,151 @@
...
+    Defaults to 5 years or 1825 days..
+
+  -domain
+    Domain of nomad cluster. Only used in combination with -name-constraint. 
+    Defaults to nomad..
+
+  -name-constraint
+    Add name constraints for the CA. Results in rejecting
+    certificates for other DNS than specified. If turned on localhost and 
+    -domain will be added to the allowed DNS. If the Ui is going to be served
```
jrasell, line 57:
```suggestion
    -domain will be added to the allowed DNS. If the UI is going to be served 
```

command/tls_ca_create.go
```diff
@@ -0,0 +1,151 @@
...
+
+	var (
+		days                     int
+		constraint               bool
+		domain                   string
+		commonName               string
+		additionalNameConstraint flags.StringFlag
+	)
+
+	flags := c.Meta.FlagSet(c.Name(), FlagSetClient)
```
jrasell, line 96: The flags variable collides with a package import, so we should rename this to something like `flagSet`.

command/tls_ca_create.go
```diff
@@ -0,0 +1,151 @@
...
+	if l := len(args); l < 0 || l > 1 {
+		c.Ui.Error("This command takes up to one argument")
+		c.Ui.Error(commandErrorText(c))
+		return 1
+	}
+	certFileName := fmt.Sprintf("%s-agent-ca.pem", domain)
+	pkFileName := fmt.Sprintf("%s-agent-ca-key.pem", domain)
+
+	if !(FileDoesNotExist(certFileName)) {
+		c.Ui.Error(certFileName + " already exists.")
```
jrasell, line 118: This will bring the message inline with others such as job init:
```suggestion
		c.Ui.Error(fmt.Sprintf("File '%s' already exists",certFileName ))
```

command/tls_ca_create.go
```diff
@@ -0,0 +1,151 @@
...
+		return 1
+	}
+	if !(FileDoesNotExist(pkFileName)) {
+		c.Ui.Error(pkFileName + " already exists.")
```
jrasell, line 122: This will bring the message inline with others such as job init:
```suggestion
		c.Ui.Error(fmt.Sprintf("File '%s' already exists", pkFileName))
```

command/tls_ca_create.go
```diff
@@ -0,0 +1,151 @@
...
+		return 1
+	}
+
+	constraints := []string{}
+	if constraint {
+		constraints = []string{domain, "localhost"}
+		constraints = append(constraints, additionalNameConstraint...)
+	}
```
jrasell, line 130: I believe we might be able to use [go-set](https://github.com/hashicorp/go-set) for this work.

command/tls_cert.go
```diff
@@ -0,0 +1,57 @@
...
+	return complete.PredictNothing
+}
+
+func (c *TLSCertCommand) Synopsis() string {
+	return "Helpers for managing certificates"
+}
+
+func (c *TLSCertCommand) Name() string { return "tls cert" }
+
+func (c *TLSCertCommand) Run(args []string) int {
```
jrasell, line 55:
```suggestion
func (c *TLSCertCommand) Run(_ []string) int {
```

command/tls_cert_create.go
```diff
@@ -0,0 +1,289 @@
...
+		client      bool
+		key         string
+		days        int
+		dc          string
+		domain      string
+		node        string
+		server      bool
+	)
+
+	flags := c.Meta.FlagSet(c.Name(), FlagSetClient)
```
jrasell, line 128: The flags variable collides with a package import, so we should rename this to something like `flagSet`.

command/tls_cert_create.go
```diff
@@ -0,0 +1,289 @@
...
+
+  -domain
+    Provide the domain. Matters only for -server certificates.
+  
+  -key
+    Provide path to the key. Defaults to #DOMAIN#-agent-ca-key.pem.
+
+  -node
+    When generating a server cert and this is set an additional dns name is 
+    included of the form <node>.server.<datacenter>.<domain>.
```
jrasell, line 77: I find this flag name confusing as we also use the term interchangeably to describe clients.

command/tls_cert_create_test.go
```diff
@@ -0,0 +1,296 @@
...
+import (
+	"crypto/x509"
+	"net"
+	"strings"
+	"testing"
+
+	"github.com/hashicorp/nomad/testutil"
+	"github.com/mitchellh/cli"
+	"github.com/stretchr/testify/require"
+	// caCreate "github.com/hashicorp/nomad/command/tls/ca/create"
```
jrasell, line 12:
```suggestion
```

command/tls_cert_create_test.go
```diff
@@ -0,0 +1,296 @@
...
+				require.Len(t, cert.ExtKeyUsage, 0)
+			}
+			require.False(t, cert.IsCA)
+			require.Equal(t, tc.expectDNS, cert.DNSNames)
+			require.Equal(t, tc.expectIP, cert.IPAddresses)
+		}))
+	}
+}
+
+// func expectFiles(t *testing.T, certPath, keyPath string) (*x509.Certificate, crypto.Signer) {
```
jrasell, line 248: Do we want to uncomment these? Otherwise, lets remove it.

jrasell (changes requested):
Nice one @lhaig and thanks for working on this. I have some overall comments which apply generally or in more than one place, along with some inline. I have not currently used the output certificates to run a Nomad cluster due to time constraints.

- The `New` functions are only ever used within testing. Can we remove these and instead within the test instantiate the struct like we do [here](https://github.com/hashicorp/nomad/blob/546bdb8b95a85ec7c8a5ae390a90da774f6d3d7c/command/acl_bootstrap_test.go#L29)?
- A couple of tests rely on `t.Fatal` and it would be nice if we could use `require` instead.
- The create commands take an optional argument which determines the output file prefix. Is there a reason why this couldn't be a flag instead, which seems to suit the optional nature better?
- A lot of the help and flag wording seems overly terse, however, this could be personal preference.
- I wonder if the CLI output when writing files can be more descriptive to help users; something like "CA Certificate saved to: <file_path>" rather than "Saved: <file_path>".
- The certificate create command details setting `dc` but not region, which seems at odds with our [Learn Guide](https://learn.hashicorp.com/tutorials/nomad/security-enable-tls#node-certificates)
- Each create command has a number of vars instantiated to store the flag values; it might be nice to have these as part of the command like we do for [eval delete](https://github.com/hashicorp/nomad/blob/main/command/eval_delete.go#L17-L31) and others.
- Use of `ioutil.ReadFile` is deprecated and we should use `os.Readfile` instead.

command/tls_cert_create.go
lhaig: Will do

lhaig:
> * The create commands take an optional argument which determines the output file prefix. Is there a reason why this couldn't be a flag instead, which seems to suit the optional nature better?

I have removed this as the prefix did nothing in the code.

lhaig:
> * A couple of tests rely on `t.Fatal` and it would be nice if we could use `require` instead.

I have changed the test @jrasell but now I am thinking I should change the name of the test from 
TestValidateTLSCACreateCommand_noTabs to TestValidateTLSCACreateCommand_hasNoTabs

command/tls_cert_create.go
lhaig: Do we need this granularity? I can remove it

helper/tlsutil/generate.go
```diff
@@ -229,6 +229,21 @@ func keyID(raw interface{}) ([]byte, error) {
...
 
+// ParseCert parses the x509 certificate from a PEM-encoded value.
+func ParseCert(pemValue string) (*x509.Certificate, error) {
+	// The _ result below is not an error but the remaining PEM bytes.
+	block, _ := pem.Decode([]byte(pemValue))
+	if block == nil {
+		return nil, fmt.Errorf("no PEM-encoded data found")
+	}
+
+	if block.Type != "CERTIFICATE" {
```
tgross, line 240: I was going to suggest using a constant from `encoding/pem` here but apparently they don't have any defined there? Weird.

lib/file/atomic.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	tempPath := fmt.Sprintf("%s-%s.tmp", path, uuid)
+
+	if err := os.MkdirAll(filepath.Dir(path), dirPerms); err != nil {
+		return err
+	}
+	fh, err := os.OpenFile(tempPath, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, filePerms)
+	if err != nil {
+		return err
+	}
```
tgross, line 31: The stdlib has methods for everything we're doing here that use platform-specific implementations under the hood. Let's use those so we know we'll work everywhere.

tgross (changes requested):
Hi @lhaig! This is a great start!

command/tls_cert_create.go
```diff
@@ -0,0 +1,289 @@
...
+		}
+		DNSNames = append(DNSNames, name)
+		DNSNames = append(DNSNames, "localhost")
+
+		IPAddresses = append(IPAddresses, net.ParseIP("127.0.0.1"))
+		extKeyUsage = []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth, x509.ExtKeyUsageClientAuth}
+		prefix = fmt.Sprintf("%s-server-%s", dc, domain)
+
+	} else if client {
+		name = fmt.Sprintf("client.%s.%s", dc, domain)
```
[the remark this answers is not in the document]
lhaig: Good Catch @tgross will update it accordingly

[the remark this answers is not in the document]
lhaig: I cant use the word -region as this is declared and used elsewhere so I was thinking that I create -cluster-region as the flag and cluster_region as the variable. 
Do you think this will be a good way forward?

testutil/io.go
```diff
@@ -0,0 +1,60 @@
+package testutil
+
+import (
+	"io/ioutil"
+	"os"
+	"strings"
+	"testing"
+)
+
+var noCleanup = strings.ToLower(os.Getenv("TEST_NOCLEANUP")) == "true"
```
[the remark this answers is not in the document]
lhaig: I am happy to take this out. 
It did help me to make sure the certs had the alternative names

command/tls_cert_create.go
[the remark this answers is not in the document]
tgross: The `-region` flag only gets used for the parts of the CLI that make API calls. So you can reuse it here without needing to declare a flag; you'll need to modify the help text is all.

command/tls_ca_create.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func (c *TLSCACreateCommand) Help() string {
+	helpText := `
+Usage: nomad tls ca create [options] filename-prefix
+
+This command has subcommands for interacting with Certificate Authorities.
+
+Here are some simple examples, and more detailed examples are available
+in the subcommands or the documentation.
```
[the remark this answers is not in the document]
lhaig: Correct I needed something in there for the help text

command/tls_cert_create.go
lhaig: The tests fail if I change the if statement to what you suggest. I need to understand your suggestion a bit better.

command/tls_cert_create.go
```diff
@@ -0,0 +1,289 @@
...
+	var pkFileName, certFileName string
+	max := 10000
+	for i := 0; i <= max; i++ {
+		tmpCert := fmt.Sprintf("%s-%d.pem", prefix, i)
+		tmpPk := fmt.Sprintf("%s-%d-key.pem", prefix, i)
+		if fileDoesNotExist(tmpCert) && fileDoesNotExist(tmpPk) {
+			certFileName = tmpCert
+			pkFileName = tmpPk
+			break
+		}
```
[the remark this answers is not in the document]
tgross: (Which made me realize that in a future PR we should add a `nomad tls info` that can _read_ the cert and produce the typical Nomad table format for the values we care about too!)

command/tls_cert_create.go
lhaig: I have decided to leave this as is for now.

command/tls_cert_create.go
lhaig: After thinking about this I have removed the option for this.

command/tls_cert_create.go
```diff
@@ -0,0 +1,289 @@
...
+
+  -cli
+    Generate cli certificate.
+
+  -client
+    Generate client certificate.
+
+  -days
+    Provide number of days the certificate is valid for from now on. 
+    Defaults to 1 year.
```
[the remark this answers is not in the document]
tgross: Oh I see, we're matching Consul exactly. Works for me.

command/tls_cert_create.go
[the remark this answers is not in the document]
lhaig: @tgross What would you want to see in a `nomad tls cert info` command?
I would think everything that is possible to add right?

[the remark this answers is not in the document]
lhaig: I made the changes and added the info command

lib/file/atomic.go
lhaig: Hey @tgross,
Thanks for all the great comments and support. 
I am not sure what I have missed here as the os package is part of the standard library 
https://pkg.go.dev/os@go1.19.1#pkg-index
It would be great to know where to start thanks

tgross: Sorry, I should have called out specifically what I meant: [`os.CreateTemp`](https://pkg.go.dev/os#CreateTemp). And you can do a `defer os.RemoveAll` on the resulting file path to ensure it's always removed on any return, if it still exists.

lhaig: I have been looking at this and os.CreateTemp does not allow you to write the file with permissions already set. We need specific permissions on the key file as it is a sensitive file.
I have removed the duplicated os.Remove(tempPath)

website/content/docs/commands/tls/index.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+  # ...
+
+Subcommands:
+  ca      Helpers for CAs
+  cert    Helpers for certificates
+```
+
+For more information, examples, and usage about a subcommand, click on the name
+of the subcommand in the sidebar or one of the links below:
```
tgross, line 52: We're missing the links here

website/content/docs/commands/tls/index.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+```shell-session
+$ nomad tls cert create -client
+==> Using nomad-agent-ca.pem and nomad-agent-ca-key.pem
+==> Saved dc1-client-nomad.pem
+==> Saved dc1-client-nomad-key.pem
+```
+
+For more examples, ask for subcommand help or view the subcommand documentation
+by clicking on one of the links in the sidebar.
```
tgross, line 32: This mostly repeats the text below:

```suggestion
```

website/content/docs/commands/tls/index.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+## Usage
+
+Usage: `nomad tls <subcommand> <subcommand> [options]`
+
+For the exact documentation for your Nomad version, run `nomad tls -h` to
+view the complete list of subcommands.
```
tgross, line 39: We have versioned documented, so the version you're looking at here already should be the exact documentation.

```suggestion
```

website/content/docs/commands/tls/index.mdx
```diff
@@ -0,0 +1,52 @@
+---
+layout: docs
+page_title: 'Commands: TLS'
+---
```
tgross, line 4: This page should just be an overview with links to subcommands. See https://www.nomadproject.io/docs/commands/alloc for an example. We don't need pointers to the sidebar or the `--help` text or examples. All of that belongs in the individual subcommand docs.

website/content/docs/commands/tls/cert.mdx
```diff
@@ -0,0 +1,84 @@
+---
+layout: docs
+page_title: 'Commands: TLS Cert Create'
+---
+
+# Nomad TLS Cert Create
+
+Command: `nomad tls cert create`
```
tgross, line 8: There's a format that all the command docs share, so copy those:

```suggestion
# Command: nomad tls cert create
```

(Applies to all the docs)

website/content/docs/commands/tls/cert.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+The `tls cert create` command is used to create certificates to be used in 
+for your Nomad cluster for TLS setup. You can then copy these to your server
+as part fo the configurstion. This command will not automatically update the 
+settings in the agents.
```
tgross, line 13: Reworded for clarity/spelling. Also, because this is a web page and not the CLI help text, we can link out to the guide:

```suggestion
The `tls cert create` command is used to create certificates to be used for
[TLS encryption][] for your Nomad cluster. You can then copy these to your 
servers and clients. This command will not automatically update the 
configuration of the agents.
```

And then add a footnote at the bottom to make it a clickable link:

```
[TLS encryption]: https://learn.hashicorp.com/tutorials/nomad/security-enable-tls?in=nomad/transport-security
```

website/content/docs/commands/tls/ca.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+- `-additional-name-constraint=<value>` - Add name constraints for the CA.
+  Results in rejecting certificates for other DNS than specified. Can be used
+  multiple times. Only used in combination with `-name-constraint`.
+
+- `common-name` - Common Name of CA. Defaults to Nomad Agent CA.
+
+- `-days=<int>` - Provide number of days the CA is valid for from now on, defaults to 5 years.
+
+- `-domain=<string>` - Domain of nomad cluster. Only used in combination with
```
tgross, line 36: Nomad is always capitalized unless it's a command/binary/etc:

```suggestion
- `-domain=<string>` - Domain of Nomad cluster. Only used in combination with
```

website/content/docs/commands/tls/ca.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+## Usage
+
+Usage: `nomad tls ca create [filename-prefix] [options]`
+
+#### Command Options
+
+- `-additional-name-constraint=<value>` - Add name constraints for the CA.
+  Results in rejecting certificates for other DNS than specified. Can be used
+  multiple times. Only used in combination with `-name-constraint`.
```
tgross, line 30: Let's turn these into full sentences (and similar for the `-name-constraint` option):

```suggestion
- `-additional-name-constraint=<value>` - Add name constraints for the CA.
  The server will reject certificates for DNS names other than those specified. Can
  be used multiple times. This option can only used in combination with `-name-constraint`.
```

website/content/docs/commands/tls/ca.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+- `common-name` - Common Name of CA. Defaults to Nomad Agent CA.
```
tgross, line 32:
```suggestion
- `common-name` - Common Name of CA. Defaults to `Nomad Agent CA`.
```

command/tls.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	return false
+}
+
+func (c *TLSCommand) Help() string {
+	helpText := `
+Usage: nomad tls <subcommand> <subcommand> [options]
+
+This command groups subcommands for creating certificates for Nomad TLS configuration. 
+The TLS command allows operators to generate self signed certificates to use
```
tgross, line 26: Just a quibble around wording here. These aren't "self-signed certificates" in the traditional meaning of that term. They're signed by a CA, it's just that we're _also_ creating the CA.

command/tls.go
```diff
@@ -0,0 +1,58 @@
...
+
+Create a server certificate
+
+    $ nomad tls cert create -server
+
+Create a client certificate
+
+    $ nomad tls cert create -client
+
+For more examples, ask for subcommand help or view the documentation.
```
tgross, line 44: This repeats the sentence above the examples. Let's pick one or the other.

command/tls_ca.go
```diff
@@ -0,0 +1,53 @@
...
+  ==> CA Certificate saved to: nomad-agent-ca.pem
+  ==> CA Certificate key saved to: nomad-agent-ca-key.pem
+  $ nomad tls ca info nomad-agent-ca.pem
+  nomad-agent-ca.pem
+  Issuer CN              Nomad Agent CA 58896012363767591697986789371079092261
+  Common Name            CN=Nomad Agent CA 58896012363767591697986789371079092261,O=HashiCorp Inc.,...
+  Expiry Date            2027-09-24 22:24:08 +0000 UTC
+  Permitted DNS Domains  []
+
+For more examples, ask for subcommand help or view the documentation.
```
tgross, line 35: Repeated above.

command/tls_ca.go
```diff
@@ -0,0 +1,53 @@
...
+
+type TLSCACommand struct {
+	Meta
+}
+
+func (c *TLSCACommand) Help() string {
+	helpText := `
+Usage: nomad tls ca <subcommand> [options]
+
+This command has subcommands for interacting with Certificate Authorities.
```
tgross, line 18:
```suggestion
This command has subcommands for interacting with a local Certificate Authority.
```

command/tls_ca_create.go
```diff
@@ -0,0 +1,162 @@
...
+  -domain
+    Domain of nomad cluster. Only used in combination with -name-constraint. 
+    Defaults to nomad.
+
+  -name-constraint
+    Add name constraints for the CA. Results in rejecting
+    certificates for other DNS than specified. If turned on localhost and 
+    -domain will be added to the allowed DNS. If the UI is going to be served 
+    over HTTPS its DNS has to be added with -additional-constraint. It is not
+    possible to add that after the fact! Defaults to false.
```
tgross, line 78: I left some comments on the web documentation about the wording that applies here as well, but we've also got some language here referring to "DNS" when we really mean... "domains" I think?

command/tls_ca_create.go
```diff
@@ -0,0 +1,162 @@
...
+	if l := len(args); l < 0 || l > 1 {
+		c.Ui.Error("This command takes up to one argument")
+		c.Ui.Error(commandErrorText(c))
+		return 1
+	}
+	certFileName := fmt.Sprintf("%s-agent-ca.pem", c.domain)
+	pkFileName := fmt.Sprintf("%s-agent-ca-key.pem", c.domain)
+
+	if !(fileDoesNotExist(certFileName)) {
+		c.Ui.Error(fmt.Sprintf("CA Certificate File '%s' already exists", certFileName))
```
tgross, line 129:
```suggestion
		c.Ui.Error(fmt.Sprintf("CA certificate file %q already exists", certFileName))
```

command/tls_ca_create.go
```diff
@@ -0,0 +1,162 @@
...
+		return 1
+	}
+	if !(fileDoesNotExist(pkFileName)) {
+		c.Ui.Error(fmt.Sprintf("CA Key File '%s' already exists", pkFileName))
```
tgross, line 133:
```suggestion
		c.Ui.Error(fmt.Sprintf("CA key file %q already exists", pkFileName))
```

command/tls_ca_create.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		c.Ui.Error(err.Error())
+		return 1
+	}
+
+	if err := file.WriteAtomicWithPerms(certFileName, []byte(ca), 0755, 0666); err != nil {
+		c.Ui.Error(err.Error())
+		return 1
+	}
+	c.Ui.Output("==> CA Certificate saved to: " + certFileName)
```
tgross, line 153:
```suggestion
	c.Ui.Output("==> CA certificate saved to: " + certFileName)
```

command/tls_ca_create.go
```diff
@@ -0,0 +1,162 @@
...
+
+	if err := file.WriteAtomicWithPerms(pkFileName, []byte(pk), 0755, 0600); err != nil {
+		c.Ui.Error(err.Error())
+		return 1
+	}
+	c.Ui.Output("==> CA Certificate key saved to: " + pkFileName)
```
tgross, line 159:
```suggestion
	c.Ui.Output("==> CA certificate key saved to: " + pkFileName)
```

command/tls_ca_create.go
```diff
@@ -0,0 +1,162 @@
...
+
+
+CA Create Options:
+  -additional-name-constraint
+    Add name constraints for the CA. Results in rejecting certificates
+    for other DNS than specified. Can be used multiple times. Only used 
+    in combination with -name-constraint.
+
+  -common-name
+    Common Name of CA. Defaults to Nomad Agent CA..
```
tgross, line 63:
```suggestion
    Common Name of CA. Defaults to "Nomad Agent CA".
```

command/tls_ca_create.go
```diff
@@ -0,0 +1,162 @@
...
+Here are some simple examples, more details can be found below or in 
+the documentation.
+
+Create a new Nomad CA
+
+  $ nomad tls ca create
+  ==> CA Certificate saved to: nomad-agent-ca.pem
+  ==> CA Certificate key saved to: nomad-agent-ca-key.pem
+
+  Create a new Nomad CA with a Custiom Domain
```
tgross, line 49:
```suggestion
  Create a new Nomad Certificate Authority
```

But this is also repeated above

testutil/io.go
```diff
@@ -0,0 +1,49 @@
...
+	name = strings.ReplaceAll(name, "/", "_")
+	d, err := os.MkdirTemp("", name)
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
+	t.Cleanup(func() {
+		os.RemoveAll(d)
+	})
+	return d
+}
```
tgross, line 27: What's wrong with [`t.Tempdir`](https://pkg.go.dev/testing#T.TempDir) exactly?

testutil/tls.go
```diff
@@ -0,0 +1,68 @@
...
+//
+//	defer switchToTempDir(t, testDir)()
+//
+// This exploits the fact that the body of a defer is evaluated
+// EXCEPT for the final function call invocation inline with the code
+// where it is found. Only the final evaluation happens in the defer
+// at a later time. In this case it means we switch to the temp
+// directory immediately and defer switching back in one line of test
+// code.
+func SwitchToTempDir(t *testing.T, testDir string) func() {
```
tgross, line 61: I think?

```suggestion
func SwitchToTempDir(t *testing.T, testDir string) func() {
	t.Helper()
```

lib/file/atomic_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+	// Write
+	expected := []byte("hello")
+	require.NoError(t, WriteAtomicWithPerms(path, expected, 0700, 0600))
+
+	// Read and verify
+	actual, err := os.ReadFile(path)
+	require.NoError(t, err)
+	require.Equal(t, expected, actual)
+}
```
tgross, line 30: Let's kill this file. It doesn't even claim to test what it says 😀

testutil/io.go
```diff
@@ -0,0 +1,49 @@
...
+	name = strings.ReplaceAll(name, "/", "_")
+	f, err := os.CreateTemp("", name)
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
+	t.Cleanup(func() {
+		os.Remove(f.Name())
+	})
+	return f
+}
```
tgross, line 49: I can't find the caller for this... but usually if you're writing files in a test it's best just to call `t.Tempdir` and then write into there.

command/tls_cert_info.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	var certFile []byte
+	var err error
+	var file string
+
+	if len(args) == 1 {
+		file = args[0]
+		certFile, err = os.ReadFile(file)
+		if err != nil {
+			c.Ui.Error(fmt.Sprintf("Error reading Certifiate file: %v", err))
```
tgross, line 74:
```suggestion
			c.Ui.Error(fmt.Sprintf("Error reading certificate file: %v", err))
```

command/tls_cert_create.go
```diff
@@ -0,0 +1,298 @@
...
+		return 1
+	}
+	caKey, err := os.ReadFile(keyFile)
+	if err != nil {
+		c.Ui.Error(fmt.Sprintf("Error reading CA key: %s", err))
+		return 1
+	}
+
+	if c.server {
+		c.Ui.Info(
```
tgross, line 244: There's actually a `c.Ui.Warning` that does colored text we might want to use here.

tgross (changes requested):
Hi @lhaig I've taken another pass over this but sort of ran out of steam. The code looks to be getting there but now the textual bits need a good solid scrub with spell checker and some rewording for clarity and grammar.

(Don't forget to add a changelog entry.)

command/tls_ca_create_test.go
```diff
@@ -0,0 +1,96 @@
...
+			"nomad-agent-ca.pem",
+			"nomad-agent-ca-key.pem",
+			func(t *testing.T, cert *x509.Certificate) {
+				require.True(t, strings.HasPrefix(cert.Subject.CommonName, "Nomad Agent CA"))
+			},
+		},
+	}
+	for _, tc := range cases {
+		tc := tc
+		require.True(t, t.Run(tc.name, func(t *testing.T) {
```
tgross, line 76: You don't need the `require.True` here. Just call `t.Run` and it'll do the right thing.

command/tls_cert_create.go
```diff
@@ -0,0 +1,298 @@
...
+    server and access all state in the cluster including root keys
+    and all ACL tokens. Do not distribute them to production hosts
+    that are not server nodes. Store them as securely as CA keys.
+==> Using nomad-agent-ca.pem and nomad-agent-ca-key.pem
+==> Server Certificate saved to: global-server-nomad.pem
+==> Server Certificate key saved to: global-server-nomad-key.pem
+$ nomad tls cert create -client
+==> Using CA file nomad-agent-ca.pem and CA key nomad-agent-ca-key.pem
+==> Client Certificate saved to global-client-nomad.pem
+==> Client Certificate key saved to global-client-nomad-key.pem
```
tgross, line 64: Having examples in the docstring is going to look confusing, especially with the big WARNING flags.

website/content/docs/commands/tls/cert-create.mdx
```diff
@@ -0,0 +1,86 @@
...
+---
+
+# Command: nomad tls cert create
+
+The `tls cert create` command is used to create certificates to be used for
+[TLS encryption][] for your Nomad cluster. You can then copy these to your 
+servers and clients. This command will not automatically update the 
+configuration of the agents.
+
+## Examples
```
tgross, line 15: Examples always go last in our docs.

testutil/tls.go
```diff
@@ -0,0 +1,70 @@
...
+	t.Helper()
+
+	previousDirectory, err := os.Getwd()
+	require.NoError(t, err)
+	require.NoError(t, os.Chdir(testDir))
+	return func() {
+		os.Chdir(previousDirectory)
+	}
+}
```
tgross, line 70: This feels excessively clever for something that's a test helper only used in two tests. Can't the `cmd.Run` calls output their files directly to the tempdir anyways?

lhaig: I am happy to remove the file ad add the code directly to the two tests.

tgross:
Preview link https://nomad-nttruonxz-hashicorp.vercel.app/nomad/docs/commands/tls

tgross (approved):
LGTM. @jrasell has open comments that I want to make sure he's satisfied about, but we'll merge once he's had a second chance at it on Monday.

command/tls_ca_create.go
lhaig: This is used to populate the list of additional domains that the CA will permit certificates to be created for. 
All it needs to do is add any additional domains it finds from the flags provided and then pass them to the tlsutil.GenerateCA function. I am not sure how I could do that with go-set.

jrasell (approved):
LGTM; thanks @lhaig! I am very excited about using this for automagic Nomad TLS clusters.

tgross:
Let's get this merged so it can go out in the next release of Nomad (likely 1.4.4).

Issue #14079, opened 2022-08-10
Unable to start tasks "logmon: fork/exec /usr/bin/nomad;62edd7f7: no such file or directory"


ktf:
### Nomad version

Nomad v1.3.3 (428b2cd8014c48ee9eae23f02712b7219da16d30)

### Operating system and Environment details

Centos 7

### Issue

Today a given machine in our cluster was failing to run some services and the only message which was there is:

"logmon: fork/exec /usr/bin/nomad;62edd7f7: no such file or directory"

restarting the machine did not help. Then suddenly after I pushed a raw_exec batch job for test to the same machine, things got back into business. Any idea of what might have been going on?

### Reproduction steps

No idea.

#### Expected Result

Having tasks running on that machine, as usual.

#### Actual Result

Tasks were not running.

shoenig:
Hi @ktf, can't say I've seen anything like this.

Were the failures coming from tasks being launched for the first time? Or were these tasks that Nomad was trying to reattach to? What task driver were these tasks using?

ktf:
No, the tasks have been there since a while and they have a constrain of the given machine. They are stateless long running services. They are using the exec driver. Could it be that some internal nomad state was not fully committed and that the scheduling of the raw_exec batch job cleaned up something? Does the error message mean that the executable /usr/bin/nomad is not found?

shoenig:
@ktf, was the `nomad` executable itself modified / replaced / upgraded while the tasks were running? The error message suggests `os.Executable` returned the literal string: `/usr/bin/nomad;62edd7f7` (i.e. including the odd `;62edd7f7` suffix). I'm not aware of anything in Nomad code that would do that :thinking: 

https://github.com/hashicorp/nomad/blob/v1.3.3/client/logmon/plugin.go#L19

ktf:
Yes, I suspect this correlates with the move from 1.3.2 to 1.3.3, which was done via yum update in the background. I will pin the versions for the future, but right now I am still unable  to launch jobs on many of the machines (but not all).

ktf:
for the record... I also tried adding a symlink to nomad;62edd7f7 and it still does not work.

ktf:
also, the hash is different on every machine  which does not work.

shoenig:
Can we get the exact kernel you're using via `uname -a` ?

ktf:
[...]

PR #14297, opened 2022-08-24
client/logmon: acquire executable in init block

shoenig:
This PR causes the logmon task runner to acquire the path of binary of the
Nomad executable file in an `init` block, so as to almost certainly acquire
the name while the nomad file still exists.

This is an attempt at fixing the case where a deleted Nomad file
(e.g. during upgrade) may be getting renamed with a suffix before the deletion
(maybe SELinux specific?).

If this doesn't work, as a last resort we can literally just trim the mystery string.

Fixes: #14079

### DIFF
--- a/.changelog/14297.txt
+++ b/.changelog/14297.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+client/logmon: fixed a bug where logmon cannot find nomad executable
+```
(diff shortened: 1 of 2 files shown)


tgross (approved):
LGTM

Issue #14233, opened 2022-08-23
Vault health check fails with "unsupported path"


t-davies:
### Nomad version
`Nomad v1.3.2 (bf602974112964e9691729f3f0716ff2bcdb3b44)`

### Issue
Nomad is unable to call Vault's health check endpoint with an expired Vault token, this should be possible since this Vault endpoint is not namespaced and does not require a token. The Vault client making the call should _not_ be attempting to send credentials.

There appears to have been some weird behaviour where Nomad didn't correctly renew the Vault token, not sure if that's a bug or down to configuration on our end? See the logs below. Even if this is the case, Nomad should still be able to call the health check endpoint with an expired token - it believing that Vault is not healthy causes further problems.

Previously reported something similar, although we had some other errors at the time that made it a bit ambiguous as to what was the cause, see https://github.com/hashicorp/nomad/issues/13710.

#### Expected Result
 - Nomad should renew the Vault token successfully.
 - Nomad should be able to call Vault's health check endpoint regardless of whether the Vault token has expired.

#### Actual Result
 - Nomad fails to renew its Vault token.
 - Nomad fails to call the Vault health check endpoint after the token expires.

### Nomad Server logs (if appropriate)

```
2022-08-20T20:52:19.216Z [INFO]  nomad.vault: successfully renewed token: next_renewal=37m43.499968891s
2022-08-20T21:07:25.877Z [ERROR] nomad.event_broker: failed resolving ACL for secretID, closing subscriptions: error="no token for secret ID"
2022-08-20T21:29:54.291Z [ERROR] nomad.event_broker: failed resolving ACL for secretID, closing subscriptions: error="no token for secret ID"
2022-08-20T21:30:03.412Z [INFO]  nomad.vault: successfully renewed token: next_renewal=18m51.499975211s
2022-08-20T21:35:37.207Z [ERROR] nomad.event_broker: failed resolving ACL for secretID, closing subscriptions: error="no token for secret ID"
2022-08-20T21:48:55.502Z [INFO]  nomad.vault: successfully renewed token: next_renewal=9m25.499982629s
2022-08-20T21:49:38.045Z [ERROR] nomad.event_broker: failed resolving ACL for secretID, closing subscriptions: error="no token for secret ID"
2022-08-20T21:58:21.825Z [INFO]  nomad.vault: successfully renewed token: next_renewal=4m42.499986896s
[...]

PR #14298, opened 2022-08-24
vault: detect namespace change in config reload

tgross:
The `namespace` field was not included in the equality check between old and new
Vault configurations, which meant that a Vault config change that only changed
the namespace would not be detected as a change and the clients would not be
reloaded.

Also, the comparison for boolean fields such as `enabled` and
`allow_unauthenticated` was on the pointer and not the value of that pointer,
which results in spurious reloads in real config reload that is easily missed in
typical test scenarios.

Includes a minor refactor of the order of fields for `Copy` and `Merge` to match
the struct fields in hopes it makes it harder to make this mistake in the
future, as well as additional test coverage.

---

This comes out of https://github.com/hashicorp/nomad/issues/14233 but doesn't fix it as far as I can tell.

### DIFF
--- a/command/agent/agent_test.go
+++ b/command/agent/agent_test.go
@@ -1085,6 +1085,39 @@ func TestServer_Reload_TLS_DowngradeFromTLS(t *testing.T) {
 	assert.True(agent.config.TLSConfig.IsEmpty())
 }
 
+func TestServer_Reload_VaultConfig(t *testing.T) {
+	ci.Parallel(t)
+
+	logger := testlog.HCLogger(t)
+
+	agentConfig := &Config{
+		TLSConfig: &config.TLSConfig{},
+		Vault: &config.VaultConfig{
+			Enabled:   pointer.Of(true),
+			Token:     "vault-token",
+			Namespace: "vault-namespace",
+		},
+	}
+
+	agent := &Agent{
+		auditor: &noOpAuditor{},
+		logger:  logger,
+		config:  agentConfig,
+	}
+
+	newConfig := &Config{
+		TLSConfig: &config.TLSConfig{},
+		Vault: &config.VaultConfig{
+			Enabled:   pointer.Of(true),
+			Token:     "vault-token",
+			Namespace: "vault-namespace",
+		},
+	}
+
+	must.NoError(t, agent.Reload(newConfig))
+	must.Equals(t, agent.config.Vault, newConfig.Vault)
+}
+
 func TestServer_ShouldReload_ReturnFalseForNoChanges(t *testing.T) {
 	ci.Parallel(t)
 	assert := assert.New(t)
--- a/command/agent/agent_test.go
+++ b/command/agent/agent_test.go
@@ -1088,34 +1088,34 @@ func TestServer_Reload_TLS_DowngradeFromTLS(t *testing.T) {
 func TestServer_Reload_VaultConfig(t *testing.T) {
 	ci.Parallel(t)
 
-	logger := testlog.HCLogger(t)
-
-	agentConfig := &Config{
-		TLSConfig: &config.TLSConfig{},
-		Vault: &config.VaultConfig{
+	agent := NewTestAgent(t, t.Name(), func(c *Config) {
+		c.Server.NumSchedulers = pointer.Of(0)
+		c.Vault = &config.VaultConfig{
 			Enabled:   pointer.Of(true),
 			Token:     "vault-token",
 			Namespace: "vault-namespace",
-		},
+			Addr:      "https://vault.consul:8200",
+		}
+	})
+	defer agent.Shutdown()
+
+	newConfig := agent.GetConfig().Copy()
+	newConfig.Vault = &config.VaultConfig{
+		Enabled:   pointer.Of(true),
+		Token:     "vault-token",
+		Namespace: "another-namespace",
+		Addr:      "https://vault.consul:8200",
 	}
 
-	agent := &Agent{
-		auditor: &noOpAuditor{},
-		logger:  logger,
-		config:  agentConfig,
-	}
+	sconf, err := convertServerConfig(newConfig)
+	must.NoError(t, err)
+	agent.finalizeServerConfig(sconf)
 
-	newConfig := &Config{
-		TLSConfig: &config.TLSConfig{},
-		Vault: &config.VaultConfig{
-			Enabled:   pointer.Of(true),
-			Token:     "vault-token",
-			Namespace: "vault-namespace",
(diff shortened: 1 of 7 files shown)


nomad/structs/config/vault_test.go
```diff
@@ -66,69 +68,75 @@ func TestVaultConfig_Merge(t *testing.T) {
 func TestVaultConfig_IsEqual(t *testing.T) {
```
lgfa29, line 68:
```suggestion
func TestVaultConfig_Equals(t *testing.T) {
```

nomad/structs/config/vault_test.go
```diff
@@ -66,69 +68,77 @@ func TestVaultConfig_Merge(t *testing.T) {
...
 		TLSCertFile:          "1",
 		TLSKeyFile:           "1",
-		TLSSkipVerify:        &trueValue,
+		TLSSkipVerify:        pointer.Of(true),
 		TLSServerName:        "1",
 	}
 
-	require.True(c1.IsEqual(c2))
+	// TODO: must.Equals(t, c, c2) should work here?
+	require.True(t, c.Equals(c2))
```
tgross, line 106: @shoenig this seems like a possible bug in `test`?

command/agent/agent_test.go
```diff
@@ -1085,6 +1085,39 @@ func TestServer_Reload_TLS_DowngradeFromTLS(t *testing.T) {
...
+		config:  agentConfig,
+	}
+
+	newConfig := &Config{
+		TLSConfig: &config.TLSConfig{},
+		Vault: &config.VaultConfig{
+			Enabled:   pointer.Of(true),
+			Token:     "vault-token",
+			Namespace: "vault-namespace",
+		},
```
lgfa29, line 1114: Is `newConfig` supposed to be different from `agentConfig`? It think you would want to test if `newConfig` was applied after a `Reload`?

lgfa29 (approved):
Minor comments on the test, but LGTM!

command/agent/agent_test.go
tgross: Oh good call. And I should be checking `agent.GetConfig().Vault` here too to get the updated config.

nomad/structs/config/vault_test.go
shoenig: i believe the bug is the pointer comparison of TLSSkipVerify

````
	if c.TLSSkipVerify != b.TLSSkipVerify {
		return false
	}
```

shoenig: In fact if you fmt.Println(c1.Equals(c2)) it's false; testify is doing bad magic here to make it work

command/agent/agent_test.go
tgross: Well as it turns out, reloading the config _doesn't_ set the server config we keep in the server struct. That still holds on to the old version of the config as far as I can tell, because we're never supposed to read from it at runtime. That seems like a bug, tbh, but unrelated and there's a lot of threads to pull on that.

We do overwrite the `VaultConfig` that we write into the server's `VaultClient`, but I can't get to the one in the agent in this test. I've added an accessor method so that we can access it in the `nomad/server_test.go` test at least.

tgross: Done in [9692586](https://github.com/hashicorp/nomad/pull/14298/commits/9692586e77c3121737f6fe2c14fd6037520410c4)

nomad/structs/config/vault_test.go
tgross: Oh ho! Thanks!

tgross: Done in [3f63499](https://github.com/hashicorp/nomad/pull/14298/commits/3f6349922eab6c6c77ec925cf36095ffeb4f5aa1)

nomad/structs/config/vault.go
```diff
@@ -245,9 +245,15 @@ func (c *VaultConfig) Equals(b *VaultConfig) bool {
 	if c.TLSKeyFile != b.TLSKeyFile {
 		return false
 	}
-	if c.TLSSkipVerify != b.TLSSkipVerify {
+
+	if c.TLSSkipVerify == nil || b.TLSSkipVerify == nil {
+		if c.TLSSkipVerify != b.TLSSkipVerify {
+			return false
+		}
+	} else if *c.TLSSkipVerify != *b.TLSSkipVerify {
```
shoenig, line 253: i was thinking, maybe we could have `pointer.Compare` similar to (and replacing) `helper.CompareTimePtrs` - caveat being it should only be used on basic types (not on structs with also pointer fields)

tgross: Good idea, lemme take a crack at that.

shoenig: > caveat being it should only be used on basic types

what am I thinking, we can enforce that with a type parameter!

shoenig: pretty sure this is effectively what we'd want

```
// Compare returns whether a and b contain the same value.
func Compare[A constraints.Ordered](a, b *A) bool {
	return *a == *b
}```

Comment (schmichael) [reply nomad/structs/config/vault.go:253]: Yeah, something like this? https://go.dev/play/p/iZx7yLMGIoU

**Review by schmichael — APPROVED** [review]

File: nomad/structs/config/vault.go (hunk 4 above)
Comment (shoenig) [reply nomad/structs/config/vault.go:253]: the problem is I want to prevent comparisons like this:

https://go.dev/play/p/Lh26c-YWmBd

using `constraints.Ordered` as a constraint prevents comparing pointers of anything but underlying Numeric/String types which are safe

Comment (tgross) [reply nomad/structs/config/vault.go:253]: hrm, I seem to be stuck there.

```go
type primitivePointer interface {
	*bool | *string | *int // etc, we can add the rest here
}

// Compare is an equality check for pointers that point to primitives.
func Compare[A primitivePointer](a, b A) bool {
	if a == nil || b == nil {
		return a == b
	}

	return *a == *b
}
```

Gives me:

```
# github.com/hashicorp/nomad/helper/pointer
helper/pointer/pointer.go:28:10: invalid operation: pointers of a (variable of type A constrained by primitivePointer) must have identical base types
helper/pointer/pointer.go:28:16: invalid operation: pointers of b (variable of type A constrained by primitivePointer) must have identical base types
```

For now I'm going to merge the bugfix and then we'll come up with some nicer comparison functions in a separate PR.

tgross: Ah, I posted that without seeing the other solutions. In any case, probably better for another PR.

PR #12416, opened 2022-03-31
docs: remove deprecated client options parameters docs

tgross:
The client configuration options for drivers have been deprecated
since 0.9. We haven't torn them out completely but because they're
deprecated it's been hard to guarantee correct behavior. Remove the
documentation so that users aren't misled about their viability.

Closes https://github.com/hashicorp/nomad/issues/12030

PR #14300, opened 2022-08-24
Allow jobs not requiring network resources with network fingerprinter disabled

wjordan:
Jobs not requiring any network resources should be allowed even when the network fingerprinter is disabled.

My use-case: Nomad Client running on a host with thousands of virtual network interfaces where the `network` fingerprinter takes a long time (several minutes), and the job definitions aren't allocating any network resources. I would like to add `network` to the `fingerprint.denylist` option, but this makes the scheduler fail to place any job because the current `NetworkChecker` fails if the node doesn't have any network. Allowing jobs not requiring any network resources would let me disable the `network` fingerprinter while still placing jobs properly.

### DIFF
--- a/scheduler/feasible.go
+++ b/scheduler/feasible.go
@@ -381,6 +381,11 @@ func (c *NetworkChecker) SetNetwork(network *structs.NetworkResource) {
 }
 
 func (c *NetworkChecker) Feasible(option *structs.Node) bool {
+	// Allow jobs not requiring any network resources
+	if c.networkMode == "none" {
+		return true
+	}
+
 	if !c.hasNetwork(option) {
 
 		// special case - if the client is running a version older than 0.12 but
--- a/.changelog/14300.txt
+++ b/.changelog/14300.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+scheduler: Allow jobs not requiring network resources even when no network is fingerprinted
+```


scheduler/feasible.go
```diff
@@ -381,6 +381,11 @@ func (c *NetworkChecker) SetNetwork(network *structs.NetworkResource) {
 }
 
 func (c *NetworkChecker) Feasible(option *structs.Node) bool {
+	// Allow jobs not requiring any network resources
+	if c.ports == nil {
+		return true
+	}
```
tgross, line 387: There's a [`network.mode = "none"`](https://www.nomadproject.io/docs/job-specification/network#none) that I would expect to be set in the scenario you've described. Rather than using nil ports as an implied "no network", it'd be better if we checked explicitly on that mode. That way we're not confusing potential misconfiguration for an intentional user instruction.

tgross (changes requested):
Hi @wjordan! This is a great idea -- I always love finding ways to do less work in the scheduler. I've left a suggestion about a way we could make this a bit more intentional in behavior.

Also, you can add a changelog file at `.changelog/14300.txt` similar to [this one](https://github.com/hashicorp/nomad/blob/main/.changelog/11245.txt)? Something like "scheduler: Allow jobs not...", etc. Thanks!

wjordan:
@tgross I've updated the PR based on your suggestion to allow `network.mode = "none"` to not require network resources, please take another look.

tgross (approved):
LGTM, thanks @wjordan! We're in the release candidate window for Nomad 1.4.0 and this didn't get in, but I'll merge this once we've shipped GA (next week-ish).

Tested with the following job:

```hcl
job "example" {
  type        = "batch"
  datacenters = ["dc1"]

  parameterized {
    payload = "required"
  }

  group "group" {

    network {
      mode = "none"
    }

    task "task" {

      driver = "docker"

      config {
        image   = "busybox:1"
        command = "/bin/sh"
        args    = ["-c", "cat local/payload.txt; sleep 1"]
      }

      dispatch_payload {
        file = "payload.txt"
      }

      resources {
        cpu    = 64
        memory = 500
      }

    }
  }
}
```

</details>

And with the following in my client config:

```
client {
  options {
    fingerprint.denylist = "network"
  }
}
```

(Which I'm now realizing isn't available without the deprecated `options` block, so I really should not have removed that from the docs in https://github.com/hashicorp/nomad/pull/12416 and I should revert that.)

tgross:
Hi @wjordan we had to push out a quick Nomad 1.4.1 but this has now been merged and will ship in the next regular patch release of Nomad. Thanks again!

PR #14301, opened 2022-08-24
testing: fix flakey check status test

shoenig:
This PR fixes a flakey test where we did not wait on the check
status to actually become failing (go too fast and you just get
a pending check).

Instead add a helper for waiting on any check in the alloc to become
the state we are looking for.


pkazmierczak (approved):
nice 👍

PR #14311, opened 2022-08-24
Add Namespace, Job and Group to envoy stats

jorgemarey:
This PR adds more tags (nomad namespace, job and group) to the envoy stats tags. 
It's useful when performing scaling based on envoy request being able to filter the service by nomad job/group.

Follows the work done in #12543

### DIFF
--- a/client/serviceregistration/workload.go
+++ b/client/serviceregistration/workload.go
@@ -9,27 +9,15 @@ import (
 // WorkloadServices describes services defined in either a Task or TaskGroup
 // that need to be syncronized with a service registration provider.
 type WorkloadServices struct {
-	AllocID string
-
-	// Group in which the service belongs for a group-level service, or the
-	// group in which task belongs for a task-level service.
-	Group string
-
-	// Task in which the service belongs for task-level service. Will be empty
-	// for a group-level service.
-	Task string
-
-	// JobID provides additional context for providers regarding which job
-	// caused this registration.
-	JobID string
+	AllocInfo structs.AllocInfo
 
 	// Canary indicates whether, or not the allocation is a canary. This is
 	// used to build the correct tags mapping.
 	Canary bool
 
-	// Namespace is the provider namespace in which services will be
+	// ProviderNamespace is the provider namespace in which services will be
 	// registered, if the provider supports this functionality.
-	Namespace string
+	ProviderNamespace string
 
 	// Restarter allows restarting the task or task group depending on the
 	// check_restart stanzas.
@@ -88,8 +76,8 @@ func (ws *WorkloadServices) Copy() *WorkloadServices {
 }
 
 func (ws *WorkloadServices) Name() string {
-	if ws.Task != "" {
-		return ws.Task
+	if ws.AllocInfo.Task != "" {
+		return ws.AllocInfo.Task
 	}
-	return "group-" + ws.Group
+	return "group-" + ws.AllocInfo.Group
 }
--- a/command/agent/consul/connect.go
+++ b/command/agent/consul/connect.go
@@ -3,6 +3,7 @@ package consul
 import (
 	"fmt"
 	"net"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -277,6 +278,7 @@ func injectNomadInfo(cfg map[string]interface{}, defaultTags map[string]string) 
 		tag := key + value
 		tags = append(tags, tag)
 	}
+	sort.Strings(tags) // mostly for test stability
 	cfg[configKey] = tags
 }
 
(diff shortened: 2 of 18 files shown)


client/serviceregistration/workload.go
```diff
@@ -23,6 +23,11 @@ type WorkloadServices struct {
 	// caused this registration.
 	JobID string
 
+	// NomadNamespace provides additional context for providers regarding which
+	// nomad namespace caused this registration.
+	// Note: this can be different to Namespace if the provider is not Nomad
+	NomadNamespace string
```
tgross, line 29: This name is really awkward now that we have Nomad native service discovery, because there's a `Namespace` field that may _also_ be the Nomad namespace. We're passing a bunch of individual fields from the allocation runner, which don't have an obvious purpose until we get way down in the weeds here. These should probably get all wrapped up in some AllocInfo struct that we can instantiate in the alloc runner's group service hook.

(cc @jrasell and @shoenig on that though... would love your opinions on whether this smells like a bad suggestion to you)

command/agent/consul/connect.go
```diff
@@ -226,38 +233,59 @@ func connectMeshGateway(in structs.ConsulMeshGateway) api.MeshGatewayConfig {
...
+			pair := k + value
+			tags = append(tags, pair)
+		}
+		if tags != nil {
+			cfg[key] = tags
+		}
 		return
 	}
-
 	switch v := tags.(type) {
```
tgross, line 273: I realize this is existing code but let's rename this variable while we're here (something like `existingTags` or `cfgTags`?). "v" suggest it's a "value", which is really confusing in the context where we've got key, value pairs being iterated over.

command/agent/consul/connect.go
```diff
@@ -226,38 +233,59 @@ func connectMeshGateway(in structs.ConsulMeshGateway) api.MeshGatewayConfig {
...
+// injectNomadInfo merges nomad information into cfg=>envoy_stats_tags
 //
 // cfg must not be nil
-func injectAllocID(cfg map[string]interface{}, allocID string) {
+func injectNomadInfo(cfg map[string]interface{}, defaultTags map[string]string) {
 	const key = "envoy_stats_tags"
-	const prefix = "nomad.alloc_id="
-	pair := prefix + allocID
 	tags, exists := cfg[key]
 	if !exists {
```
tgross, line 259: We're only checking for existence here and not that it's the expected shape, so we end up falling thru entirely if it's not. Maybe we should do the type switch here and also overwrite if len==0 ? That'll be the most common case by far and will save us from entering the more complex loop below.

tgross:
Thanks for the PR @jorgemarey! I've left some comments for discussion.

jorgemarey:
Thanks @tgross  for taking the time to review this. I'll update the PR with the suggested changes as soon as I can!

client/serviceregistration/workload.go
jorgemarey: Yeah, I know. I didn't know how to call this. It felt weird setting that name there. I'll wait for suggestions on the changes here.

jrasell: If we go with adding a new field for the short-term, I'd vote for `JobNamespace` which aligns with `JobID` and would allow me to half-reason each namespace value has a slightly different meaning. This is wilful ignorance to the fact when using Nomad service discovery they are exactly the same though.

I would agree having an easier to reason about and understand structure to instantiate this information would be amazing. It is worth noting this would need to be available to both the AllocRunner and TaskRunner as both include service hooks for register, update, and deregister actions. Off the top of my head though, I can't quite envisage how best this would look.

shoenig: What if the existing `Namespace` field was renamed to `ProviderNamespace`, letting this field be name `Namespace`?

But yeah, I'm also on board with @tgross in just creating some kind of `AllocInfo` struct with the common alloc specific fields and embedding that into `WorkloadServices`

command/agent/consul/connect.go
jorgemarey: Tried to do this, but I don't know if I understood you correctly. One thing that's changed respect to the previous behavior is that right now if it's not a slice the value would be overridden by the nomad default tags. Should we leave the value provided by the user in that case?

client/serviceregistration/workload.go
jorgemarey: I created a new struct called AllocInfo for this (in nomad/structs) and changed references all around (there a couple of changed files...). Also renamed the workload namespace to providerNamespace.

tgross (approved):
Hi @jorgemarey! Thanks for your patience on this. The updates you've done here LGTM 👍 . I'm going to push a changelog entry on top of the PR, rebase on main, and fix the merge conflicts and then we'll get this merged.

tgross:
Merged! This will ship in Nomad 1.4.0

PR #14316, opened 2022-08-24
fix backport of #14310

lgfa29:
`must` was not imported in the file yet in `release/1.3.x` and `Config.Copy()` was added in https://github.com/hashicorp/nomad/pull/14139, which was not backported.

I wasn't sure how to deal with `Copy()`, I thought of just copying the function over to `release/1.3.x` but it seemed unnecessary for the test.

[CI errors](https://app.circleci.com/pipelines/github/hashicorp/nomad/30498/workflows/e84f203f-c4b2-41ab-a64a-bc8b54ae441a/jobs/340995).

### DIFF
--- a/command/agent/agent_test.go
+++ b/command/agent/agent_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/hashicorp/nomad/helper/testlog"
 	"github.com/hashicorp/nomad/nomad/structs"
 	"github.com/hashicorp/nomad/nomad/structs/config"
+	"github.com/shoenig/test/must"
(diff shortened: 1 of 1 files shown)


tgross (approved):
LGTM.

Kind of unfortunate that the backport assistant doesn't wait for green tests, but we'd be waiting all day if it did. 😀

lgfa29:
> Kind of unfortunate that the backport assistant doesn't wait for green tests, but we'd be waiting all day if it did. 😀

Yeah, I was thinking about it the other day. Slow runs + flaky tests would probably mean a lot more of CI babysitting 😅

PR #14318, opened 2022-08-24
cleanup: create pointer.Compare helper function

shoenig:
This PR creates a pointer.Compare helper for comparing equality of
two pointers. Strictly only works with primitive types we know are
safe to derefence and compare using `==`.

### DIFF
--- a/helper/pointer/pointer.go
+++ b/helper/pointer/pointer.go
@@ -1,6 +1,10 @@
 // Package pointer provides helper functions related to Go pointers.
 package pointer
 
+import (
+	"golang.org/x/exp/constraints"
+)
+
 // Of returns a pointer to a.
 func Of[A any](a A) *A {
 	return &a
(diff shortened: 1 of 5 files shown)


helper/pointer/pointer.go
```diff
@@ -14,3 +18,21 @@ func Copy[A any](a *A) *A {
...
+type Primitive interface {
+	constraints.Ordered
+}
+
+// Compare returns whether a and b contain the same value.
+//
+// May only be used on pointers to primitive types, where the comparison is
+// guaranteed to be sensible. For complex types (i.e. structs) consider implementing
+// an Equals method.
+func Compare[P Primitive](a, b *P) bool {
```
schmichael, line 33: I always think of "Compare" as returning differences (like some sort of Diff struct) and "Equals" as a bool. No big deal though. Folks will know what's up either way I'm sure.

schmichael (approved):

tgross (approved):

helper/pointer/pointer.go
shoenig: The thing with "Equals" is it sounds the same as the built-in pointer equals `==`. Maybe `pointer.Eq` ? `pointer.Same` ?

PR #14319, opened 2022-08-25
Job Services: fixtures and acceptance tests

philrenaud:
Adds Mirage fixtures and basic acceptance tests for the Job Services tab

image
image


ui/mirage/factories/service.js
```diff
@@ -19,16 +24,38 @@ export default Factory.extend({
...
-            DestinationName: dasherize(faker.hacker.noun()),
-            LocalBindPort: faker.random.number({ min: 5000, max: 60000 }),
-          },
-        ],
-      },
-    },
+
+  afterCreate(service, server) {
+    if (!service.namespace) {
+      const namespace = pickOne(server.db.jobs)?.namespace || 'default';
```
ChaiWithJai, line 30: This is randomly going to pick a job and assign the namespace of that job to this Service instead of taking the related job's namespace. Which could cause weird test related bugs down the line.

ui/mirage/factories/service.js
```diff
@@ -19,16 +24,38 @@ export default Factory.extend({
...
+        nodeId: node.id,
+      });
+    }
+
+    if (server.db.jobs.findBy({ id: 'service-haver' })) {
+      if (!service.jobId) {
+        service.update({
+          jobId: 'service-haver',
+        });
+      }
```
ChaiWithJai, line 48: Should this logic be tied to the `namespace` assignment? Will this logic mix well with `allScenarios` in mirage?

ChaiWithJai (approved):
LGTM

ui/mirage/factories/service.js
philrenaud: In testing re: allScenarios, so far so good.

I don't think it should be tied to the namespace assignment; it's in the same afterCreate block but otherwise separated.

PR #14320, opened 2022-08-25
acl: add token expiration and ACL role functionality

jrasell:
This PR adds the ACL Token Expiration feature and the new ACL Role construct. All feature code has been previously reviewed before being merged into the feature branch.

I believe the feature branch includes the use of some require test calls within the command package which were recently removed. I will raise a separate PR to address this so I can keep this PR straight forward.

I feel the best merge method will be a merge commit rather than squash to preserve some of the history. I do not mind though if others prefer a squash.

Related #13120
Targets main


shoenig (approved):
LGTM! (via previous reviews)

PR #14325, opened 2022-08-25
SV API: return upserted variable to caller

angrycub:
Now that the CAS is all the way to the write transaction and we plumbed the responses back out the RPC, provide them to the HTTP API and the Go API implementation.

### DIFF
--- a/api/secure_variables.go
+++ b/api/secure_variables.go
@@ -5,16 +5,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"html/template"
 	"io"
 	"net/http"
-	"sort"
 	"strings"
 	"time"
-
-	"github.com/mitchellh/cli"
-	"github.com/mitchellh/colorstring"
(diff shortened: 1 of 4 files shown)


api/secure_variables.go
```diff
@@ -445,3 +462,83 @@ func generateUnexpectedResponseCodeError(resp *http.Response) error {
 	resp.Body.Close()
 	return fmt.Errorf("Unexpected response code: %d (%s)", resp.StatusCode, buf.Bytes())
 }
+
+type VarUI interface {
+	GetConcurrentUI() cli.ConcurrentUi
+	Colorize() *colorstring.Colorize
+}
```
tgross, line 469: None of this presentation-level stuff has callers in this PR. And I don't think it belongs here rather in the `api` package rather than our `command` package in any case (it won't build, either).

tgross (changes requested):
The API changes here LGTM, just need to clean up the build

Issue #14293, opened 2022-08-24
node drain returning complete before jobs have stopped


srisch:
### Nomad version
```
Nomad v1.2.4 (9f21b724acf115a58d9ecdc97d281aef01d36387)
Nomad v1.3.3 (latest)
```

### Operating system and Environment details
`Ubuntu 20.04`
### Issue
When draining a node that is running service jobs with a long kill timeout ( say 1h ) to finish existing traffic - the node drain command instantly returns "Drain Complete" while allocations are still in the stop/running state and waiting to safely finish. 

```
nomad node drain -enable -deadline 1h 89213336
Are you sure you want to enable drain mode for node "89213336-5dda-e359-bc0a-f34e463fda79"? [y/N] y
2022-08-24T15:30:55Z: Ctrl-C to stop monitoring: will not cancel the node drain
2022-08-24T15:30:55Z: Node "89213336-5dda-e359-bc0a-f34e463fda79" drain strategy set
2022-08-24T15:30:56Z: Drain complete for node 89213336-5dda-e359-bc0a-f34e463fda79
2022-08-24T15:30:56Z: Alloc "92d17994-9a31-be8f-edbe-a635882d05c8" marked for migration
2022-08-24T15:30:56Z: Alloc "a1c99835-cdaa-45a3-ac04-ba2a007d45f8" marked for migration
2022-08-24T15:30:56Z: Alloc "92d17994-9a31-be8f-edbe-a635882d05c8" draining
2022-08-24T15:30:56Z: Alloc "a1c99835-cdaa-45a3-ac04-ba2a007d45f8" draining
```

This behavior currently causes the nomad-autoscaler with AWS to pre-emptively shut down the node before the allocation is finished.  Documentation here https://www.nomadproject.io/docs/commands/node/drain seems to indicate that
```
By default the node drain command blocks until a node is done draining and all allocations have terminated.
```   

### Reproduction steps
Set a job with a long kill_timeout so that it stays running when the node is drained - example below for HAproxy to continue handling traffic
```
kill_signal = "SIGUSR1"
kill_timeout = "1h"
```
#### Expected Result
Node drain shouldn't be complete until all allocations have stopped
#### Actual Result
Node is marked complete before all allocations have stopped

#### Idea
I do wonder if the below code is what is returning that the drain is complete.
[...]

PR #14348, opened 2022-08-26
drain: use client status to determine drain is complete

tgross:
WIP for https://github.com/hashicorp/nomad/issues/14293

If an allocation is slow to stop because of `kill_timeout`, the node drain is
marked as complete prematurely, even though drain monitoring will continue to
report allocation migrations. This impacts the UI or API clients that monitor
node draining to shut down nodes.


shoenig (approved):
LGTM!

PR #14353, opened 2022-08-26
Update policy.mdx

kanno41:
Documentation update specified by 
https://github.com/hashicorp/nomad-autoscaler/issues/574

Original issue this fixed
https://github.com/hashicorp/nomad-autoscaler/issues/565

### DIFF
--- a/website/content/tools/autoscaling/policy.mdx
+++ b/website/content/tools/autoscaling/policy.mdx
@@ -66,6 +66,9 @@ horizontal application scaling or horizontal cluster scaling.
 
 - `query_window` - Defines how far back to query the APM for metrics. It should
   be provided as a duration (e.g.: `"5s"`, `"1m"`). Defaults to `1m`.
+  
+- `group` - Groups checks together that query that same data so that there can be 
+  non-overlapping threshold checks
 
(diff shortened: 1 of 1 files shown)


lgfa29 (changes requested):
Thanks for the contribution @kanno41!

This is actually a pretty intricate configuration that will require a bit more of explanation and examples to properly guide users on how to use it (and I should've written it sooner 😅).

Looking at the list in https://github.com/hashicorp/nomad-autoscaler/issues/574, I think `enable_debug` would be a good one to document, the description is the same as in the command help message: 
https://github.com/hashicorp/nomad-autoscaler/blob/main/command/agent.go#L54-L55

The `source` option also a good one to take on. We need to link to https://www.nomadproject.io/tools/autoscaling/agent/source from https://www.nomadproject.io/tools/autoscaling/agent/policy. So something like:

```md
### `policy` Parameters
...
- `source` <code>([Source][source]: nil)</code> - Configure scaling policy sources.
```

Where that `[source]` link points to https://www.nomadproject.io/tools/autoscaling/agent/source.

I'm going to close this PR, but feel free to reach out if you have any questions 🙂

PR #14356, opened 2022-08-26
Update ephemeral_disk.mdx to explain how to use the ephemeral disk

brettplarson:
It is really unclear on how to use this feature. it took me a while to find this, so I thought I would purpose how to use this by explaining where it is mounted and providing a link to more detailed documentation.


lgfa29 (approved):
Thanks for the improvemtns @brettplarson!

PR #14368, opened 2022-08-28
docs: template change script example correction

quoing:
"path" parameter doesn't work, should be command


pkazmierczak (approved):
thanks for catching this 👍

PR #14385, opened 2022-08-29
cgroups: refactor v2 kill path to use cgroups.kill interface file

### LINKED ISSUE #14371: cgroups: v2 should use kill interface instead of freezer
When implementing support for cgroups v2 we mostly tried to keep the v1 logic but retrofit it onto v2. This includes the block of code for forcefully killing processes related to a task https://github.com/hashicorp/nomad/blob/v1.3.4/client/lib/cgutil/group_killer.go#L98

In v2, there is now the `cgroup.kill` interface file, which we can simply write a `1` into to kill all processes present in the cgroup (i.e. show up in `cgroup.procs`). Unlike `cgroup.freeze`, `cgroup.kill` handles racy forkers.

```
Writing “1” to the file causes the cgroup and all descendant cgroups to be killed. This means that all processes located in the affected cgroup tree will be killed via SIGKILL.

Killing a cgroup tree will deal with concurrent forks appropriately and is protected against migrations.
```

shoenig:
This PR refactors the cgroups v2 group kill code path to use the
`cgroups.kill` interface file for destroying the cgroup. Previously
we copied the freeze + sigkill + unfreeze pattern from the v1 code,
but v2 provides a more efficient and more race-free way to handle
this.

Closes #14371

Backport only to 1.3.x (when v2 support was added)

### DIFF
--- a/client/lib/cgutil/group_killer.go
+++ b/client/lib/cgutil/group_killer.go
@@ -6,13 +6,12 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"path/filepath"
+	"strconv"
 	"time"
 
 	"github.com/hashicorp/go-hclog"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/cgroups/fs"
-	"github.com/opencontainers/runc/libcontainer/cgroups/fs2"
 	"github.com/opencontainers/runc/libcontainer/configs"
 )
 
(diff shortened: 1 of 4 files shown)


client/lib/cgutil/group_killer.go
```diff
@@ -96,56 +95,24 @@ func (d *killer) v1(cgroup *configs.Cgroup) error {
...
+	if err := editSelf.write("cgroup.procs", strconv.Itoa(d.pid)); err != nil {
+		return err
 	}
 
-	// do the common kill logic
-
-	if err = d.kill(path, freeze, thaw); err != nil {
+	// write "1" to cgroup.kill
+	editTask := &editor{cgroup.Path}
+	if err := editTask.write("cgroup.kill", "1"); err != nil {
```
tgross, line 110: Are the paths we try to write guaranteed to exist? And if not, should we maybe wrap the errors we get from here so that they're a little less inscrutable to the end user?

tgross (approved):
LGTM

PR #14397, opened 2022-08-30
search: enforce correct ACL for search over variables

tgross:
When the Search API was updated for variables, we didn't have the ACLs finished yet. We missed a spot when updating this endpoint for the updated ACLs.


tgross:
Whoops, missed that there were tests over in command/agent for this. Back to draft till that's fixed.

shoenig (approved):
LGTM!

PR #14398, opened 2022-08-30
e2e: add ACL test suite with ACL Role test.

jrasell:
This adds a new ACL test suite to the e2e framework which includes
an initial test for ACL roles. The ACL test includes a helper to
track and clean created Nomad resources which keeps the test
cluster clean no matter if the test fails early or not.

Related #13120

### DIFF
--- a/e2e/acl/acl_role_test.go
+++ b/e2e/acl/acl_role_test.go
@@ -1,0 +1,153 @@
+package acl
+
+import (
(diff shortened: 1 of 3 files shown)


DerekStrickland (approved):
Very nice!

shoenig (approved):
LGTM!

PR #14399, opened 2022-08-30
Vars: Update CT dependency to support variables.

angrycub:
Updates go.mod and go.sum to use a version of consul-template provided by  https://github.com/hashicorp/consul-template/pull/1632; specifically at commit https://github.com/hashicorp/consul-template/pull/1632/commits/21d2c9bb9752c77b3905c14282824efd2d67cf3a


tgross (approved):
LGTM!

PR #14400, opened 2022-08-30
Vars: CLI commands for get, put, purge [wip]

angrycub:
- Update return func; delete mergapasta
- Add snake case HCL identifiers
- SV CLI: remaining commands initial commit
- Add shared presentation code
- Update creates; share output implementations
- Update for secure variable rename


command/var_put.go
```diff
@@ -0,0 +1,472 @@
...
+		out.Path = path
+		out.Namespace = c.Meta.namespace
+		out.Items = make(map[string]string)
+		return out, nil
+	}
+	switch c.inFmt {
+	case "json":
+		err = json.Unmarshal(c.contents, out)
+		if err != nil {
+			return nil, fmt.Errorf("error unmarshaling json: %w", err)
```
tgross, line 329: We chatted previously about why automatically detecting the input format is challenging. But it might be nice in the case where there's a parsing error if we could suggest trying a different input format? Here's what passing HCL to the default format looks like:

```
$ nomad var put -namespace prod yet/another/path @spec.nsv.hcl
Failed to parse variable data: error unmarshaling json: invalid character '#' looking for beginning of value
```

command/var_put.go
```diff
@@ -0,0 +1,472 @@
...
+		}
+	case "hcl":
+		out, err = parseVariableSpec(c.contents)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing hcl: %w", err)
+		}
+	case "":
+		fmt.Println("format flag required")
```
tgross, line 337:
```suggestion
		c.Ui.Error("format flag required")
```

command/var_get.go
```diff
@@ -0,0 +1,182 @@
...
+			return 1
+		}
+		c.Ui.Error(fmt.Sprintf("Error retrieving variable: %s", err))
+		return 1
+	}
+	// If the user provided an item key, return that value instead of the whole
+	// object
+	if item != "" {
+		if v, ok := sv.Items[item]; ok {
+			fmt.Print(v)
```
tgross, line 129: Should this be `c.Ui.Output`?

command/var_put.go
```diff
@@ -0,0 +1,472 @@
...
+		return nil, fmt.Errorf("unknown format flag value")
+	}
+
+	// Handle cases where values are provided by CLI flags that modify the
+	// the created variable. Typical of a "copy" operation, it is a convenience
+	// to reset the Create and Modify metadata to zero.
+	var resetIndex bool
+
+	// Step on the namespace in the object if one is provided by flag
+	if c.Meta.namespace != "" && c.Meta.namespace != out.Path {
```
tgross, line 349:
```suggestion
	if c.Meta.namespace != "" && c.Meta.namespace != out.Namespace {
```

Found this because it results in weird CAS errors if you pipe in a variable you get from `var get`, but _only_ when you include the `-namespace` flag unnecessarily!

command/var_get.go
```diff
@@ -0,0 +1,182 @@
...
+  -item <item key>
+     Print only the value of the given item. Specifying this option will
+     take precedence over other formatting directives. The result will not
+     have a trailing newline making it ideal for piping to other processes.
+
+  -out ( go-template | hcl | json | none | pretty-json | table )
+     Format to render the variable in. When using "go-template", you must
+     provide the template content with the "-template" option. Defaults
+     to "table" when stdout is a terminal and to "json" when stdout is
+     redirected.
```
tgross, line 44: This little detail turns out to be _really_ elegant for what you've done for taking input from stdin. :+1:

command/var_put.go
```diff
@@ -0,0 +1,472 @@
...
+}
+
+func (c *VarPutCommand) Help() string {
+	helpText := `
+Usage: nomad var put [options] <path> [<key>=<value>]...
+
+  The 'var put' command is used to create or update an existing variable.
+  The items to be stored in the variable can be supplied as a series of
+  key-value pairs. The value for a key-value pair can be a string, an
+  @-prefixed file reference, or a '-' to get the value from standard input.
```
tgross, line 36: We're not really documenting that you can pass the entire variable in from file/stdin, only that you can pass in individual key values from string/file/stdin. i.e. that both `nomad var put @myfile.json` and `nomad var put some/path user=me password@mypassword.txt` work. You can kind of pick it up from the `-in` flag below but probably better to be explicit about that here.

command/var_put.go
```diff
@@ -0,0 +1,472 @@
...
+		// detect format based on file extension
+		switch strings.ToLower(p[len(p)-1]) {
+		case "json":
+			c.inFmt = "json"
+		case "hcl":
+			c.inFmt = "hcl"
+		default:
+			c.Ui.Error(fmt.Sprintf("Unable to determine format of %s; Use the -in flag to specify it.", filepath))
+			return 1
+		}
```
tgross, line 196: There's a nice helper in the stdlib for this:

```suggestion
		// detect format based on file extension
		fpath := arg[1:]
		switch filepath.Ext(fpath) {
		case "json":
			c.inFmt = "json"
		case "hcl":
			c.inFmt = "hcl"
		default:
			c.Ui.Error(fmt.Sprintf("Unable to determine format of %s; Use the -in flag to specify it.", filepath))
			return 1
		}
```

command/var_put.go
```diff
@@ -0,0 +1,472 @@
...
+
+	args = args[1:]
+	switch {
+	// Handle second argument: can be -, @file, or kv
+	case len(args) == 0:
+		// no-op
+	case isArgStdinRef(args[0]):
+		verbose(fmt.Sprintf("Creating variable %q using specification from stdin", path))
+		stat, _ := os.Stdin.Stat()
+		if (stat.Mode() & os.ModeCharDevice) == 0 {
```
tgross, line 217: Is the idea here that you're preventing someone from accidentally doing something like `cat /dev/urandom | nomad var put`?

tgross (approved):
I checked this PR out and put the CLI through its paces and this looks great @angrycub ! I've left a whole bunch of comments, but most are tiny UX polishes that I can take or leave. Just make sure the couple of little bugs I've identified get fixed and the debugging code gets stripped out and then I say let's get this merged!

command/var_get.go
angrycub: c.Ui.Output was forcing a terminal linefeed on me, where in this case I want the output to be as raw as possible.

tgross: Ah! Right, makes sense.

command/var_put.go
angrycub: Took a swing at being more explicit here. Please give it a new read.

command/var_list_test.go
```diff
@@ -95,11 +97,11 @@ func TestVarListCommand_Offline(t *testing.T) {
...
 }
 
-// TestVarListCommand_Online contains all of the tests that use a testagent.
-// They reuse the same testagent so that they can run in parallel and minimize
+// TestVarListCommand_Online contains all of the tests that use a testServer.
+// They reuse the same testServer so that they can run in parallel and minimize
 // test startup time costs.
 func TestVarListCommand_Online(t *testing.T) {
-	ci.Parallel(t)
+	//ci.Parallel(t)
```
tgross, line 104: Should this line just be deleted, or was it commented-out during test debugging?

command/var_list_test.go
```diff
@@ -13,13 +13,15 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// cSpell:ignore Pather
```
tgross, line 16: Do we run this in CI?

command/var_get.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+func (c *VarGetCommand) Help() string {
+	helpText := `
+Usage: nomad var get [options] <path>
+
+  The 'var get' command is used to get the contents of an existing variable.
+
+  If ACLs are enabled, this command requires a token with the 'var:read'
+  capability.
```
tgross, line 27: I'm just noticing how we've described the ACL requirements here and while I know what this means, it doesn't match up with how the ACL policy is really specified. I'd probably write this out as:

```suggestion
  If ACLs are enabled, this command requires a token with the 'variables:read'
  capability for the variable's namespace.
```

(Similar for the other commands.)

tgross (approved):
LGTM!

command/var_get.go
```diff
@@ -0,0 +1,182 @@
...
+		flags.StringVar(&c.outFmt, "out", "json", "")
+	}
+	flags.StringVar(&c.tmpl, "template", "", "")
+	flags.StringVar(&item, "item", "", "")
+
+	if err := flags.Parse(args); err != nil {
+		return 1
+	}
+
+	// Check that we got one or two arguments
```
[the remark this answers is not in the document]
angrycub: Fixed the comment. Yeah, not sure if flags can cope with this case well.

Perhaps we could use a [TextVar](https://pkg.go.dev/flag#TextVar) and do extra validation on the `out` flag. However, then we'd still  have the same problem for namespace if you changed the order. Since `-namespace` comes from the common flagset, it might present an adoption challenge for TextVar—a 1.19 addition—into code where backport is a higher probability.

There's [Var](https://pkg.go.dev/flag#Var) flags, which seem like they would give us the validation layer we need IF we want to add the extra scaffolding that they require.

command/var_put.go
```diff
@@ -0,0 +1,472 @@
...
+  capability.
+
+General Options:
+
+  ` + generalOptionsUsage(usageOptsDefault) + `
+
+Apply Options:
+  -check-index
+     If set, the variable is only purged if the server side version's modify
+     index matches the provided value.
```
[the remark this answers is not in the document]
angrycub: I tried to make the check-index text more portable across commands so that it could be more copypastable and I think I got to the heart of your point here. Can rework as needed.

command/var_put.go
```diff
@@ -0,0 +1,472 @@
...
+		sv.ModifyIndex = checkIndex
+	}
+
+	if forceOverwrite {
+		sv, _, err = client.Variables().Update(sv, nil)
+	} else {
+		sv, _, err = client.Variables().CheckedUpdate(sv, nil)
+	}
+	if err != nil {
+		c.Ui.Error(fmt.Sprintf("Error creating variable: %s", err))
```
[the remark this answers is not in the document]
angrycub: This one got a huge rework. More error detail and a better call to action like you mentioned here.

command/var_list_test.go
angrycub: Not _yet_. It was some unintended bleedover from some code quality work I was doing in another branch.

command/var_list_test.go
angrycub: I was having local challenges running all of the tests without commenting out parallelism on this one test. The odd thing is, it's set up with its own `testServer` so I am now scratching my head about that...

command/var_put.go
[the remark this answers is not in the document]
tgross: Looks great

PR #14405, opened 2022-08-30
testing: setting env var incompatible with parallel tests

tgross:
Neither the `os.Setenv` nor `t.Setenv` helper are safe to use in parallel tests
because environment variables are process-global. The stdlib panics if you try
to do this. Remove the `ci.Parallel()` call from all tests where we're setting
environment variables.

Switched to `t.Setenv` where possible to take advantage of automatic cleanup.
Fixes panics trying to run these tests locally with `go test ./$packagename`


shoenig (approved):
LGTM!

PR #12130, opened 2022-02-24
tests: deflake test that joins a server with non-voting servers to form quorum

shoenig:
This PR
 - upgrades the serf library
 - has the test start the join process using the un-joined server first
 - disables schedulers on the servers
 - uses the WaitForLeader and wantPeers helpers

Not sure which, if any of these actually improves the flakiness of this test.

Closes https://github.com/hashicorp/nomad/issues/12115 ... hopefully :cat2:

PR #14407, opened 2022-08-30
ci: fix TestNomad_BootstrapExpect_NonVoter test

lgfa29:
PR #12130 refactored the test to use the `wantPeers` helper, but this
function only returns the number of voting peers, which in this test
should be equal to 2.

I think the tests were passing back them because of a bug in Raft
(https://github.com/hashicorp/raft/pull/483) where a non-voting server
was able to transition to candidate state.

One possible evidence of this is that a successful test run would have
the following log line:

```
raft@v1.3.5/raft.go:1058: nomad.raft: updating configuration: command=AddVoter server-id=127.0.0.1:9101 server-addr=127.0.0.1:9101 servers="[{Suffrage:Voter ID:127.0.0.1:9107 Address:127.0.0.1:9107} {Suffrage:Voter ID:127.0.0.1:9105 Address:127.0.0.1:9105} {Suffrage:Voter ID:127.0.0.1:9103 Address:127.0.0.1:9103} {Suffrage:Voter ID:127.0.0.1:9101 Address:127.0.0.1:9101}]"
```

This commit reverts the test logic to check for peer count, regardless
of voting status.

I ran it [a few times in CI](https://app.circleci.com/pipelines/github/hashicorp/nomad-enterprise?branch=ci-fix-bootstrap-expect-non-voter) and there were no failures in this test anymore.

### DIFF
--- a/nomad/serf_test.go
+++ b/nomad/serf_test.go
@@ -326,10 +326,10 @@ func TestNomad_BootstrapExpect_NonVoter(t *testing.T) {
 	})
 	defer cleanupS4()
 
-	servers := []*Server{s1, s2, s3, s4}
+	// Start with 4th server for higher chance of success when joining servers.
+	servers := []*Server{s4, s3, s2, s1}
 
 	// Join with fourth server (now have quorum)
-	// Start with 4th server for higher chance of success
 	TestJoin(t, servers...)
 
 	// Assert leadership with 4 peers
--- a/nomad/serf_test.go
+++ b/nomad/serf_test.go
@@ -11,7 +11,6 @@ import (
 	"github.com/hashicorp/nomad/testutil"
 	"github.com/hashicorp/raft"
 	"github.com/hashicorp/serf/serf"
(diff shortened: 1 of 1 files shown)


nomad/serf_test.go
```diff
@@ -327,16 +326,35 @@ func TestNomad_BootstrapExpect_NonVoter(t *testing.T) {
 	})
 	defer cleanupS4()
 
+	servers := []*Server{s1, s2, s3, s4}
+
 	// Join with fourth server (now have quorum)
 	// Start with 4th server for higher chance of success
```
tgross, line 332: This line of the comment is no longer relevant I think?

```suggestion
```

tgross (approved):
LGTM

nomad/serf_test.go
lgfa29: Oh I didn't notice that the order was important. I pushed a commit to store the server in reverse order and moved this comment 👍

PR #14410, opened 2022-08-30
[ui] Jobs/jobid/services table sorting

### LINKED ISSUE #14387: [ui] Service Discovery: table sorting
at jobs/:jobid/services

philrenaud:
Resolves #14387

Implements sorting on the Services table, and makes a style adjustment to move provider icon inline next to service name


ChaiWithJai (approved):
great! but how does this forward query parameters when we make the network request? did we already handle that behavior?

philrenaud:
> great! but how does this forward query parameters when we make the network request? did we already handle that behavior?

Sorting and filtering job services is not supported as part of our API, per https://www.nomadproject.io/api-docs/jobs#job-services, so all this PR seeks to do is handle it within the UI.

PR #14418, opened 2022-08-31
e2e: add acl test for token expiration.

jrasell:
In order to add an E2E test to cover token expiration, the server
config has been updated to include a low minimum allowed TTL
value. For ease of reading, the max value is also set.

### DIFF
--- a/e2e/acl/acl_token_test.go
+++ b/e2e/acl/acl_token_test.go
@@ -1,0 +1,157 @@
+package acl
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/nomad/api"
+	"github.com/hashicorp/nomad/e2e/e2eutil"
+	"github.com/hashicorp/nomad/helper/uuid"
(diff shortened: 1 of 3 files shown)


DerekStrickland (approved):
:shipit:

e2e/acl/acl_token_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+	cleanUpProcess.add(tokenQuickExpiryCreateResp.AccessorID, aclTokenTestResourceType)
+
+	// Block the test (sorry) until the token has expired.
+	timer, timerStopFn := helper.NewSafeTimer(tokenQuickExpiry.ExpirationTTL)
+	defer timerStopFn()
+
+	select {
+	case <-timer.C:
+	}
```
shoenig, line 129: could probably just be `time.Sleep(expireTTL)` (not like we have a context to cancel from)

shoenig (approved):
LGTM!

PR #14424, opened 2022-08-31
cli: ignore VaultToken when generating job diff

### LINKED ISSUE #14423: nomad job plan leaks VAULT_TOKEN in output
### Nomad version
`Nomad v1.3.4+ent`

Debian 11

### Issue
`nomad job plan` now requires a vault token to successfully run. The diff returned by the plan now includes the vault token, even if the token was submitted via environment variable or CLI option. This is a potential security risk - we run plans as part of our deploy pipeline, and this would result in the tokens being leaked in logs.

### Reproduction steps
```
$  export VAULT_TOKEN=<THE ACTUAL VAULT TOKEN>
$ nomad job plan -verbose  /tmp/20220831121752/nomad.hcl.final
+/- Job: "sysad-sandbox-corp-dc6"
+/- Meta[DOCKER_TAG]: "0.0.8-master-121" => "master-latest"
+/- Meta[GIT_REPO]:   "https://bitbucket.savagebeast.com/scm/SYSAD/sysad-sandbox.git" => "ssh://git@bitbucket.savagebeast.com:2222/sysad/sysad-sandbox.git"
+   VaultToken:       "<THE ACTUAL VAULT TOKEN>"
+/- Task Group: "sysad-sandbox-group" (1 create/destroy update)
  +/- Task: "sysad-sandbox" (forces create/destroy update)
    +/- Config {
    ...
```

#### Expected Result
If the VAULT_TOKEN was not included in the original hcl file (i.e. passed via environment variable or CLI option), `nomad job plan` should not include it in the diff output.

#### Actual Result
The vault token is there, naked and for all to see ;)

#### Workaround
We can add a grep or other filter to scrub the vault token from the output, but this is both inconvenient and fragile

See also: #14422

lgfa29:
Backport only to 1.3.x because that's when `-vault-token` flag was added to the  `plan` command.

From our [docs](https://www.nomadproject.io/docs/commands/job/run#vault-token):

> This token is cleared from the job after validating and cannot be used within the job executing environment.

So the token shouldn't be considered in the diff as it's not used internally.

Closes #14423

### DIFF
--- a/nomad/structs/diff.go
+++ b/nomad/structs/diff.go
@@ -69,7 +69,7 @@ func (j *Job) Diff(other *Job, contextual bool) (*JobDiff, error) {
 	diff := &JobDiff{Type: DiffTypeNone}
 	var oldPrimitiveFlat, newPrimitiveFlat map[string]string
 	filter := []string{"ID", "Status", "StatusDescription", "Version", "Stable", "CreateIndex",
-		"ModifyIndex", "JobModifyIndex", "Update", "SubmitTime", "NomadTokenID"}
+		"ModifyIndex", "JobModifyIndex", "Update", "SubmitTime", "NomadTokenID", "VaultToken"}
 
 	if j == nil && other == nil {
 		return diff, nil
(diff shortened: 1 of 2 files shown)


shoenig (approved):
LGTM!

PR #14433, opened 2022-09-01
chore: remove use of "err" a log line context key for errors.

jrasell:
Log lines which include an error should use the full term "error"
as the context key. This provides consistency across the codebase
and avoids a Go style which operators might not be aware of.


DerekStrickland (approved):
Nice.  I wonder if we could find a linter rule for this moving forward.

jrasell:
> Nice. I wonder if we could find a linter rule for this moving forward.

I tried coming up with a semgrep rule, but I couldn't figure it out. It likely someone with better knowledge will be able to though.

shoenig (approved):

Issue #9570, opened 2020-12-08
Adopt raft-autopilot library


notnoop:
Nomad currently directly imports Consul's private autopilot package (`github.com/hashicorp/consul/agent/consul/autopilot`). Last month, Consul extracted autopilot to be its own module to ease adoption in other products: https://github.com/hashicorp/raft-autopilot 

Let's move off our direct dependency on the internal repo in preference to the new module.

tgross:
Note to implementor: probably worth doing https://github.com/hashicorp/nomad/issues/13219 at the same time.

shoenig:
Took a brief look into making the switch; looks like it's going to be a bit more complicated than just renaming imports. In addition to `raft-autopilot` there is also `raft-autopilot-enterprise`, and the features referenced by Nomad in the old Consul autopilot library have been split up between the two.

Still very much want to make this happen - without it Nomad is forced to import consul v1.7 :disappointed:

Issue #13931, opened 2022-07-28
test flake in TestAutopilot_RollingUpdate


tgross:
Example run https://app.circleci.com/pipelines/github/hashicorp/nomad/28734/workflows/9c27bd15-423e-437e-be7e-eaf456d23624/jobs/317342. CircleCI has identified this test as flaky with multiple failures on `main` in the last 30 days.

<details><summary>test logs</summary>

> Failed
=== RUN   TestAutopilot_RollingUpdate
2022-07-27T11:17:24.636Z [TRACE] eventer/eventer.go:74: plugin_loader.exec: task event loop shutdown: plugin_dir=""
2022-07-27T11:17:24.636Z [TRACE] eventer/eventer.go:74: plugin_loader.mock_driver: task event loop shutdown: plugin_dir=""
2022-07-27T11:17:24.636Z [TRACE] eventer/eventer.go:74: plugin_loader.raw_exec: task event loop shutdown: plugin_dir=""
2022-07-27T11:17:24.636Z [TRACE] eventer/eventer.go:74: plugin_loader.java: task event loop shutdown: plugin_dir=""
2022-07-27T11:17:24.636Z [TRACE] eventer/eventer.go:74: plugin_loader.qemu: task event loop shutdown: plugin_dir=""
2022-07-27T11:17:24.636Z [TRACE] eventer/eventer.go:74: plugin_loader.docker: task event loop shutdown: plugin_dir=""
2022-07-27T11:17:24.636Z [TRACE] eventer/eventer.go:74: plugin_loader.docker: task event loop shutdown: plugin_dir=""
2022-07-27T11:17:24.636Z [TRACE] eventer/eventer.go:74: plugin_loader.mock_driver: task event loop shutdown: plugin_dir=""
2022-07-27T11:17:24.636Z [TRACE] eventer/eventer.go:74: plugin_loader.raw_exec: task event loop shutdown: plugin_dir=""
node-063 2022-07-27T11:17:24.637Z [TRACE] nomad/worker.go:330: worker: changed workload status: worker_id=32cec098-6878-ee8d-8777-6169e91b319c from=WaitingToDequeue to=Paused
2022-07-27T11:17:24.637Z [TRACE] eventer/eventer.go:74: plugin_loader.qemu: task event loop shutdown: plugin_dir=""
2022-07-27T11:17:24.637Z [TRACE] eventer/eventer.go:74: plugin_loader.exec: task event loop shutdown: plugin_dir=""
[...]

Issue #13930, opened 2022-07-28
test flake in TestLeader_RollRaftServer


tgross:
Example run: https://app.circleci.com/pipelines/github/hashicorp/nomad/28563/workflows/e087b535-aa13-470a-97d2-384c4cb9ae05/jobs/315033. This has failed frequently on `main`, but I've seen it a bunch on other branches as well. Note that when this fails it can take >1min to do so, which also slows down test runs. See the CircleCI link for logs, which are very long for this one.

This test exercises similar code paths to https://github.com/hashicorp/nomad/issues/13931

PR #14441, opened 2022-09-01
migrate autopilot implementation to raft-autopilot

tgross:
Nomad's original autopilot was importing from a private package in Consul. It
has been moved out to a shared library. Switch Nomad to use this library so that
we can eliminate the import of Consul, which is necessary to build Nomad ENT
with the current version of the Consul SDK. This also will let us pick up
autopilot improvements shared with Consul more easily.

Fixes https://github.com/hashicorp/nomad/issues/9570 (associated with https://github.com/hashicorp/nomad-enterprise/pull/836)


go.mod
```diff
@@ -40,7 +40,6 @@ require (
 	github.com/gorilla/websocket v1.5.0
 	github.com/gosuri/uilive v0.0.4
 	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0
-	github.com/hashicorp/consul v1.7.8
```
lgfa29, line 43: 🎉

nomad/autopilot.go
```diff
@@ -20,88 +22,227 @@ const (
 	AutopilotVersionTag = "ap_version"
 )
 
-// AutopilotDelegate is a Nomad delegate for autopilot operations.
+// AutopilotDelegate is a Nomad delegate for autopilot operations. It implements
+// the autopilot.ApplicationIntegration interface and methods required for that
+// have been documented as such below.
```
lgfa29, line 27:
```suggestion
// the autopilot.ApplicationIntegration interface and methods required that
// have been documented as such below.
```

nomad/autopilot.go
```diff
@@ -20,88 +22,227 @@ const (
...
-	future := d.server.raft.GetConfiguration()
-	if err := future.Error(); err != nil {
-		return nil, fmt.Errorf("failed to get raft configuration: %v", err)
+// RemoveFailedServer will be called by autopilot to notify Nomad to remove the
+// server in a failed state. This method is required to implement the
+// ApplicationIntegration interface. (Note this is expected to return
+// immediately so we'll spawn a goroutine for it.)
+func (d *AutopilotDelegate) RemoveFailedServer(failedSrv *autopilot.Server) {
+	go func() {
+		err := d.server.RemoveFailedNode(failedSrv.Name)
```
lgfa29, line 88: Is `Name` here the [agent's `name`](https://www.nomadproject.io/docs/configuration#name)? Would it be possible to remove by ID instead in case the cluster has servers with duplicate names (which I _think_ is allow, though not a good idea 😅)?

tgross:
I was hoping this would fix the test flakes in https://github.com/hashicorp/nomad/issues/13931 and https://github.com/hashicorp/nomad/issues/13930 but given the behavior isn't supposed to have changed I guess I see why it doesn't. I'll pick those up as part of improving our E2E story for autopilot before 1.4.0-GA.

lgfa29 (approved):
This looks great! It may be good to test (if you haven't yet) interoperability between servers using the old Consul autopilot and this one, like it would happen during a rolling cluster upgrade.

nomad/autopilot.go
tgross: It's the `serf.Member.Name`, which I think is the same thing. I originally had this by ID (which is what Consul does) but our existing `RemoveFailedNode` code only accepts the name and not the ID. I didn't want to change the behavior here.

nomad/autopilot.go
tgross: Reworked a bit for clarity

Issue #14429, opened 2022-09-01
Data erased by moving a raft region leader into another region 


zizon:
### Nomad version
Output from `nomad version`
Nomad v1.2.6

### Issue
Moving a raft leader under region A to another region B, by changing region field in config file, causing region B raft state being overwrite by A thus causing all running state in region B being lost.

We had encounter this severe disaster recently.
For brief summery, we had,
1. **10.129.100.233** in a region named **not-assign** which are aimed to be moving into another region named **non-live-stable**(raft leader of the cluster that time was **10.129.110.18** )
2. we did not cleanup/remove the server files of 10.129.100.233, meaning we keep the **/var/lib/nomad/server** as it was.
3. after change  region in config from not-assign to non-live-stable, and restarted 10.129.100.233, we notice that  10.129.110.18 are being raft log force overwrite.

I will attach the associated log of each latter at bottom.

The root cause of such disaster is in involve a operation out of standard, but maybe we should revise the raft restore procedure.

1. when 10.129.100.233 restarted, it first restore the old, and incorrect raft state, from persisted state.
2. then 10.129.100.233 had the raft state of **not-assign**  but configured as a **non-live-stable** server
3. via serf, 10.129.100.233 and 10.129.110.18 discovered each other, and of course all other federated servers.
4. 10.129.110.18, as the that time leader of **non-live-stable** try to append raft logs to 10.129.100.233, but being rejected as occasionally 10.129.100.233, as was in  **not-assign** , has higher terms.
5. meanwhile, 10.129.100.233 still running as a leader of  **not-assign** after restart and re-election. 
6. as members of **non-live-stable** notice a step down of original leader(10.129.110.18), and 10.129.100.233 had add all such members into formed **not-assign**  as non-voters, via serf discovering, a log overwrite occurs.

### Nomad Server logs (if appropriate)
```
# before restarted
10.129.100.233:
 
2022-08-30T16:21:40.834+0800 [INFO]  nomad: shutting down server
2022-08-30T16:21:40.834+0800 [WARN]  nomad: serf: Shutdown without a Leave
2022-08-30T16:21:40.834+0800 [INFO]  nomad.raft: aborting pipeline replication: peer="{Voter df0e75eb-6503-cbd3-1889-6ba00b25781a 10.129.104.207:4647}"
[...]

PR #14443, opened 2022-09-01
docs: add warning about changing region config

lgfa29:
As pointed out in #14429, changing an agent's `region` while it still has local state can result in data loss, so we should warn about this in our documentation.

Preview:
image


tgross (approved):
LGTM!

PR #14445, opened 2022-09-01
connect: interpolate task env in config values

### LINKED ISSUE #14403: Nomad job with envoy escape hatch has no dynamic port interpolation available
### Nomad version
Nomad v1.2.9, v1.3.2

### Issue
Envoy uses a default [stream_idle_timeout](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-stream-idle-timeout) of 5 minutes which is in conflict with a few long running APIs we are using.  There are a few direct [override knobs](https://www.consul.io/docs/connect/proxies/envoy#local_request_timeout_ms) for envoy timeouts in Consul, but there is no direct knob for this `stream_idle_timeout`.  Advanced Consul Escape hatches are [available](https://www.consul.io/docs/connect/proxies/envoy#advanced-configuration), and in this case, configuring an escape hatch override for [envoy_public_listener_json](https://www.consul.io/docs/connect/proxies/envoy#envoy_public_listener_json) appears that it would solve the problem, as we could add a route public listener `idle_timeout` which will override envoy's default `stream_idle_timeout`, or directly modify the `stream_idle_timeout` itself.

However, when Nomad sets up the consul job, the dynamic port the envoy listener will use ahead of time is not known, so some Nomad interpolation appears to be necessary to be able to declare this snippet properly.  Taking the example of the Consul `envoy_public_listener_json` from the Consul link above, with slight modifications, I believe we'd need to be able to set the `connect.sidecar_service.proxy.config.envoy_public_listener_json` stanza in the Nomad job declaration to something like the following where the Nomad assigned port is interpolated and passed to the Consul escape hatch json override:

```
{
  "@type": "type.googleapis.com/envoy.config.listener.v3.Listener",
  "name": "public_listener:0.0.0.0:${NOMAD_PORT_connect_proxy_<SERVICE_NAME>}",
  "address": {
    "socket_address": {
      "address": "0.0.0.0",
      "port_value": ${NOMAD_PORT_connect_proxy_<SERVICE_NAME>}
    }
  },
  "filter_chains": [
    {
      "filters": [
        {
          "name": "envoy.filters.network.http_connection_manager",
          "typed_config": {
            "@type": "type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager",
            "stat_prefix": "public_listener",
            "http_filters": [
              {
                "name": "envoy.filters.http.router"
              }
            ],
            "route_config": {
              "name": "public_listener",
              "virtual_hosts": [
                {
                  "name": "public_listener",
                  "domains": ["*"],
                  "routes": [
                    {
                      "match": {
                        "prefix": "/"
                      },
                      "route": {
                        "cluster": "local-app",
                        "idle_timeout": "10m"
                      }
                    }
                  ]
                }
              ]
            }
          }
        }
      ]
    }
  ],
 "traffic_direction": "INBOUND"
}
```
However, interpolation of the Nomad assigned dynamic port to the Consul connect service doesn't appear available to the passed json escape hatch override in the testing I've done, which seems to eliminate the possibility of using envoy escape hatches to override parameters there aren't already direct Consul overrides for.

I've tested the above approach by deploying the job with a random hardcoded port substituted in the parameterization above for the escape hatch which will be incorrect.  Then, once the job is deployed, adjusting the above job definition for the escape hatch snippet and assigning the correct port that Nomad has utilized for the connect proxy and re-deploying/updating the job, at which point the escape hatch override does work as intended.  So it appears the only remaining issue is the ability to interpolate a Nomad dynamic port and pass it to the escape hatch snippet.

Perhaps Nomad interpolation can done in this snippet and I'm not aware of it?

lgfa29:
When configuring Consul Service Mesh, it's sometimes necessary to
provide dynamic value that are only known to Nomad at runtime. By
interpolating configuration values (in addition to configuration keys),
user are able to pass these dynamic values to Consul from their Nomad
jobs.

Closes #14403

### DIFF
--- a/.changelog/14445.txt
+++ b/.changelog/14445.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+consul: Allow interpolation of task environment values into Consul Service Mesh configuration
+```
--- a/client/taskenv/services.go
+++ b/client/taskenv/services.go
@@ -81,7 +81,12 @@ func interpolateMapStringInterface(taskEnv *TaskEnv, orig map[string]interface{}
 
 	m := make(map[string]interface{}, len(orig))
 	for k, v := range orig {
-		m[taskEnv.ReplaceEnv(k)] = v
+		envK := taskEnv.ReplaceEnv(k)
+		if vStr, ok := v.(string); ok {
+			m[envK] = taskEnv.ReplaceEnv(vStr)
+		} else {
+			m[envK] = v
+		}
 	}
 	return m
 }
--- a/client/taskenv/services_test.go
+++ b/client/taskenv/services_test.go
@@ -239,6 +239,7 @@ func TestInterpolate_interpolateConnect(t *testing.T) {
 				},
 				Config: map[string]interface{}{
 					"${config1}": 1,
+					"port":       "${port1}",
 				},
 			},
 		},
@@ -347,6 +348,7 @@ func TestInterpolate_interpolateConnect(t *testing.T) {
 				},
 				Config: map[string]interface{}{
 					"_config1": 1,
+					"port":     "12345",
 				},
 			},
 		},
--- a/website/content/docs/job-specification/gateway.mdx
+++ b/website/content/docs/job-specification/gateway.mdx
@@ -80,6 +80,7 @@ envoy_gateway_bind_addresses "<service>" {
   This option applies to terminating gateways that route to services addressed by a
   hostname.
 - `config` `(map: nil)` - Escape hatch for [Advanced Configuration] of Envoy.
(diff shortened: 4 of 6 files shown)


jrasell (approved):
LGTM.

Do we want to add any detail to the docs to mention this; seems quite a subtle enhancement which might be hard for users to find.

PR #14457, opened 2022-09-02
ui: remove extra space in menu footer

lgfa29:
Before:
image

After:
image

### DIFF
--- a/.changelog/14457.txt
+++ b/.changelog/14457.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Remove extra space when displaying the version in the menu footer.
+```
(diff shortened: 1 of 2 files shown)


ChaiWithJai (approved):
LGTM. Great catch. @philrenaud this "bug" has happened a few times wtrg to handlebars and whitespace. Can we look into our linting/prettier set-up?

philrenaud:
Updating prettier for this would be ideal. We can also start expanding our visual diff tests to try and catch these kinds of inconsistencies in the future. Thanks @lgfa29 for catching this!

philrenaud (approved):

lgfa29:
I think the Prettier config may be OK now since it didn't try to revert this change. But yeah, it may be good to double check 👍

PR #14463, opened 2022-09-05
hcl2: add strlen function and update docs.

### LINKED ISSUE #14462: HCL2 length function not taking string?
### Nomad version
v1.3.2

FreeBSD 13.1-STABLE

### Issue
HCL2 length() function, [according to the docs](https://www.nomadproject.io/docs/job-specification/hcl2/functions/collection/length), should accept an argument of any of the following types: string, array, map, or tuple.  
It seems to only accept collections though, not a string.  I know I can remove the default assignment and effectively achieve the same thing, but I'd like to solve why length() isn't taking a string. I tried strlen() as well since that's in the Terraform docs and I wondered if Nomad might have a version mismatch with what's in the docs, but nope.

### Reproduction steps
See above snippet.

#### Expected Result
Return numeric length of string.

#### Actual Result
HCL2 semantic error when running `nomad job validate myjob.hcl`. Complains that length() expects a map, array, or tuple.  

```
Error getting job struct: Error parsing job file from ./ports/_job.hcl:
_job.hcl:36,17-24: Error in function call; Call to function "length" failed: collection must be a list, a map or a tuple.
```

### Job file (if appropriate)
```
variable "portsList" {
  description = "Path to file with list of ports"
  type = string
  default = ""
  validation  {
    condition = length(var.portsList) > 0
    error_message = "Path to file containing list of ports is required."
  }
}
```

jrasell:
closes #14462

### DIFF
--- a/.changelog/14463.txt
+++ b/.changelog/14463.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+jobspec: Added `strlen` HCL2 function to determine the length of a string
+```
--- a/jobspec2/functions.go
+++ b/jobspec2/functions.go
@@ -81,6 +81,7 @@ func Functions(basedir string, allowFS bool) map[string]function.Function {
 		"slice":           stdlib.SliceFunc,
 		"sort":            stdlib.SortFunc,
 		"split":           stdlib.SplitFunc,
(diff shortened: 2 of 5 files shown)


pkazmierczak (approved):
👍

Issue #9920, opened 2021-01-29
network bandwidth limits should be removed from quotas


henrikjohansen:
The quota API [docs](https://www.nomadproject.io/api-docs/quotas) and the [quota example docs](https://learn.hashicorp.com/tutorials/nomad/quotas) still advocate the ability to limit network bandwidth usage.

Likewise the CLI also still exposes a **Network Usage** column : 

```
$ nomad quota status default-quota    
Name        = default-quota
Description = Limit the shared default namespace
Limits      = 1

Quota Limits
Region  CPU Usage  Memory Usage  Network Usage
global  0 / 2500   0 / 4096      0 / inf
```

However, the ability to express bandwidth usage is not available in the new group level network stanza and the usage of **network {}**  in the resource stanza is deprecated.

tgross:
Thanks @henrikjohansen! We already have a PR open in the ENT repo and our docs to do that, it's been waiting on some other work to complete.

tgross:
That PR and https://github.com/hashicorp/nomad/pull/9940 have been merged. That will land on the website documentation with the 1.0.4 release.

henrikjohansen:
@tgross FYI - the CLI `quota` command still lists ***Network Usage*** in Nomad v1.3.5+ent : 

```
$ nomad quota status default-quota
Name        = default-quota
Description = Limit the shared default namespace
Limits      = 1

Quota Limits
Region  CPU Usage  Memory Usage  Memory Max Usage  Network Usage
global  0 / 2500   0 / 4096      0 / inf           0 / inf
```

tgross:
Thanks @henrikjohansen. Fixed in https://github.com/hashicorp/nomad/pull/14468

PR #14468, opened 2022-09-06
cli: remove network from `quota status` output

tgross:
As reported in https://github.com/hashicorp/nomad/issues/9920#issuecomment-1236376964

Network quotas were removed in Nomad 1.0.4. Remove the fields no longer in use
from the `quota status` output.

---

This PR won't run the test here because it's in the OSS repo, but I've verified the test passes locally in the ENT repo. Adding a `fmt.Println` to the test in the appropriate spot shows the following new output:

```
Name        = test
Description = <none>
Limits      = 1

Quota Limits
Region  CPU Usage  Memory Usage  Memory Max Usage  Variables Usage
global  0 / 100    0 / inf       0 / inf           0 / inf
```

### DIFF
--- a/.changelog/14468.txt
+++ b/.changelog/14468.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+cli: Removed deprecated network quota fields from `quota status` output
+```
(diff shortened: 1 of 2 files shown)


jrasell (approved):
LGTM!

PR #14469, opened 2022-09-06
e2e: save tokens to file to use in parameterized tests

ChaiWithJai:
This PR updates the `globalSetup` script to save the tokens created during this process to a file to be used later in parameterized tests.

### DIFF
--- a/e2e/ui/global-setup.js
+++ b/e2e/ui/global-setup.js
@@ -1,3 +1,4 @@
+import fs from "fs";
 import { chromium } from "@playwright/test";
 import { client, NOMAD_ADDR, NOMAD_TOKEN } from "./api-client.js";
 import {
@@ -6,6 +7,8 @@ import {
   OPERATOR_POLICY_JSON,
   DEV_POLICY_JSON,
 } from "./utils/index.js";
+
+const formatToken = token => `export const token = "${token}"`;
 
 async function globalSetup() {
   if (NOMAD_TOKEN === undefined || NOMAD_TOKEN === "") {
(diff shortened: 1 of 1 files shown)


philrenaud (approved):
Q: are these tokens short-lived / do they get wiped after tests run? I'm a bit wary of writing out tokens to the filesystem in this way.

tgross (changes requested):
>Q: are these tokens short-lived / do they get wiped after tests run? I'm a bit wary of writing out tokens to the filesystem in this way.

The E2E cluster itself is destroyed after each test run. However, the E2E test runner environment is only destroyed after each execution when it's running on CI. If you run the E2E tests locally against a cluster, we'd be leaving tokens sitting around in files on your laptop. If it's at all possible, we should be putting these files in the same `keys` directory that we use for everything else: https://github.com/hashicorp/nomad/blob/main/e2e/terraform/scripts/bootstrap-nomad.sh#L16. That all gets `.gitignore`d too, which is nice.

e2e/ui/global-setup.js
```diff
@@ -17,12 +20,18 @@ async function globalSetup() {
...
     await client(`/acl/policy/operator`, { data: OPERATOR_POLICY_JSON });
     await client(`/acl/policy/dev`, { data: DEV_POLICY_JSON });
-    await client(`/acl/token`, {
+    
+    // Create Operator Token and save to local director to use in tests
+    const {data: {SecretID: operatorToken}} = await client(`/acl/token`, {
       data: { Name: "Operator", Type: "client", Policies: ["operator"] },
     });
-    await client(`/acl/token`, {
+    fs.writeFileSync('input/tokens/operator.js', formatToken(operatorToken));
```
ChaiWithJai, line 28: @tgross @philrenaud Great feedback

> Q: are these tokens short-lived / do they get wiped after tests run? I'm a bit wary of writing out tokens to the filesystem in this way.

> The E2E cluster itself is destroyed after each test run. However, the E2E test runner environment is only destroyed after each execution when it's running on CI. If you run the E2E tests locally against a cluster, we'd be leaving tokens sitting around in files on your laptop. If it's at all possible, we should be putting these files in the same keys directory that we use for everything else: https://github.com/hashicorp/nomad/blob/main/e2e/terraform/scripts/bootstrap-nomad.sh#L16. That all gets .gitignored too, which is nice.

So is the solution here to save to `../keys/CREATE/OWN/path.js`?

And then in the `globalTeardown` and `catch` hook, I can delete the files that were created for good measure?

tgross: I think it'd be `../terraform/keys/path.js` but I'm not sure I understand where the working directory for this execution is. But then again, this whole thing gets run in an ephemeral Docker container, right? (ref [run.sh](https://github.com/hashicorp/nomad/blob/main/e2e/ui/run.sh#L42-L52)) If that's the case, we don't need to worry about it because we throw the whole file system when we're done.

tgross (approved):

e2e/ui/global-setup.js
ChaiWithJai: > I think it'd be ../terraform/keys/path.js but I'm not sure I understand where the working directory for this execution is. But then again, this whole thing gets run in an ephemeral Docker container, right? (ref [run.sh](https://github.com/hashicorp/nomad/blob/main/e2e/ui/run.sh#L42-L52)) If that's the case, we don't need to worry about it because we throw the whole file system when we're done.

So should I research more? Or does "this whole thing gets run in an ephemeral Docker container" mean that we're good to proceed without changes?

tgross: I think we're good-to-go.

Issue #14471, opened 2022-09-06
[ui] [bug] Browser becomes laggy when polling `/checks` endpoint.


ChaiWithJai:
The browser becomes laggy when we keep the `/allocations/:allocationId/index` view open. This behavior is likely because we're long polling both the `/checks` and `/stats` endpoints even when the browser isn't active. This behavior didn't occur prior to long-polling the `/checks` endpoint on the same route.

The goal of this issue is to only long poll `/checks` when we're actively on the route.

![Image](https://user-images.githubusercontent.com/41024828/188675185-c38c9f52-daa4-490b-b880-320bd3ef1a81.png)

PR #14472, opened 2022-09-06
ui:  only long poll `/checks` endpoint when route is active

ChaiWithJai:
Closes [14471](https://github.com/hashicorp/nomad/issues/14471).

### DIFF
--- a/ui/app/routes/allocations/allocation.js
+++ b/ui/app/routes/allocations/allocation.js
@@ -48,5 +48,5 @@ export default class AllocationRoute extends Route.extend(WithWatchers) {
   @watchRecord('allocation') watch;
   @watchNonStoreRecords('allocation') watchHealthChecks;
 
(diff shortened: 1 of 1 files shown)


ui/app/routes/allocations/allocation.js
```diff
@@ -48,5 +48,5 @@ export default class AllocationRoute extends Route.extend(WithWatchers) {
   @watchRecord('allocation') watch;
   @watchNonStoreRecords('allocation') watchHealthChecks;
 
-  @collect('watch') watchers;
+  @collect('watch', 'watchHealthChecks') watchers;
```
philrenaud, line 51: Does this work because `watchers` are subject to [visibilityHandler()](https://github.com/hashicorp/nomad/blob/main/ui/app/mixins/with-watchers.js#L28-L34) ?

philrenaud (approved):

PR #14481, opened 2022-09-07
core: verify leadership before marking clients down via heartbeating

shoenig:
This PR changes failed hearbeat processing so that we use `.VerifyLeader()`
to ensure the server is still the leader, instead of `.IsLeader()`. The
difference is that the later is a cached value indicating the raft state
for the agent, and the former enforces raft-read consensus.

With this, we can create an FSM snapshot and lookup the latest Node
status for the node that failed to update its node status in time,
just to make sure we didn't receive an update in the interim.

A more extreme version of this idea would be to use `.Barrier()`, which
enforces a write through the raft concensus, ensuring that all preceding
raft operations are processed before moving forward. This might be
useful in the case where the leader has a large queue of raft operations
to process, and node updates are stuck waiting in the queue while heartbeat
timers are being processed async regardless.

### DIFF
--- a/nomad/heartbeat.go
+++ b/nomad/heartbeat.go
@@ -8,7 +8,6 @@ import (
 	metrics "github.com/armon/go-metrics"
 	log "github.com/hashicorp/go-hclog"
 	memdb "github.com/hashicorp/go-memdb"
-
 	"github.com/hashicorp/nomad/helper"
 	"github.com/hashicorp/nomad/nomad/structs"
 )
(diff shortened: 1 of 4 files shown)


nomad/heartbeat.go
```diff
@@ -142,15 +145,37 @@ func (h *nodeHeartbeater) invalidateHeartbeat(id string) {
...
+	// reset the heartbeat timer.
+	latestUpdate := time.Unix(node.StatusUpdatedAt, 0)
+	if latestUpdate.After(fencepost) {
+		_, _ = h.resetHeartbeatTimer(id)
+		h.logger.Debug("reset node TTL because a newer status update was found", "node_id", id)
 		return
 	}
 
-	h.logger.Warn("node TTL expired", "node_id", id)
+	h.logger.Warn("setting node status down due to expired TTL", "node_id", id, "fencepost", fencepost, "latest_update", latestUpdate)
```
tgross, line 178: A nitpick: we might not want to call it a "fencepost" in the logs. It's obvious to us reading the code but if you're an operator what does that mean? Maybe something like "missed_at"?

tgross (approved):
LGTM

nomad/heartbeat.go
```diff
@@ -142,15 +145,37 @@ func (h *nodeHeartbeater) invalidateHeartbeat(id string) {
 	}
 	h.heartbeatTimersLock.Unlock()
 
-	// Do not invalidate the node since we are not the leader. This check avoids
+	// Do not invalidate the node if we are no longer the leader. This check avoids
 	// the race in which leadership is lost but a timer is created on this
 	// server since it was servicing an RPC during a leadership loss.
-	if !h.IsLeader() {
-		h.logger.Debug("ignoring node TTL since this server is not the leader", "node_id", id)
+	if future := h.raft.VerifyLeader(); future.Error() != nil {
```
schmichael, line 151: An alternative approach would be refactoring Node.UpdateStatus's implementation into a new method that *skipped forwarding to the leader* and call that instead. If this server has lost leadership it will return [`raft.ErrNotLeader`](https://pkg.go.dev/github.com/hashicorp/raft#pkg-variables) and fail to apply.

The imprecise `IsLeader` check would be a fine way to skip some work in that case. While `VerifyLeader` is _more_ precise, it's impossible to guarantee leadership hasn't been lost between checking it and calling Node.UpdateStatus anyway.

PR #14483, opened 2022-09-07
Add client scheduling eligibility to heartbeat

### LINKED ISSUE #8965: Metric label `node_scheduling_availability` does not update with changes.
### Nomad version

Nomad v0.12.5

### Issue

The `node_scheduling_eligibility` label on the Nomad client metrics does not update as the eligibility status is changed.

### Reproduction steps

Create a telemetry.nomad file with the basic configuration

```bash
telemetry {
  publish_allocation_metrics = true
  publish_node_metrics       = true
}
```

Run a dev agent with the telemetry configuration

```bash
nomad agent -dev -config=telemetry.nomad
```

Watch a metric with the label

```bash
watch "curl -s http://localhost:4646/v1/metrics | jq '.Gauges[] | select(.Name==\"nomad.client.uptime\")'"
```

Note that it appears enabled.  Set the eligibility to disabled

```bash
nomad node eligibility -self -disable
```

Watch for a stat update, observe that enabled doesn't change.

angrycub:
Adds the client's scheduling eligibility to the NodeUpdateResponse so that the client is made aware of the server-side value. Takes a heartbeat to get the data to the client and one metrics window to shed the value with the former label, but it appears to be a net improvement over the current behavior.

Fixes #8965

### DIFF
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -1361,7 +1361,17 @@ type NodeUpdateResponse struct {
 	// region.
 	Servers []*NodeServerInfo
 
+	// ClientStatus is used to inform clients what the server-side
+	// has for their scheduling status during heartbeats.
+	ClientStatus *ClientStatus
+
 	QueryMeta
+}
+
+// ClientStatus is used to inform clients what the server-side
+// has for their scheduling status during heartbeats.
+type ClientStatus struct {
+	SchedulingEligibility string
 }
 
 // NodeDrainUpdateResponse is used to respond to a node drain update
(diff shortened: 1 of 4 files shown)


nomad/structs/structs.go
```diff
@@ -1361,9 +1361,19 @@ type NodeUpdateResponse struct {
 	// region.
 	Servers []*NodeServerInfo
 
+	// ClientStatus is used to inform clients what the server-side
+	// has for their scheduling status during heartbeats.
+	ClientStatus *ClientStatus
```
schmichael, line 1366: `SchedulingEligibility string` would be fine here. Would just need to check for `SchedulingEligibility != ""` instead of `ClientStatus != nil`.

schmichael (approved):

PR #14491, opened 2022-09-08
e2e: fix-up token expiration test to account for longer forced GC.

jrasell:
This has been hard to prove and test, but my current theory is that the force GC is taking longer to run than the test waits before checking that the token has been deleted. This is likely in the e2e run as the ACL suite runs towards the end of the full suite and there are likely more than just the expired token to collect.


DerekStrickland (approved):
:shipit:

shoenig (approved):
LGTM!

Issue #2012, opened 2016-11-18
nomad plan for periodic tasks


holtwilkins:
Nomad version 0.5.0.

For a periodic job, `nomad plan` seems to always exit with exitcode 1, indicating pending changes, even when there are no changes pending.

This output is when the job doesn't exist, this is expected

```
[centos@ip-10-2-211-31 ~]$ nomad plan hello-periodic.hcl
+ Job: "hello-periodic"
+ Task Group: "hello-periodic" (1 create)
  + Task: "hello-periodic" (forces create)

Scheduler dry-run:
- All tasks successfully allocated.
- If submitted now, next periodic launch would be at 11/18/16 23:10:00 UTC (2m22s from now).

Job Modify Index: 0
To submit the job with version verification run:

nomad run -check-index 0 hello-periodic.hcl

When running the job with the check-index flag, the job will only be run if the
server side version matches the job modify index returned. If the index has
changed, another user has modified the job and the plan's results are
potentially invalid.
[centos@ip-10-2-211-31 ~]$ echo $?
1
```

However, I expect that once the job is running, that subsequent plans are noops, but that's not what happens.

```
[centos@ip-10-2-211-31 ~]$ nomad run hello-periodic.hcl
Job registration successful
Approximate next launch time: 11/18/16 23:10:00 UTC (1m2s from now)
[centos@ip-10-2-211-31 ~]$ nomad plan hello-periodic.hcl
+/- Job: "hello-periodic"
Datacenters {
  Datacenters: "us-east-1"
}
Task Group: "hello-periodic" (1 create)
  Task: "hello-periodic"

Scheduler dry-run:
- All tasks successfully allocated.
- If submitted now, next periodic launch would be at 11/18/16 23:10:00 UTC (57s from now).

Job Modify Index: 13593
To submit the job with version verification run:

nomad run -check-index 13593 hello-periodic.hcl

When running the job with the check-index flag, the job will only be run if the
server side version matches the job modify index returned. If the index has
changed, another user has modified the job and the plan's results are
potentially invalid.
[centos@ip-10-2-211-31 ~]$ echo $?
1
```

EDIT: Including the jobfile as an attachment, but I renamed it to `hello-periodic.txt` to make github happy.
[hello-periodic.txt](https://github.com/hashicorp/nomad/files/601088/hello-periodic.txt)

holtwilkins:
[...]

PR #14492, opened 2022-09-08
Check that JobPlanResponse Diff Type is None before checking for changes on getExitCode

gabivlj:
Hi everyone! 

* This PR is strongly related to https://github.com/hashicorp/nomad/issues/2012, where if you:
`nomad plan ...`
and there are new allocations it should give an exit status code of 1.
* However if you do:
`nomad run ...`
and then:
`nomad plan ...`
It still returns an exit status code of 1.

These changes prevent that from happening, as if there wasn't any edits to the deployment file `resp.Diff.Type` is 'None', and it gives an exit status code of 0.
I'm wondering if this is the right solution for this problem or we are searching for a better one.

### DIFF
--- a/.changelog/14492.txt
+++ b/.changelog/14492.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cli: Fixed a bug where plans for periodic jobs would return exit code 1 when the job was already register
+```
--- a/command/job_plan.go
+++ b/command/job_plan.go
@@ -391,6 +391,10 @@ type namespaceIdPair struct {
 // * 0: No allocations created or destroyed.
 // * 1: Allocations created or destroyed.
 func getExitCode(resp *api.JobPlanResponse) int {
+	if resp.Diff.Type == "None" {
+		return 0
+	}
+
 	// Check for changes
 	for _, d := range resp.Annotations.DesiredTGUpdates {
 		if d.Stop+d.Place+d.Migrate+d.DestructiveUpdate+d.Canary > 0 {


th0m:
@tgross unless we have missed something (which we might have :grin:) this seems to be a straightforward way of fixing #2012

th0m:
@jrasell do you think you could have a look at this one? It should be relatively quick to review. Thank you

tgross (changes requested):
LGTM, thanks @gabivlj!


<details><summary>test</summary>

```
$ nomad job plan ./jobs/periodic.nomad
+ Job: "example"
+ Task Group: "group" (1 create)
  + Task: "task" (forces create)

Scheduler dry-run:
- All tasks successfully allocated.
- If submitted now, next periodic launch would be at 2022-10-03T21:30:00-07:00 (8h11m20s from now).

Job Modify Index: 0

nomad job run -check-index 0 ./jobs/periodic.nomad

potentially invalid.

$ echo $?
1

$ nomad job run ./jobs/periodic.nomad
Job registration successful

Approximate next launch time: 2022-10-03T21:30:00-07:00 (8h11m15s from now)

$ nomad job plan ./jobs/periodic.nomad
Job: "example"
Task Group: "group" (1 create)
  Task: "task"

Scheduler dry-run:
- All tasks successfully allocated.
- If submitted now, next periodic launch would be at 2022-10-03T21:30:00-07:00 (8h11m12s from now).

Job Modify Index: 11

nomad job run -check-index 11 ./jobs/periodic.nomad

potentially invalid.

$ echo $?
0
```

</details>

This just missed the merge window for Nomad 1.4.0, but I'll make sure it gets merged once that goes out and it'll land in the first regular release of Nomad 1.4.x.

In the meantime, if you want you can add a changelog file to this PR. It'll be at `.changelog/14492.txt` and have the format you can see in [.changelog/14749.txt](https://github.com/hashicorp/nomad/blob/main/.changelog/14749.txt). For content, probably something like:

```
cli: Fixed a bug where plans for periodic jobs would return exit code 1 when the job was already register
```

(If you don't get to it before I'm ready to merge this, no worries. I see the PR has been open a little bit and it'll only take me a few minutes to add it.)

tgross (approved):

tgross:
Hi @gabivlj! We had to push out a quick Nomad 1.4.1 patch, but this has now been merged and will go out in the next regular patch release! Thanks again!

PR #14495, opened 2022-09-08
[ui] Service health checks unique by allocation

philrenaud:
- Services/Service Fragments exist independent from a given allocation
- but service health checks exist within the context of an allocation
- and we still want some way to show all health checks for a given service

So, we bundle all health checks across all allocs into service.healthChecks, add an `Alloc` ID, and filter where we want to show that.

This gives us flexibility to show general service health in the future, plus a historical record of health within the confines of a given alloc.

### DIFF
--- a/ui/app/serializers/service-fragment.js
+++ b/ui/app/serializers/service-fragment.js
@@ -3,9 +3,5 @@ import classic from 'ember-classic-decorator';
 
 @classic
 export default class ServiceFragmentSerializer extends ApplicationSerializer {
-  attrs = {
-    connect: 'Connect',
-  };
-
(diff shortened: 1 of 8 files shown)


ui/app/serializers/service-fragment.js
```diff
@@ -3,9 +3,5 @@ import classic from 'ember-classic-decorator';
 
 @classic
 export default class ServiceFragmentSerializer extends ApplicationSerializer {
-  attrs = {
```
philrenaud, line 6: Side-effect; we're no longer worrying about the connect flag on service fragments.

ChaiWithJai (approved):

PR #14497, opened 2022-09-08
helper: guard against negative inputs into random stagger

shoenig:
This PR modifies `RandomStagger` to protect against negative input
values. If the given interval is negative, the value returned will
be somewhere in the stratosphere. Instead, treat negative inputs
like zero, returning zero.

Before:
```
=== RUN   TestCluster_RandomStagger
    interface.go:24: 
        cluster_test.go:27: expected 2409629h13m43.082153551s <= 1s
--- FAIL: TestCluster_RandomStagger (0.00s)
```

### DIFF
--- a/helper/cluster_test.go
+++ b/helper/cluster_test.go
@@ -1,0 +1,29 @@
+package helper
+
+import (
+	"testing"
+	"time"
+
+	"github.com/shoenig/test/must"
+)
+
+func TestCluster_RandomStagger(t *testing.T) {
+	cases := []struct {
+		name  string
+		input time.Duration
+	}{
+		{name: "positive", input: 1 * time.Second},
+		{name: "negative", input: -1 * time.Second},
+		{name: "zero", input: 0},
+	}
+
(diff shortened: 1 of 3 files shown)


helper/cluster_test.go
```diff
@@ -0,0 +1,29 @@
...
+	}
+
+	abs := func(d time.Duration) time.Duration {
+		return Max(d, -d)
+	}
+
+	for _, tc := range cases {
+		result := RandomStagger(tc.input)
+		must.GreaterEq(t, result, 0)
+		must.LessEq(t, result, abs(tc.input))
```
shoenig, line 27: Just now realizing I got these parameters backwards, https://github.com/shoenig/test/issues/33

tgross: fwiw, having the parameters in the same "direction" as the operator feels more intuitive to me. ex. `must.GreaterEq(t, a, b)` reads like it should be checking `a > b`. Whereas for something like `must.Eq` there isn't an obvious expectation vs result in the way the parameters are named anyways.

PR #14500, opened 2022-09-08
Backport of helper: guard against negative inputs into random stagger into release/1.2.x

hc-github-team-nomad-core:
## Backport

This PR is auto-generated from #14497 to be assessed for backporting due to the inclusion of the label backport/1.2.x.

---

This PR modifies `RandomStagger` to protect against negative input
values. If the given interval is negative, the value returned will
be somewhere in the stratosphere. Instead, treat negative inputs
like zero, returning zero.

Before:
```
=== RUN   TestCluster_RandomStagger
    interface.go:24: 
        cluster_test.go:27: expected 2409629h13m43.082153551s <= 1s
--- FAIL: TestCluster_RandomStagger (0.00s)
```


shoenig:
In 1.2.X this function is defined in Consul

PR #14513, opened 2022-09-08
docs: keyring HTTP API documentation

tgross:
Preview: https://nomad-8bi9ilj87-hashicorp.vercel.app/api-docs/operator/keyring

PR #14514, opened 2022-09-08
remove root keyring install API

tgross:
While working on https://github.com/hashicorp/nomad/pull/14513 and running tests with the API, I realized that our root key install API doesn't actually work anymore following the KMS wrapping we're doing in https://github.com/hashicorp/nomad/pull/14388. The outer wrapper isn't supposed to be shared between servers, and the user has no way to decrypt the key to install the raw key material (without writing their own unwrapper).

Because Nomad servers automatically load the key material in the keystore at startup and replicate the keys to each other, it's not clear this API and associated commands is actually needed or even a good idea to expose to operators. Remove it for now so that we're not committing to having it in Nomad 1.4.0. We can always come back to add it back later.

(Also fixes a minor bug where the rotate API would work with GET)


jrasell (approved):
LGTM

PR #14517, opened 2022-09-09
e2e: fixup service discovery and ACL expiration tests.

jrasell:
The NSD checks tests were racey, whereby the check may not have been triggered by the time it was queried. This change wraps the check so it can account for this.

This removes the current ACL expiration GC section in order to get the tests passing and allow more time to investigate the test. I have full confidence the feature is working as expected and have tested extensively locally.


DerekStrickland (approved):
:shipit:

PR #14519, opened 2022-09-09
job_endpoint: check spec for all regions

DerekStrickland:
This PR introduces a helper method to the Job RPC endpoint called specChanged. This method checks whether an incoming jobspec has changed relative to the currently deployed jobspec in all regions. This handles the case where a job has been stopped in one or more regions, and the operator now wants to restart it. By checking all regions, the spec change is detected, and a new job version can be computed and deployed to all multiregion targets.

### DIFF
--- a/nomad/job_endpoint.go
+++ b/nomad/job_endpoint.go
@@ -349,7 +349,12 @@ func (j *Job) Register(args *structs.JobRegisterRequest, reply *structs.JobRegis
 	}
 
 	// Check if the job has changed at all
-	if existingJob == nil || existingJob.SpecChanged(args.Job) {
+	specChanged, err := j.specChanged(existingJob, args)
+	if err != nil {
+		return err
+	}
+
+	if existingJob == nil || specChanged {
 
 		// COMPAT(1.1.0): Remove the ServerMeetMinimumVersion check to always set args.Eval
 		// 0.12.1 introduced atomic eval job registration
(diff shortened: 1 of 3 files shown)


nomad/job_endpoint.go
```diff
@@ -433,6 +438,53 @@ func (j *Job) Register(args *structs.JobRegisterRequest, reply *structs.JobRegis
 	return nil
 }
 
+// specChanged checks to see if the job spec has changed. If the job is multiregion,
+// it checks all regions. This handles the case of a multiregion job that has been
+// stopped in one or more regions being restarted via the run command.
+func (j *Job) specChanged(existingJob *structs.Job, args *structs.JobRegisterRequest) (bool, error) {
```
tgross, line 444: This method covers much more behavior than its internal call to `Job.SpecChanged`, and all of the behavior changes are specific to multiregion deployments:
* The method name is misleading.
* This should probably be a stub call in OSS with all the real logic only in ENT (which is the only repo it can be tested in anyways).

nomad/job_endpoint.go
```diff
@@ -433,6 +438,53 @@ func (j *Job) Register(args *structs.JobRegisterRequest, reply *structs.JobRegis
 	return nil
 }
 
+// specChanged checks to see if the job spec has changed. If the job is multiregion,
+// it checks all regions. This handles the case of a multiregion job that has been
+// stopped in one or more regions being restarted via the run command.
```
tgross, line 443: The doc strings here says it's "handling" this case but it's a "what not why" comment: we should be explaining why checking that the spec has changed is meaningful and why we need to specially handle this case at all.

nomad/job_endpoint.go
```diff
@@ -433,6 +438,53 @@ func (j *Job) Register(args *structs.JobRegisterRequest, reply *structs.JobRegis
...
+func (j *Job) specChanged(existingJob *structs.Job, args *structs.JobRegisterRequest) (bool, error) {
+	if existingJob == nil {
+		return false, nil
+	}
+
+	if existingJob.SpecChanged(args.Job) {
+		return true, nil
+	}
```
tgross, line 451: This could be pulled out to the caller or the OSS stub function could consist of just this block, leaving it clear that there are two different things going on here: we're checking whether the spec changed _and_ we're checking other regions.

nomad/job_endpoint.go
```diff
@@ -433,6 +438,53 @@ func (j *Job) Register(args *structs.JobRegisterRequest, reply *structs.JobRegis
...
+
+	if !existingJob.IsMultiregion() {
+		return false, nil
+	}
+
+	// If the spec hasn't changed, but this is a multiregion job, check other regions.
+	for _, region := range existingJob.Multiregion.Regions {
+		// Copy the job so we can mutate it by region and compare it with the response.
+		incomingJob := args.Job.Copy()
```
tgross, line 460: I think you can move this copy above the loop and then mutate the copy for each loop iteration.

nomad/job_endpoint.go
```diff
@@ -433,6 +438,53 @@ func (j *Job) Register(args *structs.JobRegisterRequest, reply *structs.JobRegis
...
+
+	if existingJob.SpecChanged(args.Job) {
+		return true, nil
+	}
+
+	if !existingJob.IsMultiregion() {
+		return false, nil
+	}
+
+	// If the spec hasn't changed, but this is a multiregion job, check other regions.
```
tgross, line 457:
```suggestion
	// If the local region's spec hasn't changed, but this is a multiregion job, check other regions.
```

nomad/job_endpoint.go
```diff
@@ -433,6 +438,53 @@ func (j *Job) Register(args *structs.JobRegisterRequest, reply *structs.JobRegis
...
+		req := &structs.JobSpecificRequest{
+			JobID: incomingJob.ID,
+			QueryOptions: structs.QueryOptions{
+				Region:    region.Name,
+				Namespace: incomingJob.Namespace,
+				AuthToken: args.AuthToken,
+			},
+		}
+		resp := structs.SingleJobResponse{}
+		j.logger.Debug("checking spec for multiregion job")
```
tgross, line 471: If we're going to log here at all we should include context like which job, namespace, and region we're querying. Same for the error log below.

nomad/job_endpoint.go
```diff
@@ -433,6 +438,53 @@ func (j *Job) Register(args *structs.JobRegisterRequest, reply *structs.JobRegis
...
+		err := j.GetJob(req, &resp)
+		if err != nil {
+			j.logger.Error("job lookup failed", "error", err)
+			return false, err
+		}
+
+		// If the jobspec changed or is nil (purged) for that region, assume it's being re-registered
+		// and treat like a spec change.
+		if resp.Job.SpecChanged(incomingJob) || resp.Job == nil {
```
tgross, line 480: `SpecChanged` is safe if the receiver is nil but it's still better to flip this conditional so that the next person who reads this code doesn't have to _check_ that it is. 😁

tgross (changes requested):
I think the overall approach here is sound @DerekStrickland

nomad/job_endpoint.go
DerekStrickland: Good point. On it.

tgross (approved):
LGTM!

Issue #13961, opened 2022-08-02
build: use make target for ec2info


shoenig:
Everytime I go to run `ec2info` I forget how. Add a make target like the one for `make missing` that uses `-modfile` to just do the right thing. Before that works though we need to fixup the default CLI arguments to assume being run from the nomad directory. 

Also while we're in here add better error messages for when you try to run without the required AWS environment variables.

PR #14520, opened 2022-09-09
build: make ec2info command usable from GNUMakefile

shoenig:
The ec2info was never intuitive to run - needing to set the AWS environment variables, cd'ing into `tools/ec2info`, and knowing to invoke the command.

This PR makes it so we can run `ec2info` just by running

```shell
make ec2info
```

The command now also checks for the AWS environment variables being set, and provides a useful error if they are not.

Note to reviewers: there's no new code here, just moving everything into `main.go` so it can be conveniently invoked via `go run` without globbing paths, etc. 

Closes https://github.com/hashicorp/nomad/issues/13961


tools/ec2info/main.go
```diff
@@ -59,3 +57,167 @@ func main() {
...
+func environment() error {
+	if value := os.Getenv("AWS_ACCESS_KEY_ID"); value == "" {
+		return errors.New("AWS_ACCESS_KEY_ID must be set")
+	}
+	if value := os.Getenv("AWS_SECRET_ACCESS_KEY"); value == "" {
+		return errors.New("AWS_SECRET_ACCESS_KEY must be set")
+	}
+	if value := os.Getenv("AWS_SESSION_TOKEN"); value == "" {
+		return errors.New("AWS_SESSION_TOKEN must be set")
+	}
```
tgross, line 70: If you're using doormat to set your credentials these values won't be set in the environment, you'll instead have short-lived credentials in `~/.aws/credentials`. A better way to do this might be to capture the error from `NewSession` and make a suggestion that the AWS credentials need to be setup (either via env var or doormat)

tgross (changes requested):

tools/ec2info/main.go
shoenig: Good catch, fixed!

```
➜ make ec2info 
==> Generating AWS EC2 specifications ...
2022/09/09 08:47:57 failed to create AWS session; make sure environment is setup
2022/09/09 08:47:57 must have environment variables AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN
2022/09/09 08:47:57 or ~/.aws/credentials configured properly
2022/09/09 08:47:57 NoCredentialProviders: no valid providers in chain. Deprecated.
	For verbose messaging see aws.Config.CredentialsChainVerboseErrors
exit status 1
```

Issue #12261, opened 2022-03-10
Increase in "Vault: server failed to derive vault token: Allocation "xxx" does not exist" errors after 1.06 to 1.1.10 upgrade


dpn:
### Nomad version
Output from `nomad version`

Previous: `Nomad v1.0.6 (592cd4565bf726408a03482da2c9fd8a3a1015cf)`
Current: `Nomad v1.1.10 (028cef251052e629db4321abc5dfd415f6045b78)`

### Operating system and Environment details

RHEL 7
Control plane (per region): 3x 32 Core, 128G RAM
Clients: On the order of hundreds of clients per region, c1_m2-sized hosts all the way up to c32_128. Mix of EC2 and colocated hardware.

### Issue

As we've been upgrading our clusters from `1.0.6` to `1.1.10`, we've been seeing a significant increase in the number of allocations failing to start with the following error:

```
ID                  = 0d4aaf0a-2c78-74cc-a39f-ead09ca53fdb
Eval ID             = 5681f411
Name                = a-nomad-job/dispatch-1646951139-93b8a50a.a-job[0]
Node ID             = 29ff4074
Node Name           = a-nomad-node.somewhere.someplace
Job ID              = a-nomad-job/dispatch-1646951139-93b8a50a
Job Version         = 19
Client Status       = failed
Client Description  = Failed tasks
Desired Status      = stop
Desired Description = alloc was rescheduled because it failed
Created             = 45s ago
Modified            = 22s ago

Task "a-nomad-job-task" is "dead"
Task Resources
CPU       Memory  Disk     Addresses
7200 MHz  16 GiB  300 MiB

Task Events:
Started At     = N/A
Finished At    = 2022-03-10T22:27:45Z
Total Restarts = 0
Last Restart   = N/A

Recent Events:
Time                       Type        Description
2022-03-10T14:27:48-08:00  Template    Missing: vault.read(super/secret/secret), vault.read(a/different/super/secret/secret)
2022-03-10T14:27:45-08:00  Killing     Vault: server failed to derive vault token: Allocation "0d4aaf0a-2c78-74cc-a39f-ead09ca53fdb" does not exist
2022-03-10T14:27:45-08:00  Task Setup  Building Task Directory
2022-03-10T14:27:45-08:00  Received    Task received by client
```

[...]

PR #14521, opened 2022-09-09
Respect `MinQueryIndex` in `QueryOptions` on getting state snapshot

hntrmrrs:
In an effort to fix the problems leading to https://github.com/hashicorp/nomad/issues/12261#issuecomment-1235367464 this uses `SnapshotMinIndex` instead of `Snapshot` with a minimum index from the `QueryOptions` provided. Is this approach reasonable, or is my understanding of the state snapshotting way off base?

I have had trouble reliably reproducing the behaviour described in #12261 in a test environment.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -2706,11 +2706,15 @@ func (c *Client) deriveSIToken(alloc *structs.Allocation, taskNames []string) (m
 	}
 
 	req := &structs.DeriveSITokenRequest{
-		NodeID:       c.NodeID(),
-		SecretID:     c.secretNodeID(),
-		AllocID:      alloc.ID,
-		Tasks:        tasks,
-		QueryOptions: structs.QueryOptions{Region: c.Region()},
+		NodeID:   c.NodeID(),
+		SecretID: c.secretNodeID(),
+		AllocID:  alloc.ID,
+		Tasks:    tasks,
+		QueryOptions: structs.QueryOptions{
+			Region:        c.Region(),
+			AllowStale:    false,
+			MinQueryIndex: alloc.CreateIndex,
+		},
 	}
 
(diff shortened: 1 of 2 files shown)


tgross (commented):
Hi @hntrmrrs! Sorry about the delay in reviewing this.

There's one small gap here that means this isn't going to help out. When RPCs are sent from the client to the server, they're forwarded to the leader unless the client specifically asks for a stale result. The leader's index is by definition going to be current, because the leader's index is the basis by which raft says the value is current! 

The only time you see us checking for a minimum snapshot index on the leader is in the plan applier (ref [`plan_apply.go#L102-L110`](https://github.com/hashicorp/nomad/blob/main/nomad/plan_apply.go#L102-L110)) and that's because it's running a concurrent operation _outside_ of a raft transaction and that can get out of sync with raft.

In the case of the `Node.DeriveVaultToken` and `Node.DeriveSIToken`, the RPCs are not marked as stale. The client can't possibly have an allocation more recent than the leader. But the client _can_ have a stale allocation because the `Alloc.GetAlloc` request it's sending has the allow-stale flag set (ref [`client.go#L2084-L2092`](https://github.com/hashicorp/nomad/blob/v1.3.5/client/client.go#L2084-L2092).

So I don't think is going to help here. I'm going to leave this PR open for further discussion though and I'll also reopen https://github.com/hashicorp/nomad/issues/12261

tgross:
Doing a little bit of cleanup and I'm going to close this PR out. The problem in #12261 is still open but we're fairly confident this isn't the right approach. I'll be revisiting #12261 as part of upcoming work for Nomad 1.7.0 to refresh our Vault integration. See https://github.com/hashicorp/nomad/issues/15617

PR #14528, opened 2022-09-09
CI: make `make check` clean on macOS

tgross:
Running `make check` on macOS identifies some dead code because the code is used only with the Linux build tag. Move this code into appropriately-tagged code files.


shoenig (approved):
LGTM!

PR #14529, opened 2022-09-09
On window blur, immediately hide keyboard hints

### LINKED ISSUE #14526: [ui] Keyboard nav: disable hints on window defocus
If you're trying to cmd+shift+4 on a mac for a screenshot, for example, keyboard hints persist. Disable them when you're not focused on the browser/window in question.

philrenaud:
Resolves #14526

Solves an issue where you can have shift held and click away and hints are still shown.

This will **not** fix this for when you're taking screenshots, as cmd+shift+5 (and etc.) doesn't fire any window-or-document-detectable events, but it's still an improvement.

### DIFF
--- a/ui/app/components/keyboard-shortcuts-modal.js
+++ b/ui/app/components/keyboard-shortcuts-modal.js
@@ -8,6 +8,13 @@ export default class KeyboardShortcutsModalComponent extends Component {
   @service keyboard;
   @service config;
 
+  constructor() {
+    super(...arguments);
+    window.addEventListener('blur', (a, b, c) => {
+      this.keyboard.displayHints = false;
+    });
+  }
+
   escapeCommand = {
     label: 'Hide Keyboard Shortcuts',
     pattern: ['Escape'],
--- a/ui/app/components/keyboard-shortcuts-modal.js
+++ b/ui/app/components/keyboard-shortcuts-modal.js
@@ -1,3 +1,4 @@
+import { set } from '@ember/object';
 import Component from '@glimmer/component';
 import { inject as service } from '@ember/service';
 import { computed } from '@ember/object';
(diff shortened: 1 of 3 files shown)


ui/app/templates/csi/volumes/index.hbs
```diff
@@ -38,7 +38,7 @@
         @source={{p.list}}
         @sortProperty={{this.sortProperty}}
         @sortDescending={{this.sortDescending}}
-        @class="with-foot {{if this.keyboard.displayHints "display-hints"}}" as |t|
+        @class="with-foot" as |t|
```
philrenaud, line 41: Side-effect: .display-hints was not being used for anything anymore.

ui/app/components/keyboard-shortcuts-modal.js
```diff
@@ -8,6 +9,13 @@ export default class KeyboardShortcutsModalComponent extends Component {
   @service keyboard;
   @service config;
 
+  constructor() {
+    super(...arguments);
+    window.addEventListener('blur', () => {
+      set(this, 'keyboard.displayHints', false);
```
ChaiWithJai, line 15: `question`: Should this get cleaned up after the component unmounts?

ChaiWithJai (approved):
I think this technically has a memory leak because we don't remove the event listener when the component unmounts. This might probably be better off as an element modifier. But its no big deal and probably not worth the effort.

philrenaud:
> I think this technically has a memory leak

@ChaiWithJai How might you fix it, if it were worth doing? This is an app-level always-on component.

ui/app/components/keyboard-shortcuts-modal.js
philrenaud: I don't think so. displayHints = false is the default / at-rest state.

PR #14538, opened 2022-09-10
2 small data race fixes in logmon and check tests

schmichael:
Trying to make some progress on #14236

While the logmon race is in runtime code, I don't think it could account for any logging related issues I see. I think the worst it could do is cause purging to happen too often or not often enough, but unless the *other* logic is wrong in purgeOldFiles that shouldn't cause logging to fail. Just use too much disk space or not as much as it could.

The WithLock helper really shouldn't be used outside of tests where we often "peek into" the state of structs that require locking semantics. I could move it into `testutil` if you think that's a better idea.

### DIFF
--- a/client/logmon/logging/rotator.go
+++ b/client/logmon/logging/rotator.go
@@ -38,10 +38,13 @@ type FileRotator struct {
 	MaxFiles int   // MaxFiles is the maximum number of rotated files allowed in a path
 	FileSize int64 // FileSize is the size a rotated file is allowed to grow
 
-	path             string // path is the path on the file system where the rotated set of files are opened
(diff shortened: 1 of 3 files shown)


client/logmon/logging/rotator.go
```diff
@@ -54,7 +54,7 @@ type FileRotator struct {
 	doneCh      chan struct{}
 
 	closed     bool
-	closedLock sync.Mutex
+	closedLock sync.Mutex // also used to guard oldestLogFileIdx
```
shoenig, line 57: Might be worth giving the lock a general name and moving the locked values under it

shoenig (approved):
LGTM

PR #14546, opened 2022-09-12
client: refactor check watcher to be reusable

shoenig:
This PR refactors `agent/consul/check_watcher` under `client/serviceregistration/`,
and abstracts away the Consul-specific API usage.

In doing so we should be able to reuse the existing check watcher implementation for
also watching NSD checks in a followup PR.

A chunk of `consul/unit_test.go` is removed - we'll cover that in e2e tests
in a follow PR if needed. In the long run I'd like to remove this whole file.
https://github.com/hashicorp/nomad/issues/14548


command/agent/consul/service_client.go
```diff
@@ -419,13 +419,43 @@ type ServiceClient struct {
 	deregisterProbationExpiry time.Time
 
 	// checkWatcher restarts checks that are unhealthy.
-	checkWatcher *checkWatcher
+	// checkWatcher *checkWatcher
```
jrasell, line 422: can we remove this comment?

jrasell (approved):
LGTM!

Issue #14244, opened 2022-08-23
restrict variable path names


tgross:
We should ensure that variable paths don't include characters that are going to be problematic in the UI and Consul Template and return validation errors on them. `@` and `.` in particular.

PR #14547, opened 2022-09-12
variables: restrict allowed paths for variables

tgross:
Fixes https://github.com/hashicorp/nomad/issues/14244

Restrict variable paths to RFC3986 URL-safe characters that don't conflict with the use of characters "@" and "." in `template` blocks. This prevents users from writing variables that will require tricky templating syntax or that they simply won't be able to use.

Also restrict the length so that a user can't make queries in the state store unusually expensive (as they are O(k) on the key length).

### DIFF
--- a/nomad/structs/variables.go
+++ b/nomad/structs/variables.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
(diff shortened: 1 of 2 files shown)


angrycub (approved):
LGTM

PR #14550, opened 2022-09-12
[vars] Make `var list` flags match other var commands

angrycub:
Renames `-q` to `-out=terse`. Removed the combination of `-json` and `-q` since it was of uncertain value because the expectation is that json output would be post-processed anyway


tgross (approved):
LGTM!

PR #14551, opened 2022-09-12
Stabilizing percy snapshots with faker

philrenaud:
Adding faker.seed() in relevant tests, just before a snapshot takes place. Looks like it will prevent instability, for example, in our management token name (which has been changing on every percy run, but seems stable (and named Manager Daphnee Altenwerth forevermore)

### DIFF
--- a/ui/mirage/scenarios/default.js
+++ b/ui/mirage/scenarios/default.js
@@ -2,6 +2,7 @@ import config from 'nomad-ui/config/environment';
 import * as topoScenarios from './topo';
 import * as sysbatchScenarios from './sysbatch';
 import { pickOne } from '../utils';
+import faker from 'nomad-ui/mirage/faker';
 
 const withNamespaces = getConfigValue('mirageWithNamespaces', false);
 const withTokens = getConfigValue('mirageWithTokens', true);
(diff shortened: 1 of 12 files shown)


ui/mirage/scenarios/default.js
```diff
@@ -2,6 +2,7 @@ import config from 'nomad-ui/config/environment';
 import * as topoScenarios from './topo';
 import * as sysbatchScenarios from './sysbatch';
 import { pickOne } from '../utils';
+import faker from 'nomad-ui/mirage/faker';
```
ChaiWithJai, line 5: `question`: They have a pinned version of `faker` before the **incident** occurred, right?

philrenaud: ¿ incident ?

ChaiWithJai: https://stackoverflow.com/questions/70597019/what-happened-with-faker-js

philrenaud: Oh, had totally forgotten that nonsense. Yeah, we're pinned at ^4.1.0.

ChaiWithJai (approved):
LGTM

PR #14556, opened 2022-09-13
client: add `OTEL_RESOURCE_ATTRIBUTES` env var.

lgfa29:
Add new task hook to inject a `OTEL_RESOURCE_ATTRIBUTES` environment variable with Nomad attributes into tasks. The attributes set are related to the alloc and specific task that is running, the node where the alloc is running, and the job and eval that generated the alloc.

These attributes are merged if the task already defines a `OTEL_RESOURCE_ATTRIBUTES` environment variable, or disabled if the value defined by the task is an empty string.


client/allocrunner/taskrunner/otel_hook.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		attrs = resourceAttrs.String()
+	}
+	resp.Env = map[string]string{
+		envKeyOtelResourceAttrs: attrs,
+	}
+	return nil
+}
+
+func generateBaggage(alloc *structs.Allocation, task *structs.Task, node *structs.Node) (baggage.Baggage, error) {
```
tgross, line 101: Why should this function ever return errors (which will stop the allocation)? Aren't all the contents of this under Nomad's control?

client/allocrunner/taskrunner/otel_hook.go
```diff
@@ -0,0 +1,144 @@
...
+			for _, m := range taskBaggage.Members() {
+				k, v := m.Key(), m.Value()
+				logger.Trace("found member", "key", k, "value", v)
+
+				// TODO(luiz): don't create new member once baggage.Members()
+				// returns values with `hasData` set to `true`.
+				// https://github.com/open-telemetry/opentelemetry-go/issues/3164
+				member, err := baggage.NewMember(k, v)
+				if err != nil {
+					logger.Warn("failed to create new baggage member", "key", k, "value", v, "error", err)
```
tgross, line 76: The ridiculous name "baggage" aside, which isn't your fault, I'm not sure we should expose that inside-baseball terminology to end-users?

tgross (commented):
The code mostly looks fine here, but I'm not sure about some of the design motivations of this one @lgfa29. There's a hidden backwards compatibility issue here as well, which we've run into when discussing adding Docker labels: if you _are_ running OTEL already by adding the OTEL env var to your tasks, Nomad just ballooned your OTEL provider costs. You can only opt-out of the env var entirely, not out of having Nomad add to it.

client/allocrunner/taskrunner/otel_hook.go
```diff
@@ -0,0 +1,144 @@
...
+		newMember("nomad.job.name", job.Name, mErr),
+		newMember("nomad.job.region", job.Region, mErr),
+		newMember("nomad.job.type", job.Type, mErr),
+		newMember("nomad.namespace", alloc.Namespace, mErr),
+		newMember("nomad.node.id", node.ID, mErr),
+		newMember("nomad.node.name", node.Name, mErr),
+		newMember("nomad.node.datacenter", node.Datacenter, mErr),
+		newMember("nomad.task.name", task.Name, mErr),
+		newMember("nomad.task.driver", task.Driver, mErr),
+	}
```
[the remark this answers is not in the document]
lgfa29: Ah I see. I thought of some kind of configuration (either jobspec or client side), but that would be even more commitment to keep it updated, or `meta` values, but they are not really well defined and should probably remain opaque to Nomad.

Maybe an external project may be better for now then. I will think more about it, thanks!

client/allocrunner/taskrunner/otel_hook.go
```diff
@@ -0,0 +1,144 @@
...
+			}
+		}
+	}
+
+	// TODO(luiz): remove decode step once the Otel SDK handles it internally.
+	// https://github.com/open-telemetry/opentelemetry-go/pull/2963
+	attrs, err := url.QueryUnescape(resourceAttrs.String())
+	if err != nil {
+		attrs = resourceAttrs.String()
+	}
```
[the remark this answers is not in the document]
lgfa29: The problem was a mismatch between how the baggage spec and the collector handled encoding values. Only the baggage required encoding/decoding so the more general "fix" was [an update to the spec](https://github.com/open-telemetry/opentelemetry-specification/pull/2670). Other languages will need to be updated to handle this as well, but good point on supporting older versions. I will make sure to test it.

Attempting to double-encode would result in an error that is handled by the using the original string.

client/allocrunner/taskrunner/otel_hook.go
lgfa29: This function returns the error so it can be logged by the hook. The hook itself never returns an error, so it doesn't impact the allocation. Keeping it a separate function (instead of a method of the hook) helps with testing, but I can make it part of struct as well.

client/allocrunner/taskrunner/otel_hook.go
lgfa29: Ah right, good point 👍

schmichael:
This is a great candidate for Task Hook Plugins... if such a thing existed. The hook API was designed to seamlessly be able to add them someday. :shrug:

PR #14561, opened 2022-09-13
docs: include path in ACL requirements for variables

tgross:
Also add links to the ACL policy reference and variables concepts docs near the top of the page.


angrycub (approved):
LGTM

PR #14573, opened 2022-09-13
Tooltips for client (node) names

### LINKED ISSUE #14421: Add client name tooltip to allocation view
### Proposal
The new client name tooltips when hovering over the client_id in the allocations tables is very helpful. I'm wondering if it would be possible to add that same functionality to the client_id that appears on the allocation view:
image

### Use-cases
When drilled down to a the allocation view, it currently takes extra clicks/context switches to identify what node the allocation is running on.

### Attempted Solutions
We can navigate up a level, or lookup the client name by other means, but again, this requires context switching which can be frustrating when debugging an issue.

philrenaud:
Resolves #14421 

Adds tooltips to a few places we only show node shortIDs

image

### DIFF
--- a/ui/app/components/allocation-service-sidebar.hbs
+++ b/ui/app/components/allocation-service-sidebar.hbs
@@ -77,12 +77,15 @@
 						<span class="term">
 							Client
 						</span>
-						<LinkTo
-							@route="clients.client"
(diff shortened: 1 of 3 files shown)


ChaiWithJai (approved):
Nice

PR #14576, opened 2022-09-13
docs: Correct driver name for 'Nomad Task Group' autoscaler target

krarey:
This PR updates the Nomad Autoscaler documentation to correctly reference the Nomad Task Group target plugin as `nomad-target`. The current example references the plugin as `nomad`, which causes the Autoscaler agent to fail to start:

```
2022-09-13T19:24:07.280Z [ERROR] agent: failed to start agent:
  error=
  | failed to setup plugins: 1 error occurred:
  | 	* failed to dispense plugin nomad: failed to instantiate plugin nomad client: fork/exec /plugins/nomad: no such file or directory
  | 
  ```


jrasell (approved):
LGTM, thanks @krarey!

PR #14577, opened 2022-09-13
cleanup: create interface for check watcher and mock it in nsd tests

shoenig:
This PR refactors `serviceregistration.CheckWatcher` to be an interface, that we then mock in some service registration tests where we can keep a tally of `Watch` and `Unwatch` calls, making sure they meet expectations.


jrasell (approved):
LGTM!

PR #14581, opened 2022-09-13
[release/1.4.0-beta.1] Temporarily turn off binary secret scanning

mdeggies:
Temporarily turn off binary secret scanning until false negatives are resolved


DerekStrickland (approved):
:shipit:

PR #14583, opened 2022-09-14
build: always set osusergo

lgfa29:
The way this tag was previous declared would cause it to be overwritten when the GO_TAGS env var was set.


DerekStrickland (approved):
:shipit:

PR #14587, opened 2022-09-14
E2E/UI: set `NOMAD_PROXY_ADDR` separately from `NOMAD_ADDR`

tgross:
The Nomad UI will be proxied by the proxy job, but the test setup also requires making direct HTTP API requests that bypass the UI. For clusters with mTLS configured (ex. nightly), we need to bind-mount the user's mTLS certs picked up by the API client into the test runner.

Includes a new README that explains the topology of the test environments in detail and provides some advice for running locally.

---

Tested against both an E2E test cluster and local development environment. Once we've merged it I'll need to make a PR over in the [private test infra repo](https://github.com/hashicorp/nomad-e2e) to set the `NOMAD_PROXY_ADDR` value.

README preview link: https://github.com/hashicorp/nomad/blob/d24570eb342884e79535a64c8fb06055139859f1/e2e/ui/README.md

### DIFF
--- a/e2e/ui/api-client.js
+++ b/e2e/ui/api-client.js
@@ -25,9 +25,9 @@ export const client = async (
 ) => {
   const url = `${NOMAD_ADDR}/v1`.concat(path);
 
-  var httpsAgent = new https.Agent({keepAlive: true});
+  let httpsAgent = new https.Agent({keepAlive: true});
 
-  if (NOMAD_CLIENT_CERT != "") {
+  if (NOMAD_CLIENT_CERT !== "") {
     httpsAgent = new https.Agent({
       keepAlive: true,
       cert: fs.readFileSync(NOMAD_CLIENT_CERT),
@@ -47,7 +47,7 @@ export const client = async (
       "X-Nomad-Token": NOMAD_TOKEN,
       ...customHeaders,
     },
-    httpsAgent: httpsAgent,
+    httpsAgent,
     ...customConfig,
   };
 
--- a/e2e/ui/api-client.js
+++ b/e2e/ui/api-client.js
@@ -1,11 +1,21 @@
 import axios from "axios";
 
+export let NOMAD_PROXY_ADDR = process.env.NOMAD_PROXY_ADDR;
 export let NOMAD_ADDR = process.env.NOMAD_ADDR;
 export let NOMAD_TOKEN = process.env.NOMAD_TOKEN;
+export let NOMAD_CLIENT_CERT = process.env.NOMAD_CLIENT_CERT;
+export let NOMAD_CLIENT_KEY = process.env.NOMAD_CLIENT_KEY;
+export let NOMAD_CA_CERT = process.env.NOMAD_CA_CERT;
+export let NOMAD_TLS_SERVER_NAME = process.env.NOMAD_TLS_SERVER_NAME;
 
 if (NOMAD_ADDR == undefined || NOMAD_ADDR == "") {
   NOMAD_ADDR = "http://localhost:4646";
 }
+
+if (NOMAD_PROXY_ADDR == undefined || NOMAD_PROXY_ADDR == "") {
+  NOMAD_PROXY_ADDR = NOMAD_ADDR;
+}
+
 
 export const client = async (
   path,
--- a/e2e/ui/api-client.js
+++ b/e2e/ui/api-client.js
@@ -1,11 +1,13 @@
 import axios from "axios";
+import https from "https";
+import fs from "fs";
 
 export let NOMAD_PROXY_ADDR = process.env.NOMAD_PROXY_ADDR;
 export let NOMAD_ADDR = process.env.NOMAD_ADDR;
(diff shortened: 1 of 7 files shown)


e2e/ui/global-setup.js
```diff
@@ -1,6 +1,6 @@
 import fs from "fs";
 import { chromium } from "@playwright/test";
-import { client, NOMAD_ADDR, NOMAD_TOKEN } from "./api-client.js";
+import { client, NOMAD_PROXY_ADDR, NOMAD_ADDR, NOMAD_TOKEN } from "./api-client.js";
```
ChaiWithJai, line 3: `nit`: But we should be able to remove `NOMAD_ADDR`  from this list of imports as its unused now, right?

tgross: I think so, assuming that `client` is already configured for the HTTP API at that point and doesn't need to be handed addresses?

tgross: Ok, great! That sounds right to me!

e2e/ui/api-client.js
```diff
@@ -1,19 +1,41 @@
...
 export const client = async (
   path,
   { data, method, headers: customHeaders, ...customConfig } = {},
   parameters
 ) => {
   const url = `${NOMAD_ADDR}/v1`.concat(path);
 
+  var httpsAgent = new https.Agent({keepAlive: true});
+
+  if (NOMAD_CLIENT_CERT != "") {
```
ChaiWithJai, line 30: `question`: Do we want loose or strict equality here? As it stands:

```
undefined != "" // true
null != "" // true
0 != "" // false
```

e2e/ui/api-client.js
```diff
@@ -1,19 +1,41 @@
...
+}
+
 export const client = async (
   path,
   { data, method, headers: customHeaders, ...customConfig } = {},
   parameters
 ) => {
   const url = `${NOMAD_ADDR}/v1`.concat(path);
 
+  var httpsAgent = new https.Agent({keepAlive: true});
```
ChaiWithJai, line 28: `nit`: Can we declare this as a `let`?

e2e/ui/api-client.js
```diff
@@ -25,6 +47,7 @@ export const client = async (
       "X-Nomad-Token": NOMAD_TOKEN,
       ...customHeaders,
     },
+    httpsAgent: httpsAgent,
```
ChaiWithJai, line 50: `nit`: this can just be `httpsAgent` instead of `httpsAgent: httpsAgent`

`question`: Do we always want this present? Or are we conditionally passing this as a configuration?

ChaiWithJai (approved):
Some minor improvements could be made, but not necessary.

e2e/ui/api-client.js
tgross: I think so. Will do.

e2e/ui/api-client.js
tgross: Good catch!

e2e/ui/api-client.js
tgross: > question: Do we always want this present? Or are we conditionally passing this as a configuration?

I think we do. We won't have configured the client certs for mTLS if the env vars aren't set appropriately. But we still can take advantage of the config if `NOMAD_ADDR` is https but mTLS isn't configured for some reason (a somewhat oddball local dev environment, I guess).

e2e/ui/api-client.js
```diff
@@ -1,19 +1,41 @@
...
 export const client = async (
   path,
   { data, method, headers: customHeaders, ...customConfig } = {},
   parameters
 ) => {
   const url = `${NOMAD_ADDR}/v1`.concat(path);
 
+  let httpsAgent = new https.Agent({keepAlive: true});
+
+  if (NOMAD_CLIENT_CERT !== "") {
```
ChaiWithJai, line 30: @tgross I did a poor job explaining the comment attached to this last time.

I was prodding that `NOMAD_CLIENT_CERT` could and would be `undefined` and if we want to compare based on both the case of an empty string `""` and `undefined` then we'd want this to read:

`if (!!NOMAD_CLIENT_CERT)`

These are quirks of JavaScript that I wouldn't expect anyone to know, however. I'll get a patch up for this.

Mea culpa. I'll be more direct and clear with PR comments in the future.

tgross: Oh, thanks. I'd tested this out in a console but I think I might have just screwed up my testing because of the "double negative" of `("undefined !== "")`.

ChaiWithJai: I'll do better on pull requests feedbacks and comments in the future.

I follow this [Conventional Comments](https://conventionalcomments.org/) paradigm. However, I'll do better to be more explicit and with explanations... and less "prodding" comments.

Personally, I view PRs as part of the learning process hence prodding for more back and forth. But, I realize its not normal and untenable to have a lot of back and forth on PRs because we're a lean team.

I'll be better next time around.

PR #14598, opened 2022-09-15
e2e: rewrite spread suite to use new e2e style.

jrasell:
The rewrite refactors the suite to use the new style along with other recent testing improvements. In order to ensure the spread tests do not impact each other, there is new cleanup functionality to ensure both the job and allocations are removed from state before the test exits completely.

Whilst I cannot be 100% sure this fixes the failing test, I have run the spread suite 200 times in succession against a full E2E cluster without a single failure. Further failures will indicate a problem with these tests in relation to other suites.


tgross (approved):
LGTM!

PR #14600, opened 2022-09-15
tests: use longer window in testing exec streaming w/ child process

tgross:
Our conformance tests for exec streaming include a test where, when the exec is done with a tty and a child process is forked and backgrounded, we only wait for the main process to exit before dropping the stream. Under heavy load in the test runner (or CPU steal from the CI environment), it's possible that we don't tear down the stream before the background process emits its message and this causes a spurious test failure.

We can safely extend the sleep for the background process by a lot without impacting the test semantics, because the test is only ensuring that we close the stream after the main process exits and not that we exit within a certain timeframe.

Note that this entire test suite is a good candidate for refactoring, but getting CI green is more important at the moment.


tgross:
Last failed CircleCI test run shows:
> --- FAIL: TestExec_ExecTaskStreaming/basic:_tty:_children_processes (10.02s)

Which suggests this is selectively broken in this environment; it works just fine on, say, Vagrant.

tgross:
Closing this as a dead-end

PR #14602, opened 2022-09-15
Release/1.4.0 beta.1

DerekStrickland:


.release/security-scan.hcl
```diff
@@ -13,6 +13,6 @@ binary {
   nvd        = false
 
   secrets {
-    all = true
+    all = false
```
lgfa29, line 16: Oh, we can flip this back to `true`. The false positive problem has a fix that will be available soon as well 🤞

lgfa29 (approved):
Just need to flip that config back 🙂

PR #14604, opened 2022-09-16
e2e: use unique names for Connect ACL Consul policy names.

jrasell:
In the event a single test fails to clear up properly after itself, all other tests will fail as they attempt to create ACL policies with the same names. This change ensures they use unique ACL names, so when a single test fails, it is easy to identify it is a problem with the test rather than the suite.


shoenig (approved):
LGTM!

PR #14606, opened 2022-09-16
refact: convert to boolean primitive not strict eq

ChaiWithJai:


philrenaud (approved):
What's this solving for? Are there cases where "" is a valid cert?

ChaiWithJai:
> What's this solving for? Are there cases where "" is a valid cert?

Hey, buddy! Great question. We're solving for the case where `NOMAD_CLIENT_CERT` is undefined which can happen when the environment variable isn't set.

`!!""` will always resolve to `false`.

Using the Double Logic `NOT` operator (`!!`) will have the following values resolve to `false`:

```
null;
NaN;
0;
empty string ("" or '' or ``);
undefined;
```

PR #14610, opened 2022-09-16
operator debug: write NDJSON for large collections

tgross:
The `operator debug` command writes JSON files from API responses as a single line containing an array of JSON objects. But some of these files can be extremely large (GB's) for large production clusters, which makes it difficult to parse them using typical line-oriented Unix command line tools that can stream their inputs without consuming a lot of memory.

For collections that are typically large, instead emit newline-delimited JSON.

This changeset includes some first-pass refactoring of this command. It breaks up monolithic methods that validate a path, create a file, serialize objects, and write them to disk into smaller functions, some of which can now be standalone to take advantage of generics.

---

There are mostly likely a bunch more improvements we can make here to memory usage, including using pagination so we're not beating on the servers so much (although that makes the CLI more sensitive to server version). But also I didn't want to try to rework the whole darn thing at once 😀

### DIFF
--- a/command/operator_debug.go
+++ b/command/operator_debug.go
@@ -738,7 +738,7 @@ func (c *OperatorDebugCommand) mkdir(paths ...string) error {
 		return fmt.Errorf("file path escapes capture directory")
 	}
 
-	return os.MkdirAll(joinedPath, 0755)
+	return escapingfs.EnsurePath(joinedPath, true)
 }
 
 // startMonitors starts go routines for each node and client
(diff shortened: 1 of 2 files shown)


DerekStrickland (approved):
LGTM.

command/operator_debug.go
```diff
@@ -1293,22 +1298,155 @@ func (c *OperatorDebugCommand) writeBytes(dir, file string, data []byte) error {
...
+	// Replace invalid characters in filename
+	filename := helper.CleanFilename(file, "_")
+
+	relativePath := filepath.Join(dir, filename)
+	c.manifest = append(c.manifest, relativePath)
+	dirPath := filepath.Join(c.collectDir, dir)
+	filePath := filepath.Join(dirPath, filename)
+
+	// Ensure parent directories exist
+	err := os.MkdirAll(dirPath, os.ModePerm)
```
shoenig, line 1315: There is also https://github.com/hashicorp/nomad/blob/main/helper/escapingfs/escapes.go#L116

Which also sets permissions to a more reasonable `0755`; `os.ModePerm` is just a mask of permission bits, resulting in `0777`.

command/operator_debug.go
```diff
@@ -1293,22 +1298,155 @@ func (c *OperatorDebugCommand) writeBytes(dir, file string, data []byte) error {
...
+
+// writeNDJSON writes a slice of Nomad API objects to the archive file as
+// newline-delimited JSON objects.
+func writeNDJSON[T any](data []T, writer io.Writer) error {
+	for _, obj := range data {
+		err := writeJSON(obj, writer)
+		if err != nil {
+			return fmt.Errorf("failed to write to file: %w", err)
+		}
+		_, err = writer.Write([]byte{'\n'})
```
shoenig, line 1441: Does NDJSON say anything about newlines on Windows?

shoenig (changes requested):

command/operator_debug.go
tgross: Fixed in [5423f92](https://github.com/hashicorp/nomad/pull/14610/commits/5423f923b5a30b6f8f0a66a289b32c20b645e3e5), and I've gone thru the rest of the file and fixed the other uses cases of the same thing. 

(I also fixed all the places we were doing `\"%s\"` in error messages while I was in there 😁 )

shoenig (approved):
LGTM!

PR #14612, opened 2022-09-16
[ui] task logs in sidebar

### LINKED ISSUE #14579: [ui] Task logs in sidebar
image

philrenaud:
Resolves #14579

Adds the ability to view logs for a given task from a parent context:
- Job page
- Client page
- Task group page
- Job/allocations page

In other words: from within any context that we display an allocations table.

This adds a portalTarget to the application level to reduce overhead, and hoists the sidebar from the recently-added `<TaskSubRow />` component.

The logs feed is using the same component and layout that its dedicated page uses.

image

### TODO
- [x] Add for non-service jobs (each job part has its own template file)
- [x] Smoke testing in periodic/parameterized (child) jobs
- [x] Acceptance tests
- [x] Visual regression tests


ui/app/components/allocation-service-sidebar.js
```diff
@@ -10,7 +10,7 @@ export default class AllocationServiceSidebarComponent extends Component {
   }
   keyCommands = [
     {
-      label: 'Close Evaluations Sidebar',
+      label: 'Close Service Sidebar',
```
philrenaud, line 13: Side-effect; adding correct name.

ui/app/styles/components/toggle.scss
```diff
@@ -31,7 +31,9 @@ $size: 12px;
   .toggler {
     display: inline-block;
     position: relative;
-    vertical-align: middle;
+    vertical-align: baseline;
+    position: relative;
+    top: 1px;
```
philrenaud, line 36: This affects (positively) a few others instances where we use our <Toggle> element and it was slightly misaligned.

ui/app/templates/components/job-page.hbs
```diff
@@ -19,7 +19,7 @@
         "job-page/parts/latest-deployment" job=@job handleError=this.handleError
       )
       TaskGroups=(component "job-page/parts/task-groups" job=@job)
-      RecentAllocations=(component "job-page/parts/recent-allocations" job=@job)
+      RecentAllocations=(component "job-page/parts/recent-allocations" job=@job activeTask=@activeTask setActiveTaskQueryParam=@setActiveTaskQueryParam)
```
philrenaud, line 22: This, and further job-page/*.hbs file changes, were tricky to figure out how to best handle (and thank you to @DingoEatingFuzz for unblocking me!)

The nature of our higher-order components is such that, to get query parameters at the /job level, we have to declare our passed variables more verbosely than desired.

ui/app/components/task-sub-row.hbs
```diff
@@ -1,16 +1,16 @@
...
-			{{this.task.name}}
-		</LinkTo>
-		{{!-- TODO: in-page logs --}}
-		{{!-- <FlightIcon @name="logs" /> --}}
+		<div class="name-grid">
+			<LinkTo title={{this.task.name}} class="task-name" @route="allocations.allocation.task" @models={{array this.task.allocation this.task}}>{{this.task.name}}</LinkTo>
+			<button type="button" class="logs-sidebar-trigger button is-borderless is-inline is-compact" onclick={{action "handleTaskLogsClick" this.task}}>
+				<FlightIcon @name="logs" />View Logs
+			</button>
+		</div>
```
ChaiWithJai, line 13: `comment`: I think this is going to cause your test to fail. Maybe add a `data-test` attribute to your `LinkTo` component and assert on that selector instead of `.name-grid`.

ui/app/templates/application.hbs
```diff
@@ -26,6 +26,8 @@
 
 <KeyboardShortcutsModal />
 
+<PortalTarget @name="log-sidebar-portal" />
```
ChaiWithJai, line 29: This is pretty high up in the DOM tree. Why not have this be a child of the `PageLayout`?

ui/app/templates/components/job-page/parameterized-child.hbs
```diff
@@ -21,7 +21,7 @@
     <jobPage.ui.Summary @forceCollapsed={{@job.hasClientStatus}} />
     <jobPage.ui.PlacementFailures />
     <jobPage.ui.TaskGroups @sortProperty={{@sortProperty}} @sortDescending={{@sortDescending}} />
-    <jobPage.ui.RecentAllocations />
+    <jobPage.ui.RecentAllocations @activeTask={{@activeTask}} @setActiveTaskQueryParam={{@setActiveTaskQueryParam}} />
```
ChaiWithJai, line 24: Looks like you're passing this prop into each component template and adding it directly into `app/templates/components/job-page.hbs`.

Was there some type of error when working with parent jobs? I think this muddies the component API a bit... but if there is an error arising with parent jobs which prevents us from following the contextual component pattern, then this is fine.

https://guides.emberjs.com/v3.28.0/in-depth-topics/patterns-for-components/#toc_contextual-components

Great [talk](https://youtu.be/MpFudGJn2J0) on component APIs in Ember.

ChaiWithJai (approved):
There's some areas for improvement, but not worth blocking on them.

ui/app/templates/components/job-page/parameterized-child.hbs
philrenaud: I'm familiar with and a fan of the Camba talk. In this case, the nature of these higher-order components makes it so we have to do more verbose declaration than we'd like to. See for example the `<jobPage.ui.TaskGroups` `@properties` just above this line — this behaves in a very similar way.

ui/app/templates/application.hbs
philrenaud: You get `Assertion Failed: Portal target with name log-sidebar-portal already exists` at that level.

mikenomitch (approved):
LGTM!

PR #14615, opened 2022-09-17
Invoke FingerprintManager's Reload() func during agent's SIGHUP (Fixes #14614)

### LINKED ISSUE #14614: Reloading agent does detect CNI plugin changes (due to FingerprintManager.Reload() not being invoked)
### Nomad version
Nomad v1.3.5 (1359c2580fed080295840fb888e28f0855e42d50)

Linux/amd64

### Issue
While trying to integrate cilium with nomad (see: https://github.com/hashicorp/nomad/issues/12120), I've found nomad agent's is not checking changes on CNI during SIGHUP, which means it cannot detect changes applied to CNI's configuration files. However, by inspecting the code, all the needed infrastructure is in place, it just looks like the call to FingerprintManager.Reload() is missing from SIGHUP handling code.

### Reproduction steps
start nomad
make changes to /etc/cni/net.d/*
kill -HUP $NOMAD_PID

#### Expected Result
Reloading of CNI config changes, including errors on logfile in case CNI config can't be processed. At any point, at least the message "reloading fingerprinter" from ./client/fingerprint_manager.go:114 should appear on log file.

#### Actual Result
CNI changes are ignored, and nothing (related to fingerprint manager, nor CNI) appears on logfile.

### Job file (if appropriate)
N/A

pruiz:
This PR addresses https://github.com/hashicorp/nomad/issues/14614, by adding the missing call to Fingerprint's Reload() method from within SIGHUP handling on agent's code.

### DIFF
--- a/.changelog/14615.txt
+++ b/.changelog/14615.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+client: Fixed a bug where network fingerprinters were not reloaded when the client configuration was reloaded with SIGHUP
+```
--- a/client/client.go
+++ b/client/client.go
@@ -297,6 +297,9 @@ type Client struct {
 	endpoints     rpcEndpoints
 	streamingRpcs *structs.StreamingRpcRegistry
 
+	// fingerprintManager is the FingerprintManager registered by the client
+	fingerprintManager *FingerprintManager
+
 	// pluginManagers is the set of PluginManagers registered by the client
 	pluginManagers *pluginmanager.PluginGroup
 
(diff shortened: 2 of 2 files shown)


tgross (changes requested):
LGTM!

This has missed the merge window for Nomad 1.4.0, but I'll merge it once the GA goes out (shortly) and then it'll go out in the next regular release of Nomad. 

@pruiz can you add a changelog file to this PR? It'll be at `.changelog/14615.txt` and have the format you can see in [.changelog/14749.txt](https://github.com/hashicorp/nomad/blob/main/.changelog/14749.txt). For content, probably something like:

```
client: Fixed a bug where fingerprinters were not reloaded when the client configuration was reloaded with SIGHUP
```

tgross (approved):

pruiz:
Added changelog file.

Issue #6480, opened 2019-10-11
Better metric name for nomad.broker.total_blocked


preetapan:
The metric `broker.total_blocked` is poorly named. It is actually counting the number of pending evals for the same job in Nomad's eval broker. 

Proposing renaming this (better name TBD) with a deprecation path. Plan is to log the metric with two names in a point release. The old name will be removed in a subsequent major release with deprecation notes.

PR #14621, opened 2022-09-19
eval broker: shed all but one blocked eval per job after ack

tgross:
When an evaluation is acknowledged by a scheduler, the resulting plan is guaranteed to cover up to the `ModifyIndex` ("wait index") set by the worker based on the most recent evaluation for that job in the state store. At that point, we no longer need to retain blocked evaluations in the broker that are older than that index.

Move these stale evals into a canceled set. When the `Eval.Ack` RPC returns from the eval broker it will retrieve a batch of canceable evals to write to raft. This paces the cancelations limited by how frequently the schedulers are acknowledging evals; this should reduce the risk of cancelations from overwhelming raft relative to scheduler progress. 

Note that the evals will still need to be deleted during garbage collection, but there's not much we can do about that without preventing the evals from being created in the first place.

<details><summary>original approach</summary>

When a node updates its fingerprint or status, we need to create new evaluations to ensure that jobs waiting for resources get a chance to be evaluated. But in the case of a cluster with a large backup of evaluations and flapping nodes, we can get a large backlog of evaluations for the same job. Most of these will be canceled but we still need to write the evaluation in raft and then write its deletion in raft.

This changeset proposes that we avoid creating evals for jobs that already have a blocked eval in the eval broker. A blocked eval means that the broker already has work in-flight *and* work waiting to be re-enqueued, so it's safe to drop the evaluation.

</details>

### DIFF
--- a/nomad/eval_broker.go
+++ b/nomad/eval_broker.go
@@ -147,7 +147,7 @@ func NewEvalBroker(timeout, initialNackDelay, subsequentNackDelay time.Duration,
 		evals:                make(map[string]int),
 		jobEvals:             make(map[structs.NamespacedID]string),
 		blocked:              make(map[structs.NamespacedID]BlockedEvaluations),
-		cancelable:           []*structs.Evaluation{},
+		cancelable:           make([]*structs.Evaluation, 0, structs.MaxUUIDsPerWriteRequest),
 		ready:                make(map[string]PendingEvaluations),
 		unack:                make(map[string]*unackEval),
 		waiting:              make(map[string]chan struct{}),
@@ -760,7 +760,7 @@ func (b *EvalBroker) flush() {
 	b.evals = make(map[string]int)
 	b.jobEvals = make(map[structs.NamespacedID]string)
 	b.blocked = make(map[structs.NamespacedID]BlockedEvaluations)
-	b.cancelable = []*structs.Evaluation{}
+	b.cancelable = make([]*structs.Evaluation, 0, structs.MaxUUIDsPerWriteRequest)
 	b.ready = make(map[string]PendingEvaluations)
 	b.unack = make(map[string]*unackEval)
 	b.timeWait = make(map[string]*time.Timer)
@@ -1023,10 +1023,7 @@ func (p *BlockedEvaluations) MarkForCancel() []*structs.Evaluation {
 	// will want to cancel most of them. Using heap.Remove requires we re-sort
 	// for each eval we remove. Because we expect to have at most one remaining,
 	// we'll just create a new heap.
-	retain := BlockedEvaluations{}
-
-	raw := heap.Pop(p)
-	heap.Push(&retain, raw)
+	retain := BlockedEvaluations{(heap.Pop(p)).(*structs.Evaluation)}
 
 	cancelable := make([]*structs.Evaluation, len(*p))
 	copy(cancelable, *p)
--- a/nomad/eval_broker.go
+++ b/nomad/eval_broker.go
@@ -64,7 +64,11 @@ type EvalBroker struct {
 	jobEvals map[structs.NamespacedID]string
 
 	// blocked tracks the blocked evaluations by JobID in a priority queue
-	blocked map[structs.NamespacedID]PendingEvaluations
+	blocked map[structs.NamespacedID]BlockedEvaluations
+
+	// cancelable tracks previously blocked evaluations (for any job) that are
+	// now safe for the Eval.Ack RPC to cancel in batches
+	cancelable []*structs.Evaluation
 
 	// ready tracks the ready jobs by scheduler in a priority queue
 	ready map[string]PendingEvaluations
@@ -115,10 +119,13 @@ type unackEval struct {
 	NackTimer *time.Timer
 }
 
-// PendingEvaluations is a list of waiting evaluations.
-// We implement the container/heap interface so that this is a
-// priority queue
+// PendingEvaluations is a list of ready evaluations across multiple jobs. We
+// implement the container/heap interface so that this is a priority queue.
 type PendingEvaluations []*structs.Evaluation
+
+// BlockedEvaluations is a list of blocked evaluations for a given job. We
+// implement the container/heap interface so that this is a priority queue.
+type BlockedEvaluations []*structs.Evaluation
 
 // NewEvalBroker creates a new evaluation broker. This is parameterized
 // with the timeout used for messages that are not acknowledged before we
@@ -139,7 +146,8 @@ func NewEvalBroker(timeout, initialNackDelay, subsequentNackDelay time.Duration,
 		stats:                new(BrokerStats),
 		evals:                make(map[string]int),
 		jobEvals:             make(map[structs.NamespacedID]string),
-		blocked:              make(map[structs.NamespacedID]PendingEvaluations),
+		blocked:              make(map[structs.NamespacedID]BlockedEvaluations),
+		cancelable:           []*structs.Evaluation{},
 		ready:                make(map[string]PendingEvaluations),
 		unack:                make(map[string]*unackEval),
 		waiting:              make(map[string]chan struct{}),
@@ -586,15 +594,28 @@ func (b *EvalBroker) Ack(evalID, token string) error {
 
 	// Check if there are any blocked evaluations
 	if blocked := b.blocked[namespacedID]; len(blocked) != 0 {
-		raw := heap.Pop(&blocked)
+
+		// Any blocked evaluations with ModifyIndexes older than the just-ack'd
+		// evaluation are no longer useful, so it's safe to drop them.
+		cancelable := blocked.MarkForCancel()
+		b.cancelable = append(b.cancelable, cancelable...)
+		b.stats.TotalCancelable = len(b.cancelable)
+		b.stats.TotalBlocked -= len(cancelable)
+
+		// If any remain, enqueue an eval
+		if len(blocked) > 0 {
+			raw := heap.Pop(&blocked)
+			eval := raw.(*structs.Evaluation)
+			b.stats.TotalBlocked -= 1
+			b.enqueueLocked(eval, eval.Type)
+		}
+
+		// Clean up if there are no more after that
 		if len(blocked) > 0 {
 			b.blocked[namespacedID] = blocked
 		} else {
 			delete(b.blocked, namespacedID)
 		}
-		eval := raw.(*structs.Evaluation)
-		b.stats.TotalBlocked -= 1
-		b.enqueueLocked(eval, eval.Type)
 	}
 
 	// Re-enqueue the evaluation.
@@ -733,11 +754,13 @@ func (b *EvalBroker) flush() {
 	b.stats.TotalUnacked = 0
 	b.stats.TotalBlocked = 0
 	b.stats.TotalWaiting = 0
+	b.stats.TotalCancelable = 0
 	b.stats.DelayedEvals = make(map[string]*structs.Evaluation)
 	b.stats.ByScheduler = make(map[string]*SchedulerStats)
 	b.evals = make(map[string]int)
 	b.jobEvals = make(map[structs.NamespacedID]string)
-	b.blocked = make(map[structs.NamespacedID]PendingEvaluations)
+	b.blocked = make(map[structs.NamespacedID]BlockedEvaluations)
+	b.cancelable = []*structs.Evaluation{}
 	b.ready = make(map[string]PendingEvaluations)
 	b.unack = make(map[string]*unackEval)
 	b.timeWait = make(map[string]*time.Timer)
@@ -830,6 +853,7 @@ func (b *EvalBroker) Stats() *BrokerStats {
 	stats.TotalUnacked = b.stats.TotalUnacked
 	stats.TotalBlocked = b.stats.TotalBlocked
 	stats.TotalWaiting = b.stats.TotalWaiting
+	stats.TotalCancelable = b.stats.TotalCancelable
 	for id, eval := range b.stats.DelayedEvals {
 		evalCopy := *eval
 		stats.DelayedEvals[id] = &evalCopy
@@ -839,6 +863,24 @@ func (b *EvalBroker) Stats() *BrokerStats {
 		stats.ByScheduler[sched] = &subStatCopy
 	}
 	return stats
+}
+
+// Cancelable retrieves a batch of previously-blocked evaluations that are now
+// stale and ready to mark for canceling. The eval RPC will call this with a
+// batch size set to avoid sending overly large raft messages.
+func (b *EvalBroker) Cancelable(batchSize int) []*structs.Evaluation {
+	b.l.RLock()
+	defer b.l.RUnlock()
+
+	if batchSize > len(b.cancelable) {
+		batchSize = len(b.cancelable)
+	}
+
+	cancelable := b.cancelable[:batchSize]
+	b.cancelable = b.cancelable[batchSize:]
+
+	b.stats.TotalCancelable = len(b.cancelable)
+	return cancelable
 }
 
 // EmitStats is used to export metrics about the broker while enabled
@@ -856,6 +898,7 @@ func (b *EvalBroker) EmitStats(period time.Duration, stopCh <-chan struct{}) {
 			metrics.SetGauge([]string{"nomad", "broker", "total_unacked"}, float32(stats.TotalUnacked))
 			metrics.SetGauge([]string{"nomad", "broker", "total_blocked"}, float32(stats.TotalBlocked))
 			metrics.SetGauge([]string{"nomad", "broker", "total_waiting"}, float32(stats.TotalWaiting))
+			metrics.SetGauge([]string{"nomad", "broker", "total_cancelable"}, float32(stats.TotalCancelable))
 			for _, eval := range stats.DelayedEvals {
 				metrics.SetGaugeWithLabels([]string{"nomad", "broker", "eval_waiting"},
 					float32(time.Until(eval.WaitUntil).Seconds()),
@@ -878,12 +921,13 @@ func (b *EvalBroker) EmitStats(period time.Duration, stopCh <-chan struct{}) {
 
 // BrokerStats returns all the stats about the broker
 type BrokerStats struct {
-	TotalReady   int
-	TotalUnacked int
-	TotalBlocked int
-	TotalWaiting int
-	DelayedEvals map[string]*structs.Evaluation
-	ByScheduler  map[string]*SchedulerStats
+	TotalReady      int
+	TotalUnacked    int
+	TotalBlocked    int
+	TotalWaiting    int
+	TotalCancelable int
+	DelayedEvals    map[string]*structs.Evaluation
+	ByScheduler     map[string]*SchedulerStats
 }
 
 // SchedulerStats returns the stats per scheduler
@@ -933,4 +977,60 @@ func (p PendingEvaluations) Peek() *structs.Evaluation {
 		return nil
 	}
 	return p[n-1]
+}
+
+// Len is for the sorting interface
+func (p BlockedEvaluations) Len() int {
+	return len(p)
+}
+
+// Less is for the sorting interface. We flip the check
+// so that the "min" in the min-heap is the element with the
+// highest priority or highest modify index
+func (p BlockedEvaluations) Less(i, j int) bool {
+	if p[i].Priority != p[j].Priority {
+		return !(p[i].Priority < p[j].Priority)
+	}
+	return !(p[i].ModifyIndex < p[j].ModifyIndex)
+}
+
+// Swap is for the sorting interface
+func (p BlockedEvaluations) Swap(i, j int) {
+	p[i], p[j] = p[j], p[i]
+}
+
+// Push implements the heap interface and is used to add a new evaluation to the slice
+func (p *BlockedEvaluations) Push(e interface{}) {
+	*p = append(*p, e.(*structs.Evaluation))
+}
+
+// Pop implements the heap interface and is used to remove an evaluation from the slice
+func (p *BlockedEvaluations) Pop() interface{} {
+	n := len(*p)
+	e := (*p)[n-1]
+	(*p)[n-1] = nil
+	*p = (*p)[:n-1]
+	return e
+}
+
+// MarkForCancel is used to clear the blocked list of all but the one with the
+// highest modify index and highest priority. It returns a slice of cancelable
+// evals so that Eval.Ack RPCs can write batched raft entries to cancel
+// them. This must be called inside the broker's lock.
+func (p *BlockedEvaluations) MarkForCancel() []*structs.Evaluation {
+
+	// In pathological cases, we can have a large number of blocked evals but
+	// will want to cancel most of them. Using heap.Remove requires we re-sort
+	// for each eval we remove. Because we expect to have at most one remaining,
+	// we'll just create a new heap.
+	retain := BlockedEvaluations{}
+
+	raw := heap.Pop(p)
+	heap.Push(&retain, raw)
+
+	cancelable := make([]*structs.Evaluation, len(*p))
+	copy(cancelable, *p)
+
+	*p = retain
+	return cancelable
 }
--- a/nomad/eval_broker_test.go
+++ b/nomad/eval_broker_test.go
@@ -1,16 +1,21 @@
 package nomad
 
 import (
+	"container/heap"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"testing"
 	"time"
 
+	msgpackrpc "github.com/hashicorp/net-rpc-msgpackrpc"
 	"github.com/hashicorp/nomad/ci"
 	"github.com/hashicorp/nomad/nomad/mock"
+	"github.com/hashicorp/nomad/nomad/state"
 	"github.com/hashicorp/nomad/nomad/structs"
 	"github.com/hashicorp/nomad/testutil"
+	"github.com/shoenig/test"
+	"github.com/shoenig/test/must"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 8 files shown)


nomad/eval_broker.go
```diff
@@ -876,6 +876,14 @@ func (b *EvalBroker) EmitStats(period time.Duration, stopCh <-chan struct{}) {
 	}
 }
 
+// HasBlockedEval is used to check if the eval broker already has a blocked eval for this job.
+func (b *EvalBroker) HasBlockedEval(namespacedJobID structs.NamespacedID) bool {
+	b.l.Lock()
+	defer b.l.Unlock()
+	_, ok := b.blocked[namespacedJobID]
+	return ok
+}
```
tgross, line 885: Node update evals always come in with their priority set to the job priority. Maybe we should check if the blocked eval has a lower priority here as well, so that we can bump-over an eval that was submitted by a user with a low priority?

tgross: My initial concerns around this idea are that things like `SnapshotIndex` and `NodeModifyIndex` aren't getting evals with updated values, but `NodeModifyIndex` is just informational (never consumed by Nomad) and the `SnapshotIndex` is set at _scheduling_ time so there's nothing for us to screw up with that either -- none of the evals in the `PendingEvals` list we call "blocked" here will have `SnapshotIndex` set either.

tgross:
```mermaid
sequenceDiagram
    Node RPC-->>Broker: node-update 1: has blocked eval? (no)
    Node RPC->>Raft: node-update 1
    Raft-->>Broker: enqueue eval
    Node RPC-->>Broker: node-update 2: has blocked eval? (no)
    Node RPC->>Raft: node-update 2
    Raft-->>Broker: can't enqueue eval (blocked)
    Node RPC-->>Broker: node-update 3: has blocked eval? (yes)    
    Broker-->>Scheduler: GetEval
    Node RPC-->>Broker: node-update 4: has blocked eval? (no)
    Node RPC->>Raft: node-update 4
    Raft-->>Broker: can't enqueue eval (blocked)
```

nomad/eval_broker.go
schmichael: Great analysis. I had assumed `{Job,Node}ModifyIndex` would be used for the minimum index of the state store when initially dequeuing the eval in the worker, but that doesn't appear to be the case. The index is the max `ModifyIndex` of _all evals for the job._ :bangbang: 

(That actually seems like an overly pessimistic index to use as it ensures the oldest eval for a job evaluates against the same index as the _latest_ eval for a job... why even bother processing those subsequent evals then since there's no guarantee the state will be later than when the first eval was committed? If a new eval comes in in the mean time, _it_ would need to get processed.)

Unfortunately I think this logic poses a problem for your otherwise fantastic patch: you will no longer be creating an eval with a potentially higher `ModifyIndex` than all of the blocked evals for the job. So the job may end up not getting evaluated against the latest index... which is particularly a problem if the whole point is to reevaluate the job against _this_ index of the node.

To try to illustrate: The old behavior ("Blocked" means "dupe pending eval for a job" -- BlockedEvals don't factor in here AFAICT)

1. Eval E1 for Job J1 created @ Index 10
2. Eval E2 for Job J1 created and blocked @ Index 20
3. Worker dequeues E1, gets a WaitIndex of **20**
4. Node.Update occurs, updates Node and creates Eval E3 for J1 at Index 30
5. Worker dequeues E2, gets a WaitIndex of **30**
6. Worker dequeues E3, gets a WaitIndex of **30**

With this PR:

1. 1. Eval E1 for Job J1 created @ Index 10
2. Eval E2 for Job J1 created and blocked @ Index 20
3. Worker dequeues E1, gets a WaitIndex of **20**
4. Node.Update occurs, updates Node at index 30, **sees E2 blocked and skips eval creation**
5. Worker dequeues E2, gets a WaitIndex of **20**

So we may never evaluate the node at its latest ModifyIndex (30).

I could be wrong though... lots of moving parts here. Honestly I'm not sure why we calculate WaitIndex the way we do... using `Eval.{Job,Node}ModifyIndex` (based on `TriggeredBy`) initially seems *correct.* Using `max(Eval.ModifyIndex, Eval.SnapshotIndex)` always seems safe and potentially optimal though (I would assume Eval.ModifyIndex is always >= its `{Job,Node}ModifyIndex`)

I almost want to remove `{Job,Node}ModifyIndex` if they're not used. I can't imagine them being useful for end users. :shrug: 

If my assessment of the current behavior is correct, then obviously the eval you're skipping creating is useless to actually process, but it *is* necessary to ensure the modify index is high enough.

tgross: Yeah. ☹️ I hit this problem from a different direction last night. 

I was trying to work up a model of the schedulers/broker/raft inspired by the [aws s3 lightweight formal methods](https://www.amazon.science/publications/using-lightweight-formal-methods-to-validate-a-key-value-storage-node-in-amazon-s3) paper (which is an excellent read, btw). But I needed an invariant to test: we want any change to cluster state (i.e. state at a given index) to be eventually reflected in the last evaluation processed. But without writing the evaluation to raft we don't have a valid index to compare against anywhere in the scheduler.

So for some reason I had it in my head we only had a wait index concept in the planner but it's obvious in retrospect that we'd need it in the scheduler too.

tgross: > (That actually seems like an overly pessimistic index to use as it ensures the oldest eval for a job evaluates against the same index as the _latest_ eval for a job... why even bother processing those subsequent evals then since there's no guarantee the state will be later than when the first eval was committed? If a new eval comes in in the mean time, it would need to get processed.)

This observation could inform an approach to shed duplicates from the eval broker, even if we can't prevent them from getting written to raft. Don't try to avoid adding them to the `blocked` list of `PendingEvals`. But if we get an ack for an eval at `ModifyIndex = n`, then I think we can immediately discard all evals in `PendingEvals` for that job with `ModifyIndex` < `n` and batch cancel them for efficiency.

tgross: Hacking on this idea over in https://github.com/hashicorp/nomad/compare/main..f-evaluation-load-shedding-take-2 but will update this PR with the results once it's at all suitable for discussion.

tgross: Something I've run into here is that when we `heap.Pop` the `blocked` queue, we're always getting the oldest evaluation instead of the newest one. So when it later gets ack'd, it's never newer than the others. The `PendingEvals` type sorts by Priority and then CreateIndex FIFO, which is fine for `pending` (we probably need to do that for fairness across jobs). But for the `blocked` queue we want to sort by Priority but CreateIndex LIFO for this idea to work. Will have to split the types and see where that goes next.

schmichael: > we're always getting the oldest evaluation instead of the newest one. So when it later gets ack'd, it's never newer than the others.

While this is true, I don't think it matters because `WaitIndex` given to workers as the min index to work from is calculated to be the _max modify index for all evals for the job:_ https://github.com/hashicorp/nomad/blob/v1.4.0-beta.1/nomad/eval_endpoint.go#L195-L208

So for a given job it doesn't matter which of its evals we process first, we'll always process it against the latest known index at the time of dequeue. I think when completing an eval we could go ahead and cancel all other evals for the same job with a <= ModifyIndex to the plan's _SnapshotIndex_ as that's guaranteed to be >= the WaitIndex.

Put another way, if you have the following evals for the same job:

1. node-update @ 100
2. preemption @ 200

Then even though the `node-update @ 100` is dequeued, the `WaitIndex` will be `200`. The resulting Plan's `SnapshotIndex` will be >= `200` so if the `node-update @ 100` eval is complete, the `preemption @ 200` eval can be canceled as well.

`Oldest Eval ModifyIndex <= WaitIndex at Dequeue time <= Worker's Snapshot Index`

tgross: Ok, that behavior doesn't show up at a unit testing level because we're not running a real worker, just making `Dequeue` and `Ack` calls, but that eliminates some extra code I've had to write here. Just need to update the tests to match that behavior and I think I've actually got this working.

tgross:
@schmichael @lgfa29 @shoenig I've now got this into working shape (and green on CI except for the usual test flakes in unrelated areas of code). I'm reasonably comfortable with the idea of landing this in 1.4.0-rc1 for next week but it's probably worth having a discussion about how good we feel about this change coming so late.

tgross:
Oops, looks like I need to fix expectations on `TestJobEndpoint_Register_NonOverlapping`

lgfa29 (approved):
Great job! 

I was trying to think if there was another place where the batch raft update happens to keep the logic closer together (like some kind of cancelled evals GC). I was also a bit confused initially why the endpoint was calling `Cancelable(1000)` since it would return evals not necessarily related to the eval being ACK'ed, but I don't think the broker has access to raft.

nomad/eval_broker.go
```diff
@@ -559,6 +564,7 @@ func (b *EvalBroker) Ack(evalID, token string) error {
 		return fmt.Errorf("Token does not match for Evaluation ID")
 	}
 	jobID := unack.Eval.JobID
+	recentlyAckedIndex := unack.Eval.ModifyIndex
```
schmichael, line 567: **Update:** See my comment below about *canceling all but 1 eval.* I don't think the rest of this comment matters at all, but I spent like 4 hours on it so by golly I'm leaving it. :face_with_spiral_eyes: 

**tl;dr** -> I'm not sure this will cancel any evals. I think having the worker put `Eval.SnapshotIndex` into `structs.AckRequest` would allow the broker to cancel the maximum number of evals safely. This gets extremely difficult to reason about as it seems every layer of scheduler (broker, rpc, worker, scheduler) seems to do its own funny little index handling, so please check my work! :grimacing: 

AFAICT `Eval.ModifyIndex` does *not* include the index at which the eval was updated after committing a plan. `unack` is pulled out of `EvalBroker.unack` which contains the original copy of the eval from when it was first inserted into the broker as ready.

This means that `Eval.ModifyIndex` is pessimistic in the sense that it is the *minimum* possible value the Worker's snapshot was at when this evaluation was processed. Luckily we have a whole smorgasbord of indexes in the worker to choose from:

- `EvalDequeueResponse.WaitIndex` - the maximum Eval.ModifyIndex of all evals for this job when it was dequeued.
- `Eval.SnapshotIndex` - the _actual_ Raft index the worker's state snapshot used. *Only available on the _updated_ Eval.*
- `PlanResult.AllocIndex / PlanResponse.Index` - returned by the `Plan.Submit` RPC to the worker who submitted the plan, but never used. I think the worker only ever uses the `PlanResult.RefreshIndex` and _that_ index never bubbles up to where the Worker calls `Ack`.

I think the following holds as an invariant:

```
PlanResult.AllocIndex > {Updated}Eval.SnapshotIndex >= EvalDequeueResponse.WaitIndex >= {Broker}Eval.ModifyIndex
          ^                          ^                               ^                              ^
          |                          |                               |                              |
          |    no point in processing new evals < this index         |                 this is what's being used
          |                                                          |
too new! cluster may have changed prior to this                      |
                                                                     |      
     safe choice as if it's > ModifyIndex that indicates there are pending evals that can be optimized out!

```

I think `Eval.SnapshotIndex` is precisely what we're looking for. There's no point processing a new eval for the same job if we're not going to process it at a higher index than this because we will come up with the same result as before since we haven't allowed the cluster state to change!

Since there could be updates between `PR.AllocIndex` and `SnapshotIndex`, we can't drop any evals based on `AllocIndex`.

nomad/eval_broker_test.go
```diff
@@ -1506,3 +1383,40 @@ func TestEvalBroker_NamespacedJobs(t *testing.T) {
 	require.Equal(1, len(b.blocked))
 
 }
+
+func TestEvalBroker_BlockedEvals_MarkForCancel(t *testing.T) {
```
schmichael, line 1387: I wonder if we could add an integration test to avoid creating unrealistically ideal evals and indexes in tests.

Since we can manipulate NumSchedulers on a live server it should be possible to start a TestServer or TestAgent with num_schedulers=0 and create some interesting pending evals by:

1. Registering a system job (creates an initial eval)
2. Register a node (creates a second eval for the system job)
3. Set num_schedulers=1
4. ...wait...
5. Assert the initial eval was `complete`'d successfully and the second eval was `canceled`

I think that should assert that your optimization works in somewhat realistic conditions. Obviously we aren't concerned with folks actually flipping num_schedulers, but I think that's a reasonable bit of unrealistic interaction to let us get a queue of evals for a single job in such a way that is currently *not* optimized but would *should be optimized by this PR.*

nomad/eval_endpoint.go
```diff
@@ -229,6 +229,27 @@ func (e *Eval) Ack(args *structs.EvalAckRequest,
...
+		}
+
+		update := &structs.EvalUpdateRequest{
+			Evals:        cancelable,
+			WriteRequest: structs.WriteRequest{Region: args.Region},
+		}
+		_, _, err = e.srv.raftApply(structs.EvalUpdateRequestType, update)
+		if err != nil {
+			e.logger.Error("eval cancel failed", "error", err, "method", "ack")
+			return err
```
schmichael, line 249: I think we can just `WARN` here as we're only bypassing an optimization and not affecting correctness.

I'd say we could skip returning the error as well, but worker.go just logs it and throws it away anyway. I suppose we can't really control the error logging on that side, so it may not be worth fiddling with the level here. :shrug:

nomad/eval_endpoint.go
```diff
@@ -229,6 +229,27 @@ func (e *Eval) Ack(args *structs.EvalAckRequest,
 	if err := e.srv.evalBroker.Ack(args.EvalID, args.Token); err != nil {
 		return err
 	}
+
+	const cancelDesc = "cancelled after more recent eval was processed"
```
schmichael, line 233:
```suggestion
	const cancelDesc = "canceled after more recent eval was processed"
```

I know the const is spelled wrong, but its value is actually spelled right! We should probably just correct the spelling of `structs.EvalStatusCancelled` since we don't promise API stability of that package.

nomad/eval_endpoint.go
```diff
@@ -229,6 +229,27 @@ func (e *Eval) Ack(args *structs.EvalAckRequest,
...
+
+	cancelable := e.srv.evalBroker.Cancelable(1000)
+	if len(cancelable) > 0 {
+		for _, eval := range cancelable {
+			eval.Status = structs.EvalStatusCancelled
+			eval.StatusDescription = cancelDesc
```
schmichael, line 239:
```suggestion
			eval.Status = structs.EvalStatusCancelled
			eval.StatusDescription = cancelDesc
			eval.UpdateModifyTime()
```

nomad/eval_broker.go
```diff
@@ -934,3 +969,43 @@ func (p PendingEvaluations) Peek() *structs.Evaluation {
...
 	return p[n-1]
 }
+
+// PopN removes and returns the minimum N evaluations from the slice.
+func (p *PendingEvaluations) PopN(n int) []*structs.Evaluation {
+	if n > len(*p) {
+		n = len(*p)
+	}
+
+	popped := []*structs.Evaluation{}
```
schmichael, line 979: Since `n` can be significant here (1000), let's try to avoid the slice resizing.
```suggestion
	popped := make([]*structs.Evaluation, 0, n)
```

Although who knows, the Go compiler might be able to figure that out on its own.

nomad/eval_broker.go
```diff
@@ -586,15 +592,28 @@ func (b *EvalBroker) Ack(evalID, token string) error {
...
+		// evaluation are no longer useful, so it's safe to drop them.
+		cancelable := blocked.MarkForCancel(recentlyAckedIndex)
+		b.cancelable = append(b.cancelable, cancelable...)
+		b.stats.TotalCancelable = b.cancelable.Len()
+		b.stats.TotalBlocked -= cancelable.Len()
+
+		// If any remain, enqueue an eval
+		if len(blocked) > 0 {
+			raw := heap.Pop(&blocked)
+			eval := raw.(*structs.Evaluation)
```
schmichael, line 606: Surely @shoenig has a generic heap for us to avoid having to assert a type? :grin:

nomad/eval_broker.go
```diff
@@ -934,3 +969,43 @@ func (p PendingEvaluations) Peek() *structs.Evaluation {
...
+		if eval.ModifyIndex >= index {
+			heap.Push(&retain, eval)
+		} else {
+			heap.Push(&cancelable, eval)
+		}
+	}
+
+	*p = retain
+	return cancelable
+}
```
schmichael, line 1011: Hold on... due to my "favorite" logic in the `Eval.Dequeue` RPC: https://github.com/hashicorp/nomad/blob/v1.4.1/nomad/eval_endpoint.go#L185-L211

...I think we can cancel *every eval except the one with the highest ModifyIndex?!*

Since on Dequeue we always set `WaitIndex = max(ModifyIndex)` it doesn't matter which eval we dequeue next for a job: we'll always process it against a new enough snapshot to capture every cluster change that triggered evals!

...it almost makes me think we could cancel these evals in the FSM, but we'd still need to cancel them in EvalBroker since the broker isn't a "live" view of the evals in the statestore... might not be impossible, but when in doubt I like to keep the FSM simple.

schmichael (commented):
I've spent most of my day trying to reason about this and my brain is fully and completely melted. :sweat_smile: 

I think everything you have done is correct and safe, but I don't think it's optimal. The current approach may not cancel any evals in realistic situations, but I'm not sure about that.

nomad/eval_broker.go
tgross: I think you're right on this, but I'm going to work up a diagram of what indexes we have and what eval states we have in the broker just to double-check. 

My next instinct here is to mark the eval as cancelable at the time we enqueue rather than at the time we ack. Have `EvalBroker.blocked` contain a spot for at most one eval, and have `EvalBroker.Enqueue` drop the eval into `EvalBroker.cancelable` immediately if it's full. (As you've said, I'd love to be able to have the FSM detect that and delete the eval right away as well, but the followers wouldn't see the eval broker state. This would be the next best thing.) Let me work up that diagram first and we can come back to this discussion.

tgross: I think one of the things that's misleading about the existing eval broker code is the `PendingEvaluations` sorting by `CreateIndex`. The `PendingEvaluations` alias over a slice of evals gets used by both `blocked` and `ready` but only `ready` actually cares about the index b/c it contains evals from multiple jobs. The `blocked` slice is full of evals from the same job, so the order we pick them off doesn't matter.

nomad/eval_broker.go
tgross: Agreed. Based on the testing I did for https://github.com/hashicorp/nomad/pull/15117 I think we're probably going to want to crank the slice of the max size down quite a bit as well.

nomad/eval_broker_test.go
tgross: > I wonder if we could add an integration test to avoid creating unrealistically ideal evals and indexes in tests.

Yeah this updated `TestEvalBroker_Serialize_DuplicateJobID` test is totally broken b/c it's assuming that a change to the Eval's modify index is at all visible to the broker, which it is not. I'm not sure how I got it in my head that we were getting an update Eval. 😊

nomad/eval_broker.go
tgross: Well merging a state diagram together with the sequence diagram for the RPCs turned out to be a bad idea. But after a couple hours re-re-reading this code I've got a state diagram here that makes most of the problem clear, which I can annotate.

```mermaid
stateDiagram-v2

    isPending: check if jobEvals\n(pending JobID) is empty
    state if_state <<choice>>

    [*] --> isPending: Enqueue
    isPending --> if_state
    if_state --> ready: no pending JobID.\nready is sorted\nby priority and CreateIndex
    if_state --> blocked: pending JobID

    blocked --> ready: Ack

    ready --> unack: Dequeue
    unack --> [*]: Ack

    unack --> isPending: Nack
```

When we enqueue, we check `jobEvals` (better described as "pending jobs") to see if its non-empty.
- If it's empty, we insert the job ID into `jobEvals` and push the eval onto the `ready` heap.
- If it's non-empty, we insert the eval into `blocked` and return.

When we dequeue, we:
- `scanForSchedulers`:
  - peeks evals from `ready` for each scheduler
  - finds the eval with the highest priortity (or picks between ones with the same)
  - in `dequeueForSched`, pop the selected eval from `ready` and adds it to `unack`
- `getWaitIndex(eval.ns, eval.jobid, eval.ModifyIndex)` for the eval we got:
  - gets highest modify index for any eval for the job, or dequeued eval's modify index, _whichever is greater_

When we submit the plan:
- [updateEvalModifiyInxex](https://github.com/hashicorp/nomad/blob/v1.4.2/nomad/state/state_store.go#L403) updates the eval, but this will not be reflected in eval broker

> // updateEvalModifyIndex is used to update the modify index of an evaluation that has been
> // through a scheduler pass. This is done as part of plan apply. It ensures that when a subsequent
> // scheduler workers process a re-queued evaluation it sees any partial updates from the plan apply.

When we ack the eval:
- remove from `unack`
- delete jobID from `jobEvals` so the job isn't marked "pending"
- enqueue top `blocked` item (which will mark us as "pending" again)

So `ready` sorting is the only thing in the eval broker that cares about `CreateIndex`, and only `getWaitIndex` checks the modify index. As long as we're always setting the eval with the highest modify index as ready, we're good. I _think_ we can cancel the entire `blocked` list when we do so.

```mermaid
stateDiagram-v2

    isPending: check if jobEvals\n(pending JobID) is empty
    state if_state <<choice>>

    [*] --> isPending: Enqueue
    isPending --> if_state
    if_state --> ready: no pending JobID.\nready is sorted\nby priority and CreateIndex
    if_state --> blocked: pending JobID

    blocked --> ready: on Ack\neval w/ highest ModifyIndex
    blocked --> canceled: cancel all the rest

    ready --> unack: Dequeue
    unack --> [*]: Ack

    unack --> isPending: Nack
```

Alternately, if we wanted to be extra paranoid to ensure we haven't missed something we could cancel all but one eval out of `blocked`. That still helps us for the "large number of blocked evals" scenario but makes for bimodal operation where we won't hit this new cancel workflow except in cases where there's a backlog, which I don't love.

tgross: I'll implement this and make sure we've got an integration test that exercises it.

tgross: Started on implementation and immediately remembered that evals have their own priority, so we can't just naively use the highest modify index either: we could get a higher-priority eval (ex 100) and then a lower-priority one (ex 50). If we just use highest modify index the lower-priority eval would be the one that gets set to ready and we'd discard the high-priority one. Then that low-priority eval could get starved by middling-priority evals for other jobs (ex 70). We'll need to account for both fields.

nomad/eval_broker_test.go
tgross: Done in `TestEvalBroker_IntegrationTest`. I've got it so that it flaps the node status and then asserts that we only get 2 evals processed (the first one and then the one from the top of the `blocked` queued once that one is ack'd), with the rest being canceled.

nomad/eval_endpoint.go
```diff
@@ -766,3 +767,35 @@ func (e *Eval) Allocations(args *structs.EvalSpecificRequest,
 		}}
 	return e.srv.blockingRPC(&opts)
 }
+
+// cancelCancelableEvals pulls a batch of cancelable evaluations from the eval
+// broker and updates their status to canceled.
+func cancelCancelableEvals(srv *Server) error {
```
tgross, line 773: Note for reviewers: this is used by both the `Eval.Ack` endpoint and the leader's reaper, but we don't have a great spot for it in the code. Definitely open to suggestions for a better spot if you've got one.

tgross:
I've run a middling-scale performance test of this PR and the results look pretty good. I stood up a cluster of 3 `t3.large` instances on AWS (8GiB RAM).

* Disable the schedulers on all 3 instances by setting `server.num_schedulers = 0` and reloading.
* Deploy 10 sytem jobs.
* Deploy 5000 simulated client nodes via [`nomad-nodesim`](https://github.com/schmichael/nomad-nodesim).
* Wait until all 5000 nodes are registered.


At this point, there are 49730 pending evaluations in the state store and in the `blocked` queue on the broker. After commenting-out `server.num_schedulers = 0` and reloading, the schedulers are restarted and begin processing.
(19:30)

After 15s (19:30):
* The `blocked` queue is empty.
* Because some evals have been acked, we've already updated 16159 evals to canceled in raft, with 33694 evals in the `canceleable` queue on the broker.
* At this point, the eval broker has no more work in-flight (in the `ready` queue).
* No more acks are happening to drive cancellations, so we've fallen back to moving batches of evals out of the `canelable` list and writing them to raft as `canceled` every 5 sec.

After ~2m (19:32), there are still 16950 evals in the `pending` state (still on the `cancelable` queue).

At this point, I stopped the 5000 client nodes. Within 10s, all remaining evals on the `cancelable` queue have been writting to raft as `caneled`.

As each of the 5000 nodes miss heartbeats they spawn 10 evaluations each. We can see from the broker metrics that at no point do we have any backlog on the `blocked` list, because we're immediately moving batches off to `cancelable` and clearing them to `canceled` on the next ack. So long as you've got new evals coming in, the `cancelable` list is getting quickly cleared out.

Screen Shot 2022-11-09 at 2 58 45 PM

schmichael:
@tgross That test is phenomenal! Any chance you have the terraform or whatever laying around to run it again against 1.4.2? I'd love to compare the graphs.

nomad/eval_endpoint.go
schmichael: Short of a sizable refactor, I can't think of anywhere better.

Evaluations themselves have some methods for deriving _new_ evals, but that's not what's happening here.

The scheduler has a little `setStatus(...)` func that does effectively this ... and it's jammed in a `util.go` file, so clearly we don't have a good place to put eval munging. :sweat_smile:

nomad/eval_endpoint.go
```diff
@@ -230,7 +230,8 @@ func (e *Eval) Ack(args *structs.EvalAckRequest,
 	if err := e.srv.evalBroker.Ack(args.EvalID, args.Token); err != nil {
 		return err
 	}
-	return nil
+
+	return cancelCancelableEvals(e.srv)
```
schmichael, line 234: Hm, so this introduces a raft operation in `Ack(...)` which didn't have one before.

Since cancelable evals are already moved out of the blocked queue, we don't *need* to commit them to make the optimization work. However the longer we wait between updating the Ack'd eval and committing canceled evals, the longer the statestore will confusingly have `ready` evals that aren't be processed because they're sitting in the cancelable queue... :thinking: 

Since your benchmarking demonstrated this approach is already a massive improvement, let's leave this be.

However, I think we could make this async (maybe by just making it wakeup the reapCancelableEvaluations loop?) and potentially speed things up further.


If I'm right it might be worth commenting because it will be very hard for futures folks to load the context to make this determination:

```suggestion
	// It's not necessary to cancel evals before Ack returns, but it's done here to
	// commit canceled evals as close to the Ack'd eval being committed as possible.
	return cancelCancelableEvals(e.srv)
```

nomad/leader.go
```diff
@@ -992,6 +995,24 @@ func (s *Server) reapDupBlockedEvaluations(stopCh chan struct{}) {
 	}
 }
 
+// reapCancelableEvaluations is used to reap evaluations that were marked
+// cancelable by the eval broker and should be cancelled. These get swept up
```
schmichael, line 999:
```suggestion
// cancelable by the eval broker and should be canceled. These get swept up
```

literally impossible to remember to be consistent with this spelling... stupid english

schmichael (approved):
No blockers or even logic changes! Merge away and fantastic work. This is the most significant eval broker improvement in years. :tada:

nomad/eval_endpoint.go
tgross: Actually, the wake-up channel is probably a safer implementation -- if it's a buffered channel that'll ensure that `Ack` never gets blocked on raft. I've done that here. That also lets me move the `cancelCancelableEvals` into `leader.go`, as it now has only that one caller.

nomad/eval_endpoint.go
tgross: Moved into leader.go because `Eval.Ack` is no longer using it.

nomad/eval_broker.go
```diff
@@ -115,11 +119,14 @@ type unackEval struct {
...
-// PendingEvaluations is a list of waiting evaluations.
-// We implement the container/heap interface so that this is a
-// priority queue
+// PendingEvaluations is a list of ready evaluations across multiple jobs. We
+// implement the container/heap interface so that this is a priority queue.
 type PendingEvaluations []*structs.Evaluation
 
+// BlockedEvaluations is a list of blocked evaluations for a given job. We
+// implement the container/heap interface so that this is a priority queue.
+type BlockedEvaluations []*structs.Evaluation
```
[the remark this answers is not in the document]
tgross: Plus I still need to fix the metric name for https://github.com/hashicorp/nomad/issues/6480 so we can do that at the same time.

nomad/eval_endpoint.go
schmichael: Did you mean to change this call to use the wakeup chan, because it's still performing the raft apply synchronously on main.

I don't think it's a big deal, especially because your benchmark demonstrated it's an immense speedup already.

tgross: 🤦  Yes. Thanks for catching that. This line should now be 

```go
s.reapCancelableEvalsCh <-struct{}
return nil
```

tgross: Fixed in https://github.com/hashicorp/nomad/pull/15294

PR #14629, opened 2022-09-20
api: provide more detail on ACL bootstrap request error

lgfa29:
JSON decode error messages can be quite cryptic and hard to pinpoint, so this provides more context that the error happen while trying to decode the request body.


tgross (approved):
LGTM

PR #14630, opened 2022-09-20
[CI-only] Update build trigger to run on pushes to main and the release branches

mdeggies:
This updates nomad to always builds on commits to `main` and the release branches. This will avoid backport issues in the build workflow (from `main` into the release branches) and let us avoid having to manually allowlist new release branches in the build workflow. This also matches what is in the release configuration file: https://github.com/hashicorp/nomad/blob/main/.release/ci.hcl#L14-L17.


.github/workflows/build.yml
```diff
@@ -2,8 +2,9 @@ name: build
 
 on:
   push:
-    branches:
-      - "main"
+    branches-ignore:
+      - docs/**
+      - backport/docs/**
```
tgross, line 7: We don't run builds tests for branches prefixed by E2E, as those changes only impact our nightly E2E suite (and takes 1hr+):

```suggestion
    branches-ignore:
      - docs/**
      - backport/docs/**
      - e2e-*
```

claire-labry (approved):

DerekStrickland (approved):
:shipit:

tgross:
Leaving a note that there's still some open discussion on this one internally: https://hashicorp.slack.com/archives/CUYKT2A73/p1663623216766039

mdeggies:
I updated this PR to always trigger builds in `main` and the `release/**` branches. This will avoid backport issues in the build workflow (from `main` into the release branches) and let us avoid having to manually allowlist new release branches in the build workflow. This also matches what is in the release configuration file: https://github.com/hashicorp/nomad/blob/main/.release/ci.hcl#L14-L17.

tgross (approved):
:+1: LGTM

PR #14631, opened 2022-09-20
Allow specification of a custom job name/prefix for parameterized jobs

### LINKED ISSUE #13594: Allow specification of a custom job name/prefix for parameterized jobs
### Proposal

Currently, when a parameterized job is dispatched, the jobId of the child job is generated using a fixed template as described here:

https://github.com/hashicorp/nomad/blob/49270ed0013a90ecc5a4ff7e7ede24858372fa30/nomad/structs/structs.go#L5398-L5401

When running many versions of a parameterized job, tracking individual dispatches is hard.

I was wondering if it would be possible to allow specifying a optional "prefix" for an individual dispatch or the ability to specify a child job id as a template in the parent parameterized job. e.g.: `ffmpeg-parameterized/dispatch-{prefix}-uuid`

### Use-cases

This would be helpful when using a parameterized job for multiple different clients or tasks, checking status of jobs either through the UI or through the CLI is difficult. Taking the `ffmpeg` sample task from the docs, this would allow a user to specify which project a specific dispatch might be associated with.

### Attempted Solutions

I am currently using metadata fields to track this but this workflow is hard to use through the CLI and through the UI.

geovannyAvelar:
Now it's possible to use a custom prefix in the job ID of the parameterized jobs. It's just necessary to use a new flag called `id-prefix-template` on `nomad job dispatch` command.

This PR fixes #13594.

### DIFF
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -431,12 +431,13 @@ func (j *Jobs) Summary(jobID string, q *QueryOptions) (*JobSummary, *QueryMeta, 
 }
 
 func (j *Jobs) Dispatch(jobID string, meta map[string]string,
-	payload []byte, q *WriteOptions) (*JobDispatchResponse, *WriteMeta, error) {
+	payload []byte, idPrefixTemplate string, q *WriteOptions) (*JobDispatchResponse, *WriteMeta, error) {
 	var resp JobDispatchResponse
 	req := &JobDispatchRequest{
-		JobID:   jobID,
-		Meta:    meta,
-		Payload: payload,
+		JobID:            jobID,
+		Meta:             meta,
+		Payload:          payload,
+		IdPrefixTemplate: idPrefixTemplate,
 	}
 	wm, err := j.client.write("/v1/job/"+url.PathEscape(jobID)+"/dispatch", req, &resp, q)
 	if err != nil {
@@ -1342,9 +1343,10 @@ type DesiredUpdates struct {
 }
 
 type JobDispatchRequest struct {
-	JobID   string
-	Payload []byte
-	Meta    map[string]string
+	JobID            string
+	Payload          []byte
+	Meta             map[string]string
+	IdPrefixTemplate string
 }
 
 type JobDispatchResponse struct {
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -768,6 +768,7 @@ type JobDispatchRequest struct {
 	Payload []byte
 	Meta    map[string]string
 	WriteRequest
+	IdPrefixTemplate string
 }
 
 // JobValidateRequest is used to validate a job
@@ -5366,8 +5367,13 @@ func (d *ParameterizedJobConfig) Copy() *ParameterizedJobConfig {
 
 // DispatchedID returns an ID appropriate for a job dispatched against a
 // particular parameterized job
-func DispatchedID(templateID string, t time.Time) string {
+func DispatchedID(templateID, idPrefixTemplate string, t time.Time) string {
 	u := uuid.Generate()[:8]
+
+	if idPrefixTemplate != "" {
+		return fmt.Sprintf("%s%s%s-%d-%s", templateID, DispatchLaunchSuffix, idPrefixTemplate, t.Unix(), u)
+	}
+
 	return fmt.Sprintf("%s%s%d-%s", templateID, DispatchLaunchSuffix, t.Unix(), u)
 }
 
--- a/.changelog/14631.txt
+++ b/.changelog/14631.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+cli: Added `-id-prefix-template` option to `nomad job dispatch`
+```
--- a/command/job_dispatch.go
+++ b/command/job_dispatch.go
@@ -107,6 +107,7 @@ func (c *JobDispatchCommand) Run(args []string) int {
 	var detach, verbose bool
 	var idempotencyToken string
 	var meta []string
+	var idPrefixTemplate string
 
 	flags := c.Meta.FlagSet(c.Name(), FlagSetClient)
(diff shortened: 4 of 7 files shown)


api/jobs.go
```diff
@@ -836,9 +836,10 @@ func (p *PeriodicConfig) GetLocation() (*time.Location, error) {
 
 // ParameterizedJobConfig is used to configure the parameterized job.
 type ParameterizedJobConfig struct {
-	Payload      string   `hcl:"payload,optional"`
-	MetaRequired []string `mapstructure:"meta_required" hcl:"meta_required,optional"`
-	MetaOptional []string `mapstructure:"meta_optional" hcl:"meta_optional,optional"`
+	Payload        string   `hcl:"payload,optional"`
+	MetaRequired   []string `mapstructure:"meta_required" hcl:"meta_required,optional"`
+	MetaOptional   []string `mapstructure:"meta_optional" hcl:"meta_optional,optional"`
+	TemplatePrefix string   `mapstructure:"template_prefix" hcl:"template_prefix,optional"`
```
tgross, line 842: @geovannyAvelar what you've got here will work to change the dispatched job ID, but I'm not sure it really accomplishes the desired goal from #13594. This prefix is defined in the jobspec, so all the resulting jobs from [`nomad job dispatch`](https://www.nomadproject.io/docs/commands/job/dispatch) would have the same prefix. That is, if I have a job `example` with the following jobspec:

```hcl
parameterized {
    payload       = "required"
    meta_required = ["dispatcher_email"]
    meta_optional = ["pager_email"]
    template_prefix = "myprefix"
  }
```

If I then dispatch multiple jobs, it's going to be `example/dispatch-myprefix-d68869b7` and `example/dispatch-myprefix-aabbccddef`, which doesn't help differentiate them beyond what we're already doing today.

I think we either want a new flag on `nomad job dispatch` or to extend what you've done here to interpolate the metadata set by the user with the [`-meta`](https://www.nomadproject.io/docs/commands/job/dispatch#meta) option.

nomad/structs/structs.go
```diff
@@ -5366,9 +5368,14 @@ func (d *ParameterizedJobConfig) Copy() *ParameterizedJobConfig {
...
 // DispatchedID returns an ID appropriate for a job dispatched against a
 // particular parameterized job
-func DispatchedID(templateID string, t time.Time) string {
+func DispatchedID(prefix, templateID string, t time.Time) string {
 	u := uuid.Generate()[:8]
-	return fmt.Sprintf("%s%s%d-%s", templateID, DispatchLaunchSuffix, t.Unix(), u)
+
+	if prefix == "" {
+		prefix = fmt.Sprint(t.Unix())
+	}
```
tgross, line 5376: Getting rid of the unix timestamp entirely means that these won't sort by name anymore.

nomad/structs/structs.go
```diff
@@ -5366,9 +5368,14 @@ func (d *ParameterizedJobConfig) Copy() *ParameterizedJobConfig {
...
+
+	return fmt.Sprintf("%s%s%s-%s", templateID, DispatchLaunchSuffix, prefix, u)
```
tgross, line 5378: I wrote a batch job from this branch and wanted to mess around a bit with some options for how I might want to use interpolation here. Here's an example job with a `template_prefix`:

```hcl
job "example" {
  type        = "batch"
  datacenters = ["dc1"]

  parameterized {
    payload         = "required"
    meta_required   = ["foo"]
    template_prefix = "foo-${meta.foo}"
  }

  group "group" {
    task "task" {

      driver = "raw_exec"

      config {
        command = "/bin/sh"
        args    = ["-c", "cat local/payload.txt; sleep 1"]
      }

      dispatch_payload {
        file = "payload.txt"
      }

    }
  }
}
```

If I dispatch this with `nomad job dispatch -meta foo=bar example ./payload.txt`, I'd want that to be interpolated into something like `example/dispatch-foo-bar-b87be30b`. Right now that's not getting interpolated so of course we get the following instead:

```
$ nomad job status
ID                                               Type                 Priority  Status   Submit Date
example                                          batch/parameterized  50        running  2022-09-21T09:13:46-04:00
example/dispatch-foobar-${meta.foo}-b87be30b  batch                50        dead     2022-09-21T09:13:48-04:00
```

The tricky bit here is that the `${meta.foo}` syntax is HCL, and the HCL is getting rendered into JSON long before it hits the scheduler so we can't use HCL interpolation here. We should be able to support go templating though?

api/jobs.go
tgross, line 842: Super bikesheddy nitpick (😀): the field name `template_prefix` doesn't have an obvious relationship to the purpose. Maybe "id_prefix_template" or something?

tgross (changes requested):
Thanks for the PR @geovannyAvelar! As I noted in https://github.com/hashicorp/nomad/issues/13594#issuecomment-1253662571 I'm also going to tag @fumblehool here to see if they have additional thoughts on the implementation.

api/jobs.go
geovannyAvelar: I consider adding new a flag idea a better approach to solve that. I'm working on it.

geovannyAvelar: Surely this isn't the better name for this field. I'm not good with names. `id_prefix_template` sounds better, think I'll use it.

tgross (approved):
LGTM! Thanks @geovannyAvelar!

I've added a changelog entry for you. This missed the merge window for Nomad 1.4.0, which is going out very soon. Once that's been released, I'll merge this PR and it'll go out with the next regular release of Nomad (probably 1.4.1).

tgross:
Hi @geovannyAvelar! We had to push out a quick Nomad 1.4.1 patch, but this has now been merged and will go out in the next regular patch release.

PR #14635, opened 2022-09-21
cli: fix a bug in operator API when setting HTTPS via address.

### LINKED ISSUE #14625: Nomad Operator API ignores https protocol from NOMAD_ADDR
### Nomad version
nomad version used: 1.3.3

No needed
### Issue
`nomad operator api` creates  http connection address ignoring the `https` protocol in  NOMAD_ADDR env variable
### Reproduction steps

```
export NOMAD_ADDR=https://hashicorp.nomad:443
$ nomad operator api -dryrun /v1/jobs
curl \
  -X GET \
  http://hashicorp.nomad:443/v1/jobs ###Ignores https


nomad status
Error querying jobs: Get "https://hashicorp.nomad:443/v1/jobs": dial tcp: lookup hashicorp.nomad: no such host ###Does not ignore https protocol
```

#### Expected Result
`nomad operator api ` should constant the URL with HTTPS  when nomad address is with HTTPS

#### Actual Result
nomad is constructing HTTP address instead of HTTPS

jrasell:
Operators may have a setup whereby the TLS config comes from a source other than setting Nomad specific env vars. In this case, we should attempt to identify the scheme using the config setting as a fallback.

closes #14625


schmichael (approved):
:tada:

PR #14645, opened 2022-09-21
ci: replace hosted `semver` with an action

madhead:
I've noticed, that you are using a `curl` to download a [semver tool](https://github.com/fsaintjacques/semver-tool) from https://raw.githubusercontent.com to work with semantic versions.

I propose to use [my reusable GitHub Action](https://github.com/marketplace/actions/semver-utils) to accomplish the same task. Using the reusable action has, from my point of view, some benefits, like being more native and natural and less surprising way of doing things in GitHub Actions Workflows.

If you would like to ensure it works the same way as before, I've tested it in my fork:

- A successful run is [here](https://github.com/madhead/nomad/actions/runs/3101684238/jobs/5023285204#step:13:2), take a look at the way it parsed the input. It's identical to the old way, isn't it?
- The workflow I used for the check is [here](https://github.com/madhead/nomad/actions/runs/3101684238/workflow). I've commented some things out because they are optional for the purpose.
- Guarding against invalid input also [works](https://github.com/madhead/nomad/actions/runs/3101794529).

_Thank you!_

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,21 +37,12 @@ jobs:
         run: |
           echo "::notice::Release v${{ github.event.inputs.version }} from branch ${{ github.ref_name }}"
 
-      - name: Install semver CLI
-        run: |
-          local_bin="${HOME}/.local/bin"
-          mkdir -p "${local_bin}"
-          curl -L --output "${local_bin}/semver" \
-            https://raw.githubusercontent.com/fsaintjacques/semver-tool/3.3.0/src/semver
-          chmod +x "${local_bin}/semver"
-          echo "${local_bin}" >> $GITHUB_PATH
-
-      - name: Validate release version
-        run: |
-          if [ "$(semver validate ${{ github.event.inputs.version }})" == "invalid" ]; then
-            echo "::error::Version ${{ github.event.inputs.version }} is invalid"
-            exit 1
-          fi
+      - name: Parse the version
+        id: version
+        uses: madhead/semver-utils@v2.4.0
+        with:
+          version: ${{ github.event.inputs.version }}
+          lenient: false
 
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


tgross (commented):
Hi @madhead! Thanks for this PR but there may be internal policy issues with third-party Actions that prevent us from accepting this PR. I'm going to follow-up on that here but just a heads up that it may be a bit before I can get back to you.

madhead:
Yes, sure. I understand. That's the reason I used a fixed tag, and not `@latest` or `master`. I saw you already used external actions in a few places:

- https://github.com/hashicorp/nomad/blob/ff1a30fe8dd4a0da38ddd34915b0be2210d39614/.github/workflows/release.yml#L144
- https://github.com/hashicorp/nomad/blob/ff1a30fe8dd4a0da38ddd34915b0be2210d39614/.github/workflows/ember-test-audit.yml#L20
- https://github.com/hashicorp/nomad/blob/ff1a30fe8dd4a0da38ddd34915b0be2210d39614/.github/workflows/ember-test-audit.yml#L60
- https://github.com/hashicorp/nomad/blob/ff1a30fe8dd4a0da38ddd34915b0be2210d39614/.github/workflows/ember-test-audit.yml#L68
- https://github.com/hashicorp/nomad/blob/ff1a30fe8dd4a0da38ddd34915b0be2210d39614/.github/workflows/test-core.yaml#L40
- https://github.com/hashicorp/nomad/blob/ff1a30fe8dd4a0da38ddd34915b0be2210d39614/.github/workflows/semgrep.yml#L18

So I might have a chance :relaxed:

.github/workflows/release.yml
```diff
@@ -37,21 +37,12 @@ jobs:
...
-
-      - name: Validate release version
-        run: |
-          if [ "$(semver validate ${{ github.event.inputs.version }})" == "invalid" ]; then
-            echo "::error::Version ${{ github.event.inputs.version }} is invalid"
-            exit 1
-          fi
+      - name: Parse the version
+        id: version
+        uses: madhead/semver-utils@v2.4.0
```
tgross, line 42: Circling back on this a bit, I'm confused by this tag. The https://github.com/marketplace/actions/semver-utils URL shows the latest version being 1.0.0. And the Action tag doesn't seem to have any relationship to the semver utils version (for which we were using 3.3.0).

madhead: The action version indeed doesn't have any relation to the semver library version. That's intentional.

The repository does actually contain a `v2.4.0` tag: https://github.com/madhead/semver-utils/tree/v2.4.0. There is also a `v2.4.0` release in the releases: https://github.com/madhead/semver-utils/releases. I don't know why the Markeplace uses 1.0.0 as the version and I don't see any knobs to refresh it there.

lgfa29:
Oh, sorry for the delay here, I didn't see this PR before and this was actually a workflow that I created. 

For now I'm thinking of keeping it as-is because I'm not 100% sure where we will go in terms of automating our release process, so keeping it more script-like makes it easier for us to port it to whatever direction we go.

But thank you very much for the work here @madhead. I will keep your action in mind as we cement our release process 🙂

PR #10446, opened 2021-04-25
scheduler: stopped-yet-running allocs are still running

notnoop:
_This is a straw-man PR to kick the discussion going. Needs more testing; before merging._

Updates the scheduler so that it treats the alloc as running and using its resources until it truly completes as reported by the client. Typically, an alloc is terminates quickly upon being stopped, but an alloc may take a "long" time to properly shutdown and free up exclusive resources: e.g. ports, CSI/host volumes.

Previously, the scheduler optimistically treats stopped-yet-running allocs as terminated and may schedule an alloc in their place. This introduces a widow where both stopped-yet-running and new allocs are running on the client. During the window, the client resources can be oversubscribed unexpectedly (with OOM trigger) or lead to port/volume collision. Avoiding such window seems like the correct behavior.

I conjecture in a typical cluster, the schedule will simply place the alloc to a new client without user visible downside. For some constrained jobs, the alloc start may delayed until the previous alloc stopped.

### An Alternative

Alternatively, the client may order alloc operations so it starts new allocs only after conflicting stopped allocs complete. Ordering the operations is more complex to implement. It also delays starting the task unnecessarily - the new alloc would have started faster had it be placed on another client without a conflicting alloc.

Fixes #10440

PR #14646, opened 2022-09-22
test: add e2e for non-overlapping placements

schmichael:
Followup to #10446

Fails (as expected) against 1.3.x at the wait for blocked eval (because the allocs are allowed to overlap).

Passes against 1.4.0-beta.1 (as expected).

### DIFF
--- a/e2e/overlap/overlap_test.go
+++ b/e2e/overlap/overlap_test.go
@@ -64,7 +64,7 @@ func TestOverlap(t *testing.T) {
 		if n := len(a); n == 0 {
 			return false, fmt.Errorf("timed out before an allocation was found for %s", jobID1)
 		}
-		must.Eq(t, 1, len(a))
+		must.Len(t, 1, a)
 
 		origAlloc = a[0]
 		return origAlloc.ClientStatus == "running", fmt.Errorf("timed out before alloc %s for %s was running: %s",
--- a/e2e/overlap/overlap_test.go
+++ b/e2e/overlap/overlap_test.go
@@ -1,0 +1,130 @@
+package overlap
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/nomad/api"
+	"github.com/hashicorp/nomad/e2e/e2eutil"
+	"github.com/hashicorp/nomad/helper/uuid"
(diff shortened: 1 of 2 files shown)


e2e/overlap/overlap_test.go
```diff
@@ -0,0 +1,130 @@
...
+	defer e2eutil.WaitForJobStopped(t, nomadClient, jobID1)
+
+	var origAlloc *api.AllocationListStub
+	testutil.Wait(t, func() (bool, error) {
+		a, _, err := nomadClient.Jobs().Allocations(jobID1, false, nil)
+		must.NoError(t, err)
+		if n := len(a); n == 0 {
+			return false, fmt.Errorf("timed out before an allocation was found for %s", jobID1)
+		}
+		must.Eq(t, 1, len(a))
```
jrasell, line 67:
```suggestion
		must.Len(t, 1, a)
```

e2e/overlap/overlap_test.go
```diff
@@ -0,0 +1,130 @@
...
+	})
+
+	// Assert replacement job unblocked and running
+	testutil.Wait(t, func() (bool, error) {
+		a, _, err := nomadClient.Jobs().Allocations(jobID2, false, nil)
+		must.NoError(t, err)
+		if n := len(a); n == 0 {
+			return false, fmt.Errorf("timed out before an allocation was found for %s", jobID2)
+		}
+		must.Eq(t, 1, len(a))
```
jrasell, line 125:
```suggestion
		must.Len(t, 1, a)
```

jrasell (approved):
LGTM.

Two optional inline suggestions; it seems standard to also use `input` as the directory name that holds test jobs, again this is very optional nit-pick.

schmichael:
> Two optional inline suggestions; it seems standard to also use `input` as the directory name that holds test jobs, again this is very optional nit-pick.

Ah yeah, forgot that:

```
$ find e2e -type d -name input | wc -l
30
$ find e2e -type d -name testdata | wc -l
3
```

I'd kind of like to start using `testdata` since it appears we're trying to move e2e tests closer to "normal" go tests (or at least that was my impression from looking at #14598) since [`testdata` is the standard for inputs](https://pkg.go.dev/cmd/go#hdr-Test_packages).

PR #14649, opened 2022-09-22
scheduler: handle MRD jobs correctly in computeDeploymentPaused

pkazmierczak:
when making a check for non-existing deployment in `computeDeploymentPaused`, make sure to check whether the deployment hasn't existed in the past. This affects MRD and other type of evaluations that do not have deployments.

Resolves https://github.com/hashicorp/nomad-enterprise/issues/737


scheduler/reconcile.go
```diff
@@ -275,19 +275,36 @@ func (a *allocReconciler) computeDeploymentUpdates(deploymentComplete bool) {
...
+func (a *allocReconciler) computeMRDDeploymentPaused() {
+	// consider only jobs without current deployment
+	if a.deployment != nil {
+		return
+	}
+
+	// consider only jobs that are MRD and that use deployments
+	if !(a.job.IsMultiregion() && a.job.UsesDeployments() && !(a.job.IsPeriodic() || a.job.IsParameterized())) {
+		return
+	}
```
tgross, line 300: I'd break this up even further:

```suggestion
	// consider only jobs that are MRD 
	if !a.job.IsMultiregion() {
		return
	}	
	// consider only jobs that use deployments
	if !a.job.UsesDeployments() || (a.job.IsPeriodic() || a.job.IsParameterized()) {
		return
	}
```

Because returning early here reverses the sense of the original set of conditionals, it might also be worth considering if the `IsPeriodic`/`IsParameterized` even makes sense to include here now, as it should always be false if `!a.job.UsesDeployments`?

pkazmierczak: You're right of course, I will remove superfluous checks, thanks. (this is what happens when you have long conditions!)

.changelog/14649.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+scheduler: Fixed a bug where MRD jobs wouldn't get deployments even if they previously had one
```
tgross, line 2: I don't think "MRD" is in any of the user-facing documentation, that's just our internal acronym for it:

```suggestion
scheduler: Fixed a bug where multi-region jobs wouldn't get deployments even if they previously had one
```

tgross (approved):
LGTM

lgfa29:
Marking it as draft for now as we found more cases that need to be handled.

lgfa29:
Closing this in favour of #15325

PR #14650, opened 2022-09-22
Add Namespace, Job and Group to envoy stats

tgross:
_WIP fix for weird build issue on #14311_


tgross:
Weird, #14311 seems to be running CI now. 🤷

Issue #13101, opened 2022-05-23
Change nomad eval status -json to display on selected evaluations


mmcquillan:
### Proposal

“In Nomad 1.4.0, nomad eval status -json will be changed to display only the selected evaluation in JSON format.” 

Ref:
https://www.nomadproject.io/docs/upgrade/upgrade-specific

PR #14651, opened 2022-09-22
cli: remove deprecated `eval status -json` list behavior

tgross:
Fixes https://github.com/hashicorp/nomad/issues/13101

In Nomad 1.2.6 we shipped `eval list`, which accepts a `-json` flag, and deprecated the usage of `eval status` without an evaluation ID with an upgrade note that it would be removed in Nomad 1.4.0. This changeset completes that work.


lgfa29 (approved):
🧹

tgross:
Oops, gotta fix a test.

PR #14659, opened 2022-09-22
scheduler: Fix bug where the would treat multiregion jobs as paused for job types that don't use deployments

DerekStrickland:
This PR fixes a bug in the `allocReconciler` where multiregion job types that don't support deployments would be flagged as paused. 

- `structs.Job` - Added a `UsesDeployments` function that can be expanded as more job types support deployments
- `scheduler.allocReconciler` - Added check to `computeDeploymentPaused` to guard against setting `deploymentPaused = true` if the job type doesn't use deployments.
-

### DIFF
--- a/.changelog/14659.txt
+++ b/.changelog/14659.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-scheduler: Fixed bug where the would treat multiregion jobs as paused for job types that don't use deployments
+scheduler: Fixed bug where the scheduler would treat multiregion jobs as paused for job types that don't use deployments
 ```
--- a/.changelog/14659.txt
+++ b/.changelog/14659.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+scheduler: Fixed bug where the would treat multiregion jobs as paused for job types that don't use deployments
+```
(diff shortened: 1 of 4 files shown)


nomad/structs/structs.go
```diff
@@ -5746,6 +5746,17 @@ func (j *Job) GetScalingPolicies() []*ScalingPolicy {
...
 
+// UsesDeployments returns a boolean indicating whether the job configuration
+// results in a deployment during scheduling.
+func (j *Job) UsesDeployments() bool {
+	switch j.Type {
+	case JobTypeService:
+		return true
+	default:
+		return false
+	}
```
pkazmierczak, line 5757: Could this not be simplified?

```suggestion
        return j.Type == JobTypeService
```

DerekStrickland: It could, but I intentionally did this since a goal is to support Deployments for more job types in the future.  Maybe this is premature?  I was hoping the team would discuss so thanks for taking the bait!

.changelog/14659.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+scheduler: Fixed bug where the would treat multiregion jobs as paused for job types that don't use deployments
```
tgross, line 2:
```suggestion
scheduler: Fixed bug where the scheduler would treat multiregion jobs as paused for job types that don't use deployments
```

tgross (approved):
LGTM

Issue #13595, opened 2022-07-05
cleanup: break up nomad/mock/mock.go into more files


shoenig:
In the `nomad/mock` package we have a bunch of mock objects that are widely used in tests. Apart from `acl.go`, all mocks are stuffed into `mock.go`, making it the file huge and difficult to find anything. 

We should break up `mock.go` into more logical files, e.g. `node.go`, `alloc.go`, `job.go`, etc.

PR #14660, opened 2022-09-22
cleanup: split up mocks/mocks.go file

shoenig:
This PR splits up the nomad/mock package into more files. Specific features
that have a lot of mocks get their own file (e.g. acl, variables, csi, connect, etc.).
Above that, functions that return jobs/allocs/nodes are in the job/alloc/node file. And
lastly other mocks/helpers are in mock.go

Closes https://github.com/hashicorp/nomad/issues/13595


tgross (approved):
LGTM!

PR #14661, opened 2022-09-22
[vars:api] Return fake QueryMeta on 403s with Peek()

angrycub:
Consul-template expects that the Response Metadata is non-zero, but on a 403 we do not pass back QueryMeta headers.


tgross (approved):
LGTM. Kind of weird but seems like the right answer to dealing with the consumer's behavior. 👍

PR #14662, opened 2022-09-22
Backport of scheduler: Fix bug where the would treat multiregion jobs as paused for job types that don't use deployments into release/1.2.x

hc-github-team-nomad-core:
## Backport

This PR is auto-generated from #14659 to be assessed for backporting due to the inclusion of the label backport/1.2.x.

---

This PR fixes a bug in the `allocReconciler` where multiregion job types that don't support deployments would be flagged as paused. 

- `structs.Job` - Added a `UsesDeployments` function that can be expanded as more job types support deployments
- `scheduler.allocReconciler` - Added check to `computeDeploymentPaused` to guard against setting `deploymentPaused = true` if the job type doesn't use deployments.
-


lgfa29:
I can't reproduce the problem in the 1.2.x release line, so I'm going to assume the problem was introduced during the refactoring of `reconcile.go` for the 1.3.0 release.

PR #14665, opened 2022-09-22
[ui] Warn users when they leave an edited but unsaved variable page

### LINKED ISSUE #14658: [ui] Variables: warn users before leaving a partially-filled form

philrenaud:
Performs an equality lookup for key values and path when attempting to leave the variables/new or variables/edit routes.

Note: early attempts at using `model.hasDirtyAttributes()` in parent routes proved to be really cumbersome — would have had to do it in a few places. @ChaiWithJai made a good suggestion to try it in the component and so here we are.

image

### Limitation
This will not notify a user when they have changes in JSON mode, as we don't observe (nor want to observe) keystroke changes to that textarea. This could be done but the perf tradeoff would be non-trivial.

Resolves #14658

### DIFF
--- a/ui/app/components/variable-form.js
+++ b/ui/app/components/variable-form.js
@@ -364,7 +364,8 @@ export default class VariableFormComponent extends Component {
   }
 
   confirmExit(transition) {
-    if (transition.isAborted) return;
+    const queryParamUpdateOnly = transition.to.name === transition.from.name;
+    if (transition.isAborted || queryParamUpdateOnly) return;
 
     if (this.hasUserModifiedAttributes) {
       if (
--- a/ui/app/components/variable-form.js
+++ b/ui/app/components/variable-form.js
@@ -364,8 +364,7 @@ export default class VariableFormComponent extends Component {
   }
 
   confirmExit(transition) {
-    const queryParamUpdateOnly = transition.to.name === transition.from.name;
-    if (transition.isAborted || queryParamUpdateOnly) return;
+    if (transition.isAborted || transition.queryParamsOnly) return;
 
     if (this.hasUserModifiedAttributes) {
       if (
@@ -374,6 +373,8 @@ export default class VariableFormComponent extends Component {
         )
       ) {
         transition.abort();
+      } else {
+        this.removeExitHandler();
       }
     }
   }
--- a/ui/app/components/variable-form.js
+++ b/ui/app/components/variable-form.js
@@ -1,7 +1,7 @@
 // @ts-check
 
 import Component from '@glimmer/component';
-import { action } from '@ember/object';
+import { action, computed } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { inject as service } from '@ember/service';
 import { trimPath } from '../helpers/trim-path';
@@ -12,6 +12,7 @@ import MutableArray from '@ember/array/mutable';
 import { A } from '@ember/array';
 import { stringifyObject } from 'nomad-ui/helpers/stringify-object';
 import notifyConflict from 'nomad-ui/utils/notify-conflict';
+import isEqual from 'lodash.isequal';
 
 const EMPTY_KV = {
   key: '',
@@ -43,6 +44,7 @@ export default class VariableFormComponent extends Component {
   constructor() {
     super(...arguments);
     set(this, 'path', this.args.model.path);
+    this.addExitHandler();
   }
 
   @action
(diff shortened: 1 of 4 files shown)


ui/app/components/variable-form.js
```diff
@@ -93,7 +95,7 @@ export default class VariableFormComponent extends Component {
     if (!this.args.model?.isNew) {
       keyValues.pushObject(copy(EMPTY_KV));
     }
-    this.keyValues = keyValues;
+    set(this, 'keyValues', keyValues);
```
philrenaud, line 98: since we're now using this in a computed getter elsewhere, we should use proper `set`ters when establishing our list.

ui/app/components/variable-form.js
```diff
@@ -206,6 +208,7 @@ export default class VariableFormComponent extends Component {
         destroyOnClick: false,
         timeout: 5000,
       });
+      this.removeExitHandler();
```
philrenaud, line 211: Don't bother the user when transitioning because of a successful save

ui/app/components/variable-form.js
```diff
@@ -326,4 +329,53 @@ export default class VariableFormComponent extends Component {
...
+    return unequal;
+  }
+
+  addExitHandler() {
+    this.router.on('routeWillChange', this, this.confirmExit);
+  }
+
+  removeExitHandler() {
+    this.router.off('routeWillChange', this, this.confirmExit);
+  }
```
ChaiWithJai, line 359: `comment`: If its good for [Ballint](https://balinterdi.com/blog/warning-about-losing-unsaved-changes-in-ember-octane/), then its good for us.

ui/app/components/variable-form.js
```diff
@@ -326,4 +329,53 @@ export default class VariableFormComponent extends Component {
...
+
+  confirmExit(transition) {
+    if (transition.isAborted) return;
+
+    if (this.hasUserModifiedAttributes) {
+      if (
+        !confirm(
+          'Your variable has unsaved changes. Are you sure you want to leave?'
+        )
```
ChaiWithJai, line 368: `question`: Wouldn't this nested conditional always resolve to a `falsy` value?

Taking a look at the event handler `confirm` it doesn't have a `return` and would thus return `void` a `falsy` value every time.

What are you trying to guard against? Maybe that the confirmation action has successfully dispatched... in which case the only thing that can confirm that would be being in the `idle` state.

ChaiWithJai (approved):

ui/app/components/variable-form.js
ChaiWithJai: Hey, buddy! Looks like we end up removing the event listener twice and that causes our tests to fail.

The bottom of this [blog](https://balinterdi.com/blog/warning-about-losing-unsaved-changes-in-ember-octane/) should have a quick patch.

ChaiWithJai (changes requested):
We're removing the event listener twice which causes tests to fail. There should be a quick [solution](https://balinterdi.com/blog/warning-about-losing-unsaved-changes-in-ember-octane/) here. Scroll to the last few headers in the post.

ui/app/components/variable-form.js
philrenaud: Confirm returns whatever the user selects. If "OK", it comes back true. If "Cancel", false.

philrenaud:
Good eye, @ChaiWithJai! Fixed.

lgfa29 (changes requested):
I noticed two problems:

- We can probably ignore navigation due to the JSON toggle, as that doesn't cause any data loss
- Sometimes I get double alerts? I haven't figured out a pattern for when this happens, but it seems more frequent at the very first time I try to navigate away with unsaved changes.

https://user-images.githubusercontent.com/775380/191859854-a8c52dde-7393-49b8-bfbe-c0260edeeb17.mov

ChaiWithJai (approved):

philrenaud:
Great catch @lgfa29 — updated to remove the double-hit and query-params-only transitions

(TIL: Ember's router transitions have a `queryParamsOnly` property built in!)

ui/app/templates/variables/variable/edit.hbs
```diff
@@ -11,7 +11,7 @@
   Edit
   {{this.model.path}}
   <Toggle
-    data-test-memory-toggle
+    data-test-json-toggle
```
philrenaud, line 14: Side-effect: was mis-copied originally.

lgfa29 (approved):
The problems I noticed are fixed now, thanks! 🎉

PR #14670, opened 2022-09-22
docs: update description for failover_heartbeat_ttl with explicit formula

shoenig:
Update docs for `failover_heartbeat_ttl` with more in-depth description of what happens as cluster size increases

### DIFF
--- a/website/content/docs/configuration/server.mdx
+++ b/website/content/docs/configuration/server.mdx
@@ -135,15 +135,33 @@ server {
   a tradeoff as it lowers failure detection time of nodes at the tradeoff of
   false positives and increased load on the leader.
 
-- `failover_heartbeat_ttl` `(string: "5m")` - Specifies the TTL applied to
-	heartbeats after a new leader is elected, since we no longer know the status
-	of all the heartbeats. This is specified using a label suffix like "30s" or
-	"1h".
+- `failover_heartbeat_ttl` `(string: "5m")` - Specifies the initial TTL applied to
+  Client heartbeats after a new leader is elected. This TTL must be high enough
+  such that a Client previously connected to a now-unresponsive server will be
+  able to find a new healthy server through which to send heartbeats to the new
+  leader. The strict minimum value depends on the number of nodes in the cluster,
+  the value of `max_heartbeats_per_second`, `min_heartbeat_ttl`, and `heartbeat_grace`.
+  As the number of nodes in the cluster increases, the longer the TTL each client
+  is assigned. This value is uniformly randomized between the values computed by
+  these two formula:
 
-  ~> Lowering the `failover_heartbeat_ttl` is a tradeoff as it lowers failure
+```
+<min ttl> = max(((<num nodes> / <max_heartbeats_per_second>) + <heartbeat_grace>), <min_heartbeat_ttl>)
+<max ttl> = max(((2 * (<num nodes> / <max_heartbeats_per_second>)) + <heartbeat_grace>), <min_heartbeat_ttl>)
+```
+
+  The default value of `"5m"` is fine for small and medium sized clusters of fewer
+  than approximately 5k nodes - the value of `failover_heartbeat_ttl` must be
+  higher than `<max ttl>`, ideally with some extra padding for clusters growing
+  over time. If `<max ttl>` exceeds the failover heartbeat ttl, the cluster may
+  experience Clients being marked down unexpectedly after leader transitions - in
+  particular by Clients that were connected to a leader that has suddenly gone
+  offline.
+
+  Lowering the `failover_heartbeat_ttl` is a tradeoff as it lowers failure
   detection time of nodes at the tradeoff of false positives. False positives
-  could cause all clients to stop their allocations if a leadership transition
-  lasts longer than `heartbeat_grace + failover_heartbeat_ttl`.
+  will cause clients to be marked as node status down, and their allocations will
+  be marked as lost.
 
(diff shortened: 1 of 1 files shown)


website/content/docs/configuration/server.mdx
```diff
@@ -135,15 +135,33 @@ server {
...
-  ~> Lowering the `failover_heartbeat_ttl` is a tradeoff as it lowers failure
+```
+<min ttl> = max(((<num nodes> / <max_heartbeats_per_second>) + <heartbeat_grace>), <min_heartbeat_ttl>)
+<max ttl> = max(((2 * (<num nodes> / <max_heartbeats_per_second>)) + <heartbeat_grace>), <min_heartbeat_ttl>)
+```
+
+  The default value of `"5m"` is fine for small and medium sized clusters of fewer
+  than approximately 5k nodes - the value of `failover_heartbeat_ttl` must be
+  higher than `<max ttl>`, ideally with some extra padding for clusters growing
+  over time. If `<max ttl>` exceeds the failover heartbeat ttl, the cluster may
```
schmichael, line 156: Seems unfortunate to use a lot of approximation right after giving a formula that seems to guide toward the right answer.

For example a 5,000 node cluster with everything else on defaults gets you:

```
min -> (5000 / 50) + 10 = 110 seconds (1m50s)
max -> (2 * (5000 / 50)) + 10 = 210 seconds (3m30s)
```

So on paper it would seem `failover_heartbeat_ttl = "5m"` would be ample time for all Clients to heartbeat before being considered `down` by the new leader. If we're suggesting otherwise I'm curious why.

Given the scenario:

1. `t=1s` Client C1 heartbeats to Server S1 who is the Leader, is informed to heartbeat again within 1m40s (min - grace)
2. `t=2s` Client C2 heartbeats to Server S1, is informed to heartbeat again within 3m20s (max - grace)
3. `t=10s` Server S1 crashes
4. `t=20s` Server S2 elected Leader, sets all clients ttls to`t=5m20s` (wallclock + `5m`)
5. `t=1m1s` Client C1 attempts to heartbeat to dead server...
6. `t=3m22s` Client C2 attempts to heartbeat to dead server...
7. ??? ...118 seconds pass... ???
8. `t=5m20s` Server S2 marks Clients C1 & C2 as `down` and allocations as `lost`

In the absence of all servers known to the clients being down and a Consul failure preventing finding new servers, I don't know how to account for a Client not being able to find a live server within `3m35s`. That being said I can't find where we set TCP timeouts (other than a dial timeout of 10s which wouldn't apply to existing connections), so if the clients do not receive an `connection reset by peer` error (they should) perhaps they block indefinitely? There's a 5s RPC timeout that only seems to affect how retries are performed *after* the attempted RPC fails... so if the RPC's write simply blocks indefinitely... perhaps this could occur?

website/content/docs/configuration/server.mdx
```diff
@@ -135,15 +135,33 @@ server {
...
+  experience Clients being marked down unexpectedly after leader transitions - in
+  particular by Clients that were connected to a leader that has suddenly gone
+  offline.
+
+  Lowering the `failover_heartbeat_ttl` is a tradeoff as it lowers failure
   detection time of nodes at the tradeoff of false positives. False positives
-  could cause all clients to stop their allocations if a leadership transition
-  lasts longer than `heartbeat_grace + failover_heartbeat_ttl`.
+  will cause clients to be marked as node status down, and their allocations will
+  be marked as lost.
```
schmichael, line 164: Nitpick as written: capitalize Client to be consistent and put `down` in backticks

...but I think we can do better on this whole paragraph. It's not wrong per se, but it lacks the leader election context that this particular setting uniquely applies to.

Perhaps:

> Lowering the `failover_heartbeat_ttl` gives Clients less time to potentially failover from a dead Server to a live one in the case of a Server crash causing a leader election. The failover TTL must be higher than the max heartbeat TTL as determined above with additional buffer to account for network latency, timeouts, and potentially rediscovering Server addresses via Consul.
> Raising the `failover_heartbeat_ttl` increases the time in which a down Client will *not* have its work rescheduled. This can be a particular problem if the loss of a datacenter causes a Server and number of Clients to be lost. The remaining quorum of Servers and Clients will not reschedule the down datacenter's work until this failover TTL has elapsed.

...but idk... I might be getting too wordy.

schmichael (commented):
Thank you for tackling this parameter! It's been underdocumented for far too long.

That being said I think there are some improvements we can make.

We don't have to block documentation improvements on figuring out my mysterious step `#7 ???`, but that's probably worth further investigation on its own.

website/content/docs/configuration/server.mdx
tgross: > Seems unfortunate to use a lot of approximation right after giving a formula that seems to guide toward the right answer.

So long as we're noting that we're giving room for growth (but also dynamic node scaling which will make the count less stable), I think this sentence says to users "if you keep the defaults and you're below roughly this size, don't worry about this", which I think is good messaging to when you should _start_ thinking about it as a cluster admin.

schmichael: > 7. ??? ...118 seconds pass... ???

For completeness: I found a Yamux timeout that could theoretically add **150s** of latency here (since the timeout is hit twice if a client fails over *and* the server it connects to fails over as well): https://github.com/hashicorp/yamux/blob/0bc27b27de87381b58e35462a8dcbd69040279c1/mux.go#L66

Now whether that timeout *should* be that or not is a question for a different issue/PR...

schmichael:
Closing this in favor of #14679

PR #14673, opened 2022-09-23
fingerprint: don't clear Consul/Vault attributes on failure

tgross:
Clients periodically fingerprint Vault and Consul to ensure the server has updated attributes in the client's fingerprint. If the client can't reach Vault/Consul, the fingerprinter clears the attributes and requires a node update. Although this seems like correct behavior so that we can detect intentional removal of Vault/Consul access, it has two serious failure modes:

(1) If a local Consul agent is restarted to pick up configuration changes and the client happens to fingerprint at that moment, the client will update its fingerprint and result in evaluations for all its jobs and all the system jobs in the cluster.

(2) If a client loses Vault connectivity, the same thing happens. But the consequences are much worse in the Vault case because Vault is not run as a local agent, so Vault connectivity failures are highly correlated across the entire cluster. A 15 second Vault outage will cause a new `node-update` evalution for every system job on the cluster times the number of nodes, plus one `node-update` evaluation for every non-system job on each node. On large clusters of 1000s of nodes, we've seen this create a large backlog of evaluations. (See also https://github.com/hashicorp/nomad/pull/14621 for mitigations of this.)

This changeset updates the fingerprinting behavior to keep the last fingerprint if Consul or Vault queries fail. This prevents a storm of evaluations at the cost of requiring a client restart if Consul or Vault is intentionally removed from the client.

### DIFF
--- a/.changelog/14673.txt
+++ b/.changelog/14673.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+fingerprint: Consul and Vault attributes are no longer cleared on fingerprinting failure
+```
--- a/client/fingerprint/consul.go
+++ b/client/fingerprint/consul.go
@@ -75,14 +75,6 @@ func (f *ConsulFingerprint) Periodic() (bool, time.Duration) {
 	return true, 15 * time.Second
 }
 
-// clearConsulAttributes removes consul attributes and links from the passed Node.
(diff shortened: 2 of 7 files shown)


shoenig (approved):
LGTM! We might want to leave a note (upgrade guide?) about how to uninstall Consul/Vault, since the attributes will no longer be automatically cleared (just a Client restart?)

PR #14675, opened 2022-09-23
CSI: ensure initial unpublish state is checkpointed

tgross:
A test flake revealed a bug in the CSI unpublish workflow, where an unpublish
that comes from a client that's successfully done the node-unpublish step will
not have the claim checkpointed if the controller-unpublish step fails. This
will result in a delay in releasing the volume claim until the next GC.

This changeset also ensures we're using a new snapshot after each write to raft,
and fixes two timing issues in test where either the volume watcher can
unpublish before the unpublish RPC is sent or we don't wait long enough in
resource-restricted environments like GHA.

---

Two notes for reviewers:
* This is somewhat related to https://github.com/hashicorp/nomad/pull/14484 but is really a distinct bug, so I've given it its own changelog and it'll need backports.
* This doesn't fix the longstanding issue of Nomad thumbing its nose at the concurrency requirements of the CSI spec, inasmuch that the volumewatcher can kick off after we've checkpointed (which doesn't actually result in bad behavior so long as plugins are mostly spec-compliant themselves, but...). I've got some thoughts about fixing that, but it's not going to land in 1.4.0.

### DIFF
--- a/.changelog/14675.txt
+++ b/.changelog/14675.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+csi: Fixed a bug where a volume that was successfully unmounted by the client but then failed controller unpublishing would not be marked free until garbage collection ran.
+```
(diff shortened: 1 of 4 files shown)


DerekStrickland (approved):
LGTM

PR #14677, opened 2022-09-23
[ui] Service Healthchecks: styles for pseudo-timestamp axis

philrenaud:
image

### DIFF
--- a/ui/app/styles/components/services.scss
+++ b/ui/app/styles/components/services.scss
@@ -95,23 +95,77 @@ table.health-checks {
         width: calc(
           750px - 3rem - 50px
         ); //Sidebar width - padding - table padding
-        height: 20px;
+        height: 12px;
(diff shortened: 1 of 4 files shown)


ui/app/styles/components/services.scss
```diff
@@ -95,23 +95,77 @@ table.health-checks {
...
+              border-top: 1px solid $grey-blue;
+              content: '';
+              position: absolute;
+              left: calc(50% - 1px);
+              margin-left: -6px;
+              top: calc(-50% - 1px);
+            }
+          }
+
+          &:nth-child(8n + 1) > .timestamp {
```
philrenaud, line 159: Only show the timestamp itself every 8th element, beginning with the first one, so as to avoid overlap.

ChaiWithJai (approved):

PR #14679, opened 2022-09-23
docs: write a lot of words about heartbeats

schmichael:
Alternative to #14670

Well if wordcount is what matters I sure win, but I'm pretty sure wordcount is uncorrelated or perhaps even slightly inversely proportional to utility when it comes to docs.

Let me know what you think. If this approach seems good I might do something similar for the `gc` parameters.

Preview: https://nomad-f8oqghj31-hashicorp.vercel.app/docs/configuration/server#client-heartbeats

### DIFF
--- a/website/content/docs/configuration/server.mdx
+++ b/website/content/docs/configuration/server.mdx
@@ -375,7 +375,7 @@ and rescheduled.
 The various heartbeat related parameters allow you to tune the following
 tradeoffs:
 
-- The longer the heartbeat period, the longer a down Client's workload will
+- The longer the heartbeat period, the longer a `down` Client's workload will
   take to be rescheduled.
 - The shorter the heartbeat period, the more likely transient network issues,
   leader elections, and other temporary issues could cause a perfectly
@@ -400,15 +400,15 @@ Other factors modify this base TTL:
   [thundering herd][herd] problem where a large number of clients attempt to
   heartbeat at exactly the same time.
 - [`min_heartbeat_ttl`](#min_heartbeat_ttl) is used as the lower bound to
-  prevent small clusters from needlessly heartbeating extremely quickly.
+  prevent small clusters from sending excessive heartbeats.
 - [`heartbeat_grace`](#heartbeat_grace) is the amount of _extra_ time the
   leader will wait for a heartbeat beyond the base heartbeat.
 - After a leader election all Clients are given up to `failover_heartbeat_ttl`
   to successfully heartbeat. This gives Clients time to discover a functioning
   Server in case they were directly connected to a leader that crashed.
 
-Given the default values for heartbeat parameters, different sized clusters
-will use the following TTLs for the heartbeats. Note that the `Server TTL`
+For example, given the default values for heartbeat parameters, different sized 
+clusters will use the following TTLs for the heartbeats. Note that the `Server TTL`
 simply adds the `heartbeat_grace` parameter to the TTL Clients are given.
 
 | Clients | Client TTL  | Server TTL  | Safe after elections |
@@ -424,8 +424,8 @@ Regardless of size, all clients will have a Server TTL of
 than the maximum Client TTL for your cluster size in order to prevent marking
 live Clients as `down`.
 
-For clusters over 5000 Clients it is recommended to increase
-`failover_heartbeat_ttl` using the following:
+For clusters over 5000 Clients you should increase `failover_heartbeat_ttl` 
+using the following formula:
 
 ```
 (2 * (<number of Clients> / <max_heartbeats_per_second>)) + (10 * <min_heartbeat_ttl>)
--- a/website/content/docs/configuration/server.mdx
+++ b/website/content/docs/configuration/server.mdx
@@ -118,38 +118,24 @@ server {
   example section](#configuring-scheduler-config) for more details
   `default_scheduler_config` was introduced in Nomad 0.10.4.
 
-- `heartbeat_grace` `(string: "10s")` - Specifies the additional time given as a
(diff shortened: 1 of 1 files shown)


shoenig (approved):
LGTM! just the one note

website/content/docs/configuration/server.mdx
```diff
@@ -118,38 +118,24 @@ server {
...
-	of all the heartbeats. This is specified using a label suffix like "30s" or
-	"1h".
-
-  ~> Lowering the `failover_heartbeat_ttl` is a tradeoff as it lowers failure
-  detection time of nodes at the tradeoff of false positives. False positives
-  could cause all clients to stop their allocations if a leadership transition
-  lasts longer than `heartbeat_grace + failover_heartbeat_ttl`.
+- `failover_heartbeat_ttl` `(string: "5m")` - The time by which all Clients
+  must heartbeat after a Server leader election. This is specified using a label
+  suffix like "30s" or "1h". See [below](#client-heartbeats) for details.
```
tgross, line 133: Including the name of the header section helps orient the reader once they land (particularly for folks with non-default text zoom settings for reading assistance):

```suggestion
  suffix like "30s" or "1h". See [Client Heartbeats](#client-heartbeats) below for details.
```

website/content/docs/configuration/server.mdx
```diff
@@ -369,6 +361,88 @@ server {
...
+Nomad Clients periodically heartbeat to Nomad Servers to confirm they are
+operating as expected. Nomad Clients which do not heartbeat in the specified
+amount of time are considered `down` and their allocations are marked as `lost`
+or `disconnected` (if [`max_client_disconnect`][max_client_disconnect] is set)
+and rescheduled.
+
+The various heartbeat related parameters allow you to tune the following
+tradeoffs:
+
+- The longer the heartbeat period, the longer a down Client's workload will
```
tgross, line 378:
```suggestion
- The longer the heartbeat period, the longer a `down` Client's workload will
```

website/content/docs/configuration/server.mdx
```diff
@@ -369,6 +361,88 @@ server {
...
+<number of Clients> / <max_heartbeats_per_second> 
+```
+
+Other factors modify this base TTL:
+
+- A random factor up to `2x` is added to the base TTL to prevent the
+  [thundering herd][herd] problem where a large number of clients attempt to
+  heartbeat at exactly the same time.
+- [`min_heartbeat_ttl`](#min_heartbeat_ttl) is used as the lower bound to
+  prevent small clusters from needlessly heartbeating extremely quickly.
```
tgross, line 403:
```suggestion
- [`min_heartbeat_ttl`](#min_heartbeat_ttl) is used as the lower bound to
  prevent small clusters from sending excessive heartbeats.
```

website/content/docs/configuration/server.mdx
```diff
@@ -369,6 +361,88 @@ server {
...
+- [`heartbeat_grace`](#heartbeat_grace) is the amount of _extra_ time the
+  leader will wait for a heartbeat beyond the base heartbeat.
+- After a leader election all Clients are given up to `failover_heartbeat_ttl`
+  to successfully heartbeat. This gives Clients time to discover a functioning
+  Server in case they were directly connected to a leader that crashed.
+
+Given the default values for heartbeat parameters, different sized clusters
+will use the following TTLs for the heartbeats. Note that the `Server TTL`
+simply adds the `heartbeat_grace` parameter to the TTL Clients are given.
```
tgross, line 412: We might want to specifically call out that these are examples so that it's clear this isn't new information that contradicts the above:

```suggestion
For example, given the default values for heartbeat parameters, different sized 
clusters will use the following TTLs for the heartbeats. Note that the `Server TTL`
simply adds the `heartbeat_grace` parameter to the TTL Clients are given.
```

website/content/docs/configuration/server.mdx
```diff
@@ -369,6 +361,88 @@ server {
...
+| 5000    | 100s - 200s | 110s - 210s | yes                  |
+| 10000   | 200s - 400s | 210s - 410s | NO (see below)       |
+
+Regardless of size, all clients will have a Server TTL of
+`failover_heartbeat_ttl` after a leader election. It should always be larger
+than the maximum Client TTL for your cluster size in order to prevent marking
+live Clients as `down`.
+
+For clusters over 5000 Clients it is recommended to increase
+`failover_heartbeat_ttl` using the following:
```
tgross, line 428: I think we can make this sentence more direct, as it's not all that optional:

```suggestion
For clusters over 5000 Clients you should increase `failover_heartbeat_ttl` 
using the following formula:
```

tgross (approved):
LGTM

website/content/docs/configuration/server.mdx
schmichael: Thanks! Fixed up all the references.

website/content/docs/configuration/server.mdx
```diff
@@ -369,6 +361,88 @@ server {
...
+```
+
+This ensures Clients have some additional time to failover even if they were
+told to heartbeat after the maximum interval.
+
+The actual value used should take into consideration how much tolerance your
+system has for a delay in noticing crashed Clients. A `failover_heartbeat_ttl`
+of 30 minutes may give even the largest cluster ample time to heartbeat after
+an election, but could cause a 30 minute outage for services on crashed Clients
+instead of a 30 second outage if `failover_heartbeat_ttl`.
```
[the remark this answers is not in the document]
schmichael: Ha, good catch. I'm going to rework that whole unwieldy sentence.

PR #9435, opened 2020-11-24
client: always wait 200ms before sending updates

schmichael:
Always wait 200ms before calling the Node.UpdateAlloc RPC to send
allocation updates to servers.

Prior to this change we only reset the update ticker when an error was
encountered. This meant the 200ms ticker was running while the RPC was
being performed. If the RPC was slow due to network latency or server
load and took >=200ms, the ticker would tick during the RPC.

Then on the next loop only the select would randomly choose between the
two viable cases: receive an update or fire the RPC again.

If the RPC case won it would immediately loop again due to there being
no updates to send.

When the update chan receive is selected a single update is added to the
slice. The odds are then 50/50 that the subsequent loop will send the
single update instead of receiving any more updates.

This could cause a couple of problems:

1. Since only a small number of updates are sent, the chan buffer may
   fill, applying backpressure, and slowing down other client
   operations.
2. The small number of updates sent may already be stale and not
   represent the current state of the allocation locally.

A risk here is that it's hard to reason about how this will interact
with the 50ms batches on servers when the servers under load.

A further improvement would be to completely remove the alloc update
chan and instead use a mutex to build a map of alloc updates. I wanted
to test the lowest risk possible change on loaded servers first before
making more drastic changes.

PR #14693, opened 2022-09-26
fingerprint: lengthen Vault check after seen

schmichael:
@tgross Yikes! #14673 made me realize how pathological the Vault fingerprinting was. What do you think about going a step further and backing off fingerprints as long as Vault is available?

-- Original commit message --

Extension of #14673

Once Vault is initially fingerprinted, extend the period since changes should be infrequent and the fingerprint is relatively expensive since it is contacting a central Vault server.

Also move the period timer reset *after* the fingerprint. This is similar to #9435 where the idea is to ensure the retry period starts *after* the operation is attempted. 15s will be the *minimum* time between fingerprints now instead of the *maximum* time between fingerprints.

In the case of Vault fingerprinting, the original behavior might cause the following:

1. Timer is reset to 15s
2. Fingerprint takes 16s
3. Timer has already elapsed so we immediately Fingerprint again

Even if fingerprinting Vault only takes a few seconds, that may very well be due to excessive load and backing off our fingerprints is desirable. The new bevahior ensures we always wait at least 15s between fingerprint attempts and should allow some natural jittering based on server load and network latency.

### DIFF
--- a/client/fingerprint/vault.go
+++ b/client/fingerprint/vault.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	log "github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/nomad/helper"
 	vapi "github.com/hashicorp/vault/api"
 )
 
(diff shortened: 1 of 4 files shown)


tgross (approved):
LGTM! 👍 

At first glance I thought we could add the same large jitter for the `Periodic` method of the Consul fingerprinter, but there's _so_ much less value there because the agent/self endpoint on the Consul doesn't need to talk to the Consul servers. Probably better to keep it so we can respond to local agent config changes more quickly.

Issue #14676, opened 2022-09-23
"cores" parameter in resource stanza not behaving as expected


bjornbyte:
### Nomad version
Output from `nomad version`
Nomad v1.3.5 (1359c2580fed080295840fb888e28f0855e42d50)

### Operating system and Environment details
Windows 11 WSL

### Issue
When attempting to schedule a job with 
```
resources {
  cores = 1
}
```
in the `task` stanza, plan is returning `Dimension cores exhausted on 1 node` even though there are no other jobs running and the client attributes in the UI shows `cpu.numcores | 16`

### Reproduction steps
in windows WSL ubuntu prompt run
`nomad agent -dev`
try to submit the below simple job

#### Expected Result
it should get scheduled

#### Actual Result
```
hello 1 unplaced
Resources exhausted on 1 node
Dimension cores exhausted on 1 node
```

### Job file (if appropriate)
```
job "example" {
  datacenters = ["dc1"]

  group "hello" {
    task "hello" {
      driver = "docker"

      config {
	image   = "redis:3.2"
        command = "/bin/bash"
	args    = ["-c", "echo 'Hello World!' && sleep infinity"]
      }

      resources {
        cores    = 1
        memory = 20
      }
    }
  }
}
```

### Client Node details
I do notice that in the below output, although `Attributes.cpu.numcores` is 16, `Resources.Cores` is zero, so I expect that's the problem here, but not sure why that would be.

```
> curl http://localhost:4646/v1/node/ae67729b-44e7-9bcf-002f-0fcf7ea3351d
{
    "Attributes": {
        "cpu.frequency": "2303",
        "nomad.version": "1.3.5",
        "cpu.totalcompute": "36848",
        "unique.storage.bytesfree": "247538786304",
        "cpu.modelname": "11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz",
        "kernel.version": "5.10.16.3-microsoft-standard-WSL2",
        "kernel.arch": "x86_64",
        "nomad.service_discovery": "true",
        "cpu.numcores": "16",
        "unique.storage.bytestotal": "269490393088",
        "driver.docker": "1",
        "unique.storage.volume": "/dev/sdc",
        "driver.docker.os_type": "linux",
        "unique.hostname": "Byter",
        "os.version": "20.04",
        "unique.cgroup.version": "v1",
        "nomad.revision": "1359c2580fed080295840fb888e28f0855e42d50",
[...]

PR #14694, opened 2022-09-26
fingerprint: add node attr for reserverable cores

schmichael:
Add an attribute for the number of reservable CPU cores as they may differ from the existing `cpu.numcores` due to client configuration or OS support.

Hopefully clarifies some confusion in #14676

### DIFF
--- a/client/fingerprint/cpu.go
+++ b/client/fingerprint/cpu.go
@@ -2,6 +2,7 @@ package fingerprint
 
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/hashicorp/nomad/lib/cpuset"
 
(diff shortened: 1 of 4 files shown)


client/fingerprint/cpu.go
```diff
@@ -80,6 +81,7 @@ func (f *CPUFingerprint) Fingerprint(req *FingerprintRequest, resp *FingerprintR
 			f.logger.Debug("detected reservable cores", "cpuset", reservableCores)
 		}
 	}
+	resp.AddAttribute("cpu.num_reservable_cores", strconv.Itoa(len(reservableCores)))
```
shoenig, line 84: the other `cpu.` attributes contain no underscores, 

```
cpu.num_reservable_cores  = 24
cpu.numcores              = 24
cpu.totalcompute          = 112128
```

what about naming this `cpu.reservablecores` ?

shoenig (approved):

client/fingerprint/cpu.go
schmichael: Good call, thanks. Fixing.

PR #14710, opened 2022-09-26
e2e: convert chroot env unit tests into e2e tests

shoenig:
This PR converts two of our most flakey unit tests into
e2e tests where they are fit much more naturally.


tgross (approved):
LGTM, thanks!

Issue #2158, opened 2017-01-05
Users allowlist


schmichael:
> nomad needs a user.whitelist. having to blacklist hundreds of user accounts to satisfy security reqs is a PITA.

 -- from: https://twitter.com/iKryten/status/817129940005122048

Currently nomad allows denylisting running tasks as certain users (such as root). For locking down a system a allowlist is much more succinct and easier to implement correctly.

Nomad should have a user.allowlist setting to allow specifying exactly what users tasks can run as.

tgross:
Closed by https://github.com/hashicorp/nomad/pull/24343

Issue #12420, opened 2022-03-31
remove deprecated client options for plugin config


tgross:
Pre-0.9 clients used the deprecated client options configuration for plugins. The documentation was removed in https://github.com/hashicorp/nomad/pull/12416 so that we discourage users from using it. But we should be able to remove these options entirely in a major release of Nomad.

This work may also end up blocking https://github.com/hashicorp/nomad/issues/12390 (or at least can be done at the same time).

iwinux:
If these options are removed, how would I disable unwanted task drivers (e.g.: `docker` or `qemu`) for certain client nodes?

tgross:
Hi @iwinux! I just double-checked because I thought we had an `enabled` flag for all the task drivers, but after testing it turns out we only have it for `raw_exec`! So yes, we'll need to add it to all the built-in plugins as well.

By the way, in Nomad 1.3.0 you'll also be able to add task driver allow/deny lists to namespaces, which some folks will find handy for limiting what workloads can happen most of the time without tying their hands in the case where they want to allow for example platform teams the ability to run `raw_exec` workloads.

tgross:
`fingerprint.denylist` also needs to be reimplemented as new config.

PR #14714, opened 2022-09-27
feat(2158): Adding user allowlist logic and test

bneil:
This commit should help with issue #2158

### DIFF
--- a/client/allocrunner/taskrunner/validate_hook.go
+++ b/client/allocrunner/taskrunner/validate_hook.go
@@ -48,10 +48,21 @@ func validateTask(task *structs.Task, taskEnv *taskenv.TaskEnv, conf *config.Con
 		[]string{"user.denylist", "user.blacklist"},
 		config.DefaultUserDenylist,
 	)
+
+	allowedUsers := conf.ReadStringListAlternativeToMapDefault(
+		[]string{"user.allowlist", "user.whitelist"},
+		config.DefaultUserAllowlist,
+	)
+
+	// if you are on the allowlist, you are allowed
+	// if you are not on the allowlist, check the denylist
+	// if you are on the denylist, you are disallowed
 	checkDrivers := conf.ReadStringListToMapDefault("user.checked_drivers", config.DefaultUserCheckedDrivers)
 	if _, driverMatch := checkDrivers[task.Driver]; driverMatch {
-		if _, unallowed := unallowedUsers[task.User]; unallowed {
-			mErr.Errors = append(mErr.Errors, fmt.Errorf("running as user %q is disallowed", task.User))
+		if _, allowed := allowedUsers[task.User]; !allowed {
+			if _, unallowed := unallowedUsers[task.User]; unallowed {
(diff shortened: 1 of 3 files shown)


client/allocrunner/taskrunner/validate_hook.go
```diff
@@ -48,10 +48,21 @@ func validateTask(task *structs.Task, taskEnv *taskenv.TaskEnv, conf *config.Con
...
 	checkDrivers := conf.ReadStringListToMapDefault("user.checked_drivers", config.DefaultUserCheckedDrivers)
 	if _, driverMatch := checkDrivers[task.Driver]; driverMatch {
-		if _, unallowed := unallowedUsers[task.User]; unallowed {
-			mErr.Errors = append(mErr.Errors, fmt.Errorf("running as user %q is disallowed", task.User))
+		if _, allowed := allowedUsers[task.User]; !allowed {
+			if _, unallowed := unallowedUsers[task.User]; unallowed {
+				mErr.Errors = append(mErr.Errors, fmt.Errorf("running as user %q is disallowed", task.User))
+			}
 		}
 	}
```
tgross, line 67: The logic here currently "fails open". Consider the following configuration:

```hcl
client {
  options = {
    "user.allowlist" = "nobody"
  }
}
```

If I try to run a job with `user = "root"`, we'd check the allowlist and not find them, but not find them on the denylist either, so they'd be allowed to go ahead. Whereas an allowlist should _only_ allow those users on the allowlist and not on the denylist.

Likewise, consider the following misconfiguration:

```hcl
client {
  options = {
    "user.allowlist" = "root"
    "user.denylist" = "root,ubuntu"
  }
}
```

In this case, `root` would be allowed even though under a "default deny" approach the denylist should override the allowlist.

client/allocrunner/taskrunner/validate_hook_test.go
```diff
@@ -25,6 +25,11 @@ func TestTaskRunner_Validate_UserEnforcement(t *testing.T) {
 		t.Fatalf("expected error running as root with exec")
 	}
 
+	// Try to run an allowlisted user with exec.
+	config.DefaultUserAllowlist = "allow-fubar"
+	task.User = "allow-fubar"
+	require.NoError(t, validateTask(task, taskEnv, conf))
```
tgross, line 31: The tests added here should cover the problems discussed above.

tgross (changes requested):
Hi @bneil! 

You'll note https://github.com/hashicorp/nomad/issues/2158 is a very old issue. And as it turns out the documentation for the denylist feature was actually removed because it's in the deprecated section of the configuration. But we don't have a replacement for the behavior in the task driver configs yet either. See https://github.com/hashicorp/nomad/issues/12420 for that.

You can fix up the logic here and we could get that merged, but we'd just be tearing the code out again once we clean out those old options. It might be better if added the denylist/allowlist feature directly to the `plugin` configs for `exec` and `java` drivers and then leave this in-place to be removed later.

tgross:
Superceded by https://github.com/hashicorp/nomad/pull/24343

PR #14721, opened 2022-09-27
backport to 1.1.x: CSI: skip node unpublish on GC'd or down nodes

tgross:
https://github.com/hashicorp/nomad/pull/13301 was supposed to have been backported to 1.1.x but there's no backport tag. Weirdly, https://github.com/hashicorp/nomad/pull/13311 refers to the code having been backported, and the test changes from #13301 are already here, just not the code changes. Not sure how that happened but #13301 was one of the PRs impacted by the weirdness around the 1.3.2 vs 1.3.5 release. In any case, this bug should get released in the next 1.1.x.

(2/2 PRs, the other was for 1.2.x)

### DIFF
--- a/.changelog/14720.txt
+++ b/.changelog/14720.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+csi: Fixed a bug where volume claims on lost or garbage collected nodes could not be freed
+```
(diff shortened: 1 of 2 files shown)


shoenig (approved):

lgfa29:
Sorry, late to the party, but just to explain what seems to have happened:

- #13301 didn't get backported
- #13311 references the wrong PR. The test that it removes (`TestServiceSched_CSITopology`) was added in #13274, not #13301

tgross:
Ah! Thanks @lgfa29! 😀

PR #4856, opened 2018-11-08
supported numerical comparison in constrains #4729

MrJacek:

PR #14722, opened 2022-09-27
core: numeric operands comparisons in constraints

### LINKED ISSUE #4729: Constraint numeric comparison
For some reason job with

~~~~
      constraint {
        attribute = "${attr.cpu.numcores}"
        operator  = ">"
        value     = "10"
      }
~~~~

Is placed to machines with `cpu.numcores == 8`.

Why? Even if constraints are not mandatory but merely advisory a job should have been placed to client with 10+ cores, right?

### LINKED ISSUE #14719: Provide method of defining greater than/less than constraints on numeric attributes
There appears to be no method of defining a >, >=. <, <= constraint on a numeric attribute such as memory.totalbytes or cpu.numcores.
I naively created such a constraint and it didn't work as expected. The documentation does say that the comparison is done lexicographically.

Could a means be provided to do a numeric comparison for attributes? Something like a custom driver attribute, or a user metadata attribute could reasonably be defined as numeric and you might want to do a numeric comparison on it.

shoenig:
This PR changes constraint comparisons to be numeric rather than
lexical if both operands are integers or floats.

Inspiration #4856
Closes #4729
Closes #14719

### DIFF
--- a/scheduler/feasible.go
+++ b/scheduler/feasible.go
@@ -12,6 +12,7 @@ import (
 	"github.com/hashicorp/nomad/helper/constraints/semver"
 	"github.com/hashicorp/nomad/nomad/structs"
 	psstructs "github.com/hashicorp/nomad/plugins/shared/structs"
+	"golang.org/x/exp/constraints"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


scheduler/feasible.go
```diff
@@ -850,35 +854,73 @@ func checkAttributeAffinity(ctx Context, operand string, lVal, rVal *psstructs.A
...
+	}
+	if result, ok := checkFloatOrder(operand, left, right); ok {
+		return result
 	}
+	return checkLexicalOrder(operand, left, right)
+}
+
+// checkIntegralOrder compares lVal and rVal as integers if possible, or false otherwise.
+func checkIntegralOrder(op, lVal, rVal string) (bool, bool) {
+	left, lErr := strconv.Atoi(lVal)
```
tgross, line 876: It _probably_ doesn't matter, but this restricts the integer size to a machine-sized int instead of a int64, which means on the small number of 32-bit platforms we support we could end up with different behavior. `strconv.ParseInt(lVal, 10, 64)` would be more specific. (But also, do we care?)

tgross (approved):
LGTM

scheduler/feasible.go
shoenig: I could see this being an issue if someone started using this comparison for `unique.storage.bytesfree`; I'll add the fix

PR #14728, opened 2022-09-27
Post 1.4.0 rc.1 release

DerekStrickland:

PR #14729, opened 2022-09-27
Fix and remove changelog files

DerekStrickland:


lgfa29 (approved):
I missed in the previous PR, but `14551` was removed as well:
https://github.com/hashicorp/nomad/pull/14728/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edL54

PR #14739, opened 2022-09-29
test: fix data race in taskrunner restore test

schmichael:
The test was using the mock driver and in-memory client statestore. Since the in-memory client statestore just stores pointers, two problems were occurring:

1. The race was due to the task state being shared between the new and old task runners since an in-memory statestore was used.

2. The mock driver's restore code relies on the unexported TaskConfig.rawDriverConfig field to be maintained between restarts. This can't happen with a "real" statestore that must encode and decode state: the rawDriverConfig field is expected to be dropped!

I'm not sure it's worth fixing the mock driver's restore path since restoring an in-memory driver between invocations of a Nomad agent can't work in the same way as other driver's which run external processes (eg exec) or communicate with external daemons (eg docker).

Overall these changes should make the test more realistic and reliable at the cost of requiring bash to be available.

### DIFF
--- a/client/allocrunner/taskrunner/task_runner_test.go
+++ b/client/allocrunner/taskrunner/task_runner_test.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


schmichael:
Converted to draft to inspect CI failure.

PR #14742, opened 2022-09-29
client: protect user lookups with global lock

shoenig:
This PR updates Nomad client to always do user lookups while holding
a global process lock. This is to prevent concurrency unsafe implementations
of NSS, but still enabling NSS lookups of users (i.e. cannot not use osusergo).


tgross (approved):
LGTM

Issue #14744, opened 2022-09-29
[ui] Give users an option to run a job from URL, or by dragging/uploading a job spec


philrenaud:
Add URL and drag/upload options here:

![image](https://user-images.githubusercontent.com/713991/193062106-50301e32-9879-4ae9-a09c-8033fa51ea76.png)

PR #14748, opened 2022-09-29
Fetch job spec from URL

philrenaud:
Contributes to #14744 
image


philrenaud:
Going to close this PR after weighing the balance of adding more interactivity to this page and the relative low likelihood of use,

PR #14749, opened 2022-09-29
template: apply splay value on change_mode script

### LINKED ISSUE #14734: After rendering the template, the script runs constantly restarts
Nomad v1.3.5 
I'm trying to set up nginx to reload the config generated from a template:
```
{{- $forbiddenNames := parseJSON `["serviceslanding","webconfigurator"]` -}}
{{ range services -}}
  {{if .Tags | contains "api" -}}
    location /{{ .Name }}/ {
      #proxy_pass http://{{ env `NOMAD_IP_nginxhttp` }}:{{ scratch.Get .Name }}/;
      proxy_pass http://{{ env `NOMAD_IP_nginxhttp` }}:{{ key (print .Name "/http.port") }}/;
      include /etc/nginx/conf.d/proxy.conf;
      proxy_set_header X-Forwarded-Path /{{ .Name }};
   }
  {{- end -}}
{{- end -}}
```
To do this, I set the change_mode = "script" value in the job and add the change_script section.
```
# services
      artifact {
        source      = "http://consul.service.consul:8500/v1/kv/nginx/services.conf.tpl?raw=true"
        ${artifact_headers}
      }
      template {
        source        = "local/services.conf.tpl"
        destination   = "local/services.conf"
        change_mode   = "script"
        change_script  {
          command       = "/usr/sbin/nginx"
          args          = ["-s", "reload"]
          timeout       = "5s"
          fail_on_error = false
        }
      }
```

When the template changes, the script is executed. However, after that it starts to run continuously.

image

At the same time, the config modification time does not change. This means that the script restarts are not due to a change in the config.

I can't understand why this might be happening. Maybe there are thoughts about this? Maybe I missed something?

Thank you!

lgfa29:
Previously, the splay timeout was only applied if a template re-render caused a restart or a signal action. The `change_mode = "script"` was running after the `if restart || len(signals) != 0` check, so it was invoked at all times.

This change refactors the logic so it's easier to notice that new `change_mode` options should start only after `splay` is applied.

Closes #14734

### DIFF
--- a/.changelog/14749.txt
+++ b/.changelog/14749.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+template: Fixed a bug where the `splay` timeout was not being applied when `change_mode` was set to `script`.
+```
--- a/client/allocrunner/taskrunner/template/template.go
+++ b/client/allocrunner/taskrunner/template/template.go
@@ -463,52 +463,65 @@ func (tm *TaskTemplateManager) onTemplateRendered(handledRenders map[string]time
 		handling = append(handling, id)
 	}
 
-	if restart || len(signals) != 0 {
-		if splay != 0 {
-			ns := splay.Nanoseconds()
-			offset := rand.Int63n(ns)
-			t := time.Duration(offset)
+	shouldHandle := restart || len(signals) != 0 || len(scripts) != 0
+	if !shouldHandle {
+		return
+	}
 
-			select {
-			case <-time.After(t):
-			case <-tm.shutdownCh:
-				return
-			}
-		}
+	// Apply splay timeout to avoid applying change_mode too frequently.
+	if splay != 0 {
+		ns := splay.Nanoseconds()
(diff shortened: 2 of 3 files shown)


shoenig (approved):
LGTM! But what happened to the GHA runners :thinking:

lgfa29:
Oh, weird. I think it skipped https://github.com/hashicorp/nomad/pull/14749/commits/a33f0963419e3d61950aa1cdfe071b46bdf8065e because it's just a `.changelog` change (https://github.com/hashicorp/nomad/blob/main/.github/workflows/test-core.yaml#L10).

It ran for https://github.com/hashicorp/nomad/pull/14749/commits/9d888394cea181fc59ce66edec7fb59bff5c015e: https://github.com/hashicorp/nomad/actions/runs/3154058295

pkazmierczak (approved):
:shipit:

Issue #13731, opened 2022-07-12
document "Deployments" in the Concepts section


wimax-grapl:
### Proposal

I'd like to see a minimal description of a Deployment on https://www.nomadproject.io/docs/concepts/architecture that compare-contrasts it to, say, an Allocation. 

### Use-cases

My use case is the following: I was trying to reverse-engineer the job-success monitoring that goes on when you `nomad job run`, which was introduced here:
https://github.com/hashicorp/nomad/pull/10661
It relies on the Deployment api, which is not well documented publicly. Or rather, you'll get stuff like ["The /deployment endpoints are used to query for and interact with deployments."](https://www.nomadproject.io/api-docs/deployments), which skirts telling you what concept this represents! 

While it looks like it's a collection of Allocation IDs, I'm genuinely just unsure. 

### Attempted Solutions

this is a documentation request

schmichael:
Great idea! We're hoping to have a little docs sprint soon, and this is on our TODO list.

wimax-grapl:
Glad to hear!

tgross:
Notes from dupe #10581:
> The deployment logic is currently hand-wavy and interactions: canaries, deployment timeouts, node health, etc. Create concept documentation for deployments that contains diagrams showing the state transitions that occur during deployments.

tgross:
Fixed by https://github.com/hashicorp/nomad/pull/14750

Issue #13468, opened 2022-06-23
document expected evaluation counts


tgross:
The number of evaluations generated by Nomad is an important metric for operators to understand its performance. We should improve the documentation on the expected number of evaluations we create and what this means for how many nodes are processed by the scheduler.

Some example events that create Evaluations:
* For one job registration, a single Evaluation is created in raft.
* For one allocation failure, a single Evaluation is created in raft.
* For each chunk of a deployment, a single Evaluation is created in raft.
* For each run of a periodic job, a single Evaluation is created in raft.
* For each periodic server garbage collection event, a single Evaluation is created in raft.
* For one node registration event (including missed heartbeats):
  * One Evaluation is created in raft for each `system` job in the cluster, plus...
  * One Evaluation is created for each non-system job that has an allocation on that node.
  * ref [`node_endpoint.go#L1445-L1548`](https://github.com/hashicorp/nomad/blob/v1.3.1/nomad/node_endpoint.go#L1445-L1548)

For each Evaluation in raft:
* A scheduler worker must dequeue it from the eval broker on the leader.
* For each allocation in a batch job or service job without `spread`, the scheduler will evaluate (process) nodes until it finds 2 feasible nodes to score.
* For each allocation in a batch job or service job with `spread`, the scheduler will evaluate (process) a number of nodes equal to the task group count, or 100, whichever is less. 
* For a system job, the scheduler will evaluate (process) all nodes on the cluster.
* If there are changes to make, the scheduler worker must submit a plan to the plan applier on the leader. No-op evaluations are not submitted as plans.
* If the scheduler can't place all the allocations, one new Evaluation will be created in the "blocked" state and those will start from the top once dequeued by the next pass thru a scheduler.

Some concrete examples:  
* If you have 90 system jobs and 10k nodes existing, and add 100 nodes:
  * 90 * 100 = 9000 Evaluations will be created in raft (each one written to disk, replicated to a quorum of followers, etc.)
  * for each of those Evaluations created in raft, 10100 nodes will be evaluated ("processed") by the scheduler (909M total, all in-memory operations on the scheduler workers)
  * we'd expect between 90 and 9000 plans submitted, depending on how close together node updates land
* If you have 90 system jobs and 10k nodes existing, and add 100 nodes, but after placement all the allocations fail once b/c of external dependencies:
  * 90 * 100 = 9000 initial Evaluations + 9000 Evaluations for the failed allocs
[...]

Issue #8588, opened 2020-08-03
Need documentation about different events that trigger evaluations 


abhip:
The results returned when we call evaluation API (When I call evaluation API (https://www.nomadproject.io/api-docs/evaluations) contain a field called "TriggeredBy" which contain various values such as deployment-watcher, job-register and node-update . Some of these values are self -explanatory but others are not. 

I did not find any documentation about these fields in the website or any comment in the code explaining about them. 

Here is where they defined in the code - https://github.com/hashicorp/nomad/blob/940817aeb52e740ef5d35f1be925f2c61b8ecd0f/nomad/structs/structs.go#L9377

It would be very helpful if documentation is provided about them

tgross:
Fixed by https://github.com/hashicorp/nomad/pull/14750

PR #14750, opened 2022-09-29
internals documentation with diagrams

tgross:
Fixes https://github.com/hashicorp/nomad/issues/13731 
Fixes https://github.com/hashicorp/nomad/issues/13468
Fixes https://github.com/hashicorp/nomad/issues/8588

This changeset adds new architecture internals documents to the contributing guide. These are intentionally here and not on the public-facing website because the material is not required for operators and includes a lot of diagrams that we can cheaply maintain with mermaid syntax but would involve art assets to have up on the main site that would become quickly out of date as code changes happen and be extremely expensive to maintain. However, these should be suitable to use as points of conversation with expert end users.

Included:
* A description of Evaluation triggers and expected counts, with examples. ([preview](https://github.com/hashicorp/nomad/blob/a01c4a8e61169993f8d841fe934738d4ddbae9fd/contributing/architecture-eval-triggers.md))
* A description of Evaluation states and implicit states. This is taken from an internal document in our team wiki. ([preview](https://github.com/hashicorp/nomad/blob/a01c4a8e61169993f8d841fe934738d4ddbae9fd/contributing/architecture-eval-states.md))
* A description of how writing to the State Store works. This is taken from a diagram I put together a few months ago for internal education purposes. ([preview](https://github.com/hashicorp/nomad/blob/a01c4a8e61169993f8d841fe934738d4ddbae9fd/contributing/architecture-state-store.md))
* A description of Evaluation lifecycle, from registration to running Allocations. This is mostly lifted from @lgfa29's amazing mega-diagram, but broken into digestible chunks and without multi-region deployments, which I'd like to cover in a future doc. ([preview](https://github.com/hashicorp/nomad/blob/a01c4a8e61169993f8d841fe934738d4ddbae9fd/contributing/architecture-eval-lifecycle.md))

Also includes adding Deployments to our public-facing glossary. ([preview](https://nomad-bvo1eisdp-hashicorp.vercel.app/docs/concepts/architecture#deployment))

### DIFF
--- a/contributing/architecture-eval-lifecycle.md
+++ b/contributing/architecture-eval-lifecycle.md
@@ -19,8 +19,8 @@ have been written to the state store of the server nodes.
 * Note: parameterized or periodic batch jobs don't create an Evaluation at
   registration, only once dispatched.
 * Note: The scheduler is very fast! This means that once the CLI gets a
-  response it can immediately start making blocking queries to get information
-  about the next steps, but those are happening _asynchronously_.
+  response it can immediately start making queries to get information
+  about the next steps.
 * Note: This workflow is different for Multi-Region Deployments (found in Nomad
   Enterprise). That will be documented separately.
 
@@ -102,7 +102,7 @@ steps:
   creates new Allocations in this step. But note that _these Allocations are not
   yet persisted to the state store_.
 * Feasibility Check: For each Allocation it needs, the scheduler iterates over
-  Nodes until it finds 2 Nodes that match the Allocation's resource requirements
+  Nodes until it finds up to 2 Nodes that match the Allocation's resource requirements
   and constraints.
   * Note: for system jobs or jobs with with `spread` blocks, the scheduler has
     to check all Nodes.
--- a/contributing/architecture-eval-lifecycle.md
+++ b/contributing/architecture-eval-lifecycle.md
@@ -10,6 +10,13 @@ Allocations on the clients. The process can be broken into 4 parts:
 * [Deployment Watcher](#deployment-watcher)
 * [Client Allocs](#client-allocs)
 
+Note that in all the diagrams below, writing to the State Store is considered
+atomic. This means the Nomad leader has replicated to all the followers, and all
+the servers have applied the raft log entry to their local FSM. So long as
+`raftApply` returns without an error, we have a guarantee that all servers will
+be able to retrieve the entry from their state store at some point in the
+future.
+
 ## Job Registration
 
 Creating or updating a Job is a _synchronous_ API operation. By the time the
@@ -34,8 +41,7 @@ sequenceDiagram
     participant cli as CLI
     participant httpAPI as HTTP API
     participant leaderRpc as Leader RPC
-    participant leaderFSM as Leader State Store
-    participant workerFSM as Worker State Store
+    participant stateStore as State Store
 
     user ->> cli: nomad job run
     activate cli
@@ -44,25 +50,20 @@ sequenceDiagram
 
     activate leaderRpc
 
-    leaderRpc ->> leaderFSM: write Job
-    activate leaderFSM
-    leaderFSM ->> workerFSM: replicate Job to followers
+    leaderRpc ->> stateStore: write Job
+    activate stateStore
+    stateStore -->> leaderRpc: ok
+    deactivate stateStore
 
-    workerFSM -->> leaderFSM: ok
-    leaderFSM -->> leaderRpc: ok
-    deactivate leaderFSM
+    leaderRpc ->> stateStore: write Evaluation
 
-    leaderRpc ->> leaderFSM: write Evaluation
+    activate stateStore
+    stateStore -->> leaderRpc: ok
 
-    activate leaderFSM
-    leaderFSM ->> workerFSM: replicate Evaluation to followers
-    workerFSM -->> leaderFSM: ok
-    leaderFSM -->> leaderRpc: ok
+    Note right of stateStore: EvalBroker.Enqueue
+    Note right of stateStore: (see Scheduling below)
 
-    Note right of leaderFSM: EvalBroker.Enqueue
-    Note right of leaderFSM: (see Scheduling below)
-
-    deactivate leaderFSM
+    deactivate stateStore
 
     leaderRpc -->> httpAPI: Job.Register response
     deactivate leaderRpc
@@ -151,28 +152,24 @@ been persisted to the state store.
 sequenceDiagram
 
     participant leaderRpc as Leader RPC
-    participant leaderFSM as Leader State Store
+    participant stateStore as State Store
     participant planner as Planner
     participant broker as Eval Broker
-    participant workerFSM as Worker State Store
     participant sched as Scheduler
 
-    leaderRpc ->> leaderFSM: write Evaluation (see above)
-    leaderFSM ->> broker: EvalBroker.Enqueue
+    leaderRpc ->> stateStore: write Evaluation (see above)
+    stateStore ->> broker: EvalBroker.Enqueue
     activate broker
     broker -->> broker: enqueue eval
-    broker -->> leaderFSM: ok
+    broker -->> stateStore: ok
     deactivate broker
 
     sched ->> broker: Eval.Dequeue (blocks until work available)
     activate sched
     broker -->> sched: Evaluation
 
-    Note right of workerFSM: Schedulers run on all servers
-    Note right of workerFSM: So this may query the leader state store directly
-
-    sched ->> workerFSM: query to get job and cluster state
-    workerFSM -->> sched: results
+    sched ->> stateStore: query to get job and cluster state
+    stateStore -->> sched: results
 
     sched -->> sched: Reconcile (how many allocs are needed?)
     deactivate sched
@@ -193,12 +190,10 @@ sequenceDiagram
             Note left of sched: Not enough room! (But we can submit a partial plan)
             sched ->> leaderRpc: Eval.Upsert (blocked)
             activate leaderRpc
-            leaderRpc ->> leaderFSM: write Evaluation
-            activate leaderFSM
-            leaderFSM ->> workerFSM: replicate Evaluation to followers
-            workerFSM -->> leaderFSM: ok
-            leaderFSM -->> leaderRpc: ok
-            deactivate leaderFSM
+            leaderRpc ->> stateStore: write Evaluation
+            activate stateStore
+            stateStore -->> leaderRpc: ok
+            deactivate stateStore
             leaderRpc --> sched: ok
             deactivate leaderRpc
         end
@@ -217,12 +212,10 @@ sequenceDiagram
 
             planner ->> leaderRpc: Allocations.Upsert + Deployment.Upsert
             activate leaderRpc
-            leaderRpc ->> leaderFSM: write Allocations and Deployment
-            activate leaderFSM
-            leaderFSM ->> workerFSM: replicate Allocations and Deployment to followers
-            workerFSM -->> leaderFSM: ok
-            leaderFSM -->> leaderRpc: ok
-            deactivate leaderFSM
+            leaderRpc ->> stateStore: write Allocations and Deployment
+            activate stateStore
+            stateStore -->> leaderRpc: ok
+            deactivate stateStore
             leaderRpc -->> planner: ok
             deactivate leaderRpc
             planner -->> sched: ok
@@ -239,12 +232,10 @@ sequenceDiagram
 
     sched ->> broker: Eval.Ack (Eval.Nack if failed)
     activate broker
-    broker ->> leaderFSM: complete Evaluation
-    activate leaderFSM
-    leaderFSM ->> workerFSM: replicate Evaluation to followers
-    workerFSM -->> leaderFSM: ok
-    leaderFSM -->> broker: ok
-    deactivate leaderFSM
+    broker ->> stateStore: complete Evaluation
+    activate stateStore
+    stateStore -->> broker: ok
+    deactivate stateStore
     broker -->> sched: ok
     deactivate broker
     deactivate sched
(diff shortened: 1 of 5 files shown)


contributing/architecture-eval-lifecycle.md
```diff
@@ -0,0 +1,326 @@
...
+
+```
+
+## Scheduling
+
+A goroutine on the Nomad leader called the Eval Broker runs a continuous
+blocking query for Evaluations on the state store. Those Evaluations are
+enqueued in the broker.
+
+Schedulers are goroutines running on all server nodes. Typically this will be
```
lgfa29, line 76: I'm not sure if it's worth mentioning this distinction here, but something that confused me early on (and I think it confused @pkazmierczak as well) is that that what we call "scheduler" in the code (like `GenericScheduler` and `SystemScheduler`) is actually just a short-lived struct that is instantiated on-demand per eval. The long-lived goroutines are the workers.

So when you're reading the code for these schedulers, they keep mutating themselves without a lock 😅

contributing/architecture-eval-lifecycle.md
```diff
@@ -0,0 +1,326 @@
...
+If the job is a `service` job, the scheduler will also create (or update) a
+Deployment. When the scheduler determines the number of Allocations to create,
+it examine the job's [`update`][] block. Only the number of Allocations needed
+to complete the next phase of the update will be created in a given
+allocation. The Deployment is used by the deployment watcher (see below) to
+monitor the health of allocations and create new Evaluations to continue the
+update.
+
+If the scheduler cannot place all Allocations, it will create a new Evaluation
+in the "blocked" state and submit it to the leader. The Eval Broker will
```
lgfa29, line 105:
```suggestion
in the `blocked` state and submit it to the leader. The Eval Broker will
```

contributing/architecture-eval-lifecycle.md
```diff
@@ -0,0 +1,326 @@
...
+
+* The scheduler took a snapshot of cluster state at the start of its work, so
+  that state may have changed in the meantime.
+* Schedulers run concurrently across the cluster, so they may generate
+  conflicting plans (particularly on heavily-packed clusters).
+
+The leader processes the plan in the plan applier. If the plan is valid, the
+plan applier will write the Allocations (and Deployment) update to the state
+store. If not, it will reject the plan. (The plan submit process is the second
+green box in the sequence diagram below.)
```
lgfa29, line 121:
```suggestion
store. If not, it will reject the plan and the scheduler will try to create a new plan with a refreshed state. If the scheduler fails to submit a valid plan too many times it submits a `blocked` Evaluation that is triggered by `max-plan-attempts` type. (The plan submit process is the second
green box in the sequence diagram below.)
```

I don't know if this helps to explain the `plan for node rejected` class of problems. Feel free to ignore if it's too much detail.

Another detail here is that the plan applier can have a partial apply, and so the scheduler only needs to retry what's left. But again, it may be too much to add here.

contributing/architecture-eval-lifecycle.md
```diff
@@ -0,0 +1,326 @@
...
+
+Once the scheduler has a response from the leader, it will tell the Eval Broker
+to Ack the Evaluation (if it successfully submitted the plan) or Nack the
+Evaluation (if it failed to do so).
```
lgfa29, line 125:
```suggestion
Evaluation (if it failed to do so) so that another scheduler can try processing it.
```

contributing/architecture-eval-lifecycle.md
```diff
@@ -0,0 +1,326 @@
...
+sequenceDiagram
+
+    participant leaderRpc as Leader RPC
+    participant leaderFSM as Leader State Store
+    participant planner as Planner
+    participant broker as Eval Broker
+    participant workerFSM as Worker State Store
+    participant sched as Scheduler
+
+    broker ->> leaderFSM: blocking query for new evals
```
lgfa29, line 141: Is it a blocking query or do evals get enqueued directly using [`EvalBroker.Enqueue`](https://github.com/hashicorp/nomad/blob/main/nomad/eval_broker.go#L189-L194)?

contributing/architecture-eval-lifecycle.md
```diff
@@ -0,0 +1,326 @@
...
+    broker -->> broker: enqueue eval
+    deactivate broker
+
+    sched ->> broker: Eval.Dequeue (blocks until work available)
+    activate sched
+    broker -->> sched: Evaluation
+
+
+    Note right of workerFSM: Schedulers run on all servers
+    Note right of workerFSM: So this may query the leader state store directly
```
lgfa29, line 153: I tried to make these fit into the yellow box using `<br />` but Mermaid is not having it 😅

contributing/architecture-eval-lifecycle.md
```diff
@@ -0,0 +1,326 @@
...
+
+```
+
+## Client Allocs
+
+Once the plan applier has persisted Allocations to the state store (with an
+associated Node ID), they become available to get placed on the client. Clients
+_pull_ new allocations (and changes to allocations), so a new allocation will be
+in the `pending` state until it's been pulled down by a Client and the
+allocation has been instantiated.
```
lgfa29, line 296: Do clients report allocation status back to the servers in the heartbeat messages?

lgfa29 (approved):
These are fantastic! Great job 👏

contributing/architecture-eval-triggers.md
```diff
@@ -0,0 +1,274 @@
...
+  `blocked` state to be processed later when node updates arrive.
+* **preemption**: When Allocations are preempted, the plan applier creates 1
+  Evaluation for each Job that has been preempted.
+* **job-scaling**: Scaling a Job will result in 1 Evaluation created, plus any
+  follow-up Evaluations associated with scheduling, planning, or deployments.
+* **max-disconnect-timeout**: When an Allocation is in the `unknown` state for
+  longer than the [`max_client_disconnect`][] window, the scheduler will create
+  1 Evaluation.
+* **reconnect**: When a Node in the `disconnected` state reconnects, Nomad will
+  create 1 Evaluation.
```
schmichael, line 81:
```suggestion
* **reconnect**: When a Node in the `disconnected` state reconnects, Nomad will
  create 1 Evaluation per job with an allocation on the reconnected Node.
```

I think? We actually create 1 eval per *allocation* but the batching in `node_endpoint.go` dedupes evals by job.

Interesting side note: when a down down reconnects I think it will call Node.Register and Node.UpdateAlloc in nondeterministic order, and if they occur within 50ms of each other the node endpoints batching will drop subsequent evals for a job from whichever RPC was processed last. I think this means when reconnecting an alloc you may get the following evals:

1. TriggeredBy: node-update (Register called first, UpdateAlloc called within 50ms later)
2. TriggeredBy: reconnect (UpdateAlloc called first, Register called within 50ms later)
3. Both node-update and reconnect evals because they occurred more than 50ms apart (or at least were not in the same batch)

contributing/architecture-eval-lifecycle.md
tgross: Yeah it's... imperfect. :grin:

contributing/architecture-eval-lifecycle.md
tgross: Good catch, I was confusing the [restoreEvals](https://github.com/hashicorp/nomad/blob/v1.4.0-beta.1/nomad/leader.go#L571-L597) that happens on leader transition with [upserting evals](https://github.com/hashicorp/nomad/blob/v1.4.0-beta.1/nomad/leader.go#L571-L597). Will fix!

contributing/architecture-state-store.md
```diff
@@ -0,0 +1,129 @@
...
+
+* The `fsm.Apply` functions must be deterministic over their inputs for a given
+  state. You can never generate random IDs or assign wall-clock timestamps in
+  the state store. These values must be provided as parameters from the RPC
+  handler.
+
+* Every object you read from the state store must be copied before it can be
+  mutated, because mutating the object modifies it outside the raft
+  workflow. The result can be servers having inconsistent state, transactions
+  breaking, or even server panics.
```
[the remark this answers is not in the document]
tgross: Love it!

contributing/architecture-eval-lifecycle.md
tgross: Yeah, lemme see if I can add some language around that here.

contributing/architecture-eval-lifecycle.md
tgross: It's a separate message `Node.UpdateAlloc` but we should include that here so that it's obvious how the deployment watcher gets its source of information.

contributing/architecture-eval-states.md
```diff
@@ -0,0 +1,219 @@
...
+evaluation takes through the leader, worker, and plan applier. But it doesn't
+cover in any detail the various `Evaluation.Status` values, or where the
+`PreviousEval`, `NextEval`, or `BlockedEval` ID pointers are set.
+
+The state diagram below describes the transitions between `Status` values as
+solid arrows. The dashed arrows represent when a new evaluation is created. The
+parenthetical labels on those arrows are the `TriggeredBy` field for the new
+evaluation.
+
+The states are:
```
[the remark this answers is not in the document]
tgross: You're right, and I should continue that into the "quasi-state" discussion below as well.

contributing/architecture-eval-lifecycle.md
```diff
@@ -0,0 +1,326 @@
...
+
+Creating or updating a Job is a _synchronous_ API operation. By the time the
+response has return to the API consumer, the Job and an Evaluation for that Job
+have been written to the state store of the server nodes.
+
+* Note: parameterized or periodic batch jobs don't create an Evaluation at
+  registration, only once dispatched.
+* Note: The scheduler is very fast! This means that once the CLI gets a
+  response it can immediately start making blocking queries to get information
+  about the next steps, but those are happening _asynchronously_.
```
shoenig, line 23:
```suggestion
* Note: The scheduler is very fast! This means that once the CLI gets a
  response it can immediately start making queries to get information
  about the next steps.
```

Do we need to know the async details about how that works, here?

contributing/architecture-eval-lifecycle.md
```diff
@@ -0,0 +1,326 @@
...
+blocking query for Evaluations on the state store. Those Evaluations are
+enqueued in the broker.
+
+Schedulers are goroutines running on all server nodes. Typically this will be
+one per core on followers and 1/4 that number on the leader. The schedulers poll
+for Evaluations from the Eval Broker with the `Eval.Dequeue` RPC.
+
+Once a scheduler has an evaluation, it takes a snapshot of that server node's
+state store so that it has a consistent current view of the cluster state. The
+scheduler executes 3 main steps:
```
shoenig, line 82:
```suggestion
Once a scheduler has an evaluation, it takes a snapshot of that server node's
state store so that it has a constant current view of the cluster state. The
scheduler executes 3 main steps:
```

"consistent" implies cluster consistency, which the snapshot cannot provide

contributing/architecture-eval-lifecycle.md
```diff
@@ -0,0 +1,326 @@
...
+
+* Reconcile: compare the cluster state and job specification to determine what
+  changes need to be made -- starting or stopping Allocations. The scheduler
+  creates new Allocations in this step. But note that _these Allocations are not
+  yet persisted to the state store_.
+* Feasibility Check: For each Allocation it needs, the scheduler iterates over
+  Nodes until it finds 2 Nodes that match the Allocation's resource requirements
```
shoenig, line 89:
```suggestion
  Nodes until it finds up to 2 Nodes that match the Allocation's resource requirements
```

contributing/architecture-eval-lifecycle.md
tgross: Yeah good call, that's minutia.

contributing/architecture-eval-lifecycle.md
tgross: Done

contributing/architecture-eval-lifecycle.md
tgross: Done

tgross:
(@mikenomitch I originally flagged you for review but really that was intended to be informational as we've got some field folks anticipating these docs, I'll remove you as a reviewer but just FYI 😀 )

contributing/architecture-eval-lifecycle.md
```diff
@@ -0,0 +1,354 @@
...
+    httpAPI ->> leaderRpc: Job.Register
+
+    activate leaderRpc
+
+    leaderRpc ->> leaderFSM: write Job
+    activate leaderFSM
+    leaderFSM ->> workerFSM: replicate Job to followers
+
+    workerFSM -->> leaderFSM: ok
+    leaderFSM -->> leaderRpc: ok
```
schmichael, line 52: *Not a blocker!*

I think it might be beneficial to wrap this up in some single stage called `raftApply` or similar. It could then be reused in any graph we have a `raftApply` call.

The distinction is subtle, but I think worth making:

> `raftApply` returning without an error guarantees all servers _will be able to retrieve this object from their statestore at some point in the future._

The distinction is important because when a follower dequeues an evaluation it _may not be able to retrieve that evaluation from its statestore despite having applied the raft log!_ This is why the worker immediately ensures its statestore has the eval's WaitIndex: https://github.com/hashicorp/nomad/blob/v1.4.0-rc.1/nomad/worker.go#L415


Also I think making folks think about `raftApply` as an atomic operation is helpful. Sure there are network roundtrips in there to consider, but the whole point of `raftApply` is that if it succeeds a quorum of servers is guaranteed to have committed the log atomically.

tgross: Had a sidebar with @schmichael about the notation to make that work, and we're going to collapse the workerFSM/leaderFSM into a single "raft" participant. That'll let us refer to the raft workflow outside this already huge diagram 😀

tgross: Done

PR #14753, opened 2022-09-29
Revert removing deprecated client options docs

tgross:
This reverts PR #12416 and commit 6668ce022ac561f75ad113cc838b1fb786f11f79.

While the driver options are well and truly deprecated, this documentation also covers features like `fingerprint.denylist` that are not available any other way. Let's revert this until #12420 is ready.


lgfa29 (approved):
It would be good to add `backport/1.1.x` as well to restore in the versioned docs as well.

DerekStrickland (approved):
:shipit:

tgross:
> It would be good to add backport/1.1.x as well to restore in the versioned docs as well.

Done. At some point we should have a discussion about how far we show versioned docs for "unsupported" versions, but :+1:

PR #14759, opened 2022-09-30
cleanup: rename Equals to Equal for consistency

shoenig:
This PR aims to make equality comparisons of our structs consistent with modern Go convention. In the past we have used `Equals`, but the cool kids are all using `Equal`. 

Important example - `go-cmp` has special behavior to detect and use `Equal` when present.
https://github.com/google/go-cmp#package-for-equality-of-go-values
> Types that have an Equal method may use that method to determine equality. This allows package authors to determine the equality operation for the types that they define.

Likewise, the standard library is naming things with `Equal`
https://pkg.go.dev/bytes#Equal

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -1566,7 +1566,7 @@ func (c *Client) updateNodeFromFingerprint(response *fingerprint.FingerprintResp
 		response.Resources.Networks = updateNetworks(
 			response.Resources.Networks,
 			newConfig)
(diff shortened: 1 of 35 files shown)


shoenig:
Will do backports after 1.4 GA to avoid confusion

pkazmierczak (approved):

PR #9142, opened 2020-10-21
Support HCLv2 for Nomad jobs

notnoop:
Support HCLv2 syntax for Nomad jobs.  HCLv2 offers significant expressibility by adding variables and expression supports, and this PR builds the foundation to support that.  In order to experiment, I enabled some functions and dynamic block support to illustrate the power of HCLv2 and to experiment with new syntax in development, but I expect to iterate on these before cutting the final 1.0 release.

## UX and User Visible Changes

This PR updates the CLI and job render API to use HCLv2 parser for Nomad job files.  For the initial draft, I adopted the pPacker Configuration Functions](https://www.packer.io/docs/from-1.5/functions).  Using functions can enable concise HCLv2 jobs - consider the following template blocks:

```hcl
      template {
        destination = "local/motd.txt"
        data        = file("motd.txt")
      }

      template {
        destination = "local/config.json"
        data        = jsonencode({ data = [for s in ["val1", "val2", "val3"] : upper(s)] })
      }
```

Here, Nomad will load the local `motd.txt` and insert it as the template data for the first template.  The second template illustrates the HCLv2 functions, for-looping, as well as function capability: The second template data will be `{"data":["VAL1", "VAL2","VAL3"]}`.

Also, it supports having some input variables to be passed as CLI arguments: `${vars.region}` appearing in job file will be populated by values passed as CLI argument (e.g. `--var region=eu-west-1`).

The exact functions, template, and CLI will iterated on, probably after this PR is merged.

### Backward Compatible Changes

The HCLv2 syntax closely resembles the HCLv1 syntax, so we anticipate that operators can upgrade smoothly. However, there are some notable differences.

In particular, HCLv2 is stricter about block syntax and schema than HCLv1.  HCLv2 requires that block attributes to be unquoted, and doesn't accept map attribute syntax as an alternative to blocks:

```hcl
# valid in HCLv1 and HCLv2
env {
  key1 = "val1"
}

# valid in HCLv1 but NOT HCLv2
env {
  "key1" = "val1"
}

## this is technically a valid HCLv2 but cannot be accepted with above syntax
env = {
  key1 = val1
}
```

To ease adoption, we are making the following concessions:
1. The job run API remain unmodified: frameworks and tools that submit jobs to the job run API endpoint function can function unmodified.  This PR doesn't change the json format the API accept
[...]

PR #14767, opened 2022-09-30
Docs(job-specification/periodic): Add enabled toggle

pop:
This is probably undocumented for a reason, but the `enabled` toggle in the `periodic` stanza is very useful so I figured I try adding it to the docs.

The feature has been secretly avaliable since #9142 and was [called out in that PR as being a dubious addition](https://github.com/hashicorp/nomad/pull/9142/files#r510132150), only added to avoid regressions in the update to hcl2.

The use case for disabling a periodic job in this way is to prevent it from running without modifying the schedule. Ideally Nomad would make it more clear that this was the case, and allow you to force a run of the job, but even with those rough edges I think users would benefit from knowing about this toggle.

### DIFF
--- a/website/content/docs/job-specification/periodic.mdx
+++ b/website/content/docs/job-specification/periodic.mdx
@@ -48,6 +48,10 @@ consistent evaluation when Nomad spans multiple time zones.
   be parsable by Golang's
   [LoadLocation](https://golang.org/pkg/time/#LoadLocation).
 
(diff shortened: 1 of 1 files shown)


tgross:
Preview link: https://nomad-ai2ax6iwl-hashicorp.vercel.app/docs/job-specification/periodic#enabled

PR #14775, opened 2022-10-02
enhancement: UpdateTask when Task is waiting for ShutdownDelay

### LINKED ISSUE #14170: emit TaskEvent for task stops blocked by `shutdown_delay`
When a task shuts down with a `shutdown_delay` set, there's no Task Event that indicates we're waiting for the task to shut down. It can be useful for debugging what would otherwise look like a "stuck" task to have this in the output of `nomad alloc status` (and the corresponding pages on the UI).

We've already got a debug-level log entry at the right spot in [`task_runner.go#L931`](https://github.com/hashicorp/nomad/blob/v1.3.3/client/allocrunner/taskrunner/task_runner.go#L931), so it would be a matter of creating a new Task Event and calling `UpdateState` for it (similar to [`task_runner.go#L1143-L1145`](https://github.com/hashicorp/nomad/blob/v1.3.3/client/allocrunner/taskrunner/task_runner.go#L1143-L1145)).

DarthBenro008:
Signed-off-by: Hemanth Krishna <hkpdev008@gmail.com>

- Fixes #14170 
- Adds a new Task Event named `TaskWaitingShuttingDownDelay ` (Open to suggestions)

### DIFF
--- a/client/allocrunner/taskrunner/task_runner.go
+++ b/client/allocrunner/taskrunner/task_runner.go
@@ -982,6 +982,10 @@ func (tr *TaskRunner) handleKill(resultCh <-chan *drivers.ExitResult) *drivers.E
 	if delay := tr.Task().ShutdownDelay; delay != 0 {
 		tr.logger.Debug("waiting before killing task", "shutdown_delay", delay)
 
+		ev := structs.NewTaskEvent(structs.TaskWaitingShuttingDownDelay).
+			SetDisplayMessage(fmt.Sprintf("Waiting for shutdown_delay of %s before killing the task.", delay))
+		tr.UpdateState(structs.TaskStatePending, ev)
+
 		select {
 		case result := <-resultCh:
 			return result
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -8153,6 +8153,10 @@ const (
 
 	// TaskClientReconnected indicates that the client running the task disconnected.
 	TaskClientReconnected = "Reconnected"
+
+	// TaskWaitingShuttingDownDelay indicates that the task is waiting for
+	// shutdown delay before being TaskKilled
+	TaskWaitingShuttingDownDelay = "Waiting for shutdown delay"
 )
 
 // TaskEvent is an event that effects the state of a task and contains meta-data
--- a/.changelog/14775.txt
+++ b/.changelog/14775.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+client: Added a TaskEvent when task shutdown is waiting on shutdown_delay
+```


cesarvspr:
this looks good @DarthBenro008 ! 

Would you please provide instructions on how can I test it?

cesarvspr (commented):
see https://github.com/hashicorp/nomad/pull/14775#issuecomment-1264695659

DarthBenro008:
@cesarvspr , you can create a [Task with shutdown delay](https://www.nomadproject.io/docs/job-specification/task#shutdown_delay) and then run `nomad alloc status` . It should show that the Task is waiting for the shutdown delay

nomad/structs/structs.go
```diff
@@ -8153,6 +8153,10 @@ const (
 
 	// TaskClientReconnected indicates that the client running the task disconnected.
 	TaskClientReconnected = "Reconnected"
+
+	// TaskWaitingShuttingDownDelay indicates that the task is waiting for
+	// shutdown delay before being TaskKilled
+	TaskWaitingShuttingDownDelay = "Task waiting for shutdown delay"
```
tgross, line 8159: If we look at the output of these, this name is a little wide and pushes everything to the right:

```
Recent Events:
Time                       Type                             Description
2022-10-03T14:25:34-04:00  Task waiting for shutdown delay  Waiting for shutdown delay before killing the task.
2022-10-03T14:25:34-04:00  Killing                          Sent interrupt. Waiting 5s before force killing
2022-10-03T14:25:15-04:00  Started                          Task started by client
2022-10-03T14:25:15-04:00  Task Setup                       Building Task Directory
2022-10-03T14:25:14-04:00  Received                         Task received by client
```

Screen Shot 2022-10-03 at 2 38 26 PM

The event description can have more information about how we're waiting specifically on the `shutdown_delay` field and for how long, so for this event title, what if we trimmed it down to:

```suggestion
	TaskWaitingShuttingDownDelay = "Waiting for shutdown"
```

That also gives us some wiggle room to add new "waiting for shutdown" events with different sources/descriptions in the future.

client/allocrunner/taskrunner/task_runner.go
```diff
@@ -982,6 +982,10 @@ func (tr *TaskRunner) handleKill(resultCh <-chan *drivers.ExitResult) *drivers.E
 	if delay := tr.Task().ShutdownDelay; delay != 0 {
 		tr.logger.Debug("waiting before killing task", "shutdown_delay", delay)
 
+		ev := structs.NewTaskEvent(structs.TaskWaitingShuttingDownDelay).
+			SetDisplayMessage("Waiting for shutdown delay before killing the task.")
```
tgross, line 986: For the display message we usually try to provide a little more information than is in the event type name. What if we told the user how long they'll be waiting:

```suggestion
			SetDisplayMessage(
				fmt.Sprintf("Waiting for shutdown_delay of %s before killing the task", delay))			
```

tgross (changes requested):
Hi @DarthBenro008, this is great!

I've left some suggestions this could be made even better. Also, can you add a changelog entry file? It'll be a file at `.changelog/14775.txt` that's in the format like [this one](https://github.com/hashicorp/nomad/blob/main/.changelog/14742.txt). For the text, it can be something like:

```
client: Added a TaskEvent when task shutdown is waiting on shutdown_delay
```

This just missed the window for Nomad 1.4.0 GA, but once this is merged you can expect this to ship in the next regular release of Nomad. Thanks so much!

DarthBenro008:
Hi @tgross, thanks a lot for the detailed review! 😃 . I have made the necessary amendments!

tgross (approved):
LGTM! Thanks @DarthBenro008! This should ship in the next regular patch release of Nomad

Issue #11757, opened 2021-12-30
Please add `nomad fmt` command


nahsi:
Please provide command `fmt` to format nomad jobs analogous to `terraform fmt` and `packer fmt`.

Right now I'm using `hclfmt` from [hcl repo](https://github.com/hashicorp/hcl/tree/main/cmd/hclfmt).

To install it I use this command `go install github.com/hashicorp/hcl/v2/cmd/hclfmt@latest`.

jrasell:
Hi @nahsi and thanks for raising this request. I believe this addition would be very useful!

rounakdatta:
Hi @jrasell @mikenomitch is the `nomad fmt` command expected to be just a usage of `hclfmt` or would we like to re-implement the formatting specifically for the Nomad specification?

danishprakash:
@jrasell I'm not entirely sure but I guess this should involve using `tfexec` to invoke `hclfmt` internally and overwrite the source file? I've seen something similar happen over at `terraform-ls` for terraform fmt. If that sounds like the right direction, I'd like to work on this.

mikenomitch:
Hey @rounakdatta, yeah I think just using `hclfmt` would be fine!

If you base the implementation off of [Packer's code](https://github.com/hashicorp/packer/blob/main/command/fmt.go#L16), that could be a really good starting point if you'd like to contribute.

@danishprakash, I think anything that relies on TF is going to be a no-go from our perspective. While we of course love TF, we don't want that as a hard dependency for Nomad.

danishprakash:
@mikenomitch you're right, I think I thought we could do something similar to how we do it in terraform-ls for the formatting even from the client but I now realize, as the name suggests, `tfexec` is in fact built for executing `terraform` commands. The link you shared from Packer looks more aligned and reasonable. I'd like to give this a shot if no one else has started working on this already. Thanks

rounakdatta:
Alright @mikenomitch , I've started looking into Packer's implementation. It's clear, and I think I'll be able to make some progress tonight/tomorrow. @danishprakash I was willing to try my first Nomad contribution through this issue 🙂

danishprakash:
Sure @rounakdatta, please feel free to take this up!

sharmaansh21:
@rounakdatta Are you working on this ?

Trojan295:
Hello, if there is nobody working on that right now, I would like to implement the `nomad fmt`.

[...]

PR #14779, opened 2022-10-03
feat: Add nomad fmt

Trojan295:
This PR adds the `nomad fmt` command, which can be used to format and check Nomad configuration and jobspec files.

It formats the files according to the HCL syntax using [hashicorp/hcl](https://github.com/hashicorp/hcl/). It does not validate, if the files are correct Nomad or Jobspec files.

Closes https://github.com/hashicorp/nomad/issues/11757


command/fmt.go
```diff
@@ -0,0 +1,177 @@
...
+	with .nomad and .hcl extensions in the directory.
+
+Format Options:
+
+	-check
+		Check if the files are valid HCL files. If not, exit status of the command
+		will be 1 and the incorrect files will not be formatted.
+
+	-overwrite=false
+		Print the formatted files to stdout, instead of overwriting the file content.
```
tgross, line 82: Let's call this `write` to match `terraform fmt` (ref https://developer.hashicorp.com/terraform/cli/commands/fmt#write-false).

command/fmt.go
```diff
@@ -0,0 +1,177 @@
...
+	flags := f.Meta.FlagSet(f.Name(), FlagSetClient)
+	flags.Usage = func() { f.Ui.Output(f.Help()) }
+	flags.BoolVar(&fmtArgs.Check, "check", false, "")
+	flags.BoolVar(&fmtArgs.Overwrite, "overwrite", true, "")
+
+	if err := flags.Parse(args); err != nil {
+		return 1
+	}
+
+	fmtArgs.Paths = flags.Args()
```
tgross, line 119: As my colleague @mikenomitch noted in https://github.com/hashicorp/nomad/issues/11757#issuecomment-1266090490, let's also allow this to accept `-` to read files from stdin.

command/fmt.go
```diff
@@ -0,0 +1,177 @@
...
+	return 0
+}
+
+func (*FormatCommand) Help() string {
+	helpText := `
+Usage: nomad fmt [flags] paths ...
+
+	Formats Nomad agent configuration and job file to a canonical format.
+	If a path is a directory, it will recursively format all files
+	with .nomad and .hcl extensions in the directory.
```
tgross, line 73: `terraform fmt` doesn't do this by default and has a separate [-recursive](https://developer.hashicorp.com/terraform/cli/commands/fmt#recursive) option. We should probably do the same.

command/fmt_test.go
```diff
@@ -0,0 +1,106 @@
...
+	ui := cli.NewMockUi()
+
+	cmd := &FormatCommand{
+		Meta: Meta{Ui: ui},
+	}
+
+	bytes, err := os.ReadFile("testdata/fmt/bad-nomad.hcl")
+	require.NoError(t, err)
+
+	f, err := os.CreateTemp("", "nomad-fmt-")
```
tgross, line 66: This leaves temp files on the test host when we're done. You can use [`t.TempDir`](https://pkg.go.dev/testing#T.TempDir) to get a directory that'll automatically get cleaned up afterwards.

tgross (changes requested):
Hi @Trojan295! This is great!

I've left some comments that can tighten this up and make it more consistent with both `terraform fmt` and the rest of the Nomad code base. Two more things:

* You can add a changelog file. You'll make that at `.changelog/14779.txt` and it'll look like https://github.com/hashicorp/nomad/blob/main/.changelog/14742.txt. 
* There's documentation at [./website/content/docs/commands](https://github.com/hashicorp/nomad/tree/main/website/content/docs/commands). You can take a crack at mimicking the files in the top of that directory, but if that's too overwhelming just let me know and I can knock that out pretty quickly once this is ready to merge.

Thanks again!

command/fmt.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+func (*FormatCommand) AutocompleteArgs() complete.Predictor {
+	return complete.PredictFiles("*")
+}
+
+func (*FormatCommand) AutocompleteFlags() complete.Flags {
+	return complete.Flags{
+		"-check":           complete.PredictNothing,
+		"-overwrite=false": complete.PredictNothing,
```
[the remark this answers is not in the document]
Trojan295: My idea behind this was, that by default the `overwrite` defaults to `true`, so someone would provide the flag to change this flag to `false`. That's why I think it's easier, if this autocompletes to `false` already.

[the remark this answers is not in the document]
tgross: Ah yeah, I misread the docstring here.

website/content/docs/commands/fmt.mdx
```diff
@@ -0,0 +1,58 @@
+---
```
tgross, line 1: I forgot that we need to add this page to the navbar for it to be rendered. If you add this bit of JSON:

```
      {
        "title": "fmt",
        "path": "commands/fmt"
      },
```

between the "commands/eval" and "commands/jobs" at https://github.com/hashicorp/nomad/blob/main/website/data/docs-nav-data.json#L431-L432, that'll allow our preview build for the docs to work.

tgross (changes requested):
Looking great @Trojan295! Just a few last items to fix up and we'll be able to land this one.

Trojan295:
Thanks @tgross for the review! I will update this PR tomorrow. :)

command/testdata/fmt/nomad.in.hcl
```diff
@@ -0,0 +1,8 @@
+server {
```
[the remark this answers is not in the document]
Trojan295: What do you think about ignoring the `./command/testdata` directory from the `make hclfmt`?
```
   -o -path './command/testdata' -prune \
```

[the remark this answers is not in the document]
tgross: Eh, that'll work.

tgross (approved):
Looks great, @Trojan295! Thanks for sticking with it, I know PRs for new commands can feel like a whole lot of nitpicking 😀 

CircleCI seems to have gotten confused about the PR and can't run the build. So I just rebased this branch on `main` to get that moving again. Once that finishes, we can get this merged. ~(I'll probably have to do that tomorrow morning, as I'm close to end-of-day here.)~ _Done_

tgross:
@Trojan295 this has been merged and a backport bot will merge it to the 1.4.x release branch for the next patch release of Nomad . If that bot can't figure things out and requests a review from you, don't worry about it and I'll fix it in the morning; it's one of those fiddly process things we try not to expose to contributors but the bot is working thru a bit of a queue at the moment and I won't be there to immediately catch it. 😀

Trojan295:
> Looks great, @Trojan295! Thanks for sticking with it, I know PRs for new commands can feel like a whole lot of nitpicking 😀

Wasn't so bad :). I just had a though, that it would be nice, if some of those docs/CLI help could be generated. E.g. we have the help in `mdx` and this gets automatically added to the website and also the CLI help is generated out of it.

tgross:
> I just had a though, that it would be nice, if some of those docs/CLI help could be generated.

Yes! I know a few folks have been tinkering around with better CLI generation internally for that reason.

Issue #14785, opened 2022-10-03
build: bootstrap target fails due to invalid cert


desmana:
### Nomad version

Not applicable.

### Operating system and Environment details

Fedora Linux 5.19.11-200.fc36.x86_64

### Issue

An error occures when trying to compile nomad from source.

`deps` target contains reference to gophers.dev which is currently returning certificate mismatch error.

### Reproduction steps

Run this snippet from official documentation using a clean system (no installed Go packages):

```sh
git clone https://github.com/hashicorp/nomad.git
cd nomad
make bootstrap
```

#### Expected Result

Bootstrap completes without error.

#### Actual Result

```
go: gophers.dev/cmds/hc-install/cmd/hc-install@v1.0.2: unrecognized import path "gophers.dev/cmds/hc-install/cmd/hc-install": https fetch: Get "https://gophers.dev/cmds/hc-install/cmd/hc-install?go-get=1": x509: certificate is valid for cpanel.ctzsupport.duckdns.org, cpcalendars.ctzsupport.duckdns.org, cpcontacts.ctzsupport.duckdns.org, ctzsupport.duckdns.org, mail.ctzsupport.duckdns.org, webdisk.ctzsupport.duckdns.org, webmail.ctzsupport.duckdns.org, www.ctzsupport.duckdns.org, not gophers.dev
```

tgross:
Hi @desmana it looks like this is a known issue for hc-install: https://github.com/hashicorp/hc-install/issues/70 But I'm not sure why that gophers.dev URL is in use in the first place except that the CLI hasn't actually been stabilized. We got that version in https://github.com/hashicorp/nomad/pull/12937 so that we were using my colleague @shoenig's fork at the time. But that's all been merged at this point.

`go install github.com/hashicorp/hc-install/cmd/hc-install@4487b02cbcbb92204e3416cef9852b6ad44487b21` should get the job done, and I'm going to add that to our makefile.

tgross:
I've opened https://github.com/hashicorp/nomad/pull/14786 with what should fix it.

desmana:
Hi,

thanks for your prompt response. The revision in your comment seems to be invalid, but it looks like simply using HEAD is not going to be complete fix anyway.

FYI `tools/go.mod` also refers to a package from this currently problematic source: `gophers.dev/pkgs/ignore`. This should not impact installation, but `make tidy` is currently failing with the same error:

```sh
$ make tidy
==> Tidy up submodules
[...]

PR #14789, opened 2022-10-04
deps: remove gophers.dev dependency

tgross:
Fixes https://github.com/hashicorp/nomad/issues/14785 by removing the dependency on `gophers.dev/pkg/ignore`. This package registry has a broken cert. The package is really our own @shoenig's https://github.com/shoenig/ignore, but the package is declared with the gophers.dev domain so we can't fix it directly (and he's on vacation :grin:). We're using a one-liner function from it, so to get all the tools build working again, just remove it and we can have a discussion about that when @shoenig gets back.


jrasell (approved):
LGTM :D

Issue #14654, opened 2022-09-22
`Ports` table should always show.


ChaiWithJai:
### Nomad version
Output from `nomad version 1.3.5`

### Issue
The `Ports` table should always show.

### Replay of the Bug
[Watch the Replay](https://app.replay.io/recording/allocation-ig-bridge-demoingress-group0-nomad--677d0c42-4762-4f20-8bdd-64aa5529a101)

#### Expected Result
`Ports` table always shows.

#### Actual Result
`Ports` tables sometimes shows.

ChaiWithJai:
@lgfa29 just opened an issue regarding this. I'll take a peak into this on Monday and we'll aim to have a patch out for `1.3.5.1`(?) not sure if I referenced the version number properly, but you know what I mean.

PR #14798, opened 2022-10-04
refact: upgrade Promise.then to async/await

ChaiWithJai:
In the process of debugging [14654](https://github.com/hashicorp/nomad/issues/14654). This will not resolve the issue, but part of the process of cleaning things up as I see them.

### DIFF
--- a/ui/app/routes/allocations/allocation.js
+++ b/ui/app/routes/allocations/allocation.js
@@ -40,7 +40,7 @@ export default class AllocationRoute extends Route.extend(WithWatchers) {
     try {
       // Preload the job for the allocation since it's required for the breadcrumb trail
       const allocation = await super.model(...arguments);
-      const jobId = allocation.belongsTo('job').id();
+      const jobId = allocation?.belongsTo('job').id();
       const getJob = this.store.findRecord('job', jobId);
       const getNamespaces = this.store.findAll('namespace');
       await Promise.all([getJob, getNamespaces]);
--- a/ui/app/routes/allocations/allocation.js
+++ b/ui/app/routes/allocations/allocation.js
@@ -36,18 +36,18 @@ export default class AllocationRoute extends Route.extend(WithWatchers) {
     }
   }
 
-  model() {
-    // Preload the job for the allocation since it's required for the breadcrumb trail
-    return super
-      .model(...arguments)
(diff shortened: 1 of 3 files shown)


ui/app/routes/allocations/allocation.js
```diff
@@ -36,18 +36,18 @@ export default class AllocationRoute extends Route.extend(WithWatchers) {
...
-          .findRecord('job', jobId)
-          .then(() => this.store.findAll('namespace')) // namespaces belong to a job and are an asynchronous relationship so we can peak them later on
-          .then(() => allocation);
-      })
-      .catch(notifyError(this));
+  async model() {
+    try {
+      // Preload the job for the allocation since it's required for the breadcrumb trail
+      const allocation = await super.model(...arguments);
+      const jobId = allocation.belongsTo('job').id();
```
philrenaud, line 43: Should we consider a defensive `allocation?.belongsTo` here, now that we're not encapsulated within a return object?

philrenaud (approved):

lgfa29 (approved):

ui/app/routes/allocations/allocation.js
```diff
@@ -36,18 +36,18 @@ export default class AllocationRoute extends Route.extend(WithWatchers) {
...
-          .findRecord('job', jobId)
-          .then(() => this.store.findAll('namespace')) // namespaces belong to a job and are an asynchronous relationship so we can peak them later on
-          .then(() => allocation);
-      })
-      .catch(notifyError(this));
+  async model() {
+    try {
+      // Preload the job for the allocation since it's required for the breadcrumb trail
+      const allocation = await super.model(...arguments);
+      const jobId = allocation?.belongsTo('job').id();
```
ChaiWithJai, line 43: @philrenaud based on the test suite, I believe that we actually need to throw an error if we don't find the allocation so using the [elvis operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), will fail a test and not propagate the error message we're expecting. Will revert this.

ChaiWithJai: Meh.... there's more to the story than just this comment though... slight change that using the `try/catch` block and `await` syntax is causing something funky.

PR #14807, opened 2022-10-05
website: add autoscaler agent enable_debug config detail.

jrasell:
I was going to add detail on the pprof endpoints, but it seems we do not have this data for Nomad either. I therefore thought this would be a good docs followup which would be useful to both Nomad and the Nomad Autoscaler.


lgfa29:
Ops, sorry I missed this. I ended up documenting it here https://github.com/hashicorp/nomad/pull/14941 so I'm going to close this one 😅

PR #14809, opened 2022-10-05
ui: remove `RSVP` instances in routes for `async/await`

ChaiWithJai:
This PR updates the model hooks to move away from using `RSVP` to organize asynchronous code, and use the `async/await` syntactic sugar instead.

### DIFF
--- a/ui/app/routes/clients.js
+++ b/ui/app/routes/clients.js
@@ -1,6 +1,5 @@
 import { inject as service } from '@ember/service';
 import Route from '@ember/routing/route';
-import RSVP from 'rsvp';
 import WithForbiddenState from 'nomad-ui/mixins/with-forbidden-state';
 import notifyForbidden from 'nomad-ui/utils/notify-forbidden';
 import classic from 'ember-classic-decorator';
@@ -14,10 +13,14 @@ export default class ClientsRoute extends Route.extend(WithForbiddenState) {
     return this.get('system.leader');
   }
 
-  model() {
-    return RSVP.hash({
-      nodes: this.store.findAll('node'),
-      agents: this.store.findAll('agent'),
-    }).catch(notifyForbidden(this));
+  async model() {
+    try {
+      return {
+        nodes: await this.store.findAll('node'),
+        agents: await this.store.findAll('agent'),
+      };
+    } catch (e) {
+      notifyForbidden(this)(e);
+    }
   }
 }
--- a/ui/app/routes/clients.js
+++ b/ui/app/routes/clients.js
@@ -15,9 +15,14 @@ export default class ClientsRoute extends Route.extend(WithForbiddenState) {
 
   async model() {
     try {
+      const [nodes, agents] = await Promise.all([
+        this.store.findAll('node'),
+        this.store.findAll('agent'),
+      ]);
+
       return {
-        nodes: await this.store.findAll('node'),
-        agents: await this.store.findAll('agent'),
+        nodes,
+        agents,
       };
     } catch (e) {
       notifyForbidden(this)(e);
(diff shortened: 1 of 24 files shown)


ui/app/routes/clients.js
```diff
@@ -14,10 +13,14 @@ export default class ClientsRoute extends Route.extend(WithForbiddenState) {
...
-    return RSVP.hash({
-      nodes: this.store.findAll('node'),
-      agents: this.store.findAll('agent'),
-    }).catch(notifyForbidden(this));
+  async model() {
+    try {
+      return {
+        nodes: await this.store.findAll('node'),
+        agents: await this.store.findAll('agent'),
+      };
```
deanmarano, line 21: This is a change in behavior. Since this is using two `await`s, we're running these requests synchronously, not simultaneously.

I would still use RSVP's `hash` function for this (since you want both `findAll`s to run at the same time) and await on the hash, rather than each promise.

Also, it was there before your change, but it strikes me as odd that we're catching the error here and not in the [Route.error](https://guides.emberjs.com/release/routing/loading-and-error-substates/#toc_the-error-event). Might be a larger refactor than you're going for with just the `async/await`.

ui/app/routes/jobs/index.js
```diff
@@ -19,13 +18,15 @@ export default class IndexRoute extends Route.extend(
...
-    });
+  async model(params) {
+    try {
+      return {
+        jobs: await this.store.query('job', { namespace: params.qpNamespace }),
+        namespaces: await this.store.findAll('namespace'),
+      };
+    } catch (e) {
+      notifyForbidden(this)(e);
+    }
```
fivetanley, line 29: This migration definitely works in the sense that it returns the same data. However, the difference here between `RSVP.hash` or `Promise.all` is that by adding `await`, we reduce the parallelism here because the `await` will resolve sequentially, rather than running both requests at the same time. 

For example, the request tree _before_ this change looked like:

```
model
  ---> jobs
  ---> namespace
```

The data flow after the change using `await` in both keys now looks like this:

```
model
  ---> jobs
           ---> namespaces
```

In other words, users have to wait a little longer for stuff because they have to wait for jobs to resolve before getting namespaces, rather than fetching in parallel.

Rewriting the code to use `Promise.all` isn't as syntactically convenient, but it does allow us to make both requests at the same time and preserve the parallelism we got with `RSVP.hash`. What do you think about this suggestion?

```suggestion
  async model(params) {
    try {
      const [jobs, namespace] = await Promise.all([
        this.store.query('job', { namespace: params.qpNamespace }),
        this.store.findAll('namespace')
       ]);
       
     return { jobs, namespaces };
    } catch (e) {
      notifyForbidden(this, e);
    }
```

ui/tests/acceptance/task-logs-test.js
```diff
@@ -75,9 +75,13 @@ module('Acceptance | task logs', function (hooks) {
       run.cancelTimers();
     }, 500);
 
-    await click('button.logs-sidebar-trigger');
+    const taskButton = document.querySelectorAll(
+      'button.logs-sidebar-trigger'
+    )[1];
+    await click(taskButton);
```
ChaiWithJai, line 81: This test should be using `data-test` attributes to ensure that the correct Task is being selected. However, that's outside of the scope of this PR.

The test, in its previous form was selecting the wrong task which is not a concern of this test:

image

ui/app/routes/jobs/index.js
ChaiWithJai: OK, this is great feedback made the update!

ui/app/routes/clients.js
ChaiWithJai: That's such a good call out! We break pretty significantly from the Ember way on many things. Unsure if we're doing something funky with errors where we use the `notifyError` utility along with a `mixin` to update the controller to have an error. But that's a can of worms that I don't dare to open.

PR #14811, opened 2022-10-05
test: simplify overlap job placement logic

### LINKED ISSUE #14806: e2e: test Overlap consistent failure
Test failure:
```
Failed
=== RUN   TestOverlap
    overlap_test.go:71: 1500/2000 retries reached for github.com/hashicorp/nomad/e2e/overlap.TestOverlap (err=timed out before an allocation was found for overlap6f928495)
    overlap_test.go:71: timeout: timed out before an allocation was found for overlap6f928495
--- FAIL: TestOverlap (24.85s)
```

schmichael:
Trying to fix #14806

Both the previous approach as well as this one worked on e2e clusters I spun up.

The old approach added a constraint to both the initial job and the followup job on a specific node. This would fail when a Windows node was selected and fixed in #14780. I have no idea why #14780 didn't fit it entirely though. It worked in my own e2e cluster.

But I realized I could simplify the test and make it more robust at the same time! Instead of picking a node before registering the job: let the initial job get placed, and then ensure the subsequent job has a constraint on that placement's node!

### DIFF
--- a/e2e/overlap/overlap_test.go
+++ b/e2e/overlap/overlap_test.go
@@ -20,54 +20,24 @@ func TestOverlap(t *testing.T) {
 	nomadClient := e2eutil.NomadClient(t)
 	e2eutil.WaitForLeader(t, nomadClient)
 
-	// Wait for at least 1 feasible node to be ready and get its ID
-	var node *api.Node
-	testutil.Wait(t, func() (bool, error) {
-		nodesList, _, err := nomadClient.Nodes().List(nil)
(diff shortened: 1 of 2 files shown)


schmichael:
It passed last night which make me think your offline suggestion that the failure was timing related was right. I'll make the timing match e2eutils or use e2eutils when possible.

schmichael:
After quite a few days passing the old e2e test version failed again. Merging this as-is to see if it squashes the flakiness.

PR #14818, opened 2022-10-06
e2e: fixes the ordering on greater than checks within spread test.

jrasell:
I think this occurred as the `must` and `require` libs use the opposite order for performing greater than checks.


tgross (approved):
LGTM

Issue #14819, opened 2022-10-06
Nomad server 1.3.5 isn't able to join 1.4.0


freef4ll:
### Nomad version
Nomad running on Linux is 1.4.0 and another is 1.3.5, which panics upon trying to upgrade:

### Operating system and Environment details

1.4.0:
```
# uname -a
Linux test 5.4.0-126-generic #142-Ubuntu SMP Fri Aug 26 12:15:55 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux
```

1.3.5 from brew:
```
$ uname -a
Darwin laptop.local 21.6.0 Darwin Kernel Version 21.6.0: Wed Aug 10 14:28:23 PDT 2022; root:xnu-8020.141.5~2/RELEASE_ARM64_T6000 arm64
```

### Issue

Trying to bring up a cluster causes the 1.3.5 to panic:

```
    2022-10-06T12:05:56.238+0300 [INFO]  nomad.raft: updating configuration: command=AddVoter server-id=259e58a9-7df4-9e68-6217-c0602b5784cf server-addr=192.168.66.7:4647 servers="[{Suffrage:Voter ID:a63aa225-36ec-2644-98b7-dd902f5ccc72 Address:192.168.66.1:4647} {Suffrage:Voter ID:259e58a9-7df4-9e68-6217-c0602b5784cf Address:192.168.66.7:4647}]"
    2022-10-06T12:10:19.447+0300 [WARN]  nomad.raft: failed to contact quorum of nodes, stepping down
    2022-10-06T12:10:19.480+0300 [INFO]  nomad.raft: entering follower state: follower="Node at 192.168.66.1:4647 [Follower]" leader-address= leader-id=
    2022-10-06T12:10:19.480+0300 [INFO]  nomad.raft: aborting pipeline replication: peer="{Nonvoter 259e58a9-7df4-9e68-6217-c0602b5784cf 192.168.66.7:4647}"
    2022-10-06T12:10:19.628+0300 [INFO]  nomad: cluster leadership lost
    2022-10-06T12:10:19.628+0300 [ERROR] worker: failed to dequeue evaluation: worker_id=2f5d738b-bff9-0e16-8a62-9adac20068ac error="eval broker disabled"
    2022-10-06T12:10:19.628+0300 [ERROR] worker: failed to dequeue evaluation: worker_id=69995186-c93e-57c7-7ef3-9934aac44bb0 error="eval broker disabled"
[...]

PR #14821, opened 2022-10-06
fix panic from keyring raft entries being written during upgrade

tgross:
Fixes https://github.com/hashicorp/nomad/issues/14819

During an upgrade to Nomad 1.4.0, if a server running 1.4.0 becomes the leader before one of the 1.3.x servers, the old server will crash because the keyring is initialized and writes a raft entry.

Wait until all members are on a version that supports the keyring before initializing it.

### DIFF
--- a/nomad/leader.go
+++ b/nomad/leader.go
@@ -294,10 +294,7 @@ func (s *Server) establishLeadership(stopCh chan struct{}) error {
 	schedulerConfig := s.getOrCreateSchedulerConfig()
 
 	// Create the first root key if it doesn't already exist
-	err := s.initializeKeyring()
-	if err != nil {
-		return err
-	}
+	go s.initializeKeyring(stopCh)
 
 	// Initialize the ClusterID
 	_, _ = s.ClusterID()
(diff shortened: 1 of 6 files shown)


jrasell (approved):
LGTM.

I tested the patch locally through a previously failing scenario which didn't cause the non-1.4.0 servers to panic when a 1.4.0 server become leader.

```
$ nomad server members                                                                                                                     
Name            Address       Port  Status  Leader  Raft Version  Build      Datacenter  Region
server1.global  192.168.0.14  4648  alive   true    3             1.4.0-dev  dc1         global
server2.global  192.168.0.14  5648  alive   false   3             1.3.5      dc1         global
server3.global  192.168.0.14  6648  alive   false   3             1.4.0-dev  dc1         global
```

nomad/leader.go
```diff
@@ -1966,12 +1963,13 @@ func (s *Server) getOrCreateSchedulerConfig() *structs.SchedulerConfiguration {
...
 
+var minVersionKeyring = version.Must(version.NewVersion("1.4.0"))
+
 // initializeKeyring creates the first root key if the leader doesn't
 // already have one. The metadata will be replicated via raft and then
 // the followers will get the key material from their own key
 // replication.
-func (s *Server) initializeKeyring() error {
-
+func (s *Server) initializeKeyring(stopCh <-chan struct{}) error {
```
jrasell, line 1972: This function now gets called in a go routine, so do we want to remove the error return and log any error messages?

tgross: Good call.

tgross: Done in [3845b9c](https://github.com/hashicorp/nomad/pull/14821/commits/3845b9cf8b88e934ff4aa524f42d8af1fbf37071)

nomad/leader.go
```diff
@@ -1966,43 +1963,79 @@ func (s *Server) getOrCreateSchedulerConfig() *structs.SchedulerConfiguration {
...
-func (s *Server) initializeKeyring() error {
+func (s *Server) initializeKeyring(stopCh <-chan struct{}) {
+
+	logger := s.logger.Named("keyring")
 
 	store := s.fsm.State()
 	keyMeta, err := store.GetActiveRootKeyMeta(nil)
 	if err != nil {
-		return err
+		logger.Error("failed to get active key: %v", err)
```
schmichael, line 1979: Is there anything else we can/should say here? If it's an ignorable part of an upgrade process let's drop it to warn.

nomad/leader.go
```diff
@@ -1966,43 +1963,79 @@ func (s *Server) getOrCreateSchedulerConfig() *structs.SchedulerConfiguration {
...
+		members := s.serf.Members()
+		for _, member := range members {
+			build := member.Tags["build"]
+			memberVersion := version.Must(version.NewVersion(build))
+			if memberVersion.Core().LessThan(minVersionKeyring) {
+				return false
+			}
+		}
+		return true
+	}
```
schmichael, line 1998: Could this use https://github.com/hashicorp/nomad/blob/v1.4.0/nomad/util.go#L146-L162?

I like the actual version checking better here, but that helper we normally use includes some safety/liveness checks that seem important. I think this approach breaks for folks who don't upgrade in place because the old rolled-out servers will still be in serf?

nomad/leader.go
tgross: This is one of those annoying errors that bubbles up from the state store's `txn.Get` that can't actually ever fail unless we give `txn.Get` the wrong table name, which isn't ever going to happen in prod builds. We could probably eliminate a lot of spurious error checking if we panicked in the state store on those cases, but the current idiom in the state store code is to bubble it up.

nomad/leader.go
tgross: Good catch. Will fix.

tgross: Done. We could probably update that helper to use `Core()` as well, but that's future work 😀

schmichael (approved):

PR #14831, opened 2022-10-06
docs: clarify nomad vars vs vault

schmichael:
I think we should make the difference in root key management between Nomad and Vault clear in the concept docs. I didn't see anywhere else in the docs we compared it.

I also s/secrets/variables everywhere except the first sentence since the feature is intended to be more generic than secrets. Right now it's more of a compliment to Consul's kv than Vault due to root key handling and featureset.

### DIFF
--- a/website/content/docs/concepts/variables.mdx
+++ b/website/content/docs/concepts/variables.mdx
@@ -7,24 +7,24 @@ description: Learn about the Nomad Variables feature
 # Nomad Variables
 
 Most Nomad workloads need access to config values or secrets. Nomad has a
-`template` block to provide such configuration to tasks, but has left the role
-of storing that configuration to external services such as [HashiCorp Consul]
-and [HashiCorp Vault].
+`template` block to provide such configuration to tasks, but prior to Nomad 1.4
+has left the role of storing that configuration to external services such as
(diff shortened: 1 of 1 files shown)


website/content/docs/concepts/variables.mdx
```diff
@@ -7,24 +7,24 @@ description: Learn about the Nomad Variables feature
...
 
 Note that the Variables feature is intended for small pieces of configuration
 data needed by workloads. Because writing to the Nomad state store uses
 resources needed by Nomad, it is not well-suited for large or fast-changing
 data. For example, do not store batch job results as Variables - these should be
 stored in an external database. Variables are also not intended to be a full
-replacement for HashiCorp Vault. If you need powerful options like dynamic
-secrets or transit encryption, continue using Vault.
+replacement for HashiCorp Vault. Unlike Vault, Nomad stores the root encryption
+key on the servers. See [Key Management] for details.
```
tgross, line 27: The links work fine without the extra `[]` but it syntax highlights in most markdown modes in editors if you have it in there, which is handy:

```suggestion
key on the servers. See [Key Management][] for details.
```

tgross (approved):
LGTM!

PR #14833, opened 2022-10-06
[ui] Adds meta to job list stub and displays a pack logo on the jobs index

### LINKED ISSUE #11886: Show pack label on the main jobs view
### Proposal
It would be great to differentiate between "normal" jobs and jobs created by nomad-pack. Jobs created by nomad-pack already display a little logo when drilling down into the specific jobs, it would be great if this information is available on the main jobs view already.

### Use-cases
Distinguish between "normal" jobs and jobs created by nomad-pack from the main jobs view instead of drilling down into specific jobs

### Attempted Solutions
N/A

philrenaud:
Resolves #11886 

Adds a Pack logo to the jobs list index's job row:
image

### DIFF
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -1063,6 +1063,7 @@ type JobListStub struct {
 	ModifyIndex       uint64
 	JobModifyIndex    uint64
 	SubmitTime        int64
+	Meta              map[string]string `json:",omitempty"`
 }
 
 // JobIDSort is used to sort jobs by their job ID's.
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -175,7 +175,7 @@ func (j *Jobs) ListOptions(opts *JobListOptions, q *QueryOptions) ([]*JobListStu
 
 	if opts != nil && opts.Fields != nil {
 		qp := url.Values{}
-		qp.Add("Meta", fmt.Sprint(opts.Fields.Meta))
+		qp.Add("meta", fmt.Sprint(opts.Fields.Meta))
 		destinationURL = destinationURL + "?" + qp.Encode()
 	}
 
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -155,10 +155,31 @@ func (j *Jobs) RegisterOpts(job *Job, opts *RegisterOptions, q *WriteOptions) (*
 	return &resp, wm, nil
 }
 
+type JobListFields struct {
+	Meta bool
+}
+type JobListOptions struct {
+	Fields *JobListFields
+}
+
 // List is used to list all of the existing jobs.
 func (j *Jobs) List(q *QueryOptions) ([]*JobListStub, *QueryMeta, error) {
+	return j.ListOptions(nil, q)
+}
+
+// List is used to list all of the existing jobs.
+func (j *Jobs) ListOptions(opts *JobListOptions, q *QueryOptions) ([]*JobListStub, *QueryMeta, error) {
 	var resp []*JobListStub
-	qm, err := j.client.query("/v1/jobs", &resp, q)
+
+	destinationURL := "/v1/jobs"
+
+	if opts != nil && opts.Fields != nil {
+		qp := url.Values{}
+		qp.Add("Meta", fmt.Sprint(opts.Fields.Meta))
+		destinationURL = destinationURL + "?" + qp.Encode()
+	}
+
+	qm, err := j.client.query(destinationURL, &resp, q)
 	if err != nil {
 		return nil, qm, err
 	}
--- a/command/agent/job_endpoint.go
+++ b/command/agent/job_endpoint.go
@@ -37,6 +37,16 @@ func (s *HTTPServer) jobListRequest(resp http.ResponseWriter, req *http.Request)
 		return nil, nil
 	}
 
+	args.Fields = &structs.JobStubFields{}
+	// Parse meta query param
+	jobMeta, err := parseBool(req, "meta")
+	if err != nil {
+		return nil, err
+	}
+	if jobMeta != nil {
+		args.Fields.Meta = *jobMeta
+	}
+
 	var out structs.JobListResponse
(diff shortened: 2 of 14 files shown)


ChaiWithJai (approved):

command/agent/job_endpoint.go
```diff
@@ -37,6 +37,16 @@ func (s *HTTPServer) jobListRequest(resp http.ResponseWriter, req *http.Request)
...
 
+	args.Fields = &structs.JobStubFields{}
+	// Parse meta query param
+	jobMeta, err := parseBool(req, "meta")
+	if err != nil {
+		return nil, err
+	}
+	if jobMeta != nil {
+		args.Fields.Meta = *jobMeta
+	}
```
ChaiWithJai, line 48: `praise`: Good job writing some `Go` code!

ui/app/routes/jobs/index.js
```diff
@@ -32,7 +32,7 @@ export default class IndexRoute extends Route.extend(
     controller.set('namespacesWatch', this.watchNamespaces.perform());
     controller.set(
       'modelWatch',
-      this.watchJobs.perform({ namespace: controller.qpNamesapce })
+      this.watchJobs.perform({ namespace: controller.qpNamesapce, meta: true })
```
ChaiWithJai, line 35: `unrelated` `nit`: Do you see that typo to the left of your change?

ChaiWithJai: `unrelated` `question`: We don't poll `/jobs` though? I was thinking about removing this. Any concerns you can think of while we're here?

ui/app/routes/jobs/job.js
```diff
@@ -35,7 +35,7 @@ export default class JobRoute extends Route {
         const relatedModelsQueries = [
           job.get('allocations'),
           job.get('evaluations'),
-          this.store.query('job', { namespace }),
+          this.store.query('job', { namespace, meta: true }),
```
ChaiWithJai, line 38: `question` `non-blocking`: We're making a second request to `/jobs/:jobId` here. Maybe we should refactor line 33 `.findRecord` to `.query` instead?

ChaiWithJai (approved):
Good job buddy.

ui/app/routes/jobs/index.js
philrenaud: We do poll existing jobs for changes; if you are on the index page and a `pending` job changes to `ready`, for example, that update will propagate to the list.

nomad/structs/structs.go
```diff
@@ -4720,6 +4734,7 @@ type JobListStub struct {
 	ModifyIndex       uint64
 	JobModifyIndex    uint64
 	SubmitTime        int64
+	Meta              map[string]string
```
tgross, line 4737: We should make sure this field is included on `api.JobListStub` as well. 

We should probably add a `api.JobListWithFields` method to the `api` package as well (under api/jobs.go), but I'm fine with shipping this PR without and we can come back to that later: right now the UI will be the only consumer of this field.

tgross (approved):
LGTM! Let's get that field added to `api.JobListStub` and ship!

nomad/structs/structs.go
lgfa29, line 4737:
```suggestion
	Meta              map[string]string `json:",omitempty"`
```

Welcome the API side of things, this is where we make the life of UI engineers harder 😅 

I think omitting the `Meta` field if not requested makes sense? Otherwise the job list will always return `null`:

```console
$ nomad operator api '/v1/jobs?pretty'
[
  {
    "CreateIndex": 11,
    ... 
    "Meta": null,
    "ModifyIndex": 22,
    "Multiregion": null,
    ...
  }
]
``` 

But I don't know if this makes things harder for the UI since the field may not be present if `Meta` is empty.

ui/app/templates/components/topo-viz/datacenter.hbs
```diff
@@ -19,3 +19,4 @@
     </FlexMasonry>
   </div>
 </div>
+
```
lgfa29, line 22: Seems like some leftover

lgfa29 (approved):
Nice one!

In addition to the update to the `api` package we need to update the docs as well:
https://github.com/hashicorp/nomad/blob/main/website/content/api-docs/jobs.mdx?plain=1#L27-L47

My comment about the `omitempty` can be ignored, I'm not sure if it's a good idea or not give how it behaves kind of weird.

nomad/structs/structs.go
tgross: Good catch, @lgfa29

philrenaud: Ahhhh this is awesome and will definitely simplify the work we have to do in JS, as "property exists but is null" is treated differently than "property doesn't exist" (Ember will overwrite existing meta details in the former case).

TIL omitempty!

api/jobs.go
```diff
@@ -155,10 +155,31 @@ func (j *Jobs) RegisterOpts(job *Job, opts *RegisterOptions, q *WriteOptions) (*
...
+// List is used to list all of the existing jobs.
+func (j *Jobs) ListOptions(opts *JobListOptions, q *QueryOptions) ([]*JobListStub, *QueryMeta, error) {
 	var resp []*JobListStub
-	qm, err := j.client.query("/v1/jobs", &resp, q)
+
+	destinationURL := "/v1/jobs"
+
+	if opts != nil && opts.Fields != nil {
+		qp := url.Values{}
+		qp.Add("Meta", fmt.Sprint(opts.Fields.Meta))
```
tgross, line 178: I checked this out locally to give it a try, and the server is sensitive to capitalization:

```suggestion
		qp.Add("meta", fmt.Sprint(opts.Fields.Meta))
```

tgross (approved):
LGTM @philrenaud! I've left the one fix we'll need for capitalization of the field in the `api` package, but other than that it's ready to ship!

Issue #14822, opened 2022-10-06
Services with a task defined fails if there are tcp/http checks


dani:
For one of my job (to run a traefik instance), I need both native connect support (for traefik to reach the proxied services) and an Envoy sidecar (to expose two external services, LDAP and outbound web proxy which are required for traefik, and are reachable through a mesh terminating gateway). I used to do this by defining 2 services, one with connect = "native" and the other with a sidecar.

For this to work in Nomad 1.3, I had to define task = "traefik" in the connect native service, so Nomad could identify which task is using the native connect.

Attached is a sample / simplified job which used to run in Nomad 1.3, but now fails in Nomad 1.4.
[traefik-test.nomad.txt](https://github.com/hashicorp/nomad/files/9725549/traefik-test.nomad.txt)

The error with Nomad 1.4 is
```
Error submitting job: Unexpected response code: 500 (rpc error: 1 error occurred:
	* Task group traefik-test validation failed: 1 error occurred:
	* Task group service validation failed: 1 error occurred:
	* Check service: "traefik" check invalid: only script and gRPC checks should have tasks)
```

If I remove the task = "traefik" stanza from the traefik service, the job fails with
```
Error submitting job: Unexpected response code: 500 (rpc error: error in job mutator connect: task for Consul Connect Native service traefik-test->traefik is ambiguous and must be set)
```
This was already the case in Nomad 1.3, and was the reason I've added task = "traefik".

Now in Nomad 1.4 I have to either

- Run without any test
- Run with script tests, which would be really suboptimal

gmichalec-pandora:
I'll also add that the documentation for task does not indicate it cannot be used for non-grpc/script checks: 
>[task](https://developer.hashicorp.com/nomad/docs/job-specification/service#task-1) (string: "") - Specifies the name of the Nomad task associated with this service definition. Only available on group services. Must be set if this service definition represents a Consul Connect-native service and there is more than one task in the task group.

This also is not mentioned in the upgrade guide. We have lots of service definitions that specify a task. Is there a reason why having a task specified in a regular health check is an error vs a warning?

maxramqvist:
I see the same issue with checks for "http"-services on the task-level on Nomad 1.4.1. Worked fine on 1.3.5.
```
Task group cats validation failed: 1 error occurred:
	* Task group service validation failed: 1 error occurred:
	* Check service: "cats" check invalid: only script and gRPC checks should have tasks
``` 
The check is under the Services level and im running Connect, not native.

[...]

PR #14849, opened 2022-10-07
checks: don't validate task/check compatibility for Connect services

tgross:
Fixes https://github.com/hashicorp/nomad/issues/14822

In Nomad 1.4.0 we plumbed-through a task name to all checks in order to support native service discovery. For Connect tasks with a non-script/gRPC check, this triggers a validation error that setting the task name on a check isn't compatible with http/tpc checks. Normally this is correct behavior but for Connect tasks we need to include the `task` field. Bypass this validation for Connect services.

Note to reviewers: see my comment in https://github.com/hashicorp/nomad/issues/14822#issuecomment-1271714556. I've tested that this fixes the immediate problem but I'm not familiar enough with the original changes for service disco checks yet to feel super confident that this is the right approach.

### DIFF
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -6681,7 +6681,7 @@ func (tg *TaskGroup) validateServices() error {
 
 		for _, check := range service.Checks {
 			if check.TaskName != "" {
-				if check.Type != ServiceCheckScript && check.Type != ServiceCheckGRPC {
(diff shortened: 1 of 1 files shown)


jrasell (commented):
This LGTM, but I'd like to wait on merging until Seth has also taken a look. I think we'd also want a changelog entry before this is merged.

tgross:
Closed in lieu of https://github.com/hashicorp/nomad/pull/14864

Issue #11781, opened 2022-01-05
Ability to run the command directly from a mounted hosted volume


wojciak:
### Proposal
I'd like to use an fsx mount as a hosted volume for my nomad cluster, the issue is that I cannot run any apps from it, because the volume is not mounted by the time of the app execution.

The proposal is to make sure (or add a lifecycle toggle) that when the config.command stanza is run, all the hosted volume mounts are resolved.

### Use-cases
This is very useful for hotfix deployment and any ci/cd pipeline

### Attempted Solutions
Prestart lifecycle hook with a different task, poststart lifecycle hook - as all tasks have their own allocations this still won't make the volume available when the config.command stanza is run

wojciak:
Ok I did some digging and it is possible with some workarounds see:
- https://discuss.hashicorp.com/t/running-a-command-from-a-mounted-hosted-volume/33927/2
- https://github.com/hashicorp/nomad/issues/8410

Would be cool to have this behavior build-in somehow (like disable_checker = true on the exec driver)

DerekStrickland:
Hi @wojciak 

Thanks for all your work and thoughts on this. We'll review this suggestion, and think about feasibility, etc. From there we'll update this issue with next steps.  

Derek

tgross:
Hi @wojciak effectively you're running into the exact same issue I commented on here https://github.com/hashicorp/nomad/issues/11690#issuecomment-996050659:

> you're running into a consequence of the order in which the filesystem for a task is built. If you take a look at the [filesystem docs](https://www.nomadproject.io/docs/internals/filesystem#templates-artifacts-and-dispatch-payloads) you'll see that the task gets started with the volume mount in a single step at the end. That happens in the task driver by calling out to [`libcontainer`](https://github.com/hashicorp/nomad/blob/v1.2.3/drivers/shared/executor/executor_linux.go#L100-L119). So you can't use binaries from a volume mount for the `exec` task driver.
>
> As a workaround, you _might_ be able to get away with using `/bin/sh` as the command and have it exec into your binary, because it won't be started until after the volume is mounted. Something like:
> 
```hcl
commands = "/bin/sh"
args = ["-c", "exec /foo/var"]
```

The source of this issue is arising from libcontainer and not something we have control over in Nomad. It might be possible to do in some future task drivers not based on libcontainer but right now that's virtually all of them. I'm going to keep this issue open but I want to be honest with you that I would not expect to be implemented anytime soon. 

[...]

PR #14851, opened 2022-10-07
exec: allow running commands from host volume

tgross:
Fixes https://github.com/hashicorp/nomad/issues/11781

The exec driver and other drivers derived from the shared executor check the path of the command before handing off to libcontainer to ensure that the command doesn't escape the sandbox. But we don't check any host volume mounts, which should be safe to use as a source for executables if we're letting the user mount them to the container in the first place.

Check the mount config to verify the executable lives in the mount's host path, but then return an absolute path within the mount's task path so that we can hand that off to libcontainer to run.

Includes a good bit of refactoring here because the anchoring of the final task path has different code paths for inside the task dir vs inside a mount. But I've fleshed out the test coverage of this a good bit to ensure we haven't created any regressions in the process.

---

Longer-term with some sort of `exec` v2 I'd like to switch all of this to an overlay FS, but we'll need a new FSIsolation mode so as not to break any 3rd-party drivers relying on the existing alloc dir behavior.

Example jobspec that this enables, assuming a binary named `hello` at the host volume named `shared_data`:

```hcl
job "example" {
  datacenters = ["dc1"]

  group "hello" {

    volume "host_data" {
      type      = "host"
      read_only = false
      source    = "shared_data"
    }

    task "hello" {

      driver = "exec"

      config {
        command = "/host_data/hello"
      }

      volume_mount {
        volume      = "host_data"
        destination = "/host_data"
        read_only   = true
      }

    }
  }
}
```

### DIFF
--- a/drivers/shared/executor/executor_linux.go
+++ b/drivers/shared/executor/executor_linux.go
@@ -120,32 +120,15 @@ func (l *LibcontainerExecutor) Launch(command *ExecCommand) (*ProcessState, erro
 	l.container = container
 
 	// Look up the binary path and make it executable
-	absPath, err := lookupTaskBin(command)
-
+	taskPath, hostPath, err := lookupTaskBin(command)
 	if err != nil {
 		return nil, err
 	}
-
-	if err := makeExecutable(absPath); err != nil {
+	if err := makeExecutable(hostPath); err != nil {
 		return nil, err
 	}
 
-	path := absPath
-
-	// Ensure that the path is contained in the chroot, and find it relative to the container
-	rel, err := filepath.Rel(command.TaskDir, path)
-	if err != nil {
-		return nil, fmt.Errorf("failed to determine relative path base=%q target=%q: %v", command.TaskDir, path, err)
-	}
-
-	// Turn relative-to-chroot path into absolute path to avoid
-	// libcontainer trying to resolve the binary using $PATH.
-	// Do *not* use filepath.Join as it will translate ".."s returned by
-	// filepath.Rel. Prepending "/" will cause the path to be rooted in the
-	// chroot which is the desired behavior.
-	path = "/" + rel
-
-	combined := append([]string{path}, command.Args...)
+	combined := append([]string{taskPath}, command.Args...)
 	stdout, err := command.Stdout()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


mikenomitch:
Would close https://github.com/hashicorp/nomad/issues/11781

schmichael:
> The exec driver and other drivers derived from the shared executor check the path of the command before handing off to libcontainer to ensure that the command doesn't escape the sandbox

Is the only reason for this to "speed up" errors due to the executable not being reachable to pre-chroot instead of post? Seems like just letting the command fail to execute in the chroot would fix this surprising behavior and simplify code at the cost of "slower" task failures?

tgross:
> Is the only reason for this to "speed up" errors due to the executable not being reachable to pre-chroot instead of post?

It turns out we also set the executable bit on the command binary for the user. For the `exec` driver (or any driver with `FSIsolationChroot` instead of `FSIsolationImage`) the task dir includes copies of the `client.chroot_env`. So it's safe to set the executable bit... so long as we make sure that we're doing so to one of those files copied into the task dir sandbox. Or now, in the mounted host volume.

drivers/shared/executor/executor_linux.go
```diff
@@ -120,32 +120,15 @@ func (l *LibcontainerExecutor) Launch(command *ExecCommand) (*ProcessState, erro
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to determine relative path base=%q target=%q: %v", command.TaskDir, path, err)
-	}
-
-	// Turn relative-to-chroot path into absolute path to avoid
-	// libcontainer trying to resolve the binary using $PATH.
-	// Do *not* use filepath.Join as it will translate ".."s returned by
-	// filepath.Rel. Prepending "/" will cause the path to be rooted in the
-	// chroot which is the desired behavior.
-	path = "/" + rel
```
tgross, line 146: Note for reviewers: this has been folded in to `lookupTaskBin`'s inner functions. The alternative was to have `lookupTaskBin` pass back flags to the caller about where it got the paths, and that was all pretty gross.

drivers/shared/executor/executor_linux.go
```diff
@@ -805,52 +788,125 @@ func cmdMounts(mounts []*drivers.MountConfig) []*lconfigs.Mount {
...
-		if dir == "" {
-			// match unix shell behavior, empty path element == .
-			dir = "."
-		}
-		path := filepath.Join(root, dir, bin)
-		f, err := os.Stat(path)
-		if err != nil {
-			continue
-		}
-		if m := f.Mode(); !m.IsDir() {
```
tgross, line 849: Note for reviewers: this was copied almost verbatim out of the stdlib but passed a mock PATH. But since we control the mock path there's no need to do any PATH parsing and we can just iterate over a known set of directories instead. That lets us use `getPathInTaskDir` for searching the task dir, the local dir, and these paths rather than having special cases.

drivers/shared/executor/executor_linux.go
```diff
@@ -805,52 +788,125 @@ func cmdMounts(mounts []*drivers.MountConfig) []*lconfigs.Mount {
 	return r
 }
 
-// lookupTaskBin finds the file `bin` in taskDir/local, taskDir in that order, then performs
-// a PATH search inside taskDir. It returns an absolute path. See also executor.lookupBin
-func lookupTaskBin(command *ExecCommand) (string, error) {
+// lookupTaskBin finds the file `bin` in taskDir/local, taskDir in that order,
+// then performs a PATH search inside taskDir. It returns an absolute path
```
shoenig, line 792: needs a mention of the mount paths in the ordering - might be nice to format this as a bullet list, including the PATH style lookup dirs

shoenig: and the search order of the mounts is the order of mounts listed on the task, correct?

shoenig: might be good to detail all this on the exec docs page

drivers/shared/executor/executor_linux.go
```diff
@@ -805,52 +788,125 @@ func cmdMounts(mounts []*drivers.MountConfig) []*lconfigs.Mount {
...
+// inside the container that will get passed as arg[0] to the launched process,
+// and the absolute path to that binary as seen by the host (these will be
+// identical for binaries that don't come from mounts).
+//
+// See also executor.lookupBin for a version used by non-isolated drivers.
+func lookupTaskBin(command *ExecCommand) (string, string, error) {
 	taskDir := command.TaskDir
 	bin := command.Cmd
```
shoenig, line 800: does it make sense to search `task/local/` and `task/` if the `bin` is given as absolute? don't think i've ever tried that, but it sounds confusing if it works

shoenig (approved):
LGTM; see thoughts

drivers/shared/executor/executor_linux.go
tgross: Agreed, will do.

drivers/shared/executor/executor_linux.go
tgross: It _does_ sound confusing, and I spent much of yesterday puzzled over that! But it's the existing behavior (ref [`executor_linux.go#L814-L819`](https://github.com/hashicorp/nomad/blob/main/drivers/shared/executor/executor_linux.go#L814-L819)) so changing it now will break backwards compatibility. Let's save changing it for `exec` v2.

drivers/shared/executor/executor_linux.go
```diff
@@ -805,52 +788,125 @@ func cmdMounts(mounts []*drivers.MountConfig) []*lconfigs.Mount {
...
+
+	hostPath := filepath.Join(mountHostPath, mountRel)
+
+	f, err := os.Stat(hostPath)
+	if err != nil {
+		return "", "", err
+	}
+	if m := f.Mode(); m.IsDir() {
+		return "", "", fmt.Errorf("path was directory, not file")
+	}
```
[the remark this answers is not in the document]
tgross: Looks like there's a `IsRegular()` method on file mode that covers us nicely here. I've pull this out to a function.

drivers/shared/executor/executor_linux.go
shoenig: Just like I imagined!

PR #14863, opened 2022-10-10
build: move imports into the transitive require block

shoenig:


tgross (approved):
LGTM! I was wondering how those ended up being stragglers and they make backports and OSS->ENT merges messy. Thanks!

shoenig:
Yeah sometimes GoLand or other tools will add imports into a new block instead of the existing one, no idea why

PR #14864, opened 2022-10-10
services: remove assertion on 'task' field being set

### LINKED ISSUE #14822: Services with a task defined fails if there are tcp/http checks

shoenig:
This PR removes the assertion around when the 'task' field of
a check may be set. Starting in Nomad 1.4 we automatically set
the task field on all checks in support of the NSD checks feature.

This is causing validation problems elsewhere, e.g. when a group
service using the Consul provider sets 'task' it will fail
validation that worked previously.

The assertion of leaving 'task' unset was only about making sure
job submitters weren't expecting some behavior, but in practice
is causing bugs now that we need the task field for more than it
was originally added for.

We can simply update the docs, noting when the task field set by
job submitters actually has value.

Closes #14822
Closes #14849

### DIFF
--- a/.changelog/14864.txt
+++ b/.changelog/14864.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+services: Fixed a regression where check task validation stopped allowing some configurations
+```
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -6681,10 +6681,6 @@ func (tg *TaskGroup) validateServices() error {
 
 		for _, check := range service.Checks {
 			if check.TaskName != "" {
-				if check.Type != ServiceCheckScript && check.Type != ServiceCheckGRPC {
-					mErr.Errors = append(mErr.Errors,
-						fmt.Errorf("Check %s invalid: only script and gRPC checks should have tasks", check.Name))
-				}
 				if check.AddressMode == AddressModeDriver {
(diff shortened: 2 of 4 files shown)


tgross (approved):
LGTM! Thanks!

PR #14866, opened 2022-10-10
helpers: lockfree lookup of nobody user on linux and darwin

shoenig:
This PR continues the nobody user lookup saga, by making the nobody
user lookup lock-free on linux and darwin.

By doing the lookup in an init block this originally broke on Windows,
where we must avoid doing the lookup at all. We can get around that
breakage by only doing the lookup on linux/darwin where the nobody
user is going to exist.

Also return the nobody user by value so that a copy is created that
cannot be modified by callers of Nobody().

`unix` build tag is so new I couldn't find official docs for it, but it exists
https://groups.google.com/g/golang-checkins/c/_kGDaBrYk6k


tgross (approved):
LGTM. This is probably what we should've done when we fixed this, but we were under some time pressure for the release.

Aside: typically if I'm using NSS and LDAP, I'm going to have a line in `/etc/nsswitch.conf` like `passwd: files ldap`. And most folks using NSS + LDAP will have the results cached in `nscd` but that won't help us on first boot. But presumably just `passwd: ldap` is possible and then your nobody user would be dependent on the LDAP client being up, which means you've got ordering requirements for Nomad.

But maybe this seems like a "please make better choices" scenario and if you've made that particular choice, you should also realize you need to configure your systemd unit file dependencies appropriately?

schmichael (approved):

shoenig:
What do ya'll think about adding a section to https://www.nomadproject.io/docs/operations/nomad-agent about user management - e.g. beyond this PR there is still the issue where you cannot use `nobody` in `exec` (or any other chroot driver) unless your `nobody` user is defined in `/etc/passwd`. Might be nice to document that along with the systemd ordering suggestion.

PR #14868, opened 2022-10-10
[no ci] servicedisco: implicit constraint for nomad v1.4 when using nsd checks

### LINKED ISSUE #14862: nsd: add implicit Nomad version constraint for tasks which include checks
In clusters running mixed versions of clients at 1.3.x and 1.4.x, allocations that include Nomad service discovery health checks could be placed on clients that do not include the functionality due to their version. It may therefore be useful to add an implicit constraint on tasks which utilise Nomad service discovery checks to ensure they get placed on clients running Nomad version 1.4.0 or greater.

shoenig:
This PR adds a jobspec mutator to constrain jobs making use of checks
in the nomad service provider to nomad clients of at least v1.4.0.

Before, in a mixed client version cluster it was possible to submit
an NSD job making use of checks and for that job to land on an older,
incompatible client node.

Closes #14862


jrasell (approved):
LGTM!

Just noting to future readers, if we need to backport a fix for the NSD mutator in 1.3.X we will likely need two PRs to account for the updated use of go-set. This is fairly unlikely though I expect.

PR #14883, opened 2022-10-12
Corrected example

fraenku:
The description of the second example was not in sync with the example provided.


jrasell (changes requested):
Hi @fraenku and thanks for raising this. I have suggested a change to the wording to clarify the documentation as I feel both version are a little ambiguous. Let me know what you think.

We'll also need you to sign the CLA before we can merge this, thanks.

Issue #14896, opened 2022-10-15
Not upgrading servers across all regions to 1.4 causes "keyring has not been initialized yet"


optiz0r:
### Nomad version

Main region: 1.4.1
Alternate region: 1.3.5

### Issue

After upgrading 3x servers in the main region (which included the leader node) from 1.3.5 to 1.4.1, it was impossible to schedule any new jobs until the a second region containing a single server was also upgraded from 1.3.3 to 1.4.1. All evals failed with error "keyring has not been initialized yet". As soon as the alternate region server was upgraded, jobs in the main region started to be allocated once more.

### Reproduction steps

* Upgrade region containing leader form 1.3.5 to 1.4.1, leaving a server in a second region at 1.3.3.
* Attempt to schedule any new jobs

#### Expected Result

Jobs to continue to be scheduled in the main cluster once all servers had been upgraded to 1.4.1

#### Actual Result

Cluster was in-operable until all servers across all regions were upgraded.

#### Possible remedies

The upgrade-specific instructions hint that there could be problems if the leader is allowed to rollback to an older cluster version, but not that servers across all regions need to be upgraded simultaneously in order to maintain cluster function.
Either the upgrade nodes should call out this requirement, or the leader should be able to handle servers in another region still running on an older version. Typically I would only upgrade one region at a time.

### Nomad Server logs

```
Oct 14 21:58:31 leader.example.com nomad[276724]:     2022-10-14T21:58:31.265Z [ERROR] worker: error invoking scheduler: worker_id=4ba20503-dafb-ee08-c327-7e0d1779e4ae error="failed to process evaluation: rpc error: keyring has not been initialized yet"
Oct 14 21:58:31 leader.example.com nomad[276724]:     2022-10-14T21:58:31.725Z [ERROR] worker: failed to submit plan for evaluation: worker_id=4ba20503-dafb-ee08-c327-7e0d1779e4ae eval_id=6c8ce736-2a39-836b-1d83-e51714187792 error="rpc error: keyring has not been initialized yet"
Oct 14 21:58:31 leader.example.com nomad[276724]:     2022-10-14T21:58:31.726Z [ERROR] worker: error invoking scheduler: worker_id=4ba20503-dafb-ee08-c327-7e0d1779e4ae error="failed to process evaluation: rpc error: keyring has not been initialized yet"
[...]

PR #14901, opened 2022-10-17
keyring: filter by region before checking version

tgross:
In #14821 we fixed a panic that can happen if a leadership election happens in the middle of an upgrade. That fix checks that all servers are at the minimum version before initializing the keyring (which blocks evaluation processing during trhe upgrade). But the check we implemented is over the serf membership, which includes servers in any federated regions, which don't necessarily have the same upgrade cycle.

Filter the version check by the leader's region.

Fixes https://github.com/hashicorp/nomad/issues/14896 but this fix makes me wonder if we have other lurking bugs in this version check.

### DIFF
--- a/.changelog/14901.txt
+++ b/.changelog/14901.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+keyring: Fixed a bug where keyring initialization is blocked by un-upgraded federated regions
+```
--- a/nomad/encrypter.go
+++ b/nomad/encrypter.go
@@ -474,7 +474,7 @@ START:
 					// new leader has not yet replicated the key from
 					// the old leader before the transition. Ask all
 					// the other servers if they have it.
-					krr.logger.Debug("failed to fetch key from current leader",
+					krr.logger.Warn("failed to fetch key from current leader, trying peers",
 						"key", keyID, "error", err)
 					getReq.AllowStale = true
 					for _, peer := range krr.getAllPeers() {
@@ -494,7 +494,7 @@ START:
 					krr.logger.Error("failed to add key", "key", keyID, "error", err)
 					goto ERR_WAIT
 				}
-				krr.logger.Trace("added key", "key", keyID)
+				krr.logger.Info("added key", "key", keyID)
 			}
 		}
 	}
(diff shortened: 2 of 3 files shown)


tgross:
(An alternate approach here could be to get the localPeers as done in the keyring replication: https://github.com/hashicorp/nomad/blob/9cdbbbb583f77c6c0bd8e799a588173babc80ba9/nomad/encrypter.go#L514-L523)

tgross:
Got this failure twice... it doesn't look like a flake and it looks like it's potentially relevant, but I can't figure out where it's coming from yet.

```
=== RUN   TestServer_ReconcileMember
    leader_test.go:1463: found leader: f00dc7a1-7097-8bbe-2659-ecdcd4ff3622 127.0.0.1:9391
    leader_test.go:1472: found leader: f00dc7a1-7097-8bbe-2659-ecdcd4ff3622 127.0.0.1:9391
Error: 3 2022-10-17T14:37:55.265Z [ERROR] raft-net: failed to flush response: error="write tcp 127.0.0.1:9387->127.0.0.1:35878: write: broken pipe"
    leader_test.go:1496: got 2 server ids want 3: []raft.Server{raft.Server{Suffrage:0, ID:"f00dc7a1-7097-8bbe-2659-ecdcd4ff3622", Address:"127.0.0.1:9391"}, raft.Server{Suffrage:0, ID:"cefda963-7d8f-9a8c-131d-bec0073fe8c2", Address:"127.0.0.1:9383"}}
--- FAIL: TestServer_ReconcileMember (0.19s)
```

schmichael (approved):
As discussed offline we need to check our other uses of ServersMeetMinimumVersion and ensure if they're local-only they do this region check, but if they're global (eg ACL related objects) they need to check *all* members.

Great find on this, I can't believe it hasn't bitten us before. Although I wonder if it has in such a subtle way that we didn't realize it. In other spots this would more or less silently prevent using a new code path.

tgross:
I'm going to merge this one, and then take a pass thru the existing uses of `ServersMeetMinimumVersion` and will push up a separate PR for that (because that'll likely need different backporting anyways). We'll also want to pick up https://github.com/hashicorp/nomad/pull/14908 for 1.4.2

PR #14902, opened 2022-10-17
cleanup: remove another string-set helper function

shoenig:
This fix for https://github.com/hashicorp/nomad/issues/14914 should make use of the changes here.


tgross (approved):
LGTM!

PR #14908, opened 2022-10-17
acl: gate ACL role write and delete RPC usage on v1.4.0 or greater.

jrasell:
This ensures all federated servers are running v1.4.0 or greater before the use of ACL roles is allowed.

### DIFF
--- a/nomad/acl_endpoint.go
+++ b/nomad/acl_endpoint.go
@@ -1097,6 +1097,13 @@ func (a *ACL) UpsertRoles(
 	}
 	defer metrics.MeasureSince([]string{"nomad", "acl", "upsert_roles"}, time.Now())
 
+	// ACL roles can only be used once all servers, in all federated regions
+	// have been upgraded to 1.4.0 or greater.
+	if !ServersMeetMinimumVersion(a.srv.Members(), AllRegions, minACLRoleVersion, false) {
+		return fmt.Errorf("all servers should be running version %v or later to use ACL roles",
+			minACLRoleVersion)
+	}
+
 	// Only tokens with management level permissions can create ACL roles.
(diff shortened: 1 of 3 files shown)


nomad/acl_endpoint.go
```diff
@@ -1233,6 +1240,13 @@ func (a *ACL) DeleteRolesByID(
 	}
 	defer metrics.MeasureSince([]string{"nomad", "acl", "delete_roles"}, time.Now())
 
+	// ACL roles can only be used once all servers, in all federated regions
```
tgross, line 1243: Some of other RPC methods in this endpoint support like `ACL.UpsertTokens` write Role information. Do we want to gate those RPCs as well if `len(token.Roles) > 0`, so that we're not writing tokens that are assuming that the roles exist. I see we have a `fixTokenRoleLinks` in the state store code, but I'm not sure I understand exactly what it's doing, so maybe that data integrity problem is already accounted for?

tgross (approved):
LGTM if my comment about writing tokens with roles is off-base, otherwise we might want to look at that too.

nomad/acl_endpoint.go
jrasell: An ACL token cannot be created with an ACL role link where the ACL role does not exist; the RPC handler [performs this check](https://github.com/hashicorp/nomad/blob/1b434184e4a1c3b5df87052d8022a36866fbbb9e/nomad/acl_endpoint.go#L559-L578). Therefore we would never be able create the token without the ACL role RPC endpoints being unblocked by the feature gate.

It would of course be possible to add the check, but it feels to me we are guarded. I'll hold on on merging so you have had a chance to read and respond.

tgross: Ah, cool. LGTM!

PR #14910, opened 2022-10-17
make version checks specific to region (1.2.x) 

tgross:
* One-time tokens are not replicated between regions, so we don't want to enforce
  that the version check across all of serf, just members in the same region.
* Cleans up a bunch of legacy checks.

This changeset is specific to 1.2.x and the changes for other versions of
Nomad will be manually backported in separate PRs.

(And yes the 1.1 in the branch name is wrong 😁 )

### DIFF
--- a/nomad/util.go
+++ b/nomad/util.go
@@ -13,6 +13,7 @@ import (
 	"github.com/hashicorp/nomad/nomad/state"
 	"github.com/hashicorp/nomad/nomad/structs"
 	"github.com/hashicorp/serf/serf"
+	"golang.org/x/exp/slices"
 )
 
 // MinVersionPlanNormalization is the minimum version to support the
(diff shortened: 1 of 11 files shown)


nomad/util.go
```diff
@@ -148,7 +148,29 @@ func isNomadServer(m serf.Member) (bool, *serverParts) {
...
+	}
+
+	return true
+}
+
+// RegionServersMeetMinimumVersion returns whether the Nomad servers in this
+// region are at least on the given Nomad version. The checkFailedServers
+// parameter specifies whether version for the failed servers should be
+// verified.
+func RegionServersMeetMinimumVersion(members []serf.Member, region string, minVersion *version.Version, checkFailedServers bool) bool {
```
shoenig, line 169: Since this is the same as `ServersMeetMinimumVersion` I'm tempted to suggest just expanding that functions parameter list to include `region string`, which matches all regions if unset.

shoenig: While we're in here we can probably kill `slicesMatch` for `helper.SliceSetEq`

shoenig (approved):
LGTM; maybe fold these two functions together?

nomad/util.go
tgross: > Since this is the same as `ServersMeetMinimumVersion` I'm tempted to suggest just expanding that functions parameter list to include `region string`, which matches all regions if unset. 

We have a whole bunch of call sites for this for legacy bits I didn't want to tear out while trying to fix a bug at the same time. If we go and touch those now, we'd want to change the behavior too, otherwise it's going to look like we intentionally set them to check all regions. I was intentionally not touching those but honestly it's probably safe to go ahead and do it at this point. Will do.

> While we're in here we can probably kill `slicesMatch` for `helper.SliceSetEq`

Isn't `SliceSetEq` specifically for a _set_ (i.e. unordered). The segments of a version have meaningful order.

tgross: Turns out I can't mess with this signature without breaking all the backports in the process. So I'm going to rework all the open PRs for this problem into one PR per backport.

shoenig: oh derp, yeah. in that case there is https://pkg.go.dev/golang.org/x/exp/slices#Equal but yeah that's just optional cleanup

shoenig: You could also just keep both public signatures and have them call into a shared private helper

tgross: I've squooshed it all into https://github.com/hashicorp/nomad/pull/14912 and I'll rework this PR to just cover 1.2.x and https://github.com/hashicorp/nomad/pull/14911 to just cover 1.3.x.

nomad/util_test.go
```diff
@@ -188,7 +188,45 @@ func TestServersMeetMinimumVersionIncludingFailed(t *testing.T) {
...
+
+func TestServersMeetMinimumVersionSuffix(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		members  []serf.Member
+		ver      *version.Version
+		expected bool
+	}{
+		// Multiple servers, meets req version
```
shoenig, line 206: it's nice to put these as a string field in the test case, then execute each case inside a 
```
t.Run(tc.name, ...) {
}
```

shoenig (approved):
LGTM! just the test case suggestion

nomad/util_test.go
tgross: Oops, I didn't mean to copy this over from the cherry-pick at all.

PR #9733, opened 2021-01-06
UI: Guard against nodes running an old version of the Nomad agent

DingoEatingFuzz:
Fixes #9702 

If a cluster has nodes running a version of Nomad <0.9.0, they won't have the requisite `NodeResources` property. Instead of blowing up, now the UI will filter those nodes out of the viz and report the situation to the user.

Screen Shot 2021-01-05 at 4 17 05 PM

PR #14913, opened 2022-10-17
Adds searching and filtering for nodes on topology view

### LINKED ISSUE #10002: filter topology view by node status
### Nomad version
1.0.3

Ubuntu 20.04.2 LTS (GNU/Linux 5.4.0-1029-aws x86_64)

### Issue
This is a feature request to add filters to the Topology page, similarly to the Clients page.

In a datacenter with lots of spot instances that get terminated suddenly, the Topology view give a false impression of unused clients hanging in the cluster, when in fact their status is "down", which can be confirmed once clicking on them.

Maybe adding the same set of filters from the Clients page will improve the benefit of that view.

philrenaud:
Adds a toolbar for searching and filtering clients within the Topology view, and really ties the room together.

image

Resolves #10002

### DIFF
--- a/ui/app/controllers/topology.js
+++ b/ui/app/controllers/topology.js
@@ -143,7 +143,9 @@ export default class TopologyControllers extends Controller.extend(Searchable) {
         (selectionClass.length
           ? selectionClass.includes(node.nodeClass)
           : true) &&
-        node.name.includes(searchTerm)
+        (node.name.includes(searchTerm) ||
+          node.datacenter.includes(searchTerm) ||
+          node.nodeClass.includes(searchTerm))
       );
     });
   }
--- a/ui/app/controllers/topology.js
+++ b/ui/app/controllers/topology.js
@@ -42,6 +42,10 @@ export default class TopologyControllers extends Controller.extend(Searchable) {
   ];
 
   @tracked searchTerm = '';
+  qpState = '';
+  qpVersion = '';
+  qpClass = '';
+  qpDatacenter = '';
 
   setFacetQueryParam(queryParam, selection) {
     this.set(queryParam, serialize(selection));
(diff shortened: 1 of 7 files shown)


ChaiWithJai (approved):

ui/app/controllers/topology.js
```diff
@@ -1,23 +1,152 @@
...
+      );
+    });
+
+    return versions.sort().map((v) => ({ key: v, label: v }));
+  }
+
   @alias('userSettings.showTopoVizPollingNotice') showPollingNotice;
 
-  @tracked filteredNodes = null;
+  @tracked pre09Nodes = null;
```
philrenaud, line 123: Because we want to be able to filter nodes now, I've renamed what was `filteredNodes` to `pre09Nodes`. This modifies the code added from https://github.com/hashicorp/nomad/pull/9733 but I think in a healthy way. cc @DingoEatingFuzz

philrenaud:
image
yay Percy!

ui/app/controllers/topology.js
```diff
@@ -1,23 +1,152 @@
...
+        (selectionVersion.length
+          ? selectionVersion.includes(node.version)
+          : true) &&
+        (selectionDatacenter.length
+          ? selectionDatacenter.includes(node.datacenter)
+          : true) &&
+        (selectionClass.length
+          ? selectionClass.includes(node.nodeClass)
+          : true) &&
+        node.name.includes(searchTerm)
```
mikenomitch, line 146: I feel like node name or datacenter might be nice here

philrenaud: Sounds good, added for class and datacenter as well.

cyrilgdn:
Many many many many thanks :tada: ! It makes topology view finally useful for our Nomad usage :tada:

PR #14921, opened 2022-10-18
website: redirects to empty array

zchsh:
🎟 [Asana task](https://app.asana.com/0/0/1203168563666035/f)

This PR replaces the website redirects with an empty array. These redirects will have been migrated to the hashicorp/dev-portal project, via https://github.com/hashicorp/dev-portal/pull/1243.


zchsh:
A note for posterity: As noted in [the Asana task](https://app.asana.com/0/0/1203168563666035/f):

> If there are redirects present on main, but not on stable-website, keep those

As of right now, `git diff main:website/redirects.js stable-website:website/redirects.js` shows `website/redirects.js` as being identical between `main` and `stable-website`, so I think there's nothing extra we need to keep.

BRKalow (approved):

PR #14923, opened 2022-10-18
acl: add ACL roles to event stream topic and resolve policies.

### LINKED ISSUE #14777: events: add ACL roles to event stream as topic subscription
The events stream should support ACL role create and delete subscriptions. This new stream addition can follow the ACL policies stream work as a base. The state store implementation already supports this work and so should not need any updates.

jrasell:
This changes adds ACL role creation and deletion to the event stream. It is exposed as a single topic with two types; the filter is primarily the role ID but also includes the role name.

While conducting this work it was also discovered that the events stream has its own ACL resolution logic. This did not account for ACL tokens which included role links, or tokens with expiry times. ACL role links are now resolved to their policies and tokens are checked for expiry correctly.

Related to #14922 and includes a modification in the RPC handler to accommodate the linked PR.
Closes #14777

### DIFF
--- a/nomad/structs/acl.go
+++ b/nomad/structs/acl.go
@@ -233,13 +233,10 @@ func (a *ACLToken) IsExpired(t time.Time) bool {
 	return a.ExpirationTime.Before(t) || t.IsZero()
 }
 
-// RoleSubset checks if a given set of role IDs are assigned to the ACL token.
-func (a *ACLToken) RoleSubset(roleIDs []string) bool {
-
-	// Hot-path: management tokens allows access to all roles.
-	if a.Type == ACLManagementToken {
-		return true
-	}
+// HasRoles checks if a given set of role IDs are assigned to the ACL token. It
+// does not account for management tokens, therefore it is the responsibility
+// of the caller to perform this check, if required.
+func (a *ACLToken) HasRoles(roleIDs []string) bool {
 
 	// Generate a set of role IDs that the token is assigned.
 	roleSet := set.FromFunc(a.Roles, func(roleLink *ACLTokenRoleLink) string { return roleLink.ID })
--- a/nomad/structs/acl.go
+++ b/nomad/structs/acl.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-multierror"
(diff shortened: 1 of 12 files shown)


nomad/structs/acl.go
```diff
@@ -232,6 +233,27 @@ func (a *ACLToken) IsExpired(t time.Time) bool {
 	return a.ExpirationTime.Before(t) || t.IsZero()
 }
 
+// RoleSubset checks if a given set of role IDs are assigned to the ACL token.
+func (a *ACLToken) RoleSubset(roleIDs []string) bool {
```
tgross, line 237: This might make the signature more obvious at a glance:

```suggestion
func (a *ACLToken) HasRoles(roleIDs []string) bool {
```

nomad/structs/acl.go
```diff
@@ -232,6 +233,27 @@ func (a *ACLToken) IsExpired(t time.Time) bool {
...
+
+	// Hot-path: management tokens allows access to all roles.
+	if a.Type == ACLManagementToken {
+		return true
+	}
```
tgross, line 242: It looks like this is safe for the consumers we have today. But a role can have a policy with `deny` as well. Maybe we shouldn't have this bit of convenience and _require_ that any consumers of this are explicitly checking the management role instead, so that future consumers can't accidentally lock out the management token from something?

tgross (approved):
LGTM! I've left a couple suggestions but consider them non-blocking.

nomad/structs/acl.go
jrasell: Yeah this makes perfect sense and is a better approach. I have made the change and updated the function comment to call this out within commit: https://github.com/hashicorp/nomad/pull/14923/commits/3481a620406ba5b1f2c4d1c29047e0a9d82cef67

nomad/structs/acl.go
jrasell: Agreed and has been updated within commit: https://github.com/hashicorp/nomad/pull/14923/commits/3481a620406ba5b1f2c4d1c29047e0a9d82cef67

Thanks.

lgfa29 (approved):

Issue #10482, opened 2021-04-30
mishandling of service deregistration with script healthcheck


fredwangwang:
### Nomad version
v1.0.4

### Operating system and Environment details
linux

### Issue
when stop a job having a service with script check, the following warn msg gets shown in nomad:

```
[WARN]  client.alloc_runner.task_runner.task_hook.script_checks: updating check failed: alloc_id=4f0f5dac-3497-4bcc-e248-90bb8364ac73 task=huan-fake error="Unexpected response code: 500 (Unknown check "_nomad-check-5c9d58746b0a2a10bfaf3fc6ae306eb22a61e4a1")"
```

and in consul:

```
[ERROR] agent.http: Request error: method=PUT url=/v1/agent/check/update/_nomad-check-5c9d58746b0a2a10bfaf3fc6ae306eb22a61e4a1 from=127.0.0.1:36812 error="Unknown check "_nomad-check-5c9d58746b0a2a10bfaf3fc6ae306eb22a61e4a1""
[ERROR] agent.http: Request error: method=PUT url=/v1/agent/check/update/_nomad-check-5c9d58746b0a2a10bfaf3fc6ae306eb22a61e4a1 from=127.0.0.1:36812 error="Unknown check "_nomad-check-5c9d58746b0a2a10bfaf3fc6ae306eb22a61e4a1""
```

### Reproduction steps

Deploy a job having a service with script check, then stop the job.

#### Expected Result

No warn/error msg for a normal job shutdown

#### Actual Result

Warn/error msg as shown above

### Job file (if appropriate)

<details>
  <summary>Job file</summary>
  
```
job "huan" {
  region      = "global"
  datacenters = ["dc1"]
  type        = "service"

  group "huan-group" {
    count = 1

    network {
      mode = "bridge"
      port "api" {}
    }
    service {
      name = "huan-api"
      port = "api"

      check {
        type     = "script"
        name     = "check"
        command  = "echo"
        args     = ["1"]
        task     = "huan-fake"
        interval = "60s"
        timeout  = "5s"
      }
    }

    task "huan-fake" {
      driver = "docker"

      config {
        image = "nicholasjackson/fake-service:v0.12.0"
      }

      env {
        LISTEN_ADDR             = "0.0.0.0:${NOMAD_PORT_api}"
        NAME                    = "${NOMAD_TASK_NAME}"
[...]

PR #14944, opened 2022-10-18
consul: register checks along with service on initial registration

### LINKED ISSUE #3935: Register Consul Service and Checks atomically
### Context
We use Nomad with Consul and Docker to deploy Grpc services to a cluster. We use [Linkerd](https://linkerd.io/docs/) as the loadbalancer, it uses Consul to get the availability information of the services.

### Scenario
We observe transient errors at the consumer when a new instance of the service is getting started during the service update with Nomad.
(Tcp health check is used to avoid routing request to it when the Docker container has started and the service inside is not ready yet to receive requests. Linkerd is configured to only consider services with passing health check status)

### Expected behavior
Being able to update the services seamlessly with the above setup.

### Versions
Nomad v0.6.3
Consul v1.0.3
Linkerd v1.3.5

### Analysis
Linkerd uses blocking queries to get availability info from Consul. 
Example: `/v1/health/service/testservice?index=1366&dc=dc1&passing=true`
I captured the network traffic and this type of query returns the new service instance without the Tcp health check being defined first, then shortly afterwards it returns it the with Tcp health check.
So it seems Consul considers the service “passing” when it does not have health checks defined, 
and Nomad first registers the new service instance and then it registers its health checks in a separate call.
The captured network packets (Nomad -> Consul) confirms that it happens separately:
```
PUT /v1/agent/service/register HTTP/1.1
Host: 127.0.0.1:8500
User-Agent: Go-http-client/1.1
Content-Length: 165
Accept-Encoding: gzip
{"ID":"_nomad-executor-47313988-2a65-66e0-46af-491023330cca-session-testservice",
"Name":"testservice","Port":27123,"Address":"10.0.75.1","Check":null,"Checks":null}

HTTP/1.1 200 OK
Date: Fri, 02 Mar 2018 14:33:31 GMT
Content-Length: 0
Content-Type: text/plain; charset=utf-8
```
```
PUT /v1/agent/check/register HTTP/1.1
Host: 127.0.0.1:8500
User-Agent: Go-http-client/1.1
Content-Length: 233
Accept-Encoding: gzip
{"ID":"942d6c5b20bea9520f70ced61336a2987bf9c530","Name":"service: \"testservice\" check",
"ServiceID":"_nomad-executor-47313988-2a65-66e0-46af-491023330cca-session-testservice",
"Interval":"10s","Timeout":"2s","TCP":"10.0.75.1:27123"}

HTTP/1.1 200 OK
Date: Fri, 02 Mar 2018 14:33:31 GMT
Content-Length: 0
Content-Type: text/plain; charset=utf-8
```
Nomad should register the service and its health checks in one call in Consul,
otherwise the new service instance is considered healthy even before Nomad registers its health check, I believe.

shoenig:
This PR updates Nomad's Consul service client to include checks in
an initial service registration, so that the checks associated with
the service are registered "atomically" with the service. Before, we
would only register the checks after the service registration, which
causes problems where the service is deemed healthy, even if one or
more checks are unhealthy - especially problematic in the case where
SuccessBeforePassing is configured.

Fixes #3935

Note this doesn't address service _de_-registrations; which we'll get to in https://github.com/hashicorp/nomad/issues/10482

### DIFF
--- a/.changelog/14944.txt
+++ b/.changelog/14944.txt
@@ -1,3 +1,3 @@
-```release-note:improvement
+```release-note:bug
 consul: atomically register checks on initial service registration
 ```
--- a/.changelog/14944.txt
+++ b/.changelog/14944.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+consul: atomically register checks on initial service registration
+```
--- a/client/serviceregistration/service_registration.go
+++ b/client/serviceregistration/service_registration.go
@@ -131,7 +131,7 @@ type ServiceRegistration struct {
 	// services/checks registered in Consul. It is used to materialize the other
 	// fields when queried.
 	ServiceID string
-	CheckIDs  map[string]struct{}
+	CheckIDs  map[string]struct{} // todo: use a Set?
 
 	// CheckOnUpdate is a map of checkIDs and the associated OnUpdate value
 	// from the ServiceCheck It is used to determine how a reported checks
--- a/command/agent/consul/catalog_testing.go
+++ b/command/agent/consul/catalog_testing.go
@@ -275,14 +275,29 @@ func (c *MockAgent) CheckRegister(check *api.AgentCheckRegistration) error {
 	if c.checks[check.Namespace] == nil {
 		c.checks[check.Namespace] = make(map[string]*api.AgentCheckRegistration)
 	}
+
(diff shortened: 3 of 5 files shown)


tgross (approved):
LGTM, but should we also be removing the bit of code that registers the checks non-atomically? Or will that be a separate PR too?

PR #14948, opened 2022-10-18
Add ReconnectModifyIndex to handle reconnect lifecycle

### LINKED ISSUE #14925: Job updates not applied to reconnected allocations
### Nomad version
Confirmed on 1.3.x, 1.4.x

Ubuntu 20.04
3 servers
3 clients

### Issue

nomad job with `max_client_disconnect` becomes unmanageable  (no updates can be occur after successful updating job definition ) unless restarted, after job was in state "unknown" (nomad client was disconnected) 
 
### Reproduction steps

- Run the jobspec below
- Disconnect the nomad client where job is running
- After Job becomes in status `unknown` reconnect the nomad client
- After job s reconnected change the `image` to `httpd:latest` and run job again
- It will be successful but if you make curl request it still return nginx default response (allocation will be still using nginx container)
- If you look in the UI you can see that the Alloc that was placed on the disconnected/reconnected client does not get replaced and is running with the previous job version

#### Expected Result

Reconnected alloc will be replaced with newer jobspec config

#### Actual Result

Allocs that never transitioned to`unknown` get replaced, but the disconnected/reconnected alloc may not get replaced. In at least one manual bench test, I have seen it work as desired, but not deterministically.

### Job file (if appropriate)

```hcl
job "nginx-job" {
  datacenters = ["dc1"]

  group "nginx-group" {
    max_client_disconnect = "24h"

    count = 3

    spread {
      attribute = "${attr.unique.hostname}"
    }

    network {
      port "http" {
        to = 80
      }
    }

    task "nginx-task" {
      driver = "docker"

      config {
        // toggle these lines to create new job version
        image = "nginx:latest"
        // image = "httpd:latest"
        ports = ["http"]
      }

    }
  }
}
```

Should have place 3 and stop 3

```shell
Oct 18 10:52:20 node-server-1 nomad[1819]:   results=
Oct 18 10:52:20 node-server-1 nomad[1819]:   | Total changes: (place 2) (destructive 0) (inplace 0) (stop 2) (disconnect 0) (reconnect 0)
```

DerekStrickland:
Closes #14925 

This PR fixes a bug where if an allocation with `max_client_disconnect` configured is on a node that disconnects, and then the node reconnects, future jobspec changes for that job get ignored until the `max_client_disconnect` interval expires. Previous to this change, `Allocation.Reconnected` naively just checked the last reconnect event time and the expiry.

This PR:

- Adds a `ReconnectModifyIndex` field to the `Allocation` struct.
- Updates the alloc runner to update the alloc `ReconnectModifyIndex` when a reconnect is processed by the client
- Modifies `Client.allocSync` to send the `ReconnectModifyIndex` when syncing client managed attributes
- Modifies `Node.UpdateAlloc` to persist the incoming `ReconnectModifyIndex` when generating reconnect evals
- Renames `Allocation.Reconnected` to `Allocation.IsReconnecting`
- Refactors `Allocation.IsReconnecting` to compare the `ReconnectModifyIndex` to the `AllocModifyIndex` to determine if an allocation is reconnecting
- Updates all related code to match the new name and test the new logic
- Updates `GenericScheduler.computeJobAllocs` to reset the `ReconnectModifyIndex` to `0` when processing `reconnectUpdates` and appends them to `Plan.NodeAllocation` so that the updates get persisted


client/allocrunner/alloc_runner.go
```diff
@@ -1348,6 +1348,9 @@ func (ar *allocRunner) Reconnect(update *structs.Allocation) (err error) {
 	// Build the client allocation
 	alloc := ar.clientAlloc(states)
 
+	// Set the reconnect modify index so that the scheduler can track that the reconnect has not been processed.
+	alloc.ReconnectModifyIndex = ar.Alloc().ReconnectModifyIndex
```
tgross, line 1352: I don't think I understand what ever sets `ReconnectModifyIndex` to a non-zero value outside of tests. The approach seems right here but I can't figure out the source of the index.

DerekStrickland: I'm not entirely sure what happened with that refactoring. It was getting set in `setIndexex` and then I moved it over here.  I just updated to `alloc.ReconnectModifyIndex = ar.Alloc().AllocModifyIndex`.

nomad/node_endpoint.go
```diff
@@ -1220,6 +1220,7 @@ func (n *Node) UpdateAlloc(args *structs.AllocUpdateRequest, reply *structs.Gene
 		if evalTriggerBy != structs.EvalTriggerJobDeregister &&
 			alloc.ClientStatus == structs.AllocClientStatusUnknown {
 			evalTriggerBy = structs.EvalTriggerReconnect
+			alloc.ReconnectModifyIndex = allocToUpdate.ReconnectModifyIndex
```
tgross, line 1223: This assignment corrupts the state store because `alloc` hasn't been copied after being queried from the state store. I'm fairly certain this line isn't needed at all, as the `allocToUpdate` is what's getting added to the batch of updates and not `alloc`.

tgross (changes requested):
@lgfa29 it looks like there are some correctness issues here around the state store; let's chat internally about carrying this PR.

tgross:
Per our discussion, moving this out of 1.4.2 so that we don't risk rushing it out.

client/allocrunner/alloc_runner.go
```diff
@@ -1348,6 +1348,9 @@ func (ar *allocRunner) Reconnect(update *structs.Allocation) (err error) {
 	// Build the client allocation
 	alloc := ar.clientAlloc(states)
 
+	// Set the reconnect modify index so that the scheduler can track that the reconnect has not been processed.
+	alloc.ReconnectModifyIndex = ar.Alloc().AllocModifyIndex
```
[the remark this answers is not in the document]
DerekStrickland: So in client.go line 2033 it get's added to the stripped alloc during allocSync, that then gets sent to Node Update, which then updates state, and triggers an eval, When the  eval fires, the index is set. We then have to unset it when applying the plan.  Have you tried it out?  I had logging in here during development showing it all flowed through.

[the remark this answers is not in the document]
tgross: > which then updates state

That's the bit where I don't see how it's happening. Any update of an existing object takes a copy first (ref [`state_store.go#L3474`](https://github.com/hashicorp/nomad/blob/main/nomad/state/state_store.go#L3474)) and then modifies the copy before inserting it. So if we haven't pulled in information that the client is authoritative on, the state isn't getting updated for the transaction. 

I haven't had a chance to test it out thoroughly (still trying to get 1.4.2 out! 😁 ) but I suspect the reason it's "working" right now is because of the state store corruption on line 1223. That'll appear correct under some circumstances but won't have gone thru raft correctly.

mikenomitch:
Passing in some feedback from a customer. I think it is related to this PR, but I wanted to make sure

> I found scenario, where I see duplicated ALLOC_INDEXes in one JOB.
> Below are required steps (Nomad 1.3.5):
>  
> We have job with count=2 and set option max_client_disconnect running as below:
> NOMAD_ALLOC_INDEX=0 - NodeA
> NOMAD_ALLOC_INDEX=1 - NodeB
>  
> We stop Nomad Agent on NodeA, after that we have temporarily 3 allocations:
> NOMAD_ALLOC_INDEX=0 - NodeA (Unknown state)
> NOMAD_ALLOC_INDEX=1 - NodeB
> NOMAD_ALLOC_INDEX=0 - NodeC
>  
> When we start agent on NodeA, then again we have 2 allocations - first was recovered:
> NOMAD_ALLOC_INDEX=0 - NodeA
> NOMAD_ALLOC_INDEX=1 - NodeB
> NodeC - allocation terminated here - as expected
>  
> I'm changing count from 2 to 3 and 3rd allocation appears with the same ALLOC_INDEX as the first one!!!
> NOMAD_ALLOC_INDEX=0 - NodeA
> NOMAD_ALLOC_INDEX=1 - NodeB
> NOMAD_ALLOC_INDEX=0 - NodeD 
>  
> Apart from that, in the state with duplicated ALLOC_INDEXes, EXEC feature in Nomad UI stopped working properly (for affected JOB only)

Does this seem related or should I make a new issue?

lgfa29:
I think I understand the problem now 😅 

I have an alternative approach in https://github.com/hashicorp/nomad/pull/15068 that I think makes the disconnect/reconnect flows more similar and so easier to understand, but it's still an early work. I will keep investigating the problem to see which solution would be better.

lgfa29:
@mikenomitch I think this may be related to this problem. From our [docs on `NOMAD_ALLOC_INDEX`](https://developer.hashicorp.com/nomad/docs/runtime/environment#summary): 

> The index is unique within a given version of a job

I think #14925 may prevent the job version from changing, which means you could end up with reused indexes.

But it may be better to open a separate issue just in case. If it's the same problem we can close both issues.

lgfa29:
Closing this in favour of #15068. 

Thanks for the all the work and guidance on this issue @DerekStrickland!

DerekStrickland:
@lgfa29  I'm glad you found a good solution!

PR #14949, opened 2022-10-18
e2e: convert flaky exec download in chroot unit test into e2e test

shoenig:
Similar to https://github.com/hashicorp/nomad/pull/14710, convert flaky `TestTaskRunner_Download_ChrootExec` into e2e test.

This job
1. Creates an executable script and creates a tar (prestart task)
2. Starts an http server to serve the tar (prestart sidecar task)
3. Runs the script as the command of an exec task 

The e2e test then checks the output from the script is present in the task logs.


tgross (approved):
LGTM

PR #14955, opened 2022-10-19
deps: update consul-template to `61e288a`

jrasell:
accounts temporarily for a force push removing the previous reference.


tgross (approved):
Oops! We're going to get that branch merged today-ish, but let's get this fixed in the meantime in case it takes a little longer.

PR #14957, opened 2022-10-19
build: add ability to specify release targets

schmichael:
My make knowledge is very very limited, so if there's a better way to do this please let me know! This seems to work and lets me cut one off builds easily.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -75,6 +75,11 @@ ifeq (FreeBSD,$(THIS_OS))
 ALL_TARGETS = freebsd_amd64
 endif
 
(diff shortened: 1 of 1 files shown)


tgross (approved):
It looks like you're wanting to be able to run something like `make release TARGETS=linux_amd64` so that you're getting the codegen bits too? (Otherwise you can do `make pkg/linux_amd64/nomad`).

If so, LGTM

schmichael:
> It looks like you're wanting to be able to run something like make release TARGETS=linux_amd64 so that you're getting the codegen bits too? (Otherwise you can do make pkg/linux_amd64/nomad).

Yeah, sorry for not calling that out.

PR #14982, opened 2022-10-20
acl: allow tokens to read policies linked via roles to the token.

jrasell:
ACL tokens are granted permissions eithe rby direct policy links or via ACL role links. Callers should therefore be able to read policies directly assigned to the caller token or indirectly by ACL role links.

### DIFF
--- a/nomad/acl_endpoint.go
+++ b/nomad/acl_endpoint.go
@@ -139,7 +139,7 @@ func (a *ACL) ListPolicies(args *structs.ACLPolicyListRequest, reply *structs.AC
 
 	// If it is not a management token determine the policies that may be listed
 	mgt := acl.IsManagement()
-	var policies map[string]struct{}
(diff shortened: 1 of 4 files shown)


nomad/acl_endpoint.go
```diff
@@ -1593,3 +1611,59 @@ func (a *ACL) GetRoleByName(
 		},
 	})
 }
+
+// policyNamesFromRoleLinks resolves the policy names which are linked via the
+// passed role links. This is useful when you need to understand what polices
+// an ACL token has access to and need to include role links. The function will
+// not return a nil set object, so callers can use this without having to check
+// this.
```
shoenig, line 1619: Was there a reason for not returning a nil set if we do return a non-nil error? Seems like all callers bail on an err without looking at the set.

shoenig (approved):
LGTM, just the one thought

PR #14983, opened 2022-10-20
deps: update go-memdb for goroutine leak fix

shoenig:
May as well slip this in for the bugfix releases


jrasell (approved):
LGTM!

Might want a changelog entry so it's called out and easy to find if someone experiences behaviour that fits.

PR #14987, opened 2022-10-20
keyring: fixes for keyring replication on cluster join

tgross:
Fixes https://github.com/hashicorp/nomad/issues/14981. Note to reviewers: these are definitely bugs but I don't have 100% confidence we've solved the problem because there's a log line missing in the reports from both the issue author and our internal user that I would expect to see. So I'm still trying to repro exactly, but it's worth getting some early eyes on this PR anyways.

* Don't unblock early if rate limit burst exceeded. The rate limiter returns an error and unblocks early if its burst limit is exceeded (unless the burst limit is Inf). Ensure we're not unblocking early, otherwise we'll only slow down the cases where we're already pausing to make external RPC requests.

* Set `MinQueryIndex` on stale queries. When keyring replication makes a stale query to non-leader peers to find a key the leader doesn't have, we need to make sure the peer we're querying has had a chance to catch up to the most current index for that key. Otherwise it's possible for newly-added servers to query another newly-added server and get a non-error nil response for that key ID.

* Note that the "not found" case does not return an error, just an empty key. Update the handling of empty responses so that we don't break the loop early if we hit a server that doesn't have the key. (Peers aren't shuffled so we'd expect to hit the same server repeatedly.)

* Move the keyring initialize step to wait until we're sure the FSM is current.

* If a key is rotated immediately following a leader election, plans that are in-flight may get signed before the new leader has the key. Allow for a short timeout-and-retry to avoid rejecting plans

### DIFF
--- a/nomad/encrypter.go
+++ b/nomad/encrypter.go
@@ -155,12 +155,33 @@ const keyIDHeader = "kid"
 // SignClaims signs the identity claim for the task and returns an
 // encoded JWT with both the claim and its signature
 func (e *Encrypter) SignClaims(claim *structs.IdentityClaims) (string, error) {
-	e.lock.RLock()
-	defer e.lock.RUnlock()
 
-	keyset, err := e.activeKeySetLocked()
+	getActiveKeyset := func() (*keyset, error) {
+		e.lock.RLock()
+		defer e.lock.RUnlock()
+		keyset, err := e.activeKeySetLocked()
+		return keyset, err
+	}
+
+	// If a key is rotated immediately following a leader election, plans that
+	// are in-flight may get signed before the new leader has the key. Allow for
+	// a short timeout-and-retry to avoid rejecting plans
+	keyset, err := getActiveKeyset()
 	if err != nil {
-		return "", err
+		ctx, cancel := context.WithTimeout(e.srv.shutdownCtx, 5*time.Second)
+		defer cancel()
+		for {
+			select {
+			case <-ctx.Done():
+				return "", err
+			default:
+				time.Sleep(50 * time.Millisecond)
+				keyset, err = getActiveKeyset()
+				if keyset != nil {
+					break
+				}
+			}
+		}
 	}
 
 	token := jwt.NewWithClaims(&jwt.SigningMethodEd25519{}, claim)
--- a/nomad/encrypter.go
+++ b/nomad/encrypter.go
@@ -464,7 +464,8 @@ START:
 				getReq := &structs.KeyringGetRootKeyRequest{
 					KeyID: keyID,
 					QueryOptions: structs.QueryOptions{
-						Region: krr.srv.config.Region,
+						Region:        krr.srv.config.Region,
+						MinQueryIndex: keyMeta.ModifyIndex - 1,
 					},
 				}
 				getResp := &structs.KeyringGetRootKeyResponse{}
@@ -482,7 +483,7 @@ START:
 					getReq.AllowStale = true
 					for _, peer := range krr.getAllPeers() {
 						err = krr.srv.forwardServer(peer, "Keyring.Get", getReq, getResp)
-						if err == nil {
+						if err == nil && getResp.Key != nil {
 							break
 						}
 					}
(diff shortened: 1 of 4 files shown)


tgross:
> Don't unblock early if rate limit burst exceeded. The rate limiter returns an error and unblocks early if its burst limit is exceeded (unless the burst limit is Inf). 

Aside: there's a couple other existing cases of this in `leader.go` which I want to fix as well, but I don't want to muddy up this PR with that.

tgross:
https://app.circleci.com/pipelines/github/hashicorp/nomad/33354/workflows/dbbd75a4-25cc-4d3d-9205-a817f7d68159/jobs/376989 shows what might be a reproduction of the original problem. I'm investigating that.

nomad/encrypter.go
```diff
@@ -435,7 +456,10 @@ START:
 			return
 		default:
 			// Rate limit how often we attempt replication
-			limiter.Wait(ctx)
+			err := limiter.Wait(ctx)
+			if err != nil {
+				goto ERR_WAIT // rate limit exceeded
+			}
```
shoenig, line 462: Seems like if we have our own blocking mechanism, we should be using `Allow` instead? Otherwise we are kind of double-waiting, which is fine but awkward given the error. (no need to change anything now)

https://pkg.go.dev/golang.org/x/time/rate#Limiter

shoenig (approved):
LGTM, just thinking about Limiters

nomad/encrypter.go
tgross: Yeah this API is weirdly hard to use correctly and the error case is _only_ for exceeding the burst. I first thought we probably wanted to use `Reserve` but that has the same double-waiting problem. Maybe we should just remove the burst (set it to `Inf`) so that we guarantee we'll wait and not have to handle this error so awkwardly?

tgross: I misread the API again 😊 . Only when the _rate_ is set to `rate.Inf` is the burst-limit ignored. So this is good as-is, awkward as it seems.

tgross:
I've got a clue that the core problem here is actually related to bugs in garbage collection, but I'm going to merge this in and push up a new PR with that fix once I've got that pinned down.

tgross:
Oops, forgot the changelog entry. Will add that to https://github.com/hashicorp/nomad/pull/15009

zansity:
Hello! Any idea when `1.4.2` will be released? One of my environments is currently unstable due to key rotations becoming "stuck", and would like to test the new fix. Thank you!

tgross:
> Hello! Any idea when 1.4.2 will be released? One of my environments is currently unstable due to key rotations becoming "stuck", and would like to test the new fix. Thank you!

I can't commit to exact dates but I can tell you we're in the process of finalizing the release now, so it'll be "very soon."

nomad/encrypter.go
```diff
@@ -155,12 +155,33 @@ const keyIDHeader = "kid"
...
-	e.lock.RLock()
-	defer e.lock.RUnlock()
 
-	keyset, err := e.activeKeySetLocked()
+	getActiveKeyset := func() (*keyset, error) {
+		e.lock.RLock()
+		defer e.lock.RUnlock()
+		keyset, err := e.activeKeySetLocked()
+		return keyset, err
+	}
```
schmichael, line 164: Sorry for the after-the-fact drive-by review:

Looks like we could replace `activeKeySetLocked` with this func as the only other place the key is read is `Encrypt` and it similarly only needs the lock for the duration of fetching the key.

The fact neither reader of this field need the lock other than for this field signals to me that the lock should be encapsulated in the getter and not the caller's concern.

Not a big deal though.

tgross: That's a good catch. `activeKeySetLocked` ends up holding onto the lock longer than it needs to as well -- it doesn't need to hold the lock while querying the state store. I'll open another quickie PR for that.

tgross: Follow-up PR: https://github.com/hashicorp/nomad/pull/15026

Issue #14787, opened 2022-10-03
Nomad UI painfully slow when job counts goes from hundreds to thousands


djenriquez:
### Nomad version
Output from `nomad version`
`Nomad v1.3.3 (428b2cd8014c48ee9eae23f02712b7219da16d30)`
### Operating system and Environment details
`Amazon Linux release 2 (Karoo)`

### Issue
We have a particular use case where Nomad is used to orchestrate full sandboxes for our developers in our development environment. These sandboxes represent our complete stack of services, which means ~100 jobs, including periodic batch jobs.

The higher the number of total jobs, the slower the Nomad UI becomes. Initially, we thought this might be an issue with the actual Nomad servers handling the sheer amount of work, but thats is not the case. Nomad's core is able to handle, at one point, over 10,000 jobs /w ~maybe 50,000 allocations just fine. RPC calls through its API were responsive and the metrics we track showed no struggle whatsoever.

However, the UI was a different story, as it would sit on the Nomad loader graphic for a period of time that seemed to grow linearly with the amount of jobs being run. Interestingly the API requests the UI made to the Nomad servers were responsive, according to chrome dev tools, providing supporting evidence that the backend is not the issue.

Also, when looking at the waterfall chart from chrome dev tools, we see a call to `/v1/namespaces?index=1` that eventually is canceled by the browser. Not sure if this request is misleading, but the page renders once that request pops up in the network analyzer, so it seems there is some blocking call at that part of the flow.
<img width="909" alt="Screen Shot 2022-10-03 at 1 34 34 PM" src="https://user-images.githubusercontent.com/4636740/193683706-31400b9a-c370-4e26-a306-40255dac6e2c.png">

### Reproduction steps
Spin up atleast 1000 jobs /w ~3000 allocations then navigate to the UI.

#### Expected Result
UI load time grows proportionately with the API response time for requests made to the Nomad server.

#### Actual Result
UI load time degrades as more jobs and allocations are running on the Nomad cluster while the API responds performantly.

We're open to scheduling a remote session if that makes it easier to see the issue.

philrenaud:
Hi @djenriquez, thanks for raising this — we'll take a look and update this once we have more info.

ChaiWithJai:
Hey @djenriquez! Nice to meet you. We're super grateful that you raised this issue and it looks like the Nomad Community at large is also noticing this problem.

[...]

Issue #10555, opened 2021-05-10
ui: Jobs list doesn’t show a new job (on a locally-built binary?)


backspace:
### Nomad version

```
Nomad v1.1.0-dev (b890a1f7eb298d25fcb39fd2adbc9fe3d8ee14c2+CHANGES)
```

### Operating system and Environment details

macOS 10.14.6

### Issue

When viewing the Nomad UI at `localhost:4646` on a locally-built binary, the jobs list doesn’t show a job newly submitted elsewhere. It’s not happening with a released binary 🤔

### Reproduction steps

Build a Nomad binary that includes the UI with `make dev-ui`, run it, open the jobs list at `localhost:4646/ui/jobs`, submit a job via CLI.

#### Expected Result

The job should appear in the jobs list.

#### Actual Result

The job only appears upon refresh, or by leaving and returning to the jobs list.

![jobs-no-live-reload](https://user-images.githubusercontent.com/43280/117687586-8ed0a980-b17d-11eb-880b-aae46e0d4cfd.gif)

herter4171:
@backspace, I was going to make my own issue until I saw this.  I am on Amazon Linux 2 with the Nomad version below, and we recently started using two head nodes instead of one.  What I have found is that I need to refresh the page for new jobs to show up, which wasn't necessary for a single head node.  I hope this gets resolved soon.

```
Nomad v1.1.2 (60638a086ef9630e2a9ba1e237e8426192a44244)
```

fideloper:
Adding to this - I updated Nomad from `0.12.*` to `1.2.6` and saw this behavior.

With 3 servers, and an auto-scaled number of clients (Ubuntu 18.04 AMD64), the Jobs UI does not refresh automatically any longer. I have to refresh the browser.

I'm not sure if this is Chrome-specific, perhaps when you leave the page it stops long-lived HTTP requests:

<img width="2010" alt="image" src="https://user-images.githubusercontent.com/467411/171225919-736c3fe6-e000-435a-9197-c4600973ff00.png">

I can see here that job requests seem to get cancelled. I usually keep the tab open forever-ish and check back occasionally.

josegonzalez:
I'm also seeing this locally, where it definitely wasn't an issue before. It makes debugging issues quite a bit more annoying as I typically have 5-6 tabs open as I trigger various actions in Nomad.

PR #14989, opened 2022-10-20
ui: fix long loading times on `/jobs/:jobId` page

ChaiWithJai:
This PR is part of resolving [14787]( https://github.com/hashicorp/nomad/issues/14787).

This PR does not solve long loading times on the `jobs.index` route. I believe we can and should use pagination and we can also patch [10555](https://github.com/hashicorp/nomad/issues/10555) while we do that to keep PRs small.


philrenaud (approved):

djenriquez:
> > Hi @ChaiWithJai, I'm looking at creating a local branch where I merge `v1.5.3` into this branch (since thats the release we're on internally). Do you foresee any incompatibilities with that?
> 
> Merging this branch might not be suitable for your needs as it does not solve the problem for the `jobs.index` page (The main Jobs list page). And I think that @philrenaud solves this same from this PR in his current block of work which we'll be out soon.

Ok, just to confirm, theres a PR that exists which we think will handle the problems in GH-14787?

philrenaud:
Hi @djenriquez — the PR that Jai is referring to is https://github.com/hashicorp/nomad/pull/17214 but that doesn't solve for the slow index page issue of https://github.com/hashicorp/nomad/issues/14787 — it's meant to make child jobs specifically less painful to use within the UI, but won't impact the /jobs index page.

The correct solution for that is for us to paginate `jobs` index requests; we're working on setting that up but have some plans to depend on the unpaginated "all jobs" request for some upcoming features. As of now, the plan is to set up a new "all jobs' summaries" request, and to use a paginated jobs request otherwise.

This seems like something worth prioritizing, so we'll kick off work related to it soon. We appreciate your patience on this!

djenriquez:
Completely understood @philrenaud , thank you for the reply.

philrenaud:
Made less necessary with the work in https://github.com/hashicorp/nomad/pull/20452

PR #14994, opened 2022-10-20
test: use port collision instead of cpu exhaustion

schmichael:
Originally this test relied on Job 1 blocking Job 2 until Job 1 had a terminal *ClientStatus.* Job 2 ensured it would get blocked using 2 mechanisms:

1. A constraint requiring it is placed on the same node as Job 1.
2. Job 2 would require all unreserved CPU on the node to ensure it would be blocked until Job 1's resources were free.

That 2nd assertion breaks if *any previous job is still running on the target node!* That seems very likely to happen in the flaky world of our e2e tests. In fact there may be some jobs we intentionally want running throughout; in hindsight it was never safe to assume my test would be the only thing scheduled when it ran.

*Ports to the rescue!* Reserving a static port means that both Job 2 will now block on Job 1 being terminal. It will only conflict with other tests if those tests use that port *on every node.* I ensured no existing tests were using the port I chose.

Other changes:
- Gave job a bit more breathing room resource-wise.
- Tightened timings a bit since previous failure ran into the `go test` time limit.
- Cleaned up the DumpEvals output. It's quite nice and handy now!

### DIFF
--- a/e2e/overlap/testdata/overlap.nomad
+++ b/e2e/overlap/testdata/overlap.nomad
@@ -10,11 +10,19 @@ job "overlap" {
   group "overlap" {
     count = 1
 
+    network {
+      # Reserve a static port so that the subsequent job run is blocked until
+      # the port is freed
+      port "hack" {
+        static = 7234
+      }
+    }
+
     task "test" {
       driver = "raw_exec"
 
       # Delay shutdown to delay next placement
-      shutdown_delay = "10s"
+      shutdown_delay = "8s"
 
       config {
         command = "bash"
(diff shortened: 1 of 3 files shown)


e2e/overlap/testdata/overlap.nomad
```diff
@@ -10,11 +10,19 @@ job "overlap" {
   group "overlap" {
     count = 1
 
+    network {
+      # Reserve a static port so that the subsequent job run is blocked until
+      # the port is freed
+      port "hack" {
+        static = 7234
+      }
+    }
```
schmichael, line 19: This is the entire fix.

jrasell (approved):
LGTM!

PR #14995, opened 2022-10-21
ci: try gotestsum in gha

shoenig:
This PR changes `make test-nomad` and `make test-nomad-module` to use `gotestsum` in GHA,
and combine the new behavior with Circle. Also bumps `gotestsum` to `v1.8.2`. Removes some
unused bits of GNUMakefile.

`gotestsum` is configured to retry failed individual test cases up to 3 times, and with a total of up to 10 retries across all tests (per runner) per run (which is the default).

The run below was set with a deliberately broken test to confirm real test failures still fail the build
https://github.com/hashicorp/nomad/actions/runs/3294509753

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -32,12 +32,6 @@ ifndef NOMAD_NO_UI
 GO_TAGS := ui $(GO_TAGS)
 endif
 
-ifeq ($(CIRCLECI),true)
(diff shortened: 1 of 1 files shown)


tgross (approved):
LGTM

jrasell (approved):

PR #15005, opened 2022-10-21
docs: remove stale prometheus / grafana integrations examples

shoenig:
These examples are > 5 years old and likely stale. A better resource are the Community Nomad Packs
- [prometheus](https://github.com/hashicorp/nomad-pack-community-registry/tree/main/packs/prometheus)
- [grafana](https://github.com/hashicorp/nomad-pack-community-registry/tree/main/packs/grafana)


jrasell (commented):
Agreed we should remove them; do we know if there are any links from Learn that we should handle as part of this?

Issue #15000, opened 2022-10-21
ci: cut down on cache spam


shoenig:
A screenshot is worth a thousand words. Currently `test-core` creates a Go module cache for each type of job - `check`, `api`, `compile * [windows, linux, macos]`, `api`, `package`, where each one is nearly a gigabyte. I'm fairly sure we can reduce this all to 1 cache if we be smart and have the `check` job run first, and also have it do a a `make bootstrap` which accounts for a large chunk of the cache size.

![Screenshot from 2022-10-21 08-51-45](https://user-images.githubusercontent.com/394887/197212137-d6650cd1-2ab7-48ae-b6bf-09c2e9bdb12a.png)

PR #15006, opened 2022-10-21
ci: use the same go mod cache across test-core jobs

shoenig:
This PR configures the `test-core` Action(?) with a new `mods` job on which the
other jobs depend on. The mods job just runs `mod tidy` and `bootstrap` to download
all the Go modules that will be needed and caches them. The dependent jobs then make
use of the same cache so they do not need to download / cache the near 1GB of modules
themselves.

Closes https://github.com/hashicorp/nomad/issues/15000

### DIFF
--- a/.github/workflows/test-core.yaml
+++ b/.github/workflows/test-core.yaml
@@ -40,13 +40,14 @@ jobs:
       - uses: magnetikonline/action-golang-cache@v1
         with:
           go-version: ${{env.GO_VERSION}}
-          cache-key-suffix: -checks
+          cache-key-suffix: -core
       - name: Run make check
         run: |
           make missing
           make bootstrap
           make check
   compile:
+    needs: [checks]
     strategy:
       fail-fast: false
       matrix:
@@ -58,7 +59,7 @@ jobs:
       - uses: magnetikonline/action-golang-cache@v1
         with:
           go-version: ${{env.GO_VERSION}}
-          cache-key-suffix: -compile
+          cache-key-suffix: -core
       - name: Run make dev
         env:
           GOBIN: ${{env.GOROOT}}/bin # windows kludge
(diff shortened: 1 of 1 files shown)


.github/workflows/test-core.yaml
```diff
@@ -58,22 +72,23 @@ jobs:
       - uses: magnetikonline/action-golang-cache@v1
         with:
           go-version: ${{env.GO_VERSION}}
-          cache-key-suffix: -compile
+          cache-key-suffix: -core
       - name: Run make dev
         env:
           GOBIN: ${{env.GOROOT}}/bin # windows kludge
         run: |
           make bootstrap
```
tgross, line 80: If we're caching the results of `make bootstrap` in `mods`, do we need this anymore?

shoenig: We do because GOBIN isn't part of the cache - the executables aren't cached across runners, only the GOMODCACHE and GOCACHE contents. https://github.com/magnetikonline/action-golang-cache/issues/1 mentions the possibility of adding GOBIN but hasn't gotten traction. The resolution of https://github.com/actions/setup-go/issues/278 might also impact what happens to the binaries.

.github/workflows/test-core.yaml
```diff
@@ -40,13 +53,14 @@ jobs:
       - uses: magnetikonline/action-golang-cache@v1
         with:
           go-version: ${{env.GO_VERSION}}
```
lgfa29, line 55: Not part of this work, so feel free to ignore it, but we could read the Go version from the `.go-version` file like the release workflow.

lgfa29 (approved):

.github/workflows/test-core.yaml
shoenig: yeah we should be able to set `go-version-file: go.mod` instead; i'm gonna try that in a follow up pr

PR #15009, opened 2022-10-21
keyring: fix missing GC config, don't rotate on manual GC

tgross:
For https://github.com/hashicorp/nomad/issues/14981

The configuration knobs for root keyring garbage collection are present in the consumer and present in the user-facing config, but we missed the spot where we copy from one to the other. Fix this so that users can set their own thresholds.

The root key is automatically rotated every ~30d, but the function that does both rotation and key GC was wired up such that `nomad system gc` caused an unexpected key rotation. Split this into two functions so that `nomad system gc` cleans up old keys without forcing a rotation, which will be done periodically or by the `nomad operator root keyring rotate` command.

### DIFF
--- a/.changelog/14987.txt
+++ b/.changelog/14987.txt
@@ -1,0 +1,15 @@
+```release-note:bug
+keyring: Fixed a bug where the root keyring replicator's rate limiting would be skipped if the keyring replication exceeded the burst rate.
+```
+
+```release-note:bug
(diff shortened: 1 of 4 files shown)


tgross:
Includes the changelog entry from https://github.com/hashicorp/nomad/pull/14987

shoenig (approved):
LGTM

PR #15015, opened 2022-10-21
ci: add -core suffix to mods action

shoenig:
Forgot to add this line to the new mods job; without it, it
creates a cache different from the one used by the other jobs. :facepalm:


tgross (approved):
LGTM

PR #15018, opened 2022-10-24
ci: use groups of tests in gha

shoenig:
(hold until after 1.4.2)

This PR changes our `Core CI Tests` GHA to run packages in groups of packages instead of one runner per package.

The plumbing is a bit more complex, so it's now documented in `ci/README.md`.

Grouping packages this way dramatically cuts down on the number of runners invoked by each `push` event. We were triggering ~50 concurrent runners; now we should use about 7. Packages are organized into groups as declared in `ci/test-core.json`. The idea is to create groups that are reasonably well balanced in terms of run time. We can tweak the groups going forward as needed.

The `tools/missing` tool has been updated to scan `ci/test-core.json` against Nomad source. It also now serves as a translator for group -> package list.

In a followup PR we can probably have the `drivers` group run on all platforms, eliminating the need for the Circle `test-windows` runner.


contributing/testing.md
```diff
@@ -6,26 +6,35 @@ demonstrating correct functionality.
...
 
 ## API tests
 
 Testing in the `api` package requires an already-built Nomad
 binary. If you're writing `api` tests, you'll need to build a Nomad
 binary (ex. with `make dev`) that includes any changes your API
 exercises.
+
+
+# CI Plubming
```
tgross, line 37:
```suggestion
# CI Plumbing
```

tgross (approved):
LGTM 👍  

I was worried about how long it was going to take but we're bottlenecked on the giant `nomad` package anyways right now; I can imagine if we were to break that up we'd be able to break up these groups a bit more as well.

PR #15026, opened 2022-10-24
keyring: refactor to hold locks for less time

tgross:
Follow-up from https://github.com/hashicorp/nomad/pull/14987/files#r1003611644

We don't need to hold the lock when querying the state store, so move the read-lock to the interior of the `activeKeySet` function.


shoenig (approved):
LGTM

Issue #14494, opened 2022-09-08
cgroups: unable to initialize cpuset manager on CentOS9


shoenig:
Out of the box, CentOS 9 is disabling `cpuset` (among other controllers) from being created by subgroups. We can probably turn it back on ourselves, but this is probably an artifact of them RH trying to push apps into using a systemd cgroups driver.

```
➜ cat /sys/fs/cgroup/cgroup.subtree_control 
cpu memory pids
```

```
➜ nomad version 
Nomad v1.3.5 (1359c2580fed080295840fb888e28f0855e42d50)
```

```
2022-09-08T07:46:30.088-0500 [WARN]  client.cpuset.v2: failed to lookup cpus from parent cgroup; disable cpuset management: error="openat2 /sys/fs/cgroup/nomad.slice/cpuset.cpus.effective: no such file or directory"
```

Procsiab:
Hello there, are you running the Nomad binary with a rootless account or with root privileges?

PR #15027, opened 2022-10-24
client: ensure minimal cgroup controllers enabled

shoenig:
This PR fixes a bug where Nomad could not operate properly on operating
systems that set the root `cgroup.subtree_control` to a set of controllers that
do not include the minimal set of controllers needed by Nomad.

Nomad needs these controllers enabled to operate:
- cpuset
- cpu
- io
- memory
- pids

Now, Nomad will ensure these controllers are enabled during Client initialization,
adding them to `cgroup.subtree_control` as necessary. This should be particularly
helpful on the RHEL/CentOS/Fedora family of system. Ubuntu systems should be
unaffected as they enable all controllers by default.

Fixes: https://github.com/hashicorp/nomad/issues/14494

Backports to 1.4.x and 1.3.x (1.2.x predates cgroups v2 support)

### DIFF
--- a/client/lib/cgutil/cpuset_manager_v2.go
+++ b/client/lib/cgutil/cpuset_manager_v2.go
@@ -56,20 +56,20 @@ type cpusetManagerV2 struct {
 
 func NewCpusetManagerV2(parent string, reservable []uint16, logger hclog.Logger) CpusetManager {
 	if err := minimumRootControllers(); err != nil {
-		logger.Error("failed to enabled minimum set of cgroup controllers; disable cpuset management", "error", err)
+		logger.Error("failed to enabled minimum set of cgroup controllers; disabling cpuset management", "error", err)
 		return new(NoopCpusetManager)
 	}
 
 	parentAbs := filepath.Join(CgroupRoot, parent)
 	if err := os.MkdirAll(parentAbs, 0o755); err != nil {
-		logger.Error("failed to ensure nomad parent cgroup exists; disable cpuset management", "error", err)
+		logger.Error("failed to ensure nomad parent cgroup exists; disabling cpuset management", "error", err)
 		return new(NoopCpusetManager)
 	}
 
 	if len(reservable) == 0 {
 		// read from group
 		if cpus, err := GetCPUsFromCgroup(parent); err != nil {
-			logger.Error("failed to lookup cpus from parent cgroup; disable cpuset management", "error", err)
+			logger.Error("failed to lookup cpus from parent cgroup; disabling cpuset management", "error", err)
 			return new(NoopCpusetManager)
 		} else {
 			reservable = cpus
@@ -95,14 +95,21 @@ func minimumRootControllers() error {
 	if err != nil {
 		return err
 	}
+
 	required := set.From[string]([]string{"cpuset", "cpu", "io", "memory", "pids"})
 	enabled := set.From[string](strings.Fields(s))
 	needed := required.Difference(enabled)
+
+	if needed.Size() == 0 {
+		return nil // already sufficient
+	}
+
 	sb := new(strings.Builder)
 	for _, controller := range needed.List() {
 		sb.WriteString("+" + controller + " ")
 	}
-	activation := sb.String()
+
+	activation := strings.TrimSpace(sb.String())
 	return e.write("cgroup.subtree_control", activation)
 }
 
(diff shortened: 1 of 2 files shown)


client/lib/cgutil/cpuset_manager_v2.go
```diff
@@ -80,6 +86,30 @@ func NewCpusetManagerV2(parent string, reservable []uint16, logger hclog.Logger)
 	}
 }
 
+// minimumControllers sets the minimum set of required controllers on the
+// /sys/fs/cgroup/cgroup.subtree_control file. Some systems like Ubuntu turn on
+// all controllers by default, and will be unaffected. Other systems like RHEL,
+// CentOS, Fedora turn of most controllers, and provide a default that excludes
+// controllers needed by Nomad. This helper ensures all of:
+// [cpuset, cpu, io, memory, pids]
+// are enabled.
```
shoenig, line 95:
```suggestion
// minimumControllers sets the minimum set of required controllers on the
// /sys/fs/cgroup/cgroup.subtree_control file - ensuring [cpuset, cpu, io, memory, pids]
// are enabled.
```

client/lib/cgutil/cpuset_manager_v2.go
```diff
@@ -80,6 +86,26 @@ func NewCpusetManagerV2(parent string, reservable []uint16, logger hclog.Logger)
...
+	s, err := e.read("cgroup.subtree_control")
+	if err != nil {
+		return err
+	}
+	required := set.From[string]([]string{"cpuset", "cpu", "io", "memory", "pids"})
+	enabled := set.From[string](strings.Fields(s))
+	needed := required.Difference(enabled)
+	sb := new(strings.Builder)
+	for _, controller := range needed.List() {
+		sb.WriteString("+" + controller + " ")
```
tgross, line 103: Is having a trailing ` ` ok or do we need to trim whitespace when we're done?

client/lib/cgutil/cpuset_manager_v2.go
```diff
@@ -54,16 +55,21 @@ type cpusetManagerV2 struct {
...
 func NewCpusetManagerV2(parent string, reservable []uint16, logger hclog.Logger) CpusetManager {
+	if err := minimumRootControllers(); err != nil {
+		logger.Error("failed to enabled minimum set of cgroup controllers; disable cpuset management", "error", err)
+		return new(NoopCpusetManager)
+	}
+
 	parentAbs := filepath.Join(CgroupRoot, parent)
 	if err := os.MkdirAll(parentAbs, 0o755); err != nil {
-		logger.Warn("failed to ensure nomad parent cgroup exists; disable cpuset management", "error", err)
+		logger.Error("failed to ensure nomad parent cgroup exists; disable cpuset management", "error", err)
```
tgross, line 65: Can we change these log entries to say "disabling cpuset management"? "disable" sounds like we're telling the cluster administrator to do it.

tgross (approved):

client/lib/cgutil/cpuset_manager_v2.go
shoenig: experimentally it's safe but yeah I can clean this up. can also avoid writing an empty string in the case there is nothing to write

shoenig:
spot check on CentOS 9 AMI

```
[centos@ip-172-31-72-243 ~]$ ./nomad node status -self -verbose | grep cpu 
cpu.arch                                 = amd64
cpu.frequency                            = 2200
cpu.modelname                            = AMD EPYC 7571
cpu.numcores                             = 2
cpu.reservablecores                      = 2
cpu.totalcompute                         = 4400
[centos@ip-172-31-72-243 ~]$ ./nomad node status -self -verbose | grep 'os\.'
os.name                                  = centos
os.signals                               = SIGSTOP,SIGTRAP,SIGUSR2,SIGHUP,SIGILL,SIGIO,SIGIOT,SIGABRT,SIGALRM,SIGPROF,SIGTSTP,SIGXFSZ,SIGKILL,SIGTERM,SIGWINCH,SIGXCPU,SIGTTIN,SIGTTOU,SIGUSR1,SIGNULL,SIGCONT,SIGPIPE,SIGQUIT,SIGFPE,SIGSEGV,SIGBUS,SIGINT,SIGSYS
os.version                               = 9

```

PR #15034, opened 2022-10-25
keyring: remove root key GC

tgross:
While trying to repro https://github.com/hashicorp/nomad/issues/14981 I discovered that the approach we're using to GC keys by timeboxing them based on allocation `CreateIndex` simply won't work. Allocations would need to have a reference to the key ID that signed their WI. The current check we have is based on allocation index but it's bugged: we can only GC inactive keys newer than the newest allocation, which is useless.

We only rotate keys every 30d by default, so remove the key GC for now until we can come up with a better long-term fix for this problem.

### DIFF
--- a/website/content/docs/configuration/server.mdx
+++ b/website/content/docs/configuration/server.mdx
@@ -200,15 +200,11 @@ server {
   rejoin the cluster.
 
 - `root_key_gc_interval` `(string: "10m")` - Specifies the interval between
-  [encryption key][] metadata garbage collections.
-
-- `root_key_gc_threshold` `(string: "1h")` - Specifies the minimum time that an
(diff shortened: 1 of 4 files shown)


website/content/docs/configuration/server.mdx
```diff
@@ -200,15 +200,11 @@ server {
   rejoin the cluster.
 
 - `root_key_gc_interval` `(string: "10m")` - Specifies the interval between
```
tgross, line 202: Note: I'm intentionally leaving the name here because we'll want to use this interval for GC once we circle back to it, and I don't want to have to add a new interval for that too. Same applies to the name of the core job emitted by the leader.

angrycub (approved):
LGTM

Issue #14768, opened 2022-10-01
templates: rewrite despite no content change if owner / group configured


harningt:
### Nomad version
Nomad v1.3.5 (1359c2580fed080295840fb888e28f0855e42d50)

### Operating system and Environment details
Fedora Linux clients and servers.
User Namespace configured docker instance as the primary use case.

### Issue
Adjusting the ownership / group of templates (so they are owned by the namespaced user vs "nobody") results in the template being re-rendered + update trigger every 3-5 minutes despite no content changes.

### Reproduction steps

 - Setup user namespaced client (I assume this is a requirement - maybe not).
 - Create a job that has a template that has a configured numeric uid/gid in the namespace (ex: 5550000 because the offset I configured for the namespaced user was 550000)
 
#### Expected Result
 - Files written out to match the prescribed uid/gid.
 - Files re-written on change to contents
 - `change_mode` action is run on contents change
 

#### Actual Result
 - Files written out to match the prescribed uid/gid.
 - Files re-written every 3-5 minutes (observed timestamp updates) with no content changes (sha1sum checked for sanity)
 -  `change_mode` action is run every 3-5 minutes

### Job file (if appropriate)

Reduced to bare minimum
```
job "test-template-issue" {
  type = "service"
  datacenters = ["dc1"]

  group "template-group" {
    task "template-task" {
      driver = "docker"

      resources {
        memory = 64
        cpu = 500
      }

      vault {
        policies = ["default"]
      }

      template {
        destination = "${NOMAD_SECRETS_DIR}/ca.pem"
        uid = "55500000"
        data = <<EOL
{{ with secret "pki-int/cert/ca_chain" }}{{ .Data.ca_chain }}{{ end }}
EOL
      }
      config {
        image        = "alpine:3.16"
        command     = "sleep"
        args        = ["infinity"]
      }
    }
  }
}
```

### Nomad Client logs (if appropriate)
```
Sep 30 22:09:04 client-host-1 nomad[3190]:     2022-09-30T22:09:04.881-0400 [INFO]  client.alloc_runner.task_runner.task_hook.logmon.nomad: opening fifo: alloc_id=f2ab3f86-c0fb-c1e0-82cd->
Sep 30 22:09:04 client-host-1 nomad[3190]:  client.alloc_runner.task_runner.task_hook.logmon.nomad: opening fifo: alloc_id=f2ab3f86-c0fb-c1e0-82cd-836984c1798b task=template-task path=/va>
[...]

PR #15045, opened 2022-10-26
[deps] Update consul-template dependency

angrycub:
Updates consul-template to https://github.com/hashicorp/consul-template/commit/90370e07bf621811826b803fb633dadbfb4cf287

Fixes https://github.com/hashicorp/nomad/issues/14768

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,8 +40,8 @@ require (
 	github.com/gorilla/websocket v1.5.0
 	github.com/gosuri/uilive v0.0.4
 	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0
-	github.com/hashicorp/consul-template v0.29.6-0.20221017191245-61e288a79cca
-	github.com/hashicorp/consul/api v1.15.2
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -266,20 +261,27 @@ require (
...
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+require (
+	cloud.google.com/go/compute v1.11.0 // indirect
+	cloud.google.com/go/iam v0.6.0 // indirect
+	github.com/googleapis/enterprise-certificate-proxy v0.2.0 // indirect
+	golang.org/x/lint v0.0.0-20210508222113-6edffad5e616 // indirect
+)
```
tgross, line 287: Can we move these up into the block above and then re-run go mod tidy? I dunno why the toolchain always adds them down here but it reduces merge conflicts with ENT if all the `// indirect` is in a single block.

tgross (approved):

tgross:
LGTM @angrycub

PR #15055, opened 2022-10-26
Top nav auth dropdown

### LINKED ISSUE #15051: UI: SSO: Top nav auth dropdown
image

philrenaud:
Resolves #15051 

Signed in: 
image

Signed out:
image

Active: 
image

### DIFF
--- a/ui/app/components/global-header.js
+++ b/ui/app/components/global-header.js
@@ -8,7 +8,28 @@ import { attributeBindings } from '@ember-decorators/component';
 export default class GlobalHeader extends Component {
   @service config;
   @service system;
+  @service token;
+  @service router;
 
   'data-test-global-header' = true;
   onHamburgerClick() {}
+
+  profileOptions = [
+    {
+      label: 'Authorization',
+      key: 'authorization',
+      action: () => {
+        this.router.transitionTo('settings.tokens');
+      },
+    },
+    {
+      label: 'Sign Out',
+      key: 'sign-out',
+      action: () => {
+        this.router.transitionTo('jobs.index');
+      },
+    },
+  ];
+
+  profileSelection = this.profileOptions[0];
 }
--- a/ui/app/components/global-header.js
+++ b/ui/app/components/global-header.js
@@ -10,6 +10,7 @@ export default class GlobalHeader extends Component {
   @service system;
   @service token;
   @service router;
+  @service store;
 
   'data-test-global-header' = true;
   onHamburgerClick() {}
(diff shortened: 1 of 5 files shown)


ui/app/components/global-header.js
```diff
@@ -8,7 +8,37 @@ import { attributeBindings } from '@ember-decorators/component';
...
+      action: () => {
+        this.token.setProperties({
+          secret: undefined,
+        });
+
+        // Clear out all data to ensure only data the anonymous token is privileged to see is shown
+        this.store.unloadAll();
+        this.token.reset();
+
+        // this.router.transitionTo('jobs.index');
```
ChaiWithJai, line 38: `nit`: Did you want to uncomment this?

ui/app/components/global-header.js
```diff
@@ -8,7 +8,37 @@ import { attributeBindings } from '@ember-decorators/component';
...
   @service config;
   @service system;
+  @service token;
+  @service router;
+  @service store;
 
   'data-test-global-header' = true;
   onHamburgerClick() {}
+
+  profileOptions = [
```
ChaiWithJai, line 18: `question`: Should this functionality be extracted from this component into a separate service? I think that could help with testability and maintainability. How do you decide when to include logic into the component and when to extract?

ui/app/templates/components/global-header.hbs
```diff
@@ -60,9 +60,27 @@
...
-      ACL Tokens
-    </LinkTo>
+    {{#if this.token.selfToken}}
+      <PowerSelect
+        {{keyboard-shortcut menuLevel=true pattern=(array "g" "p") }}
+        @options={{this.profileOptions}}
+        @onChange={{action (queue 
+          (fn (mut this.profileSelection))
+          this.profileSelection.action
+        )}}
```
ChaiWithJai, line 70: `learning opportunity`: This is unique. Is this an instance of why you want the ergonomic benefits of `Ember 4.5`? Why are you calling the `mut` helper here? 

Are we using it for the child component `PowerSelect` to mutate the value of a parent `GlobalHeader`?

So you're creating a piped function that first mutates the value of the `profileSelection` to display to the user when the option is selected in the dropdown menu and then you're dispatching `profileSelection.action` to fire after we set the `profileSelection` value? Do I have that correct? And this is ergonomically better because now we're not writing that function in the backing component class?

ChaiWithJai (approved):
Left a few questions, but LGTM.

ui/app/templates/components/global-header.hbs
philrenaud: This lets us attach actions to the `profileOptions` objects themselves. PowerSelect doesn't give us a "currentlySelectedOption" entity by default in its template, so we use `profileSelection` to store that.

We don't do this to show the selection to the user; we always just show the word "Profile" to them.

PR #15067, opened 2022-10-27
[no merge] ci: try large runners again

shoenig:


shoenig:
Probably shouldn't turn `Parallel` back on until we fix our tests to work concurrently 

just one example (they get hidden by gotestsum) 
```
   Error Trace:	/home/runner/work/nomad/nomad/nomad/client_agent_endpoint_test.go:570
   Error:      	Received unexpected error:
        	       RPC Error:: 500,cpu profiling already in use
   Test:       	TestAgentProfile_RemoteClient
```

PR #15068, opened 2022-10-27
Update alloc after reconnect and enforece client heartbeat order

### LINKED ISSUE #12680: disconnected clients: Reconnecting allocation can cause a replacement that is immediately stopped
### Nomad version

Nomad 1.3.0-beta1

Nomad Vagrantfile 
1 Server
2 Clients

### Issue

Creating a job that is constrained to a single node, can result in a second allocation being scheduled for that client being started and then immediately stopped when a disconnected client reconnects

### Reproduction steps

- `nomad run` the jobfile below that constrains by `node.unique.name`
- In a new session, start a watch on the job `watch nomad status <jobname>`
- Wait for the deployment to finish
- Start a watch for the job - you should see something similar to the following
- Simulate a network outage on the targeted client (e.g. `sudo iptables -I INPUT -s 192.168.56.11 -j DROP`)
- Wait for the allocation to transition to the unknown status
- Simulate a network reconnect for the targeted client (e.g. `sudo iptables -D INPUT -s 192.168.56.11 -j DROP`)
- When the allocation transition back to `running`, notice that a second allocation was started and immediately stopped
- Notice the original allocation did not stop

```shell
Allocations
ID                Node ID      Task Group  Version  Desired  Status       Created         Modified
2f8d72f2    94027db7   cache           0            stop       complete  2s ago            1s ago
043e25bd  94027db7   cache           0            run         running     1m44s ago    2s ago
```

#### Expected Result

The original allocation reconnects and no additional allocations are scheduled

#### Actual Result

An extra allocation was scheduled and then immediately stopped

### Job file (if appropriate)

```hcl
job "no-replace" {
  datacenters = ["dc1"]

  group "cache" {
    count = 1

    max_client_disconnect = "2m"

    constraint {
      attribute = "${node.unique.name}"
      value     = "nomad-client02"
      operator  = "="
    }

    network {
      port "db" {
        to = 6379
      }
    }

    task "redis" {
      driver = "docker"

      config {
        image = "redis:3.2"

        ports = ["db"]
      }

      resources {
        cpu    = 500
        memory = 256
      }
    }
  }
}
```

### LINKED ISSUE #14925: Job updates not applied to reconnected allocations

lgfa29:
When a Nomad client with at least one alloc with [max_client_disconnect](https://developer.hashicorp.com/nomad/docs/job-specification/group#max-client-disconnect) misses a heartbeat the leader [updates its status to `disconnected`](https://github.com/hashicorp/nomad/blob/v1.4.2/nomad/heartbeat.go#L167-L173) which results in [evals for all jobs that have an allocation in that client](https://github.com/hashicorp/nomad/blob/v1.4.2/nomad/node_endpoint.go#L499-L500). This eval is reconciled by the scheduler, and [updates the alloc `ClientStatus` to `unknown` and appends a new state value](https://github.com/hashicorp/nomad/blob/v1.4.2/scheduler/reconcile.go#L1337-L1344) to indicate the allocation is considered disconnected.

When the client reconnects again [its status is set back to `ready`](https://github.com/hashicorp/nomad/blob/v1.4.2/nomad/node_endpoint.go#L489) since the heartbeat succeeds. The `ClientStatus` of the allocation is still `unknown` as it may have failed while the client was disconnected. This status will only be updated once the client calls the `Node.UpdateAlloc` against a server, which will [overwrite the `unknown` `ClientStatus` value set in the server](https://github.com/hashicorp/nomad/blob/v1.4.2/nomad/node_endpoint.go#L1247) with the correct status.

From the leader perspective this is what happens:

node-disconnect drawio

The problem described in #14925 happens because any eval after this disconnect/reconnect flow happens (such as a job update) is essentially indistinguishable from the last eval in the diagram above: the node is `ready` and the alloc `ClientStatus` is `running`, but these two scenarios need to be handle differently and so we need to store something in state to be able to detect this difference.

Current implementation uses the presence of a [`TaskClientReconnected`](https://github.com/hashicorp/nomad/blob/v1.4.2/nomad/structs/structs.go#L8156-L8157) task event to detect if [an alloc needs to reconnect](https://github.com/hashicorp/nomad/blob/v1.4.2/nomad/structs/structs.go#L10401-L10422), but this event is still present even after the alloc reconnects, so the scheduler always consider the alloc as still reconnecting.

While testing the fix for #14925 I would often hit #12680, which prevented the issue from being triggered since the extra alloc would not be considered as "reconnecting", so I included a fix for #12680 in this PR as well since their root cause is similar.

Clients use three main RPC methods to communicate its state with servers:

  - `Node.GetAllocs` reads allocation data _from the server_ and writes it _to the client_.
  - `Node.UpdateAlloc` reads allocation _from from the client_ and writes them _to the server_.
  - `Node.UpdateStatus` writes the client status to the server and is used as the heartbeat mechanism.

These RPC methods are called periodically by the client, and independently from each other. The usual mental model is that clients heartbeat first with `Node.UpdateStatus` and then update their allocation data with `Node.UpdateAlloc`, but this is not always true. If these two operations are reversed the diagram above will look like this:

node-disconnect drawio2

The state at the second last eval (`Node updates allocs`) looks just like the state when the node missed its heartbeat (`Node misses heartbeat`): `Node.Status: disconnected` and `Alloc.ClientStatus: running`.

This PR addresses these problems in the following ways:
- Don't rely on task events since only a limited number of them are kept in state 
- Record when a allocation reconnects as an `AllocState` entry to keep it consistent with how disconnects are recorded
- Check if the alloc needs to reconnect by verifying the last `AllocState`, it it's `unknown` the alloc reconnect has not been processed yet
- Enforce that nodes register and heartbeat successfully before being able to update their allocs. This makes the reconnect flow easier to reason about and more predictable.

The commits are split by different work chunks:

- 882904bfd04ae9cae81f1e3b3d9a91cde651059c is the main fix for #14925, both for system and service jobs.
- 720513f00214a3004b120ce41457252f1d0c21a8 fixes #12680 by properly handling `running` allocs in disconnected clients
- 956a3fa93ee9b34a02cb686c2add6ff48db37869 enforces clients to heartbeat before they can update their allocs. While 720513f00214a3004b120ce41457252f1d0c21a8 would be enough to fix #12680 I noticed other edge cases that I was not able to fix in the scheduler.

This is an alternative fix for #14948.

Closes #14925
Closes #12680


nomad/node_endpoint.go
```diff
@@ -1150,6 +1152,26 @@ func (n *Node) UpdateAlloc(args *structs.AllocUpdateRequest, reply *structs.Gene
...
+	// Ensure the node is allowed to update allocs.
+	// The node needs to successfully heartbeat before updating its allocs.
+	nodeID := args.NodeID
+	if nodeID == "" {
+		// COMPAT 1.14
+		// Maintain backwards compatibility with clients that don't set the
+		// NodeID field in the request.
+		nodeID = args.Alloc[0].NodeID
+	}
+	node, err := n.srv.State().NodeByID(nil, nodeID)
```
lgfa29, line 1164: I wasn't sure how to handle this, but I think it's fine to expect that all allocations being sent here are from the same client?

An alternative would be to just skip this check of there is no node ID.

nomad/structs/structs.go
```diff
@@ -10393,27 +10397,22 @@ func (a *Allocation) LastUnknown() time.Time {
...
+// "unknown" and so the allocation did not reconnect yet.
+func (a *Allocation) NeedsToReconnect() bool {
+	disconnected := false
+
+	for i := len(a.AllocStates) - 1; i >= 0; i-- {
+		s := a.AllocStates[i]
+		if s.Field != AllocStateFieldClientStatus {
+			continue
 		}
-	}
```
lgfa29, line 10410: This assumes that states are order chronologically, which I think it's fine since entries are added using [`alloc.AppendState`](https://github.com/hashicorp/nomad/blob/v1.4.2/nomad/structs/structs.go#L10183-L10191)? Another option would be to traverse all values.

nomad/structs/structs.go
```diff
@@ -10393,27 +10397,22 @@ func (a *Allocation) LastUnknown() time.Time {
 	return lastUnknown.UTC()
 }
 
-// Reconnected determines whether a reconnect event has occurred for any task
-// and whether that event occurred within the allowable duration specified by MaxClientDisconnect.
-func (a *Allocation) Reconnected() (bool, bool) {
```
lgfa29, line 10398: I changed the return values here because the double boolean was a bit confusing and also kind of unnecessary since there's already a [`alloc.Expired`](https://github.com/hashicorp/nomad/blob/v1.4.2/nomad/structs/structs.go#L10354-L10382) method.

scheduler/reconcile_test.go
```diff
@@ -5282,6 +5315,7 @@ func TestReconciler_Disconnected_Client(t *testing.T) {
 		jobVersionIncrement          uint64
 		nodeScoreIncrement           float64
 		disconnectedAllocStatus      string
+		disconnectedAllocStates      []*structs.AllocState
```
lgfa29, line 5318: It ended up a bit repetitive, but I think it's important to highlight that `AllocStates` is also used for scheduling decisions.

scheduler/reconcile_util_test.go
```diff
@@ -791,37 +778,138 @@ func TestAllocSet_filterByTainted(t *testing.T) {
...
-			require.Equal(t, tc.lost, lost, "with-nodes: %s", "lost")
-			require.Equal(t, tc.disconnecting, disconnecting, "with-nodes: %s", "disconnecting")
-			require.Equal(t, tc.reconnecting, reconnecting, "with-nodes: %s", "reconnecting")
-			require.Equal(t, tc.ignore, ignore, "with-nodes: %s", "ignore")
+			assert.Equal(t, tc.untainted, untainted, "with-nodes: %s", "untainted")
+			assert.Equal(t, tc.migrate, migrate, "with-nodes: %s", "migrate")
+			assert.Equal(t, tc.lost, lost, "with-nodes: %s", "lost")
+			assert.Equal(t, tc.disconnecting, disconnecting, "with-nodes: %s", "disconnecting")
+			assert.Equal(t, tc.reconnecting, reconnecting, "with-nodes: %s", "reconnecting")
+			assert.Equal(t, tc.ignore, ignore, "with-nodes: %s", "ignore")
```
lgfa29, line 899: I switched to `assert` because these fields are all independent and it was annoying to find errors one by one 😅 . Now the test will print all failures at once.

nomad/node_endpoint.go
```diff
@@ -1150,6 +1152,26 @@ func (n *Node) UpdateAlloc(args *structs.AllocUpdateRequest, reply *structs.Gene
 		return fmt.Errorf("must update at least one allocation")
 	}
 
+	// Ensure the node is allowed to update allocs.
+	// The node needs to successfully heartbeat before updating its allocs.
+	nodeID := args.NodeID
+	if nodeID == "" {
+		// COMPAT 1.14
```
shoenig, line 1159: 1.4?

nomad/structs/structs.go
```diff
@@ -10393,27 +10397,22 @@ func (a *Allocation) LastUnknown() time.Time {
...
-			eventTime := time.Unix(0, taskEvent.Time).UTC()
-			if lastReconnect.IsZero() || lastReconnect.Before(eventTime) {
-				lastReconnect = eventTime
-			}
+// NeedsToReconnect returns true if the last known ClientStatus value is
+// "unknown" and so the allocation did not reconnect yet.
+func (a *Allocation) NeedsToReconnect() bool {
+	disconnected := false
+
+	for i := len(a.AllocStates) - 1; i >= 0; i-- {
```
shoenig, line 10405: does the reverse order matter? if so can you comment why

scheduler/reconcile_test.go
shoenig: you could make a little helper function to tidy up the repetition

shoenig (approved):
Awesome investigation, @lgfa29! I just had small comments

nomad/node_endpoint.go
lgfa29: Ooops 😅

scheduler/reconcile_test.go
lgfa29: Ah yeah, it was actually simpler to just create a variable with the most common value. Thanks!

lgfa29:
Ops, I found a regression while doing some extra testing where the alloc keeps running when the job is stopped while the client is still disconnected. I'm still investigating it.

lgfa29:
> Ops, I found a regression while doing some extra testing where the alloc keeps running when the job is stopped while the client is still disconnected. I'm still investigating it.

More race condition fun: f5ce8a96b3b9a7676c23fba78faeb583988a364a 😄 

This time between alloc reads and writes in the client. If the client reads the alloc state before writing it, it will trigger the `Reconnect` flow even if `DesiredStatus: stop`. While this was existing bug, by delaying the alloc status write this PR just made this problem more likely to happen.

nomad/structs/structs.go
```diff
@@ -950,6 +950,10 @@ type ApplyPlanResultsRequest struct {
 // to cause evictions or to assign new allocations. Both can be done
 // within a single transaction
 type AllocUpdateRequest struct {
+	// NodeID is the ID of the node sending the request.
+	// It may be empty if this is not an RPC request from a client.
```
schmichael, line 954: Hm I'd really like to avoid more conditionally-initialized fields as this is exactly what makes figuring out things like `stripped allocs sent from clients` so hard.

AFAICT this struct is used for *three mostly disjoint requests:*

1. Stripped allocs from clients Node.UpdateAlloc RPC using the Allocs, NodeID (new), and WriteRequest fields
2. Raft log message using the Allocs, Evals, and WriteRequest fields
3. Plan updates using the AllocsStopped, AllocsUpdated, and Job fields

1 & 2 using the same alloc is defensible but 3 just makes no sense these days. 3 isn't your problem though at all.

I can't think of a way to change this that I'd be comfortable backporting, so maybe let's put a big ugly comment here and file an issue to clean this up someday? I really don't want to hold up your PR with refactoring this mess.

scheduler/reconcile.go
```diff
@@ -1201,6 +1201,10 @@ func (a *allocReconciler) computeReconnecting(reconnecting allocSet) {
 			continue
 		}
 
+		// Record the new ClientStatus to indicate to future evals that the
+		// alloc has already reconnected.
+		alloc.AppendState(structs.AllocStateFieldClientStatus, alloc.ClientStatus)
```
schmichael, line 1206: Is this alloc a new alloc to be applied via a plan? Just making sure we're not mutating an object that's in the statestore.

schmichael (commented):
In the description:

> When a Nomad client misses a heartbeat the leader [updates its status to disconnected](https://github.com/hashicorp/nomad/blob/v1.4.2/nomad/heartbeat.go#L167-L173)

should probably read:

> When a Nomad client _with at least one alloc with [`max_client_disconnect`](https://developer.hashicorp.com/nomad/docs/job-specification/group#max-client-disconnect)_ misses a heartbeat...

Clients without an alloc with `max_client_disconnect` set will transition to `down` and allocs to `lost`. Minor pedantic point, but I suspect we may be linking to this PR in the future to explain this behavior so I think it's worth being precise. :grin: 

> Don't rely on task events since only a limited number of them are kept in state

:tada: I'm really happy to be keeping task events for humans only.

scheduler/reconcile_util.go
```diff
@@ -229,21 +229,25 @@ func (a allocSet) filterByTainted(taintedNodes map[string]*structs.Node, serverS
...
+		// Only compute reconnect for unknown, running, and failed since they
+		// need to go through the reconnect logic.
 		if supportsDisconnectedClients &&
 			(alloc.ClientStatus == structs.AllocClientStatusUnknown ||
 				alloc.ClientStatus == structs.AllocClientStatusRunning ||
 				alloc.ClientStatus == structs.AllocClientStatusFailed) {
-			reconnected, expired = alloc.Reconnected()
+			reconnect = alloc.NeedsToReconnect()
+			if reconnect {
+				expired = alloc.Expired(time.Now())
```
schmichael, line 243: filterByTainted has `time.Now()` passed in to help with testing. 
```suggestion
				expired = alloc.Expired(now)
```

scheduler/reconcile_util.go
```diff
@@ -258,6 +262,15 @@ func (a allocSet) filterByTainted(taintedNodes map[string]*structs.Node, serverS
 				if supportsDisconnectedClients {
 					// Filter running allocs on a node that is disconnected to be marked as unknown.
 					if alloc.ClientStatus == structs.AllocClientStatusRunning {
+						// Handle the case where the client updates its allocs
+						// before sending a heartbeat. The alloc will be set to
+						// running, but the client is still considered to be
+						// disconnected.
```
schmichael, line 268: I thought this was prevented by the node readiness check added in Node.UpdateAlloc?

scheduler/util.go
```diff
@@ -184,6 +195,20 @@ func diffSystemAllocsForNode(
 				node.Status == structs.NodeStatusDisconnected &&
 				exist.ClientStatus == structs.AllocClientStatusRunning {
 
+				// Handle the case where the client updates its allocs before
+				// sending a heartbeat. The alloc will be set to running, but
+				// the client is still considered to be disconnected.
```
schmichael, line 200: Same question as above: I thought this was prevented in the node endpoint changes?

Since the eval's WaitIndex is _at least_ its CreatedIndex, and it will have been created by the node heartbeat at the earliest, I don't think there's any chance of a "the statestore might not have applied the raft log the scheduler needs to make the right decision yet" issue, so as long as the node endpoint prevents this logic I'm not sure why we also need to prevent it here?

schmichael (approved):
Absolutely fantastic work from the repro to the description to the code! What a complex situation to debug and fixup.

An e2e test would be nice since there's so much emergent nondeterministic behavior at play. Your unit tests are great but come with some assumptions about the state of the data being tested that may not cover everything that can happen at runtime.

I forget if our `Test{Agent,Server}` make it possible to write an integration test which at least exercises these changes with a real `client.Client`.

The obvious problem with an e2e testing is how to simulate a disconnected client. Do I have a terrible idea for you! :sweat_smile: 

Schedule count=2 nomad agents via raw exec with the "fake" nomad's datacenter = `your-test-name`, and then schedule a job with `max_client_disconnect` set that targets that datacenter.

Grab the node ID of which of the 2 your alloc got scheduled on, send it a kill signal, let it restart via the parent Nomad, and then assert this intended behavior.

I think the nested-nomad agent should work, and the unique datacenter trick should prevent any flakiness from other jobs colliding with your careful placement scheme.

**Does _not_ block merging this though.** Probably best as a followup.

scheduler/reconcile_util.go
lgfa29: Yup, the RPC update is enough for this particular issue, but I was wondering if there was another scenario that we don't know about that could trigger this condition (like some data corruption? or bad alloc in state?) and so I kept the reconciler fix as well as a kind of layered defense. 

But I can remove this as well, less code to think about? 😅

scheduler/reconcile.go
lgfa29: Nice catch! I don't think it's being copied. Tracing it back, this `alloc` comes from these [`reconnecting[alloc.ID] = alloc`](https://github.com/hashicorp/nomad/blob/v1.4.2/scheduler/reconcile_util.go#L249). 

It's so easy to break stuff 😅

.changelog/15068.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:bug
+scheduler: Fixed a bug that prevented allocations with `max_client_disconnect` set to be updated after they reconnected.
+```
+
+```release-note:bug
+scheduler: Prevent unnecessary placements when allocations with `max_client_disconnect` set reconnect.
```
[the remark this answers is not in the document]
lgfa29: Yeah, makes sense. I didn't know how to call these either. I probably called them disconnected allocs in code comments as well 🤷

nomad/node_endpoint.go
```diff
@@ -1150,6 +1152,26 @@ func (n *Node) UpdateAlloc(args *structs.AllocUpdateRequest, reply *structs.Gene
...
 
+	// Ensure the node is allowed to update allocs.
+	// The node needs to successfully heartbeat before updating its allocs.
+	nodeID := args.NodeID
+	if nodeID == "" {
+		// COMPAT 1.4
+		// Maintain backwards compatibility with clients that don't set the
+		// NodeID field in the request.
+		nodeID = args.Alloc[0].NodeID
+	}
```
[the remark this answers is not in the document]
lgfa29: I tried to use `n.ctx.NodeID` but you're right, it's not present when the request is forwarded. I think it would take quite a bit of refactoring since we would need to pass it in the `n.src.forward` call and that affects all endpoints.

Another option would be to add it to [`InternalRPCInfo`](https://github.com/hashicorp/nomad/blob/v1.4.2/nomad/structs/structs.go#L228-L233) before forwarding the request, but I'm not sure if that's a good 😅

tgross (commented):
I haven't finished reviewing yet but wanted to surface my comments before I wrap up for the day 😀 . Will pick this back up first thing tomorrow.

scheduler/reconcile_util.go
lgfa29: That's what I meant by the layered approach. The RPC change fixes things externally but without this change the reconciler can still _in theory_  generate an invalid result given a these specific conditions. 

I would've liked to be able to fix everything in the scheduler, but I found one case for system jobs that I couldn't and only the RPC ordering fixed it, and I suspect there may be other edge cases like this, so I kept both approaches.

But I agree that a comment stating this should not happen would be good.

lgfa29:
I think I found a bug in the new `NeedsToReconnect` method that is preventing an alloc to disconnect twice. I will investigate further tomorrow.

nomad/node_endpoint.go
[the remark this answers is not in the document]
schmichael: I think if you just grab it before calling `forward` it would work, but that's a bit odd as forward is almost always the first thing we do.

What if we just dropped the field and always plucked it off the alloc? That would avoid adding another only-used-in-1-of-3-places field to the overloaded request struct. We already always have it on the stripped allocs, so we'd finally make that decision sensible.

We can ship it with the new NodeID field as its not worth holding up these excellent fixes for.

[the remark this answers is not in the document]
lgfa29: > I think if you just grab it before calling `forward` it would work

The problem with this is that the leader would then need to skip the validation, or more specifically, a server would need to ignore the check if the request was forward and assume that the check has been done at some point. 

I thought this would be a bigger problem for cross-region request since the first server would not find the node, but now that I think about it, it's kind of weird to have a cross-region request in this endpoint 😅 

> What if we just dropped the field and always plucked it off the alloc? 

Yeah, that sounds good to me. I wasn't sure which way to go any way and this is probably the most reliable option without adding more code 👍

scheduler/reconcile_test.go
```diff
@@ -5272,16 +5272,56 @@ func TestReconciler_Node_Disconnect_Updates_Alloc_To_Unknown(t *testing.T) {
 	})
 }
 
+func TestReconciler_Disconnect_UpdateJobAfterReconnect(t *testing.T) {
+	ci.Parallel(t)
+
+	// Create 2 allocs and simulate one have being previously disconnected.
```
tgross, line 5278:
```suggestion
	// Create 2 allocs and simulate one have being previously disconnected and then reconnected.
```

I think, based on these alloc states?

tgross (approved):
LGTM! (Once the docstrings are fixed up as being discussed)

nomad/node_endpoint.go
lgfa29: Accessing Node ID from allocs will be the default:
https://github.com/hashicorp/nomad/pull/15068#discussion_r1013503431

nomad/structs/structs.go
lgfa29: Removed the new field as per suggestion [above](https://github.com/hashicorp/nomad/pull/15068#discussion_r1013503431). Thanks!

scheduler/reconcile_util.go
lgfa29: I'm sold now 😄 

Adding code to the scheduler that we don't expect to hit just add stuff to an already complex part of the code, so I removed this check and its equivalent for `system` scheduling. Thanks for the feedback!

scheduler/reconcile_test.go
lgfa29: Oh that's right, good catch!

PR #15070, opened 2022-10-28
refactor eval delete safety check

tgross:
The `Eval.Delete` endpoint has a helper that takes a list of jobs and allocs and determines whether the eval associated with those is safe to delete (based on their state). Filtering improvements to the `Eval.Delete` endpoint are going to need this check to run in the state store itself for consistency.

Refactor to push this check down into the state store to keep the eventual diff for that work reasonable.


jrasell (approved):
LGTM!

PR #15073, opened 2022-10-28
[ui, sso] Handle token expiry 500s

philrenaud:
Part 1 of handling #15061 ; second part will involve giving a notification ~10m prior to expiry.

### Handles a recently-expired token
Redirects a user to the tokens page where they'll see:
image

### Handles a long-expired token (that's been garbage-collected)
image

### Lets you know when your token will be expiring
image

#### Implementation details
Nomad's UI has token-aware stuff happening in several places, at several scopes:
- Anything that ingests a token has access to that token's model. That's where I've chosen to add the `isExpired` computed property. It updates in realtime / reactively to javascript's `new Date()` object changing.
- Anything that specifically asks for it has access to the `Token service`. This both holds the token itself, has `fetch` and `reset` methods, etc. I've chosen to set a `tokenNotFound` property here, since by its nature there would be no model to set it on in the event that a token was not found. This is mostly used in the "token has been GC'd" case.
- The `/settings/tokens` route has a few methods on it, like the ability to fetch policies for a given token, etc. I've chosen to do very little here, instead opting for it to read needed props from the tokens service.

#### TODO
- [x] Tests for general redirect-on-token-expired
- [x] Tests for general redirect-on-token-not-found

### DIFF
--- a/.changelog/15073.txt
+++ b/.changelog/15073.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: redirect users to Sign In should their tokens ever come back expired or not-found
+```
--- a/ui/app/controllers/application.js
+++ b/ui/app/controllers/application.js
@@ -69,6 +69,11 @@ export default class ApplicationController extends Controller {
     return this.errorCodes.includes('500');
   }
 
+  @computed('errorCodes.[]')
+  get isExpiredTokenError() {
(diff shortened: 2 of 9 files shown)


pkazmierczak (approved):

jrasell (approved):
LGTM and I believe I understood it \o/

Issue #15014, opened 2022-10-21
[plugin] Plugin Client does not serialize MountPropagation.


stswidwinski:
### Nomad version
1.4.1

### Operating system and Environment details
Any unix kernel. Centos 8.

### Issue

The exec driver advertises the ability to specify mounts within the configuration:

https://github.com/hashicorp/nomad/blob/v1.4.1/drivers/exec/driver.go#L489

The mounts are defined using the following four properties:

https://github.com/hashicorp/nomad/blob/2aa7e66bdb526e25f59883952d74dad7ea9a014e/plugins/drivers/driver.go#L434-L439

However, the protobuf serialization of the mounts does not respect the mount propagation settings:

https://github.com/hashicorp/nomad/blob/e34fa583f9eade2e77759108cc1b97b33c069e22/plugins/drivers/proto/driver.proto#L557-L567
https://github.com/hashicorp/nomad/blob/e34fa583f9eade2e77759108cc1b97b33c069e22/plugins/drivers/utils.go#L326-L336

Resulting in any mount propagation defaulting to `""` which is translated to `rprivate`:

https://github.com/hashicorp/nomad/blob/e34fa583f9eade2e77759108cc1b97b33c069e22/drivers/shared/executor/executor_linux.go#L812-L835

### Reproduction steps

Attempt to mount anything using the `Mount` configuration and observe that within the private namespace created by `libcontainer` all mounts are `private` independent of the input setting.

#### Expected Result
The mount propagation is respected.

#### Actual Result
The mount propagation is not respected.

### Job file (if appropriate)

tgross:
Hi @stswidwinski! That might also be at the root of https://github.com/hashicorp/nomad-driver-podman/issues/192 Thanks for opening this!

lgfa29:
Closed by #15096

PR #15096, opened 2022-11-01
Add mount propagation to protobuf definition of mounts

stswidwinski:
A simple fix for https://github.com/hashicorp/nomad/issues/15014 and the related https://github.com/hashicorp/nomad-driver-podman/issues/192.

Tests pass for the `drivers` package.


stswidwinski:
cc: @tgross @ejweber

lgfa29 (changes requested):
Thanks for the contribution @stswidwinski 🙂 

Would you mind add tests for this change? I think adding `PropagationMode` to [`TestTaskConfigRoundTrip`](https://github.com/hashicorp/nomad/blob/v1.4.2/plugins/drivers/utils_test.go#L77-L82) may be a good one, but feel free to add other tests if you see fit.

We also require contributors to sing our CLA before merging work, so you could you this as well?

stswidwinski:
Sure thing! I'll do that on Monday -- I was looking for related roundtrip
tests but didn't find them before. Thanks for the pointer!

stswidwinski:
The CLA has been signed. Thanks for your patience @lgfa29

lgfa29 (approved):
I just pushed a commit to add a changelog entry for this. Thank you for the contribution and for taking care of the CLA 🙂

Issue #15090, opened 2022-10-31
Infinite memory growth in Nomad Agent and Nomad Client when using batch jobs.


stswidwinski:
### Nomad version
Nomad v1.3.1 (2b054e38e91af964d1235faa98c286ca3f527e56)
But this is broken on `v.1.4.2` as well.

### Operating system and Environment details
Unix.

### Issue

Evaluations and Allocations for Batch jobs are not garbage collected as long as the batch job is not stopped. The growth continues ad-infinitum causing two side-effects:

1. The RSS of the Scheduler grows linearly in the number of evals and allocs
2. The RSS of the Agent grows linearly in the number of allocs
3. The Agent's goroutine count grows linearly in the number of allocs.

There are only two ways to remove this memory growth:

1. Run `nomad system gc` (which removes any history and data)
2. Stop the job and start a new one

Both of which put the burden of cleanup on the end-user. Now, let me point to the exact lines of code which are responsible for the behavior and go through the logic in detail. First, the GC of evaluations and allocations for batch jobs is somewhat special-cased:

https://github.com/hashicorp/nomad/blob/main/nomad/core_sched.go#L303-L328

The evaluation is never garbage collected unless the corresponding job is fully stopped. `allowBatch` is somewhat moot because it is never `true` unless we force gc. Now, infinite growth of allocations would be bad, but would not cause the infinite memory growth on clients. I believe that this is the reasoning behind line 325 which attempts to GC old allocations:

https://github.com/hashicorp/nomad/blob/main/nomad/core_sched.go#L347-L357

However, please notice that an allocation is never "re-parented" between different jobs and the `.Job` object is a pointer. Hence, we always get the equality of `alloc.Job == job` which yields `alloc.Job.CreateIndex == job.CreateIndex`. Hence, this is never garbage collected. Please note that this is due to the fact that the `CreateIndex` is set only once during the first raft-level serialization of the object:

https://github.com/hashicorp/nomad/blob/c725c1b519389d737faa7aad1ed056f2cd47338d/nomad/state/state_store.go#L1643

The Job pointer of an allocation is similarly only ever set when the job is (de-)serialized for the first time:

https://github.com/hashicorp/nomad/blob/c725c1b519389d737faa7aad1ed056f2cd47338d/nomad/state/state_store.go#L3641

### History.

Please note that this logic has been broken for quite some time. In 2018 the logic shifted from using ModifyIndex to CreateIndex:

[...]

Issue #14842, opened 2022-10-07
Memory leak on nomad servers


D4GGe:
### Nomad version
`Nomad v1.3.1 (bc49012743edc0e2adb565e0fbcbc36c184f8ce4+CHANGES)`

### Operating system and Environment details

**3 nomad servers**

```
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.1 LTS
Release:	20.04
Codename:	focal

```
```
8 cores
32 GiB ram
```

**~10 large nomad clients** 

```
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.1 LTS
Release:	20.04
Codename:	focal

```
```
96 Cores
384 GiB ram
8 x nvidia t4
```
No nvidia plugin installed

**~20 nomad clients of various sizes**

### Issue
**Memory leak on nomad servers**

We have a memory leak on our nomad servers, the cluster have around 30 clients and the memory leak shows when we run a lot of of short lived jobs (parameterized job ~100 new allocs per hour), but seams to build up even if we lower the load just slower.

The only way for us to fix this is to run an `nomad system gc` on one of the servers. Otherwise it uses up all memory on the server until the server needs a reboot to be usable again. 
When this happens the only way to start nomad without nomad using 100% memory is to remove the `data_dir` and start over.

Right now in our production we run automatically `nomad system gc`  each time memory goes over 30% and this is how it looks.

![screenshot](https://user-images.githubusercontent.com/8104271/194497272-deb1a277-0629-43df-bcc8-279faf015511.png)

**Are there any way of analyzing what actually take up this amount of memory?
Can I analyze the `data_dir` in any meaningful way?
This log show up ever now and then can it have anything to do with it?**

```
2022-10-07T03:53:13.369Z [ERROR] worker: error waiting for Raft index: worker_id=599a4aeb-925e-0764-362f-844469f18e71 error="timed out after 5s waiting for index=23307547" index=23307547
```

### Configuration
***Server***
```HCL
data_dir = "/var/lib/nomad"
bind_addr = "0.0.0.0"
region = "{{region}}"
datacenter = "dc-master"
advertise {
	rpc = "{{ipify_public_ip}}"
}
server {
  enabled = true
  bootstrap_expect = 3
  encrypt = "{{gossipkey}}"
   server_join {
    retry_join = ["{{ipa}}", "{{ipb}}", "{{ipc}}"]
    retry_max = 0
    retry_interval = "15s"
  }
[...]

Issue #4532, opened 2018-07-25
Nomad doens't make GC for manualy stoped batch allocations


tantra35:
### Nomad version
Nomad v0.8.4 (dbee1d7d051619e90a809c23cf7e55750900742a)

### Issue
If we made stop foar batch jobs then launch it again throw `nomad run`. Old Allocations will not be gc, manual gc also doens' help

### Reproduction steps
for example we have test job:

```
job "test"
{
        datacenters = ["test"]
        type = "batch"

        constraint
        {
                attribute = "${attr.kernel.name}"
                value = "linux"
        }

        constraint
        {
                attribute = "${node.unique.name}"
                operator = "="
                value = "dockerworker-1"
        }

        task "diamondbcapacitycollector"
        {
                driver = "exec"

                config
                {
                        command = "sleep"
                        args = ["600"]
                }

                logs
                {
                        max_files = 3
                        max_file_size = 10
                }

                resources
                {
                        cpu = 100
                        memory = 300
                }
        }
}
```

then we launch it  throw `nomad run ./test.nomad`, then stop(`nomad stop test`) then again `nomad run ./test.nomad`, then again `nomad stop test` and finally `nomad run ./test.nomad`, so we have follow state of job

```
vagrant@consulnomad-1:~$ nomad status test
ID            = test
Name          = test
Submit Date   = 2018-07-25T16:34:18+03:00
Type          = batch
Priority      = 50
Datacenters   = test
Status        = dead
Periodic      = false
Parameterized = false

Summary
Task Group                 Queued  Starting  Running  Failed  Complete  Lost
diamondbcapacitycollector  0       0         0        0       3         0

Allocations
ID        Node ID   Task Group                 Version  Desired  Status    Created     Modified
7657bd07  2b5a3435  diamondbcapacitycollector  4        run      complete  10m43s ago  39s ago
490ccb3f  2b5a3435  diamondbcapacitycollector  2        stop     complete  11m30s ago  11m8s ago
324a29b9  2b5a3435  diamondbcapacitycollector  0        stop     complete  13m30s ago  11m8s ago
```

After that manipulations we try to made manual GC(we made it 2-3 times):

```
curl -XPUT http://localhost:4646/v1/system/gc
```

And nothing happens, old allocations doesn't clean. Only if we fully stop batch job GC for it will clean all allocations. All this we made on test stand, but on real environment we have following situations:

```
ID            = bobrovnik-a-jupiternoteBook
[...]

PR #15097, opened 2022-11-01
[15090] Ensure no leakage of evaluations for batch jobs.

stswidwinski:
Prior to https://github.com/hashicorp/nomad/commit/2409f72d0d42a3225d59347d2604f005bc132636 the code compared the modification index of a job to itself. Afterwards, the code compared the creation index of the job to itself. In either case there should never be a case of re-parenting of allocs causing the evaluation to trivially always result in `false`, which leads to memory leakage as reported here:

https://github.com/hashicorp/nomad/issues/15090
https://github.com/hashicorp/nomad/issues/14842
https://github.com/hashicorp/nomad/issues/4532

The comments within the code and tests are self-contradictory. Some state that evals of batch jobs should never be GCed others claim that they should be GCed if a new job version is created (as determined by comparing the indexes). I believe that the latter is the expected state.

Thus, we compare the creation index of the allocation with the modification index of the job to determine whether or not an alloc belongs to the current job version or not. 

This pull request specifically:

1. Removes references to `alloc.Job` which is unsafe given the deserialization of allocs from memDB which may include invalid pointers
2. Fixes the logical unity described above (compare creation with modification and not creation with creation)
3. Fixes the test (the test breaks the assumption of alloc reparenting, thus testing the mock rather than production code)
4. Adds more tests

Any and all feedback welcome.

For more details, please see the issues linked. Specifically https://github.com/hashicorp/nomad/issues/15090 describes the issue in detail.

EDIT:
I am assuming here that an update to the job results in an eval and consequently allocs for the job. If this is not correct than we must take into account the ordering of evals/allocs which has not been done prior.


stswidwinski:
cc: @jrasell

stswidwinski:
I need to test this a little bit more, but I believe that the current code actually references relatively random memory which is the result of dereferences of pointers which are serialized and de-serialized with the underlying object being non-static (that is: rellocable).

In particular the de-serialization proceeds by using the eval id as an index (https://github.com/hashicorp/nomad/blob/main/nomad/core_sched.go#L294). The GC logic is quite careful not to address pointers in structs except for this one place.

nomad/core_sched.go
```diff
@@ -344,16 +343,20 @@ func (c *CoreScheduler) gcEval(eval *structs.Evaluation, thresholdIndex uint64,
...
-func olderVersionTerminalAllocs(allocs []*structs.Allocation, job *structs.Job) []string {
+// olderVersionTerminalAllocs returns a tuplie ([]string, bool). The first element is the list of
+// terminal allocations which may be garbage collected for batch jobs. The second element indicates
+// whether or not the allocation itself may be garbage collected.
+func olderVersionTerminalAllocs(allocs []*structs.Allocation, job *structs.Job, thresholdIndex uint64) ([]string, bool) {
 	var ret []string
+	var mayGCEval = true
 	for _, alloc := range allocs {
-		if alloc.Job != nil && alloc.Job.CreateIndex < job.CreateIndex && alloc.TerminalStatus() {
+		if alloc.CreateIndex < job.JobModifyIndex && alloc.ModifyIndex < thresholdIndex && alloc.TerminalStatus() {
```
stswidwinski, line 353: Are these checks enough to ensure that the following scenario doesn't exhibit?

1. Batch job submitted
2. Eval created
3. Alloc created
4. Alloc finishes
5. Batch job updated
6. Is it possible for no new Eval/Alloc pair to correspond to the update?

D4GGe:
Any progress of merging this? Having huge problems in our environment with this!

stswidwinski:
At this point we're waiting for Nomad folks to reproduce and reason about
the initial reason for the existing code. Given that it's not uncommon for
bugs to hide bugs, I appreciate the scrutiny. I hope we will hear back soon.

Take a look at the issue attached to the PR for the discussion so far.

shantanugadgil:
@tgross Thanks for the update. :pray:  Looking forward to the backported releases of 1.4.x. :+1:

nomad/core_sched.go
tgross: I think so. `(*Job).JobModifyIndex` is only updated when the user-facing version of the job is updated (as opposed to `(*Job).ModifyIndex` which is updated whenever the job is updated in raft for any reason, like being marked complete or whatever.

nomad/core_sched.go
```diff
@@ -344,16 +347,20 @@ func (c *CoreScheduler) gcEval(eval *structs.Evaluation, thresholdIndex uint64,
 	return gcEval, gcAllocIDs, nil
 }
 
-// olderVersionTerminalAllocs returns terminal allocations whose job create index
-// is older than the job's create index
-func olderVersionTerminalAllocs(allocs []*structs.Allocation, job *structs.Job) []string {
+// olderVersionTerminalAllocs returns a tuplie ([]string, bool). The first element is the list of
+// terminal allocations which may be garbage collected for batch jobs. The second element indicates
+// whether or not the allocation itself may be garbage collected.
```
tgross, line 352:
```suggestion
// olderVersionTerminalAllocs returns a tuple ([]string, bool). The first element is the list of
// terminal allocations which may be garbage collected for batch jobs. The second element indicates
// whether or not the evaluation itself may be garbage collected.
```

Also, the caller can determine if its safe to GC the eval because the length of the `allocs` parameter has to match the length of the return value, right? And if there are no allocs left, is it safe to just go ahead and GC it without calling this function?

command/agent/config.go
```diff
@@ -467,9 +467,14 @@ type ServerConfig struct {
...
 	// EvalGCThreshold controls how "old" an eval must be to be collected by GC.
 	// Age is not the only requirement for a eval to be GCed but the threshold
-	// can be used to filter by age.
+	// can be used to filter by age. Please note that batch job evaluations are
+	// controlled by 'BatchEvalGCThreshold' instead.
 	EvalGCThreshold string `hcl:"eval_gc_threshold"`
 
+	// BatchEvalGCThreshold controls how "old" an evaluation must be to be eligible
+	// for GC if the eval belongs to a batch job.
+	BatchEvalGCThreshold string `hcl:"batch_eval_gc_threshold"`
```
tgross, line 476: There's a `ParseDuration` call we'll need to make in `command/agent.go` as well. See the equivalent for `EvalGCThreshold` at [agent.go#L359-L365](https://github.com/hashicorp/nomad/blob/v1.4.3/command/agent/agent.go#L359-L365) for an example

nomad/core_sched_test.go
```diff
@@ -439,26 +441,68 @@ func TestCoreScheduler_EvalGC_Batch_OldVersion(t *testing.T) {
...
+		allocsShouldExist []*structs.Allocation,
+		allocsShouldNotExist []*structs.Allocation,
+	) {
+		out, err := store.EvalByID(ws, eval.ID)
+		if err != nil {
+			t.Fatalf("err: %v", err)
+		}
+		if out == nil {
+			t.Fatalf("bad: %v", out)
+		}
```
tgross, line 484: We're migrating over to https://github.com/shoenig/test as we touch old tests and write new ones. For all the new assertions, let's use that.

```suggestion
		must.NoError(t, err)
		must.NotNil(t, out)
```

nomad/core_sched_test.go
```diff
@@ -439,26 +441,68 @@ func TestCoreScheduler_EvalGC_Batch_OldVersion(t *testing.T) {
...
-	// Update the time tables to make this work
-	tt := s1.fsm.TimeTable()
-	tt.Witness(2000, time.Now().UTC().Add(-1*s1.config.EvalGCThreshold))
+	// A little helper for assertions
+	assertCorrectEvalAlloc := func(
+		ws memdb.WatchSet,
+		eval *structs.Evaluation,
+		allocsShouldExist []*structs.Allocation,
+		allocsShouldNotExist []*structs.Allocation,
+	) {
```
tgross, line 477: This makes test outputs a lot more legible if a test fails:

```suggestion
	) {
	t.Helper()
```

nomad/core_sched_test.go
```diff
@@ -439,26 +441,68 @@ func TestCoreScheduler_EvalGC_Batch_OldVersion(t *testing.T) {
...
-	alloc3 := mock.Alloc()
-	job2 := job.Copy()
+	err = store.UpsertAllocs(structs.MsgTypeTestSetup, jobModifyIdx+2, []*structs.Allocation{alloc, alloc2})
+	if err != nil {
+		t.Fatalf("err: %v", err)
+	}
 
-	alloc3.Job = job2
-	alloc3.JobID = job2.ID
+	// Insert allocs with indexes older than job.ModifyIndex. Two cases:
```
tgross, line 449: For clarity, because of what we'll actually check:

```suggestion
	// Insert allocs with indexes older than job.ModifyIndex and job.JobModifyIndex. Two cases:
```

nomad/core_sched_test.go
```diff
@@ -467,44 +511,174 @@ func TestCoreScheduler_EvalGC_Batch_OldVersion(t *testing.T) {
...
+		t.Fatalf("err: %v", err)
+	}
+	if outB == nil {
+		t.Fatalf("bad: %v", outB)
+	}
+}
+
+// An EvalGC should reap allocations from jobs with a newer modify index and reap the eval itself
+// if all allocs are reaped.
+func TestCoreScheduler_EvalGC_Batch_OldVersionReapsEval(t *testing.T) {
```
tgross, line 565: These tests have a _lot_ of setup code (as I'm sure you noticed!) and the assertion is only subtly different than the previous one. It might make the whole test more understandable if we collapsed these into a single test and had this one be the final assertion that the eval is GC'd once the conditions are right.

tgross (commented):
Hi @stswidwinski! Thanks for your patience on this. I've given it a review and I've left a few comments to tighten up the code that'll hopefully make it easier for the next folks who read this :grin: 

But I feel pretty good about this approach overall. I'm getting towards the end of my week here so I'm going to give this one more pass early next week and hopefully we can wrap it up then. Thanks!

stswidwinski:
Sounds great. Acking the comment and I'll come back to fix up the code
early next week.

stswidwinski:
Please note that I cannot run the full suite of tests locally and must leave my computer at this time. If any tests are failing, expect that I will fix them before I ask for meaningful review and/or feedback.

stswidwinski:
@tgross, please let me know if there is anything outstanding from my end. I want to make sure that I'm not blocking you

shoenig:
Thanks for the effort @stswidwinski, just FYI most folks are out until January; we'll likely get this merged first thing then and cut a release shortly after

stswidwinski:
Happy new year! I'm wondering if we know when to expect another round of
review?

stswidwinski:
Happy new year! Hope you had a good time away :-) Thanks for the heads up!

tgross (changes requested):
Hi @stswidwinski! Again, sorry for the delay in reviewing this. It is a very tricky problem and as you mentioned above it's easy for bugs to hide bugs. 

Something that complicated testing this end-to-end which took a little while for me to rememeber is that the time table has a hard-coded 5min resolution. I've pulled down this PR locally and patched the time table out to something tighter to make testing faster. I've found one problem around what happens if you purge a job and the `batch_eval_gc_threshold` is lower than the `eval_gc_threshold`. But once that's fixed I think this is ready to go.

Once you've done that, I'll push a commit up to add a changelog entry and docs for the new config flag. But if you'd like to do that instead, there's a `make cl` command and the docs are at [server.mdx](https://github.com/hashicorp/nomad/blob/main/website/content/docs/configuration/server.mdx)

stswidwinski:
All fair points, thank you. I have considered the configuration conflict
you mention, but decided that the result is surprising in either case.
Thank you for resolving the problem!

I'll make the requested changes early next week as I'm already out for the
weekend. Have a good day!

lgfa29 (changes requested):
Looking good @stswidwinski!

I pushed a couple of commits to test the case that @tgross found where the `EvalGCThreshold` was actually affecting the GC of batch jobs, evals, and allocs.

I tested the suggestions I made and they do seem to fix the problem. The changes I made are in https://github.com/hashicorp/nomad/commit/0d80be04fa594201110383b1687734d5e555c8e5. Feel free to cherry-pick them if they look correct, but also feel free to fix the test with a different approach 🙂

nomad/core_sched.go
```diff
@@ -297,16 +304,21 @@ func (c *CoreScheduler) gcEval(eval *structs.Evaluation, thresholdIndex uint64,
...
-		// allowBatch and the job is dead
+		// Can collect if either holds:
+		//   - Job doesn't exist
+		//   - Job is Stopped and dead
+		//   - allowBatch and the job is dead
+		//
+		// If we cannot collect outright, check if a partial GC may occur
 		collect := false
 		if job == nil {
 			collect = true
```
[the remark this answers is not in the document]
stswidwinski: Agreed. Changed.

nomad/core_sched.go
```diff
@@ -246,7 +248,7 @@ func (c *CoreScheduler) evalGC(eval *structs.Evaluation) error {
 
 		// The Evaluation GC should not handle batch jobs since those need to be
 		// garbage collected in one shot
-		gc, allocs, err := c.gcEval(eval, oldThreshold, false)
+		gc, allocs, err := c.gcEval(eval, oldThreshold, batchOldThreshold, false)
```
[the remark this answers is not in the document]
stswidwinski: Agreed. Done.

nomad/core_sched.go
```diff
@@ -137,7 +137,7 @@ OUTER:
 		allEvalsGC := true
 		var jobAlloc, jobEval []string
 		for _, eval := range evals {
-			gc, allocs, err := c.gcEval(eval, oldThreshold, true)
+			gc, allocs, err := c.gcEval(eval, oldThreshold, oldThreshold, true)
```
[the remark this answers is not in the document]
stswidwinski: Agreed. Applied changes as suggested by @lgfa29

nomad/config.go
```diff
@@ -444,6 +451,7 @@ func DefaultConfig() *Config {
 		ReconcileInterval:                60 * time.Second,
 		EvalGCInterval:                   5 * time.Minute,
 		EvalGCThreshold:                  1 * time.Hour,
+		BatchEvalGCThreshold:             168 * time.Hour,
```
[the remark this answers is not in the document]
stswidwinski: I think it makes sense to have this value "close to" the value of EvalGCThreshold for some definition of "close to." 24 hours is definitely closer, so this makes sense to me. 

That said, I harbor no strong opinion since it seems to me that users will very likely have wildly different approaches to this value (ranging from: "we don't care" to "we care strongly and think that value X is the only correct one -- the former won't care what the default is and the latter will disagree with any value we pick).

Changed to 24 hours by default.

> I don't have any strong opinion on the default value, but looking at it again whatever value we pick should be noted in the upgrade guide since we're going from infinity to some finite time 

Given that this is going to be a bug fix ported onto other versions, where should a note about this behavior change be placed? At each version in which this is included or just one version?

Let me know if you think I should include a note about this in the current pull request, or if it's something to be added prior to the release being cut

stswidwinski:
> I tested the suggestions I made and they do seem to fix the problem. The changes I made are in https://github.com/hashicorp/nomad/commit/0d80be04fa594201110383b1687734d5e555c8e5. Feel free to cherry-pick them if they look correct, but also feel free to fix the test with a different approach slightly_smiling_face

This sounds great. I cherrypicked the changes on top.

stswidwinski:
Note: I'll add the changelog entries soon :-)

tgross (approved):
I've pulled down this branch and re-ran the bench testing I did previously. LGTM! Thanks for your patience in seeing this one through @stswidwinski!

stswidwinski:
:fireworks: Thank you for bearing with me

lgfa29 (approved):
Thanks for this work @stswidwinski!

I pushed a commit to expand the changelog entry a bit more since this will be a breaking change. 

For the upgrade note, I will add it in a separate PR since this one will be backported.

nomad/config.go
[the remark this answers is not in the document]
lgfa29: I will open a follow-up PR to add the upgrade notice since this one will be backported.

lgfa29:
Sorry, just noticed  a mistake in my changes 😅 

I pushed a commit to fix it.

PR #15099, opened 2022-11-01
build: update to go1.19.3

shoenig:
May as well grab this today; 1.19.2 was a bugfix release we should pick up anyway.


shoenig:
Circle: `docker.mirror.hashicorp.services/golang:1.19.3 not found: manifest unknown`

seems not even official image is out yet https://hub.docker.com/_/golang/tags?page=1&name=1.19

tgross (approved):
LGTM!

Issue #15095, opened 2022-11-01
panic in volume watcher


tgross:
Reported through our support org:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x110 pc=0x1d647f3]

goroutine 22850374 [running]:
github.com/hashicorp/nomad/nomad/volumewatcher.(*volumeWatcher).volumeReapImpl(0x0?, 0x0)
    github.com/hashicorp/nomad/nomad/volumewatcher/volume_watcher.go:208 +0x33
github.com/hashicorp/nomad/nomad/volumewatcher.(*volumeWatcher).volumeReap(0xc002c84f00, 0xc0020ea1a0?)
    github.com/hashicorp/nomad/nomad/volumewatcher/volume_watcher.go:193 +0x58
github.com/hashicorp/nomad/nomad/volumewatcher.(*volumeWatcher).watch(0xc002c84f00)
    github.com/hashicorp/nomad/nomad/volumewatcher/volume_watcher.go:121 +0x5f
created by github.com/hashicorp/nomad/nomad/volumewatcher.(*volumeWatcher).Start
    github.com/hashicorp/nomad/nomad/volumewatcher/volume_watcher.go:89 +0x125
```

This is bubbling up from [`volume_watcher.go#L208`](https://github.com/hashicorp/nomad/blob/v1.4.2/nomad/volumewatcher/volume_watcher.go#L208), which suggests the volume is nil here. A quick look at the code shows we're missing a nil-check at [`volume_watcher.go#L120-L121`](https://github.com/hashicorp/nomad/blob/v1.4.2/nomad/volumewatcher/volume_watcher.go#L120-L121)

tgross:
#15101 has been merged and will ship in the next release of Nomad, with backports to the 1.3.x and 1.2.x series.

PR #15101, opened 2022-11-01
volumewatcher: prevent panic on nil volume

tgross:
Fixes https://github.com/hashicorp/nomad/issues/15095

If a GC claim is written and then volume is deleted before the `volumewatcher` enters its run loop, we panic on the nil-pointer access. Simply doing a nil-check at the top of the loop reveals a race condition around shutting down the loop just as a new update is coming in.

Have the parent `volumeswatcher` send an initial update on the channel before returning, so that we're still holding the lock. Update the watcher's `Stop` method to set the running state, which lets us avoid having a second context and makes stopping synchronous. This reduces the cases we have to handle in the run loop.

Updated the tests now that we'll safely return from the goroutine and stop the runner in a larger set of cases. Ran the tests with the `-race` detection flag and fixed up any problems found here as well, and tested this on my local `democratic-csi` cluster.

### DIFF
--- a/.changelog/15101.txt
+++ b/.changelog/15101.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+csi: Fixed race condition that can cause a panic when volume is garbage collected
+```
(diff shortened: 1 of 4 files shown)


shoenig (approved):
LGTM!

PR #15102, opened 2022-11-01
variables: limit rekey eval to half the nack timeout

tgross:
In order to limit how much the rekey job can monopolize a scheduler worker, we limit how long it can run to 1min before stopping work and emitting a new eval. But this exactly matches the default nack timeout, so it'll fail the eval rather than getting a chance to emit a new one.

Set the timeout for the rekey eval to half the configured nack timeout.


schmichael (approved):

schmichael:
Does this need backport labels?

PR #15110, opened 2022-11-02
[ui, epic] SSO and Auth improvements

philrenaud:
(UI Epic branch)

### DIFF
--- a/ui/app/models/token.js
+++ b/ui/app/models/token.js
@@ -11,6 +11,11 @@ export default class Token extends Model {
   @attr('string') type;
   @hasMany('policy') policies;
   @attr() policyNames;
+  @attr('date') expirationTime;
 
   @alias('id') accessor;
+
+  get isExpired() {
+    return this.expirationTime && this.expirationTime < new Date();
+  }
 }
--- a/ui/app/routes/application.js
+++ b/ui/app/routes/application.js
@@ -13,6 +13,7 @@ export default class ApplicationRoute extends Route {
   @service system;
   @service store;
   @service token;
+  @service router;
 
   queryParams = {
     region: {
@@ -140,7 +141,17 @@ export default class ApplicationRoute extends Route {
   @action
   error(error) {
     if (!(error instanceof AbortError)) {
-      this.controllerFor('application').set('error', error);
+      if (
+        error.errors?.any(
+          (e) =>
+            e.detail === 'ACL token expired' ||
+            e.detail === 'ACL token not found'
+        )
+      ) {
+        this.router.transitionTo('settings.tokens');
+      } else {
+        this.controllerFor('application').set('error', error);
+      }
     }
   }
 }
--- a/ui/app/routes/settings/tokens.js
+++ b/ui/app/routes/settings/tokens.js
@@ -1,0 +1,12 @@
+// @ts-check
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class SettingsTokensRoute extends Route {
+  @service store;
+  model() {
+    return {
+      authMethods: this.store.findAll('auth-method'),
+    };
+  }
+}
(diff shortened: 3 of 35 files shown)


ui/app/models/token.js
```diff
@@ -11,6 +11,11 @@ export default class Token extends Model {
   @attr('string') type;
   @hasMany('policy') policies;
   @attr() policyNames;
+  @attr('date') expirationTime;
 
   @alias('id') accessor;
+
+  get isExpired() {
+    return this.expirationTime && this.expirationTime < new Date();
```
ChaiWithJai, line 19: `question`: `ember-data` sets the date attribute to use `ISO 8601` format whereas JavaScript's `Date` constructor gives us a Unix time. Should we have concern re:  date formatting in this comparison?

ui/app/routes/application.js
```diff
@@ -140,7 +141,17 @@ export default class ApplicationRoute extends Route {
   @action
   error(error) {
     if (!(error instanceof AbortError)) {
-      this.controllerFor('application').set('error', error);
+      if (
+        error.errors?.any(
+          (e) =>
+            e.detail === 'ACL token expired' ||
+            e.detail === 'ACL token not found'
```
ChaiWithJai, line 148: `question`: Does this mean we're hard coding the error message that we're expecting from the API here? And thus, if they added a period to the end of the message, we wouldn't be able to match here?

ui/app/routes/settings/tokens.js
```diff
@@ -0,0 +1,12 @@
+// @ts-check
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class SettingsTokensRoute extends Route {
+  @service store;
+  model() {
+    return {
+      authMethods: this.store.findAll('auth-method'),
+    };
```
ChaiWithJai, line 10: `nit`: If this is the only property returned on the model couldn't we just write:

```
async model() {
  return this.store.findAll('auth-method')
}
```

ui/app/routes/application.js
```diff
@@ -140,7 +141,17 @@ export default class ApplicationRoute extends Route {
...
+        )
+      ) {
+        this.router.transitionTo('settings.tokens');
```
ChaiWithJai, line 151: `question`: Should a flash message accompany this transition?

ChaiWithJai (approved):
LGTM. Left a few questions.

ui/app/routes/settings/tokens.js
philrenaud: For sure, but I kind of like the explicitness of asking for `this.model.authMethods` on a tokens page.

(If you asked me in a vacuum what `this.model` would iterate on a page called `tokens`, I would not assume it would be `auth-method`s, so IMO better to make it explicit)

ui/app/routes/application.js
philrenaud: Token errors (not found, expired) are shown as part of the [Tokens service](https://github.com/hashicorp/nomad/pull/15073/files#diff-06cbf9ecb213715e26c0786cc7972366d6dd1074a4063d4d785283f49eecd0cfR45-R47) — I've added screenshots and some explanations why I'm using errors in service/model instead of here in the application route in the PR notes: https://github.com/hashicorp/nomad/pull/15073#

ui/app/routes/application.js
philrenaud: That's right — this is a pretty fragile way of doing a match. Luckily this PR also includes visual diff and acceptance tests for this situation, so if this ever changes we should be getting plenty of alarm bells.

ui/app/models/token.js
philrenaud: Hm, this is a good concern to raise but in manual and acceptance testing I have found that the comparison works as expected.

PR #15114, opened 2022-11-02
[ui] if ACLs are disabled, remove the Sign In link from the top of the UI

philrenaud:


ui/app/templates/components/global-header.hbs
```diff
@@ -60,7 +60,9 @@
     >
       Documentation
     </a>
-    <ProfileNavbarItem />
+    {{#unless (eq this.system.agent.config.ACL.Enabled false)}}
```
lgfa29, line 63: I feel bad about nit-picking in a 3 line commit, but this double-double (:tim-hortons:) negative seems a bit confusing. Looking at the code I think it's because `this.system.agent.config.ACL` can be `null` while the user is not logged in because the service calls `v1/agent/self` and that would fail without a token?

Would it be better to have it as a template helper, then the template would be something like:

```hbs
{{#if aclEnabled}}
  <ProfileNavbarItem />
{{/if}}
```

lgfa29 (approved):
Minor nit-picking, feel free to ignore 👍

ui/app/templates/components/global-header.hbs
philrenaud: @lgfa29 this is a good point; will fix. Your assumption about the positive `if` condition failing (and thereby locking a user out of signing in if they had no token, since they wouldn't be able to load agent.config). 

I'll make it a computed property now

PR #15117, opened 2022-11-02
eval delete: move batching of deletes into RPC handler and state

tgross:
During unusual outage recovery scenarios on large clusters, a backlog of millions of evaluatons can appear. In these cases, the `eval delete` command can put excessive load on the cluster by listing large sets of evals to extract the IDs and then sending larges batches of IDs. Although the command's batch size was carefully tuned, we still need to be JSON deserialize, reserialize to messagepack, send the log entries through raft, and get the FSM applied.

To improve performance of this recovery case, move the batching process into the RPC handler and the state store. The design here is a little weird, so let's look at the failed options first:

* A naive solution here would be to just send the filter as the raft request and let the FSM apply delete the whole set in a single operation. Benchmarking with 1M evals on a 3 node cluster demonstrated this can block the FSM apply for several minutes, which puts the cluster at risk if there's a leadership failover (the barrier write can't be made while this apply is in-flight).

* A less naive but still bad solution would be to have the RPC handler filter and paginate, and then hand a list of IDs to the existing raft log entry. Benchmarks showed this blocked the FSM apply for 20-30s at a time and took roughly an hour to complete.

Instead, we're filtering and paginating in the RPC handler to find a page token, and then passing both the filter and page token in the raft log. The FSM apply recreates the paginator using the filter and page token to get roughly the same page of evaluations, which it then deletes. The pagination process is fairly cheap (only about 5% of the total FSM apply time), so counter-intuitively this rework of the pagination ends up being much faster. A benchmark of 1M evaluations showed this blocked the FSM apply for 20-30ms at a time (typical for normal operations) and completes in less than 4 minutes.

Note that, as with the existing design, this delete is not consistent: a new evaluation inserted "behind" the cursor of the pagination will fail to be deleted.

### DIFF
--- a/command/eval_delete.go
+++ b/command/eval_delete.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"strings"
-	"time"
 
 	"github.com/hashicorp/nomad/api"
 	"github.com/hashicorp/nomad/api/contexts"
@@ -148,32 +147,7 @@ func (e *EvalDeleteCommand) Run(args []string) int {
 		e.deleteByArg = true
 		exitCode, err = e.handleEvalArgDelete(args[0])
 	default:
-
-		// Track the next token, so we can iterate all pages that match the
-		// passed filter.
-		var nextToken string
-
-		// It is possible the filter matches a large number of evaluations
-		// which means we need to run a number of batch deletes. Perform
-		// iteration here rather than recursion in later function, so we avoid
-		// any potential issues with stack size limits.
-		for {
-			exitCode, nextToken, err = e.handleFlagFilterDelete(nextToken)
-
-			// If there is another page of evaluations matching the filter,
-			// iterate the loop and delete the next batch of evals. We pause
-			// for a 500ms rather than just run as fast as the code and machine
-			// possibly can. This means deleting 13million evals will take
-			// roughly 13-15 mins, which seems reasonable. It is worth noting,
-			// we do not expect operators to delete this many evals in a single
-			// run and expect more careful filtering options to be used.
-			if nextToken != "" {
-				time.Sleep(500 * time.Millisecond)
-				continue
-			} else {
-				break
-			}
-		}
+		exitCode, err = e.handleDeleteByFilter(e.filter)
 	}
 
 	// Do not exit if we got an error as it's possible this was on the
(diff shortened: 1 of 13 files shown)


command/eval_delete.go
```diff
@@ -404,3 +291,40 @@ func correctGrammar(word string, num int) string {
...
+	// if !e.yes && !e.deleteByArg {
+	// 	code, deleteEvals := e.askQuestion(fmt.Sprintf(
+	// 		"Are you sure you want to delete %v evals? [y/N]",
+	// 		len(evals)), "Cancelling eval deletion")
+	// 	e.Ui.Output("")
+
+	// 	if !deleteEvals {
+	// 		return code, nil
+	// 	}
+	// }
```
tgross, line 318: Open item for discussion: how do we feel about adding a `Eval.Count` RPC that does the exact same thing as `Eval.List` but returns a count instead of serializing potentially millions of evals to JSON?

jrasell: This sounds useful to me.

jrasell (approved):
LGTM; baring the lint failure and any updates based on adding count/len endpoint for evals.

nomad/eval_endpoint_test.go
```diff
@@ -1838,3 +1932,99 @@ func TestEvalEndpoint_Reblock(t *testing.T) {
 		t.Fatalf("ReblockEval didn't insert eval into the blocked eval tracker")
 	}
 }
+
+func TestEvalEndpoint_BigStateDelete(t *testing.T) {
```
tgross, line 1936: Oops, forgot to delete this! I was using it as a helper to build myself a snapshot for manual testing 😊

tgross:
As noted in https://github.com/hashicorp/nomad/pull/15120 I've got an import cycle here in the paginator tests, so I've got to fix that.

command/eval_delete.go
tgross: Implemented this in https://github.com/hashicorp/nomad/pull/15117/commits/0ecf61c5d480658e39c5dcd8e11804f4ea408b31 by adding a count field to the `Eval.List` RPC and reusing the same HTTP API with a new shape on `?count=true`. I don't exactly love this approach for the HTTP API, but it seems like it's better than doing `/v1/evaluations/count`?

tgross:
I've pulled the `Eval.Count` work out into https://github.com/hashicorp/nomad/pull/15147 and will rebase this PR on that once it's been merged.

command/eval_delete.go
tgross: Backed this out in lieu of https://github.com/hashicorp/nomad/pull/15147 and will rebase on that.

nomad/eval_endpoint.go
```diff
@@ -438,19 +438,40 @@ func (e *Eval) Delete(
 		return structs.ErrPermissionDenied
 	}
 
+	if args.Filter != "" && len(args.EvalIDs) > 0 {
```
schmichael, line 441: Should we reject this request if all servers don't support it?

tgross: Good catch, yes.

nomad/state/state_store.go
```diff
@@ -3143,6 +3144,69 @@ func (s *StateStore) updateEvalModifyIndex(txn *txn, index uint64, evalID string
...
+		}
+		raw := iter.Next()
+		if raw == nil {
+			break
+		}
+		eval := raw.(*structs.Evaluation)
+		if !pageTokenFound {
+			if eval.ID < pageToken {
+				continue
+			}
```
schmichael, line 3189: `pageTokenFound` is never set to `true`. I think you're missing:
```suggestion
		if !pageTokenFound {
			if eval.ID < pageToken {
				continue
			}
			pageTokenFound = true
```

schmichael (approved):
Don't forget to update docs.

nomad/state/state_store.go
tgross: A dumber mistake than that: it's not used because we don't need it. We'd just `continue` for each loop thru until we find the one we want and only then move on to deleting in the line below.

PR #15120, opened 2022-11-03
testing: fix import cycle in paginator test

tgross:
The benchmark in the paginator test somehow includes reference cycle that hasn't been previously detected in CI. Move this up into the `nomad/state` package.

@shoenig I think this implies that the check that we're running all our tests isn't working correctly for some of the nested packages?


shoenig:
@tgross I'm not able to reproduce an import cycle on `main`. Do you have some other changes or a `go.work` file doing things?

```
➜ git rev-parse HEAD
85f472189a2ebe4a96480a728c53c91df830ecad

nomad/state/paginator on main
➜ go test
PASS
ok      github.com/hashicorp/nomad/nomad/state/paginator        0.009s
```

```
➜ go test -bench=.
goos: linux
goarch: amd64
pkg: github.com/hashicorp/nomad/nomad/state/paginator
cpu: AMD Ryzen 5 PRO 5650G with Radeon Graphics
BenchmarkEvalListFilter/filter_with_index-12                 416           2794013 ns/op
BenchmarkEvalListFilter/filter_with_go-bexpr-12                8         140127661 ns/op
BenchmarkEvalListFilter/paginated_filter_with_index-12            177604              7179 ns/op
BenchmarkEvalListFilter/paginated_filter_with_go-bexpr-12           6981            170749 ns/op
BenchmarkEvalListFilter/paginated_filter_with_index_last_page-12                    1333            855848 ns/op
BenchmarkEvalListFilter/paginated_filter_with_go-bexpr_last_page-12                 6829            167608 ns/op
PASS
ok      github.com/hashicorp/nomad/nomad/state/paginator        86.463s
```

shoenig:
And we should have CI coverage over these, 

```
➜ cat ci/test-core.json | jq -r .quick[] | grep nomad/state
nomad/state/...
```

tgross:
Very weird... I could have sworn I was running that against `main`. But I can't repro on main either, only from branches cut from the branch I have for https://github.com/hashicorp/nomad/pull/15117, which I now see is [failing for the same reason](https://github.com/hashicorp/nomad/actions/runs/3385837896/jobs/5624485049):

```
$ git checkout eval-safe-delete-filter
Switched to branch 'eval-safe-delete-filter'

$ go test -v -count=1 ./nomad/state/paginator
# github.com/hashicorp/nomad/nomad/state/paginator
package github.com/hashicorp/nomad/nomad/state/paginator
        imports github.com/hashicorp/nomad/nomad/state: import cycle not allowed in test
FAIL    github.com/hashicorp/nomad/nomad/state/paginator [setup failed]
FAIL
```

Issue #15115, opened 2022-11-02
1.4.2: Jobs killed due to nomadService templating failures


iSchluff:
### Nomad version
Output from `nomad version`
Nomad v1.4.2 (039d70eeef5888164cad05e7ddfd8b6f8220923b)

### Operating system and Environment details
Linux nomad1 5.4.0-131-generic 147-Ubuntu 

### Issue
I have a set up where a job with multiple groups has various template cross-references using nomadService to dynamically fill in the address/port details.

With 1.4.x I see jobs getting killed because of template failures like these, even though the referenced service is running fine:
**alertmanager service**
``` 
Nov 02, '22 20:56:09 +0100 | Killed | Task successfully killed
Nov 02, '22 20:56:09 +0100 | Terminated | Exit Code: 0
Nov 02, '22 20:56:09 +0100 | Killing | Template failed: nomad.service(promteams): Unexpected response code: 403 (Permission denied)
Nov 02, '22 20:51:08 +0100 | Template | Missing: nomad.service(promteams)
Nov 02, '22 20:13:35 +0100 | Started | Task started by client
Nov 02, '22 20:13:35 +0100 | Task Setup | Building Task Directory
Nov 02, '22 20:13:35 +0100 | Received | Task received by client
```

**promteams service**
```
Nov 02, '22 17:39:44 +0100 | Started | Task started by client1
Nov 02, '22 17:39:43 +0100 | Driver | Downloading image
Nov 02, '22 17:39:43 +0100 | Task Setup | Building Task Directory
Nov 02, '22 17:39:42 +0100 | Received | Task received by client
```

Additionally sometimes a template will not update even though the nomad service changed.

#### Expected Result
Tasks should not be randomly killed, and templates should be properly updated when the service changes.

lgfa29:
Thanks for the report @iSchluff. 

I haven't been able to reproduce it yet. Would you happen to have any client and server logs around the time this happened?

lgfa29:
@iSchluff the team found the root cause for the `403` errors and #15121 has a fix for that.

But now I'm curious about the job failing due to the template render error. By default, Nomad should retry rendering forever, which is configured by the [`client.template.nomad`](https://developer.hashicorp.com/nomad/docs/configuration/client#nomad_retry) config. Did you happen to change this value?

iSchluff:
Hi @lgfa29, no the client config is completely default
```
client {
  enabled = true
  servers = ["{{ nomad_address }}"]
[...]

PR #15121, opened 2022-11-03
WI: set identity to client secret if missing

tgross:
For #15115

Allocations created before 1.4.0 will not have a workload identity token. When the client running these allocs is upgraded to 1.4.x, the identity hook will run and replace the node secret ID token used previously with an empty string. This causes service discovery queries to fail.

Fallback to the node's secret ID when the allocation doesn't have a signed identity. Note that pre-1.4.0 allocations won't have templates that read Variables, so there's no threat that this new node ID secret will be able to read data that the allocation shouldn't have access to.

I've smoke-tested this by creating a build of Nomad that doesn't sign identities in the plan applier, and verified that we can query service registrations as expected now.

### DIFF
--- a/.changelog/15121.txt
+++ b/.changelog/15121.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+wi: Fixed a bug where clients running pre-1.4.0 allocations would erase the token used to query service registrations after upgrade
+```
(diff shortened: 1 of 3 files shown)


lgfa29 (approved):
Testes locally in an upgrade scenario that failed before, and it works now. Thanks for the fix!

Issue #1113, opened 2016-04-20
Downloading artifact hosted in S3 returns a 403 despite IAM role being properly configured. 


c4milo:
### Nomad version
Output from `nomad version`
0.3.1
### Operating system and Environment details
Linux

### Issue
Downloading artifact hosted in S3 returns a 403 despite IAM role being properly configured. 

### Reproduction steps
1. Host a file in a private S3 bucket
2. Configure an EC2 instance with an IAM role that allows getting files from the bucket created in 1.
3. Schedule a nomad job with an artifact pointing to the mentioned file.

c4milo:
I also reported this issue at https://github.com/hashicorp/go-getter/issues/22

dadgar:
@c4milo I just vendored the change so if you build from master we can see if this is resolved 👍

c4milo:
@dadgar works well. Thank you Alex!

c4milo:
@dadgar, I'm seeing this problem again with 0.4.0

c4milo:
To make sure it wasn't my fault, I manually tested using https://github.com/rlmcpherson/s3gof3r and it downloaded the artifacts correctly. However, nomad does not and returns the following: 

![](https://cldup.com/wmp2zdpat6.png)

dadgar:
@c4milo Did anything change on your end since you marked it as closed? We haven't changed go-getter so wondering if there is another underlying bug in that library

c4milo:
@dadgar I haven't changed any of my configurations. I only upgraded nomad to the latest changes in master branch and redeployed it.

c4milo:
oh wait, I think the URL format for artifacts caught me again: 

Instead of: mybucket.s3.amazonaws.com/containers/vault.tar.gz

I kept using https://mybucket.s3.amazonaws.com/containers/vault.tar.gz

c4milo:
I'm closing this one in favor of https://github.com/hashicorp/nomad/issues/1219

PR #15123, opened 2022-11-03
Fix s3 example URLs in the artifacts docs

twunderlich-grapl:
Unfortunately, s3 urls prefixed with https:// do _NOT_ work with the underlying go-getter library. As such, this fixes the examples so that they are working examples that won't cause problems for people reading the docs. 

This is a known limitation of the go-getter lib since at least 2016 per https://github.com/hashicorp/nomad/issues/1113

I've also put up a PR with the underlying go-getter lib to fix their docs as well: https://github.com/hashicorp/go-getter/pull/394

### DIFF
--- a/website/content/docs/job-specification/artifact.mdx
+++ b/website/content/docs/job-specification/artifact.mdx
@@ -210,7 +210,7 @@ This example uses path-based notation on a publicly-accessible bucket:
 
 ```hcl
 artifact {
-  source = "my-bucket-example.s3-us-west-2.amazonaws.com/my_app.tar.gz"
+  source = "s3://my-bucket-example.s3-us-west-2.amazonaws.com/my_app.tar.gz"
 }
 ```
 
@@ -240,7 +240,7 @@ Alternatively you can use virtual hosted style:
 
 ```hcl
 artifact {
-  source = "my-bucket-example.s3-eu-west-1.amazonaws.com/my_app.tar.gz"
+  source = "s3://my-bucket-example.s3-eu-west-1.amazonaws.com/my_app.tar.gz"
 }
 ```
 
--- a/website/content/docs/job-specification/artifact.mdx
+++ b/website/content/docs/job-specification/artifact.mdx
@@ -210,7 +210,7 @@ This example uses path-based notation on a publicly-accessible bucket:
 
 ```hcl
 artifact {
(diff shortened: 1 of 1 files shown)


website/content/docs/job-specification/artifact.mdx
```diff
@@ -210,7 +210,7 @@ This example uses path-based notation on a publicly-accessible bucket:
 
 ```hcl
 artifact {
-  source = "https://my-bucket-example.s3-us-west-2.amazonaws.com/my_app.tar.gz"
+  source = "my-bucket-example.s3-us-west-2.amazonaws.com/my_app.tar.gz"
```
lgfa29, line 213:
```suggestion
  source = "s3://my-bucket-example.s3-us-west-2.amazonaws.com/my_app.tar.gz"
```

website/content/docs/job-specification/artifact.mdx
```diff
@@ -240,7 +240,7 @@ Alternatively you can use virtual hosted style:
 
 ```hcl
 artifact {
-  source = "https://my-bucket-example.s3-eu-west-1.amazonaws.com/my_app.tar.gz"
+  source = "my-bucket-example.s3-eu-west-1.amazonaws.com/my_app.tar.gz"
```
lgfa29, line 243:
```suggestion
  source = "s3://my-bucket-example.s3-eu-west-1.amazonaws.com/my_app.tar.gz"
```

lgfa29 (changes requested):
Thanks for the fix @twunderlich-grapl!

Per your findings and existing documentation ("non-Amazon S3-compatible endpoints like Minio are supported, but you must explicitly set the "s3::" prefix.") using the explicit `s3://` protocol may be the example that is most likely to work in different scenarios.

How do you feel about setting it in the docs?

website/content/docs/job-specification/artifact.mdx
twunderlich-grapl: I've updated this in a follow-up commit

twunderlich-grapl:
@lgfa29,
I've updated the example based on your suggestion. I agree that its the most likely way to avoid confusion since it will have an explicit protocol.

lgfa29 (approved):
Awesome, thanks!

PR #11983, opened 2022-02-02
cleanup: prevent leaks from time.After

shoenig:
This PR replaces use of time.After with a safe helper function
that creates a time.Timer to use instead. The new function returns
both a time.Timer and a Stop function that the caller must handle.

Unlike time.NewTimer, the helper function does not panic if the duration
set is <= 0.

Fixes #11982

PR #15134, opened 2022-11-03
Add missing timer reset

### LINKED ISSUE #15128: nomad.nomad.plan.queue_depth metric missing
### Nomad version

Nomad v1.3.6-zd84215+ent (fe6f9f9ed302ea13fffa4417375d2b89824a6ea8)

### Issue

We no longer get the nomad.nomad.plan.queue_depth metric since our upgrade from 1.2.6+ent. 

I believe the bug was introduced here:

https://github.com/hashicorp/nomad/pull/11983/files#diff-e76df9f1d1a93123add63e2828b97cbf4d55bd7fa0c42bb97d788bfe2fb598d0

In nomad/plan_queue.go we are missing a call to timer.Reset(period).

It is also not clear to me if calls to https://pkg.go.dev/time#Timer.Reset are being done safely. From those docs: "For a Timer created with NewTimer, Reset should be invoked only on stopped or expired timers with drained channels."

I do not completely understand why this has led to us not seeing the metric, as I'd have expected to still get a single point when the process started, even if the timer was not reset.

angrycub:
In applying the fixes in #11983, a timer.Reset was missed for the Plan Queue's EmitStats function.

Fixes #15128

### DIFF
--- a/.changelog/15134.txt
+++ b/.changelog/15134.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cleanup: fixed missing timer.Reset for plan queue stat emitter
+```
--- a/nomad/plan_queue.go
+++ b/nomad/plan_queue.go
@@ -201,6 +201,8 @@ func (q *PlanQueue) EmitStats(period time.Duration, stopCh <-chan struct{}) {
 	defer stop()
 
 	for {
+		timer.Reset(period)
+
 		select {
 		case <-timer.C:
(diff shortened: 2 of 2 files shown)


shoenig (approved):
LGTM!

lgfa29 (approved):
Nice catch! I wonder if semgre could be used to detect this 🤔

PR #15141, opened 2022-11-04
template: error on missing key

angrycub:
This PR adds the consul-template configuration option `error_on_missing_key`. This allows jobspec authors to specify that a template should fail if it references a map key that does not exist. The default value is `false` and should be fully backward compatible.

This PR does not prevent the template engine from performing normal blocking behavior for functions that use blocking queries. This solely addresses the missing key in a map case. The other cases can be managed by using a non-blocking alternative and the [`sprig_fail «message»` function](http://masterminds.github.io/sprig/flow_control.html)

This PR replaces #14002, just because it was easier to hand recreate it then to tease through the merge conflicts.


lgfa29 (approved):
LGTM! 

> The other cases can be managed by using a non-blocking alternative and the [sprig_fail «message» function](http://masterminds.github.io/sprig/flow_control.html)

An example of this would be a nice addition to the docs 🙂

PR #15145, opened 2022-11-04
e2e: explicitly wait on task status in chroot download exec test

shoenig:
Also add some debug log lines for this test, because it doesn't make sense
for the allocation to be complete yet a task in the allocation to be not
started yet, which is what the test failures are implying.

Adds `WaitForAllocTaskComplete` for convenience.


jrasell (approved):
LGTM, thanks!

PR #15147, opened 2022-11-04
API for `Eval.Count`

tgross:
Add a new `Eval.Count` RPC and associated HTTP API endpoints. This API is designed to support interactive use in the `nomad eval delete` command to get a count of evals expected to be deleted before doing so. (ref https://github.com/hashicorp/nomad/pull/15117)

The state store operations to do this sort of thing are somewhat expensive, but it's cheaper than serializing a big list of evals to JSON. Note that although it seems like this could be done as an extra parameter and response field on `Eval.List`, having it as its own endpoint avoids having to change the response body shape and lets us avoid handling the legacy filter params supported by `Eval.List`.

### DIFF
--- a/.changelog/15147.txt
+++ b/.changelog/15147.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+api: Added an API for counting evaluations that match a filter
+```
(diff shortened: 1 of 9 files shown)


pkazmierczak (approved):
LGTM!

PR #15155, opened 2022-11-04
[SSO] Auth Methods and Mock OIDC Flow

philrenaud:
This PR establishes the OIDC sign-in flow by implementing `auth-methods` and mocking a redirect flow based on which auth methods are present.

Running in Mirage, Nomad will generate 3 auth methods which can be used to simulate sign-in or sign-in failure.

image
image


ui/app/adapters/auth-method.js
```diff
@@ -0,0 +1,41 @@
+// @ts-check
+import { default as ApplicationAdapter, namespace } from './application';
+import { dasherize } from '@ember/string';
+import classic from 'ember-classic-decorator';
+
+@classic
+export default class AuthMethodAdapter extends ApplicationAdapter {
```
ChaiWithJai, line 7: `question`: I noticed we're not making any calls to `super` and thus we might not need to extend the `ApplicationAdapter` nor use the `classic` decorator as a result. Is it possible for us to create the adapter without it? What might be the drawbacks of my suggestion?

I believe that Adapters and Serializers may be [deprecated](https://github.com/emberjs/rfcs/pull/860) soon!

ui/app/controllers/oidc-mock.js
```diff
@@ -0,0 +1,32 @@
...
+
+  @action
+  failToSignIn() {
+    const url = `${this.redirect_uri.split('?')[0]}?state=failure`;
+    if (Ember.testing) {
+      this.router.transitionTo(url);
+    } else {
+      window.location = url;
+    }
+  }
```
ChaiWithJai, line 31: `question`: This controller isn't going to make it into production, and only our test builds, no? Are we planning to refactor this later to no longer be a mock?

ui/app/controllers/settings/tokens.js
```diff
@@ -66,4 +78,74 @@ export default class Tokens extends Controller {
...
+      })
+      .then(async (response) => {
+        if (response.ok) {
+          let json = await response.json();
+          this.token.set('secret', json.ACLToken);
+          this.verifyToken();
+          this.code = null;
+          this.state = null;
+        }
+      });
```
ChaiWithJai, line 145: `question`: Because we're using `.then` instead of `async/await` this callback will be fired once the request resolves and the rest of the thread of execution will resume? However, we have an `isValidatingToken` that will cause a conditional part of the template to render. I wonder what would happen if we simulate a slow network request. Could we trigger an unhandled error as a result? Would this be a good argument to advocate for `async/await` instead?

ui/app/router.js
```diff
@@ -98,4 +98,8 @@ Router.map(function () {
       path: '/path/*absolutePath',
     });
   });
+  // Mirage-only route for testing OIDC flow
+  if (config['ember-cli-mirage']) {
+    this.route('oidc-mock');
+  }
```
ChaiWithJai, line 104: `praise`: Wow, I didn't know you could do that!

ui/tests/acceptance/token-test.js
```diff
@@ -317,7 +315,8 @@ module('Acceptance | tokens', function (hooks) {
         .dom('.flash-message.alert-error')
         .doesNotExist('No notification yet for a token with 10m5s left');
       notificationNotRendered();
-      setTimeout(() => {
+      setTimeout(async () => {
+        await percySnapshot(assert);
```
ChaiWithJai, line 319: `thought`: Whoa! This is super interesting! Why did we do this?

ChaiWithJai (approved):
Great job! Left some questions so I could keep up with everything that's happening in here.

ui/app/controllers/oidc-mock.js
philrenaud: As it stands, this controller/route *will* be in the production binary, but because of the condition in the Router, will only exist if we're in Mirage mode — that is, no running agent / customer would see it.

As such, the most important thing here is ease of internal testing for this route in particular.

ui/app/adapters/auth-method.js
philrenaud: Interesting about the deprecation! I added the `@classic` decorator in the interest of getting our tests (which swallow up adapter errors) to progress but I'll take a look at `super`ing this.

ui/tests/acceptance/token-test.js
philrenaud: Thanks, I like this one a lot. We're testing that a soon-to-expire token generates an alert for the user.

So, we create a token that expires in 10min5sec, and check to make sure that there's no message
Then we wait 5sec, and check to make sure there IS a message. We take a screenshot for visual diffs.

ui/app/adapters/auth-method.js
philrenaud: Checked, and because we leverage `buildURL` and `ajax` methods from the application adapter, I don't think we can avoid the decorator.

ui/app/controllers/settings/tokens.js
philrenaud: Great minds think alike! A slow network request is exactly what we're simulating here: https://github.com/hashicorp/nomad/blob/60f7b6073d70ed2e07a105ebb56025bf48aa5eee/ui/mirage/config.js#L955

(timing: 1000 will wait a second before returning our token)

As for .thenning vs async/awaiting, I will make that switch! good suggestion.

PR #15203, opened 2022-11-10
ci: notify on backport-assistant errors

lgfa29:
I'm not sure how to test this properly, so we may need to iterate a bit. I manually tested as much as I could and you can see some samples in my test channel `#feed-nomad-alerts`

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -27,3 +27,43 @@ jobs:
           BACKPORT_LABEL_REGEXP: "backport/(?P<target>\\d+\\.\\d+\\.[+\\w]+)"
           BACKPORT_TARGET_TEMPLATE: "release/{{.target}}"
           GITHUB_TOKEN: ${{ secrets.ELEVATED_GITHUB_TOKEN }}
+  handle-failure:
+    needs:
+      - backport
+    if: always() && needs.backport.result == 'failure'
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yml
```diff
@@ -27,3 +27,43 @@ jobs:
...
+                          "type": "mrkdwn",
+                          "text": "*Pull Request:*\n<${{ github.event.pull_request.html_url}}|${{ github.repository }}#${{ github.event.pull_request.number}}>"
+                        },
+                        {
+                          "type": "mrkdwn",
+                          "text": "*From:*\n@${{ github.event.sender.login }}"
+                        },
+                        {
+                          "type": "mrkdwn",
+                          "text": "*Run:*\n<${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|${{ github.run_id }}>"
```
tgross, line 59: Is this supposed to be a link? If I look on Slack it shows as:

> </hashicorp/nomad-enterprise/actions/runs/1|1>

lgfa29: Ah yeah, those were the tests I did using [`act`](https://github.com/nektos/act), which don't have a GitHub Action metadata. Running from GitHub we get something like this:
image

tgross (approved):
LGTM!

PR #15204, opened 2022-11-10
ci: re-enable tests on main

lgfa29:
Now that the tests are grouped more tightly we don't use as many runners as before, so we can re-enable these without clogging the queue.


shoenig (approved):
LGTM!

PR #15212, opened 2022-11-10
docs: clarify how to access task meta values in templates

shoenig:
This PR updates template and meta docs pages to give examples of accessing
meta values in templates. To do so one must use the environment variable form
of the meta key name, which isn't obvious and wasn't yet documented.


tgross (approved):
LGTM

PR #15214, opened 2022-11-11
[bug] Return a spec on reconnect

### LINKED ISSUE #11096: network namespaces are not reconciled
### Nomad version
Output from `nomad version`
```
$ nomad version
Nomad v1.1.4 (acd3d7889328ad1df2895eb714e2cbe3dd9c6d82)
```

Linux

### Issue
Network namespaces associated to allocations that complete or fail while Nomad is stopped do not get cleaned up when Nomad starts back up and recovers the task.

This is a similar issue to https://github.com/hashicorp/nomad/issues/6385 but for network namespaces.

I am assuming this is due to https://github.com/hashicorp/nomad/blob/main/client/allocrunner/network_hook.go#L131-L140 not running when Nomad is stopped and the allocation completes.

### Job file
**sleep.nomad**
```
job "sleep" {
  datacenters = ["dc1"]

  group "sleep-group" {
    network {
      mode = "bridge"
    }
    task "sleep-task" {
      driver = "exec"

      config {
        command = "/bin/sleep
        args    = ["60"]
      }

      resources {
        cpu    = 500
        memory = 256
      }
    }
  }
}

```
### Reproduction steps

1. Run the job with `nomad job run sleep.nomad`
2. A network namespace gets created with the same id as the allocation
```
$ ip netns | grep 33a997ad
33a997ad-02f0-686b-8289-dcc9c4817711 (id: 4)
```
2. Stop nomad with `systemctl stop nomad`
3. Let the `sleep 60` task complete
4. Observe the network namespace from the completed task is still there
```
$ ip netns | grep 33a997ad
33a997ad-02f0-686b-8289-dcc9c4817711 (id: 4)
```
4. Start nomad `systemctl start nomad`
5. Observe the network namespace from the completed task is still there
```
$ ip netns | grep 33a997ad
33a997ad-02f0-686b-8289-dcc9c4817711 (id: 4)
```
6. Observe Nomad created a new allocation for the job
```
$ nomad job status sleep
[...]
Allocations
ID        Node ID   Task Group   Version  Desired  Status   Created    Modified
e08c8f0e  c1d19991  sleep-group  0        run      running  30s ago    18s ago
```
5. And that a new associated network namespace has been created in addition to the one from the now defunct previous allocation.
```
$ ip netns | grep e08c8f0e
e08c8f0e-6a29-4b09-2900-2dda8acb2075 (id: 5)
```

#### Expected Result
I would expect the network namespaces from previous allocations to be regularly reconciled and garbage collected.

#### Actual Result
The network namespace associated to the now defunct `33a997ad` allocation is still present on the machine

Feel free to let me know if you need any more information.
Thank you!

angrycub:
This PR changes the behavior on reconnect from the behavior introduced in #9757. 
https://github.com/hashicorp/nomad/blob/83b31ea44f7153a482de85992dcf77b52057047d/client/allocrunner/network_manager_linux.go#L100-L104

If the client enters this case and is returned a nil spec. When the network_hook's Postrun func is called to clean things up, the nil spec causes a fast exit.

https://github.com/hashicorp/nomad/blob/83b31ea44f7153a482de85992dcf77b52057047d/client/allocrunner/network_hook.go#L115-L118

This PR changes the behavior when we encounter the existing network namespace to return the spec so that the Postrun behavior will run properly to reap the namespace. 
 
Fixes #11096

### DIFF
--- a/.changelog/15214.txt
+++ b/.changelog/15214.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+client: fixed a bug where non-`docker` tasks with network isolation would leak network namespaces and iptables rules if the client was restarted while they were running
+```
--- a/client/allocrunner/network_manager_linux.go
+++ b/client/allocrunner/network_manager_linux.go
@@ -122,7 +122,18 @@ func (*defaultNetworkManager) CreateNetwork(allocID string, _ *drivers.NetworkCr
 			nsPath := path.Join(nsutil.NetNSRunDir, allocID)
 			_, err := os.Stat(nsPath)
 			if err == nil {
-				return nil, false, nil
+				// Let's return a spec that points to the tested nspath, but indicate
+				// that we didn't make the namespace. That will stop the network_hook
+				// from calling its networkConfigurator.Setup function in the reconnect
+				// case, but provide the spec value necessary for the network_hook's
+				// Postrun function to not fast exit.
+				spec := &drivers.NetworkIsolationSpec{
+					Mode:   drivers.NetIsolationModeGroup,
+					Path:   nsPath,
+					Labels: make(map[string]string),
+				}
+
+				return spec, false, nil
 			}
 		}
 		return nil, false, err


shoenig (approved):
LGTM! Be sure to add a CL entry

tgross (approved):
LGTM!

From https://github.com/hashicorp/nomad/pull/9757:
> safely handle

🤦 Guess not! What's insidious about this bug is that there's a "resources may have leaked" error just after the nil check in the `PostRun`, where we could have potentially caught this behavior.

PR #15221, opened 2022-11-11
acl: sso auth method RPC endpoints

pkazmierczak:
This PR implements RPC endpoints for SSO auth methods. 

This PR is part of the SSO work captured under ☂️ ticket https://github.com/hashicorp/nomad/issues/13120.

Consult NMD-155 RFC for documentation details.


jrasell (commented):
A couple of high level points before I dive into the code properly:

I think we can remove the changelog entry, this will be added once the feature is complete as a top level entry.

When checking for errors within a test, we can use `must.NoError` rather than a conditional check and call to `t.Fatalf`.

Can the entries into `structs/structs.go` be moved into `structs/acl.go` in order to keep the former file from growing even more?

nomad/acl_endpoint_test.go
```diff
@@ -2678,3 +2678,540 @@ func TestACL_GetRoleByName(t *testing.T) {
...
+	s1, root, cleanupS1 := TestACLServer(t, nil)
+	defer cleanupS1()
+	state := s1.fsm.State()
+	codec := rpcClient(t, s1)
+	testutil.WaitForLeader(t, s1.RPC)
+
+	// Create the authMethod
+	authMethod := mock.ACLAuthMethod()
+
+	// Upsert eval triggers watches
```
jrasell, line 3085:
```suggestion
	// Upsert auth-method triggers watches
```

nomad/acl_endpoint_test.go
```diff
@@ -2678,3 +2678,540 @@ func TestACL_GetRoleByName(t *testing.T) {
...
+		t.Fatalf("err: %v", err)
+	}
+
+	if elapsed := time.Since(start); elapsed < 100*time.Millisecond {
+		t.Fatalf("should block (returned in %s) %#v", elapsed, resp)
+	}
+	must.Eq(t, uint64(2), resp.Index)
+	if len(resp.AuthMethods) != 1 || resp.AuthMethods[0].Name != authMethod.Name {
+		t.Fatalf("bad: %#v", resp.AuthMethods)
+	}
```
jrasell, line 3111: Can we use must here?

nomad/acl_endpoint_test.go
```diff
@@ -2678,3 +2678,540 @@ func TestACL_GetRoleByName(t *testing.T) {
...
+
+	if elapsed := time.Since(start); elapsed < 100*time.Millisecond {
+		t.Fatalf("should block (returned in %s) %#v", elapsed, resp2)
+	}
+	must.Eq(t, uint64(3), resp2.Index)
+	must.Eq(t, 0, len(resp2.AuthMethods))
+}
+
+func TestACLEndpoint_DeleteAuthMethods(t *testing.T) {
```
jrasell, line 3134: Can we test some additional behaviour here, such as attempting to delete an auth method that does not exist?

nomad/acl_endpoint_test.go
```diff
@@ -2678,3 +2678,540 @@ func TestACL_GetRoleByName(t *testing.T) {
...
+	}
+	must.NotEq(t, uint64(0), resp.Index)
+
+	// Check we created the authMethod
+	out, err := s1.fsm.State().GetACLAuthMethodByName(nil, p1.Name)
+	must.Nil(t, err)
+	must.NotNil(t, out)
+}
+
+func TestACLEndpoint_UpsertACLAuthMethods_Invalid(t *testing.T) {
```
jrasell, line 3192: Can this test be added to the above test, so we avoid all the setup?

nomad/acl_endpoint_test.go
```diff
@@ -2678,3 +2678,540 @@ func TestACL_GetRoleByName(t *testing.T) {
...
+			Region:    "global",
+			AuthToken: root.SecretID,
+		},
+	}
+	var resp structs.GenericResponse
+	err := msgpackrpc.CallWithCodec(codec, structs.ACLUpsertAuthMethodsRPCMethod, req, &resp)
+	must.NotNil(t, err)
+	if !strings.Contains(err.Error(), "failed to parse") {
+		t.Fatalf("bad: %s", err)
+	}
```
jrasell, line 3216: Can we use must here?

nomad/structs/structs.go
```diff
@@ -12316,6 +12388,15 @@ func (a *ACLAuthMethod) Copy() *ACLAuthMethod {
 	return c
 }
 
+func (a *ACLAuthMethod) Validate() error {
+	var mErr multierror.Error
+
+	// TODO what are validity conditions for an auth method?
+	// ¯\_(ツ)_/¯
```
jrasell, line 12395: 😄 

Lets chat about this synchronously; but we probably want this to check all non-internally set fields have values and ensure they are not too big.

jrasell (changes requested):
Great first pass, thanks @pkazmierczak!

Along with the inline comments, the RPC write/delete handlers should version gate use of the feature such as [we do here for ACL roles](https://github.com/hashicorp/nomad/blob/1217a96edf7075c414f1496980944dbb5318352f/nomad/acl_endpoint.go#L1120-L1123).

Additionally within the tests there are a number of state calls which do not assign or check the error return. It would be nice to wrap these in `must.NoError`.

nomad/acl_endpoint.go
```diff
@@ -1667,3 +1667,263 @@ func (a *ACL) policyNamesFromRoleLinks(roleLinks []*structs.ACLTokenRoleLink) (*
...
+	}
+	defer metrics.MeasureSince([]string{"nomad", "acl", "get_auth_method_name"}, time.Now())
+
+	// Resolve the token and ensure it has some form of permissions.
+	acl, err := a.srv.ResolveToken(args.AuthToken)
+	if err != nil {
+		return err
+	} else if acl == nil {
+		return structs.ErrPermissionDenied
+	}
```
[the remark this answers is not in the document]
pkazmierczak: added a `|| !acl.IsManagement()` check if that's what you meant.

nomad/structs/structs.go
pkazmierczak: I added a simple name validation based on the same regexp policy names are validated against, and wrote a TODO note to revisit this in the future.

nomad/structs/acl.go
```diff
@@ -78,6 +79,40 @@ const (
...
 	ACLGetRoleByNameRPCMethod = "ACL.GetRoleByName"
+
+	// ACLUpsertAuthMethodsRPCMethod is the RPC method for batch creating or
+	// modifying auth methods.
+	//
+	// Args: ACLAuthMethodsUpsertRequest
+	// Reply: GenericResponse
+	ACLUpsertAuthMethodsRPCMethod = "ACL.UpsertAuthMethods"
+
+	// ACLDeleteAuthMethodsRPCMethod is the RPC method for batch deleting auth
```
jrasell, line 90:
```suggestion
	// ACLDeleteAuthMethodsByNameRPCMethod is the RPC method for batch deleting auth
```

nomad/structs/acl.go
```diff
@@ -517,3 +555,227 @@ type ACLRoleByNameResponse struct {
 	ACLRole *ACLRole
 	QueryMeta
 }
+
```
jrasell, line 558: Can we add some tests for the functionality from this point on?

nomad/structs/acl.go
```diff
@@ -78,6 +79,40 @@ const (
 	// Args: ACLRoleByNameRequest
 	// Reply: ACLRoleByNameResponse
 	ACLGetRoleByNameRPCMethod = "ACL.GetRoleByName"
+
+	// ACLUpsertAuthMethodsRPCMethod is the RPC method for batch creating or
+	// modifying auth methods.
+	//
+	// Args: ACLAuthMethodsUpsertRequest
+	// Reply: GenericResponse
```
jrasell, line 87: Thoughts about this, and ACLDeleteAuthMethodsByNameRPCMethod using custom structs rather than the generic?

jrasell (changes requested):
Lovely work, couple of additional questions, suggestions, and comments.

nomad/fsm.go
```diff
@@ -2069,6 +2073,36 @@ func (n *nomadFSM) applyACLRolesDeleteByID(msgType structs.MessageType, buf []by
...
+
+	if err := n.state.UpsertACLAuthMethods(index, req.AuthMethods); err != nil {
+		n.logger.Error("UpsertACLAuthMethods failed", "error", err)
+		return err
+	}
+
+	return nil
+}
+
+func (n *nomadFSM) applyACLAuthMethodsDelete(buf []byte, index uint64) interface{} {
```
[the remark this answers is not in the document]
pkazmierczak: Right, so this makes me think: should we just simplify `DeleteACLAuthMethodsByName` to `DeleteACLAuthMethods` (same for respective `Get`)? since we don't delete or get by IDs or anything else?

nomad/structs/acl.go
pkazmierczak: yup, updated comments.

nomad/fsm.go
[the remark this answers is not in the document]
jrasell: That's a good shout; mostly consistency across them will be good. Naming is hard!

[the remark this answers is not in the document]
pkazmierczak: I simplified by dropping `ByName` from all method names and RPCs like we agreed.

nomad/structs/acl.go
pkazmierczak: added.

tgross (commented):
This looks great @pkazmierczak!

(I'll let @jrasell mark for approval because he had some change requests, but LGTM in general)

PR #15227, opened 2022-11-11
keyring: update handle to state inside replication loop

tgross:
Fixes https://github.com/hashicorp/nomad/issues/14981

When keyring replication starts, we take a handle to the state store. But whenever a snapshot is restored, this handle is invalidated and no longer points to a state store that is receiving new keys. This leaks a bunch of memory too!

In addition to operator-initiated restores, when fresh servers are added to existing clusters with large-enough state, the keyring replication can get started quickly enough that it's running before the snapshot from the existing clusters have been restored.

Fix this by updating the handle to the state store when we query.

### DIFF
--- a/nomad/encrypter.go
+++ b/nomad/encrypter.go
@@ -18,10 +18,11 @@ import (
 	log "github.com/hashicorp/go-hclog"
 	kms "github.com/hashicorp/go-kms-wrapping/v2"
 	"github.com/hashicorp/go-kms-wrapping/v2/aead"
-	"golang.org/x/time/rate"
+	memdb "github.com/hashicorp/go-memdb"
 
 	"github.com/hashicorp/nomad/helper"
 	"github.com/hashicorp/nomad/helper/crypto"
+	"github.com/hashicorp/nomad/nomad/state"
 	"github.com/hashicorp/nomad/nomad/structs"
 )
 
@@ -434,15 +435,11 @@ func (krr *KeyringReplicator) stop() {
 const keyringReplicationRate = 10
 
 func (krr *KeyringReplicator) run(ctx context.Context) {
-	limiter := rate.NewLimiter(keyringReplicationRate, keyringReplicationRate)
 	krr.logger.Debug("starting encryption key replication")
 	defer krr.logger.Debug("exiting key replication")
 
-	retryErrTimer, stop := helper.NewSafeTimer(time.Second * 1)
-	defer stop()
-
-START:
 	store := krr.srv.fsm.State()
+	minIndex := uint64(0)
 
 	for {
 		select {
@@ -450,19 +447,26 @@ START:
 			return
 		case <-ctx.Done():
 			return
+		case <-store.AbandonCh():
+			store = krr.srv.fsm.State()
+			minIndex = uint64(0)
 		default:
-			// Rate limit how often we attempt replication
-			err := limiter.Wait(ctx)
-			if err != nil {
-				goto ERR_WAIT // rate limit exceeded
-			}
+			// ensure that we can abandon the blocking query periodically and
+			// give our other contexts a chance to fire
+			queryCtx, queryCancel := context.WithTimeout(ctx, time.Second)
+			defer queryCancel()
 
-			ws := store.NewWatchSet()
-			iter, err := store.RootKeyMetas(ws)
+			var rawIter any
+			var err error
+			rawIter, minIndex, err = store.BlockingQuery(getRootKeyMetas, minIndex, queryCtx)
 			if err != nil {
-				krr.logger.Error("failed to fetch keyring", "error", err)
-				goto ERR_WAIT
+				if queryCtx.Err() == nil {
+					// we get errors for closed context but don't want to log on that
+					krr.logger.Error("failed to fetch keyring", "error", err)
+				}
+				continue
 			}
+			iter := rawIter.(memdb.ResultIterator)
 
 			for {
 				raw := iter.Next()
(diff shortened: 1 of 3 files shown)


nomad/encrypter.go
```diff
@@ -457,6 +456,7 @@ START:
 				goto ERR_WAIT // rate limit exceeded
 			}
 
+			store := krr.srv.fsm.State()
 			ws := store.NewWatchSet()
```
tgross, line 460: I'm going to admit I feel like I'm supposed to be able to have this `store.NewWatchSet` at the top of the loop and then I can `select` on it with the rest of the context to replace the state handle... but I can't figure out which is the right thing to be polling on there and this diff fixes the test case I've got.

schmichael: So I think this can restructured a bit, but I could be wrong as I'm new to this area of code:

`(*state.StateStore).AbandonCh()` will get closed when a restore finishes, so I think just toss it in the select above and then you only need to do `store = krr.srv.fsm.State()` in that case block.

The `WatchSet` can be dropped and `nil` apssed t`WatchSets` are internal to iradix indexes while Abandoning is a StateStore concept. I think this is an accurate mental model:

```
Raft
 |
 v
FSM   <readers>
 |      ^
 v      |
StateStore <- Abandon
  |
  v
MemDB
  |
  v
iradix indexes <- Watches
```

If replicators were waiting for _their local root key meta to be updated via the FSM_ then I think the watchset could be added to the `select {}` to be notified when it was updated. Since the root key meta is replicated by RPC outside of the FSM, I don't think a watchset is useful here.

But like I said: new code to me, so let's maybe zoom if this all sounds way off.

schmichael (changes requested):

tgross:
> `(*state.StateStore).AbandonCh()` will get closed when a restore finishes, so I think just toss it in the select above and then you only need to do `store = krr.srv.fsm.State()` in that case block.

That's what I was looking for! Thank you!

> If replicators were waiting for _their local root key meta to be updated via the FSM_ then I think the watchset could be added to the select {} to be notified when it was updated. Since the root key meta is replicated by RPC outside of the FSM, I don't think a watchset is useful here.

The `RootKeyMeta` is replicated in the FSM; it's the root key material that's not (the random bytes that make up the key). So a watchset would normally be what we want here. But I couldn't figure out how to make this work when bringing up a fresh node. While the snapshot restores, any watchsets that fire will be on the new iradix tree which we're not yet watching on, so we'd never detect an update.

But I think we could do this:
* select on the state store abandon channel. when that channel is closed:
  * replace the `store` variable
  * start a new watchset
  * query for _all_ the keys like we do now
* select on the watchset. when that fires:
  * query for new keys like we do now (or given there's so few keys we can query for all of them as a belt-and-suspenders)
* select on a timer every few seconds. That allows for periodic attempts to repair if the on-disk keystore needs to be repaired by the cluster administrator after a botched recovery from backup.

That would let us lean on the efficiency of the watchset in the common case, while still handling new nodes and cases where the cluster needs to self-repair.

tgross:
@schmichael I've updated this to `select` over the abandon channel and to use a blocking query (to take advantage of the wait channel).

tgross:
Bah, I need to rework this, as it caused a panic during shutdown somewhere in https://github.com/hashicorp/nomad/actions/runs/3474236511/jobs/5807491485

nomad/encrypter.go
```diff
@@ -434,35 +435,42 @@ func (krr *KeyringReplicator) stop() {
...
-			// Rate limit how often we attempt replication
-			err := limiter.Wait(ctx)
+			// ensure that we abandon the blocking query periodically to give
+			// other contexts a chance to fire
+			queryCtx, queryCancel := context.WithTimeout(ctx, time.Second)
+			defer queryCancel()
+
+			var rawIter any
+			var err error
+			rawIter, minIndex, err = store.BlockingQuery(getRootKeyMetas, minIndex, queryCtx)
```
lgfa29, line 465: Hum..just thought of something, but I'm not sure if this how the blocking query works. 

Instead of a timeout could we instead check `abandonCh` after the query returns and loop back if the query returned because the snapshot was abandoned? So something like:

```go
for {
  select {
    // ...
    default: 
      // ...
      rawIter, minIndex, err = store.BlockingQuery(getRootKeyMetas, minIndex, ctx)
      select {
        case <-abandonCh:
          store = krr.srv.fsm.State()
          // ...
          continue 
        default:
      }
      // ...
}
```

lgfa29: Oh too bad. I thought the problem was that we were getting out of the blocking query too early 😅

tgross: I think there's either some very wrong with my implementation of the blocking query or there's a bug in `BlockingQuery`, because that's generating _many many_ queries rather than waiting on the wait channel as I'd expect. That in turn seems to be destableizing leadership in the unit tests.

In the interest of shipping this critical fix sooner rather than later, I've backed out the blocking query in lieu of a rate-limited one, and that seems to have stabilized the tests. Lemme do a little bit of cleanup of the resulting code and I'll flag this for review one last time.

tgross: What's really weird about this is that the timeout we implemented makes this very similar to the `blockingRPC` function, which sets a context that the `ws.WatchCtx` uses.

tgross:
@schmichael I've backed out the blocking query change and just left refreshing the `State()` call in the loop (and clean up some of the rest of the code in the process). There's something I'm missing with blocking queries where everything would be working as I'd expect -- the blocking query would return on either an update or the context timeout. But then it'd randomly get into a state where it'd enter a tight spin loop and ignore the context timeout entirely. This didn't seem to be correlated with the snapshot restore either. 

This tight loop was destabilizing the whole test suite, as it was causing enough load that leadership was failing. In the interest of shipping a critical fix I'm using the rate-limited query and I'll see about revisiting blocking queries in later work -- I have a sneaking suspicion there's may be a bug in our blocking query implementation that's hard to hit in RPCs. You've got a request-changes on this so I'll need your 👍 here to ship.

shoenig (approved):
LGTM! `goto -= 3` :tada:

schmichael (approved):

PR #15242, opened 2022-11-14
fix create snapshot request docs

ayrat555:


jrasell (commented):
Hi @ayrat555 and thanks for raising this PR. While we are there, would you be able to also update the sample request, as this also looks incorrect?

jrasell (approved):
LGTM, thanks @ayrat555!

Issue #10830, opened 2021-06-29
Topology view access denied until selecting namespace with at least one job


optiz0r:
### Nomad version

`1.1.2+ent`

### Issue

The topology ui view shows an access denied error if the selected namespace (typically `default` on first access) has no defined jobs. This caused some wild goose chasing through policies. Selecting a namespace on the jobs view, or crafting the URL to contain a namespace where that namespace has at least one job defined then displays the topology view as expected.

### Reproduction steps

- Possibly have namespaces enabled
- Have no jobs running in default namespace
- Try to navigate to the topology view, get access denied error
- Switch to another namespace which does contain a job
- Try to navigate to topology view, now works

#### Expected Result

Access denied error message not displayed if the user does indeed have the right permissions.
Blank topology to be displayed if no jobs are visible to user.

#### Actual Result

Access denied view, suggesting user authenticate with a different token.

shoenig:
Thanks for reporting, @optiz0r ! I'll hand this over to our UI team to look into.

ChaiWithJai:
Should be [resolved](https://github.com/hashicorp/nomad/pull/15244) now.

PR #15244, opened 2022-11-14
refact: remove unused API

ChaiWithJai:
Related to [10830](https://github.com/hashicorp/nomad/issues/10830).

The Topology Visualization makes a request to the `/jobs` endpoint and by default, requests all jobs in the `default` namespace. Upon investigating the issue, there is no usage of `model.jobs` in the controller, nor the template. It appears that we may not need this API call.


philrenaud (approved):
Nice find!

PR #15248, opened 2022-11-15
Fix example in docs

nikita-b:
A comma was missed in the payload example

### DIFF
--- a/website/content/api-docs/variables.mdx
+++ b/website/content/api-docs/variables.mdx
@@ -179,7 +179,7 @@ $ curl \
   "CreateIndex": 1457,
   "ModifyIndex": 1457,
   "CreateTime": 1662061225600373000,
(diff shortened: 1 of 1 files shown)


website/content/api-docs/variables.mdx
```diff
@@ -131,7 +131,7 @@ This endpoint creates or updates a variable.
 
 | Method | Path      | Produces           |
 |--------|-----------|--------------------|
```
jrasell, line 133:
```suggestion
| Method | Path               | Produces           |
|--------|--------------------|--------------------|
```

jrasell (changes requested):
HI @nikita-b and thanks for this. I have suggested a minor change to keep the table formatting as expected. Once this is added to the change set, I will get this merged.

jrasell (approved):
LGTM, thanks @nikita-b!

PR #15250, opened 2022-11-15
agent: ensure all HTTP Server methods are pointer receivers.

jrasell:
Per https://go.dev/doc/faq#methods_on_values_or_pointers

> If some of the methods of the type must have pointer receivers, the rest should too, so the method set is consistent regardless of how the type is used.


tgross (approved):
LGTM

PR #15252, opened 2022-11-15
api: ensure all request body decode error return a 400 status code.

jrasell:
Related to #15253 but split, so this PR can be back-ported further.


lgfa29 (approved):
> Bad request, no 🍩 for you

PR #15253, opened 2022-11-15
api: ensure ACL role upsert decode error returns a 400 status code.

jrasell:
Related to #15252 and forms the changes that will be back-ported into 1.4.x. This was built from the mentioned PR, and therefore that should be merged first.

It also includes a semgrep rule to catch this behaviour moving forward.

### DIFF
--- a/.semgrep/http_endpoint.yml
+++ b/.semgrep/http_endpoint.yml
@@ -1,0 +1,22 @@
+rules:
+  - id: "http-endpoint-request-decode-error-code"
+    patterns:
+      - pattern: |
+          if err := decodeBody(...); err != nil {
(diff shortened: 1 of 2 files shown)


lgfa29 (approved):
Oh nice catch. This may be, in theory, an API breaking change? Either way, having a changelog entry would be good.

jrasell:
> Oh nice catch. This may be, in theory, an API breaking change? Either way, having a changelog entry would be good.

Ah yeah I should have mentioned that the related PR has a changelog entry that will cover this.

lgfa29:
Oh right, you did mention that this part of another PR. Don't mind me 😅

Issue #15169, opened 2022-11-07
[UI] Add namespace to breadcrumb navigation


wdmeest:
### Proposal

Add namespace name to breadcrumb navigation (or another visible location)

![image](https://user-images.githubusercontent.com/23402402/200304700-dce81126-997c-49e6-a092-9afc7d1f86c0.png)

### Use-cases

We have different jobs with the same name but in a different namespace (example: job "intranet" we have in namespace "testing", "development" and "production"). You open the desired job in the specific namespace, do your stuff and leave the web page open. When returning to the web page (after a certain period of time) for example to restart the allocation, you want to be certain that you are restarting the allocation in the correct namespace, so not accidently restarting the production allocation.

### Attempted Solutions

\

tomiles:
Agreed, once you navigate to the task group level or deeper there is no way to tell what namespace the job is based on the UI, navigating back to the job level.

jrasell:
Hi @wdmeest and thanks for raising this idea that sounds really useful. I'll mark this so that our UI engineers and designers can take a further look.

ChaiWithJai:
Hey, @wdmeest! Thanks for raising this idea. We've started discussion internally regarding how to resolve this issue. We are doing to do our best to find ways to resolve this issue while maintaining the integrity of our design.

Please feel free to join the [conversation](https://github.com/hashicorp/nomad/pull/15261) or file any new tickets for challenges or even tiny papercuts in the UI. We'd love to make your experience with the Nomad UI better!

wdmeest:
@ChaiWithJai  Added a possible solution to the conversation: https://github.com/hashicorp/nomad/pull/15261#issuecomment-1319686548

PR #15261, opened 2022-11-15
refact: add namespace to nested detail pages

ChaiWithJai:
Closes [15169](https://github.com/hashicorp/nomad/issues/15169)

Add `namespace` to Allocation Details and Task Details on the respective overview pages.

image

image

image

### DIFF
--- a/ui/app/components/breadcrumbs/job.hbs
+++ b/ui/app/components/breadcrumbs/job.hbs
@@ -48,9 +48,6 @@
           </dt>
           <dd>
             {{this.job.trimmedName}}
-            {{#if this.job.namespace}}
-              <span>@{{this.job.namespace.name}}</span>
-            {{/if}}
           </dd>
         </dl>
       </LinkTo>
--- a/ui/app/components/breadcrumbs/job.hbs
+++ b/ui/app/components/breadcrumbs/job.hbs
@@ -49,6 +49,9 @@
           <dd>
             {{this.job.trimmedName}}
           </dd>
+            {{#if this.job.namespace}}
+              <dt>{{this.job.namespace.name}}</dt>
+            {{/if}}
         </dl>
       </LinkTo>
     </li>
--- a/ui/app/components/breadcrumbs/job.hbs
+++ b/ui/app/components/breadcrumbs/job.hbs
@@ -48,10 +48,10 @@
           </dt>
           <dd>
             {{this.job.trimmedName}}
-          </dd>
             {{#if this.job.namespace}}
-              <dt>{{this.job.namespace.name}}</dt>
+              <span>@{{this.job.namespace.name}}</span>
             {{/if}}
+          </dd>
         </dl>
       </LinkTo>
     </li>
--- a/ui/app/styles/core/breadcrumb.scss
+++ b/ui/app/styles/core/breadcrumb.scss
@@ -31,4 +31,9 @@
   dl dt {
     font-size: small;
   }
+
+  dl span {
+    display: block;
+    font-size: x-small;
+  }
 }
--- a/ui/app/styles/core/breadcrumb.scss
+++ b/ui/app/styles/core/breadcrumb.scss
@@ -31,9 +31,4 @@
   dl dt {
     font-size: small;
   }
-
-  dl span {
-    display: block;
-    font-size: x-small;
-  }
 }
--- a/ui/app/templates/allocations/allocation/index.hbs
+++ b/ui/app/templates/allocations/allocation/index.hbs
@@ -124,6 +124,14 @@
           </LinkTo>
         </Tooltip>
       </span>
+      <span class="pair">
+        <span class="term">
+          Namespace
+        </span>
+        <span>
+          {{this.model.job.namespace.name}}
+        </span>
+      </span>
     </div>
   </div>
   <div class="boxed-section">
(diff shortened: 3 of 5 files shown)


ui/app/components/breadcrumbs/job.hbs
```diff
@@ -49,6 +49,9 @@
           <dd>
             {{this.job.trimmedName}}
           </dd>
+            {{#if this.job.namespace}}
+              <dt>{{this.job.namespace.name}}</dt>
```
philrenaud, line 53: Semantically, we'd want to avoid using `<dt>` here as it implies something that needs to be defined. I think you'll want to use a less semantic-laden tag (span?) and apply some styles to it in the breadcrumb stylesheet.

ChaiWithJai: Ahhhh! But then `span` doesn't create a new line!! Is it OK if I do

```
<br>
<span>Namespace</span>
```

joshklekamp (dismissed):
I'll defer to @juliezzhou here, but I'm not in love with the formatting here. The misalignment feels like an error. Would it be better to display this in a single line and format it like the route/url does (i.e. "$job-ID@$namespace-ID")?

ChaiWithJai:
> I'll defer to @juliezzhou here, but I'm not in love with the formatting here. The misalignment feels like an error. Would it be better to display this in a single line and format it like the route/url does (i.e. "$job-ID@$namespace-ID")?

That's a great suggestion! I gave it a try just now... one challenge that I see with a single line is that we have to deal with a deeply nested view that goes 5-6 levels deep. In the event of a parameterized job, we'd have overflow. That's probably too much to work around right now considering we have a redesign coming around the corner. IMHO, the most swift solution that solves a user's problem and doesn't require too much work before the redesign, would be to just include the namespace as a smaller text below the name (and even consider just adding the `@` sign).

image

Per the ticket, it sounds like user's are looking for a quick fix:

> We have different jobs with the same name but in a different namespace (example: job "intranet" we have in namespace "testing", "development" and "production"). You open the desired job in the specific namespace, do your stuff and leave the web page open. When returning to the web page (after a certain period of time) for example to restart the allocation, you want to be certain that you are restarting the allocation in the correct namespace, so not accidently restarting the production allocation.

Would it be unreasonable to get a quick fix in now to resolve the issue, and then consider challenges we may face with balancing look & feel and maintaining a responsive design?

juliezzhou:
My concern is if it is not called "namespace-1", but "apps" or "default", will it still be clear to the users that it means the namespace?

ChaiWithJai:
> My concern is if it is not called "namespace-1", but "apps" or "default", will it still be clear to the users that it means the namespace?

@juliezzhou what would you propose a solution?

The simplest solution I could see would be pattern matching what we do in the URL:

image

Because it matches the pattern that we have in the URL, and the text is smaller that could be enough guidance for the user.

juliezzhou:
read through the comment from the customer, the other way we can do is to have a hover state on the job to show the namespace? I know it is a little hidden, but then it won't make the breadcrumb more crowded, and we don't know if all the customers will need to see the namespace?
Screen Shot 2022-11-15 at 3 37 21 PM

joshklekamp:
We could also just use `$job-ID@$namespace-ID` in the page header instead of the breadcrumb. That would reserve space for us in the breadcrumb while making it clear which namespace the job was in. 

CleanShot 2022-11-15 at 14 42 57@2x

juliezzhou:
Re: Josh's comments on the misalignment, maybe we can align the name like this
Screen Shot 2022-11-15 at 3 40 01 PM
But i still don't think that every user will need this, should we make it so prominent in the breadcrumb?

ChaiWithJai:
> Re: Josh's comments on the misalignment, maybe we can align the name like this Screen Shot 2022-11-15 at 3 40 01 PM But i still don't think that every user will need this, should we make it so prominent in the breadcrumb?

@juliezzhou @joshklekamp 

The community user filing the ticket mentioned the pain point was leaving a tab, coming back and wanting to take an action to "restart" (or take another action) on an allocation, but being unsure about the job.

I'm unsure if changing the title of the Job Detail page solves the problem, because the user will still need to click back into the page.

And re:  the tooltip, that seems like an elegant solution but it still seems like a paper cut in the user experience because I think they just want to visually see which job they're on (jobs are unique by namespace and name).

I think @juliezzhou's suggestion of smaller text with the `@` sign is going to be our best path forward.
image

joshklekamp:
Yeah, good call out on leaving the page @ChaiWithJai! While it doesn't have to be the breadcrumb, we do need something visually that is persistent across pages to contextualize the namespace. 

I generally would prefer we do design exploration first because we are trying solve a customer problem, not ship features our users directly ask for.  My preference given all of the conversation here is that we give Julie a little time to thing through this problem in Figma before doing any more development. If we end up deciding we want to do this via the breadcrumb, we haven't lost any work. If we decide on a different solution, we will have done right by our users.  

If we go with a breadcrumb, I think I would still prefer the text on one line and the same size purely from a design perspective. Visually, having three different type sizes there looks a little random.

wdmeest:
Possible solution is to add a new layer to the breadcrumb (to keep the 2 level look)?:

image

Clicking on the namespace gives you all the jobs for that namespace

joshklekamp:
Nice @wdmeest, that's a good suggestion too! And thanks for the original issue! 

We're doing a little exploration on our side and we'll definitely add this as an option to consider. We're going to do some design work to consider all possible options (both already considered and some not discussed yet) to figure out what might best solve this problem for you and other users.

philrenaud (approved):

Issue #11068, opened 2021-08-19
cli: add extra flag to specify timeout/waittime with the new -monitor on run/eval/deployment feature


42wim:
### Proposal

Add an option to increase the WaitTime in the deployment monitor: https://github.com/hashicorp/nomad/blob/main/command/deployment_status.go#L237

The current implementation in https://github.com/hashicorp/nomad/pull/10661 works fine for CLI / tty but for running in a CI (eg jenkins) this results in printing new data every 2 seconds because they don't support a erase line.

In this kind environment you would only need output when something actually changes, which works fine if we increase the WaitTime. State changes will still be shown immediately when a new index is reached.

### Use-cases

To decrease the output spamming when running in CI environments.

### Attempted Solutions

Works fine in our environment when WaitTime is set to 30 seconds

tgross:
> The current implementation in https://github.com/hashicorp/nomad/pull/10661 works fine for CLI / tty but for running in a CI (eg jenkins) this results in printing new data every 2 seconds because they don't support a erase line.

For environments like Jenkins we should be detecting that there's no tty and then simply not running the monitor. But I suspect Jenkins is doing something janky here, as I seem to recall some weird behaviors around tty's last I ran it (which is admittedly years ago now).

In any case, apologies for not responding to this feature request earlier @42wim. I'll mark it for roadmapping but it's also a fairly small intervention so I'll mark it as a good first issue for new contributors as well.

jdockerty:
@tgross Hopefully this one hasn't been forgotten about :eyes: I've taken a bit of a "work in progress" shot at it in a linked PR.

Any feedback/help is much appreciated for my first contribution to Nomad :blush:

tgross:
Thanks @jdockerty! I've marked the PR so that it gets triaged for review.

tgross:
This has been implemented in https://github.com/hashicorp/nomad/pull/15262, which we'll ship in the next regular version of Nomad (likely 1.4.4)

PR #15262, opened 2022-11-15
cli: `wait` flag for use with `deployment status -monitor`

jdockerty:
Addresses comments laid out in https://github.com/hashicorp/nomad/issues/11068

### DIFF
--- a/command/deployment_status.go
+++ b/command/deployment_status.go
@@ -91,7 +91,7 @@ func (c *DeploymentStatusCommand) Name() string { return "deployment status" }
 
 func (c *DeploymentStatusCommand) Run(args []string) int {
 	var json, verbose, monitor bool
-	var waitTime time.Duration
+	var wait time.Duration
 	var tmpl string
 
 	flags := c.Meta.FlagSet(c.Name(), FlagSetClient)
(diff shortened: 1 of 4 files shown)


command/deployment_status.go
```diff
@@ -48,6 +48,9 @@ Status Options:
   -monitor
     Enter monitor mode to poll for updates to the deployment status.
 
+  -wait-time
+    How long to wait before polling an update, used in conjunction with monitor mode. Defaults to 2s.
```
tgross, line 52: Two requests here: let's change the name to `-wait` as that matches what we use for similar configuration values elsewhere. And let's wrap this to 80col:

```suggestion
-wait
    How long to wait before polling an update, used in conjunction with monitor 
    mode. Defaults to 2s.
```

(And then we'll need to change the actual flag parsing to match.)

tgross (changes requested):
Looks great @jdockerty! 

To get this into mergeable state:
* Address the comments.
* Run `make cl` to add a changelog entry for the flag.
* Add the docs for the flag to the `deployment status` docs in [status.mdx](https://github.com/hashicorp/nomad/blob/main/website/content/docs/commands/deployment/status.mdx)
* Mark it ready for review!

I think at some point someone is going to want to come along and add this feature to `job run` too, but looking at the code for that the monitoring is a bit more complex because we monitor an eval and then switch to monitoring the deployment. What you've got here does the job asked and is worth shipping on its own. 😀

command/deployment_status.go
jdockerty: Makes sense :blush: 

I debated myself whether to go with `wait`, but I always seem to land on the more verbose options :smile: This has changed now :rocket:

tgross (approved):
LGTM! Thanks @jdockerty!

PR #15267, opened 2022-11-16
test: ensure leader is still valid in reelection test

tgross:
The `TestLeader_Reelection` test waits for a leader to be elected and then makes some other assertions. But it implcitly assumes that there's no failure of leadership before shutting down the leader, which can lead to a panic in the tests. Assert there's still a leader before the shutdown.


jrasell (approved):
LGTM, thanks!

PR #15273, opened 2022-11-16
Tokens page styles

### LINKED ISSUE #15269: Authentication page styles

philrenaud:
Resolves #15269 

- Renames a few computed properties to be more clear: `tokenIsValid` to `successfulSignIn` (because it only fires immediately upon auth request), `this.tokenRecord` to `this.shouldShowPolicies`, etc.
- Modify styles to match intended designs: stacked notifications -> tokens -> sso
image

### DIFF
--- a/ui/app/controllers/settings/tokens.js
+++ b/ui/app/controllers/settings/tokens.js
@@ -19,8 +19,11 @@ export default class Tokens extends Controller {
 
   @reads('token.secret') secret;
 
-  successfulSignIn = false;
-  unsuccessfulSignIn = false;
+  /**
+   * @type {(null | "success" | "failure")} signInStatus
+   */
+  @tracked
+  signInStatus = null;
 
   @alias('token.selfToken') tokenRecord;
 
@@ -34,10 +37,7 @@ export default class Tokens extends Controller {
       secret: undefined,
       tokenNotFound: false,
     });
-    this.setProperties({
-      successfulSignIn: false,
-      unsuccessfulSignIn: false,
-    });
+    this.signInStatus = null;
     // Clear out all data to ensure only data the anonymous token is privileged to see is shown
     this.resetStore();
     this.token.reset();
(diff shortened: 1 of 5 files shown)


ui/app/controllers/settings/tokens.js
```diff
@@ -14,14 +14,13 @@ export default class Tokens extends Controller {
...
-  @service flashMessages;
 
   queryParams = ['code', 'state'];
 
   @reads('token.secret') secret;
 
-  tokenIsValid = false;
-  tokenIsInvalid = false;
+  successfulSignIn = false;
+  unsuccessfulSignIn = false;
```
ChaiWithJai, line 23: `question`: I'm noticing that these properties are wired together. And complimentary opposites. Could it be possible for this to be combined into a field that is an enum? Potentially:

`signInStatus: 'success' | 'failed' | 'idle'`

Also should this be a tracked property that updates reactively?

ui/app/controllers/settings/tokens.js
```diff
@@ -52,9 +51,11 @@ export default class Tokens extends Controller {
   @action
   verifyToken() {
     const { secret } = this;
+    this.clearTokenProperties();
```
ChaiWithJai, line 54: `question`: This is action dispatches a few asynchronous actions via `ember-concurrency` and `ember-data`. But, I'm noticing that there is no `catch` for any failures or waiting until these actions have completed to set the secret on the token. How should we think about this? Is this a non-issue because some JavaScript magic?

ui/app/controllers/settings/tokens.js
```diff
@@ -52,9 +51,11 @@ export default class Tokens extends Controller {
...
     const TokenAdapter = getOwner(this).lookup('adapter:token');
 
     this.set('token.secret', secret);
+    this.set('secret', null);
```
ChaiWithJai, line 58: `thought`: This is interesting because when we dispatch `clearTokenProperties` we're setting the `token` to be `undefined` and then resetting the `secret` to `null` here. How are we thinking about the difference between `null` and `undefined` here?

ChaiWithJai (approved):

ui/app/controllers/settings/tokens.js
philrenaud: Yeah, this is a good point. There's not much stopping us from updating this, so I think I will modernize it. Thanks for the suggestion!

ChaiWithJai (approved):
LGTM

Issue #15019, opened 2022-10-24
deprecate raft log entry AllocUpdateRequestType


tgross:
While looking into metrics emitted by the raft FSM, I noticed that the `AllocUpdateRequestType` raft entry has no non-test callers other than the event stream that reads those non-existent raft entries. This entry appears to have been deprecated by the new-in-0.6.0 Deployments in 64d7fd8330c3f4d4f8bbc2bac2826f5a9e10d197

I took a quick look at removing the metrics, events, and docs (ex references to `nomad.nomad.fsm.alloc_update` in the metrics guide), but there's a bunch of tests I'll need to clean up as well. I won't get a chance to get to that in the next day or two, so leaving this issue for myself to remind me to finish cleaning this up.

PR #15285, opened 2022-11-17
remove deprecated `AllocUpdateRequestType` raft entry

tgross:
Fixes https://github.com/hashicorp/nomad/issues/15019

After Deployments were added in Nomad 0.6.0, the `AllocUpdateRequestType` raft log entry was no longer in use. Mark this as deprecated, remove the associated dead code, and remove references to the metrics it emits from the docs. We'll leave the entry itself just in case we encounter old raft logs that we need to be able to safely load.

---

I don't think this needs a changelog entry. All the real consumers of the log entry were removed in Nomad 0.6.0, so this is code cleanup without a behavior change. ~Not sure if we should backport to 1.3.x and 1.2.x as well?~ Eh, for abundance of caution I'll only do 1.4.x


jrasell (commented):
LGTM!

tgross:
Ok so weirdly the `// DEPRECATED` comment in structs.go caused a change in the generated files, which was the exact opposite of what I wanted. So I've removed that comment and now the checks for the generate files should pass.

shoenig (approved):
LGTM

PR #15286, opened 2022-11-17
e2e: disable systemd stub dns in jammy image

shoenig:
make sure systemd stub resolver does not try to start and steal port 53

### DIFF
--- a/e2e/terraform/packer/ubuntu-jammy-amd64/setup.sh
+++ b/e2e/terraform/packer/ubuntu-jammy-amd64/setup.sh
@@ -119,10 +119,14 @@ fi
 
 echo "Configuring dnsmasq"
 
+# disable systemd stub resolver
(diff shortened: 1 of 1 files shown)


e2e/terraform/packer/ubuntu-jammy-amd64/setup.sh
```diff
@@ -119,10 +119,14 @@ fi
 
 echo "Configuring dnsmasq"
 
+# disable systemd stub resolver
+sudo sed -i 's|#DNSStubListener=yes|DNSStubListener=no|g' /etc/systemd/resolved.conf
```
tgross, line 123: This is maybe nitpicky, but I'm not sure I trust the systemd project not to randomly decide that the commented-out lines should have an extra space and for that to get picked up in a seemingly innocent patch downstream. Maybe we should change this to append the `DNSStubListener=no` line?

tgross (approved):
LGTM!

PR #15294, opened 2022-11-17
make eval cancelation async with `Eval.Ack`

tgross:
In #14621 we added an eval canelation reaper goroutine with a channel that allowed us to wake it up. But we forgot to actually send on this channel from `Eval.Ack` and are still committing the cancelations synchronously. Fix this by sending on the buffered channel to wake up the reaper instead.

(ref https://github.com/hashicorp/nomad/pull/14621/files#r1025630828)

### DIFF
--- a/nomad/eval_endpoint.go
+++ b/nomad/eval_endpoint.go
@@ -234,10 +234,9 @@ func (e *Eval) Ack(args *structs.EvalAckRequest,
 		return err
 	}
 
-	// It's not necessary to cancel evals before Ack returns, but it's done here
-	// to commit canceled evals as close to the Ack'd eval being committed as
-	// possible.
-	return cancelCancelableEvals(e.srv)
+	// Wake up the eval cancelation reaper
+	e.srv.reapCancelableEvalsCh <- struct{}{}
+	return nil
 }
 
 // Nack is used to negative acknowledge completion of a dequeued evaluation.


shoenig (approved):
LGTM!

nomad/eval_endpoint.go
```diff
@@ -234,10 +234,9 @@ func (e *Eval) Ack(args *structs.EvalAckRequest,
 		return err
 	}
 
-	// It's not necessary to cancel evals before Ack returns, but it's done here
-	// to commit canceled evals as close to the Ack'd eval being committed as
-	// possible.
-	return cancelCancelableEvals(e.srv)
+	// Wake up the eval cancelation reaper
+	e.srv.reapCancelableEvalsCh <- struct{}{}
```
schmichael, line 238: This could still block if a raft apply is already in progress. I think to asynchronously trigger another loop you want to use a buffered chan:

```suggestion
	// Wake up the eval cancelation reaper
	select {
		case e.srv.reapCancelableEvalsCh <- struct{}{}:
		default:
	}
```

and make the chan buffered of size 1: `make(struct{}, 1)`

This ensures `Eval.Ack` never blocks but that the cancellation loop always runs at least once after an Eval.Ack (either due to the Ack causing a send, or because a prior Ack's sent is already buffered and waiting to be recv'd).

schmichael: Your implementation isn't wrong though, just not always asynchronous.

tgross: The channel is already buffered, with the thinking that we wanted to wait until the buffer was empty before continuing. But you're right we could avoid blocking entirely in that case b/c we already know the buffer is full. I merged right before your comment unfortunately, but I can fix that up.

tgross: Take 2 (3?) is here: https://github.com/hashicorp/nomad/pull/15298

PR #15309, opened 2022-11-18
client: accommodate Consul 1.14.0 gRPC and agent self changes.

### LINKED ISSUE #15266: Consul grpc_tls usage cannot deploy nomad jobs anymore
### Nomad version
1.4.2
### Consul version
1.14.0

I updated my consul agent today. Consul 1.40 have a breaking change 

BREAKING CHANGES:
config: Add new ports.grpc_tls configuration option.

I remove the grpc falg form the config and replace it with grpc_tls . After that nomad is not able to deploy jobs with the error:

Placement Failure
Task Group "keycloak":
  * Constraint "${attr.consul.grpc} > 0": 1 nodes excluded by filter

jrasell:
Consul 1.14.0 changed the way in which gRPC listeners are configured, particularly when using TLS. Prior to the change, a single listener was responsible for handling plain-text and encrypted gRPC requests. In 1.14.0 and beyond, separate listeners will be used for each, defaulting to 8502 and 8503 for plain-text and TLS respectively.

The change means that Nomad’s Consul Connect integration would not work when integrated with Consul clusters using TLS and running 1.14.0 or greater.

The Nomad Consul fingerprinter identifies the gRPC port Consul has exposed using the "DebugConfig.GRPCPort" value from Consul’s “/v1/agent/self” endpoint. In Consul 1.14.0 and greater, this only represents the plain-text gRPC port which is likely to be disbaled in clusters running TLS. In order to fix this issue, Nomad now takes into account the Consul version and configured scheme to optionally use “DebugConfig.GRPCTLSPort” value from Consul’s agent self return.

The “consul_grcp_socket” allocrunner hook has also been updated so that the fingerprinted gRPC port attribute is passed in. This provides a better fallback method, when the operator does not configure the “consul.grpc_address” option.

Closes #15266

### DIFF
--- a/client/fingerprint/consul.go
+++ b/client/fingerprint/consul.go
@@ -3,6 +3,7 @@ package fingerprint
 import (
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	consulapi "github.com/hashicorp/consul/api"
@@ -207,8 +208,8 @@ func (f *ConsulFingerprint) grpc(scheme string) func(info agentconsul.Self) (str
 		// Now that we know we are querying a Consul agent running v1.14.0 or
 		// greater, we need to select the correct port parameter from the
 		// config depending on whether we have been asked to speak TLS or not.
-		switch scheme {
-		case "https", "HTTPS":
+		switch strings.ToLower(scheme) {
+		case "https":
 			return f.grpcTLSPort(info)
 		default:
 			return f.grpcPort(info)
--- a/client/fingerprint/consul.go
+++ b/client/fingerprint/consul.go
@@ -7,12 +7,22 @@ import (
 
 	consulapi "github.com/hashicorp/consul/api"
 	log "github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-version"
 	agentconsul "github.com/hashicorp/nomad/command/agent/consul"
+	"golang.org/x/exp/slices"
 )
 
 const (
 	consulAvailable   = "available"
 	consulUnavailable = "unavailable"
+)
+
+var (
+	// consulGRPCPortChangeVersion is the Consul version which made a breaking
+	// change to the way gRPC API listeners are created. This means Nomad must
+	// perform different fingerprinting depending on which version of Consul it
+	// is communicating with.
+	consulGRPCPortChangeVersion = version.Must(version.NewVersion("1.14.0"))
 )
 
 // ConsulFingerprint is used to fingerprint for Consul
@@ -97,7 +107,7 @@ func (f *ConsulFingerprint) initialize(req *FingerprintRequest) error {
 			"consul.datacenter":    f.dc,
 			"consul.segment":       f.segment,
 			"consul.connect":       f.connect,
-			"consul.grpc":          f.grpc,
+			"consul.grpc":          f.grpc(consulConfig.Scheme),
 			"consul.ft.namespaces": f.namespaces,
 		}
 	}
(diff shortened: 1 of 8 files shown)


client/fingerprint/consul.go
```diff
@@ -173,11 +183,54 @@ func (f *ConsulFingerprint) connect(info agentconsul.Self) (string, bool) {
...
+		}
+
+		consulVersion, err := version.NewVersion(v)
+		if err != nil {
+			return "", false
+		}
+
+		// This check ensures that users running a 1.14.0 pre-release get the
+		// desired behaviour, otherwise 1.14.0-beta1 < 1.14.0 will be truthy.
+		versionsMatch := slices.Equal(consulGRPCPortChangeVersion.Segments(), consulVersion.Segments())
```
tgross, line 204: We probably don't use it everywhere we should be but just FYI [`Version.Core`](https://pkg.go.dev/github.com/hashicorp/go-version?utm_source=godoc#Version.Core) lets you extract a version without `-beta1`, etc.

client/fingerprint/consul_test.go
```diff
@@ -296,29 +296,127 @@ func TestConsulFingerprint_grpc(t *testing.T) {
...
+	})
+
+	t.Run("grpc disabled pre-1.14 https", func(t *testing.T) {
+		s, ok := fp.grpc("https")(agentconsul.Self{
+			"Config":      {"Version": "1.14.0"},
+			"DebugConfig": {"GRPCTLSPort": -1.0}, // JSON numbers are floats
+		})
+		require.True(t, ok)
+		require.Equal(t, "-1", s)
+	})
```
tgross, line 369: This test is repeated below on line 398. But it also looks weird to me: it says "pre-1.14" but the version is 1.14.0 and I wouldn't expect a pre-1.14 to have the `GRPCTLSPort` field at all? Maybe just the name is wrong?

client/fingerprint/consul.go
jrasell: Thanks that is a lot cleaner. The update can be seen within commit: https://github.com/hashicorp/nomad/pull/15309/commits/cebd6b59ed3afbc521678cb3441be1c2ce5b64d2

client/fingerprint/consul_test.go
jrasell: I think I had a mare with copy/pasting the test outlines; this has been cleaned up within commit: https://github.com/hashicorp/nomad/pull/15309/commits/cebd6b59ed3afbc521678cb3441be1c2ce5b64d2

tgross (approved):
LGTM

client/fingerprint/consul.go
```diff
@@ -173,11 +182,50 @@ func (f *ConsulFingerprint) connect(info agentconsul.Self) (string, bool) {
...
+		// than 1.14.0 we use the original single gRPC port.
+		if consulVersion.Core().LessThan(consulGRPCPortChangeVersion.Core()) {
+			return f.grpcPort(info)
+		}
+
+		// Now that we know we are querying a Consul agent running v1.14.0 or
+		// greater, we need to select the correct port parameter from the
+		// config depending on whether we have been asked to speak TLS or not.
+		switch scheme {
+		case "https", "HTTPS":
```
schmichael, line 211: Not a blocker, but I think lowering (or raising) is a better approach generally:

```suggestion
		switch strings.ToLower(scheme) {
		case "https":
```

schmichael (approved):
Thanks for jumping on this @jrasell! LGTM

client/fingerprint/consul.go
jrasell: Agreed; I was trying to be clever and avoid the func overhead but I can now see what can go wrong with that approach. This has been fixed in commit: https://github.com/hashicorp/nomad/pull/15309/commits/497eb98b161768cb8ccc54e9a3f577d90908470e

shoenig (approved):
Thanks @jrasell !

Issue #15265, opened 2022-11-16
After upgrading Nomad to 1.4.2, services perpetually register in consul (/PUT to local consul agent)


usovamaria:
### Nomad version and environment details
Nomad `1.1.2` -> `1.4.2`
Consul `1.9.5`
Ubuntu `20.04`

Environment configuration:
Servers: 3 nomad servers, consul servers (services register here) + consul servers with ACL.
Clients: we've installed nomad client and consul agent.

### Issue
After upgrading Nomad from 1.1.2 to 1.4.2 we saw that the rate of services registration in consul (using `/PUT` to consul agent) significantly increased according to metrics and logs.

We use metric `rate(consul_http_PUT_v1_agent_service_register_count)` in order to monitor consul agent API usage. Here is the graph based on this metric: 

<img width="1072" alt="Screen Shot 2022-11-16 at 13 25 55" src="https://user-images.githubusercontent.com/62989519/202169438-968f6614-c9bb-4638-9535-b1b99574a7ef.png">

Also we've noticed that consul logs amount related to registering service increased too, e.g.:

`Nov 16 13:29:18 hostname consul[48446]:     2022-11-16T13:29:18.585+0300 [DEBUG] agent.http: Request finished: method=PUT url=/v1/agent/service/register from=[::1]:23594 latency=7.206663ms`.  

Log amount obviously strongly correlates to the number of service containers on a host.

In contrast, our production environment, where nomad was not upgraded, the `/PUT` rate to consul agent is normal (0-0.1 per minute).

We do not experience performance issues on a current environment (with ~3K running containers) but we are going to update a production environment (with ~6K) and we can't predict an impact on consul and nomad performance. The only thing we noticed is that the number of goroutines on consul servers increased (~30K -> ~80K).

We assume that Nomad and Consul interaction during the process of service registration has changed.

### Reproduction steps

1. Bootstrap an environment with Nomad `1.1.2`, Consul `1.9.5`.
2. Setup up monitoring of `/PUT` rate to a local consul agent (`consul_http_PUT_v1_agent_service_register_count`) and log rate in DEBUG mode.
3. Run containers via Nomad.
4. Update Nomad servers and clients to `1.4.2` according to official documentation.
5. See the change in monitoring.

#### Expected Behaviour

Register rate of services remains the same or increases insignificantly after Nomad update

shoenig:
Thanks for reporting, @usovamaria 

[...]

PR #14917, opened 2022-10-17
consul: do not re-register already registered services

shoenig:
This PR updates Nomad's Consul service client to do map comparisons
using `maps.Equal` instead of `reflect.DeepEqual`. The bug fix is in how
`DeepEqual` treats nil slices different from empty slices, when actually
they should be treated the same.

Also, we need to avoid comparing `TaggedAddress` on agent registrations if
none are set on the Nomad side - because Consul will helpfully set that field
for us on its side.

And some drive-by cleanup `map[string]struct{}` for `set.Set`

Fixes #14914

Issue #14914, opened 2022-10-17
consul: all services being re-registered all the time


shoenig:
While looking into https://github.com/hashicorp/nomad/issues/3935, it looks like we are improperly re-registering every service on every iteration of the reconciliation loop. This is quite inefficient, we should only re-register a service if the service definition has changed on the Consul side from what the Nomad job definition declares.

E.g.
```
GET_SERVICES namespace=default why=sync
got services keys=[_nomad-server-o5s5vxlab3cgpyi3x2i5cm24xa2r2hbz _nomad-server-tl5dgauqxnxsq7pgcvloyq7d6gf75mj7 _nomad-client-rxajoy6bie7cqxttksivmlv2ucz7qxfl _nomad-server-icsfum2knxzafc7jtsv5hmelx35hkufm]
different H
SERVICE_REGISTER id=_nomad-server-o5s5vxlab3cgpyi3x2i5cm24xa2r2hbz name=nomad why=missing exists=true
different H
SERVICE_REGISTER id=_nomad-server-icsfum2knxzafc7jtsv5hmelx35hkufm name=nomad why=missing exists=true
different H
SERVICE_REGISTER id=_nomad-client-rxajoy6bie7cqxttksivmlv2ucz7qxfl name=nomad-client why=missing exists=true
different H
SERVICE_REGISTER id=_nomad-server-tl5dgauqxnxsq7pgcvloyq7d6gf75mj7 name=nomad why=missing exists=true
GET_CHECKS namespace=default
GET_SERVICES namespace=default why=sync
got services keys=[_nomad-server-tl5dgauqxnxsq7pgcvloyq7d6gf75mj7 _nomad-client-rxajoy6bie7cqxttksivmlv2ucz7qxfl _nomad-server-icsfum2knxzafc7jtsv5hmelx35hkufm _nomad-server-o5s5vxlab3cgpyi3x2i5cm24xa2r2hbz]
different H
SERVICE_REGISTER id=_nomad-server-tl5dgauqxnxsq7pgcvloyq7d6gf75mj7 name=nomad why=missing exists=true
different H
SERVICE_REGISTER id=_nomad-server-o5s5vxlab3cgpyi3x2i5cm24xa2r2hbz name=nomad why=missing exists=true
different H
SERVICE_REGISTER id=_nomad-server-icsfum2knxzafc7jtsv5hmelx35hkufm name=nomad why=missing exists=true
different H
[...]

PR #15311, opened 2022-11-18
consul: add trace logging around service registrations

shoenig:
This PR adds trace logging around the differential done between a Nomad service registration and its corresponding Consul service registration, in an effort to shed light on why a service registration request is being made.

Should help with https://github.com/hashicorp/nomad/issues/15265


jrasell (approved):
LGTM; just needs the required backport labels to be added.

tgross:
I've added the 1.4.x backport label but I'm looking at https://github.com/hashicorp/nomad/pull/14917 and I _think_ we probably need to backport to 1.3.x and 1.2.x as well?

shoenig:
1.4 introduced the now-suspect changes of https://github.com/hashicorp/nomad/issues/14914; I think that's as far back as we need to backport. If it turns out the actual for https://github.com/hashicorp/nomad/issues/15265 predates 1.4 we can backport that

PR #15322, opened 2022-11-18
acl: sso auth methods cli commands

pkazmierczak:
This PR implements CLI commands to interact with SSO auth methods.

This PR is part of the SSO work captured under ☂️ ticket https://github.com/hashicorp/nomad/issues/13120.

Consult NMD-155 RFC for documentation details.

### DIFF
--- a/command/acl_auth_method.go
+++ b/command/acl_auth_method.go
@@ -1,8 +1,10 @@
 package command
 
 import (
+	"fmt"
 	"strings"
 
+	"github.com/hashicorp/nomad/api"
 	"github.com/mitchellh/cli"
 )
 
@@ -54,3 +56,40 @@ func (a *ACLAuthMethodCommand) Name() string { return "acl auth-method" }
 
 // Run satisfies the cli.Command Run function.
 func (a *ACLAuthMethodCommand) Run(_ []string) int { return cli.RunResultHelp }
+
+// formatAuthMethod formats and converts the ACL auth method API object into a
+// string KV representation suitable for console output.
+func formatAuthMethod(authMethod *api.ACLAuthMethod) string {
+	return formatKV([]string{
+		fmt.Sprintf("Name|%s", authMethod.Name),
+		fmt.Sprintf("Type|%s", authMethod.Type),
+		fmt.Sprintf("Locality|%s", authMethod.TokenLocality),
+		fmt.Sprintf("MaxTokenTTL|%s", authMethod.MaxTokenTTL.String()),
+		fmt.Sprintf("Default|%t", authMethod.Default),
+		fmt.Sprintf("Config|%s", formatAuthMethodConfig(authMethod.Config)),
+		fmt.Sprintf("Create Index|%d", authMethod.CreateIndex),
+		fmt.Sprintf("Modify Index|%d", authMethod.ModifyIndex),
+	})
+}
+
+func formatAuthMethodConfig(config *api.ACLAuthMethodConfig) string {
+	return formatKV([]string{
+		fmt.Sprintf("OIDC Discovery URL|%s", config.OIDCDiscoveryURL),
+		fmt.Sprintf("OIDC Client ID|%s", config.OIDCClientID),
+		fmt.Sprintf("OIDC Client Secret|%s", config.OIDCClientSecret),
+		fmt.Sprintf("Bound audiences|%s", strings.Join(config.BoundAudiences, ",")),
+		fmt.Sprintf("Allowed redirects URIs|%s", strings.Join(config.AllowedRedirectURIs, ",")),
+		fmt.Sprintf("Discovery CA pem|%s", strings.Join(config.DiscoveryCaPem, ",")),
+		fmt.Sprintf("Signing algorithms|%s", strings.Join(config.SigningAlgs, ",")),
+		fmt.Sprintf("Claim mappings|%s", formatMap(config.ClaimMappings)),
+		fmt.Sprintf("List claim mappings|%s", formatMap(config.ListClaimMappings)),
+	})
+}
+
+func formatMap(m map[string]string) string {
+	out := []string{}
+	for k, v := range m {
+		out = append(out, fmt.Sprintf("%s/%s", k, v))
+	}
+	return formatKV(out)
+}
--- a/command/acl_auth_method.go
+++ b/command/acl_auth_method.go
@@ -20,6 +20,28 @@ func (a *ACLAuthMethodCommand) Help() string {
 Usage: nomad acl auth-method <subcommand> [options] [args]
 
   This command groups subcommands for interacting with ACL auth methods. 
+
+  Create an ACL auth method:
+
+      $ nomad acl auth-method create -name="name" -policy-name="policy-name"
+
+  List all ACL auth methods:
+
+      $ nomad acl auth-method list
+
+  Lookup a specific ACL auth method:
+
+      $ nomad acl auth-method info <acl_auth_method_name>
+
+  Update an ACL auth method:
+
+      $ nomad acl auth-method update -name="updated-name" <acl_auth_method_name>
+
+  Delete an ACL auth method:
+
+      $ nomad acl auth-method delete <acl_auth_method_name>
+
+  Please see the individual subcommand help for detailed usage information.
 `
 	return strings.TrimSpace(helpText)
 }
--- a/command/acl_auth_method.go
+++ b/command/acl_auth_method.go
@@ -60,20 +60,28 @@ func (a *ACLAuthMethodCommand) Run(_ []string) int { return cli.RunResultHelp }
 // formatAuthMethod formats and converts the ACL auth method API object into a
 // string KV representation suitable for console output.
 func formatAuthMethod(authMethod *api.ACLAuthMethod) string {
-	return formatKV([]string{
+	out := []string{
 		fmt.Sprintf("Name|%s", authMethod.Name),
 		fmt.Sprintf("Type|%s", authMethod.Type),
 		fmt.Sprintf("Locality|%s", authMethod.TokenLocality),
 		fmt.Sprintf("MaxTokenTTL|%s", authMethod.MaxTokenTTL.String()),
 		fmt.Sprintf("Default|%t", authMethod.Default),
-		fmt.Sprintf("Config|%s", formatAuthMethodConfig(authMethod.Config)),
-		fmt.Sprintf("Create Index|%d", authMethod.CreateIndex),
-		fmt.Sprintf("Modify Index|%d", authMethod.ModifyIndex),
-	})
+	}
+
+	if authMethod.Config != nil {
+		out = append(out, formatAuthMethodConfig(authMethod.Config)...)
+	}
+	out = append(out,
+		[]string{fmt.Sprintf("Create Index|%d", authMethod.CreateIndex),
+			fmt.Sprintf("Modify Index|%d", authMethod.ModifyIndex),
+		}...,
+	)
+
+	return formatKV(out)
 }
 
-func formatAuthMethodConfig(config *api.ACLAuthMethodConfig) string {
-	return formatKV([]string{
+func formatAuthMethodConfig(config *api.ACLAuthMethodConfig) []string {
+	return []string{
 		fmt.Sprintf("OIDC Discovery URL|%s", config.OIDCDiscoveryURL),
 		fmt.Sprintf("OIDC Client ID|%s", config.OIDCClientID),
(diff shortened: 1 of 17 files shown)


command/acl_auth_method_create.go
```diff
@@ -0,0 +1,174 @@
...
+// Run satisfies the cli.Command Run function.
+func (a *ACLAuthMethodCreateCommand) Run(args []string) int {
+
+	flags := a.Meta.FlagSet(a.Name(), FlagSetClient)
+	flags.Usage = func() { a.Ui.Output(a.Help()) }
+	flags.StringVar(&a.name, "name", "", "")
+	flags.StringVar(&a.methodType, "type", "", "")
+	flags.StringVar(&a.tokenLocality, "token-locality", "local", "")
+	flags.DurationVar(&a.maxTokenTTL, "max-token-ttl", 0, "")
+	flags.BoolVar(&a.isDefault, "default", true, "")
```
jrasell, line 100: I wonder whether true is the right default here? I don't have the answer, just the curiosity.

command/acl_auth_method_create.go
```diff
@@ -0,0 +1,174 @@
...
+	// Perform some basic validation
+	if a.name == "" {
+		a.Ui.Error("ACL auth method name must be specified using the -name flag")
+		return 1
+	}
+	if !slices.Contains([]string{"global", "local"}, a.tokenLocality) {
+		a.Ui.Error("Token locality must be set to either 'local' or 'global'")
+		return 1
+	}
+	if strings.ToLower(a.methodType) != "oidc" {
```
jrasell, line 122: In the RPC we check for `OIDC` within validation, so I wonder here if it would be worth using `ToUpper` and writing that value back on the config? I am thinking of cases where one or more of the letters is lower case, which would pass validation here but fail within the RPC.

command/acl_auth_method_create.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		a.Ui.Error(fmt.Sprintf("Error creating ACL auth method: %v", err))
+		return 1
+	}
+
+	a.Ui.Output(fmt.Sprintf("Created ACL auth method %s", a.name))
+	return 0
+}
+
+func configStringToAuthConfig(conf string) (*api.ACLAuthMethodConfig, error) {
```
jrasell, line 165: If we have time before release or as a followup, it would be nice to support loading files and even reading from stdin also.

command/acl_auth_method_info.go
```diff
@@ -0,0 +1,125 @@
...
+	client, err := a.Meta.Client()
+	if err != nil {
+		a.Ui.Error(fmt.Sprintf("Error initializing client: %s", err))
+		return 1
+	}
+
+	var (
+		method *api.ACLAuthMethod
+		apiErr error
+	)
```
jrasell, line 98: I think we can remove this and create and assign the variables below as there is no switching.

command/acl_auth_method_update.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/hashicorp/nomad/api"
+	"github.com/mitchellh/cli"
+	"github.com/posener/complete"
+	"golang.org/x/exp/slices"
+)
+
+// Ensure ACLAuthMethodUpdateCommand satisfies the cli.Command interface.
+var _ cli.Command = &ACLAuthMethodUpdateCommand{}
+
+// ACLRoleUpdateCommand implements cli.Command.
```
jrasell, line 18:
```suggestion
// ACLAuthMethodUpdateCommand implements cli.Command.
```

command/acl_auth_method_create.go
```diff
@@ -0,0 +1,174 @@
...
+    Auth method configuration in JSON format.
+`
+	return strings.TrimSpace(helpText)
+}
+
+func (a *ACLAuthMethodCreateCommand) AutocompleteFlags() complete.Flags {
+	return mergeAutocompleteFlags(a.Meta.AutocompleteFlags(FlagSetClient),
+		complete.Flags{
+			"-name":           complete.PredictAnything,
+			"-type":           complete.PredictAnything,
```
jrasell, line 73: We could use `complete.PredictSet("OIDC")` here as we will always be looking for known values.

command/acl_auth_method_create.go
```diff
@@ -0,0 +1,174 @@
...
+			"-max-token-ttl":  complete.PredictAnything,
+			"-token-locality": complete.PredictAnything,
```
jrasell, line 75: We could use complete.PredictSet("local", "global") here as we will always be looking for known values.

command/acl_auth_method_update.go
```diff
@@ -0,0 +1,191 @@
...
+    Updates auth method configuration (in JSON format).
+`
+
+	return strings.TrimSpace(helpText)
+}
+
+func (a *ACLAuthMethodUpdateCommand) AutocompleteFlags() complete.Flags {
+	return mergeAutocompleteFlags(a.Meta.AutocompleteFlags(FlagSetClient),
+		complete.Flags{
+			"-type":           complete.PredictAnything,
```
jrasell, line 68: We could use complete.PredictSet("OIDC") here as we will always be looking for known values.

command/acl_auth_method_update.go
```diff
@@ -0,0 +1,191 @@
...
+			"-max-token-ttl":  complete.PredictAnything,
+			"-token-locality": complete.PredictAnything,
```
jrasell, line 70: We could use complete.PredictSet("local", "global") here as we will always be looking for known values.

command/acl_auth_method_update.go
```diff
@@ -0,0 +1,191 @@
...
+			"-default":        complete.PredictAnything,
```
jrasell, line 71: We could use complete.PredictSet("true", "false") here as we will always be looking for known values.

command/acl_auth_method_create.go
```diff
@@ -0,0 +1,174 @@
...
+			"-default":        complete.PredictAnything,
```
jrasell, line 76: We could use complete.PredictSet("true", "false") here as we will always be looking for known values.

command/acl_auth_method_create.go
```diff
@@ -0,0 +1,174 @@
...
+
+// Run satisfies the cli.Command Run function.
+func (a *ACLAuthMethodCreateCommand) Run(args []string) int {
+
+	flags := a.Meta.FlagSet(a.Name(), FlagSetClient)
+	flags.Usage = func() { a.Ui.Output(a.Help()) }
+	flags.StringVar(&a.name, "name", "", "")
+	flags.StringVar(&a.methodType, "type", "", "")
+	flags.StringVar(&a.tokenLocality, "token-locality", "local", "")
+	flags.DurationVar(&a.maxTokenTTL, "max-token-ttl", 0, "")
```
jrasell, line 99: This flag doesn't get validated unlike the others, but if unset will fail validation on the RPC. Could we add validation for this also?

command/acl_auth_method_create.go
```diff
@@ -0,0 +1,174 @@
...
+func (a *ACLAuthMethodCreateCommand) Name() string { return "acl auth-method create" }
+
+// Run satisfies the cli.Command Run function.
+func (a *ACLAuthMethodCreateCommand) Run(args []string) int {
+
+	flags := a.Meta.FlagSet(a.Name(), FlagSetClient)
+	flags.Usage = func() { a.Ui.Output(a.Help()) }
+	flags.StringVar(&a.name, "name", "", "")
+	flags.StringVar(&a.methodType, "type", "", "")
+	flags.StringVar(&a.tokenLocality, "token-locality", "local", "")
```
jrasell, line 98: This default create an inconsistent UX between the API and CLI. I wonder if we want to move this defaulting into the RPC layer via a canonicalize function and default to empty here, allowing that through the CLI validation? The RPC part can be done as a follow up.

jrasell (changes requested):
Looking great! Some comments about the UX and how this ties into the HTTP API.

I noted about being able to read the config additionally from a file or stdin, which I would like to see before merge. I find writing out formatted JSON strings on the CLI a little annoying and error prone.

command/acl_role_delete.go
```diff
@@ -42,7 +42,7 @@ func (a *ACLRoleDeleteCommand) AutocompleteArgs() complete.Predictor { return co
 func (a *ACLRoleDeleteCommand) Synopsis() string { return "Delete an existing ACL role" }
 
 // Name returns the name of this command.
-func (a *ACLRoleDeleteCommand) Name() string { return "acl token delete" }
+func (a *ACLRoleDeleteCommand) Name() string { return "acl role delete" }
```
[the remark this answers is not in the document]
pkazmierczak: Yeah I'll make another PR for this.

nomad/structs/acl.go
```diff
@@ -770,6 +770,7 @@ type ACLAuthMethodUpsertRequest struct {
 // ACLAuthMethodUpsertResponse is a response of the upsert ACL auth methods
 // operation
 type ACLAuthMethodUpsertResponse struct {
+	AuthMethods []*ACLAuthMethod
```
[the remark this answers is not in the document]
pkazmierczak: good points, I'll move this to a separate PR.

command/acl_auth_method.go
```diff
@@ -0,0 +1,103 @@
...
+// Help satisfies the cli.Command Help function.
+func (a *ACLAuthMethodCommand) Help() string {
+	helpText := `
+Usage: nomad acl auth-method <subcommand> [options] [args]
+
+  This command groups subcommands for interacting with ACL auth methods. 
+
+  Create an ACL auth method:
+
+      $ nomad acl auth-method create -name="name" -policy-name="policy-name"
```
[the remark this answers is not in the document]
pkazmierczak: yikes, copy pasta bug. nice catch.

command/acl_auth_method_create.go
pkazmierczak: This is a general issue with default values for non-string flags. I don't know either, but I think we'd want at least one auth method to be the default, wouldn't we? Same problem for the TTL: I set it to 0 by default, but the validation function won't accept anything below min TTL set in the server config. 

Is there a clever way not to set default values here?

command/acl_auth_method_create.go
pkazmierczak: Fair enough, we should be consistent.

command/acl_auth_method_info.go
pkazmierczak: yes sorry 🤦 this looks like another one of my classic copy pasta fails.

command/acl_auth_method_update.go
pkazmierczak: 🤦 🙏

command/acl_auth_method_create.go
jrasell: The [helper/flags](https://github.com/hashicorp/nomad/blob/main/helper/flags/autopilot_flags.go) package has some helpers that allow the use of pointer values as flags. I wonder if this would helper, although its general use purpose is for updates where you merge two values.

I think defaulting to false is fine for this flag in general. My gut says that having an administrator have to set the method to be default, is better than assuming it will be the default.

The TTL has to be set by the administrator when using this command as we cannot predict (baring looking it up via an API) what the TTL min/max values are. Therefore doing a `if ttl < 1` seems fine to me.

command/acl_auth_method_create.go
pkazmierczak: nice, I wasn't familiar with the lib. I'll apply this and all your commends about prediction below. thanks!

command/acl_auth_method_create.go
pkazmierczak: I made a change in 0a6a7a33929f75890264fb5fea023b77b47cdb0c, let me know if you find it ok.

command/acl_auth_method_create.go
pkazmierczak: Added in 0a6a7a33929f75890264fb5fea023b77b47cdb0c.

command/acl_auth_method_create.go
pkazmierczak: Agreed. I'm removing the default here.

command/acl_auth_method_create.go
pkazmierczak: Good idea!

I added support in 234393f47d2c77d1d3e0d824ad72981b867f6ae6 and tests in 92cf8246f61f04c09e033a38376bc83abc4968bb.

jrasell (approved):
LGTM!

PR #15330, opened 2022-11-20
refact: add conditional table logic

### LINKED ISSUE #8667: Deployment allocations table should handle empty state
If this table can be empty:

image

It could look more like this:

image

ChaiWithJai:
Resolves #8667 

image


philrenaud (approved):
Looks good! You might not need the `data-test-...` attributes on the new elements, but it couldn't hurt to have them, if only for consistency with our recent-allocations template.

PR #15355, opened 2022-11-21
[ui] Stabilize Percy tests for Exec and Logs Sidebar views

### LINKED ISSUE #15349: Percy is flakey on exec and log screens
image
image

philrenaud:
Resolves #15349 

- Adds a 1-second timer to allow [xterm](https://xtermjs.org/) to finish writing our example lines for the Exec view
- Adds conditional CSS to hide the (highly variable) allocation created-at fields that were showing up in the background of our sidebar snapshot

The result should be no more default red Xs for percy tests on Nomad. At least until I accidentally add more volatile screenshots again in the future :)

### DIFF
--- a/ui/tests/acceptance/exec-test.js
+++ b/ui/tests/acceptance/exec-test.js
@@ -297,7 +297,11 @@ module('Acceptance | exec', function (hooks) {
       } /bin/bash`
     );
 
-    await percySnapshot(assert);
+    const terminalTextRendered = assert.async();
+    setTimeout(async () => {
+      await percySnapshot(assert);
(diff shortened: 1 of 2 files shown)


ChaiWithJai (approved):

philrenaud:
> percy is still failing on this very PR, though. I don't have access to see what exactly the problem is, but is that intended?

@pkazmierczak yep, expected! Fails initially because this is the first time it's seeing the screens "fixed" by the code from this PR.

Triggers me to do a manual review, which I've approved; once in main, these new screenshots will be the new "standard" for others to compare to.

Issue #8575, opened 2020-07-31
Nomad `host_network` documentation and example improvments


nickethier:
The Multi-Host Interface feature released in 0.12 can be tricky to configure properly as there are numerous different ways users can configure their host's network interfaces. It's also not clear from existing documentation how a `host_network` interfaces with some driver's networking features (ex. #8488).

The following documentation improvments are underway to help this.

 * Updated examples in the docker driver documentation
 * Multi-Host Interface learn guide
 * New internals documentation on how to debug/understand how packets flow through Nomad's `bridge` networking mode, including the use of `host_network`.

Issue #10915, opened 2021-07-20
Nomad host_network default on first start


EtienneBruines:
### Nomad version
Output from `nomad version`
```
Nomad v1.1.2 (60638a086ef9630e2a9ba1e237e8426192a44244)
```
### Operating system and Environment details
```
Ubuntu 20.04 LTS
Linux nomad02 5.4.0-77-generic #86-Ubuntu SMP Thu Jun 17 02:35:03 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

### Issue
The `host_network "default"` is ignored for the [`bridge_network_subnet`](https://www.nomadproject.io/docs/configuration/client#bridge_network_subnet), if no jobs have been allocated yet:

```
client {
  enabled = true
  host_network "default" {
    cidr = "172.26.64.0/24"
  }
}
```

After some debugging, it seems like the issue is that the subnet / bridge has not been created yet (this happens when the first job is allocated and not before), and therefore cannot be referenced creating the host_network default.

```
# At the start, when no jobs have been allocated yet
root@nomad02:/etc/nomad.d# ip -br a
lo               UNKNOWN        127.0.0.1/8 ::1/128 
ens224           UP             172.16.17.136/24 172.16.17.2/24 172.16.17.4/24 fe80::250:56ff:fe8b:105e/64 
docker0          DOWN           172.17.0.1/16 

# After at least one job (e.g. hello-world) has been allocated:
root@nomad02:/etc/nomad.d# ip -br a
lo               UNKNOWN        127.0.0.1/8 ::1/128 
ens224           UP             172.16.17.136/24 172.16.17.2/24 172.16.17.4/24 fe80::250:56ff:fe8b:105e/64 
docker0          DOWN           172.17.0.1/16 fe80::42:3dff:fee1:810e/64 
nomad            UP             172.26.64.1/20 fe80::887d:b7ff:fefb:a8b4/64 
veth5f13f062@if3 UP             fe80::24b1:78ff:fec9:fbe4/64 
veth4a71ec15@if3 UP             fe80::106f:e2ff:fe53:fc2b/64 
```

### Reproduction steps
1. Install a nomad-client on a server
2. Define the `network_mode "default"` in the `nomad.hcl` file with the subnet as defined in `bridge_network_subnet`
3. Attempt to allocate a job, and then seeing the error (using the `example.nomad` file here):

```
      * Constraint "missing host network \"default\" for port \"db\"": 2 nodes excluded by filter
```

[...]

PR #15358, opened 2022-11-22
docs: networking

### LINKED ISSUE #9662: docs: remove references to removed NOMAD_ networking vars
We still have references to environment variables that are set when the `network` block is embedded in the task rather than the group. These should be removed from places like [/docs/runtime/interpolation](https://www.nomadproject.io/docs/runtime/interpolation) and [/docs/runtime/environment](https://www.nomadproject.io/docs/runtime/environment) (maybe more places?)

- `NOMAD_IP_<task>_<label>`
- `NOMAD_PORT_<task>_<label>`
- `NOMAD_ADDR_<task>_<label>`
- `NOMAD_HOST_PORT_<task>_<label>`
-  maybe others?

### LINKED ISSUE #10851: Inherit Docker Task networking configuration from Group Networking Block
### Proposal
When configuring a nomad job the user can specify a networking configuration inside different job groups. Each task can also have a separate configuration and when that task is a docker image, that configuration can include newtorking specifications. If a user specifies a networking mode in the group configuration they must also specify the same networking mode in the docker task driver. This can create confusion until/unless the user ends up doing more investigation into the docker task driver configuration (which defaults to `bridge`). It'd reduce confusion to make the behavior inherited between from the group networking block into the task networking block.

This could be complicated by the fact that the group networking block defaults to `host` but the task networking block defaults to `bridge` in docker

### Use-cases
Reduce required duplication and confusion in networking configuration

### Attempted Solutions
Duplicate configuration and add extra documentation/gotcha's for group vs task networking configuration.

### LINKED ISSUE #15187: Clarify bridge network isolation
### Proposal

I want to suggest to clarify the documentation on bridge network isolation and/or isolate bridge networks by default.

The [current documentation](https://developer.hashicorp.com/nomad/docs/job-specification/network#network-modes) says:

> When the network stanza is defined with bridge as the networking mode, all tasks in the task group share the same network namespace. [...] Tasks running within a network namespace are not visible to applications outside the namespace on the same host.

But looking at the actual configuration (Nomad latest, Consul latest, Connect enabled, no ACLs) I noticed that:

- All tasks share one bridge under `172.26.64.1/20`
- Forwarding is globally allowed to this bridge, looking at iptables rules I see things like:
```
-A FORWARD -m comment --comment "CNI firewall plugin rules" -j CNI-FORWARD
-A CNI-FORWARD -m comment --comment "CNI firewall plugin admin overrides" -j NOMAD-ADMIN
-A NOMAD-ADMIN -d 172.26.64.0/20 -o nomad -j ACCEPT
```

I was able to connect between tasks on the same host (assuming they bind to `0.0.0.0`) and was able to connect from another machine to the tasks running in a network namespace (assuming they bind to `0.0.0.0` and pointing to the Nomad box as the gateway for the bridge IP range).

This was not really clear from the docs to me and I had expected default firewalling between the bridge network and the outside world to be in place.

Of course, most of this is resolved by binding to `localhost`, or possibly by putting my own firewalling rules in place. But I don't want to break anything.

I think at least the docs could be clarified so it does not suggest that tasks are entirely invisible to applications outside of the namespace, because as long as you know the internal interface IP address of the task it can be reached. Another option would be to actually firewall this from the outside world and between task groups, but I have no idea what the consequences of that would be.

### LINKED ISSUE #15279: docs: correctly detail what files are read and parsed within `cni_config_dir`
The [client `cni_config_dir`](https://developer.hashicorp.com/nomad/docs/configuration/client#cni_config_dir) config parameter details that "*Filenames should use the .conflist extension.*" which is only half the truth. The fingerprinter will also look for files with [`json` and `conf`](https://github.com/hashicorp/nomad/blob/322c6b3dce06f91a34ba5674de48e28022ffd4a1/client/fingerprint/cni.go#L31) extensions and will attempt to parse these as a different object as files with the `conflist` extension. This undocumented behaviour can cause confusion when parsing errors occur.

The documentation should therefore be updated to detail all the extension types that the Nomad client will attempt to read and parse, as to avoid unnecessary confusion.

lgfa29:
Improve and expand Nomad networking documentation by creating a new top-level section with information and reference on how networking works in Nomad. Also removes some deprecated content and add more noticeable callouts in important configuration areas.  

Preview link: https://nomad-git-website-networking-hashicorp.vercel.app/nomad/docs/networking

Closes #15279
Closes #15187
Closes #10851
Closes #9662

https://github.com/hashicorp/nomad/issues/8575 ended up being left out for lack of time, but at lest we now have _where_ to write those docs.


website/content/docs/networking/index.mdx
```diff
@@ -0,0 +1,330 @@
...
+## Bridge networking
+
+Linux clients support a network [`mode`][network_mode] called [`bridge`]. A
+bridge network acts like a virtual network switch allowing processes connected
+to the bridge to reach each other while isolating them from others.
+
+When a Nomad agent starts in a Linux system it creates a bridge network called
+`nomad` (or the value set in [`bridge_network_name`]) using the [`bridge` CNI
+plugin][cni_bridge]. Before using this mode you must first [install the CNI
+plugins][cni_install] into your clients.
```
lgfa29, line 101: This should like to https://github.com/hashicorp/tutorials/pull/1005 (internal repo) once the new content is available.

website/content/docs/networking/index.mdx
```diff
@@ -0,0 +1,330 @@
...
+components behind the scenes, such as DNS servers and load balancers.
+
+This can be confusing at first since it is quite different from what you may
+be used to from other tools. This section explains how networking works in
+Nomad, some of the different patterns and configurations you are likely to find
+and use, and how Nomad differs from other tools in this aspect.
+
+## Allocation networking
+
+The base unit of scheduling in Nomad is an [allocation], which means that all
```
tgross, line 26:
```suggestion
The base unit of scheduling in Nomad is an [allocation][], which means that all
```

website/content/docs/networking/index.mdx
```diff
@@ -0,0 +1,330 @@
...
+group "..." {
+  network {
+    mode = "bridge"
+  }
+
+  task "..." {
+    driver = "docker"
+
+    config {
+      # This conclicts with the group-level network.mode configuration and
```
tgross, line 182:
```suggestion
      # This conflicts with the group-level network.mode configuration and
```

website/content/docs/networking/service-discovery.mdx
```diff
@@ -0,0 +1,143 @@
...
+# Service discovery
+
+Service discovery is the process of retrieving the low-level information
+necessary to access a service, such as its IP and port, via a high-level
+identifier, such as a human-friendly service name.
+
+The service information is stored in a service catalog which provide interfaces
+that can be used to query data. In Nomad there are two service catalogs:
+
+- the **native service discovery** catalog is embedded in Nomad and requires no
```
tgross, line 17:
```suggestion
- The **native service discovery** catalog is embedded in Nomad and requires no
```

(also a bunch of other locations where we have bullet points with full sentences)

tgross (commented):
This looks great!

website/content/docs/networking/service-mesh.mdx
```diff
@@ -0,0 +1,151 @@
...
+Since the mesh defines a closed boundary that only selected services can
+participate there are specialized proxies called gateways that can be used for
+mesh-wide connectivity. The types of gateways provided by Consul Service Mesh
+are:
+
+- **mesh gateways** allow communication between different service meshes.
+- **ingress gateways** allow services outside the mesh to connect to services
+  inside the mesh.
+- **egress gateways** allow services inside the mesh to communication with
+  services outside the mesh.
```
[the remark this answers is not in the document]
lgfa29: Yeah, good point. A shuffled things around a bit and linked each gateway to its corresponding Nomad jobspec config 👍

website/content/docs/concepts/plugins/cni.mdx
```diff
@@ -0,0 +1,58 @@
...
+page_title: Network Plugins
+description: Learn how Nomad manages custom user-specified network configurations.
+---
+
+# Network plugins
+
+Nomad has built-in support for scheduling compute resources such as
+CPU, memory, and networking. Nomad's network plugin support extends
+this to allow scheduling tasks with externally created storage
+volumes. Network plugins are third-party plugins that conform to the
```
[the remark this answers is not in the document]
lgfa29: Good catch, fixed!

website/content/docs/networking/cni.mdx
```diff
@@ -0,0 +1,196 @@
...
+addresses for the allocations from that same subnet using the host-local plugin.
+The configuration also specifies a default route for the allocations of the
+host-side bridge address.
+
+### firewall
+
+The firewall plugin creates firewall rules to allow traffic to/from the
+allocation's IP address via the host network. Nomad uses the iptables backend
+for the firewall plugin. This configuration creates two new iptables chains,
+CNI-FORWARD and NOMAD-ADMIN, in the filter table and add rules that allow the
```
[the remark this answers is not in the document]
lgfa29: I made an executive decision to have backticks everywhere 😄

website/content/docs/networking/cni.mdx
```diff
@@ -0,0 +1,196 @@
...
+
+### bridge
+
+The `bridge` plugin creates a bridge (virtual switch) named `nomad` that resides
+in the host network namespace. Because this bridge is intended to provide
+network connectivity to allocations, it's configured to be a gateway by setting
+`isGateway` to `true`. This tells the plugin to assign an IP address to the
+bridge interface
+
+The bridge plugin connects allocations (on the same host) into a bridge (virtual
```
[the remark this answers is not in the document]
lgfa29: Good point, I added this clarification here and in https://github.com/hashicorp/nomad/pull/15358/files#diff-26fb23e7998b097aa61b2edb406f5e2531b8b7e04f29ec4787feb268391a3382R92

website/content/docs/networking/index.mdx
```diff
@@ -0,0 +1,330 @@
...
+network topology and routing configuration allow the Nomad clients to
+communicate with each other.
+
+## Bridge networking
+
+Linux clients support a network [`mode`][network_mode] called [`bridge`][]. A
+bridge network acts like a virtual network switch allowing processes connected
+to the bridge to reach each other while isolating them from others.
+
+When a Nomad agent starts in a Linux system it creates a bridge network called
```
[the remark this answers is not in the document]
lgfa29: Oh great catch, thanks!

I updated this PR to reflect the current behaviour. Once it's merged I will add a note to #10915 that this page will need to be updated as well.

website/content/docs/networking/index.mdx
```diff
@@ -0,0 +1,331 @@
...
+
+Bridge networking is at the core of [service mesh][] and a requirement when
+using [Consul Service Mesh][consul_service_mesh].
+
+### Bridge networking with Docker
+
+The Docker daemon manages its own network configuration and creates its own
+[bridge network][docker_bridge], network namespaces, and [`iptable`
+rules][docker_iptables]. Tasks using the `docker` task driver connect to the
+Docker bridge instead of using the one create by Nomad and, by default, each
```
tgross, line 156:
```suggestion
Docker bridge instead of using the one created by Nomad and, by default, each
```

website/content/docs/networking/index.mdx
```diff
@@ -0,0 +1,331 @@
...
+incorrectly.
+
+<!-- Source: https://drive.google.com/file/d/1q4a2ab0TyLEPdWiO2DIianAPWuPqLqZ4/view?usp=share_link -->
+[![Nomad Bridge](/img/networking/docker_bridge.png)](/img/networking/docker_bridge.png)
+
+The tasks in the rightmost allocation are not able to communicate with each
+other using their loopback interface because they were placed in different
+network namespaces.
+
+Since the group `network.mode` is `bridge`, Nomad creates the pause cointainer
```
tgross, line 202:
```suggestion
Since the group `network.mode` is `bridge`, Nomad creates the pause container
```

website/content/docs/networking/cni.mdx
```diff
@@ -0,0 +1,208 @@
...
+
+CNI plugins are installed and configured on a per-client basis. Nomad consults
+the path given in the client's [`cni_path`][] to find CNI plugin executables.
+
+## CNI configuration files
+
+The CNI specification defines a network configuration format for administrators.
+It contains directives for both the orchestrator and the plugins to consume.
+At plugin execution time, this configuration format is interpreted by the
+runtime and transformed in to a form to be passed to the plugins.
```
tgross, line 42: "The runtime" is ambiguous here. I'd probably do something like:

```suggestion
At plugin execution time, this configuration format is interpreted by Nomad
and transformed into arguments for the plugins.
```

tgross (commented):
I've got a few more pages to re-review here @lgfa29 but this is looking great. Submitting the comments I have so far to get them to you sooner rather than later.

website/content/docs/networking/service-discovery.mdx
```diff
@@ -0,0 +1,144 @@
...
+
+The service information is stored in a service catalog which provide interfaces
+that can be used to query data. In Nomad there are two service catalogs:
+
+- The **native service discovery** catalog is embedded in Nomad and requires no
+  additional infrastructure.
+
+- The **[Consul service discovery catalog][consul_sd]** requires access to a
+  Consul cluster, but it provides additional features, including a [DNS
+  interface][consul_dns][] and [service mesh][] capabilities.
```
tgross, line 22: Looks like some extra syntax snuck in:

```suggestion
  interface][consul_dns] and [service mesh][] capabilities.
```

tgross (approved):
Ok, I've left a few more remarks but nothing blocking once the typos are fixed. Let's get this merged and then we can iterate on it if we want. This is great work!

website/content/docs/concepts/plugins/cni.mdx
```diff
@@ -0,0 +1,55 @@
...
+reference plugins][cni_ref] because certain Nomad networking features, like
+`bridge` network mode and Consul service mesh, leverage them to provide an
+operating-system agnostic interface to configure workload networking.
+
+Custom networking in Nomad is accomplished with a combination of CNI plugin
+binaries and CNI configuration files.
+
+## CNI plugins
+
+Spec-compliant plugins should work with Nomad. However, it's possible a plugin
```
jrasell, line 26: feel free to ignore, but I was always told not to start sentence with however
```suggestion
Spec-compliant plugins should work with Nomad, however, it's possible a plugin
```

website/content/docs/concepts/plugins/cni.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+## CNI configuration files
+
+CNI defines a network configuration format for administrators. It contains
+directives for both the orchestrator and the plugins to consume.
+At plugin execution time, this configuration format is interpreted by the
+runtime and transformed in to a form to be passed to the plugins.
+
+Nomad reads all files with a `.conflist` extension in the [`cni_config_dir`][]—
+`/opt/cni/config` by default.
```
jrasell, line 43: It would be nice to update this based on the detail included within https://github.com/hashicorp/nomad/issues/15279 to avoid future confusion with users.

website/content/docs/networking/cni.mdx
```diff
@@ -0,0 +1,208 @@
...
+reference plugins][cni_ref] because certain Nomad networking features, like
+`bridge` network mode and Consul service mesh, leverage them to provide an
+operating-system agnostic interface to configure workload networking.
+
+Custom networking in Nomad is accomplished with a combination of CNI plugin
+binaries and CNI configuration files.
+
+## CNI plugins
+
+Spec-compliant plugins should work with Nomad. However, it's possible a plugin
```
jrasell, line 28: feel free to ignore, but I was always told not to start sentence with however
```suggestion
Spec-compliant plugins should work with Nomad, however, it's possible a plugin
```

website/content/docs/networking/cni.mdx
```diff
@@ -0,0 +1,208 @@
...
+
+## CNI configuration files
+
+The CNI specification defines a network configuration format for administrators.
+It contains directives for both the orchestrator and the plugins to consume.
+At plugin execution time, this configuration format is interpreted by the
+runtime and transformed in to a form to be passed to the plugins.
+
+Nomad reads all files with a `.conflist` extension in the [`cni_config_dir`][]—
+`/opt/cni/config` by default.
```
jrasell, line 45: It would be nice to update this based on the detail included within https://github.com/hashicorp/nomad/issues/15279 to avoid future confusion with users.

jrasell (approved):
Looks great and will be extremely useful. I made a couple of minor comments, but nothing that should be considered blocking or that can't be fixed in later PRs.

website/content/docs/networking/cni.mdx
```diff
@@ -0,0 +1,208 @@
...
+
+## Using CNI networks with Nomad jobs
+
+To specify that a job should use a CNI network, set the task group's
+network's [`mode`][] attribute to the value `cni/<your_cni_config_name>`.
```
shoenig, line 50:
```suggestion
network [`mode`][] attribute to the value `cni/<your_cni_config_name>`.
```

website/content/docs/networking/service-discovery.mdx
```diff
@@ -0,0 +1,144 @@
...
+necessary to access a service, such as its IP and port, via a high-level
+identifier, such as a human-friendly service name.
+
+The service information is stored in a service catalog which provide interfaces
+that can be used to query data. In Nomad there are two service catalogs:
+
+- The **native service discovery** catalog is embedded in Nomad and requires no
+  additional infrastructure.
+
+- The **[Consul service discovery catalog][consul_sd]** requires access to a
```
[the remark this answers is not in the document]
lgfa29: Good catch, moving them inside did seem to fix it. Thanks!

website/content/docs/networking/index.mdx
```diff
@@ -0,0 +1,331 @@
...
+    }
+  }
+}
+```
+
+The diagram below illustrates what happens when a Docker task is configured
+incorrectly.
+
+<!-- Source: https://drive.google.com/file/d/1q4a2ab0TyLEPdWiO2DIianAPWuPqLqZ4/view?usp=share_link -->
+[![Nomad Bridge](/img/networking/docker_bridge.png)](/img/networking/docker_bridge.png)
```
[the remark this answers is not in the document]
lgfa29: Yeah, good point. There is actually a correct and an incorrect allocation in the diagram. I updated the image so it labels the incorrect allocation as misconfigured. I also hope to be able to clean this up a bit using Figma 😅 

docker_bridge

website/content/docs/concepts/plugins/cni.mdx
lgfa29: Thanks! Updated with the extra file extensions, though I'm not sure if my description and understanding is 100% correct:
https://nomad-git-website-networking-hashicorp.vercel.app/nomad/docs/networking/cni#cni-configuration-files

website/content/docs/networking/index.mdx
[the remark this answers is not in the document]
tgross: That's great!

website/content/docs/concepts/plugins/cni.mdx
jrasell: That makes sense to me; the CNI spec is a very light on any real detail.

PR #15365, opened 2022-11-22
bug:  enable `jobs.index` page to update in real-time

### LINKED ISSUE #10555: ui: Jobs list doesn’t show a new job (on a locally-built binary?)

ChaiWithJai:
Resolves #10555

The `jobs.index` view is bound the model set on the `jobs.index` route. This model does not reactively update to changes in the `ember-data` store (instead it is subscribed to a set of `RecordIdenitfiers`... I believe).

That's why we're able to see job updates, but when a new job is added, our view does not update.

To fix the problem, we need to send a notification to the controller after the `query` and associated callback resolves meaning that our payload has successfully been added/updated to our `ember-date` store.

We also have one more challenge handling the wildcard (`*`) namespace which is a defacto `findAll` query. We have updated our `matchingRecord` algorithm to handle this case.

### DIFF
--- a/ui/app/adapters/watchable.js
+++ b/ui/app/adapters/watchable.js
@@ -143,7 +143,12 @@ export default class Watchable extends ApplicationAdapter {
         );
 
         matchingRecords.forEach((record) => {
-          removeRecord(store, record);
+          const IDValue = record.get('plainId') || record.get('id');
+          const storedRecordNotFoundInPayload =
+            IDValue && !payload.find((r) => r.ID === IDValue);
+          if (storedRecordNotFoundInPayload) {
+            removeRecord(store, record);
+          }
         });
 
         return payload;
--- a/ui/app/adapters/watchable.js
+++ b/ui/app/adapters/watchable.js
@@ -133,7 +133,8 @@ export default class Watchable extends ApplicationAdapter {
                 return true;
               } else if (mapping.attr === 'namespace') {
                 return (
-                  get(record, 'namespace.name') === query[mapping.queryParam]
+                  record.belongsTo('namespace').id() ===
+                  query[mapping.queryParam]
                 );
               }
             }
--- a/ui/app/adapters/watchable.js
+++ b/ui/app/adapters/watchable.js
@@ -125,16 +125,17 @@ export default class Watchable extends ApplicationAdapter {
 
       // Remove existing records that match this query. This way if server-side
       // deletes have occurred, the store won't have stale records.
-      store
-        .peekAll(type.modelName)
-        .filter((record) =>
-          queryParamsToAttrs.some(
-            (mapping) => get(record, mapping.attr) === query[mapping.queryParam]
-          )
-        )
-        .forEach((record) => {
-          removeRecord(store, record);
-        });
+      const matchingRecords = store.peekAll(type.modelName).filter((record) =>
+        queryParamsToAttrs.some((mapping) => {
+          if (mapping.queryParam === 'namespace' && query.namespace === '*')
+            return true;
+          return get(record, mapping.attr) === query[mapping.queryParam];
+        })
+      );
+
+      matchingRecords.forEach((record) => {
+        removeRecord(store, record);
+      });
 
       return payload;
     });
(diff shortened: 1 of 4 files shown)


ui/app/adapters/watchable.js
```diff
@@ -111,33 +111,36 @@ export default class Watchable extends ApplicationAdapter {
...
+          attr: adapter.queryParamsToAttrs[key],
+        }));
+
+        // Remove existing records that match this query. This way if server-side
+        // deletes have occurred, the store won't have stale records.
+        const matchingRecords = store.peekAll(type.modelName).filter((record) =>
+          queryParamsToAttrs.some((mapping) => {
+            if (mapping.queryParam === 'namespace' && query.namespace === '*')
+              return true;
+            return get(record, mapping.attr) === query[mapping.queryParam];
```
ChaiWithJai, line 133: @philrenaud This algorithm needs more work. I tried to make it minimal change by only adding one life, which works great if no filters are applied... however, if we apply the `namespace` filter, we're could run into issues.

ui/app/controllers/jobs/index.js
```diff
@@ -187,14 +189,26 @@ export default class IndexController extends Controller.extend(
...
-    if (!this.model || !this.model.jobs) return [];
-    return this.model.jobs
+    let jobs = null;
+    if (this.currentCount) {
+      jobs = this.store.peekAll('job');
+    } else {
+      jobs = this.model.jobs;
+    }
+
+    if ((!this.model || !this.model.jobs) && jobs) return [];
```
ChaiWithJai, line 210: `bug`: This should say if 

```
if (!jobs) return []
```

ui/app/services/watch-list.js
```diff
@@ -1,11 +1,16 @@
...
+import { computed, set } from '@ember/object';
 import { readOnly } from '@ember/object/computed';
+import { inject as service } from '@ember/service';
 import { copy } from 'ember-copy';
 import Service from '@ember/service';
 
 let list = {};
 
 export default class WatchListService extends Service {
+  @service store;
```
ChaiWithJai, line 10: `nit`: remove unused dependency

ui/app/services/watch-list.js
```diff
@@ -1,11 +1,16 @@
...
+
+  jobUpdateCount = 0;
```
ChaiWithJai, line 12: @philrenaud I tried to make the minimal surface change here. We could try to make this more generic by converting this into a dictionary, however, I imagine we could gut the watcher system once its time for a redesign.

philrenaud (changes requested):
I am still not sure this is working as intended. I am using a management token with 2 jobs running (shown in the return here) on the default namespace, and I see no jobs in my index. (on the main branch, I see both jobs) 
image

lgfa29:
For me it never shows the job, even after a refresh. Here's a replay:
https://app.replay.io/recording/jobs-nomad--8a2cadb9-0747-4eb6-b7e9-a07244a22fdd

ChaiWithJai:
> I am still not sure this is working as intended. I am using a management token with 2 jobs running (shown in the return here) on the default namespace, and I see no jobs in my index. (on the main branch, I see both jobs) image

Looks like a namespace issue. We have this wonky behavior where the namespace filter only appears when there are multiple namespaces configured. I didn't handle the case where a user isn't using namespaces. Good catch!

philrenaud:
Resolved in https://github.com/hashicorp/nomad/pull/20452

PR #15366, opened 2022-11-23
acl: replicate auth-methods from federated cluster leaders.

jrasell:
I rejigged some on the leader code so it was easier and made more sense when scrolling.

The hash has been added to the stub object so this can be used when performing replication diffs as is done with other ACL objects. It also now includes the create and modify indexes.

The auth method struct object has JSON marhsal/unmarshal implementations to allow for the time.Duration parameter.

Related https://github.com/hashicorp/nomad/issues/13120

### DIFF
--- a/api/acl.go
+++ b/api/acl.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"encoding/json"
 	"errors"
 	"time"
 )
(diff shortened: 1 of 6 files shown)


pkazmierczak (approved):
LGTM!

Issue #15094, opened 2022-10-31
sysbatch/system type jobs fail to be scheduled when using a multi-node-multi-writer CSI volume


mmeier86:
### Nomad version
Nomad v1.4.1 (2aa7e66bdb526e25f59883952d74dad7ea9a014e)

### Operating system and Environment details
Underlying systems are either Ubuntu or Debian Linux, with up to date patches. Debian 11, Ubuntu 22.04.

The Ceph CSI plugin is used in version 3.6.2.

### Issue

When adding a CSI `volume` stanza to a `sysbatch` job, Nomad consistently says "missing CSI Volume" when trying to run the job.

The same behavior appears when setting the `type` to `system`, so it seems that the `system` part is the problem here, while pure `batch` or `service` jobs with the `type` being the only difference work without a problem.

### Reproduction steps

First: Create a Ceph cluster and a Nomad cluster and configure Ceph CSI for the Nomad cluster. Then create a volume as follows:

```hcl
id = "vol-backup-s3"
name = "vol-backup-s3"
type = "csi"
plugin_id = "ceph-csi-cephfs"

capability {
  access_mode     = "multi-node-multi-writer"
  attachment_mode = "file-system"
}

secrets {
  adminID  = "<ceph client ID here>"
  adminKey = "<ceph client key here>"
}

parameters {
  clusterID = "<ceph cluster id here>"
  fsName = "<name of the Ceph FS here>"
  pool = "<name of the pool here>"
}
```

Create the volume with `nomad volume create`.

Then Create a simple job like this:
```hcl
job "csi-test" {
  datacenters = ["homenet"]

  type = "sysbatch"
  priority = 60

  constraint {
    attribute = "${node.class}"
    value     = "internal"
  }

  group "csi-test" {

    volume "vol-backup-s3" {
      type            = "csi"
      source          = "vol-backup-s3"
      attachment_mode = "file-system"
      access_mode     = "multi-node-multi-writer"
    }

    task "csi-test" {
      driver = "docker"

      config {
        image = "alpine:3.16.0"

        args = [
          3600
        ]
        command = "sleep"

      }

      volume_mount {
        volume      = "vol-backup-s3"
        destination = "/hn-data/buckets"
      }

      resources {
        cpu = 200
        memory = 200
        memory_max = 1024
      }
    }
  }
}
```
Plan the job with `nomad job plan` and observe the error message showing up:
```
Job: "csi-test"
[...]

PR #15372, opened 2022-11-23
scheduler: set job on system stack for CSI feasibility check

tgross:
Fixes https://github.com/hashicorp/nomad/issues/15094

When the scheduler checks feasibility of each node, it creates a "stack" which carries attributes of the job and task group it needs to check feasibility for. The `system` and `sysbatch` scheduler use a different stack than `service` and `batch` jobs. This stack was missing the call to set the job ID and namespace for the CSI check. This prevents CSI volumes from being scheduled for system jobs whenever the volume is in a non-default namespace.

Set the job ID and namespace to match the generic scheduler.

### DIFF
--- a/.changelog/15372.txt
+++ b/.changelog/15372.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+csi: Fixed a bug where volumes in non-default namespaces could not be scheduled for system or sysbatch jobs
+```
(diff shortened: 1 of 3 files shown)


tgross:
This is currently failing the linter check because of https://github.com/hashicorp/nomad/pull/15373 but once that's merged I can rebase this on `main` (or we can merge as-is if everything else is green).

lgfa29 (approved):
Oh nice catch! I was investigating this, but got sidetracked and didn't have the chance to get back to it. Thanks for taking it over!

PR #15380, opened 2022-11-24
Serialize auth-methods response to use "name" as primary key

philrenaud:


jrasell (commented):
The UI tests seem to be failing due to the `@classic` decorator.

philrenaud:
> The UI tests seem to be failing due to the `@classic` decorator.

Yep, had it commented in there just in case. Re-added in https://github.com/hashicorp/nomad/pull/15380/commits/757cd9d5ec6f1782fbdcb4ede6455304f0daba9d and all are passing now.

jrasell (approved):

jrasell:
> > The UI tests seem to be failing due to the `@classic` decorator.
> 
> Yep, had it commented in there just in case. Re-added in [757cd9d](https://github.com/hashicorp/nomad/commit/757cd9d5ec6f1782fbdcb4ede6455304f0daba9d) and all are passing now.

Ah my bad; I was clearly looking at the wrong commit test runs. :(

PR #11872, opened 2022-01-18
deps: upgrade docker and runc

shoenig:
(going in 1.3.0)

This PR upgrades
 - docker dependency to the latest tagged release (v20.10.12)
 - runc dependency to the latest tagged release (v1.0.3)

Docker does not abide by [semver](https://github.com/moby/moby/issues/39302), so it is marked +incompatible,
and transitive dependencies are upgrade manually.

Runc made three relevant breaking changes

 * cgroup manager .Set changed to accept Resources instead of Cgroup
   https://github.com/opencontainers/runc/commit/3f6594675675d4e88901c782462f56497260b1d2

 * config.Device moved to devices.Device
   https://github.com/opencontainers/runc/pull/2679

 * mountinfo.Mounted now returns an error if the specified path does not exist
   https://github.com/moby/sys/blob/mountinfo/v0.5.0/mountinfo/mountinfo.go#L16

Closes #10832

PR #15397, opened 2022-11-25
deps: apply fixes after Docker update

lgfa29:
https://github.com/hashicorp/nomad/pull/15229 updated Docker in `release/1.2.x`, but it had some breaking changes that [caused CI to fail](https://app.circleci.com/pipelines/github/hashicorp/nomad/35387/workflows/04f7a731-7f34-44dc-8d97-45173a34d6b2/jobs/391854). The fixes had already been done in https://github.com/hashicorp/nomad/pull/11872 for 1.3.0+, that's why only `release/1.2.x` fails. 

I applied the Docker-related changes here, I believe the other changes are related the `runc` update.


tgross (approved):
LGTM!

At the time we did the fixes for 1.3.x I think we weren't being as rigorous about getting dependency updates backported, so makes sense that we missed this. Thanks for picking this up!

shoenig (approved):

PR #15410, opened 2022-11-28
acl: sso auth methods RPC/API/CLI should return created or updated objects

pkazmierczak:
Currently CRUD code that operates on SSO auth methods does not return created or updated object upon creation/update. This is bad UX and inconsistent behavior compared to other ACL objects like roles, policies or tokens. 

This PR fixes it. 

Relates to #13120

### DIFF
--- a/nomad/acl_endpoint.go
+++ b/nomad/acl_endpoint.go
@@ -1735,7 +1735,9 @@ func (a *ACL) UpsertAuthMethods(
 		if err != nil {
 			return structs.NewErrRPCCodedf(400, "ACL auth method lookup failed: %v", err)
 		}
-		reply.AuthMethods = append(reply.AuthMethods, lookupAuthMethod)
+		if lookupAuthMethod != nil {
+			reply.AuthMethods = append(reply.AuthMethods, lookupAuthMethod)
+		}
 	}
 
 	// Update the index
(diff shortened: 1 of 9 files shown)


nomad/acl_endpoint.go
```diff
@@ -1724,6 +1724,20 @@ func (a *ACL) UpsertAuthMethods(
...
+	stateSnapshot, err := a.srv.State().Snapshot()
+	if err != nil {
+		return err
+	}
+	for _, method := range args.AuthMethods {
+		lookupAuthMethod, err := stateSnapshot.GetACLAuthMethodByName(nil, method.Name)
+		if err != nil {
+			return structs.NewErrRPCCodedf(400, "ACL auth method lookup failed: %v", err)
+		}
+		reply.AuthMethods = append(reply.AuthMethods, lookupAuthMethod)
```
tgross, line 1738: `GetACLAuthMethodByName` can return nil without an error if the method doesn't exist. Practically speaking I can't see this happening but if we had a bug that allowed `raftApply` to return without error but not write the method, we could be appending a nil pointer to the results (which would crash in the CLI because the formatted doesn't consider the nil case).

For belt-and-suspenders can we check that `lookupAuthMethod` isn't nil before adding it to the reply?

tgross (approved):
LGTM!

PR #15417, opened 2022-11-29
Pre forwarding authentication

tgross:
Upcoming work to instrument the rate of RPC requests by consumer (and eventually rate limit) require that we authenticate a RPC request before forwarding. Add a new top-level `Authenticate` method to the server and have it return an `AuthenticatedIdentity` struct. RPC handlers will use the relevant fields of this identity for performing authorization.

I've split this PR into several commits to make it a little easier to review:
* The main implementation of `Authenticate`
* Provide a new RPC `ACL.WhoAmI` for debugging authentication. This endpoint returns the same `AuthenticatedIdentity` that will be used by RPC handlers. At some point we might want to give this an equivalent HTTP endpoint but I didn't want to add that to our public API until some of the other Workload Identity work is solidified, especially if we don't need it yet.
* A full coverage test of the `Authenticate` method. This sets up two server nodes with mTLS and ACLs, some tokens, and some allocations with workload identities.
* Wire up an example of using `Authenticate` in the `Namespace.Upsert` RPC and see how authorization happens after forwarding.
* A new semgrep rule for `Authenticate`, which we'll need to update once we're ready to wire up more RPC endpoints with authorization steps.


nomad/acl.go
```diff
@@ -2,15 +2,117 @@ package nomad
...
 	lru "github.com/hashicorp/golang-lru"
 	"github.com/hashicorp/nomad/acl"
+	"github.com/hashicorp/nomad/helper"
 	"github.com/hashicorp/nomad/nomad/state"
 	"github.com/hashicorp/nomad/nomad/structs"
 )
 
+// Authenticate extracts an AuthenticatedIdentity from the request context or
+// provided token. The caller can extract an acl.ACL, WorkloadIdentity, or other
+// identifying token to use for authorization.
```
tgross, line 18: Note for reviewers: something that's a little quirky here is that because we're making a stale query against the state store, it's possible a follower gets a different value than the leader if an ACL token or alloc has just been created. Because we don't expect the follower to actually serve the request unless the user has specifically asked for a stale query, I think this ends up being ok but will produce slightly off metrics. (The race is also just really unlikely unless a follower is far behind.)

shoenig: Want to add this clarification to the doc string?

nomad/acl.go
```diff
@@ -2,15 +2,117 @@ package nomad
...
+	case structs.ErrTokenExpired:
+		return nil, err
+
+	case structs.ErrTokenInvalid:
+		// if it's not a UUID it might be an identity claim
+		claims, err := s.VerifyClaim(secretID)
+		if err != nil {
+			return nil, err
+		}
+		// unlike the state queries, errors here are invalid tokens
```
shoenig, line 41: Is this really saying, "we tried to read the ACL from state store and it was invalid, now we tried to read it as an identity claim and it was invalid, and now there are no other options." ?

nomad/acl.go
```diff
@@ -2,15 +2,117 @@ package nomad
...
+		return identity, nil
+
+	case structs.ErrTokenNotFound:
+		// Check if the secret ID is the leader's secret ID, in which case treat
+		// it as a management token and record the server's node ID
+		leaderAcl := s.getLeaderAcl()
+		if leaderAcl != "" && secretID == leaderAcl {
+			identity.ACLToken = structs.LeaderACLToken
+		}
+
```
shoenig, line 52: going from the comment, do we want to bail here if the secretID is _not_ the leader's secretID?

nomad/acl.go
```diff
@@ -2,15 +2,117 @@ package nomad
...
+		remoteAddr = ctx.Session.RemoteAddr().(*net.TCPAddr)
+	}
+	if remoteAddr == nil && ctx.Conn != nil {
+		remoteAddr = ctx.Conn.RemoteAddr().(*net.TCPAddr)
+	}
+	if remoteAddr != nil {
+		identity.RemoteIP = remoteAddr.IP
+	}
+
+	return identity, nil
```
shoenig, line 102: Would we want to special case an error here where identity contains no information? Probably not even possible to be in that condition though

nomad/acl.go
```diff
@@ -185,27 +293,28 @@ func (s *Server) ResolveSecretToken(secretID string) (*structs.ACLToken, error)
...
-			return nil, structs.ErrTokenExpired
-		}
+	token, err := snap.ACLTokenBySecretID(nil, secretID)
+	if err != nil {
+		return nil, err
+	}
+	if token == nil {
+		return nil, structs.ErrTokenNotFound
+	}
+	if token.IsExpired(time.Now().UTC()) {
```
shoenig, line 316: Just to nitpick - it'd be nice to get away from calling `time.Now` directly since it makes testing more clumsy; alternatives could be passing in "now" as a parameter, or turn the function into a method and have like a libtime.Clock as a struct field which can then be mocked

shoenig (commented):
Really like the test coverage here! Left some Q's/Comments mostly to help clarify and build my own mental model

nomad/acl.go
```diff
@@ -2,15 +2,117 @@ package nomad
...
+	// get the user ACLToken or anonymous token
+	aclToken, err := s.ResolveSecretToken(secretID)
+	identity := &structs.AuthenticatedIdentity{ACLToken: aclToken}
+
+	switch err {
+	case nil:
+		// missing or anonymous ACL will fall-through for further annotation
+		// based on the RPC context
+		if aclToken == nil || aclToken != structs.AnonymousACLToken {
+			return identity, nil
```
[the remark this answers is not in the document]
tgross: This is meant to indicate something in my head like "we're not worrying here about the missing/anon case" but yeah that's totally confusing in retrospect. Will fix!

nomad/acl.go
tgross: Yes! Will clarify.

nomad/acl.go
tgross: If it's not, we need to fallthru to check if the secret ID is a node secret instead. I missed that we should be breaking out of the switch here though so that we don't bother checking the node secret if we _do_ have the leader ACL.

nomad/acl.go
tgross: Yeah I don't think you can even hit this code path if you're calling it as a static handler (because `ctx` would be nil). It might be worth returning an error here just so we know if we missed something.

nomad/acl.go
tgross: Yeah I like the idea of having a mockable clock. This is moved code rather than new code in this PR though, so I'll look at returning to that in a separate PR if that works for you?

nomad/acl.go
tgross: Will do

nomad/acl.go
tgross: Also, the comment should be on the error above, not this line 🤦

nomad/acl.go
tgross: Done

nomad/acl_endpoint.go
```diff
@@ -1964,3 +1964,20 @@ func (a *ACL) GetAuthMethods(
...
+		return err
+	}
+	args.SetIdentity(identity)
+
+	if done, err := a.srv.forward("ACL.WhoAmI", args, args, reply); done {
+		return err
+	}
+	defer metrics.MeasureSince([]string{"nomad", "acl", "whoami"}, time.Now())
+
+	reply.Identity = args.GetIdentity()
```
schmichael, line 1981: What are the security implications of roundtripping this? I know that right now it's only reachable by devs and uniquely motivated users (because they'd need access to the rpc port, manually perform the rpc, and have a valid tls cert).

The only secret it returns is `ACLToken.SecretID`, and AFAICT the only way to have that set is to have passed that in.

So I think we're safe?

That being said perhaps we should at implement a `GetRedactedIdentity()` (or redact secrets from `GetIdentity()` by default and add an `UnsafeGetIdentity()` or similar if we callers need to reach in and use the ACLToken.SecretID directly).

As we strengthen our notion of identity more and more I think this (or similar) is going to be a really useful endpoint to expose.

nomad/structs/structs.go
```diff
@@ -435,6 +447,30 @@ func (w WriteRequest) AllowStaleRead() bool {
...
 
+func (w *WriteRequest) SetIdentity(identity *AuthenticatedIdentity) {
+	w.identity = identity
+}
+
+func (w WriteRequest) GetIdentity() *AuthenticatedIdentity {
+	return w.identity
+}
+
+type AuthenticatedIdentity struct {
```
schmichael, line 458: docstring at a least enough to hint at folks where to learn more about this abstraction

Especially the fact that the fields are largely used independently makes its actual use different from what someone might assume at a glance.

nomad/structs/structs.go
```diff
@@ -435,6 +447,30 @@ func (w WriteRequest) AllowStaleRead() bool {
...
+	ClientID string
+	ServerID string
+	TLSName  string
+	RemoteIP net.IP
+}
+
+func (ai *AuthenticatedIdentity) GetACLToken() *ACLToken {
+	if ai == nil {
+		return nil
+	}
```
schmichael, line 470: Does Claims need similar treatment?

schmichael (approved):
I'm really excited about this abstraction for auth! 

None of my comments are blockers, so feel free to ship as-is.

That being said that `Authenticate` method is a lot to take in, and we've had a CVE from not quite getting complicated ACL checks right in the past, so I left some comments that might make it more maintainable.

nomad/acl.go
```diff
@@ -1,16 +1,128 @@
...
+	if ctx == nil {
+		return identity, nil
+	}
+
+	// Previously-connected clients will have a NodeID set and will be a large
+	// number of the RPCs sent, so we can fast path this case
+	if ctx.NodeID != "" {
+		identity.ClientID = ctx.NodeID
+		return identity, nil
+	}
```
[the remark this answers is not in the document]
tgross: Good call, and that makes some of the later bits easier to understand as well.

nomad/acl.go
```diff
@@ -1,16 +1,128 @@
...
+// provided token. The caller can extract an acl.ACL, WorkloadIdentity, or other
+// identifying token to use for authorization.
+//
+// Note: when called on the follower we'll be making stale queries, so it's
+// possible if the follower is behind that the leader will get a different value
+// if an ACL token or allocation's WI has just been created.
+func (s *Server) Authenticate(ctx *RPCContext, secretID string) (*structs.AuthenticatedIdentity, error) {
+	// get the user ACLToken or anonymous token
+	aclToken, err := s.ResolveSecretToken(secretID)
+	identity := &structs.AuthenticatedIdentity{ACLToken: aclToken}
```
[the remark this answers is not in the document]
tgross: Ok, I've done that throughout the first half of the function. For everything after that, where we reach the server identification through the context, it gets noticeably less readable because you've got to get those non-disjoint pieces of data together.

nomad/acl.go
tgross: The two blocks inside this case are very different, so it made for a weird helper. I've fixed that by just using an `else` block and that seems less subtle to trace.

nomad/acl_endpoint.go
tgross: Typically we've done redaction at the structs->API layer, especially with the serialization extensions that Chris introduces on Node (which we also use on CSIVolume these days and I have an open issue to do for Job too).

PR #15418, opened 2022-11-30
docs: clarify autoscaling factor and threshold for target-value plugin

lgfa29:
How `threshold` is used: https://github.com/hashicorp/nomad-autoscaler/blob/v0.3.7/plugins/builtin/strategy/target-value/plugin/plugin.go#L161-L162

How `factor` and `new_count` are calculated: https://github.com/hashicorp/nomad-autoscaler/blob/v0.3.7/plugins/builtin/strategy/target-value/plugin/plugin.go#L108-L134


jrasell (approved):
LGTM!

PR #15421, opened 2022-11-30
docs: correct "User" attribute in Podman Task Driver Docs

PJDuszynski:
The Podman task driver implements the `user` attribute outside of the task config stanza, however the current Nomad docs have the task within the task config stanza.  This PR corrects this error.

https://github.com/hashicorp/nomad-driver-podman


lgfa29:
Thanks for the PR @PJDuszynski!

Since this is not a task driver configuration value I think we can actually just remove this item since it's covered in the [`task`](https://developer.hashicorp.com/nomad/docs/job-specification/task#user) documentation?

lgfa29:
I updated the page so it matches how the Docker driver documents this.

lgfa29 (approved):

PR #15430, opened 2022-11-30
provide `RPCContext` to all RPC handlers

tgross:
Upcoming work to instrument the rate of RPC requests by consumer (and eventually rate limit) requires that we thread the `RPCContext` through all RPC handlers so that we can access the underlying connection. This changeset adds the context to everywhere we intend to initially support it and therefore intentionally excludes streaming RPCs and client RPCs.

To improve the ergonomics of adding the context everywhere its needed and to clarify the requirements of dynamic vs static handlers, I've also done a good bit of refactoring here:

* canonicalized the RPC handler fields so they're as close to identical as possible without introducing unused fields (i.e. I didn't add loggers if the handler doesn't use them already).
* canonicalized the imports in the handler files.
* added a `NewExampleEndpoint` constructor function for each handler that ensures we're constructing the handlers with the required arguments.
* reordered the registration in server.go to match the order of the files (to make it easier to see if we've missed one), and added a bunch of commentary there as to what the difference between static and dynamic handlers is.

---

Related to https://github.com/hashicorp/nomad/pull/15417. This will need a matching Enterprise PR as well.

No changelog for this as it should be entirely invisible to users.

### DIFF
--- a/nomad/acl_endpoint.go
+++ b/nomad/acl_endpoint.go
@@ -9,10 +9,11 @@ import (
 	"strings"
 	"time"
 
-	metrics "github.com/armon/go-metrics"
-	log "github.com/hashicorp/go-hclog"
-	memdb "github.com/hashicorp/go-memdb"
+	"github.com/armon/go-metrics"
+	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-memdb"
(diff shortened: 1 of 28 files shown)


shoenig (approved):
:rocket:

PR #15431, opened 2022-11-30
bug:  hitting refresh on `/jobs/:jobId/services/:serviceId` doesn't display data

### LINKED ISSUE #15416: hitting refresh on `/jobs/:jobId/services/:serviceId` doesn't display data
There is a data loading issue when hitting page refresh on `/jobs/:jobId/services/:serviceId`. The rendered view looks like:

image

If you navigate away from the page and then back information displays correctly. This is likely an issue regarding asynchronous relationships not being handled in the `route` and thus making a request when the template requests the data from the relationship.

ChaiWithJai:
resolves #15416

### DIFF
--- a/ui/app/helpers/async-escape-hatch.js
+++ b/ui/app/helpers/async-escape-hatch.js
@@ -1,0 +1,7 @@
+import Helper from '@ember/component/helper';
+
+export function asyncEscapeHatch([model, relationship]) {
+  return model[relationship].content;
+}
+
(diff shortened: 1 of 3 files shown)


philrenaud (approved):
Approved, but looks like there is an editor config change that may have been made on your end / noticing that line length changes are being made in unrelated parts of your changed file. Could you see if you could please undo those?

PR #15434, opened 2022-11-30
ui:  move "add more" button in variable create and edit forms

ChaiWithJai:
Resolves [#15345](https://github.com/hashicorp/nomad/issues/15345) and [#15346](https://github.com/hashicorp/nomad/issues/15346).

Users are accidentally adding more variables when the mean to save the variable because of the placement of "add more" button.

### Edit Form
image

### Create Form
image

### DIFF
--- a/ui/app/components/variable-form.hbs
+++ b/ui/app/components/variable-form.hbs
@@ -137,16 +137,17 @@
   {{/if}}
 
   <footer>
-    {{#if (eq this.view "table")}}
+    {{#unless (eq this.view "json")}}
     <button
       class="add-more button is-info is-inverted"
       type="button"
       disabled={{not (and this.keyValues.lastObject.key this.keyValues.lastObject.value)}}
       {{on "click" this.appendRow}}
+      data-test-add-kv
     >
       Add More
     </button>
-    {{/if}}
+    {{/unless}}
     <button
       disabled={{this.shouldDisableSave}}
       class="button is-primary save"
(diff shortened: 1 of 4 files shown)


juliezzhou:
looks good, just a small suggestion, for the space between "add more" and "save", can we keep it the same as the space between the input field and the delete button?

ui/app/components/variable-form.hbs
```diff
@@ -148,9 +137,20 @@
   {{/if}}
 
   <footer>
+    {{#unless (eq this.view "json")}}
```
philrenaud, line 140: There's an ember idiomatic style error here in order to prevent double-negative etc. template logic. Straightforward fix is to `@alias` `this.viewIsJson` to this equality.

ui/app/styles/components/variables.scss
```diff
@@ -184,3 +180,14 @@ table.variable-items {
     opacity: 1;
   }
 }
+
+footer {
+  .button.is-primary.save {
+    margin-left: 8px;
+  }
```
philrenaud, line 187: Adding margin to individual block-level elements within a row is harder to scale than setting things like spacing at the top level. Try this instead and use CSS grid (or flexbox if you're more comfortable with it):

```
footer {
    display: grid;
    grid-auto-columns: max-content;
    grid-auto-flow: column;
    gap: 1rem;
}
```

and you can remove the 8px margin from the button itself.

This has the added benefit of, if we add more buttons later, we don't have to apply new styles to them. Same for if we rename this class in the future.

philrenaud (approved):
Approved pending UI test failure and css grid change. Looks good!

PR #13976, opened 2022-08-03
[ui] Policies UI

philrenaud:
Adds a new section to the UI sidebar for Policies, restricted to Management token users.

This interface can be used to view, create, update or delete existing policies. With time, we can expand this to include token management and other ACL / role management for cluster users.

image

image

PR #15435, opened 2022-11-30
[ui] Token management interface on policy pages

philrenaud:
Adds the ability to view, delete, and create tokens associated with a given policy in the new [Policies UI](https://github.com/hashicorp/nomad/pull/13976).

Ability to do any of this is available only to `manager` tokens only.

image

### DIFF
--- a/ui/app/abilities/token.js
+++ b/ui/app/abilities/token.js
@@ -1,0 +1,10 @@
+import AbstractAbility from './abstract';
+import { alias } from '@ember/object/computed';
+
+export default class extends AbstractAbility {
+  @alias('selfTokenIsManagement') canRead;
+  @alias('selfTokenIsManagement') canList;
+  @alias('selfTokenIsManagement') canWrite;
+  @alias('selfTokenIsManagement') canUpdate;
+  @alias('selfTokenIsManagement') canDestroy;
+}
--- a/ui/app/components/policy-editor.js
+++ b/ui/app/components/policy-editor.js
@@ -25,9 +25,10 @@ export default class PolicyEditorComponent extends Component {
         throw {
           errors: [
             {
-              detail: 'Policy name must be 1-128 characters long and can only contain letters, numbers, and dashes.'
-            }
-          ]
+              detail:
+                'Policy name must be 1-128 characters long and can only contain letters, numbers, and dashes.',
+            },
+          ],
         };
       }
 
--- a/ui/app/components/policy-editor.js
+++ b/ui/app/components/policy-editor.js
@@ -2,7 +2,6 @@ import Component from '@glimmer/component';
 import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
 import { alias } from '@ember/object/computed';
-import messageForError from 'nomad-ui/utils/message-from-adapter-error';
 
 export default class PolicyEditorComponent extends Component {
   @service flashMessages;
@@ -22,14 +21,9 @@ export default class PolicyEditorComponent extends Component {
     try {
       const nameRegex = '^[a-zA-Z0-9-]{1,128}$';
       if (!this.policy.name?.match(nameRegex)) {
-        throw {
-          errors: [
-            {
-              detail:
-                'Policy name must be 1-128 characters long and can only contain letters, numbers, and dashes.',
-            },
-          ],
-        };
+        throw new Error(
+          `Policy name must be 1-128 characters long and can only contain letters, numbers, and dashes.`
+        );
       }
 
       const shouldRedirectAfterSave = this.policy.isNew;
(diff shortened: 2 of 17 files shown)


ui/mirage/factories/token.js
```diff
@@ -14,6 +14,7 @@ export default Factory.extend({
   oneTimeSecret: () => faker.random.uuid(),
 
   afterCreate(token, server) {
+    if (token.policyIds && token.policyIds.length) return;
```
philrenaud, line 17: Don't generate mirage policyIDs if we're explicitly creating them elsewhere

ui/app/styles/components/policies.scss
```diff
@@ -25,3 +25,48 @@ table.policies {
...
+    }
+  }
+}
+
+@media #{$mq-hidden-gutter} {
+  .token-operations {
+    grid-auto-flow: row;
+    grid-template-columns: 1fr;
+  }
+}
```
philrenaud, line 67: De-columnizes things if you're on a narrow screen
image

mikenomitch:
Feeling really good.

Small ignorable request: When you create an example token, do you think we could auto-copy to the clipboard? Seems marginally nicer than having the user copy from the alert.

Also, I think the policy name has to be wrapped in quotes in case they have a space. This command will fail for instance:
`nomad acl token create -name="<TOKEN_NAME>" -policy=back up -type=client -ttl=<8h>`

Should Policy Deleted and Token Deleted alerts be auto-removed? I see why the Example Token created one wouldn't be (since you have to copy), but it might be nice to remove the others after N seconds.

philrenaud:
@mikenomitch Added a "Copy to Clipboard" button and set timeout on that particular notification to 30sec.

Changed timeouts for token and policy deletion from 60sec (`¯\_(ツ)_/¯`) to 5sec

Added quotes to the example CLI command, thanks for spotting that!

ui/app/abilities/token.js
```diff
@@ -0,0 +1,10 @@
+import AbstractAbility from './abstract';
+import { alias } from '@ember/object/computed';
+
+export default class extends AbstractAbility {
+  @alias('selfTokenIsManagement') canRead;
+  @alias('selfTokenIsManagement') canList;
+  @alias('selfTokenIsManagement') canWrite;
+  @alias('selfTokenIsManagement') canUpdate;
+  @alias('selfTokenIsManagement') canDestroy;
```
ChaiWithJai, line 9: `question`: Shouldn't these permissions be based on multiple properties?

ui/app/components/policy-editor.js
```diff
@@ -22,11 +22,18 @@ export default class PolicyEditorComponent extends Component {
...
+          errors: [
+            {
+              detail:
+                'Policy name must be 1-128 characters long and can only contain letters, numbers, and dashes.',
+            },
+          ],
+        };
       }
 
+      const shouldRedirectAfterSave = this.policy.isNew;
```
ChaiWithJai, line 35: `question`: Looks we're shaping this error message to match the schema of an `AdapterError`. Are we just doing this because we don't want to update the utility function `messageError` that we have?

If `ember-data` updates the schema of the `AdapterError` then we'll have a pretty challenging situation to understand if we broke something.

Could we patch `messageError` to work with errors that don't stem from an HTTP response because I imagine we'll throw more errors in components because we've just recently started using flash messages?

ChaiWithJai (approved):

ui/app/abilities/token.js
philrenaud: Token access and operations will only ever be available to Management tokens. If we want to have some kind of `policiesSupportTokenList` functionality later, it can be added to this file, but for now it's managers only — same as policies.

ui/app/controllers/policies/policy.js
```diff
@@ -43,4 +55,69 @@ export default class PoliciesPolicyController extends Controller {
...
+    try {
+      const newToken = this.store.createRecord('token', {
+        name: `Example Token for ${this.policy.name}`,
+        policies: [this.policy],
+        // New date 10 minutes into the future
+        expirationTime: new Date(Date.now() + 10 * 60 * 1000),
+        type: 'client',
+      });
+      yield newToken.save();
+      this.refreshTokens();
```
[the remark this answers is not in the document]
philrenaud: Good catch; adding a yield!

ui/app/routes/policies/policy.js
```diff
@@ -2,15 +2,23 @@ import Route from '@ember/routing/route';
...
+      policy: this.store.findRecord('policy', decodeURIComponent(params.name), {
+        reload: true,
+      }),
+      tokens: this.store
+        .peekAll('token')
+        .filter((token) =>
+          token.policyNames?.includes(decodeURIComponent(params.name))
+        ),
     });
   }
```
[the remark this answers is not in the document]
philrenaud: I'm not opposed to using Promise.all, but it seems less readable and idiomatic for our models to have otherwise-unused consts that are destructured later. Are there good reasons to avoid rsvp.hash?

PR #15436, opened 2022-11-30
client: stop some taskrunner hooks when task exits

### LINKED ISSUE #15419: Prestart / Init task continues to render templates
- Nomad v1.4.3
- Ubuntu 22.04.1 LTS (Linux 5.15.0-53-generic)

When a given init/prestart task completes successfully, it appears that the any `template` values rendered by that task which contain Vault secrets are continually updated.

The expected behavior is that, once a task concludes, all templates associated with it are no longer updated.

Here's an example job file that demonstrates the issue:

```
job "my-job" {
  datacenters = ["default"]

  group "primary" {
    count = 1

    task "main" {
      driver = "exec"

      config {
        command = "${NOMAD_ALLOC_DIR}/usr/bin/sleep"
        args    = ["infinity"]
      }
    }

    task "init" {
      driver = "exec"

      lifecycle { hook = "prestart" } # init task, once it exits, the template shouldn't be re-rendered

      vault { policies = [ "vault-policy-here" ] }

      config {
        command = "${NOMAD_ALLOC_DIR}/usr/bin/echo" # arbitrary command; typically much more happening here
        args    = ["done"]
      }

      # new secret.txt file is re-rendered every few minutes even after completion of init (prestart) task
      template {
        change_mode = "noop"
        destination = "${NOMAD_SECRETS_DIR}/secret.txt"
        data        = <<-EOT
          {{- with secret "gcp/my-service-account/my-reader/token" -}}
          {{ .Data.token }}
          {{- end -}}
        EOT
      }
    }
  }
}
```

lgfa29:
Non-sidecar prestart and poststart tasks run to completion and exit, but their Stop hook are not called until the job is stopped. This leaves some background operations running even when the task itself is not running anymore.

Implementing the Exited hook stops these background processes on task exit. If the task is restarted their Prestart and Poststart hooks will run again and resume the background operations.

I skipped the Vault hook because token renewal has been quite finicky and trying to pause it could cause even more problems. Since it's just a goroutine that calls the Vault API not that frequently it doesn't consume much resources.

Closes #15419

### DIFF
--- a/client/allocrunner/taskrunner/logmon_hook_test.go
+++ b/client/allocrunner/taskrunner/logmon_hook_test.go
@@ -55,10 +55,10 @@ func TestTaskRunner_LogmonHook_LoadReattach(t *testing.T) {
 	require.Equal(t, orig, cfg)
 }
 
-// TestTaskRunner_LogmonHook_StartStop asserts that a new logmon is created the
-// first time Prestart is called, reattached to on subsequent restarts, and
-// killed on Stop.
-func TestTaskRunner_LogmonHook_StartStop(t *testing.T) {
+// TestTaskRunner_LogmonHook_StartRestartStop asserts that a new logmon is
+// created the first time Prestart is called, reattached to on subsequent
+// restarts, killed on Exited, restarted on restart, and killed on Stop.
+func TestTaskRunner_LogmonHook_StartRestartStop(t *testing.T) {
 	ci.Parallel(t)
 
 	alloc := mock.BatchAlloc()
@@ -93,6 +93,17 @@ func TestTaskRunner_LogmonHook_StartStop(t *testing.T) {
 	require.False(t, resp.Done)
 	origHookData = resp.State[logmonReattachKey]
 	require.Equal(t, origHookData, req.PreviousState[logmonReattachKey])
+
+	// Runnig exited should shutdown logmon
+	exitReq := interfaces.TaskExitedRequest{}
+	require.NoError(t, hook.Exited(context.Background(), &exitReq, nil))
+
+	// Restarting the task should run logmon again
+	req.PreviousState = map[string]string{
+		logmonReattachKey: origHookData,
+	}
+	require.NoError(t, hook.Prestart(context.Background(), &req, &resp))
+	require.False(t, resp.Done)
 
 	// Running stop should shutdown logmon
 	stopReq := interfaces.TaskStopRequest{
--- a/client/allocrunner/taskrunner/script_check_hook.go
+++ b/client/allocrunner/taskrunner/script_check_hook.go
@@ -151,6 +151,22 @@ func (h *scriptCheckHook) upsertChecks() error {
 	return nil
 }
 
+// Exited runs when the task exits, but it may run again on restart.
+// Stop any running scripts but don't shutdown the script check since we may
+// still need it.
+func (h *scriptCheckHook) Exited(_ context.Context, _ *interfaces.TaskExitedRequest, _ *interfaces.TaskStopResponse) error {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+
+	// Cancel all running scripts, but don't close the shutdownCh since the
+	// task may still be restarted.
+	for _, script := range h.runningScripts {
+		script.cancel()
+	}
+
+	return nil
+}
+
 // Stop implements interfaces.TaskStopHook and blocks waiting for running
 // scripts to finish (or for the shutdownWait timeout to expire).
 func (h *scriptCheckHook) Stop(ctx context.Context, req *interfaces.TaskStopRequest, resp *interfaces.TaskStopResponse) error {
--- a/client/allocrunner/taskrunner/template_hook.go
+++ b/client/allocrunner/taskrunner/template_hook.go
@@ -161,6 +161,25 @@ func (h *templateHook) newManager() (unblock chan struct{}, err error) {
 	return unblock, nil
 }
 
+// Exited runs when the task exits, but it may run again on restart.
+// Stop the template manager to prevent updates to templates of exited tasks.
+func (h *templateHook) Exited(ctx context.Context, req *interfaces.TaskExitedRequest, resp *interfaces.TaskExitedResponse) error {
+	h.managerLock.Lock()
+	defer h.managerLock.Unlock()
+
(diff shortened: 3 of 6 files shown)


client/allocrunner/taskrunner/logmon_hook_test.go
```diff
@@ -55,10 +55,10 @@ func TestTaskRunner_LogmonHook_LoadReattach(t *testing.T) {
 	require.Equal(t, orig, cfg)
 }
 
-// TestTaskRunner_LogmonHook_StartStop asserts that a new logmon is created the
-// first time Prestart is called, reattached to on subsequent restarts, and
-// killed on Stop.
-func TestTaskRunner_LogmonHook_StartStop(t *testing.T) {
+// TestTaskRunner_LogmonHook_StartRestartStop asserts that a new logmon is
+// created the first time Prestart is called, reattached to on subsequent
```
schmichael, line 59:
```go
// a new logmon is created the first time Prestart is called
```

To me `first time` implies that a single logmon process should be reused and reconnected to between restarts. Mind rewriting this comment to make sure it's unambiguous?

> TestTaskRunner_LogmonHook_StartRestartStop asserts that logmon's lifecycle matches the tasks: it is killed when the task exits and started (or restarted or reattached) whenever the task starts.

Read on for context if your bored :sweat_smile:

<details>
I think logmon was written such that the process lived for the lifetime of the *alloc,* not the lifetime of an individual task invocation.

Most hooks that are implemented that way do so by setting `response.Done = true` somewhere in Prestart so that TaskRunner never bothers calling it again. However logmon does not since it *does* need to *reconnect* to the running logmon process on task restart or restart logmon if the entire host was restarted and logmon no longer exists...

...and because we have to handle that `restart logmon if it did die` case, I think just killing it every time will work fine functionally!
</details>

client/allocrunner/taskrunner/logmon_hook_test.go
```diff
@@ -94,6 +94,17 @@ func TestTaskRunner_LogmonHook_StartStop(t *testing.T) {
 	origHookData = resp.State[logmonReattachKey]
 	require.Equal(t, origHookData, req.PreviousState[logmonReattachKey])
 
+	// Runnig exited should shutdown logmon
```
schmichael, line 97:
```suggestion
	// Running exited should shutdown logmon
```

client/allocrunner/taskrunner/script_check_hook.go
```diff
@@ -151,6 +151,22 @@ func (h *scriptCheckHook) upsertChecks() error {
...
+// still need it.
+func (h *scriptCheckHook) Exited(_ context.Context, _ *interfaces.TaskExitedRequest, _ *interfaces.TaskStopResponse) error {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+
+	// Cancel all running scripts, but don't close the shutdownCh since the
+	// task may still be restarted.
+	for _, script := range h.runningScripts {
+		script.cancel()
+	}
```
schmichael, line 165: This will cause script checks with `check.interval > restart.delay` to be run more frequently. I can't imagine why that would be a _bad thing_ since it just means we're heartbeating the TTL check more often, but I thought I'd mention it in case someone can dream up a way it could cause problems.

client/allocrunner/taskrunner/template_hook.go
```diff
@@ -161,6 +161,25 @@ func (h *templateHook) newManager() (unblock chan struct{}, err error) {
...
+func (h *templateHook) Exited(ctx context.Context, req *interfaces.TaskExitedRequest, resp *interfaces.TaskExitedResponse) error {
+	h.managerLock.Lock()
+	defer h.managerLock.Unlock()
+
+	// Shutdown any created template
+	if h.templateManager != nil {
+		h.templateManager.Stop()
+
+		// Set templateManager to nil so it's recreated by Prestart on restart.
+		h.templateManager = nil
```
schmichael, line 175: This will block the task from restarting until the template has been *re-rendered.*

I'm a little worried this could cause correlated failures if something like this happens:

- A template dependency is unreachable
- Task dies for an unrelated reason

Prior to this change the task would restart after `restart.delay`, use whatever templated files it used last, and get back to work.

After this change the task would be blocked from restarting until all dependencies were reachable.

To put it another way: prior to this change client agents could handle _local_ "scheduling" decisions while the servers were in charge of _cluster_ scheduling decisions. (I think that's broadly true? Artifacts and Vault for example both allow disconnected local operation as long as they were able to successfully run _once._)

I _think_ we need to maintain that behavior although I'd love to be convinced otherwise! :sweat_smile: 

A quick peek at `TaskTemplateManager` makes me think adding the ability to `Pause/Resume` it would be a significant effort. Adding a new parameter or other bit of plumbing to make `TaskTemplateManager.Start()` blocking on `handleFirstRender` _optional_ might be an easier route.

schmichael (changes requested):

client/allocrunner/taskrunner/template_hook.go
schmichael: A potentially interesting case to consider:

1. Task A happily talks to Service X
2. Service X crashes or is otherwise partitioned from A
3. Task A crashes due to Service X going away
4. Task A is down for `restart.delay` duration
5. Service Y is registered as a replacement for Service X
6. Task A restarts and happily talks to Service Y

I think the behavior of Task A in this situation is:

<table>
    <tr>
        <td></td>
        <td>During restart.delay</td>
        <td>After restart.delay</td>
    </tr>
    <tr>
        <td>Before change</td>
        <td>Immediate restart</td>
        <td>Crash loop until Y is discovered</td>
    </tr>
    <tr>
        <td>After change</td>
        <td colspan=2>Blocked until Y is discovered</td>
    </tr>
</table>

I think in the happy cases of Service Y being ~immediately available, your changes are optimal!

However there are some convoluted situations in which I think it could make an outage worse:

If the only way to discover Service Y was being _rescheduled_ we're better off with the existing crash-loop behavior as that will eventually result in a rescheduling of the whole alloc. That's a pretty complex outage situation I think: Nomad would have fine connectivity but ...Task A's DC would lose connectivity to Consul and need rescheduling in a new DC? idk... in that complex of a failure scenario there may be a number of other failure conditions I'm not thinking of that alter the behavior of Task A.

PR #15437, opened 2022-11-30
style: Remove "is-primary" class from TaskRow

### LINKED ISSUE #12881: task log is hidden/inaccessible/hard-to-find
Steps:
1) Set up a docker job for an image that doesn't exist (presumably other mistakes recreate this situation)
2) Attempt to run job
3) Try to find the real error

The error is hidden under this UI element:
Allocation_statsd-exporter_statsd-exporter_0__-_Nomad

which is this route:
/ui/allocations/4dce572b-6c85-b6b6-acfb-e502a893aa43/statsd-exporter

and it very plainly states: `Failed to pull 'prom/statsd_exporter:v0.22.4'` which is great.  The problem is before I found that I clicked around what I thought was every possible nomad page (and didn't find it) for a solid 10 minutes then ssh'd into the client and tailed /var/log/messages.

**Request:**
Is it possible to make this row more obviously clickable?

Saintpreston:
Resolves #12881 

The task row is clickable however since it was not formatted like a hyperlink users report that its not obvious that its clickable.

This PR removes the "is-primary" class on the task id which is overriding the default a tag styling. This PR follows the linking pattern in the HashiCorp Design System.

Here's the current appearance:
image

Here's how linking is in the HashiCorp Design System:
Screenshot 2022-11-30 at 6 22 57 PM

PS: I'd love an internship with the HashiCorp Design System team!

### DIFF
--- a/ui/app/templates/components/task-row.hbs
+++ b/ui/app/templates/components/task-row.hbs
@@ -13,7 +13,6 @@
   <LinkTo
     @route="allocations.allocation.task"
     @models={{array this.task.allocation this.task}}
-    class="is-primary"
   >
     {{this.task.name}}
     {{#if this.task.isConnectProxy}}


ChaiWithJai:
Hey @philrenaud and Nomad team! @Saintpreston is one of my star students. He's wanted to try to learn about the HashiCorp Design System because he'd like to intern with the Design Systems team so we pair programmed on this PR.

cc:  @didoo @MelSumner

didoo (approved):
I don't have a Nomad instance that I can use for testing this PR, but I _imagine_* removing this class solves the issue.

\* Looking at the code, probably is this CSS rule? https://github.com/hashicorp/nomad/blob/5bec70723deb05f92673c31b2275bcdd2fe50f6b/ui/app/styles/core/table.scss#L138-L142
in which case one thing I noticed is that it also adds a `font-weight: $weight-semibold` to the text, so by removing `is-primary` is also removing the bolder weight; is this an issue?

MelSumner:
+1 to what @didoo said. I'd look at removing line 139 in your table.scss file (which overrides the color) instead of removing the entire CSS class since that class also does other things. 

I would also note that a table row (tr element) is not eligible to be interactive; but you seem to have contained the link in a table cell (td element) which is the correct way to do it. Just something to keep in mind. 👍

philrenaud (approved):
Hi @Saintpreston and thanks for the PR! This looks good to me. Longer term, we might consider amending the `.is-primary` rule to have a `table td > .is-primary` application that maintains other styles (like the font weight / boldness), but this is a good change.

Regarding an earlier review comment:
> so by removing is-primary is also removing the bolder weight; is this an issue?

My feeling is that this is not going to be an issue. This increases the obviousness of the link in a way that the semibold was not itself doing. IMO should be fine :)

Thanks again!

ChaiWithJai (approved):
Great job! LGTM.

Saintpreston:
> I don't have a Nomad instance that I can use for testing this PR, but I _imagine_* removing this class solves the issue.
> 
> * Looking at the code, probably is this CSS rule?
> 
> https://github.com/hashicorp/nomad/blob/5bec70723deb05f92673c31b2275bcdd2fe50f6b/ui/app/styles/core/table.scss#L138-L142
> 
> 
> in which case one thing I noticed is that it also adds a `font-weight: $weight-semibold` to the text, so by removing `is-primary` is also removing the bolder weight; is this an issue?

Hey @didoo , Jai had a conversation with Josh and Julie on the design team in light of this PR. Their guidance was to remove the `text-decoration: none after having a meeting discussing this. Super grateful for your code review and excited to learn more about HashiCorp Design System.

PR #15446, opened 2022-12-01
fix enterprise endpoint registration

tgross:
In #15430 we refactored the RPC endpoint configuration to make adding the RPC context easier. But when implementing the change on the Enterprise side, I discovered that the registration of enterprise endpoints was being done incorrectly -- this doesn't show up on OSS because the registration is always a no-op here.


shoenig (approved):
:thinking:

tgross:
> 🤔

Yeah, registration for `net/rpc` takes an `any` and then looks for "suitable methods" to construct a RPC handler for via reflection. So there's very little type assertion that much of anything is actually wired up correctly until something tries to call it. Pretty awful.

PR #15451, opened 2022-12-01
remove most static RPC handlers

tgross:
Nomad server components that aren't in the `nomad` package like the deployment watcher and volume watcher need to make RPC calls but can't import the Server struct to do so because it creates a circular reference. These components have a "shim" object that gets populated to pass a "static" handler that has no RPC context.

Most RPC handlers are never used in this way, but during server setup we were constructing a set of static handlers for most RPC endpoints anyways. This is slightly wasteful but also confusing to developers who end up being encouraged to just copy what was being done for previous RPCs.

This changeset includes the following refactorings:
* Remove the static handlers field on the server
* Instead construct just the specific static handlers we need to pass into the deployment watcher and volume watcher.
* Remove the unnecessary static handler from heartbeater
* Update various tests to avoid needing the static endpoints and instead use an endpoint constructed on the spot.
* Make sure every atypical handler is documented as such, so that it's obvious what the happy path is when creating new endpoints.

Follow-up work will examine whether we can remove the RPCs from deployment watcher and volume watcher entirely, falling back to raft applies like node drainer does currently. This PR does leave a nil `*RPCContext` in deployment watcher and volume watcher constructors, which would be nice to remove for safety.

### DIFF
--- a/nomad/server.go
+++ b/nomad/server.go
@@ -153,10 +153,6 @@ type Server struct {
 	rpcTLS    *tls.Config
 	rpcCancel context.CancelFunc
 
-	// staticEndpoints is the set of static endpoints that can be reused across
-	// all RPC connections
-	staticEndpoints endpoints
-
 	// streamingRpcs is the registry holding our streaming RPC handlers.
 	streamingRpcs *structs.StreamingRpcRegistry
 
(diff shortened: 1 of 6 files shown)


nomad/server.go
```diff
@@ -1197,69 +1193,44 @@ func (s *Server) setupRPC(tlsWrap tlsutil.RegionWrapper) error {
...
-	server.Register(s.staticEndpoints.ClientCSI)
-	server.Register(s.staticEndpoints.FileSystem)
-	server.Register(s.staticEndpoints.Agent)
-
-	// Dynamic endpoints are endpoints that include the connection context and
-	// are created on each connection. Register all the dynamic endpoints with
-	// the RPC server.
+	// The endpoints are client RPCs and don't include a connection
+	// context. They also need to be registered as streaming endpoints in their
+	// register() methods.
```
tgross, line 1203: I strongly considered having the constructors register the streaming RPCs here because it would make this whole function straightforward, but having that as a side effect in the constructor seems terrible. And it would make it hard to use the constructors in tests (which we've done a few places).

shoenig (approved):

PR #15454, opened 2022-12-02
[ui] Add namespaces to exec window

### LINKED ISSUE #13921: [ui] job exec window refresh for jobs in non-default namespaces
### Nomad version
`Nomad v1.3.2 (bf602974112964e9691729f3f0716ff2bcdb3b44)`

Ubuntu 22.04 LTS on amd64

### Issue
In the web UI, when refreshing the `Exec` window, it shows a `Not Found` error.

### Reproduction steps
- Open the web UI
- Click `Exec` on an allocation that is in a non-default namespace (important!)
- Refresh the popup (using CMD+R, F5, ...)
- See a nice `Not Found` page.

Notice that F12 / Devtools reports an API request to `/v1/job/my-super-job-name` without the namespace in the URL, resulting in a 404 because that job does not exist in the defualt namespace. 

#### Expected Result
Not a 404. 

#### Actual Result
A 404. 

### Job file (if appropriate)

Any job file with a non-default namespace.

Shouldn't be important.

Shouldn't be important.

philrenaud:
Fixes an issue where namespaces weren't being respected in Exec window URLs upon refresh (they were on initial load, by virtue of having the full job object passed in via openExecUrl). 

image
image

Resolves #13921

### DIFF
--- a/ui/app/controllers/exec.js
+++ b/ui/app/controllers/exec.js
@@ -17,7 +17,7 @@ export default class ExecController extends Controller {
   @service system;
   @service token;
 
-  queryParams = ['allocation'];
+  queryParams = ['allocation', 'namespace'];
 
   @localStorageProperty('nomadExecCommand', '/bin/bash') command;
   socketOpen = false;
--- a/.changelog/15454.txt
+++ b/.changelog/15454.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fixed a bug where the exec window would not maintain namespace upon refresh
+```


ui/app/controllers/exec.js
```diff
@@ -17,7 +17,10 @@ export default class ExecController extends Controller {
   @service system;
   @service token;
 
-  queryParams = ['allocation'];
+  queryParams = [
+    'allocation',
+    'namespace'
```
ChaiWithJai, line 22: `thought`: `namespace` is a reserved keyword in Ember. We normally alias to `qpNamespace` as a result. However, since we have no logic that depends on this variable, we should be fine. But worth noting.

ChaiWithJai (approved):
LGTM

ui/app/controllers/exec.js
ChaiWithJai, line 22: `nit`: `prettier` is expecting this to be on one line.

PR #15455, opened 2022-12-02
scheduler: allow using device ID as attribute

lgfa29:
Devices are fingerprinted as groups of similar devices. This prevented specifying specific device by their ID in constraint and affinity rules.

This commit introduces the `${device.ids}` attribute that returns a comma separated list of IDs that are part of the device group. Users can then use the set operators to write rules.

Closes https://github.com/hashicorp/nomad-device-nvidia/issues/11


shoenig (approved):
LGTM!

PR #15458, opened 2022-12-02
[ui] Adding canary_tags the web UI

### LINKED ISSUE #13018: canary_tags are not shown in ui when instance is canary
### Nomad version
Output from `nomad version`

Nomad v1.2.6 (a6c6b475db5073e33885377b4a5c733e1161020c)

aws linux 

### Issue

Spin up a new canary instance with `canary_tags` defined -- I'd expect to see the canary_tags in the ui when navigating to the allocation.

### Reproduction steps

1. spin up canary instance with canary_tags defined2. 
2. navigate to instance deatils in nomad ui3. 
3. observe that tags are shown on page but nothing about the canary_tags

### Expected Result

I'd expect to see the consul `canary_tags` somewhere

#### Actual Result

I see only the regular consul `tags`

### Job file (if appropriate)

```
    service {
      name = "health-check-tool"
      port = "http"

      tags = [
        "traefik.enable=true",
        "traefik.http.services.health_check_tool.loadbalancer.healthcheck.path=/healthCheckAlwaysSuccessTraefik",
        "traefik.http.services.health_check_tool.loadbalancer.healthcheck.interval=10s",
        "traefik.http.routers.deploy_automation_tester.rule=Host(`${var.domain}`)",
        "traefik.http.routers.deploy_automation_tester.tls=true",
        "traefik.http.routers.deploy_automation_tester.entrypoints=https",
        "traefik.http.middlewares.strip_canary_header.headers.customrequestheaders.canary=",
        "traefik.http.routers.deploy_automation_tester.middlewares=strip_canary_header@consulcatalog"
      ]

      canary_tags = [
        "traefik.consulnamesuffix=`canary`",
        "traefik.http.services.health_check_tool_canary.loadbalancer.healthcheck.path=/healthCheckAlwaysSuccessTraefik",
        "traefik.http.services.health_check_tool.loadbalancer.healthcheck.interval=10s",
        "traefik.enable=true",
        "traefik.http.routers.deploy_automation_tester_canary.rule=Host(`${var.domain}`) && Headers(`canary`, `true`)",
        "traefik.http.routers.deploy_automation_tester_canary.tls=true",
        "traefik.http.routers.deploy_automation_tester_canary.entrypoints=https"
      ]

      connect {
        sidecar_service {}
        sidecar_task {
          shutdown_delay = "31s"
          kill_timeout   = "35s"
          resources {
            cpu    = 250 # MHz
            memory = 64  # MB
          }
        }
      }

      check {
        name     = "alive"
        type     = "http"
        path     = "/healthCheckAlwaysSuccessNomad"
        interval = "10s"
        timeout  = "2s"
      }
    }
  }
```

philrenaud:
Appending, with a custom style, to anyplace we show service tags:

image
image
image

Resolves #13018

### DIFF
--- a/ui/app/components/allocation-service-sidebar.hbs
+++ b/ui/app/components/allocation-service-sidebar.hbs
@@ -1,159 +1,159 @@
 <div
-	class="sidebar has-subnav service-sidebar {{if this.isSideBarOpen "open"}}"
-	{{on-click-outside
-		@fns.closeSidebar
-		capture=true
-	}}
+  class="sidebar has-subnav service-sidebar {{if this.isSideBarOpen "open"}}"
+  {{on-click-outside
+    @fns.closeSidebar
+    capture=true
+  }}
 >
-	{{#if @service}}
-		{{keyboard-commands this.keyCommands}}
-		<header class="detail-header">
-			<h1 class="title">
-				{{@service.name}}
-				{{#if (not-eq @service.provider "consul")}}
-					<span class="aggregate-status">
-						{{#if (eq this.aggregateStatus 'Unhealthy')}}
-							<FlightIcon @name="x-square-fill" @color="#c84034" />
-							Unhealthy
-						{{else}}
-							<FlightIcon @name="check-square-fill" @color="#25ba81" />
-							Healthy
-						{{/if}}
-					</span>
-				{{/if}}
-			</h1>
-			<button
-				data-test-close-service-sidebar
-				class="button is-borderless"
-				type="button"
-				{{on "click" @fns.closeSidebar}}
-			>
-				{{x-icon "cancel"}}
-			</button>
-		</header>
+  {{#if @service}}
+    {{keyboard-commands this.keyCommands}}
+    <header class="detail-header">
+      <h1 class="title">
+        {{@service.name}}
+        {{#if (not-eq @service.provider "consul")}}
+          <span class="aggregate-status">
+            {{#if (eq this.aggregateStatus 'Unhealthy')}}
+              <FlightIcon @name="x-square-fill" @color="#c84034" />
+              Unhealthy
+            {{else}}
+              <FlightIcon @name="check-square-fill" @color="#25ba81" />
+              Healthy
+            {{/if}}
+          </span>
+        {{/if}}
+      </h1>
+      <button
+        data-test-close-service-sidebar
+        class="button is-borderless"
+        type="button"
+        {{on "click" @fns.closeSidebar}}
+      >
+        {{x-icon "cancel"}}
+      </button>
+    </header>
 
-		<div class="boxed-section is-small">
-			<div
-				class="boxed-section-body inline-definitions"
-			>
-				<span class="label">
-					Service Details
-				</span>
+    <div class="boxed-section is-small">
+      <div
+        class="boxed-section-body inline-definitions"
+      >
+        <span class="label">
+          Service Details
+        </span>
 
-				<div>
-					{{#if @service.connect}}
-						<span class="pair">
-							<span class="term">
-								Connect
-							</span>
+        <div>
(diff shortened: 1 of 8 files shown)


ui/app/styles/components/services.scss
```diff
@@ -172,3 +172,21 @@ table.health-checks {
...
   }
 }
+
+$tagPadding: 0.75em;
+
+.tag {
+  &.canary {
+    overflow: hidden;
+    &:before {
+      content: "Canary";
```
philrenaud, line 182: Fun CSS trick to add copy to an element without adding it in the template itself.

ui/app/components/allocation-service-sidebar.hbs
```diff
@@ -94,6 +94,9 @@
               {{#each @service.tags as |tag|}}
                 <span class="tag is-service">{{tag}}</span>
               {{/each}}
+							{{#each @service.canary_tags as |tag|}}
+								<span class="tag canary is-service">{{tag}}</span>
+							{{/each}}
```
philrenaud, line 99: Rather than make these their own conditional table column, decided to append them to the end of the rest of the `tags` on a service, should they exist.

ui/app/styles/components/services.scss
```diff
@@ -172,3 +172,21 @@ table.health-checks {
...
+      background-color: $blue-light;
+      color: $black;
+      line-height: 1.5em;
+      margin-left: -$tagPadding;
+      margin-right: $tagPadding;
+      padding: 0 $tagPadding;
+      align-self: normal;
+    }
+  }
+}
```
lgfa29, line 192: nit-picking: would it be better to have this in [`tags.scss`](https://github.com/hashicorp/nomad/blob/main/ui/app/styles/core/tag.scss) and call the class `is-canary` to match others?

ui/app/components/allocation-service-sidebar.hbs
lgfa29, line 99: Tab alert 🚨 😄

ui/app/templates/components/job-service-row.hbs
```diff
@@ -35,7 +35,10 @@
 	</td>
   <td>
     {{#each @service.tags as |tag|}}
-			<span class="tag">{{tag}}</span>
+			<span class="tag is-service">{{tag}}</span>
+    {{/each}}
+    {{#each @service.canary_tags as |tag|}}
+			<span class="tag canary is-service">{{tag}}</span>
```
lgfa29, line 41: Tabs here as well

lgfa29 (approved):
Minor comments, but LGTM.

I wonder if it would be possible to apply the correct set of tags instead of always displaying them all depending on the allocation deployment status:

```console
$ nomad operator api /v1/allocations | jq '.[].DeploymentStatus'
{
  "Canary": false,
  "Healthy": true,
  "ModifyIndex": 24,
  "Timestamp": "2022-12-05T15:32:23.764038-03:00"
}
{
  "Canary": true,
  "Healthy": true,
  "ModifyIndex": 35,
  "Timestamp": "2022-12-05T15:33:13.774562-03:00"
}
```

So the `canary_tags` would only show for canary allocs, and the `tags` would show for non-canary.

philrenaud:
> I wonder if it would be possible to apply the correct set of tags instead of always displaying them all depending on the allocation deployment status:

@lgfa29 I was thinking about this, but figured it might be worthwhile to say something like "... and if this alloc is ever in a `canary` state, these are the tags it would have", as the canary state might be short-lived and fleeting.

lgfa29:
> as the canary state might be short-lived and fleeting.

I think the use of `canary_tags` is usually associated with manual promotion after validating the changes are correct, so I would imagine they are not as short-lived. But it would be a big lift to parse that API response, so just  something to maybe consider in the future 👍

PR #15473, opened 2022-12-05
networking: Inject implicit constraints on CNI plugins when using bridge mode

### LINKED ISSUE #14022: fingerprint: should be aware of available cointainernetworking plugin
To use `bridge` networking mode, one must first install containernetworking plugins into `/opt/cni/bin` (or where configured).

Nomad doesn't fingerprint the availability of these plugins, and you find out about their absence only when a task is set to run and crashloop on looking for them. Tracking down the problem ends up taking a number of steps:

Allocs dying:

```
Allocations
ID        Node ID   Task Group  Version  Desired  Status  Created    Modified
e28426a9  1d5216fa  fake        0        run      failed  1m57s ago  1m52s ago
ff79d1d6  1d5216fa  fake        0        stop     failed  3m ago     1m57s ago
b1da3290  1d5216fa  fake        0        stop     failed  3m34s ago  3m ago
```

No alloc logs:

```
➜ nomad alloc logs e2
Error reading file: Unexpected response code: 404 (task "faketask" not started yet. No logs available)
```

Client logs:

```
    2022-08-04T14:17:37.815-0500 [WARN]  client.alloc_runner.runner_hook: failed to configure network: alloc_id=87fe5f48-74a4-5afe-0308-5436d0fe961f err="plugin type=\"loopback\" failed (add): failed to find plugin \"loopback\" in path [/opt/cni/bin]" attempt=1
```

At this point, if you're lucky, you remember to lookup the instructions under `Consul Service Mesh` (?) to setup the plugins to make bridge networking work

https://www.nomadproject.io/docs/integrations/consul-connect#cni-plugins

shoenig:
This PR adds a job mutator which injects constraints on the job taskgroups
that make use of bridge networking. Creating a bridge network makes use of the
CNI plugins: bridge, firewall, host-local, loopback, and portmap. Starting
with Nomad 1.5 these plugins are fingerprinted on each node, and as such we
can ensure jobs are correctly scheduled only on nodes where they are available,
when needed.

Targets v1.5.0

Closes #14022

### DIFF
--- a/nomad/structs/job.go
+++ b/nomad/structs/job.go
@@ -136,7 +136,7 @@ func (j *Job) RequiredNUMA() set.Collection[string] {
 // RequiredBridgeNetwork identifies which task groups, if any, within the job
 // contain networks requesting bridge networking.
 func (j *Job) RequiredBridgeNetwork() set.Collection[string] {
-	result := set.New[string](10)
+	result := set.New[string](len(j.TaskGroups))
 	for _, tg := range j.TaskGroups {
 		if tg.Networks.Modes().Contains("bridge") {
 			result.Insert(tg.Name)
--- a/nomad/structs/job.go
+++ b/nomad/structs/job.go
@@ -132,3 +132,15 @@ func (j *Job) RequiredNUMA() set.Collection[string] {
 	}
 	return result
 }
+
+// RequiredBridgeNetwork identifies which task groups, if any, within the job
+// contain networks requesting bridge networking.
+func (j *Job) RequiredBridgeNetwork() set.Collection[string] {
+	result := set.New[string](10)
+	for _, tg := range j.TaskGroups {
+		if tg.Networks.Modes().Contains("bridge") {
+			result.Insert(tg.Name)
+		}
+	}
+	return result
+}
--- a/.changelog/15473.txt
+++ b/.changelog/15473.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+networking: Inject constraints on CNI plugins when using bridge networking
+```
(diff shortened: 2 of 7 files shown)


angrycub:
Looks good to me. Will be nice to let folks validate the plugin versions if they need to.

lgfa29 (approved):
Neat! It could be helpful to have a comment next to [`nomadCNIConfigTemplate`](https://github.com/hashicorp/nomad/blob/v1.4.3/client/allocrunner/networking_bridge_linux.go#L141) with a reminder to add or remove these implicit constraints if the template changes.

This also makes me thing about mixed version clusters 🤔 

If you happen to have clients with different `nomadCNIConfigTemplate`, these constraints may not necessarily be true since they are applied in the servers. But that's a very edge case. 

It does make me wonder if we could generate these constraints dynamically from `nomadCNIConfigTemplate` 😅

shoenig:
Kinda missed the boat on 1.7; pushing to 1.8

tgross:
I've rebased this PR on main, added handling of the early exit we introduced for NUMA, and added an upgrade note.

nomad/structs/job.go
```diff
@@ -132,3 +132,15 @@ func (j *Job) RequiredNUMA() set.Collection[string] {
 	}
 	return result
 }
+
+// RequiredBridgeNetwork identifies which task groups, if any, within the job
+// contain networks requesting bridge networking.
+func (j *Job) RequiredBridgeNetwork() set.Collection[string] {
+	result := set.New[string](10)
```
lgfa29, line 139:
```suggestion
	result := set.New[string](len(j.TaskGroups))
```

Not sure if it matters much 🤷

lgfa29 (approved):
LGTM! Maybe worth adding a note to the bridge config that these constraints may need to be updated if the bridge spec is modified.

https://github.com/hashicorp/nomad/blob/c3e7b13d54eaf868974a0a31a721d04ff97c024b/client/allocrunner/networking_bridge_linux.go#L150-L152

tgross:
> Maybe worth adding a note to the bridge config that these constraints may need to be updated if the bridge spec is modified.

Done!

PR #15475, opened 2022-12-05
acl: SSO auth methods API documentation

pkazmierczak:
This PR provides documentation for the ACL Auth Methods API endpoints. 

Relates to #13120

### DIFF
--- a/website/content/api-docs/acl/auth-methods.mdx
+++ b/website/content/api-docs/acl/auth-methods.mdx
@@ -39,7 +39,7 @@ The table below shows this endpoint's support for
   set to either "local" or "global"
 
 - `MaxTokenTTL` `(duration: <required>)` - Defines the maximum life of a token created
-  by this method. When set it will initialize the `ExpirationTime` field on all
+  by this method. When set, it will initialize the `ExpirationTime` field on all
   tokens to a value of `Token.CreateTime + AuthMethod.MaxTokenTTL`. This field is
   not persisted beyond its initial use. Can be specified in the form of `"60s"` or
   `"5m"` (i.e., 60 seconds or 5 minutes, respectively). 
@@ -107,7 +107,7 @@ The table below shows this endpoint's support for
       "http://example.com/last_name": "last_name"
     },
     "ListClaimMappings": {
-      "http://consul.com/groups": "groups"
+      "http://nomad.com/groups": "groups"
     }
   }
 }
@@ -149,7 +149,7 @@ $ curl \
             "http://example.com/last_name": "last_name"
         },
         "ListClaimMappings": {
-            "http://consul.com/groups": "groups"
+            "http://nomad.com/groups": "groups"
         }
     },
     "CreateTime": "0001-01-01T00:00:00Z",
--- a/website/content/api-docs/acl/auth-methods.mdx
+++ b/website/content/api-docs/acl/auth-methods.mdx
@@ -199,7 +199,8 @@ queries](/api-docs#blocking-queries) and [required ACLs](/api-docs#acls).
   set as default when running `nomad login` command.
 
 - `Config` `(ACLAuthMethodConfig: nil)` - The raw configuration to use for
-  the auth method.
+  the auth method. This parameter is part of the auth method configuration, not
+  specific to Nomad.
 
   - `OIDCDiscoveryURL` `(string: "")` - The OIDC Discovery URL, without
     any .well-known component (base path).
@@ -238,11 +239,29 @@ queries](/api-docs#blocking-queries) and [required ACLs](/api-docs#acls).
 
 ```json
 {
-    "name": "example-acl-auth-method",
-    "type": "OIDC",
-    "tokenlocality": "global",
-    "maxtokenttl": "1h0m0s",
-    "default": true
+  "Name": "example-acl-auth-method",
+  "Type": "OIDC",
+  "Tokenlocality": "global",
+  "Maxtokenttl": "1h0m0s",
+  "Default": true,
+  "Config": {
+    "OIDCDiscoveryURL": "https://my-corp-app-name.auth0.com/",
+    "OIDCClientID": "V1RPi2MYptMV1RPi2MYptMV1RPi2MYpt",
+    "OIDCClientSecret": "example-client-secret",
+    "BoundAudiences": [
+      "V1RPi2MYptMV1RPi2MYptMV1RPi2MYpt"
+    ],
+    "AllowedRedirectURIs": [
+      "http://localhost:4646/oidc/callback"
+    ],
+    "ClaimMappings": {
+      "http://example.com/first_name": "first_name",
+      "http://example.com/last_name": "last_name"
+    },
+    "ListClaimMappings": {
+      "http://nomad.com/groups": "groups"
+    }
+  }
 }
 ```
 
@@ -265,7 +284,24 @@ $ curl \
     "Type": "OIDC",
     "TokenLocality": "global",
     "Default": true,
-    "Config": null,
+    "Config": {
+        "OIDCDiscoveryURL": "https://my-corp-app-name.auth0.com/",
+        "OIDCClientID": "V1RPi2MYptMV1RPi2MYptMV1RPi2MYpt",
+        "OIDCClientSecret": "example-client-secret",
+        "BoundAudiences": [
+          "V1RPi2MYptMV1RPi2MYptMV1RPi2MYpt"
+        ],
+        "AllowedRedirectURIs": [
+          "http://localhost:4646/oidc/callback"
+        ],
+        "ClaimMappings": {
+          "http://example.com/first_name": "first_name",
+          "http://example.com/last_name": "last_name"
+        },
+        "ListClaimMappings": {
+          "http://nomad.com/groups": "groups"
+        }
+    }
     "CreateTime": "2022-12-08T11:04:43.46206Z",
     "ModifyTime": "2022-12-08T11:04:43.46206Z",
     "CreateIndex": 12,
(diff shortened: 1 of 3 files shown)


website/content/api-docs/acl/auth-methods.mdx
```diff
@@ -0,0 +1,392 @@
...
+
+- `Type` `(string: <required>)` - ACL Auth Role SSO identifier. Currently, the
+  only supported Type is "OIDC." 
+
+- `TokenLocality` `(string: <required>)` - Defines whether the ACL Auth Method
+  creates a local or global token when performing SSO login. This field must be
+  set to either "local" or "global"
+
+- `MaxTokenTTL` `(duration: <required>)` - Defines the maximum life of a token created
+  by this method. When set it will initialize the `ExpirationTime` field on all
```
jrasell, line 42:
```suggestion
  by this method. When set, it will initialize the `ExpirationTime` field on all
```

website/content/api-docs/acl/auth-methods.mdx
```diff
@@ -0,0 +1,392 @@
...
+    ],
+    "AllowedRedirectURIs": [
+      "http://localhost:4646/oidc/callback"
+    ],
+    "ClaimMappings": {
+      "http://example.com/first_name": "first_name",
+      "http://example.com/last_name": "last_name"
+    },
+    "ListClaimMappings": {
+      "http://consul.com/groups": "groups"
```
jrasell, line 110:
```suggestion
      "http://nomad.com/groups": "groups"
```

website/content/api-docs/acl/auth-methods.mdx
```diff
@@ -0,0 +1,392 @@
...
+  by this method. When set it will initialize the `ExpirationTime` field on all
+  tokens to a value of `Token.CreateTime + AuthMethod.MaxTokenTTL`. This field is
+  not persisted beyond its initial use. Can be specified in the form of `"60s"` or
+  `"5m"` (i.e., 60 seconds or 5 minutes, respectively). 
+
+- `Default` `(bool: false)` - Defines whether this ACL Auth Method is to be
+  set as default when running `nomad login` command. 
+
+- `Config` `(ACLAuthMethodConfig: <required>)` - The raw configuration to use for
+  the auth method.
```
jrasell, line 51: I wonder if it is worth expanding this to explain these are specific to the SSO provider, whereas the top level are specific to Nomad?

website/content/api-docs/acl/auth-methods.mdx
```diff
@@ -0,0 +1,392 @@
...
+
+  - `OIDCDiscoveryURL` `(string: <required>)` - The OIDC Discovery URL, without
+    any .well-known component (base path).
```
jrasell, line 54: Clarifying for myself, what is `.well-known`?

website/content/api-docs/acl/auth-methods.mdx
```diff
@@ -0,0 +1,392 @@
...
+            "http://localhost:4646/oidc/callback"
+        ],
+        "DiscoveryCaPem": null,
+        "SigningAlgs": null,
+        "ClaimMappings": {
+            "http://example.com/first_name": "first_name",
+            "http://example.com/last_name": "last_name"
+        },
+        "ListClaimMappings": {
+            "http://consul.com/groups": "groups"
```
jrasell, line 152:
```suggestion
            "http://nomad.com/groups": "groups"
```

website/content/api-docs/acl/auth-methods.mdx
```diff
@@ -0,0 +1,392 @@
...
+        }
+    },
+    "CreateTime": "0001-01-01T00:00:00Z",
+    "ModifyTime": "0001-01-01T00:00:00Z",
```
jrasell, line 156: I think it would be nice to have these as non-default time.Time values, to make it feel a little more real.

website/content/api-docs/acl/auth-methods.mdx
```diff
@@ -0,0 +1,392 @@
...
+| `POST` | `/acl/auth-method/:method_name` | `application/json` |
+
+The table below shows this endpoint's support for [blocking
+queries](/api-docs#blocking-queries) and [required ACLs](/api-docs#acls).
+
+| Blocking Queries | ACL Required |
+| ---------------- | ------------ |
+| `NO`             | `management` |
+
+### Parameters
```
jrasell, line 178: The comments I made in the create parameter section I believe also apply here.

website/content/api-docs/acl/auth-methods.mdx
```diff
@@ -0,0 +1,392 @@
...
+```json
+{
+    "MaxTokenTTL": "1h0m0s",
+    "Name": "example-acl-auth-method",
+    "Type": "OIDC",
+    "TokenLocality": "global",
+    "Default": true,
+    "Config": null,
+    "CreateTime": "0001-01-01T00:00:00Z",
+    "ModifyTime": "0001-01-01T00:00:00Z",
```
jrasell, line 269: I think it would be nice to have these as non-default time.Time values, to make it feel a little more real.

website/content/api-docs/acl/auth-methods.mdx
```diff
@@ -0,0 +1,392 @@
...
+    is singular (such as an attribute).
+
+    When mapped, the values in each list can be any of a number, string, or
+    boolean and will all be stringified when returned.
+
+  - `ListClaimMappings` `(map[string]string)` - Mappings of claims (key) will be
+    copied to a metadata field (value). Use this if the claim you are capturing is
+    list-like (such as groups).
+
+### Sample Payload
```
jrasell, line 236: Do we want to include a config in the example, so the output is not null?

website/content/api-docs/acl/auth-methods.mdx
```diff
@@ -0,0 +1,392 @@
...
+| `GET`  | `/acl/auth-methods` | `application/json` |
+
+The table below shows this endpoint's support for
+[blocking queries](/api-docs#blocking-queries),
+[consistency modes](/api-docs#consistency-modes) and
+[required ACLs](/api-docs#acls).
+
+| Blocking Queries | Consistency Modes | ACL Required                                                                                                                             |
+| ---------------- | ----------------- | ---------------------------------------------------------------------------------------------------------------------------------------- |
+| `YES`            | `all`             | `management` for all auth methods.<br />Output when given a non-management token will be limited to the auth methods on the token itself |
```
jrasell, line 291: I believe this is incorrect as we don't know which auth method created the token and provide this functionality.

website/content/api-docs/acl/auth-methods.mdx
```diff
@@ -0,0 +1,392 @@
...
+
+### Sample Response
+
+```json
+[
+    {
+        "CreateIndex": 12,
+        "Default": true,
+        "ModifyIndex": 32,
+        "Name": "example-acl-auth-method"
```
jrasell, line 309: I think we need to add type here also, based on the update made after this PR.

website/content/api-docs/acl/auth-methods.mdx
```diff
@@ -0,0 +1,392 @@
...
+| `GET`  | `/acl/auth-method/:method_name` | `application/json` |
+
+The table below shows this endpoint's support for
+[blocking queries](/api-docs#blocking-queries),
+[consistency modes](/api-docs#consistency-modes) and
+[required ACLs](/api-docs#acls).
+
+| Blocking Queries | Consistency Modes | ACL Required                              |
+| ---------------- | ----------------- | ----------------------------------------- |
+| `YES`            | `all`             | `management` or token with access to role |
```
jrasell, line 331: I believe this can only be read by management tokens.

website/content/api-docs/acl/auth-methods.mdx
```diff
@@ -0,0 +1,392 @@
...
+### Sample Response
+
+```json
+{
+    "MaxTokenTTL": "1h0m0s",
+    "Name": "example-acl-auth-method",
+    "Type": "OIDC",
+    "TokenLocality": "global",
+    "Default": true,
+    "Config": null,
```
jrasell, line 355: Do we want to include an example config, even if its generic data?

website/content/api-docs/acl/auth-methods.mdx
```diff
@@ -0,0 +1,392 @@
...
+    "CreateTime": "0001-01-01T00:00:00Z",
+    "ModifyTime": "0001-01-01T00:00:00Z",
```
jrasell, line 357: I think it would be nice to have these as non-default time.Time values, to make it feel a little more real.

jrasell (commented):
Looking good and thanks for taking this on @pkazmierczak; I have made some inline suggestions that I think might be useful. let me know your thoughts.

website/content/api-docs/acl/auth-methods.mdx
pkazmierczak: yikes

website/content/api-docs/acl/auth-methods.mdx
pkazmierczak: 😬

website/content/api-docs/acl/auth-methods.mdx
pkazmierczak: Yeah so I thought about writing something along the lines of: "config differs per auth method" but abandoned it since we only support 1 auth method for now. But yes, mentioning that top-level and auth-specific are different domains is a good idea! 👍

website/content/api-docs/acl/auth-methods.mdx
pkazmierczak: Fixed in 4c7a08eb3737d331051ded3a836a91637128340a

website/content/api-docs/acl/auth-methods.mdx
pkazmierczak: Right! we don't require an ACL for `List`, we just return stubs anyway.

jrasell (approved):
LGTM, thanks @pkazmierczak!

PR #15484, opened 2022-12-06
bugfix: make sure streaming endpoints are only registered once

### LINKED ISSUE #15474: Job Endpoint concurrent map writes
Had a CI fail due to (almost certainly) unrelated changes; concurrent map writes on the Job endpoint

@pkazmierczak 
@jrasell 

```
FAIL nomad
=== RUN   TestACLEndpoint_GetAuthMethods
=== PAUSE TestACLEndpoint_GetAuthMethods
FAIL nomad.TestACLEndpoint_GetAuthMethods (-1.00s)
=== RUN   TestClientEndpoint_UpdateAlloc_Evals_ByTrigger
=== PAUSE TestClientEndpoint_UpdateAlloc_Evals_ByTrigger
FAIL nomad.TestClientEndpoint_UpdateAlloc_Evals_ByTrigger (-1.00s)
=== RUN   TestPlanEndpoint_ApplyConcurrent
=== PAUSE TestPlanEndpoint_ApplyConcurrent
FAIL nomad.TestPlanEndpoint_ApplyConcurrent (-1.00s)
=== RUN   TestRPC_streamingRpcConn_goodMethod_TLS
fatal error: concurrent map writes

goroutine 91524 [running]:
github.com/hashicorp/nomad/nomad/structs.(*StreamingRpcRegistry).Register(...)
	/home/runner/work/nomad/nomad/nomad/structs/streaming_rpc.go:42
github.com/hashicorp/nomad/nomad.(*ClientAllocations).register(...)
	/home/runner/work/nomad/nomad/nomad/client_alloc_endpoint.go:32
github.com/hashicorp/nomad/nomad.(*Server).setupRpcServer(0xc0046aedc0, 0xc005f65880?, 0xc0050bb220)
	/home/runner/work/nomad/nomad/nomad/server.go:1206 +0x16b
github.com/hashicorp/nomad/nomad.(*rpcHandler).handleMultiplexV2(0xc005bfd280, {0x29e92a0?, 0xc0066e5440}, {0x29f51f0?, 0xc005f65880}, 0xc0050bb220)
	/home/runner/work/nomad/nomad/nomad/rpc.go:494 +0x225
github.com/hashicorp/nomad/nomad.(*rpcHandler).handleConn(0xc005bfd280, {0x29e92a0, 0xc0066e5440}, {0x29f51f0, 0xc005f65880?}, 0xc0050bb220)
	/home/runner/work/nomad/nomad/nomad/rpc.go:352 +0xe45
github.com/hashicorp/nomad/nomad.(*rpcHandler).handleConn(0xc005bfd280, {0x29e92a0, 0xc0066e5440}, {0x29f74a8, 0xc005d9b148?}, 0xc0050bb220)
	/home/runner/work/nomad/nomad/nomad/rpc.go:332 +0x9bf
created by github.com/hashicorp/nomad/nomad.(*rpcHandler).listen
	/home/runner/work/nomad/nomad/nomad/rpc.go:183 +0x452
```

pkazmierczak:
Streaming RPCs should only be registered once, not on every RPC call, because they set keys in [`StreamingRpcRegistry.registry` map](https://github.com/hashicorp/nomad/blob/main/nomad/structs/streaming_rpc.go#L42). This PR fixes it by checking whether endpoints are already registered before calling `.register()` method. Fixes #15474

### DIFF
--- a/nomad/server.go
+++ b/nomad/server.go
@@ -1201,28 +1201,52 @@ func (s *Server) setupRpcServer(server *rpc.Server, ctx *RPCContext) error {
 	// The endpoints are client RPCs and don't include a connection
 	// context. They also need to be registered as streaming endpoints in their
 	// register() methods.
+	//
+	// Streaming RPCs should only be registered once, so before we register
+	// them, we make a check if they haven't been already registered.
 
 	clientAllocs := NewClientAllocationsEndpoint(s)
-	clientAllocs.register()
-	_ = server.Register(clientAllocs)
-
 	fsEndpoint := NewFileSystemEndpoint(s)
-	fsEndpoint.register()
-	_ = server.Register(fsEndpoint)
-
 	agentEndpoint := NewAgentEndpoint(s)
-	agentEndpoint.register()
-	_ = server.Register(agentEndpoint)
 
 	// Event is a streaming-only endpoint so we don't want to register it as a
 	// normal RPC
 	eventEndpoint := NewEventEndpoint(s)
-	eventEndpoint.register()
 
 	// Operator takes a RPC context but also has a streaming RPC that needs to
 	// be registered
 	operatorEndpoint := NewOperatorEndpoint(s, ctx)
-	operatorEndpoint.register()
+
+	// streamingEndpoint is a helper interface to make working with endpoint
+	// structs easier
+	type streamingEndpoint interface {
+		register()
+	}
+
+	streaming_endpoints := map[string]streamingEndpoint{
+		"Allocations.Exec": clientAllocs,
+		"Agent.Monitor":    agentEndpoint,
+		// fsEndpoint registers "FileSystem.Logs" and "FileSystem.Stream", but
+		// if at least one of them is registered we know the method's been
+		// called
+		"FileSystem.Logs": fsEndpoint,
+		"Event.Stream":    eventEndpoint,
+		// operatorEndpoint registers "Operator.SnapshotSave" and
+		// "Operator.SnapshotRestore"
+		"Operator.SnapshotSave": operatorEndpoint,
+	}
+
+	// If we haven't registered the streaming endpoints before, register them
+	// now.
+	for method, endpoint := range streaming_endpoints {
+		if _, err := s.streamingRpcs.GetHandler(method); err != nil {
+			endpoint.register()
+		}
+	}
+
+	_ = server.Register(clientAllocs)
+	_ = server.Register(fsEndpoint)
+	_ = server.Register(agentEndpoint)
 	_ = server.Register(NewOperatorEndpoint(s, ctx))
 
 	// These endpoints are client RPCs and don't include a connection context
--- a/nomad/server.go
+++ b/nomad/server.go
@@ -1132,10 +1132,10 @@ func (s *Server) setupVaultClient() error {
 // setupRPC is used to setup the RPC listener
 func (s *Server) setupRPC(tlsWrap tlsutil.RegionWrapper) error {
 	// Populate the static RPC server
-	err := s.setupRpcServer(s.rpcServer, nil)
-	if err != nil {
-		return err
-	}
+	s.setupRpcServer(s.rpcServer, nil)
+
+	// Setup streaming endpoints
+	s.setupStreamingEndpoints(s.rpcServer, nil)
 
 	listener, err := s.createRPCListener()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


nomad/server.go
```diff
@@ -1201,28 +1201,52 @@ func (s *Server) setupRpcServer(server *rpc.Server, ctx *RPCContext) error {
...
 	operatorEndpoint := NewOperatorEndpoint(s, ctx)
-	operatorEndpoint.register()
+
+	// streamingEndpoint is a helper interface to make working with endpoint
+	// structs easier
+	type streamingEndpoint interface {
+		register()
+	}
+
+	streaming_endpoints := map[string]streamingEndpoint{
```
tgross, line 1226: Hm, this ends up having to allocate this map every time we connect. Trying to test this without allocation seems like we'd need a lock or an atomic, which would be nice to avoid if we can.

What if we move the streaming registration out of this function entirely? This `setupRpcServer` gets called by both `setupRPC` (which is called once) and for each connection. That is, we can move the streaming endpoint registration into `setupRPC` (or into a new function it calls) and we know it'll only ever get called once.

tgross, line 1226: Hm, this ends up having to allocate this map every time we connect. Trying to test this without allocation seems like we'd need a lock or an atomic, which would be nice to avoid if we can.

What if we move the streaming registration out of this function entirely? This `setupRpcServer` gets called by both `setupRPC` (which is called once) and for each connection. That is, we can move the streaming endpoint registration into `setupRPC` (or into a new function it calls) and we know it'll only ever get called once.

tgross (changes requested):

nomad/server.go
pkazmierczak: I think I understand better now, thanks. Refactored.

nomad/server.go
```diff
@@ -1193,62 +1193,41 @@ func (s *Server) setupRPC(tlsWrap tlsutil.RegionWrapper) error {
...
 }
 
-// setupRpcServer is used to populate an RPC server with endpoints. This gets
-// called at startup but also once for every new RPC connection so that RPC
-// handlers can have per-connection context.
-func (s *Server) setupRpcServer(server *rpc.Server, ctx *RPCContext) error {
-
+// setupStreamingEndpoints is used to populate an RPC server with streaming
+// endpoints. This only gets called at server startup.
+func (s *Server) setupStreamingEndpoints(server *rpc.Server, ctx *RPCContext) {
```
tgross, line 1198: This method never gets a `*RPContext`, so we can eliminate that parameter.

nomad/server.go
```diff
@@ -1193,62 +1193,41 @@ func (s *Server) setupRPC(tlsWrap tlsutil.RegionWrapper) error {
...
-		if _, err := s.streamingRpcs.GetHandler(method); err != nil {
-			endpoint.register()
-		}
-	}
-
-	_ = server.Register(clientAllocs)
-	_ = server.Register(fsEndpoint)
-	_ = server.Register(agentEndpoint)
+	operatorEndpoint.register()
 	_ = server.Register(NewOperatorEndpoint(s, ctx))
```
tgross, line 1224: The client allocs, FS, agent, and operator endpoints all have non-streaming RPCs as well, so we should move these `_ = server.Register(..)` method calls into `setupRpcServer`. It looks like the event endpoint is the only one without a non-streaming method.

tgross (changes requested):

nomad/server.go
pkazmierczak: 🤦‍♂️ right...

nomad/server.go
pkazmierczak: Good point.

nomad/server.go
```diff
@@ -1221,14 +1216,26 @@ func (s *Server) setupRpcServer(server *rpc.Server, ctx *RPCContext) error {
...
 	_ = server.Register(NewClientCSIEndpoint(s))
 	_ = server.Register(NewClientStatsEndpoint(s))
 
+	// These endpoints have their streaming component registered in
+	// setupStreamingEndpoints, but their non-streaming RPCs are registered
+	// here.
+	_ = server.Register(NewClientAllocationsEndpoint(s))
+	_ = server.Register(NewFileSystemEndpoint(s))
+	_ = server.Register(NewAgentEndpoint(s))
+	_ = server.Register(NewOperatorEndpoint(s, nil))
```
tgross, line 1237:
```suggestion
	_ = server.Register(NewOperatorEndpoint(s, ctx))
```

tgross (approved):
I left one last item to make sure the operator endpoint has a context, but once that's included LGTM! Thanks for fixing this!

nomad/server.go
pkazmierczak: yikes, thanks for catching.

PR #15492, opened 2022-12-08
acl: canonicalize ACL Auth Method object

pkazmierczak:
Relates to #13120

### DIFF
--- a/nomad/structs/acl.go
+++ b/nomad/structs/acl.go
@@ -710,6 +710,16 @@ func (a *ACLAuthMethod) Copy() *ACLAuthMethod {
 	return c
 }
 
+// Canonicalize performs basic canonicalization on the ACL auth method object.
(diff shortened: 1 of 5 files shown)


nomad/structs/acl.go
```diff
@@ -710,6 +710,16 @@ func (a *ACLAuthMethod) Copy() *ACLAuthMethod {
 	return c
 }
 
+// Canonicalize performs basic canonicalization on the ACL auth method object.
+func (a *ACLAuthMethod) Canonicalize() {
+	t := time.Now().UTC()
+
+	if a.CreateTime.IsZero() {
+		a.CreateTime = t
+		a.ModifyTime = t
```
jrasell, line 719: I think we will always want to update the modify time. We call this whenever an auth method is being upserted, so it will always be a modification to the state.

jrasell (approved):
LGTM!

Issue #15493, opened 2022-12-08
Event Streams: A missing ACL policy on token will cause GET /v1/event/stream to fail


wjnicholson:
### Nomad version
Nomad v1.4.1 (2aa7e66bdb526e25f59883952d74dad7ea9a014e)

### Operating system and Environment details

Tested using curl and a nomad server in AWS.

### Issue

Minting a token with a policy in the list that doesn't exist causes GET requests to /v1/event/stream to return a 500 permission denied response, even if there is another policy that should allow the request to pass.

### Reproduction steps

Create token with two policies, one that exists with a default write permission and then another policy name that doesn't exist.
i.e
```
namespace "default" {
  policy = "write"
}
```

Call 
```
curl -vv -sN --header "X-Nomad-Token: ${NOMAD_TOKEN}" "${NOMAD_ADDR}/v1/event/stream?namespace=default&topic=Allocation"
```

#### Expected Result

A 200 response and an event stream.

#### Actual Result

500 response, with body "Permission denied"

Discovered issue by creating tokens with different policy subsets until we accidentally found it was due to a policy that didn't exist.
Mentioned here that it was possible to mint a token without a corresponding policy:
https://github.com/hashicorp/nomad/issues/14058#issuecomment-1342595062

I believe the code at https://github.com/hashicorp/nomad/blob/main/nomad/stream/event_broker.go#L302 is the issue, in method aclObjFromSnapshotForTokenSecretID. If it cannot find a policy it eagerly returns an error. Based on other instances of that method call (e.g https://github.com/hashicorp/nomad/blob/47c2d4ab341a7c8dc1f73e4ca77c2811d1aee02b/nomad/acl.go#L248) it looks like the code should be changed to continue:

```go
		if err != nil {
			return nil, err
		}
		if policy == nil {
			// Ignore policies that don't exist, since they don't grant any
			// more privilege.
			continue
		}
```

Separately - we found it hard to debug this as we could not see the missing policy in `ui/settings/tokens`, as the request for the policy document had failed and we hadn't opened dev tools to check for any errors.

tgross:
Leaving a note here about the open discussion on the PR: https://github.com/hashicorp/nomad/pull/15495#pullrequestreview-1295962292

Issue #14058, opened 2022-08-09
acl: an ACL token can be created referencing a non-existent ACL policy


jrasell:
### Nomad version
`Nomad v1.3.3 (428b2cd8014c48ee9eae23f02712b7219da16d30)`

### Operating system and Environment details
Tested locally on macOS using a development agent.

### Issue
A Nomad ACL client token can be created referencing a policy that does not exist. This could be a typo or other error which is hard to identify when the create request succeeds without problem. Using the subsequent token results in a permission denied error.

### Reproduction steps
Run a development agent using the following config snippet to ensure ACLs are enabled:
```
acl {
  enabled = true
}
```

Bootstrap the ACL system via the `nomad acl boostrap` command.

Create an ACL token referencing a non-existent ACL policy using the `nomad acl token create -policy=this-doesnt-exist -token=<ROOT_TOKEN>` command.

#### Expected Result
The token creation should fail and explain that the ACL policy does not exist.

#### Actual Result
The token is created successfully.

wjnicholson:
Hi 
> Empty (or absent) policies are treated as a no-op, but allow for policies to be applied to a live token.

This doesn't seem to be the case for event streams. We had a situation with a token with two policies attached, the first with 
```
namespace "will" {
  policy = "write"
}
``` 
and a second with a non-existent policy.

Calling
`curl -vv -sN --header "X-Nomad-Token: ${NOMAD_TOKEN}" "${NOMAD_ADDR}/v1/event/stream?namespace=will&topic=Allocation&topic=Job"`

We expected the call to succeed, but we get back a 500 permission denied error.
Upon recreating a token without the non-existent policy we receive 200 and the event stream.

I believe the code at https://github.com/hashicorp/nomad/blob/main/nomad/stream/event_broker.go#L302 is the issue, in method aclObjFromSnapshotForTokenSecretID. If we cannot find a policy we error eagerly.

jrasell:
Hi @wjnicholson, could you raise this as a separate issue? Although related, the problem you mention seems different from the discussion here and is worthy of its own issue as the behaviour certainly seems wrong.

wjnicholson:
Thanks @jrasell, raised https://github.com/hashicorp/nomad/issues/15493

PR #15495, opened 2022-12-08
bug: Handle missing policy documents in event streams

wjnicholson:
Fixes https://github.com/hashicorp/nomad/issues/15493

Will need a test too.

### DIFF
--- a/nomad/stream/event_broker.go
+++ b/nomad/stream/event_broker.go
@@ -301,8 +301,13 @@ func aclObjFromSnapshotForTokenSecretID(
 
 	for _, policyName := range aclToken.Policies {
 		policy, err := aclSnapshot.ACLPolicyByName(nil, policyName)
-		if err != nil || policy == nil {
+		if err != nil {
 			return nil, nil, errors.New("error finding acl policy")
+		}
+		if policy == nil {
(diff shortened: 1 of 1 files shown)


tgross (commented):
Hi @wjnicholson! This approach seems reasonable.

But it seems from the discussion in https://github.com/hashicorp/nomad/issues/14058 that we might not necessarily want to keep the existing behavior of this on other endpoints, in which case this PR would be adding work we want to rip out later. I'm going to ping @jrasell on this, as he's recently worked on the Roles feature and may have more context than me.

jrasell:
> Hi @wjnicholson! This approach seems reasonable.
> 
> But it seems from the discussion in #14058 that we might not necessarily want to keep the existing behavior of this on other endpoints, in which case this PR would be adding work we want to rip out later. I'm going to ping @jrasell on this, as he's recently worked on the Roles feature and may have more context than me.

Apologies for the delay. I think there is still some discussion to have on the issue and any change would come in an upcoming major release, and therefore not in the immediate term. I'm inclined to therefore say we should accept this change, which brings the stream inline with how policies work in the RPC and client caching. Any work on changing the existing behaviour would involve multiple subsystems, so this change I don't think adds too much more and the advantage of consistency seems like a good thing.

PR #15504, opened 2022-12-08
acl: make sure there is only one default Auth Method per type

pkazmierczak:
This PR adds a check that makes sure we don't insert a duplicate `default` ACL auth method for a given type. 
Relates to #13120


jrasell (commented):
Looks great and a lovely approach. QQ. do we also want the check we have in the RPC handler within the upsert state method to account for Raft latency?

pkazmierczak:
> Looks great and a lovely approach. QQ. do we also want the check we have in the RPC handler within the upsert state method to account for Raft latency?

Good call, I updated in 126a117

jrasell (approved):
LGTM!

PR #15510, opened 2022-12-09
qemu: add binary_path plugin config parameter

### LINKED ISSUE #4684: Qemu driver should allow to specify architecture (hard-coded qemu-system-x86_64)
Currently hard-coded `qemu-system-x86_64` severely undermines usefulness of Qemu driver.
It would be useful to be able to run _arm_ images on _x86_64_ architectures through `qemu-system-arm` executable, etc.

Please allow to specify qemu architecture in config.

schmichael:
Fixes #4684

This contains 3 changes to the qemu driver:

1. Adds a `binary_path` plugin configuration parameter for specifying alternative binary locations for qemu.
2. Defaults `binary_path = "qemu-system-ARCH"
3. Replaces ARCH in `binary_path` with the Nomad binary's GOARCH.

This should allow qemu to Just Work on non-x86_64 architectures like Apple's M1/M2 or MIPS if... I honestly don't know why you'd be using this on MIPS but it's a target both Go and QEMU support!

The `ARCH` replacement is ugly, and I had planned on using standard `${ARCH}` replacement. However that causes it to be handled by our HCL parser which fails with:

```
==> Error starting agent: failed to create plugin loader: failed to initialize plugin loader: parsing plugin configurations failed: 1 error occurred:
        * plugin "qemu" (driver): failed to parse config:  Variables not allowed: Variables may not be used here.
```

This is returned from `hclutils.ParseHclInterface`, and I'm not sure it's worth fighting that battle.


drivers/qemu/driver.go
```diff
@@ -234,7 +254,13 @@ func (d *Driver) buildFingerprint() *drivers.Fingerprint {
 		HealthDescription: drivers.DriverHealthy,
 	}
 
-	bin := "qemu-system-x86_64"
+	bin, err := resolveBinPath(d.binPath)
+	if err != nil {
+		fingerprint.Health = drivers.HealthStateUndetected
+		fingerprint.HealthDescription = err.Error()
```
schmichael, line 260: This makes the CLI `node status -self -verbose` output *really* wide. Should I shorten it to `QEMU binary not found` and just log the full error? Including the path makes it arbitrarily wide but makes debugging way easier than digging through client logs.

Example output with a *short* binary name:
```
Drivers
Driver       Detected  Healthy  Message                                                                    Time
docker       false     false    Failed to connect to docker daemon                                         2022-12-09T10:06:19-08:00
exec         false     false    Driver must run as root                                                    2022-12-09T10:06:19-08:00
java         false     false    Driver must run as root                                                    2022-12-09T10:06:19-08:00
mock_driver  true      true     Healthy                                                                    2022-12-09T10:06:19-08:00
qemu         false     false    unable to use binary_path="asdf-ARCH": executable file not found in $PATH  2022-12-09T10:06:19-08:00
raw_exec     true      true     Healthy                                                                    2022-12-09T10:06:19-08:00
```

shoenig: If `binary_path` is unset does it retain the `<none>` message? The verbosity when set seems fine - in that case you set bad config and should feel bad

schmichael: Oh that's a fantastic idea. I'll implement that, thanks.

tgross:
This was superceded by @mismithhisler's work in https://github.com/hashicorp/nomad/pull/27128 https://github.com/hashicorp/nomad/pull/27182, so I'm going to close this PR.

PR #15513, opened 2022-12-09
implement pre-forwarding auth on select RPCs

tgross:
In #15417 we added a new `Authenticate` method to the server that returns an `AuthenticatedIdentity` struct. This changeset implements this method for a small number of RPC endpoints that together represent all the various ways in which RPCs are sent, so that we can validate that we're happy with this approach.

A follow-up PR will do all the dirty work of adding this everywhere else.

### DIFF
--- a/client/rpc.go
+++ b/client/rpc.go
@@ -435,7 +435,7 @@ func resolveServer(s string) (net.Addr, error) {
 // a potential error.
 func (c *Client) Ping(srv net.Addr) error {
 	var reply struct{}
-	err := c.connPool.RPC(c.Region(), srv, "Status.Ping", struct{}{}, &reply)
+	err := c.connPool.RPC(c.Region(), srv, "Status.Ping", &structs.GenericRequest{}, &reply)
 	return err
 }
 
--- a/client/rpc.go
+++ b/client/rpc.go
@@ -431,11 +431,15 @@ func resolveServer(s string) (net.Addr, error) {
 	return net.ResolveTCPAddr("tcp", net.JoinHostPort(host, port))
 }
 
+// Ping never mutates the request, so reuse a singleton to avoid the extra
+// malloc
+var pingRequest = &structs.GenericRequest{}
+
 // Ping is used to ping a particular server and returns whether it is healthy or
 // a potential error.
 func (c *Client) Ping(srv net.Addr) error {
 	var reply struct{}
-	err := c.connPool.RPC(c.Region(), srv, "Status.Ping", &structs.GenericRequest{}, &reply)
+	err := c.connPool.RPC(c.Region(), srv, "Status.Ping", pingRequest, &reply)
 	return err
 }
 
--- a/nomad/node_endpoint.go
+++ b/nomad/node_endpoint.go
@@ -90,6 +90,10 @@ func NewNodeEndpoint(srv *Server, ctx *RPCContext) *Node {
 
 // Register is used to upsert a client that is available for scheduling
 func (n *Node) Register(args *structs.NodeRegisterRequest, reply *structs.NodeUpdateResponse) error {
+	// note that we trust-on-first use and the identity will be anonymous for
+	// that initial request; we lean on mTLS for handling that safely
+	authErr := n.srv.Authenticate(n.ctx, args)
+
 	isForwarded := args.IsForwarded()
 	if done, err := n.srv.forward("Node.Register", args, args, reply); done {
 		// We have a valid node connection since there is no error from the
@@ -102,6 +106,10 @@ func (n *Node) Register(args *structs.NodeRegisterRequest, reply *structs.NodeUp
 
 		return err
 	}
+	if authErr != nil {
+		return structs.ErrPermissionDenied
+	}
+
 	defer metrics.MeasureSince([]string{"nomad", "client", "register"}, time.Now())
 
 	// Validate the arguments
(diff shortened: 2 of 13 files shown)


nomad/node_endpoint.go
```diff
@@ -435,6 +444,9 @@ func (n *Node) deregister(args *structs.NodeBatchDeregisterRequest,
 
 // UpdateStatus is used to update the status of a client node
 func (n *Node) UpdateStatus(args *structs.NodeUpdateStatusRequest, reply *structs.NodeUpdateResponse) error {
+	identity, authErr := n.srv.Authenticate(n.ctx, args.AuthToken)
+	args.SetIdentity(identity)
```
schmichael, line 448: Instead of having to remember to call `args.SetIdentity` after every call to `Authenticate`, what if made `Server.Authenticate` (or a new wrapper around `Authenticate`) take an interface that could both retrieve the auth token and set the identity?

This is just a really non-idiomatic pattern to expect devs to get right in every RPC: both understanding when to call SetIdentity (always?) as well as knowing where exactly to check `authErr` (*after* forwarding?).

schmichael: An alternative might be to actually update our ancient `Server.forward` method (or make a new method to migrate RPCs piecemeal) to just encapsulate *all* of this. This would reduce the per-RPC boilerplate as well as only have to write the slightly unusual "check error *after* a subsequent operation" authErr handling.

nomad/status_endpoint.go
```diff
@@ -22,7 +22,11 @@ func NewStatusEndpoint(srv *Server, ctx *RPCContext) *Status {
 }
 
 // Ping is used to just check for connectivity
-func (s *Status) Ping(args struct{}, reply *struct{}) error {
+func (s *Status) Ping(args structs.GenericRequest, reply *struct{}) error {
+	// note: we're intentionally throwing away any auth error here
+	identity, _ := s.srv.Authenticate(s.ctx, args.AuthToken)
+	args.SetIdentity(identity)
```
schmichael, line 28: Why do we call this if we don't use either of the results? Might want to update the comment to explain why it's called at all instead of just pointing out that the `_` is intentional

nomad/acl.go
```diff
@@ -118,7 +118,7 @@ func (s *Server) Authenticate(ctx *RPCContext, secretID string) (*structs.Authen
 }
 
 func (s *Server) ResolveACL(aclToken *structs.ACLToken) (*acl.ACL, error) {
-	if !s.config.ACLEnabled {
+	if !s.config.ACLEnabled || aclToken == nil {
```
schmichael, line 121: This line is a good candidate for a helper func as well as (1) I'd be surprised if accessing `s.config` *wasn't* a race condition and therefore needs locking added someday, and (2) really easy to get a single character wrong in this line and cause a hard to spot security issue.

It's a well trod path though, so not a blocker this work.

schmichael: I was just in `Server.Reload` and it looks like `s.config` isn't swapped and `s.config.ACLEnabled` isn't touched, so this probably isn't a race condition (1)

nomad/node_endpoint.go
tgross: We could definitely call `SetIdentity` inside `Authenticate`, as we always want it. Should be easy enough to do.

But although I love the idea of wrapping it into `Server.forward` in principle, there's a bunch of leader-only RPCs like `Eval.Dequeue` that don't forward because we _always_ send them to the leader. This is why I was exploring that RPC middleware notion previously (but the type signatures are _terrible_ :grin: )

nomad/acl.go
tgross: This is actually just totally wrong, so thanks for catching it. I made this change to fix a NPE coming from variables (which may have claims instead of an ACL token), but that's not safe to do for the other consumers of this function. I'll back this bit out and check for the nil pointer in the caller in variables instead.

nomad/node_endpoint.go
tgross: Ok, I've moved the `SetIdentity` into `Authenticate`, which also cleans up `Authenticate`'s signature so that if we eventually get it to produce ACLs for claims we don't have to change it again everywhere

tgross: I've also updated the semgrep rule so that if we're using `Authenticate` we expect to return the error at some point. That should minimize the risk of misuse.

tgross:
I've extracted a subset of this PR out into https://github.com/hashicorp/nomad/pull/15734 to pick up some work @schmichael and I discussed in a sidebar (and so that we're keeping the history a little easier to read)

tgross:
Rebased and should be ready for final review.

client/rpc.go
```diff
@@ -435,7 +435,7 @@ func resolveServer(s string) (net.Addr, error) {
 // a potential error.
 func (c *Client) Ping(srv net.Addr) error {
 	var reply struct{}
-	err := c.connPool.RPC(c.Region(), srv, "Status.Ping", struct{}{}, &reply)
+	err := c.connPool.RPC(c.Region(), srv, "Status.Ping", &structs.GenericRequest{}, &reply)
```
shoenig, line 438: Could this be a singleton, avoiding the extra malloc on every request?

shoenig (approved):
LGTM - it does seem like some middleware could help here but I didn't get much of an improvement over what you already have

client/rpc.go
tgross: We don't mutate it, so yeah that should be safe. Done.

PR #15514, opened 2022-12-09
artifact: enable inheriting environment variables from client

### LINKED ISSUE #15497: artifact: enable setting custom environment variables in artifact downloader sandbox
In #15328 Nomad now downloads artifacts in a child process of the Nomad client agent. Part of this includes setting a minimal process environment, so that the go-getter internals do not have access to the potentially secretive environment variables of the Nomad agent. Some folks may be relying on environment variables to control go-getter (e.g. s3/git/hg), so we'll want to enable customizing the environment that gets set for the child process.

### LINKED ISSUE #15498: artifact: enable reading system gitconfig
In #15328 Nomad restricts access to the filesystem, including blocking reads from `/etc/` except for specific files. Previously artifact downloads were executed as the Nomad agent's user (`root`) and as such git configuration would be read from `/root/.gitconfig`. Now that the downloads are executed as the `nobody` user, we should be reading git config from the system configuration (`nobody` ostensibly has no `$HOME`).

We could _probably_ get away with enabling reads on `/etc/gitconfig` but the true git way of doing things means we need to find the `git` executable and compute the `(prefix)` value.

https://git-scm.com/docs/git-config#Documentation/git-config.txt---system

```
For writing options: write to system-wide $(prefix)/etc/gitconfig rather than the repository .git/config.

For reading options: read only from system-wide $(prefix)/etc/gitconfig rather than from all available files.
```

Might end up turning this into a helper function in `go-landlock` like we do for finding certificates, etc.

shoenig:
This PR adds client configuration for specifying environment variables that
should be inherited by the artifact sandbox process from the Nomad Client agent.

Most users should not need to set these values but the configuration is provided
to ensure backwards compatability. Configuration of go-getter should ideally be
done through the artifact block in a jobspec task.

e.g.

```hcl
client {
  artifact {
    set_environment_variables = "TMPDIR,GIT_SSH_OPTS"
  }
}
```

Closes #15497

### DIFF
--- a/nomad/structs/config/artifact.go
+++ b/nomad/structs/config/artifact.go
@@ -39,6 +39,11 @@ type ArtifactConfig struct {
 	// artifact downloader can write only to the task sandbox directory, and can
 	// read only from specific locations on the host filesystem.
 	DisableFilesystemIsolation *bool `hcl:"disable_filesystem_isolation"`
+
+	// SetEnvironmentVariables is a comma-separated list of environment
+	// variable names to inherit from the Nomad Client and set in the artifact
+	// download sandbox process.
+	SetEnvironmentVariables *string `hcl:"set_environment_variables"`
 }
 
 func (a *ArtifactConfig) Copy() *ArtifactConfig {
@@ -53,6 +58,7 @@ func (a *ArtifactConfig) Copy() *ArtifactConfig {
 		HgTimeout:                  pointer.Copy(a.HgTimeout),
 		S3Timeout:                  pointer.Copy(a.S3Timeout),
 		DisableFilesystemIsolation: pointer.Copy(a.DisableFilesystemIsolation),
+		SetEnvironmentVariables:    pointer.Copy(a.SetEnvironmentVariables),
 	}
 }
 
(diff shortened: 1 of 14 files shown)


nomad/structs/config/artifact.go
```diff
@@ -39,6 +39,11 @@ type ArtifactConfig struct {
 	// artifact downloader can write only to the task sandbox directory, and can
 	// read only from specific locations on the host filesystem.
 	DisableFilesystemIsolation *bool `hcl:"disable_filesystem_isolation"`
+
+	// SetEnvironmentVariables is a comma-separated list of environment
+	// variable names to inherit from the Nomad Client and set in the artifact
+	// download sandbox process.
+	SetEnvironmentVariables *string `hcl:"set_environment_variables"`
```
tgross, line 46: On the name "set": is there future in which we might want to let users set specific env vars, rather than having them inherit from the client?

website/content/docs/configuration/client.mdx
```diff
@@ -382,6 +382,10 @@ see the [drivers documentation](/docs/drivers).
   isolation should be disabled for artifact downloads. Applies only to systems
   where filesystem isolation via [landlock] is possible (Linux kernel 5.13+).
 
+- `set_environment_variables` `(string:"")` - Specifies a comma separated list
+  of environment variables that should be inherited by the artifact sandbox from
+  the Nomad client's environment.
```
tgross, line 387: Do we want to call out the default variables here so that people don't think they need to add `PATH`?

tgross (approved):
LGTM! I can bikeshed the name but I'm happy with it as-is as well :grinning:

nomad/structs/config/artifact.go
shoenig: I think we can continue using this field to support setting custom values, just by interpreting '=' as key+value,  e.g.

```
set_environment_variables = "BLAH,PATH=/opt/custom/bin"
```

But that would be beyond compatibility support, so that's a PR for another day

PR #15515, opened 2022-12-09
add metric for rate of RPC requests

tgross:
Implement a metric for RPC requests with labels on the identity, so that administrators can monitor the source of requests within the cluster. This changeset demonstrates the change with the new `ACL.WhoAmI` RPC, and we'll wire up the remaining RPCs once we've threaded the new pre-forwarding authentication through the all.


shoenig:
@tgross mind rebasing? Looks like this got wrecked by https://github.com/hashicorp/nomad/pull/15512

nomad/rpc_rate_metrics.go
```diff
@@ -0,0 +1,25 @@
...
+	if rpcCtx == nil {
+		return // we're the RPC caller and not the server
+	}
+	if !srv.config.ACLEnabled || identity == nil {
+		// If ACLs aren't enabled, we never have a sensible identity
+		metrics.IncrCounter([]string{"nomad", "rpc", endpoint, op}, 1)
+	}
+	metrics.IncrCounterWithLabels(
+		[]string{"nomad", "rpc", endpoint, op}, 1,
+		[]metrics.Label{{Name: "id", Value: identity.String()}})
```
shoenig, line 24: bikeshed: would "identity" instead of "id" make more sense? since this can be a non-UUID

shoenig (approved):
LGTM!

jrasell (approved):
LGTM!

tgross:
Just leaving a note that I'm having strong re-consideration of this approach following the investigation that's led to https://github.com/hashicorp/nomad/pull/15523 and https://github.com/hashicorp/nomad/pull/15522, so I'm holding off on merging this until we can have an internal discussion around it.

nomad/rpc_rate_metrics.go
```diff
@@ -0,0 +1,25 @@
...
+
+// MeasureRPCRate increments the appropriate rate metric for this endpoint,
+// with a label from the identity
+func (srv *Server) MeasureRPCRate(
+	endpoint, op string,
+	identity *structs.AuthenticatedIdentity, rpcCtx *RPCContext,
+) {
+	if rpcCtx == nil {
+		return // we're the RPC caller and not the server
+	}
```
tgross, line 17: In some local testing I'm seeing this comment isn't quite accurate; I need to debug this a bit more to verify.

schmichael (approved):

nomad/rpc_rate_metrics.go
tgross: So the intention from the RFC was to count whenever an RPC is served by a server or forwarded by a server (because that let's you detect problems like uneven distribution of RPCs from client nodes). If we only cared about metrics on the server that actually serves the request, we could dramatically simplify the problem because auth could happen after checking for forwarding. But I'm pretty sure we do want metrics pre-forwarding?

My intent with this conditional around `rpcCtx == nil` was that we probably don't want to track "static endpoint" RPCs where ex. the deploymentwatcher running on the leader is using RPCs as normal function calls. Unfortunately that also means that if you send a HTTP request to a server, the server that serves the HTTP request won't count it because there's no `rpcCtx` in that case (ex. if the HTTP request hits the leader).

schmichael: > My intent with this conditional around rpcCtx == nil was that we probably don't want to track "static endpoint" RPCs where ex. the deploymentwatcher running on the leader is using RPCs as normal function calls.

Ah that makes sense. I'd love to stop using those static endpoints anyway. I can't imagine there's not a better way to share that logic internally. Perhaps we should file an issue around this (eg `Remove static endpoints to avoid counting them toward metrics` or something) and ship this as-is.

tgross:
Will need to rebase this on https://github.com/hashicorp/nomad/pull/15513 once that's been merged, as there's some changes that impact this one

shoenig:
test failures include

```
  acl_test.go:268: expected matching error strings
  ↪ msg: "ACL token expired"
  ↪ err: "rpc error: ACL token expired"
```

tgross:
> test failures include

Bah, that keeps breaking/unbreaking as I've done rebases with the other PRs for this topic. Will fix.

tgross:
Had to rebase this on the merged #15513 and I've got a telemetry config test I need to fix, and that should wrap this up.

PR #15519, opened 2022-12-09
Backport of Use custom runners for building nomad into release/1.2.x

mdeggies:
Backport of https://github.com/hashicorp/nomad/pull/15490

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,7 @@ jobs:
         # version, because "goenv" can react to it automatically.
         run: |
           echo "Building with Go $(cat .go-version)"
-          echo "::set-output name=go-version::$(cat .go-version)"
+          echo "go-version=$(cat .go-version)" >> $GITHUB_OUTPUT
   get-product-version:
     runs-on: ubuntu-20.04
     outputs:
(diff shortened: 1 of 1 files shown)


tgross (approved):
:+1:

Just a heads up that we seem to be missing the line https://github.com/hashicorp/nomad/pull/15521/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R6 that was in the other 2 PRs

mdeggies:
Ah yeah- For the `release/1.2.x` branch, builds are triggered on the `release/1.2.x` branch already, which is what we want: https://github.com/hashicorp/nomad/blob/release/1.2.x/.github/workflows/build.yml#L6. 

For `release/1.3.x` and `release/1.4.x`, builds are only being triggered on `main`- so commits to these two release branches are not being built continuously, and if you cut a new release branch from either of these two branches, you could run into unexpected failures. I think we want is to have builds run on each respective release branch.

mdeggies:
Thanks, Tim! Going to merge these backport PR's in, then merge in https://github.com/hashicorp/nomad/pull/14630 across branches.

PR #15522, opened 2022-12-09
Add raft snapshot configuration options

angrycub:
This plumbs raft snapshot configuration out to the server config

### DIFF
--- a/.changelog/15522.txt
+++ b/.changelog/15522.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+server: Added raft snapshot arguments to server config
+```
--- a/.changelog/15522.txt
+++ b/.changelog/15522.txt
@@ -1,3 +1,4 @@
 ```release-note:improvement
 server: Added raft snapshot arguments to server config
+server: Certain raft configuration elements can now be reloaded without restarting the server
 ```
--- a/command/agent/agent.go
+++ b/command/agent/agent.go
@@ -214,10 +214,10 @@ func convertServerConfig(agentConfig *Config) (*nomad.Config, error) {
 	}
 
 	if vPtr := agentConfig.Server.RaftTrailingLogs; vPtr != nil {
-		conf.RaftConfig.TrailingLogs = uint64(*vPtr)
 		if *vPtr < 1 {
 			return nil, fmt.Errorf("raft_trailing_logs must be non-negative, got %d", *vPtr)
 		}
+		conf.RaftConfig.TrailingLogs = uint64(*vPtr)
 	}
 
 	if vPtr := agentConfig.Server.RaftSnapshotInterval; vPtr != nil {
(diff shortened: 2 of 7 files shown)


command/agent/config.go
```diff
@@ -610,6 +610,27 @@ type ServerConfig struct {
...
+	// the entire cluster from performing a snapshot at once. The value passed
+	// here is the initial setting used.
+	// TODO: This can be tuned during operation with a hot reload.
+	RaftSnapshotInterval *string `hcl:"raft_snapshot_interval"`
+
+	// RaftTrailingLogs controls how many logs are left after a snapshot. This is
+	// used so that we can quickly replay logs on a follower instead of being
+	// forced to send an entire snapshot. The value passed here is the initial
+	// setting used.
+	// TODO: This can be tuned during operation using a hot reload.
```
schmichael, line 632: Can these TODOs be removed? Looks like it's implemented below.

schmichael (approved):

.changelog/15522.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+server: Added raft snapshot arguments to server config
+```
```
tgross, line 3: I can't make the "suggestion" work here in the code review interface because of the triple backticks, but maybe we want to add a second improvement for this same entry:

> server: Raft configuration can now be reloaded without restarting the server

command/agent/agent.go
```diff
@@ -212,6 +212,33 @@ func convertServerConfig(agentConfig *Config) (*nomad.Config, error) {
 			return nil, fmt.Errorf("raft_multiplier cannot be %d. Must be between 1 and %d", *agentConfig.Server.RaftMultiplier, MaxRaftMultiplier)
 		}
 	}
+
+	if vPtr := agentConfig.Server.RaftTrailingLogs; vPtr != nil {
+		conf.RaftConfig.TrailingLogs = uint64(*vPtr)
+		if *vPtr < 1 {
+			return nil, fmt.Errorf("raft_trailing_logs must be non-negative, got %d", *vPtr)
+		}
```
tgross, line 220: It's maybe safe because we're returning an error which the caller is going to bubble up and then bail out... but we probably shouldn't do the cast and update until after we've validated `vPtr`'s value, otherwise we'll get an underflow on the config value:

```suggestion
		if *vPtr < 1 {
			return nil, fmt.Errorf("raft_trailing_logs must be non-negative, got %d", *vPtr)
		}
		conf.RaftConfig.TrailingLogs = uint64(*vPtr)
```

tgross (approved):
Left one thing around validation we should probably fix, but other than that this LGTM! Nice work

command/agent/agent.go
angrycub: Moved in both cases I saw it.  Added agent test to check the -1 cases for trailing logs and snapshot threshold.

PR #15523, opened 2022-12-10
core: backoff considerably when worker is behind raft

schmichael:
**Leaving as a draft to discuss approach and appropriate testing.**

Upon dequeuing an evaluation workers snapshot their state store at the eval's wait index or later. This ensures we process an eval at a point in time after it was created or updated. Processing an eval on an old snapshot could cause any number of problems such as:

1. Since job registration atomically updates an eval and job in a single raft entry, scheduling against indexes before that may not have the eval's job or may have an older version.
2. The older the scheduler's snapshot, the higher the likelihood something has changed in the cluster state which will cause the plan applier to reject the scheduler's plan. This could waste work or even cause eval's to be failed needlessly.

However, the workers run in parallel with a new server pulling the cluster state from a peer. During this time, which may be many minutes long, the state store is likely far behind the minimum index required to process evaluations.

This PR addresses this by adding an additional long backoff period after an eval is nacked. If the scheduler's indexes catches up within the additional backoff, it will unblock early to dequeue the next eval.

While there may be more precise ways to block scheduling until the server catches up, this approach adds little risk and covers additional cases where a server may be temporarily behind due to a spike in load or a saturated network.

### DIFF
--- a/nomad/state/state_store.go
+++ b/nomad/state/state_store.go
@@ -242,7 +242,7 @@ func (s *StateStore) SnapshotMinIndex(ctx context.Context, index uint64) (*State
 		// Get the states current index
 		snapshotIndex, err := s.LatestIndex()
 		if err != nil {
-			return nil, fmt.Errorf("failed to determine state store's index: %v", err)
+			return nil, fmt.Errorf("failed to determine state store's index: %w", err)
 		}
 
 		// We only need the FSM state to be as recent as the given index
--- a/nomad/worker.go
+++ b/nomad/worker.go
@@ -3,6 +3,7 @@ package nomad
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 	"sync"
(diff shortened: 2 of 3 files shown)


nomad/state/state_store.go
```diff
@@ -242,7 +242,7 @@ func (s *StateStore) SnapshotMinIndex(ctx context.Context, index uint64) (*State
 		// Get the states current index
 		snapshotIndex, err := s.LatestIndex()
 		if err != nil {
-			return nil, fmt.Errorf("failed to determine state store's index: %v", err)
+			return nil, fmt.Errorf("failed to determine state store's index: %w", err)
```
tgross, line 245: I'm :+1: for wrapping all the errors we can, but FWIW this is one of the many cases where we can't ever fail b/c the inner call to [`txn.Get`](https://github.com/hashicorp/go-memdb/blob/main/txn.go#L744) only returns an error if we give it the wrong table or index. (I'd be willing to make an argument we should panic in the state store helpers around those methods, but that's certainly out of scope for this PR.)

nomad/worker.go
```diff
@@ -414,8 +415,26 @@ func (w *Worker) run() {
...
+				w.sendNack(eval, token)
+
+				// Timing out here means this server is woefully behind the leader's
+				// index. This can happen when a new server is added to a cluster and
+				// must initially sync the cluster state.
+				// Backoff dequeuing another eval until there's some indication this
+				// server would be up to date enough to process it.
+				const slowServerSyncLimit = 10 * raftSyncLimit
+				if _, err := w.snapshotMinIndex(waitIndex, slowServerSyncLimit); err != nil {
+					w.logger.Warn("server is unable to catch up to last eval's index", "error", err)
```
tgross, line 430: If we get this warning, it's going to be quite a while after the previous log line (50s), so it might be hard to correlate with the previous one unless you're already familiar with the behavior, especially because we'll have quite a few workers running concurrently which should all have the same problem. Given this should be rare, maybe we can add more context here around which eval we previously dequeued?

tgross: Re-reading this I'm realizing we don't care about the previous eval by the time we've hit this, as we've already nacked it and we have the index in the error message.

tgross:
I've rebased this on main and made two changes:
* I've picked up the dropped `context.Cancelled` error from the snapshot min index, so that we have a nicer warning when the server gets shut down while we're waiting.
* I've added a test for the logic, which meant adding a parameter to `Worker.run` for testing purposes. I could add _all_ the timeout parameters as fields on `Worker` and have them set when the worker gets created, but there didn't seem a ton of value in it without writing a bunch more test code and shipping this change independently would be nice.

Going to tag @schmichael for review as it was his code to start but will pull in some other folks too.

nomad/worker_test.go
```diff
@@ -338,6 +341,54 @@ func TestWorker_sendAck(t *testing.T) {
...
+	doneCh := make(chan struct{})
+
+	go func() {
+		w.run(time.Millisecond)
+		doneCh <- struct{}{}
+	}()
+
+	// We expect to be paused for 10ms + 1ms but otherwise can't be all that
+	// precise here because of concurrency. But checking coverage for this test
+	// shows we've covered the logic
```
schmichael, line 374: :+1:

schmichael (commented):
Ha, I can't actually approve this because I'm the owner, but LGTM! Thanks for cleaning it up and finding a testing path. Looks great.

jrasell (approved):
LGTM!

PR #15529, opened 2022-12-12
ACL: add ACL binding rule RPC and HTTP API handlers.

jrasell:
Best viewed commit by commit which is split by Nomad subsystem.

Builds on #15511 
Related to #13120

### DIFF
--- a/api/acl_test.go
+++ b/api/acl_test.go
@@ -618,7 +618,6 @@ func TestACLAuthMethods(t *testing.T) {
 	must.Len(t, 1, aclAuthMethodsListResp)
 	must.Eq(t, authMethod.Name, aclAuthMethodsListResp[0].Name)
 	must.True(t, aclAuthMethodsListResp[0].Default)
-	must.SliceNotEmpty(t, aclAuthMethodsListResp[0].Hash)
 	assertQueryMeta(t, queryMeta)
 
 	// Read the auth-method.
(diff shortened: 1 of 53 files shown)


nomad/leader.go
```diff
@@ -62,6 +62,14 @@ var minACLRoleVersion = version.Must(version.NewVersion("1.4.0"))
 // 1.5, otherwise it's hard to test the functionality
 var minACLAuthMethodVersion = version.Must(version.NewVersion("1.4.3-dev"))
 
+// minACLBindingRuleVersion is the Nomad version at which the ACL binding rules
+// table was introduced. It forms the minimum version all federated servers
+// must meet before the feature can be used.
+//
+// TODO: version constraint will be updated for every beta or rc until we reach
+// 1.5, otherwise it's hard to test the functionality
+var minACLBindingRuleVersion = version.Must(version.NewVersion("1.4.3-dev"))
```
tgross, line 71: Once we go to 1.5-beta we can update this to `1.5.0` and then use [`Version.Core`](https://pkg.go.dev/github.com/hashicorp/go-version?#Version.Core) on the version we check.

nomad/acl_endpoint.go
```diff
@@ -1987,3 +1987,324 @@ func (a *ACL) WhoAmI(args *structs.GenericRequest, reply *structs.ACLWhoAmIRespo
...
+		return err
+	}
+	for _, bindingRule := range args.ACLBindingRules {
+		lookupBindingRule, err := stateSnapshot.GetACLBindingRule(nil, bindingRule.ID)
+		if err != nil {
+			return structs.NewErrRPCCodedf(400, "ACL binding rule lookup failed: %v", err)
+		}
+		if lookupBindingRule != nil {
+			reply.ACLBindingRules = append(reply.ACLBindingRules, lookupBindingRule)
+		}
```
tgross, line 2095: State store read methods don't usually return an error unless its internal. And in any case if we get either an error or a nil value for this query, that seems like an internal error (5xx) to me rather than a request error (4xx) -- we should have just written this value so if we didn't, something went very wrong!

api/acl_test.go
```diff
@@ -655,3 +654,85 @@ func TestACLAuthMethods(t *testing.T) {
 	must.Len(t, 0, aclAuthMethodsListResp)
 	assertQueryMeta(t, queryMeta)
 }
+
+func TestACLACLBindingRules(t *testing.T) {
```
tgross, line 658:
```suggestion
func TestACLBindingRules(t *testing.T) {
```

command/agent/acl_endpoint.go
```diff
@@ -676,3 +676,150 @@ func (s *HTTPServer) aclAuthMethodUpsertRequest(
...
+	var aclBindingRule structs.ACLBindingRule
+	if err := decodeBody(req, &aclBindingRule); err != nil {
+		return nil, CodedError(http.StatusBadRequest, err.Error())
+	}
+
+	// Ensure the request path ID matches the ACL binding rule ID that was
+	// decoded. Only perform this check on updates as a generic error on
+	// creation might be confusing to operators as there is no specific binding
+	// rule request path.
+	if ruleID != "" && ruleID != aclBindingRule.ID {
```
tgross, line 806: Doesn't this condition imply that we allow users to set the ID on creation if they include it in the request body? The server is supposed to create it as a UUID.

tgross (approved):
Left some comments but overall LGTM!

api/acl_test.go
jrasell: Fixed within https://github.com/hashicorp/nomad/pull/15529/commits/4754d09f3d1f935e186d336a913f7ba40b4b78f3

nomad/acl_endpoint.go
jrasell: Agreed on both counts, thanks! Fixed within https://github.com/hashicorp/nomad/pull/15529/commits/4754d09f3d1f935e186d336a913f7ba40b4b78f3

command/agent/acl_endpoint.go
jrasell: This function can be used by new creations or updates to existing binding rules. Therefore I feel it's worth checking that if the caller included an ID within the payload, that this is checked against the request path to ensure they match, otherwise we don't know which the correct one is. The RPC handler will check that the ID is found within state, indicating it is an update, before moving forward.

This comment did get me thinking about this function and I made a small modification within https://github.com/hashicorp/nomad/pull/15529/commits/4754d09f3d1f935e186d336a913f7ba40b4b78f3 which sets the aclBindingRule.ID to the request path ID, if the former has been left empty by the caller.

Issue #12174, opened 2022-03-02
Editing an existing job definition deletes expose paths


jhitt25:
### Nomad version
Nomad v1.2.6 (a6c6b475db5073e33885377b4a5c733e1161020c)

### Operating system and Environment details
8 node nomad/consul cluster, 3 control, 5 worker nodes.  All nodes running Nomad 1.2.6, Consul 1.11.3 and Oracle Linux 7.9 with kernel 5.4.17

### Issue
When editing an existing job definition in the UI manually defined expose paths are lost, even if no actual edit is performed.  Simply clicking 'Edit', then 'Plan' and 'Run' will remove existing expose paths.

### Reproduction steps
Run a simple job:
```
job "test-job" {
  datacenters = ["centralus"]
  type        = "service"

  group "test-group" {
    count = 1

    network {
        mode = "bridge"
        port "http-echo" {}
        port "health"  {}
    }

    service {
      name = "test-service"
      port = "8080"

      connect {
        sidecar_service {
          proxy {
            expose {
              path {
                path = "/"
                protocol = "http"
                local_path_port = 8080
                listener_port = "health"
              }
            }
          }
        }
      }
    }

    task "test-task" {
      driver = "docker"
      config {
        image = "hashicorp/http-echo:latest"
        args = [ "-listen=:8080", "-text='expose test'" ]
      }
    }
  }
}
```

If you view the definition, under TaskGroups[0].Services[0].Connect.SidecarService.Proxy you will have:
```
                "Expose": {
                  "Paths": [
                    {
                      "Path": "/",
                      "Protocol": "http",
                      "LocalPathPort": 8080,
                      "ListenerPort": "health"
                    }
                  ]
                },
```
Next, edit the job (there's no need to make any changes), plan, and run.

If you view the definition now, TaskGroups[0].Services[0].Connect.SidecarService.Proxy.Expose will be null.

#### Expected Result
The expose paths will be retained

#### Actual Result
The expose paths are removed

jrasell:
Hi @jhitt25 and thanks for raising this issue.

I have not been able to reproduce this locally using v1.2.6 or current main yet. Could you share the details reported when you perform the plan phase of the reproduction along with any logs from the server?

jrasell:
Hi @jhitt25 and thanks for the updated response. I notice in the file `5_run_request` that the expose section is detailed as expected which suggests the UI is sending the correct request to Nomad.

> nothing showing up in the Nomad logs

[...]

Issue #11304, opened 2021-10-13
API/Inspect loses  ConsulConnect params


pserranoa:
### Nomad version
1.0.6

### Operating system and Environment details
NAME="Red Hat Enterprise Linux"
VERSION="8.2 (Ootpa)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="8.2"
PLATFORM_ID="platform:el8"

### Issue
Some parameters of nomad job where missed after deploy a job with connect integration when you use the ui/api with json to make changes 

### Reproduction steps
deploy a job that expose something like this
```
connect {
          sidecar_service {
            proxy {
              expose {
                path {
                  path            = "/metrics"
                  protocol        = "http"
                  local_path_port = 8085
                  listener_port   = "api_expose_metrics"
                }
              }
            }
          }
        }
```

When the job is correctly running, go to definition of job, click on edit and plan without changes, you will receive a message with some changes to apply 

```
+/- Job: "my_job"
+/- Task Group: "my_group" ( 1 in-place update )
+/- Service {
    AddressMode: "auto"
    EnableTagOverride: "false"
    Name: "openidm-my_job-service"
    PortLabel: "api"
    TaskName: "my_task"
    Tags {
      Tags: "metrics"
    }
    +/- ConsulConnect {
      Native: "false"
      +/- SidecarService {
        Port: ""
      +/- ConsulProxy {
        LocalServiceAddress: ""
        LocalServicePort: "0"
        }
      }
    }
  }
Task: "connect-proxy-my_job-service"
Task: my_task"

```
#### Expected Result
If you use the  deinition at UI, click edit and make a plan you should receive a message like
```
Job: "my_job"
Task Group: "my_group" (1 in-place update)
  Task: "connect-proxy-my_job-service"
  Task: "my_task"

Scheduler dry-run:
- All tasks successfully allocated.
```

#### Actual Result
```
+/- Job: "my_job"
+/- Task Group: "my_group" ( 1 in-place update )
+/- Service {
    AddressMode: "auto"
    EnableTagOverride: "false"
    Name: "openidm-my_job-service"
    PortLabel: "api"
    TaskName: "my_task"
    Tags {
      Tags: "metrics"
    }
    +/- ConsulConnect {
      Native: "false"
      +/- SidecarService {
        Port: ""
      +/- ConsulProxy {
        LocalServiceAddress: ""
        LocalServicePort: "0"
        }
      }
    }
  }
[...]

Issue #9379, opened 2020-11-17
Expose Paths not returned from the API


jorgemarey:
### Nomad version
Nomad v0.12.7

### Issue
When running jobs with `Expose Paths` in the sidecar_service stanza those are not returned when querying the job API.

The job runs correctly when using the HCL and the nomad client, but if we query the job (via curl or `nomad inspect`) the expose configuration is not presented. This is a problem also if we try to modify the job directly via the UI.

### Reproduction steps

1) Run job file
2) nomad inspect job

- The expose config won't be present:

```
            "SidecarService": {
              "Port": "",
              "Proxy": {
                "Config": null,
                "ExposeConfig": null,
                "LocalServiceAddress": "",
                "LocalServicePort": 0,
                "Upstreams": null
              },
```

It seems that the problem is when encoding the struct back to the client. Maybe some problem with the msgpack codec? Maybe the tags are not used correctly?

### Job file (if appropriate)

```
job "test" {
  datacenters = ["dc1"]
  type        = "service"

  group "api" {
    count = 1

    network {
      mode = "bridge"
      port "api_expose_healthcheck" {
        to = -1
      }
    }

    service {
      name = "test"
      port = "3000"
      check {
        name     = "http-check"
        port     = "api_expose_healthcheck"
        type     = "http"
        path     = "/health"
        interval = "10s"
        timeout  = "2s"
      }
      connect {
        sidecar_service {
          proxy {
            expose {
              path {
                path            = "/health"
                protocol        = "http"
                local_path_port = 3000
                listener_port   = "api_expose_healthcheck"
              }
            }
          }
        }
      }
    }

    task "api" {
      driver = "docker"
      config {
        image = "..."
      }
      resources {
        cpu    = 200
        memory = 100
      }
    }
  }
}
```

tgross:
Hi @jorgemarey! I tested this with the current `master` which we're getting ready to ship as Nomad 1.0 GA soon. It looks like it's working there.... I don't see it in the changelog for any release and the original commit eef81c3b4f400a3a01705b410f3440c51b7cd12e LGTM. I'll take a look at earlier versions and see if I can reproduce and get that into the changelog properly if it was fixed.

---

I used following job derived from the jobspec you provided above:

<details><summary>test.nomad</summary>

```hcl
job "test" {
  datacenters = ["dc1"]
  type        = "service"

  group "api" {
    count = 1

    network {
      mode = "bridge"
      port "api_expose_healthcheck" {
[...]

PR #15541, opened 2022-12-13
Rename fields on proxyConfig

jorgemarey:
This renames two fields in the api for proxyConfig.

- ExposeConfig to Expose
- Path to Paths

The old fields are mantained for backwards compatibility

Fixes https://github.com/hashicorp/nomad/issues/11304
Fixes https://github.com/hashicorp/nomad/issues/12174
Fixes https://github.com/hashicorp/nomad/issues/9379

### DIFF
--- a/.changelog/15541.txt
+++ b/.changelog/15541.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+api: Fixed a bug where exposeConfig field was not provided correctly when getting the jobs via the API
+```
--- a/.changelog/15541.txt
+++ b/.changelog/15541.txt
@@ -1,3 +1,11 @@
 ```release-note:bug
 api: Fixed a bug where exposeConfig field was not provided correctly when getting the jobs via the API
 ```
+
+```release-note:deprecation
+api: The connect `ConsulProxy.ExposeConfig` field is deprecated in favor of `ConsulProxy.Expose`
+```
+
+```release-note:deprecation
+api: The connect `ConsulExposeConfig.Path` field is deprecated in favor of `ConsulExposeConfig.Paths`
+```
--- a/api/consul.go
+++ b/api/consul.go
@@ -135,13 +135,12 @@ func (st *SidecarTask) Canonicalize() {
 
 // ConsulProxy represents a Consul Connect sidecar proxy jobspec stanza.
 type ConsulProxy struct {
-	LocalServiceAddress string              `mapstructure:"local_service_address" hcl:"local_service_address,optional"`
-	LocalServicePort    int                 `mapstructure:"local_service_port" hcl:"local_service_port,optional"`
-	Expose              *ConsulExposeConfig `mapstructure:"expose" hcl:"expose,block"`
-	// Deprecated, only to maintain backwards compatibility. Use Expose instead
-	ExposeConfig *ConsulExposeConfig
-	Upstreams    []*ConsulUpstream      `hcl:"upstreams,block"`
-	Config       map[string]interface{} `hcl:"config,block"`
+	LocalServiceAddress string                 `mapstructure:"local_service_address" hcl:"local_service_address,optional"`
+	LocalServicePort    int                    `mapstructure:"local_service_port" hcl:"local_service_port,optional"`
+	Expose              *ConsulExposeConfig    `mapstructure:"expose" hcl:"expose,block"`
+	ExposeConfig        *ConsulExposeConfig    // Deprecated: only to maintain backwards compatibility. Use Expose instead.
+	Upstreams           []*ConsulUpstream      `hcl:"upstreams,block"`
+	Config              map[string]interface{} `hcl:"config,block"`
 }
 
 func (cp *ConsulProxy) Canonicalize() {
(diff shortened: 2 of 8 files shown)


jrasell (commented):
Hi @jorgemarey and thanks a lot for raising this PR. While we get around to reviewing this, I wonder if it would be possible to add a changlog entry; we have [this guide](https://github.com/hashicorp/nomad/blob/main/contributing/CHANGELOG.md#developer-guide) available detailing the format.

jrasell (commented):
Hi @jorgemarey and thanks again.

Unfortunately renaming an internal struct field will break the messagepack encoding we use to serialize to the state store. This means we therefore cannot accept a fix of this nature.

jorgemarey:
Hi @jrasell. I thought of that, but wasn't sure about it. 
I don't know if it's a problem with the msgpack library or other thing, but when returning the job via the API, those fields are not serialized correctly and so the input Job differs from the output job, which is a real issue get doing thinks like get a job, modify a parameter and run it again.

Do you have any ideas of how to best solve this?

jorgemarey:
Hi again, If this aproach doesn't work. How about chaning the API but making it compatible with the current format. It would be something like the following (we'll need to add the changes to the Path field also):

```
diff --git a/api/consul.go b/api/consul.go
index 9a76bfb329..7457ec91f4 100644
--- a/api/consul.go
+++ b/api/consul.go
@@ -133,11 +133,13 @@ func (st *SidecarTask) Canonicalize() {
 
 // ConsulProxy represents a Consul Connect sidecar proxy jobspec stanza.
 type ConsulProxy struct {
-	LocalServiceAddress string                 `mapstructure:"local_service_address" hcl:"local_service_address,optional"`
-	LocalServicePort    int                    `mapstructure:"local_service_port" hcl:"local_service_port,optional"`
-	ExposeConfig        *ConsulExposeConfig    `mapstructure:"expose" hcl:"expose,block"`
-	Upstreams           []*ConsulUpstream      `hcl:"upstreams,block"`
-	Config              map[string]interface{} `hcl:"config,block"`
+	LocalServiceAddress string              `mapstructure:"local_service_address" hcl:"local_service_address,optional"`
+	LocalServicePort    int                 `mapstructure:"local_service_port" hcl:"local_service_port,optional"`
+	Expose              *ConsulExposeConfig `mapstructure:"expose" hcl:"expose,block"`
+	// Deprecated, only to maintain backwards compatibility
+	ExposeConfig *ConsulExposeConfig
+	Upstreams    []*ConsulUpstream      `hcl:"upstreams,block"`
+	Config       map[string]interface{} `hcl:"config,block"`
 }
 
 func (cp *ConsulProxy) Canonicalize() {
@@ -145,7 +147,7 @@ func (cp *ConsulProxy) Canonicalize() {
 		return
 	}
 
-	cp.ExposeConfig.Canonicalize()
+	cp.Expose.Canonicalize()
 
 	if len(cp.Upstreams) == 0 {
 		cp.Upstreams = nil
diff --git a/command/agent/job_endpoint.go b/command/agent/job_endpoint.go
index a6440b924c..45d96f1966 100644
--- a/command/agent/job_endpoint.go
+++ b/command/agent/job_endpoint.go
@@ -1646,11 +1646,17 @@ func apiConnectSidecarServiceProxyToStructs(in *api.ConsulProxy) *structs.Consul
 	if in == nil {
 		return nil
 	}
+	// TODO: to maintain backwards compatibility
+	expose := in.Expose
+	if in.ExposeConfig != nil {
+		expose = in.ExposeConfig
+	}
+
 	return &structs.ConsulProxy{
 		LocalServiceAddress: in.LocalServiceAddress,
 		LocalServicePort:    in.LocalServicePort,
 		Upstreams:           apiUpstreamsToStructs(in.Upstreams),
-		Expose:              apiConsulExposeConfigToStructs(in.ExposeConfig),
+		Expose:              apiConsulExposeConfigToStructs(expose),
 		Config:              maps.Clone(in.Config),
 	}
 }
diff --git a/jobspec/parse_service.go b/jobspec/parse_service.go
index 7dc10214b3..8dcc9a1ed1 100644
--- a/jobspec/parse_service.go
+++ b/jobspec/parse_service.go
@@ -819,7 +819,7 @@ func parseProxy(o *ast.ObjectItem) (*api.ConsulProxy, error) {
 		if e, err := parseExpose(eo.Items[0]); err != nil {
 			return nil, err
 		} else {
-			proxy.ExposeConfig = e
+			proxy.Expose = e
 		}
 	}
 ```

Basically, change the api field to be Expose (like in the struct field but checking if in the json the value ExposeConfig has content, and in that case use that.

dpewsey:
Hi @jrasell can you check the above? Seems to be a fix and this is quite a significant issue that is broken.  
Limits us from being able to move to nomad right now.

shoenig:
Hi @jorgemarey sorry for letting this issue slip by - but yeah I think introducing the new field name while deprecating the old name but keeping it for compatibility sounds reasonable. We're expecting to ship 1.5-beta early next week so if we can get those changes in, that would be really great.

jorgemarey:
Hi @shoenig , thanks for looking into this. I'll try to make the changes this weekend.

jorgemarey:
Hi @shoenig, I think I made all the needed changes for this to work. 
Let me know if you think I should change anything else (feel free to make the changes yourself if you want to)

shoenig (approved):
LGTM; thanks @jorgemarey!

PR #15543, opened 2022-12-14
api: experimental changes to make api CI tests fast

shoenig:
This PR tries to make API tests run fast, as an experiment to later apply
to all packages. Key changes include

- Swapping freeport for [test/portal](https://github.com/shoenig/test/blob/main/portal/portal.go) for port allocations
- Swappng some uses of WaitForResult with [test/wait](https://github.com/shoenig/test/blob/main/wait/wait.go)
- Turning on parallelism in api/testutil/slow.go
- Turning on `-race` for `test-nomad-submodule` target in GNUMakefile
- Switching to custom public runner (32 vcpu)
- Upgrading to test v0.5.1

There's also chunk of cleanup brought in for the ride.

I've noticed the `xl` runners can be slow to launch from a cold start (a few minutes), but once primed they're fine

Comparison
- [recent PR](https://github.com/hashicorp/nomad/actions/runs/3711372398/jobs/6291764269) without these changes, api runs in ~11 minutes
- this PR, api runs in ~2 minutes

Future Work

Just emphasizing this is part 1 of N - I'd like to get the API package to a nice state before moving on to other packages. (Now that it's fast, let's cleanup remaining legacy patterns)

Should be fine to backport


api/internal/testutil/server.go
```diff
@@ -307,23 +296,30 @@ func (s *TestServer) waitForAPI() {
...
+	}
 
-		return true, nil
-	}, func(err error) {
-		defer s.Stop()
-		s.t.Fatalf("err: %s", err)
-	})
+	test.Wait(s.t,
+		wait.On(
+			wait.ErrorFunc(f),
```
tgross, line 315: How wedded to the name `ErrorFunc` are you vs something like `UntilNilError` or `UntilNoError` (or whatever)? The name could just as easily be saying we're waiting for an error to be non-nil.

tgross (approved):
The changes here LGTM. I like the approach in `test/portal` (it'd be amazing if we could hand off file handles instead to the server, but that's just impractical). Left one comment about the `test/wait` API.

api/internal/testutil/server.go
shoenig: Definitely open to changing the names; let me noodle on that while working on the inverse of wait.On, https://github.com/shoenig/test/issues/80

PR #15552, opened 2022-12-14
Fix rejoin_after_leave behavior

anth0d:
This will effectively change the default behavior of Serf to _not_ rejoin after leave, which (I believe...) is actually the documented behavior. But this could be perceived as a breaking change by some cluster operators who may be accustomed to server nodes ignoring leave events.

Alternative approaches which would not change the default behavior would be
* use a bool pointer in this config field in order to be able to discern `nil` (not configured) from intentionally `false`, and default to `true` as exists today
* introduce a new config field `disable_rejoin_after_leave` which would default to `false` and therefore continue to produce the current behavior

In any case, docs probably need to be updated here and if necessary, a behavior change be communicated out.

### DIFF
--- a/command/agent/agent.go
+++ b/command/agent/agent.go
@@ -309,6 +309,7 @@ func convertServerConfig(agentConfig *Config) (*nomad.Config, error) {
 	conf.RPCAddr.IP = rpcAddr.IP
 	conf.SerfConfig.MemberlistConfig.BindPort = serfAddr.Port
 	conf.SerfConfig.MemberlistConfig.BindAddr = serfAddr.IP.String()
+	conf.SerfConfig.RejoinAfterLeave = agentConfig.Server.RejoinAfterLeave
 
 	// Set up the advertise addresses
 	rpcAddr, err = net.ResolveTCPAddr("tcp", agentConfig.AdvertiseAddrs.RPC)
--- a/command/agent/config.go
+++ b/command/agent/config.go
@@ -547,7 +547,7 @@ type ServerConfig struct {
 	RetryIntervalHCL string `hcl:"retry_interval" json:"-"`
 
 	// RejoinAfterLeave controls our interaction with the cluster after leave.
-	// When set to false (default), a leave causes Consul to not rejoin
+	// When set to false (default), a leave causes Nomad to not rejoin
 	// the cluster until an explicit join is received. If this is set to
 	// true, we ignore the leave, and rejoin the cluster on start.
 	RejoinAfterLeave bool `hcl:"rejoin_after_leave"`
--- a/nomad/server.go
+++ b/nomad/server.go
@@ -1531,7 +1531,6 @@ func (s *Server) setupSerf(conf *serf.Config, ch chan serf.Event, path string) (
 			return nil, err
 		}
 	}
-	conf.RejoinAfterLeave = true
 	// LeavePropagateDelay is used to make sure broadcasted leave intents propagate
 	// This value was tuned using https://www.serf.io/docs/internals/simulator.html to
 	// allow for convergence in 99.9% of nodes in a 10 node cluster
--- a/website/content/docs/upgrade/upgrade-specific.mdx
+++ b/website/content/docs/upgrade/upgrade-specific.mdx
@@ -62,8 +62,6 @@ from the Nomad client by setting [`set_environment_variables`][artifact_env].
 The use of filesystem isolation can be disabled in Client configuration by
 setting [`disable_filesystem_isolation`][artifact_fs_isolation].
 
-## Nomad 1.4.4
-
 #### Server `rejoin_after_leave` (default: `false`) now enforced
 
 All Nomad versions prior to v1.4.4 have incorrectly ignored the Server [`rejoin_after_leave`][rejoin_after_leave]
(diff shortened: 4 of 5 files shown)


command/agent/agent.go
```diff
@@ -309,6 +309,8 @@ func convertServerConfig(agentConfig *Config) (*nomad.Config, error) {
 	conf.RPCAddr.IP = rpcAddr.IP
 	conf.SerfConfig.MemberlistConfig.BindPort = serfAddr.Port
 	conf.SerfConfig.MemberlistConfig.BindAddr = serfAddr.IP.String()
+	conf.SerfConfig.RejoinAfterLeave = agentConfig.Server.RejoinAfterLeave
+	// todo: surely there are other SerfConfig fields missing and this probably shouldn't be so heavily drilled-through
```
anth0d, line 313: this is more a note to self that should be squashed out of the commit

command/agent/config.go
```diff
@@ -547,7 +547,7 @@ type ServerConfig struct {
 	RetryIntervalHCL string `hcl:"retry_interval" json:"-"`
 
 	// RejoinAfterLeave controls our interaction with the cluster after leave.
-	// When set to false (default), a leave causes Consul to not rejoin
+	// When set to false (default), a leave causes Nomad to not rejoin
```
tgross, line 550: I see you've noticed how much of the really old code was copy-pasted from Consul :blush:

command/agent/agent.go
tgross: I think we're ok here -- most of the fields don't have Nomad-specific overrides, so we're starting from the default config value.

nomad/server.go
```diff
@@ -1531,7 +1531,6 @@ func (s *Server) setupSerf(conf *serf.Config, ch chan serf.Event, path string) (
 			return nil, err
 		}
 	}
-	conf.RejoinAfterLeave = true
```
tgross, line 1534: Oof, looks like this was left in from the original implementation: https://github.com/hashicorp/nomad/blame/main/nomad/server.go#L1534

tgross (commented):
Hi @anth0d this looks great! A few remaining items:

* If you run `make cl` that'll create a changelog entry file (it should read something like "Fixed a bug where the rejoin_after_leave configuration was not respected).
* This does seem like the sort of thing that could catch out users, so maybe we add a note to the version-specific upgrade guide for Nomad 1.4.4 as well. That'll be found in https://github.com/hashicorp/nomad/blob/main/website/content/docs/upgrade/upgrade-specific.mdx
* For some reason GHA didn't run on this PR, so we haven't run most of the CI testing on it. I'll try to figure that out but if you have any suggestions I'm all ears :grin:

tgross:
@anth0d I _think_ we've got the CI issue fixed. Would you mind rebasing on `main` to verify that?

website/content/docs/upgrade/upgrade-specific.mdx
```diff
@@ -62,6 +62,18 @@ from the Nomad client by setting [`set_environment_variables`][artifact_env].
 The use of filesystem isolation can be disabled in Client configuration by
 setting [`disable_filesystem_isolation`][artifact_fs_isolation].
 
+#### Server `rejoin_after_leave` (default: `false`) now enforced
+
+All Nomad versions prior to v1.4.4 have incorrectly ignored the Server [`rejoin_after_leave`][rejoin_after_leave]
```
tgross, line 67: This link is broken. See the preview at https://nomad-a6lg3nb9m-hashicorp.vercel.app/nomad/docs/upgrade/upgrade-specific#server-rejoin_after_leave-default-false-now-enforced.

The best way to fix this is to have this line be:

```suggestion
All Nomad versions prior to v1.4.4 have incorrectly ignored the Server [`rejoin_after_leave`][]
```

And then add the following line to the foot of this file:

```
[`rejoin_after_leave`]: /docs/configuration/server#rejoin_after_leave
```

tgross (changes requested):
One last docs item to fix and this will be good-to-go @anth0d!

anth0d:
Thanks @tgross, could you authorize the new preview app deployment?

website/content/docs/upgrade/upgrade-specific.mdx
```diff
@@ -62,6 +62,18 @@ from the Nomad client by setting [`set_environment_variables`][artifact_env].
 The use of filesystem isolation can be disabled in Client configuration by
 setting [`disable_filesystem_isolation`][artifact_fs_isolation].
 
+#### Server `rejoin_after_leave` (default: `false`) now enforced
+
+All Nomad versions prior to v1.4.4 have incorrectly ignored the Server [`rejoin_after_leave`]
+configuration option. This bug has been fixed in Nomad version v1.4.4.
+
+Previous to v1.4.4 the behavior of Nomad `rejoin_after_leave` was always `true`, regardless of 
+Nomad server configuration, while the documentation incorrectly indicated a default of `false`.
```
tgross, line 71: Gah, sorry my eyes passed right over the version number here. This should be 1.5.0 now too:

```suggestion
All Nomad versions prior to v1.5.0 have incorrectly ignored the Server [`rejoin_after_leave`]
configuration option. This bug has been fixed in Nomad version v1.5.0.

Previous to v1.5.0 the behavior of Nomad `rejoin_after_leave` was always `true`, regardless of 
Nomad server configuration, while the documentation incorrectly indicated a default of `false`.
```

anth0d: Good catch!

tgross (approved):
LGTM! Thanks for sticking this out @anth0d!

Preview link for docs: https://nomad-1wafiobcv-hashicorp.vercel.app/nomad/docs/upgrade/upgrade-specific#server-rejoin_after_leave-default-false-now-enforced

PR #15554, opened 2022-12-15
cli: add ACL binding rule commands for CRUD actions.

jrasell:
Related #13120

### DIFF
--- a/command/acl_binding_rule_list.go
+++ b/command/acl_binding_rule_list.go
@@ -1,0 +1,124 @@
+package command
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/nomad/api"
+	"github.com/mitchellh/cli"
+	"github.com/posener/complete"
+)
+
+// Ensure ACLBindingRuleListCommand satisfies the cli.Command interface.
+var _ cli.Command = &ACLBindingRuleListCommand{}
+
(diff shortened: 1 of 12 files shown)


command/acl_binding_rule_list.go
```diff
@@ -0,0 +1,124 @@
...
+	json bool
+	tmpl string
+}
+
+// Help satisfies the cli.Command Help function.
+func (a *ACLBindingRuleListCommand) Help() string {
+	helpText := `
+Usage: nomad acl binding-rule list [options]
+
+  List is used to list existing ACL binding rules. Requires a management token.
```
pkazmierczak, line 28: Just a clarification question: why do we require mgmt token for this? Wouldn't it be useful to let users list the stubs without a token? Or is there a security risk I'm not seeing?

pkazmierczak (approved):
LGTM!

command/acl_binding_rule_list.go
jrasell: I believe keeping it to management is the correct way. I don't believe users would want to see the binding rules, what is more important sometimes is seeing the policies and roles assigned to the generated token which they have the functionality to do.

The binding rule will also include information sent from the OIDC provider, which the user will likely have no access nor information about such as which Auth0 permissions they are granted. I think it's therefore fine to leave this as it is both for UX and security.

PR #15555, opened 2022-12-15
core: add ACL binding rule to replication system.

jrasell:
ACL binding rule create and deletes are always forwarded to the authoritative region. In order to make these available in federated regions, the leaders in these regions need to replicate from the authoritative.

Related #13120

### DIFF
--- a/nomad/leader.go
+++ b/nomad/leader.go
@@ -423,6 +423,7 @@ func (s *Server) establishLeadership(stopCh chan struct{}) error {
 			go s.replicateACLTokens(stopCh)
 			go s.replicateACLRoles(stopCh)
 			go s.replicateACLAuthMethods(stopCh)
+			go s.replicateACLBindingRules(stopCh)
 			go s.replicateNamespaces(stopCh)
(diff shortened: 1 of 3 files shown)


nomad/leader.go
```diff
@@ -2170,6 +2171,202 @@ func diffACLAuthMethods(
...
+
+			// Update the minimum query index, blocks until there is a change.
+			req.MinQueryIndex = resp.Index
+		}
+	}
+}
+
+// diffACLAuthMethods is used to perform a two-way diff between the local ACL
+// auth-methods and the remote auth-methods to determine which ones need to be
+// deleted or updated. The returned array's contain ACL auth-method names.
```
pkazmierczak, line 2316:
```suggestion
// diffACLBindingRules is used to perform a two-way diff between the local ACL
// binding rules and the remote binding rules to determine which ones need to be
// deleted or updated. The returned array's contain ACL binding rule names.
```

pkazmierczak (approved):
Looks good!

PR #15557, opened 2022-12-15
[ui] Make the not-auth'd messages in the app less token-centric

philrenaud:
Resolves https://github.com/hashicorp/nomad/issues/15556

Signed out, have auth-methods configured:
image

Signed out, no auth-methods:
image

Signed in, but lacking permissions: 
image

### DIFF
--- a/ui/app/templates/components/forbidden-message.hbs
+++ b/ui/app/templates/components/forbidden-message.hbs
@@ -2,27 +2,38 @@
   <h3 data-test-error-title class="empty-message-headline">Not Authorized</h3>
   <p data-test-error-message class="empty-message-body">
     {{#if this.token.secret}}
-      Your <LinkTo @route="settings.tokens">ACL token</LinkTo> does not provide the
+      You currently lack the
       {{#if this.permission}}
         <code>{{this.permission}}</code>
       {{else}}
         required
       {{/if}}
-      permission. Contact your administrator if this is an error.
+      <LinkTo @route="settings.tokens">permission</LinkTo> for this resource.<br /> Contact your administrator if this is an error.
     {{else}}
-      Provide an <LinkTo @route="settings.tokens">ACL token</LinkTo> with the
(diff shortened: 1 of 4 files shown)


ui/app/templates/components/forbidden-message.hbs
```diff
@@ -2,27 +2,38 @@
...
-      Provide an <LinkTo @route="settings.tokens">ACL token</LinkTo> with the
+      {{#if this.authMethods}}
+        Sign in with
+        {{#each this.authMethods as |authMethod|}}
+          <LinkTo @route="settings.tokens">{{authMethod.name}}</LinkTo>, 
+        {{/each}}
+        or
+      {{/if}}
+
+      {{#if this.authMethods}}p{{else}}P{{/if}}rovide a <LinkTo @route="settings.tokens">token</LinkTo> with the
```
philrenaud, line 21: There's gotta be a better way to conditionally capitalize something in handlebars. If you can think of anything, please let me know!

ChaiWithJai: Why not use a template helper?

ChaiWithJai:
```
{{#if this.authMethods}}
  {{capitalize "provide"}}
{{/else}}
  provide
{{/if}}
```

You could extract this into its own helper to simplify down to one line.

ChaiWithJai (approved):

ui/app/templates/components/forbidden-message.hbs
philrenaud: Thanks for the helper suggestion. Was trying to avoid this if I could in hope of a built-in.

Ended up making a `conditionally-capitalize` helper to make this [even simpler](https://github.com/hashicorp/nomad/pull/15557/commits/59381ecbe273d1bc6d6dbb825c190550f66aee8c)

PR #15558, opened 2022-12-15
Return buffered text from log endpoint if decoding fails

benbuzbee:
To see why I think this is a good change lets look at why I am making it

My disk was full, which means GC was happening agressively. So by the time I called the logging endpoint from the SDK, the logs were GC'd

The error I was getting before was:
```
invalid character 'i' in literal false (expecting 'l')
```

Now the error I get is:
```
failed to decode log endpoint response as JSON: "failed to list entries: open /tmp/nomad.data.4219353875/alloc/f11fee50-2b66-a7a2-d3ec-8442cb3d557a/alloc/logs: no such file or directory"
```

Still not super descriptive but much more debugable

### DIFF
--- a/.changelog/15558.txt
+++ b/.changelog/15558.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+api: improved error returned from AllocFS.Logs when response is not JSON
+```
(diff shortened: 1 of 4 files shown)


jrasell (approved):
LGTM, thanks @benbuzbee!

PR #15571, opened 2022-12-16
Expanding nomad job init to accept a template flag

### LINKED ISSUE #15563: Extend `nomad job init` to accept a `-template` flag
- Throw an error if variables are not enabled, or are disallowed by one's token
- Throw an error if the template in question is not found
- If a variable at `nomad/job-templates/${passed-template-value}` is found, write that out to example.nomad

philrenaud:
## Adds a new `-template` flag to `nomad job init`.
image

---

## Adds a new `-list-templates` flag to `nomad job init`
reads from all variables at `nomad/job-templates/...` and lists them out. 
image

---

## UI: Custom view
- changes the interface when editing a variable at `nomad/job-templates/...` to have an interface that's less free-form, and specifically supports only editing description and template.
image

Resolves #15563

### DIFF
--- a/command/job_init.go
+++ b/command/job_init.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/hashicorp/nomad/api"
 	"github.com/posener/complete"
 )
 
@@ -60,11 +61,13 @@ func (c *JobInitCommand) Name() string { return "job init" }
 func (c *JobInitCommand) Run(args []string) int {
 	var short bool
 	var connect bool
+	var template string
 
 	flags := c.Meta.FlagSet(c.Name(), FlagSetClient)
 	flags.Usage = func() { c.Ui.Output(c.Help()) }
 	flags.BoolVar(&short, "short", false, "")
 	flags.BoolVar(&connect, "connect", false, "")
+	flags.StringVar(&template, "template", "", "The name of the job template variable to initialize")
 
 	if err := flags.Parse(args); err != nil {
 		return 1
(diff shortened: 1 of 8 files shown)


command/job_init.go
```diff
@@ -36,6 +37,12 @@ Init Options:
 
   -connect
     If the connect flag is set, the jobspec includes Consul Connect integration.
+
+  -template
+    Specifies a predefined template to initialize. Must be a Nomad Variable that lives at nomad/job-templates/<template>
+
+  -list-templates
+    Display a list of possible job templates to pass to -template. Reads from all variables pathed at nomad/job-templates/<template>
```
tgross, line 45: Unfortunately we don't have automatic wrapping to the terminal for these, so let's hard-wrap them to something reasonable like 80 char (or aligned with the rest of the help block, at least)

command/job_init.go
```diff
@@ -90,27 +101,82 @@ func (c *JobInitCommand) Run(args []string) int {
...
+	if listTemplates {
+		// Get the HTTP client
+		client, err := c.Meta.Client()
+		qo := &api.QueryOptions{
+			Namespace: c.Meta.namespace,
+		}
+		// Get and list all variables at nomad/job-templates
+		vars, _, err := client.Variables().PrefixList("nomad/job-templates/", qo)
+		if err != nil {
+			c.Ui.Error(fmt.Sprintf("Error retrieving vars: %s", err))
```
tgross, line 125: Very nitpicky nitpick: it doesn't seem obvious from the command why we'd even be retrieving variables here. Maybe this should say that we got an error trying to retrieve templates from the server and where they're stored?

command/job_init.go
```diff
@@ -90,27 +101,82 @@ func (c *JobInitCommand) Run(args []string) int {
...
+			return 1
+		}
+
+		if len(vars) == 0 {
+			c.Ui.Error(fmt.Sprintf("No variables in nomad/job-templates"))
+			return 1
+		} else {
+			c.Ui.Output(fmt.Sprintf("Use nomad job init -template=<template> with any of the following:"))
+			for _, v := range vars {
+				fmt.Println(strings.TrimPrefix(v.Path, "nomad/job-templates/"))
```
tgross, line 135: We should use `c.Ui.Output` here so that we know it's all going thru that interface. Also, do we want to indent this slightly (like 2 spaces?) to make it stand out?

command/job_init.go
```diff
@@ -90,27 +101,82 @@ func (c *JobInitCommand) Run(args []string) int {
...
+		}
+		// If the user provided an item key, return that value instead of the whole
+		// object
+		if v, ok := sv.Items["template"]; ok {
+			c.Ui.Output(fmt.Sprintf("Initializing a job template from %s", template))
+			jobSpec = []byte(v)
+		} else {
+			c.Ui.Error(fmt.Sprintf("Variable does not contain %q item", args[1]))
+			return 1
+		}
```
tgross, line 167: We don't `return 0` at the end of this `else if` block: doesn't that mean we end up continuing to the `WriteFile` section below and writing an empty file? I might consider breaking out the final `else` as well so that it's in the same scope as `WriteFile`.

command/job_init.go
```diff
@@ -90,27 +101,87 @@ func (c *JobInitCommand) Run(args []string) int {
...
+			c.Ui.Error(fmt.Sprintf("Error retrieving variable: %s", err))
+			return 1
+		}
+		// If the user provided an item key, return that value instead of the whole
+		// object
+		if v, ok := sv.Items["template"]; ok {
+			c.Ui.Output(fmt.Sprintf("Initializing a job template from %s", template))
+			jobSpec = []byte(v)
+		} else {
+			c.Ui.Error(fmt.Sprintf("Variable does not contain %q item", args[1]))
```
tgross, line 165: This error message suggests we're looking for the command argument but we're actually looking for the field `template`, right? I think we want to tell the user that the variable that stores the template is malformed and missing a `template` field. Maybe even refer them to the UI page with the editor?

philrenaud: Good call, I hastily copied this from our more open-ended get variable code.

command/job_init.go
philrenaud: Just above this, we have

```
		if v, ok := sv.Items["template"]; ok {
			c.Ui.Output(fmt.Sprintf("Initializing a job template from %s", template))
			jobSpec = []byte(v)
```

where `jobSpec` is used by WriteFile below.

Should I still consider breaking the `else` out with this in mind?

tgross: Oh I missed that admidst all the error handling we have to do here. This looks fine then

tgross (approved):
LGTM!

ChaiWithJai (approved):
LGTM

command/job_init_test.go
```diff
@@ -86,6 +88,88 @@ func TestInitCommand_defaultJob(t *testing.T) {
...
+	// Doesnt work if our var lacks a template key
+	must.Eq(t, 1, jobCmd.Run([]string{"-address=" + url, "-template=invalid-template"}))
+
+	// Works if the file doesn't exist
+	must.Eq(t, 0, jobCmd.Run([]string{"-address=" + url, "-template=valid-template"}))
+
+	content, err := ioutil.ReadFile(DefaultInitName)
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
```
tgross, line 163:
```suggestion
	must.NoError(t, err)
```

command/job_init_test.go
```diff
@@ -86,6 +88,88 @@ func TestInitCommand_defaultJob(t *testing.T) {
...
+	if string(content) != string(tinyJob) {
+		t.Fatalf("unexpected file content\n\n%s", string(content))
+	}
```
tgross, line 166:
```suggestion
	must.NotEq(t, string(content), string(tinyJob))
```

PR #15580, opened 2022-12-19
acl: modify update endpoints behavior

pkazmierczak:
API and RPC endpoints for `ACLAuthMethod`s and `ACLBindingRule`s should allow users to send incomplete objects in order to, e.g., update single fields. This PR provides "merging" functionality for these endpoints. 

Relates to #13120

### DIFF
--- a/nomad/acl_endpoint.go
+++ b/nomad/acl_endpoint.go
@@ -1716,7 +1716,15 @@ func (a *ACL) UpsertAuthMethods(
 	}
 
 	// Validate each auth method, canonicalize, and compute hash
+	// merge methods in case we're doing an update
 	for idx, authMethod := range args.AuthMethods {
+		// if there's an existing method with the same name, we treat this as
+		// an update
+		existingMethod, _ := stateSnapshot.GetACLAuthMethodByName(nil, authMethod.Name)
+		if existingMethod != nil {
+			authMethod.Merge(existingMethod)
+		}
+
 		if err := authMethod.Validate(
 			a.srv.config.ACLTokenMinExpirationTTL,
 			a.srv.config.ACLTokenMaxExpirationTTL); err != nil {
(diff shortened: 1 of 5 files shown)


nomad/acl_endpoint.go
```diff
@@ -1716,7 +1716,15 @@ func (a *ACL) UpsertAuthMethods(
...
 
 	// Validate each auth method, canonicalize, and compute hash
+	// merge methods in case we're doing an update
 	for idx, authMethod := range args.AuthMethods {
+		// if there's an existing method with the same name, we treat this as
+		// an update
+		existingMethod, _ := stateSnapshot.GetACLAuthMethodByName(nil, authMethod.Name)
+		if existingMethod != nil {
+			authMethod.Merge(existingMethod)
+		}
```
shoenig, line 1726: I'd think about moving this if-case into the `Merge` implementation - as it is, it's kind of a leaky abstraction

shoenig (approved):
LGTM; any chance we can add a test case that stresses the changes to `UpsertAuthMethods` ?

shoenig:
Oooh actually looks like those test failures might be related, e.g.

`TestACL_UpsertBindingRules`

pkazmierczak:
> ~Oooh actually looks like those test failures might be related~
> 
> Well maybe related, but they're also broken on main https://github.com/hashicorp/nomad/actions/runs/3733911824/jobs/6335331833

oof that's bad. I'll fix.

nomad/acl_endpoint.go
```diff
@@ -1716,7 +1716,15 @@ func (a *ACL) UpsertAuthMethods(
 	}
 
 	// Validate each auth method, canonicalize, and compute hash
+	// merge methods in case we're doing an update
 	for idx, authMethod := range args.AuthMethods {
+		// if there's an existing method with the same name, we treat this as
+		// an update
+		existingMethod, _ := stateSnapshot.GetACLAuthMethodByName(nil, authMethod.Name)
```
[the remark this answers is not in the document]
pkazmierczak: `GetACLAuthMethodByName` will only error when it can't find the existing method, what matters to us is that it will always return nil if it can't find the method.

nomad/acl_endpoint.go
pkazmierczak: Good point.

PR #15581, opened 2022-12-19
acl: binding rules API documentation

pkazmierczak:
Relates to #13120

### DIFF
--- a/website/content/api-docs/acl/binding-rules.mdx
+++ b/website/content/api-docs/acl/binding-rules.mdx
@@ -122,29 +122,7 @@ queries](/api-docs#blocking-queries) and [required ACLs](/api-docs#acls).
 
 ```json
 {
-  "Name": "example-acl-binding-rule",
-  "Type": "OIDC",
-  "Tokenlocality": "global",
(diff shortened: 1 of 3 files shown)


philrenaud (approved):
Docs look good and cohesive to me!

2 comments, less about this particular doc and more about our ACL docs in general:
- Most of our API documentation starts with the main index GET (would be GET acl/binding-rules in this case), and then list other methods. Any reason why we oughtn't standardize this here?
- As best as I can tell, all API docs outside of /acl list `Path`s starting with `/v1`, but our ACL docs do not. Is there any good reason why there's an inconsistency?

pkazmierczak:
> * Most of our API documentation starts with the main index GET (would be GET acl/binding-rules in this case), and then list other methods. Any reason why we oughtn't standardize this here?

Good point, I'll re-order. 

> * As best as I can tell, all API docs outside of /acl list `Path`s starting with `/v1`, but our ACL docs do not. Is there any good reason why there's an inconsistency?

Wow I didn't notice this before 😬 
Let's align with James when he's back and I can make another PR that updates all of ACL docs.

PR #15582, opened 2022-12-19
RadioCard component for Job Templates

ChaiWithJai:
This PR creates a Job Template Card component to be used in the new `jobs/run/templates` view.

Side Effect:  this PR installs `@hashicorp/design-system-components` as a package and updates the configuration of our build process for `ember-cli`.

Preview
image


ui/app/components/job-template-card.hbs
```diff
@@ -0,0 +1,6 @@
+<Hds::Form::RadioCard data-test-template-card={{@label}} @checked={{@selected}} {{on "change" @onChange}} as |R|>
+    {{#if @icon}}<R.Icon @name={{@icon}} />{{/if}}
+    <R.Label data-test-template-label>{{@label}}</R.Label>
+    {{#if @badge}}<R.Badge @text={{@badge}} />{{/if}}
```
ChaiWithJai, line 4: Removing this one line should patch our problem in our build process. However, we'll need to address this naming collision challenge later.

ui/app/styles/components.scss
```diff
@@ -1,12 +1,12 @@
 @import './components/accordion';
-@import './components/badge';
+@import './components/badge-nomad-internal';
```
ChaiWithJai, line 2: @Dhaulagiri Is this a sufficient way to get around naming collision we chatted about in Slack? I clicked around our application and everything appears to be just fine.

Dhaulagiri: yes, as long as it isn't exactly `components/badge` you should avoid the conflict

ChaiWithJai: Awesome! cc:  @philrenaud

philrenaud (commented):
Code, tests, and naming changes all look good to me!

My only concern here is the (considerable) asset file size increase that comes from this PR. nomad-ui.css increases about 30% with just this PR, for example. vendor-js about +10%. That's a non-trivial impact on our users and therefore worth digging into alternatives.

Is it possible to import a lighter footprint here? Add only the needed design system component rather than the full suite?

philrenaud (approved):

PR #15585, opened 2022-12-20
Broken link

fmaste:
/docs/concepts/variablesr -> /docs/concepts/variables


jrasell:
Hi @fmaste and thanks for raising this. We already merged https://github.com/hashicorp/nomad/pull/15737 which fixed this problem, so will close this. Thanks again.

Issue #15575, opened 2022-12-17
test panic in TestStopCommand_JSON


shoenig:
```
=== RUN   TestStopCommand_JSON
    job_stop_test.go:68: [DEBUG] waiting for job to register; status code non zero saw 2
--- FAIL: TestStopCommand_JSON (1.17s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xa9 pc=0x22180ba]

goroutine 25102 [running]:
testing.tRunner.func1.2({0x275d440, 0x47d75d0})
	/opt/hostedtoolcache/go/1.19.3/x64/src/testing/testing.go:1396 +0x24e
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.19.3/x64/src/testing/testing.go:1399 +0x39f
panic({0x275d440, 0x47d75d0})
	/opt/hostedtoolcache/go/1.19.3/x64/src/runtime/panic.go:884 +0x212
github.com/hashicorp/nomad/command/agent.(*TestAgent).Shutdown(0x52d546?)
	/home/runner/work/nomad/nomad/command/agent/testagent.go:265 +0x3a
runtime.Goexit()
	/opt/hostedtoolcache/go/1.19.3/x64/src/runtime/panic.go:522 +0x185
testing.(*common).FailNow(0xc0070c0340)
	/opt/hostedtoolcache/go/1.19.3/x64/src/testing/testing.go:871 +0x4c
testing.(*common).Fatalf(0xc0070c0340, {0x2db6ea4?, 0x0?}, {0xc00338fe50?, 0x530294?, 0xc00235c750?})
	/opt/hostedtoolcache/go/1.19.3/x64/src/testing/testing.go:955 +0x65
github.com/hashicorp/nomad/command.TestStopCommand_JSON(0xc0070c0340)
	/home/runner/work/nomad/nomad/command/job_stop_test.go:68 +0x539
testing.tRunner(0xc0070c0340, 0x2e1fc38)
	/opt/hostedtoolcache/go/1.19.3/x64/src/testing/testing.go:1446 +0x10b
created by testing.(*T).Run
	/opt/hostedtoolcache/go/1.19.3/x64/src/testing/testing.go:1493 +0x35f
FAIL command.TestStopCommand_JSON (1.17s)
FAIL command
```

[...]

PR #15595, opened 2022-12-20
tests: do not return error from testagent shutdown

shoenig:
This PR changes the signature of `TestAgent.Shutdown` to return nothing, making it directly compatible with `defer`. We usually ignore the error even though we shouldn't, so instead now we call Fatal if the agent fails to shutdown or times out while doing so.

Extracting this cleanup out of fixing https://github.com/hashicorp/nomad/issues/15575


pkazmierczak (approved):
lgtm!

Issue #15546, opened 2022-12-14
Add "Upload File" and "Choose from Template" buttons to the existing Run Job route


ChaiWithJai:

tgross:
Closed by #15596

PR #15596, opened 2022-12-20
ui: add choose template button

ChaiWithJai:
Resolves [#15546](https://github.com/hashicorp/nomad/issues/15546).

This PR creates a link (button) in the `jobs.run` view to allow a user to select a template to populate the job definition.

Current Design:
image

Figma Mock (has an `a11y` contrast violation):
image

Side Effects and notes:

- This PR updates the file directory to have a `jobs/run` and `jobs/run/templates` subdirectories.
- This PR populates dummy markup for acceptance tests.
- This PR doesn't make full use of styling in the Figma mocks because of a11y violations and waiting for #15582 to be merged in for appropriate component consumption.
- The original Figma mock (screenshot above) includes an `a11y` color contrast violation. To circumvent this issue we should default to using `primary` and `secondary` buttons given to us from `HDS`. This can be addressed in the subsequent PR.


ui/app/templates/components/job-editor.hbs
```diff
@@ -22,10 +22,23 @@
...
+      {{#let (cannot "write variable" path="*" namespace="*") as |editVar|}}
+        <button
+          data-test-choose-template
+          type="button"
+          class="button is-outline is-small {{if editVar "tooltip"}}"
+          disabled={{if editVar 'disabled'}}
+          aria-label={{if editVar "You don’t have permission to choose templates."}}
+          {{action "goToJobTemplates"}}>
+          <span>Choose from a template</span>
+        </button>
```
philrenaud, line 39: I'd like to request that this be an `<a>` rather than a `<button>`, since its action is only routing. Would remove the need to modify the `job-editor` component, and would allow native browser behaviour (like right click -> open in new tab) where the proposed button here does not.

ui/app/templates/components/job-editor.hbs
```diff
@@ -22,10 +22,23 @@
...
-    <label class="job-spec-upload">
-      <input type="file" onchange={{action this.uploadJobSpec}} accept=".hcl,.json,.nomad" />
-      <span class="button">Upload a job spec file</span>
-    </label>
+    <div>
+      <label class="job-spec-upload">
+        <input type="file" onchange={{action this.uploadJobSpec}} accept=".hcl,.json,.nomad" />
+        <span class="button is-outline is-small">Upload file</span>
+      </label>
+      {{#let (cannot "write variable" path="*" namespace="*") as |editVar|}}
```
philrenaud, line 30: `editVar` is perhaps a misnomer; I would expect that this implies I _can_  edit the variable but it seems that just the opposite is true.

ui/app/router.js
```diff
@@ -14,7 +14,13 @@ Router.map(function () {
...
 
   this.route('jobs', function () {
-    this.route('run');
+    this.route('run', function () {
+      this.route('templates', function () {
+        this.route('new');
+        this.route('manage');
+        this.route('template', { path: '/:name' });
+      });
+    });
```
philrenaud, line 23: Routes look good!

philrenaud (changes requested):

ui/app/templates/components/job-editor.hbs
ChaiWithJai: The ticket did say, "Choose from Template requires `can edit` variable ability". So I tried to match that. What would you suggest?

ui/app/templates/components/job-editor.hbs
ChaiWithJai: I agree. Let's follow the `HDS` [pattern](https://design-system-components-hashicorp.vercel.app/components/button) of using a `Button` link because this is a CTA.

ChaiWithJai: However, I'm just using "button" to match the existing style until #15582 is merged. I think they have it all figured out.

ui/app/templates/components/job-editor.hbs
philrenaud: we're naming "The inability to write a variable" as "editVar", so at the very most verbose, something like "cannotEditVar" would be more correct. I think you probably just meant this to be inverted.

PR #15600, opened 2022-12-20
[CI only] Prepare workflow rollout

mdeggies:
This PR rolls out the new `prepare` CI workflow, which encapsulates all of the centralized artifact processing jobs that run after artifacts are built. Instead of explicitly listing out each job that should run for Nomad, artifact processing jobs run based on a set of organizational rules (e.g. fossa scanning will run for all repos) and the presence of artifacts (e.g. if darwin artifacts have been built for nomad, they will be notarized). More info on this simplification of the release configuration file is available here: https://hashicorp.atlassian.net/wiki/spaces/RELENG/pages/2489712686/Dec+7th+2022+-+Introducing+the+new+Prepare+workflow


tgross (approved):
LGTM

.release/ci.hcl
```diff
@@ -18,159 +18,23 @@ project "nomad" {
...
-event "build" {
-  depends = ["merge"]
-
-  action "build" {
-    organization = "hashicorp"
-    repository   = "nomad"
-    workflow     = "build"
-  }
-}
-
```
shore, line 34: Removing the `build` event?  The new `prepare` event explicitly depends on `build`.

mdeggies: This is why code reviews are important. 😆  
I was/am still planning to test this branch after the prepare workflow is green again. But thanks for catching this ahead of time!!

.release/ci.hcl
```diff
@@ -18,159 +18,31 @@ project "nomad" {
...
-event "notarize-darwin-amd64" {
-  depends = ["security-scan-binaries"]
-
-  action "notarize-darwin-amd64" {
+  action "prepare" {
     organization = "hashicorp"
     repository   = "crt-workflows-common"
-    workflow     = "notarize-darwin-amd64"
+    workflow     = "prepare"
+    depends      = ["build"]
```
shore, line 36: Does this belong in the action stanza?  The event already declares this dependency, and I don't see any other examples of `depends` within `action`.

mdeggies: Yeah, this is the way it's been configured in all other projects: https://github.com/hashicorp/crt-core-helloworld/blob/main/.release/ci.hcl#L45, https://github.com/hashicorp/tfc-agent/blob/main/.release/ci.hcl#L34, https://github.com/hashicorp/consul/blob/main/.release/ci.hcl#L33. I can't tell you the exact reason for this, though :-D Scott would likely know!

Issue #15562, opened 2022-12-16
parse `stale` querystring parameter value as boolean


tgross:
The [`stale`](https://developer.hashicorp.com/nomad/api-docs#stale) parameter on the Nomad HTTP API determines the consistency mode for query endpoints. If the `stale` flag is set, a request to a follower will be served without being forwarded to the leader. Unfortunately the API server tolerates this flag have a value, but it only _acts_ on the flag being set.

This means a request like `/v1/jobs?stale=false` is treated as a stale query. While the documentation is straightforward on this, it's bad ergonomics and can lead to mistakes. Instead, we should parse any value as a boolean (and keep the existing behavior where if the flag is set without a value, the query is stale). So in other words we want:

| querystring | is stale query? |
|-|-|
| `?stale` | yes |
| `?stale=true` | yes |
| `?stale=false` | no |

dttung2905:
If you don't mind, please allow me to work on this issue.

PR #15605, opened 2022-12-21
fix: `stale` querystring parameter value as boolean

dttung2905:
Hi Nomad core dev team,

Please let me know if there are other places I need to make changes too. Thank you !

Resolve: #15562

### DIFF
--- a/command/agent/http.go
+++ b/command/agent/http.go
@@ -783,10 +783,17 @@ func parseWait(resp http.ResponseWriter, req *http.Request, b *structs.QueryOpti
 }
 
 // parseConsistency is used to parse the ?stale query params.
-func parseConsistency(req *http.Request, b *structs.QueryOptions) {
+func parseConsistency(resp http.ResponseWriter, req *http.Request, b *structs.QueryOptions) {
 	query := req.URL.Query()
-	if _, ok := query["stale"]; ok {
-		b.AllowStale = true
+	if staleVal, ok := query["stale"]; ok {
+		if staleVal[0] == "true" || staleVal[0] == "" {
+			b.AllowStale = true
+		} else if staleVal[0] == "false" {
+			// fall through
+		} else {
+			resp.WriteHeader(400)
+			resp.Write([]byte(fmt.Sprintf("Expect `true` or `false` for `stale` query string parameter, got %s", staleVal[0])))
+		}
 	}
 }
 
@@ -884,7 +891,7 @@ func (s *HTTPServer) parseToken(req *http.Request, token *string) {
 func (s *HTTPServer) parse(resp http.ResponseWriter, req *http.Request, r *string, b *structs.QueryOptions) bool {
 	s.parseRegion(req, r)
 	s.parseToken(req, &b.AuthToken)
-	parseConsistency(req, b)
+	parseConsistency(resp, req, b)
 	parsePrefix(req, b)
 	parseNamespace(req, &b.Namespace)
 	parsePagination(req, b)
--- a/command/agent/http.go
+++ b/command/agent/http.go
@@ -792,11 +792,7 @@ func parseConsistency(resp http.ResponseWriter, req *http.Request, b *structs.Qu
 			resp.Write([]byte(fmt.Sprintf("Expect `true` or `false` for `stale` query string parameter, got %s", staleVal[0])))
 		}
 
-		if staleQuery || staleVal[0] == "" {
-			b.AllowStale = true
-		} else {
-			b.AllowStale = false
-		}
+		b.AllowStale = staleQuery || staleVal[0] == ""
 	}
 }
 
--- a/command/agent/http.go
+++ b/command/agent/http.go
@@ -794,6 +794,8 @@ func parseConsistency(resp http.ResponseWriter, req *http.Request, b *structs.Qu
 
 		if staleQuery || staleVal[0] == "" {
 			b.AllowStale = true
+		} else {
+			b.AllowStale = false
 		}
 	}
 }
--- a/command/agent/http.go
+++ b/command/agent/http.go
@@ -786,13 +786,14 @@ func parseWait(resp http.ResponseWriter, req *http.Request, b *structs.QueryOpti
 func parseConsistency(resp http.ResponseWriter, req *http.Request, b *structs.QueryOptions) {
 	query := req.URL.Query()
 	if staleVal, ok := query["stale"]; ok {
-		if staleVal[0] == "true" || staleVal[0] == "" {
-			b.AllowStale = true
-		} else if staleVal[0] == "false" {
-			// fall through
-		} else {
+		staleQuery, err := strconv.ParseBool(staleVal[0])
+		if err != nil {
 			resp.WriteHeader(400)
 			resp.Write([]byte(fmt.Sprintf("Expect `true` or `false` for `stale` query string parameter, got %s", staleVal[0])))
+		}
+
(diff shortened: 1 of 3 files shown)


command/agent/http.go
```diff
@@ -783,10 +783,17 @@ func parseWait(resp http.ResponseWriter, req *http.Request, b *structs.QueryOpti
 }
 
 // parseConsistency is used to parse the ?stale query params.
-func parseConsistency(req *http.Request, b *structs.QueryOptions) {
+func parseConsistency(resp http.ResponseWriter, req *http.Request, b *structs.QueryOptions) {
 	query := req.URL.Query()
-	if _, ok := query["stale"]; ok {
-		b.AllowStale = true
+	if staleVal, ok := query["stale"]; ok {
+		if staleVal[0] == "true" || staleVal[0] == "" {
```
shoenig, line 789: Instead of requiring `"true"` or empty string, what if we use `strconv.ParseBool` here; that will help with compatibility with callers using other languages, etc.

command/agent/http.go
```diff
@@ -783,10 +783,17 @@ func parseWait(resp http.ResponseWriter, req *http.Request, b *structs.QueryOpti
...
+			b.AllowStale = true
+		} else if staleVal[0] == "false" {
+			// fall through
```
shoenig, line 792: What if we re-organize these cases as a switch statement? Comments about control flow are usually a good sign there's a better pattern to be found

command/agent/http_test.go
```diff
@@ -476,16 +476,31 @@ func TestParseWait_InvalidIndex(t *testing.T) {
 func TestParseConsistency(t *testing.T) {
 	ci.Parallel(t)
 	var b structs.QueryOptions
+	var resp *httptest.ResponseRecorder
+
+	testCases := [3]string{"/v1/catalog/nodes?stale", "/v1/catalog/nodes?stale=true", "/v1/catalog/nodes?stale=false"}
+	for _, url := range testCases {
+		req, err := http.NewRequest("GET",
+			url, nil)
```
shoenig, line 484: no need for the line break here is there?

command/agent/http_test.go
```diff
@@ -476,16 +476,31 @@ func TestParseWait_InvalidIndex(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("err: %v", err)
+		}
```
shoenig, line 487: The surrounding tests are quite old, but lately we've been migrating to github.com/shoenig/test for assertions; 

```
must.NoError(t, err)
```

command/agent/http_test.go
```diff
@@ -476,16 +476,31 @@ func TestParseWait_InvalidIndex(t *testing.T) {
...
+		resp = httptest.NewRecorder()
+		parseConsistency(resp, req, &b)
+		if !b.AllowStale {
+			t.Fatalf("Bad: %v", b)
+		}
```
shoenig, line 492: similar,

```
must.True(t, b.AllowStale)
```

command/agent/http_test.go
```diff
@@ -476,16 +476,31 @@ func TestParseWait_InvalidIndex(t *testing.T) {
...
 	}
-
-	parseConsistency(req, &b)
-	if !b.AllowStale {
-		t.Fatalf("Bad: %v", b)
+	resp = httptest.NewRecorder()
+	parseConsistency(resp, req, &b)
+	if resp.Code != 400 {
+		t.Fatalf("bad code: %v", resp.Code)
 	}
```
shoenig, line 504: same

shoenig (changes requested):
This is looking great, @dttung2905! I left some comments mostly about style, but functionally this is good.

Since this is a bug fix, we'll need to also be sure to add a changelog entry. You can type `make cl` to start a small wizard that helps create one.

command/agent/http.go
dttung2905: Thanks you for the suggestion and the super quick review. I have a question though. If it is empty string `strconv.ParseBool` will return `false` as well right :thinking: Just want to confirm on my understanding

shoenig: It will return false yeah, so we'll need to first explicitly check for empty string and interpret that as true

command/agent/http.go
```diff
@@ -783,10 +783,20 @@ func parseWait(resp http.ResponseWriter, req *http.Request, b *structs.QueryOpti
...
+		if err != nil {
+			resp.WriteHeader(400)
+			resp.Write([]byte(fmt.Sprintf("Expect `true` or `false` for `stale` query string parameter, got %s", staleVal[0])))
+		}
+
+		if staleQuery || staleVal[0] == "" {
+			b.AllowStale = true
+		} else {
+			b.AllowStale = false
+		}
```
shoenig, line 799: can collapse this to one line, 

```
b.AllowStale = staleQuery || stalVal[0] == "" 
```

command/agent/http_test.go
```diff
@@ -476,29 +477,36 @@ func TestParseWait_InvalidIndex(t *testing.T) {
...
 
-	parseConsistency(req, &b)
-	if !b.AllowStale {
-		t.Fatalf("Bad: %v", b)
-	}
+	req, err := http.NewRequest("GET", "/v1/catalog/nodes?stale=false", nil)
+	must.NoError(t, err)
+	resp = httptest.NewRecorder()
+	parseConsistency(resp, req, &b)
+	must.True(t, !b.AllowStale)
```
shoenig, line 495: can use `must.False`

command/agent/http_test.go
```diff
@@ -476,29 +477,36 @@ func TestParseWait_InvalidIndex(t *testing.T) {
...
+
+	req, err = http.NewRequest("GET", "/v1/catalog/nodes?stale=random", nil)
+	must.NoError(t, err)
+	resp = httptest.NewRecorder()
+	parseConsistency(resp, req, &b)
+	must.EqOp(t, resp.Code, 400)
```
shoenig, line 501: The expectation (the `400` here) goes first

(it only matters when rendering failure output, but still it's nice to avoid causing confusion)

command/agent/http_test.go
```diff
@@ -476,29 +477,36 @@ func TestParseWait_InvalidIndex(t *testing.T) {
...
+	req, err = http.NewRequest("GET", "/v1/catalog/nodes?consistent", nil)
+	must.NoError(t, err)
 
-	parseConsistency(req, &b)
-	if b.AllowStale {
-		t.Fatalf("Bad: %v", b)
-	}
+	resp = httptest.NewRecorder()
+	parseConsistency(resp, req, &b)
+	must.True(t, !b.AllowStale)
```
shoenig, line 509: also `.False`

shoenig (changes requested):
Thanks @dttung2905, just a few more nitpicks and then we'll get this merged :slightly_smiling_face:

command/agent/http_test.go
dttung2905: Do you mean we should add in extra assertion like this to provide consistency to the test case? 
```
	must.False(t, b.AllowStale)
	must.EqOp(t, resp.Code, 400)

```

shoenig: Oh sorry, just mean you have the order of the arguments backwards for EqOp(t, expectation, value): https://pkg.go.dev/github.com/shoenig/test/must#EqOp

should be `EqOp(t, 400, resp.Code)`

shoenig (approved):

Issue #15589, opened 2022-12-20
tests: swap freeport for portal 


shoenig:
One driver of test failures / flakiness is in how `freeport` can regularly be exhausted of ports. When a test waits for a port to become free, it may hit a timeout and fail the test even though it never got a chance to run.

A key change in 5de8e0a153c8658fb983aefdb30b07c26acad901 is that `api/` now uses this [portal](https://github.com/shoenig/test/tree/v0.5.2/portal) package for port allocations. Seems to be working well in the `api/` test suite, so I'd like to try converting our other tests to use it as well.

e.g.
```
[WARN] freeport: 16 out of 16 pending ports are still in use; something probably didn't wait around for the port to be closed!
<< repeat 100's of times >>

event_endpoint_test.go:302: 
     Error Trace:	/home/runner/work/nomad/nomad/command/agent/event_endpoint_test.go:302
        	            /home/runner/work/nomad/nomad/command/agent/http_test.go:1457
        	            /home/runner/work/nomad/nomad/command/agent/event_endpoint_test.go:211
     Error:      	Expected value not to be nil.
     Test:       	TestHTTP_Alloc_Port_Response
```

PR #15609, opened 2022-12-22
cleanup: remove usage of consul/sdk/testutil/retry

shoenig:
This PR removes usages of `consul/sdk/testutil/retry`, as part of the
ongoing effort to remove use of any non-API module from Consul.

There is one remanining usage in the helper/freeport package, but that
will get removed as part of #15589


pkazmierczak (approved):
LGTM!

PR #15611, opened 2022-12-22
Disallows variables at exactly nomad/job-templates

philrenaud:
Realized the existing code would let you write to `nomad/job-templates` with no suffix. This disallows writing there but allows you to write to `nomad/job-templates/foo`.

In so doing, refactored the switch statement for path name error handling.


ChaiWithJai (approved):
LGTM.

PR #15612, opened 2022-12-22
ui:  update `VariableSerializer` normalize method

ChaiWithJai:
The `VariableSerializer` currently does not save key/value pairs when querying multiple `Variable` records. This PR updates the `normalize` method to use the same logic in the `serialize` method to extract the `key/value` attributes form the `Items` property on `Variable` so that we can use that logic in the `jobs.run.templates.index` view where we need to visualize this data.

image


philrenaud (approved):
Looks good to me, but can you confirm that we're not planning on reading key/value data on all variables at variables index/path level? Only on the job templates page, and on a specific variable itself? We don't want to load that data unless specifically needed.

PR #15613, opened 2022-12-22
ui: select template

### LINKED ISSUE #15547: [UI, Example Job Templates] New route: jobs/run/templates/index
image

ChaiWithJai:
Closes #15547


ui/mirage/config.js
```diff
@@ -8,6 +8,9 @@ import { copy } from 'ember-copy';
...
...602:bb72:1cd1:5d2c\"\n              }\n            ]\n          },\n          \"Resources\": {\n            \"CPU\": 2500,\n            \"MemoryMB\": 2048,\n            \"MemoryMaxMB\": 0,\n            \"DiskMB\": 0\n          },\n          \"Driver\": \"exec\",\n          \"Name\": \"task-capacitor-6\",\n          \"JobID\": \"\",\n          \"VolumeMounts\": [\n            {\n              \"Volume\": \"mazie\",\n              \"Destination\": \"/Willy_Hane/vicente/#562154\",\n              \"PropagationMode\": \"\",\n              \"ReadOnly\": false\n            }\n          ],\n          \"GroupNames\": [],\n          \"WithServices\": false,\n          \"CreateRecommendations\": false,\n          \"ID\": \"7\",\n          \"Services\": []\n        }\n      ]\n    }\n  ],\n  \"JobSummary\": {\n    \"GroupNames\": [\n      \"bandwidth-g-4\"\n    ],\n    \"Summary\": {\n      \"bandwidth-g-4\": {\n        \"Queued\": 4,\n        \"Complete\": 4,\n        \"Failed\": 7,\n        \"Running\": 8,\n        \"Starting\": 3,\n        \"Lost\": 4,\n        \"Unknown\": 2\n      }\n    },\n    \"Namespace\": \"default\",\n    \"ID\": \"4\",\n    \"JobID\": \"jbod-capacitor-3\"\n  }\n}`;
```
ChaiWithJai, line 12: TODO:  create a mirage scenario once we know the schema for `template`.

ui/app/routes/jobs/run/templates/index.js
```diff
@@ -5,21 +5,28 @@ export default class RunTemplatesRoute extends Route {
...
-    ) {
+  beforeModel(transition) {
+    const hasPermissions = this.can.can('write variable', null, {
+      namespace: '*',
+      path: '*',
+    });
+
+    // We create a job with no id in jobs.run that is populated by this form.
+    // A user cannot start at this route.
+    const wasJobModelCreated = transition.from?.name === 'jobs.run.index';
```
ChaiWithJai, line 16: `note`: Previously the `jobs.run` route created an empty `job` without an `id` upon entering the route and there were two behaviors:

1. The page acted like a form, that upon hitting `Plan` it would set the `id`, `namespace` and `plainId` and the newly created `job`.
2. Upon exiting the page, it would destroy the `model`.

Now, we're creating a multi-page form there's more nuance here. 

Challenge:  We're creating a `job` without an `id` or any queryable attributes and then we navigate to a brand new page that loses the context of the `route`.

- How do we preserve this state?
- How do we ensure proper clean-up and maintenance of our `ember-data` store?

These are architectural questions of this sub-route and not addressed in this PR. The simple solution would be to recognize that we're in a multi-step form and only allow the user to enter the form via `jobs.run` (the beginning of the form). We currently don't support page refresh behavior to preserve draft state in our current version of the `JobEditor` (this is a pretty frustrating challenge that would be worth solving while we are here).

Nomad currently has 4 forms and this is our first multi-page form.

I think we want to think about this problem architecturally and consider creating a `FormService` that would be responsible for creating/updating the new entity because trying to manage this across the 4 pages in this view could make it challenging for us to keep track of permissioning and the non-queryable `job` that we're in the process of creating.

I'd propose having a separate PR to handle this.

We totally could just keep this logic within our `routes` and use Ember API's to help us here which would be a little faster in the short-term, however, it could become a big ball of mud.

ui/app/serializers/variable.js
```diff
@@ -8,14 +8,6 @@ export default class VariableSerializer extends ApplicationSerializer {
...
     // ID is a composite of both the job ID and the namespace the job is in
     hash.ID = `${hash.Path}@${hash.Namespace || 'default'}`;
-
-    hash.KeyValues = Object.entries(hash.Items).map(([key, value]) => {
-      return {
-        key,
-        value,
-      };
-    });
-
```
ChaiWithJai, line 18: `note`: revert commit after Slack conversation.

ui/app/styles/core/forms.scss
```diff
@@ -89,6 +89,9 @@
   }
 }
 
+.radio-group {
+  padding: 16px 0px;
```
ChaiWithJai, line 93: `question`: Would you like to see some type of `grid` pattern here?

ui/app/routes/jobs/run/templates/index.js
```diff
@@ -5,21 +5,28 @@ export default class RunTemplatesRoute extends Route {
...
-  model() {
-    return this.store.query('variable', {
+  async model() {
+    const jobTemplateVariables = await this.store.query('variable', {
       prefix: 'nomad/job-templates',
-      filter: 'Template is not empty"',
     });
+    const recordsToQuery = jobTemplateVariables.map((template) =>
+      this.store.findRecord('variable', template.id)
+    );
```
ChaiWithJai, line 28: `note`: `O(n+1)` query that I mentioned earlier.

ui/app/routes/jobs/run/index.js
```diff
@@ -18,16 +18,26 @@ export default class RunRoute extends Route {
...
   }
 
-  model() {
+  async model() {
+    const transition = arguments[1];
+
+    if (transition.from?.name === 'jobs.run.templates.index') {
+      const plainId = transition.to.queryParams.plainId;
+      return this.store.peekAll('job').findBy('plainId', plainId);
+    }
```
ChaiWithJai, line 27: `note`: See [comment](https://github.com/hashicorp/nomad/pull/15613/files#r1055829677).

ui/app/routes/jobs/run/index.js
```diff
@@ -18,16 +18,26 @@ export default class RunRoute extends Route {
...
       return this.store.createRecord('job');
     });
   }
 
-  resetController(controller, isExiting) {
+  resetController(controller, isExiting, transition) {
     if (isExiting) {
+      if (transition.to.name === 'jobs.run.templates.index') {
+        return;
+      }
```
ChaiWithJai, line 40: `note`: See [comment](https://github.com/hashicorp/nomad/pull/15613/files#r1055829677).

ui/app/controllers/jobs/run/templates/index.js
```diff
@@ -0,0 +1,52 @@
...
+  @action
+  onApply() {
+    const job = getOwner(this).lookup('controller:jobs.run.index').model;
+    const { json } = this.templates?.find(
+      (template) => template.id === this.selectedTemplate
+    );
+
+    job.set('_newDefinition', json);
+    job.setIdByPayload(json);
+    this.navigateToJobsRun(job.plainId);
```
ChaiWithJai, line 50: `note`: See [comment](https://github.com/hashicorp/nomad/pull/15613/files#r1055829677).

ui/app/controllers/jobs/run/templates/index.js
```diff
@@ -0,0 +1,52 @@
...
+      })?.value;
+
+      return {
+        id: templateVariable.id,
+        label: templateVariable.path.split('nomad/job-templates/')[1],
+        description,
+        json,
+      };
+    });
+  }
```
ChaiWithJai, line 29: `question`: Is the schema correct here?

ui/app/routes/jobs/run/index.js
```diff
@@ -1,34 +1,59 @@
...
+      if (template) {
+        const templateRecord = await this.store.findRecord(
+          'variable',
+          template
+        );
+
+        return this.store.createRecord('job', {
+          _newDefinition: templateRecord.items.template,
+        });
+      }
```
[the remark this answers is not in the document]
philrenaud: And a question: should we nullify the queryParam for template once it's loaded? Once loaded, its job is pretty much done, right? Better to have a clean URL?

ui/app/templates/jobs/run/templates/index.hbs
```diff
@@ -1,10 +1,26 @@
...
-        <button data-test-apply>PLACEHOLDER USE HDS</button>
-        <button data-test-cancel>PLACEHOLDER USE HDS</button>
-    </div>
+    <header class="run-job-header">
+        <h1 class="title is-3">Choose a template</h1>
+        <p>A plan will be requested before the job is submitted.  You can pick a template or write your own job spec.</p>
+    </header>
+    <main class="radio-group" data-test-template-list>
+        <Hds::Form::RadioCard::Group @alignment="left" as |G|>
+            <G.Legend>Select a Template</G.Legend>
```
philrenaud, line 8: Let's move "Select a template" into a conditional for if templates exist

image

ui/app/templates/jobs/run/templates/index.hbs
```diff
@@ -1,10 +1,26 @@
...
+                </G.RadioCard>
+            {{/each}}
+        </Hds::Form::RadioCard::Group>
+    </main>
+    {{#if (eq this.templates.length 0)}}
+        <h3 data-test-empty-jobs-list-headline class="empty-message-headline">
+          You have no templates to choose from. Would you like to <Hds::Link::Inline @route="jobs.run.templates.new" data-test-create-inline>create</Hds::Link::Inline> one?
+        </h3>
+    {{/if}}
+    <Hds::ButtonSet>
```
philrenaud, line 22: Let's add some padding/margin to the top of this. It looks pretty cramped to either the bottom of your templates, or to the "You have no templates" message.

ui/tests/acceptance/job-run-test.js
```diff
@@ -158,9 +158,9 @@ module('Acceptance | job run', function (hooks) {
...
           {
             prefix: 'nomad/job-templates',
-            filter: 'Template is not empty"',
+            namespace: '*',
           },
-          'It makes a request to the /vars endpoint with the appropriate filter expression for job templates.'
+          'It makes a request to the /vars endpoint with the appropriate query parameters for job templates.'
         );
         return [];
       });
```
philrenaud, line 166: Follow-up item request: add tests for "When variable created at nomad/job-templates/, shows up on the run/templates page", ditto delete, description change.

ui/app/templates/jobs/run/templates/index.hbs
```diff
@@ -1,10 +1,26 @@
...
-    <div>
-        <button data-test-apply>PLACEHOLDER USE HDS</button>
-        <button data-test-cancel>PLACEHOLDER USE HDS</button>
-    </div>
+    <header class="run-job-header">
+        <h1 class="title is-3">Choose a template</h1>
+        <p>A plan will be requested before the job is submitted.  You can pick a template or write your own job spec.</p>
+    </header>
+    <main class="radio-group" data-test-template-list>
+        <Hds::Form::RadioCard::Group @alignment="left" as |G|>
```
philrenaud, line 7: Seems to be a bug with this radio card group implementation: I'm able to select multiple: 
image

ui/app/templates/jobs/run/templates/index.hbs
```diff
@@ -1,10 +1,26 @@
...
+            <G.Legend>Select a Template</G.Legend>
+            {{#each this.templates as |card|}}
+                <G.RadioCard @layout="fixed" @maxWidth="30%" id={{card.id}} data-test-template-card={{card.label}} {{on "change" this.onChange}} as |R|>
```
philrenaud, line 10: layout + maxWidth seems to work well for the design here. Nice work!

philrenaud (changes requested):
The "Can select multiple" is the only significant blocker that I can see, but I've made some other suggestions that I think would make this shippable.

ui/app/templates/jobs/run/templates/index.hbs
```diff
@@ -1,10 +1,26 @@
...
+                    <R.Label data-test-template-label>{{card.label}}</R.Label>
+                    <R.Description data-test-template-description>{{card.description}}</R.Description>
+                </G.RadioCard>
+            {{/each}}
+        </Hds::Form::RadioCard::Group>
+    </main>
+    {{#if (eq this.templates.length 0)}}
+        <h3 data-test-empty-jobs-list-headline class="empty-message-headline">
+          You have no templates to choose from. Would you like to <Hds::Link::Inline @route="jobs.run.templates.new" data-test-create-inline>create</Hds::Link::Inline> one?
```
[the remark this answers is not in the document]
ChaiWithJai: We wouldn't link internally using the `a href`. We would use the Ember default component of `LinkTo` because it accepts a `route` `query` and `model` arguments that we can forward to our component. And for styling sake, sticking with `HDS` components when possible is a good convention IMO. The component API for `Hds::Link::Inline` is [better](https://design-system-components-hashicorp.vercel.app/components/link/inline) for our needs than `LinkTo`.

ui/tests/acceptance/job-run-test.js
ChaiWithJai: This page isn't real-time. So we're really just testing given an input which is the URL, that we get a certain output and trigger the correct side-effect (dispatch a network request).

I can create some variables and ensure that those variables do not appear in the list, but that's already covered by ensuring that we dispatch a network request with the correct query parameters.

The tests you're describing sound like its to ensure that the Ember Data store isn't serving us a stale cache. So I'm a little confused about these specifications.

ChaiWithJai: Ember Data, the Ember Route and the Nomad API are implementation details and basically the blackbox of this view. But I can gladly test these details here if you think its best.

ui/app/routes/jobs/run/index.js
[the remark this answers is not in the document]
ChaiWithJai: At this layer, the `error` object that is returned is an assertion message (`string`) and a stack trace.

If we want to trigger a toast notification upon receiving a `404` error we'll want to update the `handleResponse` method on the `Variable Adapter` which would have ramifications for the `Variable` views who aren't handling `404`'s like this. It might just cause a slight inconsistency with error message handling but I'll handle it there and we can do some manual testing and see what happens.

```
message
: 
"Assertion Failed: You made a 'findRecord' request for a 'variable' with id 'Auto-conflicting Varible@default', but the adapter's response did not have any data"
stack
: 
"Error: Assertion Failed: You made a 'findRecord' request for a 'variable' with id 'Auto-conflicting Varible@default', but the adapter's re....
```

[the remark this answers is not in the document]
ChaiWithJai: > Once loaded, its job is pretty much done, right? Better to have a clean URL?

IMO, we care about URL's because the contract the user has with the browser. If the user hits refresh on the page, they'll get a different output if we clear the query parameter. And it'll impact the back/forward buttons on the browser.

I'd probably keep the query parameter.

ui/app/templates/jobs/run/templates/index.hbs
ChaiWithJai: Patched.

ui/app/templates/jobs/run/templates/index.hbs
ChaiWithJai: Is this what you're looking for

image

ui/app/routes/jobs/run/index.js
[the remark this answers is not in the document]
ChaiWithJai: image

ui/app/templates/jobs/run/templates/index.hbs
philrenaud: That looks great!

philrenaud (approved):
Looks great! Thanks for all the updates.

Issue #15087, opened 2022-10-31
agent syslog always runs at NOTICE level


tgross:
The rarely-used [`enable_syslog`](https://developer.hashicorp.com/nomad/docs/configuration#enable_syslog) configuration emits Nomad agent logs to a syslog interface. While debugging a problem for an internal user, I noticed that they were only getting high-level logs. When we configure syslog at [`command.go#L478`](https://github.com/hashicorp/nomad/blob/v1.4.2/command/agent/command.go#L478) we set the log-level to NOTICE, which means even if the agent is set to log at debug, only NOTICE-level and below will actually get emitted on the syslog interface.

(Note this is entirely unrelated to any allocation logs, which are covered under other open issues.)

sharmaansh21:
Hi @tgross Could I pick this issue?

jrasell:
Hi @sharmaansh21; we do not have a formal way to assign this issue to you, but we are not currently working on it internally, so please feel free to pick this issue up. If you have any questions, please do not hesitate to reach out to us via this issue. Thanks for showing interest!

tgross:
Fixed by #15625 and will ship in the next regular release of Nomad.

PR #15625, opened 2022-12-25
fix: Make agent syslog log level inherit from Nomad agent log

dttung2905:
Fix issue : #15087

Hi Nomad Core Dev team,

Below is my attempt to fix the issue mentioned above. Imo, we can either make an extra argument for syslog log level or make it inherit from Nomad Agent existing one. This PR is using the latter approach because there is not need to use the former for a rarely used feature.

Would appreciate getting your feedback on this PR, especially on a few issues:
- On the mapping between gsyslog and Nomad Agent log, I'm not entirely sure about how `OFF` and `TRACE` in Nomad log_level maps to [gsyslog](https://pkg.go.dev/log/syslog) log level
- I attempted to write one similar to [TestCommand_Args()](https://github.com/hashicorp/nomad/blob/v1.4.3/command/agent/command_test.go#L26-L98) but the log is not captured by either `ui.ErrorWriter` or `ui.OutputWriter`. Please let me know if a unit test is needed or there are better ways to structure the unit test

### DIFF
--- a/.changelog/15625.txt
+++ b/.changelog/15625.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-fix: Make agent syslog log level follow log_level config
+agent: Make agent syslog log level follow log_level config
 ```
--- a/.changelog/15625.txt
+++ b/.changelog/15625.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+fix: Make agent syslog log level follow log_level config
+```
--- a/command/agent/command.go
+++ b/command/agent/command.go
@@ -473,12 +473,14 @@ func SetupLoggers(ui cli.Ui, config *Config) (*logutils.LevelFilter, *gatedwrite
 	// Create a log writer, and wrap a logOutput around it
 	writers := []io.Writer{logFilter}
 	logLevelMap := map[string]gsyslog.Priority{
-		"OFF":   gsyslog.LOG_EMERG,
 		"ERROR": gsyslog.LOG_ERR,
 		"WARN":  gsyslog.LOG_WARNING,
 		"INFO":  gsyslog.LOG_INFO,
 		"DEBUG": gsyslog.LOG_DEBUG,
 		"TRACE": gsyslog.LOG_DEBUG,
+	}
+	if config.LogLevel == "OFF" {
+		config.EnableSyslog = false
 	}
 	// Check if syslog is enabled
 	if config.EnableSyslog {
--- a/command/agent/command.go
+++ b/command/agent/command.go
@@ -472,10 +472,18 @@ func SetupLoggers(ui cli.Ui, config *Config) (*logutils.LevelFilter, *gatedwrite
 
 	// Create a log writer, and wrap a logOutput around it
 	writers := []io.Writer{logFilter}
-
+	logLevelMap := map[string]gsyslog.Priority{
+		"OFF":   gsyslog.LOG_EMERG,
+		"ERROR": gsyslog.LOG_ERR,
+		"WARN":  gsyslog.LOG_WARNING,
+		"INFO":  gsyslog.LOG_INFO,
+		"DEBUG": gsyslog.LOG_DEBUG,
+		"TRACE": gsyslog.LOG_DEBUG,
+	}
 	// Check if syslog is enabled
 	if config.EnableSyslog {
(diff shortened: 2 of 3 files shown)


.changelog/15625.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+fix: Make agent syslog log level follow log_level config
```
tgross, line 2:
```suggestion
agent: Make agent syslog log level follow log_level config
```

command/agent/command.go
```diff
@@ -472,10 +472,18 @@ func SetupLoggers(ui cli.Ui, config *Config) (*logutils.LevelFilter, *gatedwrite
 
 	// Create a log writer, and wrap a logOutput around it
 	writers := []io.Writer{logFilter}
-
+	logLevelMap := map[string]gsyslog.Priority{
+		"OFF":   gsyslog.LOG_EMERG,
```
tgross, line 476: When the log level is `"OFF"` the user is saying they don't want logs at all (which is intended only for testing but it's an option). Maybe if `config.LogLevel` is `"OFF"` we can treat that exactly as if `config.EnableSyslog` is false?

tgross (changes requested):
This looks great @dttung2905! I agree that it makes the best sense to have the syslog logger inherit the log level from the agent. I've left a suggestion for how to treat the `"OFF"` level, but I think what you've done for `"TRACE"` is fine.

Thanks for the PR!

command/agent/command.go
dttung2905: Yes. It makes sense to me too. Let me update the code

command/agent/command.go
```diff
@@ -472,10 +472,20 @@ func SetupLoggers(ui cli.Ui, config *Config) (*logutils.LevelFilter, *gatedwrite
 
 	// Create a log writer, and wrap a logOutput around it
 	writers := []io.Writer{logFilter}
-
```
tgross, line 475: I was double-checking this before merging and noticed that `config.LogLevel` isn't coerced to upper-case until it's consumed (see line 464 above). It's coerced to a hclog enum in a different function in the agent code, so canonicalizing the case at the point we initially handle it is going to have a bunch of ripple effects. So let's canonicalize it here for use in this function:

```suggestion
		logLevel = strings.ToUpper(config.LogLevel)
```

command/agent/command.go
```diff
@@ -472,10 +472,20 @@ func SetupLoggers(ui cli.Ui, config *Config) (*logutils.LevelFilter, *gatedwrite
...
+	logLevelMap := map[string]gsyslog.Priority{
+		"ERROR": gsyslog.LOG_ERR,
+		"WARN":  gsyslog.LOG_WARNING,
+		"INFO":  gsyslog.LOG_INFO,
+		"DEBUG": gsyslog.LOG_DEBUG,
+		"TRACE": gsyslog.LOG_DEBUG,
+	}
+	if config.LogLevel == "OFF" {
```
tgross, line 482:
```suggestion
	if logLevel == "OFF" {
```

command/agent/command.go
```diff
@@ -472,10 +472,20 @@ func SetupLoggers(ui cli.Ui, config *Config) (*logutils.LevelFilter, *gatedwrite
...
+		config.EnableSyslog = false
+	}
 	// Check if syslog is enabled
 	if config.EnableSyslog {
-		l, err := gsyslog.NewLogger(gsyslog.LOG_NOTICE, config.SyslogFacility, "nomad")
+		ui.Output(fmt.Sprintf("Config enable_syslog is `true` with log_level=%v", config.LogLevel))
+		l, err := gsyslog.NewLogger(logLevelMap[config.LogLevel], config.SyslogFacility, "nomad")
```
tgross, line 488:
```suggestion
		l, err := gsyslog.NewLogger(logLevelMap[logLevel], config.SyslogFacility, "nomad")
```

tgross (changes requested):
Looking good! I found one minor issue around case handling that became obvious once we checked for `OFF`. Once that's fixed I think this is ready for merge.

command/agent/command.go
dttung2905: I have made changes according to your suggestion. Thanks for pointing it out !

tgross (approved):
LGTM! Thanks!

PR #15627, opened 2022-12-26
api: purge testify and pretty dependencies 

shoenig:
This PR refactors API tests to no longer depend on testify or pretty.

### DIFF
--- a/api/csi_test.go
+++ b/api/csi_test.go
@@ -4,7 +4,7 @@ import (
 	"testing"
 
 	"github.com/hashicorp/nomad/api/internal/testutil"
-	"github.com/stretchr/testify/require"
+	"github.com/shoenig/test/must"
 )
 
 // TestCSIVolumes_CRUD fails because of a combination of removing the job to plugin creation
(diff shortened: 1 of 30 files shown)


api/csi_test.go
```diff
@@ -14,89 +14,94 @@ import (
...
 	defer s.Stop()
 	v := c.CSIVolumes()
 
 	// Successful empty result
 	vols, qm, err := v.List(nil)
-	require.NoError(t, err)
-	require.NotEqual(t, 0, qm.LastIndex)
-	require.Equal(t, 0, len(vols))
+	must.NoError(t, err)
+	// must.Positive(t, qm.LastIndex) TODO(tgross), this was always broken?
```
tgross, line 25: The `require.NotEqual(t, 0, qm.LastIndex)` test above suggests it was working... but is it returning a negative value?

Also, I don't recall what the story with the `// FIXME` section is below, but I can take a look after the holidays. No need to block this PR on that though.

tgross (approved):
LGTM!

PR #15667, opened 2023-01-03
[doc] nomad `eval delete` example typo fix

Dgotlieb:
Status instead of Stauts


tgross (approved):
LGTM, thanks @Dgotlieb !

Issue #15663, opened 2023-01-03
Unable to delete external volume because of special character in secret key/value pair


digi-talo:
### Nomad version
Output from `v1.4.3`

### Operating system and Environment details
Debian buster -  3 node dev environment running both server and client mode

### Issue

Can't delete an external storage volumes because the `adminKey` contains an `=` sign.

### Reproduction steps

`nomad volume delete -secret adminID=csiAdmin -secret adminKey=xxxxxxxxxxxxxxxxxx== vol-test`

Using quotes around the key=value renders the same result. There seems to be no way to have the value correctly interpreted when it contains an equal sign.
When I remove the equal signs from the adminKey I get a rados authentication error as expected.

#### Expected Result

Deleted volume

#### Actual Result

`Secret must be in the format: -secret key=value`

tgross:
Hi @digi-talo! This looks like a bug in the command line parsing of the k/v format. See [`volume_delete.go#L105-L114`](https://github.com/hashicorp/nomad/blob/v1.4.3/command/volume_delete.go#L105-L114) but it looks like we have the exact same bug in `volume snapshot create`, `volume snapshot list`, and `volume snapshot delete` as well.

The fix for this should be straightforward: we need to grab only the first `=` and not the rest of them. I'll try to get a patch up shortly for this.

PR #15670, opened 2023-01-03
csi: Fix parsing of '=' in secrets at command line and HTTP

tgross:
Fixes https://github.com/hashicorp/nomad/issues/15663

The command line flag parsing and the HTTP header parsing for CSI secrets incorrectly split at more than one '=' rune, making it impossible to use secrets that included that rune.

### DIFF
--- a/command/agent/csi_endpoint.go
+++ b/command/agent/csi_endpoint.go
@@ -409,7 +409,7 @@ func parseCSISecrets(req *http.Request) structs.CSISecrets {
 	secrets := map[string]string{}
 	secretkvs := strings.Split(secretsHeader, ",")
 	for _, secretkv := range secretkvs {
-		kv := strings.Split(secretkv, "=")
+		kv := strings.SplitN(secretkv, "=", 2)
 		if len(kv) == 2 {
 			secrets[kv[0]] = kv[1]
 		}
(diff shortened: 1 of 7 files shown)


command/agent/csi_endpoint.go
```diff
@@ -409,7 +409,7 @@ func parseCSISecrets(req *http.Request) structs.CSISecrets {
 	secrets := map[string]string{}
 	secretkvs := strings.Split(secretsHeader, ",")
 	for _, secretkv := range secretkvs {
-		kv := strings.Split(secretkv, "=")
+		kv := strings.SplitN(secretkv, "=", 2)
 		if len(kv) == 2 {
 			secrets[kv[0]] = kv[1]
 		}
```
angrycub, line 415: What about strings.Cut? The docs for SplitN suggest it for the case of splitting at the first separator. Might be a little easier to read?

```suggestion
		if key, value, found := strings.Cut(secretkv, "="); found {
			secrets[key] = value
		}
```

angrycub (approved):
Just a suggestion about an alternative implementation, but nothing that prevents this from going as is.

command/agent/csi_endpoint.go
tgross: Good catch, will fix

shoenig: TIL, nice!

Issue #15671, opened 2023-01-03
testflake: TestJobs_Evaluations got wrong assertion during refactoring


shoenig:
previous:
```
// Check that we got the evals back, evals are in order most recent to least recent
// so the last eval is the original registered eval
idx := len(evals) - 1
if n := len(evals); n == 0 || evals[idx].ID != resp.EvalID {
  t.Fatalf("expected >= 1 eval (%s), got: %#v", resp.EvalID, evals[idx])
}
```

now:
```
must.Len(t, 1, evals)
```

Asserting exactly 1 eval is too strong of an assertion.

PR #15672, opened 2023-01-03
tests: fix assertion for slice length

shoenig:
This assertions got borked during the refactoring; should be at least one element, not exactly one element.

Closes https://github.com/hashicorp/nomad/issues/15671


tgross (approved):
LGTM

Issue #15662, opened 2023-01-03
testflake: TestAllocations_List_Resources fails semi-regularly 


shoenig:
e.g.

```
=== FAIL: . TestAllocations_List_Resources (re-run 2) (5.90s)
==> Loaded configuration from /tmp/nomad3907227128/nomad1777016878
==> Starting Nomad agent...
==> Nomad agent configuration:

       Advertise Addrs: HTTP: 127.0.0.1:40577; RPC: 127.0.0.1:44193; Serf: 127.0.0.1:43323
            Bind Addrs: HTTP: [127.0.0.1:40577]; RPC: 127.0.0.1:44193; Serf: 127.0.0.1:43323
                Client: true
             Log Level: OFF
                Region: global (DC: dc1)
                Server: true
               Version: 1.4.4-dev

==> Nomad agent started! Log data will stream in below:

make: *** [GNUmakefile:294: test-nomad-module] Error 1
    assert.go:24: 
        allocations_test.go:127: expected equality via cmp.Equal function
        ↪ Assertion | differential ↷
          strings.Join({
        - 	"3767f73b-8531-ddff-a218-fc94a80c6b8e",
        + 	"aca5c7a0-3de8-d420-8aec-9a7ac101e841",
          }, "")
==> Caught signal: interrupt
```

printing out the resp and alloc objects

```
    assert.go:24: 
        allocations_test.go:127: expected equality via cmp.Equal function
        ↪ Assertion | differential ↷
          strings.Join({
        - 	"5aafb593-2184-4429-967e-8a8f6101fe37",
        + 	"1ddbd2c0-89b0-eea7-e225-1add8cdead9b",
          }, "")
        ↪ PostScript | annotation ↷
        	resp:      &api.JobRegisterResponse{EvalID:"5aafb593-2184-4429-967e-8a8f6101fe37", EvalCreateIndex:0x7, JobModifyIndex:0x7, Warnings:"", QueryMeta:api.QueryMeta{LastIndex:0x0, LastContact:0, KnownLeader:false, RequestTime:0, NextToken:""}}
        ↪ PostScript | annotation ↷
[...]

PR #15677, opened 2023-01-03
test: wait for node in api tests that register job

shoenig:
Pretty sure this is just a race between `makeClient` producing a usable node and us trying to register a job. If we lose the race, the job registration produces a queued eval rather than the eval for running the job, which has a different ID.

Adds a little more debug output to help with debugging if it still breaks.

Closes https://github.com/hashicorp/nomad/issues/15662


tgross (approved):
:+1: LGTM

PR #15682, opened 2023-01-03
Backport of test: wait for node in api tests that register job into release/1.2.x

hc-github-team-nomad-core:
## Backport

This PR is auto-generated from #15677 to be assessed for backporting due to the inclusion of the label backport/1.2.x.

---

Pretty sure this is just a race between `makeClient` producing a usable node and us trying to register a job. If we lose the race, the job registration produces a queued eval rather than the eval for running the job, which has a different ID.

Adds a little more debug output to help with debugging if it still breaks.

Closes https://github.com/hashicorp/nomad/issues/15662


shoenig:
irrelevant

PR #15695, opened 2023-01-05
docs: clarify shutdown_delay jobspec param and service behaviour.

### LINKED ISSUE #15602: Shutdown Delay should deregister a nomad service
### Nomad version
Output from `nomad version`

`Nomad v1.4.3 (f464aca721d222ae9c1f3df643b3c3aaa20e2da7)`

Ubuntu 22.04

### Issue

The `group` in the below jobspec has a `shutdown_delay` param configured to be 30s.

According to the docs [here](https://developer.hashicorp.com/nomad/docs/job-specification/group) and [here](https://developer.hashicorp.com/nomad/docs/job-specification/task#shutdown_delay), only `consul` is configured to de-register the service.

> The delay occurs between Consul deregistration and sending each task a shutdown signal
> Specifies the duration to wait when killing a task between removing it from Consul and sending it a shutdown signal

If I infer this correctly, once the task is sent a SIGINT/SIGTERM, the service will get de-registered immediately however the actual task won't be killed until `shutdown_delay` param. Is my understanding correct here? 

In that case, the same behaviour doesn't happen with `nomad` service provider.

image

### Reproduction steps

- Run the app: `nomad run deployment.nomad`
- Trigger a new deployment, since the jobspec has a `meta` variable with `uuidv4` value, re-running `nomad run deployment.nomad` would do the same.
- Watch the list of allocs and services and notice that the alloc which is given a `stop` signal, still is available in the registrations:

image

#### Expected Result

The service should be de-registered immediately.

#### Actual Result

The service isn't removed until the alloc itself is stopped. Thus, configuring `shutdown_delay` isn't helping here.

### Job file (if appropriate)

```
job "doggo" {
  datacenters = ["dc1"]
  type        = "service"
  namespace   = "default"

  group "app" {
    count          = 3
    shutdown_delay = "30s"

    # Rolling update.
    update {
      max_parallel = 1
      stagger      = "15s"
    }


    network {
      port "http" {
        to = 8080
      }
    }

    task "web" {
      driver = "docker"

      service {
        provider = "nomad"
        name     = "doggo-web"
        tags     = ["doggo", "web"]
        port     = "http"
        check {
          type     = "http"
          port     = "http"
          path     = "/api/ping/"
          interval = "5s"
          timeout  = "2s"
        }
      }

      meta {
        deployment_id = uuidv4()
      }

      config {
        image   = "ghcr.io/mr-karan/doggo-api:latest"
        ports   = ["http"]
        command = "/bin/sh"
        args    = ["-c", "sleep 10 && /app/doggo-api.bin"]
      }

      resources {
        cpu    = 200
        memory = 200
      }
    }
  }
}
```

jrasell:
Clarifies that the task and group level `shutdown_delay` parameters do not influence each other and that they apply to both Nomad and Consul service registrations.

The change also clarifies that service blocks apply to both Nomad and Consul service registrations.

Closes #15602

### DIFF
--- a/website/content/docs/job-specification/group.mdx
+++ b/website/content/docs/job-specification/group.mdx
@@ -66,20 +66,21 @@ job "docs" {
   all tasks in this group. If omitted, a default policy exists for each job
   type, which can be found in the [restart stanza documentation][restart].
 
-- `service` <code>([Service][]: nil)</code> - Specifies integrations with
-  [Consul](/docs/configuration/consul) for service discovery.
-  Nomad automatically registers each service when an allocation
-  is started and de-registers them when the allocation is destroyed.
+- `service` <code>([Service][]: nil)</code> - Specifies integrations with Nomad
+  or [Consul](/docs/configuration/consul) for service discovery. Nomad
+  automatically registers each service when an allocation is started and
+  de-registers them when the allocation is destroyed.
 
 - `shutdown_delay` `(string: "0s")` - Specifies the duration to wait when
-  stopping a group's tasks. The delay occurs between Consul deregistration
-  and sending each task a shutdown signal. Ideally, services would fail
-  healthchecks once they receive a shutdown signal. Alternatively
+  stopping a group's tasks. The delay occurs between Consul or Nomad service
+  deregistration and sending each task a shutdown signal. Ideally, services
+  would fail health checks once they receive a shutdown signal. Alternatively,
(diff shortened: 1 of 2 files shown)


shoenig (approved):
LGTM. Not to be done here, but I think we should have docs section somewhere with an elaborate example of using

> Ideally services would fail health checks
  once they receive a shutdown signal.

to control readiness checks

PR #15697, opened 2023-01-05
acl: binding rules evaluation

pkazmierczak:
`Binder` provides an interface for binding claims and ACL roles/policies of Nomad.

### DIFF
--- a/lib/auth/oidc/claims.go
+++ b/lib/auth/oidc/claims.go
@@ -13,7 +13,7 @@ import (
 
 // SelectorData returns the data for go-bexpr for selector evaluation.
 func SelectorData(
-	am *structs.ACLAuthMethod, idClaims, userClaims json.RawMessage) (map[string]interface{}, error) {
+	am *structs.ACLAuthMethod, idClaims, userClaims json.RawMessage) (*structs.ACLAuthClaims, error) {
 
 	// Extract the claims into a map[string]interface{}
 	var all map[string]interface{}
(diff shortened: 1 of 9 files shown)


jrasell (changes requested):
Looking great, thanks for this! I have made a couple of suggestions, the policy name list is the only blocker as the other items can be looked up as we see fit.

lib/auth/oidc/claims.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	list, err := extractListMappings(all, am.Config.ListClaimMappings)
+	if err != nil {
+		return nil, err
+	}
+
+	return map[string]interface{}{
+		"value": values,
+		"list":  list,
```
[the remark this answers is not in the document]
pkazmierczak: Good idea.

nomad/structs/acl.go
```diff
@@ -829,6 +829,14 @@ func (a *ACLAuthMethodConfig) Copy() *ACLAuthMethodConfig {
 	return c
 }
 
+// ACLAuthClaims is the claim mapping of the OIDC auth method in a format that
+// can be used with go-bexpr. This structure is used during rule binding
+// evaluation.
+type ACLAuthClaims struct {
+	Value any
+	List  any
```
jrasell, line 837: I think we can make these map types, so we avoid asserting it each time to grab the data.
```suggestion
	Value map[string]string{}
	List  map[string][]string{}
```

pkazmierczak: Ah right I forgot these will always be maps 🤦‍♂️

PR #15701, opened 2023-01-05
consul: add client configuration for grpc_ca_file

### LINKED ISSUE #15360: Consul Connect sidecar proxies require additional configuration for gRPC-TLS listener
### Nomad version

### Issue
Consul 1.14 introduced changes to the gRPC listener used for Envoy xDS configuration, splitting support for plaintext and TLS-enabled listeners for the protocol across two ports.

An initial fix for discovering the TLS-enabled listener via Consul's `/agent/self` endpoint landed in #15309, however when utilizing gRPC over TLS the `consul connect envoy [...]` command internally invoked by Nomad to bootstrap Envoy requires the addition of the [-grpc-ca-file](https://developer.hashicorp.com/consul/commands/connect/envoy#grpc-ca-file) or [-grpc-ca-path](https://developer.hashicorp.com/consul/commands/connect/envoy#grpc-ca-path) flags (or environment variable equivalents) to inject a trusted CA certificate into the generated Envoy config.

When the gRPC CA has not been explicitly passed, Envoy sidecar proxy tasks will initially deploy but will never retrieve their running configuration from the local Consul agent, sitting in a fail/retry loop unable to negotiate a connection with the gRPC-TLS listener..

To support this, Nomad will likely need to make a CA certificate available ~~within the allocation filesystem wherever a `sidecar_service{}` stanza has been defined,~~ on disk on each scheduler, or the Consul CLI will need to be extended to allow other means of providing the CA certificate (i.e. as an inline PEM rather than file path).

As a temporary workaround to restore proxy connectivity, the plaintext Consul HTTP and gRPC endpoints may be enabled and bound to a loopback interface within the local Consul agent, as Nomad will default to discovering the plaintext gRPC listener as long as HTTP is the URL schema in the `consul.address` configuration.

shoenig:
This combined with https://github.com/hashicorp/consul/pull/15913 should be enough to get folks unblocked in upgrading to Consul 1.14 with TLS enabled and using Connect.

Unfortunately the [breaking change on the Consul side](https://developer.hashicorp.com/consul/docs/upgrading/upgrade-specific#changes-to-grpc-tls-configuration) combined with the new config required on the Nomad side implies the upgrade path for affected folks requires reconfiguration and restarting the Consul and Nomad agents together. Further, users who now specify a CA for grpc connections distinct from the Consul HTTP CA will need to drain each node of existing Connect services making use of the Envoy proxy - otherwise existing services will begin to fail to receive xDS config changes from Consul.

Note that Nomad users will also need to either configure `consul.grpc_address` to the 8503 default tls grpc port, or modify Consul config to listen to grpc tls on port 8502, e.g.

in nomad.hcl
```
consul {
  grpc_ca_file = "/etc/tls/consul-agent-ca.pem"
  ca_file = "/etc/tls/consul-agent-ca.pem"
  cert_file = "/etc/tls/dc1-client-consul-0.pem"
  key_file = "/etc/tls/dc1-client-consul-0-key.pem"
  ssl = true
  address = "127.0.0.1:8501"
  grpc_address = "127.0.0.1:8503"
}
```

Closes #15360


jrasell (approved):
LGTM!

Do we want a changelog entry and upgrade guide note with this PR, or will this be done separately?

shoenig:
Whoops of course, yes

PR #15703, opened 2023-01-05
e2e: disable disconnected clients test(s)

shoenig:
The e2e suite is not in good shape right now; let's disable the tests that modify
agent / node state until we can get things working again. Also the one DC test
that was enabled still doesn't work anyway.


tgross (approved):
LGTM

Issue #15711, opened 2023-01-06
e2e: broken consul template test case


shoenig:
`go test -v -suite=ConsulTemplate .`

Probably something to do with 

```
&api.TaskEvent{
    Type:             "Failed Artifact Download",
    Time:             1673027676696340729,
    DisplayMessage:   "failed to download artifact \"https://google.com\": getter subprocess failed: exit status 1",
    Details:          {"download_error":"failed to download artifact \"https://google.com\": getter subprocess failed: exit status 1"},
```

cases:

```
--- FAIL: TestE2E/ConsulTemplate (307.73s)
    --- FAIL: TestE2E/ConsulTemplate/*consultemplate.ConsulTemplateTest (307.73s)
        --- PASS: TestE2E/ConsulTemplate/*consultemplate.ConsulTemplateTest/TestConsulTemplate_NomadServiceLookups (14.70s)
        --- PASS: TestE2E/ConsulTemplate/*consultemplate.ConsulTemplateTest/TestTemplatePathInterpolation_Bad (6.56s)
        --- PASS: TestE2E/ConsulTemplate/*consultemplate.ConsulTemplateTest/TestTemplatePathInterpolation_Ok (9.82s)
        --- FAIL: TestE2E/ConsulTemplate/*consultemplate.ConsulTemplateTest/TestTemplatePathInterpolation_SharedAllocDir (90.13s)
        --- FAIL: TestE2E/ConsulTemplate/*consultemplate.ConsulTemplateTest/TestTemplateUpdateTriggers (186.19s)
```

PR #15712, opened 2023-01-06
artifact: fix sandbox behavior when destination is shared alloc directory

shoenig:
This PR fixes the artifact sandbox (new in Nomad 1.5) to allow downloading
artifacts into the shared 'alloc' directory made available to each task in
a common allocation. Previously we assumed the 'alloc' dir would be mounted
under the 'task' dir, but this is only the case in fs isolation: chroot; in
other modes the alloc dir is elsewhere.

Closes https://github.com/hashicorp/nomad/issues/15711


tgross (approved):
LGTM

PR #15715, opened 2023-01-07
deps: update shoenig/test to v0.6.0

shoenig:
Adds support for custom cmp.Options; needed to fix one minor thing
causing api breakage.


tgross (approved):
LGTM!

PR #15734, opened 2023-01-09
Authenticate method improvements

tgross:
This changeset covers a sidebar discussion that @schmichael and I had around the
design for pre-forwarding auth. This includes some changes extracted out of
#15513 to make it easier to review both and leave a clean history. 

* Remove fast path for NodeID: Previously-connected clients will have a NodeID
set on the context, and because this is a large portion of the RPCs sent we
fast-pathed it at the top of the `Authenticate` method. But the context is
shared for all yamux streams over the same yamux session (and TCP
connection). This lets an authenticated HTTP request to a client use the NodeID
for authentication, which is a privilege escalation. Remove the fast path and
annotate it so that we don't break it again.

* Add context to decisions. The `Authenticate` method taken on its own looks
like it wants to return an `acl.ACL` that folds over all the various identity
types (creating an ephemeral ACL on the fly if neccessary). But keeping these
fields idependent allows RPC handlers to differentiate between internal and
external origins so we most likely want to avoid this. Leave some docstrings as
a warning as to why this is built the way it is.

* Mutate the request rather than returning a value. When reviewing #15513 we
decided that forcing the request handler to call `SetIdentity` was repetitive
and error prone. Instead, the `Authenticate` method mutates the request by
setting its `AuthenticatedIdentity`.

---

I'm marking folks other than @schmichael for review on this as he's discussed these changes with me already but isn't going to be available for a review this week, so spreading the design decisions around seems like a good idea.

### DIFF
--- a/nomad/acl.go
+++ b/nomad/acl.go
@@ -16,7 +16,7 @@ import (
 
 // Authenticate extracts an AuthenticatedIdentity from the request context or
 // provided token and sets the identity on the request. The caller can extract
-// an acl.ACL, WorkloadIdentity, or other identifying token to use for
+// an acl.ACL, WorkloadIdentity, or other identifying tokens to use for
 // authorization. Keeping these fields independent rather than merging them into
 // an ephemeral ACLToken makes the original of the credential clear to RPC
 // handlers, who may have different behavior for internal vs external origins.
--- a/nomad/acl.go
+++ b/nomad/acl.go
@@ -15,29 +15,31 @@ import (
 )
 
 // Authenticate extracts an AuthenticatedIdentity from the request context or
-// provided token. The caller can extract an acl.ACL, WorkloadIdentity, or other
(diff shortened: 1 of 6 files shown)


shoenig (approved):
LGTM

nomad/acl.go
```diff
@@ -15,32 +15,31 @@ import (
 )
 
 // Authenticate extracts an AuthenticatedIdentity from the request context or
-// provided token. The caller can extract an acl.ACL, WorkloadIdentity, or other
-// identifying token to use for authorization.
+// provided token and sets the identity on the request. The caller can extract
+// an acl.ACL, WorkloadIdentity, or other identifying token to use for
```
jrasell, line 19:
```suggestion
// an acl.ACL, WorkloadIdentity, or other identifying tokens to use for
```

jrasell (approved):
LGTM!

Issue #15548, opened 2022-12-14
[UI, Example Job Templates] New route: jobs/run/templates/new


ChaiWithJai:
![image](https://user-images.githubusercontent.com/713991/207902366-e35afcff-4bfa-48eb-9e2d-a2038692c4df.png)

tgross:
Closed by #15744

PR #15744, opened 2023-01-10
ui: create new job template

ChaiWithJai:
Closes [#15548](https://github.com/hashicorp/nomad/issues/15548)

### DIFF
--- a/ui/app/controllers/jobs/run/templates/new.js
+++ b/ui/app/controllers/jobs/run/templates/new.js
@@ -1,0 +1,71 @@
+import { A } from '@ember/array';
+import Controller from '@ember/controller';
+import { inject as service } from '@ember/service';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+const ALL_NAMESPACE_WILDCARD = '*';
+
+export default class JobsRunTemplatesController extends Controller {
+  @service router;
+  @tracked templateName = null;
+  @tracked templateNamespace = 'default';
+
+  keyValues = A([]);
+
+  get namespaceOptions() {
+    const namespaces = this.store
+      .peekAll('namespace')
+      .map(({ name }) => ({ key: name, label: name }));
+
+    if (namespaces.length <= 1) return null;
+
+    // Create default namespace selection
+    namespaces.unshift({
+      key: ALL_NAMESPACE_WILDCARD,
+      label: 'All (*)',
+    });
+
+    return namespaces;
+  }
+
+  @action
+  updateKeyValue(key, value) {
+    if (this.keyValues.find((kv) => kv.key === key)) {
+      this.keyValues.find((kv) => kv.key === key).value = value;
+    } else {
+      this.keyValues.pushObject({ key, value });
+    }
+  }
+
+  @action
+  async save(e, overwrite = false) {
+    if (e.type === 'submit') {
+      e.preventDefault();
+    }
+
+    if (this.model?.isNew) {
+      if (this.namespaceOptions) {
+        this.model.set('namespace', this.templateNamespace);
+      } else {
+        const [namespace] = this.store.peekAll('namespace').toArray();
+        this.model.set('namespace', namespace.id);
+      }
+    }
+
+    this.model.set('keyValues', this.keyValues);
+    this.model.set('path', `nomad/job-templates/${this.templateName}`);
+    this.model.setAndTrimPath();
+    await this.model.save({ adapterOptions: { overwrite } });
+
+    this.flashMessages.add({
+      title: 'Job template saved',
+      message: `${this.templateName} successfully saved`,
+      type: 'success',
+      destroyOnClick: false,
+      timeout: 5000,
+    });
+
+    this.router.transitionTo('jobs.run.templates');
+  }
+}
--- a/ui/app/routes/jobs/run/index.js
+++ b/ui/app/routes/jobs/run/index.js
@@ -69,7 +69,7 @@ export default class RunRoute extends Route {
 
   resetController(controller, isExiting) {
     if (isExiting) {
-      controller.model.deleteRecord();
+      controller.model?.deleteRecord();
       controller.set('template', null);
     }
   }
--- a/ui/app/templates/jobs/run/templates/index.hbs
+++ b/ui/app/templates/jobs/run/templates/index.hbs
@@ -3,8 +3,9 @@
         <h1 class="title is-3">Choose a template</h1>
         <p>A plan will be requested before the job is submitted.  You can pick a template or write your own job spec.</p>
     </header>
+    <Hds::Button @text="Create New Template" @route="jobs.run.templates.new" data-test-create-new-button />
     {{#if (eq this.templates.length 0)}}
-        <h3 data-test-empty-jobs-list-headline class="empty-message-headline">
+        <h3 data-test-empty-templates-list-headline class="empty-message-headline">
           You have no templates to choose from. Would you like to <Hds::Link::Inline @route="jobs.run.templates.new" data-test-create-inline>create</Hds::Link::Inline> one?
         </h3>
         <Hds::Button class="button-group" @text="Back to editor" @route="jobs.run" @icon="arrow-left" data-test-cancel />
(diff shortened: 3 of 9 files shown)


ui/app/controllers/jobs/run/templates/new.js
```diff
@@ -0,0 +1,71 @@
...
+
+    this.flashMessages.add({
+      title: 'Job template saved',
+      message: `${this.templateName} successfully saved`,
+      type: 'success',
+      destroyOnClick: false,
+      timeout: 5000,
+    });
+
+    this.router.transitionTo('jobs.run.templates');
```
ChaiWithJai, line 69: @philrenaud there's some funky behavior for how we handle deleting a record on exiting the controller that has a slight conflict with saving. I handled the problem slightly differently than you did with variables by addressing this logic in `resetController`.

ui/app/templates/jobs/run/templates/new.hbs
```diff
@@ -0,0 +1,34 @@
...
+                @selection={{this.templateNamespace}}
+                @onSelect={{fn (mut this.templateNamespace)}}
+            />
+        </div>
+        <VariableForm::JobTemplateEditor
+            @keyValues={{this.model.keyValues}}
+            @updateKeyValue={{this.updateKeyValue}}
+        />
+        <footer>
+            <Hds::Button @text="Save" @route="jobs.run.templates" disabled={{this.shouldDisableSave}} {{on "click" this.save}} data-test-save-template />
```
ChaiWithJai, line 30: `shouldDisableSave` logic hasn't been set yet.

ui/app/templates/jobs/run/templates/new.hbs
```diff
@@ -0,0 +1,34 @@
...
+                data-test-template-name
+            />
+            </label>
+            <SingleSelectDropdown
+                data-test-namespace-facet
+                @label="Namespace"
+                @options={{this.namespaceOptions}}
+                @selection={{this.templateNamespace}}
+                @onSelect={{fn (mut this.templateNamespace)}}
+            />
```
ChaiWithJai, line 23: `styling`: Needs to get onto one row.

ui/app/templates/jobs/run/templates/index.hbs
```diff
@@ -3,8 +3,9 @@
         <h1 class="title is-3">Choose a template</h1>
         <p>A plan will be requested before the job is submitted.  You can pick a template or write your own job spec.</p>
     </header>
+    <Hds::Button @text="Create New Template" @route="jobs.run.templates.new" data-test-create-new-button />
```
ChaiWithJai, line 6: `styling` needs to be inline with `header`.

ui/app/routes/jobs/run/index.js
```diff
@@ -69,7 +69,7 @@ export default class RunRoute extends Route {
 
   resetController(controller, isExiting) {
     if (isExiting) {
-      controller.model.deleteRecord();
+      controller.model?.deleteRecord();
```
ChaiWithJai, line 72: In the event that a user has a bad URL, we will throw an error on fetching the template before the job model is created. So we need some type safety here to account for this situation.

ChaiWithJai (commented):
@philrenaud This PR can successfully handle the happy path of creating and saving a new variable. However, I need some guidance re: error handling.

- Happy Path -- Should we have a confirmation message on hitting `Cancel` or navigating away on creation?
- Sad Path -- Using duplicate template name?
- Sad Path -- Form Validation -- bad template name that isn't compliant with [RFC3986](https://www.rfc-editor.org/rfc/rfc3986#section-2) URL-safe characters
- Sad Path -- Form Validation -- no description
- Sad Path -- Invalid JSON

ui/app/controllers/jobs/run/templates/new.js
```diff
@@ -0,0 +1,71 @@
...
+
+    return namespaces;
+  }
+
+  @action
+  updateKeyValue(key, value) {
+    if (this.keyValues.find((kv) => kv.key === key)) {
+      this.keyValues.find((kv) => kv.key === key).value = value;
+    } else {
+      this.keyValues.pushObject({ key, value });
```
ChaiWithJai, line 37: I noticed that `Variables` have warnings. Are they needed here?

ui/app/templates/jobs/run/templates/new.hbs
```diff
@@ -0,0 +1,34 @@
...
+            <Input
+                @type="text"
+                @value={{mut this.templateName}}
+                placeholder="your-template-name-here"
+                class="input path-input"
+                {{autofocus}}
+                data-test-template-name
+            />
+            </label>
+            <SingleSelectDropdown
```
philrenaud, line 17: Running into a fatal error here: 
image

(I don't have namespaces other than default)

I think you probably want to wrap this in a `{{#if this.system.shouldShowNamespaces}}` conditional

ui/app/templates/jobs/run/templates/new.hbs
```diff
@@ -0,0 +1,34 @@
+<section class="section">
```
philrenaud, line 1: Let's add a page title here: "Create custom template"

ui/app/controllers/jobs/run/templates/new.js
philrenaud: Hm, I appreciate the different solution attempted here but I"m running into a breaking error when I try this:

`Error: Attempted to handle event `deleteRecord` on <variable:nomad/job-templates/my fun template@default> while in state root.loaded.created.inFlight. `

ui/app/templates/jobs/run/templates/new.hbs
```diff
@@ -0,0 +1,34 @@
...
+                @options={{this.namespaceOptions}}
+                @selection={{this.templateNamespace}}
+                @onSelect={{fn (mut this.templateNamespace)}}
+            />
+        </div>
+        <VariableForm::JobTemplateEditor
+            @keyValues={{this.model.keyValues}}
+            @updateKeyValue={{this.updateKeyValue}}
+        />
+        <footer>
```
philrenaud, line 29: Margin between editor and footer please 
image

ui/app/templates/jobs/run/templates/new.hbs
```diff
@@ -0,0 +1,34 @@
...
+            </span>
+            <Input
+                @type="text"
+                @value={{mut this.templateName}}
+                placeholder="your-template-name-here"
+                class="input path-input"
+                {{autofocus}}
+                data-test-template-name
+            />
+            </label>
```
philrenaud, line 16: I'm not sure we want this here; we should instead add it into the VariableForm::JobTemplateEditor. Name is already a field there.

ui/app/controllers/jobs/run/templates/new.js
philrenaud: They're helpful for storing information about problems in a variable (whether it comes back from a server, has a duplicate name, etc.) but are probably not needed here.

philrenaud (changes requested):
Code looks pretty good, but smoke testing revealed some bugs.

ui/app/controllers/jobs/run/templates/new.js
ChaiWithJai: I'm unable to reproduce this. Here's a [Replay](https://app.replay.io/recording/reproduction-of-error--ff185084-7610-4771-9474-92090d5bbd0d) showing that I'm getting to success.

Could you record a Replay so that I can trace what's happening to start debugging here?

ui/app/templates/jobs/run/templates/new.hbs
ChaiWithJai: OK! I'm following the pattern in `variable-form.hbs`, where `path` isn't included in the `variable-form.job-template-editor` component.

Do you want me to create this as a conditionally rendered part of the template in that component?

ui/app/controllers/jobs/run/templates/new.js
ChaiWithJai: Look like calling `unloadRecord` patches our problem here.

ui/app/templates/jobs/run/templates/new.hbs
ChaiWithJai: Should be patched by some conditional logic here for when `namespaces` are enabled.

philrenaud (approved):

Issue #15629, opened 2022-12-31
nomad: vault integration > nomad doesnt set user-agent header when making requests to vault


noahehall:
### Nomad version
Output from `nomad version`
```sh
08:29 AM (nomad *$|u=) 
$ nomad -version
Nomad v1.4.3 (f464aca721d222ae9c1f3df643b3c3aaa20e2da7)
```

### Operating system and Environment details

```sh
08:29 AM (nomad *$|u=) 
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.1 LTS
Release:	22.04
Codename:	jammy

```
### Issue
- following the [vault integration docs](https://developer.hashicorp.com/nomad/docs/integrations/vault-integration)  (really good by the way)
- nomad doesnt set user-agent header when making requests to vault over http
- haproxy [ACL documentation](https://www.haproxy.com/blog/introduction-to-haproxy-acls/) recommends blocking requests lacking proper user agent header
```sh
Length of the user-agent string

Some attackers will attempt to bypass normal 
user agent strings by using a random md5sum, 
which can be identified by length and immediately blocked:

Attackers can vary more with their attacks, 
so you can rely on the fact that legitimate 
user agents are longer while also being set 
to a minimum length:

http-request deny if { req.hdr(user-agent) -m len le 32 }

This will then block any requests which 
have a user-agent header shorter than 32 characters.
```

### Reproduction steps

- following the [vault integration docs](https://developer.hashicorp.com/nomad/docs/integrations/vault-integration) 
- use these [haproxy configs](https://github.com/nirv-ai/configs/tree/develop/haproxy/configs) to start a tcp >  http reverse proxy > vault stack
  - this is the [reverse proxy for vault](https://github.com/nirv-ai/configs/blob/develop/haproxy/configs/002-001-vault.cfg)
  - these are the [reverse proxy defaults](https://github.com/nirv-ai/configs/blob/develop/haproxy/configs/002-000-http-defaults.cfg)

#### Expected Result

- nomad  correclty implements [http rfc 2048](https://www.ietf.org/rfc/rfc2068.txt) and sets a user agent string
- vault (IMO) should exceed the standard set by [github](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#user-agent-required) and not accept requests over http lacking user agent header
- nomad doesnt require me to modify this line

```sh
[...]

PR #15745, opened 2023-01-10
vault: configure user agent on Nomad vault clients

shoenig:
This PR attempts to set the User-Agent header on each Vault API client
created by Nomad. Still need to figure a way to set User-Agent on the
Vault client created internally by consul-template.

Partial of https://github.com/hashicorp/nomad/issues/15629

### DIFF
--- a/client/vaultclient/vaultclient.go
+++ b/client/vaultclient/vaultclient.go
@@ -4,13 +4,13 @@ import (
 	"container/heap"
 	"fmt"
 	"math/rand"
-	"net/http"
 	"strings"
 	"sync"
 	"time"
 
 	metrics "github.com/armon/go-metrics"
 	hclog "github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/nomad/helper/useragent"
 	"github.com/hashicorp/nomad/nomad/structs"
 	"github.com/hashicorp/nomad/nomad/structs/config"
 	vaultapi "github.com/hashicorp/vault/api"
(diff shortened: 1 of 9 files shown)


client/vaultclient/vaultclient.go
```diff
@@ -353,7 +352,7 @@ func (c *vaultClient) renew(req *vaultClientRenewalRequest) error {
 	var renewalErr error
 	leaseDuration := req.increment
 	if req.isToken {
-		// Set the token in the API client to the one that needs
+		// SetHeaders the token in the API client to the one that needs
```
lgfa29, line 355: Was this a find-and-replace gone wild? There are some others like this in `nomad/vault.go` as well 😅

helper/useragent/useragent.go
```diff
@@ -27,3 +28,15 @@ func String() string {
...
+
+// HeaderSetter is anything that implements SetHeaders(http.Header).
+type HeaderSetter interface {
+	SetHeaders(http.Header)
+}
+
+// SetHeaders configures the User-Agent http.Header for the client.
+func SetHeaders(client HeaderSetter) {
+	client.SetHeaders(http.Header{
+		"User-Agent": []string{String()},
```
lgfa29, line 40: Hum...maybe I'm missing something, but the user-agent used to be [`hashicorp/nomad`](https://github.com/hashicorp/nomad/pull/15745/files#diff-61b0f81ca4505d23bb2adf7025d2c36c4e02232a2f13ac39dce779617c954416L155) and now it's going to be empty?

client/vaultclient/vaultclient.go
shoenig: Ah crap, I dunno how I didn't see this

helper/useragent/useragent.go
lgfa29: Ops..I'm still warming up my brain from the time off 😄

lgfa29 (approved):

Issue #8647, opened 2020-08-12
Nomad can not use consul ingress-gateways because tasks use protocol tcp


spuder:
Nomad 0.11.1  
Consul 1.8.2  

Consul Ingress-Gateways support `tcp` and `http` listeners. Http listeners are preferred because they allow for multiple services to listen on a single port and use Host header identification. 

## Problem

Nomad jobs default to service type of `tcp`.  There  does not appear to be a documented way to change a nomad job to use `http` as the service type. As a result the user will get the following error when they attempt to create a listener for it. 

```
Error writing config entry ingress-gateway/ingress-ngproxy: Unexpected response code: 500 (rpc error making call: service "count-dashboard" has protocol "tcp", which does not match defined listener protocol "http")
```

## Steps to reproduce

1. Submit the standard count-dash example

<details><summary>count-dash.job</summary>
<p>

#### yes, even hidden code blocks!

```
job "countdash" {
   datacenters = ["dc1"]
   group "api" {
     network {
       mode = "bridge"
     }

     service {
       name = "count-api"
       port = "9001"

       connect {
         sidecar_service {}
       }
     }

     task "web" {
       driver = "docker"
       config {
         image = "hashicorpnomad/counter-api:v1"
       }
     }
   }

   group "dashboard" {
     network {
       mode ="bridge"
       port "http" {
         static = 9002
         to     = 9002
       }
     }

     service {
       name = "count-dashboard"
       port = "9002"
       # This is slightly modified from the stock count-dash examples
       # By adding an 'http' health check, the hope was to force nomad to use 'http' over 'tcp'
       check  {
         name = "count-dashboard-health"
         type = "http"
         protocol = "http"
         path = "/health"
         port = 9002
         interval = "10s"
         timeout = "5s"

       }
       connect {
         sidecar_service {
           proxy {
             upstreams {
               destination_name = "count-api"
               local_bind_port = 8080
             }
           }
         }
       }
     }

     task "dashboard" {
       driver = "docker"
       env {
         COUNTING_SERVICE_URL = "http://${NOMAD_UPSTREAM_ADDR_count_api}"
       }
       config {
         image = "hashicorpnomad/counter-dashboard:v1"
       }
     }
   }
 }
 ```

</p>
</details>

2. Create an ingress controller and register it with consul config

```
consul config write ingress-service.hcl
```
```

Listeners = [
 {
   Port = 8080
   Protocol = "http"
   Services = [
     {
	Name = "count-dashboard",
[...]

PR #15749, opened 2023-01-10
connect: ingress gateway validation for http hosts and wildcards

### LINKED ISSUE #10955: Consul Ingress Gateway config incorrectly requires Hosts value
### Nomad version
1.1.2

### Issue

When configuring a Consul Ingress Gateway using Nomad, if you do not provide a "hosts" value, Nomad throws an error. The relevant code is here -  https://github.com/hashicorp/nomad/blob/v1.1.2/nomad/structs/services.go#L1749-L1751

In the Consul documentation, it shows hosts as optional though: https://www.consul.io/docs/connect/config-entries/ingress-gateway#hosts - Have confirmed with the Consul team that this is accurate.

Additionally, the [ingress example](https://www.nomadproject.io/docs/job-specification/gateway#ingress-gateway) on the website doesn't include hosts, so it errors.

#### Expected Result

Job with consul ingress gateway deployed successfully.

#### Actual Result

Error `Consul Ingress Service requires one or more hosts when using HTTP protocol
")`

### Job file

```
ingress {
    tls {
        enabled = false
    }
    listener {
        port     = 8080
        protocol = "http"
        service {
            name  = "freshrss"
            # hosts = ["freshrss.ingress.consul"] - fails unless this line is added
        }
    }
}
```

gulducat:
This makes code match the documentation, and reality 😋 

Applies to all non-"tcp" protocols: `http`, `http2`, and `grpc`, which support "hosts" and tests now cover all of them as well.  I could maybe be convinced to remove the extra test coverage if it seems superfluous, but it's intended to guard against potential future regressions.

per https://developer.hashicorp.com/nomad/docs/job-specification/gateway#service-parameters,

> `service` Parameters
> * [`hosts`](https://developer.hashicorp.com/nomad/docs/job-specification/gateway#hosts) `(array<string>: nil)` - A list of hosts that specify what requests will match this service. This cannot be used with a `tcp` listener, and cannot be specified alongside a wildcard (`*`) service name. If not specified, the default domain `<service-name>.ingress.*` will be used to match services.

<details><summary>e.g. this will now work:</summary>

```hcl
listener {
  port     = 8080
  protocol = "http"
  service {
    name  = "uuid-api"
    # hosts = no longer required
  }
}
```

and this will no longer work (by "work" I mean pass to consul, which errors less-specifically):

```hcl
listener {
  port     = 8080
  protocol = "http"
  service {
    name  = "*"
    hosts = ["anything"]
  }
}
```

error before:

> Error submitting job: Unexpected response code: 500 (Unexpected response code: 500 (Associating hosts to a wildcard service is not supported (listener on port 8080)))

error after:

> Error submitting job: Unexpected response code: 500 (1 error occurred:
        * Task group ingress-group validation failed: 1 error occurred:
        * Service[0] my-ingress-service validation failed: 1 error occurred:
        * Consul Ingress Service with a wildcard "*" service name can not also specify hosts)

</details>

Closes #10955 

Note: to use these non-"tcp" protocols, users will still need to manually write a service-defaults config entry as described in https://github.com/hashicorp/nomad/issues/8647#issuecomment-691279667

### DIFF
--- a/nomad/structs/services_test.go
+++ b/nomad/structs/services_test.go
@@ -1328,13 +1328,6 @@ func TestConsulIngressService_Validate(t *testing.T) {
 		require.EqualError(t, err, "Consul Ingress Service requires a name")
 	})
 
-	t.Run("http missing hosts", func(t *testing.T) {
-		err := (&ConsulIngressService{
-			Name: "service1",
-		}).Validate("http")
-		require.EqualError(t, err, `Consul Ingress Service requires one or more hosts when using "http" protocol`)
-	})
-
 	t.Run("tcp extraneous hosts", func(t *testing.T) {
 		err := (&ConsulIngressService{
 			Name:  "service1",
@@ -1343,25 +1336,10 @@ func TestConsulIngressService_Validate(t *testing.T) {
 		require.EqualError(t, err, `Consul Ingress Service doesn't support associating hosts to a service for the "tcp" protocol`)
 	})
 
-	t.Run("ok tcp", func(t *testing.T) {
+	t.Run("tcp ok", func(t *testing.T) {
 		err := (&ConsulIngressService{
 			Name: "service1",
 		}).Validate("tcp")
-		require.NoError(t, err)
-	})
-
-	t.Run("ok http", func(t *testing.T) {
-		err := (&ConsulIngressService{
-			Name:  "service1",
-			Hosts: []string{"host1"},
-		}).Validate("http")
-		require.NoError(t, err)
-	})
-
-	t.Run("http with wildcard service", func(t *testing.T) {
-		err := (&ConsulIngressService{
-			Name: "*",
-		}).Validate("http")
 		require.NoError(t, err)
 	})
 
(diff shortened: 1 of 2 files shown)


nomad/structs/services_test.go
```diff
@@ -1343,34 +1336,52 @@ func TestConsulIngressService_Validate(t *testing.T) {
...
 	})
+
+	// non-"tcp" protocols should be all treated the same.
+	for _, proto := range []string{"http", "http2", "grpc"} {
+		t.Run(proto+" ok", func(t *testing.T) {
+			err := (&ConsulIngressService{
+				Name:  "service1",
+				Hosts: []string{"host1"},
+			}).Validate(proto)
+			require.NoError(t, err)
```
shoenig, line 1360: We've been slowly migrating away from stretchr/testify in favor of shoenig/test (NMD-140 for more info). Generally new code should use the new library.

shoenig (approved):
LGTM! just the one suggestion. Be sure to set backport labels for 1.4.x, 1.3.x, and 1.2.x

nomad/structs/services_test.go
gulducat: I thought about switching in this one test func I'm changing, but my intuition was that a separate PR to convert this whole file would feel cleaner?  Are my eyes bigger perhaps than my stomach to think this?

gulducat: alrighty I went ahead and converted `TestConsulIngressService_Validate` to use the new lib.  how's that look?

shoenig: LGTM!

Issue #15614, opened 2022-12-22
Workload Identity


mikenomitch:
### Proposal

As part of the Variables project for Nomad 1.4, each Nomad Task is given a signed token that encodes information about its namespace, job, task group, and task. This Nomad Workload Identity JWT is used in order to validate that the Nomad task in question has access to the secure variables requested.

We should expand upon this work in several ways:

- Task should have access to their Workload Identity token and Nomad’s public key
- Standard OIDC keys and values to workload identity tokens
- A JWKS URL for public keys should be added
- An OIDC Configuration URL for OIDC Discovery should be added

This would allow us to greatly expand the use of the workload identity tokens for more use cases.

### Use-cases

- This would allow users to use Nomad as an OIDC provider. They could configure OIDC consumers to give access to resources based on identity. For instance, they could [AWS resources](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_oidc.html) such as S3 buckets could be exposed to certain jobs. Azure and GCP have similar mechanisms
- This would allow users to have fine grained permission checks based on JWT. For instance, if a certain route in app A should only allow requests from app B, the JWT from app B could prove identity and grant permission.
- This could be used as the basis to redo the Vault integration to remove the need for Vault ACL Tokens on every job, allow for multiple Vault clusters per Nomad cluster, and make it easier to run Vault as a Nomad job
- This could be used as the basis to redo the Consul integration to remove the need for Consul ACL Tokens on every job, allow for multiple Consul clusters per Nomad cluster, and make it easier to run Consul as a Nomad job

mikenomitch:
@anarsen, the public URL issue is something we'll still need to figure out.

To be honest, I'm not sure what our solution will be, but I'm glad you've called it out here, as it'll force us to think about this early. I wouldn't want to force people to deploy a job to proxy only specific routes to the Nomad servers, but that'd be at least one way to expose just these routes. If there's some other option you, or anybody else reading this, has in mind, let me know.

I'll update this thread once we zero in on a solution for this.

anarsen:
My initial thought is that the OIDC configuration endpoints are cacheable, and offloading the Nomad servers from serving these sounds like a good idea.

# Proposal 1: OIDC agent

A separate Nomad agent serving only OIDC configurations. Nomad servers would push OIDC config to these agents when there are updates.

[...]

PR #15755, opened 2023-01-11
Add option to expose workload token to task

angrycub:
This PR creates a configuration flag that, when set, outputs the Nomad task identity token to the task secrets folder and loads it into the environment. The goal being to provide users credentials that they could use to perform actions based on the Nomad Workload Identity token.


tgross (commented):
In the upcoming WI epic @mikenomitch wrote up https://github.com/hashicorp/nomad/issues/15614 it seems like we want to _always_ expose the WI token to the task. Is there a reason we want to do this as a configuration flag?

angrycub:
Making it user selectable creates parity with the ability to emit or not emit vault tokens to the tasks, since not all tasks might be trusted, but might require workload identity for variables to be used in templates.

tgross:
> Making it user selectable creates parity with the ability to emit or not emit vault tokens to the tasks, since not all tasks might be trusted, but might require workload identity for variables to be used in templates.

Arguably there's a difference between the Nomad WI and Vault tokens, so it's not obvious to me there _needs_ to be 1:1 parity. Also, we have to support jobspec flags basically _forever_ once they're created. I'm open to this idea but, I'd rather have a more holistic discussion about how we want WI to be exposed to tasks than a one-off PR for a config flag.

mikenomitch:
Oooh, I'm excited for this :)

I think since the token's purpose is to declare "I am this task", it probably makes sense to have it available by default and the flag isn't worth it. Would remove one step from configuring workflows that use it too.

tgross:
> I think since the token's purpose is to declare "I am this task", it probably makes sense to have it available by default and the flag isn't worth it.

I think you're on the right track here. Having the WI token just lets the workload claim an _identity_ -- it doesn't expose any new _capabilities_ (other than workload-scoped variables, which we already grant the workload via `template`). You still need to configure a policy for that identity for it to be at all interesting to the workload.

client/allocrunner/taskrunner/identity_hook.go
```diff
@@ -11,11 +14,21 @@ import (
 
 // identityHook sets the task runner's Nomad workload identity token
 // based on the signed identity stored on the Allocation
+
+const (
+	// wiTokenFile is the name of the file holding the Vault token inside the
```
anarsen, line 19:
```suggestion
	// wiTokenFile is the name of the file holding the Nomad token inside the
```

Right?

schmichael: Yup! Thanks!

schmichael:
I pushed cae00049787f6b66ec19101b32d3481d1346abea to split the single bool into a block similar to `vault` with `env` and `file` bools like:

```hcl
identity {
  env  = true
  file = true
}
```

It defaults to disabled and each parameter must be enabled individually which *differs* from `vault` which defaults to `env = true` and always writes the token file.

To match Vault we could make file and env default to true if identity is set so that:

```hcl
task "foo" {
  identity {}

  # ...
}
```

...gets both an env var and file by default.

Regardless of which direction we go I think we should get #13343 merged ASAP after this as it's nice to give users fine grained control.

client/allocrunner/taskrunner/identity_hook.go
```diff
@@ -34,21 +47,42 @@ func (*identityHook) Name() string {
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// writeToken writes the given token to disk
+func (h *identityHook) writeToken(token string) error {
+	if err := ioutil.WriteFile(h.tokenPath, []byte(token), 0666); err != nil {
```
shoenig, line 83:
```suggestion
	if err := ioutil.WriteFile(h.tokenPath, []byte(token), 0666); err != nil {
```

Does this file need to be world writable?

schmichael: Hm, this copied the Vault token file's behavior. My assumptions are:

1. This is necessary because we don't know what user the task will run as
2. This is safe because of directory permissions on the parent dir

If these aren't true, I'd love to fix them. If nothing else it's an awful smell.

shoenig: Hmm yeah, maybe there's something we could do if Task.User happens to be set?

But my real question is why does the file need to be _writable_? Would `0664` or even `0644` be sufficient?

schmichael: Just realized my 1-3 workflow above is broken by the fact that alloc updates will rewrite the token, so that use case is right out...

Maybe it shouldn't be world writable :thinking:

schmichael: The good: Checkout what I came up with in `helper/users.WriteFileFor`. It appears to degrade gracefully when `Task.User` is unset (even without the explicit guard I added) as well as properly set the mode and owner when running as `root` on Linux. All of this without any OS specific implementations so it should even work on any Unix.

The bad: I don't think there's anything bad here, but it definitely increases risk. I think that's tolerable for a new off-by-default feature.

The ugly: Check out the Dockerfile for our example job: https://github.com/docker-library/redis/blob/master/Dockerfile.template#L3-L4 It creates a user within the container and runs Redis as that. Setting `task.user = "nobody"` does still work. Redis runs as nobody within the container, and `nomad_secret` gets properly owned by nobody. Without setting `task.user` the fallback is used, so Redis, running as `redis`, is still able to access `nomad_secret`! So... it's ugly but it works?

I haven't tested it with Docker and user namespaces.

helper/users/lookup_test.go
```diff
@@ -39,3 +43,37 @@ func TestLookup_NobodyIDs(t *testing.T) {
 	must.Eq(t, 65534, uid) // ubuntu
 	must.Eq(t, 65534, gid) // ubuntu
 }
+
+func TestWriteFileFor_Linux(t *testing.T) {
+	// This is really how you have to retrieve umask. See `man 2 umask`
+	umask := syscall.Umask(0)
```
shoenig, line 49: `syscall` is (mostly?) deprecated in favor of `unix`

https://pkg.go.dev/golang.org/x/sys/unix#Umask

e2e/workload_id/workload_id_test.go
```diff
@@ -0,0 +1,173 @@
...
+			must.Len(t, 2, lines, ps)
+		}
+	}
+
+}
+
+func parseEnv(t *testing.T, line string, ps must.Setting) string {
+	must.StrHasPrefix(t, "NOMAD_TOKEN=", line, ps)
+	token := strings.Split(line, "=")[1]
+	must.Positive(t, len(token), ps)
```
shoenig, line 129: is the token not always a UUID? If so you could just `must.UUIDv4(token)`

helper/users/lookup.go
```diff
@@ -30,3 +35,63 @@ func Current() (*user.User, error) {
...
+		}
+	}
+
+	// Fallback to world readable
+	if err := os.WriteFile(path, contents, 0o666); err != nil {
+		if origErr != nil {
+			// Return both errors
+			return &multierror.Error{
+				Errors: []error{origErr, err},
+			}
```
shoenig, line 65: Can't wait for this to be `errors.Join` https://pkg.go.dev/errors@master#Join

helper/users/lookup.go
```diff
@@ -30,3 +35,63 @@ func Current() (*user.User, error) {
...
+func writeFileFor(path string, contents []byte, username string) error {
+	user, err := Lookup(username)
+	if err != nil {
+		return err
+	}
+
+	uid, err := strconv.Atoi(user.Uid)
+	if err != nil {
+		return fmt.Errorf("error parsing uid: %w", err)
+	}
```
shoenig, line 83: Will this work on Windows? Where user.Uid is a SID string. I suspect where we convert to a UID in other places is always gated on unix build tags. And `chown` will always fail on Windows anyway. https://pkg.go.dev/os#Chown

e2e/workload_id/workload_id_test.go
tgross: This is the token from the workload identity, so it'll be a signed JWT (which is not shaped like the UUID bearer tokens)

client/allocrunner/taskrunner/identity_hook.go
schmichael: @shoenig mind peeking at this and resolving the comment thread if it looks good to you?

shoenig: I added this comment about Windows
https://github.com/hashicorp/nomad/pull/15755/files#r1092065794

website/content/docs/job-specification/identity.mdx
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }
+}
+```
+
+## `identity` Parameters
+
+- `env` `(bool: true)` - If true the workload identity will be available in the
+  task's `NOMAD_TOKEN` environment variable.
+- `file` `(bool: true)` - If true the workload identity will be available in
```
tgross, line 44: Defaults are false now, right?

```suggestion
- `env` `(bool: false)` - If true the workload identity will be available in the
  task's `NOMAD_TOKEN` environment variable.
- `file` `(bool: false)` - If true the workload identity will be available in
```

helper/users/lookup.go
schmichael: Yeah, I think this func will always fail on Windows. I could make os specific implementations, but instead I tried the approach of having a universal fallback in case of error. I'll add a Windows specific test to assert the expected behavior (fails internally, succeeds with fallback).

website/content/docs/job-specification/identity.mdx
schmichael: Yes! Thanks for catching that.

api/tasks.go
```diff
@@ -1110,3 +1112,10 @@ func (t *TaskCSIPluginConfig) Canonicalize() {
 		t.HealthTimeout = 30 * time.Second
 	}
 }
+
+// WorkloadIdentity is the jobspec block which determines if and how a workload
+// identity is exposed to tasks.
+type WorkloadIdentity struct {
+	Env  *bool `hcl:"env,optional"`
+	File *bool `hcl:"file,optional"`
```
shoenig, line 1120: Now that these default false, do they need to be pointers?

helper/users/lookup.go
shoenig: The call to `Lookup` here could be quite expensive (network hops as Windows does domain lookups, etc.) but even worse it's a global mutex. The fallback is probably fine for beta but I do think we should optimize this out if we know Windows will always take the error path.

api/tasks.go
schmichael: They don't, but I kind of wanted to reserve the ability to distinguish between `false` and `unset` in the future (without a breaking API change).

For example if we add `file_name = "..."` or `file_owner = "..."` parameters, do we still need to make set `file` explicitly? We may even want to change the default to `true` in the future if the unauthenticated HTTP endpoints issue can be addressed.

I'm fine removing the pointer though if this is too strange.

shoenig: We've shipped so many nil pointer panics already I would lean toward using them only when they're necessary. Having these as pointers is also annoying to the API user, who now has to do the 

```
var b bool
t.Env = &b
```

hoop jumping.

helper/users/lookup.go
schmichael: Good point about the scary combo of `network lookup + global lock`. Wrote up #16010 - please feel free to edit/comment/fix. :grin:

helper/users/lookup_test.go
schmichael: Can we just deprecate `umask` instead? (kidding kidding fix pushed but syscall import remains due to use of `Stat_t`)

helper/users/lookup.go
schmichael: It nearly killed me, but I added a test for Windows to assert it hits the fallback behavior! :tada: It passes as both a normal user and a local Administrator because (as expected/documented) `chown` is Unix-only.

scheduler/util.go
```diff
@@ -599,6 +599,11 @@ func tasksUpdated(jobA, jobB *structs.Job, taskGroup string) bool {
 		} else if !ar.Devices.Equal(&br.Devices) {
 			return true
 		}
+
+		// Inspect Identity being exposed
+		if !reflect.DeepEqual(at.Identity, bt.Identity) {
+			return true
+		}
```
shoenig, line 606: nitpick: prefer to avoid `reflect.DeepEqual` which is both slow and dangerous - we've fixed at least half a dozen bugs where `DeepEqual` treats things like nil vs empty slice differently.

shoenig (approved):
LGTM, just left one last nitpick :)

scheduler/util.go
schmichael: Good call. Added a WorkloadIdentity.Equal method + a test, and used that method instead of DeepEqual here.

PR #15764, opened 2023-01-12
sso: add OIDC login RPC, HTTP, and CLI workflow

jrasell:
Each commit targets a different section of the code and attempts to make reviewing easier.

Raising in draft initially so I can perform some additional local testing in parallel.

Related: https://github.com/hashicorp/nomad/issues/13120

### DIFF
--- a/command/login_test.go
+++ b/command/login_test.go
@@ -1,0 +1,57 @@
+package command
+
+import (
+	"testing"
+
+	"github.com/hashicorp/nomad/ci"
+	"github.com/hashicorp/nomad/command/agent"
+	"github.com/hashicorp/nomad/testutil"
+	"github.com/mitchellh/cli"
+	"github.com/shoenig/test/must"
+)
+
+func TestLoginCommand_Run(t *testing.T) {
+	ci.Parallel(t)
+
+	// Build a test server with ACLs enabled.
+	srv, _, agentURL := testServer(t, false, func(c *agent.Config) {
+		c.ACL.Enabled = true
+	})
+	defer srv.Shutdown()
+
+	// Wait for the server to start fully.
+	testutil.WaitForLeader(t, srv.Agent.RPC)
+
+	ui := cli.NewMockUi()
+	cmd := &LoginCommand{
+		Meta: Meta{
+			Ui:          ui,
+			flagAddress: agentURL,
+		},
+	}
+
+	// Test the basic validation on the command.
+	must.Eq(t, 1, cmd.Run([]string{"-address=" + agentURL, "this-command-does-not-take-args"}))
+	must.StrContains(t, ui.ErrorWriter.String(), "This command takes no arguments")
+
+	ui.OutputWriter.Reset()
+	ui.ErrorWriter.Reset()
+
+	// Attempt to call it with an unsupported method type.
+	must.Eq(t, 1, cmd.Run([]string{"-address=" + agentURL, "-type=SAML"}))
+	must.StrContains(t, ui.ErrorWriter.String(), `Unsupported authentication type "SAML"`)
+
+	ui.OutputWriter.Reset()
+	ui.ErrorWriter.Reset()
+
+	// Use a valid method type but with incorrect casing so we can ensure this
+	// is handled.
+	must.Eq(t, 1, cmd.Run([]string{"-address=" + agentURL, "-type=oIdC"}))
+	must.StrContains(t, ui.ErrorWriter.String(), "Must specify an auth method name, no default found")
+
+	ui.OutputWriter.Reset()
+	ui.ErrorWriter.Reset()
+
+	// TODO(jrasell) find a way to test the full login flow from the CLI
+	//  perspective.
+}
--- a/lib/auth/oidc/provider.go
+++ b/lib/auth/oidc/provider.go
@@ -1,0 +1,189 @@
+package oidc
+
+import (
+	"context"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/hashicorp/cap/oidc"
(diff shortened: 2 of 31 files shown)


command/login_test.go
```diff
@@ -0,0 +1,57 @@
...
+	// Use a valid method type but with incorrect casing so we can ensure this
+	// is handled.
+	must.Eq(t, 1, cmd.Run([]string{"-address=" + agentURL, "-type=oIdC"}))
+	must.StrContains(t, ui.ErrorWriter.String(), "Must specify an auth method name, no default found")
+
+	ui.OutputWriter.Reset()
+	ui.ErrorWriter.Reset()
+
+	// TODO(jrasell) find a way to test the full login flow from the CLI
+	//  perspective.
```
tgross, line 56: It's probably a huge chore but useful for E2E if we could spin up a dummy OIDC provider server. Does `cap` have something like that already?

lib/auth/oidc/provider.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	// Get any current provider for the named auth-method.
+	var (
+		current *oidc.Provider
+		ok      bool
+	)
+
+	// This may be the first time the cache is being used, therefore we need to
+	// ensure we don't panic when attempting to access the map.
```
tgross, line 91: Isn't this guaranteed by the constructor?

lib/auth/oidc/provider.go
```diff
@@ -0,0 +1,189 @@
...
+// shutting down.
+func (c *ProviderCache) Shutdown() {
+	c.cancel()
+	c.clear()
+}
+
+// runCleanupLoop runs an infinite loop that clears the cache every cacheExpiry
+// duration. This ensures that we force refresh our provider info periodically
+// in case anything changes.
+func (c *ProviderCache) runCleanupLoop(ctx context.Context) {
```
tgross, line 159: We don't need to do it in this PR but we might want to consider whether the RPC handlers can/should force a cache flush of the provider if they hit errors that are caused by the upstream provider (ex. this is a very long lived HTTP client and I don't know how `cap` has its TCP connections configured to handle a DNS change at the provider; ref https://github.com/golang/go/issues/23427)

tgross (commented):
@jrasell I've taken an initial pass over this and it's looking good so far. I really appreciate the solid comments you've left on some of the gritty bits so that folks coming along later can orient themselves.

The double-forwarding in the RPC handler seems unfortunate but given that we need to query the state to know whether we need to forward to the authoritative region I don't really see any way around it unless `AuthMethod` was assumed immutable, which as I understand it isn't the case.

nomad/acl_endpoint.go
```diff
@@ -2347,3 +2370,254 @@ func (a *ACL) GetBindingRule(
...
+
+	// Generate the data used by the go-bexpr selector that is an internal
+	// representation of the claims that can be understood by Nomad.
+	oidcInternalClaims, err := oidc.SelectorData(authMethod, idTokenClaims, userClaims)
+	if err != nil {
+		return err
+	}
+
+	// Create a new binder object based on the current state.
+	oidcBinder := oidc.NewBinder(a.srv.State())
```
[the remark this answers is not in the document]
jrasell: Yes, we should absolutely do that, thanks!

command/login_test.go
jrasell: Cap does have a local test provider we can use. I think I have figured out the other blocker I had, so will look into adding that to this PR. Apologies for not doing it sooner, my initial aim with this PR was to allow us to manually run full e2e tests using multiple real providers.

lib/auth/oidc/provider.go
jrasell: Great shout; I'll note this for future thought and raise as a followup issue.

command/login_test.go
tgross: Oh, sorry, I didn't mean to suggest we had to block this PR on that. If we've manually tested this already it seems totally reasonable to add automated E2E in a separate PR!

pkazmierczak (approved):
This looks good @jrasell! I couldn't find anything but it is a pretty large PR and even though it has neat commits that separate by functionality (🙏), it's still a lot of code and at some point fatigue kicks in and I might've missed something. This being said, I did quite some testing of this code using vault as oidc provider and it seems to be working exactly as expected.

nomad/acl_endpoint.go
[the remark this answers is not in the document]
jrasell: Resolved within commit https://github.com/hashicorp/nomad/pull/15764/commits/ebc76d23da5f673adfdfb993d0b537376d9d7e4a

lib/auth/oidc/provider.go
jrasell: Resolved within commit https://github.com/hashicorp/nomad/pull/15764/commits/ebc76d23da5f673adfdfb993d0b537376d9d7e4a

lib/auth/oidc/provider.go
```diff
@@ -0,0 +1,189 @@
...
+		algs,
+		authMethod.Config.AllowedRedirectURIs,
+		oidc.WithAudiences(authMethod.Config.BoundAudiences...),
+		oidc.WithProviderCA(strings.Join(authMethod.Config.DiscoveryCaPem, "\n")),
+	)
+}
+
+// ProviderCache is a cache for OIDC providers. OIDC providers are something
+// you don't want to recreate per-request since they make HTTP requests
+// themselves.
```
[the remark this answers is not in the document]
jrasell: Resolved within commit https://github.com/hashicorp/nomad/pull/15764/commits/ebc76d23da5f673adfdfb993d0b537376d9d7e4a

PR #15761, opened 2023-01-11
consul/connect: support for proxy upstreams opaque config

shoenig:
This PR adds support for configuring `proxy.upstreams[].config` for
Consul Connect upstreams. This is an opaque config value to Nomad -
the data is passed directly to Consul and is unknown to Nomad.

Closes #11392

PR #15765, opened 2023-01-12
consul/connect: use block not optional for opaque map

shoenig:
I forgot that I created an uncommitted change while testing https://github.com/hashicorp/nomad/pull/15761. Adds the fix and adds a relevant parsing test too.


gulducat (approved):
nice to see a full jobspec being parsed for this 👍

PR #15767, opened 2023-01-12
ui: add error handling and form validation to job create template

### LINKED ISSUE #15766: Handle error messages on Create Job Template form

ChaiWithJai:
Closes #15766

### DIFF
--- a/ui/app/controllers/jobs/run/templates/new.js
+++ b/ui/app/controllers/jobs/run/templates/new.js
@@ -8,6 +8,7 @@ import { trimPath } from '../../../../helpers/trim-path';
 export default class JobsRunTemplatesController extends Controller {
   @service router;
   @service store;
+  @service system;
   @tracked templateName = null;
   @tracked templateNamespace = 'default';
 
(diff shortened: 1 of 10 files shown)


ChaiWithJai:
An unrelated test is failing because `pretender` isn't running an `assertion`. Unsure why this is the case here.

ui/app/controllers/jobs/run/templates/new.js
```diff
@@ -19,6 +22,17 @@ export default class JobsRunTemplatesController extends Controller {
     return namespaces;
   }
 
+  get isDuplicateTemplate() {
+    const templates = this.store.peekAll('variable');
```
philrenaud, line 26: Suggestion: either use the queried variables from the route's model at jobs/run/templates/new, or else alias this somewhere outside the isDuplicateTemplate() getter. As it stands, this is going to make a store.peekAll() request every time you type a character into the path input element, which is probably doing more work than it needs to.

philrenaud (approved):
Just a small suggestion, otherwise this seems like it works well. test-ui seems to be failing in a couple spots; once that's good, this seems good to merge.

ui/app/controllers/jobs/run/templates/new.js
ChaiWithJai: Great callout! I'll address this in the `edit` template PR!

PR #15779, opened 2023-01-13
[WIP 🚧] docs: Migrate link formats

ashleemboyer:
## What

- Migrates the formats of links in Markdown content as part of the [Docs Content Link Rewrites project](https://docs.google.com/document/d/1WaSyvoVPS-YCNiSPX0ynGpvc1gySEcpbrRoQVhimQCA/edit)
- Adds two new GitHub workflow files:
  - `.github/workflows/check-legacy-links-format.yml` - This is a workflow that checks if links in markdown content are in the new format. It does this by running the link rewrite script from `dev-portal`, and then asserting that no "rewriteable" links were found. This workflow will be used on subsequent PRs to maintain the new link format.
  - `.github/workflows/test-link-rewrites.yml` - This is a workflow that executes e2e tests of this PR's changes against the `main` branch's deploy preview. The tests only run for branches named `docs/amb.migrate-link-formats`, so future PRs will not have the tests run against them. This is a temporary workflow.

## Reviewing

With the two new workflows in place, this PR should only need a spot-check.

## Testing

- [ ] The latest run of the "Legacy Link Format Checker" workflow for this branch should pass (TODO - link)
- [ ] The latest run of the "Test Link Rewrites" workflow for this branch should pass (TODO - link)


website/content/api-docs/index.mdx
```diff
@@ -47,7 +47,7 @@ There are five primary nouns in Nomad:
 - deployments
 - evaluations
 
-[![Nomad Data Model](/img/nomad-data-model.png)](/img/nomad-data-model.png)
+[![Nomad Data Model](https://nomadproject.io/img/nomad-data-model.png)](https://nomadproject.io/img/nomad-data-model.png)
```
kendallstrautman, line 50: @ashleemboyer interesting, we don't want to rewrite this I would presume. Looks like the markdown syntax for the link is _off_ though.

kendallstrautman, line 50:
```suggestion
[![Nomad Data Model](/img/nomad-data-model.png)](/img/nomad-data-model.png)
```

kendallstrautman: Syntax is right actually, just reads confusing since its an image link wrapped by a link

ashleemboyer: I think we have a couple of things to do. Let's double check with @BRKalow on how images are referenced across content repos (is it consistent?), and then update the script to ignore internal links/paths to images. I'm pretty sure a marketing content workflow handles rewriting these in content before they reach the server-side plugins, so we shouldn't need to make any changes to the plugins.

BRKalow: It depends on where authors put images in the `public` folder, but they are mostly consistent. We should be able to ignore any relative paths to images as part of the rewrites, like you said 👍

kendallstrautman: Update: we decided to fix the links in this repo manually, then we'll circle back to update the script.

topfunky (approved):
👍🏻

kendallstrautman (approved):

PR #15786, opened 2023-01-14
e2e: fixup reference to exported test type

shoenig:


tgross:
> Raises the question, how is the e2e runner still running on Circle?

[It's failing](https://app.circleci.com/pipelines/github/hashicorp/nomad-e2e/742/workflows/bb16fa1d-7d58-444a-9d6a-7fea5c0ee4b0/jobs/4254/steps) (look at the bottom), but each test suite is its own package, so that only breaks that package.

tgross (approved):
LGTM

Issue #15740, opened 2023-01-10
`nomad fmt -check` flag doesn't return error code


WebLeash:
I'm using nomad verision v1.4.2
$ nomad --version
Nomad **v1.4.2** (039d70eeef5888164cad05e7ddfd8b6f8220923b)

https://developer.hashicorp.com/nomad/docs/commands/fmt#check
The documentation says it will not change and return value 1.
Observing the behaviour this doesn't look to be working. 

I want to fail our CI process if the nomad **fmt** command detects any issue with the source file (HCL).

Currently the output of the _$ nomad fmt -check_ command lists the file in the output and returns 0 value to the shell if issues with formatting detected. 

 $ nomad fmt -check   http.nomad
http.nomad
$ echo $?
0

tgross:
Hi @WebLeash! Yup, I just checked the source and we're completely ignoring that argument. Embarrassing for me because I reviewed that PR https://github.com/hashicorp/nomad/pull/14779 :blush: 

I'm going to reword the title a bit here and mark this as a bug to get fixed. We'd also be happy to review a patch if you want to have a go at one.

dttung2905:
Hi @tgross I can help with this.

PR #15797, opened 2023-01-16
fix bug in nomad fmt -check does not return error code

dttung2905:
Hi team,

This PR is to resolve issue #15740. Could you help to review when you have time ?


WebLeash (commented):
Changes look good from what I understand of golang.

PR #15804, opened 2023-01-16
[sso] OIDC Updates for the UI

philrenaud:
This changes some properties to match expected endpoint props (AuthMethod -> AuthMethodName), and adds redirect URIs to requests.

Additionally, makes a change to the `list auth methods` API endpoint to not check for authentication (as we want to provide auth methods for users who are signed out, have invalid or expired tokens, etc.)

### DIFF
--- a/ui/app/controllers/settings/tokens.js
+++ b/ui/app/controllers/settings/tokens.js
@@ -90,13 +90,20 @@ export default class Tokens extends Controller {
     window.localStorage.setItem('nomadOIDCNonce', nonce);
     window.localStorage.setItem('nomadOIDCAuthMethod', provider);
 
+    let redirectURL;
+    if (Ember.testing) {
+      redirectURL = this.router.currentURL;
+    } else {
+      redirectURL = new URL(window.location.toString());
+      redirectURL.search = '';
+      redirectURL = redirectURL.href;
+    }
+
     method
       .getAuthURL({
-        AuthMethod: provider,
+        AuthMethodName: provider,
         ClientNonce: nonce,
-        RedirectUri: Ember.testing
-          ? this.router.currentURL
-          : window.location.toString(),
+        RedirectUri: redirectURL,
       })
       .then(({ AuthURL }) => {
         if (Ember.testing) {
@@ -111,7 +118,7 @@ export default class Tokens extends Controller {
   @tracked state = null;
 
   get isValidatingToken() {
-    if (this.code && this.state === 'success') {
+    if (this.code && this.state) {
       this.validateSSO();
       return true;
     } else {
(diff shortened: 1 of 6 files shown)


jrasell (approved):
LGTM, although please take my review of the JS sections with a caution.

pkazmierczak (approved):
backend part looks good!

ui/app/controllers/settings/tokens.js
```diff
@@ -90,13 +90,20 @@ export default class Tokens extends Controller {
     window.localStorage.setItem('nomadOIDCNonce', nonce);
     window.localStorage.setItem('nomadOIDCAuthMethod', provider);
 
+    let redirectURL;
+    if (Ember.testing) {
+      redirectURL = this.router.currentURL;
+    } else {
+      redirectURL = new URL(window.location.toString());
+      redirectURL.search = '';
+      redirectURL = redirectURL.href;
```
ChaiWithJai, line 99: `nit`: This line nullifies the work you've just done on line 97 and 98, no?

ui/app/controllers/settings/tokens.js
```diff
@@ -120,25 +127,43 @@ export default class Tokens extends Controller {
   }
 
   async validateSSO() {
+    let redirectURL;
+    if (Ember.testing) {
+      redirectURL = this.router.currentURL;
+    } else {
+      redirectURL = new URL(window.location.toString());
+      redirectURL.search = '';
+      redirectURL = redirectURL.href;
```
ChaiWithJai, line 136: ditto

ui/app/controllers/settings/tokens.js
```diff
@@ -120,25 +127,43 @@ export default class Tokens extends Controller {
...
+          RedirectURI: redirectURL,
         }),
       }
     );
 
     if (res.ok) {
       const data = await res.json();
-      this.token.set('secret', data.ACLToken);
-      this.verifyToken();
+      this.clearTokenProperties();
```
ChaiWithJai, line 155: `nit`: You're unloading the entire store here which is going to dispatch a large number of transactions. Maybe we should coordinate that this action is complete before moving to the next line of code via the `await` keyword.

ui/app/controllers/settings/tokens.js
```diff
@@ -120,25 +127,43 @@ export default class Tokens extends Controller {
...
+      this.token.set('secret', data.SecretID);
       this.state = null;
       this.code = null;
+
+      this.resetStore();
```
ChaiWithJai, line 160: So we're unloading the store on line 155, then editing a token (from an empty store), and then unloading the store, again here. And then we fetch the token on line 163.

Maybe we might be doing too much work here. There's a small chance that once the store has a large number of items that we'll run into challenges with asynchronocity.

Should we get an integration test on this?

ChaiWithJai (approved):
My gut says there might be a lurking bug in here with how we're clearing out the store in the `validateSSO` logic. But I'm imagining you've already done a fair amount of smoke testing here anyhow. Did we test what happens on a slow network or having a large amount of items in the store?

ui/app/controllers/settings/tokens.js
ChaiWithJai: What's hard for me to grok here is that, I'm reading this as:

```
variable `redirectURL` is assigned to a new object in memory using URL constructor
`redirectURL.search` is assigned a value of empty string
`redirectURL` is assigned to the value of `redirectURL.href` -- this would override what just happened in the previous step, no?
```

philrenaud: If the URL constructor generated a POJO or something that would be true, but it differs in a few ways.

Consider if I was running this from a hypothetical `site.com/page?foo=bar`

```
redirectURL = { href:"site.com/page?foo=bar", search="foo=bar", ... }
redirectURL.search = "" // since URL.href is dependent upon search and other props, it is modified to become "site.com/page"
redirectURL = "site.com/page"
```

ui/app/controllers/settings/tokens.js
philrenaud: Noting that this is an existing established pattern from `verifyToken()` (sign in without SSO); the events being being triggered are not asynchronous (`store.unloadAll()` might be?) but all subsequent actions here are healthy ones to continue down regardless of what's happening int he background: we're clearing out state/code (meant for a single immediate transaction, not meant for persistence) and moving forward with a logged in state.

ui/app/controllers/settings/tokens.js
philrenaud: You may be correct here! This looks like it could be a redundant unload of the store. 

FWIW, we do run this in a regular acceptance test [here](https://github.com/hashicorp/nomad/blob/c314661baff456c7ea04ff531b1c2b9f5a4b2150/ui/tests/acceptance/token-test.js#L187) with a non-trivial amount of store load.

ChaiWithJai: Interesting! There's a `forEach` being called when you [inspect](https://github.com/emberjs/data/blob/v4.7.3/packages/store/addon/-private/store-service.ts#L650) `InstanceCache` which is called on `unloadAll`. But, I'm not complaining! Less Work!!!

ui/app/controllers/settings/tokens.js
ChaiWithJai: Aha! Couldn't we have just take `redirectURL.host` or `redirectURL.origin` instead? But either way, that's clear now.

PR #15805, opened 2023-01-16
Clarify conflicting metric documentation, fix typo

karelorigin:
The description for the `nomad.nomad.blocked_evals.total_blocked` states that this could include evals blocked due to reached quota limits, but the `total_quota_limit` mentions being exclusive to its own metric.  I personally interpret `total_blocked` as encompassing any blocked evals for any reason, as written in the docs. Though someone will have to verify the validity of that statement and possibly rectify the other metric description.

Fixed a typo: `limtis` vs `limits`.

**Note: I made the commit via the GitHub UI and assumed I was writing a PR description, which is why the commit description contains the message above and might not make sense outside the context of this PR. Sorry about that.**


tgross (approved):
LGTM, thanks @karelorigin!

PR #15815, opened 2023-01-18
ui: job templates manager view

### LINKED ISSUE #15550: [UI, Example Job Templates] New route: jobs/run/templates/manage
image

ChaiWithJai:
Resolves #15550

### DIFF
--- a/ui/app/helpers/create-columns.js
+++ b/ui/app/helpers/create-columns.js
@@ -1,7 +1,0 @@
-import { helper } from '@ember/component/helper';
-
-function createColumns(positional) {
-  return positional.map((column) => ({ label: column }));
-}
-
-export default helper(createColumns);
--- a/ui/app/helpers/create-columns.js
+++ b/ui/app/helpers/create-columns.js
@@ -1,0 +1,7 @@
+import { helper } from '@ember/component/helper';
+
+function createColumns(positional) {
+  return positional.map((column) => ({ label: column }));
+}
+
+export default helper(createColumns);
--- a/ui/app/routes/jobs/run/templates/manage.js
+++ b/ui/app/routes/jobs/run/templates/manage.js
@@ -1,0 +1,31 @@
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class RunTemplatesRoute extends Route {
+  @service can;
+  @service router;
+  @service store;
+
+  beforeModel() {
+    const hasPermissions = this.can.can('write variable', null, {
+      namespace: '*',
+      path: '*',
+    });
+
+    if (!hasPermissions) {
+      this.router.transitionTo('jobs');
+    }
+  }
+
+  async model() {
+    const jobTemplateVariables = await this.store.query('variable', {
+      prefix: 'nomad/job-templates',
+      namespace: '*',
+    });
+    const recordsToQuery = jobTemplateVariables.map((template) =>
+      this.store.findRecord('variable', template.id)
+    );
+
+    return Promise.all(recordsToQuery);
+  }
+}
--- a/ui/app/templates/jobs/run/templates/manage.hbs
+++ b/ui/app/templates/jobs/run/templates/manage.hbs
@@ -15,7 +15,7 @@
             <:body as |B|>
                 <B.Tr>
                     <B.Td>
-                        <LinkTo @route="jobs.run.templates.template" @model={{concat B.data.path "@" B.data.namespace}}>
+                        <LinkTo @route="jobs.run.templates.template" @model={{concat B.data.path "@" B.data.namespace}} data-test-edit-template={{B.data.path}}>
                             {{B.data.path}}
                         </LinkTo>
                     </B.Td>
--- a/ui/app/templates/jobs/run/templates/manage.hbs
+++ b/ui/app/templates/jobs/run/templates/manage.hbs
@@ -11,12 +11,12 @@
         <Hds::Button class="button-group" @text="Back to editor" @route="jobs.run" @icon="arrow-left" data-test-cancel />
     {{else}}
     <main class="radio-group" data-test-template-list>
-        <Hds::Table @model={{this.model}} @columns={{ create-columns "Name" "Namespace" "Description"}}>
+        <Hds::Table @model={{this.model}} @columns={{this.columns}}>
             <:body as |B|>
                 <B.Tr>
                     <B.Td>
                         <LinkTo @route="jobs.run.templates.template" @model={{concat B.data.path "@" B.data.namespace}} data-test-edit-template={{B.data.path}}>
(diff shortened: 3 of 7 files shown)


ui/app/routes/jobs/run/templates/manage.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  async model() {
+    const jobTemplateVariables = await this.store.query('variable', {
+      prefix: 'nomad/job-templates',
+      namespace: '*',
+    });
+    const recordsToQuery = jobTemplateVariables.map((template) =>
+      this.store.findRecord('variable', template.id)
+    );
```
philrenaud, line 27: Seeing all variables [here](https://nomad-storybook-and-ee1v8mego-hashicorp.vercel.app/ui/jobs/run/templates/manage/) in the Mirage mocks: 
image - is `prefix` the correct property, or is `path`?

philrenaud: Note: also seeing more variables than expected on the non-management list page (templates index)

ui/app/templates/jobs/run/templates/manage.hbs
```diff
@@ -0,0 +1,34 @@
...
+        <p>A plan will be requested before the job is submitted.  You can pick a template or write your own job spec.</p>
+    </header>
+    {{#if (eq this.model.length 0)}}
+        <h3 data-test-empty-templates-list-headline class="empty-message-headline">
+          You have no templates to choose from. Would you like to <Hds::Link::Inline @route="jobs.run.templates.new" data-test-create-inline>create</Hds::Link::Inline> one?
+        </h3>
+        <Hds::Button class="button-group" @text="Back to editor" @route="jobs.run" @icon="arrow-left" data-test-cancel />
+    {{else}}
+    <main class="radio-group" data-test-template-list>
+        <Hds::Table @model={{this.model}} @columns={{ create-columns "Name" "Namespace" "Description"}}>
```
philrenaud, line 14: Hm, the extra abstraction (new helper just for columns) and ergonomics (`<ListTable><t.head>` at least used real `<th>` elements) makes me think that this is the wrong time to bring Hds::Table in.

ui/app/templates/jobs/run/templates/manage.hbs
```diff
@@ -0,0 +1,34 @@
...
+            <:body as |B|>
```
philrenaud, line 15: Could we change `B` to `Template` or something more indicative of what we're iterating over?

ui/app/helpers/create-columns.js
```diff
@@ -0,0 +1,7 @@
+import { helper } from '@ember/component/helper';
+
+function createColumns(positional) {
+  return positional.map((column) => ({ label: column }));
+}
+
+export default helper(createColumns);
```
philrenaud, line 7: I'm not convinced we want a helper just to make tables work with the fields we want, but if we decided we did:
- I'd like this made more generalized (like `map-to-objects` instead of `create-columns`)
- a second argument (either positional or named) for `key`, which would let you customize what you currently have hardcoded as `label` currently

My first suggestion would still be to consider if we really want to go down the road of invoking a helper every time we display a table, though.

philrenaud (changes requested):

ui/app/routes/jobs/run/templates/manage.js
ChaiWithJai: In `mirage` mocks you will see all the variables here because we're using the `/vars` endpoint which does not have any logic for accounting for the `prefix`. Touching our `Mirage` setup would spill over into the `variables` view so I'm not going to touch it until after all the work is complete and we don't have more feature requests or updates on the feature.

Its annoying but we'll have to rely on a Nomad server here.

ChaiWithJai: Hey! Get back to being on vacation and recharge your batteries!!!

ui/app/templates/jobs/run/templates/manage.hbs
ChaiWithJai: This isn't a template. Rather it's a [contextual components](https://guides.emberjs.com/v4.9.0/components/block-content/) that is yielding out `TableRow` and `TableData` components. We could change the name to `Body` to reflect this. And its a pattern used by [other teams](https://github.com/hashicorp/boundary-ui/blob/a6b2acc14d0de00ff5b8aa2a3568f1d57db99393/ui/admin/app/components/form/target/details/index.hbs#L17), in our own codebase and in line with the [official documentation](https://helios.hashicorp.design/components/table?tab=code).

Changing this to `Body` could make sense. But that would defeat the purpose of the `alias` and we could just write `:body.Tr` instead.

Your call. I'm open to suggestions!

ui/app/helpers/create-columns.js
ChaiWithJai: Roger that. Normally we put this logic in our controllers check any controller that is using the `SingleSelectDropdown` feature and you'll see redundant code. IMHO the code isn't adding value when its in the `controller` or `component` because its not real state, just data transformation for working with another component API.

You'll see logic like [this](https://github.com/hashicorp/nomad/blob/main/ui/app/controllers/jobs/index.js#L84-L103) across 26 different views (about 50% of our views).

On pages where we have multiple dropdowns, it really muddies the controller and makes it challenging to find the code that matters. Most of the code in this [file](https://github.com/hashicorp/nomad/blob/main/ui/app/controllers/jobs/index.js) is just derived state that matches the component API for `SingleSelectDropdown` and all the code boils down to is iterating over a list to get a `key` and `label`.

ui/app/templates/jobs/run/templates/manage.hbs
philrenaud: This makes a ton of sense. Thanks for clearing it up for me!

ui/app/templates/jobs/run/templates/manage.hbs
```diff
@@ -11,12 +11,12 @@
...
     {{else}}
     <main class="radio-group" data-test-template-list>
-        <Hds::Table @model={{this.model}} @columns={{ create-columns "Name" "Namespace" "Description"}}>
+        <Hds::Table @model={{this.model}} @columns={{this.columns}}>
             <:body as |B|>
                 <B.Tr>
                     <B.Td>
                         <LinkTo @route="jobs.run.templates.template" @model={{concat B.data.path "@" B.data.namespace}} data-test-edit-template={{B.data.path}}>
-                            {{B.data.path}}
+                            {{invoke-fn this this.formatTemplateLabel B.data.path}}
```
philrenaud, line 19: Nice and reminds me of svelte attributes!

philrenaud (approved):
Looks great, nice work

PR #15819, opened 2023-01-18
Adds docker isolation option

nickwales:
This change enables the --isolation=hyper-v flag in the docker driver for Windows users. 

Prior to this change attempts to use the flag would cause an error in Nomad.

### DIFF
--- a/drivers/docker/driver.go
+++ b/drivers/docker/driver.go
@@ -837,8 +837,8 @@ func (d *Driver) createContainerConfig(task *drivers.TaskConfig, driverConfig *T
 
 	// Only windows supports alternative isolations modes
 	isolationMode := driverConfig.Isolation
-	if runtime.GOOS != "windows" && driverConfig.Isolation != "process" {
-		return c, fmt.Errorf("Failed to create container configuration for image %q, cannot use isolation mode \"%v\" on %v", driverConfig.Image, isolationMode, runtime.GOOS)
+	if runtime.GOOS != "windows" && isolationMode != "" {
+		return c, fmt.Errorf("Failed to create container configuration, cannot use isolation mode \"%s\" on %s", isolationMode, runtime.GOOS)
 	}
 
 	memory, memoryReservation := memoryLimits(driverConfig.MemoryHardLimit, task.Resources.NomadResources.Memory)
--- a/drivers/docker/driver.go
+++ b/drivers/docker/driver.go
@@ -835,6 +835,12 @@ func (d *Driver) createContainerConfig(task *drivers.TaskConfig, driverConfig *T
 		return c, fmt.Errorf("requested runtime %q is not allowed", containerRuntime)
 	}
 
+	// Only windows supports alternative isolations modes
+	isolationMode := driverConfig.Isolation
+	if runtime.GOOS != "windows" && driverConfig.Isolation != "process" {
+		return c, fmt.Errorf("Failed to create container configuration for image %q, cannot use isolation mode \"%v\" on %v", driverConfig.Image, isolationMode, runtime.GOOS)
+	}
+
 	memory, memoryReservation := memoryLimits(driverConfig.MemoryHardLimit, task.Resources.NomadResources.Memory)
 
 	var pidsLimit int64
--- a/drivers/docker/driver.go
+++ b/drivers/docker/driver.go
@@ -865,6 +865,7 @@ func (d *Driver) createContainerConfig(task *drivers.TaskConfig, driverConfig *T
 		// used to share data between different tasks in the same task group.
 		Binds: binds,
 
+		Isolation:    driverConfig.Isolation,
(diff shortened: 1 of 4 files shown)


tgross (changes requested):
@nickwales it looks like the HCL config plumbing here is fine (although I haven't checked this branch out to test it).

But I'm fairly certain we need to set this configuration value on the [HostConfig](https://pkg.go.dev/github.com/fsouza/go-dockerclient#HostConfig) in [`driver.go#L855-L874`](https://github.com/hashicorp/nomad/blob/v1.4.3/drivers/docker/driver.go#L855-L874) as well, otherwise this configuration won't be sent to the Docker API.

Also: can you run `make cl` for this PR to add a changelog entry?

tgross (changes requested):
Hi @nickwales! So I tested this on my machine trying to break it by running with `isolation = "hyper-v"` on Linux. That fails as expect. It results in the following debug log entry:

```
    2023-01-23T14:48:35.465-0500 [DEBUG] client.driver_mgr.docker: failed to create container: driver=docker container_name=http-6a5f3d74-f5cb-1a39-bef2-2981ace92f37 image_name=busybox:1 image_id=busybox:1 attempt=2 error="API error (500): Invalid isolation: \"hyper-v\" - linux only supports 'default'"
```

And _eventually_ the job submitter sees a task event like the following:

```
Recent Events:
Time                       Type             Description
2023-01-23T14:49:47-05:00  Alloc Unhealthy  Unhealthy because of failed task
2023-01-23T14:49:43-05:00  Not Restarting   Error was unrecoverable
2023-01-23T14:49:43-05:00  Driver Failure   failed to create container: API error (500): Invalid isolation: "hyper-v" - linux only supports 'default'
2023-01-23T14:48:35-05:00  Task Setup       Building Task Directory
2023-01-23T14:48:34-05:00  Received         Task received by client
```

But as you can see from the timestamps, we end up having to wait a while to get that error message. Whereas if I quickly hack in a validation into `createContainerConfig`, I get the following:

```
Recent Events:
Time                       Type             Description
2023-01-23T14:52:48-05:00  Alloc Unhealthy  Unhealthy because of failed task
2023-01-23T14:52:44-05:00  Not Restarting   Error was unrecoverable
2023-01-23T14:52:44-05:00  Driver Failure   Failed to create container configuration for image "busybox:1" ("sha256:827365c7baf137228e94bcfc6c47938b4ffde26c68c32bf3d3a7762cd04056a5"): cannot use isolation mode "hyper-v" on Linux, dummy!
2023-01-23T14:52:44-05:00  Task Setup       Building Task Directory
2023-01-23T14:52:44-05:00  Received         Task received by client
```

Which happens basically instantly, which is a much better UX for operators for an unrecoverable error. So let's add some validation based on the client host OS. While we're at it, we can validate that the isolation we've passed in is one of the ones allowed by Docker.

drivers/docker/driver.go
```diff
@@ -835,6 +835,12 @@ func (d *Driver) createContainerConfig(task *drivers.TaskConfig, driverConfig *T
 		return c, fmt.Errorf("requested runtime %q is not allowed", containerRuntime)
 	}
 
+	// Only windows supports alternative isolations modes
+	isolationMode := driverConfig.Isolation
+	if runtime.GOOS != "windows" && driverConfig.Isolation != "process" {
```
tgross, line 840: It doesn't look like the configuration sets `"process"` as the default, so I think it's `""`, right? Won't that cause us to hit this error for every non-Windows OS?

drivers/docker/driver.go
```diff
@@ -835,6 +835,12 @@ func (d *Driver) createContainerConfig(task *drivers.TaskConfig, driverConfig *T
...
+		return c, fmt.Errorf("Failed to create container configuration for image %q, cannot use isolation mode \"%v\" on %v", driverConfig.Image, isolationMode, runtime.GOOS)
```
tgross, line 841: Looking at the other errors in this area, I'm pretty sure the caller wraps this in a top-level error message as it bubbles up. We probably don't need the image name here, either, because this will get bubbled up to a specific taskrunner's logger and the TaskEvent (which is specific to the task too). Also, we know these are strings so we can use `%s` (or `%q` for quoted strings without manually writing the escaping).

```suggestion
		return c, fmt.Errorf("cannot use isolation mode %q on %s", isolationMode, runtime.GOOS)
```

tgross (changes requested):

drivers/docker/driver.go
nickwales: Yes, can confirm now from testing.

drivers/docker/driver.go
nickwales: I see that now and have pushed those changes up.

tgross (approved):
LGTM

PR #15823, opened 2023-01-19
ui: template edit and delete flow

### LINKED ISSUE #15551: [UI, Example Job Templates] New route: jobs/run/templates/:name
image

ChaiWithJai:
Resolves #15551

### DIFF
--- a/ui/app/controllers/jobs/run/templates/template.js
+++ b/ui/app/controllers/jobs/run/templates/template.js
@@ -3,8 +3,10 @@ import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
 import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
+import { task } from 'ember-concurrency';
 
 export default class JobsRunTemplatesController extends Controller {
+  @service flashMessages;
   @service router;
   @service system;
 
@@ -24,33 +26,6 @@ export default class JobsRunTemplatesController extends Controller {
   @action
   toggleModal() {
     this.formModalActive = !this.formModalActive;
-  }
-
-  @action
-  async deleteTemplateAndClose() {
-    try {
-      this.toggleModal();
-      const name = this.model.path;
-      await this.model.destroyRecord();
-
-      this.flashMessages.add({
-        title: 'Job template deleted',
-        message: `${name} successfully deleted`,
-        type: 'success',
-        destroyOnClick: false,
-        timeout: 5000,
-      });
-
-      this.router.transitionTo('jobs.run.templates');
-    } catch (e) {
-      this.flashMessages.add({
-        title: 'Job template could not be deleted.',
-        message: e,
-        type: 'error',
-        destroyOnClick: false,
-        timeout: 5000,
-      });
-    }
   }
 
   @action
@@ -83,4 +58,28 @@ export default class JobsRunTemplatesController extends Controller {
       });
     }
   }
+
+  @task(function* () {
+    try {
+      yield this.model.destroyRecord();
+
+      this.flashMessages.add({
+        title: 'Job template deleted',
+        message: `${this.model.path} successfully deleted`,
+        type: 'success',
+        destroyOnClick: false,
+        timeout: 5000,
+      });
+      this.router.transitionTo('jobs.run.templates.manage');
+    } catch (err) {
+      this.flashMessages.add({
+        title: `Job template could not be deleted.`,
+        message: err,
+        type: 'error',
+        destroyOnClick: false,
+        sticky: true,
+      });
+    }
+  })
+  deleteTemplate;
 }
--- a/ui/app/controllers/jobs/run/templates/template.js
+++ b/ui/app/controllers/jobs/run/templates/template.js
@@ -1,0 +1,85 @@
+import { A } from '@ember/array';
+import Controller from '@ember/controller';
+import { inject as service } from '@ember/service';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class JobsRunTemplatesController extends Controller {
+  @service router;
+  @service system;
+
+  @tracked formModalActive = false;
+
+  keyValues = A([]);
+
+  @action
+  updateKeyValue(key, value) {
+    if (this.keyValues.find((kv) => kv.key === key)) {
+      this.keyValues.find((kv) => kv.key === key).value = value;
+    } else {
+      this.keyValues.pushObject({ key, value });
+    }
+  }
+
+  @action
+  toggleModal() {
+    this.formModalActive = !this.formModalActive;
+  }
+
+  @action
+  async deleteTemplateAndClose() {
+    try {
+      this.toggleModal();
+      await this.model.destroyRecord();
+
+      this.flashMessages.add({
+        title: 'Job template deleted',
+        message: `${this.templateName} successfully deleted`,
+        type: 'success',
+        destroyOnClick: false,
+        timeout: 5000,
+      });
+
+      this.router.transitionTo('jobs.run.templates');
+    } catch (e) {
+      this.flashMessages.add({
+        title: 'Job template could not be deleted.',
+        message: e,
+        type: 'error',
+        destroyOnClick: false,
+        timeout: 5000,
+      });
+    }
+  }
+
+  @action
+  async save(e, overwrite = false) {
+    if (e.type === 'submit') {
+      e.preventDefault();
+    }
+
+    this.model.set('keyValues', this.keyValues);
+
+    try {
+      await this.model.save({ adapterOptions: { overwrite } });
+
+      this.flashMessages.add({
+        title: 'Job template saved',
+        message: `${this.templateName} successfully editted`,
+        type: 'success',
+        destroyOnClick: false,
+        timeout: 5000,
+      });
+
+      this.router.transitionTo('jobs.run.templates');
+    } catch (e) {
+      this.flashMessages.add({
+        title: 'Job template cannot be editted.',
+        message: e,
+        type: 'error',
+        destroyOnClick: false,
+        timeout: 5000,
+      });
+    }
+  }
+}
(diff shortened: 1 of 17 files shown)


ui/app/templates/jobs/run/templates/template.hbs
```diff
@@ -0,0 +1,76 @@
+{{page-title "Edit template"}}
+<section class="section">
+    <header class="run-job-header">
+        <div>
+            <h1 class="title is-3">Edit template</h1>
+        </div>
+        <Hds::Dropdown as |dd|>
+            <dd.ToggleButton @text="Manage" data-test-manage-dd />
+            <dd.Interactive @icon="trash" @text="Delete" @color="critical" {{on "click" this.toggleModal}} data-test-delete />
+        </Hds::Dropdown>
```
philrenaud, line 10: If we're never planning on having a second button here, could we save the operator a click and put the Delete button outside of it?

ui/app/routes/jobs/run/templates/template.js
```diff
@@ -0,0 +1,28 @@
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+import notifyForbidden from 'nomad-ui/utils/notify-forbidden';
+
+export default class RunRoute extends Route {
```
philrenaud, line 5: Naming convention request: I think this is borrowed from routes/jobs/run and duplicated here (jobs/run/templates/template) as well as jobs/run/templates.

Could we use the more standard RunTemplatesRoute and RunTemplatesTemplateRoute for easier grepping?

ui/app/controllers/jobs/run/templates/template.js
```diff
@@ -0,0 +1,86 @@
...
+      this.flashMessages.add({
+        title: 'Job template cannot be editted.',
+        message: e,
+        type: 'error',
+        destroyOnClick: false,
+        timeout: 5000,
+      });
+    }
+  }
+}
```
philrenaud, line 86: Should we prevent save unless the user has added _some_ template? We don't have to do careful hcl parsing or anything, but we should at least ensure they haven't accidentally left this blank.

Preferred place would be as a `disabled` condition on the `save1 button in template.

philrenaud (changes requested):

ui/app/templates/jobs/run/templates/template.hbs
ChaiWithJai: I'm matching the pattern that we see in `cloud-ui`, `vault` and `boundary` here.

I'm imagining we will eventually be adding to this manage dropdown because we may decide that we want to be a read-only view by default across the application now that we have more editing scenarios or say we want to allow additional key value pairs to be added later on.

Generally, I like the pattern of the managerial buttons being grouped and stored together in a nice dropdown.

Screen Shot 2023-01-20 at 11 34 56 AM

Screen Shot 2023-01-20 at 11 34 30 AM

ui/app/templates/jobs/run/templates/template.hbs
```diff
@@ -0,0 +1,76 @@
...
+        <M.Footer as |F|>
+        <Hds::ButtonSet>
+            <Hds::Button type="submit" @text="Confirm"
+            {{on "click" this.deleteTemplateAndClose}}
+            data-test-delete-template
+            />
+            <Hds::Button type="button" @text="Cancel" @color="secondary" {{on "click" F.close}} />
+        </Hds::ButtonSet>
+        </M.Footer>
+    </Hds::Modal>
```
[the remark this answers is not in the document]
ChaiWithJai: I'm trying to match the patterns of the other teams here. Every other team is using this pattern. Personally, I think 2-step delete was created when we have a super clunky view like the Client Detail Overview and you need to take an action.

HashiCorp Design System advises to not use a button "to display multiple actions under a single button, use [Dropdown](https://helios.hashicorp.design/components/dropdown)." The 2-step button in this location appears to be a violation of those guidelines and out of sync with the other applications.

The dimming effect should be an easy patch.

Screen Shot 2023-01-20 at 11 42 18 AM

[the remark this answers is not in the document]
ChaiWithJai: Patched the styling issue!

ui/app/controllers/jobs/run/templates/template.js
```diff
@@ -0,0 +1,86 @@
+import { A } from '@ember/array';
+import Controller from '@ember/controller';
+import { inject as service } from '@ember/service';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class JobsRunTemplatesController extends Controller {
```
ChaiWithJai, line 7: Will address naming when I patch delete in row functionality.

ui/app/routes/jobs/run/templates/template.js
ChaiWithJai: Will patch this when I throw the commit up for deleting inside of a row.

ui/app/templates/jobs/run/templates/template.hbs
philrenaud: Right, but the button in boundary, for example, has further examples other than just Delete (has Add Members etc.)

I like dropdown buttons for sets of actions too, but it's adding unnecessary overhead if there's just one thing you can do beneath it.

philrenaud: Also just noting: is there no way in HDS for me to use keyboard navigation in a dropdown? Opening the Manage button here doesn't give me the ability to select Delete with my keyboard that I can tell.

philrenaud (changes requested):
It looks like some of the recent changes (z-index changes to account for HDS' modal pattern) are causing breaks in other places
image

I'd like to strongly suggest not overlapping this feature with HDS inclusion; we're seeing that that deserves its own dedicated effort.

ui/app/templates/jobs/run/templates/template.hbs
ChaiWithJai: I'm successfully using the normal pattern for using tabs to access the manage dropdown. But this sounds like a question to drop into the `#design-systems` Slack channel. I'm typing `tab` + `enter` + `tab` + `enter` to complete the entire action of deleting a template.

ChaiWithJai: > Right, but the button in boundary, for example, has further examples other than just Delete (has Add Members etc.)
>
> I like dropdown buttons for sets of actions too, but it's adding unnecessary overhead if there's just one thing you can do beneath it.

OK, no worries! 😅 I don't have any strong opinions here -- to clarify:  are you saying that you want me to change this because you don't like it? I'm totally happy to accommodate here! 

If you don't like this pattern I can just put a delete button next to the edit button. Is that what you're looking for?

Its a little challenging for me to move forward on this PR because I'm unsure of about the language of the comment. It would be a little more helpful for me to know if we're having a discussion or if this is something that you're requesting a change on.

Just tell me what you're looking for and I'll gladly make the change! 😄

ChaiWithJai:
> It looks like some of the recent changes (z-index changes to account for HDS' modal pattern) are causing breaks in other places image
> 
> I'd like to strongly suggest not overlapping this feature with HDS inclusion; we're seeing that that deserves its own dedicated effort.

OK! Do want me to re-do the PR? I think our `z-index` woes are fairly trivial here. But I can just re-do this part feature because we're already ahead of schedule. LMK what you want to see.

ui/app/templates/jobs/run/templates/template.hbs
philrenaud: > If you don't like this pattern I can just put a delete button next to the edit button. Is that what you're looking for?
Am I missing something? I don't see any edit button for this: 
image

I'd like to just see the delete button here, no manage button that you have to click first.

philrenaud:
> OK! Do want me to re-do the PR? 

I think you can keep this PR, but I would use the conventions we've typically used in Nomad. The changes in behaviour (buttons, dropdowns) and styles (tables) on just these pages is going to be jarring to users.

ui/app/templates/jobs/run/templates/manage.hbs
```diff
@@ -21,6 +21,17 @@
...
+                        <TwoStepButton
+                            data-test-delete
+                            @idleText="Delete"
+                            @cancelText="Cancel"
+                            @confirmText="Yes, Delete Text"
+                            @awaitingConfirmation={{this.deleteTemplate.isRunning}}
+                            @disabled={{cannot "destroy variable" namespace="*"}}
+                            @onConfirm={{perform this.deleteTemplate B.data}}
+                        />
+                    </B.Td>
```
ChaiWithJai, line 34: @philrenaud the delete button causes a jumpy column, what should I do here?

ui/app/controllers/jobs/run/templates/template.js
ChaiWithJai: Will get this after we confirm that I addressed the real changes. My bad.

ui/app/templates/jobs/run/templates/template.hbs
```diff
@@ -0,0 +1,82 @@
...
+    <header class="run-job-header">
+        <div>
+            <h1 class="title is-3">Edit template</h1>
+        </div>
+        <TwoStepButton
+            data-test-delete
+            @alignRight={{true}}
+            @idleText="Delete"
+            @cancelText="Cancel"
+            @confirmText="Yes, Delete Text"
```
philrenaud, line 12:
```suggestion
            @confirmText="Yes, Delete Template"
            @inlineText={{true}}
            @confirmationMessage="Are you sure?"
```

ui/tests/acceptance/job-run-test.js
```diff
@@ -434,6 +434,7 @@ module('Acceptance | job run', function (hooks) {
         path: 'nomad/job-templates/foo',
         namespace: 'default',
         id: 'nomad/job-templates/foo',
+        Items: {},
       });
```
philrenaud, line 438: Would like to see some tests for delete / edit template. Future PR is fine.

philrenaud (approved):
Looks good! Made some suggestions for making the delete confirmations more understandable for users, and would like to see a few tests here and route name changes. But the logic and the flow feels great!

PR #15835, opened 2023-01-20
Rename `nomad.broker.total_blocked` metric

tgross:
Fixes https://github.com/hashicorp/nomad/issues/6480 and addresses comments from https://github.com/hashicorp/nomad/pull/14621#discussion_r1023293596

This changeset fixes a long-standing point of confusion in metrics emitted by the eval broker. The eval broker has a queue of "blocked" evals that are waiting for an in-flight ("unacked") eval of the same job to be completed. But this "blocked" state is not the same as the `blocked` status that we write to raft and expose in the Nomad API to end users. There's a second metric `nomad.blocked_eval.total_blocked` that refers to evaluations in that state. This has caused ongoing confusion in major customer incidents and even in our own documentation! (Fixed in this PR.)

There's little functional change in this PR aside from the name of the metric emitted, but there's a bit refactoring to clean up the names in `eval_broker.go` so that there aren't name collisions and multiple names for the same state. Changes included are:
* Everything that was previously called "pending" referred to entities that were associated witht he "ready" metric. These are all now called "ready" to match the metric.
* Everything named "blocked" in `eval_broker.go` is now named "pending", except for a couple of comments that actually refer to blocked RPCs.
* Added a note to the upgrade guide docs for 1.5.0.
* Fixed the [scheduling performance metrics](https://developer.hashicorp.com/nomad/docs/operations/monitoring-nomad#performance) docs because the description for `nomad.broker.total_blocked` was actually the description for `nomad.blocked_eval.total_blocked`.

Once this is reviewed I'll open a second PR backporting just the docs fixes.

### DIFF
--- a/nomad/eval_broker.go
+++ b/nomad/eval_broker.go
@@ -71,7 +71,7 @@ type EvalBroker struct {
 	cancelable []*structs.Evaluation
 
 	// ready tracks the ready jobs by scheduler in a priority queue
-	ready map[string]PendingEvaluations
+	ready map[string]ReadyEvaluations
 
 	// unack is a map of evalID to an un-acknowledged evaluation
 	unack map[string]*unackEval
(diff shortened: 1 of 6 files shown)


nomad/eval_broker.go
```diff
@@ -896,7 +895,7 @@ func (b *EvalBroker) EmitStats(period time.Duration, stopCh <-chan struct{}) {
 			stats := b.Stats()
 			metrics.SetGauge([]string{"nomad", "broker", "total_ready"}, float32(stats.TotalReady))
 			metrics.SetGauge([]string{"nomad", "broker", "total_unacked"}, float32(stats.TotalUnacked))
-			metrics.SetGauge([]string{"nomad", "broker", "total_blocked"}, float32(stats.TotalBlocked))
+			metrics.SetGauge([]string{"nomad", "broker", "total_pending"}, float32(stats.TotalPending))
```
schmichael, line 898: :tada:

PR #15836, opened 2023-01-20
acl: CLI login command should not default to "OIDC" type

pkazmierczak:
Since we're expecting to be supporting more auth method types soon, we don't want the CLI to default to "OIDC." It would be poor UX if this behavior suddenly changed.

Relates to #13120


pkazmierczak:
closing in favor of https://github.com/hashicorp/nomad/pull/15869

PR #15841, opened 2023-01-20
docs: backport fix for metrics on blocked_evals vs broker

tgross:
In #15835 we renamed the `nomad.broker.total_blocked` metric to `nomad.broker.total_pending`, but in the process identified that the existing scheduling performance monitoring guide mixed up the `broker.total_blocked` metric with the `blocked_evals.total_blocked` metric. This changeset backports the fix to the docs without renaming the metric (for backwards compatibility).

### DIFF
--- a/website/content/docs/operations/monitoring-nomad.mdx
+++ b/website/content/docs/operations/monitoring-nomad.mdx
@@ -187,16 +187,14 @@ points in the scheduling process.
   evaluation at a time, entirely in-memory. If this metric increases,
   examine the CPU and memory resources of the scheduler.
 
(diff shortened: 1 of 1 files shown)


gulducat (approved):
Good on ya updating past docs.

This will be reflected on developer.hashicorp.com when using the little version dropdown yeah?  e.g. here for v1.3.x: https://developer.hashicorp.com/nomad/docs/v1.3.x/operations/monitoring-nomad#performance

tgross:
> This will be reflected on developer.hashicorp.com when using the little version dropdown yeah?

Right. I've added the backport labels to this PR so that'll get done automagically.

PR #15844, opened 2023-01-21
Dynamic Node Metadata

### LINKED ISSUE #14617: Dynamic node metadata
### Proposal

Adding a `HTTP PUT` for `/v1/node/:node_id/meta` to update the metadata stanza of that node. 

Potential issues: what if the node reloads? Which metadata should be used? One solution would be to always use the "latest", e.g. upon reloading the file-config, using that which is defined. Ideally, the dynamic metadata would persist a reload/restart. 

### Use-cases

The end-goal would be to set dynamic properties for specific nodes, to micro-manage the job allocation process a bit. For example dynamically figuring out which nodes are acting as load-balancers / ingress, and then marking those nodes as such using the API without restarting nomad. (Mostly because restarting is VERY destructive, restarting/rescheduling all jobs that are present.)

That way, this metadata can be used by new jobs for scheduling (e.g. `affinity`). 

### Attempted Solutions

Manually tagging all nodes that might potentially end up being a load-balancer (and then restarting the nodes), but I'd rather not restart my nodes every time I change that metadata. 

An alternative solution (e.g. a different feature request, since this isn't possible as far as I know) to my use-case would be to reference other jobs in the `affinity` stanza. To tell it to "please allocate me next to this job", or "please do not allocate me next to that job". (Within the same namespace, I'd assume.) As a `constraint` that'd potentially cause issues, since then you could create mutually-dependent jobs - but for `affinity` that should be fine. 

Another alternative solution (that doesn't quite work yet) would be to support the metadata as part of `systemctl reload nomad`.

schmichael:
Fixes #14617

**WIP**

Dynamic Node Metadata allows Nomad users, and their jobs, to update Node metadata through an API. Currently Node metadata is only reloaded when a Client agent is restarted.

This is a draft implementation using a Client-agent API and state and is intended for use by node-local system jobs that provide extra monitoring and node annotation over Nomad's fingerprinting facilities.

Cons:
1. The Node being manipulated must be online.
2. There is up to 10 seconds of latency before schedulers use the new metadata.

Pros:
1. Servers are protected from DDoS if a whole fleet of system jobs start mutating metadata simultaneously or in a tight loop.
2. The API works even if the node is disconnected from servers, so a monitoring system job can continue to update the metadata locally and it will be propagated as soon as connectivity is restored. This is similar to Consul's Agent API anti-entropy.

**TODO**
* [ ] Docs
* [ ] Tests

### DIFF
--- a/api/acl.go
+++ b/api/acl.go
@@ -31,7 +31,7 @@ func (a *ACLPolicies) Upsert(policy *ACLPolicy, q *WriteOptions) (*WriteMeta, er
 	if policy == nil || policy.Name == "" {
 		return nil, errors.New("missing policy name")
 	}
-	wm, err := a.client.write("/v1/acl/policy/"+policy.Name, policy, nil, q)
+	wm, err := a.client.put("/v1/acl/policy/"+policy.Name, policy, nil, q)
 	if err != nil {
 		return nil, err
 	}
@@ -77,7 +77,7 @@ func (c *Client) ACLTokens() *ACLTokens {
 // Bootstrap is used to get the initial bootstrap token
 func (a *ACLTokens) Bootstrap(q *WriteOptions) (*ACLToken, *WriteMeta, error) {
 	var resp ACLToken
-	wm, err := a.client.write("/v1/acl/bootstrap", nil, &resp, q)
+	wm, err := a.client.put("/v1/acl/bootstrap", nil, &resp, q)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -94,7 +94,7 @@ func (a *ACLTokens) BootstrapOpts(btoken string, q *WriteOptions) (*ACLToken, *W
 	}
 
 	var resp ACLToken
-	wm, err := a.client.write("/v1/acl/bootstrap", req, &resp, q)
+	wm, err := a.client.put("/v1/acl/bootstrap", req, &resp, q)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -117,7 +117,7 @@ func (a *ACLTokens) Create(token *ACLToken, q *WriteOptions) (*ACLToken, *WriteM
 		return nil, nil, errors.New("cannot specify Accessor ID")
 	}
 	var resp ACLToken
-	wm, err := a.client.write("/v1/acl/token", token, &resp, q)
+	wm, err := a.client.put("/v1/acl/token", token, &resp, q)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -130,7 +130,7 @@ func (a *ACLTokens) Update(token *ACLToken, q *WriteOptions) (*ACLToken, *WriteM
 		return nil, nil, errors.New("missing accessor ID")
 	}
 	var resp ACLToken
-	wm, err := a.client.write("/v1/acl/token/"+token.AccessorID,
+	wm, err := a.client.put("/v1/acl/token/"+token.AccessorID,
 		token, &resp, q)
 	if err != nil {
 		return nil, nil, err
@@ -176,7 +176,7 @@ func (a *ACLTokens) Self(q *QueryOptions) (*ACLToken, *QueryMeta, error) {
 // UpsertOneTimeToken is used to create a one-time token
 func (a *ACLTokens) UpsertOneTimeToken(q *WriteOptions) (*OneTimeToken, *WriteMeta, error) {
 	var resp *OneTimeTokenUpsertResponse
-	wm, err := a.client.write("/v1/acl/token/onetime", nil, &resp, q)
+	wm, err := a.client.put("/v1/acl/token/onetime", nil, &resp, q)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -193,7 +193,7 @@ func (a *ACLTokens) ExchangeOneTimeToken(secret string, q *WriteOptions) (*ACLTo
 	}
 	req := &OneTimeTokenExchangeRequest{OneTimeSecretID: secret}
 	var resp *OneTimeTokenExchangeResponse
-	wm, err := a.client.write("/v1/acl/token/onetime/exchange", req, &resp, q)
+	wm, err := a.client.put("/v1/acl/token/onetime/exchange", req, &resp, q)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -243,7 +243,7 @@ func (a *ACLRoles) Create(role *ACLRole, w *WriteOptions) (*ACLRole, *WriteMeta,
 		return nil, nil, errors.New("cannot specify ACL role ID")
 	}
 	var resp ACLRole
-	wm, err := a.client.write("/v1/acl/role", role, &resp, w)
+	wm, err := a.client.put("/v1/acl/role", role, &resp, w)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -256,7 +256,7 @@ func (a *ACLRoles) Update(role *ACLRole, w *WriteOptions) (*ACLRole, *WriteMeta,
 		return nil, nil, errMissingACLRoleID
 	}
 	var resp ACLRole
-	wm, err := a.client.write("/v1/acl/role/"+role.ID, role, &resp, w)
+	wm, err := a.client.put("/v1/acl/role/"+role.ID, role, &resp, w)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -328,7 +328,7 @@ func (a *ACLAuthMethods) Create(authMethod *ACLAuthMethod, w *WriteOptions) (*AC
 		return nil, nil, errMissingACLAuthMethodName
 	}
 	var resp ACLAuthMethod
-	wm, err := a.client.write("/v1/acl/auth-method", authMethod, &resp, w)
+	wm, err := a.client.put("/v1/acl/auth-method", authMethod, &resp, w)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -341,7 +341,7 @@ func (a *ACLAuthMethods) Update(authMethod *ACLAuthMethod, w *WriteOptions) (*AC
 		return nil, nil, errMissingACLAuthMethodName
 	}
 	var resp ACLAuthMethod
-	wm, err := a.client.write("/v1/acl/auth-method/"+authMethod.Name, authMethod, &resp, w)
+	wm, err := a.client.put("/v1/acl/auth-method/"+authMethod.Name, authMethod, &resp, w)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -397,7 +397,7 @@ func (a *ACLBindingRules) List(q *QueryOptions) ([]*ACLBindingRuleListStub, *Que
 // Create is used to create an ACL binding rule.
 func (a *ACLBindingRules) Create(bindingRule *ACLBindingRule, w *WriteOptions) (*ACLBindingRule, *WriteMeta, error) {
 	var resp ACLBindingRule
-	wm, err := a.client.write("/v1/acl/binding-rule", bindingRule, &resp, w)
+	wm, err := a.client.put("/v1/acl/binding-rule", bindingRule, &resp, w)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -410,7 +410,7 @@ func (a *ACLBindingRules) Update(bindingRule *ACLBindingRule, w *WriteOptions) (
 		return nil, nil, errMissingACLBindingRuleID
 	}
 	var resp ACLBindingRule
-	wm, err := a.client.write("/v1/acl/binding-rule/"+bindingRule.ID, bindingRule, &resp, w)
+	wm, err := a.client.put("/v1/acl/binding-rule/"+bindingRule.ID, bindingRule, &resp, w)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/api/acl.go
+++ b/api/acl.go
@@ -456,7 +456,7 @@ func (c *Client) ACLOIDC() *ACLOIDC {
 // be visited in order to sign in to the provider.
 func (a *ACLOIDC) GetAuthURL(req *ACLOIDCAuthURLRequest, q *WriteOptions) (*ACLOIDCAuthURLResponse, *WriteMeta, error) {
 	var resp ACLOIDCAuthURLResponse
-	wm, err := a.client.write("/v1/acl/oidc/auth-url", req, &resp, q)
+	wm, err := a.client.put("/v1/acl/oidc/auth-url", req, &resp, q)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -467,7 +467,7 @@ func (a *ACLOIDC) GetAuthURL(req *ACLOIDCAuthURLRequest, q *WriteOptions) (*ACLO
 // appropriate claims attached.
 func (a *ACLOIDC) CompleteAuth(req *ACLOIDCCompleteAuthRequest, q *WriteOptions) (*ACLToken, *WriteMeta, error) {
 	var resp ACLToken
-	wm, err := a.client.write("/v1/acl/oidc/complete-auth", req, &resp, q)
+	wm, err := a.client.put("/v1/acl/oidc/complete-auth", req, &resp, q)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/api/node_meta.go
+++ b/api/node_meta.go
@@ -25,7 +25,7 @@ func (n *Nodes) Meta() *NodeMeta {
 // receiving the request is modified.
 func (n *NodeMeta) Apply(meta *NodeMetaApplyRequest, qo *WriteOptions) (*NodeMetaResponse, error) {
 	var out NodeMetaResponse
-	_, err := n.client.write("/v1/client/metadata", meta, &out, qo)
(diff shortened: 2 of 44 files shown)


client/client.go
```diff
@@ -1521,6 +1547,34 @@ func (c *Client) setupNode() error {
...
+	for dk, dv := range dynamicMeta {
+		if dv == nil {
+			_, ok := node.Meta[dk]
+			if ok {
+				// Unset static node metadata
+				delete(node.Meta, dk)
+			} else {
+				// Forget dynamic node metadata tombstone as there's no
+				// static value to erase.
+				delete(dynamicMeta, dk)
```
tgross, line 1564: Why throw this value away? It doesn't have an impact on the resulting merged metadata, but if we discard the tombstone then a subsequent config update (from restart) can cause the field to reappear without any trace of having discarded it previously. That seems like it could be confusing. (And the operator can always submit a new `nomad node config set` to wipe out any unwanted tombstones, but then that's under their control.)

tgross (commented):
Hi @schmichael! I took a pass over this at the architectural level and it's looking pretty good! I've left a couple of comments here and on the RFC, but nothing that should be show-stopping.

command/node_config.go
```diff
@@ -13,11 +13,11 @@ type NodeConfigCommand struct {
...
 	helpText := `
-Usage: nomad node config [options]
+Usage: nomad node meta [options]
 
-  View or modify a client node's configuration details. This command only works
-  on client nodes, and can be used to update the running client configurations
-  it supports.
+  View or modify a client node's metadata. This command only works
+  on client nodes, and can be used to update the scheduling metadata the node
+  registers.
```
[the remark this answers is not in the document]
schmichael: Yikes, yeah, sorry about that. Must have edited the wrong file. And yeah, I was surprised to stumble across this command to.

Reverted.

client/client.go
schmichael: > And the operator can always submit a new `nomad node config set` to wipe out any unwanted tombstones, but then that's under their control.

(`nomad node meta apply` now). They can't, at least right now. Since `unset` and `empty` are both valid values for node metadata keys, we don't have a distinct delete/purge/forget/use-static option yet... This pruning seemed simpler than explaining the difference between `unset` and a hypothetical `delete` but it seems more like I'm making *implicit* complexity instead of *explicit* complexity.

I think implicit pruning is fairly safe because it only does something unexpected in the following situation:

1. Node has `foo="i'll be back"` set in HCL
2. Node has `foo=null` from the API
3. *Operator removes `foo` from HCL entirely*
4. Agent restarts, pruning occurs
5. Operator adds `foo="helo"` back into HCL
6. Agent restarts and `foo="helo"` appears

I just have trouble imagining a case in which this happens and the result is particularly surprising. If anything it seems most likely to occur when considerable time passes between 4 & 5, and the end result is what the operator intends (`foo` to come back)... but idk... I can also see how this could be magical and confusing.

tgross: Fair enough. This solution effectively garbage-collects unused keys too. :+1:

api/acl.go
```diff
@@ -31,7 +31,7 @@ func (a *ACLPolicies) Upsert(policy *ACLPolicy, q *WriteOptions) (*WriteMeta, er
 	if policy == nil || policy.Name == "" {
 		return nil, errors.New("missing policy name")
 	}
-	wm, err := a.client.write("/v1/acl/policy/"+policy.Name, policy, nil, q)
+	wm, err := a.client.put("/v1/acl/policy/"+policy.Name, policy, nil, q)
```
schmichael, line 34: Sorry for sneaking in this little refactor... I don't think I've ever noticed we only ever did PUTs! I really don't want to refight the REST Wars, but I just couldn't bear to put my explicitly not-PUT-like API behind a PUT.

api/node_meta.go
```diff
@@ -0,0 +1,59 @@
...
+func (n *NodeMeta) Set(meta *NodeMetaSetRequest, qo *WriteOptions) (*NodeMetaResponse, error) {
+	var out NodeMetaResponse
+	_, err := n.client.post("/v1/client/metadata", meta, &out, qo)
+	if err != nil {
+		return nil, err
+	}
+	return &out, nil
+}
+
+// Read Node metadata (dynamic and static merged) from a Node directly. May
```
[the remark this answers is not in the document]
schmichael: It is. The current logic is that the merged metadata is returned (which will _eventually_ match the server's Node.Meta) + Static & Dynamic maps of those individual values to hopefully make it really clear where each value is coming from.

client/client.go
```diff
@@ -743,7 +749,9 @@ func (c *Client) Reload(newConfig *config.Config) error {
 	}
 
 	if shouldReloadTLS {
-		return c.reloadTLSConnections(newConfig.TLSConfig)
+		if err := c.reloadTLSConnections(newConfig.TLSConfig); err != nil {
+			return err
+		}
 	}
 
 	c.fingerprintManager.Reload()
```
schmichael, line 757: Unrelated minor fix. In theory if an operator was updating tls settings and something that affected fingerprinting in one fell swoop, we could have missed whatever affected fingerprinting.

e2e/workload_id/workload_id_test.go
```diff
@@ -1,4 +1,4 @@
-package main
+package workload_id
```
schmichael, line 1: Fixing an oopsie from my last couple PRs. :grimacing:

shoenig:
@schmichael I believe these permission denied errors got wrapped during the grpc refactoring, maybe try `must.ErrorIs` instead?

```
assert.go:11: 
    meta_endpoint_test.go:40: expected matching error strings
    ↪ msg: "Permission denied"
    ↪ err: "rpc error: Permission denied"
```

api/node_meta.go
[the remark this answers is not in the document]
tgross: :+1:

command/node_meta_apply.go
```diff
@@ -0,0 +1,119 @@
...
+	meta := make(map[string]*string)
+	for _, pair := range args {
+		kv := strings.SplitN(pair, "=", 2)
+		meta[kv[0]] = &kv[1]
+	}
+	for _, k := range strings.Split(unset, ",") {
+		if k != "" {
+			meta[k] = nil
+		}
+	}
```
shoenig, line 94: would be good to get some tests around this - not just to prove correctness but also the cases will self describe the expected inputs/outputs

command/node_meta_apply.go
```diff
@@ -0,0 +1,119 @@
...
+
+type NodeMetaApplyCommand struct {
+	Meta
+}
+
+func (c *NodeMetaApplyCommand) Help() string {
+	helpText := `
+Usage: nomad node meta apply [-node-id ...] [-unset ...] key1=value1 ... kN=vN
+
+  Modify a node's metadata. This command only works on client agents, and can
```
shoenig, line 19:
```suggestion
  Modify a node's metadata. This command only applies to client agents, and can
```

"only works on" sounds like you cannot run this command against a server (but you can, with -node-id, right?)

command/node_meta_read.go
```diff
@@ -0,0 +1,142 @@
...
+
+General Options:
+
+  ` + generalOptionsUsage(usageOptsDefault|usageOptsNoNamespace) + `
+
+Node Meta Options:
+
+  -node-id
+		Reads metadata from the specified node. If not specified the node receiving
+		the request will be used by default.
```
shoenig, line 36: indentation doesn't match the other options

nomad/client_rpc.go
```diff
@@ -212,6 +212,40 @@ func (s *Server) serverWithNodeConn(nodeID, region string) (*serverParts, error)
 	return mostRecentServer, nil
 }
 
+// forwardClientRPC forwards the RPC specified by method to the node specified
+// by nodeID. Must be done after region forwarding, metrics, and permissions
+// checks.
+//
+// This is a wrapper method for getNodeForRpc, getNodeConn, etc that Client
+// RPCs which only need Servers to forward requests can use.
+func (s *Server) forwardClientRPC(method, nodeID string, args, reply any) error {
```
tgross, line 221: Nice, we have this pattern in a few places in `client_csi_endpoint.go` where we could use this too.

nomad/client_meta_endpoint.go
```diff
@@ -0,0 +1,69 @@
+package nomad
+
+import (
+	"time"
+
+	metrics "github.com/armon/go-metrics"
+	log "github.com/hashicorp/go-hclog"
```
shoenig, line 7:
```suggestion
	"github.com/armon/go-metrics"
	"github.com/hashicorp/go-hclog"
```

command/node_meta_read.go
```diff
@@ -0,0 +1,142 @@
...
+Usage: nomad node meta read [-json] [-node-id ...]
+
+	Read a node's metadata. This command only works on client agents. The node
+	status command can be used to retrieve node metadata from any agent.
+
+  Changes via the "node meta apply" subcommand are batched and may take up to
+  10 seconds to propagate to the servers and affect scheduling. This command
+  will always return the most recent node metadata while the "node status"
+  command can be used to view the metadata that is currently being used for
+  scheduling.
```
tgross, line 26: It's hard to tell for sure in the PR reviewer but I think these two paragraphs aren't intended the same. The top one looks like it has tabs?

client/meta_endpoint.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	for k := range args.Meta {
+		if k == "" {
+			return structs.NewErrRPCCoded(http.StatusBadRequest, "metadata keys must not be empty")
+		}
+
+		// Actually do a modicum of validation
+		if strings.Contains(k, "*") {
+			return structs.NewErrRPCCoded(http.StatusBadRequest, "metadata keys cannot contain *")
+		}
```
tgross, line 44: Do we want to be more aggressive about this constraint, considering these values will want to be interpolated with `${meta.<key>}`? 

On the one hand I don't think we've formalized the constraint in docs so existing metadata in HCL might not match any validation we do here. But on the other hand, it's easier to open up the space of characters allowed later than the other way around. On the other other hand, _maybe_ we can assume cluster administrators know what they want here?

e2e/workload_id/nodemeta_test.go
```diff
@@ -0,0 +1,139 @@
...
+			for _, alloc = range allocs {
+				if alloc.ClientStatus == "complete" {
+					found = true
+					break
+				}
+			}
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
+	must.True(t, found, must.Sprintf("did not find completed alloc"))
```
tgross, line 129: Not for this PR, but there's a `e2eutil.WaitForAllocStatusExpected` which doesn't _quite_ have the API we want I think, because there's a bunch of these little loops all over the e2e tests. Something we should regularize at some point.

tgross:
(bah, I didn't refresh the page before submitting my review, which means @shoenig said a lot of the same things I did :blush: )

e2e/workload_id/nodemeta_test.go
schmichael: Agreed, but I also spend a lot of time sprinkling logs and other checks in these loops when iterating on tests.

I think spending some time stuffing those shared funcs full of The Usual Debugging Stuff (eg if an alloc fails, dump Task States and maybe even the last log lines; if a placement fails, dump evals). You may even notice some loops I use WaitIndex while this one I didn't... because it could fail in a way that all of the returned indexes were 0 and handling that was harder than just polling. :grimacing: I think getting some helpers in our *api* package to make some of these common patterns easier to access would payoff for users and tests alike.

I have some more tests I'd like to write post-beta, so we'll see if I make any progress... anything in the *api* package is likely too big of an effort to sneak in though as we really have to make sure to craft the right behavior+api.

nomad/client_meta_endpoint.go
schmichael: Huh... wonder if my editor is misconfigured. Thanks.

api/node_meta_test.go
```diff
@@ -0,0 +1,72 @@
...
+	metaResp, err := c.Nodes().Meta().Read(node.ID, nil)
+	must.NoError(t, err)
+	must.MapEq(t, node.Meta, metaResp.Meta)
+	must.MapEq(t, node.Meta, metaResp.Static)
+	must.MapEmpty(t, metaResp.Dynamic)
+
+	staticKey := ""
+	for staticKey = range node.Meta {
+		break
+	}
```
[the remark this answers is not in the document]
schmichael: Except `nodeMeta` is a map, so I'm just plucking a random key out.

api/node_meta_test.go
```diff
@@ -0,0 +1,72 @@
...
+	must.Eq(t, "bar", metaResp.Meta["foo"])
+
+	// Wait up to 10s (plus buffer) for node to re-register
+	deadline = time.Now().Add(11 * time.Second)
+	found := false
+	for !found && time.Now().Before(deadline) {
+		node, _, err = c.Nodes().Info(node.ID, nil)
+		must.NoError(t, err)
+		found = node.Meta["foo"] == "bar"
+	}
```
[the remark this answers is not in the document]
schmichael: I saw that but it seemed like about the same amount of code and more indirect. As I rambled in a comment to Tim: What I'd really like is an *api package* helper that made our WaitIndex API easier to use as that keeps the agent logs from getting spammed with GETs and minimizes latency between the change and detecting it in the test.

client/meta_endpoint_test.go
```diff
@@ -0,0 +1,95 @@
...
+	applyReq := &structs.NodeMetaApplyRequest{
+		NodeID: c1.NodeID(),
+		Meta: map[string]*string{
+			"foo": pointer.Of("bar"),
+		},
+	}
+
+	resp := structs.NodeMetaResponse{}
+	err := c1.ClientRPC("NodeMeta.Apply", applyReq, &resp)
+	must.EqError(t, err, structs.ErrPermissionDenied.Error())
```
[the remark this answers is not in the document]
schmichael: Sadly no, that doesn't work at all. Seems like this error message changes depending on whether the test is run as root or non-root, so I'm switching to ErrorContains.

command/node_meta_apply.go
schmichael: `-node-id` only routes to client agents, but I like your wording better. Thanks!

client/meta_endpoint.go
schmichael: https://pkg.go.dev/github.com/hashicorp/hcl/v2@v2.16.0/hclsyntax#ValidIdentifier has a nice interface and the spec seems reasonable: https://github.com/hashicorp/hcl/blob/main/hclsyntax/spec.md#identifiers

Seems like a reasonable thing to try.

command/node_meta_apply.go
schmichael: Good call, Tim found a panic in it!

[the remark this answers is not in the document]
schmichael: Yikes! Great catch. Fixed and added a table test.

client/meta_endpoint.go
schmichael: Just a note that I used that hcl function, but I split on dots as Nomad itself uses dotted identifiers in node metadata (eg `connect.log_level`). Added a table test around the validation.

command/node_meta_apply_test.go
```diff
@@ -0,0 +1,145 @@
...
+		{
+			name:  "WeirdArgs",
+			input: []string{"=", "foo==bar"},
+			exp: map[string]*string{
+				"":    pointer.Of(""),
+				"foo": pointer.Of("=bar"),
+			},
+		},
+		{
+			name:  "WeirderArgs",
```
tgross, line 60: > WeirdArgs
> WeirderArgs
:laughing:

command/node_meta_apply_test.go
```diff
@@ -0,0 +1,145 @@
+package command
+
+import (
+	"testing"
+
+	"github.com/hashicorp/nomad/helper/pointer"
+	"github.com/shoenig/test/must"
+)
+
+func TestNodeMeta_parseMapFromArgs(t *testing.T) {
```
shoenig, line 10: ci.Parallel

command/node_meta_apply_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	for i := range cases {
+		tc := cases[i]
+		t.Run(tc.name, func(t *testing.T) {
+			must.MapEq(t, tc.exp, parseMapFromArgs(tc.input))
+		})
+	}
+}
+
+func TestNodeMeta_applyNodeMetaUnset(t *testing.T) {
```
shoenig, line 77: Parallel

nomad/structs/node_test.go
```diff
@@ -63,3 +64,97 @@ func TestDriverInfoEquals(t *testing.T) {
 		require.Equal(testCase.expected, first.HealthCheckEquals(second), testCase.errorMsg)
 	}
 }
+
+func TestNodeMeta_Validate(t *testing.T) {
```
shoenig, line 68: parallel

shoenig (approved):
LGTM! just missing parallel in tests

Didn't look at anything ui/

tgross (approved):
LGTM (didn't look at the UI work, which was previously-reviewed)

Issue #6172, opened 2019-08-21
incorrect calculation of free storage after client restart


ygersie:
### Version
I've tested and reproduced on 2 older versions of Nomad (0.5.6 and 0.7.1) and this seems to be still an issue in 0.9.4 as well.

### Issue
Nomad can not allocate resources under the false presumption that we ran out of disk resources.

### Reproduction steps
Start a Nomad agent. Compare the Allocated Resources section of a `nomad node-status -self` with `df -h` reported available space. Now try to schedule a job with an ephemeral disk size similar to what is available to Nomad. This should succeed. Now fill up the disk using some random data: `dd if=/dev/zero of=/root/test.img bs=1M count=4096`. Try a `nomad plan` again, this should still succeed. Now restart the Nomad agent and try a `nomad plan` again, it should now fail stating it ran out of disk resources. 

The behaviour w.r.t. disk resources is different than for example CPU and Memory:
- Nomad allocated resources are reporting disk "free" fingerprinted during agent startup, instead of "total" disk space available.
- Nomad seems to deduct "allocated" GiB's from fingerprint reported "free" GiB's. This is not correct as the reported "free" GiB's changes over time when you restart the Nomad agent and have running jobs occupying disk space.

A real world example which clearly shows we should have plenty of disk resources available on this node but due the deduction: (free (566 GiB) - allocated (440 GiB)) we ran out:

```sh
ygersie@worker059:~$ nomad node-status -self
<snip>...
Allocated Resources
CPU              Memory         Disk             IOPS
12000/26388 MHz  24 GiB/62 GiB  440 GiB/566 GiB  0/0

Allocation Resource Utilization
CPU             Memory
1427/26388 MHz  20 GiB/62 GiB

Host Resource Utilization
CPU             Memory         Disk
1970/26388 MHz  28 GiB/63 GiB  354 GiB/984 GiB
<snip>....
```
```sh
ygersie@worker059:~$ df -h /
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda1       985G  355G  591G  38% /
```
and here an example job which asks to schedule Redis on `worker059`:
```sh
ygersie@nomad001:~$ cat example.nomad
job "example" {
  region = "ams1"
  datacenters = ["zone2"]
  type = "service"
  constraint {
    attribute = "${node.unique.name}"
    value     = "worker059"
  }

  group "cache" {
    count = 1
    restart {
      attempts = 10
      interval = "5m"
      delay = "25s"
[...]

PR #15852, opened 2023-01-23
Add disk_total_mb and disk_free_mb client config options

kjschubert:
@tgross as suggested in https://github.com/hashicorp/nomad/issues/6172#issuecomment-1377479562 and in regard to your positive reply I implemented the config options `disk_total_mb` and `disk_free_mb`. This is my first PR to nomad and none of the checklists in the contributing folder seems to fit those changes. Therefore I'm not sure if something is missing or how to proceed. I'm looking forward to your feedback.

### DIFF
--- a/.changelog/15852.txt
+++ b/.changelog/15852.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+client/fingerprint/storage: Added config options disk_total_mb and disk_free_mb to override detected disk space
+```
--- a/client/config/config.go
+++ b/client/config/config.go
@@ -113,6 +113,14 @@ type Config struct {
 	// determined dynamically.
 	MemoryMB int
 
+	// DiskTotalMB is the default node total disk space in megabytes if it cannot be
+	// determined dynamically.
+	DiskTotalMB int
+
+	// DiskFreeMB is the default node free disk space in megabytes if it cannot be
+	// determined dynamically.
+	DiskFreeMB int
+
 	// MaxKillTimeout allows capping the user-specifiable KillTimeout. If the
 	// task's KillTimeout is greater than the MaxKillTimeout, MaxKillTimeout is
 	// used.
--- a/client/fingerprint/storage.go
+++ b/client/fingerprint/storage.go
@@ -42,10 +42,10 @@ func (f *StorageFingerprint) Fingerprint(req *FingerprintRequest, resp *Fingerpr
 	}
 
 	if cfg.DiskTotalMB > 0 {
-		total = uint64(cfg.DiskTotalMB) * 1024 * 1024
+		total = uint64(cfg.DiskTotalMB) * bytesPerMegabyte
 	}
 	if cfg.DiskFreeMB > 0 {
-		free = uint64(cfg.DiskFreeMB) * 1024 * 1024
+		free = uint64(cfg.DiskFreeMB) * bytesPerMegabyte
 	}
 
 	if total < free {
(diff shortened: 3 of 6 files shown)


tgross (changes requested):
Hi @kjschubert! Thanks for opening this PR! Can you run `make cl` to add a changelog entry as well? We should get that sort of thing added to the contributing guide.

I pulled this patch down locally for testing and it mostly looks good. The attributes are picked up correctly and we can schedule only the amount expected. The `disk_free_mb` shows up under the `node status` Allocated Resources section as I'd expect.

But it turns out the `disk_total_mb` do doesn't actually much of anything here. Weirdly, that field isn't actually used by Nomad's scheduling at all except for fingerprinting the `unique.storage.bytestotal` attribute (which I guess someone could be using for job constraints?). The only place a total storage is surfaced is in the `node status`'s Host Resource Utilization, which isn't in this code path at all. That value gets reported with the expensively-gathered values from the client stats endpoint.

As such, we can keep the `disk_total_mb` but we'll need to extend the docs a bit to make sure it's clear what the purpose is. I've left suggestions on the PR for improving the doc.

tgross:
**Testing Notes**

If I set the following in my configuration:

```hcl
client {
  disk_total_mb = 5000
  disk_free_mb = 500
}
```

Then my node reports the correct available space under Allocated Resources and under the fingerprinted Attributes section:

```
$ nomad node status -self -verbose
...
Allocated Resources
CPU           Memory      Disk
0/100000 MHz  0 B/62 GiB  0 B/500 MiB

Allocation Resource Utilization
CPU           Memory
0/100000 MHz  0 B/62 GiB

Host Resource Utilization
CPU             Memory         Disk
653/100000 MHz  15 GiB/62 GiB  10 GiB/868 GiB

Allocations
No allocations placed

Attributes
...
unique.storage.bytesfree            = 524288000
unique.storage.bytestotal           = 5242880000
```

If I run an allocation with the default 300 MiB ephemeral disk, this gets counted properly against it:

```
Allocated Resources
CPU             Memory          Disk
128/100000 MHz  128 MiB/62 GiB  300 MiB/500 MiB
```

And if I bump the ephemeral disk up to 501 MiB or run a second allocation of the same size, I get a scheduler failure just as I'd expect:

```
 nomad job plan ./jobs/httpd.nomad
+ Job: "httpd"
+ Task Group: "web" (1 create)
  + Task: "http" (forces create)

Scheduler dry-run:
  Task Group "web" (failed to place 1 allocation):
    * Dimension "disk" exhausted on 1 nodes
```

But the host resource utilization looks all wrong:

```
Host Resource Utilization
CPU             Memory         Disk
653/100000 MHz  15 GiB/62 GiB  10 GiB/868 GiB
```

I followed through the CLI code and found that we get this value via two API calls. The first collects all the node attributes, among which is the `unique.storage.volume` key. Then it looks at the client stats API for that key and gets the used/total values from the device it finds with that key. The HTTP API equivalent calls are as follows:

```
$ nomad operator api "/v1/node/177f326c-b44c-61d5-68b9-9e74c8d5e5cd" | jq .Attributes | grep volume
  "unique.storage.volume": "rpool/ROOT/ubuntu_6blepg",

$ nomad operator api "/v1/client/stats?node_id=177f326c-b44c-61d5-68b9-9e74c8d5e5cd" | jq '.DiskStats[] | select(.Device == "rpool/ROOT/ubuntu_6blepg")'
{
  "Available": 920474746880,
  "Device": "rpool/ROOT/ubuntu_6blepg",
  "InodesUsedPercent": 0.022815878313334436,
  "Mountpoint": "/",
  "Size": 931683106816,
  "Used": 11208359936,
  "UsedPercent": 1.2030227717989055
}
```

Now admittedly I've got a weird setup on my machine with ZFS so the "Used" number is just plain wrong for me (and that's a separate bug). But the "Size" value is accurate for that volume. It looks like all of this gets done in [`client/stats/host.go`](https://github.com/hashicorp/nomad/blob/main/client/stats/host.go).

tgross (approved):
LGTM! Nice work, @kjschubert and thanks for contributing!

PR #15855, opened 2023-01-23
client: always run alloc cleanup hooks on final update

### LINKED ISSUE #15477: Health check routine leaks using new nomad provider
### Nomad version

### Issue
I have a service that logs HTTP requests and noticed that the endpoint given for health checking is being executed a few hundred times per second. There is a pretty aggressive restart policy on this job and we had a netsplit issue last night which lead to the service restarting around 600 times, so the logs are quite busy to say the least.

### Reproduction steps
Run the job below and either stop the job and resubmit or have the process crash. The number of requests hitting the service increase until nomad is restarted.

### Job file (if appropriate)

```
job "signaling" {
  datacenters = ["cloud"]
  type        = "service"

  reschedule {
    unlimited      = true
    delay          = "15s"
    delay_function = "constant"
    attempts       = 0
  }

  group "signaling" {
    restart {
      attempts = 2
      delay    = "1s"
      interval = "15s"
      mode     = "fail"
    }

    volume "opt" {
      type      = "host"
      source    = "opt"
      read_only = true
    }

    network {
      mode = "host"
      port "api" {
        static = 8000
      }
    }

    task "signaling" {
      driver = "exec"

      config {
        command = "entry.bash"
      }

      template {
        data = <<EOH
#!/bin/bash

/local/opt/signaling -bind=:8000
EOH

        destination = "local/entry.bash"
        perms       = "755"
      }

      template {
        data        = <<EOH
{{ with nomadVar "proapps" -}}
PROAPPS_ID={{ .id }}
PROAPPS_SECRET={{ .secret }}
{{- end }}
EOH
        destination = "local/file.env"
        env         = true
      }

      volume_mount {
        volume      = "opt"
        destination = "/local/opt"
      }

      resources {
        cpu    = 100 # Mhz
        memory = 512 # Mb
      }

      service {
        name     = "api"
        port     = "api"
        provider = "nomad"
        check {
          type     = "http"
          path     = "/debug/pprof/"
          interval = "5s"
          timeout  = "1s"
          method   = "GET"
        }
      }
    }
  }
}
```

shoenig:
This PR fixes a bug where alloc pre-kill hooks were not run in the
edge case where there are no live tasks remaining, but it is also
the final update to process for the (terminal) allocation. We need
to run cleanup hooks here, otherwise they will not run until the
allocation gets garbage collected (i.e. via Destroy()), possibly
at a distant time in the future.

Fixes #15477

### DIFF
--- a/client/allocrunner/alloc_runner.go
+++ b/client/allocrunner/alloc_runner.go
@@ -590,9 +590,9 @@ func (ar *allocRunner) handleTaskStateUpdates() {
 		} else {
 			// there are no live runners left
 
-			// run AR pre-kill hooks if this alloc is terminal; any post-stop
-			// tasks would regularly run in this state anyway (?)
-			if done {
+			// run AR pre-kill hooks if this alloc is done, but not if it's because
+			// the agent is shutting down.
+			if !ar.isShuttingDown() && done {
 				ar.preKillHooks()
 			}
 
--- a/client/allocrunner/alloc_runner.go
+++ b/client/allocrunner/alloc_runner.go
@@ -125,7 +125,7 @@ type allocRunner struct {
 	allocDir *allocdir.AllocDir
 
 	// runnerHooks are alloc runner lifecycle hooks that should be run on state
-	// transistions.
+	// transitions.
 	runnerHooks []interfaces.RunnerHook
 
 	// hookState is the output of allocrunner hooks
@@ -546,7 +546,9 @@ func (ar *allocRunner) handleTaskStateUpdates() {
 			}
 		}
 
+		// kill remaining live tasks
 		if len(liveRunners) > 0 {
+
 			// if all live runners are sidecars - kill alloc
 			onlySidecarsRemaining := hasSidecars && !hasNonSidecarTasks(liveRunners)
 			if killEvent == nil && onlySidecarsRemaining {
@@ -586,6 +588,14 @@ func (ar *allocRunner) handleTaskStateUpdates() {
 				}
 			}
 		} else {
+			// there are no live runners left
+
+			// run AR pre-kill hooks if this alloc is terminal; any post-stop
+			// tasks would regularly run in this state anyway (?)
+			if done {
+				ar.preKillHooks()
+			}
+
 			// If there are no live runners left kill all non-poststop task
 			// runners to unblock them from the alloc restart loop.
 			for _, tr := range ar.tasks {
--- a/.changelog/15477.txt
+++ b/.changelog/15477.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+client: Fixed a bug where allocation cleanup hooks would not run
+```
--- a/client/allocrunner/alloc_runner_hooks.go
+++ b/client/allocrunner/alloc_runner_hooks.go
@@ -329,6 +329,7 @@ func (ar *allocRunner) destroy() error {
 func (ar *allocRunner) preKillHooks() {
 	for _, hook := range ar.runnerHooks {
 		pre, ok := hook.(interfaces.RunnerPreKillHook)
+
 		if !ok {
 			continue
 		}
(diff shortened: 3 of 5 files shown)


client/allocrunner/alloc_runner.go
```diff
@@ -586,6 +588,14 @@ func (ar *allocRunner) handleTaskStateUpdates() {
 				}
 			}
 		} else {
+			// there are no live runners left
+
+			// run AR pre-kill hooks if this alloc is terminal; any post-stop
+			// tasks would regularly run in this state anyway (?)
+			if done {
+				ar.preKillHooks()
+			}
```
schmichael, line 597: I think this will erroneously run prekill hooks one client agent shutdown.

shoenig: I added a guard with `ar.isShuttingDown()`

gulducat: if `preKillHooks()` shouldn't be run during agent shutdown, would it make sense to put the guard inside of `preKillHooks()` instead of here?  more broadly, if this func is a potential foot-shoot, is there a way to make it less foot-shoot-able?

schmichael: Good question @gulducat. I'm afraid it needs to be up to the caller to decide whether prekill hooks should run or not because only the callers knows whether a kill event (eg a task dying) was triggered or not. Once your in prekill, a shutdown may be issued concurrently, but that doesn't change the fact that the prekill was *triggered by a kill event.* So I think this is the right approach as @shoenig's code is able to differentiate a kill event from an agent shutdown event.

That being said every operation in Nomad needs to be crash safe which is why prekill hooks are also run when garbage collecting dead allocations. So prekill should always eventually be called on every terminal allocation regardless of the triggering event.

gulducat (approved):
💀 ✅

client/allocrunner/alloc_runner.go
lgfa29: I think I understand things now 😅 

`preKillHooks` is only called by [`killTasks()`](https://github.com/hashicorp/nomad/blob/v1.4.3/client/allocrunner/alloc_runner.go#L646-L654) and `killTasks` is called in a few places:
- [`destroyImpl`](https://github.com/hashicorp/nomad/blob/v1.4.3/client/allocrunner/alloc_runner.go#L974) which is quite expected.
- [`handleTaskStateUpdates`](https://github.com/hashicorp/nomad/blob/v1.4.3/client/allocrunner/alloc_runner.go#L574-L575) which also makes sense since we may need to kill all tasks if one of the tasks fails.
- [`handleAllocUpdate`](https://github.com/hashicorp/nomad/blob/v1.4.3/client/allocrunner/alloc_runner.go#L963-L966) which again makes sense since the alloc is stopped we need to kill the tasks.

The part that was confusing me was why `handleAllocUpdate` was not triggering the `preKillHooks`: the allocation does eventually die and is marked with `ClientStatus: failed`. 

The reason is that this is a client-only alloc update. When watching allocations the client [ignores changes that it already knows about](https://github.com/hashicorp/nomad/blob/v1.4.3/client/client.go#L2188-L2194) and so the `ar.Update()` method is not called when the allocation fails client-side. 

We don't want to destroy the alloc runner so the only place left to call the hooks is in `handleTaskStateUpdates` like @shoenig did.

The second thing that was confusing to me is why we only [check for `killEvent` if there are live runners](https://github.com/hashicorp/nomad/blob/v1.4.3/client/allocrunner/alloc_runner.go#L556-L557), I feel like we should always be checking that and calling `killTasks()` if not nil? Something like https://github.com/hashicorp/nomad/compare/wip-luiz-kill-ar (this probably breaks other stuff since task lifecycle is finicky 😅).

All of this to say that I think we can use `if killEvent != nil` here to guard the `preKillHooks` so it matches the other conditional.

PR #15856, opened 2023-01-23
Initialized default jobs

philrenaud:
Stands up 4 Default job templates for users to select from when running a job via the UI

Establishes a new a utils const directory of job templates to be included. These are currently lightly-commented and I hope to improve commenting before this merges.

image
image

### DIFF
--- a/ui/app/controllers/jobs/run/templates/index.js
+++ b/ui/app/controllers/jobs/run/templates/index.js
@@ -8,10 +8,10 @@ export default class JobsRunTemplatesIndexController extends Controller {
   get templates() {
     return this.model.map((templateVariable) => {
       // THIS LOGIC SHOULD LIKELY MOVE TO THE SERIALIZATION LAYER
-      const description = templateVariable.items.description;
+      const description = templateVariable.keyValues.findBy('key', 'description')?.value;
       return {
         id: templateVariable.id,
-        label: templateVariable.path.split('nomad/job-templates/')[1],
+        label: templateVariable.id.split('nomad/job-templates/')[1].split('@')[0].replace(/-/g, ' '),
         description,
       };
     });
--- a/ui/app/controllers/jobs/run/templates/index.js
+++ b/ui/app/controllers/jobs/run/templates/index.js
@@ -8,11 +8,7 @@ export default class JobsRunTemplatesIndexController extends Controller {
 
   get templates() {
     return this.model.map((templateVariable) => {
-      // THIS LOGIC SHOULD LIKELY MOVE TO THE SERIALIZATION LAYER
-      const description = templateVariable.keyValues.findBy(
-        'key',
-        'description'
-      )?.value;
+      const description = templateVariable.items.description;
 
       // Removes the preceeding nomad/job-templates/default/, as well as the namespace, from the ID
       let label;
--- a/ui/app/controllers/jobs/run/templates/index.js
+++ b/ui/app/controllers/jobs/run/templates/index.js
@@ -7,27 +7,31 @@ export default class JobsRunTemplatesIndexController extends Controller {
   @tracked selectedTemplate = null;
 
   get templates() {
-    return this.model.map((templateVariable) => {
-      const description = templateVariable.items.description;
-
-      // Removes the preceeding nomad/job-templates/default/, as well as the namespace, from the ID
-      let label;
-      const delimiter = templateVariable.id.lastIndexOf('/');
-      if (delimiter !== -1) {
-        label = templateVariable.id.slice(delimiter + 1);
-      } else {
-        label = templateVariable.id;
-      }
-
-      label = capitalize(label.split('@')[0].replace(/-/g, ' '));
-
-      return {
-        id: templateVariable.id,
-        label,
-        description,
-      };
-    });
+    return this.model;
   }
+
+  // get templates() {
+  //   return this.model.map((templateVariable) => {
+  //     const description = templateVariable.items.description;
+
+  //     // Removes the preceeding nomad/job-templates/default/, as well as the namespace, from the ID
+  //     let label;
+  //     const delimiter = templateVariable.id.lastIndexOf('/');
+  //     if (delimiter !== -1) {
+  //       label = templateVariable.id.slice(delimiter + 1);
+  //     } else {
+  //       label = templateVariable.id;
+  //     }
+
+  //     label = capitalize(label.split('@')[0].replace(/-/g, ' '));
+
+  //     return {
+  //       id: templateVariable.id,
+  //       label,
+  //       description,
+  //     };
+  //   });
+  // }
 
   @action
   onChange(e) {
--- a/ui/app/controllers/jobs/run/templates/index.js
+++ b/ui/app/controllers/jobs/run/templates/index.js
@@ -5,6 +5,10 @@ import { tracked } from '@glimmer/tracking';
 export default class JobsRunTemplatesIndexController extends Controller {
   @tracked selectedTemplate = null;
 
+  get templates() {
+    return [...this.model.variables.toArray(), ...this.model.default];
+  }
+
   @action
   onChange(e) {
     this.selectedTemplate = e.target.id;
--- a/ui/app/controllers/jobs/run/templates/index.js
+++ b/ui/app/controllers/jobs/run/templates/index.js
@@ -1,6 +1,7 @@
 import Controller from '@ember/controller';
 import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
+import { capitalize } from '@ember/string';
 
 export default class JobsRunTemplatesIndexController extends Controller {
   @tracked selectedTemplate = null;
@@ -8,10 +9,25 @@ export default class JobsRunTemplatesIndexController extends Controller {
   get templates() {
     return this.model.map((templateVariable) => {
       // THIS LOGIC SHOULD LIKELY MOVE TO THE SERIALIZATION LAYER
-      const description = templateVariable.keyValues.findBy('key', 'description')?.value;
+      const description = templateVariable.keyValues.findBy(
+        'key',
+        'description'
+      )?.value;
+
+      // Removes the preceeding nomad/job-templates/default/, as well as the namespace, from the ID
+      let label;
+      const delimiter = templateVariable.id.lastIndexOf('/');
+      if (delimiter !== -1) {
+        label = templateVariable.id.slice(delimiter + 1);
+      } else {
+        label = templateVariable.id;
+      }
+
+      label = capitalize(label.split('@')[0].replace(/-/g, ' '));
+
       return {
         id: templateVariable.id,
-        label: templateVariable.id.split('nomad/job-templates/')[1].split('@')[0].replace(/-/g, ' '),
+        label,
         description,
       };
     });
--- a/ui/app/controllers/jobs/run/templates/index.js
+++ b/ui/app/controllers/jobs/run/templates/index.js
@@ -1,37 +1,9 @@
 import Controller from '@ember/controller';
 import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
-import { capitalize } from '@ember/string';
 
 export default class JobsRunTemplatesIndexController extends Controller {
   @tracked selectedTemplate = null;
-
-  get templates() {
-    return this.model;
-  }
-
-  // get templates() {
-  //   return this.model.map((templateVariable) => {
-  //     const description = templateVariable.items.description;
-
-  //     // Removes the preceeding nomad/job-templates/default/, as well as the namespace, from the ID
-  //     let label;
-  //     const delimiter = templateVariable.id.lastIndexOf('/');
-  //     if (delimiter !== -1) {
-  //       label = templateVariable.id.slice(delimiter + 1);
-  //     } else {
-  //       label = templateVariable.id;
-  //     }
-
-  //     label = capitalize(label.split('@')[0].replace(/-/g, ' '));
-
-  //     return {
-  //       id: templateVariable.id,
-  //       label,
-  //       description,
-  //     };
-  //   });
-  // }
 
   @action
   onChange(e) {
--- a/ui/app/controllers/jobs/run/templates/manage.js
+++ b/ui/app/controllers/jobs/run/templates/manage.js
@@ -2,6 +2,7 @@ import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
 import { tracked } from '@glimmer/tracking';
 import { task } from 'ember-concurrency';
+import { getOwner } from '@ember/application';
 
 export default class JobsRunTemplatesManageController extends Controller {
   @service flashMessages;
@@ -23,7 +24,7 @@ export default class JobsRunTemplatesManageController extends Controller {
   @task(function* (model) {
     try {
       yield model.destroyRecord();
-
+      getOwner(this).lookup('route:jobs.run.templates.manage').refresh();
       this.flashMessages.add({
         title: 'Job template deleted',
         message: `${model.path} successfully deleted`,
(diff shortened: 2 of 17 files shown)


ui/app/controllers/jobs/run/templates/index.js
```diff
@@ -1,17 +1,30 @@
...
 import { tracked } from '@glimmer/tracking';
+import { capitalize } from '@ember/string';
 
 export default class JobsRunTemplatesController extends Controller {
   @tracked selectedTemplate = null;
 
   get templates() {
     return this.model.map((templateVariable) => {
       // THIS LOGIC SHOULD LIKELY MOVE TO THE SERIALIZATION LAYER
-      const description = templateVariable.items.description;
```
philrenaud, line 11: Moving from .items to .keyValues here because the default templates are not Variables, but just Plain Old Javascript Objects.

ui/app/controllers/jobs/run/templates/index.js
```diff
@@ -1,17 +1,30 @@
...
+        label = templateVariable.id.slice(delimiter+1);
+      } else {
+        label = templateVariable.id
+      }
+
+      label = capitalize(label.split('@')[0].replace(/-/g, ' '));
+  
       return {
         id: templateVariable.id,
-        label: templateVariable.path.split('nomad/job-templates/')[1],
```
philrenaud, line 14: Moving from .path to .id here because the default templates are not Variables, but just Plain Old Javascript Objects.

ui/app/routes/jobs/run/index.js
```diff
@@ -28,18 +29,20 @@ export default class RunRoute extends Route {
...
 
+      let templateRecord;
       // If template is set in URL, create the job model and add the definition
       if (template) {
-        const templateRecord = await this.store.findRecord(
-          'variable',
-          template
-        );
-
+        if (DEFAULT_JOB_TEMPLATES.find((t) => t.id === template)) {
```
philrenaud, line 41: Don't make a network request if the job template passed in is among defaults

ui/app/routes/jobs/run/index.js
```diff
@@ -67,6 +70,17 @@ export default class RunRoute extends Route {
...
 
+  async establishDefaultJobTemplates() {
+    // Iterates over DEFAULT_JOB_TEMPLATES and creates a variable record for each of them, if they dont already exist
+    return await Promise.all(
+      DEFAULT_JOB_TEMPLATES.map((template) => {
+        if (!this.store.peekRecord('variable', template.id)) {
+          return this.store.createRecord('variable', template);
+        }
+      })
+    );
```
philrenaud, line 81: At this stage, scaffold all our default job templates as Variable objects, so they get access to all the goodness (.items, etc.) that the model provides.

ui/mirage/config.js
```diff
@@ -887,7 +891,12 @@ export default function () {
 
   //#region Variables
 
-  this.get('/vars', function (schema, { queryParams: { namespace } }) {
+  this.get('/vars', function (schema, { queryParams: { namespace, prefix } }) {
+    if (prefix === 'nomad/job-templates') {
+      return schema.variables
+        .all()
+        .filter((v) => v.path.includes('nomad/job-templates'));
+    }
```
ChaiWithJai, line 899: `question`: Do we have a branching issue, again? I think I already addressed this in a [previous PR](https://github.com/hashicorp/nomad/pull/15827/commits/d2ebb99c0b7a519fee2cd92c048de777eb13f756#).

ui/mirage/config.js
```diff
@@ -979,26 +988,37 @@ export default function () {
...
+      return new Response(
+        200,
+        {},
+        {
+          ACLToken: token.secretId,
+        }
+      );
+    },
+    { timing: 1000 }
+  );
```
ChaiWithJai, line 1021: `question`: Is this a prettier change? Happens to me a lot and I end up needing to use `git add -p` to patch add the changes.

ui/mirage/config.js
```diff
@@ -979,26 +988,37 @@ export default function () {
...
+    const { AuthMethod, ClientNonce, RedirectUri, Meta } = JSON.parse(
+      req.requestBody
+    );
+    return new Response(
+      200,
+      {},
+      {
+        AuthURL: `/ui/oidc-mock?auth_method=${AuthMethod}&client_nonce=${ClientNonce}&redirect_uri=${RedirectUri}&meta=${Meta}`,
+      }
+    );
```
ChaiWithJai, line 1000: `question`: Is this a prettier change? Happens to me a lot and I end up needing to use `git add -p` to patch add the changes.

ui/mirage/config.js
```diff
@@ -548,9 +548,14 @@ export default function () {
...
-      token.update({ policyIds: token.policyIds.filter(pid => pid !== id) });
-    });
+    schema.tokens
+      .all()
+      .models.filter((token) => token.policyIds.includes(id))
+      .forEach((token) => {
+        token.update({
+          policyIds: token.policyIds.filter((pid) => pid !== id),
+        });
+      });
```
ChaiWithJai, line 558: `question`: Is this a prettier change? Happens to me a lot and I end up needing to use `git add -p` to patch add the changes.

ui/mirage/config.js
```diff
@@ -443,7 +443,7 @@ export default function () {
     return JSON.stringify(findLeader(schema));
   });
 
-  this.get('/acl/tokens', function ({tokens}, req) {
+  this.get('/acl/tokens', function ({ tokens }, req) {
```
ChaiWithJai, line 446: `question`: Is this a prettier change? Happens to me a lot and I end up needing to use `git add -p` to patch add the changes.

ui/app/controllers/jobs/run/templates/index.js
ChaiWithJai: This would be resolved if we just moved the logic into the adapter layer and we could avoid this. It could lead to a bug down the line.

ui/app/controllers/jobs/run/templates/index.js
ChaiWithJai: ditto.

ChaiWithJai (commented):
I really think we should consider moving this logic into the Adapter layer. It would be less code and testable. I could see some changes stemming from `templates.index` being bugs down the line. But, I'm very happy to have a discussion on this.

Is there a reason why handling this in the adapter layer doesn't make sense for us?

ui/mirage/config.js
philrenaud: Yep, but that is not yet merged to epic/job-templates which this is based from. This duplicates that code and we should not have any merge conflicts.

ui/app/utils/default_jobs/variables.js
```diff
@@ -0,0 +1,49 @@
...
+      template {
+        data        = "<html>hello, {{ with nomadVar \"nomad/jobs/variables-example\" }}{{ .name }}{{ end }}</html>"
+        destination = "local/index.html"
+      }
+
+      resources {
+        cpu    = 128
+        memory = 128
+      }
+
```
mikenomitch, line 46: nit: I'd run these thru HCL format

ui/app/utils/default_jobs/variables.js
```diff
@@ -0,0 +1,49 @@
...
+job "variables-example" {
+  datacenters =  ["dc1"]
+
+  group "web" {
+
+    network {
+      # Task group will have an isolated network namespace with
+      # an interface that is bridged with the host
+      mode = "bridge"
```
mikenomitch, line 13: This relies on CNI plugins being installed - is there a good way to do this without bridge?

ui/app/utils/default_jobs/service-discovery.js
```diff
@@ -0,0 +1,41 @@
+export default `job "service-discovery-example" {
+  datacenters = ["dc1"]
+
+  group "cache" {
```
mikenomitch, line 4: IMO this one should have two groups with a task each, and they'll talk to eachother.

Maybe the "outer" task just curls with exec and the inner is a busybox with some version hello world (like below)?

ui/app/utils/default_jobs/hello-world.js
```diff
@@ -0,0 +1,38 @@
...
+      }
+    }
+
+    # Tasks are individual units of work that are run by Nomad.
+    task "redis" {
+      # This particular task starts a redis server within a Docker container
+      driver = "docker"
+
+      config {
+        image          = "redis:7"
```
mikenomitch, line 25: Note: we should have a non-redis example

ui/app/utils/default-job-templates.js
```diff
@@ -0,0 +1,61 @@
...
+    keyValues: [
+      {
+        key: 'template',
+        value: helloWorld,
+      },
+      {
+        key: 'description',
+        value: 'A simple job that runs a single task on a single node.',
+      },
+    ],
```
ChaiWithJai, line 18: `blocking`: We should get the schema for the `keyValues` right here because these default job templates will not be able to be deleted.

ui/app/utils/default_jobs/hello-world.js
```diff
@@ -0,0 +1,38 @@
+export default `job "hello-world" {
+  # Specifies the datacenters within which this job should be run.
+  # Leave as "dc1" for the default datacenter.
+  datacenters = ["dc1"]
```
mikenomitch, line 4: Ugh, I just realized the datacenters will probably mess up every job

Maybe we can get datacenter globbing into 1.5 for this. I was on the fence on trying to get this one in, but it'd make this feature much nicer.

ui/app/utils/default-job-templates.js
philrenaud: This is our standard variable keyValue typedef; https://github.com/hashicorp/nomad/blob/bd82a512c5a334c06012aa4dfa6a7bddbfd22451/ui/app/models/variable.js#L10-L15

ui/app/routes/jobs/run/index.js
```diff
@@ -33,29 +34,25 @@ export default class JobsRunIndexRoute extends Route {
...
-          template
-        );
-
+        const VariableAdapter = this.store.adapterFor('variable');
+        const templateRecord = await VariableAdapter.getJobTemplate(template);
         return this.store.createRecord('job', {
           _newDefinition: templateRecord.items.template,
         });
+      } else {
+        return this.store.createRecord('job');
```
ChaiWithJai, line 44: `nit` We don't need an `else` block here because we return out of the `if` statement.

ui/app/routes/jobs/run/templates/index.js
```diff
@@ -18,18 +19,9 @@ export default class JobsRunTemplatesIndexRoute extends Route {
...
-    await Promise.all(
-      jobTemplateVariables.map((template) =>
-        this.store.findRecord('variable', template.id)
-      )
-    );
-
-    return jobTemplateVariables;
+    const VariableAdapter = this.store.adapterFor('variable');
+    const jobTemplates = await VariableAdapter.getJobTemplates();
+    return jobTemplates;
```
ChaiWithJai, line 24: `nit` `code-golf`: you could just write

```
return VariableAdapter.getJobTemplates();
```

ChaiWithJai (changes requested):
Hey! The create flow is breaking as a result of this. Made some comments on my hunches as to why.

ui/app/utils/default_jobs/service-discovery.js
lgfa29: Yeah, it would be nice to have two groups. Here's an example:

```hcl
job "service-discovery-example" {
  datacenters = ["dc1"]

  group "client" {
    task "curl" {
      driver = "docker"

      config {
        image   = "curlimages/curl:7.87.0"
        command = "/bin/ash"
        args    = ["local/script.sh"]
      }

      template {
        data        = <<EOF
#!/usr/bin/env ash

while true; do
{{range nomadService "nomad-service-discovery-example-server"}}
  curl -L -v http://{{.Address}}:{{.Port}}/
{{end}}
  sleep 3
done
EOF
        destination = "local/script.sh"
      }

      resources {
        cpu    = 10
        memory = 50
      }
    }
  }

  group "server" {
    network {
      port "www" {
        to = 8001
      }
    }

    service {
      name     = "nomad-service-discovery-example-server"
      provider = "nomad"
      port     = "www"

      check {
        type     = "http"
        path     = "/"
        interval = "5s"
        timeout  = "1s"
      }
    }

    task "http" {
      driver = "docker"

      config {
        image   = "busybox:1"
        command = "httpd"
        args    = ["-v", "-f", "-p", "${NOMAD_PORT_www}", "-h", "/local"]
        ports   = ["www"]
      }

      template {
        data        = <<EOF
hello world
EOF
        destination = "local/index.html"
      }

      resources {
        cpu    = 10
        memory = 50
      }
    }
  }
}
```

Feel free to put something more meaningful instead of `hello world` 😅 

I also reduced the values in `resource` since these are very small containers.

ui/app/utils/default_jobs/parameterized.js
```diff
@@ -0,0 +1,37 @@
+export default `job "parameterized-job" {
+  datacenters = ["dc1"]
+  # Unlike service jobs, Batch jobs are intended to run until they exit successfully.
+  type = "batch"
```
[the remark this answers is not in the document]
tgross: Having this be parameterized also surfaces that workflow to new users, which is kind of nice.

ChaiWithJai (commented):
There's a failing test FYI.

Also I got a funky situation when overwriting a default variable. The API doesn't prevent me from using `/default/**`

image

ui/app/controllers/jobs/run/templates/manage.js
```diff
@@ -23,7 +24,7 @@ export default class JobsRunTemplatesManageController extends Controller {
   @task(function* (model) {
     try {
       yield model.destroyRecord();
-
+      getOwner(this).lookup('route:jobs.run.templates.manage').refresh();
```
ChaiWithJai, line 27: Funky! Guessing this is our reactivity workaround. Thankfully we have solid testing around this so we're covered on maintainability.

ChaiWithJai (approved):

ui/app/controllers/jobs/run/templates/manage.js
philrenaud: Ended up nixing this, but FWIW this is an established pattern from the Evaluations controller's refresh() method. Screams "global state editing" to me but ¯\_(ツ)_/¯

philrenaud:
@ChaiWithJai FWIW, I get the following errors when trying to save a variable at /default; you may just need to re-`make dev`.

image
image

ui/app/controllers/jobs/run/templates/manage.js
ChaiWithJai: Evaluations has a Refresh button in the UI because that view isn't real-time. We wanted to give users a fast way to clear all filters and refresh the page. We added the logic in when a user wanted to click a button to `refresh` the page per the designs. Evaluations was designed with reactivity in mind, however! That's why you see the statecharts! 😄 

```
  @action
  refresh() {
    const isDefaultParams = this.nextToken === null && this.status === null;
    if (isDefaultParams) {
      getOwner(this).lookup('route:evaluations.index').refresh();
      return;
    }

    this._resetTokens();
    this.status = null;
    this.pageSize = this.userSettings.pageSize;
  }
```

ui/app/utils/default_jobs/variables.js
philrenaud: Thanks! Noting here for posterity that the updated  example job spec will have `\\"nomad...` instead of `\"nomad...` for JS escape reasons.

HCL Format also did a couple funky things like added spaces between the `/` in my variable path in-template.

ui/app/utils/default_jobs/hello-world.js
philrenaud: Switched it over to busybox; thanks for the suggestions

PR #15860, opened 2023-01-24
sso: allow binding rules to create management ACL tokens.

jrasell:
This change allows binding rules to be created with a new type
named management. When this binding rule matches against a users
SSO claims, the resulting token will be of type management and not
client.

A match will supercede previous role or policy matches. This is
required as all binding rules are evaluated on login via
an auth method. It would be highly complex for users to ensure
those with rights to management tokens, do not match any other
binding rules.

When creating a binding rule of management type, the bind name
must be empty. This is account for within the validation logic, so
errors are clear for operators.

### DIFF
--- a/command/acl_binding_rule_create.go
+++ b/command/acl_binding_rule_create.go
@@ -53,11 +53,12 @@ Create Options:
 
   -bind-type
     Specifies adjusts how this binding rule is applied at login time to internal
-    Nomad objects. Valid options are "role" and "policy".
+    Nomad objects. Valid options are "role", "policy", or "management".
 
   -bind-name
     Specifies is the target of the binding used on selector match. This can be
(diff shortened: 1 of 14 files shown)


pkazmierczak (approved):
LGTM!

command/acl_binding_rule_create.go
```diff
@@ -53,7 +53,7 @@ Create Options:
 
   -bind-type
     Specifies adjusts how this binding rule is applied at login time to internal
-    Nomad objects. Valid options are "role" and "policy".
+    Nomad objects. Valid options are "role", "policy", or "management".
 
   -bind-name
```
lgfa29, line 58: Should the `-bind-name` help message and docs state that it must not be used if `-bind-type=management`?

lgfa29 (approved):

PR #15863, opened 2023-01-24
acl: Fix panic when bogus token is passed

tgross:
If a consumer of the new `Authenticate` method gets passed a bogus token that's a correctly-shaped UUID, it will correctly get an identity without a ACL token. But most consumers will then panic when they consume this nil `ACLToken` for authorization.

Because no API client should ever send a bogus auth token, update the `Authenticate` method to create the identity with remote IP (for metrics tracking) but also return an `ErrPermissionDenied`.

---

I hit this while adding the `Authenticate` to a broader set of APIs, in particular the CSI endpoint that has special authorization handling b/c it takes both nodes and tokens.


shoenig (approved):
LGTM

PR #15864, opened 2023-01-25
Task API via Unix Domain Socket

schmichael:
The goal of this PR is to provide universal workload access to Nomad's HTTP API via Unix Domain Sockets (UDS).

Dynamic Node Metadata #15844 is one potential use case.

Inspired by https://github.com/hashicorp/nomad/commit/e90807ec080c72025c434eb7ba7247e6adb84304

TODO
- [ ] docs - might do in a followup since dynamic node metadata / task api / workload id all need to interlink
- [ ] Unit tests for auth middleware
- [ ] Caching for auth middleware
- [ ] Rate limiting on negative lookups for auth middleware

### DIFF
--- a/client/allocrunner/taskrunner/api_hook.go
+++ b/client/allocrunner/taskrunner/api_hook.go
@@ -1,0 +1,83 @@
+package taskrunner
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"os"
+	"path/filepath"
+
+	log "github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/nomad/client/allocrunner/interfaces"
+	"github.com/hashicorp/nomad/client/config"
+)
+
+type apiHook struct {
+	srv    config.APIListenerRegistrar
+	logger log.Logger
+	ln     net.Listener
+}
+
+func newAPIHook(srv config.APIListenerRegistrar, logger log.Logger) *apiHook {
+	h := &apiHook{
+		srv: srv,
+	}
+	h.logger = logger.Named(h.Name())
+	return h
+}
+
+func (*apiHook) Name() string {
+	return "api"
+}
+
+func (h *apiHook) Prestart(ctx context.Context, req *interfaces.TaskPrestartRequest, resp *interfaces.TaskPrestartResponse) error {
+	if h.ln != nil {
+		//TODO(schmichael) remove me
+		h.logger.Trace("huh, called apiHook.Prestart twice")
+		return nil
+	}
+
+	//TODO(schmichael) what dir & perms
+	udsDir := filepath.Join(req.TaskDir.SecretsDir, "run")
+	if err := os.MkdirAll(udsDir, 0o775); err != nil {
+		return fmt.Errorf("error creating api socket directory: %w", err)
+	}
+
+	//TODO(schmichael) what name
+	udsPath := filepath.Join(udsDir, "nomad.socket")
+	if err := os.RemoveAll(udsPath); err != nil {
+		return fmt.Errorf("could not remove existing api socket: %w", err)
+	}
+
+	udsln, err := net.Listen("unix", udsPath)
+	if err != nil {
+		return fmt.Errorf("could not create api socket: %w", err)
+	}
+
+	go func() {
+		if err := h.srv.Serve(ctx, udsln); err != nil {
+			//TODO(schmichael) probably ignore http.ErrServerClosed
+			h.logger.Warn("error serving api", "error", err)
+		}
+	}()
+
+	h.ln = udsln
+	return nil
+}
+
+func (h *apiHook) Stop(ctx context.Context, req *interfaces.TaskStopRequest, resp *interfaces.TaskStopResponse) error {
+	if h.ln == nil {
+		//TODO(schmichael) remove me
+		h.logger.Trace("huh, no listener")
+		return nil
+	}
+
+	if err := h.ln.Close(); err != nil {
+		if errors.Is(err, net.ErrClosed) {
+			return nil
+		}
+		h.logger.Trace("error closing task listener: %v", err)
+	}
+	return nil
+}
--- a/client/allocrunner/taskrunner/api_hook.go
+++ b/client/allocrunner/taskrunner/api_hook.go
@@ -46,6 +46,11 @@ func (*apiHook) Name() string {
 }
 
 func (h *apiHook) Prestart(_ context.Context, req *interfaces.TaskPrestartRequest, resp *interfaces.TaskPrestartResponse) error {
+	if h.ln != nil {
+		// Listener already set. Task is probably restarting.
+		return nil
+	}
+
 	udsPath := apiSocketPath(req.TaskDir)
 	if err := os.RemoveAll(udsPath); err != nil {
 		h.logger.Warn("error removing task api socket", "path", udsPath, "error", err)
--- a/client/allocrunner/taskrunner/api_hook.go
+++ b/client/allocrunner/taskrunner/api_hook.go
@@ -3,26 +3,39 @@ package taskrunner
 import (
 	"context"
 	"errors"
-	"fmt"
 	"net"
 	"net/http"
 	"os"
 	"path/filepath"
 
 	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/nomad/client/allocdir"
 	"github.com/hashicorp/nomad/client/allocrunner/interfaces"
 	"github.com/hashicorp/nomad/client/config"
 )
 
+// apiHook exposes the Task API. The Task API allows task's to access the Nomad
+// HTTP API without having to discover and connect to an agent's address.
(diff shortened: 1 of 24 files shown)


client/allocrunner/taskrunner/task_runner.go
```diff
@@ -426,7 +426,11 @@ func NewTaskRunner(config *Config) (*TaskRunner, error) {
 
 	// Use the client secret only as the initial value; the identity hook will
 	// update this with a workload identity if one is available
-	tr.setNomadToken(config.ClientConfig.Node.SecretID)
+	//TODO(schmichael) can we remove this entirely? seems like a huge
+	//security problem if we use accidentally use it instead of the
+	//workload identity
```
tgross, line 431: We probably should. But the original reason for this was because of client upgrades from pre-1.4.0. Suppose you've got a Nomad 1.3.x cluster that starts an allocation with a `template` block that reads from Nomad native service discovery. That alloc has no Workload Identity b/c that was shipped in 1.4.0 and we mint the WI on alloc creation. When you upgrade the client to 1.4.x, if we didn't fallback to the node secret then all those allocations would fail.

command/agent/agent.go
```diff
@@ -1002,9 +1006,15 @@ func (a *Agent) setupClient() error {
...
 	// load this into the client config, therefore this needs to happen before
 	// we call NewClient.
+	//TODO migrate to APIListenerRegistrar
 	a.builtinListener, a.builtinDialer = bufconndialer.New()
 	conf.TemplateDialer = a.builtinDialer
 
+	// Initialize builtin API server here for use in the client, but it won't
+	// accept connections until the HTTP servers are created.
+	a.builtinServer = newBuiltinAPI()
+	conf.APIListenerRegistrar = a.builtinServer
```
shoenig, line 1016: Is this necessary? The implemenation of `builtinAPI` is clever and unfamiliar, which is usually a bad thing.

schmichael: **tl;dr I would _love_ a better way of doing this.**

> Is this necessary?

Kind of? Basically giving the Client access to the Agent HTTP server creates a circular dependency that needs solving somehow:

1. [NewAgent creates Client](https://github.com/hashicorp/nomad/blob/v1.4.3/command/agent/agent.go#L148-L150)
2. [agent.Command creates HTTP servers](https://github.com/hashicorp/nomad/blob/v1.4.3/command/agent/command.go#L538-L545) which accept the Agent as a dependency

That worked out fine when the Client was oblivious to the existence of the HTTP API.

When we needed access to the HTTP API for `template` support we used [`bufconn` to create a Dialer connected to a Listener](https://github.com/hashicorp/nomad/blob/v1.4.3/command/agent/agent.go#L976-L981), returned the Listener to Agent, and [Agent would slap the HTTP API on it](https://github.com/hashicorp/nomad/blob/v1.4.3/command/agent/http.go#L173-L186). `bufconn` [blocks Dials until the Listener accepts](https://github.com/grpc/grpc-go/blob/v1.52.3/test/bufconn/bufconn.go#L93-L106) them.

Unix Domain Sockets (UDS) are _listeners_ themselves though, so we couldn't just reuse the Dialer (unless we wanted to create 1 Accept goroutine per Task and 2 proxy goroutines for every UDS connection to proxy reads and writes from the UDS to the `bufconn`).

So I more or less copied `bufconn` except instead of blocking a Dial attempt on its corresponding Accept, I block a `Register(listener)` attempt on the HTTP Server existing.

> schmichael why don't you just swap the order we start the api and client?

I think that could work. The `Agent.{Client,Server}` getters for the API to access the client and server would need locks and setters since they would be nil until NewAgent returned them. The dependencies that the API has on Agent itself are harder to untangle: `Agent.RPC(` being one of the harder bits. Do we do a `bufconn`/`builtinServer` style `block RPCs from the HTTP API until the Agent is created`? We don't really gain much from swapping the order if we don't also detangle the circular dependencies. So it gets very complicated especially because both Client and Server have their own RPC implementations and methods of routing requests around and the HTTP API just grabs one and uses it.

client/allocrunner/taskrunner/api_hook.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"os"
+	"path/filepath"
+
+	log "github.com/hashicorp/go-hclog"
```
shoenig, line 11: nitpick: I've been trying to cleanup ineffective package aliases like this

client/allocrunner/taskrunner/task_runner.go
schmichael: k, reverting for now and we can followup in a separate pr

schmichael:
Marked as ready for review despite some outstanding TODOs. All of them are fairly orthogonal to this and may be best suited to followup PRs anyway.

shoenig:
@schmichael can you look into these test failures?

```
=== RUN   TestAllocRunner_Restore_RunningTerminal
    alloc_runner_unix_test.go:130: 
        	Error Trace:	/home/runner/work/nomad/nomad/client/allocrunner/alloc_runner_unix_test.go:130
        	Error:      	"[{remove 16566603-2d92-a388-a6ac-becf18a65b32 web 2023-02-04 01:29:21.8589[730](https://github.com/hashicorp/nomad/actions/runs/4089214192/jobs/7051705048#step:4:731)08 +0000 UTC m=+0.094414412} {remove 16566603-2d92-a388-a6ac-becf18a65b32 group-web 2023-02-04 01:29:22.860232985 +0000 UTC m=+1.095674389} {remove 16566603-2d92-a388-a6ac-becf18a65b32 group-web 2023-02-04 01:29:22.860298886 +0000 UTC m=+1.095740290}]" should have 2 item(s), but has 3
```

```
# github.com/hashicorp/nomad/command/agent/consul_test [github.com/hashicorp/nomad/command/agent/consul.test]
Error: command/agent/consul/int_test.go:168:3: unknown field APIListenerRegistrar in struct literal of type taskrunner.Config
```

client/allocrunner/taskrunner/api_hook.go
```diff
@@ -0,0 +1,113 @@
...
+		h.logger.Warn("error removing task api socket", "path", udsPath, "error", err)
+	}
+
+	udsln, err := net.Listen("unix", udsPath)
+	if err != nil {
+		h.logger.Warn("error listening on task api socket", "path", udsPath, "error", err)
+		return nil
+	}
+
+	if err := os.Chmod(udsPath, 0o777); err != nil {
```
tgross, line 65: I think our logic here for why this is safe in the context of the task dir is sound, but it's also going to result in us getting dinged by third-party auditors armed with sophisticated source code analysis tools like `grep` :grinning: 

Can we leave a comment above here about why it's ok in hopes of reducing the amount of noise we get from those discussions?

client/allocrunner/taskrunner/api_hook.go
```diff
@@ -0,0 +1,113 @@
...
+	h.logger = logger.Named(h.Name())
+	return h
+}
+
+func (*apiHook) Name() string {
+	return "api"
+}
+
+func (h *apiHook) Prestart(_ context.Context, req *interfaces.TaskPrestartRequest, resp *interfaces.TaskPrestartResponse) error {
+	if h.ln != nil {
```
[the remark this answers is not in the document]
schmichael: Correct, but that seems needlessly fragile. I'll add locks to prevent easy mistakes in the future (like if we add Update or Exited hooks).

client/allocrunner/taskrunner/api_hook.go
schmichael: Good points. I was going to say `well we require auth anyway so who cares` but...

These are the sorts of decisions we have to live with forever, and respecting `Task.User` when its set (as `nomad_token` does too) seems like a valuable road to go down.

So I followed #15755 and made this attempt to chown the socket to `Task.User` with `0o600` perms. Like #15755 it falls back to `0o666` if that fails (which it always will on Windows or as a non-root user).

e2e/workload_id/input/api-auth.nomad.hcl
```diff
@@ -0,0 +1,99 @@
+job "api-auth" {
```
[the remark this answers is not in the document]
schmichael: Oops, sorry for the messy PR. Fixed.

tgross (approved):
LGTM

apollo13:
Hi there, I realize I am kinda late to the party here -- sorry for that. I was wondering if it is a good idea to expose the API sockets to tasks by default. I realize that the access to it requires authentication, but not all tasks are trusted and a pre-auth vulnerability in the API would allow "malicious" tasks to exploit this. From an security POV I'd very much prefer this to be opt in (it would keep the attack surface smaller).

schmichael:
That's a valid concern @apollo13! Mind opening a new issue? We discussed the ability to opt out internally but couldn't decide who should control it: Client Agent, jobspec, namespace, ...some combination of those to allow defaults in one place to be overridden in another place?

PR #15867, opened 2023-01-25
ui: add client links to topo viz

### LINKED ISSUE #15843: Add client node links from topology page
### Proposal

When viewing the topology page, it might be nice to jump directly to a client node's show page. We should swap out the plain text title of each client node with a link.

### Use-cases

Quickly, figuring out how a client is configured (ie looking at metadata) or viewing all allocations on it more easily.

ChaiWithJai:
Closes #15843 

image

### DIFF
--- a/ui/app/templates/components/topo-viz/node.hbs
+++ b/ui/app/templates/components/topo-viz/node.hbs
@@ -6,7 +6,7 @@
       {{else if (not @node.node.isEligible)}}
         <span data-test-status-icon class="tooltip" aria-label="Client is ineligible">{{x-icon "lock-closed" class="is-warning"}}</span>
       {{/if}}
-      <strong>{{@node.node.name}}</strong>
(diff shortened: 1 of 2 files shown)


ui/app/templates/components/topo-viz/node.hbs
```diff
@@ -6,7 +6,7 @@
       {{else if (not @node.node.isEligible)}}
         <span data-test-status-icon class="tooltip" aria-label="Client is ineligible">{{x-icon "lock-closed" class="is-warning"}}</span>
       {{/if}}
-      <strong>{{@node.node.name}}</strong>
+      <strong><LinkTo @route="clients.client" @model={{@node.node.id}}>{{@node.node.name}}</LinkTo></strong>
```
ChaiWithJai, line 9: This causes a break in our accessibility testing, however. This might be an issue where our test suite doesn't have a property of `node.node.id` so I'll need to patch the test.

philrenaud (approved):

PR #15869, opened 2023-01-25
acl: make auth method default across all types

pkazmierczak:
As discussed with the team, we want to have `default` auth method across all types, for better UX. 

Relates to https://github.com/hashicorp/nomad/issues/13120

### DIFF
--- a/.changelog/15869.txt
+++ b/.changelog/15869.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+acl: made auth method default across all method types
+```
--- a/.changelog/15869.txt
+++ b/.changelog/15869.txt
@@ -1,3 +1,0 @@
-```release-note:improvement
-acl: made auth method default across all method types
-```
--- a/command/login.go
+++ b/command/login.go
@@ -143,6 +143,13 @@ func (l *LoginCommand) Run(args []string) int {
 				if l.authMethodType == "" {
 					l.authMethodType = authMethod.Type
 				}
+				if l.authMethodType != authMethod.Type {
+					l.Ui.Error(fmt.Sprintf(
+						"Specified type: %s does not match the type of the default method: %s",
+						l.authMethodType, authMethod.Type,
+					))
+					return 1
+				}
 			}
 		}
 
--- a/command/login.go
+++ b/command/login.go
@@ -48,7 +48,8 @@ Login Options:
     has configured a default, this flag is optional.
 
   -type
-    Type of the auth method to login to. Defaults to "OIDC".
+	Type of the auth method to login to. If the cluster administrator has
+	configured a default, this flag is optional.
 
   -oidc-callback-addr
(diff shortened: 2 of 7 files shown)


command/login.go
```diff
@@ -141,11 +140,14 @@ func (l *LoginCommand) Run(args []string) int {
 		for _, authMethod := range authMethodList {
 			if authMethod.Default {
 				l.authMethodName = authMethod.Name
+				if l.authMethodType == "" {
+					l.authMethodType = authMethod.Type
+				}
```
jrasell, line 145: What happens if the user specifies the auth method type, but this does not match the default one found? I assume the login will fail further down the flow, and wonder if we either want to error on this, or overwrite the passed flag value?

nomad/state/state_store_acl_sso.go
```diff
@@ -186,10 +185,10 @@ func (s *StateStore) GetACLAuthMethodByName(ws memdb.WatchSet, authMethod string
 	return nil, nil
 }
 
-// GetDefaultACLAuthMethodByType returns a default ACL Auth Methods for a given
+// GetDefaultACLAuthMethod returns a default ACL Auth Methods for a given
 // auth type. Since we only want 1 default auth method per type, this function
 // is used during upserts to facilitate that check.
```
jrasell, line 190: I think we should update the function comment as I don't believe this is now correct based on the change of the logic.

jrasell (approved):
Looking good, thanks! I have noted a couple of minor things I think are worth changing and a question I am curious of your thoughts on.

command/login.go
pkazmierczak: Right, it crossed my mind. I think it's fine to let it fail down the line, I don't think we should be overwriting things. Are you ok with this? or do you want to make a check here and fail quickly?

nomad/state/state_store_acl_sso.go
pkazmierczak: resolved in [a56579e](https://github.com/hashicorp/nomad/pull/15869/commits/a56579ec97e83cd84f20ab0cdbbf93374a0af948)

.changelog/15869.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+acl: made auth method default across all method types
+```
```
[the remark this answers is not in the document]
pkazmierczak: resolved in [a56579e](https://github.com/hashicorp/nomad/pull/15869/commits/a56579ec97e83cd84f20ab0cdbbf93374a0af948)

command/login.go
pkazmierczak: ok maybe it's best to fail quickly. 
[3e7dab8](https://github.com/hashicorp/nomad/pull/15869/commits/3e7dab826d82bb438deab46137b52b45059a2740)

PR #15870, opened 2023-01-25
WI: allow workloads to use RPCs associated with HTTP API

tgross:
This changeset allows Workload Identities to authenticate to all the RPCs that
support HTTP API endpoints, for use with PR #15864.

* Extends the work done for pre-forwarding authentication to all RPCs that
  support a HTTP API endpoint.
* Consolidates the auth helpers used by the CSI, Service Registration, and Node
  endpoints that are currently used to support both tokens and client secrets.

Intentionally excluded from this changeset:
* The Variables endpoint still has custom handling because of the implicit
  policies. Ideally we'll figure out an efficient way to resolve those into real
  policies and then we can get rid of that custom handling.
* The RPCs that don't currently support auth tokens (i.e. those that don't
  support HTTP endpoints) have not been updated with the new pre-forwarding auth
  We'll be doing this under a separate PR to support RPC rate metrics.

### DIFF
--- a/nomad/acl.go
+++ b/nomad/acl.go
@@ -156,7 +156,29 @@ func (s *Server) remoteIPFromRPCContext(ctx *RPCContext) (net.IP, error) {
 	return nil, structs.ErrPermissionDenied
 }
 
-func (s *Server) ResolveACL(aclToken *structs.ACLToken) (*acl.ACL, error) {
+func (s *Server) ResolveACL(args structs.RequestWithIdentity) (*acl.ACL, error) {
+	identity := args.GetIdentity()
+	if !s.config.ACLEnabled || identity == nil {
+		return nil, nil
+	}
+	aclToken := identity.GetACLToken()
+	if aclToken != nil {
+		return s.ResolveACLForToken(aclToken)
+	}
+	claims := identity.GetClaims()
+	if claims != nil {
+		return s.ResolveClaims(claims)
+	}
+	return nil, nil
+}
+
+// ResolveACLForToken resolves an ACL from a token only. It should be used only
+// by Variables endpoints, which have additional implicit policies for their
+// claims so we can't wrap them up in ResolveACL.
+//
+// TODO: figure out a way to the Variables endpoint implicit policies baked into
+// their acl.ACL object so that we can avoid using this method.
+func (s *Server) ResolveACLForToken(aclToken *structs.ACLToken) (*acl.ACL, error) {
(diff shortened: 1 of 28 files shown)


nomad/acl.go
```diff
@@ -156,7 +156,29 @@ func (s *Server) remoteIPFromRPCContext(ctx *RPCContext) (net.IP, error) {
 	return nil, structs.ErrPermissionDenied
 }
 
-func (s *Server) ResolveACL(aclToken *structs.ACLToken) (*acl.ACL, error) {
+func (s *Server) ResolveACL(args structs.RequestWithIdentity) (*acl.ACL, error) {
```
schmichael, line 159: Sorry it's not formatted properly and probably wordier than it needs to be, but I thought I'd at least leave a sample for you to grab.

```suggestion
// ResolveACL is an authentication wrapper which handles resolving both ACL tokens and Workload Identities. If both are provided the ACL token is preferred, but it is best for the RPC caller to only include the credentials for the identity they intend the operation to be performed with.
func (s *Server) ResolveACL(args structs.RequestWithIdentity) (*acl.ACL, error) {
```

schmichael (commented):
LGTM! Thanks for jumping on this. I'll give it a +1 once you move it out of draft.

tgross:
In addition to running the unit tests I've tested this out by spinning up a 3 server cluster along with a client, and verified that the job deployment workflow is running, that client APIs (ex. `alloc fs`) are working, that CSI plugins can register, and that the event stream works. Tested all those with no ACLs, with ACLs, and with mTLS.

nomad/node_endpoint.go
```diff
@@ -890,7 +898,7 @@ func (n *Node) GetNode(args *structs.NodeSpecificRequest,
 	if err != nil {
 		return err
 	}
-	if !aclObj.AllowNodeRead() {
+	if aclObj != nil && !aclObj.AllowNodeRead() {
```
schmichael, line 901: Huh, how did this sneak by? I would think it would panic on `nomad node status ...` for everyone with ACLs disabled?

schmichael (approved):

nomad/node_endpoint.go
tgross: Yeah that burned me here... the old `ResolveToken` code returns `nil, nil` in that case, so we just skip over the `AllowNodeRead` check. Fortunately we have good unit test coverage of non-ACLs vs with-ACLs and that caught it.

PR #15876, opened 2023-01-25
metrics: measure rate of RPC requests that serve API

tgross:
This changeset configures the RPC rate metrics that were added in #15515 to all the RPCs that support authenticated HTTP API requests. These endpoints already configured with pre-forwarding authentication in #15870, and a handful of others were done already as part of the proof-of-concept work. So this changeset is entirely copy-and-pasting one method call into a whole mess of handlers.

Upcoming PRs will wire up pre-forwarding auth and rate metrics for the remaining set of RPCs that have no API consumers or aren't authenticated, in smaller chunks that can be more thoughtfully reviewed. (And add the metrics reference docs update.)

### DIFF
--- a/nomad/acl_endpoint.go
+++ b/nomad/acl_endpoint.go
@@ -79,6 +79,7 @@ func (a *ACL) UpsertPolicies(args *structs.ACLPolicyUpsertRequest, reply *struct
 	if done, err := a.srv.forward("ACL.UpsertPolicies", args, args, reply); done {
 		return err
 	}
+	a.srv.MeasureRPCRate("acl", structs.RateMetricWrite, args)
 	if authErr != nil {
 		return structs.ErrPermissionDenied
 	}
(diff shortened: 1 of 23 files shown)


nomad/acl_endpoint.go
```diff
@@ -79,6 +79,7 @@ func (a *ACL) UpsertPolicies(args *structs.ACLPolicyUpsertRequest, reply *struct
 	if done, err := a.srv.forward("ACL.UpsertPolicies", args, args, reply); done {
 		return err
 	}
+	a.srv.MeasureRPCRate("acl", structs.RateMetricWrite, args)
```
lgfa29, line 82: Is there any meaningful difference between measuring the RPC rate before or after forwarding? The only thing I can think of is that requests that fail to be forwarded would not be counted, but that's probably OK 🤔

lgfa29 (approved):

nomad/acl_endpoint.go
lgfa29: Oh, measuring before forwarding would count the request twice (at least). Never mind 😬

tgross: By doing it after forwarding, we ensure that we only measure a RPC on the server where it's being "finally" processed. 

For example, a write gets measured on the leader, but not the follower (or other region!) it's been forwarded from. Whereas a stale query gets measured on the follower that serves the query. Counting for each hop would be confusing for administrators and would make eventual rate limiting really weird to handle.

PR #15892, opened 2023-01-26
cli: separate auth method config output for easier reading.

jrasell:
"easier reading" is subjective, so interested in overall feedback on how this looks and what could be improved.


pkazmierczak (approved):
I think this is good, it _does_ make the output more readable. Are we sure it doesn't break any formatting conventions? (I don't think our other commands output this way?)

Then again does it matter? If users want a particular way of output formatting, they can use json or template flags.

shoenig:
After some random clicking in the docs, I think the only tweak I'd suggest is removing the use of colons on the headers, e.g.

https://developer.hashicorp.com/nomad/docs/commands/status#examples
https://developer.hashicorp.com/nomad/docs/commands/acl/policy/info#examples

And perhaps something more descriptive than "Config"

shoenig (approved):

PR #15897, opened 2023-01-26
acl: JWT as SSO auth method

pkazmierczak:
This PR introduces a new ACL Auth Method: JWT, which allows users to exchange 3rd party JSON Web Tokens for Nomad ACL Tokens. We achieve this by:
- extending ACLAuthMethodConfig fields with JWT-specific properties
- a new RPC endpoint `ACL.Login`
- a new HTTP API endpoint `/v1/acl/login`
- adjustments to the CLI that allow for new `-type=JWT` when creating and updating auth methods, and to the `nomad login` command
- and adjustments to the sign in UI which now detects if there are JWT auth methods present, whether they are multiple and modifies the "sign in with token" field accordingly. 

All the changes mentioned above were described in [NMD-167](https://docs.google.com/document/d/1MqhdDpMqOYMT72PgdhxsGHmF0-WcK5425oxEBAkQRXo/edit), and all the commits in this pull request (except for changelog) were previously reviewed.

### DIFF
--- a/lib/auth/jwt/validator.go
+++ b/lib/auth/jwt/validator.go
@@ -1,0 +1,125 @@
+package jwt
+
+import (
+	"context"
+	"crypto"
+	"fmt"
+	"time"
+
+	"github.com/armon/go-metrics"
+	"github.com/hashicorp/cap/jwt"
+	"golang.org/x/exp/slices"
+
+	"github.com/hashicorp/nomad/helper"
+	"github.com/hashicorp/nomad/nomad/structs"
+)
+
+// Validate performs token signature verification and JWT header validation,
+// and returns a list of claims or an error in case any validation or signature
+// verification fails.
+func Validate(ctx context.Context, token string, methodConf *structs.ACLAuthMethodConfig) (map[string]any, error) {
+	var (
+		keySet jwt.KeySet
+		err    error
+	)
+
+	// JWT validation can happen in 3 ways:
+	// - via embedded public keys, locally
+	// - via JWKS
+	// - or via OIDC provider
+	if len(methodConf.JWTValidationPubKeys) != 0 {
+		keySet, err = usingStaticKeys(methodConf.JWTValidationPubKeys)
+		if err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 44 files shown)


lib/auth/jwt/validator.go
```diff
@@ -0,0 +1,125 @@
...
+		return nil, fmt.Errorf("unable to get validation keys from JWKS: %v", err)
+	}
+	return keySet, nil
+}
+
+func usingOIDC(ctx context.Context, oidcurl string, oidccapem []string) (jwt.KeySet, error) {
+	// Measure the OIDC endpoint performance.
+	defer metrics.MeasureSince([]string{"nomad", "acl", "jwt", "oidc_jwt"}, time.Now())
+
+	// TODO why do we have DiscoverCaPem as an array but JWKSCaPem as a single string?
```
tgross, line 114: Just a quick check that this TODO is ok to leave before merging?

nomad/acl_endpoint.go
```diff
@@ -2777,3 +2783,152 @@ func (a *ACL) OIDCCompleteAuth(
...
+	// respect stale queries.
+	if done, err := a.srv.forward(structs.ACLLoginRPCMethod, args, args, reply); done {
+		return err
+	}
+
+	// Measure the login endpoint performance.
+	defer metrics.MeasureSince([]string{"nomad", "acl", "login"}, time.Now())
+
+	// This endpoint can only be used once all servers in all federated regions
+	// have been upgraded to 1.5.2 or greater, since JWT Auth method was
```
tgross, line 2806: Should this read 1.5.3? Or should we just remove the specific version here and let that live in the `minACLJWTAuthMethodVersion` constant so we don't need to update it if we have to ship a security update or whatever before this goes out?

tgross (approved):
LGTM!

lib/auth/jwt/validator.go
pkazmierczak: Right, this is "a note for future us." Something we'll definitely revisit.

PR #15899, opened 2023-01-26
metrics: Add metrics to unauthenticated endpoints

tgross:
Continues the work done in #15876 by extending rate metrics measurement over the various unauthenticated RPCs.


shoenig (approved):
LGTM

Issue #15875, opened 2023-01-25
create ephemeral policies for task access to Variables


tgross:
[Workload-Associated ACL Policies](https://developer.hashicorp.com/nomad/docs/concepts/workload-identity#workload-associated-acl-policies) are ordinary ACL policies with the appropriate job/group/task fields set. This is suitable for use with most APIs, as shown in https://github.com/hashicorp/nomad/pull/15870. But the implicit [Task Access to Variables](https://developer.hashicorp.com/nomad/docs/concepts/variables#task-access-to-variables) pseudo-policies need special handling for the Variables RPC, which feels likely to result in a security bug eventually. 

The `acl.ACL` objects are expensive to create, so we cache them on the server after they've been resolved from the token (or claim). The cache is keyed by a hash of the policies and not the token, so without considering Task Access to Variables most `acl.ACL` objects we get from WIs are going to be the same and will cache efficiently. Unfortunately when you add Task Access to Variables implicit policies, every WI will have a different cache key and that'll explode the number of `acl.ACL` objects we need to cache if there are a lot of them. This will push user-tokens out of the LRU.

A couple of ideas we'll want to explore:
* Is there a way to create an efficient "pattern matching" policy that can be matched against the Variables paths without having to have a separate policy for each and every WI? (That's also https://github.com/hashicorp/nomad/issues/14764)
* Is the memory usage so much that we can't just split out the WI ACL objects to their own cache, so that we don't push user-tokens out of the LRU cache?

tgross:
Will be closed by #18744. That doesn't exactly do what's described here, but instead makes it unnecessary.

PR #15901, opened 2023-01-26
metrics: Add remaining server RPC rate metrics

tgross:
Continues the work done in #15876 by extending rate metrics measurement over a few stragglers that got missed in the original PR (probably because of concurrent PRs landing).

This changeset includes an update to the semgrep rule to add a rule for rate metrics, and another so that we're matching on only the currently used patterns. I've also left a some TODOs in the auth rule pointing to https://github.com/hashicorp/nomad/issues/15875, which isn't likely to ship in 1.5.0

### DIFF
--- a/.semgrep/rpc_endpoint.yml
+++ b/.semgrep/rpc_endpoint.yml
@@ -6,77 +6,83 @@ rules:
           if done, err := $A.$B.forward($METHOD, ...); done {
             return err
           }
+      # Pattern used by typical endpoints that take an auth token or workload
+      # identity. Some of these endpoints have no context for Authenticate
       - pattern-not-inside: |
+          authErr := $A.$B.Authenticate(...)
+          ...
           if done, err := $A.$B.forward($METHOD, ...); done {
             return err
           }
           ...
-          ... := $X.$Y.ResolveToken(...)
+          ... := $A.$B.ResolveACL(...)
           ...
+      # Pattern used by endpoints that are used by both ACLs and Clients.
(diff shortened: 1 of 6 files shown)


tgross:
Looks like I've got a failing set of tests from clients getting their tokens. Moving to draft till that's fixed.

tgross:
Tests fixed! There's a fussy little bit around whether errors get returned wrapped with `rpc error: %v` or not depending on where the error originate, and that broke some tests.

.semgrep/rpc_endpoint.yml
```diff
@@ -100,4 +107,4 @@ rules:
     severity: "WARNING"
     paths:
       include:
-        - "*_endpoint.go"
+        - "nomad/*_endpoint.go"
```
lgfa29, line 110: This will exclude client endpoints, do they follow a different pattern for authorization now?

lgfa29 (approved):
LGTM!

.semgrep/rpc_endpoint.yml
tgross: There's two kinds of "client endpoints": the server-side client RPCs are in files like `nomad/client_*_endpoint.go` and get picked up with this glob. The client-side client RPCs in files like `client/*_endpoint.go` have always been excluded, because they don't match the first pattern with the RPC forwarding:

```
    patterns:
      - pattern: |
          if done, err := $A.$B.forward($METHOD, ...); done {
            return err
          }
```

This change just makes sure we don't bother checking the files that will never have this pattern.

PR #15903, opened 2023-01-26
artifact: enable reading system git/mercurial configuration

### LINKED ISSUE #15498: artifact: enable reading system gitconfig

shoenig:
This PR adjusts the artifact sandbox on Linux to enable reading from known
system-wide git or mercurial configuration, if they exist.

Folks doing something odd like specifying custom paths for global config will
need to use the standard locations, or disable artifact filesystem isolation.

Closes #15498

### DIFF
--- a/client/allocrunner/taskrunner/getter/util_linux.go
+++ b/client/allocrunner/taskrunner/getter/util_linux.go
@@ -3,6 +3,7 @@
 package getter
 
 import (
+	"os"
 	"path/filepath"
 	"syscall"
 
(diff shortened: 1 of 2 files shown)


tgross (approved):
LGTM

PR #15904, opened 2023-01-26
bug: reactively update view

ChaiWithJai:
Bug associated with `O(n+1)` query causing us to lose reference to Ember Data Store because we're using the `map` method.

### DIFF
--- a/ui/app/routes/jobs/run/templates/index.js
+++ b/ui/app/routes/jobs/run/templates/index.js
@@ -22,11 +22,14 @@ export default class JobsRunTemplatesIndexRoute extends Route {
       prefix: 'nomad/job-templates',
       namespace: '*',
     });
-    const recordsToQuery = jobTemplateVariables.map((template) =>
-      this.store.findRecord('variable', template.id)
+
+    await Promise.all(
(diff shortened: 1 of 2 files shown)


philrenaud (approved):
I think this looks good, but I'm struggling to think of where this bug would surface. What would it take to demonstrate a non-reactive model return here?

ChaiWithJai:
> I think this looks good, but I'm struggling to think of where this bug would surface. What would it take to demonstrate a non-reactive model return here?

1. If the page goes real-time
2. If there's any editing. Like lets say we decide the `manage` view is one too many clicks and you should be able to delete from this page.

But ultimately, its ensuring the proper pattern is put into place because we will copy/paste styles around the codebase. So if we ever do an `O(n+1)` query elsewhere, we'll come to this page and probably do a copy pasta.

ChaiWithJai:
> What would it take to demonstrate a non-reactive model return here?

The model is reactive. However, ensuring what it's in reference to. We're just assuring that we're pointing to the Ember Data store and not a new array that we're creating in the model hook.

Kind of like a mild derivative of the bug we have on the jobs list page.

PR #15905, opened 2023-01-26
metrics: Add rate metrics to Client CSI endpoints

tgross:
Continues the work done in https://github.com/hashicorp/nomad/pull/15876 by extending rate metrics measurement over client CSI endpoints (the last batch of stuff!)

Also tightens up authentication for these endpoints by enforcing the server certificate name is valid. We protect these endpoints currently by mTLS and can't use an auth token because these endpoints are (uniquely) called by the leader and followers for a given node won't have the leader's ephemeral ACL token. Add a certificate name check that requests come from a server and not a client, because no client should ever send these RPCs directly.


shoenig (approved):
LGTM

PR #10444, opened 2021-04-24
UI: Namespaces redesign

DingoEatingFuzz:
This rethinks namespaces as a filter on list pages rather than a global setting. Namespaces are still part of job and CSI volume identifiers, so there was lots of plumbing involved.

The biggest net-new feature here is being able to select `All (*)` to list all jobs or CSI volumes across namespaces.

Jobs list page with new namespaces facet open to show options, including All

This PR has the following in it (in hindsight it should have been multiple PRs, sorry)

1. New `SingleSelectDropdown` component that wraps Ember Power Select and gives it a similar API as `MultiSelectDropdown`
2. Refactored multi select dropdown stories as facet stories and includes new ones for the single select dropdown
3. Removes the namespaces selector from the gutter menu
4. Conditionally adds a namespaces facet to the jobs, optimize, and csi volumes pages (when there are namespaces other than `default`)
5. Conditionally adds a namespaces column to jobs and volume tables
6. Removes some of the namespaces logic from the system service
7. Changes some data plumbing for jobs and volumes now that query is being used to add the namespace param rather than it being added invisibly within the data layer based on the global namespace setting.

PR #15906, opened 2023-01-26
ui: fix navigation for namespaced jobs in search and job version

### LINKED ISSUE #15893: Jumping to a Job via the global search does not work
### Nomad version
1.4.3

### Issue

Search for "traefik" in the global search box redirects to `ui/jobs/traefik?namespace=infra` once clicking on the link. This results in a 404. The proper link seems to be: `ui/jobs/traefik@infra`. 

### Reproduction steps

Create  a job in a namespace and try to search for it.

lgfa29:
Use the new URL structure for navigating to a job page.

I started in the global search, but did a quick search to find other places with the same problem. I found this [`gotoJobsForNamespace` function](https://github.com/hashicorp/nomad/blob/v1.4.3/ui/app/components/gutter-menu.js#L46-L60) that I think it's also wrong, but it doesn't seem to be called from anywhere so maybe we can delete it?

Closes #15893

### DIFF
--- a/ui/app/components/global-search/control.js
+++ b/ui/app/components/global-search/control.js
@@ -13,6 +13,7 @@ const MAXIMUM_RESULTS = 10;
 export default class GlobalSearchControl extends Component {
   @service router;
   @service token;
+  @service store;
 
   searchString = null;
 
@@ -180,14 +181,20 @@ export default class GlobalSearchControl extends Component {
   @action
   selectOption(model) {
     if (model.type === 'job') {
-      this.router.transitionTo('jobs.job', model.id, {
-        queryParams: { namespace: model.namespace },
+      const fullId = JSON.stringify([model.id, model.namespace]);
+      this.store.findRecord('job', fullId).then((job) => {
+        this.router.transitionTo('jobs.job', job.idWithNamespace);
       });
     } else if (model.type === 'node') {
       this.router.transitionTo('clients.client', model.id);
     } else if (model.type === 'task-group') {
-      this.router.transitionTo('jobs.job.task-group', model.jobId, model.id, {
-        queryParams: { namespace: model.namespace },
+      const fullJobId = JSON.stringify([model.jobId, model.namespace]);
+      this.store.findRecord('job', fullJobId).then((job) => {
+        this.router.transitionTo(
+          'jobs.job.task-group',
+          job.idWithNamespace,
+          model.id
+        );
       });
     } else if (model.type === 'plugin') {
       this.router.transitionTo('csi.plugins.plugin', model.id);
--- a/.changelog/15906.txt
+++ b/.changelog/15906.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fix navigation to pages for jobs that are not in the default namespace
+```
--- a/ui/app/components/gutter-menu.js
+++ b/ui/app/components/gutter-menu.js
@@ -42,20 +42,4 @@ export default class GutterMenu extends Component {
   transitionTo(destination) {
     return this.router.transitionTo(destination);
   }
-
-  gotoJobsForNamespace(namespace) {
-    if (!namespace || !namespace.get('id')) return;
-
-    // Jobs and CSI Volumes are both namespace-sensitive. Changing namespaces is
-    // an intent to reset context, but where to reset to depends on where the namespace
-    // is being switched from. Jobs take precedence, but if the namespace is switched from
-    // a storage-related page, context should be reset to volumes.
-    const destination = this.router.currentRouteName.startsWith('csi.')
-      ? 'csi.volumes'
-      : 'jobs';
-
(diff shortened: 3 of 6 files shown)


ui/app/components/global-search/control.js
```diff
@@ -180,14 +181,20 @@ export default class GlobalSearchControl extends Component {
   @action
   selectOption(model) {
     if (model.type === 'job') {
-      this.router.transitionTo('jobs.job', model.id, {
-        queryParams: { namespace: model.namespace },
+      const fullId = JSON.stringify([model.id, model.namespace]);
+      this.store.findRecord('job', fullId).then((job) => {
+        this.router.transitionTo('jobs.job', job.idWithNamespace);
       });
```
lgfa29, line 187: If we extract the logic to generate the proper namespaced job ID into a helper function we could avoid this store query, but I think it's fine this way?

philrenaud (approved):
Nice find, and this solves the bug well. I think you're safe to remove `gotoJobsForNamespace` in gutter-menu also.

lgfa29:
Ok, I found where `gotoJobsForNamespace` was being used and it was removed in https://github.com/hashicorp/nomad/pull/10444/files#diff-bce0972dc9d8b2d20b097153494f982afa17c33099a1611d9574e390c6d2ebbdL36, so I feel more confident that we can indeed remove it 😄

lgfa29:
@philrenaud the Percy failure seems flaky so I'm going to merge this one.

ui/app/components/global-search/control.js
```diff
@@ -180,14 +181,20 @@ export default class GlobalSearchControl extends Component {
   @action
   selectOption(model) {
     if (model.type === 'job') {
-      this.router.transitionTo('jobs.job', model.id, {
-        queryParams: { namespace: model.namespace },
+      const fullId = JSON.stringify([model.id, model.namespace]);
+      this.store.findRecord('job', fullId).then((job) => {
```
ChaiWithJai, line 185: `nit`: I think this API call is unnecessary here and we can't guarantee when the `router` transition will occur now. It could happen `> ~150ms` later. When we use `then` we don't guarantee when the JavaScript scheduler (event loop) will dispatch the callback.

I think we might have been able to get away with:

```
job.belongsTo('namespace').id()
```

This would have been synchronous.

ChaiWithJai (commented):
Great job! I think we could have avoided the API call, however, that's not a real blocker here.

ui/app/components/global-search/control.js
lgfa29: Ah interesting. Yeah, I thought about the added network request but I thought it would OK in this case since we're redirecting to the job page so we would need to fetch it anyway. If the job is not in the state store is the user impact a delay between the click and the redirect?

> I think we might have been able to get away with:
>
> ```
> job.belongsTo('namespace').id()
> ```

I think this would have the same problem? `job` is only available after the `this.store.findRecord('job', fullId)` call. We we could do to avoid this is to have something like:

```javascript
this.router.transitionTo('jobs.job', `${model.id}@{model.namespace}`);
```

But that duplicates the logic in `idWithNamespace()`. We could extract the logic to a shared helper function but that seemed a bit much in this case, but maybe not?

ChaiWithJai: A few things to chew on here:

1. Super subtle yet important difference between `model.namespace` and `.belongsTo('model').id()`. The prior makes a network request, the latter is synchronous.
2. What happens under the hood when you use `.then`? Our `.then` syntax is a misnomer. `.then` is really a setter function that says add this function to the array on my Promise object (that we can just call `onComplete`). So we dispatch a network request, continue the thread of execution, when the call stack is cleared, we'll check our callback queue where our `job` will be and then execute the callback function of `router.transitionTo`, then navigate to the `job` page and dispatch the same network request when we're on the `jobs.job.index` route to get fresh data.

The key thing worth noting here is when `router.transitionTo` will execute. A decent way of testing this behavior would be opening up your Developer Tools, go to the Network tab, simulate a slow network.

You did a great job on this PR and I'm really looking forward to more contributions from you! I figured you would just like to nerd out on what's actually happening under the hood.

lgfa29: Oh yeah, I'm taking this opportunity to learn more 😄 

One thing to note is that this component doesn't user Ember Data (or didn't prior to this PR). The variable names `model` is not _actually_ a model, it's just an object like this: https://github.com/hashicorp/nomad/blob/v1.4.3/ui/app/components/global-search/control.js#L69-L74, so `model.namespace` is just accessing a string in the object.

The `.then` part is interesting. There could be a ghost redirect in a slow connection where the user clicks a search result, then clicks somewhere else, and it's then redirected to the search result. Something to keep an eye on. Would making the action `async` and `await` the fetch prevent the early return until the promise finishes?

PR #15908, opened 2023-01-26
docs: how to troubleshoot consul connect envoy

### LINKED ISSUE #12389: Improve logging on Connect proxy error
### Proposal

When there is an error creating the bootstrap configuration for Connect proxy sidecar, the error is not properly surfaced. Currently there is no indication of what happened when Envoy fails to start allocations

### Use-cases

Ideally Nomad could pass this error through to the end user so they could better diagnose and fix the issue.

For instance, an ACL mismatch in Consul and/or notifications on expired Certificates would be logged/surfaced.

### Attempted Solutions

The following code is used to surface the bug in a local Nomad fork:

```
diff --git a/client/allocrunner/taskrunner/envoy_bootstrap_hook.go b/client/allocrunner/taskrunner/envoy_bootstrap_hook.go
index fa9c63a43..494c7398a 100644
--- a/client/allocrunner/taskrunner/envoy_bootstrap_hook.go
+++ b/client/allocrunner/taskrunner/envoy_bootstrap_hook.go
@@ -274,6 +274,7 @@ func (h *envoyBootstrapHook) Prestart(ctx context.Context, req *ifs.TaskPrestart

 	// keep track of latest error returned from exec-ing consul envoy bootstrap
 	var cmdErr error
+	var cmdOut string

 	// Since Consul services are registered asynchronously with this task
 	// hook running, retry until timeout or success.
@@ -323,6 +324,8 @@ func (h *envoyBootstrapHook) Prestart(ctx context.Context, req *ifs.TaskPrestart
 		// fails.
 		_ = os.Remove(bootstrapFilePath)

+		cmdOut = buf.String()
+
 		return true, cmdErr
 	}, exptime.BackoffOptions{
 		MaxSleepTime:   h.envoyBootstrapWaitTime,
@@ -332,7 +335,7 @@ func (h *envoyBootstrapHook) Prestart(ctx context.Context, req *ifs.TaskPrestart
 		// Wrap the last error from Consul and set that as our status.
 		_, recoverable := cmdErr.(*exec.ExitError)
 		return structs.NewRecoverableError(
-			fmt.Errorf("error creating bootstrap configuration for Connect proxy sidecar: %v", cmdErr),
+			fmt.Errorf("error creating bootstrap configuration for Connect proxy sidecar: %v \n %v", cmdErr, cmdOut),
 			recoverable,
 		)
 	}
```

gulducat:
Largely a doc-ification of this commit message: d47678074bf8ae9ff2da3c91d0729bf03aee8446

This doesn't spell out all (or any of) the possible failure modes, but should be a good starting point for folks.

Relates to issue #12389 which doesn't have a straightforward solution, but this documentation at least puts some steps out on our website for people to try out.

Closes #12389

### DIFF
--- a/client/allocrunner/taskrunner/envoy_bootstrap_hook.go
+++ b/client/allocrunner/taskrunner/envoy_bootstrap_hook.go
@@ -43,6 +43,10 @@ const (
 	envoyBootstrapMaxJitter = 500 * time.Millisecond
 )
 
+var (
+	errEnvoyBootstrapError = errors.New("error creating bootstrap configuration for Connect proxy sidecar")
+)
+
 type consulTransportConfig struct {
 	HTTPAddr   string // required
 	Auth       string // optional, env CONSUL_HTTP_AUTH
@@ -373,7 +377,10 @@ func (h *envoyBootstrapHook) Prestart(ctx context.Context, req *ifs.TaskPrestart
 		// Wrap the last error from Consul and set that as our status.
 		_, recoverable := cmdErr.(*exec.ExitError)
 		return structs.NewRecoverableError(
-			fmt.Errorf("error creating bootstrap configuration for Connect proxy sidecar: %v", cmdErr),
+			fmt.Errorf("%w: %v; see: <https://www.nomadproject.io/s/envoy-bootstrap-error>",
+				errEnvoyBootstrapError,
+				cmdErr,
+			),
 			recoverable,
 		)
 	}
--- a/website/content/docs/integrations/consul-connect.mdx
+++ b/website/content/docs/integrations/consul-connect.mdx
@@ -371,6 +371,32 @@ dashes (`-`) are converted to underscores (`_`) in environment variables so
 - Prior to Consul 1.9, the Envoy sidecar proxy will drop and stop accepting
   connections while the Nomad agent is restarting.
 
+## Troubleshooting
+
+If the sidecar service is not running correctly, you can investigate
+potential `envoy` failures in the following ways:
+
+* Task logs in the associated `connect-*` task
+* Task secrets (may contain sensitive information):
+  * envoy CLI command: `secrets/.envoy_bootstrap.cmd`
+  * environment variables: `secrets/.envoy_bootstrap.env`
+* An extra Allocation log file: `alloc/logs/envoy_bootstrap.stderr.0`
+
+For example, with an allocation ID starting with `b36a`:
+
+```shell-session
+nomad alloc status -short b36a  # to get the connect-* task name
+nomad alloc logs -task connect-proxy-count-api -stderr b36a
+nomad alloc exec -task connect-proxy-count-api b36a cat secrets/.envoy_bootstrap.cmd
+nomad alloc exec -task connect-proxy-count-api b36a cat secrets/.envoy_bootstrap.env
+nomad alloc fs b36a alloc/logs/envoy_bootstrap.stderr.0
+```
+
+Note: If the alloc is unable to start successfully, debugging files may
+only be accessible from the host filesystem. However, the sidecar task secrets
+directory may not be available in systems where it is mounted in a temporary
+filesystem.
+
 [count-dashboard]: /img/count-dashboard.png
 [consul_acl]: https://github.com/hashicorp/consul/issues/7414
 [gh-9907]: https://github.com/hashicorp/nomad/issues/9907
--- a/.changelog/15908.txt
+++ b/.changelog/15908.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+docs: link to an envoy troubleshooting doc when envoy bootstrap fails
+```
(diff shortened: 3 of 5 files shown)


website/content/docs/integrations/consul-connect.mdx
```diff
@@ -371,6 +371,30 @@ dashes (`-`) are converted to underscores (`_`) in environment variables so
...
+```shell-session
+nomad alloc status -short b36a  # to get the connect-* task name
+nomad alloc logs -task connect-proxy-count-api -stderr b36a
+nomad alloc exec -task connect-proxy-count-api b36a cat secrets/.envoy_bootstrap.cmd
+nomad alloc exec -task connect-proxy-count-api b36a cat secrets/.envoy_bootstrap.env
+nomad alloc fs b36a alloc/logs/envoy_bootstrap.stderr.0
+```
+
+Note: If the alloc is unable to start successfully, debugging files
+may only be accessible from the host filesystem.
```
lgfa29, line 396: I think the `secrets` dir is mounted in a `tmpfs` in Linux, so they would not be available in the host.

lgfa29 (approved):
LGTM! Minor nit-pick on the secrets dir

website/content/docs/integrations/consul-connect.mdx
gulducat: I went with this to soften the wording from the original commit message, which was

> If the alloc is unable to start successfully, the debugging files **are**
> only accessible from the host filesystem.

I debated leaving this note off entirely, but thought it might be useful for folks to consider, even if it doesn't provide a complete answer.  Would you suggest I mention secrets and tmpfs in particular here?

gulducat:
I added a link to the error message, and took a little detour to get tests to pass via error wrapping instead of literal string matching.  How does that look to ya @lgfa29 ?  It appears in a task event as

> Task hook failed | envoy_bootstrap: error creating bootstrap configuration  for Connect proxy sidecar: <nil>; see:  <https://www.nomadproject.io/s/envoy-bootstrap-error>

The redirect link doesn't exist yet, so probably should hold off on merging this until that's set up.  The PR for that is here: hashicorp/dev-portal#1585

client/allocrunner/taskrunner/envoy_bootstrap_hook.go
```diff
@@ -373,7 +377,11 @@ func (h *envoyBootstrapHook) Prestart(ctx context.Context, req *ifs.TaskPrestart
 		// Wrap the last error from Consul and set that as our status.
 		_, recoverable := cmdErr.(*exec.ExitError)
 		return structs.NewRecoverableError(
-			fmt.Errorf("error creating bootstrap configuration for Connect proxy sidecar: %v", cmdErr),
+			fmt.Errorf("%w: %v; see: %s",
+				errEnvoyBootstrapError,
+				cmdErr,
+				"<https://www.nomadproject.io/s/envoy-bootstrap-error>",
+			),
```
lgfa29, line 384:
```suggestion
			fmt.Errorf("%w: %v; see: <https://www.nomadproject.io/s/envoy-bootstrap-error>",
				errEnvoyBootstrapError,
				cmdErr,
			),
```

I think we can just put the URL directly in the string template? It doesn't seem to add much to have it as a param.

website/content/docs/integrations/consul-connect.mdx
lgfa29: I think a explicit referent that `secrets` may not be available at the host may be better, specially with this being a troubleshooting document, something went wrong so we should provide the most accurate information we can.

Maybe something like this? Kind of too long though 😅 

> Note: If the alloc is unable to start successfully, debugging files may only be accessible from the host filesystem. The sidecar task `secrets` directory may not be available in systems where it is mounted in a temporary file system.

lgfa29 (approved):

client/allocrunner/taskrunner/envoy_bootstrap_hook.go
gulducat: Oh yeah, that was from a previous iteration where I had a variable just for the link.  I'm still a little intrigued by the idea that all doc links in code might be set as variables all in one place, but that's a big tangent I don't want to take right now.

website/redirects.js
```diff
@@ -24,6 +24,22 @@ module.exports = [
     permanent: true,
   },
   */
+
+  // /s/* redirects for useful links that need a stable URL but we may need to
+  // change its destination in the future.
+  {
+    source: '/s/port-plan-failure',
+    destination:
```
kendallstrautman, line 32: This is pointing at a redirected path. It should point to a relative developer path.

`/nomad/docs/operations/monitoring-nomad#progress`

website/redirects.js
```diff
@@ -24,6 +24,22 @@ module.exports = [
...
+      'https://www.nomadproject.io/docs/operations/monitoring-nomad#progress',
+    permanent: false,
+  },
+  {
+    source: '/s/envoy-bootstrap-error',
+    destination:
+      'https://developer.hashicorp.com/nomad/docs/integrations/consul-connect#troubleshooting',
```
kendallstrautman, line 39:
```suggestion
      '/nomad/docs/integrations/consul-connect#troubleshooting',
```

website/redirects.js
lgfa29: Unfortunately redirects with path fragments (`#`) don't work if they are relative, the fragment gets lost between hops. That's why we have to use absolute addresses here.

BRKalow: I believe the issue with fragments should have been fixed earlier this week 👀

PR #15909, opened 2023-01-26
ui: fix alloc memory stats to match CLI output

### LINKED ISSUE #15830: Exec driver is showing cache+RSS in memory usage UI view
### Nomad version

Amazon Linux 2
### Issue
Nomad exec driver seems to be reporting cache+RSS in UI view, causing confusion among operators that memory usage of the application is too high.
### Reproduction steps
Launch an exec driver job that uses an external volume and reads a large file
#### Expected Result
Nomad's memory usage UI view should indicate memory used by the process as seems to be the case with docker driver.
#### Actual Result
Nomad's memory usage UI shows cache+RSS.

UI allocation view
Screenshot from 2023-01-20 09-07-12

nomad alloc stats <allocation>
Screenshot from 2023-01-20 09-07-35

lgfa29:
Match memory usage logic with the [CLI](https://github.com/hashicorp/nomad/blob/v1.4.3/command/alloc_status.go#L623-L628).

Closes #15830

### DIFF
--- a/.changelog/15909.txt
+++ b/.changelog/15909.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fix allocation memory chart to display the same value as the CLI
+```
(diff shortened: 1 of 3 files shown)


philrenaud (approved):

ChaiWithJai (approved):
LGTM

PR #15912, opened 2023-01-26
[ui] Dynamic Node Metadata editing

philrenaud:
Allows for editing node metadata via the Web UI. New metadata can be added and existing metadata can be edited or removed.

Some nice things like escape press within an input are handled as you'd expect.

Requires `can write node` access levels so, if testing against vercel/mirage, please "sign in" as a manager

image

### DIFF
--- a/ui/app/adapters/node.js
+++ b/ui/app/adapters/node.js
@@ -61,4 +61,16 @@ export default class NodeAdapter extends Watchable {
       },
     });
   }
+
+  getMeta(node) {
+    const url = `/v1/client/metadata?node_id=${node.id}`;
+    return this.ajax(url, 'GET');
+  }
+
+  addMeta(node, newMeta) {
+    const url = `/v1/client/metadata?node_id=${node.id}`;
+    return this.ajax(url, 'POST', {
+      data: {Meta: newMeta},
+    });
+  }
 }
--- a/ui/app/adapters/node.js
+++ b/ui/app/adapters/node.js
@@ -62,11 +62,6 @@ export default class NodeAdapter extends Watchable {
     });
   }
 
-  getMeta(node) {
-    const url = `/v1/client/metadata?node_id=${node.id}`;
-    return this.ajax(url, 'GET');
-  }
-
   addMeta(node, newMeta) {
     const url = `/v1/client/metadata?node_id=${node.id}`;
     return this.ajax(url, 'POST', {
--- a/ui/app/adapters/node.js
+++ b/ui/app/adapters/node.js
@@ -65,7 +65,7 @@ export default class NodeAdapter extends Watchable {
   addMeta(node, newMeta) {
     const url = `/v1/client/metadata?node_id=${node.id}`;
     return this.ajax(url, 'POST', {
-      data: {Meta: newMeta},
+      data: { Meta: newMeta },
     });
   }
 }
--- a/ui/app/controllers/clients/client/index.js
+++ b/ui/app/controllers/clients/client/index.js
@@ -293,8 +293,9 @@ export default class ClientController extends Controller.extend(
     console.log("VALID KEY?", key, this.model.meta);
   }
 
-  @action async addDynamicMetaData() {
+  @action async addDynamicMetaData(e) {
     try {
+      e.preventDefault();
       await this.model.addMeta({[this.newMetaData.key]: this.newMetaData.value});
       // Clear newMetaData
       this.newMetaData = {
--- a/ui/app/controllers/clients/client/index.js
+++ b/ui/app/controllers/clients/client/index.js
@@ -16,12 +16,16 @@ import {
 } from 'nomad-ui/utils/qp-serialize';
 import classic from 'ember-classic-decorator';
 import localStorageProperty from 'nomad-ui/utils/properties/local-storage';
+import { inject as service } from '@ember/service';
 
 @classic
 export default class ClientController extends Controller.extend(
   Sortable,
   Searchable
 ) {
+
+  @service flashMessages;
+
   queryParams = [
     {
       currentPage: 'page',
@@ -277,4 +281,39 @@ export default class ClientController extends Controller.extend(
       this.set('activeTask', null);
     }
   }
+  
+  // #region metadata
+  newMetaData = {
+    key: '',
+    value: '',
+  }
+
+  @action validateKey(key) {
+    console.log("VALID KEY?", key, this.model.meta);
+  }
+
+  @action async addDynamicMetaData() {
+    try {
+      await this.model.addMeta({[this.newMetaData.key]: this.newMetaData.value});
+      this.flashMessages.add({
+        title: 'Metadata added',
+        message: `${this.newMetaData.key} successfully saved`,
+        type: 'success',
+        destroyOnClick: false,
+        timeout: 3000,
+      });
+
+    } catch (err) {
+      const error = messageFromAdapterError(err) || 'Could not save new dynamic metadata';
+      this.flashMessages.add({
+        title: `Error saving Metadata`,
+        message: error,
+        type: 'error',
+        destroyOnClick: false,
+        sticky: true,
+      });
+    }
+
+  }
+  // #endregion metadata
 }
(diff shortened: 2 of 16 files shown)


ui/app/templates/components/attributes-section.hbs
```diff
@@ -5,14 +5,11 @@
         {{#if this.prefix}}<span class="is-faded" data-test-prefix>{{this.prefix}}.</span>{{/if}}{{key}}
       </td>
     </tr>
-    <AttributesSection @prefix={{if this.prefix (concat this.prefix "." key) key}} @attributes={{value}} />
+    <AttributesSection @prefix={{if this.prefix (concat this.prefix "." key) key}} @attributes={{value}}
+    @key={{key}} @value={{value}} @editable={{@editable}} @onKVEdit={{@onKVEdit}} @onKVSave={{@onKVSave}} />
```
ChaiWithJai, line 9: `question`: The component is invoking itself?

ui/app/templates/components/attributes-table.hbs
```diff
@@ -1,11 +1,16 @@
...
       <th class="is-one-third">Name</th>
       <th>Value</th>
     </tr>
   </thead>
   <tbody>
-    <AttributesSection @attributes={{@attributePairs}} />
+    <AttributesSection @attributes={{@attributePairs}}
+    @editable={{@editable}}
+    @onKVEdit={{@onKVEdit}}
+    @onKVSave={{@onKVSave}}
```
ChaiWithJai, line 12: `nit`: we're prop drilling these down 4 components. We could use a [closure component](https://guides.emberjs.com/release/components/block-content/) in `AttributesSection` to yield these values out? Adding a new function that floats down to `MetadataEditor` would make this a little challenging to track down.

ui/app/templates/components/attributes-section.hbs
philrenaud: Yep. This is a nested / tree structure and has been there awhile by the looks of things. For properties that contain a `.` delimiter, it splits them up visually (see items under "connect"):

image

ui/app/templates/clients/client/index.hbs
```diff
@@ -835,5 +838,40 @@
...
+          <MetadataEditor
+            @kv={{this.newMetaData}}
+            @onEdit={{this.validateKey}}
+          >
+            <button
+              disabled={{or (not this.newMetaData.key) (not this.newMetaData.value)}}
+              type="submit"
+              class="button is-primary"
+              {{on "click" (queue
+                (action this.addDynamicMetaData this.newMetaData)
```
ChaiWithJai, line 855: If this promise is rejected, the functions that come after this will not execute. We're not relying on the return value of any of these functions, just the promise resolution before we reset.

I like handling things in templates and I don't think you need to change this. But just documenting this, if we run into a problem where state doesn't update... we'll know why.

ui/app/models/structured-attributes.js
```diff
@@ -8,6 +8,11 @@ const { unflatten } = flat;
 export default class StructuredAttributes extends Fragment {
   @attr() raw;
 
+  recomputeRawProperties(incoming) {
+    this.set('raw', incoming);
+    this.notifyPropertyChange('structured');
```
ChaiWithJai, line 13: `this.set` [internally calls](https://discuss.emberjs.com/t/notifypropertychange-behavior/14981) `this.notifyPropertyChange`.

I think the issue is that we're not setting `raw` to a new reference in memory since we're just setting it to `incoming`.

I think we might be able to get away with:

```
  recomputeRawProperties(incoming) {
    this.set('raw', {...incoming}); // this will set raw to a new reference in memory that should trigger the dependency key to notice that a change actually occured
  }

```

ChaiWithJai (commented):
Looks great to me!

ui/app/models/structured-attributes.js
philrenaud: Good call, that's cleaner.

ui/app/models/structured-attributes.js
```diff
@@ -8,6 +8,11 @@ const { unflatten } = flat;
 export default class StructuredAttributes extends Fragment {
   @attr() raw;
 
+  recomputeRawProperties(incoming) {
+    this.set('raw', incoming);
```
ChaiWithJai, line 12: Calling `set` means we'll have to use the `classic` decorator here which means we're not Octane compatible.

The [docs](https://guides.emberjs.com/release/models/creating-updating-and-deleting-records/) are saying we might be able to get away with dot notation.

ChaiWithJai: As in `this.raw = {...incoming}`

ui/app/templates/components/attributes-section.hbs
ChaiWithJai: Cool!

ui/app/models/structured-attributes.js
philrenaud: `this.raw = incoming` seems to work so going with that, without object deconstruction cloning

ui/app/templates/components/attributes-table.hbs
philrenaud: I ended up playing with this idea a bit yesterday and no matter how I worked it, it felt either a little less legible or a bit clunky. I'm going to opt to leave these as passed functions. The names are specific enough that they will be easy to track down.

ui/app/controllers/clients/client/index.js
```diff
@@ -277,4 +281,60 @@ export default class ClientController extends Controller.extend(
...
+
+  @action resetNewMetaData() {
+    this.newMetaData = {
+      key: '',
+      value: '',
+    };
+  }
+
+  @action validateMetadata(event, b, c) {
+    console.log('validatin', event, b, c);
```
ChaiWithJai, line 308: `nit`: log statement

ui/app/models/node.js
```diff
@@ -153,4 +153,17 @@ export default class Node extends Model {
...
   }
+
+  async addMeta(newMeta) {
+    let metaResponse = await this.store
+      .adapterFor('node')
+      .addMeta(this, newMeta);
+
+    if (!this.meta) {
+      this.set('meta', this.store.createFragment('structured-attributes'));
+    }
```
philrenaud, line 164: `null !== []` so we create a fragment with raw/recompute/structured methods on it if there's nothing at all.

ui/app/controllers/clients/client/index.js
```diff
@@ -277,4 +281,59 @@ export default class ClientController extends Controller.extend(
...
+      this.flashMessages.add({
+        title: 'Metadata added',
+        message: `${key} successfully saved`,
+        type: 'success',
+        destroyOnClick: false,
+        timeout: 3000,
+      });
+    } catch (err) {
+      const error =
+        messageFromAdapterError(err) || 'Could not save new dynamic metadata';
```
ChaiWithJai, line 328: `nit`: `messageFromAdapterError` never returns `void` because it returns `Unknown error` if no cases are met. You can remove the logical OR operator.

ui/app/adapters/node.js
```diff
@@ -61,4 +61,11 @@ export default class NodeAdapter extends Watchable {
       },
     });
   }
+
+  addMeta(node, newMeta) {
```
ChaiWithJai, line 65: `nit`: first param should probably just be `nodeId`. I was confused when we were passing `this` via the model.

ui/app/models/node.js
ChaiWithJai: `question`: So this is a defensive technique we're using before calling `recomputeRawProperties` which then computes the derived state for `structured`?

Where are we comparing `null` to `[]`? Is this something we're doing for the template layer?

ChaiWithJai (approved):
Left some comments.

schmichael (approved):
Not sure my +1 should mean much, but LGTM!

ui/app/models/node.js
philrenaud: If we have no metadata, we get a null back. If we had metadata but use dynamic node metadata to overwrite it all to null values, we get `[]` back. In both cases we need to have actual `structured-attributes`.

PR #15918, opened 2023-01-27
acl: RPC endpoints for JWT auth

pkazmierczak:
This PR introduces RPC endpoints for JWT auth.


nomad/structs/acl.go
```diff
@@ -898,6 +926,90 @@ func (a *ACLAuthMethodConfig) Copy() *ACLAuthMethodConfig {
...
+		ClockSkewLeeway  any
+		*Alias
+	}{
+		Alias: (*Alias)(a),
+	}
+	if err = json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+	if aux.ExpirationLeeway != nil {
+		switch v := aux.ExpirationLeeway.(type) {
```
shoenig, line 972: should we have a `default` case that returns error on unexpected type?

shoenig (approved):
disclaimer: I'm not a JWT/auth expert

nomad/structs/acl.go
```diff
@@ -1378,9 +1493,38 @@ func (a *ACLOIDCCompleteAuthRequest) Validate() error {
...
+
+// ACLLoginRequest is the request object to begin auth with an external bearer
+// token provider.
+type ACLLoginRequest struct {
+
+	// AuthMethodName is the name of the auth method being used to login. This
+	// is a required parameter.
+	AuthMethodName string
+
+	// LoginToken is the token used to login. This is a required parameter.
```
jrasell, line 1511: Would you be able to expand where this token comes from? It's not clear to me exactly where this comes from, and whether this is different from a Nomad ACL token.

nomad/mock/acl.go
```diff
@@ -228,16 +233,17 @@ func ACLAuthMethod() *structs.ACLAuthMethod {
...
-			OIDCClientID:        "mock",
-			OIDCClientSecret:    "very secret secret",
-			OIDCScopes:          []string{"groups"},
-			BoundAudiences:      []string{"audience1", "audience2"},
-			AllowedRedirectURIs: []string{"foo", "bar"},
-			DiscoveryCaPem:      []string{"foo"},
-			SigningAlgs:         []string{"bar"},
-			ClaimMappings:       map[string]string{"foo": "bar"},
-			ListClaimMappings:   map[string]string{"foo": "bar"},
+			JWTValidationPubKeys: []string{},
```
jrasell, line 236: Would it be better to have separate mocks for OIDC and JWT based auth methods?

nomad/acl_endpoint.go
```diff
@@ -2735,3 +2741,144 @@ func (a *ACL) OIDCCompleteAuth(
 	reply.ACLToken = tokenUpsertReply.Tokens[0]
 	return nil
 }
+
+// Login RPC performs non-interactive auth using a given AuthMethod. This method
+// can not be used for OIDC login flow.
+func (a *ACL) Login(args *structs.ACLLoginRequest, reply *structs.ACLCompleteAuthResponse) error {
```
jrasell, line 2747: Do we want to version gate this RPC? In a situation where one server in the region or one region in a federated cluster has not been updated to a version that includes this, I believe the call will always error.

lib/auth/jwt/validator.go
```diff
@@ -0,0 +1,126 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to parse public key for JWT auth: %v", err)
+		}
+	}
+	return jwt.NewStaticKeySet(parsedKeys)
+}
+
+func usingJWKS(ctx context.Context, jwksurl, jwkscapem string) (jwt.KeySet, error) {
+	// Measure the JWKS endpoint performance.
+	defer metrics.MeasureSince([]string{"nomad", "acl", "jwks"}, time.Now())
```
jrasell, line 102: What would an operator to use this to measure? I find the name clashes with RPC telemetry names and therefore initially seems an RPC measure to me.

lib/auth/jwt/validator.go
```diff
@@ -0,0 +1,126 @@
...
+	keySet, err := jwt.NewJSONWebKeySet(ctx, jwksurl, jwkscapem)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get validation keys from JWKS: %v", err)
+	}
+	return keySet, nil
+}
+
+func usingOIDC(ctx context.Context, oidcurl string, oidccapem []string) (jwt.KeySet, error) {
+	// Measure the OIDC endpoint performance.
+	defer metrics.MeasureSince([]string{"nomad", "acl", "oidc_jwt"}, time.Now())
```
jrasell, line 113: What would an operator to use this to measure? I find the name clashes with RPC telemetry names and therefore initially seems an RPC measure to me.

lib/auth/jwt/validator.go
```diff
@@ -0,0 +1,126 @@
...
+			)
+		}
+		var iss string
+		var ok bool
+		if iss, ok = claims["iss"].(string); !ok {
+			return nil, fmt.Errorf("unable to read iss property of provided token")
+		}
+		if !slices.Contains(methodConf.BoundIssuer, iss) {
+			return nil, fmt.Errorf("invalid JWT issuer: %v", claims["iss"])
+		}
```
jrasell, line 82: very much a take it or leave it suggestion:
```suggestion
		if iss, ok := claims["iss"].(string); !ok {
			return nil, fmt.Errorf("unable to read iss property of provided token")
		} else if !slices.Contains(methodConf.BoundIssuer, iss) {
			return nil, fmt.Errorf("invalid JWT issuer: %v", claims["iss"])
		}
```

lib/auth/jwt/validator.go
```diff
@@ -0,0 +1,126 @@
...
+		return nil, fmt.Errorf("unable to get validation keys from JWKS: %v", err)
+	}
+	return keySet, nil
+}
+
+func usingOIDC(ctx context.Context, oidcurl string, oidccapem []string) (jwt.KeySet, error) {
+	// Measure the OIDC endpoint performance.
+	defer metrics.MeasureSince([]string{"nomad", "acl", "oidc_jwt"}, time.Now())
+
+	// TODO why do we have DiscoverCaPem as an array but JWKSCaPem as a single string?
```
jrasell, line 115: Is this something we want to figure out before merging, or is OK for a later date?

lib/auth/jwt/validator.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/hashicorp/nomad/helper"
+	"github.com/hashicorp/nomad/nomad/structs"
+)
+
+// Validate performs token signature verification and JWT header validation,
+// and returns a list of claims or an error in case any validation or signature
+// verification fails.
+func Validate(ctx context.Context, token string, methodConf *structs.ACLAuthMethodConfig) (map[string]any, error) {
+	var keySet jwt.KeySet
+	var err error
```
jrasell, line 22: typical go style to declare multiple vars:

```suggestion
	var (
		keySet jwt.KeySet
		err    error
	)
```

jrasell (changes requested):
Looking great, thanks for taking this on!  

Along with the inline comments, I feel we should expand the auth-method validate function, so that we have validation workflows for either JWT or OIDC. This will make the UX better and also removes current ambiguity in the `jwt.Validate` function whereby all three options could have been configured, but the first will only ever be used.

nomad/structs/acl.go
```diff
@@ -1378,9 +1493,38 @@ func (a *ACLOIDCCompleteAuthRequest) Validate() error {
 	return mErr.ErrorOrNil()
 }
 
-// ACLOIDCCompleteAuthResponse is the response when the OIDC auth flow has been
+// ACLCompleteAuthResponse is the response when the auth flow has been
 // completed successfully.
-type ACLOIDCCompleteAuthResponse struct {
+type ACLCompleteAuthResponse struct {
```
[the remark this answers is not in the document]
pkazmierczak: resolved in [eccc9a6](https://github.com/hashicorp/nomad/pull/15918/commits/eccc9a6941e346af13adc9303f2d2250949c0bdf)

nomad/mock/acl.go
pkazmierczak: I think that's actually a good shout. Separated in [92ef601](https://github.com/hashicorp/nomad/pull/15918/commits/92ef60197693583d59358eae059909de0fe6c00c)

nomad/acl_endpoint.go
pkazmierczak: Good shout, I gated to 1.5.2 with a TODO comment in the leader. See [28efef6](https://github.com/hashicorp/nomad/pull/15918/commits/28efef61a282d311e4414d37c4d387e6eacd54f8) if you're ok with the implementation.

Ayomi-gh (approved):
All looks good 👍

jrasell (approved):
LGTM; minor comment inline but that's not a blocker.

nomad/acl_endpoint.go
```diff
@@ -2735,3 +2741,152 @@ func (a *ACL) OIDCCompleteAuth(
...
+			Region:    a.srv.Region(),
+			AuthToken: a.srv.getLeaderAcl(),
+		},
+	}
+
+	var tokenUpsertReply structs.ACLTokenUpsertResponse
+
+	if err := a.srv.RPC(structs.ACLUpsertTokensRPCMethod, &tokenUpsertRequest, &tokenUpsertReply); err != nil {
+		return err
+	}
```
tgross, line 2884: Calling this RPC will work but we'll have to go thru authentication again, check for forwarding again, write metrics as though a second RPC was called, etc. 

It'd probably be better to pull the shared bits out of the `ACLUpsertTokens` RPC into its own method and have that called from both `ACLUpsetTokens` and `ACLLogins`.

tgross: (Fine to do in a follow-up PR)

pkazmierczak: I'll make a PR against `main` once the feature branch is merged.

pkazmierczak: addressed in https://github.com/hashicorp/nomad/pull/17194

PR #15919, opened 2023-01-27
docs: add post-install steps for CNI to main install docs page

tgross:
The getting started Tutorial has a post-installation steps section that includes installing CNI plugins. Many users will want to use `bridge` networking right out of the gate, so adding these same post-install instructions to the main docs will be a better Day 0 experience for them.

### DIFF
--- a/website/content/docs/install/index.mdx
+++ b/website/content/docs/install/index.mdx
@@ -13,7 +13,7 @@ operating systems. You can also [build Nomad from source](#from-source).
 [Quickstart](/nomad/docs/install/quickstart) for options to get started with Nomad.
 
 <Tabs>
-<Tab heading="Manual installation">
+<Tab heading="Manual installation" group="manual">
 
 You can download a [precompiled binary](/nomad/downloads) and
 run it on your machine locally. You can also verify the binary using the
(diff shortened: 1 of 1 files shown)


website/content/docs/install/index.mdx
```diff
@@ -21,32 +21,6 @@ available SHA-256 sums. After downloading Nomad, unzip the package. Make sure
...
-with the command `source ~/.bashrc`, replacing `bash` with the name of your
-shell.
-
--> **Tip (Windows):** Add a location to your path via the GUI by navigating to
-`Environment Variables` in your system settings, and looking for the variable
-called `PATH`. You will see a semicolon-separated list of locations. Add the
-Nomad binary's location to that list and then launch a new console window.
-
 </Tab>
 <Tab heading="Linux Packages">
```
lgfa29, line 25:
```suggestion
<Tab heading="Linux Packages" group="linux">
```

Having the matching `group` here keeps the tabs in sync with the ones below. There's also `manual`, `homebrew`, and `chocolatey`.

Also, for some reason the tab name and order here is different from the the Tutorial, so it would be nice for them to match.

image

image

lgfa29 (changes requested):

website/content/docs/install/index.mdx
tgross: Oh, cool. Done in 6640357

tgross:
Had to squash to nudge Vercel along (I think), but your suggested changes have been done @lgfa29

website/content/docs/install/index.mdx
```diff
@@ -145,54 +163,140 @@ $ sudo yum -y install nomad
...
-First, install the HashiCorp tap, a repository of all of the HashiCorp Homebrew
-packages.
+These steps are considered optional but can be helpful for running Nomad and to
+take advantage of additional Nomad functionalities.
+
+<Tabs>
+<Tab heading="Manual installation" group="manual">
+
+<Tabs>
+<Tab heading="Linux">
```
lgfa29, line 177:
```suggestion
<Tab heading="Linux Packages">
```

Matching the name with the tab above. The tutorial uses just `Linux` for this tab, but looking at it again I think `Linux Packages` is more accurate.

tgross:
Forgot to backport this and that created a conflict when I backported https://github.com/hashicorp/nomad/pull/15926. I've cherry-picked this into 1.4.x, 1.3.x, and 1.2.x with https://github.com/hashicorp/nomad/pull/15931 https://github.com/hashicorp/nomad/pull/15930 and https://github.com/hashicorp/nomad/pull/15929

PR #15926, opened 2023-01-27
docs: add more warnings about running agent as root on Linux

tgross:


shoenig (approved):
LGTM

website/content/docs/install/quickstart.mdx
```diff
@@ -44,6 +44,9 @@ If you would like to try Nomad locally, you can install Nomad on your local
 machine using the same steps that you would for a production environments and
 run a single-node development instance using the `nomad agent -dev` command
 
+-> **Note:** If you are running Nomad on Linux, you'll need to run with `sudo`
+so that cpuset accounting and network namespaces work correctly.
+
```
schmichael, line 49: These instructions are already very `-> Note` heavy, so I think it would be sufficient to just include sudo in the command we tell users to run.

```suggestion
run a single-node development instance using the `sudo nomad agent -dev` command
```

website/content/docs/operations/nomad-agent.mdx
```diff
@@ -38,6 +38,10 @@ The agent is started with the [`nomad agent` command](/nomad/docs/commands/agent
 command blocks, running forever or until told to quit. The agent command takes a variety
 of configuration options, but most have sane defaults.
 
+-> **Note:** If you are running Nomad on Linux, you'll need to run client agents
+as root (or with `sudo`) so that cpuset accounting and network namespaces work
+correctly.
+
 When running `nomad agent`, you should see output similar to this:
 
 ```shell-session
```
schmichael, line 47: Argh, one line away from being able to leave this as a suggestion.

I think we should add `sudo` to the example command below.

schmichael (approved):
Thanks!

Issue #15359, opened 2022-11-22
Specify that `-` is replaced by `_` when changing meta keys into environment variables


Kamilcuk:
On the documentation page https://developer.hashicorp.com/nomad/docs/runtime/environment#meta it is not explained that meta keys have replaced `-` for `_` when transformed into environment variables. Although variables with `-` in the name can't be accessed with unix shell, they are technically perfectly valid environment variable names. I tried to follow the source code https://github.com/hashicorp/nomad/blob/main/client/taskenv/env.go#L115 but with no success.

### Proposal

Please specify in the documentation that meta keys containing `-` are replaced by `_` when transformed into environment variables.

Please also specify if other characters are replaced? Do all non-alphanumeric characters in meta keys are replaced by underscores?

### Use-cases

Currently, this behavior I believe would count as "unspecified". Or it is specified in the documentation, but I have not found it, that's on me.

### Attempted Solutions

```
job "test_meta_dash" {
  datacenters = [
    "wee-dev",
  ]
  type = "service"
  namespace = "dev"
  meta {
    key-key = "value"
  }
  group "test" {
    task "test" {
      driver = "docker"
      config {
        image = "busybox"
        command = "sh"
        args = [
          "-xc",
          "env | grep NOMAD_META_"
        ]
      }
    }
  }
}
```

outputs:

```
NOMAD_META_KEY_KEY=value
NOMAD_META_key_key=value
```

The environment variable is `NOMAD_META_key_key` not `NOMAD_META_key-key`.

tgross:
Hi @Kamilcuk! This definitely needs better documentation. We should probably also update the docs for the [`meta`](https://developer.hashicorp.com/nomad/docs/job-specification/meta) block. I'll mark this for fixing but if you're interested in opening a PR with the fix I'd be happy to review it as well.

In the meantime, the code that calls this is in [`taskenv/env.go#L611-L622`](https://github.com/hashicorp/nomad/blob/v1.4.3/client/taskenv/env.go#L611-L622), and it "cleans" any of the Nomad-owned environment variable paths. That [`CleanEnvVar`](https://github.com/hashicorp/nomad/blob/v1.4.3/helper/funcs.go#L206-L223) function removes anything that isn't in `[A-Za-z0-9_\.]`. This was introduced in https://github.com/hashicorp/nomad/issues/2405

Kamilcuk:
Amazing, thank you for code links and clarification.

PR #15940, opened 2023-01-29
Fix documentation for `meta` block: string replacement in key from `-` to `_`

dttung2905:
Hi team,

This PR will solve the issue #15359

### DIFF
--- a/.changelog/15940.txt
+++ b/.changelog/15940.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+docs: Added further explanation that any `-` string in key in `meta` block will be replaced with `_`
+```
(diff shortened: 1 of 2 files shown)


jrasell (changes requested):
Hi @dttung2905 and thanks for raising this PR. It is looking good, however, we don't require changelog entries for documentation updates. Could you remove this, I can then get this merged. Thanks.

dttung2905:
Thank @jrasell for the super quick review.  I have removed the changelog like you suggested. You also answered what I had in mind already when I searched for past PR by `docs` tag but found no related changelog too. :pray:

jrasell (approved):
LGTM, thanks @dttung2905!

PR #15941, opened 2023-01-30
renamed "stanza" to "block" for consistency with other projects

### LINKED ISSUE #15916: Globally replace "stanza" with "block" when referring to HCL
Nomad is lagging behind Terraform and other HCL-enabled projects by calling `foo { ... }` a "stanza" instead of a "block" in docs.

We should fix that and do a big global search and replace on `stanza -> block`. I can't think of anywhere this would cause backward compatibility issues, but if any are encountered we will have to stop and consider each one.

pkazmierczak:
resolves #15916

### DIFF
--- a/.release/ci.hcl
+++ b/.release/ci.hcl
@@ -170,7 +170,7 @@ event "fossa-scan" {
 }
 
 ## These are promotion and post-publish events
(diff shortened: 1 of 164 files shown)


tgross (approved):
LGTM!

PR #15961, opened 2023-01-30
client: Add option to enable hairpinMode on Nomad bridge

angrycub:
- Add `bridge_network_hairpin_mode` client config setting
- Add hairpin mode to bridge template; convert to go template
- add node attribute: nomad.bridge.hairpin_mode
- Add documentation

### DIFF
--- a/client/allocrunner/network_manager_linux.go
+++ b/client/allocrunner/network_manager_linux.go
@@ -184,7 +184,7 @@ func newNetworkConfigurator(log hclog.Logger, alloc *structs.Allocation, config 
 
 	switch {
 	case netMode == "bridge":
-		c, err := newBridgeNetworkConfigurator(log, config.BridgeNetworkName, config.BridgeNetworkAllocSubnet, config.CNIPath, ignorePortMappingHostIP)
+		c, err := newBridgeNetworkConfigurator(log, config.BridgeNetworkName, config.BridgeNetworkAllocSubnet, config.BridgeNetworkHairpinMode, config.CNIPath, ignorePortMappingHostIP)
 		if err != nil {
 			return nil, err
 		}
--- a/client/allocrunner/networking_bridge_linux.go
+++ b/client/allocrunner/networking_bridge_linux.go
@@ -1,8 +1,10 @@
 package allocrunner
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"text/template"
 
 	"github.com/coreos/go-iptables/iptables"
 	hclog "github.com/hashicorp/go-hclog"
@@ -35,14 +37,16 @@ type bridgeNetworkConfigurator struct {
 	cni         *cniNetworkConfigurator
 	allocSubnet string
 	bridgeName  string
+	hairpinMode bool
 
 	logger hclog.Logger
 }
 
-func newBridgeNetworkConfigurator(log hclog.Logger, bridgeName, ipRange, cniPath string, ignorePortMappingHostIP bool) (*bridgeNetworkConfigurator, error) {
+func newBridgeNetworkConfigurator(log hclog.Logger, bridgeName, ipRange string, hairpinMode bool, cniPath string, ignorePortMappingHostIP bool) (*bridgeNetworkConfigurator, error) {
 	b := &bridgeNetworkConfigurator{
 		bridgeName:  bridgeName,
 		allocSubnet: ipRange,
+		hairpinMode: hairpinMode,
 		logger:      log,
 	}
 
@@ -54,7 +58,7 @@ func newBridgeNetworkConfigurator(log hclog.Logger, bridgeName, ipRange, cniPath
 		b.allocSubnet = defaultNomadAllocSubnet
 	}
 
-	c, err := newCNINetworkConfiguratorWithConf(log, cniPath, bridgeNetworkAllocIfPrefix, ignorePortMappingHostIP, buildNomadBridgeNetConfig(b.bridgeName, b.allocSubnet))
+	c, err := newCNINetworkConfiguratorWithConf(log, cniPath, bridgeNetworkAllocIfPrefix, ignorePortMappingHostIP, buildNomadBridgeNetConfig(*b))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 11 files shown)


client/allocrunner/networking_bridge_linux.go
```diff
@@ -134,8 +138,34 @@ func (b *bridgeNetworkConfigurator) Teardown(ctx context.Context, alloc *structs
...
+	tIn := templInput{
+		AllocSubnet:       b.allocSubnet,
+		BridgeName:        b.bridgeName,
+		HairpinMode:       b.hairpinMode,
+		CNIAdminChainName: cniAdminChainName,
+	}
+
+	if err != nil {
+		panic(err)
+	}
```
gulducat, line 162: I'd move this err check to right under its declaration above the templInput stuff, and maybe add a comment about why panic() is ok/desirable vs returning an error up the stack?

client/allocrunner/networking_bridge_linux_test.go
```diff
@@ -0,0 +1,45 @@
...
+			},
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			tc := tc
+			var out []byte
+			require.NotPanics(t, func() { out = buildNomadBridgeNetConfig(*tc.b) })
+			if tc.name == "bad_name" {
+				fmt.Println(string(out))
```
gulducat, line 41: I'm curious why print the output for "bad_name" ?  and not any others?

and I find myself wanting to assert more than just "doesn't panic" but on the fence about what specifically to suggest short of full-text string matching...

website/content/docs/configuration/client.mdx
```diff
@@ -147,12 +147,18 @@ client {
...
   client.
 
 - `bridge_network_subnet` `(string: "172.26.64.0/20")` - Specifies the subnet
   which the client will use to allocate IP addresses from.
 
+- `bridge_network_hairpin_mode` `(bool: false)` - Specifies if hairpin mode
+  is enabled on the network bridge created by Nomad for allocations running
+  with bridge networking mode on this client. This value is exposed to the
+  scheduler as a node attribute—`nomad.bridge.hairpin_mode`—for use in
+  constraints.
```
gulducat, line 160: as we discussed, we like a more active voice here and remove the em-dash

```suggestion
  with bridge networking mode on this client. You may use the corresponding
  node attribute `nomad.bridge.hairpin_mode` in constraints.
```

client/allocrunner/networking_bridge_linux_test.go
angrycub: The bad_name test was to see if JSON stuffing was a concern. While it appears to be, it will cause the [`generateAdminChainRule` function](https://github.com/hashicorp/nomad/blob/778ab0fcd6d8f76ce8a03ed7cb37bea4bc0af47c/client/allocrunner/networking_bridge_linux.go#L121-L125) to break.  This PR doesn't actually introduce the JSON stuffing issue and ultimately, bridge names would benefit from some additional validation since they aren't built until they are needed the first time.

angrycub: I removed the test since it doesn't actually test whether or not the template _rendering_ breaks.  Even in the original code, we dropped in `%s` inside of quotes rather than leveraging `%q` for json safety.  That could be a nice "make it better than it was" item we could add though.

gulducat (approved):
I might like to see the value-add of extra unit test coverage to confirm we produce valid json, but this is still quite a good, so LGTM!

client/allocrunner/network_manager_linux.go
```diff
@@ -184,7 +184,7 @@ func newNetworkConfigurator(log hclog.Logger, alloc *structs.Allocation, config
 
 	switch {
 	case netMode == "bridge":
-		c, err := newBridgeNetworkConfigurator(log, config.BridgeNetworkName, config.BridgeNetworkAllocSubnet, config.CNIPath, ignorePortMappingHostIP)
+		c, err := newBridgeNetworkConfigurator(log, config.BridgeNetworkName, config.BridgeNetworkAllocSubnet, config.BridgeNetworkHairpinMode, config.CNIPath, ignorePortMappingHostIP)
```
lgfa29, line 187: The list of arguments is getting uncomfortably long, I wouldn't oppose a new config struct the be passed here. Although `bridgeNetworkConfiguratorConfig` is kind of weird 😬 

Another option would be to pass `config`, but that has the downside of not making it clear which config values are relevant.

client/allocrunner/networking_bridge_linux.go
```diff
@@ -54,7 +58,7 @@ func newBridgeNetworkConfigurator(log hclog.Logger, bridgeName, ipRange, cniPath
 		b.allocSubnet = defaultNomadAllocSubnet
 	}
 
-	c, err := newCNINetworkConfiguratorWithConf(log, cniPath, bridgeNetworkAllocIfPrefix, ignorePortMappingHostIP, buildNomadBridgeNetConfig(b.bridgeName, b.allocSubnet))
+	c, err := newCNINetworkConfiguratorWithConf(log, cniPath, bridgeNetworkAllocIfPrefix, ignorePortMappingHostIP, buildNomadBridgeNetConfig(*b))
```
lgfa29, line 61:
```suggestion
	c, err := newCNINetworkConfiguratorWithConf(log, cniPath, bridgeNetworkAllocIfPrefix, ignorePortMappingHostIP, buildNomadBridgeNetConfig(b))
```

And then you receive a pointer in `buildNomadBridgeNetConfig`.

client/allocrunner/networking_bridge_linux.go
```diff
@@ -134,8 +138,44 @@ func (b *bridgeNetworkConfigurator) Teardown(ctx context.Context, alloc *structs
...
+func buildNomadBridgeNetConfig(b bridgeNetworkConfigurator) []byte {
+	// Parse the internal template that generates the Nomad bridge interface.
+	// The template's parsablity is tested in networking_bridge_linux_test.go
+	// so any template issues should be caught by the test rather than at this
+	// point.
+	tmpl, err := template.New("cniConf").Parse(nomadCNIConfigTemplate)
+	if err != nil {
+		// Panic on error for catching issues in testing
+		panic(err)
+	}
```
lgfa29, line 150: I think you can move this to the top of file so the template is parse just once at initialization, and the use [`Must`](https://pkg.go.dev/text/template#Must) to panic, like:

```go
var nomadBridgeTmpl = template.Must(template.New("cniConf").Parse(nomadCNIConfigTemplate))
```

client/allocrunner/networking_bridge_linux.go
```diff
@@ -134,8 +138,44 @@ func (b *bridgeNetworkConfigurator) Teardown(ctx context.Context, alloc *structs
...
+	var out bytes.Buffer
+
+	// Since the input object provides all of the variable elements used in the
+	// template and out is a buffer (rather than a closable writer) the template
+	// execution should not error.
+	err = tmpl.Execute(&out, tIn)
+	if err != nil {
+		// Panic on error for catching issues in testing
+		panic(err)
+	}
```
lgfa29, line 177: The docs are not very clear in all possible error modes here, so I think it may be better to return `err` and let the allocrunner handle that? It may be a transient issue that a retry would fix.

website/content/docs/configuration/client.mdx
```diff
@@ -147,12 +147,17 @@ client {
...
+  created by Nomad for allocations running with bridge networking mode on the
   client.
 
 - `bridge_network_subnet` `(string: "172.26.64.0/20")` - Specifies the subnet
   which the client will use to allocate IP addresses from.
 
+- `bridge_network_hairpin_mode` `(bool: false)` - Specifies if hairpin mode
+  is enabled on the network bridge created by Nomad for allocations running
+  with bridge networking mode on this client. You may use the corresponding
+  node attribute `nomad.bridge.hairpin_mode` in constraints.
```
lgfa29, line 159:
```suggestion
- `bridge_network_hairpin_mode` `(bool: false)` - Specifies if hairpin mode
  is enabled on the network bridge created by Nomad for allocations running
  with bridge networking mode on this client. You may use the corresponding
  node attribute `nomad.bridge.hairpin_mode` in constraints. When hairpin mode 
  is enabled allocations are able to reach their own IP and port. Changing this
  value requires a reboot of the client host to take effect.
```

I think? 😅 

Feel free to modify any of this.

lgfa29 (approved):
Some minor suggestions, but LGTM!

website/content/docs/configuration/client.mdx
angrycub: For now, I think I'd like to omit the additional text so we can get the merge done. Then test the behavior around this on a running system, and then if needed come back and add it in a followup PR. I've added #16023 to track this

A-Helberg:
Awesome work!
What release will this be a part of? 
I have been using a private build of https://github.com/hashicorp/nomad/pull/13834 will be great to get back to upstream 😄

tgross:
Hi @A-Helberg this shipped in Nomad 1.5.0-beta.1 and will ship in the GA release.

Issue #15760, opened 2023-01-11
unable to delete docker image with multiple tags on same id


BDelacour:
### Nomad version
Nomad v1.4.3 (f464aca721d222ae9c1f3df643b3c3aaa20e2da7)

### Operating system and Environment details
Debian Bullseye (11)

### Issue

Hello !

When a job is deployed with a different image tag pointing to the same digest (two tags, one image), Docker driver is unable to delete the previous image later on.

Ex :
```
$> docker images
REPOSITORY  TAG        IMAGE ID       CREATED              SIZE
project1    0a8b653c   26b79d6079da   About a minute ago   924MB
project2    0a8b653c   13f41a2b8456   7 minutes ago        465MB
project2    530817fd   13f41a2b8456   7 minutes ago        465MB
project1    530817fd   21cccefce913   12 minutes ago       924MB
project1    7e991dbe   21cccefce913   12 minutes ago       924M
$> docker ps -a
CONTAINER ID   IMAGE               COMMAND                  CREATED          STATUS          PORTS                      NAMES
45756fcc7b18   project2:0a8b653c   "docker-entrypoint.s…"   34 minutes ago   Up 34 minutes   3000/tcp, 3000/udp         project2-3beab3d0-7124-385b-d0f2-015f0b575ba4
0016185272ec   project1:0a8b653c   "/entrypoint.sh /usr…"   34 minutes ago   Up 34 minutes   9000/tcp, 80/tcp, 80/udp   project1-ec844e41-4303-ffb6-572b-e77c5786c770
```

Nomad logs shows
```
[DEBUG] client.driver_mgr.docker: image id reference count decremented: driver=docker image_id=sha256:21cccefce913c5ead81f429c28810c3f7b487c6558a622a1bf1bbdcad7b1efad references=0
[DEBUG] client.driver_mgr.docker: unable to cleanup image, still in use: driver=docker image_id=sha256:21cccefce913c5ead81f429c28810c3f7b487c6558a622a1bf1bbdcad7b1efad
```

Image `21cccefce913` is no longer in use but because we're trying to delete using the image ID, we encounter a conflict here : https://github.com/hashicorp/nomad/blob/0af476263b5c6f12a43a069070400e1762ad17c2/drivers/docker/coordinator.go#L332-L335
But 409 only means `conflict` (https://docs.docker.com/engine/api/v1.23/#remove-an-image)

### Reproduction steps

[...]

PR #15962, opened 2023-01-30
docker: set force=true on remove image to handle images referenced by multiple tags

shoenig:
This PR changes our call of docker client `RemoveImage()` to `RemoveImageExtended` with
the `Force=true` option set. This fixes a bug where an image referenced by more than
one tag could never be garbage collected by Nomad. The Force option only applies to
[stopped containers](https://docs.docker.com/engine/api/v1.41/#tag/Image/operation/ImageDelete); it does not affect running workloads.

Fixes https://github.com/hashicorp/nomad/issues/15760

### DIFF
--- a/.changelog/15962.txt
+++ b/.changelog/15962.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+docker: Fixed a bug where images referenced by multiple tags would not be GC'd
+```
--- a/drivers/docker/coordinator.go
+++ b/drivers/docker/coordinator.go
@@ -59,7 +59,7 @@ func (p *pullFuture) set(imageID string, err error) {
 type DockerImageClient interface {
 	PullImage(opts docker.PullImageOptions, auth docker.AuthConfiguration) error
 	InspectImage(id string) (*docker.Image, error)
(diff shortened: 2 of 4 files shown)


tgross (approved):
LGTM

Looking at the [dockerCoordinator](https://github.com/hashicorp/nomad/blob/v1.4.3/drivers/docker/coordinator.go#L305) code, I suspect this will subvert the [`image_delay`](https://developer.hashicorp.com/nomad/docs/drivers/docker#image_delay) when folks have multiple tags on the same image if ref counts for those tags all go to zero within the delay window. But I also think if you care a lot about `image_delay` then you are going to care about canonicalizing which image tag you're using?

jrasell (approved):
LGTM!

shoenig:
> I suspect this will subvert the image_delay

Indeed, this swaps one bug for another, but I think for the less bad one.

With this change there's potential for missed optimization in re-using a downloaded image. Without, images can pile up on a node and require external maintenence as described in https://github.com/hashicorp/nomad/issues/15760. I added a note to the `image_delay` docs noting the caveat with referencing an image by multiple tags.

PR #15963, opened 2023-01-30
docs (typo fix): removed extra 'end' in one of the code blocks in template stanza documentation

Neutrollized:
In the example code block in [Simple Load Balancing with Nomad Services](https://developer.hashicorp.com/nomad/docs/job-specification/template#simple-load-balancing-with-nomad-services), there is an extra `{{- end }}` which will cause that snippet of code to produce an error


angrycub (approved):
Thanks

PR #15966, opened 2023-01-30
docs: Add info about variable item key name restrictions

angrycub:
While you can use any string value for a variable Item's key name using characters that are outside of the set [unicode.Letter, unicode.Number, `_`] will require the `index` function for direct access.

[[Preview 🔎]](https://nomad-9zsrqb18n-hashicorp.vercel.app/nomad/docs/commands/var/put#restrictions)


tgross (approved):
LGTM

PR #15969, opened 2023-01-30
15154/alloc redirect

### LINKED ISSUE #15154: Redirect with error if navigating to a missing alloc
### Proposal

When clicking on a link to an allocation that no longer exists, the web UI just shows the gray screen with "Go to Jobs" and "Go to Clients".

Instead of going to this page, which removed the user from their current context, we could redirect back to the page that they were on and show an error alerting them that the alloc doesn't exist.

ChaiWithJai:
Resolves #15154


philrenaud (approved):
LGTM! Triggered a percy diff for some reason. ¯\_(ツ)_/¯

PR #15984, opened 2023-01-31
doc: specifiy the default output is of json format for `nomad quota inspect` command

dttung2905:
Hi team,

This PR will add the `-json` flag under Inspect options for `nomad quota inspect` cli command.
The command is under enterprise offering so its abit challenging to test locally. If you know of a way to do so, please let me know too!
Also, if you want to create a separate issue to track the addition of `-json` flag for various commands, I can create it as well

### DIFF
--- a/website/content/docs/commands/quota/inspect.mdx
+++ b/website/content/docs/commands/quota/inspect.mdx
@@ -9,7 +9,7 @@ description: >
 # Command: quota inspect
 
 The `quota inspect` command is used to view raw information about a particular
-quota.
+quota. The default output is in JSON format
 
 ~> Quota commands are new in Nomad 0.7 and are only available with Nomad
 Enterprise.
@@ -29,7 +29,6 @@ capability and access to any namespaces that the quota is applied to.
 
 ## Inspect Options
 
-- `-json` : Output quota information in its JSON format.
 - `-t`    : Format and display the quota using a Go template.
 
 ## Examples
--- a/website/content/docs/commands/quota/inspect.mdx
+++ b/website/content/docs/commands/quota/inspect.mdx
@@ -9,7 +9,7 @@ description: >
 # Command: quota inspect
 
 The `quota inspect` command is used to view raw information about a particular
-quota. The default output is in JSON format
+quota. The default output is in JSON format.
 
 ~> Quota commands are new in Nomad 0.7 and are only available with Nomad
 Enterprise.
--- a/website/content/docs/commands/quota/inspect.mdx
+++ b/website/content/docs/commands/quota/inspect.mdx
@@ -29,7 +29,7 @@ capability and access to any namespaces that the quota is applied to.
 
 ## Inspect Options
 
-- `-t`    : Format and display the quota using a Go template.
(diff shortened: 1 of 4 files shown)


dttung2905:
@jrasell if you could help reviewing this PR and provide insight on the issues I highlighted above 🙏

jrasell (commented):
Hi @dttung2905 and thanks for raising this PR. I took a closer look into the quota inspect command (I've not previously used it) and it looks like JSON output is the current default unless you specify and Go template to use via the `-t` flag. This change therefore means the command requires a flag, whereas ideally flags should be optional.

An approach could be to have the `-json` flag default to true, and any supplied `-t` overrides this and results in the template being used. I don't believe this is a better approach to keeping the command the way it is though, where JSON is the default output. Apologies I didn't see this before you raised the PR. Please let me know your thoughts on this.

dttung2905:
> Hi @\dttung2905 and thanks for raising this PR. I took a closer look into the quota inspect command (I've not previously used it) and it looks like JSON output is the current default unless you specify and Go template to use via the `-t` flag. This change therefore means the command requires a flag, whereas ideally flags should be optional.
> 
> An approach could be to have the `-json` flag default to true, and any supplied `-t` overrides this and results in the template being used. I don't believe this is a better approach to keeping the command the way it is though, where JSON is the default output. Apologies I didn't see this before you raised the PR. Please let me know your thoughts on this.

Hi @jrasell , thanks for confirming this behavior. I had my suspicion when looking at the example result in the [official doc](https://developer.hashicorp.com/nomad/docs/commands/quota/inspect) where the format is already in JSON :laughing: 

Another way I could think of is below. However, since the output is deeply nested, formatting the output data in the second last line will be complicated :point_down:  ( I could not test this idea out because this is Enterprise feature though ) 

```
	if json || len(tmpl) > 0 {
		out, err := Format(json, tmpl, data)
		if err != nil {
			c.Ui.Error(err.Error())
			return 1
		}

		c.Ui.Output(out)
		return 0
	}

	c.Ui.Output(aNewFunctionToFormatData(data))
	return 0
```
I guess what we need now is just extra doc (under this [section](https://github.com/hashicorp/nomad/blob/main/website/content/docs/commands/quota/inspect.mdx#usage) for example) to clarify the default JSON output behavior? Wdyt?

dttung2905:
Hi @jrasell, I have reverted all the changes and go with the extra documentation approach. Could you help to review it again ? :pray:

website/content/docs/commands/quota/inspect.mdx
```diff
@@ -9,7 +9,7 @@ description: >
 # Command: quota inspect
 
 The `quota inspect` command is used to view raw information about a particular
-quota.
+quota. The default output is in JSON format
```
jrasell, line 12:
```suggestion
quota. The default output is in JSON format.
```

jrasell (changes requested):
Hi @dttung2905 and thanks for making the changes here. I have suggested a tiny grammar change, then I will get this merged.

jrasell (approved):
LGTM, thanks @dttung2905!

PR #15988, opened 2023-01-31
Backport of [15090] Ensure no leakage of evaluations for batch jobs. into release/1.3.x

hc-github-team-nomad-core:
## Backport

This PR is auto-generated from #15097 to be assessed for backporting due to the inclusion of the label backport/1.3.x.

---

Prior to https://github.com/hashicorp/nomad/commit/2409f72d0d42a3225d59347d2604f005bc132636 the code compared the modification index of a job to itself. Afterwards, the code compared the creation index of the job to itself. In either case there should never be a case of re-parenting of allocs causing the evaluation to trivially always result in `false`, which leads to memory leakage as reported here:

https://github.com/hashicorp/nomad/issues/15090
https://github.com/hashicorp/nomad/issues/14842
https://github.com/hashicorp/nomad/issues/4532

The comments within the code and tests are self-contradictory. Some state that evals of batch jobs should never be GCed others claim that they should be GCed if a new job version is created (as determined by comparing the indexes). I believe that the latter is the expected state.

Thus, we compare the creation index of the allocation with the modification index of the job to determine whether or not an alloc belongs to the current job version or not. 

This pull request specifically:

1. Removes references to `alloc.Job` which is unsafe given the deserialization of allocs from memDB which may include invalid pointers
2. Fixes the logical unity described above (compare creation with modification and not creation with creation)
3. Fixes the test (the test breaks the assumption of alloc reparenting, thus testing the mock rather than production code)
4. Adds more tests

Any and all feedback welcome.

For more details, please see the issues linked. Specifically https://github.com/hashicorp/nomad/issues/15090 describes the issue in detail.

EDIT:
I am assuming here that an update to the job results in an eval and consequently allocs for the job. If this is not correct than we must take into account the ordering of evals/allocs which has not been done prior.


tgross:
Fixed the backport by bringing the `getThreshold` helper back to 1.3.x

PR #15995, opened 2023-01-31
e2e: remove unused consulacls directory

shoenig:
This pile was deprecated when we starting using HCP Consul for e2e
instead of standing up our own cluster and managing Consul ACLs at test
runtime.


tgross (approved):
Red diffs are the best diffs!

PR #15997, opened 2023-01-31
Change `job init` default to example`.nomad.hcl` and recommend in docs

### LINKED ISSUE #13350: Change job spec name convention to `*.nomad.hcl`
## Background

Nomad job spec files currently follow a documented convention of `*.nomad`. This is currently "just" a convention as users can name these files any way they wish since Nomad expects full file path when submitting (`nomad job run <FILE_PATH>`).

This may become a more strictly-enforced convention in the context of https://github.com/hashicorp/nomad/issues/7954

The motivation here is to:

1. Align naming convention with Packer, which also uses HCL2 (`*.pkr.hcl` & `*.pkrvars.hcl`)
2. Provide better "out of the box" editor experience by allowing editors to register files as `*.hcl` and highlight them accordingly as HCL (i.e. editors/grammars don't need an extra matching rule for Nomad). See https://github.com/hashicorp/vscode-hcl/issues/116
3. Pave the path towards some future dedicated editor plugins/extensions which can still claim `*.nomad.hcl` - that's especially important for VS Code, which exhibits some unfortunate behaviours when multiple extensions claim the same files, see https://github.com/hashicorp/vscode-terraform/issues/692

## Proposal

Change all occurrences of `*.nomad` to `*.nomad.hcl`

 - [ ] https://www.nomadproject.io
 - [ ] https://learn.hashicorp.com/nomad
 - [ ] Update the file generated by `nomad job init`

## Notes

It _not_ proposed to set any convention for [Nomad Pack files](https://github.com/hashicorp/nomad-pack) here _yet_, but if there was a convention, it should be _different_ from Nomad job spec files. This is to allow editors which need to provide anything more than just highlighting to distinguish these files and provide relevant IntelliSense for each file type.

For the same reason, it is _not_ proposed to set any conventions for [Levant](https://github.com/hashicorp/levant) but if there was a convention I imagine it would be more like `*.nomad.yaml`, i.e. the templating engine being more important/dominant component of the name, which editors should use for highlighting. The truth is though that most editors today struggle with highlighting files with embedded languages, such as templates like these and there are no proven ways in that area.

gulducat:
Historically we have recommended the file extension `.nomad` for HCL job specification files.  This results in downstream issues like ~"Please support '.nomad' files in the HCL VS Code extension" where one solution is to append `.hcl` to the filename.

This formalizes that recommendation by:
 * `nomad job init` now produces `example.nomad.hcl` by default:
 ```shell-session
$ nomad job init
Example job file written to example.nomad.hcl
 ```
 * references to that and similar filenames are updated in website docs

This **does not**
 * stop `fmt` from also formatting `.nomad` files
 * change all our tests and such that use various `*.nomad` files and "files"
 * update docs that reference files outside of this repo

Closes #13350


tgross (approved):
LGTM!

We probably should follow up with a pass over the Tutorials repo as well.

PR #15998, opened 2023-01-31
changelog: fix entries for #15522 and #15819

lgfa29:
I noticed these while previewing the changelog from `main`.

### DIFF
--- a/.changelog/15522.txt
+++ b/.changelog/15522.txt
@@ -1,4 +1,7 @@
 ```release-note:improvement
 server: Added raft snapshot arguments to server config
+```
+
+```release-note:improvement
 server: Certain raft configuration elements can now be reloaded without restarting the server
 ```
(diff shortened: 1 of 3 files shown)


jrasell (approved):
LGTM!

shoenig (approved):
We could probably get the regex in `tools/cl-entry` to exclude `"` ... @lgfa29

lgfa29:
> We could probably get the regex in `tools/cl-entry` to exclude `"` ... @lgfa29

I fiddle around with this, but I don't think it's actually possible without false positives. We would need to detect if the entire changelog text is enclosed in quotes, but something like `^[^"][a-z0-9/\s]+: .+[^"]$` prevents ending in quotes.

Maybe just adding a caret to require the input to start with a letter or number like `^[a-z0-9/\s]+: .+` would be enough for now?

shoenig:
We could also just do the blunt `strings.HasPrefix/Suffix`

lgfa29:
But that removes quotes that are valid, like:

```
bug: Fix config "my_config"
```

Becomes this when trimmed:

```
bug: Fix config "my_config
```

lgfa29:
I'm going with just the `^` for now. If we get more errors like this we can adjust it more. Thanks for the help!

PR #15999, opened 2023-02-01
acl: fix a bug in token creation when parsing expiration TTLs. 

### LINKED ISSUE #15960: ACL Token ExpirationTTL is broken or not documented properly
### Nomad version
`Nomad v1.4.3 (f464aca721d222ae9c1f3df643b3c3aaa20e2da7)`

### Issue
According to the [documentation](https://developer.hashicorp.com/nomad/api-docs/acl/tokens#create-token):

**Create Token** `POST /acl/token`

Parameters:
 - [ExpirationTTL](https://developer.hashicorp.com/nomad/api-docs/acl/tokens#expirationttl) (duration: 0s) - This is a convenience field and if set will initialize the ExpirationTime field to a value of CreateTime + ExpirationTTL

As described above, one would assume that they can use a human readable string such as `"1h"` or `"30m"`. However, this is not the case. Seems like the nomad API does not properly unmarshal this field into `time.Duration`. As such, **only an integer representing nano seconds** seems to be an acceptable value for this field, but that is not documented. Also it is not documented that there is an acceptable range: `between 1m0s and 24h0m0s` according to the API responses I am receiving when using an integer.

### Reproduction steps
wont accept friendly string (despite being documented as a "convenience field" ;)
```
< POST /v1/acl/token HTTP/1.1
< Host: 127.0.0.1:4646
< User-Agent: Ansible Nomad API Module
< Accept-Encoding: gzip, deflate, br
< Accept: application/json
< Connection: keep-alive
< Content-Type: application/json
< X-Nomad-Token: <REDACTED>
< Content-Length: 98
< 
< {"Name": "gbolo", "Type": "client", "Policies": ["gbolo"], "Global": false, "ExpirationTTL": "1h"}

> HTTP/1.1 400 Bad Request
> Content-Encoding: gzip
> Vary: Accept-Encoding
> Date: Mon, 30 Jan 2023 16:58:03 GMT
> Content-Length: 109
> 
json: cannot unmarshal string into Go struct field ACLToken.ExpirationTTL of type time.Duration
```

apparently there are also undocumented min/max values for this field:
```
< POST /v1/acl/token HTTP/1.1
< Host: 127.0.0.1:4646
< User-Agent: Ansible Nomad API Module
< Accept-Encoding: gzip, deflate, br
< Accept: application/json
< Connection: keep-alive
< Content-Type: application/json
< X-Nomad-Token: <REDACTED>
< Content-Length: 101
< 
< {"Name": "gbolo", "Type": "client", "Policies": ["gbolo"], "Global": false, "ExpirationTTL": 1000000}

> HTTP/1.1 400 Bad Request
> Content-Encoding: gzip
> Vary: Accept-Encoding
> Date: Mon, 30 Jan 2023 17:00:11 GMT
> Content-Length: 110
> 
token 0 invalid: 1 error occurred:
        * expiration time cannot be less than 1m0s in the future (was 1ms)
```

```
< POST /v1/acl/token HTTP/1.1
< Host: 127.0.0.1:4646
< User-Agent: Ansible Nomad API Module
< Accept-Encoding: gzip, deflate, br
< Accept: application/json
< Connection: keep-alive
< Content-Type: application/json
< X-Nomad-Token: <REDACTED>
< Content-Length: 109
< 
< {"Name": "gbolo", "Type": "client", "Policies": ["gbolo"], "Global": false, "ExpirationTTL": 100000000000000}

> HTTP/1.1 400 Bad Request
> Content-Encoding: gzip
> Vary: Accept-Encoding
> Date: Mon, 30 Jan 2023 17:01:15 GMT
> Content-Length: 118
> 
token 0 invalid: 1 error occurred:
        * expiration time cannot be more than 24h0m0s in the future (was 27h46m40s)
```

#### Expected Result
I expect that the friendly strings should work, or that it is documented that this field should be an integer that represents how many nano seconds in the future that this token will be valid for. Also please document that acceptable range.

jrasell:
The ACL token decoding was not correctly handling time duration
syntax such as "1h" which forced people to use the nanosecond
representation via the HTTP API.

The change adds an unmarshal function which allows this syntax to
be used, along with other styles correctly.

It also includes a docs clarification on the validation params used for
TTL values.

closes #15960

### DIFF
--- a/api/acl.go
+++ b/api/acl.go
@@ -545,6 +545,53 @@ type ACLTokenRoleLink struct {
 	Name string
 }
 
+// MarshalJSON implements the json.Marshaler interface and allows
+// ACLToken.ExpirationTTL to be marshaled correctly.
+func (a *ACLToken) MarshalJSON() ([]byte, error) {
+	type Alias ACLToken
+	exported := &struct {
+		ExpirationTTL string
+		*Alias
+	}{
+		ExpirationTTL: a.ExpirationTTL.String(),
+		Alias:         (*Alias)(a),
+	}
+	if a.ExpirationTTL == 0 {
+		exported.ExpirationTTL = ""
+	}
+	return json.Marshal(exported)
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface and allows
+// ACLToken.ExpirationTTL to be unmarshalled correctly.
+func (a *ACLToken) UnmarshalJSON(data []byte) (err error) {
+	type Alias ACLToken
+	aux := &struct {
+		ExpirationTTL interface{}
+		*Alias
+	}{
+		Alias: (*Alias)(a),
+	}
+
+	if err = json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+	if aux.ExpirationTTL != nil {
+		switch v := aux.ExpirationTTL.(type) {
+		case string:
+			if v != "" {
+				if a.ExpirationTTL, err = time.ParseDuration(v); err != nil {
+					return err
+				}
+			}
+		case float64:
+			a.ExpirationTTL = time.Duration(v)
+		}
+
+	}
+	return nil
+}
+
 type ACLTokenListStub struct {
 	AccessorID string
 	Name       string
--- a/api/acl.go
+++ b/api/acl.go
@@ -567,7 +567,7 @@ func (a *ACLToken) MarshalJSON() ([]byte, error) {
 func (a *ACLToken) UnmarshalJSON(data []byte) (err error) {
 	type Alias ACLToken
 	aux := &struct {
-		ExpirationTTL interface{}
+		ExpirationTTL any
 		*Alias
 	}{
 		Alias: (*Alias)(a),
--- a/nomad/structs/acl.go
+++ b/nomad/structs/acl.go
@@ -345,6 +345,57 @@ func (a *ACLToken) HasRoles(roleIDs []string) bool {
 	return true
 }
 
+// MarshalJSON implements the json.Marshaler interface and allows
+// ACLToken.ExpirationTTL to be marshaled correctly.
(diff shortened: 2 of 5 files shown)


pkazmierczak (approved):
LGTM!

nomad/structs/acl.go
```diff
@@ -345,6 +345,38 @@ func (a *ACLToken) HasRoles(roleIDs []string) bool {
 	return true
 }
 
+// UnmarshalJSON implements the json.Unmarshaler interface and allows
```
tgross, line 348: The code here looks fine but in the other cases we have to support this kind of thing we do it at the `api` package layer. Ex. `api/acl.go` has this method for `ACLAuthMethod`. Should we do this parsing at the "boundary" in the API?

jrasell: Ah yes totally; I am not sure why I didn't use `ACLAuthMethod` as a basis considering I wrote that :|

I've force pushed the changes which includes `MarshalJSON` implementations and a slight change to account for parsing at the boundary which showed up during the bootstrap command. Thanks for spotting!

tgross (approved):

api/acl.go
```diff
@@ -545,6 +545,53 @@ type ACLTokenRoleLink struct {
...
+	}
+	return json.Marshal(exported)
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface and allows
+// ACLToken.ExpirationTTL to be unmarshalled correctly.
+func (a *ACLToken) UnmarshalJSON(data []byte) (err error) {
+	type Alias ACLToken
+	aux := &struct {
+		ExpirationTTL interface{}
```
shoenig, line 570:
```suggestion
		ExpirationTTL any
```

gotta be like the cool kids

shoenig (approved):
LGTM!

PR #16009, opened 2023-02-01
scheduler: keep track of placement metrics when in preempt mode

shoenig:
- scheduler: keep track of failure reasons when in preempt mode
- scheduler: drive by cleanup of some scheduler bits


shoenig:
wrong way of doing this

Issue #15235, opened 2022-11-14
nil pointer dereference in deployment monitor


mxplusb:
### Nomad version

`Nomad v1.4.2`

### Operating system and Environment details

<img width="392" alt="image" src="https://user-images.githubusercontent.com/7537841/201113190-a58d033c-75c4-466c-846c-c592a8a2ce40.png">

### Issue

There is a nil pointer dereference when running `nomad job run`.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x8 pc=0x1042cf598]

goroutine 1 [running]:
github.com/hashicorp/nomad/command.(*DeploymentStatusCommand).ttyMonitor(0x14000a9f928, 0x1400083ec40, {0x140009fdad0, 0x24}, 0x0, 0x0)
	github.com/hashicorp/nomad/command/deployment_status.go:329 +0x1e48
github.com/hashicorp/nomad/command.(*DeploymentStatusCommand).monitor(0x1400083ee00?, 0x14000c26040?, {0x140009fdad0, 0x24}, 0x2?, 0x0?)
	github.com/hashicorp/nomad/command/deployment_status.go:185 +0x84
github.com/hashicorp/nomad/command.(*monitor).monitor(0x14000bc2690, {0x140009fda10, 0x24})
	github.com/hashicorp/nomad/command/monitor.go:302 +0xbc8
github.com/hashicorp/nomad/command.(*JobRunCommand).Run(0x1400052aa00, {0x1400004e070, 0x1, 0x1})
	github.com/hashicorp/nomad/command/job_run.go:376 +0x1074
github.com/mitchellh/cli.(*CLI).Run(0x140000e3540)
	github.com/mitchellh/cli@v1.1.4/cli.go:262 +0x4a8
main.RunCustom({0x1400004e050?, 0x3, 0x3})
	github.com/hashicorp/nomad/main.go:117 +0x350
main.Run(...)
	github.com/hashicorp/nomad/main.go:87
main.main()
	github.com/hashicorp/nomad/main.go:83 +0x50
```

### Reproduction steps

Here is my `waypoint.nomad` job, it's using a [GCP CSI](https://gist.github.com/angrycub/54ccec8215b4c5f405f95b98f1faa3b5) volume on the backend.

```hcl
job "waypoint" {
  region      = "americas"
  datacenters = ["us-central1"]
  type        = "service"

  group "server" {
    count = 1

    update {
[...]

PR #16011, opened 2023-02-01
cli: Fix a panic in `deployment status` when scheduling is slow

tgross:
Fixes https://github.com/hashicorp/nomad/issues/15235

If a deployment fails, the `deployment status` command can get a nil deployment when it checks for a rollback deployment if there isn't one (or at least not one at the time of the query). Fix the panic.

Note that I'm intentionally punting on solving the old TODO here, as it'd just take too long for me to page all this code in. Dropping a quick panic fix was a small win between other tasks :grin:

### DIFF
--- a/.changelog/16011.txt
+++ b/.changelog/16011.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cli: Fixed a panic in `deployment status` when rollback deployments are slow to appear
+```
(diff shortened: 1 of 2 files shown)


angrycub:
Does this need to happen at line 444 also?

tgross:
> Does this need to happen at line 444 also?

Good catch. Fixed! (Plus changelog!)

shoenig (approved):
LGTM

Issue #12513, opened 2022-04-08
Jobspec using validation with missing constraint causes panic


shoenig:
While looking over https://github.com/hashicorp/nomad/pull/12508 I had a typo in my own example and triggered a panic of the CLI. A missing `condition` in the `validation` block of a `variable` block gets through jobspec validation itself, and causes a NPE. 

```hcl
variable "image_id" {
  type          = string
  description   = "Check on image names."

  validation {
    # condition  =   substr(var.image_id, 0, 4) == "foo-" 
    error_message = "Must use a foo image."
  }
}

job "example" {
  datacenters = ["dc1"]

  group "cache" {
    network {
      port "db" {
        to = 6379
      }
    }

    task "redis" {
      driver = "docker"
      config {
        image = "${var.image_id}"
        ports = ["db"]
      }
    }
  }
}
```

```
➜ nomad job run -var 'image_id=foo-' example.nomad
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x1be87d9]

goroutine 1 [running]:
github.com/hashicorp/nomad/jobspec2.(*Variable).validateValue(0xc000909ea0, {{0x25b2ed8, 0x3}, {{{0x2b067c8, 0xc0004bcd69}}, {0x1fa0820, 0xc000b4d1b0}}, {0x7f03143459f0, 0xc000c304e0}})
	github.com/hashicorp/nomad/jobspec2/types.variables.go:99 +0x359
github.com/hashicorp/nomad/jobspec2.(*Variable).Value(0x120)
	github.com/hashicorp/nomad/jobspec2/types.variables.go:159 +0x145
github.com/hashicorp/nomad/jobspec2.Variables.Values(0xc0003a1730)
	github.com/hashicorp/nomad/jobspec2/types.variables.go:176 +0xd9
github.com/hashicorp/nomad/jobspec2.(*jobConfig).decodeBody(0xc0003a1730, {0x2b054c0, 0xc0002d3550})
	github.com/hashicorp/nomad/jobspec2/types.config.go:68 +0x2da
github.com/hashicorp/nomad/jobspec2.decode(0xc0003a1730)
	github.com/hashicorp/nomad/jobspec2/parse.go:105 +0xc9
github.com/hashicorp/nomad/jobspec2.ParseWithConfig(0xc0002d2f20)
	github.com/hashicorp/nomad/jobspec2/parse.go:44 +0xe5
[...]

PR #16018, opened 2023-02-01
job parsing: fix panic when variable validation is missing condition

tgross:
Fixes https://github.com/hashicorp/nomad/issues/12513

---

Test using the example jobspec from #12513:

```
$ nomad job run -var 'image_id=foo-' example.nomad
Error getting job struct: Error parsing job file from example.nomad:
example.nomad:5,14-14: Missing required argument; The argument "condition" is required, but no definition was found.
example.nomad:5,3-13: Invalid variable validation specification; validation requires a condition.
```


shoenig (approved):
LGTM!

Issue #1851, opened 2016-10-24
NOMAD_ALLOC_INDEX is 0 on all nodes when using system scheduler.


siffiejoe:
### Nomad version

Nomad v0.4.0
### Operating system and Environment details

```
Linux host01 3.13.0-100-generic #147-Ubuntu SMP Tue Oct 18 16:48:51 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```
### Issue

Two different nodes get the same allocation index (0) when using the system scheduler.
### Reproduction steps

Create a nomad cluster of at least two nodes, copy the following script onto all nodes as `/tmp/x.sh` and run the job script below.

``` sh
#!/bin/sh

while true; do
  echo "NOMAD_ALLOC_INDEX: $NOMAD_ALLOC_INDEX"
  sleep 1
done
```

All (both in my case) print the same allocation index (i.e. 0).
### Job file (if appropriate)

```
job "xxx" {
    type = "system"

    datacenters = ["dc1"]

    task "x" {
    driver = "raw_exec"
    config {
        command = "/tmp/x.sh"
        args = []
    }
    resources {
        memory = 10
    }
    kill_timeout = 10s
    }
}
```

dadgar:
Hey,

This is currently the desired behavior. The allocation index is used to indicate the instance out of a desired count size. System jobs do not have an explicit count but are based more on the size of a the targeted cluster, making the index less informative.

I'm curious what you have in mind for the index. Further there is an associated cost with making it a monotonically increasing counter (need to store additional meta data) and I am not sure of the benefit.

siffiejoe:
I was trying to use the allocation index to get separate host names for each instance via consul. I need this because the service needs host names/addresses of all nodes that are going to be part of the cluster in advance (zookeeper is the culprit). I also want to wait until all instances are up before I start dependent services, and checking for `zk-N.zookeeper.service.consul` is an easy way to do that.

jippi:
I've been helping @smuthali with zookeeper as well, and he got something working out of that, perhaps drop by the gitter channel and give him a poke? :)

dadgar:
@siffiejoe I think you could accomplish that by using `node.unique.id` or `node.unique.name`.

I am going to close this issue as it is desired behavior. Hope you get the help you need from @jippi and @smuthali

Best,
Alex

jippi:
a hack is to have a docker entrypoint that echo ${NOMAD_ALLOC_INDEX} into the required file :)

PR #16030, opened 2023-02-02
System and sysbatch jobs always have zero index

tgross:
System jobs do not have unique allocation Names because the index is intended to indicated the instance out of a desired count size. Because system jobs do not have an explicit count but the results are based on the targeted nodes, the index is less informative and this was intentionally omitted from the original design. (See https://github.com/hashicorp/nomad/issues/1851). This behavior has confused a lot of users and Nomad engineers, including me!

This changeset includes:
* Updating the documentation to make this existing behavior clear to users.
* Added a test condition to the scheduler tests so that the behavior is documented in code.
* Fixed a bug where `per_alloc` volumes should not have been allowed for system and sysbatch jobs.

Note that changing this behavior rather than merely documenting it is likely to be a huge lift and not practical in the short term. We'll need to backport this but I'm going to open that as a separate PR against `release/1.4.x` and backport from there because the current `per_alloc` code includes a new-in-1.5.0 feature.


scheduler/scheduler_system_test.go
```diff
@@ -66,6 +66,17 @@ func TestSystemSched_JobRegister(t *testing.T) {
 	// Ensure all allocations placed
 	require.Len(t, out, 10)
 
+	// Note that all system allocations have the same name derived from Job.Name
+	allocNames := []string{}
+	for _, alloc := range out {
+		allocNames = append(allocNames, alloc.Name)
+	}
```
shoenig, line 73: one liner with `helper.ConvertSlice`: https://github.com/hashicorp/nomad/blob/main/helper/funcs.go#L401

shoenig (approved):
LGTM! just the code golf suggestion

scheduler/scheduler_system_test.go
tgross: Cool, will do.

tgross:
I'm going to pick up these suggestions @shoenig but I'm also realizing that the code that creates the names is _incredibly_ misleading, because it's never called for service jobs but suggests that it is. I'm going to fix this up here as well.

lgfa29 (approved):

tgross:
I'm going to go ahead with merging this without touching that test code after all. This PR covers the problem with the alloc index. But when I started messing with the test code that uses a service job to test the system diffs (oops!) it breaks in weird ways that seem unrelated to this bug entirely.

PR #16044, opened 2023-02-02
Backport #16030 System and sysbatch jobs always have zero index to 1.2.x

tgross:
This hand-backports https://github.com/hashicorp/nomad/pull/16030 to 1.2.x, because the original PR touched some code that has a minor new feature for 1.5.0 (per_alloc host volumes) and because there are some significant conflicts over the env vars page, for which the new table syntax was never backported.

---

Service jobs should have unique allocation Names, derived from the Job.ID. System jobs do not have unique allocation Names because the index is intended to indicated the instance out of a desired count size. Because system jobs do not have an explicit count but the results are based on the targeted nodes, the index is less informative and this was intentionally omitted from the original design.

Update docs to make it clear that NOMAD_ALLOC_INDEX is always zero for system/sysbatch jobs

Validate that `volume.per_alloc` is incompatible with system/sysbatch jobs. System and sysbatch jobs always have a `NOMAD_ALLOC_INDEX` of 0. So interpolation via `per_alloc` will not work as soon as there's more than one allocation placed. Validate against this on job submission.

### DIFF
--- a/scheduler/scheduler_system_test.go
+++ b/scheduler/scheduler_system_test.go
@@ -9,10 +9,12 @@ import (
 
 	memdb "github.com/hashicorp/go-memdb"
 	"github.com/hashicorp/nomad/ci"
+	"github.com/hashicorp/nomad/helper"
 	"github.com/hashicorp/nomad/helper/pointer"
 	"github.com/hashicorp/nomad/helper/uuid"
 	"github.com/hashicorp/nomad/nomad/mock"
 	"github.com/hashicorp/nomad/nomad/structs"
+	"github.com/shoenig/test/must"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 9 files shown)


scheduler/scheduler_system_test.go
```diff
@@ -65,6 +66,15 @@ func TestSystemSched_JobRegister(t *testing.T) {
...
 	require.Len(t, out, 10)
 
+	// Note that all system allocations have the same name derived from Job.Name
+	allocNames := helper.ConvertSlice(out,
+		func(alloc *structs.Allocation) string { return alloc.Name })
+	expectAllocNames := []string{}
+	for i := 0; i < 10; i++ {
+		expectAllocNames = append(expectAllocNames, fmt.Sprintf("%s.web[0]", job.Name))
+	}
+	must.SliceContainsAll(t, expectAllocNames, allocNames)
```
lgfa29, line 76: Missing `must` import:
https://app.circleci.com/pipelines/github/hashicorp/nomad/38593/workflows/542a7cb9-06ba-4034-a075-0b09d7b96ff2/jobs/410325?invite=true#step-108-227

lgfa29 (approved):
Just need to fix the missing import 👍

PR #16050, opened 2023-02-03
Add a parameter to the template config render_templates_once

jeffreybolle:
This changes the re-render behaviour of nomad templates so that they do not get re-rendered when the values associated with consul keys used in the templates change.

### DIFF
--- a/client/allocrunner/taskrunner/template/template.go
+++ b/client/allocrunner/taskrunner/template/template.go
@@ -932,6 +932,8 @@ func newRunnerConfig(config *TaskTemplateManagerConfig,
 		}
 	}
 
+	conf.Once = cc.TemplateConfig.RenderTemplatesOnce
+
 	conf.Finalize()
 	return conf, nil
 }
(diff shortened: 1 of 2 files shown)


jrasell (changes requested):
Hi @jeffreybolle and thanks for raising this PR.

This approach being on a per client basis raises concern that in heterogeneous clusters, allocations which require templates to be re-rendered on change, could be placed on clients which have this disabled. In order to avoid this, operators would need to ensure jobs have custom constraints which feels like a poor UX and unexpected behaviour.

I think therefore to have this feature added, the control of this would need to be performed via a parameter within the template block of the job specification. It would be great to also include some testing in any followup work. Please let me know your thoughts or if you have any questions.

jeffreybolle:
> Hi @jeffreybolle and thanks for raising this PR.
> 
> This approach being on a per client basis raises concern that in heterogeneous clusters, allocations which require templates to be re-rendered on change, could be placed on clients which have this disabled. In order to avoid this, operators would need to ensure jobs have custom constraints which feels like a poor UX and unexpected behaviour.
> 
> I think therefore to have this feature added, the control of this would need to be performed via a parameter within the template block of the job specification. It would be great to also include some testing in any followup work. Please let me know your thoughts or if you have any questions.

Thanks for the feedback, your reasoning makes sense.  I will update the pull request when I get a chance.

jrasell:
Hi @jeffreybolle, I am going to close this PR. If you managed to get time to make the suggested changes, then please feel free to reopen or create a new PR.

PR #16051, opened 2023-02-03
scheduler: move utils into files specific to their scheduler type

tgross:
Many of the functions in the `utils.go` file are specific to a particular scheduler, and very few of them have guards (or even names) that help avoid misuse with features specific to a given scheduler type. Move these functions (and their tests) into files specific to their scheduler type without any functionality changes to make it clear which bits go with what.

(I normally try to avoid large code moves like this, but that's why this has no functionality changes.)


tgross:
(I'm going to backport this by hand, because I don't trust backport assistant to safely backport without accidentally adding features to old versions in the process.)

scheduler/system_util.go
```diff
@@ -0,0 +1,348 @@
...
+	if job.Stopped() {
+		return out
+	}
+
+	for _, tg := range job.TaskGroups {
+		for i := 0; i < tg.Count; i++ {
+			name := fmt.Sprintf("%s.%s[%d]", job.Name, tg.Name, i)
+			out[name] = tg
+		}
+	}
```
tgross, line 29: Note to reviewers if this jumps out at you: this code is unchanged from the function originally in `util.go` and this is what inspired the spelunking that resulted in this PR. This is likely a bug because `tg.Count` is always 0 for system and sysbatch jobs, but I don't want to try to fix a bug and do a large refactoring in the same changeset. I'm working on that bug next.

shoenig (approved):
LGTM

PR #16063, opened 2023-02-06
docker: disable driver when running as non-root on cgroups v2 hosts

### LINKED ISSUE #15927: docker: disable driver if running on cgv2 but without root
In the case where someone might run `nomad agent -dev` (without sudo), the docker driver will be enabled, but will not actually be able to run tasks. The reason is complicated but boils down to how Nomad client manages the cgroup when cgv2 is enabled, which currently requires elevated privileges.

For a better experience than trying to run tasks that just never start, have the docker driver fingerprint the case where cgv2 is present but the driver is not running as root, and have it disable itself.

shoenig:
This PR modifies the docker driver to behave like exec when being run
as a non-root user on a host machine with cgroups v2 enabled. Because
of how cpu resources are managed by the Nomad client, the nomad agent
must be run as root to manage docker-created cgroups.

Fixes #15927

### DIFF
--- a/.changelog/7794.txt
+++ b/.changelog/7794.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+docker: disable driver when running as non-root on cgv2 hosts
+```
--- a/client/testutil/driver_compatible.go
+++ b/client/testutil/driver_compatible.go
@@ -15,6 +15,14 @@ func RequireRoot(t *testing.T) {
 	}
 }
 
+// RequireNonRoot skips tests unless:
+// - running as non-root
+func RequireNonRoot(t *testing.T) {
+	if syscall.Geteuid() == 0 {
+		t.Skip("Test requires non-root")
+	}
+}
+
 // RequireConsul skips tests unless:
 // - "consul" executable is detected on $PATH
 func RequireConsul(t *testing.T) {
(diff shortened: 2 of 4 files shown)


shoenig:
Spot check

running as non root
```
➜ ps -ef | grep -v grep | grep nomad | fields 1
shoenig
```

```
Drivers
Driver       Detected  Healthy  Message                  Time
docker       false     false    Driver must run as root  2023-02-06T11:21:06-06:00
exec         false     false    Driver must run as root  2023-02-06T11:21:06-06:00
java         false     false    Driver must run as root  2023-02-06T11:21:06-06:00
mock_driver  true      true     Healthy                  2023-02-06T11:21:06-06:00
qemu         false     false    <none>                   2023-02-06T11:21:06-06:00
raw_exec     true      true     Healthy                  2023-02-06T11:21:06-06:00
```

running as root

```
Drivers
Driver       Detected  Healthy  Message  Time
docker       true      true     Healthy  2023-02-06T11:22:29-06:00
exec         true      true     Healthy  2023-02-06T11:22:29-06:00
java         true      true     Healthy  2023-02-06T11:22:29-06:00
mock_driver  true      true     Healthy  2023-02-06T11:22:29-06:00
qemu         false     false    <none>   2023-02-06T11:22:29-06:00
raw_exec     true      true     Healthy  2023-02-06T11:22:29-06:00
```

schmichael (approved):

PR #16066, opened 2023-02-06
chore: Convert assets from bindatafs to go embeds

angrycub:
Per discussion, starting a move toward go:embed where possible.


lgfa29 (approved):
Nice! 

Since we're moving them already, how about placing the job HCL files in a folder like `command/asset/jobs`? That way if we add more `init` type commands for things, like policies, volumes etc. they would be more organized.

PR #16067, opened 2023-02-06
build: update hc-install and golangci-lint tools

shoenig:
Update golangci-lint again, and also hc-install which now has a real version.


tgross (approved):
LGTM

PR #16069, opened 2023-02-06
Add cpuset_mems to docker driver.

shishir-a412ed:
Currently, we have NUMA support for CPU pinning in Nomad docker driver using [cpuset_cpus](https://developer.hashicorp.com/nomad/docs/drivers/docker#cpuset_cpus) option. However, we don't have support for pinning the workload to a particular set of memory nodes.

Without both (`cpuset_cpus` and `cpuset_mems`) the feature is not complete, as one can pin the workload to a particular NUMA node, however the memory can still spread across two NUMA nodes.

This PR will add support for pinning the workload to a particular set of memory nodes.

Added a unit test and updated nomad docker driver docs on the website.

cpuset_mems

Signed-off-by: Shishir Mahajan <smahajan@roblox.com>

### DIFF
--- a/drivers/docker/driver_test.go
+++ b/drivers/docker/driver_test.go
@@ -1575,6 +1575,27 @@ func TestDockerDriver_CPUSetMEMs(t *testing.T) {
 	require.Equal(t, cfg.CPUSetMEMs, container.HostConfig.CPUSetMEMs)
 }
 
+func TestDockerDriver_InvalidCPUSetMEMs(t *testing.T) {
+	ci.Parallel(t)
+	testutil.DockerCompatible(t)
+	if runtime.GOOS == "windows" {
+		t.Skip("Windows does not support cpuset_mems")
+	}
+
+	task, cfg, _ := dockerTask(t)
+
+	cfg.CPUSetMEMs = "0-8"
+	require.NoError(t, task.EncodeConcreteDriverConfig(cfg))
+
+	d := dockerDriverHarness(t, nil)
+
+	_, _, err := d.StartTask(task)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "Requested memory nodes are not available")
+
+	d.DestroyTask(task.ID, true)
+}
+
 func TestDockerDriver_MemoryHardLimit(t *testing.T) {
 	ci.Parallel(t)
 	testutil.DockerCompatible(t)
--- a/drivers/docker/driver_test.go
+++ b/drivers/docker/driver_test.go
@@ -1553,6 +1553,28 @@ func TestDockerDriver_CPUSetCPUs(t *testing.T) {
 	}
 }
 
+func TestDockerDriver_CPUSetMEMs(t *testing.T) {
+	ci.Parallel(t)
+	testutil.DockerCompatible(t)
+	if runtime.GOOS == "windows" {
+		t.Skip("Windows does not support cpuset_mems")
+	}
+
+	task, cfg, _ := dockerTask(t)
+
+	cfg.CPUSetMEMs = "0"
+	require.NoError(t, task.EncodeConcreteDriverConfig(cfg))
+
+	client, d, handle, cleanup := dockerSetup(t, task, nil)
+	defer cleanup()
+	require.NoError(t, d.WaitUntilStarted(task.ID, 5*time.Second))
+
+	container, err := client.InspectContainer(handle.containerID)
+	require.NoError(t, err)
+
+	require.Equal(t, cfg.CPUSetMEMs, container.HostConfig.CPUSetMEMs)
+}
+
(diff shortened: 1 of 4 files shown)


drivers/docker/driver_test.go
```diff
@@ -1553,6 +1553,48 @@ func TestDockerDriver_CPUSetCPUs(t *testing.T) {
 	}
 }
 
+func TestDockerDriver_CPUSetMEMs(t *testing.T) {
+	ci.Parallel(t)
+	testutil.DockerCompatible(t)
+	testutil.CgroupsCompatible(t)
+
+	testCases := []struct {
```
georgeli-roblox, line 1561: Thanks! LGTM

nit:  maybe add a test case of invalid  `CPUSetMems` ?

shishir-a412ed: Added `TestDockerDriver_InvalidCPUSetMEMs`

tgross:
Hi @shishir-a412ed! As you might have seen, we shipped NUMA support in Nomad Enterprise 1.7.0. I'm going to close this PR, but thanks for kicking off our discussions around this!

neomantra:
@tgross I was just looking for this feature in the Docker Driver and was even going to add it if it wasn't there.   

I did similar pull requests with Docker itself over 10 years ago [1] and discussions in Kubernetes over 8 years ago [2].   Control over this has been in Docker for a long time and is commonly used across many metal deployments.

Nomad Enterprise 1.7.0 shipped with support for NUMA in the **Scheduler**.   Congratulations, it is not an easy problem -- such an advanced feature is worthy of an Enterprise tier.

----

The Docker Driver Config and the NUMA-Aware Scheduler operate at different levels.

Both `cpuset_cpus` and `cpuset_mems` are advanced  `docker run` knobs for controlling process placement.  When used an operator, they are **NOT using the Nomad scheduler at all** and relying upon their specific placement directives.    Like `cpuset_cpus`, the `cpuset_mems` implementation is a simple pass-through if config to driver; clearly this PR is short, simple, and harmonious with the  `cpuset_cpus` implementation.

The NUMA-Aware scheduler operates at a much higher level, understanding existing workloads and resources, dynamically and optimally placing jobs.   

The Docker Driver Configurations are only manageable (but very useful!) at small scale... the NUMA-Aware Scheduler is not Enterprise tier because it can pin jobs to a NUMA-node -- it is Enterprise-tier because it can pin jobs to the most optimal NUMA node given the all resources and constraints of the Cluster.

---

I personally combine these settings along with node-specifying `constraints`, that are all data-driven by Terraform.  But as noted, the scalability of this is limited and not dynamic like a scheduler.

Lastly, it may actually impair performance to ignore `cpuset_mems` when also using `cpuset_cpus`.  These are potential footguns for the uninitiated. [3]

The alternative I'm about to try is running `numactl` in a sidecar, or maintaining our own Nomad build with this one patch. 

[1] https://github.com/moby/moby/issues/439
[2] https://github.com/kubernetes/kubernetes/issues/10570#issuecomment-255441875
[3] https://gist.github.com/neomantra/3c9b89887d19be6fa5708bf4017c0ecd#the-foot-gun (hope to  update for Nomad, as this is what I'm doing right now  🤞 )

neomantra:
I channeled my frustration into OSS love / empowerment and started this project right after that comment -- something I've been wanting to make for a few years now.  It actually wasn't that hard, given the great Skeleton tooling.

 * https://github.com/neomantra/nomad-onload

One can use [Onload](https://github.com/Xilinx-CNS/onload) without the NICs for userspace `epoll`. or with user-space TCP/UDP acceleration via XDP.  So it is generally applicable to Nomad Clusters.

I don't personally have the workloads or scale for it -- but for sure there's something amazing to behold with kernel-bypassed Nomad Jobs running on a pool of GPUs and NICs smoothly connected over a fabric of shared memory and hardware acceleration, all optimally orchestrated for mechanical sympathy by a NUMA-aware scheduler that maintains a dynamic model of the entire cluster's CPUs, GPUs, NICs, and Memory and Job's affinities.   There's an AI wet dream just sitting here unassembled.  

Please pull on that thread for Enterprise!  Don't limit Nomad by closing this issue without merging. 🥺

tgross:
@neomantra from an architectural perspective Nomad tries to avoid having resource constraints in the task driver that can conflict with each other after placement without a corresponding scheduler component. The reason being is that if the scheduler isn't aware of those resources then it can make multiple placements to the wrong node in a short period of time and cause deployment failures. This definitely doesn't cover 100% of features of all task drivers (because of other unfortunate architectural decisions like not having the task driver config schema part of the fingerprint), but it's a general goal.

If you've got a specific proposal that you think you could make a case for given that context, we'd be happy to discuss it in a new issue.

PR #16071, opened 2023-02-06
changelog: add changelog update for vendor label for linux packaging

claire-labry:
This PR is just an update to the changelog to reflect a bug that was fixed in our [actions-packaging-linux](https://github.com/hashicorp/actions-packaging-linux) action. We received a report from a user stating that vendor labels were missing when downloading linux packages (for products that were released via CRT) We fixed this in [this PR](https://github.com/hashicorp/actions-packaging-linux/pull/14) and we added a vendor label and set the default to `HashiCorp` This will only affect future releases that go through our linux promotion pipeline -- historical releases will not have this vendor label.

### DIFF
--- a/.changelog/_14.txt
+++ b/.changelog/_14.txt
@@ -1,3 +1,0 @@
-```release-note: bug
-core: Linux packages now have vendor label and set the default label to HashiCorp. This fix is implemented for any future releases, but will not be updated for historical releases.
-```
--- a/.changelog/_14.txt
+++ b/.changelog/_14.txt
@@ -1,0 +1,3 @@
+```release-note: bug
+core: Linux packages now have vendor label and set the default label to HashiCorp. This fix is implemented for any future releases, but will not be updated for historical releases.
+```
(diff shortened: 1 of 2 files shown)


.changelog/_14.txt
```diff
@@ -0,0 +1,3 @@
+```release-note: bug
+core: Linux packages now have vendor label and set the default label to HashiCorp. This fix is implemented for any future releases, but will not be updated for historical releases.
```
jrasell, line 2: we typically use `core` to mean core Nomad functionality, so changing this would help differentiate.
```suggestion
build: Linux packages now have vendor label and set the default label to HashiCorp. This fix is implemented for any future releases, but will not be updated for historical releases.
```

jrasell (changes requested):
I just wanted to double check, that this applies to our 1.5.0-beta.1 release that went out yesterday? If so, I'll makes sure this gets into the changelog for this release train.

claire-labry:
@jrasell that's correct, I've updated the PR to reflect your feedback and I ended up changing the changelog # to reflect the #PR number here instead of the PR number in the linux-packaging action. Feel free to merge this whenever you're ready.

lgfa29 (approved):

jrasell (approved):

PR #16073, opened 2023-02-07
docs: update CLI contrib checklist

lgfa29:
I think these were the things we found today. Anything else we should add?


jrasell (approved):
LGTM!

PR #16074, opened 2023-02-07
main: remove deprecated uses of rand.Seed

shoenig:
go1.20 deprecates rand.Seed, and seeds the rand package automatically. Remove cases where we seed the random package, and cleanup the one test case where we intentionally create a known random source.

https://pkg.go.dev/math/rand#Seed
https://go.dev/doc/go1.20


jrasell (approved):
LGTM!

Checked locally also:
```
$ ag "rand.Seed" |wc -l 
       0
```

PR #16075, opened 2023-02-07
e2e: mark framework package as deprecated

shoenig:
Nothing more motivating than lots of deprecation warnings to get some code refactored.


tgross (approved):
:+1:

PR #16076, opened 2023-02-07
agent: fix agent HTTP server audit event implementation access.

jrasell:
https://github.com/hashicorp/nomad/pull/15864 changed the way in which the agent is passed to the HTTP servers by wrapping it in an interface. This removed access to the audit eventer, which is used by the HTTP server to log events for the enterprise codebase.

This change therefore explicitly passes the `event.Auditor` interface to the agent HTTP servers for use. This change will be needed along with an enterprise specific one, to slightly alter the audit calls to use the new field.

Please see the internal nightly runs to see the original error.

### DIFF
--- a/command/agent/event/event.go
+++ b/command/agent/event/event.go
@@ -6,10 +6,10 @@ import (
 
 // Auditor describes the interface that must be implemented by an eventer.
 type Auditor interface {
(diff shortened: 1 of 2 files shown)


tgross (approved):
LGTM! Thanks for picking this up!

PR #16082, opened 2023-02-07
docs: update example `license_path`

tgross:
In #13374 we updated the commented-out `license_path` in the packaged example configuration file to match the existing documentation. Although this config value was commented-out, it was reported that changing the value was confusing. Update the commented-out line to the previous value and update the documented examples to match that. This matches most of the examples for Consul/Vault licensing as well. I've double-checked the tutorials and it looks like it'd been left on the previous value there, so no additional work to be done.


shoenig (approved):
LGTM

Issue #15516, opened 2022-12-09
allow configurable range of priorities >100


tgross:
Users who want fine-grained control over job priorities combined with a large volume of jobs being submitted at once (ex. large batch workloads) have reported that having only 100 levels of priority isn't fine-grained enough. The range from 0 to 100 is arbitrary and hard-coded as a integer literal in all the various places it's used.

We can instead turn this constant into a [server configuration](https://developer.hashicorp.com/nomad/docs/configuration/server) value with a default of 100 (the default eval/job priority can be left at 50, or we could make that a configuration value as well). Because priority 0 is the lowest-priority in Nomad, this can be done without impacting backwards compatibility. It'd just be up for operators to set sensible values when they restart servers with new values.

cc @sofixa

alessio-perugini:
:wave: @tgross If you don't mind I'd like to take care of this. :nerd_face: 
I was thinking about some open points though:
1. What would it be the minimum and the maximum value of `JobDefaultPriority` and `JobMaxPriority`?
Do we want to keep the current default values as a lower bound and allow only to go upper? Or we do want to allow going also below our default current values? (0-100)
2. I guess we also want to change the default value of `CoreJobPriority = JobMaxPriority * 2` to avoid possible int overflow. 

For the second point, I can see 2 options:
- `CoreJobPriority = math.MaxInt` 
- `CoreJobPriority = JobMaxPriority+1`

tgross:
> What would be the minimum and maximum values of `JobDefaultPriority` and `JobMaxPriority`?

The goal here is to give cluster admins greater range, so the minimum value should be the current hard-coded value. `JobDefaultPriority = 50` and `JobMaxPriority = 100`.

It's a good catch to note that `CoreJobPriority` is set to the max * 2. To your point about `CoreJobPriority`, we need to ensure that `CoreJobPriority` is always greater than any user-defined value. Off the top of my head I don't see any issues with setting `CoreJobPriority` so high as to where it'd cause overflow. But maybe out of an abundance of caution we cap priority to `math.MaxInt16` during validation without changing the field value from `int` (which we need to keep for backwards compatibility of serialized values). That'll give us headroom for new features if we need it, too.

So we that would give us the following:

* `CoreJobPriority` is hard-coded to `math.MaxInt16` (32767)
* `JobMaxPriority` defaults to 100 and can be configured in the range `100 <= JobMaxPriority < math.MaxInt16`.
[...]

PR #16084, opened 2023-02-07
Allow configurable range of Job priorities

alessio-perugini:
#15516
## Summary
To allow our users to better fine-grain the job priorities, we're expanding the upper limit from `100` to `MaxInt16-1`. 

- `CoreJobPriority` is hard-coded to `math.MaxInt16` (32767)
- `JobMaxPriority` defaults to 100 and can be configured in the range `100 <= JobMaxPriority < math.MaxInt16`.
- `JobDefaultPriority` defaults to 50 and can be configured in the range `50 <= JobDefaultPriority < JobMaxPriority`
- The new configuration value is set into our [server configuration](https://developer.hashicorp.com/nomad/docs/configuration/server).

## Notes for the reviewer
This quick draft aims to gather feedback to understand if the solution is on the right path. After that, I'll add tests, docs, proper commits names, and the other missing stuff.

Bear with me, this is my first time contributing to the codebase. 🙏 I'm looking forward to your comments. :nerd_face:


api/jobs.go
```diff
@@ -907,6 +913,7 @@ type Job struct {
 	CreateIndex              *uint64
 	ModifyIndex              *uint64
 	JobModifyIndex           *uint64
+	MaxPriority              *int
```
tgross, line 916: The Job struct doesn't need a `MaxPriority`, as it only ever has a single priority -- the one the user has set or the default.

tgross (commented):
Hi @alessio-perugini! It looks like you've discovered how gross it is to wire up all the configuration so that it makes its way from the agent to the server, and how it gets merged to the default config. So yes, this looks like it's on the right track!

api/jobs.go
alessio-perugini: :thinking: I'm not sure how we can send the user defined MaxPriority in the `Job` struct definied in `nomad/structs/structs.go` without touching the one in the api/jobs.go.
My idea was to pass this information so we could update the check at line `nomad/structs/structs.go:4454`. But maybe we can move this in another part, so we don't need the MaxPriority value inside the job struct?

tgross: Apologies, I'd given a very high-level read and then didn't communicate well. Because scheduling workers are shared between all jobs, the maximum priority value has to be shared between all jobs and can't be configured on a per-job basis. The server config is the maximum value of the priority field.

So the `structs.Job` struct doesn't need a user-defined max priority either. The check that we have in the `(*Job).Validate` method should be checking against the server configured max priority, not a job-internal maximum. So something like the following:

```diff
-if j.Priority < JobMinPriority || j.Priority > JobMaxPriority {
+if j.Priority < JobMinPriority || j.Priority > srv.Server.config.JobMaxPriority {
```

Note that the server config isn't currently exposed to the `Validate` method, so you'll need to thread that through.

nomad/structs/structs.go
```diff
@@ -4156,17 +4156,19 @@ const (
...
 	JobMinPriority = 1
 
-	// JobDefaultPriority is the default priority if not
-	// not specified.
+	// JobDefaultPriority is the default priority if not specified.
 	JobDefaultPriority = 50
 
-	// JobMaxPriority is the maximum allowed priority
-	JobMaxPriority = 100
+	// JobDefaultMaxPriority is the default maximum allowed priority
```
tgross, line 4162: Going to nitpick on the comment text here:

```suggestion
	// JobDefaultMaxPriority is the maximum allowed value for the default priority
```

api/jobs.go
alessio-perugini: No worries, thank you for this deeper explanation! :raised_hands: 
In the previous draft, I leaked the JobMaxPriority config from the server directly in the new struct field. 
With your suggestion, the cleanest way I could come up with was to move this check directly into the `jobValidate.Validate` (nomad/job_endpoint_hooks.go:264).
Do you see any downside to this approach? :thinking: 

(I did an amend with the new solution to the previous commit)

website/content/docs/configuration/server.mdx
```diff
@@ -249,6 +249,12 @@ server {
 - `search` <code>([search][search]: nil)</code> - Specifies configuration parameters
   for the Nomad search API.
 
+- `job_max_priority` `(int: 100)` - Specifies the maximum priority that can be assigned to a job.
+   A valid value must be between `100` and `32766`
+
+- `job_default_priority` `(int: 50)` - Specifies the default priority assigned to a job.
+   A valid value must be between `50` and `job_max_priority`
```
tgross, line 256:
```suggestion
- `job_max_priority` `(int: 100)` - Specifies the maximum priority that can be assigned to a job.
   A valid value must be between `100` and `32766`.

- `job_default_priority` `(int: 50)` - Specifies the default priority assigned to a job.
   A valid value must be between `50` and `job_max_priority`.
```

nomad/job_endpoint_test.go
```diff
@@ -6651,6 +6651,61 @@ func TestJobEndpoint_ValidateJobUpdate_ACL(t *testing.T) {
...
+		require.Contains(t, err.Error(), "job priority must be between")
+	})
+
+	t.Run("job with invalid max priority", func(t *testing.T) {
+		j := mock.Job()
+		j.Priority = 101
+
+		err := validateJob(j)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "job priority must be between")
```
tgross, line 6705: We're gradually migrating over to https://github.com/shoenig/test/ which has much better equality detection. So new tests should use that where possible:

```suggestion
		must.Error(t, err)
		must.ErrorContains(t, err, "job priority must be between")
```

(similarly for the other new tests you've introduced in this PR)

command/agent/agent.go
```diff
@@ -326,6 +326,23 @@ func convertServerConfig(agentConfig *Config) (*nomad.Config, error) {
 		}
 	}
 
+	jobMaxPriority := int(structs.JobDefaultMaxPriority)
```
tgross, line 329: Isn't this already an int?

```suggestion
	jobMaxPriority := structs.JobDefaultMaxPriority
```

command/agent/agent.go
```diff
@@ -326,6 +326,23 @@ func convertServerConfig(agentConfig *Config) (*nomad.Config, error) {
...
+	if agentConfig.Server.JobMaxPriority != nil && *agentConfig.Server.JobMaxPriority != 0 {
+		jobMaxPriority = *agentConfig.Server.JobMaxPriority
+		if jobMaxPriority < structs.JobDefaultMaxPriority || jobMaxPriority > structs.JobMaxPriority {
+			return nil, fmt.Errorf("job_max_priority cannot be %d. Must be between %d and %d", *agentConfig.Server.JobMaxPriority, structs.JobDefaultMaxPriority, structs.JobMaxPriority)
+		}
+	}
+	jobDefaultPriority := int(structs.JobDefaultPriority)
```
tgross, line 336:
```suggestion
	jobDefaultPriority := structs.JobDefaultPriority
```

nomad/structs/structs.go
```diff
@@ -4156,17 +4156,19 @@ const (
...
 
-	// JobDefaultPriority is the default priority if not
-	// not specified.
+	// JobDefaultPriority is the default priority if not specified.
 	JobDefaultPriority = 50
 
+	// JobDefaultMaxPriority is the default maximum allowed priority
+	JobDefaultMaxPriority = 100
+
 	// JobMaxPriority is the maximum allowed priority
```
tgross, line 4165: Just to make the purpose of this const clear:

```suggestion
	// JobMaxPriority is the maximum allowed configuration value for maximum job priority
```

tgross (changes requested):
This is looking good, @alessio-perugini! 

I've left some comments to address. In addition, if you can run `make cl` that'll ask you some questions to generate a changelog file for this PR.

api/jobs.go
```diff
@@ -938,7 +941,7 @@ func (j *Job) Canonicalize() {
 		j.Namespace = pointerOf(DefaultNamespace)
 	}
 	if j.Priority == nil {
-		j.Priority = pointerOf(50)
+		j.Priority = pointerOf(DefaultPriority)
```
tgross, line 944: Sorry I missed this when I was reviewing the code, but this isn't quite going to work. When I checked this PR out locally and ran it, I discovered that the new `job_default_priority` isn't being respected. That's because when you submit a job without a `priority` set, it's defaulting to this hard-coded value instead of the value from the server.

I think what we're going to want to do here is set the value to 0 and then on the server treat `0` as "unset" so that it falls back to the default.

tgross (changes requested):

api/jobs.go
alessio-perugini: Nice catch! All those mutations are a bit confusing :exploding_head:

I've pushed a new commit, but I'm not sure if it's enough. :thinking:  

Should we also update the `ValidateJobRequest`? (command/agent/job_endpoint.go)

```go
func (s *HTTPServer) ValidateJobRequest(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
...
	job := ApiJobToStructJob(validateRequest.Job)
	if job.Priority == 0 {
		job.Priority = s.agent.Server().GetConfig().JobDefaultPriority
	}
...
```

command/agent/job_endpoint.go
```diff
@@ -196,6 +196,11 @@ func (s *HTTPServer) ValidateJobRequest(resp http.ResponseWriter, req *http.Requ
 
 	job := ApiJobToStructJob(validateRequest.Job)
 
+	// If the job priority is not set, we fallback on the defaults specified in the server config
+	if job.Priority == 0 {
+		job.Priority = s.agent.Server().GetConfig().JobDefaultPriority
+	}
```
tgross, line 202: I see where you're going here, but we've already got a `jobCanonicalizer` in that same `nomad/job_endpoint_hooks.go` where you've updated the `jobValidator`. It calls `(*structs.Job).Canonicalize`, to which we could pass the config value.

website/content/docs/job-specification/job.mdx
```diff
@@ -105,8 +105,10 @@ job "docs" {
   at fixed times, dates or intervals.
 
 - `priority` `(int: 50)` - Specifies the job priority which is used to
-  prioritize scheduling and access to resources. Must be between 1 and 100
-  inclusively, with a larger value corresponding to a higher priority.
+  prioritize scheduling and access to resources.
+  Must be between 1 and [`server` block `job_max_priority`] inclusively,
```
tgross, line 109: I'd put this as "Must be between 1 and the server's `job_max_priority` configuration"

tgross (changes requested):
I see you picked up the tests, thanks. I think we're getting close on this one

command/agent/job_endpoint.go
alessio-perugini: :exploding_head: I completely forgot about the existence of this hook! It was much simpler than I thought, and it feels way nicer this way!

tgross (approved):
LGTM! I know reviews for seemingly minor features can end up taking a surprising amount of back and of forth. Thanks for sticking with this, @alessio-perugini!

This will ship in Nomad 1.5.0 GA

alessio-perugini:
@tgross Thank you so much for your patience and help. This was my first time contributing to a project of this scale. I had so much fun working on this. Looking forward to helping you with other issues.
Keep rocking! 🤘

PR #16089, opened 2023-02-07
generate serialization code for IdentityClaims

tgross:
Originally IdentityClaims never were passed over the wire, but they're now used for the `ACL.WhoAmI` endpoint. Generate the serialization code for that.

### DIFF
--- a/nomad/structs/generate.sh
+++ b/nomad/structs/generate.sh
@@ -8,5 +8,4 @@ codecgen \
     -d 100 \
     -t codegen_generated \
     -o structs.generated.go \
-    -nr="^IdentityClaims$" \
     ${FILES}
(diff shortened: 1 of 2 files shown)


tgross:
Moving to draft as it turns out we can't generate this automatically because of the `jwt.NumericDate` field. We'll need a custom encoder I think (or just eat the tiny perf hit for this little-used endpoint):

> Error: nomad/structs/structs.generated.go:185141:19: x.ExpiresAt.CodecEncodeSelf undefined (type *"github.com/golang-jwt/jwt/v4".NumericDate has no field or method CodecEncodeSelf)

tgross:
Going to close this one out for now. See https://github.com/hashicorp/nomad/issues/16129 for the reasoning why and how we'll tackle this in the future when we have some time.

Issue #16010, opened 2023-02-01
Cache and/or Memoize users.Lookup results


schmichael:
[`users.Lookup`](https://github.com/hashicorp/nomad/blob/v1.4.3/helper/users/lookup.go#L29-L34) is a wrapper around Go's [`os/user.Lookup`](https://pkg.go.dev/os/user#Lookup) which adds a global lock to workaround thread safety issues in Unix NSS implementations (the Go stdlib itself makes no goroutine safety claims in its version, so it is implied to be **not threadsafe**).

Prior to Nomad v1.5 this lookup was only called [on Linux when starting the executor](https://github.com/hashicorp/nomad/blob/v1.4.3/drivers/shared/executor/executor_universal_linux.go#L24).

As of #15755 it will also be called for every task specifying `identity { file = true }`, so in many places the lookup will be redundant.

Since this helper uses a global lock *and* may incur a network request (Windows Domain users, directory lookups on Unixes), it should be cached and/or memoized.

It may be possible/optimal to attach the result to TaskRunner for reuse for the duration of the task's/agent's lifecycle.

PR #16093, opened 2023-02-08
users: eliminate LookupGroupId and its one use case

shoenig:
This PR deletes the `user.LookupGroupId` function as it was only being used
in a single test case, and its value was not important to the test.

Prep for https://github.com/hashicorp/nomad/issues/16010

No CL - only affects test case


tgross (approved):
LGTM

Issue #11052, opened 2021-08-13
Multiple instances of a periodic job are run simultaneously, when prohibit_overlap is true


madsholden:
### Nomad version
Nomad v1.1.0 (2678c3604bc9530014208bc167415e167fd440fc)

### Operating system and Environment details

Ubuntu 20.04.2 on EC2

### Issue

I have a periodic job set to run every half hour, with prohibit_overlap set to true. It is usually finished in much shorter time, but now it's running for longer, which is expected. However, Nomad has started three instances of this job. When checking the Nomad server logs I noticed both times when it started a new job, the servers were in the middle of a server election.

### Reproduction steps

I can't reproduce it, it usually doesn't start a new instance when one is running.

#### Expected Result

Nomad should never run two instances of the same job.

#### Actual Result

Three instances of the job is running.

### Job file (if appropriate)

Only relevant parts included. Please let me know if you want more of the job file.

```
job "data-remover" {
  type = "batch"

  periodic {
    cron = "30 * * * *"
    prohibit_overlap = true
  }

  group "data-remover" {
    task "data-remover" {
      driver = "docker"
```

### Nomad Server logs (if appropriate)

From the Nomad UI:
Aug 03, '21 01:33:58 +0200 | Received | Task received by client
Aug 12, '21 03:11:02 +0200 | Received | Task received by client

Server logs around those times:

```
2021-08-12 01:11:02.673	ip-10-49-15-7
{"@level":"info","@message":"pipelining replication","@module":"nomad.raft","@timestamp":"2021-08-12T01:11:02.673964Z","peer":{"Suffrage":0,"ID":"674dedad-4ea5-6e68-d59e-abac98e1c353","Address":"10.49.11.101:4647"}}
2021-08-12 01:11:02.672	ip-10-49-15-7
{"@level":"info","@message":"pipelining replication","@module":"nomad.raft","@timestamp":"2021-08-12T01:11:02.672838Z","peer":{"Suffrage":0,"ID":"f720e37e-203b-0d55-af95-19783d1663b5","Address":"10.49.13.249:4647"}}
2021-08-12 01:11:02.666	ip-10-49-15-7
{"@level":"info","@message":"cluster leadership acquired","@module":"nomad","@timestamp":"2021-08-12T01:11:02.666016Z"}
2021-08-12 01:11:02.665	ip-10-49-15-7
[...]

PR #16097, opened 2023-02-08
scheduler: system scheduler should reconcile overlapping live allocs

tgross:
Under conditions that are not yet understood, it's possible for there to be overlapping live system job allocations on the same node. The scheduler does not reconcile this condition by stopping any extra allocations, and the extras also do not show up in the plan either.

When allocations need exclusive resources such as reserved ports, this results in the plan applier rejecting the plan. As far as the scheduler knows the plan is good, and because this is for a system job, the scheduler will repeatedly submit this bad plan. This results in the dreaded "plan for node rejected" loop. Although we should definitely figure out why we get into this bad state to begin with, it's the job of the scheduler to properly reconcile the cluster state even in the face of errors.

This changeset includes two major chunks of work:
* Add a new `ensureMaxSystemAllocCount` function that enforces the invariant that there can be only a single desired-running allocation for a given system job on a given node (or a number == `count` for `sysbatch` jobs).
* The tests for the system allocs reconciling code path (`diffSystemAllocs`) include many impossible test environments, such as passing allocs for the wrong node into the function. This makes the test assertions nonsensible for use in walking yourself through the correct behavior. This changeset breaks up a couple of tests, expands test coverage, and makes test assertions more clear.

Note this doesn't completely close out the Plan For Node Rejected bug class, because we're pretty sure this can also happen with service jobs. I haven't tracked that down yet, but this PR should eliminate a large chunk of known problems we've seen on some large clusters.

Note for reviewers: you might want to read this commit-by-commit
Closes https://github.com/hashicorp/team-nomad/issues/347

### DIFF
--- a/scheduler/scheduler_system_test.go
+++ b/scheduler/scheduler_system_test.go
@@ -8,14 +8,16 @@ import (
 	"time"
 
 	memdb "github.com/hashicorp/go-memdb"
+	"github.com/shoenig/test"
+	"github.com/shoenig/test/must"
+	"github.com/stretchr/testify/require"
+
 	"github.com/hashicorp/nomad/ci"
 	"github.com/hashicorp/nomad/helper"
 	"github.com/hashicorp/nomad/helper/pointer"
 	"github.com/hashicorp/nomad/helper/uuid"
 	"github.com/hashicorp/nomad/nomad/mock"
 	"github.com/hashicorp/nomad/nomad/structs"
-	"github.com/shoenig/test/must"
-	"github.com/stretchr/testify/require"
 )
 
 func TestSystemSched_JobRegister(t *testing.T) {
@@ -3083,4 +3085,67 @@ func TestSystemSched_CSITopology(t *testing.T) {
 	must.Eq(t, "", h.Evals[0].BlockedEval, must.Sprint("did not expect a blocked eval"))
 	must.Eq(t, structs.EvalStatusComplete, h.Evals[0].Status)
 
+}
+
+func TestSystemSched_OverlappingAllocations(t *testing.T) {
+	ci.Parallel(t)
+	h := NewHarness(t)
+
+	node := mock.Node()
+	node.Name = "good-node-1"
+	must.NoError(t, h.State.UpsertNode(structs.MsgTypeTestSetup, h.NextIndex(), node))
+
+	// Create 2 allocs for 2 versions of the same system job on the same node.
+	// This should not happen but has been observed, so the scheduler must "fix"
+	// this by stopping the old version.
+	oldJob := mock.SystemJob()
+	oldJob.ID = "my-job"
+	must.NoError(t, h.State.UpsertJob(structs.MsgTypeTestSetup, h.NextIndex(), oldJob))
+
+	job := oldJob.Copy()
+	job.Version++
+	must.NoError(t, h.State.UpsertJob(structs.MsgTypeTestSetup, h.NextIndex(), job))
+
+	allocOld := mock.Alloc()
+	allocOld.Job = oldJob
+	allocOld.JobID = oldJob.ID
+	allocOld.NodeID = node.ID
+	allocOld.Name = "my-job.web[0]"
+
+	allocCurrent := allocOld.Copy()
+	allocCurrent.ID = uuid.Generate()
+	allocCurrent.Job = job
+
+	must.NoError(t, h.State.UpsertAllocs(structs.MsgTypeTestSetup, h.NextIndex(),
+		[]*structs.Allocation{
+			allocOld, allocCurrent}))
+
+	// Create a mock evaluation to deregister the job
+	eval := &structs.Evaluation{
+		Namespace:   structs.DefaultNamespace,
+		ID:          uuid.Generate(),
+		Priority:    job.Priority,
+		TriggeredBy: structs.EvalTriggerJobRegister,
+		JobID:       job.ID,
+		Status:      structs.EvalStatusPending,
+	}
+	lastIndex := h.NextIndex()
+	must.NoError(t, h.State.UpsertEvals(structs.MsgTypeTestSetup,
+		lastIndex, []*structs.Evaluation{eval}))
+
+	// Process the evaluation
+	must.NoError(t, h.Process(NewSystemScheduler, eval))
+
+	// Ensure a single plan
+	must.Len(t, 1, h.Plans)
+	plan := h.Plans[0]
+	must.Nil(t, plan.Annotations, must.Sprint("expected no annotations"))
+
+	test.Len(t, 1, plan.NodeUpdate[node.ID], test.Sprint("expected 1 evict/stop"))
+	if len(plan.NodeUpdate[node.ID]) > 0 {
+		test.Eq(t, allocOld.ID, plan.NodeUpdate[node.ID][0].ID,
+			test.Sprintf("expected alloc=%s to be evicted/stopped", allocOld.ID))
+	}
+
+	test.Len(t, 0, plan.NodeAllocation[node.ID], test.Sprint("expected no updates"))
 }
--- a/scheduler/system_util.go
+++ b/scheduler/system_util.go
@@ -8,13 +8,15 @@ package scheduler
 
 import (
 	"fmt"
+	"sort"
 	"time"
 
 	"github.com/hashicorp/nomad/nomad/structs"
 )
 
-// materializeSystemTaskGroups is used to materialize all the task groups
-// a system or sysbatch job requires.
+// materializeSystemTaskGroups is used to materialize all the task groups a
+// system or sysbatch job requires. Returns a map of allocation names to task
+// groups.
 func materializeSystemTaskGroups(job *structs.Job) map[string]*structs.TaskGroup {
 	out := make(map[string]*structs.TaskGroup)
 	if job.Stopped() {
@@ -45,24 +47,34 @@ func diffSystemAllocsForNode(
 	eligibleNodes map[string]*structs.Node,
 	notReadyNodes map[string]struct{}, // nodes that are not ready, e.g. draining
 	taintedNodes map[string]*structs.Node, // nodes which are down (by node id)
-	required map[string]*structs.TaskGroup, // set of allocations that must exist
+	required map[string]*structs.TaskGroup, // set of task groups (by alloc name) that must exist
 	allocs []*structs.Allocation, // non-terminal allocations that exist
 	terminal structs.TerminalByNodeByName, // latest terminal allocations (by node, id)
 	serverSupportsDisconnectedClients bool, // flag indicating whether to apply disconnected client logic
 ) *diffResult {
-	result := new(diffResult)
 
-	// Scan the existing updates
-	existing := make(map[string]struct{}) // set of alloc names
+	// Track a map of task group names (both those required and not) to
+	// diffResult for allocations of that name. This lets us enforce task-group
+	// global invariants before we merge all the results together for the node.
+	results := map[string]*diffResult{}
+
+	// Track the set of allocation names we've
+	// seen, so we can determine if new placements are needed.
+	existing := make(map[string]struct{})
+
 	for _, exist := range allocs {
-		// Index the existing node
 		name := exist.Name
 		existing[name] = struct{}{}
 
-		// Check for the definition in the required set
-		tg, ok := required[name]
+		result := results[name]
+		if result == nil {
+			result = new(diffResult)
+			results[name] = result
+		}
 
-		// If not required, we stop the alloc
+		// Check if the allocation's task group is in the required set (it might
+		// have been dropped from the jobspec). If not, we stop the alloc
+		tg, ok := required[name]
 		if !ok {
 			result.stop = append(result.stop, allocTuple{
 				Name:      name,
@@ -88,19 +100,9 @@ func diffSystemAllocsForNode(
 			}
 		}
 
-		// If we have been marked for migration and aren't terminal, migrate
-		if !exist.TerminalStatus() && exist.DesiredTransition.ShouldMigrate() {
+		// If we have been marked for migration, migrate
+		if exist.DesiredTransition.ShouldMigrate() {
 			result.migrate = append(result.migrate, allocTuple{
-				Name:      name,
-				TaskGroup: tg,
-				Alloc:     exist,
-			})
-			continue
-		}
-
-		// If we are a sysbatch job and terminal, ignore (or stop?) the alloc
-		if job.Type == structs.JobTypeSysBatch && exist.TerminalStatus() {
-			result.ignore = append(result.ignore, allocTuple{
 				Name:      name,
 				TaskGroup: tg,
 				Alloc:     exist,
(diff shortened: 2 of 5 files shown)


scheduler/system_util.go
```diff
@@ -98,16 +100,6 @@ func diffSystemAllocsForNode(
...
 
-		// If we are a sysbatch job and terminal, ignore (or stop?) the alloc
-		if job.Type == structs.JobTypeSysBatch && exist.TerminalStatus() {
-			result.ignore = append(result.ignore, allocTuple{
-				Name:      name,
-				TaskGroup: tg,
-				Alloc:     exist,
-			})
-			continue
-		}
```
tgross, line 109: Note to reviewers: this is dead code, confirmed by test coverage from the original tests plus the refactored ones. Only non-terminal allocs are passed in the list we're iterating over here for the `exist` block.

scheduler/system_util.go
```diff
@@ -291,6 +287,66 @@ func diffSystemAllocsForNode(
 	return result
 }
 
+// ensureSingleSystemAlloc enforces the invariant that a system or sysbatch job
+// should never have more than one allocation in a desired-running state.
+func ensureSingleSystemAlloc(result *diffResult) {
```
tgross, line 292: Note for reviewers: I don't _love_ this implementation but because we can't rely on the order of allocations being passed in to `diffSystemAllocsForNode` (because they come from a map), it's hard to make decisions between loop iterations without hugely expanding a loop that's already huge and hard to follow. This possibly costs us a couple of extra slice memory allocations, but I think its worth it to keep it understandable?

scheduler/system_util.go
```diff
@@ -291,6 +287,64 @@ func diffSystemAllocsForNode(
...
+		return
+	}
+
+	// sort by JobModifyIndex, then CreateIndex
+	sortTuples := func(tuples []allocTuple) {
+		sort.Slice(tuples, func(i, j int) bool {
+			if tuples[i].Alloc.Job.JobModifyIndex == tuples[j].Alloc.Job.JobModifyIndex {
+				return tuples[i].Alloc.CreateIndex > tuples[j].Alloc.CreateIndex
+			}
+			return tuples[i].Alloc.Job.JobModifyIndex > tuples[j].Alloc.Job.JobModifyIndex
```
shoenig, line 304: golf: sometimes is is nice to assign intermediate variables in these so the statements are less verbose, e.g.

```
func(i, j int) bool {
  I, J := tuples[i].Alloc, tuples[j].Alloc
  if I.Job.ModifyIndex == J.Job.ModifyIndex {
  ...
}
```

scheduler/system_util.go
```diff
@@ -291,6 +287,64 @@ func diffSystemAllocsForNode(
...
 
+// ensureSingleSystemAlloc enforces the invariant that a system or sysbatch job
+// should never have more than one allocation in a desired-running state.
+func ensureSingleSystemAlloc(result *diffResult) {
+
+	if len(result.ignore)+len(result.update)+len(result.reconnecting) <= 1 {
+		return
+	}
+
+	// sort by JobModifyIndex, then CreateIndex
```
shoenig, line 298: clarify descending; abnormal from typical use of `.Slice`

scheduler/system_util.go
shoenig: actually we probably don't need to sort these - we only need to iterate through and find the index of the max, then iterate again copying to []stop but skipping the max, then creating the [1] with the max in it

if these slices are always small the Sort is probably fine and definitely more readable

scheduler/scheduler_system_test.go
```diff
@@ -3084,3 +3086,63 @@ func TestSystemSched_CSITopology(t *testing.T) {
...
+
+	// Process the evaluation
+	must.NoError(t, h.Process(NewSystemScheduler, eval))
+
+	// Ensure a single plan
+	must.Len(t, 1, h.Plans)
+	plan := h.Plans[0]
+	must.Nil(t, plan.Annotations, must.Sprint("expected no annotations"))
+
+	test.Len(t, 1, plan.NodeUpdate[node.ID], test.Sprint("expected 1 evict/stop"))
```
shoenig, line 3141: was the switch to `test.` here and below intentional?

scheduler/system_util.go
shoenig: Is the first condition in the block above also unnecessary then?

```
if !exist.TerminalStatus() && exist.DesiredTransition.ShouldMigrate() {
```

shoenig (commented):
(will continue tmrw)

scheduler/system_util.go
tgross: Looks like it!

scheduler/system_util.go
tgross: The inputs are the ignore/update/reconnecting allocations that were assigned to a single node, so that constrains the size of these slices quite a bit. Most of the time if we don't hit the bail-out early condition I'd expect there to be only 2 or 3 items between all 3 slices.

shoenig (approved):
LGTM!

lgfa29 (changes requested):
Great investigation! I think the change to `materializeSystemTaskGroups` introduces a bug for `sysbatch` jobs.

scheduler/scheduler_system_test.go
```diff
@@ -3084,3 +3086,63 @@ func TestSystemSched_CSITopology(t *testing.T) {
...
+
+func TestSystemSched_OverlappingAllocations(t *testing.T) {
+	ci.Parallel(t)
+	h := NewHarness(t)
+
+	node := mock.Node()
+	node.Name = "good-node-1"
+	must.NoError(t, h.State.UpsertNode(structs.MsgTypeTestSetup, h.NextIndex(), node))
+
+	oldJob := mock.SystemJob()
```
schmichael, line 3098: This is a great test but let's leave some context for future-us to remember why it has a seemingly impossible setup.

```suggestion
	// Create 2 allocs for 2 versions of the same system job on the same node.
	// This should not happen but has been observed, so the scheduler must "fix" this by stopping the old version.
	oldJob := mock.SystemJob()
```

scheduler/system_util.go
```diff
@@ -152,11 +144,12 @@ func diffSystemAllocsForNode(
 		// If we are on a tainted node, we must migrate if we are a service or
 		// if the batch allocation did not finish
 		if nodeIsTainted {
-			// If the job is batch and finished successfully, the fact that the
-			// node is tainted does not mean it should be migrated or marked as
-			// lost as the work was already successfully finished. However for
-			// service/system jobs, tasks should never complete. The check of
-			// batch type, defends against client bugs.
+			// If the job is batch and finished successfully (but not yet marked
+			// terminal), the fact that the node is tainted does not mean it
```
schmichael, line 148: Huh I'm surprised this is possible because of: https://github.com/hashicorp/nomad/blob/v1.5.0-beta.1/client/allocrunner/alloc_runner.go#L812-L843

When sending alloc updates from client->server, the client computes the alloc's ClientStatus based on the state of every task precisely so that from the scheduler's perspective `all tasks completed -> alloc completed` is in an atomic update and therefore an invariant.

This suggests that state transition isn't atomic/invariant... I don't mind being defensible against misbehaving clients here (as I think we should do that more!), but it does make me wonder if we should rethink the approach the linked code in the client does entirely.... perhaps the fsm should _also_ enforce the invariant that `all task states being terminal == alloc's client status is terminal`?

lgfa29:
> Although we should definitely figure out why we get into this bad state to begin with

I don't know if it's related by https://github.com/hashicorp/nomad/issues/11052 reports a problem with overlapping allocs during leader election.

scheduler/system_util.go
```diff
@@ -8,24 +8,26 @@ package scheduler
...
 
 	for _, tg := range job.TaskGroups {
-		for i := 0; i < tg.Count; i++ {
-			name := fmt.Sprintf("%s.%s[%d]", job.Name, tg.Name, i)
-			out[name] = tg
-		}
+		// note that all system/sysbatch allocs for the same task group share
+		// the same alloc index 0
+		name := fmt.Sprintf("%s.%s[0]", job.Name, tg.Name)
+		out[name] = tg
```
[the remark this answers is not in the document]
tgross: > I believe tg.Count is always 1 for system jobs, but not for sysbatch?

`tg.Count` is ignored for system jobs, because we always run 1 per node that meets the requirements. You can omit the field, which means that most of the time the value will be 0.

I suspect allowing `tg.Count > 1` for `sysbatch` jobs is just a bug in validation. If you actually run a `sysbatch` job with `count=3` with 2 eligible clients, you get 1 per node. The same behavior we get with `system` jobs, which is what I'd expect.

schmichael (approved):
Fantastic work. I had a big long comment about `being defensive in the scheduler` vs `being defensive in rpcs and cleaning up bad state in the fsm`, but I hit the back button and lost it all..

...the thrilling conclusion was that your approach in this PR seems like the optimal approach to me: bad state is caught by the scheduler, so no matter how much work we do to try to prevent it from existing, being defensive in the scheduler is always attacking the problem where it impacts the cluster.

scheduler/system_util.go
tgross: It's entirely possible that it _isn't_ possible to reach this state. This block was introduced at the same time as the dead code I removed above that assumes we can have terminal allocations here at all, which we can't. I added the `(but not yet marked terminal)` parenthetical here because I assumed it was correct, but on reflection maybe it's unreachable outside of unit tests.

scheduler/system_util.go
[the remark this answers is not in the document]
lgfa29: > If you actually run a sysbatch job with count=3 with 2 eligible clients, you get 1 per node.

Did you test using a binary built with this branch? From `main` and 1.4.3 I get 4 allocations if you run a job like this in a cluster with 2 clients.

```hcl
job "example" {
  type        = "sysbatch"
  datacenters = ["dc1"]

  group "echo" {
    count = 2

    task "echo" {
      driver = "raw_exec"

      config {
        command = "echo"
        args    = ["hi"]
      }
    }
  }
}
```

```console
$ nomad job status example
ID            = example
Name          = example
Submit Date   = 2023-02-09T14:04:09-05:00
Type          = sysbatch
Priority      = 50
Datacenters   = dc1
Namespace     = default
Status        = dead
Periodic      = false
Parameterized = false

Summary
Task Group  Queued  Starting  Running  Failed  Complete  Lost  Unknown
echo        0       0         0        0       4         0     0

Allocations
ID        Node ID   Task Group  Version  Desired  Status    Created  Modified
65b2be15  cd72b386  echo        0        run      complete  8s ago   7s ago
7f67849c  327ab09f  echo        0        run      complete  8s ago   7s ago
93a4a3b2  327ab09f  echo        0        run      complete  8s ago   7s ago
f244a8ff  cd72b386  echo        0        run      complete  8s ago   7s ago
```

Running a binary from this branch gives me only 2 allocs.

I don't have any preference on either allowing or preventing count > 1 for sysbatch, but if we add a validation to prevent it it would be a breaking change.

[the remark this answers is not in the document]
tgross: Shoot, I was certain I'd run it against `main` but maybe I forgot to rebuild after I checked out `main`. :blush: You're right that this is how sysbatch works, which means this entire patch is broken for sysbatch as-is, because we're enforcing a single ignored/updated/reconnected alloc.

I'm kinda of the opinion that this is a sysbatch bug, but the [RFC](https://docs.google.com/document/d/1FlYV71mRmC_qTHfXggsXD7ONICtyFqe5_yQDqHNsuU8/edit) doesn't say either way. Lemme bring this up with the team.

lgfa29 (approved):
After talking more about it the `count` behaviour for `sysbatch` jobs is not well specified, so we'll need to handle it in follow-up PRs

scheduler/system_util.go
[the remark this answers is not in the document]
tgross: Oof, just realized there's a fairly serious bug here if the system job has more than one task group

tgross:
I've reworked the implementation so that there's now a `ensureMaxSystemAllocCount` instead of a function that ensures we have a single one. This covers `sysbatch` jobs when `count > 1`. I've also moved the assertion to inside the loop for each task group, so that we've properly covered cases where a job has multiple task groups.

The other good news about this change is that I suspect this same function could now be used for the service jobs as well, but I haven't figured out the right place to hook it in yet, so I'll do that in a separate PR.

scheduler/system_util.go
tgross: Returning to this, I think we should probably address it, but I want to pull that out to a separate PR just because it's not _obviously_ correct to remove this and I don't want to muddy up this commit with that if we have to git bisect it later.

scheduler/system_util.go
```diff
@@ -288,7 +309,78 @@ func diffSystemAllocsForNode(
...
+
+	return finalResult
+}
+
+// ensureMaxSystemAllocCount enforces the invariant that the per-node diffResult
+// we have for a system or sysbatch job should never have more than "count"
+// allocations in a desired-running state.
+func ensureMaxSystemAllocCount(result *diffResult, count int) {
+
+	// sort descending by TaskGroup name, then JobModifyIndex, then
```
shoenig, line 326: where do we sort by TaskGroup name?

shoenig (dismissed):
LGTM. This has got me wondering about whether sysbatch jobs with count > 1 could be better validated, e.g. checking for duplicate static ports, etc.

scheduler/system_util.go
tgross: Oops, we don't. I added that in my first attempt and forgot to remove it. Will fix.

tgross: Fixed in 661fd31

scheduler/system_util.go
```diff
@@ -288,7 +309,78 @@ func diffSystemAllocsForNode(
...
+			result.update = []allocTuple{}
+			result.reconnecting = []allocTuple{}
+			return
+		}
+	}
+
+	// updated allocs are for updates of the job (in-place or destructive), so
+	// we can pick the most recent one and stop all the rest of the running
+	// allocs on the node
+	if len(result.update) > 0 {
```
tgross, line 361: Something has been bugging me about this PR for the last few days and I finally realized what it was: we determine destructive vs non-destructive updates _after_ this per-node call is made and then limit the allocation changes based on `MaxParallel`. 

If we treat the non-destructive and destructive updates the same here, I'm fairly certain we'll break the system not-really-deployments feature of `MaxParallel/Stagger`, which makes decisions only based on the destructive endpoints. For example, suppose we have 3 nodes with `MaxParallel = 1`. One of the nodes has both a destructive and non-destructive update to make, but the alloc getting the non-destructive update is newer. In a single pass, we could end up stopping the alloc with the destructive update, update the alloc with the non-destructive update, and update an alloc on another node with a destructive update. That'd be more than `MaxParallel = 1` destructive updates (inasmuch as a "stop" is destructive).

I think we need to move the destructive vs non-destructive checking into the diff code and have a separate diff field for destructive vs non-destructive updates. But that'll ripple out to the generic scheduler as well. I'm going to need a little bit of time to get this locked down, so I think it's going to have to miss 1.5.0 GA.

tgross:
The root cause for the original bug here will be solved in https://github.com/hashicorp/nomad/pull/16401. So I'm going to pull the test improvements out of this PR to land, and then see about tackling the diff changes as second body of work when I next dig into the reconciler code.

tgross:
I'm going to close this stale PR. The issue this was intended to solve is believed to be solved by https://github.com/hashicorp/nomad/pull/16401 which shipped long ago. The approach here is dubious on a bunch of levels, and to be honest between refactoring and other bug fixes I'm not sure this could be reasonably merged without completely reworking it anyways.

PR #16100, opened 2023-02-08
users: create cache for user lookups

### LINKED ISSUE #16010: Cache and/or Memoize users.Lookup results

shoenig:
This PR introduces a global cache for OS user lookups. This should
relieve pressure on the OS domain/directory lookups, which would be
queried more now that Task API exists.

Hits are cached for 1 hour, and misses are cached for 1 minute. These
values are fairly arbitrary - we can tweak them if there is any reason to.

Closes #16010

### DIFF
--- a/helper/users/cache.go
+++ b/helper/users/cache.go
@@ -1,0 +1,81 @@
+package users
+
+import (
+	"os/user"
+	"sync"
+	"time"
+
+	"github.com/hashicorp/nomad/lib/lang"
+	"oss.indeed.com/go/libtime"
+)
+
+const (
+	cacheTTL   = 1 * time.Hour
+	failureTTL = 1 * time.Minute
+)
+
+type entry[T any] lang.Pair[T, time.Time]
+
+func (e *entry[T]) expired(now time.Time, ttl time.Duration) bool {
+	return now.After(e.Second.Add(ttl))
+}
+
+type (
+	userCache        map[string]*entry[*user.User]
+	userFailureCache map[string]*entry[error]
+)
+
+type lookupUserFunc func(string) (*user.User, error)
+
+type cache struct {
+	clock      libtime.Clock
+	lookupUser lookupUserFunc
+
+	lock         sync.Mutex
+	users        userCache
+	userFailures userFailureCache
+}
+
+func newCache() *cache {
+	return &cache{
+		clock:        libtime.SystemClock(),
+		lookupUser:   internalLookupUser,
+		users:        make(userCache),
+		userFailures: make(userFailureCache),
+	}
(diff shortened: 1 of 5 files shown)


helper/users/cache.go
```diff
@@ -0,0 +1,81 @@
...
+
+type entry[T any] lang.Pair[T, time.Time]
+
+func (e *entry[T]) expired(now time.Time, ttl time.Duration) bool {
+	return now.After(e.Second.Add(ttl))
+}
+
+type (
+	userCache        map[string]*entry[*user.User]
+	userFailureCache map[string]*entry[error]
```
lgfa29, line 25: Any concerns about unlimited memory growth? I don't expect these to grow too much, and creating another GC routine is a bit much 😅  but maybe we can delete entries on read once they expire?

helper/users/cache_test.go
```diff
@@ -0,0 +1,151 @@
...
+)
+
+func TestCache_real_hit(t *testing.T) {
+	ci.Parallel(t)
+
+	c := newCache()
+
+	// fresh lookup
+	u, err := c.GetUser("nobody")
+	must.NoError(t, err)
```
lgfa29, line 21: Would this work on Windows?

lgfa29 (approved):
Minor comments, but LGTM!

helper/users/cache.go
shoenig: Oh yeah that's a nice idea. In practice this _should_ always be zero, but yeah people certainly do stuff to their clusters.

helper/users/cache_test.go
shoenig: 🤦

PR #16102, opened 2023-02-09
Workload Identity, Task API, and Dynamic Node Metadata Docs

schmichael:
Better late than never! Here is some much needed documentation for new features in the freshly released 1.5.0-beta.1 release.

A new tutorial demonstrating how to use the Task API to serve the Nomad UI would be neat, but that's a job for another PR.

I added Task API to the little section above the list of endpoints:

Screenshot from 2023-02-08 16-51-09

### DIFF
--- a/website/content/api-docs/client.mdx
+++ b/website/content/api-docs/client.mdx
@@ -67,10 +67,6 @@ The table below shows this endpoint's support for
 $ nomad operator api /v1/client/metadata
 ```
 
-```shell-session
-$ curl http://127.0.0.1:4646/v1/client/metadata
-```
-
 ### Sample Response
 
 Formatted by appending `?pretty` above.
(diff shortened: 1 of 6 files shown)


website/content/api-docs/client.mdx
```diff
@@ -20,15 +20,190 @@ When accessing the endpoints via the server, if the desired node is ambiguous
...
+
+### Sample Request
+
+```shell-session
+$ nomad operator api /v1/client/metadata
+```
+
+```shell-session
+$ curl http://127.0.0.1:4646/v1/client/metadata
+```
```
shoenig, line 72: probably unnecessary but it doesn't hurt (we wouldn't want curl alternatives in every case we use opapi in an example?)

shoenig (approved):
LGTM!

website/content/api-docs/client.mdx
schmichael: We took a vote internally:

nomad operator api - :six: 
curl - :five: 

I'll change this page's examples but not do a big old s/curl/noa/g in this PR

PR #16105, opened 2023-02-09
cli: fix use of the sanitized method type for the login command.

jrasell:
When an auth method was not supplied and the OIDC type was given in lower case, the CLI was not matching the default method due to casing and responded with a confusing user message.

This change fixes the above problem, along with making use of the santized type easier.


pkazmierczak (approved):
Thanks @jrasell!

shoenig (approved):
LGTM

PR #16107, opened 2023-02-09
tests: reduce timeout on test-api

shoenig:
Sometimes GHA gets wedged, and there is no need to waste 30 minutes
on a job that usually takes < 3.

e.g. stuck loading module cache, which is out of our control
https://github.com/hashicorp/nomad/actions/runs/4135098558/jobs/7147128419


lgfa29 (approved):
I wish there were more analytic to help us measure these things

Issue #16098, opened 2023-02-08
[ACL] bootstrapping with an old NOMAD_TOKEN throws a 403


philrenaud:
In 1.5.beta, running `nomad acl bootstrap` throws a `Error bootstrapping: Unexpected response code: 403 (Permission denied)` error if I had a previously-set `NOMAD_TOKEN` value set.

I can get past this by running `export NOMAD_TOKEN=""` or similar, but prior to this version, the existence of a NOMAD_TOKEN wasn't relevant to bootstrapping that I know of.

PR #16108, opened 2023-02-09
acl: never return auth errors for `ACL.Bootstrap` RPC

tgross:
Fixes https://github.com/hashicorp/nomad/issues/16098 introduced in 1.5.0-beta.1

In #15901 we introduced pre-forwarding authentication for RPCs so that we can grab the identity for rate metrics. The `ACL.Bootstrap` RPC is an unauthenticated endpoint, so any error message from authentication is not particularly useful. This would be harmless, but if you try to bootstrap with your `NOMAD_TOKEN` already set (perhaps because you were talking to another cluster previously from the same shell session), you'll get an authentication error instead of just having the token be ignored. This is a regression from the existing behavior, so have this endpoint ignore auth errors the same way we do for every other unauthenticated endpoint (ex `Status.Peers`)


pkazmierczak (approved):
Makes sense to me @tgross 👍

PR #16111, opened 2023-02-09
[core] Do not start the plugin loader on non-clients

angrycub:
The plugin loader loads task and device driver plugins which are not
used on server nodes.

### DIFF
--- a/command/agent/plugins.go
+++ b/command/agent/plugins.go
@@ -10,6 +10,10 @@ import (
 
 // setupPlugins is used to setup the plugin loaders.
 func (a *Agent) setupPlugins() error {
+	if a.GetConfig().Client == nil || !a.GetConfig().Client.Enabled {
+		a.logger.Debug("skipping plugin setup because node is not a client")
+		return nil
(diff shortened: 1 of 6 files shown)


tgross (approved):
LGTM

command/agent/plugins.go
```diff
@@ -10,6 +10,10 @@ import (
 
 // setupPlugins is used to setup the plugin loaders.
 func (a *Agent) setupPlugins() error {
+	if a.GetConfig().Client == nil || !a.GetConfig().Client.Enabled {
+		a.logger.Debug("skipping plugin setup because node is not a client")
+		return nil
+	}
```
schmichael, line 16: Instead of an early return what if we just called this method from within `setupClient`? I just find the conditions a bit confusing because they should always match but may not depending on _when_ you check them.

schmichael (approved):
Had another idea for how to structure it, but thank you for doing this! Tests are very appreciated.

schmichael (approved):

PR #16114, opened 2023-02-09
[chore] Move TestUtil_loadVersionControlGlobalConfigs into build flagged file

angrycub:
Since this test tests a function that in a file controlled by a build flag, when not on Linux, the implementation is not found by the Go language server and makes IDEs sad.


tgross (approved):
LGTM

shoenig (approved):

PR #16115, opened 2023-02-09
ui: hide custom label if not text is provided

lgfa29:
Running a simple `nomad agent -dev` results in an empty label because `this.system.agent.config.UI.Label` is an object shaped like this:

```
{
  "BackgroundColor": "",
  "Text": "",
  "TextColor": ""
}
```
image

After the change the label is not displayed unless a text is provided:
image
image

I wasn't able to figure out how to modify the label config to write a proper test 😅


philrenaud (approved):
Good catch, thank you!

PR #16120, opened 2023-02-09
tests: don't mutate global structs in core scheduler tests

tgross:
Some of the core scheduler tests need the maximum batch size for writes to be smaller than the usual `structs.MaxUUIDsPerWriteRequest`. But they do so by unsafely modifying the global struct, which creates test flakes in other tests.

Modify the functions under test to take a batch size parameter. Production code will pass the global while the tests can inject smaller values. Turn the `structs.MaxUUIDsPerWriteRequest` into a constant, and add a semgrep rule for avoiding this kind of thing in the future.

Discovered while trying to debug https://github.com/hashicorp/nomad/pull/16112, but this should get backported to all supported versions whereas that will be only to 1.4.x

### DIFF
--- a/nomad/structs/uuid.go
+++ b/nomad/structs/uuid.go
@@ -3,5 +3,5 @@ package structs
 // MaxUUIDsPerWriteRequest is the maximum number of UUIDs that can be included
 // within a single write request. This is to ensure that the Raft message does
 // not become too large. The resulting value corresponds to 0.25MB of IDs or
-// 7282 UUID strings.
-var MaxUUIDsPerWriteRequest = (1024 * 256) / 36
+// 7281 UUID strings.
+const MaxUUIDsPerWriteRequest = 7281 // (1024 * 256) / 36
--- a/.semgrep/protect_globals.yml
+++ b/.semgrep/protect_globals.yml
@@ -1,0 +1,13 @@
+rules:
+  - id: "no-overriding-struct-globals"
+    patterns:
+      - pattern: |
+          structs.$A = ...
+    message: "Mutating global structs is never safe"
+    languages:
+      - "go"
+    severity: "ERROR"
+    fix: " "
+    paths:
+      # including tests!
+      include: ["*"]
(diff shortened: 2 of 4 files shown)


tgross:
New semgrep rule from this PR, without the patch to the tests:

```
$ semgrep scan -c ./.semgrep/protect_globals.yml
METRICS: Using configs from the Registry (like --config=p/ci) reports pseudonymous rule metrics to semgrep.dev.
To disable Registry rule metrics, use "--metrics=off".
Using configs only from local files (like --config=xyz.yml) does not enable metrics.

More information: https://semgrep.dev/docs/metrics

Scanning 1683 files.

Findings:

  nomad/core_sched_test.go
     semgrep.no-overriding-struct-globals
        Mutating global structs is never safe

       1853┆ structs.MaxUUIDsPerWriteRequest = 2
       1896┆ structs.MaxUUIDsPerWriteRequest = 2
       1930┆ structs.MaxUUIDsPerWriteRequest = 2
       1932┆ structs.MaxUUIDsPerWriteRequest = originalMaxUUIDsPerWriteRequest

Some files were skipped or only partially analyzed.
  Scan was limited to files tracked by git.
  Partially scanned: 3 files only partially analyzed due to a parsing or internal Semgrep error
  Scan skipped: 1 files larger than 1.0 MB, 1 files matching .semgrepignore patterns
  For a full list of skipped files, run semgrep with the --verbose flag.

Ran 1 rule on 1683 files: 4 findings.
```

nomad/structs/uuid.go
```diff
@@ -4,4 +4,4 @@ package structs
 // within a single write request. This is to ensure that the Raft message does
 // not become too large. The resulting value corresponds to 0.25MB of IDs or
 // 7282 UUID strings.
-var MaxUUIDsPerWriteRequest = (1024 * 256) / 36
+const MaxUUIDsPerWriteRequest = 7822 // (1024 * 256) / 36
```
lgfa29, line 7:
```suggestion
const MaxUUIDsPerWriteRequest = 7821 // (1024 * 256) / 36
```

Based on the errors in `TestHTTP_EvalsDelete` I guess this was being rounded down?

lgfa29 (approved):

nomad/structs/uuid.go
tgross: Good catch. The truncating of integer division (to 0!) is why this caused the failure in #16112 in the first place!

PR #16137, opened 2023-02-12
fix: correct Makefile to run unit tests locally

farbodahm:
I wanted to run tests locally to start contributing to Nomad; So as [current documentation mentions](https://github.com/hashicorp/nomad/blob/main/contributing/README.md?plain=1#L49), I had to run `make test` to run the tests.
But after checking the [GNUMakefile](https://github.com/hashicorp/nomad/blob/main/GNUmakefile), found that it doesn't exist any more and documentation should be updated.

Also after trying to run the unit tests with `make test-nomad`, I found that it doesn't work properly as `GOTEST_GROUP` is not set in local run and it gets set in CI like [this](https://github.com/hashicorp/nomad/blob/main/.github/workflows/test-core.yaml#L110).

So this PR:

1. Fix the documentation regarding how to run the unit tests
2. Fix the Makefile so anyone can run the unit tests localy

I'm not sure if it was the correct place to fix the running unit tests locally; So please let me know if anything is needed.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -32,6 +32,11 @@ ifndef NOMAD_NO_UI
 GO_TAGS := ui $(GO_TAGS)
 endif
 
+#GOTEST_GROUP is set in CI pipelines. We have to set it for local run.
+ifndef GOTEST_GROUP
+GOTEST_GROUP := nomad client command drivers quick
+endif
+
 # tag corresponding to latest release we maintain backward compatibility with
 PROTO_COMPARE_TAG ?= v1.0.3$(if $(findstring ent,$(GO_TAGS)),+ent,)
 
@@ -408,3 +413,15 @@ ec2info: ## Generate AWS EC2 CPU specification table
 .PHONY: cl
 cl: ## Create a new Changelog entry
 	@go run -modfile tools/go.mod tools/cl-entry/main.go
+
+.PHONY: test
+test: GOTEST_PKGS := $(foreach g,$(GOTEST_GROUP),$(shell go run -modfile=tools/go.mod tools/missing/main.go ci/test-core.json $(g)))
+test: ## Use this target as a smoke test
+	@echo "==> Running Nomad smoke tests on groups: $(GOTEST_GROUP)"
+	@echo "==> with packages: $(GOTEST_PKGS)"
+	gotestsum --format=testname --packages="$(GOTEST_PKGS)" -- \
+		-cover \
+		-timeout=20m \
+		-count=1 \
+		-tags "$(GO_TAGS)" \
+		$(GOTEST_PKGS)
--- a/contributing/README.md
+++ b/contributing/README.md
@@ -44,10 +44,14 @@ Developing without Vagrant
    ```sh
    $ [ "$(ulimit -n)" -lt 1024 ] && ulimit -n 1024
    ```
-1. Verify you can run tests
+1. Verify you can run smoke tests
    ```sh
    $ make test
    ```
+   **Note:** You can think of this as a `smoke` test which runs a subset of
+   tests and some may fail because of `operation not permitted` error which
+   requires `root` access. You can use `go test` to test the specific subsystem
+   which you are working on and let the CI run rest of the tests for you.
 
 Running a development build
 ---


tgross (commented):
Hi @farbodahm! Thanks for opening this PR, we definitely missed that when reworking CI recently! I suspect the core Nomad developers are mostly running unit tests via `go test` directly, over subsets of tests because the test suite is so large, and then relying on CI to exercise the rest.

That being said, so long as we've got two different code paths here, maybe this is an opportunity to have a separate `make test` and `make test-nomad` that do different things. For example, when running unit tests locally for development, you generally don't want to use `--rerun-fails=3`. The approach you've got here also ends up skipping the `api` package tests which would need to get run through `test-nomad-module`.

farbodahm:
@tgross Thanks for the answer; it also make sense to me to do so, As I saw how much it takes to run the tests with the current approach.

What is your suggestion on this? What tests should be run for local development? I will be happy to help on this.

tgross:
Hi @farbodahm! So I had a conversation with the team about this and it's pretty clear we all use `go test` directly (or an IDE runner that does that under the hood) because of the complicated requirements around the test graph. For example, some tests require running as `root` because they're not really "unit" tests and touch the local operating system. We want to catch those and `t.Skip` them but haven't got around to it yet.

So what I'd suggest we do here is have a new `make test` target that runs over `nomad client command drivers quick` without retries, and update the docs to note that this is a "smoke test" that runs a subset of the tests, and that you'll want to use `go test` to test the specific subsystem they're working on. That'll at least get us into better shape than we are now and make it easier for folks to get started.

tgross (changes requested):
Hi @farbodahm! So I had a conversation with the team about this and it's pretty clear we all use `go test` directly (or an IDE runner that does that under the hood) because of the complicated requirements around the test graph. For example, some tests require running as `root` because they're not really "unit" tests and touch the local operating system. We want to catch those and `t.Skip` them but haven't got around to it yet.

So what I'd suggest we do here is have a new `make test` target that runs over `nomad client command drivers quick` without retries, and update the docs to note that this is a "smoke test" that runs a subset of the tests, and that you'll want to use `go test` to test the specific subsystem they're working on. That'll at least get us into better shape than we are now and make it easier for folks to get started.

farbodahm:
@tgross Thank you for the detailed explanation Tim, I applied the changes you requested for; Please let me know if anything else is needed.

Also I can start working on the part that you mentioned regarding catching the tests which requires root access and skip them if they are not running with root access; Please let me know if that's ok; Also we can discuss the details which you need for those.

Thanks.

GNUmakefile
```diff
@@ -408,3 +413,16 @@ ec2info: ## Generate AWS EC2 CPU specification table
 .PHONY: cl
 cl: ## Create a new Changelog entry
 	@go run -modfile tools/go.mod tools/cl-entry/main.go
+
+# Use this target as a smoke test
+.PHONY: test
+test: GOTEST_PKGS := $(foreach g,$(GOTEST_GROUP),$(shell go run -modfile=tools/go.mod tools/missing/main.go ci/test-core.json $(g)))
+test:
```
tgross, line 420:
```suggestion
.PHONY: test
test: GOTEST_PKGS := $(foreach g,$(GOTEST_GROUP),$(shell go run -modfile=tools/go.mod tools/missing/main.go ci/test-core.json $(g)))
test: ## Use this target as a smoke test
```

This convention for "docstrings" lets the target get picked up by the silly little awk script we have in `make help`

contributing/README.md
```diff
@@ -44,10 +44,11 @@ Developing without Vagrant
    ```sh
    $ [ "$(ulimit -n)" -lt 1024 ] && ulimit -n 1024
    ```
-1. Verify you can run tests
+1. Verify you can run smoke tests
    ```sh
    $ make test
    ```
+   **Note:** You can think of this as a `smoke` test which runs a subset of tests and some may fail because of `operation not permitted` error which requires `root` access. You can use `go test` to test the specific subsystem which you are working on and let the CI run rest of the tests for you.
```
tgross, line 51: We're unfortunately not 100% consistent with it everywhere, but we try to hard-wrap long lines in markdown like this:

```suggestion
   **Note:** You can think of this as a `smoke` test which runs a subset of
   tests and some may fail because of `operation not permitted` error which
   requires `root` access. You can use `go test` to test the specific subsystem
   which you are working on and let the CI run rest of the tests for you.
```

farbodahm:
@tgross Thank you so much for the review, Applied the changes requested.

And thank you for the head start for applying the changes on the tests as you asked for. I will start working on it and mention you in its PR if I need your help ;)

tgross (approved):
LGTM! Thanks!

PR #16145, opened 2023-02-13
dev: remove use of cfssl and use Nomad CLI for TLS certs.

jrasell:
should be timed with https://github.com/hashicorp/tutorials/pull/1140


lgfa29 (approved):
Noting here my dislike for `foo` as a region, but that's not part of this PR 😄

PR #16165, opened 2023-02-13
docs: add variable specification docs

schmichael:


tgross (approved):
LGTM. I know it's redundant but should we include the [Restrictions](https://developer.hashicorp.com/nomad/docs/commands/var/put#restrictions) block in this new page as well? Just to make sure folks see it wherever they land first.

PR #16166, opened 2023-02-13
Fix: missing Node Status "disconnected" in API

pierreca:
Looks like a new node status was introduced with nomad 1.3.x but this isn't reflected in the API


tgross (approved):
Thanks @pierreca!

Issue #16170, opened 2023-02-14
nomad fmt should return non zero exit code when changing files


jrnijboer:
### Nomad version
Nomad v1.4.3 (f464aca721d222ae9c1f3df643b3c3aaa20e2da7)

### Operating system and Environment details
VERSION="20.04.5 LTS (Focal Fossa)"
running under wsl2 in Windows 11

### Issue
There are 2 problems I have with nomad fmt. I'm not actually sure if they are bugs, or if they should be feature requests, so for now I'll call them "issues".

The first issue is that 'nomad fmt' returns a zero exit code when it has formatted files. I believe this should return a non zero exit code, this makes it easier to check if nomad files weren't properly formatted and have been changed (nice for pre-commit hooks). This is also how terraform fmt works. I think it would be nice for users to have some consistency between the Hashicorp products.

The other issue is that nomad fmt -check actually alters files (by formatting them). According to https://developer.hashicorp.com/nomad/docs/commands/fmt?product_intent=nomad#check this should only check the files if they are valid HCL. Running nomad fmt -check should not change files. However, I also believe this also should work more like terraform fmt -check, so not only should it check if the file is valid HCL, but also if it is properly formatted. So, return zero exit code if valid HCL and properly formatted, else return non zero exit code (when HCL invalid or not properly formatted) to make it consistent with other Hashicorp products (see [terraform fmt -check](https://developer.hashicorp.com/terraform/cli/commands/fmt#check)) 

### First issue (non zero exit code when nomad has changed files)
```
mkdir -p /tmp/nomad
rm -rf /tmp/nomad/*
git init
cat <<EOF >> job.hcl
job "nginx" {
    datacenters = ["dc1"]
    group "nginx" {
        task "container" {
            driver = "docker"
            config {
                image = "nginx-alpine"
            }
        }
    }
}
EOF
git add .
git status
nomad fmt job.hcl
echo "exit code: $?"
git status
```
the last git status shows that the file has been changed, but the exit code from nomad is zero

### Second issue (running with -check does change files)
```
mkdir -p /tmp/nomad
rm -rf /tmp/nomad/*
git init
cat <<EOF >> job.hcl
job "nginx" {
    datacenters = ["dc1"]
    group "nginx" {
        task "container" {
            driver = "docker"
            config {
                image = "nginx-alpine"
            }
        }
    }
}
EOF
git add .
git status
nomad fmt -check job.hcl
echo "exit code: $?"
git status
```
[...]

PR #16174, opened 2023-02-14
cli: `fmt -check` should return early on diff

tgross:
Fixes https://github.com/hashicorp/nomad/issues/16170

The `nomad fmt -check` command incorrectly writes to file because we didn't
return before writing the file on a diff. Fix this bug and update the command
internals to differentiate between the write-to-file and write-to-stdout code
paths, which are activated by different combinations of options and flags.

The docstring for the `-list` and `-write` flags is also unclear and can be
easily misread to be the opposite of the actual behavior. Clarify this and fix
up the docs to match.

This changeset also refactors the tests quite a bit so as to make the test
outputs clear when something is incorrect and avoid reusing the UI state.

### DIFF
--- a/command/fmt_test.go
+++ b/command/fmt_test.go
@@ -20,8 +20,6 @@ func TestFmtCommand(t *testing.T) {
 	const inSuffix = ".in.hcl"
 	const expectedSuffix = ".out.hcl"
 
-	tmpDir := t.TempDir()
-
 	tests := []struct {
 		name        string
 		testFile    string
@@ -59,6 +57,10 @@ func TestFmtCommand(t *testing.T) {
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
+			tc := tc
+			ci.Parallel(t)
+
+			tmpDir := t.TempDir()
 			inFile := filepath.Join("testdata", "fmt", tc.testFile+inSuffix)
 			expectedFile := filepath.Join("testdata", "fmt", tc.testFile+expectedSuffix)
 			fmtFile := filepath.Join(tmpDir, tc.testFile+".hcl")
--- a/command/fmt_test.go
+++ b/command/fmt_test.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/hashicorp/nomad/ci"
 	"github.com/mitchellh/cli"
+	"github.com/shoenig/test/must"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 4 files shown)


command/fmt_test.go
```diff
@@ -18,80 +19,115 @@ func TestFmtCommand(t *testing.T) {
...
+			expectWrite: true,
+			expectCode:  0,
+		},
+	}
 
-			input, err := os.ReadFile(inFile)
-			require.NoError(t, err)
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			inFile := filepath.Join("testdata", "fmt", tc.testFile+inSuffix)
```
angrycub, line 62: Maybe we could parallelize these cases? If so, we can scratch the tmpDir on line 23
```suggestion
			tc := tc
			ci.Parallel(t)
			tmpDir := t.TempDir()
			inFile := filepath.Join("testdata", "fmt", tc.testFile+inSuffix)
```

command/fmt_test.go
```diff
@@ -18,80 +19,115 @@ func TestFmtCommand(t *testing.T) {
...
 	const inSuffix = ".in.hcl"
 	const expectedSuffix = ".out.hcl"
-	tests := [][]string{
-		{"nomad", "-check"},
-		{"nomad", ""},
-		{"job", "-check"},
-		{"job", ""},
-	}
+
 	tmpDir := t.TempDir()
```
angrycub, line 23: (see later suggestion) Should create the tempdir in the test case if we're parallelizing them
```suggestion
```

website/content/docs/commands/fmt.mdx
```diff
@@ -46,9 +49,9 @@ client {
 }
 
 $ nomad fmt
-
 agent.hcl
-$ cat agent.hcl 
+
+$ cat agent.hcl
```
angrycub, line 54: Should be its own shell-session block

website/content/docs/commands/fmt.mdx
```diff
@@ -46,9 +49,9 @@ client {
 }
 
 $ nomad fmt
```
angrycub, line 51: Should be its own shell session block

angrycub (approved):
Needs a couple of small touches to the docs markup, and I shared some testing thoughts that are totally optional.

shoenig (approved):
LGTM

PR #16180, opened 2023-02-14
cgutil: handle panic from runc helper method

### LINKED ISSUE #16179: Nomad panic during RPM post install script execution.
I am adding nomad to a fedora coreos system of mine during the build process. I am not exactly sure in which kind of chroot or whatever an rpm-ostree build runs. Nevertheless the postinstall script of the nomad RPM fails like this:

```
nomad.post: panic: cannot statfs cgroup root: no such file or directory
nomad.post: 
nomad.post: goroutine 1 [running]:
nomad.post: github.com/opencontainers/runc/libcontainer/cgroups.IsCgroup2UnifiedMode.func1()
nomad.post: 	github.com/opencontainers/runc@v1.1.4/libcontainer/cgroups/utils.go:45 +0x1d9
nomad.post: sync.(*Once).doSlow(0x4b97dd?, 0x320c780?)
nomad.post: 	sync/once.go:74 +0xc2
nomad.post: sync.(*Once).Do(...)
nomad.post: 	sync/once.go:65
nomad.post: github.com/opencontainers/runc/libcontainer/cgroups.IsCgroup2UnifiedMode()
nomad.post: 	github.com/opencontainers/runc@v1.1.4/libcontainer/cgroups/utils.go:35 +0x31
nomad.post: github.com/hashicorp/nomad/client/lib/cgutil.init()
nomad.post: 	github.com/hashicorp/nomad/client/lib/cgutil/cgutil_linux.go:21 +0x17
```

I assume this is because the nomad postinstall script executes the following:
```
if [[ $(nomad version) == *+ent* ]]; then
    echo "
The following shall apply unless your organization has a separately signed Enterprise License Agreement or Evaluation Agreement governing your use of the software:
Software in this repository is subject to the license terms located in the software, copies of which are also available at https://eula.hashicorp.com/ClickThruELA-Global.pdf or https://www.hashicorp.com/terms-of-evaluation as applicable. Please read the license terms prior to using the software. Your installation and use of the software constitutes your acceptance of these terms. If you do not accept the terms, do not use the software.
"
fi
```
and `nomad version` might already panic? I think it would be nice if `nomad version` would run without to many checks and if that is not possible then maybe at least fix the postinstall script to hide errors?

shoenig:
This PR wraps the `cgroups.IsCgroup2UnifiedMode()` helper method from
runc in a `defer/recover` block because it might panic in some cases.

Upstream fix in: https://github.com/opencontainers/runc/pull/3745

Closes #16179

Backports as far back as 1.3.x, when support for cgv2 was added.

### DIFF
--- a/.changelog/16180.txt
+++ b/.changelog/16180.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cgutil: handle panic coming from runc helper method
+```
--- a/client/lib/cgutil/cgutil_linux.go
+++ b/client/lib/cgutil/cgutil_linux.go
@@ -18,7 +18,19 @@ import (
 // cgroups.v1
 //
 // This is a read-only value.
-var UseV2 = cgroups.IsCgroup2UnifiedMode()
+var UseV2 = safelyDetectUnifiedMode()
+
+// Currently it is possible for the runc utility function to panic
+// https://github.com/opencontainers/runc/pull/3745
+func safelyDetectUnifiedMode() (result bool) {
+	defer func() {
+		if r := recover(); r != nil {
+			result = false
+		}
+	}()
+	result = cgroups.IsCgroup2UnifiedMode()
+	return
+}
 
 // GetCgroupParent returns the mount point under the root cgroup in which Nomad
 // will create cgroups. If parent is not set, an appropriate name for the version


tgross (approved):
LGTM

PR #16193, opened 2023-02-15
Minor post-1.5-beta1 API, code, and docs cleanups

schmichael:
1. Remove `TODO` that isn't going to get addressed
2. Return an error when `parse` fails
3. Rename `cap -> policy` because `cap` is a builtin, and the variables we pass in are called `SomethingPolicy`
4. Fix a misspelling
5. Mention anonymous policy doesn't work for task api
6. Remove old `Using WI` section


shoenig (approved):
LGTM

PR #16195, opened 2023-02-16
client: add option to configure CSIVolumeClaimGCInterval

visweshs123:
The default setting is 5 minutes. If we lose an instance using an EBS volume with the CSI EBS volume plugin, it takes up to 5 minutes before the claims are GC'ed and we can get the job rescheduled. We would like to be able to configure this to be lower.

### DIFF
--- a/.changelog/16195.txt
+++ b/.changelog/16195.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+csi: Added server configuration for `csi_volume_claim_gc_interval`
+```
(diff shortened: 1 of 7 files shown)


tgross (approved):
Hi @visweshs123! This looks great, thanks for the PR. I've pushed an extra commit with changelog, docs update, a minor bit of validation, and a test fix. I'll get this merged once CI is green and this will ship in Nomad 1.5.0 GA

PR #16199, opened 2023-02-16
fix test flake for RPC TLS enforcement test

tgross:
The RPC TLS enforcement test was frequently failing with broken connections. The
most likely cause was that the tests started to run before the server had
started its RPC server. Wait until it self-elects to ensure that the RPC server
is up. This seems to have corrected the error; I ran this 3 times without a
failure (even accounting for `gotestsum` retries).

Also, fix a minor test bug that didn't impact the test but showed an incorrect
usage for `Status.Ping.`


shoenig (approved):
LGTM

lgfa29 (approved):
Thanks for digging into it!

PR #16205, opened 2023-02-16
docs: clarify sysbatch supports count

schmichael:
Also remove old version indicators.

Inspired by discussion arising from #16097 -- specifically that I forgot sysbatch supported `count > 1`!

Not sure we have a best practice on when to remove version spam from our docs. Removing "Since Nomad 0.9" seems safe, but is it too soon to remove the `Nomad 1.2` note?

### DIFF
--- a/website/content/docs/schedulers.mdx
+++ b/website/content/docs/schedulers.mdx
@@ -52,17 +52,15 @@ should be present on every node in the cluster. Since these tasks are
 managed by Nomad, they can take advantage of job updating,
 service discovery, and more.
 
-Since Nomad 0.9, the system scheduler will preempt eligible lower priority
-tasks running on a node if there isn't enough capacity to place a system job.
-See [preemption] for details on how tasks that get preempted are chosen.
+The system scheduler will preempt eligible lower priority tasks running on a
+node if there isn't enough capacity to place a system job.  See [preemption]
+for details on how tasks that get preempted are chosen.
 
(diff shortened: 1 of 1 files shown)


website/content/docs/schedulers.mdx
```diff
@@ -82,6 +80,11 @@ Sysbatch jobs are intended to run until successful completion, explicitly stoppe
 by an operator, or evicted through [preemption]. Sysbatch tasks that exit with an
 error are handled according to the job's [restart] block.
 
+ Like the `batch` scheduler, system batch jobs may have a [`count`] greater
+ than 1 to control how many instances are run. Instances that cannot be
+ immediately placed will be scheduled when resources become available,
+ potentially on a node that has already run another instance of the same job.
```
tgross, line 86: Missing a link for this (see https://nomad-4k5la2m65-hashicorp.vercel.app/nomad/docs/schedulers#system-batch). We could probably just say `count` here, although there's also the detail that `count` refers to the task group, not the job. So maybe:

```suggestion
 Like the `batch` scheduler, task groups in system batch jobs may have a `count` 
 greater than 1 to control how many instances are run. Instances that cannot be
 immediately placed will be scheduled when resources become available,
 potentially on a node that has already run another instance of the same job.
```

tgross (approved):

website/content/docs/schedulers.mdx
schmichael: Thanks!

PR #16210, opened 2023-02-17
acl: add validation to binding rule selector on upsert.

### LINKED ISSUE #16206: SSO expressions don't work when value has - in it
### Nomad version
1.5 beta 1

### Issue
When SSO claim values have a "-" in them, the binding-rule expressions fail to correctly match.

If I remove the "-" from both the role itself in our identity provider and the selector in Nomad it works as expected. I also tried with the dash but a shorter name, just to make sure it wasn't a length issue, and the problem persisted.

### Reproduction steps
auth-method command:
```bash
nomad acl auth-method create \
    -default \
    -name="azure" \
    -type="OIDC" \
    -max-token-ttl="1h" \
    -token-locality="local" \
    -config="@specs/auth-method/azure.json"
```

Auth method JSON:
```json
{
    "OIDCDiscoveryURL": "https://login.microsoftonline.com/.../v2.0",
    "OIDCClientID": "...",
    "OIDCClientSecret": "...",
    "OIDCScopes": [
        "openid"
    ],
    "BoundAudiences": [
    ],
    "AllowedRedirectURIs": [
        "https://example.com/ui/settings/tokens",
        "http://localhost:4649/oidc/callback"
    ],
    "ClaimMappings": {
    },
    "ListClaimMappings": {
        "roles": "roles"
    }
}
```

binding-rule command:
```bash
nomad acl binding-rule create \
    -auth-method="azure" \
    -selector="project-developer in list.roles" \
    -bind-type="policy" \
    -bind-name="project-developer" \
    -description="project developers"
```

#### Expected Result
I expect users that have a claim with a "roles" attribute that is a list, and that contains the string "project-developer" to be given the "project-developer" ACL policy.

#### Actual Result
```bash
nomad login -type="oidc" -method="azure"
Error performing login: Unexpected response code: 400 (no role or policy bindings matched)
```

jrasell:
The selector should be validated on upsert, otherwise an expression that is invalid will cause mapping failures at login which can be confusing to the user and operator. The docs have also been updated to include a little information about this and an example of dealing with claim names that include dashes.

This update targets an unreleased feature, so doesn't require a changelong entry.

Closes #16206

### DIFF
--- a/website/content/api-docs/acl/binding-rules.mdx
+++ b/website/content/api-docs/acl/binding-rules.mdx
@@ -127,7 +127,11 @@ The table below shows this endpoint's support for
 
 - `Selector` `(string: "")` - A boolean expression that matches against verified
   identity attributes returned from the auth method during login.  This is
-  optional and when not set, provides a catch-all rule.
+  optional and when not set, provides a catch-all rule. If set, it must be a
+  valid [go-bexpr][] expression; for example, a dash in the claim name will
+  require it to be encased in quotes and escaped such as
+  `"\"project-developer\" in list.roles"`.
+
 
 - `BindType` `(string: <required>)` - Adjusts how this binding rule is applied
   at login time. Valid values are `role`, `policy`, and `management`.
@@ -200,7 +204,10 @@ queries](/nomad/api-docs#blocking-queries) and [required ACLs](/nomad/api-docs#a
 
 - `Selector` `(string: "")` - A boolean expression that matches against verified
   identity attributes returned from the auth method during login.  This is
-  optional and when not set, provides a catch-all rule.
+  optional and when not set, provides a catch-all rule. If set, it must be a
+  valid [go-bexpr][] expression; for example, a dash in the claim name will
+  require it to be encased in quotes and escaped such as
+  `"\"project-developer\" in list.roles"`.
 
 - `BindType` `(string: "")` - Adjusts how this binding rule is applied at login
   time. Valid values are `role`, `policy`, and `management`.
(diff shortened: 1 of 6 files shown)


website/content/api-docs/acl/binding-rules.mdx
```diff
@@ -200,7 +204,10 @@ queries](/nomad/api-docs#blocking-queries) and [required ACLs](/nomad/api-docs#a
 
 - `Selector` `(string: "")` - A boolean expression that matches against verified
   identity attributes returned from the auth method during login.  This is
-  optional and when not set, provides a catch-all rule.
+  optional and when not set, provides a catch-all rule. If set, it must be a
+  valid [go-bexpr][] expression; for example, a dash in the claim name will
+  require it to be encased in quotes and escaped such as
+  `"\"project-developer\" in list.roles"`.
```
tgross, line 210: For my own clarity, the escaping is for the shell or JSON body, right? If I was using the command line can I do `-selector '"product-developer" in list.roles'` and that could get escaped when it's serialized to JSON?

tgross (approved):

website/content/api-docs/acl/binding-rules.mdx
jrasell: Yes exactly. I use the double quote escape example as it seems more obvious and visible what is going on to me.

```
$ nomad acl binding-rule create \
    -description "example binding rule" \
    -auth-method "auth0" \
    -bind-type "role" \
    -bind-name "eng-ro" \
    -selector '"product-developer" in list.roles'
ID           = 5a6dba47-0c6c-5d3c-98de-cbb06ed97c23
Description  = example binding rule
Auth Method  = auth0
Selector     = "\"product-developer\" in list.roles"
Bind Type    = role
Bind Name    = eng-ro
Create Time  = 2023-02-17 13:51:47.929953 +0000 UTC
Modify Time  = 2023-02-17 13:51:47.929953 +0000 UTC
Create Index = 17
Modify Index = 17
```

PR #16216, opened 2023-02-17
build/cli: Add BuildDate

gulducat:
Populate a new `BuildDate` value based on the latest commit at build time.

Using commit time instead of a simple `$(date)` to be consistent across multiple builds, as done in Consul and Vault.

`BuildDate` will be used in Enterprise to compare to license expiration time, but here in OSS it only changes the output of `nomad version`. I could be convinced to revert that if we don't want to futz with it here, but I can imagine cases where it would be nice for folks to have available.

Relates to: hashicorp/nomad-enterprise#855

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -3,15 +3,15 @@ PROJECT_ROOT := $(patsubst %/,%,$(dir $(abspath $(lastword $(MAKEFILE_LIST)))))
 THIS_OS := $(shell uname | cut -d- -f1)
 THIS_ARCH := $(shell uname -m)
 
-REPO = github.com/hashicorp/nomad
+GO_MODULE = github.com/hashicorp/nomad
 
 GIT_COMMIT := $(shell git rev-parse HEAD)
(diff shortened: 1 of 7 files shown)


GNUmakefile
```diff
@@ -3,10 +3,17 @@ PROJECT_ROOT := $(patsubst %/,%,$(dir $(abspath $(lastword $(MAKEFILE_LIST)))))
 THIS_OS := $(shell uname | cut -d- -f1)
 THIS_ARCH := $(shell uname -m)
 
+REPO = github.com/hashicorp/nomad
```
shoenig, line 6: will ent need a different value?

shoenig (approved):
LGTM!

GNUmakefile
gulducat: that crossed my mind, but nope!  actually though since we both bumped into it, I'll rename it -- it's not actually the repo, it's the go module, so `GO_MODULE` is more accurate (and I don't think conflicts with any other go env names).

Issue #16211, opened 2023-02-17
Nomad client tries to connect to serf IP address, not to RPC


Kamilcuk:
### Nomad version
1.4.3

### Operating system and Environment details
Fedora 29

### Issue
Nomad client does not use RPC IP to connect to Nomad servers. He uses serf IP. He should be using RPC IP.

### Reproduction steps
Configured the server with the following advertise block:

```
advertise {
  http = "172.29.248.59:4646"
  rpc  = "172.29.248.59:4647"
  serf = "10.120.18.153:4648"
}
```

Nomad servers have registered services in consul:

![image](https://user-images.githubusercontent.com/15679217/219639118-d77669c1-8a92-4b43-a2c1-8d77e9feb893.png)

```
$ curl 'http://localhost:8500/v1/catalog/service/nomad' | ..... | jq -c '.[]'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  5180    0  5180    0     0  1686k      0 --:--:-- --:--:-- --:--:-- 1686k
{"ID":"6172a751-c027-acc1-8c32-d1bb749760f9","Node":"--","Address":"172.29.248.59","Datacenter":"dev","TaggedAddresses":{"lan":"172.29.248.59","lan_ipv4":"172.29.248.59","wan":"172.29.248.59","wan_ipv4":"172.29.248.59"},"NodeMeta":{"consul-network-segment":""},"ServiceKind":"","ServiceID":"_nomad-server-3vwcahrnlg4szdzuqmlmufy5gqq6xvov","ServiceName":"nomad","ServiceTags":["rpc"],"ServiceAddress":"172.29.248.59","ServiceTaggedAddresses":{"lan_ipv4":{"Address":"172.29.248.59","Port":4647},"wan_ipv4":{"Address":"172.29.248.59","Port":4647}},"ServiceWeights":{"Passing":1,"Warning":1},"ServiceMeta":{"external-source":"nomad"},"ServicePort":4647,"ServiceSocketPath":"","ServiceEnableTagOverride":false,"ServiceProxy":{"Mode":"","MeshGateway":{},"Expose":{}},"ServiceConnect":{},"CreateIndex":5707990,"ModifyIndex":5707990}
[...]

PR #11895, opened 2022-01-21
fix wrong control plane discovery, when serf advertised address located in network, unreachable from clients

tantra35:
Suppose follow configuration, when we use nomad federation:

1. We have 2 nomad regions (`A` and `B`)
2. controlplane of regions located in common network: `10.168.0.0/16`, rpc and http addresses located in private networks, unique for both regions(`192.168.102.0/24` for region `A`, and `192.168.103.0/24` for region `B`)
3. nomad clients located only in private networks and can't communicate with controlplane throw `10.168.0.0/24`
4. on controllplane we advertise addreses like this:
    1. for region `A`  on first server
        ```
        advertise
        {
            http = "192.168.102.11"
            rpc =  "192.168.102.11"
            serf = "10.168.0.11"
        } 
        ```  
     1. for region `B`  on first server
        ```
        advertise
        {
            http = "192.168.103.11"
            rpc =  "192.168.103.11"
            serf = "10.168.1.11"
        } 
        ```

In such configuration nomad clients when discover controlplane by consul will get addresses from network `10.168.0.0/16`, and of course nothing will work, because  they can't reach this network

This patch try fix this situation by adding new RPC method `RpcPeers` which will return private RPC adresses instead of Serf

PR #16217, opened 2023-02-17
client: use RPC address and not serf after initial Consul discovery

tgross:
Fixes https://github.com/hashicorp/nomad/issues/16211

Nomad servers can advertise independent IP addresses for `serf` and `rpc`. Somewhat unexpectedly, the `serf` address is also used for both Serf and server-to-server RPC communication (including Raft RPC). The address advertised for `rpc` is only used for client-to-server RPC. This split was introduced intentionally in Nomad 0.8.

When clients are using Consul discovery for connecting to servers, they get an initial discovery set from Consul and use the correct `rpc` tag in Consul to get a list of adddresses for servers. The client then makes a `Status.Peers` RPC to get the list of those servers that are raft peers. But this endpoint is shared between servers and clients, and provides the address used for Raft.

Most of the time this is harmless because servers will bind on 0.0.0.0 anyways., But in topologies where servers are on a private network and clients are on separate subnets (or even public subnets), clients will make initial contact with the server to get the list of peers but then populate their local server set with unreachable addresses.

Cluster administrators can work around this problem by using `server_join` with specific IP addresses (or DNS names), because the `Node.UpdateStatus` endpoint returns the correct set of RPC addresses when updating the node. So once a client has registered, it will get the correct set of RPC addresses.

This changeset updates the client logic to query `Status.Members` instead of `Status.Peers`, and then extract the correctly advertised address and port from the response body.

---

Notes:
* The end-to-end testing required for this means it's going to just miss the 1.5.0 GA. This is probably safe to land in a patch release without worrying about backwards compatibility, because anyone with this topology is already using the `server_join` workaround described above.
* This also closes https://github.com/hashicorp/nomad/pull/11895, which unfortunately got abandoned because we had some trouble understanding exactly what the problem was.
* I've done some bench testing of this with a local development cluster, some extra IP addresses, and iptables rules. 
* I also stood it up on a real multi-host cluster using the changes in our E2E config from https://github.com/hashicorp/nomad/pull/16218 and everything seems to work as usual.
* This is outside the code path of the non-Consul cloud discovery features, but I also tested it by disabling server join via Consul (by just breaking the `consul` config block) and used [AWS EC2 auto-join via tags](https://developer.hashicorp.com/nomad/docs/configuration/server_join#amazon-ec2) and that works just fine as well.

### DIFF
--- a/.changelog/16217.txt
+++ b/.changelog/16217.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+client: Fixed a bug where clients used the serf advertise address to connect to servers when using Consul auto-discovery
+```
--- a/client/client.go
+++ b/client/client.go
@@ -2944,23 +2944,29 @@ DISCOLOOP:
 				mErr.Errors = append(mErr.Errors, err)
 				continue
 			}
-			var peers []string
-			if err := c.connPool.RPC(region, addr, "Status.Peers", rpcargs, &peers); err != nil {
+
+			// Query the members from the region that Consul gave us, and
+			// extract the client-advertise RPC address from each member
+			var membersResp structs.ServerMembersResponse
+			if err := c.connPool.RPC(region, addr, "Status.Members", rpcargs, &membersResp); err != nil {
 				mErr.Errors = append(mErr.Errors, err)
 				continue
 			}
-
-			// Successfully received the Server peers list of the correct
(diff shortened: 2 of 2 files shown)


lgfa29 (approved):
LGTM!

PR #16218, opened 2023-02-17
E2E: add multi-home networking to test infrastructure

tgross:
Add an Elastic Network Interface (ENI) to each Linux host, on a secondary subnet we have provisioned in each AZ. Revise security groups as follows:

* Split out client security groups from servers so that we can't have clients accidentally accessing serf addresses or other unexpected cross-talk.
* Add new security groups for the secondary subnet that only allows communication within the security group so we can exercise behaviors with multiple IPs.
* Drop inbound Consul/Vault access, as we're running those on HCP these days

This changeset doesn't include any Nomad configuration changes needed to take advantage of the extra network interface. I'll include those with testing for PR #16217. (And I've verified the cluster comes up normally with this change.)

### DIFF
--- a/e2e/terraform/.gitignore
+++ b/e2e/terraform/.gitignore
@@ -1,1 +1,2 @@
 *.zip
+uploads/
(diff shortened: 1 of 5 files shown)


shoenig (approved):
LGTM

schmichael (approved):
Worked for #16219

jrasell (approved):
LGTM!

PR #16219, opened 2023-02-18
Task API / Dynamic Node Metadata E2E test fixes

schmichael:
Fixes based on #16218 

Peeking at the commits is best, but there they're grouped by file neatly as well:

- TaskAPI middlewhere needs to turn RPC Permission Denied errors into Forbidden status codes to ensure acl.enabled=true behavior matches acl.enabled=false behavior. The e2e test covers this (and was failing as it should in our e2e environment!)
- Removed datacenters from all of my recent e2e jobspecs because I love the new default.
- I forgot to skip the Windows node in the Dynamic Node Metadata test. Fixed that.


tgross (approved):
LGTM!

You might want to rebase this on `main` as the commits are still showing for the infra changes even though the merge base was changed.

PR #16221, opened 2023-02-18
Fix unsorted Node value in `nomad operator raft list-peers` command

### LINKED ISSUE #13711: Consistently sort output of Nomad CLI
### Proposal

The output of `nomad server members` and `nomad node status` is sorted alphabetically by Name, but `nomad operator raft list-peers` does not seem to be sorted, or at least not by Node. I think it would make sense that CLI output is sorted by Name/Node consistently.

### Use-cases

This would make it easier to find a particular node within the output of the CLI.

### Attempted Solutions

`nomad operator raft list-peers | sort` kind of works, but it's not a great solution. :smile:

dttung2905:
Hi team,

This PR presents a small fix to sort the Node value in `nomad operator raft list-peers` command.
It will resolve #13711

### DIFF
--- a/.changelog/16221.txt
+++ b/.changelog/16221.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+cli: Sort output by Node name of the command `nomad operator raft list-peers`
+```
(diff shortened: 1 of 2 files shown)


tgross (approved):
LGTM, thanks @dttung2905!

PR #16222, opened 2023-02-18
Skip tests running locally without root access  

farbodahm:
[After a short discussion](https://github.com/hashicorp/nomad/pull/16137#pullrequestreview-1301517321) with @tgross, I learned that some of the tests will fail if we run them locally and without root access.

This PR will add a functionality to skip tests if the tests:
- Are **not** running in CI environment (they are running locally)
- and if they are **not** running with Root access

Also it will log that the test is skipped, so if user wants to run them, she/he has to run it with root access.

In the current behavior, They are failing by saying something like: `operation not permitted`. 

**Note:** I used `make test | grep -C 5 "operation not permitted"` to find which tests require Root access, But it may not contain all of the tests which require root access.

Please let me know if anything is needed; Thanks.

### DIFF
--- a/ci/skip_non_root.go
+++ b/ci/skip_non_root.go
@@ -1,0 +1,20 @@
+package ci
+
+import (
+	"os"
+	"strconv"
+	"syscall"
+	"testing"
+)
+
+// SkipTestWithoutRootAccess will skip test t if it's not running in CI environment
+// and test is not running with Root access.
+func SkipTestWithoutRootAccess(t *testing.T) {
+	ciVar := os.Getenv("CI")
+	isCI, err := strconv.ParseBool(ciVar)
+	isCI = isCI && err == nil
+
+	if !isCI && syscall.Getuid() != 0 {
+		t.Skipf("Skipping test %s. To run this test, you should run it as root user", t.Name())
+	}
+}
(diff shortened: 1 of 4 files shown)


tgross (commented):
Hi @farbodahm! The CI run [7312099268]https://github.com/hashicorp/nomad/actions/runs/4212805695/jobs/7312099268) is showing a test panic in `TestGenericNotifier` in `helper/broker/notify_test.go` that you'll want to look into.

farbodahm:
@tgross I double checked that the part which I changed didn't have any effect on the test which was failing. Also I ran the test with different parameters locally and it was fine; So I just re-run the CI (by force pushing) and everything is fine now.

tgross:
Hey @farbodahm I'm tied up in another mini-project I'm trying to ship by end of the week. I'll pick this back up next week. Just wanted to let you know that I haven't forgotten about this! :grinning:

farbodahm:
@tgross No worries man, Please take your time and thanks for mentioning!

tgross (changes requested):
Hi @farbodahm! This looks pretty good once the changes I've posted are resolved.

tgross (approved):
LGTM!

I'll have to wait a day to merge this -- we're in a very brief merge freeze on `main` because 1.5.0 GA is coming out shortly. But once that's live I can get this in. Thanks again, @farbodahm!

farbodahm:
@tgross No worries, Thank you so much for the review Tim.

PR #16233, opened 2023-02-20
docs: Fix missing `namespace`  query parameter in job doc

### LINKED ISSUE #14197: Namespace parameter missing in api documentation for stop job
The api docs for stop job do not reference the namespace parameter https://www.nomadproject.io/api-docs/jobs#stop-a-job

There's a bit of a suggestion from the ACL being at the namespace level, but for clarity and to save time for future users of this API it is probably worth explicitly documenting the namespace parameter for this endpoint.

dttung2905:
Hi team,
This MR will fix the missing `namespace` query parameter in job doc
Closes #14197


shoenig (approved):
LGTM, thanks @dttung2905 !

PR #16237, opened 2023-02-21
api: fix a panic and tweak some exported types

### LINKED ISSUE #16234: api: variables exported Err values are not error types
https://github.com/hashicorp/nomad/blob/main/api/variables.go#L15-L16

These have a prefix of Err suggesting they are of type `error`, but in fact they are of type `string` and do not satisfy the `error` interface. Trying to use them as `error` leads to compilation problems, e.g.

```
if errors.Is(err, api.ErrVariableNotFound) { // does not compile
   ...
}
```

### LINKED ISSUE #16235: api: variables api returns double pointer for GetItems
https://github.com/hashicorp/nomad/blob/main/api/variables.go#L164

`GetItems` returns a pointer to `VariableItems`, but the `VariableItems` type is already a pointer type (map). In practice, this means callers have to do a useless and unexpected deference to get to the actual value.

```
items, _, err := ns.client.Variables().GetItems(varPath, nil)
if err != nil {
   return value, err
}
s := (*items)[key] // double deference 
```

### LINKED ISSUE #16236: api: variables GetItems methods causes panic if variable does not exist
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x10c8a61]

goroutine 1 [running]:
github.com/hashicorp/nomad/api.(*Variables).GetItems(0xc000d7eeb8?, {0x7fffef67316b?, 0x1?}, 0x4b79476?)
	/home/shoenig/Work/go/nomad/api/variables.go:174 +0xe1
github.com/traefik/traefik/v2/pkg/provider/acme.get[...](0xc0013cda40, {0x7fffef67316b, 0xc}, {0x4b80e20, 0x7})
	/home/shoenig/Work/go/traefik/pkg/provider/acme/nomad_store.go:128 +0x77
github.com/traefik/traefik/v2/pkg/provider/acme.(*NomadStore).GetAccount(0xc0013cda40, {0xc000bce8ee, 0x2})
	/home/shoenig/Work/go/traefik/pkg/provider/acme/nomad_store.go:66 +0x2bc
```

This should return an `error`, but we'll have to fix https://github.com/hashicorp/nomad/issues/16234 first

shoenig:
This PR
 - fixes a panic in `GetItems` when looking up a variable that does not exist
 - deprecates `GetItems` in favor of `GetVariableItems` which avoids returning a pointer to a `map`
 - deprecates `ErrVariableNotFound` in favor of `ErrVariablePathNotFound` which is an actual `error` type
 - eliminates unused constant `ErrVariableMissingItems`
 - does some minor code cleanup to make linters happier

Fixes #16236
Fixes #16235
Fixes #16234

No backport - we don't tag the api module anyway

### DIFF
--- a/api/variables.go
+++ b/api/variables.go
@@ -12,8 +12,16 @@ import (
 )
 
 const (
-	ErrVariableNotFound     = "variable not found"
-	ErrVariableMissingItems = "variable missing Items field"
+	// ErrVariableNotFound was used as the content of an error string.
+	//
+	// Deprecated: use ErrVariablePathNotFound instead.
+	ErrVariableNotFound = "variable not found"
+)
+
+var (
+	// ErrVariablePathNotFound is returned when trying to read a variable that
+	// does not exist.
+	ErrVariablePathNotFound = errors.New("variable not found")
 )
 
 // Variables is used to access variables.
@@ -27,11 +35,10 @@ func (c *Client) Variables() *Variables {
 }
 
 // Create is used to create a variable.
-func (sv *Variables) Create(v *Variable, qo *WriteOptions) (*Variable, *WriteMeta, error) {
-
+func (vars *Variables) Create(v *Variable, qo *WriteOptions) (*Variable, *WriteMeta, error) {
 	v.Path = cleanPathString(v.Path)
 	var out Variable
-	wm, err := sv.client.put("/v1/var/"+v.Path, v, &out, qo)
+	wm, err := vars.client.put("/v1/var/"+v.Path, v, &out, qo)
 	if err != nil {
 		return nil, wm, err
 	}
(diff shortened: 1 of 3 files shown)


api/variables.go
```diff
@@ -12,8 +12,16 @@ import (
 )
 
 const (
-	ErrVariableNotFound     = "variable not found"
-	ErrVariableMissingItems = "variable missing Items field"
```
shoenig, line 16: note: this wasn't used so ... should be safe to just remove?

tgross: I checked in CT's [`nomad_var_get.go`](https://github.com/hashicorp/consul-template/blob/main/dependency/nomad_var_get.go) and [`nomad_var_list.go`](https://github.com/hashicorp/consul-template/blob/main/dependency/nomad_var_list.go#L80) and found no consumers either. _Technically_ breaking backcompat but I seriously doubt there are any real consumers.

api/variables.go
```diff
@@ -124,87 +124,98 @@ func (sv *Variables) Delete(path string, qo *WriteOptions) (*WriteMeta, error) {
...
 }
 
 // readInternal exists because the API's higher-level read method requires
 // the status code to be 200 (OK). For Peek(), we do not consider 403 (Permission
 // Denied or 404 (Not Found) an error, this function just returns a nil in those
 // cases.
-func (sv *Variables) readInternal(endpoint string, out **Variable, q *QueryOptions) (*QueryMeta, error) {
+func (vars *Variables) readInternal(endpoint string, out **Variable, q *QueryOptions) (*QueryMeta, error) {
+	// todo(shoenig): seems like this could just return a *Variable instead of taking
+	// in a **Variable and modifying it?
```
tgross, line 192: I went back to look at the original PR for this and don't see any discussion of it. Maybe @angrycub will remember why we did this?

schmichael (approved):

PR #16254, opened 2023-02-24
Accept Workload Identities for Client RPCs

schmichael:
Nomad 1.5.0-beta.1 was missing the ability to use workload identities for authorization when calling client RPCs like Dynamic Node Metadata. _Authentication_ using workload identity worked in the Task API, but if ACLs were enabled and the client RPCs tried to _authorize_ the workload identity... they had no idea how to turn a JWT into policies.

So that's what this does: Client RPCs can now turn JWTs -> Policies. The general approach is:

1. Treat JWTs like an ACL Token and look them up via `ACL.WhoAmI` instead of the ACLToken-specific `ACL.ResolveToken`
2. When using a JWT, call the new `ACL.GetClaimPolicies` RPC to retrieve policies instead of the ACLToken-specific `ACL.GetPolicies`

I tried stuffing policies into the `ACL.WhoAmI` response or even the `AuthenticatedIdentity` itself, but went with the separate RPC approach for to avoid having to maintain policies in those structs forever. Maybe we'll put policy names in the JWT someday (1.6 even?) and can use pre-1.5 ACL RPCs to resolve them? If so we just have to live with the new `GetClaimPolicies` RPC instead of having to live with deprecated fields in otherwise commonly used structs like `AuthenticatedIdentity`.

Put another way: adding a new RPC seemed like the lowest risk approach at the cost of performance.

Let me know if you want to discuss it over zoom as it's pretty subtle stuff that took me a lot of tries to settle on.

Sorry for the messy names/comments. I didn't want to devote too much time to prettying up if we didn't like the overall approach.

TODO
- [ ] Remove TODOs
- [ ] Fix and update tests
- [ ] Caching (or file an issue to add re-evaluate caching in 1.5.x)

### DIFF
--- a/client/acl.go
+++ b/client/acl.go
@@ -36,8 +36,9 @@ type clientACLResolver struct {
 	// policyCache is used to maintain the fetched policy objects
 	policyCache *structs.ACLCache[*structs.ACLPolicy]
 
+	//TODO(schmichael) rename?
 	// tokenCache is used to maintain the fetched token objects
-	tokenCache *structs.ACLCache[*structs.ACLToken]
+	tokenCache *structs.ACLCache[*structs.AuthenticatedIdentity]
 
 	// roleCache is used to maintain a cache of the fetched ACL roles. Each
 	// entry is keyed by the role ID.
@@ -48,10 +49,11 @@ type clientACLResolver struct {
 func (c *clientACLResolver) init() {
 	c.aclCache = structs.NewACLCache[*acl.ACL](aclCacheSize)
 	c.policyCache = structs.NewACLCache[*structs.ACLPolicy](policyCacheSize)
-	c.tokenCache = structs.NewACLCache[*structs.ACLToken](tokenCacheSize)
+	c.tokenCache = structs.NewACLCache[*structs.AuthenticatedIdentity](tokenCacheSize)
 	c.roleCache = structs.NewACLCache[*structs.ACLRole](roleCacheSize)
 }
 
+// TODO(schmichael) rename to ResolveBearerToken? or Authenticate?
 // ResolveToken is used to translate an ACL Token Secret ID into
 // an ACL object, nil if ACLs are disabled, or an error.
 func (c *Client) ResolveToken(secretID string) (*acl.ACL, error) {
@@ -64,6 +66,7 @@ func (c *Client) ResolveSecretToken(secretID string) (*structs.ACLToken, error) 
 	return t, err
 }
 
+// TODO(schmichael) rename?
 func (c *Client) resolveTokenAndACL(secretID string) (*acl.ACL, *structs.ACLToken, error) {
 	// Fast-path if ACLs are disabled
 	if !c.GetConfig().ACLEnabled {
@@ -115,10 +118,11 @@ func (c *Client) resolveTokenAndACL(secretID string) (*acl.ACL, *structs.ACLToke
 	return aclObj, token, nil
 }
 
+// TODO(schmichael) rename?
 // resolveTokenValue is used to translate a secret ID into an ACL token with caching
 // We use a local cache up to the TTL limit, and then resolve via a server. If we cannot
 // reach a server, but have a cached value we extend the TTL to gracefully handle outages.
-func (c *Client) resolveTokenValue(secretID string) (*structs.ACLToken, error) {
+func (c *Client) resolveTokenValue(secretID string) (*structs.AuthenticatedIdentity, error) {
 	// Hot-path the anonymous token
 	if secretID == "" {
 		return structs.AnonymousACLToken, nil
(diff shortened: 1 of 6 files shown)


client/acl.go
```diff
@@ -48,100 +49,121 @@ type clientACLResolver struct {
...
+			return nil, nil, err
+		}
+		policies = make([]*structs.ACLPolicy, 0, len(policyReply.Policies))
+		for _, p := range policyReply.Policies {
+			policies = append(policies, p)
+		}
 	}
 
 	// Resolve the ACL object
 	aclObj, err := structs.CompileACLObject(c.aclCache, policies)
```
tgross, line 131: Note for anyone else reviewing: this is where the ACL policy hash gets checked against the cache, so that we're always using the freshest known ACL policies.

client/acl.go
```diff
@@ -48,100 +49,121 @@ type clientACLResolver struct {
...
 	// Hot-path the anonymous token
 	if secretID == "" {
-		return structs.AnonymousACLToken, nil
+		return &structs.AuthenticatedIdentity{ACLToken: structs.AnonymousACLToken}, nil
 	}
 
 	// Lookup the token entry in the cache
 	entry, ok := c.tokenCache.Get(secretID)
 	if ok {
+		//TODO(schmichael) use a different cache for workload identities since we
```
tgross, line 151: The "expensive" cache of compiled `acl.ACL` objects is shared between tokens and WI because the claims just give us policies that compile down into the same `acl.ACL`. 

Maybe the thing that really wants to happen here is that the `tokenCacheSize` should be larger than the `aclCacheSize`, because it's likely that tokens are in a N:1 ratio with sets of policies. (To use a trivial example, lots of WI's will have a policy set of `[]*structs.ACLPolicy{}` or "no additional policies")

client/acl.go
```diff
@@ -48,100 +49,121 @@ type clientACLResolver struct {
...
 
 	// Lookup the token
-	req := structs.ResolveACLTokenRequest{
-		SecretID: secretID,
+	req := structs.GenericRequest{
 		QueryOptions: structs.QueryOptions{
+			AuthToken:  secretID,
 			Region:     c.Region(),
-			AllowStale: true,
+			AllowStale: false, //TODO(schmichael) I don't think we should allow this
```
tgross, line 162: Using a stale query for the tokens/claims themselves probably doesn't hurt, so long as the query for the policies is not stale (which looks to be the case).

tgross (commented):
The general approach here looks good to me @schmichael

client/acl.go
schmichael: Good call, doubled token cache size to 128 to accommodate workload IDs. The TTL is awfully short (30s), and workload ids give us the opportunity to be less naive about caching... but today they're bearer tokens as far as clients are concerned so I'm going to treat them like one.

.semgrep/rpc_endpoint.yml
```diff
@@ -89,6 +89,7 @@ rules:
           patterns:
             # Endpoints that are expected not to have authentication.
             - pattern-not: '"ACL.Bootstrap"'
+            - pattern-not: '"ACL.GetClaimPolicies"'
```
schmichael, line 92: Could a semgrep expert quick peek at these patterns and my RPC and make sure this is the right choice?

The intent of `ACL.GetClaimPolicies` is to do a lookup using the authenticated claim which is a fairly new pattern, so I think it just escapes semgreps matching.

client/acl.go
```diff
@@ -17,9 +17,10 @@ const (
 	// construction cost, so we keep the hot objects cached to reduce the ACL token resolution time.
 	aclCacheSize = 64
 
-	// tokenCacheSize is the number of ACL tokens to keep cached. Tokens have a fetching cost,
-	// so we keep the hot tokens cached to reduce the lookups.
-	tokenCacheSize = 64
+	// tokenCacheSize is the number of bearer tokens, ACL and workload identity,
+	// to keep cached. Tokens have a fetching cost, so we keep the hot tokens
+	// cached to reduce the lookups.
+	tokenCacheSize = 128
```
schmichael, line 23: Doubled the cache size since we're now stuffing workload identities in here. No science was used to pick this number (or I suspect the previous number). The [TTL is only 30s by default](https://developer.hashicorp.com/nomad/docs/configuration/acl#token_ttl), so I suspect this is more than enough to ensure all lookups are cached and only evicted by TTL.

e2e/e2eutil/acl.go
```diff
@@ -0,0 +1,49 @@
+package e2eutil
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	api "github.com/hashicorp/nomad/api"
```
jrasell, line 8:
```suggestion
	"github.com/hashicorp/nomad/api"
```

jrasell (approved):
lgtm; one tiny item, but not a blocker

.semgrep/rpc_endpoint.yml
tgross: Yeah that's right, it's similar to the `ACL.ResolveToken` case in that way. Arguably the name of this rule `rpc-potentially-unauthenticated` should be changed... once authentication and authorization got split out this is really testing for authorization.

tgross (approved):
LGTM!

e2e/e2eutil/acl.go
schmichael: What is wrong with my editor config that it did this?! :thinking:

.semgrep/rpc_endpoint.yml
schmichael: > once authentication and authorization got split out this is really testing for authorization.

Ah thanks for making this distinction explicit! It's quite the mental shift for me after years of not having to think about them independently.

I updated a comment but am too nervous to make a more invasive change in this PR. :sweat_smile:

PR #16256, opened 2023-02-24
CSI: fix potential state store corruptions

tgross:
The `CSIVolume` struct has references to allocations that are "denormalized"; we don't store them on the `CSIVolume` struct but hydrate them on read. Tests detecting potential state store corruptions (ref https://github.com/hashicorp/nomad/pull/16257) found two locations where we're not copying the volume before denormalizing:

* When garbage collecting CSI volume claims.
* When checking if it's safe to force-deregister the volume.

There are no known user-visible problems associated with these bugs but both have the potential of mutating volume claims outside of a FSM transaction. This changeset also cleans up state mutations in some CSI tests so as to avoid having working tests cover up potential future bugs.

### DIFF
--- a/.changelog/16256.txt
+++ b/.changelog/16256.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+csi: Fixed potential state store corruption when garbage collecting CSI volume claims or checking whether it's safe to force-deregister a volume
+```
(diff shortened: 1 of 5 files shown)


jrasell (approved):
LGTM!

Issue #15559, opened 2022-12-16
`api` package needs tests for force-leave


tgross:
While working on some internal testing which resulted in discovering #15552, @anth0d pointed out we're missing a test as noted in this very old TODO: https://github.com/hashicorp/nomad/blob/v1.4.3/api/agent_test.go#L137. Opening this issue to put that on our radar.

tgross:
Closed by #16260

PR #16260, opened 2023-02-25
Add new test case for force-leave

dttung2905:
Hi team,

This MR aims to add the test case for force-leave on exsisting node

It will solve issue #15559

### DIFF
--- a/api/agent_test.go
+++ b/api/agent_test.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"fmt"
 	"sort"
 	"strings"
 	"testing"
@@ -8,6 +9,7 @@ import (
 
 	"github.com/hashicorp/nomad/api/internal/testutil"
 	"github.com/shoenig/test/must"
+	"github.com/shoenig/test/wait"
 )
 
 func TestAgent_Self(t *testing.T) {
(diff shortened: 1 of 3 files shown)


api/agent_test.go
```diff
@@ -110,7 +110,31 @@ func TestAgent_ForceLeave(t *testing.T) {
...
+
+	time.Sleep(3 * time.Second)
+
+	membersAfter, err := a.MembersOpts(&QueryOptions{})
+
+	for _, node := range membersAfter.Members {
+		if node.Name == membersBefore.Members[1].Name {
+			must.Eq(t, node.Status, "leaving")
+		}
+	}
```
tgross, line 137: For this kind of thing, rather than waiting a static period of time, we can use [`shoenig/test/wait`](https://pkg.go.dev/github.com/shoenig/test@v0.6.1/wait) to wait until the assertion is true or the timeout expires. That lets us move the test along if the desired state is reached quickly.

For example, the following takes <3sec on my machine instead of >6sec. This adds up given the number of tests we run!

```suggestion
	f := func() error {
		membersAfter, err := a.MembersOpts(&QueryOptions{})
		if err != nil {
			return err
		}

		for _, node := range membersAfter.Members {
			if node.Name == membersBefore.Members[1].Name {
				if node.Status != "leaving" {
					return fmt.Errorf("node did not leave")
				}
			}
		}
		return nil
	}

	must.Wait(t, wait.InitialSuccess(
		wait.ErrorFunc(f),
		wait.Timeout(3*time.Second),
		wait.Gap(100*time.Millisecond),
	))
```

tgross (changes requested):
Thanks for this PR @dttung2905!

api/agent_test.go
dttung2905: Thanks alot for the suggestion and the code snippet. This is exactly what I'm looking for when writing this test. Totally agree that we can greatly reduce the test run time by making the `wait.Gap()` more aggressive instead of waiting for the whole duration.

tgross (approved):
LGTM!

PR #16263, opened 2023-02-26
build: set GOAMD64=v3 for amd64 builds

shoenig:
This PR sets GOAMD64=v3 for amd64 builds. By default, the Go compiler uses
GOAMD64=v1, which [at least partly] limits the generated machine code from
using modern instructions, particularly SSE3, SSE4.1/2, AVX, AVX2 which all
help with String-y operations, of which Nomad makes much use of.

By setting v3, the output binary will require processors make after ~2015,
(Haswell/Zen1 era) otherwise you'll get a SIGILL.

https://github.com/golang/go/wiki/MinimumRequirements#amd64
https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels


shoenig:
Let's punt this to 1.6 so it isn't rushed

shoenig:
@josegonzalez basically if `lscpu | grep Flags` contains `avx` (and your CPU is made by Intel/AMD) you are good to go - other architectures would be unaffected.

I think what we would end up doing is continue publishing builds for legacy hardware as attachments of GitHub releases - it's something we're already considering for other unsupported platforms like FreeBSD and friends

josegonzalez:
Heh my Minnowboard Max servers (which I use to run my local intel servers in homelab) don't support avx.

PR #16273, opened 2023-02-28
docs: add binding-rule selector escape example on Windows PS

jrasell:


website/content/docs/commands/acl/binding-rule/create.mdx
```diff
@@ -86,3 +87,25 @@ Modify Time  = 2022-12-20 11:15:22.582568 +0000 UTC
...
+Create a new ACL Binding Rule where the selector needs to be escaped on Windows
+machines via PowerShell:
+
+```shell-session
+$ nomad.exe acl binding-rule create \
+    -description "example binding rule" \
+    -auth-method "auth0" \
+    -bind-type "role" \
+    -bind-name "eng-ro" \
+    -selector="\`"project-developer\`" in list.roles"
```
pkazmierczak, line 100: we escape the backtick but not the `"`? is that correct?

pkazmierczak (approved):

PR #16274, opened 2023-02-28
[ui] Fix: Wildcard-datacenter system/sysbatch jobs stopped showing client links/chart

philrenaud:
Will now return all clients when `datacenters = ["*"]`, `["dc*"]`, etc. Otherwise will use the same filter as before.

Added some acceptance test cases for this as well. These tests will break without [the change to job-client-status.js](https://github.com/hashicorp/nomad/compare/16270-ui-job-status-in-client-panel-always-shows-all-zero-data-when-datacenters-%3D?expand=1#diff-58500552bd4b064fb387fb5c61302dcb7614aa7f4ae46536ff296f56b0268175R29-R32)

image


ui/app/utils/properties/job-client-status.js
```diff
@@ -26,7 +26,10 @@ export default function jobClientStatus(nodesKey, jobKey) {
 
       // Filter nodes by the datacenters defined in the job.
       const filteredNodes = nodes.filter((n) => {
-        return job.datacenters.indexOf(n.datacenter) >= 0;
+        return (
+          job.datacenters.includes('*') ||
```
lgfa29, line 30: `*` in datacenters is not like namespace, where `*` is a special wildcard to match all. In datacenters it's actually a glob pattern so, for example, `datacenters = ["dc*"]` would match `dc1` and `dc2` but not `us-east-1`. This check would return all clients, even the ones not matched by the pattern

philrenaud: Ahh thanks for keeping an eye out. I'll update to match on glob.

ChaiWithJai: Returning all clients here could unnecessarily throw a `403`.

ChaiWithJai: +1 to @lgfa29 -- you probably want to extract the logic from `_doesMatchPattern` into a utility and use it here.

ChaiWithJai (changes requested):

ui/app/utils/properties/job-client-status.js
lgfa29: Yup, that looks pretty good! If you need some test cases to match with the Go implementation:
https://github.com/ryanuber/go-glob/blob/master/glob_test.go

ui/app/abilities/variable.js
```diff
@@ -207,41 +208,41 @@ export default class Variable extends AbstractAbility {
...
+
+  //     if (!doesPathIncludeSubPattern) {
+  //       return false;
+  //     }
+
+  //     subPathToMatchOn = subPathToMatchOn.slice(0, idx + path.length);
+  //   }
+
+  //   return hasTrailingGlob || path.endsWith(lastPartOfPattern);
+  // }
```
ChaiWithJai, line 245: `blocking` `nit`: I don't think we want these here.

ui/app/abilities/variable.js
```diff
@@ -2,6 +2,7 @@
 import { computed, get } from '@ember/object';
 import { or } from '@ember/object/computed';
 import AbstractAbility from './abstract';
+import { default as doesMatchPattern } from 'nomad-ui/utils/match-glob';
```
ChaiWithJai, line 5: `nit` `non-blocking`: This `import` statement looks funky. You should be able to just write:

```
import doesMatchPattern from 'nomad-ui/utils/match-glob';
```

ui/app/utils/properties/job-client-status.js
```diff
@@ -26,7 +27,9 @@ export default function jobClientStatus(nodesKey, jobKey) {
 
       // Filter nodes by the datacenters defined in the job.
       const filteredNodes = nodes.filter((n) => {
-        return job.datacenters.indexOf(n.datacenter) >= 0;
+        return job.datacenters.filter((dc) => {
+          return !!matchGlob(dc, n.datacenter);
+        }).length;
```
ChaiWithJai, line 32: `nit` `non-blocking`:  might want to use `find` in the nested loop for an early exit and remove the need for `.length`.

```
const filteredNodes = nodes.filter((n) => job.datacenters.find((dc) => !!matchGlob(dc, n.datacenter));
```

ChaiWithJai (approved):
LGTM. There's a large block of commented code that should be removed.

PR #16278, opened 2023-02-28
cli: job restart command

### LINKED ISSUE #698: Add ability to restart all running tasks/allocs of a job
So I would love the ability to restart tasks, at the very least restart an entire job, but preferably single allocations. This is very useful for when a particular allocation or job happens to get in a bad state.

I am thinking something like `nomad restart <job>` or `nomad alloc-restart <alloc-id>`.

One of my specific use cases, is I have a cluster of rabbitmq nodes, and at some point one of the nodes gets partitioned from the rest of the cluster. I would like to restart that specific node (allocation in nomad parlance), or be able to preform a rolling restart to the entire cluster (job in nomad parlance).

Does this sound useful?

### LINKED ISSUE #10391: [feature request] Force(fast) job restart
### Proposal
It would be great if Nomad would have the ability to do a fast\force restart of the job.

### Use-cases
We need to do a quick restart during the incidents or on the development environment where we don't want to wait for a proper rolling update restart.

### Attempted Solutions
Currently, Nomad can restart job using the ~restart~ update stanza or alloc restart.

The problem with the update stanza is that I want to keep max_parallel small in the majority of the cases, like 1 to do rolling updates. If I'd change max_parallel to Job group count it will help but now I have my job with incorrect configuration and if I change it back - it will trigger restart one more time which is far from perfect.

With alloc restart it's just scary and hard to do programmatically. Seems like there is a lot of issues of restarting allocation with pre\post jobs(#9464 , #9841) and I'm uncertain how Consul Connect envoy side-car will handle that and if I need to do it in a specific order and how to get that order, etc, etc.

lgfa29:
Implement the new `nomad job restart` command that allows operators to restart allocations tasks or reschedule then entire allocation.

Restarts can be batched to target multiple allocations in parallel. Between each batch the command can stop and hold for a predefined time or until the user confirms that the process should proceed.

This implements the "Stateless Restarts" alternative from the original RFC
(https://gist.github.com/schmichael/e0b8b2ec1eb146301175fd87ddd46180). The original concept is still worth implementing, as it allows this functionality to be exposed over an API that can be consumed by the Nomad UI and other clients. But the implementation turned out to be more complex than we initially expected so we thought it would be better to release a stateless CLI-based implementation first to gather feedback and validate the restart behaviour.

Closes #698
Closes #10391


command/job_restart.go
```diff
@@ -0,0 +1,850 @@
...
+		return 1
+	}
+
+	restartAllocs := make([]string, 0, len(allocs))
+	for _, alloc := range allocs {
+		shortAllocID := limit(alloc.ID, c.length)
+
+		// Skip allocations that are not running.
+		allocRunning := alloc.ClientStatus == api.AllocClientStatusRunning && alloc.DesiredStatus == api.AllocDesiredStatusRun
+		if !allocRunning {
```
schmichael, line 238: Hm, should this be an `||` instead of an `&&` so the command works on allocs that **either** running or desired to be run?

For example:
1. If an alloc has `Client=Running, Desired=Stop`, it may be in the user still may want `-no-shutdown-delay` to kick the alloc immediately.
2. If an alloc has `Client=Pending, Desired=Run`, users may still want to reschedule it off a problematic client (that is failing to start it maybe?)

These are pretty narrow edge cases, so I'm not too worried about it. However, I think it's pretty safe to be aggressive here and try to kick more things instead of fewer things. The scheduler should never run too many allocations or resurrect dead allocs.

schmichael (commented):
Sorry, WIP. Will return to it.

command/job_restart.go
```diff
@@ -0,0 +1,850 @@
...
+func (c *JobRestartCommand) Name() string { return "job restart" }
+
+func (c *JobRestartCommand) Run(args []string) int {
+	c.ui = &cli.ConcurrentUi{Ui: c.Ui}
+
+	// Handle SIGINT to prevent accidental cancellations.
+	// activeCh is blocked while a signal is being handled.
+	activeCh := make(chan interface{})
+	sigsCh := make(chan os.Signal, 1)
+	signal.Notify(sigsCh, syscall.SIGINT)
```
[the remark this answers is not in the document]
lgfa29: I think the problem in this case is that the user may choose [not to cancel the command](https://github.com/hashicorp/nomad/pull/16278/files#diff-331d4a10a3b907d64a5d41fda14da628e6b34efd7f7a0095e7ba52782279ff10R789-R792), but by then the context would have cancelled already and so we would need to recreate it, which is kind of awkward 😬

command/job_restart.go
```diff
@@ -0,0 +1,850 @@
...
+}
+
+func (c *JobRestartCommand) Name() string { return "job restart" }
+
+func (c *JobRestartCommand) Run(args []string) int {
+	c.ui = &cli.ConcurrentUi{Ui: c.Ui}
+
+	// Handle SIGINT to prevent accidental cancellations.
+	// activeCh is blocked while a signal is being handled.
+	activeCh := make(chan interface{})
```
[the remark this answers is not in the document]
lgfa29: I keep forgetting all the new cool stuff 😄

command/job_restart.go
lgfa29: Ah good point. I first thought it wouldn't make sense to restart an alloc that is already marked to stop anyway, but it could be stuck or something else. I agree that changing this seems fine 👍

command/job_restart.go
```diff
@@ -0,0 +1,850 @@
...
+		c.ui.Output("No allocations to restart")
+		return 0
+	}
+
+	// Calculate absolute batch size based on the number of eligible
+	// allocations if the value provided is a percentage.
+	// Round values up to increase parallelism.
+	if c.batchSizePercent {
+		c.batchSize = int(math.Ceil(float64(len(restartAllocs)*c.batchSize) / 100))
+	}
```
lgfa29, line 298: Should we worry about large absolute batch sizes? I first thought of a percent batch size as a convenience option, but a `nomad job restart -reschedule -batch-size=100%` in a job with hundreds of allocations could cause havoc in a cluster 😅

command/job_restart.go
```diff
@@ -0,0 +1,850 @@
...
+	for {
+		newAlloc, qm, err := c.client.Allocations().Info(nextAllocID, q)
+		if err != nil {
+			return fmt.Errorf("Error retrieving replacement allocation %s: %v", limit(nextAllocID, c.length), err)
+		}
+		if newAlloc.ClientStatus == api.AllocClientStatusRunning {
+			break
+		}
+		q.WaitIndex = qm.LastIndex
+	}
```
lgfa29, line 768: Would it be worth to have a flag that skips this check for a faster restart? Or maybe something like `-wait-for-status=pending` so it doesn't always wait for `running`, which could take a while for some workloads.

command/job_restart.go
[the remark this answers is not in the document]
tgross: Starting the signal handler at the top of the command but not polling on `activeCh` creates some weird artifacts where if the user hits `Ctrl-C` early (perhaps the initial queries are hanging) the command will continue to run and log until we hit the main restart loop:

```
$ nomad job restart example
^C
Are you sure you want to stop the restart process?
Allocations not restarted yet will not be restarted. [y/N] y

Canceling job restart process
==> 2023-03-01T10:04:25-05:00: Restarting 2 allocations
```

I see where you're going with the signal handler here in terms of making it "friendly", and it looks like your implementation handles double-Ctrl-C correctly (immediate exit). But if we're going to do this, we should push the signal handler down to right before we enter the main restart loop so that we're only checking for signals on the parts we're actually prepared to interrupt.

command/job_restart.go
```diff
@@ -0,0 +1,849 @@
...
+			c.ui.Output(c.Colorize().Color(fmt.Sprintf(
+				"[bold]==> %s: Restarting %s batch of %d allocations[reset]",
+				formatTime(time.Now()),
+				humanize.Ordinal(batchNumber),
+				helper.Min(c.batchSize, remaining),
+			)))
+		}
+
+		// Restart allocation. Wrap the callback function to capture the
+		// allocID loop variable and prevent if from changing inside the
```
tgross, line 329:
```suggestion
		// allocID loop variable and prevent it from changing inside the
```

command/job_restart.go
```diff
@@ -0,0 +1,849 @@
...
+	}
+
+	c.ui.Output("\nAll allocations restarted successfully!")
+	return 0
+}
+
+// parseAndValidate parses and validates the arguments passed to the command.
+//
+// This function mutates the command and is not thread-safe so it must be
+// called only once and early in the command lifecycle.
```
tgross, line 398: I :heart: when we have these kinds of warnings around to protect us from our future self eagerly refactoring :grin:

website/content/docs/commands/job/restart.mdx
```diff
@@ -0,0 +1,206 @@
...
+The command can operate in batches and wait until all restarted or
+rescheduled allocations are running again before proceeding to the next batch.
+It is also possible to specify additional time to wait between batches.
+
+Allocations can be restarted in-place or rescheduled. When restarting
+in-place the command may target specific tasks in the allocations, restart
+only tasks that are currently running, or restart all tasks, even the ones
+that have already run. Allocations can also be targeted by groups and tasks.
+When both groups and tasks are defined only the tasks for the allocations of
+those groups are restarted.
```
tgross, line 31: It seems worth specifically calling out that we're only checking client status here and not health checks.

command/job_restart.go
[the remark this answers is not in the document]
lgfa29: Oh nice catch. Yeah, that signal handler is only useful for the long-running restart loop  so it makes sense to only start it when the loop starts 👍

command/job_restart.go
lgfa29: `Alloc.Stop`, which is used when `-reschedule` is set, [does create a Raft entry](https://github.com/hashicorp/nomad/blob/v1.4.4/nomad/alloc_endpoint.go#L310-L325) for each allocation, but the FSM operation [already supports batching multiple allocations in the same Raft transaction](https://github.com/hashicorp/nomad/blob/main/nomad/state/state_store.go#L3888-L3893), so that's something we can improve in the future.

command/job_restart.go
lgfa29: Oh that's a good point. Not waiting for a successful outcome can only lead to bad things 😅

command/job_restart.go
```diff
@@ -0,0 +1,849 @@
...
+	}
+
+	// Check for placement failures.
+	var mErr *multierror.Error
+	for tg, metrics := range eval.FailedTGAllocs {
+		failures := metrics.CoalescedFailures + 1
+		mErr = multierror.Append(mErr, fmt.Errorf(
+			"Task Group %q (failed to place %s):\n%s",
+			tg,
+			english.Plural(failures, "allocation", "allocations"),
```
schmichael, line 733: Neat!

website/content/docs/commands/job/restart.mdx
```diff
@@ -0,0 +1,206 @@
...
+- `-reschedule`: If set, allocations are stopped and rescheduled instead of
+  restarted in-place. Since the group is not modified the restart does not
+  create a new deployment, and so values defined in [`update`][] blocks, such
+  as [`max_parallel`][], are not taken into account. This option cannot be use
+  with `-task`.
+
+- `-task=<task-name>`: Specify the task to restart. Can be specified multiple
+  times. If groups are also specified the task must exist in at least of them.
+  If no task is set only tasks that are currently running are restarted. This
+  option cannot be used with `-all-tasks` or `-reschedule`.
```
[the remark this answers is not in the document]
lgfa29: Updated `-all-tasks` and `-task` to mention task lifecycle implications:

```
- `-all-tasks`: If set, all tasks in the allocations are restarted, even the
  ones that have already run, such as non-sidecar tasks. Tasks will restart
  following their [`lifecycle`][] order. This option cannot be used with
  `-task`.

- `-task=<task-name>`: Specify the task to restart. Can be specified multiple
  times. If groups are also specified the task must exist in at least one of
  them. If no task is set only tasks that are currently running are restarted.
  For example, non-sidecar tasks that already ran are not restarted unless
  `-all-tasks` is used instead. This option cannot be used with `-all-tasks` or
  `-reschedule`.
```

command/job_restart.go
schmichael: If the followup allocation terminates before this catches it running it will loop forever (well... until the alloc is GC'd and `Info(...)` returns error). Even watching the event stream and processing every update for the alloc isn't enough as clients and servers batch updates, so I think it's possible for an alloc to go directly from `pending -> failed`.

Seems like if we error on `ClientStatusTerminal()` with a recoverable error, the batch will be allowed to proceed unless `-fail-on-error` is set? That seems reasonable?

For allocs that are stuck in pending or can't be schedule (the FailedTGAllocs check above won't failures like an eval being canceled or failing), I'm not sure what to do. This will block forever in a fairly silent and mysterious way.

Perhaps we could slightly change `-batch-wait` to be more like `-batch-duration` and be the time the command allots to performing a batch of operations. These polling queries could log to inform the user if they want to stop before the next batch.

I'm not sure what the right answer here is, but I am worried trying this path only works without edge cases if you effectively recreate deployments in the CLI... at which point we should just create a helper for the `bump a meta{} var` trick.

website/content/docs/commands/job/restart.mdx
lgfa29: Added some content about it:

```
When rescheduling, the current allocations are stopped triggering the Nomad
scheduler to create replacement allocations that may be placed in different
clients. The command waits until the new allocations have client status `ready`
before proceeding with the next batch. Services health checks are not taken
into account.
```

lgfa29:
@tgross @schmichael sorry for push new code in https://github.com/hashicorp/nomad/pull/16278/commits/164e233b70114c58dfce3091bb722ee3fe3b748e, but I noticed a bug while testing some failure cases. 

More specifically, when rescheduling more than one allocation for the same task group, and the new allocations fail to be placed, the eval broker doesn't chain the evals and just cancels one of them without referencing the blocked eval or the failed metrics. This caused one of the goroutines to get stuck waiting the replacement to be running, but that would never happen.

<details>
<summary>Repro</summary>

Job
```hcl
job "example" {
  group "cache" {
    count = 2

    constraint {
      attribute = meta.test
      value     = "1"
    }

    task "redis" {
      driver = "docker"

      config {
        image = "redis:7"
      }
    }
  }

  group "cache2" {
    count = 2

    task "redis" {
      driver = "docker"

      config {
        image = "redis:7"
      }
    }
  }
}
```

Set a dynamic node metadata (🎉) and run the job:

```
$ nomad node meta apply test=1
$ nomad run example.nomad
```

Change the metadata and reschedule the job to cause placement failures. Use `-batch-size=3` to cause 2 allocations for the same task group to fail:

```
$ nomad node meta apply test=2
$ nomad job restart -reschedule -batch-size=3 example
```

One of the evals will be cancelled without information about the placement problems, causing the goroutine to block forever.
</details>

I changed the implementation so that it now checks for a blocked eval for the job and a successful replacement in parallel, so if any of these happen the goroutines will exit.

command/job_restart.go
lgfa29: One idea on error handling. Instead of having `-fail-on-error` have a flag `-on-error` that can be set to :
* `ask` (default for TTY sessions) behaving like `-batch-wait` and asking the user how to proceed if an error happens during a batch.
* `fail` that behaves like the current `-fail-on-error`
* `continue` (default for non-TTY?) that behaves like the current default and accumulates errors until the command exits.

command/job_restart.go
```diff
@@ -0,0 +1,955 @@
...
+			}
+
+			failures := eval.FailedTGAllocs[alloc.TaskGroup]
+			if failures != nil {
+				errCh <- fmt.Errorf("%s:\n%s",
+					jobRestartFailedToPlaceNewAllocation,
+					formatAllocMetrics(failures, false, strings.Repeat(" ", 4)),
+				)
+				return
+			}
```
tgross, line 794: This will throw an error and stop the restart if there are _any_ blocked evaluations for the job, including ones that were for unrelated issues. Consider the following scenario:

* `nomad job run`
* 5 allocs placed and running, 1 alloc queued (which means we have 1 blocked eval)
* `nomad job restart -reschedule`
* 1 running alloc is stopped, and the restart command exits here

We might be able to make this work if we used the wait index we got back from the initial stop, so that we're only detecting _new_ blocked evaluations that happened after the stop. That still wouldn't account for random other allocations that crash and fail to get placed. But I think that's an ok gap to have -- if that starts happening you probably want to stop what you're doing anyways.

command/job_restart.go
```diff
@@ -0,0 +1,955 @@
...
+	}
+}
+
+// monitorReplacementAlloc waits for the allocation to have a follow-up
+// placement and for the new allocation be running.
+//
+// Returns an error in errCh if anything goes wrong or nil when the new
+// allocation is running.
+func (c *JobRestartCommand) monitorReplacementAlloc(ctx context.Context, alloc *api.Allocation, errCh chan<- error) {
+	q := &api.QueryOptions{WaitIndex: 1}
```
tgross, line 806: Similar to the `monitorBlockedEvals`, I'd use the index we got back from stopping the alloc as the starting point. That ensures we're only getting new information here.

command/job_restart.go
```diff
@@ -0,0 +1,955 @@
...
+	}
+
+	return fmt.Sprintf(
+		"%d errors occurred:\n%s%s\n\n",
+		len(es), space, strings.Join(points, fmt.Sprintf("\n%s", space)))
+}
+
+// formatSliceOf returns a string with the length and the given singular or
+// plural noun depending on how many elements are present.
+func formatSliceOf(in []string, singular string, plural string) string {
```
tgross, line 940: Not in this PR but maybe for a follow-up, we have a bunch of these lovely formatting helpers all over the `command` package. It'd be nice if we could centralize them in a utils/helpers file so folks can find them and reuse them for UX consistency across commands.

website/content/docs/commands/job/restart.mdx
```diff
@@ -0,0 +1,206 @@
...
+layout: docs
+page_title: 'Commands: job restart'
+description: |
+  The job restart command is used to restart allocations for a job.
+---
+
+# Command: job restart
+
+The `job restart` command is used to restart or reschedule allocations for a
+particular job.
```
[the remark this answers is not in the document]
lgfa29: Added some description about their difference closer to the top.

```
Restarting the job calls the [Restart Allocation][api_alloc_restart] API
endpoint to restart the tasks inside allocations, so the allocations themselves
are not modified but rather restarted in-place.

Rescheduling the job uses the [Stop Allocation][api_alloc_stop] API endpoint to
stop the allocations and trigger the Nomad scheduler to compute new placements.
This may cause the new allocations to be scheduled in different clients from
the originals.
```

command/job_restart.go
```diff
@@ -0,0 +1,849 @@
...
+		return 1
+	}
+
+	// Check if the job exists.
+	// Avoid prefix matching to make sure we only restart the expected job.
+	job, _, err := c.client.Jobs().Info(c.jobID, nil)
+	if err != nil {
+		c.ui.Error(fmt.Sprintf("Error retrieving job %q: %s", c.jobID, err))
+		return 1
+	}
```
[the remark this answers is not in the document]
lgfa29: Extracted job prefix match login into `Meta` in https://github.com/hashicorp/nomad/pull/16306. Once that's merged I will rebase this PR and update the command 👍

command/job_restart.go
```diff
@@ -0,0 +1,849 @@
...
+	// Dedup tasks and groups.
+	c.groups = set.From(groups)
+	c.tasks = set.From(tasks)
+
+	return nil, 0
+}
+
+// validateGroupsAndTasks validates that the combination of groups and tasks
+// defined in the command are valid for the target job.
+func (c *JobRestartCommand) validateGroupsAndTasks(job *api.Job) error {
```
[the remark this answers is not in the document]
lgfa29: Hum....good point. Your [previous comment](https://github.com/hashicorp/nomad/pull/16278#discussion_r1122364290) made me think about job versions and then I arrived at the same problem you described here.

I think it makes sense to not have this validation, but I think I need to fix the allocation filter logic to take the allocation's job version into consideration.

The only thing this would catch is if you pass a task or group that doesn't exist in any job version, but that's probably OK? If none of your groups or tasks matches then we will just end-up with a list of empty allocations to restart, which is fine as well.

I will refactor this validation and alloc filtering logic to be version aware, thanks!

command/job_restart.go
lgfa29: > We might be able to make this work if we used the wait index we got back from the initial stop, so that we're only detecting new blocked evaluations that happened after the stop.

I had this in my first implementation, but that's how I found out that `api.Allocations().Stop()` doesn't return an index 😅 

I will fix the `api` package in a separate PR and merge it here so I can use the right index.

> That still wouldn't account for random other allocations that crash and fail to get placed. But I think that's an ok gap to have -- if that starts happening you probably want to stop what you're doing anyways.

Yeah, I thought about that as well, but couldn't come up with a good solution. Ideally the eval broker would make sure the eval chain is kept, but I found some situations where the eval doesn't reference any follow-up (like in [`MarkForCancel`](https://github.com/hashicorp/nomad/blob/v1.5.0/nomad/eval_broker.go#L1015-L1032)). I tried fixing some of those, but I was still getting some evals breaking the chain, so checking for _any_ blocked eval with placement failures was the best I could think of.

Stopping allocs while there are placement failures is problematic because it can leave you with zero allocations running, so I thought it would be better to stop the process.

tgross: Agreed :+1:

command/job_restart.go
```diff
@@ -0,0 +1,849 @@
...
+	// Check if passed groups and tasks are valid.
+	err = c.validateGroupsAndTasks(job)
+	if err != nil {
+		c.ui.Error(err.Error())
+		return 1
+	}
+
+	// Fetch all allocations for the job and filter out the ones that are not
+	// eligible for restart.
+	allocs, _, err := c.client.Jobs().Allocations(c.jobID, false, nil)
```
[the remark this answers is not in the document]
lgfa29: Ah good catch!

command/job_restart.go
```diff
@@ -0,0 +1,849 @@
...
+
+		// Skip allocations that don't have any of the requested tasks.
+		if c.tasks.Size() > 0 {
+			hasTask := false
+			for task := range alloc.TaskStates {
+				if c.tasks.Contains(task) {
+					hasTask = true
+					break
+				}
+			}
```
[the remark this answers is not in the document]
lgfa29: My variable naming was not great, but `alloc` here is a `api.AllocationListStub` which doesn't have `LookupTask` nor the `AllocatedResources`. 

I noticed while working on https://github.com/hashicorp/nomad/pull/16278/commits/7c7e7483c237e8e01e3e09b3b9f736857bb5649d that I actually don't need to request the full allocation so I have even less information to work with here 😅 

I encapsulated the search for a task in the [`HasTask`](https://github.com/hashicorp/nomad/blob/7c7e7483c237e8e01e3e09b3b9f736857bb5649d/command/job_restart.go#L891-L924) method that checks `TaskStates` first but then fallbacks to iterating over the job if not available.

command/job_restart.go
```diff
@@ -0,0 +1,849 @@
...
+	flags.StringVar(&batchSizeStr, "batch-size", "1", "")
+	flags.StringVar(&batchWaitStr, "batch-wait", "0s", "")
+	flags.BoolVar(&c.failOnError, "fail-on-error", false, "")
+	flags.BoolVar(&c.noShutdownDelay, "no-shutdown-delay", false, "")
+	flags.BoolVar(&c.reschedule, "reschedule", false, "")
+	flags.BoolVar(&c.verbose, "verbose", false, "")
+	flags.Var((funcVar)(func(s string) error {
+		groups = append(groups, s)
+		return nil
+	}), "group", "")
```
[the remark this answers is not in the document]
lgfa29: 1.5.1? 😅

command/job_restart.go
```diff
@@ -0,0 +1,849 @@
...
+			"Invalid -batch-size value %q: number value must be greater than zero",
+			batchSizeStr,
+		), 1
+	}
+
+	// Parse and validate -batch-wait.
+	if strings.ToLower(batchWaitStr) == jobRestartWaitAsk {
+		if !isTty() {
+			return fmt.Errorf(
+				"Invalid -batch-wait value %[1]q: %[1]q cannot be used when terminal is not interactive",
```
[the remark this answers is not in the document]
lgfa29: I know it exists but I always have to lookup the syntax 😬

command/job_restart.go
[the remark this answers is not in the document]
lgfa29: Refactored in https://github.com/hashicorp/nomad/pull/16278/commits/7c7e7483c237e8e01e3e09b3b9f736857bb5649d.

lgfa29:
`TestJobRestartCommand_shutdownDelay_reschedule` is flaky. I will refactor it so it's less time sensitive.

command/job_restart.go
schmichael: > > For allocs that are stuck in pending or can't be schedule (the FailedTGAllocs check above won't failures like an eval being canceled or failing), I'm not sure what to do. This will block forever in a fairly silent and mysterious way.
>
> Is there an easy way to simulate this? All my tests had FailedTGAllocs populated thinking

Canceled occurs when multiple evals are enqueued for the same job. Probably only occurs under scheduler load and/or when concurrent cluster activity causes the job to be eval'd. I think it could be exercised by setting NumSchedulers=0, kicking off a `restart -reschedule job-with-multiple-allocs`, and then setting NumSchedulers=1.

https://github.com/hashicorp/nomad/blob/b07af5761846a59ac91e4826c9a0f8e38c91f70b/nomad/eval_broker.go#L594-L610

I think failed evals are even weirder and rarer edge cases that occur when an eval has been nack'd too many times. Nacks occur when the worker/scheduler hits an error/timeout. Not sure that's worth trying to test. :grimacing: 

https://github.com/hashicorp/nomad/blob/b07af5761846a59ac91e4826c9a0f8e38c91f70b/nomad/worker.go#L417-L423


> One idea on error handling. Instead of having -fail-on-error have a flag -on-error that can be set to :

I think this works. `continue` is pretty scary as it allows for Nomad killing your job while you're not paying attention. Maybe the non-TTY default should be `fail` and require users to opt into the scary option?

command/job_restart.go
schmichael: (Technically doesn't _ensure_ new information because it could just timeout, but it helps!)

command/job_restart.go
lgfa29: I implemented this in https://github.com/hashicorp/nomad/pull/16278/commits/9552e091678a2e07aaac99d71dbd620e78e339ef but then I noticed that the blocking query index only controls when the request is unblocked, not the data returned, so we still receive the full list of evaluations.

I fixed this in https://github.com/hashicorp/nomad/pull/16278/commits/b0cc1234d39300bc7fcacf03d8188ed6e21d61cb by skipping evaluations crated before the job stop index.

I also had to remove the check for [`eval.Status != api.EvalStatusBlocked`](https://github.com/hashicorp/nomad/pull/16278/commits/9552e091678a2e07aaac99d71dbd620e78e339ef#diff-331d4a10a3b907d64a5d41fda14da628e6b34efd7f7a0095e7ba52782279ff10L729-L731) because when a new placement failure happens the new eval is cancelled with description `existing blocked evaluation exists for job "example"` so it never gets blocked.

command/job_restart.go
lgfa29: Yeah, I don't think setting the index here will give us only _new information_, it just controls _when_ the request unblocks. More details in https://github.com/hashicorp/nomad/pull/16278#discussion_r1125169155.

Since the allocation stop index is guaranteed to happen before `monitorReplacementAlloc` I'm not sure if there's a difference between using `1` or the stop index 🤔

command/job_restart.go
lgfa29: It turns out that I don't need this method anymore since we're not validating tasks and groups 😅 

But I will re-usability in mind for future commands.

command/job_restart.go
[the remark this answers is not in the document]
lgfa29: Implemented in 971ec5d4b55e01f719c0aa294c39b1f9f2066323.

lgfa29:
Ohh I think I broke ACL requirements for a few commands in https://github.com/hashicorp/nomad/pull/16306 😅 

To check job prefix it's necessary to have the `list-job` capability, which seems a bit much for a convenience functionality. I will modify the helper function to fallback to exact match with `Jobs().Info()` if `Jobs().PrefixList()` fails with a `403`.

command/job_restart.go
lgfa29: > I think it could be exercised by setting NumSchedulers=0, kicking off a restart -reschedule job-with-multiple-allocs, and then setting NumSchedulers=1.

Oh nice, I will try that next week. Thanks!

> `continue` is pretty scary as it allows for Nomad killing your job while you're not paying attention. Maybe the non-TTY default should be `fail` and require users to opt into the scary option?

Good point. I added a new [`-yes` flag](https://github.com/hashicorp/nomad/pull/16278/files#diff-7a75798ae34efc4d2d03c5950d047c4d5efe878631f5a43712ddcecbde45c696R99) to handle multi-region jobs and I think it can be useful here as well.

`ask` will be the default but fail validation on non-TTY session like `-batch-wait`. To always proceed even in case of errors you will need to add a `-yes`.

lgfa29:
> Ohh I think I broke ACL requirements for a few commands in #16306 😅
> 
> To check job prefix it's necessary to have the `list-job` capability, which seems a bit much for a convenience functionality. I will modify the helper function to fallback to exact match with `Jobs().Info()` if `Jobs().PrefixList()` fails with a `403`.

Fixed in https://github.com/hashicorp/nomad/pull/16380.

lgfa29:
> `TestJobRestartCommand_shutdownDelay_reschedule` is flaky. I will refactor it so it's less time sensitive.

Fixed by adding a [safety margin](https://github.com/hashicorp/nomad/pull/16278/files#diff-8d95103f4cdb1ce322cbcafc0a56d1dc561f2b5bd3e2d183ea450456bba77b0cR1064-R1066)

lgfa29:
@tgross @schmichael this PR is ready for another round of review. There were a few changes from the last round that, unfortunately, didn't fall neatly into separate commits, but these were the latest changes:

- Batch error handling is now done using the `-on-error` flag. Since we now have up to two questions that we ask users between batches (`-batch-wait` and `-on-error`) I refactor the logic to make sure we only ask them once.
- I realized that I could wrap the `Meta.Ui` in a `*cli.ConcurrentUi` instead of having two UIs in the command, which was confusing. But this required me to do some [`reflect` magic](https://github.com/hashicorp/nomad/pull/16278/files/971ec5d4b55e01f719c0aa294c39b1f9f2066323..8cdcc1411716ec4a60ca881b4e2559898c8c2748#diff-81269586aaa3bcddb4c844929b12481a6eb625350ba2e6673d1fed8e20b94d2cR179-R207) to "unwrap" the UIs.
- I added a few more output messages, specially when `-reschedule` is used because the process can take a while (or not progress at all!). I then noticed that calling `monitorPlacementFailures()` per alloc is kind of redundant because evals are per job anyway, but I couldn't find a nice way to have a single monitor goroutine that didn't involve a major refactoring. In hindsight the event stream would probably be better here.
- Handle multi-region job by asking the user for confirmation that it's OK to restart the job in a single region. Coordinated  multi-region job restart is left as a future work.
- Since there were so many questions to ask, I created `askQuestion()` to format the question and handle inputs in a more standardized way.

Here's the diff of the new commits:
https://github.com/hashicorp/nomad/pull/16278/files/971ec5d4b55e01f719c0aa294c39b1f9f2066323..8cdcc1411716ec4a60ca881b4e2559898c8c2748

command/job_restart.go
```diff
@@ -0,0 +1,1185 @@
...
+		})
+}
+
+// shouldProceed blocks and waits for the user to provide a valid input on how
+// to proceed. Returns true if the answer is positive.
+//
+// The flags -batch-wait and -on-error have an 'ask' option. This function
+// handles both to prevent asking the user twice in case they are both set to
+// 'ask' and an error happens.
+func (c *JobRestartCommand) shouldProceed(err error) bool {
```
lgfa29, line 699: I couldn't find a way to test this function since it requires user input and messing with `os.Stdin` during tests seemed dangerous 😬

command/job_restart.go
```diff
@@ -0,0 +1,1185 @@
...
+	)
+
+	// Let ui.Ask() handle interrupt signals.
+	signal.Stop(c.sigsCh)
+	defer func() {
+		signal.Notify(c.sigsCh, syscall.SIGINT)
+	}()
+
+	for {
+		answer, err := c.Ui.Ask(c.Colorize().Color(prefixedQuestion))
```
tgross, line 792: While doing some testing I noticed that `Ask` has some weird behavior around buffering where some errant keypresses after `Ctrl-C` got dropped into the input. I'm not sure if we can actually do anything about it here in Nomad, but just leaving a note in case we encounter it again:

```
    2023-03-17T10:06:07-04:00: Still waiting for allocation "1b86450c" to be replaced
^C^[[A^[[A^[^[==> 2023-03-17T10:06:13-04:00: Restart interrupted, no more allocations will be restarted.
                               Are you sure you want to stop the restart process? [y/N] y
                               Invalid answer "\x1b[A\x1b[A\x1b\x1by"
==> 2023-03-17T10:06:13-04:00: Restart interrupted, no more allocations will be restarted.
                               Are you sure you want to stop the restart process? [y/N] y
```

command/job_restart.go
```diff
@@ -0,0 +1,1185 @@
...
+		c.Ui.Error(fmt.Sprintf("Error initializing client: %v", err))
+		return 1
+	}
+
+	// Use prefix matching to find job.
+	job, err := c.JobByPrefix(c.client, c.jobID, nil)
+	if err != nil {
+		c.Ui.Error(err.Error())
+		return 1
+	}
```
tgross, line 270: During testing I discovered that passing `-reschedule` while there's an ongoing deployment pretty much always results in this command hanging. I used a job with a canary and manual promotion in its update block.

First attempt was running the restart while the deployment was waiting for manual promotion:

```
$ nomad job restart -reschedule -batch-wait=ask example
==> 2023-03-17T09:55:39-04:00: Restarting 7 allocations
    2023-03-17T09:55:39-04:00: Rescheduling allocation "3958e815" for group "group"
^C==> 2023-03-17T09:57:36-04:00: Restart interrupted, no more allocations will be restarted.
                               Are you sure you want to stop the restart process? [y/N]
    2023-03-17T09:57:39-04:00: Still waiting for allocation "3958e815" to be replaced
    2023-03-17T09:58:39-04:00: Still waiting for allocation "3958e815" to be replaced

$ nomad job allocs example
ID        Node ID   Task Group  Version  Desired  Status    Created    Modified
c7bfb048  22e1813e  group       1        run      running   52s ago    21s ago
3958e815  22e1813e  group       1        stop     complete  1m36s ago  46s ago
65110380  22e1813e  group       0        run      running   2m5s ago   1m48s ago
976fdbf3  22e1813e  group       0        run      running   2m13s ago  1m57s ago
a422b9db  22e1813e  group       0        run      running   2m22s ago  2m6s ago
a7c3945c  22e1813e  group       0        run      running   2m32s ago  2m16s ago
6fbff862  22e1813e  group       0        run      running   2m40s ago  2m24s ago
16472afa  22e1813e  group       0        run      running   4m44s ago  4m27s ago

$ nomad deployment list
ID        Job ID   Job Version  Status      Description
04edbf0c  example  1            running     Deployment is running but requires manual promotion
02cf29ff  example  0            successful  Deployment completed successfully
```

After canceling that and promoting the deployment, I deployed again. This time I waited until after promotion. In the example below, the allocation `1b86450c` was in a stopped state, having been replaced by the deployment:

```
$ nomad job restart -reschedule -batch-wait=ask example
==> 2023-03-17T10:03:37-04:00: Restarting 6 allocations
    2023-03-17T10:03:37-04:00: Rescheduling allocation "95f54f2c" for group "group"
    2023-03-17T10:03:37-04:00: Allocation "95f54f2c" replaced by "f254e087", waiting for "f254e087" to start running
    2023-03-17T10:03:43-04:00: Allocation "f254e087" is "running"
==> 2023-03-17T10:03:43-04:00: Proceed with the next batch? [Y/n/<wait duration>] 1s
==> 2023-03-17T10:03:59-04:00: Proceeding restarts with new wait time of 1s
==> 2023-03-17T10:03:59-04:00: Waiting 1s before restarting the next batch
    2023-03-17T10:04:00-04:00: Rescheduling allocation "57654725" for group "group"
    2023-03-17T10:04:00-04:00: Allocation "57654725" replaced by "7bce4be0", waiting for "7bce4be0" to start running
    2023-03-17T10:04:06-04:00: Allocation "7bce4be0" is "running"
==> 2023-03-17T10:04:06-04:00: Waiting 1s before restarting the next batch
    2023-03-17T10:04:07-04:00: Rescheduling allocation "1b86450c" for group "group"
    2023-03-17T10:05:07-04:00: Still waiting for allocation "1b86450c" to be replaced
    2023-03-17T10:06:07-04:00: Still waiting for allocation "1b86450c" to be replaced
```

My take on this is that we probably shouldn't support running `job restart` during an active deployment at all, because it makes it impossible for the deployment watcher to reason about what's going on. We could block that at this line by checking for active deployments and returning an error.

tgross (approved):
LGTM! This whole feature is a masterclass in the difference between incidental complexity and intrinsic complexity. :grinning: 

I've left a fairly long comment around what happens during deployments, but that's a one-liner to block so I'm :+1: on shipping once that's handled.

command/meta.go
```diff
@@ -176,7 +177,34 @@ func (m *Meta) allNamespaces() bool {
...
+		_, coloredUi = ui.(*cli.ColoredUi)
+		if coloredUi {
+			break
+		}
+
+		v := reflect.ValueOf(ui)
+		if v.Kind() == reflect.Ptr {
+			v = v.Elem()
+		}
+		for i := 0; i < v.NumField(); i++ {
```
tgross, line 195: Missed this before I approved, but there's a test panic at this line:

```
=== FAIL: command TestMeta_Colorize (0.00s)
panic: reflect: call of reflect.Value.NumField on zero Value [recovered]
	panic: reflect: call of reflect.Value.NumField on zero Value

goroutine 51085 [running]:
testing.tRunner.func1.2({0x2905000, 0xc0088c53f8})
	/opt/hostedtoolcache/go/1.20.2/x64/src/testing/testing.go:1526 +0x24e
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.20.2/x64/src/testing/testing.go:1529 +0x39f
panic({0x2905000, 0xc0088c53f8})
	/opt/hostedtoolcache/go/1.20.2/x64/src/runtime/panic.go:884 +0x213
reflect.flag.mustBe(...)
	/opt/hostedtoolcache/go/1.20.2/x64/src/reflect/value.go:223
reflect.Value.NumField({0x0?, 0x0?, 0x2e26140?})
	/opt/hostedtoolcache/go/1.20.2/x64/src/reflect/value.go:2041 +0x8d
github.com/hashicorp/nomad/command.(*Meta).Colorize(0x35aac08?)
	/home/runner/work/nomad/nomad/command/meta.go:195 +0x23b
github.com/hashicorp/nomad/command.TestMeta_Colorize.func8(0xc001930f60?)
	/home/runner/work/nomad/nomad/command/meta_test.go:155 +0x16b
testing.tRunner(0xc00df2a4e0, 0xc008a40920)
	/opt/hostedtoolcache/go/1.20.2/x64/src/testing/testing.go:1576 +0x10b
created by testing.(*T).Run
	/opt/hostedtoolcache/go/1.20.2/x64/src/testing/testing.go:1629 +0x3ea
```

command/job_restart.go
lgfa29: Ah interesting! We could try to remove signal characters from the input `Ask` gives us, but the list of signals and their values may change from system to system. 

It may also be possible to [`Ignore`](https://pkg.go.dev/os/signal#Ignore) all other signals while an input is being provided, but that can be pretty annoying since it breaks the expected flow of signal handling.

All of this to say that I agree there's not much we can do here...

command/job_restart.go
lgfa29: Nice catch! We do the same thing for the autoscaler because messing with allocations during deployments is a recipe for disaster 😅 

I think there's a link missing in your comment, but I implemented the check in https://github.com/hashicorp/nomad/pull/16278/commits/069d2c6a06a052c34c4ea2913f509ce6e5257a4b before each restart. I also didn't gate it on `-reschedule` because I think even an in-place restart can be dangerous.

While working on this I realized that the core issue of the command getting blocked during a deployment was a bug in `monitorReplacementAlloc`. We need to keep following `NextAllocation` because, for example, if the first replacement fails Nomad will keep creating follow-up allocations. I fixed this in https://github.com/hashicorp/nomad/pull/16278/commits/fde22d65221dff71c5ac0a5dc1824691b73d03bf.

command/meta.go
lgfa29: Oof...`reflect` is scary 😅 

Fixed in https://github.com/hashicorp/nomad/pull/16278/commits/e53fbe00e4d0bc5d95fa25ee3cb4cf3ea8b1a768, I needed to check for `nil` at the top of the loop 🤦

command/job_restart.go
tgross: > I think there's a link missing in your comment, but I implemented the check in 069d2c6 before each restart.

By "this line" I meant "here, right after `JobByPrefix`. But where you put it is much better, because that'll protect us from a time-of-check-time-of-use bug. :+1: 

> I also didn't gate it on `-reschedule` because I think even an in-place restart can be dangerous.

Good call

> While working on this I realized that the core issue of the command getting blocked during a deployment was a bug in `monitorReplacementAlloc`. We need to keep following `NextAllocation` because, for example, if the first replacement fails Nomad will keep creating follow-up allocations. I fixed this in https://github.com/hashicorp/nomad/commit/fde22d65221dff71c5ac0a5dc1824691b73d03bf.

Nice catch!

PR #16134, opened 2023-02-10
[ui] Service job status panel

philrenaud:
Adds a new "Job Status" panel concept, with "current" and "historical" modes, to represent the running/pending/failed/unplaced status of a job.

image

Note: This is just the Service job, and "steady-state" implementation of a [much larger effort](https://github.com/hashicorp/nomad/issues/16128). Active deployments, and other job types, will be getting variations of this that are not yet considered outside of test fixes herein.

DX changes and side-effects:
- A new `allocStatusDistribution` prop on mirage jobs to distribute allocations according to desired ratios
- job-page/parts/summary broken into a sub-component and a parent wrapper

Issue #15982, opened 2023-01-31
☂️ Nomad HCL in UI


ChaiWithJai:
[RFC](https://docs.google.com/document/d/1XaludieCDUniGlpFDz8SeaFmYLNz1G5500aZwQRqfYI/edit#)
[Design Mocks](https://www.figma.com/file/Ug0J1mgeNUjUqtgCB3ZeH7/HCL-in-the-UI?node-id=4572%3A2063&t=gCmu837NxE7ZiSm3-0)

UI
- [x] Read-Only Flow in `jobs.job.definition` view
  - [ ] ~~Syntax-Highlighting Logic for `HCL2` in `CodeMirror`~~
  - [X] Toggle Logic to switch between `JSON` and `HCL`
  - [x] #16510
  - [x] Create read-only `Table` for viewing `HCL` env and local variables.
  - [x] #16509
- [x] Edit Flow in `jobs.job.definition` view
  - [ ] ~~Update `JobDiff` component~~
  - [x] #16733
  - [x] #16714
- [x] Tablestakes
  - [x] #16589
  - [x] #16571
  - [X] Update `JobEditor` component API (shared across `jobs.run` and `jobs.job.definition`)

V1 Requests
- [x] #16735
- [x] #16736

API
- [ ] Add `jobspec` field to `Job` struct in REST API (and Raft)
- [ ] Add `variables` field  to `Job` struct in REST API (and Raft) for environment and local variables
- [ ] Update `POST /parse` endpoint to accept variables

CLI
- [ ] `job inspect` includes a flag to return jobspec

ChaiWithJai:
@mikenomitch The [design mocks](https://www.figma.com/file/Ug0J1mgeNUjUqtgCB3ZeH7/HCL-in-the-UI?node-id=4572%3A2063&t=gCmu837NxE7ZiSm3-0) show that users are able to edit HCL variables. However, your document says that should not be included.

Can you take a peak and determine if there's any functionality from that mock that won't be included?

ChaiWithJai:
@mikenomitch Also, just want to confirm that we're scoping this to the `jobs.job.definition` view and we will not be making any changes to:

- Job Dispatch (parent jobs)
- Job Run

mikenomitch:
[...]

PR #16279, opened 2023-02-28
ui:  Toggle for `read-only` view

### LINKED ISSUE #16269: [ui, hcl] Toggle Button
Part of #15982

ChaiWithJai:
Closes #16269 for creating a Toggle Button to display both the Job Specification and Full Definition.

**Functional Changes**
This PR adds a toggle button when the `JobEditor` component is in a read-only state. 

**Refactoring**
This PR refactors the JobEditor component to improve its flexibility and maintainability. The original code tightly coupled the component's state with its rendering logic, which made it difficult to modify and extend. This PR introduces several improvements to make the code more modular and easier to understand.

**Changes**
- Uses Ember's `component helper` to dynamically render the appropriate child component based on the value of the stage property.
- Reorganizes the code to separate the component state from the rendering logic, which makes the component more modular and easier to test.

### DIFF
--- a/ui/app/components/job-editor.js
+++ b/ui/app/components/job-editor.js
@@ -12,6 +12,21 @@ export default class JobEditor extends Component {
 
   @tracked error = null;
   @tracked planOutput = null;
+  @tracked view = 'job-spec';
+  @tracked isEditing = false;
+
+  @action
+  edit() {
+    this.args.job.set(
+      '_newDefinition',
+      JSON.stringify(this.args.definition, null, 2)
+    );
+    this.isEditing = true;
+  }
+
+  onCancel() {
+    this.isEditing = false;
+  }
 
   get stage() {
     return this.planOutput ? 'plan' : 'editor';
@@ -94,5 +109,34 @@ export default class JobEditor extends Component {
 
     const [file] = event.target.files;
     reader.readAsText(file);
+  }
+
+  @action
+  toggleView() {
+    const opposite = this.view === 'job-spec' ? 'full-definition' : 'job-spec';
+    this.view = opposite;
+  }
+
+  get definition() {
+    if (this.view === 'full-definition') {
+      return this.args.definition;
+    } else {
+      return this.args.specification;
+    }
+  }
+
+  get data() {
+    return {
+      view: this.view,
+      definition: this.definition,
+      job: this.args.job,
+    };
+  }
+
+  get fns() {
+    return {
+      onEdit: this.edit,
+      onToggle: this.toggleView,
+    };
   }
 }
--- a/ui/app/components/job-editor.js
+++ b/ui/app/components/job-editor.js
@@ -12,12 +12,13 @@ export default class JobEditor extends Component {
 
   @tracked error = null;
   @tracked planOutput = null;
-  @tracked view = 'job-spec';
   @tracked isEditing;
+  @tracked view;
 
   constructor() {
     super(...arguments);
     this.isEditing = !!(this.args.context === 'new');
+    this.view = this.args.specification ? 'job-spec' : 'full-definition';
   }
 
   toggleEdit(bool) {
@@ -141,6 +142,7 @@ export default class JobEditor extends Component {
     return {
       cancelable: this.args.cancelable,
       definition: this.definition,
+      hasSpecification: !!this.args.specification,
       job: this.args.job,
       planOutput: this.planOutput,
       shouldShowPlanMessage: this.shouldShowPlanMessage,
--- a/ui/app/components/job-editor.js
+++ b/ui/app/components/job-editor.js
@@ -30,7 +30,7 @@ export default class JobEditor extends Component {
   }
 
   get stage() {
-    if (this.planOutput) return 'plan';
+    if (this.planOutput) return 'review';
     if (this.isEditing) return 'edit';
     else return 'read';
   }
@@ -135,7 +135,6 @@ export default class JobEditor extends Component {
       job: this.args.job,
       planOutput: this.planOutput,
       shouldShowPlanMessage: this.shouldShowPlanMessage,
-      stage: this.stage,
       view: this.view,
     };
   }
--- a/ui/app/components/job-editor.js
+++ b/ui/app/components/job-editor.js
@@ -13,7 +13,16 @@ export default class JobEditor extends Component {
   @tracked error = null;
   @tracked planOutput = null;
   @tracked view = 'job-spec';
-  @tracked isEditing = false;
+  @tracked isEditing;
+
+  constructor() {
+    super(...arguments);
+    this.isEditing = !!(this.args.context === 'new');
+  }
+
+  toggleEdit(bool) {
+    this.isEditing = bool || !this.isEditing;
+  }
 
   @action
   edit() {
@@ -21,12 +30,12 @@ export default class JobEditor extends Component {
       '_newDefinition',
       JSON.stringify(this.args.definition, null, 2)
     );
-    this.isEditing = true;
+    this.toggleEdit(true);
   }
 
   @action
   onCancel() {
-    this.isEditing = false;
+    this.toggleEdit(false);
   }
 
   get stage() {
--- a/ui/app/components/job-editor.js
+++ b/ui/app/components/job-editor.js
@@ -24,15 +24,18 @@ export default class JobEditor extends Component {
     this.isEditing = true;
   }
 
+  @action
   onCancel() {
     this.isEditing = false;
   }
 
   get stage() {
-    return this.planOutput ? 'plan' : 'editor';
+    if (this.planOutput) return 'plan';
+    if (this.isEditing) return 'edit';
+    else return 'read';
   }
 
-  @localStorageProperty('nomadMessageJobPlan', true) showPlanMessage;
+  @localStorageProperty('nomadMessageJobPlan', true) shouldShowPlanMessage;
 
   @(task(function* () {
     this.reset();
@@ -127,16 +130,27 @@ export default class JobEditor extends Component {
 
   get data() {
     return {
-      view: this.view,
+      cancelable: this.args.cancelable,
       definition: this.definition,
       job: this.args.job,
+      planOutput: this.planOutput,
+      shouldShowPlanMessage: this.shouldShowPlanMessage,
+      stage: this.stage,
+      view: this.view,
     };
   }
 
   get fns() {
     return {
+      onCancel: this.onCancel,
       onEdit: this.edit,
+      onPlan: this.plan,
+      onReset: this.reset,
+      onSaveAs: this.args.handleSaveAsTemplate,
+      onSubmit: this.submit,
       onToggle: this.toggleView,
+      onUpdate: this.updateCode,
+      onUpload: this.uploadJobSpec,
     };
   }
 }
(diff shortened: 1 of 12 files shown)


ui/app/templates/components/job-editor/read.hbs
```diff
@@ -0,0 +1,48 @@
...
+        </button>
+        <button 
+            class="button is-small is-borderless {{if (eq @data.view "full-definition") "is-active"}}"
+            type="button"
+            {{on "click" @fns.onToggle}}
+            data-test-toggle-full
+        >
+            Full Definition
+        </button>
+        </div>
```
ChaiWithJai, line 21: This should be conditionally rendered because old jobs will not have this functionality. This wasn't addressed in the mocks so I will tackle this issue separately.

ChaiWithJai (commented):
FYI `it allows users to toggle between full specification and JSON definition` is a test that is made to fail until the next PR.

philrenaud:
Small note: if you use the GitHub web ui's "Create a branch for...." from an issue, it'll auto-track it. Makes it a little easier to traverse from issue to pr and back.

ui/app/templates/components/job-editor/edit.hbs
```diff
@@ -0,0 +1,47 @@
+<div class="boxed-section">
+    <div class="boxed-section-head">
+    Job Definition
+    {{#if @data.cancelable}}
+        <button
+            class="button is-light is-compact pull-right"
+            onclick={{fn @fns.onCancel}}
```
philrenaud, line 7: this'll throw you a requires at least 1 argument error; try this:

```suggestion
            onclick={{@fns.onCancel}}
```

ui/app/components/job-editor.js
```diff
@@ -95,4 +122,43 @@ export default class JobEditor extends Component {
...
+      onEdit: this.edit,
+      onPlan: this.plan,
+      onReset: this.reset,
+      onSaveAs: this.args.handleSaveAsTemplate,
+      onSubmit: this.submit,
+      onToggle: this.toggleView,
+      onUpdate: this.updateCode,
+      onUpload: this.uploadJobSpec,
+    };
+  }
```
philrenaud, line 163: Hey, this is a neat pattern! Is this something we use anyplace else in Nomad?

ui/app/components/job-editor.js
```diff
@@ -12,12 +12,39 @@ export default class JobEditor extends Component {
...
+  onCancel() {
+    this.toggleEdit(false);
+  }
 
   get stage() {
-    return this.planOutput ? 'plan' : 'editor';
+    if (this.planOutput) return 'review';
+    if (this.isEditing) return 'edit';
+    else return 'read';
   }
```
philrenaud, line 45: Request: Now that this is becoming a more complicated stateful page, I'd like to see that state represented as queryParams in the URL somehow.

`jobs/:jobid/definition?mode=edit&spec=hcl` vs `jobs/:jobid/definition?mode=read&spec=json` and the like.

ui/app/routes/jobs/job/definition.js
```diff
@@ -1,14 +1,24 @@
...
 
-    return job.fetchRawDefinition().then((definition) => ({
+    const definition = await job.fetchRawDefinition();
+
+    const definitionWithoutSpecification = { ...definition };
+    delete definitionWithoutSpecification.Specification;
+
+    const specification = await new Blob([
+      definition?.Specification?.Definition,
+    ]).text();
```
philrenaud, line 15: Can you help me understand what this is accomplishing? `delete prop` outside of a serializer has me wondering if this is the optimal place for this.

ui/tests/acceptance/job-definition-test.js
```diff
@@ -8,6 +8,556 @@ import Definition from 'nomad-ui/tests/pages/jobs/job/definition';
...
+    Definition:
+      'job "docs" {\n  namespace = "madness"\n  group "example" {\n    task "server" {\n      service {\n        tags = ["leader", "mysql"]\n\n        port = "db"\n\n        meta {\n          meta = "for your service"\n        }\n\n        check {\n          type     = "tcp"\n          port     = "db"\n          interval = "10s"\n          timeout  = "2s"\n        }\n\n        check {\n          type     = "script"\n          name     = "check_table"\n          command  = "/usr/local/bin/check_mysql_table_status"\n          args     = ["--verbose"]\n          interval = "60s"\n          timeout  = "5s"\n\n          check_restart {\n            limit = 3\n            grace = "90s"\n            ignore_warnings = false\n          }\n        }\n      }\n    }\n  }\n}\n',
+    Type: 'hcl',
+    Variables: {
+      datacenters: ['west'],
+      external_port: 4000,
+    },
+  },
+};
+
```
philrenaud, line 560: Q: why not use a factory to accomplish this? They're extensible and most things you have listed here are either already job defaults or else unrelated to the acceptance test below.

ui/app/styles/components/codemirror.scss
```diff
@@ -156,3 +156,29 @@ header.run-job-header {
   display: flex;
   justify-content: space-between;
 }
+
+.select-mode {
```
philrenaud, line 160: Request: have this be more specific, either by class name or by nesting. As it stands, I could see someone creating another component with a class of `select-mode` in the future and not being sure why there's suddenly a fresh looking grid applied.

ui/app/components/job-editor.js
```diff
@@ -12,12 +12,39 @@ export default class JobEditor extends Component {
...
+  }
+
+  @action
+  edit() {
+    this.args.job.set(
+      '_newDefinition',
+      JSON.stringify(this.args.definition, null, 2)
+    );
+    this.toggleEdit(true);
+  }
```
philrenaud, line 34: Request: could you have the default/smallcluster mirage case include a job with an HCL definition? As it stands I just get `undefined` for all cases right now and it's hard to evaluate if _newDefinition is being set as intended.

philrenaud (changes requested):
A few changes and requests for clarification.

Some of the terminology could be changed to be clearer IMO. For example, I'm able to grok that `definition` means JSON and `specification` means HCL in this context, but that wasn't immediately clear.

ui/app/components/job-editor.js
ChaiWithJai: This is a component, not a controller. I don't think URL based state makes sense here because this component is rendered across multiple views and meant to expand.

ui/app/styles/components/codemirror.scss
ChaiWithJai: The #2 problem of programming is naming things. I can have ChatGPT name this or take a name from you! Is that unreasonable?

ui/app/templates/components/job-editor/edit.hbs
ChaiWithJai: The tests would capture that, no? Per the [RFC](https://github.com/emberjs/rfcs/blob/master/text/0470-fn-helper.md) the `fn` helper helps with:

- binding the `this` context
- applying a list of arguments
- partial application to prevent prop drilling

None of these apply and we have test coverage for this. I manually tested this and not seeing this error either.

Can you reproduce?

ui/app/components/job-editor.js
ChaiWithJai: Yes! We use this in `breadcrumbs` and `evaluations`! Any major net new features I've worked on!

ui/app/routes/jobs/job/definition.js
ChaiWithJai: `fetchRawDefinition` specifically is made to skip the Ember Data and thus should not be serialized because we serialize payloads to format them before they enter the Ember Data store.

That leaves me with 2 places to handle this concerns:  the Route who makes the `fetch` request outside of `Ember Data` or creating a `Service` and extract our logic there. I opted for minimally changing the API surface and handling the concern where we are loading the data.

ui/app/components/job-editor.js
ChaiWithJai: We have test coverage for this via the `job-editor` integration test and in the `job-definition` acceptance test suite.

Touching Mirage deserves its own PR as this PR is big enough for just separate state and presentation logic. 

I originally highlighted this as adjacent work in the [umbrella ticket](https://github.com/hashicorp/nomad/issues/15982) because I don't think it matters. When this goes into production, users will have a mix of jobs that have specifications and jobs that do not.

The most appropriate thing to do to account for this unforeseen UX would be [conditionally rendering](https://github.com/hashicorp/nomad/pull/16279/files#r1122238137) the toggle which is something we will need to do in production.

ui/app/templates/components/job-editor/edit.hbs
philrenaud: The first line of that RFC includes "to allow clear argument passing for functions in templates"; fn is for passing args and you're not passing any here. See also the current lint error on this PR.

ui/app/components/job-editor.js
philrenaud: Well why not have `stage` exist as a property that gets passed down to this component? `<JobEditor @stage={{this.qp.stage}} ...` etc. from the controller?

and you'd want to transform your `get stage()` in this component to something like

```
  get stage() {
    return this.args.stage || <your current logic>
  }
```

ui/app/styles/components/codemirror.scss
ChaiWithJai: I didn't actually write this code FYI. I [stole](https://github.com/hashicorp/nomad/pull/16134/files#diff-2a532068e46093a6e751d50034579f500880b133d580718bab64a1ec5096630fR2-R26) it from your work.

ui/app/components/job-editor.js
ChaiWithJai: I could totally do that, this PR is for separating the concerns of state and presentation logic. I could write a separate PR for updating the component's API and changing this to be addressable state flowing in from the controller, instead of UI state on the component would require a refactoring of the associated routes and controllers everywhere this component is invoked. So it will require its own PR because that's a larger change that shouldn't scope creep into this PR.

ui/app/templates/components/job-editor/edit.hbs
ChaiWithJai: Totally missed that. I read this as vice-versa. Like you were advocating for that change. So disregard my comments and good catch because I made a mistake here.

ui/app/routes/jobs/job/definition.js
philrenaud: Okay, so why are we deleting `.Specification` here?

What's the format of `definition` ? Why does `Specification` need to be removed from it? I'm struggling to understand what this model is trying to accomplish.

ChaiWithJai: We're doing what we would do in our `Serializer` here. We don't require this field. We could keep it, as we're not concerned with it and it won't break anything.

ui/tests/acceptance/job-definition-test.js
ChaiWithJai: I'm not in the camp of factory or writing test logic. It becomes painful to maintain and that's personally the part of the test suite I hate the most. If you want it, I'm happy to create it. But it's just more code to maintain.

ChaiWithJai: There's a few things here:

1. This data isn't going to be modeled because this is a raw request. We're explicitly using the API response to visualize the response itself in code mirror.
2. We could overcome this using the `afterCreate` hook. But we already have 200 lines of logic in that hook and I simply don't want to touch it.

The best solution would be to just move this into a utility file if you wish to separate this fixture from the test code. And I can address that in a follow up PR.

philrenaud (approved):

PR #16281, opened 2023-03-01
client: don't emit task shutdown delay event if not waiting

lgfa29:
https://github.com/hashicorp/nomad/pull/14775 introduced a new task event when the task runner is waiting for the `shutdown_delay`, but some requests can be configured to [ignore the `shutdown_delay`](https://developer.hashicorp.com/nomad/docs/commands/alloc/stop#no-shutdown-delay), in which case it can confusing to have this event, so emit one that notes the delay is being skipped instead.

~No changelog or backports necessary since #14775 has not being released yet.~

I just realized that I should've merged this for 1.5.0 🤦 

Not big deal, but I added a changelog entry now.

### DIFF
--- a/client/allocrunner/taskrunner/task_runner_test.go
+++ b/client/allocrunner/taskrunner/task_runner_test.go
@@ -1108,6 +1108,16 @@ func TestTaskRunner_NoShutdownDelay(t *testing.T) {
 
 	err := <-killed
 	require.NoError(t, err, "killing task returned unexpected error")
+
+	// Check that we only emit the expected events.
+	hasEvent := false
+	for _, ev := range tr.state.Events {
+		require.NotEqual(t, structs.TaskWaitingShuttingDownDelay, ev.Type)
+		if ev.Type == structs.TaskSkippingShutdownDelay {
+			hasEvent = true
+		}
+	}
+	require.True(t, hasEvent)
 }
 
 // TestTaskRunner_Dispatch_Payload asserts that a dispatch job runs and the
--- a/client/allocrunner/taskrunner/task_runner_test.go
+++ b/client/allocrunner/taskrunner/task_runner_test.go
@@ -1107,17 +1107,17 @@ func TestTaskRunner_NoShutdownDelay(t *testing.T) {
 	}
 
 	err := <-killed
-	require.NoError(t, err, "killing task returned unexpected error")
+	must.NoError(t, err)
 
 	// Check that we only emit the expected events.
 	hasEvent := false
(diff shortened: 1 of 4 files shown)


client/allocrunner/taskrunner/task_runner_test.go
```diff
@@ -1108,6 +1108,16 @@ func TestTaskRunner_NoShutdownDelay(t *testing.T) {
...
+
+	// Check that we only emit the expected events.
+	hasEvent := false
+	for _, ev := range tr.state.Events {
+		require.NotEqual(t, structs.TaskWaitingShuttingDownDelay, ev.Type)
+		if ev.Type == structs.TaskSkippingShutdownDelay {
+			hasEvent = true
+		}
+	}
+	require.True(t, hasEvent)
```
tgross, line 1120: Not a blocker, but I think this kind of loop is what [`must.Contains`](https://pkg.go.dev/github.com/shoenig/test#Contains)  was designed for?

tgross (approved):
LGTM!

client/allocrunner/taskrunner/task_runner_test.go
lgfa29: I couldn't quite make that work. The best I was able to do was something like this:

```go
want := &structs.TaskEvent{Type: structs.TaskSkippingShutdownDelay}
must.SliceContainsFunc(t, tr.state.Events, want, func(a, b *structs.TaskEvent) bool {
	return a.Type == b.Type
})
```

But `SliceNotContainsFunc`, which I would have to use to check for the absence of  `TaskWaitingShuttingDownDelay` doesn't exist yet, so I left the initial implementation but replacing `require` with `must`.

shoenig: > SliceNotContainsFunc

probably should exist, created https://github.com/shoenig/test/issues/110

shoenig (approved):
LGTM!

client/allocrunner/taskrunner/task_runner_test.go
lgfa29: Awesome, thanks!

I was thinking about the ergonomics of this use case, and I think it could be helpful to have something like these:

```go
s := []string{"hit", "miss"}

must.SliceAtLeastOne(s, func(i string) bool {
  return i == "hit"
}) // success

must.SliceAll(s, func(i string) bool {
  return i == "hit"
}) // fail

must.SliceNone(s, func(i string) bool {
  return i == "none"
}) // success
```

PR #16287, opened 2023-03-01
handle `FSM.Apply` errors in `raftApply`

tgross:
The signature of the `raftApply` function requires that the caller unwrap the first returned value (the response from `FSM.Apply`) to see if it's an error. This puts the burden on the caller to remember to check two different places for errors, and we've done so inconsistently.

Update `raftApply` to do the unwrapping for us and return any `FSM.Apply` error as the error value. Similar work was done in Consul in https://github.com/hashicorp/consul/pull/9991. This eliminates some boilerplate and surfaces a few minor bugs in the process:

* job deregistrations of already-GC'd jobs were still emitting evals
* reconcile job summaries does not return scheduler errors
* node updates did not report errors associated with inconsistent service discovery or CSI plugin states

Note that although _most_ of the `FSM.Apply` functions return only errors (which makes it tempting to remove the first return value entirely), there are few that return `bool` for some reason and Variables relies on the response value for proper CAS checking.

### DIFF
--- a/api/jobs_test.go
+++ b/api/jobs_test.go
@@ -1554,7 +1554,7 @@ func TestJobs_Deregister(t *testing.T) {
 	must.NoError(t, err)
 	assertWriteMeta(t, wm)
 
-	// Attempting delete on non-existing job returns an error
+	// Attempting delete on non-existing job does not return an error
 	_, _, err = jobs.Deregister("nope", false, nil)
 	must.NoError(t, err)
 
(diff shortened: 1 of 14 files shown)


api/jobs_test.go
```diff
@@ -1554,7 +1554,7 @@ func TestJobs_Deregister(t *testing.T) {
 	must.NoError(t, err)
 	assertWriteMeta(t, wm)
 
-	// Attempting delete on non-existing job returns an error
+	// Attempting delete on non-existing job does not return an error
```
schmichael, line 1557: Good catch! I can imagine that has provoked some unnecessary :thinking:

schmichael (approved):

PR #16289, opened 2023-03-01
nsd: always set deregister flag after deregistration of group

shoenig:
This PR fixes a bug where the group service hook's deregister flag was not set in `Postrun()`, causing the hook to attempt deregistrations twice during job updates (alloc replacement), where `PreKill()` would also do a deregistration. With the Nomad service provider, this shows up as Error logs on every job update.

First commit is the bug fix, second is drive by test cleanup.

### DIFF
--- a/.changelog/16289.txt
+++ b/.changelog/16289.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+services: Fixed a bug where a service would be deregistered twice
+```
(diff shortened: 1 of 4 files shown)


jrasell (commented):
LGTM! TIL `helper.WithLock`

jrasell (approved):

PR #16292, opened 2023-03-02
deps: update go-plugin to 1.4.9

### LINKED ISSUE #16288: logging: WARN messages on every agent start
Lately I've been seeing these WARN level agent log messages on every agent startup. Either something is actually wrong and we should investigate, or these should not be warnings.

```
2023-03-01T10:29:42.765-0600 [WARN]  client.alloc_runner.task_runner.task_hook.logmon: plugin configured with a nil SecureConfig: alloc_id=828b90e5-1fa6-2a31-9acb-dafe46a90d9b task=task
2023-03-01T10:29:43.058-0600 [WARN]  client.driver_mgr.docker.docker_logger: plugin configured with a nil SecureConfig: driver=docker
```

tgross:
Fixes #16288. An earlier version of `go-plugin` introduced a warning log if `SecureConfig` is unset. For Nomad and other applications that have "internal" `go-plugin` consumers where the application runs itself as a plugin, this causes spurious warn-level logs. For Nomad in particular this means every task driver and logmon invocation emits the log, which is our primary operation.

The change was reverted upstream, so this changeset picks up the reverted version.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -61,7 +61,7 @@ require (
 	github.com/hashicorp/go-memdb v1.3.4
 	github.com/hashicorp/go-msgpack v1.1.5
 	github.com/hashicorp/go-multierror v1.1.1
-	github.com/hashicorp/go-plugin v1.4.8
+	github.com/hashicorp/go-plugin v1.4.9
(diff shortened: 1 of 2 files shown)


shoenig (approved):
LGTM

PR #16304, opened 2023-03-02
deps: remove replace statement for go-discover

shoenig:
Which we no longer need since we no longer have consul as a dependency (the thing that also imported go-discover with it's own desired version). 

backport only to 1.5.x and 1.4.x (first without Consul dependency)


shoenig:
@shoenig remember to update github.com/hashicorp/nomad/issues/11826

tgross (approved):
LGTM!

PR #16305, opened 2023-03-02
remove backcompat support for non-atomic job registration

tgross:
In Nomad 0.12.1 we introduced atomic job registration/deregistration, where the new eval was written in the same raft entry. Backwards-compatibility checks were supposed to have been removed in Nomad 1.1.0, but we missed that. This is long safe to remove.


schmichael (approved):
> supposed to have been removed in Nomad 1.1.0, but we missed that.

missed by just a bit :sweat_smile: 

Thanks for cleaning this up Tim!

tgross:
I was on the fence, but probably can't hurt to include it. Will fix.

PR #16306, opened 2023-03-02
cli: use shared logic for resolving job prefix

lgfa29:
Several `nomad job` subcommands had duplicate or slightly similar logic for resolving a job ID from a CLI argument prefix, while others did not have this functionality at all.

This commit pulls the shared logic to the command Meta and updates all `nomad job` subcommands to use it.


schmichael (approved):
One linter error, but otherwise looks great! Thanks for cleaning all of this up!

tgross (approved):
LGTM!

PR #16309, opened 2023-03-03
docs: fix typos in task-api.mdx and workload-identity.mdx

aofei:
According to [envvars.mdx](https://github.com/hashicorp/nomad/blob/f89910d4650c47006838a7b6de17cbacdc62fe1c/website/content/partials/envvars.mdx), I don't think there is a `SECRETS_DIR` environment variable here, it should be `NOMAD_SECRETS_DIR`.


tgross (approved):
Thanks @aofei!

Issue #16276, opened 2023-02-28
Unexpected loss of capabilities when using Workload Associated ACL Policies


aofei:
### Nomad version

Nomad v1.5.0-rc.1
BuildDate 2023-02-27T21:58:15Z
Revision d65756c82fb53a977da741ac3319a189e8a3e7bd

### Operating system and Environment details

Linux nomad-bug-16276 https://github.com/hashicorp/nomad/pull/36-Ubuntu SMP Mon Jan 23 21:04:15 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

### Issue

https://github.com/hashicorp/nomad/blob/542b23e999b5338cdcfa56bfdae1db9aa834cb3c/website/content/docs/concepts/workload-identity.mdx?plain=1#L46-L52

According to the above docs, I can associate **additional ACL policies** with workload identities. IIUC, by "associate additional ACL policies", it means only associate new ACL policies but not disassociate the existing default ACL policy (or any of its capabilities). So the workload identity should just get more capabilities, not lose some implicitly. But it didn't in my test.

### Reproduction steps

1. Create file `foobar.nomad.hcl`:

```hcl
job "foobar" {
	group "foobar" {
		network {
			port "foobar" {}
		}
		task "foobar" {
			driver = "exec"
			config {
				command = "tail"
				args = ["-f", "/dev/null"]
			}
			service {
				port = "foobar"
				provider = "nomad"
			}
			identity {
				env = true
			}
		}
	}
}
```

2. Create file `foobar-job.policy.hcl`:

```hcl
namespace "default" {
	variables {
		path "foobar/*" {
			capabilities = ["read", "list", "write", "destroy"]
		}
	}
}
```

3. Open terminal tab 1 and execute:

```bash
$ sudo nomad agent -dev -acl-enabled
```

4. Open terminal tab 2 and execute:

```bash
$ nomad acl bootstrap
$ export NOMAD_TOKEN=<The Bootstrap Token>
$ nomad run foobar.nomad.hcl
$ # Wait until the job deploys successfully.
$ nomad alloc exec -job foobar nomad service list
$ # The output should be:
$ # Service Name          Tags
$ # foobar-foobar-foobar  []
$ nomad acl policy apply -namespace default -job foobar foobar-job-policy foobar-job.policy.hcl
$ nomad alloc exec -job foobar nomad service list
$ # The output should be:
$ # Error listing service registrations: Unexpected response code: 403 (Permission denied)
```

[...]

PR #16316, opened 2023-03-03
service: fix regression in task access to list/read endpoint

tgross:
Fixes https://github.com/hashicorp/nomad/issues/16276

When native service discovery was added, we used the node secret as the auth token. Once Workload Identity was added in Nomad 1.4.x we needed to use the claim token for `template` blocks, and so we allowed valid claims to bypass the ACL policy check to preserve the existing behavior. (Invalid claims are still rejected, so this didn't widen any security boundary.)

In reworking authentication for 1.5.0, we unintentionally removed this bypass. For WIs without a policy attached to their job, everything works as expected because the resulting `acl.ACL` is nil. But once a policy is attached to the job the `acl.ACL` is no longer nil and this causes permissions errors.

Fix the regression by adding back the bypass for valid claims. In future work, we should strongly consider getting turning the implicit policies into real `ACLPolicy` objects (even if not stored in state) so that we don't have these kind of brittle exceptions to the auth code.

### DIFF
--- a/.changelog/16316.txt
+++ b/.changelog/16316.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+service: Fixed a bug where attaching a policy to a job would prevent workload identities for the job from reading the service registration API
+```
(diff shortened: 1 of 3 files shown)


pkazmierczak (approved):
LGTM!

PR #16318, opened 2023-03-03
deps: Update ioutil deprecated library references to os and io respectively in the client package

lhaig:
No user facing changes so I assume no change log is required.

### DIFF
--- a/client/allocdir/alloc_dir.go
+++ b/client/allocdir/alloc_dir.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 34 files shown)


client/allocdir/task_dir.go
```diff
@@ -184,12 +183,13 @@ func (t *TaskDir) embedDirs(entries map[string]string) error {
...
 		// Enumerate the files in source.
-		dirEntries, err := ioutil.ReadDir(source)
+		dirEntries, err := os.ReadDir(source)
 		if err != nil {
 			return fmt.Errorf("Couldn't read directory %v: %v", source, err)
 		}
 
-		for _, entry := range dirEntries {
+		for _, fileEntry := range dirEntries {
+			entry, _ := fileEntry.Info()
```
shoenig, line 192: I think we need to not ignore this error - it would have previously been reported by `ioutil.ReadDir`

The deprecation notice suggests the same, 
https://github.com/golang/go/blob/master/src/io/ioutil/ioutil.go#L59-L68

client/allocdir/alloc_dir.go
```diff
@@ -357,12 +356,13 @@ func (d *AllocDir) List(path string) ([]*cstructs.AllocFileInfo, error) {
...
 	p := filepath.Join(d.AllocDir, path)
-	finfos, err := ioutil.ReadDir(p)
+	finfos, err := os.ReadDir(p)
 	if err != nil {
 		return []*cstructs.AllocFileInfo{}, err
 	}
 	files := make([]*cstructs.AllocFileInfo, len(finfos))
-	for idx, info := range finfos {
+	for idx, file := range finfos {
+		info, _ := file.Info()
```
shoenig, line 365: Same as the note below

shoenig (changes requested):
Thanks @lhaig - just the one fix needed in 2 places, where we shouldn't ignore an error.

shoenig (approved):
LGTM!

PR #16319, opened 2023-03-03
api: set last index and request time on alloc stop

lgfa29:
Some of the methods in `Allocations()` incorrectly use the `putQuery` in API calls where `put` is more appropriate since they are not reading information back. These methods are also not returning request metadata such as `LastIndex` back to callers, which can be useful to have in some scenarios.

They also provide poor developer experience as they take an `*api.Allocation` struct when only the allocation ID is necessary. This can lead consumers to make unnecessary API calls to fetch the full allocation.

Fixing these problems require updating the methods' signatures so they take `*WriteOptions` instead of `*QueryOptions` and return `*WriteMeta`, but this is a breaking change that requires advanced notice to consumers.

This commit adds a future breaking change notice and also fixes the `Stop` method so it properly returns request metadata in a backwards compatible way.

### DIFF
--- a/.changelog/16319.txt
+++ b/.changelog/16319.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+api: Fix `Allocations().Stop()` method to properly set the request `LastIndex` and `RequestTime` in the response
+```
--- a/.changelog/16319.txt
+++ b/.changelog/16319.txt
@@ -1,3 +1,11 @@
 ```release-note:bug
 api: Fix `Allocations().Stop()` method to properly set the request `LastIndex` and `RequestTime` in the response
 ```
+
+```release-note:deprecation
+api: The `Restart()`, `Stop()`, and `Signal()` methods in the `Allocations` struct will have their signatures modified in Nomad 1.6.0
+```
+
+```release-note:deprecation
+api: The `RestartAllTasks()` method in the `Allocations` struct will be removed in Nomad 1.6.0
+```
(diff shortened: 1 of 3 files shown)


.changelog/16319.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+api: Fix `Allocations().Stop()` method to properly set the request `LastIndex` and `RequestTime` in the response
```
lgfa29, line 2: This is the only user facing change, but should I note the future breaking changes somehow?

api/allocations.go
```diff
@@ -179,9 +185,27 @@ func (a *Allocations) RestartAllTasks(alloc *Allocation, q *QueryOptions) error
...
+			Headers:   q.Headers,
+			ctx:       q.ctx,
+		}
+	}
+
 	var resp AllocStopResponse
-	_, err := a.client.putQuery("/v1/allocation/"+alloc.ID+"/stop", nil, &resp, q)
+	wm, err := a.client.put("/v1/allocation/"+alloc.ID+"/stop", nil, &resp, w)
+	resp.LastIndex = wm.LastIndex
+	resp.RequestTime = wm.RequestTime
```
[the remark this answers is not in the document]
lgfa29: Ooof..great catch! I also added a test case to cover this scenario.

.changelog/16319.txt
lgfa29: Ah nice. I added the deprecation notes and I also remembered that I added `RestartAllTasks()` method to [prevent a breaking change](https://github.com/hashicorp/nomad/pull/14127#discussion_r948577573). Since we're breaking things already we may as well clean that up.

PR #16331, opened 2023-03-04
deps: Update ioutil library references to os and io respectively for drivers package

lhaig:
No user facing changes so I assume no change log is required


shoenig (approved):
LGTM!

PR #16345, opened 2023-03-06
[ui, job status] Make panel status mode a queryParam

### LINKED ISSUE #16340: Service Job Panel: Make current/historical a query parameter
Make it so the URL is affected by and affects these:
image

philrenaud:
Resolves #16340 

Passes `statusMode` down to the job status panel; changes behaviour to be at job index level.

### DIFF
--- a/ui/app/components/job-status/panel.hbs
+++ b/ui/app/components/job-status/panel.hbs
@@ -21,7 +21,7 @@
     {{#if (eq @statusMode "current")}}
       <h3 class="title is-4 running-allocs-title"><strong>{{@job.runningAllocs}}/{{this.totalAllocs}}</strong> Allocations Running</h3>
 
-      aaa<JobStatus::AllocationStatusRow @allocBlocks={{this.allocBlocks}} />bbb
+      <JobStatus::AllocationStatusRow @allocBlocks={{this.allocBlocks}} />
 
       <footer>
         <legend>
--- a/ui/app/components/job-status/panel.hbs
+++ b/ui/app/components/job-status/panel.hbs
@@ -1,27 +1,27 @@
-<div class="job-status-panel boxed-section" data-test-job-status-panel data-test-status-mode={{this.mode}}>
+<div class="job-status-panel boxed-section" data-test-job-status-panel data-test-status-mode={{@statusMode}}>
   <div class="boxed-section-head">
     <h2>Status</h2>
     <div class="select-mode">
       <button type="button"
         data-test-status-mode-current
-        class="button is-small is-borderless {{if (eq this.mode "current") "is-active"}}"
-        {{on "click" (action (mut this.mode) "current")}}
+        class="button is-small is-borderless {{if (eq @statusMode "current") "is-active"}}"
+        {{on "click" (action @setStatusMode "current")}}
       >
         Current
       </button>
       <button type="button"
         data-test-status-mode-historical
-        class="button is-small is-borderless {{if (eq this.mode "historical") "is-active"}}"
-        {{on "click" (action (mut this.mode) "historical")}}>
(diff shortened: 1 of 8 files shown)


ui/app/controllers/jobs/job/index.js
```diff
@@ -31,6 +33,11 @@ export default class IndexController extends Controller.extend(
   sortDescending = false;
   activeTask = null;
 
+  /**
+   * @type {('current'|'historical')}
+   */
+  statusMode = 'current';
```
ChaiWithJai, line 39: `question`: Shouldn't this be a `tracked` property? I think despite being a `classic` controller `tracked` should be [backwards compatible](https://guides.emberjs.com/release/upgrading/current-edition/tracked-properties/#toc_backwards-compatibility). We do this in the `Tokens` controller I believe.

ui/app/controllers/jobs/job/index.js
```diff
@@ -39,4 +46,12 @@ export default class IndexController extends Controller.extend(
       this.set('activeTask', null);
     }
   }
+
+  /**
+   * @param {('current'|'historical')} mode
+   */
+  @action
+  setStatusMode(mode) {
+    this.set('statusMode', mode);
```
ChaiWithJai, line 55: Once you edit to `tracked` it should be:

```
this.statusMode = mode;
```

ui/app/controllers/jobs/job/index.js
ChaiWithJai: `question`: Are you going to want to fetch data when this query parameter changes? If so, you'll need to update the associated `Route`?

ui/app/components/job-status/panel.hbs
```diff
@@ -1,27 +1,27 @@
...
       </button>
     </div>
   </div>
   <div class="boxed-section-body">
-    {{#if (eq this.mode "current")}}
+    {{#if (eq @statusMode "current")}}
       <h3 class="title is-4 running-allocs-title"><strong>{{@job.runningAllocs}}/{{this.totalAllocs}}</strong> Allocations Running</h3>
 
-      <JobStatus::AllocationStatusRow @allocBlocks={{this.allocBlocks}} />
+      aaa<JobStatus::AllocationStatusRow @allocBlocks={{this.allocBlocks}} />bbb
```
ChaiWithJai, line 24: Looks like we don't want `aaa` and `bbb` in prod.

ui/app/controllers/jobs/job/index.js
ChaiWithJai, line 39: `question`: Is this better off as a boolean?

ChaiWithJai (changes requested):
Minor changes required.

ui/app/controllers/jobs/job/index.js
philrenaud: I can try to make this tracked, but we don't typically do this in the Nomad UI I've found so I've got some concerns with bringing glimmer concepts in.

philrenaud: I don't think so; we don't save anything on perf and keeping it a specific string lets us extend this when we want to.

philrenaud: Also just from a legibility standpoint, `statusMode=historical` reads cleaner than `statusModeHistorical=true` for an end-user.

ChaiWithJai (approved):
My only critique here is that we're not dispatching a new network request upon updating this query parameter by calling a refresh in the associated route. We're relying on derived state computations for `allocations` and `taskGroups` for this. Thankfully, we do have watchers which **should** update upon the server detecting a change. However, as we've noted, some browsers can only keep 2 concurrent requests open and `jobs.job.index` is watching about half a dozen entities so we might be breaking on real-time updates and not realize that.

Issue #16338, opened 2023-03-05
Tasks using Docker drivers frequently failed to restart


dani:
### Nomad version
Nomad v1.5.0
BuildDate 2023-03-01T10:11:42Z
Revision fc40c491cacec3d8ec3f2f98cd82b9068a50797c

### Operating system and Environment details
AlmaLinux 8.7 x86_64
Nomad installed from the pre-built standalone binary
Docker CE 23.0.1 (from the official Docker repo)

### Issue

Since I've updated Nomad from 1.4.4 to 1.5.0, I have frequent errors when a task is to be restarted :

![image](https://user-images.githubusercontent.com/526826/222984091-b1c4028c-d29c-4a06-9d07-43cf839ab8f7.png)

It can be triggered by a template re-rendered with the default restart value for on_change, or by manually restarting a task from the GUI. The task is correctly stopped, but when starting it again, it fails.

### Reproduction steps
Not sure yet what the exact conditions to reproduce are. It doesn't seem to be 100% of the time. But I can reproduce it easily by restarting from the GUI an connect sidecar.

#### Expected Result
The task is correctly restarted

#### Actual Result
The task fails to be started and is marked as failed

### Job file (if appropriate)
Still trying to find a simple reproducer (can trigger it easily with internal jobs which I can't share)

### Nomad Client logs (if appropriate)
Sample logs when the error is triggered. In this example, I restarted the connect proxy sidecar task (connect-proxy-pharma-ws-stub) from the GUI
```
mars 05 21:32:01 ct-wn-4 nomad[579417]:     2023-03-05T21:32:01.096+0100 [INFO]  client.alloc_runner.task_runner: Task event: alloc_id=62698c1a-3e79-c080-5c82-902f7eb228d4 task=connect-proxy-pharma-ws-stub type="Restart Signaled" msg="User requested task to restart" failed=false
mars 05 21:32:01 ct-wn-4 nomad[579417]:     2023-03-05T21:32:01.314+0100 [INFO]  client.driver_mgr.docker: stopped container: container_id=2a748bc8df048a2dcee5a759369b9802fff0cc3510fab52cd711ab1a66f973ec driver=docker
[...]

PR #16352, opened 2023-03-06
docker: fix bug where network pause containers would be erroneously reconciled 

shoenig:
This PR adds tracking of pause containers to the docker driver, fixing a bug introduced by https://github.com/hashicorp/nomad/pull/15898 where the containers are now subject to dangling container reconciliation. The pause container created for allocs with docker tasks making use of bridge networking is not created in the same flow as a normal Task - which have a `TaskHandle` state. The set of tasks _not_ to reconcile was identified by scanning the set of these states, which does not include pause containers. 

To remedy this, this PR now tracks pause containers in their own little store. Since the Nomad Client may be restarted, we scan existing running containers on startup to reload the store from existing running containers. 

Fixes: https://github.com/hashicorp/nomad/issues/16338

### DIFF
--- a/drivers/docker/driver.go
+++ b/drivers/docker/driver.go
@@ -171,7 +171,7 @@ func NewDockerDriver(ctx context.Context, logger hclog.Logger) drivers.DriverPlu
 		ctx:             ctx,
 		logger:          logger,
 	}
-	go driver.recoverPauseContainers()
+	go driver.recoverPauseContainers(ctx)
 	return driver
 }
 
@@ -274,6 +274,9 @@ func (d *Driver) RecoverTask(handle *drivers.TaskHandle) error {
 	}
 
 	d.tasks.Set(handle.Config.ID, h)
+
+	// find a pause container?
+
 	go h.run()
 
 	return nil
@@ -745,7 +748,7 @@ func (d *Driver) containerBinds(task *drivers.TaskConfig, driverConfig *TaskConf
 	return binds, nil
 }
 
-func (d *Driver) recoverPauseContainers() {
+func (d *Driver) recoverPauseContainers(ctx context.Context) {
 	// On Client restart, we must rebuild the set of pause containers
 	// we are tracking. Basically just scan all containers and pull the ID from
 	// anything that has the Nomad Label and has Name with prefix "/nomad_init_".
@@ -757,7 +760,8 @@ func (d *Driver) recoverPauseContainers() {
 	}
 
 	containers, listErr := dockerClient.ListContainers(docker.ListContainersOptions{
-		All: false, // running only
+		Context: ctx,
+		All:     false, // running only
 		Filters: map[string][]string{
 			"label": {dockerLabelAllocID},
 		},
--- a/drivers/docker/driver.go
+++ b/drivers/docker/driver.go
@@ -20,6 +20,7 @@ import (
 	hclog "github.com/hashicorp/go-hclog"
 	multierror "github.com/hashicorp/go-multierror"
 	plugin "github.com/hashicorp/go-plugin"
+	"github.com/hashicorp/go-set"
 	"github.com/hashicorp/nomad/client/lib/cgutil"
 	"github.com/hashicorp/nomad/client/taskenv"
 	"github.com/hashicorp/nomad/drivers/docker/docklog"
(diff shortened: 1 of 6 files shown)


drivers/docker/driver.go
```diff
@@ -709,6 +745,39 @@ func (d *Driver) containerBinds(task *drivers.TaskConfig, driverConfig *TaskConf
...
+CONTAINER:
+	for _, c := range containers {
+		for _, name := range c.Names {
+			if strings.HasPrefix(name, "/nomad_init_") {
+				d.pauseContainers.add(c.ID)
+				continue CONTAINER
+			}
+		}
+	}
+}
```
tgross, line 779: I think this implementation can lead to us tracking pause containers that have been orphaned by their task. For example:
- the client stops
- the task fails, but the pause container is left running
- the client restarts
- the task cannot be recovered
- the pause container will be picked up here and tracked

We currently have the `RecoverTask` method in this file. Could we try to find the pause container for a task that's been successfully recovered there, so that we only track the pause containers for live tasks?

shoenig: I'm pretty sure this implementation is correct - a pause container is associated with an allocation, not a task. If a task cannot be recovered the alloc is replaced, and the destruction of the alloc will call [DestroyNetwork](https://github.com/hashicorp/nomad/blob/main/client/allocrunner/network_hook.go#L185) the implementation of which now purges the pause container associated with the alloc from the tracking store.

Testing described steps, 

```hcl
job "bug" {

  group "cache" {
    network {
      mode = "bridge"
      port "db" {
        to = 6379
      }
    }
    
    task "sleep" {
      driver = "docker"
      
      restart {
        attempts = 0
        mode = "fail"
      }
      
      config {
        image = "bash:5"
        command = "sleep"
        args = ["infinity"]
      }
      
      resources {
        cpu = 100
        memory = 32
      }
    }

    task "redis" {
      driver = "docker"

      config {
        image          = "redis:7"
        ports          = ["db"]
        auth_soft_fail = true
      }

      resources {
        cpu    = 100
        memory = 128
      }
    }
  }
}
```
</details>

```
Work/tickets/docker-infra 
➜ docker ps -a
CONTAINER ID   IMAGE                                      COMMAND                  CREATED          STATUS          PORTS     NAMES
5a7c52cb1854   redis:7                                    "docker-entrypoint.s…"   30 seconds ago   Up 29 seconds             redis-9b58cff0-d8ad-b1b0-186f-adcc2b53ed17
c4af36a0a5fa   bash:5                                     "docker-entrypoint.s…"   32 seconds ago   Up 31 seconds             sleep-9b58cff0-d8ad-b1b0-186f-adcc2b53ed17
86b102d2d006   gcr.io/google_containers/pause-amd64:3.1   "/pause"                 33 seconds ago   Up 32 seconds             nomad_init_9b58cff0-d8ad-b1b0-186f-adcc2b53ed17

Work/tickets/docker-infra 
➜ sudo pkill -9 nomad

Work/tickets/docker-infra 
➜ docker stop c4
c4

Work/tickets/docker-infra 10s
➜ docker ps -a
CONTAINER ID   IMAGE                                      COMMAND                  CREATED              STATUS                        PORTS     NAMES
5a7c52cb1854   redis:7                                    "docker-entrypoint.s…"   About a minute ago   Up About a minute                       redis-9b58cff0-d8ad-b1b0-186f-adcc2b53ed17
c4af36a0a5fa   bash:5                                     "docker-entrypoint.s…"   About a minute ago   Exited (137) 10 seconds ago             sleep-9b58cff0-d8ad-b1b0-186f-adcc2b53ed17
86b102d2d006   gcr.io/google_containers/pause-amd64:3.1   "/pause"                 About a minute ago   Up About a minute                       nomad_init_9b58cff0-d8ad-b1b0-186f-adcc2b53ed17

Work/tickets/docker-infra 
➜ # start nomad


Work/tickets/docker-infra 
➜ docker ps -a
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES
```

tgross: Ah, I forgot that we call `DestroyNetwork` if we failed to recover. That works :+1:

drivers/docker/driver.go
```diff
@@ -709,6 +745,39 @@ func (d *Driver) containerBinds(task *drivers.TaskConfig, driverConfig *TaskConf
...
+	// we are tracking. Basically just scan all containers and pull the ID from
+	// anything that has the Nomad Label and has Name with prefix "/nomad_init_".
+
+	_, dockerClient, err := d.dockerClients()
+	if err != nil {
+		d.logger.Error("failed to recover pause containers", "error", err)
+		return
+	}
+
+	containers, listErr := dockerClient.ListContainers(docker.ListContainersOptions{
```
tgross, line 759: `ListContainersOptions` has a `Context` field we could thread from the `Driver.ctx`. This is one of the places for the Docker client where it seems unambiguously correct to bail out.

tgross (approved):
LGTM. Left one last note about threading a context thru but that's fairly straightforward

shoenig:
Hi @gbolo we're currently expecting to release v1.5.1 with this fix, plus a few other important things on Monday the 13th.

gbolo:
@shoenig will this be back ported to v1.4.x? Seems like `1.4.5` is also affected

shoenig:
yes, good catch @gbolo!

PR #16353, opened 2023-03-06
Add k8s support for go-discover.

### LINKED ISSUE #16351: Enable k8s discovery support in go-discover
We are working towards running a nomad server cluster within our pre-existing kubernetes cluster. Just the server, and not agents. We plan on running agents/clients outside of kube, but leveraging kube as something we have pre-existing and easy to deploy stuff.

I'm not sure if this is an oversight, or something we're against, but it seems consul has this enabled: https://github.com/hashicorp/consul/blob/main/agent/retry_join.go#L110-L122 whereas nomad does not: https://github.com/hashicorp/nomad/blob/main/command/agent/command.go#L819

We're trying to run a cluster entirely without consul since, we can, and have no real need for consul in our architecture.

I'd be happy to submit a PR to enable this similar to how it's done in consul, but this sorta seems like maybe an oversight.

Pinging @schmichael tho. 👋

scottjab:
Resolves #16351

Method is largely the same as Consul's retry-join getDiscover method.


schmichael (commented):
Hi @scottjab! Thanks for the PR although you should really make @mattrobenolt do his own dirty work. :sweat_smile: 

The code looks great. The dependencies... add 23mb to the `nomad` binary. :disappointed: 

From `90mb -> 133mb` for the Nomad binary after running `make dev` on main (b3f75593517fe2ff660eafb45a7f055d55fc5c4f) vs this PR.

While we don't worry too much about our binary size, a quick unscientific peek at memory usage shows a 5-10MB bump in RSS for `sudo nomad agent -dev` on my machine with k8s nowhere in sight. I think we pay a similar (or worse) RSS penalty for AWS's SDK, but I'm hesitant to add more.

I'll chat with the team and see if anyone has ideas/opinions. Sadly I doubt there are many easy answers here unless you or someone else notices a way to trim down `go-discover`'s requirements.

mattrobenolt:
Yeah, that's very fair, I'm not a fan of pulling in the entirety of the kube client either for this.

My only justification is consul already does it? I guess that just means it'd be doubling up for that.

Right now, we're just running and building a fork of `nomad` with this patch. We can maintain a fork since I'm sure this path isn't going to be radically different. We also only need it for the server, and plan to use normal upstream binaries for the client.

Need a light version of the kube API client for use cases like this. :)

mattrobenolt:
Would an alternative option be to add an `exec` provider in go-discover? Then we could have something like:

```hcl
retry_join = [ "provider=exec cmd=/bin/thing" ]
```

just trying to think of some middleground here that we can agree on.

schmichael:
> My only justification is consul already does it? I guess that just means it'd be doubling up for that.

@lgfa29 noticed they also hardcoded a Consul const. Is this not causing you problems? Do you just set that annotation for Nomad?

https://github.com/hashicorp/go-discover/blob/70b45b3eb8b7a54c62311682dbe48a9259cb3abb/provider/k8s/k8s_discover.go#L27

> Would an alternative option be to add an `exec` provider in go-discover? Then we could have something like:
> 
> ```terraform
> retry_join = [ "provider=exec cmd=/bin/thing" ]
> ```
> 
> just trying to think of some middleground here that we can agree on.

Oooh, I like this. Makes for a nice escape hatch no matter what provider you use. consul-dataplane uses https://github.com/hashicorp/go-netaddrs instead of go-discover, and it's exactly that. Perhaps we could reuse it or at least match its interface?

Alternatively: DIYing a k8s provider without the SDK got support from folks at HashiCorp.

From glancing at [`k8s_discover.go`](https://github.com/hashicorp/go-discover/blob/70b45b3eb8b7a54c62311682dbe48a9259cb3abb/provider/k8s/k8s_discover.go#L108-L114) I think there's only 1 HTTP request required to get this information, so implementing that "manually" seems tenable...

...however I have no ideal complex the k8s client configuration is: https://github.com/hashicorp/go-discover/blob/70b45b3eb8b7a54c62311682dbe48a9259cb3abb/provider/k8s/k8s_discover.go#L66-L101

Reimplementing that manually could be really risky and error prone.

I'd definitely review and help to merge any upstream approaches you want to try. I don't think I can promise a solution from us at this point though.

mattrobenolt:
The kube API itself is the easy part, it's the auth that is kinda being glossed over and gotten for free.

https://github.com/hashicorp/go-discover/blob/70b45b3eb8b7a54c62311682dbe48a9259cb3abb/provider/k8s/k8s_discover.go#L19-L21

kube auth is... complex unfortunately.

TIL https://github.com/hashicorp/go-netaddrs though. Yeah, I think having some `exec` escape hatch would get us far without needing to bloat the nomad binary itself. I haven't looked much, could discovery be swapped out for netaddrs as some drop in replacement? Rather than adding `exec` support into `go-discover`.

Anyways, just kinda spitballing here and down the rabbithole. We're most likely going to just maintain a patch and build our own `nomad` binary in the short term.

mattrobenolt:
I spoke before looking at the code, `netaddrs` doesn't do much, and the docs say to exec out to `go-discover`. :)

So yeah, i think similarly passing along the same interface into go-discover to exec out to a script that spits back a list of IPs would seemingly be the least amount of friction.

scottjab:
Sorry for taking a bit to respond, I was AFK for most of yesterday.  

It is unfortunate that the k8s dependencies add too much bloat to nomad, but I totally understand.  I tend to agree with @mattrobenolt that maybe the least friction is allowing some kind of exec to get a list of IPs is a fine compromise. 

That said until that is figured out, we will probably maintain this patch as a fork.

PR #16362, opened 2023-03-07
scheduler: correctly detect inplace update with wildcard datacenters

tgross:
Wildcard datacenters introduced a bug where a system job with any wildcard datacenters will always be treated as a destructive update when we check whether a datacenter has been removed from the jobspec.

Includes updating the helper so that callers don't have to loop over the job's datacenters.

### DIFF
--- a/scheduler/util_test.go
+++ b/scheduler/util_test.go
@@ -320,6 +320,7 @@ func TestTaskUpdatedSpread(t *testing.T) {
 
 	require.False(t, tasksUpdated(j5, j6, name))
 }
+
 func TestTasksUpdated(t *testing.T) {
 	ci.Parallel(t)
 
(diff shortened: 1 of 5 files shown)


scheduler/util_test.go
```diff
@@ -972,6 +973,35 @@ func TestInplaceUpdate_Success(t *testing.T) {
 	}
 }
 
+func TestInplaceUpdate_WildcardDatacenters(t *testing.T) {
+	ci.Parallel(t)
+
+	state, ctx := testContext(t)
```
jrasell, line 979: nit: this variable name collides with an import

jrasell (approved):
LGTM! Very minor comment, feel free to ignore.

scheduler/util_test.go
tgross: Good catch, that always annoys me but I copy-pasted a lot of the setup code in this test :grinning: Fixed!

PR #16378, opened 2023-03-07
[ui] Perform common job tasks with keyboard shortcuts

philrenaud:
Traversal has been a strong-suit of keyboard navigation for long enough that I feel confident in its behaviour in the Nomad UI.

Now, this PR adds some basic action-running for jobs.

From the jobs index:
- run a new job with `r u n`

From a job page:
- exec with `e x e c`
- stop with `s t o p`
- purge with `p u r g e`
- start with `s t a r t`

image

### DIFF
--- a/ui/app/controllers/jobs/index.js
+++ b/ui/app/controllers/jobs/index.js
@@ -2,7 +2,7 @@
 import { inject as service } from '@ember/service';
 import { alias, readOnly } from '@ember/object/computed';
 import Controller from '@ember/controller';
-import { computed } from '@ember/object';
+import { computed, action } from '@ember/object';
 import { scheduleOnce } from '@ember/runloop';
 import intersection from 'lodash.intersection';
 import Sortable from 'nomad-ui/mixins/sortable';
@@ -20,6 +20,7 @@ export default class IndexController extends Controller.extend(
 ) {
   @service system;
   @service userSettings;
+  @service router;
 
   isForbidden = false;
 
@@ -253,5 +254,10 @@ export default class IndexController extends Controller.extend(
 
   setFacetQueryParam(queryParam, selection) {
     this.set(queryParam, serialize(selection));
+  }
+
+  @action
+  goToRun() {
+    this.router.transitionTo('jobs.run');
   }
 }
--- a/ui/app/services/keyboard.js
+++ b/ui/app/services/keyboard.js
@@ -342,7 +342,7 @@ export default class KeyboardService extends Service {
           this.displayHints = true;
         } else {
           if (!DISALLOWED_KEYS.includes(key)) {
-            this.addKeyToBuffer.perform(key, shifted);
+            this.addKeyToBuffer.perform(key, shifted, event);
           }
         }
       } else if (type === 'release') {
@@ -382,7 +382,7 @@ export default class KeyboardService extends Service {
    * @param {string} key
    * @param {boolean} shifted
    */
-  @restartableTask *addKeyToBuffer(key, shifted) {
+  @restartableTask *addKeyToBuffer(key, shifted, event) {
     // Replace key with its unshifted equivalent if it's a number key
     if (shifted && key in DIGIT_MAP) {
       key = DIGIT_MAP[key];
@@ -411,6 +411,7 @@ export default class KeyboardService extends Service {
             command.label === 'Show Keyboard Shortcuts' ||
             command.label === 'Hide Keyboard Shortcuts'
           ) {
+            event.preventDefault();
             command.action();
           }
         });
--- a/ui/app/templates/components/job-page/parts/title.hbs
+++ b/ui/app/templates/components/job-page/parts/title.hbs
@@ -23,7 +23,12 @@
         @confirmText="Yes, Stop Job"
         @confirmationMessage="Are you sure you want to stop this job?"
         @awaitingConfirmation={{this.stopJob.isRunning}}
-        @onConfirm={{perform this.stopJob}} />
+        @onConfirm={{perform this.stopJob}}
+        {{keyboard-shortcut 
+          label="Exec"
+          pattern=(array "s" "t" "o" "p")
+          action=(perform this.stopJob)
+        }} />
     {{else}}
       <TwoStepButton
         data-test-purge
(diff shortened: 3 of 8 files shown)


ui/app/services/keyboard.js
```diff
@@ -411,6 +411,7 @@ export default class KeyboardService extends Service {
             command.label === 'Show Keyboard Shortcuts' ||
             command.label === 'Hide Keyboard Shortcuts'
           ) {
+            event.preventDefault();
```
philrenaud, line 414: Added this so that the letter `n` (from having typed `run`) doesn't show up in the job template editor box (keyboard service eagerly performs a task on keydown, but typing waits for keyup)

ui/app/templates/components/job-page/parts/title.hbs
```diff
@@ -23,7 +23,12 @@
         @confirmText="Yes, Stop Job"
         @confirmationMessage="Are you sure you want to stop this job?"
         @awaitingConfirmation={{this.stopJob.isRunning}}
-        @onConfirm={{perform this.stopJob}} />
+        @onConfirm={{perform this.stopJob}}
+        {{keyboard-shortcut 
+          label="Exec"
```
lgfa29, line 28: Are these `label="Exec"` copy-paste errors?

ui/app/templates/components/job-page/parts/title.hbs
```diff
@@ -33,7 +38,13 @@
         @confirmText="Yes, Purge Job"
         @confirmationMessage="Are you sure? You cannot undo this action."
         @awaitingConfirmation={{this.purgeJob.isRunning}}
-        @onConfirm={{perform this.purgeJob}} />
+        @onConfirm={{perform this.purgeJob}}
+        {{keyboard-shortcut 
+          label="Exec"
+          pattern=(array "p" "u" "r" "g" "e")
+          action=(perform this.purgeJob)
+        }}
```
lgfa29, line 46: It's a pretty long shortcut, but I'm kind of wary about the potential for unintended destructive operations here. Would it be possible to still have the confirmation pop-up? Keyboard shortcuts make things quick but sometimes it's good to slow down and double-check you are in the right namespace or something like that 😅

ui/app/templates/components/job-page/parts/title.hbs
```diff
@@ -42,7 +53,13 @@
         @confirmText="Yes, Start Job"
         @confirmationMessage="Are you sure you want to start this job?"
         @awaitingConfirmation={{this.startJob.isRunning}}
-        @onConfirm={{perform this.startJob}} />
+        @onConfirm={{perform this.startJob}}
+        {{keyboard-shortcut 
+          label="Exec"
+          pattern=(array "s" "t" "a" "r" "t")
+          action=(perform this.startJob)
+        }}
```
lgfa29, line 61: Would it be possible to add some kind of toast confirmation on some of these? They happen rather quickly and it may be hard to tell if they were triggered by a shortcut or something that happened in the backend at the same time.

lgfa29 (changes requested):
I think the `label` is the main thing that needs to be fixed (though I'm not sure what's the impact of having them wrong).

As for the confirmation I would gather a second opinion. My spidey-sense makes me worry a little about them in production.

ui/app/templates/components/job-page/parts/title.hbs
philrenaud: Yes! Thank you

ui/app/templates/components/job-page/parts/title.hbs
philrenaud: I've been giving this some thought. A couple things:
- Our two-step buttons are probably not a convention that we're going to continue with in the near future; [Helios](https://helios.hashicorp.design/components/modal) indicates that we should either use modals to warn, or toast/modals to make undoing easy. Once we start using modals, adding them to a queue of keyboard shortcut actions will be possible, but still not an ideal flow. As it stands now, the visual effect probably wouldn't be enough to make the user notice and they might have assumed the purge command went through without noticing otherwise.
- The act of typing the full word is a preeeeeetty deliberate action - something much harder to do than accidentally mouse-clicking or tab+entering on the button while enroute to do something else on the page.

As such, I think I'd like to move forward with this change. I am worried about the "double-checking you're in the right namespace" problem as well but the benefit here seems to outweigh the potential cost.

ui/app/templates/components/job-page/parts/title.hbs
philrenaud: This is a great suggestion, yeah.

lgfa29 (approved):

ui/app/controllers/jobs/index.js
```diff
@@ -20,6 +20,7 @@ export default class IndexController extends Controller.extend(
 ) {
   @service system;
   @service userSettings;
+  @service router;
```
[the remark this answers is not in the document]
philrenaud: > I think we got the action API in the Keyboard

Can you expand on what you mean here? There are several non-router actions that use this modifier throughout the app.

Would your suggestion have been to instead inject the router into the service?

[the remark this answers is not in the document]
ChaiWithJai: There's 27 uses of this. 25 of them use are transitions. The 2 that are not, are changing a value that should be a query parameter.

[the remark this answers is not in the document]
ChaiWithJai: Using partial application:

```
function someAction() {
  const [routeName, ...args] = arguments;
  this.routerTransitionTo(routeName, { /* apply arguments here */}
}
```

And then use the `fn` helper to partially apply arguments as normal in the template layer.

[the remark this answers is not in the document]
philrenaud: > changing a value that should be a query parameter.
This strikes me as wrong in the case of what I think you're referring to (show/hide individual variable).

Regardless, if you're suggesting that the action API in the keybaord service should be specifically for routing, that's a pretty limiting way of viewing this feature and isn't exactly what I'd call future-proof.

ui/app/templates/components/job-page/parts/title.hbs
```diff
@@ -42,7 +53,13 @@
         @confirmText="Yes, Start Job"
         @confirmationMessage="Are you sure you want to start this job?"
         @awaitingConfirmation={{this.startJob.isRunning}}
-        @onConfirm={{perform this.startJob}} />
+        @onConfirm={{perform this.startJob}}
+        {{keyboard-shortcut 
+          label="Start"
+          pattern=(array "s" "t" "a" "r" "t")
```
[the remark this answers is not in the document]
philrenaud: I suspect under the hood mousetrap is doing something like parsing their string into an array; the basic mechanism of a key buffer is that it'll have to step through it iteratively. More efficient to do with an array than a string.

I could have the template interface for this take in a string and split it in the keyboard service, but that seems like an hasty abstraction.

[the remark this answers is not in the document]
philrenaud: You're right that the keyboard shortcut modifier has pattern set to an empty string, but it's not strongly typed or typechecked. The [typedef in our keyboard service](https://github.com/hashicorp/nomad/blob/ee057488f0aa8a3d33f155ea4f382cdd4abc5480/ui/app/services/keyboard.js#L21) is typechecked and provides a better example.

PR #16379, opened 2023-03-07
packaging: set limited permissions on nomad data directory

shoenig:
Setting `0700` on the Nomad `data_dir` brings it in line with Nomad's production security hardening guide.

https://developer.hashicorp.com/nomad/docs/install/production/requirements#hardening-nomad


shoenig:
TBD - we'll likely end up splitting Nomad Client and Server specific Linux packages to make it possible to adhere to best practices for each variant.

PR #16380, opened 2023-03-08
cli: remove hard requirement on `list-jobs`

### LINKED ISSUE #16356: Review ACL requirements for `nomad job` subcommands
https://github.com/hashicorp/nomad/pull/16306 refactored the `nomad job` subcommands to share logic for job prefix match, but this logic requires the `list-job` permission to be able to query jobs by prefix.

As noted in the CHANGELOG entries, some of the commands did not do prefix match, meaning that they did not require the `list-job` capability. https://github.com/hashicorp/nomad/pull/16306 broke this and so the logic must be fixed to prevent this breaking change.

lgfa29:
Most job subcommands allow for job ID prefix match as a convenience functionality so users don't have to type the full job ID.

But this introduces a hard ACL requirement that the token used to run these commands have the `list-jobs` permission, even if the token has enough permission to execute the basic command action and the user passed an exact job ID.

This change softens this requirement by not failing the prefix match in case the request results in a permission denied error and instead using the information passed by the user directly.

Closes #16356

### DIFF
--- a/.changelog/16380.txt
+++ b/.changelog/16380.txt
@@ -1,0 +1,3 @@
+``release-note:improvement
+cli: Remove requirement for `list-jobs` capability on several job subcommands that didn't strictly needed it
+```
--- a/command/job_allocs.go
+++ b/command/job_allocs.go
@@ -19,8 +19,9 @@ Usage: nomad job allocs [options] <job>
 
   Display allocations for a particular job.
 
-  When ACLs are enabled, this command requires a token with the 'read-job' and
-  'list-jobs' capabilities for the job's namespace.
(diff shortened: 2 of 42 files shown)


tgross (approved):
LGTM

That `read-job` doesn't imply `list-jobs` seems like an ACL design mistake to me, but I suspect most people who want to allow `read-job` just set `read` on the namespace and call it a day. It's the same way for `csi-read-volume` (but not Variables, oh well). In any case, it's probably not something we could change without breaking the ACL policies of the folks who really do care about that distinction.

schmichael (approved):

PR #16384, opened 2023-03-08
[ui] Fix: New toast notifications no longer last forever

philrenaud:
Also, removed an errant console.log

### DIFF
--- a/ui/app/services/notifications.js
+++ b/ui/app/services/notifications.js
@@ -29,7 +29,7 @@ export default class NotificationsService extends FlashService {
     }
 
     if (!('sticky' in notificationObject)) {
-      notificationObject.sticky = true;
+      notificationObject.sticky = false;
     }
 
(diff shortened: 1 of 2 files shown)


ui/app/services/notifications.js
```diff
@@ -29,7 +29,7 @@ export default class NotificationsService extends FlashService {
     }
 
     if (!('sticky' in notificationObject)) {
-      notificationObject.sticky = true;
+      notificationObject.sticky = false;
     }
```
lgfa29, line 33: `false` seems to be the [default for `sticky`](https://github.com/adopted-ember-addons/ember-cli-flash#custom-messages-api) so we can probably remove this altogether?

lgfa29 (approved):

ui/app/services/notifications.js
philrenaud: Good call; removed!

Issue #14850, opened 2022-10-07
Allocation stopped immediately after start


valodzka:
### Nomad version
1.3.6

### Issue
Sometimes during deployment I see that freshly started allocation immediately stopped by nomad. During the same deployment multiple allocations with identical alloc index (name) created. Example, all allocs share same Eval ID and Job Version (full alloc status here: https://gist.github.com/valodzka/231f3942203a5f39528b52241905a06e):

Alloc 1:

```
ID                  = d1421efd-6b25-295a-4832-348c09955a8f
Name                = bk.client[35]
Client Status       = complete
Desired Description = alloc not needed due to job update
....

Recent Events:
Time                  Type      Description
2022-10-07T15:09:45Z  Killing   Sent interrupt. Waiting 2m0s before force killing
2022-10-07T15:09:45Z  Received  Task received by client
```

Alloc 2:
```
ID                  = f278d901-b4d0-8b7b-8f20-4f1c46a36a10
Name                = bk.client[24]
Client Status       = running

Recent Events:
Time                  Type        Description
2022-10-07T15:10:00Z  Started     Task started by client
2022-10-07T15:09:57Z  Task Setup  Building Task Directory
2022-10-07T15:09:45Z  Received    Task received by client

Placement Metrics
  * Constraint "${meta.only_low_cpu_tasks} = false": 1 nodes excluded by filter
  * Resources exhausted on 31 nodes
  * Dimension "network: port collision" exhausted on 1 nodes
  * Dimension "cpu" exhausted on 13 nodes
  * Dimension "memory" exhausted on 17 nodes
```

Alloc 3:
```
ID                  = 246dec93-d65e-9498-1d69-816d727083d7
Name                = bk.client[24]
Client Status       = running

Recent Events:
Time                  Type        Description
2022-10-07T15:09:57Z  Started     Task started by client
2022-10-07T15:09:57Z  Task Setup  Building Task Directory
2022-10-07T15:09:45Z  Received    Task received by client
```

What is strange here:
1. Alloc 2 and 3 have identical alloc index/name (with same job version and no canaries used)
2. Alloc 1 killed the same second as it was started
3. Alloc 1 "Desired description" is "alloc not needed due to job update" but I'm sure that job definition wasn't updated
4. Alloc 2 has `Dimension "network: port collision" exhausted on 1 nodes`. Not sure if it's relevant here but I guess it might be related.

Issue #12797 might be relevant.

#### Expected Result

[...]

Issue #8421, opened 2020-07-10
Setting reserved_ports results in jobs being blocked 


evandam:
### Nomad version

```
Nomad v0.11.3
```

### Issue

Similar to https://github.com/hashicorp/nomad/issues/1046, we're seeing jobs stuck in pending state when `reserved_ports` is set.

<details>
  <summary><b>/etc/nomad/client.json</b></summary>

```json
{
  "client": {
    "disable_remote_exec": false,
    "enabled": true,
    "max_kill_timeout": "30s",
    "reserved": {
      "reserved_ports": "0-19999,24224"
    },
    "meta": {
      "chef_role": "nomad-compute",
      "role": "nomad-compute"
    },
    "no_host_uuid": false,
    "node_class": "compute",
    "server_join": {
      "retry_join": [
        "provider=aws tag_key=role tag_value=nomad-server region=us-west-2 addr_type=private_v4"
      ]
    }
  }
}
```
</details>

Running a job results in it being stuck in a pending state, creating evals with errors like so:

```
ID                 = 2a3d639d-d030-7313-923f-829ad9b8e125
Create Time        = 2020-07-09T19:10:32-07:00
Modify Time        = 2020-07-09T19:10:32-07:00
Status             = blocked
Status Description = created due to placement conflicts
Type               = batch
TriggeredBy        = max-plan-attempts
Previous Eval      = e9674379-8362-cf7f-c004-d42eab81a88b
Priority           = 50
Placement Failures = N/A - In Progress
Previous Eval      = e9674379-8362-cf7f-c004-d42eab81a88b
Next Eval          = <none>
Blocked Eval       = <none>
```

```
ID                 = e9674379-8362-cf7f-c004-d42eab81a88b
Create Time        = 2020-07-09T19:09:32-07:00
Modify Time        = 2020-07-09T19:10:32-07:00
Status             = failed
Status Description = maximum attempts reached (2)
Type               = batch
TriggeredBy        = max-plan-attempts
Previous Eval      = 40c308c7-2caa-ee63-0e6e-ad6e7dea9b71
Priority           = 50
Placement Failures = false
Previous Eval      = 40c308c7-2caa-ee63-0e6e-ad6e7dea9b71
Next Eval          = <none>
Blocked Eval       = 2a3d639d-d030-7313-923f-829ad9b8e125
```

Here is the job HCL being used:

<details>
  <summary><b>prod-platform-core-rake.hcl</b></summary>

```
[...]

PR #16401, opened 2023-03-08
scheduling: prevent self-collision in dynamic port network offerings

### LINKED ISSUE #9506: intermittent error "reserved port collision" when running new job
### Nomad version
Nomad v1.0.0-beta3 (fcb32ef7ba037b2b83e7a0fd1c53c7410a2990db)

Distributor ID: Debian
Description:    Debian GNU/Linux 10 (buster)
Release:        10
Codename:       buster

### Issue
Sometimes, whenever I update or run a new job. I got error `max-plan-attempts failed` on its UI

### Reproduction steps
Just run a job file

### Job file (if appropriate)

Taken from: https://www.hashicorp.com/blog/consul-connect-native-tasks-in-hashicorp-nomad-0-12

```
job "cn-demo" {
      datacenters = ["dc1"]

      group "generator" {
        network {
          port "api" {}
        }

        service {
          name = "uuid-api"
          port = "${NOMAD_PORT_api}"

          connect {
            native = true
          }
        }

        task "generate" {
          driver = "docker"

          config {
            image        = "hashicorpnomad/uuid-api:v3"
            network_mode = "host"
          }

          env {
            BIND = "0.0.0.0"
            PORT = "${NOMAD_PORT_api}"
          }
        }
      }

      group "frontend" {
        network {
          port "http" {
            static = 9800
          }
        }

        service {
          name = "uuid-fe"
          port = "9800"

          connect {
            native = true
          }
        }

        task "frontend" {
          driver = "docker"

          config {
            image        = "hashicorpnomad/uuid-fe:v3"
            network_mode = "host"
          }

          env {
            UPSTREAM = "uuid-api"
            BIND     = "0.0.0.0"
            PORT     = "9800"
          }
        }
      }
    }
```

```
2020-12-03T05:32:38.511+0100 [DEBUG] http: request complete: method=GET path=/v1/agent/health?type=client duration=442.672µs
    2020-12-03T05:32:48.516+0100 [DEBUG] http: request complete: method=GET path=/v1/agent/health?type=client duration=1.008414ms
    2020-12-03T05:32:58.534+0100 [DEBUG] http: request complete: method=GET path=/v1/agent/health?type=client duration=900.343µs
```

```
2020-12-03T05:33:01.954+0100 [DEBUG] worker: submitted plan for evaluation: eval_id=098a520b-4cd9-78a3-6ba2-613d94b33823
    2020-12-03T05:33:01.954+0100 [DEBUG] worker: refreshing state: refresh_index=345 eval_id=098a520b-4cd9-78a3-6ba2-613d94b33823
    2020-12-03T05:33:01.954+0100 [DEBUG] worker.service_sched: refresh forced: eval_id=098a520b-4cd9-78a3-6ba2-613d94b33823 job_id=cn-demo namespace=default
    2020-12-03T05:33:01.954+0100 [DEBUG] worker.service_sched: reconciled current state with desired state: eval_id=098a520b-4cd9-78a3-6ba2-613d94b33823 job_id=cn-demo namespace=default results="Total changes: (place 2) (destructive 0) (inplace 0) (stop 0)
Desired Changes for "generator": (place 1) (inplace 0) (destructive 0) (stop 0) (migrate 0) (ignore 0) (canary 0)
Desired Changes for "frontend": (place 1) (inplace 0) (destructive 0) (stop 0) (migrate 0) (ignore 0) (canary 0)"
    2020-12-03T05:33:01.987+0100 [DEBUG] nomad: plan for node rejected: node_id=9fa84370-c713-b914-d329-f6485951cddc reason="reserved port collision" eval_id=098a5
```

### Nomad Server Config

```
bind_addr = "0.0.0.0"
log_level = "DEBUG"
data_dir = "/etc/nomad"
name = "vmi483776.contaboserver.net"
server {
  enabled = true
  bootstrap_expect = 1
  raft_protocol = 3
}
advertise {
  http = "161.97.158.38"
  rpc = "161.97.158.38"
  serf = "161.97.158.38"
}
consul {
  address = "127.0.0.1:8500"
  auto_advertise = true
  server_auto_join = true
  client_auto_join = true
}
autopilot {
  cleanup_dead_servers      = true
  last_contact_threshold    = "200ms"
  max_trailing_logs         = 2500
  server_stabilization_time = "10s"
  enable_redundancy_zones   = false
  disable_upgrade_migration = false
  enable_custom_upgrades    = false
}
```

### Nomad Client config

```
bind_addr = "0.0.0.0"
log_level = "DEBUG"
data_dir = "/etc/nomad"
name = "vmi483778.contaboserver.net"
client {
  enabled = true
  reserved {
    reserved_ports = "22,53,68,111,8301,8500,8600,32768-60999"
  }
}
advertise {
  http = "161.97.158.40"
  rpc = "161.97.158.40"
  serf = "161.97.158.40"
}
consul {
  address = "127.0.0.1:8500"
  auto_advertise = true
  server_auto_join = true
  client_auto_join = true
}
```

### Notes
Btw you can really visit this url: http://161.97.158.38:4646/ui/jobs/cn-demo. It's my testbed server for this nomad beta version. It's opened and unauthenticated, in which you can try to see it live

Thanks

tgross:
Fixes #9506 #14850 (and several internally-tracked issues)

When the scheduler tries to find a placement for a new allocation, it iterates over a subset of nodes. For each node, we populate a `NetworkIndex` bitmap with the ports of all existing allocations and any other allocations already proposed as part of this same evaluation via its `SetAllocs` method. Then we make an "ask" of the `NetworkIndex` in `AssignPorts` for any ports we need and receive an "offer" in return. The offer will include both static ports and any dynamic port assignments.

The `AssignPorts` method was written to support group networks, and it shares code that selects dynamic ports with the original `AssignTaskNetwork` code. `AssignTaskNetwork` can request multiple ports from the bitmap at a time. But `AssignPorts` requests them one at a time and does not account for possible collisions, and doesn't return an error in that case.

What happens next varies:

1. If the scheduler doesn't place the allocation on that node, the port conflict is thrown away and there's no problem.
2. If the node is picked and this is the only allocation (or last allocation), the plan applier will reject the plan when it calls `SetAllocs`, as we'd expect.
3. If the node is picked and there are additional allocations in the same eval that iterate over the same node, their call to `SetAllocs` will detect the impossible state and the node will be rejected. This can have the puzzling behavior where a second task group for the job without any networking at all can hit a port collision error!

It looks like this bug has existed since we implemented group networks, but there are several factors that add up to making the issue rare for many users yet frustratingly frequent for others:

* You're more likely to hit this bug the more tightly packed your range for dynamic ports is. With 12000 ports in the range by default, many clusters can avoid this for a long time.
* You're more likely to hit case (3) for jobs with lots of allocations or if a scheduler has to iterate over a large number of nodes, such as with system jobs, jobs with `spread` blocks, or (sometimes) jobs using `unique` constraints.

For unlucky combinations of these factors, it's possible that case (3) happens repeatedly, preventing scheduling of a given job until a client state change (ex. restarting the agent so all its allocations are rescheduled elsewhere) re-opens the range of dynamic ports available.

This changeset:

* Fixes the bug by accounting for collisions in dynamic port selection in `AssignPorts`.
* Adds test coverage for `AssignPorts`, expands coverage of this case for the deprecated `AssignTaskNetwork`, and tightens the dynamic port range in a scheduler test for spread scheduling to more easily detect this kind of problem in the future.
* Adds a `String()` method to `Bitmap` so that any future "screaming" log lines have a human-readable list of used ports.

### DIFF
--- a/nomad/structs/bitmap.go
+++ b/nomad/structs/bitmap.go
@@ -88,7 +88,7 @@ func (b Bitmap) IndexesInRangeFiltered(set bool, from, to uint, filter []int) []
 	var indexes []int
 	for i := from; i <= to && i < b.Size(); i++ {
 		c := b.Check(i)
-		if c && set || !c && !set {
+		if c == set {
 			if len(filter) < 1 || !slices.Contains(filter, int(i)) {
 				indexes = append(indexes, int(i))
 			}
--- a/nomad/structs/bitmap.go
+++ b/nomad/structs/bitmap.go
@@ -1,6 +1,10 @@
 package structs
 
-import "fmt"
+import (
+	"fmt"
+
+	"golang.org/x/exp/slices"
+)
 
 // Bitmap is a simple uncompressed bitmap
 type Bitmap []byte
@@ -75,4 +79,26 @@ func (b Bitmap) IndexesInRange(set bool, from, to uint) []int {
 	}
 
 	return indexes
+}
+
+// IndexesInRangeFiltered returns the indexes in which the values are either set
+// or unset based on the passed parameter in the passed range, and do not appear
+// in the filter slice
+func (b Bitmap) IndexesInRangeFiltered(set bool, from, to uint, filter []int) []int {
+	var indexes []int
+	for i := from; i <= to && i < b.Size(); i++ {
+		c := b.Check(i)
+		if c && set || !c && !set {
+			if len(filter) < 1 || !slices.Contains(filter, int(i)) {
+				indexes = append(indexes, int(i))
+			}
+		}
+	}
+
+	return indexes
+}
+
+// String represents the Bitmap the same as slice of the Bitmap's set values
+func (b Bitmap) String() string {
+	return fmt.Sprintf("%v", b.IndexesInRange(true, 0, b.Size()))
 }
--- a/nomad/structs/network.go
+++ b/nomad/structs/network.go
@@ -758,7 +758,7 @@ func getDynamicPortsStochastic(nodeUsed Bitmap, portsInOffer []int, minDynamicPo
 		}
 		// the pick conflicted with a previous pick that hasn't been saved to
 		// the index yet
-		if portsInOffer != nil && slices.Contains(portsInOffer, randPort) {
+		if slices.Contains(portsInOffer, randPort) {
 			goto PICK
 		}
 
--- a/nomad/structs/network.go
+++ b/nomad/structs/network.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 
 	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
 )
 
 const (
@@ -502,6 +503,7 @@ func incIP(ip net.IP) {
 // AssignTaskNetwork supports the deprecated task.resources.network block.
 func (idx *NetworkIndex) AssignPorts(ask *NetworkResource) (AllocatedPorts, error) {
 	var offer AllocatedPorts
+	var portsInOffer []int
 
 	// index of host network name to slice of reserved ports, used during dynamic port assignment
 	reservedIdx := map[string][]Port{}
@@ -543,6 +545,7 @@ func (idx *NetworkIndex) AssignPorts(ask *NetworkResource) (AllocatedPorts, erro
 		}
 
 		offer = append(offer, *allocPort)
+		portsInOffer = append(portsInOffer, allocPort.Value)
 	}
 
 	for _, port := range ask.DynamicPorts {
@@ -554,10 +557,14 @@ func (idx *NetworkIndex) AssignPorts(ask *NetworkResource) (AllocatedPorts, erro
 			// lower memory usage.
 			var dynPorts []int
 			// TODO: its more efficient to find multiple dynamic ports at once
-			dynPorts, addrErr = getDynamicPortsStochastic(used, idx.MinDynamicPort, idx.MaxDynamicPort, reservedIdx[port.HostNetwork], 1)
+			dynPorts, addrErr = getDynamicPortsStochastic(
+				used, portsInOffer, idx.MinDynamicPort, idx.MaxDynamicPort,
+				reservedIdx[port.HostNetwork], 1)
 			if addrErr != nil {
 				// Fall back to the precise method if the random sampling failed.
-				dynPorts, addrErr = getDynamicPortsPrecise(used, idx.MinDynamicPort, idx.MaxDynamicPort, reservedIdx[port.HostNetwork], 1)
+				dynPorts, addrErr = getDynamicPortsPrecise(used, portsInOffer,
+					idx.MinDynamicPort, idx.MaxDynamicPort,
+					reservedIdx[port.HostNetwork], 1)
 				if addrErr != nil {
 					continue
 				}
@@ -583,6 +590,7 @@ func (idx *NetworkIndex) AssignPorts(ask *NetworkResource) (AllocatedPorts, erro
 			return nil, fmt.Errorf("no addresses available for %s network", port.HostNetwork)
 		}
 		offer = append(offer, *allocPort)
+		portsInOffer = append(portsInOffer, allocPort.Value)
 	}
 
 	return offer, nil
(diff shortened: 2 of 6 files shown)


nomad/structs/network.go
```diff
@@ -554,10 +557,14 @@ func (idx *NetworkIndex) AssignPorts(ask *NetworkResource) (AllocatedPorts, erro
 			// lower memory usage.
 			var dynPorts []int
 			// TODO: its more efficient to find multiple dynamic ports at once
```
tgross, line 559: Note to reviewers: this comment is really only true in the case where the dynamic range is full enough that we fail to get randomly selected ports. Optimizing for this will involve heavily refactoring this method to figure out the count of dynamic ports we need (per network/address), getting the offered ports, and then dolling them out for each of the asked ports. I want to do this performance work in the very near term, but not in the same PR as a major bug fix.

nomad/structs/bitmap.go
```diff
@@ -76,3 +80,25 @@ func (b Bitmap) IndexesInRange(set bool, from, to uint) []int {
...
 }
+
+// IndexesInRangeFiltered returns the indexes in which the values are either set
+// or unset based on the passed parameter in the passed range, and do not appear
+// in the filter slice
+func (b Bitmap) IndexesInRangeFiltered(set bool, from, to uint, filter []int) []int {
+	var indexes []int
+	for i := from; i <= to && i < b.Size(); i++ {
+		c := b.Check(i)
+		if c && set || !c && !set {
```
schmichael, line 91:
```suggestion
		if c == set {
```
but it might be best to just match the above func and leave it be.

nomad/structs/network.go
```diff
@@ -742,6 +756,12 @@ func getDynamicPortsStochastic(nodeUsed Bitmap, minDynamicPort, maxDynamicPort i
 				goto PICK
 			}
 		}
+		// the pick conflicted with a previous pick that hasn't been saved to
+		// the index yet
+		if portsInOffer != nil && slices.Contains(portsInOffer, randPort) {
```
schmichael, line 761: The nil check (and len check above) aren't technically necessary as Contains uses `range` which just doesn't loop over empty slices.

A map or set would work for portsInOffer too, but I suspect it should be small enough that the slice is more/as efficient for common cases (<10 ports per offer).

schmichael (approved):
Congratulations on tracking down this nasty and persistent bug!

nomad/structs/network.go
shoenig: Sounds like a fun problem! I would even go so far as to suggest reconsidering the use of a `Bitmap`, perhaps instead thinking of our own data structure that provides an efficient API surface that actually makes sense here (eliminating the need for random tries with precise fallback). No reason it can't be done, it's just off the beaten path.

nomad/structs/network.go
shoenig: The real advantage of `Set` is it makes code like this more readable, e.g.

`slices.Contains(portsInOffer, randPort)`

becomes

`portsInOffer.Contains(randPort)`

while _also_ being more efficient `O(1)` vs `O(n)` lookups - though it would be interesting to benchmark here!

shoenig (approved):
LGTM, awesome work @tgross!

nomad/structs/bitmap.go
tgross: Wow this got copy-and-pasted from the other function and I didn't even notice how weird a thing to do this is. Will fix.

nomad/structs/network.go
tgross: Fixed the nil check (not sure which len check you mean?). I've kept the slice over the map/set because as you noted it doesn't make a meaningful performance difference at the size we're talking about and on top of that when I do the optimization pass mentioned in https://github.com/hashicorp/nomad/pull/16401/files#r1130056010 the return value ends up being a slice anyways.

tgross: Eh, I was writing this comment still when @shoenig posted. I'm sold... I'll swap it out for a set.

tgross:
> Sounds like a fun problem! I would even go so far as to suggest reconsidering the use of a `Bitmap`, perhaps instead thinking of our own data structure that provides an efficient API surface that actually makes sense here (eliminating the need for random tries with precise fallback). No reason it can't be done, it's just off the beaten path.

Yeah I was thinking about this and one of the interesting things about using a `Bitmap` is that it uses a fixed but large amount of space, it's cheap to check, and it's cheap to get a new port _so long as the bitmap is sparsely-populated_. So it has a weird performance cliff when the bitmap gets full. We'll want to benchmark the space vs time tradeoff at varying levels of sparseness.

tgross:
> I stumbled across https://github.com/hashicorp/nomad/issues/8421 while spelunking for other network issues. Think we could close it as well? Not sure this is the exact fix for it, but combined with other work in the past 2 years I doubt that old issue is relevant anymore?

Agreed. Done!

PR #16402, opened 2023-03-09
allocrunner: fix health check monitoring for Consul services

### LINKED ISSUE #16307: [1.5.0] Job scrolling upgrade compatibility failure
### Nomad version
1.5.0

### Issue
1.Upgrade from 1.4.4 to 1.5.0
2.Release a new version of the job
3.image

### job
```
job "srv-third" {
  datacenters = ["dc1"]
  type        = "service"
  namespace   = "ai-dev"

  update {
    max_parallel     = 1
    min_healthy_time = "10s"
    healthy_deadline = "3m"
    auto_revert      = true
    auto_promote     = true
    canary           = 1
  }

  reschedule {
    attempts       = 15
    interval       = "1h"
    delay          = "15s"
    delay_function = "exponential"
    max_delay      = "120s"
    unlimited      = false
  }

  group "service" {

    restart {
      interval = "3m"
      attempts = 3
      delay    = "15s"
      mode     = "delay"
    }

    service {
      name         = "${NOMAD_JOB_NAME}"
      port         = "http"
      address_mode = "host"

      check {
        type           = "http"
        port           = "http"
        path           = "/v1.0/healthz"
        interval       = "12s"
        timeout        = "6s"

        check_restart {
          limit           = 3
          grace           = "10s"
        }
      }
    }

    task "app" {
      driver = "docker"
      config {
        image   = "alpine:3.15"
        command = "local/app"
        ports   = ["app"]
        args = [

        ]
      }
    }


    task "daprd" {
      lifecycle {
        hook = "poststart"
        sidecar = true
      }
      driver = "docker"

      config {
        image   = "alpine:3.15"
        ports   = ["http", "grpc"]
        command = "local/daprd"
        args = [
        ...
        ]
      }
    }
  }
}
```

### LINKED ISSUE #16382: Nomad does not mark deployment as successful with dynamic service names
### Nomad version
Output from `nomad version`
Nomad v1.4.5,  v1.5.0
Linux Mint 21.1
### Issue
Services with dynamic names never report healthy deployment status, although services are healthy in Consul view. Seems to be occuring after #15822 , could it be that dynamic service name is not being evaluated before validating the service checks?
#### Expected Result
Service healthiness result from Consul is forwarded to Nomad and deployment is successful.
#### Actual Result
Service healthiness is not received by Nomad and deployment is marked as failed after a while.
### Job file (if appropriate)
A service which uses custom name that is dynamically evaluated (NOMAD_GROUP_NAME, NOMAD_JOB_NAME or similar)
```
service {
  name = "${NOMAD_JOB_NAME}"
  port = "grpc"
  check {
    type = "grpc"
    interval = "10s"
    timeout = "5s"
    check_restart {
      limit = 3
      grace = "90s"
      ignore_warnings = false
    }
  }
}
```

lgfa29:
Services must be canonicalized and have its values interpolated before comparing with the values returned by Consul.

Close #16382
Close #16307

### DIFF
--- a/client/allochealth/tracker.go
+++ b/client/allochealth/tracker.go
@@ -519,18 +519,18 @@ OUTER:
 
 		// interpolate services to replace runtime variables
 		consulServices := t.tg.ConsulServices()
-		interpolatedServices := make([]*structs.Service, len(consulServices))
-		for i, service := range consulServices {
+		interpolatedServices := make([]*structs.Service, 0, len(consulServices))
+		for _, service := range consulServices {
 			env := t.taskEnvs[service.TaskName]
 			if env == nil {
 				// This is not expected to happen, but guard against a nil
-				// task environment by using the group environment since it has
-				// most of the same values.
-				t.logger.Warn("task environment not found, using group level environment for variable interpolation",
+				// task environment that could case a panic.
+				t.logger.Error("failed to interpolate service runtime variables: task environment not found",
 					"alloc_id", t.alloc.ID, "task", service.TaskName)
-				env = t.taskEnvs[""]
+				continue
 			}
-			interpolatedServices[i] = taskenv.InterpolateService(env, service)
+			interpolatedService := taskenv.InterpolateService(env, service)
+			interpolatedServices = append(interpolatedServices, interpolatedService)
 		}
 
 		// scan for missing or unhealthy consul checks
--- a/client/allochealth/tracker.go
+++ b/client/allochealth/tracker.go
@@ -12,6 +12,7 @@ import (
 	"github.com/hashicorp/nomad/client/serviceregistration"
 	"github.com/hashicorp/nomad/client/serviceregistration/checks/checkstore"
 	cstructs "github.com/hashicorp/nomad/client/structs"
+	"github.com/hashicorp/nomad/client/taskenv"
 	"github.com/hashicorp/nomad/helper"
 	"github.com/hashicorp/nomad/nomad/structs"
 	"golang.org/x/exp/maps"
@@ -96,6 +97,10 @@ type Tracker struct {
 	// name -> state
 	taskHealth map[string]*taskHealthState
 
+	// taskEnvs maps each task in the allocation to a *taskenv.TaskEnv that is
+	// used to interpolate runtime variables used in service definitions.
+	taskEnvs map[string]*taskenv.TaskEnv
+
 	// logger is for logging things
 	logger hclog.Logger
 }
@@ -111,6 +116,7 @@ func NewTracker(
 	logger hclog.Logger,
 	alloc *structs.Allocation,
 	allocUpdates *cstructs.AllocListener,
+	taskEnvBuilder *taskenv.Builder,
 	consulClient serviceregistration.Handler,
 	checkStore checkstore.Shim,
 	minHealthyTime time.Duration,
@@ -132,6 +138,12 @@ func NewTracker(
 		lifecycleTasks:      map[string]string{},
 	}
 
+	// Build the map of TaskEnv for each task. Create the group-level TaskEnv
+	// first because taskEnvBuilder is mutated in every loop and we can't undo
+	// a call to UpdateTask().
+	t.taskEnvs = make(map[string]*taskenv.TaskEnv, len(t.tg.Tasks)+1)
+	t.taskEnvs[""] = taskEnvBuilder.Build()
+
 	t.taskHealth = make(map[string]*taskHealthState, len(t.tg.Tasks))
 	for _, task := range t.tg.Tasks {
 		t.taskHealth[task.Name] = &taskHealthState{task: task}
@@ -139,6 +151,8 @@ func NewTracker(
 		if task.Lifecycle != nil && !task.Lifecycle.Sidecar {
 			t.lifecycleTasks[task.Name] = task.Lifecycle.Hook
 		}
+
+		t.taskEnvs[task.Name] = taskEnvBuilder.UpdateTask(alloc, task).Build()
 
 		c, n := countChecks(task.Services)
 		t.consulCheckCount += c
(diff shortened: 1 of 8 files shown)


nomad/structs/structs.go
```diff
@@ -6460,12 +6460,20 @@ func (tg *TaskGroup) Canonicalize(job *Job) {
...
 	return tg.filterServices(func(s *Service) bool {
 		return s.Provider == ServiceProviderNomad
 	})
 }
 
+// ConsulServices returns a list of all group and task - level services in tg
+// that are making use of the consul service provider.
+//
+// The returned services must be canonicalized and go though a taskenv.TaskEnv
+// to have its values interpolated.
```
lgfa29, line 6476: I kind of want to force the caller to pass in the job name, namespace, and a `TaskEnv` so we can return services that are already interpolated, but maybe that's too much?

client/allochealth/tracker.go
```diff
@@ -523,14 +529,20 @@ OUTER:
 	}
 }
 
-func evaluateConsulChecks(tg *structs.TaskGroup, registrations *serviceregistration.AllocRegistration) bool {
+func evaluateConsulChecks(
+	alloc *structs.Allocation,
+	tg *structs.TaskGroup,
+	taskEnv *taskenv.TaskEnv,
+	registrations *serviceregistration.AllocRegistration,
+) bool {
```
lgfa29, line 537: We're now passing a bunch of stuff here. Another option would be to perform the interpolation outside `evaluateConsulChecks` and just pass the list of services, or make this a method of `Tracker`.

client/allocrunner/alloc_runner_hooks.go
```diff
@@ -149,7 +149,7 @@ func (ar *allocRunner) initRunnerHooks(config *clientconfig.Config) error {
 		newCgroupHook(ar.Alloc(), ar.cpusetManager),
 		newUpstreamAllocsHook(hookLogger, ar.prevAllocWatcher),
 		newDiskMigrationHook(hookLogger, ar.prevAllocMigrator, ar.allocDir),
-		newAllocHealthWatcherHook(hookLogger, alloc, hs, ar.Listener(), ar.consulClient, ar.checkStore),
+		newAllocHealthWatcherHook(hookLogger, alloc, builtTaskEnv, hs, ar.Listener(), ar.consulClient, ar.checkStore),
```
lgfa29, line 152: I think re-using it here is safe since these are read-only? Another option is to call `envBuilder.Build()` in each function that needs a `TaskEnv`.

nomad/structs/structs.go
lgfa29: After stewing on this a bit more I think I'm going to add the arguments after all, but allow passing empty values. It's like checking and signing a "I have read and agree to the Terms and Conditions" that you know the values returned are not interpolated 😅

lgfa29: Small refactoring of service interpolation in https://github.com/hashicorp/nomad/pull/16413. I will rebase this PR when that one gets merged.

client/allochealth/tracker.go
lgfa29: Discussion covered in https://github.com/hashicorp/nomad/pull/16402#discussion_r1130228868

lgfa29:
After all my failed attempts to refactor the service interpolation so it's closer to the `Service` struct I decided to take a step back and understand this part of the code better.

Then I realized that my initial implementation was quite wrong. As the name suggests, each `TaskEnv` holds information for a specific task (or nothing task-related for a group). Since `allocHealthWatcherHook` is an alloc runner hook we actually need one `TaskEnv` _per task_ (+1 for the group).

`allocHealthWatcherHook` also implements the `Update()` hook, so the allocation may change while the hook is active, in which case we need to update the allocation in the `taskenv.Builder` and recreate the `TaskEnv`.

And since `taskenv.Builder` mutates itself in several different ways, we can't share instances across hooks. `taskenv.Builder` also mutate themselves in ways that cannot be undone (for example, calling `UpdateAlloc(alloc, nil)` will not reset the task to empty, it will just [ignore the change](https://github.com/hashicorp/nomad/blob/v1.5.0/client/taskenv/env.go#L701-L703)).

So `allocHealthWatcherHook` now takes a `taskenv.Builder` factory to creates a fresh builder when the allocation updates to prevent spurious task information at the group level `TaskEnv`.

client/allocrunner/alloc_runner_hooks.go
lgfa29: _Narrator: it was not safe._

nomad/structs/structs.go
lgfa29: Not refactoring here. There are quite a few intricacies that would require a more thoughtful process.

client/allochealth/tracker.go
```diff
@@ -503,8 +517,24 @@ OUTER:
...
+		interpolatedServices := make([]*structs.Service, len(consulServices))
+		for i, service := range consulServices {
+			env := t.taskEnvs[service.TaskName]
+			if env == nil {
+				// This is not expected to happen, but guard against a nil
+				// task environment by using the group environment since it has
+				// most of the same values.
+				t.logger.Warn("task environment not found, using group level environment for variable interpolation",
+					"alloc_id", t.alloc.ID, "task", service.TaskName)
+				env = t.taskEnvs[""]
```
shoenig, line 531: we [should] already guard against this in Validate/Canonicalize; instead of trying to make things _maybe_ work in an unexpected situation I think we should just do nothing, e.g.

```
if env != nil {
 // assign
} else {
  logger.Error("you found a bug!")
}
```

shoenig (approved):
LGTM; just one suggestion you could ignore. my thinking is we _want_ folks to complain about bugs if we are validating / canonicalizing things incorrectly. if we try to do the right thing they may just ignore what is otherwise a serious bug

client/allochealth/tracker.go
lgfa29: Hum...good point. Changed to this in 925d34798

tgross (approved):
:+1:

PR #16422, opened 2023-03-10
scheduler: add simple benchmark for tasksUpdated

shoenig:
In preperation for some refactoring to `tasksUpdated`, add a benchmark to the
old code so it's easy to compare with the changes, making sure nothing goes
off the rails for performance.

inaugural run:

```
➜ go test -count=3 -bench=BenchmarkTasksUpdated -run BenchmarkTasksUpdated
BenchmarkTasksUpdated-24           91621             15052 ns/op
BenchmarkTasksUpdated-24           80002             15023 ns/op
BenchmarkTasksUpdated-24           78248             15514 ns/op
```

### DIFF
--- a/nomad/mock/job.go
+++ b/nomad/mock/job.go
@@ -568,3 +568,93 @@ func MaxParallelJob() *structs.Job {
 	job.Canonicalize()
 	return job
 }
+
+// BigBenchmarkJob creates a job with many fields set, ideal for benchmarking
+// stuff involving jobs.
+//
+// Should not be used outside of benchmarking - folks should feel free to add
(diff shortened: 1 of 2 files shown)


jrasell (approved):
LGTM!

one question: do we want to add the benchmark results as a comment in the code? That way, it's easy for somewhere to see the previous results and compare to the current run with changes.

shoenig:
You really have to always re-run the benchmark on the old code anyway, since the hardware is unlikely to be identical.

tgross (approved):
LGTM!

PR #16434, opened 2023-03-11
plugin: add missing fields to `TaskConfig`

lgfa29:
While testing https://github.com/hashicorp/nomad-driver-podman/pull/215 several of my tests failed with empty strings even though everything looked right. The problem is that we were missing these fields when encoding/decoding `TaskConfig` objects to send to external plugins.


jrasell (approved):
LGTM, nice find!

PR #16437, opened 2023-03-11
feat: show warning if policy doesn't exist

farbodahm:
This pr is related to #14058 .

I wanted to work on that issue, but as I saw still it's not quite obvious on how we are going to fix it, I decided to only show a warning so at-least user can be aware the policy he/she is referring to, doesn't exist at the moment.

Also I can continue fixing this one if the expected result is known now.
Thanks.

Fixes: #14058

### DIFF
--- a/.changelog/16437.txt
+++ b/.changelog/16437.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+cli: `acl token create` will now emit a warning if the token has a policy that does not yet exist
+```
--- a/command/acl_token_create.go
+++ b/command/acl_token_create.go
@@ -154,6 +154,14 @@ func (c *ACLTokenCreateCommand) Run(args []string) int {
 		return 1
 	}
 
+	// Show warning if policy doesn't exist
+	for _, policy := range tk.Policies {
+		_, _, err := client.ACLPolicies().Info(policy, nil)
+		if err != nil {
+			c.Ui.Warn(fmt.Sprintf("Error fetching info on %s policy: %s", policy, err))
+		}
+	}
+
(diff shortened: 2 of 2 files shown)


tgross:
Hey @farbodahm! Just a heads up that we're having a fairly intense (but friendly! :grinning: ) internal debate about how we actually want to handle this whole missing-policy issue, because it has a lot of backwards compatibility issues as discussed in #14058. We'll probably at least want to have this warning, so thanks for that. We'll try to get back to you on this soon.

farbodahm:
@tgross Hey man. Thank you for the update. 
I'll be happy if I can give any help regarding this issue; So please let me know if anything is needed.

angrycub:
Hey @farbodahm, I apologize for not responding to you sooner.  I was reviewing your request and I had what I thought at the time would be a simple organizational ask about consolidating the policies that weren't found into a single warning line.

In looking more deeply, I realized that, other than string-scraping, there is currently not a way to know if an unexpected response error was because of a not found or because of a server error. To address this issue, I started working on #16743, which can then be used by code to more concretely validate the error reasons and provide more specific warning text in these cases.

I am going to assign this PR to myself and will follow up with you once 16743's fate is decided.

tgross (approved):
Looking to do a little bit of PR and issue cleanup and found this issue. While it'd be great to rework this and other bits of the CLI to take advantage of the shapely error work done in #16743, better to just go ahead and land this if we don't have the time to follow-up. I've rebased this PR on main and pushed a commit with a changelog entry.

Will merge once green and then this will ship in Nomad 1.8.3

PR #16442, opened 2023-03-13
cli: Add  `-json` and `-t` flags to `namespace status` command

Juanadelacuesta:


command/namespace_status.go
```diff
@@ -24,13 +24,26 @@ Usage: nomad namespace status [options] <namespace>
...
+  ` + generalOptionsUsage(usageOptsDefault|usageOptsNoNamespace) + `
+
+  Status Specific Options:
+  
+	-json
+	  Output the latest namespace status information in a JSON format.
+  
+	-t
+	  Format and display namespace status information using a Go template.
+  `
```
jrasell, line 36: slight fix to the output formatting:

```suggestion
Status Specific Options:
  
  -json
    Output the latest namespace status information in a JSON format.
  
  -t
    Format and display namespace status information using a Go template.
`
```

command/namespace_status_test.go
```diff
@@ -52,6 +54,7 @@ func TestNamespaceStatusCommand_Good(t *testing.T) {
 	// Create a namespace
 	ns := &api.Namespace{
 		Name: "foo",
+		//Description: "Test namespace",
```
jrasell, line 57: If this isn't being used, can we remove it?

jrasell (commented):
Looking good, thanks!

The PR will need a changelog entry added. Also if we could move the additional tests added to use must, that would be great. An optional change would be to move all the new and current tests in the file to use must, but this isn't required on this PR. :D

command/namespace_status_test.go
Juanadelacuesta: It can, my bad :)

jrasell (approved):
LGTM, thanks @Juanadelacuesta! Lets add the backport 1.5.x label before merging this in.

PR #16463, opened 2023-03-13
acl: update job eval requirement to `submit-job`

lgfa29:
The job evaluate endpoint creates a new evaluation for the job which is a write operation. This change modifies the necessary capability from `read-job` to `submit-job` to better reflect this.

No backport since this is targeted for Nomad 1.6.0


shoenig (approved):
LGTM!

Issue #16291, opened 2023-03-02
Limit "resources.cores" does not work after nomad client restart


TheSpbra1n:
Hello.
We test `resources.cores` for nomad jobs and found a bug: after nomad client restart, cores limit does not work. 

### Nomad version
Nomad v1.4.2 (039d70eeef5888164cad05e7ddfd8b6f8220923b)

### Operating system and Environment details
Codename:	focal

### Reproduction steps
1. Run job:
```
# job name
job "cpu-loader" {
  region      = "global"
  datacenters = ["dc1"]

  type = "service"

  meta {
      ...
  }

  group "cpu-loader-group" {
    count = 1

    # task
    task "cpu-loader-task" {
      kill_signal = "SIGTERM"
      driver = "docker"

      config {
        image        = "docker-image"
      }

      env {
          ...
      }

      resources {
       memory = 1000
       cores  = 4 
      }
    }
  }
}
```
2. Stop job
3. Restart nomad client
4. Run job again

(for loader we used: https://github.com/vikyd/go-cpu-load )

#### Expected Result
Job used only 4 cores

#### Actual Result
Job used more than 4 core

RobloxMatt:
We've seen something that sounds like this problem as well.  Do you see something like this in your logs?

`client.cpuset.v1: failed to ensure reserved cpuset.cpus interface exists; disable cpuset management: error="mkdir /sys/fs/cgroup/cpuset/nomad/reserved: file exists"`

or

`[WARN]  client.cpuset.v1: failed to ensure reserved cpuset.cpus interface exists; disable cpuset management: error="mkdir /sys/fs/cgroup/cpuset/nomad/reserved: file exists"`

shoenig:
Thanks for the report @TheSpbra1n and for the logs @RobloxMatt, it seems this got broken during a refactoring in https://github.com/hashicorp/nomad/commit/0a3d57f7e4603e264daed73a569ac4369e25163b where we (I) missed the edge case where the cgroup structure is already set on Client initialization. Should only affect machines using cgroups v1.

PR #16467, opened 2023-03-13
cgv1: do not disable cpuset manager if reserved interface already exists

shoenig:
This PR fixes a bug where restarting a Nomad Client on a machine using cgroups
v1 (e.g. Ubuntu 20.04 or earlier) would cause the cpuset cgroups manager to disable itself.

This is being caused by incorrectly interpreting a "file exists" error as
problematic when ensuring the reserved cpuset exists. If we get a "file exists"
error, that just means there is nothing to do, and the cpuset manager should continue
initializing normally.

Note that a machine reboot would fix the issue - the interfaces are ephemeral.

Fixes https://github.com/hashicorp/nomad/issues/16291

### DIFF
--- a/.changelog/16467.txt
+++ b/.changelog/16467.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+client: Fixed a bug where cpuset initialization fails after Client restart
+```
--- a/client/lib/cgutil/cpuset_manager_v1.go
+++ b/client/lib/cgutil/cpuset_manager_v1.go
@@ -4,6 +4,7 @@ package cgutil
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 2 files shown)


shoenig:
spot check

```
ubuntu@ip-172-31-19-2:~$ grep PRETTY /etc/os-release
PRETTY_NAME="Ubuntu 20.04.5 LTS"
```

After Nomad Client restart,

```
2023-03-13T19:56:22.242Z [INFO]  client.cpuset.v1: initialized cpuset cgroup manager: parent=/nomad cpuset=0-1
2023-03-13T19:56:22.254Z [INFO]  client.fingerprint_mgr.cgroup: cgroups are available
```

tgross (approved):
LGTM!

lgfa29 (approved):

Issue #11898, opened 2022-01-21
alloc/exec: WebSocket 1011 acl token lookup failed: index error: UUID must be 36 characters


Thunderbottom:
### Nomad version
```bash
$ nomad version
Nomad v1.2.3 (a79efc8422082c4790046c3f5ad92c542592a54f)
```

### Operating system and Environment details
```bash
$ cat /etc/*release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.3 LTS"
NAME="Ubuntu"
VERSION="20.04.3 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.3 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

### Issue

If the ACL Token set in localstorage contains spaces, any RPC calls to `Allocations.Exec` results in:

```
  Connection Closed: 1011 acl token lookup failed: index error: UUID must be 36 characters
```

##### Allocation Exec Websocket:
![alloc exec ws](https://user-images.githubusercontent.com/11243138/150521518-955f1b52-dc5c-4197-b6e9-af4d5dc01661.png)

##### Request Headers:
![ws headers](https://user-images.githubusercontent.com/11243138/150521528-51529b9b-ceaa-4d10-b9c4-3ce11404f9d8.png)

### Reproduction steps

1. Set token in `https://<nomad-instance>/ui/settings/tokens`, preferably with an added space before the token. The token gets set correctly, which means that the system is handling spaces just fine. Example Token: `" XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX"`

2. Exec into any allocation through the web UI, which would fail as described above.

#### Expected Result

The system should consistently handle spaces in the ACL tokens correctly. Since it accepts tokens with prefixed/suffixed spaces while authenticating, `Allocations.Exec` should not fail either and users should be allowed to exec into allocations.

#### Actual Result

[...]

PR #11897, opened 2022-01-21
fix(alloc/exec): trim spaces on secretID to prevent error on lookup

Thunderbottom:
If the ACL Token set in localstorage contains prefixed/suffixed spaces, any RPC calls to `Allocations.Exec` results in:

```
  Connection Closed: 1011 acl token lookup failed: index error: UUID must be 36 characters
```

### How to reproduce:

1. Set token in `https://<nomad-instance>/ui/settings/tokens`, preferably with an added space before the token. The token gets set correctly, which means that the system is handling spaces just fine.

2. Exec into any allocation through the web UI, which would fail as described above.

### Sample Request Payload:

```json
{
  "version": 1,
  "auth_token": " XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX",
}
```

Notice the prefixed added space in the `auth_token` value which causes the error.

### Proposed fix:

RPC to `Allocations.Exec` calls `ACLTokenBySecretID()`, which should pass the space-trimmed `secretID` to lookup the ACL token. A simple `strings.TrimSpace()` call around `secretID` should fix the issue.

PR #16469, opened 2023-03-13
agent: trim space when parsing X-Nomad-Token header

tgross:
Fixes https://github.com/hashicorp/nomad/issues/11898 and will replace https://github.com/hashicorp/nomad/pull/11897

Our auth token parsing code trims space around the `Authorization` header but not around `X-Nomad-Token`. When using the UI, it's easy to accidentally introduce a leading or trailing space, which results in spurious authentication errors. Some browsers fix this automatically, but apparently not all do. Trim the space at the HTTP server.


jrasell (approved):
LGTM!

Issue #16470, opened 2023-03-13
clients require node:read on anon ACL policy to join cluster via Consul discovery (1.5.1)


CarelvanHeerden:
### Nomad version
v1.5.1

### Operating system and Environment details
Ubuntu 20.04

### Issue
Our Test system was running v1.4.3, with ACL system configured. Upgraded to V1.5.1 using a rolling upgrade (replacing all servers and clients)
The servers started up correctly and joined the cluster.
The clients however could not join, getting a RPC error: Permission denied

The only way to get the clients to Join, was to add back the Anonymous policy

### Client Config
etc/nomad.d/acl.hcl 
```
acl {
  enabled = true
}
```

/etc/nomad.d/client.hcl 
```
client {
 enabled = true
 disable_remote_exec = true
 max_kill_timeout = "90s"
  options = {
  docker.privileged.enabled = true
  docker.volumes.enabled = true
  }
  reserved {
  cpu  = 500
  memory         = 512
  }
}
```
/etc/nomad.d/nomad.hcl 
```
datacenter = "DMZ"
data_dir = "/opt/nomad"
region = "sa"
telemetry {
  collection_interval = "1s"
  disable_hostname = true
  prometheus_metrics = true
  publish_allocation_metrics = true
  publish_node_metrics = true
}
```
### Server Config
etc/nomad.d/acl.hcl 
```
acl {
  enabled = true
}
```
/etc/nomad.d/server.hcl 
```
server {
  enabled = true
  bootstrap_expect = 3
  default_scheduler_config {
    scheduler_algorithm = "spread"
    memory_oversubscription_enabled = true
    preemption_config {
      batch_scheduler_enabled    = true
      enable_event_broker    = true
      system_scheduler_enabled   = true
      service_scheduler_enabled  = true
      sysbatch_scheduler_enabled = true
    }
  }
}
```

/etc/nomad.d/nomad.hcl 
```
datacenter = "GENERAL"
data_dir = "/opt/nomad"
region = "sa"
telemetry {
  collection_interval = "1s"
  disable_hostname = true
  prometheus_metrics = true
  publish_allocation_metrics = true
  publish_node_metrics = true
}
```

### CLI output
nomad acl token list
```
Name             Type        Global  Accessor ID                           Expired
Bootstrap Token  management  true    4a03f056-XXXXX-XXXXX-XXXXX  false
Terraform        management  true    6e26d344-XXXXX-XXXXX-XXXXX  false
```

### Reproduction steps
[...]

PR #16490, opened 2023-03-14
client: use `Status.RPCServers` RPC for Consul discovery

tgross:
Fixes https://github.com/hashicorp/nomad/issues/16470

In #16217 we switched clients using Consul discovery to the `Status.Members`
endpoint for getting the list of servers so that we're using the correct
address. This endpoint has an authorization gate, so this fails if the anonymous
policy doesn't have `node:read`. We also can't check the `AuthToken` for the
request for the client secret, because the client hasn't yet registered so the
server doesn't have anything to compare against.

Instead of hitting the `Status.Peers` or `Status.Members` RPC endpoint, use the
Consul response directly. Update the `registerNode` method to handle the list of
servers we get back in the response; if we get a "no servers" or "no path to
region" response we'll kick off discovery again and retry immediately rather
than waiting 15s.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -2950,12 +2950,12 @@ DISCOLOOP:
 
 			// Query the client-advertise RPC addresses from the region that
 			// Consul gave us
-			var members []string
-			if err := c.connPool.RPC(region, addr, "Status.RPCServers", rpcargs, &members); err != nil {
+			var resp *structs.RPCServersResponse
+			if err := c.connPool.RPC(region, addr, "Status.RPCServers", rpcargs, &resp); err != nil {
 				mErr.Errors = append(mErr.Errors, err)
 				continue
 			}
-			for _, member := range members {
+			for _, member := range resp.Addresses {
 				addr, err := net.ResolveTCPAddr("tcp", member)
 				if err != nil {
 					mErr.Errors = append(mErr.Errors, err)
--- a/client/client.go
+++ b/client/client.go
@@ -58,7 +58,6 @@ import (
 	vaultapi "github.com/hashicorp/vault/api"
 	"github.com/shirou/gopsutil/v3/host"
 	"golang.org/x/exp/maps"
-	"golang.org/x/exp/slices"
 )
 
 const (
@@ -1944,7 +1943,7 @@ func (c *Client) retryRegisterNode() {
 		}
 
 		retryIntv := registerRetryIntv
-		if err == noServersErr {
+		if err == noServersErr || structs.IsErrNoRegionPath(err) {
 			c.logger.Debug("registration waiting on servers")
 			c.triggerDiscovery()
 			retryIntv = noServerRetryIntv
@@ -1971,6 +1970,11 @@ func (c *Client) registerNode() error {
 		return err
 	}
 
+	err := c.handleNodeUpdateResponse(resp)
+	if err != nil {
+		return err
+	}
+
 	// Update the node status to ready after we register.
 	c.UpdateConfig(func(c *config.Config) {
 		c.Node.Status = structs.NodeStatusReady
@@ -1985,6 +1989,7 @@ func (c *Client) registerNode() error {
 	defer c.heartbeatLock.Unlock()
 	c.heartbeatStop.setLastOk(time.Now())
 	c.heartbeatTTL = resp.HeartbeatTTL
+
 	return nil
 }
 
@@ -2036,6 +2041,25 @@ func (c *Client) updateNodeStatus() error {
 		}
 	})
 
+	err := c.handleNodeUpdateResponse(resp)
+	if err != nil {
+		return fmt.Errorf("heartbeat response returned no valid servers")
+	}
+
+	// Begin polling Consul if there is no Nomad leader.  We could be
+	// heartbeating to a Nomad server that is in the minority of a
+	// partition of the Nomad server quorum, but this Nomad Agent still
+	// has connectivity to the existing majority of Nomad Servers, but
+	// only if it queries Consul.
+	if resp.LeaderRPCAddr == "" {
+		c.triggerDiscovery()
+	}
+
+	c.EnterpriseClient.SetFeatures(resp.Features)
+	return nil
+}
+
+func (c *Client) handleNodeUpdateResponse(resp structs.NodeUpdateResponse) error {
 	// Update the number of nodes in the cluster so we can adjust our server
 	// rebalance rate.
 	c.servers.SetNumNodes(resp.NumNodes)
@@ -2052,20 +2076,9 @@ func (c *Client) updateNodeStatus() error {
 		nomadServers = append(nomadServers, e)
 	}
 	if len(nomadServers) == 0 {
-		return fmt.Errorf("heartbeat response returned no valid servers")
+		return noServersErr
 	}
 	c.servers.SetServers(nomadServers)
-
-	// Begin polling Consul if there is no Nomad leader.  We could be
-	// heartbeating to a Nomad server that is in the minority of a
-	// partition of the Nomad server quorum, but this Nomad Agent still
-	// has connectivity to the existing majority of Nomad Servers, but
-	// only if it queries Consul.
-	if resp.LeaderRPCAddr == "" {
-		c.triggerDiscovery()
-	}
-
-	c.EnterpriseClient.SetFeatures(resp.Features)
 	return nil
 }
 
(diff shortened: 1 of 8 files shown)


tgross:
Moved back to draft, as this is still failing in E2E testing.

tgross:
I've reworked this PR heavily based on the results of a discussion I had with @shoenig and @gulducat in order to get this working in the pre-registration phase. Because the new RPC handler is substantially the same as https://github.com/hashicorp/nomad/pull/11895/commits/ee3e9522ffa5aa6f203ffe4fd0c8185a7a9e1fbf which we closed unmerged, I've added Ruslan as a co-author on this commit.

I need to do one more round of testing with this on E2E with the advertise addresses set correctly, and then I'll be ready for review.

tgross:
Ok, I've tested this out on a cluster on AWS using HCP Consul for discovery and the networking tweaks discussed in https://github.com/hashicorp/nomad/issues/16507 to ensure that we're using the serf advertisement:

```
$ nomad operator api "/v1/status/peers"
["172.31.119.47:4647","172.31.112.244:4647","172.31.125.250:4647"]%

$ nomad operator api "/v1/agent/members" | jq -r '.Members[].Addr'
172.31.125.250
172.31.112.244
172.31.119.47

$ nomad operator api "/v1/agent/members" | jq -r '.Members[].Tags.rpc_addr'
172.31.92.121
172.31.93.95
172.31.88.76
```

And both the servers and client end up happy:

```
$ nomad server members
Name                     Address         Port  Status  Leader  Raft Version  Build      Datacenter  Region
ip-172-31-88-76.global   172.31.119.47   4648  alive   false   3             1.5.2-dev  dc1         global
ip-172-31-92-121.global  172.31.125.250  4648  alive   true    3             1.5.2-dev  dc1         global
ip-172-31-93-95.global   172.31.112.244  4648  alive   false   3             1.5.2-dev  dc1         global

$ nomad node status
ID        DC   Name              Class   Drain  Eligibility  Status
700d45a8  dc1  ip-172-31-88-231  <none>  false  eligible     ready
```

nomad/status_endpoint.go
```diff
@@ -78,6 +78,31 @@ func (s *Status) Peers(args *structs.GenericRequest, reply *[]string) error {
 	return nil
 }
 
+// RPCServers is used to get all the RPC server addresses in a region
+func (s *Status) RPCServers(args *structs.GenericRequest, reply *[]string) error {
```
tgross, line 82: Note to reviewers, I've intentionally left off adding a HTTP API for this endpoint. We already have a `/v1/agent/servers` endpoint but it's an agent endpoint and currently seems to work in a weird and possibly buggy way (`nomad operator api /v1/agent/servers` always returns "wrong method" for me!). It's not necessary to do this to fix the regression, so I'd like to do any polish we might want in a follow-up PR.

shoenig, line 82: Would it make sense to future-proof this a bit by wrapping the `[]string` with a typical RPC response struct? (onto which we could add more fields later). It'd also help not have pointer-to-slice which always feels weird.

tgross: Yeah, totally. I'd copied it right out of `Status.Peers` but signatures like those are terrible. :grin: Will fix.

client/client.go
```diff
@@ -2944,26 +2945,24 @@ DISCOLOOP:
...
-			// extract the client-advertise RPC address from each member
-			var membersResp structs.ServerMembersResponse
-			if err := c.connPool.RPC(region, addr, "Status.Members", rpcargs, &membersResp); err != nil {
+			srv := &servers.Server{Addr: addr}
+			nomadServers = append(nomadServers, srv)
+
+			// Query the client-advertise RPC addresses from the region that
+			// Consul gave us
+			var members []string
+			if err := c.connPool.RPC(region, addr, "Status.RPCServers", rpcargs, &members); err != nil {
```
schmichael: If Nomad Servers are properly advertising their client-to-server IP address and port, then why do we have to call _any_ Nomad RPC to resolve server addresses? Aren't the addresses in Consul fine? If the addresses in Consul aren't fine, can we have Nomad Servers register a new Consul Service for Nomad Clients to discover without having to use an ancillary RPC?

tgross: That's what I wondered too! But I dug _way_ back in the code and it looks like the reason it was done this way was that multiple regions could be registered in the same Consul cluster (?!). So we hit the RPC endpoint to say "ok but which of these are _our_ servers?" Although to be honest, this seems like a very strange topology that probably breaks a million other things.

tgross: That being said, maybe we could tag the registration in Consul with the region name and then just query on that. That seems like it'd be nicer. Let me take a look at how much of a lift that'd be.

tgross: Done in 094d3ba49675a36f57405ba7f497d3bf4cb0ea27 @schmichael. This seems like it cuts thru the extra nonsense here nicely. (And lets clients become live a smidgen faster!)

command/agent/agent.go
```diff
@@ -891,7 +891,8 @@ func (a *Agent) setupServer() error {
 		rpcServ := &structs.Service{
 			Name:      a.config.Consul.ServerServiceName,
 			PortLabel: a.config.AdvertiseAddrs.RPC,
-			Tags:      append([]string{consul.ServiceTagRPC}, a.config.Consul.Tags...),
+			Tags: append([]string{consul.ServiceTagRPC, a.config.Region},
+				a.config.Consul.Tags...),
```
tgross, line 895: @schmichael what do you think of adding the region tag to the other 2 services too? I can't think of any reason not to, and it'd be nice for consistency when people query.

client/client.go
```diff
@@ -2933,43 +2929,27 @@ DISCOLOOP:
...
-						addr, err := net.ResolveTCPAddr("tcp",
-							fmt.Sprintf("%s:%s", addrTag, portTag))
-						if err != nil {
-							mErr.Errors = append(mErr.Errors, err)
-							continue
-						}
-						srv := &servers.Server{Addr: addr}
-						nomadServers = append(nomadServers, srv)
-					}
+			if slices.Contains(s.ServiceTags, region) {
```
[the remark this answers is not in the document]
tgross: I've updated the PR (and the description above) to drop the extra tags/meta on Consul. And then I've updated the `registerNode` method to use the servers we get back from the response. If we get no servers or a "no path to region", we'll bubble that state up to the retry loop so that it triggers discovery again and hits the fast-path retry rather than the longer retry of 15s.

client/client.go
```diff
@@ -2035,6 +2041,25 @@ func (c *Client) updateNodeStatus() error {
...
+	err := c.handleNodeUpdateResponse(resp)
+	if err != nil {
+		return fmt.Errorf("heartbeat response returned no valid servers")
+	}
+
+	// Begin polling Consul if there is no Nomad leader.  We could be
+	// heartbeating to a Nomad server that is in the minority of a
+	// partition of the Nomad server quorum, but this Nomad Agent still
+	// has connectivity to the existing majority of Nomad Servers, but
+	// only if it queries Consul.
```
schmichael, line 2053: Let's clean this up a bit while we're moving it:

```suggestion
	// If there's no Leader in the response we may be talking to a partitioned
	// server. Redo discovery to ensure our server list is up to date.
```

Perhaps? It doesn't mention Consul, but in a perfect world it would trigger go-discover as well.

nomad/acl.go
```diff
@@ -161,9 +161,14 @@ func (s *Server) remoteIPFromRPCContext(ctx *RPCContext) (net.IP, error) {
 // for the identity they intend the operation to be performed with.
 func (s *Server) ResolveACL(args structs.RequestWithIdentity) (*acl.ACL, error) {
 	identity := args.GetIdentity()
-	if !s.config.ACLEnabled || identity == nil {
+	if !s.config.ACLEnabled {
 		return nil, nil
```
schmichael, line 165: OT: I don't think we should do this for this PR, but I am curious if we could get away from using `nil` as a sentinel value by switching these helpers to return a management token:

```suggestion
	if !s.config.ACLEnabled {
		return structs.ManagementToken(), nil
```

Maybe I'll find an opportunity to poke at that more in my WID work... :thinking:

schmichael (approved):

nomad/acl.go
tgross: Good call. As per our sidebar discussion, I've backed out the changes to this file in this PR

PR #16494, opened 2023-03-14
Added section of needed rules for Nomad UI

ron-savoia:
Added a UI section noting the needed read rules (node and agent) when using the Nomad UI

### DIFF
--- a/website/content/docs/other-specifications/acl-policy.mdx
+++ b/website/content/docs/other-specifications/acl-policy.mdx
@@ -384,7 +384,9 @@ The `policy` field for the plugin rule can have one of the following values:
 
 ## Configuring ACLs for the web UI
 
-Since the Nomad UI provides a much richer user experience than the CLI, additional API endpoints `/v1/agent` `/v1/node` are used when rendering pages.  Adding the rules below to your ACL when utilizing the Nomad UI is recommended.
+The Nomad web UI uses the API endpoints `/v1/agent` and `/v1/node` for nearly
+every page. ACL policies for users who will use the Nomad UI should include the
+following rules.
 
 ```hcl
 node {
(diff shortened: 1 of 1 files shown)


website/content/docs/other-specifications/acl-policy.mdx
```diff
@@ -382,6 +382,20 @@ The `policy` field for the plugin rule can have one of the following values:
 - `deny`: do not allow the resource to be read or modified. Deny takes
   precedence when multiple policies are associated with a token.
 
+## UI rules
```
tgross, line 385: Taken with the rest of this doc, this header implies there's a `ui` section. Let's call this:

```suggestion
## Configuring ACLs for the web UI
```

website/content/docs/other-specifications/acl-policy.mdx
```diff
@@ -382,6 +382,20 @@ The `policy` field for the plugin rule can have one of the following values:
...
+
+Since the Nomad UI provides a much richer user experience than the CLI, additional API endpoints `/v1/agent` `/v1/node` are used when rendering pages.  Adding the rules below to your ACL when utilizing the Nomad UI is recommended.
```
tgross, line 387: Generally we want to keep marketing language out of the docs and have it be more "just the facts". Also, we hard-wrap markdown:

```suggestion
The Nomad web UI uses the API endpoints `/v1/agent` and `/v1/node` for nearly
every page. ACL policies for users who will use the Nomad UI should include the
following rules.
```

tgross (changes requested):
Thanks @ron-savoia! I'm going to tag-in a couple other folks for review here as well to see if there's a better policy to use or way to word this.

PR #16496, opened 2023-03-14
update `default` Mirage scenario

ChaiWithJai:
This PR updates the `default` Mirage scenario to include a `Job` that has the new `Specification` field required to display HCL in the UI. Note:  Older jobs will not have this field.


philrenaud (approved):
Hell yeah dude, this is perfect.

PR #16497, opened 2023-03-14
ui:  conditionally render toggle in `jobs.job.definition`

ChaiWithJai:
Older jobs do not have a `Specification` field. Therefore, we'll need to conditionally render the toggle button.

image

### DIFF
--- a/ui/app/components/job-editor.js
+++ b/ui/app/components/job-editor.js
@@ -12,12 +12,13 @@ export default class JobEditor extends Component {
 
   @tracked error = null;
   @tracked planOutput = null;
-  @tracked view = 'job-spec';
   @tracked isEditing;
+  @tracked view;
 
   constructor() {
     super(...arguments);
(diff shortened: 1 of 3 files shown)


ui/app/components/job-editor.js
```diff
@@ -12,12 +12,13 @@ export default class JobEditor extends Component {
...
   @tracked error = null;
   @tracked planOutput = null;
-  @tracked view = 'job-spec';
   @tracked isEditing;
+  @tracked view;
 
   constructor() {
     super(...arguments);
     this.isEditing = !!(this.args.context === 'new');
+    this.view = this.args.specification ? 'job-spec' : 'full-definition';
```
philrenaud, line 21: Question: is this also accomplishable by having `view` be a getter with this logic embedded? Is there a particular reason to do it in a constructor instead?

philrenaud (approved):
LGTM! Nice work.

PR #16499, opened 2023-03-14
test: set BuildDate in default TestAgent config

gulducat:
Without this, various enterprise tests will fail due to the default zero time being guarded against in an upcoming enterprise change.  Here it should do nothing.

relates to: hashicorp/nomad-enterprise#855


jrasell (approved):
LGTM!

Note: if we want this backported, we can add the able post-merge.

PR #16504, opened 2023-03-15
cli: nomad login command should not require a -type flag and should respect default auth method

### LINKED ISSUE #16501: `nomad login` errors before determining the default auth method type
### Nomad version
`v1.5.0`

Ubuntu 22.04.2 LTS

### Issue
The [documentation for the `nomad login` command](https://developer.hashicorp.com/nomad/docs/commands/login#type) and the integration [guide here](https://developer.hashicorp.com/nomad/tutorials/single-sign-on/sso-oidc-vault#authenticate-with-nomad) state that the `type` flag is optional if an admin has configured a default, however this is not the case.

Line [`117` in `login.go`](https://github.com/hashicorp/nomad/blob/1a01e871927284b2c484d7745e1d7df1df6312d8/command/login.go#L117) explicitly rejects any calls to `nomad login` that don't have `OIDC` set as the type:
```go
switch l.authMethodType {
case api.ACLAuthMethodTypeOIDC:
default:
	l.Ui.Error(fmt.Sprintf("Unsupported authentication type %q", l.authMethodType))
	return 1
}
```

There is a check further down for a default auth method, at [line 143](https://github.com/hashicorp/nomad/blob/1a01e871927284b2c484d7745e1d7df1df6312d8/command/login.go#L143) that will set the type to be the type of the default auth method, but that section of the code is never reached because the switch statement above returns an error before it can try to determine the default type.

### Reproduction steps

 1. Register an auth method with Nomad (i.e. following [this guide](https://developer.hashicorp.com/nomad/tutorials/single-sign-on/sso-oidc-vault#authenticate-with-nomad))
 2. Make sure the auth method is set as default (`nomad acl auth-method info vault`)
 3. Try to run `nomad login` without specifying the type, you'll receive an error `Unsupported authentication type ""`

#### Expected Result

It should be selecting the default auth method and using the type from that. These is code defined in the command to do so, but the `switch` statement further up cancels it out before it can try to determine the default.

#### Actual Result

```
Unsupported authentication type ""
```

pkazmierczak:
`nomad login` command does not need to know ACL Auth Method's type, since all method names are unique. I discovered this while fixing a bug in which `login` ignored the existence of default auth method and errored when users didn't provide a method name or type. I believe this bug comes from the fact that originally we intended to have default-per-type design, and later we decided it's better UX to have global-default auth methods. 

Closes #16501.

### DIFF
--- a/command/login.go
+++ b/command/login.go
@@ -8,10 +8,11 @@ import (
 	"strings"
 
 	"github.com/hashicorp/cap/util"
-	"github.com/hashicorp/nomad/api"
-	"github.com/hashicorp/nomad/lib/auth/oidc"
 	"github.com/mitchellh/cli"
 	"github.com/posener/complete"
+
+	"github.com/hashicorp/nomad/api"
+	"github.com/hashicorp/nomad/lib/auth/oidc"
 )
 
 // Ensure LoginCommand satisfies the cli.Command interface.
@@ -21,7 +22,6 @@ var _ cli.Command = &LoginCommand{}
 type LoginCommand struct {
 	Meta
 
-	authMethodType string
 	authMethodName string
 	callbackAddr   string
 
@@ -47,10 +47,6 @@ Login Options:
     The name of the ACL auth method to login to. If the cluster administrator
     has configured a default, this flag is optional.
 
-  -type
-    Type of the auth method to login to. If the cluster administrator has
-    configured a default, this flag is optional.
-
   -oidc-callback-addr
     The address to use for the local OIDC callback server. This should be given
     in the form of <IP>:<PORT> and defaults to "localhost:4649".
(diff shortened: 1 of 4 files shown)


command/login.go
```diff
@@ -104,57 +98,52 @@ func (l *LoginCommand) Run(args []string) int {
...
+	for _, authMethod := range authMethodList {
+		if authMethod.Default {
+			defaultMethod = authMethod
+		}
+	}
 
-		authMethodList, _, err := client.ACLAuthMethods().List(nil)
-		if err != nil {
-			l.Ui.Error(fmt.Sprintf("Error listing ACL auth methods: %s", err))
+	// If there is a default method available, and the called did not pass method
```
jrasell, line 124:
```suggestion
	// If there is a default method available, and the caller did not pass method
```

jrasell (approved):
LGTM, with a tiny grammar comment change. It might be worth pinging @mikenomitch to double check he is cool with the UX change.

It would also be worth preparing PR's for any tutorials that also need updating based on this, so they can be released at the right time.

mikenomitch (approved):
+1 to the UX change from me

Issue #16477, opened 2023-03-14
Upgrading from v1.4.3 to v1.5.1: Nomad Job that uses a remote artifact, that requires SSH auth fails after upgrade


CarelvanHeerden:
### Nomad version
v1.5.1

### Operating system and Environment details
Ubuntu 20.04

### Issue
A Nomad job, that downloads artifacts from Gitlab using SSH auth, that was working fails after the upgrade to v1.5.1

### Reproduction steps
Nomad job, that has an artifact block that requires SSH auth fails after upgrade

#### Expected Result
Nomad Job to Continue as per normal

#### Actual Result
Job Fails
```
msg="failed to download artifact \"git::git@gitlab.com:xxxxxxxxxxxx.git\": getter subprocess failed: exit status 1" 
```

### Job file
Removed some tags that are not relevant
```
job "JOB" {
  type = "batch"
    periodic {
    cron = "5 10 * * 2"
    prohibit_overlap = true
  }

  group "JOB" {
    count = 1
    task "JOB" {
      driver = "docker"


      artifact {
          source      = "git::git@gitlab.com:XXXXXXXXXX.git"
          destination = "local/repo"
          options {
            ref = "main"
            depth = 1
            sshkey = "${ID_RSA_BASE64}"
          }
        }

      config {
        image        = "dockerrepo/XXXX/docker:${DOCKER_VERSION}"
        force_pull = "false"
        auth {
          username = "XXXXXXX"
          password = "${DOCKER_REPO_TOKEN}"
        }

        dns_servers = ["$${attr.unique.network.ip-address}"]

        volumes = [
          "local/repo:/tmp",
        ]

      }


      resources {
        cpu    = 1024
        memory = 2056

        network {
        }
      }
    }
  }
}


```

### Nomad Client logs 
```
Mar 14 11:43:38 nomad-client-secure00000V nomad[2878]:     2023-03-14T11:43:38.109Z [INFO]  client.alloc_runner.task_runner: Task event: alloc_id=901ad884-9552-2228-aa74-10cbe023a856 task=Ansible-CIS type="Downloading Artifacts" msg="Client is downloading artifacts" failed=false
[...]

PR #16508, opened 2023-03-15
artifact: git needs more files for private repositories

shoenig:
This PR fixes artifact downloading so that git may work when cloning from
private repositories. It needs

- file read on `/etc/passwd`
- dir read on `/root/.ssh`
- file write on `/root/.ssh/known_hosts`

Add these rules to the landlock rules for the artifact sandbox.

Fixes https://github.com/hashicorp/nomad/issues/16477

### DIFF
--- a/client/allocrunner/taskrunner/getter/util_linux.go
+++ b/client/allocrunner/taskrunner/getter/util_linux.go
@@ -8,6 +8,7 @@ import (
 	"syscall"
 
 	"github.com/shoenig/go-landlock"
+	"golang.org/x/sys/unix"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


client/allocrunner/taskrunner/getter/util_linux.go
```diff
@@ -42,12 +43,35 @@ func credentials() (uint32, uint32) {
...
+
+	// if home is set, just use that
+	if home, err := os.UserHomeDir(); err == nil {
+		return home
+	}
+
+	// if we are the root user, use "/root"
+	if unix.Getuid() == 0 {
+		return "/root"
+	}
```
tgross, line 61: We have `mitchell/go-homedir` in our dependencies already, which checks `getent` to get the homedir out of `/etc/passwd` if `$HOME` isn't set: https://github.com/mitchellh/go-homedir/blob/main/homedir.go#L114 Should we use that to handle weird edge cases like `$HOME` being unset for non-root users?

client/allocrunner/taskrunner/getter/util_linux.go
```diff
@@ -71,26 +95,60 @@ func lockdown(allocDir, taskDir string) error {
...
 		_, err := os.Stat(p)
 		return err == nil
 	}
-	result := make([]*landlock.Path, 0, 3)
+
+	result := make([]*landlock.Path, 0, 6)
+	if exists(sshDir) {
+		result = append(result, landlock.Dir(sshDir, "r"))
+	}
+	if exists(knownHosts) {
```
tgross, line 146: Does landlock require that the file already exists? I'm thinking of the case where the user has just spun up a VM from scratch and `/root/.ssh/known_hosts` hasn't been created yet.

client/allocrunner/taskrunner/getter/util_linux.go
```diff
@@ -42,12 +43,35 @@ func credentials() (uint32, uint32) {
...
+		return home
+	}
+
+	// if we are the root user, use "/root"
+	if unix.Getuid() == 0 {
+		return "/root"
+	}
+
+	// nothing safe to do
+	return "/dev/null"
```
schmichael, line 64:
```suggestion
	return "/nonexistent"
```

Let's use what `nobody` uses for home. I'd hate for an accidental misuse of this func (even in a test) to overwrite someone's `/dev/null` or similar. :sweat_smile:

client/allocrunner/taskrunner/getter/util_linux.go
shoenig: > Does landlock require that the file already exists? 

Yup. But this code is running in the artifact downloader child process - if you get to this point without your ssh config, you're gonna have a bad time anyway.

client/allocrunner/taskrunner/getter/util_linux.go
shoenig: Good call! Swapped out UserHomeDir for go-homedir; this makes the tests a bit more interesting to setup but seems worth it.

tgross (approved):
LGTM :+1:

PR #16511, opened 2023-03-15
artifact: do not set process attributes on darwin

shoenig:
This PR fixes the non-root macOS use case where artifact downloads
stopped working. It seems setting a Credential on a SysProcAttr
used by the exec package will always cause fork/exec to fail -
even if the credential contains our own UID/GID or nil UID/GID.

Technically we do not need to set this as the child process will
inherit the parent UID/GID anyway... and not setting it makes
things work again ... /shrug

### DIFF
--- a/client/allocrunner/taskrunner/getter/util_default.go
+++ b/client/allocrunner/taskrunner/getter/util_default.go
@@ -7,23 +7,19 @@ import (
 	"syscall"
 )
 
-// attributes returns the system process attributes to run
-// the sandbox process with
+// attributes is not implemented by default
 func attributes() *syscall.SysProcAttr {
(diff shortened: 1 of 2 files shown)


tgross (approved):
LGTM

This looks like it should have worked as-is to me, but I'm having a heck of a time following what's going on with the Darwin behavior here... I wonder if this is a SIP problem? It might have been nice to specifically split out the Darwin behavior if we thought non-Darwin Unixen worked, but that's not really something we have in our test environment at this point so it seems safe to ship as-is and if someone really wants to improve artifact sandboxing on FreeBSD or whatever they can contribute a patch for it :grinning:

PR #16513, opened 2023-03-15
artifact: use specific version link for zipbomb artifact

shoenig:
Fix the e2e case where we download the go-getter bomb.zip test file, which
is being removed on main. We can still get it from the version tag - yay git!


tgross (approved):
LGTM :+1:

PR #16519, opened 2023-03-16
acl: HTTP endpoints for JWT auth

pkazmierczak:
This PR introduces HTTP endpoints for JWT auth.

### DIFF
--- a/api/acl.go
+++ b/api/acl.go
@@ -442,19 +442,19 @@ func (a *ACLBindingRules) Get(bindingRuleID string, q *QueryOptions) (*ACLBindin
 	return &resp, qm, nil
 }
 
-// ACLOIDC is used to query the ACL OIDC endpoints.
-type ACLOIDC struct {
+// ACLAuth is used to query the ACL OIDC endpoints.
+type ACLAuth struct {
 	client *Client
 }
 
-// ACLOIDC returns a new handle on the ACL auth-methods API client.
-func (c *Client) ACLOIDC() *ACLOIDC {
-	return &ACLOIDC{client: c}
+// ACLAuth returns a new handle on the ACL auth-methods API client.
+func (c *Client) ACLAuth() *ACLAuth {
+	return &ACLAuth{client: c}
 }
 
 // GetAuthURL generates the OIDC provider authentication URL. This URL should
 // be visited in order to sign in to the provider.
-func (a *ACLOIDC) GetAuthURL(req *ACLOIDCAuthURLRequest, q *WriteOptions) (*ACLOIDCAuthURLResponse, *WriteMeta, error) {
+func (a *ACLAuth) GetAuthURL(req *ACLOIDCAuthURLRequest, q *WriteOptions) (*ACLOIDCAuthURLResponse, *WriteMeta, error) {
 	var resp ACLOIDCAuthURLResponse
 	wm, err := a.client.put("/v1/acl/oidc/auth-url", req, &resp, q)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


api/acl.go
```diff
@@ -793,6 +790,132 @@ func (m *ACLAuthMethod) UnmarshalJSON(data []byte) error {
...
+		*Alias
+	}{
+		Alias: (*Alias)(c),
+	}
+	if err := json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+	var err error
+	if aux.ExpirationLeeway != nil {
+		switch v := aux.ExpirationLeeway.(type) {
```
tgross, line 881: Is the idea that if we're neither type we just fall thru and the value on `ACLAuthMethodConfig` is unset so that it's handled in validation later?

pkazmierczak: No, it's a mistake. I will add a `default` case, we should error if the type for any of those leeways is other than string or float.

pkazmierczak: addressed in [4c1d047](https://github.com/hashicorp/nomad/pull/16519/commits/4c1d0474cd8d5cbce93d2d3c59018b09a37c87d5)

command/agent/acl_endpoint.go
```diff
@@ -874,3 +874,22 @@ func (s *HTTPServer) ACLOIDCCompleteAuthRequest(resp http.ResponseWriter, req *h
 	setIndex(resp, out.Index)
 	return out.ACLToken, nil
 }
+
+// ACLLoginRequest performs a non-interactive authentication request
+func (s *HTTPServer) ACLLoginRequest(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
```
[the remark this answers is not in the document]
pkazmierczak: [4c1d047](https://github.com/hashicorp/nomad/pull/16519/commits/4c1d0474cd8d5cbce93d2d3c59018b09a37c87d5)

api/acl.go
```diff
@@ -442,19 +442,19 @@ func (a *ACLBindingRules) Get(bindingRuleID string, q *QueryOptions) (*ACLBindin
 	return &resp, qm, nil
 }
 
-// ACLOIDC is used to query the ACL OIDC endpoints.
-type ACLOIDC struct {
+// ACLAuth is used to query the ACL OIDC endpoints.
+type ACLAuth struct {
 	client *Client
 }
```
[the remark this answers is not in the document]
pkazmierczak: addressed in [133b2a0](https://github.com/hashicorp/nomad/pull/16519/commits/133b2a09192cb8e21cddecce86d0fb8a13bdd3c0)

api/acl.go
```diff
@@ -443,18 +444,32 @@ func (a *ACLBindingRules) Get(bindingRuleID string, q *QueryOptions) (*ACLBindin
...
-	client *Client
+	ACLAuth
 }
 
 // ACLOIDC returns a new handle on the ACL auth-methods API client.
-func (c *Client) ACLOIDC() *ACLOIDC {
-	return &ACLOIDC{client: c}
+//
+// Deprecated: c.ACLOIDC() is deprecated, use c.ACLAuth() instead.
+func (c *Client) ACLOIDC() *ACLAuth {
```
tgross, line 456: Although I think the methods still work, I'm pretty sure this changes the signature in a backwards incompatible way. Consider the case where someone has this code:

```go
var oidcAPI *ACLOIDC
oidcAPI = c.cACLOIDC

token , wm, err := oidcAPI.CompleteAuth(...)
```

It will no longer compile because of the type error and you'll get  `cannot use c.ACLOIDC() (value of type *ACLAuth) as *ACLOIDC value in assignment`. So we need to have `c.ACLOIDC` return `*ACLOIDC` still.

pkazmierczak: oof, nice catch, thanks. I didn't consider it. Addressed in [36563a5](https://github.com/hashicorp/nomad/pull/16519/commits/36563a5c5587553d728028472ea4b2ec895e0bd8)

tgross (approved):
LGTM!

PR #16530, opened 2023-03-16
Enable ACLs on E2E test clients

schmichael:
I also updated the docs to hopefully make it more clear that ACLs should be uniformly enabled or disabled on all nodes. I wouldn't mind making it bold or a note or something because I really don't want to give the impression we support a mixed-auth world. Hopefully in the near future it will be an auth-only world and this easy-get-wrong config will go away!


tgross (approved):
LGTM! I always forget about enabling them on the clients during development :grimacing:

PR #16531, opened 2023-03-17
acl: fix canonicalization of JWT auth method mock

pkazmierczak:


jrasell (approved):
LGTM! Not sure if we need to backport these; so making a note here in case labels need to be added.

shoenig (approved):

PR #16532, opened 2023-03-17
acl: JWT auth CLI

pkazmierczak:
This PR introduces CLI command adjustments for JWT-based auth.

### DIFF
--- a/command/acl_auth_method_create.go
+++ b/command/acl_auth_method_create.go
@@ -50,8 +50,8 @@ ACL Auth Method Create Options:
     between 1-128 characters and is a required parameter.
 
   -type
-    Sets the type of the auth method. Currently the only supported type is
-    'OIDC'.
+    Sets the type of the auth method. Currently the only supported types are
+    'OIDC' and 'JWT'.
 
   -max-token-ttl
     Sets the duration of time all tokens created by this auth method should be
@@ -83,7 +83,7 @@ func (a *ACLAuthMethodCreateCommand) AutocompleteFlags() complete.Flags {
 	return mergeAutocompleteFlags(a.Meta.AutocompleteFlags(FlagSetClient),
 		complete.Flags{
 			"-name":           complete.PredictAnything,
-			"-type":           complete.PredictSet("OIDC"),
+			"-type":           complete.PredictSet("OIDC", "JWT"),
 			"-max-token-ttl":  complete.PredictAnything,
 			"-token-locality": complete.PredictSet("local", "global"),
 			"-default":        complete.PredictSet("true", "false"),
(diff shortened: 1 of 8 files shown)


command/acl_auth_method_create.go
```diff
@@ -50,8 +50,8 @@ ACL Auth Method Create Options:
     between 1-128 characters and is a required parameter.
 
   -type
-    Sets the type of the auth method. Currently the only supported type is
-    'OIDC'.
+    Sets the type of the auth method. Currently the only supported types are
+    'OIDC' and 'JWT'.
```
lgfa29, line 54:
```suggestion
    Sets the type of the auth method. Supported types are 'OIDC' and 'JWT'.
```

Same for `update`. We also need to update the [website](https://github.com/hashicorp/nomad/blob/main/website/content/docs/commands/acl/auth-method/create.mdx?plain=1#L32-L33).

command/login.go
```diff
@@ -50,7 +51,12 @@ Login Options:
 
   -oidc-callback-addr
     The address to use for the local OIDC callback server. This should be given
-    in the form of <IP>:<PORT> and defaults to "localhost:4649".
+    in the form of <IP>:<PORT> and defaults to "localhost:4649". It is only
+    required if using OIDC auth method type.
```
lgfa29, line 55:
```suggestion
    in the form of <IP>:<PORT> and defaults to "localhost:4649".
```

I think the flag name make this clear already?

command/login.go
```diff
@@ -50,7 +51,12 @@ Login Options:
 
   -oidc-callback-addr
     The address to use for the local OIDC callback server. This should be given
-    in the form of <IP>:<PORT> and defaults to "localhost:4649".
+    in the form of <IP>:<PORT> and defaults to "localhost:4649". It is only
+    required if using OIDC auth method type. 
+
+  -login-token
```
lgfa29, line 57:
```suggestion
  -login-token=<token>
```

We usually document flags that take values to differentiate them from boolean types (like `-json` below), but I don't know how consistent we are about it 😅

command/login.go
```diff
@@ -71,6 +77,7 @@ func (l *LoginCommand) AutocompleteFlags() complete.Flags {
 		complete.Flags{
 			"-method":             complete.PredictAnything,
 			"-oidc-callback-addr": complete.PredictAnything,
+			"-login-token":        complete.PredictNothing,
```
lgfa29, line 80:
```suggestion
			"-login-token":        complete.PredictAnything,
```

It doesn't _really_ matter in this case, but `PredictAnything` calls the predictor function, while `PredictNothing` doesn't.

lgfa29 (approved):
Minor nit-picking, but LGTM!

command/login.go
pkazmierczak: eh we don't do this for any of the auth method commands...

PR #16548, opened 2023-03-17
[ui] Copyable server and client attribute values

### LINKED ISSUE #16539: [ui] Make it easy to copy values from client/server pages
Would be nice to have a copy to clipboard button for the many client values, for example

philrenaud:
Resolves #16539 

Adds Copy buttons to server and client attribute values:

image

In order to get this working within the confines of a (potentially overflow:hidden) table row, added the `@inset` argument, which will modify the background to be transparent, the dom placement to be inline, and to not show a "Copied!" message and defer instead to just the checkmark icon.

### DIFF
--- a/ui/app/components/copy-button.js
+++ b/ui/app/components/copy-button.js
@@ -1,10 +1,11 @@
 import Component from '@ember/component';
 import { task, timeout } from 'ember-concurrency';
-import { classNames } from '@ember-decorators/component';
+import { classNames, classNameBindings } from '@ember-decorators/component';
 import classic from 'ember-classic-decorator';
 
 @classic
 @classNames('copy-button')
+@classNameBindings('inset')
 export default class CopyButton extends Component {
   clipboardText = null;
   state = null;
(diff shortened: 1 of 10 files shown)


schmichael (approved):
ty! :grin:

ui/app/components/copy-button.js
```diff
@@ -1,10 +1,11 @@
 import Component from '@ember/component';
 import { task, timeout } from 'ember-concurrency';
-import { classNames } from '@ember-decorators/component';
+import { classNames, classNameBindings } from '@ember-decorators/component';
 import classic from 'ember-classic-decorator';
 
 @classic
 @classNames('copy-button')
+@classNameBindings('inset')
```
ChaiWithJai, line 8: `nit`: There are only 6 invocations of the `CopyButton` component. Rather than persisting `ember-classic` pattern, this seems like an easy win for us to migrate to Ember Octane via a Glimmer component!

ChaiWithJai (commented):
Looks like a pretty easy win to migrate `CopyButton` to Glimmer while we're here.

ui/app/components/copy-button.js
ChaiWithJai: Roger dodger!

PR #16556, opened 2023-03-20
cli: stream both stdout and stderr when following an alloc.

### LINKED ISSUE #2123: Mixing stdout and stderr logs
Hi guys,

I use the nomad logs -f <job-name> command pretty often, to display stdout and stderr logs, but I always need to execute it two times. What about:

- nomad logs -stdout: only logs on 1
- nomad logs -stderr: only logs on 2
- nomad logs: logs on 1 and 2

Docker respects this rule.

jrasell:
This update changes the behaviour when following logs from an
allocation, so that both stdout and stderr files streamed when the
operator supplies the follow flag. The previous behaviour is held
when all other flags and situations are provided.

Opening in draft to get initial UX feedback.

Todo: figure out tests, likely e2e.
Closes #2123


command/alloc_logs.go
```diff
@@ -195,77 +208,99 @@ func (l *AllocLogsCommand) Run(args []string) int {
...
 		return 1
 	}
 
-	logType := "stdout"
-	if stderr {
-		logType = "stderr"
+	// In order to run the mixed log output, we can only follow the files from
+	// their current positions. There is no way to interleave previous log
+	// lines as there is no timestamp references.
+	if !l.stderr && !l.stdout && l.follow && !l.tail && l.numLines < 0 && l.numBytes < 0 {
```
lgfa29, line 232:
```suggestion
	if l.follow && !(l.stderr || l.stdout || l.tail || l.numLines > 0 || l.numBytes > 0) {
```

I was a little confused with this check so I think applying a DeMorgan on the negative components makes the rule a bit more clear? You need `-f` and can't have any of the others.

command/alloc_logs.go
```diff
@@ -275,30 +310,103 @@ func (l *AllocLogsCommand) followFile(client *api.Client, alloc *api.Allocation,
...
+	signal.Notify(signalCh, os.Interrupt, syscall.SIGTERM)
+
+	// This go routine blocks until the command receives an interrupt or
+	// terminate signal, at which point we close the ReadCloser.
 	go func() {
 		<-signalCh
-
-		// End the streaming
-		r.Close()
+		_ = r.Close()
```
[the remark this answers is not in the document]
jrasell: Good spot; I can fix this up in another PR.

helper/cli/ui.go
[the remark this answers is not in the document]
jrasell: I don't want to put it into the `command` package as that is already huge, but you're right and I'll move it.

command/alloc_logs.go
```diff
@@ -195,77 +208,95 @@ func (l *AllocLogsCommand) Run(args []string) int {
...
-		logType = "stderr"
+	// In order to run the mixed log output, we can only follow the files from
+	// their current positions. There is no way to interleave previous log
+	// lines as there is no timestamp references.
+	if l.follow && !(l.stderr || l.stdout || l.tail || l.numLines > 0 || l.numBytes > 0) {
+		if err := l.tailMultipleFiles(client, alloc); err != nil {
+			l.Ui.Error(fmt.Sprintf("Failed to tail stdout and stderr files: %v", err))
+			return 1
+		}
+	} else {
```
lgfa29, line 237:
```suggestion
	if l.follow && !(l.stderr || l.stdout || l.tail || l.numLines > 0 || l.numBytes > 0) {
		if err := l.tailMultipleFiles(client, alloc); err != nil {
			l.Ui.Error(fmt.Sprintf("Failed to tail stdout and stderr files: %v", err))
			return 1
		}
		return 0
	}
```

By returning early we can remove the `else` clause and dedent the rest of the code, which may be nice, but very much a style nit-pick 😄

command/alloc_logs.go
```diff
@@ -195,77 +208,95 @@ func (l *AllocLogsCommand) Run(args []string) int {
...
-			r = NewLineLimitReader(r, int(numLines), int(numLines*bytesToLines), 1*time.Second)
+		if l.numLines != -1 {
+			r = NewLineLimitReader(r, int(l.numLines), int(l.numLines*bytesToLines), 1*time.Second)
 		}
 
 		if readErr != nil {
-			readErr = fmt.Errorf("Error tailing file: %v", readErr)
+			return fmt.Errorf("error tailing file: %v", readErr)
 		}
 	}
```
lgfa29, line 292: Not part of the PR, but it may be worth flipping the lines to check for `readErr != nil` earlier since `r` will be `nil`

command/alloc_logs.go
```diff
@@ -195,77 +208,95 @@ func (l *AllocLogsCommand) Run(args []string) int {
...
+}
+
+func (l *AllocLogsCommand) handleSingleFile(client *api.Client, alloc *api.Allocation, logType string) error {
 	// We have a file, output it.
 	var r io.ReadCloser
 	var readErr error
-	if !tail {
-		r, readErr = l.followFile(client, alloc, follow, task, logType, api.OriginStart, 0)
+	if !l.tail {
+		r, readErr = l.followFile(client, alloc, l.follow, l.task, logType, api.OriginStart, 0)
```
lgfa29, line 264:
```suggestion
		r, readErr = l.followFile(client, alloc, logType, api.OriginStart, 0)
```

Since `followFile` is a method and now `follow` and `task` are in `l`, we could remove these, but it's not really necessary 😄

command/alloc_logs.go
```diff
@@ -275,30 +306,103 @@ func (l *AllocLogsCommand) followFile(client *api.Client, alloc *api.Allocation,
...
+	// Use a single cancel channel for both log streams, so we only have to
+	// close one.
+	cancel := make(chan struct{})
+
+	// Ensure the channel is closed in order to notify listeners whenever we
+	// exit.
+	defer close(cancel)
+
+	stdoutFrames, stdoutErrCh := client.AllocFS().Logs(
+		alloc, true, l.task, api.FSLogNameStdout, api.OriginEnd, 1, cancel, nil)
```
lgfa29, line 352: Reading the [docs](https://github.com/hashicorp/nomad/blob/43e25410d0bdd2ce0fc067bd4946cc50c9e54c58/api/fs.go#L238) I _think_ passing a `ctx.Done()` instead of creating a `cancel` channel would work, and it would make the code look more like modern Go.

You can then even pass the context to the function (`tailMultipleFiles(ctx, ...)`) and have let the parent `Run()` handle the lifecycle of the request, so that things like signal handling are handled outside of these sub-functions.

```go

func (l *AllocLogsCommand) Run(...) {
	// ...
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()
	// ...
	signalCh := make(chan os.Signal, 1)
	signal.Notify(signalCh, os.Interrupt, syscall.SIGTERM)
	go func() {
		for {
			select {
			case <-signalCh:
				return 0
			}
		}
	}()
	// ...
	tailMultipleFiles(ctx, ...)
	// ...
	followFile(ctx, ...)
}

func (l *AllocLogsCommand) tailMultipleFiles(ctx ...) error {
	// ...
	stderrFrames, stderrErrCh := client.AllocFS().Logs(
		alloc, true, l.task, api.FSLogNameStderr, api.OriginEnd, 1, ctx.Done(), nil)
	// ...
}

```

lgfa29 (approved):
Looking good!
image

There seems to always be some empty lines first, but I don't know where those are coming from. Left some random comments, but nothing major so feel free to ignore them

website/content/docs/commands/alloc/logs.mdx
```diff
@@ -35,6 +35,10 @@ When ACLs are enabled, this command requires a token with the `read-logs`,
 
 ## Logs Options
 
+- `-stdout`: Display stdout logs. This is used as the default value in all
+  commands expect when using the `-f` flag where both stdout and stderr are
```
jrasell, line 39:
```suggestion
  commands except when using the `-f` flag where both stdout and stderr are
```

e2e/alloc_logs/alloc_logs_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	// Run the alloc logs command which we expect to capture both stdout and
+	// stderr logs. The command will reach its timeout and therefore return an
+	// error. We want to ignore this, as it's expected. Any other error is
+	// terminal.
+	out, err := e2eutil.Command("nomad", "alloc", "logs", "-f", allocStubs[0].ID)
+	if err != nil {
+		must.ErrorContains(t, err, "failed: signal: killed")
+	}
+	must.StrContains(t, out, "stdout\nstderr\nstdout\nstderr\nstdout\nstderr")
```
lgfa29, line 44:
```suggestion
	must.StrContains(t, out, "stdout\nstderr")
```
Is there a risk of a timing issue if the alloc doesn't run long enough to generate 3 sets of `stdout\nstderr`? Checking just for one occurrence may be enoug? 😅

lgfa29 (approved):

e2e/alloc_logs/alloc_logs_test.go
jrasell: I think we are safe, because the test waits until the allocation is running before continuing to this portion and the context will mean the command collects logs for 10 seconds. That being said, I think you're right that checking a single mixed set is enough.

PR #16569, opened 2023-03-20
contrib: architecture guide to the drainer

tgross:
The drainer component is fairly complex. As part of upcoming work to fix some of the drainer's rough edges, document the drainer's architecture from a Nomad developer perspective.

[Preview link for the diagrams](https://github.com/hashicorp/nomad/blob/drain-legibility-arch-docs/contributing/architecture-drainer.md)

### DIFF
--- a/contributing/architecture-drainer.md
+++ b/contributing/architecture-drainer.md
@@ -151,7 +151,8 @@ And the third is when nodes pass their deadline. The `NodeWatcher` is
 responsible for adding and removing the watch in the `DeadlineNotifier`. The
 `DeadlineNotifier` is responsible for watching the timer. If the node isn't
 removed before the deadline, the `DeadlineNotifier` tells the `NodeDrainer` and
-the `NodeDrainer` updates the state via the `NodesDrainComplete` shim.
+the `NodeDrainer` updates the state via the `NodesDrainComplete` shim. At this
+point the remaining allocs will be forced to shutdown immediately.
 
 ```mermaid
 flowchart TD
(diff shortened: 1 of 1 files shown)


lgfa29 (approved):
Nice! A minor question about drain deadline, but that's me not knowing about how Nomad works 😅

contributing/architecture-drainer.md
```diff
@@ -0,0 +1,185 @@
+# Architecture: Drainer
+
+The drainer is a component that runs on the leader that services requests from
+the [`nomad node drain`][] command and related workflows in the web UI. For a
+play-by-play from the user's perspective, see [node drain tutorial][]. This
+document describes the internals of the drainer for Nomad developers.
+
+The high-level workflow is that:
+* The user sets the drain state of the client ("node") in the state store.
```
schmichael, line 9: Should we capitalize Client or both nouns? We've started capitalizing Client in some docs, and I think that's probably a good idea to differentiate the _Nomad Client Agent (represented by the Node object)_ from _clients_ like the API package, the CLI, and whatever else that extremely common software noun conjures up.

Some pages where we Uppercase Client:
 - https://developer.hashicorp.com/nomad/docs/operations/metrics-reference
 - https://developer.hashicorp.com/nomad/docs/upgrade/upgrade-specific
 - https://developer.hashicorp.com/nomad/docs/configuration/server

```suggestion
* The user sets the drain state of the Client ("Node") in the state store.
```

schmichael: Out of scope for this, but a Styleguide would be nice

schmichael (approved):
Thanks! :tada:

contributing/architecture-drainer.md
tgross: @angrycub has turned me on to https://vale.sh/ which is great and would let us have automated style guides for the different audiences we write prose for (ex. tutorials vs user-facing docs vs developer docs vs marketing).

PR #16572, opened 2023-03-20
chore: add mirage fixture

ChaiWithJai:
This PR adds a variable table so the Vercel deployment displays a variables table.


philrenaud (approved):
This should probably be a commit in #16570 rather than its own PR

Jordy1997 (approved):

PR #16573, opened 2023-03-20
contrib: developer docs for `mock_driver`

tgross:
We don't have any handy documentation for the `mock_driver` that we use for internal testing. This isn't exposed to end users so it doesn't belong in the real documentation, but it can live in the contributing guide.


philrenaud (approved):
This is awesome, and thanks for shining light on it! I imagine I'll use it pretty regularly now that I get the gist.

PR #16575, opened 2023-03-20
Adds public roadmap project to readme

mikenomitch:
TODO: Roadmap must be make public before merging


apollo13:
Just coming here to say that this would be great to have. The roadmap doesn't even have to have dates, just an ordering what you are trying to do next would be great! :+1:

jrasell (approved):
LGTM.

PR #16593, opened 2023-03-21
ui:  update `parse` method on `Job` model to accept variables

### LINKED ISSUE #16589: Refactor `Mirage` for `Job` model changes

ChaiWithJai:
Resolves #16589. This PR updates the `parse` adapter and model methods to enable passing `HCL Variables` to the `/parse` endpoint and also creates a new private property on the `Job` model to store `HCL Variables`.

Note:  We do not know the request schema for the `/parse` endpoint. We have not addressed how we'll handle complex data types like tuples, maps and etcetera when sending this request. That will be resolved in a later PR when its time to integrate the API.

### DIFF
--- a/ui/app/adapters/job.js
+++ b/ui/app/adapters/job.js
@@ -34,11 +34,12 @@ export default class JobAdapter extends WatchableNamespaceIDs {
     return this.ajax(url, 'DELETE');
   }
 
-  parse(spec) {
+  parse(spec, jobVars) {
     const url = addToPath(this.urlForFindAll('job'), '/parse?namespace=*');
     return this.ajax(url, 'POST', {
       data: {
         JobHCL: spec,
+        JobVariables: jobVars,
         Canonicalize: true,
       },
     });
--- a/ui/app/models/job.js
+++ b/ui/app/models/job.js
@@ -248,6 +248,7 @@ export default class Job extends Model {
 
   parse() {
     const definition = this._newDefinition;
+    const variables = this._newDefinitionVariables;
     let promise;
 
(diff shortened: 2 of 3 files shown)


ui/app/adapters/job.js
```diff
@@ -34,11 +34,12 @@ export default class JobAdapter extends WatchableNamespaceIDs {
     return this.ajax(url, 'DELETE');
   }
 
-  parse(spec) {
+  parse(spec, jobVars) {
     const url = addToPath(this.urlForFindAll('job'), '/parse?namespace=*');
     return this.ajax(url, 'POST', {
       data: {
         JobHCL: spec,
+        JobVariables: jobVars,
```
philrenaud, line 42: since we don't know what the eventual format expected for these will be, could you add a default value here? Whether that's an empty string, or a stringified object, or whatever the presumed format might be.

This'll help with rough edges if we try to call jobAdapter.parse from any other situation without jobVars included later, and as a side benefit we'll get some type hinting.

philrenaud (approved):

ui/app/adapters/job.js
ChaiWithJai: This shouldn't have a default value other than `undefined` because it's an optional argument. Jobs do not need to include variables even if the specification was written in HCL2.

In order to think through that edge case, the only viable option I could think of was including defensive logic in the `model` method that is responsible for calling the `adapter` method here by ensuring variables are validly written in `JSON`. Otherwise, we'll dispatch the request and rely on the API to throw an error.

PR #16594, opened 2023-03-21
taskapi: use HasSuffix to detect errors from rpcs

schmichael:
Matches the "normal" HTTP error detection logic in the same file.

Tested by `TestTaskAPI/testTaskAPI_Auth` (which is currently failing in the nightly run, but passes against this patch).


lgfa29 (approved):
I ran into a similar problem but trying to use `errors.Is` 😬

PR #12916, opened 2022-05-06
api: remove `mapstructure` tags from`Port` struct

DerekStrickland:
This PR solves a defect in the deserialization of `api.Port` structs when returning structs from the`EventStream`. 

Previously, the `api.Port` struct's fields were decorated with both `mapstructure` and `hcl` tags to support the `network.port` stanza's use of the keyword `static` when posting a static port value. This works fine when posting a job and when retrieving any struct that has an embedded `api.Port` instance as long as the value is deserialized using `JSON` decoding. The `EventStream`, however, uses `mapstructure` to decode event payloads in the `api` package. `mapstructure` expects an underlying field named `static` which does not exist. The result was that the `Port.Value` field would always be set to `0`.

Upon further inspection, a few things became apparent.

- The struct already has `hcl` tags that support the indirection during job submission.
- Serialization/deserialization with both the `json` and `hcl` packages produce the desired result.
- The use of of the `mapstructure` tags provided no value as the `Port` struct contains only fields with primitive types.

This PR:
- Removes the `mapstructure` tags from the `api.Port` structs
- Updates the job parsing logic to use `hcl` instead of `mapstructure` when decoding `Port` instances.

Closes #11044

PR #16595, opened 2023-03-21
E2E: fix events tests

tgross:
In #12916 we updated the events test as part of a larger set of changes around mapstructure serialization fixes. But the changes to the jobs we're deploying in the tests had invalid task configs so they never result in good deployments and the test will always fail. Make the before/after jobs identical (except for the version bump) and make them valid. Also wait for allocations for the 2nd job run to appear before checking the deployment list, so that we don't race with the scheduler.


schmichael (approved):

schmichael:
Merging since Tim is out for the day, and I'd like this to get into the e2e test suite ASAP.

PR #16608, opened 2023-03-22
[ui] Deployment History search

philrenaud:
- Adds a "loading shimmer" and rolling slide-in for events:
image

- Adds a searchbox:
image

Search term persists as a deployment rolls forward; which is to say, if you search a term when there are no events that match it, but a placed allocation matches it later, it will show up with no further user action needed.

### DIFF
--- a/ui/app/components/job-status/deployment-history.js
+++ b/ui/app/components/job-status/deployment-history.js
@@ -55,6 +55,7 @@ export default class JobStatusDeploymentHistoryComponent extends Component {
             .flat()
         )
         .flat()
+        .filter((a) => this.filterOnSearchTerm(a))
         .sort((a, b) => a.get('time') - b.get('time'))
         .reverse();
     } catch (e) {
@@ -71,4 +72,25 @@ export default class JobStatusDeploymentHistoryComponent extends Component {
       color: 'critical',
     });
   }
+
+  // #region search
+
+  /**
+   * @type { string }
+   */
+  @tracked searchTerm = '';
+
+  /**
+   * @param { import('../../models/task-event').default } taskEvent
+   * @returns { boolean }
+   */
+  filterOnSearchTerm(taskEvent) {
+    return (
+      taskEvent.message.toLowerCase().includes(this.searchTerm.toLowerCase()) ||
+      taskEvent.type.toLowerCase().includes(this.searchTerm.toLowerCase()) ||
+      taskEvent.state.allocation.shortId.includes(this.searchTerm.toLowerCase())
+    );
+  }
+
+  // #endregion search
 }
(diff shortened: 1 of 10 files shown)


ui/mirage/factories/allocation.js
```diff
@@ -13,7 +13,7 @@ const REF_TIME = new Date();
 export default Factory.extend({
   id: (i) => (i >= 100 ? `${UUIDS[i % 100]}-${i}` : UUIDS[i]),
 
-  jobVersion: 1,
+  jobVersion: 0,
```
philrenaud, line 16: TBD: jobVersion is set to 1 in our tests, but Nomad jobVersions start at 0. Setting this to 0 to try to make Mirage a little closer to real life. If this breaks a lot of things, will revert.

ui/mirage/factories/task-event.js
```diff
@@ -12,5 +12,7 @@ export default Factory.extend({
   exitCode: () => null,
   time: () => faker.date.past(2 / 365, REF_TIME) * 1000000,
 
-  displayMessage: () => faker.lorem.sentence(),
+  // TODO: I think this is probably an archaic property name. See if tests collapse with this removal and decide to remove then.
+  // displayMessage: () => faker.lorem.sentence(),
+  message: () => faker.lorem.sentence(),
```
ChaiWithJai, line 17: `blocking`: I'm getting 99 test failures. ~97 of them are associated to this diff. Unsure if this is the root of all the test failures...

ui/app/components/job-status/deployment-history.js
```diff
@@ -71,4 +72,25 @@ export default class JobStatusDeploymentHistoryComponent extends Component {
       color: 'critical',
     });
   }
+
+  // #region search
+
+  /**
+   * @type { string }
+   */
+  @tracked searchTerm = '';
```
ChaiWithJai, line 81: `question`: I noticed that this is not associated to a query parameter, as you've requested in previous PRs. Any color about why we're not using an associated query parameter here?

ChaiWithJai (changes requested):
There's many test failures associated with the Mirage change.

ui/mirage/factories/task-event.js
philrenaud: Hi @ChaiWithJai -- please see the code comment TODO in the code here. This was run to see if tests collapse.

ui/app/components/job-status/deployment-history.js
philrenaud: Good question! In this case, this is a conditional component — it only exists when a deployment is active (a much smaller % of the time than when it's in steady state).

If I wanted to add a queryParam that controlled this from the job index route, there would be a majority of time where that queryParam did nothing.

There's an argument to be made for including it anyway, but it seems pretty outweighed by the argument that each deployment is separate and search terms aren't necessarily something you'd want to maintain between all of them.

ChaiWithJai: Gotcha!

ChaiWithJai (approved):

PR #16612, opened 2023-03-22
drainer: stop watching down or disconnected nodes

tgross:
When a node is down or disconnected, we can no longer gracefully migrate its allocations. Any evaluations we need to replace the allocations will have already been created by the heartbeater, so there's no more work for the drainer to do. Stop watching nodes in this state.

This changeset also refactors the tests of the draining node watcher so that we don't mock the node watcher's `Remove` and `Update` methods for its own tests. Instead we'll mock the node watcher's dependencies (the job watcher and deadline notifier) and now unit tests can cover the real code.


nomad/drainer/drain_testing.go
```diff
@@ -1,52 +1,142 @@
...
+		enabled: false,
+		logger:  logger,
+		nodes:   map[string]*drainingNode{},
+		jobWatcher: &MockJobWatcher{
+			drainCh:    make(chan *DrainRequest),
+			migratedCh: make(chan []*structs.Allocation),
+			jobs:       map[structs.NamespacedID]struct{}{},
+		},
+		deadlineNotifier: &MockDeadlineNotifier{
+			expiredCh: make(<-chan []string),
```
schmichael, line 130: Hm, not sure I understand the point of drainCh, migratedCh, and expiredCh. They don't seem to ever have any writes, so they're all effectively nil chans?

nomad/drainer/drain_testing.go
```diff
@@ -1,52 +1,142 @@
...
+type MockRaftApplierShim struct {
+	state *state.StateStore
+}
+
+// AllocUpdateDesiredTransition mocks a write to raft as a state store update
+func (m *MockRaftApplierShim) AllocUpdateDesiredTransition(
+	allocs map[string]*structs.DesiredTransition, evals []*structs.Evaluation) (uint64, error) {
+	index, _ := m.state.LatestIndex()
+	index++
+	err := m.state.UpdateAllocsDesiredTransitions(structs.MsgTypeTestSetup, index, allocs, evals)
```
schmichael, line 87: I think this needs locking to ensure data races can't cause bad interleaving with concurrent index reads and increments.

nomad/drainer/drain_testing.go
```diff
@@ -1,52 +1,142 @@
...
+	return index, err
+}
+
+// NodesDrainComplete mocks a write to raft as a state store update
+func (m *MockRaftApplierShim) NodesDrainComplete(
+	nodes []string, event *structs.NodeEvent) (uint64, error) {
+	index, _ := m.state.LatestIndex()
+	index++
```
schmichael, line 95: Likewise might need locking as above?

nomad/drainer/watch_nodes.go
```diff
@@ -179,21 +174,31 @@ func (w *nodeDrainWatcher) watch() {
...
+				// Down or Disconnected nodes can't be migrated, so skip
+				// tracking them
+
 			case !tracked && newDraining:
+				// If the node is not being tracked but is draining, track
 				w.tracker.Update(node)
 
-				// If the node is being tracked but has changed, update:
+			case tracked && node.UnresponsiveStatus():
+				// Down or Disconnected nodes can't be migrated, so stop
```
schmichael, line 191: What about `unknown` allocs on `disconnected` nodes? Will something else ensure they're marked with DesiredStatus=stop if they reconnect?

nomad/drainer/watch_nodes.go
```diff
@@ -241,15 +245,6 @@ func (w *nodeDrainWatcher) getNodesImpl(ws memdb.WatchSet, state *state.StateSto
 
 		node := raw.(*structs.Node)
 		resp[node.ID] = node
-		if maxIndex < node.ModifyIndex {
-			maxIndex = node.ModifyIndex
```
schmichael, line 245: This was unreachable right?! It seems like if a `Node.ModifyIndex > Index("nodes")` we have a bug?! I know I :heavy_check_mark: the original PR, but I can't reason about what this did...

nomad/drainer/watch_nodes.go
```diff
@@ -179,21 +174,31 @@ func (w *nodeDrainWatcher) watch() {
...
 
 			switch {
-			// If the node is tracked but not draining, untrack
+
 			case tracked && !newDraining:
+				// If the node is tracked but not draining, untrack
 				w.tracker.Remove(nodeID)
 
-				// If the node is not being tracked but is draining, track
+			case !tracked && node.UnresponsiveStatus():
```
schmichael, line 182: Since this is a noop should we even bother computing it? I suppose the compiler might elide it, but that seems awfully aggressive.

schmichael (approved):
Some questions, but definitely seems like a step in the right direction.

nomad/drainer/watch_nodes.go
jrasell: This computation takes 1ns or less, so I feel having it clearly understandable when looking at the code is worth it. Alternatively, we could have a big old comment on the default case to detail the matchers.

tgross: This is a switch case so we'll always take the first matching branch. Having this no-op check prevents the node from falling thru to one of the other `!tracked` cases that come later.

nomad/drainer/drain_testing.go
tgross: Good catch, will fix.

nomad/drainer/watch_nodes.go
tgross: Removed in 4c8b5566ab0f681df2d0cabd823467f149eb8ad2

tgross:
Although there's ultimately no fix here to get backported, I'm going to backport this PR because I anticipate wanting to backport one or more bug fixes for other drain issues I'm looking at. Having these test improvements available on the backport branches will be handy for ensuring nothing is broken there.

schmichael (approved):

nomad/drainer/watch_nodes.go
schmichael: (Great job pointing out a `down` node can come back before we reschedule its work! Distributed systems :flushed:)

nomad/drainer/watch_nodes.go
tgross: > Since this wakes up any time any node is updated (both before and after this patch AFAICT), I wonder if we should rate limit it a bit to avoid iterating over every node as fast as possible. 

There's a rate limiter set to 100 state queries / second, which is shared between the job watcher and node watcher. The job watcher does a blocking query on allocs for just the tracked jobs.

> Also I don't even know if using an index at all is useful here if we're just waking up on every "nodes" table change. I think we could just block on the WatchSet ourselves. Performance-wise I don't expect this to really matter though, especially if we start coalescing iterations over a short timeframe.

From a code structure standpoint `(*StateStore).BlockingQuery` gives us nice QoL things like exiting when the context is canceled that we'd have to implement separately for a long-lived poll like this. I did some very rough testing and even running a very tight test with a single node I'm seeing queries about ~100ms apart. The real limiting factor is going to be how quickly we can write raft entries and the rest of the time it'll just block. So in practice I don't think we're going to keep this particular goroutine spinning on-CPU all that much.

That being said, I'm going to be taking a deep dive into the job watcher as my next task (to try and finally land https://github.com/hashicorp/nomad/pull/14348), so I can do some more detailed performance analysis as part of that work.

PR #16618, opened 2023-03-23
Updates complete alloc colour to green for new visualizations only

philrenaud:
Changes the completed colour from dark green to pale green (for the new visualizations only)
image

image


ChaiWithJai (approved):
`nit`: There is no "u" in color. 😉

jk LGTM

PR #16625, opened 2023-03-23
[ui] Web sign-in with JWT

philrenaud:
image

### DIFF
--- a/ui/app/adapters/token.js
+++ b/ui/app/adapters/token.js
@@ -32,6 +32,22 @@ export default class TokenAdapter extends ApplicationAdapter {
     });
   }
 
+  loginJWT(LoginToken, AuthMethodName) {
+    return this.ajax(`${this.buildURL()}/login`, 'POST', {
+      data: {
+        AuthMethodName,
+        LoginToken,
+      },
+    }).then((token) => {
+      const store = this.store;
+      store.pushPayload('token', {
+        tokens: [token],
+      });
+
+      return store.peekRecord('token', store.normalize('token', token).data.id);
+    });
+  }
+
   exchangeOneTimeToken(oneTimeToken) {
     return this.ajax(`${this.buildURL()}/token/onetime/exchange`, 'POST', {
       data: {
--- a/ui/app/adapters/token.js
+++ b/ui/app/adapters/token.js
@@ -21,31 +21,23 @@ export default class TokenAdapter extends ApplicationAdapter {
     return `${this.buildURL()}/${singularize(modelName)}/${identifier}`;
   }
 
-  findSelf() {
-    return this.ajax(`${this.buildURL()}/token/self`, 'GET').then((token) => {
-      const store = this.store;
-      store.pushPayload('token', {
-        tokens: [token],
-      });
-
-      return store.peekRecord('token', store.normalize('token', token).data.id);
-    });
+  async findSelf() {
+    const response = await this.ajax(`${this.buildURL()}/token/self`, 'GET');
+    const normalized = this.store.normalize('token', response);
+    const tokenRecord = this.store.push(normalized);
+    return tokenRecord;
   }
 
-  loginJWT(LoginToken, AuthMethodName) {
-    return this.ajax(`${this.buildURL()}/login`, 'POST', {
+  async loginJWT(LoginToken, AuthMethodName) {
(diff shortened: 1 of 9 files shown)


ui/app/controllers/settings/tokens.js
```diff
@@ -48,31 +49,83 @@ export default class Tokens extends Controller {
...
-        // Clear out all data to ensure only data the new token is privileged to see is shown
-        this.resetStore();
+      this.clearTokenProperties();
 
-        // Refetch the token and associated policies
-        this.get('token.fetchSelfTokenAndPolicies').perform().catch();
+      // Set bearer token instead of findSelf etc.
+      TokenAdapter.loginJWT(secret, methodName).then(
+        (token) => {
+          this.set('token.secret', token.secret);
```
ChaiWithJai, line 90: `question`: Should this be:

```
this.token.setProperties({ secret: token.secret, tokenNotFound: false}) // see line 100
```

ChaiWithJai (approved):

ui/app/controllers/settings/tokens.js
```diff
@@ -48,31 +49,83 @@ export default class Tokens extends Controller {
...
+    const isJWT =
+      secret.length > 36 &&
+      secret.match(/^[a-zA-Z0-9-_]+\.[a-zA-Z0-9-_]+\.[a-zA-Z0-9-_]+$/);
+
+    if (isJWT) {
+      // TODO: Discussion: There may be multiple JWT methods sometime in the future.
+      // A quick assumption would be to just use the first one, but that
+      // won't hold up forever. We should probably have a way to select which one
+      // to use, or at least a way to know which one was used to generate the token.
+      const methodName = await this.authMethods.findBy('type', 'JWT')?.name;
```
philrenaud, line 76: Wanted to make sure this was noted in the PR for discussion: If the agent has multiple authMethods of type `JWT`, this will simply pick the first one and assume that's what the user is attempting to sign in with.

It's very easy for this assumption to be wrong — a user could have multiple JWT auth methods and intend on using one other than the first — but it seems like a real edge case of auth method config.

A straightforward solution to this would be to add a "Which JWT method did you want to sign in with?" prompt, or a button set similar to how we handle OIDC auth methods, but then it becomes pretty cumbersome to an end-user.

My current feeling is that we should ship this with a single JWT auth method supported, and worry about interface expansion later if it becomes pressing.

cc @pkazmierczak @juliezzhou @ChaiWithJai

ui/app/controllers/settings/tokens.js
philrenaud: My fear in changing this was that there was something to resetStore() or fetchSelfTokenAndPolicies that would necessitate them taking place before setting tokenNotFound below. I'll test it out now for code consolidation's sake.

ui/app/adapters/token.js
```diff
@@ -32,6 +32,22 @@ export default class TokenAdapter extends ApplicationAdapter {
...
+        AuthMethodName,
+        LoginToken,
+      },
+    }).then((token) => {
+      const store = this.store;
+      store.pushPayload('token', {
+        tokens: [token],
+      });
+
+      return store.peekRecord('token', store.normalize('token', token).data.id);
```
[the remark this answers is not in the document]
philrenaud: Thanks for the suggestion; made the change to both loginJWT and findSelf, from which this was copied.

ui/app/controllers/settings/tokens.js
pkazmierczak: I agree that from a UI perspective your current design is neater (kudos btw), but I'm not sure about this: 

> a user could have multiple JWT auth methods and intend on using one other than the first — but it seems like a real edge case of auth method config.

Here's why: JWT is mostly going to be used as a service-to-service/machine-to-machine auth method. Thus it's rather likely a user could have multiple JWT-type auth methods with different role/policy bindings or different methods of validation depending on the use case, or depending where they got that token from. I might be wrong but in contrast to OIDC I'd say it's more likely for people to have multiple JWT methods and a single OIDC method. 

@mikenomitch what do you think about this?

philrenaud:
Nice little convenience method added for reviewers using mirage (mocks); any JWT-shaped token that ends in `management` will work the way you might expect. Ditto one that ends with `bad` to simulate an authentication error.

Issue #16623, opened 2023-03-23
Nomad v1.5.2 panics with a segmentation violation


henrikjohansen:
### Nomad version
Nomad `v1.5.2+ent`

### Operating system and Environment details
Ubuntu 20.04 LTS

### Issue
After installing Nomad v1.5.2 over v1.5.1 the nomad agent keeps crashing and never comes online. 

```
Mar 23 16:46:31 prappai03 nomad[14294]: panic: runtime error: invalid memory address or nil pointer dereference
Mar 23 16:46:31 prappai03 nomad[14294]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x18d8e37]
Mar 23 16:46:31 prappai03 nomad[14294]: goroutine 625 [running]:
Mar 23 16:46:31 prappai03 nomad[14294]: github.com/hashicorp/nomad/client/structs.(*AllocHookResources).GetCSIMounts(0x0?)
Mar 23 16:46:31 prappai03 nomad[14294]:         github.com/hashicorp/nomad/client/structs/allochook.go:18 +0x37
Mar 23 16:46:31 prappai03 nomad[14294]: github.com/hashicorp/nomad/client/allocrunner/taskrunner.(*volumeHook).prepareCSIVolumes(0xc001807560, 0xc000d719b0?, >
Mar 23 16:46:31 prappai03 nomad[14294]:         github.com/hashicorp/nomad/client/allocrunner/taskrunner/volume_hook.go:169 +0x1ef
Mar 23 16:46:31 prappai03 nomad[14294]: github.com/hashicorp/nomad/client/allocrunner/taskrunner.(*volumeHook).Prestart(0xc001807560, {0x3165e73?, 0xc00180756>
Mar 23 16:46:31 prappai03 nomad[14294]:         github.com/hashicorp/nomad/client/allocrunner/taskrunner/volume_hook.go:219 +0x38e
Mar 23 16:46:31 prappai03 nomad[14294]: github.com/hashicorp/nomad/client/allocrunner/taskrunner.(*TaskRunner).prestart(0xc001796c00)
Mar 23 16:46:31 prappai03 nomad[14294]:         github.com/hashicorp/nomad/client/allocrunner/taskrunner/task_runner_hooks.go:262 +0x7c5
Mar 23 16:46:31 prappai03 nomad[14294]: github.com/hashicorp/nomad/client/allocrunner/taskrunner.(*TaskRunner).Run(0xc001796c00)
[...]

PR #16638, opened 2023-03-24
client: allow incomplete allocrunners to be removed on restore

tgross:
If an allocrunner is persisted to the client state but the client stops before task runner can start, we end up with an allocation in the database with allocrunner state but no taskrunner state. This ends up mimicking an old pre-0.9.5 state where this state was not recorded and that hits a backwards compatibility shim. This leaves allocations in the client state that can never be restored, but won't ever be removed either.

Update the backwards compatibility shim so that we fail the restore for the allocrunner and remove the allocation from the client state. Taskrunners persist state during graceful shutdown, so it shouldn't be possible to leak tasks that have actually started. This lets us "start over" with the allocation, if the server still wants to place it on the client.

---

This work came out of discussions in https://github.com/hashicorp/nomad/issues/16623 where old state was kicking around and making log noise that was not useful to the user and making it harder to debug the real problem.

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -749,6 +749,10 @@ func TestClient_SaveRestoreState(t *testing.T) {
 		wait.Gap(time.Millisecond*30),
 	))
 
+	// Create a corrupted allocation that will be removed during restore
+	corruptAlloc := mock.Alloc()
+	c1.stateDB.PutAllocation(corruptAlloc)
+
 	t.Log("shutting down client")
 	must.NoError(t, c1.Shutdown()) // note: this saves the client state DB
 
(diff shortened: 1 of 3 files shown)


client/client_test.go
```diff
@@ -617,25 +619,32 @@ func TestClient_SaveRestoreState(t *testing.T) {
...
-	})
+	must.Wait(t, wait.InitialSuccess(
+		wait.ErrorFunc(func() error {
+			c2.allocLock.RLock()
+			defer c2.allocLock.RUnlock()
+
+			_, corruptStillLives := c2.allocs[corruptAlloc.ID]
+			if corruptStillLives {
+				return fmt.Errorf("corrupted allocation should not have been restored")
+			}
```
tgross, line 634: Note to reviewers: I've updated this test to the current idioms, but this is the meat of the change to ensure we handle the corrupted alloc state correctly.

shoenig (approved):
LGTM!

schmichael (approved):
Wow, great debugging. Good example of how ignoring that old compat code can cause issues. :grimacing:

schmichael:
Backport since it's a bug fix?

tgross:
> Backport since it's a bug fix?

I've updated the changelog to mark it as a bug. Will merge and run the backports once CI is green again

PR #16641, opened 2023-03-24
[testing] Set RequireRoot to be a test helper

angrycub:
Noticed that it was logging RequireRoot's line number and not the test.


shoenig (approved):
LGTM!

PR #16651, opened 2023-03-27
dev: modify Go install to support arm64 and non-vagrant machines.

jrasell:


scripts/linux-priv-go.sh
```diff
@@ -0,0 +1,67 @@
...
+
+install_go
+
+# Ensure that the GOPATH tree is owned by the correct user.
+sudo mkdir -p /opt/gopath
+sudo chown -R $USER:$USER /opt/gopath
+
+# Ensure Go is on PATH
+if [ ! -e /usr/bin/go ] ; then
+	sudo ln -s /usr/local/go/bin/go /usr/bin/go
```
shoenig, line 53: Seems like this should link to `/usr/local/bin/go` which conventionally appears first on PATH before `/usr/bin`. It would make a difference if someone were to install Go via the package manager and we want ours to take priority

shoenig (approved):
LGTM! just the one suggestion

tgross (approved):
LGTM

scripts/linux-priv-go.sh
jrasell: Changing this results in Go not being available, at least on my local VM and this is the configuration we had in the previous script. Git couldn't handle the file rename too well. I'd prefer to leave it as is for now, and follow up with any changes, if you're OK with that?

Issue #16640, opened 2023-03-24
Restoring a Nomad Snapshot into a fresh cluster with new clients breaks jobs in the cluster


megalan247:
### Nomad version
1.5.2 

### Operating system and Environment details
Ubuntu Linux (ARM64) running in AWS

### Issue
When restoring a snapshot of the nomad state, nomad also restores all the clients and where the allocs are placed. However, when using an AWS autoscaling group, often we have a situation where the "old" clients no longer exist. Therefore Nomad "hangs" in a situation where it thinks all the allocs are running, but they are not. 

The clients then after a long time (5 mins) eventually transition to a "down" state, however the allocs are never marked as "lost" and therefore are never rescheduled. This results in a broken cluster where none of the running jobs can be interacted with as no evaluations for them can be processed.

There are logs about 
```
nomad.keyring.replicator: failed to fetch key from any peer: key=9cc16249-4278-53c6-f0a7-50101efa705b error="rpc error: no such key \"9cc16249-4278-53c6-f0a7-50101efa705b\" in keyring"
```
After running a `nomad operator root keyring rotate -full` it appeared to "free" up the allocs and they were all rescheduled. 

### Reproduction steps
- Create two autoscaling groups in AWS, one for nomad servers, one for nomad clients. 
- Run some jobs and take a nomad snapshot. 
- Terminate all the nomad servers and clients at the same time in AWS, which will result in the servers being recreated, and a fresh cluster bootstrapped
- Restore the nomad snapshot to the cluster, and observe no rescheduling taking place.

#### Expected Result
On restore of the state, the jobs would be immediately marked as "lost" and rescheduled onto new nodes.

#### Actual Result
The allocs are never rescheduled and the cluster is broken

N/A

### Nomad Server logs (if appropriate)
Mar 24 16:47:05 s-i-085b661d4ea83f713 nomad[4235]:     2023-03-24T16:47:05.492Z [ERROR] nomad.keyring.replicator: failed to fetch key from any peer: rpc error: no such key "9cc16249-4278-53c6-f0a7-50101efa705b" in keyring: key=9cc16249-4278-53c6-f0a7-50101efa705b
[...]

Issue #14852, opened 2022-10-07
External key wrapping for keyring


schmichael:
Nomad 1.4.0 shipped with [Variables](https://developer.hashicorp.com/nomad/docs/concepts/variables) which are [stored encrypted with server-specific keys](https://developer.hashicorp.com/nomad/docs/operations/key-management).

The next step toward securing variables would be to allow the use of external root encryption keys (the key-encrypting-key used to decrypt the server-local data-encryption-key) so that they are never on the server either in memory or on disk.

Luckily HashiCorp has a library for this and we already use it: https://github.com/hashicorp/go-kms-wrapping :tada: 

Enabling the use of external KMSes (key management services) should be as easy as plumbing through the configuration options, testing, and documentation. Supporting Vault and major cloud KMSes being the primary targets for support.

## Security Model

The end result of this change is that servers would be safe from **offline** attacks to retrieve variables. For example in Nomad 1.4 if an attacker has `root` access on a server, even if that server is not running, the attacker can access the plaintext KEK (key-encryption-key), decrypt the DEK (data-encryption-key), and then decrypt all variables.

If Nomad used a KMS as proposed in this issue the KEK would not be present on the server regardless of whether the Nomad agent was running or not. The KEK would never leave the KMS so that the DEK could not be decrypted without compromising the KMS itself, and therefore no variables could be decrypted.

### Shortcomings

Nomad operators should still consider `root*` access to **online** Nomad servers as having full access to all Nomad Variables. There are a variety of ways `root` could access variables even when a KMS was in use:

1. `root` users can [reset ACL management tokens](https://developer.hashicorp.com/nomad/tutorials/access-control/access-control-bootstrap#re-bootstrap-acl-system) and management tokens have access to all variables.
2. `root` users have access to the running Nomad server agent's memory and therefore could extract the DEK.
3. Most cloud KMSes use intrinsic instance identity/credentials for auth and therefore obtaining `root` on the Nomad server instance may grant use of the KEK via the KMS.

I'm sure a motivated attacker can come up with even more ways to extract secrets from a live Nomad server agent, but the point is that while using a KMS would significantly protect variables from a wide variety of attacks, it is not a panacea. Specifically `Shortcoming #3` above is something Nomad operators would need to prevent as Nomad has no ability to deauth or unenroll itself from a KMS on server shutdown.

`*` or whatever user the Nomad server agent is running as. `root` is not required (or recommended) for Nomad server agents.

### Out of scope

[...]

PR #16663, opened 2023-03-27
docs: add notes about keyring to snapshot restore

tgross:
When cluster administrators restore from Raft snapshot, they also need to ensure the keyring is in place. For on-prem users doing in-place upgrades this is less of a concern but for typical cloud workflows where the whole host is replaced, it's an important warning (at least until #14852 has been implemented).

Ref https://github.com/hashicorp/nomad/issues/16640

### DIFF
--- a/website/content/docs/commands/operator/snapshot/restore.mdx
+++ b/website/content/docs/commands/operator/snapshot/restore.mdx
@@ -15,6 +15,9 @@ designed to handle server failures during a restore. This command is primarily
 intended to be used when recovering from a disaster, restoring into a fresh
 cluster of Nomad servers.
 
+This command only restores the Raft snapshot. If you are recovering a cluster,
+you will also need to [restore the keyring][] onto at least one server.
+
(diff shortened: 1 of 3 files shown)


website/content/docs/commands/operator/snapshot/restore.mdx
```diff
@@ -15,6 +15,9 @@ designed to handle server failures during a restore. This command is primarily
 intended to be used when recovering from a disaster, restoring into a fresh
 cluster of Nomad servers.
 
+This command only restores the Raft snapshot. If you are recovering a cluster,
+you will also need to [restore the keyring][] onto at least one server.
```
lgfa29, line 19:
```suggestion
~> **Warning:** This command only restores the Raft snapshot. If you are recovering a cluster,
you will also need to [restore the keyring][] onto at least one server.
```

> but for typical cloud workflows where the whole host is replaced, it's an important warning

This is a good point, and I haven't thought of this before. So maybe it's worth highlighting these more?

lgfa29 (approved):

PR #16665, opened 2023-03-27
[ui] Allow for multiple JWT auth methods in the UI

philrenaud:
In #16625 , we allow a user to sign in using a JWT. However, this only looks for the first `authMethod` of type JWT and assumes that's good enough.

This PR allows the user to have multiple JWTs and select from among them upon token input.

The select dropdown does not appear until the secret provided is "JWT-shaped", at which point it will give options if more than one JWT auth method exists.

## Side-effects:
- updated our jsconfig to let us do more type-safety and module interpretative type work. Your editor should give you good type hinting for any authMethods, etc. in this file and others, using in-comment type imports.

### Screenshots
3 JWT auth methods, but the token provided is not JWT: 
image

3 JWT auth methods, token provided is JWT-shaped:
image

User can modify the method upon form submission:
image

### DIFF
--- a/ui/app/controllers/settings/tokens.js
+++ b/ui/app/controllers/settings/tokens.js
@@ -1,6 +1,5 @@
 // @ts-check
 import { inject as service } from '@ember/service';
-import { reads } from '@ember/object/computed';
 import Controller from '@ember/controller';
 import { getOwner } from '@ember/application';
 import { alias } from '@ember/object/computed';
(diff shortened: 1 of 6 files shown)


pkazmierczak (approved):
Looks great!

ui/app/controllers/settings/tokens.js
```diff
@@ -45,8 +49,11 @@ export default class Tokens extends Controller {
...
 
+  /**
+   * @returns {import('@ember/array/mutable').default<import('../../models/auth-method').default>}
+   */
   get authMethods() {
-    return this.store.peekAll('auth-method');
+    return this.model?.authMethods || [];
   }
 
   get hasJWTAuthMethods() {
```
ChaiWithJai, line 59: `nit`: This should be named `nonJWTAuthMethods`. This returns a list of methods matching the criteria, not a boolean.

ChaiWithJai (approved):

ui/app/controllers/settings/tokens.js
philrenaud: Hey @ChaiWithJai , is this about `hasJWTAuthMethods` ? If so, the .any() type signature does in fact return a boolean ([docs](https://api.emberjs.com/ember/2.18/classes/EmberArray/methods/mapBy?anchor=any))

Issue #16656, opened 2023-03-27
Nomad does not respect the default behavior of unlimited retry when an error is returned from Vault `vault_retry {attempts=0}` 


margiran:
### Nomad version

Nomad v1.4.3+ent (bf2274db6b44e17c96d9b16361bcd77016708353)

### Operating system and Environment details

Ubuntu 20.04.6 LTS

### Issue
Nomad does not respect the default behavior of unlimited retry when an error is returned from Vault `vault_retry {attempts=0}` mentioned in the [documents](https://developer.hashicorp.com/nomad/docs/configuration/client#vault_retry).

### Reproduction steps

1. Setup a Vault server (needs a secret and policy which the nomad job can access)
2. Setup a Nomad Server 
3. A Nomad Clients with default configuration ( NOT setting the vault_retry { attempts=0 … } in configuration to observe the default behavior) 
4. Run the provided job ( a job that has a template to retrieve secrets from the vault) 
5. Shut down the vault server
6. Restarting the nomad service in the clients ( to make it try to update the template )
7. After 12 retries the Nomad client with the default config shuts the allocation down despite documented default behavior  which is  unlimited retry (attempts=0)

#### Expected Result
unlimited retry  when an error is returned from Vault ( vault_retry { attempts=0 })

#### Actual Result
After 12 retries the Nomad client with the default config shuts the allocation down

### Job file (if appropriate)
```
job "noretry" {
  datacenters = ["dc1"]

  group "webcheck" {
    network {
      mode = "bridge"
      port "http" {
        to = 80
      }
    }
    max_client_disconnect = "504h"
    count                 = 1
    task "website" {
      driver = "docker"
      config {
        image = "python:latest"
        ports = ["http"]
        command = "python3"
        args    = ["-m", "http.server", "80"]
      }
        vault {
            policies = ["access-tables"]
        }
        template {
            data = "{{ with secret \"test/data/test-data\"}}{{.Data.key1}}{{ end }}"
            destination = "local/test-data-bcpdr.txt"
        }
    }
  }
}
```

### Nomad Client logs (if appropriate)

```
2023-03-18T23:30:03.063+0100 [INFO]  agent: (runner) creating new runner (dry: false, once: false)
[...]

PR #16667, opened 2023-03-27
docs: fix template retry attempts default documentation

tgross:
Fixes https://github.com/hashicorp/nomad/issues/16656

The configuration docs for `client.template.vault_retry`, `consul_retry`, and `nomad_retry` incorrectly document the default number of attempts to be unlimited (0). When we added these config blocks, we defaulted the fields to `nil` for backwards compatibility, which causes them to fall back to the default consul-template configuration values.

### DIFF
--- a/website/content/docs/configuration/client.mdx
+++ b/website/content/docs/configuration/client.mdx
@@ -470,7 +470,7 @@ see the [drivers documentation](/nomad/docs/drivers).
   "blocking queries". A blocking query is used to wait for a potential change
   using long polling.
 
-- `consul_retry` `(map: { attempts = 0 backoff = "250ms" max_backoff = "1m" })`-
+- `consul_retry` `(map: { attempts = 12 backoff = "250ms" max_backoff = "1m" })`-
   This controls the retry behavior when an error is returned from Consul. The template
   runner will not exit in the face of failure. Instead, it uses exponential back-off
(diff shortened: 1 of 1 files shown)


lgfa29 (approved):
:sad-panda:

jorgemarey:
Hi  @tgross, happen to get to this as I was expecting unlimited retries on the template if vault fails (I saw the documentation before this was changed). Looking at the code I see [this](https://github.com/hashicorp/nomad/blob/0040427c6dce152d1fe20a1b0925141166abb6c9/client/config/config.go#L801) .I think it's not being used, but maybe we should remove that from the code to avoid future problems?

tgross:
Yeah IIRC the actual default value is coming from the consul-template library. An unfortunate bit of this config parsing is that Go's zero values are meaningful values here, so the config is a pointer to an int instead of an int. So I suspect the reason these are getting set in this method is to avoid nil pointer exceptions elsewhere in the parsing code (or maybe even in test code).

Seems worth hunting down though because it's quite confusing, especially with the consul-template default not being visible in the code base here in Nomad. Looks like the whole set of similar configs (ref [`config.go#L798-L806`](https://github.com/hashicorp/nomad/blob/0040427c6dce152d1fe20a1b0925141166abb6c9/client/config/config.go#L798-L806)) has the same problem.

Issue #16666, opened 2023-03-27
ui: Refactor for Variable Parsing changes


ChaiWithJai:
We will need to remove the Variables table in the read-only view because we cannot parse variables out of job specification file.
This will require us to also inform users that the job specification file they're reading will require them to check the CLI to learn about the current values of their variables.

PR #16668, opened 2023-03-27
ui:  validate job specification file for variables

ChaiWithJai:
This PR kicks off the refactoring work required for #16666.

This PR specifically creates a utility function that will parse a string representation of a  job specification file that has a `variable` declaration or reference.

### DIFF
--- a/ui/app/utils/has-variable-declarations.js
+++ b/ui/app/utils/has-variable-declarations.js
@@ -1,0 +1,6 @@
+export default function hasVariableDeclarationsAndReferences(spec) {
+  const VAR_DECLARATION_AND_REFERENCE_PATTERN = '\\b(?:variable|var\\.)\\b';
+  const expression = new RegExp(VAR_DECLARATION_AND_REFERENCE_PATTERN);
+
+  return expression.test(spec);
+}
--- a/ui/tests/unit/utils/has-variable-declarations-test.js
+++ b/ui/tests/unit/utils/has-variable-declarations-test.js
@@ -1,0 +1,39 @@
+import hasVariableDeclarationsAndReferences from 'nomad-ui/utils/has-variable-declarations';
+import { module, test } from 'qunit';
+
+module(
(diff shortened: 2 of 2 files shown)


philrenaud (approved):
If the has-variable-declarations util is only going to be used in tests, let's move it under tests/utils; otherwise, LGTM!

ui/app/utils/has-variable-declarations.js
```diff
@@ -0,0 +1,6 @@
+export default function hasVariableDeclarationsAndReferences(spec) {
+  const VAR_DECLARATION_AND_REFERENCE_PATTERN = '\\b(?:variable|var\\.)\\b';
```
lgfa29, line 2:
```suggestion
  const VAR_DECLARATION_AND_REFERENCE_PATTERN = /\b(?:variable[s]?|var\.)\b/;
```

Sorry, late to the party 😅 

There's a succinct way to declare variables that looks like this:

```hcl
variables {
  image = "redis:7"
  cpu   = 500
}
```

I think this regex will capture both? 

Using `/.../` also sidesteps the need to escape slashes, making the expression easier to read, and also compiles it automatically, so it's not necessary to call `new RegExp`. Since this pattern is static you can also declare the constant outside the function so it's computed only once instead of every time the function is called.

It's also worth keeping in mind that this can generate false positives, for example, a job like this would return `true` even though it doesn't use variables:

```hcl
job "example" {
  group "cache" {
    task "redis" {
      driver = "raw_exec"

      config {
        command = "/bin/bash"
        args    = ["${NOMAD_TASK_DIR}/script.sh"]
      }

      template {
        data = <<EOF
{{ $var := sprig_dict "a" "b" }}

while true; do
  echo "{{ $var.a }}"
  sleep 3
done
EOF

        destination = "${NOMAD_TASK_DIR}/script.sh"
      }
    }
  }
}
```

It's a pretty weird example, and you have is probably good enough depending on how this function is going to be used 👍

PR #16669, opened 2023-03-27
feat:  visualize HCL Job Specification in the Nomad UI `jobs.job.definition` view

### LINKED ISSUE #15982: ☂️ Nomad HCL in UI

ChaiWithJai:
Closes #15982


ui/app/styles/components/codemirror.scss
```diff
@@ -138,7 +138,7 @@ $dark-bright: lighten($dark, 15%);
 header.run-job-header {
   display: grid;
   grid-template-columns: 1fr auto;
-  margin-bottom: 2rem;
+  margin: 2rem 0rem;
```
ChaiWithJai, line 141: @philrenaud this is it.

juliezzhou (commented):
I tested this branch, and I think it meets the UX requirements from design side.

ui/app/templates/components/job-editor/read.hbs
```diff
@@ -0,0 +1,78 @@
...
+                    content=@data.job._newDefinitionVariables
+                    mode="ruby"
+                    theme="hashi-read-only"
+                    readOnly=true
+                }}
+            />
+        </div>
+    </div>
+    {{/if}}
+</div>
```
philrenaud, line 78: Is @data.hasVariables working as intended? I'm getting the following display: 
image

I have the `All HCL variable values Not Shown` warning at the top of the editor, but shouldn't we not show this empty editor if we have nothing to show within it?

ui/app/templates/components/job-editor/alert.hbs
```diff
@@ -0,0 +1,29 @@
...
+          <A.Description data-test-error-message>{{@data.error.message}}</A.Description>
+      </Hds::Alert>
+    {{/if}}
+    {{#if (and (eq @data.stage "read") @data.hasVariables (eq @data.view "job-spec"))}}
+      {{#if this.shouldShowAlert}}
+        <Hds::Alert @type="inline" @onDismiss={{this.dismissAlert}} data-test-variable-notification as |A|>
+          <A.Title>All HCL variable values Not Shown</A.Title>
+          <A.Description>All HCL variable values provided to this job version may not be shown. Ensure HCL variable values are properly set when submitting the job.</A.Description>
+        </Hds::Alert>
+      {{/if}}
```
philrenaud, line 14: Is this just always on?  "Ensure that..." in an alert reads to me like I've made some kind of mistake that I could correct in the future, in order to not see it.

Noting this because it seems like `shouldShowAlert` is set to true, not dependent on anything else other than a user dismissing it. Is there something I'm missing here?

image

What impact are we hoping to have with the user in showing them this?

philrenaud: A second thought, but even if we **did** intend to show this to a user on all jobs, could we possibly make it localStorage dismissable? If the goal is to show it to all users, but it isn't actually impacted by the job in question, could we make it a `@localStorageProperty` so that once dismissed, it stays dismissed (in the user's current browser)

ui/tests/utils/generate-raw-json-job.js
```diff
@@ -0,0 +1,549 @@
...
+  Specification: {
+    Definition:
+      'job "docs" {\n  namespace = "madness"\n  group "example" {\n    task "server" {\n      service {\n        tags = ["leader", "mysql"]\n\n        port = "db"\n\n        meta {\n          meta = "for your service"\n        }\n\n        check {\n          type     = "tcp"\n          port     = "db"\n          interval = "10s"\n          timeout  = "2s"\n        }\n\n        check {\n          type     = "script"\n          name     = "check_table"\n          command  = "/usr/local/bin/check_mysql_table_status"\n          args     = ["--verbose"]\n          interval = "60s"\n          timeout  = "5s"\n\n          check_restart {\n            limit = 3\n            grace = "90s"\n            ignore_warnings = false\n          }\n        }\n      }\n    }\n  }\n}\n',
+    Type: 'hcl',
+    Variables: {
+      datacenters: ['west'],
+      external_port: 4000,
+    },
+  },
+};
```
philrenaud, line 549: I'd rather this be generated in Mirage rather than have its JSON pasted verbatim. If we want to vary this test and expand on it later, it would make it easier to do if we could modify it at mirage server.schema level.

philrenaud (commented):
Code looks good, but I have a couple product/ui questions I wanted to raise.

I'm also generally unable to get "Edit HCL Variable Values" to have any content in it, even on jobs like this:

```
variable "datacenters" {
  type        = list(string)
  description = "List of Nomad datacenters to run the job in. Defaults to `[\"dc1\"]`"
  default     = ["dc1"]
}

variable "docker_image" {
  type        = string
  description = "Docker image for the job to run"
  default     = "redis"
}

variable "image_version" {
  type        = string
  description = "Version of the docker image to run"
  default     = 3
}

job "job1" {
  datacenters = var.datacenters

  group "job1" {
    task "job1" {
      driver = "docker"

      config {
        image = "${var.docker_image}:${var.image_version}"
      }
    }
  }
}
```

@ChaiWithJai 's let me know that this is because the variables are not externally set, but instead are the defaults as set in the job. The UX of this doesn't seem right to me; I'd expect to see them listed alongside any manually-set ones below. cc @mikenomitch

philrenaud (commented):
Code looks good, but I have a couple product/ui questions I wanted to raise.

I'm also generally unable to get "Edit HCL Variable Values" to have any content in it, even on jobs like this:

```
variable "datacenters" {
  type        = list(string)
  description = "List of Nomad datacenters to run the job in. Defaults to `[\"dc1\"]`"
  default     = ["dc1"]
}

variable "docker_image" {
  type        = string
  description = "Docker image for the job to run"
  default     = "redis"
}

variable "image_version" {
  type        = string
  description = "Version of the docker image to run"
  default     = 3
}

job "job1" {
  datacenters = var.datacenters

  group "job1" {
    task "job1" {
      driver = "docker"

      config {
        image = "${var.docker_image}:${var.image_version}"
      }
    }
  }
}
```

@ChaiWithJai 's let me know that this is because the variables are not externally set, but instead are the defaults as set in the job. The UX of this doesn't seem right to me; I'd expect to see them listed alongside any manually-set ones below. cc @mikenomitch

ui/app/templates/components/job-editor/read.hbs
ChaiWithJai: Can you provide the job specification file so I can reproduce?

ui/app/components/job-editor.js
```diff
@@ -24,6 +24,15 @@ export default class JobEditor extends Component {
     if (this.definition) {
       this.setDefinitionOnModel();
     }
+
+    if (this.args.variables) {
+      this.args.job.set(
+        '_newDefinitionVariables',
+        this.jsonToHcl(this.args.variables.flags).concat(
+          this.args.variables.literal
```
ChaiWithJai, line 32: `nit`: reverse the order. There is a chance for dupes and we don't care about dupes showing, but we'd rather prioritize the literal over the flags.

ui/tests/utils/generate-raw-json-job.js
ChaiWithJai: Deleted my previous comment. The reason that this file is here is because of `fetchRawRequest`. This is what hydrates `CodeMirror` previously we never tested how this API request was responsible for hydrating this view to ensure that its correct. Take a look into the acceptance test for `Job Definition` to see how this is being used.

The entire purpose of `fetchRawRequest` is that its dispatching a request using an adapter, however its side stepping the Ember Data store. Its the same request URL as fetching a job, however, we don't want to serialize the API response and thus we must override the default behavior of Mirage. To accomplish this, we use a passthrough recipe where we call a `pretender` method instead of a Mirage request to ensure that we get a proper `JSON` response.

This is Mirage/Pretender pattern and this is how we handle the evaluation response in the `Evaluations` view.

I believe this is the correct pattern for the acceptance test.

ui/tests/acceptance/job-definition-test.js
```diff
@@ -130,3 +130,48 @@ module('Acceptance | job definition', function (hooks) {
...
+    assert.equal(
+      codeMirror.getValue(),
+      specification_response.Source,
+      'Shows the full definition as written by the user'
+    );
+
+    await click('[data-test-select-full]');
+    codeMirror = getCodeMirrorInstance('[data-test-editor]');
+    assert.propContains(JSON.parse(codeMirror.getValue()), JOB_JSON);
+  });
```
philrenaud, line 176: Suggestion: add a `percySnapshot` here for a little more reassurance down the road! I can see us making changes to our editor over time and this test would be a nice canary in the coal mine.

philrenaud (approved):
Feels and looks great! Nice work.

ui/tests/acceptance/job-definition-test.js
ChaiWithJai: We're expecting a visual regression where we may change the order of concatenation, as well as, the removal of an alert because of API changes. It might make sense to wait until this issue is resolved and then tackle adding Percy snapshots into another PR. I'm filing a [ticket](https://github.com/hashicorp/nomad/issues/17129) for this.

PR #16691, opened 2023-03-28
func: add warnings function to Service definition

Juanadelacuesta:


nomad/structs/structs.go
```diff
@@ -7036,6 +7036,13 @@ func (tg *TaskGroup) Warnings(j *Job) error {
 		}
 	}
 
+	for _, s := range tg.Services {
+		if err := s.Warnings(); err != nil {
+			err = multierror.Prefix(err, fmt.Sprintf("Service :%q on TaksGroup: %q", s.Name, tg.Name))
```
tgross, line 7041:
```suggestion
			err = multierror.Prefix(err, fmt.Sprintf("Service: %q on TaskGroup: %q", s.Name, tg.Name))
```

tgross (commented):
The code LGTM but I have two comments in terms of our intent:

* The wording of the warning suggests that it's an error.
* Do we really want to make this change? I realize that having consistency with the Consul services would be nice, but having a stricter validation for our own services means we could more easily add DNS support for Nomad SD without worrying about backwards compatibility later on.

Issue #16692, opened 2023-03-28
docker driver `allow_caps` and `cap_add` don't work as documented (fails closed)


tgross:
While testing https://github.com/hashicorp/nomad/pull/16643 which fixes the behavior of `cap_add` for the `exec` driver, I discovered that the `docker` driver has the same bug. This is because upstream Docker has the same bug (see https://github.com/moby/moby/issues/38664 https://github.com/moby/moby/issues/8460) but has intentionally decided not to fix it. This results in unexpected behavior in the `docker` task driver as well.

**Note: this bug "fails closed". It does not open capabilities unexpectedly, so it's a security bug.**

tl;dr:
* [`allow_caps`](https://developer.hashicorp.com/nomad/docs/drivers/docker#allow_caps) can only reduce, not expand, the permitted set of capabilities.
* [`cap_add`](https://developer.hashicorp.com/nomad/docs/drivers/docker#cap_add) can only remove selected capabilities from the `cap_drop` set (making them allowed again).

Unfortunately there's no workaround for this issue short of getting Docker to fix it upstream, which they've previously rejected. So I'm going to update the documentation to note the current behavior and point to this issue.

---

To verify this behavior, I used the following jobspec. Note that in order to properly test `net_bind_service` you need to set the `sysctl net.ipv4.ip_unprivileged_port_start 1024`. This is because upstream Docker has worked around this whole thing for the common case of binding to low ports by setting the default `sysctl net.ipv4.ip_unprivileged_port_start 0`.

<details><summary>job spec</summary>

```hcl
job "example" {

  group "group" {

    task "task" {
      driver = "docker"
      # user   = "nobody"

      config {
        image    = "busybox:1"
        # cap_add  = ["net_bind_service"]
        # cap_drop = ["all"]

        command  = "/bin/sleep"
        args = ["600"]
        sysctl = {
          "net.ipv4.ip_unprivileged_port_start" = "1024"
        }
      }
    }
  }
}
```

</details>

I tested this jobspec along several dimensions: user namespace remapping, the `allow_caps` field on the plugin, the `user` field on the task, and the `cap_add`/`cap_drop` field. For each test I ran the job, `alloc exec`'d into the allocation (if it ran at all), and ran two tests for the netbind and netraw capabilities:

```
/ $ nc -lvp 443
nc: bind: Permission denied
```

```
/ $ ping 8.8.8.8
PING 8.8.8.8 (8.8.8.8): 56 data bytes
[...]

PR #16693, opened 2023-03-28
docs: clarify capabilities options for `docker` driver

tgross:
Closes https://github.com/hashicorp/nomad/issues/16692

The `docker` driver cannot expand capabilities beyond the default set when the task is a non-root user. Clarify this in the documentation of `allow_caps` and update the `cap_add` and `cap_drop` to match the `exec` driver, which has more clear language overall.

### DIFF
--- a/website/content/docs/drivers/docker.mdx
+++ b/website/content/docs/drivers/docker.mdx
@@ -504,11 +504,11 @@ config {
   }
   ```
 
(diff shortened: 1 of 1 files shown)


lgfa29 (approved):
Would it make sense to fingerprint the client's capabilities? It seems like this would fail silently so being able to add a `constraint` when a specific `cap_add` is needed could be useful.

PR #16715, opened 2023-03-29
refact:  uniform alerts

ChaiWithJai:
The PR is pre-work for #16714 to create some uniformity in alerts on the `JobEditor` component before adding a brand new alert.

Loom Walkthrough:  https://www.loom.com/share/ac82a522ba40423ab5c625799153a077

### DIFF
--- a/ui/app/routes/jobs/job/definition.js
+++ b/ui/app/routes/jobs/job/definition.js
@@ -31,6 +31,12 @@ export default class DefinitionRoute extends Route {
 
   setupController(controller, model) {
     super.setupController(controller, model);
-    controller.view = model?.specification ? 'job-spec' : 'full-definition';
+
+    const view = controller.view
+      ? controller.view
+      : model?.specification
+      ? 'job-spec'
+      : 'full-definition';
+    controller.view = view;
   }
 }
--- a/ui/app/routes/jobs/job/definition.js
+++ b/ui/app/routes/jobs/job/definition.js
@@ -31,6 +31,6 @@ export default class DefinitionRoute extends Route {
 
   setupController(controller, model) {
     super.setupController(controller, model);
-    controller.view = model.specification ? 'job-spec' : 'full-definition';
+    controller.view = model?.specification ? 'job-spec' : 'full-definition';
   }
 }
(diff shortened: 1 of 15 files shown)


ui/app/routes/jobs/job/definition.js
```diff
@@ -31,6 +31,6 @@ export default class DefinitionRoute extends Route {
 
   setupController(controller, model) {
     super.setupController(controller, model);
-    controller.view = model.specification ? 'job-spec' : 'full-definition';
+    controller.view = model?.specification ? 'job-spec' : 'full-definition';
```
philrenaud, line 34: Mentioned elsewhere but: this seems like something that might be better as a defaultValue / constructor in the definition controller, rather than the route.

ui/app/templates/components/job-editor.hbs
```diff
@@ -1,16 +1,7 @@
...
-    </div>
-  {{/if}}
-  {{#if this.data.hasVariables}}
-    <div data-test-variable-notification class="notification">
-      <h3 class="title is-4">Job Variable Values Not Shown</h3>
-      <p>Please refer to Nomad CLI to see the current value of variables.</p>
-    </div>
-  {{/if}}
+  <JobEditor::Alert 
+    @data={{merge (hash error=this.error stage=this.stage) this.data}}
```
philrenaud, line 3: TIL Ember has `merge`

ui/app/templates/components/job-editor/alert.hbs
```diff
@@ -0,0 +1,21 @@
+  {{#if @data.error}}
+    <Hds::Alert @type="inline" @color="critical" data-test-error={{@data.error.type}} as |A|>
+        <A.Title data-test-error-title>{{conditionally-capitalize @data.error.type true}}</A.Title>
```
philrenaud, line 3: I don't feel great about handling this in JS, especially when the condition here is always true.

I'd suggest setting up a style (feel free to even make a new job-editor-alert.scss) that amounts to `.job-editor-alert .title { text-transform: capitalize; }`

philrenaud (approved):

ui/app/routes/jobs/job/definition.js
ChaiWithJai: The challenge here is that the `model` is not available when the `Controller` is instantiated. Therefore using the `constructor` method of the `Controller` isn't the appropriate place. If we want `model` access, the "Ember way" would be to use `setupController`.

ui/app/templates/components/job-editor.hbs
ChaiWithJai: I created that!

PR #16722, opened 2023-03-29
CSI: set mounts in alloc hook resources atomically

tgross:
Fixes https://github.com/hashicorp/nomad/issues/16623

The allocrunner has a facility for passing data written by allocrunner hooks to taskrunner hooks. Currently the only consumers of this facility are the allocrunner CSI hook (which writes data) and the taskrunner volume hook (which reads that same data).

The allocrunner hook for CSI volumes doesn't set the alloc hook resources atomically. Instead, it gets the current resources and then writes a new version back. Because the CSI hook is currently the only writer and all readers happen long afterwards, this should be safe but #16623 shows there's some sequence of events during restore where this breaks down.

Refactor hook resources so that hook data is accessed via setters and getters that hold the mutex, and ensure the object is instantiated synchronously at the time the `AllocRunner` is created.

---

Note to reviewers: the reproduction for the crash is extremely complicated and timing dependent, so we're not going to be able to test this in a unit test. See my comment at https://github.com/hashicorp/nomad/issues/16623#issuecomment-1492415089 for how this has been manually tested. There's some follow-up work we're going to need to do in https://github.com/hashicorp/nomad/issues/16746

### DIFF
--- a/client/structs/allochook.go
+++ b/client/structs/allochook.go
@@ -4,6 +4,7 @@ import (
 	"sync"
 
 	"github.com/hashicorp/nomad/client/pluginmanager/csimanager"
+	"github.com/hashicorp/nomad/helper"
 )
 
 // AllocHookResources contains data that is provided by AllocRunner Hooks for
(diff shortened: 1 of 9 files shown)


client/structs/allochook.go
```diff
@@ -7,23 +7,31 @@ import (
...
+	}
+}
+
 func (a *AllocHookResources) GetCSIMounts() map[string]*csimanager.MountInfo {
 	a.mu.RLock()
 	defer a.mu.RUnlock()
 
-	return a.CSIMounts
+	return a.csiMounts
 }
```
shoenig, line 30: What's the policy on modifying this map? Maybe add a comment here on what is expected of callers, or return a copy.

shoenig (approved):
LGTM! much easier to follow.

client/structs/allochook.go
tgross: Good call. It's probably safe to "guard" this with a comment because the volume hook is just reading that data and not writing it, but it's way to easy to break later unexpectedly. Returning a copy in a9b2ac844093863c731540caf57cb6cc8d7626fd

tgross:
This will go out in the next regular patch release.

PR #16723, opened 2023-03-30
fix host port handling for ipv6, fixes #16550

### LINKED ISSUE #16550: NOMAD_ADDR_* variables incorrect with ipv6
### Env
nomad 1.5.1, linux node with only ipv6 address

### Issue

Nomad env vars have invalid format:
```
NOMAD_HOST_ADDR_http=2a01:4f8:111:1ae1::2:25013
NOMAD_ADDR_http=2a01:4f8:111:1ae1::2:25013
```
#### Expected Result

They should be like this:

```
NOMAD_HOST_ADDR_http=[2a01:4f8:111:1ae1::2]:25013
NOMAD_ADDR_http=[2a01:4f8:111:1ae1::2]:25013
```

#### Related issues

No address for services with IPv6 address in UI:
image

valodzka:
Closes #16550

### DIFF
--- a/command/agent/agent_endpoint_test.go
+++ b/command/agent/agent_endpoint_test.go
@@ -15,6 +15,7 @@ import (
 	"net/url"
 	"os"
 	"reflect"
+	"strconv"
 	"strings"
 	"sync"
 	"syscall"
@@ -158,7 +159,7 @@ func TestHTTP_AgentJoin(t *testing.T) {
 	httpTest(t, nil, func(s *TestAgent) {
 		// Determine the join address
 		member := s.Agent.Server().LocalMember()
-		addr := net.JoinHostPort("%s:%d", member.Addr, member.Port)
+		addr := net.JoinHostPort(string(member.Addr), strconv.Itoa(int(member.Port)))
 
 		// Make the HTTP request
 		req, err := http.NewRequest("PUT",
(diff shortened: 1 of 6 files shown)


command/agent/agent_endpoint_test.go
```diff
@@ -155,7 +155,7 @@ func TestHTTP_AgentJoin(t *testing.T) {
 	httpTest(t, nil, func(s *TestAgent) {
 		// Determine the join address
 		member := s.Agent.Server().LocalMember()
-		addr := fmt.Sprintf("%s:%d", member.Addr, member.Port)
+		addr := net.JoinHostPort("%s:%d", member.Addr, member.Port)
```
lgfa29, line 158:
```suggestion
		addr := net.JoinHostPort(member.Addr, member.Port)
```

lgfa29 (changes requested):
Thanks for the PR @valodzka!

Do you think you would be able to add some tests? I think something like [`TestEnvironment_AsList`](https://github.com/hashicorp/nomad/blob/3f7143f50b64afeb1e461295bde3d4f8f1077b66/client/taskenv/env_test.go#L155-L250) but using IPv6 addresses would do the trick. 

Feel free to reach out if you need any help 🙂

valodzka:
> Do you think you would be able to add some tests? 

@lgfa29  I'm not very familiar with go but here is my try :)

lgfa29 (approved):
Thanks for the additional tests @valodzka, you're practically a Go pro at this point 😄 

I pushed a couple of commits to add a changelog entry and also to fix a tiny problem in one of the test.

Issue #11097, opened 2021-08-30
Nomad service name validation stricter than Consul validation


craigday:
### Nomad version
Nomad v1.1.3 (8c0c8140997329136971e66e4c2337dfcf932692)

### Operating system and Environment details
Linux ip-172-25-216-30.ap-southeast-2.compute.internal 3.10.0-1160.36.2.el7.x86_64 #1 SMP Wed Jul 21 11:57:15 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

### Issue
The validation of service name when registering services in Consul via the service{} block is stricter than that of Consul. Consul allows registering service names that don't comply with RFC1123, since that restriction is purely to satisfy DNS based service discovery, and not everyone uses Consul DNS.

Example Error:

Service name must be valid per RFC 1123 and can contain only alphanumeric characters or dashes: "com.daml.ledger.api.v1.ActiveContractsService"

### Reproduction steps

      service {
        name = "com.daml.ledger.api.v1.ActiveContractsService"
        port = "service"
        address_mode = "host"
        check {
          type = "tcp"
          address_mode = "host"
          port = "service"
          interval = "10s"
          timeout = "5s"
        }
     }

#### Expected Result

The service is registered with Consul.

#### Actual Result

Service name must be valid per RFC 1123 and can contain only alphanumeric characters or dashes: "com.daml.ledger.api.v1.ActiveContractsService"

lgfa29:
Thanks for the report @craigday. Consul's [documentation](https://www.consul.io/api-docs/catalog#service) does mention that valid DNS labels are _recommended_, but not required.

> We recommend using valid DNS labels for service definition names for compatibility with external DNS. 

Testing a manual service registration also works in Consul
```console
$ consul catalog services
consul
web.1
```

Juanadelacuesta:
After going through some possible solutions for this issue, it was decided to maintain the current validation in place because it will make a future transition to an external DNS compatible with all currently existing services and also, even though Consul does not enforce any type of validation, Consul templates does.

PR #16729, opened 2023-03-30
[Docs] Update comments on service name validation

Juanadelacuesta:
This PR addresses the bug reported on https://github.com/hashicorp/nomad/issues/11097

After some consideration, it was decided that it was better to keep the validation in place:

- If in the future external DNS is used, putting the restriction back wont be backwards compatible.
- consul templates also has a similar validation.


tgross (approved):
LGTM :+1:

We should probably leave a note in #11097 explaining our reasoning here when we close it.

lgfa29 (approved):

PR #16730, opened 2023-03-30
ci: use `BACKPORT_MERGE_COMMIT` option

lgfa29:
Instead of attempting to pick each individual commit in a PR using `BACKPORT_MERGE_COMMIT` only picks the commit that was merged into `main`.

This reduces the amount of work done during a backport, generating cleaner merges and avoiding potential issues on specific commits.

With this setting PRs that are not squashed will fail to backport and must be handled manually, but those are considered exceptions.


tgross (approved):
LGTM :+1:

shoenig (approved):
:rocket:

PR #16734, opened 2023-03-30
ui:  enable editing of HCL vars in the UI

ChaiWithJai:
Loom Walk Through:  https://www.loom.com/share/e44928734f314deca5840c247f249eaf

### DIFF
--- a/ui/app/components/job-editor.js
+++ b/ui/app/components/job-editor.js
@@ -110,9 +110,13 @@ export default class JobEditor extends Component {
   }
 
   @action
-  updateCode(value) {
+  updateCode(value, type = 'job') {
     if (!this.args.job.isDestroying && !this.args.job.isDestroyed) {
-      this.args.job.set('_newDefinition', value);
+      if (type === 'hclVars') {
+        this.args.job.set('_newDefinitionVariables', value);
+      } else {
+        this.args.job.set('_newDefinition', value);
+      }
     }
   }
 
--- a/ui/app/routes/jobs/job/definition.js
+++ b/ui/app/routes/jobs/job/definition.js
@@ -31,7 +31,12 @@ export default class DefinitionRoute extends Route {
 
   setupController(controller, model) {
     super.setupController(controller, model);
-    controller.view =
-      model?.specification && !controller.view ? 'job-spec' : 'full-definition';
+
+    const view = controller.view
+      ? controller.view
+      : model?.specification
+      ? 'job-spec'
+      : 'full-definition';
+    controller.view = view;
   }
 }
--- a/ui/app/templates/components/job-editor/edit.hbs
+++ b/ui/app/templates/components/job-editor/edit.hbs
@@ -13,17 +13,32 @@
     {{/if}}
     </div>
     <div class="boxed-section-body is-full-bleed">
-    <div
-        data-test-editor
-        {{code-mirror
-        screenReaderLabel="Job definition"
-        content=@data.job._newDefinition
-        theme="hashi"
-        onUpdate=@fns.onUpdate
-        mode="javascript"
-        }}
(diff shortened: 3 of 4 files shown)


ui/app/components/job-editor.js
```diff
@@ -110,9 +110,13 @@ export default class JobEditor extends Component {
   }
 
   @action
-  updateCode(value) {
+  updateCode(value, type = 'job') {
```
philrenaud, line 113: This would be a great spot to include some jsdoc comments, if type can only ever equal `job` or `hclVars`!

```
  /**
   * @param {string} value
   * @param {("job" | "hclVars")} job
   */
```

or something like that

ui/app/routes/jobs/job/definition.js
```diff
@@ -31,7 +31,12 @@ export default class DefinitionRoute extends Route {
...
     super.setupController(controller, model);
-    controller.view =
-      model?.specification && !controller.view ? 'job-spec' : 'full-definition';
+
+    const view = controller.view
+      ? controller.view
+      : model?.specification
+      ? 'job-spec'
+      : 'full-definition';
+    controller.view = view;
```
philrenaud, line 40: This makes me want to further handle these as `@tracked` values with defaults and possibly a constructor in the definition controller. I'd be worried as a future editor that I'd miss where this logic lives.

philrenaud (approved):

ui/app/routes/jobs/job/definition.js
ChaiWithJai: The query parameter is a `tracked` property! However, for instantiating the value on constructing the controller, this logic must be handled in the `setupController` hook because the `model` isn't available.

philrenaud: There are a few things using the model in the controller currently — we're doing

```
@alias('model.job') job;
@alias('model.definition') definition;
```
there for example. 

Is there something that prevents `specification` from being referenced?

ui/app/templates/components/job-editor/edit.hbs
```diff
@@ -13,17 +13,32 @@
...
+        />
+    </div>
+    {{#if (eq @data.view "job-spec")}}
+    <div>
+        <h1 style="font-size: 2rem; font-weight: 600">Edit HCL Variables</h1>
+        <div
+            data-test-variable-editor
+            {{code-mirror
+            screenReaderLabel="HLC Variables for Job Spec"
+            content=""
```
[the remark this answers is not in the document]
ChaiWithJai: This is a brilliant comment. No, it won't. This will instead, take us into `review` stage of the component on success. Unless, there is a `parse` failure. As of now, we do not have a variable parser and it won't be in scope for the project, so this may always start as blank.

It may benefit us to create a collection of API requests/responses to inform us about the nuances of the user experience here.

ui/app/routes/jobs/job/definition.js
ChaiWithJai: This brings up a discussion that I wish Ember had a "cookbook" for. I had these same thoughts running the process and tried implementing almost exactly what you're talking about. Specifically I wanted a recipe for "how do I initialize a query parameter?".

If we're initializing query parameters: we need a resolved model.

The `setupController` hook is called after the model hook resolves, so you have access to both the resolved model and any query parameters at this stage.

We also need to support one more thing:

Responding to query parameter changes: If you need to perform some action whenever a query parameter changes, you can use the `setupController` hook in combination with queryParams configuration to respond to those changes.

Another area when we'd want to use this would be setting up properties on the controller based on query parameters: when you want to set up properties on the controller that depend on the current value of the query parameters (as I normally refer to as "derived state".

PR #16738, opened 2023-03-30
ent: move all license info into LicenseConfig{}

gulducat:
and add new `TestConfigForServer()` to get a valid `nomad.Config` to use in tests.

Should be noop here in OSS.

Relates to hashicorp/nomad-enterprise#855
These OSS files originated in ent PR hashicorp/nomad-enterprise#1060

### DIFF
--- a/nomad/config.go
+++ b/nomad/config.go
@@ -127,9 +127,6 @@ type Config struct {
 	// operators track which versions are actively deployed
 	Build string
 
-	// BuildDate is the time of the git commit used to build the program.
-	BuildDate time.Time
-
 	// Revision is a string that carries the version.GitCommit of Nomad that
 	// was compiled.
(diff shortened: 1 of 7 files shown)


nomad/config.go
```diff
@@ -389,10 +386,8 @@ type Config struct {
 	// connections from a single IP address. nil/0 means no limit.
 	RPCMaxConnsPerClient int
 
-	// LicenseConfig is a tunable knob for enterprise license testing.
+	// LicenseConfig stored information about the Enterprise license loaded for the server.
```
lgfa29, line 389:
```suggestion
	// LicenseConfig stores information about the Enterprise license loaded for the server.
```

I don't know how to type 🤦

lgfa29 (approved):

PR #16743, opened 2023-03-31
[api] Return a shapely error for unexpected response

angrycub:
**Motivation**
Currently the majority of the Nomad Go API methods use the `api.Client`'s generic methods like query, putQuery, etc. which wrap the response with a `requireOK` to report unexpected status errors. However, this results in 404's also appearing as unexpected responses. Additionally, each function that handled unexpected response errors implemented their own error string creation (even though it was identical behavior) risking drift between implementations.

This PR seeks to do the following:

- **Create a structured error type** - In considering feedback to #16437, I wanted to be able to group the warnings by unexpected response error type, which would have depended on brittle string matches to implement. Providing an error that can be cast to a structured error would reduce the fragility around screen scraping and make the proposed warning code mode resilient.

- **Use a common error pathway** - Prevent risk of drift by using the same implementation

Some considerations in this PR:

- A main goal was to maintain the structure of the current error message. This is the reason that go errors stored in the UnexpectedResponseError are not printed out in the error string—it would have changed `api.websocket()`'s error text. However, rather than being dropped, it is now retained inside of the UnexpectedResponseError and can be consulted if necessary.

- If a go error occurs that prevents normal reading of the response.Body it is recorded in `err`; `HasError()` => `true` and `Error()` => `err`.

- Any errors that occur _as a consequence of creating the error_ are recorded in `additional`; `HasAdditional()` => `true` and  `Additional() => additional`.

- UnexpectedResponseErrors that have `additional` set add the following text when their String() func is called. This additional information in the error string is helpful, since the highest likelihood of encountering an error while making an `UnexpectedResponseError` using `FromHTTPResponse()` is during the `io.Copy` call.
  ```
  Additionally, an error occurred while constructing this error (%s); the body might be truncated or missing.
  ```

### DIFF
--- a/api/error_unexpected_response.go
+++ b/api/error_unexpected_response.go
@@ -7,6 +7,8 @@ import (
 	"net/http"
 	"strings"
 	"time"
+
+	"golang.org/x/exp/slices"
 )
 
 // UnexpectedResponseError tracks the components for API errors encountered when
@@ -33,16 +35,16 @@ func (e UnexpectedResponseError) Unwrap() error             { return e.err }
 func (e UnexpectedResponseError) HasAdditional() bool       { return e.additional != nil }
 func (e UnexpectedResponseError) Additional() error         { return e.additional }
 func NewUnexpectedResponseError(src UnexpectedResponseErrorSource, opts ...UnexpectedResponseErrorOption) UnexpectedResponseError {
-	new := src()
+	nErr := src()
 	for _, opt := range opts {
-		opt(new)
+		opt(nErr)
 	}
-	if new.statusText == "" {
+	if nErr.statusText == "" {
 		// the stdlib's http.StatusText function is a good place to start
-		new.statusFromCode(http.StatusText)
+		nErr.statusFromCode(http.StatusText)
 	}
 
-	return *new
+	return *nErr
 }
 
 // Use textual representation of the given integer code. Called when status text
@@ -102,10 +104,7 @@ func WithStatusText(st string) UnexpectedResponseErrorOption {
 // expected to receive. This can be used by API callers to provide more feedback
 // to end-users.
 func WithExpectedStatuses(s []int) UnexpectedResponseErrorOption {
-	return func(u *UnexpectedResponseError) {
-		u.expected = make([]int, len(s))
-		copy(u.expected, s)
-	}
+	return func(u *UnexpectedResponseError) { u.expected = slices.Clone(s) }
 }
 
 // UnexpectedResponseErrorSource provides the basis for a NewUnexpectedResponseError.
@@ -137,9 +136,9 @@ func FromHTTPResponse(resp *http.Response) UnexpectedResponseErrorSource {
 	}
 }
 
-// FromStatusCode is the "thinnest" source for an UnexpectedResultError. It
-// will attempt to resolve the status code to status text using a resolving
-// function provided inside of the NewUnexpectedResponseError implementation.
+// FromStatusCode attempts to resolve the status code to status text using
+// the resolving function provided inside of the NewUnexpectedResponseError
+// implementation.
 func FromStatusCode(sc int) UnexpectedResponseErrorSource {
 	return func() *UnexpectedResponseError { return &UnexpectedResponseError{statusCode: sc} }
 }
--- a/api/error_unexpected_response.go
+++ b/api/error_unexpected_response.go
@@ -1,0 +1,164 @@
+package api
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+)
+
+// UnexpectedResponseError tracks the components for API errors encountered when
+// requireOK and requireStatusIn's conditions are not met.
+type UnexpectedResponseError struct {
+	expected   []int
+	statusCode int
+	statusText string
+	body       string
+	err        error
+}
+
+func (e UnexpectedResponseError) HasExpectedStatuses() bool { return len(e.expected) > 0 }
+func (e UnexpectedResponseError) ExpectedStatuses() []int   { return e.expected }
+func (e UnexpectedResponseError) HasStatusCode() bool       { return e.statusCode != 0 }
+func (e UnexpectedResponseError) StatusCode() int           { return e.statusCode }
+func (e UnexpectedResponseError) HasStatusText() bool       { return e.statusText != "" }
+func (e UnexpectedResponseError) StatusText() string        { return e.statusText }
+func (e UnexpectedResponseError) HasBody() bool             { return e.body != "" }
+func (e UnexpectedResponseError) Body() string              { return e.body }
+func (e UnexpectedResponseError) HasError() bool            { return e.err != nil }
+func (e UnexpectedResponseError) Unwrap() error             { return e.err }
+
+func NewUnexpectedResponseError(src UnexpectedResponseErrorSource, opts ...UnexpectedResponseErrorOption) UnexpectedResponseError {
+	new := src()
+	for _, opt := range opts {
+		opt(new)
+	}
+	if new.statusText == "" {
+		// the stdlib's http.StatusText function is a good place to start
+		new.statusFromCode(http.StatusText)
+	}
+
+	return *new
+}
+
+// Use textual representation of the given integer code. Called when status text
+// is not set using the WithStatusText option.
+func (e UnexpectedResponseError) statusFromCode(f func(int) string) {
+	e.statusText = f(e.statusCode)
+	if !e.HasStatusText() {
+		e.statusText = "unknown status code"
+	}
+}
+
+func (e UnexpectedResponseError) Error() string {
+	var eTxt strings.Builder
+	eTxt.WriteString("Unexpected response code")
+	if e.HasBody() || e.HasStatusCode() {
+		eTxt.WriteString(": ")
+	}
+	if e.HasStatusCode() {
+		eTxt.WriteString(fmt.Sprint(e.statusCode))
+		if e.HasBody() {
+			eTxt.WriteRune(' ')
+		}
+	}
+	if e.HasBody() {
+		eTxt.WriteString(fmt.Sprintf("(%s)", e.body))
+	}
+
+	return eTxt.String()
+}
+
+// UnexpectedResponseErrorOptions are functions passed to NewUnexpectedResponseError
+// to customize the created error.
+type UnexpectedResponseErrorOption func(*UnexpectedResponseError)
+
+// WithError allows the addition of a Go error that may have been encountered
+// while processing the response. For example, if there is an error constructing
+// the gzip reader to process a gzip-encoded response body.
+func WithError(e error) UnexpectedResponseErrorOption {
+	return func(u *UnexpectedResponseError) { u.err = e }
+}
+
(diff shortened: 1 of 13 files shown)


api/error_unexpected_response.go
```diff
@@ -0,0 +1,164 @@
...
+type UnexpectedResponseErrorSource func() *UnexpectedResponseError
+
+// FromHTTPResponse read an open HTTP response, drains and closes its body as
+// the data for the UnexpectedResponseError.
+func FromHTTPResponse(resp *http.Response) UnexpectedResponseErrorSource {
+	return func() *UnexpectedResponseError {
+		// collect and close the body
+		var buf bytes.Buffer
+		_, _ = io.Copy(&buf, resp.Body)
+		_ = resp.Body.Close()
```
tgross, line 115: `io.Copy` doesn't return an error on EOF, so if we ignore the error here are we ignoring some underlying problem that would be useful to know? Also, is it safe to close this `resp.Body` twice? We've not done so in the only consumer we have here, but it would be easy for someone to do `defer resp.Body.Close()` in a caller and miss that it's being closed here.

By the time we call this, we're already in an error state and things could be screwy, so maybe we should verify that `resp` and `resp.Body` are non-nil?

api/operator.go
```diff
@@ -338,13 +338,15 @@ func (op *Operator) LicenseGet(q *QueryOptions) (*LicenseReply, *QueryMeta, erro
...
 		return nil, nil, errors.New("Nomad Enterprise only endpoint")
 	}
 
-	if resp.StatusCode != 200 {
-		body, _ := io.ReadAll(resp.Body)
-		return nil, nil, fmt.Errorf("Unexpected response code: %d (%s)", resp.StatusCode, body)
+	if resp.StatusCode != http.StatusOK {
+		return nil, nil, NewUnexpectedResponseError(
+			FromHTTPResponse(resp),
+			WithExpectedStatuses([]int{http.StatusOK, http.StatusNoContent}),
```
tgross, line 348: Doesn't this say that we expected 200 or 204 here? If we did get a 204, we would have returned up above and never hit this code path I think, so shouldn't this just be expected 200?

api/error_unexpected_response.go
```diff
@@ -0,0 +1,164 @@
...
+		if e.HasBody() {
+			eTxt.WriteRune(' ')
+		}
+	}
+	if e.HasBody() {
+		eTxt.WriteString(fmt.Sprintf("(%s)", e.body))
+	}
+
+	return eTxt.String()
+}
```
tgross, line 72: I think this needs a block for `e.HasError` too so that we can report the inner error if we have one.

api/operator.go
angrycub: Since 204s are handled by the function (and consequently aren't unexpected) is why I included them in the list.

api/error_unexpected_response.go
angrycub: I avoided that since it would change the error message in https://github.com/hashicorp/nomad/blob/50a4c1376a52c2b070055c8d2d63492ddc640fce/api/api.go#L901-L904, but I'm not opposed overall

tgross: Right, but the original code for that just threw out the error, so why call `WithError(err)` if we're not going to report it? The caller won't be able to report it either (if this is used in ex. an external package). So if we handle the `e.HasError` case and not call `WithError` at `api/api.go#L904`, I think we get both?

api/operator.go
tgross: Works for me.

api/error_unexpected_response_test.go
```diff
@@ -0,0 +1,268 @@
...
+			method:    r.Method,
+			ipaddr:    ipAddrFromRemoteAddr(r.RemoteAddr),
+			code:      m.Code,
+			duration:  m.Duration,
+			size:      m.Written,
+			userAgent: r.UserAgent(),
+		}
+		t.Logf(ri.String())
+	})
+}
```
tgross, line 231: It doesn't seem like we assert on any of the metrics we're getting from `httpsnoop`, but there's enough indirection that I'm having trouble knowing for sure. Are we adding this wrapper just to add log lines to the test?

tgross (approved):
LGTM!

api/error_unexpected_response_test.go
angrycub: Yep, it is only for that mock test server to make a basic request log. It's definitely not super critical, but it did help me validate that my test mux was routing as it should have been.

api/error_unexpected_response.go
```diff
@@ -0,0 +1,176 @@
...
+func (e UnexpectedResponseError) HasStatusText() bool       { return e.statusText != "" }
+func (e UnexpectedResponseError) StatusText() string        { return e.statusText }
+func (e UnexpectedResponseError) HasBody() bool             { return e.body != "" }
+func (e UnexpectedResponseError) Body() string              { return e.body }
+func (e UnexpectedResponseError) HasError() bool            { return e.err != nil }
+func (e UnexpectedResponseError) Unwrap() error             { return e.err }
+func (e UnexpectedResponseError) HasAdditional() bool       { return e.additional != nil }
+func (e UnexpectedResponseError) Additional() error         { return e.additional }
+func NewUnexpectedResponseError(src UnexpectedResponseErrorSource, opts ...UnexpectedResponseErrorOption) UnexpectedResponseError {
+	new := src()
```
shoenig, line 36:
```suggestion
	new := src()
```

please don't shadow keywords

api/error_unexpected_response.go
```diff
@@ -0,0 +1,176 @@
...
+	return func(u *UnexpectedResponseError) { u.statusText = st }
+}
+
+// WithExpectedStatuses provides a list of statuses that the receiving function
+// expected to receive. This can be used by API callers to provide more feedback
+// to end-users.
+func WithExpectedStatuses(s []int) UnexpectedResponseErrorOption {
+	return func(u *UnexpectedResponseError) {
+		u.expected = make([]int, len(s))
+		copy(u.expected, s)
```
shoenig, line 107: one line with https://pkg.go.dev/golang.org/x/exp/slices#Clone

api/error_unexpected_response.go
```diff
@@ -0,0 +1,176 @@
...
+			u.statusText = strings.TrimSpace(strings.TrimPrefix(resp.Status, fmt.Sprint(resp.StatusCode)))
+			u.body = body
+		}
+		return u
+	}
+}
+
+// FromStatusCode is the "thinnest" source for an UnexpectedResultError. It
+// will attempt to resolve the status code to status text using a resolving
+// function provided inside of the NewUnexpectedResponseError implementation.
```
shoenig, line 142: The first sentence doesn't add any value, just start with the second sentence.

api/error_unexpected_response_test.go
```diff
@@ -0,0 +1,268 @@
...
+	"github.com/felixge/httpsnoop"
+	"github.com/hashicorp/nomad/api"
+	"github.com/shoenig/test/must"
+	"github.com/shoenig/test/portal"
+)
+
+const mockNamespaceBody = `{"Capabilities":null,"CreateIndex":1,"Description":"Default shared namespace","Hash":"C7UbjDwBK0dK8wQq7Izg7SJIzaV+lIo2X7wRtzY3pSw=","Meta":null,"ModifyIndex":1,"Name":"default","Quota":""}`
+
+func TestUnexpectedResponseError(t *testing.T) {
+	t.Parallel()
```
shoenig, line 24: the `testutil.Parallel` helper gives us a convenient hook for manipulating parallel behavior - please use it so that we can control these tests down the line if we need to 

(same for all uses of `t.Parallel`)

api/error_unexpected_response_test.go
```diff
@@ -0,0 +1,268 @@
...
+		must.True(t, ure.HasBody())
+		must.Eq(t, "{", ure.Body()) // The body is truncated to the first byte
+	})
+}
+
+// testServer creates a httptest.Server that can be used to serve simple mock
+// data, which is faster than starting a real Nomad agent.
+func testServer(t *testing.T) string {
+	grabber := portal.New(t)
+	ports := grabber.Grab(1)
```
shoenig, line 127: By default this will use the same `127.0.0.1` address as the `portal.Grabber` being used by `testutil.TestServer`. We should either configure this one to use a different address, or reactor both callers to use a common `Grabber` like we do for the non-api packages in https://github.com/hashicorp/nomad/blob/main/ci/ports.go#L15-L20

api/error_unexpected_response.go
```diff
@@ -0,0 +1,176 @@
...
+		copy(u.expected, s)
+	}
+}
+
+// UnexpectedResponseErrorSource provides the basis for a NewUnexpectedResponseError.
+type UnexpectedResponseErrorSource func() *UnexpectedResponseError
+
+// FromHTTPResponse read an open HTTP response, drains and closes its body as
+// the data for the UnexpectedResponseError.
+func FromHTTPResponse(resp *http.Response) UnexpectedResponseErrorSource {
```
shoenig, line 116: The callsite of this would look like

```
err := api.FromHTTPResponse(resp)
```

which doesn't convey much about what the function does or returns - seems like this should have a name like `ErrorFromHTTPResponse` ?

api/error_unexpected_response_test.go
```diff
@@ -0,0 +1,268 @@
...
+		// content-length itself.
+		io.Copy(w, er)
+	})
+}
+
+// testLogRequestHandler wraps a http.Handler with a logger that writes to the
+// test log output
+func testLogRequestHandler(t *testing.T, h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		t := t
```
shoenig, line 217: are you sure this shallow copy is necessary? Unlike `for`, the parent `t` isn't overwritten by subsequent callers, they all get their own `t`

shoenig (approved):
LGTM! just some nitpicky suggestions, and the portal thing probably needs to be fixed

api/error_unexpected_response.go
angrycub: `FromHTTPResponse(resp)` doesn't return `error`, it makes a required function parameter that `NewUnexpectedResponseError(...)` consumes to populate the error it's building. The expectation is that you call `NewUnexpectedResponseError(FromHTTPResponse(resp))`. 

Is there a better conventional for naming the `UnexpectedResponseErrorSource` funcs? Also, since their is internal to the API package, maybe they should be unimported so Go API consumers don't think that they can/should use them.

api/error_unexpected_response_test.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+// ipAddrFromRemoteAddr removes the port from the address:port in remote addr
+func ipAddrFromRemoteAddr(s string) string {
+	idx := strings.LastIndex(s, ":")
+	if idx == -1 {
+		return s
+	}
+	return s[:idx]
+}
```
shoenig, line 256: I believe this will choke on IPv6 addresses, instead there is https://pkg.go.dev/net/netip#ParseAddrPort which should do the trick

shoenig (approved):
LGTM - i do think we want to leave the IP parsing to the stdlib

Issue #8487, opened 2020-07-21
Prevent kill_timeout greater than progress_deadline


threemachines:
### Nomad version
Nomad v0.12.0 (8f7fbc8e7b5a4ed0d0209968faf41b238e6d5817)

### Operating system and Environment details
Ubuntu 18.04 on AWS

### Issue
If a kill_timeout is greater than the job's progress_deadline, allocations may keep running (after initial kill signal) for long enough that the deploy fails. However, the allocations that were scheduled by that job will still be pending after job failure, and will be placed and started once the previous allocations do exit.

I'm honestly split between this being a bug or a feature request. I can't say with certainty that the current behavior is *wrong*, but we certainly found it surprising and unpleasant. We found it by accidental misconfiguration, and I'm hard-pressed to imagine a situation where you'd really want your job to work this way, so why not add a guard-rail?

The alternative fix I can imagine is that the timer for progress_deadline could not start until allocations have been placed, but that would have other problems.

### Reproduction steps
1) Deploy the below job file. Should be fine.
2) Change the environment variable to force redeploys and deploy it again. Deploy will fail on timeout.
3) Do it again, just for fun. (Changing environment variables again.)
4) Wait until five minutes after the second deploy.

You now have fifteen allocations from deploy 0, and five allocations from deploy 2, which all started several minutes after deploy 2 failed. (The five allocations from deploy 1 went directly from pending to completed.)

### Job file
```
job "example" {
  datacenters = ["sandbox"]
  type = "service"
  update {
    max_parallel = 5
    min_healthy_time = "10s"
    healthy_deadline = "30s"
    progress_deadline = "1m"
    auto_revert = false
    canary = 0
  }
  migrate {
    max_parallel = 5
    health_check = "checks"
    min_healthy_time = "10s"
    healthy_deadline = "5m"
  }
  group "cache" {
    count = 20
    restart {
      attempts = 2
      interval = "30m"
      delay = "15s"
      mode = "fail"
    }
    ephemeral_disk {
      size = 300
    }
    task "redis" {
      driver = "docker"
      ## send redis a sighup so it doesn't actually exit
      ## this lets us play with the kill_timeouts
      kill_signal = "SIGHUP"
      kill_timeout = "5m"
      config {
        image = "redis:3.2"
        port_map {
          db = 6379
        }
      }
      env {
        asdf = "asdf"
      }
      resources {
        cpu    = 500
        memory = 256
        network {
          mbits = 10
          port "db" {}
        }
[...]

PR #16761, opened 2023-04-03
Prevent kill_timeout greater than progress_deadline 

Juanadelacuesta:
This PR addresses the bug reported on https://github.com/hashicorp/nomad/issues/8487

If a kill_timeout is greater than the job's progress_deadline, allocations may keep running (after initial kill signal) for long enough that the deploy fails. However, the allocations that were scheduled by that job will still be pending after job failure, and will be placed and started once the previous allocations do exit. This PR adds a validation to avoid this situation.


jrasell (changes requested):
Looking good, thanks @Juanadelacuesta! It looks like the new validation has meant some tests are now breaking, which need to be fixed up. Lets also add a changelog entry to this PR.

nomad/structs/structs.go
```diff
@@ -5145,6 +5153,38 @@ func (u *UpdateStrategy) Rolling() bool {
 	return u.Stagger > 0 && u.MaxParallel > 0
 }
 
+func (u *UpdateStrategy) Canonicalize() {
```
jrasell, line 5156: I worry that having this new functionality will mean users upgrading will see their job specification change between versions. I wonder why we need this, when the canonicalization should be handled by before it reaches the RPC.

I took a look at the use of `structs.DefaultUpdateStrategy` and it seems this only currently gets used by mocks and tests, which suggests we currently rely on the API default instantiation as mentioned above.

.changelog/16761.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+code: Prevent Kill Timeout greater than Progress Deadline on Update block
```
jrasell, line 2: code or core? It'd be nice if the parameters were detailed in their jobspec form, as it makes it easier for users to grok.

jrasell (changes requested):

nomad/structs/structs.go
```diff
@@ -5136,6 +5136,12 @@ func (u *UpdateStrategy) IsEmpty() bool {
 		return true
 	}
 
+	// When the Job is transformed from api to struct, the Update Strategy block is
+	// copied into the existing task groups, the only things that are passed along
+	// are MaxParallel and Stagger, because they are enforced at job level.
+	// That is why checking if MaxParallel is cero is enough to know if the
```
jrasell, line 5142: thanks for adding this!
```suggestion
	// That is why checking if MaxParallel is zero is enough to know if the
```

jrasell (approved):
Couple of nit pick changes, but LGTM, thanks!

We will need to add backport labels before merging this in.

PR #16763, opened 2023-04-03
api: enable support for setting original job source

shoenig:
Note to reviewers: this PR isn't as large as it looks - most file changes are just due to `UpsertJob` getting a new signature, and that is used in many, many tests.

This PR adds support for setting job source material along with the registration of a job.

This includes a new HTTP endpoint and a new RPC endpoint for making queries for the original source of a job. The HTTP endpoint is `/v1/job/<id>/submission?version=<version>` and the RPC method is `Job.GetJobSubmission`. This endpoint is limited by ACLs with `read-job` capability. It is quite similar to the JobVersions endpoint, in that data is upserted and removed along with a Job, but queryable through a separate endpoint.

The job source (if submitted, and doing so is always optional), is stored in the `job_submission` memdb table, separately from the actual job. This way we do not incur overhead of reading the potentially huge string field throughout normal job operations.

The server config now includes `job_max_source_size` for configuring the maximum size the job source may be, before the server simply drops the source material. This should help prevent Bad Things from happening when huge jobs are submitted. If the value is set to 0, all job source material will be dropped.

Will add e2e tests in a followup PR, this one is already too large.

No backports, this is for a 1.6 feature.

### DIFF
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -1312,6 +1312,7 @@ type JobRevertRequest struct {
 type JobRegisterRequest struct {
 	Submission *JobSubmission
 	Job        *Job
+
 	// If EnforceIndex is set then the job will only be registered if the passed
 	// JobModifyIndex matches the current Jobs index. If the index is zero, the
 	// register only occurs if the job is new.
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/cronexpr"
+	"golang.org/x/exp/maps"
 )
 
 const (
@@ -68,6 +69,11 @@ type JobsParseRequest struct {
 	// HCLv1 indicates whether the JobHCL should be parsed with the hcl v1 parser
 	HCLv1 bool `json:"hclv1,omitempty"`
 
+	// Variables are HCL2 variables associated with the job. Only works with hcl2.
+	//
+	// Interpreted as if it were the content of a variables file.
+	Variables string
+
 	// Canonicalize is a flag as to if the server should return default values
 	// for unset fields
 	Canonicalize bool
@@ -78,7 +84,7 @@ func (c *Client) Jobs() *Jobs {
 	return &Jobs{client: c}
 }
 
-// ParseHCL is used to convert the HCL repesentation of a Job to JSON server side.
+// ParseHCL is used to convert the HCL representation of a Job to JSON server side.
 // To parse the HCL client side see package github.com/hashicorp/nomad/jobspec
 // Use ParseHCLOpts if you need to customize JobsParseRequest.
 func (j *Jobs) ParseHCL(jobHCL string, canonicalize bool) (*Job, error) {
@@ -89,10 +95,8 @@ func (j *Jobs) ParseHCL(jobHCL string, canonicalize bool) (*Job, error) {
 	return j.ParseHCLOpts(req)
 }
 
-// ParseHCLOpts is used to convert the HCL representation of a Job to JSON
-// server side. To parse the HCL client side see package
-// github.com/hashicorp/nomad/jobspec.
-// ParseHCL is an alternative convenience API for HCLv2 users.
+// ParseHCLOpts is used to request the server convert the HCL representation of a
+// Job to JSON on our behalf. Accepts HCL1 or HCL2 jobs as input.
 func (j *Jobs) ParseHCLOpts(req *JobsParseRequest) (*Job, error) {
 	var job Job
 	_, err := j.client.put("/v1/jobs/parse", req, &job, nil)
@@ -116,6 +120,7 @@ type RegisterOptions struct {
 	PolicyOverride bool
 	PreserveCounts bool
 	EvalPriority   int
+	Submission     *JobSubmission
 }
 
 // Register is used to register a new job. It returns the ID
@@ -134,9 +139,7 @@ func (j *Jobs) EnforceRegister(job *Job, modifyIndex uint64, q *WriteOptions) (*
 // returns the ID of the evaluation, along with any errors encountered.
 func (j *Jobs) RegisterOpts(job *Job, opts *RegisterOptions, q *WriteOptions) (*JobRegisterResponse, *WriteMeta, error) {
 	// Format the request
-	req := &JobRegisterRequest{
-		Job: job,
-	}
+	req := &JobRegisterRequest{Job: job}
 	if opts != nil {
 		if opts.EnforceIndex {
 			req.EnforceIndex = true
@@ -145,6 +148,7 @@ func (j *Jobs) RegisterOpts(job *Job, opts *RegisterOptions, q *WriteOptions) (*
 		req.PolicyOverride = opts.PolicyOverride
 		req.PreserveCounts = opts.PreserveCounts
 		req.EvalPriority = opts.EvalPriority
+		req.Submission = opts.Submission
 	}
 
 	var resp JobRegisterResponse
@@ -250,6 +254,19 @@ func (j *Jobs) Versions(jobID string, diffs bool, q *QueryOptions) ([]*Job, []*J
 		return nil, nil, nil, err
 	}
 	return resp.Versions, resp.Diffs, qm, nil
+}
+
+// Submission is used to retrieve the original submitted source of a job given its
+// namespace, jobID, and version number. The original source might not be available,
+// which case nil is returned with no error.
+func (j *Jobs) Submission(jobID string, version int, q *QueryOptions) (*JobSubmission, *QueryMeta, error) {
+	var sub JobSubmission
+	s := fmt.Sprintf("/v1/job/%s/submission?version=%d", url.PathEscape(jobID), version)
+	qm, err := j.client.query(s, &sub, q)
+	if err != nil {
+		return nil, nil, err
+	}
+	return &sub, qm, nil
 }
 
 // Allocations is used to return the allocs for a given job ID.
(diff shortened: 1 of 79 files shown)


website/content/api-docs/jobs.mdx
```diff
@@ -122,6 +122,10 @@ The table below shows this endpoint's support for
 
 - `Job` `(Job: <required>)` - Specifies the JSON definition of the job.
 
+- `Submission` `(JobSubmission: <optional>)` - Specifies the original HCL/HCL2/JSON
+  definition of the job. This data is useful for reference only, it is not considered
+  for the actual registration of `Job`.
```
schmichael, line 127:
```suggestion
  for the actual scheduling of `Job`.
```

A bit pedantic so nbd either way. I think arguably it _is_ considered by the Job.Register operation, just not for any operational logic. Just saying "scheduling" might make the distinction more clear even if it's ignored by more than just scheduling logic.

command/job_run.go
```diff
@@ -317,6 +317,7 @@ func (c *JobRunCommand) Run(args []string) int {
 		PolicyOverride: override,
 		PreserveCounts: preserveCounts,
 		EvalPriority:   evalPriority,
+		Submission:     sub,
```
schmichael, line 320: Could we add a `-drop-source` flag (or similar) to allow people to opt out of submitting the source? I'm thinking it might be useful for batch/sysbatch jobs where you might be submitting a relatively large number in an automated way and don't want the overhead of the source.

schmichael (commented):
This is really exciting! Great work.

nomad/job_endpoint.go
```diff
@@ -1183,6 +1182,52 @@ func (j *Job) Scale(args *structs.JobScaleRequest, reply *structs.JobRegisterRes
...
+		return structs.ErrPermissionDenied
+	}
+
+	// Setup the blocking query
+	opts := blockingOptions{
+		queryOpts: &args.QueryOptions,
+		queryMeta: &reply.QueryMeta,
+		run: func(ws memdb.WatchSet, state *state.StateStore) error {
+			// Look for the submission
+			out, err := state.JobSubmission(ws, args.RequestNamespace(), args.JobID, args.Version)
```
tgross, line 1209: We're submitting the jobspec with the parent job and there's no pointer from `structs.Job` to the `structs.JobSubmission` (which is totally reasonable). But that means for periodic jobs when we call [`deriveJob`](https://github.com/hashicorp/nomad/blob/v1.5.3/nomad/periodic.go#L404-L432) (and similarly for dispatch jobs) we're copying the `structs.Job` and changing the job ID, which leaves no way to get from the derived job to the job submission anymore.

Maybe we should detect the case where there's a parent job ID and pass that as the job ID parameter here instead?

api/jobs.go
```diff
@@ -863,6 +880,51 @@ type ParameterizedJobConfig struct {
...
+	// Format indicates what the Source content was (hcl1, hcl2, or json).
+	Format string
+
+	// VariableFlags contains the CLI "-var" flag arguments as submitted with the
+	// job (hcl2 only).
+	VariableFlags map[string]string
+
+	// Variables contains the opaque variables configuration as coming from
+	// a var-file or the WebUI variables input (hcl2 only).
+	Variables string
```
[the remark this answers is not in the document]
shoenig: Since the parse endpoint is really only used by the webUI and the webUI's variable content is coming from a html form, I don't think it quite make sense to associate the content with a file name

command/agent/job_endpoint.go
```diff
@@ -702,6 +747,37 @@ func (s *HTTPServer) jobDispatchRequest(resp http.ResponseWriter, req *http.Requ
 	return out, nil
 }
 
+// writeVariablesFile writes content to a temporary file that is to be read by
+// the hcl parser. If content is empty nothing is written and nil is returned.
+// The return value is otherwise a one element slice with the filename of the
+// temporary file. Also returned is a cleanup function that must be called by
+// the caller for removing the temporary file once it is no longer needed.
+func writeVariablesFile(content string) ([]string, func(), error) {
```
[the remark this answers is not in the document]
shoenig: For some reason I thought it was the hcl library requiring file names, but yeah it's just our own `jobspec2` package. I expanded the `ParseOptions` to also enable setting raw variable file content, and now just use that eliminating the need for files. https://github.com/hashicorp/nomad/pull/16763/commits/eaddac3c6ed9a0b92c8694ec63b6d00eea2861d7

command/agent/job_endpoint.go
```diff
@@ -787,6 +872,32 @@ func (s *HTTPServer) jobServiceRegistrations(resp http.ResponseWriter, req *http
...
+	// allowable size as set by client config
+	totalSize := len(submission.Source)
+	totalSize += len(submission.Variables)
+	for key, value := range submission.VariableFlags {
+		totalSize += len(key)
+		totalSize += len(value)
+	}
+
+	if totalSize > maxSize {
+		return nil
```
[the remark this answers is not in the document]
shoenig: In 169988301 I move this check out of the HTTP layer and into the RPC layer, where we do lots of job admission controller things. That way we get to set a warning here, and avoid reading server config on clients which as you point out below doesn't exist.

nomad/state/state_store.go
```diff
@@ -5315,6 +5350,20 @@ func (s *StateStore) updateJobScalingPolicies(index uint64, job *structs.Job, tx
 	return nil
 }
 
+// updateJobSubmission stores the original job source and variables associated that the
+// job structure originates from. It is up to the job submitter to include the source
+// material, and as such sub may be nil, in which case nothing is stored.
+func (s *StateStore) updateJobSubmission(index uint64, sub *structs.JobSubmission, namespace, jobID string, version uint64, txn *txn) error {
+	if sub == nil || namespace == "" || jobID == "" {
```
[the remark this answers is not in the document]
shoenig: In f19721c47 we now return an error if `namespace` or `jobID` are not set

nomad/structs/structs.go
```diff
@@ -4246,6 +4266,44 @@ const (
...
+	// Version is managed internally, not set.
+	//
+	// The version of the Job this submission is associated with.
+	Version uint64
+
+	// JobIndex is managed internally, not set.
+	//
+	// The raft index the Job this submission is associated with.
+	JobIndex uint64
+}
```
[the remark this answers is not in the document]
shoenig: Renamed to clarify in 37e71b237

nomad/job_endpoint.go
schmichael: I like keeping this simple and leaving that up to higher layers (UI and API). I think job invocations (aka child dispatch or periodic jobs) should be as minimal as possible, and I think reflecting that in the lowlevel RPCs/APIs is important to let users know they're minimal.

I would hate to include the parent source and give a user the impression we're "wasting" space saving the same job submission for every invocation of a job. That could lead them to disable the feature erroneously thinking they'll get a big performance improvement.

command/job_run.go
schmichael: From offline discussion it's become apparent we need to add something to jobspecs to say "Don't save me!" so that operators don't have to change _how_ they submit jobs depending on whether or not they want the source persisted.

shoenig:
In ca315b5ec the state store is updated to prune all but the last `structs.JobTrackedVersions` (6) job submissions, ensuring we do not balloon the state store on stored job submissions.

website/content/api-docs/jobs.mdx
```diff
@@ -235,6 +239,12 @@ The table below shows this endpoint's support for
 - `Canonicalize` `(bool: false)` - Flag to enable setting any unset fields to
   their default values.
 
+- `Variables` `(string: "")` - Specifies HCL2 variables to use during parsing of
+  the job in the var file format.
+
+- `VariableFlags` `(map[string]string: nil)` - Specifies HCL2 variables to use
+  during parsing of the job in key = value format.q
```
tgross, line 246:
```suggestion
  during parsing of the job in key = value format.
```

website/content/api-docs/jobs.mdx
```diff
@@ -547,6 +557,62 @@ $ curl \
   - `batch`: Allocations are intended to exit.
   - `system`: Each client gets an allocation.
 
+## Read Job Submission
+
+This endpoint reads original source information about a specific version of a single
+job. The data this endpoint provides is only available if it was provided with the
+original job during job registration.
```
tgross, line 564: Should we include details here about the maximum number of submissions to expect to remain?

command/agent/config.go
```diff
@@ -606,6 +606,11 @@ type ServerConfig struct {
 	// for the EventBufferSize is 1.
 	EventBufferSize *int `hcl:"event_buffer_size"`
 
+	// JobMaxSourceSize limits the maximum size of a jobs source hcl/json
+	// before being discarded automatically. If unset, the maximum size defaults
+	// to 1 MiB. If the value is zero, no job sources will be stored.
```
tgross, line 611: We're setting it to `1M` which is 1 million bytes, not 1024^2 bytes (https://go.dev/play/p/_F5kNg8Mtqn):

```suggestion
	// to 1 MB. If the value is zero, no job sources will be stored.
```

That being said, we set almost all our other byte size configs in powers-of-2 bytes and not powers-of-10 bytes, so I wouldn't be opposed to setting the default to `1 MiB` either.

command/helpers.go
```diff
@@ -476,24 +473,34 @@ func (j *JobGetter) Get(jpath string) (*api.Job, error) {
...
+	jobfile = io.TeeReader(jobfile, &source)
 	var err error
 	switch {
 	case j.HCL1:
 		jobStruct, err = jobspec.Parse(jobfile)
+
+		// include the hcl1 source as the submission
+		jobSubmission = &api.JobSubmission{
+			Source: source.String(),
+			Format: formatHCL1,
```
tgross, line 500: If we store HCL1 will the UI need to have the `-hcl1` flag too?

nomad/job_endpoint.go
tgross: That's reasonable, but we'll need to let the UI folks know that and we should document that you're responsible for sending the parent ID in the API docs.

nomad/state/state_store.go
```diff
@@ -5315,6 +5352,74 @@ func (s *StateStore) updateJobScalingPolicies(index uint64, job *structs.Job, tx
...
+
+	// prune old job submissions
+	return s.pruneJobSubmissions(namespace, jobID, txn)
+}
+
+func (s *StateStore) pruneJobSubmissions(namespace, jobID string, txn *txn) error {
+	// although the number of tracked submissions is the same as the number of
+	// tracked job versions, do not assume a 1:1 correlation, as there could be
+	// holes in the submissions (or none at all)
+	limit := structs.JobTrackedVersions
```
tgross, line 5386: Won't the number of tracked submissions always be `<=` the number of tracked job versions though? There might be "holes" but there won't be excess, so won't the job submissions automatically fall out as the tracked job versions fall out?

schmichael (approved):

command/helpers.go
tgross: Yeah, good point

tgross (approved):

PR #16776, opened 2023-04-03
cni: fix plugin fingerprinting versions

EtienneBruines:
CNI plugins v1.2.0 and above output a second line, containing supported protocol versions. Nomad tests should support this.

### DIFF
--- a/.changelog/16776.txt
+++ b/.changelog/16776.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+client: Fix CNI plugin version fingerprint when output includes protocol version
+```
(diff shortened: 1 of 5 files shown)


lgfa29:
Ah thanks for letting us know @EtienneBruines.

As expected, this change breaks our tests 😅 
https://github.com/hashicorp/nomad/actions/runs/4598672300/jobs/8185888257?pr=16776

Do you think you would be able to fix them?

lgfa29 (changes requested):
Ah thanks for letting us know @EtienneBruines.

As expected, this change breaks our tests 😅 
https://github.com/hashicorp/nomad/actions/runs/4598672300/jobs/8185888257?pr=16776

Do you think you would be able to fix them?

PR #16781, opened 2023-04-04
cli: fix up failing quota inspect enterprise test.

jrasell:


command/quota_inspect_test.go
```diff
@@ -78,7 +76,7 @@ func TestQuotaInspectCommand_Run(t *testing.T) {
 	must.Zero(t, code)
 
 	out = ui.OutputWriter.String()
-	must.StrContains(t, out, "test-quota")
+	must.StrContains(t, out, "test")
```
tgross, line 79: Two minor nitpicks that come from using the shared `testQuotaSpec` function for the object under test:
* Should this also be `qs.Name` as we've done on line 62R above so that we can't unintentionally break it in the future?
* Seeing as how our assertions are just checking for the string, should we generate a name for the quota spec rather than calling it `"test"` so that we make this _pass_ unintentionally later by changing the region name to `"test"`, or adding some other quota spec field that we name `"test"`?

tgross (approved):
LGTM!

command/quota_inspect_test.go
jrasell: Both totally great points; I'll get those changed.

Issue #16789, opened 2023-04-04
Artifact getter breaks when Nomad is run as non-root


the-nando:
### Nomad version
Nomad v1.5.1+ent

### Issue

The artifact getter now runs as a subprocess since v1.5.0. We run nomad as a non-root user and the getter fails with:
```
fork/exec /usr/bin/nomad: operation not permitted
```
The error originates from this syscall:
```
[pid 18934] setgroups(0, NULL <unfinished ...>
[pid 16745] epoll_pwait(4,  <unfinished ...>
[pid 18934] <... setgroups resumed>)    = -1 EPERM (Operation not permitted)
```
Which comes from setting the GID and UID in SysProcAttr:
https://github.com/hashicorp/nomad/blob/v1.5.1/client/allocrunner/taskrunner/getter/util.go#L140
https://github.com/hashicorp/nomad/blob/v1.5.1/client/allocrunner/taskrunner/getter/util_linux.go#L29

The code was intended to be used to drop privileges and run the getter as nobody when the nomad binary runs as root but that has been reverted in https://github.com/hashicorp/nomad/commit/95359b8c4cc9afbfe4db9b612d07f5116bad1066. 
Unfortunately when nomad is run as a non user, setting the GID for the child doesn't work as the user lacks `cap_setgid`.
A workaround is:
```
# setcap 'cap_setgid=ep' /usr/sbin/nomad 
```
or removing setting the attributes altogether as they are technically non required and they'll be inherited from the parent:
```
# diff --git a/client/allocrunner/taskrunner/getter/util.go b/client/allocrunner/taskrunner/getter/util.go

index 1a71017f3..8edf5a449 100644
--- a/client/allocrunner/taskrunner/getter/util.go
+++ b/client/allocrunner/taskrunner/getter/util.go
@@ -137,7 +137,7 @@ func (s *Sandbox) runCmd(env *parameters) error {
 	cmd.Stdin = env.reader()
 	cmd.Stdout = output
 	cmd.Stderr = output
-	cmd.SysProcAttr = attributes()
+	// cmd.SysProcAttr = attributes()

 	// start & wait for the subprocess to terminate
 	if err := cmd.Run(); err != nil {
```

Same issue was addressed here for non-Linux systems: https://github.com/hashicorp/nomad/commit/78d8ae01afc6a2efb7a8a610c5b085127eb2e4a6

PR #16791, opened 2023-04-05
Do not set attributes when spawning the getter child

the-nando:
https://github.com/hashicorp/nomad/commit/95359b8c4cc9afbfe4db9b612d07f5116bad1066 disabled running the artifact downloader on Linux as the user `nobody` but left behind the code to set UID / GID, which are currently set as the same ones the user Nomad runs under. The current implementation however doesn't work, as described in https://github.com/hashicorp/nomad/issues/16789.

This PR removes setting attributes altogether as they are technically non required and they'll be inherited from the parent.

Fixes https://github.com/hashicorp/nomad/issues/16789

### DIFF
--- a/.changelog/16791.txt
+++ b/.changelog/16791.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+client: Remove setting attributes when spawning the getter child
+```
--- a/client/allocrunner/taskrunner/getter/testing.go
+++ b/client/allocrunner/taskrunner/getter/testing.go
@@ -28,20 +28,15 @@ func TestSandbox(t *testing.T) *Sandbox {
 //
 // returns alloc_dir, task_dir
 func SetupDir(t *testing.T) (string, string) {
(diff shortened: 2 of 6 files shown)


shoenig (commented):
Thanks for the PR @the-nando! 

I believe we can eliminate the `credentials()` helpers altogether now as well since they aren't being used in any case. Do you mind cleaning those up as well? 

And can you run `make cl` to add a Changelog entry for this PR?

Thanks!

the-nando:
The `credentials()` helper is still referenced in https://github.com/hashicorp/nomad/blob/511d0c1e705543944a19f2174087b66be02101fc/client/allocrunner/taskrunner/getter/testing.go#L31
Does it still make sense to test for directory ownership as the getter now can only run under the same user Nomad runs?
If not, I can remove the `chown`'s assertions and cleanup the other files too.

Issue #13660, opened 2022-07-08
Deployment Status Command Does Not Respect -namespace Wildcard


MikaelHashiCorp:
### Nomad version
v1.3.1
v1.3.2-dev (c0f2f7ec19e5df889d2d0dd15f037e8edefc8a97+CHANGES)

### Operating system and Environment details

Ubuntu 20.04.4 LTS

### Issue

When passing in the -namespace="*" flag into the "deployment status" command, it returns a error instead of returning the proper deployment.  Error message:
```
Error retrieving deployment: Deployment ID "8faa0e1a" matched no deployments
```

### Reproduction steps

Get a deployment ID:
```
nomad deployment list
```

Get its info using the status command:
```
nomad deployment status -namespace="default" 8faa0e1a
```

Note it returns the info:
```
ID          = 8faa0e1a
Job ID      = grafana
Job Version = 0
Status      = successful
Description = Deployment completed successfully

Deployed
Task Group  Desired  Placed  Healthy  Unhealthy  Progress Deadline
grafana     1        1       1        0          2022-07-07T00:04:08Z
```

Try to get its info using a namespace wildcard:
```
nomad deployment status -namespace="*" "
```

Note it says "matched no deployment":
```
Error retrieving deployment: Deployment ID "8faa0e1a" matched no deployments
```

#### Expected Result

The wildcard value should allow getting deployment info from any namespace.

#### Actual Result

An error is returned instead:

![2022-07-08_15-07-17](https://user-images.githubusercontent.com/89491078/178076478-245e4159-29d2-44a0-a685-b963844e5035.png)

tgross:
A quick glance at the code suggests that the `Deployment.List` RPC has the same bug as was fixed for `Eval.List` in https://github.com/hashicorp/nomad/pull/13530

PR #16792, opened 2023-04-05
Deployment Status Command Does Not Respect -namespace Wildcard

Juanadelacuesta:
This PR addresses the bug reported [here](https://github.com/hashicorp/nomad/issues/13660)

The deployments Namespace filter did not take into account the namespace wildcard, resulting on an empty response every time it was used in conjunction with a prefix.  This PR adds the missing wildcard check as well as filtering by the allowed namespaces according to the ACL token.

### DIFF
--- a/nomad/deployment_endpoint.go
+++ b/nomad/deployment_endpoint.go
@@ -450,11 +450,15 @@ func (d *Deployment) List(args *structs.DeploymentListRequest, reply *structs.De
 
 	// Check namespace read-job permissions against request namespace since
 	// results are filtered by request namespace.
-	if aclObj, err := d.srv.ResolveACL(args); err != nil {
+	aclObj, err := d.srv.ResolveACL(args)
+	if err != nil {
 		return err
-	} else if aclObj != nil && !aclObj.AllowNsOp(namespace, acl.NamespaceCapabilityReadJob) {
+	}
+	if aclObj != nil && !aclObj.AllowNsOp(namespace, acl.NamespaceCapabilityReadJob) {
 		return structs.ErrPermissionDenied
 	}
+
+	allow := aclObj.AllowNsOpFunc(acl.NamespaceCapabilityReadJob)
 
 	// Setup the blocking query
 	sort := state.SortOption(args.Reverse)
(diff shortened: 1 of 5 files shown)


nomad/deployment_endpoint.go
```diff
@@ -450,20 +450,33 @@ func (d *Deployment) List(args *structs.DeploymentListRequest, reply *structs.De
...
 		queryMeta: &reply.QueryMeta,
 		run: func(ws memdb.WatchSet, store *state.StateStore) error {
+			allowableNamespaces, err := allowedNSes(aclObj, store, allow)
+			if err != nil {
+				if err == structs.ErrPermissionDenied {
+					reply.Deployments = make([]*structs.Deployment, 0)
+					return nil
+				}
+				return err
+			}
```
tgross, line 477: Not blocking but something we might want to get agreement on with the team and then document in our RPC checklist...

There's a small inconsistency here between the behavior of the first permissions check (on line 458) and this permissions check in the blocking query:
* If you have no permission to any namespace when you make the query, you get `ErrPermissionDenied`
* If you have permissions and make a blocking query, and while you're waiting on the blocking query the permissions are removed, you get an empty list.

I spot-checked some of the other `List` RPCs and it looks like we're inconsistent with this across the code base; some RPCs have both checks (ex. `Eval.List`, `Job.List`) while others only have the second check (ex. `Namespaces.ListNamespace`, `Variables.List`).

tgross (approved):
LGTM!

nomad/deployment_endpoint.go
```diff
@@ -491,8 +504,14 @@ func (d *Deployment) List(args *structs.DeploymentListRequest, reply *structs.De
...
 
+			filters := []paginator.Filter{
+				paginator.NamespaceFilter{
+					AllowableNamespaces: allowableNamespaces,
+				},
+			}
+
 			var deploys []*structs.Deployment
-			paginator, err := paginator.NewPaginator(iter, tokenizer, nil, args.QueryOptions,
+			paginator, err := paginator.NewPaginator(iter, tokenizer, filters, args.QueryOptions,
```
jrasell, line 514: While we are in this area, could we change the name of the `paginator` variable as it clashes with an import.

jrasell (approved):
LGTM, minor inline suggestion which is not blocking. The PR will also need a changelog entry, so everyone can bask in the fix!

Something that isn't needed before merge would be to use must for all the new test additions, rather than a mix of must and assert/require.

PR #16793, opened 2023-04-05
deps: update go-m1cpu with fix for BigSur users

shoenig:
https://github.com/shoenig/go-m1cpu/releases/tag/v0.1.5

which may also help with the Nix package for Nomad, which makes use of
the older IOKit framework version

https://github.com/shoenig/go-m1cpu/issues/5


jrasell (approved):
LGTM!

Issue #16798, opened 2023-04-05
rate metrics from some client-to-server RPCs incorrect (1.5.3+)


tgross:
In Nomad 1.5.3 we fixed a security bug that allowed bypass of ACL checks if the request came thru a client node first. But this fix broke (knowingly) the identification of many client-to-server RPCs. These will be now measured as if they were anonymous. The reason for this is that many client-to-server RPCs do not send the node secret and instead rely on the protection of mTLS.

The fix for this is to ensure that the node secret is being sent with every client-to-server RPC request. This is going to be complicated by:
* `Node.Register` RPC, which is Trust on First Use (TOFU)
* `Client.GetAllocs` RPC, which is sent in parallel with `Node.Register`, which means that clients would show errors during their initial startup.

PR #16799, opened 2023-04-05
client: send node secret with every client-to-server RPC

tgross:
In Nomad 1.5.3 we fixed a security bug that allowed bypass of ACL checks if the request came thru a client node first. But this fix broke (knowingly) the identification of many client-to-server RPCs. These will be now measured as if they were anonymous. The reason for this is that many client-to-server RPCs do not send the node secret and instead rely on the protection of mTLS.

This changeset ensures that the node secret is being sent with every client-to-server RPC request. In a future version of Nomad we can add enforcement on the server side, but this was left out of this changeset to reduce risks to the safe upgrade path.

Sending the node secret as an auth token introduces a new problem during initial introduction of a client. Clients send many RPCs concurrently with `Node.Register`, but until the node is registered the node secret is unknown to the server and will be rejected as invalid. This causes permission denied errors.

To fix that, this changeset introduces a gate on having successfully made a `Node.Register` RPC before any other RPCs can be sent (except for `Status.Ping`, which we need earlier but which also ignores the error because that handler doesn't do an authorization check). This ensures that we only send requests with a node secret already known to the server. This also makes client startup a little easier to reason about because we know `Node.Register` must succeed first, and it should make for a good place to hook in future plans for secure introduction of nodes. The tradeoff is that an existing client that has running allocs will take slightly longer (a second or two) to transition to ready after a restart, because the transition in `Node.UpdateStatus` is gated at the server by first submitting `Node.UpdateAlloc` with client alloc updates.

Fixes: https://github.com/hashicorp/nomad/issues/16798
Fixes: https://github.com/hashicorp/nomad-enterprise/issues/1069

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -302,6 +302,10 @@ type Client struct {
 	// applied to the node
 	fpInitialized chan struct{}
 
+	// registeredCh is closed when Node.Register has successfully run once.
+	registeredCh   chan struct{}
+	registeredOnce sync.Once
+
 	// serversContactedCh is closed when GetClientAllocs and runAllocs have
 	// successfully run once.
 	serversContactedCh   chan struct{}
@@ -376,6 +380,8 @@ func NewClient(cfg *config.Config, consulCatalog consul.CatalogAPI, consulProxie
 		invalidAllocs:        make(map[string]struct{}),
 		serversContactedCh:   make(chan struct{}),
 		serversContactedOnce: sync.Once{},
+		registeredCh:         make(chan struct{}),
+		registeredOnce:       sync.Once{},
 		cpusetManager:        cgutil.CreateCPUSetManager(cfg.CgroupParent, cfg.ReservableCores, logger),
 		getter:               getter.New(cfg.Artifact, logger),
 		EnterpriseClient:     newEnterpriseClient(logger),
@@ -1832,6 +1838,15 @@ func (c *Client) periodicSnapshot() {
 
 // run is a long lived goroutine used to run the client. Shutdown() stops it first
 func (c *Client) run() {
+
+	// Block until we've registered at least once so that we know the server has
+	// our node secret and we can authenticate
+	select {
+	case <-c.registeredCh:
+	case <-c.shutdownCh:
+		return
+	}
+
 	// Watch for changes in allocations
 	allocUpdates := make(chan *allocUpdates, 8)
 	go c.watchAllocations(allocUpdates)
@@ -1866,8 +1881,11 @@ func (c *Client) submitNodeEvents(events []*structs.NodeEvent) error {
 		nodeID: events,
 	}
 	req := structs.EmitNodeEventsRequest{
-		NodeEvents:   nodeEvents,
-		WriteRequest: structs.WriteRequest{Region: c.Region()},
+		NodeEvents: nodeEvents,
+		WriteRequest: structs.WriteRequest{
+			Region:    c.Region(),
+			AuthToken: c.secretNodeID(),
+		},
 	}
 	var resp structs.EmitNodeEventsResponse
 	if err := c.RPC("Node.EmitEvents", &req, &resp); err != nil {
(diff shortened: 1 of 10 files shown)


lgfa29 (approved):
Getting a more predictable node registration flow is a nice benefit 😄

client/client.go
```diff
@@ -2129,8 +2166,11 @@ func (c *Client) allocSync() {
 
 			// Send to server.
 			args := structs.AllocUpdateRequest{
-				Alloc:        toSync,
-				WriteRequest: structs.WriteRequest{Region: c.Region()},
+				Alloc: toSync,
+				WriteRequest: structs.WriteRequest{
+					Region:    c.Region(),
+					AuthToken: c.secretNodeID(),
```
schmichael, line 2172: My only concern here is that it doesn't *explicitly* block on registration. It *implicitly* blocks on registration because the `Client.run` method blocks and therefore no updates can occur until that's unblocked.

However if for some reason we change that behavior so that runners restart prior to registration, this will implicitly start sending un-authed AllocUpdateRequests.

I _think_ we should explicitly block here to enforce `no RPCs until secret is registered` as an invariant.

We could even change `secretNodeID() string` to be `secretNodeID() (secret string, shutdown bool)`, have it wait on `c.shutdownCh`, and return `"", true` if a shutdown occurred.

I don't know... I don't feel particularly strongly about this, but we do have a history of `change X + change Y much later = bug Z from emergent behavior`

client/client.go
```diff
@@ -1832,6 +1838,15 @@ func (c *Client) periodicSnapshot() {
...
 // run is a long lived goroutine used to run the client. Shutdown() stops it first
 func (c *Client) run() {
+
+	// Block until we've registered at least once so that we know the server has
+	// our node secret and we can authenticate
+	select {
+	case <-c.registeredCh:
+	case <-c.shutdownCh:
+		return
+	}
```
schmichael, line 1848: This substantially changes what happens if a disconnected client agent with workloads restarts. Currently if a workload crashes, it can be restarted locally without reconnecting to the server. With this change _after a restart_ the client will be blocked here and not monitoring workloads for exits, so they will not be restarted.

I think this also breaks disconnected client agents with Consul based TTL health checks as Nomad won't heartbeat those checks until the runners start.

Luckily I think there's a pretty easy fix that's also a nice optimization: **persist whether first registration has occurred.** This means we only ever block on for first registration when there's no workloads running anyway! If there are runners running then we _know_ we've successfully registered so we should start running them again ASAP.

### Refactoring

This does make me realize I don't think anyone writing a Nomad Client from scratch would structure the code this way... breaking the currently monolithic `NewClient` into its 3 actual phases:

1. NewClient - initialize the struct, no side effects or IO
2. Register - only blocks on first run
3. Run - ...the runners! all the non-registration goroutines get spawned here

...would maybe make all of this easier to reason about, but that would be a pretty major refactor and touch innumerable tests. I highly doubt this is worth pursuing, but it's fun to dream.

schmichael (changes requested):

client/client.go
tgross: Gah, I totally missed the restart case :facepalm: Persisting registration is simple enough, will fix. And that'll probably make an easy way to block the RPCs specifically as well (as noted above).

I sort of wish the server was what handed out the node secret because that would make it really easy to identify a node that hadn't yet registered because it wouldn't yet have a secret. Maybe we can revisit that workflow when we look at secure introduction of nodes in the near-ish future, and that might be a good time to consider the refactoring you're talking about here as well.

tgross:
Ok @schmichael I think I've got this working as discussed. I do want to do a little more end-to-end testing of this before merging but I figure in the meantime I'd try to get feedback on the approach.

client/client.go
schmichael: Resolved by the new RPC method code

client/client.go
```diff
@@ -1947,17 +1967,61 @@ func (c *Client) retryRegisterNode() {
...
 
+// getRegistrationToken gets the node secret to use for the Node.Register call.
+// Registration is trust-on-first-use so we can't send the auth token with the
+// initial request, but we want to add the auth token after that so that we can
+// capture metrics.
+func (c *Client) getRegistrationToken() string {
+
+	select {
+	case <-c.registeredCh:
+		return c.GetConfig().Node.SecretID
```
schmichael, line 1978:
```suggestion
		return c.secretNodeID()
```

client/client.go
```diff
@@ -1947,17 +1967,61 @@ func (c *Client) retryRegisterNode() {
...
+		// the 1st time or we've just restarted. Check the local state to see if
+		// we've written a successful registration previously so that we don't
+		// block allocrunner operations on disconnected clients.
+		registration, err := c.stateDB.GetNodeRegistration()
+		if err != nil {
+			c.logger.Error("could not determine previous node registration", "error", err)
+		}
+		if registration != nil && registration.HasRegistered {
+			c.registeredOnce.Do(func() { close(c.registeredCh) })
+			return c.GetConfig().Node.SecretID
```
schmichael, line 1990:
```suggestion
			return c.secretNodeID()
```

client/rpc.go
```diff
@@ -437,13 +457,14 @@ func resolveServer(s string) (net.Addr, error) {
...
-// Ping never mutates the request, so reuse a singleton to avoid the extra
-// malloc
-var pingRequest = &structs.GenericRequest{}
-
 // Ping is used to ping a particular server and returns whether it is healthy or
 // a potential error.
 func (c *Client) Ping(srv net.Addr) error {
+	pingRequest := &structs.GenericRequest{
+		QueryOptions: structs.QueryOptions{
+			AuthToken: c.GetConfig().Node.SecretID,
```
schmichael, line 465:
```suggestion
			AuthToken: c.secretNodeID(),
```

schmichael (approved):

PR #16800, opened 2023-04-05
agent: add top-level warning if mTLS is not configured

tgross:
Nomad's security model requires mTLS in order to secure client-to-server and server-to-server communications. Configuring ACLs alone is not enough. Loudly warn the user if mTLS is not configured in non-dev modes.

---

Note to reviewers: I'm totally happy to bikeshed the language used here  :grin: But it matches the other top-level pre-logging warnings like the bootstrap warning. 

Example output (first line of banner):

```
==> WARNING: mTLS is not configured - Nomad is not secure without mTLS!
==> WARNING: Bootstrap mode enabled! Potentially unsafe operation.
==> Loaded configuration from /home/tim/ws/nomad/etc/local/standalone.hcl
==> Starting Nomad agent...
==> Nomad agent configuration:

       Advertise Addrs: HTTP: 192.168.1.162:4646; RPC: 192.168.1.162:4647; Serf: 192.168.1.162:4648
            Bind Addrs: HTTP: [0.0.0.0:4646]; RPC: 0.0.0.0:4647; Serf: 0.0.0.0:4648
                Client: true
             Log Level: info
                Region: global (DC: dc1)
                Server: true
               Version: 1.5.4-dev

==> Nomad agent started! Log data will stream in below:

    2023-04-05T13:44:30.765-0400 [INFO]  nomad: setting up raft bolt store: no_freelist_sync=false
    2023-04-05T13:44:30.765-0400 [INFO]  nomad.raft: initial configuration: ...
```

### DIFF
--- a/command/agent/command.go
+++ b/command/agent/command.go
@@ -328,6 +328,10 @@ func (c *Command) IsValidConfig(config, cmdConfig *Config) bool {
 			c.Ui.Error(fmt.Sprintf("WARNING: Error when parsing TLS configuration: %v", err))
 		}
 	}
+	if !config.DevMode && (config.TLSConfig == nil ||
+		!config.TLSConfig.EnableHTTP || !config.TLSConfig.EnableRPC) {
+		c.Ui.Error("WARNING: mTLS is not configured - Nomad is not secure without mTLS!")
+	}
 
 	if config.Server.EncryptKey != "" {
 		if _, err := config.Server.EncryptBytes(); err != nil {


schmichael (approved):

lgfa29 (approved):
How annoying do we want to be? 😅 

There are a few more high-viz places we could output this (like CLI commands, the UI etc.) but without a general config it may be too much.

tgross:
> There are a few more high-viz places we could output this (like CLI commands, the UI etc.) but without a general config it may be too much.

Having in the UI might be nice, but probably would require that we pass the warning in the API because the UI gets proxied a lot.

PR #16810, opened 2023-04-05
hashicorp/go-msgpack v2

### LINKED ISSUE #16808: Using retracted version v1.1.5 of hashicorp/go-msgpack
### Nomad version
`master`

n/a
### Issue
The current version listed in `go.mod` is a retracted version, v1.1.5. Apparently version v2.0.0 is identical and should be used instead. This prevents updating to new versions of `github.com/hashicorp/nomad` in other projects (such as [github.com/hashicorp/terraform-provider-nomad](https://github.com/hashicorp/terraform-provider-nomad)
### Reproduction steps
Check `go.mod` file for `github.com/hashicorp/go-msgpack v1.1.5`
#### Expected Result
Find `github.com/hashicorp/go-msgpack v2.0.0` in `go.mod`
#### Actual Result
Find `github.com/hashicorp/go-msgpack v1.1.5` in `go.mod`

IamTheFij:
Updates to v2 of both hashicorp/go-msgpack and hashicorp/net-rpc-msgpackrpc to resolve the dependency on a the retracted version, v1.1.5, of hashicorp/go-msgpack.

hashicorp/raft still depends on hashicorp/go-msgpack v0.5.5, which is now an indirect dependency.

Fixes #16808


lgfa29 (approved):
Thanks for the work here @IamTheFij!

I was catching-up on this change and it seems like the `v2` tag was more of a way to help the Consul team manage its dependencies since they still use v0.5.5 directly and, without this tag, Go would pull v1.5.5 which does not work for them. Creating the `v2` tag makes Go aware of this incompatibility.

I pushed a commit to pin our dependency on v2.0.0 instead of v2.1.0 though. v2.1.0 does have some code changes that we the Nomad team will have to evaluate before adopting it.

Using v2.0.0 is no-op for us, so I will keep it at that for now.

PR #16815, opened 2023-04-06
consul/connect: fixed a bug where restarting proxy tasks failed.

### LINKED ISSUE #9887: envoyproxy/envoy:v${NOMAD_envoy_version} error="API error (400): invalid tag format"
### Nomad version
```
Nomad v1.0.2 (4c1d4fc6a5823ebc8c3e748daec7b4fda3f11037)
```
### Issue
After initiating a restart of a allocation, nomad cannot find envoy sidecar image anymore. I did a couple of restarts earlier today (same config) and did not run into this problem and consul does not report any problems.

```
Jan 26 09:30:08 node-3 nomad[6127]:     2021-01-26T09:30:08.751Z [INFO]  client.alloc_runner.task_runner: restarting task: alloc_id=5fa76958-04fe-e6b4-ce6f-9e4647fee2b8 task=connect-proxy-app-backend reason="Restart within policy" delay=16.074283377s
Jan 26 09:30:08 node-3 nomad[6127]:     2021-01-26T09:30:08.840Z [ERROR] client.driver_mgr.docker: failed pulling container: driver=docker image_ref=envoyproxy/envoy:v${NOMAD_envoy_version} error="API error (400): invalid tag format"
Jan 26 09:30:08 node-3 nomad[6127]:     2021-01-26T09:30:08.842Z [ERROR] client.alloc_runner.task_runner: running driver failed: alloc_id=c141dfaa-72c5-e589-1f7d-91628e8e501c task=connect-proxy-app-backend-service error="Failed to pull `envoyproxy/envoy:v${NOMAD_envoy_version}`: API error (400): invalid tag format"
```

Any way to find out what `NOMAD_envoy_version` is set to?

Cheers,
bert2002

jrasell:
The first start of a Consul Connect proxy sidecar triggers a run of the envoy_version hook which modifies the task config image entry. The modification takes into account a number of factors to correctly populate this. Importantly, once the hook has run, it marks itself as done so the taskrunner will not execute it again.

When the client receives a non-destructive update for the allocation which the proxy sidecar is a member of, it will update and overwrite the task definition within the taskerunner. In doing so it overwrite the modification performed by the hook. If the allocation is restarted, the envoy_version hook will be skipped as it previously marked itself as done, and therefore the sidecar config image is incorrect and causes a driver error.

The fix removes the hook in marking itself as done to the view of the taskrunner, and instead tracks this internally. The hook also implements the TaskPreKillHook interface, so that it's state can be correctly modified when required by task events.

closes #9887

The linked issue has reproduction steps for testing.

### DIFF
--- a/client/allocrunner/taskrunner/envoy_version_hook.go
+++ b/client/allocrunner/taskrunner/envoy_version_hook.go
@@ -62,7 +62,7 @@ func (envoyVersionHook) Name() string {
 	return envoyVersionHookName
 }
 
-func (h *envoyVersionHook) Prestart(_ context.Context, request *ifs.TaskPrestartRequest, response *ifs.TaskPrestartResponse) error {
+func (h *envoyVersionHook) Prestart(_ context.Context, request *ifs.TaskPrestartRequest, _ *ifs.TaskPrestartResponse) error {
 	// First interpolation of the task image. Typically this turns the default
 	// ${meta.connect.sidecar_task} into envoyproxy/envoy:v${NOMAD_envoy_version}
 	// but could be a no-op or some other value if so configured.
@@ -73,7 +73,6 @@ func (h *envoyVersionHook) Prestart(_ context.Context, request *ifs.TaskPrestart
 	// - task is a connect sidecar or gateway
 	// - task image needs ${NOMAD_envoy_version} resolved
 	if h.skip(request) {
-		response.Done = true
 		return nil
 	}
 
@@ -95,7 +94,6 @@ func (h *envoyVersionHook) Prestart(_ context.Context, request *ifs.TaskPrestart
 	// Set the resulting image.
 	h.logger.Trace("setting task envoy image", "image", image)
 	request.Task.Config["image"] = image
-	response.Done = true
 	return nil
 }
 
--- a/client/allocrunner/taskrunner/envoy_version_hook_test.go
+++ b/client/allocrunner/taskrunner/envoy_version_hook_test.go
@@ -14,7 +14,7 @@ import (
 	"github.com/hashicorp/nomad/helper/testlog"
 	"github.com/hashicorp/nomad/nomad/mock"
 	"github.com/hashicorp/nomad/nomad/structs"
-	"github.com/stretchr/testify/require"
+	"github.com/shoenig/test/must"
 )
 
 var (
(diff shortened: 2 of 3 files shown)


client/allocrunner/taskrunner/envoy_version_hook.go
```diff
@@ -95,7 +108,18 @@ func (h *envoyVersionHook) Prestart(_ context.Context, request *ifs.TaskPrestart
...
 	h.logger.Trace("setting task envoy image", "image", image)
 	request.Task.Config["image"] = image
-	response.Done = true
+	h.done = true
+	return nil
+}
+
+// PreKilling implements the ifs.TaskPreKillHook interface.
+//
+// When a sidecar task is restarting, it is possible a previous non-destructive
```
tgross, line 117: "is restarting" seems a little ambiguous here because I think we'd only get `PreKilling` if the task has been intentionally restarted? (Or restarted because a separate main task failed, too, IIRC.) If the task has exited on its own and is getting restarted because of that, we're keeping `h.done = true` in place because we don't want to do the expensive request to Consul again.

That being said, is it possible a non-destructive update could change the image modification for that case as well? Or in other words, would it make more sense not to track `h.done` at all and take the expense of re-running the request to Consul every time?

client/allocrunner/taskrunner/envoy_version_hook_test.go
```diff
@@ -422,15 +422,72 @@ func TestTaskRunner_EnvoyVersionHook_Prestart_error(t *testing.T) {
...
+	must.True(t, h.done)
+	must.Eq(t, "envoyproxy/envoy:v1.15.0", request.Task.Config["image"])
+
+	// Overwrite the previously modified image. This is the same behaviour that
+	// occurs when the server sends a non-destructive allocation update.
+	request.Task.Config["image"] = "${meta.connect.sidecar_image}"
+
+	// Run the PreKilling function which gets called when an allocation or the
+	// task is being restarted.
+	must.NoError(t, h.PreKilling(nil, nil, nil))
```
tgross, line 483: Similarly to my comment above, we might want this test to exercise a non-requested task kill too.

tgross (commented):
In terms of "what it says on the tin", this looks great @jrasell. I've left a comment about unrequested restarts though, which I think is another case where this will come up.

client/allocrunner/taskrunner/envoy_version_hook.go
schmichael: > Or in other words, would it make more sense not to track h.done at all and take the expense of re-running the request to Consul every time?

I'd be +1 on this.

I think we could modify the `skip()` func to mark `response.Done` for tasks we know will never need this hook to be done (eg non-docker non-sidecar tasks), but otherwise let's just always run it? :shrug:

jrasell: I have ripped out the internal tracking, which keeps it all a lot simpler.

client/allocrunner/taskrunner/envoy_version_hook_test.go
jrasell: I have removed the addition of PreKilling as we are not tracking done anymore and made a slight modification to the test which mimics multiple calls.

tgross (approved):
LGTM!

Issue #12314, opened 2022-03-17
E2E: `TestNodeDrainEphermeralMigrate` should work with client ACLs


tgross:
Our E2E environment doesn't have `acls { enabled = true }` on the clients, just the servers. If this setting is enabled, the `TestNodeDrainEphermeralMigrate` fails because the client-to-client ACL check fails and the data isn't migrated. We need to figure out the right client ACL policy to support this. This predates #12267 but I noticed it while working on it.

Note that failure happens _silently_. The old allocation is simply stopped and the data isn't migrated and the user isn't warned of this, because ephemeral disk migration is best effort. https://github.com/hashicorp/nomad/pull/12267 includes a commit to make the test failure more legible in this case.

tgross:
Example error on the client:
> 2023-04-06T15:12:28.346Z [WARN]  client.alloc_runner.runner_hook.migrate_disk: error migrating data from previous alloc: alloc_id=c2c411f6-942f-0977-62fb-7911d96e4cf3 error="error getting snapshot from previous alloc \"b36b3f11-ebfa-a568-27c6-3444c391baa2\": Unexpected response code: 403 (Permission denied)"

PR #16826, opened 2023-04-07
ephemeral disk: `migrate` should imply `sticky`

tgross:
The `ephemeral_disk` block's `migrate` field allows for best-effort migration of the ephemeral disk data to new nodes. The documentation says the `migrate` field is only respected if `sticky=true`, but in fact if client ACLs are not set the data is migrated even if `sticky=false`.

The existing behavior when client ACLs are disabled has existed since the early implementation, so "fixing" that case now would silently break backwards compatibility. Additionally, having `migrate` not imply `sticky` seems nonsensical: it suggests that if we place on a new node we migrate the data but if we place on the same node, we throw the data away!

Update so that `migrate=true` implies `sticky=true` as follows:

* The failure mode when client ACLs are enabled comes from the server not passing along a migration token. Update the server so that the server provides a migration token whenever `migrate=true` and not just when `sticky=true` too.
* Update the scheduler so that `migrate` implies `sticky`.
* Update the client so that we check for `migrate || sticky` where appropriate.
* Refactor the E2E tests to move them off the old framework and make the intention of the test more clear.

---

Fixes https://github.com/hashicorp/nomad/issues/12314

E2E test results:

```
$ go test -v -count=1 ./nodedrain
=== RUN   TestNodeDrain
=== RUN   TestNodeDrain/IgnoreSystem
    node_drain_test.go:239: alloc has drained from node=b39d3578 after 17.453100094s
=== RUN   TestNodeDrain/EphemeralMigrate
    node_drain_test.go:239: alloc has drained from node=b39d3578 after 16.54164811s
=== RUN   TestNodeDrain/KeepIneligible
--- PASS: TestNodeDrain (50.41s)
    --- PASS: TestNodeDrain/IgnoreSystem (25.99s)
    --- PASS: TestNodeDrain/EphemeralMigrate (22.94s)
    --- PASS: TestNodeDrain/KeepIneligible (1.26s)
PASS
ok      github.com/hashicorp/nomad/e2e/nodedrain        50.413s
```

### DIFF
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -10635,11 +10635,7 @@ func (a *Allocation) ShouldMigrate() bool {
 	}
 
 	// We won't migrate any data if the user hasn't enabled migration
-	if !tg.EphemeralDisk.Migrate {
-		return false
-	}
-
-	return true
+	return tg.EphemeralDisk.Migrate
 }
 
 // SetEventDisplayMessages populates the display message if its not already set,
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -10634,9 +10634,8 @@ func (a *Allocation) ShouldMigrate() bool {
 		return false
 	}
 
-	// We won't migrate any data is the user hasn't enabled migration or the
-	// disk is not marked as sticky
-	if !tg.EphemeralDisk.Migrate || !tg.EphemeralDisk.Sticky {
+	// We won't migrate any data if the user hasn't enabled migration
+	if !tg.EphemeralDisk.Migrate {
 		return false
 	}
 
--- a/scheduler/generic_sched.go
+++ b/scheduler/generic_sched.go
@@ -789,7 +789,8 @@ func updateRescheduleTracker(alloc *structs.Allocation, prev *structs.Allocation
 
 // findPreferredNode finds the preferred node for an allocation
 func (s *GenericScheduler) findPreferredNode(place placementResult) (*structs.Node, error) {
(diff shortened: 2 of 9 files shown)


nomad/structs/structs.go
```diff
@@ -10634,9 +10634,8 @@ func (a *Allocation) ShouldMigrate() bool {
 		return false
 	}
 
-	// We won't migrate any data is the user hasn't enabled migration or the
-	// disk is not marked as sticky
-	if !tg.EphemeralDisk.Migrate || !tg.EphemeralDisk.Sticky {
+	// We won't migrate any data if the user hasn't enabled migration
+	if !tg.EphemeralDisk.Migrate {
```
schmichael, line 10638: `return tg.EphemeralDisk.Migrate` works here, but nbd

scheduler/generic_sched.go
```diff
@@ -789,7 +789,8 @@ func updateRescheduleTracker(alloc *structs.Allocation, prev *structs.Allocation
 
 // findPreferredNode finds the preferred node for an allocation
 func (s *GenericScheduler) findPreferredNode(place placementResult) (*structs.Node, error) {
-	if prev := place.PreviousAllocation(); prev != nil && place.TaskGroup().EphemeralDisk.Sticky {
+	if prev := place.PreviousAllocation(); prev != nil &&
```
schmichael, line 792: I wouldn't mind breaking this up into multiple `if`s with an early return for things like `prev == nil`, but nbd.

schmichael (approved):
Looks great!

Issue #2052, opened 2016-12-01
Need a way to cleanly shut down nodes


blalor:
Nomad v0.5.0

There doesn't appear to be a way to cleanly shut down a client node in a way that allows allocations to be moved to other nodes and also accounts for the data in sticky ephemeral disks to be migrated.  I wrote a script to help my systemd service delay stopping the service until allocations have been moved, but there doesn't appear to be a way to monitor for the status of the migrated data.  If the data's not moved quickly, it could be lost when the node shuts down.

Something like `nomad shutdown` that blocks until the agent is completely idle would be ideal.

groggemans:
Consul has the `leave` command. Would be nice to have a similar command in nomad, which would trigger a node drain, wait for it to complete, and then gracefully leave the cluster.

preetapan:
@groggemans Nomad 0.8 added advanced node draining features. Some useful links:

[Node drain command](https://www.hashicorp.com/blog/advanced-node-draining-in-hashicorp-nomad)
[Blog post that explains node draining features](https://www.hashicorp.com/blog/advanced-node-draining-in-hashicorp-nomad)

groggemans:
I know, and it solves/implements te draining part, but then the node should still gracefully leave the cluster. And I think the only way to do this now is by stopping/interrupting the service (with `leave_on_terminate = true` or `leave_on_interrupt = true`).

Setting `leave_on_interrupt` or `leave_on_terminate` to `true` isn't always desirable, but it should still be possible to do a graceful leave from the cli even when both options are `false` (default).

For servers there's the `force-leave` option, but for clients there's no command to do a graceful leave. A universal `leave` that would work for both servers and clients which also triggers a node drain seems to be missing.

onlyjob:
A relevant discussion happened in #4305. @insanejudge, @schmichael.

Indeed draining node on shutdown is the best and service file could be adjusted to do that. However graceful restart can not be implemented in systemd service because it can not distinguish shut down from restart. Regardless, `KillMode=control-group` (default) is better than `KillMode=process` because the latter does not guarantee cleanup. It is important to leave no unmanaged processes behind.

dcparker88:
Not sure if this is relevant or related - but even when I have `leave_on_terminate` set in my config - it doesn't seem to fully leave. I've been doing some testing with the stuff above, and I can see in my logs the node is cleanly shutting down:

```
nomad: ==> Caught signal: terminated
nomad: ==> Gracefully shutting down agent...
[...]

PR #16827, opened 2023-04-07
client: allow `drain_on_shutdown` configuration

tgross:
Fixes https://github.com/hashicorp/nomad/issues/2052

Adds a new configuration to clients to optionally allow them to drain their workloads on shutdown. The client sends the `Node.UpdateDrain` RPC targeting itself and then monitors the drain state as seen by the server until the drain is complete or the deadline expires.

---

In addition to the unit tests here, I've run a bench test. Clients have the following config:

```hcl
leave_on_interrupt = true

client {
  drain_on_shutdown {
    deadline = "20s"
    ignore_system_jobs = true
  }
```

Then I run a service job and a system job.

In the first test the `service_job` has a 5s `kill_timeout`. Then I stopped the node where the service job is running. The alloc takes a little over 5s to be stopped, and then the client immediately exits because the drain strategy is nil. But because the client has exited by that point, the node's drain isn't marked complete by the client status update until the full 20s deadline expires. This is a little quirky but I think it's what we want because the local node can't decide the drain is complete on its own (because it depends on migration health).

```
Time                       Subsystem  Message
2023-04-13T16:52:26-04:00  Drain      Node drain complete
2023-04-13T16:52:05-04:00  Drain      Node drain strategy set
```

In the second test the `service_job` has a 1m `kill_timeout`. This respects the deadline as I'd expect and the node status gets updated.

```
Time                       Subsystem  Message
2023-04-13T16:54:56-04:00  Drain      Node drain complete
2023-04-13T16:54:35-04:00  Drain      Node drain strategy set
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -761,8 +761,12 @@ func (c *Client) Reload(newConfig *config.Config) error {
 
 // Leave is used to prepare the client to leave the cluster
 func (c *Client) Leave() error {
-	// TODO
-	return nil
+	if c.GetConfig().DevMode {
+		return nil
+	}
+
+	// In normal mode optionally drain the node
+	return c.DrainSelf()
 }
 
 // GetConfig returns the config of the client. Do *not* mutate without first
--- a/client/drain.go
+++ b/client/drain.go
@@ -1,0 +1,163 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/nomad/client/config"
+	"github.com/hashicorp/nomad/helper"
+	"github.com/hashicorp/nomad/nomad/structs"
+)
+
+func (c *Client) DrainSelf() error {
+	drainSpec := c.GetConfig().Drain
+	if drainSpec == nil {
+		return nil
+	}
+
+	logger := c.logger.Named("drain")
+
+	now := time.Now()
+	drainReq := &structs.NodeUpdateDrainRequest{
+		NodeID: c.NodeID(),
+		DrainStrategy: &structs.DrainStrategy{
+			DrainSpec: structs.DrainSpec{
+				Deadline:         drainSpec.Deadline,
+				IgnoreSystemJobs: drainSpec.IgnoreSystemJobs,
+			},
+			StartedAt: now,
+		},
+		MarkEligible: false,
+		Meta:         map[string]string{"message": "shutting down"},
+		WriteRequest: structs.WriteRequest{
+			Region: c.Region(), AuthToken: c.secretNodeID()},
+	}
+	if drainSpec.Deadline > 0 {
+		drainReq.DrainStrategy.ForceDeadline = now.Add(drainSpec.Deadline)
+	}
+
+	var drainResp structs.NodeDrainUpdateResponse
+	err := c.RPC("Node.UpdateDrain", drainReq, &drainResp)
+	if err != nil {
+		return err
+	}
+
+	// note: the default deadline is 1hr but could be set to "". letting this
+	// run forever seems wrong but init system (ex systemd) will almost always
+	// force kill the client eventually
+	ctx := context.Background()
+	var cancel context.CancelFunc
+	if drainSpec.Deadline > 0 {
+		// if we set this context to the deadline, the server will reach the
+		// deadline but not get a chance to record it before this context
+		// expires, resulting in spurious errors. So extend the deadline here by
+		// a few seconds
+		ctx, cancel = context.WithTimeout(context.Background(), drainSpec.Deadline+(5*time.Second))
+		defer cancel()
+	}
+	statusCheckInterval := time.Second
+
+	logger.Info("monitoring self-drain")
+	err = c.pollServerForDrainStatus(ctx, statusCheckInterval)
+	switch err {
+	case nil:
+		logger.Debug("self-drain complete")
+		return nil
+	case context.DeadlineExceeded, context.Canceled:
+		logger.Error("self-drain exceeded deadline")
+		return fmt.Errorf("self-drain exceeded deadline")
+	default:
+		logger.Error("could not check node status, falling back to local status checks", "error", err)
+	}
+
+	err = c.pollLocalStatusForDrainStatus(ctx, statusCheckInterval, drainSpec)
+	if err != nil {
+		return fmt.Errorf("self-drain exceeded deadline")
+	}
+
+	logger.Debug("self-drain complete")
+	return nil
+}
+
+// pollServerForDrainStatus will poll the server periodically for the client's
+// drain status, returning an error if the context expires or get any error from
+// the RPC call. If this function returns nil, the drain was successful.
+func (c *Client) pollServerForDrainStatus(ctx context.Context, interval time.Duration) error {
+	timer, stop := helper.NewSafeTimer(0)
+	defer stop()
+
+	statusReq := &structs.NodeSpecificRequest{
+		NodeID:   c.NodeID(),
+		SecretID: c.secretNodeID(),
+		QueryOptions: structs.QueryOptions{
+			Region: c.Region(), AuthToken: c.secretNodeID()},
+	}
+	var statusResp structs.SingleNodeResponse
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-timer.C:
+			err := c.RPC("Node.GetNode", statusReq, &statusResp)
+			if err != nil {
+				return err
+			}
+			if &statusResp != nil && statusResp.Node.DrainStrategy == nil {
+				return nil
+			}
+			timer.Reset(interval)
+		}
+	}
+}
+
+// pollLocalStatusForDrainStatus polls the local allocrunner state periodicially
+// for the client status of all allocation runners, returning an error if the
+// context expires or get any error from the RPC call. If this function returns
+// nil, the drain was successful. This is a fallback function in case polling
+// the server fails.
+func (c *Client) pollLocalStatusForDrainStatus(ctx context.Context,
+	interval time.Duration, drainSpec *config.DrainConfig) error {
+
+	// drainIsDone is its own function scope so we can release the allocLock
+	// between poll attempts
+	drainIsDone := func() bool {
+		c.allocLock.RLock()
+		defer c.allocLock.RUnlock()
+		for _, runner := range c.allocs {
+
+			// note: allocs in runners should never be nil or have a nil Job but
+			// if they do we can safely assume the runner is done with it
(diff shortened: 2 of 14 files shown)


client/drain.go
```diff
@@ -0,0 +1,80 @@
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			logger.Error("self-drain exceeded deadline")
+			return ctx.Err()
+		case <-timer.C:
+			err := c.RPC("Node.GetNode", statusReq, statusResp)
+			if err != nil {
+				logger.Error("could not check node status", "error", err)
```
schmichael, line 73: We really need our RPC layer to support cancellation, but I think it requires adopting a fork of `net/rpc` which way too much scope to add for this effort. :sob: 

Even without that I think we could do a couple improvements:

1. On `ctx.Done()` try the RPC once more to avoid spurious error logs when everything actually worked as desired (since the 10s wait leaves a pretty sizable window for the deadline to be hit before we do another check).
   - If we're worried about an additional RPC blocking for too long, what if we just checked to see if there are any non-terminal allocations locally? We'd have to manually implement the `ignore system job` logic here as well which is a bit gross, but I love the idea of letting Clients make _local_ decisions.
2. Disconnected nodes without a deadline will never shutdown cleanly. I think we need some escape hatch like `if 10 minutes have passed without a successful RPC, exit`. I would just hate large clusters with fairly automated operations to have orphaned nodes hanging around because they never expected them to block forever.

That being said I'm not sure 10 minutes is sufficiently low to be safe. We could just rely on the user configuring their init system to force kill Nomad after a timeout of their choosing. The docs for this configuration block would be a natural place to call that out.

client/client.go
```diff
@@ -859,7 +859,9 @@ func (c *Client) Shutdown() error {
 			arGroup.AddCh(ar.DestroyCh())
 		}
 	} else {
-		// In normal mode call shutdown
+		// In normal mode optionally drain the node before shutting down any
+		// remaining alloc runners.
+		c.DrainSelf()
```
schmichael, line 864: As we discussed offline while I think `cancel the drain` will be a great escape hatch to document for users, I'm not sure it's sufficient for the inplace Nomad upgrade case.

I'm assuming there are 2 common reasons for shutting down a Nomad agent:
1. Upgrading the `nomad` binary
2. Decommissioning a node

(I think there are other reasons like "something is broken and turning things off and on again is always worth a shot" -- but _hopefully_ that's far less common than the 2 above.)

This behavior suits `#2` perfectly, and manually cancelling a drain to speed up `#2` is totally reasonable.

However for `#1` (upgrading Nomad), not only does it require an extra command/API-call (and the error handling considerations that goes with any operation), but it also incurs 2 useless Raft writes per Node (start the drain / stop the drain). For large clusters looking to roll clients as quickly as possible that could be kind of annoying as restarts likely already increase Leader load when they re-register and sync allocs on startup.

So it'd be nice to tell users `When decommissioning use SIGINT and when upgrading use SIGTERM` or similar.

Windows will require an API (and associated `nomad node shutdown ...` command which would be pretty cool), but I think we can leave that to a future PR.

schmichael (commented):
This approach looks great! I'm almost to saying :shipit: as-is, but the inplace upgrade case I outline warrants some additional consideration.

client/client.go
tgross: Aye, I can get something wired into the signal handler for this.

client/drain.go
tgross: I really like the idea of having the logic all be local and it _mostly_ solves problem (2) as well. We don't retry the initial `Node.UpdateDrain`, so if we're already disconnected at the time we shutdown that's fine -- we just stop the allocrunners and leave the tasks up as usual (which allows our disconnected clients logic to work as well). So long as we don't disconnect in the narrow window between when we call `Node.UpdateDrain` and the first `Client.GetAllocs` call, we don't need to talk to the server any more after that. So instead of calling `Node.GetNode` we can poll the alloc runners with a check like:

```go
check := func() {
	c.allocLock.RLock()
	defer c.allocLock.RUnlock()
	for _, runner := range c.allocs {
		// ...
	}
}
```

The ignore-system logic is pretty trivial.

client/client.go
tgross: @schmichael I dug into this and remembered that we've got an existing facility for graceful exit with [`leave_on_interrupt`](https://developer.hashicorp.com/nomad/docs/configuration#leave_on_interrupt) and [`leave_on_terminate`](https://developer.hashicorp.com/nomad/docs/configuration#leave_on_terminate) on either SIGINT or SIGTERM, which for clients does not actually do anything today because it ends up [here](https://github.com/hashicorp/nomad/blob/main/client/client.go#L759-L763
):

```go
// Leave is used to prepare the client to leave the cluster
func (c *Client) Leave() error {
	// TODO
	return nil
}
```

So I think we can move the call to `DrainSelf` into the `Leave` method, and that'll let the existing `leave_on_interrupt` and `leave_on_terminate` configurations actually do something for clients. Cluster adminstrators get their pick of which signal they want to use for graceful exits with drain, and if neither are set we don't drain at all.

tgross: That `TODO` is from 2015! Finally done! :grinning:

schmichael: :facepalm: Wow I had totally forgotten about those flags. Fantastic work.

client/drain.go
```diff
@@ -0,0 +1,88 @@
...
+	var drainResp *structs.NodeDrainUpdateResponse
+	err := c.RPC("Node.UpdateDrain", drainReq, drainResp)
+	if err != nil {
+		logger.Error("failed to drain self on shutdown", "error", err)
+		return err
+	}
+
+	// note: letting this run forever seems wrong but init system (ex systemd)
+	// will almost always force kill the client eventually
+	ctx := context.Background()
```
tgross, line 47: Of course, we could always cap this to something really long like 24h, so long as we document that behavior in the config docs.

client/drain.go
tgross: @schmichael implementing local-only behavior as we discussed here was easy but then I realized it's easy because it doesn't work. :grinning: 

We only remove an allocrunner from `c.allocs` when the client has been notified by the server that we've GC'd on the server (ref [`client.go#L2540-L2541`](https://github.com/hashicorp/nomad/blob/v1.5.3/client/client.go#L2540-L2541)). We also can't simply track the local client status because ephemeral disk migration relies on receiving the client status update to the server before making the data migration HTTP call, so we're supposed to be waiting on that before considering the drain complete.

Given that all the above is waiting for external requests (either server-to-client RPC or client-to-client HTTP), maybe we poll the server for drain status as I did originally, and then if we _can't_ poll the server because we're disconnected, we treat that as being safe to fall back to checking local `ClientStatus` of the allocs.

nomad/fsm.go
```diff
@@ -473,10 +473,19 @@ func (n *nomadFSM) applyDrainUpdate(reqType structs.MessageType, buf []byte, ind
 			return fmt.Errorf("error looking up ACL token: %v", err)
 		}
 		if token == nil {
-			n.logger.Error("token did not exist during node drain update")
-			return fmt.Errorf("token did not exist during node drain update")
+			node, err := n.state.NodeBySecretID(nil, req.AuthToken)
```
schmichael, line 476: We check auth tokens in the FSM?! Well that's not going to work with Workload Identities, but obviously that is not a problem for this PR.

schmichael (approved):

nomad/fsm.go
tgross: The only reason we do this is to record the accessor ID in the drain events, so we can totally fix this in a later PR by passing the [`AuthenticatedIdentity.String()`](https://github.com/hashicorp/nomad/blob/v1.5.3/nomad/structs/structs.go#L510-L524) value in the arguments somewhere. I just didn't want to go messing with that in this PR

tgross:
@schmichael before I merge this:
* do you have an opinion on whether we should set the deadline to 1h by default so that it matches the `nomad node drain` command?
* should we support the equivalent of the `nomad node drain -force` flag?:

> - `-force`: Remove allocations off the node immediately, regardless of the allocation's `migrate` block.

website/content/docs/configuration/client.mdx
```diff
@@ -602,6 +607,41 @@ client {
...
+[`TimeoutStopSec`][] value in the `nomad.service` unit file to account for
+enough time for the client to drain.
+
+```hcl
+client {
+  drain_on_shutdown {
+    deadline = ""
+    ignore_system_jobs = false
+  }
+}
```
lgfa29, line 634:
```suggestion
client {
  # Either leave_on_interrupt or leave_on_terminate must for this to take effect.
  drain_on_shutdown {
    deadline = ""
    ignore_system_jobs = false
  }
}
```

Or something like this. It took me a bit to realize why my example was not working.

website/content/docs/configuration/client.mdx
```diff
@@ -602,6 +607,41 @@ client {
...
+```
+
+- `deadline` `(string: "")` - Set the deadline by which all allocations must be
+  moved off the node. Remaining allocations after the deadline are removed from
+  the node, regardless of their [`migrate`][] block. Defaults to no deadline.
```
lgfa29, line 639:
```suggestion
  moved off the client. Remaining allocations after the deadline are removed from
  the client, regardless of their [`migrate`][] block. Defaults to no deadline.
```

I don't remember if we reached a consensus on the terminology here 😅 

Feel free to ignore it.

lgfa29 (approved):
Not to give you more work, but I think it would be good to have some docs (maybe in the [`Upgrade Process`](https://developer.hashicorp.com/nomad/docs/upgrade#upgrade-process)?) about the in-place upgrade when `drain_on_shutdown` is set.

schmichael:
> * do you have an opinion on whether we should set the deadline to 1h by default so that it matches the `nomad node drain` command?

Matching defaults makes sense to me. I can't think of a reason _not_ to.

> * should we support the equivalent of the `nomad node drain -force` flag?:

Seems dangerous, but if you're shutting down a node it's still safer than not draining at all, so ... sure I guess? Again I can't think of a strong reason not to as long as we're clear it can cause an outage (but no _worse_ of an outage than if they didn't use `drain_on_shutdown` _at all._

website/content/docs/configuration/client.mdx
tgross: Done!

website/content/docs/configuration/client.mdx
tgross: Good call

client/drain.go
tgross: Switched to 1hr.

PR #16835, opened 2023-04-10
docs: document signal handling

tgross:
Expand documentation about Nomad's signal handling behaviors, including removing incorrect information about graceful client shutdowns.

---

* preview link: https://nomad-9cb2o1urm-hashicorp.vercel.app/nomad/docs/operations/nomad-agent#stopping-an-agent
* note: I'm having this PR up separately from https://github.com/hashicorp/nomad/pull/16827 which changes the behavior because I want these docs backported whereas #16827 will only land in 1.5.x

### DIFF
--- a/website/content/docs/operations/nomad-agent.mdx
+++ b/website/content/docs/operations/nomad-agent.mdx
@@ -103,6 +103,7 @@ In addition to the optional handling of interrupt (`SIGINT`) and terminate
 signals (`SIGTERM`) described in [Stopping an Agent][#stopping-an-agent], Nomad
 supports special behavior for several other signals useful for debugging.
 
+* `SIGHUP` will cause Nomad to [reload its configuration][].
 * `SIGUSR1` will cause Nomad to print its [metrics][] without stopping the
   agent.
 * `SIGQUIT`, `SIGILL`, `SIGTRAP`, `SIGABRT`, `SIGSTKFLT`, `SIGEMT`, or `SIGSYS`
@@ -178,4 +179,5 @@ owned by `root` with filesystem permissions set to `0700`.
 [`leave_on_terminate`]: /nomad/docs/configuration#leave_on_terminate
 [`server force-leave` command]: /nomad/docs/commands/server/force-leave
 [consensus]: /nomad/docs/concepts/consensus
+[reload its configuration]: /nomad/docs/configuration#configuration-reload
 [metrics]: /nomad/docs/operations/metrics-reference
--- a/website/content/docs/operations/nomad-agent.mdx
+++ b/website/content/docs/operations/nomad-agent.mdx
@@ -82,22 +82,44 @@ There are several important messages that `nomad agent` outputs:
 
 ## Stopping an Agent
 
-An agent can be stopped in two ways: gracefully or forcefully. By default,
-any signal to an agent (interrupt, terminate, kill) will cause the agent
-to forcefully stop. Graceful termination can be configured by either
-setting `leave_on_interrupt` or `leave_on_terminate` to respond to the
+An agent can be stopped in two ways: gracefully or forcefully. By default, any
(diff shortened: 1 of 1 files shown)


website/content/docs/operations/nomad-agent.mdx
```diff
@@ -82,22 +82,44 @@ There are several important messages that `nomad agent` outputs:
...
-to forcefully stop. Graceful termination can be configured by either
-setting `leave_on_interrupt` or `leave_on_terminate` to respond to the
+An agent can be stopped in two ways: gracefully or forcefully. By default, any
+stop signal to an agent (interrupt, terminate, kill) will cause the agent to
+forcefully stop. Graceful termination can be configured by either setting
+[`leave_on_interrupt`][] or [`leave_on_terminate`][] to respond to the
 respective signals.
 
-When gracefully exiting, clients will update their status to terminal on
-the servers so that tasks can be migrated to healthy agents. Servers
```
tgross, line 92: The behavior described here for clients has never been the case. Only the node GC job ever calls `Node.Deregister` and the client calls `Node.UpdateStatus` only to heartbeat (i.e. status is only ever "ready"). 

Arguably if we have it documented here that makes the lack of it a bug which we should fix and backport. I'd definitely be open to that idea, seeing as how you have to opt-in to `leave_on_interrupt` or `leave_on_terminate` anyways so that's not going to break backwards compatibility unexpectedly.

website/content/docs/operations/nomad-agent.mdx
```diff
@@ -82,22 +82,44 @@ There are several important messages that `nomad agent` outputs:
...
+
+## Signal Handling
+
+In addition to the optional handling of interrupt (`SIGINT`) and terminate
+signals (`SIGTERM`) described in [Stopping an Agent][#stopping-an-agent], Nomad
+supports special behavior for several other signals useful for debugging.
+
+* `SIGUSR1` will cause Nomad to print its [metrics][] without stopping the
+  agent.
+* `SIGQUIT`, `SIGILL`, `SIGTRAP`, `SIGABRT`, `SIGSTKFLT`, `SIGEMT`, or `SIGSYS`
```
gulducat, line 108: I see SIGTERM as mentioned above, and SIGHUP which isn't mentioned, here:

https://github.com/hashicorp/nomad/blob/v1.5.3/command/agent/command.go#L890

for the `agent` command signal handling, but I'm not finding these other signals handled anywhere explicitly..  Am I missing something obvious, or something implicit that I'm not aware of?

tgross: The `SIGUSR1` behavior is provided by the go-metrics library (see [`inmem_signal.go`](https://github.com/hashicorp/go-metrics/blob/master/inmem_signal.go), which I kinda don't _love_ but :shrug: ), and the signals listed in the 2nd bullet point are covered by the Go runtime. See [Default behavior of signals in Go programs](https://pkg.go.dev/os/signal#hdr-Default_behavior_of_signals_in_Go_programs). I'm not documenting the system default handlers for things like job control because that doesn't seem all that useful?

tgross: Add a bullet point for `SIGHUP` with a link to where we define that behavior in more detail.

gulducat (approved):
LGTM! 📶

website/content/docs/operations/nomad-agent.mdx
schmichael: Sorry I missed the review, but I would be +1 on rewriting this paragraph altogether when your drain PRs land. I find these to be surprising and unhelpful definitions of "graceful" and "forceful." I suspect the only reason these wrong docs lived so long is that they're so wrong and unhelpful it wasn't clear whether they or Nomad needed fixing as you pointed out.

tgross: Will do!

PR #16841, opened 2023-04-11
e2e: add e2e tests for job submission api

shoenig:
This PR adds e2e tests around the new job submissions api, added in https://github.com/hashicorp/nomad/pull/16763

No backport, feature is in 1.6

### DIFF
--- a/e2e/e2eutil/allocs.go
+++ b/e2e/e2eutil/allocs.go
@@ -70,7 +70,7 @@ func WaitForAllocStatusComparison(query func() ([]string, error), comparison fun
 }
 
 // SingleAllocID returns the ID for the first allocation found for jobID in namespace
-// at the specified job version number. Will retry for a ten seconds before returning
+// at the specified job version number. Will retry for ten seconds before returning
 // an error.
 //
 // Should only be used with jobs containing a single task group.


e2e/e2eutil/allocs.go
```diff
@@ -65,11 +69,38 @@ func WaitForAllocStatusComparison(query func() ([]string, error), comparison fun
 	return err
 }
 
+// SingleAllocID returns the ID for the first allocation found for jobID in namespace
+// at the specified job version number. Will retry for a ten seconds before returning
```
lgfa29, line 73:
```suggestion
// at the specified job version number. Will retry for ten seconds before returning
```

e2e/jobsubmissions/input/xyz.hcl
```diff
@@ -0,0 +1,36 @@
...
+}
+
+variable "Y" {
+  type = number
+}
+
+variable "Z" {
+  type = bool
+}
+
```
lgfa29, line 12: Not sure if it's relevant as a test case here, but there's also a shorter version for declaring variables:

```hcl
variables {
  A = "B"
}
```

shoenig: Yeah these kinds of variables that have the value set directly in the original job are "invisible", at least not without doing some work with the hcl parser.

PR #16843, opened 2023-04-11
consul/connect: fixed a bug where restarting proxy tasks failed.

jrasell:
manual backport of #16815 

The first start of a Consul Connect proxy sidecar triggers a run of the envoy_version hook which modifies the task config image entry. The modification takes into account a number of factors to correctly populate this. Importantly, once the hook has run, it marks itself as done so the taskrunner will not execute it again.

When the client receives a non-destructive update for the allocation which the proxy sidecar is a member of, it will update and overwrite the task definition within the taskerunner. In doing so it overwrite the modification performed by the hook. If the allocation is restarted, the envoy_version hook will be skipped as it previously marked itself as done, and therefore the sidecar config image is incorrect and causes a driver error.

The fix removes the hook in marking itself as done to the view of the taskrunner.

### DIFF
--- a/.changelog/16815.txt
+++ b/.changelog/16815.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+client: Fixed a bug where restarting proxy sidecar tasks failed
+```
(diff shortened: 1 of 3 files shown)


pkazmierczak (approved):

tgross (approved):
LGTM

PR #16845, opened 2023-04-11
drainer: fix codec race condition in integration test

tgross:
While working on #14348 I encountered corruption in the RPC decoding in the drainer integration test.

msgpackrpc codec handles are specific to a connection and cannot be shared between goroutines; this can cause corrupted decoding. Fix the drainer integration test so that we create separate codecs for the goroutines that the test helper spins up to simulate client updates.

This changeset also refactors the drainer integration test to bring it up to current idioms and library usages, make assertions more clear, and reduce duplication. I'm backporting this work for consistency between release branches but there's no changelog because this is a test-only PR.

### DIFF
--- a/nomad/drainer_int_test.go
+++ b/nomad/drainer_int_test.go
@@ -6,15 +6,16 @@ package nomad
 import (
 	"context"
 	"fmt"
-	"net/rpc"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


shoenig (approved):
LGTM!

PR #16861, opened 2023-04-12
[ui, compliance] Remove the newline after .hbs copyright headers

philrenaud:
We were seeing test failures in main related to log-content equivalence tests, as well as some flakey percy vertical spacing issues, that seem to point to the newline after copyright headers. This PR attempts to remove that newline across all our .hbs files in the ui, to see if those problems are resolved.

image
image

### DIFF
--- a/ui/app/components/allocation-service-sidebar.hbs
+++ b/ui/app/components/allocation-service-sidebar.hbs
@@ -1,7 +1,8 @@
 {{!
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
-}}
+~}}
+
 <div
   class="sidebar has-subnav service-sidebar {{if this.isSideBarOpen "open"}}"
   {{on-click-outside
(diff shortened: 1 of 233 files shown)


lgfa29:
Have you tested with `~}}` for [white-space control](https://handlebarsjs.com/guide/expressions.html#whitespace-control)? I'm afraid that without fixing the bot we will keep running into this problem and having to manually fix new files.

tgross:
> I'm afraid that without fixing the bot we will keep running into this problem and having to manually fix new files.

It looks like we can run [`copywrite`](https://github.com/hashicorp/copywrite) locally to see if it'll fail these files again.

tgross (approved):
LGTM but worth verifying it won't get undone by the bot again

philrenaud:
> ~}}

@lgfa29 very pleased to report that the `~}}` style syntax works exactly the same as removing the newline altogether!

PR #16865, opened 2023-04-12
Backport to 1.5.x: Deployment Status Command Does Not Respect -namespace Wildcard

Juanadelacuesta:
* func: add namespace support for list deployment

* func: add wildcard to namespace filter for deployments

* Update deployment_endpoint.go

* style: use must instead of require or asseert

* style: rename paginator to avoid clash with import

* style: add changelog entry

* fix: add missing parameter for upsert jobs


tgross (approved):
LGTM!

Issue #10366, opened 2021-04-12
Safely running jobs with shared data


NomAnor:
### Proposal
Implement a configuration option to disable automatic resheduling of lost tasks and implement a way to
reshedule such allocations explicitly with a command.

### Use-cases
I'm looking to migrate some legacy applications from a pacemaker cluster to nomad.
The applications use shared storage so only one intance is allowed to run in the cluster.
This is only a small 3-node internal cluster, I mostly look at nomad for easier use to run containers and vms.

As far as I can tell nomad automatically reshedules tasks when they are marked as lost when
a client is disconnected from the servers. 

But the servers don't know if the task is still running, sheduling a new instance is fatal for the shared data.
`stop_after_client_disconnect` only works if the client agent is still running, which it might not be.

If the automatic resheduling can be disabled an operator can ensure that no instance is running and then reshedule/remove the
lost allocation which would then allow the cluster to start the instance somewhere else.

This is a manual step and would lead to an outage for that service, but in this case this is exactly what I want because the
alternative would be data corruption. Ideally there is a way to automatically ensure no running instance like the STONITH configuration in pacemaker but that might be out of scope for nomad. 

In https://github.com/hashicorp/nomad/issues/2185#issuecomment-288606154 someone mentioned a similar problem

### Attempted Solutions
I tried setting `restart.attemps = 0` and `reschedule.attemps = 0`.

schmichael:
Hi @NomAnor!

This is an interesting use case and as you point out other people have hit it. Anyone else coming across this issue please feel free to :+1: it and/or leave your use case in the comments!

As you have discovered `restart.attempts` and `reschedule.attempts` do not cover when allocations become `lost`. Also you correctly pointed out the limitation of `stop_after_client_disconnect` to only work if the Nomad client agent is still running.

Ideally the shared storage would prevent multiple writers, but I know that that's not always the case and often out of your hands.

I think what Nomad could do is add a new jobspec parameter like `group.reschedule_on_lost = true` (name TBD). It would default to `true` to maintain backward compatibility and the more common case of wanting to reschedule allocations on doown nodes.

However, when `reschedule_on_lost = false` the scheduler could treat `lost` allocations as `running` and wait for an operator to intervene. The next problem is that we lack an API to unblock allocations in that state and would need to add it, perhaps `nomad alloc reschedule` much like we have `nomad alloc restart` for local restarts.

Does that seem like it would solve your issue?

hungdqht:
[...]

PR #16867, opened 2023-04-12
Add the ability to disable rescheduling of lost tasks. Implements issue #10366

DominicLavery:
This adds the ability to prevent tasks being rescheduled when their allocations are lost as suggested in https://github.com/hashicorp/nomad/issues/10366

Jobs with lost allocations are left in the running state and require intervention. Currently, jobs can be stopped manually via the CLI or API. e.g. `nomad job stop`. A future improvement would be to implement `nomad alloc reschedule` as suggested in the issue.

### DIFF
--- a/.changelog/16867.txt
+++ b/.changelog/16867.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+**Reschedule on Lost**: Adds the ability to prevent tasks on down nodes from being rescheduled
+```
--- a/api/jobs_test.go
+++ b/api/jobs_test.go
@@ -310,8 +310,9 @@ func TestJobs_Canonicalize(t *testing.T) {
 				},
 				TaskGroups: []*TaskGroup{
 					{
-						Name:  pointerOf(""),
-						Count: pointerOf(1),
+						Name:             pointerOf(""),
+						Count:            pointerOf(1),
+						RescheduleOnLost: pointerOf(true),
 						EphemeralDisk: &EphemeralDisk{
 							Sticky:  pointerOf(false),
 							Migrate: pointerOf(false),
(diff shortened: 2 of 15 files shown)


tgross:
Hi @DominicLavery! Sorry about the delayed review on this. The issue #10366 was really still a little bit in the design stage, so there's some open questions about how this should interact with the existing `max_client_disconnect` feature, which prevents tasks from being marked as lost but replaces them temporarily until the allocation reconnects or the `max_client_disconnect` window expires. It might make sense for `reschedule_on_lost=false` to interact with that so that we don't create those replacements either (that's certainly how I'd want to use it). I'm going to have an internal discussion with the team in the next day or so and then I'll pop back over here and provide some direction.

DominicLavery:
Hey @tgross. Thanks for the response. Sorry to skip ahead in the process a bit, we had a bit of a pressing need for this kind of feature. I'd be happy to rework it based off of the results of your discussions. I agree that `max_client_disconnect` should be taken in to account here.

I've rebased (and re-enabled GH actions on the fork 😅) but I still only get the one check.

Juanadelacuesta:
He @DominicLavery! Thank you for all the work on this issue.  On reviewing the code, it all seems to be working, but on a closer inspection, there are two problems:

1. The new option `reschedule_on_lost` defaults to `true`, causing a bit of a problem for older job definitions that now need to be updated to include the new option.
2. Your PR successfully manages to avoid rescheduling of a lost allocation, but if any other event happens, including the lost node coming back online, a new reconciliation on the server is triggered and  the lost allocation is rescheduled and starts again. 

We think this is a very interesting feature and there is an internal discussion about how to move on with it.

DominicLavery:
Hey @Juanadelacuesta! Thanks for the feedback!
On point 1: rescheduling on lost is the current behaviour, so it defaults to true for that backwards compatibility. The option only needs to be added if you want to disable it. Older job definitions will work as the did previously. 

On point 2: I hadn't considered that!

Happy to work on point 2, and @tgross's point about `max_client_disconnect` once your internal discussions are finished and your happy about the path forward

roman-vynar:
Same as @Juanadelacuesta saying
```
The new option reschedule_on_lost defaults to true, causing a bit of a problem for older job definitions that now need to be updated to include the new option.
Your PR successfully manages to avoid rescheduling of a lost allocation, but if any other event happens, including the lost node coming back online, a new reconciliation on the server is triggered and the lost allocation is rescheduled and starts again.
```

It is a big problem for old jobs.
Moreover, despite `reschedule_on_lost` defaults to `true` in the code it is not `true` when you don't specify this option in the job spec. So it is for new jobs as well.

And have to say when a node comes back online an alloc may get lost and it will never get re-scheduled. 
This feature in 1.7.2 is not backward compatible with previous versions.

PR #16875, opened 2023-04-13
Demo: NFS CSI Plugins

gulducat:
Hopefully the Readme explains itself, but the goal here is to demo (and easily reproduce, locally) a CSI configuration that has separate controller and node plugins, to compliment the simple `hostpath` monolith.

This approach uses NFS in a container backed by a host volume on the node.

Hi @apollo13 I believe `rocketDuck` on GitLab is you?  I found your [Python NFS CSI plugin](https://gitlab.com/rocketduck/csi-plugin-nfs) to be pretty delightful. :)  Would love your feedback here if you'd like to chime in!

Initially I used [democratic-csi's NFS plugin](https://github.com/democratic-csi/democratic-csi/blob/v1.8.3/docs/nomad.md), which was pretty neat and did work, but ended up preferring rocketDuck's mainly because its controller plugin does not need to write directly to the host volume, so it can run anywhere that can reach NFS, more like other controller plugins.  I left the democratic-csi commit in this branch, but intend to squash it away on merge.  I may still go back and submit a PR for their Nomad setup doc, since some if it is pretty different nowadays.


demo/csi/nfs/volume.hcl
```diff
@@ -0,0 +1,9 @@
+id        = "csi-nfs"
+name      = "csi-nfs"
+type      = "csi"
+plugin_id = "rocketduck-nfs"
+
+capability {
+  access_mode     = "multi-node-multi-writer"
+  attachment_mode = "file-system"
+}
```
tgross, line 9: CSI's capabilities don't enclose each other the way you'd expect. The first job that claims the volume sets the active capability from the set of capabilities in the volume spec. Having a single `capability` means that the volume _cannot_ be mounted in a non-multi-node-multi-writer way by jobs. If your application doesn't safely support multi-writer semantics or you never want the volume to be writable, you'll probably want to be able to tell Nomad to lock that out. 

So I'd expand this to at least cover:

```suggestion
capability {
  access_mode     = "multi-node-multi-writer"
  attachment_mode = "file-system"
}

capability {
  access_mode     = "multi-node-single-writer"
  attachment_mode = "file-system"
}

capability {
  access_mode     = "multi-node-reader-only"
  attachment_mode = "file-system"
}
```

demo/csi/nfs/setup.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+
+# Set up all the demo components.
+# This can be run repeatedly as it is fairly idempotent.
+
+set -xeuo pipefail
```
tgross, line 6: Do we want `-x` still or was that from debugging?

tgross (approved):
LGTM!

demo/csi/nfs/volume.hcl
apollo13: Something new to learn every day, thanks!

apollo13:
Hi @gulducat, thanks -- this looks nice. It would be great if you could also add parameters to the example:

https://gitlab.com/rocketduck/csi-plugin-nfs/-/blob/main/nomad/example.volume#L10-17

It is imo rather common that containers expect the volume mounted under a certain uid

demo/csi/nfs/setup.sh
gulducat: Personally I like `-x` in general, but especially for demos I like to be transparent about what's actually being done, even if the script is right there to be read.  Sometimes it's a bit annoying, like with the while loops polling until next step, but overall I'm a fan.

tgross: Works for me!

demo/csi/nfs/jobs/controller-plugin.nomad.hcl
```diff
@@ -0,0 +1,32 @@
...
+        image = "registry.gitlab.com/rocketduck/csi-plugin-nfs:0.6.1"
+        args = [
+          "--type=controller",
+          "--endpoint=${CSI_ENDPOINT}", # provided by csi_plugin{}
+          "--node-id=${attr.unique.hostname}",
+          "--nfs-server=${NFS_ADDRESS}:/srv/nfs",
+          "--log-level=DEBUG",
+        ]
+        privileged   = true   # this particular controller mounts NFS in itself
+        network_mode = "host" # allows rpc.statd to work for remote NFS locking
```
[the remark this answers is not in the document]
gulducat: Oh thank you!  This must be left over from my previous implementation where I specifically encountered an rpc.statd error that this resolved, I think from democratic-csi's node plugin in particular.

Looks like here too I only need it on the node plugin, due to other tasks' mounts hanging when it's restarted as you say.  For the controller, I think it's the only user of its internal NFS mount, so it seems okay with only privileged=true and default network_mode.  Does that sound right to you?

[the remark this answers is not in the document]
apollo13: Yeah with NFSv4 you don't need a NLM anymore (I think it is handled in the kernel or so?); with NFSv3 you might have to run with `-o nolock` or similar (https://gitlab.com/rocketduck/csi-plugin-nfs/-/issues/9 is similar).

You are right that one most likely won't need `network_mode = "host"` for the controller since it is an internal mount and not "shared". That said I will probably have to double check whether the controller properly unmounts the share or leaves around stray mounts.

PR #16882, opened 2023-04-13
[ui, deployments] Conversions on long nanosecond update params

philrenaud:
Changes the nanoseconds in Update Params to seconds/mins/etc. using our formatDuration utility.

image

### DIFF
--- a/ui/app/components/job-status/update-params.js
+++ b/ui/app/components/job-status/update-params.js
@@ -48,21 +48,18 @@ export default class JobStatusUpdateParamsComponent extends Component {
   @tracked rawDefinition = null;
 
   get updateParamGroups() {
-    if (this.rawDefinition) {
-      return this.rawDefinition.TaskGroups.map((tg) => {
-        return {
-          name: tg.Name,
-          update: Object.keys(tg.Update).reduce((newUpdateObj, key) => {
-            newUpdateObj[key] = PARAMS_REQUIRING_CONVERSION.includes(key)
-              ? formatDuration(tg.Update[key])
-              : tg.Update[key];
-            return newUpdateObj;
-          }, {}),
-        };
-      });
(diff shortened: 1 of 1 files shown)


ui/app/components/job-status/update-params.js
```diff
@@ -44,7 +52,12 @@ export default class JobStatusUpdateParamsComponent extends Component {
...
-          update: tg.Update,
+          update: Object.keys(tg.Update).reduce((newUpdateObj, key) => {
+            newUpdateObj[key] = PARAMS_REQUIRING_CONVERSION.includes(key)
+              ? formatDuration(tg.Update[key])
+              : tg.Update[key];
+            return newUpdateObj;
+          }, {}),
         };
       });
     } else {
```
lgfa29, line 63: Not necessarily part of the PR, but I think this function can be simplified a bit by returning early and removing the nested `return` in the arrow function:

```js
  get updateParamGroups() {
    if (!this.rawDefinition) {
      return null;
    }

    return this.rawDefinition.TaskGroups.map((tg) => ({
      name: tg.Name,
      update: Object.keys(tg.Update).reduce((newUpdateObj, key) => {
        newUpdateObj[key] = PARAMS_REQUIRING_CONVERSION.includes(key)
          ? formatDuration(tg.Update[key])
          : tg.Update[key];
        return newUpdateObj;
      }, {}),
    }));
  }
```

But definitely a style nit-pick 😄

lgfa29 (approved):

ui/app/components/job-status/update-params.js
philrenaud: Returning early is a good suggestion / something I see often in our .go code. I'll try to implement it here and elsewhere as it comes up. Thank you!

PR #16884, opened 2023-04-13
[server] Serve API over Unix domain sockets

angrycub:
This PR pairs with #16872 to provide the server side configuration and implementation for serving the Nomad API over unix domain sockets.

For local testing, you can use `socat` to provide a suitable proxy

```bash
nomad agent -dev -api-socket-path /tmp/run/nomad.sock &
socat -d -d TCP4-LISTEN:14646,reuseaddr,fork UNIX:/tmp/run/nomad.sock
```

In another terminal

```bash
export NOMAD_ADDR=http://127.0.0.1:14646
nomad namespace list
```

Contents:
- Add api_socket config
- Update tests for new config
- Enable socket listener
- Add flags to agent for socket config


command/agent/http.go
```diff
@@ -193,6 +194,58 @@ func NewHTTPServers(agent *Agent, config *Config) ([]*HTTPServer, error) {
...
 	}
 
+	// Initialize the API socket if configured
+	startAPISocket := func() {
+		if config.APISocket.Path == "" {
+			return
+		}
+
+		const (
+			MaxSocketPathLength = 104
```
schmichael, line 204: This varies depending on OS. So I think we should just try to create the socket and return the awful OS error if we fail.

If Go can find the `sun_path` length for the OS its on, that would be amazing. I think even then we need to be careful to subtract 1 to leave space for a null terminator: otherwise the man page implies no one may be able to connect to it?! What a mess. https://man7.org/linux/man-pages/man7/unix.7.html

command/agent/http.go
```diff
@@ -193,6 +194,58 @@ func NewHTTPServers(agent *Agent, config *Config) ([]*HTTPServer, error) {
 		srvs = append(srvs, srv)
 	}
 
+	// Initialize the API socket if configured
+	startAPISocket := func() {
```
schmichael, line 198: I would prefer a top level func that takes a config and returns an error just because this func is already so huge and only readable because it's the same code over and over slight slight variations. nbd though

command/agent/http.go
```diff
@@ -193,6 +194,58 @@ func NewHTTPServers(agent *Agent, config *Config) ([]*HTTPServer, error) {
...
+			logger:       agent.httpLogger,
+			Addr:         sl.Addr().String(),
+			wsUpgrader:   wsUpgrader,
+		}
+		srv.registerHandlers(config.EnableDebug)
+
+		// Create HTTP server with timeouts
+		httpServer := http.Server{
+			Addr:      srv.Addr,
+			Handler:   handlers.CompressHandler(srv.mux),
```
schmichael, line 236: Since the unix socket lacks mTLS, I think we need to require authentication. We have a few HTTP endpoints that are unauthenticated because we consider mTLS a "good enough" baseline of access control.

It could be argued that the unix file permissions play the role of mTLS here, but (1) UDS permissions are a mess of OS-specific details and (2) now we're adding yet-another-dimension to our already highly complex auth story.

There are some options here:
1. Require auth only if ACLs are enabled. I think this would make the most sense to users and be met with little friction.
2. Require auth only if TLS (and ACLs?) are enabled. That actually matches the security properties we're going after, but I feel like it's going to get confusing for users.
3. Require auth but add a `api_unix_socket.allow_unauthenticated` option that **defaults to `false`.** The nice thing about this is that it gives us somewhere to drop the big long explanation of why we require it in the first place.

Thoughts? I really don't want to diminish the usefulness of this feature, but I also don't want to turn around have even a low sev CVE because we subtly changed our security model.

command/agent/config.go
```diff
@@ -92,6 +92,10 @@ type Config struct {
 	// Use normalizedAddrs if you need the host+port to bind to.
 	Addresses *Addresses `hcl:"addresses"`
 
+	// APISocket is used to configure a unix domain socket listener for the
+	// HTTP API
+	APISocket *SocketConfig `hcl:"api_socket"`
```
schmichael, line 97: The pedant in me wants to name this `api_unix_socket` because it's the "unix" bit that makes this interesting. All `addresses{}` are all sockets. Using `unix` in the name would match Consul and Vault better as well.

That being said if you don't think anyone wants to type that many characters I will shutup. It is nbd. :sweat_smile:

command/agent/command.go
```diff
@@ -122,6 +123,12 @@ func (c *Command) readConfig() *Config {
 	flags.BoolVar(&cmdConfig.LogJson, "log-json", false, "")
 	flags.StringVar(&cmdConfig.NodeName, "node", "", "")
 
+	// API Socket Config options
+	flags.StringVar(&cmdConfig.APISocket.Path, "api-socket-path", "", "")
+	flags.StringVar(&cmdConfig.APISocket.User, "api-socket-user", "", "")
+	flags.StringVar(&cmdConfig.APISocket.Group, "api-socket-group", "", "")
+	flags.StringVar(&cmdConfig.APISocket.Mode, "api-socket-mode", "", "")
```
schmichael, line 130: I think `api-unix-*` makes more sense here (see below for discussion).

schmichael (commented):
This doesn't align directly with [Consul](https://developer.hashicorp.com/consul/docs/agent/config/config-files#unix_sockets) or [Vault](https://developer.hashicorp.com/vault/docs/configuration/listener/unix), but I think that's ok:

- Consul makes you choose between TCP or Unix, and I don't think we want to do that. They also have a distinct `unix_socket{}` block anyway, so we're not far off.
- We match Vault in all the ways that matter. Vault just has that neat `listener{}` block to reuse which we lack. `api_listener{}` blocks probably would have been a better choice for Nomad than `addresses{}`, but that ship sailed ages ago.

command/agent/http.go
schmichael: Looks like syscall exposes it: https://pkg.go.dev/syscall#RawSockaddrUnix

Still might need to subtract 1 to leave room for a null terminator?

command/agent/http.go
tgross: I'm not sure it's all that bad? Most HTTP API calls get mapped to RPCs and those RPCs happen between agents over a separate TCP connection, so the HTTP API's mTLS doesn't ever get consulted as part of all that complex auth logic. The only HTTP APIs that don't get mapped to RPCs are those that hit the local agent anyways. So the HTTP API's mTLS is only used for establishing the connection between the API client and the HTTP server.

So treating the UDS permissions as the equivalent "transport security" for a host-local connection seems ok to me. Unless the UDS socket's permissions are wide-open, anyone who can talk to the UDS can probably grab the TLS cert anyways and spoof that.

That being said, I also suspect a lot of folks run with ACLs disabled on clients, because it's the default and most features work fine without it. Defaulting to requiring ACLs here could encourage wider use of client ACLs and give us a better path to always requiring ACLs (similar to our discussion around the Task API).

command/agent/http.go
```diff
@@ -193,6 +194,58 @@ func NewHTTPServers(agent *Agent, config *Config) ([]*HTTPServer, error) {
...
+			MaxSocketPathLength = 104
+			msgPrefix           = "error starting api socket:"
+		)
+
+		if len(config.APISocket.Path) > MaxSocketPathLength {
+			serverInitializationErrors = multierror.Append(serverInitializationErrors,
+				fmt.Errorf("%s path must be %v characters or less", msgPrefix, MaxSocketPathLength))
+		}
+
+		sl, err := listenerutil.UnixSocketListener(config.APISocket.Path, config.APISocket.UnixSocketsConfig())
```
tgross, line 213: If we're going to use this library we should make sure we've fully documented the expected behavior for users. You can set `user` by name or ID, but `group` only by ID (well, that's how it's _documented_ but not how it behaves if you look at the source). If you set `path` without setting `user`, `group`, and `mode`, the socket is owned by the Nomad agent's user and the mode will be determined by the umask (which is usually different for login shells vs services, etc.)

Maybe we should have an explicit default configuration here rather than relying on the default from the library? i feel like that's likely to get changed out from under us.

command/agent/config.go
```diff
@@ -1089,6 +1093,57 @@ func (n *NormalizedAddrs) Copy() *NormalizedAddrs {
 	return &nn
 }
 
+// SocketConfig contains the path and configuration for a unix domain socket
+// listener.
+type SocketConfig struct {
+	Path  string `hcl:"path"`
+	User  string `hcl:"user"`
+	Mode  string `hcl:"mode"`
+	Group string `hcl:"group"`
```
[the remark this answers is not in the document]
schmichael: Some prior art: 
- [Vault](https://developer.hashicorp.com/vault/docs/configuration/listener/unix) has the bare minimum docs here. 
- [Consul](https://developer.hashicorp.com/consul/docs/agent/config/config-files#unix_sockets) is probably a good one to copy (except for the lack of Windows support... I wonder if it just-started-working for them when Go+Windows added support).
- [Boundary](https://developer.hashicorp.com/boundary/docs/configuration/listener/unix) is clearly an iteration on Vault's approach, and I don't think has anything relevant for us. It's interesting they appear to enable tls by default while Vault totally lacks tls for their unix socket. I think we can follow Vault here until somebody gives us a reason to do otherwise.

command/agent/http.go
schmichael: Oh actually maybe you already peeked because it looks like 104 is the floor!

Peeking at what Go sets across platforms:
```
go/src/syscall$ rg -A3 RawSockaddrUnix | rg Path.*int | awk '{print $3}' | sort -u
[1023]uint8
[104]int8
[108]int8
[UNIX_PATH_MAX]int8
```

Somehow AIX manages to be the nice one here with 1023. What a world. `UNIX_PATH_MAX` is Windows only and 108.

Although I'm not sure if that takes the null terminator weirdness into account? So maybe we should set this to 103?

command/agent/http.go
schmichael: **tl;dr - Ignore me. This is fine as-is.** Feel free to mark resolved.

> So treating the UDS permissions as the equivalent "transport security" for a host-local connection seems ok to me. Unless the UDS socket's permissions are wide-open, anyone who can talk to the UDS can probably grab the TLS cert anyways and spoof that.

Yeah, this makes sense. I think I was ignoring that fact that *for users on Client hosts unix permissions are already the de facto "transport" security.* It's just right now it's unix permissions on the *mTLS client certificates,* not a UDS file. So while adding a UDS is adding one extra file to a an Agent's security perimeter, it doesn't introduce any new concepts, implicit environment requirements, or complexity that mTLS didn't already.

Also I keep forgetting that we shouldn't even really worry about the `acl.enabled = false` configuration as it's outside our security model (despite being the default configuration).

command/agent/http.go
angrycub: +1 on observing the differences in implementation and documentation on that function. What are your thoughts on a reasonable default configuration, or are you saying that we should just codify the current defaults in Nomad so that if they _do change in the library_ that we won't be surprised/broken by it?

tgross: The current defaults from the library seem reasonable, so let's codify those defaults in Nomad so that we control them going forward.

command/agent/http.go
angrycub: Yeah, I'd peeked, but I think you're right about the null terminator. I think we can go for something more OS-specific by changing that line to: 

```go
			MaxSocketPathLength = len(syscall.RawSockaddrUnix{}.Path) - 1 // remove 1 byte for null terminator
```

command/agent/http.go
angrycub: I can ensure the behavior around user and group by making the defaults concrete, but the behavior around `mode` is a moving target since it's solely determined by how the library function handles that argument. I did add a test to validate whether or not the behavior of the library call has changed from we're expecting. Seem okay?

command/agent/http.go
angrycub: I'll drop a TODO here so that I can come back to is but it doesn't stall the world.

command/agent/config.go
```diff
@@ -1089,6 +1093,57 @@ func (n *NormalizedAddrs) Copy() *NormalizedAddrs {
...
+	}
+	if b.User != "" {
+		result.User = b.User
+	}
+	if b.Group != "" {
+		result.Group = b.Group
+	}
+	if b.Mode != "" {
+		result.Mode = b.Mode
+	}
```
shoenig, line 1133: These can be 1-liners with https://github.com/hashicorp/nomad/blob/main/helper/funcs.go#L492-L500

command/agent/config.go
```diff
@@ -1089,6 +1093,57 @@ func (n *NormalizedAddrs) Copy() *NormalizedAddrs {
...
+
+// Merge merges two SocketConfigs together, preferring values from the argument.
+func (s *SocketConfig) Merge(b *SocketConfig) *SocketConfig {
+	if b == nil {
+		return s
+	}
+	var result SocketConfig
+	if s != nil {
+		result = *s
+	}
```
shoenig, line 1121: We just declared `Copy()` above, let's use that instead of implementing it twice

tgross:
Ref: https://hashicorp.atlassian.net/browse/NET-10375

PR #16889, opened 2023-04-14
docs: update licensing FAQ to match current documentation

tgross:
Although we are changing licensing behavior in 1.6.0, the documentation for 1.5.x and previous is only mostly correct. It doesn't include the separate notion of license termination, which is well-described in Vault and Consul docs. The FAQ's section on expiring licenses was also never updated when we updated the behavior around expiration back in 1.1.0.

Make all the pre-1.6.0 licensing documentation accurately describe Nomad's actual behavior.

Note this PR is directly to the `release/1.5.x` branch.

### DIFF
--- a/website/content/docs/enterprise/license/faq.mdx
+++ b/website/content/docs/enterprise/license/faq.mdx
@@ -7,18 +7,23 @@ description: >-
 
 # Frequently Asked Questions (FAQ)
 
-This FAQ is for the license changes introduced in Nomad Enterprise version v1.1.0+ent. 
+This FAQ is for the license changes introduced in Nomad Enterprise version v1.1.0+ent.
 Nomad Enterprise automatically load Nomad licenses when a Nomad server agent starts using Nomad Enterprise.
 
 ## Q: Can I get a quick summary of the Nomad changes?
 
 Starting with Nomad Enterprise v1.1.0, the license enablement process is different.
 
-HashiCorp Enterprise servers will no longer start without a license. When the server instance initializes the license watcher, the server reads from either an environment variable or file. If the license is missing, invalid, or expired, the server will immediately exit. 
-This check is part of the server boot-up process.
+HashiCorp Enterprise servers will no longer start without a license. When the
+server instance initializes the license watcher, the server reads from either an
+environment variable or file. If the license is missing, invalid, or terminated,
+the server will immediately exit. This check is part of the server boot-up
+process.
 
-In previous versions of HashiCorp enterprise products, one server could distribute a license to other servers via the Raft protocol. 
-This will no longer work since each server must be able to find a valid license during the startup process.
+In previous versions of HashiCorp enterprise products, one server could
+distribute a license to other servers via the Raft protocol. This will no longer
+work since each server must be able to find a valid license during the startup
+process.
 
 
 ## Q: What resources are available?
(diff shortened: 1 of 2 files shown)


lgfa29 (approved):
I _think_ you will need to manually backport this to `stable-website` to make it live. Backport Assistant won't help here because it only acts on PRs that target `main`.

We can check the website after merge to see if this is necessary.

website/content/docs/enterprise/license/faq.mdx
```diff
@@ -65,8 +65,17 @@ If you are an existing HashiCorp enterprise customer you may contact your organi
...
+As a Nomad Enterprise license approaches its expiration time, Nomad servers will
+periodically log a warning message about the approaching expiration.
+
+When the license expires, enterprise functionality will become limited. Only
+read operations on enterprise endpoints will be supported, and write operations
+will return an error.
+
+Note that if the server is restarted with an expired license, it will
+immediately stop. Clients are unaffected by expired licenses because clients do
+not read the license.
```
tgross, line 78: @gulducat has pointed out that this doesn't explain how _terminated_ licenses work, which is different from expiration. The docs don't seem to cover this either, so maybe we should expand on the docs in this PR too?

tgross: Following up on this: the only way to actually verify the difference between terminated and expired licenses is to issue a license, but we can't issue expired licenses. So I need to issue myself a license with a near-future expired date far-future terminated date and then wait a day for it to expire. I'll issue that license now and then test it on Monday before we wrap this up.

website/content/docs/enterprise/license/faq.mdx
```diff
@@ -54,19 +59,30 @@ If you have built automation around `nomad license put`, this command will fail
...
-## Q: What happens when a license expires?
-If a license expires after a server node joins the cluster, the existing Enterprise features will continue to operate until a grace period expires. 
-The grace period duration is 1 day, and upon the end of the grace period the Nomad servers will shut down. 
+## Q: What happens when a license expires or terminates?
+
+As a Nomad Enterprise license approaches its expiration time, Nomad servers will
+periodically log a warning message about the approaching expiration. The grace
+period is defined as the time between license expiration and license
+termination. (This is typically one day for evaluation licenses, and much longer
+for non-evaluation licenses.) Customers must provide a valid license before the
```
tgross, line 78: The Vault docs from pre-1.12 say 10 years, but I don't know for sure that this is what was actually issued to Nomad customers because I don't have access to that data. Rather than promising anything in particular, I figure this language will send folks to their sales person.

website/content/docs/enterprise/license/index.mdx
```diff
@@ -21,8 +21,8 @@ for more details.
 
 ## Expiring Licenses
 
-Nomad Enterprise license have an expiration time. You can read the license on
-a server with the `nomad license get` command:
+Nomad Enterprise licenses have an expiration time and a termination time. You
+can read the license on a server with the `nomad license get` command:
```
gulducat, line 25: fun tidbit: `license get` does not show TerminationTime

https://github.com/hashicorp/nomad/blob/v1.5.3/command/license_get.go#L61-L67
=> https://github.com/hashicorp/nomad/blob/v1.5.3/command/license.go#L54

I imagine it just got overlooked in the past.  Do we want to add that too?

gulducat (approved):
I have a couple lil suggestions, but thanks for making this better reflect operational reality! 🥇

website/content/docs/enterprise/license/faq.mdx
```diff
@@ -54,21 +59,32 @@ If you have built automation around `nomad license put`, this command will fail
...
+grace period expires.
+
+When the license terminates, enterprise functionality will become limited. Only
+read operations on enterprise endpoints will be supported, and write operations
+will return an error. Note that if the server is restarted with a terminate
+license, it will immediately stop. Clients are unaffected by terminated licenses
+because clients do not read the license.
```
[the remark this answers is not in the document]
tgross: Done

website/content/docs/enterprise/license/index.mdx
tgross: Probably. I can do that under separate PR; this should just be a docs PR.

tgross: Done in https://github.com/hashicorp/nomad/pull/16892

tgross:
Hand-backported to 1.4.x and 1.3.x, force-pushed `release/1.5.x` onto `stable-website`

PR #16905, opened 2023-04-17
services: un-mark group services as deregistered if restart hook runs

### LINKED ISSUE #16616: Services not unregistered
Just upgraded to Nomad 1.5.2. Since then, services are not always unregistered from Consul service catalog when they are shuted down / upgraded. So old services versions appear as failed, eg

image

Environment : 
- Nomad 1.5.2 (with ACL), using the prebuilt binary
- Consul 1.15.1 (with ACL), using the pre-built binary
- Alma Linux 8

Haven't found yet a pattern to reproduce it 100% of the time

shoenig:
This PR may fix a bug where group services will never be deregistered if the group undergoes an alloc restart.

Fixes #16616

Adds an e2e test case covering the repro in ^

### DIFF
--- a/e2e/consul/alloc_restart_test.go
+++ b/e2e/consul/alloc_restart_test.go
@@ -5,11 +5,14 @@ package consul
 
 import (
 	"context"
+	"errors"
 	"testing"
+	"time"
 
 	"github.com/hashicorp/nomad/e2e/e2eutil"
 	"github.com/hashicorp/nomad/helper/uuid"
 	"github.com/shoenig/test/must"
+	"github.com/shoenig/test/wait"
 )
 
 func testAllocRestart(t *testing.T) {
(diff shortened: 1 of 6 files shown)


e2e/consul/alloc_restart_test.go
```diff
@@ -0,0 +1,59 @@
...
+	// make sure our service is still registered
+	services, _, err = cc.Service("alloc-restart-http", "", nil)
+	must.NoError(t, err)
+	must.Len(t, 1, services)
+
+	err = e2eutil.StopJob(jobID)
+	must.NoError(t, err)
+
+	// make sure our service is no longer registered
+	services, _, err = cc.Service("alloc-restart-http", "", nil)
```
tgross, line 56: I'm pretty sure by the time the alloc is marked running the service will be registered. But service deregistration isn't synchronous with the `Job.Deregister` RPC, right? I think we need to `must.Wait` for this API call.

tgross (approved):
LGTM

e2e/consul/alloc_restart_test.go
shoenig: Good catch, will add that

PR #16907, opened 2023-04-17
[scheduler] Honor `false` for distinct hosts constraint

### LINKED ISSUE #13861: Make distinct_hosts as a variable not working
I want to disable `distinct_hosts` in certain env, but it not working as expected.
- Template
```
job "" {
  constraint {
    operator  = "distinct_hosts"
    value     = var.distinct_hosts
  }
}

variable "distinct_hosts" {
  description = "This job should be scheduled on different hosts"
  type        = string
  default     = "true"
}
```

- Usage
```
distinct_hosts = "false"
```

Then I find the source code, is not reading the value config.

```
func (iter *DistinctHostsIterator) hasDistinctHostsConstraint(constraints []*structs.Constraint) bool {
	for _, con := range constraints {
		if con.Operand == structs.ConstraintDistinctHosts {
			return true
		}
	}

	return false
}
```

angrycub:
Currently, when checking a nodes feasibility against the `distinct_hosts` constraint, the mere presence of the constraint is sufficient to activate it; however, the documentation indicates that setting the value to false is a legitimate answer.

The `jobspec` HCL1 parser would look for cases where the value was set to false and strip it from the generated job specification before submission. The `jobspec2` HCL2 parser writes it into the specification with the provided value whether true or false.

This has the consequence of causing jobs or tasks with a distinct_hosts constraint to always act as though the value is set to true.
 
Fixes #13861 

- Honor value for distinct_hosts constraint
- Add test for feasibility checking for `false`

### DIFF
--- a/scheduler/feasible.go
+++ b/scheduler/feasible.go
@@ -572,7 +572,7 @@ func (iter *DistinctHostsIterator) SetJob(job *structs.Job) {
 func (iter *DistinctHostsIterator) hasDistinctHostsConstraint(constraints []*structs.Constraint) bool {
 	for _, con := range constraints {
 		if con.Operand == structs.ConstraintDistinctHosts {
-			// Maintain the old behavior around unset RTargets
+			// distinct_hosts defaults to true
 			if con.RTarget == "" {
 				return true
 			}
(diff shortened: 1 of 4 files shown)


scheduler/feasible.go
```diff
@@ -572,7 +572,14 @@ func (iter *DistinctHostsIterator) SetJob(job *structs.Job) {
 func (iter *DistinctHostsIterator) hasDistinctHostsConstraint(constraints []*structs.Constraint) bool {
 	for _, con := range constraints {
 		if con.Operand == structs.ConstraintDistinctHosts {
-			return true
+			// Maintain the old behavior around unset RTargets
```
schmichael, line 575:
```suggestion
			// distinct_hosts defaults to true
```

schmichael (approved):
Nice find!

tgross (commented):
Looks like @schmichael beat me to the punch on reviewing this, but just wanted to say LGTM and the PR description here was top notch!

PR #16916, opened 2023-04-17
ci: fix missing import

lgfa29:
Add missing import to fix https://github.com/hashicorp/nomad/actions/runs/4726130915/jobs/8385391229#step:4:3833

Targets `release/1.3.x` only.


jrasell (approved):
LGTM!

PR #16923, opened 2023-04-18
rpc: use "+" concatenations in hot path RPC rate limit metrics.

jrasell:
Using the "+" operator is faster than the `fmt` package functions which I think makes sense to use here as they are simple concatenations and in a hot-path. I added some simple tests to ensure the change doesn't alter the behaviour.


tgross (approved):
LGTM!

PR #16930, opened 2023-04-19
docs: add missing API field JobACL and fix workload identity headers

lgfa29:
For some reason the headers are not displayed if they're a level 1 header:
https://developer.hashicorp.com/nomad/docs/concepts/workload-identity


tgross (approved):
LGTM

Issue #16939, opened 2023-04-19
SIGSEGV on nomad plan


pschichtel:
### Nomad version
Nomad v1.5.3

### Operating system and Environment details

Archlinux (Stable), up to date

### Issue

While the nomad plan command generally works, it crashes with a SIGSEGV at the end, bash -x output:

```
+ nomad job plan --diff=false example.nomad
Scheduler dry-run:
- All tasks successfully allocated.

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x55de36ab6a93]

goroutine 1 [running]:
github.com/hashicorp/nomad/command.getExitCode(...)
	/build/nomad/src/nomad/command/job_plan.go:394
github.com/hashicorp/nomad/command.(*JobPlanCommand).outputPlannedJob(0xc0005c0700, 0xc0006029c0?, 0xc000220d20, 0xb0?, 0x13?)
	/build/nomad/src/nomad/command/job_plan.go:327 +0x2f3
github.com/hashicorp/nomad/command.(*JobPlanCommand).Run(0xc0005c0700, {0xc0000520d0, 0x2, 0x2})
	/build/nomad/src/nomad/command/job_plan.go:265 +0xcb0
github.com/mitchellh/cli.(*CLI).Run(0xc000a25e00)
	/build/nomad/src/pkg/mod/github.com/mitchellh/cli@v1.1.5/cli.go:262 +0x5f8
main.Run({0xc0000520b0, 0x4, 0x4})
	/build/nomad/src/nomad/main.go:107 +0x28a
main.main()
	/build/nomad/src/nomad/main.go:77 +0x4e
+ exit_code=2
+ '[' 2 == 1 ']'
+ echo 'Job planning failed: 2'
Job planning failed: 2
+ exit 2
```

The plan did indeed happen, but apparently it crashed when trying to determine the exit code.

### Reproduction steps

1. setup local single-node consul
2. setup local single-node nomad
3. write simple job file
4. run `nomad plan job.nomad`

#### Expected Result

Program exits normally

#### Actual Result

Program crashes with aforementioned SIGSEGV.

lgfa29:
Thank you for the report @pschichtel!

This is the result of a new check added in 1.5 that did not expect the `-diff=false` to behave the way it currently does. I opened https://github.com/hashicorp/nomad/pull/16944 with a fix for this problem.

[...]

PR #16944, opened 2023-04-20
cli: fix panic on job plan when -diff=false

lgfa29:
PR #14492 introduced a new check to return 0 when the `nomad job plan` command returns a diff of type `None`.

But the `-diff` CLI flag was also being used to control whether the plan request should return the diff of not instead of just controlling if the diff was printed.

This means that when `-diff=false` is set the response does not include any diff information, and so the new check panics.

This commit fixes the problem by always requesting a diff and using the `-diff` only for controlling output, as it's currently documented.

Closes https://github.com/hashicorp/nomad/issues/16939

### DIFF
--- a/.changelog/16944.txt
+++ b/.changelog/16944.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cli: Fix panic on job plan when -diff=false
+```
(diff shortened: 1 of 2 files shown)


lgfa29:
https://github.com/hashicorp/nomad/pull/14492 mentions that the change will be released in the next 1.4 patch release, but the change wasn't actually backported so it only came out in 1.5.0.

schmichael (approved):

PR #16960, opened 2023-04-21
[ui] Fixed: Evaluations sidebar/response not scrollable

philrenaud:
This PR establishes CSS flexbox regions in the Evaluations sidebar so both it, and the evaluation response, can be scrolled.

### DIFF
--- a/ui/app/styles/components/sidebar.scss
+++ b/ui/app/styles/components/sidebar.scss
@@ -12,7 +12,7 @@ $subNavOffset: 49px;
   width: 750px;
   padding: 24px;
   right: 0%;
-  overflow: visible;
+  overflow: auto;
   bottom: 0;
   top: $topNavOffset;
   transform: translateX(100%);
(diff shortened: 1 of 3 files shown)


ui/app/styles/components/sidebar.scss
```diff
@@ -41,6 +41,26 @@ $subNavOffset: 49px;
...
+  &.evaluations-sidebar {
+    display: flex;
+    gap: 1.5rem;
+    flex-direction: column;
+
+    & > .evaluation-response {
+      display: grid;
+      min-height: 200px;
+      grid-template-rows: auto 1fr;
+      overflow: hidden;
```
philrenaud, line 54: The parent evaluation response container has overflow: hidden deliberately: it contains both the header and the json viewer, the latter of which has auto overflow.

ui/app/styles/components/sidebar.scss
```diff
@@ -41,6 +41,26 @@ $subNavOffset: 49px;
...
+      & > .boxed-section-body {
+        overflow: auto;
+      }
+    }
+    & > div,
+    h1.title {
+      margin: 0;
```
philrenaud, line 61: Let the flex/grid parent handle gap/margin where possible

PR #16971, opened 2023-04-24
copywrite: excempt example assets from copywrite headers

shoenig:


tgross (approved):
LGTM! Examples were called out in the [RFC](https://docs.google.com/document/d/1rUmuAGNDULSbVe2MJOXm9guCHGdYQ2wlpt3XL8-EPz0/edit#) (internal) as something we could exempt.

PR #16972, opened 2023-04-24
[ui, deployments] Restarted and Rescheduled panel cells

philrenaud:
Adds indicators and links for those allocations that have been restarted and rescheduled, either during a deployment or due to failures in an otherwise steady state.

image

Additionally, adds the concept of "permanent failure" to allocations we show in the main chart. For example, an allocation that has failed past its restart attempts will go into a "failed" clientStatus, but we don't want to show it in the main chart if there's a rescheduled allocation to take it over. The new `allocation.hasBeenRescheduled` computed property handles this.

### DIFF
--- a/ui/app/components/job-status/panel/deploying.js
+++ b/ui/app/components/job-status/panel/deploying.js
@@ -144,6 +144,12 @@ export default class JobStatusPanelDeployingComponent extends Component {
     ];
   }
 
+  get failedOrLostAllocs() {
+    return this.job.allocations.filter(
+      (a) => a.jobVersion === this.deployment.get('versionNumber') && (a.clientStatus === 'failed' || a.clientStatus === 'lost' || a.clientStatus === 'unknown')
+    );
+  }
+
   // #region legend
   get newAllocsByStatus() {
     return Object.entries(this.newVersionAllocBlocks).reduce(
--- a/ui/app/components/job-status/panel/deploying.js
+++ b/ui/app/components/job-status/panel/deploying.js
@@ -14,7 +14,7 @@ export default class JobStatusPanelDeployingComponent extends Component {
     'pending',
     'failed',
     // 'unknown',
-    // 'lost',
+    'lost',
     // 'queued',
     // 'complete',
     'unplaced',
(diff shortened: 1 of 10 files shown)


ui/app/components/job-status/panel/steady.js
```diff
@@ -25,10 +25,10 @@ export default class JobStatusPanelSteadyComponent extends Component {
     let availableSlotsToFill = this.totalAllocs;
     // Only fill up to 100% of totalAllocs. Once we've filled up, we can stop counting.
     let allocationsOfShowableType = this.allocTypes.reduce((blocks, type) => {
-      const jobAllocsOfType = this.args.job.allocations.filterBy(
-        'clientStatus',
-        type.label
-      );
+      const jobAllocsOfType = this.args.job.allocations
+        .sortBy('jobVersion') // Try counting from latest deployment's allocs and work backwards if needed
```
philrenaud, line 29: Side-effect; will count backwards from the most recent by job version, rather than iterating through them in their default order.

ui/app/components/job-status/panel/deploying.js
```diff
@@ -61,7 +61,7 @@ export default class JobStatusPanelDeployingComponent extends Component {
   fail;
 
   @alias('job.latestDeployment') deployment;
-  @alias('deployment.desiredTotal') desiredTotal;
+  @alias('totalAllocs') desiredTotal;
```
philrenaud, line 64: This was a change to address a bug found when restarting a dead job. This bug is present in Nomad's UI today. The latest deployment on a job that you start from dead have a desired count that differs from the sum of the job's groups.count, which is the number the operator is really going for.

ui/app/components/job-status/panel/deploying.js
```diff
@@ -104,6 +104,14 @@ export default class JobStatusPanelDeployingComponent extends Component {
       let canary = alloc.isCanary ? 'canary' : 'nonCanary';
 
       if (allocationCategories[status]) {
+        // If status is failed or lost, we only want to show it IF it's used up its restarts/rescheds.
+        // Otherwise, we'd be showing an alloc that had been replaced.
+        if (alloc.willNotRestart) {
+          if (!alloc.willNotReschedule) {
+            // Dont count it
+            continue;
```
philrenaud, line 112: Style thing: we don't use `continue` very often for breaking out of an iteration! It seemed apt here, though.

ChaiWithJai (approved):

PR #16974, opened 2023-04-24
tools: update dependencies and use tree set

shoenig:
- tools: bump go mod deps for tools module
- tools: use treeset in tools/missing


jrasell (approved):
LGTM!

PR #16987, opened 2023-04-26
ui: visualize variables

ChaiWithJai:
In this PR, we introduce a feature to display specification literals via variable flags from the API response in the read-only and edit views. The PR also adds styling and spacing adjustments to ensure that the displayed literals look visually appealing and that code mirror doesn't take up too much space. Additionally, the PR includes logic to manually convert JSON objects to HCL format using loose logic, focusing on variable flags. 

Side Effect: Note that the text displayed in the Job Editor header is now only conditionally rendered to show on the job run page, which utilizes the same editor component.

Loom Walk Through:  https://www.loom.com/share/07a0d2cde466416890d3aaa3988e5559

### DIFF
--- a/ui/app/adapters/job.js
+++ b/ui/app/adapters/job.js
@@ -19,6 +19,15 @@ export default class JobAdapter extends WatchableNamespaceIDs {
     return this.ajax(url, 'GET');
   }
 
+  fetchRawSpecification(job) {
+    const url = addToPath(
+      this.urlForFindRecord(job.get('id'), 'job', null, 'submission'),
+      '',
+      'version=' + job.get('version')
+    );
+    return this.ajax(url, 'GET');
+  }
+
   forcePeriodic(job) {
     if (job.get('periodic')) {
       const url = addToPath(
--- a/ui/app/components/job-editor.js
+++ b/ui/app/components/job-editor.js
@@ -144,6 +144,22 @@ export default class JobEditor extends Component {
     }
   }
 
+  get variables() {
+    function jsonToHcl(obj) {
+      const hclLines = [];
+
+      for (const key in obj) {
+        const value = obj[key];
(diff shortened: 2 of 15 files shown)


philrenaud:
Seems like this one's still conflicting with its parent; mind doing a rebase before review?

ui/app/adapters/job.js
```diff
@@ -19,6 +19,15 @@ export default class JobAdapter extends WatchableNamespaceIDs {
     return this.ajax(url, 'GET');
   }
 
+  fetchRawSpecification(job) {
+    const url = addToPath(
+      this.urlForFindRecord(job.get('id'), 'job', null, 'submission'),
+      '',
+      'version=' + job.get('version')
```
philrenaud, line 26: `version=${job.get('version')}` would help you avoid future concatenation + line return readability woes at some point, but this is fine either way

ui/app/components/job-editor.js
```diff
@@ -144,6 +144,22 @@ export default class JobEditor extends Component {
     }
   }
 
+  get variables() {
+    function jsonToHcl(obj) {
+      const hclLines = [];
+
+      for (const key in obj) {
+        const value = obj[key];
+        const hclValue = typeof value === 'string' ? `"${value}"` : value;
```
philrenaud, line 153: This is interesting! I would have expected that HCL values would already come to you wrapped in "" when strings. Wouldn't this result in `""string""` issues?

philrenaud (approved):

ui/app/components/job-editor.js
ChaiWithJai: hmmm... good point! I guess this really is determined by the API. I tested manually and that's what worked for me. Reading through the tests in the Go code, there's nothing that would tell me what shape to expect. Maybe we'll need to play around with this a bit more and adjust accordingly!

PR #16990, opened 2023-04-26
testing: improve fidelity of mock driver task restore

tgross:
While working on client status update improvements, I encountered problems getting tests with the mock driver to correctly restore.

Unlike typical drivers the mock driver doesn't have an external source of truth for whether the task is running (ex. making API calls to `dockerd` or looking for a running PID), and so in order to make up that information, it re-parses the original task config. But the taskrunner doesn't call the encoding step for `RecoverTask`, only `StartTask`, so the task config the mock driver gets is missing data.

Update the mock driver to stash the "external" state in the task state that we'll get from the task runner, so that we don't have to try to recover from the original `TaskConfig` anymore. This should bring the mock driver closer to the behavior of the other drivers.

### DIFF
--- a/drivers/mock/driver.go
+++ b/drivers/mock/driver.go
@@ -266,6 +266,15 @@ type TaskConfig struct {
 
 type MockTaskState struct {
 	StartedAt time.Time
+
+	// these are not strictly "state" but because there's no external
+	// reattachment we need somewhere to stash this config so we can properly
+	// restore mock tasks
+	Command         Command
+	ExecCommand     *Command
+	PluginExitAfter time.Duration
+	KillAfter       time.Duration
+	ProcState       drivers.TaskState
 }
 
 func (d *Driver) PluginInfo() (*base.PluginInfoResponse, error) {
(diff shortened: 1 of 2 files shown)


drivers/mock/driver.go
```diff
@@ -358,21 +367,39 @@ func (d *Driver) RecoverTask(handle *drivers.TaskHandle) error {
...
+	killCtx, killCancel := context.WithCancel(context.Background())
+	h := &taskHandle{
+		logger:          d.logger.With("task_name", handle.Config.Name),
+		pluginExitAfter: taskState.PluginExitAfter,
+		killAfter:       taskState.KillAfter,
+		waitCh:          make(chan any),
+		taskConfig:      handle.Config,
+		command:         taskState.Command,
+		execCommand:     taskState.ExecCommand,
+		//		procState:       drivers.TaskStateRunning,
```
shoenig, line 396: we don't want this?

shoenig (approved):
LGTM; just the Q

drivers/mock/driver.go
tgross: Oops, that's left over from debugging the test to make sure we don't _need_ to set it, because it should get set when we call `go h.run()` below and that'll account for when the `run_for` expires. However, we also should be setting it here for fidelity with the other task drivers. Will fix.

tgross: Done. In order to make sure the test doesn't just read that value, I also had to change it from `InitialSuccess` to `ContinualSuccess`

PR #16995, opened 2023-04-27
vars: ensure struct receiver names are consistent.

jrasell:


tgross (approved):
LGTM! Thanks, we did get a little sloppy with the receivers during the Great Feature Rename :grinning:

PR #17005, opened 2023-04-27
docs: add more notes about artifact breaking changes in 1.5.0

### LINKED ISSUE #17000: breaking change should be documetned
artifact: enable inheriting environment variables from client https://github.com/hashicorp/nomad/pull/15514

regarding the above commit, we were relying on the client to pass credentials to the getter subprocess. After upgrading to nomad 1.5.3 our testing clusters start to fail to fetch artifacts, despite seeing no breaking change related to this aspect.

Open an issue since the orignal pr above is locked, just for some awareness, since a nice change could have very realistic downstream effect, thats all :) 

still, thanks for the continuous improvements to the product! That is the reason keeping us wanting to upgrade :D

shoenig:
- changelog: note artifact breaking changes for 1.5.0
- docs: add note about environment variables to artifact job spec docs

Closes #17000

I didn't add a .changelog file for the compression limits entry because there already is one (originally got sorted under improvements, now calling it out as a breaking change)

### DIFF
--- a/website/content/docs/job-specification/artifact.mdx
+++ b/website/content/docs/job-specification/artifact.mdx
@@ -246,7 +246,7 @@ artifact {
 
 ## Environment
 
-Starting with Nomad 1.5.0, the `artifact` downloader by default does not have access to the environment variables
+The `artifact` downloader by default does not have access to the environment variables
 set for the Nomad client. Inheritence of environment variables can be managed through the [`artifact.set_environment_variables`][client_artifact]
 client configuration.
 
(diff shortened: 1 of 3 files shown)


website/content/docs/job-specification/artifact.mdx
```diff
@@ -244,6 +244,12 @@ artifact {
 }
 ```
 
+## Environment
+
+Starting with Nomad 1.5.0, the `artifact` downloader by default does not have access to the environment variables
```
lgfa29, line 249:
```suggestion
The `artifact` downloader by default does not have access to the environment variables
```

With versioned docs we don't need to reference specific versions on the docs anymore, just make sure this gets backported to `release/1.5.x` only.

lgfa29 (approved):

PR #17021, opened 2023-04-28
[ui] Stabilize Percy visual diff tests for new Job Status feature

philrenaud:
Our acceptance tests generate some relative-time-from-now text that varies on each test run. This hides those values.

### DIFF
--- a/ui/tests/acceptance/job-status-panel-test.js
+++ b/ui/tests/acceptance/job-status-panel-test.js
@@ -37,7 +37,11 @@ module('Acceptance | job status panel', function (hooks) {
     await visit(`/jobs/${job.id}`);
     assert.dom('.job-status-panel').exists();
     await a11yAudit(assert);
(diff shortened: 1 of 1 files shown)


lgfa29 (approved):
I guess this works, but I think the root cause is that the allocation factory uses an unstable date seed:
https://github.com/hashicorp/nomad/blob/0aae32a346f49cfcc160cd5e9f02b496ff4a1954/ui/mirage/factories/allocation.js#L16

Maybe if we change it to a specific date we would not have this problem?
```js
const REF_TIME = new Date('2023-05-28T01:23:45');
```

philrenaud:
> Maybe if we change it to a specific date we would not have this problem?

Yeah, that's a good option too. My thinking here is we already have several screenshots in percy that specifically look at the alloc and other similar tables with hardcoded timestamps; in this case it seems fine to just blank it out.

PR #17041, opened 2023-05-01
connect: do not restrict auto envoy version to docker task driver

shoenig:
This PR updates the envoy_bootstrap_hook to no longer disable itself if
the task driver in use is not docker. In other words, make it work for
podman and other image based task drivers. The hook now only checks that

1. the task is a connect sidecar
2. the task.config block contains an "image" field


gulducat (approved):
just to phrase it differently, this is to allow this kind of config in a [`sidecar_task` block](https://developer.hashicorp.com/nomad/docs/job-specification/sidecar_task)

```hcl
# job..group..service {
connect {
  sidecar_task {
    driver = "podman"
  }
}
```

without needing to also explicitly specify an envoy `image` for the sidecar task, right?

shoenig:
@gulducat yup, this combined with https://github.com/hashicorp/nomad/pull/17045 should get us to the point of using your config block. After that I'd like to go one step further, and adding a `meta.connect.driver` client config option (similar to the already existing meta.connect.image and meta.connect.log_level options) so that Podman oriented clusters can make use of a default `connect{}` block with no additional job level configuration.

lgfa29:
Sorry, late to the party 😅 

Would this approach cause problems for task drivers that use `image` for non-OCI paths, like the [`pot`](https://developer.hashicorp.com/nomad/plugins/drivers/community/pot#image) driver?

I think we document that Connect is currently only supported in Docker, so that's probably OK, but I wonder if it would make sense to create a new task driver capability.

shoenig:
The envoy version hook is also gated on the task being a Connect sidecar, which can only be true if the task is using bridge networking, which wouldn't apply to most task drivers.

Adding a new task driver capability is an interesting idea, but I suspect it would end up mirroring the same image/sidecar checks already in the hook.

PR #17045, opened 2023-05-01
connect: set default envoy version domain

shoenig:
This PR modifies references to the `envoyproxy/envoy` docker image to
explicitly include the `docker.io` prefix. This does not affect existing
[docker] users, but makes things easier for Podman users, who otherwise need to
specify the full name because Podman does not default to `docker.io`


tgross (approved):
LGTM! :+1:

PR #17051, opened 2023-05-02
[internal] Add WriterUI

angrycub:
This special purpose UI provides commands that can benefit from direct access to the `io.Reader` and `io.Writers` of a `cli.BasicUi`(or cli.MockUi for testing). It can also descend into a chain of `cli.ColoredUi`s to find a BasicUi or MockUi. 

The `nomad operator api` currently copies the api response directly to os.Stdout to avoid the cost of buffering a potentially huge response before output. Using the WriterUI, the response can be copied directly to the UI's writer. This also facilitates testing, since we can use MockUi's to provide the readers and writers to the WriterUi.

Renames ui.go and ui_test.go to log_ui.go and log_ui_test.go


tgross (approved):
LGTM! Nice work!

PR #17056, opened 2023-05-02
build: upgrade from go 1.20.3 to 1.20.4

schmichael:
Includes CVE fixes that do *not* impact Nomad:

https://groups.google.com/g/golang-announce/c/MEb0UyuSMsU


tgross (approved):
LGTM but it'll usually take a little bit before the build image gets updated.

shoenig (approved):

PR #17063, opened 2023-05-02
[ui, deployments] Fix order in which allocations are ascribed to the status chart

philrenaud:
The order in which we want to show allocations (that may have previously been relevant but perhaps no longer are?) in our new visualization has gone through a few changes.

First, we figured "Let's just show them in order of what the user cares about! Running first, then pending, then failed, then lost...." and so on.

Then, we realized that this would clump all the failed allocations from older versions together before we got to the lost allocations of the newest version.

Because the user would expect the "latest" allocs to exist there instead, we can walk backwards by job version instead.

Ah! But that's not *quite right*. Running/Pending still matter above all else. So the new, nuanced solution looks like this:
- Look at all job allocations and grab the running/pending ones first
- Then, sort reverse by jobVersion and walk backwards to fill up `totalAllocs` (the desired total)
- Any that are left over / would otherwise be taken up by now-GC'd allocs are set as `unplaced`

===

We can most easily see this error in the previous version with the following situation:
- job has a count of 2
- job v0 has an alloc placement that goes `failed`
- job v1 has an alloc placement that goes `lost`

Before this fix, the 2 allocations that would have been shown would be "running" and "failed", since failed was earlier in the allocTypes list.

Now, it will show "running" and "lost", which is closer to the user's expectations in reverse-job-version ordering.

image
image

### DIFF
--- a/ui/app/components/job-status/panel/deploying.js
+++ b/ui/app/components/job-status/panel/deploying.js
@@ -30,9 +30,7 @@ export default class JobStatusPanelDeployingComponent extends Component {
   // allocations we can track throughout a deployment
   establishOldAllocBlockIDs() {
     this.oldVersionAllocBlockIDs = this.job.allocations.filter(
-      (a) =>
-        a.clientStatus === 'running' &&
-        a.jobVersion !== this.deployment.get('versionNumber')
+      (a) => a.clientStatus === 'running' && a.isOld
     );
   }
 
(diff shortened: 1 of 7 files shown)


ui/app/components/job-status/panel/deploying.js
```diff
@@ -30,9 +30,7 @@ export default class JobStatusPanelDeployingComponent extends Component {
   // allocations we can track throughout a deployment
   establishOldAllocBlockIDs() {
     this.oldVersionAllocBlockIDs = this.job.allocations.filter(
-      (a) =>
-        a.clientStatus === 'running' &&
-        a.jobVersion !== this.deployment.get('versionNumber')
+      (a) => a.clientStatus === 'running' && a.isOld
```
philrenaud, line 33: Side-effect: I found `isOld` for the first time and it'll help make a bunch of this more concise.

ChaiWithJai (approved):

PR #17064, opened 2023-05-02
styling:  UX requests

ChaiWithJai:
This PR tackles several styling requests:

- Disabling the toggle with tooltip messaging
- Spacing out alerts
- Dismissing the plan message
- Sticky job plan button
- Autofocusing the code mirror editor to prevent automatic scrolling when there's too much code in the editor or a second editor is present

Video Walk Through:  https://www.loom.com/share/74a9dc40dc2d4d5586b4e4f98f0538b6

### DIFF
--- a/ui/app/templates/components/tooltip.hbs
+++ b/ui/app/templates/components/tooltip.hbs
@@ -3,6 +3,10 @@
   SPDX-License-Identifier: MPL-2.0
 ~}}
 
-<span class="tooltip" aria-label="{{this.text}}">
+{{#if this.condition}}
+  <span class={{concat "tooltip" (if @isFullText " multiline")}} aria-label={{if @isFullText @text this.text}}>
+    {{yield}}
+  </span>
+{{else}}
   {{yield}}
-</span>
+{{/if}}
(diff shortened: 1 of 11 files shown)


ui/app/templates/components/tooltip.hbs
```diff
@@ -3,6 +3,10 @@
   SPDX-License-Identifier: MPL-2.0
 ~}}
 
-<span class="tooltip" aria-label="{{this.text}}">
+{{#if this.condition}}
```
philrenaud, line 6: Request: could we add some tests for condition? Is it expected to always be a boolean? If so, let's consider adding an assert/check to make sure of its type.

In particular I worry about a condition that results in a falsy value being passed in, like `@condition=job.version` where version could be 0 / result in a failed condition.

philrenaud (changes requested):

ui/app/templates/components/tooltip.hbs
ChaiWithJai: We can design the component API to have an assertion where only booleans can be passed here to prevent this. That's probably the most appropriate solution IMO. Are you OK with that?

philrenaud (approved):

PR #17065, opened 2023-05-02
connect: use heuristic to detect sidecar task driver

### LINKED ISSUE #17042: connect: simplify envoy sidecar when using podman
When running the envoy connect sidecar as a podman job, one currently needs to configure the `image` manually, partly due to https://github.com/hashicorp/nomad/issues/17040 and partly because the image name needs to be prefixed with `docker.io/`.

Ideally, podman oriented clusters could use Connect jobs just like docker oriented clusters, where the `connect{}` block by default Just Works.

Incorporates 
- #17040
- #17045
- #17065

shoenig:
This PR adds a heuristic to detect whether to use the podman task driver
for the connect sidecar proxy. The podman driver will be selected if there
is at least one task in the task group configured to use podman, and there
are zero tasks in the group configured to use docker. In all other cases
the task driver defaults to docker.

After this change, we should be able to run typical Connect jobspecs
(e.g. nomad job init [-short] -connect) on Clusters configured with the
podman task driver, without modification to the job files.

Closes #17042

### DIFF
--- a/nomad/job_endpoint_hook_connect.go
+++ b/nomad/job_endpoint_hook_connect.go
@@ -230,6 +230,29 @@ func injectPort(group *structs.TaskGroup, label string) {
 	})
 }
 
+// groupConnectGuessTaskDriver will scan the tasks in g and try to decide which
+// task driver to use for the default sidecar proxy task definition.
+//
+// If there is at least one podman task and zero docker tasks, use podman.
+// Otherwise default to docker.
+//
+// If the sidecar_task block is set, that takes precedence and this does not apply.
+func groupConnectGuessTaskDriver(g *structs.TaskGroup) string {
+	foundPodman := false
+	for _, task := range g.Tasks {
+		switch task.Driver {
+		case "docker":
+			return "docker"
+		case "podman":
+			foundPodman = true
+		}
+	}
+	if foundPodman {
+		return "podman"
+	}
+	return "docker"
+}
+
 func groupConnectHook(job *structs.Job, g *structs.TaskGroup) error {
 	// Create an environment interpolator with what we have at submission time.
 	// This should only be used to interpolate connect service names which are
@@ -254,7 +277,8 @@ func groupConnectHook(job *structs.Job, g *structs.TaskGroup) error {
 
 			// If the task doesn't already exist, create a new one and add it to the job
 			if task == nil {
-				task = newConnectSidecarTask(service.Name)
+				driver := groupConnectGuessTaskDriver(g)
+				task = newConnectSidecarTask(service.Name, driver)
 
 				// If there happens to be a task defined with the same name
 				// append an UUID fragment to the task name
@@ -477,12 +501,12 @@ func newConnectGatewayTask(prefix, service string, netHost, customizedTls bool) 
 	}
 }
 
-func newConnectSidecarTask(service string) *structs.Task {
+func newConnectSidecarTask(service, driver string) *structs.Task {
 	return &structs.Task{
 		// Name is used in container name so must start with '[A-Za-z0-9]'
 		Name:          fmt.Sprintf("%s-%s", structs.ConnectProxyPrefix, service),
 		Kind:          structs.NewTaskKind(structs.ConnectProxyPrefix, service),
-		Driver:        "docker",
+		Driver:        driver,
 		Config:        connectSidecarDriverConfig(),
 		ShutdownDelay: 5 * time.Second,
 		LogConfig: &structs.LogConfig{
--- a/nomad/job_endpoint_hook_connect.go
+++ b/nomad/job_endpoint_hook_connect.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/hashicorp/go-set"
 	"github.com/hashicorp/nomad/client/taskenv"
 	"github.com/hashicorp/nomad/helper/envoy"
 	"github.com/hashicorp/nomad/helper/pointer"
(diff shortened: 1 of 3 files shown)


shoenig:
Spot check

<details><summary>nomad agent config</summary>

```hcl
server {
  enabled = true
}

client {
  enabled = true

  options = {
    driver.allowlist = "podman"
  }
}

plugin_dir = "/home/shoenig/Go/podman/build"

plugin "nomad-driver-podman" {
}
```

</details>

```hcl
job "podcon" {

  group "api" {
    network {
      mode = "bridge"
    }

    service {
      name = "count-api"
      port = "9001"

      connect {
        sidecar_service {}
      }
    }

    task "web" {
      driver = "podman"

      config {
        image = "docker.io/hashicorpdev/counter-api:v3"
      }
    }
  }

  group "dashboard" {
    network {
      mode = "bridge"

      port "http" {
        static = 9002
        to     = 9002
      }
    }

    service {
      name = "count-dashboard"
      port = "9002"

      connect {
        sidecar_service {
          proxy {
            upstreams {
              destination_name = "count-api"
              local_bind_port  = 8080
            }
          }
        }
      }
    }

    task "dashboard" {
      driver = "podman"

      env {
        COUNTING_SERVICE_URL = "http://${NOMAD_UPSTREAM_ADDR_count_api}"
      }

      config {
        image = "docker.io/hashicorpdev/counter-dashboard:v3"
      }
    }
  }
}
```

</details>

plan is ok

```
➜ nomad job plan podcon.hcl
+ Job: "podcon"
+ Task Group: "api" (1 create)
  + Task: "connect-proxy-count-api" (forces create)
  + Task: "web" (forces create)

+ Task Group: "dashboard" (1 create)
  + Task: "connect-proxy-count-dashboard" (forces create)
  + Task: "dashboard" (forces create)

Scheduler dry-run:
- All tasks successfully allocated.

Job Modify Index: 0

nomad job run -check-index 0 podcon.hcl

potentially invalid.
```

deployment ok

```
➜ nomad job run podcon.hcl
==> 2023-05-02T21:15:18Z: Monitoring evaluation "859d0de3"
    2023-05-02T21:15:18Z: Evaluation triggered by job "podcon"
    2023-05-02T21:15:19Z: Evaluation within deployment: "35a39ae8"
    2023-05-02T21:15:19Z: Allocation "7ff9e075" created: node "4d26c0cf", group "api"
    2023-05-02T21:15:19Z: Allocation "b3b163c5" created: node "4d26c0cf", group "dashboard"
    2023-05-02T21:15:19Z: Evaluation status changed: "pending" -> "complete"
==> 2023-05-02T21:15:19Z: Evaluation "859d0de3" finished with status "complete"
==> 2023-05-02T21:15:19Z: Monitoring deployment "35a39ae8"
  ✓ Deployment "35a39ae8" successful

    2023-05-02T21:15:30Z
    ID          = 35a39ae8
    Job ID      = podcon
    Job Version = 0
    Status      = successful
    Description = Deployment completed successfully

    Deployed
    api         1        1       1        0          2023-05-02T21:25:29Z
    dashboard   1        1       1        0          2023-05-02T21:25:29Z
```

nomad/job_endpoint_hook_connect.go
```diff
@@ -230,6 +230,29 @@ func injectPort(group *structs.TaskGroup, label string) {
...
+			return "docker"
+		case "podman":
+			foundPodman = true
+		}
+	}
+	if foundPodman {
+		return "podman"
+	}
+	return "docker"
+}
```
lgfa29, line 254: It's a simple function, so I hate to bikeshed on it 😅 

But I found this logic a bit confusing to grok at first. I think we could instead have a task driver counter map and then something like this so it more explicitly matches the function docstring:

```go
if numTasks["podman"] > 1 && numTasks["docker"] == 0 {
  return "podman"
}

return "docker"
```

nomad/job_endpoint_hook_connect.go
```diff
@@ -477,12 +501,12 @@ func newConnectGatewayTask(prefix, service string, netHost, customizedTls bool)
...
 
-func newConnectSidecarTask(service string) *structs.Task {
+func newConnectSidecarTask(service, driver string) *structs.Task {
 	return &structs.Task{
 		// Name is used in container name so must start with '[A-Za-z0-9]'
 		Name:          fmt.Sprintf("%s-%s", structs.ConnectProxyPrefix, service),
 		Kind:          structs.NewTaskKind(structs.ConnectProxyPrefix, service),
-		Driver:        "docker",
+		Driver:        driver,
 		Config:        connectSidecarDriverConfig(),
```
lgfa29, line 510: Would it be worth adding a docsting to `connectSidecarDriverConfig()` to note that it should be compatible with Docker _and_ Podman? 

Or maybe even pass the driver choice here. It can be ignored in the current implementation, but may be handy if there are conflicting configs between the drivers or we add support for other plugins.

lgfa29 (approved):
> After this change, we should be able to run typical Connect jobspecs
> (e.g. nomad job init [-short] -connect) on Clusters configured with the
> podman task driver, without modification to the job files.

🥳

nomad/job_endpoint_hook_connect.go
shoenig: How about

```go
func groupConnectGuessTaskDriver(g *structs.TaskGroup) string {
    drivers := set.FromFunc(g.Tasks, func(t *structs.Task) string {
        return t.Driver
    })
    if drivers.Contains("podman") && !drivers.Contains("docker") {
        return "podman"
    }
    return "docker"
}
```

nomad/job_endpoint_hook_connect.go
shoenig: Added a comment; with a note about passing in the parameter one day if needed

shoenig:
> Is there a doc we could update

yeah, let me cover that in another PR along with an e2e test

PR #17066, opened 2023-05-02
Fixing max job source size typo

mikenomitch:
Small typo fix in the docs to match the real value. Also had some linter auto-fixes in the markdown.


shoenig:
@mikenomitch Ahh sorry, I actually meant for this to be `job_max_source_size`, because that follows the pattern of other names `job_default_priority` and `job_max_priority`. Do you want to fix it the other way? Otherwise I can push a commit to clean it up.

mikenomitch:
@shoenig no problem - Would be great if you took the fix! I'll close this PR.

PR #17068, opened 2023-05-03
docs: move CNI reference plugins installation to CNI overview page

### LINKED ISSUE #17038: docs: CNI networking docs should contain or at least link to reference plugin installation
After a complete read through of https://developer.hashicorp.com/nomad/docs/networking/cni I still have no instructions or even a link to instructions for installing the CNI reference plugins necessary for Connect / `bridge` networking.

The actual instructions are buried in the Connect integration guide
https://developer.hashicorp.com/nomad/docs/integrations/consul-connect#cni-plugins

shoenig:
This PR moves the instruction steps for install the CNI reference plugins
from the Consul Mesh integration page to the general Networking CNI page.

These plugins are required for bridge networking, not just Consul Mesh,
so it makes sense to have them on the general CNI page.

Closes #17038

### DIFF
--- a/website/content/docs/integrations/consul-connect.mdx
+++ b/website/content/docs/integrations/consul-connect.mdx
@@ -146,37 +146,9 @@ $ sudo nomad agent -dev-connect
 
 ### CNI Plugins
 
-Nomad uses CNI plugins to configure the network namespace used to secure the
+Nomad uses CNI reference plugins to configure the network namespace used to secure the
 Consul service mesh sidecar proxy. All Nomad client nodes using network namespaces
-must have CNI plugins installed.
-
-The following commands install CNI plugins:
-
-```shell-session
-curl -L -o cni-plugins.tgz "https://github.com/containernetworking/plugins/releases/download/v1.0.0/cni-plugins-linux-$( [ $(uname -m) = aarch64 ] && echo arm64 || echo amd64)"-v1.0.0.tgz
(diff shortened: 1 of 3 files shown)


lgfa29 (commented):
To avoid duplicating these instructions in multiple places we added them to the [`Linux` `Post-Install` step](https://developer.hashicorp.com/nomad/docs/install#post-installation-steps) and [linked there from this section](https://github.com/hashicorp/nomad/blob/main/website/content/docs/networking/index.mdx?plain=1#L101).

But we have like 3 different install pages, and the link above is now broken for some reason and the instructions I added in https://github.com/hashicorp/tutorials/pull/1005 are now gone 😞 

So maybe that's not a good approach....

MDX allows [importing other files](https://github.com/hashicorp/nomad/blob/main/website/content/docs/runtime/environment.mdx?plain=1#L16), would it be better to keep these instructions in one specific file and then importing them where needed?

shoenig:
Ooh I didn't even see the Post-Install steps. How about we just keep that as the source of truth and use links? (basically what you had before)

lgfa29 (approved):
LGTM!

Would you mind updating this link as well? It's currently broken...

https://github.com/hashicorp/nomad/blob/06e877f26b08d9a4eceb3e9e94bd1c1f28f6a339/website/content/docs/networking/index.mdx?plain=1#L308

Issue #17070, opened 2023-05-03
Canceled evaluation provide circular or wrong debug information.


stswidwinski:
### Nomad version
Built off tip

### Operating system and Environment details
Unix

### Issue
When an evaluation is canceled, it produces confusing or wrong debug information. In particular:
```
$ nomad eval status  -namespace '*' -json 81235e65
{
    [...]
    "CreateIndex": 6445487,
    "CreateTime": 1682965847060638834,
    [...]
    "ModifyIndex": 6445493,
    "ModifyTime": 1682965847060638834,
    [...]
    "Status": "canceled",
    "StatusDescription": "evaluation \"81235e65-7e71-9595-0a34-00674bff2b75\" successful",
    [...]
}
```
There are a few interesting points:

1. `CreateIndex` and `ModifyIndex` are different but `CreateTime` and `ModifyTime` are the same. In practice the times should be 30  seconds apart.
2. `StatusDescription` reports the successful evaluation as the same evaluation which is canceled. This is not the intended information.

I will submit a proposed patch to fix these fields below.

jrasell:
Hi @stswidwinski thanks for raising this issue and the corresponding PR. We will take a look into this shortly.

tgross:
Fixed in https://github.com/hashicorp/nomad/pull/17071 which will ship in the next release of Nomad 1.5.x, with backports to 1.4.x and 1.3.x.

PR #17071, opened 2023-05-03
[17070] Correct the status description and modify time of canceled evals.

stswidwinski:
Fix for https://github.com/hashicorp/nomad/issues/17070. Corrected the status description and modify time of evals which are canceled due to another eval having completed in the meantime. 

Tests added. It seems that the testing has originally been written a rather long time before the new-style of testing started within the Nomad codebase. Please let me know if you'd rather I moved the entire test to the new style, or if added if clauses seem like a reasonable tradeoff given the size of the change.

### DIFF
--- a/nomad/state/state_store.go
+++ b/nomad/state/state_store.go
@@ -3160,11 +3160,12 @@ func (s *StateStore) nestedUpsertEval(txn *txn, index uint64, eval *structs.Eval
 		}
 
 		// Go through and update the evals
-		for _, eval := range blocked {
-			newEval := eval.Copy()
+		for _, blockedEval := range blocked {
+			newEval := blockedEval.Copy()
 			newEval.Status = structs.EvalStatusCancelled
-			newEval.StatusDescription = fmt.Sprintf("evaluation %q successful", newEval.ID)
+			newEval.StatusDescription = fmt.Sprintf("evaluation %q successful", eval.ID)
 			newEval.ModifyIndex = index
+			newEval.ModifyTime = eval.CreateTime
 
 			if err := txn.Insert("evals", newEval); err != nil {
 				return fmt.Errorf("eval insert failed: %v", err)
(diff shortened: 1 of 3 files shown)


nomad/state/state_store.go
```diff
@@ -3160,11 +3160,12 @@ func (s *StateStore) nestedUpsertEval(txn *txn, index uint64, eval *structs.Eval
...
 		// Go through and update the evals
-		for _, eval := range blocked {
-			newEval := eval.Copy()
+		for _, blockedEval := range blocked {
+			newEval := blockedEval.Copy()
 			newEval.Status = structs.EvalStatusCancelled
-			newEval.StatusDescription = fmt.Sprintf("evaluation %q successful", newEval.ID)
+			newEval.StatusDescription = fmt.Sprintf("evaluation %q successful", eval.ID)
 			newEval.ModifyIndex = index
+			newEval.ModifyTime = eval.CreateTime
```
tgross, line 3168: The Evaluation from the `eval` argument we're upserting is `EvalStatusComplete`, so I'd expect the `eval.ModifyTime` to be updated to be later than the `eval.CreateTime` by this point. And we're updating the `ModifyIndex` to the same `ModifyIndex` as the `eval` argument (as I'd expect). So shouldn't this be?:

```suggestion
			newEval.ModifyTime = eval.ModifyTime
```

stswidwinski: I was convinced that this is what I initially did. Nice catch and sorry for crossing my wires :)

jrasell (approved):
LGTM and thanks @stswidwinski!

Would you be able to run `make cl` to add a changelog entry for this fix; with that I will get this merged.

stswidwinski:
All done. Thank you!

tgross:
Backport assistant screwed up all the backports. I'll do those by hand.

PR #17073, opened 2023-05-03
[ui, deployments] handle node read permissions for system job panel

philrenaud:
We depend on iterating over `nodes` to determine how many allocations should be running in a system job. So what happens when the user's ACL doesn't allow for `node-read` permissions?

An easy solution would be to just drop a "You require node-read" block over top of the panel, like we do for the `JobStatusInClient` panel.

But we _can_ know a bit about the nodes — their IDs, anyway — and we can know that for system jobs at most one alloc is supposed to run per node.

So in this way, we can do something, which is better than nothing: show them the actual state, if not the desired state, of their job allocations.

This drops the "/Y" part of "X/Y allocations running" for the steady state panel when you lack node-read permissions and are viewing a system job. It also can never show "Unplaced" allocs, as it has no idea about which nodes exist that lack them.

image

### DIFF
--- a/ui/app/components/job-status/panel/steady.js
+++ b/ui/app/components/job-status/panel/steady.js
@@ -2,8 +2,10 @@
 import Component from '@glimmer/component';
 import { alias } from '@ember/object/computed';
 import matchGlob from 'nomad-ui/utils/match-glob';
+import { inject as service } from '@ember/service';
 
 export default class JobStatusPanelSteadyComponent extends Component {
+  @service can;
   @alias('args.job') job;
 
   // Build note: allocTypes order matters! We will fill up to 100% of totalAllocs in this order.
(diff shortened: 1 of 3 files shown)


ui/app/components/job-status/panel/steady.js
```diff
@@ -70,23 +72,34 @@ export default class JobStatusPanelSteadyComponent extends Component {
...
-      return filteredNodes.length;
-      // return this.args.job.allocations.uniqBy('node.id').length
-      // return this.args.job.allocations
-      //   .filter((a) => a.jobVersion === this.args.job.version)
-      //   .uniqBy('node.id').length;
+        return filteredNodes.length;
+      } else {
+        // You don't have node read access, so do the best you can: uniqBy node IDs on the allocs you can see.
+        return this.args.job.allocations.uniqBy('nodeID').length;
+      }
```
lgfa29, line 87: Do we need this part? In the template, `totalAllocs` is not called in the code path that would trigger this.

This function also seems to return the _desired_ count of allocations, which is a hard value to derive for system and sysbatch jobs. 

I think I used this node filter by DC approach in the `Job Status in Client` panel but in hindsight I don't know if it provides value of it it's more confusing.

The biggest problem is that DC is only one of the aspects that can affect the desired count. Constraints (either explicit as a `constraint` block in the job, or implicit like the choice of task `driver`) can also affect this count and are hard to determine if they are expected or not. 

For example, if you have a client without Docker, and you know that it doesn't have Docker, you don't expect an alloc with a Docker task to run there, so displaying something like `2/3 Allocations Running` is misleading.

All of this to say that I think I would drop the `/Y` for system and sysbatch jobs in general and just report the running count 🤔

lgfa29 (approved):
LGTM. Just shared some thoughts about figuring out desired alloc count for system/sysbatch jobs. It's not an easy thing to do 😅

ui/app/components/job-status/panel/steady.js
philrenaud: Ended up simplifying it in https://github.com/hashicorp/nomad/pull/17073/commits/847339976e7bc7580f21c1518d959501fb3699e4 to not worry about a desired count, like you suggested.

As to whether or not the else is needed: it's required as part of the `get allocBlocks()` in steady.js; it needs to know what number to fill up to.

lgfa29: > As to whether or not the else is needed: it's required as part of the get allocBlocks() in steady.js; it needs to know what number to fill up to.

I was just looking into that and don't have a good answer...As I mentioned in the sysbatch UI RFC, these types of jobs are a weird bunch 😅 

The main impact to your current work is that `Unplaced` doesn't make sense for system/sysbatch since these jobs are supposed to run wherever is feasible, and we can't really tell what that means outside of the Nomad scheduler.

One way we can sidestep this for now is to have `totalAllocs()` return the number of non-terminal allocs for the job, which I think is what we care for system/sysbatch.

PR #15808, opened 2023-01-17
core: enforce strict steps for clients reconnect

lgfa29:
When a Nomad client that is running an allocation with `max_client_disconnect` set misses a heartbeat the Nomad server will update its status to `disconnected`.

Upon reconnecting, the client will make three main RPC calls:

- `Node.UpdateStatus` is used to set the client status to `ready`.
- `Node.UpdateAlloc` is used to update the client-side information about
  allocations, such as their `ClientStatus`, task states etc.
- `Node.Register` is used to upsert the entire node information,
  including its status.

These calls are made concurrently and are also running in parallel with the scheduler. Depending on the order they run the scheduler may end up with incomplete data when reconciling allocations. 

#15068 already enforced clients to heartbeat before updating their allocation data, but there are still scenarios that can generate wrong results.

For example, a client disconnects and its replacement allocation cannot be placed anywhere else, so there's a pending eval waiting for resources.

When this client comes back the order of events may be:

1. Client calls `Node.UpdateStatus` and is now `ready`.
2. Scheduler reconciles allocations and places the replacement alloc to
   the client. The client is now assigned two allocations: the original
   alloc that is still `unknown` and the replacement that is `pending`.
3. Client calls `Node.UpdateAlloc` and updates the original alloc to
   `running`.
4. Scheduler notices too many allocs and stops the replacement.

node-disconnect drawio

This creates unnecessary placements or, in a different order of events, may leave the job without any allocations running until the whole state is updated and reconciled.

To avoid problems like this clients must update _all_ of its relevant information before they can be considered `ready` and available for scheduling.

To achieve this goal the RPC endpoints mentioned above have been modified to enforce strict steps for nodes reconnecting:

- `Node.Register` does not set the client status anymore.
- `Node.UpdateStatus` sets the reconnecting client to the `initializing`
  status until it successfully calls `Node.UpdateAlloc`.

node-disconnect drawio

These changes are done server-side to avoid the need of additional coordination between clients and servers. Clients are kept oblivious of these changes and will keep making these calls as they normally would.

The verification of whether allocations have been updates is done by storing and comparing the Raft index of the last time the client missed a heartbeat and the last time it updated its allocations.

Closes #15483

PR #17074, opened 2023-05-03
client: de-duplicare alloc updates and gate during restore

### LINKED ISSUE #16381: Node restart causes modify index of all allocs on that node to be bumped
### Nomad version
Nomad v1.4.2

Unix

### Issue
When client nodes are restarted, all allocations that have been scheduled on the node have their modify index changed to the index at the time of restart, including allocations that are in terminal state. This problem affects both allocations that have and have not been GC-ed from the client's local state as long as they are in the scheduler's raft state.

**System Impact**

1. Restarts of nodes at a cadence higher than the GC limits lead to infinite accumulation of data in the system, OOMs of the client and server nodes and high CPU, disk and network load within the entire system.
2. Client node restarts lead to high load on the system – both servers and clients
3. Rolling, splayed restart of client nodes leads to stampeding herd behaviors and cascading failures

**Root Cause**
The rough flow is:
1. On restart the client adds terminal allocs back to its local state
2. It then sends the server updates about these terminal allocs
3. The server writes these allocs back to the raft state bumping the modify index in the process

More details:

1. An alloc runner is created for a terminal allocation

    _Case 1: The terminal allocation has been GC-ed from the client's local state_

    When the client node starts back up, it's `watchAllocations` routine calls the `Node.GetClientAllocs` RPC to request the allocations it  should know about from the server:
    https://github.com/hashicorp/nomad/blob/6f52a9194f1e0e13a5a85f7fbd1317b3df17ede6/client/client.go#L2219

    The handler for `GetClientAllocs` calls `AllocsByNode` to get all the allocations that have been scheduled on the node (including those that are already terminal):
    https://github.com/hashicorp/nomad/blob/6f52a9194f1e0e13a5a85f7fbd1317b3df17ede6/nomad/node_endpoint.go#L1152

    The `watchAllocations` enriches those allocations returned by the server and puts them into the `allocUpdates` channel:
    https://github.com/hashicorp/nomad/blob/6f52a9194f1e0e13a5a85f7fbd1317b3df17ede6/client/client.go#L2102
 

    Which is then read and handled by `runAllocs`:
    https://github.com/hashicorp/nomad/blob/6f52a9194f1e0e13a5a85f7fbd1317b3df17ede6/client/client.go#L1854-L1864

    Since the allocations that have already been GC-ed from client's local state, they are not in the client's `allocs` and thus will be considered as allocations to be added (see client logs from repro below):
    https://github.com/hashicorp/nomad/blob/6f52a9194f1e0e13a5a85f7fbd1317b3df17ede6/client/client.go#L2441

    Which does a write to the client's local state DB:
    https://github.com/hashicorp/nomad/blob/6f52a9194f1e0e13a5a85f7fbd1317b3df17ede6/client/client.go#L2583
    (Note that this write is actually redundant since it will immediately be marked for GC once client has breached its max allocs)

    And creates an alloc runner for the allocation:    
    https://github.com/hashicorp/nomad/blob/6f52a9194f1e0e13a5a85f7fbd1317b3df17ede6/client/client.go#L2633

    _Case 2: The terminal allocation has not been GC-ed from the client's local state_

    When the client node starts back up, it calls `restoreState`, which finds the alloc in it's local state DB, and creates an alloc runner for it:
    https://github.com/hashicorp/nomad/blob/6f52a9194f1e0e13a5a85f7fbd1317b3df17ede6/client/client.go#L1250

4. The alloc runner realises that the allocation is terminal, and schedules an update to the server about the allocation before terminating
https://github.com/hashicorp/nomad/blob/6f52a9194f1e0e13a5a85f7fbd1317b3df17ede6/client/allocrunner/alloc_runner.go#L627

5. The server receives the update about the allocation, and writes it to the raft state bumping the modify index in the process
https://github.com/hashicorp/nomad/blob/5d5740b5339cdd26e34f6f68d09a995471ee6734/nomad/state/state_store.go#L3681

If the number of terminal allocations for the node is large, then the client does a large number of redundant writes (both additions and deletions) to the state DB, and the server has to send large number of raft messages to update allocations in the raft state.

### Reproduction steps
```
# Server config
$ cat server.conf
data_dir = "/tmp/nomad/server"

advertise {
  http = "127.0.0.1"
  rpc = "127.0.0.1"
  serf = "127.0.0.1"
}

server {
  enabled = true
  bootstrap_expect = 1
}

# Start the server
$ nomad agent -config server.conf

# Client config
$ cat client.conf
data_dir = "/tmp/nomad/client"
log_level = "debug"

advertise {
  http = "127.0.0.1"
  rpc = "127.0.0.1"
  serf = "127.0.0.1"
}

ports {
  http = "9876"
  rpc = "9875"
  serf = "9874"
}

client {
  enabled = true
  servers = ["127.0.0.1"]
  gc_max_allocs = 1
}

plugin "raw_exec" {
  config {
    enabled = true
  }
}

# Start the client
$ nomad agent -config client.conf

# Show the job specs, they are identical except for the job name
$ cat example-1.nomad
job "example-1" {
  datacenters = ["dc1"]
  type = "batch"

  group "test-group" {
    task "test-task" {
      driver = "raw_exec"

      config {
	command = "/usr/bin/true"
      }
    }

    reschedule {
      attempts  = 0
      unlimited = false
    }
  }
}
$ cat example-2.nomad
job "example-2" {
  datacenters = ["dc1"]
  type = "batch"

  group "test-group" {
    task "test-task" {
      driver = "raw_exec"

      config {
	command = "/usr/bin/true"
      }
    }

    reschedule {
      attempts  = 0
      unlimited = false
    }
  }
}

# Run both jobs - the second allocation causes the first to get GC-ed (because we set gc_max_allocs to 1)
$ nomad job run example-1.nomad
==> 2023-03-07T22:42:56+08:00: Monitoring evaluation "4ba37e42"
    2023-03-07T22:42:56+08:00: Evaluation triggered by job "example-1"
    2023-03-07T22:42:57+08:00: Allocation "d538290d" created: node "a52d4f78", group "test-group"
    2023-03-07T22:42:57+08:00: Evaluation status changed: "pending" -> "complete"
==> 2023-03-07T22:42:57+08:00: Evaluation "4ba37e42" finished with status "complete"
$ nomad job run example-2.nomad
==> 2023-03-07T22:43:03+08:00: Monitoring evaluation "11aa17f4"
    2023-03-07T22:43:03+08:00: Evaluation triggered by job "example-2"
    2023-03-07T22:43:03+08:00: Allocation "f8720e5e" created: node "a52d4f78", group "test-group"
    2023-03-07T22:43:04+08:00: Allocation "f8720e5e" status changed: "pending" -> "complete" (All tasks have completed)
    2023-03-07T22:43:04+08:00: Evaluation status changed: "pending" -> "complete"
==> 2023-03-07T22:43:04+08:00: Evaluation "11aa17f4" finished with status "complete"
 
# Kill client...

# Get the modify index of the allocations, and show that they are terminal
$ nomad alloc status -json d538290d | jq '.JobID,.ClientStatus,.ModifyIndex'
"example-1"
"complete"
18
$ nomad alloc status -json f8720e5e | jq '.JobID,.ClientStatus,.ModifyIndex'
"example-2"
"complete"
19

# Start client back up
$ nomad agent -config client.conf

# Show that the allocations have their modify index bumped
$ nomad alloc status -json d538290d | jq '.JobID,.ClientStatus,.ModifyIndex'
"example-1"
"complete"
37
$ nomad alloc status -json f8720e5e | jq '.JobID,.ClientStatus,.ModifyIndex'
"example-2"
"complete"
37 
```

Some interesting/relevant logs from the nomad client on the second startup, which shows that it thinks it needs to add the terminal alloc:
```
2023-03-07T22:47:05.749+0800 [INFO]  client.gc: marking allocation for GC: alloc_id=f8720e5e-9e78-fbc2-1072-310bb8fd5546
2023-03-07T22:47:05.750+0800 [DEBUG] client: updated allocations: index=19 total=2 pulled=1 filtered=1
2023-03-07T22:47:05.750+0800 [DEBUG] client: allocation updates: added=1 removed=0 updated=0 ignored=1
2023-03-07T22:47:05.750+0800 [INFO]  client.gc: garbage collecting allocation: alloc_id=f8720e5e-9e78-fbc2-1072-310bb8fd5546 reason="new allocations and over max (1)"
```

#### Expected Result
Modify index of terminal allocations should not be updated

#### Actual Result
The modify index of terminal allocations were updated

tgross:
When client nodes are restarted, all allocations that have been scheduled on the node have their modify index updated, including terminal allocations. There several contributing factors:

* The `allocSync` method that updates the servers isn't gated on first contact with the servers. This means that if a server updates the desired state while the client is down, the `allocSync` races with the `Node.ClientGetAlloc` RPC. This will typically result in the client updating the server with "running" and then immediately thereafter "complete".

* The `allocSync` method unconditionally sends the `Node.UpdateAlloc` RPC even if it's possible to assert that the server has definitely seen the client state. The allocrunner may queue-up updates even if we gate sending them. So then we end up with a race between the allocrunner updating its internal state to overwrite the previous update and `allocSync` sending the bogus or duplicate update.

This changeset adds tracking of server-acknowledged state to the allocrunner. This state gets checked in the `allocSync` before adding the update to the batch, and updated when `Node.UpdateAlloc` returns successfully. To implement this we need to be able to equality-check the updates against the last acknowledged state. We also need to add the last acknowledged state to the client state DB, otherwise we'd drop unacknowledged updates across restarts.

The client restart test has been expanded to cover a variety of allocation states, including allocs stopped before shutdown, allocs stopped by the server while the client is down, and allocs that have been completely GC'd on the server while the client is down. I've also bench tested scenarios where the task workload is killed while the client is down, resulting in a failed restore.

Fixes #16381


client/client.go
```diff
@@ -2144,10 +2153,32 @@ func (c *Client) allocSync() {
...
+				if ar, ok := c.allocs[allocID]; ok {
+					if !ar.LastAcknowledgedStateIsCurrent(update) {
+						sync = append(sync, update)
+					}
+				} else {
+					// no allocrunner (typically a failed placement), so we need
+					// to send update
+					sync = append(sync, update)
+				}
+			}
```
shoenig, line 2175: nit: can you pull this into a helper function like `deduplicateUpdates(updates, allocs) sync`

easier to read and write full coverage tests that way

client/client.go
```diff
@@ -2144,10 +2153,32 @@ func (c *Client) allocSync() {
...
+			if len(sync) == 0 {
+				// No updates to send
+				updates = make(map[string]*structs.Allocation, len(updates))
+				syncTicker.Stop()
+				syncTicker = time.NewTicker(allocSyncIntv)
```
shoenig, line 2180: do we need a new one instead of `Reset()`?

client/client.go
```diff
@@ -2144,10 +2153,32 @@ func (c *Client) allocSync() {
...
+			case <-c.serversContactedCh:
+			default:
+				continue
+			}
 
 			sync := make([]*structs.Allocation, 0, len(updates))
-			for _, alloc := range updates {
-				sync = append(sync, alloc)
+			for allocID, update := range updates {
+				if ar, ok := c.allocs[allocID]; ok {
```
shoenig, line 2166: c.allocs needs to hold `allocLocks` right?

nomad/structs/structs.go
```diff
@@ -8437,6 +8437,35 @@ func (ts *TaskState) Successful() bool {
...
+	}
+	if ts.FinishedAt != o.FinishedAt {
+		return false
+	}
+	if len(ts.Events) != len(o.Events) {
+		return false
+	}
+	if ts.TaskHandle != o.TaskHandle {
+		return false
+	}
```
shoenig, line 8464: `Events` and `TaskHandle` are complex structs - should we not defer to `Equal()` implementation on each of them?

nomad/structs/structs.go
```diff
@@ -11257,6 +11286,41 @@ func (a *AllocNetworkStatus) Copy() *AllocNetworkStatus {
...
+func (a *AllocNetworkStatus) IsZero() bool {
+	if a == nil {
+		return true
+	}
+	if a.InterfaceName != "" || a.Address != "" {
+		return false
+	}
+	if a.DNS != nil && (len(a.DNS.Options) > 0 || len(a.DNS.Searches) > 0 || len(a.DNS.Servers) > 0) {
+		return false
+	}
```
shoenig, line 11320: reaching into implementation detail of `DNS` - how about adding `DNS.IsZero()` and deferring to that

nomad/structs/structs.go
```diff
@@ -11333,6 +11397,24 @@ func (a *AllocDeploymentStatus) Copy() *AllocDeploymentStatus {
...
 }
 
+func (a *AllocDeploymentStatus) Equal(o *AllocDeploymentStatus) bool {
+	if a == nil || o == nil {
+		return a == o
+	}
+
+	switch {
+	case *a.Healthy != *o.Healthy:
+		return false
```
shoenig, line 11407: possible nil pointer dereference? 

safer with https://github.com/hashicorp/nomad/blob/main/helper/pointer/pointer.go#L44

shoenig (commented):
Added some suggestions; will make a second pass later on

client/client.go
tgross: Oof, yes. Will fix.

client/client.go
```diff
@@ -2167,6 +2198,20 @@ func (c *Client) allocSync() {
 				continue
 			}
 
+			c.allocLock.RLock()
+			for _, update := range sync {
+				if ar, ok := c.allocs[update.ID]; ok {
+					ar.AcknowledgeState(&arstate.State{
```
schmichael, line 2204: If you wanted to avoid storing a bunch of duplicated client alloc state and implementing a deep equality, I think we could make this a `uint64`, increment it on every `Client.AllocStateUpdated` call, and have the Client persist it here.

That relies on ARs/TRs not calling `Client.AllocStateUpdated` even when nothing has been updated though, so I think your approach is safer.

schmichael (approved):
Great work fixing a long standing issue!

Due to #15808 I think this will cause a reconnecting Client to remain in `initializing` until `Node.GetClientAllocs` returns since it relies on `Node.UpdateAlloc` being called as a signal that a node is ready again.

I can't imagine how this is an issue. I don't think there's any opportunity for further simplification or cleanup either, but I wanted to call it out in case somebody else can think of a way this effects behavior.

client/client.go
```diff
@@ -1270,6 +1272,13 @@ func (c *Client) restoreState() error {
 			continue
 		}
 
+		allocState, err := c.stateDB.GetAcknowledgedState(alloc.ID)
+		if err != nil {
+			c.logger.Error("error restoring last acknowledged alloc state, will update again",
+				err, "alloc_id", alloc.ID)
+		}
+		ar.AcknowledgeState(allocState)
```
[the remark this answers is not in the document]
tgross: It'll still be nil in this case, so probably better just to put the `ar.AcknowledgeState` in the `else` block for that error so that if we change `AcknowledgeState` later it's that much harder to introduce a NPE.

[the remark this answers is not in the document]
tgross: Resolved in f62c33f7e4fc40db35ed3287d8089e6258a3fe58

client/client.go
tgross: Resolved in f62c33f7e4fc40db35ed3287d8089e6258a3fe58

nomad/structs/structs.go
tgross: Good call. Resolved in f62c33f7e4fc40db35ed3287d8089e6258a3fe58

tgross:
Somehow in addressing comments I've managed to break the test. Will need to figure out what I broke here :blush:

tgross:
I've made one more pass over this and added an additional unit test for the check so that we have a little more coverage of the equality check. Will merge once I've got the :heavy_check_mark: from CI and this will go out with the 1.6.0 release.

PR #17077, opened 2023-05-04
docs: update artifact jobspec sshkey example path.

### LINKED ISSUE #17027: Documentation - Wrong file specified for sshkey in git artifact example
In the second example of the [Artifact Block > Download using Git](https://developer.hashicorp.com/nomad/docs/job-specification/artifact#download-using-git) section the `ssh_known_hosts` file is shown instead of the private SSH key used to connect to the git repository.

Rather than
```
    sshkey = "${base64encode(file("/etc/ssh/ssh_known_hosts"))}"
```
I would have expected it to be something like
```
    sshkey = "${base64encode(file("/path/to/privatekey"))}"
```

jrasell:
closes #17027


tgross (approved):
LGTM

PR #17092, opened 2023-05-05
scale: fixed a bug where evals could be created with wrong type.

### LINKED ISSUE #17091: scale: eval creation on scaling assuming service job type
Tested on main, locally on M2 MacOS.

When scaling a job, the resulting evaluation is always created with `Type = "Service"` regardless of the type parameter within the jobspec that is being scaled. This can result in jobs of type batch being scaled and the created evaluation detailing type service.

JobSpec:
```
job "example" {
  datacenters = ["dc1"]
  type        = "batch"

  group "cache" {
    count = 2
    network {
      port "db" {
        to = 6379
      }
    }
    task "redis" {
      driver = "docker"
      config {
        image = "redis:7"
        ports = ["db"]
      }
    }
  }
}
```

When scaling the above job via `nomad job scale example 10` the following evaluation is created.
```
ID                 = 97bc2abd
Create Time        = 31s ago
Modify Time        = 31s ago
Status             = complete
Status Description = complete
Type               = service
TriggeredBy        = job-scaling
Job ID             = example
Namespace          = default
Priority           = 50
Placement Failures = false
```

jrasell:
The job scale RPC endpoint hard-coded the eval creation to use the type of service. This meant scaling events triggered on jobs of type batch would create evaluations with the wrong type, which does not seem to cause any problems, just confusion when correlating the two.

closes #17091

### DIFF
--- a/.changelog/17092.txt
+++ b/.changelog/17092.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+scale: Fixed a bug where evals could be created with the wrong type
+```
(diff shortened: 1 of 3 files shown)


tgross (approved):
LGTM

PR #17097, opened 2023-05-05
api: set the job submission during job reversion

shoenig:
This PR fixes a bug where the job submission would always be nil when
a job goes through a reversion to a previous version. Basically we need
to detect when this happens, lookup the submission of the job version
being reverted to, and set that as the submission of the new job being
created.

No CL/BP; this is for a 1.6 feature


tgross (approved):
LGTM

PR #17103, opened 2023-05-05
Standardize on GitHub Actions for CI

hc-github-team-es-release-engineering:
The primary goal here is to migrate our remaining CI workflows from Circle CI to GitHub Actions, namely: test-e2e, test-ui, and test-windows.

Related test changes to enable that were split out to separate PRs: #17401 and #17399

Extra-curricularly, as part of the overall migration effort company-wide, this also includes some standardization such as:
 * explicit permissions:read on various workflows
 * pinned action version shas (per [hashicorp/security-public-tsccr](https://github.com/hashicorp/security-public-tsccr))
 * actionlint, which among other things runs shellcheck on GHA run steps

### DIFF
--- a/.github/actionlint.yml
+++ b/.github/actionlint.yml
@@ -1,0 +1,14 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+self-hosted-runner:
+  # Labels of self-hosted runner in array of string
+  labels:
+    - custom
+    - xl
+    - 22.04
+    - linux
+    - xxl
+    - 20.04
+    - custom-windows-medium
+    - windows-2019-16core
--- a/.github/actions/install-vault/action.yml
+++ b/.github/actions/install-vault/action.yml
@@ -1,0 +1,33 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+name: install-vault
+inputs:
+  version:
+    required: false
+    default: 1.2.4
+runs:
+  using: composite
+  steps:
+  - name: Install Vault
+    run: |-
+      set -x
+
+      case "${OSTYPE}" in
+        linux*)  os="linux" ;;
+        darwin*) os="darwin" ;;
+        msys*)   os="windows" ;;
+        *)       echo "unknown os: ${OSTYPE}"; exit 1 ;;
+      esac
+
+      curl -SL --fail -o /tmp/vault.zip https://releases.hashicorp.com/vault/"${{ inputs.version }}"/vault_"${{ inputs.version }}"_"${os}"_amd64.zip
+
+      if [[ "${os}" != "windows" ]]; then
+        sudo unzip -d /usr/local/bin /tmp/vault.zip
+        rm -rf /tmp/vault*
+      else
+        rm -rf c:\Windows\vault.exe
+        unzip /tmp/vault.zip -d "/c/Windows"
+        rm -rf /tmp/vault*
+      fi
+    shell: bash
--- a/.github/workflows/ember-assets.yml
+++ b/.github/workflows/ember-assets.yml
@@ -22,3 +22,5 @@ jobs:
         with:
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
           working-directory: "ui"
+permissions:
+  contents: read
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -15,4 +15,7 @@ jobs:
     if: (github.actor != 'dependabot[bot]')
     steps:
       - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
-      - uses: returntocorp/semgrep-action@245bf11ddb2f3d4e35f116608cf6e27ae0f9aa04 # v1
+      - uses: returntocorp/semgrep-action@8689a93eca50c91f8e8cfa61e3c4fb7e96d12bf2 # v1
+permissions:
+  contents: read
+
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -1,0 +1,48 @@
+name: test-e2e
+on:
+  pull_request:
+    paths-ignore:
+      - 'README.md'
+      - 'CHANGELOG.md'
+      - '.changelog/**'
+      - '.tours/**'
+      - 'contributing/**'
+      - 'demo/**'
+      - 'dev/**'
+      - 'integrations/**'
+      - 'pkg/**'
+      - 'scripts/**'
+      - 'terraform/**'
+      - 'ui/**'
+      - 'website/**'
+  push:
+    branches:
+      - main
+      - release/**
+    paths-ignore:
+      - 'README.md'
+      - 'CHANGELOG.md'
+      - '.changelog/**'
+      - '.tours/**'
+      - 'contributing/**'
+      - 'demo/**'
+      - 'dev/**'
+      - 'integrations/**'
+      - 'pkg/**'
+      - 'scripts/**'
+      - 'terraform/**'
+      - 'ui/**'
+      - 'website/**'
+
+jobs:
+  test-e2e:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+      - name: Setup go
+        uses: actions/setup-go@4d34df0c2316fe8122ab82dc22947d607c0c91f9 # v4.0.0
+      - run: make deps
+      - run: make integration-test
+      - run: make e2e-test
+permissions:
+  contents: read
--- a/.github/workflows/test-ui.yml
+++ b/.github/workflows/test-ui.yml
@@ -1,0 +1,57 @@
+name: test-ui
+on:
+  pull_request:
+    paths-ignore:
+      - 'README.md'
+      - 'CHANGELOG.md'
+      - '.changelog/**'
+      - '.tours/**'
+      - 'contributing/**'
+      - 'demo/**'
+      - 'dev/**'
+      - 'e2e/**'
+      - 'integrations/**'
+      - 'pkg/**'
+      - 'scripts/**'
+      - 'terraform/**'
+      - 'website/**'
+  push:
+    branches:
+      - main
+      - release/**
+      - test-ui
+    paths-ignore:
+      - 'README.md'
+      - 'CHANGELOG.md'
+      - '.changelog/**'
+      - '.tours/**'
+      - 'contributing/**'
+      - 'demo/**'
+      - 'dev/**'
+      - 'e2e/**'
+      - 'integrations/**'
+      - 'pkg/**'
+      - 'scripts/**'
+      - 'terraform/**'
+      - 'website/**'
+
+jobs:
+  test-ui:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+      - name: yarn install
+        run: cd ui && yarn install --frozen-lockfile
+      - name: lint:js
+        run: cd ui && yarn run lint:js
+      - name: lint:hbs
+        run: cd ui && yarn run lint:hbs
+      - name: Ember tests
+        env:
+          PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
+        run: mkdir -p /tmp/test-reports && cd ui && yarn test
+      - uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
+        with:
+          path: "/tmp/test-reports"
+permissions:
+  contents: read
--- a/.github/workflows/test-windows.yml
+++ b/.github/workflows/test-windows.yml
@@ -1,0 +1,97 @@
+name: test-windows
+on:
+  pull_request:
+    branches:
+      - main
+      - release/**
+    paths-ignore:
+      - 'README.md'
+      - 'CHANGELOG.md'
+      - '.changelog/**'
+      - '.tours/**'
+      - 'contributing/**'
+      - 'demo/**'
+      - 'dev/**'
+      - 'e2e/**'
+      - 'integrations/**'
+      - 'pkg/**'
+      - 'scripts/**'
+      - 'terraform/**'
+      - 'ui/**'
+      - 'website/**'
+  push:
+    branches:
+      - main
+      - release/**
+    paths-ignore:
+      - 'README.md'
+      - 'CHANGELOG.md'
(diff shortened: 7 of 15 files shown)


.github/workflows/build-test.yml
```diff
@@ -0,0 +1,134 @@
+name: build-test
+on:
+  push:
+    branches:
+      - releng/fix-circle-migration
+      - '!**-ui*'
+      - '!docs-*'
+      - '!backport/docs-*'
+      - '!stable-website'
```
tgross, line 9: This should be `on: pull_request`, and `on: push` for the branches we've filtered for in test-core.yaml

.github/actions/install-vault/action.yml
```diff
@@ -0,0 +1,30 @@
...
+
+      if [[ "${os}" != "windows" ]]; then
+        sudo unzip -d /usr/local/bin /tmp/vault.zip
+        rm -rf /tmp/vault*
+      else
+        rm -rf c:\Windows\vault.exe
+        unzip /tmp/vault.zip -d "/c/Windows"
+        rm -rf /tmp/vault*
+      fi
+    shell: bash
```
tgross, line 30: Heads up that whatever automation we're using to copy these from CircleCI to GitHub is missing the trailing newline.

gulducat:
Converted this to a draft to signal that it should not be merged while our CI friends and myself are out all next week.

tgross:
@gulducat this will need to get rebased because having to land #17218 #17222 #17238 this week

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,169 @@
...
+      - 'scripts/**'
+      - 'terraform/**'
+      - 'ui/**'
+      - 'website/**'
+
+env:
+  VAULT_VERSION: 1.4.1
+  PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
+jobs:
+  build-binaries:
```
tgross, line 45: The GitHub Actions we've migrated previously already include a job named `compile` (in `workflows/test-core.yml`). Doesn't this job overlap significantly with that one? If there's any differences here can we merge them to that `compile` job (alternately, should `compile` be pulled out to a separate workflow?)

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,169 @@
+name: build-test
```
tgross, line 1: Calling this `build-test` ports over the CircleCI name directly, but we already have a significant footprint of GHA tests so this name doesn't tell us why it's in its own workflow.

.github/workflows/ember-assets.yml
```diff
@@ -1,15 +1,11 @@
...
 on:
   pull_request:
     paths:
-    - 'ui/**'
-
+      - 'ui/**'
 jobs:
   build:
-
     runs-on: ubuntu-latest
```
tgross, line 8: Is there some kind of yaml linter being used in this PR? There's a lot of fairly arbitrary line and spacing changes going on here.

.github/workflows/build-test.yml
emilymianeil: It doesn't loop like `test-api` is reliant on `compile` but afaik the `compile` and `build-binaries` steps are slightly different:

1. `compile` - calls `make bootstrap && make dev`
2. `build-binaries` - compiles `ui` static assets then makes binaries

I think the biggest difference being that `build-binaries` needs that `ui` static assets steps, so I'm not sure they can be combined.
I think why `build-binaries` even exists in the first place (despite not being used downstream in the current workflow) is to fetch in the `nomad-e2e` repo [here](https://github.com/hashicorp/nomad-e2e/blob/main/circleci/scripts/fetch-binaries.sh)

.github/workflows/build-test.yml
emilymianeil: Yeah I totally agree - there doesn't seem to be a clear reason as to why these tests are separated from the rest in `test-core`? Unsure if y'all want to keep the split or just combine them all into one workflow?

.github/workflows/build-test.yml
tgross: > I think the biggest difference being that build-binaries needs that ui static assets steps, so I'm not sure they can be combined.
> I think why build-binaries even exists in the first place (despite not being used downstream in the current workflow)

Hm, good point. The `compile` ends up being shorter (because compiling the UI takes so long), which is handy for unit tests. But we need the `build-binaries` with the UI to be available for launching our nightly E2E tests. So it might be worth filtering the `build-binaries` down to running just on the `main` and the `release/*` branches. 

Actually making those binaries available to the `nomad-e2e` repo's actions is a whole other question for y'all to solve :grinning: but it might be worth considering moving the `nomad-e2e` nightly runs into `nomad-enterprise` so that we can just use the binaries we build in this action for the nightly run.

.github/workflows/build-test.yml
emilymianeil: Ah ok gotcha - makes sense. I've gone ahead and split these tests out by _function_ then to hopefully add more clarity:

1. e2e
2. ui
3. windows

**note the e2e includes the builds that are used by the `nomad-e2e` repo

.github/workflows/build-test.yml
emilymianeil: Yeah this is something we'll have to chat through when Daniel gets back - I think moving the `nomad-e2e` actions into `ent` probably makes sense since it's using those binaries anyways. I've left it for the time being with the caveat that we're moving on to the `nomad-e2e` repo next anyways, so we'll revisit this then.

drivers/shared/capabilities/defaults.go
```diff
@@ -36,7 +36,6 @@ func NomadDefaults() *Set {
 // https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities
 func DockerDefaults() *Set {
 	defaults := NomadDefaults()
-	defaults.Add("NET_RAW")
 	return defaults
```
dekimsey, line 39: Just wanting to confirm this removal is necessary and appropriate. I thought this was in place specifically because on Linux Docker grants NET_RAW which is problematic and Nomad explicitly removes it. So I'm not sure if removing it from here is intended.

dekimsey (approved):
Looks good.

I noted one issue I want us to double-check on. And a clean-up item to remove the references to `convert-hashicorp-nomad-to-actions-20230505-212504` prior to merging

drivers/shared/capabilities/defaults.go
tgross: @gulducat any code change we're making in this PR should get pulled out into its own PR so those details can be hammered out (and so that we're not comingling commits of OSS production code with internal tooling code).

gulducat: yep, agreed.  I vaguely recall before I was out on vacay/illness that we found this to be unnecessary (with windows 2019 instead of 2022 iirc), so hopefully we can just drop it, but otherwise yes we'll move any functional changes like this to separate PR(s).

.github/workflows/test-windows.yml
```diff
@@ -0,0 +1,97 @@
+name: test-windows
+on:
+  pull_request:
+    paths-ignore:
```
gulducat, line 4: here I think we can use [paths:](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths) instead of paths-ignore, since the gotestsum command targets these specific packages:

```
nomad/drivers/docker
nomad/client/lib/fifo
nomad/client/logmon
```

(note there is a "nomad" dir in this nomad repo, so do need the leading "nomad/")

.github/workflows/test-ui.yml
```diff
@@ -0,0 +1,62 @@
...
+      - 'contributing/**'
+      - 'demo/**'
+      - 'dev/**'
+      - 'e2e/terraform/**'
+      - 'e2e/ui/**'
+      - 'integrations/**'
+      - 'pkg/**'
+      - 'scripts/**'
+      - 'terraform/**'
+      - 'ui/**'
```
gulducat, line 18: def don't want to ignore `ui/` here, nor `e2e/ui/` probably

gulducat (changes requested):
Getting pretty close I think!  We should double-check our "paths-ignore" for the new workflows, and I have a couple other questions/comments for ya.

.github/workflows/ember-assets.yml
gulducat: This PR adds `.github/workflows/actionlint.yml` => [rhysd/actionlint](https://github.com/rhysd/actionlint), which does many things, but it doesn't appear to do generic yaml linting, so I wonder if these might be from a person's IDE?

drivers/shared/capabilities/defaults.go
gulducat: I was mistaken, we can't do nothing, but I can move the change into the test harness instead: [e677ac2](https://github.com/hashicorp/nomad/pull/17103/commits/e677ac2b90ef5f5ff2c721dc5615758e49996069) -- any objections to that until we get around to addressing the real issue?

philrenaud (approved):

.github/actionlint.yml
```diff
@@ -0,0 +1,11 @@
+self-hosted-runner:
```
tgross, line 1: Something I just noticed we're missing: all the new files will need copyright headers so that we don't get a follow-up PR from the copyright bot

.github/workflows/semgrep.yml
```diff
@@ -15,4 +15,7 @@ jobs:
     if: (github.actor != 'dependabot[bot]')
     steps:
       - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
-      - uses: returntocorp/semgrep-action@245bf11ddb2f3d4e35f116608cf6e27ae0f9aa04 # v1
+      - uses: returntocorp/semgrep-action@8689a93eca50c91f8e8cfa61e3c4fb7e96d12bf2 # v1
+permissions:
+  contents: read
+
```
tgross, line 21: We're making a lot of changes in this PR to existing actions that have nothing to do with the CircleCI migration. There's no reasoning why in the PR description (which I expect we'll update with human-authored description at some point?). I don't think any of this belongs here.

.github/workflows/test-ui.yml
tgross: `e2e/ui` is only for E2E tests. There's no reason to run anything in `e2e/` for this workflow.

tgross (changes requested):

.github/actionlint.yml
dekimsey: Copyright wasn't a work item on the migration [ticket](https://github.com/hashicorp/GHA-migration/issues/241), which is why it wasn't in scope/addressed. When I run the plan locally (`copywrite headers --plan --spdx MPL-2.0`) it doesn't flag any of the .github files. So this may not be an issue

.github/workflows/semgrep.yml
dekimsey: Hi Tim, these changes were added as part of the "Restrict Workflow Permissions" work item on the GHA migration [ticket](https://github.com/hashicorp/GHA-migration/issues/241). It was an automated change introduced by the bot and has been done in all migrated repositories to date.

We can certainly link back to that ticket if you'd like too, is there some kind of details or information you are looking to see in the description?

.github/workflows/test-e2e.yml
```diff
@@ -0,0 +1,95 @@
...
+      - 'e2e/ui/**'
+      - 'integrations/**'
+      - 'pkg/**'
+      - 'scripts/**'
+      - 'terraform/**'
+      - 'ui/**'
+      - 'website/**'
+
+jobs:
+  build-binaries:
```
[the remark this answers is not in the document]
dekimsey: You know... I have no idea. I'm gonna wait to see if Emily knows. But upon reading this workflow, it accomplishes nothing

.github/actionlint.yml
dekimsey: I'd like to suggest maybe that get fixed outside of this PR as the plan appears to flag a large number of other files in the repository. Perhaps as part of an effort to also enable a lint action (we have a .golangci.yml, but no lint workflow in the CI)?

.github/workflows/semgrep.yml
tgross: That the internal ticket is poorly scoped to include unrelated work isn't really relevant.  Each chunk of work should be in its own PR, just as we would for any other work on Nomad.

This is an open source project and changes should be as transparent as possible to contributors -- not to mention Nomad Engineering contributors, for whom that link or statement of work had no link to this PR.

.github/actionlint.yml
dekimsey: Added copyright to several of the files that were being picked up by the cli tool

.github/workflows/test-e2e.yml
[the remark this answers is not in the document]
emilymianeil: Ah this is the part where to build binaries for the `nomad-e2e` repo to read 😄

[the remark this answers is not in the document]
emilymianeil: I suppose we could just grab built artifacts off of the `build.yml` build as well - although I'm not entirely sure that the build processes are the same, so it could introduce unexpected behaviour.
If we do keep this, I think it's probably worth adding a note as to why these builds appear to be unused.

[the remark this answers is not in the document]
dekimsey: Myself, I like the idea of us pulling build.yml's artifacts since then we'd be e2e testing the binaries we'd ship

.github/actionlint.yml
dekimsey: @tgross Do the changes I've mentioned above satisfy this concern?

In short:
1. Add copyright headers to files created here.
2. Introduce a separate PR to add a lint check for the project going forward.

.github/workflows/test-windows.yml
gulducat: I'm second-guessing myself now, because those packages have dependencies on other things (also would not need leading "nomad/" because those packages are not in the nomad dir...)

So I think instead we stick with the same list as test-core, though I wish we didn't have to maintain the same list separately...

.github/actionlint.yml
tgross: Works for me!

.github/workflows/semgrep.yml
gulducat: I updated the PR description to clearly state the goals, both primary (migration) and secondary (standardization), and the resulting changes.

I agree with you in general about scope, but I'm hoping the description (which will be commit-message'd too after squash) is now sufficiently clear for us and for our open source community, which I take to be your main concern, laudably.  How's it look to ya?

gulducat (approved):
:rocket:

.github/workflows/semgrep.yml
gulducat: Alrighty, we removed all the non-workflow code here (and I updated the description to point to those PRs for posterity).  How's it look to ye now?

tgross (approved):
:shipit:

PR #17104, opened 2023-05-06
Fix task leak during client restore when allocrunner prerun hook fails

### LINKED ISSUE #17102: Task leak if alloc prerun hook fails after client restart
If a prerun hook fails when restoring alloc state, as with a client agent restart, tasks don't get fully cleaned up and may leave orphan resources like a running container and network configuration (e.g. iptables rules).

This was pointed out in #13028 where specifically a CSI prerun hook fails, but it's an issue more generally with alloc runner prerun hooks.

I encountered it myself while investigating that issue, and as @ygersie put it,

> The worst thing is that Nomad garbage collects the failed allocation but doesn't actually shutdown the docker container (checked the docker logs it never received the api call to stop it either), leaving a zombie container.

### Reproduction steps

I made a strange hook to be able to poison an alloc on disk, so it can succeed first pass but fail after a client agent stop/start.

#### Expected Result

All of the failed task's resources are cleaned up.

#### Actual Result

The task is marked as failed and dead and gets replaced, but the old container remains running.

Also if the task uses a static port, the new one will fail to start because the port is held by the "failed" task.

gulducat:
Fixes #17102 -- I describe the issue more fully there.

My approach here is to stop skipping over `allocRunner.runTasks()` on prerun error.  That way, instead of duplicating more cleanup code, which may change in the future, the same `TaskRunner.Run()` code that usually handles task cleanup can do what it needs to as appropriate with tasks that fail prerun during the alloc restore process.

In pursuit of that, I made an error-inducing `FailHook` and added the ability to include it as part of client Config for the client integration test.  I could remove the non-Prerun interface implementations, but I figured while I'm at it, may as well make a thing that can be induced to fail at any stage in case it's useful?

### DIFF
--- a/client/allocrunner/taskrunner/task_runner_test.go
+++ b/client/allocrunner/taskrunner/task_runner_test.go
@@ -16,6 +16,12 @@ import (
 	"time"
 
 	"github.com/golang/snappy"
+	"github.com/kr/pretty"
+	"github.com/shoenig/test"
+	"github.com/shoenig/test/must"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
 	"github.com/hashicorp/nomad/ci"
 	"github.com/hashicorp/nomad/client/allocdir"
 	"github.com/hashicorp/nomad/client/allocrunner/interfaces"
@@ -41,10 +47,6 @@ import (
 	"github.com/hashicorp/nomad/plugins/device"
 	"github.com/hashicorp/nomad/plugins/drivers"
 	"github.com/hashicorp/nomad/testutil"
-	"github.com/kr/pretty"
-	"github.com/shoenig/test/must"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 )
 
 type MockTaskStateUpdater struct {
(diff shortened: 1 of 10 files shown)


client/allocrunner/taskrunner/task_runner_test.go
```diff
@@ -662,6 +664,61 @@ func TestTaskRunner_Restore_System(t *testing.T) {
...
+	// this gets set directly by MarkFailedKill()
+	test.True(t, state.Failed, test.Sprint("task should have failed"))
+	// this is set in Run()
+	test.Eq(t, structs.TaskStateDead, state.State, test.Sprint("task should be dead"))
+	// reason "because i said so" should be a task event message
+	foundMessages := make(map[string]bool)
+	for _, event := range state.Events {
+		foundMessages[event.DisplayMessage] = true
+	}
+	test.True(t, foundMessages[reason], test.Sprintf("expected '%s' in events: %#v", reason, foundMessages))
```
tgross, line 719: "Map contains value that meets this condition" seems like it'd be nice new frequently-used assertion for @shoenig's test library. But out of scope for this PR.

tgross (commented):
LGTM overall. We need a changelog entry via `make cl`, but also my comment about the commit description applies generally. If we squash-merge this as we typically do, we'll end up with a PR description that's just a reference somewhere else, instead of being self-contained within the commit message.

tgross (approved):
LGTM!

PR #17113, opened 2023-05-08
Backport of [17070] Correct the status description and modify time of canceled evals. into release/1.5.x

hc-github-team-nomad-core:
## Backport

This PR is auto-generated from #17071 to be assessed for backporting due to the inclusion of the label backport/1.5.x.

---

Fix for https://github.com/hashicorp/nomad/issues/17070. Corrected the status description and modify time of evals which are canceled due to another eval having completed in the meantime. 

Tests added. It seems that the testing has originally been written a rather long time before the new-style of testing started within the Nomad codebase. Please let me know if you'd rather I moved the entire test to the new style, or if added if clauses seem like a reasonable tradeoff given the size of the change.


tgross:
Fixed by hand, thanks BPA!

PR #17115, opened 2023-05-08
docs: add note to upgrade guide about yanked version

tgross:
Nomad 1.5.4 shipped with a logmon bug that we rolled out a fix for in Nomad 1.5.5. Unfortunately we can't yank the release but we should leave a note in the upgrade guide telling users to avoid it.

I'll backport this to `release/1.5.x` and the `stable-website` branch once merged.

### DIFF
--- a/website/content/docs/upgrade/upgrade-specific.mdx
+++ b/website/content/docs/upgrade/upgrade-specific.mdx
@@ -42,6 +42,21 @@ registered before the upgrade would no longer collect allocation logs. The
 by a `logs.disabled` field that defaults to false. The `logs.enabled` field value
 will be ignored in 1.5.5 and will be removed in Nomad 1.6.0.
 
+## Nomad 1.5.4
+
+Nomad 1.5.4 included a bug where allocations that are rescheduled for jobs
+registered before the upgrade would no longer collect allocation logs. The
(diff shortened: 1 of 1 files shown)


lgfa29 (approved):

website/content/docs/upgrade/upgrade-specific.mdx
```diff
@@ -42,6 +42,19 @@ registered before the upgrade would no longer collect allocation logs. The
...
+Nomad 1.5.4 included a bug where allocations that are rescheduled for jobs
+registered before the upgrade would no longer collect allocation logs. The
+client will emit debug-level logs like the following:
+
+```
+client.alloc_runner.task_runner.task_hook: log collection is disabled by task
+```
+
+You should avoid this version of Nomad and instead install the latest version of
+Nomad 1.5.
```
gulducat, line 56: Are there any steps one could take to remediate this for running allocs exhibiting this bug?  As in, "After upgrading to 1.5.5+, run `____` to reschedule affected allocations." ?

tgross: You don't need to do anything, as the client will spin up the missing logmon automatically for you during the task restore. I'll add that as a note.

PR #17116, opened 2023-05-08
Revert "ci: use `BACKPORT_MERGE_COMMIT` option (#16730)"

lgfa29:
This reverts commit 1721e687c0832bea3d9b7eec5dcd3c4e7a924d71.

The change was expected to solve the sporadic problems we were having with Backport Assistant, but it end up creating even more failures.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -19,6 +19,8 @@ jobs:
         env:
           BACKPORT_LABEL_REGEXP: "backport/(?P<target>website)"
           BACKPORT_TARGET_TEMPLATE: "stable-{{.target}}"
+          # Enabling this option increased the number of backport failures.
+          BACKPORT_MERGE_COMMIT: false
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yml
```diff
@@ -19,15 +19,13 @@ jobs:
         env:
           BACKPORT_LABEL_REGEXP: "backport/(?P<target>website)"
           BACKPORT_TARGET_TEMPLATE: "stable-{{.target}}"
-          BACKPORT_MERGE_COMMIT: true
```
shoenig, line 22: how about leaving these in but with `false` and a comment pointing to this issue

shoenig (approved):

tgross (approved):
LGTM

.github/workflows/backport.yml
lgfa29: Yeah, good idea. Added in 74ee31a8f

Issue #15861, opened 2023-01-24
nomad process dies with `panic: counter cannot decrease in value`


rbastiaans-tc:
### Nomad version
```
Nomad v1.3.1 (2b054e38e91af964d1235faa98c286ca3f527e56)
```
### Operating system and Environment details

```
Distributor ID:	Debian
Description:	Debian GNU/Linux 10 (buster)
Release:	10
Codename:	buster
```

### Issue

The Nomad agent process died with a message: `panic: counter cannot decrease in value`

See full panic output below.

### Reproduction steps

Unknown

#### Expected Result

Nomad stay running

#### Actual Result

Nomad process exits

### Nomad Client logs (if appropriate)

```
Sun 2023-01-22 16:57:52 UTC nomadclient013 nomad[3403]: ==> Newer Nomad version available: 1.4.3 (currently running: 1.3.1)
Sun 2023-01-22 16:58:41 UTC nomadclient013 nomad[3403]: panic: counter cannot decrease in value
Sun 2023-01-22 16:58:41 UTC nomadclient013 nomad[3403]: goroutine 7598137 [running]:
Sun 2023-01-22 16:58:41 UTC nomadclient013 nomad[3403]: github.com/prometheus/client_golang/prometheus.(*counter).Add(...)
Sun 2023-01-22 16:58:41 UTC nomadclient013 nomad[3403]:         github.com/prometheus/client_golang@v1.12.0/prometheus/counter.go:109
Sun 2023-01-22 16:58:41 UTC nomadclient013 nomad[3403]: github.com/prometheus/client_golang/prometheus.(*goCollector).Collect(0xc000280080, 0xc00235bf60)
Sun 2023-01-22 16:58:41 UTC nomadclient013 nomad[3403]:         github.com/prometheus/client_golang@v1.12.0/prometheus/go_collector_go117.go:147 +0x5ec
Sun 2023-01-22 16:58:41 UTC nomadclient013 nomad[3403]: github.com/prometheus/client_golang/prometheus.(*Registry).Gather.func1()
Sun 2023-01-22 16:58:41 UTC nomadclient013 nomad[3403]:         github.com/prometheus/client_golang@v1.12.0/prometheus/registry.go:446 +0x102
Sun 2023-01-22 16:58:41 UTC nomadclient013 nomad[3403]: created by github.com/prometheus/client_golang/prometheus.(*Registry).Gather
[...]

PR #17133, opened 2023-05-09
deps: update go-metrics to prevent panic

lgfa29:
#15861 describes intermittent panics caused by go-metrics Prometheus client. We have not been able to further debug this problem due to the lack of information when the panic happens.

https://github.com/hashicorp/go-metrics/pull/146 prevents the panic from happening and also logs additional information that can help us understand the root cause of the problem.

This commits pins the go-metric dependency to this branch until we can better debug the issue.


tgross (approved):
LGTM

tgross (approved):
LGTM

PR #17146, opened 2023-05-10
core: eliminate second index on job_submissions table

shoenig:
This PR refactors the `job_submissions` state store code to eliminate the
use of a second index formerly used for purging all versions of a given
job. In practice we ended up with duplicate entries on the table. Instead,
use index prefix scanning on the primary index and also tidy up any potential
for creating (or removing) duplicates.

Currently on main attempting to GC a job results in, 
```
    2023-05-10T20:15:42.657Z [ERROR] nomad.fsm: deregistering job failed: job=http error="DeleteJob failed: deleting job submission failed: not found"
    2023-05-10T20:15:42.657Z [ERROR] nomad.job: batch deregister failed: error="DeleteJob failed: deleting job submission failed: not found"
    2023-05-10T20:15:42.657Z [ERROR] core.sched: batch job reap failed: error="DeleteJob failed: deleting job submission failed: not found"
```

The "not found" is a funny symptom - the problem is that we have 2 references to the same job submission, we apply the delete to the first one, and then we get an error trying to delete the second.

### DIFF
--- a/nomad/state/schema.go
+++ b/nomad/state/schema.go
@@ -283,7 +283,6 @@ func jobVersionSchema() *memdb.TableSchema {
 
 // jobSubmissionSchema returns the memdb table schema of job submissions
 // which contain the original source material of each job, per version.
-// Unique index by Namespace, JobID, and Version.
 func jobSubmissionSchema() *memdb.TableSchema {
 	return &memdb.TableSchema{
 		Name: "job_submission",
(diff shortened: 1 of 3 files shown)


nomad/state/schema.go
```diff
@@ -283,46 +283,32 @@ func jobVersionSchema() *memdb.TableSchema {
...
-// Unique index by Namespace, JobID, and Version.
 func jobSubmissionSchema() *memdb.TableSchema {
 	return &memdb.TableSchema{
 		Name: "job_submission",
 		Indexes: map[string]*memdb.IndexSchema{
 			"id": {
 				Name:         "id",
 				AllowMissing: false,
-				Unique:       true,
+				Unique:       true, // basically worthless
```
schmichael, line 293:
```suggestion
				Unique:       true,
```

While arguably true, let's leave this off so we don't distract readers. AFAICT `Unique` always small optimizations in a handful of cases. It does not seem to enforce uniqueness, but I think your `uniqueness is DIY` comment below conveys that.

nomad/state/state_store.go
```diff
@@ -1907,8 +1907,36 @@ func (s *StateStore) deleteJobScalingPolicies(index uint64, job *structs.Job, tx
 }
 
 func (s *StateStore) deleteJobSubmission(job *structs.Job, txn *txn) error {
-	_, err := txn.DeleteAll("job_submission", "by_jobID", job.Namespace, job.ID)
-	return err
+	// find submissions associated with job
+	remove := *set.NewHashSet[*structs.JobSubmission, string](6)
```
schmichael, line 1911: Is there a const we can use for 6 (because I assume it's meant to match the max number of versions we intend to keep around)

schmichael: Actually can you use `txn.DeletePrefix("job_submission", "id_prefix", job.Namespace, job.ID)` here instead?

nomad/structs/structs.go
```diff
@@ -4307,6 +4307,29 @@ type JobSubmission struct {
 	JobModifyIndex uint64
 }
 
+// Hash returns a value representative of the intended uniquness of a
+// JobSubmission in the job_submission state store table (namespace, jobID, version).
+func (js *JobSubmission) Hash() string {
+	return fmt.Sprintf("%s \x00 %s \x00 %d", js.Namespace, js.JobID, js.Version)
+}
```
schmichael, line 4314: Where is this used?

schmichael (approved):

schmichael (approved):

nomad/state/state_store.go
schmichael: Oh no. :flushed: Is it ever possible to safely use `DeletePrefix` on indexes with user controlled fields?!

PR #17161, opened 2023-05-11
allocrunner: provide factory function so we can build mock ARs

tgross:
Tools like the [`nomad-nodesim`](https://github.com/schmichael/nomad-nodesim) are unable to implement a minimal implementation of an allocrunner so that we can test the client communication without having to lug around the entire allocrunner/taskrunner code base. The allocrunner was implemented with an interface specifically for this purpose, but there were circular imports that made it challenging to use in practice.

Move the `AllocRunner` interface into an inner package and provide a factory function type. Provide a minimal test that exercises the new function so that consumers have some idea of what the minimum implementation required is.


client/allocrunner/alloc_runner.go
```diff
@@ -209,7 +209,8 @@ type RPCer interface {
 }
 
 // NewAllocRunner returns a new allocation runner.
-func NewAllocRunner(config *Config) (*allocRunner, error) {
+func NewAllocRunner(cfg any) (interfaces.AllocRunner, error) {
+	config := cfg.(*Config)
```
schmichael, line 213: How awful would it be to move allocrunner.Config somewhere like client/structs? :grimacing:

client/allocrunner/testing.go
```diff
@@ -99,7 +100,7 @@ func testAllocRunnerConfig(t *testing.T, alloc *structs.Allocation) (*Config, fu
 	return conf, cleanup
 }
 
-func TestAllocRunnerFromAlloc(t *testing.T, alloc *structs.Allocation) (*allocRunner, func()) {
+func TestAllocRunnerFromAlloc(t *testing.T, alloc *structs.Allocation) (interfaces.AllocRunner, func()) {
```
schmichael, line 103: Do we need to update this test helper? It seems like returning the "real" implementation is the whole point.

client/client.go
```diff
@@ -3306,6 +3279,18 @@ func (c *Client) GetTaskEventHandler(allocID, taskName string) drivermanager.Eve
...
+	defer c.allocLock.Unlock()
+	c.allocrunnerFactory = factory
+}
+
+func (c *Client) newAllocRunner(conf *allocrunner.Config) (interfaces.AllocRunner, error) {
+	c.allocLock.RLock()
+	defer c.allocLock.RUnlock()
+	return c.allocrunnerFactory(conf)
+}
+
```
schmichael: Since the alloc runner is set by the client's configuration, why do we need to mutate it at runtime? Seems only useful for test cases where you know that know allocs will be run before you can swap factories.

If updating the factory at runtime is necessary, let's slap a big "For testing purposes only!" warning on it.

If updating the factory at runtime is _not_ necessary, let's kill the Set method and remove the locking from the newAR helper.

client/client_interface_test.go
```diff
@@ -0,0 +1,165 @@
...
+func TestEmptyAllocRunner(t *testing.T) {
+	ci.Parallel(t)
+
+	s1, _, cleanupS1 := testServer(t, nil)
+	defer cleanupS1()
+
+	c1, cleanup := TestClient(t, func(c *config.Config) {
+		c.RPCHandler = s1
+	})
+	c1.SetAllocRunnerFactory(newEmptyAllocRunnerFunc(c1))
```
schmichael, line 33: can't the AR factory be set in config.Config above instead?

schmichael (approved):
:tada: :tada: :tada: 

Step 2: group drivers & allocrunner plugins
Step 3: profit???

client/allocrunner/alloc_runner.go
tgross: Kinda rough. But maybe I can find a spot for it still.

tgross:
I've moved this to draft as with the new changes I've got a version that compiles but probably doesn't pass tests yet. Will pick it up in the morning.

tgross:
@schmichael @shoenig I've reworked based on @schmichael's feedback by moving the `allocrunner.Config` into `client/config.AllocRunnerConfig`, which we can use without creating an import cycle. This suggests long-term that pushing some more of the core types for the client into `client/structs` might be nice at some point in terms of untangling all of this, but this gives us an incremental improvement and a new interface for testing `client.Client` behavior.

tgross:
> Should we backport this to ease future backports that touch this code? It seems very safe

Will do.

shoenig (approved):
LGTM!

PR #17173, opened 2023-05-12
Backport of allocrunner: provide factory function so we can build mock ARs into release/1.4.x

hc-github-team-nomad-core:
## Backport

This PR is auto-generated from #17161 to be assessed for backporting due to the inclusion of the label backport/1.4.x.

---

Tools like the [`nomad-nodesim`](https://github.com/schmichael/nomad-nodesim) are unable to implement a minimal implementation of an allocrunner so that we can test the client communication without having to lug around the entire allocrunner/taskrunner code base. The allocrunner was implemented with an interface specifically for this purpose, but there were circular imports that made it challenging to use in practice.

Move the `AllocRunner` interface into an inner package and provide a factory function type. Provide a minimal test that exercises the new function so that consumers have some idea of what the minimum implementation required is.


tgross:
Had to remove some 1.6.0-specific bits of the `AllocRunner` interface.

PR #17190, opened 2023-05-15
[ui] Remove unnecessary subnav for parent jobs

### LINKED ISSUE #17189: [ui] Parent jobs (periodic, parameterized) should not have certain subnav items
Parent jobs do not have their own allocations, evaluations, and services, and so the subnav should not show them as they'll always be empty.

philrenaud:
Allocations, Evaluations, and Services don't exist for parents of child jobs, but we show their menu items anyway. This adds a check to see if the job has children, and if so, hides the relevant nav. "Clients" is also swept up on this, and is a redundant check, but that's fine.

Resolves #17189


ChaiWithJai (approved):
LGTM

PR #17192, opened 2023-05-15
[ui] Latest Deployment component removed

philrenaud:
The Latest Deployment panel is made obsolete by the new Job Status - Current panel, and so we're removing it and its backing logic.


ChaiWithJai (approved):
LGTM but looks like you have some test failures.

PR #17194, opened 2023-05-15
acl: refactor acl.UpsertTokens to avoid unnecessary RPC calls

pkazmierczak:
New RPC endpoints introduced during OIDC and JWT auth perform unnecessary many RPC calls when they upsert generated ACL tokens, as [pointed out](https://github.com/hashicorp/nomad/pull/15918/files/223c764e82408f02c306d9ebf98477167a4c6a05#r1138684122) by @tgross. 

This PR moves the common logic from `acl.UpsertTokens` method into a helper method that contains common logic, and sidesteps authentication, metrics, etc.


tgross (approved):
LGTM! :+1:

Issue #11823, opened 2022-01-11
implicit spread target results in incorrect scoring


ygersie:
### Nomad version
v1.2.3

### Issue
A fairly minor issue but according to the [spread documentation](https://www.nomadproject.io/docs/job-specification/spread#spread-with-target-percentages) providing a percent of 50 for a single DC should lead to half of the allocs being scheduled in that DC. This isn't the case until you specify all of the targets. With a 3 DC dev setup and the following job:
```hcl
job "example" {
  datacenters = ["dc1", "dc2", "dc3"]

  update {
    max_parallel     = 30
    min_healthy_time = "5s"
  }

  spread {
    attribute = node.datacenter

    target "dc1" {
      percent = 50
    }
  }

  group "example" {
    count = 12
    task "example" {
      driver = "docker"
      config {
        image = "alpine"
        args  = [ "tail", "-f", "/dev/null" ]
      }

      resources {
        cpu    = 100
        memory = 64
      }
    }
  }
}
```
The result would be 4 allocations per node, an even spread. According to the docs this should have lead to 6 allocs being placed in dc1 and the remainder in the others:
```sh
$ nomad status example | grep -A100 Allocations | tail -n+3 | sort -k2
736eb996  4c862aa8  example     0        run      running  10s ago  4s ago
c6d637f3  4c862aa8  example     0        run      running  10s ago  4s ago
f5b55722  4c862aa8  example     0        run      running  10s ago  4s ago
47ed2c33  4c862aa8  example     0        run      running  10s ago  5s ago
239c67d4  732bbfb8  example     0        run      running  10s ago  5s ago
b0f6f499  732bbfb8  example     0        run      running  10s ago  5s ago
c915fa61  732bbfb8  example     0        run      running  10s ago  5s ago
cb0f60db  732bbfb8  example     0        run      running  10s ago  5s ago
038f6167  8fb0da83  example     0        run      running  10s ago  1s ago
0c6c2d0a  8fb0da83  example     0        run      running  10s ago  1s ago
87ecb255  8fb0da83  example     0        run      running  10s ago  1s ago
9ad5b804  8fb0da83  example     0        run      running  10s ago  1s ago
```
When I specify the `target` for each dc as follows:
```hcl
job "example" {
  datacenters = ["dc1", "dc2", "dc3"]

  update {
    max_parallel     = 30
    min_healthy_time = "5s"
  }

  spread {
    attribute = node.datacenter

    target "dc1" {
      percent = 50
    }
[...]

PR #17195, opened 2023-05-15
scheduler: count implicit spread targets as a single target

tgross:
When calculating the score in the `SpreadIterator`, the score boost is proportional to the difference between the current and desired count. But when there are implicit spread targets, the current count is the sum of the possible implicit targets, which results in incorrect scoring unless there's only one implicit target.

This changeset updates the `propertySet` struct to accept a set of explicit target values so it can detect when a property value falls into the implicit set and should be combined with other implicit values.

Fixes: #11823

### DIFF
--- a/scheduler/propertyset.go
+++ b/scheduler/propertyset.go
@@ -9,6 +9,8 @@ import (
 
 	log "github.com/hashicorp/go-hclog"
 	memdb "github.com/hashicorp/go-memdb"
+	"github.com/hashicorp/go-set"
+
 	"github.com/hashicorp/nomad/helper"
 	"github.com/hashicorp/nomad/nomad/structs"
 )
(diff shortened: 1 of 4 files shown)


scheduler/propertyset.go
```diff
@@ -347,3 +362,14 @@ func getProperty(n *structs.Node, property string) (string, bool) {
 
 	return resolveTarget(property, n)
 }
+
+// targetedPropertyValue transforms the property value to combine all implicit
+// target values into a single wildcard placeholder so that we get accurate
+// counts when we compare an explicitly-defined target against multiple implicit
+// targets.
+func (p *propertySet) targetedPropertyValue(propertyValue string) string {
+	if len(p.targetValues) == 0 || slices.Contains(p.targetValues, propertyValue) {
```
shoenig, line 371: if this is the key operation you could probably use `set.Set` and make this `O(1)`, though it's also just fine as is

shoenig (approved):
LGTM!

scheduler/propertyset.go
tgross: The number of values here will typically be really small, but it can't hurt. Done.

Issue #15686, opened 2023-01-04
Drivers: Make InternalCapabilities.DisableLogCollection Public


endocrimes:
### Proposal

The current set of Internal Driver Capabilities have been relatively stable since the introduction of the plugin interface, and some would greatly benefit 3rd party users of Nomad and Custom Task Drivers. This proposal suggests moving the DisableLogCollection feature to the public capabilities API.

While not perfect in isolation (Nomad offers no way to disable log collection for individual tasks, which means when scheduling to a driver with log collection disabled, the task still has an associated set of resources allocated for log storage), this goes a long way towards reducing overhead in a range of cases.

### Use-cases

There are many use-cases where a driver author might want to give operators the ability to save on the overhead of deploying an unnecessary `logmon` for every task. Similarly to the Nomad Docker Driver's usage, the most obvious is where a runtime offers the ability to manage logs externally to Nomad.

This is particularly important when building high-density runtimes (and task drivers) - where no-op logmon overhead can quickly become an issue when deploying X000 allocations/node and logging is handled externally.

### Attempted Solutions

We currently deploy a fork of Nomad that has a _slightly_ jankier approach 😅.

tgross:
Hi @endocrimes! Exposing this capability makes sense to me. 

A little while back I spent a few days hacking on some thoughts for something like this (and https://github.com/hashicorp/nomad/issues/14636 https://github.com/hashicorp/nomad/issues/14056 https://github.com/hashicorp/nomad/issues/9211 and a variety of [logging-related issues](https://github.com/hashicorp/nomad/issues?q=is%3Aopen+is%3Aissue+label%3Atheme%2Flogging)). I've got a disorganized branch with a couple different experiements at https://github.com/hashicorp/nomad/compare/main...hack-logging-plugins which I'm going to carve out some time on in the next major release cycle. The larger "logging plugins" concept may not land but we definitely want to support the simpler lower-resource use case you're talking about here.

endocrimes:
@tgross _Awesome_ - always happy to chat if you wanna bounce ideas around - from that branch I love the new logmon prospect (and the logging plugin concept in general).

It'd go a long way to alleviating one of the harder things to do cleanly with non-Docker drivers 😅 too, which is pretty good for everyone.

tgross:
Done in #17196 which will ship in Nomad 1.6.0! I'll have a PR up (probably later today) adding the capability to the other internal drivers.

Issue #14636, opened 2022-09-21
driver: allow operators to disable logmon process spawning


jrasell:
PR https://github.com/hashicorp/nomad/pull/6820 added an experimental driver interface which allowed operators to disable logmon process spawning. This interface is only available to builtin drivers as it was not added to the driver capabilities functionality to allow for UX improvements and modifications to the interface.

In some cases, cluster operators would like to disable spawning logmon processes to avoid this additional overhead where the functionality it delivers is not required. It would be good therefore to discuss what changes we should make to the interface, if any, and what needs to be done in order to have this functionality, or equivalent exposed to external drivers.

PR #16962, opened 2023-04-21
logs: allow disabling log collection in jobspec

tgross:
Some Nomad users ship application logs out-of-band via syslog. For these users having `logmon` (and `docker_logger`) running is unnecessary overhead. Allow disabling the logmon and pointing the task's stdout/stderr to /dev/null.

This changeset is the first of several incremental improvements to log collection short of full-on logging plugins. The next step will likely be to extend the internal-only task driver configuration so that cluster administrators can turn off log collection for the entire driver.

---

Fixes: #11175

PR #17196, opened 2023-05-15
drivers: make internal `DisableLogCollection` capability public

tgross:
The `DisableLogCollection` capability was introduced as an experimental interface for the Docker driver in 0.10.4. The interface has been stable for a while, and allowing third-party task drivers the same capability would be useful for those drivers that don't need the additional overhead of logmon.

This PR only makes the capability public. It doesn't yet add it to the configuration options for the other internal drivers. (I'll do that in a separate PR just to keep review size reasonable.)

Fixes: #14636 
Fixes: #15686

Unlike #16962 this has no visible impact on the jobspec or server state, but just to be sure I've tested upgrade paths on clients between 1.5.5 and this branch, both with and without the flag enabled on the `docker` driver.


jrasell (approved):
LGTM!

pkazmierczak (approved):

PR #4936, opened 2018-11-29
Refactor and repackage client/driver

dantoml:
This PR migrates future relevant code out of the client/driver/* package, and then removes the legacy driver interfaces.

A side effect of this is that we no longer get submission time job validation, but that should be coming as a follow up to the 0.9 release with included support for driver plugin validation.

❓
- Can we rm the allocrunnerdeprecated (and executor_pluginpre_09_0) trees now?

PR #17197, opened 2023-05-15
Remove unused helper/fields package

schmichael:
Has been unused since we switched task drivers to plugins and used hclschema for config in #4936 (v0.9.0-beta1)


shoenig (approved):
:broom:

Issue #8863, opened 2020-09-10
`-Inf` in allocation scoring JSON output


mrkurt:
This looks identical to #5698 to me, but we have allocations/nodes that suddenly cause errors with both `nomad` and the web UI.

### Nomad version
`Nomad v0.11.0 (5f8fe0afc894d254e4d3baaeaee1c7a70a44fdc6)`

### Operating system and Environment details
Ubuntu 18.04, physical Intel Xeon based server.

### Issue

JSON returned from nomad server includes `-Inf`:

```json
{"NodeID":"233150e3-6991-9a0c-101b-64124eca941a","Scores":{"binpack":0.15161654193467708,"job-anti-affinity":0.0,"node-reschedule-penalty":0.0,"node-affinity":0.3333333333333333,"allocation-spread":-Inf},"NormScore":-Inf}
```

tgross:
Hi @mrkurt! Sorry to hear about that. Can you clarify which command or HTTP API you're hitting to get that JSON back?

mrkurt:
@tgross it seems like it comes from the `allocations` lookup for a node. `nomad node status -self` would do it, and the client wouldn't load in the UI at all (console errors parsing what I'm pretty sure was the allocations response).

tgross:
Ok, thanks @mrkurt. It looks like the last logic change to that area of the code was in https://github.com/hashicorp/nomad/commit/ea843a507a54b7eb3d8de8babf0f950e90624a40#diff-7c9a6a9bdf41c1724b4f427d5c39daec which shipped in 0.9.4, so it looks like there's an unfortunate edge case here. Can you share the jobspec for the allocation involved? That might help us narrow the problem down.

mrkurt:
These are Spreads/Constraints/Affinities from the job (I can email you the whole thing if it's helpful):

```json
  "Spreads": [
    {
      "Weight": 100,
      "Attribute": "${meta.fly_region}",
      "SpreadTarget": [
        {
          "Value": "vin",
          "Percent": 100
        }
      ]
    }
  ],
  "Affinities": [
    {
      "Weight": 50,
      "LTarget": "${meta.fly_region}",
      "Operand": "set_contains_any",
      "RTarget": "vin"
    },
    {
      "Weight": -100,
      "LTarget": "${meta.fly_region}",
      "Operand": "set_contains_any",
      "RTarget": "scl1,syd"
    }
  ],
  "Constraints": [
    {
      "LTarget": "${meta.fly_region}",
[...]

PR #17198, opened 2023-05-15
scheduler: prevent -Inf in spread scoring 

tgross:
When spread targets have a percent value of zero it's possible for them to return -Inf scoring because of a float divide by zero. This is very hard for operators to debug because the string "-Inf" is returned in the API and that breaks the presentation of debugging data.

Most scoring iterators are bracketed to -1/+1, but spread iterators do not so that they can handle greatly unbalanced scoring so we can't simply return a -1 score without generating a score that might be greater than the negative scores set by other spread targets. Instead, track the lowest-seen spread boost and use that as the spread boost for any cases where we'd divide by zero.

Fixes: #8863

### DIFF
--- a/.changelog/17198.txt
+++ b/.changelog/17198.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+scheduler: Fixed a bug where scores for spread scheduling could be -Inf
+```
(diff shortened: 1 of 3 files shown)


tgross:
For what it's worth, it's still a little unclear to me whether we _should_ be allowing `spread` to dominate the scoring so much by having unclamped values. But I'm preserving the existing behavior here.

shoenig (approved):
LGTM!

PR #17207, opened 2023-05-16
Backport of Prevent kill_timeout greater than progress_deadline  into release/1.5.x

hc-github-team-nomad-core:
## Backport

This PR is auto-generated from #16761 to be assessed for backporting due to the inclusion of the label backport/1.5.x.

---

This PR addresses the bug reported on https://github.com/hashicorp/nomad/issues/8487

If a kill_timeout is greater than the job's progress_deadline, allocations may keep running (after initial kill signal) for long enough that the deploy fails. However, the allocations that were scheduled by that job will still be pending after job failure, and will be placed and started once the previous allocations do exit. This PR adds a validation to avoid this situation.


pavanrangain:
Wonder why this got back ported on a patch release. We were waiting for fix of https://github.com/hashicorp/nomad/issues/17079 but this backport is preventing us from moving to 1..5.6 as many jobs may now need a change. 
Isn't this considered a breaking change to have been back ported to patch release ?

PR #17212, opened 2023-05-16
services: un-mark group services as deregistered if restart hook runs…

shoenig:
… (#16905)

* services: un-mark group services as deregistered if restart hook runs

This PR may fix a bug where group services will never be deregistered if the group undergoes a task restart.

* e2e: add test case for restart and deregister group service

* cl: add cl

* e2e: add wait for service list call


tgross (approved):
LGTM!

PR #17213, opened 2023-05-16
ui:  explain why `alloc` is failing

### LINKED ISSUE #16942: [ui] Give explanation for hasStoppedRescheduling
It would be nice to know why an allocation has stopped rescheduling:
https://files.slack.com/files-pri/T04J1HLKT-F0543UZA8DT/image.png

Relevant computed prop: https://github.com/hashicorp/nomad/tree/main/ui/app/models/allocation.js#L166-L177

Ideally we could provide a `hasStoppedReschedulingReason` string property to users that gives better specifics.

ChaiWithJai:
Resolves #16942 

This PR creates a new derived state property on the `Allocation` model to show why an allocation has stopped rescheduling and updates the template to provide a better reason about why the allocation has stopped rescheduling and links to the follow up evaluation to enable the user to debug what's happening.

### DIFF
--- a/ui/app/models/allocation.js
+++ b/ui/app/models/allocation.js
@@ -128,6 +128,10 @@ export default class Allocation extends Model {
 
   @belongsTo('evaluation') followUpEvaluation;
 
+  get followUpEvaluationId() {
+    return this.belongsTo('evaluation').id();
+  }
+
   @computed('clientStatus')
   get statusClass() {
     const classMap = {
@@ -199,6 +203,27 @@ export default class Allocation extends Model {
       !this.get('followUpEvaluation.content') &&
       this.clientStatus === 'failed'
     );
+  }
+
+  @computed(
+    'hasStoppedRescheduling',
+    'node.status',
+    'resources.length',
+    'task.{config.errors.length,code.errors.length}'
+  )
+  get failureReason() {
+    switch (this.hasStoppedRescheduling) {
+      case this.node.status === 'failed':
+        return 'the node that the allocation was scheduled on failed.';
+      case this.resources.length === 0:
+        return 'the resources that the allocation was scheduled on were not available.';
+      case this.task.config.errors.length > 0:
+        return "the task or service's configuration was incorrect.";
+      case this.task.code.errors.length > 0:
+        return "there was an error in the task or service's code.";
+      default:
+        return 'of an unknown failure reason.';
+    }
   }
 
   stop() {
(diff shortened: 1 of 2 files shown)


ui/app/models/allocation.js
```diff
@@ -201,6 +205,27 @@ export default class Allocation extends Model {
...
+  get failureReason() {
+    switch (this.hasStoppedRescheduling) {
+      case this.node.status === 'failed':
+        return 'the node that the allocation was scheduled on failed.';
+      case this.resources.length === 0:
+        return 'the resources that the allocation was scheduled on were not available.';
+      case this.task.config.errors.length > 0:
+        return "the task or service's configuration was incorrect.";
+      case this.task.code.errors.length > 0:
+        return "there was an error in the task or service's code.";
```
philrenaud, line 223: Will this prevent reschedule, or simply throw an error but otherwise continue on? Asking because I want to make sure that we're giving the right cause when we say "stopped .... because of ________"

ChaiWithJai: This is a derived state computation to show a reason why an `allocation` has stopped rescheduling to be shown in the `allocations.allocation.index` view. This is only shown when the current allocation (the one that's associated with the view) has stopped rescheduling.

ChaiWithJai: See the invocation in the template layer for usage details and you'll see that this only is applied for the top level allocation on the reschedule timeline because the rest of the items in the timeline are `RescheduleEventRow` which only provides links and no text.

tgross:
@ChaiWithJai this doesn't seem like it actually solves the problem described in https://github.com/hashicorp/nomad/issues/16942. The description you have here (and the code) tells us why the allocation failed, not why the scheduler stopped scheduling it.

ui/app/models/allocation.js
```diff
@@ -201,6 +205,27 @@ export default class Allocation extends Model {
...
   }
 
+  @computed(
+    'hasStoppedRescheduling',
+    'node.status',
+    'resources.length',
+    'task.{config.errors.length,code.errors.length}'
+  )
+  get failureReason() {
+    switch (this.hasStoppedRescheduling) {
```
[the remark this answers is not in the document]
ChaiWithJai: Mmmmm.... good call.

ChaiWithJai:
> @ChaiWithJai this doesn't seem like it actually solves the problem described in #16942. The description you have here (and the code) tells us why the allocation failed, not why the scheduler stopped scheduling it.

Yeah... looks I conflated the concepts. There doesn't appear to be a solution on this problem. There's no follow up evaluation to link for the evaluations view and we can't predictably compute derived state to give the user a better explanation.

PR #17226, opened 2023-05-17
ui:  add option to filter for jobs that are packs

### LINKED ISSUE #15736: Add type filter option for packs
### Proposal
I would love to be able to filter on the which jobs are packs

Screenshot from 2023-01-09 23-38-35

### Use-cases
to visually limited what jobs are packs and which arent

### Attempted Solutions
n/a

ChaiWithJai:
Resolves #15736 

This PR adds a UI filter to the `jobs.index` view to enable filtering for jobs that are packs.

### DIFF
--- a/ui/app/controllers/jobs/index.js
+++ b/ui/app/controllers/jobs/index.js
@@ -91,6 +91,7 @@ export default class IndexController extends Controller.extend(
   get optionsType() {
     return [
       { key: 'batch', label: 'Batch' },
+      { key: 'pack', label: 'Pack' },
       { key: 'parameterized', label: 'Parameterized' },
       { key: 'periodic', label: 'Periodic' },
       { key: 'service', label: 'Service' },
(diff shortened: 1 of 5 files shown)


ui/app/controllers/jobs/index.js
```diff
@@ -224,7 +225,13 @@ export default class IndexController extends Controller.extend(
...
     // Always return early to prevent unnecessary facet predicates.
     return this.visibleJobs.filter((job) => {
-      if (types.length && !types.includes(job.get('displayType'))) {
+      const shouldShowPack = types.includes('pack') && job.displayType.isPack;
+
+      if (types.length && shouldShowPack) {
+        return true;
+      }
+
+      if (types.length && !types.includes(job.get('displayType.type'))) {
```
philrenaud, line 234: Something seems different about pack here; shown is a non-pack job where I have service + pack selected. Shouldn't this be hidden?


image

philrenaud (changes requested):
Pack doesn't seem to have the same exclusivity rules in filtering as to the job types

ui/app/controllers/jobs/index.js
philrenaud: I guess my confusion is from Pack not really being a job type as much as an attribute of the job / making it the only non-mutually-exclusive item here. I don't think this will lead to much user error, so I think leave it in, but another way we might have tackled this is under a separate dropdown for job attributes.

philrenaud (approved):

ui/app/controllers/jobs/index.js
ChaiWithJai: Yeah, it really does deserve its own filter. But then we just clutter the page. Our filtering is pretty far off from the rest of the other UIs. So it might be worth addressing later on down the line.

Issue #16128, opened 2023-02-10
☂️ Job Page Redesign


philrenaud:
[Figma](https://www.figma.com/file/u0bjC7eDS2eikXMDJoZhYQ/job-show-page-redesign?node-id=5085%3A127012&t=UnExc1OE8N4qMnOJ-0)
[PRD](https://docs.google.com/document/d/1hJyeW585VQGST5BTtIocu6ml8zbceOYpt9-Eq0isv24/edit?usp=sharing)

## Service Jobs
- [x] #16133
  - [x] #16340
  - [x] #17240
- [x] #16350
  - [x] #16512
  - [x] #16961
  - [x] #17191

## System Jobs
- [x] #17036
- [x] #17073
- [x] #17099

## Batch Jobs
- [x] #17139

## Other Job Types
- [x] #17114
- [x] #17242
- [x] #17216

## Status panel, generally:
- [x] #17149
- [x] #16941
- [x] #16821
- [x] #16881
- [x] #16724
- [x] #16802
- [x] #16797
- [ ] Consider refactoring allocation-status-block (and represented-alloc?) into classes per https://github.com/hashicorp/nomad/pull/16766#discussion_r1158921786
- [x] #17159
- [x] #17011
- [ ] #17058
- [x] #17513
- [x] #17523
- [x] #17525
- [x] #17567
- [x] #17627
- [x] #17565
- [x] #17597
- [x] #17643
- [ ] #17683

## Misc:
- [ ] Mirage Deployment simulator
- [ ] Remove obsoleted components (job-client-status-summary)
- [x] #17193
- [x] #17221

## Jobs Index page
- [x] #17283

mikenomitch:
Note: An allocation that is completed and the rescheduled should be marked as rescheduled

PR #17227, opened 2023-05-17
[ui] Adds a "Scheduling" filter to the job.allocations page

### LINKED ISSUE #17221: Add a "Will not Reschedule" filter to the job/allocations page

philrenaud:
Adds a new filter, "Scheduling", to the job allocations page. This helps facilitate Rescheduled/Restarted links from the upcoming [Job Status Panel](https://github.com/hashicorp/nomad/issues/16128) and gives us future options for letting users drill down on this list page.

Because of some seen inconsistencies in testing, this adds `NextAllocation` to Allocations in their list/stub form from our API, where before they only existed when an allocation was being fetched individually. This, combined with a check for FollowUpEvalID, gives us a better picture of "Is a terminal allocation never going to be rescheduled?", which the UI is seeking to show to users here.

image

image

Resolves #17221

### DIFF
--- a/api/allocations.go
+++ b/api/allocations.go
@@ -326,6 +326,7 @@ func (a *Allocation) Stub() *AllocationListStub {
 		TaskStates:            a.TaskStates,
 		DeploymentStatus:      a.DeploymentStatus,
 		FollowupEvalID:        a.FollowupEvalID,
+		NextAllocation:        a.JobID,
 		RescheduleTracker:     a.RescheduleTracker,
 		PreemptedAllocations:  a.PreemptedAllocations,
 		PreemptedByAllocation: a.PreemptedByAllocation,
@@ -379,6 +380,7 @@ type AllocationListStub struct {
 	TaskStates            map[string]*TaskState
 	DeploymentStatus      *AllocDeploymentStatus
 	FollowupEvalID        string
+	NextAllocation        string
 	RescheduleTracker     *RescheduleTracker
 	PreemptedAllocations  []string
 	PreemptedByAllocation string
--- a/ui/app/components/job-status/failed-or-lost.hbs
+++ b/ui/app/components/job-status/failed-or-lost.hbs
@@ -9,14 +9,28 @@
       <FlightIcon @name="info" />
     </span>
   </h4>
-  <ConditionalLinkTo
-    @condition={{this.shouldLinkToAllocations}}
-    @route="jobs.job.allocations"
-    @model={{@job}}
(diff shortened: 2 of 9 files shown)


ui/app/components/job-status/panel/steady.hbs
```diff
@@ -48,7 +48,7 @@
               @condition={{not (eq type.label "unplaced")}}
               @route="jobs.job.allocations"
               @model={{@job}}
-              @query={{hash status=(concat '["' type.label '"]') version=(concat '[' (keys this.versions) ']')}}
+              @query={{hash status=(concat '["' type.label '"]') version=(concat '[' (map-by "version" this.versions) ']')}}
```
philrenaud, line 51: Noted that I was being inconsistent after changing my version sorting logic a few PRs ago.

ChaiWithJai (approved):

api/allocations.go
```diff
@@ -326,6 +326,7 @@ func (a *Allocation) Stub() *AllocationListStub {
 		TaskStates:            a.TaskStates,
 		DeploymentStatus:      a.DeploymentStatus,
 		FollowupEvalID:        a.FollowupEvalID,
+		NextAllocation:        a.JobID,
```
lgfa29, line 329:
```suggestion
		NextAllocation:        a.NextAllocation,
```

ui/app/components/job-status/failed-or-lost.hbs
```diff
@@ -9,14 +9,28 @@
...
+    >
+      {{@allocs.length}}
+    </ConditionalLinkTo>
+  {{/if}}
+  {{#if (eq @title "Restarted")}}
+    <ConditionalLinkTo
+      @condition={{this.shouldLinkToAllocations}}
+      @route="jobs.job.allocations"
+      @model={{@job}}
+      @query={{hash scheduling=(concat '["has-been-restarted"]') version=(concat '[' @job.latestDeployment.versionNumber ']')}}
```
lgfa29, line 29: What is being concatened in `(concat '["has-been-rescheduled"]')` and `(concat '["has-been-restarted"]')`? It seems like it's just a string?

lgfa29 (changes requested):
Just need to fix `api/allocations.go`, otherwise it will return the wrong value.

A `.changelog` entry would be good as well since it's a change to the API return content.

ui/app/components/job-status/failed-or-lost.hbs
philrenaud: Good catch; this was an artefact from earlier! Will fix.

PR #17238, opened 2023-05-18
build: move GitHub actions to versions allowed by prodsec

tgross:
The `backspace/ember-asset-size` action we're using is unmaintained and has a bunch of vulns in it, so it won't pass security screening (this is a NodeJS action so it has piles of dependencies, 99% of which won't be in use but fails automated screening anyways). Move this to the upstream version.

The `machine-learning-apps/pr-comment` action also presents a problem for the ProdSec security screening because it's archived and also runs an external Docker image. Move this to a likely-ok maintained action for now, until we can spare some time to remove this in lieu of something more reasonable that isn't a GitHub action.

Once this is tested, I'll then have to follow up in https://github.com/hashicorp/security-tsccr/pull/428 to pin the SHA, and then I'll have to follow-up once that other PR is merged to pin the SHAs here.

### DIFF
--- a/.github/workflows/ember-assets.yml
+++ b/.github/workflows/ember-assets.yml
@@ -18,7 +18,7 @@ jobs:
         uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3.6.0
         with:
           node-version: '14'
-      - uses: backspace/ember-asset-size-action@edit-comment # TSCCR: no entry for repository "backspace/ember-asset-size-action"
+      - uses: mainmatter/ember-asset-size-action@v3
         with:
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
           working-directory: "ui"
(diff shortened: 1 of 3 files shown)


philrenaud (approved):
Thanks, Tim! Any idea why we were using forked / lesser-used versions before?

tgross:
> At a glance yeah I think using @DingoEatingFuzz 's ember-test-audit is the right move (Michael, please let us know if you think otherwise!)

Sorry, I mixed up `ember-test-audit` (which is the program being run) with https://github.com/backspace/ember-test-audit-comparison-action (which is the thing with the `action.yml` that lets us run in GHA). That is _not_ a fork. I'm an advocate of vendoring things like this into the HashiCorp organization, but I don't need to do that to ship this.

I'm going to merge this (and backport) and then update https://github.com/hashicorp/security-tsccr/pull/428 with the resulting changes.

Issue #16277, opened 2023-02-28
Document default ACL policy for Workload


aofei:
To work around #16276, I dug into the source code to find the default ACL policy for Workload, which took me a while. I was actually curious about the default ACL policy before I encountered #16276. But I'm too lazy to search the code so gave up on figuring it out. So I thought it might be a good idea for people to have a "Workload Default ACL Policy" section in [workload-identity.mdx](https://github.com/hashicorp/nomad/blob/542b23e999b5338cdcfa56bfdae1db9aa834cb3c/website/content/docs/concepts/workload-identity.mdx).

tgross:
Hi @aofei! This is definitely a documentation gap. The default policy isn't a "real" policy but some special exceptions for Variables and Service Registration (those exceptions are the source of https://github.com/hashicorp/nomad/issues/16276 which I'm working on with you). The Variables policy is documented at [Task Access to Variables](https://developer.hashicorp.com/nomad/docs/concepts/variables#task-access-to-variables) but that's neither complete nor the right place to have it. Marking this as a documentation gap to fix.

tgross:
Opened #17245 to get that documentation up.

PR #17245, opened 2023-05-18
docs: describe the default Workload Identity ACL policy

tgross:
Workload Identities have an implicit default policy. This policy can't currently be described via HCL because it includes task interpolation for Variables and access to the Services API (which doesn't exist as its own ACL capbility). Describe this in our WI documentation.

Fixes: #16277

### DIFF
--- a/website/content/docs/concepts/workload-identity.mdx
+++ b/website/content/docs/concepts/workload-identity.mdx
@@ -43,6 +43,14 @@ task "example" {
 }
 ```
 
+## Default Workload ACL Policy
+
+By default, a Workload Identity has access to a implicit ACL policy. This policy
+grants access to Nomad Variables associated with the job, group, and task, as
+described in [Task Access to Variables][]. The implicit policy also allows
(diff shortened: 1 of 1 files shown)


website/content/docs/concepts/workload-identity.mdx
```diff
@@ -43,6 +43,14 @@ task "example" {
 }
 ```
 
+## Default Workload ACL Policy
+
+By default, a Workload Identity has access to a implicit ACL policy. This policy
+grants access to Nomad Variables associated with the job, group, and task, as
+described in [Task Access to Variables][]. The implicit policy also allows
+access to list or read any Nomad service registration as with the [List Services
+API][] or [Read Service API][].
```
jrasell, line 52:
```suggestion
described in [Task Access to Variables][]. The implicit policy also allows
access to list or read any Nomad service registration in the same namespace
as the job, as with the [List Services API][] or [Read Service API][].
```

jrasell (approved):
LGTM, although which a clarifying suggestion around the service registration detail.

PR #17276, opened 2023-05-22
post 1.5.6 release

shoenig:
- Generate files for 1.5.6 release
- Prepare for next release
- Merge release 1.5.6 files


tgross (changes requested):
@philrenaud and @angrycub recently did a one-off commit of a new version of the `bindata_assetfs.go` from a build from main b/c the old one was super-broken with the current `main` API (ref https://github.com/hashicorp/nomad/pull/17263). Let's remove this file from this PR so that we don't overwrite that.

tgross (approved):
LGTM!

PR #17280, opened 2023-05-23
[api] Add NetworkStatus to allocation response

deverton-godaddy:
Service discovery or mesh network systems consuming the Nomad event stream or API need to know the CNI assigned IP for the allocation. This data is returned by the underlying Nomad API but isn't mapped in the response struct in the Go client.

### DIFF
--- a/.changelog/17280.txt
+++ b/.changelog/17280.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+api: add missing field NetworkStatus for Allocation
+```
--- a/api/allocations.go
+++ b/api/allocations.go
@@ -270,6 +270,7 @@ type Allocation struct {
 	PreviousAllocation    string
 	NextAllocation        string
 	RescheduleTracker     *RescheduleTracker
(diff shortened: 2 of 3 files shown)


lgfa29 (approved):
Thank you for the contribution @deverton-godaddy!

I pushed a commit to add a CHANGELOG entry and a quick test. I also reordered the struct fields so it matches the order in `structs.go`, making it easier to finding missing fields (and I did notice some others missing here 😅)

Before we can merge this we will need you to sign the CLA. Would you be able to do that?

Thanks!

deverton-godaddy:
Thank you for taking the time to update this PR. I'm still working on getting CLA approval and will update as soon as I can. I'm hoping it should be in the next few days.

deverton-godaddy:
@lgfa29 I've signed the CLA so this should be good to go.

lgfa29:
> @lgfa29 I've signed the CLA so this should be good to go.

Ops, sorry I missed this @deverton-godaddy. Thanks for handling the CLA!

PR #17296, opened 2023-05-23
docs: fixup example of readiness check

shoenig:
A "readiness" check implies a failing healthcheck will not cause the
deployment of a service to stop - i.e. it is only used as a liveness
probe in the context of service discoverability.

Fix our docs example to reflect that a readiness check is created by
setting on_update to "ignore" (as opposed to "ignore_warnings").


tgross (approved):
LGTM

PR #17301, opened 2023-05-24
ui:  create node pool model

### LINKED ISSUE #17292: Create a Node Pool Model in Ember Data

ChaiWithJai:
Resolves #17292 

This PR adds a new NodePool model and serializer. The NodePool model is used to represent a pool of nodes that can be used to run applications. The serializer is used to serialize the JSON response of a NodePool object to a JSON:API object.

The NodePool model has the following properties:

```
name: The name of the node pool
description: A description of the node pool
meta: A map of additional metadata about the node pool
schedulerConfiguration: The configuration of the scheduler for the node pool
```

The serializer serializes NodePool objects to JSON:API by converting the properties of the objects to strings. The following is an example of the serialized object that is generated for a NodePool object:

```
{
  "name": "prod-eng",
  "description": "Production workloads",
  "meta": {
    "env": "production",
    "team": "engineering"
  },
  "schedulerConfiguration": {
    "SchedulerAlgorithm": "spread" // pattern match:  We don't apply camelCasing for nested objects in other models
  }
}
```

### DIFF
--- a/ui/app/models/node-pool.js
+++ b/ui/app/models/node-pool.js
@@ -1,0 +1,22 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Model from '@ember-data/model';
+import { attr } from '@ember-data/model';
+import { computed } from '@ember/object';
+import classic from 'ember-classic-decorator';
+
+@classic
+export default class NodePool extends Model {
+  @attr('string') name;
+  @attr('string') description;
+  @attr() meta;
+  @attr() schedulerConfiguration;
+
+  @computed('schedulerConfiguration.SchedulerAlgorithm')
+  get schedulerAlgorithm() {
+    return this.get('schedulerConfiguration.SchedulerAlgorithm');
+  }
+}
--- a/ui/app/serializers/node-pool.js
+++ b/ui/app/serializers/node-pool.js
@@ -1,0 +1,12 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import ApplicationSerializer from './application';
+import classic from 'ember-classic-decorator';
+
+@classic
+export default class NodePool extends ApplicationSerializer {
+  primaryKey = 'Name';
+}
--- a/ui/tests/unit/serializers/node-pool-test.js
+++ b/ui/tests/unit/serializers/node-pool-test.js
@@ -1,0 +1,46 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { setupTest } from 'ember-qunit';
+import { module, test } from 'qunit';
+
+module('Unit | Serializer | NodePool', function (hooks) {
+  setupTest(hooks);
+
+  hooks.beforeEach(function () {
+    this.store = this.owner.lookup('service:store');
+    this.subject = () => this.store.serializerFor('node-pool');
+  });
+
+  test('should serialize a NodePool', function (assert) {
+    const nodePool = this.store.createRecord('node-pool', {
+      name: 'prod-eng',
+      description: 'Production workloads',
+      meta: {
+        env: 'production',
+        team: 'engineering',
+      },
+      schedulerConfiguration: {
+        SchedulerAlgorithm: 'spread',
+      },
+    });
+
+    const serializedNodePool = this.subject().serialize(
+      nodePool._createSnapshot()
+    );
+
+    assert.deepEqual(serializedNodePool, {
+      Name: 'prod-eng',
+      Description: 'Production workloads',
+      Meta: {
+        env: 'production',
+        team: 'engineering',
+      },
+      SchedulerConfiguration: {
+        SchedulerAlgorithm: 'spread',
+      },
+    });
+  });
+});


philrenaud (approved):

ui/app/models/node-pool.js
```diff
@@ -0,0 +1,22 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Model from '@ember-data/model';
+import { attr } from '@ember-data/model';
+import { computed } from '@ember/object';
+import classic from 'ember-classic-decorator';
+
+@classic
```
lgfa29, line 11: Does it need to be a classic model or can we start to leverage Octane for new models?

ui/app/serializers/node-pool.js
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import ApplicationSerializer from './application';
+import classic from 'ember-classic-decorator';
+
+@classic
+export default class NodePool extends ApplicationSerializer {
+  primaryKey = 'Name';
```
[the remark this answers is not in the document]
lgfa29: Node pool are not namespaced and their names are unique. They are also replicated across regions (like namespaces and ACL policies), so it's OK to consider them IDs 👍

lgfa29 (approved):

ui/tests/unit/serializers/node-pool-test.js
```diff
@@ -0,0 +1,46 @@
...
+      name: 'prod-eng',
+      description: 'Production workloads',
+      meta: {
+        env: 'production',
+        team: 'engineering',
+      },
+      schedulerConfiguration: {
+        SchedulerAlgorithm: 'spread',
+      },
+    });
```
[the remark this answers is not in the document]
ChaiWithJai: That's a great callout.

ui/app/models/node-pool.js
ChaiWithJai: We're stuck on `classic` because we rely on base models to extend that are `classic`. We will likely stay on `classic` for the foreseeable future.

lgfa29: Ah got it, thanks!

ui/app/controllers/clients/index.js
```diff
@@ -164,11 +169,35 @@ export default class IndexController extends Controller.extend(
     return volumes.sort().map((volume) => ({ key: volume, label: volume }));
   }
 
+  @computed('selectionNodePools', 'model.nodePools.[]')
+  get optionsNodePools() {
+    const availableNodePools = this.model.nodePools;
```
lgfa29, line 174: One interesting bit here that I hadn't considered is the special `all` node pool. It's kind of like a "virtual" node pool, in the sense that it always includes all nodes, no matter what node pool they are actually part of.

So I think it would make to remove it from the `availableNodePools` so it's not a dropdown option. Otherwise we would try to match nodes with node pool set to `all`, but this is not allowed:
https://github.com/hashicorp/nomad/blob/84e7cf39f6c7b0aee49d174ba03bd95824c83c59/command/agent/command.go#L390-L393

ui/app/routes/topology.js
lgfa29, line 1: In the RFC I proposed that we add a `Node Pool` filter and the node pool to the node box, next to datacenter.
image

But looking at the code I now realize that each box is a _datacenter_, not a node so the proposal doesn't make much sense 😅 
image

That complicate things a bit, so we may need more time to think through this. I think a toggle or dropdown like 
```
Group By
  - Datacenters
  - Node pools
```

would work, but that's quite a bit of extra work.


So maybe for now we can add the `Node Pool` filter and the node pool name in the node box, but with a tooltip to indicate what it represents:
image

We can even add the tooltip to the other labels so it doesn't look weird that node pools is the only one that has it:
image

I think this would a simple implementation, with just a line like this in [`topo-viz/node.hbs`](https://github.com/hashicorp/nomad/blob/main/ui/app/templates/components/topo-viz/node.hbs#L14-L18), but would give us enough value for now.

```hbs
<span class="bumper-left is-faded tooltip" aria-label="Node pool">{{@node.node.nodePool}}</span>
```

@philrenaud @juliezzhou any thoughts on this?

ui/app/styles/components/dashboard-metric.scss
```diff
@@ -17,6 +17,12 @@
     font-weight: $weight-bold;
     font-size: $size-3;
 
+    &.topo {
```
lgfa29, line 20: Nit-picking on the name here a bit, the `dashboard-metric` class is already restricted to the topoviz page, so `topo` doesn't describe this class very well. It also leaks some assumption on where this is used.

Looking at the page, what differentiates this `metric` from other (like CPU and memory) is that they are justified so maybe naming it so would make the intention more clear?

```suggestion
    &.justify {
```

ui/app/templates/topology.hbs
```diff
@@ -370,21 +370,29 @@
...
+                    <p data-test-alloc-count class="metric topo">
                       {{this.scheduledAllocations.length}}
                       <span class="metric-label">
                         Allocations
                       </span>
                     </p>
                   </div>
+                  <div class="dashboard-metric column">
+                    <p data-test-node-pool-count class="metric topo">
+                      {{this.model.nodePools.length}}
```
lgfa29, line 390: Trying to think about the `all` pool here as well, I'm not sure if I would expect to see it counted 🤔 

But it may be weird to _not_ count it when you try to compare it with something like the `nomad node pool list` output, so it's probably better to count it? 

I think I will ask the rest of the teams for seconds (and thirds, fourths....) opinions 😄

ui/app/templates/components/job-row.hbs
```diff
@@ -30,6 +30,9 @@
     {{this.job.namespace.name}}
   </td>
 {{/if}}
+<td data-test-job-nodepool>
+  {{this.job.nodePool}}
+</td>
```
lgfa29, line 35: This are getting a bit crowded when displaying child jobs 🤔 

image

Perhaps we could omit node pool and namespace if the context is a child job since that information is already in the parent bar?
image

lgfa29 (changes requested):
Looking great!

I think removing the `all` node pool from the client list dropdown is the main blocker, otherwise you would always get an empty list result. 

It would nice to have the topoviz updates as well, since it's such a useful page 😄

ui/app/controllers/clients/index.js
lgfa29: Fixed in https://github.com/hashicorp/nomad/pull/17633

ui/app/routes/topology.js
lgfa29: Fixed in https://github.com/hashicorp/nomad/pull/17634

ui/app/styles/components/dashboard-metric.scss
lgfa29: Fixed in https://github.com/hashicorp/nomad/pull/17634

ui/app/templates/components/job-row.hbs
lgfa29: Fixed in https://github.com/hashicorp/nomad/pull/17640

ui/tests/unit/serializers/node-pool-test.js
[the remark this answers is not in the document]
lgfa29: Fixed in https://github.com/hashicorp/nomad/pull/17638

ui/tests/acceptance/allocation-detail-test.js
```diff
@@ -736,14 +741,18 @@ module('Acceptance | allocation detail (services)', function (hooks) {
   });
 
   test('Allocation has a list of services with active checks', async function (assert) {
-    await visit('jobs/service-haver@default');
-    await click('.allocation-row');
+    const runningAlloc = server.db.allocations.findBy({
+      jobId: 'service-haver',
+      clientStatus: 'running',
+    });
+    await Allocation.visit({ id: runningAlloc.id });
```
lgfa29, line 748: This test was flaky because the order of allocation could change, so the test could end up in an allocation that was not running.

ui/tests/acceptance/job-status-panel-test.js
```diff
@@ -819,6 +821,7 @@ module('Acceptance | job status panel', function (hooks) {
           lost: 0,
           complete: 0.2,
         },
+        groupsCount: 1,
```
lgfa29, line 824: Test was flaky without this because sometimes it would create 2 tasks groups, so the total number of allocations doubled.

ui/tests/integration/components/topo-viz/datacenter-test.js
```diff
@@ -119,15 +119,15 @@ module('Integration | Component | TopoViz::Datacenter', function (hooks) {
...
+        `${formatBytes(memoryReserved, 'MiB')} / ${formatBytes(
           memoryTotal,
           'MiB'
         )}`
       )
     );
     assert.ok(
       TopoVizDatacenter.label.includes(
-        `${formatHertz(cpuReserved, 'MHz')}/${formatHertz(cpuTotal, 'MHz')}`
+        `${formatHertz(cpuReserved, 'MHz')} / ${formatHertz(cpuTotal, 'MHz')}`
```
lgfa29, line 130: Side-effect of adding the tooltips, which meant this text is now split between two `<span>` so there's a space.

ui/tests/acceptance/token-test.js
lgfa29, line 1: I think these changes were also made to help stabilize tests.

ui/tests/acceptance/allocation-detail-test.js
lgfa29: Extracted to https://github.com/hashicorp/nomad/pull/17648

ui/tests/acceptance/job-status-panel-test.js
lgfa29: Extracted to https://github.com/hashicorp/nomad/pull/17648

ui/tests/integration/components/topo-viz/datacenter-test.js
lgfa29: Extracted to https://github.com/hashicorp/nomad/pull/17647

ui/tests/acceptance/token-test.js
lgfa29: Extracted to https://github.com/hashicorp/nomad/pull/17648

philrenaud (approved):

lgfa29 (approved):

Issue #17099, opened 2023-05-05
Look into whether a `job.numDesiredAllocs` property is feasible


philrenaud:
For the updated job status panel, we'd like to show "X of Y Allocations Running", where Y is the desired/intended count.

For Service jobs, this can be determined by adding up group counts.

For System jobs, this is harder: we need to know the logic of the constraint scheduler, among other things.

Would it be possible to get a tally of the nodes the job SHOULD be running on as a property on said job?

schmichael:
I'm a bit embarrassed to say I don't have a good solution to this top of mind, but I am curious if you'd be able to run an experiment:

1. Get a copy of the jobspec.
2. Strip any `update` blocks out of the job so that we get total numbers instead of just want the next step of a rolling update would do.
3. Post this to https://developer.hashicorp.com/nomad/api-docs/jobs#create-job-plan
4. For each task group in `JobPlanResponse.Annotations.DesiredTGUpdates` see if `Desired = Place + InPlaceUpdate + DestructiveUpdate` looks good.

I think that ought to be accurate for system and sysbatch jobs... maybe all jobs... sadly the plan results are a bit more "defined by implementation rather than specification" than I'd like, hence why I'm framing this as an experiment.

We would dig in to validate the approach and results before shipping. I feel like Job Summaries _should_ fit this use case, but I'm afraid they don't. This may uncover a clearer way of reporting "job scheduling status" than we currently provide.

PR #17303, opened 2023-05-24
Versioned job summary (allocation clientStatuses by version and group)

philrenaud:
Gives me a "versioned" summary object to mirror the existing summary, so I can show, for example, only recent clientStatus aggregate info on the jobs index list page.

In practice, this'll give me output that looks like this on a JobSummary:

image

---
Known edge cases and potential improvements:
- **Alloc version update timing**: An allocation that gets its version updated, but does not change clientStatus (such as being bumped from version 0 to 1 when a non-destructive update happens, such as getting a task service's tags array updated) does not trigger the state store. Reconcile Summaries does fix this, though.
- **Indicating "Failed, but replaced so it's fine" status to a user**: Inside a job page, we're able to read all allocs directly and ascertain their .Job.Version and .ClientStatus, and also reason about the job's DesiredStatus. We've waxed about having an easily-accessible [Job.DesiredAllocCount](https://github.com/hashicorp/nomad/issues/17099) before, but there are reasons why that might be tricky to implement as well. From the jobs index page and elsewhere, getting info about something like DesiredCount is not currently possible without making potentially a lot of separate requests.

### DIFF
--- a/nomad/fsm_test.go
+++ b/nomad/fsm_test.go
@@ -2847,6 +2847,126 @@ func TestFSM_ReconcileSummaries(t *testing.T) {
 	}
 }
 
+func TestFSM_ReconcileVersionedSummary(t *testing.T) {
+	logger := testlog.HCLogger(t)
+	logger.Info("+++++")
+
+	ci.Parallel(t)
+	// Add some state
+	fsm := testFSM(t)
+	state := fsm.State()
+
+	// Add a node
+	node := mock.Node()
+	require.NoError(t, state.UpsertNode(structs.MsgTypeTestSetup, 800, node))
+
+	// Make a job
+	job1 := mock.Job()
+	job1.TaskGroups[0].Tasks[0].Resources.CPU = 5000
+
+	// // make a job which can make partial progress
+	alloc := mock.Alloc()
+	alloc.NodeID = node.ID
+	// Not sure why, but have to set both .Job and .JobID for this alloc to show up in the job summary??
+	alloc.Job = job1
+	alloc.JobID = job1.ID
+	alloc.ClientStatus = structs.AllocClientStatusFailed
+
+	req := structs.GenericRequest{}
+	buf, err := structs.Encode(structs.ReconcileJobSummariesRequestType, req)
+	if err != nil {
+		t.Fatalf("err: %v", err)
+	}
+
+	require.NoError(t, state.UpsertJob(structs.MsgTypeTestSetup, 1000, nil, job1))
+	require.NoError(t, state.UpsertAllocs(structs.MsgTypeTestSetup, 1011, []*structs.Allocation{alloc}))
+
+	resp := fsm.Apply(makeLog(buf))
+	if resp != nil {
+		t.Fatalf("resp: %v", resp)
+	}
+
+	ws := memdb.NewWatchSet()
+	out1, err := state.JobSummaryByID(ws, job1.Namespace, job1.ID)
+	require.NoError(t, err)
+	logger.Info("++++THRU 1", "JOBVERSION", job1.Version, "JOBALLOCS", out1)
+
+	expected := structs.JobSummary{
+		JobID:     job1.ID,
+		Namespace: job1.Namespace,
+		Summary: map[string]structs.TaskGroupSummary{
+			"web": {
+				Queued: 10,
+				Failed: 1,
+			},
+		},
+		VersionedSummary: map[uint64]structs.VersionedSummary{
+			0: {
+				Version: 0,
+				Groups: map[string]structs.TaskGroupSummary{
+					"web": {
+						Failed: 1,
+					},
+				},
+			},
+		},
+		CreateIndex: 1000,
+		ModifyIndex: out1.ModifyIndex,
+	}
+	if !reflect.DeepEqual(&expected, out1) {
+		t.Fatalf("expected: %#v, actual: %#v", &expected, out1)
+	}
+
+	// Bump job to 1 and add a couple allocs
+
+	alloc.ClientStatus = structs.AllocClientStatusRunning
+	require.NoError(t, state.UpsertAllocs(structs.MsgTypeTestSetup, 1021, []*structs.Allocation{alloc}))
+	require.NoError(t, state.UpsertJob(structs.MsgTypeTestSetup, 1020, nil, job1))
+
+	// Apply ReconcileJobSummaries request
+	resp = fsm.Apply(makeLog(buf))
+	if resp != nil {
+		t.Fatalf("resp: %v", resp)
+	}
+
+	out2, err := state.JobSummaryByID(ws, job1.Namespace, job1.ID)
+	require.NoError(t, err)
+
+	expected = structs.JobSummary{
+		JobID:     alloc.Job.ID,
+		Namespace: alloc.Job.Namespace,
+		Summary: map[string]structs.TaskGroupSummary{
+			"web": {
+				Queued:  9,
+				Running: 1,
+			},
+		},
+		VersionedSummary: map[uint64]structs.VersionedSummary{
+			// TODO: Why doesn't version 0 show up here?
+			// 0: {
+			// 	Version: 0,
+			// 	Groups: map[string]structs.TaskGroupSummary{
+			// 		"web": {},
+			// 	},
+			// },
+			1: {
+				Version: 1,
+				Groups: map[string]structs.TaskGroupSummary{
+					"web": {
+						Running: 1,
+					},
+				},
+			},
+		},
+		CreateIndex: 1000,
+		ModifyIndex: out2.ModifyIndex,
+	}
+	if !reflect.DeepEqual(&expected, out2) {
+		t.Fatalf("Diff % #v", pretty.Diff(&expected, out2))
+	}
+
+}
+
 // COMPAT: Remove in 0.11
 func TestFSM_ReconcileParentJobSummary(t *testing.T) {
 	// This test exercises code to handle https://github.com/hashicorp/nomad/issues/3886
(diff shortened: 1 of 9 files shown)


nomad/structs/structs.go
```diff
@@ -5058,11 +5060,27 @@ type JobSummary struct {
...
 	*newJobSummary = *js
+
 	newTGSummary := make(map[string]TaskGroupSummary, len(js.Summary))
 	for k, v := range js.Summary {
 		newTGSummary[k] = v
 	}
+
+	newVTGSummary := make(map[uint64]VersionedSummary, len(js.VersionedSummary))
+	for k, v := range js.VersionedSummary {
+		uk := uint64(k)
```
philrenaud, line 5071: Would be thrilled to find out there was a better way to do this!

nomad/state/state_store.go
```diff
@@ -5694,6 +5797,11 @@ func (s *StateStore) updateSummaryWithAlloc(index uint64, alloc *structs.Allocat
 		return fmt.Errorf("unable to find task group in the job summary: %v", alloc.TaskGroup)
 	}
 
+	vtgSummary, ok := jobSummary.VersionedSummary[alloc.Job.Version]
+	if !ok {
+		return fmt.Errorf("unable to find versioned summary for job version: %v", alloc.Job.Version)
+	}
+
 	summaryChanged := false
 	if existingAlloc == nil {
```
philrenaud, line 5806: What should I be trying to do with a new alloc in terms of versioned summary? If it has no version, what would I increment/decrement? In practice, my clientStatus counts for all allocs in the "versioned" summary seem to match the non-versioned one, so I'm not sure if anything is warranted here but wanted to note it.

(I don't think we need to mess with "queued" in the versioned summary.)

nomad/structs/structs.go
philrenaud: nvm — this was from a version where I was trying to set the key to a string well outside of serialization.

nomad/fsm_test.go
```diff
@@ -2847,6 +2847,126 @@ func TestFSM_ReconcileSummaries(t *testing.T) {
...
+	require.NoError(t, state.UpsertNode(structs.MsgTypeTestSetup, 800, node))
+
+	// Make a job
+	job1 := mock.Job()
+	job1.TaskGroups[0].Tasks[0].Resources.CPU = 5000
+
+	// // make a job which can make partial progress
+	alloc := mock.Alloc()
+	alloc.NodeID = node.ID
+	// Not sure why, but have to set both .Job and .JobID for this alloc to show up in the job summary??
```
philrenaud, line 2870: If anybody knows why this happens, please lmk! If I only set .Job or .JobID, the allocation is uncounted in .summary or .versionedSummary

nomad/fsm_test.go
```diff
@@ -2879,7 +2877,10 @@ func TestFSM_ReconcileVersionedSummary(t *testing.T) {
 	}
 
 	require.NoError(t, state.UpsertJob(structs.MsgTypeTestSetup, 1000, nil, job1))
-	require.NoError(t, state.UpsertAllocs(structs.MsgTypeTestSetup, 1011, []*structs.Allocation{alloc}))
+	require.NoError(t, state.UpsertAllocs(structs.MsgTypeTestSetup, 1011, []*structs.Allocation{alloc1}))
+
+	require.NoError(t, state.DeleteJobSummary(1030, job1.Namespace, job1.ID))
+	require.NoError(t, state.DeleteJobSummary(1040, alloc1.Namespace, alloc1.Job.ID))
```
shoenig, line 2883: these seem like 2 ways of saying the same thing?

shoenig: i think these indexes are higher than ones used below; they should be increasing

nomad/fsm_test.go
```diff
@@ -2922,15 +2922,27 @@ func TestFSM_ReconcileVersionedSummary(t *testing.T) {
...
+	alloc3.NodeID = node.ID
+	alloc3.Job = job2
+	alloc3.JobID = job2.ID
+	alloc3.ClientStatus = structs.AllocClientStatusRunning
+	require.NoError(t, state.UpsertJob(structs.MsgTypeTestSetup, 1020, nil, job2))
+	require.NoError(t, state.UpsertAllocs(structs.MsgTypeTestSetup, 1021, []*structs.Allocation{alloc2, alloc3}))
+
+	require.NoError(t, state.DeleteJobSummary(1030, job1.Namespace, job1.ID))
+	require.NoError(t, state.DeleteJobSummary(1035, job2.Namespace, job2.ID))
+	require.NoError(t, state.DeleteJobSummary(1040, alloc1.Namespace, alloc1.Job.ID))
```
shoenig, line 2945: same, should be redundant right?

nomad/state/state_store.go
```diff
@@ -5342,6 +5357,28 @@ func (s *StateStore) updateSummaryWithJob(index uint64, job *structs.Job,
 		}
 	}
 
+	for v := uint64(0); v <= job.Version; v++ {
```
shoenig, line 5360: We'll need to limit the versioned summaries we keep track of; there are use cases where jobs are constantly being updated and will have many thousands of versions, which would be a huge waste of space here.

philrenaud: I think the main thing I want out of this is a "latest version", and having them by older versions seemed like a nice feature complete thing but the more I work on it the less interested in old version statuses I am.

Maybe I should just change the gist of this from a VersionedSummary to a LatestVersionSummary?

nomad/fsm_test.go
philrenaud: Fixed and fixed, thx

nomad/fsm_test.go
philrenaud: Fixed, thanks

nomad/state/state_store.go
philrenaud: I think I'm going to start by keeping the structure the same and only having a latest version in there. This way if we want to give some control to how many versions we store here later, we can do that without completely changing the format.

philrenaud: Ack right, we might have allocs running on a job of the not-latest version that matter. For example in a failed deployment situation. Hm.

nomad/fsm_test.go
```diff
@@ -2836,6 +2848,183 @@ func TestFSM_ReconcileSummaries(t *testing.T) {
...
+			return false, err
+		}
+		if alloc3result != nil {
+			return false, fmt.Errorf("alloc %+v wasn't garbage collected", alloc3)
+		}
+
+		return true, nil
+	}, func(err error) {
+		t.Fatalf("err: %s", err)
+	})
```
philrenaud, line 3024: This part of the test is not working right now; I've decided to get eyes on the PR anyway. I'd like to have this verify that the alloc got GC'd and for it to show up in the summary regardless.

nomad/fsm_test.go
```diff
@@ -2836,6 +2848,183 @@ func TestFSM_ReconcileSummaries(t *testing.T) {
...
+					},
+				},
+			},
+		},
+		CreateIndex: 1000,
+		ModifyIndex: out1.ModifyIndex,
+	}
+	if !reflect.DeepEqual(&expected, out1) {
+		t.Fatalf("expected: %#v, actual: %#v", &expected, out1)
+	}
```
tgross, line 2918:
```suggestion
	must.Eq(t, &expected, out1)
```

nomad/fsm_test.go
```diff
@@ -2836,6 +2848,183 @@ func TestFSM_ReconcileSummaries(t *testing.T) {
...
+	alloc2.JobID = job2.ID
+	alloc2.ClientStatus = structs.AllocClientStatusRunning
+
+	// Create a new alloc here, have it be running
+	alloc3 := mock.Alloc()
+	alloc3.NodeID = node.ID
+	alloc3.Job = job2
+	alloc3.JobID = job2.ID
+	alloc3.ClientStatus = structs.AllocClientStatusRunning
+	require.NoError(t, state.UpsertJob(structs.MsgTypeTestSetup, 1020, nil, job2))
```
tgross, line 2940: We're gradually moving Nomad off of `testify` and towards `shoenig/test`, so let's use `must.NoError` for these kinds of assertions.

nomad/fsm_test.go
tgross: The logic around what gets GC'd is pretty complex and not super relevant to what you're trying to test here. If you look at `nomad/core_sched.go`, you can find that ultimately the GC is going to call `Eval.Reap` with the alloc ID in its request. So you could call that directly, or even go one level down and call `state.DeleteEval` because that's the only logic you'll find in the `Eval.Reap` RPC's write.

At the very least, looking `state.DeleteEval` might help you figure out why the test is failing.

tgross (commented):
This is looking really good @philrenaud! I think we're on the right track here.

> Alloc version update timing: An allocation that gets its version updated, but does not change clientStatus (such as being bumped from version 0 to 1 when a non-destructive update happens, such as getting a task service's tags array updated) does not trigger the state store. Reconcile Summaries does fix this, though.

Oh right, the `Allocation` doesn't get updated in that case, the `Job` does. The interesting problem with this is that the `VersionedJobSummary` can't be updated correctly on `Job.Register` either because we don't actually know whether the existing allocations will need to be destroyed or not until the scheduler runs! Maybe we could transfer the allocation counts to the new versioned summary during the plan applier?

> Indicating "Failed, but replaced so it's fine" status to a user

Yeah, fine that we're not covering that in this PR. I also feel like this wants to live in the client code (UI/CLI) rather than in the API proper, as it's transforming the raw data in a way that's making an interpretation of events.

Issue #17305, opened 2023-05-24
Lookup bridge kernel module in /sys/module/ to enable running on WSL2


jeteve:
### Proposal

To find out if the host linux system is capable of bridge networking, the client fingerprinting looks up in /proc/modules , /lib/modules/<kernel>/modules.builtin and in /lib/modules/<kernel/modules.dep.

Sadly none of those places are populated in WSL2 (tested using ubuntu WSL2) in Windows, even though the WSL2 kernel provided by MS has many modules built-in (including the bridge module).

### Use-cases

End goal: run nomad on WSL2 with bridge networking capability.

### Attempted Solutions

Add a lookup into /sys/module/<module name> . To attempt detecting the availability of the bridge kernel module.

A branch with this solution is provided against this issue

### References
See https://www.kernel.org/doc/Documentation/ABI/stable/sysfs-module

hardkeo:
### Hello @jeteve, today could be your lucky day!

Yesterday I completed a survey on how to do this with minimal steps and/or interventions and I've already tested it in 2 different environments using WSL2 with Ubuntu 22.04.2 vanilla.

---

## Enabling bridge and br_netfilter modules in vanilla WSL2 (Ubuntu-22.04)
### 1. Downloading the necessary requirements
```shell
sudo apt install build-essential flex bison dwarves libssl-dev libelf-dev bc libncurses-dev unzip
```

### 2. Set variable with kernel version
####    ex: WSL_RELEASE=linux-msft-wsl-5.15.90.1
```shell
WSL_RELEASE="linux-msft-wsl-$(echo $(uname -r) | cut -d"-" -f1)"
```

### 3. Download the source code of your kernel version used by wsl2
```shell
curl -LO https://github.com/microsoft/WSL2-Linux-Kernel/archive/refs/tags/${WSL_RELEASE}.zip
```

### 4. Unzip the source code 
```shell
unzip ${WSL_RELEASE}.zip
```

### 5. Load your WSL2 kernel configuration file (Kconfig)
```shell
cd WSL2-Linux-Kernel-${WSL_RELEASE}
zcat /proc/config.gz > .config
```

### 6. Prepare the modules
```shell
make prepare modules_prepare -j $(nproc)
```

### 7. By default, the modules are already enabled in the ".config", I suggest checking them when opening the config menu
####    Modules must be enabled with "*" (built-in)
```shell
make menuconfig
```

####   Settings locations in menuconfig
>\> Networking support
>>    \> Networking options
[...]

PR #17306, opened 2023-05-24
client checks kernel module in /sys/module for WSL2 bridge networking

jeteve:
This is a solution for issue #17305 and will enable nomad to have greater chances to detect the `bridge` kernel module on WSL2 hosts, thus enabling bridge networking.


client/fingerprint/bridge_linux.go
```diff
@@ -89,6 +97,14 @@ func (f *BridgeFingerprint) detect(module string) error {
 	return errs
 }
 
+func (f *BridgeFingerprint) findDir(dirname string) error {
+	if _, err := os.Stat(dirname); os.IsNotExist(err) {
+		return fmt.Errorf("failed to find %s: %v", dirname, err)
+	} else {
+		return nil
+	}
```
lgfa29, line 105: This is only checking for `IsNotExist`, which results in any other kind of error being ignored. We can probably remove the `IsNotExist` call and just check for `err != nil`.

Moreover, this seems like a trivial enough check that could be done in the `detect` method directly:

```go
loadedOrBuiltinPath := fmt.Sprintf("/sys/module/%s", module)
if _, err := os.Stat(loadedOrBuiltinPath); err != nill {
    errs = multierror.Append(errs, err)
} else {
    return nil
}
```

client/fingerprint/bridge_linux.go
```diff
@@ -56,6 +56,14 @@ func (f *BridgeFingerprint) detect(module string) error {
 	// accumulate errors from every place we might find the module
 	var errs error
 
+	// Check if the module is in /sys/modules
+	loadedOrBuiltinPath := fmt.Sprintf("/sys/module/%s", module)
```
lgfa29, line 60: Naming is always hard, so feel free to ignore this, but `loadedOrBuiltinPath` doesn't convey much to me 🤔 

I'm also not a Linux expert, so that very well may be due to my lack of knowledge 😅 

Skimming over [this paper](https://mirrors.edge.kernel.org/pub/linux/kernel/people/mochel/doc/papers/ols-2005/mochel.pdf) it seems that `/sys` is part of a broader feature set called  `sysfs`, so maybe this variable could be called something like `sysfsModulePath`?

lgfa29 (changes requested):
Thanks for the contribution @jeteve!

I left some minor comments about error handling and variable naming but I was able to get this to work after the small adjustment described in https://github.com/microsoft/WSL/issues/6044#issuecomment-1115137696 to enable legacy mode for `iptables`. Did you run into this as well?

I was also not able to get the `docker` driver to work (the `network` pre-run hook fails with the error `failed to Statfs "/var/run/docker/netns/...`), but that looks like a fix for another time 😄 

Using the `exec` driver in `bridge` mode worked well 👍 

I will also ping @shoenig for another check since he has more experience on this area than me.

client/fingerprint/bridge_linux.go
jeteve: Makes sense. Thanks!

jeteve: On the 'no function', I'd like to keep things consistent with the way search File pushes specific errors into the multierr. That's actually how I found where the code was!

jeteve:
Hi @lgfa29 , thanks a lot for your review. I've pushed some amended code.

lgfa29 (approved):
Changes look good to me, thanks @jeteve!

I just pushed a commit to add a CHANGELOG entry.

jeteve:
Thanks a lot @lgfa29 , and apologies for skipping the CHANGELOG myself.

I see all checks are good except 'Vercel – nomad'. I assume unblocking this is part of your release flow?

Thanks!

shoenig (approved):
LGTM!

PR #17309, opened 2023-05-25
Add the ability to customise the details of the CA

lhaig:
Allow operators to customise the CA created using `nomad tls ca create`

### DIFF
--- a/.changelog/17309.txt
+++ b/.changelog/17309.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-tls: Add the ability to customize the details of the CA
+cli: Add the ability to customize the details of the CA when running `nomad tls ca create`
 ```
--- a/.changelog/17309.txt
+++ b/.changelog/17309.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+tls: Add the ability to customize the details of the CA
+```
--- a/command/tls_ca_create.go
+++ b/command/tls_ca_create.go
@@ -34,6 +34,27 @@ type TLSCACreateCommand struct {
 	// additionalDomain provides a list of restricted domains to the CA which
 	// will then reject any domains other than these.
 	additionalDomain flags.StringFlag
+
+	// country is used to set a country code for the CA
+	country string
+
+	// postalCode is used to set a postal code for the CA
+	postalCode string
+
+	// province is used to set a province for the CA
+	province string
+
+	// locality is used to set a locality for the CA
+	locality string
+
+	// streetAddress is used to set a street address for the CA
+	streetAddress string
+
+	// organization is used to set an organization for the CA
+	organization string
+
+	// organizationalUnit is used to set an organizational unit for the CA
+	organizationalUnit string
 }
 
 func (c *TLSCACreateCommand) Help() string {
@@ -53,6 +74,9 @@ CA Create Options:
   -common-name
     Common Name of CA. Defaults to "Nomad Agent CA".
 
+  -country
+    Country of the CA. Defaults to "US".
+
   -days
     Provide number of days the CA is valid for from now on.
     Defaults to 5 years or 1825 days.
@@ -61,12 +85,31 @@ CA Create Options:
     Domain of Nomad cluster. Only used in combination with -name-constraint.
     Defaults to "nomad".
 
+  -locality
+    Locality of the CA. Defaults to "San Francisco".
+
   -name-constraint
     Enables the DNS name restriction functionality to the CA. Results in the CA
     rejecting certificates for any other DNS zone. If enabled, localhost and the
     value of -domain will be added to the allowed DNS zones field. If the UI is
     going to be served over HTTPS its hostname must be added with
     -additional-domain. Defaults to false.
+
+  -organization
+    Organization of the CA. Defaults to "HashiCorp Inc.".
+
+  -organizational-unit
+    Organizational Unit of the CA. Defaults to "Nomad".
+
+  -postal-code
+    Postal Code of the CA. Defaults to "94105".
+
+  -province
+    Province of the CA. Defaults to "CA".
+
+  -street-address
+    Street Address of the CA. Defaults to "101 Second Street".
+
 `
 	return strings.TrimSpace(helpText)
 }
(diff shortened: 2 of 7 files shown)


.changelog/17309.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+tls: Add the ability to customize the details of the CA
```
jrasell, line 2: We tend to group them by process, so this would be `cli`:
```suggestion
cli: Add the ability to customize the details of the CA when running `nomad tls ca create`
```

command/tls_ca_create.go
```diff
@@ -61,24 +85,50 @@ CA Create Options:
...
     going to be served over HTTPS its hostname must be added with
     -additional-domain. Defaults to false.
+
+  -organization
+    Organization of the CA. Defaults to "HashiCorp Inc.".
+
+  -organizational-unit
+    Organizational Unit of the CA. Defaults to "Nomad".
+
+  -postal-code
```
jrasell, line 104: This is defined in the flags as `post-code`, so I believe either this or the flag declaration need to change.

command/tls_ca_create.go
```diff
@@ -61,24 +85,50 @@ CA Create Options:
...
+			"-additional-domain":   complete.PredictAnything,
+			"-common-name":         complete.PredictAnything,
+			"-days":                complete.PredictAnything,
+			"-country":             complete.PredictAnything,
+			"-domain":              complete.PredictAnything,
+			"-locality":            complete.PredictAnything,
+			"-name-constraint":     complete.PredictAnything,
+			"-organization":        complete.PredictAnything,
+			"-organizational-unit": complete.PredictAnything,
+			"-postcode":            complete.PredictAnything,
```
jrasell, line 129: This is declared in the help as `post-code` and the flag setup as `postal-code` so will need to be updated along with either of those.

helper/tlsutil/generate.go
```diff
@@ -116,17 +124,53 @@ func GenerateCA(opts CAOpts) (string, string, error) {
...
+
+	organization := opts.Organization
+	if organization == "" {
+		organization = ("HashiCorp Inc.")
+	}
+
+	organizationalUnit := opts.OrganizationalUnit
+	if organizationalUnit == "" {
+		organizationalUnit = ("Nomad")
+	}
```
jrasell, line 160: The assignments do not require parentheses; so `organizationalUnit = ("Nomad")` can become `organizationalUnit = "Nomad"` for example.

jrasell (changes requested):
I have added some initial comments inline along with an internal note about a change I believe is required for this to work. It would also be nice if new tests used `must` rather than require, so we continue moving towards this library.

command/tls_ca_create.go
lhaig: yes I messed up on this one.
I need to change everything to postal-code

helper/tlsutil/generate.go
```diff
@@ -81,57 +89,122 @@ type CertOpts struct {
...
+		if opts.Country == "" {
+			return "", "", errors.New("please provide the -country flag when customizing the CA")
+		}
+
+		if opts.Organization == "" {
+			return "", "", errors.New("please provide the -organization flag when customizing the CA")
+		}
+
+		if opts.OrganizationalUnit == "" {
+			return "", "", errors.New("please provide the -organizational-unit flag when customizing the CA")
```
jrasell, line 187: The errors are specific to the CLI implementation; I wonder if we want to make these more generic, so future internal callers do not emit errors related to CLI flags when they do not implement them?

jrasell (changes requested):
Thanks for the changes @lhaig; I left some comments inline, mostly stylistic ones and some areas we could reduce the amount of code.

I feel we could make some changes to the validation logic as well after testing this locally. I would expect to be able to modify only the `organizational-unit` for example, however, when running this I get an error:
```
./pkg/darwin_arm64/nomad tls ca create --organizational-unit="nomad engineering"
please provide the -common-name flag when customizing the CA
```

I therefore think we need some type of merge function, which merges default values onto a potentially partially populated opts struct.

helper/tlsutil/generate.go
```diff
@@ -81,57 +89,122 @@ type CertOpts struct {
...
+	return caOpts.Days == 0 &&
+		caOpts.Country == "" &&
+		caOpts.PostalCode == "" &&
+		caOpts.Province == "" &&
+		caOpts.Locality == "" &&
+		caOpts.StreetAddress == "" &&
+		caOpts.Organization == "" &&
+		caOpts.OrganizationalUnit == "" &&
+		caOpts.Name == ""
+}
```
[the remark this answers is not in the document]
lhaig: @jrasell thanks for your comments. 
The reason for the error when just updating the -organisational-unit covers the scenario where you would have a completely unrelated org Unit mixed in with the official HashiCorp address and details.
My logic on this is if you want to update the org you will most probably not want to use the defaults and should thus provide the other required details.

command/tls_ca_create.go
```diff
@@ -163,3 +259,17 @@ func (c *TLSCACreateCommand) Run(args []string) int {
 
 	return 0
 }
+
+// IsEmpty checks whether any of TLSCACreateCommand parameters have been populated with
+// non-default values.
+func (c TLSCACreateCommand) IsEmpty() bool {
```
jrasell, line 265: Use a pointer otherwise `TLSCACreateCommand` has methods on both value and pointer receivers.
```suggestion
func (c *TLSCACreateCommand) IsEmpty() bool {
```

This function name is also a little misleading, as it doesn't check whether the command is totally empty, just certain parameters.

helper/tlsutil/generate.go
```diff
@@ -93,45 +118,74 @@ func GenerateCA(opts CAOpts) (string, string, error) {
...
+		opts.OrganizationalUnit = "Nomad"
+	} else if opts.IsEmpty() && opts.Days != 0 {
+		opts.Name = fmt.Sprintf("Nomad Agent CA %d", sn)
+		opts.Country = "US"
+		opts.PostalCode = "94105"
+		opts.Province = "CA"
+		opts.Locality = "San Francisco"
+		opts.StreetAddress = "101 Second Street"
+		opts.Organization = "HashiCorp Inc."
+		opts.OrganizationalUnit = "Nomad"
```
jrasell, line 152: The only difference between the two conditionals is the days; could we therefore add a check for the days inside the `IsEmpty` conditional?

helper/tlsutil/generate.go
```diff
@@ -81,10 +88,28 @@ type CertOpts struct {
 	ExtKeyUsage []x509.ExtKeyUsage
 }
 
+// IsEmpty checks whether any of CAOpts parameters have been populated with
+// non-default values.
+func (c CAOpts) IsEmpty() bool {
```
jrasell, line 93: Is this function name accurate? There are a number of parameters which can be non-default and this would return true.

command/tls_ca_create.go
```diff
@@ -112,6 +169,34 @@ func (c *TLSCACreateCommand) Run(args []string) int {
...
+		}
+
+		if c.organization == "" {
+			c.Ui.Error("please provide the -organization flag when customizing the CA")
+			c.Ui.Error(commandErrorText(c))
+			return 1
+		}
+
+		if c.organizationalUnit == "" {
+			c.Ui.Error("please provide the -organizational-unit flag when customizing the CA")
```
jrasell, line 195: These errors are directly printed to the users console, so we should title the sentence.

jrasell (changes requested):
Some inline comments. It would probably also be nice to expand on the command description, given that defaults don't take effect if you modify certain parameters.

jrasell (approved):
LGTM, thanks so much @lhaig!

Issue #17108, opened 2023-05-08
Add group_add to job docker config


Kamilcuk:
### Proposal
This is a proposal to add a `group_add` argument to docker task `"config"`-uration that will mimic the `docker --group_add` and `docker compose` `group_add:` options.

```
task "..." {
   driver = "docker"
   user = "user1:group1"
    config {
       image = "alpine"
       group_add = [
          "group1",
          "group2",
          "1234",
       ]
   }
}
```

From `docker run --help`:

      --group-add list                 Add additional groups to join

### Use-cases
The use case is to run a user with more than one group. For example, there is an NFS share where directories are owned by specific, multiple, different groups.

### Attempted Solutions
The solution or workaround is to create a custom image with Dockerfile where you `RUN useradd && groupadd` the required groups to the user.

Related:
https://github.com/hashicorp/nomad/pull/8066
@dnrce could you sign CLA? Thank you.
https://discuss.hashicorp.com/t/inject-default-user-group-for-docker-by-nomad-client/18895

tgross:
@Kamilcuk can you do us a favor and revisit what you've written here and try to turn it into a specific proposal? This is a bunch of links and @-mentions of unrelated folks on a closed PR.

schmichael:
Thanks for clarifying the issue @Kamilcuk. Looks like a straightforward case of Nomad needing to implement an additional Docker configuration option. I updated the labels so we can pick it up someday.

These sorts of plugin improvements are good opportunities for contributions as well, so I added the `good first issue` label. The Docker config parameters generally follow a pretty standard path in https://github.com/hashicorp/nomad/blob/v1.5.5/drivers/docker/config.go

PR #17313, opened 2023-05-25
Add group_add docker option

Kamilcuk:
fix https://github.com/hashicorp/nomad/issues/17108


lgfa29 (approved):
Thank you for the contribution @Kamilcuk!

I pushed a commit to add a CHANGELOG entry but everything looks to me.

PR #17314, opened 2023-05-25
fix: job canonicalization should set job priority to 50, not 0.

pkazmierczak:
Nomad API will reject jobs with priority set to 0. 
```
        	* Job priority must be between [1, 100])
```


tgross (changes requested):
@pkazmierczak this looks right at first in terms of the canonicalization but in the `Job.Register` RPC we set the priority to the default priority (50) if the priority is 0. So this would break that and make it so that jobs with unset priority are set to 1 instead of 50.

tgross (approved):
LGTM!

PR #17315, opened 2023-05-25
refact:  update models for relationship

### LINKED ISSUE #17293: Update Ember Data models to add `nodePool` attribute onto `Job` and `Node` models.

ChaiWithJai:
Resolves #17293 

This PR adds a relationship between the Node and NodePool models. In addition, The Job model now has an @attr('string') property called nodePool.

The nodePool property on the Node model is used to store the name of the NodePool that the job is associated with. The nodes property on the NodePool model is used to store a list of the Job models that are associated with the NodePool.

This change was made to allow users to easily associate jobs with node pools. For example, a user could create a job and specify the name of the node pool that the job should be run on.

**Testing**
The changes in this PR have been tested using the following tests:

`ui/tests/unit/models/node-pool-test.js`

The tests have been written to ensure that the Node and NodePool models can be associated with each other correctly.

**Impact**
This change will have a minor impact on the codebase. The Node and NodePool models will now have a relationship with each other.

**Next Steps**
The next steps for this change are to seed the Mirage environments such that `node pools` are reflected. Once that is ready, we'll be ready to visualize the changes in the UI against both a Mirage scenario so that design can sign off and against the real API changes to smoke test the changes.

### DIFF
--- a/ui/app/models/job.js
+++ b/ui/app/models/job.js
@@ -28,7 +28,7 @@ export default class Job extends Model {
   @attr('number') createIndex;
   @attr('number') modifyIndex;
   @attr('date') submitTime;
-  @attr('string') nodePool; // Jobs are related to Node Pools via Allocations, but no relationship.
+  @attr('string') nodePool; // Jobs are related to Node Pools either directly or via its Namespace, but no relationship.
 
   @fragment('structured-attributes') meta;
 
--- a/ui/app/models/job.js
+++ b/ui/app/models/job.js
@@ -28,6 +28,7 @@ export default class Job extends Model {
   @attr('number') createIndex;
   @attr('number') modifyIndex;
   @attr('date') submitTime;
+  @attr('string') nodePool; // Jobs are related to Node Pools via Allocations, but no relationship.
 
   @fragment('structured-attributes') meta;
 
--- a/ui/tests/unit/models/node-test.js
+++ b/ui/tests/unit/models/node-test.js
@@ -1,0 +1,28 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('Unit | Model | NodePool', function (hooks) {
+  setupTest(hooks);
+
+  test('a node pool can be associated with multiples nodes', function (assert) {
+    const store = this.owner.lookup('service:store');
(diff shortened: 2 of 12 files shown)


ui/app/models/job.js
```diff
@@ -28,6 +28,7 @@ export default class Job extends Model {
   @attr('number') createIndex;
   @attr('number') modifyIndex;
   @attr('date') submitTime;
+  @attr('string') nodePool; // Jobs are related to Node Pools via Allocations, but no relationship.
```
lgfa29, line 31:
```suggestion
  @attr('string') nodePool; // Jobs are related to Node Pools either directly or via its Namespace, but no relationship.
```

The job can be directly assigned to a node pool in the jobspec or inherit the default pool from the namespace. 

I'm not sure what `but no relationship` means in this context, is it something like `@belongsTo('node-pool')`?

ui/tests/unit/models/node-test.js
```diff
@@ -0,0 +1,28 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('Unit | Model | NodePool', function (hooks) {
+  setupTest(hooks);
+
+  test('a node pool can be associated with multiples nodes', function (assert) {
```
lgfa29, line 12:
```suggestion
  test('a node pool can have multiples nodes', function (assert) {
```

Minor nit-picking, but I think this describes the expected relationship better.

ui/tests/unit/models/node-test.js
```diff
@@ -0,0 +1,28 @@
...
+    const node2 = store.createRecord('node', { name: 'Node 2' });
+
+    nodePool.get('nodes').pushObject(node1);
+    nodePool.get('nodes').pushObject(node2);
+
+    const nodes = nodePool.get('nodes');
+
+    assert.equal(nodes.length, 2);
+    assert.ok(nodes.includes(node1));
+    assert.ok(nodes.includes(node2));
```
lgfa29, line 26: It may be worth checking that the inverse relationship is also set, so check that `node1` and `node2` belong to `nodePool`.

lgfa29 (approved):
Looking good! Just left some minor comments and suggestions

ui/tests/unit/models/node-test.js
ChaiWithJai: That's a good call!

philrenaud (approved):

PR #17316, opened 2023-05-25
[core] nil check and error handling for client status in heartbeat responses

### LINKED ISSUE #17310: Nomad v1.4.3 crashes with 'panic: runtime error: invalid memory address or nil pointer dereference'
### Nomad version
v1.4.3
Autoscaler version - v0.3.7

Ubuntu 20.04
GCP - 3 server cluster, @180 client nodes, @1300 allocations.
Servers are n2-standard-16 - CPUx16, 64Gb RAM, 200Gb SSD - Usage on Cpu, Memory, Network, Disk is all low @20% max.
Config files, logs leading to crash and startup logs attached.

### Issue
Nomad server crashes, cluster loses a node.
The issue occurred as Nomad was scaling in a node class from 144 -> 126 over a period of 7m
But this may be a red herring as we regularly have large scaling events, shortly afterwards the impaired cluster scaled the node class from 150 -> 5 nodes over a period of 10m without issue.

### Reproduction steps
Cannot reproduce, this is the first time.

#### Expected Result
Nomad server does not crash.

#### Actual Result
Nomad server crashes, cluster loses a node.

### Job file (if appropriate)
Many jobs running - the large node class is running a CPU intensive batch job.
The drain strategy is empty_ignore_system

Will send logs and config files by email.

angrycub:
Add a nil check to `constructNodeServerInfoResponse` to manage an apparent race between deregister and client heartbeats.

Should fix #17310

### DIFF
--- a/.changelog/17316.txt
+++ b/.changelog/17316.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core: Fix panic around client deregistration and pending heartbeats
+```
--- a/client/client.go
+++ b/client/client.go
@@ -2024,7 +2024,8 @@ func (c *Client) updateNodeStatus() error {
 	}
 
 	// Check heartbeat response for information about the server-side scheduling
-	// state of this node
+	// state of this node. If there are errors on the server side, this will come
+	// back as an empty string.
 	c.UpdateConfig(func(c *config.Config) {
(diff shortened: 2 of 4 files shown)


tgross (approved):
LGTM!

PR #17321, opened 2023-05-25
chore: setup mirage models + factory hooks

### LINKED ISSUE #17294: Update Mirage Factories for Jobs and Nodes to include Node Pools.

ChaiWithJai:
Resolves #17294 

This PR adds a new `node-pool` Mirage model and updates the `node` Mirage model that has a `node-pool` relationship. It also updates the factories for `node` to create an associated `node-pool` and the `job` factory to associate itself to a `node-pool`.

**Impact**
This PR impacts our Mirage server and the fake calls we make.

**Side Effect**
This PR runs prettier on `mirage/factories/node` so just pay attention to the `afterCreate` hook changes.

**Next Steps**
This is the final prerequisite we need to start developing against the real API. We'll start PRs that impact the view layer next.

### DIFF
--- a/ui/mirage/factories/job.js
+++ b/ui/mirage/factories/job.js
@@ -214,6 +214,15 @@ export default Factory.extend({
       });
     }
 
+    if (!job.nodePool) {
+      const nodePool = server.db.nodePools.length
+        ? pickOne(server.db.nodePools).name
+        : server.create('node-pool').name;
+      job.update({
+        nodePool,
+      });
+    }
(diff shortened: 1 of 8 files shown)


ChaiWithJai (commented):
@philrenaud could you pull this down and run the test suite? I'm failing unrelated tests that I believe might have to do with your work, but I'm unsure why that would happen.

philrenaud (approved):

Issue #17260, opened 2023-05-19
Fail an update where Nomad stops attempting to reschedule allocations


nicoche:
### Proposal

Hi!

I'm not sure if it is a bug or an unimplemented feature, but here goes.

Given a deployment where the reschedule policy is not unlimited, the deployment does not fail when the max number of reschedules is met.

For example, given this spec:

```hcl
job "a-broken-job" {
  type = "service"

  update {
    max_parallel = 1
    healthy_deadline = "5m"
    progress_deadline = "10m"
    auto_promote = true
    canary = 1
  }

  reschedule {
    attempts = 1
    interval = "24h"
    delay = "5s"
    delay_function = "constant"
    unlimited = false
  }

  group "a-broken-group" {
    count = 1

    task "a-broken-task" {
      driver = "docker"

      config {
        image = "this-image-does-not-exist:latest"
      }
    }
  }
}
```

Nomad will attempt to run exactly two allocations, which will both fail in seconds. Then, the Deployment will be stuck in Pending while the `progress_deadline` is met. When it's met, the deployment will fail.

As far as I know, when the reschedule limit is met, nothing can save that deployment from failing.
I would like to know if there is a mechanism that would allow the deployment to fail whenever that reschedule limit is met.

I hope I have not overlooked any configuration block that does exactly this - if no mechanism like this one exists, I'm happy to provide a PR 🙂.

By advance, thanks!

______

<details>
  <summary>Screenshots of submitted spec</summary>
  
<img width="879" alt="image" src="https://github.com/hashicorp/nomad/assets/78445450/22d22ca4-47c8-4c30-a833-cd773dd563b8">

<img width="888" alt="image" src="https://github.com/hashicorp/nomad/assets/78445450/33142043-ba43-452c-ba30-2fe64dd6ca1c">
  ```
  
</details>

nicoche:
Hi @tgross !

Thanks for you fast answer and the code pointer 🙂.

I'll take a look at this at the beginning of next week and submit a PR

nicoche:
Hi @tgross ! 

I've pushed a patch. Before I add tests, could you let me know if the overall approach looks good to you? I also have left a few questions in the PR and would appreciate your input.


______________


A few notes:
The patch seems to work properly on one of our staging environments.

> One element to throw in here is that the update block doesn't apply when first deploying the job

It does, actually 🤔.
[...]

PR #17341, opened 2023-05-29
deploymentwatcher: fail early whenever possible

nicoche:
Given a deployment that has a progress_deadline, if a task group runs out of reschedule attempts, allow it to fail at this time instead of waiting until the progress_deadline is reached.

See #17260

### DIFF
--- a/nomad/deploymentwatcher/deployment_watcher.go
+++ b/nomad/deploymentwatcher/deployment_watcher.go
@@ -615,12 +615,30 @@ func (w *deploymentWatcher) handleAllocUpdate(allocs []*structs.AllocListStub) (
 			continue
 		}
 
-		// Determine if the update block for this group is progress based
-		progressBased := dstate.ProgressDeadline != 0
+		hasProgressDeadline := dstate.ProgressDeadline != 0
+		failDeployment := false
+
+		// Fail on the first unhealthy allocation if no progress deadline is specified.
+		if !hasProgressDeadline && alloc.DeploymentStatus.IsUnhealthy() {
+			w.logger.Debug("failing deployment because an allocation failed and the deployment is not progress based", alloc.ID)
+			failDeployment = true
+		}
+
+		if hasProgressDeadline && alloc.DeploymentStatus.IsUnhealthy() &&
+			deployment.Active() {
+			reschedulePolicy := w.j.LookupTaskGroup(alloc.TaskGroup).ReschedulePolicy
+			isRescheduleEligible := alloc.RescheduleEligible(reschedulePolicy, time.Unix(alloc.ModifyTime, 0))
+			if !isRescheduleEligible {
+				// We have run out of reschedule attempts: do not wait for the progress deadline to expire because
+				// we can fail early
+				w.logger.Debug("failing deployment because an allocation has failed and the task group has run out of reschedule attempts", alloc.ID)
+				failDeployment = true
+			}
+		}
 
 		// Check if the allocation has failed and we need to mark it for allow
 		// replacements
-		if progressBased && alloc.DeploymentStatus.IsUnhealthy() &&
+		if !failDeployment && alloc.DeploymentStatus.IsUnhealthy() &&
 			deployment.Active() && !alloc.DesiredTransition.ShouldReschedule() {
 			res.allowReplacements = append(res.allowReplacements, alloc.ID)
 			continue
@@ -631,19 +649,12 @@ func (w *deploymentWatcher) handleAllocUpdate(allocs []*structs.AllocListStub) (
 			res.createEval = true
 		}
 
-		// If the group is using a progress deadline, we don't have to do anything.
-		if progressBased {
-			continue
-		}
-
-		// Fail on the first bad allocation
-		if alloc.DeploymentStatus.IsUnhealthy() {
+		if failDeployment {
 			// Check if the group has autorevert set
 			if dstate.AutoRevert {
 				res.rollback = true
 			}
 
-			// Since we have an unhealthy allocation, fail the deployment
 			res.failDeployment = true
 		}
 
--- a/nomad/deploymentwatcher/deployment_watcher.go
+++ b/nomad/deploymentwatcher/deployment_watcher.go
@@ -702,7 +702,7 @@ func (w *deploymentWatcher) shouldFailEarly(deployment *structs.Deployment, allo
 
 	// Fail on the first unhealthy allocation if no progress deadline is specified.
 	if dstate.ProgressDeadline == 0 {
-		w.logger.Debug("failing deployment because an allocation failed and the deployment is not progress based", alloc.ID)
+		w.logger.Debug("failing deployment because an allocation failed and the deployment is not progress based", "alloc", alloc.ID)
 		return true
 	}
 
@@ -712,7 +712,7 @@ func (w *deploymentWatcher) shouldFailEarly(deployment *structs.Deployment, allo
 		if !isRescheduleEligible {
 			// We have run out of reschedule attempts: do not wait for the progress deadline to expire because
 			// we know that we will not be able to try to get another allocation healthy
-			w.logger.Debug("failing deployment because an allocation has failed and the task group has run out of reschedule attempts", alloc.ID)
+			w.logger.Debug("failing deployment because an allocation has failed and the task group has run out of reschedule attempts", "alloc", alloc.ID)
 			return true
 		}
 	}
--- a/nomad/deploymentwatcher/deployment_watcher.go
+++ b/nomad/deploymentwatcher/deployment_watcher.go
@@ -700,10 +700,8 @@ func (w *deploymentWatcher) shouldFailEarly(deployment *structs.Deployment, allo
 		return false
 	}
 
-	hasProgressDeadline := dstate.ProgressDeadline != 0
-
 	// Fail on the first unhealthy allocation if no progress deadline is specified.
-	if !hasProgressDeadline {
+	if dstate.ProgressDeadline == 0 {
 		w.logger.Debug("failing deployment because an allocation failed and the deployment is not progress based", alloc.ID)
 		return true
 	}
--- a/nomad/deploymentwatcher/deployment_watcher.go
+++ b/nomad/deploymentwatcher/deployment_watcher.go
@@ -615,30 +615,12 @@ func (w *deploymentWatcher) handleAllocUpdate(allocs []*structs.AllocListStub) (
 			continue
 		}
 
-		hasProgressDeadline := dstate.ProgressDeadline != 0
-		failDeployment := false
-
-		// Fail on the first unhealthy allocation if no progress deadline is specified.
-		if !hasProgressDeadline && alloc.DeploymentStatus.IsUnhealthy() {
-			w.logger.Debug("failing deployment because an allocation failed and the deployment is not progress based", alloc.ID)
-			failDeployment = true
-		}
-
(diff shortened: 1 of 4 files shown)


nomad/structs/structs.go
```diff
@@ -11069,6 +11069,48 @@ type AllocListStub struct {
 	ModifyTime            int64
 }
 
+// RescheduleEligible returns if the allocation is eligible to be rescheduled according
+// to its ReschedulePolicy and the current state of its reschedule trackers
+func (a *AllocListStub) RescheduleEligible(reschedulePolicy *ReschedulePolicy, failTime time.Time) bool {
```
nicoche, line 11074: For now, this is a copy/paste from `Allocation.RescheduleEligible`. What approach you would recommend in order to prevent this code duplication?

nomad/deploymentwatcher/deployment_watcher.go
```diff
@@ -615,12 +615,30 @@ func (w *deploymentWatcher) handleAllocUpdate(allocs []*structs.AllocListStub) (
...
+		// Fail on the first unhealthy allocation if no progress deadline is specified.
+		if !hasProgressDeadline && alloc.DeploymentStatus.IsUnhealthy() {
+			w.logger.Debug("failing deployment because an allocation failed and the deployment is not progress based", alloc.ID)
+			failDeployment = true
+		}
+
+		if hasProgressDeadline && alloc.DeploymentStatus.IsUnhealthy() &&
+			deployment.Active() {
+			reschedulePolicy := w.j.LookupTaskGroup(alloc.TaskGroup).ReschedulePolicy
+			isRescheduleEligible := alloc.RescheduleEligible(reschedulePolicy, time.Unix(alloc.ModifyTime, 0))
```
nicoche, line 630: Can we use `alloc.ModifyTime` here? I'm afraid that this timestamp could also refer to other kind of updates, not only the one we are interested in: the time it turned to Failed 😕

tgross: Looking at the logic in `node_endpoint.go` and `RescheduleEligible`, shouldn't the time we're using here be the current time? I don't think we care about the time it was last written to Raft (which could be a little bit ago at this point).

nomad/structs/structs.go
tgross: It looks like the only thing these methods reference from `Allocation` is the `RescheduleTracker`. Could we change it to be a method on `RescheduleTracker` itself? That's the same type between `Allocation` and `AllocListStub`.

nomad/deploymentwatcher/deployment_watcher.go
```diff
@@ -615,12 +615,30 @@ func (w *deploymentWatcher) handleAllocUpdate(allocs []*structs.AllocListStub) (
 			continue
 		}
 
-		// Determine if the update block for this group is progress based
-		progressBased := dstate.ProgressDeadline != 0
+		hasProgressDeadline := dstate.ProgressDeadline != 0
+		failDeployment := false
```
tgross, line 619: Because everything between line 619 and line 637 is there to set `failDeployment` with no side-effects (other than logging), it might make sense to extract that to it own function. That may make it easier to write tests of just that logic, too.

tgross (commented):
This is looking great so far @nicoche! A few things we'll want before it's ready to merge:
* Run `make cl` to add a changelog entry
* Tests: we should make sure the existing tests work (or are altered as needed) of course, but we should probably include a test that exercises these specific behaviors.

nicoche:
Hey @tgross !

Thanks for your comments. I incorporated them into the PR. I haven't squashed the commits because I figured out that it would make things easier for you if you wanted to check the changes. They should have all been addressed.

Btw, did you see https://github.com/hashicorp/nomad/issues/17260#issuecomment-1566949370 ? Especially the part regarding the behavior of an `update` block for a first deployment?

nomad/structs/structs.go
```diff
@@ -10527,20 +10569,12 @@ func (a *Allocation) RescheduleEligible(reschedulePolicy *ReschedulePolicy, fail
...
-	if (a.RescheduleTracker == nil || len(a.RescheduleTracker.Events) == 0) && attempts > 0 {
-		return true
+
+	if a.RescheduleTracker == nil {
+		return reschedulePolicy.Attempts > 0
 	}
-	attempted, _ := a.rescheduleInfo(reschedulePolicy, failTime)
-	return attempted < attempts
+
+	return a.RescheduleTracker.RescheduleEligible(reschedulePolicy, failTime)
```
tgross, line 10577: By pushing all the checks _except_ the `ReschedulerTracker == nil` into this function, haven't we changed the order of checks such that we're returning based on the reschedule policy attempts even if `reschedulePolicy.Unlimited` is set? 

We could safely handle the `RescheduleTracker == nil` case at the top of the `RescheduleEligible` method (you can call a method with a pointer receiver even if the pointer is nil). That would let both the `Allocation` and `AllocListStub` version of this share their code entirely.

nomad/deploymentwatcher/deployment_watcher.go
```diff
@@ -702,6 +695,33 @@ func (w *deploymentWatcher) shouldFail() (fail, rollback bool, err error) {
...
+		return true
+	}
+
+	if deployment.Active() {
+		reschedulePolicy := w.j.LookupTaskGroup(alloc.TaskGroup).ReschedulePolicy
+		isRescheduleEligible := alloc.RescheduleEligible(reschedulePolicy, time.Now())
+		if !isRescheduleEligible {
+			// We have run out of reschedule attempts: do not wait for the progress deadline to expire because
+			// we know that we will not be able to try to get another allocation healthy
+			w.logger.Debug("failing deployment because an allocation has failed and the task group has run out of reschedule attempts", alloc.ID)
```
tgross, line 717: hclog requires pairs of key-value pairs following the first param.

```suggestion
			w.logger.Debug("failing deployment because an allocation has failed and the task group has run out of reschedule attempts", "alloc", alloc.ID)
```

Otherwise you end up with logs like this:

> 2023-06-12T11:47:52.630-0400 [DEBUG] nomad.deployments_watcher: failing deployment because an allocation has failed and the task group has run out of reschedule attempts: deployment_id=d73fc44f-d9a9-583a-d563-13a758dca813 job="<ns: \"default\", id: \"mock\">" EXTRA_VALUE_AT_END=a947bcfa-81b4-2e81-8865-d92474fcdf87

nomad/deploymentwatcher/deployment_watcher.go
```diff
@@ -702,6 +695,33 @@ func (w *deploymentWatcher) shouldFail() (fail, rollback bool, err error) {
...
+func (w *deploymentWatcher) shouldFailEarly(deployment *structs.Deployment, alloc *structs.AllocListStub, dstate *structs.DeploymentState) bool {
+	if !alloc.DeploymentStatus.IsUnhealthy() {
+		return false
+	}
+
+	hasProgressDeadline := dstate.ProgressDeadline != 0
+
+	// Fail on the first unhealthy allocation if no progress deadline is specified.
+	if !hasProgressDeadline {
+		w.logger.Debug("failing deployment because an allocation failed and the deployment is not progress based", alloc.ID)
```
tgross, line 707:
```suggestion
		w.logger.Debug("failing deployment because an allocation failed and the deployment is not progress based", "alloc", alloc.ID)
```

nomad/deploymentwatcher/deployments_watcher_test.go
```diff
@@ -1784,6 +1784,91 @@ func TestDeploymentWatcher_Watch_StartWithoutProgressDeadline(t *testing.T) {
 	})
 }
 
+// Test that we exit before hitting the Progress Deadline when we run out of reschedule attempts
+// for a failing deployment
+func TestDeploymentWatcher_Watch_FailEarly(t *testing.T) {
+       ci.Parallel(t)
+       require := require.New(t)
```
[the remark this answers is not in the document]
nicoche: I've migrated this test to make it use `shoenig/test` ✔️

nomad/deploymentwatcher/deployment_watcher.go
```diff
@@ -702,6 +695,33 @@ func (w *deploymentWatcher) shouldFail() (fail, rollback bool, err error) {
...
+		return true
+	}
```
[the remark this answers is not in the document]
nicoche: Indeed, fixed ✔️

nomad/structs/structs.go
nicoche: > you can call a method with a pointer receiver even if the pointer is nil)

I was not aware of that, thanks for the tip! I just applied your suggestion, that solves everything: simpler and the confusion around the order of checks is gone

nomad/deploymentwatcher/deployment_watcher.go
nicoche: 🤦 Nice catch, thanks for that

nicoche:
Hello @tgross 

Thanks for the feedbacks, you caught a lot of blunders that I made 🤦

 >  especially given the note I left around the unlimited case

I applied your suggestion. I think that it mostly solves the issue.

> Do you have a Nomad job you're using as your smoke test that you could share?

I'll try the whole thing with this new version and get back to you with the specs that I'm using

nicoche:
Here are a few manifests that I tried which had the expected behavior:

<details>
  <summary>A job with a unlimited reschedules and a progress deadline will reschedule many times until hitting the progress deadline</summary>
  
```
job "broken-unlimited-reschedules" {
  type = "service"

  update {
    max_parallel      = 1
    auto_promote      = true
    canary            = 1
    progress_deadline = "180s"
    healthy_deadline = "179s"
  }

  reschedule {
    attempts       = 0
    interval       = 0
    delay          = "5s"
    delay_function = "constant"
    unlimited      = true
  }

  group "some-broken-group" {
    count = 1

    restart {
      delay = "1s"
      mode  = "fail"
    }

    task "some-broken-task" {
      driver = "docker"

      config {
        image = "redis:zzzzzzzzzz"
      }
    }
  }
}
```
  
</details>

<details>
  <summary>A job with limited reschedules and a progress deadline will be stopped after hitting the max number of reschedules</summary>
  
```
job "broken-finite-reschedules" {
  type = "service"

  update {
    max_parallel      = 1
    auto_promote      = true
    canary            = 1
    progress_deadline = "15m"
  }

  reschedule {
    attempts       = 1
    interval       = "24h"
    delay          = "20s"
    delay_function = "constant"
    unlimited      = false
  }

  group "some-broken-group" {
    count = 1

    restart {
      delay = "1s"
      mode  = "fail"
    }

    task "some-broken-task" {
      driver = "docker"

      config {
        image = "redis:qqqqqqqqqq"
      }
    }
  }
}
```
  
</details>

I just tested canaries the following way:
- I deployed a working redis job, everything OK
- I deployed a new version with `update.canary=1`, `update.progress_deadline=180s`, `reschedule.unlimited=true` and edit the image version to something that does not exist
- I validated that the deployment ended in Error and the job was still running the latest stable version

Then, I did the same but with `update.canary=1`, `update.progress_deadline=15m`, `reschedule.unlimited=false`, `reschedule.attempts=1` and validated the same thing, except that the deployment failed after running out  of reschedule attempts

nicoche:
Hey @tgross !

Is there any way I can help to move this forward? I hear that you might want to do more tests. Let me know if I can perform them on your behalf or add specific, automated tests to the PR

nicoche:
Hey @tgross ! Thanks for the thorough review and for pushing this through 🙂

PR #17342, opened 2023-05-29
core: fix kill_timeout validation when progress_deadline is 0

### LINKED ISSUE #17335: progress_deadline set to 0 fails to validate
Nomad 1.5.6 with #16761 introduces a check for kill_timeout, but fails to consider progress_deadline set to 0.

It fails with

```
Job validation errors:
1 error occurred:
	* Task group some-group validation failed: 1 error occurred:
	* Task some-task has a kill timout (5s) longer than the group's progress deadline (0s)
```

We don't have a kill timeout set, so 5s is the default.

As per documentation, progress_deadline set to 0 has special meaning:

> If the progress_deadline is set to 0, the first allocation to be marked as unhealthy causes the deployment to fail. 

We now have jobs that fails to validate/deploy. Is this a bug, or is progress_deadline set to 0 no longer supported?

lgfa29:
Fixes the validation for `kill_timeout` when `progress_deadline` is set to 0, which is a special case that should not be validated.

Closes #17335

### DIFF
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -6831,7 +6831,7 @@ func (tg *TaskGroup) Validate(j *Job) error {
 
 		// Validate the group's Update Strategy does not conflict with the Task's kill_timeout for service type jobs
 		if isTypeService && tg.Update != nil {
-			// progress_deadline = 0 has a special meaning show it should not
+			// progress_deadline = 0 has a special meaning so it should not be
 			// validated against the task's kill_timeout.
 			if tg.Update.ProgressDeadline > 0 && task.KillTimeout > tg.Update.ProgressDeadline {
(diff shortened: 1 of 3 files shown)


nomad/structs/structs.go
```diff
@@ -6831,12 +6831,13 @@ func (tg *TaskGroup) Validate(j *Job) error {
 
 		// Validate the group's Update Strategy does not conflict with the Task's kill_timeout for service type jobs
 		if isTypeService && tg.Update != nil {
-			if task.KillTimeout > tg.Update.ProgressDeadline {
+			// progress_deadline = 0 has a special meaning show it should not
```
jrasell, line 6834:
```suggestion
			// progress_deadline = 0 has a special meaning, so it should not be
```

jrasell (approved):
LGTM with a minor comment change.

PR #17343, opened 2023-05-29
Observe newlines when displaying variables

### LINKED ISSUE #17183: preserve newlines when showing Nomad variables in UI
### Proposal
Preserve the newlines when showing Nomad variables in UI.

How is it now:
image
image

How could it be:
image

### Use-cases
The use case is any variable that has newlines.

### Attempted Solutions
The attempted solution is to use openstyles chome extension and add a custom style with `code { white-space: pre-wrap; word-wrap: break-word; }`.

Thank you!

philrenaud:
Resolves #17183 

pre-wrap-formatting will honour newline characters and still attempt to wrap at the appropriate point (in case of a particularly long variable with spaces, etc.)

image
image

### DIFF
--- a/.changelog/17343.txt
+++ b/.changelog/17343.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: preserve newlines when displaying shown variables in non-json mode
+```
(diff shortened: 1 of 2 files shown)


lgfa29 (approved):
LGTM.

A changelog here would be nice as well 🙂

ChaiWithJai (approved):

PR #17344, opened 2023-05-29
Refactor task validation

lgfa29:
Move all validations related to task fields to `Task.Validate()`. Prior to this, some task validations were being done inside `TaskGroup.Validate()` because they required access to some group values.

But similarly to how `TaskGroup.Validate()` takes the job as parameter, it's fair to expect the task to receive its group.

Note: this could be considered an alternative implementation for #17342, but we could also just merge https://github.com/hashicorp/nomad/commit/a5f3e0d34a4765d3c095f9b3108f6d80ca7f0287, or nothing at all if we don't think this is worth it 😄

### DIFF
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -6822,36 +6822,11 @@ func (tg *TaskGroup) Validate(j *Job) error {
 		mErr.Errors = append(mErr.Errors, outer)
 	}
 
-	isTypeService := j.Type == JobTypeService
-
 	// Validate the tasks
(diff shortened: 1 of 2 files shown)


Juanadelacuesta (approved):

lgfa29:
After looking at it again I thought it would be best to merge and backport https://github.com/hashicorp/nomad/pull/17342 and only merge this refactoring to `main`. 

I think it's a fairly stable part of the code that is not expected to receive backports (famous last words 😅)

PR #17345, opened 2023-05-29
[ui] Text type to password type input on profile sign-in page

### LINKED ISSUE #16901: Security FR: ui/settings/tokens input field type as 'password' to guide autofill vs. password manager logic
# Problem:
- Currently the `ui/settings/tokens` input field `token-input` is flagged as a `type="text"` field and does not have a defined `autocomplete` value.
- For example Chrome thus treats it as a generic form field for which the values should be stored to provide autocompletion.
- This bypasses the browser level settings for storing sensitive tokens ("passwords") and instead the field is saved depending on the general form autofill setting.

## Repro example:
1. Create an ACL token with write policies attached.
2. Ensure browser does not save or prompt saving passwords for security.
3. Enter said token at `ui/settings/tokens`

#### Expected outcome:
- Token is not saved in browser store.
- I can use a password manager such as the built-in one of the Browser, or an external one such as Bitwarden to fill in the token.

#### Actual outcome:
- The sensitive token is now saved and available for autofill when clicking the field or typing the first characters.
- Bitwarden is unable to fill the token: `Unable to autofill the selected item on this page.`

# Solutions:
- Setting `type="password"` for the field allows handling it via the password manager APIs instead of as a form field autofill.
- This has further implications since then the token would not be visible as entering it, but you could just add a `show` icon to solve that.

philrenaud:
Resolves #16901

Changes ACL Token/JWT input field to be of type "password", enabling password managers and triggering browsers to store it sensitively.

I considered adding a "show token" button but opted against it for keeping overhead low.

### DIFF
--- a/.changelog/17345.txt
+++ b/.changelog/17345.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: change token input type from text to password
+```
--- a/ui/app/templates/settings/tokens.hbs
+++ b/ui/app/templates/settings/tokens.hbs
@@ -112,7 +112,7 @@
             <Input
               id="token-input"
               class="input"
-              @type="text"
+              @type="password"
               placeholder="{{if this.hasJWTAuthMethods "36-character token secret or JWT" "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX"}}"
               {{autofocus}}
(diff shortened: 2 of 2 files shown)


lgfa29 (approved):
LGTM. It may be worth a changelog entry since it's a user-facing change.

ChaiWithJai (approved):

Issue #17302, opened 2023-05-24
issue upgrading along the 1.3.x path - prevents nomad allocations


josh-m-sharpe:
Our stable versions:
consul: 1.13.2
nomad: 1.3.5

Decided to start upgrading, so I figured, sure, the last patch release of 1.13.x and 1.3.x respectively.  
Attempted version combination:
consul: 1.13.8
nomad: 1.3.14 (I started this process before 1.3.15 dropped)

When I launch nomad with these versions, I see this output in /var/log/messages:

```
nomad: 2023-05-24T16:50:05.821Z [WARN]  client.fingerprint_mgr.consul: unable to fingerprint consul: attribute=consul.grpc
nomad: 2023-05-24T16:50:05.821Z [WARN]  client.fingerprint_mgr.consul: unable to fingerprint consul: attribute=consul.sku
```

This seems to have the effect of preventing nomad from allocating containers there.  This only happened once, and it took out my site, so I'm a bit reluctant to "test" that part again.  I'm super interested in understanding and eliminating this warning though.  It all seems odd though, the consul agent joins the cluster and looks like it's healthy.  The nomad agent joins as well but just shows an empty client - nothing gets provisioned there.

I can't find much on the internet about what those attributes are and/or what nomad is looking for.  So I don't have much to go on.

By trial and error I've sorta determined that with consul 1.13.7 this issue doesn't show up, so by downgrading to:
Consul: 1.13.7
Nomad: 1.3.14 
...I sorta have a new stable set of versions.

I saw this release: https://github.com/hashicorp/nomad/releases/tag/v1.3.10 which mentions `consul: add client configuration for grpc_ca_file` so I attempted nomad 1.3.9 with consul 1.13.8 and that still produces these warnings.

Before I go try all the versions of nomad to figure out what's going on - is there anything I've missed.  I didn't see much in any of the release notes between Nomad 1.3.5 and 1.3.15.  Thanks!

erulabs:
We're seeing the same thing, occurred when we upgraded to 1.5.6 from 1.5.5 and consul to 1.13.8 from 1.13.7

I suspect the issue here is consul 1.13.8 instead of Nomad due to @josh-m-sharpe's versions - will test this shortly.

josh-m-sharpe:
[...]

PR #17349, opened 2023-05-29
client: fix Consul version finterprint

lgfa29:
Consul v1.13.8 was released with a [breaking change](https://github.com/hashicorp/consul/issues/17503) in the /v1/agent/self endpoint version where a line break was being returned.

This caused the Nomad finterprint to fail because `NewVersion` errors on parse.

This commit removes any extra space from the Consul version returned by the API.

Closes https://github.com/hashicorp/nomad/issues/17302


jrasell (approved):
LGTM!

shoenig (approved):

Issue #9451, opened 2020-11-25
client: make batching of allocation updates smarter


schmichael:
## Nomad version
`Nomad v1.0.0-beta3`

## Issue

Load testing revealed pathological behavior in the client loop that sends allocation updates to servers. See #9435 for details.

#9435 is the most naive improvement to client allocation updating logic. Further improvements could have better performance characteristics while lowering the latency for operations that depend on allocation updates (`nomad alloc status`, deployments, drains, and rescheduling).

### Split batching from sending

As of #9435 the alloc update batching and sending are all performed in a single goroutine. While the code is easy to understand and eases error+retry handling, it potentially causes unnecessary backpressure into alloc updates if the chan's buffer ever fills. Newer updates should *always* overwrite older updates, so there's no reason to keep a serialized buffer of updates and consume them one at a time.

A more efficient design would be to share the `update` map, and directly update it (using a mutex) instead of using a chan for serialization. This removes the potential for processing stale updates and minimizes the number of allocations in memory.

### Scale Alloc Update Rate like Heartbeats

The node heartbeat rate is proportional to cluster size: the larger the size, the long between updates.

Alloc updates should apply similar logic, although potentially with a small upper bound (1s? 3s?) to cap the staleness of alloc information which is used for other time sensitive operations (eg deployments and drains).

### Count alloc updates as heartbeats

Right now alloc updates and heartbeats are entirely separate code paths. It is possible, although has not been observed, that a node's alloc update succeeds while its heartbeat fails, resulting in a false `down` node.

If alloc updates were counted as heartbeats on both the clients and servers it could reduce a lot of spurious RPC calls and reduce opportunities for false `down` detection.

### Prioritized updates

Not all alloc updates are equal: task events are purely for human consumption whereas fields like `Allocation.ClientStatus` affects a wide range of scheduler behavior (deployments, drains, rescheduling, and more).

Therefore we could update the batch interval based on the highest priority update in the current batch. The optimal behavior being something like:

1. A couple docker allocs are started concurrently on a client
2. 100ms later `Received` task event is created for both (2 updates)
3. 100ms later `Task dir created` event is created for both (2 updates)
4. 100ms later `Image downloading` event is created for both (2 updates)
5. 10s later 1 alloc is started (1 update)
6. 100ms later the other alloc fails (1 update)

(For realistic jobs with artifacts and templates and deployments there would be a lot more updates.)

[...]

PR #17354, opened 2023-05-30
prioritized client updates

tgross:
The allocrunner sends several updates to the server during the early lifecycle of an allocation and its tasks. Clients batch-up allocation updates every 200ms, but experiments like the [C2M challenge](https://www.hashicorp.com/blog/hashicorp-nomad-meets-the-2-million-container-challenge) have shown that even with this batching, servers can be overwhelmed with client updates during high volume deployments. Benchmarking done in #9451 has shown that client updates can easily represent ~70% of all Nomad Raft traffic.

Each allocation sends many updates during its lifetime, but only those that change the `ClientStatus` field are critical for progressing a deployment or kicking off a reschedule to recover from failures.

Add a priority to the client allocation sync and update the `syncTicker` receiver so that we only send an update if there's a high priority update waiting, or on every 5th tick. This means when there are no high priority updates, the client will send updates at most every 1s instead of 200ms. Benchmarks have shown this can reduce overall Raft traffic by 10%, as well as reduce client-to-server RPC traffic.

This changeset also switches from a channel-based collection of updates to a shared buffer, so as to split batching from sending and prevent backpressure onto the allocrunner when the RPC is slow. This doesn't have a major performance benefit in the benchmarks but makes the implementation of the prioritized update simpler.

Fixes: #9451

### DIFF
--- a/client/allocrunner/alloc_runner.go
+++ b/client/allocrunner/alloc_runner.go
@@ -1420,30 +1420,36 @@ func (ar *allocRunner) persistLastAcknowledgedState(a *state.State) {
 	}
 }
 
-// LastAcknowledgedStateIsCurrent returns true if the current state matches the
-// state that was last acknowledged from a server update. This is called from
-// the client in the same goroutine that called AcknowledgeState so that we
-// can't get a TOCTOU error.
-func (ar *allocRunner) LastAcknowledgedStateIsCurrent(a *structs.Allocation) bool {
+// GetUpdatePriority returns the update priority based the difference between
+// the current state and the state that was last acknowledged from a server
+// update. This is called from the client in the same goroutine that called
+// AcknowledgeState so that we can't get a TOCTOU error.
+func (ar *allocRunner) GetUpdatePriority(a *structs.Allocation) cstructs.AllocUpdatePriority {
 	ar.stateLock.RLock()
 	defer ar.stateLock.RUnlock()
 
 	last := ar.lastAcknowledgedState
 	if last == nil {
-		return false
+		return cstructs.AllocUpdatePriorityTypical
 	}
 
 	switch {
 	case last.ClientStatus != a.ClientStatus:
-		return false
+		return cstructs.AllocUpdatePriorityUrgent
 	case last.ClientDescription != a.ClientDescription:
-		return false
+		return cstructs.AllocUpdatePriorityTypical
 	case !last.DeploymentStatus.Equal(a.DeploymentStatus):
-		return false
+		return cstructs.AllocUpdatePriorityTypical
 	case !last.NetworkStatus.Equal(a.NetworkStatus):
-		return false
+		return cstructs.AllocUpdatePriorityTypical
 	}
-	return maps.EqualFunc(last.TaskStates, a.TaskStates, func(st, o *structs.TaskState) bool {
+
+	if !maps.EqualFunc(last.TaskStates, a.TaskStates, func(st, o *structs.TaskState) bool {
 		return st.Equal(o)
-	})
+
+	}) {
+		return cstructs.AllocUpdatePriorityTypical
+	}
+
+	return cstructs.AllocUpdatePriorityNone
 }
--- a/client/client.go
+++ b/client/client.go
@@ -43,6 +43,7 @@ import (
 	"github.com/hashicorp/nomad/client/serviceregistration/wrapper"
 	"github.com/hashicorp/nomad/client/state"
 	"github.com/hashicorp/nomad/client/stats"
+	cstructs "github.com/hashicorp/nomad/client/structs"
 	"github.com/hashicorp/nomad/client/vaultclient"
 	"github.com/hashicorp/nomad/command/agent/consul"
 	"github.com/hashicorp/nomad/helper"
@@ -213,8 +214,10 @@ type Client struct {
 	invalidAllocs     map[string]struct{}
 	invalidAllocsLock sync.Mutex
 
-	// allocUpdates stores allocations that need to be synced to the server.
-	allocUpdates chan *structs.Allocation
+	// allocUpdates stores allocations that need to be synced to the server, and
+	// allocUpdatesLock guards access to it for concurrent updates.
+	allocUpdates     map[string]*structs.Allocation
+	allocUpdatesLock sync.Mutex
 
 	// consulService is the Consul handler implementation for managing services
 	// and checks.
@@ -366,7 +369,7 @@ func NewClient(cfg *config.Config, consulCatalog consul.CatalogAPI, consulProxie
 		logger:               logger,
 		rpcLogger:            logger.Named("rpc"),
 		allocs:               make(map[string]interfaces.AllocRunner),
-		allocUpdates:         make(chan *structs.Allocation, 64),
+		allocUpdates:         make(map[string]*structs.Allocation, 64),
 		shutdownCh:           make(chan struct{}),
 		triggerDiscoveryCh:   make(chan struct{}),
 		triggerNodeUpdate:    make(chan struct{}, 8),
@@ -1322,10 +1325,11 @@ func (c *Client) handleInvalidAllocs(alloc *structs.Allocation, err error) {
 
 	// Mark alloc as failed so server can handle this
 	failed := makeFailedAlloc(alloc, err)
-	select {
-	case c.allocUpdates <- failed:
-	case <-c.shutdownCh:
-	}
+
+	c.allocUpdatesLock.Lock()
+	defer c.allocUpdatesLock.Unlock()
+
+	c.allocUpdates[alloc.ID] = failed
 }
 
 // saveState is used to snapshot our state into the data dir.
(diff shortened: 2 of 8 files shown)


tgross:
Moving this to draft because as I was re-reading it I realized there's another minor optimization we can make here that should make the implementation simpler in the allocrunner as well.

client/client_test.go
```diff
@@ -765,6 +765,24 @@ func TestClient_SaveRestoreState(t *testing.T) {
 						return fmt.Errorf("expected running client status, got %v",
 							ar.AllocState().ClientStatus)
 					}
+
```
tgross, line 768: Note for reviewers: this test was actually already racy but it's a _very_ tight race before the changes in this PR. So these fixes for the test remove the race.

jrasell (approved):
LGTM! I added a couple of very minor comments and a question, but nothing blocking.

client/client.go
```diff
@@ -2174,33 +2179,76 @@ func (c *Client) allocSync() {
...
+
+	// Clear here so that allocrunners can queue up the next set of updates
+	// while we're waiting to hear from the server
+	c.allocUpdates = make(map[string]*structs.Allocation, len(c.allocUpdates))
+
+	return toSync
+}
+
+// filteredAcknowledgedUpdates returns a list of client alloc updates with the
+// already-acknowledged updates removed, and the highest priority of any update.
```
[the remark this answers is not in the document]
shoenig: I'd go a step further and suggest extracting the `allocUpdates` map into its own little data structure and let it handle its own locking. There's quite a bit of implementation detail around what it's used for splattered all over these Client functions.

client/client.go
```diff
@@ -2174,33 +2179,76 @@ func (c *Client) allocSync() {
...
+	toSync, urgent := c.filterAcknowledgedUpdates(c.allocUpdates)
+
+	// Only update every 5th tick if there's no priority updates
+	if updateTicks%5 != 0 && !urgent {
+		return nil
+	}
+
+	// Clear here so that allocrunners can queue up the next set of updates
+	// while we're waiting to hear from the server
+	c.allocUpdates = make(map[string]*structs.Allocation, len(c.allocUpdates))
```
shoenig, line 2222: https://pkg.go.dev/golang.org/x/exp/maps#Clear

will produce less work for the GC

client/client.go
```diff
@@ -2099,10 +2103,10 @@ func (c *Client) AllocStateUpdated(alloc *structs.Allocation) {
...
 	stripped.NetworkStatus = alloc.NetworkStatus
 
-	select {
-	case c.allocUpdates <- stripped:
-	case <-c.shutdownCh:
-	}
+	c.allocUpdatesLock.Lock()
+	defer c.allocUpdatesLock.Unlock()
+
+	c.allocUpdates[stripped.ID] = stripped
```
shoenig, line 2109: I think this was always true, but we can overwrite previous pending alloc updates for a given alloc, even if both updates were urgent priority? 

The `updates[alloc.ID] = alloc` would be the previous behavior. Just want to make sure.

client/client.go
```diff
@@ -2174,33 +2179,76 @@ func (c *Client) allocSync() {
...
+				urgent = true
+			case cstructs.AllocUpdatePriorityTypical:
 				sync = append(sync, update)
+			case cstructs.AllocUpdatePriorityNone:
+				// update is dropped
 			}
 		} else {
 			// no allocrunner (typically a failed placement), so we need
 			// to send update
 			sync = append(sync, update)
```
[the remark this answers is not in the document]
tgross: Exactly!

client/client.go
```diff
@@ -2174,33 +2179,76 @@ func (c *Client) allocSync() {
...
+	c.allocUpdates = make(map[string]*structs.Allocation, len(c.allocUpdates))
+
+	return toSync
+}
+
+// filteredAcknowledgedUpdates returns a list of client alloc updates with the
+// already-acknowledged updates removed, and the highest priority of any update.
+func (c *Client) filterAcknowledgedUpdates(updates map[string]*structs.Allocation) ([]*structs.Allocation, bool) {
+	var urgent bool
 	sync := make([]*structs.Allocation, 0, len(updates))
```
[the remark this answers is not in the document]
tgross: Good call, I changed it to `toSync` elsewhere for that very reason. Will fix.

client/client.go
tgross: Oh, cool. IIRC that's coming to the stdlib soon too?

shoenig: In go1.21 https://go-review.googlesource.com/c/go/+/464343

client/client.go
tgross: Right. The updates we get from the allocrunner are lightweight copies of its `*structs.Allocation` (i.e. just the client state), serialized with an internal lock. So whatever the latest update we've got here is the currently authoritative one. And the server doesn't keep versions of `*structs.Allocation` so it also only cares about the most recent version from the client.

client/client.go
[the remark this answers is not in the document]
tgross: > If I understand correctly, the caller must hold at least a read lock on `c.allocUpdatesLock` when calling this function? If so, would it be worth adding a note to the function comment about this for future readers?

Well it's not a `RWMutex`, but yes. Really the only reason this is in its own function at all now is that we need to take the `c.allocLock` to read from the allocrunners and having it in its own function lets us scope-down that lock.

> I'd go a step further and suggest extracting the `allocUpdates` map into its own little data structure and let it handle its own locking. There's quite a bit of implementation detail around what it's used for splattered all over these Client functions.

So in other words, make `updatesToSync` and `filterAcknowledgedUpdates` methods on the `allocUpdates` data structure? We'd need to pass the `*Client` as a parameter so that we can access the allocrunners. But yeah that seems reasonable. Let me take a quick try at that.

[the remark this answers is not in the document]
tgross: @shoenig @jrasell I've refactored this by pulling out the `allocUpdates` to their own struct `pendingAllocUpdates` (so as not to conflict with the other struct named `allocUpdates` which is the updates received from the server!). I think this makes the whole locking situation a lot cleaner. Let me know what you think.

client/client.go
```diff
@@ -3306,3 +3274,102 @@ func (g *group) AddCh(ch <-chan struct{}) {
...
+
+// filteredAcknowledgedUpdatesLocked returns a list of client alloc updates with the
+// already-acknowledged updates removed, and the highest priority of any update.
+func (p *pendingClientUpdates) filterAcknowledgedUpdatesLocked(c *Client) ([]*structs.Allocation, bool) {
+	var urgent bool
+	sync := make([]*structs.Allocation, 0, len(p.updates))
+	c.allocLock.RLock()
+	defer c.allocLock.RUnlock()
+
+	for allocID, update := range p.updates {
```
shoenig, line 3357: I think this also needs to hold `p.lock`? Since this also grabs `c.alloc`, be super careful to grab and release in the ABBA order so it can't deadlock

shoenig (approved):
LGTM; just the one locking thing

client/client.go
tgross: The only caller holds `p.lock`, which is why this has the suffix "Locked" on its name. I can add an extra note about that.

client/allocrunner/alloc_runner.go
```diff
@@ -1420,30 +1420,36 @@ func (ar *allocRunner) persistLastAcknowledgedState(a *state.State) {
...
 	switch {
 	case last.ClientStatus != a.ClientStatus:
-		return false
+		return cstructs.AllocUpdatePriorityUrgent
 	case last.ClientDescription != a.ClientDescription:
-		return false
+		return cstructs.AllocUpdatePriorityTypical
 	case !last.DeploymentStatus.Equal(a.DeploymentStatus):
-		return false
+		return cstructs.AllocUpdatePriorityTypical
```
schmichael, line 1442: Technically deployments are gated by this field, so it could considered critical since it can cause a scheduling decision...

...but nothing about deployments is concerned with sub-second latencies, so I think it's fine to leave this as `Typical`.

If you're in this code again maybe add a comment pointing out that while deployment status changes are not urgent, they _can_ affect scheduling but not in a way that sub-second skew is significant.

If you really want to tidy things up the PR description misses this too:

> Each allocation sends many updates during its lifetime, but only those that change the `ClientStatus` field are critical for progressing a deployment or kicking off a reschedule to recover from failures.

`DeploymentStatus` is critical for progressing a deployment as well.

client/allocrunner/alloc_runner.go
```diff
@@ -1420,30 +1420,36 @@ func (ar *allocRunner) persistLastAcknowledgedState(a *state.State) {
...
+// update. This is called from the client in the same goroutine that called
+// AcknowledgeState so that we can't get a TOCTOU error.
+func (ar *allocRunner) GetUpdatePriority(a *structs.Allocation) cstructs.AllocUpdatePriority {
 	ar.stateLock.RLock()
 	defer ar.stateLock.RUnlock()
 
 	last := ar.lastAcknowledgedState
 	if last == nil {
-		return false
+		return cstructs.AllocUpdatePriorityTypical
```
schmichael, line 1433: If we don't know what it was before, how can we assume the change is typical? Seems worth a comment especially since all of the other code in this method *must* check from Highest Priority to Lowest in order to ensure a change to a low priority field doesn't demote an actually high priority update.

tgross: You're right that we don't know for sure. In practice an allocation will never become healthy quickly enough that the first update we send is that update. That being said we probably should account for allocations that quickly _fail_ because there's a bunch of things that can go unrecoverably wrong on the client before we ever hit the task runner, and it'd be nice to be able to send those failure states to the server more quickly.

client/allocrunner/alloc_runner.go
tgross: > Technically deployments are gated by this field, so it could considered critical since it can cause a scheduling decision...
>
> ...but nothing about deployments is concerned with sub-second latencies, so I think it's fine to leave this as Typical

Somehow I missed that, so yeah I would've set it to urgent based on the reasoning I had in the PR. I'll keep (for now at least) and I'll add some commentary here around reasoning for things.

PR #17365, opened 2023-05-30
np: implement ACL for node pools

lgfa29:
Implement ACL concepts for node pool management.

Website preview documenting the details:
https://nomad-git-f-node-pools-acl-hashicorp.vercel.app/nomad/docs/other-specifications/acl-policy#node-pools-rules

### DIFF
--- a/acl/policy.go
+++ b/acl/policy.go
@@ -56,6 +56,25 @@ var (
 )
 
 const (
+	// The following are the fine-grained capabilities that can be granted for
+	// node volume management.
+	//
+	// The Policy block is a short hand for granting several of these. When
+	// capabilities are combined we take the union of all capabilities. If the
+	// deny capability is present, it takes precedence and overwrites all other
+	// capabilities.
+
+	NodePoolCapabilityDelete = "delete"
+	NodePoolCapabilityDeny   = "deny"
+	NodePoolCapabilityRead   = "read"
+	NodePoolCapabilityWrite  = "write"
+)
+
+var (
+	validNodePool = regexp.MustCompile("^[a-zA-Z0-9-_*]{1,128}$")
+)
+
+const (
 	// The following are the fine-grained capabilities that can be granted for a volume set.
 	// The Policy block is a short hand for granting several of these. When capabilities are
 	// combined we take the union of all capabilities. If the deny capability is present, it
@@ -84,6 +103,7 @@ const (
 // Policy represents a parsed HCL or JSON policy.
 type Policy struct {
 	Namespaces  []*NamespacePolicy  `hcl:"namespace,expand"`
+	NodePools   []*NodePoolPolicy   `hcl:"node_pool,expand"`
 	HostVolumes []*HostVolumePolicy `hcl:"host_volume,expand"`
 	Agent       *AgentPolicy        `hcl:"agent"`
 	Node        *NodePolicy         `hcl:"node"`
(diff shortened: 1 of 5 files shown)


website/content/docs/other-specifications/acl-policy.mdx
```diff
@@ -248,6 +248,62 @@ The `policy` field for the node rule can have one of the following values:
...
+## Node Pools rules
+
+Node pool rules are defined with a `node_pool` block. An ACL policy can
+include zero, one, or more node pool rules.
+
+<!-- TODO(luiz): add link to node pools API docs -->
+Node pool rule controls access to the Node Pool API such as create, update, and
+list node pools.
+
+Each node pool rule is labeled with the node pool name it applies to. You may
```
tgross, line 260: Is omitting the label allowed so that it falls back to `"default"`, as we do for namespaces?

acl/policy.go
```diff
@@ -55,6 +55,25 @@ var (
 	validNamespace = regexp.MustCompile("^[a-zA-Z0-9-*]{1,128}$")
 )
 
+const (
+	// The following are the fine-grained capabilities that can be granted for
+	// node volume management.
+	//
+	// The Policy block is a short hand for granting several of these. When
```
tgross, line 62:
```suggestion
	// The Policy field is a short hand for granting several of these. When
```

acl/policy.go
```diff
@@ -247,6 +275,33 @@ func expandNamespacePolicy(policy string) []string {
 	}
 }
 
+func isNodePoolCapabilityValidd(cap string) bool {
```
tgross, line 278:
```suggestion
func isNodePoolCapabilityValid(cap string) bool {
```

acl/policy.go
```diff
@@ -352,6 +407,25 @@ func Parse(rules string) (*Policy, error) {
...
 
+	for _, np := range p.NodePools {
+		if !validNodePool.MatchString(np.Name) {
+			return nil, fmt.Errorf("Invalid node pool name: %#v", np)
+		}
+		if np.Policy != "" && !isPolicyValid(np.Policy) {
+			return nil, fmt.Errorf("Invalid node pool policy: %#v", np)
+		}
+		for _, cap := range np.Capabilities {
+			if !isNodePoolCapabilityValidd(cap) {
```
tgross, line 418:
```suggestion
			if !isNodePoolCapabilityValid(cap) {
```

acl/policy.go
```diff
@@ -352,6 +407,25 @@ func Parse(rules string) (*Policy, error) {
 
 	}
 
+	for _, np := range p.NodePools {
+		if !validNodePool.MatchString(np.Name) {
+			return nil, fmt.Errorf("Invalid node pool name: %#v", np)
```
tgross, line 412: Won't this result in a go struct representation being written to logs? The structs aren't usually user-facing and aren't really 1:1 with the HCL. I think we probably want to report the specific name that's invalid here. Similar for the "invalid node pool policy" error below -- let's tell the user why it's invalid (in this case that it has an invalid capability)

tgross (approved):
LGTM! I've left a few nitpicks and suggestions but nothing major blocking.

acl/policy.go
lgfa29: Ah that's right, it was a copy-pasta from other rule types. We're not very consistent how this is reported so I will create another format for now 😅

acl/policy.go
lgfa29: Fixed here and in the other places as well.

website/content/docs/other-specifications/acl-policy.mdx
lgfa29: After some internal discussions we're going to leave this one like this for now.

PR #17368, opened 2023-05-31
ui:  node pools data on `jobs.index` page

### LINKED ISSUE #17287: Jobs list pages will receive a new dropdown and table column to allow users to filter and sort entries by node pool.

ChaiWithJai:
Resolves #17287 

This PR adds a filter and table column for Node Pools to the `jobs.index` page.

Side Effect:  to do this, we must update `JobRow` component which is also used in the `children` template and will appear on the Job Detail Overview page for parent jobs.


ui/app/controllers/jobs/index.js
```diff
@@ -194,6 +199,29 @@ export default class IndexController extends Controller.extend(
...
+      this.set(
+        'qpNodePool',
+        serialize(
+          intersection(
+            availableNodePools.map(({ name }) => name),
+            this.selectionNodePools
+          )
+        )
+      );
+    });
```
philrenaud, line 217: Would this be better off run from a method elsewhere, or in a constructor?

philrenaud (approved):
Looks good; just had one question about the side-effect within the optionsNodePools() computed prop.

The thing with `<JobRow>` meaning you have to add this to the child jobs table feels pretty bad IMO, as those will always necessarily have the same node pool, right? Good argument against abstracting things like our job rows into reusable components in the future.

ui/app/controllers/jobs/index.js
ChaiWithJai: A `constructor` isn't the appropriate place to run this callback function because it runs whenever the controller is instantiated and there is a change to the dependent keys listed. Therefore it is best to keep the functionality here. I don't think its worth engineering effort until we upgrade to a `tracked` paradigm for the controller.

PR #17377, opened 2023-06-01
repo: block pushing to release branches in git hook

shoenig:


tgross (approved):
LGTM!

tgross:
Yes but you can use `git push --no-verify` in that case, which bypasses the hook for that one case.

shoenig:
Oh dangit, I can update the checklist to include `--no-verify`

PR #17378, opened 2023-06-01
Create a lint workflow for copyright checking

dekimsey:
This change introduces a lint workflow that verifies files have the required copyright headers. This allows us to be more proactive regarding copyright changes _before_ the bot comes around and asks for changes.

I've created this as a generic workflow "Lint" so that we have a place to possibly surface `make check` tests to our contributors.

Discovered during #17103 

Since we are adding a lint to an existing repository, I've gone ahead and applied the fix to failing files. Attached is an example of the copyright lint action failing:
image

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,5 +15,7 @@ jobs:
     steps:
       - name: Setup Copywrite
         uses: hashicorp/setup-copywrite@867a1a2a064a0626db322392806428f7dc59cb3e # v1.1.2
+      - name: Checkout
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
       - name: Check Header Compliance
         run: copywrite headers --plan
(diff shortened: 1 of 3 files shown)


lgfa29:
Thanks @dekimsey! Detecting these missing headers early is a good improvement 😄 

Do you know which version of the CLI is used and if it would it include this fix? https://github.com/hashicorp/copywrite/pull/59

.github/workflows/lint.yml
```diff
@@ -0,0 +1,21 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  copyright:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Copywrite
+        uses: hashicorp/setup-copywrite@867a1a2a064a0626db322392806428f7dc59cb3e # v1.1.2
```
tgross, line 17: I thought we didn't need to pin `hashicorp/*` namespaced actions? It would be nice to be able to automatically pick up updates to the action without having to deal with a dependabot update

PR #17379, opened 2023-06-01
node pools: add `node_pool` field to job spec

tgross:
This changeset only adds the `node_pool` field to the jobspec, and ensures that it gets picked up correctly as a change. Without the rest of the implementation landed yet, the field will be ignored. Includes required updates to the docs and `nomad job status` command.


nomad/structs/structs.go
```diff
@@ -163,6 +163,9 @@ const (
 	DefaultNamespace            = "default"
 	DefaultNamespaceDescription = "Default shared namespace"
 
+	// DefaultNamespace is the default node pool.
+	DefaultNodePool = "default"
```
lgfa29, line 167: I have this in `structs_node_pool.go` but with the name flipped 😅 
https://github.com/hashicorp/nomad/blob/45fcc59a82137bf69cb1f6087e5ce614c378cd34/nomad/structs/node_pool.go#L20-L23

website/content/api-docs/json-jobs.mdx
```diff
@@ -256,6 +257,8 @@ The `Job` object supports the following keys:
 - `Datacenters` - A list of datacenters in the region which are eligible
   for task placement. This must be provided, and does not have a default.
 
+- `NodePool` - The node pool in which the job can be placed. Defaults to "default".
```
lgfa29, line 260:
```suggestion
- `NodePool` - The node pool in which the job can be placed. Defaults to `"default"`.
```

lgfa29 (approved):
I have those constants somewhere in my commit log, but I can clean them up once I get to them.

nomad/structs/structs.go
tgross: Hrm, we have it as `DefaultNodePool` in `api` in your [private branch](https://github.com/hashicorp/nomad-enterprise/blob/f-node-pools-nodes-api/api/node_pools.go#L8) which is why I did that here. It also matches `DefaultNamespace`. We should probably be consistent between `api` and `nomad/structs` just so we don't frustrate ourselves. Which would you prefer?

tgross: Eh, because there's also a `NodePoolAll` it makes for nicer autocomplete to have it be `NodePoolDefault`. Will fix.

lgfa29: Ah sorry, I didn't notice the inconsistency. Yeah, I think `NodePoolDefault` and `NodePoolAll` work better. I reconciled their use in https://github.com/hashicorp/nomad/pull/17388.

PR #17380, opened 2023-06-01
workflows: delete check-legacy-links-format.yml

BRKalow:
This workflow has been replaced by our content checks and is no longer needed.


tgross (approved):
LGTM! Thanks @BRKalow!

Issue #7376, opened 2020-03-18
API/command to initiate graceful change in leadership


ketzacoatl:
Hello,

Maybe I am misunderstanding the internals, so please tell me this makes no sense if that is the case. Or if there is an existing thing I should instead that would accomplish the same. thank you :)

Let's say you have 5 servers running in production, 1 of them is the active leader. You need to roll out an update to those servers. There are consistent scheduling events and changes going onto the log during the maintenance, so the update must be as graceful as possible. You iterate on the list of servers, taking one down at a time, bringing up a replacement, and waiting for that replacement to join the cluster and for quorum numbers to balance with those changes.

One method would pull each server indiscriminately, such that at any point, you could be taking down the active leader. While one of the other servers will pick up and take over, depending on load, there are likely to be significant delays and failures associated with the transition (for new scheduling events, not running allocations). This method is also more than likely to result in multiple transitions, and does not do anything to limit that to one transition. When this transition happens, the operator has no idea what scheduling event they may be interfering with.

Another method has the operator identify the active leader ahead of time, rolling all other nodes first. If the active leader changes, it may end up on a new node, or an old one. Either way, we'd leave an active leader on an old node until it is the last to roll.

In either case, it would be ideal if the operator could tell nomad: "I'd like to take the active leader offline for maintenance, can you please transition the active leader to a different server at some point in the next few operations, kthxbye".

The active leader would then not be "ripped out" while doing its work, more than is necessary (eg some other hardware failure or whatever). The cluster would have an opportunity to put it's work on hold for a moment, and transition gracefully without losing active requests.

ketzacoatl:
@notnoop / @tgross / @jrasell / @lgfa29 / @shoenig 

Later this month, we'll have sat with this issue for a year.

Just yesterday I had to decommission an instance on AWS (AWS would kill the instance for me if I didn't), and it just happened to be the active leader running in production (YAY!). I could have really used this functionality (or a reasonable work-around) to carry out that maintenance.

Overall, the transition to a new leader happened and the cluster stayed healthy, but allocations were missed and the transition was not fully graceful. Part of the problem was that, as an operator, I didn't have a clear/obvious path for making the transition. If there were an obvious-to-use command for this situation, the operator could complete the transition with confidence and without taking down the cluster (worst case, but still a risk).

[...]

PR #2305, opened 2017-02-10
Add nomad operator command for interacting with Raft configuration

dadgar:
This PR adds:

* `nomad operator raft list-peers` and `nomad operator raft remove-peer` subcommands
* Corresponding endpoints
* A guides section to our website
* Moves bootstrapping to guides from docs
* Adds an outage recovery guide.

@sethvargo For guidance on redirects from the old bootstrap page

PR #17383, opened 2023-06-01
[server] Directed leadership transfer CLI and API

angrycub:
This PR adds an API endpoint and a CLI command to allow cluster operators to move leadership from the current leader to a specified server node. This can reduce leadership churn during rolling updates by allowing the operator to migrate leadership to the first upgraded node.

**Commits**
- Add directed leadership transfer func
- Add leadership transfer RPC endpoint
- Add ACL tests for leadership-transfer endpoint
- Add HTTP API route and implementation
- Add to Go API client
- Implement CLI command
- Add documentation
    - [🔍 cli docs](https://nomad-qglbkcast-hashicorp.vercel.app/nomad/docs/commands/operator/raft/transfer-leadership)
    - [🔍 api docs](https://nomad-qglbkcast-hashicorp.vercel.app/nomad/api-docs/operator/raft#transfer-leadership-to-another-raft-peer)

Fixes: https://github.com/hashicorp/nomad/issues/7376

### DIFF
--- a/api/operator.go
+++ b/api/operator.go
@@ -120,6 +120,46 @@ func (op *Operator) RaftRemovePeerByID(id string, q *WriteOptions) error {
 	return nil
 }
 
+// RaftTransferLeadershipByAddress is used to transfer leadership to a
+// different peer using its address in the form of "IP:port".
+func (op *Operator) RaftTransferLeadershipByAddress(address string, q *WriteOptions) error {
+	r, err := op.c.newRequest("PUT", "/v1/operator/raft/transfer-leadership")
+	if err != nil {
+		return err
+	}
+	r.setWriteOptions(q)
+
+	r.params.Set("address", address)
+
+	_, resp, err := requireOK(op.c.doRequest(r))
+	if err != nil {
+		return err
+	}
+
+	resp.Body.Close()
+	return nil
+}
+
+// RaftTransferLeadershipByID is used to transfer leadership to a
+// different peer using its Raft ID.
+func (op *Operator) RaftTransferLeadershipByID(id string, q *WriteOptions) error {
+	r, err := op.c.newRequest("PUT", "/v1/operator/raft/transfer-leadership")
+	if err != nil {
+		return err
+	}
+	r.setWriteOptions(q)
+
+	r.params.Set("id", id)
+
+	_, resp, err := requireOK(op.c.doRequest(r))
+	if err != nil {
+		return err
+	}
+
+	resp.Body.Close()
+	return nil
+}
+
 // SchedulerConfiguration is the config for controlling scheduler behavior
 type SchedulerConfiguration struct {
 	// SchedulerAlgorithm lets you select between available scheduling algorithms.
@@ -362,4 +402,13 @@ func (op *Operator) LicenseGet(q *QueryOptions) (*LicenseReply, *QueryMeta, erro
 	qm.RequestTime = rtt
 
 	return &reply, qm, nil
+}
+
+type LeadershipTransferResponse struct {
+	From RaftServer
+	To   RaftServer
+	Noop bool
+	Err  error
+
+	QueryMeta
 }
--- a/api/operator.go
+++ b/api/operator.go
@@ -410,5 +410,5 @@ type LeadershipTransferResponse struct {
 	Noop bool
 	Err  error
 
-	QueryMeta
+	WriteMeta
 }
--- a/command/operator_raft_leadership_transfer.go
+++ b/command/operator_raft_leadership_transfer.go
@@ -36,10 +36,10 @@ General Options:
 Remove Peer Options:
 
   -peer-address="IP:port"
-	Transfer leadership to the Nomad server with given Raft address.
+    Transfer leadership to the Nomad server with given Raft address.
 
   -peer-id="id"
-	Transfer leadership to the Nomad server with given Raft ID.
+    Transfer leadership to the Nomad server with given Raft ID.
 `
 	return strings.TrimSpace(helpText)
 }
--- a/command/operator_raft_leadership_transfer.go
+++ b/command/operator_raft_leadership_transfer.go
@@ -21,11 +21,13 @@ Usage: nomad operator raft transfer-leadership [options]
 
   Transfer leadership to the Nomad server with given -peer-address or
   -peer-id in the Raft configuration. All server nodes in the cluster
-	must be running at least Raft protocol v3 in order to use this command.
+  must be running at least Raft protocol v3 in order to use this command.
 
   There are cases where you might desire transferring leadership from one
   cluster member to another, for example, during a rolling upgrade. This
   command allows you to designate a new server to be cluster leader.
+
+  Note: This command requires a currently established leader to function.
 
   If ACLs are enabled, this command requires a management token.
 
@@ -41,6 +43,7 @@ Remove Peer Options:
   -peer-id="id"
     Transfer leadership to the Nomad server with given Raft ID.
 `
+
 	return strings.TrimSpace(helpText)
 }
 
--- a/command/operator_raft_leadership_transfer.go
+++ b/command/operator_raft_leadership_transfer.go
@@ -1,0 +1,122 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package command
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/nomad/api"
+	"github.com/posener/complete"
+)
+
+type OperatorRaftTransferLeadershipCommand struct {
+	Meta
+}
+
+func (c *OperatorRaftTransferLeadershipCommand) Help() string {
+	helpText := `
+Usage: nomad operator raft transfer-leadership [options]
+
+  Transfer leadership to the Nomad server with given -peer-address or
+  -peer-id in the Raft configuration. All server nodes in the cluster
+	must be running at least Raft protocol v3 in order to use this command.
+
+  There are cases where you might desire transferring leadership from one
+	cluster member to another, for example, during a rolling upgrade. This
+	command allows you to designate a new server to be cluster leader.
+
+  If ACLs are enabled, this command requires a management token.
+
+General Options:
+
+  ` + generalOptionsUsage(usageOptsDefault|usageOptsNoNamespace) + `
+
+Remove Peer Options:
+
+  -peer-address="IP:port"
+	Transfer leadership to the Nomad server with given Raft address.
+
+  -peer-id="id"
+	Transfer leadership to the Nomad server with given Raft ID.
+`
+	return strings.TrimSpace(helpText)
+}
+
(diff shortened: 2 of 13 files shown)


api/operator.go
```diff
@@ -363,3 +403,12 @@ func (op *Operator) LicenseGet(q *QueryOptions) (*LicenseReply, *QueryMeta, erro
...
 }
+
+type LeadershipTransferResponse struct {
+	From RaftServer
+	To   RaftServer
+	Noop bool
+	Err  error
+
+	QueryMeta
+}
```
tgross, line 414: I really like this object but it doesn't look like we use it?

command/operator_raft_leadership_transfer.go
```diff
@@ -0,0 +1,122 @@
...
+
+  Transfer leadership to the Nomad server with given -peer-address or
+  -peer-id in the Raft configuration. All server nodes in the cluster
+	must be running at least Raft protocol v3 in order to use this command.
+
+  There are cases where you might desire transferring leadership from one
+	cluster member to another, for example, during a rolling upgrade. This
+	command allows you to designate a new server to be cluster leader.
+
+  If ACLs are enabled, this command requires a management token.
```
tgross, line 30: Looks like there's some weird indentation issues here? Might just be GitHub being GitHub though

nomad/operator_endpoint.go
```diff
@@ -228,6 +228,134 @@ REMOVE:
...
+	}
+
+TRANSFER:
+	log := op.logger.With("peer_id", id, "peer_addr", addr)
+	if err = op.srv.leadershipTransferToServer(id, addr); err != nil {
+		log.Error("failed transferring leadership", "error", err)
+		return err
+	}
+
+	log.Info("transferred leadership")
```
tgross, line 355: Might be nice to have the `from` and `to` in here if we can. It'll make for good post-incident forensics. Alternately, we have the `log.Info` already in `leadershipTransferToServer` so it's redundant here and we could have `from` and `to` fields there.

nomad/operator_endpoint.go
```diff
@@ -228,6 +228,134 @@ REMOVE:
...
+	}
+
+	// TransferLeadership is not supported until Raft protocol v3 or greater.
+	if minRaftProtocol < 3 {
+		op.logger.Warn("unsupported minimum common raft protocol version", "required", "3", "current", minRaftProtocol)
+		return fmt.Errorf("unsupported minimum common raft protocol version")
+	}
+
+	// This is in a scope so that future will go out of scope once we're done
+	// checking the configuration, enabling us to reuse the name later.
```
tgross, line 303: It doesn't look like we reuse `future`?

tgross (commented):
This is awesome @angrycub. I've left a few comments on how we might be able to refine this.

nomad/operator_endpoint.go
```diff
@@ -228,6 +228,134 @@ REMOVE:
 	return nil
 }
 
+// TransferLeadershipToServerAddress is used to transfer leadership away from the
+// current leader to a specific target peer. This can help prevent leadership
+// flapping during a rolling upgrade by allowing the cluster operator to target
+// an already upgraded node before upgrading the remainder of the cluster.
+func (op *Operator) TransferLeadershipToServerAddress(args *structs.RaftPeerByAddressRequest, reply *struct{}) error {
```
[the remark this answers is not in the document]
angrycub: Can we refactor that struct for both the TransferLeadership case and the RemovePeer case, or will that break the API for RemovePeer? I used RemovePeer for inspiration, but it's not strictly necessary that we use the same input structs I guess.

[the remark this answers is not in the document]
tgross: Oh, I missed that RemovePeer is where you got this. It's sort of weird that we have two request types there too, so I dug up https://github.com/hashicorp/nomad/pull/2305 and it looks like we originally had by-address only, probably because we only had Raft v2 at the time!

It'd be nice if we had a common `RaftPeerRequest` but to use that for RemovePeer we'd need to implement a new RPC and then deprecate the old one across a major version. I imagine any peer operations are most likely to happen during upgrades, after all, so it'd be really easy to break people's upgrade workflows there. 

I'd suggest creating a common `RaftPeerRequest` here and then we can come back to clean up RemovePeer in a separate PR (which doesn't have to be any time soon; we can leave a TODO for it or open an issue to do on a rainy day).

website/content/api-docs/operator/raft.mdx
```diff
@@ -100,18 +130,98 @@ The table below shows this endpoint's support for
...
     --request DELETE \
-    https://localhost:4646/v1/operator/raft/peer?address=1.2.3.4:4646
+    --header "X-Nomad-Token: ${NOMAD_TOKEN}"
+    https://127.0.0.1:4646/v1/operator/raft/peer?address=1.2.3.4:4647
 ```
 
+</Tab>
+</Tabs>
+
+## Transfer Leadership to another Raft Peer
```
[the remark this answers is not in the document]
angrycub: Maybe not? It appears that all of our currently supported versions **must** use raft protocol v3

```
Error starting agent: server config setup failed: raft_protocol must be 3 in Nomad v1.4 and later, got 2
```

I validated that against a 1.4.0 binary to make sure that I'd get the same error there and did. However, even though I wouldn't document it, I do like still having the checks in code there to handle the upgrade case.

MattJustMatt:
Thank you for this! Was looking for this command and excited to see this PR.

nomad/operator_endpoint.go
```diff
@@ -228,6 +230,132 @@ REMOVE:
 	return nil
 }
 
+// TransferLeadershipToServerID is used to transfer leadership away from the
+// current leader to a specific target peer. This can help prevent leadership
+// flapping during a rolling upgrade by allowing the cluster operator to target
+// an already upgraded node before upgrading the remainder of the cluster.
+func (op *Operator) TransferLeadershipToPeer(req *structs.RaftPeerRequest, reply *api.LeadershipTransferResponse) error {
```
tgross, line 237: This should be a `structs.LeadershipTransferResponse`.

command/operator_raft_leadership_transfer.go
```diff
@@ -0,0 +1,122 @@
...
+	Meta
+}
+
+func (c *OperatorRaftTransferLeadershipCommand) Help() string {
+	helpText := `
+Usage: nomad operator raft transfer-leadership [options]
+
+  Transfer leadership to the Nomad server with given -peer-address or
+  -peer-id in the Raft configuration. All server nodes in the cluster
+	must be running at least Raft protocol v3 in order to use this command.
```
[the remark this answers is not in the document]
tgross: Also, we should clarify here and in the docs that this requires there's already an elected leader. It doesn't work if there isn't one.

nomad/leader.go
```diff
@@ -149,6 +149,34 @@ func (s *Server) monitorLeadership() {
 	}
 }
 
+func (s *Server) leadershipTransferToServer(id raft.ServerID, addr raft.ServerAddress) error {
+	if lAddr, lID := s.raft.LeaderWithID(); id == lID && addr == lAddr {
+		s.logger.Debug("leadership transfer to current leader is a no-op")
+		return nil
+	}
+	retryCount := 3
+	for i := 0; i < retryCount; i++ {
```
tgross, line 158: Just out of curiosity, what are the circumstances in which this will need to retry, and can safely do so such that we wouldn't want to immediately return to the user?

nomad/operator_endpoint_test.go
```diff
@@ -368,6 +375,209 @@ func TestOperator_RaftRemovePeerByID_ACL(t *testing.T) {
...
+}
+
+func TestOperator_TransferLeadershipToServerID_ACL(t *testing.T) {
+	ci.Parallel(t)
+	tc := newTestCluster(t, tcArgs{enableACL: true})
+	s1 := tc.leader()
+	codec := rpcClient(t, s1)
+	state := s1.fsm.State()
+
+	_, ldrID := s1.raft.LeaderWithID()
```
tgross, line 532: Nitpick: setting up the test cluster is so expensive relative to the test cases, that it might make sense to stand up one test cluster for both `TestOperator_TransferLeadershipToServerID_ACL` and `TestOperator_TransferLeadershipToServerAddress_ACL` and then just send different requests for the only two cases that actually care about that difference. This will help avoid making test run times worse.

tgross (changes requested):
Looking really good @angrycub! I think just a few more tweaks and we can land this.

nomad/operator_endpoint_test.go
tgross: We can do this in a separate PR, if we feel like it later. Test setup is kinda rough.

nomad/leader.go
tgross: Resolved!

nomad/operator_endpoint.go
```diff
@@ -228,6 +229,127 @@ REMOVE:
...
 }
 
+// TransferLeadershipToServerID is used to transfer leadership away from the
+// current leader to a specific target peer. This can help prevent leadership
+// flapping during a rolling upgrade by allowing the cluster operator to target
+// an already upgraded node before upgrading the remainder of the cluster.
+func (op *Operator) TransferLeadershipToPeer(req *structs.RaftPeerRequest, reply *structs.LeadershipTransferResponse) error {
+	// Populate the reply's `To` with the arguments. Only one of them is likely
+	// to be filled. We don't get any additional information until after auth
+	// to prevent leaking cluster details vis the error response.
```
tgross, line 239:
```suggestion
	// to prevent leaking cluster details via the error response.
```

website/content/api-docs/operator/raft.mdx
```diff
@@ -100,18 +130,100 @@ The table below shows this endpoint's support for
...
+
+  Either `address` or `id` must be provided, but not both.
+
+  </Note>
 
 ### Sample Request
 
+
+<Tabs>
+<Tab heading="Nomad CLI">
```
tgross, line 151: Leave this for now, but adding a brand new way to demonstrate API examples is the sort of thing that really doesn't belong in a feature PR. We end up with a bunch of one-off decisions that way that aren't applied equally across the API docs. In the future, if you want to do this sort of thing, make a separate PR for it so that we can discuss it independently of critical feature work.

nomad/leader.go
```diff
@@ -149,6 +149,50 @@ func (s *Server) monitorLeadership() {
...
+		if err == raft.ErrUnsupportedProtocol {
+			return fmt.Errorf("leadership transfer not supported with Raft version lower than 3")
+		}
+
+		// ErrEnqueueTimeout: This seems to be the valid time to retry.
+		s.logger.Error("failed to transfer leadership attempt, will retry",
+			"attempt", i,
+			"retry_limit", retryCount,
+			"error", err,
+		)
```
tgross, line 190: This is real nice. :+1:

tgross (approved):
LGTM! :shipit:

PR #17384, opened 2023-06-01
node pools: add CRUD API

lgfa29:
Implement RPC and HTTP endpoints for basic node pool CRUD operations.

Docs preview:
https://nomad-git-f-node-pools-crud-hashicorp.vercel.app/nomad/api-docs/node-pools


nomad/structs/node_pool.go
```diff
@@ -133,3 +138,39 @@ func (n *NodePoolSchedulerConfiguration) Validate() error {
...
+type NodePoolUpsertRequest struct {
+	NodePools []*NodePool
+	WriteRequest
+}
+
+// NodePoolDeleteRequest is used to make a request to delete a node pool.
+type NodePoolDeleteRequest struct {
+	Names []string
+	WriteRequest
+}
```
lgfa29, line 176: These requests take a list to help synchronizing node pools with mutiregion federation.

website/content/api-docs/node-pools.mdx
```diff
@@ -0,0 +1,213 @@
...
+- `Name` `(string: <required>)` - Specifies the node pool to create or update.
+  Must have fewer than 128 characters. Only alphanumeric, `-`, and `_` are
+  allowed.
+
+- `Description` `(string: "")` - Specifies the optional human-readable
+  description of the node pool. Must have fewer than 256 characters.
+
+  - `Meta` `(object: null)` - Optional object with string keys and values of
+    metadata to attach to the node pool. Node pool metadata is not used by
+    Nomad and is intended for use by operators and third party tools.
```
tgross, line 152: extraneous indentation here:

```suggestion
- `Meta` `(object: null)` - Optional object with string keys and values of
  metadata to attach to the node pool. Node pool metadata is not used by
  Nomad and is intended for use by operators and third party tools.
```

website/content/api-docs/node-pools.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+- `prefix` `(string: "")`- Specifies a string to filter node pools based on a
+  name prefix. This is specified as a query string parameter.
+
+- `next_token` `(string: "")` - This endpoint supports paging. The `next_token`
+  parameter accepts a string which identifies the next expected node. This
+  value can be obtained from the `X-Nomad-NextToken` header from the previous
+  response.
+
+- `per_page` `(int: 0)` - Specifies a maximum number of nodes to return for
```
tgross, line 37:
```suggestion
- `per_page` `(int: 0)` - Specifies a maximum number of node pools to return for
```

website/content/api-docs/node-pools.mdx
```diff
@@ -0,0 +1,213 @@
...
+| ---------------- | ----------------- |
+| `YES`            | `node_pool:read`  |
+
+### Parameters
+
+- `prefix` `(string: "")`- Specifies a string to filter node pools based on a
+  name prefix. This is specified as a query string parameter.
+
+- `next_token` `(string: "")` - This endpoint supports paging. The `next_token`
+  parameter accepts a string which identifies the next expected node. This
```
tgross, line 33:
```suggestion
  parameter accepts a string which identifies the next expected node pool. This
```

command/agent/node_pool_endpoint.go
```diff
@@ -0,0 +1,121 @@
...
+	case "PUT", "POST":
+		return s.nodePoolUpsert(resp, req, poolName)
+	case "DELETE":
+		return s.nodePoolDelete(resp, req, poolName)
+	default:
+		return nil, CodedError(http.StatusMethodNotAllowed, ErrInvalidMethod)
+	}
+}
+
+func (s *HTTPServer) nodePoolList(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
```
tgross, line 45: I think we can always use this?

```suggestion
func (s *HTTPServer) nodePoolList(resp http.ResponseWriter, req *http.Request) (any, error) {
```

nomad/mock/acl.go
```diff
@@ -4,11 +4,13 @@
 package mock
 
 import (
+	"bytes"
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
+	"html/template"
```
tgross, line 13: I think?

```suggestion
	"text/template"
```

nomad/node_pool_endpoint_test.go
```diff
@@ -0,0 +1,1112 @@
...
+			MinQueryIndex: 1,
+		},
+	}
+	start := time.Now()
+	var resp structs.NodePoolListResponse
+	err := msgpackrpc.CallWithCodec(codec, "NodePool.List", req, &resp)
+	must.NoError(t, err)
+
+	elapsed := time.Since(start)
+	must.Greater(t, 100*time.Millisecond, elapsed)
```
tgross, line 395: Isn't this racy? The 100ms `UpsertNodePools` timer has already started by the time we call `start := time.Now()`, so it's possible to fall under this if the blocking query returns quickly enough after the watch fires. I don't think the specific timing is actually all that relevant to the test as much as the index is?

tgross (approved):
LGTM

nomad/node_pool_endpoint_test.go
lgfa29: Hum...true. I guess the time checks for the request being blocked for a while, but if it doesn't than the index check fails anyway. This was mostly copy-past from other similar tests so I didn't think too much about them 😅

nomad/mock/acl.go
lgfa29: Yeah, makes sense. This was auto-imported. Thanks!

command/agent/node_pool_endpoint.go
lgfa29: I keep forgetting about it 😬

PR #17385, opened 2023-06-01
node pool: add search support

lgfa29:
Add node pools support to the fuzzy and prefix search endpoints. 

These endpoints are used by the CLI for prefix match and shell tab-completion.

Doc preview:
https://nomad-git-f-node-pools-search-hashicorp.vercel.app/nomad/api-docs/search

### DIFF
--- a/acl/acl.go
+++ b/acl/acl.go
@@ -410,12 +410,12 @@ func (a *ACL) AllowNodePool(pool string) bool {
 	return !capabilities.Check(PolicyDeny)
 }
 
-// AllowAnyNodePool returns true if any operation is allowed in at least one
+// AllowNodePoolSearch returns true if any operation is allowed in at least one
 // node pool.
 //
 // This is a very loose check and is expected that callers perform more precise
 // verification later.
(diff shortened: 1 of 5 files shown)


lgfa29:
Moving this to draft for now because I thought it was an independent chunk of work, but it actually needs https://github.com/hashicorp/nomad/pull/17384 😅

acl/acl.go
```diff
@@ -410,6 +410,35 @@ func (a *ACL) AllowNodePool(pool string) bool {
 	return !capabilities.Check(PolicyDeny)
 }
 
+// AllowAnyNodePool returns true if any operation is allowed in at least one
+// node pool.
+//
+// This is a very loose check and is expected that callers perform more precise
+// verification later.
+func (a *ACL) AllowAnyNodePool() bool {
```
tgross, line 418: Maybe this could be `AllowNodePoolSearch` so that it's clear when people autocomplete in their editors exactly the difference? I think we did that for Variables.

lgfa29: Ah nice, I didn't notice that. I think I was going off the namespace methods.

PR #17386, opened 2023-06-01
node pools: validate pool exists on job registration

tgross:
Add a new job admission hook for node pools that enforces the pool exists on registration. Also provide the skeleton hooks we need for Enterprise enforcement functions we'll implement later.


tgross:
Note: when I merge this I'll need to fix up the missing `_ent.go` file in ENT, so I'll wait until I have time to land this and then immediately do that so the nightly OSS->ENT merge isn't broken.

lgfa29 (approved):

Issue #16968, opened 2023-04-23
Nomad config validate does not check for missing host_volume directories and daemon fails to start


maxadamo:
### Proposal

`nomad config validate <config_file>` does not detect missing files/directories for the `host_volume` parameter and Nomad fails to start

### Use-cases

if a directory/file is missing, nomad won't start. 

### Attempted Solutions

With Puppet we created a custom config validation, which checks that the paths are valid on the system: 
https://github.com/voxpupuli/puppet-nomad/pull/85

but I'd expect this same functionality to be available on `nomad config validate`

maxadamo:
an issue has been open against Nomad: https://github.com/hashicorp/nomad/issues/16968
Meantime the user can choose between the ruby script and nomad validation.

dttung2905:
Hi team,

I would like to attempt to fix this issue. I just added a PR above :pray:

gulducat:
By my eye, that PR ensures that `path` is present in `host_volume` config and not an empty string, which is a nice improvement, but it doesn't check for the presence of the directory on disk.

Seems the latter is the desired outcome of the issue, so I'll go ahead and reopen.

Can repro with this (partial) config:

```hcl
client {
  enabled = true

  host_volume "nope" {
    path = "/nope" # assuming your system lacks this dir
  }
}
```

```shell
$ nomad config validate client.hcl
Configuration is valid!
```

but when trying to start the agent:

```shell
$ nomad agent -config client.hcl
.....
[ERROR] agent: error starting agent: error="client setup failed: node setup failed: failed to validate volume nope, err: stat /nope: no such file or directory"
```

lgfa29:
Ops, my bad I missed that in the review.

@dttung2905 would you be interested in extending the work you did in https://github.com/hashicorp/nomad/pull/17393 to also check if the host volume `path` is actually a directory that exists?

dttung2905:
Sure @lgfa29 , please allow me some time to take a look at it again :pray:

maxadamo:
> @dttung2905 would you be interested in extending the work you did in #17393 to also check if the host volume `path` is actually a directory that exists?

@lgfa29 , and @dttung2905 for clarity: it's doesn't have to be a directory. It can also be a file.

lgfa29:
Ah interesting, like a symlink to a directory? I don't think I've ever tried that 😄

maxadamo:
@lgfa29 I just checked my configuration, to ensure that I was remembering correctly. 
[...]

PR #17393, opened 2023-06-02
Add check for missing `path` in client `host_volume` config

dttung2905:
Hi team,

This PR aims to add additional check for missing `path` value in client `host_volume` config. I have added 2 more test cases to validate this change. Please let me know if I need to make any additional changes for this :pray: 

Fixes : https://github.com/hashicorp/nomad/issues/16968

### DIFF
--- a/command/agent/command.go
+++ b/command/agent/command.go
@@ -393,6 +393,13 @@ func (c *Command) IsValidConfig(config, cmdConfig *Config) bool {
 		}
 	}
 
+	for _, volumeConfig := range config.Client.HostVolumes {
+		if volumeConfig.Path == "" {
+			c.Ui.Error("Missing path in host_volume config")
+			return false
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


command/agent/command.go
```diff
@@ -379,6 +379,15 @@ func (c *Command) IsValidConfig(config, cmdConfig *Config) bool {
 		return false
 	}
 
+	if !reflect.ValueOf(config.Client.HostVolumes).IsZero() {
```
lgfa29, line 382: Is this condition necessary? If the list `IsZero` then the range below would be skipped anyway right?

lgfa29 (commented):
Thanks for the PR @dttung2905!

I left a quick question there just to make sure I'm not missing anything.

command/agent/command.go
dttung2905: Hi @lgfa29,
Thanks for the super quick review! It is indeed redundant here .

lgfa29 (approved):
Awesome, thanks for checking that 😄

PR #17396, opened 2023-06-02
node pools: implement RPC to list jobs in a given node pool

tgross:
Implements the `NodePool.ListJobs` RPC, with pagination and filtering based on the existing `Job.List` RPC.


nomad/node_pool_endpoint.go
```diff
@@ -252,3 +253,128 @@ func (n *NodePool) DeleteNodePools(args *structs.NodePoolDeleteRequest, reply *s
 	reply.Index = index
 	return nil
 }
+
+// ListJobs is used to retrieve a list of jobs for a given node pool. It supports
+// pagination and filtering.
+func (n *NodePool) ListJobs(args *structs.NodePoolJobsRequest, reply *structs.NodePoolJobsResponse) error {
+	authErr := n.srv.Authenticate(n.ctx, args)
+	if done, err := n.srv.forward("NodePool.Jobs", args, args, reply); done {
```
shoenig, line 261: doesn't this need to be `NodePool.ListJobs` ?

tgross: oof, yes.

tgross: Fixed

nomad/node_pool_endpoint.go
```diff
@@ -252,3 +253,128 @@ func (n *NodePool) DeleteNodePools(args *structs.NodePoolDeleteRequest, reply *s
...
+			_, err := store.NodePoolByName(ws, args.Name)
+			if err != nil {
+				return err
+			}
+
+			var iter memdb.ResultIterator
+
+			// Get the namespaces the user is allowed to access.
+			allowableNamespaces, err := allowedNSes(aclObj, store, allowNsFunc)
+			if err == structs.ErrPermissionDenied {
```
shoenig, line 297: This is probably copy/paste, but should we be using `errors.Is` for these now?

nomad/node_pool_endpoint.go
```diff
@@ -252,3 +253,128 @@ func (n *NodePool) DeleteNodePools(args *structs.NodePoolDeleteRequest, reply *s
...
+func (n *NodePool) ListJobs(args *structs.NodePoolJobsRequest, reply *structs.NodePoolJobsResponse) error {
+	authErr := n.srv.Authenticate(n.ctx, args)
+	if done, err := n.srv.forward("NodePool.ListJobs", args, args, reply); done {
+		return err
+	}
+	n.srv.MeasureRPCRate("node_pool", structs.RateMetricRead, args)
+	if authErr != nil {
+		return structs.ErrPermissionDenied
+	}
+	defer metrics.MeasureSince([]string{"nomad", "node_pool", "list_jobs"}, time.Now())
```
lgfa29, line 269: This needs to be documented in https://developer.hashicorp.com/nomad/docs/operations/metrics-reference#server-metrics

shoenig (approved):
LGTM! would be nice if we could one day refactor both the auth to be declarative, and the `opts` query to be ... less manual?

nomad/node_pool_endpoint.go
```diff
@@ -252,3 +253,128 @@ func (n *NodePool) DeleteNodePools(args *structs.NodePoolDeleteRequest, reply *s
...
+	// an additional paginator filter for the node pool.
+	opts := blockingOptions{
+		queryOpts: &args.QueryOptions,
+		queryMeta: &reply.QueryMeta,
+		run: func(ws memdb.WatchSet, store *state.StateStore) error {
+			// ensure the node pool exists
+			_, err := store.NodePoolByName(ws, args.Name)
+			if err != nil {
+				return err
+			}
```
lgfa29, line 292: I think you need to check for a `nil` pool result as well.

nomad/node_pool_endpoint.go
```diff
@@ -252,3 +253,128 @@ func (n *NodePool) DeleteNodePools(args *structs.NodePoolDeleteRequest, reply *s
...
+			} else if err != nil {
+				return err
+			} else {
+				if prefix := args.QueryOptions.Prefix; prefix != "" {
+					iter, err = store.JobsByIDPrefix(ws, namespace, prefix)
+				} else if namespace != structs.AllNamespacesSentinel {
+					iter, err = store.JobsByNamespace(ws, namespace)
+				} else {
+					iter, err = store.Jobs(ws)
+				}
```
lgfa29, line 311: When I was working on the `/v1/node/pool/:name/nodes` endpoint I created a new index for the `NodePool` field in the nodes table. Would it make sense to have one for jobs as well? Or maybe even a `node_pool/namespace` compound index.

I think we could also drop the prefix option here. For `Job.List` I think the main use case is for CLI completion, which is not needed for this endpoint.

lgfa29 (approved):
Looking good. I think the index may worth considering, but I'm not sure if it would result in too much overhead.

nomad/node_pool_endpoint.go
tgross: Node pools are going to be very low cardinality relative to the number of jobs, so an index could cut out big chunks of jobs to iterate. But indexes on jobs are pretty big too. It'd be easy to add later if we find it's a problem or we want it for the scheduler for some reason. Lemme get a rough idea of what the real size is and see if it matters.

> I think we could also drop the prefix option here. For `Job.List` I think the main use case is for CLI completion, which is not needed for this endpoint.

Come to think of it, it's probably confusing to use too -- is it the prefix of the node pool or the job that we're getting here? Will fix.

lgfa29: For reference, I pushed my work on the `/nodes` endpoint
https://github.com/hashicorp/nomad/pull/17413

tgross: I ran the following benchmark and found that the index increased the state store size by only around ~54MiB for 10k jobs (that's ~11% of the state store in this benchmark, but this benchmark has no nodes or allocations so percentages are misleading). Seems worth it, and trivial to drop later if we want.

```go
// added to jobs table
"pool": {
    Name:         "pool",
    AllowMissing: false,
    Unique:       true,
    Indexer: &memdb.CompoundIndex{
        Indexes: []memdb.Indexer{
            &memdb.StringFieldIndex{
                Field: "NodePool",
            },
            &memdb.StringFieldIndex{
                Field: "Namespace",
            },

            &memdb.StringFieldIndex{
                Field: "ID",
            },
        },
    },
},
```

```go
func TestNodePool_IndexSize(t *testing.T) {
	n := 10000

	res := testing.Benchmark(func(b *testing.B) {
		b.ReportAllocs()

		store, _ := state.NewStateStore(&state.StateStoreConfig{
			Logger: testlog.HCLogger(t),
			Region: "global",
		})

		for i := 0; i < n; i++ {
			j := mock.MinJob()
			store.UpsertJob(structs.MsgTypeTestSetup, uint64(i), nil, j)
		}
	})

	fmt.Println("size: ", res.MemBytes)
}
```

tgross: Oops, I don't actually want the `ID` in that index, so that makes it even smaller.

tgross: Added the index for use with wildcard namespaces only, as adding it for both wildcard and specific namespaces would end up doubling the size of the index. When we have a specific namespace, the namespace itself is a better index and the filter is relatively cheap, but for wildcard namespace having the index has a good payoff.

PR #17399, opened 2023-06-02
UI-specific GHA test changes re-implemented

philrenaud:
UI-specific tests that were pulled from https://github.com/hashicorp/nomad/pull/17103 — they were necessitated because GHA appears to run our acceptance tests under a different order / set of circumstances than Circle.

This hardens the tests a bit and gives them some sensible defaults (like "Batch jobs do not have deployments", etc.)


tgross (approved):
:+1:

PR #17400, opened 2023-06-02
np: fix node pool search permission check

lgfa29:
When checking if a token is allowed to query the search endpoints we need to return an error if the search context includes `node_pool` and the token doesn't have access to _any_ pool. This prevents returning an empty list instead of a permission denied error.


tgross (approved):
LGTM!

PR #17405, opened 2023-06-02
node pools: register a node in a node pool

lgfa29:
Add new client configuration `node_pool` to allow assigning a node to a pool. If the node pool does not exist yet, it will be created automatically when the node registers. The built-in `all` node pool is not allowed to be used.

Doc preview:
https://nomad-git-f-node-pools-register-node-hashicorp.vercel.app/nomad/docs/configuration/client#node_pool

### DIFF
--- a/command/agent/command.go
+++ b/command/agent/command.go
@@ -25,7 +25,6 @@ import (
 	checkpoint "github.com/hashicorp/go-checkpoint"
 	discover "github.com/hashicorp/go-discover"
 	hclog "github.com/hashicorp/go-hclog"
-	"github.com/hashicorp/go-set"
 	gsyslog "github.com/hashicorp/go-syslog"
 	"github.com/hashicorp/logutils"
 	"github.com/hashicorp/nomad/helper"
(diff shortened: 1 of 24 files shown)


lgfa29:
I was running the `api` tests with a stale binary 🤦

nomad/node_endpoint.go
```diff
@@ -135,6 +136,17 @@ func (n *Node) Register(args *structs.NodeRegisterRequest, reply *structs.NodeUp
 	if args.Node.SecretID == "" {
 		return fmt.Errorf("missing node secret ID for client registration")
 	}
+	// Validate node pool value if provided. The node is canonicalized in the
+	// FSM, where an empty node pool is set to "default".
+	if args.Node.NodePool != "" {
+		invalidNames := set.From([]string{
+			structs.NodePoolAll,
+		})
+		err := structs.ValidateNodePoolName(args.Node.NodePool, invalidNames)
```
tgross, line 145: Are we ever going to call this without a single item set of `structs.NodePoolAll`?

nomad/fsm_test.go
```diff
@@ -241,6 +241,98 @@ func TestFSM_UpsertNode_Canonicalize_Ineligible(t *testing.T) {
 	require.Equal(structs.NodeSchedulingIneligible, n.SchedulingEligibility)
 }
 
+func TestFSM_UpsertNode_NodePool(t *testing.T) {
```
tgross, line 244: I really like how this test hits all the invariants we really care about :+1:

tgross (approved):
Looks great!

nomad/node_endpoint.go
shoenig: It would probably make sense to extract `invalidNames` into a "constant" `var` in `structs.go` and hide its details in `ValidateNodePoolName` - since we never call this helper with any other variation. That way we don't duplicate the creation of `invalidNames` (here and in agent/command.go)

command/agent/command.go
```diff
@@ -1407,6 +1422,10 @@ Client Options:
     Mark this node as a member of a node-class. This can be used to label
     similar node types.
 
+  -node-pool
+    Register this node under this node pool. If the node pool does not exist it
```
shoenig, line 1426:
```suggestion
    Register this node under this node pool. If the node pool does not exist it
```

Maybe s/under/in ? I think of stuff being _in_ a pool, rather than under it

nomad/node_endpoint.go
lgfa29: I thought I had a case where `default` was also not valid, but I think it may have gone away as the code evolve. I will rethink this bit 😄

lgfa29: Yeah, I couldn't find the case I was thinking about. I removed the second argument and checked for `all` in the command and register endpoint directly.

PR #17412, opened 2023-06-02
node pools: add event stream support

lgfa29:
Add event stream support for node pools. Upsert and delete actions are 

Doc preview:
https://nomad-git-f-node-pools-event-stream-hashicorp.vercel.app/nomad/api-docs/events

### DIFF
--- a/api/event_stream_test.go
+++ b/api/event_stream_test.go
@@ -42,6 +42,10 @@ func TestTopic_String(t *testing.T) {
 			expectedOutput: "Node",
 		},
 		{
+			inputTopic:     TopicNodePool,
+			expectedOutput: "NodePool",
+		},
+		{
 			inputTopic:     TopicService,
 			expectedOutput: "Service",
 		},
@@ -299,6 +303,19 @@ func TestEventStream_PayloadValueHelpers(t *testing.T) {
 				must.Eq(t, &Node{
 					ID:         "some-id",
 					Datacenter: "some-dc-id",
+				}, n)
+			},
+		},
+		{
+			desc:  "node_pool",
+			input: []byte(`{"Topic":"NodePool","Payload":{"NodePool":{"Description":"prod pool","Name":"prod"}}}`),
+			expectFn: func(t *testing.T, event Event) {
+				must.Eq(t, TopicNodePool, event.Topic)
+				n, err := event.NodePool()
+				must.NoError(t, err)
+				must.Eq(t, &NodePool{
+					Name:        "prod",
+					Description: "prod pool",
 				}, n)
 			},
 		},
--- a/nomad/state/state_store_node_pools.go
+++ b/nomad/state/state_store_node_pools.go
@@ -24,7 +24,7 @@ func (s *StateStore) nodePoolInit() error {
 	}
 
 	return s.UpsertNodePools(
-		structs.NodePoolUpsertRequestType,
+		structs.SystemInitializationType,
 		1,
 		[]*structs.NodePool{allNodePool, defaultNodePool},
 	)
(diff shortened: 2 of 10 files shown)


nomad/stream/event_broker.go
```diff
@@ -367,6 +367,12 @@ func aclAllowsSubscription(aclObj *acl.ACL, subReq *SubscribeRequest) bool {
 			if ok := aclObj.AllowNodeRead(); !ok {
 				return false
 			}
+		case structs.TopicNodePool:
+			// Require management token for node pools since we can't filter
+			// out node pools the token doesn't have access to.
+			if ok := aclObj.IsManagement(); !ok {
+				return false
+			}
```
lgfa29, line 375: A management token here feels like an overkill, but I couldn't find a more granular way to check this. One option could be to check for `nomad_poll "*"` read capability or create a new capability that can only be used with `*`, but these felt a bit off.

lgfa29:
It appears I have broken stuff 😅 

Converting to draft for now.

nomad/structs/structs.go
```diff
@@ -134,6 +134,11 @@ const (
 )
 
 const (
+	// SystemInitializationType is used for messages that initialize parts of
+	// the system, such as the state store. These messages are not included in
+	// the event stream.
+	SystemInitializationType MessageType = 127
```
lgfa29, line 140: I didn't know how to call this, so I'm open to suggestions 😅 

I wanted to keep it somewhat generic because `MessageType` is an `uint8`, so we don't have that many slots open. I also thought it would be interesting to have a `IgnoreEventStreamTypeFlag` that you could `&` to the appropriate message type, but we don't have any bits left since `IgnoreEventStreamTypeFlag` already uses 128.

lgfa29:
Tests are fixed so this is ready for review now.

api/event_stream_test.go
```diff
@@ -302,6 +306,19 @@ func TestEventStream_PayloadValueHelpers(t *testing.T) {
...
+			expectFn: func(t *testing.T, event Event) {
+				must.Eq(t, TopicNodePool, event.Topic)
+				n, err := event.NodePool()
+				must.NoError(t, err)
+				must.Eq(t, &NodePool{
+					Name:        "prod",
+					Description: "prod pool",
+				}, n)
+			},
+		},
```
tgross, line 321: This new test case is fine but this test sure is weird to me. Do you know why we're testing that we can unmarshal the event JSON to structs? I feel like I'm missing something :grinning:

nomad/state/state_store_node_pools.go
```diff
@@ -24,7 +24,7 @@ func (s *StateStore) nodePoolInit() error {
 	}
 
 	return s.UpsertNodePools(
-		structs.NodePoolUpsertRequestType,
+		structs.SystemInitializationType,
```
tgross, line 27: Even though we don't currently emit events for namespaces, should we set this for the `namespaceInit` as well?

api/event_stream_test.go
lgfa29: Hum...right, that's kind of weird 😅 

I guess it's testing the `event.NodePool()` method that extracts a `NodePool`object from the event? But that is also mostly just mapstructure 🤷

nomad/state/state_store_node_pools.go
lgfa29: I can do this in a follow-up PR. `UpsertNamesapce` doesn't currently take a `MessageType` so there will be dozens of places that need to be updated because of the method signature changed.

tgross: Oh, sorry I thought we'd just be replacing in `UpsertNamespace`. Probably not worth messing with before the 1.6.0 beta then.

lgfa29: Ah yes, there's no message type being set by any namespace operation (that's why it's ignored by the event stream 😅)

PR #17413, opened 2023-06-02
node pools: list nodes in pool

lgfa29:
Implement HTTP and RPC for listing nodes that are part of a node pool.

Doc preview:
https://nomad-git-f-node-pools-nodes-api-hashicorp.vercel.app/nomad/api-docs/node-pools#list-node-pool-nodes

I think there's a small conflict with https://github.com/hashicorp/nomad/pull/17431 in the HTTP handler, but I adjusted the code to match the style of #17396 and #17431 so I think this is ready for review. I will just rebase it before merge after it's approved.


nomad/node_pool_endpoint.go
```diff
@@ -252,3 +254,92 @@ func (n *NodePool) DeleteNodePools(args *structs.NodePoolDeleteRequest, reply *s
...
+		queryMeta: &reply.QueryMeta,
+		run: func(ws memdb.WatchSet, store *state.StateStore) error {
+			// Verify node pool exists.
+			pool, err := store.NodePoolByName(ws, args.Name)
+			if err != nil {
+				return err
+			}
+			if pool == nil {
+				return fmt.Errorf("node pool %q not found", args.Name)
+			}
```
lgfa29, line 300: Should this be checked in the state store? Other endpoints like this (`/v1/job/:id/allocations` for example) don't even check if `:id` exists 😅

tgross: Having it here lets us bail out early so we don't have to do all the paginator setup for what's inevitably going to be an empty iterator. We just return `nil` and check again on the next pass if the pool exists, which is particularly nice for a blocking query where the node pool will get created automatically when the first node in the pool registers.

lgfa29: Updated to return `nil` instead of error.

command/agent/node_pool_endpoint.go
```diff
@@ -119,3 +123,37 @@ func (s *HTTPServer) nodePoolDelete(resp http.ResponseWriter, req *http.Request,
...
+		Name: poolName,
+	}
+	if s.parse(resp, req, &args.Region, &args.QueryOptions) {
+		return nil, nil
+	}
+
+	// Parse node fields selection.
+	fields, err := parseNodeListStubFields(req)
+	if err != nil {
+		return nil, CodedError(http.StatusBadRequest, fmt.Errorf("Failed to parse node list fields: %v", err).Error())
```
tgross, line 138: code golf:

```suggestion
		return nil, CodedError(http.StatusBadRequest, fmt.Sprintf("Failed to parse node list fields: %v", err))
```

nomad/node_pool_endpoint.go
```diff
@@ -383,3 +383,93 @@ func (n *NodePool) ListJobs(args *structs.NodePoolJobsRequest, reply *structs.No
...
 }
+
+// ListNodes is used to retrieve a list of nodes for a give node pool. It
+// supports pagination and filtering.
+func (n *NodePool) ListNodes(args *structs.NodePoolNodesRequest, reply *structs.NodePoolNodesResponse) error {
+	authErr := n.srv.Authenticate(n.ctx, args)
+	if done, err := n.srv.forward("NodePool.ListNodes", args, args, reply); done {
+		return err
+	}
+	n.srv.MeasureRPCRate("node_pool", structs.RateMetricRead, args)
```
tgross, line 394: I sort of regret that we have this metric split between read and list, but:

```suggestion
	n.srv.MeasureRPCRate("node_pool", structs.RateMetricList, args)
```

tgross (approved):
LGTM!

nomad/node_pool_endpoint.go
lgfa29: Ah thanks. I thought this was like "I'm reading the node pool", but it does return a list, so I guess from a rate limiting point of view this is a more expensive endpoint. 

I fixed this is `ListJobs` as well 👍

tgross: D'oh, I missed it in `ListJobs` :grinning:

PR #17430, opened 2023-06-05
node pools: fix node upsert and state mutation tests

lgfa29:
Move tests that assumed canonicalization was been done in the state store to the FSM, where it actually happens, and fix some state store pointer mutation.


tgross (approved):
LGTM!

PR #17431, opened 2023-06-05
node pools: implement HTTP API to list jobs in pool

tgross:
Implements the HTTP API associated with the `NodePool.ListJobs` RPC, including the `api` package for the public API. Also fixes a bug in the RPC handler where we were missing handling of the special `all` node pool.

### DIFF
--- a/api/node_pools.go
+++ b/api/node_pools.go
@@ -5,6 +5,7 @@ package api
 
 import (
 	"errors"
+	"fmt"
 	"net/url"
 )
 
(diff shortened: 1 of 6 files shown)


api/node_pools.go
```diff
@@ -103,3 +104,15 @@ type NodePool struct {
...
+func (n *NodePools) ListJobs(poolName string, q *QueryOptions) ([]*JobListStub, *QueryMeta, error) {
+	var resp []*JobListStub
+	qm, err := n.client.query(
+		fmt.Sprintf("/v1/node/pool/%s/jobs", url.PathEscape(poolName)),
+		&resp, q)
+	if err != nil {
+		return nil, nil, err
+	}
+	return resp, qm, nil
+}
```
lgfa29, line 118: It would be nice to keep this up there along with the other methods, or maybe move the structs up?

lgfa29 (approved):
LGTM.

We need to document this endpoint, which could be done here or is a separate PR.

command/agent/node_pool_endpoint.go
```diff
@@ -125,6 +128,10 @@ func (s *HTTPServer) nodePoolDelete(resp http.ResponseWriter, req *http.Request,
 }
 
 func (s *HTTPServer) nodePoolNodesList(resp http.ResponseWriter, req *http.Request, poolName string) (interface{}, error) {
+	if req.Method != http.MethodGet {
+		return nil, CodedError(http.StatusMethodNotAllowed, ErrInvalidMethod)
+	}
```
tgross, line 133: @lgfa29 when I rebased on `main` I added this assertion so we don't list on `POST` or whatever.

lgfa29: Ah good catch, thanks!

PR #17432, opened 2023-06-05
node pools: implement CLI for `node pool jobs` command

tgross:
_WIP_ this PR current includes everything in https://github.com/hashicorp/nomad/pull/17431. I'll rebase once that's been merged.


lgfa29 (commented):
Leaving some early review because I'm nosy 😅

PR #17434, opened 2023-06-05
Making Workload ID OIDC compliant

schmichael:
## Rough roadmap

- Milestone 1: Allow Workload Identities to be used with [Consul JWT Auth Method](https://developer.hashicorp.com/consul/docs/security/acl/auth-methods/jwt)
- Milestone 1.1: Allow WI to be used with Vault's JWT Auth Method
- Milestone 2: Allow WI to be used with Cloud IAM auth methods: [AWS](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_oidc.html) - [GCP](https://cloud.google.com/iam/docs/workload-identity-federation-with-other-providers)
- Milestone X: Replace existing uses of hardcoded Consul and Vault tokens in Nomad agents with Workload Identities

## Milestone 1 TODO

- [X] jobspec/struct fields
- [ ] Set additional claims (audience, expiration, issuer, etc)
- [x] ~SigningKey Index removal~ _Index must be maintained so pre-1.6 Clients with old tokens don't have their signing keys revoked after 30 days._
- [ ] Token Refresh: Add Server RPC for Clients to derive tokens
- [ ] Root Key Signing Key Rotation: default `root_key_rotation_threshold` to 30 days and switch from TimeTable GC to CreateTime
- [ ] JWKS - new roots should be published *before* being made active as consumers will cache this endpoint
- [ ] Consul Acceptance Testing
- [ ] Upgrade Path: Clients should rotate pre-1.6 tokens on startup

### Gotchas

- [ ] Must publish new root keys via JWKS *before* they're active
- [ ] Must ignore expiration when authenticating old Nomad JWTs in WhoAmI (otherwise 1.5 Clients talking to 1.6 Servers will start getting auth errors after the default 30 day expiration)

### Out of scope

- Expose WhoAmI RPC as /validate/token and follow [RFC-7662](https://datatracker.ietf.org/doc/html/rfc7662)
- OIDC metadata endpoint (`/.well-known/openid-configuration`) - This will be added for Milestone 2.


schmichael:
Closing this horribly outdated PoC branch. It served its purpose and `main` now has better versions of this code.

PR #17443, opened 2023-06-06
node pools: implement support in scheduler

tgross:
When a scheduler is invoked, we get a set of the ready nodes in the DCs that are allowed for that job. Extend the filter to include the node pool, ensure that changes to a job's node pool are picked up as destructive allocation updates, and add `NodesInPool` as a metric to all reporting done by the scheduler.


command/monitor.go
```diff
@@ -324,8 +324,11 @@ func formatAllocMetrics(metrics *api.AllocationMetric, scores bool, prefix strin
 		out += fmt.Sprintf("%s* No nodes were eligible for evaluation\n", prefix)
 	}
 
-	// Print a helpful message if the user has asked for a DC that has no
+	// Print a helpful message if the user has asked for a DC or pool that has no
 	// available nodes.
+	if metrics.NodesInPool == 0 {
+		out += fmt.Sprintf("%s* No nodes are available in pool\n", prefix)
```
lgfa29, line 330: It would be nice to display the node pool name here because, while you could derive it from the job, it may not be so obvious in the context where this information is displayed, like in the CLI.

```
==> 2023-06-06T19:04:36-04:00: Monitoring evaluation "9abb17c1"
    2023-06-06T19:04:36-04:00: Evaluation triggered by job "example"
    2023-06-06T19:04:37-04:00: Evaluation within deployment: "d6f075ce"
    2023-06-06T19:04:37-04:00: Evaluation status changed: "pending" -> "complete"
==> 2023-06-06T19:04:37-04:00: Evaluation "9abb17c1" finished with status "complete" but failed to place all allocations:
    2023-06-06T19:04:37-04:00: Task Group "shared" (failed to place 1 allocation):
      * No nodes were eligible for evaluation
      * No nodes are available in pool
    2023-06-06T19:04:37-04:00: Evaluation "ff453ae0" waiting for additional capacity to place remainder
```

But I think that would need some extra plumbing, or to store the node pool name in the metric, either as a new field or have this be a map of 1, which is kind of weird 🤔

lgfa29 (approved):
LGTM! I did a bit of manual testing and everything seems to be working 🎉 

One thing I tried was to have something like this:
```hcl
job "example" {
  node_pool = "all"

  constraint {
    attribute = "${node.node_pool}"
    value     = "default"
  }
  # ...
}
```

Which didn't quite work. I think we need to update [`resolveTarget`](https://github.com/hashicorp/nomad/blob/main/scheduler/feasible.go#L791-L825) to interpolate `${node.node_pool}`.

After I tried that I would sometimes get this as the placement failure message, which I'm not sure what it means, so there may be more missing 😅 
```
    2023-06-06T19:36:44-04:00: Task Group "shared" (failed to place 1 allocation):
      * Constraint "computed class ineligible": 1 nodes excluded by filter
```

scheduler/generic_sched.go
```diff
@@ -591,6 +591,7 @@ func (s *GenericScheduler) computePlacements(destructive, place []placementResul
 
 			// Store the available nodes by datacenter
 			s.ctx.Metrics().NodesAvailable = byDC
+			s.ctx.Metrics().NodesInPool = len(nodes)
```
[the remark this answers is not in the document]
tgross: That is odd. I suspect it's happening because each loop is adding other scoring metadata to the metrics, and we didn't think about where to set this. If the `s.ctx` changed I could see it, but that's set once for the whole eval.

command/monitor.go
tgross: That would make for a weird API. The caller of the formatter has the `api.Job` though, so we can thread it down from there into this function.

tgross: As it turns out, we don't always have the `api.Job`, because this gets called in the context of `nomad eval status` too, which only has strings of the `JobID` and not the Job itself. 

But I'm also realizing that we'll always get the "No nodes were eligible for evaluation" and "No nodes available in pool" at the same time, as we don't have any way of differentiating between the two cases so the added error message doesn't tell us anything at all! I'm going to remove it because it's just going to generate confusion.

scheduler/generic_sched.go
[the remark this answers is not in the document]
lgfa29: Ah nice find, thanks for digging! I'm glad we have good test coverage here 😄

command/monitor.go
lgfa29: Ah right, `NodesEvaluated` will always be 0 because `readyNodesInDCsAndPool` will have filtered them all out. 

It could be helpful to hint that node pool is the issue, specially in the case where the namespace default pool is being used, but other than adding the node pool name to AllocMetric I can't think of a good way to do it 🤔 

The `all` node pool is a bit weird here as well. It's kind of like `datacenters = ["*"]`, so maybe it does make sense to have a map for node pools? Not sure 😅

Issue #17023, opened 2023-04-28
docker task driver overrides max_kill_timeout


louievandyke:
#### Describe the bug
 
The Docker Task drivers' [dockerTimeout setting](https://github.com/hashicorp/nomad/blob/891999789689240006b2a6076b73ddc67249d48d/drivers/docker/config.go#L39-L41) supersedes Nomad's `max_kill_timeout` setting.
 
#### Steps to reproduce the behavior.
Configure a `max_kill_timeout` for a period longer than 5 minutes.  Run a Nomad job that utilizes the docker driver and then initiate a stop.  Once an initial signal is sent via nomad alloc stop <allocid> the docker driver will send another signal in `5 * time.Min`
 
#### Expected behavior
The expected behavior is that Nomad's max_kill_timeout will override task driver timeout settings.
 
Product Version: _Current_ 1.5.x

```
ubuntu@ip-172-31-17-63:~$ cat /etc/nomad.d/nomad.hcl | grep max
  max_kill_timeout = "10m"
ubuntu@ip-172-31-17-63:~$
```
```
job "simple" {
  datacenters = ["dc1"]
  type = "batch"

  periodic {
    cron             = "*/1 * * * *"
    prohibit_overlap = true
  }

  group "simple" {
    task "simple" {
      driver       = "docker"
      kill_timeout = "10m"

      config {
        image   = "ubuntu:latest"
        command = "bash"
        args    = ["-c", "/local/script.sh"]
      }

      template {
        data        = <<EOF
#!/bin/bash

# set trap
trap 'echo "Received SIGTERM doint nothing..."' SIGTERM
i=1

for i in {1..250}; do
#Print message with counter i
echo "[$(date -u)] running the loop for $i times; sleeping"
sleep 5
#Increment the counter by one
((i++))
done
EOF
        perms       = "755"
        destination = "local/script.sh"
      }

      resources {
        cpu    = 100
        memory = 100
      }
    }
  }
}
```
```
ubuntu@ip-172-31-17-63:~$ nomad job status
ID                          Type            Priority  Status   Submit Date
example                     service         50        running  2022-07-22T15:15:01Z
plugin-aws-ebs-controller   service         50        running  2022-07-23T01:11:39Z
plugin-aws-ebs-nodes        system          50        running  2022-07-23T01:11:35Z
plugin-efs                  system          50        running  2022-07-23T01:08:48Z
simple                      batch/periodic  50        running  2022-08-02T05:45:57Z
[...]

PR #17445, opened 2023-06-06
Use wait client for stopping docker tasks

mikenomitch:
Fixes https://github.com/hashicorp/nomad/issues/17023

The [docker client timeout value](https://github.com/hashicorp/nomad/blob/84e7cf39f6c7b0aee49d174ba03bd95824c83c59/drivers/docker/config.go#LL41C6-L41C6) was causing a SIGTERM to be sent to the process 5 minutes after an initial kill signal was sent. This was confirmed by changing this value to various other values, recompiling, and then watching the timestamps of various SIGTERM's sent to the task.

By using the waitClient, which does not use a timeout, instead of the standard client, we can get around the additional SIGTERM being sent. In this case, there is no timeout and we wait until the client's max_kill_timeout value or job's kill_timeout value.

Note: I have not tested this with alternative signals yet, just the default. There is a different codepath for that. 

Also, this could totally be the wrong direction and have implications I don't understand, but I figured it's at least a start of the convo :)

### DIFF
--- a/drivers/docker/handle.go
+++ b/drivers/docker/handle.go
@@ -159,7 +159,7 @@ func (h *taskHandle) Kill(killTimeout time.Duration, signal string) error {
 	// Signal is used to kill the container with the desired signal before
 	// calling StopContainer
 	if signal == "" {
-		err = h.client.StopContainer(h.containerID, uint(killTimeout.Seconds()))
+		err = h.waitClient.StopContainer(h.containerID, uint(killTimeout.Seconds()))
 	} else {
 		ctx, cancel := context.WithTimeout(context.Background(), killTimeout)
 		defer cancel()
--- a/drivers/docker/handle.go
+++ b/drivers/docker/handle.go
@@ -192,7 +192,7 @@ func (h *taskHandle) Kill(killTimeout time.Duration, signal string) error {
 		}
 
 		// Stop the container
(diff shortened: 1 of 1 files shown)


drivers/docker/handle.go
```diff
@@ -159,7 +159,7 @@ func (h *taskHandle) Kill(killTimeout time.Duration, signal string) error {
 	// Signal is used to kill the container with the desired signal before
 	// calling StopContainer
 	if signal == "" {
-		err = h.client.StopContainer(h.containerID, uint(killTimeout.Seconds()))
+		err = h.waitClient.StopContainer(h.containerID, uint(killTimeout.Seconds()))
```
mikenomitch, line 162: TBH, I am not sure of all the implications of this, but I can confirm that this does "solve" the issue in the ticket.

mikenomitch:
After talking with the team, the actual content of this PR probably isn't the answer, but it is hopefully a good start in the debugging process.

I don't currently know why (non-wait) client's timeout is sending a SIGTERM.

Is it automatic retry behavior in Nomad?
Is it a bug in the go-dockerclient library?
Is it something in a dialer that we set up?
Is this behavior happening on Podman as well? (if so, can we fix it)

All of these questions should probably be answered before we move forward with a solution.

Going to close this out as I probably won't be able to follow all of these threads myself.

PR #17457, opened 2023-06-07
node pools: protect against deleting occupied pools

tgross:
We don't want to delete node pools that have nodes or non-terminal jobs. Add a check in the `DeleteNodePools` RPC to check locally and in federated regions, similar to how we check that it's safe to delete namespaces.

### DIFF
--- a/nomad/node_pool_endpoint.go
+++ b/nomad/node_pool_endpoint.go
@@ -11,6 +11,7 @@ import (
 
 	metrics "github.com/armon/go-metrics"
 	"github.com/hashicorp/go-memdb"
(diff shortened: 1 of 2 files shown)


lgfa29 (approved):
LGTM! It doesn't look like we do it for namespaces, but I think it would be helpful to document these requirements in the CLI and API docs.

tgross:
> It doesn't look like we do it for namespaces, but I think it would be helpful to document these requirements in the CLI and API docs.

Agreed. I'm going to add that to our checklist and will pick it up later this week.

PR #17465, opened 2023-06-09
[ui] Don't show a service as healthy when its parent alloc is not running

philrenaud:
Adds a "Health Unknown" indicator and tooltip when the allocation a service is running on is not in a running state.

Previously, any /checks would return `null` and it would optimistically show Healthy for its aggregate status.

image


tgross (approved):
LGTM!

PR #17474, opened 2023-06-09
node pools: prevent panic on upsert during upgrades

tgross:
Whenever we write a Raft log entry for node pools, we need to first make sure that all servers can safely apply the log without panicking. Gate upsert and delete RPCs on all servers being upgraded to the minimum version.


nomad/leader.go
```diff
@@ -75,6 +75,11 @@ var minACLBindingRuleVersion = version.Must(version.NewVersion("1.5.0-beta.1"))
 // servers must meet before the feature can be used.
 var minNomadServiceRegistrationVersion = version.Must(version.NewVersion("1.3.0"))
 
+// Any writes to node pools requires that all servers are on version 1.6.0 to
+// prevent older versions of the server from crashing.
+// TODO: this is set to 1.5.7 until we ship the beta
```
lgfa29, line 80: Is this because of `version.go`?
https://github.com/hashicorp/nomad/blob/667d0026cde50db770979b0341ac918228cae9ae/version/version.go#L21-L22

I think we can have it as `1.6.0` in `main` (then `1.7.0` when `release/1.6.x` is cut).

lgfa29 (approved):

nomad/leader.go
tgross: That's a good point, at this point there won't be any 1.5.7 releases cut from `main`. Will fix.

shoenig (approved):
LGTM

PR #17476, opened 2023-06-09
replication: fix potential panic during upgrades

tgross:
If the authoritative region has been upgraded to a version of Nomad that has new replicated objects (such as ACL Auth Methods, ACL Binding Rules, etc.), the non-authoritative regions will start replicating those objects as soon as their leader is upgraded. If a server in the non-authoritative region is upgraded and then becomes the leader before all the other servers in the region have been upgraded, then it will attempt to write a Raft log entry that the followers don't understand. The followers will then panic.

Add same the minimum version checks that we do for RPC writes to the leader's replication loop.


nomad/leader.go
```diff
@@ -1848,6 +1848,17 @@ func (s *Server) replicateACLRoles(stopCh chan struct{}) {
 			// parameters are controlled internally.
 			_ = limiter.Wait(context.Background())
 
+			if !ServersMeetMinimumVersion(
+				s.serf.Members(), s.Region(), minACLRoleVersion, true) {
+				s.logger.Trace(
+					"all servers must be upgraded to 1.4.0 before ACL Roles can be replicated")
```
lgfa29, line 1854:
```suggestion
					"all servers must be upgraded to 1.4.0 or later before ACL Roles can be replicated")
```

lgfa29 (approved):

pkazmierczak (approved):

jrasell (approved):
LGTM, thanks!

PR #17479, opened 2023-06-09
node pools: implement `node pool init` command

tgross:
Implement a `nomad node pool init` command that generates an example spec file in either HCL or JSON format.

### DIFF
--- a/command/node_pool_init.go
+++ b/command/node_pool_init.go
@@ -56,7 +56,8 @@ func (c *NodePoolInitCommand) Synopsis() string {
 
 func (c *NodePoolInitCommand) AutocompleteFlags() complete.Flags {
 	return complete.Flags{
-		"-out": complete.PredictSet("hcl", "json"),
+		"-out":   complete.PredictSet("hcl", "json"),
+		"-quiet": complete.PredictNothing,
 	}
 }
 
--- a/command/node_pool_init.go
+++ b/command/node_pool_init.go
@@ -1,0 +1,162 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package command
+
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"strings"
+
+	"github.com/posener/complete"
+)
+
+const (
+	// DefaultHclNodePoolInitName is the default name we use when initializing
+	// the example node pool spec file in HCL format
+	DefaultHclNodePoolInitName = "pool.np.hcl"
+
+	// DefaultJsonNodePoolInitName is the default name we use when initializing
+	// the example node pool spec in JSON format
+	DefaultJsonNodePoolInitName = "pool.np.json"
+)
+
+// NodePoolInitCommand generates a new variable specification
+type NodePoolInitCommand struct {
+	Meta
+}
+
+func (c *NodePoolInitCommand) Help() string {
+	helpText := `
+Usage: nomad node pool init <filename>
+
+  Creates an example node pool specification file that can be used as a starting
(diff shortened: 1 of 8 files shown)


command/node_pool_init.go
```diff
@@ -0,0 +1,162 @@
...
+	"os"
+	"strings"
+
+	"github.com/posener/complete"
+)
+
+const (
+	// DefaultHclNodePoolInitName is the default name we use when initializing
+	// the example node pool spec file in HCL format
+	DefaultHclNodePoolInitName = "pool.np.hcl"
```
lgfa29, line 19:
```suggestion
	DefaultHclNodePoolInitName = "pool.nomad.hcl"
```

I think text editors would prefer if we were consistent with file extension? Wrapping things in blocks then provides a consistent grammar.

command/node_pool_init.go
```diff
@@ -0,0 +1,162 @@
...
+
+func (c *NodePoolInitCommand) AutocompleteArgs() complete.Predictor {
+	return complete.PredictNothing
+}
+
+func (c *NodePoolInitCommand) Name() string { return "node pool init" }
+
+func (c *NodePoolInitCommand) Run(args []string) int {
+	var outFmt string
+	var quiet bool
```
lgfa29, line 67: Was this supposed to be a flag?

command/node_pool_init.go
```diff
@@ -0,0 +1,162 @@
...
+  "Name": "example",
+  "Description": "Example node pool",
+  "Meta": {
+    "environment": "prod",
+    "owner": "sre"
+  },
+  "SchedulerConfiguration": {
+    "SchedulerAlgorithm": "spread"
+  }
+}
```
lgfa29, line 161: Good candidates for `go:embed?` 😄 
https://github.com/hashicorp/nomad/blob/main/command/asset/asset.go

lgfa29 (approved):
Just added some bikeshedding 😄

command/node_pool_init.go
tgross: Makes sense. We didn't do that for variables (or volumes, or quotas, or namespaces... _sigh_ :grinning: ) but if we can do it going forward it'll let us define that grammar as a single "nomad grammar"

command/node_pool_init.go
tgross: :man_facepalming:

command/node_pool_init.go
```diff
@@ -0,0 +1,162 @@
...
+  # configuration as described at:
+  # https://developer.hashicorp.com/nomad/docs/commands/operator/scheduler/set-config
+  #
+  # Available only in Nomad Enterprise.
+  scheduler_configuration {
+
+    # scheduler_algorithm is the scheduling algorithm to use for the pool.
+    # If not defined, the global cluster scheduling algorithm is used.
+    scheduler_algorithm = "spread"
+  }
```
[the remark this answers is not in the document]
tgross: Good call.

command/node_pool_init.go
tgross: Good call. Done.

website/content/docs/commands/node-pool/init.mdx
```diff
@@ -0,0 +1,34 @@
...
+nomad node pool init <filename>
+```
+
+When no filename is supplied, a default filename of "pool.np.hcl" or
+"pool.np.json" will be used depending on the output format.
+
+## Init Options
+
+- `-out` `(enum: hcl | json)`: Format of generated node pool
+  specification. Defaults to `hcl`.
```
lgfa29, line 25: Missing the new `quiet` flag.

PR #17503, opened 2023-06-12
node pool: node pool upsert on multiregion node register

lgfa29:
When registering a node with a new node pool in a non-authoritative region we can't create the node pool because this new pool will not be replicated to other regions.

This commit modifies the node registration logic to only allow automatic node pool creation in the authoritative region.

In non-authoritative regions, the client is registered, but the node pool is not created. The client is kept in the `initialing` status until its node pool is created in the authoritative region and replicated to the client's region.


nomad/fsm_test.go
```diff
@@ -244,143 +244,58 @@ func TestFSM_UpsertNode_Canonicalize_Ineligible(t *testing.T) {
 func TestFSM_UpsertNode_NodePool(t *testing.T) {
```
lgfa29, line 244: I realized that I basically had the same test cases as in `TestStateStore_UpsertNode_NodePool` here, and that seem quite excessive and unnecessary, so I updated this test to only exercise the logic specific to the FSM (node canonicalization and setting the upsert options).

nomad/node_endpoint.go
```diff
@@ -491,15 +509,24 @@ func (n *Node) deregister(args *structs.NodeBatchDeregisterRequest,
...
+//	                │                  ┌────────▼──▼───────┐
+//	                │                  │ Node pool exists? │
+//	                │                  └─────────┬─────────┘
+//	                │                            │
+//	              ready                         Yes
+//	                │                            │
+//	         ┌──────┴───────┐                ┌───▼───┐         ┌──────┐
 //	         │ disconnected ◄─ disconnected ─┤ ready ├─ down ──► down │
 //	         └──────────────┘                └───▲───┘         └──┬───┘
 //	                                             │                │
```
lgfa29, line 532: This is how it's supposed to look like 😅 

```
                ┌────────────────────────────────────── No ───┐
                │                                             │
             ┌──▼───┐          ┌─────────────┐       ┌────────┴────────┐
── Register ─► init ├─ ready ──► Has allocs? ├─ Yes ─► Allocs updated? │
             └──▲───┘          └─────┬───────┘       └────────┬────────┘
                │                    │                        │
                │                    └─ No ─┐  ┌─────── Yes ──┘
                │                           │  │
                │                  ┌────────▼──▼───────┐
                │                  │ Node pool exists? │
                │                  └─────────┬─────────┘
                │                            │
              ready                         Yes
                │                            │
         ┌──────┴───────┐                ┌───▼───┐         ┌──────┐
         │ disconnected ◄─ disconnected ─┤ ready ├─ down ──► down │
         └──────────────┘                └───▲───┘         └──┬───┘
                                             │                │
                                             └──── ready ─────┘
```

nomad/state/state_store.go
```diff
@@ -896,15 +905,17 @@ func (s *StateStore) ScalingEventsByJob(ws memdb.WatchSet, namespace, jobID stri
 // UpsertNode is used to register a node or update a node definition
 // This is assumed to be triggered by the client, so we retain the value
 // of drain/eligibility which is set by the scheduler.
-func (s *StateStore) UpsertNode(msgType structs.MessageType, index uint64, node *structs.Node) error {
+func (s *StateStore) UpsertNode(msgType structs.MessageType, index uint64, node *structs.Node, opts ...NodeUpsertOption) error {
```
lgfa29, line 908: I don't know if a variadic here is the best option. I picked this for a few reasons:
1. `StateStore.UpsertNode()` is called in several places (mostly tests, but still), so we will need to update it everywhere (kind of a lazy argument 😅)
2. Prevent random boolean arguments that may be hard to know what they represent from callers side.

Reasons I don't like it:
1. It's kind of weird since no other function here uses this pattern.
2. It only represents a boolean, so it may be too restrictive to expand this to other use cases.

One alternative implementation I had at some point was to have two methods: `UpserNode()`, which doesn't create a node pool, and `UpsertNodeWithNodePool()` which always creates a node pool if necessary. Then it's up to the caller (the FSM in this case) to call the appropriate function for their use case.

nomad/structs/structs.go
```diff
@@ -581,6 +581,11 @@ type WriteMeta struct {
 type NodeRegisterRequest struct {
 	Node      *Node
 	NodeEvent *NodeEvent
+
+	// CreateNodePool is used to indicate that the node's node pool should be
+	// create along with the node registration if it doesn't exist.
+	CreateNodePool bool
```
lgfa29, line 587: I think defaulting to `false` is the safe option here because it will keep the node in the `initializing` status with the node event as a hint to users.

If this were default to `true` (like, `SkipNodePoolCreation`) we would risk placing a node in a node pool that is not replicated to other regions.

nomad/state/state_store.go
lgfa29: Good points, thanks!

Issue #5887, opened 2019-06-26
Get back reason of killing a task into log


pznamensky:
### Nomad version
Nomad v0.9.3 (c5e8b66c3789e4e7f9a83b4e188e9a937eea43ce)

### Issue
In our case it's very important to know when a task was killed by OOM-killer.
Nomad has metric which indicates that task was killed by OOM, but currently it works not as expected:
- it doesn't increment for a task only for allocation (yes, you can summarize it but task, but see the reason number 2)
- it disappears quickly, so it's not very possible to find out that a task was killed
- it doesn't have initial value (it's a problem for prometheus: https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics)

For now we have to parse nomad logs, find lines like this:
`2019/04/08 06:03:16.608873 [INFO] client: task "ebook-similarities-service-worker" for alloc "c5d0e905-946b-d847-412f-4d727da98ab2" failed: Wait returned exit code 137, signal 0, and error OOM Killed`
and increment counter in prometheus exporter.
It works good, but in v0.9.x I can't find this message in log file.
Could you please get it back?

### Reproduction steps
- execute docker task which is killed by OOM killer
- there is no reason why the task was killed in log files

scalp42:
The weird part is that the UI does show the message though at the task level.

pznamensky:
Is there any news? :)

IvanVlasic:
Are there any news or workarounds?

We also relied on parsing nomad logs for "OOM Killed" but now that doesn't work anymore with newer releases.

kpweiler:
It looks like this issue was addressed several years ago here:
https://github.com/hashicorp/nomad/issues/2203

but the code didn't make it to the plugin refactor:

https://github.com/hashicorp/nomad/blob/20f8227c0a9fa5bfff9414edbcafef0ee455c9a3/drivers/docker/handle.go#L204-L206

The error is handled but not logged.

On a larger note - it's pretty clear to me that "Allocation Status" should be logged somewhere. If the allocation gets GC'd before you have a chance to investigate - you've lost vital information and context.

tgross:
As noted, this is closed by @2203.

scalp42:
@tgross it's still bugged as it won't show in the logs, not sure why you closed the issue.

tgross:
> not sure why you closed the issue.

[...]

PR #17518, opened 2023-06-14
OOM kill logging for Docker driver

vxjohanforssell:
Suggested fix for https://github.com/hashicorp/nomad/issues/5887, with an explicit error log of the docker ID and container image name


drivers/docker/handle.go
```diff
@@ -246,6 +246,8 @@ func (h *taskHandle) run() {
 	if ierr != nil {
 		h.logger.Error("failed to inspect container", "error", ierr)
 	} else if container.State.OOMKilled {
+		h.logger.Error("OOM Killed", "containerID", h.containerID, "containerImage", h.containerImage)
```
pkazmierczak, line 249: our convention is that log context keys should always follow "snake case":

```suggestion
		h.logger.Error("OOM Killed", "container_id", h.containerID, "containerImage", h.containerImage)
```

pkazmierczak (changes requested):
Hey @vxjohanforssell, thanks for the PR! This looks useful to me, even though we emit OOM task events it's useful for administrators to see this in the logs. Especially because [we used to log it in the past](https://github.com/hashicorp/nomad/pull/3459). 

Please have a quick look at one comment I left, and once addressed we can merge.

drivers/docker/handle.go
vxjohanforssell: Of course, I've corrected it!

Also added two additional points of data. It's possible to run the same container with different arguments, where some will trigger OOM and other will not. It would be useful to glean as much info as possible from the log.

drivers/docker/handle.go
```diff
@@ -246,6 +246,8 @@ func (h *taskHandle) run() {
 	if ierr != nil {
 		h.logger.Error("failed to inspect container", "error", ierr)
 	} else if container.State.OOMKilled {
+		h.logger.Error("OOM Killed", "container_id", h.containerID, "container_image", h.containerImage, "docker_command", h.task.Command, "docker_command_args", h.task.Args)
```
pkazmierczak, line 249: Hey @vxjohanforssell, sadly this won't even compile. `*drivers.TaskConfig` has no field or method `Command` or `Args`.

pkazmierczak (changes requested):

drivers/docker/handle.go
vxjohanforssell: I'm sorry, that was a waste of your time! 

I was editing through the web interface and I managed to look at the wrong struct when picking things to print. I've updated it now.

drivers/docker/handle.go
```diff
@@ -246,6 +246,8 @@ func (h *taskHandle) run() {
 	if ierr != nil {
 		h.logger.Error("failed to inspect container", "error", ierr)
 	} else if container.State.OOMKilled {
+		h.logger.Error("OOM Killed", "container_id", h.containerID, "container_image", h.containerImage, "nomad_job_name", h.task.JobName, "nomad_task_name", h.task.Name, "nomad_alloc_id", h.task.AllocID)
```
pkazmierczak, line 249:
```suggestion
		h.logger.Error("OOM Killed",
			"container_id", h.containerID,
			"container_image", h.containerImage,
			"nomad_job_name", h.task.JobName,
			"nomad_task_name", h.task.Name,
			"nomad_alloc_id", h.task.AllocID,
		)
```

pkazmierczak (approved):
Nice one @vxjohanforssell, thanks!

If you could just make the formatting a tiny bit more readable, I'll happily merge.

drivers/docker/handle.go
vxjohanforssell: Done!

PR #17526, opened 2023-06-14
cli: fix missing `-quiet` flag for `var init`

tgross:
The `var init` command was intended to have support for a `-quiet` flag but it was not documented and never parsed.

### DIFF
--- a/command/var_init.go
+++ b/command/var_init.go
@@ -44,6 +44,8 @@ Init Options:
   -out (hcl | json)
     Format of generated variable specification. Defaults to "hcl".
 
+  -quiet
+    Do not print success message.
 `
 	return strings.TrimSpace(helpText)
 }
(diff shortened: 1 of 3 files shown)


command/var_init.go
```diff
@@ -54,7 +54,8 @@ func (c *VarInitCommand) Synopsis() string {
 
 func (c *VarInitCommand) AutocompleteFlags() complete.Flags {
 	return complete.Flags{
-		"-out": complete.PredictSet("hcl", "json"),
+		"-out":   complete.PredictSet("hcl", "json"),
+		"-quiet": complete.PredictNothing,
```
angrycub, line 58: We need to add the flag to the command help text too.

angrycub (approved):
Just one addition. Approving so you can merge once the help text is added.

command/var_init.go
tgross: Oops, thanks!

PR #17528, opened 2023-06-14
node pools: add pool as label on client metrics

tgross:
This changeset adds the node pool as a label anywhere we're already emitting labels with additional information such as node class or ID about the client.


lgfa29 (approved):
I'm glad this wasn't as big and annoying as I thought it would be 😄

PR #17537, opened 2023-06-14
rpc: fix log message in Node.UpdateStatus

lgfa29:
Fix log messages that currently look like this:

```
[DEBUG] nomad.client: marking node as %s due to missing node pool: EXTRA_VALUE_AT_END=initializing
```


tgross (approved):
LGTM! (I hate it when I miss those! :blush: )

PR #17548, opened 2023-06-15
node pools: support `node.pool` constraint in scheduler

tgross:
Although most of the time jobs will be assigned to a single node pool, users may want to set the node pool to "all" and then constraint to a subset of node pools. Add support for setting a contraint like `${node.pool}`.

### DIFF
--- a/scheduler/feasible.go
+++ b/scheduler/feasible.go
@@ -809,6 +809,9 @@ func resolveTarget(target string, node *structs.Node) (string, bool) {
 	case "${node.class}" == target:
 		return node.NodeClass, true
 
+	case "${node.pool}" == target:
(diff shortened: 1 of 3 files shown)


lgfa29 (approved):
I'm not sure why, but I sometimes get an error like this:

```
==> 2023-06-16T11:23:17-04:00: Evaluation "adf6c271" finished with status "complete" but failed to place all allocations:
    2023-06-16T11:23:17-04:00: Task Group "cache" (failed to place 1 allocation):
      * Constraint "computed class ineligible": 1 nodes excluded by filter
```

But I noticed that it happens with other constraints as well so it doesn't seem related to this change.

PR #17475, opened 2023-06-09
[ui] Parallelization down from 4 to 3 and circle workflow for test-ui removed

philrenaud:
Noticing a non-trivial % of test runs on main timing out / running for hours, so reducing the parallelization from 4 to 3 for Ember tests.

Removing Circle test-ui to remove the redundancy and for Percy-run related reasons

PR #17549, opened 2023-06-15
tests: set timeout on test-ui

shoenig:
This seems to finish in about 20 minutes, or run for 6+ hours until hitting
a default timeout. Set a timeout to 30 minutes so we aren't wasting time
and runners.


philrenaud (approved):
Thanks; there's some ongoing work at https://github.com/hashicorp/nomad/pull/17475 to prevent the long runs as well that I'll pick up again shortly.

PR #17558, opened 2023-06-15
e2e: cleanup podman installation in jammy image

shoenig:
The original steps were copied over from the bionic image and does a lot
of hoop jumping we do not need anymore.

For the moment just hard-code installing the v0.4.2 version of the podman driver,
but I may follow up and modify hc-install to support installing `@latest`
like go does.


e2e/terraform/packer/ubuntu-jammy-amd64/setup.sh
```diff
@@ -27,7 +27,10 @@ sudo apt-get upgrade -y
 sudo apt-get install -y \
      software-properties-common \
      dnsmasq unzip tree redis-tools jq curl tmux awscli nfs-common \
-     apt-transport-https ca-certificates gnupg2
+     apt-transport-https ca-certificates gnupg2 golang-go
+
+# Install hc-install
+sudo go install github.com/hashicorp/hc-install/cmd/hc-install@latest
```
tgross, line 33: We _could_ get hc-install from an official release: https://releases.hashicorp.com/hc-install rather than have to ship the toolchain. Not sure if they have a deb in the official repo.

e2e/terraform/packer/ubuntu-jammy-amd64/setup.sh
```diff
@@ -70,49 +73,31 @@ distro=$(lsb_release -si | tr '[:upper:]' '[:lower:]')
...
-
-sudo apt-get update
-
+# Docker
 echo "Installing Docker"
 sudo apt-get install -y docker-ce
 
+# Java
 echo "Installing Java"
+sudo add-apt-repository -y ppa:openjdk-r/ppa
```
tgross, line 82: Did we mean to include this? Should we remove the other JDK if so?

tgross (approved):
LGTM!

e2e/terraform/packer/ubuntu-jammy-amd64/setup.sh
shoenig: Nice, didn't know we had that

e2e/terraform/packer/ubuntu-jammy-amd64/setup.sh
shoenig: Turns out I don't think we need the PPA for this anymore - jammy has openjdk 17 packaged already

PR #17562, opened 2023-06-16
node pools: namespace integration

lgfa29:
This PR adds the necessary fields and structs to support the Enterprise components of node pools. It also modifies job registration to allow an empty node pool value since we need to distinguish between a job that uses the `default` node pool and one that did not specify any.


api/namespace.go
```diff
@@ -70,20 +70,27 @@ func (n *Namespaces) Delete(namespace string, q *WriteOptions) (*WriteMeta, erro
...
+	CreateIndex           uint64
+	ModifyIndex           uint64
 }
 
 type NamespaceCapabilities struct {
 	EnabledTaskDrivers  []string `hcl:"enabled_task_drivers"`
 	DisabledTaskDrivers []string `hcl:"disabled_task_drivers"`
 }
 
+type NamespaceNodePoolConfiguration struct {
```
tgross, line 88: These structs are part of our public Go API so we should have a docstring on them it shows up in pkg.go.dev docs.

nomad/structs/structs_oss_test.go
```diff
@@ -0,0 +1,48 @@
...
+				},
+			},
+			expectedErr: "unlicensed",
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			err := tc.namespace.Validate()
+			fmt.Println(err)
```
tgross, line 40: Debugging leftovers
```suggestion
```

nomad/structs/structs.go
```diff
@@ -5386,6 +5387,26 @@ type NamespaceCapabilities struct {
...
+// namespace.
+type NamespaceNodePoolConfiguration struct {
+	// Default is the node pool used by jobs in this namespace that don't
+	// specify a node pool of their own.
+	Default string
+
+	// Allowed is a list of glob patterns to match node pool names that jobs in
+	// this namespace are allowed to use. If defined, node pools that do not
+	// match any of these patterns are not allowed to be used.
+	// Conflicts with Denied.
```
tgross, line 5400: Minor nitpick: we say "conflicts with denied" here which is a little more vague than the "cannot be used with" that the docs say. It might be nice in these internal docstrings to be clear that they can't be set at the same time (or they get merged, or whatever the behavior ultimately becomes once the ENT work ships).

nomad/structs/structs.go
```diff
@@ -4420,7 +4420,11 @@ type Job struct {
 	// Datacenters contains all the datacenters this job is allowed to span
 	Datacenters []string
 
-	// NodePool specifies the node pool this job is allowed to run on
+	// NodePool specifies the node pool this job is allowed to run on.
+	//
+	// An empty value is allowed during job registration, in which case the
+	// namespace default node pool is used in Enterprise and the 'default' node
+	// pool in OSS, but a node pool must be set before the job is stored.
```
tgross, line 4427: Just to help out future readers:

```suggestion
	// pool in OSS, but a node pool must be set before the job is stored, so
	// that will happen in admission mutators.
```

tgross (approved):
LGTM!

nomad/structs/structs.go
lgfa29: Yeah, I think I wrote these first then wrote a better version in the docs. I will make them the same.

PR #17564, opened 2023-06-16
e2e: modernize podman test suite

shoenig:
Use the new style of e2e test for the podman suite ... which is all of
one test case that was skipped out. Turn the case back on, and we will
add more tests in the near future.


gulducat (approved):
:candy: (sweet suite)

lgfa29 (approved):

PR #17566, opened 2023-06-16
[ui] Job status panel legend should only show "health unknown" for running allocs

philrenaud:
"Health Unknown" no longer counts unplaced, failed, etc. allocations.

image


juliezzhou (approved):
lgtm

PR #17571, opened 2023-06-16
cli: prevent panic if job node pool is nil

lgfa29:
If the `nomad` CLI is used to access a cluster running a version that does not include node pools the command will `nil` panic when trying to resolve the job's node pool.

Thanks @gulducat for the report!


tgross (approved):
Classic. LGTM!

gulducat (approved):
> Node Pool     = \<none>

in `nomad status job-id` output is much better than a panic! :yum:

PR #17572, opened 2023-06-16
e2e: fix windows client docker

gulducat:
The windows docker install script in e2e packer builds has stopped working, sometime in the last week or so.

After trying various things to fix the script, I opted instead for a base image that comes with docker already installed.

I confirmed this works manually from my laptop with `packer build -debug` (ssh-ing in there and running docker commands and nomad agent), and with a little hack in the E2E test suite, example run [here](https://github.com/hashicorp/nomad-e2e/actions/runs/5292693291/jobs/9579872191#step:8:2) (can click through to e2e-runner job there too if you'd like to see the full results).

For posterity, the error output during build that prompted this was:

```
Installing Docker.
WARNING: Cannot find path 'C:\Users\Administrator\AppData\Local\Temp\DockerMsftProvider\DockerDefault_DockerSearchIndex.json' because it does not exist.
WARNING: Cannot bind argument to parameter 'downloadURL' because it is an empty string.
WARNING: The property 'AbsoluteUri' cannot be found on this object. Verify that the property exists.
WARNING: The property 'RequestMessage' cannot be found on this object. Verify that the property exists.
Failed to install Docker.
Install-Package : No match was found for the specified search criteria and package name 'docker'.
```


tgross (approved):
LGTM!

PR #17575, opened 2023-06-17
np: scheduler configuration updates

lgfa29:
This PR makes a few general changes to node pool scheduler configuration:
* Rename the HCL spec to `scheduler_config` to match other configuration blocks.
* Add the `memory_oversubscription_enabled` config.
* Make `scheduler_config` an Enterprise feature.


jrasell (approved):
LGTM!

PR #17576, opened 2023-06-17
Feat: add windows `os.build` attribute to host fingerprint

### LINKED ISSUE #16591: Fingerprint Windows Build Number
As of Nomad 1.5.1 the Windows version is fingerprinted into 2 attributes, but neither is a useful format:

```
kernel.version                           = 10.0.14393.5501 Build 14393.5501
...
os.version                               = 10.0.14393.5501 Build 14393.5501
```

### Proposal

Include just the build number as a new fingerprinted attribute like:

```
os.build = 14393.5501
```

Such that constraints could treat it like a version number for comparisons.

### Use-cases

Nomad 1.5 added the Task API as a unix socket. Windows only added Unix socket support in **build 17063.** In order to schedule tasks that expect the Task API on a version of Windows with unix socket support, we need to fingerprint the build number distinctly.

VishnuJin:
fixes #16591

### DIFF
--- a/client/fingerprint/host.go
+++ b/client/fingerprint/host.go
@@ -5,6 +5,7 @@ package fingerprint
 
 import (
 	"runtime"
+	"strings"
 
 	log "github.com/hashicorp/go-hclog"
 	"github.com/shirou/gopsutil/v3/host"
@@ -29,12 +30,21 @@ func (f *HostFingerprint) Fingerprint(req *FingerprintRequest, resp *Fingerprint
 		return err
 	}
 
+	if runtime.GOOS == "windows" {
+		platformVersion := strings.Split(hostInfo.PlatformVersion, "Build")
+		if len(platformVersion) == 2 {
+			resp.AddAttribute("os.version", strings.TrimSpace(platformVersion[0]))
+			resp.AddAttribute("os.build", strings.TrimSpace(platformVersion[1]))
+		} else {
+			f.logger.Warn("unable to retrieve 'os.build' attribute", "platform_version", hostInfo.PlatformVersion)
+		}
+	} else {
+		resp.AddAttribute("os.version", hostInfo.PlatformVersion)
+		resp.AddAttribute("kernel.version", hostInfo.KernelVersion)
+	}
 	resp.AddAttribute("os.name", hostInfo.Platform)
-	resp.AddAttribute("os.version", hostInfo.PlatformVersion)
-
 	resp.AddAttribute("kernel.name", runtime.GOOS)
 	resp.AddAttribute("kernel.arch", hostInfo.KernelArch)
-	resp.AddAttribute("kernel.version", hostInfo.KernelVersion)
 
 	resp.AddAttribute("unique.hostname", hostInfo.Hostname)
 	resp.Detected = true
--- a/client/fingerprint/host_test.go
+++ b/client/fingerprint/host_test.go
@@ -4,6 +4,7 @@
 package fingerprint
 
 import (
+	"runtime"
 	"testing"
 
 	"github.com/hashicorp/nomad/ci"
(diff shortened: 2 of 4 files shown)


client/fingerprint/host.go
```diff
@@ -28,13 +29,19 @@ func (f *HostFingerprint) Fingerprint(req *FingerprintRequest, resp *Fingerprint
 		f.logger.Warn("error retrieving host information", "error", err)
 		return err
 	}
-
+	print(hostInfo.PlatformVersion)
```
tgross, line 32: Let's remove this debugging debris:
```suggestion
```

client/fingerprint/host.go
```diff
@@ -28,13 +29,19 @@ func (f *HostFingerprint) Fingerprint(req *FingerprintRequest, resp *Fingerprint
...
+
+	if runtime.GOOS == "windows" {
+		platformVersion := strings.Split(hostInfo.PlatformVersion, "Build")
+		buildNumber := strings.TrimSpace(platformVersion[1])
+		resp.AddAttribute("os.build", buildNumber)
```
tgross, line 37: We don't control the input here, so we can't rely 100% on the string "Build" not being changed by upstream. If it changes and we drop the `os.build` attribute, that'll be no big deal but with the code as written here we'd actually panic the client. So we need to check the length of `platformVersion` before accessing `[1]` and trying to assign the attribute.

Also, do we actually want to throw away the leading `10.0` value? Isn't that still a valid value for the `os.version` attribute?

tgross (changes requested):
Hi @VishnuJin! This is looking pretty close.

In addition to the comments I've left, can you do the following:
* Run `make cl` to add a changelog entry.
* Add this new attribute to the table in the [interpolation documentation](https://github.com/hashicorp/nomad/blob/main/website/content/docs/runtime/interpolation.mdx#L118)

Thanks!

client/fingerprint/host.go
VishnuJin: agreed !!

VishnuJin:
@tgross made the necessary changes, please have a look

Thanks

client/fingerprint/host.go
```diff
@@ -29,12 +31,21 @@ func (f *HostFingerprint) Fingerprint(req *FingerprintRequest, resp *Fingerprint
 		return err
 	}
 
+	if runtime.GOOS == "windows" {
+		platformVersion := strings.Split(hostInfo.PlatformVersion, "Build")
+		if len(platformVersion) == 2 {
+			resp.AddAttribute("os.version", strings.TrimSpace(platformVersion[0]))
+			resp.AddAttribute("os.build", strings.TrimSpace(platformVersion[1]))
+		} else {
+			f.logger.Warn(fmt.Sprintf("unable to retrieve 'os.build' attribute %s", platformVersion))
```
tgross, line 40: The hclog library gives us key-value pairs for the extra info we might want to fill out at runtime. And this gives us the original text rather than the post-parsed text:

```suggestion
			f.logger.Warn("unable to retrieve 'os.build' attribute", "platform_version", hostInfo.PlatformVersion))
```

client/fingerprint/host_test.go
```diff
@@ -35,8 +36,17 @@ func TestHostFingerprint(t *testing.T) {
 		t.Fatalf("should generate a diff of node attributes")
 	}
 
+	commonAttributes := []string{"os.name", "unique.hostname", "kernel.name"}
+	nonWindowsAttributes := append(commonAttributes, "os.version")
+	windowsAttributes := append(commonAttributes, "os.build")
```
tgross, line 41: With your most recent change `os.version` should appear for both:

```suggestion
	commonAttributes := []string{"os.name", "os.version", "unique.hostname", "kernel.name"}
	nonWindowsAttributes := append(commonAttributes, "kernel.version")
	windowsAttributes := append(commonAttributes, "os.build")
```

website/content/docs/runtime/interpolation.mdx
```diff
@@ -122,6 +122,7 @@ Below is a table documenting common node properties.
 | `${attr.platform.aws.placement.availability-zone}` | Availability Zone of the client (if on AWS EC2)                                                                                                        |
 | `${attr.os.name}`                                  | Operating system of the client (e.g. `ubuntu`, `windows`, `darwin`)                                                                                    |
 | `${attr.os.version}`                               | Version of the client OS                                                                                                                               |
+| `${attr.os.build}`                                 | Build number for only Windows OS (e.g `14393.5501`)                                                                                                                              |
```
tgross, line 125: A little more consistency with the other attributes:

```suggestion
| `${attr.os.build}`                                 | Build number (e.g `14393.5501`) of the client OS (if on Windows)                                                                                                                          
|
```

tgross (changes requested):
Almost there!

VishnuJin:
@tgross I’ve made the necessary changes again, please have a look

VishnuJin:
@tgross this my first ever attempt in go, thanks for the kind feedback and guidance.
super happy 😄

PR #17578, opened 2023-06-17
Add "nomad.client.host.cpu.total_ticks" to metrics

TrueBrain:
Currently there is `nomad.client.allocs.cpu.total_ticks` and `nomad.client.allocs.cpu.total_percent` for allocations, but `nomad.client.host.cpu.total` only has that, which is a percentage.

To keep things more the same, this PR adds `total_ticks` and `total_percent` for `nomad.client.host.cpu` too. This means that all `cpu` objects have at least `system`, `user`, `total_percent` and `total_ticks`.

I also kept `nomad.client.host.cpu.total`, as I am sure there are people using this field; and removing it would break their work. Which seems like a bad idea. One could argue that `total_percent` doesn't need to be added, but this makes it a bit unclear what `total` actually is.

PS: honestly not sure this PR is worth it; my OCD triggered, that was the whole reason I have to admit :)


tgross:
@TrueBrain I'm going to close this PR in lieu of keeping this work together with what you submitted in https://github.com/hashicorp/nomad/pull/17579 (which I'm reviewing now). I do appreciate that you tried to break it up into reviewable chunks! But I think these two PRs belong together and I want to address them as one.

PR #17579, opened 2023-06-17
Add "total_ticks_count" for CPU metrics

TrueBrain:
This counter tells you the total amount of ticks for that CPU
entry since the start of Nomad.

Let me give you a bit of context why I always get a bit annoyed when there isn't a CPU counter available in metrics.

Once upon a day, when I was a small boy, I had my Nagios telling me all my servers were running at 100% CPU. But whenever I logged in, they never were above 1%. But the alarms kept going, and the frustration got high. So what was the problem?

When Nagios kicks off, it starts to call a lot of scripts on a host. The last one of the series was: measure current CPU percentage. And well .. because a fraction early other measurement scripts kicked in, at that exact moment in time, the CPU was always 100%. But that was not actually what I wanted to know .. I wanted to know the average percentage since last measurement.

Nomad does this pretty well: it calculates the percentage for all drivers I checked based on the last time it looked at that information. So percentages are average, and they avoid the above issue. But, kinda.

When pushing these averages to the metrics, they are pushed as a guage. This means I am now obligated to pull the metrics as often as Nomad pushes them, to make sure I have an accurate view of the CPU usage. But that is unlikely to work out, as timers drift, and anything polling will miss a sample from time to time.

So as I found out back in the days with Nagios, the far better way to represent CPU time, is in ticks since start, as counter. That way, who-ever reads the metric can figure out themselves what the delta was since last poll, and with that the average CPU usage. In ticks in this case, but I can imagine one could also argue the case for milliseconds.

Basically, this brings what Nomad is doing internally also external, so other tools can do the same thing as Nomad is doing internally: measure the ticks whenever, and calculate based on the time-delta what the average was.

I went for the name `counter_ticks`, although I would rather have changed `total_ticks` into a Counter .. but changing the meaning of an existing measurement is in general not the nicest thing to do :D

PS: I guess "ticks" is a bit vague, as it means MHz in this case. But it was already called "ticks" everywhere, so I stayed with that name. But `counter_mhz` might be more descriptive.


website/content/docs/operations/metrics-reference.mdx
```diff
@@ -191,6 +193,7 @@ task driver; not all task drivers can provide all metrics.
 |-----------------------------------------------|-------------------------------------------------------------------|-------------|---------|--------------------------------------------------|
 | `nomad.client.allocs.complete`                | Number of complete allocations                                    | Integer     | Counter | alloc_id, host, job, namespace, task, task_group |
 | `nomad.client.allocs.cpu.allocated`           | Total CPU resources allocated by the task across all cores        | MHz         | Gauge   | alloc_id, host, job, namespace, task, task_group |
+| `nomad.client.allocs.cpu.counter_ticks`       | CPU ticks consumed by the process since startup                   | Integer     | Counter | alloc_id, host, job, namespace, task, task_group |
```
tgross, line 196: A task can have multiple processes, so this is a little more accurate. We'll want to move this to the right spot in alphabetical order as well for the name.

```suggestion
| `nomad.client.allocs.cpu.total_ticks_count`       | Total CPU ticks consumed by the task since startup                   | Integer     | Counter | alloc_id, host, job, namespace, task, task_group |
```

client/client.go
```diff
@@ -3039,10 +3039,15 @@ func (c *Client) setGaugeForCPUStats(nodeID string, hStats *stats.HostStats, bas
 			Value: cpu.CPU,
 		})
 
-		metrics.SetGaugeWithLabels([]string{"client", "host", "cpu", "total"}, float32(cpu.Total), labels)
+		// Keep "total" around to remain compatible with older consumers of the metrics
+		metrics.SetGaugeWithLabels([]string{"client", "host", "cpu", "total"}, float32(cpu.TotalPercent), labels)
```
tgross, line 3043: We do (rarely) drop metrics, but we try to have a deprecation window of 1 major release. Once this is merged I'll add an upgrade note marking the metric deprecated for whatever release it'll ship in. (Likely the upcoming 1.6.0.)

client/client.go
```diff
@@ -3039,10 +3039,15 @@ func (c *Client) setGaugeForCPUStats(nodeID string, hStats *stats.HostStats, bas
...
+		metrics.SetGaugeWithLabels([]string{"client", "host", "cpu", "total_percent"}, float32(cpu.TotalPercent), labels)
+		metrics.SetGaugeWithLabels([]string{"client", "host", "cpu", "total_ticks"}, float32(cpu.TotalTicks), labels)
 		metrics.SetGaugeWithLabels([]string{"client", "host", "cpu", "user"}, float32(cpu.User), labels)
 		metrics.SetGaugeWithLabels([]string{"client", "host", "cpu", "idle"}, float32(cpu.Idle), labels)
 		metrics.SetGaugeWithLabels([]string{"client", "host", "cpu", "system"}, float32(cpu.System), labels)
+
+		metrics.IncrCounterWithLabels([]string{"client", "host", "cpu", "counter_ticks"}, float32(cpu.TotalTicks), labels)
```
tgross, line 3050: The name "total" for the `total_ticks` and `total_percent` is unfortunate because it's the "total" as in "across all types of CPU resources" whereas this is also a counter where "total" means "across all time". To make the relationship between the gauge and the counter clear, what do you think about naming this `total_ticks_count` like so:

```suggestion
		metrics.IncrCounterWithLabels([]string{"client", "host", "cpu", "total_ticks_count"}, float32(cpu.TotalTicks), labels)
```

That'll make `client.host.cpu.total_ticks` and `client.host.cpu.total_ticks_count` fall together in all the documentation.

(Same suggestion applies everywhere)

tgross (changes requested):
Hi @TrueBrain and thanks for the PR!

I've left some comments to address. Additionally, if you run `make cl` it'll create a changelog entry.

client/client.go
TrueBrain: That is an excellent idea, tnx; I was struggling how to do it properly :)

client/client.go
TrueBrain: Excellent idea :)

TrueBrain:
Just checking in: is there anything else needed to move this PR forward? If so, please do let me know! Tnx :)

tgross (approved):
LGTM!

tgross:
Docs look good: https://nomad-a8xmfz7gg-hashicorp.vercel.app/nomad/docs/operations/metrics-reference#client-metrics

PR #17591, opened 2023-06-19
core: use faster concatenation for alloc name generation.

jrasell:
I was poking around in this code, so thought I would experiment quickly with an improvement to the name generation. I'm not strongly in favour of this, as it makes reading slightly more difficult, but thought I'd raise it to be opinions anyway.

benchmark:
```
goos: darwin
goarch: arm64
pkg: jrasell-experiments/benchmarks
BenchmarkAllocNameOriginal-12    	 9239590	       111.2 ns/op	      80 B/op	       3 allocs/op
BenchmarkAllocNameNew-12         	34172659	        35.10 ns/op	      48 B/op	       1 allocs/op
```

### DIFF
--- a/nomad/structs/funcs.go
+++ b/nomad/structs/funcs.go
@@ -382,7 +382,7 @@ func DenormalizeAllocationJobs(job *Job, allocs []*Allocation) {
 
 // AllocName returns the name of the allocation given the input.
 func AllocName(job, group string, idx uint) string {
-	return fmt.Sprintf("%s.%s[%d]", job, group, idx)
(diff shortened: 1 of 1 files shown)


tgross (approved):
I guess the stdlib's string formatter's generality ends up being more expensive than I'd have thought. Looks like a nice encapsulated change. LGTM! :shipit:

shoenig:
https://github.com/hashicorp/go-uuid/pull/14

had a similar optimization that got ignored; the 7 -> 2 allocs per ID is probably worth the little extra complexity

schmichael (approved):

PR #8691, opened 2020-08-19
Respect alloc job version for lost/failed allocs

notnoop:
This change fixes a bug where lost/failed allocations are replaced by
allocations with the latest versions, even if the version hasn't been
promoted yet.

Now, when generating a plan for lost/failed allocations, the scheduler
first checks if the current deployment is in Canary stage, and if so, it
ensures that any lost/failed allocations is replaced one with the latest
promoted version instead.

## Implementation High Level

In high level, the fix makes the following changes:

1. First, when rescheduling (or migrating) non-canary allocations, the reconciler marks the resulting `allocPlaceResult` struct with downgrade indicator along with the minimum version
  * the minimum version is used to ensure that no allocations is accidentally rollback when rescheduling
2. When making the placement and computing resources, the generic scheduler finds the latest non-canary deployment associated with the task group and uses it for scheduling purposes
3. When inserting the resulting alloc in the plan, we ensure that the job info is attached to the alloc.
  * Typically, we don't populate Allocation.Job field in the plan, as it's typically duplicated, and nomad FSM ensures that the alloc.Job field is populated from plan.  Here, we want the alloc to differ from the plan one
4. When applying the plan results to the FSM, nomad will respect the alloc.Job field and avoid populating it
  * This is already current behavior, no changes there, but the test ensures that's the case

### Why use latest promoted or non-canary version?  Why not latest job Stable version?

The promotion semantics is somewhat group specific.  Consider a job has two TaskGroups: A, B in Version 0, and then Version 1 updated A significantly (requiring a Canary) but only updated B's count and metadata.  B allocations should always use Version 1, and non-Canary A should use version 0.  Using latest Stable job will not appropriate here

### In this approach, does the scheduler respect the count and resources found in latest promoted deployment?

The scheduler will use the resources of the appropriate job version for the alloc, if the resources changed. However, the latest job Count field is the canonical value, and this PR doesn't change that.  If a job has TG count=3 in Version 0, and changed count to 2 and requested large resources; immediately, a Version 0 allocation will be stopped, so we'd have 2 Version=0 allocations along with some canaries.

### Can this PR handle cases where we reschedule a Canary and NonCanary failed allocations

Yes, when migrating allocations, the canary will be replaced by another canary, and non-canaries will be replaced by non-canary instances, each with the expected versions for them.

Fixes  #8439

PR #17598, opened 2023-06-19
node pools: apply node pool scheduler configuration

lgfa29:
Merge global scheduler configuration with the job's node pool scheduler configuration when processing nodes.

### DIFF
--- a/helper/iterator/iterator.go
+++ b/helper/iterator/iterator.go
@@ -1,0 +1,23 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package iterator
+
+// Iterator represents an object that can iterate over a set of values one at a
+// time.
+type Iterator interface {
+	// Next returns the next element or nil if there are none left.
+	Next() any
+}
+
+// Len consumes the iterator and returns the number of elements found.
+//
+// IMPORTANT: this method consumes the iterator, so it should not be used after
+// Len() returns.
+func Len(iter Iterator) int {
+	count := 0
+	for raw := iter.Next(); raw != nil; raw = iter.Next() {
+		count++
+	}
+	return count
+}
--- a/scheduler/generic_sched.go
+++ b/scheduler/generic_sched.go
@@ -282,6 +282,18 @@ func (s *GenericScheduler) process() (bool, error) {
 	s.stack = NewGenericStack(s.batch, s.ctx)
 	if !s.job.Stopped() {
 		s.stack.SetJob(s.job)
+
+		// Fetch node pool and global scheduler configuration to determine how
+		// to configure the scheduler.
+		pool, err := s.state.NodePoolByName(ws, s.job.NodePool)
+		if err != nil {
+			return false, fmt.Errorf("failed to get job node pool '%s': %v", s.job.NodePool, err)
+		}
+		_, schedConfig, err := s.state.SchedulerConfig()
+		if err != nil {
+			return false, fmt.Errorf("failed to get scheduler configuration: %v", err)
+		}
+		s.stack.SetSchedulerConfiguration(schedConfig.WithNodePool(pool))
 	}
 
 	// Compute the target job allocations
--- a/scheduler/generic_sched.go
+++ b/scheduler/generic_sched.go
@@ -13,6 +13,7 @@ import (
 	"github.com/hashicorp/go-memdb"
 	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/go-version"
+	"github.com/hashicorp/nomad/helper"
 	"github.com/hashicorp/nomad/helper/uuid"
 	"github.com/hashicorp/nomad/nomad/structs"
 )
@@ -281,19 +282,7 @@ func (s *GenericScheduler) process() (bool, error) {
 	// Construct the placement stack
 	s.stack = NewGenericStack(s.batch, s.ctx)
 	if !s.job.Stopped() {
-		s.stack.SetJob(s.job)
-
-		// Fetch node pool and global scheduler configuration to determine how
-		// to configure the scheduler.
-		pool, err := s.state.NodePoolByName(ws, s.job.NodePool)
-		if err != nil {
-			return false, fmt.Errorf("failed to get job node pool '%s': %v", s.job.NodePool, err)
-		}
-		_, schedConfig, err := s.state.SchedulerConfig()
-		if err != nil {
-			return false, fmt.Errorf("failed to get scheduler configuration: %v", err)
-		}
-		s.stack.SetSchedulerConfiguration(schedConfig.WithNodePool(pool))
+		s.setJob(s.job)
 	}
 
 	// Compute the target job allocations
@@ -521,7 +510,7 @@ func (s *GenericScheduler) downgradedJobForPlacement(p placementResult) (string,
 // destructive updates to place and the set of new placements to place.
 func (s *GenericScheduler) computePlacements(destructive, place []placementResult) error {
 	// Get the base nodes
-	nodes, _, byDC, err := readyNodesInDCsAndPool(s.state, s.job.Datacenters, s.job.NodePool)
+	nodes, byDC, err := s.setNodes(s.job)
 	if err != nil {
 		return err
 	}
@@ -574,10 +563,17 @@ func (s *GenericScheduler) computePlacements(destructive, place []placementResul
 				continue
 			}
 
-			// Use downgraded job in scheduling stack to honor
-			// old job resources and constraints
+			// Use downgraded job in scheduling stack to honor old job
+			// resources, constraints, and node pool scheduler configuration.
 			if downgradedJob != nil {
-				s.stack.SetJob(downgradedJob)
+				s.setJob(downgradedJob)
+
+				if needsToSetNodes(downgradedJob, s.job) {
+					nodes, byDC, err = s.setNodes(downgradedJob)
+					if err != nil {
+						return err
+					}
+				}
 			}
 
 			// Find the preferred node
@@ -608,9 +604,17 @@ func (s *GenericScheduler) computePlacements(destructive, place []placementResul
 			// Compute top K scoring node metadata
 			s.ctx.Metrics().PopulateScoreMetaData()
 
-			// Restore stack job now that placement is done, to use plan job version
+			// Restore stack job and nodes now that placement is done, to use
+			// plan job version
 			if downgradedJob != nil {
-				s.stack.SetJob(s.job)
+				s.setJob(s.job)
+
+				if needsToSetNodes(downgradedJob, s.job) {
+					nodes, byDC, err = s.setNodes(s.job)
+					if err != nil {
+						return err
+					}
+				}
 			}
 
 			// Set fields based on if we found an allocation option
@@ -700,6 +704,45 @@ func (s *GenericScheduler) computePlacements(destructive, place []placementResul
 	}
 
 	return nil
+}
+
+// setJob updates the stack with the given job and job's node pool scheduler
+// configuration.
+func (s *GenericScheduler) setJob(job *structs.Job) error {
+	// Fetch node pool and global scheduler configuration to determine how to
+	// configure the scheduler.
+	pool, err := s.state.NodePoolByName(nil, job.NodePool)
+	if err != nil {
+		return fmt.Errorf("failed to get job node pool %q: %v", job.NodePool, err)
+	}
+
+	_, schedConfig, err := s.state.SchedulerConfig()
+	if err != nil {
+		return fmt.Errorf("failed to get scheduler configuration: %v", err)
+	}
+
+	s.stack.SetJob(job)
+	s.stack.SetSchedulerConfiguration(schedConfig.WithNodePool(pool))
+	return nil
+}
+
+// setnodes updates the stack with the nodes that are ready for placement for
+// the given job.
+func (s *GenericScheduler) setNodes(job *structs.Job) ([]*structs.Node, map[string]int, error) {
+	nodes, _, byDC, err := readyNodesInDCsAndPool(s.state, job.Datacenters, job.NodePool)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	s.stack.SetNodes(nodes)
+	return nodes, byDC, nil
+}
+
+// needsToSetNodes returns true if jobs a and b changed in a way that requires
+// the nodes to be reset.
+func needsToSetNodes(a, b *structs.Job) bool {
+	return !helper.SliceSetEq(a.Datacenters, b.Datacenters) ||
+		a.NodePool != b.NodePool
 }
 
 // propagateTaskState copies task handles from previous allocations to
--- a/scheduler/generic_sched.go
+++ b/scheduler/generic_sched.go
@@ -873,6 +873,11 @@ func (s *GenericScheduler) selectNextOption(tg *structs.TaskGroup, selectOptions
 	_, schedConfig, _ := s.ctx.State().SchedulerConfig()
 
 	// Check if preemption is enabled, defaults to true
+	//
+	// The scheduler configuration is read directly from state but only
+	// values that can't be specified per node pool should be used. Other
+	// values must be merged by calling schedConfig.WithNodePool() and set in
+	// the stack by calling SetSchedulerConfiguration().
 	enablePreemption := true
 	if schedConfig != nil {
 		if s.job.Type == structs.JobTypeBatch {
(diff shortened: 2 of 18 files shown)


helper/iterator/iterator.go
```diff
@@ -0,0 +1,23 @@
...
+type Iterator interface {
+	// Next returns the next element or nil if there are none left.
+	Next() any
+}
+
+// Len consumes the iterator and returns the number of elements found.
+//
+// IMPORTANT: this method consumes the iterator, so it should not be used after
+// Len() returns.
+func Len(iter Iterator) int {
```
tgross, line 17: At some point I'd love to see if we can get something like this done upstream in go-memdb (non-consuming if possible!), but it hasn't quite made it to the top of my todo pile :grinning:

scheduler/generic_sched.go
```diff
@@ -282,6 +282,18 @@ func (s *GenericScheduler) process() (bool, error) {
...
+		// to configure the scheduler.
+		pool, err := s.state.NodePoolByName(ws, s.job.NodePool)
+		if err != nil {
+			return false, fmt.Errorf("failed to get job node pool '%s': %v", s.job.NodePool, err)
+		}
+		_, schedConfig, err := s.state.SchedulerConfig()
+		if err != nil {
+			return false, fmt.Errorf("failed to get scheduler configuration: %v", err)
+		}
+		s.stack.SetSchedulerConfiguration(schedConfig.WithNodePool(pool))
```
tgross, line 296: There are two other places in the `GenericScheduler` where we call `s.stack.SetJob` with the previous version of the job as part of computing placements, and then restore it later. This will miss all those cases which will break if the job is moved between node pools. 

Because the scheduler config is now effectively tied to to the node pool and therefore a specific version of the job, we should probably move the call to `SetSchedulerConfiguration` inside of `SetJob` so that they're always in sync.

scheduler/generic_sched.go
```diff
@@ -282,6 +282,18 @@ func (s *GenericScheduler) process() (bool, error) {
 	s.stack = NewGenericStack(s.batch, s.ctx)
 	if !s.job.Stopped() {
 		s.stack.SetJob(s.job)
+
+		// Fetch node pool and global scheduler configuration to determine how
+		// to configure the scheduler.
+		pool, err := s.state.NodePoolByName(ws, s.job.NodePool)
+		if err != nil {
+			return false, fmt.Errorf("failed to get job node pool '%s': %v", s.job.NodePool, err)
```
shoenig, line 290: use `%q` instead of `'%s'`

scheduler/scheduler_system.go
```diff
@@ -157,6 +157,18 @@ func (s *SystemScheduler) process() (bool, error) {
 	s.stack = NewSystemStack(s.sysbatch, s.ctx)
 	if !s.job.Stopped() {
 		s.stack.SetJob(s.job)
+
+		// Fetch node pool and global scheduler configuration to determine
+		// which values to use.
+		pool, err := s.state.NodePoolByName(ws, s.job.NodePool)
+		if err != nil {
+			return false, fmt.Errorf("failed to get job node pool '%s': %v", s.job.NodePool, err)
```
shoenig, line 165: same

tgross (changes requested):
This looks really good @lgfa29! Just the problem with the previous-job version move to fix I think

shoenig (approved):
LGTM!

scheduler/generic_sched.go
lgfa29: I was trying to write a test that would cover this scenario but the other calls to `SetJob` were never triggered, so I looked for tests that did and the only two I found were [`TestServiceSched_Migrate_CanaryStatus`](https://github.com/hashicorp/nomad/blob/main/scheduler/generic_sched_test.go#L5895) and [`TestSpreadPanicDowngrade`](https://github.com/hashicorp/nomad/blob/main/scheduler/spread_test.go#LL932).

Copying the structure of `TestSpreadPanicDowngrade` I was able to trigger the `downgradedJob ! nil` path but I didn't see any difference between updating the scheduler configuration or not.

https://github.com/hashicorp/nomad/pull/17598/commits/de31ec2f019302e7279f10b14a101e22841b332c has the change to always update the scheduler config along with the job.

lgfa29: Uploaded the test function here just in case: https://gist.github.com/lgfa29/76aa40859e8cc808a81e6bc32f3f61fa

helper/iterator/iterator.go
lgfa29: I could swear there was one already but I couldn't find it 😅 

Not consuming the iterator would be tricky. I don't think we can create a deep copy or reset the iterator somehow.

helper/iterator/iterator.go
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package iterator
+
+// Iterator represents an object that can iterate over a set of values one at a
+// time.
+type Iterator interface {
```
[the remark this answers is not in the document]
lgfa29: > Googling how to use generic....

scheduler/generic_sched.go
tgross: > but I didn't see any difference between updating the scheduler configuration or not.

Oh! It looks like that code path calls the `selectNextOption` method, which is querying the state store for the scheduler config again (ref [`generic_sched.go#L814`](https://github.com/hashicorp/nomad/blob/v1.5.6/scheduler/generic_sched.go#L814)). So it's completely ignoring that we've set the scheduler config in the stack earlier. We should make sure we're not querying the state directly anywhere else too.

scheduler/stack.go
```diff
@@ -286,7 +293,7 @@ func NewSystemStack(sysbatch bool, ctx Context) *SystemStack {
 	}
```
tgross, line 293: This block is using the scheduler config from the state, not the one we've set earlier. It's not a bug right now because we don't have preemption in the node pool scheduler config, but we'll need to address this when we do anyways so if it's not a huge lift now we might want to do that too. (Or just move finding `enablePreemption` into the `NewBinPackIterator`.)

scheduler/scheduler_test.go
```diff
@@ -0,0 +1,222 @@
...
+	cases := []struct {
+		name                          string
+		nodePool                      string
+		cpu                           int
+		memory                        int
+		memoryMax                     int
+		memoryOversubscriptionEnabled bool
+
+		expectedTaskMemoryMax int
+		// expectedTotalMemoryMax should be SUM(MAX(memory, memoryMax)) for all tasks
```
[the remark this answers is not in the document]
lgfa29: Oh right, I copied this https://github.com/hashicorp/nomad/pull/17598/files#diff-4879baf9e9a85d036c73ec0a96cb6a16895bdc7cc9b776f4caa6efe8627178cbL127 but didn't pay attention to the comment 😅

helper/iterator/iterator.go
tgross: Yeah I think you're right; the `ResultIterator` interface always bottoms out in the iradix library, but the results are filtered before they get to the caller so we couldn't use some count object on the radix trie to help us anyways. :shrug:

helper/iterator/iterator.go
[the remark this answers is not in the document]
lgfa29: I gave it a few attempts but couldn't get this to work in a way that would accept both what we current have (`Next() any` and `Next() interface{}`) and what we would like to have (`Next() *ActualType`). So I'm leaving it as is for now.

For the record, this is more or less what I was trying:

```go
// Iterator represents an object that can iterate over a set of values one at a
// time.
type Iterator[T any] interface {
	// Next returns the next element or nil if there are none left.
	Next() any
}

// Len consumes the iterator and returns the number of elements found.
//
// IMPORTANT: this method consumes the iterator, so it should not be used after
// Len() returns.
func Len[T any](iter Iterator[*T]) int {
	count := 0
	for raw := iter.Next(); raw != nil; raw = iter.Next() {
		count++
	}
	return count
}
```

scheduler/generic_sched.go
lgfa29: It made no difference because I was always setting the same job 🤦 
https://github.com/hashicorp/nomad/pull/17598/commits/de31ec2f019302e7279f10b14a101e22841b332c#diff-7da37ffb89a2df3f8fea4a048ca3bb21b799df290033c26766743e60bc217a48R708


I now have a difference in the node scores, which makes sense, but I still don't fully understand what's going on, and I think we may have a bug if node pools or datacenter changes?

Before the fix, the resulting plan looked like this:
https://gist.github.com/lgfa29/36ef5ba7ac4c7e2c798bfe29f047ec97#file-before_fix-txt

It's very hard to understand what's going on but the main part is that there are two `NodeAllocation`, one for the first job version in the node pool that uses binpacking and one for the updated job in the node pool with spread.

The scores for the alloc in the node in the binpacking pool look like this:
```
     ScoreMetaData: ([]*structs.NodeScoreMeta) (len=3 cap=3) {
      (*structs.NodeScoreMeta)(0x1400029fd40)(2a5436c6-6311-5ff3-c454-a6ca027bbd95 0.818219 map[binpack:0.8182193588025797 job-anti-affinity:0 node-affinity:0 node-reschedule-penalty:0]),
      (*structs.NodeScoreMeta)(0x1400029fcc0)(3df6cdd1-39fb-7854-513b-5251e83ebdf4 0.818219 map[binpack:0.8182193588025797 job-anti-affinity:0 node-affinity:0 node-reschedule-penalty:0]),
      (*structs.NodeScoreMeta)(0x1400029fdc0)(e67e2a90-9c39-75b9-3ea5-b83268ad3fc9 0.137793 map[binpack:0.6755864441534736 job-anti-affinity:-0.4 node-affinity:0 node-reschedule-penalty:0])
     },
```

Which I think it's the main bug because these values are quite high for essentially empty nodes, which indicates the spread algorithm was used to score them.

And we can see that by comparing them with the scores for the alloc in the node in the spread pool:

```
     ScoreMetaData: ([]*structs.NodeScoreMeta) (len=3 cap=3) {
      (*structs.NodeScoreMeta)(0x1400029fb80)(f02fca69-2de8-f319-e1a4-315c27bbd982 0.818219 map[binpack:0.8182193588025797 job-anti-affinity:0 node-affinity:0 node-reschedule-penalty:0]),
      (*structs.NodeScoreMeta)(0x1400029fc00)(08f7edaa-7a78-e0f8-96d1-7295b0daf003 0.818219 map[binpack:0.8182193588025797 job-anti-affinity:0 node-affinity:0 node-reschedule-penalty:0]),
      (*structs.NodeScoreMeta)(0x1400029fb00)(e67e2a90-9c39-75b9-3ea5-b83268ad3fc9 0.818219 map[binpack:0.8182193588025797 job-anti-affinity:0 node-affinity:0 node-reschedule-penalty:0])
     },
```

After code change things look better. The alloc in the binpack node looks like this:

```
     ScoreMetaData: ([]*structs.NodeScoreMeta) (len=4 cap=4) {
      (*structs.NodeScoreMeta)(0x140002cd760)(07567ea6-6c0a-1cb5-4984-26250bb27bca 0.181781 map[binpack:0.18178064119742032 job-anti-affinity:0 node-affinity:0 node-reschedule-penalty:0]),
      (*structs.NodeScoreMeta)(0x140002cd6e0)(4b06224f-ac54-c922-3245-0d732555b12f 0.181781 map[binpack:0.18178064119742032 job-anti-affinity:0 node-affinity:0 node-reschedule-penalty:0]),
      (*structs.NodeScoreMeta)(0x140002cd840)(7f96eb7a-cb60-5040-b722-7b6cd7193898 0.181781 map[binpack:0.18178064119742032 job-anti-affinity:0 node-affinity:0 node-reschedule-penalty:0]),
      (*structs.NodeScoreMeta)(0x140002cd7e0)(c18cd236-b02a-7118-d800-6193b7ae839a -0.037793 map[binpack:0.3244135558465265 job-anti-affinity:-0.4 node-affinity:0 node-reschedule-penalty:0])
     },
```

And the alloc in the spread node looks like this:

```
     ScoreMetaData: ([]*structs.NodeScoreMeta) (len=3 cap=3) {
      (*structs.NodeScoreMeta)(0x140002cd5a0)(7f96eb7a-cb60-5040-b722-7b6cd7193898 0.818219 map[binpack:0.8182193588025797 job-anti-affinity:0 node-affinity:0 node-reschedule-penalty:0]),
      (*structs.NodeScoreMeta)(0x140002cd620)(ebb9fade-cd87-cbbb-0229-e268f658789d 0.818219 map[binpack:0.8182193588025797 job-anti-affinity:0 node-affinity:0 node-reschedule-penalty:0]),
      (*structs.NodeScoreMeta)(0x140002cd520)(c18cd236-b02a-7118-d800-6193b7ae839a 0.818219 map[binpack:0.8182193588025797 job-anti-affinity:0 node-affinity:0 node-reschedule-penalty:0])
     },
```

Now, why are there four nodes in the binpack alloc? I'm still not sure, but while looking into it I realized that both sets of metrics had shared node IDs, which makes no sense since each alloc is for a job in a different node pool!

I think the problem is that we only call `SetNodes()` once at the beginning of the function, but we need to get the nodes again for the `downgradedJob` if the node pool or datacenter changed.

lgfa29: Reading https://github.com/hashicorp/nomad/pull/8691 and https://github.com/hashicorp/nomad/commit/c49359ad58f0af18a5697a0b7b9b6cca9656d267 helped me understand what this part of the code is covering.

I rebased this branch to squash the commits into a single commit with the (I hope 😅) proper fix: https://github.com/hashicorp/nomad/pull/17598/commits/b99e8c5781089655e607c346013282b4cfe3f87e.

I think the problem with nodes not being reset also affects current versions of Nomad when `datacenters` change. Since this code is a bit different in `main` because of node pools I will create fix for the release branches manually.

scheduler/stack.go
lgfa29: This, along with the part mentioned in https://github.com/hashicorp/nomad/pull/17598#discussion_r1235768274, are the preemption bits that I was reluctant to touch right now. 

Despite reading the full scheduler configuration, only the preemption bits are used, so, as you mentioned, it doesn't affects us now.

Fixing this is a bit awkward because the job (and therefore the node pool) is only known once `SetJob` is called, and so the [`evict`](https://github.com/hashicorp/nomad/blob/main/scheduler/rank.go#L178) flag in the `BinPackIterator` iterator could be modified mid-scheduling, and I don't have the full grasp of what this could cause.

Maybe for now we can add a comment here and in `selectNextOption` noting that only preemption values should be used because the node pool may override the others?

scheduler/scheduler_system.go
```diff
@@ -157,6 +157,18 @@ func (s *SystemScheduler) process() (bool, error) {
...
+		// which values to use.
+		pool, err := s.state.NodePoolByName(ws, s.job.NodePool)
+		if err != nil {
+			return false, fmt.Errorf("failed to get job node pool %q: %v", s.job.NodePool, err)
+		}
+		_, schedConfig, err := s.state.SchedulerConfig()
+		if err != nil {
+			return false, fmt.Errorf("failed to get scheduler configuration: %v", err)
+		}
+		s.stack.SetSchedulerConfiguration(schedConfig.WithNodePool(pool))
```
tgross, line 171: We're not taking the same approach as we did in the generic scheduler because this is the only caller of `SetJob` and `SchedulerConfiguration`, right?

scheduler/stack.go
tgross: > Maybe for now we can add a comment here and in `selectNextOption` noting that only preemption values should be used because the node pool may override the others?

Yeah, that seems reasonable for keeping the risk of this change contained.

tgross (approved):
LGTM! :shipit:

scheduler/generic_sched.go
tgross: Nice detective work there. Sounds like a plan!

scheduler/scheduler_system.go
lgfa29: That's right, although it may be nice to keep them consistent, so I think I will follow the same pattern here 👍

scheduler/generic_sched.go
lgfa29: I got an actual reproduction and described it in https://github.com/hashicorp/nomad/issues/17651

scheduler/scheduler_system.go
lgfa29: Updated in https://github.com/hashicorp/nomad/pull/17598/commits/742263623d034bf2ca6b3e56bc5f67292200cc6b

scheduler/stack.go
lgfa29: Added some comments in https://github.com/hashicorp/nomad/pull/17598/commits/6b07aedcaa56233bcc4e41f01c169a9b1f770f00

scheduler/generic_sched.go
lgfa29: Manual backports created as well:
https://github.com/hashicorp/nomad/pull/17652
https://github.com/hashicorp/nomad/pull/17653
https://github.com/hashicorp/nomad/pull/17654

PR #17599, opened 2023-06-19
[ui] General status for steady-state jobs

philrenaud:
An abstracted status to indicate how a job is doing, based on its allocations' clientStatus, compared to what is expected/desired.

Possible states:
Batch/Sysbatch jobs:
- Complete if all are complete
- Running if any are running
- Degraded if any are failed/lost
- Failed if all are failed/lost

Service and System jobs:
- Healthy if all are running
- Recovering if any are "pending" (Note: pending in a non-deploying state means a reschedule has taken place, etc.)
- Degraded if some are failed/lost/unplaced
- Failed if all are failed/lost/unplaced

image
image
image
image
image

### DIFF
--- a/ui/app/components/job-status/panel/deploying.hbs
+++ b/ui/app/components/job-status/panel/deploying.hbs
@@ -9,7 +9,7 @@
       {{did-insert (action this.establishOldAllocBlockIDs)}}
       >
         <h2>Status:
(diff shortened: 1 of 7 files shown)


mikenomitch (approved):
played w vercel and lgtm!

PR #17600, opened 2023-06-19
ci: hclfmt files

lgfa29:
Apply `hclfmt` on files in the `terraform` directory and also add a new workflow to run `make check` on paths that are ignored by [`test-core.yaml`](https://github.com/hashicorp/nomad/blob/main/.github/workflows/test-core.yaml) but have content verified by `make check`.

https://github.com/hashicorp/nomad/pull/17601 shows an example where code is changed. Only one instance of `checks.yaml` runs.

https://github.com/hashicorp/nomad/pull/17604 shows an example where only HCL files are changed. Again, only one isntance of `checks.yaml` runs.

I think we may ended up with duplicate runs if a PR changes files in and out of the ignore list, but I think that's relative rare? Except maybe for `website/**` so we could remove it from `checks.yaml`.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -1,0 +1,36 @@
+# Run 'make check' on paths ignored by test-core.yaml.
+name: Run checks
+on:
+  pull_request:
+    paths:
+      - 'demo/**'
+      - 'e2e/terraform/**'
+      - 'terraform/**'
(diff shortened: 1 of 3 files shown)


jrasell (approved):
LGTM! I am going to get this merged, so I can rerun and fix the nightly processes.

PR #17601, opened 2023-06-19
tmp: ci test for test failures

lgfa29:


main_test.go
```diff
@@ -3,5 +3,11 @@
...
 package main
 
+import "testing"
+
 // This file is intentionally empty to force early versions of Go
 // to test compilation for tests.
+
+func TestFail(t *testing.T) {
+	t.Fatal("Boom")
+}
```
lgfa29, line 13: I think I managed to change a file not being tests 😅

PR #17604, opened 2023-06-19
tmp: test hclfmt failures

lgfa29:


.github/workflows/test-core.yaml
```diff
@@ -2,6 +2,7 @@ name: Core CI Tests
 on:
   pull_request:
     paths-ignore:
+      - '.github/**'
```
lgfa29, line 5: I don't know why this workflow was not being ignored with just the `*.tf` file changes. I think GH may have been comparing the diff all the way to `main` which included the changes in `fix-tf-hclfmt`.

PR #17614, opened 2023-06-20
test: add MultiregionMinJob mock

lgfa29:
Add `MultiregionMinJob()` mock to be used in Enterprise tests. Backporting in case it becomes useful to test bug fixes.


tgross (approved):
LGTM!

Issue #17585, opened 2023-06-18
Scheduler panic at evaluation time if there is a client with `max_dynamic_port` == `min_dynamic_port`


stswidwinski:
### Nomad version
Tip

### Operating system and Environment details
Unix. Doesn't matter.

### Issue
The scheduler panics during evaluation if there exists a client whose `max_dynamic_port` is the same as `min_dynamic_port`. The documentation does not specify whether these bounds are exclusive or inclusive, but the implementation of the `rand` function in `go` does.

The answer is: "don't misconfigure your clients" (and I agree), but it would be nice if the client itself validated the configuration and crashes if the configuration were to produce unexpected result. A corollary of the above is that it seems hard to disable dynamic port allocation.

### Reproduction steps
```
# Start a nomad server
$ cat server_config.hcl
data_dir = "/tmp/nomad/server"
log_level = "TRACE"
  
advertise {
  http = "127.0.0.1"
  rpc = "127.0.0.1"
  serf = "127.0.0.1"
}
  
server {
  enabled = true
  bootstrap_expect = 1
}
$ ./nomad agent -config server_config.hcl
 
# Start a nomad client
#
## It is important that max_dynamic_port == min_dynamic_port
$ cat client_config.hcl
data_dir = "/tmp/nomad/client-1"
log_level = "debug"
  
advertise {
  http = "127.0.0.1"
  rpc = "127.0.0.1"
  serf = "127.0.0.1"
}
  
ports {
  http = "9876"
  rpc = "9875"
  serf = "9874"
}
  
client {
  enabled = true
  servers = ["127.0.0.1"]
  max_dynamic_port = 20000
  min_dynamic_port = 20000
}
  
plugin "raw_exec" {
  config {
    enabled = true
  }
}
$ ./nomad agent -config client_config.hcl
```

The cluster will start without any issues:

```
$ ./nomad node status
ID        DC   Name             Class   Drain  Eligibility  Status
003e7e01  dc1  <name>           <none>  false  eligible     ready
--
```

Now, start a job with dynamically allocates a port:

```
$ cat one_time_job.hcl
job "example" {
  datacenters = ["dc1"]
  type = "batch"
 
  group "test-group" {
    network {
      port "http" {}
    }
 
    task "test-task" {
      driver = "raw_exec"
 
  
      config {
        command = "bash"
        args = [ "-c", "env; sleep 1000" ]
      }
    }
  }
}
$ ./nomad run one_time_job.hcl
```

[...]

PR #17619, opened 2023-06-20
scheduler: tolerate having only one dynamic port available

tgross:
If the dynamic port range for a node is set so that the min is equal to the max, there's only one port available and this passes config validation. But the scheduler panics when it tries to pick a random port. Only add the randomness when there's more than one to pick from.

Adds a test for the behavior but also adjusts the commentary on a couple of the existing tests that made it seem like this case was already covered if you didn't look too closely.

Fixes: https://github.com/hashicorp/nomad/issues/17585


shoenig (approved):
LGTM!

PR #17626, opened 2023-06-20
[UI, CI] Test splitting

DingoEatingFuzz:
UI tests have already been running parallel in CI, this changes it so they aren't just parallelized on one machine (which could be leading to local state related flakiness), instead they are run across a set of GHA jobs.

Additionally, this removes the artifact uploading step, which was dutifully migrated from the previous Circle config. Mechanically the `upload-artifact` action does indeed upload the XML file just like how the Circle job worked, however, the XML file was only being generated and uploaded in service of [test insights features exclusive to Circle](https://circleci.com/docs/collect-test-data/). So this process has become strictly vestigial.


.github/workflows/_ember-exam.yml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+      - name: yarn install
+        run: cd ui && yarn install --frozen-lockfile
+      - name: Ember tests
+        env:
+          PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
+        run: mkdir -p /tmp/test-reports && cd ui && yarn exam:parallel -- --split=${{ inputs.split }} --partition=${{ inputs.partition }}
```
gulducat, line 27: I get a little lost in the --sauce-- here...

`yarn exam:parallel` kinda expands to `percy exec --parallel -- ember exam` then these extra args get tacked on to that, right? so we end up with `percy exec --parallel -- ember exam -- --split=1 --partition=4`, right?  the double-double-dash has me wanting to make fewer leaps to know which program ends up with which arguments.

maybe that's okay, but I wanted to mention it in case something nice comes easily to mind for ya.

but aside from that, while we're renovating, maybe this is a little easier to read?

```suggestion
        working-directory: ui
        run: |-
          mkdir -p /tmp/test-reports
          yarn exam:parallel -- --split=${{ inputs.split }} --partition=${{ inputs.partition }}
```

.github/workflows/test-ui.yml
```diff
@@ -47,12 +47,29 @@ jobs:
...
       - name: lint:hbs
         run: cd ui && yarn run lint:hbs
-      - name: Ember tests
+  tests:
+    needs:
+      - lint
+    strategy:
+      matrix:
+        partition: [1, 2, 3, 4]
+        split: [4]
```
gulducat, line 56: I want there to be a `len` function of some kind available, instead of this `[4]`, but I'm not finding one, so fine. :yum:

.github/workflows/test-ui.yml
```diff
@@ -47,12 +47,29 @@ jobs:
...
+      - name: yarn install
+        run: cd ui && yarn install --frozen-lockfile
+      - name: finalize
         env:
           PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
-        run: mkdir -p /tmp/test-reports && cd ui && yarn exam
-      - uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
-        with:
-          path: "/tmp/test-reports"
+        run: cd ui && yarn percy -- build:finalize
```
gulducat, line 74: glancing at [the percy docs](https://docs.percy.io/docs/cli-build#percy-buildfinalize), I'm not sure I understand what `finalize` does exactly, and how is it different from `wait`?  I assume it brings the parallel bits back together somehow.

gulducat (approved):
who doesn't love faster tests?  :tada:

.github/workflows/_ember-exam.yml
DingoEatingFuzz: It used to be that the intermediate `--` separated flags for the `yarn` command and the flags for the command the `yarn` command ran.

Turns out it's unnecessary now, so I'm updating this to just `yarn exam:parallel --split=${{ .. }} --partition=${{ .. }}`.

And since we don't need that test-reports dir anymore, I think this will look tidy enough on a single line.

.github/workflows/test-ui.yml
DingoEatingFuzz: It works like `wait` except it coordinates across machines/containers. It also integrates natively with GHA so we don't see any of the IDing and synchronizing of partitions that it's doing behind the scenes.

Relevant docs: https://docs.percy.io/docs/parallel-test-suites#parallel-suites-split-across-different-machines

gulducat: oh neat.  a bit lower down the page there though, it indicates that GHA needs some extra hand-holding, because `GITHUB_RUN_ID` is not unique across reruns?

https://docs.percy.io/docs/parallel-test-suites#advanced-nonce-setting

the example they give though is a ~deprecated way to set GHA step outputs, so you'd want something like this instead:

```
run: echo "result=${{ github.run_id }}-$(date +%s)" >> "$GITHUB_OUTPUT"
```

ref: https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs

DingoEatingFuzz:
Hey @philrenaud, this is looking good, but can you verify things look right in Percy before I merge?

philrenaud (approved):

PR #17628, opened 2023-06-20
Use cpu_total_compute setting also for CPU usage information

### LINKED ISSUE #17577: Fix ARM64 CPU monitoring of jobs
### Proposal

There are many tickets on this issue-tracker that explains that the `cpu_total_compute` doesn't work on ARM64. This ticket is not about that. It is however about the fact that, even if you fill in that value manually, or run on AWS where it is done for you, jobs and clients still show 0% CPU usage. And I think this is just an oversight, and can be fixed.

### Use-cases

Not having CPU overview in the UI, and not receiving `ticks` via Prometheus, is just slightly annoying. Although the `percentage` works fine on Prometheus, and they are basically the same value, it is just nice to have ARM64 on-par with other platforms.

### Attempted Solutions

In short, the problem seems to be in at what place `cpu_total_compute` can be overwritten by configuration.

In `helper/stat/cpu.go` the `Init` uses `github.com/shirou/gopsutil/v3/cpu` to detect CPU MHz. This fails on ARM64, and as such, `cpuTotalTicks` is set to zero.

In `client/fingerprint/cpu.go` the config `request.Config.CpuCompute` is checked, and if set, `cpu_total_computer` is set with this value.

Now everywhere else in the code to calculate CPU ticks, `TotalTicksAvailable` is used. But this cannot be changed via the above mentioned configuration. So although capacity etc is all calculated with `cpu_total_compute`, CPU usage is not. And this feels like an oversight, as those two are based on the same value: the total available MHz.

So, as an experiment I made a very ugly patch, that sets `TotalTicksAvailable` to return `cpu_total_compute`. And that actually just works fine. I can see CPU usage in the UI, and Prometheus starts to report it too.

I wanted to write a proper Pull Request to address this issue, but ... I just don't know how. I think the proper way would be to give `helper/stat/cpu.go` access to the configuration, so when `cpu_total_compute` is set, it uses that, instead of auto-detection. This would also make `client/fingerprint/cpu.go` a bit simpler .. basically, it would move the config setting upstream. But I just didn't know how to do that :)

Anyway, I hope this makes sense, and that you would be willing to help me out getting this fixed. Either by helping me point out how to actually do this, or by possibly leaning your head on the keyboard and doing this for me :D ;) Tnx either way!

PS: another odd thing about all this, is that CPU ticks is a gauge .. I really expected that to be a counter, so you don't have to constantly monitor the value .. if it would be a counter, I could just request the value once every 5 minutes, and still calculate the average CPU usage. As it is now a gauge, it is just another way to represent the percentage .. I find this a bit odd. But not relevant to the problem at hand :)

PPS: how I solved it is like this. But it is very dirty, and just meant to proof it actually addresses the issue.

```diff
diff --git a/client/fingerprint/cpu.go b/client/fingerprint/cpu.go
index c6c69e3b1f..64394f8f59 100644
--- a/client/fingerprint/cpu.go
+++ b/client/fingerprint/cpu.go
@@ -141,6 +141,7 @@ func (f *CPUFingerprint) setTotalCompute(request *FingerprintRequest, response *
        default:
                ticks = defaultCPUTicks
        }
+       stats.SetTotalTicksAvailable(ticks)
        response.AddAttribute("cpu.totalcompute", fmt.Sprintf("%d", ticks))
        f.resources.CPU = int(ticks)
        f.nodeResources.Cpu.CpuShares = int64(ticks)
diff --git a/helper/stats/cpu.go b/helper/stats/cpu.go
index 4ddb6f517a..b1755d1f8d 100644
--- a/helper/stats/cpu.go
+++ b/helper/stats/cpu.go
@@ -106,6 +106,10 @@ func CPUModelName() string {
        return cpuModelName
 }
 
+func SetTotalTicksAvailable(new uint64) {
+       cpuTotalTicks = new
+}
+
 // TotalTicksAvailable calculates the total MHz available across all cores.
 //
 // Where asymetric cores are correctly detected, the total ticks is the sum of
```

TrueBrain:
Fixes #17577.

Currently, `cpu_total_compute` is only used for fingerprinting, and doesn't influence `TotalTicksAvailable`. But `TotalTicksAvailable` is used to calculate CPU usage ticks. If for example on ARM64 no CPU MHz can be detected, `TotalTicksAvailable` remains zero. And in result all CPU usage remains zero.

This PR sets out to address that problem by having `cpu_total_compute` override `TotalTicksAvailable` too.


drivers/shared/executor/executor.go
```diff
@@ -261,9 +261,9 @@ type UniversalExecutor struct {
 // NewExecutor returns an Executor
 func NewExecutor(logger hclog.Logger) Executor {
 	logger = logger.Named("executor")
-	if err := shelpers.Init(); err != nil {
-		logger.Error("unable to initialize stats", "error", err)
-	}
+	// if err := shelpers.Init(); err != nil {
+	// 	logger.Error("unable to initialize stats", "error", err)
+	// }
```
TrueBrain, line 266: Sadly, I am a bit out of my depth here .. there are two calls to `Init()`, but here it seems I cannot access the `cpu_total_compute`. I am not sure how to handle this situation.

I am also not really sure why the `Init` is here; I guess in case it is called before fingerprinting is done; or in cases when no fingerprinting is done at all. But this requires more domain knowledge of Nomad than I have.

TrueBrain: One thing is sure: without these statements, `raw_exec` job's CPU is not measured :)

TrueBrain: As I wanted to learn Go a bit more, I delved deeper into this rabbithole. I made a commit that addresses the problem and makes everything work .. but I am not sure it was the right way to go :)

Basically, to get the CpuCompute down here without having access to the configuration, it needs to be send through a complex path. I was hoping for a more elegant solution, but .. here we are :(

client/fingerprint/cpu.go
```diff
@@ -42,7 +42,7 @@ func NewCPUFingerprint(logger hclog.Logger) Fingerprint {
 }
 
 func (f *CPUFingerprint) Fingerprint(request *FingerprintRequest, response *FingerprintResponse) error {
```
tgross, line 44: There's a test failure covering this method:

```
=== FAIL: client/fingerprint TestCPUFingerprint_OverrideCompute (0.00s)
    cpu_default_test.go:108: expected override cpu of 4711 but found 4588
```

drivers/shared/executor/utils.go
```diff
@@ -31,6 +31,7 @@ const (
 func CreateExecutor(logger hclog.Logger, driverConfig *base.ClientDriverConfig,
 	executorConfig *ExecutorConfig) (Executor, *plugin.Client, error) {
 
+	executorConfig.CpuCompute = driverConfig.CpuCompute
```
tgross, line 34: Several tests are currently panicking at this line.

client/fingerprint/cpu.go
tgross, line 44: There's a test failure covering this method:

```
=== FAIL: client/fingerprint TestCPUFingerprint_OverrideCompute (0.00s)
    cpu_default_test.go:108: expected override cpu of 4711 but found 4588
```

And a test compilation error:

```
Error: client/stats/cpu_test.go:35:13: not enough arguments in call to shelpers.Init
	have ()
	want (uint64)
```

tgross (changes requested):
Hi @TrueBrain I see where you're going with this and how you've threaded it down into the executor. Architecturally this seems like we're having to expose fingerprinted data to the plugin, when normally I'd expect that data to flow the other direction. 

I'm going to tag-in @shoenig as I know he's looked at this area of the code a bunch recently to do https://github.com/hashicorp/nomad/pull/16672 and I want to get his thoughts here.

tgross:
Thanks @TrueBrain. We're in the middle of shipping the release candidate for 1.6.0 but will circle back to re-review once that work is done.

shoenig:
This looks good @TrueBrain! The refactoring definitely makes sense. I think we can focus on this as-is and backport it (going into 1.6.1, 1.5.x, 1.4.x). Just FYI in Nomad 1.7 we're planning some significant work around NUMA node detection, where we might go back to plumbing the cpu_total_compute through the RPC layer (along with the rest of the cpu topology). The trick here of passing it through directly to the `NewExecutor` constructor doesn't help external plugins. 

I tested these changes on ec2 and it seems *Task* utilization is still broken,

```
ubuntu@ip-172-31-80-227:~$ ./nomad.tb node status -self -verbose | grep -E '(cpu\.)|(instance-type)'
cpu.arch                                 = arm64
cpu.frequency                            = 2500
cpu.modelname                            = Neoverse-N1
cpu.numcores                             = 8
cpu.reservablecores                      = 8
cpu.totalcompute                         = 20000
platform.aws.instance-type               = t4g.2xlarge
```

```
job "example" {
  group "group" {
    task "task" {
      driver = "exec"
      config {
        command = "stress"
        args    = ["-c", "2"] # burn 2 cores
      }
      resources {
        cpu    = 500
        memory = 256
      }
    }
  }
}
```

and cpu utilization for the task is still 0, though the system CPU stats are calculated correctly.


```
ubuntu@ip-172-31-80-227:~$ ./nomad.tb alloc status -stats a8
```

```
Task "task" is "running"
Task Resources:
CPU        Memory           Disk     Addresses
0/500 MHz  480 KiB/256 MiB  300 MiB

Memory Stats
Cache  Swap     Usage
0 B    480 KiB  480 KiB

CPU Stats
Percent  System Mode  Throttled Periods  Throttled Time  User Mode
199.91%  0.00%        0                  0               199.91%
```

Where the task indicates `0/500 MHz` I'd expect to see `5000/500 MHz` (2 cores * 2500 MHz each).

Likewise the node status utilization metric is zero

```
ubuntu@ip-172-31-80-227:~$ ./nomad.tb node status -self -stats
```

```
Allocated Resources
CPU            Memory          Disk
500/20000 MHz  256 MiB/31 GiB  300 MiB/29 GiB

Allocation Resource Utilization
CPU          Memory
0/20000 MHz  480 KiB/31 GiB

Host Resource Utilization
CPU          Memory          Disk
0/20000 MHz  353 MiB/31 GiB  (/dev/root)
```


_(for comparison, on my amd64 machine)_

```
➜ nomad alloc status -stats 39
```

```
Task "task" is "running"
Task Resources:
CPU           Memory           Disk     Addresses
6000/500 MHz  592 KiB/256 MiB  300 MiB

Memory Stats
Cache  Swap     Usage
0 B    592 KiB  592 KiB

CPU Stats
Percent  System Mode  Throttled Periods  Throttled Time  User Mode
200.01%  0.00%        0                  0               200.01%
```

```
➜ nomad node status -self -stats
```

```
Node Events
Time                  Subsystem  Message
2023-07-17T14:31:16Z  Cluster    Node registered

Allocated Resources
CPU            Memory           Disk
500/96000 MHz  256 MiB/126 GiB  300 MiB/202 GiB

Allocation Resource Utilization
CPU             Memory
5998/96000 MHz  592 KiB/126 GiB

Host Resource Utilization
CPU             Memory           Disk
6119/96000 MHz  685 MiB/126 GiB  19 GiB/233 GiB
```

shoenig:
I think this PR is looking good now, thanks @TrueBrain!

> The more I work on this, the more it feels it needs a bit bigger refactor

Indeed! That refactor (in combination with refactoring cgroups) is a prerequisite for what I'm working on for 1.7

shoenig (approved):

PR #17629, opened 2023-06-20
[ui] Versions added to deploying status panel

philrenaud:
In two places:
- next to the "New allocations: x/y running and healthy" shows the latest version number
- in a "Versions" cell that matches the steady state format, which'll show counts for all allocations running between prev and new allocs of non-terminal status
image
image


mikenomitch (approved):

juliezzhou (approved):
looks great!

Juanadelacuesta (approved):

PR #17630, opened 2023-06-20
Update node version to 16 in release workflow

thefallentree:
node 16 is required to run "generating static asset" part now


tgross:
Hi @thefallentree this PR doesn't say why the upgrade is needed or include any of the changes we'd need for any of the build or test steps or changes to the contributor's tooling. I'm sure at some point we'll need to make this upgrade (cc'ing @philrenaud for that) but it's not going to be right now.

PR #17632, opened 2023-06-21
ui: setup mirage and tests for node pools

lgfa29:
Require node pools to be created before nodes and jobs to force correct environment setup and improve test repeatability.

This approach touches several files, but it makes tests more predictable than creating node pools with jobs and nodes. It also follows the approach taken for namespaces.

### DIFF
--- a/ui/mirage/factories/job.js
+++ b/ui/mirage/factories/job.js
@@ -200,6 +200,11 @@ export default Factory.extend({
   shallow: false,
 
   afterCreate(job, server) {
+    Ember.assert(
+      '[Mirage] No node pools! make sure node pools are created before jobs',
+      server.db.nodePools.length
+    );
+
     if (!job.namespaceId) {
       const namespace = server.db.namespaces.length
         ? pickOne(server.db.namespaces).id
@@ -215,11 +220,8 @@ export default Factory.extend({
     }
 
     if (!job.nodePool) {
-      const nodePool = server.db.nodePools.length
-        ? pickOne(server.db.nodePools).name
-        : server.create('node-pool').name;
       job.update({
-        nodePool,
+        nodePool: pickOne(server.db.nodePools).name,
       });
     }
 
--- a/ui/mirage/scenarios/default.js
+++ b/ui/mirage/scenarios/default.js
@@ -609,7 +609,7 @@ function allNodeTypes(server) {
 
 function everyFeature(server) {
   server.createList('agent', 3, 'withConsulLink', 'withVaultLink');
-  server.createList('node-pool', count);
+  server.createList('node-pool', 3);
 
   server.create('node', 'forceIPv4');
   server.create('node', 'draining');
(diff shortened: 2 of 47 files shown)


ui/mirage/factories/job.js
```diff
@@ -200,6 +200,11 @@ export default Factory.extend({
   shallow: false,
 
   afterCreate(job, server) {
+    Ember.assert(
+      '[Mirage] No node pools! make sure node pools are created before jobs',
+      server.db.nodePools.length
+    );
```
philrenaud, line 206: Is there always a `default` node pool, in practice, to keep this from being length=0 in real life?

ui/mirage/scenarios/default.js
```diff
@@ -602,6 +609,7 @@ function allNodeTypes(server) {
 
 function everyFeature(server) {
   server.createList('agent', 3, 'withConsulLink', 'withVaultLink');
+  server.createList('node-pool', count);
```
philrenaud, line 612: what's `count` in this context?

ui/mirage/utils.js
```diff
@@ -16,8 +16,13 @@ export function provider() {
...
 
-export function pickOne(list) {
-  return list[faker.random.number(list.length - 1)];
+export function pickOne(list, filterFn) {
+  let candidates = list;
+  if (filterFn) {
+    candidates = list.filter(filterFn);
+  }
+
+  return candidates[faker.random.number(candidates.length - 1)];
```
philrenaud, line 25: nice!

ui/tests/acceptance/allocation-detail-test.js
```diff
@@ -34,6 +34,7 @@ module('Acceptance | allocation detail', function (hooks) {
   hooks.beforeEach(async function () {
     server.create('agent');
 
+    server.create('node-pool');
```
philrenaud, line 37: This has made me desperately want a `factory.beforeCreate()` in mirage, but it looks like this isn't a thing. Thanks for putting all the work in for this!

philrenaud (approved):
Huge undertaking, thank you Luiz!

ui/mirage/factories/job.js
lgfa29: Yup! There are even 2 node pools created automatically:
https://github.com/hashicorp/nomad/blob/04042c619d68ad1f67da87c3c0c449de0f9d0637/nomad/state/state_store_node_pools.go#L13-L31

And just to clarify, this assertion is only applied on dev environments using Mirage, so there's no end user impact.

ui/mirage/scenarios/default.js
lgfa29: Ops, nice catch. It's a leftover from a previous implementation approach 😅

ui/tests/acceptance/allocation-detail-test.js
lgfa29: It would make things so much easier 😿

PR #17633, opened 2023-06-21
ui: remove node pool `all` from client list filter

lgfa29:
The built-in node pool `all` should not be a client filter because nodes are not allowed to register there.

image


philrenaud (approved):
Code and test looks good, but can you give me some context about the `all` node pool? Is that it's own entry at /nodepools?

Issue #17449, opened 2023-06-07
variables: add locking construct and functionality to Nomad Variables


jrasell:
Nomad Variables will be enhanced to support basic leadership locking. This would allow applications such as the Nomad Autoscaler to implement high-availability functionality without the initial need for additional external dependencies
 
 **Work Items** 
 
- [ ] State store and structs
- [ ] Trackers
- [ ] RPC
- [ ] HTTP API
- [ ] API SDK
- [ ] CLI
- [ ] Docs

PR #17636, opened 2023-06-21
var: add initial locking struct to core package.

jrasell:
Targets feature branch.
Related #17449


nomad/structs/variables_test.go
```diff
@@ -8,9 +8,177 @@ import (
...
+				CreateTime:  1687251815,
+				ModifyIndex: 100,
+				ModifyTime:  1687251876,
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			actualOutput := tc.inputVariableMetadata.Copy()
```
tgross, line 176: We're missing a `Copy` of the `VariableLock`, so while these will pass equality, both `Lock` fields are pointing to the exact same object and so this test doesn't fully exercise the `Equal` check.

tgross (changes requested):

nomad/structs/variables_test.go
jrasell: Oh yes, thanks for catching this. I have forced pushed a change to resolve this and wonder if I should add checks to other copy functions that might be missing this.

tgross (approved):
LGTM!

PR #17639, opened 2023-06-21
Moves to the current LTS release of Node for our build and release workflows

philrenaud:
Much like the ember-test-audit node version bumps from https://github.com/hashicorp/nomad/pull/17624, this upgrades Node to the current LTS version for our build workflows.

Without this, an older version of node would fail upon storybook installation, and possibly elsewhere.


tgross (approved):
LGTM!

We probably should update the ui/README.md as well: https://github.com/hashicorp/nomad/tree/main/ui#prerequisites Maybe just say "the current LTS" so we don't have to keep updating it? Not sure how closely we have to pin versions.

PR #17641, opened 2023-06-21
ui: update Topology page tooltips to Helios

lgfa29:
Use new Helios tooltip component with more descriptive `aria-label` values. Wrap `Hds::TooltipButton` inside the `span` so the `bumber-left` offset is not included in the tooltip position.

Also make the use of the terms "node" and "client" more consistent, switching to "node" since it was used more often.


philrenaud (changes requested):
Hm,. not sure what's going on, but on my screen, when hovering, it looks like something is being shifted and content moves around past the edge of the screen. Causing a screen shift. Maybe something with bumper-left ?
image

image

PR #17644, opened 2023-06-21
[ui] Job Status/Deployment Panel: legend items with a count of 0 no longer links

philrenaud:
Previously, the "0 pending", "0 failed" here would have been links to otherwise empty list pages. Now we check for the existence of allocations in that class before making it a link, otherwise it's just a span.

image

### DIFF
--- a/ui/app/components/job-status/panel/steady.hbs
+++ b/ui/app/components/job-status/panel/steady.hbs
@@ -51,7 +51,7 @@
         <legend>
           {{#each this.allocTypes as |type|}}
             <ConditionalLinkTo
-              @condition={{not (eq type.label "unplaced")}}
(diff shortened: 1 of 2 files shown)


ui/app/components/job-status/panel/steady.hbs
```diff
@@ -51,7 +51,7 @@
         <legend>
           {{#each this.allocTypes as |type|}}
             <ConditionalLinkTo
-              @condition={{not (eq type.label "unplaced")}}
+              @condition={{and (not (eq type.label "unplaced")) (get (get (get (get this.allocBlocks type.label) 'healthy') 'nonCanary') "length")}}
```
gulducat, line 54: good golly gracious, get get get get it good!

gulducat (approved):
have you considered linking to https://devnull-as-a-service.com ?

PR #17646, opened 2023-06-21
ci: set `continue-on-error: true` on `test-ui`

lgfa29:
Since the matrix exercises different test cases, it's better to allow all partitions to completely run, even if one of them fails, so it's easier to catch multiple test failures.


philrenaud (approved):
Nice, I always assumed this was the default. TIL!

Issue #17651, opened 2023-06-21
Replacement allocations during canary deployments are placed in the wrong datacenter


lgfa29:
### Nomad version
```
Nomad v1.5.6
BuildDate 2023-05-19T18:26:13Z
Revision 8af70885c02ab921dedbdf6bc406a1e886866f80
```

But it likely happens in previous versions as well.

### Operating system and Environment details

N/A

### Issue

When a job with canary deployments changes its `datacenters` and, during deployment, an allocation for the previous version fails, the replacement allocations are created in nodes for the new `datacenters` value instead of the original one.

### Reproduction steps

1. Start a Nomad cluster with clients in two datacenters. You start a `nomad agent -dev` and run the job below to create some extra clients.
<details>
<summary>Nomad clients jobfile</summary>

```hcl
locals {
  # Adjust to the appropriate path.
  nomad_path = "/opt/hashicorp/nomad/1.5.6/nomad"

  client_config = <<EOF
data_dir   = "{{env "NOMAD_TASK_DIR"}}/data"
name       = "%s"
datacenter = "%s"

client {
  enabled = true

  server_join {
    retry_join = ["127.0.0.1"]
  }
}

server {
  enabled = false
}

ports {
  http = "46%d6"
  rpc  = "46%[3]d7"
  serf = "46%[3]d8"
}

plugin "raw_exec" {
  config {
    enabled = true
  }
}
EOF
}

job "nomad" {
  group "clients" {
    task "client-dc2" {
      driver = "raw_exec"

      config {
        command = local.nomad_path
        args    = ["agent", "-config", "local/config.hcl"]
      }

      template {
        data        = format(local.client_config, "client-dc2", "dc2", 5)
        destination = "local/config.hcl"
      }
    }

    task "client-dc3" {
      driver = "raw_exec"

      config {
        command = local.nomad_path
        args    = ["agent", "-config", "local/config.hcl"]
      }

      template {
        data        = format(local.client_config, "client-dc3", "dc3", 6)
        destination = "local/config.hcl"
      }
    }
  }
}
```
</details>

2. Run sample job below. One of the allocations will keep failing.
3. Update the job's datacenter.

    ```diff
    job "sleep" {
    - datacenters = ["dc2"]
    + datacenters = ["dc3"]
    ```
4. Run job again and monitor its allocations
    ```shell-session
    $ watch -n1 nomad job allocs sleep
    ID        Node ID   Task Group  Version  Desired  Status   Created    Modified
[...]

PR #17653, opened 2023-06-22
core: set proper nodes on downgrade

lgfa29:
When a non-canary allocation fails during a canary deployment it must be replaced according to the original job specification, which includes resources, constraints, datacenters etc.

Prior to this change, only the job was being reset in the stack, which could lead to the replacement being placed in nodes in the wrong datacenter.

Manual backport to `release/1.4.x` from the solution discussed in https://github.com/hashicorp/nomad/pull/17598#discussion_r1235295426 which will go to `main`.

FIxes https://github.com/hashicorp/nomad/issues/17651

### DIFF
--- a/scheduler/generic_sched_test.go
+++ b/scheduler/generic_sched_test.go
@@ -853,17 +853,12 @@ func TestServiceSched_Spread(t *testing.T) {
 
 // TestServiceSched_JobRegister_Datacenter_Downgrade tests the case where an
 // allocation fails during a deployment with canaries, an the job changes its
-// datacenter. The failed alloc should be placed in the datacenter of the
-// original job.
+// datacenter. The replacement for the failed alloc should be placed in the
+// datacenter of the original job.
 func TestServiceSched_JobRegister_Datacenter_Downgrade(t *testing.T) {
 	ci.Parallel(t)
 
 	h := NewHarness(t)
-
-	// Set global scheduler configuration.
-	//h.State.SchedulerSetConfig(h.NextIndex(), &structs.SchedulerConfiguration{
-	//	SchedulerAlgorithm: structs.SchedulerAlgorithmBinpack,
-	//})
 
 	// Create 5 nodes in each datacenter.
 	// Use two loops so nodes are separated by datacenter.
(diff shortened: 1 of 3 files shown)


scheduler/generic_sched_test.go
```diff
@@ -851,6 +851,121 @@ func TestServiceSched_Spread(t *testing.T) {
 	}
 }
 
+// TestServiceSched_JobRegister_Datacenter_Downgrade tests the case where an
+// allocation fails during a deployment with canaries, an the job changes its
+// datacenter. The failed alloc should be placed in the datacenter of the
+// original job.
```
tgross, line 857:
```suggestion
// TestServiceSched_JobRegister_Datacenter_Downgrade tests the case where an
// allocation fails during a deployment with canaries, an the job changes its
// datacenter. The replacement for the failed alloc should be placed in the 
// datacenter of the original job.
```

scheduler/generic_sched_test.go
```diff
@@ -851,6 +851,121 @@ func TestServiceSched_Spread(t *testing.T) {
...
+func TestServiceSched_JobRegister_Datacenter_Downgrade(t *testing.T) {
+	ci.Parallel(t)
+
+	h := NewHarness(t)
+
+	// Set global scheduler configuration.
+	//h.State.SchedulerSetConfig(h.NextIndex(), &structs.SchedulerConfiguration{
+	//	SchedulerAlgorithm: structs.SchedulerAlgorithmBinpack,
+	//})
+
```
tgross, line 867:
```suggestion
```

scheduler/generic_sched_test.go
```diff
@@ -851,6 +851,121 @@ func TestServiceSched_Spread(t *testing.T) {
...
+		Priority:    job2.Priority,
+		TriggeredBy: structs.EvalTriggerJobRegister,
+		JobID:       job2.ID,
+		Status:      structs.EvalStatusPending,
+	}
+	must.NoError(t, h.State.UpsertEvals(structs.MsgTypeTestSetup, h.NextIndex(), []*structs.Evaluation{eval}))
+
+	processErr := h.Process(NewServiceScheduler, eval)
+	require.NoError(t, processErr, "failed to process eval")
+	require.Len(t, h.Plans, 1)
```
tgross, line 946:
```suggestion
	must.NoError(t, processErr, must.Sprint("failed to process eval"))
	must.Len(t, 1, h.Plans)
```

tgross (approved):
LGTM!

PR #17656, opened 2023-06-22
np: check for license on RPC endpoints

lgfa29:
Actual license verification will happen in the Enterprise repo.

### DIFF
--- a/nomad/node_pool_endpoint.go
+++ b/nomad/node_pool_endpoint.go
@@ -49,6 +49,9 @@ func (n *NodePool) List(args *structs.NodePoolListRequest, reply *structs.NodePo
 		return err
 	}
 
+	// Only warn for expiration of a read request.
+	_ = n.validateLicense(nil)
+
 	// Setup blocking query.
 	sort := state.SortOption(args.Reverse)
 	opts := blockingOptions{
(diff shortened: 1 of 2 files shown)


nomad/node_pool_endpoint.go
```diff
@@ -134,6 +137,9 @@ func (n *NodePool) GetNodePool(args *structs.NodePoolSpecificRequest, reply *str
 		return structs.ErrPermissionDenied
 	}
 
+	// Only warn for expiration of a read request.
+	_ = n.validateLicense(nil)
```
tgross, line 141: This only warns in the logs, right? The user running `nomad node pool list` doesn't see anything? (I guess we have no way of reporting warnings in the typical RPCs though, so :shrug: )

tgross (approved):
LGTM

nomad/node_pool_endpoint.go
lgfa29: Yeah, it's a bit weird in this context, but in Enterprise code there's a flag to emit logs that triggers this block:
https://github.com/hashicorp/nomad-enterprise/blob/main/nomad/license_watcher_ent.go#L173-L183

PR #17660, opened 2023-06-22
variables: remove unused state store functions.

jrasell:


tgross (approved):
LGTM! I don't recall why we had this when the various `GetVariable*` method exist; I think it was effectively a proposal around tombstones but if we're not using it there's no reason to keep it around (git remembers!). Should this get backported to 1.4.x where Variables were introduced as well?

PR #17661, opened 2023-06-22
Add render_template_on_restart param for task block

nvanthao:
This PR creates new parameter `render_template_on_restart` for `task` block. This allow jobspec author to specify that all templates will be explicitly re-rendered when a task is restarted.

This can be useful for re-fetching Vault secrets, even if the lease on the existing secrets has not yet expired. Currently, without blocking query support in Vault, the only way to fetch new secrets is to submit new jobspec.

The default value of this parameter is `false` and should be fully backward compatible.

Other abandoned idea is to have the parameter at `template` level. I don't know a straight way to reload specific template.


jmicceri:
@nvanthao we would really love this feature, any movement on getting it into the next Nomad version? (With a backport to 1.4.x 🙏🏼 )

schmichael (approved):
Looks good to me! Thanks for the contribution and thoroughness!

The name is a bit awkward, but I can't think of a better one.

The one side effect is that it will block disconnected tasks from restart (eg tasks on nodes that are partitioned from Consul). Since the default value is `false` this seems like a fine tradeoff to allow users to opt into.

nvanthao:
Many thanks @schmichael for reviewing! Please let me know if there's any pending action on my end.

schmichael (changes requested):
Sorry for the delay, chatted with the team and someone pointed out the [`restart`](https://developer.hashicorp.com/nomad/docs/job-specification/restart) block might be an ideal place for this to live since it applies to _all_ templates on restart.

That would also provide this parameter with the `restart` block's inheritance: users who want this behavior can opt in at the `group` level and get it for every task.

The new parameter could look like:

```hcl
  restart {
    render_templates = true
  }
```

Sorry for the delay! Let me know if you don't have time to make the change, and I can move this forward!

nvanthao:
I really like the proposal of having `restart` block updated with `render_templates`. I have not thought about this before. Let me work on it and I shall update us soon.

nvanthao:
Hi @schmichael,

Since the proposal is different with current PR, I have raised a new PR with the proposed changes.
https://github.com/hashicorp/nomad/pull/18054

I'll close this and we can continue the discussion on the other PR.

PR #17662, opened 2023-06-22
var: add TTL and delay timer functionality to new pkg

jrasell:
Targets feature branch.
Related https://github.com/hashicorp/nomad/issues/17449

### DIFF
--- a/nomad/lock/delay.go
+++ b/nomad/lock/delay.go
@@ -1,0 +1,83 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package lock
+
+import (
+	"sync"
+	"time"
+
+	"github.com/armon/go-metrics"
+	"github.com/hashicorp/nomad/helper"
+)
+
+// DelayTimer is used to mark certain locks as unacquirable. When a locks TTL
+// expires, it is subject to the LockDelay configured within the variable. This
+// prevents another caller from acquiring the lock for some period of time as a
+// protection against split-brains. This is inspired by the lock-delay in
+// Chubby.
+type DelayTimer struct {
+
+	// delayTimers has the set of active delay expiration times, organized by
+	// ID, which the caller dictates when adding entries. The lock should be
+	// used for all access.
+	delayTimers map[string]time.Time
+	lock        sync.RWMutex
+}
+
+// NewDelayTimer returns a new delay timer manager.
+func NewDelayTimer() *DelayTimer {
+	return &DelayTimer{
+		delayTimers: make(map[string]time.Time),
+	}
+}
+
+// Get returns the expiration time of a key lock delay. This must be checked on
+// the leader server only due to the variability of clocks.
+func (d *DelayTimer) Get(id string) time.Time {
+	d.lock.RLock()
+	expires := d.delayTimers[id]
+	d.lock.RUnlock()
+	return expires
+}
+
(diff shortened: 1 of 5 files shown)


nomad/lock/delay.go
```diff
@@ -0,0 +1,83 @@
...
+	_ = time.AfterFunc(delay, func() {
+		d.lock.Lock()
+		delete(d.delayTimers, id)
+		d.lock.Unlock()
+	})
+}
+
+// EmitMetrics is a long-running routine used to emit periodic metrics about
+// the Delay.
+func (d *DelayTimer) EmitMetrics(period time.Duration, shutdownCh chan struct{}) {
```
Juanadelacuesta, line 63: Why not use a context instead of a shutdown channel?

nomad/lock/delay.go
```diff
@@ -0,0 +1,83 @@
...
+		case <-timer.C:
+			metrics.SetGauge([]string{"variables", "locks", "delay_timer", "num"}, float32(d.len()))
+		case <-shutdownCh:
+			return
+		}
+	}
+}
+
+// len returns the number of registered delay timers.
+func (d *DelayTimer) len() int {
```
Juanadelacuesta, line 79:
```suggestion
func (d *DelayTimer) timersCount() int {
```

nomad/lock/ttl.go
```diff
@@ -0,0 +1,105 @@
...
+	defer t.lock.Unlock()
+
+	if tm := t.ttlTimers[id]; tm != nil {
+		tm.Stop()
+		delete(t.ttlTimers, id)
+	}
+}
+
+// StopAll stops and removes all registered timers.
+func (t *TTLTimer) StopAll() {
```
Juanadelacuesta, line 73:
```suggestion
func (t *TTLTimer) StopAndRemoveAll() {
```

nomad/lock/ttl.go
```diff
@@ -0,0 +1,105 @@
...
+
+	if tm := t.ttlTimers[id]; tm != nil {
+		tm.Reset(ttl)
+		return
+	}
+	t.ttlTimers[id] = time.AfterFunc(ttl, afterFn)
+}
+
+// Stop stops the timer with the given ID and removes it from tracking.
+func (t *TTLTimer) Stop(id string) {
```
Juanadelacuesta, line 62:
```suggestion
func (t *TTLTimer) StopAndRemove(id string) {
```

nomad/lock/delay.go
jrasell: This is to keep consistency with how Nomad does this for other metric emission processes. If we wanted to use a context, we should change all processes that use the shutdownCh to also use this.

nomad/lock/delay.go
```diff
@@ -0,0 +1,83 @@
...
+	// ID, which the caller dictates when adding entries. The lock should be
+	// used for all access.
+	delayTimers map[string]time.Time
+	lock        sync.RWMutex
+}
+
+// NewDelayTimer returns a new delay timer manager.
+func NewDelayTimer() *DelayTimer {
+	return &DelayTimer{
+		delayTimers: make(map[string]time.Time),
```
pkazmierczak, line 31: Wouldn't it be good to initialize the mutex in the constructor, too? I would expect trying to get a lock on an uninitialized mutex to cause a panic.

pkazmierczak (approved):
LGTM!

Juanadelacuesta (approved):

PR #17669, opened 2023-06-22
release pipeline: release workflow needs write permissions

tgross:
In #17103 we set read-only permissions on all the workflows. Unfortunately we missed that the `release` workflow makes git commits and pushes them to the repository, so it needs to have write permissions.


gulducat (approved):
:writing_hand:

shoenig (approved):

PR #17676, opened 2023-06-22
ci: fix flaky UI test

lgfa29:
While backporting https://github.com/hashicorp/nomad/pull/17648 I noticed that this test can still be flaky because the service can be assigned to a non-running allocation.

I fixed this during the backports, so pulling it to `main` as here.


philrenaud (approved):
nice find

PR #17690, opened 2023-06-22
CI: refer to repo secrets in an early secret step

gulducat:
The goal here is to minimize merge conflicts between OSS->ENT due to our using Vault only in private repos.

Let me know if this seems misguided, but my plan is to replace this new "Setup secrets" step in ENT with [hashicorp/vault-action](https://github.com/hashicorp/vault-action) in such a way that the steps that use the secrets can reference them the same way `${{ steps.secrets.outputs.SECRET }}` in both OSS and ENT.

Some changes to secrets themselves will still likely collide, but unrelated changes _around_ lines that happen to contain secrets should not.

I've heard other projects just copy the workflows and maintain a separate ENT copy, but I personally find that to be unappetizing.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -13,10 +13,23 @@ jobs:
     runs-on: ubuntu-latest
     container: hashicorpdev/backport-assistant:0.2.3
     steps:
-      - name: Setup secrets
+      # retrieve secrets from Vault (ENT-only right now)
+      - name: Authenticate to Vault
+        if: endsWith(github.repository, '-enterprise')
+        id: vault-auth
+        run: vault-auth
+      - name: Fetch Secrets
+        if: endsWith(github.repository, '-enterprise')
         id: secrets
-        run: |-
-          echo "GITHUB_TOKEN=${{ secrets.ELEVATED_GITHUB_TOKEN }}" >> "$GITHUB_OUTPUT"
+        uses: hashicorp/vault-action@v2.5.0
+        with:
+          url: ${{ steps.vault-auth.outputs.addr }}
+          caCertificate: ${{ steps.vault-auth.outputs.ca_certificate }}
+          token: ${{ steps.vault-auth.outputs.token }}
+          exportEnv: false
+          secrets: |
+            kv/data/github/hashicorp/${{ github.repository }}/gha ELEVATED_GITHUB_TOKEN ;
+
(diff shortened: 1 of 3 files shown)


gulducat:
> Presumably GHA has some way of doing this even its just a YAML include

That sounds nice, doesn't it?  It's an old sore spot: https://github.com/actions/starter-workflows/issues/245

GHA secrets seem to really resist being abstracted...  Maybe I'm missing something, but all the solutions I've seen are some variety of hacky.

The nearest equivalent of a YAML include for this use case would probably be a "[reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow)" (which can inherit all secrets, unlike a "[composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action)").  But even with that, there's no GHA expression to do something like `lookup(secrets, "MY_SECRET")`.  It's always static `secrets.MY_SECRET`, so an include-like reusable workflow would need to set up all secrets on each call.  Maybe that isn't the worst thing?

Another differently-hacky option beyond what I started in this PR, is to also include the Vault lookup(s) here in OSS repo, but conditionally so it's only effective in ENT.  That would put the Vault paths out in the open on a public repo, which is probably fine, but it would require all secrets for ENT (and many are ENT-_only_) be set up here in OSS.

I'm open to other suggestions, but after a decent bit of flailing around, what I started here still feels to me like a decent middle-ground, since secrets aren't added/removed all that often.

tgross:
> I'm open to other suggestions

We're stuck in the rut of trying to make GHA work the "GHA way". Run a shell script that's different in the two repos that sources the environment variables/secrets we need? We should have as much as possible in sensible shell scripts rather than magic yaml.

> but after a decent bit of flailing around, what I started here still feels to me like a decent middle-ground, since secrets aren't added/removed all that often.

I'm worried we're going to get frequent merge conflicts despite your best efforts here. Merge conflicts and backport assistance failures are _incredibly_ painful for us right now to the point where I've been seriously spending a lot of time advocating with leadership for shipping only 3 versions of Nomad instead of 6 just so to reduce the burden of dealing with the poor tooling we have. 

Adding any _new_ ways to have merge conflicts is a regression and we should not ship like that.

gulducat:
closing in favor of the strategy set up in #17775

PR #17695, opened 2023-06-23
client: remove unused nsd check allocation result diff func

jrasell:


Juanadelacuesta (approved):

shoenig:
I don't suppose anyone remembers why I added that in the first place? :sweat_smile:

shoenig (approved):

PR #17697, opened 2023-06-23
Add RPC modifications to include the lock work

Juanadelacuesta:
There are no tests in this PR because some parts of the core implementation are missing, they will be added further on.

Related: https://github.com/hashicorp/nomad/issues/17449
Targets: feature branch


helper/raftutil/msgtypes.go
```diff
@@ -65,6 +65,7 @@ var msgTypeNames = map[structs.MessageType]string{
 	structs.ACLBindingRulesDeleteRequestType:             "ACLBindingRulesDeleteRequestType",
 	structs.NodePoolUpsertRequestType:                    "NodePoolUpsertRequestType",
 	structs.NodePoolDeleteRequestType:                    "NodePoolDeleteRequestType",
+	structs.VarRenewLockRequestType:                      "VarRenewLockRequestType",
```
Juanadelacuesta, line 68: The check fails if this message type is not added, but this message is not intended to make it to raft.

jrasell (changes requested):
Thanks for taking this work on @Juanadelacuesta, I've given it an initial high level pass with some inline comments as well as those below.

I think we will need different validation logic for the acquire/release lock. When attempting to acquire a lock, we would need to check whether the lock is held or not, before continuing with any potential state change. Releasing, would need to at least ensure the caller is the holder of the lock, before allowing it to continue. There are probably other caveats in this that I am missing.

There would also need to validation on the renew, to ensure the caller is the holder of the lock. I have made some additional notes about the renewal which might help decide how this implementation would look.

Give me a shout if you want to go over any of this in more detail.

nomad/structs/structs.go
```diff
@@ -127,6 +127,7 @@ const (
 	ACLBindingRulesDeleteRequestType             MessageType = 58
 	NodePoolUpsertRequestType                    MessageType = 59
 	NodePoolDeleteRequestType                    MessageType = 60
+	VarRenewLockRequestType                      MessageType = 61
```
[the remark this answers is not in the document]
Juanadelacuesta:
```==> Formatting type map...
raftutil helper message type mapping is out of sync. Run go generate ./... and push.
make: *** [GNUmakefile:181: check] Error 1```

when running `make missing`

File: nomad/structs/variables.go [hunk 3]
```diff
@@ -492,3 +548,19 @@ type VariablesReadResponse struct {
...
+	VarMeta *VariableMetadata
+	QueryOptions
+}
+
+// VariablesRenewLockResponse is sent back to the user to inform them of success or failure
+// of the renewal process.
+type VariablesRenewLockResponse struct {
+	Error   error // Error if any
+	VarMeta *VariableMetadata
+	QueryMeta
```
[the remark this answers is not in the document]
Juanadelacuesta: it will indeed

nomad/variables_endpoint.go
```diff
@@ -113,56 +131,60 @@ func (sv *Variables) Apply(args *structs.VariablesApplyRequest, reply *structs.V
...
+			return canRead, fmt.Errorf("delete requires a Path")
+		}
+
+	case structs.VarOpLockRelease:
+		if args.Var == nil || args.Var.Lock == nil {
+			err = fmt.Errorf("release requires a lock")
+			return canRead, err
 		}
+
+		err = structs.ValidatePath(args.Var.Path)
```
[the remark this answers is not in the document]
Juanadelacuesta: yeah, the linter pointer it out as well, fixed!

command/agent/variable_endpoint.go
```diff
@@ -40,10 +47,23 @@ func (s *HTTPServer) VariableSpecificRequest(resp http.ResponseWriter, req *http
 	if len(path) == 0 {
 		return nil, CodedError(http.StatusBadRequest, "missing variable path")
 	}
+
+	queyParams := req.URL.Query()
```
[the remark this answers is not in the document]
Juanadelacuesta: and I even have a spell checker on my IDE :S

nomad/structs/structs.go
```diff
@@ -203,9 +204,10 @@ const (
 	DefaultBlockingRPCQueryTime = 300 * time.Second
 
 	// RateMetric constants are used as labels in RPC rate metrics
-	RateMetricRead  = "read"
-	RateMetricList  = "list"
-	RateMetricWrite = "write"
+	RateMetricRead      = "read"
+	RateMetricList      = "list"
+	RateMetricWrite     = "write"
+	RateMetricLockRenew = "lock.renew"
```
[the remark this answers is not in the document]
tgross: These metrics are all tied to ACLs, so using the `write` metric is the way to go here. (And that's what you did in the RPC handler anyways, so I think this is just leftover.)

nomad/structs/variables.go
```diff
@@ -505,3 +592,37 @@ type VariablesReadResponse struct {
 	Data *VariableDecrypted
 	QueryMeta
 }
+
+// VariablesRenewLockRequest is used to renew the lease on a lock. This request
+// behaves like a write because the renewal needs to be forwarded to the leader
+// where the timers and lock work is kept.
+type VariablesRenewLockRequest struct {
+	Namespace string
```
tgross, line 600: The embedded `WriteRequest` struct already has the `Namespace` field, so I don't think we need it here.

nomad/structs/variables.go
```diff
@@ -35,10 +38,41 @@ const (
...
+	VariablesRenewLockRPCMethod = "Variables.RenewLock"
+
 	// maxVariableSize is the maximum size of the unencrypted contents of a
 	// variable. This size is deliberately set low and is not configurable, to
 	// discourage DoS'ing the cluster
 	maxVariableSize = 65536
+
+	// minLockTTL and maxLockTTL determine the range of valid durations for the
+	// TTL on a lock.They come from the experience on Consul.
+	minLockTTL = 10 * time.Second
```
tgross, line 55: Nitpick: these new constants and error variables are visible at the package level and aren't going to be "obviously wrong" by name. If they were named things like `minVariableLockTTL` then it'll be obvious if accidentally misused for another purpose in this package (`structs` is a very large package).

nomad/variables_endpoint.go
```diff
@@ -579,3 +623,62 @@ func (sv *Variables) groupForAlloc(claims *structs.IdentityClaims) (string, erro
 	}
 	return alloc.TaskGroup, nil
 }
+
+// RenewLock is used to apply a SV renew lock operation on a variable to maintain the lease.
+func (sv *Variables) RenewLock(args *structs.VariablesRenewLockRequest, reply *structs.VariablesRenewLockResponse) error {
+	authErr := sv.srv.Authenticate(sv.ctx, args)
+	if authErr != nil {
+		return structs.ErrPermissionDenied
+	}
```
tgross, line 632: The auth err should be handled after forwarding and RPC rate metrics (compare to the `Apply` RPC above). I know that seems a little weird, but this is how we ensure we can get rate metrics even for invalid authentication or on stale followers.

nomad/variables_endpoint.go
```diff
@@ -579,3 +623,62 @@ func (sv *Variables) groupForAlloc(claims *structs.IdentityClaims) (string, erro
...
+	if encryptedVar == nil {
+		return errors.New("")
+	}
+	if encryptedVar.Lock == nil {
+		return errors.New("")
+	}
+	if encryptedVar.Lock.ID == args.LockID {
+		return errors.New("")
+	}
```
tgross, line 676: This is all just TODO?

command/agent/variable_endpoint.go
```diff
@@ -52,6 +72,57 @@ func (s *HTTPServer) VariableSpecificRequest(resp http.ResponseWriter, req *http
...
+func (s *HTTPServer) variableLockOperation(resp http.ResponseWriter, req *http.Request,
+	operation url.Values) (interface{}, error) {
+
+	// Parse the Variable
+	var Variable structs.VariableDecrypted
+	if err := decodeBody(req, &Variable); err != nil {
+		return nil, CodedError(http.StatusBadRequest, err.Error())
+	}
+
+	if operation[renewLockQueryParam][0] == renewLockQueryParam {
```
tgross, line 84: Don't we already know `renewLockQueryParam` is set because otherwise we wouldn't have called this method? And if not, I think this could panic on deindex?

nomad/variables_endpoint.go
```diff
@@ -113,56 +144,69 @@ func (sv *Variables) Apply(args *structs.VariablesApplyRequest, reply *structs.V
...
+			return fmt.Errorf("delete requires a Path")
 		}
+
+	case structs.VarOpLockRelease:
+		if args.Var == nil || args.Var.Path == "" ||
+			args.Var.Namespace == "" || args.Var.Lock.ID == "" {
+			return errors.New("release requires all lock information")
+		}
+
+		return structs.ValidatePath(args.Var.Path)
```
tgross, line 206: The variable already must exist in order for the operation to work. Do we need to validate it's path again here?

tgross (changes requested):

nomad/structs/variables.go
Juanadelacuesta: Got it, let's rename them then

nomad/variables_endpoint.go
Juanadelacuesta: If a user calls to acquire a lock on a variable that doesn't exit, it gets created, hence the path validation here.

Juanadelacuesta: It's just a quick check to verify the provided path is not empty and is valid, to avoid making a call to the ss if we already know the path will throw an error.

nomad/variables_endpoint.go
Juanadelacuesta: Let me look at this better, not very familiar :)

Juanadelacuesta: It is a weird flow how it is forwarded before checking the error, do you know why it was built like that?

nomad/structs/variables.go
Juanadelacuesta: Indeed, removed

tgross (approved):
LGTM! I've left a few comments but nothing super-blocking, especially as this is going to a feature branch and we can review the whole when we're done.

One recommendation in terms of reviewability for future PRs: this is a huge PR but a lot of the diffs are refactoring by moving large blocks of code around. That's fine! But it would help reviewing if you separate the refactoring into separate commits in the same PR rather than mingling it all together. That way reviewers can focus on the important changes.

nomad/variables_endpoint_test.go
```diff
@@ -22,12 +22,278 @@ import (
...
+			WriteRequest: structs.WriteRequest{
+				Region:    "global",
+				AuthToken: readToken.SecretID,
+			},
+		}
+
+		applyResp := new(structs.VariablesApplyResponse)
+		err := msgpackrpc.CallWithCodec(codec, structs.VariablesApplyRPCMethod, &applyReq, &applyResp)
+		must.EqError(t, err, structs.ErrPermissionDenied.Error())
+	})
```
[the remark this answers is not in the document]
Juanadelacuesta: If the lock acquire was successful, it will always have the lock ID, and no matter if the release was successful or not, it will never return the lock ID, no matter if the token is management or not. The show or no show is more important for the list and get endpoints and this test is for the apply one.

nomad/structs/variables.go
```diff
@@ -153,6 +154,77 @@ func (vl *VariableLock) Equal(vl2 *VariableLock) bool {
...
+	}{
+		Alias: (*Alias)(vl),
+	}
+
+	if err = json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+
+	if aux.TTL != nil {
+		switch v := aux.TTL.(type) {
```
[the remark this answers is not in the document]
Juanadelacuesta: it does:
 ```~ % curl -X PUT -H 'Content-Type: application/json' -d '{"Lock": {"TTL":[], "LockDelay":"10s"}}' "http://localhost:4646/v1/var/nomad/jobs/blah?lock-acquire" | jq '.'

{
  "CreateIndex": 12,
  "CreateTime": 1690818288003482000,
  "Items": {},
  "Lock": {
    "TTL": "15s",
    "LockDelay": "10s",
    "ID": "0a575be4-abe1-333f-440d-9b267085df1d"
  },
  "ModifyIndex": 16,
  "ModifyTime": 1690818329917095000,
  "Namespace": "default",
  "Path": "nomad/jobs/blah"
}
```

PR #17702, opened 2023-06-23
docker: add cgroupns task config

lgfa29:
Add support for Docker's [`--cgroupns`](https://docs.docker.com/engine/reference/commandline/run/) flag.

Some applications, like those requiring systemd in a cgroupv2 environment`, need to set this flag in order to run properly, but I'm not sure if this inverferes with the cgroup setup Nomad does, so opening up for discussion.

### DIFF
--- a/drivers/docker/config.go
+++ b/drivers/docker/config.go
@@ -342,6 +342,7 @@ var (
 		"auth_soft_fail": hclspec.NewAttr("auth_soft_fail", "bool", false),
 		"cap_add":        hclspec.NewAttr("cap_add", "list(string)", false),
 		"cap_drop":       hclspec.NewAttr("cap_drop", "list(string)", false),
+		"cgroupns":       hclspec.NewAttr("cgroupns", "string", false),
 		"command":        hclspec.NewAttr("command", "string", false),
 		"cpuset_cpus":    hclspec.NewAttr("cpuset_cpus", "string", false),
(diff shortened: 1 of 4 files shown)


shoenig (changes requested):
@lgfa29 where is this coming from?

lgfa29:
> @lgfa29 where is this coming from?

I was just updating the PR body 😄 

I was helping a member of the community to get an app running and they required this flag to get things working. Does it break anything in Nomad?

lgfa29:
> It'd be helpful to get more info about the context there, those words in that order don't quite clarify what's going on

In this particular case it was k0s (https://docs.k0sproject.io/v1.27.2+k0s.0/k0s-in-docker/#1-initiate-k0s), but I ran into this before with something else, just don't remember exactly what 😅

thornycrackers:
One use case I could see being related to this is trying to run the datadog agent in a container on nomad using the docker driver. From [the docs](https://docs.datadoghq.com/containers/docker/?tab=standard):

```
docker run -d --cgroupns host \
    --pid host \
    --name dd-agent \
    -v /var/run/docker.sock:/var/run/docker.sock:ro \
    -v /proc/:/host/proc/:ro \
    -v /sys/fs/cgroup/:/host/sys/fs/cgroup:ro \
    -e DD_SITE=<DATADOG_SITE> \
    -e DD_API_KEY=<DATADOG_API_KEY> \
    gcr.io/datadoghq/agent:7
```

I think the only way to do the currently would be to use raw exec?

Issue #17696, opened 2023-06-23
allocations stuck during node drains from the GUI


matthiasschoger:
### Nomad version
Output from `nomad version`
Nomad v1.5.6

### Operating system and Environment details
Ubuntu Server 22.10
Linux compute2 5.19.0-45-generic #46-Ubuntu SMP PREEMPT_DYNAMIC Wed Jun 7 09:08:58 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

All the latest patches are installed.

### Issue
In #12324, the following was implemented:

> When a node is drained, system jobs are left until last so that
> operators can rely on things like log shippers running even as their
> applications are getting drained off. Include CSI plugins in this set
> so that Controller plugins deployed as services can be handled as
> gracefully as Node plugins that are running as system jobs.

I'm currently running a 3-node cluster, with the tasks using the NFS CSI plugin to mount storage from my local NAS.

When draining a node from the UI with both "Force Drain" and "Drain System Jobs" enabled, some jobs consistently get stuck when doing a node drain from the UI.

Issue #12324 seems to address the issue for the command line, but it seems like the issue still exists for a UI drain.

One more observation: Job migration works flawlessly when I just migrate the service jobs, not system.

### Reproduction steps

1. Set up a 3-node Nomad cluster
2. Deploy the NFS CSI controller and plugin as system jobs.
3. Deploy a job which uses the NFS CSI to mount a storage volume (Gitea and Bookstack consistenty fail for me).
4. Drain the node (both services and system jobs) which contains the service job mounting the NFS share via CSI

#### Expected Result

The service jobs migrate away from the drained node.

#### Actual Result

Service jobs get stuck on the node. Docker fails to stop the containers.

### Job file (if appropriate)

Happy to provide logs and job files, but I'm sure you have the relevant info from issue #12324.

tgross:
Hi @matthiasschoger! The work done in #12324 should be agnostic to the method used to set the drain mode; it doesn't matter at all whether the API request comes from the CLI or web UI as far as I know so long as all the same options are used.

Can you provide more information about what's happening with the service jobs that aren't being drained?

tgross:
[...]

PR #17703, opened 2023-06-23
docs: clarify drain's `-force` flag behavior with system/CSI jobs

tgross:
If you use `nomad node drain -force`, the drain deadline is set to -1ns. If you have not prevented system and CSI node plugin allocations from being drained with `-ignore-system`, they will be immediately drained as well. This is typically not safe for CSI node plugins.

Also fix some broken links.

Preview link: https://nomad-4twwjcfw3-hashicorp.vercel.app/nomad/docs/commands/node/drain
Fixes: #17696


shoenig (approved):
LGTM

tgross:
BPA broke, so I backported 1.5.x by hand.

PR #17704, opened 2023-06-23
docs: split out unsupported versions in changelog

tgross:
Our changelog has become large enough that GitHub's rendering is very slow, resulting in error pages ("angry unicorns"). Split out the older unsupported versions of Nomad into their own file so that we only need to render the most recent versions, while keeping the older versions relatively searchable by having them in a single file.

This will fail automated backports so I'll do those by hand once merged.


lgfa29 (approved):
It actually loads in the first try now 😄 
https://github.com/hashicorp/nomad/blob/docs-changelog-split/CHANGELOG.md

tgross:
Backported to 1.5.x, 1.4.x, and 1.3.x

PR #17705, opened 2023-06-23
np: fix list of jobs for node pool `all`

lgfa29:
Unlike nodes, jobs are allowed to be registered in the node pool `all`, in which case all nodes are used for evaluating placements. When listing jobs for the `all` node pool only those that are explicitly in this node pool should be returned.


tgross (approved):
Good catch. LGTM!

PR #17017, opened 2023-04-27
build: add Docker image

lgfa29:
Update release process to produce a [Docker image](https://hub.docker.com/r/hashicorp/nomad). 

This image is intended for CI usage and running clients as Docker containers is not supported for the reasons described in https://github.com/hashicorp/nomad/issues/6487#issuecomment-542484234 and https://github.com/hashicorp/nomad/issues/6487#issuecomment-1473795286.

Closes https://github.com/hashicorp/nomad/issues/6487

PR #17706, opened 2023-06-23
Smoke test binaries for EL7/glibc-2.17 compatiblity

dekimsey:
## Summary

This adds a quick smoke test of our binaries to verify we haven't exceeded the maximum glibc (2.17) version supported on older OSes (EL7). Which would break our ability to execute on older OSes.

## Why

This is related to our recent efforts to upgrade our GH build environment. When CGO enabled builds are built on Ubuntu-22 they include symbols to glibc versions (2.32+) that exceed our desired minimum OS version (EL7). To mitigate we are instead building on Ubuntu-20. However, Ubuntu-20 itself also includes a newer glibc version. So we need to make sure we check the binaries to prevent us from unintentionally dropping support by adding explicit smoke testing of the binary.

For clarity, here's a breakdown of distribution and glibc versions to provide a bit more context to why we should test on EL7:

| Distro          | glibc version | [EOL](https://endoflife.date)              |
|-----------------|---------------|------------------|
| Debian Jessie   | 2.19          | June 2025        |
| Debian Stretch  | 2.24          | June 2027        |
| CentOS 7        | 2.17          | June 2024        |
| CentOS 8        | 2.28          | December 2021        |
| CentOS 8 Stream       | 2.28          | May 2024        |
| CentOS 9 Stream       | 2.34          | May 2027        |
| RedHat 7        | 2.17          | June 2024        |
| RedHat 8        | 2.28          | May 2031        |
| RedHat 9        | 2.34          | June 2034        |
| Ubuntu 20.04    | 2.31          | May 2030         |
| Ubuntu 22.04    | 2.35          | May 2033         |

## Notes

This only affects CGO_ENABLED=1 binaries (like nomad).

This smoke test is only verifying `amd64` binaries because RH doesn't have an `arm64` container image for EL7 even though it ships the OS (this is fixed as of ubi8). However, no version releases an `arm` image.

## Related

(cross-liking for HC reference)
  - [Findings in hashicorp/nomad-enterprise#951](https://github.com/hashicorp/nomad-enterprise/issues/951#issuecomment-1553338889)
  - [Fixes added to hashicorp/vault-enterprise#4242](https://github.com/hashicorp/vault-enterprise/pull/4242)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -313,5 +313,43 @@ jobs:
             docker.io/hashicorppreview/${{ env.PKG_NAME }}:${{ env.version }}-dev
             docker.io/hashicorppreview/${{ env.PKG_NAME }}:${{ env.version }}-${{ github.sha }}
 
+  minimum-os:
+    name: OS Compatibility
+    # A quick smoke test of our binaries on our minimum target OS (RHEL 7). Why RHEL 7? Because the glibc version is that old (2.17).
+    needs:
+      - get-go-version
+      - get-product-version
+      - build-linux
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      # Note: Ideally we'd test all our target archs, unfortunately availability of containers of these OS's vary.
+      # For instance there is no ubi7 image for arm64 (there is on ubi8), RHBZ#1728771. And none at all for arm.
+      # So we have to settle for only being able to validate where we can, which is just amd64.
+      matrix:
+        goos: [linux]
+        goarch: [amd64]
+    steps:
+      - uses: actions/setup-go@4d34df0c2316fe8122ab82dc22947d607c0c91f9 # v4.0.0
+        with:
+          go-version: ${{needs.get-go-version.outputs.go-version}}
+      - uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
+        with:
+          name: ${{ env.PKG_NAME }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip
+      - name: Test binary
+        env:
+          artifact_name: ${{ env.PKG_NAME }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip
+        run: |
+          echo "::group::Unpack and Prep"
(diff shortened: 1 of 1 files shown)


tgross (approved):
LGTM! Thanks @dekimsey!

Verified in https://github.com/hashicorp/nomad-enterprise/actions/runs/5359393093
and in this PR in https://github.com/hashicorp/nomad/actions/runs/5359921287/jobs/9724253179?pr=17706

I'll merge this in once CI is done.

tgross:
@dekimsey it looks like https://github.com/hashicorp/nomad/pull/17017 got merged before I managed to get this merged and now there's a merge conflict.

dekimsey:
Looks like everything is [clean](https://github.com/hashicorp/nomad/actions/runs/5360236287?pr=17706), so one last push to drop the on.push override used for testing.

image

.github/workflows/build.yml
```diff
@@ -313,5 +313,43 @@ jobs:
...
             docker.io/hashicorppreview/${{ env.PKG_NAME }}:${{ env.version }}-${{ github.sha }}
 
+  minimum-os:
+    name: OS Compatibility
+    # A quick smoke test of our binaries on our minimum target OS (RHEL 7). Why RHEL 7? Because the glibc version is that old (2.17).
+    needs:
+      - get-go-version
+      - get-product-version
+      - build-linux
+    runs-on: ubuntu-latest
```
shoenig, line 323: Generally we prefer specific OS's; we've a let a lot of these creep in so I created https://github.com/hashicorp/nomad/issues/17714 to fixup them all.

shoenig (approved):

.github/workflows/build.yml
dekimsey: Good catch!

tgross:
Looks like this failure https://github.com/hashicorp/nomad/actions/runs/5360327078/jobs/9725065223?pr=17706 was introduced by https://github.com/hashicorp/nomad/pull/17705. And that UI test is a known flake.) Don't worry about that for this PR. Let's get this merged in and I'll fix the other one.

tgross:
Fixed that command test in https://github.com/hashicorp/nomad/pull/17715

tgross:
@gulducat does this PR need to get updated in order to work with the changes we made for permissions/secrets/ENT runners?

gulducat:
@tgross doesn't look to me like it needs anything extra since there are no secrets being used :+1:

Issue #11041, opened 2021-08-11
Add ability to exclude client nodes without explicit constraint


mikenomitch:
### Proposal

Datacenters are currently the only way to have nodes be ineligible to receive a job by default. Other constraints are all opt-in by the jobspec writer.

If a Nomad operator wants other members of their team to not deploy to certain nodes, they have to either use Sentinel (which is difficult and pay gated) or build a layer of validation/templating between their teammates and Nomad.

Ideally, Nomad job submitters could submit jobs and have them deployed to a constrained set of datacenter without knowledge of these constraints. See Proposal Details and Alternatives for details.

### Use-cases

Nomad users who do not have permission to deploy to clients on certain datacenters can deploy a job without specifying this constraint themselves.

Additional benefit/use-case: Nomad operators do not have to add validation step themselves to check for datacenter eligibility.

Additional benefit/use-case: Users who deploy to different data centers can share jobspecs without replacing datacenters value.

### Proposal Details

To accomplish the goals above, I propose extending ACL Policies with two fields in the `namespace` stanza:
* allowed_datacenters: list(string)
* default_datacenters: list(string)

When submitting a job, while validating the jobspec, Nomad should reject any requests that have a datacenter outside the allowed_datacenters list.

If allowed_datacenters is not specified in a policy, it falls back to no validation.

If default_datacenters is specified in a policy, a job submitter can leave out the “datacenters” value in their jobspec and the value will be autopopulated based on their ACL policy.

If default_datacenters is not specified, datacenter must be specified on the jobspec.

Additionally, allowed_datacenters and default_datacenters default values should be configurable in the nomad server config. If these values are given at the server level, the default namespace inherits them, and any policies that do not explicitly override these values do as well.

`allowed_datacenters` would provide a backwards compatible way to make sets of clients ineligible to certain Nomad users

`default_datacenters` would simplify jobspec writing and sharing. Groups within the same organization could share jobspecs without having the overwrite or parameterize the datacenter value. Additionally, the community could more easily share job specs.

### Alternatives

An alternative approach is to keep `datacenters` the same, but add a new attribute on client nodes, like “node_pool” and then add implicit constraints onto that. This may become confusing when added on top of datacenters, but could allow for more complex hierarchies.

pznamensky:
See also: https://github.com/hashicorp/nomad/issues/2299

ygersie:
[...]

PR #17707, opened 2023-06-23
changelog: add entry for node pools

lgfa29:
I don't have a specific PR to point to, so I'm using #11041 since it's more or less where the feature started. I posted an update there describing how node pools work.


tgross (approved):
:tada:

philrenaud (approved):

PR #17708, opened 2023-06-23
[ui] alignment and spacing for job status panel

philrenaud:
Cleans up and standardizes spacing across a handful of elements in the job status panel;
makes the version tabs stand out more;
fades only the text part of faded legend items

image

### DIFF
--- a/.changelog/16134.txt
+++ b/.changelog/16134.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: Added a new Job Status Panel that helps show allocation status throughout a deployment and in steady state
+```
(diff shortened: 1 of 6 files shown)


juliezzhou (approved):
lgtm

tgross (approved):
LGTM

One thing I'm seeing in the preview and just want to make sure isn't related b/c we're discussing text fading here is that the column headers for the Recent Allocations table on this page aren't the same color at the column headers for the Task Groups table above it. ex. https://nomad-storybook-and-qfns3vjoe-hashicorp.vercel.app/ui/jobs/actively-deploying-job@default (`#b5b5b5` vs `#e8e8e8`). Not sure if that's a "this is selectable" indicator above and beyond the hoverstate though. If that's all fine and I'm barking up the wrong tree, ignore me. :grinning:

PR #17717, opened 2023-06-24
e2e: refactor pids isolation tests

shoenig:
This PR refactors some old PID isolation tests to make use of the e2e/v3
packages. Should be quite a bit easier to read. Adds basic `alloc exec` capability
to the jobs3 package.


e2e/v3/jobs3/allocexec3.go
```diff
@@ -0,0 +1,70 @@
...
+			allocID = stub.ID
+			break
+		}
+	}
+	must.NotEq(sub.t, "", allocID, must.Sprintf("no allocation found for %s", group))
+
+	// do stuff
+	allocsAPI := sub.nomadClient.Allocations()
+	alloc, _, err := allocsAPI.Info(allocID, queryOpts)
+	must.NoError(sub.t, err, must.Sprintf("failed to query allocation %s", allocID))
```
tgross, line 35: Is this bit just debugging or is the idea that we're asserting there's a real allocation? Wouldn't that just return an error below if there wasn't?

tgross (approved):
LGTM!

e2e/v3/jobs3/allocexec3.go
shoenig: I dunno, Rob Pike told me never to ignore a returned `err` so there we go

tgross: I meant more why query it at all because I didn't think we used the `alloc`. But I missed that we're using it in the `Exec` call later. :facepalm:

PR #17737, opened 2023-06-26
[ui] links to allocations explicitly go through their route model hook

### LINKED ISSUE #17521: Allocation UI page is missing information when browsed from Job UI page
### Nomad version
v1.3.6

* Chrome 113.0.5672.126 (and earlier)

### Issue

When navigating to an allocation's page in the UI from the job's page via the "Recent Allocations" link, the allocation's page is missing information.

Refreshing the page or opening the allocation page's URL in a new tab fixes the issue.

### Reproduction steps

1. For a job that has at least one running alloc, go to its page (e.g. `https://.../ui/jobs/JobName`)
2. Click on one of the running allocs in the Recent Allocations table

#### Expected Result

* The CPU/Memory graphs are non-empty (i.e. show something other than empty graphs & `0 ... / 0 ... Total`)
* The Task Lifecycle Status table is non-empty
* The Tasks/Ports/Services tables appear

#### Actual Result

See screenshot:

nomad-gh-bugreport-ui

philrenaud:
Resolves #17521 

When loading an allocation from a parent page, like the allocations list of a job, there were circumstances where that allocation would load as `isPartial`, and be missing critical data, notably `taskGroup`. We have components like our `LifecycleChart` that make use of that data and throw a fatal exception if not found.

This ensures that the model, if partial, is fully loaded when /allocations/allocation/$id is hit. This further changes our links to allocation pages to pass in the ID of the allocation rather than a (potentially partial) model itself, in order to cause it to rerun the route model hook.

### DIFF
--- a/ui/app/controllers/clients/client/index.js
+++ b/ui/app/controllers/clients/client/index.js
@@ -205,7 +205,7 @@ export default class ClientController extends Controller.extend(
 
   @action
   gotoAllocation(allocation) {
-    this.transitionToRoute('allocations.allocation', allocation);
+    this.transitionToRoute('allocations.allocation', allocation.id);
   }
 
   @action
--- a/ui/app/controllers/csi/plugins/plugin/allocations.js
+++ b/ui/app/controllers/csi/plugins/plugin/allocations.js
@@ -113,7 +113,7 @@ export default class AllocationsController extends Controller.extend(
   @action
   gotoAllocation(allocation, event) {
     lazyClick([
-      () => this.transitionToRoute('allocations.allocation', allocation),
+      () => this.transitionToRoute('allocations.allocation', allocation.id),
       event,
     ]);
   }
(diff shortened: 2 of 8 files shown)


tgross (approved):
LGTM

What's the scenario where this allocation gets loaded with the partial? I wasn't able to reproduce from the information in #17521

philrenaud:
> What's the scenario where this allocation gets loaded with the partial?

There may be other cases, but the main one I was able to reproduce with was when an allocation had been rescheduled and you could still click into the old one via a allocations index page:

image

(note the icon to the left of the alloc ID, indicating that the alloc had been rescheduled)

DingoEatingFuzz (approved):
It would be nice to have an acceptance test assert that the correct API calls are being made, but I think that's non-blocking if you're looking to get this in the next build.

PR #17738, opened 2023-06-26
[ui] Node Pools moved to after Type in jobs index columns

philrenaud:
- moves the "state" filter box on the clients index page to be first
- moves the "Node Pool" column header to after Type on the jobs index page
image


mikenomitch (approved):
LGTM

PR #17743, opened 2023-06-26
fix changelog entry typo

shoenig:


jrasell (approved):
LGTM! Does the PR need any backport labels, so it gets fixed in the relevant release branches also?

shoenig:
I don't think so, or at least the original PR wasn't backported https://github.com/hashicorp/nomad/pull/16463

@luiz in case we're missing something

lgfa29:
That's right, #16463 was not backported. IIRC the reasoning was that this was more of a semantics issue that we want to correct, but it's not relevant enough to risk breaking existing release lines.

PR #17750, opened 2023-06-27
[ui] Move Placement Failures notification above job status panel

philrenaud:
When present, a placement failure is the most succinct and actionable notification that a user is likely to see during a deployment. As such, this bumps it up to the top of the page.

Additionally, styles the notification with updated HDS/Helios styles.

image


shoenig (approved):
LGTM!

PR #17753, opened 2023-06-27
[ui] Link to allocations.allocation by ID reference, not by model

philrenaud:
Follows up on the work done in #17737 — missed two template-driven (LinkTo) spots that are covered here: the individual allocation blocks (green/red/etc. squares in the status panel), and the left-side links in Deployment History.

### DIFF
--- a/ui/app/components/job-status/deployment-history.hbs
+++ b/ui/app/components/job-status/deployment-history.hbs
@@ -33,7 +33,7 @@
       {{#each this.history as |deployment-log|}}
         <li class="timeline-object {{if (eq deployment-log.exitCode 1) "error"}}">
           <div class="boxed-section-head is-light">
-            <LinkTo @route="allocations.allocation" @model={{deployment-log.state.allocation}} class="allocation-reference">{{deployment-log.state.allocation.shortId}}</LinkTo>
+            <LinkTo @route="allocations.allocation" @model={{deployment-log.state.allocation.id}} class="allocation-reference">{{deployment-log.state.allocation.shortId}}</LinkTo>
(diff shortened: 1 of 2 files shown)


ui/app/components/job-status/deployment-history.hbs
```diff
@@ -33,7 +33,7 @@
       {{#each this.history as |deployment-log|}}
         <li class="timeline-object {{if (eq deployment-log.exitCode 1) "error"}}">
           <div class="boxed-section-head is-light">
-            <LinkTo @route="allocations.allocation" @model={{deployment-log.state.allocation}} class="allocation-reference">{{deployment-log.state.allocation.shortId}}</LinkTo>
+            <LinkTo @route="allocations.allocation" @model={{deployment-log.state.allocation.id}} class="allocation-reference">{{deployment-log.state.allocation.shortId}}</LinkTo>
```
tgross, line 36: If I look at the preview: https://nomad-storybook-and-pufvhadvm-hashicorp.vercel.app/ui/jobs/actively-deploying-job@default there are no deployment history events showing. Is that just a function of not having test data there in the storybook preview?

tgross (approved):
LGTM! (Assuming the history events is a storybook data issue.)

ui/app/components/job-status/deployment-history.hbs
philrenaud: Yeah exactly, it's just that the mocked data there doesn't exist in Mirage (point of clarification: Mirage is the mocked data you see in a deployment, but the dashboard is otherwise the same as you'd see in a real build. Storybook is the components in isolation, like https://nomad-storybook-and-pufvhadvm-hashicorp.vercel.app/storybook/?path=/story/theme-colors--colors)

PR #17755, opened 2023-06-28
Update the revision used by the docker build action.

sarahethompson:
This should always reflect the commit that's being built, as this may differ from the default `github.sha `that the workflow was invoked at.

Goes with https://github.com/hashicorp/actions-docker-build/pull/59 - and should not be merged until the PR is merged and a new version of the action is cut.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -300,7 +300,7 @@ jobs:
         run: |
           echo "revision=${{ github.event.inputs.build-ref }}" >> "$GITHUB_ENV"
       - name: Docker Build (Action)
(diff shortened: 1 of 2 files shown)


sarahethompson:
Tested by running a release build on this branch:
release build: https://github.com/hashicorp/nomad/actions/runs/5402001479 (using the build revision for the docker action)
prepare - https://github.com/hashicorp/crt-workflows-common/actions/runs/5402083028 (successful docker promotion to dev)

tgross (approved):
LGTM! We'll get a chance to try this all out again for 1.6.0-rc.1

PR #17759, opened 2023-06-28
e2e: use DNS instead of HTTP to get my_public_ipv4

gulducat:
`ipify.org` is frequently flakey, causing Terraform apply/destroy to errors in E2E tests, e.g.

```
│ Error: HTTP request error. Response code: 502
│ 
Terraform planned the following actions, but then encountered a problem:
│   with data.http.my_public_ipv4,
│   on network.tf line 23, in data "http" "my_public_ipv4":
│   23: data "http" "my_public_ipv4" {
```

This replaces the HTTP lookup with DNS, which is faster (not that speed matters a whole lot here), but more importantly, OpenDNS should (hopefully) be more reliable.

### DIFF
--- a/e2e/terraform/network.tf
+++ b/e2e/terraform/network.tf
@@ -20,12 +20,18 @@ data "aws_subnet" "secondary" {
   }
 }
 
-data "http" "my_public_ipv4" {
-  url = "https://api.ipify.org"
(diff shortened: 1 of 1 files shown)


shoenig (approved):
LGTM!

schmichael (approved):

PR #17763, opened 2023-06-29
leadership: add lock timer leadership transition functionality.

jrasell:
Lock timers are tracked within memory, the TTLs are monitored from the leader only. This means leadership transition must start/stop certain background processes to track TTL timers, and emit metrics about the background processes.

Related: #17449 
Targets: feature branch

### DIFF
--- a/nomad/leader.go
+++ b/nomad/leader.go
@@ -397,6 +397,17 @@ func (s *Server) establishLeadership(stopCh chan struct{}) error {
 	// Periodically publish job status metrics
 	go s.publishJobStatusMetrics(stopCh)
 
+	// Populate the variable lock TTL timers, so we can start tracking renewals
+	// and expirations.
+	if err := s.restoreLockTTLTimers(); err != nil {
+		return err
+	}
+
+	// Periodically publish metrics for the lock timer trackers which are only
+	// run on the leader.
+	go s.lockTTLTimer.EmitMetrics(1*time.Second, stopCh)
+	go s.lockDelayTimer.EmitMetrics(1*time.Second, stopCh)
+
 	// Setup the heartbeat timers. This is done both when starting up or when
 	// a leader fail over happens. Since the timers are maintained by the leader
 	// node, effectively this means all the timers are renewed at the time of failover.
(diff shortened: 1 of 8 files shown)


nomad/leader.go
```diff
@@ -948,6 +959,88 @@ func (s *Server) restorePeriodicDispatcher() error {
...
+
+	// Iterate the variables, identifying each one that is associated to a lock
+	// and adding a TTL timer for each.
+	for varInterface := varIterator.Next(); varInterface != nil; varInterface = varIterator.Next() {
+		if realVar, ok := varInterface.(*structs.VariableEncrypted); ok && realVar.IsLock() {
+
+			// The variable will be modified in order to show that it no longer
+			// is held. We therefore need to ensure we perform modifications on
+			// a copy.
+			realVarCopy := realVar.Copy()
```
pkazmierczak, line 980: I don't think I get this? `realVar` is itself a reference to `structs.VariableEncrypted`, is it not? By modifying it, won't we be modifying the underlying object?

nomad/leader.go
```diff
@@ -948,6 +959,88 @@ func (s *Server) restorePeriodicDispatcher() error {
...
+// passed ID is expected to be generated via the variable NamespacedID
+// function.
+func (s *Server) createVariableLockTTLTimer(variable *structs.VariableEncrypted) {
+
+	// Adjust the given TTL by multiplier of 2. This is done to give a client a
+	// grace period and to compensate for network and processing delays. The
+	// contract is that a variable lock is not expired before the TTL expires,
+	// but there is no explicit promise about the upper bound so this is
+	// allowable.
+	lockTTL := variable.Lock.TTL * 2
```
pkazmierczak, line 999: Multiplier of 2 is somewhat arbitrary, no? For very small TTLs this won't really give a good grace period.

pkazmierczak (approved):
LGTM, I only have a few minor comments but won't block on that.

nomad/leader.go
Juanadelacuesta: Why not use a factor?

Juanadelacuesta: if the leader transition can take up to 2 ttls, will all the calls to renew will fail in that time?

Juanadelacuesta (approved):

nomad/leader.go
jrasell: TTLs cannot be lower than a 10s time duration, and although arbitrary, I feel it's a good enough number to pick as does Consul. I am open to other suggestions though.

nomad/leader.go
jrasell: Yes, which is why we create a copy, so we are not modifying the underlying object which has been pulled from state and therefore is a pointer reference to the state object.

pkazmierczak: right right 🤦

nomad/leader.go
pkazmierczak: I think if minimum is 10s then you're totally fine.

PR #17765, opened 2023-06-29
job: ensure node pool is canonicalized for state restores.

### LINKED ISSUE #17764: Upgrading to 1.6.0-beta fails restoring Raft snapshots
### Nomad version
Nomad v1.6.0-beta.1+ent
BuildDate 2023-06-27T16:32:15Z
Revision b26558fbf8f4cf89dbaa29068620241746c583bb

(Upgrade from 1.5.6)

Ubuntu 22.04, 3 nodes being server+client

### Issue

Doing an in-place upgrade (run Ansible to replace the binary with the 1.6.0-beta release, then started restarting the Nomad service one by one). The service fails with the following log:

```
Jun 29 10:05:44 nuc-02 systemd[1]: nomad.service: Failed with result 'exit-code'.
Jun 29 10:06:26 nuc-02 systemd[1]: nomad.service: Scheduled restart job, restart counter is at 1.
Jun 29 10:06:26 nuc-02 systemd[1]: Stopped nomad agent.
Jun 29 10:06:26 nuc-02 systemd[1]: Started nomad agent.
Jun 29 10:06:26 nuc-02 nomad[298802]: ==> WARNING: mTLS is not configured - Nomad is not secure without mTLS!
Jun 29 10:06:26 nuc-02 nomad[298802]: ==> Config enable_syslog is `true` with log_level=INFO
Jun 29 10:06:26 nuc-02 nomad[298802]: ==> Loaded configuration from /etc/nomad.d/base.hcl, /etc/nomad.d/client.hcl, /etc/nomad.d/server.hcl
Jun 29 10:06:26 nuc-02 nomad[298802]: ==> Starting Nomad agent...
Jun 29 10:06:26 nuc-02 nomad[298802]:  nomad: setting up raft bolt store: no_freelist_sync=false
Jun 29 10:06:26 nuc-02 nomad[298802]:  nomad.raft: starting restore from snapshot: id=3-8205-1687786395074 last-index=8205 last-term=3 size-in-bytes=83425
Jun 29 10:06:26 nuc-02 nomad[298802]:  nomad.raft: snapshot restore progress: id=3-8205-1687786395074 last-index=8205 last-term=3 size-in-bytes=83425 read-bytes=34130 percent-complete="40.91%"
Jun 29 10:06:26 nuc-02 nomad[298802]: nomad.raft: failed to restore snapshot: id=3-8205-1687786395074 last-index=8205 last-term=3 size-in-bytes=83425 error="job insert failed: missing value for index 'pool'"
Jun 29 10:06:26 nuc-02 nomad[298802]:  nomad: shutting down server
Jun 29 10:06:26 nuc-02 nomad[298802]: ==> Error starting agent: server setup failed: Failed to start Raft: failed to load any existing snapshots
Jun 29 10:06:26 nuc-02 nomad[298802]:     2023-06-29T10:06:26.460Z [INFO]  nomad: setting up raft bolt store: no_freelist_sync=false
Jun 29 10:06:26 nuc-02 nomad[298802]:     2023-06-29T10:06:26.462Z [INFO]  nomad.raft: starting restore from snapshot: id=3-8205-1687786395074 last-index=8205 last-term=3 size-in-bytes=83425
Jun 29 10:06:26 nuc-02 nomad[298802]:     2023-06-29T10:06:26.464Z [INFO]  nomad.raft: snapshot restore progress: id=3-8205-1687786395074 last-index=8205 last-term=3 size-in-bytes=83425 read-bytes=34130 percent-c>
Jun 29 10:06:26 nuc-02 nomad[298802]:     2023-06-29T10:06:26.464Z [ERROR] nomad.raft: failed to restore snapshot: id=3-8205-1687786395074 last-index=8205 last-term=3 size-in-bytes=83425 error="job insert failed:>
Jun 29 10:06:26 nuc-02 nomad[298802]:     2023-06-29T10:06:26.464Z [INFO]  nomad: shutting down server
Jun 29 10:06:26 nuc-02 nomad[298802]:     2023-06-29T10:06:26.464Z [ERROR] nomad: failed to start Raft: error="failed to load any existing snapshots"
Jun 29 10:06:26 nuc-02 nomad[298802]:     2023-06-29T10:06:26.464Z [ERROR] agent: error starting agent: error="server setup failed: Failed to start Raft: failed to load any existing snapshots"
Jun 29 10:06:26 nuc-02 nomad[298802]: nomad: failed to start Raft: error="failed to load any existing snapshots"
Jun 29 10:06:26 nuc-02 nomad[298802]: agent: error starting agent: error="server setup failed: Failed to start Raft: failed to load any existing snapshots"
Jun 29 10:06:26 nuc-02 systemd[1]: nomad.service: Main process exited, code=exited, status=1/FAILURE
Jun 29 10:06:26 nuc-02 systemd[1]: nomad.service: Failed with result 'exit-code'.
```

### Reproduction steps
Upgrade Nomad Ent from 1.5.6 to 1.6.0-beta

#### Expected Result

The upgrade suceeds.

#### Actual Result

Raft snapshots cannot be restored, *maybe* due to Node pools.

jrasell:
I've not had a chance to test this using the linked issue.

closes #17764

### DIFF
--- a/nomad/state/state_store_restore.go
+++ b/nomad/state/state_store_restore.go
@@ -44,6 +44,17 @@ func (r *StateRestore) NodePoolRestore(pool *structs.NodePool) error {
 
 // JobRestore is used to restore a job
 func (r *StateRestore) JobRestore(job *structs.Job) error {
+
+	// When upgrading a cluster pre to post 1.6, the existing jobs will not
+	// have a node pool set. Inserting this into the table will fail, as this
+	// is indexed and cannot be empty.
+	//
+	// This cannot happen within the job canonicalize function, as it would
+	// break the node pools and governance feature.
+	if job.NodePool == "" {
+		job.NodePool = structs.NodePoolDefault
+	}
+
 	if err := r.txn.Insert("jobs", job); err != nil {
 		return fmt.Errorf("job insert failed: %v", err)
 	}
--- a/nomad/state/state_store_restore_test.go
+++ b/nomad/state/state_store_restore_test.go
@@ -50,28 +50,36 @@ func TestStateStore_RestoreJob(t *testing.T) {
 	ci.Parallel(t)
 
 	state := testStateStore(t)
-	job := mock.Job()
+	mockJob1 := mock.Job()
 
 	restore, err := state.Restore()
-	if err != nil {
-		t.Fatalf("err: %v", err)
-	}
+	must.NoError(t, err)
 
-	err = restore.JobRestore(job)
-	if err != nil {
-		t.Fatalf("err: %v", err)
-	}
-	require.NoError(t, restore.Commit())
+	err = restore.JobRestore(mockJob1)
+	must.NoError(t, err)
+	must.NoError(t, restore.Commit())
 
 	ws := memdb.NewWatchSet()
-	out, err := state.JobByID(ws, job.Namespace, job.ID)
-	if err != nil {
-		t.Fatalf("err: %v", err)
-	}
+	out, err := state.JobByID(ws, mockJob1.Namespace, mockJob1.ID)
+	must.NoError(t, err)
+	must.Eq(t, mockJob1, out)
 
-	if !reflect.DeepEqual(out, job) {
-		t.Fatalf("Bad: %#v %#v", out, job)
-	}
+	// Test upgrade to 1.6 or greater to simulate restoring a job which does
+	// not have a node pool set.
+	mockJob2 := mock.Job()
+	mockJob2.NodePool = ""
+
+	restore, err = state.Restore()
+	must.NoError(t, err)
+
+	err = restore.JobRestore(mockJob2)
+	must.NoError(t, err)
+	must.NoError(t, restore.Commit())
+
+	ws = memdb.NewWatchSet()
+	out, err = state.JobByID(ws, mockJob2.Namespace, mockJob2.ID)
+	must.NoError(t, err)
+	must.Eq(t, structs.NodePoolDefault, out.NodePool)
 }
 
 func TestStateStore_RestorePeriodicLaunch(t *testing.T) {


Juanadelacuesta (approved):

tgross (changes requested):
Thanks for jumping on this @jrasell 

Unfortunately this fix breaks the Node Pools Governance feature. If a user submits a job that has no node pool set, we use "default" only for OSS. In Enterprise, we have to look up the default node pool for the job's namespace in the mutator. Because canonicalizing happens first, the node pool would always be set to "default" by the time it hits the mutator and we'll never use the namespace's default pool.

There's an internal [slack discussion](https://hashicorp.slack.com/archives/C050N82220K/p1686859143588389) about this and it end with this important item:

> I may even validate at the state store level to guarantee we always write a node pool value

Which was done correctly in job upsert (ref [`state_store.go#L1701-L1705`](https://github.com/hashicorp/nomad/blob/main/nomad/state/state_store.go#L1701-L1705)) but was _not_ added to [`JobRestore`](https://github.com/hashicorp/nomad/blob/main/nomad/state/state_store_restore.go#L45C1-L51). I think that's where we need to fix it.

tgross (approved):
LGTM! Thanks!

lgfa29:
Thanks for the fix @jrasell and for catching the problem with `Canonicalize` @tgross, this change happened in https://github.com/hashicorp/nomad/pull/17562 which was after our upgrade path testing.

I suspect trying to set this value in `Canonicalize` will be a common mistake so I think I will add a note there.

lgfa29:
Thinking about this a bit more, I find that, strictly speaking, the restore process may also need to take the namespace's default node pool into consideration? 🤔 

I'm still not sure if this is an actual possibility, but if you somehow end up with a namespace that has a default node pool, and a job in that namespace that still doesn't have a node pool value, then on restore we need to use the namespace default.

I find this a very unlikely (impossible?) scenario since that setting the namespace node pool requires a 1.6.0 upgrade, and this PR will place all jobs in the `default` node pool when that happens.

tgross:
> I'm still not sure if this is an actual possibility, but if you somehow end up with a namespace that has a default node pool, and a job in that namespace that still doesn't have a node pool value, then on restore we need to use the namespace default.

Once you're upgraded to 1.6, I'm pretty sure any code path where we'd fail to set the pool by the time it hit the state store would be a bug. At which point, it seems better to force the job into the default pool where it'll be easily detected as buggy rather than crashing the server.

lgfa29:
> Once you're upgraded to 1.6, I'm pretty sure any code path where we'd fail to set the pool by the time it hit the state store would be a bug. At which point, it seems better to force the job into the default pool where it'll be easily detected as buggy rather than crashing the server.

Yeah, good point. Looking at the code a bit more, I don't think it would even be possible to access the namespace during a job restore.

I was also going to add the comment I mentioned in https://github.com/hashicorp/nomad/pull/17765#issuecomment-1620840110 but then I realized that we could catch that with a test and, in fact, [we already have one](https://github.com/hashicorp/nomad-enterprise/blob/1d3e9b7279fb37e62033c2b053e38d9d1571fa1c/nomad/job_endpoint_ent_test.go#L373-L378) so the original code would have broken int 😄 

Unfortunately it has to be in the ENT repo, so the break would only be detected after the merge.

All of these to say I think this PR is great as is 👍

PR #17767, opened 2023-06-29
Backport of Update the revision used by the docker build action. into release/1.3.x

hc-github-team-nomad-core:
## Backport

This PR is auto-generated from #17755 to be assessed for backporting due to the inclusion of the label backport/1.3.x.

---

This should always reflect the commit that's being built, as this may differ from the default `github.sha `that the workflow was invoked at.

Goes with https://github.com/hashicorp/actions-docker-build/pull/59 - and should not be merged until the PR is merged and a new version of the action is cut.


tgross:
Updated to backport anything that had not been backported (except for the 386 deprecation)

PR #17775, opened 2023-06-30
CI: Dynamic runs-on values for oss/ent

gulducat:
In order to fetch secrets from Vault in our Enterprise repo, we need to use self-hosted runners, but here in this public repo, we may only use public runners.

To reduce merge conflicts between OSS and ENT (and backports..), this PR adds conditions that will result in public runners here in OSS, but self-hosted runners in ENT.  I have done a lot of trial-and-error to end up with what I hope is a nice middle-ground between complexity, explicitness, and non-mergeflictery.

A subsequent PR will add the steps to (again, conditionally) fetch secrets from Vault.

### DIFF
--- a/.github/actions/setup-js/action.yml
+++ b/.github/actions/setup-js/action.yml
@@ -1,0 +1,34 @@
+name: setup-js
+description: install node and yarn, and run yarn install
+inputs:
+  node-version:
+    description: 'node version'
+    # https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md
+    # package.json has ember-cli at version ~3.28.5 currently
+    default: '16'
+runs:
+  using: composite
+  steps:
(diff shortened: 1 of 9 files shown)


tgross (approved):
LGTM!

I know we talked this thru a bunch and we were worried it was going to be super janky, but I think the end result isn't nearly as bad as we feared. Nice work.

gulducat:
Heya @philrenaud and/or @DingoEatingFuzz since ya touched the test-ui pipeline recently -- let me know if you have any concerns about https://github.com/hashicorp/nomad/pull/17775/commits/3ea680968d12e388c6bde424417e7e205e14837d ?  self-hosted runners are different enough from github-hosted that test-ui needed some extra finagling.

philrenaud (approved):

PR #17783, opened 2023-06-30
[ui] Affix the page header to the top of the page

philrenaud:
A report shows a 3rd party browser extension puts a banner at the top of page and awkwardly shifts nav; this fixes that

image

### DIFF
--- a/ui/app/styles/components/page-layout.scss
+++ b/ui/app/styles/components/page-layout.scss
@@ -12,6 +12,7 @@
     position: fixed;
     width: 100%;
     z-index: $z-header;
+    top: 0;
 
     // Defensive styles in case header height goes over 100px, causing
     // the left gutter menu to be on top of the header.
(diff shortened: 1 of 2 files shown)


ui/app/styles/components/page-layout.scss
```diff
@@ -12,6 +12,7 @@
     position: fixed;
     width: 100%;
     z-index: $z-header;
+    top: 0;
```
philrenaud, line 15: A little context for this: this is a `position: fixed` element, which means that you can scroll your browser generally, and wherever this element is when it started, it'll remain there. Perfect for headers, footers, etc.

A `position: <anything but static>` element's `top`, `bottom`, `right` or `left` properties will affect how near it is to some container. for `position: absolute`, this is the distance from the closest parent element with `relative` positioning. For `position: fixed` like this one, this is the distance from the browser's border.

As such, `top: 0` indicates that it should always start at the very upper edge of the browser. Which it was doing already, unless some extension slots in before it to bump it down a few pixels. This just says "Even if that happens, start at y=0"

tgross (approved):
LGTM!

PR #17792, opened 2023-07-04
Deprecate the Original Bootstrap Token Code

lhaig:
Remove Deprecated BootStrap Code

### DIFF
--- a/api/acl.go
+++ b/api/acl.go
@@ -77,8 +77,9 @@ func (c *Client) ACLTokens() *ACLTokens {
 	return &ACLTokens{client: c}
 }
 
-// DEPRECATED: will be removed in Nomad 1.5.0
 // Bootstrap is used to get the initial bootstrap token
+//
+// See BootstrapOpts to set ACL bootstrapping options.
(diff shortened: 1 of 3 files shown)


api/acl.go
```diff
@@ -77,17 +77,6 @@ func (c *Client) ACLTokens() *ACLTokens {
 	return &ACLTokens{client: c}
 }
 
-// DEPRECATED: will be removed in Nomad 1.5.0
-// Bootstrap is used to get the initial bootstrap token
-func (a *ACLTokens) Bootstrap(q *WriteOptions) (*ACLToken, *WriteMeta, error) {
```
schmichael, line 82: If calling this method continues to perform the operation it always has (bootstrapping with a server-generated token), then I say we just leave it. If we didn't want to break backward compatibility by adding options to this method, then we shouldn't break backward compatibility by removing the method.

Switching the tests is fine. Perhaps switch `// DEPRECATED:` to:

```
// Bootstrap is used to get the initial bootstrap token
//
// See BootstrapOpts to set ACL bootstrapping options.
```

schmichael (approved):
Thanks Lance!

PR #17794, opened 2023-07-04
Add acquire and release lock functions on the state store

Juanadelacuesta:
This PR adds the implementation of the acquire and release calls on the state store. If an acquire call is made on a variable that doesn't exist, it will be created. For the release, the lock ID is verified before proceeding with the release.

Related: https://github.com/hashicorp/nomad/issues/17449
Targets: feature branch

### DIFF
--- a/nomad/fsm.go
+++ b/nomad/fsm.go
@@ -2249,7 +2249,8 @@ func (f *FSMFilter) Include(item interface{}) bool {
 	return true
 }
 
-func (n *nomadFSM) applyVariableOperation(msgType structs.MessageType, buf []byte, index uint64) interface{} {
+func (n *nomadFSM) applyVariableOperation(msgType structs.MessageType, buf []byte,
+	index uint64) interface{} {
 	var req structs.VarApplyStateRequest
 	if err := structs.Decode(buf, &req); err != nil {
 		panic(fmt.Errorf("failed to decode request: %v", err))
@@ -2265,6 +2266,10 @@ func (n *nomadFSM) applyVariableOperation(msgType structs.MessageType, buf []byt
 		return n.state.VarDeleteCAS(index, &req)
 	case structs.VarOpCAS:
 		return n.state.VarSetCAS(index, &req)
+	case structs.VarOpLockAcquire:
+		return n.state.VarLockAcquire(index, &req)
+	case structs.VarOpLockRelease:
+		return n.state.VarLockRelease(index, &req)
 	default:
 		err := fmt.Errorf("Invalid variable operation '%s'", req.Op)
 		n.logger.Warn("Invalid variable operation", "operation", req.Op)
--- a/nomad/fsm.go
+++ b/nomad/fsm.go
@@ -2250,7 +2250,7 @@ func (f *FSMFilter) Include(item interface{}) bool {
 }
 
 func (n *nomadFSM) applyVariableOperation(msgType structs.MessageType, buf []byte,
-	index uint64) interface{} {
+	index uint64) any {
 	var req structs.VarApplyStateRequest
 	if err := structs.Decode(buf, &req); err != nil {
 		panic(fmt.Errorf("failed to decode request: %v", err))
--- a/nomad/state/state_store_variables.go
+++ b/nomad/state/state_store_variables.go
@@ -4,6 +4,7 @@
 package state
 
 import (
+	"errors"
 	"fmt"
 	"math"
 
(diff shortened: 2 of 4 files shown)


nomad/fsm.go
```diff
@@ -2249,7 +2249,8 @@ func (f *FSMFilter) Include(item interface{}) bool {
 	return true
 }
 
-func (n *nomadFSM) applyVariableOperation(msgType structs.MessageType, buf []byte, index uint64) interface{} {
+func (n *nomadFSM) applyVariableOperation(msgType structs.MessageType, buf []byte,
+	index uint64) interface{} {
```
tgross, line 2253: As long as we're touching it:

```suggestion
	index uint64) any {
```

nomad/state/state_store_variables.go
```diff
@@ -511,3 +518,83 @@ func (s *StateStore) VariablesQuotaByNamespace(ws memdb.WatchSet, namespace stri
...
+
+	sv, _ := raw.(*structs.VariableEncrypted)
+
+	if sv.Lock == nil || sv.Lock.ID == "" {
+		// Should this be a conflict?
+		return req.ErrorResponse(idx, errLockNotFound)
+	}
+
+	if sv.Lock.ID != req.Var.Lock.ID {
+		// Should this be a conflict?
```
tgross, line 582: This does feel like it should be a conflict, because this condition implies we tried to release someone else's lock.

nomad/state/state_store_variables.go
```diff
@@ -511,3 +518,83 @@ func (s *StateStore) VariablesQuotaByNamespace(ws memdb.WatchSet, namespace stri
...
+	defer tx.Abort()
+
+	// Look up the entry in the state store.
+	raw, err := tx.First(TableVariables, indexID, req.Var.Namespace, req.Var.Path)
+	if err != nil {
+		return req.ErrorResponse(idx, fmt.Errorf("failed variable lookup: %s", err))
+	}
+
+	if raw == nil {
+		// Should this be a conflict?
```
tgross, line 570: We could make this and the "lock not found" error below return an error saying it's a conflict, but I think that makes concurrency bugs harder to debug in the caller. Ex. if they delete a variable and then try to release it in a different goroutine, it'll be easier for them to figure out what they did wrong if we say "not found" than calling it a conflict.

nomad/state/state_store_variables.go
```diff
@@ -511,3 +518,83 @@ func (s *StateStore) VariablesQuotaByNamespace(ws memdb.WatchSet, namespace stri
...
+
+	// If the variable exist, we must make sure it doesn't hold a lock already
+	if raw != nil {
+		sv := raw.(*structs.VariableEncrypted)
+		if sv.VariableMetadata.Lock != nil {
+			return req.ErrorResponse(idx, errVarAlreadyLocked)
+		}
+
+		// Update the data to avoid overwriting the variable's items while creating the lock
+		req.Var.VariableData = sv.VariableData
```
tgross, line 543: I'm not sure I understand this bit. I'm looking at the [NMD-178 (internal link)](https://docs.google.com/document/d/1OHUUNRMewMCFPi6z4VFUs0reE_AELE9zpJUg8I-rkuk/edit#heading=h.t91xls314i) RFC and it looks like the intent was that we'd use the existing Upsert API for the lock upserts too. Which means that this would drop any updates to the variable contents that were included with the lock upsert.

It'd definitely be weird for a user to do that, but it doesn't seem like the API prevents it and it would silently drop their changes.

That also makes me wonder how we'd handle someone locking _and_ setting the CAS flags.

Juanadelacuesta: it was indeed the intent, reuse the variable upserts for both renew and release lock, because the operations end up being a write to the variable in the end, but it does leaves us with some interesting choices on the usability side.

We have not talked about locks + CAS so far,  but it is possible for the user to end up modifying or deleting the lock while upserting the variable, any ideas on that one?

tgross: > it is possible for the user to end up modifying or deleting the lock while upserting the variable, any ideas on that one?

In which case, maybe we don't actually want to have a separate state store "operation" for this feature and instead add handling of the lock to the existing `VarSet` and `VarSetCAS` operations. 

We could also treat the lock ID of a locked variable like a CAS protection as well. That way:
* If you try to change a locked variable w/o having the lock ID, you'll get a conflict
* If you try to change the locked variable while having the lock ID but having the wrong CAS index, you'll get a conflict
* If you try to change the locked variable while having the lock ID and the right CAS index, it's updated
* If you try to change the locked variable while having the lock ID and no CAS index, it's updated

nomad/state/state_store_variables.go
Juanadelacuesta: Done!

nomad/state/state_store_variables.go
Juanadelacuesta: The acquire or release a lock, some extra validations need to be done, inside one transaction. VarSet and VarSetCAS don't have them, forcing us to create a new operation.

Juanadelacuesta: lock ID check added to CAS upsert and both simple and CAS delete :D

pkazmierczak (changes requested):
Nice one @Juanadelacuesta 🎉 

I am requesting changes, because `VarLockAcquire` method demands some sort of discussion about its design. I'm somewhat confused about its intended behavior, I left a comment there with more details.

nomad/state/state_store_variables.go
```diff
@@ -177,10 +184,12 @@ func (s *StateStore) varSetCASTxn(tx WriteTxn, idx uint64, req *structs.VarApply
...
 	}
 
-	// If the existing index does not match the provided CAS index arg, then we
-	// shouldn't update anything and can safely return early here.
-	if ok && sv.ModifyIndex != svEx.ModifyIndex {
-		return req.ConflictResponse(idx, svEx)
+	if ok {
+		// If the existing index does not match the provided CAS index arg, then we
+		// shouldn't update anything and can safely return early here.
+		if sv.ModifyIndex != svEx.ModifyIndex {
```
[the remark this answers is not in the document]
Juanadelacuesta: it is, the change is a remanent of an Idea I didn't follow, let me change it back

pkazmierczak (approved):
LGTM!

Issue #13028, opened 2022-05-16
Allocations with CSI volume fail during Nomad agent restart


rbastiaans-tc:
### Nomad version

```Nomad v1.3.0 (52e95d64113e01be05d585d8b4c07f6f19efebbc)```

### Operating system and Environment details

Debian 10 Buster

### Issue

* Allocations using CSI volume will fail on client after Nomad agent restart
* while the CSI node and controller jobs stayed running and were unaffected by restart

### Reproduction steps

* Run CSI node/controller jobs as system job (on all nodes)
* Run Nomad job using CSI volume
* Restart nomad agent on client where Job with CSI volume runs by running:
  * `systemctl restart nomad`

#### Expected Result

* Allocation with CSI volume to stay running

#### Actual Result

* Allocation fails and is rescheduled

### Unit file

```
$ cat /lib/systemd/system/nomad.service
[Unit]
Description=Nomad
Documentation=https://nomadproject.io/docs/
Wants=network-online.target
After=network-online.target
ConditionPathExists=/etc/nomad.d

# When using Nomad with Consul it is not necessary to start Consul first. These
# lines start Consul before Nomad as an optimization to avoid Nomad logging
# that Consul is unavailable at startup.
Wants=consul.service
After=consul.service

[Service]
ExecReload=/bin/kill -HUP $MAINPID
ExecStart=/usr/local/bin/nomad agent -config /etc/nomad.d
KillMode=process
KillSignal=SIGINT
LimitNOFILE=infinity
LimitNPROC=infinity
Restart=on-failure
RestartSec=2
StartLimitBurst=3
StartLimitIntervalSec=10
TasksMax=infinity

[Install]
WantedBy=multi-user.target
```

### Job file (if appropriate)

CSI plugin:
```
$ cat cinder_csi.hcl 
job "cinder_csi" {
  datacenters = ["dc1"]
  type = "system"

  update {
    max_parallel = 1
    min_healthy_time = "10s"
  }

  group "node" {

    vault {
      policies = ["nomad_jobs_policy"]
    }

    task "node" {
      driver = "docker"

      config {
        image = "docker.io/k8scloudprovider/cinder-csi-plugin:v1.23.0"
        dns_servers = ["172.17.0.1"]
        privileged = true

        devices = [
          {
            host_path      = "/dev"
            container_path = "/dev"
          },
        ]

        args = [
          "/bin/cinder-csi-plugin",
          "--endpoint=unix:///csi/csi.sock",
[...]

Issue #15415, opened 2022-11-29
CSI: volume has exhausted its available writer claims


ygersie:
### Nomad version
1.4.1-ent

### Issue
We're still running into the issue of jobs getting stuck pending waiting for a volume claim to be released as described here: https://github.com/hashicorp/nomad/issues/12346#issuecomment-1135059394. The trigger seems to be related to https://github.com/hashicorp/nomad/issues/8336. Multiple jobs were stopped / started at the same time, the EBS controller threw an error `Attachment point /dev/xvdba is already in use`. The allocations were in a failed state and consequent job stops and starts now incorrectly report there's already a claim.

I've checked the AWS side but the volumes are not actually mounted to any instances, nor does a `nomad volume status` report associated allocs (possibly due to GC). When stopping + purging the respective job the volumes that failed to be claimed look like:
```
ID                 Name               Plugin ID  Schedulable  Access Mode
zookeeper[0]       zookeeper[0]       aws-ebs    true         single-node-writer
zookeeper[1]       zookeeper[1]       aws-ebs    true         single-node-writer
zookeeper[2]       zookeeper[2]       aws-ebs    true         <none>
```
The first two volumes throw claim errors and have "Access Mode" still set to `single-node-writer` and a volume status reports: `No allocations placed`. At some point after multiple job stops, running the job now fails to be placed with:
```
Constraint "CSI volume zookeeper[0] has exhausted its available writer claims and is claimed by a garbage collected allocation 138a0fe3-8f9a-6176-014a-d46b7ee37249; waiting for claim to be released
```

### Reproduction steps
Try concurrently spinning up multiple allocations at once to trigger issue #8336. Then stop/start jobs that failed before.

lgfa29:
Thanks for the report @ygersie. 

We will need some more time to investigate this as it's not immediately clear to me what's going on. I've added this to our backlog so we can triage it more.

ygersie:
@lgfa29 thank you. For now I've reduced the controller count to 1 instead of 2 since the [NewDevice method](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/563a72bc88850ace60a62a05860d3986b659801f/pkg/cloud/devicemanager/manager.go#L109) has a mutex to seemingly prevent this issue from occurring. Either way Nomad should release the volume after the controllerpublishvolume method failed with an error which it doesn't seem to do.

Xuxe:
Hey, 
[...]

PR #17840, opened 2023-07-07
CSI: persist previous mounts on client to restore during restart

tgross:
When claiming a CSI volume, we need to ensure the CSI node plugin is running before we send any CSI RPCs. This extends even to the controller publish RPC because it requires the storage provider's "external node ID" for the client. This primarily impacts client restarts because the client's initial fingerprint won't include restored plugins.

Unfortunately there's no mapping of volume to plugin ID available in the jobspec, so we don't have enough information to wait on plugins until we either get the volume from the server or retrieve the plugin ID from data we've persisted on the client.

If we always require getting the volume from the server before making the claim, a client restart for disconnected clients will cause all the allocations that need CSI volumes to fail. Even while connected, checking in with the server to verify the volume's plugin before trying to make a claim RPC is inherently racy, so we'll leave that case as-is and it will fail the claim if the node plugin needed to support a newly-placed allocation is flapping such that the node fingerprint is changing.

So instead of blocking for information from the server, this changeset persists a minimum subset of data about the volume and its plugin in the client state DB, and retrieves that data during the CSI hook's prerun to avoid re-claiming and remounting the volume unnecessarily. This fixes the client restart bug but also reduces Nomad RPC and CSI RPC traffic during client restarts.

This changeset also updates the RPC handler to use the external node ID from the claim whenever it is available.

Fixes: #13028

---

Note for reviewers: this looks like a bit of a monster of a PR because of all the plumbing required for adding a new client state DB object. The bulk of the non-plumbing changes are in `client/allocrunner/csi_hook.go`, which GitHub is helpfully folding. :grinning:  An unfortunate amount of the churn in that file is having to break up the `claimVolume` function a bit, but I think the top-level flow in `Prerun` is a lot more legible as a result. In addition to expanding the test coverage in the unit test, I've also end-to-end tested this with the AWS EBS plugin and restarting the client in various configurations. (Also note this will not do anything for #15415)

### DIFF
--- a/.changelog/17840.txt
+++ b/.changelog/17840.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+csi: Fixed a bug where CSI volumes would fail to restore during client restarts
+```
--- a/client/allocrunner/alloc_runner.go
+++ b/client/allocrunner/alloc_runner.go
@@ -1460,3 +1460,20 @@ func (ar *allocRunner) GetUpdatePriority(a *structs.Allocation) cstructs.AllocUp
 
 	return cstructs.AllocUpdatePriorityNone
 }
+
+func (ar *allocRunner) SetCSIVolumes(vols map[string]*state.CSIVolumeStub) error {
+	return ar.stateDB.PutAllocVolumes(ar.id, &state.AllocVolumes{
+		CSIVolumes: vols,
+	})
+}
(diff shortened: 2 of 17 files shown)


jrasell (approved):
LGTM!

PR #17841, opened 2023-07-07
ci: pull secrets from Vault in nomad-enterprise

gulducat:
Now that (most) secrets-requiring workflows run on self-hosted runners in nomad-enterprise (#17775), we're able to (conditionally) pull secrets from our internal CI Vault cluster.

The approach here uses a [composite action](https://docs.github.com/en/actions/creating-actions/about-custom-actions#composite-actions) to avoid repetitive boilerplate, which also means it needs* to set secrets as env vars (also the default behavior of hashicorp/vault-action, and common practice in other projects, sans the composite action).  Otherwise, it would require an odd maneuver, for GHA reasons, to have the composite output a dynamic map instead (*but it is possible, if we really want to. ask me how if you'd like).

As with the aforementioned PR, this is here in OSS Nomad repo to avoid merge conflicts with enterprise.  I've pre-tested the workflows there, so hopefully no runtime surprises when we go to merge oss->ent.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -38,6 +38,11 @@ jobs:
     if: always() && needs.backport.result == 'failure'
     runs-on: ${{ endsWith(github.repository, '-enterprise') && fromJSON('["self-hosted", "ondemand", "linux"]') || 'ubuntu-latest' }}
     steps:
+      - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+      - uses: ./.github/actions/vault-secrets
+        with:
+          paths: |-
+            kv/data/teams/nomad/slack-webhooks feed-nomad | SLACK_FEED_NOMAD ;
       - name: Send slack notification on failure
         uses: slackapi/slack-github-action@007b2c3c751a190b6f0f040e47ed024deaa72844 # v1.23.0
         with:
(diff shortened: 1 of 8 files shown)


.github/workflows/backport.yml
```diff
@@ -70,8 +75,8 @@ jobs:
               ]
             }
         env:
-          SLACK_WEBHOOK_URL: ${{ secrets.BACKPORT_ASSISTANT_FAILURE_SLACK }}
+          SLACK_WEBHOOK_URL: ${{ env.SLACK_FEED_NOMAD || secrets.BACKPORT_ASSISTANT_FAILURE_SLACK }}
           SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
 permissions:
   contents: read
-
+  id-token: write
```
tgross, line 82: What's this change?

tgross (approved):
LGTM!

.github/workflows/backport.yml
gulducat: high-level explanation is that the [`vault-auth` helper script](https://github.com/hashicorp/nomad/pull/17841/files#diff-2ccd1b5860617862a60112949549f7739cf06ecc92cc2511d60dcbdd5c96e4d6R13) on our self-hosted runners needs it to provide a Vault token based on the repo of the workflow run.

longer explanation of GHA's OIDC setup is hereabouts: https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#updating-your-actions-for-oidc
> The `id-token: write` setting allows the JWT to be requested from GitHub's OIDC provider [...]

I could not find a way to make this conditional for ent-only -- it must be a static string.

Issue #17751, opened 2023-06-27
include parent job ID as a Docker container label


EtienneBruines:
### Proposal
Next to the `container.labels.com_hashicorp_nomad_job_id` field that is already being set, have a field that excludes the `/periodic-901283123` suffix for batch jobs.

My feeling tells me the `container.labels.com_hashicorp_nomad_job_name` field would make sense for that (since right now, they have the same value) - but that would break backwards compatibility a bit. 

Any field would suffice. In a lot of cases, I only care about the job name as provided in the `.nomad.hcl` file, and not have to bother with whatever suffixes were added. 

### Use-cases
Logging / filtering / grouping based on that value, while not caring about the timestamp that the batch job.

### Attempted Solutions
In my Elasticsearch I now need to setup an ingest pipeline to remove this `/periodic-*` suffix, because I want to filter on the name of the job (without the timestamp).

tgross:
Hi @EtienneBruines! That seems like a good idea. I'll mark this for roadmapping but it's also small enough that I'll mark it as a good first issue to see if anyone from the community wants to give a try at adding it (they might start looking in [`docker/driver.go`](https://github.com/hashicorp/nomad/blob/main/drivers/docker/driver.go)

devashishTaneja:
@tgross I would like to contribute to this issue.

tgross:
Go for it! Feel free to tag me o the PR and I'll make sure it gets reviewed.

devashishTaneja:
Hi @tgross, @EtienneBruines,

This is my first contribution to Nomad so I am lacking some context. I would appreciate if you could help me with a couple of doubts. So I was able to setup Nomad in my local to start a periodic docker. When I do a _docker inspect_ on the container. I can see the following labels set up by default i.e. _dockerLabelAllocID_ but no _com_hashicorp_nomad_job_id_ label


<img width="814" alt="Screenshot 2023-07-07 at 9 36 31 PM" src="https://github.com/hashicorp/nomad/assets/32792385/a754c239-c430-437a-b484-d28e26177f6c">


But I do see a similar env set (screenshot attached below) so my question is the ask to add a new env or label itself. If labels can you suggest me how these labels are set in the first place. Below is the config file I used to start the periodic docker.


<img width="814" alt="Screenshot 2023-07-07 at 9 36 41 PM" src="https://github.com/hashicorp/nomad/assets/32792385/783406bf-232b-47bb-a392-73fbf10a04a6">



```
job "simple" {
[...]

PR #17843, opened 2023-07-07
Include parent job ID as a Docker container label #17751

devashishTaneja:
Fixes: #17751


drivers/docker/driver.go
```diff
@@ -1324,6 +1325,9 @@ func (d *Driver) createContainerConfig(task *drivers.TaskConfig, driverConfig *T
 		if glob.Glob(configurationExtraLabel, "job_id") {
 			labels[dockerLabelJobID] = task.JobID
 		}
+		if glob.Glob(configurationExtraLabel, "parent_job_id") {
+			labels[dockerLabelParentJobID] = task.ParentJobId
+		}
```
tgross, line 1330: Most jobs won't have a parent job ID (the value will be `""`), only dispatch and periodic jobs do. So we probably don't want to add the label if it's not set. That'll also change the test case you have below.

client/allocrunner/taskrunner/task_runner.go
```diff
@@ -1144,6 +1144,7 @@ func (tr *TaskRunner) buildTaskConfig() *drivers.TaskConfig {
 		Namespace:     alloc.Namespace,
 		NodeName:      alloc.NodeName,
 		NodeID:        alloc.NodeID,
+		ParentJobId:   alloc.Job.ParentID,
```
tgross, line 1147: Nitpicky golang style thing:

```suggestion
		ParentJobID:   alloc.Job.ParentID,
```

plugins/drivers/driver.go
```diff
@@ -274,6 +274,7 @@ type TaskConfig struct {
 	JobName          string
 	JobID            string
 	TaskGroupName    string
+	ParentJobId      string
```
tgross, line 277:
```suggestion
	ParentJobID      string
```

We also will need to add this field to the `TaskConfig` message in [`driver.proto`](https://github.com/hashicorp/nomad/blob/main/plugins/drivers/proto/driver.proto). The reason this works without it is because the Docker driver is internal to Nomad so we don't use the protobuf messages to talk to it. But if we leave the new field out of the protobuf it would mean that it would be unavailable to external drivers like podman or containerd.

tgross (changes requested):
Looking pretty good @devashishTaneja! In addition to the comments I've left, can you run `make cl` to add a changelog entry?

---

I've smoke tested this with the following jobspec:

```hcl
job "example" {
  type        = "batch"

  parameterized {
    payload = "required"
  }

  group "group" {
    task "task" {
      driver = "docker"
      config {
        image   = "busybox:1"
        command = "/bin/sh"
        args    = ["-c", "cat local/payload.txt; sleep 300"]
      }
      dispatch_payload {
        file = "payload.txt"
      }
    }
  }
}
```

After running and dispatching that job, I get the following labels on the resulting container:

```
$ docker inspect 120 | jq '.[0].Config.Labels'
{
  "com.hashicorp.nomad.alloc_id": "45b9fe13-0c85-9b87-0f3d-1c82774d9b5f",
  "com.hashicorp.nomad.job_id": "example/dispatch-1688760549-ec1a5470",
  "com.hashicorp.nomad.job_name": "example/dispatch-1688760549-ec1a5470",
  "com.hashicorp.nomad.parent_job_id": "example",
  "com.hashicorp.nomad.task_name": "task"
}
```

drivers/docker/driver.go
devashishTaneja: @tgross I understand that the parent job id will be empty if the job is not dispatch or periodic. So should we check for an extra condition if the  **_task.ParentJobId_** is empty string, then we skip the label?

tgross: Right!

devashishTaneja:
@tgross All the review comments are taken care. Please check

devashishTaneja:
@tgross Please have a look when possible.

.changelog/17843.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+build: Added support for parent job id for periodic/dispatch jobs
```
tgross, line 2:
```suggestion
docker: Added option for labeling container with parent job ID of periodic/dispatch jobs
```

plugins/drivers/proto/driver.proto
```diff
@@ -495,6 +495,9 @@ message TaskConfig {
 
     // NodeId is the ID of the node where the associated allocation is running
     string node_id = 21;
+
+    // ParentJobID is the parent id for dispatch and periodic jobs
+    string parent_job_id = 22;
```
tgross, line 500: `plugins/drivers/proto/driver.pb.go` is now out of sync and needs to be regenerated as well. `make proto` will do so.

tgross (changes requested):
Looking good @devashishTaneja! Just a couple more items to wrap up.

tgross (approved):
LGTM! Thanks @devashishTaneja !

Issue #17833, opened 2023-07-07
Nomad retries indefinitely to revoke non-existant SI token(s)


seanamos:
### Nomad version
v1.5.6

### Issue
We were doing DR testing in one of our testing environments and did a snapshot restore to a 1 hour old consul snapshot. Since then, nomad has been trying to revoke SI tokens from Consul that don't exist:

```
# warnings in Nomad

failed to revoke SI token accessor (alloc "8e6140ce-d940-1028-83c8-ba2e306d0678", node "4cc0e923-7d21-1f02-b100-f9baa12dc654", task "xxx"): Unexpected response code: 404 (Cannot find token to delete)
```

Consul is generating lots of `Cannot find token to delete` errors as well.

These are the same tokens that are failing to delete (total attempts for each URL over the last 3 days):

<img width="501" alt="Screenshot 2023-07-07 at 09 58 09" src="https://github.com/hashicorp/nomad/assets/10598927/a752f940-e7f6-45f0-812b-62afda37b893">

Questions:
1. Nomad should possibly handle the 404 gracefully (token does not exist)? 
2. Is there a way to force nomad to forget about these SI tokens?

tgross:
Hi @seanamos! I took a quick look and Nomad is assuming that this endpoint will not return an error (ref [`consul.go#L419-L422`](https://github.com/hashicorp/nomad/blob/v1.5.6/nomad/consul.go#L419-L422)):

```go
// Consul will no-op the deletion of a non-existent token (no error)
_, err := c.aclClient.TokenDelete(accessor.AccessorID, &api.WriteOptions{Namespace: accessor.ConsulNamespace})
return err
```

Unfortunately it looks like this assumption was broken in a breaking change in [Consul 1.15.0](https://github.com/hashicorp/consul/blob/main/CHANGELOG.md#1150-february-23-2023):

> * acl errors: Delete and get requests now return descriptive errors when the specified resource cannot be found. Other ACL request errors provide more information about when a resource is missing. Add error for when the ACL system has not been bootstrapped.
>   * Delete Token/Policy/AuthMethod/Role/BindingRule endpoints now return 404 when the resource cannot be found.
>     * New error formats: "Requested * does not exist: ACL not found", "* not found in namespace $NAMESPACE: ACL not found"

[...]

PR #17847, opened 2023-07-07
consul: handle "not found" errors from Consul when deleting tokens

tgross:
In Consul 1.15.0, the Delete Token API was changed so as to return an error when deleting a non-existent ACL token. This means that if Nomad successfully deletes the token but fails to persist that fact, it will get stuck trying to delete a non-existent token forever.

Update the token deletion function to ignore "not found" errors and treat them as successful deletions.

Fixes: #17833
Source for text of error (there's no const in the `api` package for it): https://github.com/hashicorp/consul/blob/api/v1.22.0/agent/acl_endpoint.go#L470

### DIFF
--- a/.changelog/17847.txt
+++ b/.changelog/17847.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+consul: Fixed a bug where Nomad would repeatedly try to revoke successfully revoked SI tokens
+```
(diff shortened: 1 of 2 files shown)


shoenig (approved):
LGTM

PR #17852, opened 2023-07-07
ci: more self-hosted iops for checks workflow

gulducat:
Avoid hitting the timeout on nomad-enterprise.
These values have worked consistently in my tests over yonder.


tgross (approved):
LGTM

shoenig (approved):

Issue #17804, opened 2023-07-05
Multiple cron specs


aneustroev:
### Proposal
I want to make multiple cron specifications. Sometimes it useful, because impossible to make any time conditions using cron. 

### Use-cases
Some batch job required non linear intervals between runs.
For example

- every day at 6:30  `30 6 * * *`
- every day at 15:00  `0 15 * * *`
- every day at 20:45  `45 20 * * *`

### Attempted Solutions
In source code I see that I can set only one cron string.
Make several periodic jobs is dirty and bad for observability.

aneustroev:
Can you make me a advice, how to implement it feature better?

aneustroev:
https://github.com/hashicorp/nomad/pull/17811

aneustroev:
https://github.com/hashicorp/nomad/pull/17858

lgfa29:
Thanks for the idea and the PR @aneustroev!  I left a review you there 🙂

PR #17858, opened 2023-07-10
Add multiple times in periodic jobs

aneustroev:
Closes https://github.com/hashicorp/nomad/issues/17804


ui/app/templates/components/job-page/periodic.hbs
```diff
@@ -25,7 +25,7 @@
           <span class="term">
             Cron
           </span>
-          {{@job.periodicDetails.Spec}}
+          {{#each @job.periodicDetails.Specs as |spec|}}{{spec}}; {{else}}{{@job.periodicDetails.Spec}}{{/each}}
```
lgfa29, line 28:
```suggestion
          {{#each @job.periodicDetails.Specs as |spec|}}
            <span class="bumper-right tag">{{spec}}</span>
          {{else}}
            <span class="tag">{{@job.periodicDetails.Spec}}</span>
          {{/each}}
```

While I think `;` is not a reserved character for cron expression, it does look quite a bit like a `,` which is valid, so it may be tricky to notice the separation between expression:

image

I think using a pill here makes the distinction easier to read:
image


But pinging @philrenaud and @juliezzhou for extra input 😄

lgfa29 (changes requested):
Looking great @aneustroev!

I left some comments, but I think the biggest blocker is that the PR is missing setting the `SpecType` here:
https://github.com/hashicorp/nomad/blob/437941816c6b84f4b756f694fe42cb028ff50e65/jobspec2/parse_job.go#L22-L25

Without this a simple `nomad job run` panics:
```
$ nomad run example.nomad.hcl
Job registration successful
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x10332eb78]

goroutine 1 [running]:
github.com/hashicorp/nomad/api.(*PeriodicConfig).Next(0x14000745d10?, {0x103d7bdfa?, 0x140005bfca0?, 0x0?})
        github.com/hashicorp/nomad/api@v0.0.0-20230103221135-ce00d683f9be/jobs.go:858 +0x28
github.com/hashicorp/nomad/command.(*JobRunCommand).Run(0x140003ff000, {0x1400011a1a0, 0x1, 0x1})
        github.com/hashicorp/nomad/command/job_run.go:365 +0xd88
github.com/mitchellh/cli.(*CLI).Run(0x14000bd6640)
        github.com/mitchellh/cli@v1.1.5/cli.go:262 +0x4a8
main.Run({0x1400011a190, 0x2, 0x2})
        github.com/hashicorp/nomad/main.go:110 +0x29c
main.main()
        github.com/hashicorp/nomad/main.go:80 +0x50
```

In theory we don't update the `jobspec` package anymore, but given the potential for a panic I think it may be worth updating here as well:
https://github.com/hashicorp/nomad/blob/437941816c6b84f4b756f694fe42cb028ff50e65/jobspec/parse_job.go#L252-L256

Let me know if you have any questions. Thanks for the PR!

ui/app/templates/components/job-page/periodic.hbs
philrenaud: Pills look great! I am not sure if `cron` is singular where `crons` is plural, but if that is the case, you can also switch your `term` code to be something like: `{{pluralize "Cron" @job.periodicDetails.Specs.length}}`

nomad/structs/structs.go
```diff
@@ -5610,7 +5614,10 @@ func (p *PeriodicConfig) Validate() error {
 	}
 
 	var mErr multierror.Error
-	if p.Spec == "" {
+	if p.Spec != "" && len(p.Specs) != 0 {
+		_ = multierror.Append(&mErr, fmt.Errorf("You can use only Spec or Specs"))
```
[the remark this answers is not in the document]
aneustroev: I hope that you keep only `Specs` :sweat_smile: 
Because now we use systemd timers, and it's more powerful.
Moreover, even k8s doesn't have multiple cron times.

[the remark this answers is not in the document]
aneustroev: But I understand that it break backward compatibility.

ui/app/templates/components/job-page/periodic.hbs
aneustroev: Thank you for this suggestion! Frontend is very hard for me :smile:

aneustroev:
Thank you a lot for review!

api/jobs.go
```diff
@@ -852,29 +856,48 @@ func (p *PeriodicConfig) Canonicalize() {
...
+				}
+			}
+			nextTime = times[0]
+			for _, next := range times {
+				if next.Before(nextTime) {
+					nextTime = next
+				}
+			}
+			return nextTime, nil
+		}
```
lgfa29, line 880:
```suggestion
		var nextTime time.Time
		for _, spec := range p.Specs {
			t, err := cronParseNext(fromTime, spec)
			if err != nil {
				return time.Time{}, fmt.Errorf("failed parsing cron expression %s: %v", spec, err)
			}
			if nextTime.IsZero() || t.Before(nextTime) {
				nextTime = t
			}
		}
		return nextTime, nil
```

Sorry, I should've noticed this before, but I think we can avoid iterating over the times twice since we only care about keeping the earliest next time.

And since we don't have a pointer to a slice anymore we remove the `if` check and the `range` will be safe, even if `Specs` is `nil`.

nomad/structs/structs.go
```diff
@@ -5610,7 +5614,10 @@ func (p *PeriodicConfig) Validate() error {
 	}
 
 	var mErr multierror.Error
-	if p.Spec == "" {
+	if p.Spec != "" && len(p.Specs) != 0 {
+		_ = multierror.Append(&mErr, fmt.Errorf("You can use only cron or crons"))
```
lgfa29, line 5618:
```suggestion
		_ = multierror.Append(&mErr, fmt.Errorf("Only cron or crons may be used"))
```

Should've noticed this as well 😅 

We usually avoid error messages pointing to a specific subject, like `you`.

nomad/structs/structs.go
```diff
@@ -5624,8 +5631,17 @@ func (p *PeriodicConfig) Validate() error {
...
 	case PeriodicSpecCron:
 		// Validate the cron spec
-		if _, err := cronexpr.Parse(p.Spec); err != nil {
-			_ = multierror.Append(&mErr, fmt.Errorf("Invalid cron spec %q: %v", p.Spec, err))
+		if p.Spec != "" {
+			if _, err := cronexpr.Parse(p.Spec); err != nil {
+				_ = multierror.Append(&mErr, fmt.Errorf("Invalid cron spec %q: %v", p.Spec, err))
+			}
+		}
+		if len(p.Specs) != 0 {
```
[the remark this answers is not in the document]
lgfa29: I think this one was missed.

nomad/structs/structs.go
```diff
@@ -5666,11 +5685,27 @@ func CronParseNext(e *cronexpr.Expression, fromTime time.Time, spec string) (t t
...
+				return times[i], err
+			}
 		}
-		return CronParseNext(e, fromTime, p.Spec)
+		nextTime = times[0]
+		for _, next := range times {
+			if next.Before(nextTime) {
+				nextTime = next
+			}
+		}
```
lgfa29, line 5706: The same update from `api/jobs.go` can be applied here to avoid the second iteration and storage.

lgfa29 (changes requested):
Thanks for the updates. I left some minor changes that I missed in the previous review 😅

ui/app/templates/components/job-page/periodic.hbs
lgfa29: > Pills look great! I am not sure if `cron` is singular where `crons` is plural, but if that is the case, you can also switch your `term` code to be something like: `{{pluralize "Cron" @job.periodicDetails.Specs.length}}`

Ah nice, I think in this case we may need to handle zero length, which indicates that `cron` is being used.

```diff
          <span class="term">
-           Cron
+           {{pluralize "Cron" (or @job.periodicDetails.Specs.length 1)}}
          </span>
```

aneustroev: Thanks, I lost it

nomad/structs/structs.go
aneustroev: Very good suggestion.

.changelog/17858.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+jobspec: Add 'crons' fileld for multiple `cron` entity's
```
lgfa29, line 2:
```suggestion
jobspec: Add 'crons' fileld for multiple `cron` expressions 
```

api/jobs.go
```diff
@@ -851,30 +855,48 @@ func (p *PeriodicConfig) Canonicalize() {
 // returned. The `time.Location` of the returned value matches that of the
 // passed time.
 func (p *PeriodicConfig) Next(fromTime time.Time) (time.Time, error) {
+	// Once spec parsing
```
lgfa29, line 858:
```suggestion
	// Single spec parsing
```

api/jobs.go
```diff
@@ -851,30 +855,48 @@ func (p *PeriodicConfig) Canonicalize() {
...
+			return time.Time{}, fmt.Errorf("failed parsing cron expression %s: %v", spec, err)
+		}
+	}
+	nextTime = times[0]
+	for _, next := range times {
+		if next.Before(nextTime) {
+			nextTime = next
+		}
+	}
+	return nextTime, nil
```
lgfa29, line 881: I think https://github.com/hashicorp/nomad/pull/17858#discussion_r1264138630 was skipped here 😅

nomad/structs/structs.go
```diff
@@ -5666,11 +5685,29 @@ func CronParseNext(e *cronexpr.Expression, fromTime time.Time, spec string) (t t
 func (p *PeriodicConfig) Next(fromTime time.Time) (time.Time, error) {
 	switch p.SpecType {
 	case PeriodicSpecCron:
-		e, err := cronexpr.Parse(p.Spec)
-		if err != nil {
-			return time.Time{}, fmt.Errorf("failed parsing cron expression: %q: %v", p.Spec, err)
+		// Once spec parsing
```
lgfa29, line 5688:
```suggestion
		// Single spec parsing
```

nomad/structs/structs.go
lgfa29: This was missed as well 😄

website/content/api-docs/json-jobs.mdx
```diff
@@ -336,6 +336,9 @@ The `Job` object supports the following keys:
     [here](https://github.com/gorhill/cronexpr#implementation) for full
     documentation of supported cron specs and the predefined expressions.
 
+  - `Specs` - A cron expressions configuring the interval the job is launched
+    at. Like `Spec`, but support multiple entity.
```
lgfa29, line 340:
```suggestion
  - `Specs` - A cron expressions configuring the interval the job is launched
    at. Like `Spec`, but support multiple expressions. The job will run at the next time that matches any of the expressions.
```

website/content/docs/job-specification/periodic.mdx
```diff
@@ -38,6 +38,9 @@ consistent evaluation when Nomad spans multiple time zones.
   interval to launch the job. In addition to [cron-specific formats][cron], this
   option also includes predefined expressions such as `@daily` or `@weekly`.
 
+- `crons` - A cron expressions configuring the interval the job is launched
+  at. Like `cron`, but support multiple entity.
```
lgfa29, line 42:
```suggestion
- `crons` - A cron expressions configuring the interval the job is launched
  at. Like `cron`, but support multiple expressions. The job will run at the next time that matches any of the expressions.
```

lgfa29 (changes requested):
Thanks for the updates @aneustroev. I sent another round of review with some minor changes 😄 

I spoke with the team and were trying to figure out if there was a way to reuse the existing `cron` attribute to accept either a string or a slice of strings. I prototyped a possible approach in https://github.com/hashicorp/nomad/commit/fb2ee9eb21c5db3703e1250cc499ba02385630ff but it turns out that there's no way to do that without a breaking change, so we will need to have two fields for now.

What we can do is mark `cron` as deprecated to nudge people towards `crons`. Could you add a check here:
https://github.com/hashicorp/nomad/blob/d14362ec1903fa8d49db498c28cfdf4b5f073b22/nomad/structs/structs.go#L4772-L4801

This would generate an error like:
> cron is deprecated and may be removed in a future release. Use crons instead.

Thanks for all the work!

api/jobs.go
aneustroev:
```golang
if nextTime.IsZero() || t.Before(nextTime) {
			nextTime = t
		}
```
Very good advice! Sorry for I skip it.

lgfa29 (approved):
Thanks for the updates @aneustroev! The code is looking good to me 🙂 

I tried to push a few small touch-ups in the documentation and to make `structs.CronParseNext` to match `api.cronParseNext` but it seems like I don't have permission to edit the PR 😅 

Would you mind applying this patch?
https://github.com/hashicorp/nomad/commit/a32b851e42229955fa54bed42bf6a43c60fe6c1c.patch

nomad/structs/structs.go
```diff
@@ -4797,6 +4797,12 @@ func (j *Job) Warnings() error {
 		mErr.Errors = append(mErr.Errors, err)
 	}
 
+	// cron -> crons
+	if j.Periodic.Spec != "" {
```
lgfa29, line 4801:
```suggestion
	if j.Periodic != nil && j.Periodic.Spec != "" {
```

Ooops, one more change needed 😄

aneustroev:
Thank you for help with documentations!

lgfa29:
Thank _you_ for the great PR 🙂

PR #17889, opened 2023-07-11
ci: add copywrite action to check file headers.

### LINKED ISSUE #17887: ci: add copyright and license header check to CI runs
HashiCorp legal requires that our source code files have copyright and license headers. There is a bot which runs periodically to checks this and will raise PR's for files which are missing the header. While this process works, it adds slight overhead in PR management. It could be nice to add a CI check which ensures any new files within the PR have the header.

jrasell:
This PR's copywrite CI run: https://github.com/hashicorp/nomad/actions/runs/5521733801/jobs/10070278831?pr=17889

closes #17887

### DIFF
--- a/.github/workflows/copywrite.yml
+++ b/.github/workflows/copywrite.yml
@@ -10,6 +10,9 @@ jobs:
       - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
       - uses: hashicorp/setup-copywrite@867a1a2a064a0626db322392806428f7dc59cb3e # v1.1.2
         name: Setup Copywrite
+        with:
+          version: v0.16.4
+          archive-checksum: c299f830e6eef7e126a3c6ef99ac6f43a3c132d830c769e0d36fa347fa1af254
       - name: Check Header Compliance
         run: copywrite headers --plan
 permissions:
--- a/.github/workflows/copywrite.yml
+++ b/.github/workflows/copywrite.yml
@@ -1,0 +1,16 @@
+name: Check Copywrite Headers
+
+on:
+  push: {}
+
+jobs:
+  copywrite:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


tgross (approved):
This LGTM but will it work in the ENT repo without needing special creds? @gulducat might be able to answer that

.github/workflows/copywrite.yml
```diff
@@ -0,0 +1,19 @@
...
+      - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+      - uses: hashicorp/setup-copywrite@867a1a2a064a0626db322392806428f7dc59cb3e # v1.1.2
+        name: Setup Copywrite
+        with:
+          version: v0.16.4
+          archive-checksum: c299f830e6eef7e126a3c6ef99ac6f43a3c132d830c769e0d36fa347fa1af254
+      - name: Check Header Compliance
+        run: copywrite headers --plan
+permissions:
+  contents: read
```
gulducat, line 19: Though I don't know the full extent of what the `copywrite` command does, I imagine it should be fine.  The contents of the repo itself are pulled by `actions/checkout` in the first step, which only needs this `contents:read` here.


Beyond that, according to the [copyright repo's readme](https://github.com/hashicorp/copywrite#github-authentication), for some operations

> A GITHUB_TOKEN environment variable can be used with a Personal Access Token

I suspect the GITHUB_TOKEN (auto-provided by GHA), if used at all in our case, will mainly be to uniquely identify the caller so our private runners don't get some egress-IP rate-limited or something by GH API.

gulducat (approved):
:+1: 

this doesn't need backports though, does it?

tgross:
> this doesn't need backports though, does it?

I'd have said yes but then I actually checked the `release/1.5.x` branch and it turns out we didn't backport all the copyright headers. So I think you're right we don't need them here.

PR #17899, opened 2023-07-11
np: update docs and add test for nil lists

lgfa29:
Document and test that if a namespace does not provide an `allow` or `deny` list than those are treated as `nil` and have a different behaviour from an empty list (`[]string{}`).


tgross (approved):
LGTM!

PR #17909, opened 2023-07-11
e2e: add some e2e tests for pledge task driver

shoenig:
- e2e: setup nomad for pledge driver
- e2e: add some e2e tests for pledge task driver


pkazmierczak (approved):
LGTM!

PR #17911, opened 2023-07-12
volume-status : show namespace the volume belongs to

the-nando:
Fixes: https://github.com/hashicorp/nomad/issues/17912

### DIFF
--- a/.changelog/17911.txt
+++ b/.changelog/17911.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+cli: Display volume namespace on `nomad volume status` and `nomad node status` output
+```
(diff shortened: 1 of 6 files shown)


lgfa29 (approved):
Ah nice, thanks for the PR @the-nando!

I guess you noticed the same thing as I did in https://github.com/hashicorp/nomad/pull/17925 but did a much better job at it 😄 

We're in the release candidate period, where only significant bug fixes will be merged for now, so I will keep this open until GA happens.

the-nando:
@lgfa29 nothing urgent and thanks!
There's a couple of failing test actions which look like unrelated to this change

lgfa29:
> There's a couple of failing test actions which look like unrelated to this change

Yeah, it kind of does that sometimes 🤷

PR #17924, opened 2023-07-12
[ui] Flaky test fix: expiring token notification

philrenaud:
We've got this acceptance test in the UI that checks to make sure that, when the "10 minutes until your ACL token expires" point hits, you get a notification.

We would create a token with an expiry 10m5s out, wait 0.5s for initialization, and then wait 5s to make sure the notification is there.

With recent test runner changes (Circle -> GHA), we're noticing this fail more often. This may be because the JS runtime is different or slightly slower.

Because the user-facing notification is persistent (doesn't go away on its own), we're safe to wait a little longer before checking if the notification is present. And in order not to take any longer on our already-kind-of-long test runs, I've opted to shorten the token TTL rather than extending the setTimeout() length.


gulducat (approved):
not dead yet

Issue #17923, opened 2023-07-12
nomad node status doesn't return in use CSI volumes


the-nando:
### Nomad version

```
Nomad v1.5.6
```

### Issue

A call to `nomad node status -verbose <node_id>` for a node with an allocation using a CSI volume running on it, shows only the header for the CSI Volumes:
```
ID              = f65f89ba-1b54-71d2-bb2f-e582dac7b916
Name            = test-client
Class           = default
DC              = us-west-2b
Drain           = false
Eligibility     = eligible
Status          = ready
CSI Controllers = <none>
CSI Drivers     = aws-ebs,aws-efs
Uptime          = 150h19m3s

Host Volumes
Name                   ReadOnly  Source
test-volume  true      /data/test/foo

CSI Volumes                                                                             <--------
ID  Name  Plugin ID  Schedulable  Provider  Access Mode                                 <--------

Drivers
Driver    Detected  Healthy  Message   Time
docker    true      true     Healthy   2023-07-06T08:35:02Z
[...]
```

Volume status output:
```bash
~ nomad volume status -namespace='*' 'my-vol[1]' 
ID                   = my-vol[1]
Name                 = my-vol[1]
External ID          = vol-xxx
Plugin ID            = aws-ebs
Provider             = ebs.csi.aws.com
Version              = v1.12.1
Schedulable          = true
Controllers Healthy  = 1
Controllers Expected = 1
Nodes Healthy        = 64
Nodes Expected       = 64
Access Mode          = single-node-writer
Attachment Mode      = file-system
Mount Options        = fs_type: ext4
Namespace            = my-ns

Topologies
Topology  Segments

Allocations
ID        Node ID   Task Group    Version  Desired  Status   Created   Modified
e76669ab  f65f89ba  test  46       run      running  1d6h ago  10m20s ago
~ 
```

### Reproduction steps

1. Create a CSI volume
2. Create and run a job which uses such volume
3. Get the node ID where the allocation is scheduled on
4. Call `nomad node status -verbose <node_id>`

#### Expected Result

`nomad node status -verbose` shows all CSI volumes in use on the node.

#### Actual Result

No volumes are displayed for the node.  
I believe one of the problems is that [the call](https://github.com/hashicorp/nomad/blob/release/1.5.x/command/node_status.go#L685) to `/v1/volumes` doesn't set `namespace=*` as query params, but there seems to be more to it as setting it still doesn't return the volumes.
```
[...]

PR #17925, opened 2023-07-12
cli: search all namespaces for node volumes

lgfa29:
When looking for CSI volumes to display in the `node status` command the CLI needs to search all namespaces. ~It's also helpful to display the volume namespace in the command output.~ This has been handled by #17911.

Closes https://github.com/hashicorp/nomad/issues/17923


jrasell (approved):
LGTM!

PR #17928, opened 2023-07-13
fix: allowed metric values documentation for the Nomad APM plugin

janory:
Based on the [nomad-autoscaler](https://github.com/hashicorp/nomad-autoscaler) codebase and on the [validateMetricTaskGroupQuery](https://github.com/hashicorp/nomad-autoscaler/blob/954f26e8c56526d62f78d6991f5ff2a29dc6fa9f/plugins/builtin/apm/nomad/plugin/job.go#L273) vs [validateMetricNodeQuery](https://github.com/hashicorp/nomad-autoscaler/blob/954f26e8c56526d62f78d6991f5ff2a29dc6fa9f/plugins/builtin/apm/nomad/plugin/node.go#L197), this is how the documentation should look like.


lgfa29 (approved):
Nice catch, thanks @janory!

PR #17929, opened 2023-07-13
feat: improve policy documentation for "query_window" property

janory:
Based on the [nomad-autoscaler source code](https://github.com/hashicorp/nomad-autoscaler/blob/954f26e8c56526d62f78d6991f5ff2a29dc6fa9f/plugins/builtin/apm/nomad/plugin/metrics.go#L37), the `query_window` property is ignored, if the Nomad APM plugin is in use, because that doesn't have memory.

### DIFF
--- a/website/content/tools/autoscaling/plugins/apm/nomad.mdx
+++ b/website/content/tools/autoscaling/plugins/apm/nomad.mdx
@@ -8,8 +8,14 @@ description: The "nomad-apm" APM plugin queries the Nomad API for metrics.
 
 The Nomad APM plugin allows querying the Nomad API for metric data. This provides
 an immediate starting point without addition applications but comes at the price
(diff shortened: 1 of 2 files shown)


lgfa29 (changes requested):
Good catch @janory. But I think it would be better to document here something like:

> This value may not be supported by all APM plugins. Refer to their documentation for more details.

And then add a note to the Nomad APM plugin that this option is not used?

lgfa29 (approved):
I thought I had merged this, but went looking for it and noticed my pending review 😅 

I pushed my suggestion to expand the Nomad APM docs a bit more.

Thanks for the suggestion and contribution @janory!

janory:
> Thanks for the suggestion and contribution @janory!

You are welcome! :)

PR #17932, opened 2023-07-13
[ui] New keyboard shortcut for Create Variable

philrenaud:
Now, from a variable page, you can type in `n v`, which stands for "NeVermind why I chose those letters"
image


mikenomitch (approved):
+1 - I like "n _" for "new ____" actions generally. For anything else we have create actions for. Policies, Jobs, Roles, Tokens?

Issue #17635, opened 2023-06-21
Tuning job versions retention.


valafon:
At the moment, as far as I can see, primarily 6 versions are stored in Nomad. If you made some changes during the day, more versions are stored, but then it goes back to 6 again. I couldn't find any server documentation for tuning this behavior. I would like to increase both the number of permanently stored job versions and the frequency/parameters of retention. I found an issue on this topic way back in 2018 https://github.com/hashicorp/nomad/issues/4242. The topic creator was told that adding these changes is not difficult, but it seems like no one ever made a feature request for this change. Therefore, I will ask for it. This is very important for a convenient rollback.

Also, there is a problem, when I use nomad cli to show job versions, I see only 6 of them, while there are more at the moment.

tgross:
Hi @valafon! Yes, this is currently hard-coded (ref [`structs.go#L4240-L4246`](https://github.com/hashicorp/nomad/blob/v1.5.6/nomad/structs/structs.go#L4240-L4246)). So it'd be a matter of plumbing a new server configuration field down into where it's consumed in the state store (ref [`state_store.go#L1947-L1967`](https://github.com/hashicorp/nomad/blob/v1.5.6/nomad/state/state_store.go#L1947-L1967))

devashishTaneja:
@tgross I have raised a PR for this issue. Please have a look.

https://github.com/hashicorp/nomad/pull/17939

PR #17939, opened 2023-07-13
Tuning job versions retention. #17635

devashishTaneja:
https://github.com/hashicorp/nomad/issues/17635

### DIFF
--- a/.changelog/17939.txt
+++ b/.changelog/17939.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-server: Added option to tune job versions retention
+config: Added an option to configure how many historic versions of jobs are retained in the state store
 ```
--- a/.changelog/17939.txt
+++ b/.changelog/17939.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+server: Added option to tune job versions retention
+```
--- a/command/agent/agent.go
+++ b/command/agent/agent.go
@@ -344,8 +344,8 @@ func convertServerConfig(agentConfig *Config) (*nomad.Config, error) {
 	conf.JobDefaultPriority = jobDefaultPriority
 
 	if agentConfig.Server.JobTrackedVersions != nil {
-		if *agentConfig.Server.JobTrackedVersions < 0 {
-			return nil, fmt.Errorf("invalid Config, job_tracked_versions must be non-negative")
+		if *agentConfig.Server.JobTrackedVersions <= 0 {
+			return nil, fmt.Errorf("job_tracked_versions must be greater than 0")
 		}
 		conf.JobTrackedVersions = *agentConfig.Server.JobTrackedVersions
 	}
--- a/command/agent/agent.go
+++ b/command/agent/agent.go
@@ -343,6 +343,13 @@ func convertServerConfig(agentConfig *Config) (*nomad.Config, error) {
 	conf.JobMaxPriority = jobMaxPriority
 	conf.JobDefaultPriority = jobDefaultPriority
 
+	if agentConfig.Server.JobTrackedVersions != nil {
+		if *agentConfig.Server.JobTrackedVersions < 0 {
+			return nil, fmt.Errorf("invalid Config, job_tracked_versions must be non-negative")
+		}
+		conf.JobTrackedVersions = *agentConfig.Server.JobTrackedVersions
+	}
+
 	// Set up the bind addresses
 	rpcAddr, err := net.ResolveTCPAddr("tcp", agentConfig.normalizedAddrs.RPC)
 	if err != nil {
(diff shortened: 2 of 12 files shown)


tgross:
@devashishTaneja it's on my radar. There's no reason to keep pinging me.

.changelog/17939.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+server: Added option to tune job versions retention
```
tgross, line 2:
```suggestion
config: Added an option to configure how many historic versions of jobs are retained in the state store
```

command/agent/agent.go
```diff
@@ -343,6 +343,13 @@ func convertServerConfig(agentConfig *Config) (*nomad.Config, error) {
 	conf.JobMaxPriority = jobMaxPriority
 	conf.JobDefaultPriority = jobDefaultPriority
 
+	if agentConfig.Server.JobTrackedVersions != nil {
+		if *agentConfig.Server.JobTrackedVersions < 0 {
+			return nil, fmt.Errorf("invalid Config, job_tracked_versions must be non-negative")
```
tgross, line 348: The caller wraps this more nicely, so let's be consistent with the nearby message formats:

```suggestion
			return nil, fmt.Errorf("job_tracked_versions must be non-negative")
```

But also, is setting the value to 0 actually safe? There aren't any tests added that would exercise that case, which seems important to make sure we get right.

command/agent/config.go
```diff
@@ -1329,7 +1333,8 @@ func DefaultConfig() *Config {
 				LimitResults:  100,
 				MinTermLength: 2,
 			},
-			JobMaxSourceSize: pointer.Of("1M"),
+			JobMaxSourceSize:   pointer.Of("1M"),
+			JobTrackedVersions: pointer.Of(6),
```
tgross, line 1337: We've left `structs.JobTrackedVersions` untouched, so it would probably make more sense for this to be `pointer.Of(structs.JobTrackedVersions)`, right? But also, we should also rename `structs.JobTrackedVersions` to indicate it's the now the default (and update its docstring).

nomad/state/testing.go
```diff
@@ -16,8 +16,9 @@ import (
 
 func TestStateStore(t testing.TB) *StateStore {
 	config := &StateStoreConfig{
-		Logger: testlog.HCLogger(t),
-		Region: "global",
+		Logger:             testlog.HCLogger(t),
+		Region:             "global",
+		JobTrackedVersions: structs.JobTrackedVersions,
```
tgross, line 21: As noted above, there's no test case exercising the change in behavior (in particular around the 0 case).

tgross (changes requested):
Hi @devashishTaneja! There are quite a few tests that are failing in `nomad/state` and `nomad/fsm_test.go` that look to be associated with restoring snapshots and the job versions. So you'll want to look into those.

Also, this is a new config value so we'll need to document it. Can you add that to [the server config docs](https://github.com/hashicorp/nomad/blob/main/website/content/docs/configuration/server.mdx)?

command/agent/agent.go
devashishTaneja: @tgross setting to 0 is not safe, I have updated the condition to update this.

command/agent/config.go
devashishTaneja: Done

nomad/config.go
```diff
@@ -535,6 +538,7 @@ func DefaultConfig() *Config {
 		DeploymentQueryRateLimit: deploymentwatcher.LimitStateQueriesPerSecond,
 		JobDefaultPriority:       structs.JobDefaultPriority,
 		JobMaxPriority:           structs.JobDefaultMaxPriority,
+		JobTrackedVersions:       structs.JobTrackedVersions,
```
tgross, line 541: As I noted [here](https://github.com/hashicorp/nomad/pull/17939#discussion_r1268088124), the name `structs.JobTrackedVersions` should be changed to indicate it's a default value (along with its docstring):

```suggestion
		JobTrackedVersions:       structs.DefaultJobTrackedVersions,
```

nomad/state/testing.go
tgross: We didn't add any new tests for this behavior, but now that the 0 value is not allowed I think we're ok. I did manually test out the behavior of `auto_revert = true` with `job_tracked_versions = 1` and it seems correct and reasonable.

tgross (changes requested):
One last comment to address about the default var's name and I think this will be ready to merge @devashishTaneja!

tgross:
The failing `test-ui/finalize` is just our upload to Percy for visual diff tests failing. That's expected for external contributors, no worries. Waiting on Vercel to publish the docs preview.

tgross (approved):
LGTM! Thanks @devashishTaneja!

tgross:
This has been backported to 1.6.x and will be released with the next regular release of Nomad. Thanks!

Issue #17772, opened 2023-06-29
inelligible filter in topology doens't work


josh-m-sharpe:
![Cluster_Topology_-_Nomad](https://github.com/hashicorp/nomad/assets/39473/78dbd0d4-51cf-4909-b8fc-c97ff3a93899)

![Cluster_Topology_-_Nomad](https://github.com/hashicorp/nomad/assets/39473/df9448a9-b766-4032-9d25-3947e22d9f1c)

philrenaud:
Hi @josh-m-sharpe , thanks for raising this. We will give it a look and see what the fix looks like soon.

PR #17940, opened 2023-07-13
ui: fix Topology node state filter

lgfa29:
"Ineligible" and "Draining" are not determined by the node status, but are rather inferred from other fields.

Closes https://github.com/hashicorp/nomad/issues/17772


philrenaud (approved):
Looks great, thank you @lgfa29 !

PR #17943, opened 2023-07-14
Add missing consul grpc config

brucelok:
This Terraform folder is for quickly building an integrated Nomad/Consul cluster on the cloud. To avoid Job submission failures with Consul as the provider, it requires to include grpc port `8502` and enable the `connect` option in the configuration.


jrasell (approved):
LGTM, thanks @brucelok!

PR #17950, opened 2023-07-14
docs: note windows requirement for workload identity

shoenig:
Support for UDS sockets was added to Windows 10 (build 17063), make a note in the Task API docs. Also reference Task API from WI  pages since it's a common use case to combine the two features.

https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/


schmichael (approved):
Thanks!

PR #17963, opened 2023-07-17
updating to specify mTLS rpc endpoints

louievandyke:


website/content/docs/configuration/tls.mdx
```diff
@@ -23,6 +23,8 @@ tls {
 ~> Incorrect configuration of the TLS configuration can result in failure to
 start the Nomad agent.
 
+~> When enabling `rpc=true`, all RPC endpoints will use mutual authentication (mTLS)
+
```
tgross, line 27: We can probably just put this note inside the text of the `rpc` field below. No need for it to be a warn box.

tgross (changes requested):

tgross:
@louievandyke I'm 99% certain our employee IP agreements cover everything in the CLA but can you sign it just in case so that we don't have questions down the road?

tgross (approved):
LGTM!

PR #17964, opened 2023-07-17
[ui] Job Variables page

philrenaud:
A new route to:
- Show a user the variables that a job and its groups and tasks have automatic access to (by naming convention)
- Inform a user about how variables and automatic access work
- Give a user a quick way to create/edit variables with appropriately pathed names.

A few new conventions added to facilitate good UX here:
- We want to link to /variables/variable/edit if the variable exists, but /variables/new?path=$varpath if it does not. A new helper and component handle this for us.
- We want to show the tasks/task groups they could be adding variables for here, but not too many of them. We show up to 2 examples of each and slightly change our copy depending on number of tasks/groups.

image


ui/app/components/variable-form.hbs
```diff
@@ -72,7 +72,7 @@
           Please choose a different path, or
           <LinkTo
             @route="variables.variable.edit"
-            @model={{this.duplicatePathWarning.path}}
+            @model={{concat this.duplicatePathWarning.path "@" (or this.variableNamespace "default")}}
```
philrenaud, line 75: Side-effect: noticed this had stopped working in namespace-less environments.

ui/app/abilities/job.js
```diff
@@ -19,6 +19,14 @@ export default class Job extends AbstractAbility {
   )
   canScale;
 
+  @or(
+    'bypassAuthorization',
+    'selfTokenIsManagement',
+    'specificNamespaceSupportsReading',
+    'policiesSupportReading'
+  )
+  canRead;
```
philrenaud, line 28: This seems out of place in a PR like this, right?

When we save a variable, there are a few things that we want to update and depend on that change. One of those is `job.pathLinkedVariable`, which looks against a job's `variables` array (more on that later!) and finds the one that's named `nomad/jobs/$jobname`.

So now, in our saveVariable() function elsewhere, we can say "Oh great you saved! Do you have access to read jobs so that this might matter for them? Yes? Okay, update them."

This provides the ability to check that they can read jobs.

ui/app/components/editable-variable-link.hbs
```diff
@@ -0,0 +1,17 @@
...
+  {{#with (editable-variable-link @path existingPaths=@existingPaths namespace=@namespace) as |link|}}
+    {{#if link.model}}
+      <LinkTo @route={{link.route}} @model={{link.model}} @query={{link.query}}>{{@path}}</LinkTo>
+    {{else}}
+      <LinkTo @route={{link.route}} @query={{link.query}}>{{@path}}</LinkTo>
+    {{/if}}
+  {{/with}}
+{{else}}
+  @path
+{{/if}}
```
philrenaud, line 17: A little component that either links to variables/new?path=YourNewVar or variables/YourExistingVar/edit, depending on whether it exists or not.

All this just to save the user a click! We already inform them, when creating a new variable whose path already exists, that that variable already exists and give them a link to it. But this is nicer UX IMO.

ui/app/components/variable-form.hbs
philrenaud: PS: hey here's that "this variable path already exists" warning I mentioned a comment ago!

ui/app/helpers/editable-variable-link.js
```diff
@@ -0,0 +1,47 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
```
philrenaud, line 5: The backing logic for the `editable-variable-link.hbs` component. Could have set this up as a component script, rather than a helper, but I thought we might want to use this in a different context sometime.

ui/app/models/job.js
```diff
@@ -366,4 +366,17 @@ export default class Job extends Model {
...
+    await this.variables;
+    if (this.parent.get('id')) {
+      return await this.variables?.findBy(
+        'path',
+        `nomad/jobs/${JSON.parse(this.parent.get('id'))[0]}`
+      );
+    } else {
+      return await this.variables?.findBy('path', `nomad/jobs/${this.plainId}`);
+    }
+  }
```
philrenaud, line 381: This is, IMO, the most notable change I'm making in this PR. See the `pathLinkedVariable()` getter above this for how we've traditionally done it.

For the past few months, looking at a variable in the Web UI has been in a context where we'd have just gotten variables loaded. The operation, though, isn't immediate or promise-friendly. Frankly if this gets merged I intend to do away with `pathLinkedVariable()` in favour of `getPathLinkedVariable()` and make the whole operation more awaitable.

This lets us wait for the value of the job's (or group's, or task's, as you'll see in a moment) related variable without making any assumptions about what's already been loaded (see for example `_fetchParentJob` in the task model)

ui/app/models/task.js
```diff
@@ -79,4 +82,20 @@ export default class Task extends Fragment {
       );
     }
   }
+
+  // TODO: This async fetcher seems like a better fit for most of our use-cases than the above getter (which cannot do async/await)
+  async getPathLinkedVariable() {
+    if (!this._job) {
+      await this._fetchParentJob();
+    }
```
philrenaud, line 90: There isn't a direct link between jobs and their tasks; instead, there's an intermediary taskGroup. My take is that little helper methods like this are warranted shortcuts.

ui/app/routes/jobs/job/variables.js
```diff
@@ -0,0 +1,56 @@
...
+        allJobsVariable = null;
+      }
+    }
+
+    const variables = A([
+      allJobsVariable,
+      jobVariable,
+      ...groupVariables,
+      ...taskVariables,
+    ]).compact();
```
philrenaud, line 52: For awhile, I was doing all this logic in the controller (under `jobRelevantVariables()`), but realized that I wanted to wait for this info before loading the page, which this lets me do.

ui/app/templates/jobs/job/variables.hbs
```diff
@@ -0,0 +1,84 @@
...
+            {{#each this.firstFewTaskGroupNames as |name|}}
+              <code><EditableVariableLink @path="nomad/jobs/{{this.model.job.name}}/{{name}}" @existingPaths={{this.jobRelevantVariables.files}} @namespace={{this.model.job.namespace.name}} /></code>,
+            {{/each}}
+            etc. to grant a specific task group access
+          {{else}}
+            <code>
+              <EditableVariableLink @path="nomad/jobs/{{this.model.job.name}}/{{object-at 0 this.firstFewTaskGroupNames}}" @existingPaths={{this.jobRelevantVariables.files}} @namespace={{this.model.job.namespace.name}} />
+            </code>
+            to grant your task group access
+          {{/if}}
```
philrenaud, line 41: Lots of template logic for a little benefit, but I think it reads much nicer this way.

We give a single example if you only have a single group/task, but multiple examples (with ", etc." language) when you have multiple groups/tasks.

image
image

ui/app/templates/jobs/job/variables.hbs
```diff
@@ -0,0 +1,84 @@
...
+  </Hds::Alert>
+</header>
+
+{{#if this.jobRelevantVariables.files.length}}
+<VariablePaths
+  @branch={{this.jobRelevantVariables}}
+/>
+{{else}}
+  <p>
+    Job <strong>{{this.model.job.name}}</strong> does not currently have access to any variables.
```
philrenaud, line 68: Todo: I should make this language less severe sounding. The job may in fact have access to variables, just not automatically/by naming convention.

ui/mirage/config.js
```diff
@@ -960,6 +976,10 @@ export default function () {
   });
 
   this.get('/var/:id', function ({ variables }, { params }) {
+    let variable = variables.find(params.id);
+    if (!variable) {
+      return new Response(404, {}, {});
+    }
```
philrenaud, line 982: This emerged because I've started looking up the `nomad/jobs` variable in isolation when loading a given job/variables page now.

ui/mirage/scenarios/default.js
```diff
@@ -359,6 +359,11 @@ function mediumCluster(server) {
 function variableTestCluster(server) {
   faker.seed(1);
   createTokens(server);
+  server.create('token', {
+    name: 'Novars Murphy',
+    id: 'n0-v4r5-4cc355',
+    type: 'client',
+  });
```
philrenaud, line 366: All our recent tokens are Irish and I can no longer remember why

ui/tests/acceptance/keyboard-test.js
```diff
@@ -344,6 +347,15 @@ module('Acceptance | keyboard', function (hooks) {
...
 
+      await triggerKeyEvent('.page-layout', 'keydown', 'ArrowRight', {
+        shiftKey: true,
+      });
+      assert.equal(
+        currentURL(),
+        `/jobs/${jobID}@default/clients`,
+        'Shift+ArrowRight takes you to the next tab (Clients)'
+      );
+
```
philrenaud, line 358: Side-effect-of-a-side-effect:
- I wanted to make sure that keynav shift+right took you to the variables page, but you only see that if you have variable access
- so I said okay let's make the token a management token so that they'd definitely see it
- well the jobs created for this mock are all system jobs, which have client pages, which were otherwise unseen because before now the mocked token didn't have node read access, either, and now they do

ui/app/routes/jobs/job/variables.js
```diff
@@ -0,0 +1,56 @@
...
+    let groupVariables = await Promise.all(
+      taskGroups.map((tg) => tg.getPathLinkedVariable())
+    );
+    let taskVariables = await Promise.all(
+      tasks.map((task) => task.getPathLinkedVariable())
+    );
+
+    let allJobsVariable;
+    try {
+      allJobsVariable = await this.store.findRecord('variable', 'nomad/yobs');
```
tgross, line 40: I think?

```suggestion
      allJobsVariable = await this.store.findRecord('variable', 'nomad/jobs');
```

ui/app/models/task-group.js
```diff
@@ -38,6 +38,22 @@ export default class TaskGroup extends Fragment {
     }
   }
 
+  // TODO: This async fetcher seems like a better fit for most of our use-cases than the above getter (which cannot do async/await)
+  async getPathLinkedVariable() {
+    await this.job.variables;
+    if (this.job.parent.get('id')) {
+      return await this.job.variables?.findBy(
+        'path',
+        `nomad/jobs/${JSON.parse(this.job.parent.get('id'))[0]}/${this.name}`
```
tgross, line 47: For my education: why is there a `JSON.parse` here instead of `${this.job.plainId}` (we've done the same in `job.js`, but not in `task.js`)

ui/app/routes/jobs/job/variables.js
philrenaud: hah yeah, good catch

ui/app/templates/jobs/job/variables.hbs
```diff
@@ -0,0 +1,84 @@
...
+          {{#if (gt this.firstFewTaskNames.length 1)}}
+            {{#each this.firstFewTaskNames as |name|}}
+              <code><EditableVariableLink @path="nomad/jobs/{{this.model.job.name}}/{{name}}" @existingPaths={{this.jobRelevantVariables.files}} @namespace={{this.model.job.namespace.name}} /></code>,
+            {{/each}}
+            etc. to grant a specific task access
+          {{else}}
+            <code>
+              <EditableVariableLink @path="nomad/jobs/{{this.model.job.name}}/{{object-at 0 this.firstFewTaskNames}}" @existingPaths={{this.jobRelevantVariables.files}} @namespace={{this.model.job.namespace.name}} />
+            </code> to grant your task access
+          {{/if}}
```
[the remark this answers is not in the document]
philrenaud: Nice; digging into this. It might not be whether tasks are running or not, but whether tasks exist whatsoever; in Mirage we have this [shallow](https://github.com/hashicorp/nomad/blob/de71cc2f726802c86a0303d44d06ffab4a8b1294/ui/mirage/scenarios/default.js#L83) rule that we use to say "This is going to be a pretty heavy job, don't fill out all its tasks/groups/evals/task resources/etc.", Throughout our other mock factories we do a bit of `if (!job.shallow)` checks [before we fill them out](https://github.com/hashicorp/nomad/blob/e1e59455e4e65d5107e4c49f452c7bdb443020aa/ui/mirage/factories/task-group.js#L85-L112) . I suspect that's what's going on here, too.

ui/app/models/task-group.js
philrenaud: You know what, I don't think there's any good reason. Going to move this over to use plainId.

ui/app/models/job.js
```diff
@@ -366,4 +366,17 @@ export default class Job extends Model {
...
   }
+
+  // TODO: This async fetcher seems like a better fit for most of our use-cases than the above getter (which cannot do async/await)
+  async getPathLinkedVariable() {
+    await this.variables;
+    if (this.parent.get('id')) {
+      return await this.variables?.findBy(
+        'path',
+        `nomad/jobs/${JSON.parse(this.parent.get('id'))[0]}`
+      );
```
DingoEatingFuzz, line 377: If `this.variables` here is a RecordCollection (resolved relationship, however you want to call it), then isn't `findBy` here a synchronous array operation? If so, the `await` is doing anything.

DingoEatingFuzz: Guaranteeing that the variables are loaded without pushing that logic onto components or controllers is nice though.

ui/app/models/task-group.js
DingoEatingFuzz, line 47: Looks like you changed this to `plainId` in the `else` but not here.

ui/app/models/task.js
```diff
@@ -58,9 +58,12 @@ export default class Task extends Fragment {
   @fragmentArray('volume-mount', { defaultValue: () => [] }) volumeMounts;
 
   async _fetchParentJob() {
-    let job = await this.store.findRecord('job', this.taskGroup.job.id, {
-      reload: true,
-    });
+    let job = await this.store.peekRecord('job', this.taskGroup.job.id);
```
DingoEatingFuzz, line 61: minor: `peekRecord` is a synchronous function.

ui/app/models/task.js
```diff
@@ -58,9 +58,12 @@ export default class Task extends Fragment {
...
+    if (!job) {
```
DingoEatingFuzz, line 62: Since task is a model fragment of a task group which is a model fragment of a job, I don't believe it's possible to ever have the task in the store and not the job.

Doesn't hurt to be defensive here I suppose, but I figured it was worth mentioning.

ui/app/models/task.js
DingoEatingFuzz: Totally agree. Once upon a time Model Fragments were going to support proper relationships, but alas, here we are.

ui/app/models/task.js
```diff
@@ -79,4 +82,20 @@ export default class Task extends Fragment {
...
+  // TODO: This async fetcher seems like a better fit for most of our use-cases than the above getter (which cannot do async/await)
+  async getPathLinkedVariable() {
+    if (!this._job) {
+      await this._fetchParentJob();
+    }
+    await this._job.variables;
+    let jobID = this._job.plainId;
+    if (this._job.parent.get('plainId')) {
+      jobID = this._job.parent.get('plainId');
+    }
```
DingoEatingFuzz, line 95: Is this safe? Looking at this, it seems like this would be big mad if `_job.parent` wasn't defined (or maybe it's always defined because it's a relationship?) and it would also potentially erroneously return `undefined` because it has to load the parent job first, in the case there is a parent job that isn't resolved yet.

You can at least get the parent job's ID synchronously and without making a request with

```js
this._job.belongsTo('parent').id()
```

But I believe this would take the `['job-id','namespace']` form.

ui/app/routes/jobs/job/variables.js
```diff
@@ -0,0 +1,59 @@
...
+    );
+    let taskVariables = await Promise.all(
+      tasks.map((task) => task.getPathLinkedVariable())
+    );
+
+    let allJobsVariable;
+    try {
+      let variables = await this.store.query('variable', {
+        path: 'nomad/jobs',
+      });
```
DingoEatingFuzz, line 42: You're waterfalling api calls by using `await` like this.

Since none of the requests depend on the responses from prior requests, something like this would let the browser handle how many connections it wishes to open in parallel:

```js
    // Capture the promises themselves as variables instead of the resolutions
    let jobVariablePromise = job.getPathLinkedVariable();
    let groupVariablesPromises =  taskGroups.map((tg) => tg.getPathLinkedVariable();
    let taskVariablesPromises = tasks.map((task) => task.getPathLinkedVariable()));

    let allJobsVariable;
    try {
      let variables = await this.store.query('variable', {
        path: 'nomad/jobs',
      });
    } /* ... */
    
    // And then let them all rip at once
    const variables = await Promise.all([
      allJobsVariablesPromise,
      jobVariablePromise,
      ...groupVariablesPromises,
      ...taskVariablesPromises,
    ]);
    /* ... */
```

And if @tgross is still checking in on this PR, this is kinda gnarly. It would be groovy if the API had a way to get the whole cascade of variables for a job in one API call.

ui/mirage/config.js
DingoEatingFuzz: The HCP repo has helpers for these little common mirage things. They're quite handy; basically turns this into `return notFound()`.

ui/tests/acceptance/variables-test.js
```diff
@@ -983,4 +983,254 @@ module('Acceptance | variables', function (hooks) {
       });
     });
   });
+
+  module('Job Variables Page', function () {
```
DingoEatingFuzz, line 987: minor: I would expect these tests to be in a new `acceptance/job-variables-test.js` file to match the route hierarchy.

ui/tests/acceptance/variables-test.js
```diff
@@ -983,4 +983,254 @@ module('Acceptance | variables', function (hooks) {
...
+      assert.dom('[data-test-tab="variables"]').exists();
+      await click('[data-test-tab="variables"] a');
+      assert.equal(
+        currentURL(),
+        `/jobs/${server.db.jobs[1].id}@${server.db.jobs[1].namespace}/variables`
+      );
+      assert.dom('[data-test-no-auto-vars-message]').exists();
+      assert.dom('[data-test-create-variable-button]').doesNotExist();
+
+      window.localStorage.nomadTokenSecret = null; // Reset Token
```
DingoEatingFuzz, line 1023: minor: might be worth sticking this clean up in an `afterEach` hook

DingoEatingFuzz (approved):
This page is awesome.

Most of my comments are observations, none of them are blocking. I'd say the api call waterfall and the ED relationship stuff in the task model fragment are the most worth looking at, but still non-blocking.

ui/app/routes/jobs/job/variables.js
tgross: If we were only listing, we could use a [prefix](https://developer.hashicorp.com/nomad/api-docs/variables#prefix) on `nomad/jobs/:job_id` and get all the Variables in one API call. This is the usual "1+N queries" problem we have in the API where we only give stub objects on List. This does seems like it'd be a nice special case though.

ui/app/models/task.js
philrenaud: Hm yeah good catch, I guess parentlessJob.parent.get('x') doesn't error out by virtue of ember relationships. I'm going to switch this up regardless, so thanks for the suggestion.

philrenaud: Swapped to this:

```
  async getPathLinkedVariable() {
    if (!this._job) {
      await this._fetchParentJob();
    }
    await this._job.variables;
    let jobID = this._job.plainId;
    // not getting plainID because we dont know the resolution status of the task's job's parent yet
    let parentID = this._job.belongsTo('parent').id()
      ? JSON.parse(this._job.belongsTo('parent').id())[0]
      : null;
    if (parentID) {
      jobID = parentID;
    }
    return await this._job.variables?.findBy(
      'path',
      `nomad/jobs/${jobID}/${this.taskGroup.name}/${this.name}`
    );
  }
```

PR #17991, opened 2023-07-19
Backport of Use cpu_total_compute configuration for CPU usage too into release/1.4.x

### LINKED ISSUE #17577: Fix ARM64 CPU monitoring of jobs

hc-github-team-nomad-core:
## Backport

This PR is auto-generated from #17628 to be assessed for backporting due to the inclusion of the label backport/1.4.x.

---

Fixes #17577.

Currently, `cpu_total_compute` is only used for fingerprinting / capacity, and doesn't influence `TotalTicksAvailable`. But `TotalTicksAvailable` is used to calculate CPU usage ticks. If for example on ARM64 no CPU MHz can be detected, `TotalTicksAvailable` remains zero. And in result all CPU usage remains zero.

This PR sets out to address that problem by having `cpu_total_compute` override `TotalTicksAvailable` too.

A good way to test if this works, is by setting `cpu_total_compute` to 100. This means that all CPU metrics in the UI etc should show the percentage your CPU is loaded.
So next it is a matter of loading a job that uses 100% of a single core, and one should see 1 / CPU-cores of load. The easiest way I could find was a task that does `dd if=/dev/zero of=/dev/null`. That burns a single core for as long as you keep it running.


shoenig:
This isn't really feasible to backport to 1.4; 1.5 contained a chonky refactoring of cpu detection due to Apple Silicon support and we didn't backport those changes.

PR #17996, opened 2023-07-19
CSI: improve controller RPC reliability

tgross:
The CSI specification says that we "SHOULD" send no more than one in-flight request per *volume* at a time, with an allowance for losing state (ex. leadership transitions) which the plugins "SHOULD" handle gracefully. We mostly successfully serialize node and controller RPCs for the same volume, except when Nomad clients are lost. (See also https://github.com/container-storage-interface/spec/issues/512)

These concurrency requirements in the spec fall short because Storage Provider APIs aren't necessarily safe to call concurrently on the same host even for _different_ volumes. For example, concurrently attaching AWS EBS volumes to an EC2 instance results in a race for device names, which results in failure to attach (because the device name is taken already and the API call fails) and confused results when releasing claims. So in practice many CSI plugins rely on k8s-specific sidecars for serializing storage provider API calls globally. As a result, we have to be much more conservative about concurrency in Nomad than the spec allows.

This changeset includes four major changes to fix this:
* Add a serializer method to the CSI volume RPC handler. When the RPC handler makes a destructive CSI Controller RPC, we send the RPC thru this serializer and only one RPC is sent at a time. Any other RPCs in flight will block.
* Ensure that requests go to the same controller plugin instance whenever possible by sorting by lowest client ID out of the plugin instances.
* Ensure that requests go to _healthy_ plugin instances only.
* Ensure that requests for controllers can go to a controller on any _live_ node, not just ones eligible for scheduling (which CSI controllers don't care about)

Fixes: #15415

### DIFF
--- a/nomad/client_csi_endpoint.go
+++ b/nomad/client_csi_endpoint.go
@@ -5,7 +5,7 @@ package nomad
 
 import (
 	"fmt"
-	"math/rand"
+	"sort"
 	"strings"
 	"time"
 
@@ -262,9 +262,9 @@ func (a *ClientCSI) clientIDsForController(pluginID string) ([]string, error) {
 
 	ws := memdb.NewWatchSet()
 
-	// note: plugin IDs are not scoped to region/DC but volumes are.
-	// so any node we get for a controller is already in the same
-	// region/DC for the volume.
+	// note: plugin IDs are not scoped to region but volumes are. so any Nomad
+	// client we get for a controller is already in the same region for the
+	// volume.
 	plugin, err := snap.CSIPluginByID(ws, pluginID)
 	if err != nil {
 		return nil, fmt.Errorf("error getting plugin: %s, %v", pluginID, err)
@@ -273,13 +273,10 @@ func (a *ClientCSI) clientIDsForController(pluginID string) ([]string, error) {
 		return nil, fmt.Errorf("plugin missing: %s", pluginID)
 	}
 
-	// iterating maps is "random" but unspecified and isn't particularly
-	// random with small maps, so not well-suited for load balancing.
-	// so we shuffle the keys and iterate over them.
 	clientIDs := []string{}
 
 	for clientID, controller := range plugin.Controllers {
-		if !controller.IsController() {
+		if !controller.IsController() || !controller.Healthy {
 			// we don't have separate types for CSIInfo depending on
 			// whether it's a controller or node. this error shouldn't
 			// make it to production but is to aid developers during
@@ -295,9 +292,11 @@ func (a *ClientCSI) clientIDsForController(pluginID string) ([]string, error) {
 		return nil, fmt.Errorf("failed to find clients running controller plugin %q", pluginID)
 	}
 
-	rand.Shuffle(len(clientIDs), func(i, j int) {
-		clientIDs[i], clientIDs[j] = clientIDs[j], clientIDs[i]
-	})
+	// Many plugins don't handle concurrent requests as described in the spec,
+	// and have undocumented expectations of using k8s-specific sidecars to
+	// leader elect. Sort the client IDs so that we prefer sending requests to
+	// the same controller to hack around this.
+	clientIDs = sort.StringSlice(clientIDs)
 
 	return clientIDs, nil
 }
--- a/nomad/client_csi_endpoint.go
+++ b/nomad/client_csi_endpoint.go
@@ -4,6 +4,7 @@
 package nomad
 
 import (
+	"errors"
 	"fmt"
 	"sort"
 	"strings"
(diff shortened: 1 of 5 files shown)


nomad/csi_endpoint_test.go
```diff
@@ -1971,3 +1973,49 @@ func TestCSI_RPCVolumeAndPluginLookup(t *testing.T) {
 	require.Nil(t, vol)
 	require.EqualError(t, err, fmt.Sprintf("volume not found: %s", id2))
 }
+
+func TestCSI_SerializedControllerRPC(t *testing.T) {
```
tgross, line 1977: I've done some end-to-end testing of this new function with reschedules but I want to follow-up with some additional testing of drains, broken plugin configurations, etc., to make sure that everything is recoverable even in the corner cases.

nomad/client_csi_endpoint.go
```diff
@@ -295,9 +292,11 @@ func (a *ClientCSI) clientIDsForController(pluginID string) ([]string, error) {
 		return nil, fmt.Errorf("failed to find clients running controller plugin %q", pluginID)
```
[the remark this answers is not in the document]
gulducat: :heavy_plus_sign: it could be especially nice to know if we do find controller plugin(s) that meet the added `!controller.Healthy` condition.

nomad/csi_endpoint.go
```diff
@@ -586,6 +589,57 @@ func (v *CSIVolume) volAndPluginLookup(namespace, volID string) (*structs.CSIPlu
 	return plug, vol, nil
 }
 
+// serializedControllerRPC ensures we're only sending a single controller RPC to
+// a given plugin if the RPC can cause conflicting state changes.
```
gulducat, line 593: I think we're good, but sharing here some of my thought process: For this to work, all such RPCs need to be run on a single server to block in memory.  Looking at what I think are the relevant RPCs, it seems like they are all(?) `v.srv.forward()`ed (to the leader).  I suppose this is what you meant by "with an allowance for losing state (ex. leadership transitions)" - i.e. this serialization could be broken by switching leaders?  and we can't reasonably avoid that caveat.

nomad/csi_endpoint.go
```diff
@@ -586,6 +589,57 @@ func (v *CSIVolume) volAndPluginLookup(namespace, volID string) (*structs.CSIPlu
...
+			v.srv.volumeControllerFutures[pluginID] = future
+			v.srv.volumeControllerLock.Unlock()
+
+			err := fn()
+
+			// close the future while holding the lock and not in a defer so
+			// that we can ensure we've cleared it from the map before allowing
+			// anyone else to take the lock and write a new one
+			v.srv.volumeControllerLock.Lock()
+			futureDone()
```
gulducat, line 619: I find myself wanting simpler map thread-safety, and a [`sync.Map`](https://pkg.go.dev/sync#Map) seems just about right instead of a manual mutex, but its auto-locking-for-you behavior might not be compatible with your concern here?

nomad/client_csi_endpoint.go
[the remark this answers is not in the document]
tgross: Done. Doing so revealed another bug which I've fixed, which is that we're requiring controller's nodes to be eligible, which we don't need to do. That will likely help resolve some of the problems in https://github.com/hashicorp/nomad/issues/17756. I'm kind of wondering if we should split everything in this file out to its own PR at this point.

nomad/csi_endpoint.go
```diff
@@ -586,6 +589,57 @@ func (v *CSIVolume) volAndPluginLookup(namespace, volID string) (*structs.CSIPlu
...
+			case <-future.Done():
+				continue
+			case <-v.srv.shutdownCh:
+				// The csi_hook publish workflow on the client will retry if it
+				// gets this error. On unpublish, we don't want to block client
+				// shutdown so we give up on error. The new leader's
+				// volumewatcher will iterate all the claims at startup to
+				// detect this and mop up any claims in the NodeDetached state
+				// (volume GC will run periodically as well)
+				return structs.ErrNoLeader
```
gulducat, line 637: Worth wrapping this with any more info about where it came from, or does that happen upstream of here?

nomad/csi_endpoint_test.go
```diff
@@ -1971,3 +1973,49 @@ func TestCSI_RPCVolumeAndPluginLookup(t *testing.T) {
...
+	totals := map[string]time.Duration{}
+	for i := 0; i < 3; i++ {
+		pair := <-timeCh
+		totals[pair.First] += pair.Second
+	}
+
+	wg.Wait()
+
+	// plugin1 RPCs should block each other
+	must.GreaterEq(t, 150*time.Millisecond, totals["plugin1"])
```
gulducat, line 2015: I see the test passes, but how does `2 * 50` milliseconds add up to `> 150` ?  "plugin1" should only be blocked behind itself, so "plugin2"'s 50ms shouldn't be included, right?

gulducat (approved):
Looks good!  I think it'll cover a lot of our bases, and should also make it easier to debug any future issues.

Mainly though I'm not clear on some math in the test (which makes me feel a little not-smart, since it seems like basic multiplication?), so either something to fix there or I have a facepalm in my future.

nomad/csi_endpoint.go
tgross: Right, that's exactly it.

nomad/csi_endpoint.go
tgross: Yeah we're locking around both access to the map and the construction/cancellation of contexts. With a `sync.Map` that would end up looking like this:

```go
// spin up a context goroutine whether we need it or not!
future, futureDone := context.WithCancel(v.srv.shutdownCtx)
defer futureDone()

for {
	ok := v.srv.volumeControllerFutures.CompareAndSwap(pluginID, nil, future)
	if ok {
		// success path
		err := fn()
		futureDone()
		v.srv.volumeControllerFutures.CompareAndDelete(pluginID, future)
		return err
	} else {
		// wait path

		// racy! what if it's been deleted?
		waitOn, ok := v.srv.volumeControllerFutures.Load(pluginID) 
		if !ok {
			continue // lost the race with the delete, start again
		}
		select {
			case <-waitOn.Done():
				continue // other goroutine is done
			case <-v.srv.shutdownCh:
				return structs.ErrNoLeader
		}
	}
}
```

There are a few reasons why this isn't the best solution:
* A `sync.Map` doesn't let us compare-and-swap unless we've already constructed the object we're intending to write. So we end up awkwardly constructing the context outside the loop.
* In the "success path" we need to cancel the context before deleting it from the map, otherwise someone else can start their future before we've closed our context (and any downstream contexts we add in the future). But doing so will unblock the "wait path" before the `CompareAndSwap` will succeed, potentially leading to that goroutine trying to grab the `CompareAndSwap` multiple times before it can succeed, which is wasteful. 
* `CompareAndSwap` doesn't return the existing value, so we need to `Load` it the value and it may have been deleted in the meanwhile. This adds more contention around that lock.
* The `CompareAndSwap` operation returns `false` when it hasn't been previously set! We could probably hack around that by creating a dummy context that's always treated as "unused" but then we've spun up a goroutine per plugin just to hold that spot.

nomad/csi_endpoint_test.go
tgross: For `plugin1`: the goroutine on line 2002 runs for 50ms. The goroutine on line 2004 waits at least 50ms for the other to finish and then runs for 50ms, for a total of 100ms. We're summing up the elapsed time of all the goroutines for a given plugin, because that way we're asserting the wait time of the method under test from the perspective of each goroutine and not just the elapsed wall clock time. So it's a minimum of 50ms + 100ms.

The maximum value of 200ms is assuming that any overhead is (far) less than 50ms, so that asserts that we're not accidentally waiting behind `plugin2` as well.

nomad/csi_endpoint.go
tgross: The caller's logger will wrap it sufficiently. "No leader" is fairly foundational.

tgross:
In addition to the tests added above, I've done some end-to-end testing here:
* client restart on a node that has a job that mounts a volume and a controller + node plugin
* rescheduling a job that mounts a volume
* draining a node with a job that mounts a volume and a controller plugin on it

nomad/csi_endpoint.go
gulducat: Nice, thanks for the mock-up and detailed explanation!

Issue #17971, opened 2023-07-18
Post Stop tasks are never killed on the client node (also at GC time) resulting in task events loss.


stswidwinski:
### Nomad version
Tip.

### Operating system and Environment details
Linux.

### Issue

When the client Nomad node garbage collects an allocation, any post-stop lifecycle tasks:

1. Are not killed,
2. Lose task state events.

While one may argue that #1 is intended, #2 is most likely not and seems to be a bug.

### Reproduction steps

First, start up a single server and a single client. It is important that the client has a low `max_alloc_gc` so that we can easily evoke the garbage collection.

```
# Start up a server
$ echo 'data_dir = "/tmp/server"                                                                                                                                                                                                                                                                     
log_level = "TRACE"                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                      
advertise {                                                                                                                                                                                                                                                                                           
  http = "127.0.0.1"                                                                                                                                                                                                                                                                                  
  rpc = "127.0.0.1"                                                                                                                                                                                                                                                                                   
  serf = "127.0.0.1"                                                                                                                                                                                                                                                                                  
}                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                      
server {                                                                                                                                                                                                                                                                                              
  enabled = true                                                                                                                                                                                                                                                                                      
  bootstrap_expect = 1                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
}' > /tmp/server

$ nomad agent -config /tmp/server
 
# Start up a client
echo 'data_dir = "/tmp/client-1"                                                                                                                                                                                                                                                                   
log_level = "debug"                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                      
advertise {                                                                                                                                                                                                                                                                                           
  http = "127.0.0.1"                                                                                                                                                                                                                                                                                  
  rpc = "127.0.0.1"                                                                                                                                                                                                                                                                                   
  serf = "127.0.0.1"                                                                                                                                                                                                                                                                                  
}                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                      
ports {                                                                                                                                                                                                                                                                                               
  http = "9876"                                                                                                                                                                                                                                                                                       
  rpc = "9875"                                                                                                                                                                                                                                                                                        
  serf = "9874"                                                                                                                                                                                                                                                                                       
[...]

PR #18005, opened 2023-07-19
Retain task states for post stop tasks at the time of node GC

stswidwinski:
Persist the task states for post stop tasks at the time of killing to resolve https://github.com/hashicorp/nomad/issues/17971.

Please note that this is the safer of the two options discussed within the original issue description.

### DIFF
--- a/.changelog/17971.txt
+++ b/.changelog/17971.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+client: Fixed a bug where the state of poststop tasks could be corrupted by client gc
+```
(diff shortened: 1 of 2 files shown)


stswidwinski:
I believe that `test-ui / finalize (pull_request) ` has failed for unrelated reasons. I would appreciate it if the reviewer could double check

schmichael (approved):

schmichael:
Tested and looks good to me! Thanks for the great work debugging and fixing this. I pushed up a changelog entry, added backport labels, and will merge as soon as tests are happy.

PR #18020, opened 2023-07-20
ui: handle errors from unimplemented services

lgfa29:
When a request is made to an RPC service that doesn't exist (for example, a cross-region request from a newer version of Nomad to an older version that doesn't implement the endpoint) the application should return an empty list as well.

### DIFF
--- a/ui/app/adapters/application.js
+++ b/ui/app/adapters/application.js
@@ -44,7 +44,9 @@ export default class ApplicationAdapter extends RESTAdapter {
     return super.findAll(...arguments).catch((error) => {
       const errorCodes = codesForError(error);
 
-      const isNotImplemented = errorCodes.includes('501');
+      const isNotImplemented =
+        errorCodes.includes('501') ||
+        error.message.includes("rpc: can't find service");
 
       if (isNotImplemented) {
         return [];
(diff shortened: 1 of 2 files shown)


ui/app/adapters/application.js
```diff
@@ -44,7 +44,9 @@ export default class ApplicationAdapter extends RESTAdapter {
     return super.findAll(...arguments).catch((error) => {
       const errorCodes = codesForError(error);
 
-      const isNotImplemented = errorCodes.includes('501');
+      const isNotImplemented =
+        errorCodes.includes('501') ||
+        error.message.includes("rpc: can't find service");
```
tgross, line 49: Note for reviewers: this error comes from the stdlib: https://cs.opensource.google/go/go/+/refs/tags/go1.20.6:src/net/rpc/server.go;l=613

philrenaud: There's a change a UI dev* might short-circuit this with an isNotImplemented handler in the future that just checks for `501` in a specific adapter layer. I don't think it's likely (we've always used `super` so far to inherit from this one) but it would be safer if we eventually returned 501s for these. I don't know if changing the stdlib is at all a likelihood, though!

*it could be me! I forget stuff all the time!

philrenaud (approved):

ui/app/adapters/application.js
lgfa29: We would need to wrap the RPC error to change the status code. Not a big problem, but I suspect there may be lots of code to change 😅 

I opened https://github.com/hashicorp/nomad/issues/18050 to track this.

PR #18051, opened 2023-07-24
[ui] Sort variables ascending by key at serialization time

philrenaud:
Doesn't impact the variable view page, as it's already using a table that sorts ascending on key by default, but on the edit page, which iterates over and creates a form label for each, they'll now appear ascending on key alpha.


gulducat (approved):
consistency is `.key`

Issue #14872, opened 2022-10-11
Sorting problem with text filter


wdmeest:
### Nomad version
1.3.1

### Operating system and Environment details

Ubuntu 20.04

### Issue

In nomad dashboard the sorting of the jobs table doesn't work when using a text filter (without text filter it works perfectly).

Example (sort on namespace is not correct):

![image](https://user-images.githubusercontent.com/23402402/195071216-f2bd461c-937d-4fcb-8a88-b4c9bb3a1fd2.png)

The sorting problem is applicable for all columns so not only the namespace column

#### Expected Result
Correct sorting

jrasell:
Hi @wdmeeste1 and thanks for raising this issue. I was able to reproduce this locally with v.1.3.1 and v.1.4.1; I will therefore pass this onto the UI engineers so they can take a look when possible.

PR #18053, opened 2023-07-24
[ui] Search results are overloading filter with sorted results

philrenaud:
Bit of context here:
- Our Searchable mixin uses fuzzy-finding, so both my jobs "fakepy" and "fails_after_10" will show up when I type "fai" or "fak" into my search bar
image
- Last year, we addressed a [bug where sorting wouldn't apply at all if a search term was present](https://github.com/hashicorp/nomad/issues/14872) in PR #15432 
- More recently, we'd heard concerns that our search results appeared degraded from previous versions, see these screenshots showing 1.5.6 and 1.4.10 respectively: 
image
image

This PR attempts to both preserve the ability to sort after searching, and not overload results with searched-but-not-ordered filter results by creating a new getter() that checks to see if any sorting has been applied.

### DIFF
--- a/.changelog/18053.txt
+++ b/.changelog/18053.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: search results are no longer overridden by sorting preferences on the jobs index page
+```
--- a/ui/app/controllers/jobs/index.js
+++ b/ui/app/controllers/jobs/index.js
@@ -19,7 +19,6 @@ import {
   deserializedQueryParam as selection,
 } from 'nomad-ui/utils/qp-serialize';
 import classic from 'ember-classic-decorator';
-import { tracked } from '@glimmer/tracking';
 
 const DEFAULT_SORT_PROPERTY = 'modifyIndex';
 const DEFAULT_SORT_DESCENDING = true;
@@ -72,7 +71,7 @@ export default class IndexController extends Controller.extend(
   @readOnly('userSettings.pageSize') pageSize;
 
   sortProperty = DEFAULT_SORT_PROPERTY;
-  sortDescending = true;
+  sortDescending = DEFAULT_SORT_DESCENDING;
 
   @computed
   get searchProps() {
@@ -298,6 +297,7 @@ export default class IndexController extends Controller.extend(
     });
   }
 
+  // eslint-disable-next-line ember/require-computed-property-dependencies
   @computed('searchTerm')
   get sortAtLastSearch() {
     return {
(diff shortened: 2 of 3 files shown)


lgfa29 (approved):
Code changes LGTM, just missing a changelog entry and that missing property.

But I think there may be a challenge in terms of being able to get the list sorted by the original search score, since when we sort results there's no way to "unsort"  them.

One approach I though was for a search to always reset the sort order to the fuzzy score, but that could be annoying? But also, that's kind of what I expect to happen? 😅

philrenaud:
@lgfa29 

> always reset the sort order to the fuzzy score

My worry here is that:
- search `fai`, both "fails" and "fakepy" show up, in that order
- click to sort by `priority` or something to change order
- nothing changes because fuzzy score isn't affected

I think that would break the expected experience from https://github.com/hashicorp/nomad/issues/14872, so I'm apprehensive to apply it.

lgfa29:
Ah sorry, I think I didn't explain myself well. From your example the flow I was thinking would be like this:

1. search `fai`, `fails`, `fails-2`, and `fakepy` show up, in that order
2. click to sort by `priority` or something to change order
3. table is sorted according to selection
4. search for `fail`, only `fails` and `fails-2` show up, ordered by their fuzzy score and previous order selection is removed.

So after a search users can sort results by whichever column they would like, but changing the search term resets the ordering to fuzzy score.

lgfa29:
Hum...I think what's missing is a listener for changes to the search term that resets `sortProperty` and `sortDescending` to their default values. Then I believe `unsortedSearchState` (or `prioritizeSearchOrder`, it does sound like a better name) and `sortedJobs` would do the flow I described.

lgfa29:
I think we still need to clear the `sort` query param to make the arrow disappear from the table header:
image

philrenaud:
> I think we still need to clear the sort query param to make the arrow disappear from the table header:
Only thing I'm worried about there is upon removing the search term, I expect the sort to be maintained. I could hide the arrow while the search is the most recent thing, butI'd need to re-establish it once the search term is removed.

lgfa29:
> Only thing I'm worried about there is upon removing the search term, I expect the sort to be maintained.

Wouldn't that be a new search? I think I would expect the table to reset to its original state (sorted by `CreateIndex`) 🤔

philrenaud:
Yeah fair enough! That's actually the state I moved it to this morning, so I think I'll move forward with this.

PR #18054, opened 2023-07-25
feature: Add new field render_templates on restart block

nvanthao:
This feature is necessary when user want to explicitly re-render all templates on task restart. E.g. to fetch all new secrets from Vault, even if the lease on the existing secrets has not been expired.


schmichael:
> My apologies @schmichael, next PRs will be better! I have updated the tests.

You've clearly never looked at how many tries it takes me on my PRs then. :sweat_smile:  No worries. This is looking great, and jobspec additions have an obnoxious amount of plumbing to track down. :+1:

schmichael (approved):
Well that one set of tests just refuses to finish in time, but things look good to me. Thanks for sticking with this!

It will get released in the next 1.6.x release, but we don't generally backport new features.

nvanthao:
Many thanks for reviewing @schmichael!

PR #18069, opened 2023-07-26
build: support s390x architecture for linux (ent)

### LINKED ISSUE #18068: build: experimental support for s390x executables (enterprise)
This issue is here for tracking s390x builds enterprise feature. 

Enterprise ticket: https://github.com/hashicorp/nomad-enterprise/issues/1107

pkazmierczak:
Makefile changes required for supporting s390x builds and a corresponding changelog entry. 

Enterprise PR: https://github.com/hashicorp/nomad-enterprise/pull/1202

Resolves #18068

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -53,6 +53,7 @@ ALL_TARGETS = linux_386 \
 	linux_amd64 \
 	linux_arm \
 	linux_arm64 \
+	linux_s390x \
 	windows_386 \
 	windows_amd64
 endif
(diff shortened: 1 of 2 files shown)


GNUmakefile
```diff
@@ -53,6 +53,7 @@ ALL_TARGETS = linux_386 \
 	linux_amd64 \
 	linux_arm \
 	linux_arm64 \
+	linux_s390x \
```
shoenig, line 56: wouldn't adding this line cause s390x builds to happen in the oss pipeline?

pkazmierczak: no, this only lists `linux_s390x` as one of the potential `make pkg/` targets. You can build an s390x binary manually this way, but the oss build pipeline depends on the build workflow. (attached a screenshot of a running build pipeline for this branch)
Screenshot 2023-07-26 at 16 52 03

Issue #7056, opened 2020-02-03
Allow nomad UI in iframes


michalgm:
For reporting security vulnerabilities [please refer to the website.](https://www.nomadproject.io/security.html)

If you have a question, prepend your issue with `[question]` or preferably use the [nomad mailing list](https://www.nomadproject.io/community.html).

If filing a bug please include the following:

### Nomad version
v0.9.7+ent

### Issue
We have built a internal web app that provides a streamlined view into the various components of our cloud architecture. The app primarily acts as glue around a series of iframes to app-specific UIs. 
We have been using the 3rd-party hashi-ui for our view into nomad jobs, but would like to switch to your much-improved native nomad ui (great work, btw!). However, currently we are unable to do this, as the ui returns a restrictive content-security-policy header that disallows it being used in an iframe: 
`content-security-policy: default-src 'none'; connect-src 'self'; img-src 'self' data:; script-src 'self'; style-src 'unsafe-inline' 'self'; form-action 'none'; frame-ancestors 'none'`

Is there any possibility the `frame-ancestors 'none'` could be removed from the policy? Or the policy could be made configurable?
Relevant code line: https://github.com/hashicorp/nomad/blob/dd8fbf476e2ecc6cb756c7f9db1c242fcbcbd61f/command/agent/http.go#L367

### Reproduction steps
Load nomad-ui in an iframe. It will fail to load.

DingoEatingFuzz:
Hi @michalgm! Thank you for the issue (and the nice words!) This is definitely a use case we'd like to support, but we don't want to drop `frame-ancestors 'none'` from the currently hard-coded CSPs. Removing it means opening up room for clickjacking and phishing attacks.

We definitely want to make it configurable though! Vault [currently has an API for configuring UI headers](https://www.vaultproject.io/api-docs/system/config-ui/) that we might emulate. I think some sort of server hcl config may also be a nice path to pursue.

ebarriosjr:
Has there been any progress on this? I can't find any option in the nomad api that works like the vault one.

tgross:
A configurable CSP has been merged in #18085 and will ship in the next regular release of Nomad.

PR #18085, opened 2023-07-27
Added config flag to disable frame-ancestor for the nomad UI

ebarriosjr:
In order to be able to embed the nomad UI in an iframe as requested under #7056 I made a test implementation to add a configuration variable that disables `frame-ancestors 'none' ` 
The configuration option goes under the UI config and it is a boolen flag called: `DisableFrameAncestors`

### DIFF
--- a/command/agent/http.go
+++ b/command/agent/http.go
@@ -649,7 +649,13 @@ func (e *codedError) Code() int {
 func (s *HTTPServer) handleUI(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		header := w.Header()
-		header.Add("Content-Security-Policy", "default-src 'none'; connect-src *; img-src 'self' data:; script-src 'self'; style-src 'self' 'unsafe-inline'; form-action 'none'; frame-ancestors 'none'")
+		agentConfig := s.agent.GetConfig()
+		uiDisableFrameAncestorsEnabled := agentConfig.UI != nil && agentConfig.UI.DisableFrameAncestors
+		if uiDisableFrameAncestorsEnabled {
+			header.Add("Content-Security-Policy", "default-src 'none'; connect-src *; img-src 'self' data:; script-src 'self'; style-src 'self' 'unsafe-inline'; form-action 'none'")
+		} else {
+			header.Add("Content-Security-Policy", "default-src 'none'; connect-src *; img-src 'self' data:; script-src 'self'; style-src 'self' 'unsafe-inline'; form-action 'none'; frame-ancestors 'none'")
+		}
 		h.ServeHTTP(w, req)
 	})
 }
--- a/nomad/structs/config/ui.go
+++ b/nomad/structs/config/ui.go
@@ -11,6 +11,9 @@ type UIConfig struct {
 	// Enabled is used to enable the web UI
 	Enabled bool `hcl:"enabled"`
 
+	// Disable frame-ancestors CSP header
+	DisableFrameAncestors bool `hcl:"disable_frame_ancestors"`
+
 	// Consul configures deep links for Consul UI
 	Consul *ConsulUIConfig `hcl:"consul"`
 


nomad/structs/config/ui.go
```diff
@@ -11,6 +11,9 @@ type UIConfig struct {
 	// Enabled is used to enable the web UI
 	Enabled bool `hcl:"enabled"`
 
+	// Disable frame-ancestors CSP header
+	DisableFrameAncestors bool `hcl:"disable_frame_ancestors"`
```
tgross, line 15: We'll also need to add this field to the `(*UIConfig) Merge` method below so that configuration with multiple files works correctly.

tgross (changes requested):
Hi @ebarriosjr! Thanks for the PR!

This configuration value lets us drop the specific item you're looking for in #7056 but seems like the sort of thing we'll end up needing to extend multiple times. I'm sort of thinking it might be a good idea to allow arbitrary configuration here but that might actually be exposing users who don't know much about CSP to higher risk.

So I'd love to get some input from @picatz and @philrenaud on this. Are the other values we set for CSP even sensible for a user to want to override?

Once we've settled out any design discussion, can you:
* Run `make cl` to add a changelog entry.
* Add an appropriate documentation section to [`ui.mdx`](https://github.com/hashicorp/nomad/blob/main/website/content/docs/configuration/ui.mdx)
* Can you outline a little bit how you tested that this does the job for allowing iframes?

DingoEatingFuzz:
> I'm sort of thinking it might be a good idea to allow arbitrary configuration here

I tend to agree. I don't think it hurts to open it up even if most users aren't going to need to override any headers or add any headers.

> but that might actually be exposing users who don't know much about CSP to higher risk.

The way vault implements this is you literally [override the entire `Content-Security-Policy` HTTP header](https://developer.hashicorp.com/vault/api-docs/system/config-ui) (or any other HTTP header for that matter). This approach would definitely expose users to risk since setting a policy you care about could result in accidentally unsetting a policy you didn't know you care about.

If it's good enough for Vault maybe it's good enough for Nomad? Or maybe it's only good enough for Vault because they expect operators to be well versed in all things security related?

Either way, I like an individual policy approach where the overrides are merged with the defaults. This would also change the proposed config flag from

```
DisableFrameAncestors = true
```

To, idk, something like

```
ContentSecurityPolicy {
  FrameAncestors = "*"
}
```

tgross:
> To, idk, something like
```
ContentSecurityPolicy {
  FrameAncestors = "*"
}
```

Ooo, I would not hate that. Especially if we made it so that leaving values unset left the defaults. We'd end up with something like the following, and then there'd be a method on `ContentSecurityPolicy` that turns it into the appropriate header string with quotes as needed.

```go
// only covers the elements of
// https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP we need or care about
type ContentSecurityPolicy struct {
	ConnectSrc     []string
	DefaultSrc     []string
	FormAction     []string
	FrameAncestors []string
	ImgSrc         []string
	ScriptSrc      []string
	StyleSrc       []string
}

var defaultCSP = ContentSecurityPolicy{
	ConnectSrc:     []string{"*"},
	DefaultSrc:     []string{"none"},
	FormAction:     []string{"none"},
	FrameAncestors: []string{"none"},
	ImgSrc:         []string{"self", "data:"},
	ScriptSrc:      []string{"self"},
	StyleSrc:       []string{"self", "unsafe-inline"},
}
```

nomad/structs/config/ui.go
```diff
@@ -24,6 +24,30 @@ type UIConfig struct {
...
+
+func DefaultCSPConfig() *ContentSecurityPolicy {
+	return &ContentSecurityPolicy{
+		ConnectSrc:     []string{"*"},
+		DefaultSrc:     []string{"none"},
+		FormAction:     []string{"none"},
+		FrameAncestors: []string{"none"},
+		ImgSrc:         []string{"self", "data:"},
+		ScriptSrc:      []string{"self"},
+		StyleSrc:       []string{"self", "unsafe-inline"},
```
tgross, line 47: I mentioned in my previous comment that we'd need to figure out the quoting for these when we render the header value. The simply `fmt` interpolation you did is much easier, so that's fine, but now we have to make sure the default values are quoted properly. Looking at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/Sources#sources I think a bunch of these need to be single-quoted inside the string, because they're special values.

```suggestion
		DefaultSrc:     []string{"'none'"},
		FormAction:     []string{"'none'"},
		FrameAncestors: []string{"'none'"},
		ImgSrc:         []string{"'self'", "data:"},
		ScriptSrc:      []string{"'self'"},
		StyleSrc:       []string{"'self'", "'unsafe-inline'"},
```

helper/funcs.go
```diff
@@ -509,3 +509,10 @@ func Merge[T comparable](a, b T) T {
 	}
 	return a
 }
+
+func MergeListWithDefaults(list []string, defaults []string) []string {
```
tgross, line 513: There's a `Merge` method on all the structs in `nomad/structs/config/ui.go`. We should move this logic and most of the logic you have in `command/agent/http.go` into a `Merge` method on `ContentSecurityPolicy` that gets called by `UIConfig.Merge`.

command/agent/http.go
```diff
@@ -650,12 +652,18 @@ func (s *HTTPServer) handleUI(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		header := w.Header()
 		agentConfig := s.agent.GetConfig()
```
tgross, line 654: Note that you've dropped the nil-check of `agentConfig.UI` here, so the new code will cause the agent to panic if the `ui` block is unset. Likewise for the `ContentSecurityPolicy` field.

You'll need a check for both nils, and at that point you might want to consider assigning to a variable like `policy := agentConfig.UI.ContentSecurityPolicy` just so you don't have to write out the chain of dereferences a half-dozen times. But see also my comment below about where this merge should live.

tgross (changes requested):
I think we're getting closer, @ebarriosjr. I've left some comments, but can you also run `make cl` to add a changelog entry for this PR? We'll also need the new configuration block to be documented in https://github.com/hashicorp/nomad/blob/main/website/content/docs/configuration/ui.mdx

ebarriosjr:
@tgross thanks for all the comments, I made another commit before reading your comments. I will keep working on the comments and made another commit. :)

command/agent/http.go
ebarriosjr: The handleUI function is called after checking if the agentConfig.UI is not nil on line 509. That is why i did not recheck this.

ebarriosjr: The ContentSecurityPolicy will never be nil because I added defaults values on the nomad/structs/config/ui.go file. IMHO

website/content/docs/configuration/ui.mdx
```diff
@@ -48,6 +58,30 @@ and the configuration is individual to each agent.
...
+- `img_src` `(array<string>: [])` - Specifies the valid sources
+  for `img-src` in the Content Security Policy header. The default
+  value is `["'self'","data:"]`.
+- `script_src` `(array<string>: [])` - Specifies the valid sources for
+  `script-src` in the Content Security Policy header. The default
+  value is `["'self'"]`.
+- `style_src` `(array<string>: [])` - Specifies the
+  valid sources for `style-src` in the Content Security Policy
+  header. The default value is `["'self'","'unsafe-inline'"]`.
+
```
tgross, line 84: It'd be nice if we can give a little more context here. The defaults for `none` should read `'none'`. Also, the trailing `[]` in the type description is the default value, so we can tighten the individual field docs up like the following:

```suggestion
## `content_security_policy` Parameters

The `content_security_policy` block configures the HTTP
[Content-Security-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP)
header sent with the web UI response.

- `connect_src` `(array<string>: ["*"])` - Specifies the valid sources for
  `connect-src` in the Content Security Policy header.
- `default_src` `(array<string>: ["'none'"])` - Specifies the valid sources for
  `default-src` in the Content Security Policy header.
- `form_action` `(array<string>: ["'none'"])` - Specifies the valid sources for
  `form-action` in the Content Security Policy header.
- `frame_ancestors` `(array<string>: ["'none'"])` - Specifies the valid sources
  for `frame-ancestors` in the Content Security Policy header.
- `img_src` `(array<string>: ["'self'", "data:"])` - Specifies the valid sources
  for `img-src` in the Content Security Policy header.
- `script_src` `(array<string>: ["'self'"])` - Specifies the valid sources for
  `script-src` in the Content Security Policy header.
- `style_src` `(array<string>: ["'self'","'unsafe-inline'"])` - Specifies the
  valid sources for `style-src` in the Content Security Policy header.

```

website/content/docs/configuration/ui.mdx
```diff
@@ -16,6 +16,16 @@ The `ui` block configures the Nomad agent's [web UI].
 ui {
   enabled =  true
 
+  content_security_policy {
+    connect_src     = ["*"]
+    default_src     = ["none"]
+    form_action     = ["none"]
+    frame_ancestors = ["none"]
```
tgross, line 23:
```suggestion
    default_src     = ["'none'"]
    form_action     = ["'none'"]
    frame_ancestors = ["'none'"]
```

nomad/structs/config/ui.go
```diff
@@ -21,6 +29,80 @@ type UIConfig struct {
...
+}
+
+// Copy returns a copy of this Vault UI config.
+func (old *ContentSecurityPolicy) Copy() *ContentSecurityPolicy {
+	if old == nil {
+		return nil
+	}
+
+	nc := new(ContentSecurityPolicy)
+	*nc = *old
```
tgross, line 51: In Go this only copies primitives and won't copy `[]string`. So you'll need to do this:

```suggestion
	*nc = *old
	nc.ConnectSrc = slices.Clone(old.ConnectSrc)
	nc.DefaultSrc = slices.Clone(old.DefaultSrc)
	nc.FormAction = slices.Clone(old.FormAction)
	nc.FrameAncestors = slices.Clone(old.FrameAncestors)
	nc.ImgSrc = slices.Clone(old.ImgSrc)
	nc.ScriptSrc = slices.Clone(old.ScriptSrc)
	nc.StyleSrc = slices.Clone(old.StyleSrc)	
```

nomad/structs/config/ui.go
```diff
@@ -21,6 +29,80 @@ type UIConfig struct {
...
+}
+
+func (csp *ContentSecurityPolicy) String() string {
+	return fmt.Sprintf("default-src %s; connect-src %s; img-src %s; script-src %s; style-src %s; form-action %s; frame-ancestors %s", strings.Join(csp.DefaultSrc, " "), strings.Join(csp.ConnectSrc, " "), strings.Join(csp.ImgSrc, " "), strings.Join(csp.ScriptSrc, " "), strings.Join(csp.StyleSrc, " "), strings.Join(csp.FormAction, " "), strings.Join(csp.FrameAncestors, " "))
+}
+
+func (csp *ContentSecurityPolicy) Merge(other *ContentSecurityPolicy) *ContentSecurityPolicy {
+	result := csp.Copy()
+	if result == nil {
+		result = &ContentSecurityPolicy{}
```
tgross, line 62: Weirdly this looks like a bug in the other `Merge` implementations in this file which you've very reasonably copied here. If the original is nil, I'd expect us to return a copy of `other` instead. But unless you've got a way to verify whether that's right off the top of your head, let's keep it like this for now and I'll come back later to see if it's really a bug or not.

tgross (changes requested):

tgross (approved):
LGTM!

tgross:
Backported to the 1.6.x release branch

DingoEatingFuzz:
Thank you for your work on this @ebarriosjr! We handed you a bunch of scope creep and you implemented it all in stride 💚

PR #18093, opened 2023-07-28
GHA pinning updates

tgross:
Trusted Supply Chain Component Registry (TSCCR) enforcement starts Monday and an internal report shows our semgrep action is pinned to a version that's not currently permitted. Update all the action versions to whatever's the new hotness to maximum the time-to-live on these until we have automated pinning setup.

No backport labels because there's no possible way BPA is going to work with these changes so I'll just make the 6 PRs by hand.

### DIFF
--- a/.github/actions/setup-js/action.yml
+++ b/.github/actions/setup-js/action.yml
@@ -26,7 +26,7 @@ runs:
     # enforce node version (self-hosted default is not necessarily what we want)
     # after yarn, so we can use yarn for cache.
     - name: Setup node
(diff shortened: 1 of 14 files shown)


gulducat (approved):
funny that an action you're _not_ touching (setup-chromedriver) wanted to error in ember-test-audit.  I don't see it depending on anything special running prior to it...

tgross:
Updated the setup-chromedriver to a new pin and updated the pin in https://github.com/hashicorp/security-tsccr/pull/621

tgross:
Backported by hand to `release/1.6.x`, `release/1.5.x`, and `release/1.4.x`

PR #12097, opened 2022-02-20
Fix exec -job does not find jobs on wildcard namespace

leonardobsjr:
Right now, nomad exec -job is inconsistent with other commands like nomad status in the sense that it can't find an existing job if nomad namespace is "\*". The reason for that is that AllocsByJob doesn't handle '\*' properly as a nomad namespace value - it tries to search MemDB by using * and finds nothing because the jobs are registered in proper namespaces. To fix this, I've changed the behavior to, in case nomad namespace is '*', to search on all namespaces. The method fails if the JobID exists on multiple namespaces.

I'm pretty sure I should have written some tests over this, but I would love some guidance on where :)

PR #18095, opened 2023-07-28
cli: support wildcard namespace in alloc subcommands

tgross:
The alloc exec and filesystem/logs commands allow passing the `-job` flag to select a random allocation. If the namespace for the command is set to `*`, the RPC handler doesn't handle this correctly as it's expecting to query for a specific job. Most commands handle this ambiguity by first verifying that only a single object of the type in question exists (ex. a single node or job).

Update these commands so that when the `-job` flag is set we first verify there's a single job that matches. This also allows us to extend the functionality to allow for the `-job` flag to support prefix matching.

Fixes: #12097

---

What this looks like if there are multiple matching jobs:

```
$ nomad alloc exec -namespace '*' -job httpd /bin/sh
Prefix "httpd" matched multiple jobs

ID     Namespace  Type     Priority  Status   Submit Date
httpd  dev        service  50        running  2023-07-28T16:29:13-04:00
httpd  prod       service  50        running  2023-07-28T16:29:18-04:00
```

### DIFF
--- a/.changelog/18095.txt
+++ b/.changelog/18095.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+cli: support wildcard namespaces in alloc subcommands when the `-job` flag is used
+```
--- a/command/alloc_exec.go
+++ b/command/alloc_exec.go
@@ -51,7 +51,7 @@ Exec Specific Options:
     Sets the task to exec command in
 
   -job
-    Use a random allocation from the specified job ID.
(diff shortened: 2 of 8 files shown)


lgfa29 (approved):

tgross:
Failing test is because of https://github.com/hashicorp/nomad/pull/18100. I'll rebase this on `main` once that merges.

tgross:
Backported to 1.6.x

PR #18096, opened 2023-07-29
fix 'default' alias not added to interface specified by `network_interface` 

kevinschoonover:
closes https://github.com/hashicorp/nomad/issues/18097

consider the following config.hcl
```hcl
# config.hcl
client {
  network_interface = "tailscale0"

  host_network "tailscale" {
    interface = "tailscale0"
  }

  host_network "public" {
    interface = "wlan0"
  }
}
```

whenever you would schedule the following job 

```hcl
# job.hcl
job "docs" {
  datacenters = ["dc1"]

  group "example" {
    network {
      port "http" {}
    }
    task "server" {
      driver = "docker"

      config {
        image = "hashicorp/http-echo"
        ports = ["http"]
        args = [
          "-listen",
          ":5678",
          "-text",
          "hello world",
        ]
      }
    }
  }
}
```

you would see `* Constraint "missing host network \"default\" for port \"http\"": 1 nodes excluded by filter` because the 'default' host_network alias is being replaces with 'tailscale'. The follow PR appends the 'default' alias regardless to the addresses specified by `network_interface`.

### DIFF
--- a/.changelog/18096.txt
+++ b/.changelog/18096.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+fix: 'default' alias not added to interface specified by network_interface
+```
--- a/.changelog/18096.txt
+++ b/.changelog/18096.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-fingerprint: 'default' alias not added to interface specified by network_interface
+fingerprint: fix 'default' alias not being added to interface specified by network_interface
 ```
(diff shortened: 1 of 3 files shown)


tgross (changes requested):
LGTM! Nice work on this @kevinschoonover!

Two things:
* Can you run `make cl` here to add a changelog entry?
* Can you rebase on `main`? We accidentally landed a change in `main` that's breaking tests, which I've fixed in https://github.com/hashicorp/nomad/pull/18100 but you don't have on your branch.

tgross (approved):
LGTM!

PR #18100, opened 2023-07-31
scheduler: fix panic in `render_templates` destructive update check

tgross:
In #18054 we introduced a new field `render_templates` in the `restart` block. Previously changes to the `restart` block were always non-destructive in the scheduler but we now need to check the new field so that we can update the template runner. The check assumed that the block was always non-nil, which causes panics in our scheduler tests.

No changelog entry as this was never shipped.
Will be backported to 1.6.x


tgross:
Backported to 1.6.x.

schmichael:
Oh no, good catch Tim. Sorry for missing that in review.

PR #18102, opened 2023-07-31
[dep] bump golang.org/x/exp

angrycub:
There are some refactorings that have to be made in the getter and state_store where the api changed in `slices`


client/allocrunner/taskrunner/getter/params.go
```diff
@@ -109,13 +109,19 @@ func (p *parameters) Equal(o *parameters) bool {
...
 		return false
-	case !maps.EqualFunc(p.Headers, o.Headers, slices.Equal[string]):
+	case !maps.EqualFunc(p.Headers, o.Headers, HeadersCompareFn):
 		return false
 	}
 
 	return true
 }
 
+func HeadersCompareFn(a []string, b []string) bool {
```
shoenig, line 119: I think this can be private, since we aren't using from outside the package

nomad/state/state_store.go
```diff
@@ -5521,8 +5521,18 @@ func (s *StateStore) pruneJobSubmissions(namespace, jobID string, txn *txn) erro
...
+		var cmp int = 0
+		if a.First < b.First {
+			cmp = -1
+		}
+		if a.First > b.First {
+			cmp = +1
+		}
+
+		// Convert the sort into a descending sort by inverting the sign
+		cmp = cmp * -1
```
shoenig, line 5534: could alternatively just swap the cmp assignments above

shoenig (approved):
LGTM

tgross:

PR #18104, opened 2023-07-31
Update the test library

Juanadelacuesta:
Update the tests for the variables HTTP endpoint to use [must](https://pkg.go.dev/github.com/shoenig/test@v0.6.6) instead of [require](https://pkg.go.dev/github.com/stretchr/testify@v1.8.4/require)


tgross (approved):
LGTM!

Issue #18034, opened 2023-07-21
Allocation Logs Immediately Removed on Allocation Replacement


thetoothpick:
### Nomad version
```
# nomad -version
Nomad v1.5.6
BuildDate 2023-05-19T18:26:13Z
Revision 8af70885c02ab921dedbdf6bc406a1e886866f80
```

### Operating system and Environment details
```
# uname -a
Linux ip-10-3-6-38.ec2.internal 5.15.0-67-generic #74-Ubuntu SMP Wed Feb 22 14:14:39 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

### Issue
When our Nomad service allocations finish (either successfully or unsuccessfully) and are replaced by another allocation from the same task group on the same client, all allocation logs on the filesystem are immediately removed.  This makes it difficult to capture things that are only logged to standard out or error, especially if the allocation crashed or failed.  It seems like the entire allocation directory is garbage collected immediately when the allocation is replaced, but we haven't been able to find any logs indicating why (or that this is actually what's happening).

These are our GC-related settings:
```
server {
  job_gc_threshold              = "24h"
  eval_gc_threshold             = "24h"
  deployment_gc_threshold       = "24h"
  csi_volume_claim_gc_threshold = "3m"
}
```

We do have sticky ephemeral disks enabled, and the `alloc/data` directory is transferred to the new allocation successfully.

On some hosts that run fewer allocations, we're sometimes able to see logs for an extended period after the previous allocations stop and are replaced, so the problem isn't universal.  We've also checked the disk usage on the hosts when this happens and they're all very low (usually 2-5% used).

We have developed a "poststop" task to save the allocation logs after the allocation finishes, to try to combat this and help debug errors that occur in our jobs.  However, sometimes even this poststop task fails to start with this error:

```
logmon: rpc error: code = Unknown desc = failed to create stdout logfile for "save-nomad-logs.stdout": open /.../alloc/logs: no such file or directory.
```

This seems to indicate that the allocation (log) directory is being deleted before the allocation even completes.

We're not sure what might be causing this.

### Reproduction steps
1. Configure Nomad with the following garbage collection (GC) related settings:
   ```
   server {
     job_gc_threshold              = "24h"
     eval_gc_threshold             = "24h"
     deployment_gc_threshold       = "24h"
     csi_volume_claim_gc_threshold = "3m"
   }
   ```
2. Enable sticky ephemeral disks.
[...]

PR #18108, opened 2023-07-31
allocwatcher: don't destroy local allocdir after migration

tgross:
When `ephemeral_disk.migrate=true`, we make a best-effort attempt to migrate the allocation data between the old and new allocations. For previous allocations on other clients (the "remote" workflow), we create a local allocdir and download the data from the previous client into it. That data is then moved into the new allocdir and we delete the allocdir of the previous alloc.

For "local" previous allocations we don't need to create an extra directory for the previous allocation and instead move the files directly from one to the other. But we still delete the old allocdir _entirely_, which includes all the logs!

There doesn't seem to be any reason to destroy the local previous allocdir, as the usual client garbage collection should destroy it later on when needed. By not deleting it, the previous allocation's logs are still available for the user to read.

Fixes: #18034


shoenig (approved):
LGTM!

client/allocwatcher/alloc_watcher.go
```diff
@@ -279,15 +279,7 @@ func (p *localPrevAlloc) Migrate(ctx context.Context, dest *allocdir.AllocDir) e
 
 	p.logger.Debug("copying previous alloc")
 
-	moveErr := dest.Move(p.prevAllocDir, p.tasks)
-
-	// Always cleanup previous alloc
-	if err := p.prevAllocDir.Destroy(); err != nil {
```
schmichael, line 285: This removed a call to `p.prevAllocDir.UnmountAll()` as well, which among other benign effects, leaves `/secrets` laying around. I think we should add it back in.

tgross: Doesn't that get cleaned up with the rest of the allocation GC? I guess I'm not sure why we special case removing bits of the previous allocdir, but only if we've migrated data. Ex. if we replaced the allocation without migration we wouldn't be doing anything to the previous allocdir that wasn't done as part of that previous allocation's lifecycle.

schmichael: Oh good point. Not sure why this was special cased.

schmichael (approved):

tgross:
BPA failed at random for the 1.4.x release branch, I've backported that by hand.

PR #18109, opened 2023-08-01
docs: expand documentation on node pools

lgfa29:
Add new Concepts page about node pools and expand the Architecture page to note planning client configuration. Also include some smaller changes where node pools may affect the scheduling outcome.

Preview links for the biggest changes:
https://nomad-git-docs-node-pools-hashicorp.vercel.app/nomad/docs/concepts/architecture#client-organization
https://nomad-git-docs-node-pools-hashicorp.vercel.app/nomad/docs/concepts/node-pools

### DIFF
--- a/website/content/docs/concepts/architecture.mdx
+++ b/website/content/docs/concepts/architecture.mdx
@@ -176,105 +176,6 @@ on hardware features such as architecture and availability of GPUs, or software 
 like operating system and kernel version, or they can be business constraints like
 ensuring PCI compliant workloads run on appropriate servers.
 
-# Client Organization
-
-When the Nomad [scheduler][] receives a job registration request, it needs to
-determine which clients will run allocations for the job. By default, all nodes
-are considered for placements but this process can be adjusted by cluster
-operators and job submitters to achieve more control over where allocations
-are placed.
-
-The steps to achieve this control consists of setting certain values in
-clients' configuration and matching them in jobs, and so it's helpful to
-consider how clients will be configured based on your expected job registration
-patterns when planning your cluster deployment.
-
-There are several options that can be used depending on the desired outcome.
-
-### Affinities and Constraints
-
-Affinities and constraints allow users to define soft or hard requirements for
-their jobs. The [`affinity`][job_affinity] block specifies a soft requirement
-on certain node properties, meaning allocations for the job have a preference
-for some nodes, but may be placed elsewhere if the rules can't be matched,
-while the [`constraint`][job_constraint] block creates hard requirements and
-allocations can only be placed in nodes that match these rules. Job placement
-fails if a constraint cannot be satisfied.
-
-These rules can reference intrinsic node characteristics, which are called
-[node attributes][] and are automatically detected by Nomad, static values
-defined in the agent configuration file by cluster administrators, or dynamic
-values defined after the agent starts.
-
-Using affinities and constraints has the downside of only allowing allocations
-to gravitate towards certain nodes, but it does not prevent placements of jobs
-where the rules don't apply.
-
-Use affinities and constraints when certain jobs have certain node preferences
-but is acceptable to have other jobs sharing the same nodes.
-
-The sections below describe the node values that can be configured and used in
-job affinity and constraint rules.
-
-#### Node Class
-
-Node class is an arbitrary value that can be used to group nodes based on some
-characteristics, like the instance size or the presence of fast hard drives,
-and is specified in the client configuration file using the
-[`node_class`][config_client_node_class] parameter.
-
-#### Dynamic and Static Node Metadata
-
-Node metadata are arbitrary key-value mappings specified either in the client
-configuration file using the [`meta`][config_client_meta] parameter or
-dynamically via the [`nomad node meta`][cli_node_meta] command and the
-[`/v1/client/metadata`][api_client_metadata] API endpoint.
-
-There are no preconceived use cases for metadata values, and each team may
-choose to use them in different ways. Some examples of static metadata include
-resource ownership, such as `owner = "team-qa"`, or fine-grained locality,
-`rack = "3"`. Dynamic metadata may be used to track runtime information, such
-as jobs running in a given client.
-
-### Datacenter
-
-Datacenters represent a geographical location in a region that can be used for
-fault tolerance and infrastructure isolation.
-
-It is defined in the agent configuration file using the
-[`datacenter`][config_datacenter] parameter and, unlike affinities and
-constraints, datacenters are opt-in at the job level, meaning that a job only
-places allocations in the datacenters it uses, and, more importantly, only jobs
-in a given datacenter are allowed to place allocations in those nodes.
-
-Given the strong connotation of a geographical location, use datacenters to
-represent where a node resides rather than its intended use. The
-[`spread`][] block can help achieve fault tolerance across datacenters.
-
-### Node Pool
-
-Node pools allow grouping nodes that can be targeted by jobs to achieve
-workload isolation.
-
-Similarly to datacenters, node pools are configured in an agent configuration
-file using the [`node_pool`][config_client_node_pool] attribute, and are opt-in
-on jobs, allowing restricted use of certain nodes by specific jobs without
-extra configuration.
-
-But unlike datacenters, node pools don't have a preconceived notion and can be
-used for several use cases, such as segmenting infrastructure per environment
-(development, staging, production), by department (engineering, finance,
-support), or by functionality (databases, ingress proxy, applications).
-
-Node pools are also a first-class concept and can hold additional [metadata and
-configuration][spec_node_pool].
-
-Use node pools when there is a need to restrict and reserve certain nodes for
-specific workloads, or when you need to adjust specific [scheduler
-configuration][spec_node_pool_sched_config] values.
-
-Refer to the [Node Pools][concept_np] concept page for more information.
-
 # Getting in Depth
 
 This has been a brief high-level overview of the architecture of Nomad. There
@@ -285,20 +186,6 @@ are all documented in more detail.
 For other details, either consult the code, [open an issue on
 GitHub][gh_issue], or ask a question in the [community forum][forum].
 
-[`spread`]: /nomad/docs/job-specification/spread
 [`update`]: /nomad/docs/job-specification/update
-[api_client_metadata]: /nomad/api-docs/client#update-node-metadata
-[cli_node_meta]: /nomad/docs/commands/node/meta
-[concept_np]: /nomad/docs/concepts/node-pools
-[config_datacenter]: /nomad/docs/configuration#datacenter
-[config_client_node_class]: /nomad/docs/configuration/client#node_class
-[config_client_meta]: /nomad/docs/configuration/client#meta
-[config_client_node_pool]: /nomad/docs/configuration/client#node_pool
 [gh_issue]: https://github.com/hashicorp/nomad/issues/new/choose
 [forum]: https://discuss.hashicorp.com/c/nomad
-[job_affinity]: /nomad/docs/job-specification/affinity
-[job_constraint]: /nomad/docs/job-specification/constraint
-[node attributes]: /nomad/docs/runtime/interpolation#node-attributes
-[scheduler]: /nomad/docs/concepts/scheduling/scheduling
-[spec_node_pool]: /nomad/docs/other-specifications/node-pool
-[spec_node_pool_sched_config]: /nomad/docs/other-specifications/node-pool#scheduler_config-parameters
--- a/website/content/docs/concepts/architecture.mdx
+++ b/website/content/docs/concepts/architecture.mdx
@@ -93,6 +93,18 @@ Nomad models infrastructure as regions and datacenters. A region will contain
 one or more datacenters. A set of servers joined together will represent a
 single region. Servers federate across regions to make Nomad globally aware.
 
+In federated clusters one of the regions must be defined as the [authoritative
+region](#authoritative-and-non-authoritative-regions).
+
+#### Authoritative and Non-Authoritative Regions
+
+The authoritative region is the region in a federated multi-region cluster that
+holds the source of true for entities replicated across regions, such as ACL
+tokens, policies, and roles, namespaces, and node pools.
+
+All other regions are considered non-authoritative regions and replicate these
+entities by pulling them from the authoritative region.
+
 #### Datacenters
 
 Nomad models a datacenter as an abstract grouping of clients within a
--- a/website/content/docs/concepts/architecture.mdx
+++ b/website/content/docs/concepts/architecture.mdx
@@ -101,6 +101,20 @@ servers they are joined with, but do need to be in the same
 region. Datacenters provide a way to express fault tolerance among jobs as
 well as isolation of infrastructure.
 
+#### Node
+
+A more generic term used to refer to machines running Nomad agents in client
+mode. Unless noted otherwise, it may be used interchangeably with
+[client](#client).
+
+#### Node Pool
+
+Node pools are used to group [nodes](#node) and can be used to restrict which
+[jobs](#job) are able to place [allocations](#allocation) in a given set of
+nodes. Example use cases for node pools include segmenting nodes by environment
+(development, staging, production), by department (engineering, finance,
+support), or by functionality (databases, ingress proxy, applications).
+
 #### Bin Packing
 
 Bin Packing is the process of filling bins with items in a way that maximizes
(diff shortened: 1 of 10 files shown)


website/content/docs/concepts/architecture.mdx
```diff
@@ -162,14 +176,129 @@ on hardware features such as architecture and availability of GPUs, or software
 like operating system and kernel version, or they can be business constraints like
 ensuring PCI compliant workloads run on appropriate servers.
 
+# Client Organization
```
lgfa29, line 179: I'm not sure if this is the best place to have this, or if this is a good name, so I'm open to suggestions 😄

website/content/docs/concepts/architecture.mdx
```diff
@@ -162,14 +176,129 @@ on hardware features such as architecture and availability of GPUs, or software
...
+fails if a constraint cannot be satisfied.
+
+These rules can reference intrinsic node characteristics, which are called
+[node attributes][] and are automatically detected by Nomad, static values
+defined in the agent configuration file by cluster administrators, or dynamic
+values defined after the agent starts.
+
+Using affinities and constraints has the downside of only allowing allocations
+to gravitate towards certain nodes, but it does not prevent placements of jobs
+where the rules don't apply.
```
tgross, line 211:
```suggestion
Using affinities and constraints has the downside of only allowing allocations
to gravitate towards certain nodes, but it does not prevent placements of other
jobs for which the rules don't apply.
```

I think?

website/content/docs/concepts/architecture.mdx
```diff
@@ -162,14 +176,129 @@ on hardware features such as architecture and availability of GPUs, or software
...
+
+Use affinities and constraints when certain jobs have certain node preferences
+but is acceptable to have other jobs sharing the same nodes.
```
tgross, line 214:
```suggestion
Use affinities and constraints when certain jobs have certain node preferences
but it is acceptable to have other jobs sharing the same nodes.
```

website/content/docs/concepts/architecture.mdx
```diff
@@ -162,14 +176,129 @@ on hardware features such as architecture and availability of GPUs, or software
...
+used for several use cases, such as segmenting infrastructure per environment
+(development, staging, production), by department (engineering, finance,
+support), or by functionality (databases, ingress proxy, applications).
+
+Node pools are also a first-class concept and can hold additional [metadata and
+configuration][spec_node_pool].
+
+Use node pools when there is a need to restrict and reserve certain nodes for
+specific workloads, or when you need to adjust specific [scheduler
+configuration][spec_node_pool_sched_config] values.
```
tgross, line 274: Probably a nice place to plug that node pools can be associated with namespaces in Nomad Enterprise.

website/content/docs/concepts/node-pools.mdx
```diff
@@ -0,0 +1,366 @@
...
+client {
+  # ...
+  node_pool = "dev"
+  # ...
+}
+```
+
+To help streamline this process, nodes can create node pools on demand. If a
+client configuration references a node pool that does not exist yet, Nomad
+creates the node pool automatically on client registration.
```
tgross, line 52:
```suggestion
To help streamline this process, nodes can create node pools on demand. If a
client configuration references a node pool that does not exist yet, Nomad
creates the node pool automatically on client registration. (However, see
[Multi-region Clusters](#multi-region-clusters) for an exception.)
```

website/content/docs/concepts/node-pools.mdx
```diff
@@ -0,0 +1,366 @@
...
+group resources in isolated logical units. Jobs are grouped into namespaces and
+clients into node pools.
+
+With the Node Pool Governance feature of Nomad Enterprise it is possible to
+automatically link a node pool to a namespace, and so a job is able to simply
+specify a namespace and the node pool is inferred from the namespace
+configuration.
+
+This connection is done using the [`node_pool_config ->
+default`][ns_spec_np_default] attribute in the namespace specification.
```
tgross, line 131: I'm pretty sure we always use a dot to "dereference" field sections like this:

```suggestion
This connection is done using the [`node_pool_config.default`][ns_spec_np_default] 
attribute in the namespace specification.
```

Another way to do this is to reword to avoid it , like "This connection is done using the `default` attribute in the namespace specification's `node_pool_config` block."

website/content/docs/concepts/node-pools.mdx
```diff
@@ -0,0 +1,366 @@
...
+node pool specification for more information.
+
+### Node Pool Governance
+
+Node pools and namespaces share some similarities, with both providing a way to
+group resources in isolated logical units. Jobs are grouped into namespaces and
+clients into node pools.
+
+With the Node Pool Governance feature of Nomad Enterprise it is possible to
+automatically link a node pool to a namespace, and so a job is able to simply
```
tgross, line 126:
```suggestion
automatically assign a node pool to a namespace, and so a job is able to simply
```

Maybe?

website/content/docs/concepts/node-pools.mdx
```diff
@@ -0,0 +1,366 @@
...
+```hcl
+job "app-dev" {
+  # The "dev" node pool will be used because it is the
+  # namespace's default node pool.
+  namespace = "dev"
+  # ...
+}
+```
+
+Jobs are able o override the namespace default node pool by specifying a
```
tgross, line 156:
```suggestion
Jobs are able to override the namespace default node pool by specifying a
```

website/content/docs/concepts/node-pools.mdx
```diff
@@ -0,0 +1,366 @@
...
+```
+
+## Node Pool Patterns
+
+The sections below describe some node pool patterns that can be used to achieve
+specific goals.
+
+### Infrastructure and System Jobs
+
+It is a common for a Nomad cluster to have certain jobs that are focused on
```
tgross, line 219:
```suggestion
It is common for a Nomad cluster to have certain jobs that are focused on
```

tgross (commented):
This is looking great @lgfa29! I'm about halfway thru but wanted to leave comments and then come back to finish the rest of my read-thru later today.

website/content/docs/concepts/node-pools.mdx
```diff
@@ -0,0 +1,366 @@
...
+workload as possible in the given set of of clients. This makes it ideal for
+cloud environments where infrastructure is billed by the hour and can be
+quickly scaled in and out. By maximizing workload density a cluster running in
+cloud instances can reduce the number of clients needed to run everything that
+is necessary.
+
+The `spread` algorithm behaves in the opposite direction, making use of every
+client available to reduce density and potential noisy neighbors and resource
+contention. This makes it ideal for environments where clients are
+pre-provisioned and scale more slowly, such as on-premise deployments.
```
tgross, line 329:
```suggestion
pre-provisioned and scale more slowly, such as on-premises deployments.
```

IMO, this is one of those things the "grammar police" get overly worked up about, but better to avoid that discussion. :grinning:

tgross (approved):
This looks great @lgfa29! I've left a few more comments but I think if we resolve those we're good-to-go here.

website/content/docs/concepts/architecture.mdx
```diff
@@ -162,14 +176,129 @@ on hardware features such as architecture and availability of GPUs, or software
...
 
+# Client Organization
+
+When the Nomad [scheduler][] receives a job registration request, it needs to
+determine which clients will run allocations for the job. By default, all nodes
+are considered for placements but this process can be adjusted by cluster
+operators and job submitters to achieve more control over where allocations
+are placed.
+
+The steps to achieve this control consists of setting certain values in
```
Juanadelacuesta, line 187: If this paragraph is used to introduce Affinities and Constraints, Datacenter, Nodepools... it would be worth mentioning them.

website/content/docs/concepts/architecture.mdx
Juanadelacuesta: Im a little confused here: constraints are hard requirements and "Job placement fails if a constraint cannot be satisfied." but at the same time they don't " it does not prevent placements of jobs where the rules don't apply" ?

website/content/docs/concepts/node-pools.mdx
```diff
@@ -0,0 +1,366 @@
...
+
+Similarly to the `namespace` attribute, the node pool must exist beforehand,
+otherwise the job registration results in an error. Only nodes in the given
+node pool are considered for placement. If none are available the deployment
+is kept as pending until a client is added to the node pool.
+
+## Multi-region Clusters
+
+In federated multi-region clusters, node pools are automatically replicated
+from the authoritative region to all non-authoritative regions, and requests to
```
Juanadelacuesta, line 72: Not directly related, but I couldn't find what authoritative regions are, maybe add a little phrase about it on the glossary on the region?

website/content/docs/concepts/node-pools.mdx
```diff
@@ -0,0 +1,366 @@
...
+applied.
+
+Refer to the [`scheduler_config`][np_spec_scheduler_config] parameter in the
+node pool specification for more information.
+
+### Node Pool Governance
+
+Node pools and namespaces share some similarities, with both providing a way to
+group resources in isolated logical units. Jobs are grouped into namespaces and
+clients into node pools.
```
Juanadelacuesta, line 123: This phrase is very clear and useful, I would add it in the node pools description

website/content/docs/concepts/architecture.mdx
```diff
@@ -101,6 +101,20 @@ servers they are joined with, but do need to be in the same
 region. Datacenters provide a way to express fault tolerance among jobs as
 well as isolation of infrastructure.
 
+#### Node
+
+A more generic term used to refer to machines running Nomad agents in client
+mode. Unless noted otherwise, it may be used interchangeably with
+[client](#client).
```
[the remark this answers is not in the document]
shoenig: I kind of disagree with noting that it may be used interchangeably (even though it often is). Node is intended to refer to the machine (and operating system, etc.) while Client refers to the Nomad agent running in client mode. We should avoid conflating the terms in our own docs.

website/content/docs/concepts/architecture.mdx
lgfa29: > I think?

Yup, that's right. Thanks!

> Im a little confused here: constraints are hard requirements and "Job placement fails if a constraint cannot be satisfied." but at the same time " it does not prevent placements of jobs where the rules don't apply" ?

Yeah, it's kind of weird to describe in text, so maybe I should put an example.

But imagine you have two jobs, one with a constraint and one without any constraints:

```hcl
job "app"  {
  constraint {
    attribute = "${meta.env}"
    value      = "prod"
  }
  # ...
}
```

```hcl
job "db"  {
  # ...
}
```

The `app` job is only allowed to run in clients with metadata `env=prod`, but the `db` is free to run anywhere, including clients with `env=prod`. 

So if you want to reserve the `env=prod` clients to only run `app` then you need a negative constraint in _all other_ jobs:

```hcl
job "db"  {
  constraint {
    attribute = "${meta.env}"
    operator  = "!="
    value      = "prod"
  }
  # ...
}
```

So constraints are kind of unidirectional in this sense: you can restrict a _job_ to specific nodes, but you can't restrict a _node_ to specific jobs.

Does this help?

lgfa29: Updated this paragraph as I moved this section to the `Scheduler` section
https://github.com/hashicorp/nomad/blob/4f4ba9fdfc71eab2665c6091b249b711c6eb1f81/website/content/docs/concepts/scheduling/placement.mdx#L36-L38

Is this better or should I add an example to help illustrate the problem?

website/content/docs/concepts/architecture.mdx
lgfa29: Good idea. Added a blurb about it after moving this content.
https://github.com/hashicorp/nomad/blob/4f4ba9fdfc71eab2665c6091b249b711c6eb1f81/website/content/docs/concepts/scheduling/placement.mdx#L104-L105

website/content/docs/concepts/node-pools.mdx
lgfa29: Good call. I actually forgot about this until I wrote the multi-region section but then forgot to come back here 😅

website/content/docs/concepts/node-pools.mdx
lgfa29: Oh right. I think we only use the arrow syntax in the `Placement` tables in jobspec and others.

website/content/docs/concepts/node-pools.mdx
lgfa29: Yup! But reading this paragraph again I found it a bit confusing (the "automatically" part is the job -> node pool assignment, not node pool -> namespace 😅), so I rewrote it like this:
```
Node Pool Governance allows assigning a default node pool to a namespace that
is automatically used by every job registered to the namespace. This feature
simplifies job management as the node pool is inferred from the namespace
configuration instead of being specified in every job.
```

website/content/docs/schedulers.mdx
```diff
@@ -60,6 +60,10 @@ Systems jobs are intended to run until explicitly stopped either by an operator
 or [preemption]. If a system task exits it is considered a failure and handled
 according to the job's [restart] block; system jobs do not have rescheduling.
 
+When used with node pools, system jobs run in all nodes of the pool used by the
+job. The built-in node pool `all` allows placing allocations in all clients in
+the cluster.
```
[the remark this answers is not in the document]
lgfa29: It doesn't matter how many images of circles and boxes I see, I will get this wrong all the time 😄

website/content/docs/concepts/node-pools.mdx
lgfa29: I don't think I ever spelled out "on-premises" in full, I usually just go with "on-prem", so good to know 😄

website/content/docs/concepts/architecture.mdx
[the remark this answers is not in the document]
lgfa29: > when can it be different?

"Node" is a very generic term that different people use it to mean different things 😅 

For example, a node in a network context can be anything sending or receiving traffic, like a computer, a router, a switch, or maybe even an allocation, like a sidecar proxy. So if you're talking about Nomad's network topology, "node" may not necessarily mean this specific definition.

But to @shoenig's point, I think I didn't express myself well here. I meant more like "you may see or hear people using them interchangeably", which I have in informal conversations, issues, community blog posts etc.

Maybe this sentence would be better like this:

```
Despite being different concepts, you may find the term "node" being used interchangeably with "client" in some materials.
```

> Node is intended to refer to the machine (and operating system, etc.)

To the machine running a Nomad client right? I don't recall us using "node" to refer to machines running servers. For example, the `nomad node` commands and `/v1/node/` endpoints are only about clients.

website/content/docs/concepts/architecture.mdx
lgfa29: Hum...I don't quite follow. I think the confusion may be that I'm using "setting" as the gerund of the verb "to set", not as "configuration".

But this paragraph is gone once I moved this section to the new Scheduling -> Placements page 😅

website/content/docs/concepts/node-pools.mdx
lgfa29: Thanks! Which node pool description are you referring to?

website/content/docs/concepts/node-pools.mdx
lgfa29: Ah good point, I added an entry to the glossary about authoritative and non-authoritative regions. I used the same entry because they're opposites so it was kind of weird to explain one without explaining the other 😅 

https://github.com/hashicorp/nomad/pull/18109/files#diff-81af9e6ca20054c572009d730ea51e6a6563ef34befcc20a398fac0c3a6d68d9R96-R106

website/content/docs/concepts/node-pools.mdx
```diff
@@ -0,0 +1,366 @@
...
+    }
+  }
+  # ...
+}
+```
+
+## Node Pool Patterns
+
+The sections below describe some node pool patterns that can be used to achieve
+specific goals.
```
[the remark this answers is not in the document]
lgfa29: Just to make sure I understood correctly, were you thinking about something like this?

https://github.com/hashicorp/nomad/blob/3a53b9c542a96772fc2cdda70153aac9f8c4b2c5/website/content/docs/concepts/node-pools.mdx#L223-L227

https://github.com/hashicorp/nomad/blob/3a53b9c542a96772fc2cdda70153aac9f8c4b2c5/website/content/docs/concepts/node-pools.mdx#L324-L327

website/content/docs/concepts/architecture.mdx
lgfa29: Moved to its own page and adjusted the first section a bit:
https://github.com/hashicorp/nomad/pull/18109/files#diff-933fc372992990c21ae0cd52f2e8106d7c7c749fc34bb2617731c75477811de7R7-R19

website/content/docs/concepts/node-pools.mdx
tgross: Nice :+1:

website/content/docs/concepts/node-pools.mdx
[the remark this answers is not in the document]
tgross: Yeah

website/content/docs/concepts/architecture.mdx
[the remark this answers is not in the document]
lgfa29: I ping-ponged a bit on how to phrase this and ended up here:
https://github.com/hashicorp/nomad/blob/efc2e3746019c7c62930dab14a02c0377475a79c/website/content/docs/concepts/architecture.mdx#L116-L120

Is this a better description? Or maybe I should just leave this part out?

PR #18137, opened 2023-08-03
Add retry mechanism for put call

Juanadelacuesta:
The SDK for the locking could benefit from a retry mechanism for calls made to the nomad server, this PR introduces this mechanism on top of the put call already present on the api.

### DIFF
--- a/api/retry.go
+++ b/api/retry.go
@@ -14,11 +14,6 @@ import (
 const (
 	defaultNumberOfRetries = 5
 	defaultDelayTimeBase   = time.Second
-)
-
-var (
-	// defaultMaxBackoffDelay is defined as a variable so it can be modified
-	// to speed up the testing process.
 	defaultMaxBackoffDelay = 5 * time.Minute
 )
 
@@ -123,12 +118,16 @@ func (c *Client) calculateDelay(attempt int64) time.Duration {
 
 	// Ensure that a big attempt number or a big delayBase number will not cause
 	// a negative delay by overflowing the delay increase.
-	if math.MaxInt64/c.config.retryOptions.delayBase.Nanoseconds() < (1 << attempt) {
-		return defaultMaxBackoffDelay
+	// TODO: This value should be calculated on configuration and set the
+	// to default if not provided.
+	maxValidAttempt := int64(math.Log2(float64(math.MaxInt64 /
+		c.config.retryOptions.delayBase.Nanoseconds())))
+
+	if attempt > maxValidAttempt {
+		return c.config.retryOptions.maxBackoffDelay
 	}
 
 	newDelay := c.config.retryOptions.delayBase << (attempt - 1)
-
 	if c.config.retryOptions.maxBackoffDelay != defaultMaxBackoffDelay &&
 		newDelay > c.config.retryOptions.maxBackoffDelay {
 		return c.config.retryOptions.maxBackoffDelay
--- a/api/retry.go
+++ b/api/retry.go
@@ -6,17 +6,38 @@ package api
 import (
 	"context"
 	"errors"
+	"math"
 	"net/http"
 	"time"
 )
 
+const (
+	defaultNumberOfRetries = 5
+	defaultDelayTimeBase   = time.Second
+)
+
+var (
+	// defaultMaxBackoffDelay is defined as a variable so it can be modified
+	// to speed up the testing process.
+	defaultMaxBackoffDelay = 5 * time.Minute
+)
+
+type client interface {
+	put(endpoint string, in, out any, q *WriteOptions) (*WriteMeta, error)
+}
+
+type retryClient struct {
+	c client
+	retryOptions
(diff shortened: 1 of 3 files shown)


api/retry.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	return rc
+}
+
+func (rc *retryClient) retryPut(ctx context.Context, endpoint string, in, out any, q *WriteOptions) (*WriteMeta, error) {
+	var err error
+	var wm *WriteMeta
+
+	iDelay := time.Duration(0)
```
jrasell, line 67: `iDelay` for initial delay? If so, we use it for all delays, so maybe `delayDuration`?

jrasell (commented):
Looking good. Added some questions and think it would be good to have some more tests to cover functions such as the `calculateDelay` function.

api/retry.go
```diff
@@ -0,0 +1,136 @@
...
+func (rc *retryClient) calculateDelay(attempt uint64) time.Duration {
+	if rc.FixedDelay != 0 {
+		return rc.FixedDelay
+	}
+
+	if attempt == 0 {
+		return 0
+	}
+
+	new := rc.DelayBase << (attempt - 1)
```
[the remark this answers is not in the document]
Juanadelacuesta: Done!

api/retry.go
Juanadelacuesta: it was for iteration delay, but lets change it for attempt delay :)

api/retry.go
```diff
@@ -0,0 +1,136 @@
...
+		// don't retry
+		if !isCallRetriable(callErr.StatusCode()) {
+			break
+		}
+	}
+
+	return wm, err
+}
+
+func isCallRetriable(statusCode int) bool {
```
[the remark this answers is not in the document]
Juanadelacuesta: That was the best I could come up with jejej any ideas?

api/retry.go
```diff
@@ -0,0 +1,136 @@
...
+	if opts.MaxToLastCall != 0 {
+		rc.MaxToLastCall = opts.MaxToLastCall
+	}
+
+	if opts.FixedDelay != 0 {
+		rc.FixedDelay = opts.FixedDelay
+	}
+
+	return rc
+}
```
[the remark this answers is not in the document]
tgross: So if we added the `retryOptions` to `api.Config` then we wouldn't need to have a special `retryClient` struct, and the standard `api.Client` could use the retry options whenever it's been configured that way.

```go
type Config struct {
  // ... there's a bunch of existing field here
  retry retryOptions
}
```

And then later if we want to, we could make the retry options public for all `api` package consumers (ex. could the Terraform provider use this?)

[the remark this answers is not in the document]
Juanadelacuesta: Makes sense :)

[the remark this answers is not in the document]
tgross: I was thinking about this last night and there's no real reason to hold this PR up for this though. The new code you've added isn't exported, so it's not part of the public API and will be safe to update later on without breaking backwards compatibility. Let's ship and do any changes like this later if we want.

tgross (approved):

api/retry.go
```diff
@@ -0,0 +1,110 @@
...
+func (c *Client) calculateDelay(attempt uint64) time.Duration {
+	if c.config.retryOptions.fixedDelay != 0 {
+		return c.config.retryOptions.fixedDelay
+	}
+
+	if attempt == 0 {
+		return 0
+	}
+
+	newDelay := c.config.retryOptions.delayBase << (attempt - 1)
```
stswidwinski, line 103: I believe this will overflow quite quickly similarly to what is shown here: https://github.com/hashicorp/nomad/pull/18200

tgross: Good catch. @Juanadelacuesta note that we don't want to pull in the `helper` into this `api` package, so we'll need to duplicate the logic we added in #18200 here.

Juanadelacuesta: Should we set a limit to the max number of attempts?

api/retry.go
```diff
@@ -0,0 +1,138 @@
...
+
+	if attempt == 0 {
+		return 0
+	}
+
+	// Ensure that a big attempt number or a big delayBase number will not cause
+	// a negative delay by overflowing the delay increase.
+	if math.MaxInt64/c.config.retryOptions.delayBase.Nanoseconds() < (1 << attempt) {
+		return defaultMaxBackoffDelay
+	}
```
tgross, line 128: This doesn't work to protect against overflow in the bitshift, because it does the bitshift in order to do the check. Without a low cap on the number of attempts it'd be easy to cause this to overflow. See https://go.dev/play/p/DztXnyHL9hk for the results.

Instead, we should be tracking the current delay in the caller and returning early if that delay exceeds the max backoff (or default max backoff, if no max backoff is set).

Issue #18112, opened 2023-08-01
Cannot set GPUs using UUID as constraint


ruspaul013:
### Nomad version
nomad server: `Nomad v1.5.6`
nomad client: `Nomad v1.5.6`

### Operating system and Environment details
Plugin "nomad-driver-podman" v0.5.0
Plugin "nomad-device-nvidia" v1.0.0

### Issue
As it was mention in #15455, I could use  UUIDs of GPU as attribute to set constraints. But when I want to use specific GPUs based on their UUID, I still get random GPUs.

### Reproduction steps
Create a job file where you use multiple GPUs and set constraint based on their UUID.

#### Expected Result
running `nvidia-smi -L` inside container and see GPUs with the same UUID from job file.

#### Actual Result
running `nvidia-smi -L` inside container and get random GPUs.

### Job file (if appropriate)
```
job "test-2070-2" {
  datacenters = ["dc1"]
  group "test-2070-2" {

    restart {
        attempts=0
    }
    count=1
    task "test-2070-2" {
        driver = "podman"
        config {
            image = "image_with_gpu"
        }
        resources {
            cpu = 2650
            memory = 8192
            device "nvidia/gpu" {
                count = 2

                constraint {
                    attribute = "${device.model}"
                    value     = "NVIDIA GeForce RTX 2070 SUPER"
                }

                constraint {
                    attribute = "${device.ids}"
                    operator  = "set_contains"
                    value     = "GPU-9b5df054-6f08-f35c-9c4c-5709b19efea5,GPU-1846fc5f-8c71-bfab-00e1-9c190dd88ed7"
                }

            }
        }
    }
  }
}
```

results:
```
[root@481a2da8e0a9 /]# nvidia-smi -L
GPU 0: NVIDIA GeForce RTX 2070 SUPER (UUID: GPU-d7574813-0b3f-ee8f-39fc-2b48f9dff169)
GPU 1: NVIDIA GeForce RTX 2070 SUPER (UUID: GPU-9b5df054-6f08-f35c-9c4c-5709b19efea5)
```

tgross:
Hi @ruspaul013! Constraints only inform the scheduler to place the allocations on a client that has those GPU IDs in its fingerprint. Once the allocation has been placed, it's up to the task driver / device driver to select specific GPU devices on the host.

Can you confirm that the client your workloads are being placed on has the UUID you want, but that there just happen to be other GPUs on the same host and the selection is apparently random among them?

[...]

PR #18141, opened 2023-08-03
scheduler: filter device instance IDs by constraints

tgross:
When the scheduler assigns a device instance, it iterates over the feasible devices and then picks the first instance with availability. If the jobspec uses a constraint on device ID, this can lead to buggy/surprising behavior where the node's device matches the constraint but then the individual device instance does not.

Add a second filter based on the `${device.ids}` constraint after selecting a node's device to ensure the device instance ID falls within the constraint as well.

Fixes: #18112
See https://github.com/hashicorp/nomad/issues/18112#issuecomment-1664162029 for reproduction steps.
Will get backported to 1.6.x, 1.5.x, and 1.4.x

### DIFF
--- a/.changelog/18141.txt
+++ b/.changelog/18141.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+scheduler: Fixed a bug where device IDs were not correctly filtered in constraints
+```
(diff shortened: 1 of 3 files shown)


gulducat (approved):
LGTM!

lgfa29 (approved):

Issue #18113, opened 2023-08-01
server leader memory keeps increasing


shantanugadgil:
### Nomad version
Nomad v1.6.1
BuildDate 2023-07-21T13:49:42Z
Revision 515895c7690cdc72278018dc5dc58aca41204ccc

(this issue has always existed in all previous versions)

### Operating system and Environment details
Amazon Linux 2

### Issue
The server leader's memory keeps increasing with no sign of it flattening.
We have to periodically restart the leader to counter this problem

### Reproduction steps
You should be able to observe a graph similar to the following in you environment.

1. Leader changed within the cluster, but notice that the previous leader's memory never reduces. Also new leader's memory starts climbing rapidly.
3. Previous leader, memory never reduces but keeps increasing gradually

![image](https://github.com/hashicorp/nomad/assets/2508915/42441cb0-3899-49a1-975f-66b8df9bad17)

#### Expected Result
Server leader's memory should stay under control.

We run `nomad system gc` and `nomad system reconcile summaries` every hour, but that does not help either.

#### Actual Result
Leader memory keeps increasing

### Job file (if appropriate)

N/A

### Nomad Server logs (if appropriate)

N/A

### Nomad Client logs (if appropriate)

N/A

tgross:
Hi @shantanugadgil! 

Most of the time when folks report high/climbing memory usage on the servers it's because of evaluations piling up from large batch workloads that aren't getting GC'd. But if that were the case here I'd expect to see the memory increase across _all_ the servers (not just the leader) and `nomad system gc` would help.

I would definitely expect the leader to use moderately more memory than other servers, because it does quite a bit more work. It has long-running processes that startup at leader transition, so I'd expect there to be a spike right away. But as you've noted if the memory isn't yielding back to the OS a while after the leader transition, either.

So all that suggests that we've got a leak. Given that you can reproduce this reliably, can you generate the following profiles for me from the leader and one of the servers that has stepped down from leadership for a little while (at least a few minutes)?
* `/debug/pprof/heap`
* `/debug/pprof/goroutine`

If you don't have `enable_debug` on you can use the [Agent Runtime Profiles API](https://developer.hashicorp.com/nomad/api-docs/agent#agent-runtime-profiles) instead, just make sure you'd passing the right `server_id` param.

[...]

PR #18142, opened 2023-08-03
docs: fix URL in agent pprof examples

tgross:
From: https://github.com/hashicorp/nomad/issues/18113#issuecomment-1664520478


schmichael (approved):
:bow:

PR #18146, opened 2023-08-03
numa: enable numa topology detection

shoenig:
Reviewers: see NMD-183 (internal).

This change looks larger than it is, just due to the amount of deletions and fallout in plumbing things through. It's crudely broken up into 9 commits (by filepath):

b17d4182fe67d96a53b223b8390e7a9671dfd722 - refactor how the client handles cgroups (moving responsibility from the client cpuset manager + task drivers into mostly just task hooks).

2121b44c8de4c6173a674bb99113fb276dfc18ff - fingerprint numa topology (and eliminate aws cpu fingerprinter)

199dcb29f7f6acfdff49b2d189ed3527f53d5ec9 - plumbing numa and cgroups changes into drivers

1cd52649bf673f5ea11396dbabbeb3706e98f823 - cleaning up task resource usage accounting

200a275a3a5cd0c6710446f6a715694ad415fd47 - plumbing for client config

33b958cd8238a54230dc3e1fdfd977ea4b287508 - a small stack implementation (used by the refactoring around pid tracking / process resource utilization)

6d7401ee22fe651154f864d034217096d2669e17 - remove the ec2info tool

44c33ee3e53c39fbd337b7ac4c9a7ee97ec2ebbb - fixup testing packages in plugins/drivers

1ea6efc46e93cc51340a8f75df3d25f97b8129a4 - build step changes

Documentation and probably some e2e testing will be in a followup PR.

Not sure what's going on with the `test-e2e` workflow (vault compat matrix) but I'll look into it.


client/lib/numalib/topology.go
```diff
@@ -0,0 +1,252 @@
...
+	Cost     uint8
+)
+
+func (hz KHz) MHz() MHz {
+	return MHz(hz / 1000)
+}
+
+func (hz KHz) String() string {
+	return strconv.FormatUint(uint64(hz.MHz()), 10)
+}
```
tgross, line 63: Calling this pointer receiver `hz` feels a little gross :grinning: 

```suggestion
func (khz KHz) MHz() MHz {
	return MHz(khz / 1000)
}

func (khz KHz) String() string {
	return strconv.FormatUint(uint64(khz.MHz()), 10)
}
```

client/fingerprint/cpu.go
```diff
@@ -51,101 +45,154 @@ func (f *CPUFingerprint) Fingerprint(request *FingerprintRequest, response *Fing
...
+	default:
+		response.AddAttribute("cpu.numcores", f.cores(total))
+		f.logger.Debug("detected CPU core count", "cores", total)
 	}
-	f.nodeResources.Cpu.TotalCpuCores = uint16(power + efficiency)
+	f.nodeResources.Cpu.TotalCpuCores = uint16(total)
 }
 
-func (f *CPUFingerprint) setReservableCores(request *FingerprintRequest, response *FingerprintResponse) {
-	reservable := request.Config.ReservableCores
```
tgross, line 116: It looks like we've dropped handling of user-provided configuration here, but we specifically refer to it in the new numa package. Should this still be part of the fingerprint?

client/lib/numalib/detect_darwin.go
```diff
@@ -0,0 +1,74 @@
...
+	for i := 0; i < eCoreCount; i++ {
+		top.insert(nodeID, socketID, nthCore, efficiency, maxSpeed, eCoreSpeed)
+		nthCore++
+	}
+}
+
+func (m *MacOS) scanLegacyX86(top *Topology) {
+	coreCount, _ := unix.SysctlUint32("machdep.cpu.core_count")
+	hz, _ := unix.SysctlUint64("hw.cpufrequency")
+	coreSpeed := KHz(hz / 1_000_000)
```
tgross, line 68: This smells funny... what's the unit of `hw.cpufrequency`?

tgross (approved):
This looks great @shoenig. I really like how this API turned out. It feels like there are a lot of layers but when you start digging you can see exactly how each one has its own responsibilities (ex. splitting out setting up the cgroups vs writing the resources into them)

client/allocrunner/taskrunner/wrangler_hook.go
```diff
@@ -0,0 +1,52 @@
...
+	"context"
+
+	"github.com/hashicorp/go-hclog"
+	ifs "github.com/hashicorp/nomad/client/allocrunner/interfaces"
+	cifs "github.com/hashicorp/nomad/client/interfaces"
+	"github.com/hashicorp/nomad/client/lib/proclib"
+)
+
+const (
+	wranglerHookName = "wrangler"
```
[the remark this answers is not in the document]
shoenig: Yeah good point - I can rename this to something like `"procisolation"` for now, unless someone has something better

client/config/config.go
```diff
@@ -795,7 +794,7 @@ func DefaultConfig() *Config {
 		CNIConfigDir:       "/opt/cni/config",
 		CNIInterfacePrefix: "eth",
 		HostNetworks:       map[string]*structs.ClientHostNetworkConfig{},
-		CgroupParent:       cgutil.GetCgroupParent(""),
+		CgroupParent:       "nomad.slice", // SETH todo
```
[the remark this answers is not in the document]
shoenig: Great callout - filed https://github.com/hashicorp/nomad/issues/18211 to cover that

client/allocrunner/alloc_runner.go
```diff
@@ -450,6 +446,9 @@ func (ar *allocRunner) Restore() error {
 			return err
 		}
 		states[tr.Task().Name] = tr.TaskState()
+
+		// restore process wrangler for task
+		ar.wranglers.Setup(proclib.Task{AllocID: tr.Alloc().ID, Task: tr.Task().Name})
```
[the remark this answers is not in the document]
shoenig: Well the `ar.wranglers` field belongs to the alloc runner which the individual task runner wouldn't have access to. Which, I think is technically unnecessary for the current implementation but in the long term it could also manage child processes associated with the alloc (and not any task).

client/lib/numalib/topology.go
shoenig: whoops! :scream:

client/lib/numalib/detect_darwin.go
shoenig: Filed https://github.com/hashicorp/nomad/issues/18212 for myself to followup

client/fingerprint/cpu.go
shoenig: Filed https://github.com/hashicorp/nomad/issues/18213 - will think about this more

PR #18153, opened 2023-08-04
Add lock operations functions to the variables handler

Juanadelacuesta:
There is already a variables handler in the SDK, this PR adds methods for Acquire, Renew and Release a lock on a variable.


api/variables_test.go
```diff
@@ -307,3 +307,67 @@ func TestVariable_CreateReturnsContent(t *testing.T) {
...
+	t.Run("2 create sv1", func(t *testing.T) {
+		get, _, err := nsv.Create(sv1, nil)
+		must.NoError(t, err)
+		must.NotNil(t, get)
+		must.Positive(t, get.CreateIndex)
+		must.Positive(t, get.CreateTime)
+		must.Positive(t, get.ModifyIndex)
+		must.Positive(t, get.ModifyTime)
+		must.Eq(t, sv1.Items, get.Items)
+		*sv1 = *get
```
jrasell, line 332: Does this assignment, along with the others mean these tests must be run in this order?

api/variables.go
```diff
@@ -380,6 +414,28 @@ type VariableMetadata struct {
...
+	// This is a string version of a time.Duration like "2m".
+	TTL string
+
+	// LockDelay describes a grace period that exists after a lock is lost,
+	// before another client may acquire the lock. This helps protect against
+	// split-brains. This is a string version of a time.Duration like "2m".
+	LockDelay string
+}
+
+func (vl VariableLock) Empty() bool {
```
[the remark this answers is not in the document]
Juanadelacuesta: No, its not. I was using  must.Empty() for the test, but not any more, it can indeed be removed

api/variables.go
```diff
@@ -380,6 +414,28 @@ type VariableMetadata struct {
...
+	return len(vl.ID) == 0
```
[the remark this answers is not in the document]
Juanadelacuesta: Is the same thing :)

api/variables_test.go
Juanadelacuesta: yeas, because it verifies the flow of the 3 functions, they need to be run in order

api/variables.go
```diff
@@ -183,6 +183,44 @@ func (vars *Variables) GetVariableItems(path string, qo *QueryOptions) (Variable
...
 }
 
+// RenewLock renews the lease for the lock on the given variable. It has to be called
+// before the lock's TTL expires or the lock will be automatically released after the
+// delay period.
+func (vars *Variables) RenewLock(v *Variable, qo *WriteOptions) (*VariableMetadata, *WriteMeta, error) {
+	v.Path = cleanPathString(v.Path)
+	var out VariableMetadata
+
+	wm, err := vars.client.put("/v1/var/"+v.Path+"?renew-lock", v, &out, qo)
```
gulducat, line 193: oh!  this was a bug I found during hack week

```suggestion
	wm, err := vars.client.put("/v1/var/"+v.Path+"?lock-renew", v, &out, qo)
```

api/variables.go
```diff
@@ -380,6 +421,24 @@ type VariableMetadata struct {
...
+	Lock *VariableLock `hcl:",lock,optional" json:",omitempty"`
+}
+
+type VariableLock struct {
+	// ID is generated by Nomad to provide a unique caller ID which can be used
+	// for renewals and unlocking.
+	ID string
+
+	// TTL describes the time-to-live of the current lock holder.
+	// This is a string version of a time.Duration like "2m".
```
gulducat, line 435: might be worth adding the defaults values for TTL/LockDelay in these comments?  since for some users this may be their main interaction with the API.  and/or mention that they're optional.

api/variables_test.go
```diff
@@ -307,3 +307,67 @@ func TestVariable_CreateReturnsContent(t *testing.T) {
 	must.NotNil(t, sv1n)
 	must.Eq(t, sv1.Items, sv1n.Items)
 }
+
+func TestVariables_LockRenewRelease(t *testing.T) {
```
gulducat, line 311: looks like this tests only happy paths -- are there any error conditions that we want to test here?  or are we considering that adequately covered by tests on the server side?

relatedly, do we want to include any api-client-side validation for things like "TTL is too small" or any others?  I'm not sure how common that is in our api package.

Juanadelacuesta: According to what I have seen, the API is a thin wrapper, no validations or defaults

api/variables.go
Juanadelacuesta: Because the api is a light client, I dont want to add much logic here, the defaults will be set during validation. (Im trying to keep it in line with everything else)

api/variables_test.go
```diff
@@ -307,3 +307,67 @@ func TestVariable_CreateReturnsContent(t *testing.T) {
...
+		must.NotNil(t, get.Lock)
+
+		*sv1 = *get
+	})
+
+	t.Run("4 renew lock on sv1", func(t *testing.T) {
+		get, _, err := nsv.RenewLock(sv1, nil)
+		must.NoError(t, err)
+		must.NotNil(t, get)
+		must.Eq(t, sv1.ModifyIndex, get.ModifyIndex)
```
gulducat, line 350: just checking my understanding: these should be equal if the lock is renewed, because on the server side the lock is only held in memory (no raft writes), yes?

gulducat (approved):
feel free to add comments on TTL/LockDelay if you'd like, otherwise LGTM!

api/variables_test.go
Juanadelacuesta: yes!

PR #18162, opened 2023-08-07
job_hooks: add implicit constraint when using OCI containers

### LINKED ISSUE #18082: Detect supported architectures for docker containers
### Proposal

I run a Nomad cluster with `amd64`, `arm64`, and `riscv64` nodes. If I try to use a docker image that only supports `amd64`, nomad will sometimes schedule it on one of the non-`amd64` nodes, which causes an `exec format error`. This can be fixed by manually adding a constraint for the CPU architecture.  If possible, I'd like nomad to be able to automatically detect which platforms are supported by reading the manifest, and dynamically add constraints to the job based on that. The docker command allows you to inspect manifests using `docker manifest inspect --verbose <image>`.

### Use-cases

This feature would simplify the process of running docker images on multi-architecture clusters

Elara6331:
This PR adds an implicit constraint for the OS and architecture to a task if it uses a driver that uses OCI container images, such as `docker` and `podman`.

Fixes #18082


nomad/job_endpoint_hooks.go
```diff
@@ -209,11 +226,118 @@ func (jobImpliedConstraints) Mutate(j *structs.Job) (*structs.Job, []error, erro
...
+			RTarget: strings.Join(mapKeys(oses), ","),
+		}
+
+		arch = &structs.Constraint{
+			LTarget: attrCPUArch,
+			Operand: "set_contains_any",
+			RTarget: strings.Join(mapKeys(arches), ","),
+		}
+	} else if desc.MediaType.IsImage() {
+		img, err := remote.Image(ref, remote.WithAuthFromKeychain(authn.DefaultKeychain))
```
tgross, line 304: It looks to me like we're having the Nomad server authenticate to a remote registry during job submission here? Nomad servers are not typically going to be configured with registry credentials at all, because they don't run workloads themselves.

The approach we're taking here would make it mandatory that all users of Nomad who are using Docker/Podman drivers deploy credentials to their Nomad control plane before upgrading, and have a network topology that allows their control plane access to the registry (which isn't necessarily the case). And it would make all Nomad servers dependent on third-party registries at the time of job submission, which I don't think we can accept as a tradeoff.

tgross (changes requested):
Hi @Elara6331! I was doing some Friday afternoon mop-up and noticed we never reviewed this PR. Sorry about that.

Unfortunately I'm also fairly certain we can't accept this PR from a design standpoint. I've left a comment below getting into the details of why. I think we'd be open to other approaches that don't have those downsides though.

PR #18170, opened 2023-08-07
test: deflake job endpoint registration test

tgross:
We've seen test flakiness in the `TestJobEndpoint_Register_NonOverlapping` test, which asserts that we don't try to placed allocations for blocked evals until resources have been actually freed by setting the client status of the previous alloc to complete.

The flaky assertion includes sorting the two allocations by CreateIndex and this appears to be a non-stable sort in the context of the test run, which results in failures that shouldn't exist. There's no reason to sort the allocations instead of just examining them by ID. This changeset does so (and after a few dozen runs locally seems to have worked? He said, hopefully.)


shoenig (approved):
LGTM! nice find

tgross:

Issue #18147, opened 2023-08-03
test flake: `TestDrainer_AllTypes_NoDeadline`


tgross:
There are flaky tests on main for the drainer:

```
=== Failed
=== FAIL: nomad TestDrainer_AllTypes_NoDeadline (12.49s)
    assert.go:11: 
        drainer_int_test.go:870: expected condition to pass within wait context
        ↪ error: wait: timeout exceeded: has drain strategy still set
```

PR #18171, opened 2023-08-07
test: deflake node drain intergration test

tgross:
The `TestDrainer_AllTypes_NoDeadline` test has been flaky. It looks like this might be because the final update of batch allocations to complete is improperly updating the state store directly rather than by RPC. If the service jobs have restarted in the meantime, the `allocClientStateSimulator` will have updated the index on the allocations table and that will prevent the drainer from unblocking (and being marked complete) when the batch jobs are written with an earlier index.

This changeset attempts to fix that by making the update via RPC (as it normally would be in real code).

Fixes: https://github.com/hashicorp/nomad/issues/18147

### DIFF
--- a/nomad/drainer_int_test.go
+++ b/nomad/drainer_int_test.go
@@ -26,9 +26,9 @@ import (
 )
 
 // allocClientStateSimulator simulates the updates in state from the
-// client. allocations that are new on the server get marked with healthy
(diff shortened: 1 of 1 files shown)


shoenig (approved):
LGTM!

tgross:

PR #18179, opened 2023-08-08
e2e: modernize vaultcompat testing

shoenig:
Refactors the `e2e/vaultcompat` tests to use modern patterns. Raises the minimum tested vault version from 0.6.x to 1.1.x (latest is 1.14.x). Now uses a real nomad agent instead of TestAgent.

```
 --- PASS: TestVaultCompat (180.67s)
    --- PASS: TestVaultCompat/testVaultVersions (180.67s)
        --- PASS: TestVaultCompat/testVaultVersions/vault(1.6.7) (9.52s)
        --- PASS: TestVaultCompat/testVaultVersions/vault(1.8.12) (9.67s)
        --- PASS: TestVaultCompat/testVaultVersions/vault(1.2.7) (9.27s)
        --- PASS: TestVaultCompat/testVaultVersions/vault(1.9.10) (11.25s)
        --- PASS: TestVaultCompat/testVaultVersions/vault(1.1.5) (9.74s)
        --- PASS: TestVaultCompat/testVaultVersions/vault(1.11.12) (10.09s)
        --- PASS: TestVaultCompat/testVaultVersions/vault(1.3.10) (9.42s)
        --- PASS: TestVaultCompat/testVaultVersions/vault(1.7.10) (9.63s)
        --- PASS: TestVaultCompat/testVaultVersions/vault(1.14.1) (9.88s)
        --- PASS: TestVaultCompat/testVaultVersions/vault(1.5.9) (10.21s)
        --- PASS: TestVaultCompat/testVaultVersions/vault(1.10.11) (10.27s)
        --- PASS: TestVaultCompat/testVaultVersions/vault(1.13.5) (9.94s)
        --- PASS: TestVaultCompat/testVaultVersions/vault(1.4.7) (10.27s)
        --- PASS: TestVaultCompat/testVaultVersions/vault(1.12.9) (9.82s)
PASS
ok  	github.com/hashicorp/nomad/e2e/vaultcompat	180.686s
```


e2e/vaultcompat/vaultcompat_test.go
```diff
@@ -0,0 +1,260 @@
...
+	"github.com/hashicorp/go-version"
+	nomadapi "github.com/hashicorp/nomad/api"
+	"github.com/hashicorp/nomad/testutil"
+	vaultapi "github.com/hashicorp/vault/api"
+	"github.com/shoenig/test/must"
+	"github.com/shoenig/test/wait"
+)
+
+const (
+	binDir  = "vault-bins"
```
gulducat, line 28: bonus for consideration: might be neat to cache the whole `binDir` with [`actions/cache`](https://github.com/actions/cache)

e2e/vaultcompat/vaultcompat_test.go
```diff
@@ -0,0 +1,260 @@
...
+		downloadVaultBuild(t, b)
+		testVaultBuild(t, b)
+		return true
+	})
+}
+
+func testVaultBuild(t *testing.T, b build) {
+	t.Run("vault("+b.Version+")", func(t *testing.T) {
+		vStop, vc := startVault(t, b)
+		defer vStop()
```
gulducat, line 51: I can't resist test code golf: you could `t.Cleanup(vlt.Stop)` down in `startVault()` (ditto for `startNomad()`)

e2e/vaultcompat/vaultcompat_test.go
```diff
@@ -0,0 +1,260 @@
...
+		defer nStop()
+		runCatJob(t, nc)
+
+		// give nomad and vault time to stop
+		defer func() { time.Sleep(5 * time.Second) }()
+	})
+}
+
+func runCatJob(t *testing.T, nc *nomadapi.Client) {
+	b, err := os.ReadFile("input/cat.hcl")
```
gulducat, line 64: any particular reason these aren't marked as `t.Helper()` ?

e2e/vaultcompat/vaultcompat_test.go
```diff
@@ -0,0 +1,260 @@
...
+			if s := allocs[0].ClientStatus; s != "complete" {
+				return fmt.Errorf("expected alloc status complete, got %s", s)
+			}
+			return nil
+		}),
+		wait.Timeout(20*time.Second),
+		wait.Gap(1*time.Second),
+	))
+
+	t.Log("success running cat job")
```
gulducat, line 92: this all hinges on `cat "${NOMAD_SECRETS_DIR}/vault_token"` exiting `0` -- not `1` "no such file" or other -- to result in "complete" ClientStatus instead of "failed" or perpetually-"pending" right?  and that exercises a bulk of the machinery to produce a vault token for the task.

is there any additional value in checking the (shape of) contents of that file, or the env var?  or confirming that the token can be used with an actual `vault token lookup` or similar?

e2e/vaultcompat/vaultcompat_test.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+}
+
+type tracker map[int]*set.TreeSet[build, set.Compare[build]]
+
+func (t tracker) add(v *version.Version, b build) {
+	y := v.Segments()[1]
+	if _, exists := t[y]; !exists {
+		cmp := func(g, h build) int { return g.compare(h) }
+		t[y] = set.NewTreeSet[build, set.Compare[build]](cmp)
```
gulducat, line 222: I had to churn some extra brain cells to parse this, compared to everything else in here being very readable, but the end result of `track.add()` and then `tree.Max()` down below (if I'm reading this all right) is quite nice. :+1:

e2e/vaultcompat/vaultcompat_test.go
```diff
@@ -0,0 +1,260 @@
...
+		Period:   "72h",
+		NoParent: true,
+	})
+	must.NoError(t, err, must.Sprint("failed to create vault token"))
+	must.NotNil(t, secret)
+	must.NotNil(t, secret.Auth)
+}
+
+func startNomad(t *testing.T, vc *vaultapi.Client) (func(), *nomadapi.Client) {
+	ts := testutil.NewTestServer(t, func(c *testutil.TestServerConfig) {
```
gulducat, line 126: just to log my understanding here:

`make integration-test` depends on `make dev`, so it runs that first, then this testutil func does `discover.NomadExecutable()` to find the binary that was just built, and runs that as "a real nomad agent" process.

yes?

gulducat (approved):
LGTM!  just some questions and maybe-bonus suggestions for consideration.

GNUmakefile
```diff
@@ -312,13 +312,12 @@ e2e-test: dev ## Run the Nomad e2e test suite
 .PHONY: integration-test
 integration-test: dev ## Run Nomad integration tests
 	@echo "==> Running Nomad integration test suites:"
-	go test \
-		$(if $(ENABLE_RACE),-race) $(if $(VERBOSE),-v) \
```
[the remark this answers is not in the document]
shoenig: Previously when using `TestAgent` there was no hope of using `-race` so I had just removed it. Since switching to running a real agent which gets compiled beforehand... I think it should be fine adding it back? Though it'd really only be confirming the test case code has no races. But it shouldn't hurt.

e2e/vaultcompat/vaultcompat_test.go
shoenig: I was worried if the agents log during shutdown they will cause the test to fail, though I didn't actually try it ...

e2e/vaultcompat/vaultcompat_test.go
shoenig: `t.Helper` is really intended for generalized library functions who's presence in a stack trace on test failure would be of no benefit. The helper functions in this file are all relevant to the test - e.g. they all make assertions themselves and we wouldn't want to meddle with removing them from failure dumps.

e2e/vaultcompat/vaultcompat_test.go
shoenig: You got it!

e2e/vaultcompat/vaultcompat_test.go
shoenig: Yeah I had the same thought - for now this test just makes the same assertion as the code it replaces (that `cat`ing the token succeeds).

Let me make a followup ticket with your idea, it's probably worth doing but it's a good chunk more work than just cleaning things up. Using the Nomad API to gather logs from a specific task requires a lot of boilerplate.

gulducat: fwiw, a `vault token lookup` should exit `0` if there's a valid `VAULT_TOKEN` (and vault is reachable), and non-0 (`2`, apparently) if the token is invalid.  so adding that one more command to the "cat" job might be relatively low hanging fruit without having to check any log output.

PR #18184, opened 2023-08-09
update go1.21

shoenig:
This PR updates the build system to use go1.21. In addition,

- update golangci-lint
- eliminate the Min/Max helpers for the now builtin min/max
- swap out depguard and use semgrep instead (was failing with go1.21, and has been troublesome in the past too)
- fixup a TLS error string check that seemingly changed in go1.21


tgross (approved):
LGTM

IamTheFij:
I just want to bring some awareness to a downside of staying cutting edge here.

I think it's great for the application part of Nomad, but it makes it difficult for other projects that are trying to use `nomad/api`. I have been looking to move tasks of mine over to use the Nomad Task Socket, which requires the API client included in 1.7. It included a moderate amount of updates to get Diun capable, and I just started looking at upgrading Prometheus, but it seems that will be even more significant.

I don't know if there's really a way to work around this (maybe pulling API into it's own package based on a different Go API), but I figured I'd at least shed some light on it.

Thanks!

shoenig:
Hi @IamTheFij the `/api` submodule should still only require go1.20 per its [go.mod](https://github.com/hashicorp/nomad/blob/main/api/go.mod#L3) file; you shouldn't need go1.21 if that's the only module you have a dependency on. (let us know if that somehow isn't the case!)

IamTheFij:
@shoenig, so that's true per the go.mod file, but it's impossibe to build on go1.20 and results in the following error:

```
/go/pkg/mod/github.com/hashicorp/nomad/api@v0.0.0-20231213195942-64e3dca9274b/error_unexpected_response.go:11:2: package slices is not in GOROOT (/usr/local/go/src/slices)
```

Something that will also make it easier would be if the `/api` submodule had a release. I created an issue to track that as well. #19613 Perhaps that could be expanded to include some dedicated testing of that module as well.

PR #18185, opened 2023-08-09
[hooks] Fix pre-push hook

angrycub:
In #17560 I broke the pre-push hook. This PR should fix it and adds some more possibilities to the mix.

Since we use the presence of the sentinel file to determine if we're pushing the enterprise version, rather than checking the remote every time the hook runs, I moved all of the enterprise checking into a conditional based on that test.

The regex is nasty, because there are a lot of optional components to legit remote URLs 😢

### DIFF
--- a/dev/hooks/pre-push
+++ b/dev/hooks/pre-push
@@ -3,28 +3,31 @@
 set -euo pipefail
 
 fail () {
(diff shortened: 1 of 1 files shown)


shoenig (approved):
LGTM! Tested and works locally

PR #18192, opened 2023-08-14
Update linux package license to BUSL-1.1

sarahethompson:
This updates the linux package license to BUSL-1.1.
This PR should be merged to main so that the change is included in the next minor release that is cut from main.
License changes are not required on release branches (ongoing patch releases of current minor releases).

Also update copywrite.hcl to exclude MPL licensed subdirs.

### DIFF
--- a/.copywrite.hcl
+++ b/.copywrite.hcl
@@ -7,6 +7,9 @@ project {
   header_ignore = [
     "command/asset/*.hcl",
     "command/agent/bindata_assetfs.go",
+    "api/**",
+    "drivers/shared/**",
+    "plugins/**",
     // Enterprise files do not fall under the open source licensing. OSS-ENT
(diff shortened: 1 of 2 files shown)


shoenig (approved):

.copywrite.hcl
```diff
@@ -7,6 +7,9 @@ project {
   header_ignore = [
     "command/asset/*.hcl",
     "command/agent/bindata_assetfs.go",
+    "api/**",
+    "drivers/shared/**",
+    "plugins/**",
```
tgross, line 12: Is the copyright bot unable to distinguish between licenses it's supposed to use? We'll definitely be adding files to these in the future, which will need the MPL2 license.

tgross (approved):

PR #18195, opened 2023-08-14
build: use go-modtool to format go.mod file

shoenig:
This is the first of two uses of [go-modtool](https://github.com/shoenig/go-modtool) - here we use the `fmt` subcommand to consistently format the `go.mod` file of the Nomad project (and will do the same in the ent repo). Once the formatting is in place in both repos, we can use the `merge` subcommand during `oss-merge.sh` fallback merge resolution to (hopefully) avoid dealing with trivial `go.mod` merge conflicts. I suspect the same pattern could be used for backports, but I hadn't thought that far ahead.

### DIFF
--- a/.github/workflows/test-core.yaml
+++ b/.github/workflows/test-core.yaml
@@ -63,8 +63,8 @@ jobs:
       - uses: hashicorp/setup-golang@v1
       - name: Get Go modules
         run: |
-          make tidy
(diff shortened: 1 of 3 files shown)


tgross (approved):
LGTM!

jrasell (approved):
LGTM!

PR #18198, opened 2023-08-14
[ui] Trim variable path before save

philrenaud:
We previously had a trimPath helper for making sure that, if our paths ended/started in `/` it would be removed. This does the same for basic whitespace using [JS's string trim](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim)


angrycub (approved):
Nice!

Issue #18199, opened 2023-08-14
Recurring device driver stats failure causes zero backoff (and a lot of wasted CPU).


stswidwinski:
### Nomad version
tip.

### Operating system and Environment details
Linux.

### Issue

Repeated failure to obtain device stats causes the backoff to become `0` and burn CPU. We've noticed the CPU being entirely pegged due to this logic. This is because the backoff logic:

https://github.com/hashicorp/nomad/blob/main/client/devicemanager/instance.go#L497-L504

Can easily overflow. Here are the logs of repeated failures and the computed backoff:

```
[ERROR] client.device_mgr: stats returned an error: plugin=PLUGIN error=<nil> retry=5s
[ERROR] client.device_mgr: stats returned an error: plugin=PLUGIN error=<nil> retry=20s
[ERROR] client.device_mgr: stats returned an error: plugin=PLUGIN error=<nil> retry=1m20s
[ERROR] client.device_mgr: stats returned an error: plugin=PLUGIN error=<nil> retry=5m20s
[ERROR] client.device_mgr: stats returned an error: plugin=PLUGIN error=<nil> retry=21m20s
[ERROR] client.device_mgr: stats returned an error: plugin=PLUGIN error=<nil> retry=30m0s
[ERROR] client.device_mgr: stats returned an error: plugin=PLUGIN error=<nil> retry=30m0s
[ERROR] client.device_mgr: stats returned an error: plugin=PLUGIN error=<nil> retry=30m0s
[ERROR] client.device_mgr: stats returned an error: plugin=PLUGIN error=<nil> retry=30m0s
[ERROR] client.device_mgr: stats returned an error: plugin=PLUGIN error=<nil> retry=30m0s
[ERROR] client.device_mgr: stats returned an error: plugin=PLUGIN error=<nil> retry=30m0s
[ERROR] client.device_mgr: stats returned an error: plugin=PLUGIN error=<nil> retry=30m0s
[ERROR] client.device_mgr: stats returned an error: plugin=PLUGIN error=<nil> retry=30m0s
[ERROR] client.device_mgr: stats returned an error: plugin=PLUGIN error=<nil> retry=30m0s
[ERROR] client.device_mgr: stats returned an error: plugin=PLUGIN error=<nil> retry=30m0s
[ERROR] client.device_mgr: stats returned an error: plugin=PLUGIN error=<nil> retry=30m0s
[...]

PR #18200, opened 2023-08-14
fix multiple overflow errors in exponential backoff

tgross:
We use capped exponential backoff in several places in the code when handling failures. The code we've copy-and-pasted all over has a check to see if the backoff is greater than the limit, but this check happens after the bitshift and we always increment the number of attempts. This causes an overflow with a fairly small number of failures (ex. at one place I tested it occurs after only 24 iterations), resulting in a negative backoff which then never recovers. The backoff becomes a tight loop consuming resources and/or DoS'ing a Nomad RPC handler or an external API such as Vault. Note this doesn't occur in places where we cap the number of iterations so the loop breaks (usually to return an error), so long as the number of iterations is reasonable.

Introduce a check on the cap before the bitshift to avoid overflow in all places this can occur.

Fixes: #18199

### DIFF
--- a/drivers/docker/driver.go
+++ b/drivers/docker/driver.go
@@ -527,8 +527,8 @@ CREATE:
 		}
 
 		if attempted < 5 {
-			backoff = helper.Backoff(50*time.Millisecond, time.Minute, attempted)
 			attempted++
+			backoff = helper.Backoff(50*time.Millisecond, time.Minute, attempted)
 			time.Sleep(backoff)
 			goto CREATE
 		}
@@ -539,6 +539,7 @@ CREATE:
 		return nil, nstructs.NewRecoverableError(createErr, true)
 	} else if isDockerTransientError(createErr) && attempted < 5 {
 		attempted++
+		backoff = helper.Backoff(50*time.Millisecond, time.Minute, attempted)
 		time.Sleep(backoff)
 		goto CREATE
 	}
--- a/drivers/docker/driver.go
+++ b/drivers/docker/driver.go
@@ -533,9 +533,6 @@ CREATE:
 			goto CREATE
 		}
 
-		if attempted < 5 {
-			attempted++
-		}
 	} else if strings.Contains(strings.ToLower(createErr.Error()), "no such image") {
 		// There is still a very small chance this is possible even with the
 		// coordinator so retry.
--- a/drivers/docker/driver.go
+++ b/drivers/docker/driver.go
@@ -572,7 +572,8 @@ START:
 	return recoverableErrTimeouts(startErr)
 }
 
-// nextBackoff returns appropriate docker backoff durations after attempted attempts.
+// nextBackoff returns appropriate docker backoff durations after attempted
+// attempts. Note that the caller must cap attempted so that it doesn't overflow
 func nextBackoff(attempted int) time.Duration {
 	// attempts in 200ms, 800ms, 3.2s, 12.8s, 51.2s
 	// TODO: add randomization factor and extract to a helper
--- a/drivers/docker/driver.go
+++ b/drivers/docker/driver.go
@@ -476,7 +476,9 @@ type createContainerClient interface {
 func (d *Driver) createContainer(client createContainerClient, config docker.CreateContainerOptions,
 	image string) (*docker.Container, error) {
 	// Create a container
-	attempted := 0
+	var attempted uint64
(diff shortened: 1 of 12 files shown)


nomad/state/state_store.go
```diff
@@ -279,9 +280,11 @@ func (s *StateStore) SnapshotMinIndex(ctx context.Context, index uint64) (*State
...
 		} else {
 			// Subsequent retry, reset timer
-			deadline := 1 << (2 * retries) * backoffBase
-			if deadline > backoffLimit {
-				deadline = backoffLimit
+			if deadline < backoffLimit {
+				deadline = 1 << (2 * retries) * backoffBase
+				if deadline > backoffLimit {
+					deadline = backoffLimit
+				}
```
gulducat, line 287: care to put `retries++` from line 276 down here too?  I think the end result is the same with the new extra `deadline < limit` check guarding the `2 * retries`, but I like how the others have their `retries`-equivalent var right after where it's used.  even if it's not used in the math, I think this one would continue incrementing the var arbitrarily-long which just tastes funny to me.

nomad/worker.go
```diff
@@ -874,12 +876,17 @@ func (w *Worker) shouldResubmit(err error) bool {
...
 func (w *Worker) backoffErr(base, limit time.Duration) bool {
 	w.setWorkloadStatus(WorkloadBackoff)
-	backoff := (1 << (2 * w.failures)) * base
-	if backoff > limit {
-		backoff = limit
-	} else {
-		w.failures++
+
+	backoff := w.failureBackoff
+	if w.failureBackoff < limit {
```
gulducat, line 881: isn't this always true, as long as `limit` is greater than 0?  since nothing is setting `w.failureBackoff` to something different from its default zero value.

gulducat (approved):
I have a couple bits (heh) for consideration, which may be a result of my eyes glazing over reading these almost-but-not-quite-identical implementations.

nomad/state/state_store.go
tgross: Makes sense.

nomad/worker.go
tgross: D'oh, I didn't set the `w.failureBackoff` below. That's just a bug. Will fix.

stswidwinski:
I thought I'd propose a little bit more structured approach: https://github.com/hashicorp/nomad/pull/18201. Perhaps we can use that across the callsites to avoid future issues resulting from repeated logic?

tgross:
I've pulled in the proposed changes from #18201

drivers/docker/driver.go
```diff
@@ -572,13 +578,6 @@ START:
 	return recoverableErrTimeouts(startErr)
 }
 
-// nextBackoff returns appropriate docker backoff durations after attempted attempts.
-func nextBackoff(attempted int) time.Duration {
```
gulducat, line 576: potential lazy route: maybe some of my "is it the same?" questions above could be resolved by keeping this function, but having _it_ call the new helper?

gulducat (approved):
Yay for reusable helper!

I have two concerns in the docker driver, one small and one large.
Other than that, LGTM!

drivers/docker/driver.go
```diff
@@ -525,17 +527,19 @@ CREATE:
 		}
 
 		if attempted < 5 {
+			backoff = helper.Backoff(50*time.Millisecond, time.Minute, attempted)
 			attempted++
-			time.Sleep(nextBackoff(attempted))
+			time.Sleep(backoff)
```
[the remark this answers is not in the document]
tgross: It probably makes difference of exactly 1 extra 50ms delay. Worth fixing though to retain the existing behavior.

drivers/docker/driver.go
```diff
@@ -525,17 +527,19 @@ CREATE:
...
 		}
+
 	} else if strings.Contains(strings.ToLower(createErr.Error()), "no such image") {
 		// There is still a very small chance this is possible even with the
 		// coordinator so retry.
 		return nil, nstructs.NewRecoverableError(createErr, true)
 	} else if isDockerTransientError(createErr) && attempted < 5 {
 		attempted++
-		time.Sleep(nextBackoff(attempted))
+		time.Sleep(backoff)
```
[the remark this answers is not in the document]
tgross: Oops, good catch.

tgross:

PR #18201, opened 2023-08-14
Propose a more structured approach to computation of backoff.

stswidwinski:
Proposal for a more structured and tested backoff logic.

### DIFF
--- a/helper/backoff.go
+++ b/helper/backoff.go
@@ -1,0 +1,26 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package helper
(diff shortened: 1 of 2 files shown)


stswidwinski:
Proposal to be merged against https://github.com/hashicorp/nomad/pull/18200/files rather than to be considered in isolation

shoenig:
Nice @stswidwinski! I was going to suggest doing something similar.

tgross:
Yeah I like this. I'll pull this into the other PR and integrate it.

tgross:
I've pulled this change into https://github.com/hashicorp/nomad/pull/18200

PR #18202, opened 2023-08-15
Fix the logic for the display of a newly created lock

Juanadelacuesta:
On top of fixing the lock owner check, this PR also introduces a change that allows for the ACL object that is created when the call is made using a workload identity token to be used outside of the `authorize` function.

This code will need to be revisited after the nil ACLs is removed


tgross (approved):
LGTM

jrasell (approved):
LGTM, thanks!

PR #18209, opened 2023-08-15
admin: rename _oss files to _ce

jrasell:


schmichael (approved):
Before merging can we quick check and see if there's prior art with other HashiCorp products? I'd hate for us all to do something a bit different and make a jarring transition even worse (albeit just slightly).

schmichael:
Sounds like the Nomad team might be blazing this trail. I say go for it, and we can share our decision internally.

As discussed internally "Community Edition" isn't necessarily the most precise name we could choose, but I think it gets across our intent: the code is *available with restrictions beyond OSS.*

PR #18210, opened 2023-08-15
[ui] Use ]] and [[ to paginate index pages

philrenaud:
Adds a `[[` and `]]` key command to the `list-pager` prev/next page elements. These are included on, for example, the Jobs index page, clients index page, jobs.job.allocatinos page, etc.


gulducat (approved):
I ad-hoc tested locally; pretty neat!

PR #18214, opened 2023-08-15
followup to numa/cgroups refactor 

shoenig:
- lang: note that Stack is not concurrency-safe
- client: use more descriptive name for wrangler hook in logs
- numalib: use correct name for receiver parameter


tgross (approved):
LGTM!

PR #18215, opened 2023-08-15
test-e2e: no need to run vaultcomat tests as root

shoenig:
6747ef8803 fixes the Nomad client to support using the raw_exec
driver while running as a non-root user. Remove the use of sudo
in the test-e2e workflow for running integration (vaultcompat)
tests.

Needed to plumb `cpu_total_compute` client config for the agent
runner, since Nomad cannot detect the cpu specs on the github public
runner without root.


gulducat (approved):
hooray for de-escalating perms!

PR #18224, opened 2023-08-16
config: parsing support for multiple Vault clusters in agent config

tgross:
Add the plumbing we need to accept multiple Vault clusters in Nomad agent configuration, to support upcoming Nomad Enterprise features. The `vault` blocks are differentiated by a new `name` field, and if the `name` is omitted it becomes the "default" Vault configuration. All blocks with the same name are merged together, as with the existing behavior.

Unfortunately we're still using HCL1 for parsing configuration and the `Decode` method doesn't parse multiple blocks differentiated only by a field name without a label. So we've had to add an extra parsing pass, similar to what we've done for HCL1 jobspecs.

For now, all existing consumers will use the "default" Vault configuration, so there's no user-facing behavior change in this changeset other than the contents of the agent self API.

### DIFF
--- a/command/agent/command.go
+++ b/command/agent/command.go
@@ -73,9 +73,10 @@ func (c *Command) readConfig() *Config {
 		Server: &ServerConfig{
 			ServerJoin: &ServerJoin{},
 		},
-		Vault: &config.VaultConfig{},
-		ACL:   &ACLConfig{},
-		Audit: &config.AuditConfig{},
+		Vault:  &config.VaultConfig{},
+		Vaults: map[string]*config.VaultConfig{},
+		ACL:    &ACLConfig{},
+		Audit:  &config.AuditConfig{},
 	}
 
 	flags := flag.NewFlagSet("agent", flag.ContinueOnError)
--- a/command/agent/command.go
+++ b/command/agent/command.go
@@ -73,11 +73,11 @@ func (c *Command) readConfig() *Config {
 		Server: &ServerConfig{
 			ServerJoin: &ServerJoin{},
 		},
-		Vault:  &config.VaultConfig{},
-		Vaults: map[string]*config.VaultConfig{},
-		ACL:    &ACLConfig{},
-		Audit:  &config.AuditConfig{},
+		Vault: &config.VaultConfig{},
+		ACL:   &ACLConfig{},
+		Audit: &config.AuditConfig{},
 	}
+	cmdConfig.Vaults = map[string]*config.VaultConfig{"default": cmdConfig.Vault}
 
 	flags := flag.NewFlagSet("agent", flag.ContinueOnError)
 	flags.Usage = func() { c.Ui.Error(c.Help()) }
--- a/command/agent/config_parse_test.go
+++ b/command/agent/config_parse_test.go
@@ -959,6 +959,7 @@ func TestConfig_MultipleVault(t *testing.T) {
 
 	must.MapLen(t, 1, cfg.Vaults)
 	must.Equal(t, cfg.Vault, cfg.Vaults["default"])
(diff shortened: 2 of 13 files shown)


command/agent/config_parse_test.go
```diff
@@ -904,3 +946,50 @@ func permutations(arr []string) [][]string {
 	helper(arr, len(arr))
 	return res
 }
+
+func TestConfig_MultipleVault(t *testing.T) {
```
lgfa29, line 950: Maybe I missed, but would it be worth to have tests where the default `vault` block doesn't have a name and that we test not only that `Vault` is equal to `Vaults["default"]` but that they actually point to the same object?

nomad/structs/config/vault.go
```diff
@@ -109,6 +111,9 @@ func (c *VaultConfig) AllowsUnauthenticated() bool {
 func (c *VaultConfig) Merge(b *VaultConfig) *VaultConfig {
 	result := *c
 
+	if b.Name != "" {
+		c.Name = b.Name
+	}
```
lgfa29, line 116: Just out of curiosity, are there scenarios where we would merge `VaultConfig` with different names? Sounds like maybe a programmer error?

lgfa29 (approved):
Some minor comments, but LGTM!

command/agent/config_parse_test.go
tgross: Good call, we could test that by removing the `name` field from the default Vault in `command/agent/testdata/extra-vault.hcl`. And we can check that the pointers are the same.

command/agent/command.go
```diff
@@ -73,9 +73,10 @@ func (c *Command) readConfig() *Config {
 		Server: &ServerConfig{
 			ServerJoin: &ServerJoin{},
 		},
-		Vault: &config.VaultConfig{},
-		ACL:   &ACLConfig{},
-		Audit: &config.AuditConfig{},
+		Vault:  &config.VaultConfig{},
+		Vaults: map[string]*config.VaultConfig{},
```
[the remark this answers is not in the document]
tgross: Yes, but that also makes me realize that if someone passes the flags they're being set on that `Vault` (which gets overwritten) and not the `Vaults`. So that's a bug to fix.

nomad/structs/config/vault.go
tgross: There's probably a testing scenario where you have partial configs that you're merging. But that doesn't sound like a useful production code use case at all.

lgfa29: Thinking about this a bit more, there are probably scenarios where you would want to create a new `VaultConfig` from the `DefaultVaultConfig`, so overwriting the name does make sense.

command/agent/command.go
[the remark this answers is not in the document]
lgfa29: I haven't checked, but the [`vault` configuration reload](https://developer.hashicorp.com/nomad/docs/configuration/vault#vault-configuration-reloads) may need to be adjusted as well:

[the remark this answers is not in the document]
tgross: Good looking out. The way the reload works currently is that we call `SetConfig` on the Vault client with the new default Vault config. So until we have those multiple Vault clients instantiated (which will come in a later PR), there's nothing to do for the reload. I'll revisit this then.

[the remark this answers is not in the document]
tgross: Fixed in 3dcd0a7913d341bb6bbde1018202bf274c9f3173

command/agent/config_parse_test.go
tgross: Fixed in 3dcd0a7913d341bb6bbde1018202bf274c9f3173

PR #18225, opened 2023-08-16
docs: fix job dispatch documentation

### LINKED ISSUE #18173: Add idPrefixTemplate parameter to job dispatch API documentation
### Proposal
Please add the `idPrefixTemplate` parameter to the API docs at https://developer.hashicorp.com/nomad/api-docs/jobs#dispatch-job. It works fine but you need to read source code to figure out how to use it.

lgfa29:
- Add missing `IdPrefixTemplate` to the API docs.
- Fix casing of `IdempotencyToken` in the API docs.
- Remove `template` from the `-id-prefix-template` flag description. I found it to be confusing as, as far as I can tell, the value isn't actually templated.

Closes #18173


tgross (approved):
:+1:

PR #18234, opened 2023-08-17
csi: prevent panic on volume delete

lgfa29:
When a CSI volume is deleted when its plugin is not running, the function `volAndPluginLookup` returns a `nil` plugin value resulting a panic in the request handler.

Note to reviewers: the other place where `volAndPluginLookup` is used already checks for a `nil` plugin.
https://github.com/hashicorp/nomad/blob/e21ab7d948b96ad7c4a944abde8e703040e6eae7/nomad/csi_endpoint.go#L513-L518

### DIFF
--- a/.changelog/18234.txt
+++ b/.changelog/18234.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+csi: fixed a bug that could case a panic when deleting volumes
+```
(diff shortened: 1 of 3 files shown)


tgross (approved):
LGTM!

PR #18236, opened 2023-08-17
demo: update image for hostpath CSI plugin

lgfa29:
The main reason to update is to support ARM environments, but always good to keep it updated as well.


jrasell (approved):
LGTM!

PR #18237, opened 2023-08-17
nomad: remove custom max func and use Go 1.21.0 builtin

jrasell:


shoenig (approved):
LGTM!

lgfa29 (approved):
I was confused for a second because this function does seem to be used in several places, only to realize that the new built-in function is called `max` as well and my `gopls` was severely outdated 😅

PR #18238, opened 2023-08-17
drivers/docker: fix a hostConfigMemorySwappiness panic

pkazmierczak:
`cgroupslib.MaybeDisableMemorySwappiness` returned an incorrect type, and was
incorrectly typecast to `int64` causing a panic on non-linux and non-windows hosts.

### DIFF
--- a/drivers/docker/driver.go
+++ b/drivers/docker/driver.go
@@ -23,6 +23,9 @@ import (
 	multierror "github.com/hashicorp/go-multierror"
 	plugin "github.com/hashicorp/go-plugin"
 	"github.com/hashicorp/go-set"
+	"github.com/ryanuber/go-glob"
+	"golang.org/x/exp/slices"
+
 	"github.com/hashicorp/nomad/client/lib/cgroupslib"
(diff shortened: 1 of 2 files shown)


jrasell (approved):
LGTM and tested locally on macOS M2.

Juanadelacuesta (approved):

drivers/docker/driver.go
```diff
@@ -23,6 +23,9 @@ import (
 	multierror "github.com/hashicorp/go-multierror"
 	plugin "github.com/hashicorp/go-plugin"
 	"github.com/hashicorp/go-set"
+	"github.com/ryanuber/go-glob"
+	"golang.org/x/exp/slices"
+
```
shoenig, line 28: extra newline in these imports?

shoenig (approved):

drivers/docker/driver.go
pkazmierczak: eh I think it's my goland goimports setting. I'll revert this.

PR #18239, opened 2023-08-17
jobspec: add nomad_service and identity blocks

pkazmierczak:
This PR introduces updates to the jobspec required for workload identity support for services. 

### DIFF
--- a/nomad/structs/services.go
+++ b/nomad/structs/services.go
@@ -21,12 +21,13 @@ import (
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/go-set"
-	"github.com/hashicorp/nomad/helper"
-	"github.com/hashicorp/nomad/helper/args"
-	"github.com/hashicorp/nomad/helper/pointer"
 	"github.com/mitchellh/copystructure"
 	"golang.org/x/exp/maps"
 	"golang.org/x/exp/slices"
+
+	"github.com/hashicorp/nomad/helper"
+	"github.com/hashicorp/nomad/helper/args"
+	"github.com/hashicorp/nomad/helper/pointer"
 )
 
 const (
@@ -609,6 +610,11 @@ type Service struct {
 	// either ServiceProviderConsul or ServiceProviderNomad and defaults to the former when
 	// left empty by the operator.
 	Provider string
+
+	// Identity is a field populated automatically by the job mutating hook.
+	// Its name will be `consul-service/${service_name}`, and its contents will
+	// match the server's `consul.service_identity` configuration block.
+	Identity *WorkloadIdentity
 }
 
 // Copy the block recursively. Returns nil if nil.
@@ -634,6 +640,8 @@ func (s *Service) Copy() *Service {
 	ns.Meta = maps.Clone(s.Meta)
 	ns.CanaryMeta = maps.Clone(s.CanaryMeta)
 	ns.TaggedAddresses = maps.Clone(s.TaggedAddresses)
+
+	ns.Identity = s.Identity.Copy()
 
 	return ns
 }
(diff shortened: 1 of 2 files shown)


api/tasks.go
```diff
@@ -1151,8 +1151,9 @@ func (t *TaskCSIPluginConfig) Canonicalize() {
...
 type WorkloadIdentity struct {
-	Name     string   `hcl:"name,optional"`
-	Audience []string `mapstructure:"aud" hcl:"aud,optional"`
-	Env      bool     `hcl:"env,optional"`
-	File     bool     `hcl:"file,optional"`
+	Name         string   `hcl:"name,optional"`
+	Audience     []string `mapstructure:"aud" hcl:"aud,optional"`
+	Env          bool     `hcl:"env,optional"`
+	File         bool     `hcl:"file,optional"`
+	NomadService string   `hcl:"nomad_service,optional"`
```
tgross, line 1158: We didn't get into this detail in the RFC but I wonder if calling this field name `NomadService` is potentially confusing for API consumers? At least for 1.7.0 this is always going to be a Consul service. We're going to write a claim `nomad_service` in the JWT but that's so that we're namespacing the JWT claim properly for Consul.

```suggestion
	ServiceName string   `hcl:"service_name,optional"`
```

nomad/structs/services.go
```diff
@@ -879,6 +905,22 @@ func hashConnect(h hash.Hash, connect *ConsulConnect) {
 	}
 }
 
+func hashIdentity(h hash.Hash, identity *WorkloadIdentity) {
```
tgross, line 908: Nice job at avoiding messing with any of the existing hashes here

tgross (approved):
LGTM!

I've left one suggestion about the field name but that's not a blocker for me if you feel otherwise.

nomad/structs/services.go
```diff
@@ -808,6 +819,20 @@ func (s *Service) validateNomadService(mErr *multierror.Error) {
...
+func (s *Service) validateIdentity() error {
+	if s.Identity != nil {
+		if s.Identity.Name != "" {
+			return fmt.Errorf("Service identity name must not be explicitly set in the jobspec")
+		}
+		if len(s.Identity.Audience) == 0 {
+			return fmt.Errorf("Service identity must provide at least one target aud value")
+		}
+	}
+	return nil
```
lgfa29, line 833:
```suggestion
	if s.Identity == nil {
		return nil
	}
	
	if s.Identity.Name != "" {
		return fmt.Errorf("Service identity name must not be explicitly set in the jobspec")
	}
	if len(s.Identity.Audience) == 0 {
		return fmt.Errorf("Service identity must provide at least one target aud value")
	}

	return nil
```

Dedent all the things o/

nomad/structs/services.go
```diff
@@ -808,6 +819,20 @@ func (s *Service) validateNomadService(mErr *multierror.Error) {
 	}
 }
 
+// validateIdentity performs validation on workload identity field populated by
+// the job mutating hook
+func (s *Service) validateIdentity() error {
+	if s.Identity != nil {
+		if s.Identity.Name != "" {
+			return fmt.Errorf("Service identity name must not be explicitly set in the jobspec")
```
lgfa29, line 827: Hum....this is an interesting scenario, because there's only one point in which this is true: job registration.

After the job is registered, the _opposite_ is true: a service WI _must_ have a name (`consul-server/...`).

I wonder if it would be better to remove this check and only validate an empty name in the job registration hook?

lgfa29 (approved):
Left some thoughts, but changes are looking good

api/tasks.go
tgross: That can work for the `api` package but in `structs` we'd end up having to add extra handling for the signing code to take multiple types.

pkazmierczak: @tgross I'll rename the field to `ServiceName` as you suggest but I'll keep the `WorkloadIdentity` struct.

nomad/structs/services.go
pkazmierczak: Couldn't we have both? I think it's still worth it to check that there's no pre-set WI service name upon registration, and then we could have another check in the job registration hook.

PR #18253, opened 2023-08-17
fingerprint: add support for fingerprinting multiple Vault clusters

tgross:
Add fingerprinting we'll need to accept multiple Vault clusters in upcoming Nomad Enterprise features. The fingerprinter will create a map of Vault clients by cluster name. In Nomad CE, all but the default cluster will be ignored and there will be no visible behavior change.

_Don't merge until we've got a ENT repo PR ready for the `vault_ent.go` file._

### DIFF
--- a/client/fingerprint/vault.go
+++ b/client/fingerprint/vault.go
@@ -88,7 +88,7 @@ func (f *VaultFingerprint) fingerprintImpl(cfg *config.VaultConfig, resp *Finger
 	if err != nil {
 		// Print a message indicating that Vault is not available anymore
 		if state.isAvailable {
-			logger.Info("Vault is unavailable", "cluster", cfg.Name)
+			logger.Info("Vault is unavailable")
 		}
 		state.isAvailable = false
 		state.nextCheck = time.Time{} // always check on next interval
@@ -110,14 +110,14 @@ func (f *VaultFingerprint) fingerprintImpl(cfg *config.VaultConfig, resp *Finger
 	// If Vault was previously unavailable print a message to indicate the Agent
 	// is available now
 	if !state.isAvailable {
-		logger.Info("Vault is available", "cluster", cfg.Name)
+		logger.Info("Vault is available")
 	}
 
 	// Widen the minimum window to the next check so that if one out of a set of
 	// Vaults is unhealthy we don't greatly increase requests to the healthy
 	// ones. This is less than the minimum window if all Vaults are healthy so
 	// that we don't desync from the larger window provided by Periodic
-	state.nextCheck = time.Now().Add(30 * time.Second)
+	state.nextCheck = time.Now().Add(29 * time.Second)
 	state.isAvailable = true
 
 	resp.Detected = true
--- a/client/fingerprint/vault.go
+++ b/client/fingerprint/vault.go
@@ -12,6 +12,7 @@ import (
 	log "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/nomad/helper"
 	"github.com/hashicorp/nomad/helper/useragent"
+	"github.com/hashicorp/nomad/nomad/structs/config"
 	vapi "github.com/hashicorp/vault/api"
 )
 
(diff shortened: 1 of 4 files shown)


client/fingerprint/vault_ce.go
```diff
@@ -0,0 +1,19 @@
...
+// vaultConfigs returns the set of Vault configurations the fingerprint needs to
+// check. In Nomad CE we only check the default Vault.
+func (f *VaultFingerprint) vaultConfigs(req *FingerprintRequest) map[string]*config.VaultConfig {
+	agentCfg := req.Config
+	if agentCfg.VaultConfig == nil || !agentCfg.VaultConfig.IsEnabled() {
+		return nil
+	}
+
+	return map[string]*config.VaultConfig{"default": agentCfg.VaultConfig}
+}
```
tgross, line 19: Note for reviewers: once I get to the point where I'm ready to start instantiating multiple clients in `client/vaultclient`, I'll probably pull this out to some shared helper, but I want to keep the review sizes small in the meanwhile.

client/fingerprint/vault.go
```diff
@@ -22,71 +23,90 @@ const (
...
 func (f *VaultFingerprint) Fingerprint(req *FingerprintRequest, resp *FingerprintResponse) error {
-	config := req.Config
-
-	if config.VaultConfig == nil || !config.VaultConfig.IsEnabled() {
-		return nil
+	for _, cfg := range f.vaultConfigs(req) {
+		err := f.fingerprintImpl(cfg, resp)
+		if err != nil {
+			return err
+		}
```
lgfa29, line 45: I imagine each fingerprint is fairly independent from each other, so it may be nice to accumulate errors from all fingerprints so operators can find and solve all problems at once instead of progressing through several attempts.

client/fingerprint/vault.go
```diff
@@ -22,71 +23,90 @@ const (
...
 	// Connect to vault and parse its information
-	status, err := f.client.Sys().SealStatus()
+	status, err := client.Sys().SealStatus()
 	if err != nil {
-
 		// Print a message indicating that Vault is not available anymore
-		if f.lastState == vaultAvailable {
-			f.logger.Info("Vault is unavailable")
+		if lastState, ok := f.lastStates[cfg.Name]; !ok || lastState == vaultAvailable {
+			f.logger.Info("Vault is unavailable", "cluster", cfg.Name)
```
lgfa29, line 73: Alternatively, we could create a method-specific logger with the label always applied so we don't forget to add the cluster:

```go
logger := f.logger.With("cluster", cfg.Name)
```

client/fingerprint/vault.go
```diff
@@ -22,71 +23,90 @@ const (
...
+		resp.AddAttribute("vault.accessible", strconv.FormatBool(true))
+		resp.AddAttribute("vault.version", strings.TrimPrefix(status.Version, "Vault "))
+		resp.AddAttribute("vault.cluster_id", status.ClusterID)
+		resp.AddAttribute("vault.cluster_name", status.ClusterName)
+	} else {
+		resp.AddAttribute(fmt.Sprintf("vault.%s.accessible", cfg.Name), strconv.FormatBool(true))
+		resp.AddAttribute(fmt.Sprintf("vault.%s.version", cfg.Name), strings.TrimPrefix(status.Version, "Vault "))
+		resp.AddAttribute(fmt.Sprintf("vault.%s.cluster_id", cfg.Name), status.ClusterID)
+		resp.AddAttribute(fmt.Sprintf("vault.%s.cluster_name", cfg.Name), status.ClusterName)
+	}
```
lgfa29, line 89: Is there value in also having `vault.default.*` attributes? I imagine that's what we want people to use moving forwards?

client/fingerprint/vault.go
```diff
@@ -22,71 +23,90 @@ const (
...
 func (f *VaultFingerprint) Periodic() (bool, time.Duration) {
-	if f.lastState == vaultAvailable {
-		// Fingerprint infrequently once Vault is initially discovered with wide
-		// jitter to avoid thundering herds of fingerprints against central Vault
-		// servers.
-		return true, (30 * time.Second) + helper.RandomStagger(90*time.Second)
+	for _, lastState := range f.lastStates {
+		if lastState != vaultAvailable {
+			return true, 15 * time.Second
+		}
```
lgfa29, line 105: Probably not worth changing anything, but I think this means that a single cluster not being available will cause all cluster to be fingerprinted at this accelerated interval.

lgfa29 (approved):

client/fingerprint/vault.go
tgross: Good call.

client/fingerprint/vault.go
tgross: I'm torn on this. That's what our RFC says, but in practice I suspect _most_ users will have 1 Vault (ex. all Nomad CE users will), which means we'd be sending a bunch of extra attributes no one cares about. And we can't _only_ send the `vault.default.*` b/c of backwards compatibility; a client updated before servers will appear to lose its Vault.

client/fingerprint/vault.go
tgross: Yeah, the architecture of periodic fingerprinting means we'd end up spinning up a separate fingerprinter goroutine for each Vault (and Consul, once that's done). Maybe we could do something a little more sophisticated here though and fire the fingerprinter goroutine without necessarily sending the API calls to healthy Vaults on each pass.

tgross: I gave this a go in 05a9047. The results are a little dissatisfying. The healthy Vaults effectively get checked on a fixed 45s interval when one of the other Vaults is unhealthy. But this avoids waking up the outer loop unnecessarily and avoids desync where healthy vaults get accidentally skipped for one cycle when the group as a whole skips from unhealthy to healthy.

client/fingerprint/vault.go
lgfa29: Yeah, good point, the common case is likely going to be a single Vault cluster, so having a bunch of duplicate attributes sound unnecessary (and kind of confusing).

client/fingerprint/vault.go
```diff
@@ -20,73 +22,121 @@ const (
...
+	if !state.isAvailable {
+		logger.Info("Vault is available", "cluster", cfg.Name)
 	}
-	f.lastState = vaultAvailable
+
+	// Widen the minimum window to the next check so that if one out of a set of
+	// Vaults is unhealthy we don't greatly increase requests to the healthy
+	// ones. This is less than the minimum window if all Vaults are healthy so
+	// that we don't desync from the larger window provided by Periodic
+	state.nextCheck = time.Now().Add(30 * time.Second)
```
lgfa29, line 120:
```suggestion
	// Widen the minimum window to the next check so that if one out of a set of
	// Vaults is unhealthy we don't greatly increase requests to the healthy
	// ones. This is less than the minimum window if all Vaults are healthy so
	// that we don't desync from the larger window provided by Periodic
	state.nextCheck = time.Now().Add(29 * time.Second)
```
`RandomStargger` may return `0`, which would cause `Periodic` and `nextCheck` to be both `30s`. I think that, depending on the timing, it may result in the a healthy cluster being skipped?

lgfa29 (approved):
Nice fix for the fingerprint timing! I think the 45s period for healthy clusters is OK. There will be a longer than usual delay in case of cascading Vault cluster failures, but by then fingerprint period is the least of your problems 😄

PR #18257, opened 2023-08-18
jobspec: add `role` to `vault`

lgfa29:
Add new `role` field to the `vault` block in jobspecs that will be used to derive Vault ACL tokens from JWTs.


tgross (approved):
LGTM!

Issue #18267, opened 2023-08-20
Nomad Server may instruct Clients to stop and GC all of their allocations.


stswidwinski:
### Nomad version
tip

### Operating system and Environment details
unix

### Issue -- TL;DR

Nomad clients may occasionally receive allocation updates from Servers corresponding to an old RAFT Index (they travel back in time). Consequently, they will kill some (or all) workloads which are running on them. This may be particularly acute when a server is restarting, but may happen without that precondition.

### Issue -- details

Nomad Clients ping the Nomad Server for information about allocations which are present on them according to the server point of view. This is done via a more efficient, dedicated RPC call between client and server, called "GetClientAllocs." Even though the client uses a blocking query which steadily increases the index for which the data is requested, the response may not be up-to-date resulting in partial or complete deletion of local state (which affects running workloads). 

The client does not check that the response corresponds to a RAFT index which is at least as new as requested.This staleness is usually bounded under normal operation, but can be arbitrary if any one of the servers is recovering the RAFT state via replication.

#### Background part 1: Client <-> Server communication for allocations.

Every client will connect to a server and ask for any updates regarding their node. The system seems to be poll based rather than push based and the logic is here:

https://github.com/hashicorp/nomad/blob/main/client/client.go#L2288

The general flow is the following:

1. Ask Nomad for all existing allocations for the current node. The allocations are represented as a map between allocID and ModifyIndex.
2. All allocations are divided into two categories:
  a. Filtered → Things that we shouldn't bother with
  b. Pull → Things that we don't know about. Every Allocation in the "pull" bucket is additionally read in full via another RPC
3. The set of allocations is passed on to "runAllocs" for processing. (see: https://github.com/hashicorp/nomad/blob/main/client/client.go#L2514)
4. runAllocs splits the allocs into more categories.
  a. Removed → Allocs that we know about but the server does not
  b. Added → Allocs that we don't know about but the server does
  c. Ignored → Allocs which we know about and the server knows about but didn't change or are otherwise irrelevant

`runAllocs` will do strictly what it is instructed to do and will not question that decision. Allocations which are running, but are to be "removed" will be shut down without additional checks.

Takeaways:

1. he logic of `watchAllocations` **does** send requests which are steadily increasing in required index (see: https://github.com/hashicorp/nomad/blob/main/client/client.go#L2458). 
2. However, **there are no checks to ensure that the data returned corresponds to a steadily increasing index.**.
[...]

PR #18269, opened 2023-08-21
Ensure that the index processed by the client is at least as new as the last one processed.

stswidwinski:
A fix for https://github.com/hashicorp/nomad/issues/18267, ensuring that "time travel" of the client state is not possible.

I could not find tests which may exercise this logic meaningfully. Please point me the right direction and I'm happy to add them

Please note that this isn't an actual good fix to the underlying issue, but rather a stop-gap measure.

### DIFF
--- a/.changelog/18269.txt
+++ b/.changelog/18269.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+client: Ignore stale server updates to prevent GCing allocations that should be running
+```
--- a/client/client.go
+++ b/client/client.go
@@ -2368,7 +2368,7 @@ OUTER:
 		// For full context, please see https://github.com/hashicorp/nomad/issues/18267
 		if resp.Index <= req.MinQueryIndex {
 			c.logger.Debug("Received stale allocation information. Retrying.",
-				"received index", resp.Index, "requested min index", req.MinQueryIndex)
+				"index", resp.Index, "min_index", req.MinQueryIndex)
 			continue OUTER
 		}
 
--- a/client/client.go
+++ b/client/client.go
@@ -2366,7 +2366,7 @@ OUTER:
 		// after a prolonged downtime.
 		//
 		// For full context, please see https://github.com/hashicorp/nomad/issues/18267
-		if resp.Index < req.MinQueryIndex {
+		if resp.Index <= req.MinQueryIndex {
 			c.logger.Debug("Received stale allocation information. Retrying.",
 				"received index", resp.Index, "requested min index", req.MinQueryIndex)
 			continue OUTER
--- a/client/client.go
+++ b/client/client.go
@@ -2360,6 +2360,18 @@ OUTER:
 		default:
 		}
 
+		// We have not received any new data, or received stale data. This may happen in
+		// an array of situations, the worst of which seems to be a blocking request
+		// timeout when the scheduler which we are contacting is newly added or recovering
+		// after a prolonged downtime.
+		//
+		// For full context, please see https://github.com/hashicorp/nomad/issues/18267
+		if resp.Index < req.MinQueryIndex {
+			c.logger.Debug("Received stale allocation information. Retrying.",
+				"received index", resp.Index, "requested min index", req.MinQueryIndex)
+			continue OUTER
+		}
+
 		// Filter all allocations whose AllocModifyIndex was not incremented.
 		// These are the allocations who have either not been updated, or whose
 		// updates are a result of the client sending an update for the alloc.
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -562,10 +562,16 @@ func TestClient_WatchAllocs(t *testing.T) {
 	alloc1.JobID = job.ID
 	alloc1.Job = job
 	alloc1.NodeID = c1.Node().ID
+
 	alloc2 := mock.Alloc()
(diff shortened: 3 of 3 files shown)


stswidwinski:
I don't think the UI tests are related here. May I ask someone to re-run the tests or help me understand the failure mode?

lgfa29:
Thanks for the PR and detailed issue description @stswidwinski.

> I could not find tests which may exercise this logic meaningfully. Please point me the right direction and I'm happy to add them

Yeah, this part of the code is very tricky to test reliably since there's a lot going on, causing tests to be flaky sometimes. [`TestClient_WatchAllocs`](https://github.com/hashicorp/nomad/blob/14a38bee7bc4386e74157f6a99f3db7382d7e6a5/client/client_test.go#L539-L625) may be a good place to start. I wonder if you setup a test like that and manually mess with the server raft index we could reliably reproduce and test the problem.

> I don't think the UI tests are related here. May I ask someone to re-run the tests or help me understand the failure mode?

I'm not sure what's going with the this failure. I suspect it may be because this is an external PR the Percy token may not be exposed. But we can safely ignore this error as that step only uploads UI visual diff testing results, which this PR wouldn't affect.

The changes look good to me, but I will just chat with the rest of the team to make sure I'm not missing anything.

Thanks again for the detailed work!

stswidwinski:
> Yeah, this part of the code is very tricky to test reliably since there's a lot going on, causing tests to be flaky sometimes. [TestClient_WatchAllocs](https://github.com/hashicorp/nomad/blob/14a38bee7bc4386e74157f6a99f3db7382d7e6a5/client/client_test.go#L539-L625) may be a good place to start. I wonder if you setup a test like that and manually mess with the server raft index we could reliably reproduce and test the problem.

Please see the recent commit comments for some more context. What do you think?

stswidwinski:
@lgfa29 please let me know how you'd like to proceed

lgfa29:
Hi @stswidwinski 👋 

Apologies for the delay, but I don't have a good answer yet 😅 

I'm checking with the rest of the team for any additional ideas and I will let you know if something comes up. But since this is such a tricky behaviour to trigger in a test we may need do without out for now.

stswidwinski:
Hey @lgfa29!

Thank you for the update and sorry to keep bugging you. Let me know if I can help in some way or if any part of the bug report (or PR) makes this more confusing. Otherwise, I'll sit tight :)

schmichael (approved):
I think this is a fine shortcircuit to add, but I am struggling to see where it could fix misbehavior.

> I'm beginning to think that these tests are not quite possible without broader changes to timeout settings.

I think you're correct here. One approach we could take is adding a `rpcQueryWaitTime time.Duration` to the `Client struct` and overriding it during tests to at least exercise the "blocking query timeout" case. At (normal) runtime the zero value should get converted to the default in [`Client.rpc(...)`](https://github.com/hashicorp/nomad/blob/v1.6.1/client/rpc.go#L91-L94), but it's probably best to explicitly set it to [`structs.DefaultBlockingRPCQueryTime`](https://github.com/hashicorp/nomad/blob/v1.6.1/nomad/structs/structs.go#L203) in `client.NewClient(...)`.

The main downside I can see is that this is yet-another-field on the already enormous Client struct, and a test-only one at that. I can't think of a less invasive change we could make to exercise some of this logic though.

I think moving `Client.RPC(...)` and dependencies into a new struct that implements `RPCer interface{ RPC(...) error }` or similar might provide us with a reasonable hook to manipulate RPC behavior in tests, but (a) that's a lot of work, and (b) we could easily mock so much that we're not actually testing realistic runtime scenarios anymore.

Another option would be to refactor the function being changed here into multiple functions and unit test them. This greatly increases the `(b) not testing realistic scenarios` issue above, but there may be opportunities to assert some useful things.

Long story short: I'd love someone to explore better testing options for Client RPCs, but I don't think we need to hold up this PR to wait for them.

stswidwinski:
> I think this is a fine shortcircuit to add, but I am struggling to see where it could fix misbehavior.

I think the basic tenant of the breakage right now is that right now we accept the answer of the Scheduler independent of whether it is new or if it is stale. In the case of staleness we would like to avoid applying the changes that such an answer would imply (killing allocations which are "unexpected" and garbage collecting them). 

The stale replies may exist if the server we are using to get information from (we allow staleness so this is not the leader) times out awaiting for the raft state to catch up with an index which would allow it to process the query correctly. This happens more frequently than one would expect because of the retry logic within the client. The issue submitted is quite verbose about this, but I'm happy to provide any more detail if it's helpful.

The change in a somewhat trivial way discards the data returned by the scheduler are requires that the response corresponds to the minimal index requested.

schmichael:
> I think the basic tenant of the breakage right now is that right now we accept the answer of the Scheduler independent of whether it is new or if it is stale.

Ah you're right! I was thinking this check was sufficient: https://github.com/hashicorp/nomad/pull/18269/files#diff-bd3a55a72186f59e2e63efb4951573b2f9e4a7cc98086e922b0859f8ccc1dd09R2396

However that check is only performed for allocs that are observable by the server returning the response. If the server is too stale to see the alloc exists, the Client won't even make it to that check and think the alloc has been GC'd by the server and therefore should be GC'd locally.

Yikes. I'll add a changelog entry and get this merged. Thanks @stswidwinski!

PR #18279, opened 2023-08-22
agent: add `consul.service_identity` and `consul.template_identity` blocks

pkazmierczak:
This PR introduces updates to the agent config required for workload identity support for services.

### DIFF
--- a/api/tasks.go
+++ b/api/tasks.go
@@ -9,8 +9,6 @@ import (
 	"path/filepath"
 	"strings"
 	"time"
-
-	"golang.org/x/exp/slices"
 )
 
 const (
@@ -1159,17 +1157,4 @@ type WorkloadIdentity struct {
 	Env         bool     `hcl:"env,optional"`
 	File        bool     `hcl:"file,optional"`
 	ServiceName string   `hcl:"service_name,optional"`
-}
-
-func (wi *WorkloadIdentity) Copy() *WorkloadIdentity {
-	if wi == nil {
-		return nil
-	}
-	return &WorkloadIdentity{
-		Name:        wi.Name,
-		Audience:    slices.Clone(wi.Audience),
-		Env:         wi.Env,
-		File:        wi.File,
-		ServiceName: wi.ServiceName,
-	}
 }
--- a/api/tasks.go
+++ b/api/tasks.go
@@ -9,6 +9,8 @@ import (
 	"path/filepath"
 	"strings"
 	"time"
+
+	"golang.org/x/exp/slices"
 )
 
 const (
@@ -1157,4 +1159,17 @@ type WorkloadIdentity struct {
 	Env         bool     `hcl:"env,optional"`
 	File        bool     `hcl:"file,optional"`
 	ServiceName string   `hcl:"service_name,optional"`
+}
+
+func (wi *WorkloadIdentity) Copy() *WorkloadIdentity {
+	if wi == nil {
+		return nil
+	}
+	return &WorkloadIdentity{
+		Name:        wi.Name,
+		Audience:    slices.Clone(wi.Audience),
+		Env:         wi.Env,
+		File:        wi.File,
+		ServiceName: wi.ServiceName,
+	}
 }
--- a/command/agent/config_parse_test.go
+++ b/command/agent/config_parse_test.go
@@ -235,7 +235,17 @@ var basicConfig = &Config{
 		ChecksUseAdvertise:   &trueValue,
 		Timeout:              5 * time.Second,
 		TimeoutHCL:           "5s",
-		UseIdentity:          &falseValue,
+		UseIdentity:          &trueValue,
+		ServiceIdentity: &config.WorkloadIdentity{
+			Audience: []string{"consul.io", "nomad.dev"},
+			Env:      false,
+			File:     true,
+		},
+		TemplateIdentity: &config.WorkloadIdentity{
+			Audience: []string{"consul.io"},
+			Env:      true,
+			File:     false,
+		},
 	},
 	Consuls: map[string]*config.ConsulConfig{
 		"default": {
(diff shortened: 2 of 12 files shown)


api/tasks.go
```diff
@@ -9,6 +9,8 @@ import (
 	"path/filepath"
 	"strings"
 	"time"
+
+	"golang.org/x/exp/slices"
```
shoenig, line 13: you could just use "slices" as of go1.21

nomad/structs/consul.go
```diff
@@ -3,34 +3,102 @@
...
+	// for graceful upgrades without forcing rewrite of all jobspecs. If set, when a
+	// job has both a template block and a consul block, the Nomad server will sign a
+	// Workload Identity for that task. The client will use this identity rather than
+	// the client's Consul token for the template hook.
+	//
+	// The name field of the identity is always set to "consul".
+	//
+	// TemplateIdentity is set on the server.
+	TemplateIdentity *WorkloadIdentity
 }
```
lgfa29, line 49: Is this the right place for these changes? This struct represents the `group` `consul` block, not the `consul` agent config block, which is here:
https://github.com/hashicorp/nomad/blob/0b9cf4e7b708accb11cf6e16c1b9648ad15f62c2/nomad/structs/config/consul.go#L17-L29

nomad/structs/config/consul.go
```diff
@@ -131,12 +133,42 @@ type ConsulConfig struct {
...
+	// TemplateIdentity is intended to reduce overhead for jobspec authors and make
+	// for graceful upgrades without forcing rewrite of all jobspecs. If set, when a
+	// job has both a template block and a consul block, the Nomad server will sign a
+	// Workload Identity for that task. The client will use this identity rather than
+	// the client's Consul token for the template hook.
+	//
+	// The name field of the identity is always set to "consul".
+	//
+	// TemplateIdentity is set on the server.
+	TemplateIdentity *WorkloadIdentity `mapstructure:"template_identity"`
```
lgfa29, line 168: I wonder if it would be good to group the fields per where they should be set, like

```hcl
// Client configuration values.
...

// Server configuration values
...
```

Or maybe even different structs? Though that could require quite a bit of adjustments. But having to look at individual field docs can be confusing.

nomad/structs/config/consul.go
```diff
@@ -260,6 +293,15 @@ func (c *ConsulConfig) Merge(b *ConsulConfig) *ConsulConfig {
...
 	}
+	if b.UseIdentity != nil {
+		result.UseIdentity = b.UseIdentity
+	}
+	if b.ServiceIdentity != nil {
+		result.ServiceIdentity = b.ServiceIdentity
+	}
+	if b.TemplateIdentity != nil {
+		result.TemplateIdentity = b.TemplateIdentity
+	}
```
lgfa29, line 304: I don't immediately see other instances like this, but I think we would want to copy these values to prevent accidental changes?

nomad/structs/config/consul.go
```diff
@@ -359,6 +401,74 @@ func (c *ConsulConfig) Copy() *ConsulConfig {
 	if nc.AllowUnauthenticated != nil {
 		nc.AllowUnauthenticated = pointer.Of(*nc.AllowUnauthenticated)
 	}
+	if nc.UseIdentity != nil {
+		nc.UseIdentity = pointer.Of(*nc.UseIdentity)
+	}
```
lgfa29, line 406: We also need to copy `ServiceIdentity` and `TemplateIdentity`.

nomad/structs/config/consul.go
```diff
@@ -359,6 +401,74 @@ func (c *ConsulConfig) Copy() *ConsulConfig {
...
 
 	return nc
 }
+
+// WorkloadIdentity is the jobspec block which determines if and how a workload
+// identity is exposed to tasks similar to the Vault block.
+//
+// This is a copy of WorkloadIdentity from nomad/structs package in order to
+// avoid import cycles.
```
lgfa29, line 415: It may be good to add a similar comment in `structs` as a reminder to apply changes here as well.

nomad/structs/config/consul.go
```diff
@@ -359,6 +401,74 @@ func (c *ConsulConfig) Copy() *ConsulConfig {
...
+	// Env injects the Workload Identity into the Task's environment if
+	// set.
+	Env bool
+
+	// File writes the Workload Identity into the Task's secrets directory
+	// if set.
+	File bool
+
+	// ServiceName is used to bind the identity to a correct Consul service.
+	ServiceName string
```
lgfa29, line 432: This field now makes me wonder that this is not quite the same as the `WorkloadIdentity` in `structs`. I think it would never make sense to set this as an agent config? 🤔 

We could leave it in to keep the structs in sync, but maybe we need to make sure it's not set? Otherwise all service identities would have the same `ServiceName`

lgfa29 (approved):
I left some food for thought there. I don't think anything is particularly wrong, but some things can get confusing 😅 

Documenting the split between server config and client config will be important in the website as well.

command/agent/config_parse_test.go
```diff
@@ -260,6 +261,7 @@ var basicConfig = &Config{
 			ChecksUseAdvertise:   &trueValue,
 			Timeout:              5 * time.Second,
 			TimeoutHCL:           "5s",
+			UseIdentity:          &falseValue,
```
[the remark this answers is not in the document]
pkazmierczak: oof you are so right Luiz. Corrected in [7a0f84f](https://github.com/hashicorp/nomad/pull/18279/commits/7a0f84fabbbd2e6cb1a5efdb867d224fc6de1992)

nomad/structs/config/consul.go
pkazmierczak: corrected in [7a0f84f](https://github.com/hashicorp/nomad/pull/18279/commits/7a0f84fabbbd2e6cb1a5efdb867d224fc6de1992)

nomad/structs/config/consul.go
pkazmierczak: corrected in [7a0f84f](https://github.com/hashicorp/nomad/pull/18279/commits/7a0f84fabbbd2e6cb1a5efdb867d224fc6de1992)

nomad/structs/config/consul.go
pkazmierczak: I made `mapstructure` and `json` decoders ignore the field for agent config.

Issue #18127, opened 2023-08-02
Render customizable links in dashboard


mr-karan:
Hey!

Proposing a new feature spec that can be useful to add in the Job Spec:

```
  links {
    link "monitoring" {
      url = "https://monitoring.org.internal/grafana/d/5XxwU0GVz/allocations?orgId=1&var-cluster=my-cluster&var-namespace=${NOMAD_NAMESPACE}&var-workload=${NOMAD_JOB_ID}&var-group=${NOMAD_ALLOC_ID}"
      label = "Grafana"
    }
    link "logs" {
      url = "https://kibana.internal.org/logs/${NOMAD_ALLOC_ID}"
      label = "Kibana"
    }
    link "repo" {
      url = "https://github.com/org/code"
      label = "GitHub"
    }
  }
```

![image](https://github.com/hashicorp/nomad/assets/5689132/4d376e0f-81d4-4a27-a42d-78580a940b9e)

These can be rendered in the Job/Alloc/Task page. It'll be super handy to redirect to external platform where we can check the logs/dashboard/runbooks based on a templated URL. Will save a lot of time for folks looking to quickly troubleshoot applications using external tools.

The `url` field can support optional templating as well, which will help to quickly open relevant dashboard based on the context of the current job/group opened.

Would love community feedback on this.

Issue #15706, opened 2023-01-06
Description field 


davidlublink:
### Proposal

I would like to see a Description field that could appear on a job, task or group.

The field would contain markdown + mermaid

Markdown => https://en.wikipedia.org/wiki/Markdown

Mermaid => https://mermaid.js.org/#/

### Use-cases
1. Operator unfamiliar with a service needs more information about a service! With this idea the operator would see the information directly. Links could take them to any external tool such as a shared disk drive with a document on what to do.
2. A home network operator runs multiple services and the said operator might need to leave instructions for the kids if they need to restart the minecraft service or anything else.
3.  With a mermaid graph rendered as part of a service, it could allow for a quick understanding of the components being viewed. If the mermaid graph was rendered with links to services in the nomad group, that would be amazing...

tgross:
Hi @davidlublink! This is an interesting use case for sure. The data side of the problem has a couple options for you already: we already have the [`meta`](https://developer.hashicorp.com/nomad/docs/job-specification/meta) field which can contain arbitrary data. We also have Variables, so you could potentially have `meta` include a link to the Variable that contains the description if you didn't want to have large amounts of text in your jobspec. 

But then displaying the meta in the UI and/or rendering the resulting markdown is a whole other problem. The mermaid graphs might be the sort of thing we'd rather do in a more first-class way across jobs (via the service discovery interface). This is fairly open-ended so I'll raise this with the UI folks to see what they think.

davidlublink:
Thanks for the reply, you are right that meta tags can be used but they won't be rendered in the web interface like you mentioned.

davidlublink:
I wonder if we could just add a description to the job itself. If it supports markdown it would allow for linking outside of Nomad and leaving instructions for an operator interacting with the job in the future.

I added a field description in my PoC, but @tgross is right it could be done with a meta value removing any backend changes.

The biggest issue I ran into is that Nomad seems to include CSS libraries that reset a lot of default behaviours and the markdown libraries depend on the default behaviours, so the rendering is less than ideal. ( see unordered list in screenshot ) .
 


```
job "hello-world" {
  // Specifies the datacenter where this job should be run
  // This can be omitted and it will default to ["*"]
  datacenters = ["jo","bin"]
  description = <<EOH
[...]

PR #18292, opened 2023-08-22
[ui] Jobspec UI block: Descriptions and Links

philrenaud:
Adds a `ui` block to Nomad job specs. Users can define any number of `link`s and a single `description` that will show up on the Job index page in the web UI:

image
from spec

```
job "fails_every_10" {

  ui {
    description = "A job that fails **every 10 allocations**, but a 1x [restart](https://nomadproject.io) and 1x <a href='https://hashicorp.com'>reschedule</a> means that a given allocation would have to fail 4x (initial, restarted, rescheduled, rescheduled+restarted) to be considered permanently failed, or 0.01% permanent failure rate (1/10,000 allocs)"
    link {
      label = "nomad"
      url = "https://nomadproject.io"
    }
    link {
      label = "Some Arbitrary String Label"
      url = "http://hashicorp.com/"
    }
  }

...
```

Notably, this supports markdown, so you can, for example, add a table with a jobspec like this:

```
job "ui-block-tester" {

  ui {
    description = <<EOT
| Name              | Team                 | Position | Average | HRs | Wikipedia Article                                          |
|-------------------|----------------------|----------|---------|-----|------------------------------------------------------------|
| Babe Ruth         | New York Yankees     | OF/P     | .342    | 714 | [Link](https://en.wikipedia.org/wiki/Babe_Ruth)            |
| Hank Aaron        | Atlanta Braves       | OF       | .305    | 755 | [Link](https://en.wikipedia.org/wiki/Hank_Aaron)           |
| Ted Williams      | Boston Red Sox       | OF       | .344    | 521 | [Link](https://en.wikipedia.org/wiki/Ted_Williams)         |
| Jackie Robinson   | Brooklyn Dodgers     | 2B       | .311    | 137 | [Link](https://en.wikipedia.org/wiki/Jackie_Robinson)      |
| Roberto Clemente  | Pittsburgh Pirates   | OF       | .317    | 240 | [Link](https://en.wikipedia.org/wiki/Roberto_Clemente)     |
EOT
    link {
      label = "nomad"
      url = "https://nomadproject.io"
    }
    link {
      label = "Some Arbitrary String Label"
      url = "http://hashicorp.com/"
    }
  }

...
```

image

Resolves https://github.com/hashicorp/nomad/issues/18127
Resolves https://github.com/hashicorp/nomad/issues/15706


octdanb:
This is a great proposal, +1

nomad/structs/structs.go
```diff
@@ -4457,6 +4457,19 @@ type Job struct {
 	ModifyIndex uint64
 	// JobModifyIndex is the index at which the job *specification* last changed
 	JobModifyIndex uint64
+
+	// Links and Description fields for the Web UI
+	Ui *JobUIConfig
```
tgross, line 4462: Idiomatic Go would have this field name in all-caps (https://go.dev/wiki/CodeReviewComments#initialisms):

```suggestion
	UI *JobUIConfig
```

That change should ripple down to the `api` package, docs, etc. as well.

website/content/docs/job-specification/ui.mdx
```diff
@@ -0,0 +1,47 @@
...
+    link {
+      label = "Learn more about Nomad"
+      url = "https://developer.hashicorp.com/nomad"
+    }
+    link {
+      label = "Nomad on Github"
+      url = "https://github.com/hashicorp/nomad"
+    }
+  }
+  # ...
```
tgross, line 45: Our HCL examples should always be `hclfmt`'d:

```suggestion
job "docs" {
  ui {
    description = "A job that uses **Nomad Variables**"
    link {
      label = "Learn more about Nomad"
      url   = "https://developer.hashicorp.com/nomad"
    }
    link {
      label = "Nomad on Github"
      url   = "https://github.com/hashicorp/nomad"
    }
  }
  # ...
```

website/content/docs/job-specification/ui.mdx
```diff
@@ -0,0 +1,47 @@
...
+
+The `ui` block provides options to modify the presentation of the Job index page
+in the Web UI. When specified, a description and any number of links will be
+added to the top of the job page in question.
+
+The following will provide the Web UI with a job description and a pair of links:
+
+## `ui` Parameters
+
+- `description` `(string: "")` - The markdown-enabled description of the job
```
tgross, line 25: We're calling this "markdown-enabled" but we should probably tell the user what markdown dialect is permitted. Ex. is this GitHub-flavored markdown, commonmark, etc.? Also, is there any markdown feature that's disallowed here (ex. can the user add html markup and XSS the cluster admin, or is that disabled?)

nomad/structs/structs.go
```diff
@@ -4457,6 +4457,19 @@ type Job struct {
...
+}
+
+type JobUIConfig struct {
+	Description string
+	Links       []JobUILink
+}
```
tgross, line 4468: This will need a `Copy` and `Equal` method, and have it called from the `Copy`/`Equal` method of `Job`.

api/jobs.go
```diff
@@ -933,6 +933,16 @@ type JobSubmission struct {
 	Variables string
 }
 
+type JobUIConfig struct {
+	Description string       `hcl:"description,optional"`
+	Links       []*JobUILink `hcl:"link,block"`
+}
```
tgross, line 939: This likely needs a `Copy` and `Canonicalize` method, which is called from the same method on `Job`.

tgross (changes requested):
Overall this looks good @philrenaud! But you should take a pass through the [jobspec checklist](https://github.com/hashicorp/nomad/blob/main/contributing/checklist-jobspec.md) to make sure all the little helper methods listed there have been implemented.

website/content/docs/job-specification/ui.mdx
philrenaud: Thanks for this comment; I've added render sanitization via https://github.com/hashicorp/nomad/pull/18292/commits/ed9badf7de6b9443b0d178aaeb2356bb6bb49491

Juanadelacuesta (dismissed):

api/jobs.go
```diff
@@ -933,6 +933,56 @@ type JobSubmission struct {
...
+	Label string `hcl:"label,optional"`
+	URL   string `hcl:"url,optional"`
+}
+
+func (j *JobUIConfig) Canonicalize() {
+	if j == nil {
+		return
+	}
+
+	if j.Description == "" {
```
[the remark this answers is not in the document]
philrenaud: I think this is a remnant of when I was working with a separate cloned job object here; deleting.

nomad/structs/structs.go
```diff
@@ -4554,6 +4594,7 @@ func (j *Job) Copy() *Job {
 	nj.Constraints = CopySliceConstraints(nj.Constraints)
 	nj.Affinities = CopySliceAffinities(nj.Affinities)
 	nj.Multiregion = nj.Multiregion.Copy()
```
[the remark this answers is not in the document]
philrenaud: I don't understand the wider go context here, but I was just trying to follow the Multiregion/Affinities/Constraints/Datacenters convention. I think this may be deep-copying the UI object after the `*nj = *j` shallow copied it? At least that's how I read it.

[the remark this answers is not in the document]
Juanadelacuesta: `nj.Multiregion = nj.Multiregion.Copy()` this line is not copying anything, you are assigning the value of nj.Multiregion to nj.Multiregion, it self. You need to return a copy of j, so it should be 

```suggestion
	nj.Multiregion = j.Multiregion.Copy()
```

david-yu:
Wondering what it would take to perhaps have a config flag that gates this functionality for the UI?

philrenaud:
@david-yu I believe a sentinel policy could target this block.

david-yu:
> @david-yu I believe a sentinel policy could target this block.

That could possibly work.

nomad/structs/structs.go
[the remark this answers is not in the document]
philrenaud: Hope I've addressed this [here](https://github.com/hashicorp/nomad/pull/18292/commits/1e7dd3949e989f826e39f58af4d25ab87a0bc93b)

command/agent/job_endpoint.go
```diff
@@ -2075,6 +2076,23 @@ func ApiAffinitiesToStructs(in []*api.Affinity) []*structs.Affinity {
 	return out
 }
 
+func ApiJobUIConfigToStructs(jobUI *api.JobUIConfig) *structs.JobUIConfig {
```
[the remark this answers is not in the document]
philrenaud: Thanks! Added some [here](https://github.com/hashicorp/nomad/pull/18292/commits/9c64d9771d456366c285b4b0fbbcc9300c70efbd). Got me thinking about "Do we want to return nil or an empty array when there are no links", which I hadn't given much consideration to before.

nomad/structs/structs.go
```diff
@@ -4457,6 +4457,46 @@ type Job struct {
...
+	copy := new(JobUIConfig)
+	copy.Description = j.Description
+
+	if j.Links != nil {
+		links := make([]JobUILink, len(j.Links))
+		for i, link := range j.Links {
+			links[i] = *link.Copy()
+		}
+		copy.Links = links
+	}
```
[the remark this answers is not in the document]
tgross: That's odd. Feels like a bug in `CopySlice` tbh but not worth digging into at the moment.

command/agent/job_endpoint_test.go
```diff
@@ -4356,3 +4356,51 @@ func Test_apiWorkloadIdentityToStructs(t *testing.T) {
 		TTL:          2 * time.Hour,
 	}))
 }
+
+func TestConversion_ApiJobUIConfigToStructs(t *testing.T) {
+	t.Run("nil jobUI", func(t *testing.T) {
+		require.Nil(t, ApiJobUIConfigToStructs(nil))
```
tgross, line 4362: Use `must.Nil`/`must.Eq` on new tests, instead of `require`. We're slowing trying to migrate over to use `shoenig/test` instead of `testify`

nomad/structs/structs.go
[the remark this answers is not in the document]
tgross: Oh, I see why: this isn't a slice of pointers. `Links       []JobUILink` should be `Links       []*JobUILink` to match what's in the `api` package here.

nomad/structs/structs.go
```diff
@@ -4457,6 +4457,46 @@ type Job struct {
...
 	// JobModifyIndex is the index at which the job *specification* last changed
 	JobModifyIndex uint64
+
+	// Links and Description fields for the Web UI
+	UI *JobUIConfig
+}
+
+type JobUIConfig struct {
+	Description string
+	Links       []JobUILink
```
tgross, line 4467:
```suggestion
	Links       []*JobUILink
```

tgross (changes requested):
Last item on the pointer on `JobUIConfig.Links` and this should be good-to-go

tgross (approved):
LGTM!

PR #18315, opened 2023-08-24
ci: ensure semgrep tests all state store funcs for FSM time rule.

jrasell:


.semgrep/fsm_time.yml
```diff
@@ -30,4 +30,7 @@ rules:
     paths:
       include:
         - "nomad/fsm.*"
```
lgfa29, line 32: Seems like `fsm.*` has the same problem? 

I don't remember the syntax details, but would something like this work?

```yaml
include:
  - "nomad/fsm*"
  - "nomad/state/state_store*"
exclude:
  - "*_test.go"
```

jrasell: Good shout, modified within: https://github.com/hashicorp/nomad/pull/18315/commits/e49bb26795cf8ac1664580b6edaad570be3f7ac0

Tested locally and it catches the right files.

PR #18317, opened 2023-08-24
job_endpoint: jobIdentityCreator hook

pkazmierczak:
This PR introduces a job-mutating hook that inserts an `Identity` block for services that use the `consul` provider. 

### DIFF
--- a/nomad/job_endpoint_hooks_test.go
+++ b/nomad/job_endpoint_hooks_test.go
@@ -806,3 +806,113 @@ func TestJob_submissionController(t *testing.T) {
 		must.Nil(t, args.Submission)
 	})
 }
+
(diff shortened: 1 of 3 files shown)


nomad/job_endpoint_hooks_test.go
```diff
@@ -806,3 +806,113 @@ func TestJob_submissionController(t *testing.T) {
...
+								Provider: "consul",
+								Identity: &structs.WorkloadIdentity{
+									Name:     "test",
+									Audience: []string{"consul.io"},
+								},
+							},
+						},
+					},
+				},
+			},
```
lgfa29, line 874: Do we need to set the proper `Name` and `ServiceName`? I think it's not expected of users to provide them.

lgfa29 (approved):

PR #18320, opened 2023-08-25
job_endpoint hooks: fix identity block mutator behavior

pkazmierczak:
If there's an `identity` block present in the config, do not overwrite it
completely. Only set the `Name` and `ServiceName` fields, and keep the rest.


jrasell (approved):
LGTM, but I am curious the reason for overriding a subset of the fields? Is this something that might be useful as a comment?

pkazmierczak:
> I am curious the reason for overriding a subset of the fields? Is this something that might be useful as a comment?

good shout. I clarified in a comment in c96478583a7bd5131fe247ddda2e181799cced85

Issue #18325, opened 2023-08-25
ci: fix semgrep warnings and failures across all code


jrasell:
The [semgrep](https://github.com/hashicorp/nomad/tree/main/.semgrep) rules we have are triggered on each PR raised, however, they are only ran on the files which have changed as part of the PR. When running the ruleset locally, a number of problems are found. These should be fixed, so that the code is both better, and that we can run semgrep across all files for CI.

<details>
<summary>Semgrep Results:</summary>

```
$ semgrep --config=.semgrep/ ./

┌─────────────┐
│ Scan Status │
└─────────────┘
  Scanning 5233 files tracked by git with 17 Code rules:

  Language      Rules   Files          Origin    Rules
 ─────────────────────────────        ─────────────────
  <multilang>       2    2580          Unknown      17
  go                8    1780
  js                1     904

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:03
  
┌──────────────────┐
│ 73 Code Findings │
└──────────────────┘

    client/allocrunner/alloc_runner_test.go
       semgrep.tests-no-parallel
          Use ci.Parallel(t) instead of t.Parallel()

           ▶▶┆ Autofix ▶ ci.Parallel(t)
         2086┆ t.Parallel()

    client/allocrunner/consul_grpc_sock_hook.go
       semgrep.time-after-leak
          Potential leak of time.Timer, consider using NewSafeTimer instead

          251┆ select {
          252┆ case <-p.doneCh:
          253┆     return
  nil
          254┆ case <-time.After(socketProxyStopWaitTime):
          255┆     return
  errSocketProxyTimeout

    client/allocrunner/consul_http_sock_hook.go
       semgrep.time-after-leak
          Potential leak of time.Timer, consider using NewSafeTimer instead

          197┆ select {
          198┆ case <-p.doneCh:
          199┆ case <-time.After(socketProxyStopWaitTime):
          200┆     return
  errSocketProxyTimeout

    client/allocrunner/health_hook.go
       semgrep.time-after-leak
          Potential leak of time.Timer, consider using NewSafeTimer instead

          238┆ select {
          239┆ case <-ctx.Done():
          240┆     // Graceful
  shutdown
          241┆
  return
          242┆
          243┆ case <-tracker.AllocStoppedCh():
          244┆     // Allocation has stopped so no need to set
  health
          245┆
  return
          246┆
[...]

PR #18326, opened 2023-08-25
test: use correct parallel test setup func

jrasell:
related: https://github.com/hashicorp/nomad/issues/18325


pkazmierczak (approved):
Nice catch @jrasell, LGTM!

PR #18342, opened 2023-08-25
config: adjust workload identity struct

lgfa29:
Rename the agent configuraion for workload identity to `WorkloadIdentityConfig` to make its use more explicit.

Also removes the field `ServiceName` since it's never expected to be defined in a configuration file and add default values for `consul.service_identity` and `consul.template_identity`.


command/agent/agent.go
```diff
@@ -506,14 +506,14 @@ func convertServerConfig(agentConfig *Config) (*nomad.Config, error) {
 	}
 
 	// Add the Consul and Vault configs
-	conf.ConsulConfig = agentConfig.Consul
+	conf.ConsulConfig = agentConfig.Consul.Copy()
 	for _, consulConfig := range agentConfig.Consuls {
-		conf.ConsulConfigs[consulConfig.Name] = consulConfig
+		conf.ConsulConfigs[consulConfig.Name] = consulConfig.Copy()
```
lgfa29, line 511: I think we would want to pass a copy here to avoid accidental mutations?

nomad/structs/config/consul.go
```diff
@@ -191,6 +191,17 @@ func DefaultConsulConfig() *ConsulConfig {
...
+			Audience: []string{"consul.io"},
+			Env:      pointer.Of(false),
+			File:     pointer.Of(false),
+		},
+		TemplateIdentity: &WorkloadIdentityConfig{
+			Name:     "consul",
+			Audience: []string{"consul.io"},
+			Env:      pointer.Of(false),
+			File:     pointer.Of(false),
+		},
```
lgfa29, line 204: Are these the right defaults?

nomad/structs/config/workload_id.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	if !pointer.Eq(wi.File, other.File) {
+		return false
+	}
+
+	return true
+}
+
+func (wi *WorkloadIdentityConfig) Merge(other *WorkloadIdentityConfig) *WorkloadIdentityConfig {
```
lgfa29, line 77: The `structs` package usually merge things in-place, but on `config` the preference is to return a new struct.

command/agent/agent.go
lgfa29: Or maybe not, it has always been without copy 🤷

command/agent/config_parse_test.go
```diff
@@ -212,111 +265,13 @@ var basicConfig = &Config{
...
-		Role:                 "test_role",
-		TLSCaFile:            "/path/to/ca/file",
-		TLSCaPath:            "/path/to/ca",
-		TLSCertFile:          "/path/to/cert/file",
-		TLSKeyFile:           "/path/to/key/file",
-		TLSServerName:        "foobar",
-		TLSSkipVerify:        &trueValue,
-		TaskTokenTTL:         "1s",
-		Token:                "12345",
+		"default": basicConsulConfig,
```
lgfa29, line 270: I'm not sure which change I made that started causing tests to fail unless `default` is the exact same pointer as `Consul`, and same for `Vault`. 

I think this is the structure we want, so maybe I fixed something? 😅

command/agent/config_parse_test.go
```diff
@@ -620,6 +575,9 @@ func (c *Config) addDefaults() {
 	if c.Consul == nil {
 		c.Consul = config.DefaultConsulConfig()
 		c.Consuls = map[string]*config.ConsulConfig{"default": c.Consul}
+	} else {
+		c.Consul = config.DefaultConsulConfig().Merge(c.Consul)
+		c.Consuls = map[string]*config.ConsulConfig{"default": c.Consul}
```
lgfa29, line 580: These tests were really confusing to me, as they mix file parsing and setting default values.

I had to add this logic because `basicConfig` already has a `consul` block, so the Consul's default were not being set. 

Since `TemplateIdentity.Name` is one of these defaults, and cannot be set in the config file, `TestConfig_Parse/basic.hcl` and `TestConfig_Parse/basic.json` would fail because `basicConfig` doesn't have the identity name set, while `actual = oldDefault.Merge(actual)` will set the proper name.

With this change, the call to `basicConfig.addDefaults()` will set the service name.

(side note: it's kind of scary that we're modifying a global variable inside a test case 😬)

nomad/structs/config/consul.go
pkazmierczak: It looks right to me.

lgfa29:
Closing this in favour of https://github.com/hashicorp/nomad/pull/18363

PR #18343, opened 2023-08-26
vault: add `use_identity` and `default_identity` agent configuration and implicit workload identity

lgfa29:
Add new agent configuration `use_identity` and `default_identity` to the `vault` block.

`use_identity` is a boolean value that determines if Vault tokens should be derived from workload identity JWTs instead of the traditional flow of relying on Nomad servers for token derivation. If `use_identity` is `true` jobs with a `vault` block receive an implicit `identity` called `vault` if one is not already present.

`default_identity` is the `identity` block injected to tasks that don't define an `identity` block named `vault`.

Job validation on register becomes a little more complicated because we now need to check three places:
1. `vault` block in job
2. `vault` block in server config
3. `identity` in task

```mermaid
flowchart TD
	Start[Job Registered] --> HasVault{Job has vault block?}

    HasVault -->|No| HasVaultIdentityNoVault{Job has vault identity?}
    HasVault -->|Yes| HasVaultIdentityWithVault{Job has vault identity?}

    HasVaultIdentityNoVault -->|Yes| VaultNoIdentityWarn[Warn\nVault identity without vault block]
    HasVaultIdentityNoVault --> |No| NoVault[No Vault]
    VaultNoIdentityWarn --> NoVault

    HasVaultIdentityWithVault -->|Yes| UseIdentityWithVaultIdentity{Server vault.use_identity}
    HasVaultIdentityWithVault -->|No| UseIdentityWithoutVaultIdentity{Server vault.use_identity}

    UseIdentityWithVaultIdentity -->|True| WIDHasPolicies{Vault block has policies?}
    UseIdentityWithVaultIdentity -->|False| NoVaultIdentityWarn[Warn\nVault identity without using identity]
    
    
    WIDHasPolicies -->|No| UseWID[Use workload identity]
    WIDHasPolicies -->|Yes| DeprecatePoliciesWarn[Warn\nUsing identity with policies]
    DeprecatePoliciesWarn --> UseWID
    
    NoVaultIdentityWarn --> LegacyHasPolicies{Vault block has policies?}
    LegacyHasPolicies -->|Yes| UseLegacy[Use legacy flow]
    LegacyHasPolicies -->|No| MissingPoliciesError[Error\nLegacy flow without policies]
    
    UseIdentityWithoutVaultIdentity -->|False| LegacyHasPolicies
    UseIdentityWithoutVaultIdentity -->|True| HasDefaultIdentity{Server has default_identity?}
    
    HasDefaultIdentity -->|No| NoIdentityError[Error\nUsing indentity without identity]
    HasDefaultIdentity -->|Yes| InjectDefaultIdentity[Inject default identity]
    InjectDefaultIdentity --> WIDHasPolicies
```

_Note to reviewers: https://github.com/hashicorp/nomad/pull/18343/commits/98c09961b33706b9f6817aca9d8360aa75f65034 causes some diffs that are not necessary to the core logic because it just re-order struct fields. You can review the PR by commit or look at the diff without this commit in https://github.com/hashicorp/nomad/pull/18343/files/19fd50976f1436818c4d16267269ccac00682617_

### DIFF
--- a/command/agent/config_parse.go
+++ b/command/agent/config_parse.go
@@ -397,16 +397,16 @@ func parseVaults(c *Config, list *ast.ObjectList) error {
 
 		if o := listVal.Filter("default_identity"); len(o.Items) > 0 {
 			var m map[string]interface{}
-			defaultIdendityBlock := o.Items[0]
-			if err := hcl.DecodeObject(&m, defaultIdendityBlock.Val); err != nil {
+			defaultIdentityBlock := o.Items[0]
+			if err := hcl.DecodeObject(&m, defaultIdentityBlock.Val); err != nil {
 				return err
 			}
 
-			var defaultIdendity config.WorkloadIdentityConfig
-			if err := mapstructure.WeakDecode(m, &defaultIdendity); err != nil {
+			var defaultIdentity config.WorkloadIdentityConfig
+			if err := mapstructure.WeakDecode(m, &defaultIdentity); err != nil {
 				return err
 			}
-			c.Vaults[v.Name].DefaultIdentity = &defaultIdendity
+			c.Vaults[v.Name].DefaultIdentity = &defaultIdentity
 		}
 	}
 
--- a/command/agent/config_parse.go
+++ b/command/agent/config_parse.go
@@ -199,6 +199,17 @@ func ParseConfigFile(path string) (*Config, error) {
 		}
 	}
 
+	for name, vaultConfig := range c.Vaults {
+		if vaultConfig.DefaultIdentity != nil {
+			tds = append(tds, durationConversionMap{
+				fmt.Sprintf("vaults.%s.default_identity.ttl", name), nil, &vaultConfig.DefaultIdentity.TTLHCL,
+				func(d *time.Duration) {
+					vaultConfig.DefaultIdentity.TTL = d
+				},
+			})
+		}
+	}
+
 	// Add enterprise audit sinks for time.Duration parsing
 	for i, sink := range c.Audit.Sinks {
 		tds = append(tds, durationConversionMap{
--- a/command/agent/config_parse.go
+++ b/command/agent/config_parse.go
@@ -370,6 +370,9 @@ func parseVaults(c *Config, list *ast.ObjectList) error {
 		if err := hcl.DecodeObject(&m, obj.Val); err != nil {
 			return err
 		}
+
+		delete(m, "default_identity")
+
 		v := &config.VaultConfig{}
 		err := mapstructure.WeakDecode(m, v)
 		if err != nil {
@@ -382,6 +385,28 @@ func parseVaults(c *Config, list *ast.ObjectList) error {
 			c.Vaults[v.Name] = exist.Merge(v)
 		} else {
 			c.Vaults[v.Name] = v
+		}
+
+		// Decode the default identity.
+		var listVal *ast.ObjectList
+		if ot, ok := obj.Val.(*ast.ObjectType); ok {
+			listVal = ot.List
+		} else {
+			return fmt.Errorf("should be an object")
+		}
+
+		if o := listVal.Filter("default_identity"); len(o.Items) > 0 {
+			var m map[string]interface{}
+			defaultIdendityBlock := o.Items[0]
+			if err := hcl.DecodeObject(&m, defaultIdendityBlock.Val); err != nil {
+				return err
+			}
+
+			var defaultIdendity config.WorkloadIdentityConfig
+			if err := mapstructure.WeakDecode(m, &defaultIdendity); err != nil {
+				return err
+			}
+			c.Vaults[v.Name].DefaultIdentity = &defaultIdendity
 		}
 	}
 
--- a/nomad/job_endpoint_hook_implicit_identities.go
+++ b/nomad/job_endpoint_hook_implicit_identities.go
@@ -79,17 +79,13 @@ func (h jobImplicitIdentitiesHook) handleConsulService(s *structs.Service) {
 
 // handleVault injects a workload identity to the task if:
 //  1. The task has a Vault block.
-//  2. The server is configures with `vault.use_identity = true` and a
+//  2. The server is configured with `vault.use_identity = true` and a
 //     `vault.default_identity` is provided.
 //
 // If the task already has an identity named `vault` it sets the identity name
 // to the expected value.
 func (h jobImplicitIdentitiesHook) handleVault(t *structs.Task) {
-	if !h.srv.config.UseVaultIdentity() {
-		return
-	}
-
-	if t.Vault == nil {
+	if !h.srv.config.UseVaultIdentity() || t.Vault == nil {
 		return
 	}
 
--- a/nomad/job_endpoint_hook_implicit_identities.go
+++ b/nomad/job_endpoint_hook_implicit_identities.go
@@ -9,8 +9,9 @@ import (
 	"github.com/hashicorp/nomad/nomad/structs"
 )
 
-var (
+const (
 	consulServiceIdentityNamePrefix = "consul-service"
+	vaultIdentityName               = "vault"
 )
 
 // jobImplicitIdentitiesHook adds implicit `identity` blocks for external
@@ -33,6 +34,7 @@ func (h jobImplicitIdentitiesHook) Mutate(job *structs.Job) (*structs.Job, []err
 			for _, s := range t.Services {
 				h.handleConsulService(s)
 			}
+			h.handleVault(t)
 		}
 	}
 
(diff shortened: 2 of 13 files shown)


nomad/job_endpoint_hooks.go
```diff
@@ -307,45 +308,95 @@ func (v *jobValidate) Validate(job *structs.Job) (warnings []error, err error) {
...
+	var mErr *multierror.Error
+
+	// Prefix errors and warnings.
+	defer func() {
+		prefix := fmt.Sprintf("Task %s", t.Name)
+		for i, w := range warnings {
+			warnings[i] = fmt.Errorf("%s %s", prefix, w)
+		}
+		err = multierror.Prefix(mErr, prefix)
+	}()
```
lgfa29, line 369: I promise I tried to make these errors not 💩, but the whole `multierror` and `[]error` thing makes it very hard to generate a good output. I started to refactor this whole error system to give it more structure, but it snowballed pretty quickly.

Here's a warning output example:

image

nomad/structs/structs.go
```diff
@@ -9926,10 +9935,6 @@ func (v *Vault) Validate() error {
 	}
 
 	var mErr multierror.Error
-	if len(v.Policies) == 0 {
-		_ = multierror.Append(&mErr, fmt.Errorf("Policy list cannot be empty"))
-	}
```
lgfa29, line 9931: This has to [be checked in the hook now](https://github.com/hashicorp/nomad/pull/18343/files#diff-b0e2575ed29874321dffeb628199cce5e2c9d58e0e1d8842f73d90fcd06386b6R387) because when using identity policies don't need to specified.

nomad/structs/config/vault.go
```diff
@@ -83,6 +84,46 @@ type VaultConfig struct {
...
+	AllowUnauthenticated *bool `mapstructure:"allow_unauthenticated"`
+
+	// TaskTokenTTL is the TTL of the tokens created by Nomad Servers and used
+	// by the client.  There should be a minimum time value such that the client
+	// does not have to renew with Vault at a very high frequency
+	//
+	// Deprecated: Nomad 1.7.0 derives tokens from workload identities that
+	// receive their TTL configuration from the Vault role used. This field
+	// will be removed in a future release.
+	TaskTokenTTL string `mapstructure:"task_token_ttl"`
```
lgfa29, line 126: This was not part of the RFC, but looking at the code it seems to only be used in the legacy flow. I may be mistaken though 😅

nomad/structs/config/vault.go
```diff
@@ -18,44 +18,45 @@ const (
...
-//
-// - Renew Vault tokens/leases.
-//
-// - Pass a token for the Nomad Server to derive sub-tokens.
-//
-// - Create child tokens with policy subsets of the Server's token.
+//   - Renew Vault tokens/leases.
+//   - Pass a token for the Nomad Server to derive sub-tokens.
+//   - Create child tokens with policy subsets of the Server's token.
+//   - Create Vault ACL tokens from workload identity JTWs.
```
tgross, line 24:
```suggestion
//   - Create Vault ACL tokens from workload identity JWTs.
```

nomad/structs/config/vault.go
```diff
@@ -18,44 +18,45 @@ const (
...
 type VaultConfig struct {
+	// Servers and clients fields.
+
+	// Name is used to identity the Vault cluster related to this
```
tgross, line 28:
```suggestion
	// Name is used to identify the Vault cluster related to this
```

nomad/structs/config/vault.go
```diff
@@ -18,44 +18,45 @@ const (
...
+	// use when a job does not define a role in its `vault` block. If this
+	// config value is also unset, the default auth method or cluster global
+	// role is used.
+	//
+	// When not using workload identities, the Nomad servers will derive toknes
+	// using this role. The token given to Nomad does not have to be created
+	// from this role but must have "update" capability on
+	// "auth/token/create/<create_from_role>". If this value is unset and the
+	// token is created from a role, the value is defaulted to the role the
+	// token is from.
```
tgross, line 47:
```suggestion
	// When not using workload identities, the Nomad servers will derive tokens
	// using this role. The Vault token provided to the Nomad server config does 
	// not have to be created from this role but must have "update" capability on
	// "auth/token/create/<create_from_role>". If this value is unset and the
	// token is created from a role, the value is defaulted to the role the
	// token is from.
```

nomad/structs/config/vault.go
```diff
@@ -83,6 +84,46 @@ type VaultConfig struct {
...
 	TLSServerName string `mapstructure:"tls_server_name"`
+
+	// Servers-only fields.
+
+	// UseIdentity defines if workload identities should be used to derive
+	// Vault tokens.
+	//
+	// It is a transitional field used only during the adoption period of
+	// workload identities and will be ignored and removed in future versions.
+	UseIdentity *bool `mapstructure:"use_identity"`
```
tgross, line 95: Arguably this new field could immediately be placed in the "Deprecated" section below.

nomad/job_endpoint_hooks.go
```diff
@@ -307,45 +308,95 @@ func (v *jobValidate) Validate(job *structs.Job) (warnings []error, err error) {
...
+	var mErr *multierror.Error
+
+	// Prefix errors and warnings.
+	defer func() {
+		prefix := fmt.Sprintf("Service %s in %s", s.Name, parent)
+		for i, w := range warnings {
+			warnings[i] = fmt.Errorf("%s %s", prefix, w)
+		}
+		err = multierror.Prefix(mErr, prefix)
+	}()
```
tgross, line 342: Isn't `warnings` always empty in this method? If so, why not use `fmt.Errorf` for the errors and avoid this extra loop, and then call `multierror.Prefix` before the return?

nomad/job_endpoint_hooks.go
```diff
@@ -307,45 +308,95 @@ func (v *jobValidate) Validate(job *structs.Job) (warnings []error, err error) {
...
+	}
+
+	return
+}
+
+func (v *jobValidate) validateVaultIdentity(t *structs.Task) (warnings []error, err error) {
+	var mErr *multierror.Error
+
+	// Prefix errors and warnings.
+	defer func() {
```
tgross, line 363: Similar to the above method, why not use the `fmt.Errorf` with the prefix when we instantiate the error, rather than looping over them again?

tgross (approved):
This looks great @lgfa29! I've left a few remarks but they're mostly style or docstring nitpicks to pick up.

nomad/job_endpoint_hooks.go
lgfa29: Yeah, it's not used. I debated about not even having them (YAGNI), but I wasn't sure if we would need additional validation as the work progresses.

But back to YAGNI, I will remove this 👍

nomad/job_endpoint_hooks.go
lgfa29: At one point I think I had the suffix added by the caller, so I just stuck with it. But yeah, we don't need it.

nomad/job_endpoint_hook_implicit_identities.go
```diff
@@ -74,3 +76,40 @@ func (h jobImplicitIdentitiesHook) handleConsulService(s *structs.Service) {
...
+//     `vault.default_identity` is provided.
+//
+// If the task already has an identity named `vault` it sets the identity name
+// to the expected value.
+func (h jobImplicitIdentitiesHook) handleVault(t *structs.Task) {
+	if !h.srv.config.UseVaultIdentity() {
+		return
+	}
+
+	if t.Vault == nil {
```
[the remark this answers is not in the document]
lgfa29: They're kind of testing different things (`h.src.config.UseVaultIdentity()` is checking server config and `t.Vault` jobspec) so I kept them separated to make the flowchart more explicit, but I don't have any strong preference.

command/agent/config_parse.go
```diff
@@ -358,6 +361,28 @@ func parseVaults(c *Config, list *ast.ObjectList) error {
...
+			if err := hcl.DecodeObject(&m, defaultIdendityBlock.Val); err != nil {
+				return err
+			}
+
+			var defaultIdendity config.WorkloadIdentityConfig
+			if err := mapstructure.WeakDecode(m, &defaultIdendity); err != nil {
+				return err
+			}
+			c.Vaults[v.Name].DefaultIdentity = &defaultIdendity
+		}
```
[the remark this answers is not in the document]
lgfa29: 🤦

nomad/structs/config/vault.go
lgfa29: Hum...good point. I think I think I will leave it it here for now though as it still kind of expected to be set until 1.9 at least.

lgfa29:
It took me a while to realize the test failures are happening because GHA is testing against a merge between this branch and `main`. This causes a bit of a chicken and egg problem where the default Vault identity is generated without a TTL until https://github.com/hashicorp/nomad/pull/18457 is merged, so I will rebase and fix the tests then.

lgfa29:
Added TTL support in https://github.com/hashicorp/nomad/pull/18343/commits/30dc5739545ce6d869fe10c84cff30daf57469e4 and updated the tests.

PR #18345, opened 2023-08-28
client: include response body in output for successful HTTP checks

MattJustMatt:
This brings Nomad's service checks closer in line with those of Consul. Specifically, Consul provides the HTTP response body in the "Output" field of health checks, even those that are passing. 

This is needed for services that query the Nomad alloc checks API for more detailed service health info (for example, we return player count for game servers in our health check and use this to stop jobs that have no players for extended periods of time.)

On a side note, I'd like to think through (and would contribute to) a solid way to query for all health checks for a given service ala Consul's "/health/checks/:service" endpoint. 

I'm learning this is may be a limitation because health checks are only stored on the Nomad client and not replicated. Even an endpoint to get all health checks for a Node would be useful!

Please let me know if I'm trending towards an antipattern with this and similar changes. I'm new to the Hashicorp ecosystem and not sure if we're using health checks in a good way or if there's a better tool.

### DIFF
--- a/.changelog/18345.txt
+++ b/.changelog/18345.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+services: Include response body in output for successful HTTP checks when using Nomad native service discovery
+```
--- a/client/serviceregistration/checks/client.go
+++ b/client/serviceregistration/checks/client.go
@@ -205,16 +205,12 @@ func (c *checker) checkHTTP(ctx context.Context, qc *QueryContext, q *Query) *st
 	switch {
 	case result.StatusCode == http.StatusOK:
 		qr.Status = structs.CheckSuccess
-		qr.Output = "nomad: http ok"
-		return qr
 	case result.StatusCode < http.StatusBadRequest:
 		qr.Status = structs.CheckSuccess
 	default:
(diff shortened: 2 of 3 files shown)


lgfa29 (approved):
Thank you for the PR @MattJustMatt!

The changes LGTM, I just pushed a commit to add a changelog entry 🙂

shoenig:
Hi @MattJustMatt! first, thanks for the PR - it's always great to see folks willing to get into Nomad's code and making changes they would like to see. However after a bit of internal discussion I think we're going to back out this PR, while we consider another architecture for the health check accessibility problem in general. 

I created https://github.com/hashicorp/nomad/issues/18361 to capture what we're currently thinking about. Basically we want to enable users to consume the health check output results without adding load to the Nomad control plane, and without placing any Nomad client in the critical path for consumers of that check data.

We don't currently have an ETA on when that feature would get done, but it's been requested a number of times ever since Nomad first added service discovery.

MattJustMatt:
All makes sense @shoenig and I appreciate you taking a deeper look. Will rethink how we're doing this and give some feedback on the broader issue you opened.

PR #18346, opened 2023-08-28
[ui] Any policy checks in the UI now check for roles' policies as well as token policies

### LINKED ISSUE #17019: ui: display roles in token page
After completing the login, the OIDC flow redirects to `/ui/settings/tokens` which displays some information about the granted token:
Untitled

### Proposal
It would be great to also display which roles the token has assigned, alongside the policies, if any.

### Use-cases
It is currently confusing to our users to be displayed only an empty list of policies when using roles.

### Attempted Solutions
None yet.

philrenaud:
Since a token can have policies directly, or have policies as part of its role, we combine them when determining whether a user is allowed to take a certain action in the UI.

This shows both roles and their policies on the sign-in / profile page, and adds tests to verify that capabilities in policies derived from roles are adhered to in the UI.

image

Resolves #17019


ui/tests/acceptance/regions-test.js
```diff
@@ -194,29 +194,34 @@ module('Acceptance | regions (many)', function (hooks) {
...
-      ,
-      ,
-      ,
-      // License request
-      // Token/policies request
-      // Search feature detection
-      regionsRequest,
-      defaultRegionRequest,
-      ...appRequests
-    ] = server.pretender.handledRequests;
```
philrenaud, line 207: Previously, this was using `[, , , , ]` destructuring to say "We don't care about the first 3 requests for this test", which, while fast, is a bit haphazard, and broke as soon as we added a new request before the one being tested here. The update handles this with `.find` and `.filter`

ui/mirage/scenarios/default.js
```diff
@@ -439,6 +440,168 @@ function policiesTestCluster(server) {
   server.createList('agent', 3, 'withConsulLink', 'withVaultLink');
 }
 
+function rolesTestCluster(server) {
```
philrenaud, line 443: Our mocking library, Mirage, uses "scenarios" to scaffold clusters where you want to reuse the setup between tests. This adds a Roles test cluster scenario, like we'd previously done for variables, services, etc. tests. This (and anything else starting with /mirage) has no user-facing change.

ui/tests/acceptance/token-test.js
```diff
@@ -730,4 +730,73 @@ module('Acceptance | tokens', function (hooks) {
...
     );
   }
+
+  module('Roles', function (hooks) {
+    // Set up a token with a role
+    hooks.beforeEach(function () {
+      window.localStorage.clear();
+      window.sessionStorage.clear();
+      faker.seed(1);
+      allScenarios.rolesTestCluster(server);
```
philrenaud, line 740: This is where we call the aforementioned mirage scaffolded cluster scenario. It means that calls below using `db.tokens.findBy()`, etc. will find what they're looking for.

ui/app/routes/application.js
```diff
@@ -56,7 +56,7 @@ export default class ApplicationRoute extends Route {
         this.controllerFor('application').set('error', e);
       }
 
-      const fetchSelfTokenAndPolicies = this.get(
+      const fetchSelfTokenAndPolicies = await this.get(
```
philrenaud, line 59: Fetching policies is now a potentially multi-step process (if they have roles, its policies are subsequently fetched), so we use `await`.

pkazmierczak (approved):
LGTM!

While not a front end dev, I don't feel confident to review the code thoroughly, but I read it and it looks good to me. I also tested the feature locally and it works as advertised.

Issue #10324, opened 2021-04-07
CSI: implement ExpandVolume


tgross:
Splitting out of https://github.com/hashicorp/nomad/issues/8212, as we're most likely going to ship the create and snapshot workflows in Nomad 1.1.0 without having shipped ExpandVolume. WIP PR is https://github.com/hashicorp/nomad/pull/10281

From our internal design doc:
> The Resize workflow tells the Storage Provider to alter the size of an existing volume. Because Nomad does not treat disk resources as a scheduling concern, this doesn’t impact our existing scheduling decisions. Nomad doesn’t implement any of this workflow currently, but most of the RPC/HTTP plumbing can be lifted from the existing Controller and Node RPCs for register/deregister/status.
 > 
> We do need to validate whether the volume has “online” resizing and potentially return an error if the volume is claimed. The volume resizing operations themselves are dependent on the CSI plugin: cloud provider block storage (ex. AWS EBS, Digital Ocean block store) are less likely to have this option, whereas NFS-like or clustered file systems (gluster, Ceph) are more likely. When a volume resize request is sent, we’ll see the following workflow:
> *If the plugin does not support resizing, the user will get an error saying it’s unsupported for the plugin.
> * If the plugin supports resizing but not online resizing, and there are live claims, the user will get an error saying the volume can’t be resized while claimed, and identify the allocations that currently claims it.
> * If the plugin supports resizing including online resizing, the storage backend will get a resize operation. We’ll block on this request. Once it is complete, we’ll return a success or failure message to the user.

Command line will be: `nomad volume resize :volume_id -sizeNG [-secretkey=value]​`
Takes the volume ID and a `-size` flag that accepts the usual size strings (ex ​-size 10G​ for 10GiB).

tgross:
Rather than having a new CLI option for this, I'd like to update the `nomad volume register` and `nomad volume create` commands to automatically detect the case where we're trying to resize and send the `ExpandVolume` if permitted, returning an error if not.

taveraluis:
@tgross is this expected to ship in a recent release? Is this actively being worked on?

tgross:
I have a long-stale PR open for it in #10281 but we're not planning on shipping it in the upcoming Nomad 1.3.0. From a design perspective I want to resolve https://github.com/hashicorp/nomad/issues/11899 first, so that volume commands are more declarative the way that job commands are.

MattRiddell:
[...]

PR #18359, opened 2023-08-29
CSI: Expand Volume

gulducat:
Addresses #10324 with declarative config (instead of a new imperative command/api route).

I still need to update docs, but to trigger a volume increase, modify the [volume specification](https://developer.hashicorp.com/nomad/docs/other-specifications/volume) that was used initially to create/register the volume with a higher [`capacity_min`](https://developer.hashicorp.com/nomad/docs/other-specifications/volume#capacity_min) (and optionally `capacity_max`), then issue `create` or `register` again.

This introduces a comparatively very slow process into the create/register flow (if expand is actually needed), and modifies the behavior of `create` to be more idempotent than it currently is (presently it always issues create to the controller plugin).

The commit history is messy right now; I intend to squash it before merging.

So far this covers `ControllerExpandVolume` to increase volumes outside of the instance (e.g. an AWS EBS volume), but does not yet complete the process (when required) with `NodeExpandVolume` on the node where the volume is mounted.  I think I'll make that a separate PR.


nomad/csi_endpoint.go
```diff
@@ -369,6 +382,24 @@ func (v *CSIVolume) Register(args *structs.CSIVolumeRegisterRequest, reply *stru
...
+// depending on whether requested capacity values have changed.
+func (v *CSIVolume) reconcileVolume(plugin *structs.CSIPlugin, old *structs.CSIVolume, new *structs.CSIVolume) error {
+	// Merge does some validation, before we attempt any potential CSI RPCs,
+	// and mutates `old` with (most of) the values of `new`
+	err := old.Merge(new) // this skips capacity values
+	if err != nil {
+		return err
+	}
+	// expandVolume will mutate `old` with new capacity-related values, if needed.
+	err = v.expandVolume(old, plugin, &csi.CapacityRange{
```
jrasell, line 396: We can also return this function call, to save some lines :)

nomad/csi_endpoint.go
```diff
@@ -1062,20 +1109,37 @@ func (v *CSIVolume) Create(args *structs.CSIVolumeCreateRequest, reply *structs.
...
 	for _, valid := range validatedVols {
-		err = v.createVolume(valid.vol, valid.plugin)
-		if err != nil {
-			multierror.Append(&mErr, err)
+		if valid.current != nil {
+			// reconcile mutable fields
+			cp := valid.current.Copy()
+			err = v.reconcileVolume(valid.plugin, cp, valid.vol)
+			if err != nil {
+				multierror.Append(&mErr, err)
```
jrasell, line 1120: `mErr` is a multierror, so we can get away with using the builtin append here which is more efficient for both time and complexity. This also applies to the lines 1129 and 1141 below.
```suggestion
				mErr.Errors = append(mErr.Errors, err)
```

nomad/csi_endpoint.go
```diff
@@ -1062,20 +1109,37 @@ func (v *CSIVolume) Create(args *structs.CSIVolumeCreateRequest, reply *structs.
...
+			} else {
+				// we merged valid.vol into cp, so update state with the copy
+				regArgs.Volumes = append(regArgs.Volumes, cp)
+			}
+
 		} else {
-			regArgs.Volumes = append(regArgs.Volumes, valid.vol)
+			err = v.createVolume(valid.vol, valid.plugin)
+			if err != nil {
+				multierror.Append(&mErr, err)
```
jrasell, line 1129:
```suggestion
				mErr.Errors = append(mErr.Errors, err)
```

nomad/csi_endpoint.go
```diff
@@ -1118,6 +1182,112 @@ func (v *CSIVolume) createVolume(vol *structs.CSIVolume, plugin *structs.CSIPlug
...
+func (v *CSIVolume) expandVolume(vol *structs.CSIVolume, plugin *structs.CSIPlugin, capacity *csi.CapacityRange) error {
+	if vol == nil || plugin == nil || capacity == nil {
+		return errors.New("unexpected nil value")
+	}
+
+	newMax := capacity.LimitBytes
+	newMin := capacity.RequiredBytes
+
+	// closure for verbose logging, including changes to vol fields
+	logger := func() hclog.Logger { // TODO: keep or nah?
```
jrasell, line 1202: I like the additional detail, especially since CSI can be tricky in the first place. Do we need the closure though? Can we just assign logger via the `.Named().With()` call and then use this newly created sub-logger?

client/structs/csi.go
```diff
@@ -287,6 +287,36 @@ type ClientCSIControllerCreateVolumeResponse struct {
...
+type ClientCSIControllerExpandVolumeRequest struct {
+	ExternalVolumeID string
+	CapacityRange    *csi.CapacityRange
+	Secrets          structs.CSISecrets
+	VolumeCapability *csi.VolumeCapability
+
+	CSIControllerQuery
+}
+
+func (req *ClientCSIControllerExpandVolumeRequest) ToCSIRequest() *csi.ControllerExpandVolumeRequest {
```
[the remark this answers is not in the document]
gulducat: I do see some other nil checks in similar methods in this file, so maybe there's a reason for them?  The only place this one is called (currently) already assumes non-nil to get [`req.PluginID`](https://github.com/hashicorp/nomad/pull/18359/files#diff-e3410354fe147d82ac789ad474f8a7e36c240edee9b13a5b03c0b8d24ad11877R239), so I think possible-nil-ness would be bad there too, unless checked prior to that?  I'm not sure how much our RPC boundaries increase the possibility of that kind of nil pointer...

nomad/csi_endpoint.go
```diff
@@ -1047,7 +1082,19 @@ func (v *CSIVolume) Create(args *structs.CSIVolumeCreateRequest, reply *structs.
...
 		}
 
-		validatedVols = append(validatedVols, validated{vol, plugin})
+		// if the volume already exists, we'll update it instead
+		snap, err := v.srv.State().Snapshot()
+		if err != nil {
+			return err
+		}
+		// current will be nil if it does not exist.
+		current, err := snap.CSIVolumeByID(ws, vol.Namespace, vol.ID)
```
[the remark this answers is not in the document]
gulducat: I will admit that I don't fully understand how WatchSet works.  Based on reading about it just now, I _could_ use it to learn if the result of my query here is subsequently changed by a write, yes?  But I'm not doing that with my `ws`; I just take the result of the query (the "current" volume at the time of the call on this line), and stick with it.

so as it stands, yeah I'm pretty sure I can `nil` this one.  I'll ask internally for some more clarification on this thing in general though, to get a better handle on it.

nomad/csi_endpoint.go
```diff
@@ -1118,6 +1182,112 @@ func (v *CSIVolume) createVolume(vol *structs.CSIVolume, plugin *structs.CSIPlug
...
+// via Nomad client RPC.
+//
+// Note that capacity can only be increased; reduction in size is not possible,
+// and if the volume is already at the desired capacity, no action is taken.
+// vol Capacity-related values are mutated if successful, so callers should
+// pass in a copy, then commit changes to raft.
+func (v *CSIVolume) expandVolume(vol *structs.CSIVolume, plugin *structs.CSIPlugin, capacity *csi.CapacityRange) error {
+	if vol == nil || plugin == nil || capacity == nil {
+		return errors.New("unexpected nil value")
+	}
```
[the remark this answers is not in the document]
gulducat: this should never happen -- this is my fear of panics showing.  I actually wonder if a panic would be preferable? because at least then it'd provide a full stack trace if there is an unexpected bug...

nomad/csi_endpoint.go
gulducat: I made it a closure because the fields on `vol` change over the course of `expandVolume()`, and being a closure allows those changes to be reflected in the log lines.  It felt useful to me, but it's not a common pattern from what I've seen, so I'm on the fence. :shrug:

tgross: Changing the values feels like it's going to be confusing, as the logged values won't be what the user sent as input.

nomad/structs/csi.go
```diff
@@ -6,12 +6,13 @@ package structs
...
 	"errors"
 	"fmt"
-	"maps"
-	"slices"
 	"strings"
 	"time"
 
 	multierror "github.com/hashicorp/go-multierror"
+	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
```
tgross, line 14: I think these are in the stdlib now and there was a PR that got merged removing their use everywhere recently.

nomad/structs/csi.go
```diff
@@ -880,6 +867,20 @@ func (v *CSIVolume) Merge(other *CSIVolume) error {
 	return errs.ErrorOrNil()
 }
 
+// MergeWhileInUse updates fields that may be mutated while the volume is in use.
+func (v *CSIVolume) MergeWhileInUse(other *CSIVolume) error {
+	if v == nil || other == nil {
+		return nil // TODO: error?
```
tgross, line 873: It's probably a programmer's error in the call sites we have, but b/c it's being done in `nomad/structs` we don't have a lot of guarantees someone won't misuse it elsewhere later.

plugins/csi/plugin.go
```diff
@@ -101,6 +105,11 @@ type CSIPlugin interface {
 	// for the given volume.
 	NodeUnpublishVolume(ctx context.Context, volumeID, targetPath string, opts ...grpc.CallOption) error
 
+	// NodeExpandVolume is used to expand a volume. This MUST be called after
+	// any ControllerExpandVolume is called, but only if that RPC indicates
+	// that node expansion is required
+	NodeExpandVolume(ctx context.Context, req *NodeExpandVolumeRequest, opts ...grpc.CallOption) (*NodeExpandVolumeResponse, error)
```
tgross, line 111: Just for future consideration: this was a very large PR already so having all the `NodeExpandVolume` plumbing in this PR but not wiring it up makes the PR that much larger for review.

nomad/csi_endpoint.go
gulducat: I really considered using the (relatively) new stdlib [`errors.Join()`](https://pkg.go.dev/errors#Join) instead, but honestly multierror's error string is still easier to read, IMO.  stdlib's just joins them together with newlines.

nomad/csi_endpoint.go
gulducat: the user's input are here as "new_min" and "new_max" and those do not change along the way.  the `vol` values ("current_min", "current_max", and "capacity") in the happy case would change to align with the "new_*"s

nomad/csi_endpoint.go
```diff
@@ -1118,6 +1182,112 @@ func (v *CSIVolume) createVolume(vol *structs.CSIVolume, plugin *structs.CSIPlug
...
+	cReq := &cstructs.ClientCSIControllerExpandVolumeRequest{
+		ExternalVolumeID: vol.ExternalID,
+		Secrets:          vol.Secrets,
+		CapacityRange:    capacity,
+		VolumeCapability: capability,
+	}
+	cReq.PluginID = plugin.ID
+	cResp := &cstructs.ClientCSIControllerExpandVolumeResponse{}
+
+	logger().Info("expanding volume")
```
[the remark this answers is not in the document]
gulducat: rgr that, I didn't have a good sense of this, and just figured that it won't happen that often, so Info might be fine.  will change to Debug.

nomad/csi_endpoint.go
```diff
@@ -239,6 +242,10 @@ func (v *CSIVolume) pluginValidateVolume(req *structs.CSIVolumeRegisterRequest,
 		return nil, fmt.Errorf("no CSI plugin named: %s could be found", vol.PluginID)
 	}
 
+	if plugin.ControllerRequired && plugin.ControllersHealthy < 1 {
+		return nil, fmt.Errorf("no healthy controllers for CSI plugin: %s", vol.PluginID)
+	}
```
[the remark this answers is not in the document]
gulducat: It does have broader impact than expand, but is not totally unrelated.  Let me know if you'd like me to split this out to its own PR.

I noticed that trying to create/register/expand during a client restart (i.e. the controller was temporarily not registered) would produce a misleading error.  Worse, the error was differently misleading in the different code paths:

* create: `plugin does not support creating volumes`
* create w/ expand: `plugin does not support creating volumes`
* register: `controller validate volume: failed to find instances of controller plugin "csi-plugin-id"` (this one is accurate, but happens later than necessary)
* register w/ expand: `unable to update volume: expand is not implemented by this controller plugin`

With this change, all of them get this same error, and pretty early in the signal chain: `no healthy controllers for CSI plugin: csi-plugin-id`

nomad/structs/csi.go
gulducat: ah I fell down on the wrong side of a rebase I think

nomad/structs/csi.go
gulducat: yeah I wanted to nil check at least, but should I make it an explicit error, like "this is probably a bug" instead of just returning nil?

plugins/csi/plugin.go
gulducat: Ah yeah, these came with the original branch I picked up, but I can see how they'd be distracting here without a full implementation.  I'm happy to strip them out of here if you'd like?

tgross: It's fine, just something to keep in mind for next time you've got a big feature you're trying to land is all.

nomad/structs/csi.go
tgross: Make sense

nomad/state/state_store.go
```diff
@@ -2405,17 +2405,17 @@ func (s *StateStore) UpsertCSIVolume(index uint64, volumes []*structs.CSIVolume)
...
-			// prevent accidentally overwriting important properties, or
-			// overwriting a volume in use
+			// prevent accidentally overwriting important properties.
 			old := obj.(*structs.CSIVolume)
 			if old.ExternalID != v.ExternalID ||
 				old.PluginID != v.PluginID ||
 				old.Provider != v.Provider {
 				return fmt.Errorf("volume identity cannot be updated: %s", v.ID)
 			}
-			s.CSIVolumeDenormalize(nil, old.Copy())
```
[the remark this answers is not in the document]
tgross: `CSIVolumeDenormalize` is what makes sure the allocs are populated in the volume. Looks like we needed those for the `InUse` method call to verify we weren't changing something that was in use. Which is now ok so we can safely get rid of `CSIVolumeDenormalize`.

_But_ now that's making me realize this use of `MergeWhileInUse` doesn't make sense at all. We don't know that `old` is in use without checking `InUse`. And even if we did, the merge is mutating `old` which we never use again except to get the `CreateIndex`.

nomad/csi_endpoint.go
tgross: Right, but if we're returning an error then we'll have removed the information about what we did.

nomad/csi_endpoint.go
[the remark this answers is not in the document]
tgross: Ok, makes sense. This isn't strictly a bugfix, right? So it doesn't need backporting (which would need to get pulled out to its own PR).

nomad/state/state_store.go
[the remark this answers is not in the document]
gulducat: > `CSIVolumeDenormalize` is what makes sure the allocs are populated in the volume.

Where is it actually populating something here?  I don't see anything persistent from this call to anywhere outside of it...

Part of `MergeWhileInUse` confusion might be a naming issue? What it does is "merge stuff that is safe to merge, even while in use" whether it actually is in use or not.

But you're totally right about `old` not really being used, that's a mistake.  The intention was to take the current ("old") volume, and only update it with the ok-to-change values, instead of trying to denylist things that can't change.  I missed the extra step though that reassigns `*v = *old`

nomad/csi_endpoint.go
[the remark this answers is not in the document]
gulducat: Good point.  Strictly speaking, one may consider the misleading error text to be a bug, but to my knowledge this doesn't add or remove any error _conditions,_ only the language presented to the user.

nomad/csi_endpoint.go
gulducat: not sure I follow you here.

tgross: I'm guess I'm mostly looking at this from the perspective of error logs, because roughly nobody runs with debug logs in production. And I'm saying that we can populate all the values at the start of the RPC handler, so that when the user (well, administrator) sees the error log they can see "this is what I asked for" and "this was the error" and not some intermediate state where the values are changed from what they knew them to be. The log message itself has the context about whether something has changed or not.

_But_ it turns out you never log errors here so that's totally moot! I'm not sure there's a lot of value in repeating these values across 6 different log messages though... we're sort of using these as a status check and the values won't change until the very last line. So why not put them at the start and stop only?

nomad/state/state_store.go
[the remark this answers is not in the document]
gulducat: By "persistent" I mean I don't see any side-effects at all that reach outside of this function call.  Maybe I'm not following how `old.Copy()` works though?  Is denormalizing `old.Copy()` also affecting `old` itself?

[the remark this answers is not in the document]
gulducat: To clarify a bit, I've gathered (I think) that "denormalizing" is an in-memory thing to enrich data, usually across raft tables (in this case, csi_volumes and alloc tables), but here I'm not seeing how anything is being enriched outside of the call :shrug:

nomad/csi_endpoint.go
gulducat: The values actually change twice along the way
* `vol.RequestedCapacity(Min|Max)` are set right after they are validated as logically coherent
* `vol.Capacity` changes after the controller has done the actual expansion

But as for "status check"s, yeah these are probably more like `Trace`s, really, but I figured this operation will probably happen pretty rarely, and might be nice to be able to debug without all the other trace noise.

All that said, it sounds like the feeling is that this is overkill for not much gain, eh?  I can boil it down to just the requested values and structure it more straightforwardly.

nomad/state/state_store.go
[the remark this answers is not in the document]
tgross: _Oh._ I see. We're not returning the value by doing `old = s.CSIDenormalize(nil, old.Copy())`. That means `old.InUse()` is operating over the wrong object (the original old, and not the copy). That'd be alarming but it doesn't look like `InUse` cares about objects in the `PastClaims` field, and the length of those `ReadAllocs` and `WriteAllocs` doesn't change, so it doesn't result in a bug.

That's very strange. Looks like it the denormalize was added [here](https://github.com/hashicorp/nomad/commit/7d0f87b9106a8557896c6cc09576c846235495a5#diff-53c7d524897a192ece819816a9dcf006dee46747fbb853e7f638601ea556a72cR2172) but it never was useful to be doing. We probably should remove it from the backports then too, just so we don't do this work for nothing and leave this landmine of confusion in the codebase. :blush:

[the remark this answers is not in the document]
gulducat: Nice, phew, reassuring that I wasn't just missing something!

I'm open to some splitting-out for backporting separate PR(s), but not sure where to cleave the changes.  Maybe everything from Register/Create into reconcileVolume, but then exclude expandVolume and everything after that?  That could include the ["no healthy controllers"](https://github.com/hashicorp/nomad/pull/18359#discussion_r1318826936) change too.  Or, I could remove just this one line...?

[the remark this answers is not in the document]
tgross: I would land this in main and then do a brand new PR that targets the most recent backport branch (`release/1.6.x`) with just the backportable fixes. And then once that merges to `release/1.6.x`, cherry-pick it to the other backport branches. That way you don't need to mess with this branch at all and it lands as once nice clean body of work in `main`.

nomad/state/state_store.go
```diff
@@ -2421,21 +2421,21 @@ func (s *StateStore) UpsertCSIVolume(index uint64, volumes []*structs.CSIVolume)
...
-			if old.InUse() {
-				return fmt.Errorf("volume cannot be updated while in use")
-			}
 
-			v.CreateIndex = old.CreateIndex
+			// Update fields that are safe to change while volume is being used.
+			if err := old.UpdateSafeFields(v); err != nil {
+				return fmt.Errorf("unable to update in-use volume: %w", err)
+			}
+			*v = *old
```
tgross, line 2436: Both objects are pointers, and you're just looking to rebind the value, so there's no need to do this.

```suggestion
			v = old
```

tgross (approved):
LGTM!

PR #18363, opened 2023-08-29
config: fix identity config for Consul service

lgfa29:
Rename the agent configuraion for workload identity to `WorkloadIdentityConfig` to make its use more explicit and remove the `ServiceName` field since it is never expected to be defined in a configuration file.

Also update the job mutation to inject a service identity following these rules:

1. Don't inject identity if `consul.use_identity` is false.
2. Don't inject identity if `consul.service_identity` is not specified.
3. Don't inject identity if service provider is not `consul`.
4. Set name and service name if the service specifies an identity.
5. Inject `consul.service_identity` if service does not specify an identity.


pkazmierczak (approved):
LGTM, thanks Luiz!

PR #18366, opened 2023-08-30
Adding node_pool to job key validation

aminovpavel:
Issue where problem was found - https://github.com/hashicorp/levant/issues/499

### DIFF
--- a/.changelog/18366.txt
+++ b/.changelog/18366.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+jobspec: add `node_pool` as a valid field
+```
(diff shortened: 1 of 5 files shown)


lgfa29 (approved):
Thank you for the PR @aminovpavel!

I pushed a couple of commits to add the `mapstructure` label to the `NodePool` field because it's spelled differently (`node_pool`) in the jobspec. I also added a test case and a changelog entry.

lgfa29:
Linter error was fixed in https://github.com/hashicorp/nomad/pull/18480

PR #18370, opened 2023-08-31
client: new Consul client

pkazmierczak:
This PR introduces a new Consul client that returns SI tokens based on requests that contain JWTs. The new client does _not_ support token renewals (Consul does not support this) and does not integrate with any hooks at this point (there will be follow-up PRs). 


client/consul/consul.go
```diff
@@ -29,3 +36,101 @@ type ServiceIdentityAPI interface {
...
+	Role           string
+	AuthMethodName string
+}
+
+// ConsulClient is the interface that the nomad client uses to interact with
+// Consul.
+type ConsulClient interface {
+	// DeriveSITokenWithJWT contacts the nomad server and fetches wrapped tokens
+	// for a set of tasks. The wrapped tokens will be unwrapped using consul and
+	// returned.
```
tgross, line 52: I think there's some copypaste from the existing code in this comment? In this method we have the JWT already so we're contacting Consul directly, not the Nomad server. And I'm not sure what "unwrapped using consul" means here

client/consul/consul.go
```diff
@@ -29,3 +36,101 @@ type ServiceIdentityAPI interface {
...
+	DeriveSITokenWithJWT(map[string]JWTLoginRequest) (map[string]string, error)
+}
+
+type consulClient struct {
+	// tokenDeriver is a function pointer passed in by the client to derive
+	// tokens by making RPC calls to the nomad server.
+	tokenDeriver TokenDeriverFunc
+
+	// client is the API client to interact with vault
```
tgross, line 61:
```suggestion
	// client is the API client to interact with consul
```

client/consul/consul.go
```diff
@@ -29,3 +36,101 @@ type ServiceIdentityAPI interface {
...
+	// DeriveSITokenWithJWT contacts the nomad server and fetches wrapped tokens
+	// for a set of tasks. The wrapped tokens will be unwrapped using consul and
+	// returned.
+	DeriveSITokenWithJWT(map[string]JWTLoginRequest) (map[string]string, error)
+}
+
+type consulClient struct {
+	// tokenDeriver is a function pointer passed in by the client to derive
+	// tokens by making RPC calls to the nomad server.
+	tokenDeriver TokenDeriverFunc
```
tgross, line 59: It's not super clear to me what the `tokenDeriver` is doing here. Is this going to be how the existing code calls the `DeriveSIToken` once this new client is wired up in a future PR?

client/consul/consul.go
```diff
@@ -29,3 +36,101 @@ type ServiceIdentityAPI interface {
...
+	c := &consulClient{
+		config:       config,
+		logger:       logger,
+		tokenDeriver: tokenDeriver,
+	}
+
+	// if UseIdentity is unset of set to false, return an empty client
+	if config.UseIdentity == nil || !*config.UseIdentity {
+		return nil, nil
+	}
```
tgross, line 87: Returning `nil` seems to disagree with having the legacy token deriver as a field of the client struct. Should we `return c, nil` instead? If not, should we move this above everything else so we don't allocate a `consulClient` struct pointer?

client/consul/consul.go
```diff
@@ -29,3 +36,101 @@ type ServiceIdentityAPI interface {
...
+				"failed to authenticate with consul for identity %s: %v", k, err,
+			))
+			continue
+		}
+
+		tokens[k] = t.SecretID
+	}
+
+	if err := mErr.ErrorOrNil(); err != nil {
+		return nil, err
```
tgross, line 132: Is there any value in returning `tokens, err` so that we can return _some_ tokens if we only get an error on one of the logins?

tgross (approved):
LGTM! I've left some comments/questions but if you either fix those or explain why I'm wrong about them (likely! :grinning: ) then I'm :+1: to merge.

client/consul/consul.go
pkazmierczak: good point! Fixed in [4c925f3](https://github.com/hashicorp/nomad/pull/18370/commits/4c925f3a61b8b612482abe61fc7013dab06fd041)

client/consul/consul.go
pkazmierczak: I moved it up in [4c925f3](https://github.com/hashicorp/nomad/pull/18370/commits/4c925f3a61b8b612482abe61fc7013dab06fd041),  I think there's no use in returning a client if the feature is disabled.

client/consul/consul.go
pkazmierczak: Sloppy leftover, removed in [4c925f3](https://github.com/hashicorp/nomad/pull/18370/commits/4c925f3a61b8b612482abe61fc7013dab06fd041)

client/consul/consul.go
pkazmierczak: Changed in [4c925f3](https://github.com/hashicorp/nomad/pull/18370/commits/4c925f3a61b8b612482abe61fc7013dab06fd041)

client/consul/consul.go
pkazmierczak: This and other copypasta errors are my bad—I went back and forth a few times on this PR because I wasn't sure how to handle token renewals/re-issues, and there's some leftover mess. Fixed in [4c925f3](https://github.com/hashicorp/nomad/pull/18370/commits/4c925f3a61b8b612482abe61fc7013dab06fd041)

PR #18388, opened 2023-09-05
test: migrate TLS tests to new Nomad command.

jrasell:
This migrates the TLS test data to use certificates created by the Nomad CLI. This initially fixes test failures caused by expired TLS certificates. It also means we now have a single method for these across all active release branches.

Target: 1.4.x
It fixes test failures seen within https://github.com/hashicorp/nomad/pull/18385


gulducat (approved):
Appears to line up with main, and other active release branches :+1:

pkazmierczak (approved):
LGTM!

PR #18392, opened 2023-09-05
fingerprint: add support for fingerprinting multiple Consul clusters

tgross:
Add fingerprinting we'll need to accept multiple Consul clusters in upcoming Nomad Enterprise features. The fingerprinter will create a map of Consul clients by cluster name. In Nomad CE, all but the default cluster will be ignored and there will be no visible behavior change.

_Note: similar to https://github.com/hashicorp/nomad/pull/18253, I'll need to prep an ENT PR to go with this one_

### DIFF
--- a/client/fingerprint/consul.go
+++ b/client/fingerprint/consul.go
@@ -10,14 +10,12 @@ import (
 	"time"
 
 	consulapi "github.com/hashicorp/consul/api"
+	"github.com/hashicorp/go-hclog"
 	log "github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/go-version"
 	agentconsul "github.com/hashicorp/nomad/command/agent/consul"
-)
-
-const (
-	consulAvailable   = "available"
-	consulUnavailable = "unavailable"
+	"github.com/hashicorp/nomad/nomad/structs/config"
 )
 
 var (
@@ -30,10 +28,14 @@ var (
 
 // ConsulFingerprint is used to fingerprint for Consul
 type ConsulFingerprint struct {
-	logger     log.Logger
-	client     *consulapi.Client
-	lastState  string
-	extractors map[string]consulExtractor
+	logger log.Logger
+	states map[string]*consulFingerprintState
+}
+
+type consulFingerprintState struct {
+	client      *consulapi.Client
+	isAvailable bool
+	extractors  map[string]consulExtractor
 }
 
 // consulExtractor is used to parse out one attribute from consulInfo. Returns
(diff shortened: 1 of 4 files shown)


client/fingerprint/consul.go
```diff
@@ -88,46 +109,61 @@ func (f *ConsulFingerprint) Periodic() (bool, time.Duration) {
 	return true, 15 * time.Second
```
tgross, line 109: Unlike https://github.com/hashicorp/nomad/pull/18253, the Consul fingerprint doesn't back off once we've got a successful fingerprint. We probably should do that here too, but I don't want to change the existing behavior in this PR.

client/fingerprint/vault.go
```diff
@@ -17,20 +17,15 @@ import (
 	vapi "github.com/hashicorp/vault/api"
 )
 
-const (
-	vaultAvailable   = "available"
-	vaultUnavailable = "unavailable"
-)
```
tgross, line 23: Note to reviewers: this was dead code after https://github.com/hashicorp/nomad/pull/18253 that I missed

client/fingerprint/consul.go
```diff
@@ -88,46 +109,61 @@ func (f *ConsulFingerprint) Periodic() (bool, time.Duration) {
...
+			fmt.Sprintf("consul.%s.grpc", cfg.Name):          cfs.grpc(consulConfig.Scheme, logger),
+			fmt.Sprintf("consul.%s.ft.namespaces", cfg.Name): cfs.namespaces,
 		}
 	}
 
 	return nil
 }
 
-func (f *ConsulFingerprint) query(resp *FingerprintResponse) agentconsul.Self {
+func (cfs *consulFingerprintState) query(resp *FingerprintResponse, logger hclog.Logger) agentconsul.Self {
```
jrasell, line 157: This already existed but `resp` is not used within this function and the function doesn't seem to be implementing a specific interface, could we therefore remove the parameter?

client/fingerprint/consul.go
```diff
@@ -88,46 +109,61 @@ func (f *ConsulFingerprint) Periodic() (bool, time.Duration) {
...
 	// We'll try to detect consul by making a query to to the agent's self API.
 	// If we can't hit this URL consul is probably not running on this machine.
-	info, err := f.client.Agent().Self()
+	info, err := cfs.client.Agent().Self()
 	if err != nil {
 		// indicate consul no longer available
-		if f.lastState == consulAvailable {
-			f.logger.Info("consul agent is unavailable")
+		if cfs.isAvailable {
+			logger.Info("consul agent is unavailable")
```
jrasell, line 164: Is it worth adding the error to this log line?

client/fingerprint/consul_ce.go
```diff
@@ -0,0 +1,23 @@
...
+
+//go:build !ent
+
+package fingerprint
+
+import "github.com/hashicorp/nomad/nomad/structs/config"
+
+// consulConfigs returns the set of Consul configurations the fingerprint needs
+// to check. In Nomad CE we only check the default Consul.
+func (f *ConsulFingerprint) consulConfigs(req *FingerprintRequest) map[string]*config.ConsulConfig {
```
jrasell, line 12: Is it worth adding a test to cover this?

jrasell (approved):
LGTM, some minor suggestions but nothing blocking.

client/fingerprint/consul.go
pkazmierczak: Is there a reason for this? I mean if we fingerprint successfully, we should back off.

pkazmierczak (approved):
LGTM!

client/fingerprint/consul.go
```diff
@@ -43,29 +45,48 @@ type consulExtractor func(agentconsul.Self) (string, bool)
...
+		err := f.fingerprintImpl(cfg, nil, resp)
+		if err != nil {
+			mErr = multierror.Append(mErr, err)
+		}
+	}
+
+	return mErr.ErrorOrNil()
+}
+
+func (f *ConsulFingerprint) fingerprintImpl(cfg *config.ConsulConfig, req *FingerprintRequest, resp *FingerprintResponse) error {
```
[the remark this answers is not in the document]
tgross: Good catch. That was some leftovers from debugging I think :blush:

client/fingerprint/consul.go
tgross: > Is there a reason for this?

If I recall correctly, the original reasoning was that we would change the client's fingerprint if Consul became unavailable. We no longer do that as of [1.4.0](https://developer.hashicorp.com/nomad/docs/upgrade/upgrade-specific#removing-vault-consul-from-clients) in order to avoid a storm of `Node.Register` RPCs and tons of load on the Nomad servers whenever someone's Consul cluster is flapping. This had been implicated in a few major customer incidents!

> I mean if we fingerprint successfully, we should back off.

It should be safe for us to do so now. I just didn't want to include that in this PR because I want to change one thing at a time. I'll make a follow-up PR for that.

client/fingerprint/consul.go
tgross: Probably can't hurt. Will do.

client/fingerprint/consul_ce.go
tgross: There's not a lot of logic here. I'll add it to the ENT PR maybe that does more work?

PR #18393, opened 2023-09-05
fix panic from zero JobTrackedVersions config

gulducat:
that occurred during server fsm restore, which later produced a negative slice index when trying to upsertJobVersion.

The bug appears to have been introduced in #17939 where [the risk was identified](https://github.com/hashicorp/nomad/pull/17939#discussion_r1268092633) but not quite covered all the way.

Stack trace of the panic:

```
==> Starting Nomad agent...
panic: runtime error: index out of range [-1]

goroutine 115 [running]:
github.com/hashicorp/nomad/nomad/state.(*StateStore).upsertJobVersion(0xc000ecedb0, 0x21da, 0xc000c50b40, 0xc000d70100)
        github.com/hashicorp/nomad/nomad/state/state_store.go:2068 +0x370
github.com/hashicorp/nomad/nomad/state.(*StateStore).upsertJobImpl(0xc00146cd80?, 0x21da, 0x0, 0xc000c50b40, 0x0, 0xc000d70100)
        github.com/hashicorp/nomad/nomad/state/state_store.go:1772 +0x6f5
github.com/hashicorp/nomad/nomad/state.(*StateStore).UpsertJobTxn(...)
        github.com/hashicorp/nomad/nomad/state/state_store.go:1691
github.com/hashicorp/nomad/nomad.(*nomadFSM).handleJobDeregister(0xc000a80ee0, 0x21da, {0xc000eb1767, 0x3}, {0xc000eb1770, 0x7}, 0x0, 0x0, 0xc000d70100)
        github.com/hashicorp/nomad/nomad/fsm.go:843 +0x4bc
github.com/hashicorp/nomad/nomad.(*nomadFSM).applyDeregisterJob.func1(0xc000d67338?)
        github.com/hashicorp/nomad/nomad/fsm.go:729 +0x4d
github.com/hashicorp/nomad/nomad/state.(*StateStore).WithWriteTransaction(0xc0002514a1?, 0x5c?, 0x15c?, 0xc0012d5b88)
        github.com/hashicorp/nomad/nomad/state/state_store.go:6477 +0x6b
github.com/hashicorp/nomad/nomad.(*nomadFSM).applyDeregisterJob(0xc000a80ee0, 0xda?, {0xc0002514a1, 0x15c, 0x15c}, 0x21da)
        github.com/hashicorp/nomad/nomad/fsm.go:728 +0x152
github.com/hashicorp/nomad/nomad.(*nomadFSM).Apply(0xc000a80ee0, 0xc001226960)
        github.com/hashicorp/nomad/nomad/fsm.go:242 +0x295
github.com/hashicorp/raft.(*Raft).runFSM.func1(0xc0009e59f0)
        github.com/hashicorp/raft@v1.5.0/fsm.go:101 +0x224
github.com/hashicorp/raft.(*Raft).runFSM.func2({0xc000d6a600, 0x40, 0x4?})
        github.com/hashicorp/raft@v1.5.0/fsm.go:124 +0x462
github.com/hashicorp/raft.(*Raft).runFSM(0xc00050fb80)
        github.com/hashicorp/raft@v1.5.0/fsm.go:240 +0x3e8
github.com/hashicorp/raft.(*raftState).goFunc.func1()
        github.com/hashicorp/raft@v1.5.0/state.go:149 +0x53
created by github.com/hashicorp/raft.(*raftState).goFunc in goroutine 1
        github.com/hashicorp/raft@v1.5.0/state.go:147 +0x79
```

### DIFF
--- a/nomad/fsm.go
+++ b/nomad/fsm.go
@@ -1513,10 +1513,11 @@ func (n *nomadFSM) restoreImpl(old io.ReadCloser, filter *FSMFilter) error {
 
 	// Create a new state store
 	config := &state.StateStoreConfig{
-		Logger:          n.config.Logger,
-		Region:          n.config.Region,
-		EnablePublisher: n.config.EnableEventBroker,
-		EventBufferSize: n.config.EventBufferSize,
+		Logger:             n.config.Logger,
+		Region:             n.config.Region,
+		EnablePublisher:    n.config.EnableEventBroker,
+		EventBufferSize:    n.config.EventBufferSize,
+		JobTrackedVersions: n.config.JobTrackedVersions,
 	}
 	newState, err := state.NewStateStore(config)
 	if err != nil {
--- a/nomad/state/state_store.go
+++ b/nomad/state/state_store.go
@@ -105,6 +105,13 @@ type StateStoreConfig struct {
 	JobTrackedVersions int
 }
 
+func (c *StateStoreConfig) Validate() error {
+	if c.JobTrackedVersions <= 0 {
+		return fmt.Errorf("JobTrackedVersions must be positive; got: %d", c.JobTrackedVersions)
+	}
+	return nil
+}
+
 // The StateStore is responsible for maintaining all the Nomad
 // state. It is manipulated by the FSM which maintains consistency
 // through the use of Raft. The goals of the StateStore are to provide
@@ -139,6 +146,10 @@ func (a *streamACLDelegate) TokenProvider() stream.ACLTokenProvider {
 
 // NewStateStore is used to create a new state store
 func NewStateStore(config *StateStoreConfig) (*StateStore, error) {
+	if err := config.Validate(); err != nil {
+		return nil, err
+	}
+
 	// Create the MemDB
 	db, err := memdb.NewMemDB(stateStoreSchema())
 	if err != nil {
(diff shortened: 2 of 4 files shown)


nomad/fsm.go
```diff
@@ -1513,10 +1513,11 @@ func (n *nomadFSM) restoreImpl(old io.ReadCloser, filter *FSMFilter) error {
...
 	config := &state.StateStoreConfig{
-		Logger:          n.config.Logger,
-		Region:          n.config.Region,
-		EnablePublisher: n.config.EnableEventBroker,
-		EventBufferSize: n.config.EventBufferSize,
+		Logger:             n.config.Logger,
+		Region:             n.config.Region,
+		EnablePublisher:    n.config.EnableEventBroker,
+		EventBufferSize:    n.config.EventBufferSize,
+		JobTrackedVersions: n.config.JobTrackedVersions,
```
gulducat, line 1520: this line is the actual bugfix

nomad/state/state_store.go
```diff
@@ -139,6 +146,10 @@ func (a *streamACLDelegate) TokenProvider() stream.ACLTokenProvider {
 
 // NewStateStore is used to create a new state store
 func NewStateStore(config *StateStoreConfig) (*StateStore, error) {
+	if err := config.Validate(); err != nil {
+		return nil, err
+	}
```
tgross, line 151: We don't have very good existing patterns around parsing configuration in a centralized place but doing validation all the way down here in `NewStateStore` seems like it's "late". We've already started RPC handlers, Vault and Consul clients, etc, etc. by the time we instantiate the state store. Wouldn't it make more sense to validate the config when we parse it?

tgross (approved):
LGTM with the caveat about where we're validating.

nomad/state/state_store.go
gulducat: I agree in general yep, and that is already being done [here in agent.go](https://github.com/devashishTaneja/nomad/blob/f32dffc136107e493cb4ea5d0c3cc3fd26779306/command/agent/agent.go#L346-L351) from the previous PR, but the bug arose from a subsequent config hand-off that was missed along the way.

This error _should_ not get hit, due to the config hand-off being filled in now, so this is just an extra guardrail around a sensitive area that uses the value to protect against potential future regression, i.e. make noise in tests if someone accidentally breaks it somehow.

tgross: Fair enough. It's cheap and not in a hot path.

PR #18394, opened 2023-09-06
docs: improve diagram on jobspec overview for multi group + tasks.

### LINKED ISSUE #18355: [documentation] Mention that we can run multiple tasks for one job
### Issue

Here https://developer.hashicorp.com/nomad/docs/job-specification was not mentioned that we can configure multiple tasks for one group or multiple groups for one job. Other parts of documentation also does not mention that 'task' could be used multiple times inside group: https://developer.hashicorp.com/nomad/docs/job-specification/task
I found this information useful: https://discuss.hashicorp.com/t/how-to-configue-multitask-job-in-one-file/19466

#### Expected Result
So one `nomad job run XXX` will run multiple services at once.

#### Actual Result
By mistake I configured multiple jobs and run them individually. It's long

jrasell:
closes #18355

### DIFF
--- a/website/content/docs/job-specification/index.mdx
+++ b/website/content/docs/job-specification/index.mdx
@@ -22,12 +22,17 @@ The general hierarchy for a job is:
 ```text
 job
   \_ group
+  |     \_ task
+  |     \_ task
+  |
+  \_ group
+        \_ task
         \_ task
 ```
 
 Each job file has only a single job, however a job may have multiple groups, and
 each group may have multiple tasks. Groups contain a set of tasks that are
-co-located on a machine.
+co-located on the same Nomad client.
 
 ## Example
 


website/content/docs/job-specification/index.mdx
```diff
@@ -22,12 +22,17 @@ The general hierarchy for a job is:
...
+  |
+  \_ group
+        \_ task
         \_ task
 ```
 
 Each job file has only a single job, however a job may have multiple groups, and
 each group may have multiple tasks. Groups contain a set of tasks that are
-co-located on a machine.
+co-located the same Nomad client.
```
pkazmierczak, line 35: I think a preposition is missing here?

> co-located on the same Nomad client?

> co-located in the same Nomad client?

pkazmierczak (approved):
LGTM!

PR #18412, opened 2023-09-06
fingerprint: add warning in CE when there are multiple vaults

tgross:
Nomad CE only supports a single (default) Vault cluster, so log a warning if the user has configured multiple Vaults. Matches the logging done in https://github.com/hashicorp/nomad/pull/18392


jrasell (approved):
LGTM

pkazmierczak (approved):

PR #18413, opened 2023-09-06
[ui] Access Control with Roles and Tokens

philrenaud:
In https://github.com/hashicorp/nomad/pull/18346, we now check for a token's roles' policies (token.roles[].policies[].rules) when determining their permissions in the UI. Previously, the UI only cared about the permissions granted directly from token.policies[].rules.

This PR expands on the new ability to work with roles by adding a management-token-only interface to edit them, along with tokens and policies. Previously, this had only existed for policies.

When a user signs in with a management-type token, they're now greeted with a new Access Control section in the side-nav that introduces them to ACL concepts like tokens, roles, and policies (and in the future, auth methods, etc.)
image

From here, the management user can  create, modify, and delete tokens, roles, and policies via the UI:
image
image

There are a number of side-effects that have been added as a result of the work here:
- An update to our design systems components library (to get access to a new "masked input" component) had a second-order effect of updating our code editor library, causing it to cease working. You'll see changes to `modifier` behaviour sprinkled in here.
- Since the sign-in profile page now does more than simple authorization, we've changed up the top-right-corner drop-down to use an icon and renamed the pages a bit.
- Fixed an equality bug in acl.go when comparing the ExpirationTime of a `PUT` token; go time [should be compared](https://pkg.go.dev/time#Time.Equal) with a built-in `Equals` instead of `==`, etc. 
- and several others that I'll try to explain in-line within the code.


.github/workflows/ember-test-audit.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
         with:
           ref: ${{ github.event.pull_request.base.sha }}
-      - uses: nanasess/setup-chromedriver@6fb8f5ffa6b7dc11e631ff695fbd2fec0b04bb52 # v2.1.1
+      - uses: nanasess/setup-chromedriver@69cc01d772a1595b8aee87d52f53e71b3904d9d0 # v2.1.2
```
philrenaud, line 21: Our GHA checks were failing on this plugin, which was fixed up per https://github.com/nanasess/setup-chromedriver/issues/200

nomad/structs/acl.go
```diff
@@ -363,7 +363,7 @@ func (a *ACLToken) Validate(minTTL, maxTTL time.Duration, existing *ACLToken) er
 		if existing.ExpirationTTL != a.ExpirationTTL {
 			mErr.Errors = append(mErr.Errors, errors.New("cannot update expiration TTL"))
 		}
-		if existing.ExpirationTime != a.ExpirationTime {
+		if !existing.ExpirationTime.Equal(*a.ExpirationTime) {
```
philrenaud, line 366: Context: I found that every time I tried to update a token via http api (to change its policyIDs, for example), I would receive a 500 error saying the expirationTime couldn't be updated / must match.

Traced it back to the fact that you oughtn't do `==` comparisons of time in Go (and many other languages)  because of monotonic clock time: https://pkg.go.dev/time#hdr-Monotonic_Clocks

Go luckily has a time.Equal() built-in method.

ui/app/abilities/role.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+import AbstractAbility from './abstract';
```
philrenaud, line 6: This is a `ability` file in Ember. This lets use use syntax like `{{#if can "delete role"}}`, etc. elsewhere. In the case of roles/tokens/policies, all abilities are restricted to management tokens, but this may not always be the case.

ui/app/adapters/role.js
```diff
@@ -6,7 +6,17 @@
...
   namespace = namespace + '/acl';
+
+  urlForCreateRecord(modelName) {
+    let baseUrl = this.buildURL(modelName);
+    return singularize(baseUrl);
+  }
+
+  urlForDeleteRecord(id) {
+    return this.urlForUpdateRecord(id, 'role');
+  }
```
philrenaud, line 21: Ember adapters let us define the PUTs and DELETEs and etc. for a given resource; here, we're saying "The API expects list operations at /roles, and individual operations at /role"

ui/app/components/access-control-subnav.js
```diff
@@ -0,0 +1,13 @@
...
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+import Component from '@ember/component';
+import { tagName } from '@ember-decorators/component';
+import { inject as service } from '@ember/service';
+
+@tagName('')
+export default class AccessControlSubnav extends Component {
+  @service keyboard;
```
philrenaud, line 12: Adds `shift + arrowkey` nav to the new section subnav

ui/app/components/policy-editor.hbs
```diff
@@ -34,7 +33,7 @@
           mode="ruby"
           content=@policy.rules
           onUpdate=this.updatePolicyRules
-          autofocus=(not @policy.isNew)
+          autofocus=false
```
philrenaud, line 36: We decided to stop auto-focusing Edit Policy/Token/Role pages after internal feedback, as being automatically inside a textbox prevents using keyboard navigation to go back, etc.

ui/app/components/policy-editor.hbs
```diff
@@ -55,12 +54,12 @@
...
-				type="submit"
-			>
-				Save Policy
-			</button>
+		<Hds::Button
+			@text="Save Policy"
+			@type="submit"
+			data-test-save-policy
+			{{on "click" this.save}}
+		/>
```
philrenaud, line 62: A major part of this PR is adding new elements from our [official design system](https://helios.hashicorp.design/) —these Hds::Buttons are just some of the many throughout.

ui/app/components/profile-navbar-item.hbs
```diff
@@ -4,22 +4,16 @@
...
+  <Hds::Dropdown @color="secondary" class="profile-dropdown"
+    {{keyboard-shortcut menuLevel=true pattern=(array "g" "p") }}
+    as |dd|>
+    <dd.ToggleIcon @color="secondary" @icon="user-circle" @text="user menu" @size="small" />
+    <dd.Title @text="Signed In" />
+    <dd.Description @text={{this.token.selfToken.name}} />
+    <dd.Separator />
+    <dd.Interactive @route="settings.tokens" @text="Profile" />
+    <dd.Interactive {{on "click" this.signOut}} @text="Sign Out" @color="critical" />
+  </Hds::Dropdown>
```
philrenaud, line 16: The new icon-style dropdown in the top menu bar

ui/app/components/role-editor.hbs
```diff
@@ -0,0 +1,75 @@
...
+  </div>
+
+	<footer>
+		{{#if (can "update role")}}
+      <Hds::Button @text="Save Role" @color="primary"
+        {{on "click" this.save}}
+      />
+		{{/if}}
+	</footer>
+</form>
```
philrenaud, line 75: All 3 of roles/tokens/policies have a similar structure:
- a /new route that lets you create a new role/token/policy
- a /:id route that lets you edit an existing role/token/policy
- a `<type>-editor` that both /new and /:id leverage with the form and its functionality, to prevent code duplication

ui/app/controllers/access-control/policies/index.js
```diff
@@ -0,0 +1,50 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
```
philrenaud, line 5: In addition to routes (load the resources we need), and components (shared functionality for reusable pieces of code), there are updates to several "controller" files in this PR.

A controller in Ember is tied to a particular route, and helps define functionality for the template that can be found there.

Here, for example, we hydrate the `policies` list with the count of tokens each has, to make for easier sorting by the end-user. We also provide delete functionality with a notification on success/error, etc.

ui/app/controllers/access-control/policies/policy.js
```diff
@@ -11,7 +11,7 @@ import { tracked } from '@glimmer/tracking';
 import { alias } from '@ember/object/computed';
 import { task } from 'ember-concurrency';
 
-export default class PoliciesPolicyController extends Controller {
+export default class AccessControlPoliciesPolicyController extends Controller {
```
philrenaud, line 14: Before this PR, there was already a /policies route in the app. You'll see a few files like this, where things are moved into the new /access-control/policies section instead.

ui/app/modifiers/code-mirror.js
```diff
@@ -25,8 +25,15 @@ export default class CodeMirrorModifier extends Modifier {
...
-    this._setup();
+  element = null;
+  args = {};
+
+  modify(element, positional, named) {
+    if (!this.element) {
+      this.element = element;
+      this.args = { positional, named };
+      this._setup();
+    }
```
philrenaud, line 36: An update to our design systems library also brought about an update to our code editor library. This `modify` syntax is an update to the older `didInstall` syntax we'd been using before.

ui/app/routes/access-control.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  // Load our tokens, roles, and policies
+  model() {
+    return RSVP.hash({
+      tokens: this.store.findAll('token'),
+      roles: this.store.findAll('role'),
+      policies: this.store.findAll('policy'),
+    });
+  }
```
philrenaud, line 37: This is the main entry-point for all the new sub-pages. As such, we pre-load all the relevant data up front.

There's a world where we may have to paginate this in the future, but for now this approach is sensible.

ui/app/routes/access-control/index.js
```diff
@@ -0,0 +1,3 @@
+import Route from '@ember/routing/route';
+
+export default class AccessControlIndexRoute extends Route {}
```
philrenaud, line 3: TODO: Delete

ui/app/routes/access-control/policies.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+import Route from '@ember/routing/route';
+
+export default class AccessControlPoliciesRoute extends Route {}
```
philrenaud, line 8: TODO: Delete

ui/app/routes/access-control/roles.js
```diff
@@ -0,0 +1,3 @@
+import Route from '@ember/routing/route';
+
+export default class AccessControlRolesRoute extends Route {}
```
philrenaud, line 3: TODO: Delete

ui/app/routes/access-control/roles/new.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  resetController(controller, isExiting) {
+    if (isExiting) {
+      // If user didn't save, delete the freshly created model
+      if (controller.model.role.isNew) {
+        controller.model.role.destroyRecord();
+      }
+    }
+  }
```
philrenaud, line 36: When we hit /new in our new routes, we first generate a new, empty, instance of the resource we're attempting to create (in the browser only; nothing saved until explicitly done by the user).

We can then populate it by way of the form in the UI

If the user leaves before saving the new thing, `resetController` destroys the record so that there aren't garbage collection issues later on.

ui/app/routes/access-control/tokens.js
```diff
@@ -0,0 +1,3 @@
+import Route from '@ember/routing/route';
+
+export default class AccessControlTokensRoute extends Route {}
```
philrenaud, line 3: TODO: Delete

ui/app/serializers/role.js
```diff
@@ -14,4 +14,18 @@ export default class RoleSerializer extends ApplicationSerializer {
...
+      hash.ID = snapshot.id;
+    }
+    hash.Policies = hash.PolicyIDs.map((policy) => {
+      return {
+        Name: policy,
+      };
+    });
+    delete hash.PolicyIDs;
+    return hash;
+  }
```
philrenaud, line 30: The way our API endpoints handle hasMany relationships (a list of IDs) and how the Ember app handles them (as a list of fully-fleshed-out objects) deviate from one another. This serializer layer handles that incongruence when we read/save.

ui/app/styles/components/page-layout.scss
```diff
@@ -17,7 +17,7 @@
     // Defensive styles in case header height goes over 100px, causing
     // the left gutter menu to be on top of the header.
     height: $header-height;
-    overflow: hidden;
+    overflow: visible;
```
philrenaud, line 20: Small fix to let our new profile dropdown work

ui/app/templates/access-control/policies/policy.hbs
```diff
@@ -12,17 +12,9 @@
...
-					@confirmText="Yes, delete"
-					@confirmationMessage="Are you sure?"
-					@awaitingConfirmation={{this.deletePolicy.isRunning}}
-					@onConfirm={{perform this.deletePolicy}}
-					@onPrompt={{this.onDeletePrompt}}
-					@onCancel={{this.onDeleteCancel}}
-				/>
+        <Hds::Button @text="Delete Policy" @color="critical"
+          {{on "click" (perform this.deletePolicy)}}
+        />
```
philrenaud, line 17: In updating to user our new design systems buttons, one trade-off is that two-step buttons (where you first click something, then get an "are you sure? yes/cancel" prompt in its place) aren't supported.

This change has led to a lot of discussion internally, and I will probably try to introduce a two-step button pattern using the new design system soon. For now, however, former "are you sure?" buttons are just a bit more final.

ui/app/templates/access-control/tokens/token.hbs
```diff
@@ -0,0 +1,23 @@
...
+      <Hds::Button @text="Delete Token" @color="critical"
+        {{on "click" (perform this.deleteToken)}}
+      />
+		{{/if}}
+	</h1>
+  <TokenEditor
+    @token={{this.activeToken}}
+    @policies={{this.policies}}
+    @roles={{this.roles}}
+  />
```
philrenaud, line 22: This is how we invoke all the `token-editor`, `role-editor`, `policy-editor`, etc. components.

ui/mirage/config.js
```diff
@@ -592,11 +592,29 @@ export default function () {
     return new Response(403, {}, null);
   });
 
+  this.get('/acl/roles', function ({ roles }, req) {
+    return this.serialize(roles.all());
+  });
+
```
philrenaud, line 598: Everything under /mirage or /tests is not user-facing.

Mirage is our mocking library. We use this to build out demo environments (like on vercel), and to scaffold out test-cases for our acceptance tests.

philrenaud: In particular, the methods you see listed here are meant to mock how Nomad's http api endpoints respond.

website/content/api-docs/acl/roles.mdx
```diff
@@ -79,7 +79,7 @@ $ curl \
 }
 ```
 
-## Update Token
+## Update Role
```
philrenaud, line 82: Tiny little docs change; noticed that this was [misnamed in our api docs](https://developer.hashicorp.com/nomad/api-docs/acl/roles#update-token)

ui/app/adapters/role.js
```diff
@@ -6,7 +6,17 @@
 // @ts-check
```
Juanadelacuesta, line 6: I don't know how the license update affects you here?

philrenaud: I need the license here, and for ts-check to come afterward. ts-check is some type safety in JS.

ui/app/routes/access-control.js
Juanadelacuesta: Won't this call end up being very slow?

ui/app/adapters/role.js
Juanadelacuesta: I meant the BSL  on top of the page :)

ui/app/components/role-editor.js
```diff
@@ -0,0 +1,78 @@
...
+        throw new Error(
+          `Role name must be 1-128 characters long and can only contain letters, numbers, and dashes.`
+        );
+      }
+
+      const shouldRedirectAfterSave = this.role.isNew;
+
+      // TODO: this check currently fires on unsuccessful save. Dont do that!
+      if (this.role.isNew && this.store.peekRecord('role', this.role.name)) {
+        throw new Error(`A role with name ${this.role.name} already exists.`);
```
Juanadelacuesta, line 54: I noticed this while testing :)

ui/app/components/token-editor.hbs
```diff
@@ -0,0 +1,236 @@
...
+    </Hds::Form::Label>
+		<Input
+			data-test-token-name-input
+			@type="text"
+			@value={{@token.name}}
+			class="input"
+		/>
+	</label>
+
+{{!-- TODO: add expiration time here --}}
```
Juanadelacuesta, line 19: Isnt it already there? at least time until expiration

ui/app/components/token-editor.js
```diff
@@ -0,0 +1,112 @@
...
+      await this.activeToken.save();
+
+      this.notifications.add({
+        title: 'Token Saved',
+        color: 'success',
+      });
+
+      if (shouldRedirectAfterSave) {
+        this.router.transitionTo(
+          'access-control.tokens.token',
```
Juanadelacuesta, line 99: Its more obvious it successfully saved if it redirects to overview, no?

ui/app/routes/access-control/tokens/new.js
```diff
@@ -0,0 +1,39 @@
...
+      type: 'client',
+    });
+    return {
+      token,
+      policies: await this.store.findAll('policy'),
+      roles: await this.store.findAll('role'),
+    };
+  }
+
+  resetController(controller, isExiting) {
```
Juanadelacuesta, line 31: Same as above?

ui/app/serializers/token.js
```diff
@@ -20,6 +20,34 @@ export default class TokenSerializer extends ApplicationSerializer {
...
+    const hash = super.serialize(snapshot, options);
+
+    if (snapshot.id) {
+      hash.AccessorID = snapshot.id;
+      // If expirationTimeNanos exists, use that when saving expirationTime for equality check reasons;
+      // see note in token model.
+      hash.ExpirationTime = hash.ExpirationTimeVerbatim || hash.ExpirationTime;
+    }
+
+    // delete hash.ExpirationTime;
```
Juanadelacuesta, line 37: delete?

ui/mirage/factories/token.js
```diff
@@ -41,7 +41,7 @@ export default Factory.extend({
       // Create a special policy with variables rules in place
       if (token.id === '53cur3-v4r14bl35') {
         const variableMakerPolicy = {
-          id: 'Variable Maker',
+          id: 'Variable-Maker',
```
Juanadelacuesta, line 44: why the name change?

philrenaud: The new section allows us to create and edit policy names, and something that was added was a regex name check. Because policies' names are their IDs, spaces (and non-alphanumeric + dash characters) are not allowed. This fixes it and lets us, in the mocked version, save this particular policy (and variable-reader)

ui/app/components/token-editor.hbs
philrenaud: Yep good catch — this was an earlier TODO

ui/app/routes/access-control.js
philrenaud: Not too badly, in testing. But even if it did: this is an administrative page with deliberate inter-twining of the entities we load. I wouldn't want to emphasize performance at the cost of full control here, like I might want to do in a more general-use / read-only page.

Juanadelacuesta:
Amazing work!

Juanadelacuesta (approved):

PR #18419, opened 2023-09-07
client: fix role permission issue with duplicate policies.

### LINKED ISSUE #17201: Nomad ACL roles with duplicated policies
### Nomad version

```
Nomad v1.5.5
```

### Issue
The ACL lookup system fails if a token is issued with two roles containing policies with the same name.

### Reproduction steps
I have two roles `my-role` and `my-role-sre` which share one or more policies with the same name.  
In a simplified setup:
```
~ nomad acl role list
ID                                    Name         Description  Policies
5e437898-384e-328b-9a94-97765cfe90d3  my-role-sre  my-role-sre  cluster-user-readonly,my-namespace-power-user
dc0ec75b-9927-3147-586f-264ebc8dbca5  my-role      my-role      cluster-user-readonly,my-namespace-read-only
~
~ nomad acl policy info cluster-user-readonly   
Name        = cluster-user-readonly
Description = <none>
CreateIndex = 22
ModifyIndex = 45

Rules

# Access to a namespace
namespace "default" {
  policy = "read"
}

~ nomad acl policy info my-namespace-read-only 
Name        = my-namespace-read-only
Description = <none>
CreateIndex = 43
ModifyIndex = 43

Rules

namespace "my-namespace" {
  policy = "read"
}

~ nomad acl policy info my-namespace-power-user 
Name        = my-namespace-power-user
Description = <none>
CreateIndex = 42
ModifyIndex = 42

Rules

namespace "my-namespace" {
  policy = "write"
}

~ nomad namespace status my-namespace
Name            = my-namespace
Description     = <none>
Quota           = <none>
EnabledDrivers  = *
DisabledDrivers = <none>
```

I can then start a test job in `my-namespace` called `test` and issue a token with both roles:
```
nomad acl token create -role-name=my-role -role-name=my-role-sre
```

#### Expected Result

Given the policies defined above I expect to be able to exec in the alloc started by my `test` job:
```
nomad exec -i -t -namespace='my-namespace' -job test echo foo
```

#### Actual Result
Client:
```
failed to exec into task: rpc error: Permission denied
```
Server:
```
    2023-05-16T08:25:33.590+0200 [ERROR] client.rpc: error performing RPC to server: error="rpc error: Permission denied" rpc=ACL.GetPolicies server=127.0.0.1:4647
    2023-05-16T08:25:33.590+0200 [ERROR] client.rpc: error performing RPC to server which is not safe to automatically retry: error="rpc error: Permission denied" rpc=ACL.GetPolicies server=127.0.0.1:4647
```

The "Permission denied" error originates from [here](https://github.com/hashicorp/nomad/blob/release/1.5.5/nomad/acl_endpoint.go#L390) 
```
[...]
         // Generate a set of policy names. This is initially generated from the
	// ACL role links.
	tokenPolicyNames, err := a.policyNamesFromRoleLinks(token.Roles)
	if err != nil {
		return err
	}

	// Add the token policies which are directly referenced into the set.
	tokenPolicyNames.InsertAll(token.Policies)

	// Ensure the token has enough permissions to query the named policies.
	if token.Type != structs.ACLManagementToken && !tokenPolicyNames.ContainsAll(args.Names) {
	}
[...]
```

`tokenPolicyNames` contains the deduplicate list of resolved polices from all the roles assigned to the token, whereas `args.Names` from the RPC call contains the "raw" list.  
In my example:
```
tokenPolicyNames = [cluster-user-readonly my-namespace-power-user my-namespace-read-only]
args.Names = [my-namespace-read-only cluster-user-readonly my-namespace-power-user cluster-user-readonly]
```

Removing the duplicated `cluster-user-readonly` from one of the roles makes it to work as expected but it's not an uncommon pattern to scaffold ACL roles with common "basic read-only" policies.  
Should the client's `resolvePolicies` implement the deduplication? https://github.com/hashicorp/nomad/blob/release/1.5.5/client/acl.go#L181
  
The issue doesn't affect all operations as not all of them rely on the RPC call. `nomad job` works and so does `nomad alloc status`, albeit with an error when retrieving stats:
```
~ nomad alloc status -namespace='my-namespace' 865872cf
ID                  = 865872cf-5afd-5680-45a6-fad231a93f68
Eval ID             = 6980575f
Name                = test.test[0]
Node ID             = a8bb5dbd
Node Name           = agent-1
Job ID              = test
Job Version         = 0
Client Status       = running
Client Description  = Tasks are running
Desired Status      = run
Desired Description = <none>
Created             = 4m ago
Modified            = 3m50s ago
Deployment ID       = 07238550
Deployment Health   = healthy

Couldn't retrieve stats: Unexpected response code: 403 (Permission denied)

Task "test" is "running"
Task Resources:
CPU      Memory  Disk     Addresses
100 MHz  64 MiB  300 MiB

Task Events:
Started At     = 2023-05-16T11:52:47Z
Finished At    = N/A
Total Restarts = 0
Last Restart   = N/A

Recent Events:
Time                       Type        Description
2023-05-16T13:52:47+02:00  Started     Task started by client
2023-05-16T13:52:46+02:00  Task Setup  Building Task Directory
2023-05-16T13:52:46+02:00  Received    Task received by client
```

jrasell:
This change deduplicates the ACL policy list generated from ACL roles referenced within an ACL token on the client.

Previously the list could contain duplicates, which would cause erronous permission denied errors when calling client related RPC/ HTTP API endpoints. This is because the client calls the ACL get policies endpoint which subsequently ensures the caller has permission to view the ACL policies. This check is performed by comparing the requested list args with the policies referenced by the caller ACL token. When a duplicate is present, this check fails, as the check must ensure the slices match exactly.

The linked issue has details on how to reproduce the problem, if readers want to test this before and after the code changes.

Closes #17201

### DIFF
--- a/client/acl_test.go
+++ b/client/acl_test.go
@@ -131,13 +131,13 @@ func TestClient_resolveTokenACLRoles(t *testing.T) {
 	defer cleanup()
 
 	// Create an ACL Role and a client token which is linked to this.
-	mockACLRole := mock.ACLRole()
+	mockACLRole1 := mock.ACLRole()
 
 	mockACLToken := mock.ACLToken()
 	mockACLToken.Policies = []string{}
-	mockACLToken.Roles = []*structs.ACLTokenRoleLink{{ID: mockACLRole.ID}}
+	mockACLToken.Roles = []*structs.ACLTokenRoleLink{{ID: mockACLRole1.ID}}
 
-	err := testServer.State().UpsertACLRoles(structs.MsgTypeTestSetup, 10, []*structs.ACLRole{mockACLRole}, true)
+	err := testServer.State().UpsertACLRoles(structs.MsgTypeTestSetup, 10, []*structs.ACLRole{mockACLRole1}, true)
 	must.NoError(t, err)
 	err = testServer.State().UpsertACLTokens(structs.MsgTypeTestSetup, 20, []*structs.ACLToken{mockACLToken})
 	must.NoError(t, err)
@@ -150,12 +150,32 @@ func TestClient_resolveTokenACLRoles(t *testing.T) {
 	// Test the cache directly and check that the ACL role previously queried
 	// is now cached.
 	must.Eq(t, 1, testClient.roleCache.Len())
-	must.True(t, testClient.roleCache.Contains(mockACLRole.ID))
+	must.True(t, testClient.roleCache.Contains(mockACLRole1.ID))
 
 	// Resolve the roles again to check we get the same results.
 	resolvedRoles2, err := testClient.resolveTokenACLRoles(rootACLToken.SecretID, mockACLToken.Roles)
 	must.NoError(t, err)
 	must.SliceContainsAll(t, resolvedRoles1, resolvedRoles2)
+
+	// Create another ACL role which will have the same ACL policy links as the
+	// previous
+	mockACLRole2 := mock.ACLRole()
+	must.NoError(t,
+		testServer.State().UpsertACLRoles(
+			structs.MsgTypeTestSetup, 30, []*structs.ACLRole{mockACLRole2}, true))
+
+	// Update the ACL token so that it links to two ACL roles, which include
+	// duplicate ACL policies.
+	mockACLToken.Roles = append(mockACLToken.Roles, &structs.ACLTokenRoleLink{ID: mockACLRole2.ID})
+	must.NoError(t,
+		testServer.State().UpsertACLTokens(
+			structs.MsgTypeTestSetup, 40, []*structs.ACLToken{mockACLToken}))
+
+	// Ensure when resolving the ACL token, we are returned a deduplicated list
+	// of ACL policy names.
+	resolvedRoles3, err := testClient.resolveTokenACLRoles(rootACLToken.SecretID, mockACLToken.Roles)
+	must.NoError(t, err)
+	must.SliceContainsAll(t, []string{"mocked-test-policy-1", "mocked-test-policy-2"}, resolvedRoles3)
 }
 
 func TestClient_ACL_ResolveToken_Disabled(t *testing.T) {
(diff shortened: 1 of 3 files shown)


client/acl_test.go
```diff
@@ -150,12 +150,32 @@ func TestClient_resolveTokenACLRoles(t *testing.T) {
...
+	mockACLToken.Roles = append(mockACLToken.Roles, &structs.ACLTokenRoleLink{ID: mockACLRole2.ID})
+	must.NoError(t,
+		testServer.State().UpsertACLTokens(
+			structs.MsgTypeTestSetup, 40, []*structs.ACLToken{mockACLToken}))
+
+	// Ensure when resolving the ACL token, we are returned a deduplicated list
+	// of ACL policy names.
+	resolvedRoles3, err := testClient.resolveTokenACLRoles(rootACLToken.SecretID, mockACLToken.Roles)
+	must.NoError(t, err)
+	must.SliceContainsAll(t, []string{"mocked-test-policy-1", "mocked-test-policy-2"}, resolvedRoles3)
```
gulducat, line 178: wasn't sure of this test method based on the name, but it seems to do just what we want here :+1: 

> [SliceContainsAll](https://pkg.go.dev/github.com/shoenig/test#SliceContainsAll) asserts slice and items contain the same elements, but in no particular order. The number of elements in slice and items must be the same.

client/acl_test.go
```diff
@@ -150,12 +150,32 @@ func TestClient_resolveTokenACLRoles(t *testing.T) {
...
+	must.True(t, testClient.roleCache.Contains(mockACLRole1.ID))
 
 	// Resolve the roles again to check we get the same results.
 	resolvedRoles2, err := testClient.resolveTokenACLRoles(rootACLToken.SecretID, mockACLToken.Roles)
 	must.NoError(t, err)
 	must.SliceContainsAll(t, resolvedRoles1, resolvedRoles2)
+
+	// Create another ACL role which will have the same ACL policy links as the
+	// previous
+	mockACLRole2 := mock.ACLRole()
```
gulducat, line 162: just to log more of my thought process here: `mock.ACLRole()` generates a random `ID` each time, so this one is automatically different-enough from the previous to be separate in state (we're not just upserting the same role).

gulducat (approved):
LGTM!

PR #18425, opened 2023-09-07
ACLs: allow/deny/default config for Consul/Vault clusters by namespace

tgross:
In Nomad Enterprise when multiple Vault/Consul clusters are configured, cluster admins can control access to clusters for jobs via namespace ACLs, similar to how we've done so for node pools. This changeset updates the ACL configuration structs, but doesn't wire them up. I'll follow up with additional PRs for the job admission hooks and documentation. I'll also need a matching PR in the ENT repo for the validation/canonicalization methods.

### DIFF
--- a/nomad/structs/consul.go
+++ b/nomad/structs/consul.go
@@ -3,6 +3,11 @@
 
 package structs
 
+import (
+	"fmt"
+	"regexp"
+)
+
 // Consul represents optional per-group consul configuration.
 type Consul struct {
 	// Namespace in which to operate in Consul.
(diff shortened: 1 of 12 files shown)


pkazmierczak (approved):
LGTM!

tgross:
I've pushed an additional commit here that's going to be needed to support the ENT side, and adds a validation step to the config parsing around the Vault and Consul cluster names.

nomad/structs/consul.go
```diff
@@ -44,6 +49,20 @@ func (c *Consul) Validate() error {
 	return nil
```
tgross, line 49: Note we don't want to do the name validation here because this is allowed to be empty. The job admission hook will handle ensuring the cluster exists.

PR #18426, opened 2023-09-07
fingerprint: backoff on Consul fingerprint after initial success

tgross:
In the original design of Consul fingerprinting, we would poll every period so that we could change the client's fingerprint if Consul became unavailable. As of 1.4.0 (ref #14673) we no longer update the fingerprint in order to avoid excessive `Node.Register` RPCs when someone's Consul cluster is flapping.

This allows us to safely backoff Consul fingerprinting on success, just as we have with Vault.

Ref: https://github.com/hashicorp/nomad/pull/18392#discussion_r1318549264

### DIFF
--- a/.changelog/18426.txt
+++ b/.changelog/18426.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+fingerprint: clients now backoff after successfully fingerprinting Consul
+```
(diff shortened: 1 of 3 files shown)


pkazmierczak (approved):
LGTM, thanks Tim!

PR #18429, opened 2023-09-08
Add functional option to configure max retries

Juanadelacuesta:
Following the Consul example, where the user can configure the number of retries on the calls to the locks endpoint, this PR introduces the same functionality for the locks handler in nomad.

### DIFF
--- a/api/locks.go
+++ b/api/locks.go
@@ -30,7 +30,7 @@ var (
 )
 
 // Locks returns a new handle on a lock for the given variable.
-func (c *Client) Locks(wo WriteOptions, v Variable) (*Locks, error) {
+func (c *Client) Locks(wo WriteOptions, v Variable, opts ...LocksOption) (*Locks, error) {
 
 	if v.Path == "" {
 		return nil, LockNoPathErr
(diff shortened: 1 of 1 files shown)


api/locks.go
```diff
@@ -66,10 +71,22 @@ type Locks struct {
...
 
 	WriteOptions
 }
 
+type LocksOption = func(l *Locks)
+
+// WithMaxRetries allows access to configure the number of max retries the lock
+// handler will perform in case of an expected response while interacting with the
+// locks endpoint.
+func WithMaxRetries(maxRetries int64) LocksOption {
```
tgross, line 84: Because this is a free function and not a method, the name `WithMaxRetries` is going to be in the package scope and not scoped to the stuff in this file. I could see us wanting to have something like this for other options structs in the future, which would make this generic name awkward to workaround. I'd recommend we either name this so that it's scope is obvious, or somehow make it a method on `Locks`

tgross (approved):
Functionality LGTM. I've left a comment about the API design.

api/locks.go
Juanadelacuesta: Makes a lot of sense, let me figure it out :)

PR #18431, opened 2023-09-08
client: split identity_hook across allocrunner and taskrunner

pkazmierczak:
This PR splits `identity_hook` between the `allocrunner` and `taskrunner`. The `allocrunner`-level part of the hook signs each task identity, and the `taskrunner`-level part picks it up and stores secrets for each task. 

This code revamps the `WIDMgr`, which is now split into 2 interfaces: `IdentityManager` which manages renewals of signatures and handles sending updates to subscribers via `Watch` method, and `IdentitySigner` which only does the signing. 

This work is necessary for having a unified Consul login workflow that comes with the new Consul integration. A new, `allocrunner`-level `consul_hook` will now be the only hook doing Consul authentication. 


client/allocrunner/identity_hook.go
```diff
@@ -0,0 +1,248 @@
...
+		var err error
+		signedWIDs, err = h.getIdentities(h.ar.Alloc(), task)
+		if err != nil {
+			return fmt.Errorf("error fetching alternate identities: %w", err)
+		}
+
+		// publish task identities inside hookResources, so that taskrunner
+		// hooks can also use them.
+		h.hookResources.SignedTaskIdentities = make(chan map[string]*structs.SignedWorkloadIdentity, numberOfTasks)
+		h.hookResources.SignedTaskIdentities <- signedWIDs
```
tgross, line 81: This looks like we're creating a channel for each task, which we definitely need. But aren't we overwriting the `hookResources.SignedTaskIdentities` in the process? That means only the channel for the last task will be available at this location by the time the taskrunner `identity_hooks` run. 

(I'm pretty sure that when the other channels fall out of scope they'll get GC'd even though they're full, unlike a leaked goroutine.)

client/client.go
```diff
@@ -335,8 +335,11 @@ type Client struct {
 	// partitions is used for managing cpuset partitioning on linux systems
 	partitions cgroupslib.Partition
 
-	// widmgr retrieves workload identities
-	widmgr *widmgr.WIDMgr
+	// widsigner signs workload identities
+	widsigner widmgr.IdentitySigner
+
+	// widmgr manages workload identities
+	widmgr widmgr.IdentityManager
```
lgfa29, line 342:
```suggestion
```

I think we can remove this one since each allorunner has a `widmgr` now?

pkazmierczak: good point, fixed in [d2fa276](https://github.com/hashicorp/nomad/pull/18431/commits/d2fa276e186e900acea769d7129238cb22fe78ee)

client/allocrunner/identity_hook_test.go
```diff
@@ -0,0 +1,84 @@
...
+		},
+	})
+	must.NoError(t, err)
+
+	start := time.Now()
+	hook := newIdentityHook(logger, allocrunner)
+	must.Eq(t, hook.Name(), "identity")
+	must.NoError(t, hook.Prerun())
+
+	sid, err := hook.widmgr.Get(cstructs.TaskIdentity{TaskName: task.Name, IdentityName: task.Identities[0].Name})
```
tgross, line 71: The `hook.Prerun` runs the widmgr in a goroutine. Do we need to ensure it's run at least once before we call this `Get`?

client/widmgr/widmgr.go
```diff
@@ -4,80 +4,367 @@
...
+
+	// Create a cancel func for watchers to deregister when they exit.
+	cancel := func() {
+		m.watchersLock.Lock()
+		defer m.watchersLock.Unlock()
+
+		delete(m.watchers, id)
+
+		//TODO should we close(c) here? seems sketchy... caller shouldn't try to
+		//recv on c after calling cancel
```
tgross, line 159: We might want to close here but then we'll need to change the `Shutdown` method so we don't close there. Otherwise during shutdown all the `identity_hook`s in the task runner will call their own `Stop` which stops their watch channels. That'd result in a double-close during shutdown (and panic). We probably wouldn't be able to safely do the "so let's just close on double watches" you've done above either.

It's easier to reason about correctness if we make sure that the close can only happen in exactly one place (and one goroutine) per channel. But we have two ways in which we want to stop watching: (1) the task runner has exited because the task has stopped, and (2) the agent itself is shutting down. The current design you've got here handles (2) just fine and then for (1) we delete the channel from the list of watchers and the taskrunner is responsible for making sure it's not trying to poll on it anymore so the channel can be GC'd.

client/widmgr/signer.go
```diff
@@ -0,0 +1,89 @@
...
+//
+// Since a single rejection causes an error to be returned, SignIdentities
+// should currently only be used when requesting signed identities for a single
+// allocation.
+func (s *Signer) SignIdentities(minIndex uint64, req []*structs.WorkloadIdentityRequest) ([]*structs.SignedWorkloadIdentity, error) {
+	args := structs.AllocIdentitiesRequest{
+		Identities: req,
+		QueryOptions: structs.QueryOptions{
+			Region:        s.region,
+			MinQueryIndex: minIndex - 1,
```
tgross, line 64: Why -1?

client/widmgr/widmgr.go
```diff
@@ -4,80 +4,367 @@
...
+	}
+
+	// Prime chan with latest token to avoid a race condition where consumers
+	// could miss a token update between Get and Watch calls.
```
tgross, line 163: I really like how we take advantage of the lock in the widmgr to prime/pop values on the channel and eliminate a whole class of race conditions

client/widmgr/widmgr.go
```diff
@@ -4,80 +4,367 @@
...
+
+		// Success! Set next renewal and reset retries
+		wait = helper.ExpiryToRenewTime(minExp, time.Now, m.minWait)
+		retry = 0
+	}
+}
+
+func (m *WIDMgr) send(id cstructs.TaskIdentity, token *structs.SignedWorkloadIdentity) {
+	m.watchersLock.Lock()
+	defer m.watchersLock.Unlock()
```
tgross, line 354: `send` has only one caller, so it'd be more efficient to lock in the caller and then note that this method must be called while holding the lock (elsewhere with this pattern we've even changed the method name to something like `sendLocked`... unfortunately the type system doesn't give us a lot of help here :crying_cat_face: )

client/allocrunner/identity_hook_test.go
pkazmierczak: we should. Is there a smart way of doing it? Elsewhere in tests I just `time.Sleep(time.Second)` which is... inelegant.

client/widmgr/signer.go
pkazmierczak: I honestly don't know, I lifted this from Schmichael's old code. @schmichael can you shed some light on this?

schmichael: I should have left a comment because this took me way too long to remember. :grimacing: 

Normally blocking queries are used to wait for changes *subsequent* to the `MinQueryIndex` (the `?index=...` in HTTP APIs). So if an Allocation was created at index 123, you would want to block with index=123 to ensure the query returned only when the allocation was modified *after* creation.

This isn't our normal blocking query though. I think it's the first time we've taken a different approach:

Nomad determines the validity of a workload identity based on the liveness of the workload (allocation) it represents. Therefore Clients requesting signed workload identities via the `Alloc.SignIdentities` RPC want to ensure the Server doesn't erroneously reject the request because the Server's raft index is out of stale (either due to the server being mid-restore or just due to "fun" distributed systems race conditions).

Therefore the Client sets `MinQueryIndex = alloc.CreateIndex - 1` to instruct the Server to block at least _until the Allocation is created._ The Client doesn't care about modifications after that (or more accurately: modifications after that, like `DesiredStatus=stop`, will be handled by *other* RPCs the Client makes).

So this RPC, unlike all other RPCs AFAIK, doesn't care about "subsequent modifications" after an index. It's just trying to ensure the Server's state isn't too stale to know about the alloc.

...now to figure out how to describe this succinctly in a comment. :face_with_spiral_eyes:

client/allocrunner/identity_hook_test.go
tgross: Usually I tried to check for a side effect. Ex. because this is a unit test you could try to grab the internal lock and check whatever state that `Get` expects before calling `Get`. Or you could use `must.Wait` to retry the `Get` until the state converges the way you expect.

client/widmgr/widmgr.go
```diff
@@ -4,80 +4,367 @@
...
-		return nil, fmt.Errorf("expected %d signed identities but received %d", exp, act)
+	// Pop any unreceived tokens
+	select {
+	case <-c:
+	default:
 	}
 
-	return reply.SignedIdentities, nil
+	// Send new token, should never block since this is the only sender and watchersLock is held
+	c <- token
```
lgfa29, line 369: I mentioned earlier, but I think this blocks if callers fail to call the `stop` function, which would block the entire `renew` loop. Would it be safe to have a `select` and `default` here as well?

lgfa29 (commented):
Adding some early review. I'm still digging through the test failures.

client/widmgr/mock.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/go-jose/go-jose/v3"
+	"github.com/go-jose/go-jose/v3/jwt"
+
+	"github.com/hashicorp/nomad/helper/uuid"
+	"github.com/hashicorp/nomad/nomad/structs"
+)
+
+// MockWIDMgr allows TaskRunner unit tests to avoid having to setup a Server,
+// Client, and Allocation.
+type MockWIDMgr struct {
```
lgfa29, line 21:
```suggestion
type MockWIDSigner struct {
```

Alternatively we could embed a `*WIDMgr` here and have the mock work as both entities, but that could be a bit confusing?

client/widmgr/widmgr.go
```diff
@@ -4,80 +4,367 @@
...
 
-// New workload identity manager.
-func New(c Config) *WIDMgr {
+func NewWIDMgr(signer IdentitySigner, a *structs.Allocation, logger hclog.Logger) *WIDMgr {
+	widspecs := map[string][]*structs.WorkloadIdentity{}
+	tg := a.Job.LookupTaskGroup(a.TaskGroup)
+	for _, task := range tg.Tasks {
+		// Omit default identity as it does not expire
+		widspecs[task.Name] = helper.CopySlice(task.Identities)
+	}
```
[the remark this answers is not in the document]
lgfa29: Incorporated into https://github.com/hashicorp/nomad/pull/18431/commits/48079f4cf0fd156f087a88a08f03574eb1316d2e

client/widmgr/widmgr.go
```diff
@@ -4,80 +4,367 @@
...
+	if m.stopCtx.Err() != nil {
+		c := make(chan *structs.SignedWorkloadIdentity)
+		close(c)
+		return c, func() {}
+	}
+
+	//TODO I think we only want 1 watcher per id, so let's just close on double watches?
+	if existing, ok := m.watchers[id]; ok {
+		close(existing)
+	}
```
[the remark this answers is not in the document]
lgfa29: To add a new data point here, it's _very_ easy to accidentally watch an identity twice. 

I'm updating the `vault` hook and it needs to access the signed identity to derive a Vault token, but the `identity` hook also needs this information to set env/file.

My prototype code looks like this:
https://github.com/hashicorp/nomad/blob/829bf43cb482b91cedf40c77675a280bda17b000/client/allocrunner/taskrunner/vault_hook.go#L364-L381

This was causing the `identity` hook watcher to silently close and the renewed identity not be updated to disk (which is what I was checking). It took me a _long_ time to realize the second `Watch()` was closing it.

Arguably I should be using `Get()` in the `vault`  hook since I don't actually need to keep track of updates, but it was very easy to make this mistake and very hard to debug it.

client/allocrunner/identity_hook_test.go
pkazmierczak: Right, here it's hard because the state is only available to `WIDMgr`, the lock isn't public and I can't access it from the unit test. I'll just `time.Sleep(time.Second)` like we do in the taskrunner tests.

client/widmgr/widmgr.go
pkazmierczak: Ok I think I get it. I'll remove the comment.

client/widmgr/widmgr.go
pkazmierczak: good call, I modified in [e8ef776](https://github.com/hashicorp/nomad/pull/18431/commits/e8ef7763782f937b3915c163134604544919f735)

client/widmgr/signer.go
pkazmierczak: Thanks for clarifying @schmichael, I summarized in a comment in [e8ef776](https://github.com/hashicorp/nomad/pull/18431/commits/e8ef7763782f937b3915c163134604544919f735), I hope it's clear enough.

client/allocrunner/identity_hook.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"context"
+
+	log "github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/nomad/client/allocrunner/interfaces"
+	"github.com/hashicorp/nomad/client/widmgr"
+)
+
+type identityHook struct {
+	ar     *allocRunner
```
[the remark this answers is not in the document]
pkazmierczak: good catch! fixed in [e3dd3d3](https://github.com/hashicorp/nomad/pull/18431/commits/e3dd3d3f15eda095a73e2d73cc7fa349358808c5)

client/allocrunner/identity_hook_test.go
```diff
@@ -0,0 +1,84 @@
...
+	must.Nil(t, err)
+	must.Eq(t, sid.IdentityName, task.Identity.Name)
+	must.NotEq(t, sid.JWT, "")
+
+	// pad expiry time with a second to be safe
+	must.True(t, inTimeSpan(start.Add(ttl).Add(-1*time.Second), start.Add(ttl).Add(1*time.Second), sid.Expiration))
```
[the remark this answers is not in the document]
pkazmierczak: YES. I didn't notice `must.Between`. Fixed in [e3dd3d3](https://github.com/hashicorp/nomad/pull/18431/commits/e3dd3d3f15eda095a73e2d73cc7fa349358808c5)

client/widmgr/mock.go
pkazmierczak: Renamed in [e3dd3d3](https://github.com/hashicorp/nomad/pull/18431/commits/e3dd3d3f15eda095a73e2d73cc7fa349358808c5)

client/allocrunner/taskrunner/identity_hook.go
```diff
@@ -88,39 +72,42 @@ func (h *identityHook) Prestart(context.Context, *interfaces.TaskPrestartRequest
...
-		if signedWID == nil {
-			// The only way to hit this should be a bug as it indicates the server
-			// did not sign an identity for a task on this alloc.
-			return fmt.Errorf("missing workload identity %q", widspec.Name)
-		}
-
-		if err := h.setAltToken(widspec, signedWID.JWT); err != nil {
-			return err
-		}
+		go h.watchIdentity(widspec)
```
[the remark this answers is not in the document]
pkazmierczak: Not sure I follow? Capture as in make sure it's not nil?

client/widmgr/widmgr.go
```diff
@@ -4,80 +4,367 @@
...
 	"github.com/hashicorp/nomad/nomad/structs"
 )
 
-type RPCer interface {
-	RPC(method string, args any, reply any) error
+// IdentityManager defines a manager responsible for signing and renewing
+// signed identities. At runtime it is implemented by *widmgr.WIDMgr.
+type IdentityManager interface {
+	Run() error
+	Get(cstructs.TaskIdentity) (*structs.SignedWorkloadIdentity, error)
```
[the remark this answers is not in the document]
pkazmierczak: `Watch` is sufficient, but `Get` is convenient to have, too? If you don't want to deal with the channel `Watch` returns?

client/widmgr/widmgr.go
```diff
@@ -4,80 +4,367 @@
...
+
 	return &WIDMgr{
-		nodeSecret: c.NodeSecret,
-		region:     c.Region,
-		rpc:        c.RPC,
+		allocID:  a.ID,
+		minIndex: a.CreateIndex,
+		widSpecs: widspecs,
+		signer:   signer,
+		minWait:  10 * time.Second,
```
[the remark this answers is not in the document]
pkazmierczak: Nice, thanks.

client/widmgr/widmgr.go
```diff
@@ -4,80 +4,367 @@
...
+	}
+
+	//TODO I think we only want 1 watcher per id, so let's just close on double watches?
+	if existing, ok := m.watchers[id]; ok {
+		close(existing)
+	}
+
+	// Buffer of 1 so sends don't block on receives
+	c := make(chan *structs.SignedWorkloadIdentity, 1)
+	m.watchers[id] = c
```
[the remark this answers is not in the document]
pkazmierczak: Yeah... I suppose it's up to the watchers to make sure they `cancel()`? Do you have a better solution in mind?

client/widmgr/widmgr.go
```diff
@@ -4,80 +4,367 @@
...
+			reqs = append(reqs, &structs.WorkloadIdentityRequest{
+				AllocID:      m.allocID,
+				TaskName:     taskName,
+				IdentityName: widspec.Name,
+			})
+		}
+	}
+
+	renewNow := false
+	minExp := time.Now().Add(30 * time.Hour) // set high default expiration
```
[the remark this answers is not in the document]
pkazmierczak: hmm, but what would the initial `minExp` value be then? We only know the `token.Expiration` once we look at the tokens themselves, not sure how would we get the `minExp.IsZero()` to work in this case.

client/widmgr/widmgr.go
```diff
@@ -4,80 +4,367 @@
...
+		select {
+		case <-timer.C:
+			m.logger.Trace("getting new signed identities", "num", len(reqs))
+		case <-m.stopCtx.Done():
+			//TODO Close watchers
+			return
+		}
+
+		// Renew all tokens together since its cheap
+		tokens, err := m.signer.SignIdentities(m.minIndex, reqs)
```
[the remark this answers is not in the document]
pkazmierczak: Yeah... I'll make a comment but it feels like something to address in the future PR(s).

client/widmgr/widmgr.go
pkazmierczak: good idea, addressed in ea9635d23d20b1e6d4d9b6967ce4a3f653ac5256

client/allocrunner/taskrunner/identity_hook.go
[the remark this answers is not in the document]
lgfa29: Ops, sorry I should've been more clear.

Go's loop variables should only be referenced within its iteration, as it changes at every loop (this will be [fixed](https://go.dev/blog/loopvar-preview) in Go 1.22 🎉 ). I believe this will result in every goroutine watching the same `widspec`.

client/widmgr/widmgr.go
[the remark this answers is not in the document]
lgfa29: > hmm, but what would the initial minExp value be then? 

Zero. `minExp.IsZero()` would mean "this is the first iteration, so set the value without comparing it". It also handles the case where TTL is 0 as `minExp` will still be zero and set to the first non-zero TTL.

Looking at the code again, it could also serve as a signal that all identities have `ttl = 0` which means we could even exit the `renew` function? 🤔 

I also noticed later that this logic is being used below as well:
https://github.com/hashicorp/nomad/pull/18431/files/265f6cdc29f7424039fa749b80ca320662151c92#r1330575709

client/widmgr/widmgr.go
lgfa29: Oooh...sorry, I misread the code. Looking at it again I understand the comment now. 

Since the channel is buffered, and we drain the channel right before `c <- token` happens, this will never block: either the watcher will consume the channel or `token` will be placed in the (now empty) channel buffer.

This also covers the buffer size `1` in https://github.com/hashicorp/nomad/pull/18431#discussion_r1330553116.

So I think the original code was fine 😅

client/widmgr/widmgr.go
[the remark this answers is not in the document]
tgross: Aren't we using `Get` in the `vault_hook` PR? Or is the plan to update that?

client/widmgr/widmgr.go
```diff
@@ -4,80 +4,377 @@
...
+		allocID:   a.ID,
+		minIndex:  a.CreateIndex,
+		widSpecs:  widspecs,
+		signer:    signer,
+		minWait:   10 * time.Second,
+		lastToken: map[cstructs.TaskIdentity]*structs.SignedWorkloadIdentity{},
+		watchers:  map[cstructs.TaskIdentity]chan *structs.SignedWorkloadIdentity{},
+		stopCtx:   stopCtx,
+		stop:      stop,
+		logger:    logger.Named("identity"),
```
tgross, line 76: This logger's name is the same as the `identity_hook`. Should it have its own name like:

```suggestion
		logger:    logger.Named("widmgr"),
```

client/allocrunner/taskrunner/identity_hook.go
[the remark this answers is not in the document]
pkazmierczak: oof of course 🤦‍♂️ fixed in ddce193eabc77b42a97aaee8a48e04382de4f3d8

client/widmgr/widmgr.go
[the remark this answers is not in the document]
lgfa29: I think this decision is still pending. I can use whichever in the `vault_hook`, but if we go the `Watch`-only route it needs to support multiple watchers (https://github.com/hashicorp/nomad/pull/18431#discussion_r1330781301)

client/widmgr/widmgr.go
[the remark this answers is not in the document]
tgross: > This buffer can only protect the channel from being blocked once right? A watcher that stops listening but doesn't call the `stop` function may leak the goroutine.

I don't think there's any goroutine to leak here? 

In any case, I think it's reasonable to rely on the watcher canceling themselves, especially given that we ensure everything's been closed in the `Shutdown` method. The `stop` function for each watcher deletes the channel from the list (the channel will get GC'd), which prevents the sender (the `renew` goroutine) from blocking on the send because the channel isn't there anymore. We've left it up to `Shutdown` to close and open channels to notify the watcher they can close. So long as the watcher doesn't poll on the channel after it's called `stop`, we should be safe.

client/widmgr/widmgr.go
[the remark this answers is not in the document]
tgross: I like the idea of being able to bail out of the renew loop entirely. That implies downstream consumers will need to _either_ `Get` or `Watch` based on whether they have a TTL, but I think that's ok.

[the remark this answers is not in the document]
pkazmierczak: I addressed the `minExp` value in eb1aba34f92cb4fc5cc9bc8c1bc60788b81397d8, and for now we decided to keep the `Get` method.

client/widmgr/widmgr.go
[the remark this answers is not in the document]
tgross: In a sidebar discussion @pkazmierczak and I settled on keeping `Get` here for now and then once we finish the consumers we can decide whether to keep it.

tgross (approved):
@pkazmierczak I've left some minor remaining comments but the only one that seems potentially blocking are the ones around shutdown and double-closing the channels. Good-to-merge once that's fixed.

client/widmgr/widmgr.go
```diff
@@ -4,80 +4,384 @@
...
-			Region:        m.region,
-			MinQueryIndex: minIndex - 1,
-			AllowStale:    true,
-			AuthToken:     m.nodeSecret,
-		},
+// The caller must call the returned func to stop watching and ensure the
+// watched id actually exists, otherwise the channel never returns a result.
+func (m *WIDMgr) Watch(id cstructs.TaskIdentity) (<-chan *structs.SignedWorkloadIdentity, func()) {
+	m.watchersLock.Lock()
+	defer m.watchersLock.Unlock()
```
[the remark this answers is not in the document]
pkazmierczak: good idea

pkazmierczak:
I applied all your suggestions in dad0e651afe66d6d79b9426fb3acdac02b875157 and I'm gonna merge the PR.

PR #18434, opened 2023-09-08
CSI: Rename volume Mounter to Manager

gulducat:
to align with its broader purpose, and the existing `volumeManager` implementation.

Arguably if this were being designed from scratch, maybe a tight little Mounter interface could be valuable, but this just brings the word usage into alignment, and I won't feel bad about taking advantage of all this nice plumbing to add `ExpandVolume` to this "Manager" soon.

I figure backporting this isolated change might be good for other future bugfixes or similar in this territory, but let me know if I shouldn't backport it for some reason.


tgross (approved):
LGTM :+1:

Issue #18442, opened 2023-09-11
bubble up the go-getter failure reason


shantanugadgil:
### Proposal
Currently the failure reason for go-getter is shown in agent logs, but not in the `alloc status`. bubbling it up to the `alloc status` would be useful for debugging.

### Use-cases
If I have a tar archive with too many files (> 4096), and the extraction fails, the reason for failure is not shown in `alloc status`, although the error message is seen in agent logs.

### Attempted Solutions
N/A. The solution would be do patch some function definitions; to propagate the error upto alloc status.

EDIT: changed the `logs` to `status`, which is what I actually was trying to do! :man_facepalming:

tgross:
From a design standpoint, I'm not sure we ever want anything happening in the agent to go to the allocation logs rather than the agent's logs, because that will interfere with application logging formats. The Task Events are where we typically want to send this sort of thing, and I think that's what you've done in https://github.com/hashicorp/nomad/pull/18444 anyways.

shantanugadgil:
> From a design standpoint, I'm not sure we ever want anything happening in the agent to go to the allocation logs rather than the agent's logs, because that will interfere with application logging formats. The Task Events are where we typically want to send this sort of thing, and I think that's what you've done in #18444 anyways.

I got `alloc status` and `alloc logs` mixed up.

Showing the actual reason for the error in the *status* is what I meant to type. (I'll update the description above)

tgross:
Closed by #18444, which shipped today in Nomad 1.6.2

PR #18444, opened 2023-09-11
bubble up the error message from go-getter

shantanugadgil:
possible fix for https://github.com/hashicorp/nomad/issues/18442

EDIT:

Tested the locally built Nomad agent binary using `nomad agent -dev` with:

```bash
mkdir large_tar
cd large_tar/
mkdir data
for i in {1..5000} ; do echo "" > data/$i.txt; done
tar czvf data.tar.gz data/
python3 -m http.server 9000
```

```hcl
job "tar_fail" {

  group "cache" {
    network {
      port "db" {
        to = 6379
      }
    }

    task "redis" {
      artifact {
        source      = "http://127.0.0.1:9000/data.tar.gz"
        destination = "local/some-directory"
      }

      driver = "docker"

      config {
        image          = "redis:7"
        ports          = ["db"]
        auth_soft_fail = true
      }

      identity {
        env  = true
        file = true
      }

      resources {
        cpu    = 500
        memory = 256
      }
    }
  }
}
```

### DIFF
--- a/helper/subproc/subproc.go
+++ b/helper/subproc/subproc.go
@@ -45,12 +45,14 @@ func Print(format string, args ...any) {
 //
 // r should be a buffer containing output (typically combined stdin + stdout)
 // f should be an HCLogger Print method (e.g. log.Debug)
-func Log(r io.Reader, f func(msg string, args ...any)) {
+func Log(r io.Reader, f func(msg string, args ...any)) string {
 	scanner := bufio.NewScanner(r)
+	line := "unknown error"
 	for scanner.Scan() {
-		line := scanner.Text()
+		line = scanner.Text()
 		f("sub-process", "OUTPUT", line)
 	}
+	return line
 }
 
 // Context creates a context setup with the given timeout.
--- a/helper/subproc/subproc.go
+++ b/helper/subproc/subproc.go
@@ -54,7 +54,7 @@ func Log(r io.Reader, f func(msg string, args ...any)) string {
 		lines += line + "\n"
 		f("sub-process", "OUTPUT", line)
 	}
-	return strings.TrimSuffix(lines, "\n")
+	return strings.TrimSpace(lines)
 }
 
 // Context creates a context setup with the given timeout.
--- a/helper/subproc/subproc.go
+++ b/helper/subproc/subproc.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"os"
 	"time"
+	"strings"
 )
 
 const (
@@ -47,12 +48,13 @@ func Print(format string, args ...any) {
 // f should be an HCLogger Print method (e.g. log.Debug)
 func Log(r io.Reader, f func(msg string, args ...any)) string {
 	scanner := bufio.NewScanner(r)
-	line := "unknown error"
+	lines := ""
 	for scanner.Scan() {
-		line = scanner.Text()
+		line := scanner.Text()
+		lines += line + "\n"
 		f("sub-process", "OUTPUT", line)
 	}
-	return line
+	return strings.TrimSuffix(lines, "\n")
 }
 
 // Context creates a context setup with the given timeout.
--- a/.changelog/18444.txt
+++ b/.changelog/18444.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+status: go-getter failure reason now shown in `alloc status`
+```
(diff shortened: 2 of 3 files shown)


shantanugadgil:
@lgfa29 @jrasell @angrycub @tgross

helper/subproc/subproc.go
```diff
@@ -45,12 +45,14 @@ func Print(format string, args ...any) {
...
-func Log(r io.Reader, f func(msg string, args ...any)) {
+func Log(r io.Reader, f func(msg string, args ...any)) string {
 	scanner := bufio.NewScanner(r)
+	line := "unknown error"
 	for scanner.Scan() {
-		line := scanner.Text()
+		line = scanner.Text()
 		f("sub-process", "OUTPUT", line)
 	}
+	return line
```
tgross, line 55: This effectively returns only the last chunk of the value being scanned.

tgross (changes requested):
@shantanugadgil this doesn't actually get what you've described in https://github.com/hashicorp/nomad/issues/18442 does it? This added the inner error to the `artifact` block but that's only going to show up in Task Events, not the alloc logs. But I think that's actually what we want.

shantanugadgil:
> @shantanugadgil this doesn't actually get what you've described in #18442 does it? This added the inner error to the `artifact` block but that's only going to show up in Task Events, not the alloc logs. But I think that's actually what we want.

I meant to write `alloc status` in the original bug report. Bubbling the error into `alloc status` is good enough.

helper/subproc/subproc.go
shantanugadgil: > This effectively returns only the last chunk of the value being scanned.

OK, I guess this will have to change to a string append (collect for all .Scan() ) and then return the unified string.

I'll check how to do that (limited gopher skills)

shantanugadgil: does this seem OK? 
https://github.com/hashicorp/nomad/pull/18444/commits/e1435e3639609387527d8c8be94f559faa42f1a9

the last TrimSuffix is purely cosmetic. we could NOT do that and thus avoid having to include `"strings"`?

helper/subproc/subproc.go
```diff
@@ -45,12 +46,15 @@ func Print(format string, args ...any) {
...
-func Log(r io.Reader, f func(msg string, args ...any)) {
+func Log(r io.Reader, f func(msg string, args ...any)) string {
 	scanner := bufio.NewScanner(r)
+	lines := ""
 	for scanner.Scan() {
 		line := scanner.Text()
+		lines += line + "\n"
 		f("sub-process", "OUTPUT", line)
 	}
+	return strings.TrimSuffix(lines, "\n")
```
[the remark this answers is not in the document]
shantanugadgil: So I should skip appending the `\n` ?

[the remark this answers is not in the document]
tgross: Oh shoot, you know what, I'm totally wrong because I was looking at the docs for the methods and not the type:

> the default split function breaks the input into lines with line termination stripped.

So what you're doing here is exactly right. Probably worth swapping out `return strings.TrimSuffix(lines, "\n")` with `return strings.TrimSpace(lines)` and then you're good-to-go.

helper/subproc/subproc.go
shantanugadgil: changes incorporated in the latest commit

tgross (approved):
LGTM

I've smoke tested this with a broken `artifact` download and get task events like the following:

```
Recent Events:
Time                       Type                      Description
2023-09-13T09:35:00-04:00  Not Restarting            Policy allows no restarts
2023-09-13T09:35:00-04:00  Failed Artifact Download  failed to download artifact "s3::http://192.168.1.221:9000/test/release": getter subprocess failed: exit status 1: failed to download artifact: error downloading 'http://192.168.1.221:9000/test/release?aws_access_key_id=key&aws_access_key_secret=password1': RequestError: send request failed
caused by: Get "http://192.168.1.221:9000/test?prefix=release": dial tcp 192.168.1.221:9000: connect: connection refused
2023-09-13T09:35:00-04:00  Downloading Artifacts     Client is downloading artifacts
2023-09-13T09:34:59-04:00  Task Setup                Building Task Directory
2023-09-13T09:34:59-04:00  Received                  Task received by client
```

tgross:
I've got a backport PR baking in https://github.com/hashicorp/nomad/pull/18479 so once that's green I'll merge it and that'll get shipped in the upcoming Nomad 1.6.2

PR #18452, opened 2023-09-11
[f-gh-17449] Add flag to modify the number of time the client attempts to call the locks endpoints

Juanadelacuesta:
Following the Consul lead, this PR introduces a flag to use on the CLI when using the var lock command that allows the user to modify the number of times the client retries a call in case of an unexpected response.

### DIFF
--- a/command/var_lock.go
+++ b/command/var_lock.go
@@ -17,6 +17,8 @@ import (
 	"github.com/posener/complete"
 )
 
+const defaultMaxClientRetries = 5
+
 type VarLockCommand struct {
 	shell     bool
 	inFmt     string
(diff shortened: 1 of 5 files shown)


command/var_lock.go
```diff
@@ -64,6 +66,13 @@ Var lock Options:
   -delay
 	Optional, time the variable is blocked from locking when a lease is not renewed.	
 	Defaults to 15s.
+  -max-retry
```
tgross, line 69:
```suggestion

  -max-retry
```

tgross (approved):
LGTM. We should include documentation updates in `website/content/docs/commands/var/put.mdx` as well. And I think we need a changelog entry for this?

Juanadelacuesta:
We do need to update the documentation for the command and the SDK, coming in the next PR

PR #18456, opened 2023-09-12
Adding timeouts to group and task

### LINKED ISSUE #1782: [feature] Timeout for batch jobs
I'm currently running a handful of periodic batch jobs. It's great that nomad doesn't schedule another one if a current one is still running. However, I think it would be helpful to stop a batch job on timeout if it's running beyond a set time. Maybe there could be a script run on timeout or the script itself would just have to handle the signal.

mikenomitch:
During a hack week I took a stab at an initial implementation of timeouts for tasks. Pushing up this draft PR as a starting point for discussion. (Also note to anybody external to HashiCorp reading this, I'm a PM, not a full engineer, so don't get too excited just yet 😄)

What this does:
- This creates a new taskrunner hook and new allocrunner hook for timeouts.
- In the jobspec, if a new "timeout" block is set, the runners start a goroutine that waits for the deadline to pass. The allocation or task (depending on the block) is then killed at the deadline.

Open Product Questions:
- I am currently failing the allocations if the group timeout happens, and failing them at the task level if FailOnTimeout is set. It kind of feels like there should be a top level "timed out" status instead? If a task fails due to timeout, should a postStop task get run? If a preStart tails times out, should a main task get run? Do people need customization of this behavior?
- Right now this depends on a Nomad client running. Is that an issue for people?
- I think timezones need to be handled. I *think* you could just pass in a TZ at the jobspec level and us that? If not, default to the client's TZ? To UTC? Could we just not do this initially and always go UTC?
- This includes the functionality on service and system jobs... would that ever make sense?

Open Technical Questions:
- This is saving information about the timeout in Nomad Variables. Its sort of cool that this is inspectable by the user, but I think this should probably be "its own thing" in the Nomad state store. Any disagreement there? What's the "right way" to store this?
- Is there a better way to wait in the goroutine? I feel like this probably checks in too often and is wasting cycles?

I'm sure there are more open questions/issues (and unfortunately I've forgotten a lot since I did this over a month ago), but wanted to get the conversation going!

Fixes https://github.com/hashicorp/nomad/issues/1782

### DIFF
--- a/api/tasks.go
+++ b/api/tasks.go
@@ -459,6 +459,7 @@ type TaskGroup struct {
 	MaxClientDisconnect       *time.Duration            `mapstructure:"max_client_disconnect" hcl:"max_client_disconnect,optional"`
 	Scaling                   *ScalingPolicy            `hcl:"scaling,block"`
 	Consul                    *Consul                   `hcl:"consul,block"`
+	Timeout                   *TaskGroupTimeout         `hcl:"timeout,block"`
 }
 
 // NewTaskGroup creates a new TaskGroup.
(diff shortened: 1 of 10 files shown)


schmichael:
Superseded by #27803

Issue #18438, opened 2023-09-08
Add TTL field support to WorkloadIdentityConfig struct


schmichael:
> Just noting here that with https://github.com/hashicorp/nomad/pull/18363 merged the new `TTL` field should also be added to [`config.WorkloadIdentityConfig`](https://github.com/hashicorp/nomad/blob/main/nomad/structs/config/workload_id.go#L18).

_Originally posted by @lgfa29 in https://github.com/hashicorp/nomad/issues/18262#issuecomment-1702864958_

PR #18457, opened 2023-09-12
config: add TTL to agent identity config

lgfa29:
Add support for identity token TTL in agent configuration fields such as Consul `service_identity` and `template_identity`.

Closes https://github.com/hashicorp/nomad/issues/18438


command/agent/config_parse.go
```diff
@@ -72,6 +72,29 @@ func ParseConfigFile(path string) (*Config, error) {
...
+		if err := parseVaults(c, matches); err != nil {
+			return nil, fmt.Errorf("error parsing 'vault': %w", err)
+		}
+	}
+	matches = list.Filter("consul")
+	if len(matches.Items) > 0 {
+		if err := parseConsuls(c, matches); err != nil {
+			return nil, fmt.Errorf("error parsing 'consul': %w", err)
+		}
+	}
```
lgfa29, line 96: We need to move this earlier so that `TTLHCL` in `consul` and `vault` blocks are loaded properly before they are parsed as durations below.

pkazmierczak (approved):
👍

tgross (approved):
LGTM! Looks like the Vault default identity block is still TODO, but it'll use these same code paths?

PR #18463, opened 2023-09-12
feature: Add prune flag for nomad server force-leave command

nvanthao:
This feature will help operator to remove a failed/left node from Serf layer immediately without waiting for 24 hours for the node to be reaped

* Update CLI with prune flag
* Update API /v1/agent/force-leave with prune query string parameter
* Update CLI and API doc
* Add unit test

### DIFF
--- a/api/agent.go
+++ b/api/agent.go
@@ -35,6 +35,12 @@ type KeyringRequest struct {
 	Key string
 }
 
+// ForceLeaveOpts are used to configure the ForceLeave method.
+type ForceLeaveOpts struct {
+	// Prune indicates whether to remove a node from the list of members
+	Prune bool
+}
+
 // Agent returns a new agent which can be used to query
 // the agent-specific endpoints.
 func (c *Client) Agent() *Agent {
@@ -162,10 +168,19 @@ func (a *Agent) MembersOpts(opts *QueryOptions) (*ServerMembers, error) {
 }
 
 // ForceLeave is used to eject an existing node from the cluster.
-func (a *Agent) ForceLeave(node string, prune bool) error {
+func (a *Agent) ForceLeave(node string) error {
+	_, err := a.client.put("/v1/agent/force-leave?node="+node, nil, nil, nil)
+	return err
+}
+
+// ForceLeaveWithOptions is used to eject an existing node from the cluster
+// with additional options such as prune.
+func (a *Agent) ForceLeaveWithOptions(node string, opts ForceLeaveOpts) error {
 	v := url.Values{}
 	v.Add("node", node)
-	v.Add("prune", strconv.FormatBool(prune))
+	if opts.Prune {
+		v.Add("prune", "1")
+	}
 	_, err := a.client.put("/v1/agent/force-leave?"+v.Encode(), nil, nil, nil)
 	return err
 }
--- a/api/agent.go
+++ b/api/agent.go
@@ -169,7 +169,9 @@ func (a *Agent) MembersOpts(opts *QueryOptions) (*ServerMembers, error) {
 
 // ForceLeave is used to eject an existing node from the cluster.
 func (a *Agent) ForceLeave(node string) error {
-	_, err := a.client.put("/v1/agent/force-leave?node="+node, nil, nil, nil)
+	v := url.Values{}
+	v.Add("node", node)
+	_, err := a.client.put("/v1/agent/force-leave?"+v.Encode(), nil, nil, nil)
 	return err
 }
 
--- a/api/agent.go
+++ b/api/agent.go
@@ -162,8 +162,11 @@ func (a *Agent) MembersOpts(opts *QueryOptions) (*ServerMembers, error) {
 }
 
 // ForceLeave is used to eject an existing node from the cluster.
-func (a *Agent) ForceLeave(node string) error {
-	_, err := a.client.put("/v1/agent/force-leave?node="+node, nil, nil, nil)
+func (a *Agent) ForceLeave(node string, prune bool) error {
+	v := url.Values{}
+	v.Add("node", node)
+	v.Add("prune", strconv.FormatBool(prune))
+	_, err := a.client.put("/v1/agent/force-leave?"+v.Encode(), nil, nil, nil)
 	return err
 }
 
--- a/api/agent_test.go
+++ b/api/agent_test.go
@@ -189,8 +189,6 @@ func TestAgent_ForceLeavePrune(t *testing.T) {
 		if err != nil {
 			return err
 		}
-		t.Log("membersBefore", membersBefore.Members)
-		t.Log("membersAfter", membersAfter.Members)
 		if len(membersAfter.Members) == len(membersBefore.Members) {
 			return fmt.Errorf("node did not get pruned")
 		}
@@ -199,7 +197,7 @@ func TestAgent_ForceLeavePrune(t *testing.T) {
 	must.Wait(t, wait.InitialSuccess(
 		wait.ErrorFunc(f),
 		wait.Timeout(5*time.Second),
-		wait.Gap(1*time.Second),
+		wait.Gap(100*time.Millisecond),
 	))
 
 }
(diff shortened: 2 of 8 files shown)


api/agent.go
```diff
@@ -162,8 +162,11 @@ func (a *Agent) MembersOpts(opts *QueryOptions) (*ServerMembers, error) {
 }
 
 // ForceLeave is used to eject an existing node from the cluster.
-func (a *Agent) ForceLeave(node string) error {
-	_, err := a.client.put("/v1/agent/force-leave?node="+node, nil, nil, nil)
+func (a *Agent) ForceLeave(node string, prune bool) error {
```
tgross, line 165: Everything in this package is part of our public Go API, so we can't change the signature without breaking backwards compatibility. Unfortunately we'll need to add a new endpoint here. (I recommend something like `ForceLeaveWithOptions` and have an options struct as the second parameter, so that we can expand on that in the future if need be.)

command/server_force_leave.go
```diff
@@ -28,7 +28,13 @@ Usage: nomad server force-leave [options] <node>
 
 General Options:
 
-  ` + generalOptionsUsage(usageOptsDefault|usageOptsNoNamespace)
+  ` + generalOptionsUsage(usageOptsDefault|usageOptsNoNamespace) + `
+  
+Server Force-Leave Options:
+
+  -prune
+    Removes failed or left server from the list of members entirely.
```
tgross, line 36: I'm not sure words like "entirely" or "completely" (as in the help text below) sufficiently describe the difference between what happens here. Let's be much more specific and alter the main text to say "the failed or left server is marked as removed, and will be garbage collected after 24h" and then here "the failed or left server is removed from the Serf membership immediately." (This same suggestion applies everywhere we have docs changes in this PR.)

Also, in the above docs, it says:
> Note that if the member is actually still alive, it will eventually rejoin the cluster again.

I think that still applies here, right? So we should make doubly-sure it's clear that we're removing it from the current list of Serf members but that a live server will still rejoin.

command/agent/agent_endpoint.go
```diff
@@ -332,8 +332,15 @@ func (s *HTTPServer) AgentForceLeaveRequest(resp http.ResponseWriter, req *http.
...
+
 	// Attempt remove
-	err := srv.RemoveFailedNode(node)
+	var err error
+	if prune {
+		err = srv.RemoveFailedNodePrune(node)
+	} else {
+		err = srv.RemoveFailedNode(node)
+	}
 	return nil, err
```
tgross, line 344: This sets `prune` to `true` even if the `prune` query parameter is `?prune=false`, because it's only checking for the _existence_ of the key, for which the value is `"false"` (a string, not a bool). 

You should use the `parseBool` helper in `command/http.go` rather than the stdlib function, as it returns (`nil, nil`) if the parameter isn't present and you can have that default to the existing behavior as you expect.

api/agent_test.go
```diff
@@ -152,6 +153,50 @@ func TestAgent_ForceLeave(t *testing.T) {
...
+	defer s2.Stop()
+
+	n, err := a.Join(s2.SerfAddr)
+	must.NoError(t, err)
+	must.One(t, n)
+
+	err = a.ForceLeave(nodeName, true)
+	must.NoError(t, err)
+
+	time.Sleep(3 * time.Second)
```
tgross, line 180: This is going to be flaky. We're stopping the server and then waiting some arbitrary amount of time before seeing if the node has been removed from the member list. You can get this scenario:

* `ForceLeave`
* start waiting 3 seconds
* node leaves
* node rejoins
* 3 seconds expires
* run the `must.Wait` for 3 seconds

Even if you removed the `time.Sleep` here, the node could be removed and rejoined in the 100ms window of the `must.Wait`. So we should also be stopping the node we're removing, so that we can be sure it's not going to rejoin. And then we can safely remove this `time.Sleep`, which doesn't get us anything when we have `must.Wait` below.

tgross (changes requested):
Hi @nvanthao! Thanks for the PR, this is a good start. I've left some comments that'll need to be addressed before this can be merged.

api/agent.go
nvanthao: Many thanks for the backward compatibility advice, Tim! I have updated the code with new `ForceLeaveWithOptions` method.

command/server_force_leave.go
nvanthao: Thank you! I have updated the wording to reiterate that `-prune` won't work with alive server as well as added the part of node being garbaged collected after 24h.

command/agent/agent_endpoint.go
nvanthao: Thanks so much! This has been updated.

api/agent_test.go
nvanthao: Could you help me with the test? I have updated the test following your instructions, but it is still fail. I am not sure if it has to do with how the Test Server is created.

api/agent.go
```diff
@@ -162,10 +168,19 @@ func (a *Agent) MembersOpts(opts *QueryOptions) (*ServerMembers, error) {
 }
 
 // ForceLeave is used to eject an existing node from the cluster.
-func (a *Agent) ForceLeave(node string, prune bool) error {
+func (a *Agent) ForceLeave(node string) error {
+	_, err := a.client.put("/v1/agent/force-leave?node="+node, nil, nil, nil)
```
tgross, line 172: We've dropped the URL encoding we were doing previously on the values. Let's make sure we're still doing that here.

api/agent_test.go
```diff
@@ -152,6 +153,55 @@ func TestAgent_ForceLeave(t *testing.T) {
...
+		t.Log("membersAfter", membersAfter.Members)
+		if len(membersAfter.Members) == len(membersBefore.Members) {
+			return fmt.Errorf("node did not get pruned")
+		}
+		return nil
+	}
+	must.Wait(t, wait.InitialSuccess(
+		wait.ErrorFunc(f),
+		wait.Timeout(5*time.Second),
+		wait.Gap(1*time.Second),
```
tgross, line 202: In the happy case, this will speed up tests quite a bit to check frequently:

```suggestion
		wait.Gap(100*time.Millisecond),
```

tgross (changes requested):

tgross:
@nvanthao you'll want to rebase this on `main` and that'll get the remaining failing test working

website/content/docs/commands/server/force-leave.mdx
```diff
@@ -22,13 +24,21 @@ nomad server force-leave [options] <node>
...
 capability.
 
 ## General Options
 
 @include 'general_options_no_namespace.mdx'
 
+## Server Force-Leave Options
+
+- `-prune`: Specifies whether to forcibly remove the node from the list of members. 
+Pruning a node in the `left` or `failed` state removes it from the list altogether.
```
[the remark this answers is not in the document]
nvanthao: done!

nvanthao:
Many thanks @tgross! I have updated the PR with your latest feedback!

tgross (changes requested):
Getting really close @nvanthao, I think this is the last round of issues.

command/agent/agent_endpoint.go
```diff
@@ -332,8 +332,15 @@ func (s *HTTPServer) AgentForceLeaveRequest(resp http.ResponseWriter, req *http.
 		return nil, CodedError(400, "missing node to force leave")
 	}
 
+	prune, _ := parseBool(req, "prune")
```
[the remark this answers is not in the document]
nvanthao: thank you!

website/content/api-docs/agent.mdx
```diff
@@ -441,13 +441,16 @@ The table below shows this endpoint's support for
 ### Parameters
 
 - `node` `(string: <required>)` - Specifies the name of the node to force leave.
+- `prune` `(boolean: <optional>)` - Specifies whether to forcibly remove the node from
+the list of members. Pruning a node in the `left` or `failed` state removes it from the list altogether.
```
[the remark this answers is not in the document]
nvanthao: thank you!

nvanthao:
many thanks @tgross for being patient with me! next PRs will be better!

tgross (approved):
LGTM! Thanks for sticking with it @nvanthao!

PR #18465, opened 2023-09-12
[f-gh-17449] Remove usage of deleted function 

Juanadelacuesta:
The [PR](https://github.com/hashicorp/nomad/pull/17660) removed the function VarGet that was used for the lock renewal. This PR updates the code to use a different function that was not removed.


gulducat (approved):
Get em!

PR #18466, opened 2023-09-12
consul: check for warnings on service identity

lgfa29:
Apply workload identity warnings to group and task level Consul services that have an identity assigned.

### DIFF
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -7331,10 +7331,18 @@ func (tg *TaskGroup) Warnings(j *Job) error {
 		mErr.Errors = append(mErr.Errors, fmt.Errorf("mbits has been deprecated as of Nomad 0.12.0. Please remove mbits from the network block"))
 	}
 
+	// Validate group-level services.
+	for _, s := range tg.Services {
+		if err := s.Warnings(); err != nil {
+			err = multierror.Prefix(err, fmt.Sprintf("Service %q:", s.Name))
+			mErr = *multierror.Append(&mErr, err)
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


nomad/structs/structs.go
```diff
@@ -8136,7 +8144,15 @@ func (t *Task) Warnings() error {
...
-			mErr.Errors = append(mErr.Errors, err)
+			mErr = *multierror.Append(&mErr, err)
+		}
+	}
+
+	// Validate task-level services.
+	for _, s := range t.Services {
+		if err := s.Warnings(); err != nil {
+			err = multierror.Prefix(err, fmt.Sprintf("Service %q:", s.Name))
+			mErr = *multierror.Append(&mErr, err)
```
lgfa29, line 8155: This way of appending errors flattens inner errors so the output goes from something like this:

```
Group "web" has warnings: 2 errors occurred:
        * 1 error occurred:
        * Service "web": identities without an expiration are insecure


        * 2 errors occurred:
        * Task "web": IOPS has been deprecated as of Nomad 0.9.0. Please remove IOPS from resource block.
        * Task "web": Service "web": identities without an expiration are insecure
```

To this:

```
Group "web" has warnings: 2 errors occurred:
        * Service "web": identities without an expiration are insecure
        * Task "web" has warnings: 2 errors occurred:
        * IOPS has been deprecated as of Nomad 0.9.0. Please remove IOPS from resource block.
        * Service "web": identities without an expiration are insecure
```

Still weird, but I think a little better?

tgross (approved):
LGTM :+1:

PR #18468, opened 2023-09-13
Add more links from Variables doc to Nomad Variable template examples

jjti:
It was not obvious to me that Nomad Variables could themselves be used in templates. I read this page and wasn't exactly sure what they're used for w/o an example.

The timeline was:
1. day 0: read this page
2. day 1-2: monkey with HCL2 variables in job specs
3. day 3: come across this page on [Nomad Variables in templates](https://developer.hashicorp.com/nomad/docs/job-specification/template#nomad-variables) when it dawned on me that... that's how they can be used and it's sick

This PR adds two links and one sentence to try to get that across to Nomad novices like myself so they can go straight from Nomad Variables > "oh this is cool" faster.

I think ideally this doc also has an example like from the Nomad Variables section of the `template` page:

```md
template {
  data        = <<EOH
{{ range nomadVarList "path/to/filter" }}
  {{ . }}
{{ end }}
EOH
}
```

### DIFF
--- a/website/content/docs/concepts/variables.mdx
+++ b/website/content/docs/concepts/variables.mdx
@@ -7,12 +7,13 @@ description: Learn about the Nomad Variables feature
 # Nomad Variables
 
 Most Nomad workloads need access to config values or secrets. Nomad has a
-`template` block to provide such configuration to tasks, but prior to Nomad 1.4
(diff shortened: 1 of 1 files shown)


tgross (approved):
LGTM

PR #18469, opened 2023-09-13
[f-gh-17449] Locks: Fixes for lock visibility

Juanadelacuesta:
This PR addresses a 3 of small issues:
-  The logic to not show the lock ID unless a management token was used for the Read and List endpoints only worked if ACL was enabled, by changing `if aclObj != nil && !aclObj.IsManagement() ` to `!(aclObj != nil && aclObj.IsManagement())` the problem is fixed. 

- When a new variable is created and it includes lock parameters, the endpoint returns an error, to avoid creating locks without TTL

- When releasing a lock, the items provided in the variable are ignored because the lock release operation doesn't update the items, in this PR a new type of error is introduces to reject this requests.

### DIFF
--- a/api/locks.go
+++ b/api/locks.go
@@ -131,7 +131,14 @@ func (l *Locks) Acquire(ctx context.Context) (string, error) {
 func (l *Locks) Release(ctx context.Context) error {
 	var out Variable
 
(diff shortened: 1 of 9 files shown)


jrasell (commented):
The code looks good, but there are a number of tests failing; is this because we have corrected behaviour or because we have broken some?

PR #18504, opened 2023-09-14
drivers: plumb hardware topology via grpc into drivers

shoenig:
This PR removes the temporary use of detecting system hardware manually
in each driver and instead using the Client's detected topology by plumbing that
data over gRPC into each driver. This ensures that Client _configuration_ is taken to account
consistently in all references to system topology.

Basically, `numalib.Scan()` only happens once (in the Client) now. (But also all over the place in tests).


plugins/base/proto/base.proto
```diff
@@ -87,6 +87,40 @@ message NomadDriverConfig {
     // communicating with plugin subsystems over loopback
     // buf:lint:ignore FIELD_LOWER_SNAKE_CASE
     uint32 ClientMinPort = 2;
+
+    // Topology is the complex hardware topology detected by the client
+    // combined with client configuration.
+    ClientTopology Topology = 3;
```
tgross, line 93: Doesn't have to be in this PR, but there's documentation for the [`SetConfig`](https://developer.hashicorp.com/nomad/docs/concepts/plugins/base#setconfig-config-config-error) RPC to update with this.

plugins/base/proto/base.proto
```diff
@@ -87,6 +87,40 @@ message NomadDriverConfig {
...
+    // values is the flattened matrix of slit values
+    repeated uint32 values = 2 [packed=true];
+}
+
+// numalib/Core
+message ClientTopologyCore {
+  uint32 socket_id = 1;
+  uint32 node_id = 2;
+  uint32 core_id = 3;
+  bool core_grade = 4; // p=true, e=false
```
tgross, line 119: Not an enum here?

plugins/base/base.go
```diff
@@ -65,37 +78,130 @@ type ClientDriverConfig struct {
...
+	for row := 0; row < dimension; row++ {
+		for col := 0; col < dimension; col++ {
+			values = append(values, uint32(slit[row][col]))
+		}
+	}
+	return &proto.ClientTopologySLIT{
+		Dimension: uint32(dimension),
+		Values:    values,
+	}
+}
```
tgross, line 189: We have a single dimension on the SLIT table which implies it's always a square matrix, and I'm having trouble envisioning how that could be the case (I'm almost certainly missing something). This or the protobuf might not be a bad place to leave a tiny piece of domain knowledge in the form of a docstring about what how this table dimension works

tgross (approved):
LGTM! 

(This looked like a much more gruesome PR to review than it really was :grinning: )

plugins/base/base.go
shoenig: The SLIT table is just a way of expressing "how expensive is it for a process on one numa node to access memory on another numa node". If we have nodes [0, N] we would need a mapping from each node to [0, N], hence a square.

tgross: Ok, gotcha. I think what was confusing me is a couple of the test cases have tables that can't exist, like:

```go
NodeIDs:   idset.From[hw.NodeID]([]hw.NodeID{0, 1}),
Distances: numalib.SLIT{{10}},
```

A table for 2 nodes could only have a dimension of 1 if the node-to-self relationship was excluded, but you can't make a square matrix [0-2] if you do that.

It's fine that test cases are implausible I suppose, maybe even good b/c it means the functions are robust against weird input. But it made for that "what am I missing here?" moment. :grinning:

shoenig: Ah good catch, I've fixed up the tests to be plausible!

PR #18519, opened 2023-09-15
[f-gh-17449] Add early return and configurable backoff to the leaser

Juanadelacuesta:
This PR introduces 2 new options to the API and var lock command to allow users to return early if the lock is not acquired and to modify the sleep period between calls to lock acquire.

### DIFF
--- a/api/locks.go
+++ b/api/locks.go
@@ -209,24 +209,49 @@ type LockLeaser struct {
 	renewalPeriod time.Duration
 	waitPeriod    time.Duration
 	randomDelay   time.Duration
+	earlyReturn   bool
+	locked        bool
 
 	locker Locker
 }
 
+type LockLeaserOption = func(l *LockLeaser)
+
+// LockLeaserOptionWithEarlyReturn informs the leaser to return after the lock
+// acquire fails and to not wait to attempt again.
+func LockLeaserOptionWithEarlyReturn(er bool) LockLeaserOption {
+	return func(l *LockLeaser) {
+		l.earlyReturn = er
+	}
+}
+
+// LockLeaserOptionWithWaitPeriod is used to set a back off period between
+// calls to attempt tp acquire the lock. By default it is set to 1.1 * TTLs.
+func LockLeaserOptionWithWaitPeriod(wp time.Duration) LockLeaserOption {
+	return func(l *LockLeaser) {
+		l.waitPeriod = wp
+	}
+}
+
 // NewLockLeaser returns an instance of LockLeaser. callerID
 // is optional, in case they it is not provided, internal one will be created.
-func (c *Client) NewLockLeaser(l Locker) *LockLeaser {
+func (c *Client) NewLockLeaser(l Locker, opts ...LockLeaserOption) *LockLeaser {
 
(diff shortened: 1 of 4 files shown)


api/locks.go
```diff
@@ -202,24 +209,49 @@ type LockLeaser struct {
...
+// LockLeaserOptionWithEarlyReturn informs the leaser to return after the lock
+// acquire fails and to not wait to attempt again.
+func LockLeaserOptionWithEarlyReturn(er bool) LockLeaserOption {
+	return func(l *LockLeaser) {
+		l.earlyReturn = er
+	}
+}
+
+// LockLeaserOptionWithWaitPeriod is used to set a back off period between
+// calls to attempt tp acquire the lock. By default it is set to 1.1 * TTLs.
```
tgross, line 229:
```suggestion
// calls to attempt to acquire the lock. By default it is set to 1.1 * TTLs.
```

command/var_lock.go
```diff
@@ -73,7 +73,15 @@ Var lock Options:
...
 	disable.
-
+  
+  -early-return 
+	Optional, early-return indicates the command to return if the lock is not 
+	acquired instead of waiting on stand by to try again. Defaults to false.
+
+  -backoff
+	Optional, by default the lease algorithm waits for 1.1 times the lock TTL, 
+	using the backoff flag, this backoff period can be modified.
```
tgross, line 83: This isn't clear what the backoff is waiting _for_; the docstring you have in `api` package is more clear. I'd go with something like:

```suggestion
  -backoff
	Optional, indicates how long to wait between attempts to obtain the lock. 
	By default the lease algorithm waits for 1.1 times the lock TTL.
```

nomad/variables_endpoint.go
```diff
@@ -344,15 +365,17 @@ func (sv *Variables) Read(args *structs.VariablesReadRequest, reply *structs.Var
 			// Setup the output
 			reply.Data = nil
 			if out != nil {
+
+				if !(aclObj != nil && aclObj.IsManagement()) {
+					out.Lock = nil
+				}
```
tgross, line 371: This is modifying the original value we got out of the state store outside of a Raft transaction (and in a read request!). We can only modify copies of objects we get out of the state store, so I don't think we can make this change.

api/locks.go
```diff
@@ -131,7 +131,14 @@ func (l *Locks) Acquire(ctx context.Context) (string, error) {
 func (l *Locks) Release(ctx context.Context) error {
 	var out Variable
 
-	_, err := l.c.retryPut(ctx, "/v1/var/"+l.variable.Path+"?lock-release", l.variable, &out, &l.WriteOptions)
+	rv := &Variable{
+		Lock: &VariableLock{
+			ID: l.variable.Lock.ID,
```
tgross, line 136: Is it ever possible for `l.variable.Lock` to be nil, even by user error? This is in the `api` package, so we should probably assume accidental misuse.

tgross (changes requested):
This looks pretty good @Juanadelacuesta! I've left a few comments, but only the one about not mutating objects from the state store is critical.

tgross (approved):
LGTM!

PR #18520, opened 2023-09-15
[f-gh-17449] Add lock sdk to documentation and update missing links

Juanadelacuesta:


website/content/docs/concepts/variables.mdx
```diff
@@ -194,3 +234,5 @@ See [Workload Associated ACL Policies] for more details.
 [workload identity]: /nomad/docs/concepts/workload-identity
 [Workload Associated ACL Policies]: /nomad/docs/concepts/workload-identity#workload-associated-acl-policies
 [ACL policy namespace rules]: /nomad/docs/other-specifications/acl-policy#namespace-rules
+[The Chubby Lock Service for Loosely-Coupled Distributed Systems]: https://research.google/pubs/pub27897/
+[`nomad var lock`]: /nomad/docs/commands/var
```
gulducat, line 238: couple things:

* this links to `commands/var`, not `commands/var/lock` directly, and `commands/var` has a list of the sub-commands, but `lock` is not among them yet

* if the intent is to show implemented usage of the golang lock-lease-function-protector, then that commands/var doc doesn't really get me there.  maybe instead link to github instead?  right now it could be https://github.com/hashicorp/nomad/blob/f-gh-17449/command/var_lock.go#L240 -- but once we're sure of the release, it might be `https://github.com/hashicorp/nomad/blob/release/1.7.0/command/var_lock.go#L240`

gulducat (commented):
lookin pretty good!

It's difficult to give diagram feedback here, but I wonder if the "run protected function" box might be highlighted some more, to emphasize that it's the primary goal of the code, with everything around it just there to support/protect it.  I also don't remember, and the diagram doesn't quite explain: what happens if/when the protected func does return?  Does the leaser run it again?

gulducat (approved):
maybe change the PR title to match the added docs, but LGTM! :tada:

PR #18522, opened 2023-09-15
CSI: Expand Volume (Node edition)

gulducat:
Follow-up to #18359, wrapping up the full feature requested in #10324 (sans docs)

After the controller plugin has expanded the disk at e.g. a cloud vendor, it may say that we also need to issue [NodeExpandVolume](https://github.com/container-storage-interface/spec/blob/c918b7f/spec.md#nodeexpandvolume) for the node plugin to make the new disk space available to task(s) that have claims on the volume by e.g. expanding the filesystem on the node.

The main gap in test coverage is `sendCSINodeRPC` which I split out from `ClientCSI.NodeDetachVolume`, which was not actually tested before.  I know that it works, but I'm open to (figuring out how to) add proper coverage there, amongst all the other layers with greater coverage, if it's a considerable concern.

### DIFF
--- a/client/pluginmanager/csimanager/testing.go
+++ b/client/pluginmanager/csimanager/testing.go
@@ -1,0 +1,73 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package csimanager
+
+import (
+	"context"
+
+	"github.com/hashicorp/nomad/client/pluginmanager"
+	nstructs "github.com/hashicorp/nomad/nomad/structs"
+	"github.com/hashicorp/nomad/plugins/csi"
+)
+
+var _ Manager = &MockCSIManager{}
+
+type MockCSIManager struct {
+	VM *MockVolumeManager
+
+	NextWaitForPluginErr    error
+	NextManagerForPluginErr error
+}
+
+func (m *MockCSIManager) PluginManager() pluginmanager.PluginManager {
+	panic("implement me")
+}
+
+func (m *MockCSIManager) WaitForPlugin(_ context.Context, pluginType, pluginID string) error {
+	return m.NextWaitForPluginErr
+}
+
+func (m *MockCSIManager) ManagerForPlugin(_ context.Context, pluginID string) (VolumeManager, error) {
+	return m.VM, m.NextManagerForPluginErr
+}
+
+func (m *MockCSIManager) Shutdown() {
+	panic("implement me")
+}
+
+var _ VolumeManager = &MockVolumeManager{}
+
+type MockVolumeManager struct {
+	NextExpandVolumeErr  error
+	LastExpandVolumeCall *MockExpandVolumeCall
+}
+
+func (m *MockVolumeManager) MountVolume(_ context.Context, vol *nstructs.CSIVolume, alloc *nstructs.Allocation, usageOpts *UsageOptions, publishContext map[string]string) (*MountInfo, error) {
+	panic("implement me")
+}
+
+func (m *MockVolumeManager) UnmountVolume(_ context.Context, volID, remoteID, allocID string, usageOpts *UsageOptions) error {
+	panic("implement me")
+}
+
+func (m *MockVolumeManager) HasMount(_ context.Context, mountInfo *MountInfo) (bool, error) {
+	panic("implement me")
+}
+
+func (m *MockVolumeManager) ExpandVolume(_ context.Context, volID, remoteID, allocID string, usageOpts *UsageOptions, capacity *csi.CapacityRange) (int64, error) {
+	m.LastExpandVolumeCall = &MockExpandVolumeCall{
+		volID, remoteID, allocID, usageOpts, capacity,
+	}
+	return capacity.RequiredBytes, m.NextExpandVolumeErr
+}
+
+type MockExpandVolumeCall struct {
+	VolID, RemoteID, AllocID string
+	UsageOpts                *UsageOptions
+	Capacity                 *csi.CapacityRange
+}
+
+func (m *MockVolumeManager) ExternalID() string {
+	return "mock-volume-manager"
+}
--- a/nomad/client_csi_endpoint.go
+++ b/nomad/client_csi_endpoint.go
@@ -224,12 +224,34 @@ func (a *ClientCSI) isRetryable(err error) bool {
 
 func (a *ClientCSI) NodeDetachVolume(args *cstructs.ClientCSINodeDetachVolumeRequest, reply *cstructs.ClientCSINodeDetachVolumeResponse) error {
 	defer metrics.MeasureSince([]string{"nomad", "client_csi_node", "detach_volume"}, time.Now())
+	return a.sendCSINodeRPC(
+		args.NodeID,
+		"CSI.NodeDetachVolume",
+		"ClientCSI.NodeDetachVolume",
+		structs.RateMetricWrite,
+		args,
+		reply,
+	)
+}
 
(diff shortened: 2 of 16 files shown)


nomad/csi_endpoint.go
```diff
@@ -1272,12 +1272,62 @@ func (v *CSIVolume) expandVolume(vol *structs.CSIVolume, plugin *structs.CSIPlug
 	logger.Info("controller done expanding volume")
 
 	if cResp.NodeExpansionRequired {
-		v.logger.Warn("TODO: also do node volume expansion if needed") // TODO
+		if err = v.nodeExpandVolume(vol, plugin, capacity); err != nil {
+			// log instead of return err, because even if there are node failures,
+			// the real volume itself has been updated, and we should update raft
+			// up in Register/Create to reflect that new reality.
+			logger.Error("error from nodeExpandVolume", "error", err)
```
tgross, line 1279: I'm not sure about this in terms of consistency, but there aren't any obviously good answers other than writing intermediate states to Raft:
* If we don't successfully Node Expand, then we'll record that the volume has expanded to the state store but it won't have had any file system resize operations completed. 
* The user can't then just run Register again, because the state store will have the updated size and won't detect that as a change.
* Or, if we _didn't_ write the changed volume size to the state store, the user could Register again but that would force us to send the Controller Expand again. Can we do that that safely?

Looking at the possible errors that can be returned by Node Expand, there are some errors that the operators needs to see to change their request. Those errors wouldn't be surfaced to the operator (they may not have access to the Nomad server logs).

nomad/client_csi_endpoint.go
```diff
@@ -224,12 +224,34 @@ func (a *ClientCSI) isRetryable(err error) bool {
...
+		args.Claim.NodeID,
+		"CSI.NodeExpandVolume",
+		"ClientCSI.NodeExpandVolume",
+		structs.RateMetricWrite,
+		args,
+		reply,
+	)
+}
 
+func (a *ClientCSI) sendCSINodeRPC(nodeID, method, fwdMethod, op string, args any, reply any) error {
```
tgross, line 249: Nice :+1:

client/pluginmanager/csimanager/testing.go
```diff
@@ -0,0 +1,73 @@
...
+
+var _ VolumeManager = &MockVolumeManager{}
+
+type MockVolumeManager struct {
+	NextExpandVolumeErr  error
+	LastExpandVolumeCall *MockExpandVolumeCall
+}
+
+func (m *MockVolumeManager) MountVolume(_ context.Context, vol *nstructs.CSIVolume, alloc *nstructs.Allocation, usageOpts *UsageOptions, publishContext map[string]string) (*MountInfo, error) {
+	panic("implement me")
```
tgross, line 47: These are TODO for the follow-up refactoring PR, right?

tgross (commented):
This is looking really good @gulducat! I've made a comment on `nomad/csi_endpoint.go` that probably needs some thinking through.

client/pluginmanager/csimanager/testing.go
gulducat: Yep indeed!

nomad/csi_endpoint.go
gulducat: Yeah I've had awkward feelings about this too -- both that the user won't see it (it appears as a successful API call), and if something actually could be done to correct it, there's no mechanism to do so.

> Looking at the possible errors that can be returned by Node Expand, there are some errors that the operators needs to see to change their request.

I thiiink in reality, the values that the user can set (only min and max, maybe capabilities?) should be effectively validated by us and by the controller, before we move along to node expansion.  I think most of the validation in this PR is to wrap it in the expectations of the spec, much of which is redundant by this point in the process, as far as user input goes.

From what I can tell, the spec does not say how controller plugins should handle repeated identical calls, but for example, AWS has a pretty brutal rate limit on modifying EBS volumes (once per **6** hours(!)), so I erred on the side of caution there.

But I suppose I could just let it happen, if that's what the controller wants to do, and let the user see it?

As is, the hacky escape hatch for that would be to increment min_capacity by 1 \<byte or unit> to kick the whole thing again from the start, which could definitely be painful if you're having to wait 6 hours between attempts :grimacing: 

I could add new field(s) to the volume, to track intermediate state as you say, then retry only the node side of the equation if `min_capacity = real capacity` but the last attempt at node expansion was not successful.  or maybe new field(s) on the claim? :thinking:

tgross: Oh, I think we're in luck. From the spec on [`ControllerExpandVolume`](https://github.com/container-storage-interface/spec/blob/master/spec.md#controllerexpandvolume):

> This operation MUST be idempotent. If a volume corresponding to the specified volume ID is already larger than or equal to the target capacity of the expansion request, the plugin SHOULD reply 0 OK.

So I think it's safe for us to return the `NodeExpandVolume` error here and _not_ write the change to the Nomad state. That way the user can retry and if we end up sending the request to the controller plugin again, no worries.

tgross: And if we have more than one claim so that we end up with multiple `NodeExpandVolume` but only one of them fails, no worries there either, from the [spec](https://github.com/container-storage-interface/spec/blob/master/spec.md#nodeexpandvolume): 

> This operation MUST be idempotent. If a volume corresponding to the specified volume ID is already larger than or equal to the target capacity of the expansion request, the plugin SHOULD reply 0 OK.

gulducat: Gee wiz, how did I miss that line?  Alrighty, I'll go ahead and try that on a real EBS volume just out of paranoia, but that would sure be a convenient state of affairs!

gulducat: Yep, very good, the EBS controller plugin logs

```
cloud.go:1050] [Debug] Volume "vol-0eb67494a47872eb3" current size (15 GiB) is greater or equal to the new size (15 GiB)
```

so looks to behave its idempotent self.

And returning an error here, if a node plugin is offline:

```
$ nomad volume register ebs-vol.hcl
Error registering volume: Unexpected response code: 500 (unable to update volume: 1 error occurred:
        * CSI.NodeExpandVolume error: csi-node plugin 'aws-ebs0' did not become ready: context deadline exceeded)
```

which is quite useful -- hopefully enough for the operator to understand that they may need to check on their node plugin?  Then re-running the create/register tries everything again, since state was not updated in the error case.

gulducat: I squashed the change and force-pushed, but I added this test:
https://github.com/hashicorp/nomad/pull/18522/commits/76a5e55c966fce6f842bb8c3425d623c39d8933a#diff-214766af461ea207f297f2904489f01ac55b814950f301577516c1037501396fR1919-R1931

and it passes with the error returned here:
https://github.com/hashicorp/nomad/pull/18522/commits/76a5e55c966fce6f842bb8c3425d623c39d8933a#diff-6a3c30b17bb256ab9e3d5bab1620c3b9afee29aaa06371eae6cae0951a2fb12aL1275-R1275

tgross: LGTM :+1:

gulducat: Sweeeet, thanks for calling this out!

Unless there's anything else for me to address in this PR, mind pressing Approve for me?

tgross (approved):

nomad/csi_endpoint.go
tgross: Oops, forgot to do that. Done!

PR #18523, opened 2023-09-15
job endpoint: reorder check for disabled job registrations

tgross:
When job registrations are disabled, there's no reason to do the potentially expensive job mutation and admission hooks. Move the ACL resolution and this check before those hooks.


gulducat (approved):
I'm a little surprised this didn't trip more test cases, but the relevant behavior seems covered.
LGTM!

PR #12458, opened 2022-04-05
client: add Nomad template service functionality to runner. 

jrasell:
This change modifies the template task runner to utilise the
new consul-template which includes Nomad service lookup template
funcs.

In order to provide security and auth to consul-template, we use
a custom HTTP dialer which is passed to consul-template when
setting up the runner. This method follows Vault implementation.

This PR also includes an E2E test in lieu of integration tests. This
is due to the taskrunner/template pkg being in a long import chain
making utilising current test agents/clients impossible without a
large refactor. We will therefore step back and address this
properly in the future.

PR #18524, opened 2023-09-15
helper: reduce size of buffer used by template connections

tgross:
In #12458 we added an in-memory connection buffer so that template runners that want access to the Nomad API for Service Registration and Variables can communicate with Nomad without having to create a real HTTP client. The size of this buffer (1 MiB) was taken directly from its usage in Vault, and each connection makes 2 such buffers (send and receive). Because each template runner has its own connection, when there are large numbers of allocations this adds up to significant memory usage.

The largest Nomad Variable payload is 64KiB, and a small amount of metadata. Service Registration responses are much smaller, and we don't include check results in them (as Consul does), so the size is relatively bounded. We should be able to safely reduce the size of the buffer by a factor of 10 or more without forcing the template runner to make multiple read calls over the buffer.


schmichael (approved):
It's also worth noting that the performance expectations here should be awfully loose anyway. No one should be pull megabytes of state through templates as those sorts of assets are much better suited for the artifacts.

PR #18528, opened 2023-09-18
job endpoint: implicit constraints for multi-Vault/Consul

tgross:
Update the implicit constraint mutating hook to support multiple Vault and Consul clusters in Nomad Enterprise. This requires moving the Vault/Consul mutating hooks earlier in the list as well, because that'll ensure we've canonicalized properly for multiple clusters.

### DIFF
--- a/nomad/job_endpoint_hooks.go
+++ b/nomad/job_endpoint_hooks.go
@@ -5,11 +5,13 @@ package nomad
 
 import (
 	"fmt"
+	"sort"
 
 	"github.com/dustin/go-humanize"
 	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/nomad/helper"
 	"github.com/hashicorp/nomad/nomad/structs"
+	"golang.org/x/exp/maps"
 )
 
 const (
@@ -182,8 +184,11 @@ func (jobImpliedConstraints) Mutate(j *structs.Job) (*structs.Job, []error, erro
 	for _, tg := range j.TaskGroups {
 
 		// If the task group utilises Vault, run the mutator.
-		if _, ok := vaultBlocks[tg.Name]; ok {
-			mutateConstraint(constraintMatcherLeft, tg, vaultConstraint)
+		vaultTasks := maps.Keys(vaultBlocks[tg.Name])
+		sort.Strings(vaultTasks)
+		for _, vaultTask := range vaultTasks {
+			vaultBlock := vaultBlocks[tg.Name][vaultTask]
+			mutateConstraint(constraintMatcherLeft, tg, vaultConstraintFn(vaultBlock))
 		}
 
 		// Check whether the task group is using signals. In the case that it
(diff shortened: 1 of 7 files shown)


nomad/job_endpoint_hooks.go
```diff
@@ -182,8 +184,11 @@ func (jobImpliedConstraints) Mutate(j *structs.Job) (*structs.Job, []error, erro
 	for _, tg := range j.TaskGroups {
 
 		// If the task group utilises Vault, run the mutator.
-		if _, ok := vaultBlocks[tg.Name]; ok {
-			mutateConstraint(constraintMatcherLeft, tg, vaultConstraint)
+		vaultTasks := maps.Keys(vaultBlocks[tg.Name])
+		sort.Strings(vaultTasks)
```
lgfa29, line 188: Just out of curiosity, is the sorting relevant here?

nomad/job_endpoint_hooks.go
```diff
@@ -206,14 +211,58 @@ func (jobImpliedConstraints) Mutate(j *structs.Job) (*structs.Job, []error, erro
...
+// requested Vault cluster. This is to support Nomad Enterprise but neither the
+// fingerprint or non-default cluster are allowed well before we get here, so no
+// need to split out the behavior to ENT-specific code.
+func vaultConstraintFn(vault *structs.Vault) *structs.Constraint {
+	if vault.Cluster != "default" && vault.Cluster != "" {
+		return &structs.Constraint{
+			LTarget: fmt.Sprintf("${attr.vault.%s.version}", vault.Cluster),
+			RTarget: ">= 0.6.1",
+			Operand: structs.ConstraintSemver,
+		}
```
lgfa29, line 245: Would this generate a confusing placement error message? As a job submitter, I usually don't care about Vault version 😅 

Perhaps something like this generates a more meaningful error?

```suggestion
		return &structs.Constraint{
			LTarget: fmt.Sprintf("${attr.vault.%s}", vault.Cluster),
			Operand: structs.ConstraintAttributeIsSet,
		}
```

nomad/structs/services.go
```diff
@@ -1048,6 +1048,11 @@ func (s *Service) Equal(o *Service) bool {
 	return true
 }
 
+func (s *Service) IsConsul() bool {
+	return s.Provider == ServiceProviderConsul || s.Provider == ""
+
+}
```
lgfa29, line 1054: 🙌

lgfa29 (approved):

nomad/job_endpoint_hooks.go
tgross: It makes the error messages stable, which was necessary to deflake tests.

nomad/job_endpoint_hooks.go
tgross: I didn't want the constraint with non-default clusters to work any differently than the default cluster (at least for this PR), and this it the constraint we set for the default cluster today. Arguably we want to keep the semver check so that we're ensuring we're only landing on compatible versions of Vault, but that hasn't been meaningful in a long time because we're using very stable interfaces.

That being said, it'd be worth checking that this version constraint is still valid for the Workload Identity login -- when did Vault get JWT login capability?

lgfa29: > I didn't want the constraint with non-default clusters to work any differently than the default cluster (at least for this PR)

Makes sense 👍 

> when did Vault get JWT login capability?

0.10.4 it seems: https://github.com/hashicorp/vault/blob/main/CHANGELOG-v0.md#0104-july-25th-2018

tgross: Ok, for now I'm going to merge this and then follow-up with a PR where we define an official minimal Vault version including documentation later on.

Issue #18354, opened 2023-08-29
[cli] Allow "role" updates to tokens


philrenaud:
Currently, you can pass in both `-policy` and `-role-name` on `nomad acl token create`, but you can only modify `-policy` when running `nomad acl token update`.

The Nomad CLI should give users the ability to update roles for a given user token.

- [acl_token_create.go](https://github.com/hashicorp/nomad/blob/main/command/acl_token_create.go#L45-L53)
- [acl_token_update.go](https://github.com/hashicorp/nomad/blob/main/command/acl_token_update.go#L38-L40)

Kartik1397:
Hi @philrenaud @jrasell, Can I work on this?

jrasell:
Hi @Kartik1397 of course; although as you're not an internal member I cannot explicitly assign you the issue. If you have any questions, please reach out.

Kartik1397:
Hi @jrasell, I've submitted a pull request to introduce the ACL token update feature. I would greatly appreciate your review and any feedback you might have when you have the chance.

Thank you!

0xMALVEE:
is this issue fixed already? it's valid and I can take it for my first contribution? 
Or it's not needed.

burnerlee:
@jrasell may I take up this issue and raise a PR?

burnerlee:
@jrasell I've added a PR for the same

tgross:
Fixed in https://github.com/hashicorp/nomad/pull/18532

PR #18532, opened 2023-09-18
Add role update functionality to acl token update

Kartik1397:
This pull request adds the ability to update roles in the `acl token update` command.

Fixes: #18354

### DIFF
--- a/command/acl_token_update.go
+++ b/command/acl_token_update.go
@@ -12,6 +12,9 @@ import (
 
 type ACLTokenUpdateCommand struct {
 	Meta
+
+	roleNames []string
+	roleIDs   []string
 }
 
 func (c *ACLTokenUpdateCommand) Help() string {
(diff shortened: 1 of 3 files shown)


command/acl_token_update.go
```diff
@@ -122,6 +141,10 @@ func (c *ACLTokenUpdateCommand) Run(args []string) int {
 		token.Policies = policies
 	}
 
+	if len(c.roleNames) != 0 || len(c.roleIDs) != 0 {
+		token.Roles = generateACLTokenRoleLinks(c.roleNames, c.roleIDs)
+	}
```
lgfa29, line 146: One thing this makes me think is that these flags are additive, meaning they only allow adding new roles to the token, but there's no way to remove them.

Perhaps we also need something like `-unset-role-name` and `-unset-role-id`, similar to [`node meta -uset`](https://developer.hashicorp.com/nomad/docs/commands/node/meta/apply#unset). that would remove the given roles from the token?

command/acl_token_update.go
```diff
@@ -38,6 +41,12 @@ Update Options:
   -policy=""
     Specifies a policy to associate with the token. Can be specified multiple times,
     but only with client type tokens.
+
+  -role-id=""
+     ID of a role to use for this token. May be specified multiple times.
+
+  -role-name=""
+     Name of a role to use for this token. May be specified multiple times.
```
lgfa29, line 49:
```suggestion
  -role-id=""
    ID of a role to use for this token. Can be specified multiple times, but
    only with client type tokens.

  -role-name=""
    Name of a role to use for this token. Can be specified multiple times, but
    only with client type tokens.
```

Match the help message from `-policy`. I'm pretty sure the client type restriction also applies.

lgfa29 (changes requested):
Looking good @Kartik1397, thank you for the PR! I left a suggestion about how to handle the removal of roles from a token.

Would you also be able to update the documentation page to include these new flags?
https://github.com/hashicorp/nomad/blob/main/website/content/docs/commands/acl/token/update.mdx

And also run the command `make cl` to generate a CHANGELOG entry.

Thanks again for the contribution!

command/acl_token_update.go
tgross: The `-role-id` and `-role-name` flags are like the `-policy` flags: they completely replace the existing polices or roles if set. I've updated the documentation to make that clear.

tgross (approved):
I've rebased this PR on `main`, pushed a changelog entry, resolved the open comments, and added documentation updates. Will merge once this is green and that'll ship in Nomad 1.8.3.

PR #18533, opened 2023-09-18
vault: select Vault API client by cluster name

tgross:
Nomad Enterprise will support configuring multiple Vault clients. Instead of having a single Vault client field in the Nomad client, we'll have a function that callers can parameterize by the Vault cluster name that returns the correctly configured Vault API client wrapper.

Note: this will need a Nomad ENT PR as well.


client/client.go
```diff
@@ -2776,26 +2776,43 @@ func (c *Client) setupConsulTokenClient() error {
...
+	c.vaultClients = map[string]vaultclient.VaultClient{}
+	vaultConfigs := c.GetConfig().GetVaultConfigs(c.logger)
+	for _, vaultConfig := range vaultConfigs {
+		vaultClient, err := vaultclient.NewVaultClient(c.GetConfig().VaultConfig, c.logger, c.deriveToken)
+		if err != nil {
+			return err
+		}
+		if vaultClient == nil {
+			c.logger.Error("failed to create vault client", "name", vaultConfig.Name)
+			return fmt.Errorf("failed to create vault client")
```
lgfa29, line 2792:
```suggestion
			return fmt.Errorf("failed to create vault client for cluster %q", vaultConfig.Name)
```

client/client.go
```diff
@@ -264,8 +264,8 @@ type Client struct {
 	// Service Identity tokens through Nomad Server.
 	tokensClient consulApi.ServiceIdentityAPI
 
-	// vaultClient is used to interact with Vault for token and secret renewals
-	vaultClient vaultclient.VaultClient
+	// vaultClients is used to interact with Vault for token and secret renewals
+	vaultClients map[string]vaultclient.VaultClient
```
lgfa29, line 268: I think it's fine, but just noting it here in case I'm mistaken. We don't need a lock for this map because the only mutation happens in `NewClient` and accessing the map requires `NewClient` to be called beforehand.

lgfa29 (approved):

client/client.go
tgross: Right. `NewClient` is only ever called in `setupVaultClients`

PR #18534, opened 2023-09-18
vault: update task runner vault hook to support workload identity

lgfa29:
Update the task runner Vault hook and the Vault client so they are able to derive tokens using the task signed workload identity JWT. 

The new flow is used whenever a task has an identity with the pattern `vault_<cluster>`, where `<cluster>` matches the value of the `vault.cluster` config applied to the task.

When using JWT and workload identities, the `vault.token` value in the Nomad server configuration will likely be empty. This would cause the legacy Vault client in the Nomad servers to fail on start.

When a `vault.default_identity` is given for the default cluster, the Vault client in the servers are replaced by a no-op implementation since it won't be used in any way. 

Until Nomad 1.9, where the legacy flow will be removed, the `default` Vault cluster can still use the legacy flow. In Nomad ENT this allows operators to mix authentication flows, where the `default` cluster uses the legacy flow and additional clusters use the JWT flow. Non-default clusters are not allowed to use the legacy flow.

--------------

Note to reviewers: most code changes are for tests and moving things around to avoid circular dependencies. Main logic is implemented in https://github.com/hashicorp/nomad/pull/18534/commits/cf81d724d7ef66733d17b7603a0ba631fcd31b25


client/allocrunner/taskrunner/task_runner_hooks.go
```diff
@@ -90,15 +90,22 @@ func (tr *TaskRunner) initHooks() {
 
 	// If Vault is enabled, add the hook
 	if task.Vault != nil {
+		authMethod := vaultAuthMethodToken
+		if wid := task.GetIdentity("vault"); wid != nil {
+			authMethod = vaultAuthMethodJWT
+		}
```
tgross, line 96: The hook has access to the allocation, and therefore the task. It might be nice from the standpoint of encapsulation to push this method switching down into the hook rather than surfacing it in the hook config. But I don't feel strongly about it, especially if it's a lot of extra work to do.

nomad/server.go
```diff
@@ -1134,8 +1134,14 @@ func (s *Server) setupConsul(consulConfigEntries consul.ConfigAPI, consulACLs co
 
 // setupVaultClient is used to set up the Vault API client.
 func (s *Server) setupVaultClient() error {
+	vconfig := s.config.VaultConfig
+	if vconfig != nil && vconfig.UseIdentity != nil && *vconfig.UseIdentity {
+		s.vault = &NoopVault{}
+		return nil
+	}
```
tgross, line 1141: Good catch on this, and it makes an obvious bit of code to tear out for 1.9.

client/allocrunner/taskrunner/vault_hook.go
```diff
@@ -311,23 +328,76 @@ OUTER:
...
 		}
+	}
+}
 
+func (h *vaultHook) deriveVaultTokenJWT() (string, bool, error) {
+	// Wait for signed identity.
+	signed, err := h.widmgr.Get(cstructs.TaskIdentity{
+		TaskName:     h.taskName,
+		IdentityName: "vault",
+	})
```
tgross, line 369: Just checking my understanding here: we don't need to use `Watch` because we don't actually care about refreshed identities until we have to derive a brand new Vault token rather than renewing the Vault token, right?

client/vaultclient/vaultclient.go
```diff
@@ -256,6 +264,59 @@ func (c *vaultClient) DeriveToken(alloc *structs.Allocation, taskNames []string)
 	return tokens, nil
 }
 
+func (c *vaultClient) DeriveTokenWithJWT(ctx context.Context, reqs map[string]JWTLoginRequest) (map[string]string, error) {
```
tgross, line 267: Because the `vault_hook` is running at the task level, it looks like `len(reqs)` is always 1 for all callers. Why have this take a map instead of the `JWTLoginRequest` alone?

tgross (commented):
This is looking great @lgfa29. I've left a few comments around areas I'm unclear on

nomad/server.go
lgfa29: Yeah, at first I had `nomad.vaultClient` to be `nil` when `use_identity = true`, but that created a bunch of `nil` pointer exceptions 😅 

Having a no-op implementation feels safer and easier to remove in the future.

client/allocrunner/taskrunner/vault_hook.go
lgfa29: Yes, but also `Watch` can only be called once per identity, and the `identity` hook is already watching it. I'm OK either way, but to use `Watch` here we need to fix this:
https://github.com/hashicorp/nomad/pull/18431#discussion_r1330781301

client/vaultclient/vaultclient.go
```diff
@@ -256,6 +264,59 @@ func (c *vaultClient) DeriveToken(alloc *structs.Allocation, taskNames []string)
...
+	c.client.SetToken("")
+
+	var g multierror.Group
+	var tokensSync sync.Map
+
+	for k, v := range reqs {
+		name := k
+		req := v
+		g.Go(func() error {
+			s, err := c.client.Logical().WriteWithContext(ctx, "auth/jwt/login",
```
jorgemarey, line 288: Hi, I was looking into this and seeing this, i thought that in vault currently [only one jwks_url can be set](https://developer.hashicorp.com/vault/api-docs/auth/jwt#configure), so if that auth backend (auth/<backend>/login) is already used by another jwt backend this won't work. Can we set the <backend> configurable in the client configuration?

tgross: That's a good catch @jorgemarey. From https://developer.hashicorp.com/vault/api-docs/auth/jwt:

> This documentation assumes the plugin method is mounted at the /auth/jwt path in Vault. Since it is possible to enable auth methods at any location, please update your API calls accordingly.

lgfa29: > please update your API calls accordingly.

This feels very targeted to me 😄 

Thanks @jorgemarey, I would've 100% missed that!

@tgross I think we need a new field in the `vault` agent config to allow this value to be customized.

tgross: > @tgross I think we need a new field in the vault agent config to allow this value to be customized.

Agreed. I can pick that up.

client/allocrunner/taskrunner/template_hook.go
```diff
@@ -113,6 +116,10 @@ func (h *templateHook) Prestart(ctx context.Context, req *interfaces.TaskPrestar
 	h.vaultToken = req.VaultToken
 	h.nomadToken = req.NomadToken
 
+	if req.Task.Vault != nil {
+		h.vaultConfig = h.config.clientConfig.VaultConfigs[req.Task.Vault.Cluster]
+	}
+
 	// Set vault namespace if specified
 	if req.Task.Vault != nil {
```
tgross, line 124: Can we move this under the same conditional block as above?

client/allocrunner/taskrunner/template_hook.go
```diff
@@ -113,6 +116,10 @@ func (h *templateHook) Prestart(ctx context.Context, req *interfaces.TaskPrestar
 	h.vaultToken = req.VaultToken
 	h.nomadToken = req.NomadToken
 
+	if req.Task.Vault != nil {
+		h.vaultConfig = h.config.clientConfig.VaultConfigs[req.Task.Vault.Cluster]
```
tgross, line 120:
```suggestion
		h.vaultConfig = h.config.clientConfig.GetVaultConfigs()[req.Task.Vault.Cluster]
```

client/allocrunner/taskrunner/vault_hook.go
```diff
@@ -320,23 +337,77 @@ OUTER:
...
-		if err == nil {
-			return tokens[h.taskName], false
+		switch h.authMethod {
+		case vaultAuthMethodToken:
+			token, exit, err = h.deriveVaultTokenLegacy()
+		case vaultAuthMethodJWT:
+			token, exit, err = h.deriveVaultTokenJWT()
 		}
+		if err == nil || exit {
+			fmt.Println(err)
```
tgross, line 355: Debugging leftovers:

```suggestion
```

nomad/config.go
```diff
@@ -493,6 +493,15 @@ func (c *Config) VaultDefaultIdentity() *structs.WorkloadIdentity {
 	return workloadIdentityFromConfig(c.VaultConfig.DefaultIdentity)
 }
 
+func (c *Config) VaultIdentiyConfig(cluster string) *structs.WorkloadIdentity {
```
tgross, line 496:
```suggestion
func (c *Config) VaultIdentityConfig(cluster string) *structs.WorkloadIdentity {
```

nomad/structs/workload_id.go
```diff
@@ -38,7 +39,7 @@ var (
 	// be safe to use in filenames.
 	//
 	// Reuse validNamespaceName to save a bit of memory.
-	validIdentityName = validNamespaceName
+	validIdentityName = regexp.MustCompile("^[a-zA-Z0-9-//]{1,128}$")
```
tgross, line 42: I don't think the second `/` gets us anything here, as `/` doesn't need escaping:

```suggestion
	validIdentityName = regexp.MustCompile("^[a-zA-Z0-9-/]{1,128}$")
```

client/vaultclient/vaultclient.go
lgfa29: Two reasons: 

1. I thought we would need multiple Vault tokens per task. It turns out we don't 😅.
2. `DeriveToken` takes a list of tasks as well. But it also only ever gets called with a single task 🤷.

So I modified to only take a single `JWTLoginRequest` and return a `string.`

client/allocrunner/taskrunner/task_runner_hooks.go
lgfa29: Moved logic to `newVaultHook`. I'm now passing the whole task instead of just the task name.

client/allocrunner/taskrunner/vault_hook.go
lgfa29: As per internal discussions, `Get` was kept in #18431, so I will keep using it.

client/allocrunner/taskrunner/template_hook.go
lgfa29: I will moved the template changes to a new PR, but I will make sure this gets addressed.

client/allocrunner/taskrunner/vault_hook.go
lgfa29: Ah yes, I left this to remind myself that I needed to handle this error (it isn't immediately obvious that can also hit this path when `err != nil`).

But I refactored this code to make it simpler and more obvious when errors need to be handled 👍

nomad/config.go
lgfa29: I managed to get a through 1.6 cycle without committing a single `NodePoo` or `NodePoll` but it seems like `identity` is the real challenge 😅

nomad/structs/workload_id.go
lgfa29: Fixed in https://github.com/hashicorp/nomad/pull/18591/files#diff-d2ec4e0bb147da1d9abe8f14084958a2d4bc16b5cb1cf28f5a07c7f2bb3331e5R44.

For the record, we realized that using slashes in the identity name would be a problem because of how they're written to disk (`<task dir>/<secrets name>`) it would try to create a new directory.

client/allocrunner/taskrunner/template_hook.go
lgfa29: Same as [above](https://github.com/hashicorp/nomad/pull/18534#discussion_r1341775852)

client/allocrunner/taskrunner/template_hook.go
lgfa29: Applied in https://github.com/hashicorp/nomad/commit/65852f405f46f1bb2d0511252338ba2b9301304e#diff-23a9c60db0b49fdf4f4e68fd78ac41ba1eeaccf2b463b9898d25beb7632dc24fL116-L119

client/allocrunner/taskrunner/vault_hook_test.go
```diff
@@ -3,9 +3,567 @@
...
+		var resp interfaces.TaskPrestartResponse
+
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		t.Cleanup(cancel)
+
+		// Set unrecoverable error.
+		mockVaultClient.SetDeriveTokenWithJWTFn(func(_ context.Context, _ vaultclient.JWTLoginRequest) (string, error) {
+			// Cancel the context to simulate the task being killed.
+			cancel()
+			return "", structs.NewRecoverableError(errors.New("unrecoverable test error"), false)
```
tgross, line 346: Shouldn't this be an unrecoverable error?

client/allocrunner/taskrunner/vault_hook_test.go
```diff
@@ -3,9 +3,567 @@
...
+	})
+}
+
+func TestTaskRunner_VaultHook_tokenRenewalFail(t *testing.T) {
+	ci.Parallel(t)
+
+	testCases := []struct {
+		name                string
+		vaultBlock          *structs.Vault
+		verifyTaskLifecycle func(*trtesting.MockTaskHooks) error
```
tgross, line 462: I really like how this test works :+1:

client/vaultclient/vaultclient_test.go
```diff
@@ -0,0 +1,259 @@
...
+	rolePath := fmt.Sprintf("auth/%s/role/nomad-workloads", jwtAuthMountPathTest)
+	_, err = logical.WriteBytesWithContext(ctx, rolePath, []byte(widVaultRole))
+	if err != nil {
+		return fmt.Errorf("failed to write role: %w", err)
+	}
+
+	return nil
+}
+
+func TestVaultClient_DeriveTokenWithJWT(t *testing.T) {
```
tgross, line 155: Do we have a test helper that can skip the test if there's no Vault binary around? That might be nice to reduce wait times and spurious errors for non-CI development.

tgross (approved):
LGTM, nice work on all the added tests. My only potential concern is the handling of recoverable/non-recoverable errors. But once you're happy with that I'm :+1: to merge.

client/allocrunner/taskrunner/vault_hook.go
```diff
@@ -364,6 +389,55 @@ func (h *vaultHook) deriveVaultToken() (token string, exit bool) {
...
+		return "", structs.NewRecoverableError(
+			fmt.Errorf("failed to retrieve signed workload identity: %w", err),
+			true,
+		)
+	}
+	if signed == nil {
+		return "", structs.NewRecoverableError(
+			errors.New("no signed workload identity available"),
+			true,
+		)
```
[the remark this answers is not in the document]
lgfa29: In theory the `WIDManager`may not have the identities because of how they are only fetched when `widmgr.Run()` is called, but in practice that doesn't seem like it would be currently possible as it's created and started at the alloc level.

I can change it to an unrecoverable error.

client/allocrunner/taskrunner/vault_hook.go
```diff
@@ -364,6 +389,55 @@ func (h *vaultHook) deriveVaultToken() (token string, exit bool) {
...
+	// Derive Vault token with signed identity.
+	token, err := h.client.DeriveTokenWithJWT(h.ctx, vaultclient.JWTLoginRequest{
+		JWT:  signed.JWT,
+		Role: h.vaultBlock.Role,
+	})
+	if err != nil {
+		return "", structs.WrapRecoverable(
+			fmt.Sprintf("failed to derive Vault token for identity %s: %v", h.widName, err),
+			err,
+		)
```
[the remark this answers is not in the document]
lgfa29: This could also happen due to an intermittent connectivity issue (network blip, Vault agent restart etc.) so making it recoverable allow us to retry.

client/allocrunner/taskrunner/vault_hook_test.go
lgfa29: It's kind of a bad API 😅 

That magic boolean set to `false` is what makes this error unrecoverable. I think it would be better to have two functions `NewRecoverableError`and `NewUnrecoverableError` that abstract away the boolean.

client/vaultclient/vaultclient_test.go
lgfa29: Good point, I will poke around other tests like this 👍

client/allocrunner/taskrunner/vault_hook_test.go
tgross: Gotcha. :+1:

client/vaultclient/vaultclient_test.go
lgfa29: I didn't find any check like this so I updated `NewTestVault` in https://github.com/hashicorp/nomad/pull/18534/commits/be3cedd89c52a7083315dd14959769db7fd124a4 to skip the test if the binary is not found.

lgfa29:
Rebased against `main` to fix conflict.

PR #18535, opened 2023-09-19
OpenID Configuration Discovery Endpoint

schmichael:
WIP branch to test against AWS

Issuer (and JWKS base address) is configurable with `oidc_issuer = "..."` in your agent config.

Endpoint is at http://localhost:4646/.well-known/openid-configuration

### DIFF
--- a/command/agent/keyring_endpoint.go
+++ b/command/agent/keyring_endpoint.go
@@ -135,7 +135,7 @@ func (s *HTTPServer) OIDCDiscoveryRequest(resp http.ResponseWriter, req *http.Re
 		Keys:                jwksPath,
 		RequestParameter:    false,
 		RequestURIParameter: false,
-		IDTokenAlgs:         []string{structs.PubKeyAlgEdDSA},
+		IDTokenAlgs:         []string{structs.PubKeyAlgRS256, structs.PubKeyAlgEdDSA},
 		ResponseTypes:       []string{"code"},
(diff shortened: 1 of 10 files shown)


command/agent/keyring_endpoint.go
```diff
@@ -77,6 +78,67 @@ func (s *HTTPServer) JWKSRequest(resp http.ResponseWriter, req *http.Request) (a
...
+	}
+
+	conf := s.agent.GetConfig()
+
+	//FIXME(schmichael) should we bother implementing an RPC just to get region
+	//forwarding? I think *not* since consumers of this endpoint are code that is
+	//intended to be talking to a specific region directly.
+	if args.Region != conf.Region {
+		return nil, CodedError(400, "Region mismatch")
+	}
```
tgross, line 102: This is a HTTP endpoint, so clients expose it too, right? Don't we need RPC forwarding for that to work?

schmichael: No because there's no RPC involved at all: any _agent_ regardless of whether it's client or server can serve this request...

...but for operational simplicity I should probably make it an RPC and server-only. That way you only have to configure the `issuer` on your servers, and we can avoid having a situation where every agent in a cluster must have the exact same config value or risk breaking something.

schmichael:
Replaced by #18882 and #18691

PR #18540, opened 2023-09-20
config: remove `consul.template_identity` in lieu of `task_identity`

tgross:
The original thinking for Workload Identity integration with Consul and Vault was that we'd allow `template` blocks to specify their own identity. But because the login to Consul/Vault to get tokens happens at the task level, this would involve making the `template` block a new WID watcher on its own rather than using the Consul and Vault hooks we're building at the group/task level.

So it doesn't make sense to have separate identities for individual `template` blocks rather than at the level of tasks. Update the agent configuration to rename the `template_identity` to the more accurate `task_identity`, which will be used for any non-service hooks (just `template` today).

Update the implicit identities job mutation hook to create the identity we'll need as well.


lgfa29 (approved):
LGTM!

PR #18551, opened 2023-09-21
func: add go-census schema with billable metric

Juanadelacuesta:


tgross:
Leaving a note here just in case anyone from the community rolls by this closed PR and feels anxious about the mysterious private `hashicorp/go-census` dependency. This library is part of Nomad Enterprise usage-based billing, similar to Vault's. This PR was meant to have been opened in our private Enterprise repo for that use only.

PR #18553, opened 2023-09-21
[sentinel] Add existing job to enforceSubmitJob

angrycub:
This is the CE work required for https://github.com/hashicorp/nomad-enterprise/pull/1136


tgross (approved):
LGTM! We probably should add a changelog entry too

PR #18554, opened 2023-09-21
test: Refactor mock CSI manager

gulducat:
Right up front: I have mixed feelings about this.  Let me know if I should just throw it out :yum: 

Instead of separately implementing csimanager
interfaces in two separate places:
* client/allocrunner/csi_hook_test
* client/csi_endpoint_test

they can both use the same mocks defined in
client/pluginmanager/csimanager/
alongside the actual implementations of them.

---

My misgivings: my desire to leave the csi_hook tests mostly unchanged has resulted in a sorta franken-mock where ExpandVolume sticks out as different, since it tracks LastExpandVolumeCall, which I wanted in client/csi_endpoint, whereas allocrunner/csi_hook does call counting.

I'm not thrilled with the result... I might prefer proper call recording across the board, which could also cover call counts, but I should probably move on to the next priority that needs attention.  If the state of this is distasteful, I'd be willing to just throw it away.  :shrug:

### DIFF
--- a/client/allocrunner/csi_hook_test.go
+++ b/client/allocrunner/csi_hook_test.go
@@ -4,29 +4,24 @@
 package allocrunner
 
 import (
-	"context"
 	"errors"
 	"fmt"
-	"path/filepath"
-	"sync"
 	"testing"
 	"time"
 
 	"github.com/hashicorp/nomad/ci"
 	"github.com/hashicorp/nomad/client/allocrunner/interfaces"
 	"github.com/hashicorp/nomad/client/allocrunner/state"
-	"github.com/hashicorp/nomad/client/pluginmanager"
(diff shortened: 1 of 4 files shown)


tgross (approved):
LGTM!

> My misgivings: my desire to leave the csi_hook tests mostly unchanged has resulted in a sorta franken-mock where ExpandVolume sticks out as different, since it tracks LastExpandVolumeCall, which I wanted in client/csi_endpoint, whereas allocrunner/csi_hook does call counting.

I get where you're going with that worry, but it also suggests that some of the `csi_hook` tests could benefit from checking the last call arguments themselves, and not just the counts. It's not 100% consistent, but it's obvious where someone would add that feature to the mock down the road if they need it later. I'm :+1: to merge as-is.

PR #18556, opened 2023-09-22
config: move Consul getter out of fingerprinter

tgross:
This changeset moves the getter function for multiple Consul cluster configs out of the fingerprinter and into code that can be shared with the rest of the client. This matches the work done for Vault configuration in https://github.com/hashicorp/nomad/pull/18533 but has no consumers yet (the consuming code is going to be more complex than #18533 so I want to have this as its own smaller set of work.

This PR will need a matching ENT PR after it's been merged.


pkazmierczak (approved):
LGTM

PR #18557, opened 2023-09-22
client: new consul_hook

pkazmierczak:
This PR introduces a new allocrunner-level `consul_hook` which iterates over services and, if their provider is `consul`, fetches consul tokens for all of them, stores them in `AllocHookResources` and in task secret dirs. 

Caution: This code depends on service identity names being unique, which will be addressed in a follow-up PR.


client/allocrunner/alloc_runner_hooks.go
```diff
@@ -114,12 +114,20 @@ func (ar *allocRunner) initRunnerHooks(config *clientconfig.Config) error {
 	// newNetworkHook.
 	builtTaskEnv := newEnvBuilder().Build()
 
+	// Create a new consul client that works with workload identity based
+	// authentication (Nomad 1.7+)
+	newConsulClient, err := consul.NewConsulClient(ar.clientConfig.ConsulConfig, hookLogger)
+	if err != nil {
+		return fmt.Errorf("failed to initialize consul client: %v", err)
+	}
```
tgross, line 122: To support multi-cluster scenarios, the hook is going to need to be able to create multiple clients. So it's probably better if we pass the results of [`ar.clientConfig.GetConsulConfigs()`](https://github.com/hashicorp/nomad/blob/main/client/config/config_ce.go#L30) into the hook and then the hook can instantiate the clients it needs in `Prerun`, rather than creating the client here.

client/allocrunner/consul_hook.go
```diff
@@ -0,0 +1,165 @@
...
+		h.logger.Error(err.Error())
+		return err
+	}
+
+	mErr := multierror.Error{}
+	tokens := map[string]string{}
+
+	for _, tg := range job.TaskGroups {
+		for _, task := range tg.Tasks {
+			req, err := h.prepareConsulClientReq(task)
```
tgross, line 82: This method handles services defined at the task level, but we'll need to handle the services defined at the group level too.

client/allocrunner/alloc_runner_hooks.go
pkazmierczak: Right! I have a proposed solution in ea4d82b936ba1e8a404094264fb7d9819be752c4, have a look.

client/allocrunner/consul_hook.go
pkazmierczak: Right! Addressed in 06536f9474ed02ced4a1fba0ff8ef89630af2122

client/allocrunner/consul_hook.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+	}
+
+	// store the tokens in hookResources
+	h.hookResources.SetConsulTokens(tokens)
+
+	return mErr.ErrorOrNil()
+}
+
+func (h *consulHook) prepareConsulTokens(service *structs.Service, clients map[string]consul.Client) error {
```
tgross, line 113: This doesn't return tokens, so the map of `tokens` in the caller is never populated.

client/allocrunner/consul_hook_ce.go
```diff
@@ -0,0 +1,21 @@
...
+	structsc "github.com/hashicorp/nomad/nomad/structs/config"
+)
+
+func getConsulClients(configs map[string]*structsc.ConsulConfig, logger hclog.Logger) (map[string]consul.Client, error) {
+	conf := configs["default"] // Nomad CE only supports a single Consul client
+
+	client, err := consul.NewConsulClient(conf, logger)
+
+	return map[string]consul.Client{"default": client}, err
+}
```
tgross, line 21: It occurs to me that because we're returning a map of clients anyways, and the configs we get use CE/ENT-specific code to get the list of configs, you could just iterate over that map of configs to get the clients and safely end up with the right default for CE. And that would let you avoid having to have a separate `consul_hook_ce.go` and `consul_hook_ent.go`.

client/allocrunner/consul_hook.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	for _, tg := range job.TaskGroups {
+		for _, service := range tg.Services {
+			t, err := h.prepareConsulTokens(service, clients)
+			if err != nil {
+				mErr.Errors = append(mErr.Errors, err)
+				continue
+			}
+			tokens[service.Cluster] = t
```
tgross, line 89: If we use `service.Cluster` as the key, then if we have multiple services in the same cluster in the same group/task we'd overwrite the token. And each token has its own policies generated on the Consul side, so that won't work. I think we need to key these by something unique to the service in this allocation. (Ex. the group + task + service names or something like that.)

client/allocrunner/consul_hook_ce.go
pkazmierczak: addressed in c5ed8c9175e52566c9f35519df77e0516a808eb6

tgross (approved):
LGTM!

I've left a comment about filtering out tokens we write to the secretsdir, but I'm also totally fine with doing that in a follow-up PR.

PR #18559, opened 2023-09-22
consul: provide workload's Consul token to service client

tgross:
This is a work-in-progress changeset to provide workload-specific Consul tokens that are created by the `consul_hook` and attached to workload registration requests by the `group_service_hook` and `service_hook`.

This requires unreleased updates to Consul's `api` package, so this changeset includes a temporary `replace` directive in the go.mod file. Note this doesn't actually pass the token from the hooks, which will be in a follow-up PR after https://github.com/hashicorp/nomad/pull/18557 lands


lgfa29 (approved):
LGTM! My main question is about using the Consul API with empty token values, which I think would happen when `getServiceToken` is called but WID is not being used for Consul?

command/agent/consul/service_client.go
```diff
@@ -1498,8 +1556,15 @@ func (c *ServiceClient) AllocRegistrations(allocID string) (*serviceregistration
 // UpdateTTL is used to update the TTL of a check. Typically this will only be
 // called to heartbeat script checks.
 func (c *ServiceClient) UpdateTTL(id, namespace, output, status string) error {
+	var token string
+	check := c.checks[id]
+	if check != nil {
+		token = c.getServiceToken(check.ServiceID)
```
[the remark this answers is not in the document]
tgross: When the Consul client is constructed in `command/agent/agent.go`, we configure it with the token from the Nomad agent's `consul` block. Requests without the `Token` field fallback to that token.

Note that the `Token` field is already empty in the existing `UpdateTTLOpts` call (and all the others here), so returning an empty string here is the same zero-value for the field and the behavior is unchanged.

command/agent/consul/service_client.go
```diff
@@ -456,6 +456,12 @@ type ServiceClient struct {
 	allocRegistrations     map[string]*serviceregistration.AllocRegistration
 	allocRegistrationsLock sync.RWMutex
 
+	// serviceTokens is a map of service ID -> Consul tokens,
+	// where the token is set in the Workload by the client hooks when using
+	// Workload Identity. Access should be protected by allocRegistrationsLock
+	serviceTokens     map[string]string
```
pkazmierczak, line 462: Would this be where the new `consul_hook` stores the tokens it generates?

tgross: No, the `consul_hook` will write them to... hm, hook resources? disk? But then the `group_service_hook` and `service_hook` will add a map of service names to Consul tokens to their `RegisterWorkload` call. And then this PR is doing the work of taking those tokens from the `RegisterWorkload` args and sticking them into this map.

PR #18561, opened 2023-09-22
backport test: Refactor mock CSI manager into release/1.5.x

gulducat:
manual backport of #18554 - 7bd5c6e84e to 1.5.x

auto backport should work for 1.4.x from here.


pkazmierczak (approved):
I give a conditional LGTM as long as we all agree we're backporting changes like this. My understanding was that up until now these sort of changes weren't backported, but I guess having them in previous releases makes other backports less prone to conflicts?

PR #18570, opened 2023-09-25
backport csi: check controller health early into release/1.5.x

gulducat:
backporting bugfixes discussed in #18359, particularly [here](https://github.com/hashicorp/nomad/pull/18359/files#r1318732675) and [here](https://github.com/hashicorp/nomad/pull/18359/files#r1318744466).

this makes both create and register get the same
error, early in the process, when a controller
plugin will be needed but is offline.

also remove a stray ineffective CSIVolumeDenormalize


tgross (approved):
LGTM

You'll need to create a new changelog entry for this item, which will want to land in the other backport PR https://github.com/hashicorp/nomad/pull/18571 too.

PR #18572, opened 2023-09-25
server: Add reporting option to agent

Juanadelacuesta:
As part of the License Utilization Reporting, a new option for enterprise must be added to enable automatic report. This PR introduces this configuration, it most be enabled by default on ent. This PR will be merged into ENT where the reporting logic will be executed.

Related: [#1106](https://github.com/hashicorp/nomad-enterprise/issues/1106)


command/agent/config_parse.go
```diff
@@ -65,6 +65,7 @@ func ParseConfigFile(path string) (*Config, error) {
 		Telemetry: &Telemetry{},
 		Vault:     &config.VaultConfig{},
 		Vaults:    map[string]*config.VaultConfig{},
+		Reporting: config.DefaultReporting(),
```
lgfa29, line 68:
```suggestion
		Reporting: &config.Reporting{},
```

It doesn't matter all that much, but for consistency we initialize this struct with zero values

command/agent/config_parse.go
```diff
@@ -327,10 +328,14 @@ func extraKeys(c *Config) error {
...
 	}
 
-	// The `vault` and `consul` blocks are parsed separately from the Decode method, so it
+	// Remove reporting extra keys
+	c.ExtraKeysHCL = slices.DeleteFunc(c.ExtraKeysHCL, func(s string) bool { return s == "license" })
+
+	// The`vault` and `consul` blocks are parsed separately from the Decode method, so it
 	// will incorrectly report them as extra keys, of which there may be multiple
 	c.ExtraKeysHCL = slices.DeleteFunc(c.ExtraKeysHCL, func(s string) bool { return s == "vault" })
 	c.ExtraKeysHCL = slices.DeleteFunc(c.ExtraKeysHCL, func(s string) bool { return s == "consul" })
```
lgfa29, line 337:
```suggestion
	// The following blocks are parsed separately from the Decode method, so they must be removed from extra
	// keys to avoid incorrect report.
	c.ExtraKeysHCL = slices.DeleteFunc(c.ExtraKeysHCL, func(s string) bool { return s == "license" })
	c.ExtraKeysHCL = slices.DeleteFunc(c.ExtraKeysHCL, func(s string) bool { return s == "vault" })
	c.ExtraKeysHCL = slices.DeleteFunc(c.ExtraKeysHCL, func(s string) bool { return s == "consul" })
```

I think these are all removed for the same reason.

nomad/structs/config/reporting.go
```diff
@@ -0,0 +1,80 @@
...
+
+package config
+
+import "github.com/hashicorp/nomad/helper/pointer"
+
+func DefaultReporting() *Reporting {
+	return &Reporting{
+		&LicenseConfig{},
+	}
+}
```
lgfa29, line 12: Since this is about the `Reporting` struct it would be better if it were placed below, next to the other `Reporting` methods (or maybe pull the other methods up here).

nomad/structs/config/reporting.go
```diff
@@ -0,0 +1,80 @@
...
+
+type LicenseConfig struct {
```
lgfa29, line 14:
```suggestion
type LicenseReportingConfig struct {
```

`LicenseConfig` seems a bit too generic and may clash if we ever want to add license configuration values.

nomad/structs/config/reporting.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	if b.Enabled != nil {
+		result.Enabled = b.Enabled
+	}
+
+	return &result
+}
+
+type Reporting struct {
```
lgfa29, line 47:
```suggestion
type ReportingConfig struct {
```

It stutters a bit, but I think all other structs in this package end in `Config`.

lgfa29 (approved):
So minor nit-picks, but LGTM!

command/agent/config_parse.go
Juanadelacuesta: I thought the same, but Tim corrected me, for vault and consul they are different because multiple blocks are allowed, so they are their own special case

nomad/structs/config/reporting.go
Juanadelacuesta: sure!

PR #18573, opened 2023-09-25
CSI: Document volume expansion

### LINKED ISSUE #10324: CSI: implement ExpandVolume

gulducat:
and show `Capacity` (the actual capacity of the volume) in the `volume status <id>` command output.

I belieeeve this finally closes #10324!


website/content/docs/other-specifications/volume/index.mdx
```diff
@@ -221,6 +238,73 @@ context {
...
+Check the volume capacity again:
+
+```shell-session
+$ nomad volume status ebs_prod_db1 | grep Capacity
+Capacity             = 100 GiB
+```
+
+If you encounter errors that are not clear from the command output,
+the Nomad leader server logs may be helpful, and/or CSI plugin
+allocation logs.
```
tgross, line 306:
```suggestion
If you encounter errors that are not clear from the command output,
the CSI plugin allocation logs and/or Nomad leader server logs may be 
helpful.
```

website/content/docs/other-specifications/volume/index.mdx
```diff
@@ -173,6 +175,21 @@ fields can be updated after the volume is registered:
...
+## Volume Expansion
+
+CSI Volumes may be expanded (increased in size) if the CSI controller plugin
+(and node plugin, if required) has the `EXPAND_VOLUME` capability.
+
+To trigger a volume expansion, increase [`capacity_min`](#capacity_min)
+above the current real capacity of the volume (as seen with the
+[`volume status`][] command), and re-issue either [`volume create`][]
+or [`volume register`][] commands, whichever you used to set up the volume
+initially.
```
tgross, line 187: Does it actually matter which they used initially? If you `volume create` can you `volume register` to make that change later?

tgross (approved):
LGTM

website/content/docs/other-specifications/volume/index.mdx
gulducat: You totally can, bearing in mind the caveats about what to include or not include in the volume spec, depending whether you're doing create or register.  I thought it simpler to suggest "just use the same spec, up the capacity, and run the same command," but although they follow separate code paths, they're ~roughly interchangeable after initial setup.

It would be nice to bring create and register in ever greater alignment at some point, where one may basically be an alias of the other, but I figure that's a journey for another day.

tgross: Ok, but the language you're using here feels more prescriptive than that to me is all.

gulducat: Fair point, I'll loosen it up a bit

PR #18584, opened 2023-09-26
fix: use interpolated address when performing health checks

### LINKED ISSUE #18574: Nomad native service health checks fail when using interpolated service address in job
### Nomad version
Output from `nomad version`
1.6.2

Linux

### Issue
Screenshot 2023-09-25 at 6 19 18 PM

### Reproduction steps
1. Use Nomad service discovery
2. Set the address to a variable like the gce public IP address
3. Add an HTTP health check

#### Expected Result
The health checker uses the variable address from the service catalog (/v1/service/:name output:)
Screenshot_2023-09-25_at_6 34 35_PM

#### Actual Result
The health checks always fail with a parse error (using the string literal value)

### Job file (if appropriate)
Screenshot 2023-09-25 at 6 36 45 PM

MattJustMatt:
Fixes #18574

### DIFF
--- a/.changelog/18584.txt
+++ b/.changelog/18584.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-fix: use interpolated address when performing health checks
+services: use interpolated address when performing nomad service health checks
 ```
--- a/.changelog/18584.txt
+++ b/.changelog/18584.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+fix: use interpolated address when performing health checks
+```
--- a/client/allocrunner/alloc_runner_hooks.go
+++ b/client/allocrunner/alloc_runner_hooks.go
@@ -111,7 +111,7 @@ func (ar *allocRunner) initRunnerHooks(config *clientconfig.Config) error {
 	}
 
 	// Create a taskenv.TaskEnv which is used for read only purposes by the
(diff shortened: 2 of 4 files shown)


jrasell (changes requested):
Hi @MattJustMatt and thanks for the PR. The approach seems right to me, but. the change in function signature has broken some tests, which will need to be fixed up. It would also be nice to add some additional tests to cover the new behaviour and ensure interpolated service checks work as we now expected. Finally, a changelog entry for the addition can be included by running the `make cl` command.


Thanks again!

MattJustMatt:
@jrasell Thanks for the review! 

I've fixed the broken tests and added the changelog entry. I don't have a good enough understanding of the test infra to easily add a regression test for this, but if it's needed for this PR please let me know and I can dive in deeper.

.changelog/18584.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+fix: use interpolated address when performing health checks
```
shoenig, line 2:
```suggestion
services: use interpolated address when performing nomad service health checks
```

MattJustMatt:
100% cool with me!

shoenig (approved):

jrasell (approved):

PR #18586, opened 2023-09-26
WI: add test to verify we don't allow empty signatures for JWT

tgross:
Encoded JWTs include an `alg` header key that tells the verifier which signature algorithm to use. Bafflingly, the JWT standard allows a value of `"none"` here which bypasses signature verification.

In all shipped versions of Nomad, we explicitly configure verification to a specific algorithm and ignore the header value entirely to avoid this protocol flaw. But in #18123 we updated our JWT library to `go-jose`, which rightfully doesn't support `"none"` but this detail isn't encoded anywhere in our code base. Add a test that ensures we catch any regressions in the library.

### DIFF
--- a/nomad/encrypter_test.go
+++ b/nomad/encrypter_test.go
@@ -6,6 +6,9 @@ package nomad
 import (
 	"bytes"
 	"context"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


lgfa29 (approved):
Nice catch!

schmichael (approved):
Great idea! Thank you!

PR #18588, opened 2023-09-26
csi: fix volume updating behavior

gulducat:
fix for part of: c6dbba7cde911bb08f1f8da445a44a0125cd2047 (specifically around [here](https://github.com/hashicorp/nomad/commit/c6dbba7cde911bb08f1f8da445a44a0125cd2047#diff-53c7d524897a192ece819816a9dcf006dee46747fbb853e7f638601ea556a72cL2433-L2435))
which allowed updating volumes while in use,
because CSI expand may occur while in use.
but it mistakenly stopped copying other
important values that may be updated
whether in use or not.

this moves some of the in-use validation to
happen during Merge(), before writing to state,
leaving UpsertCSIVolume with only minimal final
sanity-checking.

### DIFF
--- a/nomad/state/state_store.go
+++ b/nomad/state/state_store.go
@@ -2428,18 +2428,10 @@ func (s *StateStore) UpsertCSIVolume(index uint64, volumes []*structs.CSIVolume)
 				old.Provider != v.Provider {
 				return fmt.Errorf("volume identity cannot be updated: %s", v.ID)
 			}
-
-			// Update fields that are safe to change while volume is being used.
-			if err := old.UpdateSafeFields(v); err != nil {
-				return fmt.Errorf("unable to update in-use volume: %w", err)
-			}
-			v = old
-			v.ModifyIndex = index
-
 		} else {
 			v.CreateIndex = index
-			v.ModifyIndex = index
 		}
+		v.ModifyIndex = index
 
 		// Allocations are copy on write, so we want to keep the Allocation ID
 		// but we need to clear the pointer so that we don't store it when we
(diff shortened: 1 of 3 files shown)


nomad/state/state_store.go
```diff
@@ -2428,18 +2428,10 @@ func (s *StateStore) UpsertCSIVolume(index uint64, volumes []*structs.CSIVolume)
...
 				return fmt.Errorf("volume identity cannot be updated: %s", v.ID)
 			}
-
-			// Update fields that are safe to change while volume is being used.
-			if err := old.UpdateSafeFields(v); err != nil {
-				return fmt.Errorf("unable to update in-use volume: %w", err)
-			}
-			v = old
-			v.ModifyIndex = index
-
```
tgross, line 2438: It looks like we've moved the logic we wanted here into `Merge`, but then don't we need to add the call to `Merge` here?

gulducat: Nah, we validate and build up new `CSIVolume`(s) in the register/create flows via [`reconcileVolume` -> `Merge`](https://github.com/hashicorp/nomad/blob/4895d708/nomad/csi_endpoint.go#L388-L392) prior to `UpsertCSIVolume` here.  Validation should always happen before getting this far, so "store this in state please" (this upsert method) only handles writing state (and a couple remaining validations above that I opted not to touch).

gulducat: It's that `v = old` line on 2436 that made any updating necessary here (via `UpdateSafeFields`), but that was the source of other intended updates not getting saved.  Without `v = old`, `old` is used only for some last-minute sanity-checking, then the already-merged-earlier `v`olume provided by the caller is what gets committed to state.

tgross (approved):
LGTM!

PR #18591, opened 2023-09-27
vault: update identity name to start with `vault_`

lgfa29:
Another workstream split from #18534.

In the original proposal, workload identities used to derive Vault tokens were expected to be called just `vault`. But in order to support multiple Vault clusters it is necessary to associate identities with specific Vault cluster configuration.

This PR implements a new proposal to have Vault identities named as `vault_<cluster>`.

To avoid the use of the hard coded string `"default"`, a new constant named `VaultDefaultCluster` was created. https://github.com/hashicorp/nomad/commit/4be533d69223c239e5e9ec5650855902c35358eb updates previous uses of `"default"` in the context of a Vault cluster to use this new constant and may ignored or reviewed separately to reduce noise. 

I wasn't sure if this is something we would want to have so we can also easily undo this change. I don't have any strong feelings either way. If we do want use the constant everywhere I can work on a similar PR for Consul.

### DIFF
--- a/nomad/job_endpoint_hook_implicit_identities.go
+++ b/nomad/job_endpoint_hook_implicit_identities.go
@@ -116,7 +116,8 @@ func (h jobImplicitIdentitiesHook) handleVault(t *structs.Task) {
 	}
 
 	// Use the Vault identity specified in the task.
-	vaultWID := t.GetIdentity(t.Vault.IdentityName())
+	vaultWIDName := t.Vault.IdentityName()
+	vaultWID := t.GetIdentity(vaultWIDName)
 	if vaultWID != nil {
 		return
 	}
@@ -131,6 +132,6 @@ func (h jobImplicitIdentitiesHook) handleVault(t *structs.Task) {
 	}
 
 	// Set the expected identity name and inject it into the task.
-	vaultWID.Name = t.Vault.IdentityName()
+	vaultWID.Name = vaultWIDName
 	t.Identities = append(t.Identities, vaultWID)
 }
--- a/nomad/job_endpoint_hook_implicit_identities.go
+++ b/nomad/job_endpoint_hook_implicit_identities.go
@@ -110,9 +110,6 @@ func (h jobImplicitIdentitiesHook) handleConsulTasks(t *structs.Task, consul *st
 //  1. The task has a Vault block.
 //  2. The server is configured with `vault.use_identity = true` and a
 //     `vault.default_identity` is provided.
-//
-// If the task already has an identity named `vault` it sets the identity name
-// to the expected value.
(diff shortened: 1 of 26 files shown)


nomad/job_endpoint_hook_implicit_identities.go
```diff
@@ -111,31 +110,27 @@ func (h jobImplicitIdentitiesHook) handleConsulTasks(t *structs.Task, consul *st
...
+	vaultWID = h.srv.config.VaultDefaultIdentity()
 	if vaultWID == nil {
 		// If no identity is found skip inject the implicit identity and
 		// fallback to the legacy flow.
 		return
 	}
 
 	// Set the expected identity name and inject it into the task.
-	vaultWID.Name = vaultIdentityName
+	vaultWID.Name = t.Vault.IdentityName()
```
tgross, line 134: A minor thing, but when we call this we have to allocate a new buffer for the string each time, and we call it twice in the same function. Might be nice if we called `vaultIdentiyName := t.Vault.IdentityName()` somewhere near the top of the function.

tgross (changes requested):
LGTM except for the bit about canonicalization and how that interacts with namespace governance.

nomad/structs/structs.go
```diff
@@ -10013,6 +10019,10 @@ func (v *Vault) Copy() *Vault {
 }
 
 func (v *Vault) Canonicalize() {
+	if v.Cluster == "" {
+		v.Cluster = VaultDefaultCluster
+	}
```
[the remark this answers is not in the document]
lgfa29: Ops, nice catch! Fixed in https://github.com/hashicorp/nomad/pull/18591/commits/5bf7efa15650e2c84e7b9f410f6035f8b384b756, it turns out I only needed for tests 😅 

I also added a note to the method to make sure future Luiz doesn't mess this up again 😬

nomad/job_endpoint_hook_implicit_identities.go
lgfa29: Fixed in https://github.com/hashicorp/nomad/pull/18591/commits/d8c142c34a7fbecd09b64151ab4a6f80b5971bd3. This method is still not used much, but I will keep this in mind in future PRs.

nomad/structs/structs.go
```diff
@@ -10013,6 +10019,9 @@ func (v *Vault) Copy() *Vault {
 }
 
 func (v *Vault) Canonicalize() {
+	// The Vault cluster name is canonicalized in the jobVaultHook during job
+	// registration because the value may be read from the server config.
```
tgross, line 10023: Good call adding the comment so we don't come back and break it later :grinning:

tgross (approved):
LGTM!

PR #18262, opened 2023-08-18
identity: support jwt expiration and rotation

schmichael:
What is in this PR:

- `WorkloadIdentity.TTL` field for allowing users to specify the lifetime of a specific jwt
- Clientside token renewal code
- Validation and testing

What's *not* in this PR:

- Still no docs or changelog.
- Still no e2e tests.
- `WorkloadIdentity.ChangeMode` - it shouldn't be necessary for the Consul/Vault refresh as they will use internal plumbing. Still planning on adding ASAP.
- Changes to root key rotation <- this is the last hard requirement as we need to prepublish keys before using them and ensure we keep old keys as long as unexpired tokens they have signed are still in use.

I felt like this was a nice sized PR and will work on root key rotation next.

PR #18592, opened 2023-09-27
vault: remove `use_identity` agent config

lgfa29:
The initial intention behind the `vault.use_identity` configuration was to indicate to Nomad servers that they would need to sign a workload identities for allocs with a `vault` block.

But in order to support identity renewal, #18262 and #18431 moved the token signing logic to the alloc runner since a new token needs to be signed prior to the TTL expiring.

So #18343 implemented `use_identity` as a flag to indicate that the workload identity JWT flow should be used when deriving Vault tokens for tasks.

But this configuration value is set on servers so it is not available to clients at the time of token derivation, making its meaning not clear: a job may end up using the identity-based flow even when `use_identity` is `false`.

The only reliable signal available to clients at token derivation time is the presence of an `identity` block for Vault, and this is already configured with the `vault.default_identity` configuration block, making `vault.use_identity` redundant.

This commit removes the `vault.use_identity` configuration and simplifies the logic on when an implicit Vault identity is injected into tasks.

```mermaid
flowchart TD
	Start[Job Registered] --> HasVault{Task has vault block?}

    HasVault -->|No| HasVaultIdentityNoVault{Task has vault identity?}
    HasVault -->|Yes| HasVaultIdentityWithVault{Task has vault identity?}

    HasVaultIdentityNoVault -->|Yes| VaultNoIdentityWarn[Warn\nVault identity without vault block]
    HasVaultIdentityNoVault --> |No| NoVault[No Vault]
    VaultNoIdentityWarn --> NoVault

    HasVaultIdentityWithVault -->|Yes| WIDHasPolicies{Vault block has policies?}
    HasVaultIdentityWithVault -->|No| HasDefaultIdentity{Server has default_identity?}

    WIDHasPolicies -->|No| UseWID[Use workload identity]
    WIDHasPolicies -->|Yes| DeprecatePoliciesWarn[Warn\nUsing identity with policies]
    DeprecatePoliciesWarn --> UseWID
    
    HasDefaultIdentity -->|Yes| InjectDefaultIdentity[Inject default identity]
    InjectDefaultIdentity --> WIDHasPolicies
    HasDefaultIdentity -->|No| LegacyHasPolicies{Vault block has policies?}

    LegacyHasPolicies -->|Yes| UseLegacy[Use legacy flow]
    LegacyHasPolicies -->|No| MissingPoliciesError[Error\nLegacy flow without policies]
```

### DIFF
--- a/command/agent/config_parse_test.go
+++ b/command/agent/config_parse_test.go
@@ -306,7 +306,6 @@ var basicConfig = &Config{
 		TLSSkipVerify:        &trueValue,
 		TaskTokenTTL:         "1s",
 		Token:                "12345",
-		UseIdentity:          pointer.Of(true),
 		DefaultIdentity: &config.WorkloadIdentityConfig{
 			Audience: []string{"vault.io", "nomad.io"},
 			Env:      pointer.Of(false),
@@ -331,7 +330,6 @@ var basicConfig = &Config{
 			TLSSkipVerify:        &trueValue,
 			TaskTokenTTL:         "1s",
 			Token:                "12345",
-			UseIdentity:          pointer.Of(true),
 			DefaultIdentity: &config.WorkloadIdentityConfig{
 				Audience: []string{"vault.io", "nomad.io"},
 				Env:      pointer.Of(false),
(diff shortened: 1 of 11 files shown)


tgross (approved):
LGTM!

PR #18594, opened 2023-09-27
sids_hook: read tokens from consul_hook when available

tgross:
The `sids_hook` runs for Connect sidecar/gateway tasks and gets Consul Service Identity (SI) tokens for use by the Envoy bootstrap hook. When Workload Identity is being used with Consul, the `consul_hook` will have already added these tokens to the alloc hook resources. Update the `sids_hook` to use those tokens instead and write them to the expected area of the taskdir.

### DIFF
--- a/client/allocrunner/taskrunner/sids_hook.go
+++ b/client/allocrunner/taskrunner/sids_hook.go
@@ -51,6 +51,7 @@ type sidsHookConfig struct {
 	sidsClient consul.ServiceIdentityAPI
 	lifecycle  ti.TaskLifecycle
 	logger     hclog.Logger
+	runner     *TaskRunner
 }
 
 // Service Identities hook for managing SI tokens of connect enabled tasks.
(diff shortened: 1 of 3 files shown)


client/allocrunner/taskrunner/sids_hook.go
```diff
@@ -51,6 +51,7 @@ type sidsHookConfig struct {
 	sidsClient consul.ServiceIdentityAPI
 	lifecycle  ti.TaskLifecycle
 	logger     hclog.Logger
+	runner     *TaskRunner
```
pkazmierczak, line 54: I think we could just use `hookResources *cstructs.AllocHookResources` like the `consul_hook` does? There's no need to have the full `TaskRunner` object I think?

pkazmierczak (approved):
LGTM!

client/allocrunner/taskrunner/sids_hook.go
tgross: Good call, will fix.

Issue #18507, opened 2023-09-14
Reduce reported precision on Nomad UI page for task


EugenKon:
### Proposal
![image](https://github.com/hashicorp/nomad/assets/139780495/81216080-eacf-4afc-b92f-8f4717e36c62)

Please print 2m 12s. This is hard to read to long number and it is not relevant, because we do not know exact time when event occur.

philrenaud:
Thanks, @EugenKon — makes perfect sense, will look to fix this soon.

PR #18595, opened 2023-09-27
[ui] Simplify times in task events

philrenaud:
Resolves https://github.com/hashicorp/nomad/issues/18507

Task restarts (and possibly other events) use more time specificity than is needed (for example, "task restarting in 10.01934802s").

This simplifies and rounds times to simple second, minute, or hour strings where relevant.

### DIFF
--- a/.changelog/18595.txt
+++ b/.changelog/18595.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: simplify presentation of task event times (10m2.230948s bceomes 10m2s etc.)
+```
--- a/ui/app/models/task-event.js
+++ b/ui/app/models/task-event.js
@@ -25,16 +25,18 @@ export default class TaskEvent extends Fragment {
 }
 
 function simplifyTimeMessage(message) {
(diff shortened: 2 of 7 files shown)


philrenaud:
Q for @DingoEatingFuzz : Is the most appropriate place for me to handle this in the TaskEvent model, or in the TaskEvent Serializer?

There's already a 
```
  attrs = {
    message: 'DisplayMessage',
  };
```

aliasing happening in the serializer which I think can probably be removed with this change, but wanted to double check.

tgross (approved):
LGTM

DingoEatingFuzz:
> is the most appropriate place for me to handle this in the TaskEvent model, or in the TaskEvent Serializer?

Personally I see this as presentation logic vs. data logic, so I like it in the model. I could also see it as a helper (though that's also a little odd since it operates on a whole string that happens to have a duration in it, not just a duration).

DingoEatingFuzz (approved):
Great time to use unit tests, you love to see it!

Issue #18266, opened 2023-08-20
Blocking queries may block longer than necessary in particular race conditions.


stswidwinski:
### Nomad version
Tip.

### Operating system and Environment details
Unix

### Issue

The blocking query logic is written such that awaiting on a trigger may race with the trigger itself, resulting in a block which is much longer than necessary. To be exact:

https://github.com/hashicorp/nomad/blob/main/nomad/rpc.go#L832-L870

Let us consider the following lines:

1. Line 832: background context (with timeout)
2. Line 865: waiting on background context (or timout)
3. Everything inbetween is "the query"

Consider the following set of events:

1. Context is created (Line 832)
2. In the middle of "the query" a relevant event triggers background context
3. The query finishes and we hit Line 865

The implementation of `ws.WatchCtx` is such that notifications from point 2 will be ignored and will not trigger a retry. This may cause long (sometimes all the way to timeout) queries where an instant query should've been used to satisfy the query.

One should likely begin observing the context prior to the beginning of "the query" to correctly handle retries.

### Reproduction steps

A simple repro is to explicitly race against state changes within "the query." Here is a simple way to achieve this with relatively high probability.

**Step 1: Start the dev server**

``` ./nomad agent -dev ```

**Step 2: Start running a tight loop of queries for a job allocations**

Here is a sample script:
```
#!/usr/bin/bash

set -euo pipefail
set -x

NOMAD="./nomad"
TMP_DIR="/tmp/test"
mkdir -p $TMP_DIR

INDEX=1
while true; do
  # Parallel queries in hope of racing just right
  for i in {1..10}; do
    $NOMAD operator api -verbose "/v1/job/test_job-1/allocations?index=$INDEX&all=true" 2>$TMP_DIR/stderr-$i &
    pids[${i}]=$!
  done

  # wait for all pids
  for pid in ${pids[*]}; do
      wait $pid
  done

  # Next index.
  INDEX=$(cat "$TMP_DIR/stderr-10" |  grep X-Nomad-Index | sed 'sx.*: xx' |  sed 's/[^0-9].*//' || echo "1")
done
```
This will just forever return information and use the latest index known. This is expected to continue to return information ~in a tight loop and pretty much never block for a very long time.

**Step 3: Start, Stop and GC a job**
```
#!/usr/bin/bash

set -euo pipefail
set -x

NOMAD="./nomad"
TMP_DIR="/tmp/test"
mkdir -p $TMP_DIR

function create_job_spec () {
[...]

PR #18601, opened 2023-09-27
client: prevent using stale allocs

schmichael:
Similar to #18269, it is possible that even if Node.GetClientAllocs retrieves fresh allocs that the subsequent Alloc.GetAllocs call retrieves stale allocs. While [`diffAlloc(existing, updated)`](https://github.com/hashicorp/nomad/blob/v1.6.2/client/util.go#L27-L46) properly ignores stale alloc *updates*, alloc deletions have no such check.

So if a client retrieves an alloc created at index 123, and then a subsequent Alloc.GetAllocs call hits a new server which returns results at index 100, the client will stop the alloc created at 123 because it will be missing from the stale response.

This change applies the same logic as #18269 and ensures only fresh responses are used.

Glossary:
* fresh - modified at an index > the query index
* stale - modified at an index <= the query index

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -2419,6 +2419,24 @@ OUTER:
 				}
 			}
 
+			// It is possible that Alloc.GetAllocs hits a different server than
+			// Node.GetClientAllocs which returns older results.
+			if allocsResp.Index <= allocsReq.MinQueryIndex {
+				retry := c.retryIntv(getAllocRetryIntv)
+				c.logger.Warn("failed to retrieve updated allocs; retrying",
+					"req_index", allocsReq.MinQueryIndex,
+					"resp_index", allocsResp.Index,
+					"num_allocs", len(pull),
+					"wait", retry,
+				)
+				select {
+				case <-time.After(retry):
+					continue
+				case <-c.shutdownCh:
+					return
+				}
+			}
+
 			// Ensure that we received all the allocations we wanted
 			pulledAllocs = make(map[string]*structs.Allocation, len(allocsResp.Allocs))
 			for _, alloc := range allocsResp.Allocs {
--- a/client/client.go
+++ b/client/client.go
@@ -2359,7 +2359,7 @@ OUTER:
 		//
 		// For full context, please see https://github.com/hashicorp/nomad/issues/18267
 		if resp.Index <= req.MinQueryIndex {
-			c.logger.Debug("Received stale allocation information. Retrying.",
+			c.logger.Debug("received stale allocation information; retrying",
 				"index", resp.Index, "min_index", req.MinQueryIndex)
 			continue OUTER
 		}
--- a/nomad/alloc_endpoint.go
+++ b/nomad/alloc_endpoint.go
@@ -257,7 +257,7 @@ func (a *Alloc) GetAllocs(args *structs.AllocsGetRequest,
 				reply.Allocs = allocs
 				reply.Index = maxIndex
 			} else {
-				// Use the last index that affected the nodes table
+				// Use the last index that affected the allocs table
 				index, err := state.Index("allocs")
 				if err != nil {
 					return err


schmichael:
@stswidwinski I love you weighing in on this as well if you have time since this is directly downstream from your fix in #18269

stswidwinski:
I think this is entirely analogous to #18269 and I agree that this should address the problem. Previously we haven't addressed this because for two reasons:

1. we have implied that the likelihood of this occuring after the prior responses was correct (here: https://github.com/hashicorp/nomad/pull/18601/files#diff-bd3a55a72186f59e2e63efb4951573b2f9e4a7cc98086e922b0859f8ccc1dd09R2388) was tiny
2. we were focusing on the somewhat larger blast radius race

I believe that https://github.com/hashicorp/nomad/issues/18267 more broadly describes this as a problem with nearly every RPC issued such that we require a minimal index. That's because a stale response is not an error which is somewhat compounded by related https://github.com/hashicorp/nomad/issues/18266.

While squashing the most painful samples seems like the right thing to do in the short term, perhaps it makes sense to mint a few wrappers which help issue queries with an index limit which convert stale responses into errors (such that the type checking helps us handle them). What do you think?

schmichael:
> While squashing the most painful samples seems like the right thing to do in the short term, perhaps it makes sense to mint a few wrappers which help issue queries with an index limit which convert stale responses into errors (such that the type checking helps us handle them). What do you think?

Absolutely. I'm auditing all RPCs Clients make as we speak since as you pointed out this is likely part of a broader pattern.

In the past we had audited these calls for properly _setting_ the index, but during that audit we had neglected to account for _checking_ the index after the response.

Once I get the audit complete and see what the damage is, we can decide whether spot fixes or more holistic changes are best.

Given the opportunity to design blocking RPC semantics from scratch I would have absolutely choose a default behavior of returning an error instead of a stale response. Giving callers valid looking data and entrusting them to double check the index is obviously error prone.

client/client.go
```diff
@@ -2419,6 +2419,24 @@ OUTER:
 				}
 			}
 
+			// It is possible that Alloc.GetAllocs hits a different server than
+			// Node.GetClientAllocs which returns older results.
+			if allocsResp.Index <= allocsReq.MinQueryIndex {
```
tgross, line 2424: Shouldn't this be `<`? Unlike the case above with `Node.GetClientAllocs`, we want to make sure the index is at least the _same_ index as the previous `Node.GetClientAllocs` RPC and we've set the `MinQueryIndex` to `pullIndex - 1` to ensure that. That is, the following is expected behavior if nothing has changed between the two calls:

* `Node.GetClientAllocs` returns index 10.
* we set `MinQueryIndex` to index 9, because we don't want to wait for index 11.
* `Alloc.GetAllocs` returns index 10.

This all springs from the unfortunate name of `MinQueryIndex`, which really is "this is the last index I saw":

```go
 	// If set, wait until query exceeds given index. Must be provided
	// with MaxQueryTime.
	MinQueryIndex uint64
```

stswidwinski: Nice catch. I think I was fooled by the fact that we're inconsistent (line `2480` in turn does not add the `-1` to the index which makes the former check `<=` instead of `<`). Should we pick one, stick to it and be consistent throughout?

tgross: Ideally it'd have been one query that gets us everything we need, but this reduces the total amount of data we have to send/receive by quite a bit for clients with a large amount of allocations.

stswidwinski: Yup, you're right. Sorry about that, in retrospect It was a silly idea on my part. Thanks for the explanation!

schmichael: Great catch. I'll update and make an attempt at a comment.

schmichael: Actually due to the -1 shouldn't this be a `<=`? In Tim's example:

1. Node.GetClientAllocs -> 10
2. 9 -> Alloc.GetAllocs
3. Alloc.GetAllocs -> 10

`<=` is fine here because 10 > 9 and equal to the index you got before. This is the happy (and hopefully common) path.

But let's say you get unlucky and hit time travel:

1. Node.GetClientAllocs -> 10
2. 9 -> Alloc.GetAllocs
3. Alloc.GetAllocs -> 9

Only retrying on `<` here would allow the client to use allocs at index 9 which might differ from the state the previous RPC saw at index 10!

Therefore I think due to the -1 we can maintain the `<=` comparison because it means "at least at the index the previous rpc saw."

tgross: Ah, you're right, we're comparing the response to the `MinQueryIndex`, not the `pullIndex`.

tgross (approved):
LGTM :+1:

schmichael:
Skipping changelog entry because #18269 covers it (and sadly I don't think I can link 2 PRs from 1 logline in our changelog building system?)

schmichael:
Backport failed: https://github.com/hashicorp/nomad/actions/runs/6342982782

Manual 1.4.x backport: https://github.com/hashicorp/nomad/commit/d2cd6dbb722222b28136decb790d604dfb8cf32b

Manual 1.5.x backport: https://github.com/hashicorp/nomad/commit/b1cf888dcbcbfad67fa42f2161b0b41b659126c5

Manual 1.6.x backport: https://github.com/hashicorp/nomad/commit/f02061fba053f0e5d8ee3194ddabfd8b11b7e1b7

PR #18605, opened 2023-09-28
[ui] Access Control CRUD: Make name fields for Policies and Roles required

philrenaud:
Using the new [Helios form input required attr](https://helios.hashicorp.design/components/form/text-input?tab=code#required-vs-optional)

This component is a little more involved than our older input component, so we have to explicitly set an updateRoleName / updatePolicyName method rather than it just implicitly knowing what to do. But the benefit is consistency with the rest of our form styles so a worthwhile tradeoff.

This PR also cleans up a bit of the old policy-specific styles and moves them into the larger access-control style section.

image


juliezzhou (approved):
LGTM

PR #18606, opened 2023-09-28
vault: add `jwt_backend_path` agent config

lgfa29:
Add agent configuration to allow cluster operators to define the path where the JWT auth method backend is mounted.


nomad/structs/config/vault.go
```diff
@@ -61,6 +61,9 @@ type VaultConfig struct {
 	// URL such as "http://vault.example.com"
 	Addr string `mapstructure:"address"`
 
+	// JWTBackendPath is the path used to access the JWT auth method.
+	JWTBackendPath string `mapstructure:"jwt_backend_path"`
```
lgfa29, line 65: I'm open to suggestions on how to call this field. I followed the language from https://developer.hashicorp.com/vault/docs/auth/jwt#via-the-cli

> The default path for the JWT authentication backend is `/jwt`

tgross: Maybe `JWTAuthBackendPath`? I don't feel strongly about it though.

website/content/docs/configuration/vault.mdx
```diff
@@ -97,6 +97,10 @@ agents with [`client.enabled`] set to `true`.
   given in the format `protocol://host:port`. If your Vault installation is
   behind a load balancer, this should be the address of the load balancer.
 
+- `jwt_backend_path` - `(string: "jwt")` - Specifies the mount
+  [path][vault_auth_enable_path] of the JWT authentication method to be used to
+  login with workload identity JWTs.
```
tgross, line 102: :+1: for linking to the Vault docs here.

command/agent/config_parse_test.go
```diff
@@ -318,6 +319,7 @@ var basicConfig = &Config{
 		structs.VaultDefaultCluster: {
 			Name:                 structs.VaultDefaultCluster,
 			Addr:                 "127.0.0.1:9500",
+			JWTBackendPath:       "nomad_jwt",
```
tgross, line 322: There's a `TestConfig_MultipleVault` test case below where it'd be nice to add an assertion for this vs the default (I think both will show up with the config changes you've made here).

tgross (approved):
LGTM!

PR #18607, opened 2023-09-28
[ui] node eligibilty taken into consideration when clients list filtered to "ready"

### LINKED ISSUE #18593: UI: filtering clients on "ready" displays ineligible hosts
### Nomad version
1.6.2

Fedora29

### Issue
/ui/clients?state=%5B"ready"%5D shows also ineligible hosts.

### Reproduction steps
Have some hosts ready and ineligible, and filter in UI for ready.

#### Expected Result
Only ready hosts should be shown when filtering for ready hosts.

#### Actual Result
image

philrenaud:
On our clients list page, we abstract node readiness, eligibility, and draining status into one `State` filter.

This leads to situations where a user selects "Ready" and expects to see eligible nodes, but a node can technically be "ready" by status and marked ineligible for new tasks.

This seems confusing with the abstraction in place. As such, selecting "Ready" will not filter out any ineligible nodes.

In resolving this, we've made some updates to the way the State filter works, relative to the others:
- It is an "inclusive" filter, with all options selected by default (others are exclusive until something is clicked)
- query parameters from it are individually appended to the URL, where in other dropdowns a stringified array of all checked options is appended
- it includes subsections for Status, Eligibility, and Drain status, where others are single-subject filters.

We realize this creates an inconsistency, but felt like this represents a practical approach to making state filter behaviour here work as expected.

image

Resolves #18593

### DIFF
--- a/.changelog/18607.txt
+++ b/.changelog/18607.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: change the State filter on clients page to split out eligibility and drain status
+```
--- a/ui/app/components/client-node-row.js
+++ b/ui/app/components/client-node-row.js
@@ -59,35 +59,29 @@ export default class ClientNodeRow extends Component.extend(
 
   @computed('node.compositeStatus')
   get nodeStatusColor() {
-    let compositeStatus = this.get('node.compositeStatus');
-
-    if (compositeStatus === 'draining') {
-      return 'neutral';
-    } else if (compositeStatus === 'ineligible') {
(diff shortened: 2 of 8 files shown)


philrenaud:
Note: converting back to draft after realizing how much test logic is tied up in state/eligibility/drain status.
A more holistic approach in the future will look like this:
image

and untie eligibility/drain status as "status" concepts / list them as tags of the node name, rather than under "state" (which should be renamed Status, to match CLI/API)

PR #18609, opened 2023-09-28
server: Add reporting configuration to  nomad server

Juanadelacuesta:
To continue with the licensing reporting, the configuration needs to be passed down to the server who will in turn be responsable for the metrics. This PR adds the a reporting manager that will only be populated and used on the enterprise version.

Related: [#1106](https://github.com/hashicorp/nomad-enterprise/issues/1106)


tgross (approved):
LGTM :+1:

PR #18610, opened 2023-09-28
WI: WIDMgr should expose default identity signatures

pkazmierczak:
Since the `identity_hook` is meant to be the central place that makes signed identities available to other hooks, it should also expose the `default` identity that is signed by the plan applier. 

### DIFF
--- a/client/widmgr/widmgr.go
+++ b/client/widmgr/widmgr.go
@@ -26,10 +26,11 @@ type IdentityManager interface {
 }
 
 type WIDMgr struct {
-	allocID  string
-	minIndex uint64
-	widSpecs map[string][]*structs.WorkloadIdentity // task -> WI
-	signer   IdentitySigner
+	allocID                 string
+	defaultSignedIdentities map[string]string // signed by the plan applier
+	minIndex                uint64
+	widSpecs                map[string][]*structs.WorkloadIdentity // task -> WI
+	signer                  IdentitySigner
 
 	// lastToken are the last retrieved signed workload identifiers keyed by
 	// TaskIdentity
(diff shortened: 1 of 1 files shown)


client/widmgr/widmgr.go
```diff
@@ -187,13 +189,34 @@ func (m *WIDMgr) Shutdown() {
 
 // getIdentities fetches all signed identities or returns an error.
 func (m *WIDMgr) getIdentities() error {
+	m.lastTokenLock.Lock()
+	defer m.lastTokenLock.Unlock()
+
+	// initialize the map
+	m.lastToken = map[cstructs.TaskIdentity]*structs.SignedWorkloadIdentity{}
```
tgross, line 196: Currently we wait to overwrite `m.lastToken` until after we've successfully called `SignIdentities`, so that we can keep the last tokens we saw if there's a transient problem (ex. the RPC connection was dropped briefly or something). So we should keep that behavior here and create this map as a local variable first and then assign it to `m.lastToken` after we've called `SignIdentities` below.

We could also avoid taking the lock until later if we did that.

lgfa29: I think we could also create this map in `NewWIDMgr` since it has all the information it needs?

client/widmgr/widmgr.go
```diff
@@ -221,6 +241,11 @@ func (m *WIDMgr) getIdentities() error {
 
 		m.lastToken[id] = swid
 	}
+	// Store default identity tokens
+	for id, token := range defaultTokens {
+		m.lastToken[id] = token
+	}
```
tgross, line 247: I think we want this to be before the previous block where we're looping over the signed identities, because that way a user can override the default.

pkazmierczak: can users override the default?

pkazmierczak: ah nevermind, I see it in the RFC now 🤦 I'll fix asap

pkazmierczak: [54440e4](https://github.com/hashicorp/nomad/pull/18610/commits/54440e4c7d1dc8e69a56d3ca1b8c731a0924effd)

tgross (approved):
LGTM! :+1:

PR #18611, opened 2023-09-28
consul: use constant instead of "default" literal

tgross:
Use the constant `structs.ConsulDefaultCluster` instead of the string literal "default", as we've done for Vault.


lgfa29 (approved):
Spot checked other places and this PR seems to cover all references 👍

PR #18612, opened 2023-09-28
client: prevent watching stale alloc state

schmichael:
When waiting on a previous alloc we must query against the leader before switching to a stale query with index set.

Also check to ensure the response is fresh before using it like #18269

### DIFF
--- a/client/allocwatcher/alloc_watcher.go
+++ b/client/allocwatcher/alloc_watcher.go
@@ -374,12 +374,14 @@ func (p *remotePrevAlloc) Wait(ctx context.Context) error {
 		resp := structs.SingleAllocResponse{}
 		err := p.rpc.RPC("Alloc.GetAlloc", &req, &resp)
 		if err != nil {
-			p.logger.Error("error querying previous alloc", "error", err)
 			retry := getRemoteRetryIntv + helper.RandomStagger(getRemoteRetryIntv)
+			timer, stop := helper.NewSafeTimer(retry)
+			p.logger.Error("error querying previous alloc", "error", err, "wait", retry)
 			select {
-			case <-time.After(retry):
+			case <-timer.C:
 				continue
 			case <-ctx.Done():
+				stop()
 				return ctx.Err()
 			}
 		}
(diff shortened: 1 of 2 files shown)


client/allocwatcher/alloc_watcher.go
```diff
@@ -381,6 +383,22 @@ func (p *remotePrevAlloc) Wait(ctx context.Context) error {
 				return ctx.Err()
 			}
 		}
+
+		if req.AllowStale && resp.Index <= req.MinQueryIndex {
```
tgross, line 387: I think this can be `<`. Unlike what we did in #18601, we're not using the `resp.Index - 1` as the next `MinQueryIndex`. So if `resp.Index == req.MinQueryIndex`, it just means that nothing has changed on the server and the blocking query has timed out. If we fall through then nothing will have changed on the `resp.Alloc`, and there's no need to write a warning to the logs in that case.

tgross (approved):
LGTM!

However, note the semgrep error on `time.After`. The `NewSafeTimer` helper provides a safer abstraction that provides compiler support in avoiding leaking the goroutine.

client/client.go
```diff
@@ -2423,16 +2423,18 @@ OUTER:
...
+				timer, stop := helper.NewSafeTimer(retry)
 				c.logger.Warn("failed to retrieve updated allocs; retrying",
 					"req_index", allocsReq.MinQueryIndex,
 					"resp_index", allocsResp.Index,
 					"num_allocs", len(pull),
 					"wait", retry,
 				)
 				select {
-				case <-time.After(retry):
+				case <-timer.C:
```
schmichael, line 2434: Sorry this is an unrelated fix from a linting error I introduced in my previous PR: #18601

tgross (approved):
LGTM

schmichael:
Backports failed: https://github.com/hashicorp/nomad/actions/runs/6356400700

Manual 1.4.x backport: https://github.com/hashicorp/nomad/commit/e531bf12c78676f47c4cc9c0022343b94fe8ffa1

Manual 1.5.x backport: https://github.com/hashicorp/nomad/commit/376a350616eac6a11cb130c3e6fa7aea3b10cd24

Manual 1.6.x backport: https://github.com/hashicorp/nomad/commit/b6f62d5bd2c93bbbfc77c1b82ffee729f132764a

PR #18621, opened 2023-09-29
[ui] bugfix: when hitting the "start" button on a job page, if it has no submission data, fallback to raw json definition

### LINKED ISSUE #18536: Could not start job with variables after stopping (UI)
### Nomad version
Nomad v1.6.2
BuildDate 2023-09-13T16:47:25Z
Revision 73e372ad94033db2ceaf53468b270a31544c23fd

debian 12 amd64

### Issue
After stopping job with variables, it could not be started again in UI. Nomad 1.6.1 was OK.

### Reproduction steps
Create job with variable (used in group/task definition):
```
variable "instances" {
  type = list(object({
    name  = string
    count = number
    tasks = list(object({
      name         = string
      memory       = number
      cpu          = number
      entrypoint   = list(string)
    }))
  }))
}
``` 
There are correct `Job Spec`,`HCL Variable Values` in `Definition` tab.
Stop the job in UI. Make no changes. Start the job in UI. Get error:
```
Could Not Start Job
Failed to parse job: <nil>: Unset variable "instances"; A used variable must be set or have a default value; see https://www.nomadproject.io/docs/job-specification/hcl2/variables for details.
```
This bash script starts the job successfully:
```
nomad job run -json -detach <(nomad job inspect -namespace "$n" "$j" | sed 's/"Stop": true,/"Stop": false,/')
```
Now we have only json data in `Definition`. It's ok. If the job is stopped again, on start (UI) we have another error:
```
Could Not Start Job
Job ID hasn't been provided
```
The same bash script still starts the job.
#### Expected Result
Job should start.
#### Actual Result
Error.

### LINKED ISSUE #18547: Job cannot be restarted from UI in 1.6.2
### Nomad version
1.6.2

Ubuntu 22.04

### Issue

Fail to restart a job in UI.

### Reproduction steps
Saying job version is 4,
* Stop the job, it will create a new version -  version  5 - of the job with Stop flag set to true.
* Try to restart: it will attempt to start the previously create job version, ie version 5 - with stop flag set to true.
The job doesn't restart

When attempting to restart the job an API call is made : https://nomad/v1/job/my-job/submission?version=5 which return a 404.

#### Expected Result

The job should start using version 4 of jobs, maybe there is no need to create a new version of the job when changing the stop flag

#### Actual Result

A new version version of the job is created with stop flag set to true preventing job from start

#### Workaround

Go to Version Tab and revert the job to version 4 where Stop flag is set to false. The workaround is not always possible this the new version is not always present in the UI.

philrenaud:
In Nomad 1.6, we introduced HCL-in-the-UI using a new /submissions endpoint among other things.

Shortly after the release, we realized we missed a spot: stop/starting via the UI would not maintain the submission data. My initial fix for this solved the issue of maintaining HCL, but caused a new bug: jobs that lacked submission data (perhaps they were submitted via curl with just the json spec, and without HCL) would 404 on /submission and fail to progress through to /parse and submit.

This PR provides a fallback to the initial method in cases where the /submission request fails.

Resolves #18547 
Resolves #18536

### DIFF
--- a/.changelog/18621.txt
+++ b/.changelog/18621.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: using start/stop from the job page in the UI will no longer fail when the job lacks HCL submission data
+```
--- a/ui/app/components/job-page/parts/title.js
+++ b/ui/app/components/job-page/parts/title.js
@@ -71,8 +71,17 @@ export default class Title extends Component {
    */
   @task(function* (withNotifications = false) {
     const job = this.job;
-    const specification = yield job.fetchRawSpecification();
-    job.set('_newDefinition', specification.Source);
+
+    // Try to get the submission/hcl sourced specification first.
+    // In the event that this fails, fall back to the raw definition.
+    try {
+      const specification = yield job.fetchRawSpecification();
+      job.set('_newDefinition', specification.Source);
+    } catch {
+      const definition = yield job.fetchRawDefinition();
+      delete definition.Stop;
+      job.set('_newDefinition', JSON.stringify(definition));
+    }
 
     try {
       yield job.parse();


tgross (approved):
LGTM! :+1: 

It occurs to me this whole feature is a bit racy in that another user can submit a new job definition between the time we get the job spec and re-run it. As a future improvement we might want to consider using [`EnforceIndex`](https://developer.hashicorp.com/nomad/api-docs/jobs#enforceindex) so that someone pressing the Start button can't accidentally roll-back the job. But that shouldn't be part of this bug fix!

PR #18623, opened 2023-09-29
config: apply defaults to extra Consul and Vault

lgfa29:
Apply the expected default values when loading additional Consul and Vault cluster configuration. Without these defaults some fields would be left empty.

### DIFF
--- a/command/agent/config.go
+++ b/command/agent/config.go
@@ -1612,7 +1612,8 @@ func mergeVaultConfigs(left, right map[string]*config.VaultConfig) map[string]*c
 		if lConfig, ok := left[name]; ok {
 			merged[name] = lConfig.Merge(rConfig)
 		} else {
-			merged[name] = rConfig.Copy()
+			merged[name] = config.DefaultVaultConfig().Merge(rConfig)
+			merged[name].Name = name
 		}
 	}
 	return merged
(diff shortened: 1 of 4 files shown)


command/agent/config.go
```diff
@@ -1612,7 +1612,8 @@ func mergeVaultConfigs(left, right map[string]*config.VaultConfig) map[string]*c
 		if lConfig, ok := left[name]; ok {
 			merged[name] = lConfig.Merge(rConfig)
 		} else {
-			merged[name] = rConfig.Copy()
+			merged[name] = config.DefaultVaultConfig().Merge(rConfig).Copy()
```
lgfa29, line 1615: Or maybe `config.DefaultVaultConfig().Merge(rConfig.Copy())`? Does it actually make a difference? 😅

tgross: I don't think we need the `Copy()` at all if we're calling `Merge`, because the field values are being copied from `rConfig` onto of the default config.

tgross (approved):
LGTM with the recommended change

lgfa29:
@tgross while fixing the test failure of `TestConfig_Merge` I think I found another bug where pointer references were being lost.

I fixed that in https://github.com/hashicorp/nomad/pull/18623/commits/9ed793391fd0b4c39480bd12401da3343027b509.

PR #18624, opened 2023-09-29
consul: add multi-cluster support to client constructors

tgross:
When agents start, they create a shared Consul client that is then wrapped as various interfaces for testability, and used in constructing the Nomad client and server. The interfaces that support workload services (rather than the Nomad agent itself) need to support multiple Consul clusters for Nomad Enterprise. Update these interfaces to be factory functions that return the Consul client for a given cluster name. Update the `ServiceClient` to split workload updates between clusters by creating a wrapper around all the clients that delegates to the cluster-specific `ServiceClient`.

### DIFF
--- a/command/agent/consul/service_client.go
+++ b/command/agent/consul/service_client.go
@@ -5,6 +5,7 @@ package consul
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"maps"
 	"net"
@@ -136,6 +137,9 @@ type ConfigAPI interface {
 	Set(entry api.ConfigEntry, w *api.WriteOptions) (bool, *api.WriteMeta, error)
 	// Delete(kind, name string, w *api.WriteOptions) (*api.WriteMeta, error) (not used)
 }
+
+// ConfigAPIFunc returns a ConfigAPI interface for the specific cluster
+type ConfigAPIFunc func(clusterName string) ConfigAPI
 
 // ACLsAPI is the consul/api.ACL API subset used by Nomad Server.
 //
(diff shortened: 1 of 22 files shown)


nomad/structs/connect.go
```diff
@@ -6,6 +6,7 @@ package structs
 // ConsulConfigEntries represents Consul ConfigEntry definitions from a job for
 // a single Consul namespace.
 type ConsulConfigEntries struct {
+	Cluster     string
 	Ingress     map[string]*ConsulIngressConfigEntry
 	Terminating map[string]*ConsulTerminatingConfigEntry
 }
```
lgfa29, line 12: I may be missing something about Consul's config entry pattern, but this struct seems a bit odd to me now 🤔  

The godoc mentions that a `ConsulConfigEntries` is for a a single Consul namespace, and I assume a Consul namespace is specific to a cluster?

If that's the case, then I think I would expect `job.ConfigEntries` below to return a `map[string]map[string]*ConsulConfigEntries` with the `cluster -> namespace -> CE` mapping? But that's also a bit odd to me, so adding both `Namespace` and `Cluster` would make things more clear?

command/agent/consul/service_client.go
```diff
@@ -423,6 +427,212 @@ func (o *operations) String() string {
...
+	}
+
+	return nil
+}
+
+func (scw *ServiceClientWrapper) RegisterAgent(role string, services []*structs.Service) error {
+	scw.lock.RLock()
+	defer scw.lock.RUnlock()
+
+	serviceClient, ok := scw.serviceClients[structs.ConsulDefaultCluster]
```
lgfa29, line 476: Just checking my understanding, this is always `default` because that's where we want to register Nomad agents, not in any of the additional clusters?

lgfa29 (approved):
Mostly me being confused about Consul 😅

command/agent/consul/service_client.go
tgross: Right! We could conceivably change that and make it configurable in the future, but for purposes of shipping this support it seemed much easier to make Nomad always register itself with the default cluster (and less prone to accidentally split-braining the Nomad cluster across Consul clusters!).

nomad/structs/connect.go
tgross: A Consul namespace is specific to a Consul cluster, but not necessarily unique across clusters. But we don't let you set a Consul namespace on the `service`, only on the group.

PR #18628, opened 2023-10-01
chore(command): improve `job` and `status` text

thiskevinwang:
Improves help text for `nomad job status` and `nomad status` CLI commands

### DIFF
--- a/command/status.go
+++ b/command/status.go
@@ -27,8 +27,8 @@ Usage: nomad status [options] <identifier>
   detect the type of resource being queried and display the appropriate
   status output.
 
-	If no arguments are provided, the command will fallback to "nomad job status",
-	which will list all jobs.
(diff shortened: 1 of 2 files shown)


command/status.go
```diff
@@ -27,6 +27,9 @@ Usage: nomad status [options] <identifier>
   detect the type of resource being queried and display the appropriate
   status output.
 
+	If no arguments are provided, the command will fallback to "nomad job status",
+	which will list all jobs.
+
```
tgross, line 32: It looks like you've got a tab (`\t`) here rather than space-aligned with the text above. An unfortunate quirk of writing big blocks of text in an editor configured to write Go code! :grinning:

tgross (changes requested):

tgross (approved):
LGTM! Thanks @thiskevinwang!

PR #18634, opened 2023-10-02
consul: make service and task identity names unique

pkazmierczak:
`consul_hook` depends on Consul service and task identity names being unique. 

### DIFF
--- a/nomad/job_endpoint_hook_implicit_identities.go
+++ b/nomad/job_endpoint_hook_implicit_identities.go
@@ -35,7 +35,7 @@ func (h jobImplicitIdentitiesHook) Mutate(job *structs.Job) (*structs.Job, []err
 				h.handleConsulService(s)
 			}
 			if len(t.Templates) > 0 {
-				h.handleConsulTasks(t, tg.Consul)
+				h.handleConsulTasks(t, tg.Name)
 			}
 			h.handleVault(t)
 		}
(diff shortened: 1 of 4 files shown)


nomad/job_endpoint_hook_implicit_identities.go
```diff
@@ -74,18 +74,20 @@ func (h jobImplicitIdentitiesHook) handleConsulService(s *structs.Service) {
 	}
 
 	// Set the expected identity name and service name.
-	serviceWID.Name = fmt.Sprintf("%s/%s", consulServiceIdentityNamePrefix, s.Name)
+	name := fmt.Sprintf("_nomad-service-%v-%v-%v", s.TaskName, s.Name, s.PortLabel)
```
tgross, line 77: Can we turn this into a function that can be imported on the client as well (so probably in `nomad/structs`), so that the downstream hooks can use the same function and we guarantee they're all trying to get the same service WID when we're done? 

Also, given that we're prefixing this with `consul-service` does it need to also start with `_nomad-service`? The long prefix `consul-service/_nomad-service-taskName-serviceName-port` seems redundant.

tgross (approved):
LGTM once comments are resolved.

nomad/job_endpoint_hook_implicit_identities.go
pkazmierczak: Have a look at ec96df1ce21f4e0b0b65b8c060affdc3606a21fa, I think it's what you suggested.

tgross: Perfect :+1:

PR #18635, opened 2023-10-02
consul: consul taskrunner hook should only write tokens that belong to its task

pkazmierczak:
vide https://github.com/hashicorp/nomad/pull/18557#discussion_r1341513988

Ref https://github.com/hashicorp/team-nomad/issues/404


tgross (approved):
LGTM :+1:

PR #18637, opened 2023-10-02
scaling: set Index on nil-job scale status reply

gulducat:
I noticed this while working on hashicorp/nomad-autoscaler#604 -- When a job went away, the blocking query on `/v1/job/{job-id}/scale` kept blocking for the full 5 minute timeout, instead of returning "not found" as soon as the job was purged/gc'd.

returning a nil error in a `blockingOptions.run()`
without increasing the reply `Index` can cause the
query to block indefinitely (until timeout).

this fixes that happening in `Job.ScaleStatus`
when the job is deleted -- the job going away
should now return as not-found and provide a new
index for the caller to try if they so please.

### DIFF
--- a/.changelog/18637.txt
+++ b/.changelog/18637.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+scaling: Unblock blocking queries to /v1/job/{job-id}/scale if the job goes away
+```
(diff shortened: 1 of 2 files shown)


jrasell (approved):
LGTM.

Do we want to add a changelog entry, to call out the potential change in behaviour for callers of this function?

gulducat:
Ah yes for the changelog - I made one yesterday, just didn't actually push it until this morning.

PR #18638, opened 2023-10-02
deps: upgrade to go-set/v2 and fixup a few broken things

shoenig:
https://github.com/hashicorp/go-set/issues/73

Using Nomad as a large project test case of the `v2.0.0-alpha.2` to make sure everything is fine before actually tagging the `v2.0.0` final release.
No functional changes, just cleaning up deprecated usages that are removed in v2, and replace one call of `.Slice` with `.ForEach` to avoid making the intermediate copy. 

Much smaller PR than it looks - it's almost all import name fixes.

### DIFF
--- a/command/var_put.go
+++ b/command/var_put.go
@@ -15,7 +15,7 @@ import (
 	"strings"
 
 	multierror "github.com/hashicorp/go-multierror"
-	"github.com/hashicorp/go-set"
+	"github.com/hashicorp/go-set/v2"
 	"github.com/hashicorp/hcl"
 	"github.com/hashicorp/hcl/hcl/ast"
 	"github.com/hashicorp/nomad/api"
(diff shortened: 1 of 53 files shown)


command/var_put.go
```diff
@@ -581,10 +581,12 @@ func formatInvalidVarKeyChars(invalid []string) string {
 
 	// Sort the characters for output
 	charList := make([]string, 0, chars.Size())
-	for _, k := range chars.List() {
+	chars.ForEach(func(k string) bool {
```
pkazmierczak, line 584: I'm ok with this but it just looks somewhat non-idiomatic

pkazmierczak (approved):
LGTM!

command/var_put.go
shoenig: Agreed, but it's either this or creating a copy of the data which is worse. The Go team is working behind the scenes to make iteration over user defined structs a thing, but who knows about the ETA

https://github.com/golang/go/discussions/54245
https://github.com/golang/go/discussions/56413

Issue #18627, opened 2023-09-29
☂️  Nomad Actions


philrenaud:
# Nomad Actions
Operator-sanctioned sets of actions in a Nomad job spec that users can execute, allowing for fine-grained control over task execution with ACL policies using the principle of least privilege.

Divided into two main phases:
- `alloc_exec`-enabled actions
- ACL-granular actions

[Draft implementation](https://github.com/hashicorp/nomad/pull/16249)
[Figma](https://www.figma.com/file/psLK28yV5t6NaXexd6Ni3v/Custom-links?type=design&node-id=4926-2088&mode=design&t=ixizaN5WBhu7nUGi-0)

## RFC ([NMD-184](https://go.hashi.co/rfc/nmd-184))
- [x] Draft RFC
- [x] Circulate and discuss
- [x] RFC Approval

## Scaffolding
- [x] [Action struct (go)](https://github.com/hashicorp/nomad/pull/18639)
- [x] Action adapter (js)
- [x] #18726
- [x] #18680
- [x] #18771
- [x] [Ember Data model relationships to jobs, tasks (js)](https://github.com/hashicorp/nomad/pull/18733)
- [x] Sockets service handling (js)
- [x] Mirage (js)

## UI
- [x] #18782
- [x] #19035
- [x] #19018
- [x] Running/Run action notifications
- [x] #19066

## API
- [x] [GET /job/:id/actions](https://github.com/hashicorp/nomad/issues/18680)
- [x] #18735

## Expanded Functionality
- [ ] Run action on one specific, one random, or many allocs
- [ ] Actions with passed arguments / user-provided params
  - Presently out of scope
- [ ] Action.tags
- [ ] Action.ui.style (=critical, =primary, etc.)
- [x] #19039

## Docs
- [x] Add Actions block to jobspec docs
- [x] Add running actions to ui tutorials
- [x] Add `nomad action` to cli docs

## ACL
(Items considered phase 2 - https://github.com/hashicorp/nomad/issues/18800)
- [ ] Blanket ACL rules (namespace.actions.policy)
- [ ] Task-level ACL rules (namespace.job.group.task.action.policy)
- [ ] Arbitrary tag-level ACL rules (action.tags.includes(x)

## Expansion
- [ ] Expose actions [metrics](https://github.com/hashicorp/nomad/issues/18627)
- [ ] Mirage mocks for running actions in the UI
- [ ] #19187

PR #10965, opened 2021-07-29
First draft at better service diffs.

apollo13:
This is a draft for #10964 -- What it does not handle now anymore is a change in port label. Still brainstorming on how to handle this properly :)

PR #18639, opened 2023-10-02
Scaffolding actions

philrenaud:
First steps in setting up Nomad Actions (#18627)
- Establishes the Action struct at task level and allows them to be included in a jobspec
- Returns them under tasks on the GET job endpoint

Also includes a (temporary) workaround to get exec working on our local Ember development environment: direct forwarding to :4646. This is marked with FIXMEs and will be removed prior to merge with main.

### DIFF
--- a/api/tasks.go
+++ b/api/tasks.go
@@ -1172,6 +1172,6 @@ type WorkloadIdentity struct {
 
 type Action struct {
 	Name    string   `hcl:"name,label"`
-	Command string  `mapstructure:"command" hcl:"command"`
+	Command string   `mapstructure:"command" hcl:"command"`
 	Args    []string `mapstructure:"args" hcl:"args,optional"`
 }
--- a/api/tasks.go
+++ b/api/tasks.go
@@ -1172,6 +1172,6 @@ type WorkloadIdentity struct {
 
 type Action struct {
 	Name    string   `hcl:"name,label"`
-	Command *string  `mapstructure:"command" hcl:"command"`
+	Command string  `mapstructure:"command" hcl:"command"`
 	Args    []string `mapstructure:"args" hcl:"args,optional"`
 }
--- a/api/tasks.go
+++ b/api/tasks.go
@@ -1174,5 +1174,4 @@ type Action struct {
 	Name    string   `hcl:"name,label"`
 	Command *string  `mapstructure:"command" hcl:"command"`
 	Args    []string `mapstructure:"args" hcl:"args,optional"`
-	// Type    *string  `mapstructure:"type" hcl:"type,optional"`
 }
--- a/api/tasks.go
+++ b/api/tasks.go
@@ -731,6 +731,8 @@ type Task struct {
 
 	// Workload Identities
 	Identities []*WorkloadIdentity `hcl:"identity,block"`
+
+	Actions []*Action `hcl:"action,block"`
 }
 
 func (t *Task) Canonicalize(tg *TaskGroup, job *Job) {
@@ -1166,4 +1168,11 @@ type WorkloadIdentity struct {
 	File        bool          `hcl:"file,optional"`
 	ServiceName string        `hcl:"service_name,optional"`
 	TTL         time.Duration `mapstructure:"ttl" hcl:"ttl,optional"`
+}
+
+type Action struct {
+	Name    string   `hcl:"name,label"`
+	Command *string  `mapstructure:"command" hcl:"command"`
+	Args    []string `mapstructure:"args" hcl:"args,optional"`
+	// Type    *string  `mapstructure:"type" hcl:"type,optional"`
 }
--- a/command/agent/job_endpoint.go
+++ b/command/agent/job_endpoint.go
@@ -1398,7 +1398,6 @@ func ApiActionToStructsAction(job *structs.Job, action *api.Action, act *structs
 	act.Name = action.Name
 	act.Args = action.Args
 	act.Command = action.Command
-	// act.Type = action.Type
 }
 
 func ApiResourcesToStructs(in *api.Resources) *structs.Resources {
--- a/command/agent/job_endpoint.go
+++ b/command/agent/job_endpoint.go
@@ -1344,13 +1344,9 @@ func ApiTaskToStructsTask(job *structs.Job, group *structs.TaskGroup,
 		}
 	}
 
-	if apiTask.Actions != nil {
-		structsTask.Actions = []*structs.Action{}
-		for _, action := range apiTask.Actions {
-			act := &structs.Action{}
-			ApiActionToStructsAction(job, action, act)
-			structsTask.Actions = append(structsTask.Actions, act)
-		}
+	for _, action := range apiTask.Actions {
+		act := ApiActionToStructsAction(job, action, &structs.Action{})
+		structsTask.Actions = append(structsTask.Actions, act)
 	}
 }
 
@@ -1394,10 +1390,12 @@ func ApiCSIPluginConfigToStructsCSIPluginConfig(apiConfig *api.TaskCSIPluginConf
 	return sc
 }
 
-func ApiActionToStructsAction(job *structs.Job, action *api.Action, act *structs.Action) {
-	act.Name = action.Name
-	act.Args = slices.Clone(action.Args)
-	act.Command = action.Command
+func ApiActionToStructsAction(job *structs.Job, action *api.Action, act *structs.Action) *structs.Action {
+	return &structs.Action{
+		Name:    action.Name,
+		Args:    slices.Clone(action.Args),
+		Command: action.Command,
+	}
 }
 
 func ApiResourcesToStructs(in *api.Resources) *structs.Resources {
--- a/command/agent/job_endpoint.go
+++ b/command/agent/job_endpoint.go
@@ -1343,6 +1343,15 @@ func ApiTaskToStructsTask(job *structs.Job, group *structs.TaskGroup,
 			Sidecar: apiTask.Lifecycle.Sidecar,
 		}
 	}
+
+	if apiTask.Actions != nil {
+		structsTask.Actions = []*structs.Action{}
+		for _, action := range apiTask.Actions {
+			act := &structs.Action{}
+			ApiActionToStructsAction(job, action, act)
+			structsTask.Actions = append(structsTask.Actions, act)
+		}
+	}
 }
 
 // apiWaitConfigToStructsWaitConfig is a copy and type conversion between the API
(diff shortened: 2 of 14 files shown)


nomad/structs/structs.go
```diff
@@ -13282,3 +13285,9 @@ func NewRpcError(err error, code *int64) *RpcError {
 func (r *RpcError) Error() string {
 	return r.Message
 }
+
+type Action struct {
+	Name    string
+	Command *string
+	Args    []string
+}
```
tgross, line 13293: You'll need to implement `Copy` and `Equals`, as well as diff logic in `nomad/structs/diff.go`, in order to update jobs that have actions and to avoid potentially corrupting the state store. There's a [jobspec checklist](https://github.com/hashicorp/nomad/blob/main/contributing/checklist-jobspec.md) which can help you make sure you've got all the required bits.

I know this is a "scaffolding" PR though so whether this is blocking depends on how much scaffolding you'll need for the next chunk of work you'll do.

api/tasks.go
```diff
@@ -1167,3 +1169,9 @@ type WorkloadIdentity struct {
 	ServiceName string        `hcl:"service_name,optional"`
 	TTL         time.Duration `mapstructure:"ttl" hcl:"ttl,optional"`
 }
+
+type Action struct {
+	Name    string   `hcl:"name,label"`
+	Command *string  `mapstructure:"command" hcl:"command"`
```
tgross, line 1175: The only reason we use pointers to primitive fields in the API package is to accommodate optional fields. But `Command` is a required field.

```suggestion
	Command string  `mapstructure:"command" hcl:"command"`
```

command/agent/job_endpoint.go
```diff
@@ -1385,6 +1394,12 @@ func ApiCSIPluginConfigToStructsCSIPluginConfig(apiConfig *api.TaskCSIPluginConf
 	return sc
 }
 
+func ApiActionToStructsAction(job *structs.Job, action *api.Action, act *structs.Action) {
+	act.Name = action.Name
+	act.Args = action.Args
```
tgross, line 1399: `Args` isn't a primitive so you'll need to copy this slice too:

```suggestion
	act.Args = slices.Clone(action.Args)
```

nomad/structs/structs.go
philrenaud: Thanks, putting those together now. Should I be handling this in a new /nomad/structs/actions.go or keep them in /structs.go, in your opinion?

tgross: It'd be nice to have them in `nomad/structs/actions.go`, just because that `structs.go` file is _huge_ :grinning:

nomad/structs/structs.go
```diff
@@ -13282,3 +13286,30 @@ func NewRpcError(err error, code *int64) *RpcError {
...
+func (a *Action) Equal(o *Action) bool {
+	if a == o {
+		return true
+	}
+	if a == nil || o == nil {
+		return false
+	}
+	return a.Name == o.Name &&
+		a.Command == o.Command &&
+		strings.Join(a.Args, ",") == strings.Join(o.Args, ",")
```
tgross, line 13314: Using https://pkg.go.dev/slices#Equal is nicer and doesn't allocate new memory:

```suggestion
		slices.Equal(a.Args, o.Args)
```

nomad/structs/diff.go
```diff
@@ -571,9 +573,152 @@ func (t *Task) Diff(other *Task, contextual bool) (*TaskDiff, error) {
...
+	// Compare Args field
+	if !bytes.Equal(oldArgsJSON, newArgsJSON) || contextual {
+		field := &FieldDiff{
+			Type: "[]string",
+			Name: "Args",
+			Old:  string(oldArgsJSON),
+			New:  string(newArgsJSON),
+		}
+		diff.Fields = append(diff.Fields, field)
+	}
```
tgross, line 662: Also in the `changeScriptDiff` function you can see the use of the `stringSetDiff` helper, which is perfect for comparing a slice of strings for `args`. The name is a bit of a misnomer, it's really a `stringSliceDiff` because it cares about order, which is what you want here.

api/tasks.go
```diff
@@ -1167,3 +1169,9 @@ type WorkloadIdentity struct {
 	ServiceName string        `hcl:"service_name,optional"`
 	TTL         time.Duration `mapstructure:"ttl" hcl:"ttl,optional"`
 }
+
+type Action struct {
+	Name    string   `hcl:"name,label"`
+	Command string  `mapstructure:"command" hcl:"command"`
```
tgross, line 1175: This misalignment should be picked up by `gofmt` but weirdly it wasn't detected by our checks:

```suggestion
	Command string   `mapstructure:"command" hcl:"command"`
```

command/agent/job_endpoint_test.go
```diff
@@ -1601,6 +1607,43 @@ func TestHTTP_JobVersions(t *testing.T) {
...
+											t.Errorf("Expected new 'Args' to be '--utc', got '%s'", argField.New)
+										}
+									}
+								}
+							}
+						}
+					}
+				}
+			}
+		}
```
tgross, line 1645: There's a test suite in `nomad/structs/diff_test.go` that's supposed to cover `Diff`. Doing it here is testing far away from the point where the method is defined.

nomad/structs/diff.go
```diff
@@ -571,9 +571,118 @@ func (t *Task) Diff(other *Task, contextual bool) (*TaskDiff, error) {
...
+
+	// Initialize matching arrays with -1 to indicate unmatched Actions.
+	oldMatches := make([]int, len(old))
+	newMatches := make([]int, len(new))
+	for i := range oldMatches {
+		oldMatches[i] = -1
+	}
+	for i := range newMatches {
+		newMatches[i] = -1
+	}
```
tgross, line 649: Is the idea here that we're allowing the order of action blocks to be switched around without it showing up as a plan update? I don't think we want to do that, because the scheduler will definitely pick this up as a change and then that'll be surprising for folks.

tgross (commented):
Looks like we're getting close!

command/agent/job_endpoint.go
```diff
@@ -1343,6 +1343,15 @@ func ApiTaskToStructsTask(job *structs.Job, group *structs.TaskGroup,
 			Sidecar: apiTask.Lifecycle.Sidecar,
 		}
 	}
+
+	if apiTask.Actions != nil {
+		structsTask.Actions = []*structs.Action{}
```
lgfa29, line 1348: I think we can remove these two lines. `range` over a `nil` slice is just a no-op and `append` will create the slice for you.
https://go.dev/play/p/koPZlOr-s_z

command/agent/job_endpoint.go
```diff
@@ -1385,6 +1394,12 @@ func ApiCSIPluginConfigToStructsCSIPluginConfig(apiConfig *api.TaskCSIPluginConf
 	return sc
 }
 
+func ApiActionToStructsAction(job *structs.Job, action *api.Action, act *structs.Action) {
+	act.Name = action.Name
+	act.Args = slices.Clone(action.Args)
+	act.Command = action.Command
+}
```
lgfa29, line 1401: `Api*ToStructs*` functions should return a new value instead of modifying an input.

nomad/structs/actions.go
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package structs
+
+import "slices"
+
+type Action struct {
```
lgfa29, line 8: Hum...I feel like this name may be too generic 🤔 

The `structs` package deal with all sorts of stuff and a struct called `Action` may not be obvious what it relates to. Perhaps `TaskAction` would be a better name?

nomad/structs/diff.go
lgfa29: I think this code is based on the service diff heuristic from https://github.com/hashicorp/nomad/pull/10965#issuecomment-914727717 where we were trying to fix an incorrect diff caused by the assumption that service names were unique. I'm pretty sure reordering blocks causes a diff, but it would be good to test 😅 

But by finding a "matching" block we can provide a better diff output when a change is detected.

I would need to look deeper into the code, but if `action.name` is expected to be unique then we don't need this heuristic.

lgfa29 (commented):
Submitting some early comments. I will try to give a more in-depth review later today.

command/agent/job_endpoint_test.go
philrenaud: Oh wow, that file is a lot more straightforward looking. Removing this and migrating to it now.

nomad/structs/actions.go
philrenaud: I'm not opposed to this (they started out with that name!) but lately we've been talking about Node Actions and similar concepts, which I think would borrow from this a bit. I'm torn!

nomad/structs/diff.go
philrenaud: ah okay, this is my misinterpretation of what would be considered a change. I'll rework it.

nomad/structs/diff.go
```diff
@@ -571,9 +571,98 @@ func (t *Task) Diff(other *Task, contextual bool) (*TaskDiff, error) {
...
+		}
+
+		if action.Name == newAction.Name {
+			indexMatch = i
+			break
+		}
+	}
+
+	return indexMatch
+}
```
tgross, line 597: This is no longer called, right?

nomad/structs/diff.go
```diff
@@ -571,9 +571,98 @@ func (t *Task) Diff(other *Task, contextual bool) (*TaskDiff, error) {
...
+	}
+
+	// Handle remaining new actions (which have been added)
+	for i := len(old); i < len(new); i++ {
+		if diff := actionDiff(nil, new[i], contextual); diff != nil {
+			diffs = append(diffs, diff)
+		}
+	}
+
+	// Optionally, sort the diffs for more predictable output
```
tgross, line 660: Optionally?

tgross (approved):
LGTM! Just a bit of now-dead code to clean up and this can be merged. Nice work!

nomad/structs/diff.go
philrenaud: Right! thanks

command/agent/job_endpoint.go
```diff
@@ -1385,6 +1390,14 @@ func ApiCSIPluginConfigToStructsCSIPluginConfig(apiConfig *api.TaskCSIPluginConf
 	return sc
 }
 
+func ApiActionToStructsAction(job *structs.Job, action *api.Action, act *structs.Action) *structs.Action {
```
lgfa29, line 1393:
```suggestion
func ApiActionToStructsAction(job *structs.Job, action *api.Action) *structs.Action {
```

nomad/structs/actions.go
lgfa29: Are they similar or the _same_? 

Reusing structs for similar concepts may result in weird interfaces, like attributes that are `nil` in some cases but not in others. And sometimes that's OK. So there's not right answer, just _feels_ at this point. 

My vote would be for `TaskAction`, but I will leave it up to you 😄

lgfa29 (approved):
Some additional cleanup, but good one!

command/agent/job_endpoint.go
philrenaud: Oh! Surprised not to get an "unused param" error here or something. Good catch.

PR #18649, opened 2023-10-03
consul: only fetch Consul tokens for Consul-specific identities

tgross:
Only the workload identities signed specifically for Consul, named `consul_$clusterName` or specific to a service, should result in authenticating to Consul to get tokens.


client/allocrunner/consul_hook.go
```diff
@@ -102,36 +102,12 @@ func (h *consulHook) prepareConsulTokensForTask(job *structs.Job, task *structs.
...
-	// FIXME Fetch from Task.Consul.Cluster once #18557 is in
-	if err := h.getConsulTokens(structs.ConsulDefaultCluster, task.Identity.Name, tokens, req); err != nil {
-		return err
-	}
-
-	// alt identities
+	// get tokens for alt identities for Consul
 	mErr := multierror.Error{}
 	for _, i := range task.Identities {
+		if i.Name != fmt.Sprintf("consul_%s", consulConfig.Name) {
```
pkazmierczak, line 108: I think this isn't correct? `i.Name` will have the new `$taskGroupName-$taskName` format (vide https://github.com/hashicorp/nomad/commit/62a07687753d0056355ec5a7ee6df059830e67ea#diff-297649e31d8e88fd89ff63dac7aff450569cca19009616f08c1d73d1e8239725R7672), so perhaps we should do a substring check or sth?

tgross: Fixed in 2b6f7686620ac62258aca0bfd3e7e5b7144e386b

pkazmierczak (approved):
LGTM!

PR #18650, opened 2023-10-03
WID manager: request signed identities for services

tgross:
Includes changes to WID Manager that make it request signed identities for services, as well as a few improvements to `WIHandle` introduced in #18672.


client/widmgr/widmgr.go
```diff
@@ -61,11 +61,24 @@ type WIDMgr struct {
 func NewWIDMgr(signer IdentitySigner, a *structs.Allocation, logger hclog.Logger) *WIDMgr {
 	widspecs := map[string][]*structs.WorkloadIdentity{}
 	tg := a.Job.LookupTaskGroup(a.TaskGroup)
+
+	for _, service := range tg.Services {
+		if service.Identity != nil {
+			widspecs[service.Identity.Name] = []*structs.WorkloadIdentity{service.Identity}
```
pkazmierczak, line 67: There's something 🐟-y going on here: if we key by `service.Identity.Name`, other parts of the code aren't able to find these. Something looks up by `task.Name`, so we need to investigate.

nomad/structs/workload_id.go
```diff
@@ -222,10 +222,18 @@ type AllocIdentitiesResponse struct {
 	QueryMeta
 }
 
+type WorkloadType int
+
+const (
+	ServiceWorkload WorkloadType = iota
+	TaskWorkload
```
tgross, line 229: Existing default identities are `TaskWorkload`, so swapping the order so that the existing value is the zero-value here makes it slightly less likely for us to mess this up across upgrades.

```suggestion
	TaskWorkload
	ServiceWorkload WorkloadType = iota
```

client/widmgr/widmgr.go
```diff
@@ -54,10 +54,23 @@ type WIDMgr struct {
 func NewWIDMgr(signer IdentitySigner, a *structs.Allocation, logger hclog.Logger) *WIDMgr {
 	widspecs := map[string][]*structs.WorkloadIdentity{}
```
tgross, line 55: Could we use the `WIHandle` as the map key directly? Especially given that the `lastToken` map is keyed by the `WIHandle` it seems like that might make this easier to reason about.

nomad/alloc_endpoint.go
```diff
@@ -566,42 +566,115 @@ func (a *Alloc) SignIdentities(args *structs.AllocIdentitiesRequest, reply *stru
...
+						continue
+					}
+
+					wid := s.Identity
+					if wid.Name != idReq.IdentityName {
+						continue
+					}
+
+					widFound = true
+					claims := structs.NewIdentityClaims(job, out, idReq.WorkloadIdentifier, wid, now)
```
tgross, line 627: We haven't updated the `NewIdentityClaims` method to correctly handle task vs service name here, so right now we're setting the `IdentityClaims.TaskName` to the service name, instead of `IdentityClaims.ServiceName` (which I think we'll need to add as a field so that Consul binding rules work?)

pkazmierczak: good catch, adjusted in f7400b42a92edcd22166b144c3ffcdc73210fbe9

client/widmgr/widmgr.go
```diff
@@ -52,12 +52,25 @@ type WIDMgr struct {
 }
 
 func NewWIDMgr(signer IdentitySigner, a *structs.Allocation, logger hclog.Logger) *WIDMgr {
-	widspecs := map[string][]*structs.WorkloadIdentity{}
+	widspecs := map[structs.WIHandle][]*structs.WorkloadIdentity{}
```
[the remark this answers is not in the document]
pkazmierczak: fixed in 5a0b7252be6d38fae6836d652dde8798f2bfe2f1

client/widmgr/widmgr.go
```diff
@@ -52,12 +52,25 @@ type WIDMgr struct {
...
+			widspecs[*service.IdentityHandle()] = []*structs.WorkloadIdentity{service.Identity}
+		}
+	}
+
 	for _, task := range tg.Tasks {
 		// Omit default identity as it does not expire
-		if len(task.Identities) > 0 {
-			widspecs[task.Name] = helper.CopySlice(task.Identities)
+		for _, i := range task.Identities {
+			widspecs[*task.IdentityHandle(i)] = helper.CopySlice(task.Identities)
```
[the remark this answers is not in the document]
pkazmierczak: fixed in 5a0b7252be6d38fae6836d652dde8798f2bfe2f1, and yes you're right, it simplifies the loops

nomad/alloc_endpoint.go
```diff
@@ -565,43 +566,141 @@ func (a *Alloc) SignIdentities(args *structs.AllocIdentitiesRequest, reply *stru
...
+
+func (a *Alloc) signServices(
+	job *structs.Job,
+	alloc *structs.Allocation,
+	idReq *structs.WorkloadIdentityRequest,
+	reply *structs.AllocIdentitiesResponse,
+	now time.Time,
+) (widFound bool, err error) {
+	// if it's a task service, we need to figure out what is the task it belongs to
+	// FIXME find a way to do this properly
```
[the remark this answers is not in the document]
pkazmierczak: I don't understand. `a.SignIdentities` method isn't aware of task or services, and the only information it gets from the incoming request are the identities. In order to verify that the request contains genuine services, the method needs to check what services does the job contain and compare them. Is there a way to figure out what tasks does the job contain? If there is, I don't know it. Otherwise there's no services to call `IdentityHandle` on. 

Or am I missing something?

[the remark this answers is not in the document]
tgross: No, you're right that we still have to loop over all the services. But supposing we added a method on `WIHandle` like the following:

```go
func (w WIHandle) Equal(o WIHandle) bool {
	return w.IdentityName == o.IdentityName &&
		w.WorkloadIdentifier == o.WorkloadIdentifier &&
		w.WorkloadType == o.WorkloadType
}
```

Then the `signServices` method could be reduced to this:

```go
func (a *Alloc) signServices(
	job *structs.Job,
	alloc *structs.Allocation,
	idReq *structs.WorkloadIdentityRequest,
	reply *structs.AllocIdentitiesResponse,
	now time.Time,
) (widFound bool, err error) {
	wid := idReq.WIHandle

	// services can be on the level of task groups or tasks
	for _, tg := range job.TaskGroups {
		for _, service := range tg.Services {
			if service.IdentityHandle().Equal(wid) {
				return true, a.signIdentities(alloc, service.Identity, idReq, reply, now)
			}
		}
		for _, task := range tg.Tasks {
			for _, service := range task.Services {
				if service.IdentityHandle().Equal(wid) {
					return true, a.signIdentities(alloc, service.Identity, idReq, reply, now)
				}
			}
		}
	}
	return
}
```

(This example also assumes we change `foundAndSignedIdentities` to just `signIdentities` because we know we've found the identity at that point.)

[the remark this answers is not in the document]
pkazmierczak: Oooh I get it now, that's good thinking, thanks.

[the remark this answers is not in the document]
pkazmierczak: addressed in c56f04a25ea6c33a4b84fce80500be710162fd75

tgross:
LGTM! But because I started the PR it turns out I can't approve it. So let's get @lgfa29 or @schmichael to give it the :heavy_check_mark:

nomad/alloc_endpoint.go
```diff
@@ -565,43 +565,113 @@ func (a *Alloc) SignIdentities(args *structs.AllocIdentitiesRequest, reply *stru
...
+) (widFound bool, err error) {
+	for _, wid := range task.Identities {
+		if wid.Name != idReq.IdentityName {
 			continue
 		}
+
+		widFound = true
+		err = a.signIdentities(alloc, wid, idReq, reply, now)
 	}
+	return
```
lgfa29, line 628: Do we expect to only have a single match her? In case there are multiple we may lose an `err`. For example, if the first match errors, but the second succeeds `err` will return `nil`.

We could make the intention of a single match more clear by `break`in the loop on match instead of `continue`ing on mismatch, or a `return` like in `signServices` below.

nomad/alloc_endpoint.go
```diff
@@ -565,43 +565,113 @@ func (a *Alloc) SignIdentities(args *structs.AllocIdentitiesRequest, reply *stru
...
+	reply *structs.AllocIdentitiesResponse,
+	now time.Time,
+) error {
+	claims := structs.NewIdentityClaims(alloc.Job, alloc, &idReq.WIHandle, wid, now)
+	token, _, err := a.srv.encrypter.SignClaims(claims)
+	if err != nil {
+		return err
+	}
+	reply.SignedIdentities = append(reply.SignedIdentities, &structs.SignedWorkloadIdentity{
+		WorkloadIdentityRequest: *idReq,
```
lgfa29, line 671: Another minor nit-picking safe to ignore, but I think we could take `structs.WorkloadIdentityRequest` instead of a pointer here and in the `sign*` methods above and dereference it way earlier in `SignIdentities` so we don't need to worry about potential `nil` inputs.

nomad/structs/workload_id.go
```diff
@@ -222,10 +222,24 @@ type AllocIdentitiesResponse struct {
 	QueryMeta
 }
 
+type WorkloadType int
+
+const (
+	TaskWorkload WorkloadType = iota
+	ServiceWorkload
```
lgfa29, line 229:
```suggestion
	WorkloadTypeTask WorkloadType = iota
	WorkloadTypeService
```

Just me being annoying again, feel free to ignore 😅 

When reading in isolation `TaskWorkload` and `ServiceWorkload` sound like they refer to the workload of a task/service.

lgfa29 (approved):
Minor comments, all safe to ignore 😄

PR #18652, opened 2023-10-04
vault: eliminate vaultclient test import cycle

pkazmierczak:
This PR eliminates the vaultclient test import cycle by putting the test file into a separate package and making vaultclient objects public. 

Ref https://github.com/hashicorp/team-nomad/issues/404

### DIFF
--- a/client/vaultclient/vaultclient.go
+++ b/client/vaultclient/vaultclient.go
@@ -55,58 +55,58 @@ type VaultClient interface {
 	StopRenewToken(string) error
 }
 
-// Implementation of VaultClient interface to interact with vault and perform
-// token and lease renewals periodically.
-type vaultClient struct {
-	// tokenDeriver is a function pointer passed in by the client to derive
+// Client is the implementation of VaultClient interface to interact with vault
+// and perform token and lease renewals periodically.
+type Client struct {
+	// TokenDeriver is a function pointer passed in by the Vault to derive
 	// tokens by making RPC calls to the nomad server. The wrapped tokens
 	// returned by the nomad server will be unwrapped by this function
 	// using the vault API client.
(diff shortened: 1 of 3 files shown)


tgross (commented):
Having a test file with a different package name in the same directory as the package under test feels a little weird to me. Can we lift this test file up into the `client` package maybe? I'm not sure which import creates the cycle.

Also, we didn't catch this originally because this package was in the list of ["uninteresting"](https://github.com/hashicorp/nomad/blob/main/tools/missing/main.go#L159) packages for the linter that verifies we're testing all packages, and it was left out of [`test-core.json`](https://github.com/hashicorp/nomad/blob/main/ci/test-core.json) that feeds packages to a matrix of GHA runners. Wherever the test lands, we should make sure we're running it. We can probably add this package to the `client` group? Maybe check how long that currently takes vs the `quick` group.

pkazmierczak:
> Having a test file with a different package name in the same directory as the package under test feels a little weird to me. 

That's because it is! But there's [a precedent](https://github.com/hashicorp/nomad/blob/main/client/widmgr/widmgr_test.go) and I thought that perhaps this is the way we usually handle import cycles in tests. 

> Can we lift this test file up into the `client` package maybe? I'm not sure which import creates the cycle.

I'm completely fine moving it to the `client` package, and yes, the `client` import causes the cycle.

tgross:
> But there's a precedent and I thought that perhaps this is the way we usually handle import cycles in tests.

Hm, I bet we're not testing that package either as a result. We should double-check that works with the way the GHA tooling is configured.

pkazmierczak:
@tgross I moved the tests to `client/`

tgross (approved):
LGTM!

PR #18661, opened 2023-10-04
WID manager: save and restore signed WIs from client state DB

tgross:
When clients are restarted and the identity hook runs when we restore allocations, the running allocations are likely to have already-signed Workload Identities that are unexpired. Save these to the client's local state DB so that we can avoid a thundering herd of RPCs during client restart. When we restore, we'll check if there's at least one expired signed WI before making any initial signing request.

Included:
* Renames `getIdentities` to `getInitialIdentities` to make the workflow more clear.
* Renames the existing `widmgr_test.go` file of integration tests, which is in its own package to avoid circular imports to `widmgr_int_test.go`

Currently WIP because we'll need https://github.com/hashicorp/nomad/pull/18650 for it to make any sense for service identities, which @pkazmierczak is currently working on.

### DIFF
--- a/client/widmgr/widmgr.go
+++ b/client/widmgr/widmgr.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-hclog"
+	cstate "github.com/hashicorp/nomad/client/state"
 	"github.com/hashicorp/nomad/helper"
 	"github.com/hashicorp/nomad/nomad/structs"
 )
@@ -30,6 +31,7 @@ type WIDMgr struct {
 	minIndex                uint64
 	widSpecs                map[string][]*structs.WorkloadIdentity // task -> WI
 	signer                  IdentitySigner
+	db                      cstate.StateDB
 
 	// lastToken are the last retrieved signed workload identifiers keyed by
 	// TaskIdentity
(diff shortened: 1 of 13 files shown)


pkazmierczak (approved):
LGTM! And thanks for fixing the widmgr test package, too.

client/widmgr/widmgr.go
```diff
@@ -186,8 +195,37 @@ func (m *WIDMgr) Shutdown() {
 	}
 }
 
-// getIdentities fetches all signed identities or returns an error.
-func (m *WIDMgr) getIdentities() error {
+// restoreStoredIdentities recreates the state of the WIDMgr from a previously
+// saved state, so that we can avoid asking for all identities again after a
+// client agent restart. It returns true if the caller should immediately call
+// getIdentities because one or more of the identities is expired.
```
lgfa29, line 201: Would re-fetching all identities (even if not expired) caused a task to restart? I think we've seen cases where tasks are unexpectedly restarted, so it would be good to avoid them here as well.

lgfa29 (commented):
Sorry, late to the party 😅 

But just had a small question on this.

client/widmgr/widmgr.go
tgross: We don't have any kind of change mode on identities (yet), so no this won't cause tasks to restart even in this restore path. The `identity_hook` will write new JWTs and expose them via hook resources and that's it. It's up to the application to detect that (if it needs to) for now.

PR #18664, opened 2023-10-04
client: remove null dynamic metadata keys

### LINKED ISSUE #18603: Unsetting Nomad metadata leaks keys
### Nomad version
We tested with Nomad v1.5.5, but we think this reproduces in tip.

Unix

### Issue
Unsetting node metadata keys leave them as null on the client node until it restarts. This causes steady increase in memory, and there doesn’t seem to be another way of removing the keys otherwise.

#### Description
The [Update Node Metadata HTTP API](https://developer.hashicorp.com/nomad/api-docs/client#update-node-metadata) allows us to dynamically set key-value metadata. The documentation mentions that one can ‘unset’ keys by requesting updates with value=null.

After waiting 10s for the node data to be propagated to the scheduler, the response from [Read Node Metadata HTTP API](https://developer.hashicorp.com/nomad/api-docs/client#read-node-metadata) shows that the unset key is removed from the ‘Meta’ field, but remains in ‘Dynamic’. In our local reproduction, we inspected the RAFT state and confirmed that the scheduler does not track Meta keys with null values, so this issue is limited to client nodes.

The only way to clear unset keys from ‘Dynamic’ was to restart the node. In code, the client clones the ‘Dynamic’ meta map ([code](https://github.com/hashicorp/nomad/blob/main/client/meta_endpoint.go#L55)) and thus does not end up deleting the keys as we do in standard ‘Meta’.

We suspect that this behavior exists in order to be able to unset static Meta keys. However, keeping keys with null values which are not in ‘Static’ metadata seems meaningless and undesirable because the client accumulates memory indefinitely. Ideally we should only persist keys with null values if those same keys exist in ‘Static’ metadata.

#### Proposal
Persist ‘Dynamic’ metadata keys with null values only if those same keys exist in ‘Static’ metadata.

### Reproduction steps
Start a local cluster with 1 client and 1 server, Nomad v1.5.5:
```
$ cat server_config.hcl
data_dir=”/tmp”
log_level = "TRACE"
advertise {
  http = "127.0.0.1"
  rpc = "127.0.0.1"
  serf = "127.0.0.1"
}
server {
  enabled = true
  bootstrap_expect = 1
  job_gc_interval = "1m"
  job_gc_threshold = "24h"
  eval_gc_threshold = "1m"
}

$ cat client_config.hcl
data_dir = "/tmp"
log_level = "debug"
advertise {
  http = "127.0.0.1"
  rpc = "127.0.0.1"
  serf = "127.0.0.1"
}
ports {
  http = "9876"
  rpc = "9875"
  serf = "9874"
}
client {
  enabled = true
  servers = ["127.0.0.1"]
  gc_max_allocs = 1
}
plugin "raw_exec" {
  config {
    enabled = true
  }
}
```

Add a “test-key” metadata with value null that doesn’t exist as static metadata:
`$ echo "{\"Meta\": { \"test-key\": null}}" | nomad operator api /v1/client/metadata?node_id=NODE_ID`

#### Expected Result
After waiting 10s, 
`$ nomad operator api /v1/client/metadata?node_id=NODE_ID`
will not include `test-key: null` in the 'Dynamic' section of metadata since `test-key` is not in 'Static'.

#### Actual Result
Regardless of waiting 10s or send SIGHUP to your nomad client, the unset key remains in Dynamic:
```
$ nomad operator api /v1/client/metadata?node_id=NODE_ID 
{
  "Meta": {
    "connect.gateway_image": "envoyproxy/envoy:v${NOMAD_envoy_version}",
    "connect.log_level": "info",
    "connect.proxy_concurrency": "1",
    "connect.sidecar_image": "envoyproxy/envoy:v${NOMAD_envoy_version}"
  },
  "Dynamic": {
    "test-key": null
  },
  "Static": {
    "connect.gateway_image": "envoyproxy/envoy:v${NOMAD_envoy_version}",
    "connect.log_level": "info",
    "connect.proxy_concurrency": "1",
    "connect.sidecar_image": "envoyproxy/envoy:v${NOMAD_envoy_version}"
  }
}
```

lgfa29:
Setting a null value to a node metadata is expected to remove it from subsequent reads. This is true both for static node metadata (defined in the agent configuration file) as well as for dynamic node metadata (defined via the Nomad API).

Null values for static metadata must be persisted to indicate that the value has been removed, but strictly dynamic metadata null values can be removed from state and client memory.

Closes #18603

### DIFF
--- a/.changelog/18664.txt
+++ b/.changelog/18664.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+client: ensure null dynamic node metadata values are removed from memory
+```
--- a/client/meta_endpoint.go
+++ b/client/meta_endpoint.go
@@ -45,6 +45,16 @@ func (n *NodeMeta) Apply(args *structs.NodeMetaApplyRequest, reply *structs.Node
 		dyn = maps.Clone(n.c.metaDynamic)
 		maps.Copy(dyn, args.Meta)
 
+		// Delete null values from the dynamic metadata if they are also not
+		// static. Static null values must be kept so their removal is
+		// persisted in client state.
+		for k, v := range args.Meta {
+			_, static := n.c.metaStatic[k]
+			if v == nil && !static {
+				delete(dyn, k)
+			}
+		}
+
 		if stateErr = n.c.stateDB.PutNodeMeta(dyn); stateErr != nil {
 			return
 		}
--- a/client/meta_endpoint_test.go
+++ b/client/meta_endpoint_test.go
@@ -96,3 +96,88 @@ func TestNodeMeta_Validation(t *testing.T) {
 	err = c1.ClientRPC("NodeMeta.Apply", applyReq, &resp)
 	must.ErrorContains(t, err, "*")
 }
+
+func TestNodeMeta_unset(t *testing.T) {
+	ci.Parallel(t)
+
+	s, cleanupS := nomad.TestServer(t, nil)
+	defer cleanupS()
+	testutil.WaitForLeader(t, s.RPC)
+
+	c1, cleanup := TestClient(t, func(c *config.Config) {
+		c.Servers = []string{s.GetConfig().RPCAddr.String()}
+		c.Node.Meta["static_meta"] = "true"
+	})
+	defer cleanup()
+
+	// Set dynamic node metadata.
+	applyReq := &structs.NodeMetaApplyRequest{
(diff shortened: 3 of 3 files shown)


schmichael (approved):

jrasell (approved):
LGTM!

Issue #18578, opened 2023-09-26
Add time zone data for Windows


wizpresso-steve-cy-fan:
### Proposal

Due to https://github.com/golang/go/issues/38453, specifying time zone for periodic job is actually broken in Windows, which will promptly fail the periodic job:

https://github.com/hashicorp/nomad/blob/20eadc7b296cc81d2c8c01fb15cb74047633788f/api/jobs.go#L897-L903

### Use-cases

For periodic job.

### Attempted Solutions

Use Local time. 

```hcl
periodic {
  cron             = "@hourly"
  time_zone        = "Local"
}
```

wizpresso-steve-cy-fan:
https://github.com/golang/go/issues/38017

There are two ways to do this.

lgfa29:
Thanks for the report @wizpresso-steve-cy-fan!

I've opened https://github.com/hashicorp/nomad/pull/18676 to add the `timetzdata` build flag on Windows so the time zone data is embedded into the binary.

PR #18676, opened 2023-10-05
build: load time/tzdata on Windows

lgfa29:
Nomad uses `time.LoadLocation()` to translate a periodic job time zone string value to a `time.Location`. From godocs:

    LoadLocation looks for the IANA Time Zone database in the following locations in order:

    * the directory or uncompressed zip file named by the ZONEINFO environment variable
    * on a Unix system, the system standard installation location
    * $GOROOT/lib/time/zoneinfo.zip
    * the time/tzdata package, if it was imported

So non-Unix systems require Go to be installed or `time/tzdata` to be imported, otherwise running periodic jobs with a specific `time_zone` value results in an error:

    Invalid time zone "America/Toronto": unknown time zone America/Toronto

This commit adds the `timetzdata` build tag on Windows to embed the time zone data into the final binary. This results in a slightly bigger binary, but from `time/tzdata` godocs:

    Importing this package will increase the size of a program by about 450 KB.
    [..]
    This package will be automatically imported if you build with -tags timetzdata.

Closes https://github.com/hashicorp/nomad/issues/18578

### DIFF
--- a/.changelog/18676.txt
+++ b/.changelog/18676.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+build: Add `timetzdata` Go build tag on Windows binaries to embed time zone data so periodic jobs are able to specify a time zone value on Windows environments
+```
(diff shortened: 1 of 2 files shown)


lgfa29:
Here's an example of running a periodic job with binaries that have and don't have the time zone data embedded:

image

gulducat (approved):
:clock1:

PR #18679, opened 2023-10-06
scheduler: fix host volume feasibility check

### LINKED ISSUE #18567: Host volume incorrect scheduling
### Nomad version
1.6.1+ent

### Issue
When using `per_alloc` on a host_volume Nomad initially schedules all allocations on the same host even though the host volume for that alloc isn't available on the respective client. Nomad then fails to start the allocation with error:
```
Time                       Type              Description
2023-09-25T17:22:47+02:00  Killing           Sent interrupt. Waiting 5s before force killing
2023-09-25T17:22:43+02:00  Not Restarting    Error was unrecoverable
2023-09-25T17:22:43+02:00  Task hook failed  volumes: host volume validation error: 1 error occurred:
	* missing test-host-volume[1]
2023-09-25T17:22:43+02:00  Task Setup        Building Task Directory
2023-09-25T17:22:35+02:00  Received          Task received by client
```

### Reproduction steps
Setup 2 local Nomad agents each with a host volume:
```hcl
client {
  host_volume "test-host-volume[0]" {
    path      = "/path/to/host_volumes/test-host-volume-0"
    read_only = false
  }
}
```
and the second one:
```hcl
client {
  host_volume "test-host-volume[1]" {
    path      = "/path/to/host_volumes/test-host-volume-1"
    read_only = false
  }
}
```

Deploy a test job:
```hcl
job "test" {
  region      = "us-east-1"
  datacenters = ["*"]

  group "test" {
    count = 2
    volume "data" {
      type      = "host"
      source    = "test-host-volume"
      per_alloc = true
    }

    task "example" {
      kill_timeout = "29s"
      driver = "docker"
      config {
        image = "alpine:3.18.2"
        args  = ["tail", "-f", "/dev/null"]
      }

      volume_mount {
        volume      = "data"
        destination = "/data"
      }

      resources {
        cpu    = 100
        memory = 32
      }
    }
  }
}
```

#### Expected Result
Without any constraint there should be 2 allocs, 1 assigned to each node.

#### Actual Result
```
$ nomad job run test.hcl
==> 2023-09-25T18:07:06+02:00: Monitoring evaluation "e10f6cf4"
    2023-09-25T18:07:06+02:00: Evaluation triggered by job "test"
    2023-09-25T18:07:07+02:00: Evaluation within deployment: "c5b2b8d2"
    2023-09-25T18:07:07+02:00: Allocation "3b654f6b" created: node "3f1c9f73", group "test"
    2023-09-25T18:07:07+02:00: Allocation "61a7b875" created: node "3f1c9f73", group "test"
    2023-09-25T18:07:07+02:00: Evaluation status changed: "pending" -> "complete"
==> 2023-09-25T18:07:07+02:00: Evaluation "e10f6cf4" finished with status "complete"
==> 2023-09-25T18:07:07+02:00: Monitoring deployment "c5b2b8d2"
```
Both allocs are scheduled to node id `3f1c9f73`. After a failure starting the alloc on the incorrect node:
```
Recent Events:
Time                       Type              Description
2023-09-25T18:07:07+02:00  Not Restarting    Error was unrecoverable
2023-09-25T18:07:07+02:00  Task hook failed  volumes: host volume validation error: 1 error occurred:
	* missing test-host-volume[1]
2023-09-25T18:07:07+02:00  Task Setup        Building Task Directory
2023-09-25T18:07:07+02:00  Received          Task received by client
```
After a reschedule action it does end up on the correct node:
```
Latest Deployment
ID          = 848ead7b
Status      = running

Deployed
test        2        3       1        1          2023-09-25T18:15:10+02:00

Allocations
ID        Node ID   Task Group  Version  Desired  Status   Created  Modified
2f775ba3  3f1c9f73  test        0        run      running  2s ago   2s ago
49473f08  b7a0c6ec  test        0        run      running  33s ago  21s ago
ec6df442  b7a0c6ec  test        0        stop     failed   33s ago  2s ago
```

lgfa29:
Host volumes were considered regular feasibility checks. This had two unintended consequences.

The first happened when scheduling an allocation with a host volume on a set of nodes with the same computed class but where only some of them had the desired host volume.

If the first node evaluated did not have the host volume, the entire node class was considered ineligible for the task group.

```go
// Run the job feasibility checks.
for _, check := range w.jobCheckers {
	feasible := check.Feasible(option)
	if !feasible {
		// If the job hasn't escaped, set it to be ineligible since it
		// failed a job check.
		if !jobEscaped {
			evalElig.SetJobEligibility(false, option.ComputedClass)
		}
		continue OUTER
	}
}
```

This results in all nodes with the same computed class to be skipped, even if they do have the desired host volume.

```go
switch evalElig.JobStatus(option.ComputedClass) {
case EvalComputedClassIneligible:
	// Fast path the ineligible case
	metrics.FilterNode(option, "computed class ineligible")
	continue
```

The second consequence is somewhat the opposite. When an allocation has a host volume with `per_alloc = true` the node must have a host volume that matches the allocation index, so each allocation is likely to be placed in different nodes.

But when the first allocation found a node match, it registered the node class as eligible for the task group.

```go
// Set the task group eligibility if the constraints weren't escaped and
// it hasn't been set before.
if !tgEscaped && tgUnknown {
	evalElig.SetTaskGroupEligibility(true, w.tg, option.ComputedClass)
}
```

This could cause other allocations to be placed on nodes without the expected host volume because of the computed node class fast path. The node feasibility for the volume was never checked.

```go
case EvalComputedClassEligible:
	// Fast path the eligible case
	if w.available(option) {
		return option
	}
	// We match the class but are temporarily unavailable
	continue OUTER
```

These problems did not happen with CSI volumes kind of accidentally. Since the `CSIVolumeChecker` was not placed in the `tgCheckers` list it did not cause the node class to be considered ineligible on failure (avoiding the first problem).

And, as illustrated in the code snippet above, the eligible node class fast path checks `tgAvailable` (where `CSIVolumeChecker` is placed) before returning the option (avoiding the second problem).

By also placing `HostVolumeChecker` in the `tgAvailable` list instead of `tgCheckers` we also avoid these problems on host volume feasibility.

Closes #18567

----

Note to reviewers: since both volume feasibility checks completely by-pass the computed class aspect of `FeasibilityWrapper` (which is kind of the whole point of this iterator 😅) we could refactor this code to extract the volume checkers to a separate iterator and keep `FeasibilityWrapper` just about computed class, but I kept the changeset minimum for now.

### DIFF
--- a/scheduler/stack_test.go
+++ b/scheduler/stack_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/hashicorp/nomad/ci"
 	"github.com/hashicorp/nomad/nomad/mock"
 	"github.com/hashicorp/nomad/nomad/structs"
+	"github.com/shoenig/test/must"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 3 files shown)


scheduler/stack_test.go
```diff
@@ -247,6 +248,86 @@ func TestServiceStack_Select_DriverFilter(t *testing.T) {
...
+	must.Eq(t, option.Node.ID, node1.ID)
+
+	// Second alloc selects node with volume per_alloc[1].
+	selectOptions = &SelectOptions{
+		AllocName: structs.AllocName(job.Name, job.TaskGroups[0].Name, 1),
+	}
+	option = stack.Select(job.TaskGroups[0], selectOptions)
+	must.NotNil(t, option)
+	must.Eq(t, option.Node.ID, node2.ID)
+}
```
tgross, line 329: For completeness, it might be nice to have a 3rd allocation so that we can verify that it doesn't get placed at all.

tgross (approved):
Nice work hunting this down @lgfa29! 

Looking at the list of other "classful" feasibility checkers, I wonder if we have a similar bug lurking for devices. Might be worth a quick scan through open issues to see if there's follow-up work we can hit with this fix too.

lgfa29:
> Looking at the list of other "classful" feasibility checkers, I wonder if we have a similar bug lurking for devices. Might be worth a quick scan through open issues to see if there's follow-up work we can hit with this fix too.

I looked through some issues but didn't find anything relevant. I also forgot to mention that I found the first problem described while writing tests, AFAICT it hasn't been reported either.

But I think devices do affect the node's computed class, so they may be safe from this bug:
https://github.com/hashicorp/nomad/blob/e3c8700ded891702b7f94636033109efd8f71c3a/nomad/structs/node_class.go#L82-L107

I should also note that I did try to add host volumes to the computed class, but it create a different problem with `per_alloc` where the node class for a node with a volume like `my_volume[1]` was marked as ineligible while looking for node with ``my_volume[0]`.

PR #18689, opened 2023-10-06
cli: remove default for ACL token type on update

lgfa29:
With a default value set to `client`, the `nomad acl token update` command can silently downgrade a management token to client on update if the command does not specify `-type=management` on every update.


jrasell (approved):
LGTM!

PR #18691, opened 2023-10-07
OpenID Configuration Discovery Endpoint (just the endpoint)

schmichael:
Added the [OIDC Discovery](https://openid.net/specs/openid-connect-discovery-1_0.html) `/.well-known/openid-configuration` endpoint to Nomad, but it is only enabled if the `server.oidc_issuer` parameter is set. Documented the parameter, but without a tutorial trying to actually _use_ this will be very hard.

I intentionally did *not* use https://github.com/hashicorp/cap for the OIDC configuration struct because it's built to be a *compliant* OIDC provider. Nomad is *not* trying to be compliant initially because compliance to the spec does not guarantee it will actually satisfy the requirements of third parties. I want to avoid the problem where in an attempt to be standards compliant we ship configuration parameters that lock us in to a certain behavior that we end up regretting. I want to add parameters and behaviors as there's a demonstrable need.

Users always have the escape hatch of providing their own OIDC configuration endpoint. Nomad just needs to know the Issuer so that the JWTs match the OIDC configuration. There's no reason the actual OIDC configuration JSON couldn't live in S3 and get served directly from there. Unlike JWKS the OIDC configuration should be static, or at least change very rarely.

This PR is just the endpoint extracted from #18535. The `RS256` algorithm still needs to be added in hopes of supporting third parties such as [AWS IAM OIDC Provider](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_oidc.html).

### DIFF
--- a/command/agent/config.go
+++ b/command/agent/config.go
@@ -701,6 +701,11 @@ type ServerConfig struct {
 
 	// JobTrackedVersions is the number of historic job versions that are kept.
 	JobTrackedVersions *int `hcl:"job_tracked_versions"`
+
+	// OIDCIssuer if set enables OIDC Discovery and uses this value as the
+	// issuer. Third parties such as AWS IAM OIDC Provider expect the issuer to
+	// be a publically accessible HTTPS URL signed by a trusted well-known CA.
+	OIDCIssuer string `hcl:"oidc_issuer"`
 }
 
 func (s *ServerConfig) Copy() *ServerConfig {
@@ -2119,6 +2124,10 @@ func (s *ServerConfig) Merge(b *ServerConfig) *ServerConfig {
 
 	if b.JobTrackedVersions != nil {
 		result.JobTrackedVersions = b.JobTrackedVersions
+	}
+
+	if b.OIDCIssuer != "" {
+		result.OIDCIssuer = b.OIDCIssuer
 	}
 
 	// Add the schedulers
--- a/command/agent/keyring_endpoint.go
+++ b/command/agent/keyring_endpoint.go
@@ -83,7 +83,7 @@ func (s *HTTPServer) JWKSRequest(resp http.ResponseWriter, req *http.Request) (a
 // See https://openid.net/specs/openid-connect-discovery-1_0.html for details.
 func (s *HTTPServer) OIDCDiscoveryRequest(resp http.ResponseWriter, req *http.Request) (any, error) {
 	if req.Method != http.MethodGet {
-		return nil, CodedError(405, ErrInvalidMethod)
+		return nil, CodedError(http.StatusMethodNotAllowed, ErrInvalidMethod)
 	}
 
 	args := structs.GenericRequest{}
--- a/command/agent/keyring_endpoint.go
+++ b/command/agent/keyring_endpoint.go
@@ -77,6 +77,31 @@ func (s *HTTPServer) JWKSRequest(resp http.ResponseWriter, req *http.Request) (a
 	return out, nil
 }
 
+// OIDCDiscoveryRequest implements the OIDC Discovery protocol for using
+// workload identity JWTs with external services.
+//
+// See https://openid.net/specs/openid-connect-discovery-1_0.html for details.
+func (s *HTTPServer) OIDCDiscoveryRequest(resp http.ResponseWriter, req *http.Request) (any, error) {
+	if req.Method != http.MethodGet {
+		return nil, CodedError(405, ErrInvalidMethod)
+	}
+
+	args := structs.GenericRequest{}
+	if s.parse(resp, req, &args.Region, &args.QueryOptions) {
+		return nil, nil
+	}
+	var rpcReply structs.KeyringGetConfigResponse
+	if err := s.agent.RPC("Keyring.GetConfig", &args, &rpcReply); err != nil {
+		return nil, err
+	}
+
+	if rpcReply.OIDCDiscovery == nil {
+		return nil, CodedError(http.StatusNotFound, "OIDC Discovery endpoint disabled")
+	}
+
+	return rpcReply.OIDCDiscovery, nil
+}
+
 // KeyringRequest is used route operator/raft API requests to the implementing
 // functions.
 func (s *HTTPServer) KeyringRequest(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
--- a/command/agent/keyring_endpoint_test.go
+++ b/command/agent/keyring_endpoint_test.go
@@ -83,6 +83,8 @@ func TestHTTP_Keyring_CRUD(t *testing.T) {
 	})
 }
 
+// TestHTTP_Keyring_JWKS asserts the JWKS endpoint is enabled by default and
+// caches relative to the key rotation threshold.
 func TestHTTP_Keyring_JWKS(t *testing.T) {
 	ci.Parallel(t)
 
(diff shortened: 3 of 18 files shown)


command/agent/config.go
```diff
@@ -701,6 +701,11 @@ type ServerConfig struct {
 
 	// JobTrackedVersions is the number of historic job versions that are kept.
 	JobTrackedVersions *int `hcl:"job_tracked_versions"`
+
+	// OIDCIssuer if set enables OIDC Discovery and uses this value as the
+	// issuer. Third parties such as AWS IAM OIDC Provider expect the issuer to
+	// be a publically accessible HTTPS URL signed by a trusted well-known CA.
+	OIDCIssuer string `hcl:"oidc_issuer"`
```
lgfa29, line 708: If there's a possibility we add more OIDC configs based on needs, would it make sense to encapsulate them in a block instead of a top-level string?

And we don't update this as much as we should, but could we validate the shape of this value in the `nomad config validate` command? It seems like there's some specific expectations to how it should look like.
https://github.com/hashicorp/nomad/blob/0410b8acea5e6ac1d032f6b98ce1b893f0230d92/command/agent/command.go#L311

command/agent/keyring_endpoint.go
```diff
@@ -77,6 +77,31 @@ func (s *HTTPServer) JWKSRequest(resp http.ResponseWriter, req *http.Request) (a
 	return out, nil
 }
 
+// OIDCDiscoveryRequest implements the OIDC Discovery protocol for using
+// workload identity JWTs with external services.
+//
+// See https://openid.net/specs/openid-connect-discovery-1_0.html for details.
+func (s *HTTPServer) OIDCDiscoveryRequest(resp http.ResponseWriter, req *http.Request) (any, error) {
+	if req.Method != http.MethodGet {
+		return nil, CodedError(405, ErrInvalidMethod)
```
lgfa29, line 86:
```suggestion
		return nil, CodedError(http.StatusMethodNotAllowed, ErrInvalidMethod)
```

command/agent/keyring_endpoint_test.go
```diff
@@ -77,3 +82,81 @@ func TestHTTP_Keyring_CRUD(t *testing.T) {
...
+		must.NoError(t, err)
+
+		_, err = s.Server.OIDCDiscoveryRequest(respW, req)
+		must.ErrorContains(t, err, "OIDC Discovery endpoint disabled")
+		codedErr := err.(HTTPCodedError)
+		must.Eq(t, http.StatusNotFound, codedErr.Code())
+	})
+}
+
+// TestHTTP_Keyring_OIDCDisco_Disabled asserts that the OIDC Discovery endpoint
```
lgfa29, line 137:
```suggestion
// TestHTTP_Keyring_OIDCDisco_Enabled asserts that the OIDC Discovery endpoint
```

website/content/docs/configuration/server.mdx
```diff
@@ -267,6 +267,14 @@ server {
 - `job_tracked_versions` `(int: 6)` - Specifies the number of historic job versions that
   are kept.
 
+- `oidc_issuer` `(string: "")` - Specifies the Issuer URL for [Workload
```
tgross, line 270: Do we want to give an example of the form that Issuer URL is supposed to take? Especially with the next sentence of the text showing the path component of the well-known endpoint, this could be confusing. Something like

> Specifies the Issuer URL for Workload Identity JWTs, which should include only the scheme and authority (and optionally port). For example, "https://nomad.example.com".


I'm also not sure if the spec cares about the trailing slash? Some of the issues in your test code in this PR have it and half don't. But if that's important, we should include a note about that here too. (I suspect _Nomad_ doesn't care with our HTTP router, but third-parties might.)

website/content/docs/configuration/server.mdx
```diff
@@ -267,6 +267,14 @@ server {
...
+  Identity][wi] JWTs. If set the `/.well-known/openid-configuration` HTTP
+  endpoint is enabled for third parties to discover Nomad's OIDC configuration.
+  Once set `oidc_issuer` *cannot be changed* without invalidating Workload
```
lgfa29, line 273: How bad would it be for this config to drift across servers, like a typo or an old machine coming back up for some reason?

tgross (approved):
LGTM!

lgfa29 (approved):
Some minor ramblings, but LGTM 😄

website/content/docs/configuration/server.mdx
lgfa29: Yeah, I was wondering about storing this in state. I don't have any strong opinions, and I don't know if we have any prescriptive guidance of what should go where 🤔 

Some aspects I just pulled out of thin air:

| Config file | State |
| ---------- | ----- |
| Required on agent start | Not required on agent start |
| Specific to each node | Must be the same in all nodes |
| Requires restart to apply change | Can be changed live |

So I feel like this falls into the "state" category? But I also don't feel strongly about my table either 😅

lgfa29: Yeah, a more holistic strategy for configuration management would be great!

I'm good with the changes here as they are, I just smelled a potentially bad footgun and wanted to confirm it was indeed a bad problem 😅

website/content/docs/configuration/server.mdx
schmichael: > Do we want to give an example of the form that Issuer URL is supposed to take?

Good call. Actually using all of this plumbing is going to take a Tutorial, so I was trying to be minimal with the parameter doc, but an example is always a good idea.

> I'm also not sure if the spec cares about the trailing slash? ... **but third-parties might.**

Emphasis mine. AWS's IAM OIDC Identity Provider works either way, but who knows about others. Nomad itself doesn't. (Nomad doesn't actually use this value for anything functionally other than "is it set or not"!)

HTTP routing ignores this because a user may configure their proxy and issuer to be something like: `https://example.com/nomad`. However there's no need for Nomad to use the `/nomad` prefix for its OIDC Discovery endpoint: the proxy can translate the paths.

> Some of the issuers in your test code in this PR have it and half don't.

:facepalm: Fixing...

Looks like [Terraform](https://app.terraform.io/.well-known/openid-configuration) and [Google](https://accounts.google.com/.well-known/openid-configuration) both omit the trailing slash, so that's good enough for me!

command/agent/config.go
schmichael: > If there's a possibility we add more OIDC configs based on needs

I brainstormed in Slack a bit and just didn't feel confident that I could come up with a new block that was any more likely to stand the test of time than this parameter. :disappointed:

`oidc { issuer = "..." }` is the obvious choice, but since Nomad has an OIDC Auth Method I thought that would be too confusing.

`workload_identity { oidc_issuer = "..." }` is wordy but at least is specific to this use case. That being said I think it's even odds we reuse this infrastructure for Node Identities (so a Node gets a JWT that it can perform operations with)... at which point we're full of regret for sticking this under `workload_identity`.

I also don't have an idea of what further parameters we'll need. We're probably stuck with `RS256` as the only reasonable algorithm. `jwks_uri` is built in a well-defined way on top of `issuer`, so I don't see why we would need to configure that distinctly.

So in the absence of a clear winner, I think putting it in `server` at least makes it clear where it's actually used (server-only). A feature that top-level blocks would lack.

> And we don't update this as much as we should, but could we validate the shape of this value in the nomad config validate command? 

Thanks for reminding me! Added. Kind of awkward we don't reuse that during agent startup, but it's better than nothing.

Issue #18538, opened 2023-09-20
Some services are being registered all the time


jorgemarey:
### Nomad version
Nomad v1.5.6

### Issue
We had a problem where we saw some consul service flapping to critical for a moment (less than a second) and then become healthy again. While doing some investigation about this we saw that nomad is re-registering the services constantly. 

In the client logs we can see:

```
2023-09-20T10:04:07.331Z [TRACE] consul.sync: execute sync: reason=periodic
2023-09-20T10:04:07.406Z [TRACE] consul.sync: registrations different: id=_nomad-task-6de92b6e-9099-654a-00f2-93ac2b255142-xxxx-8080 field=connect_sidecar wanted="\"xxxx\"" existing="\"xxxx\""
2023-09-20T10:04:07.406Z [TRACE] consul.sync: must register service: id=_nomad-task-6de92b6e-9099-654a-00f2-93ac2b255142-group-xxxx-8080 exists=true reason=periodic
2023-09-20T10:04:07.717Z [DEBUG] consul.sync: sync complete: registered_services=1 deregistered_services=0 registered_checks=0 deregistered_checks=1
```

We found that this happen with all services that have upstreams configured. We found that the problem occurs [here](https://github.com/hashicorp/nomad/blob/fdc6c2151d65df72d16d946b368dba5e96a86a96/command/agent/consul/service_client.go#L344),  when in nomad the mesh gateway mode is left empty, consul sets it to local, so when checking the difference between the value in the job and in consul, nomad finds a difference and tries to register it again.

Related to this, in that service we modified the interval of the check, and saw that in the process of registering again the service, the old and new checks are both being registered (for a moment before removing again the old check in the sync proccess.

```
  "_nomad-check-01300b3da45ff5b95964e40301deb8058883665d": {
    "Node": "wrkr-ec26d795-8a28",
    "CheckID": "_nomad-check-01300b3da45ff5b95964e40301deb8058883665d",
    "Name": "xxxx-http",
    "Status": "passing",
    "Notes": "",
    "Output": "HTTP GET http://10.15.36.66:21795/_service/v0/health: 200  Output: {\n  \"status\" : \"UP\"\n}",
    "ServiceID": "_nomad-task-6de92b6e-9099-654a-00f2-93ac2b255142-group-xxxx-8080",
    "ServiceName": "xxxx",
    "ServiceTags": [],
    "Type": "http",
    "Interval": "30s",
    "Timeout": "5s",
    "ExposedPort": 0,
[...]

PR #18692, opened 2023-10-07
consul: prevent unnecessary periodic service sync

lgfa29:
The periodic Consul service sync routine computes a diff between the service registration from Nomad's point-of-view against Consul's.

This diff logic is done over strict equality of some string values that have defaults in Consul that Nomad may not be aware of. This commit fixes two of such cases.

* `DestinationType` is an optional field that defaults (and is written to Consul state) as `service`.
* `MeshGateway.Mode` may be an empty string, in which case the final value is defined by a hierarchy of possibilities, including service defaults and overrides that are not available to Nomad.

In both cases, Nomad's view is that these fields are empty strings, but in Consul an actual final value is written.

This difference results in constant unnecessary service sync from Nomad to Consul.

Other fields may also be affected in a similar way, but these are the two I found while investigating #18538

### DIFF
--- a/command/agent/consul/service_client_test.go
+++ b/command/agent/consul/service_client_test.go
@@ -578,6 +578,7 @@ func TestSyncLogic_proxyUpstreamsDifferent(t *testing.T) {
 		return api.Upstream{
 			Datacenter:       "sfo",
 			DestinationName:  "billing",
+			DestinationType:  "service",
 			LocalBindAddress: "127.0.0.1",
 			LocalBindPort:    5050,
 			MeshGateway: api.MeshGatewayConfig{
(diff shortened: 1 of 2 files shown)


jrasell (approved):
LGTM!

I think we need a changelog entry before merging.

command/agent/consul/service_client_test.go
```diff
@@ -578,6 +578,7 @@ func TestSyncLogic_proxyUpstreamsDifferent(t *testing.T) {
 		return api.Upstream{
 			Datacenter:       "sfo",
 			DestinationName:  "billing",
+			DestinationType:  "service",
```
shoenig, line 581: I think we should still have a test where the `DestinationType` is `""` on the Nomad side and non empty on the Consul side

shoenig (approved):

lgfa29:
Marking this as draft for now as @jorgemarey noticed that this can cause job changes to these fields to be ignored when updated from a value set to the value being unset.

Issue #16515, opened 2023-03-16
Max_client_disconnect ignores no-reschedule policy.


stswidwinski:
## Advertised contract.

The advertised contract is as follows:

> These allocations may continue to run on the disconnected client. Replacement allocations will be scheduled according to the allocations' reschedule policy until the disconnected client reconnects. Once a disconnected client reconnects, Nomad will compare the "unknown" allocations with their replacements and keep the one with the best node score. If the max_client_disconnect duration expires before the client reconnects, the allocations will be marked "lost". Clients that contain "unknown" allocations will transition to "disconnected" rather than "down" until the last max_client_disconnect duration has expired.

As described here: https://developer.hashicorp.com/nomad/docs/job-specification/group#max-client-disconnect.

However, we have observed that this contract is simply not obeyed and workloads are restarted immediately at the time of node becoming unavailable. This is due to the fact that the logic which handles evaluations for `node-updated` will cause placement of allocations which are *not* related to those which are "lost" (for the disconnected node).

## Root cause.

On node state update (which occurs when the node transitions from ready  to disconnected ):

https://github.com/hashicorp/nomad/blob/v1.5.1/nomad/node_endpoint.go#L479

Nomad will issue an evaluation:

https://github.com/hashicorp/nomad/blob/v1.5.1/nomad/node_endpoint.go#L577-L587

https://github.com/hashicorp/nomad/blob/v1.5.1/nomad/node_endpoint.go#L661-L667

This evaluation will add evaluations per-job as described here:

https://github.com/hashicorp/nomad/blob/v1.5.1/nomad/node_endpoint.go#L1593-L1595

Now, the evaluation computation is where the bug lives. The computation will go through the regular allocation reconciler here:

https://github.com/hashicorp/nomad/blob/v1.5.1/scheduler/generic_sched.go#L385

And eventually to computing the allocations for placement within computeDeploymentComplete  and computeGroup:

https://github.com/hashicorp/nomad/blob/v1.5.1/scheduler/reconcile.go#L434-L436

When we filter for disconnecting nodes we ignore the returned values of untainted . This is the crux of the bug since later on when we compute the placement of allocations we will get the following variables within:

https://github.com/hashicorp/nomad/blob/v1.5.1/scheduler/reconcile.go#L502

```
(dlv) print untainted
github.com/hashicorp/nomad/scheduler.allocSet []
(dlv) print migrate
github.com/hashicorp/nomad/scheduler.allocSet []
(dlv) print reschedule
[...]

PR #18701, opened 2023-10-09
fix: handling non reschedule disconnecting and reconnecting allocs

Juanadelacuesta:
This PR fixes  [Max_client_disconnect ignores no-reschedule policy](https://github.com/hashicorp/nomad/issues/16515)

 The bug had 2 root causes:

- The code was written under the assumption that disconnecting allocs should always be rescheduled:
https://github.com/hashicorp/nomad/blob/cecd9b04729e63839428cfa67fdc983ed2182481/scheduler/reconcile.go#L461

Any disconnecting allocation that should not be rescheduled now, was ignored for the rest of the reconcile process and would always end up on a new placement, because it was not taken into consideration for the deployment final count.  

- The logic to define if an allocation should be reschedule didn't take into account disconnecting allocations, they were 
never set to be explicitly replaced, because according to the [`shouldFilter`](https://github.com/hashicorp/nomad/blob/cecd9b04729e63839428cfa67fdc983ed2182481/scheduler/reconcile_util.go#L428) logic, they are `untainted` and are ignored on line 396, and never make it to the `elegibleNow` selection.
 ```
           isUntainted, ignore := shouldFilter(alloc, isBatch)
		if isUntainted && !isDisconnecting {
			untainted[alloc.ID] = alloc
		}
		if isUntainted || ignore {
			continue
		}

		// Only failed allocs with desired state run get to this point
		// If the failed alloc is not eligible for rescheduling now we
		// add it to the untainted set. Disconnecting delay evals are
		// handled by allocReconciler.createTimeoutLaterEvals
		eligibleNow, eligibleLater, rescheduleTime = updateByReschedulable(alloc, now, evalID, deployment, isDisconnecting)
		if !isDisconnecting && !eligibleNow {
```
[L396](https://github.com/hashicorp/nomad/blob/cecd9b04729e63839428cfa67fdc983ed2182481/scheduler/reconcile_util.go#L396)
Also on the `updateByReschedulable`, disonnecting allocs where never set to reschedule now:
```
	if eligible && (alloc.FollowupEvalID == evalID || rescheduleTime.Sub(now) <= rescheduleWindowSize) {
		rescheduleNow = true
		return
	}
```

This allocations would once again end up being replaced because the deployment count would be one short, not because it should be replaced and most definitely not taking into account its reschedule policy. 

Once the previous problem was fixed and the correct reschedule policy was being taken into account, a new bug was visible: The `disconnecting` allocations don't have a fail time, so to calculate the next reschedule time, the alloc reconciler `now`  was used. The minimun delay for batch jobs is 5 sec and 30 sec for services, following the logic in the code, the next reschedule time ends up always being at least 5 sec from now, and the disconnecting allocations were always set  to `rescheduleLater`, but for disconnecting allocations, the `rescheduleLater` group was ignored, as shown previously on line 461 of the rescheduler.

The function `filterByRescheduleable` was modified to correctly process disconnecting allocation and the disconnecting untainted  and reconnect later allocations were added to the result to avoid unnecessary placement and correctly replace the `disconnecting` allocs using the reschedule policy.

Once the previous change was done, new evaluations were created with the new reschedule time to replace the `disconnecting` allocs, but by the time the evaluation was executed, the allocs don't qualify as disconnecting anymore  so a new restriction needed to be added in the `updateByReschedule` function for alloc that were not disconnecting and had not failed yet, the `unknown` allocs.
In order to avoid adding replacements for this allocs every time the reconciler run, a validation using the `followupEvalID` was added. 
When taking into account the reschedule policy, every `disconnecting` alloc generates 2 new evaluations: one for the next reschedule time, in order to place a replacement and a second one to set the alloc as lost once the `max_client_disconnect` times out. 
The way the code is written, the disconnecting alloc would get updated with the evaluation ID of the second generated evaluation, making the `nextEvalID` useless to verify if the `unknown` alloc being rescheduled was in fact a replacement for an old one. A change to overwrite the  `nextEvalID` with the ID of the evaluation corresponding to the replacement was set in place. This change also made finding if an alloc was a replacement of an older one or not on reconnecting a lot easier.

### DIFF
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -10897,7 +10897,8 @@ func (a *Allocation) NextRescheduleTime() (time.Time, bool) {
 	failTime := a.LastEventTime()
 	reschedulePolicy := a.ReschedulePolicy()
 
-	if reschedulePolicy.Attempts == 0 && reschedulePolicy.Unlimited {
+	//If reschedule is disabled, return early
+	if reschedulePolicy.Attempts == 0 && !reschedulePolicy.Unlimited {
 		return time.Time{}, false
 	}
 
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -10914,16 +10914,16 @@ func (a *Allocation) nextRescheduleTime(failTime time.Time, reschedulePolicy *Re
 	return nextRescheduleTime, rescheduleEligible
 }
 
-// NextRescheduleTimeByFailTime works like NextRescheduleTime but allows callers
+// NextRescheduleTimeByTime works like NextRescheduleTime but allows callers
 // specify a failure time. Useful for things like determining whether to reschedule
 // an alloc on a disconnected node.
-func (a *Allocation) NextRescheduleTimeByFailTime(failTime time.Time) (time.Time, bool) {
+func (a *Allocation) NextRescheduleTimeByTime(t time.Time) (time.Time, bool) {
 	reschedulePolicy := a.ReschedulePolicy()
 	if reschedulePolicy == nil {
 		return time.Time{}, false
 	}
 
-	return a.nextRescheduleTime(failTime, reschedulePolicy)
+	return a.nextRescheduleTime(t, reschedulePolicy)
 }
 
 // ShouldClientStop tests an alloc for StopAfterClientDisconnect configuration
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -135,6 +135,7 @@ const (
 )
 
 const (
+
 	// SystemInitializationType is used for messages that initialize parts of
 	// the system, such as the state store. These messages are not included in
 	// the event stream.
@@ -10895,6 +10896,11 @@ func (a *Allocation) MigrateStrategy() *MigrateStrategy {
 func (a *Allocation) NextRescheduleTime() (time.Time, bool) {
 	failTime := a.LastEventTime()
 	reschedulePolicy := a.ReschedulePolicy()
+
+	if reschedulePolicy.Attempts == 0 && reschedulePolicy.Unlimited {
+		return time.Time{}, false
+	}
+
 	if a.DesiredStatus == AllocDesiredStatusStop || a.ClientStatus != AllocClientStatusFailed || failTime.IsZero() || reschedulePolicy == nil {
 		return time.Time{}, false
 	}
@@ -11223,7 +11229,7 @@ func (a *Allocation) Expired(now time.Time) bool {
 	}
 
 	expiry := lastUnknown.Add(*tg.MaxClientDisconnect)
-	return now.UTC().After(expiry) || now.UTC().Equal(expiry)
+	return expiry.Sub(now) <= 0
 }
 
 // LastUnknown returns the timestamp for the last time the allocation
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Review by tgross — CHANGES_REQUESTED** [review]
This code changes in this PR seem to make sense @Juanadelacuesta, but unfortunately it looks like it fixes this bug but introduces a new one.

I built Nomad from this branch for testing. I deployed a 2 node cluster and a job with the following relevant bits of the jobspec:

```hcl
job "httpd" {
  reschedule {
    delay          = "2m"
    delay_function = "exponential"
    max_delay      = "1h"
    unlimited      = true
  }

  group "web" {
    max_client_disconnect = "5m"

    # ... etc
  }
}
```

<details><summary>full job spec here</summary>

```hcl
job "httpd" {

  reschedule {
    delay          = "2m"
    delay_function = "exponential"
    max_delay      = "1h"
    unlimited      = true
  }

  group "web" {

    max_client_disconnect = "5m"

    network {
      mode = "bridge"
      port "www" {
        to = 8001
      }
    }

    task "http" {

      driver = "docker"

      config {
        image   = "busybox:1"
        command = "httpd"
        args    = ["-vv", "-f", "-p", "8001", "-h", "/local"]
        ports   = ["www"]
      }

      resources {
        cpu    = 100
        memory = 100
      }

    }
  }
}
```

</details>

First, I waited for the deployment to finish. After stopping the node the alloc was on and waiting for the missed heartbeat, the alloc is marked `unknown` as expected but it gets immediately rescheduled. It doesn't wait for the `delay` window:

```
$ nomad eval list
ID        Priority  Triggered By            Job ID  Namespace  Node ID   Status    Placement Failures
90de133f  50        max-disconnect-timeout  httpd   default    <none>    pending   false
f0b4cd16  50        node-update             httpd   default    d55ab857  complete  false
768e0dbe  50        deployment-watcher      httpd   default    <none>    complete  false
ef752be3  50        job-register            httpd   default    <none>    complete  false
```

```
$ nomad job status httpd
...
Allocations
ID        Node ID   Task Group  Version  Desired  Status   Created    Modified
7baca89e  8910b70c  web         0        run      running  33s ago    21s ago
7abbdab0  d55ab857  web         0        run      unknown  1m48s ago  33s ago
```

If I start over and I disable reschedules entirely, then the bug fix works as intended.

```hcl
job "httpd" {
  reschedule {
    delay          = "2m"
    delay_function = "exponential"
    max_delay      = "1h"
    unlimited      = false
  }

  group "web" {
    max_client_disconnect = "5m"

    # ... etc
  }
}
```

Then it's not immediately rescheduled and I end up seeing see the following evals in my list:

```
$ nomad eval list
ID        Priority  Triggered By            Job ID  Namespace  Node ID   Status    Placement Failures
38554c98  50        max-disconnect-timeout  httpd   default    <none>    pending   false
e8976a61  50        node-update             httpd   default    d55ab857  complete  false
89feb842  50        deployment-watcher      httpd   default    <none>    complete  false
5a25579f  50        job-register            httpd   default    <none>    complete  false
```

tgross (approved):
Nice work on this @Juanadelacuesta!

Code LGTM, and I've pulled it down for testing manually and everything checks out. :shipit:

PR #18703, opened 2023-10-09
auth: refactor `Authenticate` into its own package

tgross:
The RPC handlers expect to see `nil` ACL objects whenever ACLs are disabled. By using `nil` as a sentinel value, we have the risk of nil pointer exceptions and improper handling of `nil` when returned from our various auth methods that can lead to privilege escalation bugs.

This patchset is the first in a series to eliminate the use of `nil` ACLs as a sentinel value for when ACLs are disabled. This one is entirely refactoring to reduce the burden of reviewing the final patchsets that have the functional changes:

* Move RPC auth into a new `nomad/auth` package, injecting the dependencies required from the server. Expose only those public methods on `nomad/auth` that are intended for use in the RPC handlers.
* Keep the existing large authentication test as an integration test.
* Add unit tests covering the methods of `nomad/auth` we intend on keeping. The assertions for many of these will change once we have no `nil` sentinels and can make safe assertions about permissions on the resulting `ACL` objects.


nomad/csi_endpoint_test.go
```diff
@@ -1995,7 +1988,7 @@ func TestCSIVolume_nodeExpandVolume(t *testing.T) {
 
 func TestCSIPluginEndpoint_RegisterViaFingerprint(t *testing.T) {
 	ci.Parallel(t)
-	srv, shutdown := TestServer(t, func(c *Config) {
+	srv, _, shutdown := TestACLServer(t, func(c *Config) {
```
tgross, line 1991: Note to reviewers: I have no idea why we were changing ACL config after startup for these CSI tests, but you can't actually do that and it's broken to even try now, so that's the only reason why these tests are changing.

nomad/auth/auth.go
```diff
@@ -0,0 +1,483 @@
...
+		tlsEnabled:   tlsEnabled,
+		logger:       logger.With("auth"),
+		getState:     stateFn,
+		getLeaderACL: getLeaderACLFn,
+		region:       region,
+		aclCache:     structs.NewACLCache[*acl.ACL](aclCacheSize),
+		encrypter:    encrypter,
+	}
+}
+
```
tgross, line 69: Note for reviewers: the code from here down is essentially untouched from the existing `nomad/acl.go`. The only changes should be those that are needed to inject the dependencies from the `nomad` package.

nomad/auth/auth.go
```diff
@@ -0,0 +1,483 @@
...
+	// aclCache is used to maintain the parsed ACL objects
+	aclCache *structs.ACLCache[*acl.ACL]
+
+	// encrypter is a pointer to the server's Encrypter that can be used to
+	// verify claims
+	encrypter Encrypter
+}
+
+func NewAuthenticator(stateFn StateGetter, logger hclog.Logger, getLeaderACLFn LeaderACLGetter, aclsEnabled, tlsEnabled bool, region string, encrypter Encrypter,
```
jrasell, line 56: Would there be a benefit to wrapping the function parameters in a struct? I am thinking if we add any in the future, the function signature will get a little unwieldily.

tgross: Can't hurt. It's only instantiated once per Nomad server, so it's not like that allocation is in a hot path. Will do.

tgross: Done!

nomad/rpc.go
```diff
@@ -140,6 +147,34 @@ func (ctx *RPCContext) ValidateCertificateForName(name string) error {
 	return fmt.Errorf("invalid certificate, %s not in %s", name, strings.Join(validNames, ","))
 }
 
+func (ctx *RPCContext) IsStatic() bool {
+	return ctx == nil
+}
+
+func (ctx *RPCContext) GetRemoteIP() (net.IP, error) {
+	var remoteAddr *net.TCPAddr
+	var ok bool
```
jrasell, line 156: I don't know how often `ctx` will be nil, but if it's a decent amount, is it worth moving these past that check?

jrasell (approved):
LGTM!

nomad/rpc.go
tgross: `ctx` is only ever nil when we're sending "static" RPCs. That is, RPCs that are always sent as (expensive) method calls rather than being sent over the wire. An example of that is the deployment watcher, which can use that for all its local RPCs because it only runs on the leader.

So that's probably a good chunk. Can't hurt to move it.

nomad/auth/auth.go
```diff
@@ -0,0 +1,492 @@
...
+package auth
+
+import (
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"net"
+	"time"
+
+	metrics "github.com/armon/go-metrics"
```
shoenig, line 13: ultra nit: the package alias is redundant

nomad/rpc.go
```diff
@@ -140,6 +147,34 @@ func (ctx *RPCContext) ValidateCertificateForName(name string) error {
 	return fmt.Errorf("invalid certificate, %s not in %s", name, strings.Join(validNames, ","))
 }
 
+func (ctx *RPCContext) IsStatic() bool {
+	return ctx == nil
+}
+
+func (ctx *RPCContext) GetRemoteIP() (net.IP, error) {
+	if ctx == nil {
+		return nil, nil
```
lgfa29, line 156: Is this an expected outcome? Would it make sense to return an error instead of a potentially dangerous `nil, nil`?

shoenig (approved):
LGTM!

lgfa29 (approved):
Minor thought, but LGTM!

nomad/rpc.go
tgross: Yes, we'll see this a lot! The `RPCContext` is nil whenever we have "static" RPCs, which are the RPCs sent as normal (albeit more expensive) method calls from components internal to the process. So for example, the deploymentwatcher only makes RPCs internal to the leader and doesn't go over wire (except for MRD), so there's no "remote IP" to get for that case.

nomad/auth/auth.go
tgross: Oops, sure is! See, I really just copy-and-pasted this from the original code with the minimum changes! :grinning:

nomad/rpc.go
lgfa29: Ah got it, thanks!

PR #18715, opened 2023-10-10
auth: add server-only ACL

tgross:
The RPC handlers expect to see `nil` ACL objects whenever ACLs are disabled. By using `nil` as a sentinel value, we have the risk of nil pointer exceptions and improper handling of `nil` when returned from our various auth methods that can lead to privilege escalation bugs. This is the second in a series to eliminate the use of `nil` ACLs as a sentinel value for when ACLs are disabled.

This patch involves creating a new "virtual" ACL object for checking permissions on server operations and a matching `AuthenticateServerOnly` method for server-only RPCs that can produce that ACL.

### DIFF
--- a/acl/acl.go
+++ b/acl/acl.go
@@ -76,8 +76,8 @@ type ACL struct {
 	quota    string
 	plugin   string
 
-	// The attribute below is avirtual policy that we never expose directly to
-	// the end user
+	// The attributes below detail a virtual policy that we never expose
+	// directly to the end user.
 	server   string
 	isLeader bool
 }
--- a/acl/acl.go
+++ b/acl/acl.go
@@ -75,6 +75,11 @@ type ACL struct {
 	operator string
 	quota    string
 	plugin   string
+
+	// The attribute below is avirtual policy that we never expose directly to
+	// the end user
+	server   string
+	isLeader bool
 }
 
 // maxPrivilege returns the policy which grants the most privilege
@@ -295,6 +300,9 @@ func NewACL(management bool, policies []*Policy) (*ACL, error) {
 
 	acl.variables = svTxn.Commit()
 	acl.wildcardVariables = wsvTxn.Commit()
+
+	acl.server = PolicyDeny
+	acl.isLeader = false
 
 	return acl, nil
 }
@@ -723,6 +731,11 @@ func (a *ACL) AllowNodeRead() bool {
 		return true
 	case a.node == PolicyRead:
 		return true
+	case a.server == PolicyRead,
+		a.server == PolicyWrite:
+		return true
+	case a.isLeader:
+		return true
 	default:
 		return false
 	}
@@ -822,6 +835,16 @@ func (a *ACL) AllowPluginList() bool {
 	default:
 		return false
 	}
+}
+
+// AllowServerOp checks if server-only operations are allowed
+func (a *ACL) AllowServerOp() bool {
+	if a == nil {
+		// ACL is nil only if ACLs are disabled
+		// TODO(tgross): return false when there are no nil ACLs
+		return true
+	}
+	return a.server != PolicyDeny || a.isLeader
 }
 
 // IsManagement checks if this represents a management token
--- a/nomad/auth/auth.go
+++ b/nomad/auth/auth.go
@@ -256,7 +256,8 @@ func validateCertificateForName(cert *x509.Certificate, expectedName string) (bo
 		return false, nil
 	}
 
-	validNames := append(cert.DNSNames, cert.Subject.CommonName)
+	validNames := []string{cert.Subject.CommonName}
+	validNames = append(validNames, cert.DNSNames...)
 	for _, valid := range validNames {
 		if expectedName == valid {
 			return true, nil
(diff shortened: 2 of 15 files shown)


acl/acl.go
```diff
@@ -75,6 +75,11 @@ type ACL struct {
 	operator string
 	quota    string
 	plugin   string
+
+	// The attribute below is avirtual policy that we never expose directly to
+	// the end user
```
jrasell, line 80:
```suggestion
	// The attributes below detail a virtual policy that we never expose
	// directly to the end user.
```

nomad/auth/auth.go
```diff
@@ -204,6 +205,69 @@ func (s *Authenticator) ResolveACL(args structs.RequestWithIdentity) (*acl.ACL,
 	return nil, nil
 }
 
+// AuthenticateServerOnly returns an ACL object for use *only* with internal
+// server-to-server RPCs. This should never be used for RPCs that serve HTTP
+// endpoints or accept ACL tokens to avoid confused deputy attacks by making a
+// request to a follower that's forwarded.
+//
+// The returned ACL object is always a acl.ServerACL but in the future this
```
jrasell, line 213:
```suggestion
// The returned ACL object is always an acl.ServerACL but in the future this
```

nomad/auth/auth.go
```diff
@@ -204,6 +205,69 @@ func (s *Authenticator) ResolveACL(args structs.RequestWithIdentity) (*acl.ACL,
...
+		tlsCert := ctx.Certificate()
+		if tlsCert == nil {
+			return nil, errors.New("missing certificate information")
+		}
+
+		// set on the identity whether or not its valid for server RPC, so we
+		// can capture it for metrics
+		identity.TLSName = tlsCert.Subject.CommonName
+
+		expected := fmt.Sprintf("server.%s.nomad", s.region)
```
jrasell, line 236: Could save ourselves a few nano-seconds:
```suggestion
		expected := "server." + s.region + ".nomad"
```

nomad/auth/auth.go
```diff
@@ -204,6 +205,69 @@ func (s *Authenticator) ResolveACL(args structs.RequestWithIdentity) (*acl.ACL,
...
+
+// validateCertificateForName returns true if the certificate is valid
+// for the given domain name.
+func validateCertificateForName(cert *x509.Certificate, expectedName string) (bool, error) {
+	if cert == nil {
+		return false, nil
+	}
+
+	validNames := []string{cert.Subject.CommonName}
+	validNames = append(validNames, cert.DNSNames...)
```
jrasell, line 260: This could be done in a single line, but I am not sure if I am unaware of a problem with doing it like this.
```suggestion
	validNames := append(cert.DNSNames, cert.Subject.CommonName)
```

acl/acl.go
```diff
@@ -75,6 +75,11 @@ type ACL struct {
...
+	server   string
+	isLeader bool
```
jrasell, line 82: Just checking my logic that currently this value will always be false, and this is expected?

jrasell (approved):
LGTM!

Just a couple of random thoughts or questions that can easily be ignored.

nomad/auth/auth.go
tgross: Oops, sure can.

tgross: LOL, apparently go-critic hated that though:

```
Error: nomad/auth/auth.go:259:16: appendAssign: append result not assigned to the same slice (gocritic)
	validNames := append(cert.DNSNames, cert.Subject.CommonName)
	              ^
```

nomad/keyring_endpoint.go
```diff
@@ -112,16 +112,10 @@ func (k *Keyring) List(args *structs.KeyringListRootKeyMetaRequest, reply *struc
...
-		if aclObj, err := k.srv.ResolveACL(args); err != nil {
-			return err
-		} else if aclObj != nil && !aclObj.IsManagement() {
-			return structs.ErrPermissionDenied
-		}
+	if aclObj, err := k.srv.ResolveACL(args); err != nil {
+		return err
+	} else if aclObj != nil && !aclObj.IsManagement() {
+		return structs.ErrPermissionDenied
 	}
```
lgfa29, line 119: When servers call this endpoint do they pass an `aclObj`?

lgfa29 (approved):
Just some minor questions, but nice refactoring!

nomad/auth/auth.go
```diff
@@ -204,6 +205,69 @@ func (s *Authenticator) ResolveACL(args structs.RequestWithIdentity) (*acl.ACL,
...
+	for _, valid := range validNames {
+		if expectedName == valid {
+			return true, nil
+		}
+	}
+
+	return false, fmt.Errorf("invalid certificate, %s not in %s",
+		expectedName, strings.Join(validNames, ","))
+}
```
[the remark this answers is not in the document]
tgross: You're seeing a bit of an artifact of breaking this work into multiple PRs. We only need the `(*RPCContext).ValidateCertificateForName` method to exist because we haven't done the "auth: add client-only ACL" PR. 

When we've completed that, the only callers for this logic will exist encapsulated in the `nomad/auth` package so there's no reason to keep it in `nomad/rpc.go` anymore. That allows us to have all the code that tests it in the `nomad/auth` package without having to copy the logic into a `ValidateCertificateForName` method on the `testContext` struct (at which point we'd be testing the mock context logic and not the production code).

nomad/keyring_endpoint.go
tgross: As it turns out, servers don't call this `Keyring.List` endpoint at all! I think we originally had this to support keyring replication between leader and followers, but the followers just read the metadata from the state store and then `Keyring.Get` the actually key material.

PR #18725, opened 2023-10-11
deps: remove Vault SDK

tgross:
Nomad imports the Vault SDK to get testing helpers, but it turns out the only thing actually in use was a single string constant for the Vault namespace header. Remove this dependency and hardcode the constant to reduce dependency churn.


jrasell (approved):
LGTM!

lgfa29 (approved):
🧹

PR #18730, opened 2023-10-11
auth: add client-only ACL

tgross:
The RPC handlers expect to see `nil` ACL objects whenever ACLs are disabled. By using `nil` as a sentinel value, we have the risk of nil pointer exceptions and improper handling of `nil` when returned from our various auth methods that can lead to privilege escalation bugs. This is the third in a series to eliminate the use of `nil` ACLs as a sentinel value for when ACLs are disabled.

This patch involves creating a new "virtual" ACL object for checking permissions on client operations and a matching `AuthenticateClientOnly` method for client-only RPCs that can produce that ACL.

Unlike the server ACLs PR, this also includes a special case for "legacy" client RPCs where the client was not previously sending the secret as it should (leaning on mTLS only). Those client RPCs were fixed in Nomad 1.6.0, but it'll take a while before we can guarantee they'll be present during upgrades.


nomad/auth/auth.go
```diff
@@ -249,6 +249,103 @@ func (s *Authenticator) AuthenticateServerOnly(ctx RPCContext, args structs.Requ
...
+//
+// The returned ACL object is always a acl.ClientACL but in the future this
+// could be extended to allow clients access only to their own pool and
+// associated namespaces, etc.
+func (s *Authenticator) AuthenticateClientOnly(ctx RPCContext, args structs.RequestWithIdentity) (*acl.ACL, error) {
+
+	remoteIP, err := ctx.GetRemoteIP() // capture for metrics
+	if err != nil {
+		s.logger.Error("could not determine remote address", "error", err)
+	}
```
shoenig, line 266: just making sure we do not want to bail here?

nomad/auth/auth.go
```diff
@@ -249,6 +249,103 @@ func (s *Authenticator) AuthenticateServerOnly(ctx RPCContext, args structs.Requ
...
+
+		expected := fmt.Sprintf("client.%s.nomad", s.region)
+		_, err := validateCertificateForName(tlsCert, expected)
+		if err != nil {
+			expected := fmt.Sprintf("server.%s.nomad", s.region)
+			_, err := validateCertificateForName(tlsCert, expected)
+			if err != nil {
+				return nil, err
+			}
+		}
```
shoenig, line 289: would it make sense to have `validateCertificateForName` accept a slice of expected values of which at least one must match?

nomad/service_registration_endpoint.go
```diff
@@ -381,12 +339,12 @@ func (s *ServiceRegistration) GetService(
...
+	aclObj, err := s.srv.ResolveACL(args)
 	if err != nil {
 		return structs.ErrPermissionDenied
 	}
-	if args.GetIdentity().Claims == nil &&
-		!aclObj.AllowNsOp(args.RequestNamespace(), acl.NamespaceCapabilityReadJob) {
+	if aclObj != nil && !aclObj.AllowServiceRegistrationReadList(args.RequestNamespace(),
+		args.GetIdentity().Claims != nil) {
 		return structs.ErrPermissionDenied
 	}
```
shoenig, line 349: didn't we want to avoid allowing `aclObj == nil` (or is that coming later)?

shoenig (approved):

nomad/service_registration_endpoint.go
tgross: The `ACLsDisabled` object is coming later, but you're right we don't need this here now because `AllowServiceRegistrationReadList` handles that correctly (with a TODO to reverse the test for a nil ACL once we do have the `ACLsDisabled`. Will fix.

nomad/auth/auth.go
tgross: That works. Will do.

nomad/service_registration_endpoint.go
tgross: Done in 9701aca

nomad/auth/auth.go
tgross: Done in b07a322. While doing that I noticed we had a straggler use of the cert check and after removing that we can remove the old cert check code in `nomad/util.go` entirely. Done in the same commit.

tgross: And then _that_ revealed that there was a bunch of places where we weren't authenticating with the node secret in test code, even though it's required in the RPC. Fixed that.

PR #18731, opened 2023-10-11
testing: fix configuration for retry tests

tgross:
The retry tests in the `api` package set up a client but don't use `NewClient`, so the address never gets parsed into a `url.URL` and that's causing some test failures.


angrycub (approved):
Looks good!

PR #18732, opened 2023-10-11
consul hook: get WIs only for own task group

tgross:
The WID manager will only sign WI tokens for the allocation's task group. We're accidentally looping over all the task groups, which for jobs with multiple task groups results in a failure in the `consul_hook`.


lgfa29 (approved):
Nice catch. Is there any test that should've caught this?

tgross:
> Nice catch. Is there any test that should've caught this?

Unfortunately the `consul_hook` is lacking in tests, so I'm going to tackle that tomorrow as my most urgent task.

PR #18740, opened 2023-10-12
vault: use an importable const for Vault header string.

jrasell:
Enterprise also requires this, therefore setting up a const that can be used within both codebases. Once merged, I will raise a PR for the enterprise side, and balance the merge processes.


tgross (approved):
LGTM!

pkazmierczak (approved):

PR #18742, opened 2023-10-12
WI: set the right identity name for Consul tasks

pkazmierczak:
Consul tasks should only have 1 identity of the form `consul/{consul_cluster_name}`.

### DIFF
--- a/client/allocrunner/consul_hook.go
+++ b/client/allocrunner/consul_hook.go
@@ -12,7 +12,6 @@ import (
 	"github.com/hashicorp/nomad/client/consul"
 	cstructs "github.com/hashicorp/nomad/client/structs"
 	"github.com/hashicorp/nomad/client/widmgr"
-	"github.com/hashicorp/nomad/nomad"
 	"github.com/hashicorp/nomad/nomad/structs"
 	structsc "github.com/hashicorp/nomad/nomad/structs/config"
 )
@@ -108,7 +107,7 @@ func (h *consulHook) prepareConsulTokensForTask(job *structs.Job, task *structs.
 	// get tokens for alt identities for Consul
 	mErr := multierror.Error{}
 	for _, i := range task.Identities {
-		if i.Name != fmt.Sprintf("%s/%s", nomad.ConsulTaskIdentityNamePrefix, task.GetConsulTaskName()) {
+		if i.Name != fmt.Sprintf("%s/%s", structs.ConsulTaskIdentityNamePrefix, task.GetConsulTaskName()) {
 			continue
 		}
 
--- a/client/allocrunner/consul_hook.go
+++ b/client/allocrunner/consul_hook.go
@@ -116,7 +116,7 @@ func (h *consulHook) prepareConsulTokensForTask(job *structs.Job, task *structs.
 	// get tokens for alt identities for Consul
 	mErr := multierror.Error{}
 	for _, i := range task.Identities {
-		if i.Name != fmt.Sprintf("%s/%s", structs.ConsulTaskIdentityNamePrefix, task.GetConsulTaskName()) {
+		if i.Name != fmt.Sprintf("%s_%s", structs.ConsulTaskIdentityNamePrefix, task.GetConsulTaskName()) {
 			continue
 		}
 
--- a/client/allocrunner/consul_hook.go
+++ b/client/allocrunner/consul_hook.go
@@ -116,7 +116,7 @@ func (h *consulHook) prepareConsulTokensForTask(job *structs.Job, task *structs.
 	// get tokens for alt identities for Consul
 	mErr := multierror.Error{}
 	for _, i := range task.Identities {
-		if i.Name != fmt.Sprintf("%s_%s", structs.ConsulTaskIdentityNamePrefix, task.GetConsulTaskName()) {
+		if i.Name != fmt.Sprintf("%s_%s", structs.ConsulTaskIdentityNamePrefix, consulClusterName) {
 			continue
 		}
 
--- a/client/allocrunner/consul_hook.go
+++ b/client/allocrunner/consul_hook.go
@@ -12,6 +12,7 @@ import (
 	"github.com/hashicorp/nomad/client/consul"
 	cstructs "github.com/hashicorp/nomad/client/structs"
 	"github.com/hashicorp/nomad/client/widmgr"
+	"github.com/hashicorp/nomad/nomad"
 	"github.com/hashicorp/nomad/nomad/structs"
 	structsc "github.com/hashicorp/nomad/nomad/structs/config"
 )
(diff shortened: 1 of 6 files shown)


client/allocrunner/consul_hook.go
```diff
@@ -99,18 +100,15 @@ func (h *consulHook) Prerun() error {
...
 		return nil
 	}
 
-	expectedIdentity := task.MakeUniqueIdentityName(tgName)
-
 	// get tokens for alt identities for Consul
 	mErr := multierror.Error{}
 	for _, i := range task.Identities {
-		if i.Name != expectedIdentity {
+		if i.Name != fmt.Sprintf("%s/%s", nomad.ConsulTaskIdentityNamePrefix, task.GetConsulTaskName()) {
```
lgfa29, line 111:
```suggestion
		if i.Name != fmt.Sprintf("%s_%s", nomad.ConsulTaskIdentityNamePrefix, task.GetConsulTaskName()) {
```

I remembered why I switched to not use `/` in the name. We write the signed identity to disk and the name is added to the path. A `/` would cause it to be a subdirectory in Unix systems:
https://github.com/hashicorp/nomad/blob/cecd9b04729e63839428cfa67fdc983ed2182481/nomad/structs/workload_id.go#L42-L44

nomad/job_endpoint_hook_implicit_identities.go
```diff
@@ -10,8 +10,8 @@ import (
 )
 
 const (
-	consulServiceIdentityNamePrefix = "consul-service"
-	consulTaskIdentityNamePrefix    = "consul"
+	ConsulServiceIdentityNamePrefix = "consul-service"
+	ConsulTaskIdentityNamePrefix    = "consul"
```
lgfa29, line 14: Since these values are shared across many different places I placed the Vault prefix constant in the `structs` package:
https://github.com/hashicorp/nomad/blob/cecd9b04729e63839428cfa67fdc983ed2182481/nomad/structs/workload_id.go#L23-L25

nomad/job_endpoint_hook_implicit_identities.go
```diff
@@ -87,7 +87,7 @@ func (h jobImplicitIdentitiesHook) handleConsulTasks(t *structs.Task, taskGroup
 	}
 
 	name := t.MakeUniqueIdentityName(taskGroup)
-	widName := fmt.Sprintf("%s/%s", consulTaskIdentityNamePrefix, name)
+	widName := fmt.Sprintf("%s/%s", ConsulTaskIdentityNamePrefix, name)
```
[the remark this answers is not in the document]
pkazmierczak: > Should this be ` fmt.Sprintf("%s/%s", nomad.ConsulTaskIdentityNamePrefix, task.GetConsulTaskName())` to match the client-side code?

yes, good catch

> 
> And if so, are there any callers of `(*Task) MakeUniqueIdentityName` left?

There aren't, I'll remove it.

lgfa29: Instead of `name` we need to use the `consul.cluster` value.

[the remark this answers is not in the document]
pkazmierczak: [10903ac](https://github.com/hashicorp/nomad/pull/18742/commits/10903ac0648c6b3790c04c47d829b9c58894c9cb)

pkazmierczak:
ah apologies, I didn't notice
```
	imports github.com/hashicorp/nomad/nomad: import cycle not allowed in test
```
I'll fix the import cycle asap.

nomad/structs/structs.go
```diff
@@ -7672,6 +7672,11 @@ func (t *Task) MakeUniqueIdentityName(taskGroup string) string {
 	return fmt.Sprintf("%v-%v", taskGroup, t.Name)
 }
 
+// GetConsulTaskName returns the consulClusterName_TaskName string.
+func (t *Task) GetConsulTaskName() string {
+	return fmt.Sprintf("%s_%s", t.Consul.Cluster, t.Name)
+}
+
```
lgfa29: We don't need the task name here, just the cluster. 

I think this got confusing in our discussion because at some point I used a method in my example. The reason I did that was because the `consul` block could be either at the group or task level, so we would need a function to pick the right cluster value. 

But I think @tgross mentioned he would work on that? For this PR I think you can just assume `task.Consul.Cluster` has what you need. Later we could change it to something `task.GetConsul().Cluster`, where `GetConsul()` would handle resolving the expected Consul config.

lgfa29 (changes requested):
We need to use `_` instead of `/` because of file paths 😅

nomad/structs/structs.go
pkazmierczak: But that would mean users could only have 1 templates Consul token per jobspec?

lgfa29: "But that would mean users could only have 1 templates Consul token per ~jobspec~ task?"

This function would be used in the context of a specific task, and in this context we can only have one identity called `consul_<cluster>`.

Look at it from the jobspec perspective:

```hcl
job "..." {
  group"a" {
    task "1" {
      identity {
        name = "consul_default"
      }
    }

    task "2" {
      identity {
        name = "consul_default"
      }
    }

  group "b" {
    task "1" {
      identity {
        name = "consul_default"
      }
    }

    task "2" {
      identity {
        name = "consul_default"
      }
    }

    task "3" {
      identity {
        name = "consul_default"
      }
    }
  }
}
```

Despite all tasks having an identity called `consul_default` they are all different because they are placed in different tasks.

client/allocrunner/consul_hook.go
```diff
@@ -98,19 +98,25 @@ func (h *consulHook) Prerun() error {
 }
 
 func (h *consulHook) prepareConsulTokensForTask(job *structs.Job, task *structs.Task, tgName string, tokens map[string]map[string]string) error {
+	var consulClusterName string
+	if task.Consul != nil && task.Consul.Cluster != "" {
+		consulClusterName = task.Consul.Cluster
+	} else {
+		consulClusterName = structs.ConsulDefaultCluster
+	}
```
lgfa29, line 106: I think this should be handled way earlier than in the alloc runner. For Vault it's set on job registration:
https://github.com/hashicorp/nomad/blob/3633ca0f8ca4c0a7d4ecb43840e1240c10f45116/nomad/job_endpoint_hook_vault_ce.go#L44-L55

But it could be handled in a separate PR as it requires some ENT changes.

nomad/job_endpoint_hook_implicit_identities.go
```diff
@@ -75,19 +70,25 @@ func (h jobImplicitIdentitiesHook) handleConsulService(s *structs.Service) {
...
-	name := t.MakeUniqueIdentityName(taskGroup)
-	widName := fmt.Sprintf("%s/%s", consulTaskIdentityNamePrefix, name)
+	var clusterName string
+	if t.Consul != nil && t.Consul.Cluster != "" {
+		clusterName = t.Consul.Cluster
+	} else {
+		clusterName = structs.ConsulDefaultCluster
+	}
+
+	widName := fmt.Sprintf("%s_%s", structs.ConsulTaskIdentityNamePrefix, clusterName)
```
lgfa29, line 91: I have this in a helper method to avoid repeating the logic here and in the alloc/task runner
https://github.com/hashicorp/nomad/blob/3633ca0f8ca4c0a7d4ecb43840e1240c10f45116/nomad/structs/structs.go#L10000-L10004

lgfa29 (approved):

PR #18744, opened 2023-10-12
auth: remove "mixed auth" special casing for Variables endpoint

tgross:
The RPC handlers expect to see `nil` ACL objects whenever ACLs are disabled. By using `nil` as a sentinel value, we have the risk of nil pointer exceptions and improper handling of `nil` when returned from our various auth methods that can lead to privilege escalation bugs. This is the fourth in a series to eliminate the use of `nil` ACLs as a sentinel value for when ACLs are disabled.

This patch involves leveraging the refactored `auth` package to remove the weird "mixed auth" helper functions that only support the Variables read/list RPC handlers. Instead, pass the ACL object and claim together into the `AllowVariableOperations` method in the usual `acl` package.

Ref: https://github.com/hashicorp/nomad/pull/18730

Fixes: https://github.com/hashicorp/nomad/issues/15875


nomad/acl_test.go
```diff
@@ -686,120 +686,3 @@ func TestResolveSecretToken(t *testing.T) {
 		})
 	}
 }
-
-func TestResolveClaims(t *testing.T) {
```
tgross, line 690: Note to reviewers: it looks like I'm throwing out a big bunch of tests here, but this was already moved to `nomad/auth/auth_test.go` in the 1st patch in this series, and I should have removed it then. :grinning:

nomad/auth/auth.go
```diff
@@ -194,24 +194,37 @@ func (s *Authenticator) Authenticate(ctx RPCContext, args structs.RequestWithIde
 	return nil
 }
 
-// ResolveACL is an authentication wrapper which handles resolving both ACL
-// tokens and Workload Identities. If both are provided the ACL token is
-// preferred, but it is best for the RPC caller to only include the credentials
-// for the identity they intend the operation to be performed with.
+// ResolveACL is an authentication wrapper which handles resolving ACL tokens,
+// Workload Identities, or client secrets into acl.ACL objects. Exclusively
+// server-to-server or client-to-server requests should by using
```
shoenig, line 199: s/by/be

nomad/auth/auth.go
```diff
@@ -363,16 +376,35 @@ func validateCertificateForNames(cert *x509.Certificate, expectedNames []string)
 
 }
 
-// ResolveACLForToken resolves an ACL from a token only. It should be used only
+// ToACLClaim returns an ACLClaim suitable for checking permissions
+func IdentityToACLClaim(ai *structs.AuthenticatedIdentity, store *state.StateStore) *acl.ACLClaim {
```
shoenig, line 380: comment name should match function name

nomad/auth/auth_test.go
```diff
@@ -840,6 +841,57 @@ func TestResolveACLToken(t *testing.T) {
...
 }
 
+func TestIdentityToACLClaim(t *testing.T) {
+
+	alloc := mock.Alloc()
+	alloc.ClientStatus = structs.AllocClientStatusRunning
+	tg := alloc.Job.LookupTaskGroup(alloc.TaskGroup)
+	task := tg.Tasks[0]
+
+	//	claims := alloc.ToTaskIdentityClaims(alloc.Job, "web")
```
shoenig, line 851: no longer needed?

shoenig (approved):
LGTM! just comment nitpicks

PR #18745, opened 2023-10-12
Add `go-netaddrs` support to `retry_join`

acaloiaro:
This PR adds auto discovery support for `go-netaddrs`, supplementing `go-discover`.

Needs:
- [x] Feedback from the Hashicorp Nomad team
- [x] Documentation updates throughout the codebase/website 
- [x] Perhaps additional test coverage over the fallback mechanism to `go-discover`

This PR is intended to be 100% backward compatible with existing `retry_join` configurations. Any deviation from that intent is a bug.


command/agent/retry_join_test.go
```diff
@@ -170,6 +171,44 @@ func TestRetryJoin_Server_MixedProvider(t *testing.T) {
...
+		errCh:         make(chan struct{}),
+	}
+
+	joiner.RetryJoin(serverJoin)
+
+	require.Equal(6, len(output))
+	require.Equal("192.168.1.1", output[0])
+	require.Equal("192.168.1.2", output[1])
+	require.Equal("::1", output[2])
+	require.Equal("127.0.0.1", output[3])
```
acaloiaro, line 207: These assertions on `::1` and `127.0.0.1` are system-dependent and will likely need to be refactored. In order to ensure that `go-netaddrs` resolves hostnames to IP addresses, a stable DNS entry must be in place. Currently, this relies on `/etc/hosts` to have entries for `::1` and `127.0.0.1` corresponding with the name `localhost`.

command/agent/retry_join_test.go
```diff
@@ -170,6 +171,44 @@ func TestRetryJoin_Server_MixedProvider(t *testing.T) {
...
+	require.Equal("100.100.100.100", output[4])
+	require.Equal("provider=aws, tag_value=foo", output[5])
```
acaloiaro, line 209: This doesn't actually assert that `go-discover` handles configurations of the form `provider=...`. An assertion will need to be made to ensure that configurations of that form reach `go-discover` as expected.

command/agent/retry_join.go
```diff
@@ -30,6 +35,57 @@ type DiscoverInterface interface {
...
+//
+// if addresses cannot be looked up using DNS or an exec command, fall back to
+// discovering addresses with go-discover
+func (d discoverProxy) Addrs(cfg string, logger *golog.Logger) (addrs []string, err error) {
+	var ipAddrs []net.IPAddr
+	ipAddrs, err = netaddrs.IPAddrs(context.Background(),
+		cfg,
+		log.FromStandardLogger(logger, &log.LoggerOptions{Name: "joiner"}))
+	if err != nil {
+		logger.Print("falling back to go-discover for auto-join config: ", cfg, "due to:", err)
```
acaloiaro, line 70: Most cloud auto-join configs will be for `go-discover` for quite some time. We probably won't want to log to standard out here since this log might imply that something is wrong, even though this is expected behavior. At the very least, we probably don't want to log the error here. 

Alternatively, maybe `Addrs` should take a logger with log levels so this can use a `Debug()` log instead.

command/agent/retry_join.go
```diff
@@ -114,17 +169,14 @@ func (r *retryJoiner) RetryJoin(serverJoin *ServerJoin) {
...
-				servers, err := r.discover.Addrs(addr, standardLogger)
-				if err != nil {
-					r.logger.Error("determining join addresses failed", "error", err)
-				} else {
-					addrs = append(addrs, servers...)
-				}
-			default:
+			servers, err := r.discover.Addrs(addr, standardLogger)
+			if err != nil {
+				r.logger.Warn("discovering join addresses failed", "join_config", addr, "error", err)
```
acaloiaro, line 174: This is also expected behavior and probably should not be a warn level log. This is Debug-worth at best, and is perhaps not necessary. It is here for debugging during development.

command/agent/retry_join.go
```diff
@@ -30,6 +35,57 @@ type DiscoverInterface interface {
 	Names() []string
 }
 
+// discoverProxy proxies retryJoiner discovery requests to either go-netaddrs or go-discover
+//
+// Addresses are first discovered using go-netaddrs, but if no addresses are able to be disocovered,
+// fall back to go-autodiscover
+type discoverProxy struct {
```
acaloiaro, line 42: `autoDiscover` may be a better name; `*Proxy` feels a bit non-specific.

command/agent/retry_join.go
acaloiaro: Additionally, nothing has necessarily "failed" at this point if `cfg` is an IP addresses, which is valid, then this error block simply means that "discovery" has yielded no results, not that `cfg` is some how invalid.

tgross: Given we're pushing some of this block's existing behavior down into the `Addrs` method, this feels like an opportunity to be more thoughtful about what we return as an error here. I would expect to see an error if:
* The discovery config is invalid.
* The discovery config is valid but behaved incorrectly. For example, when the `go-netaddrs` `exec=` fails or the `provider=` credentials are bad.

command/agent/retry_join.go
tgross: Agreed.

command/agent/retry_join.go
```diff
@@ -30,6 +35,57 @@ type DiscoverInterface interface {
...
+	goDiscover *discover.Discover
+}
+
+// Addrs looks up and returns IP addresses using the method described by cfg.
+//
+// If cfg is a DNS name, IP addresses are looked up by querying the default
+// DNS resolver for A and AAAA records associated with the DNS name.
```
tgross, line 49: This is a little misleading, so I want to call it out so it doesn't get copied into the eventual docs. The `go-netaddrs` library looks for the `exec=` prefix first and then if its not present, tries to resolve the name as if were a DNS entry.

command/agent/retry_join_test.go
tgross: The best approach is to not try to test the `go-netaddrs` and `go-discovery` libraries here by testing only the logic that's in Nomad. That might require a little bit of refactoring to make work.

tgross (changes requested):
Hi @acaloiaro! Thanks for this PR! I've left some suggestions on some of the details.

From a design standpoint, as you've got it here, everyone using a `provider=` configuration is going to make a spurious DNS query and have to wait for the NXDOMAIN. That's going to leak the provider config to whatever is answering DNS. For some providers that includes credentials (unfortunately!), so that seems problematic. Given that there's a small subset of correct prefixes here (`exec=`, `provider=`, valid DNS name), maybe it'd make more sense to switch behaviors based on the prefix rather than waiting on an error.

An even better approach might be to _require_ a `dns=` prefix that Nomad strips off before handing it to `go-netaddrs` if you want DNS lookups. That would make for a lot less "magic" behavior.

acaloiaro:
Thanks for the feedback, Tim. I'll implement your recommendations as time permits. Would you like me to ***@***.***` you when it's ready for a re-review, or would you like to get to it at your own pace?

tgross:
> Thanks for the feedback, Tim. I'll implement your recommendations as time permits. Would you like me to ***@***.***` you when it's ready for a re-review, or would you like to get to it at your own pace?

Go ahead and `@`-mention me or re-request the review. I get PR notifications on commits but that way I can hold off reviewing until you're ready. :grin:

command/agent/retry_join.go
acaloiaro: >  or the `provider=` credentials are bad.

Could you clarify what "bad" credentials are and how one might verify badness?

acaloiaro:
Hi @tgross, this PR is ready for re-review.

I believe that I've implemented all the changes for which I've resolved conversations, though I have one remaining question for you: https://github.com/hashicorp/nomad/pull/18745#discussion_r1383762556

I realize now that I probably should have asked you how you prefer to handle conversations/resolutions during CR. I think it would have been more sensible to leave those conversations unresolved until you verify that the actual implementation is to your liking. 


There is still one glaring piece missing from this PR:

- [ ] Documentation updates throughout the codebase/website

I wanted to ask about this before making any attempts to tackle documentation. Does your team prefer to write all the docs, or would you rather the implementer take a first pass at documentation, and your team refines as necessary?

command/agent/retry_join.go
tgross: Sorry, I should have been a little more clear that I'm saying we should treat those two different categories of error differently. 

* The first category is where Nomad can know the configuration is invalid. Ex. the config is `"exec="` but without arguments. In that case, we can return a specific error to the user telling them to go fix their config.
* But the second case is where the failure is outside of Nomad's control. For example, the command we execute fails or you're using `provider=aws` and the IAM policy doesn't actually work. We can't "verify badness" there, so we'd just bubble up the errors we got.

acaloiaro: Gotcha thanks. I believe the latest commit to this PR does exactly what you describe, but let me know if I've misunderstood your intention here. 

Category 1: https://github.com/hashicorp/nomad/pull/18745/files#diff-3cedac9fc5d67acba9b51036e3dce4d1d42d9e468c213124b0565069324f25a9R95
Category 2: https://github.com/hashicorp/nomad/pull/18745/files#diff-3cedac9fc5d67acba9b51036e3dce4d1d42d9e468c213124b0565069324f25a9R119

command/agent/retry_join_test.go
```diff
@@ -166,10 +185,48 @@ func TestRetryJoin_Server_MixedProvider(t *testing.T) {
...
+
+	var output []string
+	mockJoin := func(s []string) (int, error) {
+		output = s
+		return 0, nil
+	}
+
+	// 'exec=*'' tests go-netaddr
+	// 'localhost' also tests get-netaddr by ensuring it resolves to "::1" and "127.0.0.1"
+	// '100.100.100.100' ensures that bare IPs are used as-is
```
acaloiaro, line 204: Remove/update these comments

command/agent/retry_join.go
```diff
@@ -4,14 +4,28 @@
...
-	golog "log"
+	"net"
+	"regexp"
 	"strings"
 	"time"
 
+	golog "log"
+
+	hclog "github.com/hashicorp/go-hclog"
 	log "github.com/hashicorp/go-hclog"
```
tgross, line 17: This is going to sound nitpicky (:grinning:): but we've got two different named imports of `go-hclog` here. Let's pick one and stick with it. Also, let's keep the `golog "log"` import with the group of stdlib imports above.

command/agent/retry_join.go
```diff
@@ -30,6 +44,87 @@ type DiscoverInterface interface {
...
+
+	var ipAddrs []net.IPAddr
+	prefixMatches := cfgPrefixes.FindStringSubmatch(cfg)
+	cfgPrefix := prefixMatches[1]
+	cfgValue := prefixMatches[2]
+	switch cfgPrefix {
+	case "exec":
+		ipAddrs, err = d.netAddrs.IPAddrs(context.Background(),
+			cfg,
+			logger)
```
tgross, line 106: We're definitely not sticklers about 80 col and these are even shorter than that. I'd avoid breaking the parameters onto new lines:

```suggestion
		ipAddrs, err = d.netAddrs.IPAddrs(context.Background(), cfg, logger)
```

(Same for the "dns" case below.)

command/agent/retry_join.go
tgross: Yeah this is great :+1:

command/agent/retry_join_test.go
```diff
@@ -166,10 +185,48 @@ func TestRetryJoin_Server_MixedProvider(t *testing.T) {
...
+	}
+
+	joiner.RetryJoin(serverJoin)
+
+	require.Equal(4, len(output)) // [127.0.0.1 127.0.0.1 100.100.100.100 127.0.0.1]
+	require.Equal("100.100.100.100", output[2])
+	require.Equal(2, len(mockNetaddrs.ReceivedConfig))
+	require.Equal("exec=echo 127.0.0.1", mockNetaddrs.ReceivedConfig[0])
+	require.Equal("localhost", mockNetaddrs.ReceivedConfig[1])
+	require.Equal("provider=aws, tag_value=foo", mockDiscover.ReceivedConfig)
```
tgross, line 227: We haven't migrated all the tests yet but for all new tests we're using [`shoenig/test`](https://pkg.go.dev/github.com/shoenig/test).

```suggestion
	must.Len(t, 4, output) // [127.0.0.1 127.0.0.1 100.100.100.100 127.0.0.1]
	must.Eq(t, "100.100.100.100", output[2])
	must.Len(t, 2, mockNetaddrs.ReceivedConfig)
	must.Eq(t, "exec=echo 127.0.0.1", mockNetaddrs.ReceivedConfig[0])
	must.Eq(t, "localhost", mockNetaddrs.ReceivedConfig[1])
	must.Eq(t, "provider=aws, tag_value=foo", mockDiscover.ReceivedConfig)
```

(Same for those above.)

Also, given that the order is guaranteed we could probably combine the first two assertions into:

```go
	must.Eq(t, []string{"127.0.0.1", "127.0.0.1", "100.100.100.100", "127.0.0.1"}, output)
```

command/agent/retry_join_test.go
tgross: It's fine to leave these

command/agent/retry_join.go
```diff
@@ -30,6 +44,87 @@ type DiscoverInterface interface {
...
+	case "dns":
+		ipAddrs, err = d.netAddrs.IPAddrs(context.Background(),
+			cfgValue,
+			logger)
+		for _, addr := range ipAddrs {
+			addrs = append(addrs, addr.String())
+		}
+	case "provider":
+		addrs, err = d.goDiscover.Addrs(cfg, logger.StandardLogger(&hclog.StandardLoggerOptions{InferLevels: true}))
+	default:
```
tgross, line 120: Your mention of docs had me look at the existing docs, and I learned we have [existing support for DNS](https://developer.hashicorp.com/nomad/docs/configuration/server_join#via-domains-or-dns). If you follow the `serverJoin` and `clientJoin` functions from where they're injected by the caller in `command.go`, the server join code path resolves DNS names down in the Serf library, and the client join code path resolves DNS names itself.

So that complicates our error handling a bit. Fortunately `exec=` and `provider=` and `dns=` aren't valid domain names because of the `=`, so I think what we can do is try to resolve the address here and only if we can't resolve an address do we return an error. We can probably just remove the `dns=` case as well.

Sorry I missed that in my first review. There's a lot of very old code here. :blush:

tgross (commented):
This is looking really good @acaloiaro. I left a note about how we're handling DNS names that you'll want to take a look at.

> I realize now that I probably should have asked you how you prefer to handle conversations/resolutions during CR. I think it would have been more sensible to leave those conversations unresolved until you verify that the actual implementation is to your liking.

Probably, but no big deal. This PR isn't so huge where they'd get lost.

> I wanted to ask about this before making any attempts to tackle documentation. Does your team prefer to write all the docs, or would you rather the implementer take a first pass at documentation, and your team refines as necessary?

It'd be great if you want to contribute docs! You can a section to [`server_join.mdx`](https://github.com/hashicorp/nomad/blob/main/website/content/docs/configuration/server_join.mdx).

Another doc item to add is a changelog entry. If you run `make cl` there's a little wizard that'll walk you through adding the entry file.

command/agent/retry_join.go
acaloiaro: No worries! I really should have begun my implementation with a read of the docs :)

acaloiaro:
@tgross 

I'm wondering if `start_join` should be getting any love here? 

I ask because the documentation for both `retry_join` and `start_join` refer to the [the server address format documentation](https://developer.hashicorp.com/nomad/docs/configuration/server_join#server-address-format) for allowable values. So far, `start_join` has gotten no attention in this PR. As far as I can tell, there is no mention of `start_join` supporting `go-discover` in the past. If that's true, it was never correct to refer to the [server address format documentation](https://developer.hashicorp.com/nomad/docs/configuration/server_join#server-address-format) from `start_join`'s docs. 

With that said the language around both `start_join` and `retry_join` in the docs is a bit off. We know `retry_join` supports `go-discover`'s `provider=` format. 

[`retry_join`'s section](https://developer.hashicorp.com/nomad/docs/configuration/server_join#retry_join) starts off with 
>  Specifies a list of server addresses to join.

Which is obviously not accurate for `go-discover` entries. I wonder what language we should use in the docs to refer to the fact that now these entries may be specified in a number of formats.
- `go-discover`'s `provider=`-prefixed entries
- `go-netaddrs`'s `exec=`-prefixed entries [new]
- ip
- ip:port
- hostname:port
- hostname 

It's fairly misleading to refer to these as a "list of server addresses". I think it may be necessary to introduce a new paragraph, or rename the [server address format](https://developer.hashicorp.com/nomad/docs/configuration/server_join#server-address-format) section to better reflect what `retry_join` may contain.

So I have two questions:

1. Should there be functionality parity between `start_join` and `retry_join`? If not, we'll need to separate the "address format docs" section for `start_join` and `retry_join`. If so, then I think any `start_join` functionality should be addressed in a followup PR, since I believe start/retry already lack functional parity. 
2. Do you have any thoughts on what the "server address format" section should be called now that `retry_join` may contain more than simple server addresses?

command/agent/retry_join.go
acaloiaro: This change greatly simplified `Addrs()` in he latest commit. 

Everything that is not `exec=` or `provider=` prefixed is passed along as-is, which reverts this to the previous behavior. With the `exec=` functionality added, of course. 

All addrs returned by `Addrs()` that are not IPs get resolved by Serf down the line in the server's `Join()`.

website/content/docs/configuration/server_join.mdx
```diff
@@ -113,24 +129,43 @@ unless configured otherwise:
...
+automated cluster joining using cloud provider metadata. See [Cloud
 Auto-join](#cloud-auto-join) for more information.
 
 ```
 "provider=aws tag_key=..." => 1.2.3.4:4648
 ```
 
+### `go-netaddrs` format
+
+As of Nomad 1.x.x (**TODO**: What version would this land in?) `retry_join` accepts
```
acaloiaro, line 145: Calling out this TODO for a response.

website/content/docs/configuration/server_join.mdx
```diff
@@ -225,3 +260,33 @@ project which have the given `tag_value`.
...
+### `go-netaddrs` Configurations
+
+`go-netaddrs` configurations are prefixed with `exec=`.
+
+What follows the `exec=` prefix may be any executable program and its arguments.
+Commands run by `go-netaddrs` must return a list of space-delimited IPv4 or IPv6
+addresses and exit with code `0` on success and non-zero on failure.
+
+Executables may be referenced by absolute file system path or by any program
+name discoverable by `os/exec`'s [LookPath](https://pkg.go.dev/os/exec#LookPath)
```
acaloiaro, line 274: I have a feeling this may be too Go-specific for the audience. While `LookPath` is a nice system-independent way to talk about program path discovery, it's rooted in Go, which Nomad users may not be familiar with. 

Any thoughts on a system-independent way to discuss path discovery without being go-specific?

website/content/docs/configuration/server_join.mdx
```diff
@@ -225,3 +260,33 @@ project which have the given `tag_value`.
...
+Example `go-netaddrs` configuration. 
+```
+server_join {
+  retry_join = [ "exec=hcloud server list -o noheader -o columns=ipv4 -l nomad-server=true | tr '\n' ' '" ]
+}
+```
+Here,  `hcloud`'s newline-delimited output has been re-formatted as space-
+deliminted by piping its output into `tr`.
+
+Output of an exmaple `go-netaddrs` executable
```
tgross, line 286:
```suggestion
Here, `hcloud`'s newline-delimited output has been re-formatted as space-
delimited by piping its output into `tr`.

Output of an example `go-netaddrs` executable
```

website/content/docs/configuration/server_join.mdx
tgross: That's something worth considering, for sure (we run into this with our `template` docs and it's unfortunate). Here I think we can get away with a narrow description of the behavior and then refer to the Go docs:

> You can refer to the executable by its absolute file system path, or by the executable's name if it can be found on the Nomad agent's `PATH`. See [executables in the current directory](https://pkg.go.dev/os/exec#hdr-Executables_in_the_current_directory) for more details on the lookup behavior.

website/content/docs/configuration/server_join.mdx
tgross: We have versioned docs these days, so this would only appear in the major version in which the feature appears (ex. the 1.7.x docs). If we can manage to land this PR in 1.7.0, which seems likely, I think we can remove this clause entirely. Let's leave it out and if we need to adjust it I can come back in later to tweak it.

website/content/docs/configuration/server_join.mdx
```diff
@@ -28,37 +28,53 @@ server_join {
...
+  joined as cluster members. This is similar to [`start_join`](#start_join),
+  but join attempts are retried up to [retry_max](#retry_max) times.
+  Further, `retry_join` is available to both Nomad servers and clients, while
+  `start_join` is only defined for Nomad servers. This is useful for cases where
+  we know the address will become available eventually. Use `retry_join` with an
+  array as a replacement for `start_join`, **do not use both options**.
+
+  Server addresses must conform to the [server address format](#server-address-format).
 
   ```
```
tgross, line 42:
```suggestion
  ```hcl
```

website/content/docs/configuration/server_join.mdx
```diff
@@ -28,37 +28,53 @@ server_join {
...
+    retry_join = [ "1.1.1.1", "2.2.2.2" ]
   }
   ```
 
-  Using the `go-discover` interface, this can be defined both in a client or
-  server configuration as well as provided as a command-line argument.
+  Auto-join configurations must conform to the [Cloud Auto-join format](#cloud-auto-join).
+    
+  Cloud-Auto join using `go-discover` to join an AWS EC2 cluster.
+  ```
```
tgross, line 51:
```suggestion
  Cloud-Auto join using `go-discover` to join an AWS EC2 cluster.
  
  ```hcl
```

website/content/docs/configuration/server_join.mdx
```diff
@@ -28,37 +28,53 @@ server_join {
...
+    retry_join = [ "exec=hcloud server list -o noheader -o columns=ipv4 -l nomad-server=true | tr '\n' ' '" ]
   }
   ```
 
-  See the [server address format](#server-address-format) for more information about expected
-  server address formats.
+  Server addresses and Cloud Auto-join configurations may be used together.
+  This is a mixed configuration containing a server address and Cloud Auto-join
+  configuration.
+  ```
```
tgross, line 69:
```suggestion
  configuration.
  
  ```hcl
```

website/content/docs/configuration/server_join.mdx
```diff
@@ -113,24 +129,43 @@ unless configured otherwise:
 nomad-01.company.local => nomad-01.company.local:4648
 ```
 
-### Via the go-discover interface
+### `go-discover` format
 
 As of Nomad 0.8.4, `retry_join` accepts a unified interface using the
```
tgross, line 134: I have somewhere on a TODO list to remove all these ancient "as of Nomad 0.8..." references :grinning: 

```suggestion
`retry_join` accepts a unified interface using the
```

website/content/docs/configuration/server_join.mdx
```diff
@@ -225,3 +260,33 @@ project which have the given `tag_value`.
...
+What follows the `exec=` prefix may be any executable program and its arguments.
+Commands run by `go-netaddrs` must return a list of space-delimited IPv4 or IPv6
+addresses and exit with code `0` on success and non-zero on failure.
+
+Executables may be referenced by absolute file system path or by any program
+name discoverable by `os/exec`'s [LookPath](https://pkg.go.dev/os/exec#LookPath)
+function.
+
+Example `go-netaddrs` configuration. 
+```
```
tgross, line 278:
```suggestion
Example `go-netaddrs` configuration. 

```hcl
```

command/agent/retry_join.go
tgross: :+1:

command/agent/retry_join_test.go
tgross: (This comment accidentally got marked resolved but wasn't actually fixed)

.changelog/18745.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+server: Add `go-netaddrs` support to `retry_join`
```
tgross, line 2: This impacts both client and server. The bare `retry_join` config is [deprecated](https://developer.hashicorp.com/nomad/docs/configuration/server#deprecated-parameters) so let's refer specifically to the one we most care about:

```suggestion
config: Add `go-netaddrs` support to `server_join.retry_join`
```

go.mod
```diff
@@ -206,6 +206,7 @@ require (
 	github.com/gorilla/mux v1.8.0 // indirect
 	github.com/hashicorp/errwrap v1.1.0 // indirect
 	github.com/hashicorp/go-immutable-radix v1.3.1 // indirect
+	github.com/hashicorp/go-netaddrs v0.1.0 // indirect
```
tgross, line 209: This should be direct now, so let's move this up to the section above and remove the `// indirect`. Running `go mod tidy` should do it.

website/content/docs/configuration/server_join.mdx
acaloiaro: Excellent. That sounds much better.

tgross (changes requested):
This is getting really close, @acaloiaro. Thanks for sticking with it!

> I'm wondering if `start_join` should be getting any love here?
> ...
> However, I do believe that start_join's referencing retry_join's "server address format" section is incorrect, as I don't believe it ever had support for go-discover's provider syntax (and now go-netaddrs' exec= syntax).

You're right that `start_join` shouldn't be referring, however indirectly, to the cloud auto-join, and it doesn't support it. I think the changes you've made to the server address format docs make that more clear. 

I don't think we want to update `start_join`'s behavior, which is intentionally limited in scope.

acaloiaro:
@tgross This is ready for another round of review.

tgross (approved):
LGTM!

I've pushed one last commit mopping up a couple trivial unresolved items that were missed (no worries on that... GitHub's code review is interface, uh, imperfect :grinning: ). Once CI is finished I'll merge this in and you can expect to see it ship in Nomad 1.7.0-beta.2 shortly.

Thanks so much for this PR!

acaloiaro:
Thanks @tgross! I'll keep an eye on CI and try to decipher any failures / attempt fixes.

tgross:
Merged, thanks again!

command/agent/retry_join.go
```diff
@@ -107,24 +170,19 @@ func (r *retryJoiner) RetryJoin(serverJoin *ServerJoin) {
...
-					r.logger.Error("determining join addresses failed", "error", err)
-				} else {
-					addrs = append(addrs, servers...)
-				}
-			default:
-				addrs = append(addrs, addr)
+			servers, err := r.autoDiscover.Addrs(addr, r.logger)
+			if err != nil {
+				r.logger.Error("discovering join addresses failed", "join_config", addr, "error", err)
+				return
```
ryansammonaiven, line 182: @tgross @acaloiaro Is this `return` that was added here necessary and intended? We have found that this makes the process quite brittle as a single failure of the discovery causes it to give up (even if infinite retries are specified). Looking at the way the code was previously, this was not the case.

ryansammonaiven: I created https://github.com/hashicorp/nomad/issues/24560 related to this.

acaloiaro: @ryansammonaiven  I no longer have the cotext of this pull request in my head. 

However, your conclusion appears correct after a quick review of the code. I believe I got this return wrong and `L182` should be removed so that the retry count advances in `L205`.

The test suite should probably be extended to ensure retries on failure.

ryansammonaiven: To close the loop in this thread, this is being addressed in #24561.

PR #18751, opened 2023-10-13
client: correct consul block validation in the consul_hook

pkazmierczak:
vide https://github.com/hashicorp/nomad/pull/18742#discussion_r1357173520

### DIFF
--- a/client/allocrunner/consul_hook.go
+++ b/client/allocrunner/consul_hook.go
@@ -98,12 +98,11 @@ func (h *consulHook) Prerun() error {
 }
 
 func (h *consulHook) prepareConsulTokensForTask(job *structs.Job, task *structs.Task, tgName string, tokens map[string]map[string]string) error {
-	var consulClusterName string
-	if task.Consul != nil && task.Consul.Cluster != "" {
-		consulClusterName = task.Consul.Cluster
-	} else {
-		consulClusterName = structs.ConsulDefaultCluster
+	if task.Consul == nil {
+		return nil
 	}
+
+	consulClusterName := task.Consul.Cluster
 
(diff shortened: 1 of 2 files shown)


client/allocrunner/consul_hook.go
```diff
@@ -98,13 +98,12 @@ func (h *consulHook) Prerun() error {
...
 
 func (h *consulHook) prepareConsulTokensForTask(job *structs.Job, task *structs.Task, tgName string, tokens map[string]map[string]string) error {
-	var consulClusterName string
-	if task.Consul != nil && task.Consul.Cluster != "" {
-		consulClusterName = task.Consul.Cluster
-	} else {
-		consulClusterName = structs.ConsulDefaultCluster
+	if task.Consul == nil {
+		return nil
 	}
```
tgross, line 103: `task.Consul` can be nil if the task group's `Consul` block is non-nil, so I think we need to check there too.

pkazmierczak: but do we ever use task group's `Consul` block to get the Consul cluster name? I don't think we do?

tgross: Ah, we're probably not at the moment. I'll make sure that lands with the rest of the multi-cluster client work.

pkazmierczak: just to be clear, what I mean is that this method, `prepareConsulTokensForTask` is only called for task group tasks, and the only reason we're doing a `nil` check here is because we need the Consul cluster name. I'm just trying to think of  a situation in which `task.Consul` would be `nil` and we'd still need to get a Consul token for it.

PR #18752, opened 2023-10-13
client: use WI-issued consul tokens in the template_hook

pkazmierczak:


client/allocrunner/taskrunner/template_hook.go
```diff
@@ -113,6 +121,20 @@ func (h *templateHook) Prestart(ctx context.Context, req *interfaces.TaskPrestar
...
+		consulTokens := h.config.hookResources.GetConsulTokens()
+
+		var found bool
+		h.consulToken, found = consulTokens[req.Task.Consul.Cluster][req.Task.Consul.IdentityName()]
+		if !found {
+			return fmt.Errorf(
+				"consul task %s uses workload identity, but unable to find a consul token for that task",
+				req.Task.Name,
+			)
+		}
```
tgross, line 135: We're returning an error here because we (rightly, I think) want to protect ourselves if something isn't quite right here. But if that's the case, maybe we should ok-check/nil-check after `consulTokens[req.Task.Consul.Cluster]` too? Just in case that's where the problem is?

tgross (approved):
LGTM!

PR #18754, opened 2023-10-13
auth: use `ACLsDisabledACL` when ACLs are disabled

tgross:
The RPC handlers expect to see `nil` ACL objects whenever ACLs are disabled. By using `nil` as a sentinel value, we have the risk of nil pointer exceptions and improper handling of `nil` when returned from our various auth methods that can lead to privilege escalation bugs. This is the final patch in a series to eliminate the use of `nil` ACLs as a sentinel value for when ACLs are disabled.

This patch adds a new virtual ACL policy field for when ACLs are disabled and updates our authentication logic to use it. Included:

* Extends auth package tests to demonstrate that nil ACLs are treated as failed auth and disabled ACLs succeed auth.
* Adds a new `AllowDebug` ACL check for the weird special casing we have for pprof debugging when ACLs are disabled.
* Removes the remaining unexported methods (and repeated tests) from the `nomad/acl.go` file.
* Update the semgrep rules to detect improper nil ACL checking and remove the old invalid ACL checks.
* Update the contributing guide for RPC authentication.

Ref: https://github.com/hashicorp/nomad/pull/18730
Ref: https://github.com/hashicorp/nomad/pull/18744


jrasell (approved):
LGTM!

Just noting something I looked at when reviewing this PR for future: can/should we avoid building the iradix transactions when performing the client/server/disabled ACL init? Curious what the time/space gain would be, and even if it's safe.

PR #18760, opened 2023-10-16
Run Action websocket endpoint

philrenaud:
A hacky-feeling but working demo of a websocket on /job/:id/action:

1. run a job with a task with the following action block in it:
```
      action "weather" {
        command = "/usr/bin/curl"
        args = ["wttr.in/Toronto?format=3"]
      }
```
2. get an allocation ID where that task is running
3. run:
```
nomad operator api \
-H "Connection: Upgrade" \
-H "Upgrade: websocket" \
-H "Sec-WebSocket-Version: 13" \
-H "Sec-WebSocket-Key: x3JJHMbDL1EzLkh9GBhXDw==" \
'/v1/job/actions-demo/action?task=task&group=group&action=weather&allocID=THAT_ALLOC_ID'
```

4. get a bunch of base64 encoded output

### DIFF
--- a/client/alloc_endpoint.go
+++ b/client/alloc_endpoint.go
@@ -6,7 +6,6 @@ package client
 import (
 	"bytes"
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -165,7 +164,6 @@ func (a *Allocations) Checks(args *cstructs.AllocChecksRequest, reply *cstructs.
 
 // exec is used to execute command in a running task
 func (a *Allocations) exec(conn io.ReadWriteCloser) {
-	a.c.logger.Debug("task exec session started")
 	defer metrics.MeasureSince([]string{"client", "allocations", "exec"}, time.Now())
 	defer conn.Close()
 
@@ -179,13 +177,9 @@ func (a *Allocations) exec(conn io.ReadWriteCloser) {
 		handleStreamResultError(err, code, encoder)
 		return
 	}
-
-	a.c.logger.Info("task exec session ended", "exec_id", execID)
 }
 
 func (a *Allocations) execImpl(encoder *codec.Encoder, decoder *codec.Decoder, execID string) (code *int64, err error) {
-	a.c.logger.Debug("task execImpl session started")
-
 	// Decode the arguments
 	var req cstructs.AllocExecRequest
 	if err := decoder.Decode(&req); err != nil {
@@ -209,13 +203,6 @@ func (a *Allocations) execImpl(encoder *codec.Encoder, decoder *codec.Decoder, e
 		return code, err
 	}
 	alloc := ar.Alloc()
-
-	jsonString, err := json.MarshalIndent(req, "", "  ")
-	if err != nil {
-		a.c.logger.Error("Error converting req to JSON", "error", err)
-	} else {
-		a.c.logger.Info("Request Arguments", "req", string(jsonString))
-	}
 
 	aclObj, ident, err := a.c.resolveTokenAndACL(req.QueryOptions.AuthToken)
 	{
@@ -267,17 +254,10 @@ func (a *Allocations) execImpl(encoder *codec.Encoder, decoder *codec.Decoder, e
 			return nil, err
 		}
 		if jobAction != nil {
-			a.c.logger.Info("Action found", "action", jobAction.Name, "command", jobAction.Command, "args", jobAction.Args)
-			// req.Cmd = append(req.Cmd, jobAction.Command)
 			// append both Command and Args
 			req.Cmd = append([]string{jobAction.Command}, jobAction.Args...)
 		}
-		a.c.logger.Info("So full command is", "command", req.Cmd)
 	}
-
-	// // TODO: TEMP
-	// // Create an arbitrary command example here
-	// req.Cmd = []string{"bash", "-c", "echo hello world"}
 
 	if len(req.Cmd) == 0 {
 		return pointer.Of(int64(400)), errors.New("command is not present")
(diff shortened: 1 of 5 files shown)


client/alloc_endpoint.go
```diff
@@ -343,3 +375,49 @@ func (s *execStream) Recv() (*drivers.ExecTaskStreamingRequestMsg, error) {
...
+	var tg *nstructs.TaskGroup
+	for _, group := range alloc.Job.TaskGroups {
+		if group.Name == taskGroup {
+			tg = group
+			break
+		}
+	}
+	if tg == nil {
+		return nil, fmt.Errorf("task group %s not found", taskGroup)
+	}
```
tgross, line 397: An `Allocation` always belongs to exactly one task group, so there's a `TaskGroup` field with the name of the task group. You can then use the `LookupTaskGroup` helper on `structs.Job` to get the `*structs.TaskGroup` you need here.

client/alloc_endpoint.go
```diff
@@ -343,3 +375,49 @@ func (s *execStream) Recv() (*drivers.ExecTaskStreamingRequestMsg, error) {
...
+
+	var t *nstructs.Task
+	for _, taskStruct := range tg.Tasks {
```
tgross, line 400: Likewise there's a `LookupTask` method on `TaskGroup`.

command/agent/job_endpoint.go
```diff
@@ -358,6 +367,210 @@ func (s *HTTPServer) jobActions(resp http.ResponseWriter, req *http.Request, job
...
+
+	if err := s.readActionWsHandshake(conn.ReadJSON, req, &args.QueryOptions); err != nil {
+		conn.WriteMessage(websocket.CloseMessage,
+			websocket.FormatCloseMessage(toWsCode(400), err.Error()))
+		return nil, err
+	}
+
+	s.logger.Info("jobRunAction thru handshake")
+
+	return s.runActionStreamImpl(conn, &args)
```
tgross, line 433: I'd strongly consider at this point whether we can call `s.execStreamImpl` here without reimplementing a copy of it.

command/agent/alloc_endpoint.go
```diff
@@ -520,7 +520,7 @@ func (s *HTTPServer) allocExec(allocID string, resp http.ResponseWriter, req *ht
 
 // readWsHandshake reads the websocket handshake message and sets
 // query authentication token, if request requires a handshake
-func readWsHandshake(readFn func(interface{}) error, req *http.Request, q *structs.QueryOptions) error {
+func (s *HTTPServer) readWsHandshake(readFn func(interface{}) error, req *http.Request, q *structs.QueryOptions) error {
```
tgross, line 523: Now that you've removed your debugging logging, you don't need this change to make this function a method anymore. That also fixes the linter error you're getting on `command/agent/alloc_endpoint_test.go`

client/alloc_endpoint.go
```diff
@@ -247,6 +247,20 @@ func (a *Allocations) execImpl(encoder *codec.Encoder, decoder *codec.Decoder, e
 	if req.Task == "" {
 		return pointer.Of(int64(400)), taskNotPresentErr
 	}
+
+	// If an action is present, go find the command and args
+	if req.Action != "" {
+		alloc, _ := a.c.GetAlloc(req.AllocID)
+		jobAction, err := validateActionExists(req.Action, req.Task, alloc)
+		if err != nil {
+			return nil, err
```
tgross, line 256: We're returning a pointer to a HTTP code here, so we don't want to return nil on error:

```suggestion
			return pointer.Of(int64(400)), err
```

client/alloc_endpoint.go
```diff
@@ -343,3 +357,21 @@ func (s *execStream) Recv() (*drivers.ExecTaskStreamingRequestMsg, error) {
...
+			break
+		}
+	}
+
+	if action == nil {
+		return nil, fmt.Errorf("action %s not found", action)
+	}
+
+	return action, nil
+}
```
tgross, line 377: Typically we return early when possible from a loop like this. That lets you reduce it to something like:

```suggestion
	for _, action := range t.Actions {
		if action.Name == actionName {
			return action, nil
		}
	}

	return nil, fmt.Errorf("action %s not found", actionName)
}
```

(Plus fixed a bug in the error return value, which would have always been `"action <nil> not found"`)

client/structs/structs.go
```diff
@@ -183,6 +183,10 @@ type AllocExecRequest struct {
 	Cmd []string
 
 	structs.QueryOptions
+
+	// The name of a predefined command to be executed (optional)
+	Action string
+	JobID  string
```
tgross, line 189: Style nitpick so that the embedded struct falls at the end of the list of fields (we typically have it there or in the beginning of the list). Also, do we still need `JobID` in this struct?

```suggestion
	// The name of a predefined command to be executed (optional)
	Action string
	JobID  string

	structs.QueryOptions
```

tgross (commented):
This is looking great, @philrenaud

tgross (approved):
LGTM!

command/agent/job_endpoint.go
```diff
@@ -358,6 +362,47 @@ func (s *HTTPServer) jobActions(resp http.ResponseWriter, req *http.Request, job
...
 
+func (s *HTTPServer) jobRunAction(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
+
+	s.logger.Info("jobRunAction called")
+
+	// Build the request and parse the ACL token
+	task := req.URL.Query().Get("task")
+	action := req.URL.Query().Get("action")
+	allocID := req.URL.Query().Get("allocID")
+	ttyB := false
```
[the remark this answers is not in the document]
philrenaud: TIL Apps Hungarian!

PR #18768, opened 2023-10-16
agent: add config option to enable file and line log detail.

jrasell:
Most useful for developers or very curious operators. It means we do not need to grep/ag log messages in the codebase or know where they come from.

The CLI flag has been added along with docs for convince and this could be useful within development environments or when debugging Nomad issues.


Juanadelacuesta (approved):
Nice addition!

tgross (approved):
Very cool. :+1:

PR #18778, opened 2023-10-17
Action CLI command

philrenaud:
Adds the `nomad action` CLI command. Can be used to run a task action either by passing an allocation ID + job, or job + group + task:

```
nomad action \
  -group=group \
  -task=task \
  -job=actions-demo \
weather
```

```
nomad action \
  -allocation=3a5a5361 \
  -job=actions-demo \
weather
```

As a side-effect, also adds an optional `tasakGroupName` param to `getRandomJobAlloc`, in order to only select allocations that have the specified action name.

### DIFF
--- a/api/allocations.go
+++ b/api/allocations.go
@@ -91,6 +91,7 @@ func (a *Allocations) Info(allocID string, q *QueryOptions) (*Allocation, *Query
 // long pauses on this API call.
 func (a *Allocations) Exec(ctx context.Context,
 	alloc *Allocation, task string, tty bool, command []string,
+	action string,
 	stdin io.Reader, stdout, stderr io.Writer,
 	terminalSizeCh <-chan TerminalSize, q *QueryOptions) (exitCode int, err error) {
 
@@ -100,6 +101,7 @@ func (a *Allocations) Exec(ctx context.Context,
 		task:    task,
 		tty:     tty,
 		command: command,
+		action:  action,
 
 		stdin:  stdin,
 		stdout: stdout,
--- a/api/allocations.go
+++ b/api/allocations.go
@@ -91,7 +91,6 @@ func (a *Allocations) Info(allocID string, q *QueryOptions) (*Allocation, *Query
 // long pauses on this API call.
 func (a *Allocations) Exec(ctx context.Context,
 	alloc *Allocation, task string, tty bool, command []string,
-	action string,
 	stdin io.Reader, stdout, stderr io.Writer,
 	terminalSizeCh <-chan TerminalSize, q *QueryOptions) (exitCode int, err error) {
 
@@ -101,7 +100,6 @@ func (a *Allocations) Exec(ctx context.Context,
 		task:    task,
 		tty:     tty,
 		command: command,
-		action:  action,
 
 		stdin:  stdin,
 		stdout: stdout,
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -4,8 +4,10 @@
 package api
 
 import (
+	"context"
 	"errors"
 	"fmt"
+	"io"
 	"net/url"
 	"sort"
 	"strconv"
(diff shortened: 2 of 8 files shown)


command/action.go
```diff
@@ -0,0 +1,412 @@
...
+		}
+	}()
+
+	// TODO: is make([]string, 0) the right thing to pass an empty command here?
+	return client.Allocations().Exec(ctx,
+		alloc, task, tty, make([]string, 0), action, stdin, stdout, stderr, sizeCh, nil)
+}
+
+// isTty returns true if both stdin and stdout are a TTY
+// TODO: why does this matter?
```
tgross, line 330: The command at the other side may have changed behaviors if we're in a tty (ex. interactive vs non-interactive behaviors), so you want to make that available to end users.

command/action.go
```diff
@@ -0,0 +1,412 @@
...
+		l.Ui.Error(fmt.Sprintf("failed to exec into task: %v", err))
+		return 1
+	}
+
+	return code
+}
+
+// execImpl invokes the Alloc Exec api call, it also prepares and restores terminal states as necessary.
+func (l *ActionCommand) execImpl(client *api.Client, alloc *api.Allocation, task string, job string, action string, tty bool,
+	escapeChar string, stdin io.Reader, stdout, stderr io.WriteCloser) (int, error) {
```
tgross, line 262: The `execImpl` method in `command/alloc_exec.go` can be made into a free function because it doesn't use anything from its own `(l *AllocExecCommand)`. I would do that and then call it from `Run` above instead of copying all this code (and then trying to get away with dropping features).

nomad/structs/actions.go
```diff
@@ -26,6 +26,19 @@ type ActionListResponse struct {
...
 }
 
+type JobRunActionRequest struct {
+	JobID     string
+	TaskGroup string
+	Task      string
+	Action    string
+	AllocID   string
+	WriteRequest
+	QueryOptions
```
tgross, line 36: `WriteRequest` is for writes, and `QueryOptions` is for reads. You should only set one of these.

command/action.go
```diff
@@ -0,0 +1,412 @@
...
+	signal.Notify(signalCh, os.Interrupt, syscall.SIGTERM)
+	go func() {
+		for range signalCh {
+			cancelFn()
+		}
+	}()
+
+	// TODO: is make([]string, 0) the right thing to pass an empty command here?
+	return client.Allocations().Exec(ctx,
+		alloc, task, tty, make([]string, 0), action, stdin, stdout, stderr, sizeCh, nil)
```
tgross, line 326: This should be going to a new `client.Job().Action(...)` that calls the `command/agent/job_endpoint.go` for the actions HTTP API.

api/allocations.go
```diff
@@ -91,6 +91,7 @@ func (a *Allocations) Info(allocID string, q *QueryOptions) (*Allocation, *Query
 // long pauses on this API call.
 func (a *Allocations) Exec(ctx context.Context,
```
tgross, line 92: To implement the equivalent for actions, I think we want to:
* Copy this to something `(*Job) ActionExec`
* Add a field to `execSession` to change the path that it makes the initial connection to (so that it points to our new HTTP API endpoint).

philrenaud: the execSession field I ended up watching for is `action`; it modifies reqpath if present: https://github.com/hashicorp/nomad/pull/18778/files#diff-3735e959db9a859486511196a56619a36ce39c1ab4179a709ef6ccfd97d35d15R101-R106

command/commands.go
```diff
@@ -235,6 +235,11 @@ func Commands(metaPtr *Meta, agentUi cli.Ui) map[string]cli.CommandFactory {
 				Meta: meta,
 			}, nil
 		},
+		"action": func() (cli.Command, error) {
```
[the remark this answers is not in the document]
philrenaud: We're leaving the door open for things like "node actions", so opting to keep this as `nomad action` for extensibility I think.

command/action.go
```diff
@@ -0,0 +1,412 @@
...
+		l.Stderr = os.Stderr
+	}
+
+	action := args[0]
+	// Convert action to command here.
+	// TODO: Noting that this is also being (redundantly) done in client/alloc_endpoint.go's execImpl.
+	// I could pass the action name all the way down, but that would be maybe polluting the param space of
+	// api/allocations.go, api/allocations_exec.go, etc., since alloc exec normally would never use an Action
+	// except by via this command.
+	// TODO: update: I think I'll try that anyway.
```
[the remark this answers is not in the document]
philrenaud: (Opted for option 2 here after you explained the general flow of this command)

api/allocations.go
```diff
@@ -91,6 +91,7 @@ func (a *Allocations) Info(allocID string, q *QueryOptions) (*Allocation, *Query
...
 	alloc *Allocation, task string, tty bool, command []string,
+	action string,
```
tgross, line 94: I missed this the first time around, but fortunately it's showing up in broken tests: we can't add fields to methods in the `api` package without breaking the Go API's compatibility with potential third party consumers (like those our customers might have built).

So we're going to need a new method here. What I'd recommend is something with a signature like:

```go
func (a *Allocations) ExecWithOptions(ctx context.Context, execOpts *ExecOpts, q *QueryOptions) (exitCode int, err error) {
```

And then the `ExecOpts` struct will look like this:

```go
type ExecOpts struct {
	Alloc          *Allocation
	Task           string
	TTY            bool
	Command        []string
	Action         string
	Stdin          io.Reader
	Stdout         io.Writer
	Stderr         io.Writer
	TerminalSizeCh <-chan TerminalSize
}
```

Normally this is an unusually tedious way of writing Go, but for the public Go API it lets us add fields without breaking the signature of the method due to Go's handling of default zero values.

api/jobs.go
```diff
@@ -1514,3 +1516,31 @@ type JobEvaluateRequest struct {
 type EvalOptions struct {
 	ForceReschedule bool
 }
+
+// nomad action is used to run a pre-defined command inside a running task.
```
tgross, line 1520: This is "Go style" for the docstring of public API:

```suggestion
// ActionExec is used to run a pre-defined command inside a running task.
```

command/alloc_fs.go
```diff
@@ -387,7 +387,7 @@ func (f *AllocFSCommand) followFile(client *api.Client, alloc *api.Allocation,
 
 // Get Random Allocation from a known jobID. Prefer to use a running allocation,
 // but use a dead allocation if no running allocations are found
-func getRandomJobAlloc(client *api.Client, jobID, namespace string) (*api.AllocationListStub, error) {
+func getRandomJobAlloc(client *api.Client, jobID, tasakGroupName, namespace string) (*api.AllocationListStub, error) {
```
tgross, line 390:
```suggestion
func getRandomJobAlloc(client *api.Client, jobID, taskGroupName, namespace string) (*api.AllocationListStub, error) {
```

command/action.go
```diff
@@ -0,0 +1,317 @@
...
+  -job <job-id>
+    Specifies the job in which the Action is defined
+
+  -allocation <allocation-id>
+    Specifies the allocation in which the Action is defined. If not provided,
+    a group and task name must be provided and a random allocation will be
+    selected from the job.
+
+  -task <task-name>
+	  Specifies the task in which the Action is defined. Required if no
```
tgross, line 58: You've got some indentation issues in the command text. I _think_ you might have some tabs here, but we can't use them inside these giant strings without messing up the alignment.

tgross (approved):
I've left a few comments to clean up (and some of those are getting picked up by the linter too), but otherwise this is looking good once those are resolved.

PR #18780, opened 2023-10-17
`consul_hook` tests

pkazmierczak:


client/widmgr/mock.go
```diff
@@ -129,7 +133,7 @@ func (m MockWIDMgr) Run() error { return nil }
 func (m MockWIDMgr) Get(identity structs.WIHandle) (*structs.SignedWorkloadIdentity, error) {
 	sid, ok := m.swids[identity]
 	if !ok {
-		return nil, fmt.Errorf("identity not found")
+		return nil, fmt.Errorf("identity name %s for workload %s not found", identity.IdentityName, identity.WorkloadIdentifier)
```
tgross, line 136: Just in case we want to write a test that checks for specific errors, maybe this message should exactly match the one returned from the real implementation? https://github.com/hashicorp/nomad/blob/main/client/widmgr/widmgr.go#L143

client/consul/consul_testing.go
```diff
@@ -4,12 +4,28 @@
...
+}
 
 func (mc *MockConsulClient) DeriveSITokenWithJWT(reqs map[string]JWTLoginRequest) (map[string]string, error) {
+	if mc.tokens != nil && len(mc.tokens) > 0 {
+		return mc.tokens, nil
+	}
+
 	tokens := make(map[string]string, len(reqs))
 	for id := range reqs {
 		tokens[id] = uuid.Generate()
```
tgross, line 31: Should this now call `SetTokens` on these, too?

client/allocrunner/consul_hook.go
```diff
@@ -87,6 +101,9 @@ func (h *consulHook) Prerun() error {
 }
 
 func (h *consulHook) prepareConsulTokensForTask(task *structs.Task, tg *structs.TaskGroup, tokens map[string]map[string]string) error {
+	if task == nil {
+		return nil
+	}
```
tgross, line 106: Can this ever actually be nil? It seems like it'd be a programming error if so, in which case I'd expect this to return an error. Probably better to just let it panic.

client/allocrunner/consul_hook_test.go
```diff
@@ -0,0 +1,214 @@
...
+	"github.com/hashicorp/nomad/nomad/mock"
+	"github.com/hashicorp/nomad/nomad/structs"
+	structsc "github.com/hashicorp/nomad/nomad/structs/config"
+	"github.com/shoenig/test/must"
+)
+
+// statically assert consul hook implements the expected interfaces
+var _ interfaces.RunnerPrerunHook = (*consulHook)(nil)
+
+func testHarness(t *testing.T) *consulHook {
```
tgross, line 25: This is going to sound nitpicky, but this function is visible for all tests in `allocrunner`, so maybe it should be named such that it's scoped to this file?

```suggestion
func consulHookTestHarness(t *testing.T) *consulHook {
```

client/allocrunner/consul_hook_test.go
```diff
@@ -0,0 +1,214 @@
...
+	hook := testHarness(t)
+	task := hook.alloc.LookupTask("web")
+
+	tests := []struct {
+		name        string
+		task        *structs.Task
+		tokens      map[string]map[string]string
+		wantErr     bool
+		errMsg      string
+		emptyTokens bool
```
tgross, line 105: Because there are a bunch of other identities on this job, maybe we should assert that we got tokens for _specific_ identities and not just that we got some tokens at all? (Same for the test below.)

tgross (approved):
LGTM! I've left some suggestions for improvements but nothing blocking.

client/widmgr/mock.go
pkazmierczak: good thinking

client/allocrunner/consul_hook.go
pkazmierczak: I'd rather return an error than panic here? But yes, this could only be a programming error.

PR #18786, opened 2023-10-17
Do not embed *Server

gulducat:
These structs embedding Server, then Server _also embedding them_, confused my IDE, isn't necessary, and just feels wrong!

This is all I could find with

```
$ find -name '*.go' | xargs grep -P '^\s+\*Server$' -n
./heartbeat.go:38:      *Server
./rpc.go:47:    *Server
./plan_apply.go:25:     *Server
```

Server still embeds them, which isn't my favorite, but this makes my IDE happy, and Server's `*rpcHandler` has many tendrils, so maybe that can be a side quest for another day.


tgross (approved):
LGTM!

There's a semgrep complaint about some existing code that you didn't touch here, and that's mostly a lint anyways so fine to bypass.

PR #18790, opened 2023-10-18
vault: add multi-cluster support on templates

lgfa29:
In Nomad Enterprise, a task may connect to a non-default Vault cluster, requiring `consul-template` to be configured with a specific client `vault` block.

Tests require an ENT feature, so they will be done in a separate PR on the ENT repo.

### DIFF
--- a/client/allocrunner/taskrunner/template/template.go
+++ b/client/allocrunner/taskrunner/template/template.go
@@ -102,6 +102,10 @@ type TaskTemplateManagerConfig struct {
 	// VaultToken is the Vault token for the task.
 	VaultToken string
 
+	// VaultCluster is the Vault cluster to use for this template. It may be
+	// empty if the task does not use Vault.
+	VaultCluster string
+
 	// VaultNamespace is the Vault namespace for the task
 	VaultNamespace string
 
@@ -881,30 +885,32 @@ func newRunnerConfig(config *TaskTemplateManagerConfig,
 	emptyStr := ""
 	conf.Vault.RenewToken = pointer.Of(false)
 	conf.Vault.Token = &emptyStr
-	if cc.VaultConfig != nil && cc.VaultConfig.IsEnabled() {
-		conf.Vault.Address = &cc.VaultConfig.Addr
+
+	vaultConfig := cc.GetVaultConfigs(nil)[config.VaultCluster]
+	if vaultConfig != nil && vaultConfig.IsEnabled() {
+		conf.Vault.Address = &vaultConfig.Addr
 		conf.Vault.Token = &config.VaultToken
 
 		// Set the Vault Namespace. Passed in Task config has
 		// highest precedence.
-		if config.ClientConfig.VaultConfig.Namespace != "" {
-			conf.Vault.Namespace = &config.ClientConfig.VaultConfig.Namespace
+		if vaultConfig.Namespace != "" {
+			conf.Vault.Namespace = &vaultConfig.Namespace
 		}
 		if config.VaultNamespace != "" {
 			conf.Vault.Namespace = &config.VaultNamespace
 		}
 
-		if strings.HasPrefix(cc.VaultConfig.Addr, "https") || cc.VaultConfig.TLSCertFile != "" {
-			skipVerify := cc.VaultConfig.TLSSkipVerify != nil && *cc.VaultConfig.TLSSkipVerify
+		if strings.HasPrefix(vaultConfig.Addr, "https") || vaultConfig.TLSCertFile != "" {
+			skipVerify := vaultConfig.TLSSkipVerify != nil && *vaultConfig.TLSSkipVerify
 			verify := !skipVerify
 			conf.Vault.SSL = &ctconf.SSLConfig{
 				Enabled:    pointer.Of(true),
 				Verify:     &verify,
-				Cert:       &cc.VaultConfig.TLSCertFile,
-				Key:        &cc.VaultConfig.TLSKeyFile,
-				CaCert:     &cc.VaultConfig.TLSCaFile,
-				CaPath:     &cc.VaultConfig.TLSCaPath,
-				ServerName: &cc.VaultConfig.TLSServerName,
+				Cert:       &vaultConfig.TLSCertFile,
+				Key:        &vaultConfig.TLSKeyFile,
+				CaCert:     &vaultConfig.TLSCaFile,
+				CaPath:     &vaultConfig.TLSCaPath,
+				ServerName: &vaultConfig.TLSServerName,
 			}
 		} else {
 			conf.Vault.SSL = &ctconf.SSLConfig{
--- a/client/allocrunner/taskrunner/template/template.go
+++ b/client/allocrunner/taskrunner/template/template.go
@@ -25,6 +25,7 @@ import (
 	"github.com/hashicorp/nomad/client/taskenv"
 	"github.com/hashicorp/nomad/helper/pointer"
 	"github.com/hashicorp/nomad/nomad/structs"
+	structsc "github.com/hashicorp/nomad/nomad/structs/config"
 )
 
 const (
@@ -102,9 +103,9 @@ type TaskTemplateManagerConfig struct {
 	// VaultToken is the Vault token for the task.
 	VaultToken string
 
-	// VaultCluster is the Vault cluster to use for this template. It may be
-	// empty if the task does not use Vault.
-	VaultCluster string
+	// VaultConfig is the Vault configuration to use for this template. It may
+	// be nil if the task does not use Vault.
+	VaultConfig *structsc.VaultConfig
 
 	// VaultNamespace is the Vault namespace for the task
 	VaultNamespace string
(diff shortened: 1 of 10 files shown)


client/allocrunner/taskrunner/template/template.go
```diff
@@ -881,30 +885,32 @@ func newRunnerConfig(config *TaskTemplateManagerConfig,
 	emptyStr := ""
 	conf.Vault.RenewToken = pointer.Of(false)
 	conf.Vault.Token = &emptyStr
-	if cc.VaultConfig != nil && cc.VaultConfig.IsEnabled() {
-		conf.Vault.Address = &cc.VaultConfig.Addr
+
+	vaultConfig := cc.GetVaultConfigs(nil)[config.VaultCluster]
+	if vaultConfig != nil && vaultConfig.IsEnabled() {
```
lgfa29, line 890: `GetVaultConfigs()` in ENT does not return disabled configs, so it's not possible to differentiate between a cluster that is not configured (and that should maybe result in an error) from one that is disabled (expected, no error).

But I'm not sure if this is relevant. Tasks should only run in clients that do have the given config because of the constraint. 

In case the config does get changed and the client restarts, I think it's expected the task to just keep running?

client/config/config_ce.go
```diff
@@ -19,7 +19,7 @@ func (c *Config) GetVaultConfigs(logger hclog.Logger) map[string]*structsc.Vault
 		return nil
 	}
 
-	if len(c.VaultConfigs) > 1 {
+	if len(c.VaultConfigs) > 1 && logger != nil {
 		logger.Warn("multiple Vault configurations are only supported in Nomad Enterprise")
 	}
```
lgfa29, line 24: Sometimes we don't have a logger, in which case I think it's OK to skip this?

nomad/job_endpoint_hooks.go
```diff
@@ -420,6 +418,9 @@ func (v *jobValidate) validateVaultIdentity(t *structs.Task) ([]error, error) {
 		return warnings, nil
 	}
 
+	hasTaskWID := len(vaultWIDs) > 0
+	hasDefaultWID := v.srv.config.VaultIdentityConfig(t.Vault.Cluster) != nil
```
lgfa29, line 422: Moving these down allows us to take advantage of the `t.Vault == nil` conditional above and access `t.Vault.Cluster` directly.

nomad/config.go
```diff
@@ -485,14 +485,15 @@ func (c *Config) ConsulTaskIdentity() *structs.WorkloadIdentity {
...
 }
 
-// VaultDefaultIdentity returns the workload identity to be used for accessing
-// the Vault API.
-func (c *Config) VaultDefaultIdentity() *structs.WorkloadIdentity {
-	if c.VaultConfig == nil {
+// VaultIdentityConfig returns the workload identity to be used for accessing
+// the API of a given Vault cluster.
+func (c *Config) VaultIdentityConfig(cluster string) *structs.WorkloadIdentity {
+	conf := c.VaultConfigs[cluster]
```
pkazmierczak, line 491: is it ever possible that we get a panic here?

client/allocrunner/taskrunner/template/template.go
```diff
@@ -881,30 +885,32 @@ func newRunnerConfig(config *TaskTemplateManagerConfig,
 	emptyStr := ""
 	conf.Vault.RenewToken = pointer.Of(false)
 	conf.Vault.Token = &emptyStr
-	if cc.VaultConfig != nil && cc.VaultConfig.IsEnabled() {
-		conf.Vault.Address = &cc.VaultConfig.Addr
+
+	vaultConfig := cc.GetVaultConfigs(nil)[config.VaultCluster]
```
pkazmierczak, line 889: is it ever possible that we can't get that key?

pkazmierczak (approved):
LGTM, I only left some nit-picky comments.

client/allocrunner/taskrunner/template/template.go
tgross: The task _should_ keep running but the template runner isn't forked from the Nomad agent. So if you break the Vault config while the client is down, the task is going to fail to restore. I think that's expected behavior though.

client/config/config_ce.go
tgross: Really? Why wouldn't we always have a logger (even if just the test logger)?

tgross (approved):
LGTM

client/allocrunner/taskrunner/template/template.go
lgfa29: Yup, in which case `vaultConfig` is `nil` and we skip the block below. This could happen if the task doesn't have a `vault` block, and so `config.VaultCluster` is an empty string, or potentially in the case I mentioned below: https://github.com/hashicorp/nomad/pull/18790/files#r1363015410

client/allocrunner/taskrunner/template/template.go
lgfa29: > The task should keep running but the template runner isn't forked from the Nomad agent. So if you break the Vault config while the client is down, the task is going to fail to restore

Oh, good point. But I actually think that, if we don't return an error on missing key here, the task _will_ restore but without any Vault integration 🤔 

I can quickly test this, but I'm now thinking we may need to be able to differentiate between missing and disabled.

client/config/config_ce.go
lgfa29: The template manager (the context I'm using this function in this PR) doesn't have one:
https://github.com/hashicorp/nomad/blob/99e54da9a9aeb3a62a05f50af1dcc74c06cd9657/client/allocrunner/taskrunner/template/template.go#L49-L74

Alternatively, I think I could pass a whole Vault config instead of [just the cluster name](https://github.com/hashicorp/nomad/pull/18790/files#diff-796a6e8da6dbce4e38ef692cf9f9e70b5b5770f9cf367e3e13ca1f86de356d01R105-R107) and resolve the config at the hook level, which does have a logger. Not sure if it's worth it though 🤔

nomad/config.go
lgfa29: Like if `c.VaultConfigs` is `nil`? No panic since it's declared as a map: https://go.dev/play/p/MECHEgWMOIL

It _would_ panic if we try to modify it: https://go.dev/play/p/yS53iMVWy5h

pkazmierczak: yeah sorry I missed it

client/allocrunner/taskrunner/template/template.go
lgfa29: This took a bit longer than I expected because I started testing several failure modes 😅 

But the results are kind of interesting. As it currently is, the task will indeed fail on restore _but_ that's because of the `vault_hook` failing to get a client here:
https://github.com/hashicorp/nomad/blob/16d71582f68ba5a74504ae6311fe8ac6c8eaea9b/client/allocrunner/taskrunner/vault_hook.go#L166-L169

image

The task failing and being rescheduled in a node with the required Vault cluster makes sense, but if it weren't for the `vault_hook` the task would try to start and just hang for a while trying to get the secret:

image

For some reason, after 6m (kind of a random value 🤔), consul-template tries to render the template again, but fails. The interesting bit here is that it seems to be using some default value for the Vault address.

So while technically correct right now, I feel like hooks should work (and fail) independently, regardless of other hooks or the order in which they're called.

All this leads me to think that I _should_ raise an error here, meaning that I would need to access the map of Vault clients directly instead of using `GetVaultConfig()`. Or maybe `GetVaultConfig()` should have an `ok` return as well?

Any second opinions on this?

tgross: > So while technically correct right now, I feel like hooks should work (and fail) independently, regardless of other hooks or the order in which they're called.

I don't know about that... many or most of the hooks are dependent on ordering (for the prerun method). It's kind of implicit in the design that each sets up the environment needed for those that follow (ex. allocdir before identity, identity before Consul/Vault, Consul/Vault tokens before templates or services, CSI before volume mounts, etc.)

That being said, I wouldn't be opposed to raising an error here. It lets us bail out from a situation we know we'll never properly recover from as early as possible. We can probably just check for a nil config from `GetVaultConfigs` and then return an error like `vault cluster "foo" does not exist or is not enabled` and that'll avoid having to return something more complex than the map we already have.

client/config/config_ce.go
tgross: It looks like `template.go` needs the whole Vault config anyways, so looking it up in the caller lets us separate the "policy" of picking the cluster from the "mechanism" of setting up the template manager. (I would ask "what did we do for services?" but we configure the client selector _far_ away from the service registration client, so that's probably not a good example. :grimacing: )

client/allocrunner/taskrunner/template/template.go
lgfa29: > I don't know about that... many or most of the hooks are dependent on ordering (for the prerun method). It's kind of implicit in the design that each sets up the environment needed for those that follow 

Ah good point, they're called in a specific order on purpose. 

Interestingly, passing the whole Vault config to the template manager (as described in https://github.com/hashicorp/nomad/pull/18790#discussion_r1364436792) helps us solve this problem as well since the `template_hook` can better detect the error, so we get a two-for-one fix 😄

client/config/config_ce.go
lgfa29: Ah thanks! I modified this part of the code in https://github.com/hashicorp/nomad/pull/18790/commits/ac6985231d59f4161f6e0f1b0fad2caefffb938d to pass the whole config. 

This also helps to fix https://github.com/hashicorp/nomad/pull/18790#discussion_r1363015410, so win-win 🙂

nomad/config.go
lgfa29: No worries, it's always good to keep an eye out for things like this.

PR #18791, opened 2023-10-18
ui: fix websocket connections on dev proxy

lgfa29:
`ember-cli` includes a handler for websocket upgrade requests to start proxying requests. Handling the same upgrade event by also proxying in `api.js` causes some kind of conflict where the connection is closed unexpectedly.

The only change necessary on upgrade is to overwrite the Origin header so Nomad accepts the connection.


philrenaud (approved):
This works perfectly, thank you Luiz!

PR #18793, opened 2023-10-18
[ui] Actions implementation in the web UI

philrenaud:
The Web UI implementation of [Nomad Actions](https://github.com/hashicorp/nomad/pull/18794)

This adds Actions dropdowns to job pages and task rows in tables and jobs where applicable, then runs a websocket query to execute them against a new job/:id/action endpoint.

image
image

PR review note: there is quite a bit of noise here from a `groupTaskCount` to `groupAllocCount` for in many mirage mocked jobs. This name better reflects what is actually being done in the job factory, as I've added a lever to help separate out task and alloc counts when creating one (this became relevant in order to mock a job with multiple groups with differing numbers of tasks and allocs)

Another note on design: the "streaming messages in a code block in a notification" is probably not the end state of the design for this feature, but since it is still in flux, I've opted to move this PR forward and address design changes in the near future.


ui/app/adapters/job.js
```diff
@@ -159,4 +166,98 @@ export default class JobAdapter extends WatchableNamespaceIDs {
...
+
+    const protocol = window.location.protocol === 'https:' ? 'wss:' : 'ws:';
+    const region = this.system.activeRegion;
+    const applicationAdapter = getOwner(this).lookup('adapter:application');
+    const prefix = `${
+      applicationAdapter.host || window.location.host
+    }/${applicationAdapter.urlPrefix()}`;
+
+    const wsUrl =
+      `${protocol}//${prefix}/job/${job.get('id')}/action` +
```
lgfa29, line 181:
```suggestion
      `${protocol}//${prefix}/job/${encodeURIComponent(job.get('id')(}/action` +
```

Not sure if this is handled at another level, but job IDs can have pretty much any character 😬

ui/app/adapters/job.js
```diff
@@ -159,4 +166,98 @@ export default class JobAdapter extends WatchableNamespaceIDs {
...
+      `?namespace=*&action=${action.name}&allocID=${allocID}&task=${action.task.name}&group=${action.task.taskGroup.name}&tty=true&ws_handshake=true` +
```
lgfa29, line 182: I'm not sure if `namespace=*` will work in this endpoint, and even if it does I think we want to more precise here and use the job's namespace to avoid any unexpected requests.

lgfa29 (commented):
Some early feedback. I will take a close look next week 🙂

ui/app/adapters/job.js
```diff
@@ -159,4 +167,138 @@ export default class JobAdapter extends WatchableNamespaceIDs {
...
+      `${protocol}//${prefix}/job/${encodeURIComponent(job.get('id'))}/action` +
+      `?namespace=${job.get('namespace.id')}&action=${
+        action.name
+      }&allocID=${allocID}&task=${action.task.name}&group=${
+        action.task.taskGroup.name
+      }&tty=true&ws_handshake=true` +
+      (region ? `&region=${region}` : '');
+
+    // const socket = new WebSocket(wsUrl);
+    let socket;
```
DingoEatingFuzz, line 191: Does it make sense to eventually try to use the sockets service here?

ui/app/adapters/job.js
```diff
@@ -159,4 +167,138 @@ export default class JobAdapter extends WatchableNamespaceIDs {
...
+        })
+        .getFlashObject();
+
+      socket.send(
+        JSON.stringify({ version: 1, auth_token: this.token?.secret || '' })
+      );
+      socket.send(
+        JSON.stringify({
+          tty_size: { width: 250, height: 100 }, // Magic numbers, but they pass the eye test.
+        })
```
DingoEatingFuzz, line 251: I believe this is what we use the xterm-addon-fit for over in [ExecTerminal](https://github.com/hashicorp/nomad/blob/1ef99f05364b7d3739befa6a789f0d55b2314dcf/ui/app/components/exec-terminal.js).

ui/app/adapters/job.js
```diff
@@ -159,4 +167,138 @@ export default class JobAdapter extends WatchableNamespaceIDs {
...
+        // explicitly close the socket and return;
+        socket.close();
+        return;
+      }
+
+      let jsonData = JSON.parse(event.data);
+      if (jsonData.stdout && jsonData.stdout.data) {
+        messageBuffer += base64DecodeString(jsonData.stdout.data);
+        messageBuffer += '\n';
+        messageBuffer = messageBuffer.replace(/\x1b\[[0-9;]*[a-zA-Z]/g, '');
```
DingoEatingFuzz, line 267: I think this magic could use a comment. Is this some control sequence that's getting noped out? What's the control sequence?

ui/app/styles/core/notifications.scss
```diff
@@ -15,5 +15,40 @@ section.notifications {
...
+      margin-top: 1rem;
+      margin-bottom: -$codeBottomMargin;
+    }
+
+    pre {
+      display: inline-block;
+      background: inherit;
+      color: inherit;
+      border: none;
+      // white-space: pre-wrap;
```
DingoEatingFuzz, line 45: nit: could be deleted.

ui/mirage/factories/task-group.js
```diff
@@ -123,29 +127,29 @@ export default Factory.extend({
           unknown: 0.25,
           lost: 0.1,
         };
-      
+
```
DingoEatingFuzz, line 130: 🙈

ui/tests/acceptance/actions-test.js
```diff
@@ -0,0 +1,243 @@
...
+    assert
+      .dom(
+        '.job-page-header .actions-dropdown .hds-dropdown__list li.hds-dropdown-list-item--variant-generic'
+      )
+      .exists({ count: 4 });
+    assert
+      .dom(
+        '.job-page-header .actions-dropdown .hds-dropdown__list li.hds-dropdown-list-item--variant-interactive'
+      )
+      .exists({ count: 1 });
```
DingoEatingFuzz, line 134: This is just some guy's opinion, so not at all blocking, but selectors like this is why page objects were invented.

ui/mirage/factories/job.js
```diff
@@ -231,13 +234,18 @@ export default Factory.extend({
...
       allocStatusDistribution: job.allocStatusDistribution,
     };
 
     if (job.groupTaskCount) {
-      groupProps.count = job.groupTaskCount;
+      groupProps.taskCount = job.groupTaskCount;
+    }
+
+    if (job.groupAllocCount) {
+      groupProps.count = job.groupAllocCount;
```
DingoEatingFuzz, line 248: I suppose it was only a matter of time before this needed to be disambiguated.

DingoEatingFuzz (approved):
Super awesome work!

I think the headline for me (granted I haven't been in this code in a minute) is that there is a lot of exposure to websocket and xterm details in the action adapter.

There's a good chance that code never has to be looked at again, so I don't want to go so far as to say it should be refactored, but since we already have some of those utils for full-blown exec, maybe there is an opportunity to organize some of the common code.

ui/app/adapters/job.js
philrenaud: These strip out ansi escape chars (`\03`, `\x1b[` etc.) that indicate anything from carriage return to colour change.

I'm treating this less like an interactive exec window and more like a raw output display, so have opted to escape and remove these characters.

ui/tests/acceptance/actions-test.js
philrenaud: Good call, added exactly this: https://github.com/hashicorp/nomad/pull/18793/files#diff-b71a63231834a9079b61b90359e0e3459c6d9d6e776afe0a3fcd665a2dba63c1

ui/app/adapters/job.js
philrenaud: Ended up starting/stopping on this a couple times today; I think it's a good idea and will be marking it a TODO but I'm worried about drift from the exec implementation enough that I want to give it more consideration.

PR #18795, opened 2023-10-18
monitor: add log include location option on monitor CLI and API

jrasell:
This change allows the use of the `log-include-location` logger options when using the agent monitor API and CLI. This defaults to false.

The PR also makes use of this change within the `operator debug` command, enabling this by default. Debug bundles are captured when as much information as needed should be gathered from a Nomad cluster. Therefore, including this information by default makes sense for both Nomad operators and the extended Nomad engineering team.

### DIFF
--- a/command/agent/agent_endpoint.go
+++ b/command/agent/agent_endpoint.go
@@ -187,14 +187,26 @@ func (s *HTTPServer) AgentMonitor(resp http.ResponseWriter, req *http.Request) (
 		plainText = parsed
 	}
 
+	logIncludeLocation := false
+	logIncludeLocationStr := req.URL.Query().Get("log_include_location")
+	if logIncludeLocationStr != "" {
+		parsed, err := strconv.ParseBool(logIncludeLocationStr)
+		if err != nil {
+			return nil, CodedError(http.StatusBadRequest,
+				fmt.Sprintf("Unknown option for log_include_location: %v", err))
+		}
+		logIncludeLocation = parsed
+	}
+
 	nodeID := req.URL.Query().Get("node_id")
 	// Build the request and parse the ACL token
 	args := cstructs.MonitorRequest{
-		NodeID:    nodeID,
-		ServerID:  req.URL.Query().Get("server_id"),
-		LogLevel:  logLevel,
-		LogJSON:   logJSON,
-		PlainText: plainText,
+		NodeID:             nodeID,
+		ServerID:           req.URL.Query().Get("server_id"),
+		LogLevel:           logLevel,
+		LogJSON:            logJSON,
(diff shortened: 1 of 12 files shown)


website/content/docs/commands/operator/debug.mdx
```diff
@@ -55,6 +55,9 @@ true.
 
 - `-log-level=DEBUG`: The log level to monitor. Defaults to `DEBUG`.
 
+- `-log-include-location`: Include file and line information in each log line
+  monitored. The default is `true`.
```
lgfa29, line 59: Does `-log-include-location=false` work to disable this flag? And would we ever want to? I can't think of any downside of a hardcoded `true` value here, but also can't see why an extra flag would hurt 🤷

lgfa29 (approved):
🔍

website/content/docs/commands/operator/debug.mdx
jrasell: Yeah I tested this locally and it does disable the log location as expected. I don't expect it would ever be used, but adding it was trivial 🤷🏻

command/agent/agent_endpoint.go
```diff
@@ -187,14 +187,26 @@ func (s *HTTPServer) AgentMonitor(resp http.ResponseWriter, req *http.Request) (
 		plainText = parsed
 	}
 
+	logIncludeLocation := false
+	logIncludeLocationStr := req.URL.Query().Get("log_include_location")
+	if logIncludeLocationStr != "" {
+		parsed, err := strconv.ParseBool(logIncludeLocationStr)
```
gulducat, line 193: Since this is a boolean parameter, could we instead use its presence/absence as the indicator (like the CLI flag behaves) rather than parsing a string value to bool?  maybe with [Query().Has()](https://pkg.go.dev/net/url#Values.Has)?

This relates to the tests below that make sure `log_include_location=maybe` is invalid -- if the value is discarded instead, then it can't be invalid.

The only case I can think of where it might be confusing is if someone tries to set it explicitly to "false" -- maybe the error condition could instead be it being set to any value at all, instead of just being present?

I know there's some other API endpoint(s) with a boolean query param like this, but I'm drawing a blank right now.

jrasell: I believe the CLI flag behaves in both ways where you can either provide it in `-bool-flag` or `-bool-flag=true` which is confusing for sure and no doubt we handle this in both ways.

I understand the desire to simplify and seeing as this is a default false parameter on the monitor command it would make sense, however, I am leaning to keeping this behaviour as it matches the `log_json` boolean parameter on this endpoint.

Let me know if you have strong opinions on changing this.

PR #18799, opened 2023-10-18
E2E: Consul compatibility matrix tests

tgross:
Set up a new test suite that exercises Nomad's compatibility with Consul. This suite installs all currently supported versions of Consul, spins up a Consul agent with appropriate configuration, and a Nomad agent running in dev mode. Then it runs a Connect job against each pair.

Work in progress:
- [x] needs https://github.com/hashicorp/nomad/pull/18821 to pass
- [x] needs to have the appropriate bits in place to reuse as much code as possible for the ENT version of the same test, with Consul namespaces
- [x] clean up any remaining TODO remarks that won't ship in this PR

Note to run these locally for testing, you'll need to run as `sudo` and the tests will download Consul. So something like the following will let you avoid re-downloading as you're working on these:

```sh
NOMAD_E2E_CONSULCOMPAT_BASEDIR=/tmp/compat \
  NOMAD_E2E_CONSULCOMPAT=1 \
  sudo -E /usr/local/go/bin/go test -v ./e2e/consulcompat
```


e2e/consulcompat/shared_run_test.go
```diff
@@ -0,0 +1,102 @@
...
+// previous runs or from the development environment)
+func verifyConsulVersion(t *testing.T, consulAPI *consulapi.Client, version string) {
+	self, err := consulAPI.Agent().Self()
+	must.NoError(t, err)
+	vers := self["Config"]["Version"].(string)
+	must.Eq(t, version, vers)
+}
+
+// verifyConsulFingerprint ensures that we've successfully fingerprinted Consul
+func verifyConsulFingerprint(t *testing.T, nc *nomadapi.Client, version, clusterName string) {
```
lgfa29, line 29: That's a good test. I should add it to `vaultcompat` as well 😄

e2e/consulcompat/run_ce_test.go
```diff
@@ -0,0 +1,262 @@
...
+			Address: consulHTTPAddr,
+			Auth:    "",
+			Token:   consulToken,
+		}
+
+		nStop, nc := startNomad(t, consulCfg)
+		t.Cleanup(nStop)
+
+		verifyConsulFingerprint(t, nc, b.Version, "default")
+		runConnectJob(t, nc)
```
lgfa29, line 48: Should we gate this on Linux and maybe have a non-Connect test for other systems?

lgfa29 (approved):
Some minor comments, but nice work!

e2e/consulcompat/run_ce_test.go
```diff
@@ -0,0 +1,262 @@
...
+
+		// give nomad and consul time to stop
+		t.Cleanup(func() { time.Sleep(5 * time.Second) })
```
gulducat, line 51: another nit-pick, but again this seems like something the helper could handle

e2e/consulcompat/input/consul-auth-method-config.json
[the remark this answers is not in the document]
tgross: Oops, good catch.

e2e/consulcompat/run_ce_test.go
tgross: Yeah agreed we should gate it on Linux. I'm not sure it's worth doing a separate test for non-Linux. There's no OS-dependent behaviors other than the ones we just don't support at all (Connect).

gulducat (commented):
Great stuff!

I have a couple minor test helper design nits, but my :red_circle: concern is `sudo` setup will probably break in enterprise CI....

e2e/consulcompat/run_ce_test.go
```diff
@@ -0,0 +1,262 @@
...
+	consulapi "github.com/hashicorp/consul/api"
+	"github.com/hashicorp/nomad/helper/uuid"
+	"github.com/hashicorp/nomad/testutil"
+	"github.com/shoenig/test/must"
+)
+
+func testConsulBuildLegacy(t *testing.T, b build, baseDir string) {
+	t.Run("consul-legacy("+b.Version+")", func(t *testing.T) {
+		cStop, consulHTTPAddr, consulAPI := startConsul(t, b, baseDir, "")
+		t.Cleanup(func() { cStop() })
```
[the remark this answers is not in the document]
tgross: The usual reason I do this is because it lets the caller control the order of shutdown when you have multiple cleanup functions. I'm not sure it matters in this particular case, but I haven't written the ENT-side tests yet.

.github/workflows/test-e2e.yml
```diff
@@ -51,6 +51,26 @@ jobs:
...
+      - name: Git config token
+        if: endsWith(github.repository, '-enterprise')
+        run: git config --global url.'https://${{ env.ELEVATED_GITHUB_TOKEN }}@github.com'.insteadOf 'https://github.com'
+      - uses: hashicorp/setup-golang@v1
+      - name: Consul Compatability
+        run: |
+          make deps
+          sudo make cni
+          sudo sed -i 's!Defaults!#Defaults!g' /etc/sudoers
+          sudo -E env "PATH=$PATH" make integration-test-consul
```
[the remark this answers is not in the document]
tgross: So what's our path forward to test this in ENT?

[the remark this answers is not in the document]
gulducat: The most expedient thing to do is just not use self-hosted runners, and therefore also not pull the secret from vault.  I'll post a suggested change in a sec to save you some trouble.

I guess we put this on the pile of "things that actually do need root for testing."  Someday that list will need reckoning with, but so far that day has not yet come.

.github/workflows/test-e2e.yml
```diff
@@ -51,6 +51,26 @@ jobs:
...
+    runs-on: ${{ endsWith(github.repository, '-enterprise') && fromJSON('["self-hosted", "ondemand", "linux"]') || 'ubuntu-latest' }}
+    steps:
+      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+      - uses: ./.github/actions/vault-secrets
+        with:
+          paths: |-
+            kv/data/github/hashicorp/nomad-enterprise/gha ELEVATED_GITHUB_TOKEN ;
+      - name: Git config token
+        if: endsWith(github.repository, '-enterprise')
+        run: git config --global url.'https://${{ env.ELEVATED_GITHUB_TOKEN }}@github.com'.insteadOf 'https://github.com'
```
gulducat, line 65: I believe this should do the trick:

```suggestion
    runs-on: 'ubuntu-latest' # this job requires sudo, so not currently suitable for self-hosted runners
    steps:
      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
      - name: Git config token
        if: endsWith(github.repository, '-enterprise')
        run: git config --global url.'https://${{ secrets.ELEVATED_GITHUB_TOKEN }}@github.com'.insteadOf 'https://github.com'
```

other jobs in this predicament only set Git config token over in ent ([like this one](https://github.com/hashicorp/nomad-enterprise/blob/v1.6.2%2Bent/.github/workflows/test-core.yaml#L82-L83)), but I think the condition here will work fine.

tgross: Thanks! That also catches that we should pin the OS version for the sake of reliability.

e2e/consulcompat/run_ce_test.go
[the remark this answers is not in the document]
tgross: There's some grody stuff in the Consul start function though. Lemme see if I can clean this up a bit.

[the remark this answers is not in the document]
tgross: Done!

e2e/consulcompat/shared_setup_test.go
```diff
@@ -0,0 +1,117 @@
...
+	path := filepath.Join(baseDir, binDir, b.Version)
+	cwd, _ := os.Getwd()
+	os.Chdir(path)      // so that we can launch Consul from the current directory
+	defer os.Chdir(cwd) // return to the test dir so we can find job files
+
+	oldpath := os.Getenv("PATH")
+	os.Setenv("PATH", path+":"+oldpath)
+	t.Cleanup(func() {
+		os.Setenv("PATH", oldpath)
+	})
```
gulducat, line 38: I know already merged (good!), but fyi `testing.T` has a nice lil convenience method for this, which also prevents accidentally running it in parallel: [`t.Setenv`](https://pkg.go.dev/testing#T.Setenv)

```suggestion
	oldpath := os.Getenv("PATH")
	t.Setenv("PATH", path+":"+oldpath)
```

tgross: TIL! I'm almost certainly going to touch these files again after the ENT test gets written. I'll pick that up then.

PR #18808, opened 2023-10-19
cli: use single dep func for opening URLs.

jrasell:
The cap library is used by the login command and the functionality that it provides to the CLI looks the same. Therefore swapping the login command out means we can reduce our dependency imports and use a single library.


philrenaud (approved):
absolutely love stuff-removal, looks good

PR #18809, opened 2023-10-19
lang: add a helper for iterating a map in order

shoenig:
In some cases it is helpful to iterate a map in the sorted order of
the map's keyset - particularly in implementations of some function for
which the tests cannot be deterministic without order.

It's a little odd that the only use for this is in `ent` at first, but eh.


jrasell (approved):
LGTM!

PR #18810, opened 2023-10-19
vault: validate tasks using non-default clusters

lgfa29:
Since Nomad servers only start a Vault client for the default cluster, tasks using non-default clusters must provide an identity to be used for token derivation, either in the task itself or in the agent configuration.

### DIFF
--- a/nomad/job_endpoint_hooks.go
+++ b/nomad/job_endpoint_hooks.go
@@ -384,61 +384,53 @@ func (v *jobValidate) validateServiceIdentity(s *structs.Service, parent string)
 	return nil
 }
 
+// validateVaultIdentity validates that a task is properly configured to access
+// a Vault cluster.
+//
+// It assumes the jobImplicitIdentitiesHook mutator hook has been called to
+// inject task identities if necessary.
 func (v *jobValidate) validateVaultIdentity(t *structs.Task) ([]error, error) {
-	var mErr *multierror.Error
 	var warnings []error
 
-	var vaultWID *structs.WorkloadIdentity
-	vaultWIDs := make([]string, 0, len(t.Identities))
-
-	for _, wid := range t.Identities {
-		if strings.HasPrefix(wid.Name, structs.WorkloadIdentityVaultPrefix) {
-			vaultWIDs = append(vaultWIDs, wid.Name)
-
-			if t.Vault != nil && wid.Name == t.Vault.IdentityName() {
-				vaultWID = wid
+	if t.Vault == nil {
+		// Warn if task doesn't use Vault but has Vault identities.
+		for _, wid := range t.Identities {
+			if strings.HasPrefix(wid.Name, structs.WorkloadIdentityVaultPrefix) {
+				warnings = append(warnings, fmt.Errorf("Task %s has an identity called %s but no vault block", t.Name, wid.Name))
 			}
 		}
-	}
-
-	if t.Vault == nil {
-		for _, wid := range vaultWIDs {
-			warnings = append(warnings, fmt.Errorf("Task %s has an identity called %s but no vault block", t.Name, wid))
-		}
 		return warnings, nil
 	}
 
-	hasTaskWID := vaultWID != nil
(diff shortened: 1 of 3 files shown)


nomad/job_endpoint_hooks.go
```diff
@@ -415,6 +421,17 @@ func (v *jobValidate) validateVaultIdentity(t *structs.Task) ([]error, error) {
 		return warnings, nil
 	}
 
+	// Only the default cluster is allowed to use the legacy flow. Using a
+	// non-default cluster requires an identity to be provided either in the
+	// task or as a default in the agent config.
+	if t.Vault.Cluster != structs.VaultDefaultCluster {
+		mErr = multierror.Append(mErr, fmt.Errorf(
```
tgross, line 428: This is making me realize we're not taking advantage of the multierror in this method at all. We always return for the first error.

nomad/job_endpoint_hooks.go
```diff
@@ -415,6 +421,17 @@ func (v *jobValidate) validateVaultIdentity(t *structs.Task) ([]error, error) {
 		return warnings, nil
 	}
 
+	// Only the default cluster is allowed to use the legacy flow. Using a
+	// non-default cluster requires an identity to be provided either in the
+	// task or as a default in the agent config.
```
tgross, line 426: It's not clear to me how this conditional expresses what's in this comment. We're only checking for the non-default cluster and we'll always hit this unless we've hit an earlier set of errors.

lgfa29: Most of the time we expect a task that uses Vault to have the expected identity, either provided directly in the task or via the agent config, so they end up here:
https://github.com/hashicorp/nomad/blob/142510b99c530c57a9cc9286678581b0426989fc/nomad/job_endpoint_hooks.go#L411-L422

If no identity is available, then we fall through the rest of the code in this function. Previously the code assumed that the legacy flow would be used, so there was only this:
https://github.com/hashicorp/nomad/blob/142510b99c530c57a9cc9286678581b0426989fc/nomad/job_endpoint_hooks.go#L435-L439

But non-default clusters are not allowed to use the legacy flow, so we need to check for this before proceeding with legacy flow validation.

I can clarify the comment on this.

nomad/job_endpoint_hooks.go
lgfa29: Hum....true. This logic used to be _a lot_ more complex before, so I think it may have cases where multiple errors were returned. But now I think we can return a single error.

Another thing I'm realizing is that we run mutators before validators, so the `implicitiy_identities` hook should have already handled the agent config part, so I think we can only validate the task here.

nomad/job_endpoint_hooks.go
tgross: Ah, I got it now. I was misreading where the `hasTaskWID || hasDefaultWID` conditional always returns as though it only returned on a warning.

lgfa29: Yeah, it's a bit confusing. I flipped the logic in https://github.com/hashicorp/nomad/commit/ace07aa14843223106167becba1f8d3462cc4b0a to look for problems first instead of returning early in the happy path. Is it easier to follow like this?

https://github.com/hashicorp/nomad/blob/ace07aa14843223106167becba1f8d3462cc4b0a/nomad/job_endpoint_hooks.go#L387-L434

tgross: Oh that's nice. :+1:

nomad/job_endpoint_hooks.go
lgfa29: Fixed in https://github.com/hashicorp/nomad/pull/18810/commits/ace07aa14843223106167becba1f8d3462cc4b0a

nomad/job_endpoint_hooks.go
lgfa29: Cool! I ported that commit to this PR.

pkazmierczak (approved):
LGTM!

PR #18811, opened 2023-10-19
testutil: add ACL bootstrapping to test server configuration

tgross:
Some of our `api` package tests have ACLs enabled, but none of those tests also run clients and the "wait for the clients to be live" code reads from the Node API. The caller can't bootstrap ACLs until `NewTestServer` returns, and this makes for a circular dependency.

Allow developers to provide a bootstrap token to the test server config, and if it's available, have the server bootstrap the ACL system with it before checking for live clients.

Pulled out of https://github.com/hashicorp/nomad/pull/18799

### DIFF
--- a/testutil/server.go
+++ b/testutil/server.go
@@ -93,7 +93,8 @@ type VaultConfig struct {
 
 // ACLConfig is used to configure ACLs
 type ACLConfig struct {
-	Enabled bool `json:"enabled"`
+	Enabled        bool   `json:"enabled"`
+	BootstrapToken string `json:"-"` // not in the real config
 }
 
 // ServerConfigCallback is a function interface which can be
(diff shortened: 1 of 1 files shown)


testutil/server.go
```diff
@@ -195,6 +196,7 @@ func NewTestServer(t testing.T, cb ServerConfigCallback) *TestServer {
 	if nomadConfig.Stderr != nil {
 		stderr = nomadConfig.Stderr
 	}
+	t.Logf("CONFIG JSON: %s", string(configContent))
```
tgross, line 199: This now matches the equivalent logging we have for test Consul servers, and feels pretty nice when I've been working on https://github.com/hashicorp/nomad/pull/18799

lgfa29 (approved):
Nice! Testing with ACL enabled will help make the scenario more realistic.

PR #18812, opened 2023-10-19
Upgrade go-msgpack to v2 2.1.1

swenson:
And set the `time.Time` option to use the go-msgpack-1.1.5-compatible encoding in all the places, since that is the (now previous) version in `go.mod`.

v2 2.1.1 was specifically designed to honor backwards compatibility with 1.1.5 and 0.5.5, and to clean up the code base to be more maintainable. There may performance lost with the 1.1.5 to 2.1.1 migration since the fastpath code was removed, but the increased safety is probably worth it. See
[the release notes for go-msgkack 2.1.0](https://github.com/hashicorp/go-msgpack/releases/tag/v2.1.0) for more details.

I tested this by running this code, and booting up a cluster with a node also running the prior version of Nomad (before the upgrade). Before I made the changes to set the right `time.Time` option, the previous-version node would throw a bunch of time-decoding errors. After fixing the option, the node came up smoothly, even after changing leadership between them.

This relies on
- [x] https://github.com/hashicorp/serf/pull/705
- [x] https://github.com/hashicorp/raft-boltdb/pull/38
- [x] https://github.com/hashicorp/raft/pull/577
- [x] https://github.com/hashicorp/memberlist/pull/292

and maybe
- [x] https://github.com/hashicorp/net-rpc-msgpackrpc/pull/12

and will need to be updated after they are merged to get the `go.mod` fixes removed.


go.mod
```diff
@@ -286,3 +288,11 @@ require (
...
 	kernel.org/pub/linux/libs/security/libcap/psx v1.2.66 // indirect
 )
+
+replace github.com/hashicorp/raft => /Users/swenson/projects/raft
+
+replace github.com/hashicorp/raft-boltdb/v2 => /Users/swenson/projects/raft-boltdb/v2
+
+replace github.com/hashicorp/memberlist => /Users/swenson/projects/memberlist
+
+replace github.com/hashicorp/serf => /Users/swenson/projects/serf
```
gulducat, line 298: woops! I think this is causing all the current CI failures

swenson: Yep, there is a parameter being added in these (`MsgpackUseNewTimeFormat`) to help with compatibility, so this PR's build won't work until these are merged anyway.

go.mod
```diff
@@ -13,6 +13,8 @@ replace (
 // Nomad is built using the current source of the API module.
 replace github.com/hashicorp/nomad/api => ./api
 
+replace github.com/hashicorp/nomad/client => ./client
+
```
schmichael, line 17: Did you mean to commit this? `./api` is only replaced because it is also a Go module while `./client` is just a Go package.

schmichael (commented):
This LGTM but will want to do some local testing first. I'm also not sure why tests don't appear to have run. :thinking: 

Once all of the other PRs are approved, can we merge from the "inside out" so that we can switch from commit hash pins to real versions?

go.mod
swenson: Something was breaking when I was trying to build. Not sure what this was about. I'll remove it.

swenson:
I think tests weren't running because it was a draft PR? I resolved conflicts and re-pushed.

schmichael:
Replaced by #20173 and #20186

PR #18817, opened 2023-10-20
log: add Consul and Vault cluster name to output

lgfa29:
Ensure Consul and Vault loggers have the cluster name as an attribute to help differentiate log source.

### DIFF
--- a/client/consul/consul.go
+++ b/client/consul/consul.go
@@ -72,7 +72,7 @@ func NewConsulClient(config *config.ConsulConfig, logger hclog.Logger) (Client, 
 		return nil, fmt.Errorf("nil consul config")
 	}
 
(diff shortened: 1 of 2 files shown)


client/consul/consul.go
```diff
@@ -72,7 +72,7 @@ func NewConsulClient(config *config.ConsulConfig, logger hclog.Logger) (Client,
 		return nil, fmt.Errorf("nil consul config")
 	}
 
-	logger = logger.Named("consul")
+	logger = logger.Named("consul").With("name", config.Name)
```
lgfa29, line 75: I'm not sure if there's another place in Consul where it would be worth adding this, but this struct seemed to be doing most of the logs.

tgross (approved):
LGTM

PR #18820, opened 2023-10-20
cli: consul setup command

pkazmierczak:
An interactive setup helper for configuring Consul to accept Nomad WI-enabled workloads.

### DIFF
--- a/command/setup.go
+++ b/command/setup.go
@@ -22,7 +22,7 @@ Usage: nomad setup <subcommand> [options] [args]
 
   Setup Consul for Nomad:
 
-      $ nomad setup consul -method-name-services="nomad-workloads"
+      $ nomad setup consul -y
 
   Please see the individual subcommand help for detailed usage information.
 `
--- a/command/setup.go
+++ b/command/setup.go
@@ -30,10 +30,10 @@ Usage: nomad setup <subcommand> [options] [args]
 }
 
 // Synopsis satisfies the cli.Command Synopsis function.
-func (a *SetupCommand) Synopsis() string { return "Interact with setup helpers" }
+func (c *SetupCommand) Synopsis() string { return "Interact with setup helpers" }
 
 // Name returns the name of this command.
-func (a *SetupCommand) Name() string { return "setup" }
+func (c *SetupCommand) Name() string { return "setup" }
 
 // Run satisfies the cli.Command Run function.
-func (a *SetupCommand) Run(_ []string) int { return cli.RunResultHelp }
+func (c *SetupCommand) Run(_ []string) int { return cli.RunResultHelp }
--- a/command/setup.go
+++ b/command/setup.go
@@ -1,0 +1,39 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package command
+
+import (
+	"strings"
+
+	"github.com/mitchellh/cli"
+)
+
+type SetupCommand struct {
+	Meta
+}
+
+// Help satisfies the cli.Command Help function.
+func (a *SetupCommand) Help() string {
+	helpText := `
+Usage: nomad setup <subcommand> [options] [args]
+
+  This command groups helper subcommands used for setting up Consul and Vault.
+
+  Setup Consul for Nomad:
+
+      $ nomad setup consul -method-name-services="nomad-workloads"
+
+  Please see the individual subcommand help for detailed usage information.
+`
+	return strings.TrimSpace(helpText)
+}
+
+// Synopsis satisfies the cli.Command Synopsis function.
+func (a *SetupCommand) Synopsis() string { return "Interact with setup helpers" }
+
+// Name returns the name of this command.
+func (a *SetupCommand) Name() string { return "setup" }
+
+// Run satisfies the cli.Command Run function.
+func (a *SetupCommand) Run(_ []string) int { return cli.RunResultHelp }
--- a/command/setup.go
+++ b/command/setup.go
@@ -14,7 +14,7 @@ type SetupCommand struct {
 }
 
 // Help satisfies the cli.Command Help function.
-func (a *SetupCommand) Help() string {
+func (c *SetupCommand) Help() string {
 	helpText := `
 Usage: nomad setup <subcommand> [options] [args]
 
--- a/command/setup.go
+++ b/command/setup.go
@@ -18,7 +18,7 @@ func (c *SetupCommand) Help() string {
 	helpText := `
 Usage: nomad setup <subcommand> [options] [args]
 
-  This command groups helper subcommands used for setting up Consul and Vault.
+  This command groups helper subcommands used for setting up external tools.
 
   Setup Consul for Nomad:
 
--- a/command/setup_consul.go
+++ b/command/setup_consul.go
@@ -7,25 +7,31 @@ import (
 	_ "embed"
 	"encoding/json"
 	"fmt"
-	"os"
-	"path/filepath"
 	"strings"
 	"time"
 
 	"github.com/hashicorp/consul/api"
 	"github.com/mitchellh/cli"
 	"github.com/posener/complete"
-	"golang.org/x/exp/maps"
 )
 
 // Ensure SetupConsulCommand satisfies the cli.Command interface.
 var _ cli.Command = &SetupConsulCommand{}
 
 //go:embed asset/consul-wi-default-auth-method-config.json
-var defaultAuthConfigText []byte
+var authConfigBody []byte
 
 //go:embed asset/consul-wi-default-policy.hcl
-var defaultPolicyText []byte
+var policyBody []byte
+
+const (
+	authMethodServices = "nomad-services"
+	authMethodTasks    = "nomad-tasks"
+	roleTasks          = "role-nomad-tasks"
+	policyName         = "policy-nomad-tasks"
+	defaultTTL         = 8 * time.Hour
+	aud                = "consul.io"
+)
 
 type SetupConsulCommand struct {
 	Meta
@@ -34,16 +40,10 @@ type SetupConsulCommand struct {
 	// reuse the same connection.
 	client *api.Client
 
-	methodNameServices   string
-	methodNameTasks      string
-	roleTasks            string
-	policyTemplatesPaths stringSetFlags
-	aud                  stringSetFlags
-	jwksURL              string
-	ttl                  string
+	jwksURL string
 
-	json bool
-	tmpl string
+	// if set, answers "Yes" to all the interactive questions
+	autoYes bool
 }
 
 // Help satisfies the cli.Command Help function.
@@ -56,9 +56,7 @@ Usage: nomad setup consul [options]
 
   Setup Consul for Nomad:
 
-      $ nomad setup consul -auth-method-name="nomad-workloads"
-
-  Please see the individual subcommand help for detailed usage information.
+	$ nomad setup consul -y -jwks="http://nomad.example/.well-known/jwks.json"
 `
 	return strings.TrimSpace(helpText)
 }
@@ -66,14 +64,8 @@ Usage: nomad setup consul [options]
 func (s *SetupConsulCommand) AutocompleteFlags() complete.Flags {
 	return mergeAutocompleteFlags(s.Meta.AutocompleteFlags(FlagSetClient),
 		complete.Flags{
-			"-method-name-services":   complete.PredictAnything,
-			"-method-name-tasks":      complete.PredictAnything,
-			"-role-tasks":             complete.PredictAnything,
-			"-policy-templates-paths": complete.PredictFiles("*.hcl"),
-			"-aud":                    complete.PredictSet("consul.io"),
-			"-jwks-url":               complete.PredictAnything,
-			"-json":                   complete.PredictNothing,
-			"-t":                      complete.PredictAnything,
+			"-jwks-url": complete.PredictAnything,
+			"-y":        complete.PredictSet("true", "false"),
 		})
 }
 
@@ -92,15 +84,8 @@ func (s *SetupConsulCommand) Run(args []string) int {
 
 	flags := s.Meta.FlagSet(s.Name(), FlagSetClient)
 	flags.Usage = func() { s.Ui.Output(s.Help()) }
-	flags.StringVar(&s.methodNameServices, "method-name-services", "nomad-workloads", "")
-	flags.StringVar(&s.methodNameTasks, "method-name-tasks", "nomad-tasks", "")
-	flags.StringVar(&s.roleTasks, "role-tasks", "role-tasks", "")
-	flags.Var(&s.policyTemplatesPaths, "template-policy", "Path to a policy file used for the template role (accepts multiple)")
-	flags.Var(&s.aud, "aud", "consul.io")
-	flags.StringVar(&s.jwksURL, "jwks-url", "http://localhost:4646/.well-known/jwks.json", "")
-	flags.StringVar(&s.ttl, "ttl", "8h", "Maximum token TTL")
-	flags.BoolVar(&s.json, "json", false, "")
-	flags.StringVar(&s.tmpl, "t", "", "")
+	flags.BoolVar(&s.autoYes, "y", false, "")
+	flags.StringVar(&s.jwksURL, "jwks", "http://localhost:4646/.well-known/jwks.json", "")
 	if err := flags.Parse(args); err != nil {
 		return 1
 	}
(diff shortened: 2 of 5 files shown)


command/setup_consul.go
```diff
@@ -0,0 +1,449 @@
...
+	}
+
+	s.Ui.Output(`
+In order to map claims between Nomad's JWTs and Consul ACL, we need to create
+the following binding rule:
+{
+	"Description": "binding rule for Nomad workload identities (WI)",
+	"AuthMethod": "nomad-services",
+	"BindType": "service",
+	"BindName": "${value.nomad_namespace}-${value.nomad_service}"
```
tgross, line 154: After doing some end-to-end testing of reachability, it looks like we don't actually want this bit:

```suggestion
	"BindName": "${value.nomad_service}"
```

command/setup_consul.go
```diff
@@ -0,0 +1,449 @@
...
+  # configuration file.
+
+  # Consul mTLS configuration.
+  # ssl       = true
+  # ca_file   = "/var/ssl/bundle/ca.bundle"
+  # cert_file = "/etc/ssl/consul.crt"
+  # key_file  = "/etc/ssl/consul.key"
+
+  service_auth_method = "nomad-services"
+  task_auth_method    = "nomad-tasks"
```
lgfa29, line 310:
```suggestion
```

I think we're already using the default values for these.

command/setup_consul.go
```diff
@@ -0,0 +1,449 @@
...
+	return authConfig, nil
+}
+
+func (s *SetupConsulCommand) createAuthMethod(authMethodName string, authMethodConf map[string]any) error {
+	_, _, err := s.client.ACL().AuthMethodCreate(&api.ACLAuthMethod{
+		Name:          authMethodName,
+		Type:          "jwt",
+		DisplayName:   authMethodName,
+		Description:   "login method for Nomad workload identities (WI)",
+		MaxTokenTTL:   defaultTTL,
```
lgfa29, line 366:
```suggestion
```

I don't think we want a max TTL for Consul tokens since we can't renew them?

command/setup_consul.go
```diff
@@ -0,0 +1,449 @@
...
+
+func (s *SetupConsulCommand) createPolicy() error {
+	_, _, err := s.client.ACL().PolicyCreate(&api.ACLPolicy{
+		Name:  policyName,
+		Rules: string(policyBody),
+	}, nil)
+	if err != nil {
+		if strings.Contains(err.Error(), "already exists") {
+			s.Ui.Warn(fmt.Sprintf("[ ] policy %s already exists", policyName))
+		}
```
lgfa29, line 419: Do we want to return early here like it's being done in the role function above?

command/setup_consul.go
```diff
@@ -0,0 +1,449 @@
...
+	_, _, err := s.client.ACL().BindingRuleCreate(rule, nil)
+	if err != nil {
+		return fmt.Errorf("[✘] could not create Consul binding rule: %w", err)
+	}
+
+	s.Ui.Info(fmt.Sprintf("[✔] Created binding rule for auth method %s", rule.AuthMethod))
+	return nil
+}
+
+func (s *SetupConsulCommand) createRoleForTemplate() error {
```
lgfa29, line 393:
```suggestion
func (s *SetupConsulCommand) createRoleForTasks() error {
```

I think?

command/setup_consul.go
```diff
@@ -0,0 +1,449 @@
...
+// askQuestion asks question to user until they provide a valid response.
+func (s *SetupConsulCommand) askQuestion(question string) bool {
+	for {
+		answer, err := s.Ui.Ask(s.Colorize().Color(fmt.Sprintf("[?] %s", question)))
+		if err != nil {
+			if err.Error() != "interrupted" {
+				s.Ui.Output(err.Error())
+			}
+			return false
+		}
```
lgfa29, line 437:
```suggestion
		if err != nil {
			if err.Error() != "interrupted" {
				s.Ui.Output(err.Error())
				os.Exit(1)
			}
			os.Exit(0)
		}
```

If we don't exit here it's impossible to `ctrl-C` the command since it will interpreted as a `no`.

command/setup_consul.go
```diff
@@ -0,0 +1,449 @@
...
+
+		switch strings.TrimSpace(strings.ToLower(answer)) {
+		case "", "n", "no":
+			return false
+		case "y", "yes":
+			return true
```
lgfa29, line 443: The questions have `[Y/n]` which usually indicate that `Y` is the default, but `case "", "n", "no"` causes `n` to be the default.

command/setup_consul.go
```diff
@@ -0,0 +1,449 @@
...
+		default:
+			s.Ui.Output(fmt.Sprintf("%s is not a valid response, please answer \"yes\" or \"no\".", answer))
```
lgfa29, line 445:
```suggestion
			s.Ui.Output(fmt.Sprintf(`%q is not a valid response, please answer "yes" or "no".`, answer))
```

lgfa29 (changes requested):
Looking great! This will be super helpful for quick local testing. I just left some comments to improve the UX and some bugs I found.

command/setup_consul.go
```diff
@@ -0,0 +1,449 @@
...
+
+//go:embed asset/consul-wi-default-auth-method-config.json
+var authConfigBody []byte
+
+//go:embed asset/consul-wi-default-policy.hcl
+var policyBody []byte
+
+const (
+	authMethodServices = "nomad-services"
+	authMethodTasks    = "nomad-tasks"
```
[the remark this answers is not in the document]
pkazmierczak: 😬 yeah...

command/setup_consul.go
```diff
@@ -0,0 +1,449 @@
...
+		createAuthMethod = true
+	}
+
+	if createAuthMethod {
+		err = s.createAuthMethod(authMethodServices, authMethodConf)
+		if err != nil {
+			s.Ui.Error(err.Error())
+			return 1
+		}
+	}
```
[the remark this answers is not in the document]
pkazmierczak: My reasoning was: if users reply `n` this indicates they either set some parts up themselves or prefer to set them up themselves. I think otherwise it makes no sense to have the interactive helper at all.

command/setup_consul.go
pkazmierczak: ah, this is copy pasta from another command, my bad.

command/setup_consul.go
```diff
@@ -0,0 +1,449 @@
...
+  # Nomad agents still need a Consul token in order to register themselves
+  # for automated clustering. It is recommended to set the token using the
+  # CONSUL_HTTP_TOKEN environment variable instead of writing it in the
+  # configuration file.
+
+  # Consul mTLS configuration.
+  # ssl       = true
+  # ca_file   = "/var/ssl/bundle/ca.bundle"
+  # cert_file = "/etc/ssl/consul.crt"
+  # key_file  = "/etc/ssl/consul.key"
```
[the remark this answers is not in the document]
pkazmierczak: I think it's ok to omit the mTLS config for this helper.

pkazmierczak:
@tgross @lgfa29 I addressed all the comments, mind giving it another look?

command/setup_consul.go
```diff
@@ -0,0 +1,485 @@
...
+	// check if we're connecting to Consul ent
+	if _, err := s.client.Operator().LicenseGet(nil); err == nil {
+		s.consulEnt = true
+	}
+
+	authMethodMsg := `
+We need to create two JWT auth methods: one for Nomad services, and one for
+Nomad tasks. The method for services will be called %v and the
+method for tasks %v, and they will both be of jwt type. They
+will share the following config:`
```
tgross, line 149: This is a minor fit-and-finish issue, but these lines aren't evenly wrapped.

command/setup_consul.go
```diff
@@ -0,0 +1,485 @@
...
+You need to adjust your Nomad client configuration in the following way:
+
+consul {
+  enabled = true
+  address = "<Consul address>"
+
+  # Nomad agents still need a Consul token in order to register themselves
+  # for automated clustering. It is recommended to set the token using the
+  # CONSUL_HTTP_TOKEN environment variable instead of writing it in the
+  # configuration file.
```
tgross, line 299: :+1: I like how this text sets users up for a more secure approach.

command/setup_consul.go
```diff
@@ -0,0 +1,485 @@
...
+			existingNamespaces,
+			func(n *api.Namespace) bool { return n.Name == consulNamespace }) {
+			s.Ui.Warn(fmt.Sprintf("[ ] namespace %s already exists", consulNamespace))
+			return nil
+		}
+	}
+
+	_, _, err := nsClient.Create(namespace, nil)
+	if err != nil {
+		return fmt.Errorf("[✘] could not write namespace %q: %w", consulNamespace, err)
```
tgross, line 401: Another minor fit-and-finish issue: we're inconsistent about using `%q` and `%s` for the user-supplied named. It might be nice if we stuck to one or the other throughout this command.

tgross (approved):
LGTM! I've left a few comments around specific bits of text but the overall workflow looks great.

command/setup_consul.go
pkazmierczak: That's for a reason. I actually calculated the characters of method names that get inserted in this message to get perfect alignment. (I am not joking)

tgross: Ah, perfect! Ship it! :grinning:

command/setup_consul.go
pkazmierczak: good catch. Fixed in [6192f39](https://github.com/hashicorp/nomad/pull/18820/commits/6192f39c9ba160a73fbf37ae6b02f8475abb3026)

command/setup_consul.go
```diff
@@ -0,0 +1,485 @@
...
+		err = s.createPolicy()
+		if err != nil {
+			s.Ui.Error(err.Error())
+			return 1
+		}
+	}
+
+	s.Ui.Output(`
+Finally, we need to create a role role-nomad-tasks associated with the policy
+above.`)
```
lgfa29, line 267:
```suggestion
And finally, we will create an ACL role called %q associated with the policy
above.`, roleTasks)
```

command/setup_consul.go
```diff
@@ -0,0 +1,485 @@
...
+
+	var createRole bool
+	if !s.autoYes {
+		createRole = s.askQuestion(
+			"Should we create the above role in your Consul cluster? [Y/n]",
```
lgfa29, line 272:
```suggestion
			"Create the above role in your Consul cluster? [Y/n]",
```

command/setup_consul.go
```diff
@@ -0,0 +1,485 @@
...
+		if err != nil {
+			s.Ui.Error(err.Error())
+			return 1
+		}
+	}
+
+	s.Ui.Output(`
+Finally, we need to create a role role-nomad-tasks associated with the policy
+above.`)
+
```
lgfa29, line 268: I think we're missing printing the role information.

command/setup_consul.go
```diff
@@ -0,0 +1,485 @@
...
+			s.Ui.Error(err.Error())
+			return 1
+		}
+	}
+
+	servicesBindingRule := &api.ACLBindingRule{
+		Description: "binding rule for Nomad services w/ (WI)",
+		AuthMethod:  authMethodServices,
+		BindType:    "service",
+		BindName:    "${value.nomad_namespace}-${value.nomad_service}",
```
lgfa29, line 201:
```suggestion
		BindName:    "${value.nomad_service}",
```

Explained in https://github.com/hashicorp/nomad/pull/18820#discussion_r1373773861.

command/setup_consul.go
```diff
@@ -0,0 +1,485 @@
...
+	if s.consulEnt {
+		method.NamespaceRules = []*api.ACLAuthMethodNamespaceRule{{
+			Selector:      "",
+			BindNamespace: "${value.nomad_namespace}",
+		}}
+	}
+
+	existingMethods, _, _ := s.client.ACL().AuthMethodList(nil)
+	if len(existingMethods) > 0 {
+		if slices.ContainsFunc(
```
lgfa29, line 362:
```suggestion
		if slices.ContainsFunc(
```
I think we can omit this to make the code a bit easier to read? `ContainsFunc` seems to range over the slice, which will be a no-op on an empty slice anyway.

command/setup_consul.go
```diff
@@ -0,0 +1,485 @@
...
+			func(m *api.ACLAuthMethodListEntry) bool { return m.Name == method.Name }) {
+			s.Ui.Warn(fmt.Sprintf("[ ] auth method with name %s already exists", method.Name))
+			return nil
+		}
+	}
+
+	_, _, err := s.client.ACL().AuthMethodCreate(method, nil)
+	if err != nil {
+		if strings.Contains(err.Error(), "error checking JWKSURL") {
+			s.Ui.Error("error: Nomad JWKS endpoint unreachable, is your Nomad server running and is the JWKS url set correctly?")
```
lgfa29, line 373:
```suggestion
			s.Ui.Error("error: Nomad JWKS endpoint unreachable, verify that Nomad is running and that the JWKS URL %s is reachable by Consul", s.jwksURL)
```

command/setup_consul.go
```diff
@@ -0,0 +1,485 @@
...
+			BindNamespace: "${value.nomad_namespace}",
+		}}
+	}
+
+	existingMethods, _, _ := s.client.ACL().AuthMethodList(nil)
+	if len(existingMethods) > 0 {
+		if slices.ContainsFunc(
+			existingMethods,
+			func(m *api.ACLAuthMethodListEntry) bool { return m.Name == method.Name }) {
+			s.Ui.Warn(fmt.Sprintf("[ ] auth method with name %s already exists", method.Name))
```
lgfa29, line 365:
```suggestion
			s.Ui.Warn(fmt.Sprintf("[ ] auth method with name %q already exists", method.Name))
```

lgfa29 (commented):
Another round of me being annoying with details 😅

command/setup_consul.go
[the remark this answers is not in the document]
lgfa29: But if they use mTLS then this config will be incomplete and not work. Since we have access to the Consul client config we know exactly what needs to be set in the file.

Using a go template here could be useful, so something like this (totally untested 😅):

```
consul {
  enabled = true
  address = "{{.Address}}"

  # Nomad agents still need a Consul token in order to register themselves
  # for automated clustering. It is recommended to set the token using the
  # CONSUL_HTTP_TOKEN environment variable instead of writing it in the
  # configuration file.

  {{if .Namespace)}}
  namemespace = "{{.Namespace}}"
  {{end}}

  {{if (eq .Scheme "https")}}
  ssl       = true
  {{end}}
  {{if .TLSConfig.CAFile}}
  ca_file   = "{{.TLSConfig.CAFile}}"
  {{end}}
  {{if .TLSConfig.CertFile}}
  cert_file = "{{.TLSConfig.CertFile}}"
  {{end}}
  {{if .TLSConfig.CertFile}}
  key_file  = "{{.TLSConfig.KeyFile}}"
  {{end}}
  {{if .TLSConfig.InsecureSkipVerify}}
  verify_ssl  = "{{.TLSConfig.KeyFile}}"
  {{end}}
}
```

And then you can just feed it the Consul client config to get the correct Nomad config file.

Bonus point if you run the output through [`hclwrite.Format()`](https://pkg.go.dev/github.com/hashicorp/hcl2@v0.0.0-20191002203319-fb75b3253c80/hclwrite#Format) 😄

command/setup_consul.go
```diff
@@ -0,0 +1,485 @@
...
+		}
+		err = s.createAuthMethod(authMethodTasks, authMethodConf)
+		if err != nil {
+			s.Ui.Error(err.Error())
+			return 1
+		}
+	}
+
+	servicesBindingRule := &api.ACLBindingRule{
+		Description: "binding rule for Nomad services w/ (WI)",
```
lgfa29, line 198:
```suggestion
		Description: "Binding rule for Nomad services authenticated using a workload identity",
```

command/setup_consul.go
```diff
@@ -0,0 +1,485 @@
...
+		AuthMethod:  authMethodServices,
+		BindType:    "service",
+		BindName:    "${value.nomad_namespace}-${value.nomad_service}",
+	}
+
+	tasksBindingRule := &api.ACLBindingRule{
+		Description: "binding rule for Nomad tasks w/ (WI)",
```
lgfa29, line 205:
```suggestion
		Description: "Binding rule for Nomad tasks authenticated using a workload identity",
```

command/setup_consul.go
```diff
@@ -0,0 +1,485 @@
...
+
+	if createAuthMethods {
+		err = s.createAuthMethod(authMethodServices, authMethodConf)
+		if err != nil {
+			s.Ui.Error(err.Error())
+			return 1
+		}
+		err = s.createAuthMethod(authMethodTasks, authMethodConf)
+		if err != nil {
+			s.Ui.Error(err.Error())
```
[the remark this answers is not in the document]
pkazmierczak: `s.createAuthMethod` already returns an informative error.

command/setup_consul.go
```diff
@@ -0,0 +1,485 @@
...
+
+	if createBindingRules {
+		err = s.createBindingRules(servicesBindingRule)
+		if err != nil {
+			s.Ui.Error(err.Error())
+			return 1
+		}
+		err = s.createBindingRules(tasksBindingRule)
+		if err != nil {
+			s.Ui.Error(err.Error())
```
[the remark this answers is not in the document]
pkazmierczak: `s.createBindingRules` already returns an informative error.

command/setup_consul.go
pkazmierczak: a role is just a role name we print above and a link to a policy.

command/setup_consul.go
```diff
@@ -0,0 +1,485 @@
...
+
+	var createAuthMethods bool
+	if !s.autoYes {
+		createAuthMethods = s.askQuestion("Create these auth methods in your Consul cluster? [Y/n]")
+	} else {
+		createAuthMethods = true
+	}
+
+	if s.consulEnt {
+		err = s.createNamespace()
```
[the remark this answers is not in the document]
pkazmierczak: We create it only when it doesn't exists, `s.createNamespace` checks that. We don't _need_ to create a namespace, but I thought this was considered good practice for consul enterprise to bind auth methods to a namespace?

command/setup_consul.go
```diff
@@ -0,0 +1,483 @@
...
+	helpText := `
+Usage: nomad setup consul [options]
+
+  This command sets up Consul for allowing Nomad workloads to authenticate
+  themselves using Workload Identity.
+
+  This command requires acl:write permissions for Consul and respects
+  CONSUL_HTTP_TOKEN, CONSUL_HTTP_ADDR, and other Consul-related
+  environment variables as documented in
+  https://developer.hashicorp.com/nomad/docs/runtime/environment#summary.
```
tgross, line 61: At one point we were discussing having `setup consul` and `setup vault` be specifically called out as experimental/non-stable interfaces, so that we could iterate on them more freely in the near future as users hit stumbling blocks during migration. Do we still want to do that here?

pkazmierczak:
@lgfa29 I addressed all your comments and added a "cleanup" feature where we offer the possibility of removing all the configuration we created if users decide to stop ("atomicity").

command/setup_consul.go
pkazmierczak: good thinking, I added a warning in [5d12362](https://github.com/hashicorp/nomad/pull/18820/commits/5d12362b106b9d43415e0d949dad7b119fee0284)

command/setup.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// Synopsis satisfies the cli.Command Synopsis function.
+func (a *SetupCommand) Synopsis() string { return "Interact with setup helpers" }
+
+// Name returns the name of this command.
+func (a *SetupCommand) Name() string { return "setup" }
+
+// Run satisfies the cli.Command Run function.
+func (a *SetupCommand) Run(_ []string) int { return cli.RunResultHelp }
```
lgfa29, line 39:
```suggestion
// Synopsis satisfies the cli.Command Synopsis function.
func (c *SetupCommand) Synopsis() string { return "Interact with setup helpers" }

// Name returns the name of this command.
func (c *SetupCommand) Name() string { return "setup" }

// Run satisfies the cli.Command Run function.
func (c *SetupCommand) Run(_ []string) int { return cli.RunResultHelp }
```

command/setup_consul.go
```diff
@@ -0,0 +1,736 @@
...
+	client    *api.Client
+	clientCfg *api.Config
+
+	jwksURL string
+
+	consulEnt bool
+	autoYes   bool
+
+	// store IDs of objects we may need to cleanup
+	policyID       string
```
lgfa29, line 53:
```suggestion
```

I think we don't need this anymore. I avoid using the ID because this doesn't persist between runs, so the policy was not being cleaned up.

command/setup_consul.go
```diff
@@ -0,0 +1,736 @@
...
+	bindingRuleIDs []string
```
lgfa29, line 54: Which makes me think about this one...I suspect we may be "leaking" the binding rules between runs 🤔 

Maybe we could use the `AuthMethod` value to find the ones we need to delete?

command/setup_consul.go
```diff
@@ -0,0 +1,736 @@
...
+}
+
+func (s *SetupConsulCommand) handleNo() {
+	s.Ui.Warn(`
+By answering "no" to any of these questions, you are risking an incorrect Consul
+cluster configuration. Nomad workloads with Workload Identity will not be able
+to authenticate unless you create missing configuration yourself.
+`)
+
+	if !s.autoYes && s.askQuestion("Remove everything this command creates? [Y/n]") {
```
lgfa29, line 663: Can we list the things that will be deleted here? It could be a static list like:

```
[?] Remove everything this command creates? [Y/n]

The following items will be deleted if they exist:
  * Auth methods "nomad-services" and "nomad-tasks".
  * Binding rules associated with these methods.
  * Role named "role-nomad-task"
  * Policy named "nomad-workloads".
```

Or a dynamic list, where we first collect what needs to be deleted and output exactly what will be removed.

```
[?] Remove everything this command creates? [Y/n]

The following items will be deleted:
  * Auth methods "nomad-services" and "nomad-tasks".
  * Binding rules "ABC" and "XYZ".
  * Policy named "nomad-workloads.
```

command/setup_consul.go
```diff
@@ -0,0 +1,736 @@
...
+Setup Consul options:
+
+  -jwks-url <url>
+    URL of Nomad's JWKS endpoint contacted by Consul to verify JWT
+    signatures. Defaults to http://localhost:4646/.well-known/jwks.json.
+
+  -y
+    Automatically answers "yes" to all the questions, making the setup
+    non-interactive. Defaults to "false".
+
```
lgfa29, line 82: It could be useful to have a flag like `-cleanup` or `-remove` that just calls `handleNo()`. Otherwise it's impossible to remove things once the setup completes because it just run though automatically 😅

command/setup_consul.go
pkazmierczak: fixed in [d04a235](https://github.com/hashicorp/nomad/pull/18820/commits/d04a235d90260a8dbac816af91512d5e5a20d93f)

command/setup_consul.go
pkazmierczak: fixed in [d04a235](https://github.com/hashicorp/nomad/pull/18820/commits/d04a235d90260a8dbac816af91512d5e5a20d93f)

command/setup_consul.go
pkazmierczak: implemented in [599150e](https://github.com/hashicorp/nomad/pull/18820/commits/599150edb88dd997338d83c51c719014768a12fc)

command/setup_consul.go
pkazmierczak: implemented in [599150e](https://github.com/hashicorp/nomad/pull/18820/commits/599150edb88dd997338d83c51c719014768a12fc)

lgfa29 (approved):
Push some last minor changes, but this command is great! Thank you for the patience 😄

PR #18821, opened 2023-10-20
sids_hook: fix check for Consul token derived from WI

tgross:
The `sids_hook` serves the legacy Connect workflow, and we want to bypass it when using workload identities. So the hook checks that there's not already a Consul token in the alloc hook resources derived from the Workload Identity. This check was looking for the wrong key. This would cause the hook to ignore the Consul token we already have and then fail to derive a SI token unless the Nomad agent has its own token with `acl:write` permission.

Fix the lookup and add tests covering the bypass behavior.


lgfa29 (approved):

pkazmierczak (approved):
LGTM!

PR #18822, opened 2023-10-21
test: add E2E `vaultcompat` test for JWT auth flow

lgfa29:
Test the JWT auth flow using real Nomad and Vault agents.


tgross (approved):
This looks great @lgfa29! It looks like there's a minor issue to mop-up with the changes in `testutil`, but LGTM once that's fixed.

lgfa29:
Ops, syncing Nomad an Vault log level sounded like a good idea when running it locally but I didn't realize that `off` was used in CI 😅 

I capped the log level for Vault to `error` but also added another env var to allow setting a log level specific for Vault if necessary.

Issue #17742, opened 2023-06-26
Document and/or implement ordering in /jobs/:id/allocations


nh2:
The [`/v1/job/:job_id/allocations`](https://developer.hashicorp.com/nomad/api-docs/v1.5.x/jobs#list-job-allocations) docs do not say in which order allocations are returned.

They also don't seem to return allocations in any particular order; they look random.

It would be great to have these ordered to make it easier to e.g. find the logs of the most recent allocation.

In contrast [`/v1/allocations`](https://developer.hashicorp.com/nomad/api-docs/v1.5.x/allocations) documents an order (in the `reverse` parameter).

lgfa29:
Hi @nh2 👋 

In general endpoints return data ordered by their identifier (usually the `ID` UUID). This is documented in the general API docs: https://developer.hashicorp.com/nomad/api-docs#ordering

Custom sorting per-endpoint (like in `/v1/allocations`) is a fairly new feature, and we hope to expand it to more endpoints. I thought we have another issue for this, but I can't find it so I'm leaving this one open as it 😄

PR #18824, opened 2023-10-21
Add reverse query param while fetching job allocations

Sanskar531:
#17742

Job allocations can now be reverse sorted. I have also updated the docs for it.

### DIFF
--- a/website/content/api-docs/jobs.mdx
+++ b/website/content/api-docs/jobs.mdx
@@ -1246,7 +1246,7 @@ The table below shows this endpoint's support for
 enabled, this value must match a namespace that the token is allowed to
 access. This is specified as a query string parameter.
 
-- `reverse` `(bool: false)` - Specifies the list of returned allocations to be sorted in the reverse lexicographical order. By default, they are sorted by lexicographical order.
+- `reverse` `(bool: false)` - Specifies the list of returned allocations to be sorted in the reverse lexicographical order. By default, they are lexicographically sorted.
 
 ### Sample Request
 
--- a/website/content/api-docs/jobs.mdx
+++ b/website/content/api-docs/jobs.mdx
@@ -1246,6 +1246,8 @@ The table below shows this endpoint's support for
 enabled, this value must match a namespace that the token is allowed to
 access. This is specified as a query string parameter.
 
+- `reverse` `(bool: false)` - Specifies the list of returned allocations to be sorted in the reverse lexicographical order. By default, they are sorted by lexicographical order.
+
 ### Sample Request
 
 ```shell-session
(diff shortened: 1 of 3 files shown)


Sanskar531:
I am not sure if this is the right way to do it. Other implementations like the one in `DeploymentsByIDPrefix` , sort it directly when fetching the data from `memdb`. Doing so, is a big change though as there are 100 references of where `AllocsByJob` is used and I will have to add a new `sort` parameter to the function. If it's okay, I don't mind changing this pr to update all references of `AllocsByJob`.

website/content/api-docs/jobs.mdx
```diff
@@ -1246,6 +1246,8 @@ The table below shows this endpoint's support for
 enabled, this value must match a namespace that the token is allowed to
 access. This is specified as a query string parameter.
 
+- `reverse` `(bool: false)` - Specifies the list of returned allocations to be sorted in the reverse lexicographical order. By default, they are lexicographically sorted.
```
tgross, line 1249:
```suggestion
Added "by ID", and also we hard-wrap markdown text:

```suggestion
- `reverse` `(bool: false)` - Specifies the list of returned allocations to be
  sorted in the reverse lexicographical order. By default, they are
  lexicographically sorted by ID.
```
```

tgross (changes requested):
Hi @Sanskar531! Thanks for this PR! As you've noted in your PR description, we should be pushing the reverse parameter down into the state query. The reason is because of pagination. For example, suppose I have the following allocations:

* `aaaaaaaa-71ba-11ee-950b-bfbf31e28f07`
* `bbbbbbbb-71ba-11ee-b9ea-af44d029c7fa`
* `cccccccc-71ba-11ee-80f2-4be2f9fab70e`
* `dddddddd-71ba-11ee-81a1-db53adff8b79`
* `eeeeeeee-71ba-11ee-976a-3f503a3105b0`

And then I make a query with `per_page=3` and `reverse=true`. The way you've got it here, we'd return the following:

* `cccccccc-71ba-11ee-80f2-4be2f9fab70e`
* `bbbbbbbb-71ba-11ee-b9ea-af44d029c7fa`
* `aaaaaaaa-71ba-11ee-950b-bfbf31e28f07`

But what we really want to return is:

* `eeeeeeee-71ba-11ee-976a-3f503a3105b0`
* `dddddddd-71ba-11ee-81a1-db53adff8b79`
* `cccccccc-71ba-11ee-80f2-4be2f9fab70e`

Once you've got that done I'll be happy to re-review. Also, can you run `make cl` to add a changelog entry?

Sanskar531:
Hey! Thanks for the guidance. That makes a lot of sense! 

So just confirming before I rework this that I can make the change to the function signature of  `AllocsByJob` by adding a new `reverse` parameter and subsequently updating all the usages of it (I found 100 references) right?

tgross:
> So just confirming before I rework this that I can make the change to the function signature of `AllocsByJob` by adding a new `reverse` parameter and subsequently updating all the usages of it (I found 100 references) right?

Yeah, that looks like the right approach. But there's a `state.SortOption` type you can use for clarity, rather than a simple bool parameter.

PR #18832, opened 2023-10-23
cli: ensure HCL env vars are added to the job submission object.

### LINKED ISSUE #18805: Display all provided variables in UI
### Proposal
The Nomad UI should show all variable values in the UI under `HCL Variable Values`. Currently, it only shows the content of `-var-file` (5d6409bf?), not `NOMAD_VAR_foo` environment variables. AFAIK `-var 'key=value'` isn't shown as well.

image

### Use-cases
When one uses env vars or `-var` instead of a var file it is tedious to edit a Job Definition in the UI, since one has to provide all variable values again. It would be great if all variables are shown.

jrasell:
closes #18805

### DIFF
--- a/command/helpers.go
+++ b/command/helpers.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"maps"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -534,6 +535,10 @@ func (j *JobGetter) Get(jpath string) (*api.JobSubmission, *api.Job, error) {
 			return nil, nil, fmt.Errorf("Failed to parse HCL job: %w", err)
 		}
 
+		// Perform the environment listing here as it is used twice beyond this
+		// point.
+		osEnv := os.Environ()
+
 		// we are parsing HCL2, whether from a file or stdio
 		jobStruct, err = jobspec2.ParseWithConfig(&jobspec2.ParseConfig{
 			Path:     pathName,
@@ -541,7 +546,7 @@ func (j *JobGetter) Get(jpath string) (*api.JobSubmission, *api.Job, error) {
 			ArgVars:  j.Vars,
 			AllowFS:  true,
 			VarFiles: j.VarFiles,
-			Envs:     os.Environ(),
+			Envs:     osEnv,
 			Strict:   j.Strict,
 		})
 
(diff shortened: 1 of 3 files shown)


command/helpers_test.go
```diff
@@ -685,3 +685,19 @@ func Test_extractVarFlags(t *testing.T) {
...
+
+func Test_extractJobSpecEnvVars(t *testing.T) {
+	ci.Parallel(t)
+
+	t.Run("nil", func(t *testing.T) {
+		must.MapEmpty(t, extractJobSpecEnvVars(nil))
+	})
+
+	t.Run("complete", func(t *testing.T) {
+		result := extractJobSpecEnvVars([]string{"NOMAD_VAR_count=13", "GOPATH=/Users/jrasell/go", "NOMAD_VAR_image=redis:7"})
```
shoenig, line 697: can you add a test where there is no value, e.g. `BLAH=` because that's totally a thing to be aware of in parsing env vars

command/helpers.go
```diff
@@ -606,6 +620,36 @@ func extractVarFlags(slice []string) map[string]string {
...
+	m := make(map[string]string)
+
+	for _, raw := range envVars {
+		if !strings.HasPrefix(raw, jobspec2.VarEnvPrefix) {
+			continue
+		}
+
+		// Trim the prefix, so we just have the raw key=value variable
+		// remaining.
+		raw = raw[len(jobspec2.VarEnvPrefix):]
```
gulducat, line 638: `NOMAD_VAR_=blah` makes this `raw = "=blah"` so produces `m[""] = "blah"` below -- do we care to handle that edge case?

the non-env version looks like this

```
$ nomad run -var '=nah' vars.nomad.hcl
Error getting job struct: Error parsing job file from vars.nomad.hcl:
<nil>: Undefined -var variable; A "" variable was passed in the command line but was not found in known variables. To declare variable "", place this block in your job file
<nil>: Unset variable "one"; A used variable must be set or have a default value; see https://www.nomadproject.io/docs/job-specification/hcl2/variables for details.
```

which arguably is a kinda silly error since "" is not a valid variable.  maybe a silent no-op, as happens here, is preferable?

command/helpers_test.go
gulducat, line 697: nit: I find it easier to read these with one var per line

```suggestion
		result := extractJobSpecEnvVars([]string{
			"NOMAD_VAR_count=13",
			"GOPATH=/Users/jrasell/go",
			"NOMAD_VAR_image=redis:7",
		})
```

gulducat (approved):
:+1:

command/helpers_test.go
gulducat: fun that you and I honed in on emptiness on opposite sides of the `=` :yum:

shoenig (approved):
LGTM!

gulducat (approved):

Issue #18804, opened 2023-10-19
Attempting to increment Prometheus counter nomad_client_host_cpu_total_ticks_count with value negative value


lhossack:
### Nomad version
```
$ nomad version
Nomad v1.6.1
BuildDate 2023-07-21T13:49:42Z
Revision 515895c7690cdc72278018dc5dc58aca41204ccc
```

### Operating system and Environment details
```
$ uname -a
Linux <hostname> 4.19.0-20-amd64 #1 SMP Debian 4.19.235-1 (2022-03-17) x86_64 GNU/Linux
```
Hardware:
UCSB-B200-M4
UCSB-B200-M5

### Issue
Nomad is occasionally reporting negative cpu tick count since last epoch (see client log/ error messages below). 

I believe this is likely benign, but does generate several logs over extended periods of time (on average ~24 log messages/ day/ host). 

### Reproduction steps
Run the nomad service; possibly for several hours/ days and check the logs for the error message

Rarely, I also get a response to 
```
nomad node status -self
``` 
like this
```
Host Resource Utilization
CPU              Memory           Disk
-472/120000 MHz  6.8 GiB/376 GiB  (/dev/mapper/encryptedvol)
```

#### Expected Result
Time/ ticks should move forward (this number should be positive, or at least 0)

#### Actual Result
Attempting to increment Prometheus counter   nomad_client_host_cpu_total_ticks_count with value negative value -1310.6796

### Nomad Client logs (if appropriate)
Oct 19 04:40:31 c2007   nomad[60804]: agent: Attempting to increment Prometheus counter   nomad_client_host_cpu_total_ticks_count with value negative value -1310.6796
Oct 19 04:40:31 c2007 nomad[60804]:   2023-10-19T04:40:31.905Z [ERROR] agent: Attempting to increment Prometheus   counter nomad_client_host_cpu_total_ticks_count with value negative value   -1310.6796
Oct 19 04:40:31 c2007 nomad[60804]:   agent: Attempting to increment Prometheus counter   nomad_client_host_cpu_total_ticks_count with value negative value -414.2857
Oct 19 04:40:31 c2007 nomad[60804]:   2023-10-19T04:40:31.900Z [ERROR] agent: Attempting to increment Prometheus   counter nomad_client_host_cpu_total_ticks_count with value negative value   -414.2857
[...]

PR #18835, opened 2023-10-23
metrics: prevent negative counter from iowait decrease

tgross:
The iowait metric obtained from `/proc/stat` can under some circumstances decrease. The relevant condition is when an interrupt arrives on a different core than the one that gets woken up for the IO, and a particular counter in the kernel for that core gets decremented. This is documented in the man page for the `proc(5)` pseudo-filesystem, and considered an unfortunate behavior that can't be changed for the sake of ABI compatibility.

In Nomad, we get the current "busy" time (everything except for idle) and compare it to the previous busy time to get the counter increment. If the iowait counter decreases and the idle counter increases more than the increase in the total busy time, we can get a negative total. This previously caused a panic in our metrics collection (see #15861) but that is being prevented by reporting an error message.

Fix the bug by putting a zero floor on the values we return from the host CPU stats calculator.

Fixes: #15861
Fixes: #18804

Note for reviewers: this code has been moved around quite a bit in `main` since 1.6.x, so I'll need to do a separate PR for the backports.

### DIFF
--- a/.changelog/18835.txt
+++ b/.changelog/18835.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+metrics: Fixed a bug where CPU counters could report errors for negative values
+```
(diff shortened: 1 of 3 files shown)


shoenig (approved):
LGTM! and nice find :lady_beetle:

PR #18843, opened 2023-10-24
numa: fix cpu topology conversion for non linux systems

shoenig:


tgross (approved):
LGTM! 

Thanks for the pointers on the file moves... it makes it easier to review side-by-side :grinning:

PR #18845, opened 2023-10-24
cl: minor features are listed as improvements

shoenig:
The Features header is reserved for "tent-pole" features of a Nomad version.


gulducat (approved):
:+1:

PR #18849, opened 2023-10-24
metrics: prevent negative counter from iowait decrease (backport 1.6.x)

tgross:
The iowait metric obtained from `/proc/stat` can under some circumstances decrease. The relevant condition is when an interrupt arrives on a different core than the one that gets woken up for the IO, and a particular counter in the kernel for that core gets interrupted. This is documented in the man page for the `proc(5)` pseudo-filesystem, and considered an unfortunate behavior that can't be changed for the sake of ABI compatibility.

In Nomad, we get the current "busy" time (everything except for idle) and compare it to the previous busy time to get the counter incremeent. If the iowait counter decreases and the idle counter increases more than the increase in the total busy time, we can get a negative total. This previously caused a panic in our metrics collection (see #15861) but that is being prevented by reporting an error message.

Fix the bug by putting a zero floor on the values we return from the host CPU stats calculator.

Backport-of: #18835 (the code touched here was moved, so this backport had to be done by hand)

### DIFF
--- a/.changelog/18835.txt
+++ b/.changelog/18835.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+metrics: Fixed a bug where CPU counters could report errors for negative values
+```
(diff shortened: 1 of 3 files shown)


shoenig (approved):
LGTM!

tgross:
Backported manually to 1.5.x and 1.4.x.

PR #18856, opened 2023-10-24
docs: deprecate rsadecrypt

schmichael:
`rsadecrypt` uses `PKCS #1 v1.5` padding which has multiple known weaknesses. While it is possible to use safely in Nomad, we should not encourage our users to use bad cryptographic primitives.

If users want to decrypt secrets in jobspecs we should choose a cryptographic primitive designed for that purpose. `rsadecrypt` was inherited from Terraform which only implemented it to support decrypting Window's passwords on AWS EC2 instances:

https://github.com/hashicorp/terraform/pull/16647

This is not something that should ever be done in a jobspec, therefore there's no reason for Nomad to support this HCL2 function.

Preview to save you a few clicks: 
image

### DIFF
--- a/website/content/docs/job-specification/hcl2/functions/crypto/rsadecrypt.mdx
+++ b/website/content/docs/job-specification/hcl2/functions/crypto/rsadecrypt.mdx
@@ -6,11 +6,17 @@ description: The rsadecrypt function decrypts an RSA-encrypted message.
 
 # `rsadecrypt` Function
 
--> **Deprecated:** This function is deprecated and may be removed in the
-future. While it is possible to use safely, it encourages use of cryptographic
-primitives with [known weaknesses][attacks]. [Nomad Variables][nvars] and
-[HashiCorp Vault][vault] are the recommended ways to provide secrets to
-workloads.
+<Note title="Deprecated">
+This function is deprecated and may be removed in the future.
+</Note>
+
+<Warning>
+While it is possible to use safely, it encourages use of cryptographic
+primitives with <a href="https://en.wikipedia.org/wiki/PKCS_1#Attacks">known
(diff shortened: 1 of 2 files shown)


tgross (commented):
LGTM :+1:

website/content/docs/job-specification/hcl2/functions/crypto/rsadecrypt.mdx
```diff
@@ -6,6 +6,12 @@ description: The rsadecrypt function decrypts an RSA-encrypted message.
 
 # `rsadecrypt` Function
 
+-> **Deprecated:** This function is deprecated and may be removed in the
+future. While it is possible to use safely, it encourages use of cryptographic
+primitives with [known weaknesses][attacks]. [Nomad Variables][nvars] and
+[HashiCorp Vault][vault] are the recommended ways to provide secrets to
+workloads.
```
angrycub, line 13: Should we leverage the newer MDX components for these callouts?

```suggestion
<Note title="Deprecated">
This function is deprecated and may be removed in the future.
</Note>

<!--
The whitespace inside the Warning tag is significant because it contains markdown
-->
<Warning>

While it is possible to use safely, it encourages use of cryptographic
primitives with [known weaknesses][attacks]. [Nomad Variables][nvars] and
[HashiCorp Vault][vault] are the recommended ways to provide secrets to
workloads.

</Warning>

```

schmichael: I love it! Thank you! I had to rewrite the links in HTML, but nbd.

angrycub (approved):
Aww yiss!

PR #18864, opened 2023-10-25
docs: add job-specification docs for numa

shoenig:
Separate from a more comprehensive "CPU Resources" concepts docs we spoke of

### DIFF
--- a/website/content/docs/job-specification/numa.mdx
+++ b/website/content/docs/job-specification/numa.mdx
@@ -15,11 +15,15 @@ Workloads that are sensitive to memory latency can perform significantly better
 when pinned to CPU cores on the same NUMA node.
 
 <EnterpriseAlert>
-  This functionality only exists in Nomad Enterprise. This is not
-  present in the source available version of Nomad.
+
+This functionality only exists in Nomad Enterprise. This is not
+present in the source available version of Nomad.
 </EnterpriseAlert>
 
-~> NUMA aware scheduling is currently limited to Linux.
+<Note>
+
+NUMA aware scheduling is currently limited to Linux.
+</Note>
 
 ```hcl
 job "example" {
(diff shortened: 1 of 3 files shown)


shoenig:
cc @schmichael who had opinions on the wording around fragmentation / suggesting preemption (which I left out)

website/content/docs/job-specification/numa.mdx
```diff
@@ -0,0 +1,76 @@
...
+  on the same hardware NUMA node. If there are no Nomad nodes with a sufficient
+  number of available cores in a compatible configuration, task placement will
+  fail due to exhausted resources.
+
+<Note>
+
+The `require` affinity option should be used sparingly due to
+the implied fragmentation caused by reserving CPU cores based on the NUMA node
+they are associated with. Use it for workloads known to be highly sensitive
+to memory latencies.
```
schmichael, line 72: > ...used sparingly... ...implied fragmentation...

I know people often want more prescriptive guidance from us, but I would rather frame this in terms of the tradeoff being presented than to discourage the use of a feature that could reduce overall resource consumption dramatically. (Assuming if you're avoiding 300% performance penalties due to cross-node latencies, you can run fewer instances of a service to serve the same number of requests.)

So perhaps something like:

> The `require` affinity option may cause workload fragmentation by reserving CPU cores based on the NUMA node they are associated with. Use it for workloads known to be highly sensitive to memory latencies.

Might even be worth defining `workload fragmentation` somewhere as something like:

> A jobspec constraint that prevents optimal binpacking of Clients. This can waste cluster resources by leaving some Client resources free but unusable. For example when `numa.affinity = "require"`, workloads cannot be scheduled on Clients which may have ample free compute resources unless those compute resources happen to be colocated on a single NUMA node.

idk where an appropriate place for that would be though.

schmichael (approved):
LGTM. Left some comments, but not blockers... not even sure they're good ideas. :sweat_smile: 

This is just complex stuff! Not sure if we should keep the docs minimal because presumably folks who go looking for NUMA features already have a strong sense of how this stuff works, or if we should steal more from the RFC that helped educate NUMA-newbies like me.

shoenig:
> LGTM. Left some comments, but not blockers... not even sure they're good ideas. 😅
> 
> This is just complex stuff! Not sure if we should keep the docs minimal because presumably folks who go looking for NUMA features already have a strong sense of how this stuff works, or if we should steal more from the RFC that helped educate NUMA-newbies like me.

I'm working on a Concepts/CPU doc that will cover "everything" to do with how Nomad interacts with your processor. I'm thinking we should keep the jobspec doc fairly minimal and then link to the concepts doc for further reference (once it exists).

schmichael:
> I'm thinking we should keep the jobspec doc fairly minimal and then link to the concepts doc for further reference (once it exists).

Yeah I love this approach and what I plan on doing for workload identity. Leave the reference docs concise so folks can just find what they're looking for quickly. Concept docs provide context. Tutorials provide walkthroughs/howtos. :+1:

PR #18866, opened 2023-10-25
Fix the "Starting" allocations link

lattwood:
Before this commit, it would bring you to the list of allocations
filtered by status=starting. This status does not exist in the Status
drop-down on the Allocations section of a job in the UI.


philrenaud (approved):
Works like a charm. Thanks for finding this, @lattwood !

PR #18867, opened 2023-10-25
consul WI: rename default auth method for services

pkazmierczak:
It should be called `nomad-services` instead of `nomad-workloads`.


tgross (approved):
LGTM :+1:

PR #18873, opened 2023-10-26
scheduler: ensure dup alloc names are fixed before plan submit.

### LINKED ISSUE #10727: NOMAD_ALLOC_INDEX is not always unique within a single service job version
### Nomad version
```
/ # nomad version
Nomad v0.12.10 (6b50c40dc5fc045282ff2a6f978ba7850e43d0d2)
```

CentOS Linux 7 (Core)
3.10.0-1160.24.1.el7.x86_64

### Issue

According to https://github.com/hashicorp/nomad/pull/6830, `NOMAD_ALLOC_INDEX` is supposed to be unique across job versions. However, we have discovered a case in one of our clusters where a job where allocs appear to be on the same job version and have duplicate `NOMAD_ALLOC_INDEX`s:

```
/ # nomad alloc status a511e040 | grep -E "Job Version|^Name"
Name                = a-job.a-task[8]
Job Version         = 2
/ # nomad alloc status 86c0f7a1 | grep -E "Job Version|^Name"
Name                = a-job.a-task[8]
Job Version         = 2
/ #
```

This was discovered by our Prometheus metrics exporter complaining about attempting to ship duplicate metrics. Of interest, this job has a task count of 50, and only 4 out of those 50 are duped:

```
/ # nomad job status a-job | tail -n 50 | awk '{print $1}' | xargs -I {} nomad alloc status {} | grep ^Name | sort | uniq -d
Name                 = a-job.a-task[13]
Name                 = a-job.a-task[15]
Name                 = a-job.a-task[16]
Name                 = a-job.a-task[8]
```

Interestingly enough, we had the job owner redeploy in attempt to fix this situation and we actually see the exact same `NOMAD_ALLOC_INDEX`s duplicated:

```
/ # nomad alloc status 8711bd71 | grep -E "Job Version|^Name"
Name                = a-job.a-task[8]
Job Version         = 3
/ # nomad alloc status df217f1b | grep -E "Job Version|^Name"
Name                = a-job.a-task[8]
Job Version         = 3
/ #

/ # nomad job status a-job | grep running | tail -n 50 | awk '{print $1}' | xargs -I {} nomad alloc status {} | grep -E "Job Version|^Name" | sort | uniq -d
Job Version         = 3
Name                = a-job.a-task[13]
Name                = a-job.a-task[15]
Name                = a-job.a-task[16]
Name                = a-job.a-task[8]
```

### Reproduction steps

Man, I wish I could tell you. We just upgraded the cluster from 0.11.4 to 0.12.10 earlier today so it's probably related to that... We upgraded the server cluster, then rolling upgraded the clients. ~My _guess_ is that the deployment prior to the cluster upgrade failed, and that the rolling client restart caused alloc migrations which started up the replacement allocs on the "new" job version, but~ so far I haven't had the bandwidth to attempt a reproduction.

Edit: Scratch that. Deployment history seems fine:

```
/ # nomad job history a-job
Version     = 2
Stable      = true
Submit Date = 2021-06-08T21:57:04Z

Version     = 1
Stable      = true
Submit Date = 2021-06-08T21:32:10Z

Version     = 0
Stable      = true
Submit Date = 2021-03-23T22:46:13Z
/ #
```

So we put a hack into our exporter to stop the bleeding for the night. But this also means I can grab logs or anything interesting that will help on your end! Let me know what you need and I'll try to get it!

#### Expected Result

`NOMAD_ALLOC_INDEX` should be unique within a job version

#### Actual Result

`NOMAD_ALLOC_INDEX` is not unique within a job version

Nothing appears relevant here

I didn't see anything relevant in here, but let me know if you'd like me to collect them.

jrasell:
This change fixes a bug within the generic scheduler which meant duplicate alloc indexes (names) could be submitted to the plan applier and written to state. The bug originates from the placements calculation notion that names of allocations being replaced are blindly copied to their replacement. This is not correct in all cases, particularly when dealing with canaries.

The fix updates the alloc name index tracker to include minor duplicate tracking. This can be used when computing placements to ensure duplicate are found, and a new name picked before the plan is submitted. The name index tracking is now passed from the reconciler to the generic scheduler via the results, so this does not have to be regenerated, or another data structure used.

closes #10727 

### Reviewer Notes
The new test `TestServiceSched_JobModify_ProposedDuplicateAllocIndex` mimics the reproduction behaviour and can be used to see how the code change has fixed the bug.

[This reproduction](https://github.com/jrasell/dev-mess/tree/main/nomad/development/gh10727) can be used to test the code change in a manual way. The reproduction (in my testing) worked 100% of the time.

### DIFF
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -10679,13 +10679,19 @@ func (a *Allocation) JobNamespacedID() NamespacedID {
 // Index returns the index of the allocation. If the allocation is from a task
 // group with count greater than 1, there will be multiple allocations for it.
 func (a *Allocation) Index() uint {
-	l := len(a.Name)
-	prefix := len(a.JobID) + len(a.TaskGroup) + 2
+	return AllocIndexFromName(a.Name, a.JobID, a.TaskGroup)
+}
+
+// AllocIndexFromName returns the index of an allocation given its name, the
+// jobID and the task group name.
+func AllocIndexFromName(allocName, jobID, taskGroup string) uint {
+	l := len(allocName)
+	prefix := len(jobID) + len(taskGroup) + 2
 	if l <= 3 || l <= prefix {
 		return uint(0)
 	}
 
-	strNum := a.Name[prefix : len(a.Name)-1]
+	strNum := allocName[prefix : len(allocName)-1]
 	num, _ := strconv.Atoi(strNum)
 	return uint(num)
 }
--- a/scheduler/generic_sched.go
+++ b/scheduler/generic_sched.go
@@ -466,7 +466,7 @@ func (s *GenericScheduler) computeJobAllocs() error {
 		s.queuedAllocs[p.placeTaskGroup.Name] += 1
 		destructive = append(destructive, p)
 	}
-	return s.computePlacements(destructive, place)
+	return s.computePlacements(destructive, place, results.taskGroupAllocNameIndexes)
 }
 
 // downgradedJobForPlacement returns the job appropriate for non-canary placement replacement
@@ -508,7 +508,8 @@ func (s *GenericScheduler) downgradedJobForPlacement(p placementResult) (string,
 
 // computePlacements computes placements for allocations. It is given the set of
 // destructive updates to place and the set of new placements to place.
-func (s *GenericScheduler) computePlacements(destructive, place []placementResult) error {
+func (s *GenericScheduler) computePlacements(destructive, place []placementResult, nameIndex map[string]*allocNameIndex) error {
+
 	// Get the base nodes
 	nodes, byDC, err := s.setNodes(s.job)
 	if err != nil {
@@ -530,6 +531,12 @@ func (s *GenericScheduler) computePlacements(destructive, place []placementResul
 		for _, missing := range results {
 			// Get the task group
 			tg := missing.TaskGroup()
+
+			// This is populated from the reconciler via the compute results,
+			// therefore we cannot have an allocation belonging to a task group
+			// that has not generated and been through allocation name index
+			// tracking.
+			taskGroupNameIndex := nameIndex[tg.Name]
 
 			var downgradedJob *structs.Job
 
@@ -628,12 +635,34 @@ func (s *GenericScheduler) computePlacements(destructive, place []placementResul
 					resources.Shared.Ports = option.AllocResources.Ports
 				}
 
+				// Pull the allocation name as a new variables, so we can alter
+				// this as needed without making changes to the original
+				// object.
+				newAllocName := missing.Name()
+
+				// Identify the index from the name, so we can check this
+				// against the allocation name index tracking for duplicates.
+				allocIndex := structs.AllocIndexFromName(newAllocName, s.job.ID, tg.Name)
+
+				// If the allocation index is a duplicate, we cannot simply
+				// create a new allocation with the same name. We need to
+				// generate a new index and use this. The log message is useful
+				// for debugging and development, but could be removed in a
+				// future version of Nomad.
+				if taskGroupNameIndex.IsDuplicate(allocIndex) {
+					oldAllocName := newAllocName
+					newAllocName = taskGroupNameIndex.Next(1)[0]
+					taskGroupNameIndex.UnsetIndex(allocIndex)
+					s.logger.Debug("duplicate alloc index found and changed",
+						"old_alloc_name", oldAllocName, "new_alloc_name", newAllocName)
+				}
+
 				// Create an allocation for this
 				alloc := &structs.Allocation{
 					ID:                 uuid.Generate(),
 					Namespace:          s.job.Namespace,
 					EvalID:             s.eval.ID,
-					Name:               missing.Name(),
+					Name:               newAllocName,
 					JobID:              s.job.ID,
 					TaskGroup:          tg.Name,
 					Metrics:            s.ctx.Metrics(),
--- a/scheduler/generic_sched_test.go
+++ b/scheduler/generic_sched_test.go
@@ -2179,6 +2179,96 @@ func TestServiceSched_JobModify_ProposedDuplicateAllocIndex(t *testing.T) {
 	testHarness.AssertEvalStatus(t, structs.EvalStatusComplete)
 }
 
+func TestServiceSched_JobModify_ExistingDuplicateAllocIndexNonDestructive(t *testing.T) {
+	ci.Parallel(t)
+
+	testHarness := NewHarness(t)
+
+	// Create some nodes
+	var nodes []*structs.Node
+	for i := 0; i < 10; i++ {
+		node := mock.Node()
+		nodes = append(nodes, node)
+		must.NoError(t, testHarness.State.UpsertNode(structs.MsgTypeTestSetup, testHarness.NextIndex(), node))
+	}
+
+	// Generate a fake job with allocations
+	mockJob := mock.MinJob()
+	mockJob.TaskGroups[0].Count = 10
+	must.NoError(t, testHarness.State.UpsertJob(structs.MsgTypeTestSetup, testHarness.NextIndex(), nil, mockJob))
+
+	// Generate some allocations which will represent our pre-existing
+	// allocations. These have aggressive duplicate names.
+	var (
+		allocs   []*structs.Allocation
+		allocIDs []string
+	)
+	for i := 0; i < 10; i++ {
+		alloc := mock.MinAlloc()
+		alloc.Namespace = structs.DefaultNamespace
+		alloc.Job = mockJob
+		alloc.JobID = mockJob.ID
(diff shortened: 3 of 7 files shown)


nomad/structs/structs.go
```diff
@@ -10679,13 +10679,19 @@ func (a *Allocation) JobNamespacedID() NamespacedID {
 // Index returns the index of the allocation. If the allocation is from a task
 // group with count greater than 1, there will be multiple allocations for it.
 func (a *Allocation) Index() uint {
-	l := len(a.Name)
-	prefix := len(a.JobID) + len(a.TaskGroup) + 2
+	return AllocIndexFromName(a.Name, a.JobID, a.TaskGroup)
```
Juanadelacuesta, line 10682: Why is it necessary to create a new function?

jrasell: I needed a way to get the allocation name index where I didn't have the allocation object available but had the full name.

scheduler/reconcile_util.go
```diff
@@ -583,24 +584,36 @@ type allocNameIndex struct {
 	job, taskGroup string
 	count          int
 	b              structs.Bitmap
+
+	// duplicates is used to store duplicate allocation indexes which are
+	// currently present within the index tracking. The map key is the index,
+	// and the current count of duplicates. While the map is only accessed
+	// within a single routine, lock is used for access to provide future-proof
+	// safety at a very small overhead.
```
tgross, line 592: Changing the scheduler goroutine to be multi-threaded would be an enormous change because so much of placement work is serialized on a per-job basis. And having a mutex that we have to lock and unlock potentially many times per scheduler invocation seems fairly expensive relative to the vague future value. I would leave this off.

scheduler/reconcile_util.go
```diff
@@ -671,9 +719,11 @@ func (a *allocNameIndex) NextCanaries(n uint, existing, destructive allocSet) []
...
 	existingNames := existing.nameSet()
 
-	// First select indexes from the allocations that are undergoing destructive
-	// updates. This way we avoid duplicate names as they will get replaced.
-	dmap := bitmapFrom(destructive, uint(a.count))
+	// First select indexes from the allocations that are undergoing
+	// destructive updates. This way we avoid duplicate names as they will get
+	// replaced. As this process already takes into account duplicate checking,
+	// we can discard the duplicate mapping when generating the bitmap.
+	dmap, _ := bitmapFrom(destructive, uint(a.count))
```
tgross, line 726: Good explanatory comment :+1:

scheduler/reconcile.go
```diff
@@ -968,7 +977,16 @@ func (a *allocReconciler) computeStop(group *structs.TaskGroup, nameIndex *alloc
...
+	// Hot path the nothing to do case.
+	//
+	// Duplicate allocation indexes can be caused due to the way this piece of
+	// code works. The reproduction involved canaries, and performing both a
+	// destructive job change (image update) and increasing the group count by
+	// one. In this scenario, once the canary is placed and we compute the next
+	// set of stops, the following maths takes place compared to a canary
+	// deployment that does not generate duplicate indexes.
+	//   NonBuggy: Untainted = "4" and GroupCount = "3" | remove = "1"
+	//   YesBuggy: Untainted = "4" and GroupCount = "4" | remove = "0"
```
tgross, line 989: This comment is good for reviewers today, but seems anchored in the debugging process with notes about the previously buggy code, rather than the current state of the code when you're done. Maybe something more like:

```suggestion
	// Note that this path can result in duplication allocation indexes in a
	// scenario with a destructive job change (ex. image update) happens with an
	// increased group count. Once the canary is replaced and we compute the
	// next set of stops, the untainted set equals the new group count, which
	// results is missing one removal. The duplicate alloc index is corrected in
	// `computePlacements`
```

scheduler/generic_sched_test.go
```diff
@@ -1970,6 +1970,215 @@ func TestServiceSched_JobModify(t *testing.T) {
...
+		}
+		allocs = append(allocs, alloc)
+	}
+	must.NoError(t, testHarness.State.UpsertAllocs(structs.MsgTypeTestSetup, testHarness.NextIndex(), allocs))
+
+	// Generate a job modification which will force a destructive update.
+	mockJob2 := mock.Job()
+	mockJob2.ID = mockJob.ID
+	mockJob2.TaskGroups[0].Tasks[0].Config["command"] = "/bin/other"
+	must.NoError(t, testHarness.State.UpsertJob(structs.MsgTypeTestSetup, testHarness.NextIndex(), nil, mockJob2))
```
tgross, line 2012: Can we add a test for duplicate alloc indexes but with non-destructive updates, to ensure we're not destroying allocations unexpectedly? IIRC you said you tested this manually, but it'd be good to have a test to avoid regressions.

tgross (approved):
Nice work on this @jrasell. The fix is tightly scoped and very clear, and the testing and extra commentary is really helpful.

schmichael (approved):
Great work!

scheduler/reconcile.go
```diff
@@ -968,7 +977,16 @@ func (a *allocReconciler) computeStop(group *structs.TaskGroup, nameIndex *alloc
 		untainted = untainted.difference(canaries)
 	}
 
-	// Hot path the nothing to do case
+	// Hot path the nothing to do case.
+	//
+	// Duplicate allocation indexes can be caused due to the way this piece of
+	// code works. The reproduction involved canaries, and performing both a
```
lgfa29, line 983: Just to make I understood, this hot path can return early with duplicate alloc names? Is there a way to avoid that so that callers don't have to handle them?

lgfa29 (commented):
Sharing some comments early in case I don't have a chance to come back for a deeper review later today, but great work!

scheduler/generic_sched.go
```diff
@@ -628,12 +635,34 @@ func (s *GenericScheduler) computePlacements(destructive, place []placementResul
 					resources.Shared.Ports = option.AllocResources.Ports
 				}
 
+				// Pull the allocation name as a new variables, so we can alter
+				// this as needed without making changes to the original
+				// object.
+				newAllocName := missing.Name()
```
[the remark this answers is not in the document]
jrasell: `newAllocName` is potentially modified on L654 if it is found to be a duplicate.

scheduler/generic_sched.go
```diff
@@ -628,12 +635,34 @@ func (s *GenericScheduler) computePlacements(destructive, place []placementResul
...
+				allocIndex := structs.AllocIndexFromName(newAllocName, s.job.ID, tg.Name)
+
+				// If the allocation index is a duplicate, we cannot simply
+				// create a new allocation with the same name. We need to
+				// generate a new index and use this. The log message is useful
+				// for debugging and development, but could be removed in a
+				// future version of Nomad.
+				if taskGroupNameIndex.IsDuplicate(allocIndex) {
+					oldAllocName := newAllocName
+					newAllocName = taskGroupNameIndex.Next(1)[0]
```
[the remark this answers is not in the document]
jrasell: That's a good thread to pull on and something I will take a look at once this PR has been merged. This resolves a reproducible manifestation of the bug, so I would like to get that fixed given the time pressures before opening up new investigations.

scheduler/reconcile.go
jrasell: The comment has been clarified, hopefully that makes things clearer? I spent a while looking to see if I could fix the problem at the source, but I wasn't able to figure out a way.

scheduler/reconcile_util.go
jrasell: Resolved within commit https://github.com/hashicorp/nomad/pull/18873/commits/9fb64df1a75835ef9201cf1668f2890254c9ebcb

scheduler/generic_sched_test.go
jrasell: Resolved within commit https://github.com/hashicorp/nomad/pull/18873/commits/9fb64df1a75835ef9201cf1668f2890254c9ebcb

scheduler/reconcile_util.go
```diff
@@ -635,11 +648,25 @@ func bitmapFrom(input allocSet, minSize uint) structs.Bitmap {
...
 	for _, a := range input {
-		bitmap.Set(a.Index())
+
+		allocIndex := a.Index()
+
+		if bitmap.Check(allocIndex) {
+			duplicates[allocIndex]++
+		} else {
+			bitmap.Set(allocIndex)
+		}
```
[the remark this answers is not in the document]
jrasell: I agree that the name index tracking could be updated to include duplicates without using a bitmap, however, I think reworking this should be done within any follow-up. This change seems fairly low risk and targeted, whereas a larger rewrite would require substantial regression testing to be included. I'll raise an issue to discuss the idea of reworking the tracking, which I could noodle on casually.

tgross (approved):
LGTM!

PR #18874, opened 2023-10-26
agent: enable location logs by default in dev mode.

jrasell:


command/agent/config.go
```diff
@@ -71,7 +71,7 @@ type Config struct {
 	// LogIncludeLocation dictates whether the logger includes file and line
 	// information on each log line. This is useful for Nomad development and
 	// debugging.
-	LogIncludeLocation bool `hcl:"log_include_location"`
+	LogIncludeLocation *bool `hcl:"log_include_location"`
```
Juanadelacuesta, line 74: Why did you change it to a pointer?

angrycub: Typically a signal to indicate unset versus intentionally set to false.

PR #18878, opened 2023-10-26
e2e: fix bind name to allow Connect reachability

tgross:
The `BindName` for JWT authentication should always bind to the `nomad_service` field in the JWT and not include the namespace, as the `nomad_service` is what's actually registered in Consul. 

* Fix the binding rule for the `consulcompat` test 
* Add a reachability assertion so that we don't miss regressions.
* Ensure we have a clean shutdown so that we don't leak state (containers and iptables) between tests.

---

A note on debugging the original configuration problem. The allocations start up just fine and are healthy, so Nomad wasn't able to give much help here. To debug, I compared the results of `consul catalog services` with the results of `consul acl token list`, and looked for the tokens that resulted from the login. The `Service Identities` block in the token listing was the clue needed.

### DIFF
--- a/e2e/consulcompat/run_ce_test.go
+++ b/e2e/consulcompat/run_ce_test.go
@@ -204,7 +204,7 @@ func setupConsulJWTAuthForServices(t *testing.T, consulAPI *consulapi.Client, ad
 		AuthMethod:  "nomad-services",
 		Selector:    "",
 		BindType:    "service",
-		BindName:    "${value.nomad_namespace}-${value.nomad_service}",
+		BindName:    "${value.nomad_service}",
 	}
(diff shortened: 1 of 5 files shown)


e2e/consulcompat/run_ce_test.go
```diff
@@ -204,7 +204,7 @@ func setupConsulJWTAuthForServices(t *testing.T, consulAPI *consulapi.Client, ad
 		AuthMethod:  "nomad-services",
 		Selector:    "",
 		BindType:    "service",
-		BindName:    "${value.nomad_namespace}-${value.nomad_service}",
+		BindName:    "${value.nomad_service}",
```
pkazmierczak, line 207: That's a great catch.

pkazmierczak (approved):
LGTM!

PR #18879, opened 2023-10-26
docs: update link to tc-redirect-tap

p5:
The location of tc-redirect-tap has been changed and the docs are now out of date.
This PR points the link in the Firecracker task driver to the correct repository.

The new location is [here](https://github.com/awslabs/tc-redirect-tap) (https://github.com/awslabs/tc-redirect-tap)


tgross (approved):
LGTM. Thanks @p5 !

PR #18880, opened 2023-10-26
vault: derive token using `create_from_role`

lgfa29:
Fallback to the ACL role defined in the client's `create_from_role` configuration when using the JWT flow and the task does not specify a role to use.

### DIFF
--- a/client/config/config.go
+++ b/client/config/config.go
@@ -179,6 +179,8 @@ type Config struct {
 	ConsulConfigs map[string]*structsc.ConsulConfig
 
 	// VaultConfig is this Agent's default Vault configuration
+	//
+	// Deprecated: use GetVaultConfigs() instead.
 	VaultConfig *structsc.VaultConfig
 
 	// VaultConfigs is a map of Vault configurations, here to support features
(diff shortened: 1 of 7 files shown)


client/config/config.go
```diff
@@ -179,6 +179,8 @@ type Config struct {
 	ConsulConfigs map[string]*structsc.ConsulConfig
 
 	// VaultConfig is this Agent's default Vault configuration
+	//
+	// Deprecated: use GetVaultConfigs() instead.
 	VaultConfig *structsc.VaultConfig
```
lgfa29, line 184: I keep forgetting that I'm not supposed to read from this value. Wit this, my editor screams at me if I try to 😅

lgfa29:
Ops,. broke stuff 😬

client/config/config.go
tgross: You and me both. After we've shipped the beta and have a moment to breathe, I'd like to see if we could remove it entirely. It was too gnarly of a refactoring to do while we were also trying to ship the feature :grinning:

tgross (approved):
LGTM

PR #18881, opened 2023-10-26
cli: deprecate -vault-token flag

lgfa29:
Apply the same deprecation notice from #18863 to the `nomad job plan` command.


pkazmierczak (approved):
LGTM!

tgross (approved):
Oops, thanks for catching this! LGTM!

PR #18896, opened 2023-10-27
server: use same receiver name for all server funcs.

jrasell:


tgross (approved):
What, you don't like my artisanal per-file naming convention? :stuck_out_tongue_winking_eye: 

LGTM!

PR #18901, opened 2023-10-27
cli: Add file prediction for operator raft/snapshot commands

davemay99:


tgross (approved):
LGTM! Thanks @davemay99!

PR #18903, opened 2023-10-28
api: handle redirect during websocket upgrade

lgfa29:
When attempting a WebSocket connection upgrade the client may receive a redirect request from the server, in which case the request should be reattempted using the new address present in the `Location` header.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -992,15 +992,10 @@ func (c *Client) websocket(endpoint string, q *QueryOptions) (*websocket.Conn, *
 			// Connection upgrade was successful.
 
 		case http.StatusPermanentRedirect, http.StatusTemporaryRedirect, http.StatusMovedPermanently:
-			locs := resp.Header["Location"]
-			if len(locs) == 0 {
-				return nil, nil, errors.New("redirect request missing Location header")
-			}
-
-			loc := locs[0]
+			loc := resp.Header.Get("Location")
 			u, err := url.Parse(loc)
 			if err != nil {
-				return nil, nil, fmt.Errorf("invalid redirect location %s", loc)
+				return nil, nil, fmt.Errorf("invalid redirect location %q: %w", loc, err)
 			}
 			return c.websocket(u.Path, q)
 
@@ -1015,11 +1010,11 @@ func (c *Client) websocket(endpoint string, q *QueryOptions) (*websocket.Conn, *
 						withExpectedStatuses([]int{http.StatusSwitchingProtocols}),
 						withError(err))
 				}
-				io.Copy(&buf, greader)
+				_, _ = io.Copy(&buf, greader)
 			} else {
-				io.Copy(&buf, resp.Body)
+				_, _ = io.Copy(&buf, resp.Body)
 			}
-			resp.Body.Close()
+			_ = resp.Body.Close()
 
 			return nil, nil, newUnexpectedResponseError(
 				fromStatusCode(resp.StatusCode),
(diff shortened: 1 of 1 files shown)


api/api.go
```diff
@@ -986,28 +986,47 @@ func (c *Client) websocket(endpoint string, q *QueryOptions) (*websocket.Conn, *
...
-		var buf bytes.Buffer
+	if resp != nil {
+		switch resp.StatusCode {
+		case http.StatusSwitchingProtocols:
+			// Connection upgrade was successful.
+
+		case http.StatusPermanentRedirect, http.StatusTemporaryRedirect, http.StatusMovedPermanently:
+			locs := resp.Header["Location"]
+			if len(locs) == 0 {
+				return nil, nil, errors.New("redirect request missing Location header")
```
jrasell, line 997: Do we want to quote the `Location` entry to make it more obvious that's the header we are referring to?

api/api.go
```diff
@@ -986,28 +986,47 @@ func (c *Client) websocket(endpoint string, q *QueryOptions) (*websocket.Conn, *
...
-		if resp.Header.Get("Content-Encoding") == "gzip" {
-			greader, err := gzip.NewReader(resp.Body)
+			loc := locs[0]
+			u, err := url.Parse(loc)
 			if err != nil {
-				return nil, nil, newUnexpectedResponseError(
-					fromStatusCode(resp.StatusCode),
-					withExpectedStatuses([]int{http.StatusSwitchingProtocols}),
-					withError(err))
+				return nil, nil, fmt.Errorf("invalid redirect location %s", loc)
```
jrasell, line 1003: Do we want to add the `err` string to the returned error?

api/api.go
```diff
@@ -986,28 +986,47 @@ func (c *Client) websocket(endpoint string, q *QueryOptions) (*websocket.Conn, *
...
+
+			if resp.Header.Get("Content-Encoding") == "gzip" {
+				greader, err := gzip.NewReader(resp.Body)
+				if err != nil {
+					return nil, nil, newUnexpectedResponseError(
+						fromStatusCode(resp.StatusCode),
+						withExpectedStatuses([]int{http.StatusSwitchingProtocols}),
+						withError(err))
+				}
+				io.Copy(&buf, greader)
```
jrasell, line 1018:
```suggestion
				_, _ = io.Copy(&buf, greader)
```

api/api.go
```diff
@@ -986,28 +986,47 @@ func (c *Client) websocket(endpoint string, q *QueryOptions) (*websocket.Conn, *
...
+			} else {
+				io.Copy(&buf, resp.Body)
```
jrasell, line 1020:
```suggestion
				_, _ = io.Copy(&buf, resp.Body)
```

api/api.go
```diff
@@ -986,28 +986,47 @@ func (c *Client) websocket(endpoint string, q *QueryOptions) (*websocket.Conn, *
...
+			}
+			resp.Body.Close()
```
jrasell, line 1022:
```suggestion
			_ = resp.Body.Close()
```

jrasell (approved):
LGTM; just some minor comments and suggestions.

api/api.go
```diff
@@ -986,28 +986,47 @@ func (c *Client) websocket(endpoint string, q *QueryOptions) (*websocket.Conn, *
...
-	if resp != nil && resp.StatusCode != http.StatusSwitchingProtocols {
-		var buf bytes.Buffer
+	if resp != nil {
+		switch resp.StatusCode {
+		case http.StatusSwitchingProtocols:
+			// Connection upgrade was successful.
+
+		case http.StatusPermanentRedirect, http.StatusTemporaryRedirect, http.StatusMovedPermanently:
+			locs := resp.Header["Location"]
+			if len(locs) == 0 {
```
shoenig, line 996: always use the `Header` methods which do MIME ... stuff

https://pkg.go.dev/net/http#Header.Get

lgfa29: Ah thanks!

PR #18904, opened 2023-10-28
actions: minor bug fixes and improvements

lgfa29:
Some fixes and minor changes from a late review of https://github.com/hashicorp/nomad/commit/8902afe651fbd1d9c1b740b577a30216bd7eab66.

Bugs fixed:
* Validate `action` blocks have unique names.
* Validate `action` blocks have a command.
* Validate requested job ID exists.
* Validate requested allocation belongs to the requested job.
* Percent escape job ID in request path.
* Fix block name in HCLv1 parsing.
* No-op command args autocomplete since we can't predict action name.

Known bug I found during tests:
* Command doesn't work with jobs that end or start with `/` or that have multiple `/` in a row, but this is a bug in Go that was fixed recently    
   https://github.com/golang/go/issues/21955

Minor changes:
* Rename `-allocation` flag to just `-alloc` to match other commands.
* Adjustments to help message to match other commands.
* Output `-help` message when command error is due to bad input.
* Minor refactorings and removal of unnecessary functions.

Some questions I had during review:
* Are docs coming in a follow-up PR? Things we need to document:
  * New `nomad action` command.
  * New `/v1/job/<job id>/actions` and `/v1/job/<job id>/action` API endpoints.
  * New `nomad.nomad.job.get_actions` metric.
* I had the same question as https://github.com/hashicorp/nomad/pull/18778/files#r1362133596. Even with a node action concept I think they should still be separate commands, otherwise we will have a set of disparate flags for each kind of action. 
   
   For example, would `nomad action -job=example -node=123 -alloc=342` be a valid command? If not, then the number of permutations we would need to validate can balloon very quickly and make the code quite messy. It's also easier to hoist a default action type than later trying to split it. 

   So I think it would be better to have `nomad job action` and `nomad node action` as different commands and then shortcut `nomad action` to `nomad job action` (or apply some heuristic to determine what to do). 

   With this, since `-job` is a required flag I think it would make more sense to have it as an argument? Or maybe make it optional if `-alloc` is passed? Ooor have it as `nomad alloc exec -action <action> <alloc id>` and remove the `-alloc` flag from `nomad job action`?
* Probably a future work, but would it make sense to add metrics for indivudal actions? Like how many times each action is being called and how long they take to run?

### DIFF
--- a/command/action.go
+++ b/command/action.go
@@ -310,6 +310,7 @@ func (c *ActionCommand) execImpl(client *api.Client, alloc *api.Allocation, task
 		}
 	}()
 
-	return client.Jobs().ActionExec(ctx,
-		alloc, task, tty, make([]string, 0), action, stdin, stdout, stderr, sizeCh, nil)
+	return client.Jobs().ActionExec(ctx, alloc, job, task,
+		tty, make([]string, 0), action,
+		stdin, stdout, stderr, sizeCh, nil)
 }
--- a/command/action.go
+++ b/command/action.go
@@ -45,15 +45,15 @@ General Options:
 
 Action Specific Options:
 
-  -job=<job-id>
+  -job <job-id>
     Specifies the job in which the Action is defined
 
-  -alloc=<allocation-id>
+  -alloc <allocation-id>
     Specifies the allocation in which the Action is defined. If not provided,
     a group and task name must be provided and a random allocation will be
     selected from the job.
 
-  -task=<task-name>
+  -task <task-name>
     Specifies the task in which the Action is defined. Required if no
     allocation is provided.
 
@@ -68,10 +68,10 @@ Action Specific Options:
     Allocate a pseudo-tty, defaults to true if stdin is detected to be a tty session.
     Pass -t=false to disable explicitly.
 
-  -e=<escape_char>
-    Sets the escape character for sessions with a pty (default: '~').  The escape
-    character is only recognized at the beginning of a line.  The escape character
-    followed by a dot ('.') closes the connection.  Setting the character to
+  -e <escape_char>
+    Sets the escape character for sessions with a pty (default: '~'). The escape
+    character is only recognized at the beginning of a line. The escape character
+    followed by a dot ('.') closes the connection. Setting the character to
     'none' disables any escapes and makes the session fully transparent.
 `
 	return strings.TrimSpace(helpText)
--- a/command/action.go
+++ b/command/action.go
@@ -14,7 +14,6 @@ import (
 	"syscall"
 
 	"github.com/hashicorp/nomad/api"
-	"github.com/hashicorp/nomad/api/contexts"
 	"github.com/hashicorp/nomad/helper/escapingio"
 	"github.com/posener/complete"
 )
@@ -27,7 +26,7 @@ type ActionCommand struct {
 	Stderr io.WriteCloser
 }
 
-func (l *ActionCommand) Help() string {
+func (c *ActionCommand) Help() string {
 	helpText := `
 Usage: nomad action [options] <action>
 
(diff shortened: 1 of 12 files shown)


nomad/structs/structs.go
```diff
@@ -7872,6 +7881,22 @@ func (t *Task) Validate(jobType string, tg *TaskGroup) error {
...
+		if err := action.Validate(); err != nil {
+			outer := fmt.Errorf("Action %s validation failed: %s", action.Name, err)
+			mErr.Errors = append(mErr.Errors, outer)
+		}
+
+		if handled, seen := actions[action.Name]; seen && !handled {
+			mErr.Errors = append(mErr.Errors, fmt.Errorf("Action %s defined multiple times", action.Name))
+			actions[action.Name] = true
+			continue
+		}
```
lgfa29, line 7896: This way to check for duplicates is a bit different from others because it handles 3+ repetitions as a single error. Not sure if that's better, but if it's we could apply the same logic to other blocks.

tgross: :+1: I like this approach. We don't include file context from the original HCL, so I feel like having repeats of the error doesn't get us any new information.

nomad/client_alloc_endpoint.go
```diff
@@ -430,6 +431,31 @@ func (a *ClientAllocations) exec(conn io.ReadWriteCloser) {
 		return
 	}
 
+	// Handle job ID if requested.
```
tgross, line 434: If the original HTTP API request is made to the client that has the allocation, it's handled entirely locally without going through the server. This extra validation here gets avoids sending the client RPC if it shouldn't be sent, but maybe we should check the alloc client status as long as we're here too? (Recognizing that the allocation could fail after we send the RPC to the client, of course!)

command/agent/job_endpoint.go
```diff
@@ -362,16 +363,17 @@ func (s *HTTPServer) jobActions(resp http.ResponseWriter, req *http.Request, job
...
-
-	s.logger.Info("jobRunAction called")
-
-	// Build the request and parse the ACL token
+func (s *HTTPServer) jobRunAction(resp http.ResponseWriter, req *http.Request, jobID string) (interface{}, error) {
 	task := req.URL.Query().Get("task")
 	action := req.URL.Query().Get("action")
 	allocID := req.URL.Query().Get("allocID")
+
+	s.logger.Info("job action called",
```
tgross, line 371: Possibly worth noting that the existing `alloc exec` doesn't have this log line.

command/action.go
```diff
@@ -46,102 +45,98 @@ General Options:
 
 Action Specific Options:
 
-  -job <job-id>
+  -job=<job-id>
     Specifies the job in which the Action is defined
 
-  -allocation <allocation-id>
+  -alloc=<allocation-id>
```
tgross, line 51: I don't think we put `=` anywhere else? I just spot-checked a bunch of commands and didn't find any that show this.

tgross (approved):

command/agent/job_endpoint.go
philrenaud: This had been a debugging-purposes log for me at first and I thought it might still be relevant as a server log, but it probably doesn't meet the "more signal than noise" threshold.

philrenaud (approved):
Thank you for this! Thanks also for `allocation` to `alloc` shortening; wasn't really sure what the convention was there.

nomad/client_alloc_endpoint.go
lgfa29: Ah yeah, I forgot to mention that this endpoint is kind of "weird" in that sense: it can be routed from servers or sent directly to clients.

I tried to push as much validation to clients, but since they don't have access to cluster state, checking for [`req.JobID != alloc.JobID`](https://github.com/hashicorp/nomad/pull/18904/files/d53bebe75cf65518aa0b614835463ecdf0d218e9#diff-e2b2a83c57344644a143d5dbe1b6fdb28b05316170410efa904a6875d4484db8R252-R255) was all that I could do 😅 

This means that, when sending a request directly to clients, the job ID is not fully checked. But I think that's OK, since matching the alloc `JobID` carries a lot of preconceived assumptions (like that the job exists).

It also felt to me that the API design is a bit strange since the only things we actually care about to run an action is an alloc ID, a task name, and an action, so to me it feels like that this should be a `/v1/allocation/<id>/action?task=<task>&action=<action>` instead of `/v1/job/<job>/action?alloc=<alloc>&task=<task>&action=<action>` since `<job>` is kind of meaningless, but I didn't want to make any big change like this so late into the game 😅

lgfa29: Added check for terminal status in 4e2696cc3d

command/agent/job_endpoint.go
lgfa29: Hum....yeah, and I think we log every HTTP request so, to some degree we already have a log for when this happens? I think I will remove this.

command/action.go
lgfa29: Ah interesting, so every command is inconsistent because we do add the `=` in `generalOptionsUsage` 😅 
https://github.com/hashicorp/nomad/blob/4850f072955fee6ede30e9d17392e5f8dca20854/command/meta.go#L345-L413

I will remove them here and then maybe we can also remove from `generalOptionsUsage`?

PR #18910, opened 2023-10-30
cli: setup vault command

pkazmierczak:
An interactive setup helper for configuring Vault to accept Nomad WI-enabled workloads.

### DIFF
--- a/command/setup_vault.go
+++ b/command/setup_vault.go
@@ -34,6 +34,7 @@ const (
 	vaultPolicyName = "nomad-workloads"
 	vaultNamespace  = "nomad-workloads"
 	vaultAud        = "vault.io"
+	vaultPath       = "jwt-nomad"
 )
 
 type SetupVaultCommand struct {
@@ -384,7 +385,7 @@ vault {
   # Vault Enterprise only.
   # namespace = "<namespace>"
 
-  jwt_auth_backend_path = "jwt/"
+  jwt_auth_backend_path = "jwt-nomad/"
 }
 
 And your Nomad server configuration in the following way:
@@ -407,7 +408,7 @@ func (s *SetupVaultCommand) jwtEnabled() bool {
 }
 
 func (s *SetupVaultCommand) enableJWT() error {
-	err := s.vClient.Sys().EnableAuthWithOptions("jwt", &api.MountInput{Type: "jwt"})
+	err := s.vClient.Sys().EnableAuthWithOptions(vaultPath, &api.MountInput{Type: "jwt"})
 	if err != nil {
 		return fmt.Errorf("[✘] Could not enable JWT credential backend: %w", err)
 	}
@@ -416,7 +417,7 @@ func (s *SetupVaultCommand) enableJWT() error {
 }
 
 func (s *SetupVaultCommand) roleExists() bool {
-	existingRoles, _ := s.vLogical.List("/auth/jwt/role")
+	existingRoles, _ := s.vLogical.List(fmt.Sprintf("/auth/%s/role", vaultPath))
 	if existingRoles != nil {
 		return slices.Contains(existingRoles.Data["keys"].([]any), vaultRole)
 	}
@@ -441,7 +442,7 @@ func (s *SetupVaultCommand) createRole(role map[string]any) error {
 		return fmt.Errorf("[✘] Role could not be interpolated with args: %w", err)
 	}
 
-	path := "auth/jwt/role/" + vaultRole
+	path := fmt.Sprintf("auth/%s/role/%s", vaultPath, vaultRole)
 
 	_, err = s.vLogical.WriteBytes(path, buf)
 	if err != nil {
@@ -458,7 +459,7 @@ func (s *SetupVaultCommand) policyExists() bool {
 }
 
 func (s *SetupVaultCommand) renderPolicy() (string, error) {
-	secret, err := s.vLogical.Read("sys/auth/jwt")
+	secret, err := s.vLogical.Read("sys/auth/" + vaultPath)
 	if err != nil {
 		return "", fmt.Errorf("[✘] Could not retrieve JWT accessor: %w", err)
 	}
@@ -485,7 +486,7 @@ func (s *SetupVaultCommand) createPolicy(policyText string) error {
 }
 
 func (s *SetupVaultCommand) authMethodExists() bool {
-	existingConf, _ := s.vLogical.Read("/auth/jwt/config")
+	existingConf, _ := s.vLogical.Read(fmt.Sprintf("/auth/%s/config", vaultPath))
 	return existingConf != nil
 }
 
@@ -507,7 +508,7 @@ func (s *SetupVaultCommand) createAuthMethod(authConfig map[string]any) error {
 	if err != nil {
 		return fmt.Errorf("auth method could not be interpolated with args: %w", err)
 	}
-	_, err = s.vLogical.WriteBytes("auth/jwt/config", buf)
+	_, err = s.vLogical.WriteBytes(fmt.Sprintf("auth/%s/config", vaultPath), buf)
 	if err != nil {
 		if strings.Contains(err.Error(), "error checking jwks URL") {
 			s.Ui.Error(fmt.Sprintf(
@@ -650,7 +651,7 @@ func (s *SetupVaultCommand) removeConfiguredComponents() int {
 		}
 
 		if role, ok := componentsToRemove["Role"]; ok {
-			_, err := s.vLogical.Delete(fmt.Sprintf("/auth/jwt/role/%s", role))
+			_, err := s.vLogical.Delete(fmt.Sprintf("/auth/%s/role/%s", vaultPath, role))
 			if err != nil {
 				s.Ui.Error(fmt.Sprintf("[✘] Failed to delete role %q: %v", role, err.Error()))
 				exitCode = 1
@@ -660,7 +661,7 @@ func (s *SetupVaultCommand) removeConfiguredComponents() int {
 		}
 
 		if _, ok := componentsToRemove["JWT Credential backend and its configuration"]; ok {
-			if err := s.vClient.Sys().DisableAuth("jwt"); err != nil {
+			if err := s.vClient.Sys().DisableAuth(vaultPath); err != nil {
 				s.Ui.Error(fmt.Sprintf("[✘] Failed to disable JWT credential backend: %v", err.Error()))
 				exitCode = 1
 			} else {
--- a/command/setup_vault.go
+++ b/command/setup_vault.go
@@ -192,7 +192,7 @@ Please set the VAULT_NAMESPACE environment variable to the Vault namespace to us
 Since you're running Vault Enterprise, we will additionally create
 a namespace %q and create all configuration within that namespace.
 `
-		if s.namespaceExists(s.ns) {
+		if s.namespaceExists(s.ns, false) {
 			s.Ui.Info(fmt.Sprintf("[✔] Namespace %q already exists.", s.ns))
 		} else {
 			s.Ui.Output(fmt.Sprintf(namespaceMsg, s.ns))
@@ -418,7 +418,7 @@ func (s *SetupVaultCommand) enableJWT() error {
 func (s *SetupVaultCommand) roleExists() bool {
 	existingRoles, _ := s.vLogical.List("/auth/jwt/role")
 	if existingRoles != nil {
-		return slices.Contains(existingRoles.Data["keys"].([]interface{}), vaultRole)
+		return slices.Contains(existingRoles.Data["keys"].([]any), vaultRole)
 	}
 	return false
 }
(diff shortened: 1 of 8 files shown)


command/setup_vault.go
```diff
@@ -0,0 +1,675 @@
...
+
+func (s *SetupVaultCommand) renderRole() (map[string]any, error) {
+	role := map[string]any{}
+	err := json.Unmarshal(vaultRoleBody, &role)
+	if err != nil {
+		return role, fmt.Errorf("[✘] Default auth config text could not be deserialized: %w", err)
+	}
+
+	role["bound_audiences"] = vaultAud
+	role["token_policies"] = []string{vaultPolicyName}
```
lgfa29, line 431: We should probably not modify these and use the file asset as the source of truth. Otherwise it could get confusing to find out what needs to be changed.

command/setup_vault.go
```diff
@@ -0,0 +1,675 @@
...
+
+	return authConfig, nil
+}
+
+func (s *SetupVaultCommand) createAuthMethod(authConfig map[string]any) error {
+	buf, err := json.Marshal(authConfig)
+	if err != nil {
+		return fmt.Errorf("auth method could not be interpolated with args: %w", err)
+	}
+	_, err = s.vLogical.WriteBytes("auth/jwt/config", buf)
```
lgfa29, line 508: We should probably use a non-default name for this to avoid a conflict. 

I uses `jwt-nomad` in the TF module:
https://github.com/hashicorp-modules/terraform-vault-nomad-setup/blob/main/variables.tf#L16

command/setup_vault.go
```diff
@@ -0,0 +1,675 @@
...
+	if err != nil {
+		if strings.Contains(err.Error(), "error checking jwks URL") {
+			s.Ui.Error(fmt.Sprintf(
+				"error: Nomad JWKS endpoint unreachable, verify that Nomad is running and that the JWKS URL %s is reachable by Consul", s.jwksURL,
+			))
+			os.Exit(1)
+		}
+		return fmt.Errorf("[✘] Could not create Consul auth method: %w", err)
```
lgfa29, line 516:
```suggestion
				"error: Nomad JWKS endpoint unreachable, verify that Nomad is running and that the JWKS URL %s is reachable by Vault", s.jwksURL,
			))
			os.Exit(1)
		}
		return fmt.Errorf("[✘] Could not create Vault auth method: %w", err)
```

command/setup_vault.go
```diff
@@ -0,0 +1,675 @@
...
+	}
+
+	s.Ui.Info("[✔] Created JWT auth method configuration.")
```
lgfa29, line 519:
```suggestion
	s.Ui.Info("[✔] Created JWT auth method.")
```

command/setup_vault.go
```diff
@@ -0,0 +1,675 @@
...
+func (s *SetupVaultCommand) removeConfiguredComponents() int {
+	s.vClient.SetNamespace(s.ns)
+	exitCode := 0
+	componentsToRemove := map[string]string{}
+
+	if s.policyExists() {
+		componentsToRemove["Policy"] = consulPolicyName
+	}
+	if s.roleExists() {
+		componentsToRemove["Role"] = consulRoleTasks
```
lgfa29, line 594:
```suggestion
By answering "no" to any of these questions, you are risking an incorrect Vault
cluster configuration. Nomad workloads with Workload Identity will not be able
to authenticate unless you create missing configuration yourself.
 `)

	exitCode := 0
	if s.autoYes || s.askQuestion("Remove everything this command creates? [Y/n]") {
		exitCode = s.removeConfiguredComponents()
	}

	s.Ui.Output(s.Colorize().Color(`
Vault cluster has [bold][underline]not[reset] been configured for authenticating Nomad tasks
using workload identitiies.

Run the command again to finish the configuration process.`))
	os.Exit(exitCode)
}

func (s *SetupVaultCommand) removeConfiguredComponents() int {
	s.vClient.SetNamespace(s.ns)
	exitCode := 0
	componentsToRemove := map[string]string{}

	if s.policyExists() {
		componentsToRemove["Policy"] = vaultPolicyName
	}
	if s.roleExists() {
		componentsToRemove["Role"] = vaultRole
```

command/setup_vault.go
```diff
@@ -0,0 +1,675 @@
...
+ `)
+
+	exitCode := 0
+	if s.autoYes || s.askQuestion("Remove everything this command creates? [Y/n]") {
+		exitCode = s.removeConfiguredComponents()
+	}
+
+	s.Ui.Output(s.Colorize().Color(`
+Consul cluster has [bold][underline]not[reset] been configured for authenticating Nomad tasks and
+services using workload identitiies.
```
lgfa29, line 579:
```suggestion
services using workload identities.
```

Missed this in the Consul command as well.

command/setup_vault.go
```diff
@@ -0,0 +1,675 @@
...
+	}
+	if s.roleExists() {
+		componentsToRemove["Role"] = consulRoleTasks
+	}
+	if s.jwtEnabled() {
+		componentsToRemove["JWT Credential backend and its configuration"] = ""
+	}
+	if s.namespaceExists(s.ns) {
+		componentsToRemove["Namespace"] = s.ns
+	}
```
lgfa29, line 601: We should not remove a namespace we didn't create so we should check the namespace meta for the data we put there, like I did in the Consul command: https://github.com/hashicorp/nomad/pull/18820/commits/8040445e3b2337d8245b368e58b9a37194e65cf0.

I think we also need to check for Vault Enterprise?

command/setup_vault.go
```diff
@@ -0,0 +1,675 @@
...
+	q := `The following items will be deleted:
+%s`
+	if len(componentsToRemove) == 0 {
+		s.Ui.Output("Nothing to delete.")
+		return 0
+	}
+
+	if !s.autoYes {
+		s.Ui.Warn(fmt.Sprintf(q, printMapOfStrings(componentsToRemove)))
+	}
```
lgfa29, line 615:
```suggestion
	if len(componentsToRemove) == 0 {
		s.Ui.Output("Nothing to delete.")
		return 0
	}
	
	q := `The following items will be deleted:
%s`
	if !s.autoYes {
		s.Ui.Warn(fmt.Sprintf(q, printMapOfStrings(componentsToRemove)))
	}
```

Don't need to declare `q` if we're not going to need it.

command/setup_vault.go
```diff
@@ -0,0 +1,675 @@
...
+			if err != nil {
+				s.Ui.Error(fmt.Sprintf("[✘] Failed to delete role %q: %v", role, err.Error()))
+				exitCode = 1
+			} else {
+				s.Ui.Info(fmt.Sprintf("[✔] Deleted role %q.", role))
+			}
+		}
+
+		if _, ok := componentsToRemove["JWT Credential backend and its configuration"]; ok {
+			if err := s.vClient.Sys().DisableAuth("jwt"); err != nil {
```
lgfa29, line 640: Hum...it's probably safe to leave the JWT backend enabled, but we should remove the auth method we create.

lgfa29 (commented):
Round 1 😅

command/setup_vault.go
```diff
@@ -0,0 +1,675 @@
...
+// Help satisfies the cli.Command Help function.
+func (s *SetupVaultCommand) Help() string {
+	helpText := `
+Usage: nomad setup vault [options]
+
+  This command sets up Vault for allowing Nomad workloads to authenticate
+  themselves using Workload Identity.
+
+  This command requires acl:write permissions for Vault and respects
+  VAULT_TOKEN, VAULT_ADDR, and other Consul-related environment variables
```
lgfa29, line 61: One thing that I find really annoying is that the default value if `VAULT_ADDR` is not set uses `https`.  Since this command is probably going to be used in local test environments would it make sense to special override this default to use `http`?

command/setup_vault.go
pkazmierczak: So this is all a bit more tricky, because the Vault client is fully namespaced. This means that if a namespace is set, all API calls work within the namespace. Among other things this means you can't check if a namespace exists if you're within it (namespaces can be nested), you also cannot make any meaningful "is-ent" check unless you're in `""` namespace. 

> I think we also need to check for Vault Enterprise?

yeah so the way I do it now is I check `	ent = s.vClient.Namespace() != "" || license != nil` conditional after the client initialization. If we're connecting to CE vault, the `s.namespaceExists(s.ns)` will always return false, so no namespace will get added to the map.

command/setup_vault.go
pkazmierczak: Huh, I thought maybe if we clean everything then we should clean everything? Credential backends are also namespaced, so for ent this matters I guess?

command/setup_vault.go
```diff
@@ -0,0 +1,675 @@
...
+			s.ns = s.vClient.Namespace()
+		} else {
+			// Set default namespace if VAULT_NAMESPACE is not defined.
+			s.ns = vaultNamespace
+			s.vClient.SetNamespace(s.ns)
+		}
+
+		if s.cleanup {
+			return s.removeConfiguredComponents()
+		}
```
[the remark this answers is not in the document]
pkazmierczak: corrected

command/setup_vault.go
pkazmierczak: > We should not remove a namespace we didn't create

Yes! Fixed in f1d5c96

command/setup_vault.go
```diff
@@ -0,0 +1,675 @@
...
+			enableJWT = true
+		}
+
+		if enableJWT {
+			err := s.enableJWT()
+			if err != nil {
+				s.Ui.Error(err.Error())
+				return 1
+			}
+		}
```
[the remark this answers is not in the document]
pkazmierczak: see [db2f7fc](https://github.com/hashicorp/nomad/pull/18910/commits/db2f7fcb729ec506a4fc63204dc42c741ca046ad)

command/setup_vault.go
```diff
@@ -0,0 +1,675 @@
...
+
+		if createAuthMethodConf {
+			err = s.createAuthMethod(authMethodConf)
+			if err != nil {
+				s.Ui.Error(err.Error())
+				return 1
+			}
+		}
+	}
```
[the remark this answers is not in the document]
pkazmierczak: see [db2f7fc](https://github.com/hashicorp/nomad/pull/18910/commits/db2f7fcb729ec506a4fc63204dc42c741ca046ad)

command/setup_vault.go
```diff
@@ -0,0 +1,667 @@
...
+		if strings.Contains(err.Error(), "error checking jwks URL") {
+			s.Ui.Error(fmt.Sprintf(
+				"error: Nomad JWKS endpoint unreachable, verify that Nomad is running and that the JWKS URL %s is reachable by Vault", s.jwksURL,
+			))
+			os.Exit(1)
+		}
+		return fmt.Errorf("[✘] Could not create Vault auth method: %w", err)
+	}
+
+	s.Ui.Info("[✔] Created JWT auth method.")
```
lgfa29, line 490:
```suggestion
	s.Ui.Info("[✔] Created JWT auth method %q.", vaultPath)
```

command/setup_vault.go
```diff
@@ -0,0 +1,667 @@
...
+
+	if s.policyExists() {
+		componentsToRemove["Policy"] = vaultPolicyName
+	}
+	if s.roleExists() {
+		componentsToRemove["Role"] = vaultRole
+	}
+	if s.jwtEnabled() {
+		componentsToRemove["JWT Credential backend and JWT Auth method configuration"] = vaultPath
+	}
```
lgfa29, line 590:
```suggestion
	if s.authMethodExists() {
		componentsToRemove["JWT auth method"] = vaultPath
	}
```

I think `jwtEnabled` is not needed anymore after this.

command/setup_vault.go
```diff
@@ -0,0 +1,667 @@
...
+	if s.namespaceExists(s.ns, true) {
+		componentsToRemove["Namespace"] = s.ns
+	}
+	if exitCode != 0 {
+		return exitCode
+	}
```
lgfa29, line 596:
```suggestion
```

We don't modify `exitCode` anywhere prior to this, so it's always `0` here.

command/setup_vault.go
```diff
@@ -0,0 +1,667 @@
...
+			_, err := s.vLogical.Delete(fmt.Sprintf("/auth/%s/role/%s", vaultPath, role))
+			if err != nil {
+				s.Ui.Error(fmt.Sprintf("[✘] Failed to delete role %q: %v", role, err.Error()))
+				exitCode = 1
+			} else {
+				s.Ui.Info(fmt.Sprintf("[✔] Deleted role %q.", role))
+			}
+		}
+
+		if _, ok := componentsToRemove["JWT Credential backend and JWT Auth method configuration"]; ok {
```
lgfa29, line 631:
```suggestion
		if _, ok := componentsToRemove["JWT auth method"]; ok {
```

command/setup_vault.go
```diff
@@ -0,0 +1,667 @@
...
+			if err := s.vClient.Sys().DisableAuth(vaultPath); err != nil {
+				s.Ui.Error(fmt.Sprintf("[✘] Failed to disable JWT credential backend: %v", err.Error()))
+				exitCode = 1
+			} else {
+				s.Ui.Info("[✔] Disabled JWT credential backend.")
```
lgfa29, line 636:
```suggestion
				s.Ui.Error(fmt.Sprintf("[✘] Failed to disable JWT auth method %q %v", vaultPath, err.Error()))
				exitCode = 1
			} else {
				s.Ui.Info("[✔] Disabled JWT auth method %q.", vaultPath)
```

lgfa29 (approved):
Some minor comments, but LGTM!

I think https://github.com/hashicorp/nomad/pull/18910/files#r1382301949 could simplify the code a bit, but I'm not sure if it would actually work.

command/setup_vault.go
```diff
@@ -0,0 +1,675 @@
...
+				createNamespace = s.askQuestion(
+					fmt.Sprintf("Create the namespace %q in your Vault cluster? [Y/n]", s.ns))
+				if !createNamespace {
+					s.handleNo()
+				}
+			} else {
+				createNamespace = true
+			}
+
+			if createNamespace {
```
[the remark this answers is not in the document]
pkazmierczak: I'll make a separate PR to fix this in both setup vault and consul.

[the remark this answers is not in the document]
pkazmierczak: https://github.com/hashicorp/nomad/pull/19011

PR #18912, opened 2023-10-30
[ui] Helios two-step button and uniform title bar for Actions

philrenaud:
For purposes of changing the Exec and Stop buttons shown here:
image

we now have all two-step buttons using Helios-style buttons instead of our previous style.

### DIFF
--- a/ui/tests/pages/components/two-step-button.js
+++ b/ui/tests/pages/components/two-step-button.js
@@ -3,13 +3,7 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
-import {
-  attribute,
-  clickable,
-  hasClass,
-  isPresent,
-  text,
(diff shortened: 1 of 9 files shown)


ui/tests/pages/components/two-step-button.js
```diff
@@ -20,7 +14,6 @@ export default (scope) => ({
   confirm: clickable('[data-test-confirm-button]'),
   cancel: clickable('[data-test-cancel-button]'),
 
-  isRunning: hasClass('is-loading', '[data-test-confirm-button]'),
```
philrenaud, line 23: Tests that use this now test for "Loading...." text instead.

juliezzhou:
From the screenshot, looks like the height of the buttons and the dropdown menu button are still slightly off for like 1-2 px?

PR #18913, opened 2023-10-30
`identity_hook`: implement PreKill hook, not TaskStop hook

tgross:
The allocrunner's `identity_hook` implements the interface for TaskStop, but this interface is only ever called for task-level hooks. This results in a leaked goroutine that tries to periodically renew WIs until the client shuts down gracefully.

Add an implementation for the allocrunner's `PreKill` and `Destroy` hooks, so that whenever an allocation is stopped or garbage collected we stop renewing its Workload Identities. This also requires making the `Shutdown` method of `WIDMgr` safe to call multiple times.

### DIFF
--- a/client/allocrunner/identity_hook.go
+++ b/client/allocrunner/identity_hook.go
@@ -4,10 +4,7 @@
 package allocrunner
 
 import (
-	"context"
-
 	log "github.com/hashicorp/go-hclog"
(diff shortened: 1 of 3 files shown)


pkazmierczak (approved):
LGTM, thanks for catching this @tgross!

PR #18914, opened 2023-10-30
services: fix lookup for Consul tokens

tgross:
The `group_service_hook` needs to supply the Consul service client with Consul tokens for its services. The lookup in the hook resources was looking for the wrong key. This would cause the service client to ignore the Consul token we've received and use the agent's own token.

### DIFF
--- a/client/allocrunner/group_service_hook.go
+++ b/client/allocrunner/group_service_hook.go
@@ -5,7 +5,6 @@ package allocrunner
 
 import (
 	"context"
-	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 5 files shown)


client/allocrunner/group_service_hook.go
```diff
@@ -267,7 +268,9 @@ func (h *groupServiceHook) getWorkloadServicesLocked() *serviceregistration.Work
 
 	tokens := map[string]string{}
 	for _, service := range h.services {
-		if token, ok := allocTokens[service.Cluster][service.MakeUniqueIdentityName()]; ok {
+		id := fmt.Sprintf("%s_%s",
+			structs.ConsulServiceIdentityNamePrefix, service.MakeUniqueIdentityName())
+		if token, ok := allocTokens[service.Cluster][id]; ok {
```
lgfa29, line 273: I feel like we have this logic sprinkled in several places. Are there scenarios where we would want to use the result of `MakeUniqueIdentityName` without the identity prefix? Otherwise we could have it so `MakeUniqueIdentityName` already returns the prefixed identity name.

lgfa29 (approved):

client/allocrunner/group_service_hook.go
tgross: I think you're right... I just did a quick lookup and all callers of `(*Service) MakeUniqueIdentityName` are formatting it with the prefix. I can update the method and all callers.

tgross: Done in 662f80b

PR #18922, opened 2023-10-31
scheduler: split service/batch alloc logic into its own package.

jrasell:
Move the allocation name index and allocation set functionality to a new package within the scheduler. I'd like to reduce down the overall size of the scheduler package and create distinct packages based on the scheduler type. This will hopefully make navigating the code easier and allow us to make better boundaries.

Once merged, I plan to at least move the reconcile_util package within here also. I didn't do it as part of this PR, to reduce the churn and diff.


tgross (commented):
Overall approach looks good. Presumably you're planning on moving everything for batch/service currently in `scheduler` into `scheduler/generic` and not leaving it as an empty package. We'll almost certainly need to move all the interfaces definitions into their own shared package in order to do that and keep the test structure intact (I'd particularly take a look at the `benchmarks` directory).

That being said, the boundaries between `scheduler/generic` and `scheduler/generic/alloc` feel really fuzzy to me. The `NameIndex` logic can't be shared between scheduler types. But the `AllocSet` almost wants to be shared between scheduler types, except it has reconciler-specific methods that could likely be pulled out into free functions in `scheduler/generic`. I worry that by introducing nested packages that care about logic from their parent package, that we're going to run into circular imports or have to create a bunch of new interfaces. So I'm interested to see how far you plan on going with those split-ups -- breaking each scheduler type into a half dozen packages feels like it would be really painful to understand if the boundaries are already fuzzy.

Which is all to say that I wonder if it might be a good approach to break the largest chunks out first and then the smaller bits?

Juanadelacuesta (commented):
Since you are working on splitting the cod, I think it can be useful to also think a little about dividing the functionality to improve the code isolation and make it possible to have uni testing?

Issue #16566, opened 2023-03-20
[CLI] - Json format option with nomad job status


Lord-Y:
Hello guys,

Is it possible to have the `json`  format option when doing `nomad job status example-job-name`.  That will be great in order to manipulate allocations with JQ instead of using `grep`

jrasell:
Hi @Lord-Y thanks for raising this issue. This is an interesting one because most other `json` flag implementations dump a single object or array of objects in json format, whereas this command and implementation would likely wrap a number of related objects such as jobspec, allocations, evaluations, and deployments. I do think this would be very useful to have, so thanks again.

sharmaansh21:
@jrasell  I would like to pick this up, do you have any pointers for me ?

0xMALVEE:
is this issue fixed already?

Lord-Y:
@0xMALVEE I can confirm that with nomad `1.7.7` it's working: `nomad job status -json`. 
The only thing is it's not mentionned in the help which is too bad.
I'll close this issue then.

PR #18925, opened 2023-10-31
-json option on jobs status command

geovannyAvelar:
https://github.com/hashicorp/nomad/issues/16566

This modification dumps the entire `job` struct in JSON format. I don't know if just dumps the entire struct is the better approach, if not, i can expose specific fields.

### DIFF
--- a/command/job_status.go
+++ b/command/job_status.go
@@ -161,8 +161,13 @@ func (c *JobStatusCommand) Run(args []string) int {
 					ids[i] = j.ID
 				}
 
-				out, err := c.createJsonJobsOutput(client, ids...)
+				jsonJobs, err := createJsonJobsOutput(client, c.allAllocs, ids...)
+				if err != nil {
+					c.Ui.Error(err.Error())
+					return 1
+				}
 
+				out, err := Format(true, c.tmpl, jsonJobs)
 				if err != nil {
 					c.Ui.Error(err.Error())
 					return 1
@@ -201,8 +206,14 @@ func (c *JobStatusCommand) Run(args []string) int {
 	}
 
 	if c.json || len(c.tmpl) > 0 {
-		out, err := c.createJsonJobsOutput(client, *job.ID)
+		jsonJobs, err := createJsonJobsOutput(client, c.allAllocs, *job.ID)
 
+		if err != nil {
+			c.Ui.Error(err.Error())
+			return 1
+		}
+
+		out, err := Format(true, c.tmpl, jsonJobs)
 		if err != nil {
 			c.Ui.Error(err.Error())
 			return 1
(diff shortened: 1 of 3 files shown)


tgross (commented):
Hi @geovannyAvelar! I know this is still in draft but just a heads up that the approach you've got here is essentially what the existing `nomad job inspect` does. The goal described in https://github.com/hashicorp/nomad/issues/16566 is to have the whole command's output in some JSON-structured way, which includes a lot of ancillary information like the allocations, deployments, etc.

geovannyAvelar:
Hello @tgross! I made some changes. Now when `-json` option is included in `nomad job status` command, we output same data of `nomad job status`  command (evals, allocations, latest deployment and summary). Please let me know if I need to change or include something. Do I need to include something different when handling periodic or parameterized jobs?

command/job_status.go
```diff
@@ -666,6 +718,51 @@ func (c *JobStatusCommand) outputFailedPlacements(failedEval *api.Evaluation) {
...
+	for i, id := range jobIds {
+		job, _, err := client.Jobs().Info(id, &api.QueryOptions{})
+		if err != nil {
+			return nil, fmt.Errorf("Error querying job info: %s", err)
+		}
+
+		var q *api.QueryOptions
+		if job.Namespace != nil {
+			q = &api.QueryOptions{Namespace: *job.Namespace}
+		}
```
tgross, line 733: I did some testing of this branch with jobs in namespaces and discovered that we're not passing in the correct namespace for each job ID we're querying, and instead end up picking up the namespace set earlier. What that results in is wildcard namespaces being broken with the `-json` flag:

```
$ nomad job status -namespace '*'
ID      Namespace  Type     Priority  Status          Submit Date
httpd   default    service  50        dead (stopped)  2024-02-22T12:53:17-05:00
httpd   prod       service  50        running         2024-02-22T12:53:35-05:00
system  default    system   50        dead (stopped)  2024-02-22T12:53:20-05:00

$ nomad job status -json -namespace '*'
Error querying job info: Unexpected response code: 404 (job not found)
```

So I think we need to pass in namespace/ID pairs to this function.

tgross (changes requested):

command/job_status.go
```diff
@@ -26,6 +26,20 @@ type JobStatusCommand struct {
...
+	Summary          *api.JobSummary
+	Allocations      []*api.AllocationListStub
+	LatestDeployment *api.Deployment
+	Evaluations      []*api.Evaluation
+}
+
+type jobIDNamespacePair struct {
+	ID        string
+	Namespace string
 }
```
tgross, line 43: Upon seeing this struct and thinking "man, we must have this already, right?" I remembered we have `structs.NamespacedID` for just this purpose. Let's use that here.

tgross (commented):
This is looking great @geovannyAvelar. To get this merged:
* Run `make cl` to add a changelog entry
* Add documentation for the new command flag in `website/content/docs/commands/job/status.mdx`
* Mark the PR as ready-for-review

tgross (approved):
LGTM! 

I've made a minor adjustment to the changelog entry. Once CI is green I'll get this merged for the next 1.7.x release. Thanks @geovannyAvelar!

tgross:
Bah, sorry @geovannyAvelar, it looks like I led you astray with the `structs` package, as we have a linter than prevents that import (see https://github.com/hashicorp/nomad/pull/13938, which I approved :facepalm: ). I've fixed that up for you.

PR #18926, opened 2023-10-31
identity: version check multiple and implicit identities

tgross:
Job submitters cannot set multiple identities prior to Nomad 1.7, and cluster administrators should not set the identity configurations for their `consul` and `vault` configuration blocks until all servers have been upgraded. Validate these cases during job submission so as to prevent state store corruption when jobs are submitting in the middle of a cluster upgrade.

### DIFF
--- a/nomad/job_endpoint_hooks.go
+++ b/nomad/job_endpoint_hooks.go
@@ -381,19 +381,26 @@ func (v *jobValidate) Validate(job *structs.Job) (warnings []error, err error) {
 		multierror.Append(validationErrors, fmt.Errorf("job priority must be between [%d, %d]", structs.JobMinPriority, v.srv.config.JobMaxPriority))
 	}
 
+	okForIdentity := v.isEligibleForMultiIdentity()
+
 	for _, tg := range job.TaskGroups {
 		for _, s := range tg.Services {
-			serviceErrs := v.validateServiceIdentity(s, fmt.Sprintf("task group %s", tg.Name))
+			serviceErrs := v.validateServiceIdentity(
+				s, fmt.Sprintf("task group %s", tg.Name), okForIdentity)
 			multierror.Append(validationErrors, serviceErrs)
 		}
 
 		for _, t := range tg.Tasks {
+			if len(t.Identities) > 1 && !okForIdentity {
+				multierror.Append(validationErrors, fmt.Errorf("tasks can only have 1 identity block until all servers are upgraded to %s", minVersionMultiIdentities))
+			}
 			for _, s := range t.Services {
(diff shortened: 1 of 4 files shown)


nomad/job_endpoint_hooks.go
```diff
@@ -402,7 +409,19 @@ func (v *jobValidate) Validate(job *structs.Job) (warnings []error, err error) {
 	return warnings, validationErrors.ErrorOrNil()
 }
 
-func (v *jobValidate) validateServiceIdentity(s *structs.Service, parent string) error {
+func (v *jobValidate) isEligibleForMultiIdentity() bool {
+	if v.srv == nil || v.srv.serf == nil {
+		return true // handle tests w/o real servers safely
+	}
+	return ServersMeetMinimumVersion(
+		v.srv.Members(), v.srv.Region(), minVersionMultiIdentities, true)
```
schmichael, line 417: Would be nice if ServersMeetMinimumVersion was a struct that wrapped up `Members()` and `Region()` and only provided a simple `Meets(version, checkFailed)` interface for us to mock/stub out in tests.

nbd though, nothing to block this work on.

nomad/job_endpoint_hooks.go
```diff
@@ -402,7 +409,19 @@ func (v *jobValidate) Validate(job *structs.Job) (warnings []error, err error) {
...
+}
+
+func (v *jobValidate) validateServiceIdentity(s *structs.Service, parent string, okForIdentity bool) error {
+	if s.Identity != nil && !okForIdentity {
+		return fmt.Errorf("Service %s in %s cannot have an identity until all servers are upgraded to %s",
```
schmichael, line 422: Just a little pedantic nitpicking on the wording.

```suggestion
		return fmt.Errorf("Service %s in %s cannot have an identity until all servers are upgraded to >= %s",
```

or

```suggestion
		return fmt.Errorf("Service %s in %s cannot have an identity until all servers are upgraded to %s or later",
```

schmichael (approved):
Thanks. I think this is a great way to protect against truly wacky behavior in mixed version settings.

An implicit constraint on the client being >= 1.7.0 would be useful too.

nomad/job_endpoint_hooks.go
tgross: I've opened https://github.com/hashicorp/nomad/issues/18931 to follow-up on that when I have a few minutes.

tgross:
> An implicit constraint on the client being >= 1.7.0 would be useful too.

Agreed. I'll follow-up with that in a separate PR.

tgross:
Follow-up for the client version check in https://github.com/hashicorp/nomad/pull/18932

PR #18942, opened 2023-10-31
Vault small fixes

lgfa29:
Small fixes on Vault JWT auth:

* Remove `token_ttl` from `vaultcompat` since Nomad uses periodic tokens.
* Set minimum version required for JWT auth to 1.11.0+ because that's when `user_claim_json_pointer` was added.


tgross (approved):
LGTM!

PR #18943, opened 2023-10-31
identity: Implement `change_mode`

schmichael:
This implements the jobspec and client internals for `identity.change_mode`: allowing users to signal or restart tasks when their identity tokens get rotated.

## Identity Watcher Firstrun

As part of this I also blocked `taskrunner/identityHook.Prestart(...)` from completing until tokens are initially written (_or_ a timeout/context-cancellation is hit). I think before we technically had a race between `taskrunner/identityHook.Prestart(...)` writing initial tokens and the things that rely on those initial tokens starting up. Not sure if it would be realistically hit, but it seemed like a good idea to copy our `template` implementation of waiting for a first run.

Let me know if you think the 1 minute timeout should be changed or removed. `template` blocks indefinitely, but it has a lot more concerns than `identity`. I really couldn't think of a good reason to have the timer or not and was just guided by my desire to avoid dead/live-locking tasks without a good reason. If a task starts and immediately (well... after the timeout) fails due to a missing `identity` ... that seems preferable to blocking forever? At least the alloc could eventually be rescheduled elsewhere that might have a better chance to succeed?

## Empty String Default

I treat `""` and `"noop"` the same without actually Canonicalizing `"" -> "noop"` like other `change_mode`s do. I did this for 2 not weak reasons:

1. Why waste the bytes on the word "noop" when just omitting the entire field communicates the same thing?
2. We need to default to a coordinated `change_mode`. If folks end up with >1 TTL'd renewing identities per task, it's going to be hard to manage `change_mode` well. If you have 2 identities with the same TTL they will cause 2 restarts! Even worse: if you noop one, there's no telling if the other one will signal/restart right before or right after this one due to the jitter added. Add in `change_mode`s from other blocks and this could be a big headache.

Due to `#2` I'd almost rather folks just use `file = true; env = false`, and reload the token file when they get a 403 from their upstream service. :shrug: Us barraging their task with signals is probably more error prone than that.

### DIFF
--- a/client/allocrunner/taskrunner/identity_hook.go
+++ b/client/allocrunner/taskrunner/identity_hook.go
@@ -7,10 +7,13 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"time"
 
+	"github.com/hashicorp/consul-template/signals"
 	log "github.com/hashicorp/go-hclog"
 
 	"github.com/hashicorp/nomad/client/allocrunner/interfaces"
+	ti "github.com/hashicorp/nomad/client/allocrunner/taskrunner/interfaces"
 	"github.com/hashicorp/nomad/client/taskenv"
 	"github.com/hashicorp/nomad/client/widmgr"
 	"github.com/hashicorp/nomad/helper/users"
@@ -37,6 +40,7 @@ type identityHook struct {
 	task       *structs.Task
 	tokenDir   string
 	envBuilder *taskenv.Builder
+	lifecycle  ti.TaskLifecycle
 	ts         tokenSetter
 	widmgr     widmgr.IdentityManager
 	logger     log.Logger
@@ -52,6 +56,7 @@ func newIdentityHook(tr *TaskRunner, logger log.Logger) *identityHook {
 		task:       tr.Task(),
 		tokenDir:   tr.taskDir.SecretsDir,
 		envBuilder: tr.envBuilder,
+		lifecycle:  tr,
 		ts:         tr,
 		widmgr:     tr.widmgr,
 		stopCtx:    stopCtx,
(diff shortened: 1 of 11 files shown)


client/allocrunner/taskrunner/identity_hook.go
```diff
@@ -65,52 +70,138 @@ func (*identityHook) Name() string {
...
+
+	// Wait until every watcher ticks the first run chan
+	for i := range h.task.Identities {
+		select {
+		case <-firstRunCh:
+			// Identity fetched, loop
+		case <-deadlineTimer.C:
+			h.logger.Warn("timed out waiting for initial identity tokens to be fetched",
+				"num_fetched", i, "num_total", len(h.task.Identities))
+			return nil
```
tgross, line 101: The group-level `identity_hook` will trigger the WIDMgr and get the identities synchronously before we ever run this hook. So this deadline is only waiting on the WIDMgr broadcasting the notification that it got the signed identities. If we're waiting for over a minute for a send on a channel, are we likely to _ever_ get them? In which case, maybe we should return an error here so that the user sees this error rather than getting an error from downstream consumers?

This hook is early in the taskrunner so returning an error early will also prevent us from doing much more expensive setup work only to throw it away (ex. `artifact` and `volume` hooks won't get a chance to run).

contributing/checklist-jobspec.md
```diff
@@ -2,12 +2,12 @@
 
 ## Code
 
-* [ ] Consider similar features in Consul, Kubernetes, and other tools. Is there prior art we should match? Terminology, structure, etc?
-* [ ] Add structs/fields to `api/` package
+* [x] Consider similar features in Consul, Kubernetes, and other tools. Is there prior art we should match? Terminology, structure, etc?
+* [t] Add structs/fields to `api/` package
```
tgross, line 6: I think you left your working notes in here?

command/agent/job_endpoint.go
```diff
@@ -1653,11 +1640,13 @@ func apiWorkloadIdentityToStructs(in *api.WorkloadIdentity) *structs.WorkloadIde
 		return nil
 	}
 	return &structs.WorkloadIdentity{
-		Name:        in.Name,
-		Audience:    in.Audience,
-		Env:         in.Env,
-		File:        in.File,
-		ServiceName: in.ServiceName,
+		Name:         in.Name,
+		Audience:     in.Audience,
```
tgross, line 1644: Not a new bug, but now that I'm looking at this function again, shouldn't this be:

```suggestion
		Audience:     slices.Clone(in.Audience),
```

tgross (approved):
I've left a few notes but otherwise LGTM

client/allocrunner/taskrunner/identity_hook.go
tgross: (That being said, I wouldn't block this PR on this.)

client/allocrunner/taskrunner/identity_hook.go
```diff
@@ -65,52 +70,138 @@ func (*identityHook) Name() string {
...
+	// Don't block indefinitely for identities
+	deadlineTimer := time.NewTimer(time.Minute)
+	defer deadlineTimer.Stop()
+
+	// Wait until every watcher ticks the first run chan
+	for i := range h.task.Identities {
+		select {
+		case <-firstRunCh:
+			// Identity fetched, loop
+		case <-deadlineTimer.C:
```
shoenig, line 98: kind of makes me think there should be a context-aware version of WaitGroup

shoenig (approved):

client/allocrunner/taskrunner/identity_hook.go
schmichael: Ah that makes a ton more sense. So the only threat here is for Clients so busy it can't pluck some structs off a chan and throw them on disk (if file=true) in time. No racing the network or remote machines involved.

When nodes are rebooted they can be quite busy so I am happy to have this race fixed. 1 minute does seem like an eternity though, so perhaps short-circuiting with an error is better than making a best-effort by letting it trundle on. :thinking: 

It ought to be sufficiently unlikely to not matter, so I don't think I'm going to write up an issue until someone hits it? :thinking:

client/allocrunner/taskrunner/identity_hook.go
schmichael: 100% what I needed

PR #18945, opened 2023-11-01
consul: fix lookups of default cluster across upgrades

tgross:
Allocations that were created before Nomad 1.7 will not have the cluster field set for their Consul blocks. While this can be corrected server-side, that doesn't help allocations already on clients.

### DIFF
--- a/client/allocrunner/consul_hook.go
+++ b/client/allocrunner/consul_hook.go
@@ -180,7 +180,12 @@ func (h *consulHook) prepareConsulTokensForServices(services []*structs.Service,
 			continue
 		}
 
-		if err := h.getConsulTokens(service.Cluster, service.Identity.Name, tokens, req); err != nil {
+		cluster := service.Cluster
+		if cluster == "" {
+			cluster = structs.ConsulDefaultCluster
+		}
+
+		if err := h.getConsulTokens(cluster, service.Identity.Name, tokens, req); err != nil {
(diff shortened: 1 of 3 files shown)


client/allocrunner/consul_hook.go
```diff
@@ -180,7 +180,12 @@ func (h *consulHook) prepareConsulTokensForServices(services []*structs.Service,
 			continue
 		}
 
-		if err := h.getConsulTokens(service.Cluster, service.Identity.Name, tokens, req); err != nil {
+		cluster := service.Cluster
+		if cluster == "" {
+			cluster = structs.ConsulDefaultCluster
+		}
```
shoenig, line 186: just a thought - when there's special care needed when reading a value that's a good time to have a getter method which takes care of that logic, instead of repeating it at call sites which is prone for error, e.g.

```
cluster := service.GetCluster() // handles ""
```

shoenig (approved):

client/allocrunner/consul_hook.go
tgross: Yeah, totally agreed. We actually implemented a `GetConsulClusterName` method on both task and service but unfortunately it's not currently safe to call it at this call site in ENT (which is a bug that I'm going to follow-up on as part of https://github.com/hashicorp/nomad/issues/18947)

PR #18955, opened 2023-11-01
consul/vault: use accessor method to get cluster name in client

tgross:
When looking up the Consul or Vault cluster from a client hook, we should always use an accessor function rather than trying to lookup the `Cluster` field, which may be empty for jobs registered before Nomad 1.7.

Follow-up to https://github.com/hashicorp/nomad/pull/18945 and https://github.com/hashicorp/nomad/pull/18940


shoenig (approved):
LGTM! pending tests

PR #18957, opened 2023-11-01
template: fix test assertion to be compatible between CE/ENT

tgross:
The template hook emits an error when the task has a Consul block that requires WI but there's no WI. The exact error message we get depends on whether we're running in CE or ENT. Update the test assertion so that we can tolerate this difference without building ENT-specific test files.


pkazmierczak (approved):
LGTM!

lgfa29 (approved):
`TestCoreScheduler_CSIBadState_ClaimGC` may be flaky?

tgross:
> TestCoreScheduler_CSIBadState_ClaimGC may be flaky?

Yeah, I started seeing it fail intermittently after the keyring change in https://github.com/hashicorp/nomad/pull/18882 got merged. I suspect it's a timing problem or a race in the test. I can pick that up.

Issue #18959, opened 2023-11-01
`TestCoreScheduler_CSIBadState_ClaimGC` is flaky


tgross:
Seems like this started failing intermittently after #18882 got merged. Writing this down as an issue so that I remember to follow-up with a fix in the next day or so.

PR #18960, opened 2023-11-01
test: make CSI bad state GC test synchronous

tgross:
One of our core scheduler tests for GC tests that volumes with invalid allocations immediately have those claims marked as past claims and puts them into the unpublishing state. This happens synchronously with the GC evaluation processing, so there's no need for us to wait for the results.

Fixes: #18959


shoenig (approved):
LGTM!

gulducat (approved):
The time has come. Eventually is now.

PR #18964, opened 2023-11-01
identity: don't generate codec for oidc config

schmichael:
Our codec code generation doesn't honor `json:"..."` tags which breaks the OIDC Discovery endpoint.

This adds the relevant struct to the code generators ignore list (just like we have done with IdentityClaims).


tgross (approved):
LGTM!

PR #18965, opened 2023-11-01
vault: fix import cycle in `vaultclient`

lgfa29:
This PR reverts https://github.com/hashicorp/nomad/pull/18652 to avoid making all of those fields public, which exposed many sensitive values, such as `running`, `heap` and `lock`, and this scares me 😅 

It instead fixes the import cycle in https://github.com/hashicorp/nomad/commit/beceb78ccfa12bdd0325418e3c9a6d24415cad63, by avoiding the import of `github.com/hashicorp/nomad/client/config` from `client/vaultclient/vaultclient_test.go`. 

The only reason this package was being imported was to read `config.DefaultConfig()`, but in reality only `structsc.DefaultVaultConfig()` was being used, which does not create an import cycle.

Apologies in advance to future code spelunkers that land in this PR, I was not able to keep the git history on `vaultclient_test.go`.


pkazmierczak (approved):
LGTM!

tgross (approved):
:+1:

PR #18967, opened 2023-11-01
Set correct submit time to stopped jobs

### LINKED ISSUE #18966: Stopped job versions have the same submit time as the previous version
### Nomad version
Nomad v1.6.3 (but happens also at 1.5.X)

### Issue

When stopping a job, the submit time for that version takes the previous submit time (of the last version), so it appears the job wan't stopped at the time it was, but just after the last change.

### Reproduction steps

1) run job
2) stop job (without purge)

#### Expected Result
The version for the stopped job has the correct time (when the job was set to stop)

#### Actual Result
The stopped job has the submit time of the last previous version

stop

jorgemarey:
Fixes #18966

### DIFF
--- a/nomad/fsm.go
+++ b/nomad/fsm.go
@@ -838,6 +838,7 @@ func (n *nomadFSM) handleJobDeregister(index uint64, jobID, namespace string, pu
 		}
 
 		stopped := current.Copy()
+		stopped.SetSubmitTime()
 		stopped.Stop = true
 
 		if err := n.state.UpsertJobTxn(index, nil, stopped, tx); err != nil {
--- a/nomad/fsm.go
+++ b/nomad/fsm.go
@@ -726,7 +726,7 @@ func (n *nomadFSM) applyDeregisterJob(msgType structs.MessageType, buf []byte, i
 	}
 
 	err := n.state.WithWriteTransaction(msgType, index, func(tx state.Txn) error {
-		err := n.handleJobDeregister(index, req.JobID, req.Namespace, req.Purge, req.NoShutdownDelay, tx)
+		err := n.handleJobDeregister(index, req.JobID, req.Namespace, req.Purge, req.SubmitTime, req.NoShutdownDelay, tx)
 
 		if err != nil {
 			n.logger.Error("deregistering job failed",
@@ -766,7 +766,7 @@ func (n *nomadFSM) applyBatchDeregisterJob(msgType structs.MessageType, buf []by
 	// evals for jobs whose deregistering didn't get committed yet.
 	err := n.state.WithWriteTransaction(msgType, index, func(tx state.Txn) error {
 		for jobNS, options := range req.Jobs {
-			if err := n.handleJobDeregister(index, jobNS.ID, jobNS.Namespace, options.Purge, false, tx); err != nil {
+			if err := n.handleJobDeregister(index, jobNS.ID, jobNS.Namespace, options.Purge, req.SubmitTime, false, tx); err != nil {
 				n.logger.Error("deregistering job failed", "job", jobNS.ID, "error", err)
 				return err
 			}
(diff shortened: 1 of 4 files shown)


nomad/fsm.go
```diff
@@ -838,6 +838,7 @@ func (n *nomadFSM) handleJobDeregister(index uint64, jobID, namespace string, pu
 		}
 
 		stopped := current.Copy()
+		stopped.SetSubmitTime()
```
lgfa29, line 841: FSM operations must be deterministic since all servers apply every Raft log and must reach the exact same state. Since `SetSubmitTime()` calls `time.Now()` each server would reach a different value for this.

That's why for non-deterministic values like this we set them at the RPC layer:
https://github.com/hashicorp/nomad/blob/1ef99f05364b7d3739befa6a789f0d55b2314dcf/nomad/job_endpoint.go#L342-L348

So we will need to do the same for the `Deregister` method.

lgfa29 (changes requested):
Thanks for the PR @jorgemarey!

As I mentioned in https://github.com/hashicorp/nomad/issues/18966#issuecomment-1793285903 I spoke with the team and, while not the ideal field name, it does make sense to update it since a stop is technically a new job version submission.

The other point raised is that the `Scale()` endpoint is similar to `Deregister()` in that it also creates a new job version, and the `SubmitTIme` is not updated either. Would you mind applying the same fix there as well?

And should probably also clarify the godocs of `SubmitTime` to specify that `SubmitTime is the time at which the job _version_ was submitted`.
https://github.com/hashicorp/nomad/blob/1ef99f05364b7d3739befa6a789f0d55b2314dcf/nomad/structs/structs.go#L4455-L4457

Thanks again for the PR and let me know if you have any questions!

jorgemarey:
Hi, @lgfa29 I'll do it as soon as I can.

jorgemarey:
Hi @lgfa29 I'm looking into this, but I have a question about how to solve the deregister. 

Currently the fsm **applyDeregisterJob** method only receives the jobID (and other fields, but no complete job). Should I modify the **JobDeregisterRequest** to add a submittime or get the job in the Deregister() method and provide that in the **JobDeregisterRequest**? 

Another solution could be to use the time in the provided eval but I don't know if that solution is correct as maybe no eval is provided?

lgfa29:
Oof...I should've noticed that 😅 

But good catch. I think the best option here is to expand `JobDeregisterRequest` to include `SubmitTime`. The eval seem `nil` for non-system jobs.

One important thing to check in `fsm.go` is to only update the job's `SubmitTime` if it's non-zero. Since these `*Request` structs are what's written to Raft, old entries won't have this field set, so they cause jobs to be updated with a zero `SubmitTIme`.

This also means that the `SubmitTime` will still be wrong until all servers in the cluster are updated, and the fix will only be applied to new `job stop` events.

jorgemarey:
Hi @lgfa29 any news on this? Should I change anything else?

lgfa29 (approved):
Hi @jorgemarey, apologies for the delay here. 1.7 has kept as busy for the past weeks and I missed that you had pushed a commit.

GitHub is also not great at communicating status of things (even though it's kind of its job 😅), so just an FYI in case you haven't seen this yet, there's a tiny 🔄 button next to reviewers name that you can click to request another round of review. 
image

It's super easy to miss, but it does let us know when you're done with changes.

But back to the PR, the changes look great now, thanks you for the PR!

jorgemarey:
Hi @lgfa29,

> there's a tiny 🔄 button next to reviewers name that you can click to request another round of review. 


Oh, I didn't know that. Thanks for the information.

> But back to the PR, the changes look great now, thanks you for the PR!

Great!!

lgfa29:
Yeah, it also took me a while to notice it. It should be a big visible button that appears once you make changes after a review, but alas, that's what we have for now 😅

PR #18972, opened 2023-11-02
acl/client: fix incorrect denied error on calls with dangling policies.

### LINKED ISSUE #18619: Nomad ACL roles with dangling policies cause permission evaluation issues
### Nomad version
```
Nomad v1.6.2
```

### Issue

https://github.com/hashicorp/nomad/issues/17201 addressed the issue of roles with overlapping policies but there's another issue I stumbled upon related to a role containing one or more policies which have been deleted, which causes the list of allowed permissions to be erroneously computed; ultimately this results in allowed actions being denied to the user.
The involved code path is the same as for #17201, ACL.GetPolicies.

### Reproduction steps

1. Create two namespaces `monitoring` and `system` and deploy a simple job in the `system` one.
2. Create two policies:
```
cat > policy1.hcl << EOF
operator {
  policy = "read"
}

namespace "monitoring" {
  policy = "write"
}
EOF

cat > policy2.hcl << EOF
operator {
  policy = "read"
}

namespace "system" {
  policy = "write"
}
EOF

nomad acl policy apply monitoring-admin policy1.hcl
nomad acl policy apply system-admin policy2.hcl
```
3. Create a role and get a token:
```
nomad acl role create -name test-role -policy system-admin -policy monitoring-admin
export NOMAD_TOKEN=$(nomad acl token create -ttl=10m -name=test-token -role-name=test-role -t '{{ .SecretID }}')
```
4. Exec in the test container
```bash
~ nomad alloc exec -namespace='system' -job my-test-job echo foo
foo
~
```
5. With another token, delete the `monitoring-admin` policy
```
nomad acl policy delete monitoring-admin
```
6. Re-try to exec
```bash
~ nomad alloc exec -namespace='system' -job my-test-job echo foo
failed to exec into task: rpc error: Permission denied
~ 
```

#### Expected Result
```bash
~ nomad alloc exec -namespace='system' -job my-test-job echo foo
foo
~
```

#### Actual Result
```bash
~ nomad alloc exec -namespace='system' -job my-test-job echo foo
failed to exec into task: rpc error: Permission denied
~
```

jrasell:
When a user performs a client API call, the Nomad client will perform an RPC which looks up the ACL policies which the callers ACL token is assigned. If the ACL token includes dangling (deleted) policies, the call would previously fail with a permission denied error.

This change ensures this error is not returned and that the lookup will succeed in the event of dangling policies.

closes #18619

### DIFF
--- a/.changelog/18972.txt
+++ b/.changelog/18972.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+client: Fixed a bug where client API calls would fail incorrectly with permission denied errors when using ACL tokens with dangling policies
+```
--- a/nomad/acl_endpoint.go
+++ b/nomad/acl_endpoint.go
@@ -389,28 +389,33 @@ func (a *ACL) GetPolicies(args *structs.ACLPolicySetRequest, reply *structs.ACLP
 	// Add the token policies which are directly referenced into the set.
 	tokenPolicyNames.InsertSlice(token.Policies)
 
-	// Ensure the token has enough permissions to query the named policies.
-	if token.Type != structs.ACLManagementToken && !tokenPolicyNames.ContainsSlice(args.Names) {
-		return structs.ErrPermissionDenied
-	}
-
 	// Setup the blocking query
 	opts := blockingOptions{
 		queryOpts: &args.QueryOptions,
 		queryMeta: &reply.QueryMeta,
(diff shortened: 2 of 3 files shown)


tgross (approved):
LGTM :+1:

pkazmierczak (approved):
👍

PR #18985, opened 2023-11-02
vault: avoid continual renewal of invalid token

lgfa29:
A series of errors may happen when a token is invalidated while the Vault client is waiting to renew it. The token may have been invalidated for several reasons, such as the alloc finished running and it's now terminal or the token may have been change directly on Vault out-of-band.

Most of the errors are caused by retries that will never succeed until Vault fully removes the token from its state.

This commit prevents the retries by making the error `invalid lease ID` a fatal error.

In earlier versions of Vault, this case was covered by the error `lease not found or lease is not renewable`, which is already considered to be a fatal error by Nomad:

https://github.com/hashicorp/vault/blob/2d0cde4ccc0323591d9414342cb15f5cb70271d7/vault/expiration.go#L636-L639

But https://github.com/hashicorp/vault/pull/5346 introduced an earlier `nil` check that generates a different error message:

https://github.com/hashicorp/vault/blob/750ab337eaa0b049d9cf1535c00e860129e5e9a0/vault/expiration.go#L1362-L1364

Both errors happen for the same reason (`le == nil`) and so should be considered fatal on renewal.

Closes https://github.com/hashicorp/nomad/issues/12465

### DIFF
--- a/client/vaultclient/vaultclient.go
+++ b/client/vaultclient/vaultclient.go
@@ -453,6 +453,7 @@ func (c *vaultClient) renew(req *vaultClientRenewalRequest) error {
 	fatal := false
 	if renewalErr != nil &&
 		(strings.Contains(renewalErr.Error(), "lease not found or lease is not renewable") ||
+			strings.Contains(renewalErr.Error(), "invalid lease ID") ||
 			strings.Contains(renewalErr.Error(), "lease is not renewable") ||
 			strings.Contains(renewalErr.Error(), "token not found") ||
 			strings.Contains(renewalErr.Error(), "permission denied")) {
--- a/.changelog/18985.txt
+++ b/.changelog/18985.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+vault: Fixed an issue that could cause Nomad to attempt to renew a Vault token that is already expired
+```


client/vaultclient/vaultclient.go
```diff
@@ -370,6 +370,11 @@ func (c *vaultClient) RenewToken(token string, increment int) (<-chan error, err
 		increment: increment,
 	}
 
+	// Add renewal request to the heap to start tracking the token.
+	c.lock.Lock()
+	c.heap.Push(renewalReq, time.Now())
+	c.lock.Unlock()
```
tgross, line 376: Should we remove the original [`heap.Push`](https://github.com/hashicorp/nomad/pull/18985/files#diff-61b0f81ca4505d23bb2adf7025d2c36c4e02232a2f13ac39dce779617c954416R518-R523) or at least change this comment?

```go
// If the identifier is not already tracked, this is a first
// renewal request. In this case, add an entry into the heap
// with the next renewal time.
```

Actually, is that whole branch now an unreachable state because the [`if c.isTracked(req.id)`](https://github.com/hashicorp/nomad/pull/18985/files#diff-61b0f81ca4505d23bb2adf7025d2c36c4e02232a2f13ac39dce779617c954416R479) conditional will always be true?

tgross (approved):
LGTM! Nice work on this @lgfa29!

client/vaultclient/vaultclient.go
lgfa29: Hum...good point. Looking at the code again, I think this change can actually add a _new_ racing condition where the token is renewed twice!

When `RenewToken()` pushes the token to heap the `run()` loop may pick it up for renewal right away, but then `RenewToken()` calls `c.renew()` for an immediate renewal, resulting in a double renew. 

I don't _think_ that's a problem but it's certainly unnecessary. 

I think we could make it so only the `run()` loop actually renew tokens and `RenewToken()` blocks until the renewal request completes, but that could add some latency to task start.

Another option is to revert the new `!c.isTracked` check and just adjust the error so it's considered `fatal` and we only get one error max. 

The `select` race condition that it tries to address seems rather unlikely to happen, so a single sporadic error may not be too bad? 

Another option would be to have different behaviours if `c.renew()` is called from `RenewToken()` or `run()`, like a new `renewUntracked()` function that doesn't check the heap 🤔

tgross: > I think we could make it so only the run() loop actually renew tokens and RenewToken() blocks until the renewal request completes, but that could add some latency to task start.

I like this idea, because trying to call `renew` from two different places seems to be the root of the race. The only potential added latency would come from another token/lease being renewed at the same time as the task starts, so that it hits the `run` loop while that loop is otherwise occupied. But I'm pretty sure we already have that same latency today, because only one token/lease can be renewed at the same time because `renew` takes the mutex.

lgfa29: Cool! Implemented this in https://github.com/hashicorp/nomad/pull/18998.

Since it's a fair amount of changes I think it may be better to avoid backporting it. I then updated this PR to only include the extra error check which will prevent the multiple renewals and is much safer to backport.

Issue #18990, opened 2023-11-03
job submission hook on whether Vault is enabled only checks default cluster


tgross:
While working on https://github.com/hashicorp/nomad/issues/18947 I noticed that for Nomad Enterprise 1.7.0-beta.1, the job submission hook for Vault only checks the `enabled` flag on the default cluster, rather than the clusters that are used by the job being submitted. This will return an error on job registration saying that Vault is disabled.

tgross:
Will be fixed by https://github.com/hashicorp/nomad/pull/18991

Issue #18978, opened 2023-11-02
`consul.service_identity` and `consul.task_identity` only use default Consul cluster values


tgross:
While working on https://github.com/hashicorp/nomad/issues/18947 I noticed that for Nomad Enterprise 1.7.0-beta.1, it looks like the `consul.service_identity` and `consul.task_identity` are only ever taken from default Consul block. This will prevent the cluster admin from setting different identity configs across multiple Consul clusters. (It works fine for Vault though.)

tgross:
Will be fixed by https://github.com/hashicorp/nomad/pull/18991

Issue #18947, opened 2023-11-01
remove unused non-multi-cluster Consul and Vault config blocks


tgross:
While working on support for multiple Consul and Vault clusters, I left the original configuration blocks in so that I could implement the feature in manageable chunks. The default cluster is synchronized from the `Config.ConsulConfigs[]` block to the `Config.Consul` block (and same for Vault) at the time of configuration parsing. We can now safely remove this, but I didn't get a chance to do so before we ship the beta.

Also, while doing this I'll also implement an accessor function wherever possible rather than a slice index so as to avoid problems like the one in #18940 and #18945.

PR #18991, opened 2023-11-03
config: remove old Vault/Consul config blocks from server

tgross:
Remove the now-unused original configuration blocks for Consul and Vault from the server. When the server needs to refer to a Consul or Vault block it will always be for a specific cluster for the task/service. Add a helper for accessing the default clusters (for the servers' own use).

This is one of three changesets for this work. The remainder will implement the same changes in the `client` package and on the `command/agent` package.

As part of this work I discovered two bugs:
* The job submission hook for Vault only checks the enabled flag on the default cluster, rather than the clusters that are used by the job being submitted. This will return an error on job registration saying that Vault is disabled. Fix that to check only the cluster(s) used by the job.
* The `consul.service_identity` and `consul.task_identity` are only ever taken from default Consul block. This will prevent the cluster admin from setting different identity configs across multiple Consul clusters. (It works fine for Vault though.)

Ref: https://github.com/hashicorp/nomad/issues/18947
Ref: https://github.com/hashicorp/nomad/pull/18994
Fixes: https://github.com/hashicorp/nomad/issues/18990
Fixes: https://github.com/hashicorp/nomad/issues/18978

### DIFF
--- a/nomad/config.go
+++ b/nomad/config.go
@@ -307,16 +307,10 @@ type Config struct {
 	// of all the heartbeats.
 	FailoverHeartbeatTTL time.Duration
 
-	// ConsulConfig is this Agent's Consul configuration
-	ConsulConfig *config.ConsulConfig
-
 	// ConsulConfigs is a map of Consul configurations, here to support features
 	// in Nomad Enterprise. The default Consul config pointer above will be
 	// found in this map under the name "default"
 	ConsulConfigs map[string]*config.ConsulConfig
-
-	// VaultConfig is this Agent's default Vault configuration
-	VaultConfig *config.VaultConfig
 
 	// VaultConfigs is a map of Vault configurations, here to support features
 	// in Nomad Enterprise. The default Vault config pointer above will be found
@@ -457,9 +451,7 @@ func (c *Config) Copy() *Config {
 	nc.RaftConfig = pointer.Copy(c.RaftConfig)
 	nc.SerfConfig = pointer.Copy(c.SerfConfig)
 	nc.EnabledSchedulers = slices.Clone(c.EnabledSchedulers)
-	nc.ConsulConfig = c.ConsulConfig.Copy()
 	nc.ConsulConfigs = helper.DeepCopyMap(c.ConsulConfigs)
-	nc.VaultConfig = c.VaultConfig.Copy()
 	nc.VaultConfigs = helper.DeepCopyMap(c.VaultConfigs)
 	nc.TLSConfig = c.TLSConfig.Copy()
 	nc.SentinelConfig = c.SentinelConfig.Copy()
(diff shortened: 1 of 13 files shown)


nomad/config.go
```diff
@@ -472,22 +464,24 @@ func (c *Config) Copy() *Config {
 
 // ConsulServiceIdentity returns the workload identity to be used for accessing
 // the Consul API to register and manage Consul services.
-func (c *Config) ConsulServiceIdentity() *structs.WorkloadIdentity {
-	if c.ConsulConfig == nil {
+func (c *Config) ConsulServiceIdentity(cluster string) *structs.WorkloadIdentity {
+	conf := c.ConsulConfigs[cluster]
```
lgfa29, line 468: Would it be worth to have a `GetConsulConfigs` function like its done in the client side?

nomad/job_endpoint.go
```diff
@@ -246,7 +246,7 @@ func (j *Job) Register(args *structs.JobRegisterRequest, reply *structs.JobRegis
 	//   - reading kv store of each group
 	//   - establishing consul connect services
 	checkConsulToken := func(usages map[string]*structs.ConsulUsage) error {
-		if j.srv.config.ConsulConfig.AllowsUnauthenticated() {
+		if j.srv.config.GetDefaultConsul().AllowsUnauthenticated() {
```
lgfa29, line 249: We could modify the import to be a `map[string][string]*structs.ConsulUsage` where the first key is the Consul cluster, but I think this is used for the deprecated `-consul-token` right? So probably not worth modifying it.

nomad/job_endpoint_test.go
```diff
@@ -1573,7 +1573,7 @@ func TestJobEndpoint_Register_Vault_Disabled(t *testing.T) {
 	// Fetch the response
 	var resp structs.JobRegisterResponse
 	err := msgpackrpc.CallWithCodec(codec, "Job.Register", req, &resp)
-	if err == nil || !strings.Contains(err.Error(), "Vault not enabled") {
+	if err == nil || !strings.Contains(err.Error(), "Vault \"default\" not enabled") {
```
lgfa29, line 1576:
```suggestion
	if err == nil || !strings.Contains(err.Error(), `Vault "default" not enabled`) {
```

lgfa29 (approved):
Some minor comments and thoughts

nomad/job_endpoint.go
tgross: Exactly!

nomad/config.go
tgross: After this PR, the only callers that de-index `ConsulConfigs` are here in `nomad/config.go`, so it doesn't really get us any new abstractions to do that. The client-side version does some logic around ex. whether Vault is enabled.

Issue #18984, opened 2023-11-02
Consul config for template only uses default Consul cluster


tgross:
While working on https://github.com/hashicorp/nomad/issues/18947 I noticed that for Nomad Enterprise 1.7.0-beta.1, it looks like Consul configuration we use from templates always comes from the default Consul cluster's configuration, but will use the correct Consul token for the non-default cluster. This will prevent templates from being used with the non-default cluster.

tgross:
Will be fixed by https://github.com/hashicorp/nomad/pull/18994

PR #18994, opened 2023-11-03
config: remove old Vault/Consul config blocks from client

tgross:
Remove the now-unused original configuration blocks for Consul and Vault from the client. When the client needs to refer to a Consul or Vault block it will always be for a specific cluster for the task/service. Add a helper for accessing the default clusters (for the client's own use).

This is two of three changesets for this work. The remainder will implement the same changes in the `command/agent` package.

As part of this work I discovered and fixed two bugs:

* The gRPC proxy socket that we create for Envoy is only ever created using the default Consul cluster's configuration. This will prevent Connect from being used with the non-default cluster.
* The Consul configuration we use for templates always comes from the default Consul cluster's configuration, but will use the correct Consul token for the non-default cluster. This will prevent templates from being used with the non-default cluster.

Ref: https://github.com/hashicorp/nomad/issues/18947
Fixes: https://github.com/hashicorp/nomad/issues/18984
Fixes: https://github.com/hashicorp/nomad/issues/18978


client/allocrunner/alloc_runner_hooks.go
```diff
@@ -145,8 +146,8 @@ func (ar *allocRunner) initRunnerHooks(config *clientconfig.Config) error {
 			logger:            hookLogger,
 			shutdownDelayCtx:  ar.shutdownDelayCtx,
 		}),
-		newConsulGRPCSocketHook(hookLogger, alloc, ar.allocDir, config.ConsulConfig, config.Node.Attributes),
-		newConsulHTTPSocketHook(hookLogger, alloc, ar.allocDir, config.ConsulConfig),
+		newConsulGRPCSocketHook(hookLogger, alloc, ar.allocDir, config.ConsulConfigs, config.Node.Attributes),
+		newConsulHTTPSocketHook(hookLogger, alloc, ar.allocDir, config.ConsulConfigs),
```
lgfa29, line 150: Should these be using `GetConsulConfigs()`?

client/allocrunner/consul_grpc_sock_hook.go
```diff
@@ -184,55 +185,78 @@ func (p *grpcSocketProxy) run(alloc *structs.Allocation) error {
 	default:
 	}
 
-	// make sure either grpc or http consul address has been configured
-	if p.config.GRPCAddr == "" && p.config.Addr == "" {
-		return errors.New("consul address must be set on nomad client")
+	clusterNames := set.New[string](0)
```
lgfa29, line 188:
```suggestion
	clusterNames := set.New[string](1)
```

We probably expect to find at least one cluster right? 😅

client/allocrunner/consul_grpc_sock_hook.go
```diff
@@ -184,55 +185,78 @@ func (p *grpcSocketProxy) run(alloc *structs.Allocation) error {
...
-			return fmt.Errorf("error parsing Consul address %q: %v",
-				p.config.Addr, err)
+	var retErr error
+
+	clusterNames.ForEach(func(clusterName string) bool {
+		config := p.configs[clusterName]
+
+		// make sure either grpc or http consul address has been configured
+		if config.GRPCAddr == "" && config.Addr == "" {
+			retErr = errors.New("consul address must be set on nomad client")
```
lgfa29, line 201:
```suggestion
			retErr = fmt.Errorf("consul address for cluster %s must be set on nomad client", clusterName)
```

And would it make sense to use a multi-error here so we validate all possible errors instead of one by one?

client/allocrunner/consul_grpc_sock_hook.go
```diff
@@ -184,55 +185,78 @@ func (p *grpcSocketProxy) run(alloc *structs.Allocation) error {
...
-	// chmod afterwards.
-	// https://github.com/golang/go/issues/11822
-	if err := os.Chmod(hostGRPCSocketPath, os.ModePerm); err != nil {
-		return fmt.Errorf("unable to set permissions on unix socket for Consul gRPC endpoint: %v", err)
-	}
+		go func() {
+			proxy(p.ctx, p.logger, destAddr, listener)
+			p.cancel()
+			close(p.doneCh)
+		}()
```
lgfa29, line 252: Hum...this func makes me think we're iterating at the wrong level 🤔 

From what I understood from this part of the code, a `grpcSocketProxy` is used to expose Consul's gRPC endpoint to the task. By iterating over clusters here we would be overwriting `hostGRPCSocketPath` so that only the last cluster is being reached over this socket.

And, when the `proxy()` function returns here we will `close(p.doneCh)` multiple times, causing a `panic`.

I feel like the correct behaviour would be for `consulGRPCSocketHook` to hold a map of proxies, one for each cluster used in the alloc. Then each proxy is started pointing to a single Consul config.

client/allocrunner/consul_http_sock_hook.go
```diff
@@ -152,35 +153,57 @@ func (p *httpSocketProxy) run(alloc *structs.Allocation) error {
...
+		if err := os.Chmod(hostHTTPSockPath, os.ModePerm); err != nil {
+			retErr = fmt.Errorf("unable to set permissions on unix socket: %w", err)
+			return false
+		}
+
+		go func() {
+			proxy(p.ctx, p.logger, destAddr, listener)
+			p.cancel()
+			close(p.doneCh)
+		}()
```
lgfa29, line 199: Same thought from the gRPC poxy above.

client/allocrunner/taskrunner/task_runner_getters.go
```diff
@@ -72,7 +72,7 @@ func (tr *TaskRunner) setVaultToken(token string) {
 	// Update the task's environment
 	taskNamespace := tr.task.Vault.Namespace
 
-	ns := tr.clientConfig.VaultConfig.Namespace
+	ns := tr.clientConfig.VaultConfigs[tr.task.GetVaultClusterName()].Namespace
```
lgfa29, line 75:
```suggestion
	ns := tr.clientConfig.GetVaultConfigs(tr.logger)[tr.task.GetVaultClusterName()].Namespace
```

I guess that, at this point, it's safe to assume the client config exists?

client/allocrunner/taskrunner/task_runner_hooks.go
```diff
@@ -142,9 +142,11 @@ func (tr *TaskRunner) initHooks() {
 	// If this is a Connect sidecar proxy (or a Connect Native) service,
 	// add the sidsHook for requesting a Service Identity token (if ACLs).
 	if task.UsesConnect() {
+		tg := tr.Alloc().Job.LookupTaskGroup(tr.Alloc().TaskGroup)
+
 		// Enable the Service Identity hook only if the Nomad client is configured
 		// with a consul token, indicating that Consul ACLs are enabled
-		if tr.clientConfig.ConsulConfig.Token != "" {
+		if tr.clientConfig.ConsulConfigs[task.GetConsulClusterName(tg)].Token != "" {
```
lgfa29, line 149:
```suggestion
		if tr.clientConfig.GetConsulConfigs(tr.logger)[task.GetConsulClusterName(tg)].Token != "" {
```

client/config/config_ce.go
lgfa29, line 1: I think we could use the new `GetDefaultVault()` and `GetDefaultConsul()` helpers here?

lgfa29 (commented):
The Consul socket hooks smell a bit funny to me...it's a part of the code that I'm not too familiarized, but I think it's worth a double-check.

client/allocrunner/consul_grpc_sock_hook.go
tgross: Oof, good catch.

client/allocrunner/taskrunner/task_runner_getters.go
tgross: Yeah, `clientConfig` has to be set to create the taskrunner. And the `VaultConfig` we get should always be non-nil because the fingerprint from that config is what tells the server it can place this alloc on this node.

tgross: Oh no! But if the user _removes_ a Vault config after the allocation has been placed and then restarts the client, the client would potentially crash during restore instead of just killing the non-working allocation. Shoot, I'm going to have to rework this approach a good bit.

tgross: I'm going to pull that part of the work into https://github.com/hashicorp/nomad/issues/19000 just to keep this PR a manageable size for you to review @lgfa29

client/allocrunner/consul_grpc_sock_hook.go
tgross: Done, but in the outermost function after making the map of proxies as suggested.

client/allocrunner/consul_grpc_sock_hook.go
tgross: Fixed!

client/config/config_ce.go
tgross: Going to address this with the rest of #19000 so we can get a better handle on config getters.

client/allocrunner/consul_http_sock_hook.go
```diff
@@ -152,13 +187,24 @@ func (p *httpSocketProxy) run(alloc *structs.Allocation) error {
 	default:
 	}
 
+	clusterNames := set.New[string](0)
+	tg := alloc.Job.LookupTaskGroup(alloc.TaskGroup)
+	for _, s := range tg.Services {
+		clusterNames.Insert(s.GetConsulClusterName(tg))
+	}
+
```
lgfa29, line 195: I think this is leftover code from the previous implementation.

lgfa29 (approved):
LGTM! I think just needs to clean-up the leftover code

client/allocrunner/taskrunner/task_runner_getters.go
lgfa29: Ah good catch!

PR #18997, opened 2023-11-03
config: remove old Vault/Consul config blocks from parser

### LINKED ISSUE #19000: removing non-default Vault/Consul config can cause panic on client restart (ENT)
On Nomad Enterprise 1.7.0-beta.1, if you remove a non-default Vault or Consul configuration from a client and then restart the client, there are a number of places where we don't handle this gracefully and end up with a nil pointer exception in the allocrunner/taskrunner.

Create a configuration getter function that handles this gracefully by returning an error, which the caller will then be responsible for picking up. This will probably require pushing the getter function into the hook configurations, rather than the whole set of configurations.

tgross:
Remove the now-unused original configuration blocks for Consul and Vault from the agent configuration parsing. When the agent needs to refer to a Consul or Vault block it will always be for a specific cluster for the task/service (or the default cluster for the agent's own use).

This is third of three changesets for this work. It includes:
* A fix to the `Merge` methods where the `Name` field was not being correctly merged.
* Changing the map of Vault/Consul configs in the agent's `Config` to a slice. This doesn't change the resulting maps in the server/client `Config` because it's a lot easier to use them at the call site that way.
* Fixes to tests that were masking problems in parsing JSON format agent configuration.

Fixes: https://github.com/hashicorp/nomad/issues/18947
Ref: https://github.com/hashicorp/nomad/pull/18991
Ref: https://github.com/hashicorp/nomad/pull/18994

(Note this does not fix #19000 because this is already a huge PR and all of that fix will happen in the `client` package anyways.)

### DIFF
--- a/command/agent/agent.go
+++ b/command/agent/agent.go
@@ -518,15 +518,15 @@ func convertServerConfig(agentConfig *Config) (*nomad.Config, error) {
 		return nil, fmt.Errorf("server_service_name must be set when auto_advertise is enabled")
 	}
 
-	// handle system scheduler preemption default
-	if agentConfig.Server.DefaultSchedulerConfig != nil {
-		conf.DefaultSchedulerConfig = *agentConfig.Server.DefaultSchedulerConfig
-	}
-
 	conf.VaultConfigs = helper.SliceToMap[map[string]*config.VaultConfig](
 		agentConfig.Vaults,
 		func(cfg *config.VaultConfig) string { return cfg.Name },
 	)
+
+	// handle system scheduler preemption default
+	if agentConfig.Server.DefaultSchedulerConfig != nil {
+		conf.DefaultSchedulerConfig = *agentConfig.Server.DefaultSchedulerConfig
+	}
 
 	// Set the TLS config
 	conf.TLSConfig = agentConfig.TLSConfig
@@ -1164,13 +1164,7 @@ func (a *Agent) agentHTTPCheck(server bool) *structs.ServiceCheck {
 	// Resolve the http check address
 	httpCheckAddr := a.config.normalizedAddrs.HTTP[0]
 
-	var defaultConsul *config.ConsulConfig
-	for _, consulCfg := range a.config.Consuls {
-		if consulCfg.Name == structs.ConsulDefaultCluster {
-			defaultConsul = consulCfg
-			break
-		}
-	}
+	defaultConsul := a.config.defaultConsul()
 	if defaultConsul == nil {
 		return nil
 	}
--- a/command/agent/agent.go
+++ b/command/agent/agent.go
@@ -507,7 +507,14 @@ func convertServerConfig(agentConfig *Config) (*nomad.Config, error) {
 		conf.FailoverHeartbeatTTL = failoverTTL
 	}
 
-	if *agentConfig.Consul.AutoAdvertise && agentConfig.Consul.ServerServiceName == "" {
+	// Add the Consul and Vault configs
+	conf.ConsulConfigs = helper.SliceToMap[map[string]*config.ConsulConfig](
+		agentConfig.Consuls,
+		func(cfg *config.ConsulConfig) string { return cfg.Name },
+	)
+
+	consul := conf.ConsulConfigs[structs.ConsulDefaultCluster]
+	if *consul.AutoAdvertise && consul.ServerServiceName == "" {
 		return nil, fmt.Errorf("server_service_name must be set when auto_advertise is enabled")
 	}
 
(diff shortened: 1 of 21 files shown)


command/agent/config_parse.go
```diff
@@ -112,7 +107,6 @@ func ParseConfigFile(path string) (*Config, error) {
 		{"server.plan_rejection_tracker.node_window", &c.Server.PlanRejectionTracker.NodeWindow, &c.Server.PlanRejectionTracker.NodeWindowHCL, nil},
 		{"server.retry_interval", &c.Server.RetryInterval, &c.Server.RetryIntervalHCL, nil},
 		{"server.server_join.retry_interval", &c.Server.ServerJoin.RetryInterval, &c.Server.ServerJoin.RetryIntervalHCL, nil},
-		{"consul.timeout", &c.Consul.Timeout, &c.Consul.TimeoutHCL, nil},
```
lgfa29, line 115: Is this being handled somewhere else? It seems like it should be handled in https://github.com/hashicorp/nomad/pull/18997/files#diff-ae0c9f058db6b0da73449cf8e5183530091a464acb31da0690550f23c26f4f97R178 for each cluster?

command/agent/config_parse.go
```diff
@@ -181,14 +175,14 @@ func ParseConfigFile(path string) (*Config, error) {
 	//
 	// Since the map of multiple cluster configuration contains a pointer to
 	// the default block we don't need to parse it directly.
```
lgfa29, line 177:
```suggestion
```

This doesn't seem to be relevant anymore.

command/agent/config_parse.go
```diff
@@ -406,10 +400,17 @@ func parseVaults(c *Config, list *ast.ObjectList) error {
...
-			c.Vaults[v.Name] = exist.Merge(v)
-		} else {
-			c.Vaults[v.Name] = v
+
+		var vaultFound bool
+		for i, exist := range c.Vaults {
+			if exist.Name == v.Name {
+				c.Vaults[i] = exist.Merge(v)
+				vaultFound = true
+				break
```
lgfa29, line 409: Should we actually break here? This loop seems to be merging all `vault` blocks with the same name (kind of an unfortunate behaviour I feel, but we can't really change it at this point 😅)

By breaking early we're only merging once, so a config with 3+ `vault` blocks with the same name (oof...) will only contain the config from the first two.

command/agent/config_parse.go
```diff
@@ -472,10 +472,16 @@ func parseConsuls(c *Config, list *ast.ObjectList) error {
...
-		if exist, ok := c.Consuls[cc.Name]; ok {
-			c.Consuls[cc.Name] = exist.Merge(cc)
-		} else {
-			c.Consuls[cc.Name] = cc
+		var consulFound bool
+		for i, exist := range c.Consuls {
+			if exist.Name == cc.Name {
+				c.Consuls[i] = exist.Merge(cc)
+				consulFound = true
+				break
```
lgfa29, line 480: Similar to Vault above.

lgfa29 (approved):
The `consul.timeout` and early `break` may be worth a double-check, but LGTM. Thanks for handling this cleanup!

command/agent/agent.go
```diff
@@ -516,9 +523,10 @@ func convertServerConfig(agentConfig *Config) (*nomad.Config, error) {
 		conf.DefaultSchedulerConfig = *agentConfig.Server.DefaultSchedulerConfig
 	}
```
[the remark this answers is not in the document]
tgross: Oh I meant to move that. Will fix.

command/agent/agent.go
```diff
@@ -507,7 +507,14 @@ func convertServerConfig(agentConfig *Config) (*nomad.Config, error) {
...
+	// Add the Consul and Vault configs
+	conf.ConsulConfigs = helper.SliceToMap[map[string]*config.ConsulConfig](
+		agentConfig.Consuls,
+		func(cfg *config.ConsulConfig) string { return cfg.Name },
+	)
+
+	consul := conf.ConsulConfigs[structs.ConsulDefaultCluster]
+	if *consul.AutoAdvertise && consul.ServerServiceName == "" {
 		return nil, fmt.Errorf("server_service_name must be set when auto_advertise is enabled")
 	}
```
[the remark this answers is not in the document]
tgross: The non-default cluster will have these fields set because they're getting merged onto the `DefaultConsulConfig`. We don't do any validation until well after that merge, so it's possible to do this but kind of painful at the moment. I can open an issue to cover this in future refactoring though.

command/agent/config_parse.go
tgross: Yeah, exactly. This line only did anything at all (for HCL, not JSON) b/c we had a top-level `consul.timeout` from the `Config.Consul` field, which we've just removed. The work was actually then just thrown away when we parsed the individual clusters and overwrote the default cluster!

command/agent/config_parse.go
tgross: I'm pretty sure this is ok because `c.Vaults` will always start the `parseVaults` method empty. And this loop gets called for each `vault` block. So suppose we have the following silly case:

```hcl
vault { name = "foo" }
vault { name = "bar" }
vault { name = "foo" }
vault { name = "foo" }
```

The first block will be `c.Vaults[0]`. The second block will be `c.Vaults[1]`. The third block will get merged onto `c.Vaults[0]`. The fourth block (the 3rd "foo" block) will get merged onto `c.Vaults[0]`.

command/agent/agent.go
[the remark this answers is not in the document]
tgross: Opened https://github.com/hashicorp/nomad/issues/19029

PR #19011, opened 2023-11-07
cli: simplify conditionals in setup commands

pkazmierczak:
Ref: https://github.com/hashicorp/nomad/pull/18910#discussion_r1382301949


lgfa29 (approved):
LGTM! Code looks neater now 😄

PR #19012, opened 2023-11-07
e2e disable windows

shoenig:
- e2e: disable windows client
- e2e: disable windows artifact test

Note: the metrics suite also has Windows tests but it's also broken for other reasons; I'll follow up with fixing up my refactoring of that suite in https://github.com/hashicorp/nomad/compare/main...e2e-refactor-metrics and disable windows in that too for now.


tgross (approved):
LGTM

PR #19019, opened 2023-11-07
refactor: rename allocrunner's Consul service reg handler

tgross:
The allocrunner has a service registration handler that proxies various API calls to Consul. With multi-cluster support (for ENT), the service registration handler is what selects the correct Consul client. The name of this field in the allocrunner and taskrunner code base looks like it's referring to the actual Consul API client. This was actually the case before Nomad native service discovery was implemented, but now the name is misleading.

I ran into this while working on #19000.


lgfa29 (approved):
👍

PR #19024, opened 2023-11-08
Fix WorkloadIdentity.TTL handling, jobspec2 testing, and hcl1 vs 2 parsing

schmichael:
Sorry for the grab bag of little fixes, fixing `WorkloadIdentity.TTL` handling just led to a few different distractions in no particular order:

* The reason this all began: Fix the bug I introduced in https://github.com/hashicorp/nomad/pull/18943/files#diff-56b3c82fcbc857f8fb93a903f1610f6e6859b3610a4eddf92bad9ea27fdc85ec by switching WorkloadIdentity handling in job registration to use a shared function that ignored the `TTL` field.
* Missing `jobspec2` testing in CI
* Incorrect `/...` handling in our `missing` tests tool that caused us to miss `jobspec2` missing from CI (thankfully it _only_ missed jobspec2)
* Couple small jobspec (hcl1) fixes hit once `jobspec2` tests were actually run
* Inconsistent ellipses use in GNUmakefile. Obviously critical to correct. :zany_face:

### DIFF
--- a/jobspec/parse.go
+++ b/jobspec/parse.go
@@ -195,6 +195,7 @@ func parseConstraints(result *[]*api.Constraint, list *ast.ObjectList) error {
 			}
 
 			m["Operand"] = api.ConstraintDistinctHosts
+			m["RTarget"] = strconv.FormatBool(enabled)
 		}
 
 		if property, ok := m[api.ConstraintDistinctProperty]; ok {
--- a/jobspec/parse_task.go
+++ b/jobspec/parse_task.go
@@ -494,8 +494,6 @@ func parseTemplates(result *[]*api.Template, list *ast.ObjectList) error {
 			ChangeMode:    stringToPtr("restart"),
 			Splay:         timeToPtr(5 * time.Second),
 			Perms:         stringToPtr("0644"),
-			Uid:           pointer.Of(-1),
-			Gid:           pointer.Of(-1),
 			ErrMissingKey: pointer.Of(false),
 		}
 
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -178,13 +178,13 @@ check: ## Lint the source code
 	@$(MAKE) hclfmt
 	@if (git status -s | grep -q -e '\.hcl$$' -e '\.nomad$$' -e '\.tf$$'); then echo the following HCL files are out of sync; git status -s | grep -e '\.hcl$$' -e '\.nomad$$' -e '\.tf$$'; exit 1; fi
 
-	@echo "==> Check API package is isolated from rest"
+	@echo "==> Check API package is isolated from rest..."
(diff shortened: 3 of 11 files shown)


jobspec/parse.go
```diff
@@ -195,6 +195,7 @@ func parseConstraints(result *[]*api.Constraint, list *ast.ObjectList) error {
 			}
 
 			m["Operand"] = api.ConstraintDistinctHosts
+			m["RTarget"] = strconv.FormatBool(enabled)
```
angrycub, line 198: Is there a specific test requirement for this?  It is defaulted in scheduler/feasible.go (576-578) to true. Not a big deal and hard to see around it with phone review.

angrycub (approved):
Looks good to me. 😀

shoenig:
Looks like `TestParse` is unhappy 

```
=== FAIL: jobspec TestParse/basic.hcl (0.00s)
    parse_test.go:1914: Testing parse: basic.hcl
    parse_test.go:1924: 
        	Error Trace:	/home/runner/work/nomad/nomad/jobspec/parse_test.go:1924
```

jobspec/parse.go
tgross: It actually seems to break `TestParse/distinctHosts-constraint.hcl `

jobspec/parse_task.go
```diff
@@ -494,8 +494,6 @@ func parseTemplates(result *[]*api.Template, list *ast.ObjectList) error {
 			ChangeMode:    stringToPtr("restart"),
 			Splay:         timeToPtr(5 * time.Second),
 			Perms:         stringToPtr("0644"),
-			Uid:           pointer.Of(-1),
-			Gid:           pointer.Of(-1),
```
tgross, line 498: This breaks `TestParse/basic.hcl` but also looks off to me. Normally I'd expect to see an invalid value like this `-1` someplace where we _don't_ take a pointer, so that we can handle zero values that are meaningful (`0` means root in this case, which would be a bad default value). Are the downstream consumers of this value actually prepared to accept a nil pointer here?

schmichael: Great question! I did some digging, and I think this is the only place it's used (indirectly in its `structs/` form instead of this `api/` form): https://github.com/hashicorp/nomad/blob/v1.6.3/client/allocrunner/taskrunner/template/template.go#L727-L733

We actually explicitly changed code upstream from this to use `nil` instead of `-1` as part of 1.7, so I think this just got missed due to the missing `jobspec2` tests: https://github.com/hashicorp/nomad/pull/14203/files

jobspec/parse.go
schmichael: Yes, since the test explicitly sets `true` the hcl2 parser in jobspec2 explicitly sets `true`, so this change to the hcl1 parser (we really gotta drop this thing) makes it match the hcl2 behavior.

PR #19033, opened 2023-11-08
cli: fix `setup consul` binding rule config

lgfa29:
When creating the binding rule, `BindName` must match the pattern used for the role name, otherwise the task will not be able to login to Consul.

Also update the equality check for the binding rule to ensure this property is held even if the auth method already has existing binding rules attached.

### DIFF
--- a/command/asset/consul-wi-default-policy.hcl
+++ b/command/asset/consul-wi-default-policy.hcl
@@ -1,7 +1,7 @@
-service "*" {
+service_prefix "" {
   policy = "read"
 }
 
-key_prefix "*" {
+key_prefix "" {
   policy = "read"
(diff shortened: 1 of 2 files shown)


tgross (approved):
LGTM :+1:

lgfa29:
Sneaking in another fix in https://github.com/hashicorp/nomad/pull/19033/commits/1f1f0e9ad784171f5b118f0b6dc002358649a985.

I'm not sure if `"*"` would work as well, but all of Consul docs use the empty string with a `_prefix` rule to match everything.
https://developer.hashicorp.com/consul/docs/security/acl/acl-policies#matching-and-prefix-values

Issue #19030, opened 2023-11-08
Consul tokens derived from WI are never revoked


tgross:
Vault tokens derived from WI have a TTL so they'll automatically age out. But Consul tokens do not, and the `consul_hook` is not revoking them when we're done with the allocation. (The legacy workflow handles this on the server, but we won't be able to do that here.).

We should also make sure to revoke the Vault tokens even if they always expire via TTL.

tgross:
Fixed by https://github.com/hashicorp/nomad/pull/19034

PR #19034, opened 2023-11-08
revoke Consul tokens obtained via WI when alloc stops

tgross:
Add a `Postrun` hook to the allocrunner's `consul_hook` to ensure that Consul tokens we've created via WI get revoked when we're done with them. Also add the revoke to the `Prerun` hook if we've hit an error.

Fixes: https://github.com/hashicorp/nomad/issues/19030

### DIFF
--- a/client/allocrunner/consul_hook.go
+++ b/client/allocrunner/consul_hook.go
@@ -252,10 +252,26 @@ func (h *consulHook) Postrun() error {
 	return nil
 }
 
+// Destroy cleans up any remaining Consul tokens if the alloc is GC'd or fails
+// to restore after a client restart.
+func (h *consulHook) Destroy() error {
+	tokens := h.hookResources.GetConsulTokens()
+	err := h.revokeTokens(tokens)
+	if err != nil {
+		return err
+	}
+	h.hookResources.SetConsulTokens(tokens)
+	return nil
+}
+
 func (h *consulHook) revokeTokens(tokens map[string]map[string]*consulapi.ACLToken) error {
 	mErr := multierror.Error{}
 
 	for cluster, tokensForCluster := range tokens {
+		if tokensForCluster == nil {
+			// if called by Destroy, may have been removed by Postrun
+			continue
+		}
 		client, err := h.clientForCluster(cluster)
 		if err != nil {
 			mErr.Errors = append(mErr.Errors, err)


client/allocrunner/consul_hook.go
```diff
@@ -230,3 +240,38 @@ func (h *consulHook) prepareConsulClientReq(identity structs.WIHandle, authMetho
...
+// Postrun cleans up the Consul tokens after the tasks have exited.
+func (h *consulHook) Postrun() error {
+	tokens := h.hookResources.GetConsulTokens()
+	err := h.revokeTokens(tokens)
+	if err != nil {
+		return err
+	}
+	h.hookResources.SetConsulTokens(tokens)
+	return nil
+}
```
lgfa29, line 253: Does `Postrun()` runs on restore if the task exits while the Nomad agent is down? Or even worse, when the agent goes down forever 😅 

Hooks are a good best-effort, but I'm wondering if we should have a clean-up loop somewhere like we do for services. Leaking tokens that never expire could be a problem 🤔

lgfa29 (approved):

client/allocrunner/consul_hook.go
tgross: No, but we call `Destroy`. So we should be able to add a `Destroy` hook here that calls the same `revokeTokens` method to cover us. Will fix.

> Hooks are a good best-effort, but I'm wondering if we should have a clean-up loop somewhere like we do for services.

I don't disagree but it's currently impossible to do this without exposing an effectively-root token to Nomad. Because if we've failed to fire the hook we almost certainly don't have the Consul token at all, which means we can't use the logout method. This same problem exists in the legacy workflow where if the node status change the Nomad server is picking up doesn't successfully delete the token, we'll leak it. (Admittedly this is probably _less_ common.)

My take is that we want one of two things here:
* Consul should implement refreshable tokens with a TTL, the way that Vault does, or
* We should be reversing the workflow entirely with CDP, so that a hypothetical future `consul-nmd` task gets a _Nomad_ token from it's WI and it syncs with Nomad. That way we're not handling Consul tokens at all on the client (except for the agent's own very limited token).

Until we've sat down with the Consul engineering and product folks to figure out the game plan for CDP, I'm not sure I want to invest more time into fixing this leaking token problem given that we already have it today.

tgross: Pushed that `Destroy` method up and verified it removes the tokens leftover from an allocation that can't be restored after a client restart.

PR #19036, opened 2023-11-08
[ui] UI actions implementation on task index

### LINKED ISSUE #19018: [ui] Actions implementation in Task Fly-out

### LINKED ISSUE #19035: Implementation on Task page

philrenaud:
Resolves #19035

Adds an Actions dropdown on the task index page

image

Note: This formerly resolved #19018 as well, but we decided to work toward a design (persistent action service sidebar) that makes this undesirable.

### DIFF
--- a/ui/app/controllers/clients/client/index.js
+++ b/ui/app/controllers/clients/client/index.js
@@ -338,13 +338,4 @@ export default class ClientController extends Controller.extend(
     }
   }
   // #endregion metadata
-
-  // Show an additional column in the Allocations table, but only if any of them have actions.
-  get allocationsHaveActions() {
-    return this.sortedAllocations.any((a) => {
-      return a.states.any((s) => {
-        return s.get('task')?.actions.length;
-      });
-    });
-  }
 }
--- a/ui/app/services/nomad-actions.js
+++ b/ui/app/services/nomad-actions.js
@@ -1,0 +1,22 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+// Guess who just found out that "actions" is a reserved name in Ember?
+// Signed, the person who just renamed this NomadActions.
+
+// @ts-check
+import Service from '@ember/service';
+import { inject as service } from '@ember/service';
+
+export default class NomadActionsService extends Service {
+  @service can;
+
+  // Note: future Actions Governance work (https://github.com/hashicorp/nomad/issues/18800)
+  // will require this to be a computed property that depends on the current user's permissions.
+  // For now, we simply check alloc exec privileges.
+  get hasActionPermissions() {
+    return this.can.can('exec allocation');
+  }
+}
--- a/ui/app/styles/core/notifications.scss
+++ b/ui/app/styles/core/notifications.scss
@@ -7,7 +7,7 @@ section.notifications {
   position: fixed;
   bottom: 20px;
   right: 20px;
-  z-index: 100;
+  z-index: $z-notification;
   justify-items: right;
   display: grid;
 
--- a/ui/app/templates/allocations/allocation/task/index.hbs
+++ b/ui/app/templates/allocations/allocation/task/index.hbs
@@ -47,14 +47,6 @@
     </PH.Title>
     <PH.Actions>
       {{#if this.model.isRunning}}
-        <div class="two-step-button">
-          <Exec::OpenButton
-            @job={{this.model.task.taskGroup.job}}
-            @taskGroup={{this.model.task.taskGroup}}
-            @allocation={{this.model.allocation}}
-            @task={{this.model.task}}
-          />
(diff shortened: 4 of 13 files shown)


ui/app/controllers/clients/client/index.js
```diff
@@ -338,13 +338,4 @@ export default class ClientController extends Controller.extend(
     }
   }
   // #endregion metadata
-
```
philrenaud, line 341: As a side-effect here, I've decided to always show the "Actions" column on clients/:clientid pages (under alloc table).

Before, this ended up looking a bit jarring and now it'll take up a small amount of space but with the benefit of a consistent look/feel. See also the {{true}} pass in the client index handlebars file.

ui/app/templates/allocations/allocation/task/index.hbs
```diff
@@ -30,8 +30,8 @@
       </div>
     </div>
   {{/if}}
-  <h1 class="title with-flex" data-test-title>
-    <div>
+  <Hds::PageHeader class="job-page-header" as |PH|>
+    <PH.Title data-test-title>
```
philrenaud, line 34: Further added the Helios PageHeader component here as the right-aligned buttons look nicer with it.

ui/app/styles/core/notifications.scss
```diff
@@ -7,7 +7,7 @@ section.notifications {
   position: fixed;
   bottom: 20px;
   right: 20px;
-  z-index: 100;
+  z-index: $z-notification;
```
philrenaud, line 10: A fun new z-index, since we want to launch actions from the sidebar, the notifications for those actions should also show up above the sidebar.

ui/app/services/nomad-actions.js
```diff
@@ -0,0 +1,22 @@
...
+export default class NomadActionsService extends Service {
+  @service can;
+
+  // Note: future Actions Governance work (https://github.com/hashicorp/nomad/issues/18800)
+  // will require this to be a computed property that depends on the current user's permissions.
+  // For now, we simply check alloc exec privileges.
+  get hasActionPermissions() {
+    return this.can.can('exec allocation');
+  }
+}
```
philrenaud, line 22: Not much to this file yet, but expect big things soon as we use this service to persist running-in-the-background actions websockets among other things!

ui/app/controllers/clients/client/index.js
lgfa29: +1 on avoid conditional rendering as it can make the UI surprising (why do I only see this thing sometimes?)

More over though, I think I would not have an `Actions` column at all. `AllocationRow` is already quite crowded and adding more rows just makes everything more "squishy" and knowing the number of actions in an alloc doesn't seem very relevant.

Side note: not sure if this will be fixed in a follow-up PR, but the column rendering seems wrong?
https://github.com/hashicorp/nomad/blob/ff928a804590611111763632388161dc711adf88/ui/app/templates/components/allocation-row.hbs#L140-L142

ui/app/templates/allocations/allocation/task/index.hbs
```diff
@@ -44,9 +44,19 @@
...
       {{#if this.model.isRunning}}
+
+        {{#if this.shouldShowActions}}
+          <Hds::Dropdown class="actions-dropdown" as |dd|>
+            <dd.ToggleButton @color="secondary" @text="Actions" @size="medium" />
+            {{#each this.model.task.actions as |action|}}
+              <dd.Interactive {{on "click" (perform this.runAction action (get (object-at 0 action.allocations) "id"))}} @text="{{action.name}}" />
+            {{/each}}
+          </Hds::Dropdown>
+        {{/if}}
```
lgfa29, line 58: The spacing between the buttons at different levels is a bit jarring. 

I think we're also missing the `Actions` button at the alloc and group level?

image
image
image

lgfa29 (approved):

ui/app/templates/allocations/allocation/task/index.hbs
philrenaud: The spacing on the Task and Job page will be the norm for all pages; I'm rolling it the [new PageHeader component](https://helios.hashicorp.design/components/page-header) on a page-by-page basis.

I hadn't considered the spacing issue too off-putting, but this comment makes me think should do a wholesale switch-over soon.

philrenaud: Actions on Group and Alloc level, I was planning to leave until the Actions service from https://github.com/hashicorp/nomad/pull/19084 is wrapped up; I'll handle that in a future PR.

ui/app/controllers/clients/client/index.js
lgfa29: Ahh OK, I saw `actions.length` somewhere and thought this would be a number 😅 

Not sure why I didn't see the `...` button 🤔

PR #19038, opened 2023-11-08
[ui] Only run test-ui, and percy, in the event that a push/pr touches the ui directory

philrenaud:
Our percy and test-ui runs shouldn't need to be run on any non-ui-dir changes, since we fully mock our needed API endpoints etc. with mirage for our test runs.

This should save us some time and money on checks in the future.

### DIFF
--- a/.github/workflows/test-ui.yml
+++ b/.github/workflows/test-ui.yml
@@ -1,39 +1,15 @@
 name: test-ui
 on:
   pull_request:
-    paths-ignore:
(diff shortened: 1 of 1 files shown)


lgfa29 (approved):
Since these UI tests use Mirage to mock results I think this OK to have, thanks!

If we ever use a real Nomad client then it may be good to revert this because then backend code changes could break the API in ways that the UI did not expect.

PR #19040, opened 2023-11-08
client: pass alloc hook resources to template hook

lgfa29:
The task template hook uses the alloc resource to retrieve Consul tokens, so it must be passed from the allocation.


tgross (approved):
Nice work!

PR #19046, opened 2023-11-09
e2e: bump packer build instances because faster

shoenig:
`t2` is ancient hardware; bump to the latest and greatest general compute type. builds are about twice as fast or so, though "Waiting for AMI to become ready..." unfortunately isn't any faster.


tgross (approved):
LGTM

The "Waiting for AMI to become ready" is, as I understand it, a function of the disk IO when they copy the root volume. So not much we can do about that.

PR #19047, opened 2023-11-09
chore: remove comment about WI change mode

lgfa29:
Identity change mode was implemented in #18943 and handles the update at the task level, so workload identity manager receives the update as expected.


tgross (approved):
LGTM :+1:

PR #19048, opened 2023-11-09
Reframe nomad action as a wrapper around nomad job action

philrenaud:
This establishes `nomad job action` as the primary CLI command for running actions within a job, and moves `nomad action` to a wrapping/forwarding shortcut for it (for now; when future "action" contexts come about, this will determine which one to forward to)

This mimics the behaviour of `nomad status` to route its command to either `nomad job status`, `nomad node status`, etc. depending on the context passed to it, but for actions.

```
nomad job action \
  -group=group \
  -task=task \
  -job=actions-demo \
party
```

and

```
nomad action \
  -group=group \
  -task=task \
  -job=actions-demo \
party
```

should now behave the same way.

### DIFF
--- a/command/action.go
+++ b/command/action.go
@@ -4,12 +4,13 @@
 package command
 
 import (
-	"fmt"
 	"io"
 	"strings"
 
 	"github.com/posener/complete"
 )
+
+const defaultEscapeChar = "~"
 
 type ActionCommand struct {
 	Meta
@@ -34,39 +35,8 @@ Usage: nomad action [options] <action>
 
 General Options:
 
-  ` + generalOptionsUsage(usageOptsNoNamespace) + `
+  ` + generalOptionsUsage(usageOptsNoNamespace)
 
-Action Specific Options:
-
-  -job <job-id>
-    Specifies the job in which the Action is defined
-
-  -alloc <allocation-id>
-    Specifies the allocation in which the Action is defined. If not provided,
-    a group and task name must be provided and a random allocation will be
-    selected from the job.
-
-  -task <task-name>
-    Specifies the task in which the Action is defined. Required if no
-    allocation is provided.
-
-  -group=<group-name>
-    Specifies the group in which the Action is defined. Required if no
-    allocation is provided.
-
-  -i
-    Pass stdin to the container, defaults to true. Pass -i=false to disable.
-
-  -t
-    Allocate a pseudo-tty, defaults to true if stdin is detected to be a tty session.
-    Pass -t=false to disable explicitly.
-
-  -e <escape_char>
-    Sets the escape character for sessions with a pty (default: '~'). The escape
-    character is only recognized at the beginning of a line. The escape character
-    followed by a dot ('.') closes the connection. Setting the character to
-    'none' disables any escapes and makes the session fully transparent.
-`
 	return strings.TrimSpace(helpText)
 }
 
@@ -75,16 +45,7 @@ func (c *ActionCommand) Synopsis() string {
 }
 
 func (c *ActionCommand) AutocompleteFlags() complete.Flags {
-	return mergeAutocompleteFlags(c.Meta.AutocompleteFlags(FlagSetClient),
-		complete.Flags{
-			"-job":   complete.PredictAnything,
-			"-alloc": complete.PredictAnything,
-			"-task":  complete.PredictAnything,
-			"-group": complete.PredictAnything,
-			"-i":     complete.PredictNothing,
-			"-t":     complete.PredictNothing,
-			"-e":     complete.PredictAnything,
-		})
+	return nil
 }
 
 func (c *ActionCommand) AutocompleteArgs() complete.Predictor {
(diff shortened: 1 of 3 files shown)


command/action.go
```diff
@@ -128,189 +123,39 @@ func (c *ActionCommand) Run(args []string) int {
...
 	}
-
-	action := args[0]
-
-	code, err := c.execImpl(client, alloc, task, job, action, ttyOpt, escapeChar, c.Stdin, c.Stdout, c.Stderr)
-	if err != nil {
-		c.Ui.Error(fmt.Sprintf("failed to exec into task: %v", err))
-		return 1
+	if escapeChar != defaultEscapeChar {
+		commandWithFlags = append(commandWithFlags, "-e="+escapeChar)
```
philrenaud, line 148: Q: I'm only passing flags/args if they're non-default. Is this the right way to do it? I'm trying to reason about how "upstream" we want our defaults to live.

command/job_action.go
```diff
@@ -0,0 +1,324 @@
+// Copyright (c) HashiCorp, Inc.
```
philrenaud, line 1: Small modifications made to the Help text, etc., but otherwise this is the same as what previously existed in action.go

command/action.go
tgross: I think you probably want to pass the args in directly w/o messing with them. That way each downstream command implementation can deal with them on its own. That's how `nomad status` does it.

command/action.go
```diff
@@ -100,6 +93,8 @@ func (c *ActionCommand) AutocompleteArgs() complete.Predictor {
 
 func (c *ActionCommand) Name() string { return "action" }
 
+const defaultEscapeChar = "~"
```
lgfa29, line 96: Usually we define constants at the top of the file, so above `type ActionCommand struct {` in this case.

command/action.go
```diff
@@ -30,8 +23,8 @@ func (c *ActionCommand) Help() string {
 	helpText := `
 Usage: nomad action [options] <action>
 
-  Perform a predefined command inside the environment of the given allocation
-  and task.
+  Perform a predefined command inside the environment of a given context.
+  Currently this acts as a wrapper around the job action command.
 
   When ACLs are enabled, this command requires a token with the 'alloc-exec',
   'read-job', and 'list-jobs' capabilities for a task's namespace. If
```
lgfa29, line 30: I can't comment on the right line because it wasn't modified, but I think I left an `=` in `-group=<group-name>` 😅 

Moreover, should this command include all flags? The `nomad status` command only lists the ones _it_ cares about:
https://github.com/hashicorp/nomad/blob/f66eb83fc035e250569e13d392dab80e1b4cc9a4/command/status.go#L23-L40

command/action.go
```diff
@@ -100,6 +93,8 @@ func (c *ActionCommand) AutocompleteArgs() complete.Predictor {
 
 func (c *ActionCommand) Name() string { return "action" }
 
+const defaultEscapeChar = "~"
+
 func (c *ActionCommand) Run(args []string) int {
```
lgfa29, line 98: Since we're just wrapping another command for now, I don't think we need to do any logic here, something like this would be enough:

```go
func (c *ActionCommand) Run(args []string) int {
	cmd := &JobActionCommand{
		Meta:   c.Meta,
		Stdin:  c.Stdin,
		Stdout: c.Stdout,
		Stderr: c.Stderr,
	}

	return cmd.Run(args)
}
```

This also means we can get rid of all flags and have the`AutocompleteFlags` logic.

```go
func (c *ActionCommand) AutocompleteFlags() complete.Flags {
	return nil
}
```

command/job_action.go
```diff
@@ -0,0 +1,324 @@
...
+  -alloc <allocation-id>
+    Specifies the allocation in which the Action is defined. If not provided,
+    a group and task name must be provided and a random allocation will be
+    selected from the job.
+
+  -task <task-name>
+    Specifies the task in which the Action is defined. Required if no
+    allocation is provided.
+
+  -group=<group-name>
```
lgfa29, line 68:
```suggestion
  -group <group-name>
```

command/action.go
```diff
@@ -30,8 +23,8 @@ func (c *ActionCommand) Help() string {
 	helpText := `
 Usage: nomad action [options] <action>
 
-  Perform a predefined command inside the environment of the given allocation
-  and task.
+  Perform a predefined command inside the environment of a given context.
+  Currently this acts as a wrapper around the job action command.
 
   When ACLs are enabled, this command requires a token with the 'alloc-exec',
```
lgfa29, line 29: This paragraph is also specific to `nomad job action`. We can probably remove it.

lgfa29 (changes requested):
Looking good! But I think we don't need this much handling in the top-level `nomad action` command, just proxying the args directly to `nomad job action` should do the trick (but let me know if it doesn't 😬)

command/action.go
philrenaud: Yeah good call, keeping the flags on `job action` and ditching them on `action`.

command/action.go
philrenaud: Oh neat! I didn't expect this would pass the flags (-group -task -t etc.) along automatically.

lgfa29 (approved):

PR #19051, opened 2023-11-09
[ui] Shows the client/node name alongside alloc short ID if the job is system/sysbatch

### LINKED ISSUE #19050: [ui] In the Job Status panel, tooltips for allocations in a system job should show node name
Since a system/sysbatch job has a single alloc per node/client, we should show that node name when hovering an alloc, as it may be more useful than the alloc ID alone.

image

philrenaud:
On hover, shows client info on the job status panel's allocs:
image

Resolves #19050


juliezzhou (approved):
lgtm!

PR #19054, opened 2023-11-09
e2e: refactor nomadexec test suite

shoenig:
This was failing in the recent e2e run; figured may as well migrate it to the modern way of writing an e2e test. Seems to work fine, nothing about the actual test case is changed.


jrasell (approved):
LGTM!

PR #19055, opened 2023-11-09
E2E: add template block to `consulcompat` test

tgross:
The Consul compatibility test focuses on Connect, but it'd be a good idea to ensure we can successfully get template data out of Consul as well.

(`client` package test is failing because of https://github.com/hashicorp/nomad/pull/19056, will rebase once that's been merged)

### DIFF
--- a/e2e/consulcompat/input/consul-policy-for-nomad.hcl
+++ b/e2e/consulcompat/input/consul-policy-for-nomad.hcl
@@ -14,15 +14,11 @@ agent_prefix "" {
   policy = "read"
 }
 
-key_prefix "" {
-  policy = "read"
-}
-
 node_prefix "" {
   policy = "read"
 }
 
-service_prefix "" {
+service_prefix "nomad" {
   policy = "write"
 }
 
@@ -36,7 +32,7 @@ namespace_prefix "" {
     policy = "read"
   }
 
-  service_prefix "" {
+  service_prefix "nomad" {
     policy = "write"
   }
 }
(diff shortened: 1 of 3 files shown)


e2e/consulcompat/input/consul-policy-for-nomad.hcl
```diff
@@ -14,15 +14,11 @@ agent_prefix "" {
...
-key_prefix "" {
-  policy = "read"
-}
-
 node_prefix "" {
   policy = "read"
 }
 
-service_prefix "" {
+service_prefix "nomad" {
```
lgfa29, line 21: Since the Nomad agent runs in `dev` mode, this policy is used for both client and  servers, so this will prevent clients from registering in Consul and also from syncing services.
https://github.com/hashicorp/nomad/blob/5ad715b2818041674645486862415848dc425f48/command/agent/consul/service_client.go#L984-L988

If we don't care about gateways and their config entries, I think the most restrictive policy we can have that is fully functionable would be:

```hcl
service_prefix "nomad" {
  policy = "write"
}

service_prefix "nomad-client" {
  policy = "write"
}

service_prefix "" {
  policy = "read"
}
```

lgfa29 (approved):

e2e/consulcompat/input/consul-policy-for-nomad.hcl
tgross: `service_prefix` is a prefix match, so I think we can collapse the top two

PR #19057, opened 2023-11-09
e2e: more cpu and memory for java tasks and some scripts

shoenig:
The memory limits on the java tasks in particular were probably too low; bump these to be more sensible.


jrasell (approved):
LGTM!

PR #19058, opened 2023-11-10
client: fix Consul token retrievel for templates

lgfa29:
The template hook must use the Consul token for the cluster defined in the task-level `consul` block or, if `nil, in the group-level `consul` block.

The Consul tokens are generated by the allocrunner consul hook, but during the transition period we must fallback to the Nomad agent token if workload identities are not being used.

So an empty token returned from `GetConsulTokens()` is not enough to determine if we should use the legacy flow (either this is an old task or the cluster is not configured for Consul WI), or if there is a misconfiguration (task or group is `consul` block is using a cluster that doesn't have an `identity` set).

In order to distinguish between the two scenarios we must iterate over the task identities looking for one suitable for the Consul cluster being used.

### DIFF
--- a/client/allocrunner/taskrunner/template_hook.go
+++ b/client/allocrunner/taskrunner/template_hook.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"sync"
 
-	consulapi "github.com/hashicorp/consul/api"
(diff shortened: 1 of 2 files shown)


tgross (approved):
LGTM! Thanks @lgfa29!

PR #19061, opened 2023-11-10
[ui] Bugfix: expecting job.name and job.namespace passed as separate params, rather than job.id

philrenaud:
Small fix to let UI actions passing use job.name instead of job.id, since namespace is passed as an explicit param afterward


tgross (approved):
LGTM!

PR #19067, opened 2023-11-10
e2ev3: wait for logs to become ready

shoenig:
Just because an alloc is running does not mean nomad is ready to serve
task logs. In a test case where you immediatly read logs after starting
a task, it could be that nomad responds with "no logs found" when you
try to read logs, in which case you just need to wait longer. Do so in
the v3 TaskLogs helper function.


tgross (approved):
Much nicer :+1:

PR #19077, opened 2023-11-13
Considerations for Stateful Workloads

cjobermaier:
This document will help Nomad practitioners understand stateful storage options.


website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,116 @@
+---
+layout: docs
+page_title: Considerations for Stateful Workloads
+description: |-
+  Learn about the available options for persistent workload storage on Nomad
+---
+
+# Considerations for Stateful Workloads
+
+By default, HashiCorp Nomad's allocation storage is ephemeral and can be discarded at any moment (such as a new deployment, rescheduling, or losing a Nomad client). This is undesirable when running persistent workloads such as databases.
```
tgross, line 10: All the docs in the Nomad repo get hard-wrapped lines (this makes reviews easier on large paragraphs).

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,116 @@
...
+| Ephemeral disks  | <ul><li>Fast due to being local</li><li>Basic best effort persistence, including optional migration across Nomad clients</li></ul>  | <ul><li> Not fault tolerant. In case of hardware failure on a single instance, the data will be lost </li></ul>  | <ul><li>Environments that need temporary caches, somewhere to store files undergoing processing, etc. Everything which is ephemeral and can be easily rebuilt.</li></ul>  |
+
+## Next steps
+
+In this document, you learned about the different persistent storage options for Nomad.
```
tgross, line 103: This kind of language is usually reserved for Tutorials. This repo has reference and concept docs, and shouldn't include this.

website/data/docs-nav-data.json
```diff
@@ -2247,6 +2247,10 @@
       {
         "title": "Key Management",
         "path": "operations/key-management"
+      },
+      {
+        "title": "Considerations for Stateful Workloads",
+        "path": "operations/stateful-workloads"
```
tgross, line 2253: Two minor thoughts on this:
* The title pushes the sidebar navigation out quite a bit. Can it be shorter?
* Does this really belong here in `operations/` rather than `concepts/` or maybe even `integrations/`?

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+In an AWS environment, an EBS block volume is zonal and slower than the local storage on the EC2 instances, but it is readily available. Your data will persist in the case of a crash of the underlying hardware, and you can reattach the volume to a different EC2 instance within the same AZ. However, in the case of a zonal failure, the data on it is "stuck" within that zone, and the application is not redundant across Availability Zones. You must provision volumes at a fixed size, which you need to monitor and manage. 
+
+AWS EFS is a regional service that provides file storage using the NFS protocol. Since the data replicates across different Availability Zones and uses a higher level protocol (file instead of block like EBS), EFS has higher latencies and lower throughput than EBS. However, it can continue to serve data requests in case of a zonal failure. A significant advantage, and sometimes a disadvantage, for NFS and similar protocols is that you can mount and access them simultaneously on multiple machines.
```
tgross, line 34: Do we really want to talk about a specific cloud vendor's services here, as opposed to the general type of volumes with these as an example? Ex. should we talk about the difference between local block vs network filesystems, rather than EBS vs EFS?

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+While Nomad follows the CSI specification, some plugins may implement orchestrator-specific logic that makes them incompatible with Nomad. You should validate that your chosen plugin works with Nomad before using it. Refer to the plugin documentation for more information.
+
+There are three CSI plugin subtypes:
+
+- Controller, which communicates with the storage provider for the management of the volume lifecycle
+- Node, which runs on all Nomad clients and handles all local operations (for example, mounting/unmounting volumes in allocations) and have to be `privileged` to be able to perform those operations
+- Monolithic, which combines both the above roles.
+
+All types can and should be run as Nomad jobs - `system` jobs for Node and Monolithic, `service` for Controllers.
```
tgross, line 60: This is heavily covered in https://nomad-kfsnumtbi-hashicorp.vercel.app/nomad/docs/concepts/plugins/csi so we should at the very least link to it from the relevant sections.

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+### Host volumes
+
+Host volumes mount paths from the host (the Nomad client) into allocations. Nomad is aware of and takes into account their availability for scheduling but is indifferent about their underlying characteristics, such as if it is a standard folder on a local ext4 filesystem, backed by a distributed networked storage such as GlusterFS, or a mounted NFS/CIFS volume from a NAS or a public cloud service such as AWS EFS. Currently, you need to declare Host volumes in the agent configuration file and since the configuration is static until the Nomad client is restarted, host volumes are impractical if frequent storage changes are a requirement. 
+
+As administrators usually handle storage, operating system (OS), and job configurations separately, it is important to ensure that appropriate access controls are in place to grant permission to modify and update these configurations. If there is a complete separation of duties for storage, OS, and job management, good communication between the different teams is important to mitigate potential issues around storage performance, size limitations, and inappropriate mounting options.
```
tgross, line 70: We're hinting at but not directly addressing the [personas in the security model](https://developer.hashicorp.com/nomad/docs/concepts/security#personas). We should definitely use the exact same names here and probably cross link.

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+Host volumes backed by local storage help persist data that is not highly critical, such as on-disk cache that can be rebuilt if needed but preferably should persist. When backed by networked storage such as NFS/CIFS-mounted volumes or distributed storage via GlusterFS/Ceph, they provide a quick option to consume highly available and reliable storage.
+
+Refer to the [Stateful workloads with Nomad host volumes](/nomad/tutorials/stateful-workloads/stateful-workloads-host-volumes) tutorial to learn more about using host volumes with Nomad.
+
+#### NFS caveats
+
+A few caveats to note if Host volumes are backed by NFS volumes include ACLs, reliability, and performance. NFS mount options should be the same on all mounting Nomad clients.
```
tgross, line 78: This first sentence seems like a fragment?

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+A significant factor in the performance of NFS-backed storage is the `wsize` and `rsize` mount options that determine the maximum read/write size of a block. Smaller sizes mean bigger operations will be split into smaller chunks, significantly impacting performance. The underlying storage system's vendor provides the optimal sizes. For example, [AWS EFS](https://docs.aws.amazon.com/efs/latest/ug/mounting-fs-mount-cmd-general.html) recommends a value of `1048576` bytes of data for both `wsize` and `rsize`.
```
tgross, line 80: None of this seems specific to Nomad. Should we include it at all?

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,116 @@
...
+| Host volumes backed by networked storage  | <ul><li>Readily available</li><li>Require no ongoing maintenance</li></ul>  | <ul><li>Require coordination between multiple personas to configure and consume (storage admins need to provision volumes, operators running the Nomad clients need to configure them statically in the Nomad client's configuration file)</li><li>The underlying networked storage and its limitations are decoupled from the consumer, but need to be understood. For example,  is concurrent access possible</li></ul>  | <ul><li>Environments with low amounts or low frequency of change of storage that have an existing storage provider that can be consumed via NFS/CIFS.</li></ul>  |
+| Ephemeral disks  | <ul><li>Fast due to being local</li><li>Basic best effort persistence, including optional migration across Nomad clients</li></ul>  | <ul><li> Not fault tolerant. In case of hardware failure on a single instance, the data will be lost </li></ul>  | <ul><li>Environments that need temporary caches, somewhere to store files undergoing processing, etc. Everything which is ephemeral and can be easily rebuilt.</li></ul>  |
```
tgross, line 99: The "doesn't require ongoing maintenance" feels really misleading here. You certainly need to maintain the disks on hosts! Or NFS! I would leave that out entirely, or turn the comments about maintenance into something more specific.

tgross: > Environments with low amounts of ideally persistent storage requirements that could tolerate some failure, but prefer not to and/or have high performance / low latency needs.

This is a bit of a run-on sentence.

website/content/docs/operations/stateful-workloads.mdx
cjobermaier: Yup I'll add it.

tgross (changes requested):
@cjobermaier I'm just noticing that this content also appears (almost verbatim) in https://github.com/hashicorp/tutorials/pull/1766 (internal repo link). Normally we don't repeat content but should be linking to it. Can you give some context here?

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+You can find a list of plugins and their feature set in the [Kubernetes CSI Developer Documentation](https://kubernetes-csi.github.io/docs/drivers.html)
+
+While Nomad follows the CSI specification, some plugins may implement orchestrator-specific logic that makes them incompatible with Nomad. You should validate that your chosen plugin works with Nomad before using it. Refer to the plugin documentation for more information.
+
+There are three CSI plugin subtypes:
+
+- Controller, which communicates with the storage provider for the management of the volume lifecycle
+- Node, which runs on all Nomad clients and handles all local operations (for example, mounting/unmounting volumes in allocations) and have to be `privileged` to be able to perform those operations
+- Monolithic, which combines both the above roles.
```
BrianMMcClain, line 58:
```suggestion
- **Controller**: Communicates with the storage provider for the management of the volume lifecycle
- **Node**: Runs on all Nomad clients and handles all local operations (for example, mounting/unmounting volumes in allocations). Must be `privileged` to be able to perform those operations
- **Monolithic**: Combines both the above roles
```

BrianMMcClain (dismissed):

cjobermaier:
> @cjobermaier I'm just noticing that this content also appears (almost verbatim) in [hashicorp/tutorials#1766](https://github.com/hashicorp/tutorials/pull/1766) (internal repo link). Normally we don't repeat content but should be linking to it. Can you give some context here?

I will close that. Some backstory here: content like this would traditionally land in HashiCorp's [Well-Architected Framework](https://developer.hashicorp.com/well-architected-framework). We are shifting where content lives, and how we reference content in WAF. That PR was first opened when this document would live in WAF. 

As product best practice docs will now live in the product docs, I will close that PR and we will only use this one.

website/content/docs/operations/stateful-workloads.mdx
sofixa: @tgross my thinking here was to give concrete examples with the cloud providers' services names, with the respective caveats (e.g. EBS being zonal thus being unsuitable for a lot of  persistent storage needs), to make it clear which cloud service correspond to which storage option. However I see the whole paragraph has gotten a little mixed and I'll reword it to make clear which services are on which cloud provider and what they do.

website/content/docs/operations/stateful-workloads.mdx
sofixa: It's not relevant to Nomad _per se_, but can be an easy stumbling point for people starting with storage specifically for Nomad. If it's too niche, maybe instead of adding it directly we can link to third party sites that describe the same thing?

website/content/docs/operations/stateful-workloads.mdx
sofixa: @tgross regarding the first point, I was thinking purely of maintenance as the one you need on the CSI plugins (needing to update them, monitor their health, etc.) but you're right, there's all sorts of other maintenance that still needs to happen :D I'll try to reword it using the Security Model personas, that there is little to no maintenance for Nomad Administrators but there is for System Administrators.

website/content/docs/operations/stateful-workloads.mdx
judithpatudith, line 70:
```suggestion
As administrators usually handle storage, operating system (OS), and job configurations separately, make sure that appropriate access controls grant permission to modify and update these configurations. If there is a complete separation of duties for storage, OS, and job management, the different teams must communicate well to mitigate potential issues around storage performance, size limitations, and inappropriate mounting options.
```

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+As administrators usually handle storage, operating system (OS), and job configurations separately, it is important to ensure that appropriate access controls are in place to grant permission to modify and update these configurations. If there is a complete separation of duties for storage, OS, and job management, good communication between the different teams is important to mitigate potential issues around storage performance, size limitations, and inappropriate mounting options.
+
+Host volumes backed by local storage help persist data that is not highly critical, such as on-disk cache that can be rebuilt if needed but preferably should persist. When backed by networked storage such as NFS/CIFS-mounted volumes or distributed storage via GlusterFS/Ceph, they provide a quick option to consume highly available and reliable storage.
```
judithpatudith, line 72:
```suggestion
Host volumes backed by local storage help persist data that is not critical, for example an on-disk cache that you can rebuild you need to. When backed by networked storage such as NFS/CIFS-mounted volumes or distributed storage via GlusterFS/Ceph, host volumes provide a quick option to consume highly available and reliable storage.
```

website/content/docs/operations/stateful-workloads.mdx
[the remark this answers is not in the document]
judithpatudith: Something still seems wrong with this sentence. Should it be `volumes that include`?

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+To learn more about hard mounts and soft mounts, visit Red Hat's [NFS documentation](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/5/html/deployment_guide/s1-nfs-client-config-options).
+
+### Ephemeral disks
+
+Nomad has the concept of [ephemeral disks](/nomad/docs/job-specification/ephemeral_disk), which is a way to describe the best effort persistence of a Nomad allocation's folder. It supports data migrations between hosts (which requires network connectivity between the Nomad client nodes) and is size-aware for scheduling. Since persistence is the best effort, however,  you will lose data in the case of a client or underlying storage failure. Ephemeral disks are perfectly suited for data that can be rebuilt if needed, such as an in-progress cache or a local copy of data.
```
judithpatudith, line 88:
```suggestion
Nomad [ephemeral disks](/nomad/docs/job-specification/ephemeral_disk), describe the best-effort persistence of a Nomad allocation's folder. They support data migrations between hosts (which require network connectivity between the Nomad client nodes) and are size-aware for scheduling. Since persistence is the best effort, however,  you will lose data if the client or underlying storage fails. Ephemeral disks are perfect for data that you can rebuild if needed, such as an in-progress cache or a local copy of data.
```

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+Depending on your NFS version, the UID/GID (user/group IDs) can differ between the different Nomad clients, leading to issues when an allocation on another host tries to access the volume. The only way to ensure this isn't an issue is to use NFS v4 with ID mapping based on Kerberos or to have a reliable configuration management/image-building process that ensures UID/GIDs synchronize between hosts. You should use hard mounts to prevent data loss, optionally with `intr` to enable the option to interrupt NFS requests, which prevents the whole system from locking up in case of NFS server unavailability. 
+
+To learn more about hard mounts and soft mounts, visit Red Hat's [NFS documentation](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/5/html/deployment_guide/s1-nfs-client-config-options).
+
+### Ephemeral disks
```
judithpatudith, line 86: These are under the H2 heading of `## Persistent storage options` but they don't seem persistent, they seem ephemeral. Should we list them here or give them their own section? We might also want to put host volumes in this new section, sine it seems like we're saying that ephemeral disks and host volumes are bad options to store data that you need to keep?

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+Nomad has the concept of [ephemeral disks](/nomad/docs/job-specification/ephemeral_disk), which is a way to describe the best effort persistence of a Nomad allocation's folder. It supports data migrations between hosts (which requires network connectivity between the Nomad client nodes) and is size-aware for scheduling. Since persistence is the best effort, however,  you will lose data in the case of a client or underlying storage failure. Ephemeral disks are perfectly suited for data that can be rebuilt if needed, such as an in-progress cache or a local copy of data.
+
+## Storage comparison
+
+With the information laid out in this document, you can use the following table to make an informed decision on the storage options that best addresses your Nomad storage requirements.
```
judithpatudith, line 92:
```suggestion
With the information laid out in this document, use the following table to choose the storage options that best addresses your Nomad storage requirements.
```

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+| Storage option  | Advantages  | Disadvantages  | Ideal for  |
+|---|---|---|---|
+| CSI volumes  | <ul><li>Wide ecosystem with many providers</li><li>Advanced features such as snapshots, cloning, and resizing</li><li>High level of dynamism and self-service (volumes can be created on-demand by anyone with the correct ACL policies) and flexibility</li></ul>    | <ul><li>Some complexity and ongoing maintenance</li><li>Upgrades of the CSI plugins have to follow the underlying storage provider's API changes/upgrades</li><li>Not all CSI plugins implement all features</li><li>Not all CSI plugins respect the CSI spec and are Nomad compatible</li><li>Node plugins need to run in privileged mode to be able to mount the volumes in allocations</li></ul>  |  <ul><li>Environments with dynamic storage requirements where Nomad cluster operators and consumers should be able to easily add/change storage, and where the storage provider of choice has a CSI plugin that respects the CSI spec.</li></ul> |
```
judithpatudith, line 96:
```suggestion
| CSI volumes  | <ul><li>Wide ecosystem with many providers</li><li>Advanced features such as snapshots, cloning, and resizing</li><li>Dynamic, flexible, and self-service (anyone with the correct ACL policies can create volumes on-demand)</li></ul>    | <ul><li>Some complexity and ongoing maintenance</li><li>Plugin upgrades have to follow the underlying storage provider's API changes/upgrades</li><li>Not all CSI plugins implement all features</li><li>Not all CSI plugins respect the CSI spec and are Nomad compatible</li><li>Node plugins need to run in privileged mode to be able to mount the volumes in allocations</li></ul>  |  <ul><li>Environments where Nomad cluster operators and consumers need to easily add/change storage, and where the storage provider of choice has a CSI plugin that respects the CSI spec</li></ul> |
```

website/content/docs/operations/stateful-workloads.mdx
judithpatudith: One thing it might be nice to do is sort the links into categories since there are a lot of them. Maybe based on learning mode (tutorials vs docs) or maybe we should give CSI it's own little links section?

website/content/docs/operations/stateful-workloads.mdx
sofixa: I tried doing that, and after some back and forth decided to go with a table format for better readability. What do you think about the following for AWS (wanted some feedback before doing it for GCP and Azure too):
```
### AWS 

| AWS service                                                                                  | Availability                                                      | Persistence                                                                      | Performance                                                                                                                                        | Suitability                                                                                                                  | Use in Nomad                               |
|----------------------------------------------------------------------------------------------|-------------------------------------------------------------------|----------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------|
| [Instance Storage](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html) | Locally on some instance types                                    | Limited, not persistent accross instance stops/terminations or hardware failures | High throughput and low latency                                                                                                                    | Temporary storage of information that changes frequently, such as buffers, caches, scratch data, and other temporary content | via `ephemeral_disk` or `host_volume`      |
| [Elastic Block Store](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AmazonEBS.html)    | Zonal block devices attached to instances                         | Strongly persistent, with an independent lifecycle                               | [Configurable](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSPerformance.html), but higher latency than Instance Store                    | General purpose persistent storage                                                                                           | via the CSI plugin or static `host_volume` |
| [Elastic File System](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AmazonEFS.html)    | Regional file storage that can be available to multiple instances | Strongly persistent, with an independent lifecycle                               | [Configurable](https://docs.aws.amazon.com/efs/latest/ug/performance.html), but with less throughput and higher latency than Instance Store or EBS | File storage that needs to be available to multiple instances in multiple zones (even only as a failover)                    | via the CSI plugin or static `host_volume` |


```

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+The following are options for persistent storage. Since environments differ depending on application requirements, it is important to choose the appropriate storage for your environment. The most important things to consider are performance, reliability, availability, and maintenance.
+
+### CSI
+
+[Container Storage Interface](https://github.com/container-storage-interface/spec) is a vendor-neutral specification that allows storage providers to develop plugins that orchestrators such as Nomad can use. CSI plugins can support various features, such as dynamic provisioning of volumes and their lifecycle management, such as snapshots, deletion, and dynamic resizing. The exact feature set will depend on the plugin and the underlying storage platform.  
+
+You can find a list of plugins and their feature set in the [Kubernetes CSI Developer Documentation](https://kubernetes-csi.github.io/docs/drivers.html)
+
+While Nomad follows the CSI specification, some plugins may implement orchestrator-specific logic that makes them incompatible with Nomad. You should validate that your chosen plugin works with Nomad before using it. Refer to the plugin documentation for more information.
```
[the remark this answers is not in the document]
sofixa: the plugin documentation from the storage provider, I'll clarify that

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+There are three CSI plugin subtypes:
+
+- Controller, which communicates with the storage provider for the management of the volume lifecycle
+- Node, which runs on all Nomad clients and handles all local operations (for example, mounting/unmounting volumes in allocations) and have to be `privileged` to be able to perform those operations
+- Monolithic, which combines both the above roles.
+
+All types can and should be run as Nomad jobs - `system` jobs for Node and Monolithic, `service` for Controllers.
+
+CSI plugins are useful when storage requirements are quickly and constantly evolving due to their dynamic nature, but they present some challenges in terms of maintenance. Most notably, they need to run constantly, be configured (including authentication and connectivity to the storage platform), and updated to keep track with new features and bug fixes and keep compatibility with the underlying storage platform.
```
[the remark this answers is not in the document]
sofixa: Clarified that part, it's purely if there's frequent changes in the amount of persistent volumes  needed - e.g. if there's a new app needing a persistent volume every week, that's a lot of changes and CSI makes more sense.

website/content/docs/operations/stateful-workloads.mdx
sofixa: They're _somewhat_ persistent and a decent option for some use cases.

website/content/docs/operations/stateful-workloads.mdx
tgross: Unresolving because the hard-wrap wasn't fixed.

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,246 @@
...
+in Nomad. The information provided is for practitioners familiar with Nomad and
+with a foundational understanding of storage basics. 
+
+At the end of this article, we provide a [storage
+comparison](#storage-comparison) table with the advantages, disadvantages, and
+the ideal use-case for each storage option. This guidance will help you
+understand and choose the best storage option for your Nomad allocation storage.
+
+This document addresses traditional storage you can attach as a filesystem and
+does not cover object storage such as AWS S3 or MinIO.
```
tgross, line 23: These are reference docs, not articles. We don't need to say things like "at the end of this article" and there's an auto-generated table of contents.

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,246 @@
...
+This document addresses traditional storage you can attach as a filesystem and
+does not cover object storage such as AWS S3 or MinIO. 
+
+## Considerations
+
+Consider access patterns, performance, reliability and availability needs, and
+maintenance to choose the most appropriate storage strategy. 
+
+Local storage is performant and available. If it has enough capacity it does not
+need much maintenance. But, it is not redundant; if a single node, disk, or
```
tgross, line 31:
```suggestion
need much maintenance. But it is not redundant; if a single node, disk, or
```

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,246 @@
...
+group of disks fails, data loss and service interruption will occur.
+
+A geographically distributed networked storage with multiple redundancies,
+including disks, controllers, and network paths, provides higher availability
+and resilience, and can tolerate multiple hardware failures before risking data
+loss. But, it relies on network performance, can have higher latency, be
+lower-throughput, and need more maintenance.
```
tgross, line 38:
```suggestion
loss. But the performance and reliability of networked storage depends 
on the network. It can have higher latency and lower throughput than local
storage, and may require more maintenance.
```


More comma abuse. The phrase "relies on network performance" is a little vague.

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,246 @@
...
+choice will depend your organizational and application needs.
+
+### Public cloud
+
+Public cloud providers offer different storage services with various tradeoffs.
+Usually they're comprised of local disks (physically attached and very
+performant, but not very persistent), network attached block devices (performant
+and very persistent, but usually only attachable to one instance at a time), and
+networked shared storage (less performant, but can be accessed from multiple
+instances).
```
tgross, line 51: "very" is doing a lot of handwaving in this paragraph. I'd get rid of all the parentheticals and let the tables speak for themselves.

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,246 @@
...
+| [Instance Storage](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html) | Locally on some instance types                                    | Limited, not persistent across instance stops/terminations or hardware failures | High throughput and low latency                                                                                                                    | Temporary storage of information that changes frequently, such as buffers, caches, scratch data, and other temporary content |
+| [Elastic Block Store](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AmazonEBS.html)    | Zonal block devices attached to one or more instances                         | Strongly persistent, with an independent lifecycle                               | [Configurable](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSPerformance.html), but higher latency than Instance Store                    | General purpose persistent storage                                                                                           |
```
tgross, line 58: For this entirely table I'd replace "strongly persistent" with simply "persistent" everywhere. We're not making specific performance claims, and it doesn't add anything.

"Independent lifecycle" seems useful to explain either here in the table or above -- the fact that it's lifecycle is independent _of the VM it's attached to_ is the meaningful bit.

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,246 @@
...
+validate that your chosen plugin works with Nomad before using it. Refer to the
+plugin documentation from the storage provider for more information.
+
+There are three CSI plugin subtypes:
+
+- **Controller**: Communicates with the storage provider to manage the volume
+	lifecycle. - **Node**: Runs on all Nomad clients and handles all local
+	operations (for example, mounting/unmounting volumes in allocations). The
+	node must be `privileged` to perform those operations. - **Monolithic**:
+	Combines both the above roles.
```
tgross, line 116: Broken bullets

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,246 @@
...
+
+All types can and should be run as Nomad jobs - `system` jobs for Node and
+Monolithic, `service` for Controllers. More information can be found on the CSI
+concepts [documentation page](/nomad/docs/concepts/plugins/csi).
+
+CSI plugins are useful when storage requirements are quickly and constantly
+evolving due to their dynamic nature - for instance, new workloads that
+frequently need persistent storage added or removed. However, they present some
+challenges in terms of maintenance - most notably, they need to run constantly,
```
tgross, line 125: Let's convert these `-` to em-dashes or get rid of them entirely and break up these run-on sentences.

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,246 @@
...
+backed by a distributed networked storage such as GlusterFS, or a mounted
+NFS/CIFS volume from a NAS or a public cloud service such as AWS EFS. Therefore
+you can use host volumes for both local somewhat persistent storage and for
+highly persistent networked storage.
+
+Because you need to declare host volumes in the Nomad agent's configuration
+file, you must restart the Nomad client to reconfigure them. This makes host
+volumes impractical if you frequently change your storage configuration.
+Furthermore, it might require coordination between different
+[personnas](/nomad/docs/concepts/security#personas) to configure and consume
```
tgross, line 156:
```suggestion
[personas](/nomad/docs/concepts/security#personas) to configure and consume
```

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,246 @@
...
+and reliable storage.
+
+Refer to the [Stateful workloads with Nomad host
+volumes](/nomad/tutorials/stateful-workloads/stateful-workloads-host-volumes)
+tutorial to learn more about using host volumes with Nomad.
+
+#### NFS caveats
+
+A few caveats with NFS-backed host volumes include ACLs, reliability, and
+performance. NFS mount options should be the same on all mounting Nomad clients.
```
tgross, line 176: This doesn't seem to be well-described here, but probably goes better with the paragraph about UIDs. I'd consider moving this down or moving the block size discussion after the UID discussion.

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,246 @@
...
+## Additional resources
+
+In this document, you learned about the different persistent storage options for
+Nomad.
+
+To learn more about Nomad and the topics covered in this article, visit the
+following resources:
+### Allocations
+- Learn how to [manage your allocations and its
+  storage](/nomad/tutorials/manage-clusters)
```
tgross, line 229: Plural/singular disagreement, and I think the wrong link

website/content/docs/operations/stateful-workloads.mdx
```diff
@@ -0,0 +1,246 @@
...
+| Ephemeral disks  | <ul><li>Fast due to being local</li><li>Basic best effort persistence, including optional migration across Nomad clients</li></ul>  | <ul><li> Not fault tolerant. In case of hardware failure on a single instance, the data will be lost </li></ul>  | <ul><li>Environments that need temporary caches, somewhere to store files undergoing processing, etc. Everything which is ephemeral and can be easily rebuilt.</li></ul>  |
+
+## Additional resources
+
+In this document, you learned about the different persistent storage options for
+Nomad.
+
```
tgross, line 224: This isn't a blog post, we don't have to announce that it's done :grin: 

```suggestion
```

tgross (changes requested):

sofixa:
@tgross @cjobermaier I've addressed the latest feedback in #19692 19692

sofixa:
Thanks @tgross ! Which ones? IIRC all open comments are shown as "outdated" (so in theory addressed) outside of the last one which GH doesn't pick up because, following your recommendation, I inversed the order of the UID and the other NFS caveats, so the text is the same, just at a slightly different place.

tgross:
> all open comments are shown as "outdated" (so in theory addressed)

GitHub marks anything as "outdated" if it changed, which doesn't necessarily mean it was actually addressed, only that there was a new diff that touched that line (ex. a spelling fix). The authors of the PR should be verifying that each change requested was actually fixed and then marking the comment as resolved. If y'all can't do that, then I'll just take a pass through this and get it all finished up, because this has been lingering for an awful long time at this point.

sofixa:
@tgross since I'm not the PR author I can't mark them as resolved, so I 👍 -ed the ones I think I've addressed.

tgross (approved):
I've pushed one more commit mopping up some wordiness and cleaning up some textual issues like extra spaces, trailing spaces, missing EOL, tabs instead of spaces, etc. Will merge this once CI is done.

PR #19079, opened 2023-11-13
cli: add resources option to allocations

robinovitch61:
There is a documented [resources](https://developer.hashicorp.com/nomad/api-docs/allocations\#resources) parameter for the http api, but this is impossible to set as "true" right now in the CLI as far as I can tell.

I've exposed this option and enhanced the simple test for the other params.


lgfa29 (changes requested):
Hi @robinovitch61 👋 

Thanks for catching this! We certainly needs to support this, but I think the `QueryOptions` may not be the best place to put it since it's a generic struct used in most read requests, and `Resources`, as far as I can tell, is only used in  [alloc list](https://developer.hashicorp.com/nomad/api-docs/allocations#resources) and [node list](https://developer.hashicorp.com/nomad/api-docs/nodes#resources) (somewhere else to add if you feel like contributing more code 😄).

We also try to keep this struct in sync with its counterpart [in the `structs` package](https://github.com/hashicorp/nomad/blob/main/nomad/structs/structs.go#L267-L320).

Which brings us to the dilemma of Nomad's `api` package 😬 

We would need to modify the signature of `Allocations().List()` to receive a new parameter for resources, but that would be a breaking change to a public package, which we always try to avoid.

So, I think the best way forward here would be to create a new struct for options spcific to this endpoint and a new method that uses it. Something like this:

```go
type AllocationListOptions struct {
  Resources bool
}
// ...
func (a *Allocations) List(q *QueryOptions) ([]*AllocationListStub, *QueryMeta, error) {
  return a.ListOpts(nil, q)
}

func (a *Allocations) ListOpts(opts *AllocationListOptions, q *QueryOptions) ([]*AllocationListStub, *QueryMeta, error) {
  // Handle opts.Resources and the rest of the list logic
}
// ...
```

This way `Allocations().List()` is kept the same and consumers can opt-in to the `ListOpts` method if they need to. This new method also allows us to add new query params in the future if necessary.

Let us know if you have any questions!

tgross:
This has been open a while and left unresolved. I'm going to close it out but if you'd like to revisit we'd be happy to reopen.

PR #19080, opened 2023-11-13
set Vault namespace from task in `vault_hook` JWT login

tgross:
The JWT login codepath for the `vault_hook` was missing the Vault namespace, so the login request for non-default namespaces would fail.

### DIFF
--- a/client/vaultclient/vaultclient.go
+++ b/client/vaultclient/vaultclient.go
@@ -41,6 +41,10 @@ type JWTLoginRequest struct {
 	// Nomad client's create_from_role value is used, or the Vault cluster
 	// default role.
 	Role string
+
+	// Namespace is the Vault namespace to use for the login request. If empty,
+	// the Nomad client's Vault configuration namespace will be used.
+	Namespace string
 }
 
 // VaultClient is the interface which nomad client uses to interact with vault and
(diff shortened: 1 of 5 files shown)


client/vaultclient/vaultclient.go
```diff
@@ -285,7 +286,7 @@ func (c *vaultClient) DeriveToken(alloc *structs.Allocation, taskNames []string)
 }
 
 // DeriveTokenWithJWT returns a Vault ACL token using the JWT login endpoint.
-func (c *vaultClient) DeriveTokenWithJWT(ctx context.Context, req JWTLoginRequest) (string, error) {
+func (c *vaultClient) DeriveTokenWithJWT(ctx context.Context, req JWTLoginRequest, ns string) (string, error) {
```
lgfa29, line 289: Thoughts on having the namespace as part of `JWTLoginRequest`? It feels like the namespace is an important part of the request?

lgfa29 (approved):

client/vaultclient/vaultclient.go
tgross: Oh, that would be nicer. Will do.

client/vaultclient/vaultclient.go
```diff
@@ -254,6 +254,7 @@ func (c *vaultClient) Stop() {
 // lock. Helper method for deferring a call that does both.
 func (c *vaultClient) unlockAndUnset() {
 	c.client.SetToken("")
+	c.client.SetNamespace("default")
```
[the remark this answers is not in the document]
tgross: Good catch. Will fix.

PR #19082, opened 2023-11-13
cli: add zero nodes message to `node status`

lgfa29:
Display a message to indicate that there are no nodes registered when `node status` returns zero values.


angrycub (approved):
LGTM!

PR #19083, opened 2023-11-14
docs: Add section for disable restart

Juanadelacuesta:
This PR introduces a new section to the restart block on the job specs that explicitly defines how to disable the restart of a task group on the job specs.


shoenig (approved):
LGTM!

tgross (approved):
:+1:

PR #19085, opened 2023-11-14
Re-add wildcard for test-ui path restrictions

philrenaud:
Seems like I was overly restrictive and forgot my `**` in #19038 , h/t @tgross


tgross (approved):
:+1:

lgfa29 (approved):

Issue #19078, opened 2023-11-13
[qemu] Panic starting job


angrycub:
### Nomad version

```
Nomad v1.7.0-beta.1
BuildDate 2023-11-01T15:20:03Z
Revision abeaf3a309e1793cc24497dad3541877c65a531a
```

### Issue

When running the following job the QEMU job fails to start and the following panic is emitted:

```
panic: runtime error: invalid memory address or nil pointer dereference: alloc_id=777c4540-26f5-3c47-293a-4ea5e8577433 driver=qemu task_name=freedos
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x1b05271]: alloc_id=777c4540-26f5-3c47-293a-4ea5e8577433 driver=qemu task_name=freedos
alloc_id=777c4540-26f5-3c47-293a-4ea5e8577433 driver=qemu task_name=freedos
goroutine 28 [running]:: alloc_id=777c4540-26f5-3c47-293a-4ea5e8577433 driver=qemu task_name=freedos
github.com/hashicorp/nomad/drivers/shared/executor.(*UniversalExecutor).computeMemory(...): alloc_id=777c4540-26f5-3c47-293a-4ea5e8577433 driver=qemu task_name=freedos
        github.com/hashicorp/nomad/drivers/shared/executor/executor_universal_linux.go:243: alloc_id=777c4540-26f5-3c47-293a-4ea5e8577433 driver=qemu task_name=freedos
github.com/hashicorp/nomad/drivers/shared/executor.(*UniversalExecutor).configureCG2(0x7fc52808ca38?, {0x0, 0x0}, 0xc0005da240): alloc_id=777c4540-26f5-3c47-293a-4ea5e8577433 driver=qemu task_name=freedos
        github.com/hashicorp/nomad/drivers/shared/executor/executor_universal_linux.go:209 +0x31: alloc_id=777c4540-26f5-3c47-293a-4ea5e8577433 driver=qemu task_name=freedos
github.com/hashicorp/nomad/drivers/shared/executor.(*UniversalExecutor).configureResourceContainer(0xc0009a9c00, 0xc0005da240, 0x0?): alloc_id=777c4540-26f5-3c47-293a-4ea5e8577433 driver=qemu task_name=freedos
[...]

PR #19089, opened 2023-11-14
qemu: fix panic from missing resources block

tgross:
The `qemu` driver uses our universal executor to run the QEMU command line tool. Because QEMU owns the resource isolation, we don't pass in the resource block that the universal executor uses to configure cgroups and core pinning. This resulted in a panic.

Fix the panic by returning early in the cgroup configuration in the universal executor. This fixes `qemu` but also any third-party drivers that might exist and are using our executor code without passing in the resource block.

In future work, we should ensure that the `resources` block is being translated into QEMU equivalents, so that we have support for things like NUMA-aware scheduling for that driver.

Fixes: https://github.com/hashicorp/nomad/issues/19078
Ref: https://github.com/hashicorp/nomad/issues/19090

### DIFF
--- a/drivers/shared/executor/executor_universal_linux.go
+++ b/drivers/shared/executor/executor_universal_linux.go
@@ -176,6 +176,12 @@ func (e *UniversalExecutor) enterCG1(statsCgroup, cpusetCgroup string) func() {
 }
 
 func (e *UniversalExecutor) configureCG1(cgroup string, command *ExecCommand) {
+
+	// some drivers like qemu entirely own resource management
+	if command.Resources == nil || command.Resources.LinuxResources == nil {
+		return
+	}
+
 	// write memory limits
 	memHard, memSoft := e.computeMemory(command)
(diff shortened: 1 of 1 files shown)


tgross:
Stripped-down version of the job from the original issue that I used for testing:

```hcl
job "freedos" {

  group "g1" {

    network {
      mode = "bridge"
      port "webvnc" {}
    }

    task "freedos" {
      artifact {
        source      = "https://github.com/angrycub/nomad_example_jobs/raw/main/applications/vms/freedos/freedos.img.tgz"
        destination = "local"
        options {
          checksum = "sha256:8d2817126bf46ba2b4fca0b0c49eed2cc208c6f6448651e82c6d973fcba36569"
        }
      }

      driver = "qemu"

      config {
        image_path  = "local/freedos.img"
        accelerator = "kvm"
      }
    }
  }
}
```

schmichael (approved):

PR #10068, opened 2021-02-23
dist: place systemd unit options correctly

shoenig:
This PR places `StartLimitIntervalSec` and `StartLimitBurst` in the
`Unit` section of systemd unit files, rather than the `Service` section.

https://www.freedesktop.org/software/systemd/man/systemd.unit.html

Fixes #10065

Issue #10065, opened 2021-02-22
Systemd unit incorrect config


m1keil:
### Nomad version
```
Nomad v1.0.2 (4c1d4fc6a5823ebc8c3e748daec7b4fda3f11037)
```

### Operating system and Environment details
`Ubuntu 20.04.2 LTS`

### Issue
Nomad unit is incorrectly configure and includes `StartLimitIntervalSec` config under [Service] instead of [Unit].
This results in the following error when starting:

```
/usr/lib/systemd/system/./nomad.service:16: Unknown key name 'StartLimitIntervalSec' in section 'Service', ignoring.
```

### Reproduction steps
Install Nomad from the official package and start service.

shoenig:
Thanks for reporting, @m1keil !

I'm surprised this got by for so long, looks like this was added in 2018 (around the time the option was renamed from `StartLimitInterval` to `StartLimitIntervalSec` and moved from `Service` to `Unit`)

Ubuntu 18.04
```bash
$ systemd --version | head -n1
systemd 237

$ sudo systemctl daemon-reload 
$ sudo systemctl start test

$ sudo systemd-analyze verify test.service
/etc/systemd/system/./test.service:14: Unknown lvalue 'StartLimitIntervalSec' in section 'Service'
```

PR #19098, opened 2023-11-16
Move systemd unit file rate limits from [Service] to [Unit] block

sofixa:
Same issue as #10065:

```
Nov 16 06:45:52 nuc-01 systemd[1]: /lib/systemd/system/nomad.service:30: Unknown key name 'StartLimitIntervalSec' in section 'Service', ignoring.
Nov 16 09:24:11 nuc-01 systemd[1]: /lib/systemd/system/nomad.service:30: Unknown key name 'StartLimitIntervalSec' in section 'Service', ignoring.
```

 which was fixed by @shoenig in #10068, but seems to have disappeared with the [reorganisation](https://github.com/hashicorp/nomad/commits/5a0a8f606ffe87fe8b8a09aa7e827498d1ed8a1a/.release/linux/package/etc/usr/lib/systemd/system/nomad.service?browsing_rename_history=true&new_path=.release/linux/package/usr/lib/systemd/system/nomad.service&original_branch=f95697abaf1a07b7445f2d2a10ff45c519c3866c) of releases.

All the params being moved belong in [Unit] instead of [Service]: https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html

### DIFF
--- a/.release/linux/package/usr/lib/systemd/system/nomad.service
+++ b/.release/linux/package/usr/lib/systemd/system/nomad.service
@@ -10,17 +10,6 @@ After=network-online.target
 #Wants=consul.service
 #After=consul.service
 
-[Service]
-EnvironmentFile=-/etc/nomad.d/nomad.env
-ExecReload=/bin/kill -HUP $MAINPID
-ExecStart=/usr/bin/nomad agent -config /etc/nomad.d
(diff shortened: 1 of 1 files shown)


tgross (approved):
LGTM :+1:

PR #19101, opened 2023-11-16
Add a new parameter to avoid starting a replacement for lost allocs

Juanadelacuesta:
This PR introduces the parameter `preventRescheduleOnLost` which indicates that the task group can't afford to have multiple instances running at the same time. In the case of a node going down, its allocations will be registered as unknown but no replacements will be rescheduled. If the lost node comes back up, the allocs will reconnect and continue to run. 

In case of `max_client_disconnect` also being enabled, if there is a reschedule policy, an error will be returned.
Implements issue #10366

### DIFF
--- a/nomad/job_endpoint_test.go
+++ b/nomad/job_endpoint_test.go
@@ -172,6 +172,7 @@ func TestJobEndpoint_Register_NonOverlapping(t *testing.T) {
 			Namespace: structs.DefaultNamespace,
 		},
 	}
+
 	var alloc *structs.AllocListStub
 	testutil.Wait(t, func() (bool, error) {
 		resp := structs.JobAllocationsResponse{}
@@ -183,6 +184,7 @@ func TestJobEndpoint_Register_NonOverlapping(t *testing.T) {
 		alloc = resp.Allocations[0]
 		return true, nil
 	})
+
 	must.Eq(t, alloc.NodeID, node.ID)
 	must.Eq(t, alloc.DesiredStatus, structs.AllocDesiredStatusRun)
 	must.Eq(t, alloc.ClientStatus, structs.AllocClientStatusPending)
--- a/nomad/job_endpoint_test.go
+++ b/nomad/job_endpoint_test.go
@@ -172,7 +172,6 @@ func TestJobEndpoint_Register_NonOverlapping(t *testing.T) {
 			Namespace: structs.DefaultNamespace,
 		},
 	}
-
 	var alloc *structs.AllocListStub
 	testutil.Wait(t, func() (bool, error) {
 		resp := structs.JobAllocationsResponse{}
@@ -184,7 +183,6 @@ func TestJobEndpoint_Register_NonOverlapping(t *testing.T) {
 		alloc = resp.Allocations[0]
 		return true, nil
 	})
-
 	must.Eq(t, alloc.NodeID, node.ID)
 	must.Eq(t, alloc.DesiredStatus, structs.AllocDesiredStatusRun)
 	must.Eq(t, alloc.ClientStatus, structs.AllocClientStatusPending)
--- a/nomad/plan_apply.go
+++ b/nomad/plan_apply.go
@@ -738,10 +738,10 @@ func evaluateNodePlan(snap *state.StateSnapshot, plan *structs.Plan, nodeID stri
 		}
 		return false, "node is disconnected and contains invalid updates", nil
 	} else if node.Status == structs.NodeStatusDown {
-		if isValidForLostNode(plan, node.ID) {
+		if isValidForDownNode(plan, node.ID) {
 			return true, "", nil
 		}
-		return false, "node is lost and contains invalid updates", nil
+		return false, "node is down and contains invalid updates", nil
 	} else if node.Status != structs.NodeStatusReady {
 		return false, "node is not ready for placements", nil
 	}
--- a/nomad/plan_apply.go
+++ b/nomad/plan_apply.go
@@ -801,7 +801,8 @@ func isValidForDisconnectedNode(plan *structs.Plan, nodeID string) bool {
 // as non reschedulables when lost.
 func isValidForLostNode(plan *structs.Plan, nodeID string) bool {
 	for _, alloc := range plan.NodeAllocation[nodeID] {
-		if !(alloc.ClientStatus == structs.AllocClientStatusUnknown && alloc.AvoidRescheduleOnLost()) {
+		if !(alloc.ClientStatus == structs.AllocClientStatusUnknown && alloc.AvoidRescheduleOnLost()) &&
+			(alloc.ClientStatus != structs.AllocClientStatusLost) {
 			return false
 		}
 	}
--- a/nomad/plan_apply.go
+++ b/nomad/plan_apply.go
@@ -801,7 +801,7 @@ func isValidForDisconnectedNode(plan *structs.Plan, nodeID string) bool {
 // as non reschedulables when lost.
 func isValidForLostNode(plan *structs.Plan, nodeID string) bool {
 	for _, alloc := range plan.NodeAllocation[nodeID] {
-		if !(alloc.ClientStatus == structs.AllocClientStatusUnknown && alloc.SingleInstanceOnLost()) {
+		if !(alloc.ClientStatus == structs.AllocClientStatusUnknown && alloc.AvoidRescheduleOnLost()) {
 			return false
 		}
 	}
--- a/nomad/plan_apply.go
+++ b/nomad/plan_apply.go
@@ -742,12 +742,6 @@ func evaluateNodePlan(snap *state.StateSnapshot, plan *structs.Plan, nodeID stri
 			return true, "", nil
 		}
 		return false, "node is lost and contains invalid updates", nil
-
-	} else if node.Status == structs.NodeStatusDown {
-		if isValidForDisconnectedNode(plan, node.ID) {
-			return true, "", nil
-		}
-		return false, "node is lost and contains invalid updates", nil
 	} else if node.Status != structs.NodeStatusReady {
 		return false, "node is not ready for placements", nil
 	}
@@ -802,11 +796,12 @@ func isValidForDisconnectedNode(plan *structs.Plan, nodeID string) bool {
 	return true
 }
 
-// The plan is only valid for disconnected nodes if it only contains
-// updates to mark allocations as unknown.
+// The plan is only valid for lost nodes if it only contains
+// updates to mark allocations as unknown and those allocations are configured
+// as non reschedulables when lost.
 func isValidForLostNode(plan *structs.Plan, nodeID string) bool {
 	for _, alloc := range plan.NodeAllocation[nodeID] {
-		if alloc.ClientStatus != structs.AllocClientStatusUnknown && !alloc.SingleInstanceOnLost() {
+		if !(alloc.ClientStatus == structs.AllocClientStatusUnknown && alloc.SingleInstanceOnLost()) {
 			return false
 		}
 	}
--- a/nomad/plan_apply.go
+++ b/nomad/plan_apply.go
@@ -806,7 +806,7 @@ func isValidForDisconnectedNode(plan *structs.Plan, nodeID string) bool {
 // updates to mark allocations as unknown.
 func isValidForLostNode(plan *structs.Plan, nodeID string) bool {
 	for _, alloc := range plan.NodeAllocation[nodeID] {
-		if alloc.ClientStatus != structs.AllocClientStatusUnknown && !alloc.RescheduleOnLost() {
+		if alloc.ClientStatus != structs.AllocClientStatusUnknown && !alloc.SingleInstanceOnLost() {
 			return false
 		}
 	}
--- a/nomad/plan_apply.go
+++ b/nomad/plan_apply.go
@@ -796,12 +796,12 @@ func isValidForDisconnectedNode(plan *structs.Plan, nodeID string) bool {
 	return true
 }
 
-// The plan is only valid for lost nodes if it only contains
+// The plan is only valid for a node down if it only contains
 // updates to mark allocations as unknown and those allocations are configured
-// as non reschedulables when lost.
-func isValidForLostNode(plan *structs.Plan, nodeID string) bool {
+// as non reschedulables when lost or if the allocs are being updated to lost.
+func isValidForDownNode(plan *structs.Plan, nodeID string) bool {
 	for _, alloc := range plan.NodeAllocation[nodeID] {
-		if !(alloc.ClientStatus == structs.AllocClientStatusUnknown && alloc.AvoidRescheduleOnLost()) &&
+		if !(alloc.ClientStatus == structs.AllocClientStatusUnknown && alloc.PreventRescheduleOnLost()) &&
 			(alloc.ClientStatus != structs.AllocClientStatusLost) {
 			return false
 		}
--- a/nomad/plan_apply.go
+++ b/nomad/plan_apply.go
@@ -737,6 +737,17 @@ func evaluateNodePlan(snap *state.StateSnapshot, plan *structs.Plan, nodeID stri
 			return true, "", nil
 		}
 		return false, "node is disconnected and contains invalid updates", nil
+	} else if node.Status == structs.NodeStatusDown {
+		if isValidForLostNode(plan, node.ID) {
+			return true, "", nil
+		}
+		return false, "node is lost and contains invalid updates", nil
+
+	} else if node.Status == structs.NodeStatusDown {
+		if isValidForDisconnectedNode(plan, node.ID) {
+			return true, "", nil
+		}
+		return false, "node is lost and contains invalid updates", nil
 	} else if node.Status != structs.NodeStatusReady {
 		return false, "node is not ready for placements", nil
 	}
@@ -784,6 +795,18 @@ func evaluateNodePlan(snap *state.StateSnapshot, plan *structs.Plan, nodeID stri
 func isValidForDisconnectedNode(plan *structs.Plan, nodeID string) bool {
 	for _, alloc := range plan.NodeAllocation[nodeID] {
 		if alloc.ClientStatus != structs.AllocClientStatusUnknown {
+			return false
+		}
+	}
+
+	return true
+}
+
+// The plan is only valid for disconnected nodes if it only contains
+// updates to mark allocations as unknown.
+func isValidForLostNode(plan *structs.Plan, nodeID string) bool {
+	for _, alloc := range plan.NodeAllocation[nodeID] {
+		if alloc.ClientStatus != structs.AllocClientStatusUnknown && !alloc.RescheduleOnLost() {
 			return false
 		}
 	}
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -6641,6 +6641,11 @@ type TaskGroup struct {
 	// MaxClientDisconnect, if set, configures the client to allow placed
 	// allocations for tasks in this group to attempt to resume running without a restart.
 	MaxClientDisconnect *time.Duration
+
+	// RescheduleOnLost is used to control how allocations on disconnected
+	// nodes are handled. For backwards compatibility, it defaults to true.
+	// When true, such jobs are rescheduled.
+	RescheduleOnLost bool
 }
 
 func (tg *TaskGroup) Copy() *TaskGroup {
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -8876,7 +8876,7 @@ const (
 	// TaskPluginHealthy indicates that a plugin managed by Nomad became healthy
 	TaskPluginHealthy = "Plugin became healthy"
 
-	// TaskClientReconnected indicates that the client running the task disconnected.
+	// TaskClientReconnected indicates that the client running the task reconnected.
 	TaskClientReconnected = "Reconnected"
 
 	// TaskWaitingShuttingDownDelay indicates that the task is waiting for
@@ -10999,7 +10999,6 @@ func (a *Allocation) DisconnectTimeout(now time.Time) time.Time {
 	tg := a.Job.LookupTaskGroup(a.TaskGroup)
 
 	timeout := tg.MaxClientDisconnect
-
 	if timeout == nil {
 		return now
 	}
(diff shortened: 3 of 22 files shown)


nomad/job_endpoint_test.go
```diff
@@ -172,6 +172,7 @@ func TestJobEndpoint_Register_NonOverlapping(t *testing.T) {
 			Namespace: structs.DefaultNamespace,
 		},
 	}
+
```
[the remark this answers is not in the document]
Juanadelacuesta: Let's remove them :)

nomad/structs/structs.go
```diff
@@ -4666,6 +4666,13 @@ func (j *Job) Validate() error {
 				fmt.Errorf("Job task group %s has count %d. Count cannot exceed 1 with system scheduler",
 					tg.Name, tg.Count))
 		}
+
+		if tg.MaxClientDisconnect != nil &&
```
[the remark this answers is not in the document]
Juanadelacuesta: "0" is a valid configuration for `max_client_disconnect` as far as I recall :s

nomad/plan_apply.go
```diff
@@ -790,3 +795,16 @@ func isValidForDisconnectedNode(plan *structs.Plan, nodeID string) bool {
 
 	return true
 }
+
+// The plan is only valid for lost nodes if it only contains
+// updates to mark allocations as unknown and those allocations are configured
+// as non reschedulables when lost.
```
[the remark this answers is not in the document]
Juanadelacuesta: You are right, but when neither `max_client_disconnect` or `single_instance_on_lost` are configured, the node status at this point is `ready`.

nomad/structs/structs.go
```diff
@@ -4666,6 +4666,13 @@ func (j *Job) Validate() error {
...
+			tg.ReschedulePolicy.Attempts > 0 &&
+			tg.SingleInstanceOnLost {
+			err := fmt.Errorf("having max_client_disconnect enable along with a reschedule policy can lead to having multiple instances of a task running at the same time")
```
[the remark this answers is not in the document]
Juanadelacuesta: Well, the client will be considered disconnected instead of lost for the duration of `max_client_disconnect`, but that is it.

nomad/structs/structs.go
```diff
@@ -6635,6 +6642,11 @@ type TaskGroup struct {
 	// MaxClientDisconnect, if set, configures the client to allow placed
 	// allocations for tasks in this group to attempt to resume running without a restart.
 	MaxClientDisconnect *time.Duration
+
+	// SingleInstanceOnLost is used to signal if multiple instances of the same
+	// task can be running at the same time, it controls if a replacement is triggered
+	// when the task state is unknown
```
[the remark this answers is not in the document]
Juanadelacuesta: Updated!

scheduler/reconcile.go
```diff
@@ -460,31 +462,48 @@ func (a *allocReconciler) computeGroup(groupName string, all allocSet) bool {
...
-		untainted = untainted.union(untaintedDisconnecting)
-		rescheduleLater = append(rescheduleLater, laterDisconnecting...)
+			// Find delays for any disconnecting allocs that have max_client_disconnect,
+			// create followup evals, and update the ClientStatus to unknown.
+			timeoutLaterEvals = a.createTimeoutLaterEvals(disconnecting, tg.Name)
 
-		// Find delays for any disconnecting allocs that have max_client_disconnect,
-		// create followup evals, and update the ClientStatus to unknown.
-		timeoutLaterEvals = a.createTimeoutLaterEvals(disconnecting, tg.Name)
+		case tg.SingleInstanceOnLost:
```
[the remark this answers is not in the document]
Juanadelacuesta: will do!

scheduler/reconcile_util.go
```diff
@@ -264,38 +262,33 @@ func (a allocSet) filterByTainted(taintedNodes map[string]*structs.Node, serverS
...
 					lost[alloc.ID] = alloc
 					continue
 				}
-			case structs.NodeStatusReady:
-				// Filter reconnecting allocs on a node that is now connected.
-				if reconnect {
-					if expired {
-						lost[alloc.ID] = alloc
+
+			} else {
```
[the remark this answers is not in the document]
Juanadelacuesta: it is, should I change it for a switch? Despite the fact there are no more cases?

scheduler/reconcile_test.go
```diff
@@ -42,6 +42,481 @@ var (
...
+						Ignore: 3,
+					},
+				},
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			require.NotEqual(t, 0, tc.allocCount, "invalid test case: alloc count must be greater than zero")
```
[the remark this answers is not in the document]
Juanadelacuesta: It was there before, so I didnt want to touch it, but lets remove it :)

tgross (changes requested):
This is getting really close @Juanadelacuesta! I've left some comments and surfaced a few of my earlier comments that got overwritten by changes without getting resolved.

scheduler/util.go
```diff
@@ -222,6 +222,11 @@ func tasksUpdated(jobA, jobB *structs.Job, taskGroup string) comparison {
 		return difference("number of tasks", lenA, lenB)
 	}
 
+	// Check for AvoidRescheduleOnLost changes
+	if a.AvoidRescheduleOnLost != b.AvoidRescheduleOnLost {
+		return difference("reschedule on lost", a.AvoidRescheduleOnLost, b.AvoidRescheduleOnLost)
+	}
```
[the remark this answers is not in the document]
Juanadelacuesta: it got lost indeed, let me make the change

scheduler/reconcile.go
```diff
@@ -460,31 +476,52 @@ func (a *allocReconciler) computeGroup(groupName string, all allocSet) bool {
...
-		untaintedDisconnecting, rescheduleDisconnecting, laterDisconnecting := disconnecting.filterByRescheduleable(a.batch, true, a.now, a.evalID, a.deployment)
-
-		rescheduleNow = rescheduleNow.union(rescheduleDisconnecting)
-		untainted = untainted.union(untaintedDisconnecting)
-		rescheduleLater = append(rescheduleLater, laterDisconnecting...)
+		// If MaxClientDisconnect is enabled as well as tg.AvoidRescheduleOnLost,
+		// the reschedule policy won't be enable and the lost allocations
+		// wont be rescheduled, there is no need to specifically do any extra processing.
+		// If MaxClientDisconnect is not enabled, then AvoidRescheduleOnLost
+		// requires handling.
```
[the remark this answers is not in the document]
Juanadelacuesta: I've been debating myself over that one, `AvoidRescheduleOnLost` is "ignored here because it is already taken care of by the disabled reschedule policy so techically it is not necessary to check here. I could add it and it would make it explicit in the code, but the end result would be the same. I opted for simplicity, but conceptually, they are not mutually exclusive.

website/content/docs/job-specification/group.mdx
```diff
@@ -63,6 +63,20 @@ job "docs" {
...
+  or goes down, this allocations wont be rescheduled and will show up as `unknown`
+  until the node comes back up or it is manually restarted. 
+  This behaviour will only modify the reschedule process on the server.
+  To modify the allocation behaviour on the client, see `stop_after_client_disconnect`.
+
+  Setting `max_client_disconnect` and `avoid_reschedule_on_list=true` at the same time 
+  requires that [rescheduling is disabled entirely][] If [`max_client_disconnect`](#max_client_disconnect) 
+  is set and `avoid_reschedule_on_lost=true`, allocations on disconnected nodes will be
+  `unknown` until the `max_client_disconnect` window expires, at which point they 
+  become `lost` but will not be rescheduled.
```
tgross, line 78: Notes from our sidebar conversation:
* We'll need to correct this line.
* And add some documentation on how to recover the unknown allocation if you want (maybe just re-run the job)
* Once those are done this should be LGTM!

website/content/docs/job-specification/group.mdx
```diff
@@ -63,6 +63,31 @@ job "docs" {
...
+
+  Setting `max_client_disconnect` and `avoid_reschedule_on_lost=true` at the same 
+  time requires that [rescheduling is disabled entirely][].
+  If [`max_client_disconnect`](#max_client_disconnect) is set and 
+  `avoid_reschedule_on_lost=true`, allocations on disconnected nodes will be
+  `unknown` until the `max_client_disconnect` window expires, at which point 
+  the node will be transition from `disconnected` to `down`. The allocation 
+  will remain as `unknown` and won't be rescheduled.
+
+  To get the job up and running again it has to be manually stoped and run again
```
tgross, line 84:
```suggestion
  To get the job up and running again it has to be manually stopped and run again
```

nomad/plan_apply.go
```diff
@@ -737,6 +737,11 @@ func evaluateNodePlan(snap *state.StateSnapshot, plan *structs.Plan, nodeID stri
 			return true, "", nil
 		}
 		return false, "node is disconnected and contains invalid updates", nil
+	} else if node.Status == structs.NodeStatusDown {
+		if isValidForLostNode(plan, node.ID) {
+			return true, "", nil
+		}
+		return false, "node is lost and contains invalid updates", nil
```
lgfa29, line 744:
```suggestion
		if isValidForDownNode(plan, node.ID) {
			return true, "", nil
		}
		return false, "node is down and contains invalid updates", nil
```

Is `lost` a valid status for nodes? I think it only applies to allocs.

nomad/structs/structs.go
```diff
@@ -4672,6 +4672,13 @@ func (j *Job) Validate() error {
 				fmt.Errorf("Job task group %s has count %d. Count cannot exceed 1 with system scheduler",
 					tg.Name, tg.Count))
 		}
+
+		if tg.MaxClientDisconnect != nil &&
+			tg.ReschedulePolicy.Attempts > 0 &&
+			tg.AvoidRescheduleOnLost {
+			err := fmt.Errorf("max_client_disconnect and single_instance_on_lost cannot be enabled when rechedule.attempts > 0")
```
lgfa29, line 4679:
```suggestion
			err := fmt.Errorf("max_client_disconnect and avoid_reschedule_on_lost cannot be enabled when rechedule.attempts > 0")
```

I think it was renamed?

nomad/structs/structs.go
```diff
@@ -11020,6 +11032,19 @@ func (a *Allocation) SupportsDisconnectedClients(serverSupportsDisconnectedClien
...
+// when lost.
+func (a *Allocation) AvoidRescheduleOnLost() bool {
+	if a.Job != nil {
+		tg := a.Job.LookupTaskGroup(a.TaskGroup)
+		if tg != nil {
+			return tg.AvoidRescheduleOnLost
+		}
+	}
+
+	return false
```
lgfa29, line 11045: Have you hit a `panic` where `a.Job` was `nil`? 

We denormalize the allocation in some parts of the scheduler and remove `Job` from the allocation. I suspect this may be the case for `isValidForLostNode()` because it acts on the plan, so we may be incorrectly returning `false` here.

lgfa29 (commented):
Leaving some early comments here, I will need some extra brain juice to reason about the reconciler part 😄

nomad/structs/structs.go
Juanadelacuesta: I have not seen it yet, it returns the default value, im not sure what to do if the alloc does not have the job

nomad/structs/structs.go
```diff
@@ -4672,6 +4672,13 @@ func (j *Job) Validate() error {
 				fmt.Errorf("Job task group %s has count %d. Count cannot exceed 1 with system scheduler",
 					tg.Name, tg.Count))
 		}
+
+		if tg.MaxClientDisconnect != nil &&
+			tg.ReschedulePolicy.Attempts > 0 &&
+			tg.PreventRescheduleOnLost {
+			err := fmt.Errorf("max_client_disconnect and prevent_reschedule_on_lost cannot be enabled when rechedule.attempts > 0")
+			mErr.Errors = append(mErr.Errors, err)
+		}
```
lgfa29, line 4681: I'm still not sure about this validation as they seem to mix two scenarios. For example, a group configured like this:

```hcl
group "..." {
  max_client_disconnect      = "12h"
  prevent_reschedule_on_lost = true
  
  reschedule {
    attempts = 10
  }
}
```

As an user, I think my intention in writing this job would be something like this:

- Try to reschedule my allocations at least 10 times if they fail.
- If a node running an alloc for this group misses heartbeats, consider it `disconnected` instead of `down` for 12h as it may come back up, but don't create a replacement allocation. After the 12h consider the allocation as `lost` and create a replacement.

So in a scenario where this group has 3 allocs, if a node misses its hearbeats I would expect there to still be 3 allocs, 2 `running` and 1 `unknown`. After 12h, if the node doesn't reconnect, it will considered `down` and the alloc `lost`, and a replacement is created, resulting in 4 allocs: 3 `running` and 1 `lost`.

This is consistent with the documented behaviour of the `reschedule` block:
> Nomad will attempt to schedule the allocation on another node if any of its task statuses become `failed`.

Since the alloc is `unknown`, none of the tasks should be in `failed` state, so the `reschedule` block doesn't apply. By forcing `reschedule.attempts` to be 0 we're preventing users from being able to handle allocation failures.

In hindsight, I think the confusion started with `max_client_disconnect`, where the docs mention:
> Replacement allocations will be scheduled according to the allocations' reschedule policy until the disconnected client reconnects. 

This conflates cluster state (a node misses heartbeat) with allocation state (a task fails), making it impossible to properly handle both cases as any combination of them is valid.

If we were to refactor this part of the code I think I would create a new block to disentangle these two scenarios. `reschedule` is applied when allocations fail, and a new `disconnected` (or a better name 😅) block to handle nodes missing heartbeats.

This new block could then have configurations like these:

* Create a replacement for `unknown` allocs but try to reconnect the original alloc if the node comes back within 12h (equivalent to today's `max_client_disconnect`).
    ```hcl
    disconnected {
      replace_unknown     = true
      max_disconnect_time = "12h"
    }
    ```

* Never create a replacement for `unkown` allocs (I think this is what we're trying to accomplish here with `prevent_reschedule_on_lost`).
    ```hcl
    disconnected {
      replace_unknown = false
    }
    ```

* Wait 12h before creating a replacement for `unknown` allocs (I don't think this is currently possible to express, even with  this PR, and I think it's a valid use case).
    ```hcl
    disconnected {
      replace_unknown     = false
      max_disconnect_time = "12h"
    }
    ```

Parallel to all of these configs, I could set different `reschedule` policies to handle allocations that fail. Regardless of client status, I probably do want to reschedule them (or maybe I don't 😄).

This is a _lot_ to change that would require proper planning and scoping, but until then I think the relationship between all of these flags can get confusing.

For this PR specifically, if the intention is to prevent the use of `max_client_disconnect` and `prevent_reschedule_on_lost` with a `reschedule` policy that allows for new allocations to be created, then I think we also need to check the value of `reschedule.unlimited`, as its often used with with `reschedule.attempts = 0` but does create replacements.

scheduler/util_test.go
```diff
@@ -547,6 +547,7 @@ func TestTasksUpdated(t *testing.T) {
 	// Compare changed Template ErrMissingKey
 	j30.TaskGroups[0].Tasks[0].Templates[0].ErrMissingKey = true
 	must.True(t, tasksUpdated(j29, j30, name).modified)
+
```
lgfa29, line 550:
```suggestion
```

website/content/docs/job-specification/group.mdx
```diff
@@ -63,6 +63,31 @@ job "docs" {
   rescheduling strategy. Nomad will then attempt to schedule the task on another
   node if any of the group allocation statuses become "failed".
 
+- `prevent_reschedule_on_lost` `(bool: false)` - Defines the reschedule behaviour 
+  of an allocation when the node it is running on becomes disconnect or lost.
```
lgfa29, line 67:
```suggestion
- `prevent_reschedule_on_lost` `(bool: false)` - Defines the reschedule behaviour 
  of an allocation when the node it is running on misses heartbeats. 
```

website/content/docs/job-specification/group.mdx
```diff
@@ -63,6 +63,31 @@ job "docs" {
   rescheduling strategy. Nomad will then attempt to schedule the task on another
   node if any of the group allocation statuses become "failed".
 
+- `prevent_reschedule_on_lost` `(bool: false)` - Defines the reschedule behaviour 
+  of an allocation when the node it is running on becomes disconnect or lost. 
+  When enabled, if the node it is running on becomes disconnected 
+  or goes down, this allocations wont be rescheduled and will show up as `unknown`
+  until the node comes back up or it is manually restarted.
```
lgfa29, line 70:
```suggestion
  When enabled, the allocation is not rescheduled and remains in the `unknown` status
  until the node comes back up or the allocation is manually restarted. 
```

website/content/docs/job-specification/group.mdx
```diff
@@ -63,6 +63,31 @@ job "docs" {
...
   node if any of the group allocation statuses become "failed".
 
+- `prevent_reschedule_on_lost` `(bool: false)` - Defines the reschedule behaviour 
+  of an allocation when the node it is running on becomes disconnect or lost. 
+  When enabled, if the node it is running on becomes disconnected 
+  or goes down, this allocations wont be rescheduled and will show up as `unknown`
+  until the node comes back up or it is manually restarted. 
+  
+  This behaviour will only modify the reschedule process on the server.
+  To modify the allocation behaviour on the client, see
```
lgfa29, line 73:
```suggestion
  This behaviour only modifies the reschedule process on the server.
  To modify the allocation behaviour on the client, use 
```

website/content/docs/job-specification/group.mdx
```diff
@@ -63,6 +63,31 @@ job "docs" {
...
+  When enabled, if the node it is running on becomes disconnected 
+  or goes down, this allocations wont be rescheduled and will show up as `unknown`
+  until the node comes back up or it is manually restarted. 
+  
+  This behaviour will only modify the reschedule process on the server.
+  To modify the allocation behaviour on the client, see 
+  [`stop_after_client_disconnect`](#stop_after_client_disconnect) .
+
+  Setting `max_client_disconnect` and `prevent_reschedule_on_lost=true` at the same 
+  time requires that [rescheduling is disabled entirely][].
```
lgfa29, line 77: Broken link. Where should `rescheduling is disabled entirely` point to?

website/content/docs/job-specification/group.mdx
```diff
@@ -63,6 +63,31 @@ job "docs" {
...
+  of an allocation when the node it is running on becomes disconnect or lost. 
+  When enabled, if the node it is running on becomes disconnected 
+  or goes down, this allocations wont be rescheduled and will show up as `unknown`
+  until the node comes back up or it is manually restarted. 
+  
+  This behaviour will only modify the reschedule process on the server.
+  To modify the allocation behaviour on the client, see 
+  [`stop_after_client_disconnect`](#stop_after_client_disconnect) .
+
+  Setting `max_client_disconnect` and `prevent_reschedule_on_lost=true` at the same
```
lgfa29, line 76:
```suggestion
  Setting `max_client_disconnect` and `prevent_reschedule_on_lost = true` at the same 
```

website/content/docs/job-specification/group.mdx
```diff
@@ -63,6 +63,31 @@ job "docs" {
...
+  
+  This behaviour will only modify the reschedule process on the server.
+  To modify the allocation behaviour on the client, see 
+  [`stop_after_client_disconnect`](#stop_after_client_disconnect) .
+
+  Setting `max_client_disconnect` and `prevent_reschedule_on_lost=true` at the same 
+  time requires that [rescheduling is disabled entirely][].
+  If [`max_client_disconnect`](#max_client_disconnect) is set and 
+  `prevent_reschedule_on_lost=true`, allocations on disconnected nodes will be
+  `unknown` until the `max_client_disconnect` window expires, at which point
```
lgfa29, line 80:
```suggestion
  If [`max_client_disconnect`](#max_client_disconnect) is set and 
  `prevent_reschedule_on_lost = true`, allocations on disconnected nodes remains with status
  `unknown` until the `max_client_disconnect` window expires, at which point 
```

website/content/docs/job-specification/group.mdx
```diff
@@ -63,6 +63,31 @@ job "docs" {
...
+  To modify the allocation behaviour on the client, see 
+  [`stop_after_client_disconnect`](#stop_after_client_disconnect) .
+
+  Setting `max_client_disconnect` and `prevent_reschedule_on_lost=true` at the same 
+  time requires that [rescheduling is disabled entirely][].
+  If [`max_client_disconnect`](#max_client_disconnect) is set and 
+  `prevent_reschedule_on_lost=true`, allocations on disconnected nodes will be
+  `unknown` until the `max_client_disconnect` window expires, at which point 
+  the node will be transition from `disconnected` to `down`. The allocation 
+  will remain as `unknown` and won't be rescheduled.
```
lgfa29, line 82:
```suggestion
  the node transitions from `disconnected` to `down`. The allocation 
  remain in the `unknown` status and is not rescheduled.
```

website/content/docs/job-specification/group.mdx
```diff
@@ -63,6 +63,31 @@ job "docs" {
...
+
+  To get the job up and running again it has to be manually stopped and run again
+  or [restarted][restart] using the `reschedule` option:
+
+  ```plaintext
+  `nomad job restart -reschedule <job>`
+  ```
```
lgfa29, line 89:
```suggestion
  The `unknown` allocation has to be manually stopped to run it again.

  ```plaintext
  `nomad alloc stop  <alloc ID>`
  ```
```

I don't think we want to restart entire job if just a few allocations disconnect?

website/content/docs/job-specification/group.mdx
lgfa29, line 82: This interaction between `prevent_reschedule_on_lost` and `max_client_disconnect` should probably be documented in both places. So perhaps we should move this part to a separate section and link to there from both configs?

lgfa29 (approved):
Great job getting this implemented, it's such a tricky part of the code 😅 

I left a few docs suggestions and also some food for thought for potentially rethinking and refactoring this part of the code and the intended behaviours we want, but not something we should do here.

One thing to double-check is that a block like this _will_ create replacements, even with `attempts = 0`, but it seems like the intention is disallow the use of this new feature in this situation?

```hcl
restart {
  attempts  = 0
  unlimited = true
}
```

Issue #19076, opened 2023-11-13
e2e: RescheduleE2ETest/TestDefaultReschedule is failing


shoenig:
https://github.com/hashicorp/nomad-e2e/actions/runs/6845894866/job/18611827943

```
=== RUN   TestE2E/Rescheduling/*rescheduling.RescheduleE2ETest/TestDefaultReschedule
    rescheduling.go:101: 
        	Error Trace:	/home/runner/actions-runner/_work/nomad-e2e/nomad-e2e/nomad/e2e/rescheduling/rescheduling.go:101
        	Error:      	Received unexpected error:
        	            	alloc status check failed: got []string{"failed", "pending", "pending", "failed"}
        	            	allocs: []map[string]string{
        	            	    {"Created":"2023-11-13T05:55:36Z", "Desired":"run", "Eval ID":"38db7112-b05a-326f-c59d-b5667f9162ed", "ID":"76f97e53-9e34-ac58-04f9-1cf70cd0cac3", "Modified":"2023-11-13T05:55:38Z", "Node ID":"b12ca997-c368-5805-044c-9272109752fc", "Node Name":"ip-172-31-95-44", "Status":"failed", "Task Group":"t", "Version":"0"},
        	            	    {"Created":"2023-11-13T05:55:06Z", "Desired":"run", "Eval ID":"d753dbba-54a0-ada5-b4d2-aa3d124fcf1d", "ID":"206bbee8-64c1-c724-7a13-ed836d76edec", "Modified":"2023-11-13T05:55:06Z", "Node ID":"2741b635-1ae5-7fe4-ce68-1a9f71de0eff", "Node Name":"ip-172-31-89-138", "Status":"pending", "Task Group":"t", "Version":"0"},
        	            	    {"Created":"2023-11-13T05:55:06Z", "Desired":"run", "Eval ID":"d753dbba-54a0-ada5-b4d2-aa3d124fcf1d", "ID":"2ef04887-c59f-d470-6c20-d0ce42608c44", "Modified":"2023-11-13T05:55:06Z", "Node ID":"2f61b196-f370-1a63-9f88-002745842f65", "Node Name":"ip-172-31-94-250", "Status":"pending", "Task Group":"t", "Version":"0"},
[...]

PR #19105, opened 2023-11-16
e2e: fix and modernize rescheduling test

tgross:
The E2E test suite for rescheduling had a few bugs:
* Using the command line to stop a job with a failing deployment returns a non-zero exit code, which would cause an otherwise passing test to fail.
* Two of the input jobs were actually invalid but were only correctly detected as such because of #17342

This changeset also updates the whole test suite to move it off the v1 "framework". The test assertions are otherwise left untouched.

Fixes: https://github.com/hashicorp/nomad/issues/19076

---

Example run:

```
$ go test -v ./e2e/rescheduling -count=1 -timeout=20m
=== RUN   TestRescheduling_Service_NoReschedule
--- PASS: TestRescheduling_Service_NoReschedule (1.33s)
=== RUN   TestRescheduling_System_NoReschedule
--- PASS: TestRescheduling_System_NoReschedule (1.10s)
=== RUN   TestRescheduling_Default
--- PASS: TestRescheduling_Default (36.17s)
=== RUN   TestRescheduling_MaxAttempts
--- PASS: TestRescheduling_MaxAttempts (2.28s)
=== RUN   TestRescheduling_Success
--- PASS: TestRescheduling_Success (1.10s)
=== RUN   TestRescheduling_WithUpdate
--- PASS: TestRescheduling_WithUpdate (4.26s)
=== RUN   TestRescheduling_WithCanary
--- PASS: TestRescheduling_WithCanary (6.70s)
=== RUN   TestRescheduling_WithCanaryAutoRevert
--- PASS: TestRescheduling_WithCanaryAutoRevert (11.05s)
=== RUN   TestRescheduling_MaxParallel
--- PASS: TestRescheduling_MaxParallel (11.54s)
=== RUN   TestRescheduling_MaxParallelAutoRevert
--- PASS: TestRescheduling_MaxParallelAutoRevert (12.06s)
=== RUN   TestRescheduling_ProgressDeadline
--- PASS: TestRescheduling_ProgressDeadline (22.12s)
=== RUN   TestRescheduling_ProgressDeadlineFail
--- PASS: TestRescheduling_ProgressDeadlineFail (21.23s)
PASS
ok      github.com/hashicorp/nomad/e2e/rescheduling     130.965s
```


e2e/rescheduling/rescheduling_test.go
```diff
@@ -0,0 +1,543 @@
...
+	}
+
+	must.Eq(t, 1, statuses["failed"],
+		must.Sprintf("expected only 1 failed deployment, got:\n%s", out))
+	must.Eq(t, 2, statuses["successful"],
+		must.Sprintf("expected 2 successful deployments, got:\n%s", out))
+}
+
+// TestRescheduling_ProgressDeadline verifies the progress deadline is reset with
+// each healthy allocation, and that a rescheduled allocation does not.
```
gulducat, line 438: again not due to your change, but this test doesn't seem to confirm the latter point - that a rescheduled allocation does not reset the progress deadline.  it only checks that deadline _is_ reset by a healthy alloc.

edit: oh, reading below, `TestRescheduling_ProgressDeadlineFail` covers that case, so it's only this comment that's misleading, the code is fine.

e2e/rescheduling/rescheduling_test.go
```diff
@@ -0,0 +1,543 @@
...
+	must.NoError(t,
+		e2eutil.WaitForAllocStatusComparison(
+			func() ([]string, error) { return e2eutil.AllocStatuses(jobID, ns) },
+			func(got []string) bool {
+				for _, status := range got {
+					if status != "failed" {
+						return false
+					}
+				}
+				return true
```
gulducat, line 511: returning `false` within the for loop, then `true` after it like this would fix (the first part of) my comment on the `TestRescheduling_Success` test up above (swapping `!= "failed"` for `!= "running"`).  this one does appropriately confirm that _all_ allocs are "failed"

gulducat (approved):
just a couple tidbits for consideration, otherwise LGTM!

though I'm not seeing where this is being addressed?
> Two of the input jobs were actually invalid but were only correctly detected as such because of https://github.com/hashicorp/nomad/pull/17342

e2e/rescheduling/rescheduling_test.go
```diff
@@ -0,0 +1,543 @@
...
+// reschedule
+func TestRescheduling_Service_NoReschedule(t *testing.T) {
+	jobID := "test-no-reschedule-" + uuid.Generate()[0:8]
+	must.NoError(t, e2eutil.Register(jobID, "./input/norescheduling_service.nomad"))
+
+	t.Cleanup(func() {
+		e2eutil.StopJob(jobID, "-purge", "-detach")
+		_, err := e2eutil.Command("nomad", "system", "gc")
+		test.NoError(t, err)
+	})
```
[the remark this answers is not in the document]
tgross: Yeah, the thing that you'll find if you dig into a lot of tests is that "we need to cleanup" is universal but "what do we need to cleanup" tends to vary a lot between test suites. So in this file pretty everything is running one job. But a lot of the other test suites run two jobs, or a job and a namespace, or a job and a CSI volume, etc, etc. Which makes the common helper a lot less useful.

e2e/rescheduling/rescheduling_test.go
```diff
@@ -0,0 +1,543 @@
...
+
+	must.NoError(t,
+		e2eutil.WaitForAllocStatusComparison(
+			func() ([]string, error) { return e2eutil.AllocStatuses(jobID, ns) },
+			func(got []string) bool {
+				for _, status := range got {
+					if status == "running" {
+						return true
```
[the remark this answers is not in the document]
tgross: Yeah that assertion does seem bogus, right? I just traced it back and it looks like it was in the original "v0" implementation, ported to "v1", and now I'm copying the sin again porting it to "v2". Lemme see if I can fix it.

e2e/rescheduling/rescheduling_test.go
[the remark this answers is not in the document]
tgross: I'll add a helper specific to this suite though, and preserve the `NOMAD_TEST_SKIPCLEANUP` behavior with that.

e2e/rescheduling/rescheduling_test.go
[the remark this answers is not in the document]
tgross: This also discovered a bug in the jobspec, which is that it only had `attempts = 2`, so there was no guarantee we'd ever get a successful alloc, just a "more than likely". So this was a really flaky test. Fixed in cb81f51

e2e/rescheduling/rescheduling_test.go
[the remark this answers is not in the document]
tgross: Fixed in https://github.com/hashicorp/nomad/commit/cb81f5148a992cf73bfbcb71ba2a09f8572609e2

e2e/rescheduling/rescheduling_test.go
```diff
@@ -0,0 +1,513 @@
...
+
+		if os.Getenv("NOMAD_TEST_SKIPCLEANUP") == "1" {
+			return
+		}
+
+		e2eutil.StopJob(jobID, "-purge", "-detach")
+		_, err := e2eutil.Command("nomad", "system", "gc")
+		test.NoError(t, err)
+	}
+}
```
gulducat, line 37: super nit: since this is in the `rescheduling` package, that could be excluded from this func name

and differently nit-picky for helper design, I'd write this like:

```suggestion
func cleanupJob(t *testing.T, jobID string) {
	if os.Getenv("NOMAD_TEST_SKIPCLEANUP") == "1" {
		return
	}
	t.Helper()
	t.Cleanup(func() {
		e2eutil.StopJob(jobID, "-purge", "-detach")
		_, err := e2eutil.Command("nomad", "system", "gc")
		test.NoError(t, err)
	})
}
```

then test cases could just call

```go
cleanupJob(t, jobID)
```

gulducat (approved):
just one last test helper nit-pick!

e2e/rescheduling/rescheduling_test.go
tgross: Works for me, will do.

gulducat: it would return early from the `cleanupJob` func, prior to adding `t.Cleanup`, but the caller of `cleanupJob` would continue, eh?

tgross: :facepalm:

gulducat: having it inside of `t.Cleanup` is just fine too though -- 6 of one, half a dozen of the other :yum:

tgross: Fixed in 46a2c3e

PR #19108, opened 2023-11-16
e2e: use tf variable defaults

gulducat:
I noticed that despite #19012, a Windows instance was still being provisioned during E2E runs.

Turns out that's because our E2E repo has its own tfvars file(s), which means that these values are set in at least 3 different places:

* `variables.tf` defaults
* `terraform.tfvars` in this repo
* `github.tfvars` in e2e repo

To reduce the likelihood of this causing more confusion in the future, this removes most variables from tfvars, leaving only those that are actually different depending on context.


tgross (approved):
Good call, LGTM :+1:

Issue #18134, opened 2023-08-02
api: Stub() method causes panic when Allocation has a nil job


gabrielaccelbyte:
When using the Nomad SDK, I've encountered an issue where the Stub() method causes a panic if the Allocation object it's called on has a nil Job.

https://github.com/hashicorp/nomad/blob/2fd508d4f12fec139e495041b5ee1314eb1cd1dd/api/allocations.go#L322

https://github.com/hashicorp/nomad/blob/2fd508d4f12fec139e495041b5ee1314eb1cd1dd/api/allocations.go#L323

tgross:
Hi @gabrielaccelbyte yeah that should be handled more gracefully for sure but generally the expectation is that the `Allocation` object you'd call `Stub()` on came from the server where it's been canonicalized. Otherwise it doesn't have much use?

Certainly something to fix. I guess we can simply not set those fields and then the resulting stub is an invalid allocation stub, but that's ok because the original allocation was invalid as well.

gabrielaccelbyte:
Hi @tgross,

Thank you for your response. I understand the expectation for the Allocation object to come from the server where it's been canonicalized. However, in my specific use case, I'm receiving the Allocation object from the Nomad event stream, which doesn't always include the Job information.

PR #19115, opened 2023-11-17
Fix panic in Allocation.Stub() when Job is nil

codenoid:
Fixes: #18134


tgross (approved):
Thanks @codenoid! I've added a changelog entry and kicked off the CI run. Assuming that's green we'll get this merged.

Issue #16803, opened 2023-04-05
E2E: expiring tokens receive Permission Denied instead of ACL token expired


tgross:
The token expiration E2E test has been failing for a bit, most likely since my work in https://github.com/hashicorp/nomad/pull/15870 because we return `ErrPermissionDenied` on auth failures. I can pretty easily fix the error message so that we're returning the inner error from `Authenticate` when appropriate, but probably worth considering whether we really want to? cc @jrasell @pkazmierczak for discussion

```
=== FAIL: e2e/acl TestACL/TestACL_TokenExpiration (1.92s)
    acl_token_test.go:132: 
        	Error Trace:	/home/runner/actions-runner/_work/nomad-e2e/nomad-e2e/nomad/e2e/acl/acl_token_test.go:132
        	Error:      	Error "Unexpected response code: 403 (Permission denied)" does not contain "ACL token expired"
        	Test:       	TestACL/TestACL_TokenExpiration
```

PR #19119, opened 2023-11-17
E2E: fix expected error message from token expiration test

tgross:
In Nomad 1.5 we started masking the specific error returned from the authentication method and returned the "permission denied" error instead. Update the E2E test that covers token expiration to ensure we're asserting the correct error here.

(Also updates all the uses of `testify` to `shoenig/test`)

Fixes: https://github.com/hashicorp/nomad/issues/16803

---

Run against a real E2E cluster:

```
$ go test -v -count=1 ./e2e/acl -run TestACL/TestACL_TokenExpiration
=== RUN   TestACL
=== RUN   TestACL/TestACL_TokenExpiration
--- PASS: TestACL (1.39s)
    --- PASS: TestACL/TestACL_TokenExpiration (1.22s)
PASS
ok      github.com/hashicorp/nomad/e2e/acl      1.399s
```

### DIFF
--- a/e2e/acl/acl_token_test.go
+++ b/e2e/acl/acl_token_test.go
@@ -11,7 +11,7 @@ import (
 	"github.com/hashicorp/nomad/api"
 	"github.com/hashicorp/nomad/e2e/e2eutil"
 	"github.com/hashicorp/nomad/helper/uuid"
-	"github.com/stretchr/testify/require"
+	"github.com/shoenig/test/must"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


e2e/acl/acl_token_test.go
```diff
@@ -129,43 +129,43 @@ func testACLTokenExpiration(t *testing.T) {
 	defaultNSQueryMeta.AuthToken = tokenQuickExpiryCreateResp.SecretID
 
 	jobListResp, _, err = nomadClient.Jobs().List(&defaultNSQueryMeta)
-	require.ErrorContains(t, err, "ACL token expired")
-	require.Nil(t, jobListResp)
+	must.ErrorContains(t, err, "Permission denied")
```
gulducat, line 132: just flagging this as the functional test fix amongst the require->must conversions

gulducat (approved):
LGTM!

Issue #19112, opened 2023-11-16
E2E: dynamic node metadata workload ID test failure


tgross:
After fixing some bugs with the E2E test runs, we've got some failing tests to fixup. One of them is one of the workload ID tests.

```
=== FAIL: e2e/workload_id TestDynamicNodeMetadata/testDynamicNodeMetadata (31.07s)
    nodemeta_test.go:55: test config: job=node-meta-0499476e node=e170f7e1-80d1-3e21-3ea6-7238fc996ecd foo=foo-bdab5646 empty=empty-3a123614 unset=unset-a106f49c
    nodemeta_test.go:85: created policy node-meta-0499476e772ff818 for docker-nm
    nodemeta_test.go:85: created policy node-meta-0499476e4c92a282 for exec-nm
    nodemeta_test.go:109: job submitted, node metadata applied, waiting for metadata to be visible...
    nodemeta_test.go:124: checking node at index 3307
    nodemeta_test.go:124: checking node at index 3307
    nodemeta_test.go:124: checking node at index 3307
    nodemeta_test.go:124: checking node at index 3789
    nodemeta_test.go:135: waiting for job node-meta-0499476e to complete
    assert.go:14: 
        nodemeta_test.go:160: expected map to contain key
        ↪ key: empty-3a123614

=== FAIL: e2e/workload_id TestDynamicNodeMetadata (31.49s)
```

PR #19120, opened 2023-11-17
E2E: fix assertion in dynamic node metadata test

tgross:
In #18664 we change how null values worked with dynamic node metadata so that they were no longer returned if there wasn't also a static value for that key. The test assertion in E2E was not updated to match the new behavior.

Fixes: #19112


gulducat (approved):

lgfa29 (approved):
Oops...thanks! 😅

Issue #19109, opened 2023-11-16
E2E: `TestACL_Role` failure


tgross:
After fixing some bugs with the E2E test runs, we've got some failing tests to fixup. One of them is the ACL role test.

```
=== FAIL: e2e/acl TestACL/TestACL_Role (1.25s)
    acl_role_test.go:133: 
        	Error Trace:	/home/runner/actions-runner/_work/nomad-e2e/nomad-e2e/nomad/e2e/acl/acl_role_test.go:133
        	Error:      	Should be empty, but was [0xc0003cc000]
        	Test:       	TestACL/TestACL_Role
```

PR #19121, opened 2023-11-17
E2E: remove assertion from ACL role test

tgross:
The ACL role test asserts that the role has various permissions by listing jobs in namespaces. It never creates jobs, because we can make all the assertions we need by checking the error. But the test included an assertion that the namespace was empty. Usually this will be the case, but if the previous test case has not completed its GC (which is sync), then it's possible a stopped job will be in the namespace. Because this assertion is irrelevant for this test, remove it.

Fixes: https://github.com/hashicorp/nomad/issues/19109

### DIFF
--- a/e2e/acl/acl_role_test.go
+++ b/e2e/acl/acl_role_test.go
@@ -10,7 +10,7 @@ import (
 	"github.com/hashicorp/nomad/api"
 	"github.com/hashicorp/nomad/e2e/e2eutil"
 	"github.com/hashicorp/nomad/helper/uuid"
-	"github.com/stretchr/testify/require"
+	"github.com/shoenig/test/must"
 )
 
 // testACLRole tests basic functionality of ACL roles when used for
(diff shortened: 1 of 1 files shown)


e2e/acl/acl_role_test.go
```diff
@@ -98,12 +98,11 @@ func testACLRole(t *testing.T) {
 	customNSQueryMeta := api.QueryOptions{Namespace: ns.Name, AuthToken: aclTokenCreateResp.SecretID}
 	defaultNSQueryMeta := api.QueryOptions{Namespace: "default", AuthToken: aclTokenCreateResp.SecretID}
 
-	jobListResp, _, err := nomadClient.Jobs().List(&customNSQueryMeta)
-	require.NoError(t, err)
-	require.Empty(t, jobListResp)
```
gulducat, line 103: this is the offending assertion, yeah?

e2e/acl/acl_role_test.go
```diff
@@ -122,35 +121,34 @@ func testACLRole(t *testing.T) {
...
-	require.Len(t, aclRoleUpdateResp.Policies, 2)
+	must.NoError(t, err)
+	must.Eq(t, aclRoleCreateResp.ID, aclRoleUpdateResp.ID)
+	must.Len(t, 2, aclRoleUpdateResp.Policies)
 
 	// Try listing the jobs in the default namespace again to ensure we now
 	// have permission due to the updated role.
-	jobListResp, _, err = nomadClient.Jobs().List(&defaultNSQueryMeta)
-	require.NoError(t, err)
-	require.Empty(t, jobListResp)
```
gulducat, line 133: and here?

gulducat (approved):

e2e/acl/acl_role_test.go
tgross: Yes!

e2e/acl/acl_role_test.go
tgross: Yes!

PR #19133, opened 2023-11-20
client: flatten multierror messages in the consul hook

pkazmierczak:


client/allocrunner/consul_hook.go
```diff
@@ -97,6 +97,7 @@ func (h *consulHook) Prerun() error {
 
 	err := mErr.ErrorOrNil()
 	if err != nil {
+		multierror.Flatten(err)
```
lgfa29, line 100: I don't remember using this method before, but it doesn't seem to be doing any for us here?

Here's an error with 1.7.0-beta.2:

image

```
    2023-11-21T20:30:54.242-0500 [ERROR] client.alloc_runner: prerun failed: alloc_id=7a5af945-ce5a-d6f6-859c-ef013cf44420
  error=
  | pre-run hook "consul" failed: 1 error occurred:
  | \t* 1 error occurred:
  | \t* 1 error occurred:
  | \t* failed to authenticate with consul for identity consul-service_mongo-db: Unexpected response code: 403 (ACL not found: auth method "nomad-services" not found)
  |
  |
  |
  |
  |
```

And the error shows up the same way in this branch:

image

```
    2023-11-21T20:32:59.484-0500 [ERROR] client.alloc_runner: prerun failed: alloc_id=7d4984f7-3deb-3dc3-5850-9669eaf99caf
  error=
  | pre-run hook "consul" failed: 1 error occurred:
  | \t* 1 error occurred:
  | \t* 1 error occurred:
  | \t* failed to authenticate with consul for identity consul-service_mongo-db: Unexpected response code: 403 (ACL not found: auth method "nomad-services" not found)
  |
  |
  |
  |
  |
```

I haven't looked too much into the code yet, but it seems like the error multierror topology is something like this:

```mermaid
graph TB
  RootError((Root Error)) --> GroupServicesError
  RootError --> Task

  GroupServicesError((Group Services Errors)) --> E1((Error 1))
  GroupServicesError --> E2((Error 2))
  GroupServicesError --> E3((Error 3))

  Task((Task Errors)) --> TaskServicesError
  Task --> TaskErrors

  TaskServicesError((Tasks Services Errors)) --> E4((Error 1))
  TaskServicesError --> E5((Error 2))
  TaskServicesError --> E6((Error 3))

  TaskErrors((Tasks Errors)) --> E7((Error 1))
  TaskErrors --> E8((Error 2))
  TaskErrors --> E9((Error 3))
```

So instead of flattening we probably want to accumulate errors at the root with the proper context. For example, error above would be just:

```
    2023-11-21T20:32:59.484-0500 [ERROR] client.alloc_runner: prerun failed: alloc_id=7d4984f7-3deb-3dc3-5850-9669eaf99caf
  error=
  | pre-run hook "consul" failed: 1 error occurred:
  | \t* failed to authenticate with consul for identity consul-service_mongo-db: Unexpected response code: 403 (ACL not found: auth method "nomad-services" not found)
```

I think `errors.Unwrap()` may skip a level, but I haven't fully gotten used to how this works 😅 

Otherwise you could do type assertion at the caller.

Alternatively, since the `prepare*` methods are internal, they could return a `multierror.Error` directly and the caller merges all internal errors into one.

pkazmierczak: oof, apologies for not checking this thoroughly before :/ 

4767ee0 changes the implementation using `multierror.Error` returns for internal methods and type assertion for the returned error of consul client. Results in a much neater error message:

```
    2023-11-23T17:37:46.487+0100 [ERROR] client.alloc_runner: prerun failed: alloc_id=1e22a6a9-4b49-6fdd-f2c8-0d65b5685d55
  error=
  | pre-run hook "consul" failed: 1 error occurred:
  | \t* failed to authenticate with consul for identity consul_default: Unexpected response code: 403 (ACL not found: auth method "nomad-tasks" not found)
  |
```

client/allocrunner/consul_hook.go
```diff
@@ -196,25 +196,27 @@ func (h *consulHook) prepareConsulTokensForServices(services []*structs.Service,
...
 	}
 
 	// get consul acl tokens
 	t, err := client.DeriveTokenWithJWT(req)
 	if err != nil {
-		return err
+		// DeriveTokenWithJWT actually returns a multierror, so we assert the
+		// type to make error formatting from the hook level more readable.
+		return err.(*multierror.Error)
 	}
```
lgfa29, line 220: Oh, there are more layers 😅 

This gives me some pause because it makes an implicit assumption from the caller about an implementation detail. It's also tricky because `DeriveTokenWithJWT` comes from an interface, where another implementation may not return a multierror (a test mock for example).

Looking at the code again, I think the core problem stems from the use of `mErr.Errors = append(mErr.Errors, err)` everywhere. Appending an error directly like that creates a new level if the error being appended is a multierror since the top-level error is the ` 1 error occurred:` message.

What we usually want is to accumulate errors. For that you should use `mErr = multierror.Append(mErr, err)`.

https://github.com/hashicorp/nomad/commit/16d6cfaad00d43ced253a3862a2f46c65fb985cd make this change and I think works as expected

image

I also noticed that the error from `h.revokeTokens()` was being ignored. Not sure if it was done on purpose but I bundled it there with the results.

How does that look to you?

client/allocrunner/consul_hook.go
```diff
@@ -107,20 +107,20 @@ func (h *consulHook) Prerun() error {
 	return nil
 }
 
-func (h *consulHook) prepareConsulTokensForTask(task *structs.Task, tg *structs.TaskGroup, tokens map[string]map[string]*consulapi.ACLToken) error {
+func (h *consulHook) prepareConsulTokensForTask(task *structs.Task, tg *structs.TaskGroup, tokens map[string]map[string]*consulapi.ACLToken) *multierror.Error {
 	if task == nil {
 		// programming error
-		return fmt.Errorf("cannot prepare consul tokens, no task specified")
+		return &multierror.Error{Errors: []error{fmt.Errorf("cannot prepare consul tokens, no task specified")}}
```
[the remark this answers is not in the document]
pkazmierczak: yeah I was on the fence with this. I figured appending a singleton and returning it was just... weird, but so is creating a singleton and returning it? ¯\_(ツ)_/¯

client/allocrunner/consul_hook.go
pkazmierczak: > What we usually want is to accumulate errors. For that you should use mErr = multierror.Append(mErr, err).

ooooh so _that's_ the mistake I made. I'll merge your branch into this.

client/allocrunner/consul_hook.go
```diff
@@ -197,24 +198,26 @@ func (h *consulHook) prepareConsulTokensForServices(services []*structs.Service,
...
+		return &multierror.Error{Errors: []error{err}}
 	}
 
 	// get consul acl tokens
 	t, err := client.DeriveTokenWithJWT(req)
 	if err != nil {
-		return err
+		// DeriveTokenWithJWT actually returns a multierror, so we assert the
+		// type to make error formatting from the hook level more readable.
+		return err.(*multierror.Error)
```
lgfa29, line 220: Not sure what happened, but this diff doesn't include all my changes from https://github.com/hashicorp/nomad/commit/16d6cfaad00d43ced253a3862a2f46c65fb985cd. 

This is particular is the one that I'm worried about.

pkazmierczak: ugh, must've been some git merge fail, apologies. I fix.

pkazmierczak: actually it seems that 16d6cfa and your branch in general have been merged completely, they simply don't have the changes you want. I'll put them in manually.

pkazmierczak: I think [1da57ed](https://github.com/hashicorp/nomad/pull/19133/commits/1da57edb68491ac3849fee1628b502c1a658fe3e) provides a good fix.

Issue #19127, opened 2023-11-17
Workload Identity - Consul integration creates duplicate per-task tokens


krarey:
### Nomad version
Nomad v1.7.0-beta.2
BuildDate 2023-11-15T22:47:41Z
Revision ea3f6cc8798df782cbf82b1c2f55e10b7ca4a419

### Operating system and Environment details
Nomad 1.7.0-beta.2 arm64
Consul 1.17.0 Community Edition arm64
Nomad Workload Identity integration configured per recommended defaults

### Issue
When Nomad workload identity is used to authenticate with Consul, the per-task token used for template rendering is generated twice. Only one of the two tokens is later revoked when the associated allocation is stopped, leading to rapid pollution of the Consul ACL list with unused, infinitely-lived tokens.

pkazmierczak:
Thanks for reporting this, @krarey!

PR #19137, opened 2023-11-20
client: make sure consul_hook does not perform double requests for tasks

pkazmierczak:
Fixes https://github.com/hashicorp/nomad/issues/19127

### DIFF
--- a/client/allocrunner/consul_hook.go
+++ b/client/allocrunner/consul_hook.go
@@ -129,8 +129,10 @@ func (h *consulHook) prepareConsulTokensForTask(task *structs.Task, tg *structs.
 		ti := *task.IdentityHandle(i)
 		jwt, err := h.widmgr.Get(ti)
 		if err != nil {
-			h.logger.Error("error getting signed identity", "error", err)
-			mErr.Errors = append(mErr.Errors, err)
+			mErr.Errors = append(mErr.Errors, fmt.Errorf(
+				"error getting signed identity for task %s: %v",
+				task.Name, err,
+			))
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


client/allocrunner/consul_hook.go
```diff
@@ -175,8 +169,19 @@ func (h *consulHook) prepareConsulTokensForServices(services []*structs.Service,
...
 		}
 
-		req, err := h.prepareConsulClientReq(
-			*service.IdentityHandle(), consulConfig.ServiceIdentityAuthMethod)
+		req := map[string]consul.JWTLoginRequest{}
+		identity := *service.IdentityHandle()
+		jwt, err := h.widmgr.Get(identity)
+		if err != nil {
+			h.logger.Error("error getting signed identity", "error", err)
+			mErr.Errors = append(mErr.Errors, err)
```
lgfa29, line 177:
```suggestion
			mErr.Errors = append(mErr.Errors, fmt.Errorf("error getting signed identity for service %s: %v", service.Name, err)
```

And similar for `prepareConsulTokensForTask` above. Some reasons for this changed:

1. having the same error context for two different operations can make debugging harder because you may not be able to tell where the `error getting signed identity` is coming from (is it a task or a service?)
2. hook errors are already logged by the alloc runner, so also logging deep here can cause the same error to be logged twice.
https://github.com/hashicorp/nomad/blob/ff928a804590611111763632388161dc711adf88/client/allocrunner/alloc_runner.go#L361-L362

lgfa29 (approved):
Nice fix!

jrasell (approved):
LGTM!

PR #19139, opened 2023-11-20
[ui] Fix: Purging jobs via the UI when namespaces are present

### LINKED ISSUE #18677: Purging from Nomad UI does not work because parameters are passed incorrectly
### Nomad version
Nomad v1.6.2

Fedora 29

### Issue
Nomad UI sends wrong request when purging.

### Reproduction steps
Create a job. Stop it. Click to purge it.

#### Expected Result
Should be purged.

#### Actual Result
Not purged.
When clicking "purge job" from the ui I see in chrome network debug view that it sends the following query:

     <url>/v1/job/zentao?namespace=dev?purge=true

It's wrong, it should be `namespace=dev&purge=true`.

This is caused by this: https://github.com/hashicorp/nomad/blob/e3c8700ded891702b7f94636033109efd8f71c3a/ui/app/adapters/job.js#L47 .

philrenaud:
Previously, we'd just concat `?purge=true` without checking to see if other params were present. With this PR, we use our `addToPath` helper to append either via `?` or `&`, depending whether other params like namespace are present.

Resolves #18677

### DIFF
--- a/.changelog/19139.txt
+++ b/.changelog/19139.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fixed an issue where purging a job with a namespace did not process correctly
+```
--- a/ui/app/adapters/job.js
+++ b/ui/app/adapters/job.js
@@ -52,7 +52,12 @@ export default class JobAdapter extends WatchableNamespaceIDs {
   }
 
   purge(job) {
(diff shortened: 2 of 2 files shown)


pkazmierczak (approved):
LGTM!

PR #19144, opened 2023-11-21
docs: additional note about JWKS endpoints and CA certs

pkazmierczak:


apollo13:
This doc improvement makes no sense to me, am I missing something? If `tls.verify_https_client = true` is set then you'd need a way to specify a client cert. `JWKSCACert` only defines the root certificates that the client uses to validate the server cert of nomad, it doesn't specify a client cert for the consul client (or I am miss-reading the docs)

tgross:
> This doc improvement makes no sense to me, am I missing something?

Hey @apollo13! No, I don't think you are. We're having an internal discussion on how exactly we want to address this, but just wanted to ack your comment so you don't think we're ignoring you. :grinning:

pkazmierczak:
@apollo13 You're right, the note was my mistake due to misunderstanding how mTLS between Nomad and Consul works. I made a correction: https://github.com/hashicorp/nomad/pull/19207

apollo13:
Thank you, the new paragraph looks way better! Btw, is it necessary to call out mTLS wherever possible? I honestly cannot believe it is recommended practice in this case. Anything which requires user access like the nomad UI is a PITA with client certs. And having a client cert on a load balancer in front which then again accepts requests from literally anyone doesn't seem like a sensible suggestion either. Sure, a good security hygiene is worth much, but as soon as endusers with their browsers are involved mTLS is a clusterfuck imo.

tgross:
> is it necessary to call out mTLS wherever possible?

Ah, the wording of that paragraph might be improved a little bit. The relevant sentence reads:

> If you want to use mTLS, which is the recommended practice, you need to terminate TLS before the requests reach Nomad JWKS endpoint. 

Using mTLS is _required_ in Nomad for node-to-node RPC, as we outline pretty clearly in the [security requirements](https://developer.hashicorp.com/nomad/docs/concepts/security#requirements) docs. But you can terminate safely for the API (and the UI, which uses that same port), because of ACLs. So this sentence is saying "if you want to use mTLS (in general), you need to terminate TLS for JWKS". I can take another crack at maybe making that more clear.

apollo13:
@tgross Fair enough. Looking through the tls docs ( https://developer.hashicorp.com/nomad/docs/configuration/tls ). it is not exactly clear to me if it is possible to configure the http endpoint without mTLS and rpc with.

ie what does:
```
tls {
  http = false
  rpc  = true

  ca_file   = "/etc/certs/ca.crt"
  cert_file = "/etc/certs/nomad.crt"
  key_file  = "/etc/certs/nomad.key"
}
```
do? Does it disable TLS completely on the http endpoint or solely mTLS. The docs say:

> Specifies if mTLS should be enabled on the HTTP endpoints on the Nomad agent, including the API.

If I had to guess I think it probably means TLS. Also wouldn't it make sense to be able to set `[verify_https_client](https://developer.hashicorp.com/nomad/docs/configuration/tls#verify_https_client)` differently for http and the rpc endpoints?

tgross:
Ok, thanks for that feedback @apollo13. It's always good to hear where people are getting hung up on this. I'm going to put my head together with the team and see if we can make this much more clear (and bring in our product security person to make sure we're not giving people bad advice). 

But in the meantime the tl;dr on how this works is:

* The difference between mTLS and TLS isn't in the resulting transport but in authenticating the handshake.
* `tls.http=true` means that the HTTP server on the agent is served with TLS (i.e. the agent performs the server side of a TLS handshake)
* `tls.verify_https_client=true` (the default) means that the agent also requires that the client response with a client cert, and verifies it. With ACLs, I would argue this is belt-and-suspenders and probably unnecessary, which is why it's a reasonable option to deal with problems like JWKS.
* `tls.verify_https_client=false` means the agent doesn't require a client cert and will ignore one if presented by the client.
* `tls.rpc=true` establishes TLS between agents, but doesn't allow for opting out of mTLS because server-to-server communication has no ACLs so we need mTLS for authentication.

apollo13:
> The difference between mTLS and TLS isn't in the resulting transport but in authenticating the handshake.

Yes, I knew that. I mainly got confused because `tls.http` mentions mTLS when it actually completely enables/disables (m)TLS

> With ACLs, I would argue this is belt-and-suspenders and probably unnecessary, which is why it's a reasonable option to deal with problems like JWKS.

I agree here and I think the relevant docs (https://developer.hashicorp.com/nomad/tutorials/transport-security/security-enable-tls) should maybe get updated and have `tls.verify_https_client = False` since the JWKS workflow is becoming a rather common one. Currently people are probably afraid to turn it off since so many things recommend it without a clear explanation of what it does and when you don't need it.

> tls.rpc=true establishes TLS between agents, but doesn't allow for opting out of mTLS because server-to-server communication has no ACLs so we need mTLS for authentication.

Ah yes, I forgot that.

Thank for looking into this, doc improvements around the security aspects would certainly be helpful.

PR #19145, opened 2023-11-21
action: add job action name validation

jrasell:


philrenaud (approved):
Looks good. All the test actions on the js front are already single-worded (faker.hacker.verb()) so should match. Thanks!

lgfa29 (approved):

lgfa29:
Clicked the wrong place, but I meant to add a comment asking if we should note the name requirements in the docs.

jrasell:
> Clicked the wrong place, but I meant to add a comment asking if we should note the name requirements in the docs.

Yes, this should be included within the docs when the PR for these is raised.

PR #19153, opened 2023-11-22
[ui] example job with actions

philrenaud:
An example job in the UI templates to show off actions and how they work.

image
image
image

### DIFF
--- a/ui/app/utils/default-job-templates.js
+++ b/ui/app/utils/default-job-templates.js
@@ -7,6 +7,7 @@ import helloWorld from './default_jobs/hello-world';
 import parameterized from './default_jobs/parameterized';
 import serviceDiscovery from './default_jobs/service-discovery';
 import variables from './default_jobs/variables';
+import actions from './default_jobs/actions';
 
 export default [
   {
@@ -19,6 +20,19 @@ export default [
       {
         key: 'description',
         value: 'A simple job that runs a single task on a single node',
+      },
+    ],
+  },
+  {
+    id: 'nomad/job-templates/default/actions',
+    keyValues: [
+      {
+        key: 'template',
+        value: actions,
+      },
+      {
+        key: 'description',
+        value: 'A raw exec job that shows off Nomad Actions',
       },
     ],
   },
--- a/ui/app/utils/default-job-templates.js
+++ b/ui/app/utils/default-job-templates.js
@@ -32,7 +32,8 @@ export default [
       },
       {
         key: 'description',
-        value: 'A raw exec job that shows off Nomad Actions',
+        value:
+          'Nomad Actions let job authors describe commands that can be run in one click from the UI or one command from the CLI. This example job shows how to use them to simulate development on a Redis instance.',
       },
     ],
   },
--- a/ui/app/utils/default_jobs/actions.js
+++ b/ui/app/utils/default_jobs/actions.js
@@ -3,128 +3,129 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
-export default `job "job-with-actions" {
-  # Specifies the datacenter where this job should be run
-  # This can be omitted and it will default to ["*"]
+export default `job "redis-actions" {
   datacenters = ["*"]
 
-  # Run the job only on Linux or MacOS.
-  constraint {
-    attribute = "\${attr.kernel.name}"
-    operator  = "set_contains_any"
-    value     = "darwin,linux"
-  }
+  group "cache" {
+    count = 1
+    network {
+      port "db" {}
+    }
 
-  group "my_group" {
-    # Specifies the number of instances of this group that should be running.
-    # Use this to scale or parallelize your job.
-    count = 3
+    task "redis" {
+      driver = "docker"
 
-    task "sleepy" {
-      driver = "raw_exec"
-
-      # The "command" stanza specifies the command to run.
-      # This command will run a .sh file generated from the template stanza below,
-      # which will sleep for 2 seconds and repeat until the job is stopped.
       config {
-        command = "\${NOMAD_TASK_DIR}/sleepy.sh"
+        image = "redis:3.2"
+        ports = ["db"]
+        command = "/bin/sh"
+        args = ["-c", "redis-server --port \${NOMAD_PORT_db} & /local/db_log.sh"]
       }
 
       resources {
-        memory = 16
-        cpu = 16
+        cpu = 500
+        memory = 256
+      }
+
+      service {
+        name = "redis-service"
+        port = "db"
+        provider = "nomad"
+
+        check {
+          name     = "alive"
+          type     = "tcp"
+          port     = "db"
+          interval = "10s"
+          timeout  = "2s"
+        }
       }
 
       template {
-        data = <<EOH
-#!/bin/bash
-SLEEP_SECS=2
-interruptable_sleep() { for i in $(seq 1 $((2*\${1}))); do sleep .5; done; }
-sigint() { echo "$(date) - SIGTERM received; Ending."; exit 0; }
-trap 'sigint'  INT
-echo "$(date) - Starting. SLEEP_SECS=\${SLEEP_SECS}"
-while true; do echo "$(date) - Sleeping for \${SLEEP_SECS} seconds."; interruptable_sleep \${SLEEP_SECS}; done
-EOH
-        destination = "local/sleepy.sh"
+        data = <<EOF
+          #!/bin/sh
+          while true; do
+            echo "$(date): Current DB Size: $(redis-cli -p \${NOMAD_PORT_db} DBSIZE)"
+            sleep 3
+          done
+EOF
+        destination = "local/db_log.sh"
+        perms = "0755"
       }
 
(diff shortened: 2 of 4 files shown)


ui/app/utils/default-job-templates.js
```diff
@@ -22,6 +23,19 @@ export default [
...
+  {
+    id: 'nomad/job-templates/default/actions',
+    keyValues: [
+      {
+        key: 'template',
+        value: actions,
+      },
+      {
+        key: 'description',
+        value: 'A raw exec job that shows off Nomad Actions',
```
DingoEatingFuzz, line 35: Not a copy writer, but I think this teaser text can be improved to describe what actions are and what they do.

Maybe

> Nomad Actions let job authors describe commands that can be run in one click from the UI or one command from the CLI. This example job has actions for getting job info, alloc info, and the Nomad changelog.

philrenaud: This is a big improvement, thanks — I rushed this description.

ui/app/utils/default_jobs/actions.js
```diff
@@ -0,0 +1,130 @@
...
+    value     = "darwin,linux"
+  }
+
+  group "my_group" {
+    # Specifies the number of instances of this group that should be running.
+    # Use this to scale or parallelize your job.
+    count = 3
+
+    task "sleepy" {
+      driver = "raw_exec"
```
lgfa29, line 24: The `raw_exec` driver is disabled by default on non-dev clusters. Should we use `docker` instead?

The actions also assume things about the node, like having bash, awk, and curl installed, which may not be true.

ui/app/utils/default_jobs/actions.js
```diff
@@ -0,0 +1,130 @@
...
+        destination = "local/sleepy.sh"
+      }
+
+      # Action blocks can be used to run arbitrary commands on an allocation,
+      # such as to collect logs, clear caches, migrate databases, or debug issues.
+
+      # This action will run a simple echo command every second.
+      # Where the other actions run and exit upon completion, this action will run
+      # until it is manually stopped with an escape character or socket closure.
+      action "echo time" {
```
lgfa29, line 57:
```suggestion
      action "echo-time" {
```

https://github.com/hashicorp/nomad/pull/19145 makes this name invalid.

ui/app/utils/default_jobs/actions.js
```diff
@@ -0,0 +1,130 @@
...
+EOH
+        destination = "local/sleepy.sh"
+      }
+
+      # Action blocks can be used to run arbitrary commands on an allocation,
+      # such as to collect logs, clear caches, migrate databases, or debug issues.
+
+      # This action will run a simple echo command every second.
+      # Where the other actions run and exit upon completion, this action will run
+      # until it is manually stopped with an escape character or socket closure.
```
lgfa29, line 56: I don't remember seeing it the UI (probably missed it), but is there a "stop action" button?

ui/app/utils/default_jobs/actions.js
```diff
@@ -0,0 +1,130 @@
...
+          <<EOT
+          nomad alloc status \${NOMAD_ALLOC_ID}
+          EOT
+        ]
+      }
+
+      # This final action will fetch the latest Nomad changelog and parse it to get the latest 3 versions
+      # and the number of items under each section using curl and awk. This is meant to demonstrate
+      # how actions can be used to fetch information or perform other tasks that are not directly related
+      # to the running task.
```
lgfa29, line 86:
```suggestion
      # how actions can be used to fetch information or perform other activities that are not directly related
      # to the running task.
```

Can't think of a better word, but `task` is being used for two different things in this sentence.

lgfa29 (commented):
I'm a bit on the fence about the last action. While cool, the awk script is quite complex and I think a bit distracting for the purpose of the job.

If you want to curl something Nomad related you could add some ASCII art to https://github.com/hashicorp/nomad/tree/main/website/public/data

ui/app/utils/default_jobs/actions.js
philrenaud: There is (it sends the escape char behind the scenes); you're right that I should probably mention it though!

ui/app/utils/default_jobs/actions.js
philrenaud: Hm — I was secretly hoping that this could be our first "If you don't have docker installed you can still play around with Nomad" example job, but you're right. Actions with sh/curl/awk is surely not the one to try that out on.

lgfa29: Unfortunately that would be very hard to achieve and it's 95% of the motivation behind Docker being created in the first place 😅 

If you want a non-Docker job example I think it should be marked as such with all the important caveats ("you must have X, Y, and Z installed") listed before users run it.

ui/app/utils/default_jobs/actions.js
```diff
@@ -0,0 +1,130 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+export default `job "job-with-actions" {
```
lgfa29, line 6: Random thought: does ember-cli have a way to load random files into variables? I think webpack can do it, but I'm not sure if it is exposed.

If that's possible, then having these examples in their own files could help keep them organized, formatted, and accessible by random `nomad run` commands.

philrenaud: Yep, looked briefly into it and there's no reason why we shouldn't be able to do it other than time (and there's maybe an argument to be made that a web UI user and a `nomad job init` CLI user might want to see different "get started" jobspecs)

ui/app/utils/default_jobs/actions.js
```diff
@@ -0,0 +1,131 @@
...
+
+      resources {
+        cpu = 500
+        memory = 256
+      }
+
+      service {
+        name = "redis-service"
+        port = "db"
+        provider = "nomad"
```
lgfa29, line 33:
```suggestion
      config {
        image   = "redis:3.2"
        ports   = ["db"]
        command = "/bin/sh"
        args    = ["-c", "redis-server --port \${NOMAD_PORT_db} & /local/db_log.sh"]
      }

      resources {
        cpu    = 500
        memory = 256
      }

      service {
        name     = "redis-service"
        port     = "db"
        provider = "nomad"
```

image

ui/app/utils/default_jobs/actions.js
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    task "redis" {
+      driver = "docker"
+
+      config {
+        image = "redis:3.2"
+        ports = ["db"]
+        command = "/bin/sh"
+        args = ["-c", "redis-server --port \${NOMAD_PORT_db} & /local/db_log.sh"]
```
lgfa29, line 22: Like above, we can probably use port mapping to keep this a bit cleaner and avoid the extra args, but if you want to use dynamic ports something like this would be a bit easier to understand?

```hcl
job "example" {
  group "cache" {
    network {
      port "db" {}
    }

    task "redis" {
      driver = "docker"

      config {
        image   = "redis:7"
        ports   = ["db"]
        args    = ["/etc/redis/redis.conf"]
        volumes = ["local/conf:/etc/redis"]
      }

      template {
        data        = <<EOF
port {{env "NOMAD_PORT_db"}}
EOF
        destination = "local/conf/redis.conf"
      }
    }
  }
}
```

Then the `redis-cli -p \${NOMAD_PORT_db} DBSIZE` command in `db_log.sh` could be another action?

ui/app/utils/default_jobs/actions.js
```diff
@@ -0,0 +1,131 @@
...
+          fi;
+EOF
+        ]
+      }
+
+
+    }
+  }
+
+}
```
lgfa29, line 130:
```suggestion
      }
    }
  }
}
```

ui/app/utils/default_jobs/actions.js
```diff
@@ -0,0 +1,131 @@
...
+        image = "redis:3.2"
+        ports = ["db"]
+        command = "/bin/sh"
+        args = ["-c", "redis-server --port \${NOMAD_PORT_db} & /local/db_log.sh"]
+      }
+
+      resources {
+        cpu = 500
+        memory = 256
+      }
```
lgfa29, line 28: If possible, I would try to tighten up these values as it can result in placement failures in small deployments with other jobs running. I think redis can run with fewer resources.

lgfa29 (approved):
Minor nit-picks but LGTM!

ui/app/utils/default_jobs/actions.js
philrenaud: I realize it's not practical or desired for a real-world redis server on Nomad, but it seemed cool to illustrate db size output live when you make edits via action.

In the (pending) [actions tutorial](https://tutorials-git-1814-hashicorp.vercel.app/nomad/tutorials/job-specifications/job-spec-actions) ([PR](https://github.com/hashicorp/tutorials/pull/1814)), I hoped this would be something easily demonstrable to a user.

PR #19155, opened 2023-11-23
cli: skip allocs with replacements on job restart

lgfa29:
The `nomad job restart` command should skip allocations that already have replacements. Restarting an allocation with a replacement is a no-op because the allocation status is terminal and the command's replacement monitor returns immediatelly.

But by not skipping them, the effective batch size is computed incorrectly.


pkazmierczak (approved):
LGTM!

Issue #19128, opened 2023-11-17
[Feature Request] Workload Identity - Avoid requiring multiple Consul auth methods


krarey:
### Proposal
As of Nomad v1.7.0-beta.2, the workload identity integration with Consul suggests creating two distinct JWT auth methods within Consul, one for service tokens and one for task/template tokens. These auth methods have identical configuration, varying only in the associated binding rule.

With an additional claim (or other predictable pattern within the existing claims) on the workload identity JWT indicating its disposition, this could be avoided, and a binding rule selector could be used in stead to determine which rule takes precedence at auth time.

PR #19169, opened 2023-11-24
WI: use single auth method for Consul by default

pkazmierczak:
This code simplifies the default setup of Nomad workloads WI-based authentication for Consul by using a single auth method with 2 binding rules. Users can still specify separate auth methods for services and tasks. 

Resolves https://github.com/hashicorp/nomad/issues/19128


command/setup_consul.go
```diff
@@ -254,66 +251,37 @@ tasks %q.
...
 
 	s.Ui.Output(`
 Consul uses binding rules to map claims between Nomad's JWTs to Consul service
-identities and ACL roles, so we need to create a binding rule for each of the
-auth methods above.
+identities and ACL roles, so we need to create a two binding rules for the auth
+method we created above: one for services, and one for tasks.
 `)
 
 	if s.bindingRuleExists(servicesBindingRule) {
```
lgfa29, line 280: We should update `bindingRuleExists()` to also compare the `Selector`.

lgfa29 (approved):
Leaving a placeholder LGTM in case we do go with this route.

I think the only thing missing is updating the default values in the docs for `consul` config.
https://github.com/hashicorp/nomad/blob/cfbb2e8923d254bc940e1b01c3e772769600fefe/website/content/docs/configuration/consul.mdx#L134-L158

website/content/docs/integrations/consul-integration.mdx
```diff
@@ -167,55 +189,6 @@ service_prefix "" {
 }
 ```
 
-Using the same configuration file shown above, create the Consul Auth Method to
```
tgross, line 170: I'd probably add a reference back to the auth method and binding rule we created earlier just to make it extra clear how this policy is getting associated with jobs. Something like:

> The binding rule you created above for Nomad namespaces will map workloads to the role with this policy.

Alternately, you could move the binding rule creation here so that it's more directly associated with the role.

tgross (approved):
LGTM!

PR #19173, opened 2023-11-24
docs: clarify difference between /client and /node

lgfa29:
@sofixa reported that was quite hard to find the API endpoints for dynamic node metadata. I think the reason is two-fold:

1. since the feature is called dynamic **node** metadata, and the command is placed under `nomad node meta`, users expect to find the endpoints under `/node`, but it's actually under `/client`.
2. search engines have a really hard time matching the `/client` page with a query such as "nomad dynamic node metadata". From my quick experiment, Google never found the page, Bing had it at the bottom of first page, and DDG had it mid-page.

This PR applies the following changes to try to fix these problems.

Clarify the difference between the `/client` and `/node` endpoints and link from one to the other to help users discover the endpoint they are looking for.

Also update the  `/client` page description and dynamic nod metadata section headers to help the page be more discoverable by search engines.

### DIFF
--- a/website/content/api-docs/client.mdx
+++ b/website/content/api-docs/client.mdx
@@ -2,25 +2,29 @@
 layout: api
 page_title: Client - HTTP API
 description: |-
-  The /client endpoints are used to access client information and inspect
-  allocations running on a particular client.
+  The /client endpoints are used to access client information, read and write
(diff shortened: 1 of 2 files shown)


jrasell (approved):
LGTM!

We probably want this backported to the live site and previous versions.

PR #19176, opened 2023-11-26
Mark CGroups as off when missing essential controllers

DavidVentura:
On Linux systems that have been built without all necessary CGroup controllers enabled, you may get a non-descriptive error when initializing `cgroupslib`:

```
[ERROR] client.proclib.cg2: failed to create nomad cgroup: error="write /sys/fs/cgroup/cgroup.subtree_control: invalid argument"
```

A similarly-opaque error is logged on every run of `cpuparts_hook`, but I could not find a simple way of clarifying the error logs.

This PR checks whether all expected controllers are available, and will log an error if they are not.
On top of that, the controllers will be enabled one at a time, giving a more clear error when one of them is not supported.

You can validate this PR by appending `cgroup_disable=cpuset` to your kernel's command line arguments (`/etc/default/grub` or similar)

### DIFF
--- a/client/lib/cgroupslib/init.go
+++ b/client/lib/cgroupslib/init.go
@@ -152,7 +152,7 @@ func Init(log hclog.Logger, cores string) {
 		controllersRootPath := filepath.Join(root, controllersFile)
 		content, err := os.ReadFile(controllersRootPath)
 		if err != nil {
-			log.Info("failed to read cgroups controller file", "path", controllersRootPath, "error", err)
+			log.Error("failed to read cgroups controller file", "path", controllersRootPath, "error", err)
 		} else {
 			rootSubtreeControllers := strings.Split(strings.TrimSpace(string(content)), " ")
 
--- a/client/lib/cgroupslib/init.go
+++ b/client/lib/cgroupslib/init.go
@@ -6,9 +6,6 @@
 package cgroupslib
 
 import (
-	"fmt"
-	"strings"
-	"slices"
 	"bytes"
 	"os"
 	"path/filepath"
@@ -137,7 +134,7 @@ func Init(log hclog.Logger, cores string) {
 
 	case CG2:
 		// the cgroup controllers we need to activate at the root and on the nomad slice
-		controllers := []string{"cpuset", "cpu", "io", "memory", "pids"}
+		const activation = "+cpuset +cpu +io +memory +pids"
 
 		// the name of the cgroup subtree interface file
 		const subtreeFile = "cgroup.subtree_control"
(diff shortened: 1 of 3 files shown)


client/lib/cgroupslib/init.go
```diff
@@ -134,18 +137,37 @@ func Init(log hclog.Logger, cores string) {
...
 		// configuring root cgroup (/sys/fs/cgroup)
 		//
 
-		if err := writeCG(activation, subtreeFile); err != nil {
-			log.Error("failed to create nomad cgroup", "error", err)
-			return
+		controllersRootPath := filepath.Join(root, controllersFile)
+		content, err := os.ReadFile(controllersRootPath)
+		if err != nil {
+			log.Info("failed to read cgroups controller file", "path", controllersRootPath, "error", err)
```
tgross, line 155:
```suggestion
			log.Error("failed to read cgroups controller file", "path", controllersRootPath, "error", err)
```

DavidVentura:
I wasn't entirely convinced on whether the writes should be partially applied or not; the behavior for systems with missing support for a controller changes:

When applying a single write for all controllers, (`echo "cpuset cpu pids" > /sys/fs..`), if any controller is missing, none of them will be applied.
When applying a write per controller (`echo "+cpuset" > /sys/fs/..; echo "+cpu" > /sys/fs..`), only the missing controllers will not be applied.

I can also attempt to apply them all, but I'm not sure what that'd achieve, as anything dependent on the actual controller is bound to fail regardless.

DavidVentura:
I am experimenting with nomad on risc-v, the default kernel shipped in most VisionFive2 distros comes with the cgroup controllers enabled (see [PR](https://github.com/starfive-tech/linux/pull/125/files) to enable it).

It took me a while to figure out that the controllers being disabled was causing the `preexec_fn` to fail for `raw_exec`, so I wanted to make Nomad more clearly log whenever there's an unsupported cgroup configuration.

I couldn't figure out how to completely disable `raw_exec` in a case where the cgroup config was not supported, but that'd also be fine with me :grin:

shoenig:
Ahh, I have been blissfully unaware of RISC-V development so far, kinda neat if you can already get Nomad running on it! 

Thinking about this a little more, it might make sense to simply check for the enabled controllers at the point in time we switch on cgroups v1 or v2 (or off - which is already a thing). If we can detect the necessary controllers are not enabled we can return `OFF` here and Nomad should already do the "right" thing for a system without cgroups.

https://github.com/hashicorp/nomad/blob/main/client/lib/cgroupslib/mount.go#L41-L43

tgross:
In https://github.com/hashicorp/nomad/issues/19481 we've got a report with the identical symptoms but in this case the cpuset controller is activated, it's just not in the place that Nomad expects. They've got `/sys/fs/cgroup/nomad.slice/cpuset.cpus`. But not `/sys/fs/cgroup/nomad.slice/share.slice/cpuset.cpus`. My read of the `cgroupslib` suggests that path only works if we're in the `OFF` case, as @DavidVentura is saying. Which I wouldn't expect to be the case if `/sys/fs/cgroup` is there.

DavidVentura:
I've updated this per @shoenig's recommendation; in the end, I didn't need to update all the paths which ignore the OFF case, as they are not called at all.
This change results in the `java` and `exec` drivers getting disabled if you are missing some of the essential controllers.
The message is not super clear in the UI and there are currently no logs to indicate what the problem may be.

image

tgross:
Thanks @DavidVentura! I'll review this tomorrow.

client/lib/cgroupslib/mount.go
```diff
@@ -27,7 +29,30 @@ func detect() Mode {
...
+	if err != nil {
+		return false
+	}
+	rootSubtreeControllers := strings.Split(strings.TrimSpace(string(content)), " ")
+
+	for _, controller := range requiredCgroup2Controllers {
+		if !slices.Contains(rootSubtreeControllers, controller) {
+			return false
+		}
+	}
```
shoenig, line 54: This definitely works but to make this a bit cleaner you could do something like,

```go
rootSubtreeControllers := set.From[string](strings.Fields(string(content)))
return rootSubtreeControllers.ContainsSlice(requiredCgroup2Controllers)
```

(using https://pkg.go.dev/github.com/hashicorp/go-set/v2@v2.1.0#Set.ContainsSlice)

tgross (approved):
LGTM! Thanks @DavidVentura!

tgross:
Once CI is complete, I'll merge this and get it backported so it can ship in the next release of Nomad 1.7.x.

Issue #19053, opened 2023-11-09
update "Licensed Work" in license file


tgross:
Our [`release.yml`](https://github.com/hashicorp/nomad/blob/main/.github/workflows/release.yml) workflow updates the version file and a few other items as part of a commit with the message "prepare for next release". We need to also include updating our license file's "Licensed Work" to match the exact version we're shipping as well.

We also need to figure out how to get the version in the license file updated specially for pre-releases, which will end up happening at a different spot in the pipeline?

tgross:
Looks like we'll also need to update the release branch for the beta we just shipped so that the license file's exact version is correct.

tgross:
Putting a pin on this until some internal discussions are resolved. Not a blocker for 1.7.0-beta.2

tgross:
Looks like the resolution to this will be to update the license to "1.7.0 or later", but confirming one more detail around backports before opening a PR for it. (cc @mmcquillan )

tgross:
Last note on this is that after the new year the 1.6.x and 1.5.x branches will start being under BUSL as well. Our legal folks have informed me that each of those branches will need to have a license file that says "1.6.4 or later" and "1.5.11 or later", and that'll close out this issue.

tgross:
I've opened https://github.com/hashicorp/nomad/pull/19186 for 1.7, and will re-title this issue and close it out only once we've done the 1.5.x and 1.6.x license change.

tgross:
Closed by https://github.com/hashicorp/nomad/pull/19194. I think there's probably follow-up, but I'll leave that in the capable hands of RelEng.

PR #19186, opened 2023-11-27
license: note that BUSL applies to 1.7 "or later"

tgross:
The BUSL license applies to a specific version of the software. In order to avoid having to change the version with each version we ship (and ambiguity around prerelease versions), our legal folks have determined we can say "or later" for the version and be covered.

After the new year, we'll need to make an equivalent change for the license in the 1.5.x and 1.6.x branches, which will start shipping under BUSL as well.

Ref: https://github.com/hashicorp/nomad/issues/19053

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -4,7 +4,7 @@ License text copyright (c) 2020 MariaDB Corporation Ab, All Rights Reserved.
 Parameters
 
 Licensor:             HashiCorp, Inc.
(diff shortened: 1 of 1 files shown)


tgross:
Superseded by https://github.com/hashicorp/nomad/pull/19194, which didn't match what we said in the very lengthy chat. But whatever.

PR #19188, opened 2023-11-27
vault: update default JWT auth method path

lgfa29:
Update default auth method path to be `jwt-nomad` to avoid potential conflicts when Vault's `jwt` default is already being used for something else.


tgross (approved):
LGTM :+1:

Issue #18529, opened 2023-09-18
Nomad executes any file in `/plugins/` disregarding the plugin contract.


stswidwinski:
### Nomad version
tip

### Operating system and Environment details
Unix

### Issue

The advertised[ Nomad contract is](https://developer.hashicorp.com/nomad/docs/configuration/plugin) seems to be:

```
The name of the plugin is the plugin's executable name relative to to the plugin_dir. If the plugin has a suffix, such as .exe, this should be omitted.
```

Consequently, if the `plugins` directory looks as follows:

```
\- plugins
  \- foo.exe
  \- bar.exe
```

And the config looks as follows:

```
[...]
  "plugin": {
    "foo": {...}.
    [...]
   }
[...]
```

I would expect `foo.exe` to be used as a plugin and `bar.exe` ignored since it is not mentioned within the configuration at all. However, it seems that both `foo.exe` and `bar.exe` will be executed and used for plugin configuration. Even worse, if both` bar.exe` and `foo.exe` correspond to a driver whose fingerprinting endpoints return the same name and version, it is non-deterministic which one of the binaries will actually be used as the driver.

### Root cause.

When Nomad starts up, it will load plugins and fingerprint them. The rough codepath is:

1. Setup plugins: https://github.com/hashicorp/nomad/blob/main/command/agent/plugins.go#L15
2. New plugin loader: https://github.com/hashicorp/nomad/blob/main/helper/pluginutils/loader/loader.go#L115
3. Init plugin loader: https://github.com/hashicorp/nomad/blob/main/helper/pluginutils/loader/init.go#L57

Now, this has the following logic which goes against the contract described above:

4. `ls`  the plugin directory: https://github.com/hashicorp/nomad/blob/main/helper/pluginutils/loader/init.go#L68
5. execute  all files in it: https://github.com/hashicorp/nomad/blob/main/helper/pluginutils/loader/init.go#L322
6. Use the response of the plugin to get the name of it: https://github.com/hashicorp/nomad/blob/main/helper/pluginutils/loader/init.go#L279-L302

So in other words: the name of the binary is only used to derive the config json for the driver, not to control whether or not the driver should be executed in the first place.

### Reproduction steps

The simplest repro is to have an unrelated, non-executable file within `plugins` and observe that Nomad will eagerly execute it (or rather attempt and fail) despite the `config` not mentioning the file as a plugin.

[...]

PR #19189, opened 2023-11-27
client: deprecate loading plugins without config

lgfa29:
Nomad load all plugins from `plugin_dir` regardless if it is listed in the agent configuration file. This can cause unexpected binaries to be executed.

This commit begins the deprecation process of this behaviour. The Nomad agent will emit a warning log for every plugin binary found without a corresponding agent configuration block.

Ref: #18529

### DIFF
--- a/helper/pluginutils/loader/init.go
+++ b/helper/pluginutils/loader/init.go
@@ -266,6 +266,7 @@ func (l *PluginLoader) fingerprintPlugins(plugins []os.FileInfo, configs map[str
 		name := cleanPluginExecutable(p.Name())
 		c, ok := configs[name]
 		if !ok {
+			// COMPAT(1.7): Skip executing unconfigured plugins in 1.8 or later.
 			l.logger.Warn("plugin not referenced in the agent configuration file, future versions of Nomad will not load this plugin until the agent configuration is updated", "plugin", name)
 		}
 		info, err := l.fingerprintPlugin(p, c)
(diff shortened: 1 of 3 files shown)


helper/pluginutils/loader/init.go
```diff
@@ -264,7 +264,10 @@ func (l *PluginLoader) fingerprintPlugins(plugins []os.FileInfo, configs map[str
 	fingerprinted := make(map[PluginID]*pluginInfo, len(plugins))
 	for _, p := range plugins {
 		name := cleanPluginExecutable(p.Name())
-		c := configs[name]
+		c, ok := configs[name]
+		if !ok {
+			l.logger.Warn("plugin not referenced in the agent configuration file, future versions of Nomad will not load this plugin until the agent configuration is updated", "plugin", name)
```
schmichael, line 269:
```suggestion
			// COMPAT(1.7): Skip executing unconfigured plugins in 1.8 or later.
			l.logger.Warn("plugin not referenced in the agent configuration file, future versions of Nomad will not load this plugin until the agent configuration is updated", "plugin", name)
```

schmichael (approved):
The existing autoloading behavior is nice in that you can management plugins without restarting the Nomad agent. This simplifies operations, packaging, and scripting.

Furthermore it seems unlikely an attacker able to write to the `plugin_dir` couldn't stage other attacks on the system as well.

That being said "seems unlikely" is a poor security posture, and the closest HashiCorp prior art I could find is Vault's plugin system which has an explicit `register-before-use` step: https://developer.hashicorp.com/vault/docs/plugins/plugin-management

**So let's ship this.** We could add an `autoload_plugins = true|false` agent configuration knob to control this behavior, but I think we should wait for that feature request. Nomad already has no many knobs that subtlety change security properties I would hate to add more without good reason.

PR #19196, opened 2023-11-28
client: add metadata to tokens requested by Consul client

pkazmierczak:
This way tokens created by Nomad workloads are easier to keep track of.


tgross (approved):
LGTM! Having this be associated with the `WorkloadIdentifier` field might make it easier for us to implement some kind of out-of-band GC later, too.

PR #19194, opened 2023-11-27
Update LICENSE

hc-github-team-es-release-engineering:
Per legal team request, this makes the LICENSE file ASCII compliant and updates product version/language.

PR #19203, opened 2023-11-28
LICENSE: fix version in license

tgross:
In #19194 the license was incorrectly changed to say that Nomad 1.6.2 or later was under the BUSL, when in fact Nomad 1.6.2 and 1.6.3 are MPL2 licensed. Fix this so that the correct (next!) version is shown as covered only.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -4,41 +4,41 @@ License text copyright (c) 2020 MariaDB Corporation Ab, All Rights Reserved.
 Parameters
 
 Licensor:             HashiCorp, Inc.
-Licensed Work:        Nomdad Version 1.6.2 or later. The Licensed Work is (c) 2023
+Licensed Work:        Nomdad Version 1.7.0 or later. The Licensed Work is (c) 2023
                       HashiCorp, Inc.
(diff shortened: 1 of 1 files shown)


mikenomitch (approved):

LICENSE
```diff
@@ -4,41 +4,41 @@ License text copyright (c) 2020 MariaDB Corporation Ab, All Rights Reserved.
 Parameters
 
 Licensor:             HashiCorp, Inc.
-Licensed Work:        Nomdad Version 1.6.2 or later. The Licensed Work is (c) 2023
+Licensed Work:        Nomdad Version 1.7.0 or later. The Licensed Work is (c) 2023
```
philrenaud, line 7:
```suggestion
Licensed Work:        Nomad Version 1.7.0 or later. The Licensed Work is (c) 2023
```

philrenaud (approved):

mmcquillan (approved):

LICENSE
tgross: LOL, how did I miss that on the original? :laughing:

PR #19214, opened 2023-11-29
cli: fix help formatting on job stop command.

jrasell:
Before:
```
  -no-shutdown-delay
	Ignore the the group and task shutdown_delay configuration so that there is no
    delay between service deregistration and task shutdown. Note that using
    this flag will result in failed network connections to the allocations
    being stopped.
```

After:
```
  -no-shutdown-delay
    Ignore the group and task shutdown_delay configuration so that there is no
    delay between service deregistration and task shutdown. Note that using this
    flag will result in failed network connections to the allocations being
    stopped.
```

### DIFF
--- a/command/job_stop.go
+++ b/command/job_stop.go
@@ -54,10 +54,10 @@ Stop Options:
     only a single region at a time. Ignored for single-region jobs.
 
   -no-shutdown-delay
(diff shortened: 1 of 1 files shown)


gulducat (approved):
nice lil bit of polish :+1:

PR #19216, opened 2023-11-29
Clarify docs around CSI volume context updates

sofixa:
Clarify that when updating a CSI volume's `context` via `nomad volume register`, all parameters need to be specified, not only those to be updated.

### DIFF
--- a/website/content/docs/other-specifications/volume/index.mdx
+++ b/website/content/docs/other-specifications/volume/index.mdx
@@ -144,8 +144,9 @@ parameters {
   of strings passed directly to the CSI plugin to validate the volume. The
   details of these parameters are specific to each storage provider, so consult
   the specific plugin documentation for more information. Only allowed on
-  **volume registration**. Note that this block is declarative, and an update 
-  replaces it in its entirety, thefore all paramters need to be specified.
+  **volume registration**. Note that, like the rest of the volume specification, 
+  this block is declarative, and an update replaces it in its entirety, therefore 
+  all parameters need to be specified.
 
 ## Differences Between Create and Register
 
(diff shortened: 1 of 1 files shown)


website/content/docs/other-specifications/volume/index.mdx
```diff
@@ -144,7 +144,8 @@ parameters {
   of strings passed directly to the CSI plugin to validate the volume. The
   details of these parameters are specific to each storage provider, so consult
   the specific plugin documentation for more information. Only allowed on
-  **volume registration**.
+  **volume registration**. Note that this block is declarative, and an update 
+  replaces it in its entirety, thefore all paramters need to be specified.
```
gulducat, line 148:
```suggestion
  **volume registration**. Note that this block is declarative, and an update 
  replaces it in its entirety, therefore all parameters need to be specified.
```

tgross, line 148: The wording here makes it sound a little bit like this is an exception. Maybe something like:

```suggestion
  **volume registration**. Note that, like the rest of the volume specification, 
  this block is declarative, and an update replaces it in its entirety, therefore 
  all parameters need to be specified.
```

(Also fixes spelling errors)

tgross (changes requested):

gulducat (commented):
Thanks for the clarification!

just a couple little typos to fix then I'll :heavy_check_mark: :)

website/content/docs/other-specifications/volume/index.mdx
gulducat: woops, simultaneous review from Tim and me :P

tgross (approved):
LGTM!

jrasell (approved):

PR #19220, opened 2023-11-29
[ui] Title bar job start button now observes job submission variables data

### LINKED ISSUE #19198: Unable to restart a job from UI in 1.6.3
### Nomad version
1.6.3

Ubuntu 22.04.3

### Issue
If a job with variables is stopped from the UI, it cannot be started again.

### Reproduction steps
1. Create a job that has variables without default values.
2. Start the job providing values of the variables.
3. Stop the job from the UI.
4. Start the job from the UI again.

#### Expected Result
The job is started.

#### Actual Result
The job is not started with an error message in the "Overview" tab saying the following (the variable names were redacted):

Could Not Start Job
Failed to parse job: <nil>: Unset variable "<redacted>"; A used variable must be set or have a default value; see https://www.nomadproject.io/docs/job-specification/hcl2/variables for details. <nil>: Unset variable "<redacted>"; A used variable must be set or have a default value; see https://www.nomadproject.io/docs/job-specification/hcl2/variables for details. <nil>: Unset variable "<redacted>"; A used variable must be set or have a default value; see https://www.nomadproject.io/docs/job-specification/hcl2/variables for details. <nil>: Unset variable "<redacted>"; A used variable must be set or have a default value; see https://www.nomadproject.io/docs/job-specification/hcl2/variables for details.

However, the actual values of the variables can be seen in the "HCL Variable Values" section on the "Definition" tab. The job definition on the same tab is also correct.

Everything is fine in 1.6.1, so it seems to be introduced in later versions.
It seems to be related to #18536

philrenaud:
Resolves #19198 

Previously, after stopping a job with variables passed in but no default values, going to the definition page and editing would work fine (as variable literals were already present on the page), but the "start" button would not consider them.

Now they're specifically checked when the job start button is pressed.

Side-effect: utility-ified the json-to-hcl method previously internal to the job editor component.
image
image


ui/app/components/job-editor.js
```diff
@@ -37,11 +38,10 @@ export default class JobEditor extends Component {
     }
 
     if (this.args.variables) {
+      console.log('aye vars', this.args.variables);
```
jrasell, line 41: Is this expected to be here, or remains of debugging?

jrasell (commented):
I've left a minor inline comment; I also think we will need a changelog entry for this.

I tried testing this locally and I get the same error as the linked issue describes. I'm not sure if I am testing this wrong.

ui/app/components/job-editor.js
philrenaud: hah, remains of debugging. Thanks for spotting it!

philrenaud:
@jrasell I'm testing with the following jobspec (from your comment in the linked issue, I think):

```
variable "count" {}

job "example" {
  group "cache" {
    count = var.count
    task "redis" {
      driver = "docker"
      config {
        image = "redis:7"
      }
    }
  }
}
```

and running with `nomad job run  -var="count=4" ../nomad-job-specs/bug-empty-variables.nomad.hcl`

I'm able to get the "start" button to fire it back up with a count of 4 in the following ways:
- stop, and then immediately restart, via job index butons
- stop, and then refresh, and restart via job index buttons
- stop from elsewhere (cli `nomad job stop _____` and open the web UI and start again

In all of these cases, if you happen to go to the definition tab for the job, you should see the count in the variables box:

image

Issue #19111, opened 2023-11-16
E2E: node drain test failure


tgross:
After fixing some bugs with the E2E test runs, we've got some failing tests to fixup. One of them is the node drain test

```
=== FAIL: e2e/nodedrain TestNodeDrain/KillTimeout (53.34s)
    node_drain_test.go:204: draining node 87447de8-c5a5-2f3b-6448-025dfb192d74
    node_drain_test.go:215: waiting for kill_timeout to expire
    node_drain_test.go:228: waiting for migration to complete
    node_drain_test.go:394: alloc has drained from node=87447de8 after 17.640054847s
    assert.go:14: 
        node_drain_test.go:233: expected condition to pass within wait context
        ↪ error: wait: timeout exceeded
```

gulducat:
This one is flaky, failing some 20-25% of the time in nightly runs.

I can't repro locally with 2 client nodes (minimum required for the test, but CI has 4), at least not in isolation running this one test after ~30ish attempts.

When it fails in CI, it's getting past the alloc being drained, then fails to meet this condition (in time)
https://github.com/hashicorp/nomad/blob/e3c8700/e2e/nodedrain/node_drain_test.go#L237

```go
wait.BoolFunc(func() bool {
	node, _, err := nomadClient.Nodes().Info(oldNodeID, nil)
	must.NoError(t, err)
	return node.DrainStrategy == nil
}),
```

I'm not sure why `DrainStrategy` is staying non-`nil`, or what its specific value is, but that's how far I've gotten so far.

PR #19226, opened 2023-11-29
e2e: give node drain KillTimeout test more time

gulducat:
and error more verbosely if it fails

attempting to either resolve or illuminate #19111 (it's flaky, and I'm not able to reproduce)


tgross (approved):
LGTM :+1:

PR #19227, opened 2023-11-29
test: add consul namespace rules to consulcompat

lgfa29:
When configuring Consul for multi-namespace support, the JWT auth method needs to specify namespace rules. This attribute is set to `nil` in CE but is used in Nomad ENT.


pkazmierczak (approved):

tgross (approved):
LGTM :+1:

PR #19230, opened 2023-11-30
scheduler: fix task-level consul diff

lgfa29:
Fix `tasksUpdated()` to compare the task level `consul` blocks instead of the group.


tgross (approved):
LGTM!

tgross:
@lgfa29 @pkazmierczak I've also opened https://github.com/hashicorp/nomad/issues/19235 as a follow-up so that we can maybe reduce the challenges around handling this block correctly in the near-future. I can probably pick up that work as part of the Consul work I've been assigned-to for the 1.7.x-1.8 cycle.

PR #19237, opened 2023-11-30
cli: setup consul proper ns handling

pkazmierczak:
In order to correctly handle Consul namespaces, auth methods and binding rules must always be created in the `default` namespace only.

### DIFF
--- a/command/setup_consul.go
+++ b/command/setup_consul.go
@@ -623,7 +623,11 @@ func (s *SetupConsulCommand) removeConfiguredComponents() int {
 		componentsToRemove["Auth method"] = []string{consulAuthMethodName}
 	}
 
-	authMethodRules, _, err := s.client.ACL().BindingRuleList(consulAuthMethodName, nil)
+	qo := &api.QueryOptions{}
+	if s.consulEnt {
+		qo.Namespace = "default"
+	}
+	authMethodRules, _, err := s.client.ACL().BindingRuleList(consulAuthMethodName, qo)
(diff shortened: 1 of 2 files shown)


command/setup_consul.go
```diff
@@ -432,18 +438,24 @@ func (s *SetupConsulCommand) renderAuthMethod(name string, desc string) (*api.AC
 		TokenLocality: "local",
 		Config:        authConfig,
 	}
-	if s.consulEnt && (s.clientCfg.Namespace == "" || s.clientCfg.Namespace == "default") {
+	if s.consulEnt {
 		method.NamespaceRules = []*api.ACLAuthMethodNamespaceRule{{
 			Selector:      "",
-			BindNamespace: "${value.nomad_namespace}",
+			BindNamespace: s.clientCfg.Namespace,
```
lgfa29, line 444:
```suggestion
			BindNamespace: "${value.consul_namespace}",
```

We want to use the Consul namespace set in the job. This claim was introduced in https://github.com/hashicorp/nomad/pull/19228

lgfa29 (changes requested):

command/setup_consul.go
pkazmierczak:
```
[✘] Could not create Consul auth method: Unexpected response code: 500 (Invalid Auth Method: invalid NamespaceRules[0].BindNamespace: 1:3: unknown variable accessed: value.consul_namespace)
```

PR #19238, opened 2023-11-30
website: fix spellcheck path and cleanup some misspellings

shoenig:
We moved the website content directory a while back and never updated the spell checking thing used by `make check`; fix the path and fixup old spelling mistakes.


tgross (approved):
LGTM!

PR #19245, opened 2023-11-30
[ui] job/$job/services page: consider tags when looking up services

### LINKED ISSUE #17109: Nomad UI: incorrect service allocations counts under jobs/<name>/services
### Nomad version
1.5.3

### Issue
The Services tabs under for current job incorrectly counts allocations when service instances with the same name but different tag sets exists.

### Reproduction steps

1. Define two `service{}` configs for a task, each with its own port and a seperate tag.
2. Deploy the job. 
3. Check the Jobs > Services tab in the web ui.

#### Expected Result

Allocation count should be equal to 1 under each service & tag combination.

#### Actual Result

Allocation count is 2 in every table line:

CleanShot_20230508T080435Z@2x

Here's the output of `nomad service info`:

```
$ nomad service info whoami
Job ID  Address             Tags   Node ID   Alloc ID
whoami  172.30.36.64:31283  [foo]  d2191c46  70b914ec
whoami  172.30.36.64:22345  [bar]  d2191c46  70b914ec
```

### Job file (if appropriate)

```
job "whoami" {
  group "whoami" {
    count = 1

    network {
      port "http" {}
      port "http2" {}
    }

    task "whoami" {
      driver = "docker"

      config {
        image = "traefik/whoami:latest"
        ports = ["http", "http2"]

        args = [
          "--port", "${NOMAD_PORT_http}"
        ]
      }

      service {
        name     = "whoami"
        provider = "nomad"
        port     = "http"
        tags = ["foo"]
      }

      service {
        name     = "whoami"
        provider = "nomad"
        port     = "http2"
        tags = ["bar"]
      }
    }
  }
}
```

philrenaud:
Resolves #17109 

Before, a job could have multiple services with the same name but differing tags, and their allocation count would double up (their filter criteria would be satisfied more times than intended in the UI)

Now, in addition to name and level (group or task) equality, we also look up `service.tags` equality.

Example jobspec for testing:

```
job "whoami" {
  group "whoami" {
    count = 1

    network {
      port "http" {}
      port "http2" {}
    }

    task "whoami" {
      driver = "docker"

      config {
        image = "traefik/whoami:latest"
        ports = ["http", "http2", "http3"]

        args = [
          "--port", "${NOMAD_PORT_http}"
        ]
      }

      service {
        name     = "whoami"
        provider = "nomad"
        port     = "http"
        tags = ["foo"]
      }
      service {
        name     = "whoami"
        provider = "nomad"
        port     = "http2"
        tags = ["bar"]
      }
      service {
        name     = "no-tags"
        provider = "nomad"
        port     = "http2"
        tags = []
      }
      service {
        name     = "many-tags"
        provider = "nomad"
        port     = "http2"
        tags = ["foo", "bar", "baz", 3]
      }
    }
  }
}
```

NOTE: There is still potential for confusion here if the two services are identical, except for their `port` / `portLabel` This is because we do not expose the service's port in the same way when it's a serviceFragment (read from the `jobspec.groups[].services[]`, which gives us back a port label) or when it's a service proper (from the `job/$id/services endpoint, where it gives us back a port identifying number). I consider this edge-casey enough that my suggestion to jobspec authors who may run into this problem in the future is to not name your services the same if they differ in port. Or really to not name any two services on the same task/group the same name, generally.

### DIFF
--- a/.changelog/19245.txt
+++ b/.changelog/19245.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-ui: disambiguate same-named services with distinct tag arrays
+ui: disambiguate services with the same name but different tags
 ```
--- a/.changelog/19245.txt
+++ b/.changelog/19245.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: disambiguate same-named services with distinct tag arrays
+```
--- a/ui/app/components/job-service-row.js
+++ b/ui/app/components/job-service-row.js
@@ -15,7 +15,7 @@ export default class JobServiceRowComponent extends Component {
   gotoService(service) {
     if (service.provider === 'nomad') {
       this.router.transitionTo('jobs.job.services.service', service.name, {
-        queryParams: { level: service.level },
+        queryParams: { level: service.level, tags: service.tags },
         instances: service.instances,
       });
     }
--- a/ui/app/controllers/jobs/job/services/index.js
+++ b/ui/app/controllers/jobs/job/services/index.js
@@ -75,8 +75,8 @@ export default class JobsJobServicesIndexController extends Controller.extend(
         (s) =>
           s.name === fragment.name &&
           s.derivedLevel === fragment.level &&
-          s.tags.length === fragment.tags.length &&
-          s.tags.every((tag) => fragment.tags.includes(tag))
+          s.tags?.length === fragment.tags?.length &&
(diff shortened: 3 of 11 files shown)


shoenig (dismissed):
LGTM!

philrenaud:
This is a valid issue with a long-outdaed method to solve it: the UI counts same-named services as the same thing, even if they have different tag sets. We should use an ensemble identifier.

PR #19247, opened 2023-11-30
[ui] Helios upgraded to 3.3.0

philrenaud:
Upgraded our design systems library to the latest version.

This required pinning ember-element-helper to 0.6.0 for ember-power-select and probably other reasons.

There were a few small visual artifact diffs here (highlights below), but the only "noticeably broken" change was the alignment of dropdown buttons siwtched to centred by default (only noticeable when the width is extended from elsewhere, like we do in Actions dropdowns with multiple allocs)

### page-header component headers are bold:
image
image

### Resolved with this PR: Dropdown buttons centered:
image
image

### DIFF
--- a/ui/app/styles/components/actions.scss
+++ b/ui/app/styles/components/actions.scss
@@ -36,6 +36,9 @@
       padding-left: 0;
       padding-right: 0;
       border-width: 0;
+      span {
+        text-align: left;
+      }
     }
   }
 }
--- a/ui/yarn.lock
+++ b/ui/yarn.lock
@@ -3623,14 +3623,6 @@
     ember-cli-version-checker "^5.1.2"
     semver "^7.3.5"
 
-"@embroider/addon-shim@1.8.3":
-  version "1.8.3"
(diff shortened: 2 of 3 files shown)


ui/app/styles/components/actions.scss
```diff
@@ -36,6 +36,9 @@
       padding-left: 0;
       padding-right: 0;
       border-width: 0;
+      span {
+        text-align: left;
+      }
```
gulducat, line 41: I assume it's fine, but out of curiosity, is this directly related to the dep updates?

And is nothing else needed?  I might expect more breakage from a major version bump.

ui/yarn.lock
gulducat, line 1: :hmm-yes:

gulducat (approved):
:rainbow:

ui/app/styles/components/actions.scss
gulducat: PR description updated after I posted this.  thanks!  :+1:

PR #19258, opened 2023-12-01
docs: split `consul` config params into client vs server sections

tgross:
Some sections of the `consul` configuration are relevant only for clients or servers. We updated our Vault docs to split these parameters out into their own sections for clarity. Match that for the Consul docs.

Preview link: https://nomad-amgfwjney-hashicorp.vercel.app/nomad/docs/configuration/consul


pkazmierczak (approved):
LGTM, makes it much clearer!

Issue #17362, opened 2023-05-30
docs: client cpu resource reservation documentation is wrong / missing


shoenig:
I noticed the Client `reserved.cores` documentation specifies a [count of cores](https://developer.hashicorp.com/nomad/docs/configuration/client#cores) (int), whereas the actual configuration value accepts a cpuset (string).

e.g.

```hcl
client {
  reserved {
    cores = "3-8"
  }
}
```

https://github.com/hashicorp/nomad/blob/release/1.5.6/command/agent/config.go#LL1109C2-L1109C36

In addition, there is an undocumented config option `reservable_cores` which can be used to override the set of fingerprinted reservable cores.

https://github.com/hashicorp/nomad/blob/release/1.5.6/command/agent/config.go#L236-L237

PR #19266, opened 2023-12-01
docs: fix documentation of client.reserved.cores

shoenig:
Fixes: https://github.com/hashicorp/nomad/issues/17362

Leaving `reservable_cores` undocumented as it's a weird feature that probably shouldn't have been added to begin with and nobody has asked for it.


Juanadelacuesta (approved):
👍

Issue #16886, opened 2023-04-14
Poststop lifecycle task - Can't request Vault token for terminal allocation


beninghton:
### Nomad version
Nomad v1.5.2
BuildDate 2023-03-21T22:54:38Z
Revision 9a2fdb5f53dce81edf2802f0b64962e07596fd03

### Operating system and Environment details
NAME="Fedora Linux"
VERSION="35.20220313.3.1 (CoreOS)"

### Issue
Task with "poststop" hook lifecycle can't get a key/value from vault. But task in the same group but with "prestart" lifecycle is able.

### Reproduction steps
You can run any task with poststop hook lifecycle.
You can place vault policy inside the task block, or outside, under job/group blocks. Or in all the places, doesn't matter. It will not work.
    task "poststoptask" {
      lifecycle {
        hook = "poststop"
      }
      vault {
        policies = ["any_vault_policy"]
      }
    }

#### Expected Result
 Key/value is derived from vault.

#### Actual Result
 Key/value is not derived from vault.

### Nomad Client logs (if appropriate)
Apr 14, '23 15:10:03 +0700 | Template | Missing: vault.read(kv/nomad-drain-operator/token)
Apr 14, '23 15:10:00 +0700 | Killing | Vault: server failed to derive vault token: Can't request Vault token for terminal allocation
Apr 14, '23 15:10:00 +0700 | Task Setup | Building Task Directory
Apr 14, '23 15:09:08 +0700 | Received | Task received by client

lgfa29:
Thanks for the report @beninghton.

It seems like the error is coming from here, which is a very old part of the code:
https://github.com/hashicorp/nomad/blob/3067191c5197cacda71b0643bea4327f13bd8ce3/nomad/node_endpoint.go#L1793-L1796

The problem is that when task lifecycle was implemented this part of the code was not updated to consider this use case: the allocation is set to be `ServerTerminal` but it hasn't finished yet because it has to finish its `poststop` task.

I have placed this issue in our board for further trigage and roadmapping.

dani:
Does anyone has a workaround for this ? Hitting the same issue, in a poststop task which must access KV from vault

dani:
In my case, as I had both a prestart and a poststop task needing the same KV secret from vault, what I did 
- Fetch the secret from vault only in the prestart task, and store it in secrets/my-api-key.env
- In the poststop task, mount this file from the prestart task
```
      config {
        image = "myimage:latest"
[...]

PR #19268, opened 2023-12-01
vault: fix legacy token workflow for poststop tasks

tgross:
The new Workload Identity workflow for Vault tokens correctly handles post-stop tasks, however the legacy workflow does not. Attempts to get a Vault token are rejected if the allocation is server-terminal or client-terminal, but we should be waiting until the allocation is client-terminal (only) so that poststop tasks get a chance to get Vault tokens too.

Fixes: https://github.com/hashicorp/nomad/issues/16886
See https://github.com/hashicorp/nomad/issues/16886#issuecomment-1836456414 for testing details.

### DIFF
--- a/.changelog/19268.txt
+++ b/.changelog/19268.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+vault: Fixed a bug where poststop tasks would not get a Vault token
+```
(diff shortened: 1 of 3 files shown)


tgross:
Oops, I need to fix a test that's checking the server-terminal status. Will fix that right after lunch.

dpn:
@tgross Strangely enough we were _just_ looking at this ticket yesterday. Crazy coincidence. Thanks!

PR #19270, opened 2023-12-01
variables: fix access to variables for poststop tasks

tgross:
In the recent auth refactor, we accidentally fixed a bug where poststop tasks would not get access to Variables. Fix this same bug for backports by ensuring that we use client-terminal status and not server-terminal status to enforce access.

Note this is a backport-only PR.

Ref: https://github.com/hashicorp/nomad/issues/16886


tgross:
Backported to 1.5.x as well.

PR #19282, opened 2023-12-01
docs: clarify behavior and recommendations for mTLS vs TLS for HTTP

tgross:
Some of our documentation on `tls` configuration could be more clear as to whether we're referring to mTLS or TLS. Also, when ACLs are enabled it's fine to have `verify_https_client=false` (the default). Make it clear that this is an acceptably secure configuration and that it's in fact recommended in order to avoid pain of distributing client certs to user browsers.

cc @apollo13 for your thoughts as well.


schmichael (approved):

apollo13:
This looks so much better than before. Thank you for your work on it!

PR #19284, opened 2023-12-01
e2e: jobs3-submitted jobs automatically cleanup

gulducat:
Currently, when submitting a job like this:

```go
sub, cleanup := jobs3.Submit(t, "./job.nomad.hcl")
t.Cleanup(cleanup)
```

if the deployment is not successful within `Submission.run()`, or specifically if any `must` (`t.Fatal`) conditions are met before the end of `Submit()`, the test will fail and exit, prior to `t.Cleanup()` being run, so cleanup will not occur.

This change adds `t.Cleanup()` within `Submission.run()`, after the evaluation has passed, but before watching the deployment, so even a failed deployment will still be cleaned up (unless configured not to).

It still returns a cleanup func, and it's safe to call it again, so I haven't gone and changed all the callers, but I can do that if you'd like to encourage me to.


e2e/v3/jobs3/jobs3.go
```diff
@@ -315,6 +316,10 @@ EVAL:
 		}
 	}
 
+	if !sub.noCleanup {
+		sub.t.Cleanup(sub.cleanup)
+	}
```
tgross, line 321: I think this hits the same problem you described in the PR. If the eval API returns a fatal condition, we'll never cleanup. So I think you want to move this right underneath the `Register` API call, right after checking if you got an error from that.

tgross (changes requested):

e2e/v3/jobs3/jobs3.go
gulducat: Ah yeap, I got myopic about the failures I have seen, which have been "deployment wasn't happy fast enough" and neglected earlier possible errors.

tgross (approved):
LGTM

PR #19296, opened 2023-12-04
vault: claims for WI workloads should not contain nomad_group

pkazmierczak:


tgross (approved):
LGTM! I think we were talking about maybe including `nomad_group` at some point and not just the Alloc ID in (`nomad_allocation_id`), but it's still not included as of 1.7.

PR #19303, opened 2023-12-04
e2e: add a ShowState() option to cluster3.Establish options

shoenig:
This will dump much of the interesting parts of cluster state, including
available nodes and their status, existing allocations and their status,
and existing evaluations and their status.

In theory the cluster should be in a "clean" state before each e2e suite is run. By dumping this state we can see what state the cluster is actually in before running tests.

### DIFF
--- a/e2e/metrics/metrics_test.go
+++ b/e2e/metrics/metrics_test.go
@@ -50,6 +50,7 @@ func TestMetrics(t *testing.T) {
 	cluster3.Establish(t,
 		cluster3.Leader(),
 		cluster3.LinuxClients(1),
+		cluster3.ShowState(),
 	)
 
 	t.Log("tweaking podman registry auth files ...")
--- a/e2e/v3/cluster3/cluster3.go
+++ b/e2e/v3/cluster3/cluster3.go
@@ -11,10 +11,12 @@ import (
 	"time"
 
 	consulapi "github.com/hashicorp/consul/api"
+	"github.com/hashicorp/nomad/api"
 	nomadapi "github.com/hashicorp/nomad/api"
 	vaultapi "github.com/hashicorp/vault/api"
 	"github.com/shoenig/test/must"
 	"github.com/shoenig/test/wait"
+	"oss.indeed.com/go/libtime"
 )
 
 type Cluster struct {
@@ -30,6 +32,7 @@ type Cluster struct {
 	vaultReady     bool
 	linuxClients   int
 	windowsClients int
+	showState      bool
 }
 
 func (c *Cluster) wait() {
@@ -155,11 +158,14 @@ func Establish(t *testing.T, opts ...Option) {
 		t:       t,
 		timeout: 10 * time.Second,
 	}
+
 	for _, opt := range opts {
 		opt(c)
 	}
+
 	c.setClients()
 	c.wait()
+	c.dump()
 }
 
 func (c *Cluster) setClients() {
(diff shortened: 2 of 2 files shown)


shoenig:
Quick example against a dev agent with a sleep job running

```
➜ go test -v
=== RUN   TestExample

--- LEADER / SERVER STATUS ---
leader:     127.0.0.1:4647
peer (1/1): 127.0.0.1:4647

--- NODE STATUS ---
NODE diablo @ dc1 (1/1)
	ID: 875ee36f-7c64-e9dc-35fb-7780073ae470
	NodeResources: shares: 56000, cores: 32
	Pool: default, Class: ""
	Status: ready 
	Drain: false
	[raw_exec]
		Detected: true
		Healthy: true "Healthy"
	[qemu]
		Detected: false
		Healthy: false ""
	[exec]
		Detected: true
		Healthy: true "Healthy"
	[docker]
		Detected: false
		Healthy: false "Failed to connect to docker daemon"
	[java]
		Detected: true
		Healthy: true "Healthy"
	[mock_driver]
		Detected: true
		Healthy: true "Healthy"
	Events
		(1/1) Node registered @ 2023-12-04 14:54:37 +0000 UTC

--- ALLOCATIONS (found 1) ---
ALLOC (default/demo, group)
	Node: diablo, NodeID: 875ee36f-7c64-e9dc-35fb-7780073ae470
	ClientStatus: running "Tasks are running"
	ClientTerminalStatus: false
	DesiredStatus: run ""
	ServerTerminalStatus: false
	Deployment: a6394230-15ea-0149-f671-543cd3443b19, Healthy: true
	Task [sleep] shares: 500, cores: 0, memory: 256, memory_max: 0

--- EVALUATIONS ---
EVAL (1/1) default/demo on ""
	Status: complete
	CreateIndex: 572, CreateTime: 2023-12-04 15:53:10.678 +0000 UTC
	DeploymentID: a6394230-15ea-0149-f671-543cd3443b19
	QuotaLimitReached: ""
	EscapedComputedClass: false
	BlockedEval: ""
	ClassEligibility: map[]
	QueuedAllocations: map[group:0]

--- END ---

=== RUN   TestExample/testEnv
--- PASS: TestExample (1.04s)
    --- PASS: TestExample/testEnv (1.03s)
PASS
ok  	github.com/hashicorp/nomad/e2e/example	1.041s
```

e2e/metrics/metrics_test.go
```diff
@@ -50,6 +50,7 @@ func TestMetrics(t *testing.T) {
 	cluster3.Establish(t,
```
gulducat, line 50: Might be nice to (be able to) show state both before and after the test runs.

Either this new `ShowState()` option could do that (perhaps optionally), or...

If `Establish()` returned the `Cluster` it makes, and `dump` was exported instead of private, then one could dump state whenever they'd like, including after the test:

```go
c := cluster3.Establish(t, ...)
c.Dump() // before
t.Cleanup(c.Dump) // after
```

or similar.
Might even be nice to include when any test using this helper fails?  Could add a Cleanup that checks if `t.Failed()` and auto-dump.  That might produce real noisy output though, idk.

shoenig: Yeah I thought about doing something like that. What we _really_ want is a way to take a snapshot of the cluster state before a test run, then dump that snapshot after the test fails (and maybe also dump the state of the cluster after the test (or even better, a diff)). Left as an exercise for the reader, of course.

PR #19304, opened 2023-12-04
[docs] Note reboot to update `bridge_network_hairpin_mode`

angrycub:
Circling back on the comment in [#15961](https://github.com/hashicorp/nomad/pull/15961#discussion_r1092464814). This adds a note to the `bridge_network_hairpin_mode` setting indicating that if it is changed that you need to restart the node.  

I don't thing restarting is a strict requirement but it is by far easier (and more tooling friendly) than the corresponding manual steps.


lgfa29 (approved):
Oops, sorry I missed this 😄

PR #19312, opened 2023-12-05
e2e: push nomad token to test servers

gulducat:
so humans with root shell access can use it to debug

this is not ideal security, but this is a short-lived test cluster.


tgross (approved):
LGTM. As we discussed offline, this is a reasonable temporary hack until we can get a better KMS-backed solution in place.

shoenig (approved):

PR #19313, opened 2023-12-05
e2e: remove deprecated template_file dependency for tf

pkazmierczak:
This also allows running tf for our e2e suite locally on darwin.


e2e/terraform/ecs.tftpl
```diff
@@ -0,0 +1,2 @@
+security_groups = ["${sg_id}"]
+subnets         = ["${subnet_id}"]
```
tgross, line 2: Missing the EOF here

e2e/terraform/volumes.tftpl
```diff
@@ -0,0 +1,15 @@
...
+
+capability {
+  access_mode     = "single-node-writer"
+  attachment_mode = "file-system"
+}
+
+capability {
+  access_mode = "single-node-reader"
+  attachment_mode = "file-system"
+}
```
tgross, line 15: Missing EOF

e2e/terraform/nomad-acls.tf
```diff
@@ -12,24 +12,20 @@ locals {
...
 resource "null_resource" "bootstrap_nomad_acls" {
   depends_on = [module.nomad_server]
   triggers = {
-    script = data.template_file.bootstrap_nomad_script.rendered
+    script = templatefile("${local.nomad_env} ./scripts/bootstrap-nomad.sh", {})
   }
 
   provisioner "local-exec" {
-    command = data.template_file.bootstrap_nomad_script.rendered
+    command = templatefile("${local.nomad_env} ./scripts/bootstrap-nomad.sh", {})
```
tgross, line 25: The `templatefile` function takes a path, whereas this is a rendered string that was originally the rendered _output_ of a `template_file` block. I _think_ we can probably do without the `templatefile` command here entirely if we verify that the `local.nomad_env` is properly interpolated.

(Plus the `triggers` block above becomes useless once we're no longer responding to that render event.)

pkazmierczak: I made a correction in [d467da7](https://github.com/hashicorp/nomad/pull/19313/commits/d467da720c830ad7fd8c6a307b48699c007c4fe9) which I think makes sense.

e2e/terraform/nomad-acls.tf
```diff
@@ -15,21 +15,14 @@ locals {
...
   triggers = {
-    script = data.template_file.bootstrap_nomad_script.rendered
+    script = "${local.nomad_env} ./scripts/bootstrap-nomad.sh"
   }
 
   provisioner "local-exec" {
-    command = data.template_file.bootstrap_nomad_script.rendered
+    command = aws_instance.server.0.public_ip != "" ? local.nomad_env : "echo 'Nomad server not ready yet, skipping bootstrap'"
   }
 }
```
tgross, line 24: I'm not sure what this is for... if there's no public IP, wouldn't we be unable to run the bootstrapping in the first place and therefore just halt when the bootstrap skip fails?

tgross (approved):
LGTM

PR #19316, opened 2023-12-05
e2e: skip connect test that does a restart of nomad agent

shoenig:


shoenig:
Yeah, in particular we need to figure out what to do with tests that want to modify cluster state (e.g. this and the disconnected clients tests and some others). Like maybe have parallel e2e runs for those specifically so they don't potentially ruin the cluster for unrelated tests.

PR #19321, opened 2023-12-05
deps: Switch from mitchellh/cli to hashicorp/cli

angrycub:
Thanks @angrycub!

This PR switches from the achieved `mitchellh/cli` to the internally maintained `hashicorp/cli`. A semgrep rule is included to stop usage of this library in the future.

The change is like-for-like, therefore my plan is to backport this to all active release branches. Additionally, backporting CLI bug fixes might be painful without this. If any reviewers think this is a bad idea, please let me know.

The remaining indirect reference is due to `github.com/hashicorp/go-secure-stdlib/listenerutil` which has migrated the library on main, but not released this version. I have submitted a request for a release via https://github.com/hashicorp/go-secure-stdlib/issues/151. Once released, we can followup with a dependency update PR to remove the indirect dependency.


shoenig:
Should probably also add the old import to the list of forbidden packages semgrep rule
https://github.com/hashicorp/nomad/blob/main/.semgrep/imports.yml

tgross:
Ref https://hashicorp.atlassian.net/browse/NET-10377

pkazmierczak (approved):
Looks good!

pkazmierczak (approved):

PR #19327, opened 2023-12-06
cli: add `-dev-consul` and `-dev-vault` agent mode

lgfa29:
The `-dev-consul` and `-dev-vault` flags add default identities and configuration to the Nomad agent to connect and use the workload indetity integration with Consul and Vault.

### DIFF
--- a/command/agent/command.go
+++ b/command/agent/command.go
@@ -61,7 +61,6 @@ type Command struct {
 }
 
 func (c *Command) readConfig() *Config {
-	var dev *devModeConfig
 	var configPath []string
 	var servers string
 	var meta []string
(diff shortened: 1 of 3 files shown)


command/agent/config.go
```diff
@@ -1177,47 +1177,43 @@ type devModeConfig struct {
...
 	connectMode bool
+	consulMode  bool
+	vaultMode   bool
 
 	bindAddr string
 	iface    string
 }
 
-// newDevModeConfig parses the optional string value of the -dev flag
-func newDevModeConfig(devMode, connectMode bool) (*devModeConfig, error) {
```
tgross, line 1186: We inlined this constructor function because it had only one production caller, but we had a test for the validation we'll want to update too.

command/agent/command.go
```diff
@@ -1405,6 +1420,14 @@ General Options (clients and servers):
     interface rather than localhost for using Consul Connect. This
     mode is supported only on Linux as root.
 
+  -dev-consul
+    Starts the agent in development mode with a default Consul configuration
+    for Nomad workload identity.
```
tgross, line 1425: It might be a good idea to note that this can be combined with `-dev-connect` and doesn't automatically activate `-dev-connect`.

lgfa29: Also updated `-dev-connect` to reference `-dev-consul`. I wondered if `-dev-connect` should imply `-dev-consul`, but maybe something to consider later.

tgross: Once we completely get rid of the legacy workflow, having `-dev-connect` imply `-dev-consul` would make sense. But before that we'd be sticking people trying out legacy workflow in a bad spot (potentially including Nomad developers :grinning: )

tgross (approved):
LGTM!

PR #19333, opened 2023-12-06
ci: use go-modtool with config file

shoenig:
added support for a config file instead of passing arguments to flags in https://github.com/shoenig/go-modtool/releases/tag/v0.2.0

this will be nice when we invoke the `merge` utility in the merge script, so that we don't have to keep the arguments to the flags in sync in two places


pkazmierczak (approved):
LGTM!

Issue #19269, opened 2023-12-01
Nomad deployment progressing without healthy allocations


sluebbert:
### Nomad version
1.6.3

### Operating system and Environment details
Linux

### Issue
Nomad documentation describes the `min_healthy_time` settings as the minimum amount of time an allocation should be reporting as healthy before continuing on with the deployment by replacing old allocations.

Since our health checks on services run every 2 seconds, we didn't see the 10 second default for this setting as necessary. Reducing the setting to something like 1 second or 2 seconds allows Nomad to progress through deployments faster, but it seems to be waiting a few seconds after the allocation starts up, **not** after it registers as healthy in Consul.

Our real issue comes down to client side errors for a few seconds as no healthy upstreams are present.

My best way to solve this at the moment is to revert `min_healthy_time` back to the default 10 seconds to hide the issue better since an allocation is usually healthy in Consul within that time.

### Reproduction steps
1. Run a job with a group instance count of 2.
1. Cause a deployment by changing the version of a task's Docker image.
1. Watch in Consul as Nomad immediately kills off the first old service instance to replace it with the new version.
1. Watch in Nomad's deployment progress to see that it acknowledges the new allocation as healthy before Consul does, several seconds before consul does.
1. Watch in Nomad and Consul as Nomad kills off the next old allocation (the only one in Consul still available to serve requests) because Nomad thinks the first new one is ready when it is not.

I might think this is just the UI out of sync as it routinely updates state, but we also get a bunch of client upstream errors during this transition that indicates it is not just a UI thing.

#### Expected Result
Nomad should wait `min_healthy_time` from when Consul registers the allocation as healthy from checks before Nomad progresses with a deployment.

#### Actual Result
Nomad appears to wait `min_healthy_time` from when Nomad thinks the allocation successfully started, not when Consul has registered the service as healthy.

At times I can actually catch Nomad killing off the only remaining healthy instance and then waiting until the new instance is healthy before starting the next new one resulting in even more down time.
You can see this in the attached demo GIF.

### Job file (if appropriate)
```hcl
job "testwebapi" {
    type = "service"
    datacenters = ["*"]

    group "app" {
        count = 2

        restart {
            interval = "2m"
            attempts = 2
            delay    = "5s"
            mode     = "fail"
        }

        reschedule {
             delay          = "5s"
             delay_function = "exponential"
             max_delay      = "2m"
             unlimited      = true
        }

        update {
            healthy_deadline = "30s"
            progress_deadline = "31s"
[...]

PR #19334, opened 2023-12-06
connect: deployments should wait for Connect sidecar checks

tgross:
When a Connect service is registered with Consul, Nomad includes the nested `Connect.SidecarService` field that includes health checks for the Envoy proxy. Because these are not part of the job spec, the alloc health tracker created by `health_hook` doesn't know to read the value of these checks.

In many circumstances this won't be noticed, but if the Envoy health check happens to take longer than the `update.min_healthy_time` (perhaps because it's been set low), it's possible for a deployment to progress too early such that there will briefly be no healthy instances of the service available in Consul.

Update the Consul service client to find the nested sidecar service in the service catalog and attach it to the results provided to the tracker. The tracker can then check the sidecar health checks.

Fixes: https://github.com/hashicorp/nomad/issues/19269

### DIFF
--- a/command/agent/consul/service_client.go
+++ b/command/agent/consul/service_client.go
@@ -1761,7 +1761,7 @@ func (c *ServiceClient) AllocRegistrations(allocID string) (*serviceregistration
 				}
 			}
 
-			if sidecarService := services[serviceID+sidecarSuffix]; sidecarService != nil {
+			if sidecarService := getNomadSidecar(serviceID, services); sidecarService != nil {
 				sreg.SidecarService = sidecarService
 				for _, check := range checks {
 					if check.ServiceID == sidecarService.ID {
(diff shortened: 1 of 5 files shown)


command/agent/consul/service_client.go
```diff
@@ -1760,6 +1760,15 @@ func (c *ServiceClient) AllocRegistrations(allocID string) (*serviceregistration
 					sreg.Checks = append(sreg.Checks, check)
 				}
 			}
+
+			if sidecarService := services[serviceID+sidecarSuffix]; sidecarService != nil {
```
gulducat, line 1764: this feels like a bit of a mouthful, so I went to double-check it, and I found this thing:

https://github.com/hashicorp/nomad/blob/d9c9dc548a6037d9e71acdb8b717fd8784a1f581/command/agent/consul/service_client.go#L2123-L2130

nit-picky, but I think I prefer having the magic-string key encapsulated in that one spot:

```suggestion
			if sidecarService := getNomadSidecar(serviceID, services); sidecarService != nil {
```

gulducat (approved):
one lil suggest, but either way, LGTM!

command/agent/consul/service_client.go
tgross: Good catch :+1:

tgross:
I'll add the backport labels on this once 1.7.0 GA goes out.

PR #19335, opened 2023-12-06
docs: expand on Sentinel policy reference

tgross:
We have a tutorial already, but no reference documentation for the objects available. We can probably remove the last section of the tutorial when this gets deployed.


lgfa29 (approved):
Nice! I think we could also add a link to there from the Enterprise overview page:
https://github.com/hashicorp/nomad/blob/9fe1f0aadc1f364ceefbeb064653c0d2062514db/website/content/docs/enterprise/index.mdx#L135-L146

Issue #16607, opened 2023-03-22
E2E: isolation test failure


tgross:
The E2E isolation test appears to be failing during test setup:

```
=== RUN   TestChrootFS
=== RUN   TestChrootFS/testTaskEnvChroot
    utils.go:246: 
        	Error Trace:	/home/circleci/go/src/github.com/hashicorp/nomad/e2e/isolation/utils.go:246
        	            				/home/circleci/go/src/github.com/hashicorp/nomad/e2e/isolation/wait.go:64
        	            				/home/circleci/go/src/github.com/hashicorp/nomad/e2e/isolation/utils.go:228
        	            				/home/circleci/go/src/github.com/hashicorp/nomad/e2e/isolation/utils.go:221
        	            				/home/circleci/go/src/github.com/hashicorp/nomad/e2e/isolation/chroot_test.go:36
        	Error:      	Received unexpected error:
        	            	expected stopped alloc, but was: pending
        	Test:       	TestChrootFS/testTaskEnvChroot
        	Messages:   	failed to wait on alloc
--- FAIL: TestChrootFS/testTaskEnvChroot (57.63s)
=== RUN   TestChrootFS/testTaskImageChroot
    utils.go:246: 
        	Error Trace:	/home/circleci/go/src/github.com/hashicorp/nomad/e2e/isolation/utils.go:246
        	            				/home/circleci/go/src/github.com/hashicorp/nomad/e2e/isolation/wait.go:64
        	            				/home/circleci/go/src/github.com/hashicorp/nomad/e2e/isolation/utils.go:228
        	            				/home/circleci/go/src/github.com/hashicorp/nomad/e2e/isolation/utils.go:221
        	            				/home/circleci/go/src/github.com/hashicorp/nomad/e2e/isolation/chroot_test.go:60
        	Error:      	Received unexpected error:
        	            	expected stopped alloc, but was: pending
        	Test:       	TestChrootFS/testTaskImageChroot
        	Messages:   	failed to wait on alloc
--- FAIL: TestChrootFS/testTaskImageChroot (57.52s)
=== RUN   TestChrootFS/testDownloadChrootExec
    utils.go:246: 
        	Error Trace:	/home/circleci/go/src/github.com/hashicorp/nomad/e2e/isolation/utils.go:246
        	            				/home/circleci/go/src/github.com/hashicorp/nomad/e2e/isolation/wait.go:64
[...]

PR #19337, opened 2023-12-06
e2e: more occasionally slow exec tasks in isolation test

gulducat:
couple more timeout extensions for #16607


tgross (approved):
LGTM :+1:

Issue #19340, opened 2023-12-07
RSA key lost on leader election (1.7.0)


sbihel:
### Nomad version
`v1.7.0-rc.1`

### Operating system and Environment details
Ubuntu `22.04`
Consul `v1.17.0`

### Issue
I am deploying a fresh cluster and setting up workload identities for Consul and Vault. Bootstrapped the cluster with Nomad `v1.7.0-beta.2` but now all the nodes are using `v1.7.0-rc.1`.

By default, the root keyring contained a single ed25519 key. I then triggered a full rotation to generate an RSA key, but then encountered a few issues:
1. with multiple keys in the keyring, it appears Nomad picks one at random (?), which results in many `failed to verify id token signature` from Consul;
2. the old key never gets out of the `rekeying` state, even though I don't appear to use variables; and
3. after replacing the servers (hoping to address 2.), the new key is now an ed25519 key even though it kept the same UUID.

### Reproduction steps
Deploy a fresh cluster using the latest Nomad version. Perhaps using `v1.7.0-beta.2` at first caused the server to enter a bad state -- I can try to re-bootstrap the cluster if you think this might be the root cause.

#### Expected Result
1. Root key rotation is applied permanently.
2. Root key on a fresh cluster should probably be RS256 directly.

#### Actual Result
1. Root keyring rotation does not stick.
2. Rekeying never stops.

### Nomad Server logs (if appropriate)
When a new server joins the cluster, the only relevant logs appear to be:
```
[WARN]  nomad.keyring.replicator: failed to fetch key from current leader, trying peers: key=128ba7c1-baa0-3bc6-c20f-833b97a1fbe2 error=<nil>
[ERROR] nomad.keyring.replicator: failed to fetch key from any peer: key=128ba7c1-baa0-3bc6-c20f-833b97a1fbe2 error="rpc error: no such key \"128ba7c1-baa0-3bc6-c20f-833b97a1fbe2\" in keyring"
[ERROR] nomad.keyring.replicator: failed to fetch key from any peer: rpc error: no such key "128ba7c1-baa0-3bc6-c20f-833b97a1fbe2" in keyring: key=128ba7c1-baa0-3bc6-c20f-833b97a1fbe2
```

But the `128ba7c1` key was long removed. (I had previously done a non-`-full` rotation.)

### Nomad Client logs (if appropriate)
Error logs related to the id token verification by Consul:
```
msg=
[...]

PR #19350, opened 2023-12-07
keyring: replicate RSA private key via `GetKey` RPC

tgross:
When we added a RSA key for signing Workload Identities, we added it to the keystore serialization but did not also add it to the `GetKey` RPC. This means that when a key is rotated, the RSA key will not come along. The Nomad leader signs all Workload Identities, but external consumers of WI (like Consul or Vault) will verify the WI against any of the servers. If the request to verify hits a follower, the follower will not have the RSA private key and cannot use the existing ed25519 key to verify WIs with the `RS256` algorithm.

Add the RSA key material to the `GetKey` RPC.

Also remove an extraneous write to disk that happens for each key each time we restart the Nomad server.

Fixes: #19340


schmichael (approved):
Oof. Great catch.

PR #19356, opened 2023-12-07
e2e: aws ssm usage

gulducat:
As suggested in hashicorp/nomad-e2e#137 which enabled AWS SSM, this documents using it for shell access to E2E instances, and includes convenience Terraform output so folks don't need to hunt for instance IDs.

Example TF output:

```
or, with aws ssm:

    export AWS_REGION=us-east-1
    aws ssm start-session --target i-03897aacabe76119e  # server 44.203.202.18 ip-172-31-90-42
    aws ssm start-session --target i-006a55108211c3982  # server 3.94.210.232 ip-172-31-82-79
    aws ssm start-session --target i-0b1eac90a6301e32c  # server 18.205.185.2 ip-172-31-90-149
    aws ssm start-session --target i-0159d92e90d80e748  # linux client 100.26.201.240 ip-172-31-89-217
```

The `ip-xxx-xx-xx-xxx` are hostnames, which becomes Names on client nodes (`nomad node status`) and Nodes on servers (`nomad operator raft list-peers`), which I thought was nice to have if you want to get to a particular client, or raft leader.


shoenig (approved):
LGTM!

PR #19375, opened 2023-12-07
Revert "deps: update go-metrics to v0.5.3 in 1.6.x"

lgfa29:
`hashicorp/raft` still uses `armon/go-metrics` which results in the incorrect global metric sink variable being used and metrics not being reported.

`go-metrics` can only be upgraded once all dependencies are also updated to use the new path.


tgross (approved):
LGTM

PR #19377, opened 2023-12-08
[ui] Two small UI quality of life changes

philrenaud:
- Disables the "Download as .nomad.hcl" button when running a job through the UI until the template has content (otherwise, you could download a blank file)
- Removes the # of groups column from the jobs index page table


juliezzhou (approved):
lgtm!

Issue #19372, opened 2023-12-07
Panic when running Nomad inside of Proxmox LXC container


codingJWilliams:
### Nomad version

Nomad v1.7.0
BuildDate 2023-12-07T08:28:54Z
Revision e4150e9703f3be6ee2339f0e45ff0801186e022b

### Operating system and Environment details

```bash
root@services01:~# uname -a
Linux services01 6.5.11-6-pve #1 SMP PREEMPT_DYNAMIC PMX 6.5.11-6 (2023-11-29T08:32Z) x86_64 GNU/Linux
root@services01:~# cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

### Issue

When attempting to install Nomad inside of an LXC container hosted on Proxmox: VE, I receive a `panic:` error relating to numalib, I assume it's unable to fetch some data about the CPU. 

### Reproduction steps

1. Install Proxmox: VE following all default options.
2. Create LXC container with latest Debian 12
3. Install Hashicorp repo & GPG key
4. Install nomad with apt
5. `nomad agent -dev`

#### Expected Result

Nomad starts as expected

#### Actual Result

```
root@services01:~# nomad agent -dev
==> No configuration files loaded
==> Starting Nomad agent...
panic: runtime error: index out of range [2] with length 2

goroutine 1 [running]:
github.com/hashicorp/nomad/client/lib/numalib.(*Topology).insert(...)
        github.com/hashicorp/nomad/client/lib/numalib/topology.go:156
github.com/hashicorp/nomad/client/lib/numalib.(*Sysfs).discoverCores.func2.1(0x2)
        github.com/hashicorp/nomad/client/lib/numalib/detect_linux.go:128 +0x233
github.com/hashicorp/nomad/client/lib/idset.(*Set[...]).ForEach(0xc000be7230?, 0xc000b9eba0?)
        github.com/hashicorp/nomad/client/lib/idset/idset.go:181 +0x18c
github.com/hashicorp/nomad/client/lib/numalib.(*Sysfs).discoverCores.func2(0x0)
[...]

PR #19383, opened 2023-12-08
numa: account for incorrect core number on topology.insert

pkazmierczak:
Unsupported environments like containers or guests OSs inside LXD can incorrectly number of available cores thus leading to numalib having trouble detecting cores and panicking. This code adds tests for linux sysfs detection methods and fixes the panic from #19372.

### DIFF
--- a/client/lib/numalib/detect_linux.go
+++ b/client/lib/numalib/detect_linux.go
@@ -38,6 +38,10 @@ const (
 	cpuSiblingFile = sysRoot + "/cpu/cpu%d/topology/thread_siblings_list"
 )
 
+// pathReaderFn is a path reader function, injected into all value getters to
+// ease testing.
+type pathReaderFn func(string) ([]byte, error)
+
 // Sysfs implements SystemScanner for Linux by reading system topology data
 // from /sys/devices/system. This is the best source of truth on Linux and
 // should always be used first - additional scanners can provide more context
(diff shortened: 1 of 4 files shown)


client/lib/numalib/detect_linux_test.go
```diff
@@ -0,0 +1,203 @@
...
+	"os"
+	"testing"
+
+	"github.com/hashicorp/nomad/client/lib/idset"
+	"github.com/hashicorp/nomad/client/lib/numalib/hw"
+	"github.com/shoenig/test/must"
+)
+
+// badSysData are example values from sysfs on unsupported platforms, e.g.,
+// containers, virtualization guests
```
tgross, line 18: There's a good bit here, so it'd be helpful if there were a comment alongside the lines that are different indicating what's unexpected.

client/lib/numalib/detect_linux.go
```diff
@@ -105,46 +109,50 @@ func (*Sysfs) discoverCores(st *Topology) {
...
 			if err != nil {
 				return err
 			}
 
 			cores := idset.Parse[hw.CoreID](string(s))
 			_ = cores.ForEach(func(core hw.CoreID) error {
+				// if we get incorrect core number, stop
+				if int(core) >= len(st.Cores) {
+					return nil
+				}
```
tgross, line 130: This should stop the panic. But does it leave us in an inconsistent state elsewhere?

client/lib/numalib/detect_linux_test.go
pkazmierczak: I added extra comments

tgross (approved):
LGTM

Issue #19389, opened 2023-12-08
docs: add more prominent warnings around backing up keyring


tgross:
This has burned some folks, including an internal team. We have it in the docs but it needs to be "louder".

PR #19400, opened 2023-12-08
docs: add warnings about backing up keyring to snapshot commands

tgross:
The `operator snapshot` commands and agent don't back up Nomad's key material. Add some warnings about this to places where users might be looking for information on cluster recovery.

Fixes: https://github.com/hashicorp/nomad/issues/19389
cc @sofixa


gulducat (approved):
LGTM, hope it's "louder" enough!

PR #19402, opened 2023-12-08
Adds Namespace UI to Access Control

mikenomitch:
Adds Namespace UI to Access Control

Also adds two step buttons to Roles, Policies, and Tokens deletion, and improved error messages on those endpoints.

### DIFF
--- a/ui/app/abilities/namespace.js
+++ b/ui/app/abilities/namespace.js
@@ -1,0 +1,14 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+import AbstractAbility from './abstract';
+import { alias } from '@ember/object/computed';
+
+export default class Namespace extends AbstractAbility {
+  @alias('selfTokenIsManagement') canList;
+  @alias('selfTokenIsManagement') canUpdate;
+  @alias('selfTokenIsManagement') canWrite;
+  @alias('selfTokenIsManagement') canDestroy;
+}
--- a/ui/app/abilities/node-pool.js
+++ b/ui/app/abilities/node-pool.js
@@ -1,0 +1,20 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+import AbstractAbility from './abstract';
+import { alias, and } from '@ember/object/computed';
+import { computed } from '@ember/object';
+
+export default class NodePool extends AbstractAbility {
+  @alias('hasFeatureAndManagement') canConfigureInNamespace;
+
+  @and('nodePoolGovernanceIsPresent', 'selfTokenIsManagement')
+  hasFeatureAndManagement;
+
+  @computed('features.[]')
+  get nodePoolGovernanceIsPresent() {
+    return this.featureIsPresent('Node Pools Governance');
+  }
+}
--- a/ui/app/adapters/namespace.js
+++ b/ui/app/adapters/namespace.js
@@ -17,4 +17,12 @@ export default class NamespaceAdapter extends Watchable {
       }
     });
   }
+
+  urlForCreateRecord(_modelName, model) {
+    return this.urlForUpdateRecord(model.attr('name'), 'namespace');
+  }
+
+  urlForDeleteRecord(id) {
+    return this.urlForUpdateRecord(id, 'namespace');
+  }
 }
--- a/ui/app/components/namespace-editor.hbs
+++ b/ui/app/components/namespace-editor.hbs
@@ -1,0 +1,57 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: BUSL-1.1
+~}}
+
+<form class="acl-form namespace-editor" autocomplete="off" {{on "submit" this.save}}>
+	{{#if this.namespace.isNew }}
+		<Hds::Form::TextInput::Field
+			@isRequired={{true}}
+			data-test-namespace-name-input
+			@value={{@namespace.name}}
+			{{on "input" this.updateNamespaceName}}
+			{{autofocus ignore=(not @namespace.isNew)}}
+		as |F|>
+			<F.Label>Name</F.Label>
+		</Hds::Form::TextInput::Field>
+	{{/if}}
+
+	<div class="boxed-section">
+		<div class="boxed-section-head">
+			Definition
+		</div>
+		<div class="boxed-section-body is-full-bleed">
+			<div
+				class="editor-wrapper boxed-section-body is-full-bleed
+				{{if this.JSONError 'error'}}"
+			>
+				<div
+					class="namespace-editor"
+					data-test-namespace-editor
+					{{code-mirror
+						screenReaderLabel="Namespace definition"
+						theme="hashi"
+						mode="javascript"
+						content=this.definitionString
+						onUpdate=this.updateNamespaceDefinition
+						autofocus=false
+					}}
+				/>
+				{{#if this.JSONError}}
+				<p class="help is-danger">
+					{{this.JSONError}}
+				</p>
+				{{/if}}
+			</div>
+		</div>
+	</div>
+
+	<footer>
+		{{#if (can "update namespace")}}
+      <Hds::Button @text="Save Namespace" @color="primary" disabled={{this.JSONError}}
+        {{on "click" this.save}}
+		data-test-save-namespace
+      />
+		{{/if}}
+	</footer>
+</form>
--- a/ui/app/components/namespace-editor.hbs
+++ b/ui/app/components/namespace-editor.hbs
@@ -22,7 +22,7 @@
 		</div>
 		<div class="boxed-section-body is-full-bleed">
 			<div
-				class="editor-wrapper boxed-section-body is-full-bleed
+				class="namespace-editor-wrapper boxed-section-body is-full-bleed
 				{{if this.JSONError 'error'}}"
 			>
 				<div
@@ -35,6 +35,7 @@
 						content=this.definitionString
 						onUpdate=this.updateNamespaceDefinition
 						autofocus=false
+						extraKeys=(hash Cmd-Enter=this.save)
 					}}
 				/>
 				{{#if this.JSONError}}
--- a/ui/app/components/namespace-editor.js
+++ b/ui/app/components/namespace-editor.js
@@ -83,9 +83,9 @@ export default class NamespaceEditorComponent extends Component {
         );
       }
     } catch (err) {
-      let title = this.namespace.isNew
-        ? `Error creating Namespace ${this.namespace.name}`
-        : `Error updating Namespace ${this.namespace.name}`;
+      let title = `Error ${
+        this.namespace.isNew ? 'creating' : 'updating'
+      } Namespace ${this.namespace.name}`;
 
       let message = err.errors?.length
         ? messageFromAdapterError(err)
--- a/ui/app/components/namespace-editor.js
+++ b/ui/app/components/namespace-editor.js
@@ -1,0 +1,157 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+// @ts-check
+
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+import { alias } from '@ember/object/computed';
+import { tracked } from '@glimmer/tracking';
+import Component from '@glimmer/component';
+import messageFromAdapterError from 'nomad-ui/utils/message-from-adapter-error';
+
+export default class NamespaceEditorComponent extends Component {
+  @service notifications;
+  @service router;
+  @service store;
+  @service can;
+
+  @alias('args.namespace') namespace;
+
+  @tracked JSONError = null;
+  @tracked definitionString = this.definitionStringFromNamespace(
+    this.args.namespace
+  );
+
+  @action updateNamespaceName({ target: { value } }) {
+    this.namespace.set('name', value);
+  }
+
+  @action updateNamespaceDefinition(value) {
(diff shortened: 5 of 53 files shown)


ui/app/abilities/namespace.js
```diff
@@ -0,0 +1,24 @@
...
+  @alias('selfTokenIsManagement') canWrite;
+  @alias('selfTokenIsManagement') canUpdate;
+  @alias('selfTokenIsManagement') canDestroy;
+
+  @and('nodePoolGovernanceIsPresent', 'selfTokenIsManagement')
+  canConfigureNodePools;
+
+  @computed('features.[]')
+  get nodePoolGovernanceIsPresent() {
+    return this.featureIsPresent('Node Pools Governance');
```
mikenomitch, line 22: Note: this string was taken from nomad-licensing library

ui/app/adapters/namespace.js
```diff
@@ -4,17 +4,15 @@
  */
 
 import Watchable from './watchable';
-import codesForError from '../utils/codes-for-error';
 import classic from 'ember-classic-decorator';
 
 @classic
 export default class NamespaceAdapter extends Watchable {
-  findRecord(store, modelClass, id) {
```
mikenomitch, line 12: Oh whoops - should not have removed I think?

ui/app/components/policy-editor.js
```diff
@@ -63,7 +64,14 @@ export default class PolicyEditorComponent extends Component {
           this.policy.id
         );
       }
-    } catch (error) {
+    } catch (err) {
+      let error = 'Error saving policy';
```
mikenomitch, line 68: I reuse this pattern a few places - Is it worth adding a helper for or extending the existing helper?

ui/app/routes/access-control/namespaces/acl-namespace.js
```diff
@@ -0,0 +1,32 @@
...
+  async model(params) {
+    let namespace = await this.store.findRecord(
+      'namespace',
+      decodeURIComponent(params.name),
+      {
+        reload: true,
+      }
+    );
+
+    return hash({
```
mikenomitch, line 28: Should I de-hash this?

ui/app/abilities/node-pool.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+import AbstractAbility from './abstract';
+import { alias, and } from '@ember/object/computed';
+import { computed } from '@ember/object';
+
+export default class NodePool extends AbstractAbility {
```
mikenomitch, line 10: Remove extras

ui/app/controllers/access-control/namespaces/index.js
```diff
@@ -0,0 +1,39 @@
...
+      {
+        key: 'name',
+        label: 'Name',
+        isSortable: true,
+      },
+      {
+        key: 'description',
+        label: 'Description',
+      },
+    ];
```
philrenaud, line 37: I was getting confused about this column being here, but not having a dedicated "description" field, until I realized that (unlike policies), namespaces have their description field along the rest of their definition:
image
vs a policy:
image


Is there something specific about the shape of namespaces vs policies that would mandate this difference in the UI?

both their CLI apply commands use a `-description` flag that takes a string, so I kind of expect their web UI interface would be the same, too.

ui/app/components/namespace-editor.hbs
```diff
@@ -0,0 +1,57 @@
...
+				<div
+					class="namespace-editor"
+					data-test-namespace-editor
+					{{code-mirror
+						screenReaderLabel="Namespace definition"
+						theme="hashi"
+						mode="javascript"
+						content=this.definitionString
+						onUpdate=this.updateNamespaceDefinition
+						autofocus=false
```
philrenaud, line 37: Adding `extraKeys=(hash Cmd-Enter=this.save)` gives this a nice keyboardy feel, IMO!

ui/app/components/policy-editor.js
```diff
@@ -63,10 +64,14 @@ export default class PolicyEditorComponent extends Component {
           this.policy.id
         );
       }
-    } catch (error) {
+    } catch (err) {
+      let message = err.errors?.length
+        ? messageFromAdapterError(err)
+        : err.message || 'Unknown Error';
```
philrenaud, line 70: Thanks!

ui/app/components/namespace-editor.js
```diff
@@ -0,0 +1,157 @@
...
+      if (shouldRedirectAfterSave) {
+        this.router.transitionTo(
+          'access-control.namespaces.acl-namespace',
+          this.namespace.name
+        );
+      }
+    } catch (err) {
+      let title = this.namespace.isNew
+        ? `Error creating Namespace ${this.namespace.name}`
+        : `Error updating Namespace ${this.namespace.name}`;
```
philrenaud, line 88:
```suggestion
      let title = `Error ${this.namespace.isNew ? "creating" : "updating"} Namespace ${this.namespace.name}`
```

ui/app/components/namespace-editor.js
```diff
@@ -0,0 +1,157 @@
...
+        npConfig
+      );
+
+      this.namespace.set('nodePoolConfiguration', nodePoolConfiguration);
+    }
+
+    if (this.can.can('configure-in-namespace quota')) {
+      this.namespace.set('quota', definitionHash['Quota']);
+    }
+  }
```
philrenaud, line 156: Since we have these as separate fragments etc., do we want to include Quotas, "Allowed rules", "Disallowed rules", etc. as columns on the namespace index?

ui/app/models/ns-capabilities.js
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+import Fragment from 'ember-data-model-fragments/fragment';
+import { array } from 'ember-data-model-fragments/attributes';
+
+export default class NamespaceCapabilities extends Fragment {
+  @array('string') DisabledTaskDrivers;
+  @array('string') EnabledTaskDrivers;
```
philrenaud, line 11: These in particular would be cool to see spelled out on pages like namespaces index

ui/app/routes/access-control/namespaces/new.js
```diff
@@ -0,0 +1,56 @@
...
+      quota: '',
+      nodePoolConfiguration: defaultNodePoolConfig,
+    });
+  }
+
+  resetController(controller, isExiting) {
+    if (isExiting) {
+      // If user didn't save, delete the freshly created model
+      if (controller.model.isNew) {
+        controller.model.destroyRecord();
```
philrenaud, line 52: I think we probably also want to `controller.model.nodePoolConfiguration.destroyRecord()` and `controller.model.capabilities.destroyRecord()` here before destroying the model, otherwise we'll get leaky orphans.
image

ui/app/styles/components/access-control.scss
```diff
@@ -85,6 +85,26 @@
...
+      .CodeMirror {
+        box-shadow: 0 0 0 3px $red;
+      }
+      .help {
+        padding: 1rem 0;
+        font-size: 1rem;
+      }
+    }
+  }
+}
```
philrenaud, line 101: Suggest renaming to namespace-editor-wrapper to disambiguate from (my fault, also vaguely named) variables editor wrapper on greps.

ui/app/styles/components/codemirror.scss
```diff
@@ -13,6 +13,12 @@ $dark-bright: lighten($dark, 15%);
   min-height: 500px;
 }
 
+.namespace-editor {
+  .CodeMirror-scroll {
+    min-height: 170px;
```
philrenaud, line 18: Default for this is 500px; did it feel too large? Policies are typically longer in config than namespaces, so is the 170px argument just that they need less space?

ui/app/templates/access-control/namespaces/acl-namespace.hbs
```diff
@@ -0,0 +1,37 @@
...
+		{{/if}}
+	</h1>
+
+<Hds::Alert @type="inline" @color="highlight" @icon="info" class="related-entities notification" as |A|>
+  <A.Title>Related Resources</A.Title>
+  <A.Description>
+    View this namespace's <<Hds::Link::Inline @route="jobs" @query={{hash namespace=this.model.name}}>jobs</<Hds::Link::Inline>
+      or <<Hds::Link::Inline @route="variables" @query={{hash namespace=this.model.name}}>variables</<Hds::Link::Inline>.
+  </A.Description>
+</Hds::Alert>
```
philrenaud, line 34: If you wanted to, we could actually link the variables/jobs and give direct links to an example subset of them. We do this for `job/$id/variables`: https://github.com/hashicorp/nomad/blob/a9d61ea3fd9ecb51d64534b30fe5938024336787/ui/app/routes/jobs/job/variables.js#L38-L50

Just a nice-to-have though. Still useful with the general links.

ui/app/templates/access-control/policies/policy.hbs
```diff
@@ -12,9 +12,16 @@
 		</div>
 		{{#if (can "destroy policy")}}
 			<div>
-        <Hds::Button @text="Delete Policy" @color="critical"
+        <TwoStepButton
```
philrenaud, line 15: Thank you!

ui/app/utils/rollback-without-changed-attrs.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+export default function rollbackWithoutChangedAttrs(model) {
```
philrenaud, line 6: I spent awhile trying to figure out whether this is idiomatic or not and decided that it's actually kind of nice to take explicit control back over stuff like this.

It's annoying to write, but it's well explained and easily understood. It's a good solution!

philrenaud (approved):
Thrilled with this, @mikenomitch ! Great work and can't wait to use it on my own cluster. Some small notes, mostly questions, but I'm very happy with how this feels.

ui/app/controllers/access-control/namespaces/index.js
mikenomitch: I included it in the body because people might have existing json definitions for namespace that have the field, and I think in policy it is external to the file. So if you were using a json file, you could just paste it in here and be good to go.

I agree its a bit weird to be different though.

ui/app/models/ns-capabilities.js
mikenomitch: Okay if I leave out of this PR to cut it off/get it shipped?

ui/app/styles/components/access-control.scss
mikenomitch: sgtm!

ui/app/styles/components/codemirror.scss
mikenomitch: Yeah, just picked a smaller number based on what felt good. - Any gut reaction while you were using this? If not, then I think that's a good sign?

ui/app/templates/access-control/namespaces/acl-namespace.hbs
mikenomitch: Not a bad idea, but I def would want to merge then add it later if we want to do it.

ui/app/utils/rollback-without-changed-attrs.js
mikenomitch: :) Yay

ui/app/models/ns-capabilities.js
philrenaud: yeah 100%!

ui/app/styles/components/codemirror.scss
philrenaud: It seemed small to me but I don't write namespaces very often so don't take my word for it

ui/mirage/config.js
```diff
@@ -731,10 +731,6 @@ export default function () {
     });
   });
 
-  this.get('/namespaces', function ({ namespaces }, req) {
```
mikenomitch: This was overriding an existing call which had some custom logic that handled default ns. That is why those tests were failing I think.

mikenomitch: Damn... this wasn't it

ui/tests/acceptance/token-test.js
```diff
@@ -856,7 +857,9 @@ module('Acceptance | tokens', function (hooks) {
       await visit('/jobs');
       // Expect the Run button/link to work now
       assert.dom('[data-test-run-job]').hasTagName('a');
-      assert.dom('[data-test-run-job]').hasAttribute('href', '/ui/jobs/run');
+      assert
+        .dom('[data-test-run-job]')
+        .hasAttribute('href', '/ui/jobs/run?namespace=*');
```
mikenomitch, line 862: @philrenaud this is one of the instances where I'm not sure why we need this now (called out in the UI sync doc)

philrenaud: Dug into this a bit and I think I have an answer:

[this check](https://github.com/hashicorp/nomad/blob/a9d61ea3fd9ecb51d64534b30fe5938024336787/ui/app/controllers/jobs/index.js#L198-L203) in the jobs index controller sets your namespace to "*" in case your queryParameter has been deleted server-side. So say you had previously bookmarked `/jobs?namespace=myShortlivedNamespace` and it had subsequently been deleted — in order to prevent an empty state, it'd reset you to `*`.

Problem is, the conditional doesn't check for the existence of a namespace queryParam, so it would look to see if `null` was an existing namespace, and since it isn't, it would run the default-namespace-query-param logic for it.

I'm going to add an additional condition to the check and remove the changes to the href/url checks in these tests.

Issue #19405, opened 2023-12-09
worker: failed to dequeue evaluation: error="rpc error: Permission denied" after upgrade to 1.7.1


kevinschoonover:
### Nomad version
```
> nomad version
Nomad v1.7.1
BuildDate 2023-12-08T18:11:21Z
Revision 608e719430038cdeb5fe108536d90cf88a8540e3
```

### Operating system and Environment details
```
> uname -a
Linux 6.1.0-9-cloud-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.27-1 (2023-05-08) x86_64 GNU/Linux
```

### Issue
After upgrading to 1.7.1, I begin seeing the following with a fresh nomad server cluster with ACL enabled and bootstrapped

`Dec 09 07:40:48 vps-5f87f092 nomad[1273434]: worker: failed to dequeue evaluation: worker_id=124ce3e0-3840-b1b0-6cef-fcf99619d0c9 error="rpc error: Permission denied"`

restarting the nomad server does not fix the issue, but downgrading to 1.6.4 the problem goes away when the leader that is elected is a 1.6.4 node.

In the logs below you can see it affecting the nomad server even when its 1.6.4 until the leader is re-elected to another 1.6.4 node
```
Dec 09 07:43:40 vps-5f87f092 nomad[1273612]: ==> Nomad agent configuration:
Dec 09 07:43:40 vps-5f87f092 nomad[1273612]:        Advertise Addrs: HTTP: 100.91.63.45:4646; RPC: 100.91.63.45:4647; Serf: 100.91.63.45:4648
Dec 09 07:43:40 vps-5f87f092 nomad[1273612]:             Bind Addrs: HTTP: [100.91.63.45:4646 127.0.0.1:4646]; RPC: 100.91.63.45:4647; Serf: 100.91.63.45:4648
Dec 09 07:43:40 vps-5f87f092 nomad[1273612]:                 Client: false
Dec 09 07:43:40 vps-5f87f092 nomad[1273612]:              Log Level: INFO
Dec 09 07:43:40 vps-5f87f092 nomad[1273612]:                Node Id: 1878ea3d-1a86-bdef-6ff9-e52d31747022
Dec 09 07:43:40 vps-5f87f092 nomad[1273612]:                 Region: global (DC: ovh-us-west-or-fed)
Dec 09 07:43:40 vps-5f87f092 nomad[1273612]:                 Server: true
[...]

PR #19425, opened 2023-12-11
auth: respect default `tls.verify_server_hostname=false`

tgross:
In Nomad 1.7.0, we refactored much of the authentication code to eliminate nil ACLs and create "virtual" ACL objects that can be used to reduce the risk of fail-open security bugs. In doing so, we accidentally dropped support for the default `tls.verify_server_hostname=false` option.

Fix the bug by including the field in the set of conditions we check for the TLS argument we pass into the constructor (this keeps "policy" separate from "mechanism" in the auth code and reduces the number of dimensions we need to test). Change the field name in the Authenticator to better match the intent.

Fixes: https://github.com/hashicorp/nomad/issues/19405

### DIFF
--- a/.changelog/19425.txt
+++ b/.changelog/19425.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+auth: Fixed a bug where `tls.verify_server_hostname=false` was not respected, leading to authentication failures between Nomad agents
+```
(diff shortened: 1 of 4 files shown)


tgross:
I'm going to do some quick end-to-end testing of this as well. Will post results here.

shoenig (approved):
LGTM!

PR #19432, opened 2023-12-11
[ui] Fix a small regression where accessor is hidden by default when editing a token

philrenaud:
When editing or viewing a newly created token, we want the Accessor to be visible (and the secret masked).

The [Helios masked input component](https://helios.hashicorp.design/components/form/masked-input?tab=code#formmaskedinputfield-1) recently changed the property for this from `isMasked` to `isContentMasked`, so updating it here to follow up our [Helios upgrade](https://github.com/hashicorp/nomad/pull/19247):

image


tgross (approved):
LGTM!

Issue #19381, opened 2023-12-08
e2e: flaky docker artifact download test


shoenig:
Seen this test fail in the same way a few times but it's not consistent. Not sure why yet.

https://github.com/hashicorp/nomad-e2e/actions/runs/7137601211/job/19437898012

```
=== Failed
=== FAIL: e2e/artifact TestArtifact/testLinux/docker_file_custom (1.42s)
    assert.go:14: 
        artifact_test.go:44: expected string to contain substring; it does not
        ↪ substring: module github.com/hashicorp/go-set/v2
        ↪    string:

=== FAIL: e2e/artifact TestArtifact/testLinux (67.77s)
```

Agent logs from the e2e run.

```
➜ rg docker_file_custom
linux_client-44.204.50.146/logs-nomad.log
6257:    2023-12-08T05:42:12.511Z [INFO]  client.alloc_runner.task_runner: Task event: alloc_id=2738f8ce-ed44-c3d7-58c1-c40e36a5d849 task=docker_file_custom type=Received msg="Task received by client" failed=false
6292:    2023-12-08T05:42:12.607Z [DEBUG] client.alloc_runner.task_runner: lifecycle start condition has been met, proceeding: alloc_id=2738f8ce-ed44-c3d7-58c1-c40e36a5d849 task=docker_file_custom
6303:    2023-12-08T05:42:12.749Z [INFO]  client.alloc_runner.task_runner: Task event: alloc_id=2738f8ce-ed44-c3d7-58c1-c40e36a5d849 task=docker_file_custom type="Task Setup" msg="Building Task Directory" failed=false
6339:    2023-12-08T05:42:13.042Z [DEBUG] client.alloc_runner.task_runner.task_hook.logmon: starting plugin: alloc_id=2738f8ce-ed44-c3d7-58c1-c40e36a5d849 task=docker_file_custom path=/usr/local/bin/nomad args=["/usr/local/bin/nomad", "logmon"]
6340:    2023-12-08T05:42:13.048Z [DEBUG] client.alloc_runner.task_runner.task_hook.logmon: plugin started: alloc_id=2738f8ce-ed44-c3d7-58c1-c40e36a5d849 task=docker_file_custom path=/usr/local/bin/nomad pid=19887
[...]

PR #19434, opened 2023-12-11
e2e: unflake testTaskImageChroot

gulducat:
assuming this is similar to #19381, this theoretically fixes

```
=== RUN   TestChrootFS/testTaskImageChroot
    assert.go:14: 
        chroot_test.go:69: expected regexp match
        ↪  regex: (?m:^/local\b)
        ↪ string: /alloc
--- FAIL: TestChrootFS/testTaskImageChroot (3.55s)
```

which is expecting stuff in alloc logs of a very-fast docker job.


shoenig (approved):
LGTM!

PR #19435, opened 2023-12-11
Update auth-methods docs to comply with style guide

mpron:
Lower cased the title and headings in line with our company-wide style since this is being linked in an upcoming blog I was editing.  I also lowercased words such as "Auth Method" and other primitives/components when mentioned in prose - this is in line with our style guide as well where we don't capitalize auth method and we only capitalize components that are SKU/product-like in their separateness/importance.

https://docs.google.com/document/d/1MRvGd6tS5JkIwl_GssbyExkMJqOXKeUE00kSEtFi8m8/edit

Adam Trujilo should be in agreement with changes like this based on our past discussions, but feel free to bring in stake holders if you're not sure about accepting and we can discuss.

### DIFF
--- a/website/content/api-docs/acl/auth-methods.mdx
+++ b/website/content/api-docs/acl/auth-methods.mdx
@@ -1,16 +1,16 @@
 ---
 layout: api
-page_title: ACL Auth Methods - HTTP API
+page_title: ACL auth methods - HTTP API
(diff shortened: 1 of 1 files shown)


tgross (approved):
LGTM

tgross:
The Vercel deployment failed but their link as to why is just bringing me to one of their docs pages?

Issue #19380, opened 2023-12-08
Vault address is ignored in nomad clients on v1.7.0


L-P:
### Nomad version
```
# nomad version
Nomad v1.7.0
BuildDate 2023-12-07T08:28:54Z
Revision e4150e9703f3be6ee2339f0e45ff0801186e022b
```

### Operating system and Environment details
```
# lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 12 (bookworm)
Release:	12
Codename:	bookworm
# uname -a
Linux worker-plane 6.1.0-13-cloud-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.55-1 (2023-09-29) x86_64 GNU/Linux
```

Running on AWS EC2 t3.small.

### Issue
Nomad client attempts to reach local vault agent Vault using `127.0.0.1:8200`,
ignoring its configuration.

Downgrading the client to v1.6.4 fixes the issue.

### Reproduction steps
1. Follow what is now the "legacy" vault setup using a local vault agent
   configured to listen on a socket:
   ```
   vault {
     enabled = true
     address = "unix:///var/run/vault/agent.socket"
     create_from_role = "nomad-cluster"
   }
   ```

#### Expected Result
The value of `address` is used to connect to the vault agent.

#### Actual Result
Another value is used, thus failing to reach the agent.

### Job file (if appropriate)
Uses templates with `{{with secret "<path>"}}`.

### Nomad Server logs (if appropriate)
n/a

### Nomad Client logs (if appropriate)
```
2023-12-08T10:08:32.525Z [INFO]  client.alloc_runner.task_runner: Task event: alloc_id=e3289405-b9fb-3498-290a-0d1a5a44b59a task=fabio type=Killing msg="Template failed: vault.write(pki_int/issue/consul -> 1656cb6c): vault.write(pki_int/issue/consul -> 1656cb6c): Put \"https://127.0.0.1:8200/v1/pki_int/issue/consul\": dial tcp 127.0.0.1:8200: connect: connection refused" failed=true
```

tgross:
Hi @L-P! Unfortunately I wasn't able to reproduce this. I used the following Vault listener config:

```hcl
listener "unix" {
  address = "/home/tim/tmp/vault.sock"
}
```

The following Nomad `vault` config block:

```hcl
vault {
  address = "unix:///home/tim/tmp/vault.sock"
  token = "<redacted>"
  enabled = true
[...]

PR #19439, opened 2023-12-11
vault: load default config for tasks without vault

lgfa29:
It is often expected that a task that needs access to Vault defines a `vault` block to specify the Vault policy to use to derive a token.

But in some scenarios, like when the Nomad client is connected to a local Vault agent that is responsible for authn/authz, the task is not required to defined a `vault` block.

In these situations, the `default` Vault cluster should be used to render the template.

Closes https://github.com/hashicorp/nomad/issues/19380

### DIFF
--- a/client/allocrunner/taskrunner/template_hook_test.go
+++ b/client/allocrunner/taskrunner/template_hook_test.go
@@ -6,8 +6,12 @@ package taskrunner
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"path"
 	"sync"
 	"testing"
+	"time"
 
 	consulapi "github.com/hashicorp/consul/api"
 	"github.com/hashicorp/nomad/ci"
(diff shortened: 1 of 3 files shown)


client/allocrunner/taskrunner/template_hook_test.go
```diff
@@ -135,3 +141,129 @@ func Test_templateHook_Prestart_ConsulWI(t *testing.T) {
...
+			// Start in a goroutine because Prestart() blocks until first
+			// render.
+			hookErrCh := make(chan error)
+			go func() {
+				err := hook.Prestart(ctx, req, nil)
+				hookErrCh <- err
+			}()
+
+			var gotRequest bool
+		LOOP:
```
tgross, line 246: Because the send on `reqCh` is blocking, at first I thought you could probably select over `reqCh` and `ctx` without this loop and check the error afterwards, and avoid the label/break/goto. But this is a nicely elegant way of avoiding blocking 3 whole seconds for the context timeout on error. :+1:

tgross (approved):
LGTM!

client/allocrunner/taskrunner/template_hook_test.go
lgfa29: Yeah, this ensure we exit as soon as `Prestart()` finishes. I think there's a theoretical issue where a test case reads a request from another test case, because `reqCh` is shared while each test case has its own consumer. But since this is a test, and each test case doesn't run in parallel, I think it's fine.

joliver:
It appears there's one small regression in this. Our configuration is spread across multiple files which are merged together using the [Nomad configuration merge behavior](https://developer.hashicorp.com/nomad/docs/configuration#load-order-and-merging)

`00-vault.hcl`:

```
vault {
  address = "https://vault.company.com:8200"
}
```

`25-server.hcl`:
```
vault {
  enabled          = true
  create_from_role = "role-name-here"
}
```

In Vault 1.6.x and up through Vault 1.7.1, the above behavior was working properly. With the introduction of Vault 1.7.2, I believe the merge behavior has changed because of the later file not specifying a Vault address such that a default address is used which then overwrites/hide the value specified in the earlier file.

tgross:
@joliver I've opened https://github.com/hashicorp/nomad/issues/19482 for you

PR #19441, opened 2023-12-12
[CSI] Prevent stage_publish_base_dir from being subdir of mount_dir

### LINKED ISSUE #19440: [CSI] Do not allow stage_publish_base_dir to be a sub-directory of mount_dir
### Nomad version
Any nomad version 1.3.0 or greater

Any OS

### Issue
If a CSI plugin uses `mount_dir = /var/lib/test` and `stage_publish_base_dir = /var/lib/test/publish`, then this line will cause all mount points to be deleted:
https://github.com/hashicorp/nomad/blob/v1.3.0/client/allocrunner/taskrunner/plugin_supervisor_hook.go#L427-L430

### Reproduction steps
Use `mount_dir = /var/lib/test` and  `stage_publish_base_dir = /var/lib/test/publish` for any CSI driver, and the mount point will not resolve properly and data could be lost on CSI Plugin restart.

#### Expected Result
CSI Plugin Stanza validation should fail if `stage_publish_base_dir`  is inside of `mount_dir` 

#### Actual Result
CSI plugin stanza validation succeeds

### Job file (if appropriate)
n/a - CSI plugin and job configuration

n/a

n/a

ggriffiths:
Fixes #19440

### DIFF
--- a/.changelog/19441.txt
+++ b/.changelog/19441.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-csi: Added validation to prevent StagePublishBaseDir from being a subdir MountDir
+csi: Added validation to `csi_plugin` blocks to prevent `stage_publish_base_dir` from being a subdirectory of `mount_dir`
 ```
--- a/.changelog/19441.txt
+++ b/.changelog/19441.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+csi: Added validation to prevent StagePublishBaseDir from being a subdir MountDir
+```
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -7942,6 +7942,11 @@ func (t *Task) Validate(jobType string, tg *TaskGroup) error {
 			mErr.Errors = append(mErr.Errors, fmt.Errorf("CSIPluginConfig PluginType must be one of 'node', 'controller', or 'monolith', got: \"%s\"", t.CSIPluginConfig.Type))
 		}
 
+		if t.CSIPluginConfig.StagePublishBaseDir != "" && t.CSIPluginConfig.MountDir != "" &&
+			strings.HasPrefix(t.CSIPluginConfig.StagePublishBaseDir, t.CSIPluginConfig.MountDir) {
+			mErr.Errors = append(mErr.Errors, fmt.Errorf("CSIPluginConfig StagePublishBaseDir must not be a subdirectory of MountDir, got: StagePublishBaseDir=\"%s\" MountDir=\"%s\"", t.CSIPluginConfig.StagePublishBaseDir, t.CSIPluginConfig.MountDir))
+		}
+
 		// TODO: Investigate validation of the PluginMountDir. Not much we can do apart from check IsAbs until after we understand its execution environment though :(
 	}
 
--- a/website/content/docs/job-specification/csi_plugin.mdx
+++ b/website/content/docs/job-specification/csi_plugin.mdx
@@ -49,7 +49,8 @@ csi_plugin {
 - `stage_publish_base_dir` `(string: <optional>)` - The base directory
   path inside the container where the plugin will be instructed to
   stage and publish volumes. This field is typically not required.
-  Refer to your CSI plugin's documentation for details.
+  Refer to your CSI plugin's documentation for details. This can not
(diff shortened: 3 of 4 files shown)


ggriffiths:
@tgross can you take a look?

jobspec/parse_task.go
```diff
@@ -169,6 +169,10 @@ func parseTask(item *ast.ObjectItem, keys []string) (*api.Task, error) {
 		if err := hcl.DecodeObject(&m, i.Val); err != nil {
 			return nil, err
 		}
+		if cfg.StagePublishBaseDir != "" && cfg.MountDir != "" &&
+			strings.Contains(cfg.StagePublishBaseDir, cfg.MountDir) {
+			return nil, fmt.Errorf("csi_plugin -> Expected stage_publish_base_dir to not be a subdirectory of mount_dir")
+		}
```
tgross, line 175: There's two problems here:
* This isn't the right place to validate this. This package is the HCL1 jobspec parsing. So HCL2 isn't covered and neither are any job registration requests that come in through the API. You'll want to do this when we validate the [`CSIPluginConfig`](https://github.com/hashicorp/nomad/blob/71ea1deda707b706c36da7567caba7721e51db19/nomad/structs/structs.go#L7935-L7946) struct so that any job registration hits this check.
* This only checks if the path string for the staging publish base dir _contains_ the string path for the mount dir, not that they're relative to each other. These could be hypothetically non-overlapping values like `/local/csi` and `/csi` and that would fail.

website/content/docs/job-specification/csi_plugin.mdx
```diff
@@ -49,7 +49,8 @@ csi_plugin {
 - `stage_publish_base_dir` `(string: <optional>)` - The base directory
   path inside the container where the plugin will be instructed to
   stage and publish volumes. This field is typically not required.
-  Refer to your CSI plugin's documentation for details.
+  Refer to your CSI plugin's documentation for details. This can not
+  be a subdirectory of mount_dir.
```
tgross, line 53:
```suggestion
  be a subdirectory of `mount_dir`.
```

tgross (changes requested):
Hi @ggriffiths! Thanks for the PR!

In addition to addressing the comments I've posted, can you run `make cl` to add a changelog entry for the fix once you're done?

jobspec/parse_task.go
ggriffiths: Ah of course - will address this today. Thanks

nomad/structs/structs.go
```diff
@@ -7942,6 +7942,11 @@ func (t *Task) Validate(jobType string, tg *TaskGroup) error {
 			mErr.Errors = append(mErr.Errors, fmt.Errorf("CSIPluginConfig PluginType must be one of 'node', 'controller', or 'monolith', got: \"%s\"", t.CSIPluginConfig.Type))
 		}
 
+		if t.CSIPluginConfig.StagePublishBaseDir != "" && t.CSIPluginConfig.MountDir != "" &&
+			strings.HasPrefix(t.CSIPluginConfig.StagePublishBaseDir, t.CSIPluginConfig.MountDir) {
```
ggriffiths, line 7946: I think this will handle most cases. For example, if `stage_publish_base_dir=/local/csi` and `mount_dir=/local`, then we'll error out. 

However, we don't error out if stage_publish_base_dir=/csi and mount_dir=/local/mount-dir.

.changelog/19441.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+csi: Added validation to prevent StagePublishBaseDir from being a subdir MountDir
```
tgross, line 2: Let's add context that this is for plugins, and use the names from the jobspec, not the internal struct names:

```suggestion
csi: Added validation to `csi_plugin` blocks to prevent `stage_publish_base_dir` from being a subdirectory of `mount_dir`
```

tgross (changes requested):
This looks great @ggriffiths. Just the two minor suggestions left to accept and I think we can merge this.

ggriffiths:
Sounds good. Thanks for the quick review @tgross

tgross (approved):
LGTM

Issue #19442, opened 2023-12-12
Consul constraints don't use non-default Consul cluster (ENT)


SuyashHashiCorp:
I have encountered an issue where Multi-Cluster Consul is not working as expected when using a custom name in conjunction with Nomad version 1.7.0-rc.1+ent.

### Nomad version
Output from `nomad version`
`v1.7.0-rc.1+ent`

### Consul version
Output from `consul version`
`v1.17.0+ent`

### Operating system and Environment details
OS - Linux (Ubuntu) - `amd64` and `arm64` both

### Reproduction steps
- Create two Consul clusters with version v1.17.0+ent.

- Provision Nomad cluster (1 server and 1 client) with version v1.7.0-rc.1+ent. 

- Use the below consul blocks configuration for Nomad Server and Client Nodes -

```
consul {
  address = "10.211.55.11:8500"
  name = "alpha"
}

consul {
  address = "10.211.55.12:8500"
  name = "beta"
}
```

- Start the Nomad services on the Nomad server and client nodes.

- Not able to see the services in the Consul UI Dashboard or consul catalog services in both the consul clusters. 

- In the Nlmad node logs getting the below warning - 

```
Dec 05 12:37:38 c1 nomad[4568]:     2023-12-05T12:37:38.777+0530 [WARN]  client.fingerprint_mgr.consul: malformed Consul response prevented linking
Dec 05 12:37:38 c1 nomad[4568]:     2023-12-05T12:37:38.787+0530 [WARN]  client.fingerprint_mgr.consul: malformed Consul response prevented linking
```

- While checking on CLI using the Nomad command nomad node status -verbose <node_id> | grep -i consul I can see the details of Consul integration with Nomad. Here is the sample snippet for your reference- 

```
$ nomad server members
Name       Address      Port  Status  Leader  Raft Version  Build           Datacenter  Region
s1.global  10.211.55.6  4648  alive   true    3             1.7.0-rc.1+ent  dc1         global
```

```
$ nomad node status
ID        Node Pool  DC   Name  Class   Drain  Eligibility  Status
2f97f440  default    dc1  c1    <none>  false  eligible     ready
```

```
$ nomad node status -verbose 2f97f440 | grep -i consul
consul.alpha.connect             = true
consul.alpha.datacenter          = dc1
consul.alpha.ft.namespaces       = true
consul.alpha.grpc                = 8502
consul.alpha.revision            = 9afb1901
consul.alpha.server              = false
[...]

PR #19449, opened 2023-12-12
consul: fix constraints for non-default clusters (ENT)

tgross:
The Connect-related constraints injected by the Connect job mutating hook do not account for non-default `consul` blocks (for Nomad Enterprise). This works when both the default and non-default clusters are available and are the same version, but not when they do not.

Fixes: https://github.com/hashicorp/nomad/issues/19442

### DIFF
--- a/nomad/job_endpoint_hook_connect.go
+++ b/nomad/job_endpoint_hook_connect.go
@@ -80,40 +80,56 @@ func connectGatewayDriverConfig(hostNetwork bool) map[string]interface{} {
 // the proper Consul version is used that supports the necessary Connect
 // features. This includes bootstrapping envoy with a unix socket for Consul's
 // gRPC xDS API, and support for generating local service identity tokens.
-func connectSidecarVersionConstraint() *structs.Constraint {
-	return &structs.Constraint{
-		LTarget: "${attr.consul.version}",
-		RTarget: ">= 1.8.0",
-		Operand: structs.ConstraintSemver,
+func connectSidecarVersionConstraint(cluster string) *structs.Constraint {
+	if cluster != structs.ConsulDefaultCluster && cluster != "" {
+		return &structs.Constraint{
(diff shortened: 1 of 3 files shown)


nomad/job_endpoint_hook_connect.go
```diff
@@ -80,40 +80,56 @@ func connectGatewayDriverConfig(hostNetwork bool) map[string]interface{} {
 // the proper Consul version is used that supports the necessary Connect
 // features. This includes bootstrapping envoy with a unix socket for Consul's
 // gRPC xDS API, and support for generating local service identity tokens.
-func connectSidecarVersionConstraint() *structs.Constraint {
-	return &structs.Constraint{
-		LTarget: "${attr.consul.version}",
-		RTarget: ">= 1.8.0",
-		Operand: structs.ConstraintSemver,
+func connectSidecarVersionConstraint(cluster string) *structs.Constraint {
+	if cluster != structs.ConsulDefaultCluster && cluster != "" {
```
pkazmierczak, line 84: I understand why we're injecting the `cluster` string into the constraint, but I don't understand the if condition. Won't we want this constraint for default clusters, too? Also, should this be a `&&` and not a `||`?

tgross: We need to return the default constraint (down on line 91) when either the `consul.cluster = "default"` or `consul.cluster = ""` (which is treated as `default`). So this condition is for when the cluster is not explicitly `default` and not empty either.

tgross: For clarity, the constraint for the default cluster doesn't have the cluster name in it. It's just `attr.consul.version` and not `attr.consul.default.version`.

pkazmierczak: Oh I get it now, thanks for clarifying Tim.

pkazmierczak (approved):

PR #19452, opened 2023-12-12
[ui] When Action output is long enough, keep a user scroll-anchored to the bottom

### LINKED ISSUE #19437: [ui] Actions should keep me scrolled to the bottom in the web UI
Currently a many-lined action output will just progress down the page a bit but leave me to manually scroll.

Instead, keep me glued to the bottom unless I scroll away.

image

philrenaud:
Uses overflow-anchor (shoutout to https://css-tricks.com/books/greatest-css-tricks/pin-scrolling-to-bottom/) to keep long-running or tall Actions outputs in the UI scrolled to the bottom.

Importantly, if the user scrolls AWAY from the bottom, let's say to carefully read an earlier log or copy it, the anchoring behaviour ceases (until they scroll back to the bottom again)

image

Resolves #19437

### DIFF
--- a/ui/app/components/action-card.hbs
+++ b/ui/app/components/action-card.hbs
@@ -71,7 +71,12 @@
       <code><pre>Error: {{this.instance.error}}</pre></code>
     {{/if}}
     {{#if this.instance.messages.length}}
-      <code><pre>{{this.instance.messages}}</pre></code>
+      <code tabindex="0">
+        <pre {{did-update this.anchorToBottom this.instance.messages}}>
+          {{this.instance.messages}}
+        </pre>
+        <div class="anchor" />
+      </code>
     {{else}}
       {{#if (eq this.instance.state "complete")}}
         <p class="no-messages">Action completed with no output</p>
--- a/ui/app/styles/components/actions.scss
+++ b/ui/app/styles/components/actions.scss
@@ -117,14 +117,30 @@
       }
 
       .messages {
+        width: 100%;
         overflow: hidden;
 
-        code > pre {
-          height: 200px;
+        code {
           background-color: #0a0a0a;
           color: whitesmoke;
+          display: block;
+          overflow: auto;
+          height: 200px;
           border-radius: 6px;
           resize: vertical;
+          pre {
+            background-color: transparent;
+            color: unset;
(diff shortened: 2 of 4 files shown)


ui/app/styles/components/actions.scss
```diff
@@ -117,14 +117,30 @@
...
+          height: 200px;
           border-radius: 6px;
           resize: vertical;
+          pre {
+            background-color: transparent;
+            color: unset;
+            overflow-anchor: none;
+            min-height: 100%;
+            white-space: pre-wrap;
+          }
```
philrenaud, line 137: Part of this was making the scrollbars appear in the correct spot. An earlier attempt had me creating a scrollable area within the `<pre>` block, but that was unnecessary. Instead, we move the colo(u)r rules to the parent `<code>` element and have the `<pre>` always min-height to the code block's height.

ui/app/components/action-card.hbs
```diff
@@ -71,7 +71,12 @@
       <code><pre>Error: {{this.instance.error}}</pre></code>
     {{/if}}
     {{#if this.instance.messages.length}}
-      <code><pre>{{this.instance.messages}}</pre></code>
+      <code tabindex="0">
```
philrenaud, line 74: Side-effect: now you can focus in the action output and do things like `pgdn` etc.

ui/app/styles/components/actions.scss
```diff
@@ -117,14 +117,30 @@
...
           color: whitesmoke;
+          display: block;
+          overflow: auto;
+          height: 200px;
           border-radius: 6px;
           resize: vertical;
+          pre {
+            background-color: transparent;
+            color: unset;
+            overflow-anchor: none;
```
lgfa29, line 134: Did you have a chance to test on Safari, where `overflow-anchor` doesn't seem to be supported?
image

lgfa29 (approved):
Minor check on Safari support for `overflow-anchor`

ui/app/styles/components/actions.scss
philrenaud: Yep, this won't work in Safari; observed behaviour there is same as pre-PR (treats this like any other scrollable element, not "sticky")

Issue #19414, opened 2023-12-11
namespace: attempting to delete a namespace with attached variables returns 200 OK


jrasell:
### Nomad version
Main at `e551814df5f271a3ad2f69ed4fd812ce2807be84`

### Operating system and Environment details
Reproduced locally on MacOS but does not rely on any OS specific code.

### Issue
The problem looks like a bug in the way the namespace delete RPC validation happens. The handler [checks whether the namespace has non-terminal jobs within](https://github.com/hashicorp/nomad/blob/099ee06a607cea34389f8204b16506d1f24e1afa/nomad/namespace_endpoint.go#L111) it but doesn't check whether the namespace has variables still attached to it.

The RPC handler validation doesn't error, therefore the Raft update message is fired to remove the namespace; this triggers the RPC/HTTP response.

When the Raft message reaches the FSM, we check whether the [namespace has any variables attached](https://github.com/hashicorp/nomad/blob/5e1bbf90fca0a1d7beceba4d99e2a9d70b18e7a1/nomad/state/state_store.go#L6834), which triggers the error in the logs and results in the namespace not being deleted. This bug also seems to be present when attempting to delete a namespace [which still has CSI volumes attached](https://github.com/hashicorp/nomad/blob/5e1bbf90fca0a1d7beceba4d99e2a9d70b18e7a1/nomad/state/state_store.go#L6823).

It's likely we need to duplicate this validation to the RPC handler which will allow us to return correct errors to the external caller.

### Reproduction steps
1. Start a Nomad development agent using `nomad agent -dev`
2. Create a non-default namespace using a command such as `nomad namespace apply jrasell`
3. Create a variable attached to the non-default namespace using the command `echo "abcd1234" | nomad var put -namespace=jrasell secret/foo bar=-`
4. Attempt to delete the namespace using `nomad namespace delete jrasell`

#### Expected Result
The CLI/API should return an error that the namespace could not be deleted because it has attached variables.

#### Actual Result
The CLI/API return successfully:
```
Successfully deleted namespace "jrasell"!
```

The agent logs show an error deleting the namespace:
```
    2023-12-11T07:59:40.627Z [ERROR] nomad.fsm: DeleteNamespaces failed: error="namespace \"jrasell\" contains at least one variable. All variables in namespace must be deleted before it can be deleted"
    2023-12-11T07:59:40.627Z [DEBUG] http: request complete: method=DELETE path=/v1/namespace/jrasell duration="272.75µs"
```

[...]

PR #19483, opened 2023-12-14
Surfaces errors from namespace delete properly

mikenomitch:
Adds a "return err" that was accidentally dropped shortly after the Namespace OSS migration.

fixes https://github.com/hashicorp/nomad/issues/19414

### DIFF
--- a/.changelog/19483.txt
+++ b/.changelog/19483.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+namespaces: Failed delete calls no longer return success codes
+```
--- a/nomad/fsm_test.go
+++ b/nomad/fsm_test.go
@@ -3306,33 +3306,26 @@ func TestFSM_DeleteNamespaces(t *testing.T) {
 }
 
 func TestFSM_DeleteNamespaces_ErrorSurfacing(t *testing.T) {
-	assert := assert.New(t)
 	ci.Parallel(t)
 	fsm := testFSM(t)
 
 	ns1 := mock.Namespace()
 	// force a failure by making this the default
 	ns1.Name = "default"
-	assert.Nil(fsm.State().UpsertNamespaces(1000, []*structs.Namespace{ns1}))
+	must.NoError(t, fsm.State().UpsertNamespaces(1000, []*structs.Namespace{ns1}))
 
 	req := structs.NamespaceDeleteRequest{
 		Namespaces: []string{ns1.Name},
 	}
 
 	buf, err := structs.Encode(structs.NamespaceDeleteRequestType, req)
-	require.NoError(t, err)
-
+	must.NoError(t, err)
 	resp := fsm.Apply(makeLog(buf))
-	if resp == nil {
-		t.Fatalf("no resp: %v", resp)
-	}
+	must.NotNil(t, resp)
+
 	err, ok := resp.(error)
-	if !ok {
-		t.Fatalf("resp not of error type: %T %v", resp, resp)
-	}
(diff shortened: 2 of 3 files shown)


mikenomitch:
Note: I still need to test this manually. I will update this comment once I do.

nomad/fsm_test.go
```diff
@@ -3305,6 +3305,36 @@ func TestFSM_DeleteNamespaces(t *testing.T) {
...
+	if resp == nil {
+		t.Fatalf("no resp: %v", resp)
+	}
+	err, ok := resp.(error)
+	if !ok {
+		t.Fatalf("resp not of error type: %T %v", resp, resp)
+	}
+	if !strings.Contains(err.Error(), "default namespace can not be deleted") {
+		t.Fatalf("bad error: %v", err)
+	}
```
tgross, line 3335: We're idiomatically using `shoenig/test` unless there's a good reason not to. The assertions are typically shorter, and @shoenig has got a bunch of cool generics magic in there that makes the assertions more precise than those in testify. Not everywhere in the code base has been updated to that yet.

```suggestion
	must.NoError(t, err)

	resp := fsm.Apply(makeLog(buf))
	must.NotNil(t, resp)
	
	err, ok := resp.(error)
	must.True(t, ok, must.Sprintf("resp not of error type: %T %v", resp, resp)
	must.ErrorContains(t, err, "default namespace can not be deleted")
```

nomad/fsm_test.go
```diff
@@ -3305,6 +3305,36 @@ func TestFSM_DeleteNamespaces(t *testing.T) {
 	assert.Nil(out)
 }
 
+func TestFSM_DeleteNamespaces_ErrorSurfacing(t *testing.T) {
+	assert := assert.New(t)
```
tgross, line 3309:
```suggestion
```

nomad/fsm_test.go
```diff
@@ -3305,6 +3305,36 @@ func TestFSM_DeleteNamespaces(t *testing.T) {
...
+	ci.Parallel(t)
+	fsm := testFSM(t)
+
+	ns1 := mock.Namespace()
+	// force a failure by making this the default
+	ns1.Name = "default"
+	assert.Nil(fsm.State().UpsertNamespaces(1000, []*structs.Namespace{ns1}))
```
tgross, line 3316:
```suggestion
	must.NoError(t, fsm.State().UpsertNamespaces(1000, []*structs.Namespace{ns1}))
```

mikenomitch:
Have now manually tested this and it works (with a soon-to-come UI change in another PR):

Screenshot 2023-12-14 at 3 21 35 PM

Thanks for the review - on those changes!

nomad/fsm_test.go
mikenomitch: Oooh that is way nicer

.changelog/19483.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+namespaces: Failed delete calls no longer return success codes
```
jrasell, line 2:
```suggestion
namespaces: Fixed a bug where a 200 OK response was returned even when the namespace delete failed
```

jrasell (approved):
LGTM with wording suggestion on the changelog.

If we really wanted to we could also add a test to ensure non-default namespace deletions return an error; but I assume this is also caught by state store testing.

tgross (approved):
LGTM!

PR #19490, opened 2023-12-15
client: consul hook not called for templates

### LINKED ISSUE #19488: Consul token not being exposed to the task with workload identity
### Nomad version
1.7.2

Ubuntu 22.04 running on a virtual machine

### Issue
As stated in [the nomad documentation](https://developer.hashicorp.com/nomad/docs/job-specification/consul#access-to-token) the consul token would be available to the task via `CONSUL_TOKEN` environment variable and the `/secrets/consul_token` file, however those are not injected into the task. Access to Consul KV works with no problems in template blocks, and Vault token is also injected into the task with no problems

### Reproduction steps
Having a Consul cluster with acl enabled and a Nomad cluster with workload identities configured with Consul integration, run a job using consul as stated in the documentation linked above.

#### Expected Result
Environment variable `CONSUL_TOKEN` defined and file `/secrets/consul_token` created

#### Actual Result
Neither `CONSUL_TOKEN` environment variable nor `/secrets/consul_token` file were present

### Job file (if appropriate)
```hcl
job "example" {

  group "echo" {
    network {
      port "echo" {
        to = 5678
      }
    }

    task "echo" {
      driver = "docker"

      config {
        image = "hashicorp/http-echo"
        ports = ["echo"]
        args  = ["-text", "hello world"]
      }

      identity {
        name = "consul_default"
        aud = ["consul.io"]
      }

      consul {}
      
      template {
        data        = <<-EOT
          CONSUL_TOKEN={{env "CONSUL_TOKEN"}}
        EOT
        destination = "local/config.txt"
      }

      resources {
        cpu    = 100
        memory = 64
      }
    }
  }
}
```

pkazmierczak:
Due to some refactoring mishap, task-level Consul hook was never triggered and thus never wrote any secrets in task secret dirs. 

Fixes #19488

### DIFF
--- a/.changelog/19490.txt
+++ b/.changelog/19490.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+client: fix a bug where task level Consul hook never gets called
+```
--- a/.changelog/19490.txt
+++ b/.changelog/19490.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-client: fix a bug where task level Consul hook never gets called
+client: Fixed a bug where where the environment variable / file for the Consul token weren't written.
 ```
--- a/client/allocrunner/taskrunner/consul_hook.go
+++ b/client/allocrunner/taskrunner/consul_hook.go
@@ -18,10 +18,9 @@ import (
 )
 
 const (
-	// consulTokenFilePrefix is the begging of the name of the file holding the
-	// Consul SI token inside the task's secret directory. Full name of the file is
-	// always consulTokenFilePrefix_identityName
-	consulTokenFilePrefix = "nomad_consul"
+	// consulTokenFilename is the name of the file holding the Consul SI token
+	// inside the task's secret directory.
+	consulTokenFilename = "consul_token"
 
 	// consulTokenFilePerms is the level of file permissions granted on the file in
 	// the secrets directory for the task
@@ -57,7 +56,7 @@ func (h *consulHook) Prestart(context.Context, *interfaces.TaskPrestartRequest, 
 	tokens := h.hookResources.GetConsulTokens()
 
 	// Write tokens to tasks' secret dirs
-	for cluster, t := range tokens {
+	for _, t := range tokens {
 		for identity, token := range t {
 			// do not write tokens that do not belong to any of this task's
 			// identities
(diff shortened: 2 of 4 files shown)


.changelog/19490.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+client: fix a bug where task level Consul hook never gets called
```
tgross, line 2: For the changelog, I would word this in terms of the user-facing impact rather than the internal code ("hooks" aren't something we document). We fixed a bug where the environment variable / file for the Consul token weren't written.

client/allocrunner/taskrunner/consul_hook.go
```diff
@@ -32,15 +32,17 @@ type consulHook struct {
...
 
-func newConsulHook(logger log.Logger, tr *TaskRunner, hookResources *cstructs.AllocHookResources) *consulHook {
+func newConsulHook(logger log.Logger, tr *TaskRunner) *consulHook {
 	h := &consulHook{
 		task:          tr.Task(),
 		tokenDir:      tr.taskDir.SecretsDir,
-		hookResources: hookResources,
+		hookResources: tr.allocHookResources,
 	}
+	h.taskEnv = tr.envBuilder.Build().Map()
```
tgross, line 45: The task env is a "builder", so when you call `Build` here you're finalizing it, rather than exposing the internal map to be written. You should make the builder part of the `consulHook` struct and then call a new `SetConsulToken` method on it (similar to the existing `SetVaultToken` method).

pkazmierczak: fixed in [ad83aff](https://github.com/hashicorp/nomad/pull/19490/commits/ad83aff166ef2be0b4ceb617450fa4415280471f)

.changelog/19490.txt
pkazmierczak: rephrased in 2fbde1e9d0af0812fdaca059ca4d417c60de1757

client/allocrunner/taskrunner/task_runner_hooks.go
```diff
@@ -109,6 +109,8 @@ func (tr *TaskRunner) initHooks() {
 
 	// If there are templates is enabled, add the hook
 	if len(task.Templates) != 0 {
+		tr.runnerHooks = append(tr.runnerHooks, newConsulHook(hookLogger, tr))
```
tgross, line 112: If someone has configured a Consul identity _without_ a `template` block (so they can make their own API calls to Consul, for example), they wouldn't get the environment variable / file because this is guarded behind this conditional. I think we probably want to move this outside the `len(task.Templates) != 0` check, but still before we call the template.

tgross (approved):
LGTM once the one item I've called out is fixed (or dismissed)

PR #19496, opened 2023-12-15
[ui] Adds a copy button to Action output

philrenaud:
A happy little button on some happy little output

Stays sticky when/if you scroll

Copies action output to clipboard

image


ksatirli (approved):
Confirmed as working with and without Pack. Love it!

PR #19505, opened 2023-12-15
ui: fix `AllocationRow` for job without action

lgfa29:
The allocation table header sometimes conditionally renders the `Actions` table column, but the allocation row would render it unconditionally, resulting in broken tables when rendering allocations for jobs without actions, where rows had more columns than the header.

Also fix the conditional class for the deployments allocation table to read `length` from the right value.

Before:

image
image
image

After:

image
image
image


philrenaud (approved):
Thank you!

PR #19513, opened 2023-12-18
ci: run linter in `./api` package

lgfa29:
[Outstanding errors](https://github.com/hashicorp/nomad/actions/runs/7255037951/job/19764926088?pr=19513#step:6:31):

```
 Error: agent.go:313:46: response body must be closed (bodyclose)
	_, resp, err := requireOK(a.client.doRequest(r))
	                                            ^
Error: allocations_exec.go:112:36: response body must be closed (bodyclose)
		conn, _, _ = nodeClient.websocket(reqPath, q)
		                                 ^
Error: allocations_exec.go:116:36: response body must be closed (bodyclose)
		conn, _, err = s.client.websocket(reqPath, q)
		                                 ^
Error: services.go:254:2: structtag: struct field tag `hcl:"cluster,optional` not compatible with reflect.StructTag.Get: bad syntax for struct tag value (govet)
	Cluster string `hcl:"cluster,optional`
	^
```

`services.go` is fixed in https://github.com/hashicorp/nomad/pull/19510 but I'm not sure how to fix the `bodyclose` errors.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -161,6 +161,9 @@ check: ## Lint the source code
 	@echo "==> Linting source code..."
 	@golangci-lint run
 
+	@echo "==> Linting ./api source code..."
+	@cd ./api && golangci-lint run --config ../.golangci.yml
+
 	@echo "==> Linting hclog statements..."
 	@hclogvet .
 
--- a/api/agent.go
+++ b/api/agent.go
@@ -310,7 +310,7 @@ func (a *Agent) Monitor(stopCh <-chan struct{}, q *QueryOptions) (<-chan *Stream
 	}
 
 	r.setQueryOptions(q)
-	_, resp, err := requireOK(a.client.doRequest(r))
+	_, resp, err := requireOK(a.client.doRequest(r)) //nolint:bodyclose
 	if err != nil {
 		errCh <- err
 		return nil, errCh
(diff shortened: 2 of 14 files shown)


schmichael (approved):
> I'm not sure how to fix the bodyclose errors.

For websocket requests we don't have to: 

> The response body may not contain the entire response and does not need to be closed by the application.

https://pkg.go.dev/github.com/gorilla/websocket#Dialer.DialContext

So I think we just need some ignores for them: https://golangci-lint.run/usage/false-positives/

lgfa29:
The linter only shows a subset of errors, there's like a billion of `bodyclose`  false positives 😭 
https://github.com/hashicorp/nomad/commit/8ab00151ad4d2ef54d6c87139ccbb2b34b91b536

Should I just disable the `bodyclose` check here or is it still worth to keep all of these `nolint` comments?

shoenig:
> Should I just disable the bodyclose check here

I think no; the Client package is something that gets imported by many external consumers and we should strive to make it a good library citizen, which the linter helps us provide.

lgfa29:
Cool! So I added the `nolint` rule on false positives and maybe fixed some true positives 😅 

I would appreciate another 👀  on those to make sure I didn't break anything.

schmichael (approved):

PR #10698, opened 2021-06-03
drivers/exec: Don't inherit Nomad oom_score_adj value

notnoop:
Explicitly set the `oom_score_adj` value for `exec` and `java` tasks.

We recommend that the Nomad service to have oom_score_adj of a low value
(e.g. -1000) to avoid having nomad agent OOM Killed if the node is
oversubscriped.

However, Nomad's workloads should not inherit Nomad's process, which is
the default behavior.

Added a failed test first that you can see in https://app.circleci.com/pipelines/github/hashicorp/nomad/16419/workflows/2c6fee1f-13c5-4d8b-99c2-75bce52bb949/jobs/158264 .

Fixes #10663

PR #19515, opened 2023-12-19
drivers/executor: set oom_score_adj for raw_exec

mattrobenolt:
This might not be wholly true since I don't know all configurations of Nomad, but in our use cases, we run some of our tasks as `raw_exec` for reasons.

We observed that our tasks were running with `oom_score_adj = -1000`, which prevents them from being OOM'd. This value is being inherited from the nomad agent parent process, as configured by systemd.

Similar to #10698, we also were shocked to have this value inherited down to every child process and believe that we should also set this value to 0 explicitly.

I have no idea if there are other paths that might leverage this or other ways that `raw_exec` can manifest, but this is how I was able to observe and fix in one of our configurations.

We have been running in production our tasks wrapped in a script that does: `echo 0 > /proc/self/oom_score_adj` to avoid this issue.

### DIFF
--- a/.changelog/19515.txt
+++ b/.changelog/19515.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+rawexec: Fixed a bug where oom_score_adj would be inherited from Nomad client
+```
--- a/drivers/shared/executor/executor_universal_linux.go
+++ b/drivers/shared/executor/executor_universal_linux.go
@@ -113,8 +113,8 @@ func (e *UniversalExecutor) statCG(cgroup string) (int, func(), error) {
 func (e *UniversalExecutor) configureResourceContainer(command *ExecCommand, pid int) (func(), error) {
 	cgroup := command.StatsCgroup()
 
-	// children should not inherit Nomad agent oom_score_adj value
-	if err := os.WriteFile("/proc/self/oom_score_adj", []byte("0"), 0644); err != nil {
+	// ensure tasks do not inherit Nomad agent oom_score_adj value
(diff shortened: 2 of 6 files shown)


mattrobenolt:
To add, we applied this patch to a production host, patched over 1.7.1 tag and it behaves as expected.

shoenig:
Thanks @mattrobenolt! Hope you don't mind I've cleaned up the code just a bit and added an e2e test. Spot checking to make sure it works:

before:
```
➜ go test -v
=== RUN   TestRawExec
=== RUN   TestRawExec/testOomAdj
    assert.go:14:
        rawexec_test.go:28: expected string to contain substring; it does not
        ↪ substring: 0
        ↪    string: -999
--- FAIL: TestRawExec (4.01s)
    --- FAIL: TestRawExec/testOomAdj (3.83s)
```

after:

```
➜ go test -v
=== RUN   TestRawExec
=== RUN   TestRawExec/testOomAdj
--- PASS: TestRawExec (2.96s)
    --- PASS: TestRawExec/testOomAdj (2.77s)
PASS
ok      github.com/hashicorp/nomad/e2e/rawexec  2.962s
```

mattrobenolt:
That's great! Thank you. :) 

Do you know if this will be applicable for a 1.7.x release or need to wait for 1.8?

shoenig:
It will go into the next 1.7.x bugfix release (1.7.3 assuming we don't get a CVE release before then)

shoenig (approved):

PR #19535, opened 2023-12-19
Adds vault role to JWT claims if specified in jobspec

mikenomitch:
Adds vault_role as an attribute in the JWT if, and only if, the role is specified in the jobspec. This would allow Vault users to interpolate the Vault role in their templated policies. This could be generally useful, but was primarily added to unblock a specific customer. They have verified that this fix (including the constraint below) would unblock JWT-based Vault auth for their setup.

Ideally, we could add the role regardless of where it is specified, but since the server is minting the JWTs and the client has the default fallback auth_role values, it is difficult to insert the role into the JWT claims if it is using the fallback. There is also a potential syncing issue if you update config and then restart Nomad. To get around these issues, Vault Namespaces have the same constraint (only adding to claims if in the jobspec), so we're adding the same constraint to the role JWT claim.

### DIFF
--- a/.changelog/19535.txt
+++ b/.changelog/19535.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+identity: Added vault_role to JWT workload identity claims if specified in jobspec
+```
--- a/website/content/docs/concepts/workload-identity.mdx
+++ b/website/content/docs/concepts/workload-identity.mdx
@@ -38,6 +38,23 @@ instead of the task name.
 }
 ```
 
+Workload identities for tasks that use Vault have an additional claim for
+[`vault.role`][jobspec_vault_role] if a role is specified in the job.
+
+-> **Note:** This claim will *not* be added if the role is inherited from
+the [agent configuration][vault_role_agent_config] and is not present
+in the Nomad job specification.
+
+```json
+{
+  "nomad_namespace": "default",
+  "nomad_job_id": "example",
+  "nomad_allocation_id": "8623ac7a-28ba-20c3-24a6-e615a39bbbf3",
+  "nomad_service": "example-cache-redis"
+  "vault_role": "nomad-jwt-login"
+}
+```
+
 ### Nomad Enterprise <EnterpriseAlert inline />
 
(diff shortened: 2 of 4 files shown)


.changelog/19535.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+identity: Added vault role to jwt claims if specified in jobspec
```
lgfa29, line 2:
```suggestion
identity: Added `vault_role` to JWT workload identity claims if specified in jobspec
```

nomad/structs/workload_id.go
```diff
@@ -302,6 +302,7 @@ type WIHandle struct {
 	IdentityName string
 	// WorkloadIdentifier is either a ServiceName or a TaskName
 	WorkloadIdentifier string
+	VaultRole          string
```
lgfa29, line 305:
```suggestion
```

I don't think we need this anymore?

website/content/docs/concepts/workload-identity.mdx
```diff
@@ -38,6 +38,23 @@ instead of the task name.
 }
 ```
 
+Workload identities for Vault services have an additional claim for
+[`vault.role`][jobspec_vault_role] if role is specified in the job.
```
lgfa29, line 42:
```suggestion
Workload identities for tasks that use Vault have an additional claim for
[`vault.role`][jobspec_vault_role] if a role is specified in the job.
```

lgfa29 (approved):
LGTM! Mostly some minor suggestions, and I think we need to remove that `VaultRole` from `WIHandle` since we're not using it anymore?

nomad/structs/workload_id.go
mikenomitch: Oh damn, sorry I missed this 🤦- thought this was the Claims struct when I was self-reviewing

BGMUSTC:
When would this feature released? Do you have any plan?

pkazmierczak:
@BGMUSTC The feature will be released in the upcoming 1.7.3 release.

PR #19544, opened 2023-12-20
[ui] Multi-line variable values and helios upgrades generally

### LINKED ISSUE #19543: [ui] multi-line variable values

philrenaud:
This started as an attempt to allow multi-line variable editing in the UI (#19543) and turned into a larger effort to replace the Variables UI's use of common components with [Helios components](https://helios.hashicorp.design/)

image
image
image
image

Resolves #19543


ui/app/components/variable-form.hbs
```diff
@@ -166,25 +154,9 @@
...
-        <button
-          class="add-more button is-info is-inverted"
-          type="button"
-          disabled={{not (and this.keyValues.lastObject.key this.keyValues.lastObject.value)}}
-          {{on "click" this.appendRow}}
-          data-test-add-kv
-        >
-          Add More
-        </button>
+        <Hds::Button @text="Add More" @color="secondary" @size="medium" @icon="plus" class="add-more" type="button" disabled={{not (and this.keyValues.lastObject.key this.keyValues.lastObject.value)}} {{on "click" this.appendRow}} data-test-add-kv />
```
philrenaud, line 157: This long `disabled` attr says "Don't let the user add more fields if their last one is currently empty"

ui/app/components/variable-form.js
```diff
@@ -207,6 +208,7 @@ export default class VariableFormComponent extends Component {
    * @param {KeyboardEvent} e
    */
   @action setModelPath(e) {
+    set(this, 'path', e.target.value);
```
philrenaud, line 211: This, and the `entry` set above it, are side-effects of the Helios input component not doing two-way data binding. For some additional context, two-way data binding is kind of a bygone convention in most JS frameworks at this point so it's not that weird, but we were using that bygone convention for quite awhile. Now, we are explicit about "When you do something that indicates needing to save something, save it".

In this case, that happens in the template on keystroke (`{{on "input" (action this.setModelPath)}}`), but it's also common to do this on element blur, on form submit, etc. We're just eager beavers here.

ui/app/components/variable-paths.hbs
```diff
@@ -57,17 +62,25 @@
...
+
+
 
-  </tbody>
-      </ListTable>
+    {{!-- <B.Tr>
+      <B.Td>
+        
+      </B.Td>
+    </B.Tr> --}}
```
philrenaud, line 84: Phil TODO

ui/app/styles/components/variables.scss
```diff
@@ -5,28 +5,15 @@
...
-    text-decoration: none;
-    color: #363636;
-    position: relative;
-    top: 4px;
-  }
 }
 
+$hdsLabelTopOffset: 26px;
+$hdsInputHeight: 35px;
+
```
philrenaud, line 12: A side-effect of using multi-line variables work with multiple lines is, we want both to let users control how tall the textbox is, and we want other items to look like a grid and line up vertically with other elements.

This gets a little tricky because in some instances we hold spacing above an element (for help text) or below an element (for input errors) and this is a decent but not perfect attempt at keeping them generally lined up.

ui/app/templates/variables/variable/index.hbs
```diff
@@ -3,45 +3,54 @@
...
-    {{#unless this.isDeleting}}
+      {{on "change" (action this.toggleView)}}
+    as |F|>
+      <F.Label>JSON</F.Label>
+    </Hds::Form::Toggle::Field>
+
+    <div
+      {{keyboard-shortcut 
+        label="Copy Variable"
+        pattern=(array "c" "v")
```
philrenaud, line 29: Charlie gets his very own keyboard shortcut

ui/app/templates/variables/variable/index.hbs
```diff
@@ -72,59 +81,65 @@
...
+      label="Key"
+      isSortable=true
+      width="200px"
+    )
+    (hash
+      key="value"
+      label="Value"
+      isSortable=true
+    )
+  }}
```
philrenaud, line 106: An astute observe may note that some of the `<Hds::Table>`s here have a `@columns` property, where others do the `<:head as |H|>` and `H.Tr` thing. Helios tables allow us to do either, and the best argument for doing the former has to do with sorting being baked in.

juliezzhou (approved):
lgtm

gulducat (approved):
I found some nits to pick, but nothing really blocky, so overall LGTM!

ui/app/components/variable-paths.hbs
```diff
@@ -3,22 +3,26 @@
...
-    </th>
-    <th>
-      Namespace
-    </th>
-    <th>
-      Last Modified
-    </th>
-  </t.head>
-  <tbody>
+{{!-- TODO: flight icons etc. --}}
```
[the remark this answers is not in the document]
philrenaud: They are these little fellas: 
image

Helios supports them here: https://helios.hashicorp.design/icons/usage-guidelines

And I think they're good to go now, but I reeeally wish I'd written myself a better TODO.

philrenaud:
@gulducat growing textarea on input is a good idea here. Let me see if the Helios masked input can support `data-replicated-value` css, but even the state of the art for solving this ( https://css-tricks.com/the-cleanest-trick-for-autogrowing-textareas/ ) is kind of hacky. Going to give it some thought.

ui/app/templates/variables/index.hbs
```diff
@@ -5,46 +5,51 @@
 
 {{page-title "Variables"}}
 <section class="section">
-  <div class="toolbar">
-    <div class="toolbar-item is-right-aligned is-mobile-full-width">
+
+  <Hds::PageHeader class="variable-title" as |PH|>
+    <PH.Title>Variables</PH.Title>
```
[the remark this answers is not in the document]
philrenaud: fixed and fixed: 
image
image

ui/app/components/variable-form/namespace-filter.hbs
```diff
@@ -10,19 +10,18 @@
...
-          <SingleSelectDropdown
-            data-test-variable-namespace-filter
-            @label="Namespace"
-            @disabled={{@data.disabled}}
-            @options={{@data.namespaceOptions}}
-            @selection={{@data.selection}}
-            @onSelect={{@fns.onSelect}}
-          />
-        </label>
+        <Hds::Dropdown class="namespace-dropdown" data-test-variable-namespace-filter as |dd|>
```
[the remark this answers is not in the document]
philrenaud: This was a conscious decision by me, but I think now a bad one. I'm going to put the namespace name in parentheses like `Namespace (default)` going forward here.

ui/app/components/variable-form.hbs
```diff
@@ -166,25 +154,9 @@
...
-    <button
-      disabled={{this.shouldDisableSave}}
-      class="button is-primary save"
-      type="submit"
-      data-test-submit-var
-    >
-      Save
-      {{pluralize "Variable" @this.keyValues.length}}
-    </button>
+    <Hds::Button @text="Save {{pluralize "Variable" this.keyValues.length}}" @color="primary" type="submit" disabled={{this.shouldDisableSave}} data-test-submit-var />
```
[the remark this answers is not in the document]
philrenaud: Did it anyway and thank you for the suggestion on it! The class/type on the previous example were extraneous as you pointed out — this also lets me remove some CSS bloat, too! Great catch.

ui/app/routes/access-control/namespaces/new.js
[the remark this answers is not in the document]
philrenaud: Good catch, I must've accidentally saved it when moving between branches locally.

ui/app/styles/components/variables.scss
philrenaud: Oh, I should've responded inline here, but addressed this here: https://github.com/hashicorp/nomad/pull/19544#issuecomment-1875681046

TLDR is this is a good idea, but a little tricky! Going to try it outside-of-PR

PR #19547, opened 2023-12-21
docs: add note about `*` namespace on autoscaling

lgfa29:
Explain the behaviour when the wildcard namespace value `*` is used to configure the Nomad Autoscaler agent.

Ref.: https://github.com/hashicorp/nomad-autoscaler/issues/65#issuecomment-1620018801

### DIFF
--- a/website/content/tools/autoscaling/agent/nomad.mdx
+++ b/website/content/tools/autoscaling/agent/nomad.mdx
@@ -26,7 +26,7 @@ nomad {
 - `region` `(string: "global")` - The region of the Nomad servers to connect with.
 
 - `namespace` `(string: "")` - The target namespace for queries and actions bound
-  to a namespace.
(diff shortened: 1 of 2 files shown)


website/content/tools/autoscaling/agent/nomad.mdx
```diff
@@ -26,7 +26,7 @@ nomad {
 - `region` `(string: "global")` - The region of the Nomad servers to connect with.
 
 - `namespace` `(string: "")` - The target namespace for queries and actions bound
-  to a namespace.
+  to a namespace. If set to `*` all namespaces are queried.
```
jrasell, line 29:
```suggestion
  to a namespace. If set to `*`, all namespaces are queried.
```

jrasell (approved):
LGTM.

Do we need backport labels?

PR #19568, opened 2024-01-02
build: bump setup-golang and fix workflow yaml extensions

shoenig:
Bumps our `hashicorp/setup-golang` to v2, which itself bumps the official `cache` and `setup-go` actions.

Also fixes the file extension of some of our workflows/ files to use `.yaml` which is [technically correct](https://yaml.org/faq.html)


lgfa29 (approved):
There are a few places that are using `.yml` that need to be updated as well:
https://github.com/hashicorp/nomad/blob/bb3d2227a28a76a314572ccaadd34b4820f0fed4/.github/workflows/release.yml#L150

The Nomad Enterprise repo may have additional ones for the CE->ENT flow, and the release checklist as well 😅

shoenig:
Hah. Uhh let me just backout the yaml changes; they're beside the point.

Issue #19565, opened 2023-12-28
Nomad validate not working with new Workload Identity Vault integration


EtienneBruines:
### Nomad version
Nomad v1.7.2
BuildDate 2023-12-13T19:59:42Z
Revision 64e3dca9274b493e38a49fda3a70fd31d0485b91

### Operating system and Environment details
Ubuntu 20.04 LTS

### Issue
Twofold:
1. Upon calling `nomad validate`, this error occurs if I do not have a `VAULT_TOKEN` set:
```
1 error occurred:
	* Vault used in the job but missing Vault token
```
2. Upon calling `nomad validate`, this error occurs if I do have a `VAULT_TOKEN` set (but have my servers already configured to have a `default_identity`:
```
1 error occurred:
	* failed to lookup Vault token: Vault client not able to lookup tokens
```
This latter error occurs because the Nomad Server is configured with the [`NoopVault`](https://github.com/hashicorp/nomad/blob/76ba3e10e73dd3d880a848b64ea6915f017998d5/nomad/vault_noop.go) which returns that error. It is using the `NoopVault` instead of an actual Vault Client because the `default_identity` is set and the "legacy" code paths shouldn't be used. The validation apparently still uses the legacy code paths.

### Reproduction steps
- Have a job file that contains the `vault` stanza. 
- Have a Nomad Server that uses the new Workload Identity integration with Vault (specifically having `default_identity` configured - otherwise the legacy code path takes effect, if I understand the source code correctly).
- Call `nomad validate` for that jobfile

#### Expected Result
Short-term: If the Nomad Server uses the new Workload Identity, to not mess up the `nomad validate` command.

Long-term: Actually verifying whether the Vault parts "work", e.g. grant access to the resources they should. 

#### Actual Result
Errors, see above.

### Nomad Server logs (if appropriate)
None.

However, when I do not specify an `identity` in my job file (despite having a `default_identity` configured at the Nomad Server), I get these errors (which indicate the usage of `NoopVault`):

```
2023-12-28T15:15:13.093Z [ERROR] nomad.client: Vault token creation for alloc failed: alloc_id=0ef75eda-1bae-1e76-4343-a0d9e3b72b7b error="Vault client not able to create tokens"
2023-12-28T15:15:13.093Z [ERROR] nomad.client: Vault token revocation for alloc failed: alloc_id=0ef75eda-1bae-1e76-4343-a0d9e3b72b7b error="Vault client not able to revoke tokens"
```

[...]

PR #19585, opened 2024-01-02
vault: ignore `allow_unauthenticated` config if identity is set

tgross:
When the server's `vault` block has a default identity, we don't check the user's Vault token (and in fact, we warn them on job submit if they've provided one). But the validation hook still checks for a token if `allow_unauthenticated` is set to true. This is a misconfiguration but there's no reason for Nomad not to do the expected thing here.

Fixes: https://github.com/hashicorp/nomad/issues/19565


lgfa29 (approved):

tgross:
Backported to 1.7.x by hand

PR #19596, opened 2024-01-03
e2e: add test for raw_exec memory_max configuration

shoenig:
Nomad 1.7 started enforcing memory limits via cgroups on Linux platforms. Add an e2e test demonstrating the use of `memory_max` over subscription. And update the resources docs to mention support for the `raw_exec` driver.


gulducat (approved):
does what it says on the tin :+1:

PR #19600, opened 2024-01-03
[ui] Changelog for multi-line variables

philrenaud:
Forgot this in #19544

### DIFF
--- a/.changelog/19544.txt
+++ b/.changelog/19544.txt
@@ -1,3 +1,6 @@
 ```release-note:improvement
-ui: Replaced single-line variable value fields with multi-line textareas and updated the components to use newer styles, generally
+ui: Replaced single-line variable value fields with multi-line textarea blocks
+```
+```release-note:improvement
+ui: Updated the style of components in the Variables web ui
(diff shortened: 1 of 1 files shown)


.changelog/19544.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+ui: Replaced single-line variable value fields with multi-line textareas and updated the components to use newer styles, generally
```
gulducat, line 2: nit: "the components" refers specifically to those on variables pages only, right?  I think that vague wording and the "generally" on the end makes me think it might be broader than that.  but, I don't want to suggest being toooo wordy, because that could detract from the main feature here: multi-line variable text boxes.

could maybe make two separate \`\`\`release-note:...\`\`\` entries for the two change categories?

gulducat (approved):
ch-ch-ch-ch-chaaangeeesss

.changelog/19544.txt
philrenaud: Updated to multiple entries, thanks!

PR #19603, opened 2024-01-04
e2e: pass Nomad address to Consul WI test

pkazmierczak:
Otherwise we get an error:
```
        error: Nomad JWKS endpoint unreachable, verify that Nomad is running and that the JWKS URL http://localhost:4646/.well-known/jwks.json is reachable by Consul
```


jrasell (approved):
LGTM and thanks for fixing!

PR #19605, opened 2024-01-04
state: ensure the job submission table is persisted and restored.

### LINKED ISSUE #19561: Lost HCL job spec after upgrade (and service restart?)
### Nomad version

Previous : 1.6.2
Updated : 1.7.2

Rocky 8.9

### Issue

Hi,

After upgrading Nomad servers from 1.6.2 to 1.7.2 we lost all HCL Jobspec. The jobs start but when we try to edit the job definition we are unable to get the HCL formatted spec, only the full JSON is available.

image

"A jobspec file was not submitted when this job was run. You can still view and edit the expanded JSON format."

We tried to restore the 1.6.2 cluster backup in a test environnement but we couldn't get the HCL spec, only the JSON one is available...

How can we get back the HCL formatted job spec ?

### Reproduction steps

Upgrade a cluster from 1.6.2 to 1.7.2 or restore a 1.6.2 cluster

jrasell:
Previously the job submission table was not persisted within state snapshots, meaning servers which received a snapshot during a restore process would seemingly lose the HCL submission data for jobs. This changes ensures this table is both persisted and restored.

closes #19561

### DIFF
--- a/.changelog/19605.txt
+++ b/.changelog/19605.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core: Ensure job HCL submission data is persisted and restored during the FSM snapshot process
+```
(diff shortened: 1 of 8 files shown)


shoenig (approved):
LGTM

lgfa29 (approved):

philrenaud (approved):
Works like a charm!

PR #19609, opened 2024-01-04
state store: better handling of job deletion

### LINKED ISSUE #10502: unfinished GC may block placements of new version of the job
We received in an internal report where a job that had been stopped with the `-purge` flag and then re-run would not be placed until after a `nomad system gc` command was run. But upon further observation, it appears that while the job doesn't get placed initially, it _does_ eventually get placed.

I suspect that the `system gc` is a bit of a red herring and timing dependent. There's already a GC evalution for the job in flight because of the `-purge` flag. But the old allocation still exists in the state store because the client hasn't GC'd it yet. The old allocation's reference to the job is orphaned, but then becomes valid once the new job is registered.

The reporter of the issue has a workaround to run `system gc` (or they could simply not re-run the job immediately after purging it). So this isn't particularly urgent.

```hcl
job "example" {
  datacenters = ["dc1"]

  group "group" {
    task "task" {
      driver = "docker"
      config {
        image      = "centos:centos7"
        force_pull = true
        command    = "/blah/non-existent-file"
      }

      resources {
        cpu    = 200
        memory = 64
      }
    }
  }
}
```

</details>

To reproduce, run the failing job:

```
$ nomad job run ./example.nomad
==> Monitoring evaluation "1a2b9089"
    Evaluation triggered by job "example"
==> Monitoring evaluation "1a2b9089"
    Evaluation within deployment: "e21d83a7"
    Allocation "ef54f207" created: node "7df6e0cf", group "group"
    Evaluation status changed: "pending" -> "complete"
==> Evaluation "1a2b9089" finished with status "complete"
vagrant@linux$ nomad job stop -purge ^C
vagrant@linux$ nomad job status ex
ID            = example
Name          = example
Submit Date   = 2021-05-04T13:34:46Z
Type          = service
Priority      = 50
Datacenters   = dc1
Namespace     = default
Status        = pending
Periodic      = false
Parameterized = false

Summary
Task Group  Queued  Starting  Running  Failed  Complete  Lost
group       0       0         0        1       0         0

Future Rescheduling Attempts
Task Group  Eval ID   Eval Time
group       ab6d7799  24s from now

Latest Deployment
ID          = e21d83a7
Status      = running

Deployed
group       1        1       0        1          2021-05-04T13:44:46Z

Allocations
ID        Node ID   Task Group  Version  Desired  Status  Created  Modified
ef54f207  7df6e0cf  group       0        run      failed  7s ago   2s ago
```

Stop and purge the job, and note that no deployment is visible:

```
$ nomad job stop -purge example
==> Monitoring evaluation "b98aa4ca"
    Evaluation triggered by job "example"
==> Monitoring evaluation "b98aa4ca"
    Evaluation within deployment: "e21d83a7"
    Evaluation status changed: "pending" -> "complete"
==> Evaluation "b98aa4ca" finished with status "complete"
vagrant@linux$ nomad job run ./example.nomad
==> Monitoring evaluation "1b42cb0e"
    Evaluation triggered by job "example"
==> Monitoring evaluation "1b42cb0e"
    Evaluation within deployment: "e21d83a7"
    Evaluation status changed: "pending" -> "complete"
==> Evaluation "1b42cb0e" finished with status "complete"

$ nomad job status ex
ID            = example
Name          = example
Submit Date   = 2021-05-04T13:35:03Z
Type          = service
Priority      = 50
Datacenters   = dc1
Namespace     = default
Status        = pending
Periodic      = false
Parameterized = false

Summary
Task Group  Queued  Starting  Running  Failed  Complete  Lost
group       0       0         0        0       0         0

Allocations
No allocations placed
```

Wait a little bit and see that the deployment has been created and failed as expected:

```
$ nomad job status example
...
Latest Deployment
ID          = 9e69d457
Status      = running

Deployed
group       1        1       0        1          2021-05-04T13:45:18Z

Allocations
ID        Node ID   Task Group  Version  Desired  Status  Created  Modified
09ca1a69  7df6e0cf  group       0        run      failed  16s ago  11s ago
```

pkazmierczak:
When jobs are deleted with `-purge`, all their deployments and allocations should be deleted from the state store, and the evals status should be set to `complete`. Otherwise we end up in a situation where users could re-submit previously failing jobs, but these new jobs would not get deployments allocated unless `system gc` got called. 

Fixes #10502

### DIFF
--- a/nomad/core_sched_test.go
+++ b/nomad/core_sched_test.go
@@ -546,7 +546,6 @@ func TestCoreScheduler_EvalGC_Batch(t *testing.T) {
 			stoppedAlloc, lostAlloc,
 			activeJobRunningAlloc, activeJobLostAlloc, activeJobCompletedEvalCompletedAlloc,
 			stoppedJobStoppedAlloc, stoppedJobLostAlloc,
-			purgedJobCompleteAlloc,
 		},
 		[]*structs.Allocation{},
 	)
@@ -576,7 +575,6 @@ func TestCoreScheduler_EvalGC_Batch(t *testing.T) {
 			stoppedAlloc, lostAlloc,
 			activeJobRunningAlloc, activeJobLostAlloc, activeJobCompletedEvalCompletedAlloc,
 			stoppedJobStoppedAlloc, stoppedJobLostAlloc,
-			purgedJobCompleteAlloc,
 		},
 		[]*structs.Allocation{},
 	)
@@ -609,7 +607,6 @@ func TestCoreScheduler_EvalGC_Batch(t *testing.T) {
 		[]*structs.Allocation{
 			activeJobLostAlloc, activeJobCompletedEvalCompletedAlloc,
 			stoppedJobLostAlloc, stoppedJobLostAlloc,
-			purgedJobCompleteAlloc,
 		})
 }
 
--- a/nomad/core_sched_test.go
+++ b/nomad/core_sched_test.go
@@ -596,7 +596,7 @@ func TestCoreScheduler_EvalGC_Batch(t *testing.T) {
 	//    than that of the job).
 	//	3. The active job remains since it is active, even though the allocations are otherwise
 	//      eligible for GC. However, the inactive allocation is GCed for it.
-	//	4. The eval and allocation for the purged job are GCed.
+	//	4. The eval and allocation for the purged job are deleted.
 	assertCorrectJobEvalAlloc(
 		memdb.NewWatchSet(),
 		[]*structs.Job{deadJob, activeJob, stoppedJob},
@@ -607,6 +607,7 @@ func TestCoreScheduler_EvalGC_Batch(t *testing.T) {
 		[]*structs.Allocation{
 			activeJobLostAlloc, activeJobCompletedEvalCompletedAlloc,
 			stoppedJobLostAlloc, stoppedJobLostAlloc,
+			purgedJobCompleteAlloc,
 		})
 }
 
--- a/nomad/state/state_store.go
+++ b/nomad/state/state_store.go
@@ -1900,14 +1900,9 @@ func (s *StateStore) DeleteJobTxn(index uint64, namespace, jobID string, txn Txn
 		return fmt.Errorf("deployment lookup for job %s failed: %v", job.ID, err)
 	}
 
-	deploymentIDs := []string{}
-	for _, d := range deployments {
-		deploymentIDs = append(deploymentIDs, d.ID)
-	}
-
-	for _, deploymentID := range deploymentIDs {
+	for _, deployment := range deployments {
 		// Lookup the deployment
-		existing, err := txn.First("deployment", "id", deploymentID)
+		existing, err := txn.First("deployment", "id", deployment.ID)
 		if err != nil {
 			return fmt.Errorf("deployment lookup failed: %v", err)
 		}
@@ -1949,6 +1944,27 @@ func (s *StateStore) DeleteJobTxn(index uint64, namespace, jobID string, txn Txn
 
 		if err := txn.Insert("index", &IndexEntry{"evals", index}); err != nil {
 			return fmt.Errorf("index update failed: %v", err)
+		}
+	}
+
+	allocs, err := s.AllocsByJob(nil, namespace, job.ID, true)
+	if err != nil {
+		return fmt.Errorf("alloc lookup for job %s failed: %v", job.ID, err)
+	}
+
+	for _, alloc := range allocs {
+		// Lookup the alloc
+		existing, err := txn.First("allocs", "id", alloc.ID)
+		if err != nil {
+			return fmt.Errorf("alloc lookup failed: %v", err)
+		}
+		if existing == nil {
+			return fmt.Errorf("alloc not found")
+		}
+
+		// Delete the alloc
+		if err := txn.Delete("allocs", existing); err != nil {
+			return fmt.Errorf("deployment delete failed: %v", err)
 		}
 	}
 
--- a/nomad/state/state_store.go
+++ b/nomad/state/state_store.go
@@ -1894,6 +1894,64 @@ func (s *StateStore) DeleteJobTxn(index uint64, namespace, jobID string, txn Txn
 		return err
 	}
 
+	// Delete job deployments
+	deployments, err := s.DeploymentsByJobID(nil, namespace, job.ID, true)
+	if err != nil {
+		return fmt.Errorf("deployment lookup for job %s failed: %v", job.ID, err)
+	}
+
+	deploymentIDs := []string{}
+	for _, d := range deployments {
+		deploymentIDs = append(deploymentIDs, d.ID)
+	}
+
+	for _, deploymentID := range deploymentIDs {
+		// Lookup the deployment
+		existing, err := txn.First("deployment", "id", deploymentID)
+		if err != nil {
+			return fmt.Errorf("deployment lookup failed: %v", err)
+		}
+		if existing == nil {
+			return fmt.Errorf("deployment not found")
+		}
+
+		// Delete the deployment
+		if err := txn.Delete("deployment", existing); err != nil {
+			return fmt.Errorf("deployment delete failed: %v", err)
+		}
+	}
+
+	// Mark all evals for this job as "complete"
+	pendingEvals, err := s.EvalsByJob(nil, namespace, job.ID)
+	if err != nil {
+		return fmt.Errorf("eval lookup for job %s failed: %v", job.ID, err)
+	}
+
+	for _, eval := range pendingEvals {
+		existing, err := txn.First("evals", "id", eval.ID)
+		if err != nil {
+			return fmt.Errorf("eval lookup failed: %v", err)
+		}
+		if existing == nil {
+			continue
+		}
+
+		eval := existing.(*structs.Evaluation).Copy()
+		eval.Status = structs.EvalStatusComplete
+
+		// Insert the eval
+		if err := txn.Insert("evals", eval); err != nil {
+			return fmt.Errorf("eval insert failed: %v", err)
+		}
+		if err := txn.Insert("index", &IndexEntry{"evals", index}); err != nil {
+			return fmt.Errorf("index update failed: %v", err)
+		}
+
+		if err := txn.Insert("index", &IndexEntry{"evals", index}); err != nil {
+			return fmt.Errorf("index update failed: %v", err)
+		}
+	}
+
 	// Cleanup plugins registered by this job, before we delete the summary
 	err = s.deleteJobFromPlugins(index, txn, job)
 	if err != nil {
--- a/nomad/state/state_store.go
+++ b/nomad/state/state_store.go
@@ -1984,7 +1984,7 @@ func (s *StateStore) DeleteJobTxn(index uint64, namespace, jobID string, txn Txn
 			continue
 		}
 		eval.Status = structs.EvalStatusComplete
-		eval.StatusDescription = fmt.Sprintf("evaluation deleted while purging job %s", job.ID)
+		eval.StatusDescription = fmt.Sprintf("job %s deleted", job.ID)
 
 		// Insert the eval
 		if err := txn.Insert("evals", eval); err != nil {
--- a/nomad/state/state_store.go
+++ b/nomad/state/state_store.go
@@ -1901,7 +1901,6 @@ func (s *StateStore) DeleteJobTxn(index uint64, namespace, jobID string, txn Txn
 	}
 
 	for _, deployment := range deployments {
-		// Lookup the deployment
 		existing, err := txn.First("deployment", "id", deployment.ID)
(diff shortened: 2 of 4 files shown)


tgross (commented):
This looks generally good but it's worth trying to reason through what happens if there are allocations on a client that's disconnected when the job is purged. I'm not sure that it's safe to delete allocations until we know they're terminal?

pkazmierczak:
> what happens if there are allocations on a client that's disconnected when the job is purged.

If these are failed allocs, they remain in `dead` state because there's no scheduler to reschedule their re-run. If they correspond to non-failing jobs, they remain running. 

> I'm not sure that it's safe to delete allocations until we know they're terminal?

Yeah so this to me is more about the intended behavior of the `-purge` flag. In my mind, `purge` should delete any and all allocations on all the clients. 

The behavior introduced in this PR makes one unfortunate side-effect, that is, if a client is disconnected while the job gets purged, and if it's a failing job (with `alloc-failure` evals), after reconnecting this client will leak allocs (in `dead` state). GC cleans them of course.

tgross (changes requested):

nomad/state/state_store.go
```diff
@@ -1894,6 +1894,80 @@ func (s *StateStore) DeleteJobTxn(index uint64, namespace, jobID string, txn Txn
...
+	if err != nil {
+		return fmt.Errorf("deployment lookup for job %s failed: %v", job.ID, err)
+	}
+
+	for _, deployment := range deployments {
+		// Lookup the deployment
+		existing, err := txn.First("deployment", "id", deployment.ID)
+		if err != nil {
+			return fmt.Errorf("deployment lookup failed: %v", err)
+		}
```
lgfa29, line 1908: Why do we need to lookup the deployment? Can `DeploymentsByJobID()` return a deployment that doesn't exist?

nomad/state/state_store.go
```diff
@@ -1894,6 +1894,80 @@ func (s *StateStore) DeleteJobTxn(index uint64, namespace, jobID string, txn Txn
...
+		if existing == nil {
+			return fmt.Errorf("deployment not found")
+		}
```
lgfa29, line 1911:
```suggestion
		if existing == nil {
			continue
		}
```

Since we're trying to delete deployments, if the deployment doesn't exist then our job is already done so I think we can just continue deleting the others instead.

nomad/state/state_store.go
```diff
@@ -1894,6 +1894,80 @@ func (s *StateStore) DeleteJobTxn(index uint64, namespace, jobID string, txn Txn
...
+		existing, err := txn.First("evals", "id", eval.ID)
+		if err != nil {
+			return fmt.Errorf("eval lookup failed: %v", err)
+		}
+		if existing == nil {
+			continue
+		}
+
+		eval := existing.(*structs.Evaluation).Copy()
+		eval.Status = structs.EvalStatusComplete
```
lgfa29, line 1935: Would it be worth to also update `StatusDescription`? Depending on the original value for `Status` the `StatusDescription` may not make sense after this update.

nomad/state/state_store.go
```diff
@@ -1894,6 +1894,80 @@ func (s *StateStore) DeleteJobTxn(index uint64, namespace, jobID string, txn Txn
...
+
+	for _, alloc := range allocs {
+		// Lookup the alloc
+		existing, err := txn.First("allocs", "id", alloc.ID)
+		if err != nil {
+			return fmt.Errorf("alloc lookup failed: %v", err)
+		}
+		if existing == nil {
+			return fmt.Errorf("alloc not found")
+		}
```
lgfa29, line 1963:
```suggestion
		if existing == nil {
			continue
		}
```

Same as in deployment.

lgfa29 (commented):
Some small comments, but one general observation is that we usually have `DeleteX()` and `DeleteXTxn()` methods. I think it would nice to keep this pattern here for deployments and allocs.

nomad/state/state_store.go
```diff
@@ -1894,6 +1894,80 @@ func (s *StateStore) DeleteJobTxn(index uint64, namespace, jobID string, txn Txn
...
 	}
 
+	// Delete job deployments
+	deployments, err := s.DeploymentsByJobID(nil, namespace, job.ID, true)
+	if err != nil {
+		return fmt.Errorf("deployment lookup for job %s failed: %v", job.ID, err)
+	}
+
+	for _, deployment := range deployments {
+		// Lookup the deployment
```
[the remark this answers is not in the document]
pkazmierczak: agreed. addressed in [9360c46](https://github.com/hashicorp/nomad/pull/19609/commits/9360c464a8b31ed0ec6d424a282de1e0beb66d04)

nomad/state/state_store.go
```diff
@@ -1894,6 +1894,80 @@ func (s *StateStore) DeleteJobTxn(index uint64, namespace, jobID string, txn Txn
...
+		}
+
+		// Delete the deployment
+		if err := txn.Delete("deployment", existing); err != nil {
+			return fmt.Errorf("deployment delete failed: %v", err)
+		}
+	}
+
+	// Mark all evals for this job as "complete"
+	pendingEvals, err := s.EvalsByJob(nil, namespace, job.ID)
```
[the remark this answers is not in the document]
pkazmierczak: Fair enough. Addressed in [9360c46](https://github.com/hashicorp/nomad/pull/19609/commits/9360c464a8b31ed0ec6d424a282de1e0beb66d04)

nomad/state/state_store.go
```diff
@@ -1894,6 +1894,80 @@ func (s *StateStore) DeleteJobTxn(index uint64, namespace, jobID string, txn Txn
...
+			return fmt.Errorf("deployment lookup failed: %v", err)
+		}
+		if existing == nil {
+			return fmt.Errorf("deployment not found")
+		}
+
+		// Delete the deployment
+		if err := txn.Delete("deployment", existing); err != nil {
+			return fmt.Errorf("deployment delete failed: %v", err)
+		}
```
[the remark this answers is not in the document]
pkazmierczak: Good point, addressed in [9360c46](https://github.com/hashicorp/nomad/pull/19609/commits/9360c464a8b31ed0ec6d424a282de1e0beb66d04)

nomad/state/state_store.go
```diff
@@ -1894,6 +1894,80 @@ func (s *StateStore) DeleteJobTxn(index uint64, namespace, jobID string, txn Txn
...
+			return fmt.Errorf("alloc lookup failed: %v", err)
+		}
+		if existing == nil {
+			return fmt.Errorf("alloc not found")
+		}
+
+		// Delete the alloc
+		if err := txn.Delete("allocs", existing); err != nil {
+			return fmt.Errorf("deployment delete failed: %v", err)
+		}
```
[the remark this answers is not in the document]
pkazmierczak: Right, addressed in [9360c46](https://github.com/hashicorp/nomad/pull/19609/commits/9360c464a8b31ed0ec6d424a282de1e0beb66d04)

nomad/core_sched_test.go
```diff
@@ -609,7 +607,6 @@ func TestCoreScheduler_EvalGC_Batch(t *testing.T) {
 		[]*structs.Allocation{
 			activeJobLostAlloc, activeJobCompletedEvalCompletedAlloc,
 			stoppedJobLostAlloc, stoppedJobLostAlloc,
-			purgedJobCompleteAlloc,
```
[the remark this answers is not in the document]
pkazmierczak: This removal was actually unnecessary, the assertion just checks that these allocs are not there. I will revert the change.

tgross (approved):
LGTM once @lgfa29's comments are resolved

nomad/state/state_store_test.go
```diff
@@ -4277,27 +4277,6 @@ func TestStateStore_CSIPlugin_Lifecycle(t *testing.T) {
...
-		plug = checkPlugin(pluginCounts{
-			controllerFingerprints: 1,
-			nodeFingerprints:       2,
-			controllersHealthy:     1,
-			nodesHealthy:           2,
-			controllersExpected:    0,
-			nodesExpected:          0,
-		})
-		require.True(t, plug.ControllerRequired)
-		require.False(t, plug.IsEmpty())
```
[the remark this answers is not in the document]
pkazmierczak: I think I'm a bit confused here. We can't call `updateAllocsFn` here on any alloc, because this will cause a panic, all allocs for the job have been removed by `store.DeleteJob`.

nomad/state/state_store.go
pkazmierczak: Addressed in [d6849cc](https://github.com/hashicorp/nomad/pull/19609/commits/d6849cc849e8cbd4e5a214435a9fd98fe560b033)

nomad/state/state_store.go
pkazmierczak: 👍 Fixed in [d6849cc](https://github.com/hashicorp/nomad/pull/19609/commits/d6849cc849e8cbd4e5a214435a9fd98fe560b033)

pkazmierczak:
> we usually have `DeleteX()` and `DeleteXTxn()` methods. I think it would nice to keep this pattern here for deployments and allocs.

Not sure I understand, are you suggesting to rename `DeleteJobTxn` to `DeleteJobDeploymentsAndAllocsTxn` or sth like that? This PR doesn't introduce a new method, it just modifies the `DeleteJobTxn`.

nomad/state/state_store_test.go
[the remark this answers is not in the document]
tgross: I said `Node.UpdateStatus` above and I should have said `Node.UpdateAlloc`. My concern was what would happen if we got a `Node.UpdateAlloc` RPC concurrently with the purge, such that the validation checks in the RPC handler happened before the deletion, but then the RPC's Raft entry was operating on the now-deleted job. But I've had a quick check through the FSM + state code for that code path and I'm reasonably confident that's safe.

Changing the `updateAllocsFn` would probably have ripple effects that would make the rests of the test awkward. I might argue it'd be worth trying to reproduce the scenario I described above, but that would be adding a completely separate concern to this test than the plugin counts. So this is probably fine as you've got it.

lgfa29:
> > we usually have `DeleteX()` and `DeleteXTxn()` methods. I think it would nice to keep this pattern here for deployments and allocs.
> 
> Not sure I understand, are you suggesting to rename `DeleteJobTxn` to `DeleteJobDeploymentsAndAllocsTxn` or sth like that? This PR doesn't introduce a new method, it just modifies the `DeleteJobTxn`.

Nope, I was suggesting having `DeleteJobTxn()` look something like this:

```go
func (s *StateStore) DeleteJobTxn(index uint64, namespace, jobID string, txn Txn) error {
  // ...
  for _, deployment := range deployments {
    err := s.DeleteDeploymentTxn(txn, deployment)
    // ...
  }
  // ....
  for _, alloc := range allocs {
    err := s.DeleteAllocTxn(txn, alloc)
    // ...
  }
}
```

That way the deployment and alloc delete logic is shared and consistent across methods (for example, you wouldn't need to update their index table in `DeleteJobTxn`).

nomad/state/state_store.go
```diff
@@ -1894,6 +1894,68 @@ func (s *StateStore) DeleteJobTxn(index uint64, namespace, jobID string, txn Txn
...
+		if existing == nil {
+			continue
+		}
+
+		eval := existing.(*structs.Evaluation).Copy()
+		if eval.Status != structs.EvalStatusPending {
+			continue
+		}
+		eval.Status = structs.EvalStatusComplete
+		eval.StatusDescription = fmt.Sprintf("evaluation deleted while purging job %s", job.ID)
```
lgfa29, line 1933:
```suggestion
		eval.StatusDescription = fmt.Sprintf("job %s deleted", job.ID)
```

The eval has not been deleted.

nomad/state/state_store.go
lgfa29: Minor suggestion in https://github.com/hashicorp/nomad/pull/19609#discussion_r1449107898

pkazmierczak:
> That way the deployment and alloc delete logic is shared and consistent across methods (for example, you wouldn't need to update their index table in `DeleteJobTxn`).

oooh I get it now I think. Implemented in e918e3b, have a look if that's what you meant.

nomad/state/state_store.go
```diff
@@ -1894,6 +1949,67 @@ func (s *StateStore) DeleteJobTxn(index uint64, namespace, jobID string, txn Txn
...
+			return fmt.Errorf("eval lookup failed: %v", err)
+		}
+		if existing == nil {
+			continue
+		}
+
+		eval := existing.(*structs.Evaluation).Copy()
+		if eval.Status != structs.EvalStatusPending {
+			continue
+		}
```
lgfa29, line 1985:
```suggestion
		if existing.Status != structs.EvalStatusPending {
			continue
		}
		eval := existing.(*structs.Evaluation).Copy()
```

This avoids unnecessary memory allocation for evals that will not be touched.

PR #19611, opened 2024-01-04
Error message typo fix: Filed to Failed

philrenaud:


tgross (approved):
You've committed a snap.tmp file here. But once that's fixed this LGTM

PR #19612, opened 2024-01-04
client: add cpu scanner over /proc/cpuinfo

### LINKED ISSUE #19468: CPU fingerprinting with dmidecode fallback could be suboptimal
### Nomad version
Output from `nomad version`

```
Nomad v1.7.1
BuildDate 2023-12-08T18:11:21Z
Revision 608e719430038cdeb5fe108536d90cf88a8540e3
```

Archlinux on amd64, virtualized KVM/QEMU

### Issue

As suggested from @shoenig in #19412 this is an extra issue for the new `dmidecode` fallback for CPU fingerprinting starting with 1.7.
The usage could lead to suboptimal MHz readings.

### Reproduction steps

In virtual environments, at least KVM with qemu virtual CPU Models, which may be default selected for new VMs in different orchestrators, the dmidecode exposes "wrong" MHz while `lscpu` / `/proc/cpuinfo` do expose much better readings.
I have access for example to environments managed with Proxmox, oVirt and Virtuozzo and there these CPUs are use (compatibility stuff).

Maybe the following example is an edge case, but this shows how the `dmidecode` readings could lead to more Mhz than the host could provide.

In my homelab environment i have an (realy old) Intel(R) Celeron(R) CPU 847 @ 1.10GHz with two cores.
This system use QEMU/KVM to host some VMs with emulated CPU "Intel Xeon E312xx (Sandy Bridge, IBRS update)" on which the MHz detection was ok (2195 MHz) before the upgrade to 1.7.1.
After the upgrade the Frequency was 0 MHz until i install dmidecode as suggested in #19412.

Now I have 2x 2000 MHz because of the `dmidecode` reporting (only the relevant Information):

```
Processor Information
        Socket Designation: CPU 0
        Type: Central Processor
        Family: Other
        Manufacturer: QEMU
        Version: pc-q35-4.0
        Max Speed: 2000 MHz
        Current Speed: 2000 MHz
        Core Count: 1
        Thread Count: 1

Processor Information
        Socket Designation: CPU 1
        Type: Central Processor
        Family: Other
        Manufacturer: QEMU
        Version: pc-q35-4.0
        Max Speed: 2000 MHz
        Current Speed: 2000 MHz
        Core Count: 1
        Thread Count: 1
```

The `lscpu -Je=MHZ` and `grep -i mhz /proc/cpuinfo` still shows the real numbers:

```
{
   "cpus": [
      {
         "mhz": 1097,5060
      },{
         "mhz": 1097,5060
      }
   ]
}

cpu MHz         : 1097.506
cpu MHz         : 1097.506
```

So in this Case the detected MHz per `dmidecode` is nearly 2 times what the host real provides.

#### Expected Result

The Mhz should be detected around 2195.

#### Actual Result

The MHz is detected as 4000.

shoenig:
This PR restores a previous behavior where CPU performance characteristics
were also scanned from the `/proc/cpuinfo` special file.

These numbers are not reliable - they represent the live processor speed
for each core at the time of sampling. To hopefully find a reasonable
guess of the base speed we average the sample across all cores and use
that value. As such these values are the used as the 3rd tier of precedence,
where first try get values from sysfs, then smbios, then here.

Fixes #19468

### DIFF
--- a/.changelog/19612.txt
+++ b/.changelog/19612.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+client: Fixed a bug where cpu fingerprinter would not consider /proc/cpuinfo
+```
--- a/client/lib/numalib/detect_linux.go
+++ b/client/lib/numalib/detect_linux.go
@@ -6,8 +6,11 @@
 package numalib
 
 import (
+	"bufio"
 	"fmt"
+	"io"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -21,6 +24,7 @@ func PlatformScanners() []SystemScanner {
 	return []SystemScanner{
 		new(Sysfs),
 		new(Smbios),
+		new(Cpuinfo),
 		new(Cgroups1),
 		new(Cgroups2),
 		new(Fallback),
@@ -47,6 +51,9 @@ type pathReaderFn func(string) ([]byte, error)
 // from /sys/devices/system. This is the best source of truth on Linux and
 // should always be used first - additional scanners can provide more context
 // on top of what is initiallly detected here.
+//
+// Any data from Sysfs takes priority over Smbios.
+// Any data from Smbios takes priority over Cpuinfo.
 type Sysfs struct{}
 
 func (s *Sysfs) ScanSystem(top *Topology) {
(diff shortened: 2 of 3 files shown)


shoenig:
We basically did this in the 1.7.2 fix via the fallback mechanism.

PR #19630, opened 2024-01-05
cli: Fix dummy FSM create to ensure snapshot state command works.

jrasell:
The Nomad state store function was recently updated to validate certain parameters, fixing a panic condition. This change meant dummy FSM used for the snapshot state command was always failing this validation and the command no longer worked.

This change adds the required parameter to pass validation and therefore makes the CLI command functional again.


tgross (approved):
LGTM

Issue #13139, opened 2022-05-26
consul: support for admin partitions


shoenig:
Consul 1.11 added support for [admin partitions](https://learn.hashicorp.com/tutorials/consul/consul-admin-partitions).

Similar to the work done for adding support for Consul namespaces, Nomad should also add support for Consul admin partitions.

mikenomitch:
Wanted to comment on the current state of things for anybody interested. Currently, Nomad doesn’t “know” about Admin Partitions, but it still works with them. There’s nothing that should break using the two together assuming you split up nodes and workloads properly. The question is, how do you split that up? The rough steps would be something like:

- Step 1: Get Consul Admin Paritions running and split up your Consul nodes into partitions as you see fit.
- Step 2: Tag the nodes in Nomad appropriately so that you can send specific workloads to specific nodes. I think the easiest way to do this is to use datacenters. If Admin Paritions map directly to Nomad datacenters, then it is really easy to know which Nomad node maps to which admin partition.
- Step 3: When deploying Nomad jobs, make sure that you send in the right Consul creds for the right Admin Partition. If you are passing in a CONSUL_TOKEN that has access to that partition.

Notes:

- You don’t have to use datacenter to do this split. You could technically just use node_class or metadata to define each partition and then make sure Nomad workloads are placed onto the right node with those. In my opinion, this is more error-prone.
- Splitting up Nomad clusters would also work. Each Admin partition could be a different Nomad cluster.
- Nomad might introduce the concept of “node pools” at some point which would function somewhat like datacenter in terms of how you can split nodes up. This might eventually become the suggested way.

More official support for admin partitions is something that we've looked at briefly, but not explored in depth. If you have a use case where Nomad having first-class knowledge of admin partitions would be helpful, please let us know. A common use case where splitting by datacenter, cluster, or node_class is not sufficient would help us make the case for this internally.

davidfleming:
Hi @mikenomitch 

First thank you for commenting on this.  We have been looking forward to seeing progress on this and a discussion on the direction this will potentially take would be appreciated.

Our use case:

[...]

PR #19485, opened 2023-12-14
consul: fingerprint Consul Enterprise admin partitions

tgross:
Consul Enterprise agents all belong to an admin partition. Fingerprint this attribute when available. When a Consul agent is not explicitly configured with "default" it is in the default partition but will not report this in its `/v1/agent/self` endpoint. Fallback to "default" when missing only for Consul Enterprise.

This feature provides users the ability to add constraints for jobs to land on Nomad nodes that have a Consul in that partition. Or it can allow cluster administrators to pair Consul partitions 1:1 with Nomad node pools. We'll also have the option to implement a future `partition` field in the jobspec's `consul` block to create an implicit constraint.

Ref: https://github.com/hashicorp/nomad/issues/13139#issuecomment-1856479581

PR #19665, opened 2024-01-08
consul: support admin partitions

tgross:
Add support for Consul Enterprise admin partitions. We added fingerprinting in https://github.com/hashicorp/nomad/pull/19485. This PR adds a `consul.partition` field. The expectation is that most users will create a mapping of Nomad node pool to Consul admin partition. But we'll also create an implicit constraint for the fingerprinted value.

Fixes: https://github.com/hashicorp/nomad/issues/13139

### DIFF
--- a/nomad/job_endpoint_hook_consul_ce.go
+++ b/nomad/job_endpoint_hook_consul_ce.go
@@ -8,6 +8,7 @@ package nomad
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/hashicorp/nomad/nomad/structs"
 )
@@ -28,7 +29,11 @@ func (h jobConsulHook) Validate(job *structs.Job) ([]error, error) {
 	}
 
 	for _, group := range job.TaskGroups {
+
+		partition := ""
+
 		if group.Consul != nil {
+			partition = group.Consul.Partition
 			if err := h.validateCluster(group.Consul.Cluster); err != nil {
 				return nil, err
 			}
@@ -56,6 +61,12 @@ func (h jobConsulHook) Validate(job *structs.Job) ([]error, error) {
 			}
 
 			if task.Consul != nil {
+				if task.Consul.Partition != "" {
+					if partition != "" && task.Consul.Partition != partition {
+						return nil, fmt.Errorf("task.consul.partition %q must match group.consul.partition %q if both are set", task.Consul.Partition, partition)
+					}
+				}
+
 				if err := h.validateCluster(task.Consul.Cluster); err != nil {
 					return nil, err
 				}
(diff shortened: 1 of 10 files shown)


nomad/job_endpoint_hook_consul_ce_test.go
```diff
@@ -48,6 +51,13 @@ func TestJobEndpointHook_ConsulCE(t *testing.T) {
 	test.Eq(t, "infra", job.TaskGroups[0].Services[1].Cluster)
 	test.Eq(t, "nondefault", job.TaskGroups[0].Tasks[0].Services[0].Cluster)
 
+	test.SliceContains(t, job.TaskGroups[0].Constraints,
+		&structs.Constraint{
+			LTarget: "${attr.consul.partition}",
```
gulducat, line 56: This tests the presence of the new constraint for the default Consul cluster, but I'm not seeing any test for a non-default Consul cluster `attr.consul.%s.partition`?

gulducat (approved):
An e2e test for this might be neat, and I think there's a little unit test gap (unless I'm missing something), but otherwise LGTM!

nomad/job_endpoint_hook_consul_ce_test.go
tgross: Technically we can't have a non-default cluster in CE and having the per-cluster constraint in the first place is probably moot. But can't hurt to add a test here for it. There'll be a follow-up ENT PR as well.

tgross: Actually, I'm going to back up a bit here and remove the conditional from CE, because it's never valid so there's no point in doing that. We'll only see that in ENT.

tgross: Done in 908b445

gulducat: Looks good!  Will `consulPartitionConstraint()` have the same signature in both?  Doesn't it (maybe) need `cluster` in ENT?

tgross: It does need the `cluster` field in ENT, so that's why I moved it into the CE file.

gulducat: OH!  I missed that alllll the logic ended up in the CE file.  Don't mind me :sweat_smile:

Juanadelacuesta (approved):
I would find a way to move the empty verifications into a function, to make the code cleaner, but that is nitpicking. It looks good

nomad/job_endpoint_hook_consul_ce.go
```diff
@@ -28,7 +29,11 @@ func (h jobConsulHook) Validate(job *structs.Job) ([]error, error) {
 	}
 
 	for _, group := range job.TaskGroups {
+
+		partition := ""
```
jrasell, line 33: nit: change this to groupPartition, so it's a little easier to understand the comparison on L36. It took me a moment to figure out what was happening.

jrasell (approved):
LGTM; exceptionally minor inline comment.

PR #19670, opened 2024-01-08
drivers/raw_exec: enable configuring raw_exec task to have no memory limit

shoenig:
This PR makes it possible to configure a raw_exec task to not have an
upper memory limit, which is how the driver would behave pre-1.7.

This is done by setting memory_max = -1. The cluster (or node pool) must
have memory oversubscription enabled.

Closes #19670

Backport to 1.7 so users can re-enable behavior from before.

### DIFF
--- a/drivers/shared/executor/executor_universal_linux.go
+++ b/drivers/shared/executor/executor_universal_linux.go
@@ -1,6 +1,8 @@
 // Copyright (c) HashiCorp, Inc.
 // SPDX-License-Identifier: MPL-2.0
 
+//go:build linux
+
 package executor
 
 import (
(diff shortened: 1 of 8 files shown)


drivers/shared/executor/executor_universal_linux.go
```diff
@@ -226,7 +234,11 @@ func (e *UniversalExecutor) configureCG2(cgroup string, command *ExecCommand) {
 	// write memory cgroup files
 	memHard, memSoft := e.computeMemory(command)
 	ed := cgroupslib.OpenPath(cgroup)
-	_ = ed.Write("memory.max", strconv.FormatInt(memHard, 10))
+	if memHard == memoryNoLimit {
+		_ = ed.Write("memory.max", "max")
```
gulducat, line 238: just noting here that "max" is a magic word indicating no limit, ref:

https://git.kernel.org/pub/scm/linux/kernel/git/tj/cgroup.git/tree/Documentation/admin-guide/cgroup-v2.rst#n667

> Limits are in the range [0, max] and defaults to "max", which is noop.

gulducat (approved):
LGTM!

PR #19677, opened 2024-01-09
acl: ensure auth method configs are correctly and fully hashed.

### LINKED ISSUE #19668: Updating OIDC scope of `AuthMethod` fails silently
While working on testing https://github.com/hashicorp/nomad/issues/19318 I discovered that updating the `OIDCScope` field of a `AuthMethod` via `nomad acl auth-method update` doesn't appear to update the scopes. I took a _very_ quick look through the code and didn't see any obvious reason for it, but the way we update hashes in the RPC handler looks suspicious to me. I'll follow-up with more.

jrasell:
Adds all auth method config fields to the hash function, to ensure user initiated changes are not dropped.

closes #19668

### DIFF
--- a/.changelog/19677.txt
+++ b/.changelog/19677.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+acl: Fixed auth method hashing which meant changing some fields would be silently ignored
+```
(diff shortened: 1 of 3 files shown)


pkazmierczak (approved):
LGTM!

tgross (approved):
LGTM! Thanks for picking this up!

PR #19681, opened 2024-01-09
docs: expand docs on implicit ACL capabilities grants

tgross:
An audit of Nomad's ACLs resulted in some confusion around whether the `NamespaceValidator` method is conjunctive ("add", as implied by the docs) or disjunctive ("or", as it is by design). Clarify the ACL documentation as follows:

* Call out where fine-grained capabilities imply grants to other capabilities (for example, that `csi-read-volume` grants `csi-list-volume`).
* Fix an incorrectly documented ACL requirement for the CSI List External Volumes API.
* Clarify how ACLs are expected to work for the two search API endpoints, such that you need list/read access to the objects in the search context.

### DIFF
--- a/website/content/api-docs/search.mdx
+++ b/website/content/api-docs/search.mdx
@@ -22,14 +22,15 @@ The table below shows this endpoint's support for
 [blocking queries](/nomad/api-docs#blocking-queries) and
 [required ACLs](/nomad/api-docs#acls).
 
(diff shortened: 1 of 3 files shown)


dduzgun-security (approved):

gulducat (approved):
hooray for explicitly documenting implicit behavior!

schmichael (approved):

PR #19682, opened 2024-01-09
Add systemd time format for periodic jobs

aneustroev:


aneustroev:
If everything is ok, I will add changelog and documentations.

PR #19689, opened 2024-01-09
vault: fix token revocation during workflow migration

lgfa29:
When transitioning from the legacy token-based workflow to the new JWT workflow for Vault the previous code would instantiate a no-op Vault if the server configuration had a `default_identity` block.

This no-op client returned an error for some of its operations were called, such as `LookupToken` and `RevokeTokens`. The original intention was that, in the new JWT workflow, none of these methods should be called, so returning an error could help surface potential bugs.

But the `RevokeTokens` and `MarkForRevocation` methods _are_ called even in the JWT flow. When a leadership transition happens, the new server looks for unused Vault accessors from state and tries to revoke them. Similarly, the `RevokeTokens` method is called every time the `Node.UpdataStatus` and `Node.UpdateAlloc` RPCs are made by clients, as the Nomad server tries to find unused Vault tokens for the node/alloc.

Since the new JWT flow does not require Nomad servers to contact Vault, calling `RevokeTokens` and `MarkForRevocation` is not able to complete without a Vault address and token, so this commit changes the logic to use the no-op Vault client when one of them is not passed. It also updates the client itself to not error if these methods are called, but rather raise a warning log so operators can be made aware that there are Vault tokens created by Nomad that have not been force-expired.

There are also updates to the documentation of the migration process. When migrating an existing cluster to the new workload identity based flow, Nomad operators must first upgrade the Nomad version without removing any of the existing Vault configuration. Doing so can prevent Nomad servers from managing and cleaning-up existing Vault tokens during a leadership transition and node or alloc updates.

Operators must also resubmit all jobs with a `vault` block so they are updated with an `identity` for Vault. Skipping this step may cause allocations to fail if their Vault token expires (if, for example, the Nomad client stops running for TTL/2) or if they are rescheduled, since the new client will try to follow the legacy flow which will fail if the Nomad server configuration for Vault has already been updated to remove the Vault address and token.

### DIFF
--- a/nomad/server.go
+++ b/nomad/server.go
@@ -893,7 +893,24 @@ func (s *Server) Reload(newConfig *Config) error {
 
 	// Handle the Vault reload. Vault should never be nil but just guard.
 	if s.vault != nil {
-		if err := s.vault.SetConfig(newConfig.GetDefaultVault()); err != nil {
+		vconfig := newConfig.GetDefaultVault()
+
+		// Verify if the new configuration would cause the client type to
+		// change.
+		var err error
+		switch s.vault.(type) {
+		case *NoopVault:
+			if vconfig != nil && vconfig.Token != "" {
+				err = fmt.Errorf("setting a Vault token requires restarting the Nomad agent")
+			}
+		case *vaultClient:
+			if vconfig != nil && vconfig.Token == "" {
+				err = fmt.Errorf("removing the Vault token requires restarting the Nomad agent")
+			}
+		}
+		if err != nil {
+			_ = multierror.Append(&mErr, err)
+		} else if err := s.vault.SetConfig(newConfig.GetDefaultVault()); err != nil {
 			_ = multierror.Append(&mErr, err)
 		}
 	}
@@ -1192,8 +1209,8 @@ func (s *Server) setupConsul(consulConfigFunc consul.ConfigAPIFunc, consulACLs c
 // setupVaultClient is used to set up the Vault API client.
 func (s *Server) setupVaultClient() error {
 	vconfig := s.config.GetDefaultVault()
-	if vconfig != nil && vconfig.DefaultIdentity != nil {
-		s.vault = &NoopVault{}
+	if vconfig != nil && vconfig.Token == "" {
+		s.vault = NewNoopVault(s.logger)
 		return nil
 	}
 
--- a/nomad/vault_noop.go
+++ b/nomad/vault_noop.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 	"time"
 
+	log "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/nomad/nomad/structs"
 	"github.com/hashicorp/nomad/nomad/structs/config"
 	vapi "github.com/hashicorp/vault/api"
@@ -17,6 +18,13 @@ import (
 type NoopVault struct {
 	l      sync.Mutex
 	config *config.VaultConfig
+	logger log.Logger
+}
+
+func NewNoopVault(logger log.Logger) *NoopVault {
+	return &NoopVault{
+		logger: logger.Named("vault-noop"),
+	}
 }
 
 func (v *NoopVault) SetActive(_ bool) {}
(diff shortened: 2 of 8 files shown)


nomad/server.go
```diff
@@ -1192,8 +1192,8 @@ func (s *Server) setupConsul(consulConfigFunc consul.ConfigAPIFunc, consulACLs c
 // setupVaultClient is used to set up the Vault API client.
 func (s *Server) setupVaultClient() error {
 	vconfig := s.config.GetDefaultVault()
-	if vconfig != nil && vconfig.DefaultIdentity != nil {
-		s.vault = &NoopVault{}
+	if vconfig != nil && vconfig.Addr == config.DefaultVaultAddr && vconfig.Token == "" {
```
lgfa29, line 1195: We unfortunately can't check for an empty address string here since it's always set to this default. But I would argue that it's kind of a bad default, and I expect people that are actually using Vault to change this default address.

We must check both values because some environments may be deploy Vault "tokenless" (https://github.com/hashicorp/nomad/issues/19380).

nomad/vault_noop.go
```diff
@@ -44,12 +52,18 @@ func (v *NoopVault) LookupToken(_ context.Context, _ string) (*vapi.Secret, erro
 	return nil, errors.New("Vault client not able to lookup tokens")
 }
 
-func (v *NoopVault) RevokeTokens(_ context.Context, _ []*structs.VaultAccessor, _ bool) error {
-	return errors.New("Vault client not able to revoke tokens")
+func (v *NoopVault) RevokeTokens(_ context.Context, tokens []*structs.VaultAccessor, _ bool) error {
+	for _, t := range tokens {
+		v.logger.Warn("Vault token is no longer used, but Nomad is not able to revoke it. The token may need to be revoked manually or will expire once its TTL reaches zero.", "accessor", t.Accessor, "ttl", t.CreationTTL)
```
lgfa29, line 57: I'm not sure about these log messages. Not sure about their level (`WARN`), the actual text, or if they should even be emitted in the first place 😅

website/content/docs/integrations/vault/acl.mdx
```diff
@@ -751,18 +751,26 @@ $ VAULT_TOKEN=s.H39hfS7eHSbb1GpkdzOQLTmz.fvuLy nomad job run vault.nomad
...
-* (Optionally) Add [`vault.role`][] fields to any Nomad jobs that will not use
-  the default role.
-* (Optionally) add [`identity`][] blocks to your jobs if you want to use a
-  different identity because of how your auth method and roles are configured.
+  configurations, but **do not modify any of the existing Vault
+  configuration**.
+* Upgrade your cluster following the documented [Upgrade
+  Process][docs_upgrade].
+* Resubmit Nomad jobs that need access to Vault to redeploy them with a new
+  workload identity for Vault.
```
lgfa29, line 766: I was trying to think of ways we could help with this. Two ideas I had were:

1. A command that looks for the following artifacts and tell users if their clusters are ready for the migration.
  * jobs with `vault` blocks (and maybe `templates` with Vault-related functions? A `vault` block is not always required to access Vault) but no Vault identities.
  * unused `VaultAccessors` in state store (and potentially clean them without needing a leadership transition)
2. Some kind of metric that operators can monitor and wait until it reaches zero. But I'm not sure yet exactly what to measure.

nomad/server.go
tgross: Just to make sure that playing around in dev mode works with the legacy workflow and an env var for `VAULT_TOKEN`, we probably should leave the `DefaultIdentity` condition in there and just add the new ones. So:

```suggestion
	if vconfig != nil && vconfig.DefaultIdentity != nil && vconfig.Addr == config.DefaultVaultAddr && vconfig.Token == "" {
```

nomad/vault_noop.go
tgross: The text you've got here is great. But yeah, maybe `DEBUG` level? It's only actionable inasmuch as some cluster operators will want to go clean these up manually, but they don't _need_ to.

website/content/docs/integrations/vault/acl.mdx
tgross: I like the idea of a command... maybe we could bake this into a `-check` flag for `nomad setup vault`? 

A metric would be nice, but the only reasonable place for it would be in the state store and we'd have to repopulate it during snapshot restore. Kind of messy.

tgross (approved):
LGTM once the comments I've made are resolved.

nomad/vault_noop.go
lgfa29: Dropped to `DEBUG`.

website/content/docs/integrations/vault/acl.mdx
lgfa29: Cool! I will try to think a bit more where to place the command. I was thinking somewhere under `nomad operator` to indicate that this may require a management token since it needs to look all over the state store.

It could also be used for other upgrade checks in the future.

nomad/server.go
tgross: Yeah that makes sense. There's no such thing as Vault without ACLs.

Lord-Y:
We just hit this problem. We had rollback our config without workload identity. We'll wait for the next release.

Issue #8690, opened 2020-08-19
Add an option to disable client Vault token renewal


roman-vynar:
For Vault integration, Nomad requires that the client tokens are renewable according to the role spec like that:
```
nomad_cluster_role = {
    "allowed_policies": allowed_policies,
    "no_default_policy": True,
    "token_explicit_max_ttl": 0,  # required
    "name": "nomad-cluster",
    "orphan": True,
    "token_period": 60,
    "renewable": True  # required
}
```

However, all our application services that we deploy with Nomad read secrets from Vault during the startup, keep in memory and Vault token issued by Nomad is no longer needed. In our case we use env var VAULT_TOKEN set by Nomad into a task docker container.

Moreover, from security perspective it's not good to keep an env var VAULT_TOKEN or something like that persistent.
That's why we have set token_period=60 and patched Nomad with an easy fix to skip token renewal by doing nothing:
https://github.com/Quiq/nomad/commit/6571064d8b01f9b10e828fdd481225a1752f7a18

It works perfectly for us, a server vault token is renewed ok, a client vault token expires after 60s, it is enough time for application to read secrets.. Of course, we account that a task allocation restart is not possible because of expired vault token but it's ok for us and we do re-scheduling instead when a new alloc is created with a new vault token issued.

Thanks!

tgross:
Hi @roman-vynar! This is an interesting use case but I think it needs a bit more consideration.

mriddell-foundry:
Hi @tgross. Is there any updates on this sort of feature?

My team has a use case where we have some long running batch data processing tasks which require a secret (api key) to retrieve the initial set of data then processes that data for x amount of time afterwards.

The secret is only required for about 10 minutes during the fetching of the data but not during the processing itself which can take some time.
I feel this could be an adequate use case for this feature but was curious if there was any movement on this item.

Thanks!

tgross:
Hi @mriddell-foundry! We've got an open issue for a refresh of our Vault integration in https://github.com/hashicorp/nomad/issues/15617 and this might fall under there. That refresh is (very roughly) scheduled for the 1.7 release, which means it's not coming until late in the year at the earliest.

tgross:
Making a note here that starting in [Nomad 1.7.0-beta.1](https://github.com/hashicorp/nomad/releases/tag/v1.7.0-beta.1) we've got a new workflow where Nomad will use workload identities to sign into Vault for purposes of getting Vault tokens for those workloads. 

[...]

PR #19691, opened 2024-01-09
vault: support allowing tokens to expire without refresh

tgross:
Some users with batch workloads or short-lived prestart tasks want to derive a Vaul token, use it, and then allow it to expire without requiring a constant refresh. Add the `vault.allow_token_expiration` field, which works only with the Workload Identity workflow and not the legacy workflow.

When set to true, this disables the client's renewal loop in the `vault_hook`. When Vault revokes the token lease, the token will no longer be valid. The client will also now automatically detect if the Vault auth configuration does not allow renewals and will disable the renewal loop automatically.

Note this should only be used when a secret is requested from Vault once at the start of a task or in a short-lived prestart task. Long-running tasks should never set `allow_token_expiration=true` if they obtain Vault secrets via `template` blocks, as the Vault token will expire and the template runner will continue to make failing requests to Vault until the `vault_retry` attempts are exhausted.

Fixes: https://github.com/hashicorp/nomad/issues/8690


website/content/docs/job-specification/vault.mdx
```diff
@@ -58,6 +58,17 @@ with Vault as well.
...
+- `allow_token_expiration` `(bool: false)` - Specifies that Nomad clients should
+  not attempt to renew a task's Vault token, allowing it to expire. This should
+  only be used when a secret is requested from Vault once at the start of a task
+  or in a short-lived prestart task. Long-running tasks should never set
+  `allow_token_expiration=true` if they obtain Vault secrets via `template`
+  blocks, as the Vault token will expire and the template runner will continue
+  to make failing requests to Vault until its [`vault_retry`][] attempts are
+  exhausted, at which point the task will fail. This field is valid only when
+  Nomad has been configured to use [Workload Identity with Vault][] and will be
+  ignored when using the [legacy Vault authentication workflow][].
```
lgfa29, line 70: It's not fully ignored. I think the token property is ignored, but the config still works even with the legacy workflow.

It could also be useful to mention that, if the Vault token is already configured as non-reneweable then you don't need to set this.

client/allocrunner/taskrunner/vault_hook.go
```diff
@@ -442,6 +455,12 @@ func (h *vaultHook) deriveVaultTokenJWT() (string, error) {
 		)
 	}
 
+	// If the token cannot be renewed, it doesn't matter if the user set
+	// allow_token_expiration or not, so override the requested behavior
+	if !renewable {
+		h.allowTokenExpiration = false
```
lgfa29, line 461:
```suggestion
		h.allowTokenExpiration = true
```
I think this needs to be true? If the token cannot be renewed then we should allow it to expire.

lgfa29 (approved):
I think the boolean value is wrong when the token is not renewable, but otherwise LGTM!

It could also be useful to add a note about batch tokens to https://developer.hashicorp.com/nomad/docs/integrations/vault/acl.

client/allocrunner/taskrunner/vault_hook.go
tgross: Oops, you're right. That means I'm also missing a test. Fixed and added a test case for it.

website/content/docs/job-specification/vault.mdx
tgross: Done in 5798001 along with your suggestion to add a note to the Vault ACL docs. How's that look to you?

tgross: Expanded on that some more in 183d356

PR #19697, opened 2024-01-10
e2e: remove broken Consul WI test

pkazmierczak:
This test will not work in its current shape, since we use HCP Consul in our e2e suite and it won't connect to Nomad servers' JWKS endpoints over TLS.

https://github.com/hashicorp/nomad/issues/19698 tracks future fate of this test.


tgross (approved):
:+1:

PR #19699, opened 2024-01-10
consul: refactor job mutation hook

tgross:
The job mutation logic for Nomad CE and Nomad ENT are nearly identical except for a prelude that grabs the correct default cluster. Factor this out into a method that can be shared between both code bases.

Ref: discussion in https://github.com/hashicorp/nomad-enterprise/pull/1367/files#r1447861412

### DIFF
--- a/nomad/job_endpoint_hook_consul.go
+++ b/nomad/job_endpoint_hook_consul.go
@@ -30,3 +30,63 @@ func (jobConsulHook) validateTaskPartitionMatchesGroup(groupPartition string, ta
 	}
 	return nil
 }
+
+// mutateImpl ensures that the job's Consul blocks have been configured with the
+// correct Consul cluster if unset, and sets constraints on the Consul admin
+// partition if set. This should be called by the Mutate method.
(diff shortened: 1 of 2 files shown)


nomad/job_endpoint_hook_consul.go
```diff
@@ -30,3 +30,63 @@ func (jobConsulHook) validateTaskPartitionMatchesGroup(groupPartition string, ta
...
+			}
+			for _, service := range task.Services {
+				if service.IsConsul() && service.Cluster == "" {
+					service.Cluster = defaultCluster
+				}
+			}
+		}
+	}
+
+	return job, nil, nil
```
gulducat, line 73: Feels a little strange for the function signature to include []error and error, but for there to be no condition where they're not nil.  The `Mutate()` function needs to, to implement the interface, but here it strikes me as odd.

gulducat (approved):
I love me some code de-duplication, thanks!

nomad/job_endpoint_hook_consul.go
tgross: Eh, good point. The ENT version potentially returns errors but it can do that in the call. Will fix.

PR #19703, opened 2024-01-10
client/allocdir: use an interface in place of AllocDir structs

shoenig:
This PR replace `*allocdir.AllocDir` with `allocdir.Interface` such that we
may eventually have another implementation of alloc directories. This is
in support of the exec2 driver, which will need an implementation of the
alloc directory incompatible with the current version.

Preliminary work for https://github.com/hashicorp/team-nomad/issues/373

No backport; for 1.8


tgross (approved):
LGTM

PR #19706, opened 2024-01-11
vault: remove revoked Vault accessors from state

lgfa29:
Following up from https://github.com/hashicorp/nomad/pull/19689, when using the no-op Vault client the Nomad server still needs to delete the revoked Vault accessors from state to prevent them from lingering forever after the cluster migrates to the workload identity flow.

No changelog since https://github.com/hashicorp/nomad/pull/19689 has not been released yet.


pkazmierczak (approved):
👍

PR #19707, opened 2024-01-11
build: use setup-golang@v3 to handle auto caching

shoenig:
Use the latest hashicorp/setup-golang @ v3 which has been fixed to handle the fact that the actions/setup-go action automatically sets up the caching that we had to do manually before.

This should cleanup a bunch of errors showing in recent actions outputs and maybe prevent some double caching.


tgross (approved):
:+1:

PR #19717, opened 2024-01-11
e2e: move rawexec oversub tests into oversubscription e2e test suite

shoenig:
This PR moves two tests for raw_exec and memory oversubscription into
the oversubscription test suite, which has the necessary plumbing to
activate and restore the oversubscription configuration of the scheduler
during the test.


e2e/oversubscription/input/oversub.hcl
gulducat, line 1: `oversub.hcl` is vague in this new context -- others in this dir include `docker.hcl` and `exec.hcl` so maybe these two could be `raw_exec.hcl` and `raw_exec_max.hcl`?

gulducat (approved):
more specific filenames would be nice, but LGTM!

e2e/oversubscription/input/oversub.hcl
shoenig: You and your reasonable demands!

gulducat: I'm glad you correctly interpreted my "maybe" and "could be" as "I insist" and "_must_ be" (and a heavily implied "_or else._")

PR #19721, opened 2024-01-11
docs: fix Consul ACL requirements

lgfa29:
Even with the new workload identitiy based flow the Nomad servers still need the `acl = "write"` permission in order to revoke service identity tokens.


tgross (approved):
LGTM

PR #19779, opened 2024-01-19
Clarify some docs pages

sofixa:
Hey everyone,

A couple of small clarifications in a couple of docs:

- add a warning to `templates` that using Nomad/Consul/Vault results in a heavy dependency
- clarify that HCL2 input variables and Nomad Variables are different
- improve the wording on the `template` source param

Adrian


tgross (approved):
LGTM after I've pushed some markdown syntax fixes and reflows.

PR #19787, opened 2024-01-20
connect: rewrite envoy bootstrap on every restart

### LINKED ISSUE #19781: Envoy proxies always crash after a restart of the host
Hello :)

### Issue

After the restart of an instance, the envoy proxies are crashing.
The error in the log is:
`error initializing configuration '/secrets/envoy_bootstrap.json': Invalid path: /secrets/envoy_bootstrap.json`.
A strong hypothesis is that /secrets is a tmpfs and thus it's wiped when restarting.

The envoy_bootstrap.json is created by nomad directly in the /secrets. It doesn't seem to be possible to change that: https://github.com/hashicorp/nomad/blob/fce30f342c681343fdfe712551736a62efc609e3/client/allocrunner/taskrunner/envoy_bootstrap_hook.go#L272
And the /secrets is always using tmpfs: https://github.com/hashicorp/nomad/blob/fce30f342c681343fdfe712551736a62efc609e3/client/allocdir/fs_linux.go#L63

Is there any way we can bypass this problem ?
We use nomad in a context where there is restarts and it would be great to not have to wait for all proxies to be rescheduled.

It also prevents us from using /secrets to store secret files since it would mean we don't have it anymore after a restart and must reschedule to rebuild it.

### Reproduction steps

Start any job with a consul connect proxy on an instance and restart the instance without draining the node.
It may not work if nomad is in -dev since it may drain the node when stopping.

### Nomad version
```
Nomad v1.5.6
BuildDate 2023-05-19T18:26:13Z
Revision 8af70885c02ab921dedbdf6bc406a1e886866f80
```

Linux 4.18.0-372.9.1.el8.x86_64

Thanks for your help :)

schmichael:
Fixes #19781

Do not mark the envoy bootstrap hook as done after successfully running once. Since the bootstrap file is written to /secrets, which is a tmpfs on supported platforms, it is not persisted across reboots. This causes the task and allocation to fail on reboot (see #19781).

This fixes it by *always* rewriting the envoy bootstrap file every time the Nomad agent starts. This does mean we may write a new bootstrap file to an already running Envoy task, but in my testing that doesn't have any impact.

*Alternative 1: Use a regular file*

An alternative approach would be to write the bootstrap file somewhere other than the tmpfs, but this is *unsafe* as when Consul ACLs are enabled the file will contain a secret token:
https://developer.hashicorp.com/consul/commands/connect/envoy#bootstrap

*Alternative 2: Detect if file is already written*

An alternative approach would be to detect if the bootstrap file exists, and only write it if it doesn't.

This is just a more complicated form of the current fix. I think in general in the absence of other factors task hooks should be idempotent and therefore able to rerun on any agent startup. This simplifies the code and our ability to reason about task restarts vs agent restarts vs node reboots by making them all take the same code path.

### DIFF
--- a/client/allocrunner/interfaces/task_lifecycle.go
+++ b/client/allocrunner/interfaces/task_lifecycle.go
@@ -87,6 +87,15 @@ type TaskPrestartResponse struct {
 
 	// Done lets the hook indicate that it completed successfully and
 	// should not be run again.
+	//
+	// Use sparringly! In general hooks should be idempotent and therefore Done
+	// is unneeded. You never know at what point an agent might crash, and it can
+	// be hard to reason about how Done=true impacts agent restarts and node
+	// reboots. See #19787 for example.
+	//
+	// Done is useful for expensive operations such as downloading artifacts, or
+	// for operations which might fail needlessly if rerun while a node is
+	// disconnected.
 	Done bool
 }
 
--- a/.changelog/19787.txt
+++ b/.changelog/19787.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+connect: Fixed envoy sidecars being unable to restart after node reboots
+```
--- a/client/allocrunner/taskrunner/connect_native_hook.go
+++ b/client/allocrunner/taskrunner/connect_native_hook.go
@@ -119,8 +119,8 @@ func (h *connectNativeHook) Prestart(
 	merge(environment, h.bridgeEnv(request.TaskEnv.EnvMap))
 	merge(environment, h.hostEnv(request.TaskEnv.EnvMap))
 
-	// tls/acl setup for native task done
-	response.Done = true
+	// tls/acl setup for native task done but since SecretsDir is a tmpfs, don't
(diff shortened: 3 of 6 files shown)


client/allocrunner/interfaces/task_lifecycle.go
```diff
@@ -87,6 +87,15 @@ type TaskPrestartResponse struct {
...
+	//
+	// Use sparringly! In general hooks should be idempotent and therefore Done
+	// is unneeded. You never know at what point an agent might crash, and it can
+	// be hard to reason about how Done=true impacts agent restarts and node
+	// reboots. See #19787 for example.
+	//
+	// Done is useful for expensive operations such as downloading artifacts, or
+	// for operations which might fail needlessly if rerun while a node is
+	// disconnected.
 	Done bool
```
shoenig, line 99: great comment, this actually clarifies a lot

shoenig (approved):

tgross (approved):
LGTM once that last test is passing

PR #19820, opened 2024-01-26
remove github.com/hashicorp/nomad/helper/pointer in github.com/hashicorp/nomad/jobspec

kevinschoonover:
the hashicorp/nomad/jobspec library is licensed as MPL-2.0; however, hashicorp/nomad/jobspec imports github.com/hashicorp/nomad/helper/pointer ([src](https://github.com/hashicorp/nomad/blob/fce30f342c681343fdfe712551736a62efc609e3/jobspec/parse_task.go#L15)). github.com/hashicorp/nomad/helper/pointer is licensed at the repo root, and thus BUSL. This removes the dependency on nomad/helper/pointer to address https://github.com/pulumi/pulumi-nomad/issues/193.

### DIFF
--- a/jobspec/parse_task.go
+++ b/jobspec/parse_task.go
@@ -12,7 +12,6 @@ import (
 	"github.com/hashicorp/hcl"
 	"github.com/hashicorp/hcl/hcl/ast"
 	"github.com/hashicorp/nomad/api"
-	"github.com/hashicorp/nomad/helper/pointer"
 	"github.com/mitchellh/mapstructure"
 )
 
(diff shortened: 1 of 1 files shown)


lgfa29:
Hi @kevinschoonover 👋 

Thank you for catching this. I'm hoping to get `github.com/hashicorp/nomad/helper/pointer` licensed back to MPL-2.0 since it's an useful packaged used in several other places, including other MPL-2.0 packages.

I've opened https://github.com/hashicorp/nomad/pull/19832 to track this, and change back the license of several other packages in the same situation. I also added a Semgrep rule to help us detect and prevent this from happening again. I couldn't fix all imports right now, so I opened https://github.com/hashicorp/nomad/issues/19833 to track the remaining work.

Once I get https://github.com/hashicorp/nomad/pull/19832 merged I think we will not be needing this change, but I will keep your PR open until then just in case.

Thanks again and apologies for all the confusion.

lgfa29:
https://github.com/hashicorp/nomad/pull/19832 was merged, so we're good to use `pointer.Of` in MPL code. I've opened https://github.com/hashicorp/terraform-provider-nomad/pull/423 to update the TF provider to only use MPL code.

So I think we're good without this change, and I will go ahead and close the PR.

Thank you again @kevinschoonover for catching this and bringing it to our attention!

PR #19840, opened 2024-01-29
users: refactor method for getting UID from username

shoenig:
This PR refactors a helper function for getting the UID associated with
a given username to also return the GID and home directory. Also adds
unit tests on the known values of root and nobody user on Ubuntu Linux.


tgross (approved):
LGTM

PR #19856, opened 2024-02-01
template: run template tests on Windows where possible 

tgross:
We don't run the whole suite of unit tests on all platforms to keep CI times
reasonable, so the only things we've been running on Windows are
platform-specific.

I'm working on some platform-specific `template` related work and having these
tests run on Windows will reduce the risk of regressions. Our Windows CI box
doesn't have Consul or Vault, so I've skipped those tests for the time being,
and can follow up with that later.


client/allocrunner/taskrunner/template/template_default_test.go
```diff
@@ -0,0 +1,69 @@
...
+		t.Fatalf("Failed to stat file: %v", err)
+	}
+
+	if m := fi.Mode(); m != os.ModePerm {
+		t.Fatalf("Got mode %v; want %v", m, os.ModePerm)
+	}
+
+	sys := fi.Sys()
+	uid := pointer.Of(int(sys.(*syscall.Stat_t).Uid))
+	gid := pointer.Of(int(sys.(*syscall.Stat_t).Gid))
```
shoenig, line 65: guess i've never done this before but wow that's gross

shoenig (approved):
LGTM!

PR #19878, opened 2024-02-05
[ui] Random exec assignment chooses from only those with a matching taskGroup, if provided

### LINKED ISSUE #10962: [ui] Exec UI selects wrong allocation
### Nomad version
1.1.2

`Linux ip-<redacted> 4.9.0-16-amd64 #1 SMP Debian 4.9.272-1 (2021-06-21) x86_64 GNU/Linux`

### Issue
The exec UI is misleading about how it'll choose an allocation -- it only uses task names to filter allocations, and not group names.

### Reproduction steps
Start a job with different group names, but task names the same. 
Navigate to the UI. Select a job under the console group.

(some bad redacting done)
Screenshot 2021-07-28 at 17 25 52

#### Expected Result
The allocation you exec into belongs to the expected  group. (in the example that would be the console group)

#### Actual Result
The allocation belongs to a random group.

philrenaud:
Resolves #10962 

Previously, when selecting Exec from a group page in the UI (or selecting a task group/task from the exec window's left sidebar), we go over a given job's allocations until we find one that is active and whose name matches the selected one.

This becomes a problem in situations where a user has multiple (differently-named) groups, whose tasks share a name between them.

This PR filters down to the group name in question.

### DIFF
--- a/.changelog/19878.txt
+++ b/.changelog/19878.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fix an issue where a same-named task from a different group could be selected when the user clicks Exec from a task group page where multiple allocations would be valid
+```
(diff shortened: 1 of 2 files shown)


lgfa29 (approved):
LGTM! Probably worth backporting how far as possible.

Issue #15144, opened 2022-11-04
Nomad should differentiate between crashed/rebooted and disconnected nomad client for jobs with `max_client_disconnect`


yaroslav-007:
Nomad should differentiate between crashed/rebooted and disconnected nomad client for jobs with `max_client_disconnect` for the relocation of allocation

### Nomad version
`nomad version` : `1.4.2`

### Issue
Nomad reschedule allocations  jobs with `max_client_disconnect` on crashed system where the allocation was stopped

### Reproduction steps
Vagrant project attached. Follow the README file in the attachment

### Expected Result
Nomad should differentiate between crashed and disconnected nomad client for jobs with `max_client_disconnect`.  It should be able to see that there is no running allocation on the reconnected client and thus not take any further action

### Actual Result
Nomad will reschedule a job (with `max_client_disconnect`) on the reconnect of crashed system. 

### Attachment
[Attachment](https://github.com/hashicorp/nomad/files/9936809/Archive.zip)

tgross:
@yaroslav-007 can you clarify the expected behavior here? Nomad servers _can't_ know whether a Nomad client is netsplit vs crashed, because either way the client is no longer communicating with the server. And it doesn't actually matter in terms of the intended behavior of `max_client_disconnect`, which is to temporarily replace the allocations until the client comes back.

yaroslav-007:
Hello @tgross ,
Nomad should be able to detect that on the recovered client the allocations/docker containers are not running. In  that case the temporary one (created because of the disconnected client ) could continue to run without any need of them to be redeployed on the the recovered node. This seems as unnecessary action in that case.

tgross:
Ah yes now I understand! I think @lgfa29 fixed that behavior in https://github.com/hashicorp/nomad/pull/15068, which we just shipped today in Nomad 1.4.3

lgfa29:
More specifically, this could have been caused by https://github.com/hashicorp/nomad/issues/12680 where a new allocation was being placed unnecessarily. Hopefully this was fixed in 1.4.3 😄

yaroslav-007:
Hello,

I have checked with the latest version (1.4.3) and seems nomad continue to reschedule the jobs.

tgross:
Awesome, thanks for following up!

mikenomitch:
I asked @yaroslav-007 to retest this since we've had a lot of work on this recently. Copy/pasting his Slack reply here:

[...]

PR #19886, opened 2024-02-06
Introduces `Disconnect` block into the `TaskGroup` configuration

Juanadelacuesta:
This PR is the first on two that will implement the new `Disconnect` block. In this PR the new block is introduced to be backwards compatible with the fields it will replace. For more information refer  to [this](https://go.hashi.co/rfc/nmd-188) RFC and [this](https://github.com/hashicorp/nomad/issues/15144) ticket.

It was intended to facilitate the future deprecation of `max_client_disconnect`, `stop_after_client_disconnect` and `prevent_reschedule_on_lost` hence the duplication of tests.


scheduler/reconcile_util_test.go
```diff
@@ -1329,6 +1331,1327 @@ func TestAllocSet_filterByTainted(t *testing.T) {
 	}
 }
 
+func TestAllocSet_filterByTainted_Disconnected(t *testing.T) {
```
tgross, line 1334: It looks like this is a line-by-line recreation of the now-deprecated `TestAllocSet_filterByTainted` above.

We should consider what the maintenance burden of keeping two identical tests like this is. Even once we've marked these fields as deprecated; it'll take some time for them to get fully cycled out (we'll need to keep them in the structs for upgrade safety for several major versions).

I recommend that we create a `testAllocSet_filterByTainted` test function that accepts the 4 jobs under test as parameters, and then call it from two different tests: `TestAllocSet_filterByTainted_Deprecated` and `TestAllocSet_filterByTainted_Disconnected`.

This same comment applies to what we've done in:
* `scheduler/generic_sched_test.go` with `TestServiceSched_Client_Creates_Updates_and_Evals_Disconnect`
* `nomad/node_endpoint_test.go` with `TestClientEndpoint_UpdateStatus_Disconnect_Reconnect`

tgross (changes requested):
Overall this looks good @Juanadelacuesta. I've left a bunch of comments here but probably the most important are fixing the bugs in `Canonicalize` and the recommendations around having two copies of the tests.

client/heartbeatstop.go
```diff
@@ -43,11 +43,26 @@ func newHeartbeatStop(
 	return h
 }
 
+// This helper is meant to simplify the future depracation of
+// StopAfterClientDisconnect in favor of Disconnect.StopAfterOnClient
+// introduced in 1.8.0.
+func getDisconnectStopTimeout(tg *structs.TaskGroup) *time.Duration {
```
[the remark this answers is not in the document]
Juanadelacuesta: Done!

api/tasks.go
```diff
@@ -113,6 +123,46 @@ func (r *RestartPolicy) Merge(rp *RestartPolicy) {
...
+
+	// A boolean field used to define if the allocations should be replaced while
+	// its  considered disconnected.
+	Replace *bool `mapstructure:"replace" hcl:"replace,optional"`
+
+	// Once the disconnected node starts reporting again, it will define which
+	// instances to keep: the original allocations, the replacement, the one
+	// running on the node with the best score as it is currently implemented,
+	// or the allocation that has been running continuously the longest.
+	Reconcile *ReconcileOption `mapstructure:"reconcile" hcl:"reconcile,optional"`
```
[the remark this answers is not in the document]
Juanadelacuesta: I was more or less following what I saw on the other objects: Everything is a pointer on the API, but for the structs, only the ones that need to be a pointer, to make the diasctintion between disabled and default value, are pointers.

nomad/structs/structs.go
```diff
@@ -6250,6 +6265,95 @@ type ScalingPolicyListStub struct {
...
+func NewDisconnectStrategy() *DisconnectStrategy {
+	return &DisconnectStrategy{
+		Replace:   pointer.Of(true),
+		Reconcile: ReconcileOptionBestScore,
+	}
+}
+
+// Disconnect strategy defines how both clients and server should behave in case of
+// disconnection between them.
+type DisconnectStrategy struct {
```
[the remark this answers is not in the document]
Juanadelacuesta: Done!

nomad/structs/structs.go
```diff
@@ -6250,6 +6265,95 @@ type ScalingPolicyListStub struct {
...
+	// Defines for how long the server will consider the unresponsive node as
+	// disconnected but alive instead of lost.
+	LostAfter time.Duration `mapstructure:"lost_after" hcl:"lost_after,optional"`
+
+	// Defines for how long a disconnected client will keep its allocations running.
+	StopAfterOnClient *time.Duration `mapstructure:"stop_after" hcl:"stop_after_client"`
```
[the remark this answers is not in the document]
Juanadelacuesta: - The mapstructure and hcl tags should match, shouldn't they? :O yes, they should, fixing it!
- Yes!
- Done!

nomad/structs/structs.go
```diff
@@ -6250,6 +6265,95 @@ type ScalingPolicyListStub struct {
...
+
+	// A boolean field used to define if the allocations should be replaced while
+	// its  considered disconnected.
```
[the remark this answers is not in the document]
Juanadelacuesta: Done!

nomad/structs/structs.go
```diff
@@ -6250,6 +6265,95 @@ type ScalingPolicyListStub struct {
...
+		ds.StopAfterOnClient = cds.StopAfterOnClient
+	}
+
+	if ds.Replace == nil {
+		ds.Replace = cds.Replace
+	}
+
+	if ds.Reconcile != ReconcileOptionBestScore {
+		ds.Reconcile = cds.Reconcile
+	}
```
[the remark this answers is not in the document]
Juanadelacuesta: You are right for this and the previous ones, I did it like that because this function is currently only called if the Reconcile block is nil, but Im inadvertently introducing a dependency here, let me fix it :)

nomad/structs/structs.go
```diff
@@ -6250,6 +6265,95 @@ type ScalingPolicyListStub struct {
...
+
+func (ds *DisconnectStrategy) Canonicalize() {
+	cds := NewDisconnectStrategy()
+	if ds.LostAfter == 0 {
+		ds.LostAfter = cds.LostAfter
+	}
+
+	if ds.StopAfterOnClient != nil {
+		ds.StopAfterOnClient = cds.StopAfterOnClient
+	}
```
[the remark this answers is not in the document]
Juanadelacuesta: Initially I wan't declaring it as a pointer, I forgot to remove this

scheduler/reconcile_util_test.go
Juanadelacuesta: I was thinking about the simplest way to remove the tests in the future, but here is take  2

nomad/structs/group.go
```diff
@@ -0,0 +1,122 @@
...
+		return nil
+	}
+
+	nds := new(DisconnectStrategy)
+	*nds = *ds
+	return nds
+}
+
+func (ds *DisconnectStrategy) Canonicalize() {
+	cds := NewDefaultDisconnectStrategy()
```
tgross, line 110: My previous comment on this got lost, but we shouldn't allocate an entirely new object just to (maybe) grab default fields from it.

```suggestion
```

nomad/structs/group.go
```diff
@@ -0,0 +1,122 @@
...
+	if ds.LostAfter == 0 {
+		ds.LostAfter = cds.LostAfter
+	}
+
```
tgross, line 114: The default value for `LostAfter` is `time.Duration(0)`, so there's no need to change `ds.LostAfter` if it's already at zero.

```suggestion
```

nomad/structs/group.go
```diff
@@ -0,0 +1,122 @@
...
+	if ds.Replace == nil {
+		ds.Replace = cds.Replace
+	}
+
+	if ds.Reconcile == "" {
+		ds.Reconcile = cds.Reconcile
```
tgross, line 120: If we don't allocate the extra object with `NewDisconnectStrategy`, we can just do this, which allocates no additional memory because the string's internal pointer will point to the statically-allocated value:

```suggestion
		ds.Reconcile = ReconcileOptionBestScore
```

nomad/structs/group.go
```diff
@@ -0,0 +1,122 @@
...
+		return multierror.Append(mErr, errStopAndLost)
+	}
+
+	if ds.Reconcile != "" &&
+		ds.Reconcile != ReconcileOptionBestScore &&
+		ds.Reconcile != ReconcileOptionLongestRunning &&
+		ds.Reconcile != ReconcileOptionKeepOriginal &&
+		ds.Reconcile != ReconcileOptionKeepReplacement {
+		return multierror.Append(mErr, fmt.Errorf("%w: %s", errInvalidReconcile, ds.Reconcile))
+	}
```
tgross, line 94: Another way of doing this is that I kinda like because it's almost got the readability of pattern-matching is:

```suggestion
	switch ds.Reconcile {
		case "", ReconcileOptionBestScore, ReconcileOptionLongestRunning, ReconcileOptionKeepOriginal, ReconcileOptionKeepReplacement:
		default:
			return multierror.Append(mErr, fmt.Errorf("%w: %s", errInvalidReconcile, ds.Reconcile))
		}
	}
```

But up to you.

nomad/structs/group_test.go
```diff
@@ -0,0 +1,240 @@
...
+	// Set up a job with an invalid Disconnect.LostAfter value
+	job := testJob()
+	timeout := -1 * time.Minute
+	job.TaskGroups[0].Disconnect = &DisconnectStrategy{
+		LostAfter:         timeout,
+		StopAfterOnClient: &timeout,
+	}
+
+	err := job.Validate()
+	require.Error(t, err)
```
tgross, line 27:
```suggestion
	must.Error(t, err)
```

(Same for all the other tests in this PR.)

api/tasks.go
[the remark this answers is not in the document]
tgross: It should be that the `api` structs only have the pointer when the zero value is meaningful but non-default. But I think we haven't been as consistent about this as we should be. That being said, I don't think it hurts to have these be pointers here either, it just complicates the conversion between `structs.DisconnectStrategy` and `api.DisconnectStrategy`.

scheduler/reconcile_util_test.go
tgross: That works! :+1:

tgross (changes requested):
A few comments got missed while addressing the last round. I've resolved those and re-made the comments in the new locations to make sure they're not dropped.

nomad/structs/structs.go
```diff
@@ -10970,12 +11052,14 @@ func (a *Allocation) NextRescheduleTimeByTime(t time.Time) (time.Time, bool) {
...
+// ShouldClientStop tests an alloc for StopAfterClient on the Disconnect configuration
 func (a *Allocation) ShouldClientStop() bool {
 	tg := a.Job.LookupTaskGroup(a.TaskGroup)
+	timeout := tg.GetDisconnectStopTimeout()
+
 	if tg == nil ||
-		tg.StopAfterClientDisconnect == nil ||
-		*tg.StopAfterClientDisconnect == 0*time.Nanosecond {
+		timeout == nil ||
+		(timeout != nil && *timeout == 0*time.Nanosecond) {
```
lgfa29, line 11062:
```suggestion
		*timeout == 0*time.Nanosecond {
```

I think the previous `timeout == nil` clause already covers `timeout != nil`.

nomad/structs/structs.go
```diff
@@ -11274,11 +11360,12 @@ func (a *Allocation) Expired(now time.Time) bool {
 		return false
 	}
 
-	if tg.MaxClientDisconnect == nil && !tg.PreventRescheduleOnLost {
+	timeout := tg.GetDisconnectLostTimeout()
+	if timeout == 0 && !tg.PreventRescheduleOnLost {
```
lgfa29, line 11364: Do we need to check `tg.Disconnect.Replace` as well? Maybe a helper method would help like for the other fields.

nomad/structs/structs.go
```diff
@@ -7340,6 +7392,36 @@ func (tg *TaskGroup) GoString() string {
 	return fmt.Sprintf("*%#v", *tg)
 }
 
+// GetDisconnectLostTimeout is a helper meant to simplify the future depracation of
+// MaxClientDisconnect in favor of Disconnect.LostAfter
+// introduced in 1.8.0.
+func (tg *TaskGroup) GetDisconnectLostTimeout() time.Duration {
+	if tg.MaxClientDisconnect != nil {
+		return *tg.MaxClientDisconnect
+	}
```
lgfa29, line 7401: Maybe I'm confusing it with another field because I can't find any docs about it, but is `0` a valid (and special) value for `max_client_disconnect`?

I have a vague memory of a bug around this area that was caused by mishandling a `0` value 😅

nomad/structs/structs.go
```diff
@@ -11041,20 +11125,22 @@ func (a *Allocation) SupportsDisconnectedClients(serverSupportsDisconnectedClien
...
 	}
 
 	return false
 }
 
 // PreventRescheduleOnLost determines if an alloc allows to have a replacement
-// when lost.
-func (a *Allocation) PreventRescheduleOnLost() bool {
+// when Disconnected.
+func (a *Allocation) PreventRescheduleOnDisconnect() bool {
```
lgfa29, line 11137: Strictly speaking `lost` is a valid client status for an alloc, while `disconnect` is not, so this method is checking if an alloc should not reschedule if it were to become `lost`. 

But also the alloc would be `unknown`, not `lost`, so the new name is probably fine 😅

scheduler/generic_sched_test.go
```diff
@@ -3644,24 +3644,94 @@ func TestServiceSched_StopAfterClientDisconnect(t *testing.T) {
 	ci.Parallel(t)
 
 	cases := []struct {
-		stop        time.Duration
+		jobSpec     func() *structs.Job
```
lgfa29, line 3647:
```suggestion
		jobSpecFn     func(*structs.Job)
```

A small change if you would like to avoid having to call `mock.Job()` inside each test case. The main test loop could call it once and feed it to each case for modifications.

api/tasks.go
```diff
@@ -113,6 +123,46 @@ func (r *RestartPolicy) Merge(rp *RestartPolicy) {
...
 
+// Disconnect strategy defines how both clients and server should behave in case of
+// disconnection between them.
+type DisconnectStrategy struct {
+	// Defines for how long the server will consider the unresponsive node as
+	// disconnected but alive instead of lost.
+	LostAfter *time.Duration `mapstructure:"lost_after" hcl:"lost_after,optional"`
+
+	// Defines for how long a disconnected client will keep its allocations running.
+	StopAfterOnClient *time.Duration `mapstructure:"stop_after_on_client" hcl:"stop_after_on_client,optional"`
```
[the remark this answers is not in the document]
Juanadelacuesta: That sounds a lot better indeed

nomad/structs/group.go
Juanadelacuesta: I particularly hate the switches with one case, personal preference jejej

nomad/structs/structs.go
Juanadelacuesta: We do! my bad

nomad/structs/structs.go
Juanadelacuesta: I think the one you are thinking about is stop_after_client_disconnect

nomad/structs/group.go
```diff
@@ -0,0 +1,122 @@
...
+	var mErr *multierror.Error
+
+	if ds.StopAfterOnClient != nil {
+		if *ds.StopAfterOnClient < 0 {
+			mErr = multierror.Append(mErr, errNegativeStopAfter)
+		}
+
+		if job.Type != JobTypeService && job.Type != JobTypeBatch {
+			mErr = multierror.Append(mErr, errStopAfterNonService)
+		}
```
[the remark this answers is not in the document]
Juanadelacuesta: Yes, this is an existing constraint

nomad/structs/group.go
```diff
@@ -0,0 +1,122 @@
...
+	// Once the disconnected node starts reporting again, it will define which
+	// instances to keep: the original allocations, the replacement, the one
+	// running on the node with the best score as it is currently implemented,
+	// or the allocation that has been running continuously the longest.
+	Reconcile string `mapstructure:"reconcile" hcl:"reconcile,optional"`
+}
+
+func (ds *DisconnectStrategy) Validate(job *Job) error {
+	var mErr *multierror.Error
+
```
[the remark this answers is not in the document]
Juanadelacuesta: I hate the need to do that, this means there is something else missing somewhere else

command/testdata/example-short-bad.json
```diff
@@ -88,8 +88,10 @@
                 "Meta": null,
                 "Services": null,
                 "ShutdownDelay": null,
-                "StopAfterClientDisconnect": null,
-                "MaxClientDisconnect": null,
```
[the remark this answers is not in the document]
Juanadelacuesta: oki

nomad/structs/group_test.go
```diff
@@ -0,0 +1,240 @@
...
+	"testing"
+	"time"
+
+	"github.com/hashicorp/nomad/ci"
+	"github.com/hashicorp/nomad/helper/pointer"
+	"github.com/shoenig/test/must"
+	"github.com/stretchr/testify/require"
+)
+
+func TestJobConfig_Validate_LostAfter_Disconnect(t *testing.T) {
```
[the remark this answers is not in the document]
Juanadelacuesta: This was an existing test, I guess a job validate integration one. I just relocated it

tgross (changes requested):
Two last stray areas:
* the `api` version of `Canonicalize` hasn't been fixed to match what `structs` is doing
* some of the new tests are still using `require` instead of `must`

nomad/structs/diff_test.go
```diff
@@ -2970,6 +2970,161 @@ func TestTaskGroupDiff(t *testing.T) {
...
+					LostAfter:         time.Second,
+					Replace:           pointer.Of(false),
+					Reconcile:         ReconcileOptionLongestRunning,
+					StopOnClientAfter: pointer.Of(1 * time.Second),
+				},
+			},
+			New: &TaskGroup{
+				Disconnect: &DisconnectStrategy{
+					LostAfter:         time.Minute,
+					Replace:           pointer.Of(true),
```
tgross, line 3099: We changed `Replace` from `false` to `true`, but it doesn't show up in the diff below. That says to me that using `primitiveObjectDiff` as the diff function might not be suitable for the two pointer fields.

Issue #19836, opened 2024-01-28
TLS CA create command with name-constraint


guillaumejchauveau:
### Nomad version
Bug occurs on versions 1.6.0, 1.7.[1,2,3].

### Operating system and Environment details
Fedora 39, Rocky Linux 9.

### Issue
Using the CLI command `nomad tls ca create -name-constraint="true" -domain="foo.com"` yields `Please provide the -domain flag if you want to enable custom domain constraints`.
Nomad version 1.5.0 behaves as expected.

### Reproduction steps
Use Nomad version 1.6.0 or 1.7.[1,2,3]. Run the command `nomad tls ca create -name-constraint="true" -domain="foo.com"`.

#### Expected Result
```
==> CA certificate saved to: foo.com-agent-ca.pem
==> CA certificate key saved to: foo.com-agent-ca-key.pem
```

#### Actual Result
```
Please provide the -domain flag if you want to enable custom domain constraints.
```

lgfa29:
Thanks for the report @guillaumejchauveau.

I think the bug was introduced in https://github.com/hashicorp/nomad/pull/17309, where the `-domain` flag is being overwritten unless you specify other customization flags. I've opened https://github.com/hashicorp/nomad/pull/19892 to fix this.

PR #19892, opened 2024-02-06
cli: fix `tls ca create` command with `-domain`

lgfa29:
The current implementation of the `nomad tls ca create` command ovierrides the value of the `-domain` flag with `"nomad"` if no additional customization is provided.

This results in a certificate for the wrong domain or an error if the `-name-constraint` flag is also used.

The logic for `IsCustom()` also seemed reversed. If all custom fields are empty then the certificate is _not_ customized, so `IsCustom()` should return false.

Closes https://github.com/hashicorp/nomad/issues/19836

### DIFF
--- a/.changelog/19892.txt
+++ b/.changelog/19892.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cli: Fixed a bug where the `nomad tls ca create` command failed when the `-domain` was used without other values
+```
--- a/command/tls_ca_create.go
+++ b/command/tls_ca_create.go
@@ -147,9 +147,9 @@ func (c *TLSCACreateCommand) Run(args []string) int {
 	flagSet := c.Meta.FlagSet(c.Name(), FlagSetClient)
 	flagSet.Usage = func() { c.Ui.Output(c.Help()) }
 	flagSet.Var(&c.additionalDomain, "additional-domain", "")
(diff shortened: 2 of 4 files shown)


schmichael (approved):

schmichael (approved):

jonashaag:
Does this also fix the issue that when you overwrite the organization details, the validity of the CA is set to 0 days?

lgfa29:
Hi @jonashaag 👋 

I'm not sure, do you happen to have an issue that describes the problem?

You can also test these changes from our latest CI build (it's all the way down the page 😅):
https://github.com/hashicorp/nomad/actions/runs/7878378042

Just keep in mind that these are development builds and should be used for any production code.

jonashaag:
Thanks, I just gave it a try. It seems to be fixed in that build. With v1.7.4:

```
$ nomad tls ca create -name-constraint -domain ... -additional-domain ... -country DE -common-name ... -organization ... -organizational-unit ...

$ openssl x509 -noout -text -in *ca.pem | grep Not
            Not Before: Feb 13 08:33:23 2024 GMT
            Not After : Feb 13 08:33:23 2024 GMT
```

lgfa29:
Nice, thanks for testing it. So we got a two-for-one deal on this PR 😄 

The fix is available on Nomad 1.7.5 and 1.6.8 that were just released.

PR #19902, opened 2024-02-07
docs: add note about stub list filtering

### LINKED ISSUE #19871: HTTP API /v1/nodes endpoint fails filtering by the Address selector
### Nomad version
Output from `nomad version`

```
Nomad v1.7.3
BuildDate 2024-01-15T16:55:40Z
Revision 60ee328f97d19d2d2d9761251b895b06d82eb1a1
```

Ubuntu minimal 22.04 cloud image amd64

Using nomad-pack and dante services. No Consul integration.

### Issue

Using the HTTP API, trying to filter a node by the Address selector returns an error

### Reproduction steps

```
curl -s --get $NOMAD_ADDR/v1/nodes --data-urlencode 'filter=Address == "10.1.7.131"'
```
or

```
curl -s --get $NOMAD_ADDR/v1/nodes --data-urlencode 'filter=Address contains "10.1.7.131"'
```

or

```
curl -s --get $NOMAD_ADDR/v1/nodes --data-urlencode 'filter="10.1.7.131" in Address'
```

#### Expected Result
```
[
  {
    "Address": "10.1.7.131",
    "ID": "00f6c407-26df-e18f-52e1-abc",
    "Attributes": null,
    [...CONT...]
```

#### Actual Result

```
failed to read result page: error finding value in datum: /Address at part 0: couldn't find key: struct field with name "Address"
```

#### Additional information

The same Address selector filter, but for the v1/service/:service endpoint, works correctly

In the same endpoint as the issue, filtering by the ID selector does not cause this error.

lgfa29:
When filtering list results, the filter expression is applied to the full object, not the stub. This is useful because it allows users to filter the list on fields not present in the object stub. But it can also be confusing because some fields have different names, or only exist in the stub, so the filter expression needs to reference fields not present in returned data.

Filtering on the stub would reduce the confusion, but it would also restrict users to only be able to filter on the fields in the stub, which, by definition, are just a subset of the original fields.

Documenting this behaviour can help users understand unexpected errors and results.

Closes #19871

### DIFF
--- a/website/content/api-docs/index.mdx
+++ b/website/content/api-docs/index.mdx
@@ -123,7 +123,16 @@ parameter with the URL encoded expression when sending requests to HTTP API
 endpoints that support it.
 
 ```shell-session
-$ curl --get https://localhost:4646/v1/<path> --data-urlencode 'filter=<filter expression>'
+$ curl --get https://localhost:4646/v1/<path> \
+    --data-urlencode 'filter=<filter expression>'
+```
+
+The filter expression can also be specified in the
+[`-filter`][cli_operator_api_filter] flag of the
+[`nomad operator api`][cli_operator_api] command.
+
+```shell-session
+$ nomad operator api -filter '<filter expression>' /v1/<path>
 ```
 
 Some endpoints may have other query parameters that are used for filtering, but
@@ -145,7 +154,7 @@ node addresses should use the `HTTPAddr` field of the full node definition
 instead of `Address` field present in the stub.
 
 ```shell-session
-$ curl --get https://localhost:4646/v1/nodes --data-urlencode 'filter=HTTPAddr matches "10.0.0..+"'
+$ nomad operator api -filter 'HTTPAddr matches "10.0.0..+"' /v1/nodes
 ```
 
 ### Creating Expressions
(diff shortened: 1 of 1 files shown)


schmichael (approved):
Fantastic feature (even if accidental!) and fantastic doc update. Thanks Luiz!

website/content/api-docs/index.mdx
```diff
@@ -132,6 +132,22 @@ they can't be used with the `filter` query parameter. Doing so will result in a
...
+full resource definition. To allow more expressive filtering operations, the
+filter is applied to the full version, not the stub.
+
+If a request returns an error such as `error finding value in datum` the field
+used in filter expression may need to be adjusted. For example, filtering on
+node addresses should use the `HTTPAddr` field of the full node definition
+instead of `Address` field present in the stub.
+
+```shell-session
+$ curl --get https://localhost:4646/v1/nodes --data-urlencode 'filter=HTTPAddr matches "10.0.0..+"'
```
schmichael, line 148: Are we doing `nomad operator api` in docs? It makes filtering a little easier albeit subject to our annoying "flags must come first" CLI constraint:

```
nomad operator api -filter 'filter=HTTPAddr matches "10.0.0..+"' /v1/nodes

lgfa29: Oh, that's right! I think the command didn't exist yet when I first wrote this. Examples updated in https://github.com/hashicorp/nomad/pull/19902/commits/9f55a17ea2d1b790914014ee9308da182e64cfd0.

Issue #19834, opened 2024-01-27
bug: missing complete AllocationUpdated events from Nomad event stream


Kamilcuk:
### Nomad version
Nomad 1.7.3

### Operating system and Environment details
Archlinux Newest

### Issue
AllocationUpdated with ClientStatus=complete are no longer generated on Nomad 1.7.3.

The event is generated in Nomad 1.7.2, it is a change.

### Reproduction steps

Download Nomad and in one terminal run `nomad agent -dev`.

In a second terminal, observe the Nomad event stream Allocation topic with:

```
stdbuf -oL nomad operator api /v1/event/stream | grep --line-buffered -xFv '{}' | stdbuf -oL jq -c '.Events.[] | select(.Topic == "Allocation") | [.Topic, .Type, .Payload.Allocation.ID, .Payload.Allocation.ClientStatus]'
```

In a third terminal execute a job and then purge it.

```
nomad job run -detach ./missingcomplete.nomad.hcl && sleep 2 && nomad job stop -purge missingcomplete && nomad job status missingcomplete
```

The job can be anything.

```
# missingcomplete.nomad.hcl
job "missingcomplete" {
  group "g1" {
    reschedule { attempts = 0 }
    restart { attempts = 0 }
    task "t1" {
      driver = "docker"
      config {
        image = "busybox:stable"
        args = ["sleep", "infinity"]
      }
    }
  }
}
```

#### Expected Result
The expected result on the second terminal can be obtained by running nomad version 1.7.2 and looks like the following:

```
$ stdbuf -oL nomad operator api /v1/event/stream | grep --line-buffered -xFv '{}' | stdbuf -oL jq -c '.Events.[] | select(.Topic == "Allocation") | [.Topic, .Type, .Payload.Allocation.ID, .Payload.Allocation.ClientStatus]'
["Allocation","PlanResult","1ef87a91-f640-bb32-7a37-d0d82c478c1a","pending"]
["Allocation","AllocationUpdated","1ef87a91-f640-bb32-7a37-d0d82c478c1a","running"]
["Allocation","PlanResult","1ef87a91-f640-bb32-7a37-d0d82c478c1a","running"]
["Allocation","AllocationUpdated","1ef87a91-f640-bb32-7a37-d0d82c478c1a","running"]
["Allocation","AllocationUpdated","1ef87a91-f640-bb32-7a37-d0d82c478c1a","complete"]
```

[...]

PR #19903, opened 2024-02-07
events: emit event when job is deleted

lgfa29:
When jobs are deregistered with the `purge` flag they are immediately deleted from the state store instead of just updated to be marked as stopped.

Without tracking job deletions the event stream would not receive a `JobDeregistered` event when `purge` was set.

Ref. #19834

### DIFF
--- a/nomad/state/events_test.go
+++ b/nomad/state/events_test.go
@@ -726,7 +726,7 @@ func TestEventsFromChanges_WithDeletion(t *testing.T) {
 	event := eventsFromChanges(nil, changes)
 	require.NotNil(t, event)
 
-	require.Len(t, event.Events, 1)
+	require.Len(t, event.Events, 2)
 }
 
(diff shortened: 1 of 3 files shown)


nomad/state/events_test.go
```diff
@@ -726,7 +726,7 @@ func TestEventsFromChanges_WithDeletion(t *testing.T) {
 	event := eventsFromChanges(nil, changes)
 	require.NotNil(t, event)
 
-	require.Len(t, event.Events, 1)
+	require.Len(t, event.Events, 2)
```
lgfa29, line 729: This was testing something else (https://github.com/hashicorp/nomad/commit/5e34efadab66de20d1a64691916f10df2e8255f8), but incidentally assumed that a job deletion would not generate an event.

pkazmierczak (approved):
LGTM!

tgross (approved):
LGTM!

PR #19904, opened 2024-02-07
[ui] Categorizing and clustering task events in the Job Status panel

### LINKED ISSUE #19853: [ui] Relativize/categorize the time of events in the job status box
image

We can and should do better than these timestamps.

As an idea, something like:

```
Just now:
(event)
(event)
(event)

About 10 minutes ago:
(event)
(event)

Yesterday:
(event)
(event)
(event)
```

philrenaud:
image


philrenaud:
An attempt, like https://github.com/hashicorp/nomad/pull/25156, to add some implicit detail to deployment streams for a job. Not worth the extra logic IMO.

Issue #19870, opened 2024-02-05
Warn about unprotected /v1/metrics endpoint in docs, when `tls.verify_https_client=false`


jwoudenberg:
Thanks for building and maintaining this software, it's a joy to work with!

### Proposal
We've set up a Nomad cluster, following the security guidelines in the documentation. One thing that caught us off guard was that the `/v1/metrics` endpoint is not protected by default, and shares information about the internals of the cluster that should probably be not exposed publicly. I think adding a note to the security documentation might help others avoid this. Specifically, this part of the docs:

> Using [tls.verify_https_client=false](https://developer.hashicorp.com/nomad/docs/configuration/tls#verify_https_client) downgrades the Nomad agent HTTP server to use TLS instead of mTLS. This is safe if ACLs are enabled and is recommended if you are using the Nomad web UI to avoid the difficulty of distributing client certificates to browsers.
[source](https://developer.hashicorp.com/nomad/docs/concepts/security#requirements)

We went this path, but didn't realize a side-effect of not enabling mTLS for clients is that it leaves the /v1/metrics endpoint unprotected. We imagine this route is only safe when adding some proxy in front of servers blocking access to this endpoint.

### Use-cases
Create awareness of a possible vulnerability in a setup with `tls.verify_https_client=false`.

### Attempted Solutions
We ended up adding the following to our Nomad configuration as a patch:
```
telemetry {
  filter_default = false
}
```

This disables all telemetry though. As a longer-term solution that allows access to telemetry we're adding a layer-7 firewall (AWS ALB in our case) in front of Nomad servers to block access to this specific endpoint. That blocks public access to the metrics endpoint from outside our network, but still allows us to scrape the endpoint internally.

lgfa29:
Hi @jwoudenberg 👋 

Thank you for the suggestion and the kind words about Nomad. I'm glad you are enjoying it 😄 

That sounds like something well worth documenting. I will see if I have some time in the coming days to tackle this.

Thanks again, and feel free to reach out if you need any other help!

PR #19916, opened 2024-02-08
docs: expand impact of `verify_https_client=false`

lgfa29:
When Nomad is configured with `verify_https_client=false` endpoints that do not require an ACL token can be accessed without any other type of authentication. Expand the docs to mention this effect.

Closes https://github.com/hashicorp/nomad/issues/19870


jrasell (approved):
LGTM!

PR #19922, opened 2024-02-08
Doc-only: Added note that Autoscaler Enterprise does not support reloading

assareh:
of the nomad block

License checker does not support SIGHUP reloading and therefore this statement is not currently accurate for autoscaler enterprise.

### DIFF
--- a/website/content/tools/autoscaling/agent.mdx
+++ b/website/content/tools/autoscaling/agent.mdx
@@ -78,6 +78,8 @@ following actions.
 - reconfigure the remaining plugins with the configuration defined in their
   `config` parameter.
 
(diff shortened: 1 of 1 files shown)


lgfa29:
Hi @assareh 👋 

Thank you for the PR. Could you expand a little more on what config were your expecting to reload with a SIGHUP? The Autoscaler doesn't have a license of its own, so it checks for Nomad's license.

lgfa29:
@assareh explained the problem to me. https://github.com/hashicorp/nomad-autoscaler/pull/849 will fix the issue so the Autoscaler agent reload behaves properly.

Issue #19756, opened 2024-01-17
consul.partition fingerprint generates warning logs on CE


tgross:
The Consul fingerprint for Nomad CE generates a warning log after 1.7.3. I added support for fingerprinting the Consul admin partition in https://github.com/hashicorp/nomad/pull/19485 and tested it with combinations of Nomad CE/ENT and Consul CE/ENT. Unfortunately I apparently missed that if you fingerprint Consul CE from this code, you get a valid fingerprint and working Consul, but noise in your logs.

The "extractor" for the partition returns `"", false`, which the implementation treats as a warning. If we returned `"", true`, we wouldn't get the warning but we'd get a useless fingerprint value of `consul.partition = ""` which we don't want either. So we need to adjust the implementation to handle empty values differently too.

Reported in https://github.com/hashicorp/nomad/pull/19485#issuecomment-1895853867

maxless:
Is there a way to disable this message through config currently if you're not using Consul?

tgross:
Unfortunately no. When you leave out the `consul` config, Nomad generates a default one. This is something we've discussed removing entirely to make it nicer to run Nomad without Consul (as we do for the `vault.enabled` flag), but it hasn't managed to get to the top of the priority pile. So we've got to fix the fingerprint... I'll try to get this into the next regular release.

maxless:
Okay, thanks.

tgross:
Will be fixed in https://github.com/hashicorp/nomad/pull/19923, planned for shipping in Nomad 1.7.5

PR #19923, opened 2024-02-08
fingerprint: eliminate spurious warning logs with Consul CE

tgross:
Support for fingerprinting the Consul admin partition was added in #19485. But when the client fingerprints Consul CE, it gets a valid fingerprint and working Consul but with a warn-level log. Return "ok" from the partition extractor, but also ensure that we only add the Consul attribute if it actually has a value.

Fixes: https://github.com/hashicorp/nomad/issues/19756


lgfa29 (approved):
LGTM!

PR #19925, opened 2024-02-08
windows: remove LazyDLL calls for system modules

tgross:
On Windows, Nomad uses `syscall.NewLazyDLL` and `syscall.LoadDLL` functions to load a few system DLL files, which does not prevent DLL hijacking attacks. Hypothetically a local attacker on the client host that can place an abusive library in a specific location could use this to escalate privileges to the Nomad process. Although this attack does not fall within the Nomad security model, it doesn't hurt to follow good practices here.

We can remove two of these DLL loads by using wrapper functions provided by the stdlib in `x/sys/windows`

Fixes: https://github.com/hashicorp/nomad-enterprise/issues/991

### DIFF
--- a/command/agent/host/windows.go
+++ b/command/agent/host/windows.go
@@ -37,30 +37,24 @@ func mountedPaths() (disks []string) {
 }
 
 type df struct {
-	size       *uint64 // "systemFree" less quotas
-	avail      *uint64
-	systemFree *uint64
+	size       uint64 // "systemFree" less quotas
+	avail      uint64
+	systemFree uint64
 }
 
 func makeDf(path string) (*df, error) {
 	df := &df{}
 	err := windows.GetDiskFreeSpaceEx(
 		syscall.StringToUTF16Ptr(path),
-		df.avail, df.size, df.systemFree)
+		&df.avail, &df.size, &df.systemFree)
 
 	return df, err
 }
 
 func (d *df) total() uint64 {
-	if d.size == nil {
-		return 0
-	}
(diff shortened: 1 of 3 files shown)


tgross:
Quick smoke test of the `makeDf` change shows what I'd expect in the fingerprint:

node-status-self

dduzgun-security (approved):

command/agent/host/windows.go
```diff
@@ -36,34 +37,30 @@ func mountedPaths() (disks []string) {
...
-
-	c, err := h.FindProc("GetDiskFreeSpaceExW")
-	if err != nil {
-		return nil, err
-	}
-
 	df := &df{}
+	err := windows.GetDiskFreeSpaceEx(
+		syscall.StringToUTF16Ptr(path),
+		df.avail, df.size, df.systemFree)
```
schmichael, line 49: If instead of making all of the fields on df pointers, could we pass them by refere here?
```suggestion
		&df.avail, &df.size, &df.systemFree)
```

That would allow skipping the nil checks below, but otherwise I think it's functionally the same. :thinking:

schmichael (approved):
not a blocker

command/agent/host/windows.go
tgross: Oh, yeah that's nicer. Will fix.

tgross: Done and double-checked that we're still getting the same results, out of justified paranoia at Windows :grinning:

Issue #19506, opened 2023-12-16
panic serving: concurrent write to websocket connection


Kamilcuk:
### Nomad version

```
$ /usr/bin/nomad --version
Nomad v1.7.2
BuildDate 2023-12-13T19:59:42Z
Revision 64e3dca9274b493e38a49fda3a70fd31d0485b91
```

### Operating system and Environment details

Archlinux.

### Issue

I am executing a lot of `nomad job exec` API commands to test some stuff and Nomad logs some panics:

```
    2023-12-16T22:56:25.020+0100 [DEBUG] http: request complete: method=GET path="/v1/client/allocation/9ececd9c-a005-40e5-b2ef-38b4c2454424/exec?command=%5B%22sh%22%2C%22-c%22%
2C%22%5Cn++++++++++++find+%5C%22%242%5C%22+-maxdepth+1+-mindepth+1+-type+f+-name+%5C%22%243%2A%5C%22%5Cn++++++++++++find+%5C%22%242%5C%22+-maxdepth+1+-mindepth+1+-type+d+-name+%
5C%22%243%2A%5C%22+%7C%5Cn++++++++++++++++while+IFS%3D+read+-r+line%3B+do+echo+%5C%22%24line%2F%5C%22%3B+done%5Cn++++++++++++%22%2C%22--%22%2C%22%2Fusr%2Fbi%22%2C%22%2Fusr%2F%22
%2C%22bi%22%5D&namespace=default&region=global&task=test_nomad_cp_complete&tty=false" duration=9.321342ms                                                                        
    2023-12-16T22:56:25.020+0100 [ERROR] http: http: panic serving 127.0.0.1:50746: concurrent write to websocket connection                                                     
goroutine 3830 [running]:                                                                                                                                                        
net/http.(*conn).serve.func1()                                                                                                                                                   
        net/http/server.go:1868 +0xb9                                                                                                                                            
panic({0x28e3c80?, 0x37baf50?})                                                                                                                                                  
        runtime/panic.go:920 +0x270                                                                                                                                              
github.com/gorilla/websocket.(*messageWriter).flushFrame(0xc001227428, 0x1, {0xc00170e6c0?, 0x7fe1d8062338?, 0x7fe220b35a68?})                                                   
        github.com/gorilla/websocket@v1.5.0/conn.go:617 +0x4b8                                                                                                                   
[...]

PR #19172, opened 2023-11-24
agent: ignore websocket statuses 1000, 1001 and 1005 correctly

pkazmierczak:
These are "close" messages and not actual errors. 

Fixes https://github.com/hashicorp/nomad/issues/19164.

PR #19932, opened 2024-02-09
alloc exec: fix panics after stream close

tgross:
In #19172 we added a check on websocket errors to see if they were one of several benign "close" messages. This change inadvertently assumed that other messages used for close would not implement `HTTPCodedError`. When errors like the following are received:

> msgpack decode error [pos 0]: io: read/write on closed pipe"

they are sent from the inner loop as though they were a "real" error, but the channel is already being closed with a "close" message.

This allowed many more attempts to pass thru a previously-undiscovered race condition in the two goroutines that stream RPC responses to the websocket. When the input stream returns an error for any reason (for example, the command we're executing has exited), it will unblock the "outer" goroutine and cause a write to the websocket. If we're concurrently writing the "close error" discussed above, this results in a panic from the websocket library.

This changeset includes two fixes:
* Catch "closed pipe" error correctly so that we're not sending unnecessary error messages.
* Move all writes to the websocket into the same response streaming goroutine. The main handler goroutine will block on a results channel, and the response streaming goroutine will send on that channel with the final error when it's done so it can be reported to the user.

Fixes: https://github.com/hashicorp/nomad/issues/19506

---

In addition to the new unit test and associated websocket test infrastructure, I did some soak testing:

> $ for i in {1..200}; do nomad alloc exec 9ee04c03 echo -n "."; done
> ........................................................................................................................................................................................................%

### DIFF
--- a/command/agent/alloc_endpoint.go
+++ b/command/agent/alloc_endpoint.go
@@ -6,6 +6,7 @@ package agent
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -516,7 +517,7 @@ func (s *HTTPServer) allocExec(allocID string, resp http.ResponseWriter, req *ht
 		return nil, err
 	}
 
-	return s.execStreamImpl(conn, &args)
+	return s.execStream(conn, &args)
 }
 
 // readWsHandshake reads the websocket handshake message and sets
@@ -552,7 +553,9 @@ type wsHandshakeMessage struct {
 	AuthToken string `json:"auth_token"`
 }
 
-func (s *HTTPServer) execStreamImpl(ws *websocket.Conn, args *cstructs.AllocExecRequest) (interface{}, error) {
+// execStream finds the appropriate RPC handler and then runs the bidirectional
+// websocket-to-RPC stream
+func (s *HTTPServer) execStream(ws *websocket.Conn, args *cstructs.AllocExecRequest) (any, error) {
 	allocID := args.AllocID
 	method := "Allocations.Exec"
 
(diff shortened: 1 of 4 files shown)


command/agent/alloc_endpoint.go
```diff
@@ -586,33 +596,37 @@ func (s *HTTPServer) execStreamImpl(ws *websocket.Conn, args *cstructs.AllocExec
 		// don't close ws - wait to drain messages
 	}()
 
-	// Create a channel that decodes the results
-	errCh := make(chan HTTPCodedError, 2)
+	// Create a channel for the final result
+	resultCh := make(chan HTTPCodedError)
```
schmichael, line 600: This needs to be buffered to prevent L610 from blocking and leaking the goroutine if this func exits first.

command/agent/alloc_endpoint.go
```diff
@@ -622,39 +636,47 @@ func (s *HTTPServer) execStreamImpl(ws *websocket.Conn, args *cstructs.AllocExec
...
 
+	// retrieve any error and/or wait until goroutine stop and close errCh
+	// connection before closing websocket connection
+	result := <-resultCh
+	ws.Close()
+	return nil, result
+}
+
+// execStreamHandleError writes a CloseMessage to the websocket if we get an
+// error that isn't a ""close error" caused by the RPC pipe finishing up. Note
```
schmichael, line 664:
```suggestion
// error that isn't a "close error" caused by the RPC pipe finishing up. Note
```

schmichael (approved):

command/agent/alloc_endpoint.go
tgross: My first thought here is that there aren't any code paths out of this function that don't involve polling on the `resultCh` first. But you're right that a panic in the `handler` function could cause the outer (request-scoped) goroutine to exit, which would get recovered by the HTTP server.

(I think the original reason it was buffered was because both the goroutine below and the `forwardExecInput` both can send on the channel but we'd only ever pull one result in this request-scoped goroutine.)

Issue #19797, opened 2024-01-22
cgroup resource isolation - Invalid CPU weight value when setting it over 262144 (cgroup v1 max)


twix14:
### Nomad version
1.6.5+ent

### Operating system and Environment details
Linux

### Issue
When setting the cpu parameter, in the resources stanza, to a value larger than 262144 you'll get an invalid value in the cgroup v2 range ([1, 10000]). This happens on drivers which have resource isolation using cgroups. Resulting in an error from the Kernel.

### Reproduction steps
Use the docker driver to schedule an alloc.

```
resources {
   cpu = 300000
}
```
This value gets converted to `11443`, which is outside the cgroup v2 range.

```
ca4998934a4e8c94548c1430c19875634b8682222853f3cea42aa739067b9d8d: API error (500): failed to create shim task: 
OCI runtime create failed: runc create failed: 
unable to start container process: error during container init: error setting cgroup config for procHooks process: failed to write "11443": 
write /sys/fs/cgroup/nomad.slice/ca4998934a4e8c94548c1430c19875634b8682222853f3cea42aa739067b9d8d/cpu.weight: 
numerical result out of range: unknown
```

#### Expected Result
The value set in the CPU should accept and work with values in the range of `[1, MAX_NOMAD_CPU_AVAILABLE]`, converting that value into a cgroup v2 one. 
Example:
- Setting a CPU share of 300000MHz
- Nomad Client node with 353600MHz of cpu compute available
- Should result in a cpu weight of around 8484. Meaning the process will get 84.8% of CPU cycles and other workloads managed by nomad can use the rest. Resulting in a correct weight distribution when it comes to CPU.

#### Actual Result
Error returned by the Kernel when setting a value outside the cgroups v1 range, resulting in an invalid value in the cgroups v2 range.

lgfa29:
Hi @twix14 👋 

Thanks for the report. That's a lot of CPU you got there 😅 

Limiting the value for `resource.cpu` to `262144` is probably the only workaround for now. I think Kubernetes faced the same issue in https://github.com/kubernetes/kubernetes/issues/92855 and fixed it by clamping the maximum value allowed to `262144` (https://github.com/kubernetes/kubernetes/pull/93248). 

I will check with the rest of the team about how to handle this.

PR #19935, opened 2024-02-09
executor: limit the value of CPU shares

lgfa29:
The value for the executor cgroup CPU weight must be within the limits imposed by the Linux kernel.

Nomad used the task `resource.cpu`, an unbounded value, directly as the cgroup CPU weight, causing it to potentially go outside the imposed values.

This commit clamps the CPU shares values to be within the limits allowed.

Closes https://github.com/hashicorp/nomad/issues/19797

### DIFF
--- a/.changelog/19935.txt
+++ b/.changelog/19935.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-client: Ensure the value for CPU weight is within the allowed range
+client: Ensure the value for CPU shares are within the allowed range
 ```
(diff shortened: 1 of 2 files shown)


.changelog/19935.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+client: Ensure the value for CPU weight is within the allowed range
```
tgross, line 2:
```suggestion
client: Ensure the value for CPU shares are within the allowed range
```

tgross (approved):
LGTM

PR #19872, opened 2024-02-05
[ui] HashiCorp Design System upgraded to 3.6.0

philrenaud:
Upgrades to the latest [Helios](https://helios.hashicorp.design/whats-new/release-notes#360) (and corresponding icons/branding changes)

(Pending visual regression check w/ Percy)

PR #19953, opened 2024-02-12
[ui] Upgrades to build storybook on node v20

philrenaud:
This unblocks Vercel builds which had otherwise been failing since https://github.com/hashicorp/nomad/pull/19872

An earlier attempt, https://github.com/hashicorp/nomad/pull/19914, is still a good idea but didn't address the ultimate culprit here.

Storybook now builds locally and on Vercel with the newly-set versions, and no longer throw `@ember/component` or `babel-plugin-dynamic-import-node explicitly declared`-related errors.


gulducat (approved):
without knowing the full implications of these magic versions, LGTM!

DingoEatingFuzz (commented):
Way to get to the bottom of it! 🎊

PR #19954, opened 2024-02-12
CNI: fix deprecation warnings

tgross:
We updated our `go-cni` dependency in #17582 but this left deprecation warnings on the `cni.CNIResult` type (now `cni.Result`).

(Nothing user-facing in here, so no changelog entry)


gulducat (approved):
LGTM!

to put a fine point on it, `CNIResult` and `Result` are literally identical: https://github.com/containerd/go-cni/blob/v1.1.9/deprecated.go#L24-L25

PR #19960, opened 2024-02-12
ci: run missing tests

lgfa29:
The packages added to `ci/test-core.json` were not being tested in this release branch. Futhermore, the `client/vaultclient` package had an import cycle that was only fixed in `main`.


gulducat (approved):
:test_tube: :tada:

Issue #19805, opened 2024-01-23
Allocations stuck in pending state after a stopped node comes back online


roman-vynar:
### Issue
When a previously stopped node comes back online and gets allocations rescheduled from another node being replaced at the same moment we may get allocations stuck in `pending` state.
It happens frequently when I replace a node but not necessary. Also not necessary to all allocations. Some may get running ok, some may stuck `pending`.

### Nomad version
It is perfectly ok on Nomad 1.3.5 (server/client).
Then I was observing this after upgrade on 1.7.1, 1.7.2, 1.7.3 for a few weeks already (both server/client of course).

### Reproduction steps
It is usually happen when a node comes back online after previously being stopped. Also once I have seen this during a normal job update process.

#### Expected Result
Pending allocations should be somehow self-resolvable?

#### Actual Result
Allocations in `pending` state are stuck in this state forever, most longer I was waiting a day or two.

### Nomad Server logs
Nothing related.

### Nomad Client logs
Only 2 entries for each of such allocations:
```
    2024-01-23T09:05:45.639Z [INFO]  client.alloc_runner.task_runner: Task event: alloc_id=1d241d0a-b791-f0e5-5714-f11ac0e18ad8 task=courier type=Received msg="Task received by client" failed=false
    2024-01-23T09:05:46.940Z [INFO]  client.alloc_runner.task_runner: Task event: alloc_id=1d241d0a-b791-f0e5-5714-f11ac0e18ad8 task=courier type="Task Setup" msg="Building Task Directory" failed=false
```
<img width="971" alt="Screenshot 2024-01-23 at 14 39 12" src="https://github.com/hashicorp/nomad/assets/2220941/710694d5-d0eb-4259-8dea-590f2765068e">
<img width="1240" alt="Screenshot 2024-01-23 at 14 38 25" src="https://github.com/hashicorp/nomad/assets/2220941/d681f678-d57f-43d2-9b19-07f935a12b21">

### Workaround
In order to fix this I need to whether stop/start each job having pending allocations. When there are multiple jobs, another option is to restart nomad client.

lgfa29:
Hi @roman-vynar 👋 

A few questions to help us understand the problem:

1. Would it be possible to share the job file? Feel free to redact any private information.
2. Was the task running prior to the client agent shutdown and then it got stuck at `pending` when the client started again?
[...]

PR #19963, opened 2024-02-13
vault: revert #18998 to fix potential deadlock

lgfa29:
https://github.com/hashicorp/nomad/issues/19805 reports an issue where tasks get stuck in `pending` status. Logs shared by the user point that the problem is likely that `vault_hook.Prestart()` never completes.

```
{"log":"    2024-02-09T15:16:40.868Z [TRACE] client.alloc_runner.task_runner: running prestart hook: alloc_id=95eddd7d-*** task=*** name=vault start=\"2024-02-09 15:16:40.868679902 +0000 UTC m=+27.567861561\"\n","stream":"stdout","time":"2024-02-09T15:16:40.868767694Z"}
{"log":"    2024-02-09T15:21:40.746Z [TRACE] client.alloc_runner.runner_hook.alloc_health_watcher: deadline reached; setting unhealthy: alloc_id=95eddd7d-*** deadline=\"2024-02-09 15:21:40.746133303 +0000 UTC m=+327.445314950\"\n","stream":"stdout","time":"2024-02-09T15:21:40.746292101Z"}
```

```
{"log":"    2024-02-12T09:05:52.690Z [TRACE] client.alloc_runner.task_runner: running prestart hook: alloc_id=96dade39-*** task=*** name=vault start=\"2024-02-12 09:05:52.69095185 +0000 UTC m=+24.113560832\"\n","stream":"stdout","time":"2024-02-12T09:05:52.692175834Z"}
... user stops faulty alloc ...
{"log":"    2024-02-12T09:10:12.183Z [TRACE] client.alloc_runner.task_runner: finished prestart hook: alloc_id=96dade39-*** task=*** name=vault end=\"2024-02-12 09:10:12.183310806 +0000 UTC m=+283.605919788\" duration=4m19.492358956s\n","stream":"stdout","time":"2024-02-12T09:10:12.183382225Z"}
```

Following the code path along with the logs, the hook is likely getting stuck waiting for a Vault token:
https://github.com/hashicorp/nomad/blob/1bde7a8fb438cf5579592688b4bd745b139b49ef/client/allocrunner/taskrunner/vault_hook.go#L216-L221

`h.updater.updatedVaultToken(h.future.Get())` only contains one blocking call in `setVaultToken`, but the lock is unlikely to be blocked since the task is still being created.
https://github.com/hashicorp/nomad/blob/1bde7a8fb438cf5579592688b4bd745b139b49ef/client/allocrunner/taskrunner/task_runner_getters.go#L65-L67

`h.updater.updatedVaultToken(h.future.Get())` also triggers the the update hooks, so we would expect log entries that read `running update hooks`, but none is emitted for the task.
https://github.com/hashicorp/nomad/blob/1bde7a8fb438cf5579592688b4bd745b139b49ef/client/allocrunner/taskrunner/vault_hook.go#L54-L55

In addition to deriving a Vault token for the task, `vault_took.Prestart()` also immediately renews it, blocking until the operation completes.
https://github.com/hashicorp/nomad/blob/1bde7a8fb438cf5579592688b4bd745b139b49ef/client/allocrunner/taskrunner/vault_hook.go#L296-L306

#18998 refactored the Vault token renewal loop in order to prevent a possible race condition where expired tokens are re-added to the renewal heap, but I suspect that implementation may have introduced new race conditions.

While trying to reproduce #19805 I bumped into another issue where  the new implementation of `vaultclient.RenewToken()` can get stuck when renewing the same token concurrently. 43bec053ca07a6b9100be92e213bebeadaa18b7a introduces a test for this bug.

This scenario is unlikely to be triggered in a real Nomad cluster because each Vault token is owned by a specific entity (task runners), but it's still a bug and the new implementation may contain others that are responsible for the behaviour observed in #19805.

Since the main issue with Vault renewals was fixed in https://github.com/hashicorp/nomad/pull/18985 I think it's better to rollback this refactored code until we are able to guarantee its correctness.

### DIFF
--- a/client/vaultclient/vaultclient.go
+++ b/client/vaultclient/vaultclient.go
@@ -112,11 +112,6 @@ type vaultClient struct {
 // vaultClientRenewalRequest is a request object for renewal of both tokens and
 // secret's leases.
 type vaultClientRenewalRequest struct {
-	// renewalLoopCh is used to notify listeners every time the token goes
-	// through the renewal loop. It does not guarantee the renewal was
-	// successful, so listeners should also read from errCh for renewal errors.
-	renewalLoopCh chan struct{}
-
 	// errCh is the channel into which any renewal error will be sent to
 	errCh chan error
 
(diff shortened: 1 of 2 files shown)


tgross (approved):
LGTM. I think we should probably take a second look at the overall design of the renewal loop -- the notion that we immediately renew when we derive the token seems bogus and it introduces extra places for us to block.

Issue #10628, opened 2021-05-20
Add support for transparent connect proxies


apollo13:
Hi, I know that https://www.consul.io/docs/connect/transparent-proxy is still beta, but it would be sooooo great if Nomad could support that in the next releases :)

tgross:
Hi @apollo13! Much of what transparent proxy provides is for filling gaps in k8s (because it's a bit more "kit of parts" than combining Nomad and Consul). There's a conflict here in the security model between the two: Consul transparent proxy assumes and requires 1 service per network namespace, whereas Nomad supports _n_ services per network namespace. We already provide most of what transparent proxies support, with the major exception of preventing service mesh circumvention.

So we're still brainstorming what exactly transparent proxy support would look like in Nomad. It's likely to be implemented a set of features that together make up all the same things as the Consul tproxy, rather than a standalone "support Consul tproxy" feature. Some ideas we've considered, not all of which are compatible with each other:
* Implement a `network.mode = "transparent"` that updates the routing table in the allocation's network NS to prevent circumvention.
* Expose Consul DNS to allocations.
* Create a separate net NS for each service. This could complicate a lot of use cases for task groups with multiple tasks that communicate over localhost rather than Unix Domain Sockets.

In any case, just a heads up that this isn't on our very near-term roadmap.

apollo13:
Ok, my main goal is to get rid of the manual declaration of all the connect services that I need. If there are ideas on how to do that without manually specifying all, then it doesn't have to be the transparent proxy. Thank you for the clarifications!

shoenig:
@apollo13 if you're talking about no longer needing to define `upstreams`, then transparent proxy only helps in that regard in that upstreams are inherited from `intentions`, so that you only need to declare them once. (Unless intentions are disabled, in which case the default behavior is a free-for-all). Making services [connect-native](https://www.consul.io/docs/connect/native#connect-native-app-integration) is another way to eliminate the need for upstreams, but of course that only helps if you own the code. (And you'd still need to declare intentions)

apollo13:
Declaring them only once would imo already be a win. connect-native is certainly one way of doing things but often not easily possible for existing code (For example changing a database connection in Django to be connect native is probably close to not possible). So in that sense the proxy would still be a massive improvement (at least for me). I usually do not care about inbound connections that much (they are solved differently). But it would be really great if I could limit outgoing connections of a group to whatever the intentions allow (and not more).

kirooshu:
[...]

PR #19969, opened 2024-02-13
fingerprint: add DNS address and port to Consul fingerprint

tgross:
In order to provide a DNS address and port to Connect tasks configured for transparent proxy, we need to fingerprint the Consul DNS address and port. The client will pass this address/port to the iptables configuration provided to the `consul-cni` plugin.

Ref: https://github.com/hashicorp/nomad/issues/10628


client/fingerprint/consul.go
```diff
@@ -298,6 +303,60 @@ func (cfs *consulFingerprintState) grpcTLSPort(info agentconsul.Self) (string, b
...
+func (cfs *consulFingerprintState) dnsAddr(logger hclog.Logger) func(info agentconsul.Self) (string, bool) {
+	return func(info agentconsul.Self) (string, bool) {
+
+		// only addresses we can use for an iptables rule from a container to the
+		// host will be fingerprinted
+		isValidForTaskUse := func(addr netip.Addr) (string, bool) {
+			if !addr.IsLoopback() && !addr.IsUnspecified() && addr.IsValid() {
+				return addr.String(), true
+			}
+			return "", false
```
gulducat, line 322: this is really "6 of one, half a dozen of the other" but my instinct is to reverse the logic here

```suggestion
			if !addr.IsValid() || addr.IsLoopback() || addr.IsUnspecified() {
				return "", false
			}
			return addr.String(), true
```

up to your preference :shrug:

client/fingerprint/consul.go
```diff
@@ -298,6 +303,60 @@ func (cfs *consulFingerprintState) grpcTLSPort(info agentconsul.Self) (string, b
...
+		dnsAddrs, ok := info["DebugConfig"]["DNSAddrs"].([]string)
+		if ok {
+			for _, dnsAddr := range dnsAddrs {
+				dnsAddr = strings.TrimPrefix(dnsAddr, "tcp://")
+				dnsAddr = strings.TrimPrefix(dnsAddr, "udp://")
+
+				parsed, err := netip.ParseAddrPort(dnsAddr)
+				if err != nil {
+					logger.Warn("could not parse Consul addresses.dns config", "value", dnsAddr)
+					return "", false
```
gulducat, line 335: if any of the DNSAddrs are unparseable, then this return would prevent us from attempting the bind addr fallback below eh?

client/fingerprint/consul.go
```diff
@@ -298,6 +303,60 @@ func (cfs *consulFingerprintState) grpcTLSPort(info agentconsul.Self) (string, b
...
+			}
+		}
+
+		// fallback to the bind address
+		bindAddr, ok := info["DebugConfig"]["BindAddr"].(string)
+		if ok {
+			parsed, err := netip.ParseAddr(bindAddr)
+			if err != nil {
+				logger.Warn("could not parse Consul bind_addr config", "value", bindAddr)
+				return "", false
```
gulducat, line 349: this is the fallback, but again is `false` what we want? or maybe down the bottom on 356, would we want `true` there, if no value could be found?

I'm just trying to square these with the docstring on `consulExtractor`

> // consulExtractor is used to parse out one attribute from consulInfo. Returns
> // the value of the attribute, and whether the attribute exists.

in this case, since this is the last fallback attempt, `false` seems appropriate since we could not find a value.  but then the final return below has me a bit confused.

gulducat (approved):
I'm mainly looking for clarification around what exactly the `bool` return represents on a `consulExtractor` func, and whether we want to return early on any parse errors encountered along the way as happens here currently, or if we instead want to continue along down the line trying the other options.

other than that, lgtm!

client/fingerprint/consul.go
tgross: Yeah the only reason we'd be getting an error here is if the response body format changed in a way we didn't know about, so trying to guess and continue is the wrong approach.

client/fingerprint/consul.go
tgross: It's a little subtle and please push back if you disagree, but I'm trying to differentiate between "Consul doesn't give us a DNS address" and "Consul gave us a malformed address, which suggests something's very wrong and we shouldn't start the client"

client/fingerprint/consul.go
gulducat: ah hm, I see in the [Consul API docs](https://developer.hashicorp.com/consul/api-docs/agent#read-configuration) that

> `DebugConfig` contains the full runtime configuration but its format is subject to change without notice or deprecation.

are you looking to guard against that warning regarding the structure of the API response?  wouldn't the `ok` check on the `["DNSAddrs"].([]string)` type coercion cover that angle?  I suppose it still could change format ("proto://" etc) within the `[]string`s...

I was assuming that perhaps operators could slightly misconfigure Consul in such a way that one address could be invalid, but other(s) being right and proper could let Consul roll with it, and if that's the case, maybe we could too.  if that assumption is wrong though, and a single invalid address casts doubt on the entire effort, then bailing as early as possible as you've done here makes good sense to me.

client/fingerprint/consul.go
gulducat: that does seem a good thing to distinguish, and excellent to log as you've done here, but is that distinction meaningful elsewhere that considers "whether the attribute exists"? mainly farther down, does `return "", true` indicating that "the value exists, but it is an empty string" make sense for later use, when actually we have failed to find a value?

pardon my ignorance on the nuances.

tgross: Yeah, the more I think about it, introducing this subtlety is a bug waiting to happen. We're using a portion of the Consul API response which is marked "unstable" but in practice has been unchanged for a very long time (and in a sidebar with the Consul folks we decided that's fine). If this changes in any way, we should return `"", false` which will bubble up as an error so that we catch it in E2E testing before it ships.

So lemme fix that up and make sure it's always `"", false` if it can't be read.

tgross:
@gulducat I've reworked the logic here a good bit, so that we're not using the `bind_addr` if we get 0.0.0.0 but instead are querying the private or public IP. If we get localhost, we're returning `"", true` gracefully.

client/fingerprint/consul.go
```diff
@@ -298,6 +304,72 @@ func (cfs *consulFingerprintState) grpcTLSPort(info agentconsul.Self) (string, b
...
+
+		dnsAddrs, ok := info["DebugConfig"]["DNSAddrs"].([]string)
+		if ok {
+			for _, dnsAddr := range dnsAddrs {
+				dnsAddr = strings.TrimPrefix(dnsAddr, "tcp://")
+				dnsAddr = strings.TrimPrefix(dnsAddr, "udp://")
+
+				parsed, err := netip.ParseAddrPort(dnsAddr)
+				if err != nil || !parsed.IsValid() {
+					logger.Warn("could not parse Consul addresses.dns config", "value", dnsAddr)
```
gulducat, line 327: are the errors from `ParseAddrPort` helpful at all?  if so, I might like to include the error string in the log line, rather than bundling these two sad conditions together.  I can just imagine myself at some point in future looking at a bug report and wondering which of these got tripped.

client/fingerprint/consul.go
```diff
@@ -298,6 +304,72 @@ func (cfs *consulFingerprintState) grpcTLSPort(info agentconsul.Self) (string, b
...
+
+			privateIP, err := sockaddr.GetPrivateIP()
+			if err != nil {
+				logger.Warn("could not query network interfaces", "error", err)
+				return "", false // something is very wrong, so bail out
+			}
+			if privateIP != "" {
+				return privateIP, true
+			}
+			publicIP, err := sockaddr.GetPublicIP()
```
gulducat, line 356: I imagine the "fallback to private or public IP" test hits privateIP above so misses this path, leaving it untested, but to do anything about that (like a mockable interface) seems more trouble than its worth.  I just wanted to call it out in case you can think of a simple solution off the top of your head.

client/fingerprint/consul.go
gulducat: it occurs to me now too that this path isn't covered in tests?

client/fingerprint/consul.go
```diff
@@ -298,6 +304,72 @@ func (cfs *consulFingerprintState) grpcTLSPort(info agentconsul.Self) (string, b
...
+			}
+			if publicIP != "" {
+				return publicIP, true
+			}
+		}
+
+		// if we've hit here, Consul is bound on localhost and we won't be able
+		// to configure container DNS to use it, but we also don't want to have
+		// the fingerprinter return an error
+		return "", true
```
gulducat, line 369: I think this return is untested too? maybe this one is worth covering, and I think not too difficult to do.

gulducat (approved):
The intent is way clearer now, nice work!

just a couple comments about test coverage which you can button up or not at your discretion.

client/fingerprint/consul.go
tgross: Fixed!

client/fingerprint/consul.go
tgross: Yeah, that'd have been nice, but it'd mostly be testing the library code anyways, which hopefully is well-tested already.

client/fingerprint/consul.go
tgross: Two bits there:
* I looked at the `IsValid()` method and it's just checking the zero value. So we can throw that away and just check the error return instead. (I guess the method exists for cases where you've constructed the `netip.Addr` manually?)
* When I re-tested the new code against a real Consul, it turns out we should be casting to `[]any` and not `[]string`. So the tests all passed but it didn't fingerprint correctly in the real world.

I've updated the code to work correctly on real Consul, barf on parsing errors as we'd expect, and added a bunch more test coverage of malformed response bodies.

client/fingerprint/consul.go
tgross: Ran thru with a `-cover` profile and this is now the only bit we're missing coverage on.

gulducat: > it'd mostly be testing the library code anyways

I do try to avoid that, yeah.  the missing coverage here is not how sockaddr does its work, but that we do ask it to do its work, when appropriate.  the logic here is so simple though that it seems fine to skimp on that little bit of coverage.

PR #19972, opened 2024-02-14
Fix nil pointer dereference if alloc has nil Job

cleroux:
We encountered the following on one of our production hosts:

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -1876,6 +1876,14 @@ func TestClient_hasLocalState(t *testing.T) {
 
 	c.stateDB = cstate.NewMemDB(c.logger)
 
+	t.Run("nil Job", func(t *testing.T) {
+		alloc := mock.BatchAlloc()
+		alloc.Job = nil
+		c.stateDB.PutAllocation(alloc)
+
+		require.False(t, c.hasLocalState(alloc))
+	})
+
 	t.Run("plain alloc", func(t *testing.T) {
 		alloc := mock.BatchAlloc()
 		c.stateDB.PutAllocation(alloc)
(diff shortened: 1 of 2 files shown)


client/client_test.go
```diff
@@ -1876,6 +1876,14 @@ func TestClient_hasLocalState(t *testing.T) {
 
 	c.stateDB = cstate.NewMemDB(c.logger)
 
+	t.Run("nil Job", func(t *testing.T) {
+		alloc := mock.BatchAlloc()
+		alloc.Job = nil
+		c.stateDB.PutAllocation(alloc)
+
+		require.False(t, c.hasLocalState(alloc))
```
tgross, line 1884: We're slowly migrating all our uses of `testify` to [`shoenig/test`](https://pkg.go.dev/github.com/shoenig/test) instead, which fixes some correctness issues around equality testing. So let's make sure we're doing that on new tests:

```suggestion
		must.False(t, c.hasLocalState(alloc))
```

tgross (changes requested):
Hi @cleroux!

This looks good. Can you run `make cl` to add a changelog entry for this bug? Something like "client: Fixed a bug where corrupt client state could panic the client"

cleroux:
Thanks, @tgross!  Requested changes have been made.

tgross (approved):
LGTM!

Once CI is done I'll merge this and get it backported. Thanks @cleroux!

tgross:
The failing storybook test is unrelated and is being worked on. The flaky leadership test in `tests-groups (nomad)` is known, being tracked, and unrelated to this work.

PR #19985, opened 2024-02-14
ui: Confirmation on exit from exec as long as socket has been opened

### LINKED ISSUE #19849: Ctrl+W in Exec in web UI closes the window without prompting
### Nomad version
I don't know because I'm using the web interface and it's not telling.

Firefox

### Issue
The Exec feature in the Nomad web interface is very useful. However, one small problem is that Ctrl+W on the web means to close the current tab. I find myself frequently pressing with the intention to delete the last word, but it closes the entire window instead.

If it's not possible to capture the key bind, it would be nice if Nomad asked for confirmation when closing an Exec window that's still connected to a container.

### Reproduction steps
Open Nomad web UI
Open a job
Click Exec
Select a task
Press Enter to run the pre-filled command
Type something
Press Ctrl+W

#### Expected Result
Ideally, delete the last typed word (standard bash behaviour).
Slightly less ideally, make the browser prompt me if I really intend to close the tab.

#### Actual Result
The window closes, losing my session entirely.

Thanks!

philrenaud:
Resolves #19849 

In an exec window context, navigating away from the page is not possible by in-page links, as there are none (besides Documentation, which opens in a new tab). Otherwise, closing the window by clicking exit, or using a keyboard shortcut like ctrl+w, ctrl+q, etc. is the most common method.

Unfortunately, some of those keyboard shortcuts are also used for in-terminal activity (like `ctrl+w` to delete a word in bash, or split a terminal in vim), and users' muscle memory may cause them to close a window, losing time and context along the way.

This PR adds a beforeunload handler to the exec terminal component, triggered whenever a socket is open and the user does an action that would close the parent tab/window. It will add a prompt asking the user if they meant to close it, and if they hit cancel or escape, it will bring their context back to the cursor in the terminal. Otherwise, the window will be closed.

### DIFF
--- a/.changelog/19985.txt
+++ b/.changelog/19985.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: Prompt a user before they close an exec window to prevent accidental close-browser-tab shortcuts that overlap with terminal ones
+```
--- a/ui/app/components/exec-terminal.js
+++ b/ui/app/components/exec-terminal.js
@@ -3,15 +3,21 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
+// @ts-check
+
 import Component from '@ember/component';
 import { FitAddon } from 'xterm-addon-fit';
 import WindowResizable from '../mixins/window-resizable';
 import { classNames } from '@ember-decorators/component';
 import classic from 'ember-classic-decorator';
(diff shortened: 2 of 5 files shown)


lgfa29 (approved):
⌨️

PR #19989, opened 2024-02-14
ui: Warning and better error handling for invalid-named variables and job templates

philrenaud:
(Note: custom job templates are backed by Nomad Variables and so the variable adapter handleResponse method tackles them both)

When providing an invalid name for a variable or job template in the UI, give the user a message in realtime as they type to proactively warn of an invalid name, and if they continue to try to save regardless, transform the otherwise opaque error message (mentions an RPC error) into something more tangible.

image


shoenig (approved):
Yes this looks like javascript

PR #19990, opened 2024-02-14
connect: fix missing diff of expose block

tgross:
While working on https://github.com/hashicorp/nomad/issues/10628 I discovered that the `expose` block was missing an implementation of Diff, which means it doesn't show up correctly in `job plan` output.

I've also fixed a field comparison bug in `ServiceCheck.Equal` that doesn't appear to impact any production code.


lgfa29 (approved):
LGTM!

shoenig (approved):

PR #20001, opened 2024-02-15
Flight Icons Migration

Dhaulagiri:
This is a start at migrating existing structure icons to FlightIcons. Some caveats that will need to be investigated further by someone more familiar with the codebase

* There are a handful of `x-icon` instances left that either take in arguments or didn't have a defined mapping to a new icon when we made the list 2+ years ago. These will need to be looked at on a case-by-case basis.
* The migration of names between structure and flight was a bit more complicated than I expected. They should be mapped correctly, but if they aren't the component will fail to render.
* The actual work to remove the structure-icons dependency and `x-icon` still needs to be done


ui/app/templates/components/allocation-row.hbs
```diff
@@ -31,7 +31,7 @@
       role="tooltip"
       aria-label="Allocation was preempted"
     >
-      {{x-icon "boot" class="is-faded"}}
+      <FlightIcon @name="boot" class="is-faded" />
```
DingoEatingFuzz, line 34: There is no boot flight icon

DingoEatingFuzz: Take my money 
image

ui/app/components/nomad-icon.hbs
```diff
@@ -0,0 +1,10 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: BUSL-1.1
+~}}
+
```
philrenaud, line 5: SVG code from the formerly-separate .svg file, now as a component. Used in our `<LoadingSpinner />` component.

ui/app/models/node.js
```diff
@@ -127,9 +127,9 @@ export default class Node extends Model {
     if (this.isDraining || !this.isEligible) {
       return 'alert-circle-fill';
     } else if (this.status === 'down') {
-      return 'cancel-circle-fill';
+      return 'x-circle-fill';
     } else if (this.status === 'initializing') {
-      return 'node-init-circle-fill';
+      return 'entry-point';
```
philrenaud, line 132: Using this for initializing nodes now: 
image

ui/app/styles/core/icon.scss
```diff
@@ -57,3 +57,33 @@ $icon-dimensions-large: 2rem;
     color: $grey;
   }
 }
+
+// Flight Icons general override: notes:
```
philrenaud, line 61: I don't feel great about this but it seems necessary given Nomad's current status in a transitory period between a its historical UI design and a more Helios-centered one.

ui/app/templates/components/allocation-row.hbs
```diff
@@ -11,7 +11,7 @@
       role="tooltip"
       aria-label="Allocation depends on unhealthy drivers"
     >
-      {{x-icon "alert-triangle" class="is-warning"}}
+      <FlightIcon @name="alert-triangle-fill" @color="#fa8e23" />
```
philrenaud, line 14: General pattern: bulma classes to add colour vs expressly using the @color prop.

ui/app/templates/components/job-page/parts/task-groups.hbs
```diff
@@ -40,7 +40,6 @@
         <TaskGroupRow
           @data-test-task-group={{row.model.name}}
           @taskGroup={{row.model}}
-          @onClick={{fn this.gotoTaskGroup row.model}}
```
philrenaud, line 43: Side-effect: noticed in testing that clicking the exact centre of task group row's scaling buttons would route you into that task group. Long-term this gets fixed by upgrading to Helios tables.

ui/app/styles/core/icon.scss
Dhaulagiri: agreed it's not ideal as we would prefer these class names to be more of a private API, however, you're not the first (and won't be the last) to hook into them for specific situations that arise in our products.

philrenaud (approved):

philrenaud:
Closing in favo(u)r of https://github.com/hashicorp/nomad/pull/25353

Dhaulagiri:
farewell-solenya

PR #20006, opened 2024-02-16
[refactor] Move task directory destroy logic from alloc_dir.go to task_dir.go

angrycub:
This PR moves the task directory teardown logic to the task_dir.go file since the build logic lives there.

Some additional refactorings include:
- Made constants for common directory and file permissions that we use to avoid the "777" versus "0777" issues
- Made functions to wrap the common case of making directories and then expanding their permissions.
- Standardized uses of multierror.
- Converted fmt.Errorf calls to wrap error rather than stringifying it.


client/allocdir/task_dir.go
```diff
@@ -257,3 +237,38 @@ func (t *TaskDir) embedDirs(entries map[string]string) error {
 
 	return nil
 }
+
+func (t *TaskDir) Unmount() error {
```
schmichael, line 241:
```suggestion
// Unmount or delete task directories. Returns all errors as a multierror.
func (t *TaskDir) Unmount() error {
```

schmichael (approved):
:tada:

client/allocdir/alloc_dir.go
```diff
@@ -28,6 +28,15 @@ const (
...
 	// Windows) that don't support integer ownership identifiers.
 	idUnsupported = -1
+
+	// fileMode777 is a constant that represents the file mode rwxrwxrwx
+	fileMode777 = os.FileMode(0o777)
+
+	// fileMode755 is a constant that represents the file mode rwxr-xr-x
+	fileMode755 = os.FileMode(0o755)
+
+	// fileMode644 is a constant that represents the file mode rw-r--r--
```
tgross, line 38: The comment doesn't match the implementation:

```suggestion
	// fileMode666 is a constant that represents the file mode rw-rw-rw-
```

tgross (approved):
LGTM :+1:

client/allocdir/alloc_dir.go
```diff
@@ -28,6 +28,15 @@ const (
...
+	fileMode666 = os.FileMode(0o666)
```
shoenig, line 39: the comment is `0644` but the variable and value are `0666`

PR #20015, opened 2024-02-20
docs: fix s/envoy-bootstrap-error redirect

schmichael:
And cleanup whitespace

We use this s link in a log message: https://github.com/hashicorp/nomad/blob/v1.7.4/client/allocrunner/taskrunner/envoy_bootstrap_hook.go#L382

### DIFF
--- a/website/redirects.js
+++ b/website/redirects.js
@@ -30,16 +30,22 @@ module.exports = [
   },
   */
 
-	/**
-	 * /s/* redirects for useful links that need a stable URL but we may need to
-	 * change its destination in the future.
-	 */
-	{
-		source: '/nomad/s/vault-workload-identity-migration',
-		destination:
+  /**
+   * /s/* redirects for useful links that need a stable URL but we may need to
+   * change its destination in the future.
+   */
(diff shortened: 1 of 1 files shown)


lgfa29 (approved):
Hum...I don't think we should've changed these URLs in code. https://www.nomadproject.io/s/envoy-bootstrap-error still works fine and it's a more stable URL since we have some control over the domain. `developer.hashicorp.com` is a bit generic and may change in the future (I've seen folks confused about the use of the `developer` subdomain for example).

I suspect all URLs in https://github.com/hashicorp/web/blob/f92c006335021217a995b19129956e7323dd3c51/apps/nomad-io/src/redirects/index.js#L28-L49 are also broken if they're tried to be accessed from `developer.hashicorp.com`.

Hopefully this won't break https://www.nomadproject.io/s/envoy-bootstrap-error 😅

schmichael:
@lgfa29 yeah +1 to using nomadproject.io/s/ in the future. I just wanted to fix the link in the shipped log message ASAP.

PR #20022, opened 2024-02-21
docs: Consul Admin partition example and updates

boruszak:
This PR updates the consul Block page for admin partition support, per [CE-555](https://hashicorp.atlassian.net/browse/CE-555).

The following changes were made to the page:

- Admin partitions example added 
- `CONSUL_PARTITION` environment variable sentence and link added
- Description of a Consul partition was modified. It's not that Consul Community has access to zero admin partitions. There is one admin partition, called `default` that Community edition is limited to.

## Preview

[Link to deployment preview](https://nomad-2ncfwj0sc-hashicorp.vercel.app/nomad/docs/job-specification/consul#consul-admin-partition)

[CE-555]: https://hashicorp.atlassian.net/browse/CE-555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/content/docs/job-specification/consul.mdx
+++ b/website/content/docs/job-specification/consul.mdx
@@ -82,10 +82,6 @@ If the Nomad cluster is [configured][config_consul_namespace] to use [Consul Nam
 `CONSUL_NAMESPACE` environment variable will be injected whenever `CONSUL_TOKEN`
 is set.
 
-If the Nomad cluster is [configured][config_consul_namespace] to use [Consul Admin Partitions][], a
-`CONSUL_PARTITION` environment variable will be injected whenever `CONSUL_TOKEN`
-is set.
-
 The [`template`][template] block can use the Consul token as well.
 
 ### `consul` Parameters
@@ -104,7 +100,7 @@ The [`template`][template] block can use the Consul token as well.
 - `partition` `(string: "")` - When this field is set, a constraint will be
   added to the group or task to ensure that the allocation is placed on a Nomad
   client that has a Consul Enterprise agent in the specified Consul [admin
-  partition][]. Note that Consul Community Edition agents are limited to the `default` partition, so only Consul Enterprise users should configure this field.
+  partition][].
 
 ## `consul` Examples
 
@@ -236,15 +232,17 @@ job "docs" {
 
 ### Consul Admin Partition
 
-This example demonstrates how to configure Consul admin partitions for different tasks within a group.
+This example demonstrates how to configure Consul admin partitions for different
+tasks within a group. The Consul client agent must separately specify the admin
+partition in the agent configuration. Refer to the Consul documentation's
+[agent configuration reference][] for more information.
 
 <EnterpriseAlert />
 
-In the following example:
+In the following example, the `template` block in the `web` and `app` tasks use
+the default Consul cluster. It obtains a token that allows it access to the `prod`
+admin partition in Consul.
 
-- The `template` block in the `web` task uses the default Consul cluster. It obtains a token that allows it access to the `engineering` admin partition in Consul. 
-- The `template` block in the `app` task uses the Consul cluster named `prod-apps`. It obtains a token that allows it access to the
-`engineering` Consul admin partition.
 
 ```hcl
 job "docs" {
@@ -253,7 +251,9 @@ job "docs" {
     task "web" {
 
       consul {
-        partition = "engineering"
+        cluster   = "default"
+        namespace = "default"
+        partition = "prod"
       }
 
       template {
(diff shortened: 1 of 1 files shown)


website/content/docs/job-specification/consul.mdx
```diff
@@ -76,11 +76,16 @@ to Using Workload Identity with Consul</a>
...
-[configured][config_consul_namespace] to use [Consul Namespaces][], a
+`CONSUL_TOKEN` environment variable in the task. 
+
+If the Nomad cluster is [configured][config_consul_namespace] to use [Consul Namespaces][], a
 `CONSUL_NAMESPACE` environment variable will be injected whenever `CONSUL_TOKEN`
 is set.
 
+If the Nomad cluster is [configured][config_consul_namespace] to use [Consul Admin Partitions][], a
+`CONSUL_PARTITION` environment variable will be injected whenever `CONSUL_TOKEN`
+is set.
```
tgross, line 87: Not sure why we're making this change, as it's not correct. We don't inject the env var for partitions.

website/content/docs/job-specification/consul.mdx
```diff
@@ -99,9 +104,7 @@ The [`template`][template] block can use the Consul token as well.
 - `partition` `(string: "")` - When this field is set, a constraint will be
   added to the group or task to ensure that the allocation is placed on a Nomad
   client that has a Consul Enterprise agent in the specified Consul [admin
-  partition][]. Note that Consul Community Edition agents are not assigned to
-  any admin partition, so this field should not be used without Consul
-  Enterprise.
+  partition][]. Note that Consul Community Edition agents are limited to the `default` partition, so only Consul Enterprise users should configure this field.
```
tgross, line 107: Our testing while working on this feature showed that Consul CE doesn't return a fingerprint for the partition at all, not that it's in a "default" partition. The documentation we have here is correct for the way Nomad's fingerprinting works.

(Also, we hard-wrap markdown)

website/content/docs/job-specification/consul.mdx
```diff
@@ -231,6 +234,51 @@ job "docs" {
...
+
+This example demonstrates how to configure Consul admin partitions for different tasks within a group.
+
+<EnterpriseAlert />
+
+In the following example:
+
+- The `template` block in the `web` task uses the default Consul cluster. It obtains a token that allows it access to the `engineering` admin partition in Consul. 
+- The `template` block in the `app` task uses the Consul cluster named `prod-apps`. It obtains a token that allows it access to the
+`engineering` Consul admin partition.
```
tgross, line 247: Hard-wrap these please.

tgross (changes requested):
It's not obvious to me that we need a whole example block just for this one field, but in any case that's the only thing in this PR we can leave in as the rest is just incorrect. Let's get those fixed up and we can get this merged.

website/content/docs/job-specification/consul.mdx
```diff
@@ -99,9 +100,7 @@ The [`template`][template] block can use the Consul token as well.
 - `partition` `(string: "")` - When this field is set, a constraint will be
   added to the group or task to ensure that the allocation is placed on a Nomad
   client that has a Consul Enterprise agent in the specified Consul [admin
-  partition][]. Note that Consul Community Edition agents are not assigned to
-  any admin partition, so this field should not be used without Consul
-  Enterprise.
+  partition][].
```
tgross, line 103: As I noted before, the admin partition is _cannot be fingerprinted_ when using Consul CE, so we should leave this note here. If you want to change the wording to specifically talk about fingerprinting that's fine, but if you set this field to `"default"` with Consul CE, none of your workloads will be placed!

website/content/docs/job-specification/consul.mdx
```diff
@@ -231,6 +230,56 @@ job "docs" {
...
+This example demonstrates how to configure Consul admin partitions for different
+tasks within a group. The Consul client agent must separately specify the admin
+partition in the agent configuration. Refer to the Consul documentation's
+[agent configuration reference][] for more information.
+
+<EnterpriseAlert />
+
+In the following example, the `template` block in the `web` and `app` tasks use
+the default Consul cluster. It obtains a token that allows it access to the `prod`
+admin partition in Consul.
```
tgross, line 244: This is kind of a weird example to have without further explanation, because you can't have two `task`-level Consul blocks in different admin partitions (because tasks are placed together in the same Allocation). I'd probably move the example `consul` block up to the `group` level and then have a note here that if you have `task`-level Consul blocks that they need to have the same partition.

tgross (changes requested):

tgross (approved):
LGTM!

PR #20026, opened 2024-02-22
Add a public parser which returns hcl.Diagnostics to make life easier for community tooling (LSP)

anthonymq:
Hello,
I'm building a Nomad LSP and It would be super useful for this kind of community tooling if you could expose a parser which returns hcl.Diagnostics.
Here is a PR that add this public method without breaking changes. 
Let me know if you need some changes.
Thanks

### DIFF
--- a/jobspec2/parse.go
+++ b/jobspec2/parse.go
@@ -43,9 +43,22 @@ func ParseWithConfig(args *ParseConfig) (*api.Job, error) {
 	args.normalize()
 
 	c := newJobConfig(args)
-	err := decode(c)
+	_, err := decode(c)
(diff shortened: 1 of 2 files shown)


jrasell (commented):
Hi @anthonymq and thanks for raising this PR. I've taken a quick initial pass through the changes and I think it would be better to have the `decode` function return a single object of `hcl.Diagnostics` rather than two objects such as `return diags, diags`. The callers can then handle whether to keep the original object type, or convert this to an error.

anthonymq:
Hi @jrasell, thanks for the response. I made the change. Makes sense.

PR #20029, opened 2024-02-22
func: add new picker dependency

Juanadelacuesta:
This PR introduces the new options for reconciling a reconnecting allocation and its replacement:

- Best score (Current implementation)
- Keep original
- Keep replacement
- Keep the one that has run the longest time

It is achieved by adding a new dependency to the allocReconciler that calls the corresponding function depending on the task group's disconnect strategy. For more detailed information, refer to the [RFC](https://go.hashi.co/rfc/nmd-188)

It resolves [15144](https://github.com/hashicorp/nomad/issues/15144)


scheduler/reconnecting_picker/reconnecting_picker.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (rp *ReconnectingPicker) pickLongestRunning(original *structs.Allocation, replacement *structs.Allocation) *structs.Allocation {
+	lt := original.GetLeaderTask()
+
+	// Default to the first task in the group if no leader is found.
+	if lt.Name == "" {
+		lt = *original.Job.LookupTaskGroup(original.TaskGroup).Tasks[0]
+	}
+
+	// If the replacement has a later start time, keep the original.
```
Juanadelacuesta, line 102: Technically it can happen that the LatestStartOfTask returns Zero, but it makes no sense in reality, should we address these cases?

nomad/structs/structs.go
```diff
@@ -11074,6 +11074,21 @@ func (a *Allocation) NextRescheduleTimeByTime(t time.Time) (time.Time, bool) {
...
 
+// ShouldBeReplaced tests an alloc for replace in case of disconnection
+func (a *Allocation) ShouldBeReplaced() *bool {
+	tg := a.Job.LookupTaskGroup(a.TaskGroup)
+
+	if tg == nil {
+		return nil
+	}
+
+	if tg.Disconnect == nil || tg.Disconnect.Replace == nil {
```
tgross, line 11085: We're returning `nil` on line 11089 anyways in this case, so we can skip a comparison:

```suggestion
	if tg.Disconnect == nil {
```

scheduler/reconnecting_picker/reconnecting_picker.go
tgross: I think you can hit that case if the allocation has started but the main task hasn't yet when we disconnect and reconnect. So for example, if there's a long-running pre-start task. So I think we need to account for that.

scheduler/reconcile_util.go
```diff
@@ -505,7 +505,14 @@ func updateByReschedulable(alloc *structs.Allocation, now time.Time, evalID stri
...
 	case isDisconnecting:
-		rescheduleTime, eligible = alloc.NextRescheduleTimeByTime(now)
+		replace := alloc.ShouldBeReplaced()
+		if replace != nil {
+			eligible = *replace
+			rescheduleTime = now
+		} else {
+			// Kept to maintain backwards compatibility with behavior prior to 1.8.0
+			rescheduleTime, eligible = alloc.NextRescheduleTimeByTime(now)
+		}
```
tgross, line 515: Rather than returning a `*bool` pointer from `ShouldBeReplaced`, I wonder if we could have it call `NextRescheduleTimeByTime` in the nil case instead? That'd avoid having the potential nil pointer we need to check. I also wonder if "ShouldBeReplaced" is a good name for that method... it only applies in the case of a disconnected alloc, right? 

So all that leads me to think we want something like:


```go
func (a *Allocation) NextRescheduleTimeOnDisconnect(now time.Time) (time.Time, bool) {
	if tg == nil || tg.Disconnect == nil || tg.Disconnect.Replace == nil {
		return a.NextRescheduleTimeByTime(now)
	}

	return now, *tg.Disconnect.Replace
}
```

nomad/structs/structs.go
```diff
@@ -11074,6 +11074,21 @@ func (a *Allocation) NextRescheduleTimeByTime(t time.Time) (time.Time, bool) {
 	return a.nextRescheduleTime(t, reschedulePolicy)
 }
 
+// ShouldBeReplaced tests an alloc for replace in case of disconnection
+func (a *Allocation) ShouldBeReplaced() *bool {
```
tgross, line 11078: See my comments in `reconcile_util.go` about this method signature.

nomad/structs/structs.go
```diff
@@ -11428,6 +11443,59 @@ func (a *Allocation) NeedsToReconnect() bool {
 	return disconnected
 }
 
+// GetLeaderTask will return the leader task in the allocation
+// if there is one, otherwise it will return an empty task.
+func (a *Allocation) GetLeaderTask() Task {
```
tgross, line 11448: By returning `Task` instead of `*Task` we end up doing a shallow-copy of the struct. Given that the caller is going to need to detect the "empty task" case anyways, maybe we should return a `*Task` here and let the caller grab the first task if it's nil?

nomad/structs/structs_test.go
```diff
@@ -5891,6 +5891,251 @@ func TestAllocation_NeedsToReconnect(t *testing.T) {
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			alloc := testAlloc.Copy()
+			alloc.TaskGroup = tc.taskGroup
+
+			alloc.Job.TaskGroups[0].Disconnect = tc.disconnectGroup
+
+			got := alloc.ShouldBeReplaced()
+
+			require.Equal(t, tc.expected, got)
```
tgross, line 5941:
```suggestion
			must.Eq(t, tc.expected, got)
```

nomad/structs/structs.go
```diff
@@ -11428,6 +11443,59 @@ func (a *Allocation) NeedsToReconnect() bool {
...
+		}
+	}
+
+	return task
+}
+
+// LatestStartOfTask returns the time of the last start event for the given task
+// using the allocations TaskStates. If the task has not started, the zero time
+// will be returned.
+func (a *Allocation) LatestStartOfTask(taskName string) time.Time {
```
lgfa29, line 11479:
```suggestion
// LastTaskStart returns the time of the last start event for the given task
// using the allocations TaskStates. If the task has not started, the zero time
// will be returned.
func (a *Allocation) LastTaskStart(taskName string) time.Time {
```

Another nit-pick 😅 

This name matches the name of `LastUnknown`, which returns a similar result. It could also be useful to move the method so it's closer to that one.
https://github.com/hashicorp/nomad/blob/3193ac204f6564711004e00948e43146ce1399c4/nomad/structs/structs.go#L11394-L11409

nomad/structs/structs_test.go
```diff
@@ -5891,6 +5891,251 @@ func TestAllocation_NeedsToReconnect(t *testing.T) {
...
+			disconnectGroup: &DisconnectStrategy{
+				Replace: pointer.Of(true),
+			},
+			expected: pointer.Of(true),
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			alloc := testAlloc.Copy()
```
lgfa29, line 5934:
```suggestion
			alloc := MockAlloc()
```

This probably works? It doesn't seem like we're using `testAlloc` for much.

scheduler/reconcile.go
```diff
@@ -32,6 +33,10 @@ const (
 	rescheduleWindowSize = 1 * time.Second
 )
 
+type ReconnectingPicker interface {
+	PickReconnectingAlloc(disconnect *structs.DisconnectStrategy, original *structs.Allocation, replacement *structs.Allocation) *structs.Allocation
+}
```
lgfa29, line 38: Why do we need this interface? `reconnectingpicker.ReconnectingPicker` seems to be the only implementation.

I could see a case where the drain strategy could be an interface with one implementation per strategy, but `ReconnectingPicker` seems to be handling all of them already.

scheduler/reconnecting_picker/reconnecting_picker.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (rp *ReconnectingPicker) PickReconnectingAlloc(ds *structs.DisconnectStrategy, original *structs.Allocation, replacement *structs.Allocation) *structs.Allocation {
+	// Check if the replacement is newer.
+	// Always prefer the replacement if true.
+	replacementIsNewer := replacement.Job.Version > original.Job.Version ||
+		replacement.Job.CreateIndex > original.Job.CreateIndex
+	if replacementIsNewer {
+		rp.logger.Debug("replacement has a newer version, keeping replacement")
+		return replacement
+	}
```
lgfa29, line 31:
```suggestion
	// Check if the replacement is for a newer job version.
	// Always prefer the replacement if true.
	replacementIsNewer := replacement.Job.Version > original.Job.Version ||
		replacement.Job.CreateIndex > original.Job.CreateIndex
	if replacementIsNewer {
		rp.logger.Debug("replacement has a newer job version, keeping replacement")
		return replacement
	}
```

I think the replacement itself would always be newer right? 😅 

So just making this a bit more explicit.

scheduler/reconnecting_picker/reconnecting_picker_test.go
```diff
@@ -0,0 +1,154 @@
...
+	replacement := &structs.Allocation{
+		Job: &structs.Job{
+			Version:     2,
+			CreateIndex: 2,
+		},
+	}
+	result := rp.PickReconnectingAlloc(ds, original, replacement)
+	if result != replacement {
+		t.Fatalf("expected replacement, got %v", result)
+	}
```
lgfa29, line 35:
```suggestion
	must.Eq(t, replacement, result)
```

I think this works.

scheduler/reconnecting_picker/reconnecting_picker_test.go
```diff
@@ -0,0 +1,154 @@
...
+			Version:     1,
+			CreateIndex: 1,
+		},
+	}
+	replacement := &structs.Allocation{
+		Job: &structs.Job{
+			Version:     2,
+			CreateIndex: 2,
+		},
+	}
```
lgfa29, line 31: It may be useful to matrix test each field being modified independently. Testing for a change in both can hide bugs in one of the two checks.

nomad/structs/structs.go
Juanadelacuesta: I don't like returning pointers unless nil has a meaning, because you leave to the caller the task of checking for nil before using the result

nomad/structs/structs_test.go
Juanadelacuesta: I blame copilot!! jejeje

scheduler/reconnecting_picker/reconnecting_picker.go
```diff
@@ -0,0 +1,108 @@
...
+	if lt.Name == "" {
+		lt = *original.Job.LookupTaskGroup(original.TaskGroup).Tasks[0]
+	}
+
+	// If the replacement has a later start time, keep the original.
+	if original.LatestStartOfTask(lt.Name).Sub(replacement.LatestStartOfTask(lt.Name)) < 0 {
+		return original
+	}
+
+	return replacement
```
lgfa29, line 107: There are a few subtleties here that worry me a bit.

1. Task states are ephemeral, I think we only keep like 10 around? So for a long-living task I would expect the start event to have been long gone, specially if it uses features like `template` where each change will generate a handful of events. So in a more realistic scenario we would end up with `original.LatestStartOfTask` and `replacement.LatestStartOfTask` being both zero 😬 
2. This is only checking the leader task, or the first one, so it's not taking into consideration aspects such as task [lifecycle](https://developer.hashicorp.com/nomad/docs/job-specification/lifecycle). If the first task is a `prestart` sidecar, for example, I think this would give unexpected results.

I think a better approach would be use one of `LastRestart` or `StartedAt` (or probably a combination of both) from the alloc's `TaskState`?
https://github.com/hashicorp/nomad/blob/3193ac204f6564711004e00948e43146ce1399c4/nomad/structs/structs.go#L8816-L8826

We would also need to compare all tasks (or maybe at lest filter by `lifecycle`) so that we get a more general sense of what to consider "longest running" allocation.

lgfa29 (commented):
Exciting stuff! The reconnect logic has always been confusing to me from an user perspective 😅 

May main concern here is the `pickLongestRunning` logic. We may need to take a different approach to make this decision.

scheduler/reconcile_util.go
Juanadelacuesta: It looks cleaner for sure

scheduler/reconcile.go
Juanadelacuesta: Im aiming at removing the responsibility of the picking from the allocReconciler, that is the main reason of te interface :)

scheduler/reconnecting_picker/reconnecting_picker.go
Juanadelacuesta: Ok, let me add a new filter to use a task from the main cycle :)

scheduler/reconnecting_picker/reconnecting_picker_test.go
Juanadelacuesta: Again, I blame coPilot! jejej

scheduler/reconcile.go
lgfa29: Interfaces in Go are a bit strange, and work differently than in other languages, like Java or C#.

Since they're implicitly defined pre-creating them can often lead to a confusing implementation. Specially for people new to the code base, trying to "jump to definition" in an interface is quite frustrating 😬 

Sometimes that's unavoidable because we want to have multiple implementations (like a real thing and a test implementation), or if want to restrict what to expose (like only exposing certain methods of a struct).

But in this case I can't really think of a different implementations of `ReconnectingPicker`. The entire decision is already encapsulated in the `reconnectingpicker.ReconnectingPicker` implementation (the fact they have the same name kind of point to that as well 😅).

So we don't need this interface to remove responsibility from the reconciler. Using the concrete struct already accomplishes that. And having it creates unnecessary indirection that makes the code harder to follow.

If, later on, we do find a need for an interface, the implicit nature of Go's interface can help. If the reconciler ends up calling X(), Y(), and Z(), where each behaves differently under some conditions, we can easily add them to a new interface without much code change.

scheduler/reconnecting_picker/reconnecting_picker.go
lgfa29: Checking for main is good, but I think the biggest issue is relying on a specific task event. We can't really do that since we only store a limited number of them.

I think we need to rethink how we define longest running. The `TaskState` I mention could be helpful.

nomad/structs/structs.go
```diff
@@ -11428,6 +11438,60 @@ func (a *Allocation) NeedsToReconnect() bool {
 	return disconnected
 }
 
+// LeaderOrMainTaskInInGroup will return the leader task in the allocation
+// if there is one, otherwise it will return the first task that run as main.
+// If the task group is no longer present or there are no tasks in it, it
+// will return nil.
+func (a *Allocation) LeaderOrMainTaskInInGroup(tg *TaskGroup) *Task {
```
tgross, line 11445:
```suggestion
func (a *Allocation) LeaderOrMainTaskInGroup(tg *TaskGroup) *Task {
```

nomad/structs/structs_test.go
```diff
@@ -5891,6 +5891,249 @@ func TestAllocation_NeedsToReconnect(t *testing.T) {
...
+			taskGroup:    "missing-task-group",
+			expected:     false,
+			expectedTime: time.Time{},
+		},
+		{
+			name:            "missing_disconnect_group",
+			taskGroup:       "web",
+			disconnectGroup: nil,
+			expected:        true,
+			expectedTime:    testNow.Add(RestartPolicyMinInterval), // RestartPolicyMinInterval is de default value
```
tgross, line 5918:
```suggestion
			expectedTime:    testNow.Add(RestartPolicyMinInterval), // RestartPolicyMinInterval is the default value
```

(and copied in the case below)

tgross (approved):
LGTM! Once that discussion with @lgfa29 about the interface is resolved this should be good-to-go.

scheduler/reconcile.go
Juanadelacuesta: Having that interface in place gives us a define frontier between the alloc reconciler and the reconnecting picker: Imagine we refactor the reconciler and we want to unit test the logic. If we remove the interface, we remove the possibility of having a look at the internals of the reconciliation, we will need to write tests that also take into account the picker logic, no way to isolate components. If we leave the interface where it is, it is the perfect cutting point where we can add a test implementation of the picker just for the purpose  of verifying the behaviour of the reconciliation.

scheduler/reconnecting_picker/reconnecting_picker.go
Juanadelacuesta: Reimplemented! Now it uses the `StartAt`, `Restarts` and `LastRestart` fields.

scheduler/reconcile.go
Juanadelacuesta: It is meant to isolate the logic of the reconciler, not of the picker

scheduler/reconnecting_picker/reconnecting_picker.go
```diff
@@ -0,0 +1,125 @@
...
+
+	if orgStartTime.IsZero() && !repStartTime.IsZero() {
+		return replacement
+	}
+
+	if !orgStartTime.IsZero() && repStartTime.IsZero() {
+		return original
+	}
+
+	// In neither one of them is running yet, default to best score.
```
lgfa29, line 114:
```suggestion
	// If neither one of them is running yet, default to best score.
```

scheduler/reconnecting_picker/reconnecting_picker.go
```diff
@@ -0,0 +1,125 @@
...
+	if repStartTime.IsZero() && orgStartTime.IsZero() {
+		return rp.pickBestScore(original, replacement)
+	}
+
+	// If the replacement has a later start time, keep the original.
+	if orgStartTime.Sub(repStartTime) < 0 {
```
lgfa29, line 120:
```suggestion
	if orgStartTime.Before(repStartTime) {
```

nomad/structs/structs.go
```diff
@@ -11428,6 +11438,84 @@ func (a *Allocation) NeedsToReconnect() bool {
...
+			continue
+		}
+
+		if t.IsMain() {
+			mainTasks = append(mainTasks, t)
+		}
+	}
+
+	return leaderTasks, mainTasks
+}
```
lgfa29, line 11476: Given that this method is only used in the reconciler (a part of our code that is called _very_ often), it may be better to avoid allocating the slices and have the loop in `StartOfOldestTask`. Then we only need to keep the oldest time stored. 

Some untested sample code:

```go
func (a *Allocation) StartOfOldestTask(tg *TaskGroup) time.Time {
	now := time.Now().UTC()
	oldestStart := now

	for _, t := range tg.Tasks {
		if t.Leader {
			return a.LastStartOfTask(t.Name)
		}

		if t.IsMain() {
			ls := a.LastStartOfTask(t.Name)
			if !ls.IsZero() && ls.Before(oldestStart) {
				oldestStart = ls
			}
		}
	}

	return oldestStart
}
```

Since a group can only have a one leader, this also has the added benefit of cutting the loop short as soon as a leader is found.
https://github.com/hashicorp/nomad/blob/428103ba1265f519955d92817cdebeb673b1e58b/nomad/structs/structs.go#L6984-L6986

nomad/structs/structs.go
```diff
@@ -11428,6 +11438,84 @@ func (a *Allocation) NeedsToReconnect() bool {
...
+	if len(tasks) == 0 {
+		return time.Time{}
+	}
+
+	now := time.Now().UTC()
+	oldestStart := now
+	for _, task := range tasks {
+		ls := a.LastStartOfTask(task.Name)
+		if !ls.IsZero() && ls.Before(oldestStart) {
+			oldestStart = a.LastStartOfTask(task.Name)
```
lgfa29, line 11508:
```suggestion
			oldestStart = ls
```

We don't need to call the function again.

nomad/structs/structs.go
```diff
@@ -11428,6 +11438,84 @@ func (a *Allocation) NeedsToReconnect() bool {
...
+		}
+	}
+
+	if oldestStart == now {
+		return a.LastStartOfTask(tasks[0].Name)
+	}
```
lgfa29, line 11514: Why do we return the first one in this case? If no task has started yet I would imagine `StartOfOldestTask` to return 0 instead of a somewhat random value? 🤔

Juanadelacuesta: Zero is the oldest date possible. I needed a default value and that one seemed as good as any other.

scheduler/reconnecting_picker/reconnecting_picker.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/hashicorp/nomad/nomad/structs"
+)
+
+type ReconnectingPicker struct {
+	logger log.Logger
+}
+
+func New(logger log.Logger) *ReconnectingPicker {
+	rp := ReconnectingPicker{
+		logger: log.L().Named("reconnecting-picker"),
```
lgfa29, line 19:
```suggestion
		logger: logger.Named("reconnecting-picker"),
```

I think that, if we use the global default logger, we won't have the log configurations applied. Since the reconciler is already providing a logger here we should it instead.

lgfa29 (approved):
Some small last comments

PR #20047, opened 2024-02-28
[ui] Fixes an issue where shift+num would not open an eval on the evaluations index table

philrenaud:
Because we have multiple links on the Evaluations page table (to open the eval in a sidebar, and also to open the job to which the eval is related), we used a keyboard event "keyup" binding to let a user do something like tab + enter or tab + space to open the sidebar.

This predated the general keyboard navigation convention in the Nomad UI of shift+01, shift+02 to open a list row item action. The "which key was pressed" guard on opening the sidebar prevents this sidebar from opening with shift+n, and this PR fixes that by passing in an explicit "keynav" string on command.

image

### DIFF
--- a/.changelog/20047.txt
+++ b/.changelog/20047.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fixed an issue where keynav would not trigger evaluation sidebar expand
+```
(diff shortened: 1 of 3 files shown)


pkazmierczak (approved):
LGTM!

PR #20050, opened 2024-02-29
docs: fix autoscaler variable ACL policy example.

jrasell:
Not sure how I messed up copying it from my lab that badly.


lgfa29 (approved):
Oops...missed that as well, thanks!

PR #20069, opened 2024-03-05
exec2: implement dynamic workload users taskrunner hook

shoenig:
This PR impelements a TR hook for allocating dynamic workload users from
a pool managed by the Nomad client. This adds a new task driver Capability,
DynamicWorkloadUsers - which a task driver must indicate in order to make
use of this feature.

The client config plumbing is coming in a followup PR - in the RFC we
realized having a client.users block would be nice to have, with some
additional unrelated options being moved from the deprecated client.options
config.

### DIFF
--- a/helper/users/dynamic/users.go
+++ b/helper/users/dynamic/users.go
@@ -1,0 +1,61 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package dynamic
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+
+	"github.com/hashicorp/nomad/helper/users"
+)
+
+const (
+	// Home is the non-existent directory path to associate with dynamic
+	// workload users. Any operation on this path should cause an error.
+	//
+	// The path '/nonexistent' is consistent with what systemd uses for
+	// non-interactive service users.
+	Home = "/nonexistent"
+)
+
+// String creates a pseudo username encoding the given ugid, in the form
+// 'nomad-<id>'.
+func String(ugid UGID) string {
+	return fmt.Sprintf("nomad-%d", ugid)
+}
+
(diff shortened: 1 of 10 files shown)


plugins/drivers/driver.go
```diff
@@ -178,6 +178,12 @@ type Capabilities struct {
 	// DisableLogCollection indicates this driver has disabled log collection
 	// and the client should not start a logmon process.
 	DisableLogCollection bool
+
+	// DynamicWorkloadUsers indicates this driver is capable (but not required)
+	// of making use of UID/GID not backed by a user known to the operating system.
+	// The allocation of a unique, not-in-use UID/GID is managed by Nomad client
+	// ensuring no overlap.
+	DynamicWorkloadUsers bool
```
tgross, line 186: I think this needs to get added to the `DriverCapabilities` protobuf message as well in the `plugins/drivers/proto/driver.proto`

helper/users/dynamic/users.go
```diff
@@ -0,0 +1,61 @@
...
+
+	return UGID(i), err
+}
+
+// LookupUser will return the UID, GID, and home directory associated with the
+// given username. If username is of the form 'nomad-<id>' this indicates Nomad
+// has synthesized a dynamic workload user for the task and the UID/GID are the
+// <id> value.
+func LookupUser(username string) (int, int, string, error) {
+	// if we can succusfully parse username as an anonymous user, use that
```
tgross, line 53:
```suggestion
	// if we can successfully parse username as an anonymous user, use that
```

tgross (approved):
LGTM!

plugins/drivers/driver.go
shoenig: Yup will do in a followup; just getting the hook hooked up for now.

Issue #19833, opened 2024-01-26
Remove BUSL-1.1 imports from MPL-2.0 packages


lgfa29:
Some of the packages in this repo remained licensed under MPL-2.0, but are importing code from packages licensed under BUSL-1.1. https://github.com/hashicorp/nomad/pull/19832 fixed some of them, but there are still others remaining, mostly under `drivers/shared` and `plugins`. 

The PR also introduced a new Semgrep rule to detect them. This is the current report of remaining incorect imports.

```
$ semgrep scan --config .semgrep/mpl_busl.yml


┌─────────────┐
│ Scan Status │
└─────────────┘
  Scanning 5599 files (only git-tracked) with 1 Code rule:

  CODE RULES
  Scanning 4875 files.

  SUPPLY CHAIN RULES

  No rules to run.


  PROGRESS

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00


┌──────────────────┐
│ 49 Code Findings │
└──────────────────┘

    drivers/shared/eventer/eventer.go
   ❯❯❱ semgrep.mpl_busl
          BUSL package `github.com/hashicorp/nomad/helper` imported in MPL package `eventer`

           12┆ "github.com/hashicorp/nomad/helper"

    drivers/shared/executor/executor.go
   ❯❯❱ semgrep.mpl_busl
          BUSL package `github.com/hashicorp/nomad/client/allocdir` imported in MPL package `executor`

           23┆ "github.com/hashicorp/nomad/client/allocdir"
            ⋮┆----------------------------------------
   ❯❯❱ semgrep.mpl_busl
          BUSL package `github.com/hashicorp/nomad/client/lib/cgroupslib` imported in MPL package `executor`

           24┆ "github.com/hashicorp/nomad/client/lib/cgroupslib"
            ⋮┆----------------------------------------
   ❯❯❱ semgrep.mpl_busl
          BUSL package `github.com/hashicorp/nomad/client/lib/cpustats` imported in MPL package `executor`

           25┆ "github.com/hashicorp/nomad/client/lib/cpustats"
            ⋮┆----------------------------------------
   ❯❯❱ semgrep.mpl_busl
          BUSL package `github.com/hashicorp/nomad/client/lib/fifo` imported in MPL package `executor`

           26┆ "github.com/hashicorp/nomad/client/lib/fifo"
            ⋮┆----------------------------------------
[...]

PR #20073, opened 2024-03-05
Adds ability to restrict uid and gids in exec and raw_exec

mikenomitch:
Adds ability to restrict host uid and gids in exec and raw_exec.

To Test:

Add the following to agent config:

```
plugin "exec" {
  enabled = true
  config {
    denied_host_uids = "0-65534"
    denied_host_gids = ""
  }
}

plugin "raw_exec" {
  config {
    enabled = true
    denied_host_uids = "1,2-9"
    denied_host_gids = "0-100"
  }
}
```

Then in raw_exec or exec tasks change the "user" value to become a user in any of these ranges. Note that you should see an error like the following:
Screenshot 2024-03-05 at 10 53 06 AM

It should also error on job submit if you give it bad ranges. IE "0,1-foo"

Note: This is only *needed* on raw_exec, but since it felt like the code was 90% reusable and would be appreciated in exec too, I figured I'd add it (at the risk of a bit of scope creep). It also felt like I'd set us up better to add this to exec_v2 by just adding this in a shared location.


drivers/exec/driver.go
```diff
@@ -688,3 +709,12 @@ func (d *Driver) ExecTaskStreamingRaw(ctx context.Context,
 
 	return handle.exec.ExecStreaming(ctx, command, tty, stream)
 }
+
+func getUsername(cfg *drivers.TaskConfig) string {
```
mikenomitch, line 713: Pulled this to a shared fn just so we we'd have a single source of truth

drivers/shared/validators/validators.go
```diff
@@ -0,0 +1,154 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package validators
```
mikenomitch, line 4: I feel like pulling this stuff into a new shared package was probably the biggest "design" choice I made in this. There didn't feel like another obvious place to put shared driver validation code, but perhaps I missed something. (also, not the point of this package, but it could probably be more broadly useful)

shoenig:
@mikenomitch looks like the linter is unhappy

you should be able to run `make check` yourself locally and see the same thing

```
==> Linting source code...
Error: drivers/shared/validators/validators.go:73:29: unnecessary conversion (unconvert)
		gids = append(gids, uint64(u))
```

drivers/shared/validators/validators_test.go
```diff
@@ -0,0 +1,110 @@
...
+			err := UserInRange(getUserFn, "username", tc.uidRanges, tc.gidRanges)
+			if tc.expectedPass {
+				must.NoError(t, err)
+			} else {
+				if err == nil {
+					t.Errorf("expected error, got nil")
+				} else {
+					must.StrContains(t, err.Error(), tc.expectedErr)
+				}
+			}
```
tgross, line 107: You cut down the test verbosity here by dropped the `tc.expectedPass`, checking `tc.expectedErr == ""` instead, and then using `must.ErrContains` for the error case.

(Apply this same comment across several other tests in this PR.)

drivers/shared/validators/validators.go
```diff
@@ -0,0 +1,154 @@
...
+	uid, err := strconv.ParseUint(u.Uid, 10, 32)
+	if err != nil {
+		return fmt.Errorf("unable to convert userid %s to integer", u.Uid)
+	}
+
+	// check uids
+
+	uidRanges, err := parseRanges(deniedHostUIDs)
+	if err != nil {
+		return fmt.Errorf("invalid denied_host_uids value %q: %v", deniedHostUIDs, err)
```
tgross, line 50: Wrap your errors wherever you can:

```suggestion
		return fmt.Errorf("invalid denied_host_uids value %q: %w", deniedHostUIDs, err)
```

drivers/rawexec/driver_test.go
```diff
@@ -123,6 +123,15 @@ func TestRawExecDriver_SetConfig(t *testing.T) {
 	bconfig.PluginConfig = data
 	require.NoError(harness.SetConfig(bconfig))
 	require.Exactly(config, d.(*Driver).config)
+
+	// Turns on uid/gid restrictions, and sets the range to a bad value
+	config.DeniedHostUids = "10-0"
+	data = []byte{}
+	require.NoError(basePlug.MsgPackEncode(&data, config))
```
tgross, line 130: We're trying to get everything of testify

```suggestion
	must.NoError(basePlug.MsgPackEncode(&data, config))
```

(Apply to the rest of the tests in the PR as needed)

drivers/shared/validators/validators.go
```diff
@@ -0,0 +1,154 @@
...
+		return fmt.Errorf("failed to identify user %q: %v", usernameToLookup, err)
+	}
+	uid, err := strconv.ParseUint(u.Uid, 10, 32)
+	if err != nil {
+		return fmt.Errorf("unable to convert userid %s to integer", u.Uid)
+	}
+
+	// check uids
+
+	uidRanges, err := parseRanges(deniedHostUIDs)
```
tgross, line 48: I'm pretty sure we already do this earlier when we read the plugin's configuration? So we should be able to pass the already-parsed `IDRange` object into this function so that we only do it once at plugin startup, instead of on every task start.

drivers/shared/validators/validators.go
```diff
@@ -0,0 +1,154 @@
...
+
+		idRange.Lower = singleBoundInt
+		idRange.Upper = singleBoundInt
+	case 2:
+		boundAStr := uidDenyRangeParts[0]
+		boundBStr := uidDenyRangeParts[1]
+
+		boundAInt, err := strconv.ParseUint(boundAStr, 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("invalid range %q, invalid bound: \"%q\" ", boundsString, boundAStr)
```
tgross, line 137: A few items with this error message:
* The `%q` quotes it already.
* The caller already includes the `boundString` 
* We should wrap the inner error here so that it's clear why it's invalid. 
 
So we should do something like:

```suggestion
			return nil, fmt.Errorf("%q is an invalid bound: %w", boundAStr, err)
```

drivers/shared/validators/validators.go
```diff
@@ -0,0 +1,154 @@
...
+		singleBoundInt, err := strconv.ParseUint(singleBound, 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("range bound not valid, invalid bound: \"%q\" ", singleBoundInt)
+		}
+
+		idRange.Lower = singleBoundInt
+		idRange.Upper = singleBoundInt
+	case 2:
+		boundAStr := uidDenyRangeParts[0]
+		boundBStr := uidDenyRangeParts[1]
```
tgross, line 133: Nitpick: maybe call these lower/upper rather than a/b?

drivers/shared/validators/validators.go
tgross, line 4: The test is `//go:build !windows`, should this file be as well?

drivers/shared/validators/validators.go
```diff
@@ -0,0 +1,154 @@
...
+	uidDenyRangeParts := strings.Split(boundsString, "-")
+
+	var idRange IDRange
+
+	switch len(uidDenyRangeParts) {
+	case 0:
+		return nil, fmt.Errorf("range cannot be empty, invalid range: \"%q\" ", boundsString)
+	case 1:
+		singleBound := uidDenyRangeParts[0]
+		singleBoundInt, err := strconv.ParseUint(singleBound, 10, 64)
```
tgross, line 124: Possibly nitpicky: isn't the UID the same as the GID? In the GID parser we're assuming the value is within 32 bit because of the maximum Unix GID size. The two parsers should probably match.

website/content/docs/drivers/exec.mdx
```diff
@@ -178,16 +178,40 @@ able to make use of IPC features, like sending unexpected POSIX signals.
...
+by the operating system.
 
 !> **Warning:** Allowing more capabilities beyond the default may lead to
 undesirable consequences, including untrusted tasks being able to compromise the
 host system.
 
+- `denied_host_uids` - (Optional) Specifies a comma-separated list of host uids to
+  deny. Ranges can be specified by using a hyphen separating the two inclusive ends.
+  If a "user" value is specified in task configuration and that user has a user id in
+  the given ranges, the task will error before starting. This will not be run on Windows
```
tgross, line 194: "run" feels ambiguous, maybe "checked"?

```suggestion
  the given ranges, the task will error before starting. This will not be checked on Windows
```

drivers/rawexec/driver_unix_test.go
```diff
@@ -410,3 +412,33 @@ func TestRawExec_ExecTaskStreaming_User(t *testing.T) {
 	require.Empty(t, stderr)
 	require.Contains(t, stdout, "nobody")
 }
+
+func TestRawExec_Validate(t *testing.T) {
+	ci.Parallel(t)
+
+	current, err := users.Current()
+	require.NoError(t, err)
```
tgross, line 420: (Apply across the PR)

```suggestion
	must.NoError(t, err)
```

tgross (changes requested):
This is looking pretty good @mikenomitch, but I've left some comments we should address.

Also:
* We need a changelog entry
* Note that the `test-e2e-vault` test failure isn't a flake, but a appears to be bug introduced by this PR:

> Error:     2024-03-06T00:13:33.371Z [ERROR] client.alloc_runner.task_runner: running driver failed: alloc_id=ca431f74-e99e-8749-dc0e-62de3f1b96e9 task=cat error="failed driver config validation: failed to identify user \"\": user: unknown user "

drivers/exec/driver.go
```diff
@@ -227,6 +245,12 @@ func (tc *TaskConfig) validate() error {
 		return fmt.Errorf("cap_drop configured with capabilities not supported by system: %s", badDrops)
 	}
 
+	usernameToLookup := getUsername(cfg)
+
+	if err := validators.UserInRange(users.Lookup, usernameToLookup, driverConfig.DeniedHostUids, driverConfig.DeniedHostGids); err != nil {
+		return err
+	}
+
```
shoenig, line 253: I don't think this is the right place for this validation - we've now added 2 new arguments to pass in on this `TaskConfig.validate()` method, and then ignored the `tc` itself - suggesting we should leave this method alone and do the assertion elsewhere.

drivers/exec/driver_test.go
```diff
@@ -870,10 +871,59 @@ func TestDriver_Config_validate(t *testing.T) {
...
+				DefaultModeIPC: "private",
+				DeniedHostUids: tc.uidRanges,
+				DeniedHostGids: tc.gidRanges,
+			}).validate()
+
+			if tc.errorStr == nil {
+				require.Nil(t, validationErr)
+			} else {
+				require.Contains(t, validationErr.Error(), *tc.errorStr)
+			}
```
shoenig, line 901: switch to `shoenig/test` instead of `testify`

`must.NoError(t, err)`
`must.ErrorContains(t, err, "oops")`

drivers/exec/driver_test.go
```diff
@@ -870,10 +871,59 @@ func TestDriver_Config_validate(t *testing.T) {
...
 	ci.Parallel(t)
+
+	current, err := users.Current()
+	require.NoError(t, err)
+	currentUid, err := strconv.ParseUint(current.Uid, 10, 32)
+	require.NoError(t, err)
+	nobody, err := users.Lookup("nobody")
+	require.NoError(t, err)
+	nobodyUid, err := strconv.ParseUint(nobody.Uid, 10, 32)
+	require.NoError(t, err)
```
shoenig, line 916: you can use `LookupUser` to make this less terrible - (exec is linux only it's fine)

https://github.com/hashicorp/nomad/blob/main/helper/users/lookup.go#L27-L31

drivers/exec/driver_test.go
```diff
@@ -926,7 +976,25 @@ func TestDriver_TaskConfig_validate(t *testing.T) {
...
+			config       Config
+			driverConfig drivers.TaskConfig
+			exp          error
+		}{
+			{config: configAllowCurrent, driverConfig: driverConfigSpecifyCurrent, exp: nil},
+			{config: configDenyCurrent, driverConfig: driverConfigNoUserSpecified, exp: nil},
+			{config: configDenyCurrent, driverConfig: driverConfigSpecifyCurrent, exp: errors.New(currentUserErrStr)},
+			{config: configDenyAnonymous, driverConfig: driverConfigNoUserSpecified, exp: errors.New(anonUserErrStr)},
+		} {
+			require.Equal(t, tc.exp, (&TaskConfig{}).validate(&tc.driverConfig, &tc.config))
```
shoenig, line 997: `test` not `testify`

drivers/rawexec/driver_windows.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"github.com/hashicorp/nomad/plugins/drivers"
+)
+
+func (tc *TaskConfig) Validate(driverCofig Config, cfg drivers.TaskConfig) error {
+	// This is a noop on windows since the uid and gid cannot be checked against a range easily
+	// We could eventually extend this functionality to check for individual users IDs strings
+	// but that is not currently supported. See driverValidators.UserInRange for
+	// unix logic
```
shoenig, line 16: yesssss @mikenomitch already on board with punting on all things windows

drivers/shared/validators/validators.go
shoenig: I think in most cases any validation is driver specific. Since this is shared between exec and raw_exec, seems like a reasonably choice to put it here.

drivers/shared/validators/validators.go
```diff
@@ -0,0 +1,154 @@
...
+type IDRange struct {
+	Lower uint64 `codec:"from"`
+	Upper uint64 `codec:"to"`
+}
+
+// IDRangeValid is used to ensure that the configuration for ID ranges is valid.
+func IDRangeValid(rangeType string, deniedRanges string) error {
+	_, err := parseRanges(deniedRanges)
+	if err != nil {
+		return fmt.Errorf("invalid %s value %q: %v", rangeType, deniedRanges, err)
```
shoenig, line 23: use `%w` to wrap the error

drivers/shared/validators/validators.go
```diff
@@ -0,0 +1,154 @@
...
+			return nil, err
+		}
+
+		idRanges = append(idRanges, *idRange)
+	}
+
+	return idRanges, nil
+}
+
+func parseRangeString(boundsString string) (*IDRange, error) {
```
shoenig, line 114: I think we can replace all of this parsing with the existing generic cpuset parser

https://github.com/hashicorp/nomad/blob/main/client/lib/idset/idset.go#L72-L75

drivers/shared/validators/validators_test.go
```diff
@@ -0,0 +1,110 @@
...
+	}{
+		{name: "standard-range-is-valid", idRange: validRange, expectedPass: true},
+		{name: "same-number-for-both-bounds-is-valid", idRange: validRangeSingle, expectedPass: true},
+		{name: "lower-higher-than-upper-is-invalid", idRange: invalidRangeFlipped, expectedPass: false, expectedErr: flippedBoundsMessage},
+		{name: "missing-lower-is-invalid", idRange: invalidRangeSubstring, expectedPass: false, expectedErr: invalidBound},
+		{name: "missing-higher-is-invalid", idRange: invalidRangeEmpty, expectedPass: false, expectedErr: invalidBound},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
```
shoenig, line 39: will be unnecessary as soon as we get 1.22 (should be today)

drivers/shared/validators/validators_test.go
```diff
@@ -0,0 +1,110 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			err := IDRangeValid("uid", tc.idRange)
+			if tc.expectedPass {
+				must.NoError(t, err)
+			} else {
+				if err == nil {
+					t.Errorf("expected error, got nil")
+				} else {
+					must.StrContains(t, err.Error(), tc.expectedErr)
+				}
```
shoenig, line 49: `must.ErrorContains(t, err, "blah")`

drivers/exec/driver.go
mikenomitch: Yeah the additional args felt bad to me, but this does feel like a validation step 🤔 - I'll see what I can do to make it nicer

drivers/shared/validators/validators.go
mikenomitch: This is used by exec/driver.go which does not have !windows in it, so I think it gets tricky.

I could wrap the use in a function in exec/driver.go like I do with raw_exec and then move that function out to exec/driver_unix.go? And then this could safely be a 100% non-windows file.

Hmm... I'm also debating moving the os-level user getting bits out of this file wholesale, which might change the dynamics of this a bit. In which case, I can leave this as a general validator, then just make the test a _unix test.

drivers/shared/validators/validators.go
mikenomitch: We do do this twice, but it is happening on different machines. I'm pretty sure the initial validate is happening by the server when the job is submitted, and then this bit is happening when the task is starting on the client (and needs to happen there cause we only have the user info then). We could parse the jobspec and then save the parsed IdRanges in state somewhere and pass those down to the clients, but I figured that was not worth it versus just running the parsing again. Thoughts?

tgross: The server can't validate anything about the driver (either the driver's config or anything in the `task.config` block). Generally speaking, nothing in the driver code ever happens on the server -- not being able to validate driver config on the server is a long-standing user complaint, in fact.

drivers/shared/validators/validators.go
mikenomitch: Yeah, I also thought about reusing some of the port range code, but that was also returning all allowable port numbers, so it just didn't feel worth it for reuse.

website/content/docs/drivers/exec.mdx
```diff
@@ -178,16 +178,40 @@ able to make use of IPC features, like sending unexpected POSIX signals.
...
-  which is modeled after the capabilities allowed by [docker by default][docker_caps]
-  (without [`NET_RAW`][no_net_raw]). Allows the operator to control which capabilities
-  can be obtained by tasks using [`cap_add`][cap_add] and [`cap_drop`][cap_drop] options.
-  Supports the value `"all"` as a shortcut for allow-listing all capabilities supported
-  by the operating system.
+which is modeled after the capabilities allowed by [docker by default][docker_caps]
+(without [`NET_RAW`][no_net_raw]). Allows the operator to control which capabilities
+can be obtained by tasks using [`cap_add`][cap_add] and [`cap_drop`][cap_drop] options.
+Supports the value `"all"` as a shortcut for allow-listing all capabilities supported
+by the operating system.
```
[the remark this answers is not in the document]
mikenomitch: Auto-formatter kicked in and I rolled with it. I can revert

drivers/shared/validators/validators.go
mikenomitch: Oh blegh, sorry I'm being dumb! - Lemme see if I can thread this thru

mikenomitch: Oh man, I was being dense and (briefly) thinking this was on job submit, but its really on driver config load...

drivers/exec/driver.go
```diff
@@ -181,6 +195,23 @@ func (c *Config) validate() error {
 	return nil
 }
 
+func (c *Config) setDeniedIds() error {
```
mikenomitch, line 198: @tgross calling this out as this is the new single-spot to do the initial parsing of the strings and then those get converted to IDRanges which are stuck on the config.

Code definitely feels cleaner as a result thanks.

The Docker driver had a similar pattern with some config attributes, but I'm not positive there's not a nicer way than having two similar attrs on the struct (see above).

mikenomitch:
Update:

Most of the in-code comments have been resolved and I've done some light refactoring outside of that too. Feels like its in a much nicer state.

I still need to figure out the e2e issue. Seems that I'm trying to look up an empty user, which would only be the case if users.Current() returns an empty username "" I think.

Re the Semgrep issue:  If I want to import nomad/structs into a driver, the driver is MPL but structs is BSL, so Semgrep is yelling at me. I see that in some other driver files we do this though. I could work around this by putting my struct elsewhere (I think?) but it feels like the right place to put it. Will try to figure out what's ideal here.

drivers/exec/driver.go
```diff
@@ -688,3 +735,12 @@ func (d *Driver) ExecTaskStreamingRaw(ctx context.Context,
 
 	return handle.exec.ExecStreaming(ctx, command, tty, stream)
 }
+
+func getUsername(cfg *drivers.TaskConfig) string {
```
tgross, line 739: When I see the failing E2E test and look at this function, it makes me think that instead of having to call this function every time we want `cfg.User`, that we should _set_ `cfg.User` like this:

```go
if cfg.User == "" {
	cfg.User = "nobody"
}
```

somewhere closer to the top of `StartTask` (maybe even in `validateUserIds`?). That way you only have to do this once and now all downstream consumers of `cfg.User`, like `(*TaskConfig).Validate` know that it's definitely non-empty.

drivers/exec/driver_test.go
```diff
@@ -872,8 +907,48 @@ func TestDriver_Config_validate(t *testing.T) {
...
+		expectedErr  string
+	}{
+		{config: configAllowCurrent, driverConfig: driverConfigSpecifyCurrent, expectedErr: ""},
+		{config: configDenyCurrent, driverConfig: driverConfigNoUserSpecified, expectedErr: ""},
+		{config: configDenyCurrent, driverConfig: driverConfigSpecifyCurrent, expectedErr: currentUserErrStr},
+		{config: configDenyAnonymous, driverConfig: driverConfigNoUserSpecified, expectedErr: anonUserErrStr},
+	} {
+		err := (&TaskConfig{}).validateUserIds(&tc.driverConfig, &tc.config)
+		if tc.expectedErr == "" {
+			require.NoError(t, err)
```
tgross, line 942:
```suggestion
			must.NoError(t, err)
```

drivers/shared/validators/validators.go
tgross: > This is used by exec/driver.go which does not have !windows in it, so I think it gets tricky.

Right, but I think that's probably ok here? For example, if I later decide to add UID/GID restrictions to the `java` driver, which does run on Windows, I would want the compile to fail if I don't properly split out the behavior so that the Windows build never calls this code.

drivers/exec/driver.go
tgross: > The Docker driver had a similar pattern with some config attributes, but I'm not positive there's not a nicer way than having two similar attrs on the struct (see above).

Yeah that's pretty much it. If you dig a little bit in other config values you'll see that's how we end up having to do pretty much anything that needs special parsing (ex. parsing "3h" as `time.Duration(3 * time.Hour)`)

drivers/exec/driver.go
mikenomitch: +1 to this idea - will implement

drivers/shared/validators/validators.go
mikenomitch: What I could do is remove the uid/gid bits from this and take those as args, and then this file could run on Windows just fine and work - you just wouldn't want to because it probably wouldn't be useful, but nothing would break. And then I could use the windows/unix builds of raw_exec to gate usage (like I already do).

If I am using the validators package to keep the struct for IdRange then I can't make it all unix only because then the import into the Config won't work. I could always make a structs package to get around that though.

apollo13:
Probably a stupid question, but wouldn't it make more sense to have a whitelist for users you want to allow? I mean in general I can imagine forbidding everything below id 1000 which are usually system users. But above 1000 it is really hard to not end up with a list with plenty of items. And supporting names instead of just the raw ids would be lovely as well. Ie I want to block the group "wheel" and "sudo" on every system without bothering which exact gid it has on the system (those are commonly used for sudo on rhel or debian and derivatives).

drivers/exec/driver.go
```diff
@@ -159,6 +166,9 @@ type Config struct {
 	// AllowCaps configures which Linux Capabilities are enabled for tasks
 	// running on this node.
 	AllowCaps []string `codec:"allow_caps"`
+
+	DeniedHostUidsStr string `codec:"denied_host_uids"`
```
mismithhisler, line 170: A total nit, but it looks like these variables were suffixed with `Str` because there used to be other variables of the same name that have since been removed.  Would it make sense to remove the suffix now?

Juanadelacuesta: you are right, I picked this one from someone else :)

client/lib/numalib/hw/ids.go
```diff
@@ -14,9 +14,15 @@ type (
...
 	SocketID uint8
 
 	// A CoreID represents one logical (vCPU) core.
 	CoreID uint16
+
+	// A GroupID (GID) represents a unique numerical value assigned to each user group.
+	GroupID uint64
+
+	// A UserID represents a unique numerical value assigned to each user account.
+	UserID uint64
```
tgross, line 27: This package feels like the wrong place for these types -- uid/gid doesn't have anything to do with NUMA hardware.

drivers/rawexec/driver.go
```diff
@@ -107,6 +111,10 @@ var (
 	}
 )
 
+type UserIDValidator interface {
+	HasValidIDs(user *user.User) error
+}
```
tgross, line 116: We've defined this for both drivers, so maybe this interface type should live in the validators package?

drivers/rawexec/driver_unix.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/hashicorp/nomad/plugins/drivers"
+)
+
+func (d *Driver) Validate(cfg drivers.TaskConfig) error {
+	usernameToLookup := cfg.User
+	var user *user.User
+	var err error
+
+	// Uses the current user of the client agent process
+	// if no override is given (differs from exec)
```
tgross, line 22: This is different than `exec` but it also seems like we could push the `users.Lookup` call into the `HasValidIDs` function because it's going to be used everywhere (just with different user names).

drivers/rawexec/driver_unix.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"fmt"
+	"os/user"
+
+	"github.com/hashicorp/nomad/helper/users"
+	"github.com/hashicorp/nomad/plugins/drivers"
+)
+
+func (d *Driver) Validate(cfg drivers.TaskConfig) error {
```
tgross, line 16: We're not validating anything but the user here, right? Should this function signature be identical to `validateUserIds` in the `exec` driver?

drivers/shared/validators/validators.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func NewValidator(logger hclog.Logger, deniedHostUIDs, deniedHostGIDs string) (*validator, error) {
+	valLogger := logger.Named("id_validator")
+
+	err := validateIDRange("deniedHostUIDs", deniedHostUIDs)
+	if err != nil {
+		return nil, err
+	}
+	valLogger.Debug("user range configured", "denied range", deniedHostUIDs)
```
tgross, line 42: These two log lines were probably useful for debugging, but in the non-error case are they at all useful?

website/content/docs/drivers/exec.mdx
```diff
@@ -188,6 +188,30 @@ able to make use of IPC features, like sending unexpected POSIX signals.
...
+config {
+  denied_host_uids = "0,10-15,22"
+}
+```
+
+- `denied_host_gids` - (Optional) Specifies a comma-separated list of host uids to
+  deny. Ranges can be specified by using a hyphen separating the two inclusive ends.
+  If a "user" value is specified in task configuration and that user has is part of
+  any groups with gid's in the specified ranges, the task will error before
+  starting. This will not be checked on Windows clients.
```
mismithhisler, line 207:
```suggestion
- `denied_host_gids` - (Optional) Specifies a comma-separated list of host gids to
  deny. Ranges can be specified by using a hyphen separating the two inclusive ends.
  If a "user" value is specified in task configuration and that user is part of
  any groups with gid's in the specified ranges, the task will error before
  starting. This will not be checked on Windows clients.
```

website/content/docs/drivers/raw_exec.mdx
```diff
@@ -130,6 +130,30 @@ client {
...
+config {
+  denied_host_uids = "0,10-15,22"
+}
+```
+
+- `denied_host_gids` - (Optional) Specifies a comma-separated list of host uids to
+  deny. Ranges can be specified by using a hyphen separating the two inclusive ends.
+  If a "user" value is specified in task configuration and that user has is part of
+  any groups with gid's in the specified ranges, the task will error before
+  starting. This will not be checked on Windows clients.
```
mismithhisler, line 149:
```suggestion
- `denied_host_gids` - (Optional) Specifies a comma-separated list of host gids to
  deny. Ranges can be specified by using a hyphen separating the two inclusive ends.
  If a "user" value is specified in task configuration and that user is part of
  any groups with gid's in the specified ranges, the task will error before
  starting. This will not be checked on Windows clients.
```

mismithhisler (changes requested):

drivers/shared/validators/validators.go
```diff
@@ -0,0 +1,141 @@
...
+	v := &validator{
+		deniedUIDs: idset.Parse[hw.UserID](deniedHostUIDs),
+		deniedGIDs: idset.Parse[hw.GroupID](deniedHostGIDs),
+		logger:     valLogger,
+	}
+
+	return v, nil
+}
+
+// HasValidIds is used when running a task to ensure the
```
jrasell, line 59:
```suggestion
// HasValidIDs is used when running a task to ensure the
```

drivers/rawexec/driver_unix.go
jrasell: Alternatively, we could move the `OOMScoreAdj` check from the `StartTask` function into here?

drivers/rawexec/driver_unix.go
```diff
@@ -0,0 +1,36 @@
...
+	usernameToLookup := cfg.User
+	var user *user.User
```
jrasell, line 18: This variable collides with the imported package of the same name, could we change it?

drivers/shared/validators/validators.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/hashicorp/nomad/client/lib/numalib/hw"
+)
+
+var (
+	ErrInvalidBound = errors.New("range bound not valid")
+	ErrEmptyRange   = errors.New("range value cannot be empty")
+	ErrInvalidRange = errors.New("lower bound cannot be greater than upper bound")
+)
+
+type validator struct {
```
jrasell, line 24: Do we want to export this type considering it is an used by other packages with exported functions?

jrasell (changes requested):
Added some inline comments.

The semgrep CI check is also failing:
```
    drivers/shared/validators/validators.go 
       semgrep.mpl_busl                                                         
          BUSL package `github.com/hashicorp/nomad/client/lib/idset` imported in
  MPL package                                                                   
          `validators`                                                          
                                                                                
           14┆ "github.com/hashicorp/nomad/client/lib/idset"
       semgrep.mpl_busl                                                         
          BUSL package `github.com/hashicorp/nomad/client/lib/numalib/hw`       
  imported in MPL package                                                       
          `validators`                                                          
                                                                                
           15┆ "github.com/hashicorp/nomad/client/lib/numalib/hw"
                                                    
    drivers/shared/validators/validators_default.go 
       semgrep.mpl_busl                                                         
          BUSL package `github.com/hashicorp/nomad/client/lib/numalib/hw`       
  imported in MPL package                                                       
          `validators`                                                          
                                                                                
           11┆ "github.com/hashicorp/nomad/client/lib/numalib/hw"
                                                 
    drivers/shared/validators/validators_unix.go 
       semgrep.mpl_busl                                                         
          BUSL package `github.com/hashicorp/nomad/client/lib/numalib/hw`       
  imported in MPL package                                                       
          `validators`                                                          
                                                                                
           13┆ "github.com/hashicorp/nomad/client/lib/numalib/hw"
                            
  BLOCKING CODE RULES FIRED:
    semgrep.mpl_busl
```

drivers/shared/validators/validators.go
Juanadelacuesta: they allow to check for the correct flow of the validator creation.

drivers/shared/validators/validators.go
Juanadelacuesta: what do you mean?

client/lib/numalib/hw/ids.go
Juanadelacuesta: moved!

drivers/rawexec/driver.go
Juanadelacuesta: They might grow in different directions, I don't want to limit it from the start

drivers/rawexec/driver_unix.go
Juanadelacuesta: I was trying to keep as much as possible from the original code, but since the windows/unix diasctintion is not necessary anymore, it makes sense to remove this, lets do it.

drivers/rawexec/driver_unix.go
Juanadelacuesta: You are right, we are actually doing it twice, let's move it there.

drivers/rawexec/driver_unix.go
Juanadelacuesta: right now it only checks for users, but it leaves a clear space for validations and might help us keep the code organised later

drivers/rawexec/driver.go
tgross: I think I see where you're going with that, inasmuch as many each driver only cares about a subset of the methods on the object returned by `validator.NewValidator`.

tgross:
> The semgrep CI check is also failing:

It's detecting the issue described here https://github.com/hashicorp/nomad/issues/19833 because the files were touched. We should fix that, but probably under a separate PR.

tgross (approved):
LGTM once the test compilation error is fixed.

mismithhisler (commented):
Looks good!

mismithhisler (approved):

jrasell (approved):

PR #20074, opened 2024-03-05
[f-gh-208] Force recreation and redeployment of task if volume label changes

Juanadelacuesta:
In a [past PR](https://github.com/hashicorp/nomad/pull/19886), a new parameter was introduced as part of the volume mount configuration in the task definition. It controls a security feature in case SELinux is being used. In case the job definition changes, the security feature needs to be reevaluated and the podman plugin in charge of doing the mount re run. This PR forces the scheduler to restart the allocations in case of a change in the job definition.

For more information, refer to the [ticket](https://github.com/hashicorp/nomad-driver-podman/pull/208).

### DIFF
--- a/scheduler/util.go
+++ b/scheduler/util.go
@@ -334,6 +334,18 @@ func tasksUpdated(jobA, jobB *structs.Job, taskGroup string) comparison {
 			return difference("task log disabled", at.LogConfig.Disabled, bt.LogConfig.Disabled)
 		}
 
+		// Check volume mount updates
+		bVMMap := helper.SliceToMap[map[string]*structs.VolumeMount](bt.VolumeMounts,
+			func(vm *structs.VolumeMount) string {
+				return vm.Volume
+			})
+
+		for _, atvm := range at.VolumeMounts {
+			if c := volumeMountUpdated(atvm, bVMMap[atvm.Volume]); c.modified {
+				return c
+			}
+		}
+
 		// Check if restart.render_templates is updated
 		if c := renderTemplatesUpdated(at.RestartPolicy, bt.RestartPolicy,
 			"task restart render_templates"); c.modified {
@@ -421,6 +433,17 @@ func connectServiceUpdated(servicesA, servicesB []*structs.Service) comparison {
 			}
 		}
 	}
+	return same
+}
+
+// volumeMountUpdated returns true if the definition of the volume mount
+// has been updated in a manner that will requires the task to be recreated.
+func volumeMountUpdated(mountA, mountB *structs.VolumeMount) comparison {
+	if mountA != nil && mountB != nil &&
+		mountA.SELinuxLabel != mountB.SELinuxLabel {
+		return difference("volume mount selinux label", mountA.SELinuxLabel, mountB.SELinuxLabel)
+	}
+
 	return same
 }
 
--- a/scheduler/util.go
+++ b/scheduler/util.go
@@ -438,23 +438,7 @@ func volumeMountsUpdated(a, b []*structs.VolumeMount) comparison {
 		return same
 	}
 
-	bVMMap := helper.SliceToMap[map[string]*structs.VolumeMount](b,
-		func(vm *structs.VolumeMount) string {
-			return vm.Volume
-		})
-
-	for _, atvm := range a {
-		if c := volumeMountUpdated(atvm, bVMMap[atvm.Volume]); c.modified {
-			return c
-		}
-		delete(bVMMap, atvm.Volume)
-	}
-
-	if len(bVMMap) > 0 {
-		return difference("volume mount added", nil, bVMMap)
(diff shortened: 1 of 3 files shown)


scheduler/util.go
```diff
@@ -424,6 +436,17 @@ func connectServiceUpdated(servicesA, servicesB []*structs.Service) comparison {
...
 
+// volumeMountUpdated returns true if the definition of the volume mount
+// has been updated in a manner that will requires the task to be recreated.
+func volumeMountUpdated(mountA, mountB *structs.VolumeMount) comparison {
+	if mountA != nil && mountB != nil &&
+		mountA.SELinuxLabel != mountB.SELinuxLabel {
+		return difference("volume mount selinux label", mountA.SELinuxLabel, mountB.SELinuxLabel)
+	}
+
+	return same
```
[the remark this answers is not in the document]
Juanadelacuesta: :O Lets take a look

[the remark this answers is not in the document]
Juanadelacuesta: If mountB is nil, it means the volume is no longer mounted and there is no need to relaunch the task, right?

scheduler/util.go
```diff
@@ -334,6 +334,18 @@ func tasksUpdated(jobA, jobB *structs.Job, taskGroup string) comparison {
...
+		bVMMap := helper.SliceToMap[map[string]*structs.VolumeMount](bt.VolumeMounts,
+			func(vm *structs.VolumeMount) string {
+				return vm.Volume
+			})
+
+		for _, atvm := range at.VolumeMounts {
+			if c := volumeMountUpdated(atvm, bVMMap[atvm.Volume]); c.modified {
+				return c
+			}
+		}
```
tgross, line 347: Because we're looping over the volume mounts for (A) but not (B), this detects changes volume mounts if both versions (A) and (B) have them, and detects if the new version (A) adds a volume mount that the existing version (B) doesn't have, but I don't think it detects if version (A) removes a volume mount that version (B) has?

See my comment on the test.

scheduler/util_test.go
```diff
@@ -547,6 +547,34 @@ func TestTasksUpdated(t *testing.T) {
...
+	j32.TaskGroups[0].Tasks[0].VolumeMounts = []*structs.VolumeMount{
+		{
+			Volume:       "myvolume",
+			SELinuxLabel: "",
+		},
+	}
+
+	must.True(t, tasksUpdated(j31, j32, name).modified)
+
+	// Remove volume mount
```
tgross, line 574: This is adding a volume mount, because in all the callers for `tasksUpdated`, the right hand argument "jobB" is the existing job.

```suggestion
	// Add volume mount
```

But we should have a test for both adding and removing.

tgross (changes requested):

scheduler/util.go
shoenig: Since we only really care about _if_ the volume mounts do not match (and order does not matter) a simple way to do this could be to stuff them in a HashSet (before and after) and ask the Sets if they are equal.

Juanadelacuesta: @shoenig Done!

scheduler/util.go
```diff
@@ -424,6 +430,48 @@ func connectServiceUpdated(servicesA, servicesB []*structs.Service) comparison {
...
+	for _, atvm := range a {
+		if c := volumeMountUpdated(atvm, bVMMap[atvm.Volume]); c.modified {
+			return c
+		}
+		delete(bVMMap, atvm.Volume)
+	}
+
+	if len(bVMMap) > 0 {
+		return difference("volume mount added", nil, bVMMap)
+	}
```
shoenig, line 455: What I was trying to get at before is that all of this can now be eliminated in favor of just

```
return difference("volume mounts changed")
```

the `comparison` we create is nothing more than a debugging tool; if you get the "volume mounts changed" differential the person debugging should be able to figure it out by just looking at the submitted job(s).

Juanadelacuesta: not sure I follow what you mean

tgross (approved):
LGTM once the conversation with @shoenig is resolved

scheduler/util.go
shoenig: The only thing we care about from `volumeMountsUpdated` function is whether or not volume mounts were modified (while ignoring sort order). All of the logic for determining _which_ volume mount changed is extraneous / bug prone.

Since we don't care about _which_ volume mount changed, we can delete all this code and just use the `Equal` method between the two sets, in which case this function simplifies to something like,

```
func volumeMountsUpdated(a, b []*structs.VolumeMount) comparison {
	setA := set.HashSetFrom(a)
	setB := set.HashSetFrom(b)

	if setA.Equal(setB) {
		return same
	}

    return difference("volume mounts changed")
}

PR #20075, opened 2024-03-05
docs: skipping a major release is fine

schmichael:
Nomad has always placed an extremely high priority on backward compatibility. We have always aimed to support N-2 major releases and usually gone above and beyond that.

The new https://www.hashicorp.com/long-term-support policy also mentions that N-2 is what we have always supported, so it's probably time for our docs to reflect that reality.


tgross (approved):
LGTM

PR #20093, opened 2024-03-07
exec2: add a client.users configuration block

shoenig:
For now just add min/max dynamic user values; soon we can also absorb
the `user.denylist` and `user.checked_drivers` options from the
deprecated `client.options` map.

e.g. usage
```
client {
  enabled = true

  users {
    dynamic_user_min = 70100
    dynamic_user_max = 70200
  }
}
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -477,8 +477,8 @@ func NewClient(cfg *config.Config, consulCatalog consul.CatalogAPI, consulProxie
 
 	// Create the dynamic workload users pool
 	c.users = dynamic.New(&dynamic.PoolConfig{
-		MinUGID: 80_000, // TODO(shoenig) plumb client config
-		MaxUGID: 89_999, // TODO(shoenig) plumb client config
+		MinUGID: cfg.Users.MinDynamicUser,
+		MaxUGID: cfg.Users.MaxDynamicUser,
 	})
 
 	// Create the cpu core partition manager
--- a/client/config/config.go
+++ b/client/config/config.go
@@ -328,6 +328,9 @@ type Config struct {
 	// Drain configuration from the agent's config file.
 	Drain *DrainConfig
 
+	// Uesrs configuration from the agent's config file.
+	Users *UsersConfig
+
 	// ExtraAllocHooks are run with other allocation hooks, mainly for testing.
 	ExtraAllocHooks []interfaces.RunnerHook
 }
@@ -757,6 +760,7 @@ func (c *Config) Copy() *Config {
 	nc.TemplateConfig = c.TemplateConfig.Copy()
 	nc.ReservableCores = slices.Clone(c.ReservableCores)
 	nc.Artifact = c.Artifact.Copy()
+	nc.Users = c.Users.Copy()
 	return &nc
 }
 
@@ -805,6 +809,10 @@ func DefaultConfig() *Config {
 		CgroupParent:       "nomad.slice", // SETH todo
 		MaxDynamicPort:     structs.DefaultMinDynamicPort,
 		MinDynamicPort:     structs.DefaultMaxDynamicPort,
+		Users: &UsersConfig{
+			MinDynamicUser: 80_000,
+			MaxDynamicUser: 89_999,
+		},
 	}
 
 	return cfg
--- a/client/config/users.go
+++ b/client/config/users.go
@@ -1,0 +1,32 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package config
+
+import "github.com/hashicorp/nomad/nomad/structs/config"
+
+// UsersConfig configures things related to operating system users.
+type UsersConfig struct {
+	// MinDynamicUser is the lowest uid/gid for use in the dynamic users pool.
+	MinDynamicUser int
+
+	// MaxDynamicUser is the highest uid/gid for use in the dynamic users pool.
+	MaxDynamicUser int
+}
+
+func UsersConfigFromAgent(c *config.UsersConfig) *UsersConfig {
+	return &UsersConfig{
+		MinDynamicUser: *c.MinDynamicUser,
+		MaxDynamicUser: *c.MaxDynamicUser,
+	}
+}
+
+func (u *UsersConfig) Copy() *UsersConfig {
+	if u == nil {
+		return nil
+	}
+	return &UsersConfig{
+		MinDynamicUser: u.MinDynamicUser,
+		MaxDynamicUser: u.MaxDynamicUser,
+	}
+}
--- a/client/config/users.go
+++ b/client/config/users.go
@@ -3,7 +3,7 @@
 
 package config
 
-import "github.com/hashicorp/nomad/nomad/structs/config"
+import sconfig "github.com/hashicorp/nomad/nomad/structs/config"
 
 // UsersConfig configures things related to operating system users.
 type UsersConfig struct {
@@ -14,7 +14,7 @@ type UsersConfig struct {
 	MaxDynamicUser int
 }
 
-func UsersConfigFromAgent(c *config.UsersConfig) *UsersConfig {
+func UsersConfigFromAgent(c *sconfig.UsersConfig) *UsersConfig {
 	return &UsersConfig{
 		MinDynamicUser: *c.MinDynamicUser,
 		MaxDynamicUser: *c.MaxDynamicUser,
--- a/helper/users/dynamic/pool.go
+++ b/helper/users/dynamic/pool.go
@@ -24,6 +24,9 @@ var (
 // none indicates no dynamic user
 const none = 0
 
+// doNotEnable indicates functionality should be disabled
+const doNotEnable = -1
+
 // A UGID is a combination User (UID) and Group (GID). Since Nomad is
 // allocating these values together from the same pool it can ensure they are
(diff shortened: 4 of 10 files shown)


client/config/users.go
```diff
@@ -0,0 +1,32 @@
...
+
+func (u *UsersConfig) Copy() *UsersConfig {
+	if u == nil {
+		return nil
+	}
+	return &UsersConfig{
+		MinDynamicUser: u.MinDynamicUser,
+		MaxDynamicUser: u.MaxDynamicUser,
+	}
+}
```
Juanadelacuesta, line 32: No need to validate?

helper/users/dynamic/pool.go
```diff
@@ -81,6 +94,20 @@ func New(opts *PoolConfig) Pool {
 	}
 }
 
+// disable is an implementation of Pool that does not allow acquiring ugids
+type disable struct{}
```
Juanadelacuesta, line 98: I would give it a more explicit name, that does not requiere reading the comments to know what it is

helper/users/dynamic/pool.go
```diff
@@ -60,11 +63,21 @@ type PoolConfig struct {
...
+// indicating the client should not enable the dynamic workload users
+// functionality
+func (p *PoolConfig) disable() bool {
+	return p.MinUGID == doNotEnable || p.MaxUGID == doNotEnable
+}
+
 // New creates a Pool with the given PoolConfig options.
 func New(opts *PoolConfig) Pool {
 	if opts == nil {
 		panic("bug: users pool cannot be nil")
```
Juanadelacuesta, line 76: Why do you panic instead of returning an error?

nomad/structs/config/users.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+}
+
+// Validate whether UsersConfig is valid.
+//
+// Note that -1 is a valid value for min/max dynamic users, as this is used
+// to indicate the dynamic workload users feature should be disabled.
+func (u *UsersConfig) Validate() error {
+	if u == nil {
+		return errors.New("users must not be nil")
```
Juanadelacuesta, line 69: Creating the errors as vars helps you make cleaner tests, where you check the error itself, instead of the error message

helper/users/dynamic/pool.go
shoenig: good point ~ renamed to `noopPool` instead

helper/users/dynamic/pool.go
shoenig: Each of these conditions shouldn't be possible in a normal nomad agent - the validation occurs well before here in the config parsing. The panics here just serve to remind test authors they're trying to set an invalid configuration.

Juanadelacuesta (commented):
A couple of questions here and there :)

client/client.go
```diff
@@ -477,8 +477,8 @@ func NewClient(cfg *config.Config, consulCatalog consul.CatalogAPI, consulProxie
 
 	// Create the dynamic workload users pool
 	c.users = dynamic.New(&dynamic.PoolConfig{
-		MinUGID: 80_000, // TODO(shoenig) plumb client config
-		MaxUGID: 89_999, // TODO(shoenig) plumb client config
+		MinUGID: cfg.Users.MinDynamicUser,
+		MaxUGID: cfg.Users.MaxDynamicUser,
```
lgfa29, line 481: TODONE 🎉

client/config/users.go
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package config
+
+import "github.com/hashicorp/nomad/nomad/structs/config"
```
lgfa29, line 6:
```suggestion
package config

import sconfig "github.com/hashicorp/nomad/nomad/structs/config"
```

Minor nit-picking since you can't reference you own package, so feel free to ignore this, but since there are some structs with the same name it could be useful to highlight which `config` is being used.

client/config/config.go
```diff
@@ -805,6 +809,10 @@ func DefaultConfig() *Config {
 		CgroupParent:       "nomad.slice", // SETH todo
 		MaxDynamicPort:     structs.DefaultMinDynamicPort,
 		MinDynamicPort:     structs.DefaultMaxDynamicPort,
+		Users: &UsersConfig{
+			MinDynamicUser: 80_000,
+			MaxDynamicUser: 89_999,
+		},
```
lgfa29, line 815: Just out of curiosity, are these defaults OS specific? I'm not sure how they work on Windows 😬

helper/users/dynamic/pool.go
```diff
@@ -81,6 +94,20 @@ func New(opts *PoolConfig) Pool {
 	}
 }
 
+// noopPool is an implementation of Pool that does not allow acquiring ugids
+type noopPool struct{}
+
+func (*noopPool) Restore(UGID) {}
+func (*noopPool) Acquire() (UGID, error) {
+	return 0, errors.New("dynamic workload users disabled")
```
lgfa29, line 102: Is it always safe to return an error here? If the `UGID` returned doesn't really matter if the pool is disabled, then it may be safer to just `return 0, nil`.

I once had a noop struct that returned an error because I thought it would never happen, but then it did and broke stuff, so I'm always worried about noops that actually do something somewhere else 😅

nomad/structs/config/users.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/hashicorp/nomad/helper/pointer"
+)
+
+// UsersConfig configures things related to operating system users.
+type UsersConfig struct {
+	// MinDynamicUser is the lowest uid/gid for use in the dynamic users pool.
+	MinDynamicUser *int `hcl:"dynamic_user_min"`
+
+	// MaxDynamicUser is the highest uid/gid for use in the dynamic users pool.
+	MaxDynamicUser *int `hcl:"dynamic_user_max"`
```
lgfa29, line 18:
```suggestion
	// MinDynamicUser is the lowest uid/gid for use in the dynamic users pool.
	MinDynamicUser *int `hcl:"min_guid"`

	// MaxDynamicUser is the highest uid/gid for use in the dynamic users pool.
	MaxDynamicUser *int `hcl:"max_guid"`
```

Minor suggestion, but since this config is already scoped in an `users` block, I think it's safe to drop the `user` component? `dynamic` also didn't seem to convey much information to me, and it wasn't very clear what the value represents (is it a count? a group ID? etc.) but I may be wrong 😬 

Before:

```hcl
users {
  dynamic_user_min = 10000
  dynamic_user_max = 20000
}
```

After:

```hcl
users {
  min_guid = 10000
  max_guid = 20000
}
```

I would maybe suggest even making `users` more explicit, like `task_users`, but I don't know if that's a good name.

nomad/structs/config/users.go
```diff
@@ -0,0 +1,100 @@
...
+		return errDynamicUserMinInvalid
+	}
+	if u.MaxDynamicUser == nil {
+		return errDynamicUserMaxUnset
+	}
+	if *u.MaxDynamicUser < -1 {
+		return errDynamicUserMaxInvalid
+	}
+	return nil
+}
```
lgfa29, line 92: Minor UX thought, but does this method prevent the client from starting? 

Whenever possible I like when programs tell me as much as possible of what's wrong. For example, a config like this:

```hcl
users {
    dynamic_user_min = -12
}
```

Has two errors (`errDynamicUserMinInvalid` and `errDynamicUserMaxUnset`), but only `errDynamicUserMinInvalid` is returned. If I then fix that, I will get `errDynamicUserMaxUnset`. Returning both  allows me to fix them in one go.

lgfa29 (approved):
Minor ramblings, but LGTM!

client/config/config.go
shoenig: Even worse, 80k-90k happens to be a gap not reserved (by conveention) by systemd. Users of non-systemd linux systems will need to think about this range and whether it is suitable for them. Will be covered in the docs for exec2.

nomad/structs/config/users.go
shoenig: The problem with `users.min_guid` etc. is that this isn't a reference to generic users on the system - it's specific to the new concept of "Dynamic Workload Users" which is exposed for task drivers who expressly indicate support through a new task driver Capability. 

Even more so, a generic name would be conflicting with the very similar sounding restrictions @mikenomitch is working on in https://github.com/hashicorp/nomad/pull/20073

helper/users/dynamic/pool.go
shoenig: So the case where this error would be returned is if the user has `exec2` and a task using it, but has their client configured with `users {dynamix_user{min,max} = -1}` (disabling the anonymous workload users functionality). The point is that in this scenario they would be requiring their job submitters to set an actual value for the `task.user` field. If they don't they will get this error.

Perhaps we should set a fingerprint value on whether dynamic workload users are enabled on the client. Although I don't know how we would set a constraint - we would need to know whether the task driver makes use of dynamic workload users at job submission time, and that isn't something we know (sort of, we do know for exec2 because it's ours...)

nomad/structs/config/users.go
shoenig: An interesting point, but we don't follow that pattern for much else in client config, I don't think. It's mostly for interactive stuff like job submission.

PR #20105, opened 2024-03-08
Bump consul-template to 0.37.2

lkysow:


lkysow:
👍 the actual consul-template release hasn't finished yet so let's make sure that goes through too, I'll comment when that's complete

tgross:
Can you run `go mod tidy` as well? Looks like the linter caught something it didn't like there.

lkysow:
tidy done and ct release went through so good to merge

PR #20115, opened 2024-03-11
exec2: add client support for unveil filesystem isolation mode

shoenig:
This PR adds support for a new filesystem isolation mode, `Unveil` (in addition to `None`, `Image`, `Chroot`). The mode introduces a "alloc_mounts" directory where tasks have user-owned directory structure which are bind mounts into the real alloc directory structure. This enables a task driver to use `landlock` on Linux to isolate a task to the task owned directory structure, providing sandboxing.

This PR creates a `fsisolation` package and moves the FS Isolation enums there to avoid import cycles, now that the isolation level needs to be passed into the taskdir builder. We create alias values in the drivers package so that existing external plugins do not break and do not need to refactor.


client/allocdir/task_dir.go
```diff
@@ -145,12 +170,33 @@ func (t *TaskDir) Build(createChroot bool, chroot map[string]string) error {
...
 
+	// Only bind mount the task alloc/task dirs to the client.mounts_dir/<task>
+	if fsi == fsisolation.Unveil {
+		uid, gid, _, err := dynamic.LookupUser(username)
+		if err != nil {
+			return fmt.Errorf("Failed to lookup user: %v", err)
+		}
+
+		// create the task unique directory under the client mounts path
+		parent := filepath.Dir(t.MountsAllocDir)
```
tgross, line 187: If I'm understanding correctly, we're:
* creating `$datadir/mounts/$allocID-$task` with permissions 0710
* bind-mounting from `$datadir/allocs/$allocID` to `$datadir/mounts/$allocID-$task/alloc` with permissions 0710
* bind-mounting from `$datadir/allocs/$allocID/$task` to `$datadir/mounts/$allocID-$task/task` with permissions 0710

If so, that all seems ok... but where is `$datadir/mounts` getting created with the recommended permissions of 0700? I seem to remember that we don't set `$datadir/allocs` with the permissions we'd really recommend but now it's a backcompat issue, so if we could ensure we're creating the mounts dir with the right permissions the first time we could probably avoid some future headaches.

client/config/config.go
```diff
@@ -97,8 +97,16 @@ type Config struct {
...
 	// AllocDir is where we store data for allocations
+	//
+	// In a production environment this should be owned by root with file
+	// mode 0o700.
 	AllocDir string
 
+	// AllocMountsDir is where we bind mount paths from AllocDir for tasks making
+	// use of the unveil file isolation mode. In a production environment this
+	// should be owned  by root with file mode 0o755.
+	AllocMountsDir string
```
tgross, line 108: Should we add website docs for the `alloc_mounts` configuration in this PR?

client/allocdir/task_dir.go
```diff
@@ -8,80 +8,105 @@ import (
...
-	// Dir is the path to Task directory on the host
+	// Dir is the path to Task directory on the host.
+	//
+	// <task_dir>
 	Dir string
 
+	// MountsAllocDir is the path to the alloc directory on the host that has
+	// been bind mounted under <client.mounts_dir>
+	//
+	// <client.mounts_dir>/task/<alloc_dir> -> <alloc_dir>
```
tgross, line 34: I don't think the documented source path here matches what we ended up doing with the unique task name directory? (Same for `MountsTaskDir` below.)

client/allocdir/task_dir.go
```diff
@@ -145,12 +170,33 @@ func (t *TaskDir) Build(createChroot bool, chroot map[string]string) error {
...
+			return fmt.Errorf("Failed to lookup user: %v", err)
+		}
+
+		// create the task unique directory under the client mounts path
+		parent := filepath.Dir(t.MountsAllocDir)
+		if err = os.MkdirAll(parent, 0o710); err != nil {
+			return fmt.Errorf("Failed to create task mount directory: %v", err)
+		}
+		if err = os.Chown(parent, uid, gid); err != nil {
+			return fmt.Errorf("Failed to chown task mount directory: %v", err)
```
tgross, line 192:
```suggestion
			return fmt.Errorf("Failed to chown task mount directory: %w", err)
```

tgross (approved):
LGTM (so long as any of my open questions aren't wildly off-base :grinning: )

client/allocdir/task_dir.go
shoenig: Ah dangit I [that part](https://github.com/hashicorp/nomad/compare/main...exec2-firstpass#diff-bd3a55a72186f59e2e63efb4951573b2f9e4a7cc98086e922b0859f8ccc1dd09R692-R699) out. The permissions on `$datadir/mounts` would be `0711` - the point of all this being the outside->in protections come from `$datadir/mounts/%allocID-$task` being owned by the user of the task (`0710` as mentioned above),  which itself can be a "dynamic workload user" that does not exist from the perspective of the OS. I'll add the directory creation in another commit.

shoenig: Added directory creation in ee9e854f94

client/config/config.go
shoenig: Will do a big PR at the end for all things exec2

client/allocdir/task_dir.go
shoenig: ooh yeah that was from before I figured out how things would look, fixed in 4a45eaf355

PR #20126, opened 2024-03-12
Support insecure flag on artifact

### LINKED ISSUE #19883: Support `insecure` flag on `artifact`
### Proposal

https://github.com/hashicorp/go-getter/pull/319 added a new configuration to allow `go-getter` to fetch artifacts from sources where their TLS certificate is not trusted by the system, but Nomad has this value hard-coded to `false`.
https://github.com/hashicorp/nomad/blob/f1637bdd5fe41d65e0be7776732b7f6a5098bf02/client/allocrunner/taskrunner/getter/params.go#L160-L166

In some scenarios it may be useful to fetch artifacts from sources with self-signed certificates, which `checksum` providing some level of file tempering protection.

### Use-cases

Retrieve artifacts from sources exposed with self-signed certificates.

amirabbas8:
Hey, I'm so glad this is my first PR to nomad. 🎉

this PR supports an insecure flag for artifact download as described in #19883 

The artifact block now supports an insecure flag like the example below. this flag will prevent the go-getter from validating the TLS certificate. 

```
      artifact {
        source        = "https://expired.badssl.com/"
        destination = "local/some-directory"
        insecure     = true
      }
```

Closes #19883

### DIFF
--- a/client/allocrunner/taskrunner/getter/sandbox_test.go
+++ b/client/allocrunner/taskrunner/getter/sandbox_test.go
@@ -4,6 +4,8 @@
 package getter
 
 import (
+	"net/http"
+	"net/http/httptest"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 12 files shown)


client/allocrunner/taskrunner/getter/sandbox_test.go
```diff
@@ -51,3 +51,27 @@ func TestSandbox_Get_http(t *testing.T) {
...
+	logger := testlog.HCLogger(t)
+
+	ac := artifactConfig(10 * time.Second)
+	sbox := New(ac, logger)
+
+	_, taskDir := SetupDir(t)
+	env := noopTaskEnv(taskDir)
+
+	artifact := &structs.TaskArtifact{
+		GetterSource: "https://expired.badssl.com",
```
shoenig, line 66: As neat as this service is, we probably shouldn't make requests to 3rdparty sites in out tests / CI runs. Instead I think we can get away with just starting an http server using [httptest.NewTLSServer](https://pkg.go.dev/net/http/httptest#NewTLSServer) and using our own `http.Client` which won't have the generated certs, and should fail as expected.

shoenig:
Oh you'll also want to run `make cl` to generate a changelog entry.

amirabbas8:
@shoenig, Thanks for your review. I have made the changes 🤩

shoenig:
Hey @amirabbas8 this is very close! Looks like there's a couple failing tests that need to be tweaked. I'm pretty sure in both cases it's just a matter of the test case comparing struct equality and not taking into account the new field.

```
=== FAIL: nomad/structs TestTaskDiff/Artifacts_edited (0.00s)
    diff_test.go:9235: 
        	Error Trace:	/home/runner/work/nomad/nomad/nomad/structs/diff_test.go:9235
```

and

```
 === FAIL: client/allocrunner/taskrunner/getter TestParameters_reader (re-run 2) (0.00s)
    assert.go:14: 
        params_test.go:65: expected equality via json marshalling
        ↪ Assertion | differential ↷
          strings.Join({
          	`{"alloc_dir":"/path/to/alloc","artifact_destination":"local/out.`,
          	`txt","artifact_headers":{"X-Nomad-Artifact":["hi"]},"artifact_`,
        + 	`insecure":false,"artifact_`,
          	`mode":2,"artifact_source":"https://example.com/file.txt","decomp`,
          	`ression_limit_file_count":3,"decompression_limit_size":98765,"di`,
          	... // 248 identical bytes
          }, "")
```

amirabbas8:
@shoenig, I fixed both test cases, thanks.

shoenig (approved):
LGTM; thanks @amirabbas8!

PR #20129, opened 2024-03-13
exec2: setup RPC plumbing for dynamic workload users

shoenig:
And pass the dynamic users pool from the client into the hook.


tgross (approved):
LGTM!

PR #20130, opened 2024-03-13
New API endpoint to back UI jobs index page uplift

### LINKED ISSUE #19339: Multi-job watchable allocation endpoint
The Jobs index page in the UI long-polls a blocking query for `/jobs` in order to show the data it currently shows:
image

We presently use the `jobs[].JobSummary.Summary.$groupName` object to populate the historical state of allocations in the jobs, and display it in the small chart on the right of the job rows.
image

However, in https://github.com/hashicorp/nomad/issues/16128, a major driver of the work was to stop using the historical state-stored alloc data and to instead to a live look-up. On a job-by-job basis, looking up /allocations and watching for changes makes sense and the browser can support holding open a blocking query. But on the jobs index page, this is asking the browser to keep an eye on a few too many things and in testing, this has proven to cause network queue backups.

I'd like to have a way to watch the allocations of a subset of jobs, and for that endpoint to provide the following properties + update whenever they change:
- ClientStatus
- DeploymentStatus.Canary
- DeploymentStatus.Healthy

---
Related, this table should also be able to keep track whenever a Job has an active deployment taking place. The Nomad front-end currently handles this by watching `Job.Deployments[].{latest}.status === running` which, like allocation watching, works fine when looking at a single job but is problematic for the browser when looking at an index.

This could be part of the same query described above for allocations, and simply expanded to something like /realtimejobs; or, it could be a second blocking query.

---

### Why though?

This would solve 3 notable problems on the job index page:
- The index page doesn't tell us the present status of a job's allocations, only their historical status
- A truer picture of things: accumulation of alloc statuses in the state store is sometimes incorrect (statuses can be incremented but not decremented under certain scenarios, like quickly failed deployments or client shut-downs)
- alerting an observing user when a deployment or job update is taking place

gulducat:
Introduce a new API `/v1/jobs/statuses`, primarily for use in the UI, that collates info about jobs, their allocations, and latest deployment.

Closes #19339 

Some of noteworthy tidbits:
* reads from 3 state tables ("jobs", "allocs", "deployment") to collate info about each job.
  * the "jobs" table is the primary table used to iterate, therefore any `?filter=`s apply to `structs.Job` (i.e. one can not filter based on allocs or such)
* in order to reproduce the current UI behavior which shows most-recently-changed jobs first, this uses a new "modify_index" index, so the jobs are sorted by ModifyIndex
  * the efficiency trade-off is a whole-table scan when filtering (since we're not using a namespace/id index)
  * currently the UI loads allll of `/v1/jobs`, then sorts client-side, but we're introducing pagination here, so less efficient in one sense, but more efficient in another
* the reply.Index used to unblock the RPC is the highest of all objects' ModifyIndex, instead of the latest index on any of the state tables, in order to avoid unblocking the on changes that occur "off page"

Hopefully tests and code comments cover the rest of the nuances.

I'm leaving the whole messy commit history for posterity, but will squash on merge!


api/jobs.go
```diff
@@ -1544,3 +1544,10 @@ func (j *Jobs) ActionExec(ctx context.Context,
 
 	return s.run(ctx)
 }
+
+type JobsStatusesRequest struct {
+	Jobs []struct { // TODO: proper type
+		ID        string `json:"id"`
+		Namespace string `json:"namespace"`
+	} `json:"jobs"` // TODO: unkeyed?
+}
```
tgross, line 1553: We have the type here but not a function that lets us use it in the Go API. If we had that, I think it would call out that we should probably have the `SmartOnly` field in this struct as well.

command/agent/job_endpoint.go
```diff
@@ -2060,3 +2060,48 @@ func validateEvalPriorityOpt(priority int) HTTPCodedError {
 	}
 	return nil
 }
+
+func (s *HTTPServer) JobsStatusesRequest(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
+	args := structs.JobsStatusesRequest{}
+	if s.parse(resp, req, &args.Region, &args.QueryOptions) {
+		return nil, nil // seems whack
```
tgross, line 2067: > // seems whack

It kinda is. I'm not sure why we return `true` on error rather than the error. The inner `parseWait` does the right thing for status codes though, which we don't when we `return nil, err`. That'd be a nice thing to fix more generally around all these little parse functions. But out of scope for this PR! :grinning:

nomad/jobs_endpoint.go
```diff
@@ -0,0 +1,297 @@
...
+		ctx:    ctx,
+		logger: s.logger.Named("jobs"),
+	}
+}
+
+type Jobs struct {
+	srv    *Server
+	ctx    *RPCContext
+	logger hclog.Logger
+}
```
tgross, line 33: It's not clear to me that this should be a different RPC endpoint struct than the existing `Job` struct. Having it in its own endpoint means it gets its own metrics collection (which is a little weird) and we have to create the struct, it's logger, etc. for every RPC request.

If we're just wary of adding more methods to the big `nomad/jobs_endpoint.go` file, we can always split it out into a separate file.

nomad/jobs_endpoint.go
```diff
@@ -0,0 +1,297 @@
...
+				filters = append(filters, paginator.GenericFilter{
+					Allow: func(i interface{}) (bool, error) {
+						job := i.(*structs.Job)
+						return jobSet.Contains(job.NamespacedID()), nil
+					},
+				})
+			}
+
+			jobs := make([]structs.UIJob, 0)
+			newJobs := set.New[structs.NamespacedID](0)
```
tgross, line 142: Can we set these values with a starting point of `len(args.Jobs)` to avoid re-allocations?

nomad/structs/job.go
```diff
@@ -16,6 +16,46 @@ const (
...
 
+type UIJob struct {
+	NamespacedID
+	Name          string
+	Type          string
+	NodePool      string
+	Datacenters   []string
+	Priority      int
+	Allocs        []JobStatusAlloc
+	SmartAlloc    map[string]int
```
tgross, line 27: The whole "smart" notion in this PR is going to need some explanation in docstrings, etc. You've got an explanation buried in the implementation, but we should have it somewhere in the struct definitions.

nomad/structs/job.go
```diff
@@ -16,6 +16,46 @@ const (
...
+	GroupCountSum int
+	ChildStatuses []string
+	DeploymentID  string
```
tgross, line 30: For clarity, maybe: 

```suggestion
	LatestDeploymentID  string
```

Also, should this object have any of the indexes for the jobs?

command/agent/job_endpoint.go
```diff
@@ -2060,3 +2060,48 @@ func validateEvalPriorityOpt(priority int) HTTPCodedError {
...
+	}
+	if smartOnly, err := parseBool(req, "smart_only"); err != nil {
+		return nil, err
+	} else if smartOnly != nil {
+		args.SmartOnly = *smartOnly
+	}
+	switch req.Method {
+	case http.MethodGet:
+		// GET requests will be treated as "get all jobs" but also with filtering and pagination and such
```
tgross, line 2076: Although it's against the spec, I'm _pretty_ sure Go's server will accept a body on GET anyways. So we're silently dropping the body here, and I'm not sure that's better or worse than returning an error that says "hey, use POST".

command/agent/job_endpoint.go
```diff
@@ -2060,3 +2060,48 @@ func validateEvalPriorityOpt(priority int) HTTPCodedError {
...
+	case http.MethodPost:
+		// POST requests expect a list of Jobs in the request body, which will then be filtered/paginated, etc.
+		var in api.JobsStatusesRequest
+		if err := decodeBody(req, &in); err != nil {
+			return nil, err
+		}
+		if len(in.Jobs) == 0 {
+			return nil, CodedError(http.StatusBadRequest, "no jobs in request")
+		}
```
tgross, line 2085: Likewise, here we're returning an error when we _could_ just do what the GET does and return all the things. Making the user switch between the two could be a little annoying. But maybe it doesn't matter if both the UI and Go API "does the right thing" and picks the correct method automatically.

nomad/jobs_endpoint.go
```diff
@@ -0,0 +1,297 @@
...
+	// we may get some extra efficiency, especially for non-contiguous job IDs.
+	if len(args.Jobs) > 0 {
+		nses := set.New[string](0)
+		for _, j := range args.Jobs {
+			nses.Insert(j.Namespace)
+		}
+		if nses.Size() == 1 {
+			namespace = nses.Slice()[0]
+		}
+	}
```
tgross, line 61: This is a nice little optimization, especially when combined with the early return we'd get because of the `AllocNsOp` check on line 69. :+1:

nomad/jobs_endpoint.go
```diff
@@ -0,0 +1,297 @@
...
+			if err != nil {
+				return structs.NewErrRPCCodedf(
+					http.StatusBadRequest, "failed to create result paginator: %v", err)
+			}
+
+			nextToken, err := pager.Page()
+			if err != nil {
+				return structs.NewErrRPCCodedf(
+					http.StatusBadRequest, "failed to read result page: %v", err)
+			}
```
tgross, line 174: Aren't these both internal errors?

nomad/structs/job.go
philrenaud: I think this is `DeploymentID` because it's meant as an indicator of "Is there currently a deployment happening?"

Maybe `ActiveDeploymentID` in that case, though?

nomad/jobs_endpoint.go
gulducat: I started it like this to have a clean slate for my mental model.  the only argument I have to keep it separate is that it's unique in accepting a list of jobs as request _inputs_, but it's a pretty weak point, and plenty of other "Job" endpoints deal with plural jobs as responses.

all that to say, I'll scoot it over to `Job`!

nomad/jobs_endpoint.go
gulducat: the caller could request invalid jobs, meaning the response will be smaller than the request.  but, I don't think that will be common, and it's still probably more efficient to avoid repeated mem allocs, as you say. :+1:

nomad/structs/job.go
gulducat: the "smart" wording was not intended to be the final product, and even stand out so much as to say "hey, do something about me! at least rename me, or something!"

I've been keeping it around in case @philrenaud could make use of it _instead of_ needing to include all the alloc stubs, or if it might be useful for some other as-yet-unknown purpose, but that has been less and less likely, so I think this extra little feature will probably just get stripped out.

nomad/structs/job.go
gulducat: I dig `ActiveDeploymentID` :+1:  or I could make it a `bool` `ActiveDeployment`, or similar?  Phil do you imagine that there will actually be any use for the ID, now or later?

for the indexes, do you mean their `ModifyIndex`?  I'm happy to include that or any other fields on here, but so far have been trying to keep it to "only what Phil needs" because it's easy to add stuff later, harder to take away if users start using it for their own purposes.

command/agent/job_endpoint.go
gulducat: my instinct is to enshrine the spec, but I'll play around with this a bit and see how it feels.

nomad/jobs_endpoint.go
gulducat: good catch. this was lifted verbatim from existing jobs endpoint, not sure why it called these "bad request"s, but I can change this'n to `http.StatusInternalServerError` (500)

nomad/structs/job.go
tgross: > I dig ActiveDeploymentID 👍 or I could make it a bool ActiveDeployment, or similar? Phil do you imagine that there will actually be any use for the ID, now or later?

I'd have it be the ID because if it's at all relevant that there _is_ a deployment, we're going to want to go looking for information about it anyways. That'll save us an API call.

> for the indexes, do you mean their ModifyIndex? I'm happy to include that or any other fields on here, but so far have been trying to keep it to "only what Phil needs" because it's easy to add stuff later, harder to take away if users start using it for their own purposes.

Yes, but that's totally sensible

philrenaud: > Phil do you imagine that there will actually be any use for the ID, now or later

Yep I could see us wanting that as a quick way to link to a deployment directly!

nomad/jobs_endpoint.go
gulducat: turns out this is a bit funky.

I can do a little dance like this:

```go
i := 0
jobs := make([]structs.UIJob, len(args.Jobs))
newJobs := set.New[structs.NamespacedID](len(args.Jobs))

paginator..... {

	if len(args.Jobs) > 0 {
		jobs[i] = uiJob
		i++
	} else {
		jobs = append(jobs, uiJob)
	}
```

it feels strange, but it does work, _except_ when the caller asks for jobs that don't exist -- that ends up with zero-value `UIJob`s for each missing job (I could make them pointers, but missing jobs would show up in the results, as `nil`).  so I'd need to iterate over the results and exclude any empty ones, at which point I feel I've gone a little overboard in my optimization efforts.

maybe there's a trick here that I'm not aware of?

gulducat: I put that experiment in [3e0ab67](https://github.com/hashicorp/nomad/pull/20130/commits/3e0ab6718e5eda8275bc399f53c2a423393cc221), but was not a fan, so reverted.

let me know if you think it's worth it, or I should try it differently!

nomad/jobs_endpoint.go
gulducat: done in [1f593d9](https://github.com/hashicorp/nomad/pull/20130/commits/1f593d937dc49e887ab73311af9c31084c6a3823)

nomad/structs/job.go
gulducat: "smart" stuff removed in [f8ccbd0](https://github.com/hashicorp/nomad/pull/20130/commits/f8ccbd02da0141a2c3f3d18ae6b0635bd730178d)

nomad/structs/job.go
gulducat: we ended up going with a stubby `LatestDeployment` struct that contains more info, since I have it in hand already in the backend, saving the UI from needing to make a separate call to gather anything at all about the deployment. => [b681e8b](https://github.com/hashicorp/nomad/pull/20130/commits/b681e8b8ac207d1be8cb9bc4cad1086b11475bc1)

```shell
$ nomad operator api /v1/jobs/statuses | jq '.[].LatestDeployment'
```

```json
{
  "AllAutoPromote": false,
  "ID": "dedfb786-33b0-fcb0-10bf-1b6c63954fd0",
  "IsActive": false,
  "JobVersion": 0,
  "RequiresPromotion": false,
  "Status": "successful",
  "StatusDescription": "Deployment completed successfully"
}
```

nomad/jobs_endpoint.go
gulducat: [4447725](https://github.com/hashicorp/nomad/pull/20130/commits/4447725dcaee992fb8660229d0c79c20079f2ea3)

command/agent/job_endpoint.go
gulducat: in [924935a](https://github.com/hashicorp/nomad/pull/20130/commits/924935a54f0882fd1fe92dddf5481d2bac5943a6) I treat GET and POST the same, but I still want to error if they do give a request body, but it is malformed.

api/jobs.go
gulducat: "smart" stuff removed in [f8ccbd0](https://github.com/hashicorp/nomad/pull/20130/commits/f8ccbd02da0141a2c3f3d18ae6b0635bd730178d)

but now there's an `include_children` option to which this comment still applies.

so I should either make this more complete and proper, or alternatively, exclude it from the `api` package altogether?  perhaps that could be a follow-on...  as it stands, this new endpoint is primarily for the UI to consume.

gulducat: `IncludeChildren` is on this struct now, and split out `NamespacedID` for the "TODO: proper type" comment, in [5e641a5](https://github.com/hashicorp/nomad/pull/20130/commits/5e641a56c6ed5724d5ca06b23ffaa8311fe39d0b)

nomad/mock/alloc.go
```diff
@@ -87,14 +87,19 @@ func Alloc() *structs.Allocation {
 }
 
 func MinAlloc() *structs.Allocation {
-	job := MinJob()
+	return MinAllocForJob(MinJob())
+}
+
+func MinAllocForJob(job *structs.Job) *structs.Allocation {
```
gulducat, line 93: I could have sworn I split this out to a separate PR, like I did several other things, but I guess not.  think I should?

nomad/state/schema.go
```diff
@@ -254,6 +254,15 @@ func jobTableSchema() *memdb.TableSchema {
 					Field: "NodePool",
 				},
 			},
+			// ModifyIndex allows sorting by last-changed
+			"modify_index": {
```
gulducat, line 258: ditto here. should I split out to be able to backport?

philrenaud (approved):

nomad/mock/alloc.go
tgross: Pretty trivial. Fine to leave in.

nomad/state/schema.go
tgross: Same, fine to leave in

command/agent/job_endpoint_statuses.go
```diff
@@ -0,0 +1,79 @@
...
+
+	switch req.Method {
+	case http.MethodGet, http.MethodPost:
+		break
+	default:
+		return nil, CodedError(http.StatusMethodNotAllowed, ErrInvalidMethod)
+	}
+
+	if includeChildren, err := parseBool(req, "include_children"); err != nil {
+		return nil, err
```
tgross, line 31: Might be nice if we could wrap this in `CodedError(http.StatusBadRequest`

command/agent/job_endpoint_statuses.go
```diff
@@ -0,0 +1,79 @@
...
+	} else if includeChildren != nil {
+		args.IncludeChildren = *includeChildren
+	}
+
+	// ostensibly GETs should not accept structured body, but the HTTP spec
+	// on this is more what you'd call "guidelines" than actual rules.
```
tgross, line 37: :laughing:

command/agent/job_endpoint_statuses.go
```diff
@@ -0,0 +1,79 @@
...
+		// if all requested jobs turn out to have the same namespace,
+		// then the RPC endpoint will notice that and override this anyway.
+		if args.QueryOptions.Namespace == structs.DefaultNamespace {
+			args.QueryOptions.Namespace = structs.AllNamespacesSentinel
+		}
+
+		args.Jobs = make([]structs.NamespacedID, len(in.Jobs))
+		for i, j := range in.Jobs {
+			if j.Namespace == "" {
+				j.Namespace = "default"
```
tgross, line 58:
```suggestion
				j.Namespace = structs.DefaultNamespace
```

nomad/job_endpoint_statuses.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+	defer metrics.MeasureSince([]string{"nomad", "jobs", "statuses"}, time.Now())
+
+	namespace := args.RequestNamespace()
+	// the namespace from the UI by default is "*", but if specific jobs are
+	// requested, all with the same namespace, AllowNsOp() below may be able
+	// to quickly deny the request if the token lacks permissions for that ns,
+	// rather than iterating the whole jobs table and filtering out every job.
+	if len(args.Jobs) > 0 {
+		nses := set.New[string](0)
```
tgross, line 41: Incredibly minor optimization: we know because we have non-zero jobs that we always have at least one namespace. Setting the initial value to 0 allocates the map and then _always_ immediately has to go allocate the extra space anyways. We can save a malloc call by just giving us a minimal size up front:

```suggestion
		nses := set.New[string](1)
```

nomad/job_endpoint_statuses_test.go
```diff
@@ -0,0 +1,382 @@
...
+			deleteJob(t, job)
+		}
+		t.Cleanup(cleanup)
+		return job, cleanup
+	}
+
+	// this little cutie sets the latest state index to a predictable value,
+	// to ensure the below jobs span the boundary from 999->1000 which would
+	// break pagination without proper uint64 NextToken (ModifyIndex) comparison
+	must.NoError(t, s.State().UpsertNamespaces(996, nil))
```
tgross, line 108: Very clever, and good explanatory comment! :+1:

nomad/job_endpoint_statuses_test.go
```diff
@@ -0,0 +1,382 @@
...
+		// start the query
+		// note: queries that are expected to remain blocked leak this goroutine
+		// unless some other test coincidentally frees it up
+		go func() {
+			resp = doRequest(t, req)
+			cancel()
+		}()
+		// give it a moment for the rpc to actually start up
+		// FLAKE ALERT: if this job is flaky, this might be why.
+		time.Sleep(time.Millisecond * 10)
```
tgross, line 242: This is a little unclear. What are we waiting for here? The goroutine to start up? The request to get sent? I'd probably try to gate this with `sync.WaitGroup` instead, or have the function return a channel that the goroutine writes the response onto.

tgross (approved):
This looks great @gulducat! 

I've left a few minor comments, but the only one that really raises an eyebrow is the flaky test warning. I'd be open to fixing that in a separate PR though. Also, we'll need to add API docs. That can be a separate PR as well.

nomad/job_endpoint_statuses_test.go
gulducat: the main test that flakes without this is "delete job" (and "delete job on page"), which is supposed to unblock when a job is deleted.  I am fairly sure there is a race between these two things:

* start the blocking query
* delete the job

if the job gets deleted before the query starts blocking, then nothing triggers it to _unblock_, so the test fails.  so what this is waiting for is "everything leading up to blocking on the `WatchSet`" in `rpcHandler.blockingRPC()` (the very last thing in the RPC).  I'm not sure how to slide something in there to signal back to me in the test that it's ready, and 10ms seemed an acceptable sacrifice. :shrug: 

I suppose it doesn't crop up in other `mustUnblock` tests (like "new job") because those operations are slow enough to consistently give the RPC time to start blocking, but I just apply this lil nap across the board.

I could instead sleep down in `t.Run()` between `startQuery()` and `tc.run()` (which deletes the job), but I put it here so it would be right next to the goroutine that spurred me to add it.

phew!  if you have a suggestion for how to boil that down into a tidier comment, I'm happy to hear it! :joy:

gulducat: I added just a little extra here https://github.com/hashicorp/nomad/pull/20130/commits/1b4ce0a5e289fc8c40ffc2c7512e20fcf7556c6c#diff-e1afa22d48b0304823b263b1a4831ec09d3d39b5e9102d2b41535fed4cb5fa5eR240

> // give it a moment for the rpc to actually start up _and begin blocking_

tgross: Ah, I see what you're doing, and that makes sense. What we've typically done for this in other tests is to put the mutation behind `time.AfterFunc` and then making the blocking query. Ex. in `namespace_endpoint_test.go` you'll see something like:


```go
	// Namespace delete triggers watches
	time.AfterFunc(100*time.Millisecond, func() {
		must.NoError(t, state.DeleteNamespaces(300, []string{ns.Name}))
	})

	req.QueryOptions.MinQueryIndex = 250
	var resp structs.SingleNamespaceResponse
	must.NoError(t, msgpackrpc.CallWithCodec(codec, "Namespace.GetNamespace", req, &resp))
```

That handles both ensuring the index is what you expect and that the delay is long enough. What you've got here is fine, although I'd consider bumping the time up quite a bit because CI machines can have a _lot_ of steal and these take longer than you'd expect.

PR #20134, opened 2024-03-14
template: improve internal documentation around shutdown

tgross:
While investigating a report around possible consul-template shutdown issues, which didn't bear fruit, I found that some of the logic around template runner shutdown is unintuitive.

* Add some doc strings to the places where someone might think we should be obviously stopping the runner or returning early.
* Mark context argument for `Poststart`, `Stop`, and `Update` hooks as unused.

No functional code changes.


schmichael (approved):
:bow:

PR #20147, opened 2024-03-18
Add docs for disconnected block

Juanadelacuesta:
This PR introduces the documentation for the now disconnect block on the job configuration.

### DIFF
--- a/website/content/docs/job-specification/disconnect.mdx
+++ b/website/content/docs/job-specification/disconnect.mdx
@@ -11,7 +11,7 @@ description: |-
 
 <Placement groups={['job', 'group', 'disconnect']} />
 
-The `disconnect` block describes the systm's behaviour in case of a network 
+The `disconnect` block describes the system's behavior in case of a network 
 partition.
 
 ```hcl
--- a/website/content/docs/job-specification/disconnect.mdx
+++ b/website/content/docs/job-specification/disconnect.mdx
@@ -63,7 +63,7 @@ partition.
   the previously disconnected node regains connectivity.
   It has four possible values which are described below:
 
-    - `keep_original`: Always keep the original allocation, bare in mind 
+    - `keep_original`: Always keep the original allocation. Bear in mind 
     when choosing this option, it can have crashed while the client was 
     disconnected.
     - `keep_replacement`: Always keep the allocation that was rescheduled 
--- a/website/content/docs/job-specification/disconnect.mdx
+++ b/website/content/docs/job-specification/disconnect.mdx
@@ -197,6 +197,6 @@ group "second" {
 ```
 
 [`heartbeat_grace`]: /nomad/docs/configuration/server#heartbeat_grace
-[`stop_after`] : /nomad/docs/job-specification/disconnect#stop_after
-[`lost_after`] : /nomad/docs/job-specification/disconnect#replace_after
+[`stop_after`]: /nomad/docs/job-specification/disconnect#stop_after
+[`lost_after`]: /nomad/docs/job-specification/disconnect#replace_after
 [`reconcile`]: /nomad/docs/job-specification/disconnect#reconcile
--- a/website/content/docs/job-specification/disconnect.mdx
+++ b/website/content/docs/job-specification/disconnect.mdx
@@ -1,0 +1,204 @@
+---
+layout: docs
+page_title: disconnect Block - Job Specification
+description: |-
+  The "disconnect" block describes the behaviuor of both the Nomad server and 
+  client in case of a network partition, as well as how to reconcile the workloads
+  in case of a reconnection.
+---
+
+# `disconnect` Block
+
+<Placement groups={['job', 'group', 'disconnect']} />
+
+The `disconnect` block describes the systm's behaviour in case of a network 
+partition.
+
+```hcl
+    job "docs" {
+  	    group "example" {
+    	    disconnect {
+		        lost_after = "6h"
+		        stop_after = "2h"
+		        replace = false
+                reconcile = "keep_original"
+     	    }
+  	    }
+    }
+
+```
+
+## `disconnect` Parameters
+
+- `lost_after` `(string: "")` - Specifies a duration during which a Nomad client
+  will attempt to reconnect allocations after it fails to heartbeat
+  in the [`heartbeat_grace`] window.  It defaults to “” which is equivalent to 
+  having the disconnect block be nil.
+  
+  See [the example code below][lost_after] for more details. This setting cannot
+  be used with [`stop_after`].
+
+- `replace` `(bool: false)` - Specifies if the disconnected allocation should 
+  be replaced by a new one rescheduled on a different node. If false and the 
+  node it is running on becomes disconnected or goes down, this allocations
+  wont be rescheduled and will show up as `unknown` until the node comes back 
+  up or it are manually stoped:
+
+  ```plaintext
+  `nomad alloc stop  <alloc ID>`
+  ```
+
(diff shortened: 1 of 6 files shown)


website/content/docs/job-specification/disconnect.mdx
```diff
@@ -0,0 +1,202 @@
+---
+layout: docs
+page_title: disconnect Block - Job Specification
+description: |-
+  The "disconnect" block describes the behaviuor of both the Nomad server and
```
tgross, line 5:
```suggestion
  The "disconnect" block describes the behavior of both the Nomad server and 
```

website/content/docs/job-specification/disconnect.mdx
```diff
@@ -0,0 +1,202 @@
...
+  The "disconnect" block describes the behaviuor of both the Nomad server and 
+  client in case of a network partition, as well as how to reconcile the workloads
+  in case of a reconnection.
+---
+
+# `disconnect` Block
+
+<Placement groups={['job', 'group', 'disconnect']} />
+
+The `disconnect` block describes the systm's behaviour in case of a network
```
tgross, line 14:
```suggestion
The `disconnect` block describes the system's behavior in case of a network 
```

website/content/docs/job-specification/disconnect.mdx
```diff
@@ -0,0 +1,202 @@
...
+  	    }
+    }
+
+```
+
+## `disconnect` Parameters
+
+- `lost_after` `(string: "")` - Specifies a duration during which a Nomad client
+  will attempt to reconnect allocations after it fails to heartbeat
+  in the [`heartbeat_grace`] window.  It defaults to “” which is equivalent to
```
tgross, line 35: Some syntax highlighting of markdown will only pick up this as a link if you have a trailing empty bracket here. It's not required for rendering, but it's a nice-to-have:

```suggestion
  in the [`heartbeat_grace`][] window.  It defaults to "" which is equivalent to 
```

website/content/docs/job-specification/disconnect.mdx
```diff
@@ -0,0 +1,202 @@
...
+  having the disconnect block be nil.
+  
+  See [the example code below][lost_after] for more details. This setting cannot
+  be used with [`stop_after`].
+
+- `replace` `(bool: false)` - Specifies if the disconnected allocation should 
+  be replaced by a new one rescheduled on a different node. If false and the 
+  node it is running on becomes disconnected or goes down, this allocations
+  wont be rescheduled and will show up as `unknown` until the node comes back 
+  up or it are manually stoped:
```
tgross, line 45:
```suggestion
  node it is running on becomes disconnected or goes down, this allocation
  won't be rescheduled and will be reported as `unknown` until the node reconnects, 
  or until the allocation is manually stopped:
```

(I think that's what we mean in that last part?)

website/content/docs/job-specification/disconnect.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+  ```plaintext
+  `nomad alloc stop  <alloc ID>`
+  ```
+
+  If true, a new  alloc will be placed immediately upon the node becoming
```
tgross, line 51:
```suggestion
  If true, a new alloc will be placed immediately upon the node becoming 
```

website/content/docs/job-specification/disconnect.mdx
```diff
@@ -0,0 +1,202 @@
...
+  `nomad alloc stop  <alloc ID>`
+  ```
+
+  If true, a new  alloc will be placed immediately upon the node becoming 
+  disconnected.
+
+- `stop_after` `(string: "")` - Specifies a duration after which a disconnected 
+  Nomad client will stop its allocations. Keep in mind that if stop_after is 
+  shorter than ttl, and `replace` = false, can lead to a state where no allocations
+  are running, hence it is not admitted and will cause a validation error.
```
tgross, line 57: We're describing the "why" before the "what" here. I'd reword this to make it clear at the start of the sentence that the configuration described isn't allowed, and _then_ describe why. Also, I'm not sure what "ttl" we're referring to here?

```suggestion
  Nomad client will stop its allocations. Setting `stop_after` shorter than 
  ??? and `replace = false` at the same time is not permitted and will cause a 
  validation error, because this would lead to a state where no allocations 
  can be scheduled.
```

website/content/docs/job-specification/disconnect.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+  The Nomad client process must be running for this to occur. This setting
+  cannot be used with [`lost_after`].
+
+- `reconcile` `(string: “best_score”)` - Specifies which allocation to keep once
```
tgross, line 62: We probably don't want "smart quotes" anywhere, as I'm not sure the HCL parser will accept them if someone copy-pastes:

```suggestion
- `reconcile` `(string: "best_score")` - Specifies which allocation to keep once
```

website/content/docs/job-specification/disconnect.mdx
```diff
@@ -0,0 +1,202 @@
...
+  the previously disconnected node regains connectivity.
+  It has four possible values which are described below:
+
+    - `keep_original`: Always keep the original allocation, bare in mind
```
tgross, line 66:
```suggestion
    - `keep_original`: Always keep the original allocation. Bear in mind 
```

website/content/docs/job-specification/disconnect.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+### Lost After
+
+`lost_after` specifies a duration during which a Nomad client will
+attempt to reconnect allocations after it fails to heartbeat in the
+[`heartbeat_grace`] window.
+
+By default, allocations running on a client that fails to heartbeat will be
+marked "lost". When a client reconnects, its allocations, which may still be
+healthy, will restart because they have been marked "lost". This can cause
```
tgross, line 132: Will they restart or will they have been rescheduled, so they stop?

Also, I would consider moving this entire paragraph somewhere near the top of this page, so that we have a clear description of the default behavior before describing how `disconnect` changes it.

website/content/docs/upgrade/upgrade-specific.mdx
```diff
@@ -14,6 +14,12 @@ their upgrades as a result of new features or changed behavior. This page is
 used to document those details separately from the standard upgrade flow.
 
 ## Nomad 1.8.0
+Nomad 1.8.0 introduces a `disconnect` block meant to group all the configuration
+options related to disconnected client's and server's behavior, causing the 
+deprecation of the fileds `stop_after_client_disconnect`, `max_client_disconnec`
```
tgross, line 19:
```suggestion
deprecation of the fileds `stop_after_client_disconnect`, `max_client_disconnect`
```

website/content/docs/job-specification/disconnect.mdx
```diff
@@ -0,0 +1,199 @@
...
+  client in case of a network partition, as well as how to reconcile the workloads
+  in case of a reconnection.
+---
+
+# `disconnect` Block
+
+<Placement groups={['job', 'group', 'disconnect']} />
+
+The `disconnect` block describes the system's behavior in case of a network 
+partition.
```
tgross, line 15: Let's put this right at the very top, so it's clear what this block is for.

```suggestion
The `disconnect` block describes the system's behavior in case of a network 
partition. By default, without a `disconnect` block, if an allocation is on a 
node that misses heartbeats, the allocation will be marked `lost` and will be 
rescheduled.
```

tgross (approved):
LGTM

PR #20150, opened 2024-03-18
acl: remove unused nil ACL object handling

tgross:
As of #18754 which shipped in Nomad 1.7, we no longer need to nil-check the object returned by `ResolveACL` if there's no error return, because in the case where ACLs are disabled we return a special "ACLs disabled" ACL object. Checking nil is not a bug but should be discouraged because it opens us up to future bugs that would bypass ACLs.

While working on an unrelated feature @lindleywhite discovered that we missed removing the nil check from several endpoints with our semgrep linter. This changeset fixes that.

Co-Author: Lindley <lindley@hashicorp.com>


lindleywhite (approved):

shoenig (approved):
LGTM

Issue #19718, opened 2024-01-11
flakey test TestOperator_TransferLeadershipToServerAddress_ACL


shoenig:
I've noticed this test fail a few times, and just now failed 3 times in a row causing a build failure.

```
 === RUN   TestOperator_TransferLeadershipToServerAddress_ACL/good-token
    assert.go:14: 
        operator_endpoint_test.go:516: expected nil error
        ↪ error: Permission denied
--- FAIL: TestOperator_TransferLeadershipToServerAddress_ACL/good-token (0.00s)
FAIL nomad.TestOperator_TransferLeadershipToServerAddress_ACL/good-token (re-run 1) (0.00s)
=== RUN   TestOperator_TransferLeadershipToServerAddress_ACL
    operator_endpoint_test.go:433: bootstrap token: {ea59fd08-c7f2-bf20-f8f5-96e70e3984fe ec80755e-5092-17d1-fe18-b3cc737d6441 management 87b00b9c-f7e7-416f-f25b-1012634d7dcc management [] [] true [] 2024-01-11 19:15:49.532296339 +0000 UTC <nil> 0s 1 1}
Error: 1 2024-01-11T19:15:50.032Z [ERROR] raft-net: failed to decode incoming command: error="transport shutdown"
Error: 1 2024-01-11T19:15:50.117Z [ERROR] raft-net: failed to decode incoming command: error="transport shutdown"
--- FAIL: TestOperator_TransferLeadershipToServerAddress_ACL (0.75s)
FAIL nomad.TestOperator_TransferLeadershipToServerAddress_ACL (re-run 1) (0.75s)
coverage: 10.4% of statements
FAIL nomad (coverage: 10.4% of statements)
```

tgross:
I'll close the dupe of this over here: https://github.com/hashicorp/nomad/issues/18993

shoenig:
Oh whoops, should have checked first :goat:

PR #20154, opened 2024-03-18
[chore] Fix LeadershipTransfer tests

angrycub:
Multi-node Nomad clusters under test must use the RPC calls to bootstrap the ACL subsystem. The original implementation of `testcluster` tried naively supersizing the single node behavior in `TestRaftRemovePeer`, which mutates the single node's state directly. In the case of a multi-node cluster, the RPC calls are necessary to ensure that the data is replicated via Raft to all of the cluster members.

Fixes: https://github.com/hashicorp/nomad/issues/19718

### DIFF
--- a/nomad/operator_endpoint_test.go
+++ b/nomad/operator_endpoint_test.go
@@ -14,6 +14,7 @@ import (
 	"path"
 	"reflect"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -378,10 +379,10 @@ func TestOperator_RaftRemovePeerByID_ACL(t *testing.T) {
 
 type testcluster struct {
 	t       *testing.T
+	args    tcArgs
 	server  []*Server
 	cleanup []func()
 	token   *structs.ACLToken
-	rpc     func(string, any, any) error
 }
 
 func (tc testcluster) Cleanup() {
@@ -406,6 +407,7 @@ func newTestCluster(t *testing.T, args tcArgs) (tc testcluster) {
 	cSize := args.size
 	out := testcluster{
 		t:       t,
+		args:    args,
 		server:  make([]*Server, cSize),
 		cleanup: make([]func(), cSize),
 	}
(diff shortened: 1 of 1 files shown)


nomad/operator_endpoint_test.go
```diff
@@ -450,70 +472,119 @@ func (tc testcluster) leader() *Server {
 	}
 	return nil
 }
+func (tc testcluster) anyFollowerRaftServerID() raft.ServerID {
```
schmichael, line 475: This is an improvement over the old test, so no need to change it.

However I have a slight preference for just putting this code directly into the test instead of creating helper funcs. The helpers are only used once and are susceptible to logical races with "organic" leader elections caused by fun things like slow CI. This particular test would have a hard time avoiding that, but I don't want to encourage more tests assuming raft membership roles are static.

schmichael (approved):
:bow:

nomad/operator_endpoint_test.go
schmichael: (Adding more detail after an internal chat)

My problem with this style of helper func is that they encourage us developers to think of raft membership roles as _static_ (until we explicitly change it at least). However, leadership elections can happen at any time even during tests (slow CI builders love to trip timeouts that never happen locally).

The peerLock that is acquired prevents a _data_ race with raft membership, but not a _logical_ race. Raft can still change membership while a single peerLock is acquired. This means that the return value of these helpers might be stale or become stale by the time they're used.

For the particular test below I think that's fine. I can only think of 2 ways logical races could impact this test given a cluster of (S1: L, S2: F, S3: F):

1. anyFollower -> returns S2
2. S1 times out and S2 is elected leader "naturally"
3. TransferLeadershipTo(S2) -> Noops
4. Test passes!

So we have a false positive: the test passed but didn't really exercise the code we intended it to. Leadership was never transferred. We could detect this by looking at the Noop value of the RPC's return value and... fail? retry? Hardly seems worth it since this test is run many times and should rarely execute the incorrect code path.

The other case:

1. anyFollower -> returns S2
2. TransferLeadershipTo(S2) -> Success!
3. S2 times out and S3 is elected leader "naturally"
4. Test fails!

Oh no a false negative! Definitely bad, but the timeout would have to occur in a very small window of time. Seems incredibly unlikely, and if it's not unlikely the test will keep being flaky and we'll revisit it then. :sweat_smile: 

I can't think of a clever way to handle this false negative without a ton of effort that introduces more risks than its worth.

So the moral of the story is that these helpers cannot guarantee the freshness/accuracy of their return values. That's a pretty dangerous thing for an innocent looking helper. Few tests should care, and that's why I lean toward just hand writing the code in the 2 tests that use these ...or adding a comment or renaming the struct to newOperatorTestCluster or something to reflect its bespoke nature... just anything to avoid future us thinking these methods provide more accuracy than they actually can.

Issue #20164, opened 2024-03-19
default `client.template.wait` configuration is not passed to template runner


tgross:
The documentation for [`client.template.wait`](https://developer.hashicorp.com/nomad/docs/configuration/client#wait) says that the default value is `{ min = "5s" max = "4m" }` but while debugging https://github.com/hashicorp/nomad/issues/20163 I discovered that the default value is not being passed into the template runner. This potentially results in extra load to upstream dependencies, in particular Vault which doesn't have blocking queries.

PR #20165, opened 2024-03-19
config: fix `client.template` config merging with defaults

tgross:
When loading the client configuration, the user-specified `client.template` block was not properly merged with the default values. As a result, if the user set any `client.template` field, all the other field defaulted to their zero values instead of the documented defaults.

This changeset:
* Adds the missing `Merge` method for the client template config and ensures it's called.
* Makes a single source of truth for the default template configuration, instead of two different constructors.
* Extends the tests to cover the merge of a partial block better.

Fixes: https://github.com/hashicorp/nomad/issues/20164

### DIFF
--- a/client/config/config.go
+++ b/client/config/config.go
@@ -527,12 +527,9 @@ func (c *ClientTemplateConfig) Merge(o *ClientTemplateConfig) *ClientTemplateCon
 	if o.DisableSandbox {
 		result.DisableSandbox = true
 	}
-	if o.MaxStale != nil {
-		result.MaxStale = &*o.MaxStale
-	}
-	if o.BlockQueryWaitTime != nil {
-		result.BlockQueryWaitTime = &*o.BlockQueryWaitTime
-	}
+
+	result.MaxStale = pointer.Merge(result.MaxStale, o.MaxStale)
+	result.BlockQueryWaitTime = pointer.Merge(result.BlockQueryWaitTime, o.BlockQueryWaitTime)
 
 	if o.Wait != nil {
 		result.Wait = c.Wait.Merge(o.Wait)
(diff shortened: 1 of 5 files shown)


client/config/config.go
```diff
@@ -485,6 +508,58 @@ func (c *ClientTemplateConfig) IsEmpty() bool {
...
+		funcSet := set.From(result.FunctionDenylist)
+		funcSet.InsertSlice(o.FunctionDenylist)
+		result.FunctionDenylist = funcSet.Slice()
+	}
+
+	if len(o.FunctionBlacklist) > 0 {
+		funcSet := set.From(result.FunctionBlacklist)
+		funcSet.InsertSlice(o.FunctionBlacklist)
+		result.FunctionBlacklist = funcSet.Slice()
+	}
```
schmichael, line 531: Should we just set `result.FunctionDenylist = slices.Copy(o.FunctionDenylist)`? Making merge the union makes it impossible to allow default denied funcs.

Juanadelacuesta (approved):

client/config/config.go
tgross: Good catch, will fix.

tgross: Done in [3296220](https://github.com/hashicorp/nomad/pull/20165/commits/3296220865af93364468ba0160695abb63948da3) along with a test exercising the difference in behavior between `function_deny_list = []` and being unset.

client/config/config.go
```diff
@@ -485,6 +507,53 @@ func (c *ClientTemplateConfig) IsEmpty() bool {
 		c.NomadRetry.IsEmpty()
 }
 
+func (c *ClientTemplateConfig) Merge(o *ClientTemplateConfig) *ClientTemplateConfig {
```
shoenig, line 510: We have https://github.com/hashicorp/nomad/blob/main/helper/pointer/pointer.go#L31-L37 to help clean up these pointer merges

PR #20166, opened 2024-03-20
agent: allow configuration of in-memory telemetry sink.

jrasell:
This change adds configuration options for setting the in-memory telemetry sink collection and retention durations. This sink backs the metrics JSON API and previously had hard-coded default values.

The new options are particularly useful when running development or debug environments, where metrics collection is desired at a fast and granular rate. This is particularly useful for gauges; currently the last value written in the interval is returned from the API. In situations with spiking/flapping data, this could mean we miss information.


command/agent/config_parse_test.go
```diff
@@ -196,15 +196,19 @@ var basicConfig = &Config{
 		},
 	},
 	Telemetry: &Telemetry{
```
tgross, line 198: Judging from https://github.com/hashicorp/nomad/pull/20165 it's unfortunately easy to break config parsing with our default values. Should we add a test case that covers the "non-nil `Telemetry` block but with `in_memory_collection_interval` and `in_memory_retention_period` unset?

command/agent/command.go
```diff
@@ -1165,6 +1164,22 @@ func (c *Command) setupTelemetry(config *Config) (*metrics.InmemSink, error) {
 		telConfig = config.Telemetry
 	}
 
+	// The values are defaulted in both the dev and default config object, so
+	// we should never have empty values. Do this, so it can never happen and
+	// cause agents to panic on start.
+	inMemInt := 10 * time.Second
+	inMemRet := 1 * time.Minute
```
tgross, line 1171: I get that we're trying to be belt-and-suspenders here but this leaves two different places where the default value is being defined. Is there a reason we expect the default/dev config objects not to be enough?

jrasell: In my opinion no, the default/dev config objects should be enough as we always overlay user passed config options on a default config. The reason I put this in is because the telemetry setup also includes code to handle a nil telemetry config object, which made me suspicious of why it was added. I am happy to remove this, the nil check is potentially a remanent of code added ~10 years ago and not touched since.

command/agent/config_parse_test.go
jrasell: Added something in: https://github.com/hashicorp/nomad/pull/20166/commits/a94d996cd8542922b1a9d0636e41d52f5da078d4

Is this what you had in mind?

tgross: Perfect!

command/agent/command.go
lgfa29: Yeah, I can't think of a way we could end up with a `nil` here. Even with an explicitly empty block the default config is merged into it.

Any misconfiguration at this point would also fail loudly and early right? Meaning the agent wouldn't start and no state would be modified, so even if an error does happen it seems like it would fail safely.

If we do want to be extra safe here we could instead have a `DefaultTelemetry()` function that can be called in both `DefaultConfig` and inside the `if` above.

command/agent/config.go
```diff
@@ -1072,6 +1086,24 @@ func (a *Telemetry) PrefixFilters() (allowed, blocked []string, err error) {
...
+
+	// Ensure the in-memory durations are set correctly.
+	if t.inMemoryRetentionPeriod > 0 && t.inMemoryCollectionInterval > 0 {
+		if t.inMemoryCollectionInterval > t.inMemoryRetentionPeriod {
+			return errors.New("telemetry in-memory collection interval cannot be greater than retention period")
+		}
+	}
+
+	return nil
+}
```
lgfa29, line 1105: Should we check for 0 values here? From my quick test it seems that a `0` uses the default value. Not sure if that's expected though.

lgfa29 (approved):

command/agent/config.go
jrasell: Added within: https://github.com/hashicorp/nomad/pull/20166/commits/308f0a1fa64b1331d8b3d5de0e1622196b97abc1

tgross (approved):
LGTM!

Issue #19785, opened 2024-01-19
Consul Upstream Admin Partition Support


mmcquillan:
### Proposal

Nomad does not expose `destination_partition` in the Consul SM upstreams block as of 1.7.2. We should do so.

---

As a possible workaround, has anyone successfully tested/used the service-defaults config entry to override the upstream definition? I’m thinking something like:

```
consul config write - <<EOF
Kind = "service-defaults"
Name = "dummy"
Partition = "gamma"
Namespace = "default"
UpstreamConfig = {
  Overrides = {
    Name = "counting"
    Partition = "default"
    Namespace = "default"
  }
}
```

```
# nomad job spec snippet
    ...
    service {
      name = "dummy"
      port = "9002"

      connect {
        sidecar_service {
          proxy {
            upstreams {
              destination_name = "counting"
              local_bind_port  = 8080
            }
          }
        }
      }
    }
```

PR #20167, opened 2024-03-20
jobspec: add support for destination partition to `upstream` block

tgross:
Adds support for specifying a destination Consul admin partition in the `upstream` block.

Fixes: https://github.com/hashicorp/nomad/issues/19785


gulducat (approved):
LGTM!

PR #20173, opened 2024-03-20
Upgrade go-msgpack to v2

schmichael:
Upgraded to go-msgpack v2. Replaces #18812 because I'm better at writing new PRs than I am at resolving merge conflicts.

Manually tested with a set of 3 servers at different versions: this PR, 1.7.6, and 1.6.9. Ran a job on a 1.5.2 client agent just to really try and keep things wacky. Didn't notice anything unusual.

Please peek at `go.mod` and 772d35fa5748b812a65368d85bf4fd6d7de44828 in particular. Everything else is mechanically changed. See the commit messages on d467ab55584426afb86dca7f41cb25516bb00762 and 7e353e07c5d112d96ef28a2c34c420b302b71c16 if you want to mock my ham fisted import rewriting.


swenson (commented):
I think everything here looks good, but there still appears to be a reference to go-msgpack 1.1.5 in `GNUmakefile`

schmichael:
@swenson great catch, thanks. Added in 6328341e957908ec72dc00d6d3ae1e57a57f0dae

swenson (approved):
LGTM

shoenig:
Just to make sure I understand, we _are_ going to go through the time encoding change due to setting 

```
MsgpackUseNewTimeFormat: true,
```

but this is okay, because the decoders on the other end in `v2` can handle both cases?

```
// MsgpackUseNewTimeFormat when set to true, force the underlying msgpack
	// codec to use the new format of time.Time when encoding (used in
	// go-msgpack v1.1.5 by default). Decoding is not affected, as all
	// go-msgpack v2.1.0+ decoders know how to decode both formats.
	MsgpackUseNewTimeFormat [bool](https://pkg.go.dev/builtin#bool)
```

if so LGTM

shoenig (approved):

schmichael:
> Just to make sure I understand, we are going to go through the time encoding change due to setting

Not quite. Nomad was already on msgpack-1.1.5, so we were already using the "New" time encoding.

> but this is okay, because the decoders on the other end in v2 can handle both cases?

This would *not* be upgrade safe, so not ok. If we had been on the msgpack-0.5.5 encoding then during upgrades the old agents would not understand times from new agents. Luckily we were on msgpack-1.1.5 before.

Issue #20174, opened 2024-03-20
Nomad 1.7.6 does not pick up Docker DNS settings


Jess3Jane:
### Nomad version
Version in which this functionality is broken:
```bash
root@client-1:/# nomad version
Nomad v1.7.6
BuildDate 2024-03-12T07:27:36Z
Revision 594fedbfbc4f0e532b65e8a69b28ff9403eb822e
```

Version in which this functionality is working (I tested down to 1.7.2 and they all behave the same):
```bash
root@client-1:/# nomad version
Nomad v1.7.5
BuildDate 2024-02-13T15:10:13Z
Revision 5f5d4646198d09b8f4f6cb90fb5d50b53fa328b8
```

### Operating system and Environment details

```bash
root@client-1:/# uname -a
Linux client-1 5.15.0-67-generic #74-Ubuntu SMP Wed Feb 22 14:14:39 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
root@client-1:/# lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.2 LTS
Release:	22.04
Codename:	jammy
```

This is a completely fresh digital ocean node (I will note all of the changes I made below). Of note we have also hit this issue on systems of varying other configurations (all Ubuntu 22.04 x64).

### Issue

[Docker allows you to configure a set of DNS servers to give to each container](https://docs.docker.com/network/#dns-services). In Nomad 1.7.5 if you did not set the DNS settings for a job using the docker driver the containers in that job would have this DNS configuration. In Nomad 1.7.6 the containers instead have the host's default DNS configuration which is often undesirable.

### Reproduction steps

1. Spin up a fresh Ubuntu 22.04 server (I used a Digital Ocean droplet for our reproduction but we've noticed this happening across our fleet so I don't think they're doing anything weird).
2. Install docker-ce as per [their docs](https://docs.docker.com/engine/install/ubuntu/#install-using-the-repository) (I used Docker's apt registry to install it).
3. Install Nomad as per [the docs](https://developer.hashicorp.com/nomad/docs/install) (for the reproduction I specifically used the version of Nomad from Hashicorps repos).
4. Install the base CNI plugins by placing the contents of `https://github.com/containernetworking/plugins/releases/download/v1.0.0/cni-plugins-linux-amd64-v1.0.0.tgz` into `/opt/cni/bin`
[...]

PR #20189, opened 2024-03-22
cni: fix regression in falling back to DNS owned by `dockerd`

tgross:
In #20007 we fixed a bug where the DNS configuration set by CNI plugins was not threaded through to the task configuration. This resulted in a regression where a DNS override set by `dockerd` was not respected for `bridge` mode networking. Our existing handling of CNI DNS incorrectly assumed that the DNS field would be empty, when in fact it contains a single empty DNS struct.

Handle this case correctly by checking whether the DNS struct we get back from CNI has any nameservers, and ignore it if it doesn't. Expand test coverage of this case.

Fixes: https://github.com/hashicorp/nomad/issues/20174

---

In addition to the unit tests here, I verified this works with the Docker DNS override. Using the default values for `dockerd` and this patch, I get the following in a Nomad job using `mode="bridge"` networking:

```
$ docker exec -it 8b38 cat /etc/resolv.conf
# Generated by Docker Engine.
# This file can be edited; Docker Engine will not make further changes once it
# has been modified.

nameserver 208.67.222.222
nameserver 208.67.220.220
nameserver 192.168.1.1
search mynetworksettings.com

# Based on host file: '/run/systemd/resolve/resolv.conf' (legacy)
# Overrides: []
```

And after setting an override of `--dns=8.8.8.8` in my `docker.service` unit, I get the following for the same job:

```
$ docker exec -it ec42 cat /etc/resolv.conf
# Generated by Docker Engine.
# This file can be edited; Docker Engine will not make further changes once it
# has been modified.

nameserver 8.8.8.8
search mynetworksettings.com

# Based on host file: '/run/systemd/resolve/resolv.conf' (legacy)
# Overrides: [nameservers]
```

### DIFF
--- a/client/allocrunner/networking_cni.go
+++ b/client/allocrunner/networking_cni.go
@@ -184,12 +184,15 @@ func (c *cniNetworkConfigurator) cniToAllocNet(res *cni.Result) (*structs.AllocN
 
 	}
 
-	// Use the first DNS results.
+	// Use the first DNS results, if non-empty
 	if len(res.DNS) > 0 {
-		netStatus.DNS = &structs.DNSConfig{
-			Servers:  res.DNS[0].Nameservers,
-			Searches: res.DNS[0].Search,
-			Options:  res.DNS[0].Options,
+		cniDNS := res.DNS[0]
+		if len(cniDNS.Nameservers) > 0 {
+			netStatus.DNS = &structs.DNSConfig{
+				Servers:  cniDNS.Nameservers,
+				Searches: cniDNS.Search,
+				Options:  cniDNS.Options,
+			}
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


client/allocrunner/networking_cni.go
```diff
@@ -184,12 +184,15 @@ func (c *cniNetworkConfigurator) cniToAllocNet(res *cni.Result) (*structs.AllocN
 
 	}
 
-	// Use the first DNS results.
+	// Use the first DNS results, if non-empty
```
lgfa29, line 187: Maybe not the best place for this, but it may be useful to have a place listing the multitude of possibilities for DNS config sources 😅

lgfa29 (approved):
Validated that it fixes the problem described in #20174 👍

PR #20218, opened 2024-03-25
[ui] Namespace added to example test in exec window

### LINKED ISSUE #20213: Web UI exec command doesn't consider namespaces
### Nomad version
1.7.5

Debian 12

### Issue
If you navigate, via web UI, to a task in running in a namespace other than `default` and click `Exec`
the proposed command won't work as task's namespace is not considered.
Example:
`nomad alloc exec -i -t -task foobar 11867402 /bin/bash`
instead of
`nomad alloc exec -namespace=dev -i -t -task foobar 11867402 /bin/bash`

Screenshot 2024-03-25 at 12 15 19

philrenaud:
Adds a `-namespace $NAMESPACEID` string to the example we show when you exec in a task in the Nomad web UI (when not in the default namespace)

image

Resolves #20213

### DIFF
--- a/.changelog/20218.txt
+++ b/.changelog/20218.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Show the namespace in the web UI exec command hint
+```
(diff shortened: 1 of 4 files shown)


lgfa29 (approved):
Just missing a changelog entry 😄

PR #20219, opened 2024-03-25
CLI: remove redundant `allocs` profile from `operator debug`

tgross:
The pprof `allocs` profile is identical to the `heap` profile, just with a different default view. Collecting only one of the two is sufficient to view all of `alloc_objects`, `alloc_space`, `inuse_objects`, and `inuse_space`, and collecting only one means that both views will be of the same profile.

Also improve the docstrings on the goroutine profiles explaining what's in each so that it's clear why we might want all of debug=0, debug=1, and debug=2.


jrasell (approved):
LGTM

pkazmierczak (approved):

PR #20221, opened 2024-03-25
exec2: more tweaks to driver harness

shoenig:
Also add an explicit exit code to subproc package for when a child
process is instructed to run an unrunnable command (i.e. cannot be
found or is not executable) - with the 127 return code folks using bash
are familiar with


tgross (approved):
LGTM

PR #20231, opened 2024-03-26
subproc: lazy lookup nomad binary in self call

shoenig:
Avoid having a dependency on the `nomad` binary being present if we are invoking as a test binary and not using this functionality.


tgross (approved):
LGTM!

Issue #20163, opened 2024-03-19
goroutine leak in template runner


tgross:
A Nomad user reported an issue where template runner [`View.poll`](https://github.com/hashicorp/consul-template/blob/v0.37.2/watch/watcher.go#L172) goroutines were being leaked when using templates with many dependencies.

An investigation found that when the runner's `Run` method is in-flight and the runner concurrently receives `Stop`, it's possible for the watcher to stop dependencies via stop and then immediately recreate them. Because the `Stop` method has returned, Nomad's template hook will assume that all goroutines have been stopped, and this causes the `View.poll` goroutines to be leaked. The race condition for this case is very narrow, but is most likely to occur when the user has both a very large number of dependencies, short secret leases, and the template `Wait` configuration is nil.

I've got a PR fixing this in https://github.com/hashicorp/consul-template/pull/1898. Once that's been merged and released, we'll need to bump the dependency in Nomad. Looking to ship this fix in the next patch version of Nomad, as well as backports.

tgross:
https://github.com/hashicorp/consul-template/pull/1898 has been merged and we've sent a build to the folks who reported it so they can confirm it entirely fixes the issue. Once that's been done we'll get that dependency bump shipped here.

PR #20234, opened 2024-03-27
deps: update consul-template to 0.37.4 to fix resource leak

tgross:
A Nomad user reported an issue where template runner `View.poll` goroutines were being leaked when using templates with many dependencies. This resource leak was fixed in consul-template 0.37.4.

Fixes: https://github.com/hashicorp/nomad/issues/20163


shoenig (approved):
LGTM! :spider:

PR #20242, opened 2024-03-27
docs: update Consul compatibility matrix

tgross:
Version of Nomad and Consul that were known not to be compatible are no longer supported in general. Update the compatibility matrix for Consul to match.


david-yu (approved):
Thank you! Looks good and much cleaner now!

PR #20175, opened 2024-03-21
Connect transparent proxy support

tgross:
This is a feature branch PR for adding support for Consul transparent proxy.

Component PRs, all of which have been previously reviewed:
- [x] jobspec https://github.com/hashicorp/nomad/pull/20144
- [x] client-side network configuration (`CNI_ARGS` generation) https://github.com/hashicorp/nomad/pull/20183
- [x] job endpoint hooks for adding constraints on `plugins.cni.version.consul-cni` version https://github.com/hashicorp/nomad/pull/20244
- [x] `consul-cni` updates https://github.com/hashicorp/consul-k8s/pull/3795
- [x] `consul-cni` version string bug fix https://github.com/hashicorp/consul-k8s/pull/3829 (needed for job endpoint hook PR)
- [x] https://github.com/hashicorp/nomad/pull/20233 (needed for job endpoint hook PR)
- [x] reference documentation https://github.com/hashicorp/nomad/pull/20241
- [x] integration documentation https://github.com/hashicorp/nomad/pull/20251 (Tutorials updates under separate repo)
- [x] E2E Connect refresh https://github.com/hashicorp/nomad/pull/20278
- [x] E2E tests

Fixes: https://github.com/hashicorp/nomad/issues/10628

PR #20244, opened 2024-03-27
tproxy: job submission hooks

tgross:
Add a constraint on job submission that requires the `consul-cni` plugin fingerprint whenever transparent proxy is used. Add a validation that the `network.dns` cannot be set when transparent proxy is used, unless the `no_dns` flag is set.

This PR targets the feature branch.
Ref: https://github.com/hashicorp/nomad/pull/20175

### DIFF
--- a/nomad/job_endpoint_hooks.go
+++ b/nomad/job_endpoint_hooks.go
@@ -140,10 +140,7 @@ var (
 	// transparent proxy mode.
 	cniConsulConstraint = &structs.Constraint{
 		LTarget: attrConsulCNI,
-		// TODO(tgross): consul-cni 1.4.2 will have Nomad support and will emit
-		// the correct version string. Once that's shipped update this RTarget
-		// to "1.4.2"
-		RTarget: "1.0.0",
+		RTarget: "1.4.2",
 		Operand: structs.ConstraintSemver,
 	}
 )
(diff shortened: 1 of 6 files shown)


nomad/job_endpoint_hooks.go
```diff
@@ -134,6 +135,17 @@ var (
...
 	}
+
+	// cniConsulConstraint is an implicit constraint added to jobs making use of
+	// transparent proxy mode.
+	cniConsulConstraint = &structs.Constraint{
+		LTarget: attrConsulCNI,
+		// TODO(tgross): consul-cni 1.4.2 will have Nomad support and will emit
+		// the correct version string. Once that's shipped update this RTarget
+		// to "1.4.2"
+		RTarget: "1.0.0",
```
gulducat, line 146: if 1.0.0 here wouldn't actually work, and 1.4.2 equally wouldn't work, why not go ahead and optimistically set it to 1.4.2?

nomad/structs/job.go
```diff
@@ -144,3 +144,21 @@ func (j *Job) RequiredBridgeNetwork() set.Collection[string] {
...
+// RequiredTransparentProxy identifies which task groups, if any, within the job
+// contain Connect blocks using transparent proxy
+func (j *Job) RequiredTransparentProxy() set.Collection[string] {
+	result := set.New[string](len(j.TaskGroups))
+	for _, tg := range j.TaskGroups {
+		for _, service := range tg.Services {
+			if service.Connect != nil {
+				if service.Connect.HasTransparentProxy() {
+					result.Insert(tg.Name)
+					continue
```
gulducat, line 157: is this `continue` future-oriented?  as I read it here, it seems like a no-op, since nothing else follows it.

perhaps a `break` instead, to pop out of the inner `Services` loop, since any single service with tproxy enables it for the whole group/alloc, right?

gulducat (approved):
couple tidbits for ya, lgtm!

nomad/structs/job.go
tgross: Fixed, and added a unit test anyways :grinning:

PR #20241, opened 2024-03-27
tproxy: `transparent_proxy` reference docs

tgross:
This PR is targeting the feature branch, and only covers the jobspec reference docs, not the more involved integration docs that'll be under a separate PR.

Preview links:
* https://nomad-git-f-tproxy-docs-hashicorp.vercel.app/nomad/docs/job-specification/transparent_proxy
* https://nomad-git-f-tproxy-docs-hashicorp.vercel.app/nomad/docs/job-specification/proxy
* https://nomad-git-f-tproxy-docs-hashicorp.vercel.app/nomad/docs/job-specification/upstreams

Ref: https://github.com/hashicorp/nomad/pull/20175

PR #20251, opened 2024-03-28
docs: update service mesh integration docs for transparent proxy

tgross:
Update the service mesh integration docs to explain how Consul needs to be configured for transparent proxy. Update the walkthrough to assume that `transparent_proxy` mode is the best approach, and move the manually-configured `upstreams` to a separate section for users who don't want to use Consul DNS.

Ref: https://github.com/hashicorp/nomad/pull/20175
Ref: https://github.com/hashicorp/nomad/pull/20241

### DIFF
--- a/website/content/docs/integrations/consul/service-mesh.mdx
+++ b/website/content/docs/integrations/consul/service-mesh.mdx
@@ -134,6 +134,25 @@ service_prefix "" { policy = "read" }
 node_prefix    "" { policy = "read" }
 ```
 
+#### Consul DNS For Transparent Proxy
+
+To use Consul's [transparent proxy][] mode with virtual IP addresses, you'll
+need to configure Consul's DNS listener to be exposed to the workload network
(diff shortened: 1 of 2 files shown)


website/content/docs/integrations/consul/service-mesh.mdx
```diff
@@ -213,7 +237,7 @@ job "countdash" {
       driver = "docker"
 
       env {
-        COUNTING_SERVICE_URL = "http://${NOMAD_UPSTREAM_ADDR_count_api}"
+        COUNTING_SERVICE_URL = "http://count-api.virtual.consul:9001"
```
apollo13, line 240: Maybe note that any port can be used here (just guessing)?

tgross: Good catch. I've fixed that and made it clear as to why you don't need it in the text below this.

gulducat (approved):
lgtm!

PR #20259, opened 2024-04-01
docs: recommendation for maximum number of template dependencies

tgross:
We don't have docs on this, just the warning in the CT logs that suggests 128 is an upper limit (per task, not per template): [runner.go#L304](https://github.com/hashicorp/consul-template/blob/ee19d705e7c12ae05bac4f7302b65ede843c7258/manager/runner.go#L304). Given the debugging we did for https://github.com/hashicorp/nomad/pull/20234 https://github.com/hashicorp/nomad/issues/20163 https://github.com/hashicorp/consul-template/pull/1898 it seems like good advice to have.

### DIFF
--- a/website/content/docs/job-specification/template.mdx
+++ b/website/content/docs/job-specification/template.mdx
@@ -332,6 +332,18 @@ task "task" {
 }
 ```
 
+### Dependencies
+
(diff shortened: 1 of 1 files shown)


jrasell (approved):
LGTM!

Issue #20246, opened 2024-03-28
AllocDirStats field missing from api definition of HostStats


tomqwpl:
### Nomad version
This is in git on main as of today.

### Issue
In api/nodes.go there is a definition of HostStats. I believe that this declared what a user of the nomadapi sees as the definition of that struct.
However, in client/hosttast.go there is what I believe is the server side version of this struct. This version has AllocDirStats field in. I can see that AllocDirStats is returned in REST API responses, but the nomadapi version of HostStats doesn't declare that field so it's not available to users of nomadapi.

tgross:
Hi @tomqwpl! Yeah this looks like a bug in `api/nodes.go`. The HTTP server on the client node uses this version: [`client/hoststats/host.go#L25`](https://github.com/hashicorp/nomad/blob/main/client/hoststats/host.go#L25). Most of our structs get defined on the Nomad server under `nomad/structs`, and we have to keep them in sync as part of our development checklist for new fields. But client structs like that one don't have an equivalent checklist, so looks like we missed it.

Easy fix though. We'll get that done.

tgross:
I've got a quick fix up here: https://github.com/hashicorp/nomad/pull/20261

tgross:
Fixed in https://github.com/hashicorp/nomad/pull/20261, which will ship in the next release. Thanks @tomqwpl!

tomqwpl:
@tgross Apologies I didn't get time to create the PR before you got there, since it was a trivial fix I had done locally (albeit I didn't have the test cases)

PR #20261, opened 2024-04-02
api: add missing `AllocDirStats` field in Go API

tgross:
The JSON response for the Read Stats client API includes an `AllocDirStats` field. This field is missing in the `api` package, so consumers of the Go API can't use it to read the values we're getting back from the HTTP server.

Fixes: https://github.com/hashicorp/nomad/issues/20246

### DIFF
--- a/.changelog/20261.txt
+++ b/.changelog/20261.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+api: Fixed a bug where `AllocDirStats` field was missing from Reads Stats client API
+```
--- a/.changelog/20261.txt
+++ b/.changelog/20261.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-api: Fixed a bug where `AllocDirStats` field was missing from Reads Stats client API
+api: Fixed a bug where `AllocDirStats` field was missing from Read Stats client API
 ```
--- a/api/nodes_test.go
+++ b/api/nodes_test.go
@@ -153,6 +153,27 @@ func TestNodes_Info(t *testing.T) {
 	must.GreaterEq(t, 1, len(result.Events))
 }
 
+func TestNode_Stats(t *testing.T) {
+	testutil.Parallel(t)
+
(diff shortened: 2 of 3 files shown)


.changelog/20261.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+api: Fixed a bug where `AllocDirStats` field was missing from Reads Stats client API
```
gulducat, line 2: I suppose "Read" should be singular, and it's capitalized just to match the section heading in [api-docs client page](https://developer.hashicorp.com/nomad/api-docs/client#read-stats)?

```suggestion
api: Fixed a bug where `AllocDirStats` field was missing from Read Stats client API
```

api/nodes_test.go
```diff
@@ -153,6 +153,27 @@ func TestNodes_Info(t *testing.T) {
...
 }
 
+func TestNode_Stats(t *testing.T) {
+	testutil.Parallel(t)
+
+	c, s := makeClient(t, nil, func(c *testutil.TestServerConfig) {
+		c.DevMode = true
+	})
+	defer s.Stop()
+	nodesAPI := c.Nodes()
```
gulducat, line 163: ah this is one of those little "node" vs "client" battlegrounds, eh?

gulducat (approved):
LGTM!

Juanadelacuesta (approved):

api/nodes_test.go
gulducat: I was more pointing out that it hits the `/client` API, not `/node`!  looks like there's no `client.go` to have something like `c.Clients()` presumably to avoid the name collision of "Nomad client" and "Nomad API client" ... so the Nodes{} (go) API on the (go) API Client{} is used to hit both the /nodes (http) API and the /client (http) API :sweat_smile:

tgross: Ah! Yeah, totally.

PR #20279, opened 2024-04-03
[ui] alert a user on the jobs index page when a deploying job requires manual promotion

### LINKED ISSUE #20240: [ui] Alert a user on the jobs index page when a deploying job requires manual promotion
(when it has a non-zero number for `canary` in the `update` block of its jobspec and those canaries are running/healthy)

philrenaud:
image


ui/app/templates/components/job-row.hbs
```diff
@@ -3,71 +3,111 @@
...
-        @isNarrow={{true}}
-      />
-    {{/if}}
-  </div>
-</td>
\ No newline at end of file
+  <@tableBody.Td data-test-job-status>
+    {{#unless @job.childStatuses}}
+      <div class="status-cell">
+        <Hds::Badge @text="{{capitalize @job.aggregateAllocStatus.label}}" @color={{@job.aggregateAllocStatus.state}} @size="large" />
```
philrenaud, line 46: TODO: @size="medium" here

philrenaud:
(this is still a good idea, but this branch is so out-of-date that it's worth a re-do from scratch)

PR #20299, opened 2024-04-04
Enable numeric pagination tokens

gulducat:
For #19339 we will be creating a state index that sorts jobs by `ModifyIndex`, in order to show most-recently-changed jobs at the top of the Nomad web UI, while continuing to sort appropriately with pagination.

The current pagination logic assumes all `NextToken`s are `string`s, which results in odd paging behavior (including infinite loops!) when the token is actually a number, due to lexicographical ordering, e.g. `"10" < "2"`

This doesn't go as far as fully generalizing the `Paginator`/`Tokenizer` relationship (as might be done with `comparable` or `cmp.Ordered` generics), but it adds support for `uint64` tokens for my immediate use case, and opens a door for further future flexibility.

### DIFF
--- a/nomad/state/paginator/paginator.go
+++ b/nomad/state/paginator/paginator.go
@@ -4,7 +4,9 @@
 package paginator
 
 import (
+	"cmp"
 	"fmt"
+	"strconv"
 
 	"github.com/hashicorp/go-bexpr"
 	"github.com/hashicorp/nomad/nomad/structs"
@@ -26,6 +28,7 @@ type Paginator struct {
 	perPage        int32
 	itemCount      int32
 	seekingToken   string
+	seekingUint    uint64
 	nextToken      string
 	reverse        bool
 	nextTokenFound bool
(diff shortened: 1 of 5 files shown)


nomad/state/paginator/paginator.go
```diff
@@ -54,12 +57,16 @@ func NewPaginator(iter Iterator, tokenizer Tokenizer, filters []Filter,
 		filters = append(filters, evaluator)
 	}
 
+	// attempt to convert token to uint for iterators ordered numerically
+	seekingUint, _ := strconv.ParseUint(opts.NextToken, 10, 64)
```
tgross, line 61: Dropping errors is "weird" enough we should report why. Something like:

```suggestion
	// attempt to convert token to uint for iterators ordered numerically.
	// it's safe to ignore the error here because the `next` method ignores 
	// this field for string tokens and 0 is valid for an unset numeric NextToken
	seekingUint, _ := strconv.ParseUint(opts.NextToken, 10, 64)
```

tgross (approved):
LGTM! One minor comment about the comment and we're good-to-go.

shoenig (approved):

PR #20300, opened 2024-04-04
E2E: ensure periodic test can't fail due to cron conflicts

tgross:
The E2E test for periodic dispatch jobs has a `cron` trigger for once a minute. If the test happens to run at the top of the minute, it's possible for the forced dispatch to run from the test code, then the periodic timer triggers and leaves a running child job. This fails the test because it expects only a single job in the "dead" state.

Make it so that the `cron` expression is implausible to run during our test window, and migrate the test off the old framework while we're at it.

---

```
$ go test -v -count=1 ./e2e/periodic
=== RUN   TestPeriodicDispatch_Basic
--- PASS: TestPeriodicDispatch_Basic (6.27s)
PASS
ok      github.com/hashicorp/nomad/e2e/periodic 6.278s
```


jrasell (approved):
LGTM, thanks!

Issue #20298, opened 2024-04-04
verifying JWKS endpoint from Vault fails (Vault 1.16.1)


tgross:
Our compatibility test suite detected a regression in Vault's auth config setup that will impact Nomad users. The error we get from the Vault logs is as follows:

> 2024-04-04T13:55:35.235-0400 [ERROR] auth.jwt.auth_jwt_38ca2b23: error checking jwks URL: url=http://127.0.0.1:39647/.well-known/jwks.json error="oidc: malformed jwt: go-jose/go-jose: unexpected signature algorithm \"HS256\"; expected [\"RS256\" \"RS384\" \"RS512\" \"ES256\" \"ES384\" \"ES512\" \"PS256\" \"PS384\" \"PS512\" \"EdDSA\"]"

The bug is because an upstream dependency `go-jose` added an extra reasonableness check in https://github.com/go-jose/go-jose/pull/74, which got pulled in via our [cap](https://github.com/hashicorp/cap) library into the Vault JWT auth plugin. This bug has been fixed in https://github.com/hashicorp/vault-plugin-auth-jwt/pull/290

This will prevent the JWT auth config method from being updated with new keys if you rotate Nomad's keyring.

Nomad users should not upgrade to Vault 1.16.1 if they are using [Workload Identities with Vault](https://developer.hashicorp.com/nomad/docs/integrations/vault/acl#nomad-workload-identities). If you have already upgraded or need to upgrade before Vault 1.16.2 is released, you can download and run the current version of the JWT auth plugin separately.

MatthewJohn:
For what it's worth, (as per comment in https://github.com/hashicorp/learn-nomad-cluster-setup/pull/30), Vault 1.16.2 has now been released.

From my testing, it appears to fix this issue (at least with the setup of the auth backend - yet to test to see if it _actually_ works) :)

tgross:
Thanks @MatthewJohn! I've opened https://github.com/hashicorp/nomad/pull/20484 reverting our test case that verifies everything end-to-end and that appears to be passing too.

PR #20301, opened 2024-04-04
E2E: skip Vault 1.16.1 for JWT compatibility test

tgross:
Vault 1.16.1 has a known issue around the JWT auth configuration that will prevent this test from ever passing. Skip testing the JWT code path on 1.16.1. Once 1.16.2 ships it will no longer get skipped.

Ref: https://github.com/hashicorp/nomad/issues/20298


shoenig (approved):

lgfa29 (approved):
LGTM. 

But a thought I had was if it would make sense to instead skip the version in `scanVaultVersions()` so that we still run the test in `1.16.0`?

tgross:
> But a thought I had was if it would make sense to instead skip the version in scanVaultVersions() so that we still run the test in 1.16.0?

Well I still wanted to run the legacy test with the current version if we could. I'm going to merge this for now so that folks' unrelated work has green tests, and I'll circle back to seeing how we could improve upon it tomorrow.

PR #20304, opened 2024-04-05
docs: add Digital Ocean Spaces artifact jobspec example.

### LINKED ISSUE #20088: [docs] s3-compatible artifact URL examples are missleading
# Overview

The examples used in the [Download from an S3-compatible Bucket](https://developer.hashicorp.com/nomad/docs/job-specification/artifact#download-from-an-s3-compatible-bucket) section use actual AWS s3 URLs which, if you look at how go-getter [parses](https://github.com/hashicorp/go-getter/blob/975961f5f06346ccc282cd0d9aa16e160d26f9e3/get_s3.go#L252) them, you'll see that it'll will treat it as the actual AWS s3 URL which is probably why the examples worked when (hopefully) tested before including them in documentation.

For [other domains](https://github.com/hashicorp/go-getter/blob/975961f5f06346ccc282cd0d9aa16e160d26f9e3/get_s3.go#L298), using the `s3::https://(bucket).(host)/(key)` format is **not** supported and will yield a `URL is not a valid S3 compliant URL` error.

In such cases the URL needs to be modified so that the bucket is a path prefix, not subdomain like so: `s3::https://(host)/(bucket)/(key)`

# DigitalOcean Spaces Example

I discovered this while attempting to make artifacts work with a private DigitalOcean spaces bucket, but from what I've seen while troubleshooting this, the same might apply to Minio (which is explicitly mentioned in artifact docs):

For a bucket named `artifact-test` located in the `sfo3` region, the (Origin) Endpoint you are given is https://artifact-test.sfo3.digitaloceanspaces.com

In order to download an `example.txt` file from this bucket via nomad job artifacts the URL given in Spaces UI will work if the bucket is **public**. This is because then it's just an HTTP request and not an s3 API call so go-getter's s3 handling code is not used at all.

If the bucket, or key/file is **private**, then the URL needs to be changed from:

```
https://artifact-test.sfo3.digitaloceanspaces.com/example.txt
```

to:

```
s3::https://sfo3.digitaloceanspaces.com/artifact-test/example.txt
```

Note that in addition to forcing s3 handling via the `s3::` prefix as currently stated in the docs, the bucket portion of the URL needs to be moved as well. This extra step is not documented in either nomad artifact, nor go-getter s3 docs and the resulting error doesn't help at all

jrasell:
closes #20088

### DIFF
--- a/website/content/docs/job-specification/artifact.mdx
+++ b/website/content/docs/job-specification/artifact.mdx
@@ -244,6 +244,28 @@ artifact {
 }
 ```
 
+[DigitalOcean Spaces][do_spaces] provide S3-compatible object storage and can be
+used in conjunction with the `artifact` block. Given a bucket named
+`my-bucket-example` located in the `sfo3` region and a file named `my_app.tar.gz`,
+the following artifact block will work, provided the bucket and file are public.
+This uses the origin endpoint as detailed directly from Digital Ocean.
+
+```hcl
+artifact {
+  source = "https://my-bucket-example.sfo3.digitaloceanspaces.com/my_app.tar.gz"
+}
+```
+
+If the bucket or file are private, the artifact source string needs to be
+modified to add the `s3::` prefix. The bucket name is also moved to form part
+of the URL path component.
+
+```hcl
+artifact {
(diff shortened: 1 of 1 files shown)


pkazmierczak (approved):
LGTM!

Issue #20307, opened 2024-04-05
fmt adds a newline on each invocation using stdin/stdout


rliebz:
### Nomad version
Output from `nomad version`
```
Nomad v1.7.6+ent
BuildDate 2024-03-12T07:55:51Z
Revision a71f7e82ef60200f273a92601b1b366429b0bd5f
```

### Operating system and Environment details
macOS 14.4.1 

### Issue

Calls to `nomad fmt -` append an extra newline on each successive call

### Reproduction steps

1. Take any nomad file
2. Run it through multiple calls to `nomad fmt -`

```bash
cat example.nomad.hcl | nomad fmt - | nomad fmt - | nomad fmt -
```

#### Expected Result

Either of two behaviors:
1. The output should have exactly one trailing newline (probably ideal)
1. The output should have the original number of trailing newlines (this is what `terraform fmt -` does)

#### Actual Result

The output has one additional trailing newline for each invocation.

This can cause problems when trying to integrate `nomad fmt` into an editor, because each subsequent attempt to format will modify the file to contain an additional newline.

### Job file (if appropriate)
Doesn't matter, but for demonstration:
```nomad
job "example" {
  datacenters = ["dc1"]

  group "group1" {
    task "task1" {
      driver = "exec"
      config {
        command = "echo Hello, world!"
      }
    }
  }
}
```

### Nomad Server logs (if appropriate)
N/A

### Nomad Client logs (if appropriate)
N/A

tgross:
Hi @rliebz! I did a little testing and it looks like the bug is on the stdout end of things. I've got https://github.com/hashicorp/nomad/pull/20318 up to fix that and that'll ship in the next version of Nomad. Thanks!

PR #20318, opened 2024-04-08
cli: remove extraneous trailing newline from `nomad fmt`

tgross:
When `nomad fmt` writes to stdout instead of overwriting a file, the command was using the `UI` output, which appends an extra newline. This results in extra trailing newlines when using `nomad fmt` as part of a pipeline or editor plugin.

Update the command to write directly to stdout when in the stdout mode.

Fixes: https://github.com/hashicorp/nomad/issues/20307


gulducat (approved):
love a little win

Issue #20310, opened 2024-04-07
client: parsing of server IPv6 addresses is inconsistent with docs


jodersky:
### Nomad version

v1.7.6

### Operating system and Environment details

Debian 12

### Issue

When specifying an IPv6 address to connect to a Nomad server, it must be enclosed in brackets *and* include the port, or else the address parsing will not succeed, and the client will remain disconnected.

The [documentation](https://developer.hashicorp.com/nomad/docs/configuration/client#servers) specifically mentions that the port may be omitted, and also does not mention that the address must be given in URL form (i.e. wrapped in brackets for IPv6). This, further with by-default hidden log messages, makes this error quite hard to diagnose.

#### Using raw IPv6 address

```
client {
  enabled = true
  servers = ["2001:db8::1"]
}
log_level = "DEBUG"
```

Log:

>  [DEBUG] client: ignoring server due to resolution error: error="address 2001:db8::1: too many colons in address" server=2001:db8::1

#### Using IPv6 address as in a URL

```
client {
  enabled = true
  servers = ["[2001:db8::1]"]
}
log_level = "DEBUG"
```

Log:
>[DEBUG] client: ignoring server due to resolution error: error="address [[2001:db8::1]]:4647: missing port in address" server=[2001:db8::1]

### Workaround

Specify address as a URL and include port

```
client {
  enabled = true
  servers = ["[2001:db8::1]:4647"]
}
```

tgross:
Hi @jodersky! This is both a bug and a docs issue. I've fixed both in https://github.com/hashicorp/nomad/pull/20324. Thanks!

PR #20324, opened 2024-04-08
client: fix IPv6 parsing for `client.servers` block

tgross:
When the `client.servers` block is parsed, we split the port from the address. This does not correctly handle IPv6 addresses when they are in URL format (wrapped in brackets), which we require to disambiguate the port and address.

Fix the parser to correctly split out the port and handle a missing port value for IPv6. Update the documentation to make the URL format requirement clear.

Fixes: https://github.com/hashicorp/nomad/issues/20310


gulducat (approved):
LGTM!

the OP also pointed out that [the log](https://github.com/hashicorp/nomad/blob/v1.7.6/client/client.go#L1165) is somewhat hidden being log_level DEBUG -- perhaps we could also change that to `Info` or even `Warn`?

tgross:
> the OP also pointed out that the log is somewhat hidden being log_level DEBUG -- perhaps we could also change that to Info or even Warn?

I'm pretty sure the reason we don't debug at Info or Warn is that if it's a DNS address and you're having DNS resolution issues, it's going to create a lot of spam in the logs that's better handled in the caller (which gets the errors).

PR #20329, opened 2024-04-09
eval_broker: track enqueue and dequeue times

pkazmierczak:
Adds new metrics to the eval broker that track times of evaluations enqueueing and dequeueing.


nomad/eval_broker.go
```diff
@@ -158,6 +163,8 @@ func NewEvalBroker(ctx context.Context, timeout, initialNackDelay, subsequentNac
 		timeWait:             make(map[string]*time.Timer),
 		initialNackDelay:     initialNackDelay,
 		subsequentNackDelay:  subsequentNackDelay,
+		enqueuedTime:         make(map[string]time.Time),
+		dequeuedTime:         make(map[string]time.Time),
```
tgross, line 167: We know in advance that these maps are going to get populated and that will cause memory allocations in the hot path of the eval broker. Maybe we should pre-allocate some nice round number like 256.

Also, we're creating these empty maps in `NewEvalBroker` which gets called on every server, including followers. But when there's a leader election the new leader goes through the state store and re-enqueues all the evaluations into its own eval broker. We should make sure we're clearing these maps when the follower steps down. Otherwise if the leaders change and then change back, we'll end up with very weird metrics.

website/content/docs/operations/metrics-reference.mdx
```diff
@@ -93,13 +93,16 @@ be assumed to be 10 seconds when retrieving metrics using the above described
...
+| `nomad.nomad.broker.wait_time`               | Time elapsed while the evaluation was ready to be processed and waiting to be dequeued                                                                                                                            | ns / Evaluation Wait           | Timer   |
+| `nomad.nomad.broker.process_time`            | Time elapsed while the evaluation was dequeued and finished processing                                                                                                                                            | ns / Evaluation Process        | Timer   |
+| `nomad.nomad.broker.response_time`           | Time elapsed from when the evaluation was last enqueued and finished processing                                                                                                                                   | ns / Evaluation Response       | Timer   |
```
tgross, line 105: We're likely to get questions about this data when the numbers look weird after a leader election. I noted above that we should clear the maps, but we should note here that the metrics are only valid within a single term as well.

nomad/eval_broker.go
```diff
@@ -369,6 +385,17 @@ SCAN:
 		if timeoutTimer != nil {
 			timeoutTimer.Stop()
 		}
+		b.l.Lock()
+		if t, ok := b.enqueuedTime[eval.ID]; ok {
+			b.dequeuedTime[eval.ID] = time.Now()
```
lgfa29, line 390: We should limit the map size here as well to make sure the map doesn't grow too large.

nomad/eval_broker.go
```diff
@@ -724,6 +754,38 @@ func (b *EvalBroker) ResumeNackTimeout(evalID, token string) error {
...
+	tEnq, ok := b.enqueuedTime[eval.ID]
+	if !ok {
+		delete(b.dequeuedTime, eval.ID)
+		return
+	}
+
+	tDeq, ok := b.dequeuedTime[eval.ID]
+	if !ok {
+		return
+	}
```
lgfa29, line 771:
```suggestion
	tDeq, ok := b.dequeuedTime[eval.ID]
	if !ok {
	    delete(b.enqueuedTime, eval.ID)
		return
	}
```

Cleaning up in case there's an enqueue time but no dequeue.

nomad/eval_broker.go
```diff
@@ -724,6 +754,38 @@ func (b *EvalBroker) ResumeNackTimeout(evalID, token string) error {
 	return nil
 }
 
+func (b *EvalBroker) handleAckNackLocked(eval *structs.Evaluation) {
```
lgfa29, line 757: There's some potential weirdness here to consider. `Ack` and `Nack` are called with a string ID, but the eval may not exists in the `unack` queue, so I wonder if there's a possibility where `enqueueTime` and `dequeueTime` have an eval ID that is not in `unack`, causing them to never get cleaned up.

At some point I thought of passing `evalID` and `eval` here to handle the clean up but I was being lazy and dropped it 😅

nomad/eval_broker.go
lgfa29: Yeah, I think we need to make a more conscious decision on how to handle leadership transition. 

Pulling my comment from the internal RFC discussion:

> But looking at the the leadership transition code, when we restore evals we just call the same Enqueue() method, so the wait times are likely to go down and service time metrics dropped because we don't have the dequeue time anymore.
> 
> Ignoring the restored evals (meaning, don't emit these new metrics for them) can cause a gap until evals created after the leadership transition start to flow in.
> 
> Between the two options, I think a gap may be better than incorrect values?

Maybe a good way to handle would be:

1. Clean new timing maps on leadership step down.
2. Don't set enqueue time for restored evals on new leader. We may need to create a separate method for eval restore, like this:
    ```go
	func (b *EvalBroker) Restore(eval *structs.Evaluation) {
		b.l.Lock()
		defer b.l.Unlock()
		b.processEnqueue(eval, "", false)
	}
	
	func (b *EvalBroker) Enqueue(eval *structs.Evaluation) {
		b.l.Lock()
		defer b.l.Unlock()
		b.processEnqueue(eval, "", true)
	}
	
	func (b *EvalBroker) processEnqueue(eval *structs.Evaluation, token string, trackTime bool) {
		// ...
	}
    ```

I'm still not 100% sure with the right approach here, so it would be good to test a few options and document the intention and decision (in code, documentation, and in the RFC).

website/content/docs/operations/metrics-reference.mdx
lgfa29, line 1: We also need to document the metric details in https://developer.hashicorp.com/nomad/docs/operations/metrics-reference#server-metrics

website/content/docs/operations/metrics-reference.mdx
lgfa29, line 105: It may be worth expanding https://developer.hashicorp.com/nomad/docs/operations/monitoring-nomad#scheduling to mention how to use these metrics as well.

I think tracking `response_time` is going to be something important to track, and then `wait_time` and `process_time` can help nail down the problem. An increase in `wait_time` probably means too much load. An increase in `process_time` points more towards a server performance issue (CPU, disk, network etc.)

We likely won't know exactly what to write until we advance in our explorations, but good to keep in mind.

lgfa29 (commented):
The leadership transition is probably the key point missing. Not sure if there's a "right" answer, so feel free to explore different approaches and trade-offs.

website/content/docs/operations/metrics-reference.mdx
pkazmierczak: corrected in [7e37065](https://github.com/hashicorp/nomad/pull/20329/commits/7e37065b3271b139179bec1ba01026f5e264617f)

nomad/eval_broker.go
pkazmierczak: corrected in [268db7f](https://github.com/hashicorp/nomad/pull/20329/commits/268db7fde683a9ab3247c618855b85ae353aa365)

nomad/eval_broker.go
pkazmierczak: > We know in advance that these maps are going to get populated and that will cause memory allocations in the hot path of the eval broker. Maybe we should pre-allocate some nice round number like 256.

good thinking. Added in [383b9ce](https://github.com/hashicorp/nomad/pull/20329/commits/383b9ce2e776199fd3d62db4998ae9c1ca1bb17e)

pkazmierczak:
@lgfa29 @tgross I have a suggestion on how to handle leadership changes in a48e127. I implemented something Luiz suggested in one of his comments, basically the leader calls an `evalBroker.Restore` method now in `restoreEvals()` instead of the usual `evalBroker.Enqueue`. This isn't very elegant, because we have to pass around the information about whether to track {en,de}queue times inside eval broker methods, but then again public methods remain the same so the interface doesn't change. 

Let me know what you think about this.

tgross:
> I have a suggestion on how to handle leadership changes in https://github.com/hashicorp/nomad/commit/a48e1275aaca048779d0d3dc12508006c06b0dc6. I implemented something Luiz suggested in one of his comments, basically the leader calls an `evalBroker.Restore` method now in `restoreEvals()` instead of the usual `evalBroker.Enqueue`. This isn't very elegant, because we have to pass around the information about whether to track {en,de}queue times inside eval broker methods, but then again public methods remain the same so the interface doesn't change.

This will resolve the issue of misleading metrics by introducing a gap across leader elections. That seems fine so long as we document the expected behavior in the metrics reference.

This doesn't yet resolve the issue of clearing the maps of timings across leader elections. We should clear the maps between terms (either on step-up or step-down, doesn't really matter so whatever is convenient for the broker API).

pkazmierczak:
> This doesn't yet resolve the issue of clearing the maps of timings across leader elections. We should clear the maps between terms (either on step-up or step-down, doesn't really matter so whatever is convenient for the broker API).

Hmm, I thought clearing the map in the `flush` method as we do here https://github.com/hashicorp/nomad/pull/20329/commits/a48e1275aaca048779d0d3dc12508006c06b0dc6#diff-727104c8049d79aa6b8bcd481b366a231033732bd9ad314e408933d6ac25f891R848-R849 would solve this problem, because the `revokeLeadership` method on leader https://github.com/hashicorp/nomad/blob/main/nomad/leader.go#L1488 calls it with enabled set to `false`? I'll keep looking, perhaps I misunderstand how revoking leadership works.

tgross (approved):

Issue #20025, opened 2024-02-22
Variable Interpolation error with Consul Workload Identity


MorphBonehunter:
### Nomad version
Output from `nomad version`

```
Nomad v1.7.5
BuildDate 2024-02-13T15:10:13Z
Revision 5f5d4646198d09b8f4f6cb90fb5d50b53fa328b8
```

### Operating system and Environment details
Arch Linux

### Issue

Today i try to migrate to the workload identity mechanism.
While the vault migration went well, the consul part fails on the variable interpolation stuff.

For easiness and simple change possibility i write my jobfiles with extensive use of `NOMAD_TASK_NAME` variable,
this doesn't work anymore after migration if you use this in the consul service definition part:

```
    task "ara" {
      template {
      ...
      }
      driver = "docker"
      config {
        image = ...
        hostname = "${NOMAD_TASK_NAME}"
        command = "/opt/ara/ara-init.sh"
        mount {
          type = "volume"
          source = "${NOMAD_TASK_NAME}"
          target = "/opt/ara/data"
        }
        ports = [
          "http"
        ]
        labels {
          cpu_shares = "${NOMAD_CPU_LIMIT}"
          container_hostname = "${NOMAD_TASK_NAME}"
        }
      }
      service {
        name = "${NOMAD_TASK_NAME}"
        tags = [
          "${node.class}",
          "traefik.enable=true",
          "traefik.http.routers.${NOMAD_TASK_NAME}.entryPoints=https",
          "traefik.http.routers.${NOMAD_TASK_NAME}.tls=true",
          "traefik.http.routers.${NOMAD_TASK_NAME}.rule=Host(`${NOMAD_TASK_NAME}....`)",
          "traefik.http.routers.${NOMAD_TASK_NAME}.middlewares=compression@file,secureheaders@file,local-access-only@file",
          "traefik.tags=host_${attr.unique.hostname}"
        ]
        port = "http"
        check {
          name = "${NOMAD_TASK_NAME}"
          type = "http"
          method = "HEAD"
          path = "/"
          interval = "20s"
          timeout = "5s"
          check_restart {
            limit = 3
            grace = "10s"
          }
        }
      }
    }
```

After that the following error occurs:

```
failed to setup alloc: pre-run hook "consul" failed: 1 error occurred: * failed to derive Consul token for service ${NOMAD_TASK_NAME}: Unexpected response code: 500 (rpc error making call: invalid bind name: "${nomad_task_name}")
```

### Reproduction steps
[...]

PR #20344, opened 2024-04-09
WI: ensure Consul hook and WID manager interpolate services

tgross:
Services can have some of their string fields interpolated. The new Workload Identity flow doesn't interpolate the services before requesting signed identities or using those identities to get Consul tokens.

Add support for interpolation to the WID manager and the Consul tokens hook by providing both with a taskenv builder. Add an "interpolate workload" field to the WI handle to allow passing the original workload name to the server so the server can find the correct service to sign.

This changeset also makes two related test improvements:
* Remove the mock WID manager, which was only used in the Consul hook tests and isn't necessary so long as we provide the real WID manager with the mock signer and never call `Run` on it. It wasn't feasible to exercise the correct behavior without this refactor, as the mocks were bypassing the new code.
* Fixed swapped expect-vs-actual assertions on the `consul_hook` tests.

Fixes: https://github.com/hashicorp/nomad/issues/20025

---

Smoke tested with the following jobspec:

```hcl
job "example" {

  group "web" {

    network {
      mode = "bridge"
      port "metrics" {
        to = 9000
      }
      port "www" {
        to = 8001
      }
    }

    service {
      name = "${NOMAD_GROUP_NAME}"
      port = "metrics"
    }

    task "http" {

      driver = "docker"

      service {
        name = "${NOMAD_TASK_NAME}"
        port = "www"
      }

      config {
        image   = "busybox:1"
        command = "httpd"
        args    = ["-vv", "-f", "-p", "8001", "-h", "/local"]
        ports   = ["www"]
      }

      resources {
        cpu    = 100
        memory = 100
      }

    }
  }
}
```

</details>

success

### DIFF
--- a/client/allocrunner/alloc_runner.go
+++ b/client/allocrunner/alloc_runner.go
@@ -31,6 +31,7 @@ import (
 	"github.com/hashicorp/nomad/client/serviceregistration/wrapper"
 	cstate "github.com/hashicorp/nomad/client/state"
 	cstructs "github.com/hashicorp/nomad/client/structs"
+	"github.com/hashicorp/nomad/client/taskenv"
 	"github.com/hashicorp/nomad/client/vaultclient"
 	"github.com/hashicorp/nomad/client/widmgr"
 	"github.com/hashicorp/nomad/helper/pointer"
@@ -285,8 +286,17 @@ func NewAllocRunner(config *config.AllocRunnerConfig) (interfaces.AllocRunner, e
 	ar.shutdownDelayCtx = shutdownDelayCtx
 	ar.shutdownDelayCancelFn = shutdownDelayCancel
 
+	// Create a *taskenv.Builder for the allocation so the WID manager can
+	// interpolate services with the allocation and tasks as needed
+	envBuilder := taskenv.NewBuilder(
+		config.ClientConfig.Node,
+		ar.Alloc(),
+		nil,
+		config.ClientConfig.Region,
+	).SetAllocDir(ar.allocDir.AllocDirPath())
+
 	// initialize the workload identity manager
-	widmgr := widmgr.NewWIDMgr(ar.widsigner, alloc, ar.stateDB, ar.logger)
+	widmgr := widmgr.NewWIDMgr(ar.widsigner, alloc, ar.stateDB, ar.logger, envBuilder)
 	ar.widmgr = widmgr
 
 	// Initialize the runners hooks.
--- a/client/taskenv/services.go
+++ b/client/taskenv/services.go
@@ -4,7 +4,6 @@
 package taskenv
 
 import (
-	"github.com/hashicorp/nomad/helper"
 	"github.com/hashicorp/nomad/nomad/structs"
 )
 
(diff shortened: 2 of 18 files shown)


client/allocrunner/alloc_runner.go
```diff
@@ -285,8 +286,17 @@ func NewAllocRunner(config *config.AllocRunnerConfig) (interfaces.AllocRunner, e
 	ar.shutdownDelayCtx = shutdownDelayCtx
 	ar.shutdownDelayCancelFn = shutdownDelayCancel
 
+	// Create a *taskenv.Builder for the allocation so the WID manager can
+	// interpolate services with the allocation and tasks as needed
+	envBuilder := taskenv.NewBuilder(
```
schmichael, line 291: This is the second time allocrunner uses taskenv.Builder in a unique way compared to taskrunner (nil Task, in the other place it's used in a closure). Might be nice to reevaluate the old Builder design to see if we can split up the alloc and task concerns and make it less dangerous to hold.

client/widmgr/mock.go
```diff
@@ -113,35 +113,3 @@ func (m *MockWIDSigner) SignIdentities(minIndex uint64, req []*structs.WorkloadI
 	}
 	return swids, nil
 }
-
-// MockWIDMgr mocks IdentityManager interface allowing to only get identities
```
schmichael, line 117: I love to see a mock removed!

schmichael (approved):
The security implications of this do make me nervous. As you and @gulducat discussed offline there doesn't appear to be any immediate concern, but security issues often arise from unintended side effects down the road.

I wonder if we should deprecate this form of interpolation in this place in favor of HCL variables? I think an HCL variable with a default value could specify both the group and service names in a way that would make them static by the time they're submitted to Nomad. It would be nice to have a path away from this extremely buggy phased interpolation approach.

client/taskenv/services.go
```diff
@@ -221,3 +221,16 @@ func interpolateTaskResources(taskEnv *TaskEnv, resources *structs.Resources) {
 		}
 	}
 }
+
+// InterpolateWIHandle returns a copy of the WIHandle with only the
+// InterpolatedWorkloadIdentifier field interpolated. The original
+// WorkloadIdentifier should never be altered so the server can find
+// uninterpolated services associated with the handle.
+func InterpolateWIHandle(taskEnv *TaskEnv, orig structs.WIHandle) structs.WIHandle {
```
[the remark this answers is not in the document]
tgross: Yeah, we can do that and it just will have to handle the nil case in the alloc endpoint.

tgross:
> I wonder if we should deprecate this form of interpolation in this place in favor of HCL variables? I think an HCL variable with a default value could specify both the group and service names in a way that would make them static by the time they're submitted to Nomad. It would be nice to have a path away from this extremely buggy phased interpolation approach.

I love this idea! Right now interpolation can include data that really is allocation-specific or node-specific, like node metadata or the IP address. But there's probably some subsets of fields that we could say "you're allowed to interpolate _these_ fields with node metadata, and _these_ fields only with HCL2". Being able to document which those are would be nice for users and for us developers. I'll open a new issue for that.

tgross:
@schmichael I've opened https://github.com/hashicorp/nomad/issues/20362 to follow up on the design of interpolation more generally.

PR #20346, opened 2024-04-10
fix exhausted node metrics reporting in preemption

gabivlj:
If preemption is enabled, Nomad doesn't report back the `devices: no devices match request` when it does not find an allocation to evict. 

This allows admins to check why allocations are not being placed when preemption is configured in Nomad.


gabivlj:
Hi @tgross I think this is real useful for us, do these changes look good to you?

Thank you!

tgross (approved):
LGTM!

I've added a changelog entry and once CI is done again, I'll get this merged. Thanks @gabivlj!

Issue #19988, opened 2024-02-14
use of -purge leads to leak of goroutines 


shoenig:
When stopping a job with the `-purge` flag, a Nomad client may leak goroutines. This becomes readily apparent when rapidly starting and stopping a job as described below. There are two goroutines that show up in large quantity in a dump after applying the reproduction steps, both associated with deployment watcher. If the `-purge` flag is not set, no leak is observed.

```
goroutine 1246 [select]:
runtime.gopark(0xc0016b9f78?, 0x4?, 0x90?, 0x5e?, 0xc0016b9eb0?)
        runtime/proc.go:398 +0xce fp=0xc0016b9ce8 sp=0xc0016b9cc8 pc=0x4426ce
runtime.selectgo(0xc0016b9f78, 0xc0016b9ea8, 0x0?, 0x0, 0xedd5f13c8?, 0x1)
        runtime/select.go:327 +0x725 fp=0xc0016b9e08 sp=0xc0016b9ce8 pc=0x452e45
github.com/hashicorp/nomad/nomad/deploymentwatcher.(*deploymentWatcher).watch(0xc001826840)
        github.com/hashicorp/nomad/nomad/deploymentwatcher/deployment_watcher.go:440 +0x226 fp=0xc0016b9fc8 sp=0xc0016b9e08 pc=0x196afe6
github.com/hashicorp/nomad/nomad/deploymentwatcher.newDeploymentWatcher.func1()
        github.com/hashicorp/nomad/nomad/deploymentwatcher/deployment_watcher.go:134 +0x25 fp=0xc0016b9fe0 sp=0xc0016b9fc8 pc=0x19697c5
runtime.goexit()
        runtime/asm_amd64.s:1650 +0x1 fp=0xc0016b9fe8 sp=0xc0016b9fe0 pc=0x4764a1
created by github.com/hashicorp/nomad/nomad/deploymentwatcher.newDeploymentWatcher in goroutine 327
        github.com/hashicorp/nomad/nomad/deploymentwatcher/deployment_watcher.go:134 +0x3aa
```

```
goroutine 20345 [select]:
runtime.gopark(0xc00121fac8?, 0x21?, 0x20?, 0x0?, 0xc00121f986?)
        runtime/proc.go:398 +0xce fp=0xc00121f7e8 sp=0xc00121f7c8 pc=0x4426ce
[...]

PR #20348, opened 2024-04-10
deployment watcher: fix goroutine leak when job is purged

tgross:
The deployment watcher on the leader makes blocking queries to detect when the set of active deployments changes. It takes the resulting list of deployments and adds or removes watchers based on whether the deployment is active. But when a job is purged, the deployment will be deleted. This unblocks the query but the query result only shows the remaining deployments.

When the query unblocks, ensure that all active watchers have a corresponding deployment in state. If not, remove the watcher so that the goroutine stops.

Fixes: https://github.com/hashicorp/nomad/issues/19988

### DIFF
--- a/nomad/deploymentwatcher/deployments_watcher.go
+++ b/nomad/deploymentwatcher/deployments_watcher.go
@@ -191,20 +191,8 @@ func (w *Watcher) watchDeployments(ctx context.Context) {
 			}
 		}
 
-		// Ensure we're not tracking deployments that have been deleted because
-		// the job was purged
-		for _, watcher := range w.watchers {
-			var found bool
-			for _, d := range deployments {
-				if watcher.deploymentID == d.ID {
-					found = true
-					break
-				}
-			}
-			if !found {
-				w.removeByID(watcher.deploymentID)
-			}
-		}
+		// Ensure we've removed deployments for purged jobs
+		w.removeDeletedDeployments(deployments)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


nomad/deploymentwatcher/deployments_watcher.go
```diff
@@ -191,6 +190,21 @@ func (w *Watcher) watchDeployments(ctx context.Context) {
 				w.remove(d)
 			}
 		}
+
+		// Ensure we're not tracking deployments that have been deleted because
```
Juanadelacuesta, line 194: If you check for the number of watchers and number of deployments, you could avoid always checking even when there are no changes

tgross: Good catch. That also made me realize that we're not locking the read on `w.watchers` here, so I need to fix that.

tgross: I don't think we can't optimize this with a length check after all. Imagine the following scenario:

* watchers for deployment A and B exist, len(deployments) == 2, len(watchers) == 2
* A gets purged, B completes (is no longer active), C gets added
* after we call `add`/`remove`, we have len(deployments) == 2, len(watchers) == 2
* so now checking the lengths would be equal and we'd skip removing A

I've pulled this block of code out to a function to make the locking nicer, and I've added a note about why we can't optimize it with the length check.

shoenig (approved):
LGTM!

PR #20350, opened 2024-04-10
modifyIndex-based pagination

philrenaud:
- Adds support for the new modifyIndex-based pagination provided by the /statuses endpoint with a custom First/Prev/Next/Last pagination template on the jobs index
- Adds a new `withPagination` hook for Mirage endpoints (that builds on top of `withBlockingSupport`) that takes a customizable property to paginate upon and provide as `x-nomad-nexttoken`

### DIFF
--- a/ui/app/controllers/jobs/index.js
+++ b/ui/app/controllers/jobs/index.js
@@ -148,7 +148,7 @@ export default class JobsIndexController extends Controller {
       let prevPageToken = await this.loadPreviousPageToken();
       // If there's no nextToken, we're at the "start" of our list and can drop the cursorAt
       if (!prevPageToken.meta.nextToken) {
-        this.cursorAt = null;
+        this.cursorAt = undefined;
       } else {
         // cursorAt should be the highest modifyIndex from the previous query.
         // This will immediately fire the route model hook with the new cursorAt
@@ -162,7 +162,7 @@ export default class JobsIndexController extends Controller {
       }
       this.cursorAt = this.nextToken;
     } else if (page === 'first') {
-      this.cursorAt = null;
+      this.cursorAt = undefined;
     } else if (page === 'last') {
       let prevPageToken = await this.loadPreviousPageToken(true);
       this.cursorAt = prevPageToken
@@ -248,7 +248,7 @@ export default class JobsIndexController extends Controller {
   async loadPreviousPageToken(last = false) {
     let next_token = +this.cursorAt + 1;
     if (last) {
-      next_token = null;
+      next_token = undefined;
     }
     let prevPageToken = await this.store.query(
       'job',
--- a/ui/app/controllers/jobs/index.js
+++ b/ui/app/controllers/jobs/index.js
@@ -91,7 +91,6 @@ export default class JobsIndexController extends Controller {
       'name',
       this.system.shouldShowNamespaces ? 'namespace' : null,
       'status',
-      'modify index',
       'type',
       this.system.shouldShowNodepools ? 'node pool' : null, // TODO: implement on system service
       'priority',
@@ -164,7 +163,7 @@ export default class JobsIndexController extends Controller {
     } else if (page === 'first') {
       this.cursorAt = undefined;
     } else if (page === 'last') {
-      let prevPageToken = await this.loadPreviousPageToken(true);
+      let prevPageToken = await this.loadPreviousPageToken({ last: true });
       this.cursorAt = prevPageToken
         .sortBy('modifyIndex')
         .get('lastObject').modifyIndex;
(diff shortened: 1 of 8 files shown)


ui/app/controllers/jobs/index.js
```diff
@@ -92,6 +91,7 @@ export default class JobsIndexController extends Controller {
       'name',
       this.system.shouldShowNamespaces ? 'namespace' : null,
       'status',
+      'modify index',
```
philrenaud, line 94: Phil to remove

gulducat (commented):
just a happy thought and one possible-duplicate call. bring on the tests!

ui/mirage/config.js
```diff
@@ -30,19 +30,29 @@ export function filesForPath(allocFiles, filterPath) {
...
+    fn,
+    { pagination = false, tokenProperty = 'ModifyIndex' } = {}
+  ) {
     return function (schema, request) {
+      let handler = fn;
+      if (pagination) {
+        handler = withPagination(handler, tokenProperty);
+      }
+
+      let response = handler.apply(this, arguments);
```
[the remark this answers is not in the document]
philrenaud: Good catch, thank you!

ui/tests/acceptance/jobs-list-test.js
```diff
@@ -736,26 +736,205 @@ module('Acceptance | jobs list', function (hooks) {
...
+        assert.dom('[data-test-pager="last"]').isNotDisabled();
+        // Clicking next brings me to another full page
+        await click('[data-test-pager="next"]');
+        assert.dom('.job-row').exists({ count: 10 });
+        assert.dom('[data-test-pager="first"]').isNotDisabled();
+        assert.dom('[data-test-pager="previous"]').isNotDisabled();
+        assert.dom('[data-test-pager="next"]').isNotDisabled();
+        assert.dom('[data-test-pager="last"]').isNotDisabled();
+        // clicking next again brings me to the last page, showing jobs 20-25
+        await click('[data-test-pager="next"]');
```
gulducat, line 777: these are all "next" - how about adding a "previous" click too? maybe a couple, all the way back to page 1?  maybe even also a "first" and a "last" ?

I think I just like the simplicity of this "plenty of jobs" test, so want to push all the buttons with the concise disabled-or-not assertions.

philrenaud: Yeah, one of my back-of-napkin "test for this" cases was actually a next-prev-next loop. Let me add that!

ui/tests/acceptance/jobs-list-test.js
```diff
@@ -735,4 +735,477 @@ module('Acceptance | jobs list', function (hooks) {
...
+          assert.deepEqual(
+            modifyIndexes,
+            Array(10)
+              .fill()
+              .map((_, i) => i + 2)
+              .reverse(),
+            'Jobs are sorted by modify index'
+          );
+
+          // Simulate one of the on-page jobs getting its modify-index bumped. It should remain in place.
```
gulducat, line 1036: I expected to find more woopsies in all this copypasta, but this here comment is the only one I spotted -- in this test case, the job should not "remain in place" it should get bumped to the top, as asserted down on line 1051 (modify index 5 becomes 12)

ui/tests/acceptance/jobs-list-test.js
```diff
@@ -735,4 +735,477 @@ module('Acceptance | jobs list', function (hooks) {
...
+
+          setTimeout(async () => {
+            rows = document.querySelectorAll('.job-row');
+            modifyIndexes = Array.from(rows).map((row) =>
+              parseInt(row.getAttribute('data-test-modify-index'))
+            );
+            assert.deepEqual(
+              modifyIndexes,
+              [12, 11, 10, 9, 8, 7, 6, 4, 3, 2],
+              'Jobs are sorted by modify index, on-page job remains in-place, and off-page pending'
```
gulducat, line 1052: woop, same misleading message here too.  I think the actual logic is right for the live-updates-enabled case we're in here, just not the comment and this assertion message.

gulducat (approved):
aside from a couple misleading messages, LGTM!

ui/tests/acceptance/jobs-list-test.js
philrenaud: Thanks for the catch, and yeah validated this is the behaviour from the actual endpoint too.

That said: I have eyes on doing some in-browser post-fetch magic at some point. Remembering the order the jobs were requested, and keeping them in that order regardless of the subsequent return. That way the user wouldn't see things dancing up to the top of the page if they had jostling disabled.

But that's for later if ever!

PR #20351, opened 2024-04-11
mock_driver: fix fingreprint key

lgfa29:
The `mock_driver` is an internal task driver used mostly for testing and simulating workloads. During the allocrunner v2 work (#4792) its name changed from `mock_driver` to just `mock` and then back to `mock_driver`, but the fingreprint key was kept as `driver.mock`.

This results in tasks configured with `driver = "mock"` to be scheduled (because Nomad thinks the client has a task driver called `mock`), but fail to actually run (because the Nomad client can't find a driver called `mock` in its catalog).

Fingerprinting the right name prevents the job from being scheduled in the first place.

Also removes mentions of the mock driver from documentation since its an internal driver and not available in any production release.

No changelog since this is an internal change.

### DIFF
--- a/client/driver_manager_test.go
+++ b/client/driver_manager_test.go
@@ -174,10 +174,6 @@ func TestDriverManager_NodeAttributes_Run(t *testing.T) {
 			return false, fmt.Errorf("mock_driver driver is not marked as healthy: %+v", d)
 		}
 
-		if d.Attributes["driver.mock_driver"] != "" {
-			return false, fmt.Errorf("mock driver driver attributes contain duplicate health info: %#v", d.Attributes)
-		}
-
 		// check raw_exec
 		if node.Attributes["driver.raw_exec"] == "" {
 			return false, fmt.Errorf("raw_exec is not present in attributes: %#v", node.Attributes)
--- a/drivers/mock/driver.go
+++ b/drivers/mock/driver.go
@@ -345,7 +345,7 @@ func (d *Driver) buildFingerprint() *drivers.Fingerprint {
 		desc = "disabled"
 	} else {
 		health = drivers.HealthStateHealthy
-		attrs["driver.mock"] = pstructs.NewBoolAttribute(true)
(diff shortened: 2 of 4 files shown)


lgfa29:
Hum...I'm not sure what the failing test is checking for. 

@schmichael do you have any historical knowledge of why we check for this?

https://github.com/hashicorp/nomad/blob/8298d39e785badfbe47f0c30f2bd2bad5f9b042c/client/driver_manager_test.go#L177-L179

AFAIC every other task driver does the same, and would fail a check like this. The test has only been passing because we're checking for the wrong key 😬 

https://github.com/hashicorp/nomad/blob/8298d39e785badfbe47f0c30f2bd2bad5f9b042c/drivers/docker/fingerprint.go#L130
https://github.com/hashicorp/nomad/blob/8298d39e785badfbe47f0c30f2bd2bad5f9b042c/drivers/exec/driver.go#L366
https://github.com/hashicorp/nomad/blob/8298d39e785badfbe47f0c30f2bd2bad5f9b042c/drivers/java/driver.go#L363
https://github.com/hashicorp/nomad/blob/8298d39e785badfbe47f0c30f2bd2bad5f9b042c/drivers/qemu/driver.go#L264
https://github.com/hashicorp/nomad/blob/8298d39e785badfbe47f0c30f2bd2bad5f9b042c/drivers/rawexec/driver.go#L227

schmichael:
@lgfa29 No clue I'm afraid and feel free to change it. It seems we chose to go with `"driver.mock" = "true"` as its attribute (note the lack of `_driver` suffix there) to match other drivers... but only in the Attributes struct.

However in the `mock.Node` fixture we use extensively throughout tests we use the `"driver.mock_driver"` form in Attributes!

So I think the takeaway here is that since this driver is excluded from release builds we've been sloppy with it. A cleanup would be nice, but I think we can blame its history on devs just tweaking individual bits to match certain testing needs in certain circumstances.

jrasell (approved):

Issue #20355, opened 2024-04-11
ephemeral_disk documentation needs clarification


GuyAtTheFront:
### Proposal
The official docs for [sticky](https://developer.hashicorp.com/nomad/docs/job-specification/ephemeral_disk#sticky) and [migrate](https://developer.hashicorp.com/nomad/docs/job-specification/ephemeral_disk#migrate) uses the language:
```
... specifies that Nomad should make a best-effort attempt to ...
```

Is it possible to specify what "best-effort attempt" means? And/or possibly give examples of instances where Nomad will intentionally not or will fail to uphold `sticky` or `migrate`?

### Use-cases
It is common for services to write state data to disk to help recover from failures. State data might include checkpoints, persistent queues, or unflushed data chunks. 

Knowing exactly why, when, how and how often these services might **not recover** from failures is a critical consideration for projects, since those instance might result in permanent data loss. It also determines if additional replication factor or external state store is required to mitigate the risk. 

### Attempted Solutions
Current workaround is to avoid the `ephemeral_disk` block entirely because the behavior is unknown, and the possibility of "data-loss"  without root-cause is not ideal for production. Instead, a single nfs is mounted across all nomad client nodes, then mounted to client > group > task for state data to be written to. 

However, this solution adds a dependency that sits outside of the nomad ecosystem, which makes the job less reusable. The dependency also adds a layer of bureaucracy for deployments, since it is not in the control of the infra devs (is done by sysadmin). 

(Hi again, tgross!)

tgross:
Hi @GuyAtTheFront! Agreed that those docs could use some improvement, and I've done so in #20357. But the short version is that you definitely don't want to use ephemeral disk migration for anything you can't recreate at the destination allocation. It's best used for things like on-disk cache.

If you have data you need for correctness like non-idempotent checkpoints, I strongly recommend instead using something like the [Task API](https://developer.hashicorp.com/nomad/api-docs/task-api) in combination with [Variable Locks](https://developer.hashicorp.com/nomad/api-docs/variables/locks). Or use persistent disk via host volumes or CSI.

GuyAtTheFront:
I've somehow missed the docs you linked, will have a read. 

[...]

PR #20357, opened 2024-04-11
docs: clarify "best effort" for ephemeral disk migration

tgross:
The docs for ephemeral disk migration use the term "best effort" without outlining the requirements or the cases under which the migration can fail. Update the docs to make it obvious that ephemeral disk migration is subject to data loss.

Fixes: https://github.com/hashicorp/nomad/issues/20355
Preview link: https://nomad-nsel0s9mv-hashicorp.vercel.app/nomad/docs/job-specification/ephemeral_disk#migrate


shoenig (approved):
LGTM

PR #20376, opened 2024-04-12
tests: fix broken test in deploymentwatcher backport

tgross:
The test added in #20348 for the deployment watcher was backported to 1.5.x but is hitting a compilation error because the signature of a state store method was changed.


Juanadelacuesta (approved):
To the point :)

PR #20296, opened 2024-04-04
tproxy: E2E tests

tgross:
_WIP_

PR #20378, opened 2024-04-12
E2E: clean up go mod cache after building `consul-cni`

tgross:
In #20296 we added a Go tool chain to the AMI we use for E2E tests, so that we can build `consul-cni` for tproxy testing. This is intended to be temporary until `consul-k8s` 1.4.2 is officially released. But the Go cache from building `consul-k8s` uses up roughly 1.5GiB of space and the test machines have fairly small disks. This causes the Nomad clients to aggressively GC client allocations that stop, which breaks tests that run batch workloads and then read their logs.


pkazmierczak (approved):
🎉

gulducat (approved):
what a fun way to find out the disk was "full" :upside_down_face:

PR #20379, opened 2024-04-12
CNI: fix bugs in parsing strings to port number integers

tgross:
Ports are a maximum of uint16, but we have a few places in the recent tproxy code where we were parsing them as 64-bit wide integers and then downcasting them to `int`, which is technically unsafe and triggers code scanning alerts. In practice we've validated the range elsewhere and don't build for 32-bit platforms. This changeset fixes the parsing to make everything a bit more robust and silence the alert.

Fixes: https://github.com/hashicorp/nomad-enterprise/security/code-scanning/444


gulducat (approved):

tgross:
No backport needed as this is in unreleased code.

PR #20383, opened 2024-04-12
[ui] don't assume that an allocation has a task events array when checking if it hasBeenRestarted

philrenaud:
the hasBeenRestarted allocation property checks against its task events, which can sometimes be null

This can result on a job-index-page error when calculating the `filteredAllocations` to show (we don't show those allocations as "failed" that have been restarted/rescheduled if other more recent ones exist, up to a job's desired count)

(The `?.` is a defensive pattern that fails truthiness checks when the object in question doesn't exist, so `s.events?.content` will return `null` instead of erroring)

### DIFF
--- a/.changelog/20383.txt
+++ b/.changelog/20383.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fix an issue where the job status box would error if an allocation had no task events
+```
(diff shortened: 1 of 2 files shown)


tgross (approved):
LGTM!

lgfa29 (approved):
Elvis operator to the rescue

PR #20393, opened 2024-04-15
comment on Job.ModifyIndex vs Job.JobModifyIndex

gulducat:
Took a decent bit of brain cycles recently to understand this distinction, so writing it down to save future curious/confused persons. :writing_hand:


tgross (approved):
Nice!

Issue #20308, opened 2024-04-05
Auto-promote  for canaries changes to manual promotion when a canary can't initially get placed in the cluster


jvineet:
### Nomad version
```
Nomad v1.7.6
BuildDate 2024-03-12T07:27:36Z
Revision 594fedbfbc4f0e532b65e8a69b28ff9403eb822e
```

### Operating system and Environment details

```
Ubuntu 22.04.4 LTS (Jammy Jellyfish)
```

### Issue

We have had frequent issues recently where our nomad job deployments often end up getting their canaries stuck in "Manual promotion" state even when we have canary auto-promotion turned on. This behavior seems to have especially taken off after upgrading to Nomad v1.7.  

Upon investigating this problem, we noticed that the canaries were getting stuck in a manual promotion state only when the canary was unable to get placed in the cluster immediately due to resource constraints and had to wait for a new nomad agent to come up in order to make extra capacity to place the canary.  So the eval status would show "True" for "Placement Failures". The canary would eventually be deployed to a new nomad agent after a while and turn healthy, but the promotion strategy for the entire deployment will switch from auto-promote to manual promotion. 

<img width="525" alt="Screen Shot 2024-04-05 at 4 36 16 PM" src="https://github.com/hashicorp/nomad/assets/6412461/dcf8db0c-86a9-424b-8914-813c2629bb72">

We were able to reproduce this behavior by adding a placement constraint on a job with bogus `meta` label so the job had a placement failure when deploy started, and sure enough it put the “auto-promote” enabled canary in a “manual promotion” state.
Here are some screenshots:

<img width="1178" alt="Screen Shot 2024-04-05 at 4 29 24 PM" src="https://github.com/hashicorp/nomad/assets/6412461/b98bd917-d791-4af8-b28c-d6b26964040e">

<img width="334" alt="Screen Shot 2024-04-05 at 4 25 46 PM" src="https://github.com/hashicorp/nomad/assets/6412461/a6eacff9-3010-4182-a300-8f93fe278b8d">

<img width="1145" alt="Screen Shot 2024-04-05 at 4 30 33 PM" src="https://github.com/hashicorp/nomad/assets/6412461/dc5e3b5c-5074-4581-9d5a-9a0cda0d5a90">

<img width="529" alt="Screen Shot 2024-04-05 at 4 19 56 PM" src="https://github.com/hashicorp/nomad/assets/6412461/afddc613-90fe-4a81-88c7-bd5878c07e4b">

[...]

PR #20408, opened 2024-04-16
[ui] Fix a bug where promotion would be asked with no new canaries

philrenaud:
The "Are all your canaries healthy?" part of the "should we prompt the user to manually promote their deployment" depended on an `array.every` check without a condition that the array have some members.

This meant that, as shown in #20308, a user can run into a situation where a new canary allocation fails to place, but the system insists that "there are no unhealthy canaries!" so it's manual promotion time.

This checks for at least one new-version allocation with canary status that hasn't been rescheduled before flipping `canariesHealthy` to true. If there are none, the user will continue to be met with a "Your deployment's canaries are still being health checked" pending status.

### DIFF
--- a/.changelog/20408.txt
+++ b/.changelog/20408.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fix a bug where the UI would prompt a user to promote a deployment with unplaced canaries
+```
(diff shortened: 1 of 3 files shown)


tgross (commented):
LGTM!

gulducat (approved):
promote em if ya got em! (but only if ya got em)

PR #20410, opened 2024-04-16
state: enable more reverse sorting

gulducat:
No one asked for this, but I did it while building out the new job statuses endpoint.  I ended up using a different (new) index there, so this may be considered extra unnecessary, and it only actually adds sorting to jobs endpoint, but I think it moves the ball forward slightly, so here it is.

I could be convinced to either:
* go further, add sorting to more endpoints
* go backwards, throw out everything except the new lil helpers I want to use

If we keep this, should I backport it?  Seems like something that could cause backport headaches down the road if I don't.

### DIFF
--- a/nomad/state/sorting.go
+++ b/nomad/state/sorting.go
@@ -1,0 +1,43 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package state
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/go-memdb"
+	"github.com/hashicorp/nomad/nomad/structs"
+)
+
+// SortOption represents how results can be sorted.
+type SortOption bool
+
+const (
+	// SortDefault indicates that the result should be returned using the
+	// default go-memdb ResultIterator order.
+	SortDefault SortOption = false
+
+	// SortReverse indicates that the result should be returned using the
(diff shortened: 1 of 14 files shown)


nomad/state/sorting.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+// QueryOptionSort returns the appropriate SortOption for given QueryOptions.
+func QueryOptionSort(qo structs.QueryOptions) SortOption {
+	return SortOption(qo.Reverse)
+}
+
+// GetSorted executes either txn.Get() or txn.GetReverse()
+// depending on the provided SortOption.
+func GetSorted(txn *txn, sort SortOption, table, index string, args ...any) (memdb.ResultIterator, error) {
```
gulducat, line 33: this little func is the main helper that I like, the rest grew out of this.

tgross: Should this really be exported? This exposes internal helper functions of the state store to the caller. But more importantly, the `txn` type isn't exported, so it's not like you can call this outside of the `nomad/state` package anyways.

tgross:
> go further, add sorting to more endpoints

I would not do this unless there's a feature that's looking for it. At least not in this PR.

tgross (approved):
LGTM once we fix the unusable function export.

Issue #20432, opened 2024-04-18
Dispatch documentation and API documentation differs about the maximum size of payload


agarfer1:
This page (https://nomadproject.io/docs/commands/job/dispatch/) says:

> The payload has a size limit of 16384 bytes (16KiB).

Whereas the API documentation for Dispatch (https://nomadproject.io/api-docs/jobs/#payload) says:

> [Payload](https://developer.hashicorp.com/nomad/api-docs/jobs#payload) (string: "") - Specifies a base64 encoded string containing the payload. This is limited to 65536 bytes (64KiB).

Related: #7614

tgross:
Hi @agarfer1! Good catch, it's 16KiB. I'll fix the docs.

PR #20433, opened 2024-04-18
docs: fix incorrect dispatch payload limit in API docs

tgross:
The dispatch payload limit is limited to 16KiB, not 64KiB. It's correct in the command docs but incorrect in the API docs. It looks like this was accidentally updated in https://github.com/hashicorp/nomad/commit/c2491e9146ae5db92bb04d2eda2b5ccd58d07851 which was supposed to be for the Variables payload size.

Ref: https://github.com/hashicorp/nomad/blob/v1.7.7/nomad/job_endpoint.go#L36-L38
Fixes: https://github.com/hashicorp/nomad/issues/20432


gulducat (approved):
:+1:

Issue #12013, opened 2022-02-05
Many metrics are in Milliseconds time unit granularity but docs specify Nanoseconds


schmichael:
### Nomad version
v1.2.5

### Operating system and Environment details
Website

### Issue

Our [metrics reference documentation](https://github.com/hashicorp/nomad/blob/v1.2.5/website/content/docs/operations/metrics-reference.mdx) improperly states `Nanoseconds` is the unit of many metrics when it should be `Milliseconds`.

For example `raft.commitTime` is stated to be of type `Summary` and granularity `Nanoseconds` in Nomad's docs: https://github.com/hashicorp/nomad/blob/v1.2.5/website/content/docs/operations/metrics-reference.mdx#key-metrics

However if you look in the JSON of a dev agent, all of the values are < 0.1! Even if the operations were that fast (they're not) we're only able to measure with 1ns granularity so being <1ns is impossible to measure. The heading is also `Samples`, not `Summary`.

```json
    "Samples": [
        ...
        {
            "Count": 3,
            "Labels": {
                "host": "dean"
            },
            "Max": 0.09279300272464752,
            "Mean": 0.07419466599822044,
            "Min": 0.041811998933553696,
            "Name": "nomad.raft.commitTime",
            "Rate": 0.022258399799466134,
            "Stddev": 0.02814731797244929,
            "Sum": 0.22258399799466133
        }
```

The Prometheus format agrees with the json format (dumps are taken a few seconds apart so the values skew):

```
nomad_raft_commitTime{host="dean",quantile="0.5"} 0.04487600177526474
nomad_raft_commitTime{host="dean",quantile="0.9"} 0.11641400307416916
nomad_raft_commitTime{host="dean",quantile="0.99"} 0.11641400307416916
nomad_raft_commitTime_sum{host="dean"} 2.296792024746537
nomad_raft_commitTime_count{host="dean"} 33
```

### Expected Results

Consul seems to get things right: https://www.consul.io/docs/agent/telemetry

Both [Nomad](https://github.com/hashicorp/nomad/blob/v1.2.5/command/agent/command.go#L1043) and [Consul](https://github.com/hashicorp/consul/blob/01e633ee79c03d79d60b53db2c319cc35d59f989/lib/telemetry.go#L349) use [go-metric's default configuration (Milliseconds) for their Timer granularity.](https://github.com/armon/go-metrics/blob/d1e5690ba72f7651735eb0a618d5523a6bdccf58/start.go#L64)

[...]

PR #20446, opened 2024-04-18
docs: correct nanoseconds to milliseconds for MeasureSince metrics

pkazmierczak:
Our documentation incorrectly stated that metrics based on `MeasureSince` or `MeasureSinceWithLabels` go-metrics methods are in nanoseconds, whereas in fact they are [in milliseconds](https://github.com/hashicorp/go-metrics/blob/9d55fb233b15597348d8c257632d8211a07d3395/metrics.go#L167). Thanks for spotting this, @jrasell. 

(I also corrected `Summary` metric type to `Timer` where needed). 

Fixes: https://github.com/hashicorp/nomad/issues/12013


tgross (approved):
LGTM!

PR #20456, opened 2024-04-18
acl: remove remaining unused nil ACL object handling

tgross:
As of #18754 which shipped in Nomad 1.7, we no longer need to nil-check the object returned by ResolveACL if there's no error return, because in the case where ACLs are disabled we return a special "ACLs disabled" ACL object. Checking nil is not a bug but should be discouraged because it opens us up to future bugs that would bypass ACLs.

We fixed a bunch of these cases in https://github.com/hashicorp/nomad/pull/20150 but I didn't update the semgrep rule, which meant we missed a few more. Update the semgrep rule and fix the remaining cases.


shoenig (approved):
LGTM! :broom:

PR #20464, opened 2024-04-19
acl: token global mode can not be changed

gulducat:
True up `nomad acl token update` CLI and some docs with most docs and API reality.

Most API docs already rightly say "Note that a token cannot be switched from global to local or visa versa."

Here's what happens if you try:

```
$ nomad acl token update -global=false $accessor
Error updating token: Unexpected response code: 400 (token 0 invalid: 1 error occurred:
        * cannot toggle global mode)
```

### DIFF
--- a/website/content/docs/commands/acl/token/create.mdx
+++ b/website/content/docs/commands/acl/token/create.mdx
@@ -29,7 +29,7 @@ The `acl token create` command requires no arguments.
   "management".
 
 - `-global`: Sets the global mode of the token. Global tokens are replicated
-  to all regions. This can not be changed after creation. Defaults false.
+  to all regions. This can not be changed after creation. Defaults to false.
 
 - `-policy`: Specifies a policy to associate with the token. Can be specified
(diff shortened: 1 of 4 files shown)


website/content/docs/commands/acl/token/create.mdx
```diff
@@ -28,8 +28,8 @@ The `acl token create` command requires no arguments.
 - `-type`: Sets the type of token. Must be one of "client" (default), or
   "management".
 
-- `-global`: Toggles the global mode of the token. Global tokens are replicated
-  to all regions. Defaults false.
+- `-global`: Sets the global mode of the token. Global tokens are replicated
+  to all regions. This can not be changed after creation. Defaults false.
```
pkazmierczak, line 32:
```suggestion
  to all regions. This can not be changed after creation. Defaults to false.
```

(just a suggestion)

pkazmierczak (approved):
LGTM!

PR #20465, opened 2024-04-19
security: fine tune security-scanner

dduzgun-security:
This PR updates the security scanner configuration for better accuracy and coverage.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,3 +1,6 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: BUSL-1.1
+
 version: 2
 updates:
   - package-ecosystem: gomod
@@ -37,3 +40,13 @@ updates:
     labels:
       - "theme/dependencies"
       - "theme/website"
+  - package-ecosystem: github-actions
+    open-pull-requests-limit: 5
+    directory: /
+    labels:
+      - "theme/dependencies"
+      - "theme/ci"
+    schedule:
+      interval: "weekly"
+      day: "sunday"
+      time: "09:00"
--- a/.github/workflows/security-scan.yml
+++ b/.github/workflows/security-scan.yml
@@ -1,0 +1,67 @@
+name: Security Scan
+
+on:
+  push:
+    branches:
+      - main
+      - release/**
+    paths-ignore:
+      - 'README.md'
+      - 'CHANGELOG.md'
+      - '.changelog/**'
+      - '.tours/**'
+      - 'contributing/**'
+  pull_request:
+    branches:
+      - main
+      - release/**
+    paths-ignore:
+      - 'README.md'
+      - 'CHANGELOG.md'
+      - '.changelog/**'
+      - '.tours/**'
+      - 'contributing/**'
+
+# cancel existing runs of the same workflow on the same ref
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  scan:
+    runs-on: ${{ endsWith(github.repository, '-enterprise') && fromJSON('["self-hosted", "ondemand", "linux"]') || 'ubuntu-22.04' }}
+    # The first check ensures this doesn't run on community-contributed PRs, who
+    # won't have the permissions to run this job.
+    if: ${{ (github.repository != 'hashicorp/nomad' || (github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name))
+      && (github.actor != 'dependabot[bot]') && (github.actor != 'hc-github-team-nomad-core') }}
+
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - uses: hashicorp/setup-golang@36878950ae8f21c1bc25accaf67a4df88c29b01d # v3
+
+      - name: Clone Security Scanner repo
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          repository: hashicorp/security-scanner
+          token: ${{ secrets.PRODSEC_SCANNER_READ_ONLY }}
+          path: security-scanner
+          ref: main
+
+      - name: Scan
+        id: scan
+        uses: ./security-scanner
+        with:
+          repository: "$PWD"
+        env:
+          SECURITY_SCANNER_CONFIG_FILE: .github/scan.hcl
+          # See scan.hcl at repository .github location for config.
+
+      - name: SARIF Output
+        shell: bash
+        run: |
+          jq . < results.sarif
+
+      - name: Upload SARIF file
+        uses: github/codeql-action/upload-sarif@8fd294e26a0e458834582b0fe4988d79966c7c0a # codeql-bundle-v2.18.4
+        with:
+          sarif_file: results.sarif
--- a/.release/security-scan.hcl
+++ b/.release/security-scan.hcl
@@ -2,20 +2,25 @@
 # SPDX-License-Identifier: BUSL-1.1
 
 container {
+  local_daemon = true
+
   secrets {
-    all = false
+    all               = true
+    skip_path_strings = ["/website/content/"]
   }
 
-  dependencies    = false
-  alpine_security = false
+  dependencies    = true
+  alpine_security = true
 }
 
 binary {
   go_modules = true
-  osv        = false
+  osv        = true
+  go_stdlib  = true
   nvd        = false
 
   secrets {
-    all = true
+    all               = true
+    skip_path_strings = ["/website/content/"]
   }
 }
--- a/.github/scan.hcl
+++ b/.github/scan.hcl
@@ -1,0 +1,55 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: BUSL-1.1
+
+# Configuration for security scanner.
+# Run on PRs and pushes to `main` and `release/**` branches.
+# See .github/workflows/security-scan.yml for CI config.
+
+# To run manually, install scanner and then run `scan repository .`
+
+# Scan results are triaged via the GitHub Security tab for this repo.
+# See `security-scanner` docs for more information on how to add `triage` config
+# for specific results or to exclude paths.
+
+# This file controls scanning the repository only, not release artifacts. See
+# .release/security-scan.hcl for the scanner config for release artifacts, which
+# will block releases.
+
+repository {
+  go_modules             = true
+  npm                    = true
+  osv                    = true
+  go_stdlib_version_file = ".go-version"
+
+  secrets {
+    all               = true
+    skip_path_strings = ["/website/content/"]
+  }
+
+  github_actions {
+    pinned_hashes = true
+  }
+
+  dependabot {
+    required     = true
+    check_config = true
+  }
+
+  dockerfile {
+    pinned_hashes = true
+    curl_bash     = true
+  }
+
+  # Triage items that are _safe_ to ignore here. Note that this list should be
+  # periodically cleaned up to remove items that are no longer found by the scanner.
+  triage {
+    suppress {
+      paths = [
+        "ui/tests/*",
+        "internal/testing/*",
+        "testutil/*",
+        "website/content/*",
+      ]
+    }
+  }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -134,3 +134,6 @@ e2e/remotetasks/input/ecs.vars
 
 # Tools files
 tools/missing/missing
+
+# allow security scanner file
+!scan.hcl


.github/dependabot.yml
```diff
@@ -37,3 +40,13 @@ updates:
...
+  - package-ecosystem: github-actions
+    open-pull-requests-limit: 5
+    directory: /
+    labels:
+      - "theme/dependencies"
+      - "theme/ci"
+    schedule:
+      interval: "weekly"
+      day: "sunday"
+      time: "09:00"
```
tgross, line 52: @dduzgun-security you'll want to fix your editor to make sure files have their EOF on save. Not having this ends up with surprising behavior for various Unix tools:

```
wc -l .github/dependabot.yml
51 .github/dependabot.yml
```

.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,65 @@
...
+
+on:
+  push:
+    branches:
+      - main
+      - release/**
+  pull_request:
+    branches:
+      - main
+      - release/**
```
tgross, line 11: We don't PR _from_ `main` and `release/**`, so I don't think we want this filter.

.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,65 @@
...
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  scan:
+    runs-on: [ custom, linux, xl, 22.04 ]
+    # The first check ensures this doesn't run on community-contributed PRs, who
+    # won't have the permissions to run this job.
+    if: ${{ (github.repository != 'hashicorp/nomad' || (github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name))
+      && (github.actor != 'dependabot[bot]') && (github.actor != 'hc-github-team-nomad-core') }}
```
tgross, line 24: This reinforces my comment above about filtering on the PR; community PRs would target `main` anyways.

.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,65 @@
...
+          echo "Building with Go $(cat .go-version)"
+          echo "go-version=$(cat .go-version)" >> "$GITHUB_OUTPUT"
+
+      - name: Set up Go
+        uses: actions/setup-go@fac708d6674e30b6ba41289acaab6d4b75aa0753 # v4.0.1
+        with:
+          go-version: ${{ steps.get-go-version.outputs.go-version }}
+
+      - name: Clone Security Scanner repo
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
```
tgross, line 43: A quick survey shows we're using 4.1.1 elsewhere. Dependabot PRs are very expensive for us to deal with because of backports and CE->ENT merges. So we should absolutely minimize the number of different pins we have. (See also https://github.com/hashicorp/nomad-enterprise/issues/1209 (internal))

.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,65 @@
...
+        with:
+          repository: hashicorp/security-scanner
+          token: ${{ secrets.HASHIBOT_PRODSEC_GITHUB_TOKEN }}
```
tgross, line 46: Is this going to work when we merge this over to the ENT repo, or will the Vault requirement there make us have a separate workflow for getting the token?

.release/security-scan.hcl
```diff
@@ -2,20 +2,25 @@
 # SPDX-License-Identifier: BUSL-1.1
```
tgross, line 2: Run `hclfmt` on these HCL files please.

.release/security-scan.hcl
```diff
@@ -2,20 +2,25 @@
...
-  dependencies    = false
-  alpine_security = false
+  dependencies    = true
+  alpine_security = true
 }
 
 binary {
   go_modules = true
-  osv        = false
+  osv        = false # TODO: set to true when osv is resolved
```
tgross, line 18: So that we know who to poke about this:

```suggestion
  osv        = false # TODO(deniz): set to true when osv is resolved
```

scan.hcl
```diff
@@ -0,0 +1,53 @@
+# Copyright (c) HashiCorp, Inc.
```
tgross, line 1: Run hclfmt on this file please.

Also, this is a super nitpick :grinning:... but is it at all possible to have this somewhere other than the root of the repo? It seems like more and more tools want to pollute the root directory with their little dotfiles and it just turns into a junk drawer. If not possible, not the end of the world

.github/dependabot.yml
tgross, line 52: Generally speaking we prefer PRs to be atomic. Changing the dependabot rules is a different body of work than tuning the security scanner. Not a huge deal here but just for future reference.

scan.hcl
```diff
@@ -0,0 +1,53 @@
...
+# See .github/workflows/security-scan.yml for CI config.
+
+# To run manually, install scanner and then run `scan repository .`
+
+# Scan results are triaged via the GitHub Security tab for this repo.
+# See `security-scanner` docs for more information on how to add `triage` config
+# for specific results or to exclude paths.
+
+# .release/security-scan.hcl controls scanner config for release artifacts, which
+# unlike the scans configured here, will block releases.
```
tgross, line 15: This is worded slightly confusingly:

```suggestion
# This file controls scanning the repository only, not release artifacts. See
# .release/security-scan.hcl for the scanner config for release artifacts, which
# will block releases.
```

tgross (changes requested):
@dduzgun-security I've left a few comments on this which we should resolve. Take the PR out of draft when you've addressed those and are ready for a re-review.

I had a concern that this was going to add significantly to the CI times, but it looks like from [this run](https://github.com/hashicorp/nomad/actions/runs/8838679887/job/24270380268?pr=20465) that these are <2min, which is fine when they run in parallel.

.github/workflows/security-scan.yml
dduzgun-security: Good catch, bumped it to the latest version: https://github.com/hashicorp/nomad/pull/20465/commits/dfff95becaf6f28364e2203cd83ae0bb8574f997

.release/security-scan.hcl
dduzgun-security: Should be done, https://github.com/hashicorp/nomad/pull/20465/commits/dfff95becaf6f28364e2203cd83ae0bb8574f997

.release/security-scan.hcl
dduzgun-security: This should be resolved, switched it to `true` in here https://github.com/hashicorp/nomad/pull/20465/commits/dfff95becaf6f28364e2203cd83ae0bb8574f997

scan.hcl
dduzgun-security: hclfmt ✅ 

I think it should be doable in the CI since we have the `SECURITY_SCANNER_CONFIG_FILE` option in the security-scanner. I'll need to double check if this would cause any issues in our scheduled scanners that run twice a month internally, in case it fetches the config from the root of the project. I think @picatz could provide some additional details here?

.github/dependabot.yml
dduzgun-security: Good point, atomic PRs makes it easier to revert too, ✍️ noted it for the future

.github/workflows/security-scan.yml
dduzgun-security: Yes, the secret is added in both repositories and should be accessible by the GitHub Action

.github/workflows/security-scan.yml
dduzgun-security: I think this should be for PR from a branch to `main` or `release/**`. The addition for release is to trigger the scan on backports too if I am not mistaken. I can remove it if it's not needed?

tgross: You're right: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore

> When using the pull_request and pull_request_target events, you can configure a workflow to run only for pull requests that target specific branches.

scan.hcl
```diff
@@ -0,0 +1,55 @@
...
+
+  # Triage items that are _safe_ to ignore here. Note that this list should be
+  # periodically cleaned up to remove items that are no longer found by the scanner.
+  triage {
+    suppress {
+      paths = [
+        "ui/tests/*",
+        "internal/testing/*",
+        "testutil/*",
+        "website/content//*",
```
tgross, line 51: Does `//` have a special meaning here?

tgross (commented):
LGTM once the remaining two comments are resolved or dismissed.

scan.hcl
dduzgun-security: Good catch, [7480d2a](https://github.com/hashicorp/nomad/pull/20465/commits/7480d2addad4b1dc82c5424bc4ecc397c97682b1)

scan.hcl
dduzgun-security: This should normally now be handled by https://github.com/hashicorp/nomad/pull/20465/commits/7c13047e43134afd63ba5904691ab9d4a4b9ee6d

.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,79 @@
...
+      - 'contributing/**'
+
+# cancel existing runs of the same workflow on the same ref
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  scan:
+    runs-on: custom-linux-xl-nomad-22.04
```
tgross, line 32: If we're planning on running this in the ENT repo, we may want to do something like this: https://github.com/hashicorp/nomad/blob/main/.github/workflows/test-core.yaml#L52

.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # v5.0.2
+        with:
+          go-version: ${{ steps.get-go-version.outputs.go-version }}
+
+      - name: Clone Security Scanner repo
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          repository: hashicorp/security-scanner
+          token: ${{ secrets.PRODSEC_SCANNER_READ_ONLY }}
```
tgross, line 58: Has this secret been added for both CE and ENT?

.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,79 @@
...
+        # We use .go-version as our source of truth for current Go
+        # version, because "goenv" can react to it automatically.
+        run: |
+          echo "Building with Go $(cat .go-version)"
+          echo "go-version=$(cat .go-version)" >> "$GITHUB_OUTPUT"
+
+      - name: Set up Go
+        uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # v5.0.2
+        with:
+          go-version: ${{ steps.get-go-version.outputs.go-version }}
```
tgross, line 52: I'm pretty sure this whole section can be replaced with:

```suggestion
      - uses: hashicorp/setup-golang@v3
```

Which is what we do in [test-core.yaml#L81](https://github.com/hashicorp/nomad/blob/main/.github/workflows/test-core.yaml#L81)

.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,79 @@
...
+          ref: main
+
+      - name: Scan
+        id: scan
+        uses: ./security-scanner
+        with:
+          repository: "$PWD"
+        env:
+          SECURITY_SCANNER_CONFIG_FILE: .github/scan.hcl
+          # See scan.hcl at repository .github location for config.
```
tgross, line 69: Just for my clarification, this repo contains this [`action.yml`](https://github.com/hashicorp/security-scanner/blob/main/action.yml), so by calling `uses:` here into that directory, GitHub is treating the repo as an Action? And we need to do this workflow where we clone it because it's in a private repo?

tgross (commented):
Looking good... I've left a few comments/questions.

.github/workflows/security-scan.yml
dduzgun-security: Yes, it's an organization level GitHub Action secret available to all HashiCorp repositories

.github/workflows/security-scan.yml
dduzgun-security: https://github.com/hashicorp/nomad/pull/20465/commits/0c2d20ed6347cbe1b780d858a578be2c5a9ee885

.github/workflows/security-scan.yml
dduzgun-security: Yes that is correct, since the repo is private that is the way to retrieve/checkout the repo and use it. It's subject to change in the future when the security-scanner will be open-sourced.

.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,67 @@
...
+  scan:
+    runs-on: ${{ endsWith(github.repository, '-enterprise') && fromJSON('["self-hosted", "ondemand", "linux"]') || 'ubuntu-22.04' }}
+    # The first check ensures this doesn't run on community-contributed PRs, who
+    # won't have the permissions to run this job.
+    if: ${{ (github.repository != 'hashicorp/nomad' || (github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name))
+      && (github.actor != 'dependabot[bot]') && (github.actor != 'hc-github-team-nomad-core') }}
+
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - uses: hashicorp/setup-golang@v3
```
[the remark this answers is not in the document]
dduzgun-security: https://github.com/hashicorp/nomad/pull/20465/commits/0a3a638a71d0ccbc0a99c386af37f62618d74dee

tgross (approved):
LGTM

PR #20474, opened 2024-04-22
test: add MinAllocForJob helper

gulducat:
Useful for making multiple allocations for the same job.

### DIFF
--- a/nomad/mock/alloc.go
+++ b/nomad/mock/alloc.go
@@ -87,7 +87,10 @@ func Alloc() *structs.Allocation {
 }
 
 func MinAlloc() *structs.Allocation {
-	job := MinJob()
+	return MinAllocForJob(MinJob())
+}
+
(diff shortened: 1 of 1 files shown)


nomad/mock/alloc.go
```diff
@@ -87,14 +87,19 @@ func Alloc() *structs.Allocation {
 }
 
 func MinAlloc() *structs.Allocation {
-	job := MinJob()
+	return MinAllocForJob(MinJob())
+}
+
+func MinAllocForJob(job *structs.Job) *structs.Allocation {
 	group := job.TaskGroups[0]
 	task := group.Tasks[0]
```
gulducat, line 95: accepting any ol' Job opens up the possibility of panics if a caller provides a job with no groups or tasks, but the stack trace will not be very mysterious to a caller holding it wrong.

gulducat:
cleaning up old stuff, this PR's time has long past, and the helper made its way in another PR

PR #20479, opened 2024-04-23
cli: fix handling of scaling jobs which don't generate evals.

jrasell:
In some cases, Nomad job scaling will not generate evaluations such as parameterized jobs. This change fixes the CLI behaviour in this case, and copies the job run command for consistency.


pkazmierczak (approved):
LGTM!

PR #20481, opened 2024-04-23
drivers/raw_exec: enable setting cgroup override values

shoenig:
This PR enables configuration of cgroup override values on the `raw_exec`
task driver. **WARNING**: setting cgroup override values eliminates any
guarantee Nomad can make about resource availability for *any* task on
the client node.

For cgroup v2 systems, set a single unified cgroup path using `cgroup_v2_override`.
The path may be either absolute or relative to the cgroup root.

```hcl
config {
  cgroup_v2_override = "custom.slice/app.scope"
}
```

or

```hcl
config {
  cgroup_v2_override = "/sys/fs/cgroup/custom.slice/app.scope"
}
```

For cgroup v1 systems, set a per-controller path for each controller using
`cgroup_v1_override`. The path(s) may be either absolute or relative to
the controller root.

```hcl
config {
  cgroup_v1_override = {
    "pids": "custom/app",
    "cpuset": "custom/app",
  }
}
```

or

```hcl
config {
  cgroup_v1_override = {
    "pids": "/sys/fs/cgroup/pids/custom/app",
    "cpuset": "/sys/fs/cgroup/cpuset/custom/app",
  }
}
```


shoenig:
Example (cgroups v1):

```hcl
job "sleep" {
  group "group" {
    task "task" {
      driver = "raw_exec"
      config {
        command = "sleep"
        args    = ["infinity"]
        cgroup_v1_override = {
          "pids" : "custom/app",
          "cpuset" : "custom/app",
        }
      }
    }
  }
}
```

shoenig:
## Example (cgroups v2)

```hcl
job "sleep" {
  group "group" {
    task "task" {
      driver = "raw_exec"
      config {
        command            = "sleep"
        args               = ["infinity"]
        cgroup_v2_override = "custom.slice/app.scope"
      }
    }
  }
}
```

Ensure our custom unified cgroup exists.

```
➜ sudo mkdir -p /sys/fs/cgroup/custom.slice/app.scope
```

Run job.

Ensure our custom cgroup is in use.

```
➜ pcat custom.slice/app.scope/cgroup.procs
124226
```

drivers/shared/executor/executor_universal_linux.go
```diff
@@ -190,12 +190,23 @@ func (e *UniversalExecutor) enterCG1(statsCgroup, cpusetCgroup string) func() {
...
 	}
 
+	// if custom cgroups are set join those instead of configuring the /nomad
+	// cgroups we are not going to use
+	if len(e.command.OverrideCgroupV1) > 0 {
+		pid := unix.Getpid()
+		for controller, path := range e.command.OverrideCgroupV1 {
+			absPath := cgroupslib.CustomPathCG1(controller, path)
+			ed := cgroupslib.OpenPath(absPath)
+			_ = ed.Write("cgroup.procs", strconv.Itoa(pid))
```
schmichael, line 205: Don't we need to check the return value here or risk a task running without being added to its intended cgroup?

schmichael (commented):
Just to confirm: this doesn't appear to create cgroups if they don't exist or remove cgroups on text exit, correct?

drivers/shared/executor/executor.go
```diff
@@ -331,8 +368,12 @@ func (e *UniversalExecutor) Launch(command *ExecCommand) (*ProcessState, error)
 
 	// setup containment (i.e. cgroups on linux)
 	if cleanup, err := e.configureResourceContainer(command, os.Getpid()); err != nil {
-		// keep going; some folks run nomad as non-root and expect this driver to still work
-		e.logger.Warn("failed to configure container, process isolation will not work", "error", err)
+		e.logger.Error("failed to configure container, process isolation will not work", "error", err)
+		if os.Geteuid() == 0 {
+			return nil, fmt.Errorf("unable to configure cgroups: %w", err)
```
schmichael, line 373: Let's also return an error if an override is specified. If someone uses an override we need to fail fast if it's not possible.

drivers/shared/executor/executor_universal_linux.go
```diff
@@ -222,10 +239,11 @@ func (e *UniversalExecutor) configureCG1(cgroup string, command *ExecCommand) {
 		ed = cgroupslib.OpenPath(cpusetPath)
 		_ = ed.Write("cpuset.cpus", cpuSet)
```
schmichael, line 240: Is it safe to ignore this error?

shoenig: Historically we had to ignore these errors because people keep running Nomad as non-root or on systems where only some of the cgroup controllers are enabled. But they still want raw_exec to Just Work for them.

schmichael (approved):

PR #18122, opened 2023-08-01
[deps] Update HCL2 to 2.17.0+tweaks

angrycub:
In https://github.com/hashicorp/hcl/pull/620, I created an updated version of the Nomad-tweaked HCL2 code for HCL v2.17.0.  We leverage this version's `Extra` attribute of an `hcl.Diagnostic` in nomad-pack.

PR #20483, opened 2024-04-25
[ui] Sentinel Policies CRUD UI

philrenaud:
Provides access to Create, Edit, and Delete Sentinel Policies via the web UI.

By virtue of being within the `/administration/` route (née `/access-control/`), this requires a management token. Additionally, this requires sentinel policies in the features list provided by the agent, which is an enterprise feature.

Sentinel policies are enforced and shown at `plan` and `run` stage, and can be set as hard-mandatory, soft-mandatory, or advisory (warning but not preventing).

This PR also provides a Gallery of demo/jumping-off-point sentinel policies for things like "no friday deploys" or "must include canaries".

Side-effects:
- Renamed the "Access Control" section to "Administration" with an eye toward adding further sub-sections, like node pool CRUD.
- Administration card layout widened to accommodate a default of 5 cards (was originally built for 3)

TODO:
- Determine what we should do with the placeholder-commented-out-templates ([constraint-enforcement.js](https://github.com/hashicorp/nomad/pull/20483/files#diff-897b788676efc901549546b37d4b61ea04bc1b6c139849a512dc29df762cd5dfR6) etc.)

image
image
image


ui/app/utils/default-sentinel-policy-templates.js
```diff
@@ -0,0 +1,103 @@
...
+import alwaysFailPolicy from './sentinel_policy_templates/always-fail';
+// import enforceServiceChecksPolicy from './sentinel_policy_templates/enforce-service-checks';
+import canariesOnlyPolicy from './sentinel_policy_templates/canaries-only';
+// import enforceMeshPolicy from './sentinel_policy_templates/enforce-service-mesh';
+// import metadataPolicy from './sentinel_policy_templates/metadata-required';
+// import consulSDPolicy from './sentinel_policy_templates/consul-service-discovery-only';
+// import vaultSecretsPolicy from './sentinel_policy_templates/vault-secrets-only';
+// import dyanmicPortsPolicy from './sentinel_policy_templates/dynamic-ports-only';
+import resourceLimitsPolicy from './sentinel_policy_templates/resource-limits';
+// import constraintEnforcmentPolicy from './sentinel_policy_templates/constraint-enforcement';
```
philrenaud, line 19: A big TODO with this PR: do we launch this without the commented-out policies (which are currently just placeholders below), or can we write them in such a way that they make sense as starter templates (knowing Nomad ships with Sentinel v0.15.x?)

david-yu (approved):
Awesome, thank you. Does anyone know how to bump the sentinel version to the latest?

david-yu:
We should definitely bump up sentinel and sentinel-sdk, as our sentinel version 0.15.x is over 2 years old.

tgross:
> We should definitely bump up sentinel and sentinel-sdk, as our sentinel version 0.15.x is over 2 years old.

@david-yu I had a look at that this morning, and we can't update Sentinel without also updating our HCL2 library. But we're on a fork of HCL2 with languages changes we needed to prevent breaking backwards compatibility when we shipped HCL2 support in Nomad 1.0. We made an attempt to update in https://github.com/hashicorp/nomad/pull/18122 but that never got wrapped up and it's now behind 2.19.

Ideally we'd either (a) get our HCL2 changes upstreamed to main, but that's going to be a lift because it involves language changes, or (b) accept that we're going to break backcompat in the jobspec as we probably should have done when we originally switched to HCL2 for Nomad 1.0. In any case, unfortunately this is work we'll need to plan rather than just a quick dependency bump. :crying_cat_face:

ui/app/templates/components/administration-subnav.hbs
```diff
@@ -0,0 +1,17 @@
...
+~}}
+
+<div class="tabs is-subnav" {{did-insert this.keyboard.registerNav type="subnav"}} {{will-destroy this.keyboard.unregisterSubnav}}>
+  <ul>
+    <li><LinkTo @route="administration.index" @activeClass="is-active">Overview</LinkTo></li>
+    <li><LinkTo @route="administration.tokens" @activeClass="is-active">Tokens</LinkTo></li>
+    <li><LinkTo @route="administration.roles" @activeClass="is-active">Roles</LinkTo></li>
+    <li><LinkTo @route="administration.policies" @activeClass="is-active">Policies</LinkTo></li>
+    <li><LinkTo @route="administration.namespaces" @activeClass="is-active">Namespaces</LinkTo></li>
+    {{#if (can "list sentinel-policy")}}
```
tgross, line 13: This conditional is about whether the ENT feature is available, rather than about the ACL policy?

ui/app/templates/administration/sentinel-policies/gallery.hbs
```diff
@@ -0,0 +1,35 @@
...
+    </Hds::Form::RadioCard::Group>
+  </main>
+  <footer class="buttonset">
+    <Hds::ButtonSet class="button-group">
+      <Hds::Button @text="Apply" @route="administration.sentinel-policies.new" @query={{hash template=this.selectedTemplate}}
+        disabled={{is-equal this.selectedTemplate null}} data-test-apply />
+      <Hds::Button @text="Cancel" @route="administration.sentinel-policies.new" @color="secondary" data-test-cancel />
+    </Hds::ButtonSet>
+  </footer>
+</section>
```
tgross, line 35: Nitpicky: it probably doesn't matter for bundlers, but other Posix command line tools will miss this last line so it's a good idea to have your editor configured to ensure the EOL is always present. For example:

```
$ wc -l ui/app/templates/administration/sentinel-policies/gallery.hbs
34 ui/app/templates/administration/sentinel-policies/gallery.hbs
```

ui/app/components/sentinel-policy-editor.js
```diff
@@ -0,0 +1,86 @@
...
+    if (e instanceof Event) {
+      e.preventDefault(); // code-mirror "command+enter" submits the form, but doesnt have a preventDefault()
+    }
+    try {
+      const nameRegex = '^[a-zA-Z0-9-]{1,128}$';
+      if (!this.policy.name?.match(nameRegex)) {
+        throw new Error(
+          `Policy name must be 1-128 characters long and can only contain letters, numbers, and dashes.`
+        );
+      }
```
tgross, line 43: Do we want to enforce the description max length too? (256 characters)

tgross (approved):
LGTM!

I've left a few minor questions but nothing blocking.

ui/app/templates/components/administration-subnav.hbs
philrenaud: Yep! We let you load the Administration section as long as you can list roles, policies, tokens and namespaces (in other words: you are a management token) but it should still be a usable section without Sentinel Policies (ent).

So we don't show sentinel policies in subnav/index page, nor attempt to preload them from the parent route, if the Sentinel Policies feature (ent) is missing.

philrenaud: We also route-guard for other "is management" abilities in the [parent route](https://github.com/hashicorp/nomad/blob/8bfe03a835d108d4b2245b1da67a3eb5557cf8ce/ui/app/routes/administration.js#L20-L29)

david-yu:
@philrenaud Also wondering if Sentinel will show up as a submenu here? 
image

david-yu:
> > We should definitely bump up sentinel and sentinel-sdk, as our sentinel version 0.15.x is over 2 years old.
> 
> @david-yu I had a look at that this morning, and we can't update Sentinel without also updating our HCL2 library. But we're on a fork of HCL2 with languages changes we needed to prevent breaking backwards compatibility when we shipped HCL2 support in Nomad 1.0. We made an attempt to update in #18122 but that never got wrapped up and it's now behind 2.19.
> 
> Ideally we'd either (a) get our HCL2 changes upstreamed to main, but that's going to be a lift because it involves language changes, or (b) accept that we're going to break backcompat in the jobspec as we probably should have done when we originally switched to HCL2 for Nomad 1.0. In any case, unfortunately this is work we'll need to plan rather than just a quick dependency bump. 😿

Ok I'll see if we can announce deprecation of the HCL2 changes in 1.8.

philrenaud:
> Also wondering if Sentinel will show up as a submenu here?
@david-yu confirming that it does (when running ENT Nomad):
image

david-yu:
Submenu looks great, thank you!

angrycub:
This is looking so good.  I've been using it while developing out sample policies and noticed a couple of things:

* If I re-save the policy, could that auto-dismiss any previous result?  Sometimes I get result buildup in the corner and partially dismiss some before saving again which leaves an odd visible state.

* For some reason, I do not necessarily see the Sentinel editor UI the first pass into the Administration page and have to refresh it to get that card to appear the first time I enter the UI. (this has been intermittent)

* Since we have hard validation rules for sentinel policy names, can we prevalidate the value without having to round trip it?  Like with an invalid highlight on the box or something? (especially since the name is on the far side of the box from the submit.)

* There is the seems syntax highlight hiccup around multiline block comments. (This is probably due to the highlighter picking a language that looks _like_ Sentinel.

david-yu:
Would be nice if its possible to add a template to prevent job descriptions: https://github.com/hashicorp/nomad/pull/18292

philrenaud:
> Since we have hard validation rules for sentinel policy names, can we prevalidate the value without having to round trip it? Like with an invalid highlight on the box or something? (especially since the name is on the far side of the box from the submit.)

Noting that we throw this error without roundtripping:
image

I'll highlight the name box on this error, though

PR #20492, opened 2024-04-29
quota: ensure quota usage is freed when jobs are purged

tgross:
When a job is purged, we delete all its allocations and the client detects the absense of the allocations to clean up its resources locally. But the client won't be able to send an allocation status update in this case, which frees the quota being used by that allocation. Instead, we need to free the quota usage inside the state store immediately. To do so, we check if the allocation is already client-terminal before copying it and passing it into the Enterprise code for cleanup.

This commit also refactors the job delete to make it clear there's a single caller of this alloc deletion path. This refactoring eliminates some wasteful logic that queries the "allocs" table, allocates a slice of strings for their IDs, and then queries the "allocs" table one-by-one for each of them for deletion anyways.

Tests for this code can be found in the linked ENT repo PR. This PR is currently identical except for those tests; I'll rebase the ENT PR once this PR has been merged over to the ENT repo.

Fixes: https://github.com/hashicorp/nomad-enterprise/issues/1422
Ref: https://hashicorp.atlassian.net/browse/NOMAD-620
Ref: https://github.com/hashicorp/nomad-enterprise/pull/1432

### DIFF
--- a/.changelog/_1432.txt
+++ b/.changelog/_1432.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+quota (Enterprise): Fixed a bug where quota usage would not be freed if a job was purged
+```
--- a/nomad/state/state_store.go
+++ b/nomad/state/state_store.go
@@ -833,44 +833,29 @@ func (s *StateStore) DeleteDeploymentTxn(index uint64, deploymentIDs []string, t
 	return nil
 }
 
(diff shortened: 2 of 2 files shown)


gulducat (approved):
lgtm!

Juanadelacuesta (approved):
Makes sense :)

PR #20493, opened 2024-04-30
[ui] Handle parent/child jobs with the paginated Jobs Index route

### LINKED ISSUE #20491: Handle parent jobs' children status
Tackle 2 things related to the new [jobs index page re-work](https://github.com/hashicorp/nomad/pull/20452):
- Display the child job statuses of a parent job on the index page in visual alignment with how we show allocs of a non-parent job
- Fix an issue where child jobs aren't loaded at `/jobs/$parentID/` in the Child Jobs table (they were previously assumed in-cache by virtue of being loaded on the index page, but now that's paginated so it's no guarantee)

philrenaud:
Brings child jobs to the /statuses endpoint when a flag is present, and solves the issue of "Parent jobs presume knowledge of the entirety of a cluster's jobs" + "the index page is now paginated" from this project at large.

This also makes some quality of life updates to the existing child jobs table on a job page, and makes monitoring parent jobs from the main/index page more useful (with status badges that react to live updates and tallies of child jobs)

image
image

Resolves #20491


ui/app/components/child-job-row.hbs
```diff
@@ -0,0 +1,49 @@
+{{!
```
philrenaud, line 1: Most of this is a copy/paste of the existing <JobRow> component, but because we have sufficient deviation in what we show here, it made more sense for it to be its own component rather than deal with conditionals.

ui/app/components/job-page/parts/children.js
```diff
@@ -31,9 +31,9 @@ export default class Children extends Component.extend(Sortable) {
     return this.get('job.taskGroups') || [];
   }
 
-  @computed('job.children.[]')
+  @computed('jobs.[]')
   get children() {
-    return this.get('job.children') || [];
+    return this.jobs || [];
```
philrenaud, line 36: A big part of this PR is going from "Derive child jobs from the parent job's model's computed properties", to "accept the child jobs as passed in directly"

ui/app/components/job-status/allocation-status-row.hbs
```diff
@@ -55,7 +55,11 @@
     </div>
   {{/if}}
   {{#if @compact}}
-    {{@runningAllocs}}/{{@groupCountSum}}
+    {{#if (eq @allocationTallyMode "complete")}}
+      {{@completeAllocs}}/{{@groupCountSum}}
+    {{else}}
+      {{@runningAllocs}}/{{@groupCountSum}}
+    {{/if}}
```
philrenaud, line 62: Felt it made more sense, in the context of a list of parameterized/periodic child jobs, which are going to be batch/sysbatch, to tally up the completed allocations rather than the running ones.

This is inconsistent with what we show on the jobs index page (running/expected), but contextually probably more relevant. I'm only feeling about 75% confident of this change so open to undoing it.

ui/app/controllers/jobs/job/index.js
```diff
@@ -62,4 +66,62 @@ export default class IndexController extends Controller.extend(
...
+          );
+        }
+        this.childJobs = childJobs;
+        yield timeout(throttle);
+      }
+      if (Ember.testing) {
+        break;
+      }
+    }
+  }
```
philrenaud, line 126: Generally follows the main "duelling queries" pattern from the jobs index list of having the route's `setupController` kick off a restartable ember concurrency task, rather than use `@watchQuery`, which has some "when do I invalidate the cache" issues.

ui/app/routes/jobs/job/index.js
```diff
@@ -54,24 +49,42 @@ export default class IndexRoute extends Route.extend(WithWatchers) {
         sortProperty: 'submitTime',
         sortDescending: true,
       });
+
+      controller.resetQueryIndex({
+        id: model.get('plainId'),
+        namespace: model.get('namespace.id'),
+      });
```
philrenaud, line 56: Feels extraneous, but otherwise loading a parent job, then back out, then another parent job, then back out, then the first one again results in the child jobs from the second job being present due to the blocking query run at page load still having the previous-version's index.

There is probably another way to do this, but this appears to be working as expected.

gulducat (approved):
LGTM!

PR #20510, opened 2024-05-02
core: do not create evaluations within batch deregister endpoint.

jrasell:
The batch deregister RPC endpoint is only used by the internal garbage collection process, it is not exposed via the HTTP API or used anywhere else.

The GC process ensures that a job can only be removed from state if all related evaluations and allocations are in a state that means they can also be removed from state. This means that we do not need to create evaluations when jobs are being deregistered via this endpoint.

Seeing that the endpoint is internal only, this change also removes the ACL capability check per namespace and moves it to a simpler and more efficient management check.

Notes:

- I took this opportunity to move some structs into their related file, ever so slightly making structs.go smaller.
- The added tests build on existing tests, and help understand code for future readers as well as acted as learning for myself
- I deprecated params as we cannot remove them due to the struct being persisted via Raft


nomad/job_endpoint.go
```diff
@@ -939,6 +942,14 @@ func (j *Job) BatchDeregister(args *structs.JobBatchDeregisterRequest, reply *st
 		return err
 	}
 
+	// This RPC endpoint is only called from the Nomad server leader when
+	// performing garbage collection, therefore we can get away with a simple
+	// management check to ensure the caller can trigger this functionality as
+	// the leader token is always a management token.
+	if !aclObj.IsManagement() {
+		return structs.ErrPermissionDenied
+	}
```
tgross, line 951: The comment here isn't _quite_ right, as the GC eval can be processed on a follower using the leader's token. But because we never expect a human operator to use this RPC (it's not exposed in the HTTP API), we can narrow this further to a server op check, similar to the plan applier, etc.:

```suggestion
	// This RPC endpoint is only called during garbage collection
	if !aclObj.AllowServerOp() {
		return structs.ErrPermissionDenied
	}
```

nomad/structs/job.go
```diff
@@ -16,6 +24,53 @@ const (
...
+}
+
+// JobBatchDeregisterResponse is used to respond to a batch job deregistration.
+type JobBatchDeregisterResponse struct {
+
+	// JobEvals maps the job to its created evaluation.
+	//
+	// Deprecated: The job batch deregister endpoint is only used by internal
+	// garbage collection which no longer creates and evaluation per job GC.
+	JobEvals map[NamespacedID]string
```
tgross, line 69: I think we can safely remove this? During upgrades, a leader with the new code will send back the new struct missing the `JobEvals` field, which the caller will interpret as an empty list.

But if we're feeling paranoid about it, we could always leave a note that it can be removed in the next LTS release.

nomad/structs/job.go
```diff
@@ -16,6 +24,53 @@ const (
...
+type JobBatchDeregisterRequest struct {
+
+	// Jobs is the set of jobs to deregister.
+	Jobs map[NamespacedID]*JobDeregisterOptions
+
+	// Evals is the set of evaluations to create.
+	//
+	// Deprecated: the job batch deregister endpoint no longer generates an
+	// eval per job.
+	Evals []*Evaluation
```
tgross, line 38: I think we can safely remove this. If we send the new request to an older server and it's missing the field it'll be treated as empty on the other side. And if we send the older request to a new server, I'm pretty sure that new field gets ignored? Normally that'd be unsafe but we just end up dropping the behavior of writing new Evaluations, so that should be ok.

Worth testing across an upgrade though.

tgross (approved):

nomad/job_endpoint.go
jrasell: I did consider changing the ACL check to the server only method, but I had one reservation about how this would work during an upgrade.

In a 3 server situation:
- one follower is upgraded to the new code
- the upgraded follower picks up the core eval to run the GC 
- the upgraded follower submits the GC write to the leader without adding an ACL token to the request
- the leader running the old code attempts to perform the namespace capability check, which fails

The ACL token addition to the RPC request could be kept which I believe would avoid the situation described above, but I am skeptical of adding this due to the following comment within the `AuthenticateServerOnly` function:

> This should never be used for RPCs that serve HTTP
// endpoints or accept ACL tokens to avoid confused deputy attacks by making a
// request to a follower that's forwarded.

Keen to know you thoughts here and whether the comment would impact upgrades where a non-upgrade follower submits a request to an upgraded leader?

tgross: :facepalm: You're absolutely right. I was thinking that we could use the `AllowServerOp` with the `Authenticate` method that's already here rather than `AuthenticateServerOnly`, but that's not the intent at all. Nevermind!

nomad/structs/job.go
jrasell: Updated within: https://github.com/hashicorp/nomad/pull/20510/commits/039bf6aca7f8f83e5f783aa7760f33446e7cc1cd

Tested upgrades in AWS without issue.

nomad/structs/job.go
jrasell: Updated within: https://github.com/hashicorp/nomad/pull/20510/commits/039bf6aca7f8f83e5f783aa7760f33446e7cc1cd

Tested upgrades in AWS without issue.

tgross (approved):
LGTM!

schmichael (commented):
Love it! Sorry for the late review. I don't see any upgrade/backward compat concerns, but would it be worth manually forcing a system gc with a 1.8 leader and 1.7 followers or similar?

PR #20511, opened 2024-05-02
test: regenerate test TLS certificates.

jrasell:
Fixes broken tests due to TLS certificate expiry.


tgross (approved):
LGTM!

This gets all the tests passing but I see there's a few other certs in this directory too. Should we regen these as well? Are they even still in use?

Juanadelacuesta (approved):
Thank you!! :)

jrasell:
> LGTM!
> 
> This gets all the tests passing but I see there's a few other certs in this directory too. Should we regen these as well? Are they even still in use?

The ones that have not been regenerated are the CA certificates, which are valid for another 4 years I believe.

PR #20522, opened 2024-05-03
structs: Fix job canonicalization for array type fields

### LINKED ISSUE #20497: Job versions with empty diffs being created when updating a job before it becomes stable in some circumstances
### Nomad version
v1.6.3, tip also seems affected

Linux

### Issue
When submitting a new version of a job using a wrapper library, I observed that if the job gets resubmitted with no changes before it gets stabilized, a new version is created with an empty diff. This terminates the old allocations created for that version and creates new ones.

If the job is resubmitted repeatedly and the time between resubmissions is shorter than the time to stabilise the job, this will result in an endless loop of creating new versions.

### Issue details
It seems that this only happens if certain fields containing arrays like `Constraints`, `Affinities` or `Spreads` contain an empty list in `Job`, `TaskGroup` or `Task`.

#### Background - Job submission process
Most definitions of functions mentioned here are in `nomad/structs/structs.go`.

When a job is submitted, it seems to go through the following process (these are just the relevant bits for this bug):
1. `Job.Canonicalize` is run on the submitted job
2. In `Job.SpecChanged`, the submitted job is compared against the previous version
  a) The submitted job is copied. The copy function uses `CopySlice...` functions to copy arrays like `Constraints`, `Affinities` or `Spreads`.
  b) Certain ephemeral fields are copied from the previous version of the job into the newly created copy
  c) The newly created copy is compared against the previous version of the job using `reflect.DeepEqual`
3. If there is a difference a new version is created, with the field `Stable` set to false
4. Eventually, the job will be marked as stable in `nomad/state/state_store.go` in `StateStore.updateJobStabilityImpl` which copies the job as part of the process.

#### Bug
It seems that in `Job.Canonicalize` nothing is done to fields like `Constraints`, `Affinities` or `Spreads`. This means that the initial submission of a new version the job will have the same contents of those fields as they are submitted.

However, when the comparison is done to check if the spec changed, the previous version is compared against a copy of the new submitted version. While copying, the `CopySlice...` functions seem to canonicalize empty arrays as nil. This leads to comparing an empty array and nil which are not equal according to `reflect.DeepEqual` if the submitted job had empty arrays in those fields.

This starts working again after the job has been marked as stable because the job is copied as part of that process and future comparisons check two jobs which both have been copied at some point.

### Proposed fix

`Job.SpecChanged` should return true for jobs that are conceptually the same.

There are many possible ways to achieve that goal, I prepared a PR which canonicalizes submitted jobs so that the fields mentioned above are converted to null if they contain just an empty list, similar to how the `Meta` field is canonicalized and the same way these fields were implicitly canonicalized when copying.

### Reproduction steps
1. Submit a job containing an empty array in `Constraints`, `Affinities` or `Spreads` (example below).
6. Check the version of that job (will be 0)
7. Submit a modified version of the job (for example by changing the task name)
8. Check the version of that job (will be 1)
9. Resubmit the last version of the job before it is marked as stable

#### Expected Result
The version of the job is still 1

#### Actual Result
The version of the job is 2 and the diff between job version 1 and 2 in the nomad web ui appears to be empty.

### Job as JSON
```
{
  "ID": "test",
  "Namespace": "default",
  "Datacenters": [
    "test-dc"
  ],
  "Region": "test-region",
  "Constraints": [],
  "Type": "batch",
  "TaskGroups": [
    {
      "Name": "task-group",
      "Count": 1,
      "Tasks": [
        {
          "Name": "test",
          "Driver": "raw_exec",
          "Env": {},
          "Resources": {
            "CPU": 100,
            "MemoryMB": 100,
            "Devices": []
          },
          "Constraints": [],
          "Config": {
            "command": "sleep",
            "args": [
              "inf"
            ]
          },
          "RestartPolicy": {
            "Attempts": 0,
            "Delay": 5000000000,
            "Interval": 5000000000,
            "Mode": "fail"
          },
          "Leader": true,
          "Lifecycle": null,
          "Affinities": []
        }
      ],
      "Constraints": [],
      "ReschedulePolicy": {
        "Attempts": 0,
        "Interval": 86400000000000,
        "Unlimited": false,
        "Delay": 5000000000,
        "DelayFunction": "constant",
        "MaxDelay": null
      },
      "Affinities": []
    }
  ],
  "Meta": {},
  "Affinities": []
}
```

myszon:
Fixes #20497

### DIFF
--- a/.changelog/20522.txt
+++ b/.changelog/20522.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+structs: Fix job canonicalization for array type fields
+```
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -4510,10 +4510,22 @@ func (j *Job) Canonicalize() {
 		return
 	}
 
-	// Ensure that an empty and nil map are treated the same to avoid scheduling
+	// Ensure that an empty and nil map or array are treated the same to avoid scheduling
 	// problems since we use reflect DeepEquals.
 	if len(j.Meta) == 0 {
 		j.Meta = nil
+	}
+
+	if len(j.Constraints) == 0 {
+		j.Constraints = nil
+	}
+
+	if len(j.Affinities) == 0 {
+		j.Affinities = nil
+	}
+
+	if len(j.Spreads) == 0 {
+		j.Spreads = nil
 	}
 
 	// Ensure the job is in a namespace.
@@ -6730,10 +6742,22 @@ func (tg *TaskGroup) Copy() *TaskGroup {
 
 // Canonicalize is used to canonicalize fields in the TaskGroup.
 func (tg *TaskGroup) Canonicalize(job *Job) {
-	// Ensure that an empty and nil map are treated the same to avoid scheduling
+	// Ensure that an empty and nil map or array are treated the same to avoid scheduling
 	// problems since we use reflect DeepEquals.
 	if len(tg.Meta) == 0 {
 		tg.Meta = nil
+	}
+
+	if len(tg.Constraints) == 0 {
+		tg.Constraints = nil
+	}
+
+	if len(tg.Affinities) == 0 {
+		tg.Affinities = nil
+	}
+
+	if len(tg.Spreads) == 0 {
+		tg.Spreads = nil
 	}
 
 	// Set the default restart policy.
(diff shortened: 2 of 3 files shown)


pkazmierczak (commented):
Hey @myszon, thanks for the contribution! I think it makes sense. Like you mention in https://github.com/hashicorp/nomad/issues/20497 it's a bit of an edge case, but a bug nonetheless and we appreciate the fix. 

I only had some minor wording comments, feel free to accept or reject them. Could you add a changelog entry, though? A textfile inside the `.changelog` directory with the PR number as the filename and a short description of the fix, have a look [at an example](https://github.com/hashicorp/nomad/blob/main/.changelog/15134.txt). Body of the changelog can be something like:
```
structs: Fixed job canonicalization for array type fields
```
and you can also use the `make cl` command to generate the file. Thanks again!

myszon:
Thanks for the comments, accepted suggestions and added the changelog entry.

pkazmierczak (approved):

Issue #3885, opened 2018-02-20
Nomad dies if HUP is sent during agent initialization


burdandrei:
### Nomad version
Nomad v0.7.1 (0b295d399d00199cfab4621566babd25987ba06e)
### Operating system and Environment details
Distributor ID: Ubuntu
Description:  Ubuntu 16.04.2 LTS
Release:  16.04

### Issue
We're provisioning nomad with chef recipe that send reload on config update and restart on binary update:
```ruby
hashistack_bring_binary 'nomad' do
  url node['nomad']['download_url']
  unzip true
  notifies :restart, 'systemd_unit[nomad.service]', :delayed
end

template '/etc/nomad.d/agent.hcl' do
  source 'nomad.hcl.erb'
  owner 'nomad'
  group 'root'
  mode '0644'
  sensitive true
  notifies :reload_or_try_restart, 'systemd_unit[nomad.service]', :delayed
end

systemd_unit 'nomad.service' do
  content <<-EOU.gsub(/^\s+/, '')
[Unit]
Description=Nomad
Documentation=https://nomadproject.io/docs/
Requires=network-online.target
After=network-online.target

[Service]
LimitNOFILE=65536
Restart=on-failure
ExecStart=/usr/local/bin/nomad agent -config /etc/nomad.d
ExecReload=/bin/kill -HUP $MAINPID
RestartSec=5s

[Install]
WantedBy=multi-user.target
  EOU
  action [ :create, :enable ]
  notifies :restart, 'systemd_unit[nomad.service]', :delayed
end

```

Chef triggers restart and immediately after this reload.
On new machine if nomad data directory  is not initialized and nomad receives HUP it just exits with 0:
```
# systemctl status nomad
● nomad.service - Nomad
   Loaded: loaded (/etc/systemd/system/nomad.service; enabled; vendor preset: enabled)
   Active: inactive (dead) since Mon 2018-02-19 09:08:02 UTC; 38s ago
     Docs: https://nomadproject.io/docs/
 Main PID: 23627 (code=killed, signal=HUP)

Feb 19 09:08:02 ip-x-x-x-x systemd[1]: Started Nomad.
Feb 19 09:08:02 ip-x-x-x-x nomad[23627]:     Loaded configuration from /etc/nomad.d/agent.hcl
Feb 19 09:08:02 ip-x-x-x-x nomad[23627]: ==> Starting Nomad agent...
[...]

PR #20528, opened 2024-05-03
agent: add support for sdnotify protocol

tgross:
Nomad agents expect to receive `SIGHUP` to reload their configuration. The signal handler for this is installed fairly late in agent startup, after the client or server components are up and running. This means that configuration management tools can potentially reload the configuration before the agent can handle it, causing the agent to crash.

We don't want to allow configuration reload during client or server component startup, because it would significantly complicate initialization. Instead, we'll implement the systemd notify protocol. This causes systemd to block sending configuration reload signals until the agent is actually ready. Users can still bypass this by sending signals directly.

Note that there are several Go libraries that implement the sdnotify protocol, but most are part of much larger projects which would create a lot of dependabot burden. The bits of the protocol we need are extremely simple to implement in a just a couple of functions.

For non-Linux or non-systemd Linux systems, this feature is a no-op. In future work we could potentially implement service notification for Windows as well.

Fixes: https://github.com/hashicorp/nomad/issues/3885

---

Note on compatibility testing. Successful reloads with `Type=notify` look like this:

> Reloading Nomad Agent...
> ==> Caught signal: hangup
> ==> Reloading configuration...
>     2024-05-03T11:50:38.505-0400 [INFO]  client.fingerprint_mgr: reloading fingerprinter: fingerprinter=cni
> Reloaded Nomad Agent.

If the unit file is changed from `Type=notify` to `Type=simple` (default) but the agent is still running, we end up with an error in the logs like the following, but everything still works as expected:

> nomad.service: Got notification message from PID 2965, but reception is disabled.

### DIFF
--- a/.release/linux/package/usr/lib/systemd/system/nomad.service
+++ b/.release/linux/package/usr/lib/systemd/system/nomad.service
@@ -26,6 +26,7 @@ After=network-online.target
 # StartLimitInterval = 10s
 
 [Service]
+Type=notify
 EnvironmentFile=-/etc/nomad.d/nomad.env
 ExecReload=/bin/kill -HUP $MAINPID
 ExecStart=/usr/bin/nomad agent -config /etc/nomad.d
--- a/.changelog/20528.txt
+++ b/.changelog/20528.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+agent: Added support for systemd readiness notifications
+```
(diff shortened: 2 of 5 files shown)


.release/linux/package/usr/lib/systemd/system/nomad.service
```diff
@@ -26,6 +26,7 @@ After=network-online.target
 # StartLimitInterval = 10s
 
 [Service]
+Type=notify
```
tgross, line 29: Note to reviewers: ideally we'd have this as `Type=notify-reload` because that's a little nicer, but that's not available until systemd 253, which only the very most recent LTS distros have.

schmichael (approved):

PR #20529, opened 2024-05-03
[ui] Remove json linting while editing variables

philrenaud:
This removes the JSON Linting gutter highlight capabilities originally added in https://github.com/hashicorp/nomad/pull/13461

The main reason for this removal is the inclusion of a nested Underscore dependency: https://github.com/hashicorp/nomad/security/dependabot/110

However, our dependency (a grandparent of the insecure underscore version) has not been updated in quite some time and appears to be unmaintained.

As such, I've given consideration to whether this feature is worth forking, pinning, or abandoning. Per https://github.com/zaach/jsonlint/pull/120 a change has been merged to `master` in `jsonlint`, but no release on `npm` has been made. As such, I've opted to remove the feature for the same reasons [other projects](https://github.com/mapbox/mapbox-gl-draw/issues/1049) have done so: it doesn't give enough benefit to warrant hanging on to an un-updated dependency for marginal utility.

### DIFF
--- a/ui/app/components/variable-form.js
+++ b/ui/app/components/variable-form.js
@@ -394,14 +394,8 @@ export default class VariableFormComponent extends Component {
    *
    * @param {string} value
    */
-  @action updateCode(value, codemirror) {
-    codemirror.performLint();
+  @action updateCode(value) {
     try {
-      const hasLintErrors = codemirror?.state.lint.marked?.length > 0;
-      if (hasLintErrors || !JSON.parse(value)) {
-        throw new Error('Invalid JSON');
-      }
-
       // "myString" is valid JSON, but it's not a valid Variable.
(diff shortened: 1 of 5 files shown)


tgross (approved):
LGTM! (once tests are green) This was a nice to have but if someone is really having trouble with it, the form-editor interface is still there instead of the JSON mode. And it's all ultimately validated server side.

philrenaud:
> LGTM! (once tests are green) This was a nice to have but if someone is really having trouble with it, the form-editor interface is still there instead of the JSON mode. And it's all ultimately validated server side.

Even better, it'll still run a `JSON.parse()` on user input on keypress, so the user will still get output like this before ever hitting the server:

image

dduzgun-security (approved):

PR #20532, opened 2024-05-03
CSI: include volume namespace in staging path

tgross:
CSI volumes are namespaced. But the client does not include the namespace in the staging mount path. This causes CSI volumes with the same volume ID but different namespace to collide if they happen to be placed on the same host. The per-allocation paths don't need to be namespaced, because an allocation can only mount volumes from its job's own namespace.

Fixes: https://github.com/hashicorp/nomad/issues/18741
See https://github.com/hashicorp/nomad/issues/18741#issuecomment-2102752320 for my notes on upgrade testing.

### DIFF
--- a/testutil/mock_calls.go
+++ b/testutil/mock_calls.go
@@ -33,6 +33,12 @@ func (c *CallCounter) Get() map[string]int {
 	return maps.Clone(c.counts)
 }
 
+func (c *CallCounter) Reset() {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.counts = make(map[string]int)
+}
+
 func (c *CallCounter) AssertCalled(t testing.T, name string) {
 	t.Helper()
 	counts := c.Get()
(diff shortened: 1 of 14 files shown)


testutil/mock_calls.go
```diff
@@ -33,6 +33,12 @@ func (c *CallCounter) Get() map[string]int {
 	return maps.Clone(c.counts)
 }
 
+func (c *CallCounter) Reset() {
```
Juanadelacuesta, line 36: Im failing to see where do we use this one

Juanadelacuesta (approved):
Looks pretty straight forward, nice job

testutil/mock_calls.go
tgross: [Here](https://github.com/hashicorp/nomad/pull/20532/files#diff-e90d51a406b0aa090a361afc3207f09c62da9d012d7e397e195e9f9e34958acfR361) so that we can use the `MountVolume` code path to setup an expected environment and then reset the calls so that the test assertion only lists the unmount/unpublish calls we expect.

PR #20535, opened 2024-05-06
[ui] Splits Settings into 2 sections, sign-in/profile and user settings

philrenaud:
This all stemmed from wanting a dedicated place to set "Do/Don't jostle me on the jobs index page". This PR adds a user settings page under "settings", which was previously used just to sign in / view profile.

As such, a global-nav item is always shown (regardless of whether ACLs are disabled) and is either:
- an icon with a link to the sign-in page (if ACLs are enabled and you're not signed in)
- an icon with a link to the user settings page (if ACLs are disabled)
- a dropdown that links to the sign-in page (if ACLs are enabled and you're signed in)

These settings are localStorage-specific (and noted as such on the user settings page)

I imagine we'll be adding more settings from around the app here soon

image
image

### DIFF
--- a/ui/app/components/global-header.js
+++ b/ui/app/components/global-header.js
@@ -18,17 +18,6 @@ export default class GlobalHeader extends Component {
   'data-test-global-header' = true;
   onHamburgerClick() {}
 
-  // Show sign-in if:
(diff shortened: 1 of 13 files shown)


gulducat (approved):
can't wait to see what else folks want on this settings page!

PR #20543, opened 2024-05-08
client: fix unallocated CPU metric when reserved cpu is set.

### LINKED ISSUE #20521: Reserved CPU is counted twice
### Nomad version
Nomad v1.7.5

Ubuntu 20.04.4

### Issue

When using the `reserved` block in the nomad client config to set CPU reservation, the reservation is counted twice. We've observed this by watching the metric `nomad_client_unallocated_cpu`, and by jobs no longer being schedulable since upgrading to Nomad 1.7.

We think the issue was introduced in the this PR: https://github.com/hashicorp/nomad/pull/18681/files#diff-bd3a55a72186f59e2e63efb4951573b2f9e4a7cc98086e922b0859f8ccc1dd09R3238

From what we can tell the reservation is subtracted both on the line linked above and in the `UsableCompute` function.

### Reproduction steps

Add a `reserved` block to the nomad client config and observe `nomad_client_unallocated_cpu`.

```
 reserved {
        cpu = 1000
    }
``` 

Update the reserved block to this:
```
 reserved {
        cpu = 2000
    }
``` 

Restart nomad

#### Expected Result
The metric `nomad_client_unallocated_cpu` should decrease by 1000

#### Actual Result
The metric `nomad_client_unallocated_cpu` decreases by 2000

jrasell:
closes #20521

### DIFF
--- a/.changelog/20543.txt
+++ b/.changelog/20543.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+client: Fix unallocated CPU metric calculation when client reserved CPU is set
+```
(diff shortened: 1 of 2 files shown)


ostkrok (approved):

tgross (approved):
LGTM!

Issue #20424, opened 2024-04-17
Race condition when a CSI volume with staging gets unmounted from one alloc and mounted to another alloc


ygg-drop:
### Nomad version
Output from `nomad version`

```
Nomad v1.7.6+ent
BuildDate 2024-03-12T07:55:51Z
Revision a71f7e82ef60200f273a92601b1b366429b0bd5f
```

### Operating system and Environment details

Alpine 3.19.1

### Issue

Let's assume there are two allocations - A and B - and both of them use the same CSI volume with staging and `multi-node-multi-writer` access mode (CephFS in this case).

The following events happen:

1. alloc A has the volume mounted
2. alloc A is stopped and Nomad calls `NodeUnpublishVolume` RPC to the CSI driver followed by `NodeUnstageVolume`
3. at the same time alloc B is started and Nomad calls `NodePublishVolume` on the same volume, **missing the fact that the same volume was unstaged with alloc A**.
4. in the CSI driver, the `NodePublishVolume` RPC finishes first and the volume gets bind-mounted from `staging` path to `per-alloc` path
5. in the CSI driver, the `NodeUnstageVolume` RPC finishes second, and the volume gets unmounted from `staging` path, which results in alloc B's `per-alloc` bind mount **now pointing to the underlying local device instead of the network volume**.
6. subsequent calls to  `NodeUnpublishVolume` RPC for alloc B fail with `device or resource busy` error.

Below is a sanitized `cephcsi` node plugin's log fragment capturing just the relevant events. In this case both allocations were failing due to an unrelated error which resulted in multiple rescheduling of the jobs, which probably increased the likelihood of this bug to occur.

- `VOLUME_A` - the volume used by both allocs
- `b2bd2806-f675-41f7-8efc-8a27830ce3d1` - alloc A
- `5a77c319-f7ab-4fd5-98cb-1b73538f735f` - alloc B

```
I0101 01:04:27.675127       1 utils.go:195] ID: 11620 Req-ID: VOLUME_A GRPC call: /csi.v1.Node/NodeUnpublishVolume
I0101 01:04:27.675173       1 utils.go:206] ID: 11620 Req-ID: VOLUME_A GRPC request: {"target_path":"/local/csi/per-alloc/b2bd2806-f675-41f7-8efc-8a27830ce3d1/VOLUME_A/rw-file-system-multi-node-multi-writer","volume_id":"VOLUME_A"}
[...]

PR #20550, opened 2024-05-09
CSI: track node claim before staging to prevent interleaved unstage

tgross:
The CSI hook for each allocation that claims a volume runs concurrently. If a call to `MountVolume` happens at the same time as a call to `UnmountVolume` for the same volume, it's possible for the second alloc to detect the volume has already been staged, then for the original alloc to unpublish and unstage it, only for the second alloc to then attempt to publish a volume that's been unstaged.

The usage tracker on the volume manager was intended to prevent this behavior but the call to claim the volume was made only after staging and publishing was complete. Move the call to claim the volume for the usage tracker to the top of the `MountVolume` workflow to prevent it from being unstaged until all consuming allocations have called `UnmountVolume`.

Fixes: https://github.com/hashicorp/nomad/issues/20424

---

~Note this PR currently contains everything in https://github.com/hashicorp/nomad/pull/20532 because they'd otherwise conflict. Once that's been merged I'll rebase on main and take this out of draft. The actual changes are in https://github.com/hashicorp/nomad/commit/c01986bd871d2396571e00dec75d4787fae0bb4d~ Done

### DIFF
--- a/.changelog/20550.txt
+++ b/.changelog/20550.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+csi: Fixed a bug where concurrent mount and unmount operations could unstage volumes needed by another allocation
+```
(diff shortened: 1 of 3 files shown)


pkazmierczak (approved):
LGTM!

PR #20553, opened 2024-05-09
bug: resolve type conversion alerts

dduzgun-security:
Resolves `Incorrect conversion between integer types` alerts to catch potential overflows.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -1184,7 +1184,7 @@ func parseQueryMeta(resp *http.Response, q *QueryMeta) error {
 	if err != nil {
 		return fmt.Errorf("Failed to parse X-Nomad-LastContact: %v", err)
 	}
-	if last > uint64(math.MaxInt64) {
+	if last > math.MaxInt64 {
 		return fmt.Errorf("Last contact duration is out of range: %d", last)
 	}
 	q.LastContact = time.Duration(last) * time.Millisecond
--- a/api/api.go
+++ b/api/api.go
@@ -1184,7 +1184,7 @@ func parseQueryMeta(resp *http.Response, q *QueryMeta) error {
 	if err != nil {
 		return fmt.Errorf("Failed to parse X-Nomad-LastContact: %v", err)
 	}
-	if last == 0 || last > uint64(math.MaxInt64) {
+	if last > uint64(math.MaxInt64) {
 		return fmt.Errorf("Last contact duration is out of range: %d", last)
 	}
 	q.LastContact = time.Duration(last) * time.Millisecond
--- a/api/api.go
+++ b/api/api.go
@@ -1184,6 +1184,9 @@ func parseQueryMeta(resp *http.Response, q *QueryMeta) error {
 	if err != nil {
 		return fmt.Errorf("Failed to parse X-Nomad-LastContact: %v", err)
 	}
+	if last == 0 || last > uint64(math.MaxInt64) {
+		return fmt.Errorf("Last contact duration is out of range: %d", last)
+	}
 	q.LastContact = time.Duration(last) * time.Millisecond
 	q.NextToken = header.Get("X-Nomad-NextToken")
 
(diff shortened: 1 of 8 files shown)


command/agent/http.go
```diff
@@ -903,21 +903,23 @@ func parseWait(resp http.ResponseWriter, req *http.Request, b *structs.QueryOpti
...
+			return nil
 		}
 		staleQuery, err := strconv.ParseBool(staleVal[0])
 		if err != nil {
+			errMsg := "Expect `true` or `false` for `stale` query string parameter"
 			resp.WriteHeader(http.StatusBadRequest)
-			_, _ = resp.Write([]byte(fmt.Sprintf("Expect `true` or `false` for `stale` query string parameter, got %s", staleVal[0])))
-			return
+			resp.Write([]byte(errMsg))
+			return CodedError(http.StatusBadRequest, errMsg)
```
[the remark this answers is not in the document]
dduzgun-security: Good catch! Silly mistake in the condition 🤦‍♂️ 
This commit should fix it https://github.com/hashicorp/nomad/pull/20553/commits/63923e2f3bd11b33859a2637389ecc9281f7f8ae


```bash
curl -v "http://localhost:4646/v1/jobs?stale=random"
*   Trying [::1]:4646...
* Connected to localhost (::1) port 4646
> GET /v1/jobs?stale=random HTTP/1.1
> Host: localhost:4646
> User-Agent: curl/8.4.0
> Accept: */*
> 
< HTTP/1.1 400 Bad Request
< Vary: Accept-Encoding
< Date: Fri, 10 May 2024 19:48:40 GMT
< Content-Length: 59
< Content-Type: text/plain; charset=utf-8
< 
* Connection #0 to host localhost left intact
Expect `true` or `false` for `stale` query string parameter% 
```

```bash
curl -v "http://localhost:4646/v1/jobs?per_page=random"
*   Trying [::1]:4646...
* Connected to localhost (::1) port 4646
> GET /v1/jobs?per_page=random HTTP/1.1
> Host: localhost:4646
> User-Agent: curl/8.4.0
> Accept: */*
> 
< HTTP/1.1 400 Bad Request
< Vary: Accept-Encoding
< Date: Fri, 10 May 2024 19:49:07 GMT
< Content-Length: 53
< Content-Type: text/plain; charset=utf-8
< 
* Connection #0 to host localhost left intact
Expect a number for `per_page` query string parameter% 
```

tgross (dismissed):
LGTM! Once CI is green we can get this merged and backported.

client/lib/numalib/detect_linux.go
```diff
@@ -89,7 +89,10 @@ func (*Sysfs) discoverCosts(st *Topology, readerFunc pathReaderFn) {
 		}
 
 		for i, c := range strings.Fields(s) {
-			cost, _ := strconv.Atoi(c)
+			cost, _ := strconv.ParseUint(c, 10, 8)
+			if err != nil {
+				return err
+			}
```
shoenig, line 95: the error is ignored in the parse but then we check it's not nil? this error has nothing to do with the parsing

api/api.go
```diff
@@ -1184,6 +1184,9 @@ func parseQueryMeta(resp *http.Response, q *QueryMeta) error {
 	if err != nil {
 		return fmt.Errorf("Failed to parse X-Nomad-LastContact: %v", err)
 	}
+	if last > uint64(math.MaxInt64) {
```
shoenig, line 1187: math.MaxInt64 is an untyped constant, do we need a cast here?

client/lib/numalib/detect_linux.go
tgross: Oops, fixed that too.

api/api.go
dduzgun-security: > I think it's because this check is to see if last (which is a uint64) is greater than time.Duration (which is int64)
https://github.com/hashicorp/nomad-enterprise/pull/1433#discussion_r1583156618

Because we do `q.LastContact = time.Duration(last) * time.Millisecond` on the next line.

tgross: Right, but I think @shoenig's point is that the `math.MaxInt64` doesn't need the cast here. It compiles without it.

Fixed.

PR #20556, opened 2024-05-10
deps: vendor containernetworking/plugins functions for net NS utils

tgross:
We bring in `containernetworking/plugins` for the contents of a single file, which we use in a few places for running a goroutine in a specific network namespace. This code hasn't needed an update in a couple of years, and a good chunk of what we need was previously vendored into `client/lib/nsutil` already.

Updating the library via dependabot is causing errors in Docker driver tests because it updates a lot of transient dependencies, and it's bringing in a pile of new transient dependencies like opentelemetry. Avoid this problem going forward by vendoring the remaining code we hadn't already.

Ref: https://github.com/hashicorp/nomad/pull/20146


tgross:
Note the semgrep rule is a false positive here: it's complaining about importing BUSL-licensed code into MPL2 packages, but this package is Apache licensed.

schmichael (approved):
Great idea

Issue #19786, opened 2024-01-19
sidecar_task support for volume_mount


mmcquillan:
### Proposal

Desire to mount system certificates to enable HTTPS for external services exposed through the Gateway Block (Terminating).

Use case of system certificates configured as a host volume, and exposing natively through a sidecar_task would enable Terminating Gateway CA implementation to be aligned with their current practices outside of the service mesh integration.

Two workarounds in the interim -
1. Custom Envoy Image with certs built into the image.
2. An exec task that copies the system certs to the allocation directory for the Terminating Gateway to read.

tgross:
Draft PR for this is up here: https://github.com/hashicorp/nomad/pull/20575

tgross:
Will ship in Nomad 1.8.0

PR #20575, opened 2024-05-13
connect: support `volume_mount` blocks for sidecar task overrides

tgross:
Users can override the default sidecar task for Connect workloads. This sidecar task might need access to certificate stores on the host. Allow adding the `volume_mount` block to the sidecar task override.

Also fixes a bug where `volume_mount` blocks would not appear in plan diff outputs.

Fixes: https://github.com/hashicorp/nomad/issues/19786


tgross:
Docs build error is because of: https://github.com/hashicorp/nomad/pull/20576

david-yu:
Thanks for working on this, this helps unblock Terminating Gateways.

api/consul.go
```diff
@@ -115,6 +115,7 @@ type SidecarTask struct {
 	LogConfig     *LogConfig             `mapstructure:"logs" hcl:"logs,block"`
 	ShutdownDelay *time.Duration         `mapstructure:"shutdown_delay" hcl:"shutdown_delay,optional"`
 	KillSignal    string                 `mapstructure:"kill_signal" hcl:"kill_signal,optional"`
+	VolumeMounts  []*VolumeMount         `hcl:"volume_mount,block"`
```
Juanadelacuesta, line 118: Maybe a bit of an overkill, but no tests?

nomad/structs/services.go
```diff
@@ -1422,6 +1426,10 @@ func (t *SidecarTask) Equal(o *SidecarTask) bool {
 		return false
 	}
 
+	if !slices.Equal(t.VolumeMounts, o.VolumeMounts) {
```
Juanadelacuesta, line 1429: slices.Equal takes into account the order of the elements, should we order them before comparing?

command/agent/job_endpoint.go
```diff
@@ -1319,21 +1319,7 @@ func ApiTaskToStructsTask(job *structs.Job, group *structs.TaskGroup,
...
-						Volume:          *mount.Volume,
-						Destination:     *mount.Destination,
-						ReadOnly:        *mount.ReadOnly,
-						PropagationMode: *mount.PropagationMode,
-						SELinuxLabel:    *mount.SELinuxLabel,
-					})
-			}
-		}
-	}
+	structsTask.VolumeMounts = apiVolumeMountsToStructs(apiTask.VolumeMounts)
```
Juanadelacuesta, line 1322: nice!

tgross: Will do.

nomad/structs/services.go
tgross: Actually, I had a look for how we treat this elsewhere and we always use `!slices.Equal`, even if the ordering isn't important. That lets operators force a job update without changing anything important, so I suspect we want to keep that behavior intact.

command/agent/job_endpoint.go
tgross: Good thing you pointed this out -- I should have been using `EqualFunc` in `nomad/structs/services.go`!

Juanadelacuesta (approved):
LGTM :)

PR #20588, opened 2024-05-14
[ui] Jobs Index page min-index and default empty-state styles

philrenaud:
Does 2 things on the newly reworked jobs index page:
1. Makes the [default helios empty state component](https://helios.hashicorp.design/components/application-state?tab=code) look more Nomad-ish (grey bg, widened)
2. Makes the minimum ?index on the /statuses request 1, instead of 0, so it successfully blocks instead of polls.

image


gulducat (approved):
:1st_place_medal:

mmcquillan (approved):

PR #20589, opened 2024-05-15
client: mount shared alloc dir as nobody

shoenig:
In the Unveil filesystem isolation mode we were mounting the shared
alloc dir with the UID/GID of the user of the task dir being mounted
and 0710 filesystem permissions. This was causing the actual shared 
alloc dir to become inaccessible to other tasks in the allocation
(a race where the last mounter wins). Instead mount the shared alloc 
dir as nobodywith 0777 filesystem permissions.


tgross (approved):
LGTM!

PR #20606, opened 2024-05-16
Backport of services: retry failed Nomad service deregistrations from client into release/1.6.x

hc-github-team-nomad-core:
## Backport

This PR is auto-generated from #20596 to be assessed for backporting due to the inclusion of the label backport/1.6.x.

---

When the allocation is stopped, we deregister the service in the alloc runner's `PreKill` hook. This ensures we delete the service registration and wait for the shutdown delay before shutting down the tasks, so that workloads can drain their connections. However, the call to remove the workload only logs errors and never retries them.

Add a short retry loop to the `RemoveWorkload` method for Nomad services, so that transient errors give us an extra opportunity to deregister the service before the tasks are stopped, before we need to fall back to the data integrity improvements implemented in #20590.

---

<details>
<summary> Overview of commits </summary>

  - 953bfcc31e7e283acb44ea04c3b288bd5d65dd65 

</details>


tgross (approved):

tgross:
Needs rebase after https://github.com/hashicorp/nomad/pull/20605 merges to 1.6.x

PR #20611, opened 2024-05-16
Permit Consul Connect Gateways to be used with podman

optiz0r:
Enable use of Consul Connect Gateways (ingresss/terminating/mesh) with podman task driver.

task driver for Connect-enabled tasks for sidecar services which used podman if any other task in the same task group was using podman or fell back to docker otherwise.

That PR did not consider consul connect gateways, which remained hardcoded to using docker task driver always.

This change applies the same heuristic also to gateway tasks, enabling use of podman.

Limitations: The heuristic only works where the task group containing the gateway also contains a podman task. Therefore it does not work for the ingress example in the docs
(https://developer.hashicorp.com/nomad/docs/job-specification/gateway#ingress-gateway) which uses connect native and requires the gateway be in a separate task.

I've updated existing tests to assume the docker driver, but have not yet added any new tests to verify the auto-detection of podman.

### DIFF
--- a/.changelog/20611.txt
+++ b/.changelog/20611.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+consul/connect: Attempt autodetection of podman task driver for Connect gateways
+```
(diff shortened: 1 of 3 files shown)


shoenig:
Thanks @optiz0r! This LGTM. I'm going to rebase with main and add a changelog entry and then get this merged.

optiz0r:
@shoenig Thanks. I think I managed to fix the existing broken tests, but I haven't managed to add any tests to validate the new behaviour.

shoenig (approved):

Issue #20462, opened 2024-04-19
Batch job reschedule attempts sometimes not being respected


sagebind:
### Nomad version
Output from `nomad version`

```
Nomad v1.7.6
BuildDate 2024-03-12T07:27:36Z
Revision 594fedbfbc4f0e532b65e8a69b28ff9403eb822e
```

### Operating system and Environment details

Output from `uname -a`:

```
Linux ip-10-151-99-125 6.5.0-1017-aws #17~22.04.2-Ubuntu SMP Mon Mar 25 20:47:22 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux
```

### Issue

It seems that `attempts` for `reschedule` set on a job is sometimes not being respected. This job configuration was formerly known to be for sure working when we were running Nomad 1.4 (we recently upgraded versions to catch up with the latest). We run quite a bit of small, discrete operations using batch jobs, and we use the reschedule policy to allow Nomad to automatically retry the operation with a new allocation on a different node, in case the underlying node had an issue, but we have `attempts` set to `2`.

Normally, this does what we expect -- if a task exits with a nonzero exit code or gets `OOM Killed`, the allocation fails (because `restart.attempts` is set to zero and mode is `"fail"`), and up to 2 additional allocations may be created to attempt the task again. However, since upgrading to 1.7, we've noticed that sometimes this limit is not being enforced, and Nomad is behaving as if `unlimited` were set to `true`, which it is not. This results in some failing tasks generating an infinite number of new allocations over time and filling up our clusters with wasted work, and causing quite a bit of problems for us. We've had jobs with `attempts` set to `2` somehow end up with hundreds of failed allocations. Here's a screenshot of the dashboard for a job that already had 63 before we found and purged it manually:

<img width="1189" alt="Screenshot 2024-04-18 at 11 18 49" src="https://github.com/hashicorp/nomad/assets/2192863/2c503e53-7e38-4217-9c23-dabf0587fb87">

So far, I haven't found a rhyme or reason yet as to why it seems like only _sometimes_ this behavior happens, but over time the percentage of our cluster filled with these "zombie" jobs grows until most of our capacity is taken up by them, leaving no room for new work.

My _current_ theory is that sometimes, when the task gets OOM Killed with the docker driver, the status code is seen as `0` and thus isn't incrementing some sort of counter somewhere that keeps track of the number of failed allocations so far. Normally, the event message is

```
Exit Code: 137, Exit Message: "OOM Killed"
```

[...]

PR #20615, opened 2024-05-16
tests: update disconnected client scheduler tests to avoid blocking

tgross:
While working on #20462, I discovered that some of the scheduler tests for disconnected clients making long blocking queries. The tests used `testutil.WaitForResult` to wait for an evaluation to be written to the state store. The evaluation was never written, but the tests were not correctly returning an error for an empty query. This resulted in the tests blocking for 5s and then continuing anyways.

In practice, the evaluation is never written to the state store as part of the test harness `Process` method, so this test assertion was meaningless. Remove the broken assertion from the two top-level tests that used it, and upgrade these tests to use `shoenig/test` in the process. This will save us ~50s per test run (for just the `test-groups/quick` group)

### DIFF
--- a/scheduler/generic_sched_test.go
+++ b/scheduler/generic_sched_test.go
@@ -18,7 +18,6 @@ import (
 	"github.com/hashicorp/nomad/helper/uuid"
 	"github.com/hashicorp/nomad/nomad/mock"
 	"github.com/hashicorp/nomad/nomad/structs"
-	"github.com/hashicorp/nomad/testutil"
 	"github.com/shoenig/test/must"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


scheduler/generic_sched_test.go
```diff
@@ -3722,18 +3730,18 @@ func TestServiceSched_StopAfterClientDisconnect(t *testing.T) {
 	}
 
 	for i, tc := range cases {
-		t.Run(fmt.Sprintf(""), func(t *testing.T) {
```
shoenig, line 3725: :upside_down_face:

shoenig (approved):
LGTM!

Issue #20042, opened 2024-02-28
Nomad (v1.7.5) executor processes consume a lot of CPU when job spawns child processes on Windows Server 


Settler:
### Nomad version
Nomad v1.7.5
BuildDate 2024-02-13T15:10:13Z
Revision 5f5d4646198d09b8f4f6cb90fb5d50b53fa328b8

### Operating system and Environment details
Windows Server 2019

### Issue
If job process spawns child processes, nomad executor v1.7.5 consumes a lot of CPU on Windows Server. But nomad executor v1.6.8 doesn't.
Example from our environment with 2 bare metal servers with 16 cores each and with different nomad versions:
![windows_resources](https://github.com/hashicorp/nomad/assets/104102/1c71e359-4676-470f-9b6c-192ebff3895b)
Upper graph – Server with nomad v1.7.5
Bottom graph – Server with nomad v1.6.8
Red horizontal lines with spikes – 95th percentile of overall CPU on machine.
Blue ares – CPU consumption sum of all nomad processes
We've tested different workloads (on each server at the same time):
- 1 alloc spawns 100 child processes
- 5 allocs spawn 20 child processes each
- 10 allocs spawn 10 child processes each
- 10 allocs spawn 20 child processes each

As you can see, nomad consumption isn't linear. Upper graph shows that nomad executors v1.7.5 from 10 allocs (or 10 jobs) with 20 child processes can consume up to 60% of 16 cores in total. But server with nomad executor v1.6.8 (bottom graph, blue area) is chilling and we can only see overall CPU spikes when our example powershell script spawns new processes.

### Reproduction steps
Run example job file on Windows. Tweak alloc count and subprocesses count variables to achieve ~75% overall CPU consumption.

#### Expected Result
CPU consumption of nomad executor processes same as 1.6.x version.

#### Actual Result
Inappropriate nomad executor CPU consumption.

### Job file (if appropriate)
```HCL
locals {
  allocs = 10
  subprocesses = 20
}

job "PSTest" {
  type = "service"
  
  group "PSTest-Group" {
    count = "${local.allocs}"
    
    task "PSTest" {
      driver = "raw_exec"
      kill_timeout = "10s"

      config {
        command = "powershell"
        args = ["./test.ps1"]
      }
      
      template {
        data = <<EOH
$c=${local.subprocesses}

for (($i = 0); $i -lt 3600; $i++)
{
    Write-Host $i
	for (($j = 0); $j -lt $c; $j++)
    {
[...]

PR #20619, opened 2024-05-16
windows: fix inefficient gathering of task processes

shoenig:
This PR overhauls the way processes are scanned on Windows for the exec driver(s).

A regression was introduced in Nomad 1.7 where we would invoke an [expensive syscall](https://github.com/hashicorp/nomad/blob/release/1.7.0/drivers/shared/executor/procstats/list_default.go#L40) for every descendant of a process to build a tree of the processes associated with a task. Instead, we restore the behavior of scanning the process table once and rebuild the process tree manually.

Fixes https://github.com/hashicorp/nomad/issues/20042

### DIFF
--- a/drivers/shared/executor/procstats/list_windows.go
+++ b/drivers/shared/executor/procstats/list_windows.go
@@ -1,0 +1,73 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+//go:build windows
+
+package procstats
+
+import (
+	"github.com/hashicorp/go-set/v2"
+	"github.com/mitchellh/go-ps"
+)
+
+func gather(procs map[int]ps.Process, family set.Collection[int], root int, candidate ps.Process) bool {
+	if candidate == nil {
+		return false
+	}
+	pid := candidate.Pid()
+	if pid == 0 || pid == 1 {
+		return false
+	}
+	if pid == root {
+		return true
+	}
+	parent := procs[candidate.PPid()]
+	result := gather(procs, family, root, parent)
+	if result {
+		family.Insert(pid)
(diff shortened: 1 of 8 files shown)


drivers/shared/executor/procstats/list_windows.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	return result
+}
+
+func list(executorPID int, processes func() ([]ps.Process, error)) set.Collection[ProcessID] {
+	family := set.From([]int{executorPID})
+
+	all, err := processes()
+	if err != nil {
+		return set.New[ProcessID](0)
```
tgross, line 45: Could we at least return the executor PID here? Or is the thinking that if this call fails we can't trust any of the results?

tgross (approved):
LGTM!

Shows what a leetcode problem this is that I immediately started thinking about memoizing by checking if the `candidate.PPid()` was already in `family`... but for that to be at all meaningful the individual tasks would have to have _so_ many processes.

tgross:
@shoenig you don't need a 1.8.x backport label on this, as we'll cut the RC release from main

drivers/shared/executor/procstats/list_windows.go
shoenig: Oh good idea, let me add that

shoenig:
> but for that to be at all meaningful the individual tasks would have to have so many processes.

Yeah we could definitely add base cases for:
- if the parent is already known to be in the family 
- if the parent is already known to _not_ be in the family 

But at that point you're optimizing microseconds when the original problem is this horrible process table snapshot syscall that takes 10's of milliseconds. I'd rather the code just be readable.

PR #20620, opened 2024-05-16
doc: specify ca cert needs to be shared

schmichael:
(Certificate management is the worst.)

Had some confusion from a user thinking they _had_ to use a lb/proxy to configure the Vault integration. I noticed our docs about mTLS/verify_client lacked mentioning that Vault needs to trust Nomad's CA, so I added some words to hopefully make that more clear? The words do not seem like good words so if you have gooder words please suggest them.

### DIFF
--- a/website/content/docs/integrations/vault/acl.mdx
+++ b/website/content/docs/integrations/vault/acl.mdx
@@ -295,7 +295,8 @@ your Vault and Nomad clusters are configured and deployed.
 It is highly recommended to use [mutual TLS][tutorial_mtls] in production
 deployments of Nomad. With mTLS enabled, the [`tls.verify_https_client`][]
 configuration must be set to `false` since it is not possible to provide client
-certificates to the Vault auth method.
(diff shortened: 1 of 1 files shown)


david-yu:
I'm curious does that mean creating the Vault auth method for nomad with that public CA via the certificate parameter? https://developer.hashicorp.com/vault/docs/auth/cert#configuration

schmichael:
> I'm curious does that mean creating the Vault auth method for nomad with that public CA via the certificate parameter? https://developer.hashicorp.com/vault/docs/auth/cert#configuration

Hm, that certificate parameter is for the _certificate_ auth method, not the JWT auth method. Poking around I think [`jwks_ca_pem`](https://developer.hashicorp.com/vault/api-docs/auth/jwt#jwks_ca_pem) is the parameter people can configure the certificate with?

I'll update the docs to point people to it directly.

jrasell (approved):

david-yu (approved):

PR #22100, opened 2024-05-21
[ui] Better UX with filter expressions in the jobs index search box

philrenaud:
In the jobs index search box, we interpret non-expression-looking strings as filter expressions of `Name contains ________` type.

This is neat and probably how users expect searching for jobs to work (just type in a job name, that job shows up) but we are being over-eager with our solution, notably by prepending "Name contains" in the search box before your cursor. This is particularly troublesome if you were doing something like typing "Type contains sys" but stopped for a moment while thinking about what the operator should be. You'd end up with "Name contains Type" before you know it.

Instead, we make that search / apply the filter expression in the URL and API call, but keep the search box itself untouched.

Secondly, we now give users suggestions for when they attempt a filter expression that's lowercase (writing "type contains foo" yields "Did you mean Type?") and prompt them for when their key is generally not valid (typing "foo != bar" yields a list of possible filter expression keys)

No filter:
image

Single word filter, note the URL's filter expression:
image

Mistyped/lowercased expression key:
image

Not at all an expression key:
image

### DIFF
--- a/ui/app/controllers/jobs/index.js
+++ b/ui/app/controllers/jobs/index.js
@@ -574,5 +574,21 @@ export default class JobsIndexController extends Controller {
     this.updateFilter();
   }
 
+  // A list of combinatorial filters to show off filter expressions
+  // Make use of our various operators, and our various known keys
+  @computed('filter')
+  get exampleFilter() {
+    let examples = [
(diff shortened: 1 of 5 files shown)


gulducat (approved):
love nice training wheels for cool new features ^_^

PR #12319, opened 2022-03-18
scheduler: `RescheduleTracker` dropped if follow-up fail placements

tgross:
This PR is WIP and currently includes just a failing unit test demonstrating https://github.com/hashicorp/nomad/issues/12147

When an allocation fails it triggers an evaluation. The evaluation is
processed and the scheduler sees it needs to reschedule, which
triggers a follow-up eval. The follow-up eval creates a plan to
`(stop 1) (place 1)`. The replacement alloc has a `RescheduleTracker`
(or gets its `RescheduleTracker` updated).

But in the case where the follow-up eval can't place all allocs (there
aren't enough resources), it can create a partial plan to
`(stop 1) (place 0)`. It then creates a blocked eval. The plan applier
stops the failed alloc. Then when the blocked eval is processed, the
job is missing an allocation, so the scheduler creates a new
allocation. This allocation is _not_ a replacement from the
perspective of the scheduler, so it's not handed off a
`RescheduleTracker`.

PR #22198, opened 2024-05-22
refactor scheduler tests for node down/disconnected

tgross:
While working on #20462 #12319 I found that some of our scheduler tests around down nodes or disconnected clients were enforcing invariants that were unclear. This changeset pulls out some minor refactorings so that the bug fix PR is easier to review. This includes:

* Migrating a few tests from `testify` to `shoenig/test` that I'm going to touch in #12319 anyways.
* Adding test names to the node down test
* Update the disconnected client test so that we always re-process the pending/blocked eval it creates; this eliminates 2 redundant sub-tests.
* Update the disconnected client test assertions so that they're explicit in the test setup rather than implied by whether we re-process the pending/blocked eval.

Ref: https://github.com/hashicorp/nomad/issues/20462
Ref: https://github.com/hashicorp/nomad/pull/12319

### DIFF
--- a/scheduler/generic_sched_test.go
+++ b/scheduler/generic_sched_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/hashicorp/nomad/helper/uuid"
 	"github.com/hashicorp/nomad/nomad/mock"
 	"github.com/hashicorp/nomad/nomad/structs"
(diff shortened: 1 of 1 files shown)


jrasell (approved):
LGTM!

PR #22200, opened 2024-05-22
Add ACL auto bootstrapping option

angrycub:
This enables Nomad to start and bootstrap the ACL subsystem without having to run the `nomad acl bootstrap` command separately. This does require a certain amount of post-deployment hygeine around removing these values after using them in bootstrap.

The primary usecases for this being
- Facilitating starting debugger-built binaries in ACL-enabled mode when performing interactive debugging
- Reducing complexity of e2e tests that require ACL-enabled clusters


nomad/server.go
```diff
@@ -1653,6 +1659,42 @@ func (s *Server) setupSerf(conf *serf.Config, ch chan serf.Event, path string) (
...
+	if !s.config.ACLEnabled || s.config.ACLBootstrapToken == "" {
+		// Exit without bootstrapping the ACL system
+		return nil
+	}
+
+	resp := structs.ACLTokenUpsertResponse{}
+	err := s.RPC("ACL.Bootstrap",
+		&structs.ACLTokenBootstrapRequest{BootstrapSecret: s.config.ACLBootstrapToken},
+		&resp,
+	)
```
tgross, line 1672: I know we chatted about this a bit offline, but I'm thinking through this a bit more... because this function is blocking, we need to make an RPC call to the leader in order to start the server _at all_ (although I like that you're not returning an error on error at least). Ideally users will remove the token from the config, obviously, but I can definitely see a case where users don't, then add servers afterwards, and have a restart hang as a result.

I think we want to call `s.State().CanBootstrapACLToken()` here first, and only if that doesn't return an index should we try to bootstrap again. I don't think we really care about the reset index here?

nomad/server.go
```diff
@@ -1653,6 +1659,42 @@ func (s *Server) setupSerf(conf *serf.Config, ch chan serf.Event, path string) (
...
+		} else {
+			// Unexpected Error
+			s.logger.Warn("error in auto bootstrap; you must manually bootstrap ACLs", "err", err)
+		}
+		return nil
+	}
+
+	// Nonsense result
+	if len(resp.Tokens) != 1 {
+		return fmt.Errorf("received incorrect token count. expected 1; got %d", len(resp.Tokens))
```
tgross, line 1687: There's no prefix indicating what this token is for. Maybe something like:

```suggestion
		return fmt.Errorf("received incorrect token count during ACL bootstrap. expected 1; got %d", len(resp.Tokens))
```

tgross:
Ref: https://hashicorp.atlassian.net/browse/NET-10379

PR #22228, opened 2024-05-23
consul: set partition for gateway config entries

tgross:
When we write Connect gateway configuation entries from the server, we're not passing in the intended partition. This means we're using the server's own partition to submit the configuration entries and this may not match. Note this requires the Nomad server's token has permission to that partition.

Also, move the config entry write after we check Sentinel policies. This allows us to return early if we hit a Sentinel error without making Consul RPCs first.

Ref: https://hashicorp.atlassian.net/browse/NET-9529


pkazmierczak (approved):
LGTM!

PR #22232, opened 2024-05-23
[ui] Remove Storybook

philrenaud:
This PR removes [Storybook](https://storybook.js.org/) from Nomad.

We had been using it as a tool to preview individual components, but I had let it get out of date with our more recent components.

Integration with [Helios](https://helios.hashicorp.design/) has meant that an increasing number of the components we use in Nomad have a dedicated team outside the project but within the org to manage them, complete with galleries. Additionally, the maintenance cost of handling divergent upgrade paths for non-negotiable tools like Node and Ember has meant that Storybook ran into [compatibility concerns](https://github.com/hashicorp/nomad/pull/17573) that we lack the time to concentrate on.


dduzgun-security (approved):
LGTM! Thanks a lot for looking into that

tgross (commented):
@philrenaud we should make sure we add backport labels on this so that we're not carrying it along. And maybe once this merges we should take a pass-thru the open dependabot PRs to see if any can be closed now that we don't have this?

dduzgun-security:
@tgross luckily the alerts should auto-close on the next scans when merged in to main. I'll take a 👀 into them in case it misses some

PR #22234, opened 2024-05-23
docs: small typo

david-yu:
Small change, removal of extraneous close paren


jrasell (approved):
We should add the backport labels, so this gets into the correct versioned docs and the live website.

PR #22312, opened 2024-05-24
Consul Connect: Fix validation with multiple local_bind_socket_paths

sundbry:
When a consul connect sidecar service is defined with multiple local_bind_socket_path upstreams, validation fails due to duplicate socket address bindings on `:0` being detected:

```
Error submitting job: Unexpected response code: 500 (1 error occurred:
        * Consul Connect services "caddy-https" and "caddy-https" in group "caddy" using same address for upstreams (:0))
```

This patch validates `local_bind_socket_path` upstreams separately from `local_bind_address` sockets.

### DIFF
--- a/nomad/job_endpoint_hook_connect.go
+++ b/nomad/job_endpoint_hook_connect.go
@@ -569,7 +569,7 @@ func groupConnectValidate(g *structs.TaskGroup) error {
 }
 
 func groupConnectUpstreamsValidate(g *structs.TaskGroup, services []*structs.Service) error {
-	listeners := make(map[string]string) // address -> service
+	listeners := make(map[string]string) // address or path-> service
 
 	var connectBlockCount int
 	var hasTproxy bool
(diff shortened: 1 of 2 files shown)


nomad/job_endpoint_hook_connect.go
```diff
@@ -570,6 +570,7 @@ func groupConnectValidate(g *structs.TaskGroup) error {
 
 func groupConnectUpstreamsValidate(g *structs.TaskGroup, services []*structs.Service) error {
 	listeners := make(map[string]string) // address -> service
+	sockets := make(map[string]string) // socket -> path
```
tgross, line 573: We should be able to safely reuse the listeners map here (this would have been `// path -> service`, as well), which will save us some allocations.

```suggestion
	listeners := make(map[string]string) // address or path -> service
```

That'll let you reduce some of the branches below as well

tgross (changes requested):
Hi @sundbry! Thanks for this fix. I've left a comment about reusing the map. Also:

* You'll want to `gofmt -s` the file when you're done to make the lint check happy
* Can you run `make cl` to create a changelog entry?

sundbry:
Hi @tgross, thanks for the review - I have resubmitted it after changes + added a changelog entry.

PR #22328, opened 2024-05-24
[ui] Helios and Power Select upgrades

philrenaud:
- Upgrading [Helios](https://helios.hashicorp.design/) to 4.3.0
- ember-element-helper was previously pinned because of an old version of ember-power-select, which is now also upgraded, so pin removed.


ui/package.json
```diff
@@ -106,7 +106,7 @@
     "ember-on-resize-modifier": "^1.0.0",
     "ember-overridable-computed": "^1.0.0",
     "ember-page-title": "^7.0.0",
-    "ember-power-select": "^4.1.7",
+    "ember-power-select": "^7.2.0",
```
philrenaud, line 109: Note: I could upgrade this to v8+, but that requires adding extra template code per https://ember-power-select.com/docs/migrate-7-0-to-8-0 and I think my time would be better suited moving away from ember-power-select raw usage generally (we use it for our global search, among other things). Helios' dropdown component will be a more than sufficient replacement for it.

alex-ju (approved):
Looks good! Checked for the changes introduced in 4.x and the only relevant ones are the contextual component renames indeed

PR #22392, opened 2024-05-27
[ui] Jobs index page badge for when a job has a paused task

philrenaud:
image
image

TODO:
- Should I add logic for "only care about latest version allocs' tasks", or "only care about non-replaced allocs' tasks"?


david-yu (approved):
Looks good thanks!

schmichael (approved):

PR #22397, opened 2024-05-28
api: Add Notes field to service checks

### LINKED ISSUE #21372: Allow setting Notes on consul health checks
Hey there!

It seems that it's impossible to set the `Notes` field of [Consul health checks](https://pkg.go.dev/github.com/hashicorp/consul/api#AgentServiceCheck) using [Nomad's](https://pkg.go.dev/github.com/hashicorp/nomad/api#ServiceCheck).

Is it expected? If no, can we add it? I'm happy to provide a PR.

nicoche:
Resolves #21372

### DIFF
--- a/.changelog/22397.txt
+++ b/.changelog/22397.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+consul hc: add support for setting Notes field
+```
--- a/.changelog/22397.txt
+++ b/.changelog/22397.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-consul hc: add support for setting Notes field
+api: Add support for setting Notes field for Consul health checks
 ```
--- a/command/alloc_status.go
+++ b/command/alloc_status.go
@@ -374,14 +374,14 @@ func formatAllocNomadServiceChecks(allocID string, client *api.Client) string {
 	} else if len(statuses) == 0 {
 		return ""
 	}
-	results := []string{"Service|Task|Name|Mode|Status"}
+	results := []string{"Service|Task|Name|Mode|Status|Notes"}
 	for _, status := range statuses {
 		task := "(group)"
 		if status.Task != "" {
 			task = status.Task
 		}
-		// check | group | mode | status
-		s := fmt.Sprintf("%s|%s|%s|%s|%s", status.Service, task, status.Check, status.Mode, status.Status)
+		// check | group | mode | status | notes
+		s := fmt.Sprintf("%s|%s|%s|%s|%s|%s", status.Service, task, status.Check, status.Mode, status.Status, status.Notes)
 		results = append(results, s)
 	}
 	sort.Strings(results[1:])
--- a/command/alloc_status.go
+++ b/command/alloc_status.go
@@ -374,14 +374,14 @@ func formatAllocNomadServiceChecks(allocID string, client *api.Client) string {
 	} else if len(statuses) == 0 {
 		return ""
 	}
-	results := []string{"Service|Task|Name|Mode|Status|Notes"}
+	results := []string{"Service|Task|Name|Mode|Status"}
 	for _, status := range statuses {
 		task := "(group)"
 		if status.Task != "" {
 			task = status.Task
 		}
-		// check | group | mode | status | notes
-		s := fmt.Sprintf("%s|%s|%s|%s|%s|%s", status.Service, task, status.Check, status.Mode, status.Status, status.Notes)
+		// check | group | mode | status
+		s := fmt.Sprintf("%s|%s|%s|%s|%s", status.Service, task, status.Check, status.Mode, status.Status)
 		results = append(results, s)
 	}
 	sort.Strings(results[1:])
--- a/nomad/structs/services.go
+++ b/nomad/structs/services.go
@@ -459,6 +459,11 @@ func (sc *ServiceCheck) validateConsul() error {
 		return fmt.Errorf("failures_before_warning not supported for check of type %q", sc.Type)
 	}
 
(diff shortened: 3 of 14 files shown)


nomad/structs/services.go
```diff
@@ -68,6 +68,7 @@ type ServiceCheck struct {
 	Interval               time.Duration       // Interval of the check
 	Timeout                time.Duration       // Timeout of the response from the check before consul fails the check
 	InitialStatus          string              // Initial status of the check
+	Notes                  string              // Specifies arbitrary information for humans. This is not used by Consul internally
```
pkazmierczak, line 71: Do you think we could add a little validation on this field? Maybe we could at least limit the length of notes to 250 characters? There's already [a method](https://github.com/hashicorp/nomad/blob/d4cec2afc484917af27e98531bc43c91be232cf2/nomad/structs/services.go#L415) for validating Consul fields.

command/alloc_status.go
```diff
@@ -374,14 +374,14 @@ func formatAllocNomadServiceChecks(allocID string, client *api.Client) string {
...
+	results := []string{"Service|Task|Name|Mode|Status|Notes"}
 	for _, status := range statuses {
 		task := "(group)"
 		if status.Task != "" {
 			task = status.Task
 		}
-		// check | group | mode | status
-		s := fmt.Sprintf("%s|%s|%s|%s|%s", status.Service, task, status.Check, status.Mode, status.Status)
+		// check | group | mode | status | notes
+		s := fmt.Sprintf("%s|%s|%s|%s|%s|%s", status.Service, task, status.Check, status.Mode, status.Status, status.Notes)
```
pkazmierczak, line 384: this unit test fails, the `status.Notes` field is empty

pkazmierczak (changes requested):
Hey @nicoche, thanks for the PR! Looks good! I left some minor comments, could you have a look? Also, would you add a changelog file? You can use the `make cl` command to generate it.

nomad/structs/services.go
nicoche: Good idea, pushed a fix

command/alloc_status.go
nicoche: Nice, thanks for checking. I just pushed https://github.com/hashicorp/nomad/pull/22397/commits/bf4324cd6f23047c0fbe2fd799235261fa2068bb to try to address this, I hope it helps. Running `go test` in this package doesn't work on my laptop, even on the `main` branch.

Did `NSD` in `TestAllocStatusCommand_NSD_Checks` stand for `Nomad Service D...iscovery?`

tgross: Should we really report notes in a table like this at all? Won't the note push off the edge of the terminal most of the time?

nicoche: We intend to "abuse" this field to store structured information, so having it in the CLI table could definitely come in handy at some point for us. However, that's just our use case and we don't strictly need it, so I can remove that part from the PR if you want

nicoche: Also the test I wrote is not working, so let me know what you guys decide so I can fix it or delete it

pkazmierczak: Hey @nicoche, I took another look at this and here's the bit I don't understand. You're implementing `notes` field support for Consul service health checks. But this command, `nomad alloc status`, will not display details about services if they aren't managed by Nomad (service block `provider` field must be set to `nomad`). This is also what causes unit test issues: mock Consul service jobs with notes won't change the output of `nomad alloc status`, and for mock Nomad service jobs, the `notes` field will always be empty.

pkazmierczak: if you feel like it, you may [add](https://github.com/hashicorp/nomad/blob/953bfcc31e7e283acb44ea04c3b288bd5d65dd65/client/serviceregistration/nsd/nsd.go#L351) the `notes` field to nomad managed services, but I think this is maybe out of scope and, more importantly, not what you want?

nicoche: Hey guys! I simply removed that chunk 🙂

nomad/structs/services.go
```diff
@@ -454,6 +459,11 @@ func (sc *ServiceCheck) validateConsul() error {
 		return fmt.Errorf("failures_before_warning not supported for check of type %q", sc.Type)
 	}
 
+	// Arbitrary value, we could bump it if needed
+	if len(sc.Notes) > 255 {
+		return fmt.Errorf("notes must 255 characters or lower")
```
pkazmierczak, line 464:
```suggestion
		return fmt.Errorf("notes must not be longer than 255 characters")
```

nomad/structs/services_test.go
```diff
@@ -1992,6 +1993,26 @@ func TestService_Validate(t *testing.T) {
...
+						Type:     "http",
+						Path:     "/",
+						Interval: 1 * time.Second,
+						Timeout:  3 * time.Second,
+						Notes: strings.Repeat("A", 256),
+					},
+				},
+			},
+			expErr:    true,
+			expErrStr: "notes must 255 characters or lower",
```
pkazmierczak, line 2014:
```suggestion
			expErrStr: "notes must not be longer than 255 characters",
```

pkazmierczak (approved):
thanks @nicoche, I just left some minor comments.

.changelog/22397.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+consul hc: add support for setting Notes field
```
pkazmierczak, line 2:
```suggestion
api: Add support for setting Notes field for Consul health checks
```

tgross (commented):
LGTM once @pkazmierczak's changes are addressed.

PR #22399, opened 2024-05-29
[deps] Update go-winio fork and rebase on current upstream

angrycub:
This replaces the forked `go-winio` from endocrimes/go-winio to angrycub/go-winio. This goal being to eventually move the fork to a HashiCorp owned repo.  My fork brings the library up to current main plus @endocrimes 's patches for `ListenOnlyPipe`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.21
 
 // Pinned dependencies are noted in github.com/hashicorp/nomad/issues/11826.
 replace (
-	github.com/Microsoft/go-winio => github.com/endocrimes/go-winio v0.4.13-0.20190628114223-fb47a8b41948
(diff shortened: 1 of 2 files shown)


tgross (approved):
LGTM as far as Nomad goes.

For the forked repo, I'd suggest we might want to have the "forked from" point to the original upstream and then apply Dani's patch on top of that (preserving authorship, obviously!), so that someone can see at a glance what patch we're carrying and pull in changes cleanly. Alternately, could we get the patch upstreamed? It'd be nice to let upstream take care of it going forward.

endocrimes:
@tgross me and @notnoop have tried to little avail (https://github.com/microsoft/go-winio/pull/170)

tgross:
Ref: https://hashicorp.atlassian.net/browse/NET-10376

PR #22406, opened 2024-05-29
e2e: add tests for exec2 task driver

shoenig:
A first pass at getting some e2e tests setup on the Nomad side for the exec2 task driver. It includes running the countash example with a custom envoy task for the sidecar proxy, which is a pretty good test of the driver functionality.

I tweaked the `e2e/Makefile` to include the Consul license file (you need to have `CONSUL_LICENSE_PATH` set), which is basically a required parameter now.

Also discovered https://github.com/hashicorp/nomad-driver-exec2/issues/29 which I will get around to fixing next.

### DIFF
--- a/e2e/terraform/packer/ubuntu-jammy-amd64/setup.sh
+++ b/e2e/terraform/packer/ubuntu-jammy-amd64/setup.sh
@@ -140,7 +140,7 @@ sudo chmod +x ${NOMAD_PLUGIN_DIR}/nomad-driver-exec2
 
 # Envoy
 echo "Installing Envoy"
-sudo curl -s -S -L -o /opt/bin/envoy https://github.com/envoyproxy/envoy/releases/download/v1.30.1/envoy-1.30.1-linux-x86_64
+sudo curl -s -S -L -o /opt/bin/envoy https://github.com/envoyproxy/envoy/releases/download/v1.29.4/envoy-1.29.4-linux-x86_64
 sudo chmod +x /opt/bin/envoy
 
 # ECS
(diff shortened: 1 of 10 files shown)


e2e/terraform/packer/ubuntu-jammy-amd64/setup.sh
```diff
@@ -132,6 +133,16 @@ sudo mv /tmp/nomad-pledge-driver ${NOMAD_PLUGIN_DIR}
...
+# Exec2
+echo "Installing Exec2 Driver"
+sudo hc-install install --path ${NOMAD_PLUGIN_DIR} --version v0.1.0-alpha.2 nomad-driver-exec2
+sudo chmod +x ${NOMAD_PLUGIN_DIR}/nomad-driver-exec2
+
+# Envoy
+echo "Installing Envoy"
+sudo curl -s -S -L -o /opt/bin/envoy https://github.com/envoyproxy/envoy/releases/download/v1.30.1/envoy-1.30.1-linux-x86_64
+sudo chmod +x /opt/bin/envoy
+
```
david-yu, line 145: I'm guessing we probably want to only test to 1.29.x since Consul 1.19 (upcoming release) is targeting 1.29.x cc @zalimeni

zalimeni: Lacking context for how Envoy testing in Nomad has been done previously, yes I'd agree that targeting the same minor version line (1.29 latest) on `<latest-release-series>` in both products is probably ideal, if there's no other reason to test further out 👍🏻

shoenig: thanks for pointing this out; i've adjusted the version to be 1.29.4

david-yu (approved):

PR #22420, opened 2024-05-30
include exec2 vs exec benefits

djschnei21:


david-yu:
@shoenig Could you review for accuracy? Thanks @djschnei21!

website/content/plugins/drivers/exec2.mdx
```diff
@@ -305,6 +305,47 @@ driver.exec2.unveil.defaults    = true
 driver.exec2.unveil.tasks       = true
 ```
 
+### Benefits of exec2 Over exec
+
+#### Security Benefits
+
+##### Enhanced Filesystem Isolation with Landlock LSM
+- **Landlock LSM**: The `exec2` driver leverages the Landlock Linux Security Module (LSM) for filesystem isolation, which is more secure than the traditional `chroot` mechanism used by the `exec` driver.
```
shoenig, line 313: I think there is a misunderstanding about the value prop of exec2 - it's primarily around the performance of an exec-like driver, not the absolute security benefits. exec and exec2 provide approximately the same security - if anything exec2 is more lenient in that it allows the cluster operator to open specific parts of the host filesystem to the tasks whereas exec is fully isolated apart from mounted volumes. where exec2 shines is that it will never create a chroot in which to run each task, which can be a very slow operation on many systems depending on filesystem performance (think network drives) and the amount of content in the chroot being created.

djschnei21: Hey @shoenig, for the security benefits section I was mostly attempting to unpack the first half of the "Chroot Challenges" section of [NMD-190](https://docs.google.com/document/d/1cDH-qy0nb6Fc4PyEEqs_HfoNv_BXUwes7eAr2MqE4os/edit?usp=sharing).  Mainly contrasting these two claims:

- "The fundamental problem with chroots is this: they are not secure."
- "A process that has been locked (by Landlock LSM) cannot be unlocked, even if the process is being run as root"

website/content/plugins/drivers/exec2.mdx
```diff
@@ -305,6 +305,47 @@ driver.exec2.unveil.defaults    = true
...
+- **No Chroot Overhead**: Unlike `exec`, which relies on `chroot` and requires copying or hardlinking files into a chroot environment, `exec2` uses Landlock LSM for filesystem isolation.
+  - **Faster Task Launch**: Eliminates the need for extensive disk operations, significantly reducing the time and IOPS required to launch tasks.
+
+##### Efficient Resource Utilization
+- **Minimal Overhead**: `exec2` is designed to have very low startup times and minimal overhead in terms of CPU and memory utilization.
+  - **Lightweight Isolation**: By using native Linux tools like `nsenter` and `unshare`, `exec2` avoids the overhead associated with container runtimes and third-party libraries.
+
+##### Simplified Configuration and Management
+- **Unified Configuration**: `exec2` provides a unified and simplified configuration for filesystem and resource isolation, reducing the complexity and potential for misconfiguration.
+  - **Dynamic Defaults**: Automatically prunes default paths based on the system's runtime environment, ensuring only necessary paths are unveiled, further optimizing performance.
```
tgross, line 343: Probably syntax issues here: these don't have equal indentation. Also, we hard-wrap docs please.

website/content/plugins/drivers/exec2.mdx
```diff
@@ -305,6 +305,47 @@ driver.exec2.unveil.defaults    = true
...
+
+## Conclusion
+
+The `exec2` driver for Nomad offers significant security and performance improvements over the traditional `exec` driver. By leveraging modern Linux kernel features such as Landlock LSM, cgroups v2, and namespace isolation, `exec2` provides a more secure and efficient environment for running non-container workloads on Linux systems.
```
tgross, line 347: This is reference documentation, it doesn't need a conclusion.

website/content/plugins/drivers/exec2.mdx
djschnei21: indentation was intentional to make a sub-point
image

Hard-wrap added

website/content/plugins/drivers/exec2.mdx
david-yu: @shoenig Could you propose a different approach here or way to phrase the benefits of exec2 instead?

jrasell:
Closing this PR as the documentation has been moved out of this repo and re-organised since it was raised.

PR #22421, opened 2024-05-30
Parse hvs{} blocks in tasks

ncabatoff:


ncabatoff:
> Thanks @ncabatoff. Any idea what else is needed to get this productionized and merged?

Well, I'm not really a Nomad dev, so I have some idea, but l don't know entirely.  There's a whole [checklist](https://github.com/hashicorp/nomad/blob/main/contributing/checklist-jobspec.md) which I definitely didn't complete, so that's probably the first thing to do.

PR #22424, opened 2024-05-30
[ui] Squashing Ember deprecations

philrenaud:
Getting pretty tired of these party-poopers:
image

Note: a few of the other deprecation warnings are such that an upgrade to Ember 4+ is warranted if we want them to go away badly enough. This gets rid of the noisiest one (without having to mess with our deprecation-workflow.js file)


tgross (approved):
LGTM :+1:

PR #22439, opened 2024-05-31
[deps] Update hcl/v2 to latest patch v2.20.2-0.20240517235513-55d9c02d147d

angrycub:
Part of the foundational work for updating the Sentinel version


tgross (approved):
LGTM! (once CI is green)

We should backport this to all supported versions, just to make sure we can push any future HCL (or Sentinel) bug fixes into backports without hassle.

PR #22440, opened 2024-05-31
[tools] Changelog TUI v0.1

angrycub:
This builds off Seth's Changelog file builder and adds a TUI using the [charmbracelet/huh](https://github.com/charmbracelet/huh) form library.  Adds a makefile target, `make cltui` to run the tool.

**Current State**
The TUI application will perform the same basic workflow as the cl-entry application, including value validation.

**Future Vision**
The eventual goal is to leverage git and the github API to make draft PR creation and rough commit message generation a single process flow


GNUmakefile
```diff
@@ -431,6 +431,10 @@ missing: ## Check for packages not being tested
 cl: ## Create a new Changelog entry
 	@go run -modfile tools/go.mod tools/cl-entry/main.go
 
+.PHONY: cltui
+cltui: ## Create a new Changelog entry
```
tgross, line 435: Just so they have different help text:

```suggestion
cltui: ## Create a new Changelog entry interactively
```

Presumably we're not replacing yet but that should probably be the long-term goal so we don't have extra tools littered around :grinning:

tools/cl-entry-tui/main.go
```diff
@@ -0,0 +1,122 @@
...
+				Title("Select the change kind:").
+				Options(
+					huh.NewOption("Bug", "bug"),
+					huh.NewOption("Improvement", "improvement"),
+					huh.NewOption("Security", "security"),
+					huh.NewOption("Breaking Change", "breaking-change"),
+					huh.NewOption("Note", "note"),
+					huh.NewOption("Deprecation", "deprecation"),
+				).
+				Value(&pr.Label), // store the chosen option in the "burger" variable
```
tgross, line 65: "burger"?

tools/go.mod
```diff
@@ -5,8 +5,36 @@ go 1.21
...
+	github.com/mattn/go-runewidth v0.0.15 // indirect
+	github.com/muesli/ansi v0.0.0-20230316100256-276c6243b2f6 // indirect
+	github.com/muesli/cancelreader v0.2.2 // indirect
+	github.com/muesli/termenv v0.15.2 // indirect
+	github.com/rivo/uniseg v0.4.7 // indirect
+	github.com/xo/terminfo v0.0.0-20220910002029-abceb7e1c41e // indirect
+	golang.org/x/sync v0.7.0 // indirect
+	golang.org/x/sys v0.20.0 // indirect
+	golang.org/x/text v0.15.0 // indirect
+)
```
tgross, line 40: Oof

tgross (approved):
LGTM!

tools/go.mod
shoenig: TBH this is the only reason I'm hesitant to just rubber stamp this and also replace the original `make cl` - this is like 10's of thousands of lines of code we now "own" just for a little wizard - the price isn't really worth it IMO

angrycub:
Don't want to push it forward if we don't think it provides value; will probably just close it. I'll see if I can do something closer to the stdlib using golang.org/x/term

PR #22518, opened 2024-06-01
[bug] Fix cgroup setup for non-default devices

lukepalmer:
Using a device inside a container requires two things:
1. binding the device
2. allowing the device in cgroup rules

The nomad linux executor sets up cgroups only for default devices and does not correctly set up other devices specified in configuration.

This change sets up cgroup rules for all devices needed by the container. An important non-default device that is relevant here is /dev/fuse.

I don't believe this has worked at any point in history; a backport to 1.7 would be appreciated if that's possible.

### DIFF
--- a/drivers/shared/executor/executor_linux.go
+++ b/drivers/shared/executor/executor_linux.go
@@ -677,6 +677,10 @@ func configureIsolation(cfg *runc.Config, command *ExecCommand) error {
 		cfg.Devices = append(cfg.Devices, devs...)
 	}
 
+	for _, device := range cfg.Devices {
+		cfg.Cgroups.Resources.Devices = append(cfg.Cgroups.Resources.Devices, &device.Rule)
+	}
+
 	cfg.Mounts = []*runc.Mount{
 		{
 			Source:      "tmpfs",
@@ -844,10 +848,6 @@ func (l *LibcontainerExecutor) newLibcontainerConfig(command *ExecCommand) (*run
 		Version: "1.0.0",
 	}
 
-	for _, device := range specconv.AllowedDevices {
-		cfg.Cgroups.Resources.Devices = append(cfg.Cgroups.Resources.Devices, &device.Rule)
-	}
-
 	configureCapabilities(cfg, command)
 
 	// children should not inherit Nomad agent oom_score_adj value
@@ -897,6 +897,7 @@ func cmdDevices(driverDevices []*drivers.DeviceConfig) ([]*devices.Device, error
 			return nil, fmt.Errorf("failed to make device out for %s: %v", d.HostPath, err)
 		}
 		ed.Path = d.TaskPath
+		ed.Allow = true // rules will be used to allow devices via cgroups
 		r[i] = ed
 	}
 
--- a/drivers/shared/executor/executor_linux_test.go
+++ b/drivers/shared/executor/executor_linux_test.go
@@ -10,7 +10,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
-	"slices"
 	"strconv"
 	"strings"
 	"syscall"
@@ -997,15 +996,12 @@ func TestCgroupDeviceRules(t *testing.T) {
 	executor := execInterface.(*LibcontainerExecutor)
 	cfg, err := executor.newLibcontainerConfig(command)
 	must.NoError(t, err)
-	rules := make([]devices.Rule, len(cfg.Cgroups.Devices))
-	for i, rule := range cfg.Cgroups.Devices {
-		rules[i] = *rule
-	}
-	must.True(t, slices.Contains(rules, devices.Rule{
+
+	must.SliceContains(t, cfg.Cgroups.Devices, &devices.Rule{
 		Type:        'c',
-		Major:       0xa,
+		Major:       0x0a,
 		Minor:       0xe5,
 		Permissions: "rwm",
 		Allow:       true,
-	}))
+	})
 }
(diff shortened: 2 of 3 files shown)


lukepalmer:
Hi @tgross, I'm new here and I see you're pretty active- would you have guidance on how to get a PR reviewed?

tgross:
Hi @lukepalmer I'm marked this PR for triage, but honestly we can't really review this until we've made the CLA bot happy. I think you might have pushed a commit with a different associated email.

lukepalmer:
Thanks for looking @tgross. There is a comment that syas "All commmitters have signed the CLA" and there's only one commit (but I did force push). I can close and re-create the PR if that helps, let me know.

Hopefully the bug and fix are pretty straightforward once we get to that.

drivers/shared/executor/executor_linux.go
```diff
@@ -677,6 +677,10 @@ func configureIsolation(cfg *runc.Config, command *ExecCommand) error {
 		cfg.Devices = append(cfg.Devices, devs...)
 	}
 
+	for _, device := range cfg.Devices {
+		cfg.Cgroups.Resources.Devices = append(cfg.Cgroups.Resources.Devices, &device.Rule)
+	}
```
tgross, line 682: We've moved this so that it's happening after we've added the user-specified devices (from the [`device`]() block) to the list of default devices from libcontainer (ref [`spec_linux.go#L177-L320`](https://github.com/opencontainers/runc/blob/v1.1.12/libcontainer/specconv/spec_linux.go#L177-L320)). That seems right but not directly motivated by the example... unless you've got a `/dev/fuse` device driver?

drivers/shared/executor/executor_linux.go
```diff
@@ -897,6 +897,7 @@ func cmdDevices(driverDevices []*drivers.DeviceConfig) ([]*devices.Device, error
 			return nil, fmt.Errorf("failed to make device out for %s: %v", d.HostPath, err)
 		}
 		ed.Path = d.TaskPath
+		ed.Allow = true // rules will be used to allow devices via cgroups
```
tgross, line 900: From the libcontainer docs:

```go
	// The device nodes that should be automatically created within the container upon container start.  Note, make sure that the node is marked as allowed in the cgroup as well!
	Devices []*devices.Device `json:"devices"`
```

Oof! This was there telling us about this issue all along!

tgross (commented):
Hi @lukepalmer! I've made a pass through this and it looks good to me as far as the cgroup setup. But it's not obvious to me how this solves your motivating problem of mounting a device like `/dev/fuse`. Are you using a custom device plugin?

Two more items:
* Can you create a changelog entry? Use `make cl` to run a little tool for the right formatting.
* "a backport to 1.7 would be appreciated if that's possible" just a heads up that Nomad now has a [Long-Term Support (LTS)](https://developer.hashicorp.com/nomad/docs/enterprise#long-term-support-lts-releases) policy, so this will get backported to 1.7.x+ent and 1.6.x+ent, but only to the current version of CE (1.8.x).

(cc'ing @shoenig too as I suspect he'll have thoughts on this for the exec2 driver)

drivers/shared/executor/executor_linux.go
lukepalmer: I'm not entirely sure I've understood your question but I'll take a shot.

To have /dev/fuse usable inside a container, assuming the device is already present in the host, you can add an entry to the 'Devices' field in the executor configuration.

https://github.com/hashicorp/nomad/blob/main/drivers/shared/executor/executor.go#L141

With an entry there (and this PR) you will then have a usable /dev/fuse in the container. It's then possible to use fusermount to mount a filesystem via fuse.

An application where this matters is for [python XARs](https://github.com/facebookincubator/xar). Fuse mounting the XAR makes a huge performance difference when the archive is large and on a networked filesystem.

We are in fact using a custom driver plugin, but it's little more than a thin wrapper around the exec driver. It doesn't do anything special related to devices aside from populating the 'Devices' field.

Let me know if I've missed what you're after.

lukepalmer: Also added a changelog entry as requested. Thanks for having a look!

drivers/shared/executor/executor_linux_test.go
```diff
@@ -973,3 +975,37 @@ func TestExecutor_SignalCatching(t *testing.T) {
...
+	for i, rule := range cfg.Cgroups.Devices {
+		rules[i] = *rule
+	}
+	must.True(t, slices.Contains(rules, devices.Rule{
+		Type:        'c',
+		Major:       0xa,
+		Minor:       0xe5,
+		Permissions: "rwm",
+		Allow:       true,
+	}))
```
tgross, line 1010: We can trim this down a good bit:

```suggestion
	must.SliceContains(t, cfg.Cgroups.Devices, &devices.Rule{
		Type:        'c',
		Major:       0xa,
		Minor:       0xe5,
		Permissions: "rwm",
		Allow:       true,
	})
```

tgross (changes requested):
I've left a couple more comments but once those are resolved I think this is good-to-go. Thanks @lukepalmer!

drivers/shared/executor/executor_linux_test.go
lukepalmer: Let me try this and get back to you. When initially writing this I did not get 'must.SliceContains' to behave in a way that made sense.

tgross (approved):
LGTM! Thanks @lukepalmer!

PR #22912, opened 2024-06-03
metrics: quota utilization configuration and documentation

pkazmierczak:
CE part of the https://github.com/hashicorp/nomad-enterprise/pull/1488 change.


command/agent/config.go
```diff
@@ -981,6 +981,10 @@ type Telemetry struct {
 	// a small memory overhead.
 	DisableDispatchedJobSummaryMetrics bool `hcl:"disable_dispatched_job_summary_metrics"`
 
+	// DisableQuotaUtilizationMetrics allows to disable publishing of quota
+	// utilization metrics
+	DisableQuotaUtilizationMetrics bool `hcl:"disable_quota_utilization_metrics"`
```
tgross, line 986: I think we need to add this to the telemetry config docs?

tgross (approved):
LGTM!

PR #23166, opened 2024-06-04
testing: fix the value of NOMAD_SECRETS_DIR in test harness

shoenig:
This PR fixes the value of NOMAD_SECRETS_DIR to be the alloc_mounts
secrets directory instead of the real secrets directory, which is protected
by root 0700 even when running tests.

Needed for https://github.com/hashicorp/nomad-driver-exec2/issues/29


tgross (approved):
LGTM :+1:

PR #23169, opened 2024-06-04
scheduler: stack test should use job.ID and not job.Name

tgross:
Some of our scheduler tests use the `AllocName` function from the structs package incorrectly. This function should always receive the `Job.ID` and not the `Job.Name`. Fix this to prevent future bugs from copy-pasting usage around.


gulducat (approved):
:id: :+1:

PR #23171, opened 2024-06-04
[actions] Update hashicorp/vault-action to v3.0.0

angrycub:
Removes a dependency on a node 16 action, which are EOLed.


tgross (approved):
LGTM!

PR #23185, opened 2024-06-05
docs: fix typo in nomad quota utilization metrics

pkazmierczak:
missing `nomad` prefix


tgross (approved):
LGTM!

That we named the server package `nomad` is such a :facepalm: for these metrics :grinning:

pkazmierczak:
yeah... caught it only now while preparing a demo...

PR #23189, opened 2024-06-05
build: remove 32-bit ARM builds

tgross:
We no longer intend to release 32-bit builds for any platform. We'd previously removed the builds for i386 on both Linux and Windows, but never got around to removing the ARM builds. Add a note about this deprecation in the release notes for 1.8.x.

Closes: https://github.com/hashicorp/nomad-enterprise/issues/1053

### DIFF
--- a/website/content/docs/release-notes/nomad/v1_8_x.mdx
+++ b/website/content/docs/release-notes/nomad/v1_8_x.mdx
@@ -107,8 +107,9 @@ We are pleased to announce the following Nomad updates.
   introduces new options for allocations reconciliation if the client regains
   connectivity.
 
-- **32-bit builds**: Nomad releases will no longer include binaries built for
-  32-bit platforms.
+- **32-bit builds**: Support for 32-bit architectures is now deprecated. Nomad
+  releases will no longer include binaries built for 32-bit platforms starting
+  with Nomad 1.9.
 
 ## What's removed
 
--- a/website/content/docs/release-notes/nomad/v1_8_x.mdx
+++ b/website/content/docs/release-notes/nomad/v1_8_x.mdx
@@ -11,10 +11,11 @@ We are pleased to announce the following Nomad updates.
 
 ## Release highlights
 
-- **Long Term Support Release (Enterprise):** Nomad 1.8 introduces a [long
-  term support release](/nomad/docs/enterprise#long-term-support-lts-releases) which provides extended maintenance (two years of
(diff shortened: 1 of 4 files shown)


david-yu (approved):

schmichael (approved):
Thanks Tim!

:wave: For anyone ending up here because you're looking for 32bit ARM binaries I would encourage you to try building Nomad from source for your particular architecture. The 32 bit ARM landscape is extremely diverse and was never well served by the binary we did build. We strive to keep Nomad as easy as a `go install` to build yourself, so I encourage you read the Go toolchain's page on ARM to find the best settings for your target architecture: https://go.dev/wiki/GoArm

shoenig:
I think we should delete `build_linux_arm.go` as well

website/content/docs/release-notes/nomad/v1_8_x.mdx
```diff
@@ -87,30 +91,32 @@ We are pleased to announce the following Nomad updates.
...
-  
+  `prevent_reschedule_on_lost`:** Nomad 1.8.0 introduces a
+  [disconnect](/nomad/docs/job-specification/disconnect) block meant to group
+  all the configuration options related to disconnected client's and server's
+  behavior, causing the deprecation of the fields stop_after_client_disconnect,
+  max_client_disconnect and prevent_reschedule_on_lost. This block also
+  introduces new options for allocations reconciliation if the client regains
+  connectivity.
+
+- **32-bit builds**: Nomad releases will no longer include binaries built for
```
david-yu, line 110:
```suggestion
- **32-bit builds**: Support for 32-bit architecture is now deprecated. Nomad will no longer include binaries built for
```

website/content/docs/release-notes/nomad/v1_8_x.mdx
```diff
@@ -87,30 +91,32 @@ We are pleased to announce the following Nomad updates.
...
+  32-bit platforms.
```
david-yu, line 111:
```suggestion
  32-bit platforms starting with Nomad 1.9. 
```

tgross:
> I think we should delete build_linux_arm.go as well

Should we, given the note @schmichael left above? Or as we move away from this support is that likely to just be a red herring for anyone trying to figure out a `go build` workflow?

tgross:
I probably should have split this into 2 PRs, one for the docs and one for the build changes. I'll open separate PRs with just the docs for backporting to 1.8.x/1.8.x+ent

Issue #23174, opened 2024-06-04
testing: move any Docker image pulls off the Docker registry


tgross:
We're getting rate-limited on test runs: https://github.com/hashicorp/nomad/actions/runs/9373326002/job/25806862514?pr=23172 We should think about pulling these images from elsewhere.

```
    driver_test.go:146: 
        	Error Trace:	/home/runner/work/nomad/nomad/drivers/docker/driver_test.go:146
        	            				/home/runner/work/nomad/nomad/drivers/docker/driver_unix_test.go:680
        	Error:      	Received unexpected error:
        	            	rpc error: code = FailedPrecondition desc = Failed to pull `busybox:1.29.2`: API error (500): toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
        	Test:       	TestDockerDriver_Cleanup
```

gulducat:
there are a couple helper utils for this here: https://github.com/hashicorp/nomad/blob/v1.8.0/testutil/container_images.go

PR #23190, opened 2024-06-05
testing: pull Docker images from mirror

tgross:
In https://github.com/hashicorp/nomad/pull/17401 we added test helpers that would allow `docker` driver tests to pull from a mirror of the Docker Hub registry. Extend the use of this helper to tests that recently hit rate-limiting.

Fixes: https://github.com/hashicorp/nomad/issues/23174


gulducat (approved):
lgtm!

Issue #11087, opened 2021-08-26
Expose oom_score_adj on a per-task basis


schmichael:
### Problem
Nomad 1.1 adding [Memory Oversubscription](https://www.nomadproject.io/docs/job-specification/resources#memory-oversubscription) which allows tasks to set a soft and hard limit for their memory usage. When a system's memory is exhausted and a process needs to be oom killed, the higher a process is over its soft limit, the more likely it is to be killed.

However, this logic does not take into account the priorities of the application being run. Just because a task is using more memory does not make it "less valuable" to be running.

#### Proposal: resources.oom_score_adj

Add an `oom_score_adj` field to [`resources`](https://www.nomadproject.io/docs/job-specification/resources) that *only accepts positive values.* A task can mark itself as *more likely* to be OOM killed, but not less likely.

#### Abandoned Idea: Use Priority

It's tempting to make `job.priority` implicitly impact `oom_score_adj` since it is already used by preemption to stop lower priority jobs.

However, this impacts all tasks for a job. You may want a flaky log shipping sidecar to get killed (and restarted) before _anything_ else on a system.

So while implicitly using `priority` might be valuable in the future, allowing users to explicitly demote some tasks is a necessary first step.

pkazmierczak:
With updates to [raw_exec](https://github.com/hashicorp/nomad/pull/23308), [exec2](https://github.com/hashicorp/nomad-driver-exec2/pull/40) and [docker](https://github.com/hashicorp/nomad/pull/23297), I consider this issue closed for now. Should we want to revisit putting this in the `resources` block, we have https://github.com/hashicorp/nomad/pull/23259 as a good starting point.

PR #23259, opened 2024-06-07
resources: new oom_score_adj field

pkazmierczak:
Adds an `oom_score_adj` field to [resources](https://www.nomadproject.io/docs/job-specification/resources) that only accepts positive values. A task can mark itself as more likely to be OOM killed, but not less likely.

Resolves https://github.com/hashicorp/nomad/issues/11087

### DIFF
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -2424,6 +2424,7 @@ type Resources struct {
 	Networks    Networks
 	Devices     ResourceDevices
 	NUMA        *NUMA
+	OOMScoreAdj int
 }
 
 const (
@@ -2436,9 +2437,10 @@ const (
 // be kept in sync.
 func DefaultResources() *Resources {
 	return &Resources{
-		CPU:      100,
-		Cores:    0,
-		MemoryMB: 300,
+		CPU:         100,
+		Cores:       0,
+		MemoryMB:    300,
+		OOMScoreAdj: 0,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


nomad/structs/structs.go
```diff
@@ -2494,6 +2497,10 @@ func (r *Resources) Validate() error {
 		mErr.Errors = append(mErr.Errors, fmt.Errorf("MemoryMaxMB value (%d) should be larger than MemoryMB value (%d)", r.MemoryMaxMB, r.MemoryMB))
 	}
 
+	if r.OOMScoreAdj < 0 {
+		mErr.Errors = append(mErr.Errors, fmt.Errorf("OOMScoreAdj value (%d) must be a positive integer", r.OOMScoreAdj))
+	}
```
tgross, line 2502: The error message doesn't match the validation, because 0 is allowed.

.changelog/23259.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+resources: Added new oom_score_adj field
+```
```
tgross, line 3: We haven't wired this up to the drivers yet, so if we merge this and end up shipping Nomad 1.8.1 before that work is complete, the changelog entry won't make much sense. Maybe we should hold off adding the changelog entry until all the components (and documentation!) are in place? That way we can safely ship this code in `main` without stressing about 1.8.1.

tgross (approved):
LGTM!

.changelog/23259.txt
pkazmierczak: i can also remove the backport labels for now.

Issue #23270, opened 2024-06-07
[Documentation] Documented example regarding ACL plugin fails with error failed to parse rules: Invalid plugin policy


EugenKon:
### Version
```
Nomad v1.8.0
BuildDate 2024-05-28T17:38:17Z
Revision 28b82e4b2259fae5a62e2ed47395334bea5a24c4
```

### Issue

Having this configuration:
https://developer.hashicorp.com/nomad/docs/other-specifications/acl-policy
![image](https://github.com/hashicorp/nomad/assets/139780495/3e097647-d799-4f01-88e4-3a5d3ee80748)

Fails with error message:
```
+ nomad acl policy apply -token 50d0a579-771a-015c-9b3d-f5b60085c5f0 -description 'Policy to allow reading of agents and nodes and listing and submitting jobs in all namespaces.' node-read-job-submit /ops/shared/config/acl/nomad-acl-user.hcl
Error writing ACL policy: Unexpected response code: 400 (policy 0 invalid: 1 error occurred:
	* failed to parse rules: Invalid plugin policy: &acl.PluginPolicy{Policy:"write"})
```

### Reproduction steps
Try to bootstrap Nomad ACL using given example

#### Expected Result
No error should occur

#### Actual Result
Error happened

jrasell:
Hi @EugenKon and thanks for raising this issue. I believe the problem lies with the validation logic, so I have raised a PR to fix that.

PR #23274, opened 2024-06-10
acl: fix validation of ACL plugin policy entries.

jrasell:


tgross (approved):
LGTM

tgross:
This fixes the bug, but it occurs to me that the example in the docs being pointed out in #23270 is really dubious. There's not really a good reason to give a human user `plugin:write`, because no HTTP API call for plugins requires `plugin:write` (they're all read-only endpoints)!

jrasell:
> This fixes the bug, but it occurs to me that the example in the docs being pointed out in #23270 is really dubious. There's not really a good reason to give a human user `plugin:write`, because no HTTP API call for plugins requires `plugin:write` (they're all read-only endpoints)!

Raised https://github.com/hashicorp/nomad/pull/23277 to fix the docs. It's a separate PR, so it's easier to manage the backporting code vs. docs.

PR #23284, opened 2024-06-10
numa: fix scheduler panic due to topology serialization bug

tgross:
The NUMA topology struct field `NodeIDs` is a `idset.Set`, which has no public members. As a result, this field is never serialized via msgpack and persisted in state. When `numa.affinity = "prefer"`, the scheduler dereferences this nil field and panics the scheduler worker.

Ideally we would fix this by adding a msgpack serialization extension, but because the field already exists and is just always empty, this breaks RPC wire compatibility across upgrades. Instead, create a new field that's populated at the same time we populate the more useful `idset.Set`, and repopulate the set on demand.

Fixes: https://hashicorp.atlassian.net/browse/NET-9924
Ref: https://github.com/hashicorp/nomad-enterprise/pull/1527

### DIFF
--- a/client/fingerprint/cpu.go
+++ b/client/fingerprint/cpu.go
@@ -173,7 +173,7 @@ func (f *CPUFingerprint) setNUMA(response *FingerprintResponse) {
 		return
 	}
 
-	nodes := f.top.Nodes()
+	nodes := f.top.GetNodes()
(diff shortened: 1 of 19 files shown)


pkazmierczak (approved):
LGTM!

PR #23287, opened 2024-06-10
e2e: add prerelease builds to Consul/Vault compatibility tests

tgross:
Update the Consul/Vault build downloader functions so that we include the current prerelease build (if any) in our E2E compatibility testing we do on each PR. This will automatically cycle out when the GA build is released, because that build is "higher" in the sorted set.


jrasell (approved):
LGTM!

PR #23290, opened 2024-06-10
[ui] Rework of node/job attributes/meta using PathTree

philrenaud:
This extends and reuses PathTree logic originally written for Variables in https://github.com/hashicorp/nomad/pull/13202

This PR specifically fixes a bug where a metadata k/v item wouldn't show up in the UI if its full key was the prefix of a nested child key.

For example, this metadata:
```
{
  'phil.hair': 'yes',
  'phil.hair.colour': 'black',
  'phil.hair.length': 'short'
};
```

would produce the following:
image

and now produces:
image

### DIFF
--- a/ui/app/templates/clients/client/index.hbs
+++ b/ui/app/templates/clients/client/index.hbs
@@ -793,7 +793,7 @@
               {{capitalize a.item.name}}
               Attributes
             </div>
-            {{#if a.item.attributes.structured}}
+            {{#if a.item.attributesShort}}
               <div class="boxed-section-body is-full-bleed">
                 <AttributesTable
                   @attributePairs={{a.item.attributesShort}}
@@ -821,7 +821,7 @@
     <div class="boxed-section-body is-full-bleed">
       <AttributesTable
         data-test-attributes
-        @attributePairs={{this.model.attributes.structured}}
+        @attributePairs={{this.model.attributes.structured.root}}
         @class="attributes-table"
         @copyable={{true}}
       />
@@ -835,7 +835,7 @@
       <div class="boxed-section-body is-full-bleed">
         <AttributesTable
           data-test-meta
-          @attributePairs={{this.model.meta.structured}}
+          @attributePairs={{this.model.meta.structured.root}}
           @editable={{can "write client"}}
           @onKVSave={{this.addDynamicMetaData}}
           @onKVEdit={{this.validateMetadata}}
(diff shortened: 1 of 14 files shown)


ui/app/templates/clients/client/index.hbs
```diff
@@ -793,7 +793,7 @@
               {{capitalize a.item.name}}
               Attributes
             </div>
-            {{#if a.item.attributes.structured}}
+            {{#if a.item.attributesShort}}
```
philrenaud, line 796: The conditional here now matches the property we use, rather than being a proxy for it.

ui/app/templates/components/attributes-section.hbs
```diff
@@ -3,17 +3,23 @@
...
+{{/each}}
+{{#each-in this.attributes.children as |key value|}}
+  <AttributesSection
+    @attributes={{value}}
+    @key={{key}}
+    @value={{value}}
+    @editable={{@editable}}
+    @onKVSave={{@onKVSave}}
+    @copyable={{@copyable}}
+  />
```
philrenaud, line 24: Two notable changes here:
- We no longer show the "key-prefix-only" blue headers like we used to. These were `colspan="2"`/full-width rows that both broke table accessibility conventions and made the page taller than it needed to be (especially with a lot of client metadata)
- we place "terminal" items first, and then nested ones below, kind of like a filesystem layout might look in an explorer window, rather than alphabetizing all prefixes. I am less attached to this change but it feels right so far.

ui/app/utils/path-tree.js
```diff
@@ -39,11 +39,14 @@ export default class PathTree {
   /**
    * @param {MutableArray<VariableModel>} variables
    */
-  constructor(variables) {
+  constructor(variables, { delimiter = '/' } = {}) {
```
philrenaud, line 42: I stopped short of fully generifying this (changing `.files[]` to `.terminal[]` etc.) but did:
- parameterize the delimiter
- add `prefix` as a returned field for ease of use in the UI

ui/tests/integration/components/attributes-table-test.js
```diff
@@ -81,27 +92,4 @@ module('Integration | Component | attributes table', function (hooks) {
       'properties'
     );
   });
-
-  test('should render a row for key/value pairs even when the value is another object', async function (assert) {
```
philrenaud, line 85: Test became obsolete when we stopped showing the blue "key" rows.

philrenaud:
@tgross your comment got me to spitball with some components a bit; what would you think about the look of metadata/attributes if they were set in our `Badge` components?
image

Is this too visually heavy for what it is?

tgross:
> Is this too visually heavy for what it is?

Yeah, feels heavy to me.

PR #23297, opened 2024-06-11
docker: oom_score_adj support

pkazmierczak:
See https://github.com/hashicorp/nomad/pull/23259 and https://github.com/hashicorp/nomad/issues/11087 for context.

### DIFF
--- a/drivers/docker/driver.go
+++ b/drivers/docker/driver.go
@@ -990,8 +990,9 @@ func (d *Driver) createContainerConfig(task *drivers.TaskConfig, driverConfig *T
 	hostConfig := &docker.HostConfig{
 		// do not set cgroup parent anymore
 
-		Memory:            memory,            // hard limit
-		MemoryReservation: memoryReservation, // soft limit
+		Memory:            memory,                   // hard limit
+		MemoryReservation: memoryReservation,        // soft limit
+		OomScoreAdj:       driverConfig.OOMScoreAdj, // ignored on platforms other than linux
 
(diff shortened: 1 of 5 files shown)


tgross (approved):
LGTM!

drivers/docker/driver.go
```diff
@@ -1053,6 +1053,11 @@ func (d *Driver) createContainerConfig(task *drivers.TaskConfig, driverConfig *T
 		}
 	}
 
+	// oom_score_adj is only supported on linux
+	if runtime.GOOS == "linux" {
```
schmichael, line 1057: Is it possible the host (runtime.GOOS)  is running macOS or Windows while the container is being started in a Docker VM?

If we do need to gate this by OS then let's also at least log a `WARN` if someone is trying to use this field on an unsupported OS. Seems like a mistake, but not a big enough mistake to crash it over. Something like:

```go
logger.Warn("oom_score_adj is only supported on Linux; ignoring", "os", runtime.GOOS, "oom_score_adj", oom_score_adj)
```

schmichael (approved):

drivers/docker/driver.go
pkazmierczak: > Is it possible the host (runtime.GOOS) is running macOS or Windows while the container is being started in a Docker VM?

right. I changed the code to just always include the config option, since docker will simply ignore it on hosts that don't support it. And we mention in the documentation that this is only supported on Linux hosts, so I think we're fine.

PR #23306, opened 2024-06-11
[ui] Remove simulacral allocation stat in favor of live-updating one

### LINKED ISSUE #20627: Current Job Overview Summary page is reporting disparate numbers for running allocations (segal's law)
### Nomad version
technically it's 1.6.9, but nothing about this was mentioned in changelogs about this bug

N/A

### Issue

The Current Job Overview Summary page has incongruent numbers.

Smart aleck wording: 376 != 297

### Reproduction steps

I go to `$NOMAD_ADDR/ui/jobs/JOB-NAME@default` and see both numbers for "Running" do not match on a system job currently being deployed. Thanks to Segal's Law ("A man with one watch always knows what time it is. A man with two watches is never sure."), I now have no idea how many Allocations are Running.

#### Expected Result

I see that I have 376 Allocations Running. Oh wait, I have 297 Running. 

#### Actual Result
image

philrenaud:
While most of our job status panel allocation numbers use allocations live as they're updated from the /allocations endpoint for a given job, one holdout remained: the the "X Allocations Running" number came from the `job-summary`, which tabulates at a slightly different time than allocations do (and is sometimes susceptible to inaccuracy)

This changes the number in steady-state jobs to be what we actually see among currently running allocations from the /allocations endpoint. This doesn't discriminate between new-version and old-version allocations, only returning the number currently running. 

image

Resolves #20627

### DIFF
--- a/ui/app/components/job-status/panel/steady.js
+++ b/ui/app/components/job-status/panel/steady.js
@@ -185,6 +185,10 @@ export default class JobStatusPanelSteadyComponent extends Component {
     return this.job.allocations.filter((a) => !a.isOld && a.hasBeenRestarted);
   }
 
+  get runningAllocs() {
+    return this.job.allocations.filter((a) => a.clientStatus === 'running');
+  }
+
   get completedAllocs() {
     return this.job.allocations.filter(
       (a) => !a.isOld && a.clientStatus === 'complete'
--- a/.changelog/23306.txt
+++ b/.changelog/23306.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: unbind job detail running allocations count from job-summary endpoint
+```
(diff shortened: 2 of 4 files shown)


ui/app/components/job-status/panel/steady.js
```diff
@@ -185,6 +185,10 @@ export default class JobStatusPanelSteadyComponent extends Component {
     return this.job.allocations.filter((a) => !a.isOld && a.hasBeenRestarted);
   }
 
+  get runningAllocs() {
+    return this.job.allocations.filter((a) => a.clientStatus === 'running');
+  }
```
philrenaud, line 190: `@job.runningAllocs` comes from the job's summary as defined [here](https://github.com/hashicorp/nomad/blob/86ee56b8c58b543a245a83bf8776eacebe297c9d/ui/app/models/job.js#L406) in case you were curious

shoenig (approved):

PR #23308, opened 2024-06-12
raw_exec: oom_score_adj support

pkazmierczak:
See https://github.com/hashicorp/nomad/pull/23259 and https://github.com/hashicorp/nomad/issues/11087 for context.


website/content/docs/drivers/raw_exec.mdx
```diff
@@ -50,6 +50,9 @@ The `raw_exec` driver supports the following configuration in the job spec:
 ~> The `task.user` field cannot be set on a Task using the `raw_exec` driver if
 the Nomad client has been hardened according to the [production][hardening] guide.
 
+- `oom_score_adj` - (Optional) A positive integer to indicate the likelihood of
+  the task being OOM killed (valid only for Linux). Defaults to 0.
```
tgross, line 54: I missed this on the Docker PR, but we don't actually enforce that it's positive in the driver code, right?

tgross (approved):
LGTM!

shoenig (approved):

website/content/docs/drivers/raw_exec.mdx
pkazmierczak: Right! And also there's a mistake both here and in the docker driver docs: `oom_score_adj` is not set in the driver options, but as a task config option. I fix.

PR #23319, opened 2024-06-13
security: add escape to arbitrary file access

dduzgun-security:
Resolves `Arbitrary file write extracting an archive containing symbolic links` alerts by using the escape function.


client/allocwatcher/alloc_watcher.go
```diff
@@ -587,6 +587,14 @@ func (p *remotePrevAlloc) streamAllocDir(ctx context.Context, resp io.ReadCloser
 				p.prevAllocID, p.allocID, err)
 		}
 
+		escapes, err := escapingfs.PathEscapesAllocDir(dest, "", hdr.Name)
+		if err != nil {
+			return fmt.Errorf("error evaluating path: %w", err)
+		}
+		if escapes {
+			return fmt.Errorf("archive contains file that escapes alloc dir")
+		}
```
tgross, line 596: How would a filesystem object escape the alloc dir if it's not a symlink?

dduzgun-security: I was thinking in case a header `hdr` contains a relative path name or linkname

tgross: > I was thinking in case a header `hdr` contains a relative path name or linkname

Ah, true. In practice the `hdr.Name` is always going to be relative because that's how the other end creates the archive. But this would add some protection against a compromised client that returns an evil archive instead.

tgross (commented):
@dduzgun-security this looks good. It'd be nice if we could add some test coverage to these checks, similar to what we have in `TestPrevAlloc_StreamAllocDir_Error`.

dduzgun-security:
Looks like other tests broke, working to fix them.
```
=== Failed
=== FAIL: client/allocwatcher TestPrevAlloc_StreamAllocDir_BadSymlink (0.00s)
    assert.go:14: 
        alloc_watcher_unix_test.go:115: expected error; got nil

=== FAIL: client/allocwatcher TestPrevAlloc_StreamAllocDir_BadSymlink (re-run 1) (0.00s)
    assert.go:14: 
        alloc_watcher_unix_test.go:115: expected error; got nil

=== FAIL: client/allocwatcher TestPrevAlloc_StreamAllocDir_BadSymlink (re-run 2) (0.00s)
    assert.go:14: 
        alloc_watcher_unix_test.go:115: expected error; got nil

=== FAIL: client/allocwatcher TestPrevAlloc_StreamAllocDir_BadSymlink (re-run 3) (0.00s)
    assert.go:14: 
        alloc_watcher_unix_test.go:115: expected error; got nil
```

dduzgun-security:
🤔 hmm, running the tests locally works but the CI seems to fail, do we have some sort of caching that may cause that?
Or maybe I may be missing something

Local 👇 
```
Running tool: /opt/homebrew/bin/go test -timeout 30s -run ^TestPrevAlloc_StreamAllocDir_BadSymlink$ github.com/hashicorp/nomad/client/allocwatcher

=== RUN   TestPrevAlloc_StreamAllocDir_BadSymlink
=== PAUSE TestPrevAlloc_StreamAllocDir_BadSymlink
=== CONT  TestPrevAlloc_StreamAllocDir_BadSymlink
2024-06-17T10:05:52.743-0400 [DEBUG] .../hashicorp/nomad/client/allocwatcher/allocwatcher/alloc_watcher.go:555: streaming snapshot of previous alloc: destination=/var/folders/th/2vhwms5n1qbf8rgl7cy7902m0000gn/T/TestPrevAlloc_StreamAllocDir_BadSymlink796392596/001
--- PASS: TestPrevAlloc_StreamAllocDir_BadSymlink (0.00s)
PASS
ok      github.com/hashicorp/nomad/client/allocwatcher  0.342s
```

tgross:
@dduzgun-security this is what I'm getting locally with this branch checked out:

```
$ go test -count=1 -run ^TestPrevAlloc_StreamAllocDir_BadSymlink$ github.com/hashicorp/nomad/client/allocwatcher
2024-06-17T12:31:56.462-0400 [DEBUG] allocwatcher/alloc_watcher.go:555: streaming snapshot of previous alloc: destination=/tmp/TestPrevAlloc_StreamAllocDir_BadSymlink4141846833/001
--- FAIL: TestPrevAlloc_StreamAllocDir_BadSymlink (0.00s)
    assert.go:14:
        alloc_watcher_unix_test.go:114: expected error; got nil
FAIL
FAIL    github.com/hashicorp/nomad/client/allocwatcher  0.018s
FAIL
```

Looks like you're testing on macOS though and I'm testing on Linux (as is CI); might be worth checking that there isn't OS-specific behavior here.

dduzgun-security:
When I tried it on an ubuntu VM, the test passed
image

Maybe something related with the GHA runner 🤔 I'll give it a shot with this container image
https://github.com/actions-runner-controller/actions-runner-controller/pkgs/container/actions-runner-controller%2Factions-runner-dind/226924128?tag=ubuntu-22.04

.changelog/23319.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:security
+core: Adding escape to arbitrary file access on streamAllocDir
```
tgross, line 2: The changelog should be user facing, whereas this is more describing the internals we changed. We should instead describe what security flaw we fixed. Or, in cases where we're just silencing false positive alerts with no user-visible behavior, we should omit the changelog.

So something like this:

```suggestion
migration: Added a check for relative paths escaping the allocation directory when unpacking archive during migration, to harden clients against compromised peer clients sending malicious archives
```

(If I understand the issue we're resolving here.)

client/allocwatcher/alloc_watcher.go
```diff
@@ -587,6 +587,12 @@ func (p *remotePrevAlloc) streamAllocDir(ctx context.Context, resp io.ReadCloser
 				p.prevAllocID, p.allocID, err)
 		}
 
+		if escapes, err := escapingfs.PathEscapesAllocDir(dest, "", hdr.Name); err != nil {
+			return fmt.Errorf("error evaluating directory: %w", err)
+		} else if escapes {
+			return fmt.Errorf("archive contains directory that escapes alloc dir")
+		}
```
tgross, line 594: The error message we're returning here says "directory", but at the point we're checking this it could be any kind of archive object, right?

client/allocwatcher/alloc_watcher.go
```diff
@@ -618,12 +624,12 @@ func (p *remotePrevAlloc) streamAllocDir(ctx context.Context, resp io.ReadCloser
 				return fmt.Errorf("error creating symlink: %w", err)
 			}
 
-			escapes, err := escapingfs.PathEscapesAllocDir(dest, "", hdr.Name)
-			if err != nil {
-				return fmt.Errorf("error evaluating symlink: %w", err)
-			}
-			if escapes {
-				return fmt.Errorf("archive contains symlink that escapes alloc dir")
+			for _, path := range []string{hdr.Name, hdr.Linkname} {
```
tgross, line 627: If we're going to expand this check to cover checking the `hdr.Linkname`, we should cover that behavior with an additional test case on `TestPrevAlloc_StreamAllocDir_BadSymlink`. That should also make it clear why we've added this check, and why the check of `hdr.Name` at the top of the loop doesn't cover us.

tgross (changes requested):
This is looking good, @dduzgun-security. Just a few more issues to get it ready to ship.

client/allocwatcher/alloc_watcher.go
dduzgun-security: That's true nice catch, I edited the error message to specify `object` instead of `directory`

client/allocwatcher/alloc_watcher.go
dduzgun-security: Good point, I added `TestPrevAlloc_StreamAllocDir_BadSymlink_Linkname` to cover the Linkname based escape test https://github.com/hashicorp/nomad/commit/8fe431da2f1b1646509fa1d94b85ab4416473945

client/allocwatcher/alloc_watcher_test.go
```diff
@@ -276,3 +276,40 @@ func TestPrevAlloc_StreamAllocDir_Error(t *testing.T) {
...
+	_, err = tw.Write([]byte{'a'})
+	require.NoError(t, err)
+	err = tw.Close()
+	require.NoError(t, err)
+
+	// Attempt to stream the allocation directory
+	dest := t.TempDir()
+	err = prevAlloc.streamAllocDir(context.Background(), io.NopCloser(tarBuf), dest)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "archive contains object that escapes alloc dir")
```
tgross, line 314: We're (very slowly) migrating over to use `shoenig/test` in the Nomad codebase, which has a more robustly-type notion of equality checking than `testify` does. I've also move the `tw.Close` call into test cleanup, as we want it to get closed unconditionally and we generally don't want to make test assertions on stdlib behavior that isn't interesting to the code under test:

```suggestion
	must.NoError(t, err)
	_, err = tw.Write([]byte{'a'})
	t.Cleanup(tw.Close)
	must.NoError(t, err)

	// Attempt to stream the allocation directory
	dest := t.TempDir()
	err = prevAlloc.streamAllocDir(context.Background(), io.NopCloser(tarBuf), dest)
	must.EqError(t, err, "archive contains object that escapes alloc dir")
```

client/allocwatcher/alloc_watcher_unix_test.go
```diff
@@ -115,6 +115,28 @@ func TestPrevAlloc_StreamAllocDir_BadSymlink(t *testing.T) {
...
+		Mode:     0600,
+	}))
+	must.NoError(t, tw.Close())
+
+	newDir := t.TempDir()
+	prevAlloc := &remotePrevAlloc{logger: testlog.HCLogger(t)}
+	err := prevAlloc.streamAllocDir(context.Background(), io.NopCloser(&buf), newDir)
+
+	must.Error(t, err)
+	must.EqError(t, err, "archive contains symlink that escapes alloc dir")
```
tgross, line 137: `EqError` nil checks for us:

```suggestion
	must.EqError(t, err, "archive contains symlink that escapes alloc dir")
```

client/allocwatcher/alloc_watcher_unix_test.go
```diff
@@ -115,6 +115,28 @@ func TestPrevAlloc_StreamAllocDir_BadSymlink(t *testing.T) {
...
+	// Create a tar archive with a symlink that attempts to escape the allocation directory
+	var buf bytes.Buffer
+	tw := tar.NewWriter(&buf)
+	must.NoError(t, tw.WriteHeader(&tar.Header{
+		Typeflag: tar.TypeSymlink,
+		Name:     "symlink",
+		Linkname: "../escape_attempt",
+		Mode:     0600,
+	}))
+	must.NoError(t, tw.Close())
```
tgross, line 130:
```suggestion
	tw := tar.NewWriter(&buf)
	t.Cleanup(tw.Close)
	must.NoError(t, tw.WriteHeader(&tar.Header{
		Typeflag: tar.TypeSymlink,
		Name:     "symlink",
		Linkname: "../escape_attempt",
		Mode:     0600,
	}))
```

tgross (commented):
Looking good! Just a few minor items to pick up and this should be good-to-merge

client/allocwatcher/alloc_watcher_test.go
```diff
@@ -276,3 +276,38 @@ func TestPrevAlloc_StreamAllocDir_Error(t *testing.T) {
...
+	err := tw.WriteHeader(&tar.Header{
+		Name:     "../escape.txt",
+		Mode:     0666,
+		Size:     1,
+		ModTime:  time.Now(),
+		Typeflag: tar.TypeReg,
+	})
+	must.NoError(t, err)
+	_, err = tw.Write([]byte{'a'})
+	t.Cleanup(tw.Close)
```
tgross, line 306: whoops! 

```suggestion
	t.Cleanup(func() {tw.Close})
```

tgross (approved):
LGTM! Thanks @dduzgun-security!

PR #23328, opened 2024-06-14
[ui] Add "stopped" as a valid status on jobs index/job detail

### LINKED ISSUE #22101: [ui] A service job that stops successfully as intended shows "Failed" instead of "Complete"
We currently reserve "Complete" as a state for batch and sysbatch jobs, but we should be able to extend it to Service jobs that have been deliberately, successfully stopped.

Both on the jobs index page status badge, and on the Status badge on the job detail page:
image

philrenaud:
Our recent changes to how we label jobs in the UI considers any job "failed" if none of its expected allocations are running or pending, and it's not in active deployment.

This is coincidentally both true for jobs that failed because a user clicked the "stop" button, and also for jobs that failed because of the universe conspiring against us generally.

Luckily, jobs already had a nice `Stop` boolean to indicate the manual, deliberate stopping of a job (until it gets garbage collected, and provided it didn't have a "purge" flag passed in).

This PR observes that boolean and displays a new "Stopped" status in the UI in two places:

image
image

resolves #22101

### DIFF
--- a/nomad/job_endpoint_statuses.go
+++ b/nomad/job_endpoint_statuses.go
@@ -219,7 +219,8 @@ func jobStatusesJobFromJob(ws memdb.WatchSet, store *state.StateStore, job *stru
 		GroupCountSum:    0,
 		ChildStatuses:    nil,
 		LatestDeployment: nil,
-		Stop: 				job.Stop,
+		Stop:             job.Stop,
+		Status:           job.Status,
 	}
 
 	// the GroupCountSum will map to how many allocations we expect to run
--- a/nomad/job_endpoint_statuses.go
+++ b/nomad/job_endpoint_statuses.go
@@ -219,6 +219,7 @@ func jobStatusesJobFromJob(ws memdb.WatchSet, store *state.StateStore, job *stru
 		GroupCountSum:    0,
 		ChildStatuses:    nil,
 		LatestDeployment: nil,
+		Stop: 				job.Stop,
 	}
 
 	// the GroupCountSum will map to how many allocations we expect to run
--- a/nomad/structs/job.go
+++ b/nomad/structs/job.go
@@ -97,7 +97,8 @@ type JobStatusesJob struct {
 	// ParentID is set on child (batch) jobs, specifying the parent job ID
 	ParentID         string
 	LatestDeployment *JobStatusesLatestDeployment
-	Stop bool // has the job been manually stopped?
+	Stop             bool // has the job been manually stopped?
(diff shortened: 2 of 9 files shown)


nomad/structs/job.go
```diff
@@ -97,6 +97,7 @@ type JobStatusesJob struct {
 	// ParentID is set on child (batch) jobs, specifying the parent job ID
 	ParentID         string
 	LatestDeployment *JobStatusesLatestDeployment
+	Stop bool // has the job been manually stopped?
```
tgross, line 100: We'll want to add this to the equivalent `api` package struct as well.

ui/app/components/job-status/panel/steady.js
```diff
@@ -217,6 +217,13 @@ export default class JobStatusPanelSteadyComponent extends Component {
     // If all allocs are running, the job is Healthy
     const totalAllocs = this.totalAllocs;
 
+    if (this.job.status === 'dead' && this.job.stopped) {
```
tgross, line 220: I like that we've included both conditions here, because the job status won't be `dead` until the scheduler has had at least once chance at it to stop allocs. Is there a reason not to include that condition in `ui/app/models/job.js`?

nomad/job_endpoint_statuses.go
```diff
@@ -219,6 +219,7 @@ func jobStatusesJobFromJob(ws memdb.WatchSet, store *state.StateStore, job *stru
 		GroupCountSum:    0,
 		ChildStatuses:    nil,
 		LatestDeployment: nil,
+		Stop: 				job.Stop,
```
tgross, line 222: Needs `go fmt -s`

ui/app/components/job-status/panel/steady.js
philrenaud: Great question and I should've explained it in a comment. We don't get job.Status back on the /statuses endpoint (there's a joke here I think) is the only reason; I'll add it now.

nomad/structs/job.go
gulducat: I actually never did add this HTTP API to the `api` package, partly because I hadn't fully committed to the structure of it, long-term, only its use in the UI.  arguably I should go ahead and do it, but at least for now, Phil doesn't need to account for it in this PR.

tgross: Ok. Yeah out of scope for this PR then.

tgross (approved):
LGTM! This is a nice improvement!

Issue #20212, opened 2024-03-25
read init-p: connection reset by peer


wusikijeronii:
### Nomad version
```sh
Nomad v1.7.6
BuildDate 2024-03-12T07:27:36Z
Revision 594fedbfbc4f0e532b65e8a69b28ff9403eb822e
```

### Operating system and Environment details
```sh
NAME="Oracle Linux Server"
VERSION="8.9"
```
### Issue
After updating Nomad from 1.5.3 to 1.7.6, I can't run any job on two of the three nodes (the same job). I get the error:
```sh
failed to launch command with executor: rpc error: code = Unknown desc = unable to start container process: error during container init: read init-p: connection reset by peer
```
I also tried to create a simple job that will run /bin/bash, but I still face the issue. I also tried to reboot servers and update all packages on host machines, but that didn't help.
I also tried to remove all cache data from all servers. I thought it was a file access issue at first. If you use the default user (anonymous), the same error occurs.

### Job file
```hcl
job "test-job" {
  datacenters = ["dc1"]
  type = "service"
  group "test-group" {
    count = 3
    constraint {
      operator  = "distinct_hosts"
      value     = "true"
    }
    
    restart {
      attempts = 2
      interval = "5m"
      delay = "15s"
      mode = "fail"
    }

    ephemeral_disk {
      size = 300
    }

    task "test-task" {
      user = "cored"
      driver = "exec"
      
      config {
        command = "/bin/bash"
        args = []
      }

      resources {
        cpu = 200
        }
      }
    }
}
```

### Nomad logs
<details>

```sh
-- Logs begin at Mon 2024-03-25 10:46:06 MSK, end at Mon 2024-03-25 11:14:14 MSK. --
Mar 25 10:58:00 srv2.prod nomad[5744]:     2024-03-25T10:58:00.101+0300 [DEBUG] client: updated allocations: index=443167 total=15 pulled=13 filtered=2
Mar 25 10:58:00 srv2.prod nomad[5744]:     2024-03-25T10:58:00.101+0300 [DEBUG] client: allocation updates: added=0 removed=0 updated=13 ignored=2
Mar 25 10:58:00 srv2.prod nomad[5744]:     2024-03-25T10:58:00.124+0300 [DEBUG] client: allocation updates applied: added=0 removed=0 updated=13 ignored=2 errors=0
[...]

PR #23334, opened 2024-06-14
Backport of deps: update runc to 1.1.13 into release/1.8.x

hc-github-team-nomad-core:
## Backport

This PR is auto-generated from #23331 to be assessed for backporting due to the inclusion of the label backport/1.8.x.

---

Update `runc` to 1.1.13 to pick up build support for Go 1.22.4+, in order to ensure we've resolved errors cloning processes into Linux namespaces for libcontainer (`exec` driver) with new versions of Go and older but still supported versions of glibc.

This changeset has two minor quirks:

* Testing shows that the reported issues is already resolved on `main` by upgrading to Go 1.22.4 without this dependency bump, at least for glibc 2.31. Upgrading the dependency should make sure there isn't another glibc version where the problem will still appear.

* This version of `runc` refers to fields in `cilium/ebpf` which are not present in more recent versions of that library. So in order to build, we have to downgrade `cilium/ebpf`. Fortunately, `runc` is the only consumer of that transitive dependency.

Closes: https://github.com/hashicorp/nomad/issues/20212
Ref: https://hashicorp.atlassian.net/browse/NET-10078

---

<details>
<summary> Overview of commits </summary>

  - d9a10a629858929fdc6089ef69682baa10e6043f 

</details>


tgross (approved):
LGTM

PR #23359, opened 2024-06-17
docs: Fix parameter type and default value in client reserved configuration.

liukch:
The **reserved cores** parameter in [Nomad client agent configuration](https://developer.hashicorp.com/nomad/docs/configuration/client#cores) is `int` type, and default value is `0`.

But according to [the source codes](https://github.com/hashicorp/nomad/blob/main/command/agent/config.go#L1209), it is `string` type, and defalut value is `empty string`.

This part of the document has bothered me for a long time, so I want to fix it.

### DIFF
--- a/website/content/docs/configuration/client.mdx
+++ b/website/content/docs/configuration/client.mdx
@@ -380,7 +380,7 @@ see the [drivers documentation](/nomad/docs/drivers).
 
 - `cpu` `(int: 0)` - Specifies the amount of CPU to reserve, in MHz.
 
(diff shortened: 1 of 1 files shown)


liukch:
@tgross Would you willing to have a set to review this PR.

tgross:
@liukch we've got a "community rotation" where someone is assigned to triage issues and PRs. It'll get picked up by them.

tgross (approved):
Thanks @liukch !

PR #23362, opened 2024-06-17
e2e: add Enterprise Option for cluster3.Establish

gulducat:
I'm writing E2E tests for an enterprise feature, and while they're conventionally run against ENT nightly, I figure it might be nice to auto-skip tests that opt in to this new option explicitly, if run against a CE build.

Example usage:

```go
cluster3.Establish(t,
	cluster3.Enterprise(),
	cluster3.Leader(),
	cluster3.LinuxClients(1),
)
```

### DIFF
--- a/e2e/v3/cluster3/cluster3.go
+++ b/e2e/v3/cluster3/cluster3.go
@@ -27,6 +27,7 @@ type Cluster struct {
 	vaultClient  *vaultapi.Client
 
 	timeout        time.Duration
+	enterprise     bool
 	leaderReady    bool
 	consulReady    bool
 	vaultReady     bool
@@ -36,6 +37,8 @@ type Cluster struct {
 }
 
 func (c *Cluster) wait() {
+	c.t.Helper()
+
 	errCh := make(chan error)
 
 	statusAPI := c.nomadClient.Status()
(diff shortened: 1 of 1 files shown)


e2e/v3/cluster3/cluster3.go
```diff
@@ -149,11 +152,24 @@ func (c *Cluster) wait() {
...
 	}
+
+	// t.Skip() should not be run in a goroutine, so this check is separate,
+	// and by the time the above have passed, retries should not be necessary.
+	if c.enterprise {
+		_, _, err := c.nomadClient.Operator().LicenseGet(nil)
+		if err != nil && err.Error() == "Nomad Enterprise only endpoint" {
+			c.t.Skip("not an enterprise cluster")
+		} else {
+			must.NoError(c.t, err, must.Sprint("expect running Enterprise cluster"))
```
shoenig, line 163: doesn't the `else` need to be `else if err != nil` ? (i.e. this _is_ an enterprise cluster and we should allow the test to run)

gulducat: the `must.NoError()` covers nil-checking in that case.  if it's nil, then continue, else t.Fatal (via must)

example playground: https://go.dev/play/p/VriAyDQ3750

but if the syntax is confusing, I'm happy to tweak it to clarify intent.

shoenig: oh right, i just can't read

shoenig (approved):

PR #23365, opened 2024-06-17
[ui] Dont show keyboard hints when a user is trying to take a screenshot

### LINKED ISSUE #23361: Nomad WebUI should detect if modifiers in addition to Shift are being pressed, and hide the shortcuts
### Proposal
Nomad WebUI should detect if modifiers in addition to Shift are being pressed, and hide the shortcuts if so.

### Use-cases
It would be nice to take screenshots on MacOS with the default shortcuts, w/o those labels showing up.

### Attempted Solutions
I tried adjusting the order in which I held the modifier keys. (ie, CMD->Ctrl->Shift + 4)

No dice.

philrenaud:
Resolves #23361 

## How to test
- Buy a macbook (or pretend)
- Hold down shift, to see hints. Release and they're gone!
- Hold down cmd+shift+4 to take a screenshot
- Note that in holding the command key, the shift key no longer triggers the display of keyboard hints.

### DIFF
--- a/ui/app/services/keyboard.js
+++ b/ui/app/services/keyboard.js
@@ -348,11 +348,8 @@ export default class KeyboardService extends Service {
       const shifted = event.getModifierState('Shift');
       if (type === 'press') {
         if (key === 'Shift') {
-          // if cmd or ctrl are pressed, don't show hints — this is likely a user trying to take a screenshot.
-          if (
-            event.getModifierState('Meta') ||
-            event.getModifierState('Control')
-          ) {
+          // if cmd/windows key is pressed, don't show hints — this is likely a user trying to take a screenshot.
+          if (event.getModifierState('Meta')) {
             return;
           }
           this.displayHints = true;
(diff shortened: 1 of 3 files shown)


ui/app/services/keyboard.js
```diff
@@ -348,6 +348,13 @@ export default class KeyboardService extends Service {
       const shifted = event.getModifierState('Shift');
       if (type === 'press') {
         if (key === 'Shift') {
+          // if cmd or ctrl are pressed, don't show hints — this is likely a user trying to take a screenshot.
+          if (
+            event.getModifierState('Meta') ||
+            event.getModifierState('Control')
```
gulducat, line 354: who's `Control` for?  on my linux laptop, I just press `print sc` button (maybe with a `fn` modifier key, since it shares `F10` in my case)

gulducat (approved):
:camera_flash:

martisah (approved):

PR #23377, opened 2024-06-18
api: use the task in Allocations.GetPauseState

gulducat:
Without this, the method just doesn't work. :upside_down_face: 

For context, the "pause" state is the state of a task that has a [schedule{} block](https://developer.hashicorp.com/nomad/docs/job-specification/schedule) on it (enterprise feature).


philrenaud (approved):

schmichael (commented):
Oops! A test around this would be welcome!

Issue #20516, opened 2024-05-02
Nomad with Consul Connect Service Mesh randomly fails with a bootstrap error


ryanhulet:
### Nomad version
```
Nomad v1.7.7
BuildDate 2024-04-16T19:26:43Z
Revision 0f34c85ee63f6472bd2db1e2487611f4b176c70c
```

### Operating system and Environment details
```
NAME="Ubuntu"
VERSION="20.04.5 LTS (Focal Fossa)"
```
### Issue
Ever since upgrading Nomad to 1.7.7 and Consul to 1.18.1, nomad jobs will fail repeatedly with the error:
`envoy_bootstrap: error creating bootstrap configuration for Connect proxy sidecar: exit status 1; see: <https://developer.hashicorp.com/nomad/s/envoy-bootstrap-error>`
This will happen multiple times, until eventually the job will schedule and start correctly. This results in multiple jobs stuck in pending for hours on end because of rescheduling back off

### Reproduction steps
Literally run a job similar to the one attached, using Consul with auto-config and mTLS, envoy version 1.27.3. I am also using Vault 1.15.4 for secrets, and Nomad is using Vault and Consul workflow identities

#### Expected Result
A job to schedule and register with Consul Connect successfully

#### Actual Result
REPEATED failures with the envoy_bootstrap error

### Job file (if appropriate)
```
job "canonical-landscape" {
  datacenters = ["<DATACENTER>"]

  namespace = "ops"

  type = "service"

  vault {}
  
 affinity {
    attribute = "${node.unique.name}"
    operator = "set_contains_any"
    value = "<HOSTNAMES>"
    weight = -50
  }

  group "landscape" {

    volume "canonical-landscape" {
      type      = "host"
      read_only = false
      source    = "canonical-landscape"
    }

    volume "gpgs_keys" {
      type      = "host"
      read_only = false
      source    = "gpgs_keys"
    }

    volume "landscape-repository" {
      type      = "host"
      read_only = false
      source    = "landscape-repository"
    }

    network {
      mode = "bridge"

      port "landscape" {
        to = 443
      }

      port "landscape80" {
        to = 80
      }
    }

    service {
      name = "landscape"
      port = "landscape"
      address_mode = "alloc"

      connect {
        sidecar_service {
            proxy {
                upstreams {
                destination_name = "landscape-postgres"
                local_bind_port = 5432
                }
                upstreams {
                    destination_name = "landscape-rabbitmq"
                    local_bind_port = 5672
                }
[...]

Issue #10451, opened 2021-04-26
connect: improve envoy bootstrap coordination 


shoenig:
Currently the envoy bootstrap process needs Consul to accomplish 2 things before invoking `consul envoy -bootstrap ...`

1) For sidecar proxies, the parent service needs to finish being registered into Consul catalog

2) When ACLs are enabled, the minted Service Identity tokens need to be propagated through Consul 

The [implementation](https://github.com/hashicorp/nomad/blob/v1.0.4/client/allocrunner/taskrunner/envoy_bootstrap_hook.go#L204) right now is to simply retry 3 times on 2 second intervals. Until the final failure, errors invoking Consul are ignored. This implies the above 2 pre-requsites must be met within 6 seconds of the `envoy_bootstrap_hook` running, which is [not always the case](https://github.com/hashicorp/nomad/issues/9307#issuecomment-763384985). 

Ideally we would have blocking queries to work with, but neither the agent service registration or acl token creation APIs support blocking queries.

https://www.consul.io/api-docs/agent/service#register-service
https://www.consul.io/api-docs/acl/tokens#create-a-token

I spent a little time looking into setting a watch on the Catalog waiting for the service to be registered, but AFAICT Consul will trip the watch on _any_ service changes, not just the service we care about. In the end, using a watch was a lot more chatty over the wire than just polling every few seconds. For the time being, the best option here might be to just increase the amount of time the boostrap hook allows before giving up, maybe even making it a Nomad client config parameter. Experimenting locally it's hard to gauge how long to wait is "reasonable", so how about an increase from 6 to 60 seconds?

Ideas welcome.

TimMensch:
Looking at this error shortly after its two year anniversary of being posted with zero actual responses from the developers is _not_ giving me warm-and-fuzzy feelings about the reliability of Consul/Nomad service meshes.

I'm still seeing frequent timeouts on high-end VMs with near zero load. 

This really strikes me as fundamentally a Consul bug (or design flaw). Why isn't a registered task effectively immediately available, especially on a server with effectively zero load? Raising the timeout is a band-aid to mask the fact that it can take 60+ seconds to register a task on an otherwise idle Consul cluster, and it clearly isn't enough.

<img width="870" alt="firefox_2023-05-04_09-57-29" src="https://user-images.githubusercontent.com/530873/236277109-1126759b-b7e7-4f2c-b59c-8937bc01eaf0.png">

[...]

Issue #9307, opened 2020-11-10
consul connect is extremely unreliable at startup


kneufeld:
I've just had multiple multi-hour outages in the past two days because consul connect proxies would not start.

```
envoy_bootstrap: error creating bootstrap configuration for Connect proxy sidecar: exit status 1
```

That's it, no logs because it never actually started. Granted this is a complicated job with 5 groups and proxies between them but it used to work. It seems as soon as I had multiple jobs (built from a template) the proxies started to act up. Proxy names are all unique based on job name and group.

All I can do it just repeatedly hit restart until the gods smile on me and the proxy finally starts.

### Nomad version

Started at 0.11.6 and desperately tried to upgrade to 0.12.7 to no avail.

```
Nomad v0.12.7 (6147cb578794cb2d0c35d68fe1791728a09bb081)
```

### Operating system and Environment details

Ubuntu 18.04 and Ubuntu 20.04

### Reproduction steps

Make a job with multiple groups and multiple connect sidecar between them (but not circular). The first instance of that job will probably work. Make 10 more.

After investing _a lot_ of time and energy in nomad I'm afraid I'll have drop it for I have no idea what I'm going to do...

shoenig:
Hi @kneufeld sorry you're having trouble with this. 

Seeing the `exit status 1` error message for creating bootstrap configuration indicates something went wrong when Nomad [execs consul](https://www.consul.io/commands/connect/envoy#usage) to retrieve the envoy bootstrap configuration from Consul. Did you happen to capture Consul logs while this is happening, which may shed some light into what's going on? What version of Consul are you using?

kneufeld:
```
Nov 10 08:54:21 clients-nomad1.west.artsman.com consul[6571]:     2020-11-10T08:54:21.287-0800 [ERROR] agent.proxycfg: watch error: service_id=_nomad-task-bb39ddd4-951b-52d2-12a7-8bb290adebef-group-listener-     <redacted>-proxy-redis--sidecar-proxy id=service-http-checks:_nomad-task-bb39ddd4-951b-52d2-12a7-8bb290adebef-group-listener-<redacted>-proxy-redis- error="error filling agent cache: Internal cache failure: service '_nomad-task-bb39ddd4-951b-52d2-12a7-8bb290adebef-group-listener-<redacted>-proxy-redis-' not in agent state"
```

This is the only thing I can find that looks interesting.

```
# consul version
Consul v1.8.5
Revision 1e03567d3
[...]

PR #23381, opened 2024-06-18
Consul: add preflight checks for Envoy bootstrap

tgross:
Nomad creates Consul ACL tokens and service registrations to support Consul service mesh workloads, before bootstrapping the Envoy proxy. Nomad always talks to the local Consul agent and never directly to the Consul servers. But the local Consul agent talks to the Consul servers in stale consistency mode to reduce load on the servers. This can result in the Nomad client making the Envoy bootstrap request with a tokens or services that have not yet replicated to the follower that the local client is connected to. This request gets a 404 on the ACL token and that negative entry gets cached, preventing any retries from succeeding.

To workaround this, we'll use a method described by our friends over on `consul-k8s` where after creating the objects in Consul we try to read them from the local agent in stale consistency mode (which prevents a failed read from being cached). This cannot completely eliminate this source of error because it's possible that Consul cluster replication is unhealthy at the time we need it, but this should make Envoy bootstrap significantly more robust.

This changset adds preflight checks for the objects we create in Consul:
* We add a preflight check for ACL tokens after we login via via Workload Identity and in the function we use to derive tokens in the legacy workflow. We do this check early because we also want to use this token for registering group services in the allocrunner hooks.
* We add a preflight check for services right before we bootstrap Envoy in the taskrunner hook, so that we have time for our service client to batch updates to the local Consul agent in addition to the local agent sync.

We've added the timeouts to be configurable via node metadata rather than the usual static configuration because for most cases, users should not need to touch or even know these values are configurable; the configuration is mostly available for testing.

Fixes: https://github.com/hashicorp/nomad/issues/9307
Fixes: https://github.com/hashicorp/nomad/issues/10451
Fixes: https://github.com/hashicorp/nomad/issues/20516

Ref: https://github.com/hashicorp/consul-k8s/pull/887
Ref: https://hashicorp.atlassian.net/browse/NET-10051
Ref: https://hashicorp.atlassian.net/browse/NET-9273
Follow-up: https://hashicorp.atlassian.net/browse/NET-10138

---

Notes for reviewers: 
* This is an unfortunately large PR as I had to touch a lot of test code and there are separate code paths for supporting the legacy workflow. To help make this a little more reviewable, I've split this into 2 commits with the code changes and their specific tests, a third commit with a test helper fix, and a final commit with docs and changelog.
* I've got a follow-up described in [NET-10138](https://hashicorp.atlassian.net/browse/NET-10138) to reduce the risk of this kind of bug in the future by expanding our Consul cluster in E2E to be a 3-node cluster.
* The backport to 1.6.x+ent is not going to be clean, as all this code was heavily reworked for WI and multi-cluster support in 1.7.0. So I'm going to manually backport the relevant sections and open a new PR for that in the ENT repo.

[NET-10138]: https://hashicorp.atlassian.net/browse/NET-10138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/consul/consul.go
+++ b/client/consul/consul.go
@@ -24,7 +24,7 @@ import (
 type TokenDeriverFunc func(context.Context, *structs.Allocation, []string) (map[string]string, error)
 
 // ServiceIdentityAPI is the interface the Nomad Client uses to request Consul
-// Service Identity tokens through Nomad Server.
+// Service Identity tokens through Nomad Server. (Deprecated: will be removed in 1.9.0)
 //
 // ACL requirements
 // - acl:write (used by Server only)
@@ -54,7 +54,7 @@ type JWTLoginRequest struct {
 }
 
 // Client is the interface that the nomad client uses to interact with
-// Consul.
+// Consul tokens
 type Client interface {
 	// DeriveTokenWithJWT logs into Consul using JWT and retrieves a Consul ACL
 	// token.
@@ -202,8 +202,7 @@ func (c *consulClient) TokenPreflightCheck(pctx context.Context, t *consulapi.AC
 		retry++
 		backoff := helper.Backoff(
 			c.preflightCheckBaseInterval, c.preflightCheckBaseInterval*2, retry)
-		c.logger.Trace("waiting for Consul stale query on token",
-			"error", err, "backoff", backoff)
+		c.logger.Trace("Consul token not ready", "error", err, "backoff", backoff)
 		timer.Reset(backoff)
 		select {
 		case <-ctx.Done():
--- a/client/consul/consul.go
+++ b/client/consul/consul.go
@@ -4,20 +4,24 @@
 package consul
 
 import (
+	"context"
 	"fmt"
+	"time"
 
 	consulapi "github.com/hashicorp/consul/api"
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-multierror"
 
+	"github.com/hashicorp/nomad/helper"
 	"github.com/hashicorp/nomad/helper/useragent"
 	"github.com/hashicorp/nomad/nomad/structs"
 	"github.com/hashicorp/nomad/nomad/structs/config"
 )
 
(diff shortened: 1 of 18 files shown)


client/consul/consul.go
```diff
@@ -78,7 +92,14 @@ type ConsulClientFunc func(config *config.ConsulConfig, logger hclog.Logger) (Cl
...
 // partition
 func NewConsulClientFactory(node *structs.Node) ConsulClientFunc {
+
+	// these node values will be evaluated at the time we create the hooks, so
+	// we don't need to worry about them changing out from under us
 	partition := node.Attributes["consul.partition"]
+	preflightCheckTimeout := durationFromMeta(
+		node, "consul.token_preflight_check.timeout", time.Second*10)
+	preflightCheckBaseInterval := durationFromMeta(
+		node, "consul.token_preflight_check.base", time.Millisecond*500)
```
schmichael, line 102: Hm, I'm not super familiar with this code, but it appears `NewConsulClientFactory` will be called  from `NewClient` and therefore the timeout and interval node meta will be captured only during agent startup. Don't the durationFromMeta calls need to happen inside the closure and use `client.Node()` instead of a captured Node?

schmichael (approved):
I'm very excited to see envoy bootstrapping get more reliable! Thanks for following this through.

client/consul/consul.go
tgross: That's a good catch. For partitions, we don't expect that to change because it's part of the fingerprint we get from the agent. But metadata can be changed via the `node meta` command so yeah we'll need to move this into the closure.

tgross: Fixed! (And rebased on main) Also:

> I'm not super familiar with this code
The various dependency injection tricks we're doing here to avoid circular references doesn't help. This code all got really janky during the big WI + Consul/Vault refresh. Once we remove the legacy workflow we should be able to refactor this in the process to be a little more comprehensible.

Issue #23376, opened 2024-06-18
panic during node heartbeat reset


lattwood:
### Nomad version
Nomad v1.7.7
BuildDate 2024-04-16T19:26:43Z
Revision 0f34c85ee63f6472bd2db1e2487611f4b176c70c

### Issue
This happens:
```
{"@level":"warn","@message":"node TTL expired","@module":"nomad.heartbeat","@timestamp":"2024-06-18T03:35:10.193628Z","node_id":"a747a728-9219-d869-391f-85a7b68e9038"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xf0 pc=0x1f86009]
goroutine 909614773 [running]:
github.com/hashicorp/nomad/nomad.(*nodeHeartbeater).disconnectState(0xc000f81a40, {0xc0344d80c0, 0x24})
        github.com/hashicorp/nomad/nomad/heartbeat.go:188 +0x69
github.com/hashicorp/nomad/nomad.(*nodeHeartbeater).invalidateHeartbeat(0xc000f81a40, {0xc0344d80c0, 0x24})
        github.com/hashicorp/nomad/nomad/heartbeat.go:158 +0x22c
github.com/hashicorp/nomad/nomad.(*nodeHeartbeater).resetHeartbeatTimerLocked.func1()
        github.com/hashicorp/nomad/nomad/heartbeat.go:131 +0x1f
created by time.goFunc
        time/sleep.go:177 +0x2d
```

`StateStore.NodeByID` can return `*structs.Node, nil`, `nil, error`, and `nil, nil`. 
 `nodeHeartbeater.disconnectState` only checks for `nil, error`, assuming that only one of the returned parameters can be non-Nil.

### Relevant code sections
 https://github.com/hashicorp/nomad/blob/54115a1b3707375f629eb1b3e1768633707104d9/nomad/heartbeat.go#L181-L190
 https://github.com/hashicorp/nomad/blob/54115a1b3707375f629eb1b3e1768633707104d9/nomad/state/state_store.go#L1652-L1665

tgross:
[...]

PR #23383, opened 2024-06-18
server: fix panic if heartbeat reset happens for GC'd node

tgross:
When setting up the timer for heartbeat invalidation, there's no control that allows us to remove that timer when the node is GC'd. If the GC window is narrow enough, it's possible to GC a node that has a waiting heartbeat timer. In this case, we hit a bug where querying for the node returns `nil` and this is incorrectly handled when checking for disconnect/reconnect state. Fix this bug by correctly handling a `nil` node and allowing the `Node.Update` RPC to fire normally (which then errors correctly).

Fixes: https://github.com/hashicorp/nomad/issues/23376
Ref: https://hashicorp.atlassian.net/browse/NET-10109


pkazmierczak (approved):
LGTM!

jrasell (approved):

Issue #23314, opened 2024-06-13
CPU exhausted even with plenty available


meyerbro:
### Nomad version
1.8.0

### Operating system and Environment details
Amazon Linux 2023

### Issue
More than 4000 Mhz available on each node but can't scale extra 1000 Mhz alloc at all

### Reproduction steps
Run on EC2 (AL2023) with or without dmidecode and try to fit 90% CPU of any nomad client, it will stop at around 60% usage.
My CPU is around 10150Mhz in AWS for each node and I only reserve 500Mhz and no cpu cores for the nomad client.
Funny enough, if I set cpu_total_compute to 10000Mhz I can fit 9000Mhz without issues!
I have 7 nodes and I can fit 63 1000Mhz nginx allocs in previous versions of Nomad but can't without hacking cpu_total_compute to the same amount of Mhz I have on those nodes...
My nodes in AWS are "r5a.xlarge".

#### Expected Result
Fill 90% of the CPU, just like previous versions.

#### Actual Result
cpu exausted in x nodes

### Job file
```
job "job" {
  type = "service"
  datacenters = [REDACTED]
  namespace = "PLATFORM"
  group "group" {
    count = 63
    task "nginx" {
      resources {
        cpu    = 1000
        memory = 1024
      }
      driver = "docker"
      config {
        image = "nginx"
      }
    }
  }
}
```

jrasell:
Hi @meyerbro and thanks for raising this issue. Is it possible to add additional detail of the error(s) you are seeing? It would be useful to have the output from when you try and register a job that should be placed that doesn't along with the node status showing the available resources.

meyerbro:
Hi @jrasell, I just get "Resources exhausted on 7 nodes" and "Dimension "cpu" exhausted on 7 nodes". And in the UI I see placement failures for 21 unplaced allocations.

tgross:
This looks to be the same issue as https://github.com/hashicorp/nomad/issues/23371; I'm working to get some eyes on this. (Internal ref https://hashicorp.atlassian.net/browse/NET-10107)

pkazmierczak:
hey @meyerbro, the patch landed in main and will be released in 1.8.2. Thanks again for reporting!

meyerbro:
Thank you all!

PR #23386, opened 2024-06-19
scheduler: fix a bug where we subtract reserved node resources twice

### LINKED ISSUE #23371: Unallocated CPU resources are not being calculated correctly
### Nomad version
Nomad v1.8.0

### Issue

When declaring reserved CPU compute unallocated resources are not being correctly calculared.

### Reproduction steps

1)
❯ cat config.hcl 
```
client {
  enabled = true
  
  reserved {
    cpu            = 4000
    memory         = 1000
  }
}
```
2) nomad agent -dev -config config.hcl
3) nomad plan example.nomad.hcl
```
+ Job: "example"
+ Task Group: "test" (1 create)
  + Task: "test" (forces create)

Scheduler dry-run:
  Task Group "test" (failed to place 1 allocation):
    * Dimension "cpu" exhausted on 1 nodes
```

compute

#### Expected Result

- Job has enought resourecs to be run

#### Actual Result

- Can't schedule job

### Job file (if appropriate)
```
job "example" {
  datacenters = ["*"]
  type = "service"
  group "test" {
    task "test" {
      driver = "raw_exec"
      config {
        command = ["/bin/true"]
      }
      resources {
        cpu    = 8000 
        memory = 256
      }
    }
  }
}
```

For this job to run on this node I have to set the cpu to 6400 for it to be allocated correctly (setting 6401 will not allow it). It seems like reserved CPU is taken into account two times.

pkazmierczak:
Fixes a bug in the [`nodeResources.Comparable` method](https://github.com/hashicorp/nomad/blob/a0a717342775bce914aeb7ef052e59d3cb6dc41c/nomad/structs/structs.go#L3194), where CPU resources were accidentally offset with reserved resources, whereas functions that use this field expect total CPU resources. 

Fixes #23371 and #23314

### DIFF
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -3191,7 +3191,7 @@ func (n *NodeResources) Comparable() *ComparableResources {
 	c := &ComparableResources{
 		Flattened: AllocatedTaskResources{
 			Cpu: AllocatedCpuResources{
-				CpuShares:     int64(n.Processors.Topology.UsableCompute()),
+				CpuShares:     int64(n.Processors.Topology.TotalCompute()),
 				ReservedCores: reservableCores,
 			},
 			Memory: AllocatedMemoryResources{
--- a/.changelog/23386.txt
+++ b/.changelog/23386.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+scheduler: Fix a bug where reserved resources are not calculated correctly
+```
--- a/nomad/structs/funcs.go
+++ b/nomad/structs/funcs.go
@@ -172,6 +172,7 @@ func AllocsFit(node *Node, allocs []*Allocation, netIdx *NetworkIndex, checkDevi
 	// Check that the node resources (after subtracting reserved) are a
 	// super set of those that are being allocated
 	available := node.NodeResources.Comparable()
+	available.Subtract(node.ReservedResources.Comparable())
 	if superset, dimension := available.Superset(used); !superset {
 		return false, dimension, used, nil
 	}
--- a/nomad/structs/funcs.go
+++ b/nomad/structs/funcs.go
@@ -172,7 +172,6 @@ func AllocsFit(node *Node, allocs []*Allocation, netIdx *NetworkIndex, checkDevi
 	// Check that the node resources (after subtracting reserved) are a
 	// super set of those that are being allocated
 	available := node.NodeResources.Comparable()
-	available.Subtract(node.ReservedResources.Comparable())
 	if superset, dimension := available.Superset(used); !superset {
 		return false, dimension, used, nil
 	}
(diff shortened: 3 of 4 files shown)


nomad/structs/structs.go
```diff
@@ -3191,7 +3191,7 @@ func (n *NodeResources) Comparable() *ComparableResources {
 	c := &ComparableResources{
 		Flattened: AllocatedTaskResources{
 			Cpu: AllocatedCpuResources{
-				CpuShares:     int64(n.Processors.Topology.UsableCompute()),
+				CpuShares:     int64(n.Processors.Topology.TotalCompute()),
```
tgross, line 3194: The comment in the caller in `AllocsFit` seems to agree that this should be `TotalCompute`:

```go
	// Check that the node resources (after subtracting reserved) are a
	// super set of those that are being allocated
	available := node.NodeResources.Comparable()
	available.Subtract(node.ReservedResources.Comparable())
	if superset, dimension := available.Superset(used); !superset {
		return false, dimension, used, nil
	}
```

But it bothers me that changing this wouldn't also cause some tests to fail. Are we missing tests on `AllocsFit` that exercise this code path?

pkazmierczak: Right, so this is very tricky to test. 

I found this bug while chasing `AllocsFit` and initially just removed the `available.Subtract(node.ReservedResources.Comparable())` line, because I thought: ah, this just subtracts what's already been subtracted. It solves the issue of course, but making this change is what breaks _a lot_ of tests in the plan applier and scheduler. The reason is: we mock nodes and their resources. The only way to properly test this is an e2e test, but that's also tricky: we'd have to create a client with reserved resources, run a job that requests all the remaining available resources and make sure it succeeds. To do that, we'd have to manipulate client config in the e2e (that is tricky), and dynamically create the jobspec based on the available resources of the e2e node. Can be done, but it's a lot of work I think. Unless I'm missing something?

tgross: I suspect we could set reserved resources on the mocked nodes on some of those existing scheduler and plan applier tests such that they pass with this change and fail without it.

pkazmierczak: ok. *finally* found the culprit. every node that has reserved resources set by client conf has an additional field [`OverrideWitholdCompute`](https://github.com/hashicorp/nomad/blob/a73965f95f2c305d55f578b70cfc4c24dab82b59/client/lib/numalib/topology.go#L68) set in its `Topology` on "legacy" systems: https://github.com/hashicorp/nomad/blob/a73965f95f2c305d55f578b70cfc4c24dab82b59/nomad/structs/cpucompat_generic.go#L45
This field was missing from the mocked node in our unit tests. After adding it, tests fail without this PR's change.

PR #23406, opened 2024-06-21
docs/jobspec: Fix "task" block placement

heitorPB:
The `task` block should be inside the `group` block. The example in the page places the `task` block directly under `job`.

### DIFF
--- a/website/content/docs/job-specification/job.mdx
+++ b/website/content/docs/job-specification/job.mdx
@@ -28,6 +28,10 @@ job "docs" {
 
   group "example" {
     # ...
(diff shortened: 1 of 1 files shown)


heitorPB:
Is it really necessary to sign a CLA for such a minor change?

tgross (approved):
Thanks @heitorPB!

> Is it really necessary to sign a CLA for such a minor change?

Technically, yes we're supposed to require it for everything, but no one is going to complain if I just go ahead and merge this minor fix. :grinning:

Issue #23391, opened 2024-06-19
Deprication for `cron` is not documented


EugenKon:
### Problem
![image](https://github.com/hashicorp/nomad/assets/139780495/92925f91-94ea-4431-a09c-3715bbf3af76)

But the documentation says nothing about this:
![image](https://github.com/hashicorp/nomad/assets/139780495/47b1ac48-4bd9-4cfe-bae9-0160c798aa12)

https://developer.hashicorp.com/nomad/docs/job-specification/periodic#periodic-parameters

Antvirf:
Small fix https://github.com/hashicorp/nomad/pull/23424

PR #23424, opened 2024-06-23
docs: add deprecation notice to `cron` on `docs/job-specification/periodic`

Antvirf:
Fixes https://github.com/hashicorp/nomad/issues/23391

There are quite a few different approaches to mark deprecation across the docs, I picked this as it was the most explicit parameter-level option in my opinion.

Screenshot below:

Screenshot from 2024-06-23 21-46-17

Other refs:

- The PR that added the original deprecation notice to the tool itself https://github.com/hashicorp/nomad/pull/17858
- Changelog entry mentioning that PR for Nomad 1.6.2 https://github.com/hashicorp/nomad/blob/main/CHANGELOG.md#162-september-13-2023

### DIFF
--- a/website/content/docs/job-specification/periodic.mdx
+++ b/website/content/docs/job-specification/periodic.mdx
@@ -34,17 +34,16 @@ consistent evaluation when Nomad spans multiple time zones.
 
 ## `periodic` Parameters
 
-- `cron` (_deprecated_: Replaced by `crons` in 1.6.2) `(string: <optional>)` - Specifies a cron expression configuring the
+- `cron` (_deprecated_: Replaced by `crons` in 1.6.2) `(string)` - Specifies a cron expression configuring the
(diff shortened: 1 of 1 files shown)


tgross (changes requested):
Thanks for picking this up, @Antvirf. Changing the parameter to `(string: <optional>)` might be a little confusing though... I get where you're going with that, but it's required if `crons` is unset. Let's change it just `(string)` and change the note at the end of `crons` to:

> Either `cron` or `crons` must be set, but not both.

And then add the same note to the end of `cron` too.

Antvirf:
@tgross done

tgross (approved):
LGTM! Thanks @Antvirf!

Preview link: https://nomad-7e30x9qv1-hashicorp.vercel.app/nomad/docs/job-specification/periodic#periodic-parameters

PR #23427, opened 2024-06-24
[ui] Jobs list should handle 502s and 504s gracefully

### LINKED ISSUE #23228: [UI] The Jobs page shows "No jobs found" after being idle for two minutes.
### Nomad version
Nomad v1.8.0
BuildDate 2024-05-28T17:38:17Z
Revision 28b82e4b2259fae5a62e2ed47395334bea5a24c4

Distributor ID: Ubuntu
Description:    Ubuntu 24.04 LTS
Release:        24.04
Codename:       noble

### Issue
The Jobs page shows "No jobs found" after being idle for a few minutes. This might worry users if they have many jobs running and become lost (even if only in display). It returns to normal once you refresh the page.

### Reproduction steps
1. Upgrade Nomad Version from 1.7.5 to 1.8.0 _(not sure if this helps)_
2. Login to Nomad and navigate to Jobs menu
3. Leave the Page/Tab or become idle for 2 minutes

#### Expected Result
The page should display the jobs whether you are idle or not.
image

#### Actual Result
The Jobs page shows "No jobs found." when you become idle for two minutes.
image

image

philrenaud:
Previously, our jobs index page would handle handle all errors when fetching jobs by resetting state and trying again (with 5-second backoff).

This PR makes it so that proxy-timeout specific errors instead throw a notification, but keep the cached jobs shown on the page.

The user is then presented with two options to move forward:
- pause fetch attempts (for this session or until they click "Manually fetch jobs")
- Immediately try again (in case they extended their proxy settings in the background and want to immediately cancel pending requests and try again, or restart previously-paused attempts)

Resolves #23228 
image

### DIFF
--- a/ui/app/controllers/jobs/index.js
+++ b/ui/app/controllers/jobs/index.js
@@ -22,6 +22,7 @@ export default class JobsIndexController extends Controller {
   @service store;
   @service userSettings;
   @service watchList;
+  @service notifications;
 
   @tracked pageSize;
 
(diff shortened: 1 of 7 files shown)


ui/app/templates/jobs/index.hbs
```diff
@@ -5,337 +5,348 @@
...
+  {{#if this.showingCachedJobs}}
+    <Hds::Alert @type="inline" @color="warning" id="jobs-list-cache-warning" as |A|>
+      <A.Title>Error fetching jobs — shown jobs are cached</A.Title>
+      <A.Description>Jobs shown are cached and may be out of date. Often, this is due to a short timeout in proxy configurations. Consider increasing the timeout.</A.Description>
+      {{#if this.watchJobIDs.isRunning}}
+        <A.Button @text="Stop polling for job updates" @color="secondary" {{on "click" this.pauseJobFetching}} />
+      {{/if}}
+      <A.Button @text="Manually fetch jobs" @color="secondary" {{on "click" this.restartJobList}} />
+    </Hds::Alert>
+  {{/if}}
```
philrenaud, line 18: (Hide whitespace during review; I corrected an old indentation issue here)

ui/app/controllers/jobs/index.js
```diff
@@ -156,12 +157,70 @@ export default class JobsIndexController extends Controller {
...
+    if (notification) {
+      notification.destroyMessage();
+    }
+    this.watchList.jobsIndexIDsController.abort();
+    this.watchList.jobsIndexDetailsController.abort();
+    this.watchJobIDs.cancelAll();
+    this.watchJobs.cancelAll();
+  }
+
+  @action restartJobList() {
```
philrenaud, line 176: Note: I let the user click "Try fetching again" before they've paused here even though the current retry on failure is 5sec with an eye to the future: we might extend that timeout beyond 5sec, or more likely add something like exponential backoff.

For now, though, it's an "I want to see the network request fire immediately to see if my new proxy config settings work" button.

ui/app/controllers/jobs/index.js
```diff
@@ -156,12 +157,70 @@ export default class JobsIndexController extends Controller {
...
   //#region querying
 
+  /**
+   *
+   * Let the user know that there was difficulty fetching jobs, but don't overload their screen with notifications.
+   * Set showingCachedJobs to tell the template to prompt them to extend timeouts
+   * @param {Error} e
+   */
+  notifyFetchError(e) {
+    const firstError = e.errors[0];
```
philrenaud, line 205: I think I could `.any()` this, but ¯\_(ツ)_/¯

ui/app/controllers/jobs/index.js
```diff
@@ -257,8 +328,14 @@ export default class JobsIndexController extends Controller {
...
           this.pendingJobs = newJobs;
         }
+        if (Ember.testing) {
+          break;
+        }
         yield timeout(throttle);
       } else {
+        if (Ember.testing) {
+          break;
+        }
```
philrenaud, line 338: A small change to my behaviour for testing purposes: I want to break out of a test on error now, too, which hadn't come up before. Without these lines, my test will be caught in a terrible loop.

ui/app/serializers/job.js
```diff
@@ -71,7 +71,13 @@ export default class JobSerializer extends ApplicationSerializer {
     return super.normalize(typeHash, hash);
   }
 
-  normalizeQueryResponse(store, primaryModelClass, payload, id, requestType) {
+  normalizeQueryResponse(
+    store,
+    primaryModelClass,
+    payload = [],
```
philrenaud, line 77: Only notable change here is a default of payload to `[]` so I don't have to do a bunch of conditional guards with [elvis operators](https://en.wikipedia.org/wiki/Elvis_operator) later

tgross (approved):
LGTM :+1:

shoenig (approved):

Issue #20585, opened 2024-05-14
Nomad failing to create templating sandbox on Windows; appears to leak DACL entries on nomad.exe


tomqwpl:
### Nomad version
Nomad v1.7.5
BuildDate 2024-02-13T15:10:13Z
Revision 5f5d4646198d09b8f4f6cb90fb5d50b53fa328b8

### Operating system and Environment details
Windows 11

### Issue
Repeated errors of
```
10:21.868+0100 [ERROR] client.alloc_runner.task_runner.task_hook.template: failed to create template manager: alloc_id=03b05e98-3048-9dd0-c4ff-b6e10ef27059 task=Main error="could not create platform sandbox: could not grant object access: could not create new DACL for \"c:\\path-to-installation-of\\nomad.exe\": The parameter is incorrect."
    2024-05-13T18:
...
    2024-05-13T18:10:21.925+0100 [INFO]  client.alloc_runner.task_runner: not restarting task: alloc_id=03b05e98-3048-9dd0-c4ff-b6e10ef27059 task=Main reason="Policy allows no restarts"
    2024-05-13T18:10:21.925+0100 [INFO]  client.alloc_runner.task_runner: Task event: alloc_id=03b05e98-3048-9dd0-c4ff-b6e10ef27059 task=Main type="Not Restarting" msg="Policy allows no restarts" failed=true
```

### Reproduction steps
Hard to give specific reproduction steps at this time. I have one instance of Nomad running on a laptop, this isn't a "production" environment.

This occurred when I was performing some load testing on a solution that utilises nomad. I submitted 1000 jobs in quick succession, all of which were fairly small jobs. Many succeeded, but very many failed with this error. To it feels like a load issue. 
Note that I've deliberately got the jobs configured not to allow restarts (I don't want then to be rerun just because they return a non zero exit code)

*Possible* reproduction would be to submit a very large number of jobs, each of which has default resource requirements, and each of which is a "raw-exec" doing something simple like "dir". Nomad will attempt to schedule maybe 100 at once (32GB memory, default resource requirements are 300MB).

### Actual Result
[...]

Issue #20034, opened 2024-02-26
Template rendering through nomad job failing on windows nodes


pavanrangain:
### Nomad version
1.6.7

### Operating system and Environment details
Windows Server 2019

### Issue
Template rendering in a nomad job fails on windoes nodes since 1.6.7 (issue seen even in 1.6.8). Issue was not there with version 1.6.6

### Reproduction steps
1. Have a nomad job with a template to render to a file reading from consul KV or use default from file system
#### Expected Result
Job should get deployed successfully
#### Actual Result
Job failing with error something as below

Template failed: error rendering "(dynamic)" => "<path removed>//log_config": template render subprocess failed: exit status 0xc0000142
NOTE - actual path removed from error

### Job file (if appropriate)
```
job "test-windoes-job" {
  region      = "global"
  datacenters = ["my-test-cluster"]
  type        = "system"

  group "test-winodows-group" {
    count = 1

    constraint {
      attribute = "${attr.kernel.name}"
      value     = "windows"
    }
  
  task "test-windows-task" {
      driver = "raw_exec"

      artifact {
        source = "<some artifactory url>/filebeat.exe"
      }

      env {
        LOGZIO_CODEC          = "json"
        IP_ADDRESS            = "${attr.unique.network.ip-address}"
        HOSTNAME              = "${attr.unique.hostname}"
        CLUSTER_NAME          = "my-test-cluster"     
      }

      config {
        command = "filebeat.exe"
        args    = ["-c", "local/log_config"]
      }

      template {
        data = <<EOH
          ############################# Filebeat #####################################

          filebeat.inputs:

          - type: log
            paths:
              - ${NOMAD_ALLOC_DIR}/logs/*
            fields:
              logzio_codec: ${LOGZIO_CODEC:'json'}
              token: ${LOGZIO_TOKEN:''}
              clusterName: ${CLUSTER_NAME:''}
            fields_under_root: true
            encoding: utf-8
            ignore_older: 24h
            tail_files: true
            exclude_lines: ${excludeLines:[]}

          #The following processors are to ensure compatibility with version 7
          processors:
          - rename:
              fields:
              - from: "agent"
                to: "beat_agent"
              ignore_missing: true
          - rename:
              fields:
              - from: "log.file.path"
                to: "source"
              ignore_missing: true

          ############################# Output ##########################################
          output:
            logstash:
              enabled: false
              hosts: ["listener.logz.io:5015}"]
        EOH

        destination = "local/log_config"
[...]

PR #23432, opened 2024-06-25
template: disable sandboxed rendering on Windows

pkazmierczak:
Following https://github.com/hashicorp/nomad/pull/23443, we no longer need to sandbox template rendering on Windows. 

Relates to: https://github.com/hashicorp/nomad/issues/20585
Relates to: https://github.com/hashicorp/nomad/issues/20034
Internal ref: https://hashicorp.atlassian.net/browse/NET-9311

This is part of a solution outlined in RFC [NMD-195](https://go.hashi.co/rfc/nmd-195).


tgross (commented):
I think I understand where you're going with this, but the `TaskTemplateManagerConfig` we're passing in is still having a `ReaderFn` and `RenderFn` set. By having those functions just empty `return nil`, I think we end up not rendering at all because we never read anything. That should be detected by tests on Windows but doesn't seem to be?

I'd probably leave the `TaskTemplateManagerConfig` constructor in place for both Windows and default, and have the `ReaderFn` and `RenderFn` on Windows just call `os.Open/Read` and CT's own render function directly.

pkazmierczak:
> but the `TaskTemplateManagerConfig` we're passing in is still having a `ReaderFn` and `RenderFn` set.

yeah these `ReaderFn` and `RenderFn` functions are somewhat misleading. They are only ever used when the sandbox is enabled: https://github.com/hashicorp/nomad/blob/main/client/allocrunner/taskrunner/template/template.go#L1018-L1020 So since we no longer support sandboxing on Windows, they should always return `nil`.

tgross (approved):
> yeah these ReaderFn and RenderFn functions are somewhat misleading. They are only ever used [when the sandbox is enabled](https://github.com/hashicorp/nomad/blob/main/client/allocrunner/taskrunner/template/template.go#L1018-L1020).

Oh right, what I was missing was that the `ReaderFn` and `RenderFn` functions are "factory" functions! (Which is hilarious because I wrote all this junk! :facepalm:) So they're getting called to return a `nil` function for that configuration, which CT will accept and replace with its own default internal functions.

LGTM!

Issue #23407, opened 2024-06-21
dep: update consul-template dependency to v0.39.0


jrasell:
changelog: https://github.com/hashicorp/consul-template/blob/main/CHANGELOG.md#0390-june-20-2024

The new release adds support for Consul sameness groups. Updating the dependency will allow this feature to be used within a job specification template block.

PR #23436, opened 2024-06-25
deps: update Consul API and consul-template

tgross:
* Upgrade consul-template to 0.39.0 to allow template queries of admin partitions and sameness groups.
* Upgrade our Consul API to 1.29.1 because it's required for CT, and to remove the replacement pinned version we were using to pick up some newer Consul API features we needed in 1.7.0.

Fixes: https://github.com/hashicorp/nomad/issues/23407
Ref: https://hashicorp.atlassian.net/browse/NET-10153


pkazmierczak (approved):
LGTM!

PR #23440, opened 2024-06-26
Fixing sync to work only on Issue Link field

mmcquillan:
This will allow Jira issues to be synced from GH issue only via the Jira:Issue Link field and not issue type. 

The new tickets will come across as type:GitHub, but can be changed to bugs or any other Jira issue type and still sync.


jrasell (approved):
:jira:

PR #23443, opened 2024-06-26
docker: validate that containers do not run as ContainerAdmin on Windows

pkazmierczak:
This enables checks for `ContainerAdmin` user on docker images on Windows. It's only checked if users run docker with process isolation and not hyper-v, because hyper-v provides its own, proper sandboxing. 

Relates to: https://github.com/hashicorp/nomad/issues/20585
Relates to: https://github.com/hashicorp/nomad/issues/20034
Internal ref: https://hashicorp.atlassian.net/browse/NET-9311

This is part of a solution outlined in RFC [NMD-195](https://go.hashi.co/rfc/nmd-195).


.changelog/23443.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+docker: Validate that containers aren't running as ContainerAdmin on windows
```
tgross, line 2:
```suggestion
docker: Validate that unprivileged containers aren't running as ContainerAdmin on Windows
```

drivers/docker/driver_test.go
```diff
@@ -3143,3 +3143,68 @@ func TestDockerDriver_GroupAdd(t *testing.T) {
 
 	require.Exactly(t, cfg.GroupAdd, container.HostConfig.GroupAdd)
 }
+
+func TestDriver_createImage_validateContainerAdmin(t *testing.T) {
```
tgross, line 3147: We can move this to `drivers/docker/driver_windows_test.go` with the appropriate build flags so that the tests compile on Linux.

drivers/docker/driver_windows.go
```diff
@@ -15,3 +19,19 @@ func getPortBinding(ip string, port string) docker.PortBinding {
 func tweakCapabilities(basics, adds, drops []string) ([]string, error) {
 	return nil, nil
 }
+
+var containerAdminErrMsg = "running container as ContainerAdmin with is unsafe; change the container user, set task configuration to privileged or disable windows_allow_insecure_container_admin to disable this check"
```
tgross, line 23: I think we want to enable the flag to disable the check?

```suggestion
var containerAdminErrMsg = "running container as ContainerAdmin with is unsafe; change the container user, set task configuration to privileged or enable windows_allow_insecure_container_admin to disable this check"
```

drivers/docker/driver_windows.go
```diff
@@ -15,3 +19,19 @@ func getPortBinding(ip string, port string) docker.PortBinding {
...
+
+func (d *Driver) validateImageUser(user, taskUser string, driverConfig *TaskConfig) error {
+	// we're only interested in the case where isolation is set to "process"
+	// (it's also the default) and when windows_allow_insecure_container_admin
+	// is explicitly set to true in the config
+	if d.config.WindowsAllowInsecureContainerAdmin || driverConfig.Isolation == "hyper-v" {
+		return nil
+	}
+
+	if (user == "ContainerAdmin" || taskUser == "ContainerAdmin") && !driverConfig.Privileged {
```
tgross, line 33: The `task.user` will override the user from the image manifest. Ex. if you have `task.user = "ContainerUser"` (safe) it'll override `"ContainerAdmin"` (unsafe) from the image and the container will run as "ContainerUser". It looks like we have it right in the docs, but I don't think we want this to be and "or" condition.

drivers/docker/driver.go
```diff
@@ -616,17 +616,23 @@ func (d *Driver) createImage(task *drivers.TaskConfig, driverConfig *TaskConfig,
 	}
 
 	// Download the image
-	return d.pullImage(task, driverConfig, client, repo, tag)
+	imageID, imageUser, err := d.pullImage(task, driverConfig, repo, tag)
+
+	// validate the image user (windows only)
+	if err := d.validateImageUser(imageUser, task.User, driverConfig); err != nil {
+		return "", err
+	}
```
tgross, line 624: Suppose we pull the image and find that it's running as `"ContainerAdmin"`. If I try to run the task again on the same node, the image may not have been GC'd. In that case, we'll hit the case above:

> // We're going to check whether the image is already downloaded.

And then skip this validation entirely! I think we want to either copy this method into the case above, or move it into the caller so that any image will get checked.

tgross (changes requested):
The overall approach here is on the right track, @pkazmierczak

drivers/docker/driver_windows.go
pkazmierczak: ugh, sorry, yes. Also there was another typo there. I fix.

drivers/docker/driver_windows.go
pkazmierczak: I fixed in f32473e741d300031619ac6d4cd63d9dd8eaf68c, have a look if I got this correctly.

drivers/docker/driver.go
pkazmierczak: good catch. We validate in the caller now: 492a7501c1a2e53ac798cefc3990ac0eba4af492

tgross (approved):
LGTM!

One last thing we should do is to put a warning in the upgrade guide, but otherwise this looks ready to ship

pkazmierczak:
@tgross can you have a quick 👀 at [15e684b](https://github.com/hashicorp/nomad/pull/23443/commits/15e684bb412a5bcf00e9a543fb1a3fa6236d2185)? is the wording in the upgrade guide ok?

website/content/docs/upgrade/upgrade-specific.mdx
```diff
@@ -13,7 +13,17 @@ upgrade. However, specific versions of Nomad may have more details provided for
...
-## Nomad 1.8.1 (UNRELEASED)
+## Nomad 1.8.2 (UNRELEASED)
+
+#### New `windows_allow_insecure_container_admin` configuration option for Docker driver
+
+In 1.8.2, Nomad will refuse to run jobs that use the Docker driver on Windows
+with [Process Isolation][] that run as `ContainerAdmin`. This is in order to
+provide a more secure environment for these jobs, and this behavior can be
+overridden by setting the new `windows_allow_insecure_container_admin` Docker
+plugin configuration option to `true`.
```
tgross, line 24:
```suggestion
plugin configuration option to `true` or by setting `privileged=true`.
```

PR #23448, opened 2024-06-27
windows: remove winappcontainer and winexec helpers

pkazmierczak:
This removes helper `winappcontainer` and `winexec` helper code, since it is no longer needed after https://github.com/hashicorp/nomad/pull/23432


tgross (approved):
LGTM once https://github.com/hashicorp/nomad/pull/23432 is merged.

Note to other reviewers: I asked @pkazmierczak to pull this out to its own PR so that it's easy to find later if we want to pick this code back up for a hypothetical `winexec` driver or something like that.

PR #23452, opened 2024-06-27
docker: default to hyper-v isolation on Windows

pkazmierczak:
We should default to `hyper-v` isolation mode on Windows, as suggested in discussions between @tgross and @angrycub.

### DIFF
--- a/.changelog/23452.txt
+++ b/.changelog/23452.txt
@@ -1,0 +1,3 @@
+```release-note:breaking-change
+docker: default to hyper-v isolation mode on Windows
+```
--- a/drivers/docker/config.go
+++ b/drivers/docker/config.go
@@ -389,12 +389,9 @@ var (
 		"ipc_mode":           hclspec.NewAttr("ipc_mode", "string", false),
 		"ipv4_address":       hclspec.NewAttr("ipv4_address", "string", false),
 		"ipv6_address":       hclspec.NewAttr("ipv6_address", "string", false),
-		"isolation": hclspec.NewDefault(
-			hclspec.NewAttr("isolation", "string", false),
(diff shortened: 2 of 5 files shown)


tgross (approved):
LGTM!

josegonzalez:
Was the breaking change necessary for a security reason? It seems weird to have otherwise ported a breaking change across several releases otherwise.

pkazmierczak:
Hey @josegonzalez, this change is part of a larger effort to mitigate https://github.com/advisories/GHSA-c866-8gpw-p3mv. Our [initial solution](https://github.com/hashicorp/nomad/commit/df865033494574de14da687c999d66b799899b21) caused some problems for Windows users that ran containers using process isolation. We now provided [a solution](https://github.com/hashicorp/nomad/pull/23443) for process isolation containers, but decided to default to hyper-v since it's not vulnerable to CVE-2024-1329 in the first place and provides a more secure, better default.

tgross:
To add to that, we can assure you that making a breaking change here was not considered lightly. We had no way of fixing the initial solution we had and even our partners at Microsoft were unable to help. So this change was the only way to get `template` rendering for Windows users into an unbroken state without reintroducing CVE-2024-1329 (which is a critical pwn-the-whole-host vulnerability).

josegonzalez:
Should that have been called out as a breaking security-related change then? ATM without extra context, it just looks like ya'll broke BC for no apparent reason (the reasoning is ~fine I suppose).

tgross:
Fair enough, done. You'll also be interested to see the new page we just published with upcoming deprecations/compatibility items: https://developer.hashicorp.com/nomad/docs/release-notes/nomad/upcoming

Issue #23370, opened 2024-06-17
/v1/job/statuses doesn't filter by namespace when using a non management token


jorgemarey:
### Nomad version
nomad v1.8.0

### Issue

When ACLs  are enabled and using a non management token, the namespace filter doesn't work correctly in the UI nor the API v1/jobs/statuses directly.

### Reproduction steps

1) nomad agent -dev -acl-enabled
2) nomad acl bootstrap
3) nomad acl policy apply reader policy.hcl
namespace "*" {
   policy = "read"
}
4) nomad namespace apply test
5) nomad namespace apply dev
6) nomad run example.nomad.hcl (on namespace test)
7) nomad run example.nomad.hcl (on namespace dev)
8) nomad acl token create -ttl 1h -policy=reader -name test

Try to filter using the UI or the API by namespace using the token created in step 8. 

#### Expected Result

- Only jobs of the filtered namespace should be shown

#### Actual Result

- All jobs for all allowed namespaces are shown.

tgross:
Hi @jorgemarey! Strangely I wasn't able to confirm this behavior with Nomad 1.8.0 or the current tip of `main`. I ran thru the reproduction steps you've got here and despite having read access to both namespaces, I only see the job in the `dev` namespace in the API response:

```
$ nomad acl token self
Accessor ID  = 259b0161-16b6-7f99-6b2d-1197ccd30868
Secret ID    = <redacted>
Name         = test
Type         = client
Global       = false
Create Time  = 2024-06-21 19:32:17.162311689 +0000 UTC
Expiry Time  = <none>
Create Index = 21
Modify Index = 21
Policies     = [reader]

Roles
<none>


$ nomad operator api '/v1/jobs/statuses?filter=Namespace%20%3D%3D%20dev' | jq .
[
  {
    "ID": "example",
    "Namespace": "dev",
    "Name": "example",
    "Type": "service",
    "NodePool": "default",
    "Datacenters": [
      "*"
    ],
    "Priority": 50,
    "Version": 0,
    "SubmitTime": 1718998345549213000,
    "ModifyIndex": 35,
    "Allocs": [
      {
        "ID": "4fd3d076-ad79-b95d-7d74-3e7d90225c1d",
        "Group": "group",
        "ClientStatus": "running",
        "NodeID": "2982de71-599c-87da-dff7-ab2adf39b206",
        "DeploymentStatus": {
          "Canary": false,
          "Healthy": true
        },
        "JobVersion": 0,
[...]

PR #23456, opened 2024-06-27
job statuses: fix filtering for namespace parameter

tgross:
The job statuses endpoint does not filter jobs by the namespace query parameter unless the user passes a management token. The RPC handler creates a filter based on all the allowed namespaces but improperly conditions reducing this down to only the requested set on there being a management token. Note this does not give the user access to jobs they shouldn't have, only ignores the parameter.

Remove the RPC handler's extra condition that prevents using the requested namespace. This is safe because we specifically check the ACL for that namespace earlier in the handler.

Fixes: https://github.com/hashicorp/nomad/issues/23370

### DIFF
--- a/.changelog/23456.txt
+++ b/.changelog/23456.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fixed support for namespace parameter on job statuses API
+```
--- a/nomad/job_endpoint_statuses.go
+++ b/nomad/job_endpoint_statuses.go
@@ -70,12 +70,10 @@ func (j *Job) Statuses(
 	} else if err != nil {
 		return err
 	}
-	// since the state index we're using doesn't include namespace,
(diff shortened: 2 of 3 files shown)


philrenaud (approved):
neat! Related: we've gotten a request to change the namespace dropdown (which only appears if you have any non-default namespaces, re: issue comment) from single-select radios to multi-select checkboxes, and was planning on using filter expressions to do so. Sounds like this makes that possible for non-mgmt, so thank you!

tgross:
> and was planning on using filter expressions to do so. Sounds like this makes that possible for non-mgmt, so thank you!

As it turns out, filter expressions already worked. It was only the namespace query parameter which was broken. Which was very confusing to me in the issue! :grin:

gulducat (approved):
lgtm!

Issue #23444, opened 2024-06-26
Scaling is possible during deployment with bad results


lattwood:
### Super Quick tl;dr;

Evaluations have a different sort order when sorting by CreateIndex vs CreateTime, and this has impacts on deployment of jobs that do scaling.

### Nomad version
```
~# nomad version
Nomad v1.7.7
BuildDate 2024-04-16T19:26:43Z
Revision 0f34c85ee63f6472bd2db1e2487611f4b176c70c
```

### Operating system and Environment details
Relevant- we're using a shell script to query nomad, and then run various `nomad job scale` commands. The shell script also checks to see if a deployment is already in progress prior to scaling, but as you'll soon see due to the timeframes involved it wouldn't have helped.

### Issue
When performing a deployment of a job that is constantly being scaled, instead of getting a failure to scale and the `job scaling blocked due to active deployment` error from Nomad, spooky things happen.

This is the spooky thing.

```
==> 2024-06-26T15:07:49Z: Monitoring evaluation "46e24c94"
    2024-06-26T15:07:49Z: Evaluation triggered by job "JOB-NAME"
    2024-06-26T15:07:50Z: Evaluation status changed: "pending" -> "canceled"
==> 2024-06-26T15:07:50Z: Evaluation "46e24c94" finished with status "canceled"
```

I go and take a look at the evaluation it created.

```
> nomad eval status -json 46e24c94
{
    "AnnotatePlan": false,
    "BlockedEval": "",
    "ClassEligibility": null,
    "CreateIndex": 87987021,
    "CreateTime": 1719414468345035507,
    "DeploymentID": "",
    "EscapedComputedClass": false,
    "FailedTGAllocs": null,
    "ID": "46e24c94-b015-0f3e-4374-ae5ffe59d202",
    "JobID": "JOB-NAME",
    "JobModifyIndex": 87987021,
    "ModifyIndex": 87987127,
    "ModifyTime": 1719414469680946947,
    "Namespace": "default",
    "NextEval": "",
    "NodeID": "",
    "NodeModifyIndex": 0,
    "PreviousEval": "",
    "Priority": 50,
    "QueuedAllocations": null,
    "QuotaLimitReached": "",
    "RelatedEvals": null,
    "SnapshotIndex": 0,
    "Status": "canceled",
    "StatusDescription": "canceled after more recent eval was processed",
    "TriggeredBy": "job-register",
    "Type": "service",
    "Wait": 0,
    "WaitUntil": null
}
```

[...]

PR #23457, opened 2024-06-27
scale: add `-check-index` to `job scale` command

tgross:
The RPC handler for scaling a job passes flags to enforce the job modify index is unchanged when it makes the write to Raft. But its only checking against the existing job modify index at the time the RPC handler snapshots the state store, so it can only enforce consistency for its own validation.

In clusters with automated scaling, it would be useful to expose the enforce index options to the API, so that cluster admins can enforce that scaling only happens when the job state is consistent with a state they've previously seen in other API calls. Add this option to the CLI and API and have the RPC handler check them if asked.

Fixes: https://github.com/hashicorp/nomad/issues/23444

### DIFF
--- a/command/job_scale.go
+++ b/command/job_scale.go
@@ -161,7 +161,6 @@ func (j *JobScaleCommand) Run(args []string) int {
 		},
 		Message:        msg,
 		PolicyOverride: false,
-		EnforceIndex:   checkIndex > 0,
 		JobModifyIndex: checkIndex,
 	}
 
--- a/command/job_scale.go
+++ b/command/job_scale.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/nomad/api"
+	"github.com/hashicorp/nomad/helper/pointer"
 	"github.com/mitchellh/cli"
 	"github.com/posener/complete"
 )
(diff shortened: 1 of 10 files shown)


command/job_scale.go
```diff
@@ -144,7 +153,19 @@ func (j *JobScaleCommand) Run(args []string) int {
...
+	req := &api.ScalingRequest{
+		Count: pointer.Of(int64(count)),
+		Target: map[string]string{
+			"Job":   jobID,
+			"Group": groupString,
+		},
+		Message:        msg,
+		PolicyOverride: false,
+		EnforceIndex:   checkIndex > 0,
+		JobModifyIndex: checkIndex,
```
gulducat, line 165: why the separate `EnforceIndex` bool, when it's derived anyway?  `job_endpoint.go` below has `args.JobModifyIndex != 0` already -- am I missing what we gain by having the extra bool?

gulducat (approved):
LGTM!  personally, I might drop the `EnforceIndex` bool, but not a blocker.

command/job_scale.go
tgross: That's a good point... I was being consistent with the job register API where having both flags does matter. The Raft log we write is the `JobRegister` log entry here anyways. Will fix.

PR #23458, opened 2024-06-27
[ui] Read-based checks for viewing templates and write-based checks for saving them

### LINKED ISSUE #23438: UI: wrong policy ACLs when access Job Templates/Nomad Variables
### Nomad version
Nomad v1.8.1
BuildDate 2024-06-19T06:43:57Z
Revision 5022543e4b7b8dcec9df123f86630ae3fdcffbe6

RHEL 8.6 - RHEL 8.8

### Issue

We try to give users access to some Job Templates via UI. 
As I understand the templates (beside the default ones) are stored as Nomad Variables. So I added the following ACLs to the user policy:

```
namespace "default" {
  policy = "read"
  variables {
    # give read access to all job templates related to this namespace
    path "nomad/job-templates/*" {
      capabilities = ["read","list"]
    }
    # needed at least once specified to get access to "template" buttons in UI
    # but because of read/list above user cannot write/create new templates! (this has to be the second rule!!!)
    path "nomad/job-templates/notexisting" {
      capabilities = ["write"]
    }
  }
}
```

with this Policy read-only access works as expected. But without the second write-capability, the "choose template/upload template" Button when trying to "Run a Job" is gone.
As my policy suggest, the path of the write-capability is not existing. So it looks like the check for the Buttonvisibility only looks for the presence of any write-capability. 
I've also checked with different namespaces ... I need the write-capability only once a namespace, and then it works as expected.

If I check the "Variables" page, I could see the "nomad/job-templates" path as expected, and also access the stored custom template values. So it might only be a problem with the ui-buttons, and not with the underlying permissions checks.

### Reproduction steps

- Create a Nomad Policy with only read/list access to Nomad Variables (or one without any Variable permissions)
- Run a Job and try to load a job template

-> no Choose template button accessible

#### Expected Result

"Choose Template" Button should be available also with read-only permissions
 
And maybe access to default templates should be  possible without access to nomad Variables?

#### Actual Result

Button is only useable if at least one write permission is granted

### Job file (if appropriate)
---
---
---

thanks,
Mac

philrenaud:
This adds nuance to the ACL policy checks for a user to be able to see Job Templates (which are specifically-named Nomad Variables behind the scenes).

The way it works today has some specific checks in place:
- to even get into the job templates section, you have to first click the "Run Job" button, which only exists if you have the `submit-job` ACL capability, or a general `"write"` policy.
- Once there, we only show you a "Choose from template" button if you can write variables to `*`.

I think the original intent here was to do up-front checks for full CRUD access, but this PR moves that access check a little downstream. Here, you'll see that the check to see the "Choose from template" button requires specifically `read variable` access (and I could be convinced to make it `list variables`, but that would make the subsequent page pretty useless), and the `write variable` check is moved downstream to the "save as template" button, and to manually edit a given job template via the Management page.

Here is a sample ACL policy to test with:

```
namespace "default" {
  policy = "read"
  capabilities = ["submit-job"]
  variables {
    # give read access to all job templates related to this namespace
    path "nomad/job-templates/*" {
      capabilities = ["read","list","write"]
    }
    path "*" {
      capabilities = ["write","list"]
    }
  }
}
```
^--- the `submit-job` capability is in place, so the "Run Job" button will be present. Also, the `nomad/job-templates/*` rule has both `read` and `write`, which will show both a "Choose from template" and a "Save as template" button. Remove either read or write to see how buttons disappear accordingly.

### User has read and write, but not delete, variable capability:
image

### User has `read variables` for `nomad/job-templates/*`:
image

### User has `write variables` for `nomad/job-templates/*`:
image

Resolves #23438

### DIFF
--- a/ui/app/templates/components/job-editor.hbs
+++ b/ui/app/templates/components/job-editor.hbs
@@ -16,7 +16,7 @@
           Paste or author HCL or JSON to submit to your cluster, or select from a list of templates. A plan will be requested before the job is submitted. You can also attach a job spec by uploading a job file or dragging &amp; dropping a file to the editor.
         </p>
 
-        {{#if (can "write variable" path="*" namespace="*")}}
+        {{#if (can "read variable" path="nomad/job-templates/*" namespace="*")}}
           <Hds::ButtonSet>
             <label
               class="job-spec-upload hds-button hds-button--color-secondary hds-button--size-medium"
--- a/.changelog/23458.txt
+++ b/.changelog/23458.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fix an issue where access to Job Templates in the UI was restricted to variable.write access
+```
--- a/ui/app/templates/components/job-editor/edit.hbs
+++ b/ui/app/templates/components/job-editor/edit.hbs
@@ -101,14 +101,16 @@
     data-test-plan
     @text="Plan"
   />
-  {{#if @data.job.isNew}}
-    <Hds::Button
-      @text="Save as template"
-      @color="secondary"
-      @route="jobs.run.templates.new"
-      {{on "click" @fns.onSaveAs}}
-      data-test-save-as-template
-    />
+  {{#if (can "write variable" path="nomad/job-templates/*" namespace="*")}}
+    {{#if @data.job.isNew}}
+      <Hds::Button
+        @text="Save as template"
+        @color="secondary"
+        @route="jobs.run.templates.new"
+        {{on "click" @fns.onSaveAs}}
+        data-test-save-as-template
+      />
+    {{/if}}
   {{/if}}
   <Hds::Button
     @text="Save as .nomad.hcl"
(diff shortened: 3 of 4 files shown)


ui/app/templates/components/job-editor.hbs
```diff
@@ -16,7 +16,7 @@
           Paste or author HCL or JSON to submit to your cluster, or select from a list of templates. A plan will be requested before the job is submitted. You can also attach a job spec by uploading a job file or dragging &amp; dropping a file to the editor.
         </p>
 
-        {{#if (can "write variable" path="*" namespace="*")}}
+        {{#if (can "read variable" path="nomad/job-templates/*" namespace="*")}}
```
tgross, line 19: At first glance I thought this would get us into trouble, but the `write` policy for Variables is a little different than it is for anything else:
> Create or update Variables at this path. Includes the "list" capability but not the "read" or "destroy" capabilities.

:+1:

tgross (approved):
LGTM! :+1:

martisah (approved):

Issue #23250, opened 2024-06-06
qemu job failed at "rpc error: code = Unknown desc = unable to configur e cgroups: no such file or directory"


joechchen:
### Nomad version

Nomad v1.8.0
BuildDate 2024-05-28T17:38:17Z
Revision 28b82e4b2259fae5a62e2ed47395334bea5a24c4

### Operating system and Environment details
Ubuntu 22.04.4 LTS
### Issue

qemu job failed at:
```
...
2024-06-06T13:45:20.490-0400 [ERROR] client.alloc_runner.task_runner: running driver failed: alloc_id=40bd6120-39b6-5f19-7416-322de0c20ad5 task=freedos error="rpc error: code = Unknown desc = unable to configur
e cgroups: no such file or directory"
...
```
The same job runs fine on v1.7.1 and prior.

### Reproduction steps
```
nomad run freedos.nomad
nomad monitor -log-level=DEBUG
```
#### Expected Result

The job should run.

#### Actual Result

The job failed.

### Job file (if appropriate)
```
job "freedos" {
  datacenters = ["dc1"]

  group "g1" {
    task "freedos" {
      resources {
        cpu    = 500
        memory = 256
      }

      artifact {
        source      = "https://github.com/angrycub/nomad_example_jobs/raw/main/applications/vms/freedos/freedos.img.tgz"
        destination = "local"
        options {
          checksum  = "sha256:8d2817126bf46ba2b4fca0b0c49eed2cc208c6f6448651e82c6d973fcba36569"
        }
      }

      driver = "qemu"

      config {
        image_path  = "local/freedos.img"
        accelerator = "kvm"
        args = []
      }
    }
  }
}
```

### Nomad Client logs (if appropriate)
```
2024-06-06T13:45:20.436-0400 [DEBUG] client.artifact: sub-process: OUTPUT="artifact download was a success"
2024-06-06T13:45:20.436-0400 [TRACE] client.alloc_runner.task_runner.task_hook.procisolation: setting up client process management: alloc_id=40bd6120-39b6-5f19-7416-322de0c20ad5 task=freedos task=40bd6120/freed
os
2024-06-06T13:45:20.437-0400 [TRACE] client.proclib: setup client process management: task=40bd6120/freedos
2024-06-06T13:45:20.437-0400 [TRACE] client.proclib: initialize cgroup: task=40bd6120/freedos
[...]

Issue #19090, opened 2023-11-14
`qemu` doesn't support NUMA-aware scheduling


tgross:
As noted in https://github.com/hashicorp/nomad/pull/19089, the `qemu` driver uses our universal executor to run the QEMU command line tool. Because QEMU owns the resource isolation, we don't pass in the resource block that the universal executor uses to configure cgroups and core pinning. This prevents NUMA-aware scheduling from working properly for the `qemu` driver.

We currently support `cores` by mapping that to a `-smp` flag. It looks like QEMU supports NUMA-awareness with the `-numa` flag, so we should be able to map Nomad's NUMA-aware scheduling to the appropriate flags too.

PR #23466, opened 2024-06-28
qemu: pass task resources into driver for cgroup setup

tgross:
As part of the work for 1.7.0 we moved portions of the task cgroup setup down into the executor. This requires that the executor constructor get the `TaskConfig.Resources` struct, and this was missing from the `qemu` driver. We fixed a panic caused by this change in #19089 before we shipped, but this fix was effectively undo after we added plumbing for custom cgroups for `raw_exec` in 1.8.0. As a result, running `qemu` tasks always fail on Linux.

This was undetected in testing because our CI environment doesn't have QEMU installed. I've got all the unit tests running locally again and have added QEMU installation when we're running the drivers tests.

Fixes: https://github.com/hashicorp/nomad/issues/23250


tgross:
(Holding this pending GitHub Actions getting unstuck so we can run the tests.)

pkazmierczak (approved):
LGTM!

djthorpe:
FYI this is also happening to me using the raw_exec driver on nomad 1.8.1

Here is the output in nomad 1.8.1:

```
Jul 15, '24 13:00:39 +0200 | Driver Failure | failed to launch command with executor: rpc error: code = Unknown desc = unable to configure cgroups: no such file or directory
```

Here is the snippet from my job:
```
    task "init" {
      driver = "raw_exec"

      lifecycle {
        sidecar = false
        hook    = "prestart"
      }

      config {
        // Set permissions on the directory
        command = var.data == "" ? "/usr/bin/echo" : "/usr/bin/install"
        args = compact([
          "-d", var.data,
          "-o", "472"
        ])
      }
    } // task "init"
```

mismithhisler:
I believe this fixes [GH #19090](https://github.com/hashicorp/nomad/issues/19090)

Issue #23445, opened 2024-06-26
Unable to filter by node pool after update to v1.8.1


gscho:
### Nomad version
v1.8.1

### Issue

After updating to version 1.8.1 I can no longer filter by node pool in the nomad UI. When I select any node pool I see an error like this:

```
No jobs match your current filter selection: (NodePool == windows-amd64-instance-production). failed to create result paginator: failed to read filter expression: 1:21 (20): rule "grouping": Unmatched parentheses
```

Note this is only for node pools and all other filters I've tried seem to work fine.

### Reproduction steps

Select a node pool filter from the drop down or search for `(NodePool == my-pool)` in the search bar.

thefallentree:
we ran into same issue, and it's because the filter code doesn't work with "-"  in the pool name,   the JS needs to quote the string

tgross:
Thanks for the debugging there @thefallentree! I'll mark this for the attention or our UI folks.

philrenaud:
This should be fixed by a change in #23468 — thanks for opening the issue and for your patience!

PR #23468, opened 2024-06-28
[ui] Generalized namespace handling, generalized facet searching, node pools facet search

philrenaud:
Changes the jobs index page's Namespaces dropdown from single select (radio buttons) to checkboxes (multiple OR-based selection)

image

Also fixes an error noted in #23445  where node pools and namespaces with `-` characters would not filter correctly

Resolves NET-10159

### DIFF
--- a/ui/app/controllers/jobs/index.js
+++ b/ui/app/controllers/jobs/index.js
@@ -32,19 +32,12 @@ export default class JobsIndexController extends Controller {
     this.rawSearchText = this.searchText || '';
   }
 
-  queryParams = [
-    'cursorAt',
-    'pageSize',
-    { qpNamespace: 'namespace' },
-    'filter',
-  ];
+  queryParams = ['cursorAt', 'pageSize', 'filter'];
 
   isForbidden = false;
 
   @tracked jobQueryIndex = 0;
   @tracked jobAllocsQueryIndex = 0;
-
-  @tracked qpNamespace = '*';
 
   get tableColumns() {
     return [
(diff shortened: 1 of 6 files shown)


david-yu:
This looks beautiful! LGTM. I won't approve because that is reserved for engineering!

ui/app/controllers/jobs/index.js
```diff
@@ -202,7 +195,7 @@ export default class JobsIndexController extends Controller {
    * @param {Error} e
    */
   notifyFetchError(e) {
-    const firstError = e.errors[0];
+    const firstError = e.errors?.objectAt(0);
```
philrenaud, line 198: I should've caught this in https://github.com/hashicorp/nomad/pull/23427, but handling it here. Defensive logic in case there's an errors-less error (driving through a tunnel, etc.)

ui/app/controllers/jobs/index.js
```diff
@@ -433,40 +426,38 @@ export default class JobsIndexController extends Controller {
     label: 'NodePool',
     options: (this.model.nodePools || []).map((nodePool) => ({
       key: nodePool.name,
-      string: `NodePool == ${nodePool.name}`,
+      string: `NodePool == "${nodePool.name}"`,
```
philrenaud, line 429: Fixes an issue reported in https://github.com/hashicorp/nomad/issues/23445

pkazmierczak (approved):
LGTM!

PR #23477, opened 2024-07-01
cli: fix bold output of devices headers

shoenig:


shoenig:
before:
Screenshot from 2024-07-01 10-32-29


after: 
Screenshot from 2024-07-01 10-30-58

pkazmierczak (approved):
nice!

PR #23479, opened 2024-07-01
feat: allow role updates to token using cli

burnerlee:
This PR adds the following flag functionality to `nomad acl token update` to allow users to update token roles:
1. `-role-id`
2. `role-name`

Fixes: #18354

### DIFF
--- a/command/acl_token_update.go
+++ b/command/acl_token_update.go
@@ -12,6 +12,9 @@ import (
 
 type ACLTokenUpdateCommand struct {
 	Meta
+
+	roleNames []string
+	roleIDs   []string
 }
 
 func (c *ACLTokenUpdateCommand) Help() string {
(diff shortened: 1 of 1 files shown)


pkazmierczak (changes requested):
Hey @burnerlee, this looks great, thanks a lot for contributing! I left some comments, but nothing major. 

Would you also add a changelog message? Something like:
```
release-note:improvement
cli: Allow role updates to token
```
(you can use `make cl` command to generate it)

pkazmierczak:
ah one more thing: could you update the documentation to include these new flags, too? https://github.com/hashicorp/nomad/blob/main/website/content/docs/commands/acl/token/update.mdx

tgross:
Sorry, it looks like we had a PR in-flight for this which got abandoned, and I didn't notice that we had this open when I carried that PR. I'm going to close this in lieu of https://github.com/hashicorp/nomad/pull/18532

PR #23489, opened 2024-07-02
mbits: filter mbits out of diff

martisah:
Currently, the MBits field is not filtered and therefore always shown in diffs (regardless of MBits even being defined in an old or new network resource), and after finding that it is a deprecated field, this pr adds MBits to the list of filtered fields so it will not be shown as a diff.


tgross (approved):
LGTM!

Issue #22230, opened 2024-05-23
Nomad client's Vault client gets stuck on the wrong Vault namespace


the-nando:
### Nomad version
```
Nomad v1.7.7+ent
BuildDate 2024-04-16T19:58:29Z
Revision d0c2458a9c62314d727c3516e63d92026639cc89
```

### Issue

I've run into an odd issue which I'm trying to make sense of, hopefully someone can help.  
If I submit a job with a namespace set in the Vault stanza which fails to deploy, for instance because the role specified doesn't exist, the value of the namespace seems to persist across job versions, i.e. I'm unable to set `namespace = ""`. Interestingly enough the value seems to persist even across job purges.

### Reproduction steps
Run the following job:
```
job "test" {
  region = "region-1"
  datacenters = ["region-1a"]
  type        = "service"
  namespace = "default"

  group "nginx" {
    count = 1

     vault {
       namespace = ""
       cluster = "default"
       role = "nomad-workloads-error" 
      }

    task "nginx" {
      driver = "docker"

      config {
        image = "alpine:3.18.5"
        command = "/bin/sh"
        args = ["-c", "while true ; do echo sleeping; sleep 1 ; done"]
      }

      resources {
        cpu    = 301
        memory = 128
      }
    }
  }
}
```

Error in the client logs as expected, since the role doesn't exist:
```
  | Vault: failed to derive vault token: failed to derive Vault token for identity vault_default: failed to login with JWT: Error making API request.
  |
  | URL: PUT http://localhost:8200/v1/auth/nomad-region-1/login
  | Code: 400. Errors:
  |
  | * role "nomad-workloads-error" could not be found
```

Update the job spec and resubmit the job:
```
vault {
 namespace = "foobar"
 cluster = "default"
 role = "nomad-workloads-error" 
}
```

Error as expected:
```
  | Vault: failed to derive vault token: failed to derive Vault token for identity vault_default: failed to login with JWT: Error making API request.
  |
  | Namespace: foobar
  | URL: PUT http://localhost:8200/v1/auth/nomad-region-1/login
  | Code: 400. Errors:
  |
  | * role "nomad-workloads-error" could not be found
```

Stop and purge the job:
```
nomad job stop -purge test
```

Update the job spec and re-submit:
```
vault {
  namespace = ""
  cluster = "default"
  role = "nomad-workloads-error" 
}
```

Error:
```
[...]

PR #23491, opened 2024-07-02
vault: fix namespace reset for clients with unset namespace

tgross:
The Vault "logical" API doesn't allow configuring the namespace on a per-request basis. Instead, it's set on the client. Our `vaultclient` wrapper locks access to the API client and sets the namespace (and token, if applicable) for each request, and then resets the namespace and unlocks the API client.

The logic for resetting the namespace incorrectly assumed that if the Vault configuration didn't set the namespace that it was canonicalized to the non-empty string `"default"`. This results in the API client's namespace getting "stuck" whenever a job uses a non-default namespace if the configuration value is empty. Update the logic to always go back to the configuration, rather than accepting the "previous" namespace from the caller.

This changeset also removes some long-dead code in the Vault client wrapper.

Fixes: https://github.com/hashicorp/nomad/issues/22230
Ref: https://hashicorp.atlassian.net/browse/NET-10207


pkazmierczak (approved):
LGTM!

PR #23492, opened 2024-07-02
[ui] When a job is deleted in the background, wait until redirect before cache unload

### LINKED ISSUE #23410: UI: cannot close task sidebar if Job was deleted
### Nomad version
```
v1.8.0
```

### Issue
image

### Reproduction steps
Open "View logs". Delete the job.

#### Expected Result
I should be able to close the window even if underlying task does not exists already.

#### Actual Result
I am not able to close the window.

philrenaud:
This prevents an issue where, still on the job page, the cache-removed job would have trouble accessing properties and relationships of its now-removed Ember Data object.

Resolves #23410

### DIFF
--- a/.changelog/23492.txt
+++ b/.changelog/23492.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fix an issue where a remotely purged job would prevent redirect from taking place in the web UI
+```
--- a/ui/app/controllers/jobs/job.js
+++ b/ui/app/controllers/jobs/job.js
@@ -23,7 +23,7 @@ export default class JobController extends Controller {
     return this.model;
   }
 
-  @action notFoundJobHandler() {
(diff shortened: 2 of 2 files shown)


pkazmierczak (approved):
LGTM!

tgross (commented):
Seems like a reasonable change, but it looks like it broke a related test.

philrenaud:
Broken test is now fixed; was related to a [core piece of Ember's transition-handling](https://github.com/emberjs/ember-test-helpers/issues/332) where a transition can happen at both a high level (the application route has one) and a low level (the redirect we're patching up here) and the former is left "floating". A try/catch block prevents the `uncaught` failure.

tgross (approved):

martisah (approved):

PR #23503, opened 2024-07-03
[action] Update actions in test-ui workflow

angrycub:
Uses a deprecated node 16 action


philrenaud (approved):

angrycub:
This will need to be backported to the usual suspects.

PR #23506, opened 2024-07-04
[ui] Global token CRUD in the web UI

### LINKED ISSUE #23429: UI: allow creating global ACL tokens
### Nomad version
Output from `nomad version`
Nomad v1.7.3
BuildDate 2024-01-15T16:55:40Z
Revision 60ee328f97d19d2d2d9761251b895b06d82eb1a1

Ubuntu 22.04

### Issue
We recently established nomad federation cluster. (with replication and it works!) 
When i try to change client token from non-global to global (because some users should see other cluster) - i see following error:

``` bash
nomad acl token update -global=true -type=client fa6ebc10-a8dd-bfd6-1667-2837b415d631
Error updating token: Unexpected response code: 400 (token 0 invalid: 1 error occurred:
* cannot toggle global mode
```
#### Expected Result
I could switch client mode from non-global to global. 
Or at least **add option into UI (like checkbox) - should token be global or not.** Most tokens created via UI and it`s very inconvenient when you switched to federation recognize that now you could create some tokens only via nomad server cli interaction.   

#### Actual Result
Have not any UI option in FED mode about token global or not. 
Can't switch token type to global via cli - so need to recreate all tokens what should see two clusters.

philrenaud:
This PR allows for creation of global tokens, tokens of differing regions, and generally improves the handling of regions in the UI a bit better.

image

Resolves #23429

### DIFF
--- a/ui/app/components/region-switcher.js
+++ b/ui/app/components/region-switcher.js
@@ -22,7 +22,7 @@ export default class RegionSwitcher extends Component {
 
   async gotoRegion(region) {
     // Reset token
-    this.token.reset();
+    // this.token.reset();
     // Note: redundant but as long as we're using PowerSelect, the implicit set('activeRegion')
     // is not something we can await, so we do it explicitly here.
     this.system.set('activeRegion', region);
--- a/ui/app/components/region-switcher.js
+++ b/ui/app/components/region-switcher.js
@@ -13,16 +13,22 @@ export default class RegionSwitcher extends Component {
   @service system;
   @service router;
   @service store;
+  @service token;
 
   @computed('system.regions')
   get sortedRegions() {
     return this.get('system.regions').toArray().sort();
   }
 
-  gotoRegion(region) {
-    this.router.transitionTo('jobs', {
-      queryParams: { region },
-    });
+  async gotoRegion(region) {
+    // Reset token
+    this.token.reset();
+    // Note: redundant but as long as we're using PowerSelect, the implicit set('activeRegion')
+    // is not something we can await, so we do it explicitly here.
+    this.system.set('activeRegion', region);
+    await this.get('token.fetchSelfTokenAndPolicies').perform().catch();
+
+    this.router.transitionTo({ queryParams: { region } });
   }
 
   get keyCommands() {
--- a/ui/app/components/region-switcher.js
+++ b/ui/app/components/region-switcher.js
@@ -21,8 +21,6 @@ export default class RegionSwitcher extends Component {
   }
 
   async gotoRegion(region) {
-    // Reset token
-    // this.token.reset();
     // Note: redundant but as long as we're using PowerSelect, the implicit set('activeRegion')
     // is not something we can await, so we do it explicitly here.
     this.system.set('activeRegion', region);
--- a/ui/app/components/token-editor.js
+++ b/ui/app/components/token-editor.js
@@ -24,6 +24,14 @@ export default class TokenEditorComponent extends Component {
   @tracked tokenPolicies = [];
   @tracked tokenRoles = [];
 
+  /**
+   * When creating a token, it can be made global (has access to all regions),
+   * or non-global. If it's non-global, it can be scoped to a specific region.
+   * By default, the token is created in the active region of the UI.
+   * @type {string}
+   */
+  @tracked tokenRegion = '';
+
   // when this renders, set up tokenPolicies
   constructor() {
     super(...arguments);
@@ -32,6 +40,7 @@ export default class TokenEditorComponent extends Component {
     if (this.activeToken.isNew) {
       this.activeToken.expirationTTL = 'never';
     }
+    this.tokenRegion = this.system.activeRegion;
   }
 
   @action updateTokenPolicies(policy, event) {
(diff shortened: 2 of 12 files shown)


ui/app/components/token-editor.js
```diff
@@ -96,7 +134,15 @@ export default class TokenEditorComponent extends Component {
         this.activeToken.expirationTTL = null;
       }
 
-      await this.activeToken.save();
+      const adapterRegion = this.activeToken.global
+        ? this.system.get('defaultRegion.region')
+        : this.tokenRegion;
```
philrenaud, line 139: Note: even though the token is marked `Global: true`, I found that I still have to unset its region (or set it to the authoritative region), otherwise it gets created as global but within whatever region the user happens to be using at the time.

Once you're browsing the web UI with a region set, all requests get a queryParam of ?region=<region>. Thus you could end up with a token that, yes, is global, but is stored in the context of an agent that isn't global/authoritative.

As such it seems like you _can_ create a global, non-global token. Weird!

ui/app/controllers/jobs/index.js
```diff
@@ -240,7 +240,7 @@ export default class JobsIndexController extends Controller {
           console.log('error fetching job ids', e);
           this.notifyFetchError(e);
         }
-        if (this.jobs.length) {
+        if (this.jobs?.length) {
```
philrenaud, line 243: Fix that I should've included in https://github.com/hashicorp/nomad/pull/23427 — errors from outside of Nomad (which I was experiencing when trying to switch regions with an invalid token) didn't yield jobs so I'd get a "cannot get .length of undefined" error

ui/app/routes/clients.js
```diff
@@ -22,7 +22,8 @@ export default class ClientsRoute extends Route.extend(WithForbiddenState) {
   model() {
     return RSVP.hash({
       nodes: this.store.findAll('node'),
-      agents: this.store.findAll('agent'),
+      // TODO: needed?
+      // agents: this.store.findAll('agent'),
```
philrenaud, line 26: Note: I don't think this is a needed line for the clients index (anymore?).

As best as I can tell, non-authoritative-region management tokens do _not_ have access to the server members list (cannot do `nomad server members` either)

This hadn't been a concern in the UI before, but the following is a situation where it would be problematic:
- I am a management token for a non-authoritative region. I view the `servers` page and get a "not authorized" message. OK, that's fine, I'm not part of the global/authoritative region.
- I switch to the clients tab, which should at least show me the clients within the region that holds my token. However, since the model for this page has a check for server members (agents in Ember), this page also gives me a "not authorized" error.

This removal makes it so non-authoritative region tokens with `node list` abilities can still view this page.

ui/app/routes/clients.js
```diff
@@ -22,7 +22,8 @@ export default class ClientsRoute extends Route.extend(WithForbiddenState) {
   model() {
     return RSVP.hash({
       nodes: this.store.findAll('node'),
-      agents: this.store.findAll('agent'),
```
philrenaud, line 25: Note: I don't think this is a needed line for the clients index (anymore?).

As best as I can tell, non-authoritative-region management tokens do _not_ have access to the server members list (cannot do `nomad server members` either)

This hadn't been a concern in the UI before, but the following is a situation where it would be problematic:
- I am a management token for a non-authoritative region. I view the `servers` page and get a "not authorized" message. OK, that's fine, I'm not part of the global/authoritative region.
- I switch to the clients tab, which should at least show me the clients within the region that holds my token. However, since the model for this page has a check for server members (agents in Ember), this page also gives me a "not authorized" error.

This removal makes it so non-authoritative region tokens with `node list` abilities can still view this page.

ui/app/components/region-switcher.js
```diff
@@ -13,16 +13,22 @@ export default class RegionSwitcher extends Component {
...
-    });
+  async gotoRegion(region) {
+    // Reset token
+    // this.token.reset();
+    // Note: redundant but as long as we're using PowerSelect, the implicit set('activeRegion')
+    // is not something we can await, so we do it explicitly here.
+    this.system.set('activeRegion', region);
+    await this.get('token.fetchSelfTokenAndPolicies').perform().catch();
+
+    this.router.transitionTo({ queryParams: { region } });
```
philrenaud, line 31: Note: made the decision not to send you back to the jobs index when you switch regions anymore. This was a ~6 year old decision that made sense at the time, but we now have many more routes that may be context-shared between regions.

For example, if I sit on the /administration/policies page and switch between regions, I should now see my policies list simply update. Before, I'd have been jostled, have to click Administration, then have to click Policies again.

The drawback from this is that sometimes routes (like a particular task in a particular job) are NOT shared between regions. Switching a region on a page like this will put you on a (friendly) 404 page. This is a tradeoff that I think is worthwhile.

ui/app/routes/clients.js
tgross: > As best as I can tell, non-authoritative-region management tokens do not have access to the server members list (cannot do nomad server members either)

To close the loop on a Slack thread, the `Status.Members` RPC isn't forwarded between servers (much less regions), so a local management token can get the server members list but only when sending that request to its own region.

ui/app/components/region-switcher.js
tgross: That's reasonable, for sure. Another thing to consider is that all the administrative routes are cross-region (because the authoritative region owns all these objects), whereas everything else is single region. But having two different behaviors depending on context might be more confusing than it's worth.

ui/app/components/token-editor.js
tgross: > Thus you could end up with a token that, yes, is global, but is stored in the context of an agent that isn't global/authoritative.

Depending on what you mean by the "context" of an agent, that smells like a possible bug in the RPC handler. We have logic (ref [`acl_endpoint.go#L602-L611`](https://github.com/hashicorp/nomad/blob/v1.8.1/nomad/acl_endpoint.go#L602-L611)) to override the region flag if the token is global.

tgross (approved):
LGTM!

PR #14947, opened 2022-10-18
docs: move autoscaling `source` agent config

lgfa29:
Move the Autoscaler agent configuration `source` to the `policy` page since they are very closely related.

Also update all headers in this section so they follow the proper `h1 > h2 > h3 > ...` hierarchy.

Preview: https://nomad-qdd899fir-hashicorp.vercel.app/nomad/tools/autoscaling/agent/policy

PR #23524, opened 2024-07-09
docs: Autoscaling agent 404 and navigation fix

boruszak:
This PR removes a reference page's listing in the main navigation file, which caused a 404 and created a navigation quirk. 

It also adds a redirect. According to [this PR merged in 2022](https://github.com/hashicorp/nomad/pull/14947), the content of the `source` page is now part of the `policy` page.


tgross (approved):
LGTM, thanks @boruszak!

PR #23530, opened 2024-07-10
Improve the restrict-images Sentinel policy template

sofixa:
This PR improves the restrict-images Sentinel policy template to:

* only apply to Docker/Podman tasks
* to handle version numbers starting with v (instead of just starting with a number).
* better check for e.g. the default registry

The checks are now split in three - registry, image, and version.


philrenaud (approved):
Thank you for this!

PR #23538, opened 2024-07-10
cni: allow users to set CNI args in job spec

### LINKED ISSUE #16197: support CNI_ARGS parameters
Some CNI plugins support accepting arguments in addition to the plugin config in JSON. (See https://www.cni.dev/docs/conventions/). They typically are passed in the `CNI_ARGS` environment variable, because passing them in the JSON `args` field would mean needing separate files for each argument variant. In Nomad (and k8s for that matter), the JSON config isn't typically accessible to job submitters.

Some examples where this is useful:
* The [`static`](https://www.cni.dev/plugins/current/ipam/static/) plugin for static IP address assignment supports an address for the static IP and gateway, which are useful for cases where IPs are being assigned outside of Nomad.
* The [`tc-redirect-tap`](https://github.com/awslabs/tc-redirect-tap) plugin (used for Firecracker VMs) supports arguments for per-VM jailer users/groups and tap names.
* This likely helps more complex third-party plugins like Cilium as well, but I'm not as familiar with those use cases.

martisah:
This PR allows for users to set up CNI args in a job specification through the network block by adding a cni block and specifying an args field. 

Resolves #16197

### DIFF
--- a/client/allocrunner/networking_cni_test.go
+++ b/client/allocrunner/networking_cni_test.go
@@ -205,33 +205,13 @@ func TestCNI_cniToAllocNet_Invalid(t *testing.T) {
 	require.Nil(t, allocNet)
 }
 
-// TestCNI_addCustomCNIArgs fails if given a task group with a
-// network resource including CNI Args calling addCustomCNIArgs does not
-// result in the expected map of CNI Args created.
 func TestCNI_addCustomCNIArgs(t *testing.T) {
 	ci.Parallel(t)
 	cniArgs := map[string]string{
-		// CNI plugins are called one after the other with the same set of
-		// arguments. Passing IgnoreUnknown=true signals to plugins that they
-		// should ignore any arguments they don't understand
-		"IgnoreUnknown": "true",
+		"default": "yup",
 	}
-	alloc := mock.ConnectAlloc()
 
-	alloc.AllocatedResources.Shared.Networks = []*structs.NetworkResource{
-		{
-			Mode: "bridge",
-			CNI: &structs.CNIArgs{
-				Args: map[string]string{
-					"first_arg": "example",
-					"new_arg":   "example_2",
-				},
-			},
-		},
-	}
-	tg := alloc.Job.LookupTaskGroup(alloc.TaskGroup)
-	tg.Networks = []*structs.NetworkResource{{
-		Mode: "bridge",
+	network := []*structs.NetworkResource{{
 		CNI: &structs.CNIArgs{
 			Args: map[string]string{
 				"first_arg": "example",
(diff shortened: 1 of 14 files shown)


client/allocrunner/networking_cni_test.go
```diff
@@ -205,6 +205,48 @@ func TestCNI_cniToAllocNet_Invalid(t *testing.T) {
 	require.Nil(t, allocNet)
 }
 
+func TestCNI_addCustomCNIArgs(t *testing.T) {
```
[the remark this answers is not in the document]
martisah: Yes that sounds good!

nomad/structs/diff_test.go
```diff
@@ -5091,6 +5079,86 @@ func TestTaskGroupDiff(t *testing.T) {
 				},
 			},
 		},
+		{TestCase: "TaskGroup CNI edited",
+			Contextual: false,
+			Old: &TaskGroup{
+				Networks: Networks{
+					{
+						DNS: &DNSConfig{
+							Servers: []string{"1.1.1.1"},
```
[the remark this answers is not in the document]
martisah: I added a test case for CNIConfig added, edited and deleted, and tried to add a test specifically to codify the behavior you mentioned, though I'm not too sure about it so let me know your thoughts!

nomad/structs/diff.go
```diff
@@ -2706,6 +2710,21 @@ func (d *DNSConfig) Diff(other *DNSConfig, contextual bool) *ObjectDiff {
...
 }
 
+// Diff returns a diff of two CNIConfig structs
+func (d *CNIConfig) Diff(other *CNIConfig, contextual bool) *ObjectDiff {
+	if d.Equal(other) {
+		return nil
+	}
+	if d == nil {
+		d = &CNIConfig{}
+	}
```
gulducat, line 2720: if `d` is `nil`, then `d.Equal` would panic.  I don't think this method could even get run if `d` (`n.CNI` up above) was `nil`, so either that check can be thrown out, or at least moved up above `d.Equal`.

nomad/structs/diff_test.go
```diff
@@ -5221,6 +5221,251 @@ func TestTaskGroupDiff(t *testing.T) {
...
+										New:  "",
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+		{TestCase: "TaskGroup Networks edited behavior",
```
gulducat, line 5374: very nitpicky: "behavior" is not very descriptive -- I might say "editing Networks deletes and re-adds" or smth like that.

website/content/docs/job-specification/network.mdx
```diff
@@ -294,6 +304,24 @@ network {
...
+
+```hcl
+network {
+  mode = "cni/mynet"
+  port "http" {
+    to = 8080
+  }
+  cni {
+    args = {
+     "nomad.region" : "${node.region}"
```
gulducat, line 319: seeing the `.` here in the map key made me realize that people could include a `;` in either key or value, and accidentally cause themselves mysterious issues, since the CNI_ARGS env var ends up as a `;`-separated string.

I think we should prevent that from happening (add just a touch more validation).

gulducat: but side-note, seeing this example here with interpolation really excites the imagination!

gulducat (commented):
we are getting very close!  I have just a couple more comments, and we also should add a changelog, then let's un-draft the PR and get another set of eyes on it!

nomad/structs/cni_config.go
```diff
@@ -0,0 +1,30 @@
...
+	return &CNIConfig{
+		Args: newMap,
+	}
+}
+
+func (d *CNIConfig) Equal(o *CNIConfig) bool {
+	if d == nil || o == nil {
+		return d == o
+	}
+	return reflect.DeepEqual(d.Args, o.Args)
```
tgross, line 29: For maps of primitive types we can avoid reflection:

```suggestion
	return maps.Equal(d.Args, o.Args)
```

nomad/structs/cni_config.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func (d *CNIConfig) Copy() *CNIConfig {
+	if d == nil {
+		return nil
+	}
+	newMap := make(map[string]string)
+	for k, v := range d.Args {
+		newMap[k] = v
+	}
```
tgross, line 21: There's a `maps.Clone` helper for this.

gulducat (approved):
Huzzah!

nomad/structs/structs.go
```diff
@@ -7186,6 +7188,26 @@ func (tg *TaskGroup) validateNetworks() error {
...
+					cniArgKeys.Insert(k)
+				}
+				if strings.Contains(k, ";") {
+					err := fmt.Errorf("invalid ';' character in CNI arg key %q", k)
+					mErr.Errors = append(mErr.Errors, err)
+				}
+				if strings.Contains(v, ";") {
+					err := fmt.Errorf("invalid ';' character in CNI arg value %q", v)
+					mErr.Errors = append(mErr.Errors, err)
+				}
```
tgross, line 7208: This illegal character check feels ad hoc; are we certain that this is an actual attack vector for CNI plugins? Also we're interpolating the values on the client (although not the keys). If I were an evil job author, I could have a value that includes a client attribute with a `;` and bypass the check.

gulducat: my concern was more about ergonomics than security -- I wanted to protect job authors from doing something confusingly silly.  the semicolon here isn't dangerous like it might be when parsing something like a shell command or sql query, but if a user put a `;` in a key or value, then it could confuse plugins that read the args, since `CNI_ARGS` in the end is a `;`-separated string of `key=val`s.

e.g. given the key `"bad-test-key;is-bad"` the (common) library that the `bridge` plugin uses hits [this error](https://github.com/containernetworking/cni/blob/d5c71ad528129e306538f7ba5b796337c1898d81/pkg/types/args.go#L85) and the Nomad client logs

> [ERROR] client.alloc_runner: prerun failed: alloc_id=29822c8b-7192-150b-cc00-9a4e827384b0 error="pre-run hook \"network\" failed: failed to configure networking for alloc: failed to configure network: plugin type=\"bridge\" failed (add): ARGS: invalid pair \"bad-test-key\""

a similar `bad;value` produces similar results.

any other code (plugin, library) that parses `CNI_ARGS` would have to do its own error handling like that, and could potentially produce dangerous results (we can't control the entire CNI ecosystem), but the goal here is just to error way earlier, before the job is accepted, to save the user some headache and troubleshooting.

the idea of a client attribute with a `;` is interesting, but if you (the user) are considering client stuff like that, I suppose I'm more comfortable with you needing to hunt down client logs to track down the error?

all that said, to your point in slack, we'll put something like this comment in the code, too, to clarify the intent and danger level (low).

PR #23539, opened 2024-07-10
[ui] Agent config and User localStorage defaults for filters

### LINKED ISSUE #11970: UI: Make the default namespace filter configurable
### Proposal

In #10871 the default namespace filter was changed to "*".  Make it instead configurable to only one namespace, or a wildcard or just leave it as default by default.

### Use-cases

The way we use nomad is that we create hundreds of parameterized jobs which we dispatch against.  These jobs are not interesting and live in a namespace `forecast`.  When an operator logs into nomad, they want to see the services and infrastructure jobs running in the `default` namespace.

Because the way parameterized jobs are implemented, loading all namespaces DDOSes the interface as it fetches thousands and thousands of jobs/dispatches.  This makes working with nomad UI a bit annoying as in the beginning we have to try to switch the namespace super fast :D 

### Attempted Solutions

For now we are just loading the UI with `?namespace=default` manually added to URL.  This is not optimal.

### LINKED ISSUE #20482: [ui] Agent config for UI: Default page size, Live Updates Enabled
In the UI block of the agent config for Nomad, allow for an admin operator to set a default page size (for jobs, clients, etc.) which would otherwise default to 25, and to enable/disable live updates for the jobs index page, which would otherwise default to `true`.

philrenaud:
TODO: Add Nomad Variables as one of the ways that user preferences can get read.

Order should be:
1. config from the user agent's ui.defaults block
2. defaults from a nomad variable at `nomad/ui/defaults/token_ID` or something similar
3. defaults from localStorage, set from a form in the UI
4. manually editing the page in question (filtering by namespace on the jobs page for example, which doesn't re-set your defaults but does override them for that session)

resolves https://github.com/hashicorp/nomad/issues/11970

### DIFF
--- a/nomad/structs/config/ui.go
+++ b/nomad/structs/config/ui.go
@@ -28,6 +28,9 @@ type UIConfig struct {
 
 	// Label configures UI label styles
 	Label *LabelUIConfig `hcl:"label"`
+
+	// Defaults
+	Defaults *DefaultsUIConfig `hcl:"defaults"`
 }
 
 // only covers the elements of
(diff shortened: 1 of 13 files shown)


philrenaud:
This was an attempt to solve https://github.com/hashicorp/nomad/issues/11970, but the number of ways it could/should be handled got out of hand quickly. A simpler approach in the future might be to just use agent.ui config.

PR #5602, opened 2019-04-23
Normalize plan to increase the plan apply throughput

arshjohar:
This PR adds normalization of the plan to commit only the diff for stopped and preempted allocs to the raft log to enable better throughput. It also starts using omitempty on some of the structs during msgpack serialization to omit the empty fields.

PR #23540, opened 2024-07-10
msgpack: omit empty fields on NetworkResource

gulducat:
the ratio of optimized/unoptimized log size in TestPlanNormalize has been increased several times as people have added to various structs and coincidentally bumped into the magic limit.

we encountered another such case when adding to NetworkResource, but here we omitempty on the struct instead of bumping the limit in the test.

this has the added benefit of reducing the serialized struct size! which was the original intent (#5602) behind this test in the first place :P the actual value of the ratio is now 0.628... but here the test value is only dropped down to 0.66 to leave some wiggle room.

### DIFF
--- a/nomad/structs/structs_test.go
+++ b/nomad/structs/structs_test.go
@@ -5188,22 +5188,26 @@ func TestPlan_AppendPreemptedAllocAppendsAllocWithUpdatedAttrs(t *testing.T) {
 	assert.Equal(t, expectedAlloc, appendedAlloc)
 }
 
-func TestAllocation_MsgPackTags(t *testing.T) {
+func TestMsgPackTags(t *testing.T) {
 	ci.Parallel(t)
-	planType := reflect.TypeOf(Allocation{})
 
-	msgPackTags, _ := planType.FieldByName("_struct")
+	cases := []struct {
+		name   string
+		typeOf reflect.Type
+	}{
+		{"Allocation", reflect.TypeOf(Allocation{})},
+		{"Evaluation", reflect.TypeOf(Evaluation{})},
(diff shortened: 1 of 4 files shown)


tgross (commented):
The idea here seems solid but it looks like this is breaking some tests that assume empty fields aren't omitted.

nomad/structs/structs_test.go
```diff
@@ -5188,22 +5188,26 @@ func TestPlan_AppendPreemptedAllocAppendsAllocWithUpdatedAttrs(t *testing.T) {
 	assert.Equal(t, expectedAlloc, appendedAlloc)
 }
 
-func TestAllocation_MsgPackTags(t *testing.T) {
+func TestMsgPackTags(t *testing.T) {
```
tgross, line 5191: Non-blocking: I realize you're just expanding a test we've already got, but I sort of wonder at the utility of this test. There's no logic under test, so isn't this basically just asserting that the struct is how we defined it?

tgross (approved):
LGTM! :+1:

nomad/structs/structs_test.go
gulducat: yeah I thought it was a little strange, but I assume it was added initially as a regression test, to avoid someone mistakenly deleting the field, or perhaps typoing the otherwise not-checked-by-anything magic string?

it also serves as a central inventory of structs with this tag (if the list is kept up to date), but a pretty simple grep can find that, too (which is how I found the previously untested ones).

so, meh, these almost-good reasons in aggregate were enough for me to keep the test around. :shrug:

tgross: Works for me. It's a very cheap test anyways.

PR #23553, opened 2024-07-11
[ui] Fix: Actions should pass job id instead of job name

### LINKED ISSUE #23464: Nomad action from ui with job name does not work
### Nomad version
Nomad v1.8.0
BuildDate 2024-05-28T17:38:17Z
Revision 28b82e4b2259fae5a62e2ed47395334bea5a24c4

74-Ubuntu SMP Wed Feb 22 14:14:39 UTC 2023

### Issue
When you add a name stanza in your job, actions does not work when launched from ui
I guess that this should be job.get('id') and instead of this [job.get('name')](https://github.com/hashicorp/nomad/blob/cd3101d624f24a6c35bf7c2363b34b79e71ac7eb/ui/app/adapters/job.js#L195)
You can try to command the second line of the job in exemple to see the different behaviours

### Reproduction steps
Post job with name stanza, with an action, then run action from ui

#### Expected Result
You should see an output
It should tell `OCI runtime exec failed: exec failed: unable to start container process: exec: "ls": executable file not found in $PATH: unknown` with exemple job, but it is not what matters

#### Actual Result
The return of action say `Action completed with no output`

### Job file (if appropriate)

```
job "nomad-action-from-ui-with-job-name-does-not-work" {
  name        = "Nomad action from ui with job name does not work"
  region      = "FRA"

  group "gp" {
    task "tk" {
      driver = "docker"
      config {
        image = "hashicorp/http-echo"
        args  = ["-text", "hello world"]
      }

      resources {
        cpu = 20
      }

      action "ls" {
        command = "ls"
        args    = ["-a"]
      }
    }
  }
}
```

philrenaud:
This makes it so that jobs with a `name` field that differs from their ID will correctly route an action when run via the web UI.

Resolves #23464

### DIFF
--- a/.changelog/23553.txt
+++ b/.changelog/23553.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Actions run from jobs with explicit name properties now work from the web UI
+```
--- a/ui/app/adapters/job.js
+++ b/ui/app/adapters/job.js
@@ -192,7 +192,7 @@ export default class JobAdapter extends WatchableNamespaceIDs {
 
     const wsUrl =
       `${protocol}//${prefix}/job/${encodeURIComponent(
-        job.get('name')
+        job.get('plainId')
       )}/action` +
       `?namespace=${job.get('namespace.id')}&action=${
         action.name


gulducat (approved):
:id:

PR #23560, opened 2024-07-11
api: handle newlines in JobSubmission vars correctly

pkazmierczak:
Fixes a bug where variable values in job submissions that contained newlines weren't encoded correctly, and thus jobs that contained them couldn't be resumed once stopped via the UI. 

Internal ref: https://hashicorp.atlassian.net/browse/NET-9966

### DIFF
--- a/.changelog/23560.txt
+++ b/.changelog/23560.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+api: Fixed bug where newlines in JobSubmission vars weren't encoded correctly
+```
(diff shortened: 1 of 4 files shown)


tgross (approved):
I'm having a little trouble imagining a scenario in which multiline strings would be interpolated correctly when we render the HCL2. The reported issue doesn't seem to have an example of that, only submitting them without consuming them in the jobspec.

But this seems like a reasonable change anyways.

PR #23565, opened 2024-07-12
e2e: purge the job in the UI stop_proxy() script

pkazmierczak:
otherwise namespace deletion fails due to non-terminal allocations


jrasell (approved):
LGTM!

Issue #23528, opened 2024-07-10
Automated Key Rotation Not Occurring Every 30 Days


ChefAustin:
### Nomad version
```
$ nomad version
Nomad v1.8.0+ent
BuildDate 2024-05-28T18:08:23Z
Revision 5a7a5d80944f1568c8bcd4e71a7f923fc04f2461
```

### Operating system and Environment details
```
$ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.4 LTS"
```

### Issue
[Possibly related to #19669 ? `¯\_(ツ)_/¯`]

According to the [‘Key Rotation’ section of Hashicorp Nomad's Key Management Operation Docs](https://developer.hashicorp.com/nomad/docs/operations/key-management#key-rotation), _“Nomad automatically rotates the active encryption key every 30 days”._

```
$ nomad operator root keyring list -verbose
Key                                   State   Create Time
c5b22e0f-0fdf-5ae6-6b81-8bec90df57c9  active  2023-08-03T08:23:36+02:00
```
Note how on my cluster, the key's `Create Time` is greater than 30 days (indicating that, perhaps, the recurring 30 day task for key rotation is not occurring as it should be).

Note: It is entirely possible that I am misunderstanding some nuances of how the keyring and key rotation work.

(Also: This issue dupes internal JIRA ticket NET-10398 && Hashicorp Support request #154558.)

### Reproduction steps
- Bring online a Nomad cluster using v1.6.1
- Upgrade to v1.7.3
- Upgrade to v1.7.3-ent
- Upgrade to v1.7.4-ent
- Create a second cluster (v1.7.4-ent) in a different region and federate it with the first cluster
- Upgrade to v1.8.0-ent
- Wait 31 days
- Check keyring list for active key's "Create Time"

#### Expected Result
Active key's `Create Time` should not be more than 30 days old.

#### Actual Result
Active key's `Create Time` is older than 30 days (342 days old in my case).

tgross:
[...]

Issue #19669, opened 2024-01-08
Root signing key rotation and prepublishing


schmichael:
Nomad 1.7 uses a root encryption key to encrypt Variables at rest and a root signing key to sign Workload Identities.

These root keys should be rotated automatically using the following logic:

1. New root key should be generated at [`root_key_rotation_threshold / 2`](https://developer.hashicorp.com/nomad/docs/configuration/server#root_key_rotation_threshold) and the public signing key published before use in the JWKS endpoint.
2. At `root_key_rotation_threshold` the prepublished key will be made `active` and the old `active` key will be made `inactive`
3. After `root_key_rotation_threshold` + [`root_key_gc_threshold`](https://developer.hashicorp.com/nomad/docs/configuration/server#root_key_gc_threshold) after the old key was marked inactive, it should be garbage collected.
   - Update `root_key_*` docs to reflect that keys are not gc'd until `rotation_threshold + gc_threshold` are reached to avoid invalidating otherwise valid JWTs in use. 
4. Jobspecs with an `identity.tll > root_key_rotation_threshold` should receive a Warning on submit.

## Prior Art

https://github.com/hashicorp/vault/pull/12414

tgross:
What you're envisioning here should cover https://github.com/hashicorp/nomad/issues/19367 and https://github.com/hashicorp/nomad/issues/19368. I'm going to unassign myself from those and if you want, we can either close them out now or you can close them out with this issue.

schmichael:
Thanks for linking things together Tim.

> we can either close them out now or you can close them out with this issue.

I'm going to leave them open until this ships to ensure everything is buttoned up appropriately.

tgross:
Implemented in #23577 and will ship in the next regular release of Nomad 1.8.x, with backports to Nomad 1.7.x/1.6.x Enterprise.

Issue #19368, opened 2023-12-07
keyring: keys stuck in `rekeying` after `rotate -full`


tgross:
In https://github.com/hashicorp/nomad/issues/19340 @sbihel reported a behavior where keys would get stuck in `rekeying` after running `nomad operator root keyring rotate -full`. I was able to confirm this and build a quick unit test to cover it, but I don't yet have a fix for the underlying problem.

https://github.com/hashicorp/nomad/issues/19340 covered another critical bug and was automatically closed once the fix was merged. This issue is a follow-up.

tgross:
Ref https://github.com/hashicorp/nomad/issues/19669

tgross:
Fixed in #23577 and will ship in the next regular release of Nomad 1.8.x, with backports to Nomad 1.7.x/1.6.x Enterprise.

Issue #16359, opened 2023-03-07
 GC limits > 3 days are in effect infinite b/c of FSM timetable limit


stswidwinski:
### Nomad version
1.5.0 and anything prior.

### Operating system and Environment details
Unix.

### Issue
When Garbage collection limits are set to a value larger than 3 days, the Nomad Scheduler will never garbage collect the required object leading to infinite accumulation of data (and infinite memory and disk leak) and related resources (such as CSI volumes). The GC limits included are at least:

1. https://developer.hashicorp.com/nomad/docs/configuration/server#eval_gc_threshold
2. https://developer.hashicorp.com/nomad/docs/configuration/server#batch_eval_gc_threshold
3. https://developer.hashicorp.com/nomad/docs/configuration/server#deployment_gc_threshold
4. https://developer.hashicorp.com/nomad/docs/configuration/server#job_gc_threshold
5. https://developer.hashicorp.com/nomad/docs/configuration/server#acl_token_gc_threshold
6. https://developer.hashicorp.com/nomad/docs/configuration/server#csi_plugin_gc_threshold
7. https://developer.hashicorp.com/nomad/docs/configuration/server#csi_volume_claim_gc_interval

The expected behavior is that it is possible to set garbage collection limits at a much larger maximal value than 3 days to allow for history build up and easier debugging.

### The details of the bug.

At the time of garbage collection, Nomad will derive an approximate raft index which is used as a watermark for garbage collection. The mapping of time to such an index is handled uniformly via:

https://github.com/hashicorp/nomad/blob/v1.5.0/nomad/core_sched.go#L1133-L1143

This relies on fsm  and the TimeTable  which is initialized within. To be precise, the initialization of this table occurs here:

https://github.com/hashicorp/nomad/blob/v1.5.0/nomad/fsm.go#L170

With a hard-coded maximal time table limit:

https://github.com/hashicorp/nomad/blob/v1.5.0/nomad/fsm.go#L27-L29

If the limit is breached, the resolution of the index is going to default to zero:

https://github.com/hashicorp/nomad/blob/v1.5.0/nomad/timetable.go#L93-L106

Hence, thresholdIndex = 0  which results in any check of the form X.modifyIndex > thresholdIndex  to evaluate to true  resulting in no garbage collection. For instance, for eval s:

https://github.com/hashicorp/nomad/blob/v1.5.0/nomad/core_sched.go#L282-L288

### Repro.

[...]

PR #23577, opened 2024-07-12
keyring: support prepublishing keys

tgross:
When a root key is rotated, the servers immediately start signing Workload Identities with the new active key. But workloads may be using those WI tokens to sign into external services, which may not have had time to fetch the new public key and which might try to fetch new keys as needed.

Add support for prepublishing keys. Prepublished keys will be visible in the JWKS endpoint but will not be used for signing or encryption until their `PublishTime`. Update the periodic key rotation to prepublish keys at half the `root_key_rotation_threshold` window, and promote prepublished keys to active after the `PublishTime`.

This changeset also fixes three bugs in periodic root key rotation and garbage collection, none of which can be safely fixed without implementing prepublishing:

* Periodic root key rotation would never happen because the default `root_key_rotation_threshold` of 720h exceeds the 72h maximum window of the FSM time table. We now compare the `CreateTime` against the wall clock time instead of the time table. (We expect to remove the time table in future work, ref https://github.com/hashicorp/nomad/issues/16359)
* Root key garbage collection could GC keys that were used to sign identities. We now wait until `root_key_rotation_threshold` + `root_key_gc_threshold` before GC'ing a key.
*  When rekeying a root key, the core job did not mark the key as inactive after the rekey was complete.

Ref: https://hashicorp.atlassian.net/browse/NET-10398
Ref: https://hashicorp.atlassian.net/browse/NET-10280
Fixes: https://github.com/hashicorp/nomad/issues/19669
Fixes: https://github.com/hashicorp/nomad/issues/23528
Fixes: https://github.com/hashicorp/nomad/issues/19368

---

Notes for reviewers:
* This is strictly speaking an enhancement but because it's the only way to safely fix the 3 bugs listed (i.e. without orphaning Workload Identities), I'm planning on backporting it to supported Nomad Enterprise versions. Alternately, we could remove periodic key rotation entirely from those older versions and that would "fix" the bug there.
* I'll need to manually backport the docs changes to 1.7.x and 1.6.x CE branches.

### DIFF
--- a/nomad/core_sched.go
+++ b/nomad/core_sched.go
@@ -99,7 +99,7 @@ func (c *CoreScheduler) forceGC(eval *structs.Evaluation) error {
 	if err := c.expiredACLTokenGC(eval, true); err != nil {
 		return err
 	}
-	if err := c.rootKeyGC(eval); err != nil {
+	if err := c.rootKeyGC(eval, time.Now()); err != nil {
 		return err
 	}
 	// Node GC must occur after the others to ensure the allocations are
@@ -902,20 +902,17 @@ func (c *CoreScheduler) rootKeyRotateOrGC(eval *structs.Evaluation) error {
 	// a rotation will be sent to the leader so our view of state
 	// is no longer valid. we ack this core job and will pick up
 	// the GC work on the next interval
-	wasRotated, err := c.rootKeyRotate(eval)
+	wasRotated, err := c.rootKeyRotate(eval, time.Now())
 	if err != nil {
 		return err
 	}
 	if wasRotated {
 		return nil
 	}
-	return c.rootKeyGC(eval)
+	return c.rootKeyGC(eval, time.Now())
 }
 
-func (c *CoreScheduler) rootKeyGC(eval *structs.Evaluation) error {
-
-	oldThreshold := c.getThreshold(eval, "root key",
-		"root_key_gc_threshold", c.srv.config.RootKeyGCThreshold)
+func (c *CoreScheduler) rootKeyGC(eval *structs.Evaluation, now time.Time) error {
 
 	ws := memdb.NewWatchSet()
 	iter, err := c.snap.RootKeyMetas(ws)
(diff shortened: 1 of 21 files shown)


nomad/core_sched.go
```diff
@@ -902,40 +902,49 @@ func (c *CoreScheduler) rootKeyRotateOrGC(eval *structs.Evaluation) error {
...
 	iter, err := c.snap.RootKeyMetas(ws)
 	if err != nil {
 		return err
 	}
 
+	// the threshold is longer than we can support with the time table, and we
+	// never want to force-GC keys because that will orphan signed Workload
+	// Identities
+	rotationThreshold := now.Add(-1 *
+		(c.srv.config.RootKeyRotationThreshold + c.srv.config.RootKeyGCThreshold))
```
pkazmierczak, line 927: I wonder if this couldn't or shouldn't be pulled out somewhere to the top? I'm thinking if we'd ever want to make this a configurable option, for example.

nomad/core_sched.go
```diff
@@ -902,40 +902,49 @@ func (c *CoreScheduler) rootKeyRotateOrGC(eval *structs.Evaluation) error {
...
+		if !keyMeta.Inactive() {
+			continue // never GC keys we're still using
 		}
-		if keyMeta.CreateIndex > oldThreshold {
-			continue // don't GC recent keys
+
+		c.logger.Trace("checking inactive key eligibility for gc",
+			"create_time", keyMeta.CreateTime, "threshold", rotationThreshold.UnixNano())
+
+		if keyMeta.CreateTime > rotationThreshold.UnixNano() {
```
pkazmierczak, line 942: When could this ever happen?

nomad/structs/keyring.go
```diff
@@ -84,15 +95,17 @@ type RootKeyMeta struct {
...
 type RootKeyState string
 
 const (
-	RootKeyStateInactive RootKeyState = "inactive"
-	RootKeyStateActive                = "active"
-	RootKeyStateRekeying              = "rekeying"
+	RootKeyStateInactive     RootKeyState = "inactive"
+	RootKeyStateActive                    = "active"
+	RootKeyStateRekeying                  = "rekeying"
+	RootKeyStatePrepublished              = "prepublished"
```
pkazmierczak, line 108: the api also has a `deprecated` state, do we not want to mirror it here?

pkazmierczak (approved):
LGTM! Great work! I left some questions/comments, but nothing major.

nomad/core_sched.go
pkazmierczak: oh I get it, we set this variable to a different value in methods below, and `rootKeyRotationThreshold` is configurable. Ignore me.

nomad/structs/keyring.go
tgross: We do below. It's hidden by GitHub's interface, but if you extend the diff to lines L100/R113 you'll see it. It's split out because it is, ironically, deprecated so there's a comment to that effect.

pkazmierczak: 🤦‍♂️ sorry, missed it.

nomad/core_sched.go
tgross: If I rotate a key without prepublishing it'll be active and be used for signing. If I rotate it again, it'll be inactive and therefore eligible for GC, but it will be recent enough that there are still signed workload identities that have not expired. We don't keep a record in the state store of signed WI's (unlike Variables), so we need to make sure we're not deleting a key that's still in use.

pkazmierczak: got it, thanks

schmichael (approved):
CLI comment isn't a blocker but give it a quick think :grin:

nomad/structs/keyring.go
```diff
@@ -185,6 +198,7 @@ func (rkm *RootKeyMeta) Active() bool {
 
 func (rkm *RootKeyMeta) SetActive() {
 	rkm.State = RootKeyStateActive
+	rkm.PublishTime = 0
```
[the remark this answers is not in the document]
tgross: Good call. Will do.

nomad/core_sched.go
```diff
@@ -961,26 +970,96 @@ func (c *CoreScheduler) rootKeyGC(eval *structs.Evaluation) error {
...
+			return false, nil
+		}
+
+		rootKey, err := c.srv.encrypter.GetKey(prepublishedKey.KeyID)
+		if err != nil {
+			c.logger.Error("prepublished key does not exist in keyring", "error", err)
+			return false, nil
+		}
+		rootKey = rootKey.Copy()
+		rootKey.Meta.SetActive()
```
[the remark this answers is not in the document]
tgross: Yeah I really like this. Removing `SetActive` adds a couple of extra copies but only in very infrequent code paths (like rotation) and it sweeps away a huge category of future bugs.

PR #23580, opened 2024-07-12
keyring: support external KMS for key encryption key (KEK)

tgross:
_WIP: needs testing updates and infra for end-to-end testing_

In Nomad 1.4.0, we shipped support for encrypted Variables and signed Workload Identities, but the key material is protected only by a AEAD encrypting the KEK. Add support for Vault transit encryption and external KMS from major cloud providers. The servers call out to the external service to decrypt each key in the on-disk keystore.

Ref: https://hashicorp.atlassian.net/browse/NET-10334
Fixes: https://github.com/hashicorp/nomad/issues/14852


nomad/encrypter.go
```diff
@@ -70,25 +79,67 @@ type keyset struct {
...
+	err = encrypter.loadKeystore()
 	if err != nil {
 		return nil, err
 	}
 	return encrypter, nil
 }
 
+// fallbackVaultConfig allows the transit provider to fallback to using the
+// default Vault cluster's configuration block, instead of repeating those
+// fields
```
schmichael, line 105: Oh this is a really nice detail.

schmichael (approved):
Looks great! Very excited to see this land.

dduzgun-security (approved):
This is really awesome! 👍 Should we also add some tests for the gcp and azure external kms just in case we ever face different behaviours than the aws kms one?

tgross:
> Should we also add some tests for the gcp and azure external kms just in case we ever face different behaviours than the aws kms one?

It's a good question. Unfortunately we only have an E2E environment for AWS (see https://github.com/hashicorp/nomad/pull/23601) for those tests. But fortunately the wrapper library's API is nicely designed such that the caller really doesn't have to care about the specifics, so then the wrapper library's own tests gives us the coverage we need to feel confident about shipping this.

PR #23594, opened 2024-07-15
testing: fix serf join on keyring test and tighten serf config

tgross:
Nomad's default serf configuration has a full sync interval of 60s (the WAN default configuration in the library). If tests need to join nodes and the leader is not in the join set, the test can hang up to twice that interval waiting for the new node to be seen by the leader and added to Raft.

This changeset includes the following tweaks to improve test timings:
* Ensure that nodes introduced later in the keyring replication test are joined to all peers. (Also updates the test to `shoenig/test`.)
* Update the `TestJoin` helper so that all servers passed are joined to the full set, instead of a set that's offset by 1, and use a single `Join` call for each server to reduce the number of messages sent.
* Reduce the `PushPullInterval` from 60s to 500ms in our unit test configuration, to force faster full syncs.

### DIFF
--- a/nomad/encrypter_test.go
+++ b/nomad/encrypter_test.go
@@ -18,6 +18,7 @@ import (
 	msgpackrpc "github.com/hashicorp/net-rpc-msgpackrpc/v2"
 	"github.com/shoenig/test"
 	"github.com/shoenig/test/must"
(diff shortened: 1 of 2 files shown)


gulducat (approved):
lgtm!

PR #23597, opened 2024-07-15
e2e: test cni args

martisah:
This PR adds an e2e test for cni args implemented in #23538. 

The test uses a job specification with a custom cni plugin "cni_args.sh" to write a file containing an interpolated variable passed in as a cni arg to the task local folder. Then the test makes sure that this file exists in the correct location and checks that the value was written and interpolated correctly.

### DIFF
--- a/e2e/terraform/packer/ubuntu-jammy-amd64/cni_args.conflist
+++ b/e2e/terraform/packer/ubuntu-jammy-amd64/cni_args.conflist
@@ -1,0 +1,9 @@
+{
+	"cniVersion": "1.0.0",
+	"name": "cni_args",
+	"plugins": [
+		{
+		    "type": "cni_args.sh"
+		}
+	]
+}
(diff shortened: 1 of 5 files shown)


e2e/terraform/packer/ubuntu-jammy-amd64/cni_args.conflist
```diff
@@ -0,0 +1,9 @@
+{
+	"cniVersion": "1.0.0",
+	"name": "cni_args",
+	"plugins": [
+		{
+		    "type": "cni_args.sh"
+		}
+	]
+}
```
tgross, line 9: This file is missing a EOL at the end of the file, which will mess up various unix tools (ex. try running `wc -l` on this file and you'll see it'll only have 8 lines!). You should be able to fix-on-save in your editor.

e2e/terraform/packer/ubuntu-jammy-amd64/cni_args.sh
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env bash
+
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: BUSL-1.1
+
+set -euo pipefail
+
+# things are prefixed with "Fancy*" because this is a fancy plugin.
```
tgross, line 8: :laughing:

tgross (approved):
LGTM!

PR #23599, opened 2024-07-16
docker: adjust capabilities on Windows

### LINKED ISSUE #15181: Nomad fails to run a container with Docker-CE installed on Windows
### Nomad version
1.4.2 (and several earlier versions)

- Windows Server 2019
- Docker CE installed based on the script provided by Microsoft: [Prep Windows for containers](https://learn.microsoft.com/en-us/virtualization/windowscontainers/quick-start/set-up-environment?tabs=dockerce#windows-server-1)

### Issue
Docker CE doesn't work with Nomad's Windows Task driver (error details below).

Currently there isn't an open-source Container Runtime for Windows supported by Nomad.
Docker Enterprise Edition has been sold to Mirantis and requires licensing for any use. 
Nomad's PodMan and Containerd drivers are not supported on Windows.

Docker CE is still free for use and is supported on Windows Server, so it seems like the logical choice for dev/test scenarios, but it doesn't work with Nomad's Windows Task driver (error details below).

### Reproduction steps

Using the Mirantis Container Runtime (Docker Enterprise) runtime based on [Prep Windows for containers](https://learn.microsoft.com/en-us/virtualization/windowscontainers/quick-start/set-up-environment?tabs=dockerce#windows-server-1) works as expected, just as the previous Docker Enterprise runtime did.

Using the Docker CE/Moby runtime based on [Prep Windows for containers](https://learn.microsoft.com/en-us/virtualization/windowscontainers/quick-start/set-up-environment?tabs=dockerce#windows-server-1) does not work as expected, and produces an error when Nomad attempts to run the image.

The image is pulled but the following error occurs:

Time	Type	Description
Nov 07, '22 16:04:39 -0600	Killing	Sent interrupt. Waiting 5s before force killing
Nov 07, '22 16:04:37 -0600	Alloc Unhealthy	Unhealthy because of failed task
Nov 07, '22 16:04:37 -0600	Not Restarting	Error was unrecoverable
**Nov 07, '22 16:04:37 -0600	Driver Failure	failed to create container: API error (400): invalid CapDrop: unknown capability: "CAP_NET_RAW"**
Nov 07, '22 16:03:46 -0600	Driver	Downloading image
Nov 07, '22 16:03:46 -0600	Task Setup	Building Task Directory
Nov 07, '22 16:03:46 -0600	Received	Task received by client

The image is 'seabopo/traefik:nanoserver-1809', which is verified working with the Docker Enterprise runtime installed.

There are also no logs available in the Nomad allocation logs when the Docker CE runtime is installed, but they do appear for the Docker Enterprise (Mirantis Container Runtime) installation.

pkazmierczak:
Adjusts Docker capabilities per OS, and checks for runtime on Windows. 

Fixes #15181 
Internal ref: https://hashicorp.atlassian.net/browse/NET-9506

### DIFF
--- a/drivers/docker/driver.go
+++ b/drivers/docker/driver.go
@@ -1115,14 +1115,11 @@ func (d *Driver) createContainerConfig(task *drivers.TaskConfig, driverConfig *T
 	if err != nil {
 		return c, err
 	}
-	info, err := client.Info()
-	if err != nil {
-		return c, err
-	}
+	ver, _ := client.Version()
 
 	// set add/drop capabilities
 	if hostConfig.CapAdd, hostConfig.CapDrop, err = capabilities.Delta(
-		capabilities.DockerDefaults(info), d.config.AllowCaps, driverConfig.CapAdd, driverConfig.CapDrop,
+		capabilities.DockerDefaults(ver), d.config.AllowCaps, driverConfig.CapAdd, driverConfig.CapDrop,
 	); err != nil {
 		return c, err
 	}
--- a/drivers/docker/driver.go
+++ b/drivers/docker/driver.go
@@ -1115,7 +1115,10 @@ func (d *Driver) createContainerConfig(task *drivers.TaskConfig, driverConfig *T
 	if err != nil {
 		return c, err
 	}
-	ver, _ := client.Version()
+	ver, err := client.Version()
+	if err != nil {
+		return c, err
+	}
 
 	// set add/drop capabilities
 	if hostConfig.CapAdd, hostConfig.CapDrop, err = capabilities.Delta(
--- a/drivers/docker/driver.go
+++ b/drivers/docker/driver.go
@@ -1109,9 +1109,20 @@ func (d *Driver) createContainerConfig(task *drivers.TaskConfig, driverConfig *T
 	}
 	hostConfig.Privileged = driverConfig.Privileged
 
+	// get docker client info (we need to know the runtime to adjust
+	// OS-specific capabilities)
+	client, err := d.getDockerClient()
+	if err != nil {
+		return c, err
+	}
+	info, err := client.Info()
+	if err != nil {
+		return c, err
+	}
+
 	// set add/drop capabilities
 	if hostConfig.CapAdd, hostConfig.CapDrop, err = capabilities.Delta(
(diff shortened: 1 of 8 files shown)


drivers/shared/capabilities/defaults_windows.go
```diff
@@ -0,0 +1,33 @@
...
+// DockerDefaults is a list of Windows capabilities enabled by Docker by default
+// and is used to compute the set of capabilities to add/drop given docker driver
+// configuration.
+//
+// Doing this on windows is somewhat tricky, because capabilities differ by
+// runtime, so we have to perform some extra checks.
+func DockerDefaults(ver *docker.Env) *Set {
+	defaults := NomadDefaults()
+
+	// Docker CE doesn't support NET_RAW on Windows, Mirantis (aka Docker EE) does
```
tgross, line 23: I think you can set it manually, right? It's just not in the default list?

```suggestion
	// Docker CE doesn't include NET_RAW on Windows, Mirantis (aka Docker EE) does
```

drivers/docker/driver.go
```diff
@@ -1109,9 +1109,17 @@ func (d *Driver) createContainerConfig(task *drivers.TaskConfig, driverConfig *T
 	}
 	hostConfig.Privileged = driverConfig.Privileged
 
+	// get docker client info (we need to know the runtime to adjust
+	// OS-specific capabilities)
+	client, err := d.getDockerClient()
+	if err != nil {
+		return c, err
+	}
+	ver, _ := client.Version()
```
tgross, line 1118: We should probably return this error rather than swallowing it, as the reason for the resulting behavior of a possibly-broken list of capabilities won't be observable to the user.

tgross (approved):
LGTM!

It's really unfortunate that Docker doesn't expose this default via the API (ex. their "system" API), as I'm sure at some point in the future they're going to change it again and break this behavior. But I can't think of a better way to handle it in the face of that, either, short of doing something silly-expensive like just running the container and retrying with a different set of caps if that fails.

drivers/docker/driver.go
pkazmierczak: Yeah my thinking was: if we can't get the version object for some reason, the Windows code will just play it safe and assume we're not using Mirantis, but you're right: makes it harder to debug.

PR #23601, opened 2024-07-16
keyring: E2E testing for KMS/rotation

tgross:
In #23580 we're implementing support for encrypting Nomad's key material with external KMS providers or Vault Transit. In #23577 we're implementing support for prepublishing keys. This changeset updates the E2E infrastructure to use an external KMS and adds tests for rotation.

Ref: https://hashicorp.atlassian.net/browse/NET-10398
Ref: https://hashicorp.atlassian.net/browse/NET-10280
Ref: https://hashicorp.atlassian.net/browse/NET-10334
Ref: https://github.com/hashicorp/nomad/issues/14852
Ref: https://github.com/hashicorp/nomad/pull/23580

---

Note to reviewers: this PR requires #23580 and #23577 to be merged before we can merge this. Test run against a build that includes both PRs:

```
$ go test -v -count=1 ./keyring
=== RUN   TestKeyringRotation
--- PASS: TestKeyringRotation (0.62s)
PASS
ok      github.com/hashicorp/nomad/e2e/keyring  0.632s
```

### DIFF
--- a/e2e/keyring/keyring_test.go
+++ b/e2e/keyring/keyring_test.go
@@ -36,12 +36,12 @@ func TestKeyringRotation(t *testing.T) {
 	newCurrentKeys, newActiveKeyID := getKeyMeta(t, nc)
 	must.NotEq(t, "", newActiveKeyID, must.Sprint("expected an active key"))
 	must.Eq(t, activeKeyID, newActiveKeyID, must.Sprint("active key should not have rotated yet"))
-	must.Greater(t, len(currentKeys), len(newCurrentKeys))
+	must.Greater(t, len(currentKeys), len(newCurrentKeys), must.Sprint("expected more keys after prepublishing"))
 
 	keyset = getJWKS(t)
-	must.Len(t, len(newCurrentKeys), keyset.Keys)
+	must.Len(t, len(newCurrentKeys), keyset.Keys, must.Sprint("number of keys in jwks keyset should match keyring"))
 	for _, key := range keyset.Keys {
-		must.MapContainsKey(t, newCurrentKeys, key.KeyID)
+		must.MapContainsKey(t, newCurrentKeys, key.KeyID, must.Sprint("jwks keyset contains unexpected key"))
 	}
 	must.SliceContainsFunc(t, keyset.Keys, newKeyID, func(a jose.JSONWebKey, b string) bool {
 		return a.KeyID == b
@@ -49,6 +49,7 @@ func TestKeyringRotation(t *testing.T) {
 }
 
 func getKeyMeta(t *testing.T, nc *api.Client) (map[string]*api.RootKeyMeta, string) {
+	t.Helper()
 	keyMetas, _, err := nc.Keyring().List(nil)
 	must.NoError(t, err)
 
--- a/e2e/keyring/keyring_test.go
+++ b/e2e/keyring/keyring_test.go
@@ -30,6 +30,8 @@ func TestKeyringRotation(t *testing.T) {
 	out, err := e2eutil.Commandf("nomad operator root keyring rotate -verbose -prepublish 1h")
 	must.NoError(t, err)
 	cols, err := e2eutil.ParseColumns(out)
+	must.NoError(t, err)
+	must.Greater(t, 0, len(cols))
 	newKeyID := cols[0]["Key"]
 	must.Eq(t, "prepublished", cols[0]["State"], must.Sprint("expected new key to be prepublished"))
 
--- a/e2e/keyring/keyring_test.go
+++ b/e2e/keyring/keyring_test.go
@@ -1,0 +1,77 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package keyring
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/go-jose/go-jose/v3"
+	"github.com/hashicorp/nomad/api"
+	"github.com/hashicorp/nomad/e2e/e2eutil"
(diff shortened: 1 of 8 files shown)


e2e/keyring/keyring_test.go
```diff
@@ -0,0 +1,77 @@
...
+	newKeyID := cols[0]["Key"]
+	must.Eq(t, "prepublished", cols[0]["State"], must.Sprint("expected new key to be prepublished"))
+
+	newCurrentKeys, newActiveKeyID := getKeyMeta(t, nc)
+	must.NotEq(t, "", newActiveKeyID, must.Sprint("expected an active key"))
+	must.Eq(t, activeKeyID, newActiveKeyID, must.Sprint("active key should not have rotated yet"))
+	must.Greater(t, len(currentKeys), len(newCurrentKeys))
+
+	keyset = getJWKS(t)
+	must.Len(t, len(newCurrentKeys), keyset.Keys)
```
gulducat, line 42:
```suggestion
	must.Len(t, len(newCurrentKeys), keyset.Keys, must.Sprint("number of keys in jwks keyset should match keyring"))
```

e2e/keyring/keyring_test.go
```diff
@@ -0,0 +1,77 @@
...
+	for _, key := range keyset.Keys {
+		must.MapContainsKey(t, newCurrentKeys, key.KeyID)
```
gulducat, line 44:
```suggestion
		must.MapContainsKey(t, newCurrentKeys, key.KeyID, must.Sprint("jwks keyset contains unexpected key"))
```

e2e/keyring/keyring_test.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	must.SliceContainsFunc(t, keyset.Keys, newKeyID, func(a jose.JSONWebKey, b string) bool {
+		return a.KeyID == b
+	}, must.Sprint("expected prepublished key to appear in JWKS endpoint"))
+}
+
+func getKeyMeta(t *testing.T, nc *api.Client) (map[string]*api.RootKeyMeta, string) {
+	keyMetas, _, err := nc.Keyring().List(nil)
```
gulducat, line 52:
```suggestion
	t.Helper()
	keyMetas, _, err := nc.Keyring().List(nil)
```

e2e/keyring/keyring_test.go
```diff
@@ -0,0 +1,77 @@
...
+			activeKeyID = keyMeta.KeyID
+		}
+	}
+	must.NotEq(t, "", activeKeyID, must.Sprint("expected an active key"))
+	return currentKeys, activeKeyID
+}
+
+func getJWKS(t *testing.T) *jose.JSONWebKeySet {
+	t.Helper()
+	out, err := e2eutil.Commandf("nomad operator api /.well-known/jwks.json")
```
gulducat, line 69: tangent: intriguing that our `api` package doesn't have a method for this.

e2e/terraform/variables.tf
```diff
@@ -87,6 +87,12 @@ variable "hcp_vault_namespace" {
   default     = "admin"
 }
 
+variable "aws_kms_alias" {
+  description = "The alias for the AWS KMS key ID"
+  type        = string
+  default     = "kms-nomad-keyring"
```
gulducat, line 93: just checking: if I turn on my own cluster for my own purposes, I can leave this as-is and not collide with any other in-flight e2e clusters, right?  like I don't need to go out of my way to create a new kms key (which is done in our other nomad-e2e repo) and point this at it?

e2e/keyring/keyring_test.go
```diff
@@ -0,0 +1,77 @@
...
+	must.NotEq(t, "", activeKeyID, must.Sprint("expected an active key"))
+
+	keyset := getJWKS(t)
+
+	must.Len(t, len(currentKeys), keyset.Keys)
+	for _, key := range keyset.Keys {
+		must.MapContainsKey(t, currentKeys, key.KeyID)
+	}
+
+	out, err := e2eutil.Commandf("nomad operator root keyring rotate -verbose -prepublish 1h")
```
gulducat, line 30: re: `1h` -- this test doesn't cover the full lifecycle past prepublish.  do we want to?

gulducat (approved):
lgtm!  I mostly have little test message suggestions for failure legibility, and a couple questions.

e2e/keyring/keyring_test.go
tgross: The GC window is very long, so even if you say `-prepublish 1m` that just means the next GC will mark it active. That's too long to wait, and we have exhaustive coverage in the main PR.

e2e/keyring/keyring_test.go
tgross: Yeah, I considered adding that here but I wanted to see if there was a reason that @schmichael didn't implement it first.

e2e/terraform/variables.tf
tgross: So long as you don't also change the IAM instance profile, yes that'll work just fine. The KMS is only used to encrypt/decrypt and doesn't store the actual key material at all.

e2e/keyring/keyring_test.go
```diff
@@ -0,0 +1,77 @@
...
+	must.NoError(t, err)
+	cols, err := e2eutil.ParseColumns(out)
+	newKeyID := cols[0]["Key"]
```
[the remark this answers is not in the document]
tgross: True, but I've added a check anyways (including a `must.NoError` on the error returned from `ParseColumns`).

PR #23609, opened 2024-07-17
cni: test Setup method

martisah:
This PR adds a test for the `Setup` method, as well as an `nsOpts` field to the `cniNetworkConfigurator` struct. This field is specifically used to keep track of namespace usage for test assertions, and is used in the test written for Setup.

### DIFF
--- a/client/allocrunner/networking_cni.go
+++ b/client/allocrunner/networking_cni.go
@@ -49,21 +49,6 @@ const (
 	defaultCNIInterfacePrefix = "eth"
 )
 
-type nsOpts struct {
-	args  map[string]string
-	ports []cni.PortMapping
-}
-
-func (o *nsOpts) WithArgs(args map[string]string) cni.NamespaceOpts {
-	o.args = args
-	return cni.WithLabels(args)
-}
-
-func (o *nsOpts) WithCapabilityPortMap(ports []cni.PortMapping) cni.NamespaceOpts {
-	o.ports = ports
-	return cni.WithCapabilityPortMap(ports)
-}
-
 type cniNetworkConfigurator struct {
 	cni                     cni.CNI
 	cniConf                 []byte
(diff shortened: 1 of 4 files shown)


tgross (commented):
This looks great @martisah!

Not a blocker but something to consider is whether we really need to keep the `nsOpts` attached to the `cniNetworkConfigurator`. By doing so, we're preventing the port map slice and CNI args map from ever being garbage collected by the Go runtime for the lifetime of the allocation, even though they're never used again. These aren't _huge_ objects, but it can't hurt to be thoughtful about memory usage.

I'm thinking of this especially because we're only keeping the pointer to `nsOpts` to support tests, but we're using a mock plugin interface for the tests anyways. We could have the mock record the port mapping and args it gets, and read it out of the mock directly in the tests instead of recording it on the `cniNetworkConfigurator`.

gulducat:
We tried various things to work around the private-ness of the `go-cni` lib, and came up short. This as the cleanest option we could find, but if you have other ideas, we'd love to hear them!

I'll walk through how the real implementation of `CNI.Setup()` works, as it points to why this is hard to test purely with the mock.

* The real `Setup()` uses `Namespace`'s constructor [here](https://github.com/containerd/go-cni/blob/v1.1.9/cni.go#L159).
* [`newNamespace`](https://github.com/containerd/go-cni/blob/v1.1.9/namespace.go#L55) sets attributes to not-nil, then executes the `NamespaceOpts` functions on the `Namespace`
* Those opts mutate fields, like [`WithLabels()` -> `args`](https://github.com/containerd/go-cni/blob/v1.1.9/namespace_opts.go#L82), on the `Namespace`

Note the concrete-ness of `*Namespace` in the `CNI` interface, and the private-ness of `newNamespace()` and `Namespace.args`.

In our _mock_ `Setup()`, we can create our own `Namespace`, but without the private constructor to set non-nil private fields, `WithLabels` panics trying to set a key on a nil map.  Then even if we could set that, we'd later also need to read it to assert its value.

Options we thought up:
 * Further abstract the `CNI` interface so `Namespace` didn't have to be concrete; it could itself be an interface, which we could mock, or our own concrete type with public fields
   * this only kicks the can down the road, because (presumably) we'd use CNI eventually, anyway
 * There is a [`config()` method](https://github.com/containerd/go-cni/blob/v1.1.9/namespace.go#L70) that returns a `RuntimeConf` which does have public `Args`
   * but `config()` itself is private.
 * Try to look at the `NamespaceOpts` themselves, rather than executing them on a `Namespace`
   * I could not find a good way to assert these funcs (they are just funcs) were produced with the args we care about
 * `reflection`, `reflection` everywhere, to set and read the private fields
   * I wrote ~half of this implementation and felt real bad about it
 * We could also have made an upstream PR to public-ize what we want to set/check.

We settled on what you see in this PR, but maybe we didn't properly understand another option.  We'd love to hear a better way!

tgross:
Ah, I see what you mean about the `Namespace`. It's not obvious that we need to handle the concrete object until you start digging around in those functions. No worries. :shipit:

tgross (approved):

PR #23610, opened 2024-07-17
e2e: skip missing windows ami if windows clients=0

gulducat:
and tweak Makefile to generate a `custom.tfvars` instead of specifying vars separately via CLI.
hoping this makes it a little more obvious if there is no consul/nomad license.

I often packer build only linux if I'm testing something in linux (which is most of the time),
and as a result, there's no windows image with my git sha, so tf plan fails looking for the AMI,
even though I set windows clients to 0.

the change to `compute.tf` allows me to avoid futzing with any .tf files when testing such things.

### DIFF
--- a/e2e/terraform/.gitignore
+++ b/e2e/terraform/.gitignore
@@ -1,2 +1,3 @@
 *.zip
 uploads/
+custom.tfvars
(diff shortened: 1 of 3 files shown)


tgross (approved):
LGTM! Thanks for picking this up... it's definitely something I've run into before and hacked around by commenting out the Windows resources, which is not ideal :grinning:

PR #23612, opened 2024-07-17
e2e: hc-install consul-cni

gulducat:
now that the version with tproxy CNI_ARGS is on releases.hashicorp.com, instead of building from source.


tgross (approved):
Thanks @gulducat!

PR #23618, opened 2024-07-18
fix: Remove the changelog from v1.6.13

Juanadelacuesta:
After updating the 1.6.13 release, I noticed none of the changes of 1.8.2 applied to it, this PR updates the changelog to reflect it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -479,54 +479,6 @@ BUG FIXES:
 
 ## 1.6.13 Enterprise (July 16, 2024)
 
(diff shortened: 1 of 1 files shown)


tgross (changes requested):
This removes _all_ the changelog entries for 1.6.13 Enterprise, but there are definitely items in the bug fix list at least that should remain. For example: https://github.com/hashicorp/nomad/pull/23432

Juanadelacuesta:
Then there is a bigger problem, because the temp changelog for 1.6.13 came empty

PR #23623, opened 2024-07-18
go-memdb performance improvement experiments

absolutelightning:


nomad/state/state_store.go
```diff
@@ -221,8 +221,12 @@ func (s *StateStore) Config() *StateStoreConfig {
 // Snapshot is used to create a point in time snapshot. Because
 // we use MemDB, we just need to snapshot the state of the underlying
 // database.
-func (s *StateStore) Snapshot() (*StateSnapshot, error) {
-	memDBSnap := s.db.memdb.Snapshot()
+func (s *StateStore) Snapshot(deep bool) (*StateSnapshot, error) {
+	memDBSnap := s.db.memdb
```
tgross, line 225: Wouldn't this result in phantom reads and losing non-repeatable reads?

absolutelightning: Tried to fix this using a flag

Issue #23615, opened 2024-07-17
404 Not Found when getting volume details


deuspt:
### Nomad version
Nomad v1.8.2
BuildDate 2024-07-16T08:50:09Z
Revision 7f0822c1e4f25907d9f60e2d595411950dd1bd28

Nomad v1.8.1
BuildDate 2024-06-19T06:43:57Z
Revision 5022543e4b7b8dcec9df123f86630ae3fdcffbe6

### Operating system and Environment details
Linux 6.5.0-1022-aws #22~22.04.1-Ubuntu SMP Fri Jun 14 16:31:00 UTC 2024 x86_64

### Issue
While attempting to get a csi volume details, it returns the "404 Not found" error page.
This happens when clicking a volume from the UI on "storage > volumes > name" or via the respective allocation page (with both links in the form "https://fqdn/ui/csi/volumes/name@namespace")
The following error is seen in the browser console too:

Request URL:
https://fqdn/v1/volume/csi%2Fname?namespace=namespace
Request Method:
GET
Status Code:
404 Not Found

I notice the request URL is slightly different but maybe that is normal?

### Reproduction steps
As above.

Also in the Nomad server, this works:
curl -v https://localhost:4646/ui/volume/csi%2Fname?namespace=namespace
curl -v https://localhost:4646/ui/v1/volume/csi%2Fname?namespace=namespace

this returns 404:
curl -v https://localhost:4646/v1/volume/csi%2Fname?namespace=namespace

#### Expected Result
See the volume details (was last working on 1.7.7 before upgrading).

#### Actual Result
404 Not Found error page.

### Job file (if appropriate)
N/A

### Nomad Server logs (if appropriate)
No relevant logs in server.

### Nomad Client logs (if appropriate)
N/A

philrenaud:
Hi @deuspt , thanks for raising this. I'll take a look. Is this behaviour observed on all volume pages or only on a subset of them?

deuspt:
All volumes yes.

deuspt:
Might be unrelated but worth mentioning. Via the new "strorage > plugins" page, in the Overview tab, shows the "health" gadgets with the correct number of controller and nodes at 100%, but both the allocation panels are empty. Also the Allocations tab shows no allocations.

PR #23625, opened 2024-07-18
[ui] Unescape csi/ in plugin GET requests

philrenaud:
This PR unescapes the `/` in the `csi/PLUGIN_NAME` path that Ember Data was escaping due to how/where it was prepended at the adapter layer.

Re: #23615 : This makes it so the plugins UI is usable again, but doesn't solve the underlying no-longer-handles-unescaped-path-slashes problem.


tgross (approved):
LGTM

PR #23629, opened 2024-07-18
Fix parsing CNI .conf and .json config formats

### LINKED ISSUE #19816: CNI .conf and .json configuration files not loaded properly - "no 'plugins' key"
Our documentation [here](https://developer.hashicorp.com/nomad/docs/networking/cni#cni-configuration-files) and [here](https://developer.hashicorp.com/nomad/docs/concepts/plugins/cni#cni-configuration-files) states

> `.conf` and `.json` files are loaded as individual [plugin configurations](https://github.com/containernetworking/cni/blob/main/SPEC.md#plugin-configuration-objects) for a specific network.

That is partly true, in the sense that fingerprinting picks it up correctly, but when you try to actually use a `.conf`-style plugin like this:

```json
{
  "cniVersion": "1.0.0",
  "name": "custom-conf-loopback",
  "type": "loopback"
}
```

with a job like this:

```hcl
job "custom-cni-conf" {
  group "g" {
    network {
      mode = "cni/custom-conf-loopback" # maps to cni conf "name"
    }
    task "t" {
      driver = "docker"
      config {
        image   = "busybox:1"
        command = "sleep"
        args    = ["3600"]
      }
    }
  }
}
```

the alloc will fail with a task event like

> failed to setup alloc: pre-run hook "network" failed: failed to configure networking for alloc: cni config load failed: error parsing configuration list: no 'plugins' key: failed to load cni config

This is because the allocrunner code always treats it as a *conflist* (which expects "plugins") when it goes to load the config [here](https://github.com/hashicorp/nomad/blob/v1.7.3/client/allocrunner/networking_cni.go#L339).  Nothing has tracked the file type through the whole signal chain to say "no this one is a *conf*."

gulducat:
Our documentation states that we support `.conf` and `.json` config formats (which are the same, actually), but in reality only `.conflist` works properly.  All of them are fingerprinted correctly, but only conflist works at allocation run time.

Fixes #19816

The first commit is here for a reference, which used a boolean to keep track of whether or not `confIsList`, but in my second pass I made a little struct to make it easier (I think) to hold correctly.

I have some e2e tests set up for this too, but I'll do that in a separate PR since we generally don't backport those.

### DIFF
--- a/client/allocrunner/networking_cni_test.go
+++ b/client/allocrunner/networking_cni_test.go
@@ -10,6 +10,8 @@ import (
 	"errors"
 	"math/rand"
 	"net"
+	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/containerd/go-cni"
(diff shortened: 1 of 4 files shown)


martisah (approved):

client/allocrunner/networking_cni_test.go
```diff
@@ -25,6 +27,98 @@ import (
...
+func TestLoadCNIConf_confParser(t *testing.T) {
+	confDir := t.TempDir()
+
+	writeFile := func(t *testing.T, filename, content string) {
+		t.Helper()
+		path := filepath.Join(confDir, filename)
+		must.NoError(t, os.WriteFile(path, []byte(content), 0644))
+		t.Cleanup(func() {
+			test.NoError(t, os.Remove(path))
+		})
```
tgross, line 39: This is in the `t.TempDir` so there's no need to clean it up manually.

tgross (approved):
LGTM! Don't forget a changelog!

client/allocrunner/networking_cni_test.go
gulducat: my reasoning was to remove each config between each sub-test, so the confdir only contains the config that I'm making assertions about.

arguably I could also test more behavior like multiple config files with duplicate names, but I worried I might get strung out if I go much further than I already have.

tgross: Makes sense :+1:

Issue #23619, opened 2024-07-18
support running within a snap


stuart-warren:
### Proposal

We'd like to see support for running nomad within a snap (https://snapcraft.io/about)

There are only minimal tweaks required for this to function and maintaining our own patches is obviously undesirable
* unable to chown directories to `nobody` https://github.com/hashicorp/nomad/blob/v1.8.2/client/allocdir/fs_unix.go#L36-L66
* tmpfs volumes created by a snap are unavailable to the docker snap due to mount namespaces https://github.com/hashicorp/nomad/blob/v1.8.2/client/allocdir/fs_linux.go#L65-L109

### Use-cases

Permits running on operating systems that support snaps, this includes Ubuntu Core

### Attempted Solutions

Have tried running as a non-root user, however docker support on Ubuntu Core requires root or sudo access

We are currently maintaining our own patches

Happy to contribute a PR

tgross:
Hi @stuart-warren! Running Nomad in a container is not supported. See [running Nomad in Docker](https://developer.hashicorp.com/nomad/docs/install/production/requirements#running-nomad-in-docker). https://github.com/hashicorp/nomad/issues/13669 describes some longer-term thinking around making this feasible (that issue is for rootless Nomad but most of the same problems apply when running in a container).

stuart-warren:
Hi @tgross thanks for the reply

I'm not quite sure I follow what the problem is with this?

Are you happy for me to submit some PRs for changes that work for us and discuss any issues with them there?

tgross:
@stuart-warren we _might_ be open to patches if they don't otherwise increase our burden of maintenance, but that's not going to change whether we officially support running Nomad in containers (ex. OCI images, snaps, flatpak, etc.). 

What tends to happen with these proposals is that folks submit patches that work in their environment with a specific task driver they use but will break a bunch of other task drivers. So then at the very least those patches need to have new config flags, and those configurations come with expectations around supporting those deployments. While teams like yours are obviously experienced enough and have a deep-enough understanding of cgroups, namespaces, seccomp, apparmor, etc. to make this work... many users do not and this increases the maintenance burden on our small team significantly.

PR #23630, opened 2024-07-18
client/allocdir: chown directories as _daemon_ user if available

stuart-warren:
running within a snap (https://snapcraft.io/) means code is unable to chown to nobody user unfortunately

chown to _daemon_ user instead if available https://snapcraft.io/docs/system-usernames

ref #23619

```sh
me@mydevice:~$ sudo snap run --shell my-snap.nomad
root@mydevice:/home/me# cd
root@mydevice:~# pwd
/root/snap/my-snap/x5
root@mydevice:~# touch foo
root@mydevice:~# chown _daemon_ foo
root@mydevice:~# touch bar
root@mydevice:~# chown nobody bar
chown: changing ownership of 'bar': Operation not permitted
```

### DIFF
--- a/client/allocdir/fs_unix.go
+++ b/client/allocdir/fs_unix.go
@@ -48,6 +48,14 @@ func dropDirPermissions(path string, desired os.FileMode) error {
 		return fmt.Errorf("Unable to find nobody user: %w", err)
 	}
 
+	// Within a snap we can't chown to `nobody`, so we use `_daemon_` instead
+	// requires snap configured to use `system-usernames`
+	// https://snapcraft.io/docs/system-usernames
+	snapu, err := users.Lookup("_daemon_")
+	if err == nil {
+		u = snapu
+	}
+
 	uid, err := getUid(u)
 	if err != nil {
 		return err


tgross (changes requested):
Hi @stuart-warren!

A few issues here:
1. From the documentation, the `_daemon_` user doesn't appear to be equivalent to the `nobody` user in terms of privileges within the container. In this particular instance it probably doesn't matter because the directory is being set 0777 anyways, but...
2. There are many many more places where we're looking for or setting this user in the code base, so if we're only doing it here we're leaving broken implementations elsewhere. And maybe not all of them are 0777 situations, so point (1) above is more relevant.
3. If there's no "nobody" user, as you're saying, then `user.Lookup("nobody")` returns an error and this new code is unreachable.
4. We're unconditionally and silently overwriting `u` if the `"_daemon_"` regardless of cluster administrator intent, which seems like a nasty surprise.

I think if we were going to do this kind of thing, we'd need to instead make the "nobody" user a client configuration and then plumb that through to everywhere in the client/driver code where that's exposed.

stuart-warren:
Hi @tgross 

> 2. There are many many more places where we're looking for or setting this user in the code base, so if we're only doing it here we're leaving broken implementations elsewhere. And maybe not all of them are 0777 situations, so point (1) above is more relevant.

I'm not seeing that many places that specify "nobody"

Only
* https://github.com/hashicorp/nomad/blob/7a2c70e3f6a8d9d4549661e0bfe2af65070886dc/client/allocdir/fs_unix.go#L46
* https://github.com/hashicorp/nomad/blob/7a2c70e3f6a8d9d4549661e0bfe2af65070886dc/client/allocdir/task_dir.go#L173

Every other mention appears to be in a test file. I obviously want the default to stay the same and find some method to allow the server admin to override. Seem to be able to set the user in config for drivers https://developer.hashicorp.com/nomad/docs/job-specification/task#user

> 3. If there's no "nobody" user, as you're saying, then `user.Lookup("nobody")` returns an error and this new code is unreachable.

There is a nobody user, snapd just won't allow you to chmod file resources to other users it seems

> 4. We're unconditionally and silently overwriting `u` if the `"_daemon_"` regardless of cluster administrator intent, which seems like a nasty surprise.
> 
> I think if we were going to do this kind of thing, we'd need to instead make the "nobody" user a client configuration and then plumb that through to everywhere in the client/driver code where that's exposed.

I'm happy to find a method to set a set some client config with the default of "nobody".

`dropDirPermissions()` is a function. Do you have a preferred method for me to get a value in there?
* package level variable?
* change function signature?
* change function to a method?

Many thanks

tgross:
> Every other mention appears to be in a test file. 

Ah, looks like you're right those are mostly tests. My error there.

> I obviously want the default to stay the same and find some method to allow the server admin to override. Seem to be able to set the user in config for drivers https://developer.hashicorp.com/nomad/docs/job-specification/task#user

The `task.user` isn't the same thing unfortunately. We're setting these as owned by `nobody` even if we're using, for example, the Docker driver where the task's user may be defined by the container image.

> There is a nobody user, snapd just won't allow you to chmod file resources to other users it seems

What's the reasoning behind that? Any chance you've got a pointer to a doc on this? Nomad assumes it can do this kind of thing and I suspect that if we can't assert ownership then we're just going to hit new bugs with ex. cpusets. We don't want to play whack-a-mole with bugs around this, especially because we generally have to drag it out of people that they're running in unsupported environments so that's just more toil on the maintainers.

> I'm happy to find a method to set a set some client config with the default of "nobody".
> `dropDirPermissions()` is a function. Do you have a preferred method for me to get a value in there?

Well it's got to be runtime configurable so that the same binary can be used in different deployments (otherwise why upstream this at all?). And it can't be configured per-driver because the allocdir is shared between tasks which can use different drivers.

If you walk up the call stack a ways you'll see that all these functions ultimately get their configuration via the [`NewAllocDir`](https://github.com/hashicorp/nomad/blob/v1.8.2/client/allocdir/alloc_dir.go#L160) constructor. At the call site for `NewAllocDir` in the allocrunner (ref [`alloc_runner.go#L273`](https://github.com/hashicorp/nomad/blob/v1.8.2/client/allocrunner/alloc_runner.go#L273)), we've got access to the whole client config struct. So if you get it into the client config (see all the config parsing in `command/agent` for starting point for that), then you can thread it thru the alloc dir constructor and down to the function.

stuart-warren:
I may have worked around the issue with chown to nobody. Will need some more testing and review upstream in snapd

Issue #23616, opened 2024-07-18
Enable / document process for accessing insecure (non-https) Docker registries from Nomad


jedd:
### Proposal
Enable - or better document the process if the capability currently actually exists - non-TLS/SSL (https) access to local / private docker registries.

### Use-cases
Within local & dev environments, it's convenient to access docker registries without setting up SSL infrastructure around those endpoints.

This used to be possible I believe in 0.5.x epoch, prior to the removal of the 'ssl: false' capability in the docker driver stanza

### Attempted Solutions
Varying approaches, including specifying both no URI type, and attempting to force http: type, on the `image = ` parameter.

This, combined with docker configuration including my non-TLS endpoints, eg:

```
  "insecure-registries" : ["my.internal.host:5000"]
```

combined with Nomad configuration for docker driver replicating / confirming this:

```
client {
  enabled = true

  options {
    "docker.volumes.enabled" = "true"
    "docker.insecure_registries" = "my.internal.host:5000"
    "docker.tls.disabled"  = "true"

   ...
```

I note current documentation:  
https://developer.hashicorp.com/nomad/docs/drivers/docker#SSL
 .. says this flag is deprecated since 0.5.3 - but has no indicators as to workarounds for that deprecation.

Tracking back to the bug / pr described:  in https://github.com/hashicorp/nomad/pull/1336
does make it look like functionality was removed, rather than (as described) 'default was changed'.

jedd:
Hey @tgross - thank you for a speedy response. 

My humble apologies, as the problem seems to have been self-inflicted, with a misconfiguration of the my docker configuration (json).  For context and maybe helpful with documentation updates I'll describe.

I am using 1.8.1, and have this as my image variable in my Nomad job:  `"dg-pan-01.int.jeddi.org:5001/pdc-agent:0.0.30"` - this is the 'registry-ui' service, which fronts my cncf 'distribution' registry (running on port 5000).

I'd be chopping and changing between the two endpoints as part of this build out.

I was getting the error:

```
[...]

PR #23642, opened 2024-07-19
docs: explain how to use insecure registries with Docker

tgross:
The documentation for the `SSL` option for the Docker driver is misleading inasmuch as it's both deprecated and non-functional in current versions of Docker. Remove this option from the docs and add a section explaining how to use insecure registries.

Fixes: https://github.com/hashicorp/nomad/issues/23616
Note to reviewers: see my repro here https://github.com/hashicorp/nomad/issues/23616#issuecomment-2239280566

### DIFF
--- a/website/content/docs/drivers/docker.mdx
+++ b/website/content/docs/drivers/docker.mdx
@@ -324,9 +324,6 @@ The `docker` driver supports the following configuration in the job spec. Only
   }
   ```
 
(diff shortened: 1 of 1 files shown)


pkazmierczak (approved):

philrenaud (approved):
Looks good, and 0.5.3 deprecation seems more than an apt removal! Thanks!

Issue #15851, opened 2023-01-23
change_mode = "script" sometime fails with Docker


dani:
### Nomad version
Nomad v1.4.3 (f464aca721d222ae9c1f3df643b3c3aaa20e2da7)

### Operating system and Environment details
AlmaLinux 8.7
Nomad installed manually from the pre-built binary. Using Docker CE 20.10.22

### Issue
For some of my tasks, I trigger a script when a template changes. Typicaly, I use this to trigger a custom reload action when a vault generated certificate gets renewed. Most of the time, this is working great. But sometimes, the execution fails with this error

```
Template failed to run script /local/bin/rotate-cert.sh with arguments [] because task driver doesn't support the exec operation
```

### Reproduction steps
A job with a templated cert, a reload script and change_mode = "script", for example with :
```
      template {
        data        = <<-EOF
          {{ with pkiCert "pki/db/issue/mongo" "common_name=mongo.service.consul" "ttl=5m" }}
          {{ .Cert }}
          {{ .Key }}
          {{ end }}
        EOF
        destination = "secrets/mongo.bundle.pem"
        change_mode = "script"
        change_script {
          command = "/local/bin/rotate-cert.sh"
        }
      }
```
Using a short TTL like in this example makes the problem more visible, but it's still quite random

#### Expected Result
The /local/bin/rotate-cert.sh script is fired everytime the cert is renewed

#### Actual Result
Most of the time, it's working, but sometimes (couldn't find any common pattern), the script fails with
```
Template failed to run script /local/bin/rotate-cert.sh with arguments [] because task driver doesn't support the exec operation
```

As a workaround, I've added the fail_on_error stanza
```
        change_script {
          command = "/local/bin/rotate-cert.sh"
          # If cert rotation fails, kill the task and restart it
          fail_on_error = true
        }
```
So when this happens, the task is killed and restarted (so it gets it's new cert)

### Nomad Client logs (if appropriate)
Nothing interesting, we can see the template being rendered, and the task being stopped (because of the fail_on_error)
```
janv. 23 10:14:54 ct-wn-3 nomad[3202532]:     2023-01-23T10:14:54.138+0100 [INFO]  agent: (runner) rendered "(dynamic)" => "/opt/nomad/data/alloc/9b688daf-6269-eef6-1752-4f1db4413b3f/tracking-mongod/secrets/mongo.bundle.pem"
[...]

PR #23663, opened 2024-07-22
template: allow change_mode script to run after client restart

tgross:
For templates with `change_mode = "script"`, we set a driver handle in the poststart method, so the template runner can execute the script inside the task. But when the client is restarted and the template contents change during that window, we trigger a change_mode in the prestart method. In that case, the hook will not have the handle and so returns an errror trying to run the change mode.

We restore the driver handle before we call any prestart hooks, so we can pass that handle in the constructor whenever it's available. In the normal task start case the handle will be empty but also won't be called.

The error messages are also misleading, as there's no capabilities check happening here. Update the error messages to match.

Fixes: https://github.com/hashicorp/nomad/issues/15851
Ref: https://hashicorp.atlassian.net/browse/NET-9338

### DIFF
--- a/.changelog/23663.txt
+++ b/.changelog/23663.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+template: Fixed a bug where change_mode = "script" would not execute after a client restart
+```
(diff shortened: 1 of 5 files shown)


gulducat (approved):
LGTM!

PR #23673, opened 2024-07-23
scaling: fix state store corruption bug for job scaling events

tgross:
When updating a `JobScalingEvent`, the state store function did not copy the existing object before mutating it. This corrupts the state store because it modifies the leaf node without committing it in a transaction. It can also cause the Nomad server to crash with a "fatal error: concurrent map read and map write" if its `ScalingEvents` map is read via the `ScaleStatus` RPC at the same time as it's being written.

This changeset also removes some mostly-unused public methods on the struct that dangerously encourage you to mutate it outside of a copy.

Ref: https://hashicorp.atlassian.net/browse/NET-10529

### DIFF
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -6096,6 +6096,21 @@ type JobScalingEvents struct {
 	ModifyIndex uint64
 }
 
+func (j *JobScalingEvents) Copy() *JobScalingEvents {
+	if j == nil {
+		return nil
+	}
+	njse := new(JobScalingEvents)
+	*njse = *j
+
+	njse.ScalingEvents = map[string][]*ScalingEvent{}
+	for taskGroup, events := range j.ScalingEvents {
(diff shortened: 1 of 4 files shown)


schmichael (approved):

nomad/structs/structs.go
```diff
@@ -6096,6 +6096,21 @@ type JobScalingEvents struct {
...
 }
 
+func (j *JobScalingEvents) Copy() *JobScalingEvents {
+	if j == nil {
+		return nil
+	}
+	njse := new(JobScalingEvents)
+	*njse = *j
+
+	njse.ScalingEvents = map[string][]*ScalingEvent{}
```
jrasell, line 6106: We could probably initialise the map length using the existing object, although I don't anticipate this will give us any real noticeable performance improvements.

jrasell (approved):
LGTM, thanks @tgross!

PR #23674, opened 2024-07-23
reporting: allow export interval and address to be configurable

tgross:
The internal library for [automated license utilization reporting](https://developer.hashicorp.com/nomad/docs/enterprise/license/utilization-reporting) supports configuration to send metrics to a local development version of the collector. Add "undocumented" configuration options to the `reporting` block allow developers to debug and verify we're sending the data we expect with real Nomad servers and not just unit tests.

Note this behavior does not impact Nomad Community Edition whatsoever, which does not collect telemetry outside of the incidental data from version upgrade checks ([Checkpoint](https://developer.hashicorp.com/nomad/docs/faq#q-what-is-checkpoint-does-nomad-call-home)).

Ref: https://hashicorp.atlassian.net/browse/NET-10057
Ref: https://github.com/hashicorp/nomad-enterprise/pull/1708

### DIFF
--- a/command/agent/config_parse.go
+++ b/command/agent/config_parse.go
@@ -179,6 +179,8 @@ func ParseConfigFile(path string) (*Config, error) {
 				c.Client.TemplateConfig.NomadRetry.MaxBackoff = d
 			},
 		},
+		{"reporting.export_interval",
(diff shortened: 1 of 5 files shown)


gulducat (approved):
:+1:

jrasell (approved):
LGTM, thanks @tgross!

Issue #23510, opened 2024-07-06
Workload identity : lack of usable user_claim when using Nomad namespaces and Vault entities


the-nando:
I'm working on migrating some clusters from the legacy Vault token based integration to the new workload identity based one.

My aim is to be able to create a single Vault entity per workload, set entity specific policies and use that in addition to the generic role's token policy.

The [tutorial](https://developer.hashicorp.com/nomad/tutorials/integrate-vault/vault-acl) suggests to use `"user_claim": "/nomad_job_id"` and a templated Vault policy utilising the claim mapped metadata, something along the lines of:

```hcl
path "secrets/data/{{identity.entity.aliases.AUTH_METHOD_ACCESSOR.metadata.nomad_namespace}}/{{identity.entity.aliases.AUTH_METHOD_ACCESSOR.metadata.nomad_job_id}}" {
  capabilities = ["read"]
}
```

To cater for jobs which may require additional ad-hoc policies, I want to pre-create Vault identities for workloads that will have one or more additional identity policies.  
To get this to work I would use an entity-alias based on the `user_claim` to map it to that entity. This would allow me to setup a default token workload policy, like in the tutorial, with templated paths and for any exception I could just create a policy with the same name as the one we assign to the entity.

The problem is that the `user_claim` isn't unique when one uses `/nomad_job_id` in combination with Nomad namespace as the Job ID isn't unique within a Nomad cluster.  
The implication on the Vault side is that any job by the same name will get assigned the same implied identity which is a potential security risk and that could lead to unintended access to Vault resources.

A workaround is to create a Vault JWT role per workload and configure `bound_claims`:

```hcl
"bound_claims": {
  "nomad_namespace": "myns",
  "nomad_job_id": "myjob"
}
```

But this invalidates completely the features of Vault entity management. Furthermore, to my knowledge, a JWT user claim must be unique within the system. It would be perhaps better to recommend users to use `"user_claim": "/sub" ` if they don't intend to use `bound_claims`.

What I would like, is to be able to use a unique claim, something like `nomad_workload_id: "<namespace>:::<job_id>"` which can then be leverage on the Vault side to configure entities and aliases accordingly. `"/sub"` wouldn't work as it contains additional details, like region/taskgroup/task/identity, which are something Vault operator may not know upfront for each job.
Can such user_claim be made available?

tgross:
[...]

Issue #19438, opened 2023-12-11
Add arbitrary claims to a job's workload identity


vftaylor:
Exposing Nomad workload identity JWTs inside jobs greatly enhances the flexibility that operators have when configuring jobs to authenticate to external systems like Vault, Consul etc. Currently, a job's identity JWT is limited to 4 main claims: `nomad_namespace`, `nomad_job_id`, `nomad_allocation_id`, `nomad_task`. While a good start, even more flexibility could be obtained if an operator could configure Nomad to add arbitrary keys from the jobspec as claims to a job's identity JWT. Even allowing a subset of the jobspec keys, especially the "Meta" key, would be good enough.

### Proposal

Within the new `identity` block, allow an operator to specify arbitrary keys from the jobspec that would be added as claims to a job's JWT. The primary use case for me would be to have the contents of the "Meta" block for a job baked into the JWT. This would allow dramatically more flexibility when using templated Vault policies, since a richer source of metadata would be available for use. Example jobspec:

    job "docs" {
      group "example" {
        task "api" {

          ####
          other stuff
          ####

          identity {
            name         = "example"
            aud          = ["oidc.example.com"]
            file         = true
            ttl          = "1h"
            extra_claims = ["job.meta"]       <----------------- some suitable way to identify keys of interest
          }
        }
      }
      meta {
        foo = bar
      }
    }

Would give the following JWT:

    {
      "aud": "oidc.example.com",
      "exp": 1702121612,
      "iat": 1702118012,
      "jti": "8a5f5fb9-2a0c-a4f7-7583-b75c4a1b0766",
      "nbf": 1702118012,
      "nomad_allocation_id": "c30140d4-6106-2350-ff90-6cc9a3b9e7ab",
      "nomad_job_id": "docs",
      "nomad_namespace": "default",
      "nomad_task": "api",
      "sub": "foo",
      "job.meta": {
        "foo": "bar"               <-------------------------- and it comes out as a claim
      }
    }

### Use-cases

- Have much more control over using Vault policies that are templated with metadata claims.
- Provide richer information to external systems by securely passing the JWTs from Nomad jobs.

### Attempted Solutions

There's no real way to satisfy this requirement using existing Nomad features. This functionality requires cryptographic guarantees from the workload identity engine itself.

tgross:
[...]

PR #23675, opened 2024-07-23
workload identity: add support for extra claims config for Vault

tgross:
Although we encourage users to use Vault roles, sometimes they're going to want to assign policies based on entity and pre-create entities and aliases based on claims. This allows them to use single default role (or at least small number of them) that has a templated policy, but have an escape hatch from that.

When defining Vault entities the `user_claim` must be unique. When writing Vault binding rules for use with Nomad workload identities the binding rule won't be able to create a 1:1 mapping because the selector language allows accessing only a single field. The `nomad_job_id` claim isn't sufficient to uniquely identify a job because of namespaces. It's possible to create a JWT auth role with `bound_claims` to avoid this becoming a security problem, but this doesn't allow for correct accounting of user claims.

 Add support for an `extra_claims` block on the server's `default_identity` blocks for Vault. This allows a cluster administrator to add a custom claim on all allocations. The values for these claims are interpolatable with a limited subset of fields, similar to how we interpolate the task environment.

Fixes: https://github.com/hashicorp/nomad/issues/23510
Ref: https://hashicorp.atlassian.net/browse/NET-10372
Ref: https://hashicorp.atlassian.net/browse/NET-10387

Note to reviewers: this PR targets the branch of https://github.com/hashicorp/nomad/pull/23708 so that it includes all those changes, but I expect the refactoring PR will merge first and then this PR will target `main` again.

### DIFF
--- a/nomad/structs/workload_id.go
+++ b/nomad/structs/workload_id.go
@@ -76,6 +76,8 @@ type IdentityClaims struct {
 	VaultNamespace  string `json:"vault_namespace,omitempty"`
 	VaultRole       string `json:"vault_role,omitempty"`
 
+	ExtraClaims map[string]string `json:"extra_claims,omitempty"`
+
 	jwt.Claims
 }
 
@@ -93,6 +95,8 @@ type IdentityClaimsBuilder struct {
 	serviceName string
 	consul      *Consul
 	vault       *Vault
+	node        *Node
+	extras      map[string]string
 }
 
 // NewIdentityClaimsBuilder returns an initialized IdentityClaimsBuilder for the
@@ -111,6 +115,7 @@ func NewIdentityClaimsBuilder(job *Job, alloc *Allocation, wihandle *WIHandle, w
 		wihandle: wihandle,
 		wid:      wid,
 		tg:       tg,
+		extras:   map[string]string{},
 	}
 }
 
@@ -125,11 +130,14 @@ func (b *IdentityClaimsBuilder) WithTask(task *Task) *IdentityClaimsBuilder {
 
 // WithVault adds the task's vault block to the builder context. This should
 // only be called after WithTask.
-func (b *IdentityClaimsBuilder) WithVault() *IdentityClaimsBuilder {
+func (b *IdentityClaimsBuilder) WithVault(extraClaims map[string]string) *IdentityClaimsBuilder {
 	if !b.wid.IsVault() || b.task == nil {
 		return b
 	}
 	b.vault = b.task.Vault
+	for k, v := range extraClaims {
+		b.extras[k] = v
+	}
 	return b
 }
 
@@ -162,11 +170,18 @@ func (b *IdentityClaimsBuilder) WithService(service *Service) *IdentityClaimsBui
 	return b
 }
 
+// WithNode add the allocation's node to the builder context.
+func (b *IdentityClaimsBuilder) WithNode(node *Node) *IdentityClaimsBuilder {
+	b.node = node
+	return b
+}
+
 // Build is the terminal method for the builder and sets all the derived values
 // on the claim. The claim ID is random (nondeterministic) so multiple calls
 // with the same values will not return equal claims by design. JWT IDs should
 // never collide.
 func (b *IdentityClaimsBuilder) Build(now time.Time) *IdentityClaims {
+	b.interpolate()
 
 	jwtnow := jwt.NewNumericDate(now.UTC())
 	claims := &IdentityClaims{
@@ -178,6 +193,7 @@ func (b *IdentityClaimsBuilder) Build(now time.Time) *IdentityClaims {
 			NotBefore: jwtnow,
 			IssuedAt:  jwtnow,
 		},
+		ExtraClaims: b.extras,
 	}
 	// If this is a child job, use the parent's ID
 	if b.job.ParentID != "" {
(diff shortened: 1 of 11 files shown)


tgross:
@the-nando if you have a moment, take a look at the E2E test I've added here and see if that looks vaguely like what your team is looking to do.

nomad/structs/structs_test.go
```diff
@@ -8190,18 +8191,20 @@ func TestNewIdentityClaims(t *testing.T) {
...
 		"job/group/task/default-identity": {
-			Namespace: "default",
-			JobID:     "job",
-			TaskName:  "task",
+			NamespacedJobID: "default::job",
+			Namespace:       "default",
+			JobID:           "job",
+			TaskName:        "task",
 			Claims: jwt.Claims{
 				Subject:  "global:default:job:group:task:default-identity",
```
schmichael, line 8199: Why isn't the `sub` claim sufficient for this use case?

nomad/structs/structs.go
```diff
@@ -11725,9 +11726,10 @@ func NewIdentityClaims(job *Job, alloc *Allocation, wihandle *WIHandle, wid *Wor
 
 	jwtnow := jwt.NewNumericDate(now.UTC())
 	claims := &IdentityClaims{
-		Namespace:    alloc.Namespace,
-		JobID:        alloc.JobID,
-		AllocationID: alloc.ID,
+		NamespacedJobID: alloc.Namespace + "::" + alloc.JobID,
```
schmichael, line 11729: Is there a functional reason for double colons (`::`)? The `sub` claim uses a single colon as a separator.

Should we prepend `alloc.Region` as well for circumstances where multiple Nomad clusters use the same Vault instance?

nomad/structs/structs_test.go
schmichael: Just noticed OP addressed this somewhat so I asked for some more details on the issue: https://github.com/hashicorp/nomad/issues/23510#issuecomment-2249064479

schmichael (commented):
We should also update docs: https://developer.hashicorp.com/nomad/docs/concepts/workload-identity

They're also lacking the standard JWT claims like `aud` and `sub`. It would be nice to list out *all* of the claims Nomad adds with some explanation as the difference between `nomad_workload_id`, `nomad_job_id`, and `sub` are pretty subtle and have security implications for federated identity users.

tgross:
Moving to draft while the design gets further discussed in https://github.com/hashicorp/nomad/issues/23510

tgross:
Going to rework this PR to what's been described in https://github.com/hashicorp/nomad/issues/23510#issuecomment-2256865419

nomad/structs/workload_id.go
```diff
@@ -76,6 +76,8 @@ type IdentityClaims struct {
 	VaultNamespace  string `json:"vault_namespace,omitempty"`
 	VaultRole       string `json:"vault_role,omitempty"`
 
+	ExtraClaims map[string]string `json:"extra_claims,omitempty"`
```
schmichael, line 79:
```suggestion
	// ExtraClaims are added based on the this identity's WorkloadIdentityConfiguration 
	ExtraClaims map[string]string `json:"extra_claims,omitempty"`
```

nomad/structs/workload_id.go
```diff
@@ -203,6 +219,38 @@ func (b *IdentityClaimsBuilder) Build(now time.Time) *IdentityClaims {
...
 
+func strAttrGet[T any](x *T, fn func(x *T) string) string {
+	if x != nil {
+		return fn(x)
+	}
+	return ""
+}
+
+func (b *IdentityClaimsBuilder) interpolate() {
+
```
schmichael, line 230: nbd but might as well avoid creating replacers when there's nothing to replace

```suggestion
if len(b.extras) == 0 {
  return
}
```

website/content/docs/configuration/vault.mdx
```diff
@@ -186,6 +190,28 @@ will be removed in a future release.
 - `ttl` `(string: "")` - Specifies for how long the workload identity should be
   considered as valid before expiring.
 
+- `extra_claims` `(map[string]string: optional)` - A set of key-value pairs that
+  will be provided as extra identity claims for workloads. You can use the keys
+  as user claims in Vault role configurations. The values are interpolated. For
```
schmichael, line 195: Link `user claims in Vault role configurations` to https://developer.hashicorp.com/vault/api-docs/auth/jwt#user_claim

nomad/structs/workload_id.go
```diff
@@ -260,6 +308,9 @@ type WorkloadIdentity struct {
 	// TTL is used to determine the expiration of the credentials created for
 	// this identity (eg the JWT "exp" claim).
 	TTL time.Duration
+
+	// Note: ExtraClaims is available on config/WorkloadIdentity but not
+	// available here on jobspecs
```
schmichael, line 313: I was about to ask if this could neatly solve #19438 and noticed your comment over there.

I like shipping this PR as-is. My main concern was that if there was a strong reason *not* to expose the field here, perhaps for security reasons, we should call those out loudly.

schmichael (approved):
Thanks for sticking with this everyone! I think this solution looks great. Enables a specific useful workflow and doesn't leak into places its not needed. Gluing federated identities and auth methods together gets a bit mind bending with all these parameters and claims, but from looking at other federated identity implementations that seems unavoidable.

nomad/structs/workload_id.go
tgross: Extended the comment a bit more to expand on that reasoning.

tgross:
Going to wait until https://github.com/hashicorp/nomad/pull/23708 is reviewed and merged to merge this.

tgross:
Missing changelog added in https://github.com/hashicorp/nomad/pull/23817

PR #23686, opened 2024-07-25
Fix a Sentinel policy template

sofixa:
In my previous PR ( #23530 ) I made a small mistake - the starting multiline comment is missing an opening # on the second line, and thus gets interpreted as Sentinel code instead of a comment.


tgross (approved):
Made a minor syntax tweak to add a space between `#` and text. LGTM!

PR #23687, opened 2024-07-25
release: explicitly set Linux service file user and group param.

### LINKED ISSUE #23685: nomad.service lack of User, Group options
### Nomad version
```
Nomad v1.8.2
BuildDate 2024-07-16T08:50:09Z
Revision 7f0822c1e4f25907d9f60e2d595411950dd1bd28
```

### Issue
https://developer.hashicorp.com/nomad/tutorials/enterprise/production-deployment-guide-vm-with-consul
Documentation recommends to start nomad.service using `nomad` user, but actual Nomad does not have this configuration
image

### Reproduction steps
See the screenshot above

#### Expected Result
The `nomad.service` should contain 'User','Group'

#### Actual Result
It does not contain

jrasell:
The default root:root is used as this provides permissions to run both server and client agents. The comment details what changes can be made to operators if needed.

When running the service file prior to this change, root:root would be the default.

closes #23685

### DIFF
--- a/.release/linux/package/usr/lib/systemd/system/nomad.service
+++ b/.release/linux/package/usr/lib/systemd/system/nomad.service
@@ -26,6 +26,12 @@ After=network-online.target
 # StartLimitInterval = 10s
 
 [Service]
+
+# Nomad clients need to be run as "root" whereas Nomad servers should be run as
+# the "nomad" user. Please change this if needed.
+User=root
+Group=root
+
 Type=notify
 EnvironmentFile=-/etc/nomad.d/nomad.env
 ExecReload=/bin/kill -HUP $MAINPID


tgross (approved):
LGTM!

.release/linux/package/usr/lib/systemd/system/nomad.service
```diff
@@ -26,6 +26,12 @@ After=network-online.target
 # StartLimitInterval = 10s
 
 [Service]
+
+# Nomad clients need to be run as "root" whereas Nomad servers should be run as
+# the "nomad" user. Please change this if needed.
+User=root
+Group=root
+
```
EugenKon, line 34: It is not clear: is this `nomad.service` file used for nomad server or nomad client?

jrasell: Nomad provides a single Linux package. There is no differentiation at install time (apt/yum/dnf...) between server & client.

EugenKon: So users should tune this file to `nomad` user during EC2 instance/DO droplet provisioning for Nomad server.
Do I understand right?

jrasell: Yes (or overwrite it) as part of your configuration management step to setup Nomad and other services on the machine.

Issue #13669, opened 2022-07-11
Rootless Nomad


tgross:
Nomad client agents [must be run as root](https://www.nomadproject.io/docs/install/production/requirements#linux-capabilities). The notion of "rootless" containers has worked its way through the container ecosystem. This issue is a bit of a brain-dump to assemble some thoughts and discussion around running Nomad "rootless". Please note this isn't yet a roadmap item or even a promise that Nomad will ever support rootless operation. If we decide to pursue this direction, we'd then engage in a design process (RFC) before we could start work on this.

## What is Rootless?

Rootless operation has several criteria:
1. The container orchestrator (ex. Nomad client agent or k8s kubelet) is not running as `root`.
2. The container runtime (ex. `dockerd`, `podman`) is not running as `root`.
3. The `root` user inside the container cannot be mapped to the `root` user on the host.

User-namespace mapping (criteria 3) alone can already be done by Nomad for some task drivers, so this issue is primarily focused on running Nomad itself as an unprivileged user.

## Why Rootless?

Container runtimes and orchestrators need to perform privileged operations normally reserved to the `root` user (or to a user that can escalate via `sudo` or `doas`):

* Resource isolation via cgroups
* Namespace isolation: mount, user, pid, ipc, and network namespaces
* Allowing inbound and outbound network traffic to the workload's network namespace.

Therefore running rootless containers has two primary use cases:
* Running containers as an unprivileged user for interactive use. For single-user machines like your typical developer laptop, being able to run sandboxed software as a normal user is handy. This use case doesn't have much overlap with Nomad, as Nomad tasks aren't interactive.
* Reduce the attack surface of the container orchestrator and/or runtime. These are complicated pieces of software with network access. Running them as root exposes a large attack surface if the application is compromised. This is more interesting for us.

## Requirements for Rootless

Given the set of privileged operations needed described above, there are some specific requirements for rootless containers:

* Delegated cgroups: to safely set cgroups as an unprivileged user requires cgroups v2.
* User namespaces: on some distros this may require setting sysctls like `kernel.unprivileged_userns_clone=1`
* Creating veth pairs across network namespaces (currently implemented with CNI)
* Creating iptables rules for inbound traffic
[...]

PR #23694, opened 2024-07-26
client/allocdir: ignore mount permission errors if root

stuart-warren:
ref: #23619

On some systems (eg snaps) nomad and docker run in their own mount namespaces which means that dockerd can't see mount points created by nomad.

By a server admin denying the permission for nomad to mount tmpfs they can make the decision to keep nomad working.

There should be no other reason for root to fail to do so due to a permission error.

### DIFF
--- a/client/allocdir/fs_linux.go
+++ b/client/allocdir/fs_linux.go
@@ -78,6 +78,13 @@ func createSecretDir(dir string) error {
 		flags := uintptr(syscall.MS_NOEXEC)
 		options := fmt.Sprintf("size=%dm", secretDirTmpfsSize)
 		if err := syscall.Mount("tmpfs", dir, "tmpfs", flags, options); err != nil {
+			if err == syscall.EPERM {
+				// If we don't have permission to mount tmpfs as root,
+				// just settle for MkdirAll() and continue.
+				// This might be blocked if it in undesirable to create
+				// a mount point in a separate namespace to dockerd
+				return nil
+			}
 			return os.NewSyscallError("mount", err)
 		}
 


tgross (commented):
Hi @stuart-warren. You can't simply ignore errors like this... this _silently_ leaks secrets to disk that would otherwise be in-memory, forcing the job author to somehow know that the client is being run in an unsupported fashion and to never use the secrets dir (or accept that risk).

stuart-warren:
sure, i can understand that.

perhaps your documentation could be updated to say when it won't use tmpfs (running as a non-root user) rather than
> Where possible it is backed by an in-memory filesystem and mounted noexec
?

https://developer.hashicorp.com/nomad/docs/runtime/environment#secrets

Would a Warn level log message be sufficient, or should it be explicitly configured?

I appreciate your speedy reviews :slightly_smiling_face: 

Thanks

tgross:
Hi @stuart-warren, I took this back to the team to discuss a bit...

Ignoring the error in a new Nomad release would mean existing cluster administrators would have a new error condition to deal with which should be and has been reasonably handled by Nomad with useful messages. Having it just be a logged error wouldn't make it visible to job authors -- the Nomad security model's [Personas](https://developer.hashicorp.com/nomad/docs/concepts/security#personas) assume these are different people. And we don't feel we can simply change this and expect cluster admins to notice updated documentation in the case this error is caused for any other reason than the one you mention. Updating the docs to describe behavior changes if you're not running as root implicitly endorses doing so, which we don't support.

If we were to support running Nomad clients as unprivileged users in the future (ref https://github.com/hashicorp/nomad/issues/13669), it's the sort of thing where we'd need to re-architecture the security model for the client from top-to-bottom. We can't "whittle" our way towards supporting it. Unfortunately, we cannot reasonably accept this suggested change.

stuart-warren:
Thank you for explaining. much appreciated :+1:

Issue #6778, opened 2019-11-26
[feature request] ui/cli: keep original hcl job spec for inspection


fredrikhgrelland:
### Nomad version
v0.10.1

### Operating system and Environment details
Centos 7.7

### Issue
Posting a hcl job spec to nomad parses it leaves a _human_ unreadable json, including all implied default stanzas. I would like to extract the original specification from nomad for inspection. Currently this is not possible?

### Job file (if appropriate)
Using a slightly changed example file from https://www.nomadproject.io/docs/job-specification/index.html

```
# This declares a job named "docs". There can be exactly one
# job declaration per job file.
job "docs" {
  # Specify this job should run in the region named "us". Regions
  # are defined by the Nomad servers' configuration.
  #region = "us"

  # Spread the tasks in this job between us-west-1 and us-east-1.
  datacenters = ["green"]

  # Run this job as a "service" type. Each job type has different
  # properties. See the documentation below for more examples.
  type = "service"

  # Specify this job to have rolling updates, two-at-a-time, with
  # 30 second intervals.
  update {
    stagger      = "30s"
    max_parallel = 2
  }

  # A group defines a series of tasks that should be co-located
  # on the same client (host). All tasks within a group will be
  # placed on the same host.
  group "webs" {
    # Specify the number of these tasks we want.
    count = 5

    # Create an individual task (unit of work). This particular
    # task utilizes a Docker container to front a web application.
    task "frontend" {
      # Specify the driver to be "docker". Nomad supports
      # multiple drivers.
      driver = "docker"

      # Configuration is specific to each driver.
      config {
        image = "hashicorp/web-frontend"
      }

      # The service block tells Nomad how to register this service
      # with Consul for service discovery and monitoring.
      service {
        # This tells Consul to monitor the service on the port
        # labelled "http". Since Nomad allocates high dynamic port
        # numbers, we use labels to refer to them.
        port = "http"

        check {
          type     = "http"
          path     = "/health"
          interval = "10s"
          timeout  = "2s"
        }
      }

      # It is possible to set environment variables which will be
      # available to the task when it runs.
      env {
        "DB_HOST" = "db01.example.com"
        "DB_USER" = "web"
        "DB_PASS" = "loremipsum"
      }

      # Specify the maximum resources required to run the task,
      # include CPU, memory, and bandwidth.
      resources {
        cpu    = 500 # MHz
        memory = 128 # MB

        network {
          mbits = 100

          # This requests a dynamic port named "http". This will
[...]

PR #23699, opened 2024-07-26
cli: add option to return original HCL in `job inspect` command

tgross:
In 1.6.0 we shipped the ability to review the original HCL in the web UI, but didn't follow-up with an equivalent in the command line. Add a `-hcl` flag to the `job inspect` command.

Closes: https://github.com/hashicorp/nomad/issues/6778

### DIFF
--- a/command/job_inspect.go
+++ b/command/job_inspect.go
@@ -34,10 +34,17 @@ General Options:
 Inspect Options:
 
   -version <job version>
-    Display the job at the given job version.
+    Display the job at the given job version. Defaults to current version.
 
   -json
-    Output the job in its JSON format.
+    Output the job in its JSON format. Cannot be used with -hcl.
+
+  -hcl
+    Output the original HCL submitted with the job. Cannot be used with -json.
+
+  -with-vars
+    Include the original HCL2 variables submitted with the. Can only be used
+    with -hcl.
 
   -t
     Format and display job using a Go template.
@@ -52,9 +59,11 @@ func (c *JobInspectCommand) Synopsis() string {
 func (c *JobInspectCommand) AutocompleteFlags() complete.Flags {
 	return mergeAutocompleteFlags(c.Meta.AutocompleteFlags(FlagSetClient),
 		complete.Flags{
-			"-version": complete.PredictAnything,
-			"-json":    complete.PredictNothing,
-			"-t":       complete.PredictAnything,
+			"-version":   complete.PredictAnything,
+			"-hcl":       complete.PredictNothing,
+			"-with-vars": complete.PredictNothing,
+			"-json":      complete.PredictNothing,
+			"-t":         complete.PredictAnything,
 		})
 }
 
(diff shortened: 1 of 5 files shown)


command/job_inspect.go
```diff
@@ -90,6 +101,14 @@ func (c *JobInspectCommand) Run(args []string) int {
 	}
 	args = flags.Args()
 
+	if hcl && json {
+		c.Ui.Error("can only use one of -hcl or -json")
+		return 1
+	}
+	if withVars && !hcl {
+		c.Ui.Error("can only use -with-vars with -hcl")
```
philrenaud, line 109: Is there a reason to `return 1` in the hcl+json flag error, but not in the withVars-hcl flag error?

command/job_inspect.go
```diff
@@ -206,3 +240,58 @@ func getJob(client *api.Client, namespace, jobID string, version *uint64) (*api.
...
+	}
+	runArgs = append(runArgs, jobID)
+
+	return fmt.Sprintf(`
+To run this job as originally submitted:
+
+$ nomad job inspect -namespace %s -hcl %s |
+    nomad job run %s
+`, namespace, jobID, strings.Join(runArgs, " "))
+
```
philrenaud, line 296: Absolutely love this DX

command/job_inspect.go
tgross: Nope, that's just a miss on my part. Will fix.

command/job_inspect.go
```diff
@@ -34,10 +34,17 @@ General Options:
...
 
   -json
-    Output the job in its JSON format.
+    Output the job in its JSON format. Cannot be used with -hcl.
+
+  -hcl
+    Output the original HCL submitted with the job. Cannot be used with -json.
+
+  -with-vars
+    Include the original HCL2 variables submitted with the. Can only be used
```
shoenig, line 46: with the what?

tgross: I wanted to leave you in suspense. :grin: Will fix.

command/job_inspect.go
```diff
@@ -206,3 +241,58 @@ func getJob(client *api.Client, namespace, jobID string, version *uint64) (*api.
...
+		return nil, fmt.Errorf("job %q with version %d couldn't be found", jobID, version)
+	}
+	return submission, err
+}
+
+func getWithVarsOutput(namespace, jobID string, uiVars string, varsMap map[string]string) string {
+	runArgs := []string{}
+	if namespace != "" {
+		runArgs = append(runArgs, "-namespace")
+		runArgs = append(runArgs, namespace)
```
shoenig, line 273: nit: could just be `append(runArgs, "-namespace", namespace)

command/job_inspect.go
tgross: That makes me realize I forgot docs and a changelog. Will fix that too.

command/job_inspect.go
```diff
@@ -206,3 +241,58 @@ func getJob(client *api.Client, namespace, jobID string, version *uint64) (*api.
...
+	}
+
+	if len(varsMap) != 0 {
```
shoenig, line 276: the length check is redundant - the `range` will do nothing if the map is empty anyway

Issue #23693, opened 2024-07-26
Only one IP is recognized on host_network


sbenoistmics:
### Nomad version
Nomad v1.8.0
BuildDate 2024-05-28T17:38:17Z
Revision 28b82e4b2259fae5a62e2ed47395334bea5a24c4

### Operating system and Environment details
Linux XXX 5.10.0-31-amd64 #1 SMP Debian 5.10.221-1 (2024-07-14) x86_64 GNU/Linux

### Issue
Even if you have two valid IP on the host network, only the first one is used for placement.
You cannot use both IP to allow application with non configurable port to be on the same host  

### Reproduction steps
In nomad client config add the network :
```
  host_network "scylla" {
    cidr           = "10.0.1.248/30"
    reserved_ports = "22,80"
  }
```
Have a host with two IP in that cidr configured, e.g.:
```
4: eth2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    link/ether 08:00:27:9b:e1:25 brd ff:ff:ff:ff:ff:ff
    altname enp0s9
    inet 10.0.1.249/24 brd 10.0.1.255 scope global eth2
       valid_lft forever preferred_lft forever
    inet6 fe80::a00:27ff:fe9b:e125/64 scope link 
       valid_lft forever preferred_lft forever
5: eth3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    link/ether 08:00:27:55:be:98 brd ff:ff:ff:ff:ff:ff
    altname enp0s10
    inet 10.0.1.250/24 brd 10.0.1.255 scope global eth3
       valid_lft forever preferred_lft forever
    inet6 fe80::a00:27ff:fe55:be98/64 scope link 
       valid_lft forever preferred_lft forever
```
Submit the job spec below

#### Expected Result
Nomad deploy app on port 7000 on both IP 10.0.1.249 and 10.0.1.250 

#### Actual Result
Nomad fail to deploy with error below :
```
reserved port collision scylla=7000
```

### Job file (if appropriate)
```
job "test-network" {
  group "fake-scylla" {
    count = 2
    network {
      port "scylla" {
        static       = 7000
        host_network = "scylla"
      }
    }

    task "server" {
      driver = "exec"

      config {
        command = "sleep"
        args = ["infinity"]
      }

      resources {
[...]

PR #23702, opened 2024-07-29
Allow static ports duplication on network host with multiple matches

sbenoistmics:
When a client add two IP matching the same network host, it should be able to used both for static ports
The primary used case is when an application with not configurable port is deployed on a big bare metal host. It would be wasteful to not allow multiple time the same host to be used.

Fixes: #23693

### DIFF
--- a/.changelog/23693.txt
+++ b/.changelog/23693.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+networking: Fixed a bug where if a client have two IP on the same network host, only one can provide the same static port
+```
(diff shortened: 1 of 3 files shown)


.changelog/23693.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+networking: Fixed a bug where if a client have two IP on the same network host, only one can provide the same static port
```
gulducat, line 2: It's a little difficult to understand the improved outcome here.  how about this?

```suggestion
networking: The same static port can now be used more than once on host networks with multiple IPs
```

gulducat (approved):
This looks great, thanks for the contribution!

Let me know what you think of the changelog wording, and we'll get this merged.

.changelog/23693.txt
sbenoistmics: I had trouble to express it too, thank you for the wording

PR #23708, opened 2024-07-30
refactor identity claims constructor to builder

tgross:
Upcoming work to add extensibility to identity claims for Vault (ref #23510) will require exposing server configuration and more objects from state to the process of creating an `IdentityClaims` struct.

Depending on how we inject these parameters into the constructor, we end up creating circular dependencies or a lot more logic in the setup in the plan applier and alloc endpoint. There are three contexts where we call `NewIdentityClaims`: the plan applier (where we only care about the default identity), signing task identities, and signing service identities. Each needs different parameters. So we'll refactor the constructor as a builder with methods that the caller can decide to use (or not) depending on context. I've pulled this work out of #23675 to make it easier to review separately.

Ref: https://github.com/hashicorp/nomad/issues/23510
Ref: https://github.com/hashicorp/nomad/pull/23675
Ref: https://hashicorp.atlassian.net/browse/NET-10372
Ref: https://hashicorp.atlassian.net/browse/NET-10387

---

Note for reviewers: #23675 has some new consumers of this API that'll make it a little more clear why a functional options approach wasn't desirable here. We don't necessarily have all the options ready at once unless we allocate a slice of function pointers to append to, and it makes the already "meh" ordering requirements here easier to mess up.


shoenig (approved):
LGTM!

Issue #20131, opened 2024-03-13
allow `tls13` for `tls_min_version`


EugenKon:
### Issue

https://developer.hashicorp.com/nomad/docs/configuration/tls

![image](https://github.com/hashicorp/nomad/assets/139780495/6d89cd94-c801-459b-8d6c-2d43dd5e8313)

I see that consul already supports TLS_v1.3: https://developer.hashicorp.com/consul/docs/agent/config/config-files#tlsv1_3
I suppose that Nomad also already supports it, but this was not documented yet

#### Expected Result

Document, that `tls13` already supported.

tgross:
That's not a documentation typo. Nomad's tlsutil package only currently accepts a `tls_min_version` as documented. That seems like an oversight though, so I'm going to rename this issue to cover updating the behavior instead.

PR #23713, opened 2024-07-31
tls: support setting min version to TLS1.3

tgross:
Nomad already supports TLS1.3, but not as a minimum version configuration. Update our config validation to allow setting `tls_min_version` to 1.3. Update the documentation to match Vault and warn that the `tls_cipher_suites` field is ignored when TLS is 1.3

Fixes: https://github.com/hashicorp/nomad/issues/20131
Ref: https://hashicorp.atlassian.net/browse/NET-10530


pkazmierczak (approved):
LGTM!

dduzgun-security (approved):

PR #23738, opened 2024-08-05
security: fix write symlink escape on the same allocdir path

dduzgun-security:
Resolves symlink escape when unarchiving by removing existing paths within the same allocation directory which can occur by writing a header that points to a symlink that lives outside of the sandbox environment. This exploit requires first compromising the Nomad client agent at the source allocation.

Ref: [NET-10607](https://hashicorp.atlassian.net/browse/NET-10607) & https://github.com/hashicorp/nomad-enterprise/pull/1725

[NET-10607]: https://hashicorp.atlassian.net/browse/NET-10607?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changelog/23738.txt
+++ b/.changelog/23738.txt
@@ -1,3 +1,3 @@
 ```release-note:security
-security: Fix write symlink escape during unarchive by removing existing paths within the same allocdir.
+security: Fix symlink escape during unarchiving by removing existing paths within the same allocdir. Compromising the Nomad client agent at the source allocation first is a prerequisite for leveraging this issue.
 ```
--- a/.changelog/23738.txt
+++ b/.changelog/23738.txt
@@ -1,0 +1,3 @@
+```release-note:security
+security: Fix write symlink escape during unarchive by removing existing paths within the same allocdir.
+```
--- a/client/allocwatcher/alloc_watcher_unix_test.go
+++ b/client/allocwatcher/alloc_watcher_unix_test.go
@@ -136,6 +136,60 @@ func TestPrevAlloc_StreamAllocDir_BadSymlink_Linkname(t *testing.T) {
 	must.EqError(t, err, "archive contains symlink that escapes alloc dir")
 }
 
+func TestPrevAlloc_StreamAllocDir_SyminkWriteAttack(t *testing.T) {
+	ci.Parallel(t)
+
+	tmpDir := t.TempDir()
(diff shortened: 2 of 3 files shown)


.changelog/23738.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:security
+security: Fix write symlink escape during unarchive by removing existing paths within the same allocdir.
```
tgross, line 2: It'd be good for this to explain enough about the exploit path (i.e. you need to compromise the remote client first) so that users have enough context to know whether this is something they need to worry about in the short term.

client/allocwatcher/alloc_watcher_unix_test.go
```diff
@@ -136,6 +136,60 @@ func TestPrevAlloc_StreamAllocDir_BadSymlink_Linkname(t *testing.T) {
...
 
+func TestPrevAlloc_StreamAllocDir_SyminkWriteAttack(t *testing.T) {
+	ci.Parallel(t)
+
+	tmpDir := t.TempDir()
+	outsidePath := filepath.Join(tmpDir, "outside")
+	insidePath := "malformed_link"
+	content := "HelloWorld from outside"
+
+	// Create a tar archive with a symlink that attempts to escape the allocation directory
```
tgross, line 147: For future developers, it's important to understand that this isn't just as matter of a symlink which escapes the allocation directory, but one where the second header will write to the same path and follow the symlink out of the sandbox. We should expand upon that in this comment.

tgross (commented):
Looks great!

client/allocwatcher/alloc_watcher_unix_test.go
dduzgun-security: Good idea, added more details here https://github.com/hashicorp/nomad/pull/23738/commits/4b169330e97c48556713493c7ad644c59b4f30a1

.changelog/23738.txt
dduzgun-security: Very good point, updated the changelog to mention the prerequisite for exploiting this issue https://github.com/hashicorp/nomad/pull/23738/commits/326ff5f6a972acbb3a0bf252078e07f8d6d7665b

tgross (approved):
LGTM!

PR #23739, opened 2024-08-05
DOCS: CE-659 chroot limitations for isolated fork/exec driver

aimeeu:
[Jira issue](https://hashicorp.atlassian.net/browse/CE-659). 

Based on my research, this looks like a chroot limitation and not necessarily a Nomad issue (I could be completely wrong), so I added a new partial with a warning.

Deploy previews:
- [Fork/isolated chroot section](https://nomad-git-docs-ce659-hashicorp.vercel.app/nomad/docs/drivers/exec#chroot)
- [Client config chroot section](https://nomad-git-docs-ce659-hashicorp.vercel.app/nomad/docs/configuration/client#chroot_env-parameters)

### DIFF
--- a/website/content/partials/chroot-limitations.mdx
+++ b/website/content/partials/chroot-limitations.mdx
@@ -1,0 +1,3 @@
+!> **Warning:** chroot doesn't copy or recreate directories and files that systemd uses. Specifically on Ubuntu, `/etc/resolv.conf` is a symlink to `/run/systemd/resolve/stub-resolv.conf`, so chroot doesn't copy or link to `resolv.conf`.
+
+In these cases, configure your job's [network block](/nomad/docs/job-specification/network) for each chroot task.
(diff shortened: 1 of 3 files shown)


website/content/partials/chroot-limitations.mdx
```diff
@@ -0,0 +1,3 @@
+!> **Warning:** chroot doesn't copy or recreate directories and files that systemd uses. Specifically on Ubuntu, `/etc/resolv.conf` is a symlink to `/run/systemd/resolve/stub-resolv.conf`, so chroot doesn't copy or link to `resolv.conf`.
```
tgross, line 1: This is technically true but I feel like it's misleading if we refer to "directories and files that systemd uses", because most of the files that it uses _are_ in the default chroot (ex. everything in `/etc` and `/usr`!). What we're not copying -- by default -- are the ephemeral runtime files in `/run`.

tgross (approved):
LGTM

Issue #14101, opened 2022-08-13
support ipv6 connectivity for the bridge network


sinisterstumble:
It seems there is simply no way to setup ipv6 connectivity for the default nomad bridge.

Thank you!

jrasell:
Hi @markmartirosian and thanks for raising this feature request.

PR #23747, opened 2024-08-05
ipv6: allow ipv6 connectivity for bridge network

martisah:
This PR integrates IPv6 connectivity within the bridge network by adding IPv6 range configuration for bridge networking and a default IPv6 subnet IP address if no configuration is specified. 

Resolves https://github.com/hashicorp/nomad/issues/14101.

### DIFF
--- a/client/allocrunner/networking_bridge_linux.go
+++ b/client/allocrunner/networking_bridge_linux.go
@@ -6,6 +6,7 @@ package allocrunner
 import (
 	"context"
 	"fmt"
+
 	"github.com/coreos/go-iptables/iptables"
 	hclog "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/nomad/nomad/structs"
--- a/client/allocrunner/networking_bridge_linux.go
+++ b/client/allocrunner/networking_bridge_linux.go
@@ -6,6 +6,7 @@ package allocrunner
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/coreos/go-iptables/iptables"
 	hclog "github.com/hashicorp/go-hclog"
@@ -26,6 +27,7 @@ const (
 	// allocation when not specified by the client
 	defaultNomadAllocSubnet = "172.26.64.0/20" // end 172.26.79.255
 
+	defaultNomadAllocSubnetIPv6 = "2001:db8:1::/64" // docker ipv6 address to change
 	// cniAdminChainName is the name of the admin iptables chain used to allow
 	// forwarding traffic to allocations
 	cniAdminChainName = "NOMAD-ADMIN"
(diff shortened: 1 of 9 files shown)


client/allocrunner/networking_bridge_linux.go
```diff
@@ -195,10 +225,16 @@ const nomadCNIConfigTemplate = `{
 						{
 							"subnet": %q
 						}
+					],
+					[
+						{ 
+							"subnet": %q
```
gulducat, line 231: again I'm wondering if there's a way this could introduce problems on systems that don't ipv6.  have you tried disabling ipv6 on a test machine to see whether the bridge plugin errors trying to use this config?

the bridge plugin does not like it, then we'll need to generate this template differently somehow, so it can conditionally add ipv6 subnet/dst only if the system supports it.

client/allocrunner/networking_bridge_linux.go
```diff
@@ -195,10 +225,16 @@ const nomadCNIConfigTemplate = `{
...
+						}
 					]
 				],
 				"routes": [
-					{ "dst": "0.0.0.0/0" }
+					{ "dst": "0.0.0.0/0" },
+                    { "dst" : "::/0" }
```
gulducat, line 237: nitpick: indentation inconsistency between these
```suggestion
					{ "dst": "0.0.0.0/0" },
					{ "dst" : "::/0" }
```

client/allocrunner/networking_bridge_linux.go
```diff
@@ -207,6 +243,11 @@ const nomadCNIConfigTemplate = `{
 			"backend": "iptables",
 			"iptablesAdminChainName": %q
 		},
+		{
+			"type": "firewall",
+			"backend": "ip6tables",
```
gulducat, line 248: I remember adding this, but at the time I think we were kinda flinging spaghetti at the wall to see what stuck.  is this actually needed?  and again, is it ok on systems without ipv6?

command/agent/testdata/basic.hcl
```diff
@@ -105,6 +105,7 @@ client {
   cni_path              = "/tmp/cni_path"
   bridge_network_name   = "custom_bridge_name"
   bridge_network_subnet = "custom_bridge_subnet"
+  bridge_network_subnet_ipv6 = "custom_bridge_subnet_ipv6"
```
gulducat, line 108: nit: the `=`s should all be in alignment

gulducat (commented):
this looks really good!

I have a few comments, but mainly we'll need to:
* select a much narrower range for the default ipv6 subnet
* verify whether this works on a host with ipv6 disabled

command/agent/testdata/basic.hcl
gulducat: oh that might be what this test failure is complaining about: https://github.com/hashicorp/nomad/actions/runs/10255235476/job/28371795890?pr=23747#step:6:43

martisah: Ohh thank you! I was looking for what part of the linting I was missing 👍

client/allocrunner/networking_bridge_linux.go
```diff
@@ -88,17 +96,32 @@ func newBridgeNetworkConfigurator(log hclog.Logger, alloc *structs.Allocation, b
 // ensureForwardingRules ensures that a forwarding rule is added to iptables
 // to allow traffic inbound to the bridge network
 func (b *bridgeNetworkConfigurator) ensureForwardingRules() error {
-	ipt, err := iptables.New()
-	if err != nil {
-		return err
-	}
+	if b.allocSubnetIPv6 != "" {
```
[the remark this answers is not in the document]
martisah: Yea you're right, it's pretty useless here😅! I was thinking more specifically in the case that IPv6 is not enabled in the machine, I think this check will make more sense once I push some changes I just made.

client/allocrunner/networking_bridge_linux.go
martisah: Actually, just checked, there's no need for it! Thanks for the heads up :)

gulducat:
for simplicity, closing this one in favor of #23882

the main difference there, feature-wise, is that we do not supply a default IPv6 subnet, so the user must opt in to it by providing a subnet in the config.  that sidesteps many of the concerns I commented on here.  we can always choose to set a default later, but for now it's opt-in only.

Issue #23730, opened 2024-08-04
Nomad ACL Policy for variables can't accessed when defined with leading slash


rahadiangg:
### Nomad version
Nomad v1.8.2
BuildDate 2024-07-16T08:50:09Z
Revision 7f0822c1e4f25907d9f60e2d595411950dd1bd28

### Operating system and Environment details
Ubuntu 22.04
AWS t3.micro

### Issue
When definied acl policy with root trailing slash like below, the nomad task got 403 permission denied

```
namespace "default" {
  variables {
    path "/secret/aws/*" {
      capabilities = [
        "read",
        "list"
      ]
    }
  }
}
```

but when we remove the root trailing slash, the nomad task run as expedted

```
path "secret/aws/*"
```

### Reproduction steps
Assign policy that allow access to specified variables path with root trailing slash. E.g `nomad acl policy apply -namespace=default -job="2.access-variables-non-standard" allow-access-aws-secret.hcl`

#### Expected Result
Task running with rendered file from variables data

#### Actual Result
Pending task

### Job file (if appropriate)

<details>

```
job "2.access-variables-non-standard" {

  namespace = "default"

  group "dummy-group" {

    task "dummy-task" {
      driver = "exec"

      // driver configuration
      config {
        command = "/bin/bash"
        args = [
          "-c", "echo 'Task 1 is running'; sleep 3600"
        ]
      }

      resources {
        cpu    = 100 # MHz
        memory = 100 # MB
      }

      template {
        env         = false
        destination = "local/aws-config-demo"
        change_mode = "restart"
        data        = <<EOF
{{- with nomadVar "secret/aws/demo-account" }}
{{- range $key, $value := . }}
{{ $key }}={{ $value }}
{{- end }}
{{- end }}
EOF
      }

      template {
        env         = false
        destination = "local/aws-config-staging"
        change_mode = "restart"
        data        = <<EOF
{{- with nomadVar "secret/aws/staging-account" }}
{{- range $key, $value := . }}
{{ $key }}={{ $value }}
{{- end }}
{{- end }}
EOF
      }
    }
  }
}
```

</details>

### Nomad Server logs (if appropriate)

```
Aug 04 05:03:20 ip-172-31-0-241 nomad[5751]:     2024-08-04T05:03:20.493Z [DEBUG] http: request complete: method=GET path="/v1/var/secret/aws/staging-account?index=1&namespace=default&stale=&wait=300000ms" duration="205.364µs"
[...]

PR #23757, opened 2024-08-06
acl: disallow leading `/` on variable paths

tgross:
The path for a Variable never begins with a leading `/`, because it's stripped off in the API before it ever gets to the state store. The CLI and UI allow the leading `/` for convenience, but this can be misleading when it comes to writing ACL policies. An ACL policy with a path starting with a leading `/` will never match.

Update the ACL policy parser so that we prevent an incorrect variable path in the policy.

Fixes: https://github.com/hashicorp/nomad/issues/23730

### DIFF
--- a/acl/policy.go
+++ b/acl/policy.go
@@ -398,7 +398,7 @@ func Parse(rules string) (*Policy, error) {
 				}
 				if strings.HasPrefix(pathPolicy.PathSpec, "/") {
 					return nil, fmt.Errorf(
-						"Invalid variable path '%s' in namespace %s: cannot start with a leading '/'`",
+						"Invalid variable path %q in namespace %s: cannot start with a leading '/'`",
 						pathPolicy.PathSpec, ns.Name)
 				}
 				for _, cap := range pathPolicy.Capabilities {
--- a/acl/policy.go
+++ b/acl/policy.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
+	"strings"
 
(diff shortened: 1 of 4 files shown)


acl/policy.go
```diff
@@ -395,6 +396,11 @@ func Parse(rules string) (*Policy, error) {
 				if pathPolicy.PathSpec == "" {
 					return nil, fmt.Errorf("Invalid missing variable path in namespace %s", ns.Name)
 				}
+				if strings.HasPrefix(pathPolicy.PathSpec, "/") {
+					return nil, fmt.Errorf(
+						"Invalid variable path '%s' in namespace %s: cannot start with a leading '/'`",
```
schmichael, line 401:
```suggestion
						"Invalid variable path %q in namespace %s: cannot start with a leading '/'`",
```

acl/policy_test.go
```diff
@@ -494,6 +494,19 @@ func TestParse(t *testing.T) {
...
+			`
+			namespace "dev" {
+			  variables {
+                path "/nomad/job" {
+			      capabilities = ["read", "write"]
+                }
+			  }
+			}
+			`,
+			"Invalid variable path '/nomad/job' in namespace dev: cannot start with a leading '/'",
```
schmichael, line 507: If you accept the above change. nbd
```suggestion
			"Invalid variable path \"/nomad/job\" in namespace dev: cannot start with a leading '/'",
```

schmichael (approved):

Issue #23758, opened 2024-08-07
SignClaims : invalid memory address or nil pointer dereference


the-nando:
### Nomad version
```
Nomad v1.8.2+ent
BuildDate 2024-07-16T09:08:42Z
Revision f675c14f9ea2fada1fd1e5f1f5e63ce5d07f8cd4
```

### Issue

Nomad servers in one of our clusters started crashing out of the blue with:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x88 pc=0x1ff89b9]

goroutine 41361128 [running]:
github.com/hashicorp/nomad/nomad.(*Encrypter).SignClaims(0xc000ec54a0, 0x0)
	github.com/hashicorp/nomad/nomad/encrypter.go:206 +0x79
github.com/hashicorp/nomad/nomad.signAllocIdentities({0x3becc60, 0xc000ec54a0}, 0xc02f176008, {0xc02dc872b8, 0x1, 0xc000e86780?}, {0xdf7901?, 0xede4451aa?, 0x0?})
	github.com/hashicorp/nomad/nomad/plan_apply.go:424 +0x32d
github.com/hashicorp/nomad/nomad.(*planner).applyPlan(0xc000ef4340, 0xc02a6e8630, 0xc0336d4b00, 0xc00af3f5f0)
	github.com/hashicorp/nomad/nomad/plan_apply.go:281 +0x10ee
github.com/hashicorp/nomad/nomad.(*planner).planApply(0xc000ef4340)
	github.com/hashicorp/nomad/nomad/plan_apply.go:200 +0x7fa
created by github.com/hashicorp/nomad/nomad.(*Server).establishLeadership in goroutine 41361018
	github.com/hashicorp/nomad/nomad/leader.go:385 +0x19e
```

We've recovered it after multiple restarts of all the three servers.

### Reproduction steps

I'm not sure what triggered the issue, hopefully it's something someone has already seen.

For completeness and possibly not relevant: this cluster is still using the old token-based Vault and Consul integrations and there's an additional Vault cluster configured with the new integration but currently not in use.

tgross:
Ok @the-nando, I've had a first pass through. First, a question:

> We've recovered it after multiple restarts of all the three servers.

[...]

PR #23763, opened 2024-08-07
identity: prevent missing task identity from panicking server

tgross:
Tasks have a default identity created during canonicalization. If this default identity is somehow missing, we'll hit panics when trying to create and sign the claims in the plan applier. Fallback to the default identity if it's missing from the task.

This changeset will need a different implementation in 1.7.x+ent backports, as the constructor for identities was refactored significantly in #23708. The panic cannot occur in 1.6.x+ent.

Fixes: https://github.com/hashicorp/nomad/issues/23758

### DIFF
--- a/nomad/encrypter.go
+++ b/nomad/encrypter.go
@@ -242,6 +242,10 @@ const keyIDHeader = "kid"
 // SignClaims adds the Issuer claim prior to signing.
 func (e *Encrypter) SignClaims(claims *structs.IdentityClaims) (string, string, error) {
 
+	if claims == nil {
+		return "", "", errors.New("cannot sign empty claims")
+	}
+
 	// If a key is rotated immediately following a leader election, plans that
 	// are in-flight may get signed before the new leader has the key. Allow for
 	// a short timeout-and-retry to avoid rejecting plans
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -8055,9 +8055,10 @@ func (t *Task) Canonicalize(job *Job, tg *TaskGroup) {
 
 	// If there was no default identity, always create one.
 	if t.Identity == nil {
-		t.Identity = &WorkloadIdentity{}
+		t.Identity = DefaultWorkloadIdentity()
+	} else {
+		t.Identity.Canonicalize()
 	}
-	t.Identity.Canonicalize()
 }
 
 func (t *Task) GoString() string {
--- a/.changelog/23763.txt
+++ b/.changelog/23763.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+identity: Fixed a bug where a missing default task identity could panic the leader
+```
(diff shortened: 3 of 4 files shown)


nomad/encrypter.go
```diff
@@ -242,6 +242,10 @@ const keyIDHeader = "kid"
 // SignClaims adds the Issuer claim prior to signing.
 func (e *Encrypter) SignClaims(claims *structs.IdentityClaims) (string, string, error) {
 
+	if claims == nil {
+		return "", "", errors.New("cannot sign empty claims")
+	}
```
schmichael, line 247: How is this possible?! All of the callsites on main create the claims immediately before calling:
- https://github.com/hashicorp/nomad/blob/d131c41943cd8d2560ad044ad75981d2a0501457/nomad/alloc_endpoint.go#L639
- https://github.com/hashicorp/nomad/blob/d131c41943cd8d2560ad044ad75981d2a0501457/nomad/alloc_endpoint.go#L663
- https://github.com/hashicorp/nomad/blob/d131c41943cd8d2560ad044ad75981d2a0501457/nomad/alloc_endpoint.go#L675
- https://github.com/hashicorp/nomad/blob/d131c41943cd8d2560ad044ad75981d2a0501457/nomad/plan_apply.go#L424

If it's just to avoid future panics if code gets moved around that's fine. I just wanted to make sure I wasn't missing something.

nomad/structs/structs.go
```diff
@@ -8055,9 +8055,10 @@ func (t *Task) Canonicalize(job *Job, tg *TaskGroup) {
 
 	// If there was no default identity, always create one.
 	if t.Identity == nil {
-		t.Identity = &WorkloadIdentity{}
+		t.Identity = DefaultWorkloadIdentity()
```
schmichael, line 8058: I assume this is the fix and everything else is just defensive?

schmichael (approved):

nomad/encrypter.go
tgross: In the original issue https://github.com/hashicorp/nomad/issues/23758#issuecomment-2273636840, it turns out that the `NewIdentityClaims` constructor can return a `nil` pointer. This refactored version can't return `nil` anymore (but there are other defensive fixes in this PR). So yeah this is defensive. https://github.com/hashicorp/nomad-enterprise/pull/1755 has something closer to the original code.

nomad/structs/structs.go
tgross: The fix for this branch is the bit in `NewIdentityClaimsBuilder` where we set a nil `WorkloadIdentity` arg to the default WI: https://github.com/hashicorp/nomad/pull/23763/files#diff-d2ec4e0bb147da1d9abe8f14084958a2d4bc16b5cb1cf28f5a07c7f2bb3331e5R113

This bit isn't defensive, as the behavior is unchanged, but makes it more clear what we're doing. Originally we had an empty struct and populated it with the defaults via `Canonicalize`. Whereas this sets a nil pointer to the default WI without any further decisions to make.

Issue #14480, opened 2022-09-07
networking: allow cluster operators to specify their own bridge CNI config


jrasell:
The Nomad bridge [network mode](https://www.nomadproject.io/docs/job-specification/network#bridge) uses CNI to configure the interface via the [bridge plugin](https://www.cni.dev/plugins/current/main/bridge/#overview). The CNI plugin configuration is [embedded within the client](https://github.com/hashicorp/nomad/blob/main/client/allocrunner/networking_bridge_linux.go#L141-L180) with a small number of configuration options exposed on the [client configuration block](https://www.nomadproject.io/docs/configuration/client#bridge_network_name).

Whilst the configuration options offer enough scope for some, others would like to update fields which are not exposed. In order to add these in the current method, we would need to add these to the client configuration block. This could create an unwieldily set of config params and could also mean Nomad needs to continually update to accommodate new bridge plugin options.

It could prove useful and more flexible to allow operators to define their own CNI configuration to load for the bridge network, rather than add individual config parameters. This would immediately solve the problems mentioned within [this Discuss post](https://discuss.hashicorp.com/t/custom-mtu-for-bridge-network/44022) and https://github.com/hashicorp/nomad/issues/13352

NOBLES5E:
Is this being worked on? Right now we need to use our customized build of nomad to support hairpin mode, which is not ideal.

lgfa29:
Closing this one as completed by https://github.com/hashicorp/nomad/pull/15961.

While this issue mentions other bridge parameters, or a more flexible configuration approaches, we feel like adding more customization to the default `bridge` may result in unexpected outcomes that are hard for us to debug. The `bridge` network mode should something predictable and easily reproducible by the team.

Users that require more advanced customization are able to create their own bridge network using CNI. The main downside of this is that Consul Service Mesh requires `network_mode = "bridge"`, but this is a separate feature request that is being tracked in #8953.

Feel free to 👍 and add more comments there.

PR #23772, opened 2024-08-08
networking: refactor building nomad bridge config

martisah:
This PR refactors building the Nomad Bridge network configuration JSON, in order to allow for more manipulation of variables within the network configuration.

### DIFF
--- a/client/allocrunner/cni/bridge.go
+++ b/client/allocrunner/cni/bridge.go
@@ -1,0 +1,86 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package cni
+
+import "encoding/json"
+
+// Conflist is the .conflist format of CNI network config.
+type Conflist struct {
+	CniVersion string `json:"cniVersion"`
+	Name       string `json:"name"`
+	Plugins    []any  `json:"plugins"`
+}
+
+// Json produces indented json of the conflist.
+func (b Conflist) Json() ([]byte, error) {
+	return json.MarshalIndent(b, "", "\t")
+}
+
+// NomadBridgeConfig determines the contents of the Conflist.
+type NomadBridgeConfig struct {
+	BridgeName     string
+	AdminChainName string
+	IPv4Subnet     string
+	HairpinMode    bool
+	ConsulCNI      bool
+}
+
+// NewNomadBridgeConflist produces a full Conflist from the config.
+func NewNomadBridgeConflist(conf NomadBridgeConfig) Conflist {
+	// Update website/content/docs/networking/cni.mdx when the bridge config
+	// is modified. The json versions of the config can be found in
+	// client/allocrunner/test_fixtures/*.conflist.json
+	// If CNI plugins are added or versions need to be updated for new fields,
+	// add a new constraint to nomad/job_endpoint_hooks.go
+
+	ipRanges := [][]Range{
+		{{Subnet: conf.IPv4Subnet}},
+	}
+	ipRoutes := []Route{
+		{Dst: "0.0.0.0/0"},
+	}
+
+	plugins := []any{
+		Generic{
+			Type: "loopback",
+		},
+		Bridge{
+			Type:         "bridge",
+			Bridgename:   conf.BridgeName,
+			IpMasq:       true,
+			IsGateway:    true,
+			ForceAddress: true,
+			HairpinMode:  conf.HairpinMode,
+			Ipam: IPAM{
+				Type:   "host-local",
+				Ranges: ipRanges,
+				Routes: ipRoutes,
+			},
+		},
+		Firewall{
+			Type:           "firewall",
+			Backend:        "iptables",
+			AdminChainName: conf.AdminChainName,
+		},
+		Portmap{
+			Type: "portmap",
+			Capabilities: PortmapCapabilities{
+				Portmappings: true,
+			},
+			Snat: true,
+		},
+	}
+	if conf.ConsulCNI {
+		plugins = append(plugins, ConsulCNI{
+			Type:     "consul-cni",
+			LogLevel: "debug",
+		})
+	}
+
+	return Conflist{
+		CniVersion: "0.4.0",
+		Name:       "nomad",
+		Plugins:    plugins,
+	}
+}
--- a/client/allocrunner/cni/bridge_test.go
+++ b/client/allocrunner/cni/bridge_test.go
@@ -1,0 +1,31 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package cni
+
+import (
+	"testing"
+
+	"github.com/shoenig/test/must"
+)
+
+func TestConflist_Json(t *testing.T) {
+	conf := &Conflist{
+		CniVersion: "0.0.1",
+		Name:       "test-config",
+		Plugins: []any{
+			Generic{Type: "test-plugin"},
+		},
+	}
+	bts, err := conf.Json()
+	must.NoError(t, err)
+	must.Eq(t, `{
+	"cniVersion": "0.0.1",
+	"name": "test-config",
+	"plugins": [
+		{
+			"type": "test-plugin"
+		}
+	]
+}`, string(bts))
+}
--- a/client/allocrunner/cni/plugins.go
+++ b/client/allocrunner/cni/plugins.go
@@ -1,0 +1,58 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package cni
+
+// Generic has the one key that all plugins must have: "type"
+type Generic struct {
+	Type string `json:"type"`
+}
+
+// Bridge is the subset of options that we use to configure the "bridge" plugin.
+// https://www.cni.dev/plugins/current/main/bridge/
+type Bridge struct {
+	Type         string `json:"type"`
+	Bridgename   string `json:"bridge"`
+	IpMasq       bool   `json:"ipMasq"`
+	IsGateway    bool   `json:"isGateway"`
+	ForceAddress bool   `json:"forceAddress"`
+	HairpinMode  bool   `json:"hairpinMode"`
+	Ipam         IPAM   `json:"ipam"`
+}
+type IPAM struct {
+	Type   string    `json:"type"`
+	Ranges [][]Range `json:"ranges"`
+	Routes []Route   `json:"routes"`
+}
+type Range struct {
+	Subnet string `json:"subnet"`
+}
+type Route struct {
+	Dst string `json:"dst"`
+}
+
+// Firewall is the "firewall" plugin.
+// https://www.cni.dev/plugins/current/meta/firewall/
+type Firewall struct {
+	Type           string `json:"type"`
+	Backend        string `json:"backend"`
+	AdminChainName string `json:"iptablesAdminChainName"`
+}
+
+// Portmap is the "portmap" plugin.
+// https://www.cni.dev/plugins/current/meta/portmap/
+type Portmap struct {
+	Type         string              `json:"type"`
+	Capabilities PortmapCapabilities `json:"capabilities"`
+	Snat         bool                `json:"snat"`
+}
+type PortmapCapabilities struct {
+	Portmappings bool `json:"portMappings"`
+}
+
+// ConsulCNI is the "consul-cni" plugin used for transparent proxy.
+// https://github.com/hashicorp/consul-k8s/blob/main/control-plane/cni/main.go
+type ConsulCNI struct {
+	Type     string `json:"type"`
+	LogLevel string `json:"log_level"`
+}
--- a/client/allocrunner/networking_bridge_linux.go
+++ b/client/allocrunner/networking_bridge_linux.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/coreos/go-iptables/iptables"
 	hclog "github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/nomad/client/allocrunner/cni"
 	"github.com/hashicorp/nomad/nomad/structs"
 	"github.com/hashicorp/nomad/plugins/drivers"
 )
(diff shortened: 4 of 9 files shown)


client/allocrunner/networking_bridge_linux.go
```diff
@@ -156,69 +165,61 @@ func (b *bridgeNetworkConfigurator) Teardown(ctx context.Context, alloc *structs
 	return b.cni.Teardown(ctx, alloc, spec)
 }
 
-func buildNomadBridgeNetConfig(b bridgeNetworkConfigurator, withConsulCNI bool) []byte {
-	var consulCNI string
-	if withConsulCNI {
-		consulCNI = consulCNIBlock
+func buildNomadBridgeNetConfig(b bridgeNetworkConfigurator, withConsulCNI bool) ([]byte, error) {
+	// Update website/content/docs/networking/cni.mdx when the bridge configuration
```
martisah, line 169: This function is quite lengthy since it defines all the plugins and appends them to the overall network config, I was debating adding constructors for each plugin in the config file I created, but didn't want to overcomplicate things, but I am completely open to splitting this up!

client/allocrunner/networking_bridge_linux.go
```diff
@@ -156,69 +165,61 @@ func (b *bridgeNetworkConfigurator) Teardown(ctx context.Context, alloc *structs
...
+	allocRange := Range{Subnet: b.allocSubnet}
+	ranges = append(ranges, allocRange)
+
+	var routes []Route
+	allocRoute := Route{Dst: "0.0.0.0/0"}
+	routes = append(routes, allocRoute)
+
+	loop := Loopback{Type: "loopback"}
+
+	NewFirewall := Firewall{
```
gulducat, line 182: nit:

```suggestion
	newFirewall := Firewall{
```

and similar for the others.

although, I might skip naming them, and instead define them inline, like:

```go
plugins := []any{ // I wonder if we could swing a more specific type?
	Loopback{Type: "loopback"},
	Portmap{
		Type:         "portmap",
		Capabilities: CapabilityArgs{Portmappings: true},
		Snat:         true,
	},
	/// etc ...
}
```

but I'm curious about types again, because in the end, what we're producing is a conflist, which gets unmarshaled on the other end by the cni lib:

https://github.com/containernetworking/cni/blob/main/pkg/types/types.go#L118-L125

it may be sketchy to "just use the lib structs" but I figured I'd mention it.

client/allocrunner/networking_bridge_linux.go
```diff
@@ -156,69 +165,61 @@ func (b *bridgeNetworkConfigurator) Teardown(ctx context.Context, alloc *structs
...
+		Type:                   "firewall",
+		Backend:                "iptables",
+		IptablesAdminChainName: cniAdminChainName,
```
gulducat, line 185: to me this setup is already paying off -- we can just use the var name in the place that it's needed, without counting args in fmt.Sprintf

client/allocrunner/networking_bridge_linux.go
```diff
@@ -156,69 +165,61 @@ func (b *bridgeNetworkConfigurator) Teardown(ctx context.Context, alloc *structs
...
+		IpMasq:       true,
+		IsGateway:    true,
+		ForceAddress: true,
+		HairpinMode:  b.hairpinMode,
+		Ipam: IPAM{
+			Type:   "host-local",
+			Routes: routes,
+		},
 	}
+	NewBridge.Ipam.Ranges = append(NewBridge.Ipam.Ranges, ranges)
```
gulducat, line 205: is this split out like this because over in ipv6-land we'll have more than this one range?

client/allocrunner/networking_bridge_linux.go
```diff
@@ -156,69 +165,61 @@ func (b *bridgeNetworkConfigurator) Teardown(ctx context.Context, alloc *structs
...
-	))
-}
+	pluginsList := []any{loop, NewBridge, NewFirewall, NewPortmap}
+	CNIconfig := &BridgeCNIPlugin{
+		CniVersion: "0.4.0",
+		Name:       "nomad",
+		Plugins:    pluginsList,
+	}
+	if withConsulCNI {
+		CNIconfig.Plugins = append(CNIconfig.Plugins, ConsulCNIBlock{
```
gulducat, line 214: this feels so much better to me than Sprint-ing the json strong

client/allocrunner/networking_bridge_linux.go
```diff
@@ -156,69 +165,61 @@ func (b *bridgeNetworkConfigurator) Teardown(ctx context.Context, alloc *structs
...
+			Type:     "consul-cni",
+			Loglevel: "debug",
+		})
+	}
 
-// Update website/content/docs/networking/cni.mdx when the bridge configuration
-// is modified. If CNI plugins are added or versions need to be updated for new
-// fields, add a new constraint to nomad/job_endpoint_hooks.go
-const nomadCNIConfigTemplate = `{
```
gulducat, line 177: we will want to retain the ability to construct this json for pasting in docs.  I was thinking a unit test that prints out the different variations (and maybe also make some assertions, heh)?  that way the docs could be be kept up to date about as easily as they are now with this template situation.

gulducat (commented):
this is very promising!  some food for thought for ya.

client/allocrunner/networking_bridge_linux.go
gulducat: seems like a reasonable function length to me!

client/allocrunner/bridge_plugin_config.go
```diff
@@ -0,0 +1,54 @@
...
+	Dst string `json:"dst"`
+}
+type ConsulCNIBlock struct {
+	Type     string `json:"type"`
+	Loglevel string `json:"log-level"`
+}
+
+type Bridge struct {
+	Type         string `json:"type"`
+	Bridgename   string `json:"bridge"`
```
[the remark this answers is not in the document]
martisah: In all honesty, the cni library code ended up getting me confused so I just implemented the structs myself the way it made sense to me, but I can totally look at refactoring that if that's what we end up wanting to do!

client/allocrunner/networking_bridge_linux.go
martisah: Yep! That's the idea 👍

client/allocrunner/bridge_plugin_config.go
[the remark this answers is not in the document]
gulducat: I think using these new copies is still an improvement over formatting a flat json string, and I am thinking that keeping our own only-loosely-coupled copies allows us to keep the config tight and potentially valid for multiple versions of the plugin that users could be running.

maybe we can fiddle with the format just a little bit, but I think I like this approach!

client/allocrunner/networking_bridge_linux.go
gulducat: this confused me for a minute, because I have tried to append a slice to a slice like this before, and was mistaken in those cases, but here `Ranges` is a `[][]Range`!  so it's appending a slice to a slice of slices.

but! I think I would personally set the default ipv4 `Ranges` up in the main `&Bridge{` definition, then later we can append the ipv6 range(s) to it in a separate line like this -- akin to how you're doing below `withConsulCNI`

martisah: That makes more sense to me as well, we can do that!

client/allocrunner/networking_bridge_linux.go
gulducat: I like the rearranging, but the concrete types in the library don't easily accommodate the variety of k/v that different plugins use, so I think just duplicating all of em here (as you've done) is probably the way to go.

client/allocrunner/cni/bridge.go
```diff
@@ -0,0 +1,86 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package cni
```
tgross, line 4: It feels a little weird having a separate package called `cni` here without it including much of any of the CNI configuration. There's no code isolation happening because everything is exported (has to be because of serialization.) All CNI setup including all the configuration tests are still in the `client/allocrunner` package. It seems hard to disentangle the "bridge configurator" code from CNI as well.

client/allocrunner/cni/bridge_test.go
```diff
@@ -0,0 +1,31 @@
...
+	must.Eq(t, `{
+	"cniVersion": "0.0.1",
+	"name": "test-config",
+	"plugins": [
+		{
+			"type": "test-plugin"
+		}
+	]
+}`, string(bts))
+}
```
tgross, line 31: Doesn't this just test the stdlib?

client/allocrunner/cni/plugins.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// Portmap is the "portmap" plugin.
+// https://www.cni.dev/plugins/current/meta/portmap/
+type Portmap struct {
+	Type         string         `json:"type"`
+	Capabilities CapabilityArgs `json:"capabilities"`
+	Snat         bool           `json:"snat"`
+}
+type CapabilityArgs struct {
```
tgross, line 49: We have to export these types because of Go's rules around serialization, which means the type is visible outside this package. Given this struct belongs only to the `Portmap` plugin, might it be better named something like `PortmapCapabilities`?

client/allocrunner/test_fixtures/default.conflist.json
```diff
@@ -0,0 +1,44 @@
...
+			"forceAddress": true,
+			"hairpinMode": false,
+			"ipam": {
+				"type": "host-local",
+				"ranges": [
+					[
+						{
+							"subnet": ""
+						}
+					]
```
tgross, line 22: I think the test was called `empty` because it _wasn't_ correctly populated as it will be in the caller, which is how you get this bogus range configuration. It might not be a bad idea to have an actually-default test that populates the minimum set of items we'd expect to see populated.

tgross (approved):
LGTM to ship. My comments are for discussion and should be considered non-blocking, but the package change feels like we're trying to carve out a CNI package when maybe what we want is a "networking" package?

apollo13:
If the networking config gets touched this massively, would it also be possible to allow users to override it -- I'd really love to use a custom cni configuration for every job without manually adding it to the job

gulducat:
Hi @apollo13!  It would certainly be possible to do, and has been suggested before in #14480, but I agree with Luiz's closing comment there that we should be cautious with what the magic `"bridge"` network string represents.  If it could mean anything, then our ability to support our own "bridge" network is undercut drastically, and it's important that we be able to rely on it when designing other features.

I understand that it means a lot of `mode = "cni/my-cool-network"` in job files, but at least when that is explicit, we (engineers in github issues, and our enterprise support team) know that we're troubleshooting a custom network setup.

client/allocrunner/cni/bridge.go
gulducat: the initial motivation was to namespace the structs -- I found myself naming them like `FirewallCNIPlugin` just because they were in the `allocrunner` package, and I thought if I made a small new package, then they could be named more simply and still make sense.  I also just like small, focused packages in general.

but were I to spend more time on this, I probably would move more of the CNI stuff into here, but I figure this is an incremental improvement to be perhaps built on further, later.

regarding the "networking" package comment, a decent chunk of our networking config, especially on linux, is CNI, but I might still like to distinguish the things that are specifically satisfying CNI spec from other more-general networking concerns.  i.e. if I really went ham refactoring this area of allocrunner, I might make a "networking" package, then still put a (relatively smaller) "cni" package within there.

but as you might imagine, I am trying not to boil the ocean! :sweat_smile: 

all that said, I could move the new golden-file tests into here, since they are a direct output of this package.  I left them where they were because that was the hook into this refactor, but I did create them here in the first place...

client/allocrunner/test_fixtures/default.conflist.json
gulducat: gosh you're right, good catch!

client/allocrunner/cni/bridge.go
gulducat: I'll also admit to being a little concerned about backports, since I've been intending to go back to 1.6.x for future maintainability, but e.g. consul-cni is 1.8.x only.  backports won't be tooooo painful anyway, or I could forego them entirely, but at any rate, it was on my mind when deciding what all to move into this new package.

tgross: > but as you might imagine, I am trying not to boil the ocean!

Totally reasonable :+1:

client/allocrunner/test_fixtures/default.conflist.json
gulducat: I had half a mind to add validation to protect against this, but just stuck with what was here instead to keep it a true no-op refactor.  think I should reach a little further and check for invalid empty values like this?

tgross:
> I'd really love to use a custom cni configuration for every job without manually adding it to the job

Can you dig into what you might be looking for a bit more @apollo13? [`network.mode = "host"`](https://developer.hashicorp.com/nomad/docs/job-specification/network#mode) is the default network, so setting `network.mode = "bridge"` doesn't seem meaningfully different than setting `network.mode = "cni/example"`. The clients need to have the CNI config file of course, but you'd have to configure a "default bridge" similarly too, I'd think?

client/allocrunner/test_fixtures/default.conflist.json
tgross: I think it's fine to keep new invalidation logic out of this refactor PR.

apollo13:
Mhm, you are right @tgross. I forgot about `host` -- I always set `bridge` so in my mind this was the default. I guess part of the motiviation is wanting to run connect-enabled jobs with my custom bridge. Though that should be possible if we were to remove the current limitations and checks around `mode="bridge"` and also allow `cni/` there. Not sure though how that would work with the modifications needed for the transparent proxy though.

On a sidenote, namespaces have `enabled_task_drivers/disabled_task_drivers`. Would you accept a PR adding `enabled_network_modes/disable_network_modes`?

Issue #23770, opened 2024-08-08
New metric to track tasks in pending state


the-nando:
### Proposal

It would be great to have an exported metric to track pending tasks. There are currently metrics for pending jobs and alloc but not tasks. Ideally a job-scoped metric but client-scoped would work as well.

### Use-cases

With the introduction of Workload Identity and Multi-Vault support I have a few jobs with different tasks which source secrets from different Vault clusters. I had a few cases in which some tasks failed to render secrets from Vault, entered the "pending" state and I would like to be alerted whenever that happens.

### Attempted Solutions

Log-based alerting.

tgross:
Hi @the-nando! This looks feasible. The relevant code in the client where we emit the allocation pending/running/etc metrics already has the data handy, and it wouldn't add much to the cost of storing metrics to add a metric for each of the 3 task states ("pending", "running", "dead"). I had a few minutes between other tasks and so I've implemented that in https://github.com/hashicorp/nomad/pull/23773

the-nando:
oh wow, thanks a LOT @tgross for jumping on it this quick! Hopefully it makes it into 1.8.3 😄

PR #23773, opened 2024-08-08
metrics: add `client.tasks` state metrics

tgross:
Although we have `client.allocations` metrics to track allocation states on a client, having separate metrics for `client.tasks` will allow operators to identify that there are individual tasks in an unexpected state in an otherwise healthy allocation.

Fixes: https://github.com/hashicorp/nomad/issues/23770


pkazmierczak (approved):
LGTM!

PR #23785, opened 2024-08-09
api: only set url field in config if previously unset

tgross:
In #16872 we added support for configuring the API client with a unix domain socket. In order to set the host correctly, we parse the address before mutating the Address field in the configuration. But this prevents the configuration from being reused across multiple clients, as the next time we parse the address it will no longer be pointing to the socket. This breaks consumers like the autoscaler, which reuse the API config between plugins.

Update the `NewClient` constructor to only override the `url` field if it hasn't already been parsed. Include a test demonstrating safe reuse with a unix domain socket.

Ref: https://github.com/hashicorp/nomad-autoscaler/issues/944
Ref: https://github.com/hashicorp/nomad-autoscaler/pull/945

---

Note for reviewers: in some hypothetical "v2" of the API I'd like to have some kind of options or builder pattern for the client config where we can "consume" the config. But for the moment getting the `Config` to be safely reusable seems like the best approach to avoid further backwards-incompatibility.

### DIFF
--- a/.changelog/23785.txt
+++ b/.changelog/23785.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+api: Fixed a bug where an `api.Config` targeting a unix domain socket could not be reused between clients
+```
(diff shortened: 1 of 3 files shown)


gulducat (approved):
lgtm!

it seems a little counterintuitive that we mutate the config that's passed in at all, but at least this does so a little more predictably!

PR #23792, opened 2024-08-12
docs: remove remaining references to network_speed config

mvegter:
Follow-up to #9793


aimeeu (approved):
Thanks for fixing this!

PR #23800, opened 2024-08-13
docs: minimal Consul policy for Nomad agents needs node:write

pkazmierczak:
Since Consul 1.15 (https://github.com/hashicorp/consul/pull/16097) Consul client agent only uses the agent token to perform [anti-entropy](https://developer.hashicorp.com/consul/docs/architecture/anti-entropy) after you deregister a service via the Agent API. so the Consul client agent must have its own agent token.

Internal ref: https://hashicorp.atlassian.net/browse/NET-10360


tgross (approved):
LGTM!

Issue #18211, opened 2023-08-15
Make use of systemd delegeate cgroup when possible.


shoenig:
Per https://systemd.io/CGROUP_DELEGATION/ (at the bottom)

> 🚫 Never create your own cgroups below arbitrary cgroups systemd manages, i.e cgroups you haven’t set Delegate= in. Specifically: 🔥 don’t create your own cgroups below the root cgroup 🔥. 

Currently Nomad does exactly this - it creates the `nomad.slice` cgroup under the root cgroup regardless if systemd is in use or not. We should modify our linux packaging to set the Delegate in the systemd unit file so that we are in line with the expected usage of systemd.

However we'll need to continue supporting the mode of operation we have today - not all Linux operating systems use systemd (and thus have no delegate mechanism), and not all users use our Linux packaging. We'll also want to update our production documentation to make recommendations for such users.

tgross:
Another challenge with setting delegation is that we have a `nomad.slice` and slices aren't supported for delegation; we'd need to create a scope below `nomad.slice` and delegate that.

tgross:
This was the bit from that document:
> Let’s stress one thing: delegation is available on scope and service units only. It’s expressly not available on slice units. Why? Because slice units are our inner nodes of the cgroup trees and we freely attach services and scopes to them. If we’d allow delegation on slice units then this would mean that both systemd and your own manager would create/delete cgroups below the slice unit and that conflicts with the single-writer rule.

In an experiment I'm hacking on, turns out this doesn't really matter because we're not creating a "slice unit", we're just creating our own cgroup directory that happens to be called "slice". If we did want to create a [slice unit](https://www.man7.org/linux/man-pages/man5/systemd.slice.5.html) in the package (which we will if we want to allow less-privileged Nomad agents), we'd instead want to have 2 `Delegate=` fields pointing to the `shared.slice` and `reserved.slice` not-really-slices below that. I've smoked-tested this so far, but it needs further investigation.

PR #23803, opened 2024-08-13
cgroupslib: allow initial controller check with delegated cgroups

tgross:
During Nomad client initialization with cgroups v2, we assert that the required cgroup controllers are available in the root `cgroup.subtree_control` file by idempotently writing to the file. But if Nomad is running with delegated cgroups, this will fail file permissions checks even if the subtree control file already has the controllers we need.

Update the initialization to first check if the controllers are missing before attempting to write to them. This allows cgroup delegation so long as the cluster administrator has pre-created a Nomad owned cgroups tree and set the `Delegate` option in a systemd override. If not, initialization fails in the existing way.

Although this is one small step along the way to supporting a rootless Nomad client, running Nomad as non-root is still unsupported. I've intentionally not documented setting up cgroup delegation in this PR, as this PR is insufficient by itself to have a secure and properly-working rootless Nomad client.

Ref: https://github.com/hashicorp/nomad/issues/18211
Ref: https://github.com/hashicorp/nomad/issues/13669
Ref: https://hashicorp.atlassian.net/browse/NET-10652
Ref: https://hashicorp.atlassian.net/browse/NET-10668

### DIFF
--- a/client/lib/cgroupslib/mount.go
+++ b/client/lib/cgroupslib/mount.go
@@ -11,15 +11,17 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 
 	"github.com/hashicorp/go-set/v2"
 )
 
+// detect tries to detect which cgroups version we have by looking at the mount
+// and whether Nomad owns the cgroup.
+// - For cgroups v1 this requires root.
+// - For cgroups v2 we look for root or whether we're the owner of the slice.
+// - All other cases, including any file permission errors, return OFF.
 func detect() Mode {
-	if os.Geteuid() > 0 {
-		return OFF
-	}
-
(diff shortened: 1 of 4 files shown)


client/lib/cgroupslib/mount.go
```diff
@@ -11,15 +11,12 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 
 	"github.com/hashicorp/go-set/v2"
 )
 
 func detect() Mode {
```
shoenig, line 19: It's probably time to leave one of those huge documentation block comments describing when things are detected OFF or not.

shoenig (approved):
LGTM!

PR #23804, opened 2024-08-13
docker: disable cpuset management for non-root clients

tgross:
Nomad clients manage a cpuset cgroup for each task to reserve or share CPU cores. But Docker owns its own cgroups, and attempting to set a parent cgroup that Nomad manages runs into conflicts with how runc manages cgroups via systemd. Therefore Nomad must run as root in order for cpuset management to ever be compatible with Docker.

However, some users running in unsupported configurations felt that the changes we made in Nomad 1.7.0 to ensure Nomad was running correctly represented a regression. This changeset disables cpuset management for non-root Nomad clients. When running Nomad as non-root, the driver will not longer reconcile cpusets with Nomad and `resources.cores` will behave incorrectly (but the driver will still run).

Although this is one small step along the way to supporting a rootless Nomad client, running Nomad as non-root is still unsupported. This PR is insufficient by itself to have a secure and properly-working rootless Nomad client.

Ref: https://github.com/hashicorp/nomad/issues/18211
Ref: https://github.com/hashicorp/nomad/issues/13669
Ref: https://hashicorp.atlassian.net/browse/NET-10652
Ref: https://github.com/opencontainers/runc/blob/main/docs/systemd.md

### DIFF
--- a/.changelog/23804.txt
+++ b/.changelog/23804.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+docker: Disable cpuset management for non-root clients
+```
--- a/drivers/docker/config.go
+++ b/drivers/docker/config.go
@@ -683,6 +683,10 @@ type DriverConfig struct {
 
 	AllowRuntimesList []string            `codec:"allow_runtimes"`
 	allowRuntimes     map[string]struct{} `codec:"-"`
+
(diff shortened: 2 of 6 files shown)


shoenig (approved):
LGTM

PR #23806, opened 2024-08-14
build: apt update before apt install

shoenig:
Run `apt update` before `apt install` for `qemu-system` install step. Otherwise the cache may be out of date with Azure's own mirrors.


tgross (approved):
:+1:

PR #23807, opened 2024-08-14
build: update golangci-lint to 1.60.1

shoenig:
Helps with go1.23 compatibility. Introduces some breaking changes / newly
enforced linting patterns so those are fixed as well.


nomad/csi_endpoint.go
```diff
@@ -357,7 +357,7 @@ func (v *CSIVolume) Register(args *structs.CSIVolumeRegisterRequest, reply *stru
 
 			*vol = *existingVol
 
-		} else if vol.Topologies == nil || len(vol.Topologies) == 0 {
+		} else if len(vol.Topologies) == 0 {
```
pkazmierczak, line 360: huh, funny that this is equivalent to the nil check || length check. is this a new Go thing or has it always been like this?

pkazmierczak (approved):
LGTM!

nomad/csi_endpoint.go
shoenig: an empty slice has always been `==` to a nil slice in Go; now the linter just checks for it

Issue #23798, opened 2024-08-13
extra_claims job ID without periodic suffix for batch jobs


EtienneBruines:
### Proposal
Add a `vault.default_identity.extra_claims` interpolation for the job ID that does not include the `/periodic-1234567` suffix, the same way the `nomad_job_id` does not include it. 

Currently `${job.id}` looks something like this `my-batch-job-name/periodic-1723556632` and I'd like a variable (perhaps `${job.name}`?) that gives me just `my-batch-job-name`

### Use-cases
- Vault integration for batch jobs. Not having the periodic suffix makes it more doable to actually provide secrets/paths in the Vault policy that it can access. 

### Attempted Solutions
- No workaround is available.

Currently available interpolations:

https://github.com/hashicorp/nomad/blob/90d646a2ae696058f04bd634f596052897ec2719/nomad/structs/workload_id.go#L235-L252

Sorry @tgross for not realizing this when the PR was still a PR.

tgross:
:facepalm: That's embarrassing. Will fix.

the-nando:
hey @tgross I've run into this as well unfortunately, in my quest to migrate some clusters to workload identity. Do you know more or less when this bug fix will ship?

tgross:
I've got a PR up with the fix https://github.com/hashicorp/nomad/pull/23817 but I've been on PTO the last couple weeks. Planning on shipping this in the upcoming Nomad 1.4.0.

PR #23817, opened 2024-08-14
WI: interpolate parent job ID in `vault.default_identity.extra_claims`

tgross:
When we interpolate job fields for the `vault.default_identity.extra_claims` block, we forgot to use the parent job ID when that's available (as we do for all other claims). This changeset fixes the bug and adds a helper method that'll hopefully remind us to do this going forward.

Also added a missing changelog entry for #23675 where we implemented the `extra_claims` block originally, which shipped in Nomad 1.8.3.

Fixes: https://github.com/hashicorp/nomad/issues/23798
Ref: https://hashicorp.atlassian.net/browse/NET-10714

### DIFF
--- a/nomad/structs/workload_id_test.go
+++ b/nomad/structs/workload_id_test.go
@@ -20,6 +20,7 @@ func TestNewIdentityClaims(t *testing.T) {
 
 	job := &Job{
 		ID:        "job",
+		ParentID:  "parentJob",
 		Name:      "job",
 		Namespace: "default",
 		Region:    "global",
@@ -178,7 +179,7 @@ func TestNewIdentityClaims(t *testing.T) {
 		// group: no consul.
 		"job/group/services/group-service": {
 			Namespace:   "default",
-			JobID:       "job",
+			JobID:       "parentJob",
 			ServiceName: "group-service",
 			Claims: jwt.Claims{
 				Subject:  "global:default:job:group:group-service:consul-service_group-service-http",
@@ -190,7 +191,7 @@ func TestNewIdentityClaims(t *testing.T) {
 		// task:  no consul, no vault.
 		"job/group/task/default-identity": {
 			Namespace: "default",
-			JobID:     "job",
+			JobID:     "parentJob",
 			TaskName:  "task",
 			Claims: jwt.Claims{
 				Subject:  "global:default:job:group:task:default-identity",
@@ -200,7 +201,7 @@ func TestNewIdentityClaims(t *testing.T) {
 		},
 		"job/group/task/alt-identity": {
 			Namespace: "default",
-			JobID:     "job",
+			JobID:     "parentJob",
 			TaskName:  "task",
 			Claims: jwt.Claims{
 				Subject:  "global:default:job:group:task:alt-identity",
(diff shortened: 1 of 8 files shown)


nomad/structs/workload_id_test.go
```diff
@@ -20,6 +20,7 @@ func TestNewIdentityClaims(t *testing.T) {
 
 	job := &Job{
 		ID:        "job",
+		ParentID:  "parentJob",
```
gulducat, line 23: the new helper method is simple enough, but is there also test coverage for parentless (service/system) jobs?

gulducat (approved):
lgtm!

nomad/structs/workload_id_test.go
tgross: Not of this function specifically, but most of the downstream code doesn't test with batch workloads (which is how this slipped thru in the first place).

nomad/structs/workload_id_test.go
```diff
@@ -200,7 +201,7 @@ func TestNewIdentityClaims(t *testing.T) {
 		},
 		"job/group/task/alt-identity": {
 			Namespace: "default",
-			JobID:     "job",
+			JobID:     "parentJob",
 			TaskName:  "task",
 			Claims: jwt.Claims{
 				Subject:  "global:default:job:group:task:alt-identity",
```
tgross, line 207: @schmichael I'm just now noticed we don't use the parent ID for the `Subject` claim... is that intention or did we repeat the same bug there?

EtienneBruines (commented):
Even though I cannot come up with an actual use-case, there might be instances where someone wants to use the `/periodic-123456` suffix as part of their claim. Or something else somewhat unique to the individual run (e.g. allocation ID). 

Use-cases will probably be related to the fact that the field you select in Vault for the `user_claim` needs to be unique for the combination of claims. For example: 

Before Nomad v1.8.3 you had to configure `/sub` as `user_claim` if you wanted different dynamic database credentials (Vault `database` secret engine) per-task within the same Nomad job. For example one migrations-tasks with full admin rights to the databse and one webserver-task with limited permissions. If you configured something else as the `user_claim`, e.g. `nomad_job_id`, then the tasks would 'overwrite' each other, pointing to the wrong JWT claims. I could imagine something similar happening to someone when it comes to unique invocations of periodic/batch jobs, providing a need for something unique-ish - be that `alloc.id` or the `/periodic-123456` suffix. 

Whether that's just an additional interpolate variable or whether it's part of the `Subject` claim isn't up to me, but providing something somewhere that is unique to the allocation/run might be beneficial.

nomad/structs/workload_id_test.go
schmichael: We repeated the same bug. I'm sure I just forgot about the distinction between Parent/Child IDs when writing it initially. :(

tgross: I'll fix that in a separate PR.

tgross:
> providing a need for something unique-ish - be that alloc.id

`nomad_alloc_id` is already one of the claims. In fact, that's how Nomad adds extra validation to requests made to the Nomad API -- workload identities for terminal allocations are invalid. But we should make that available for interpolation here as well.

nomad/structs/workload_id_test.go
tgross: https://github.com/hashicorp/nomad/pull/23902

Issue #23811, opened 2024-08-14
nomad client agent incorrectly indicates that it has zero cpu


josh-m-sharpe:
### Nomad version
```
nomad --version
Nomad v1.8.3
BuildDate 2024-08-13T07:37:30Z
Revision 63b636e5cbaca312cf6ea63e040f445f05f00478
```

### Operating system and Environment details
amazonlinux 2023

### Issue
client detects 0 cpu.  therefore can't alloc jobs.

### Reproduction steps
run this agent as root:

```
datacenter = "dc1"
data_dir   = "/mnt/nomad"

log_level = "TRACE"

client {
  enabled = true

  "node_class" = "es"

  host_volume "docker-sock-ro" {
    path = "/var/run/docker.sock"
    read_only = true
  }

  # cpu_total_compute = 10000
}

plugin "docker" {
  config {
    auth {
      helper = "ecr-login"
    }

    # extra Docker labels to be set by Nomad on each Docker container with the appropriate value
    extra_labels = ["job_name", "task_group_name", "task_name", "namespace", "node_name"]

    allow_privileged = true
  }
}

telemetry {
  collection_interval = "1s"
  disable_hostname = true
  prometheus_metrics = true
  publish_allocation_metrics = true
  publish_node_metrics = true
}
```

#### Expected Result
jobs would deploy, client would report that it has cpu

#### Actual Result
visit client UI:
/ui/clients/7ecfa3d3-91ff-71a7-b7a1-8a51c87a495d 
See: `0 MHz / 0 MHz Total`

### Job file (if appropriate)
not really important other than specifying some cpu resource requirement, e.g.: `cpu = 100`

### Notes
It would appear that uncommenting `cpu_total_compute = 10000` in that client config above "fixes" this.  **_This is not a real fix._**  The agent reports that it has 10000 cpu: `4% 388 MHz / 10,000 MHz Total` (it doesn't really) but this at allows me to allocate jobs to these clients.

### Other context
I upgraded from v1.5.8 to v1.8.3.  I didn't test any versions in between.

pkazmierczak:
Ah, I see you're running amazonlinux, my bad. Still:
```
[ec2-user@ip-10-0-1-14 ~]$ cat /etc/amazon-linux-release
Amazon Linux release 2023.5.20240805 (Amazon Linux)
[ec2-user@ip-10-0-1-14 ~]$ nomad node status -self -verbose|grep cpu
cpu.arch                                 = amd64
[...]

PR #23827, opened 2024-08-15
client: log an error message if total detected cpu is zero

pkazmierczak:
It would be useful to surface CPU fingerprinting issue with an `ERROR` level log message, vide suggestion in https://github.com/hashicorp/nomad/issues/23811


tgross (approved):
LGTM!

PR #23829, opened 2024-08-15
[ui] Show "Scaled Down" as a valid job status when task groups' counts are set to zero

### LINKED ISSUE #23591: Nomad UI shows failed for jobs that are scaled to 0
Hello, 

We have this behavior on nomad when we run `nomad job scale <job> <group> 0` that the job shows as failed.

It would be nice to have a better message for this state instead of Failed, I know the job is not healthy since the service checks cant pass cuz its scaled to zero, but maybe a flag "Downscaled" would be a good thing to have.

Is this on track? are there any plans for it? 

Example situation: 
CleanShot 2024-07-14 at 12 02 03

Expected:
CleanShot 2024-07-14 at 12 10 43
or 
CleanShot 2024-07-14 at 12 13 30

philrenaud:
Resolves #23591

### DIFF
--- a/.changelog/23829.txt
+++ b/.changelog/23829.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: Badge added for Scaled Down jobs
+```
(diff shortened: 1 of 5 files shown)


pkazmierczak (approved):
LGTM!

PR #23832, opened 2024-08-15
[ui] Prevent double-open for cmd+click on jobs index links

### LINKED ISSUE #23544: UI - Cmd/Ctrl + Click won't correctly open new tabs
### Nomad version
edc56664-71d2-7a63-9574-98f1608a9fdf

Nomad v1.8.1
BuildDate 2024-06-19T06:43:57Z
Revision https://github.com/hashicorp/nomad/commit/5022543e4b7b8dcec9df123f86630ae3fdcffbe6

### Issue
In the new version when using the UI, cmd + mouseclick wont correctly open new tabs on the browser. 
It both opens the new tab but also changes your current one. This only happens when interacting with the jobs list, which is basically the updated page. Other tabs/menus don't have this problem

### Reproduction steps
Open the Jobs list on the UI and try to cmd+click any job. You would see both a new tab but also have your current one change

#### Expected Result
Open a new tab with the job selected

#### Actual Result

https://github.com/hashicorp/nomad/assets/31411471/0faf23b4-0734-4407-b79d-aa967c364a7d

philrenaud:
Previously, cmd+click (or ctrl+click) on the `<a>` tag within job rows on the jobs index page would double-open the page: once in a new tab (as cmd+click does) and another in the same tab (as it would trigger the `{{on "click"}}` Ember event for the job row.

This checks to make sure the clicked element wasn't the anchor tag itself, and if it is, it returns early.

NB: it's a bad practice, generally, to have non-interactive elements like table rows be clickable! However this follows a long-standing convention in the Nomad UI that users may be used to from years of use, and has more than one highly accessible fallback.

Resolves #23544

### DIFF
--- a/.changelog/23832.txt
+++ b/.changelog/23832.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fix an issue where cmd+click or ctrl+click would double-open a job
+```
--- a/ui/app/controllers/jobs/index.js
+++ b/ui/app/controllers/jobs/index.js
@@ -12,6 +12,8 @@ import { tracked } from '@glimmer/tracking';
 import localStorageProperty from 'nomad-ui/utils/properties/local-storage';
 import { restartableTask, timeout } from 'ember-concurrency';
 import Ember from 'ember';
+// eslint-disable-next-line no-unused-vars
+import JobModel from '../../models/job';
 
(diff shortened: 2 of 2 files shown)


gulducat (approved):
:mouse_trap:

PR #23838, opened 2024-08-16
auth: enforce use of node secret and remove legacy auth

tgross:
As of Nomad 1.6.0, Nomad client agents send their secret with all the RPCs (other than registration). But for backwards compatibility we had to keep a legacy auth method that didn't require the node secret. We've previously announced that this legacy auth method would be removed and that nodes older than 1.6.0 would not be supported with Nomad 1.9.0.

This changeset removes the legacy auth method.

Ref: https://developer.hashicorp.com/nomad/docs/release-notes/nomad/upcoming#nomad-1-9-0


tgross:
Had to resolve a merge conflict on the upgrade guide. Will merge once CI is green again.

PR #23849, opened 2024-08-21
Docs: CE-674 Add job statuses

aimeeu:
Add new page called Job and document job statuses. General content pulled from existing tutorials and docs.  Status content comes from Phil's [comment](https://hashicorp.atlassian.net/browse/CE-674?focusedCommentId=500397) in the Jira ticket.

Jira: https://hashicorp.atlassian.net/browse/CE-674

[Deploy preview](https://nomad-git-ce-674-hashicorp.vercel.app/nomad/docs/concepts/job/)

Note that I originally called the page /jobs.mdx and changed to job.mdx on commit 3.


website/content/docs/concepts/jobs.mdx
```diff
@@ -0,0 +1,79 @@
...
+spec. Nomad makes sure the actual state matches your desired state.
+
+## Job types
+
+Nomad supports the following job types:
+
+- Batch
+
+  Batch jobs are for short lived jobs that run until they exit
+  successfully. Batch jobs are paramaterized or periodic.
```
philrenaud, line 35: I don't know if it's true that batch jobs are parameterized or periodic, but it is true that there are Periodic and Parameterized job types that can have batch type child jobs.

(Batch type jobs can also be their own non-parameterized, non-periodic job type. Batch type is meant for running "scripts", or any other short-lived task)

I would consider adding Periodic and Parameterized jobs as types in addition to the ones here.

I would also consider grouping or tagging job types such that they are related across a couple important axes. For example: System and Sysbatch jobs are one-per-node, where others can have their placement configured; Batch and Sysbatch jobs are expected to be short-lived where Service and System jobs are expected to be relied about (like discoverable services)

philrenaud: Also noting that there's some overlap here with https://developer.hashicorp.com/nomad/docs/schedulers

website/content/docs/concepts/jobs.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+- Degraded: A deployment is not taking place. Some allocations are failed, lost, or unplaced.
+- Deploying: A [deployment](/nomad/docs/glossary#deployment) is actively taking place.
+- Failed: All allocations are failed, lost, or unplaced.
+- Healthy: All expected allocations are running and healthy.
+- Recovering: Some allocations are pending.
+- Removed: The job appeared in the initial query but has since been garbage
+  collected. This happens when the user remains on the page for a long time.
+- Stopped: A user has has manually stopped the job. It has not yet been purged or garbage collected.
+
```
philrenaud, line 73: We've also added `Scaled Down` as a status as of https://github.com/hashicorp/nomad/pull/23829 / https://github.com/hashicorp/nomad/issues/23591 (next 1.8.x release)

website/content/docs/concepts/jobs.mdx
```diff
@@ -0,0 +1,79 @@
...
+layout: docs
+page_title: Jobs
+description: |-
+  Learn about Nomad's Job feature, which is how you deploy your apps.
+---
+
+# Jobs
+
+This page explains the concept of a job, which is the Nomad construct for
+deploying your app and its configuration.
```
philrenaud, line 11: I don't know if others feel the same, but I've tended to shy away from the term "app" or application for what Nomad jobs do. We've used terms like workload or task in the past, but they also have their drawbacks. The implication of running an App is certainly suited to what service and system jobs _could_ be, but they also could be things like maintenance scripts, etc. In other words: I'm apprehensive about this word because it seems to sell things a little short!

philrenaud (approved):
This looks good!

I might consider adding more information about Job Statuses as the API/CLI define them: as `running`, `complete`, or `dead` (see https://developer.hashicorp.com/nomad/docs/commands/job/status#examples)

The UI statuses are an attempt to make things a little more true-to-what-users-would-expect, where the API/CLI-facing statuses are more rigid and have to do with how one might approach scheduling and garbage collection (the Nomad garbage collector doesn't care if a job is Failed or Stopped; it simply sees it as terminal/dead)

website/content/docs/concepts/jobs.mdx
aimeeu: Ok, I understand your point and will clarify on this page. Application/app is currently used throughout the docs and tutorial, so that's a cleanup for another day.

website/content/docs/concepts/jobs.mdx
aimeeu: I decided to add this status. I looked at the linked PR/issue to create the definition and its placement in the status mapping table.

website/content/docs/concepts/jobs.mdx
aimeeu: I decided to remove this section. It's more involved that I first thought, with conflicting information across older docs and tutorials. I think this topic is better addressed when we port content from tutorials to docs.

website/content/docs/concepts/job.mdx
```diff
@@ -0,0 +1,119 @@
...
+| ---------------    | --------------- |
+| `complete `        | `Complete`    |
+| `dead`             | `Failed`, `Scaled Down`, `Stopped` |
+| `running`          | `Degraded`, `Deploying`, `Healthy`, `Recovering`, `Running`, `Scaled Down` |
+
+
+### Complete
+
+The `Complete` status indicated that all expected allocations are complete. This
+status applies to only batch and sysbatch jobs.
```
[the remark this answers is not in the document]
aimeeu: Rather than repeat "In Nomad's UI, ..." in every entry, what do you think about adding a L4 heading "UI Statuses" ?

Issue #23603, opened 2024-07-16
show embedded template diffs line-by-line


EugenKon:
### Proposal
![image](https://github.com/user-attachments/assets/731f2492-e4a2-4e41-8ce7-37e175a6d75c)

For here it is not clear what was changed.
### Use-cases
It would be nice to see which lines were changed at that script.

tgross:
Related: https://github.com/hashicorp/nomad/issues/18103

vipulbhasin23:
Hi, I'd like to pick this up, but noticed #23851 already implemented this (CLI + UI) back in Aug 2024 and sat unreviewed until the author closed it a year later. Before I sink time into a fresh attempt, is this still wanted? Happy to build on that approach with tests added.

tgross:
It wasn't entirely unreviewed, but we didn't get a UI work review (the person I assigned that review to isn't on the team anymore 🤦 ). I think we would be interested in accepting a PR if you had one. It'll definitely need to include the UI work and tests, and make sure you've reviewed the contributing guide.

vipulbhasin23:
Hi @tgross — thanks for confirming! I'd like to pick this up.

I'll build on the approach from #23851 (crediting @amirabbas8's original implementation) rather than starting from scratch, since you already confirmed the direction is right. My plan:

- Same core approach: CLI (command/job_plan.go) uses cmp.Diff for EmbeddedTmpl fields instead of the flat %q => %q format; UI adds a small JobDiffTemplate component using a line-diff library, wired into job-diff-fields-and-objects.hbs.
- Adding test coverage that was missing from the original PR — a Go test for formatFieldDiff with an EmbeddedTmpl field, and an Ember integration test extending the existing job-diff-test.js pattern.
- Will review the contributing guide before opening the PR.

Will open a PR referencing this issue and #23851 once it's ready for review. Thanks for the quick response!

PR #23851, opened 2024-08-22
show embedded template diffs line-by-line

amirabbas8:
Hi team,

I’ve developed a proof of concept for displaying template differences in the job plan output. Instead of modifying the API fields, I handled the changes directly in the client. Specifically, I added a conditional check in the final step where the field is edited. This check determines whether the field name is EmbeddedTmpl and, if so, utilizes a library to display the diff.

I’d appreciate your feedback on this approach. Do you agree with handling the diff in both the command-line interface and the web UI, or would you prefer modifying the API instead? Additionally, are there any other suggestions or improvements you would recommend?

Thank you for your time.

Screenshot CMD
Screenshot Web

#23603


amirabbas8:
Hey @tgross,
Just checking if there will be a review of this merge request. 🙏🏻
Thanks!

tgross (commented):
Hi @amirabbas8!

The approach here looks great. We could do this on the server-side, but then the API client would need to transform the multi-line output anyways, so I think you've made the right choice here.

I'm a little more fuzzy on the UI side, so I'm going to tag in my colleague @philrenaud for his thoughts on the code and the new dependency you've introduced.

PR #23856, opened 2024-08-22
networking: refactor some iptables for testability

gulducat:
When adding support for ipv6 to the Nomad bridge network, I found that iptables admin chain logic was untested, so this refactor is in pursuit of that.  I also creeped a little bit to move the iptables cleanup mock to live next to the chain mock, so this iptables stuff is more centralized in one spot.

The coming-soon bridge+ipv6 PR will do more actual testing -- this is a separate no-op PR to allow easier backporting for potential future bugfixes and such.

### DIFF
--- a/client/allocrunner/networking_bridge_linux.go
+++ b/client/allocrunner/networking_bridge_linux.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"fmt"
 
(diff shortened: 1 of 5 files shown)


shoenig (approved):
LGTM

PR #23858, opened 2024-08-22
node endpoints: do not create evals for sysbatch jobs

pkazmierczak:
`node-update` triggers should never trigger sysbatch allocations, these should only ever be create by `periodic-job` or `job-register`.

An example scenario is: an allocation spawned by a sysbatch periodic job is running on a node, the allocation gets stopped, GC runs, the node becomes ineligible and eligible again, all within the parent sysbatch job period window. If this happens, `node-update` will trigger the system scheduler and prematurely start an allocation. This is not a desired behavior, and in fact a bug. 

Ref: https://hashicorp.atlassian.net/browse/NET-9323

### DIFF
--- a/nomad/node_endpoint.go
+++ b/nomad/node_endpoint.go
@@ -1695,6 +1695,12 @@ func (n *Node) createNodeEvals(node *structs.Node, nodeIndex uint64) ([]string, 
 		}
 		jobIDs[alloc.JobNamespacedID()] = struct{}{}
 
+		// If it's a sysbatch job, skip it. Sysbatch job evals should only
+		// ever be created by structs.EvalTriggerPeriodicJob
+		if alloc.Job.Type == structs.JobTypeSysBatch {
+			continue
+		}
+
 		// Create a new eval
 		eval := &structs.Evaluation{
 			ID:              uuid.Generate(),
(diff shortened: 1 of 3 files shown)


nomad/node_endpoint.go
```diff
@@ -1695,6 +1695,12 @@ func (n *Node) createNodeEvals(node *structs.Node, nodeIndex uint64) ([]string,
 		}
 		jobIDs[alloc.JobNamespacedID()] = struct{}{}
 
+		// If it's a sysbatch job, skip it. Sysbatch job evals should only
+		// ever be created by structs.EvalTriggerPeriodicJob
```
schmichael, line 1699: Hm, I'm having a hard time tracking down the relationship between `sysbatch` and `EvalTriggerPeriodicJob`. sysbatch jobs aren't _necessarily_ periodic.

I think the following logic is appropriate: non-periodic sysbatch jobs are "one shot": whatever nodes are eligible when they're scheduled, are the nodes they run on. We don't guarantee that nodes that were `down` when the sysbatch job was submitted ever receive the job if they come back and re-register.

pkazmierczak: Right! So the comment needs some re-phrasing I think, but the logic still holds. If it's a system batch job, and it's periodic, it'll get triggered by `periodic-job` and should not be triggered by anything else. If it's not periodic, `job-register` is the appropriate trigger and again, I believe the system scheduler should not be called on node update events.

schmichael (approved):
LGTM. Thanks for the expansive comment!

PR #23863, opened 2024-08-25
Tag and Untag job versions

philrenaud:
## via `curl`:

Adds a tag to Version 3:
```
curl -X PUT localhost:4646/v1/job/difftester/versions/testabc/tag -H "Content-Type: application/json" -d '{"Version": "3", "Tag": {"Name": "Foo", "Description": "Bar"}}'
```

Removes the tag:
```
curl -X DELETE localhost:4646/v1/job/difftester/versions/testabc/tag -H "Content-Type: application/json"
```

## via Nomad CLI:

Adds a tag to Version 0:
```
nomad job tag apply \
  -version=0 \
  -name="Foo" \
  -description="Bar baz" \
fails_every_10
```

Adds a tag to the latest/current version:
```
nomad job tag apply -name=foo -description=bar difftester
```

Removes a tag:
```
nomad job tag unset -name=foo fails_every_10
```


gulducat (commented):
looks pretty good!  few comments for ye

command/agent/job_endpoint.go
```diff
@@ -401,6 +404,64 @@ func (s *HTTPServer) jobRunAction(resp http.ResponseWriter, req *http.Request, j
...
+	if err != nil {
+		return nil, err
+	}
+
+	// parseWriteRequest overrides Namespace, Region and AuthToken
+	// based on values from the original http request
+	s.parseWriteRequest(req, &rpcArgs.WriteRequest)
+
+	var out structs.JobTagResponse
+	if err := s.agent.RPC("Job.TagVersion", &rpcArgs, &out); err != nil {
```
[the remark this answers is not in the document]
philrenaud: I ended up splitting these up a bit more — the differences started compounding (`unset` doesn't get `version` passed in now that we go by tag name; they both get their own structs for req/resp, etc.)

Not opposed to having a single method to handle these but they look to me like they should be separated currently

nomad/structs/structs.go
```diff
@@ -4527,6 +4529,31 @@ type JobTaggedVersion struct {
...
+	// QueryOptions
+	WriteRequest
+}
+
+type JobUnsetTagRequest struct {
+	JobID string
+	Name  string
+	// QueryOptions
+	WriteRequest // TODO: do deletes get a WriteRequest?
+}
```
tgross, line 4548: Yes, deletes get a `WriteRequest`. Also, do we really need a separate RPC message for this? It seems like we could untag a version using the same RPC as above with a little cleverness with the fields.

nomad/state/state_store.go
```diff
@@ -4885,6 +4885,99 @@ func (s *StateStore) updateJobStabilityImpl(index uint64, namespace, jobID strin
...
+	return txn.Commit()
+}
+
+func (s *StateStore) updateJobVersionTagImpl(index uint64, namespace, jobID string, jobVersion uint64, name string, description string, txn *txn) error {
+	ws := memdb.NewWatchSet()
+
+	tag := &structs.JobTaggedVersion{
+		Name:        name,
+		Description: description,
+		TaggedTime:  time.Now().UnixNano(),
```
tgross, line 4905: You can't call `time.Now()` when writing in the state store, because the FSM Apply has to be deterministic across the servers. If you want the time, it has to be in the request args.

command/agent/job_endpoint.go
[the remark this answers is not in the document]
philrenaud: (This did in fact end up being a fence I straddled and continue to straddle — see "request for reviewers" code comment)

nomad/structs/structs.go
philrenaud: Removed the second RPC message!

command/agent/job_endpoint.go
```diff
@@ -118,6 +118,11 @@ func (s *HTTPServer) JobSpecificRequest(resp http.ResponseWriter, req *http.Requ
 	case strings.HasSuffix(path, "/action"):
 		jobID := strings.TrimSuffix(path, "/action")
 		return s.jobRunAction(resp, req, jobID)
+	case strings.HasSuffix(path, "/tag"):
+		parts := strings.Split(path, "/")
+		jobID := parts[0]
+		name := parts[2] // job/<jobID>/tag/<name>
```
tgross, line 124: You need a length check here, otherwise `job/:jobid/tag` will panic rather than 404.

nomad/job_endpoint.go
```diff
@@ -2391,3 +2398,19 @@ func (j *Job) GetServiceRegistrations(
 		},
 	})
 }
+
+func (j *Job) TagVersion(args *structs.JobApplyTagRequest, reply *structs.JobTagResponse) error {
+	// Apply time to the tag if it isn't null
```
tgross, line 2403: Nitpick: this kind of comment doesn't really add anything to the code

nomad/job_endpoint.go
```diff
@@ -2391,3 +2398,19 @@ func (j *Job) GetServiceRegistrations(
 		},
 	})
 }
+
+func (j *Job) TagVersion(args *structs.JobApplyTagRequest, reply *structs.JobTagResponse) error {
```
tgross, line 2402: I think you just haven't implemented it yet, but don't forget you've still got to implement auth, metrics, and RPC forwarding at the top of this handler.

command/agent/job_endpoint.go
```diff
@@ -401,6 +406,91 @@ func (s *HTTPServer) jobRunAction(resp http.ResponseWriter, req *http.Request, j
...
+	case http.MethodPut, http.MethodPost:
+		return s.jobVersionApplyTag(resp, req, jobID, name)
+	case http.MethodDelete:
+		return s.jobVersionUnsetTag(resp, req, jobID, name)
+	default:
+		return nil, CodedError(405, ErrInvalidMethod)
+	}
+}
+
+// TODO: this is a copy of the function in command/job_history.go; any way to import it here?
```
tgross, line 426: That'd be a circular reference as `command` imports `command/agent`. You could make this function exported and then have `command` call it as `agent.ParseVersion`, but this code seems tiny enough that it's not worth bothering.

command/agent/job_endpoint.go
```diff
@@ -401,6 +406,91 @@ func (s *HTTPServer) jobRunAction(resp http.ResponseWriter, req *http.Request, j
...
 }
 
+func (s *HTTPServer) jobTagVersion(resp http.ResponseWriter, req *http.Request, jobID string, name string) (interface{}, error) {
+	switch req.Method {
+
+	// TODO: Request for reviewers: I am splitting by method here, but you'll notice both
+	// methods now call the same RPC method, which later splits them based on presence of args.Tag.
+	// So far the benefit of a method-based split here is that I can more easily create a slimmed-down
+	// TagVersionRequest for the RPC method, but I'm not sure if that's enough to justify the split.
+	// ...is this the best and most obvious place to set the .Tag or not?
```
tgross, line 416: These HTTP endpoints are definitely the best place to set the `.Tag`, because this is the first point where we go from 2 APIs (PUT and DELETE) to a single API (the RPC handler). 

Splitting the work into two methods seems fine to me. But if you didn't want to do that you could have each case here construct the args and then have the RPC call at the end of this `jobTagVersion` function after the switch statement. It makes it a little easier to read and to write tight little tests if you keep it the way you've got it here IMO, but it's a matter of taste.

command/job_tag.go
```diff
@@ -0,0 +1,220 @@
...
+func (c *JobTagCommand) Help() string {
+	helpText := `
+Usage: nomad job tag apply [options] <jobname>
+
+  Save a job version to prevent it from being garbage-collected and allow it to
+  be diffed and reverted by name.
+  
+  Example usage:
+ 
+    nomad job tag apply -name "My Golden Version" -description "The version of the job we can roll back to in the future if needed" <jobname>
```
tgross, line 26: Wrap command line help text at 80 cols if you can. For an example I'd add a `\` to move the description into the line below (or just shorten the text)

nomad/state/state_store.go
```diff
@@ -4885,6 +4900,105 @@ func (s *StateStore) updateJobStabilityImpl(index uint64, namespace, jobID strin
...
+		return err
+	}
+
+	duplicateVersionName := false
+	var job *structs.Job
+
+	for _, version := range versions {
+		// Allow for a tag to be updated (new description, for example) but otherwise don't allow a same-tagname to a different version.
+		if version.TaggedVersion != nil && version.TaggedVersion.Name == tag.Name && version.Version != *jobVersion {
+			duplicateVersionName = true
```
tgross, line 4942: There's no more useful work to do at this point. So rather than setting a bool you check later, you can just bail out early and return the error here.

nomad/state/state_store.go
```diff
@@ -4885,6 +4900,105 @@ func (s *StateStore) updateJobStabilityImpl(index uint64, namespace, jobID strin
...
+
+	if duplicateVersionName {
+		return fmt.Errorf("tag %q already exists on a different version of job %q", tag.Name, jobID)
+	}
+
+	if job == nil {
+		return fmt.Errorf("job %q version %d not found", jobID, *jobVersion)
+	}
+
+	copy := job.Copy()
```
tgross, line 4958: `copy` is a Go built-in function, which you're shadowing here. That's not _wrong_ but we try to avoid it because it looks funny.

nomad/state/state_store.go
```diff
@@ -4885,6 +4900,105 @@ func (s *StateStore) updateJobStabilityImpl(index uint64, namespace, jobID strin
...
+	for _, version := range versions {
+		if version.TaggedVersion != nil && version.TaggedVersion.Name == name {
+			job = version
+			break
+		}
+	}
+
+	if job == nil {
+		return fmt.Errorf("tag %q not found on job %q", name, jobID)
+	}
```
tgross, line 4994: Isn't this `JobVersionByTagName`?

philrenaud: It is, thanks for catching. I think I copied over the `updateTag` code here instead because it does duplicate-name checking (and this won't have to)

nomad/state/state_store.go
philrenaud: I've got a [Copy() method on JobTaggedVersion](https://github.com/hashicorp/nomad/compare/main...epic/23794-golden-versions#diff-297649e31d8e88fd89ff63dac7aff450569cca19009616f08c1d73d1e8239725R4530-R4538), (which is the tag itself) but I think I don't understand how to use it here.

Or does this look funny because of the variable name `copy` itself?

philrenaud: Update: TIL `shadowing`. Going to change the var name.

nomad/state/state_store.go
```diff
@@ -4885,6 +4900,89 @@ func (s *StateStore) updateJobStabilityImpl(index uint64, namespace, jobID strin
...
+
+	return txn.Commit()
+}
+
+func (s *StateStore) updateJobVersionTagImpl(index uint64, namespace, jobID string, jobVersion uint64, tag *structs.JobTaggedVersion, txn *txn) error {
+	ws := memdb.NewWatchSet()
+
+	// Note: could use JobByIDAndVersion to get the specific version we want here,
+	// but then we'd have to make a second lookup to make sure we're not applying a duplicate tag name
+	versions, err := s.JobVersionsByID(ws, namespace, jobID)
```
tgross, line 4931: The `JobVersionsByID` method needs to accept a watchset because it's being used by reads that might be blocking reads. But this caller is a write so it's not really relevant here. You can pass in `nil` as the watchset, because its `Add` method is designed to handle that safely.

```suggestion
	// Note: could use JobByIDAndVersion to get the specific version we want here,
	// but then we'd have to make a second lookup to make sure we're not applying a duplicate tag name
	versions, err := s.JobVersionsByID(nil, namespace, jobID)
```

nomad/state/state_store.go
```diff
@@ -4885,6 +4900,89 @@ func (s *StateStore) updateJobStabilityImpl(index uint64, namespace, jobID strin
...
+		return err
+	}
+
+	return txn.Commit()
+}
+
+func (s *StateStore) unsetJobVersionTagImpl(index uint64, namespace, jobID string, name string, txn *txn) error {
+	ws := memdb.NewWatchSet()
+
+	job, err := s.JobVersionByTagName(ws, namespace, jobID, name)
```
tgross, line 4972:
```suggestion
	job, err := s.JobVersionByTagName(nil, namespace, jobID, name)
```

command/job_tag.go
```diff
@@ -0,0 +1,219 @@
...
+	_, err = client.Jobs().TagVersion(jobID, versionStr, name, description, nil)
+	if err != nil {
+		c.Ui.Error(fmt.Sprintf("Error tagging job version: %s", err))
+		return 1
+	}
+
+	return 0
+}
+
+// #region unset
```
tgross, line 130: We always put commands in separate files because it makes it really easy to find them. So this file should be `job_tag_apply.go` and the other should be `job_tag_unset.go`. But you also end up needing a `job_tag.go` file anyways to have the short help text for both commands. (That is, if someone does `nomad job tag -help` you want that to give them a help text that shows both `job tag apply` and `job tag unset`.)

api/jobs.go
```diff
@@ -1643,3 +1643,37 @@ type JobStatusesRequest struct {
...
+			return nil, err
+		}
+		versionInt = *job.Version
+	} else {
+		var parseErr error
+		versionInt, parseErr = strconv.ParseUint(version, 10, 64)
+		if parseErr != nil {
+			return nil, fmt.Errorf("invalid version format: %q. Version must be a positive integer", version)
+		}
+	}
```
tgross, line 1668: My earlier comment here got resolved because of the pointer change (which looks great), but I think this bit still applies:

> Either way I would definitely require this function take in a uint64 though. Consider that callers outside the CLI (ex. the Terraform provider) will likely already have a uint64 from the Job object so forcing them to use a string here just makes for unnecessary conversions. Use the type system to enforce correctness here rather than parsing.

tgross (commented):
This is looking great @philrenaud!

command/job_tag.go
philrenaud: I gave this a shot in https://github.com/hashicorp/nomad/pull/23863/commits/050ac94bc3e81a6dbacf593386e1698f811fef24

and got some unusual results with respect to help text/synopsis/subcommands:

Running `nomad job --help` gives an empty entry for `tag`:
```
  ...
  Please see the individual subcommand help for detailed usage information.

Subcommands:
    action            Run a pre-defined action from a Nomad task
    ...
    stop              Stop a running job
    tag
    validate          Checks if a given job specification is valid
```

And running `nomad job tag --help` gives the `JobTagCommand.Synopsis()`  where I expected to see the `JobTagApplyCommand.Synopsis()`:

```
nomad job tag --help
This command is accessed by using one of the subcommands below.

Subcommands:
    apply    Interact with job version tags
    unset    Remove a tag from a job version.
```

I've tried comparing this to what's being done in `command/job_periodic.go` and thought my usage was similar. Am I missing something obvious?

tgross: For the `nomad job --help` you need to add `job tag` to the list of commands here: https://github.com/hashicorp/nomad/pull/23863/files#diff-761251ed6986d9fce6f663bba779a746182dd901e3ab3a1ecec1a21f8fb75debR524-R533

Your `nomad job tag --help` text looks correct -- it's not an alias for `job tag apply`.

api/jobs.go
philrenaud: My worry about using a uint64 here is that it's non-nilable, and `0` is a valid job version to pass in;

I want users to be able to not pass in a job version, and for us to infer they want the latest version to be tagged; I haven't been able to figure out how to do that with a uint64 unless it's a pointer instead.

tgross: Sure, but I wouldn't do that here in the Go API. Do that in the caller.

nomad/state/state_store.go
```diff
@@ -4885,6 +4900,96 @@ func (s *StateStore) updateJobStabilityImpl(index uint64, namespace, jobID strin
...
+		return err
+	}
+	if job == nil {
+		return fmt.Errorf("tag %q not found on job %q", name, jobID)
+	}
+
+	versionCopy := job.Copy()
+	versionCopy.TaggedVersion = nil
+	versionCopy.ModifyIndex = index
+	return s.upsertJobVersion(index, versionCopy, txn)
```
tgross, line 4990: If this version of the job is the latest version, don't we need to update the job table here too, similar to what we've done in the upsert code path?

nomad/state/state_store.go
```diff
@@ -4885,6 +4900,96 @@ func (s *StateStore) updateJobStabilityImpl(index uint64, namespace, jobID strin
...
+	txn := s.db.WriteTxn(index)
+	defer txn.Abort()
+
+	if err := s.unsetJobVersionTagImpl(index, namespace, jobID, name, txn); err != nil {
+		return err
+	}
+
+	return txn.Commit()
+}
+
```
tgross, line 4977: I think this method was part of the original design, but I it doesn't look like it has any more callers?

tgross (approved):
LGTM!

I've left a couple last items but once you've resolved those this is good-to-merge.

nomad/state/state_store.go
philrenaud: Yeah, great catch thank you

nomad/state/state_store.go
philrenaud: Good call, `unsetJobVersionTagImpl` is now called in a conditional within `UpdateJobVersionTag`

nomad/state/state_store.go
```diff
@@ -2306,6 +2306,21 @@ func (s *StateStore) JobVersionsByID(ws memdb.WatchSet, namespace, id string) ([
...
+	versions, err := s.JobVersionsByID(ws, namespace, id)
+	if err != nil {
+		return nil, err
+	}
+	for _, j := range versions {
+		if j.TaggedVersion != nil && j.TaggedVersion.Name == tagName {
+			return j, nil
+		}
+	}
+	return nil, nil
```
gulducat, line 2321: my nil pointer sense is tingling! to me `nil, nil` return for `(*pointer, error)` is a capital offense.

shouldn't it be an error if they asked for a tag by name, but no such tag exists?

nomad/state/state_store.go
```diff
@@ -4903,6 +4918,95 @@ func (s *StateStore) updateJobStabilityImpl(index uint64, namespace, jobID strin
...
+
+	latestJob, err := s.JobByID(nil, namespace, jobID)
+	if err != nil {
+		return err
+	}
+	if versionCopy.Version == latestJob.Version {
+		if err := txn.Insert("jobs", versionCopy); err != nil {
+			return err
+		}
+	}
```
gulducat, line 4980: I sure hope that `versionCopy` and `latestJob` are otherwise totally identical, since this replaces the latter with the former in the "jobs" table.

gulducat (approved):
mostly easy stuff, couple of concerns for your consideration!

command/job_tag_apply.go
```diff
@@ -0,0 +1,153 @@
...
+		return 1
+	}
+
+	// If the version is not provided, get the "active" version of the job
+	var versionInt uint64
+	if versionStr == "" {
+		q := &api.QueryOptions{
+			Namespace: namespace,
+		}
+		latestVersion, _, err := client.Jobs().Info(job, q)
```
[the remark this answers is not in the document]
tgross: Because versions can be 0, we need to query in order to have a `TagVersion` API that's both ergonomic and unambiguous.

nomad/state/state_store.go
tgross: Most of the single-object state store methods return `nil, nil` on non-existing objects, rather than some kind of `NotFound` error.

nomad/state/state_store.go
tgross: `job` and ``latestJob` are pointers to the exact same struct, so `versionCopy` is a copy of `job` that for safe upsert that needs to be done to any table where we have that pointer.

Issue #9917, opened 2021-01-29
extend quotas to include device plugins such as GPUs


henrikjohansen:
The existing resource quotas cover entities such as MHz compute and MB memory but device plugins such as GPUs are currently **not** supported. 

This is a real problem as a single job essentially can consume all available GPU resources leading to a cluster-wide starvation of critical resources.

GPU resource are already being fingerprintet : 
```
$ nomad node status -verbose 429c30c7 
...
Device Group Attributes
Device Group     = nvidia/gpu/Tesla T4
bar1             = 256 MiB
cores_clock      = 1590 MHz
display_state    = Enabled
driver_version   = 460.32.03
memory_clock     = 5001 MHz
memory           = 15109 MiB
pci_bandwidth    = 15760 MB/s
persistence_mode = Disabled
power            = 70 W
...
```

The ideal solution would be to extend the fingerprinting to include the number of GPU cores and expose those for use in resource quotas - this would bring GPU quotas on par with system CPU and memory quotas (_GPU resources could also be exposed as Mhz instead of core count to make those identical to CPU resources_).

For example : 

```
limit {
  region = "global"
  region_limit {
    cpu = 2500
    memory = 1000

    nvidia {
      cores = 2560   # number of CUDA cores
      memory = 15000 # MB GPU Memory
    }
  }
}

```

 **But** simply being able to restrict the number of devices using a resource quota would also be acceptable for now - for example :  

```
limit {
  region = "global"
  region_limit {
    cpu = 2500
    memory = 1000

    nvidia {
      devices = 2 # number of GPU devices
    }
  }
}
```

tgross:
Hi @henrikjohansen, and thanks for this suggestion! I've slightly edited the title just so that we don't confuse the ENT `quotas` feature for the `resource` block. There's some interesting trickiness to this because devices are plugins, so the resources they expose are fairly arbitrary from the perspective of the scheduler. Should be interesting to figure out!

henrikjohansen:
@tgross :+1: Well, your comment is precisely why I included the last example in :point_up:. A simple count of the number of devices is probably more realistic than exposing fine grained resources for all relevant device types :)

henrikjohansen:
@tgross This is getting a more and more prevalent issue for us since we have no way to control the utilization of GPU resources amongst our tenants leading to all sorts of problems.

I would :heart: to see the resource quotas feature of Nomad Enterprise enhanced so that the number of available GPUs could be limited per namespace :point_down: 

[...]

PR #23866, opened 2024-08-26
quota: parse device block

pkazmierczak:
related to: https://github.com/hashicorp/nomad/issues/9917
internal ref: https://hashicorp.atlassian.net/browse/NET-9858

~note to reviewers: This CE PR doesn't require an ENT counterpart, because for parsing quota spec ce and ent share the code. Follow-up PRs with updated `quota init` command and the actual logic of enforcing quotas will follow.~

Sorry for going back and forth on this,  I'm still confused about the ce/ent split in quotas. Of course we need an ent PR to handle ent structs: https://github.com/hashicorp/nomad-enterprise/pull/1787

### DIFF
--- a/api/util_test.go
+++ b/api/util_test.go
@@ -120,7 +120,7 @@ func testQuotaSpec() *QuotaSpec {
 					CPU:      pointerOf(2000),
 					MemoryMB: pointerOf(2000),
 					Devices: []*RequestedDevice{{
-						Name:  "gpu/nvidia/1080ti",
+						Name:  "nvidia/gpu/1080ti",
 						Count: pointerOf(uint64(2)),
 					}},
 				},
--- a/api/util_test.go
+++ b/api/util_test.go
@@ -119,6 +119,10 @@ func testQuotaSpec() *QuotaSpec {
 				RegionLimit: &Resources{
 					CPU:      pointerOf(2000),
 					MemoryMB: pointerOf(2000),
+					Devices: []*RequestedDevice{{
+						Name:  "gpu/nvidia/1080ti",
+						Count: pointerOf(uint64(2)),
+					}},
 				},
 			},
 		},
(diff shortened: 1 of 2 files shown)


api/util_test.go
```diff
@@ -119,6 +119,10 @@ func testQuotaSpec() *QuotaSpec {
 				RegionLimit: &Resources{
 					CPU:      pointerOf(2000),
 					MemoryMB: pointerOf(2000),
+					Devices: []*RequestedDevice{{
+						Name:  "gpu/nvidia/1080ti",
```
shoenig, line 123: order

command/quota_apply.go
```diff
@@ -280,9 +281,46 @@ func parseQuotaResource(result *api.Resources, list *ast.ObjectList) error {
...
+
+	return nil
+}
+
+func parseDeviceResource(result *[]*api.RequestedDevice, list *ast.ObjectList) error {
+	for _, o := range list.Elem().Items {
+		// Check for invalid keys
+		valid := []string{
+			"name",
+			"count",
```
shoenig, line 307: can also have `constraint` and `affinity`, no?

https://developer.hashicorp.com/nomad/docs/job-specification/device#device-parameters

pkazmierczak: it can, but do we want to include those in the quota spec? I understand what `constraint` would enable operators to do, but how would `affinity` work with a device quota?

pkazmierczak: or are you suggesting to just accept them as valid input but simply ignore at quota enforcement?

shoenig: Oh sorry I was thinking this was a normal resources block, but it's a quota block ... where these fields wouldn't exist

shoenig (approved):

PR #23881, opened 2024-08-27
quota: add device block to the quota init command

pkazmierczak:
related to: https://github.com/hashicorp/nomad/issues/9917
internal ref: https://hashicorp.atlassian.net/browse/NET-9858

### DIFF
--- a/command/quota_init.go
+++ b/command/quota_init.go
@@ -124,7 +124,7 @@ limit {
     memory     = 1000
     memory_max = 1000
     device {
-      name  = "gpu/nvidia/1080ti",
+      name  = "nvidia/gpu/1080ti",
       count = 1,
     }
   }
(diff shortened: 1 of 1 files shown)


command/quota_init.go
```diff
@@ -123,6 +123,10 @@ limit {
     cpu        = 2500
     memory     = 1000
     memory_max = 1000
+    device {
+      name  = "gpu/nvidia/1080ti",
```
shoenig, line 127: these are out of order - should be `nvidia/gpu/1080ti`

https://developer.hashicorp.com/nomad/docs/job-specification/device#-2

pkazmierczak: yikes, good catch.

shoenig: I got really far making the same mistake :upside_down_face:

shoenig (approved):

PR #23882, opened 2024-08-27
networking: option to enable ipv6 on bridge network

### LINKED ISSUE #14101: support ipv6 connectivity for the bridge network

gulducat:
Users may optionally enable IPv6 on the Nomad `bridge` network mode by setting `bridge_network_subnet_ipv6` in the client config.

Aside from that, there is a small functional change regarding our attempt to force-cleanup iptables if CNI fails to do so.  Now if the postrouting rule is not found, it only logs that instead of returning an error.  This way the new best-effort attempt to cleanup `ip6tables` can be less frightening in the logs if ipv6 isn't even enabled on the bridge.

Resolves #14101

### DIFF
--- a/client/allocrunner/cni/bridge.go
+++ b/client/allocrunner/cni/bridge.go
@@ -22,6 +22,7 @@ type NomadBridgeConfig struct {
 	BridgeName     string
 	AdminChainName string
 	IPv4Subnet     string
+	IPv6Subnet     string
 	HairpinMode    bool
 	ConsulCNI      bool
 }
@@ -39,6 +40,10 @@ func NewNomadBridgeConflist(conf NomadBridgeConfig) Conflist {
 	}
 	ipRoutes := []Route{
 		{Dst: "0.0.0.0/0"},
+	}
+	if conf.IPv6Subnet != "" {
+		ipRanges = append(ipRanges, []Range{{Subnet: conf.IPv6Subnet}})
+		ipRoutes = append(ipRoutes, Route{Dst: "::/0"})
 	}
 
 	plugins := []any{
--- a/command/agent/agent.go
+++ b/command/agent/agent.go
@@ -890,6 +890,7 @@ func convertClientConfig(agentConfig *Config) (*clientconfig.Config, error) {
 	conf.CNIConfigDir = agentConfig.Client.CNIConfigDir
 	conf.BridgeNetworkName = agentConfig.Client.BridgeNetworkName
 	conf.BridgeNetworkAllocSubnet = agentConfig.Client.BridgeNetworkSubnet
+	conf.BridgeNetworkAllocSubnetIPv6 = agentConfig.Client.BridgeNetworkSubnetIPv6
 	conf.BridgeNetworkHairpinMode = agentConfig.Client.BridgeNetworkHairpinMode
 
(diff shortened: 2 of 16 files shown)


client/allocrunner/cni/bridge.go
```diff
@@ -40,6 +41,10 @@ func NewNomadBridgeConflist(conf NomadBridgeConfig) Conflist {
 	ipRoutes := []Route{
 		{Dst: "0.0.0.0/0"},
 	}
+	if conf.IPv6Subnet != "" {
+		ipRanges = append(ipRanges, []Range{{Subnet: conf.IPv6Subnet}})
+		ipRoutes = append(ipRoutes, Route{Dst: "::/0"})
```
gulducat, line 46: these are the meat and potatoes; everything else revolves around this.

command/agent/testdata/basic.json
```diff
@@ -76,6 +76,7 @@
       "alloc_mounts_dir": "/tmp/mounts",
       "bridge_network_name": "custom_bridge_name",
       "bridge_network_subnet": "custom_bridge_subnet",
+      "bridge_network_subnet_ipv6": "custom_bridge_subnet_ipv6",
```
tgross, line 79: It's a little weird that our tests have invalid values here (I think these aren't valid?)

tgross (approved):
LGTM :+1:

command/agent/testdata/basic.json
gulducat: It's a good point - they're not validated anywhere until they actually get used at allocation-time.

I always hated finding out about bad client config that way, when an alloc fails through no fault of the jobspec.  I would prefer to have the client fail to start instead of be sitting around quietly poisoned.  Would this be a good place to add a check?

https://github.com/hashicorp/nomad/pull/23882/files#diff-430a3b367fc84668c6ae1cb6936b7a7b5c5b321f42aa754c23618f848faf87b1R893

I can also add validation for custom ipv4 subnet while I'm in here, too. A major version bump is probably a good time to go for that, eh?

command/agent/agent.go
```diff
@@ -889,7 +889,28 @@ func convertClientConfig(agentConfig *Config) (*clientconfig.Config, error) {
...
+		}
+		conf.BridgeNetworkAllocSubnet = ipv4Subnet
+	}
+	ipv6Subnet := agentConfig.Client.BridgeNetworkSubnetIPv6
+	if ipv6Subnet != "" {
+		ip, _, err := net.ParseCIDR(ipv6Subnet)
+		if err != nil {
+			return nil, fmt.Errorf("invalid bridge_network_subnet_ipv6: %w", err)
+		}
+		if ip.To16() == nil {
```
tgross, line 909: Unfortunately this doesn't work to validate an IPv6 address, as you can get a 16-byte address from an IPv4 address. See the example code for this method: https://pkg.go.dev/net#IP.To16 of a case of that.

I think what you want to do here is check `ip.To4() != nil`. You know it's a valid IP address already, so now you want to check that it's not-IPv6.

gulducat: Nice catch!  b0e4273

tgross (approved):
LGTM

PR #23889, opened 2024-08-28
Compare versions by name and version number with the nomad history command

### LINKED ISSUE #23879: `nomad job plan` `-version` flag
- Allow job diff/plan requests to take a string param for diffing against a specific (named) version, rather than against their predecessor.

philrenaud:
- Adds `diff_tag` and `diff_version` flags for the nomad job versions API endpoint, ~and for the `nomad job plan`~ and `nomad job history` commands.

- Adds `Tag` info to `nomad job history` output

## Usage:

(Prerequisite tag a version of your job with `nomad job tag apply`, per https://github.com/hashicorp/nomad/pull/23863)

~### Diff against a specific tag~
```
nomad job plan -diff-tag=my-tag-name difftester.nomad.hcl
```
**N.B.:** Decided against this, as `plan` implies "looking to change things against currently-running. To implement later: `nomad job diff` CLI command.

~### Diff against a specific version, whether tagged or not~
```
nomad job plan -diff-version=2 difftester.nomad.hcl
```
**N.B.:** Decided against this, as `plan` implies "looking to change things against currently-running. To implement later: `nomad job diff` CLI command.

### See history including tag info, with diffs
(each version will show a diff to its predecessor, except for version 0 / first non-GC'd version, which has no known predecessor)
```
nomad job history -p difftester
```

### See history with diffs against a specific tag
```
nomad job history -p -diff-tag=my-tag-name difftester
```

### See history with diffs against a specific version, whether tagged or not
```
nomad job history -p -diff-version=2 difftester
```

Resolves #23879

### DIFF
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -263,9 +263,9 @@ func (j *Jobs) ScaleStatus(jobID string, q *QueryOptions) (*JobScaleStatusRespon
 
 // Versions is used to retrieve all versions of a particular job given its
 // unique ID.
-func (j *Jobs) Versions(jobID string, diffs bool, q *QueryOptions) ([]*Job, []*JobDiff, *QueryMeta, error) {
+func (j *Jobs) Versions(jobID string, diffs bool, diffTag string, diffVersion string, q *QueryOptions) ([]*Job, []*JobDiff, *QueryMeta, error) {
 	var resp JobVersionsResponse
-	qm, err := j.client.query(fmt.Sprintf("/v1/job/%s/versions?diffs=%v", url.PathEscape(jobID), diffs), &resp, q)
+	qm, err := j.client.query(fmt.Sprintf("/v1/job/%s/versions?diffs=%t&compare_to_tag=%s&compare_to_version=%d", url.PathEscape(jobID), diffs, diffTag, diffVersion), &resp, q)
 	if err != nil {
 		return nil, nil, nil, err
 	}
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -263,9 +263,34 @@ func (j *Jobs) ScaleStatus(jobID string, q *QueryOptions) (*JobScaleStatusRespon
 
 // Versions is used to retrieve all versions of a particular job given its
 // unique ID.
-func (j *Jobs) Versions(jobID string, diffs bool, diffTag string, diffVersion string, q *QueryOptions) ([]*Job, []*JobDiff, *QueryMeta, error) {
+func (j *Jobs) Versions(jobID string, diffs bool, q *QueryOptions) ([]*Job, []*JobDiff, *QueryMeta, error) {
+	opts := &VersionsOptions{
+		Diffs: diffs,
+	}
+	return j.VersionsOpts(jobID, opts, q)
+}
+
+type VersionsOptions struct {
+	Diffs       bool
+	DiffTag     string
+	DiffVersion *uint64
+}
+
+func (j *Jobs) VersionsOpts(jobID string, opts *VersionsOptions, q *QueryOptions) ([]*Job, []*JobDiff, *QueryMeta, error) {
 	var resp JobVersionsResponse
-	qm, err := j.client.query(fmt.Sprintf("/v1/job/%s/versions?diffs=%t&diff_tag=%s&diff_version=%s", url.PathEscape(jobID), diffs, diffTag, diffVersion), &resp, q)
+
+	qp := url.Values{}
+	if opts != nil {
+		qp.Add("diffs", strconv.FormatBool(opts.Diffs))
+		if opts.DiffTag != "" {
+			qp.Add("diff_tag", opts.DiffTag)
+		}
+		if opts.DiffVersion != nil {
+			qp.Add("diff_version", strconv.FormatUint(*opts.DiffVersion, 10))
+		}
+	}
+
+	qm, err := j.client.query(fmt.Sprintf("/v1/job/%s/versions?%s", url.PathEscape(jobID), qp.Encode()), &resp, q)
 	if err != nil {
 		return nil, nil, nil, err
 	}
@@ -446,7 +471,7 @@ func (j *Jobs) PeriodicForce(jobID string, q *WriteOptions) (string, *WriteMeta,
 type PlanOptions struct {
 	Diff           bool
 	PolicyOverride bool
-	DiffVersion    string
+	DiffVersion    *uint64
 	DiffTagName    string
 }
 
@@ -470,7 +495,10 @@ func (j *Jobs) PlanOpts(job *Job, opts *PlanOptions, q *WriteOptions) (*JobPlanR
 	if opts != nil {
 		req.Diff = opts.Diff
 		req.PolicyOverride = opts.PolicyOverride
-		req.DiffVersion = opts.DiffVersion
+		if opts.DiffVersion != nil {
+			req.DiffVersion = opts.DiffVersion
+		}
+
 		req.DiffTagName = opts.DiffTagName
 	}
 
@@ -1481,7 +1509,7 @@ type JobPlanRequest struct {
 	Job            *Job
 	Diff           bool
 	PolicyOverride bool
-	DiffVersion    string
+	DiffVersion    *uint64
 	DiffTagName    string
 	WriteRequest
 }
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -471,8 +471,6 @@ func (j *Jobs) PeriodicForce(jobID string, q *WriteOptions) (string, *WriteMeta,
 type PlanOptions struct {
 	Diff           bool
 	PolicyOverride bool
-	DiffVersion    *uint64
-	DiffTagName    string
 }
 
 func (j *Jobs) Plan(job *Job, diff bool, q *WriteOptions) (*JobPlanResponse, *WriteMeta, error) {
@@ -495,9 +493,6 @@ func (j *Jobs) PlanOpts(job *Job, opts *PlanOptions, q *WriteOptions) (*JobPlanR
 	if opts != nil {
 		req.Diff = opts.Diff
 		req.PolicyOverride = opts.PolicyOverride
-		req.DiffVersion = opts.DiffVersion
-
-		req.DiffTagName = opts.DiffTagName
 	}
 
 	var resp JobPlanResponse
@@ -1507,8 +1502,6 @@ type JobPlanRequest struct {
 	Job            *Job
 	Diff           bool
 	PolicyOverride bool
-	DiffVersion    *uint64
-	DiffTagName    string
 	WriteRequest
 }
 
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -446,6 +446,8 @@ func (j *Jobs) PeriodicForce(jobID string, q *WriteOptions) (string, *WriteMeta,
 type PlanOptions struct {
 	Diff           bool
 	PolicyOverride bool
+	DiffVersion    string
+	DiffTagName    string
 }
 
 func (j *Jobs) Plan(job *Job, diff bool, q *WriteOptions) (*JobPlanResponse, *WriteMeta, error) {
@@ -468,6 +470,8 @@ func (j *Jobs) PlanOpts(job *Job, opts *PlanOptions, q *WriteOptions) (*JobPlanR
 	if opts != nil {
 		req.Diff = opts.Diff
 		req.PolicyOverride = opts.PolicyOverride
+		req.DiffVersion = opts.DiffVersion
+		req.DiffTagName = opts.DiffTagName
 	}
 
 	var resp JobPlanResponse
@@ -1477,6 +1481,8 @@ type JobPlanRequest struct {
 	Job            *Job
 	Diff           bool
 	PolicyOverride bool
+	DiffVersion    string
+	DiffTagName    string
 	WriteRequest
 }
 
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -265,7 +265,7 @@ func (j *Jobs) ScaleStatus(jobID string, q *QueryOptions) (*JobScaleStatusRespon
 // unique ID.
 func (j *Jobs) Versions(jobID string, diffs bool, diffTag string, diffVersion string, q *QueryOptions) ([]*Job, []*JobDiff, *QueryMeta, error) {
 	var resp JobVersionsResponse
-	qm, err := j.client.query(fmt.Sprintf("/v1/job/%s/versions?diffs=%t&compare_to_tag=%s&compare_to_version=%d", url.PathEscape(jobID), diffs, diffTag, diffVersion), &resp, q)
+	qm, err := j.client.query(fmt.Sprintf("/v1/job/%s/versions?diffs=%t&diff_tag=%s&diff_version=%s", url.PathEscape(jobID), diffs, diffTag, diffVersion), &resp, q)
 	if err != nil {
 		return nil, nil, nil, err
 	}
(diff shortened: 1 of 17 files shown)


nomad/job_endpoint.go
```diff
@@ -1298,19 +1299,66 @@ func (j *Job) GetJobVersions(args *structs.JobVersionsRequest,
...
+					}
 				}
 				reply.Index = maxModifyIndex
+
+				if compareStatic && compareVersion == nil {
+					if args.DiffTagName != "" {
+						return fmt.Errorf("tag %q not found", args.DiffTagName)
+					}
+					return fmt.Errorf("version %q not found", args.DiffVersion)
+				}
```
tgross, line 1346: With this block it's not really clear to me what this `compareStatic` variable is for. It looks like you're setting this only when `DiffTagName` or `DiffVersion` are set, but not when `Diffs == true`, but won't these cases have been detected earlier on lines 1307-1326?

api/jobs.go
```diff
@@ -266,9 +266,9 @@ func (j *Jobs) ScaleStatus(jobID string, q *QueryOptions) (*JobScaleStatusRespon
 
 // Versions is used to retrieve all versions of a particular job given its
 // unique ID.
-func (j *Jobs) Versions(jobID string, diffs bool, q *QueryOptions) ([]*Job, []*JobDiff, *QueryMeta, error) {
+func (j *Jobs) Versions(jobID string, diffs bool, diffTag string, diffVersion string, q *QueryOptions) ([]*Job, []*JobDiff, *QueryMeta, error) {
 	var resp JobVersionsResponse
-	qm, err := j.client.query(fmt.Sprintf("/v1/job/%s/versions?diffs=%v", url.PathEscape(jobID), diffs), &resp, q)
+	qm, err := j.client.query(fmt.Sprintf("/v1/job/%s/versions?diffs=%t&diff_tag=%s&diff_version=%s", url.PathEscape(jobID), diffs, diffTag, diffVersion), &resp, q)
```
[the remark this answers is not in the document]
philrenaud: I ended up making my options params into a `url.Values{}` which I then `.encode()` when appending into the URL string. I dug into the `url.Values{}.encode()`  code a bit and it looks like it `QueryEscape()`s, so I think this will be solved that way (I suspect it wouldn't be too bad for me to add a possibly-redundant QueryEscape here but doesn't seem necessary now)

nomad/job_endpoint.go
philrenaud: Yeah you're exactly right, good catch!

philrenaud: OH WAIT I just realized `compareStatic` is there because `compareVersion` is a uint64, not a pointer to a uint64, and therefore will instantiate at 0 here!

command/job_history.go
```diff
@@ -222,15 +246,14 @@ func parseVersion(input string) (uint64, bool, error) {
...
-		return fmt.Errorf("Number of job versions %d doesn't match number of diffs %d", vLen, dLen)
-	}
 
 	for i, version := range versions {
 		var diff *api.JobDiff
 		var nextVersion uint64
-		if i+1 <= dLen {
+		if dLen > i && diffs[i] != nil {
 			diff = diffs[i]
+		}
```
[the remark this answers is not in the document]
philrenaud: Understood, fixed! `dLen > i` is meant to prevent exactly that dereference panic.

nomad/job_endpoint.go
```diff
@@ -1816,9 +1864,46 @@ func (j *Job) Plan(args *structs.JobPlanRequest, reply *structs.JobPlanResponse)
...
+		versions, err := snap.JobVersionsByID(ws, args.RequestNamespace(), args.Job.ID)
+		if err != nil {
+			return err
+		}
+		for _, version := range versions {
+			if version.TaggedVersion != nil && version.TaggedVersion.Name == args.DiffTagName {
+				existingJob = version
+				break
+			}
+		}
```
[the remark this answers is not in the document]
philrenaud: Good suggestion, done

nomad/job_endpoint.go
```diff
@@ -1816,9 +1863,29 @@ func (j *Job) Plan(args *structs.JobPlanRequest, reply *structs.JobPlanResponse)
...
-		return err
+	var existingJob *structs.Job
+
+	if args.DiffVersion != nil {
+		existingJob, err = snap.JobByIDAndVersion(ws, args.RequestNamespace(), args.Job.ID, *args.DiffVersion)
+		if err != nil {
+			return err
+		}
+		if existingJob == nil {
+			return fmt.Errorf("version %q not found", *args.DiffVersion)
```
philrenaud, line 1874: Note: getting byte output here instead of a version number. I'm doing a bad conversion someplace.

```
nomad job plan -diff-version=7 ../nomad-job-specs/difftester5.nomad
Error during plan: Unexpected response code: 500 (version '\a' not found)

nomad job plan -diff-version=8 ../nomad-job-specs/difftester5.nomad
Error during plan: Unexpected response code: 500 (version '\b' not found)

nomad job plan -diff-version=9 ../nomad-job-specs/difftester5.nomad
Error during plan: Unexpected response code: 500 (version '\t' not found)
```

tgross: Yeah, that's just how `%q` works. You'll want to use %d there.

api/jobs.go
```diff
@@ -471,6 +498,11 @@ func (j *Jobs) PlanOpts(job *Job, opts *PlanOptions, q *WriteOptions) (*JobPlanR
 	if opts != nil {
 		req.Diff = opts.Diff
 		req.PolicyOverride = opts.PolicyOverride
+		if opts.DiffVersion != nil {
+			req.DiffVersion = opts.DiffVersion
+		}
+
```
tgross, line 504: These are both pointers, so this is safe and saves a comparison:

```suggestion
		req.DiffVersion = opts.DiffVersion
```

command/job_history.go
```diff
@@ -118,6 +134,25 @@ func (c *JobHistoryCommand) Run(args []string) int {
...
+
+	if diffTag != "" && diffVersionFlag != "" {
+		c.Ui.Error("-diff-tag and -diff-version are mutually exclusive")
+		return 1
+	}
+
+	if diffVersionFlag != "" {
+		parsedDiffVersion, err := strconv.ParseUint(diffVersionFlag, 10, 64)
+		if err != nil {
+			c.Ui.Error(fmt.Sprintf("Error parsing diff version: %s", err))
```
tgross, line 150: If we show the exact flag they were using that'll be slightly more clear:

```suggestion
			c.Ui.Error(fmt.Sprintf("Error parsing -diff-version: %s", err))
```

command/job_history.go
```diff
@@ -222,15 +262,14 @@ func parseVersion(input string) (uint64, bool, error) {
...
 	dLen := len(diffs)
-	if dLen != 0 && vLen != dLen+1 {
-		return fmt.Errorf("Number of job versions %d doesn't match number of diffs %d", vLen, dLen)
-	}
 
 	for i, version := range versions {
 		var diff *api.JobDiff
 		var nextVersion uint64
-		if i+1 <= dLen {
+		if dLen > i {
```
tgross, line 269: This still feels fuzzy to me. is `dLen >= vLen` or `vLen >= dLen`? I would expect there'd potentially be more versions than diffs, but maybe I'm misunderstanding what to expect here. In which case, we'd end up dereferencing past the last diff and panic.

Maybe we need a test to exercise this function directly?

nomad/job_endpoint.go
tgross: Ok, that makes sense but I still don't understand where the word "static" comes from in the variable name.

nomad/job_endpoint.go
```diff
@@ -1298,19 +1298,65 @@ func (j *Job) GetJobVersions(args *structs.JobVersionsRequest,
...
+						// it should either be compared to the NEXT version in out[], or to the pre-specified version from compareStatic.
+						if !compareStatic {
+							if i+1 < len(out) && out[i+1] != nil {
+								compareVersion = out[i+1]
+							} else {
+								break
+							}
+						}
+
+						old, new := compareVersion, out[i]
```
tgross, line 1359: Unconditionally assigning from the slice lets you avoid checking the length twice and doing the nil check (although I'm not sure that actually does anything anyways). We can tighten this up a bit like so:

```suggestion
					for i := 0; i < len(out)-1; i++ {
						// For each of your job's versions, counting from highest version to lowest
						// it should either be compared to the NEXT version in out[], or to the pre-specified version from compareStatic.

						old, new := out[i+1], out[i]
						if compareStatic {
							old = compareVersion
						}
```

command/job_history.go
philrenaud: I'll definitely have a test for this, but it previously made sure that "versions is at least one fewer than the last of the diffs", or in other words, "The last version (version 0) never gets a diff attached to it" (because before this PR, all diffs were against their preceding version, and version 0 didn't have a known preceding version)

That said: we assigned a `nextVersion` variable here and never used it — the line that was using it has been commented out for several years. I think I'll delete that and put the logic back to the way it was before this PR for this particular condition.

nomad/job_endpoint.go
philrenaud: changed this to `compareSpecificVersion` instead

nomad/job_endpoint.go
```diff
@@ -1298,19 +1298,72 @@ func (j *Job) GetJobVersions(args *structs.JobVersionsRequest,
...
+						return fmt.Errorf("tag %q not found", args.DiffTagName)
+					}
+					return fmt.Errorf("version %d not found", *args.DiffVersion)
+				}
+
 				// Compute the diffs
+
+				// TODO:
+				// - tests
+				// - Remove job plan getting diffs and make it into a new command, `nomad job diff`
```
tgross, line 1350: Is this TODO for this PR or intended for another PR targeting the same working branch?

command/job_history_test.go
```diff
@@ -166,3 +166,209 @@ namespace "default" {
...
+		// Diffs show up for all versions except the first one
+		must.True(t, strings.Contains(blocks[0], "Diff"))
+		must.True(t, strings.Contains(blocks[1], "Diff"))
+		must.True(t, strings.Contains(blocks[2], "Diff"))
+		must.False(t, strings.Contains(blocks[3], "Diff"))
+
+		// Check that the diffs are specifically against their predecessor
+		must.True(t, strings.Contains(blocks[0], "\"3\" => \"4\""))
+		must.True(t, strings.Contains(blocks[1], "\"2\" => \"3\""))
+		must.True(t, strings.Contains(blocks[2], "\"1\" => \"2\""))
```
tgross, line 245:
```suggestion
		// Diffs show up for all versions except the first one
		must.StrContains(t, blocks[0], "Diff")
		must.StrContains(t, blocks[1], "Diff")
		must.StrContains(t, blocks[2], "Diff")
		must.StrNotContains(t, blocks[3], "Diff")

		// Check that the diffs are specifically against their predecessor
		must.StrContains(t, blocks[0], `"3" => "4"`)
		must.StrContains(t, blocks[1], `"2" => "3"`)
		must.StrContains(t, blocks[2], `"1" => "2"`)
```

Same for all the other cases where you've got `must.True` with `strings.Contains` here.

command/job_history_test.go
```diff
@@ -166,3 +166,209 @@ namespace "default" {
...
+		out := ui.OutputWriter.String()
+		rawBlocks := strings.Split(out, "Version")
+		// trim empty blocks from whitespace in the output
+		var blocks []string
+		for _, block := range rawBlocks {
+			trimmed := strings.TrimSpace(block)
+			if trimmed != "" {
+				blocks = append(blocks, trimmed)
+			}
+		}
```
tgross, line 274: You can write a function inside the top of the test like this:


```go
	blocksFromOutput := func(out string) []string {
		t.Helper()
		rawBlocks := strings.Split(out, "Version")
		// trim empty blocks from whitespace in the output
		var blocks []string
		for _, block := range rawBlocks {
			trimmed := strings.TrimSpace(block)
			if trimmed != "" {
				blocks = append(blocks, trimmed)
			}
		}
		return blocks
	}
```

That you can then call in each of these tests, rather than copying the code block.

command/job_history_test.go
```diff
@@ -166,3 +166,209 @@ namespace "default" {
...
+
+		// Check that we have 4 versions
+		must.Eq(t, 4, len(blocks))
```
tgross, line 232:
```suggestion
		must.Len(t, 4, blocks)
```

Same for the other tests.

command/job_history_test.go
```diff
@@ -166,3 +166,209 @@ namespace "default" {
...
+	state := srv.Agent.Server().State()
+
+	// Create a job with multiple versions
+	job := mock.Job()
+
+	job.ID = "test-job-history"
+	job.TaskGroups[0].Count = 1
+	must.NoError(t, state.UpsertJob(structs.MsgTypeTestSetup, 1000, nil, job))
+
+	v2 := job.Copy()
```
tgross, line 185: This test might be a little clearer if `job` was called `v0` and then this would be `v1`.

command/job_history_test.go
```diff
@@ -166,3 +166,209 @@ namespace "default" {
...
+		must.Eq(t, 4, strings.Count(out, "Version"))
+		must.Eq(t, 3, strings.Count(out, "Diff"))
+
+		// Diffs show up for all versions except the specified one
+		must.False(t, strings.Contains(blocks[0], "Diff"))
+		must.True(t, strings.Contains(blocks[1], "Diff"))
+		must.True(t, strings.Contains(blocks[2], "Diff"))
+		must.True(t, strings.Contains(blocks[3], "Diff"))
+
+		// Check that the diffs are specifically against the tagged version (which has a count of 2)
```
tgross, line 287:
```suggestion
		// Check that the diffs are specifically against the tagged version (which has a count of 4)
```

command/job_history_test.go
```diff
@@ -166,3 +166,209 @@ namespace "default" {
...
+		must.Eq(t, 4, strings.Count(out, "Version"))
+		must.Eq(t, 3, strings.Count(out, "Diff"))
+
+		// Diffs show up for all versions except the specified one
+		must.True(t, strings.Contains(blocks[0], "Diff"))
+		must.False(t, strings.Contains(blocks[1], "Diff"))
+		must.True(t, strings.Contains(blocks[2], "Diff"))
+		must.True(t, strings.Contains(blocks[3], "Diff"))
+
+		// Check that the diffs are specifically against the tagged version (which has a count of 2)
```
tgross, line 324:
```suggestion
		// Check that the diffs are specifically against the tagged version (which has a count of 3)
```

command/job_history_test.go
```diff
@@ -166,3 +166,209 @@ namespace "default" {
...
+		must.True(t, strings.Contains(blocks[1], "\"4\" => \"3\""))
+		must.True(t, strings.Contains(blocks[2], "\"4\" => \"2\""))
+		must.True(t, strings.Contains(blocks[3], "\"4\" => \"1\""))
+
+	})
+
+	t.Run("With diffs against another specific version", func(t *testing.T) {
```
tgross, line 294: Does this test tell us something the previous one didn't?

tgross (approved):
LGTM!

I've left some suggestions to improve the tests but nothing blocking once you've got those resolved.

command/job_history_test.go
philrenaud: Only a little (that -diff-version isn't just tagged to the one-before-latest version), but possibly not enough to validate the test's run.

aimeeu:
@philrenaud I added the `theme/golden-versions` tag so I don't lose track of this when creating the new feature docs.

PR #23899, opened 2024-09-02
cli: quota status extension for devices

pkazmierczak:
This `quota status` CLI extension aggregates over all devices that are part of a quota and lists their usage/limit.

### DIFF
--- a/command/quota_status.go
+++ b/command/quota_status.go
@@ -6,6 +6,7 @@ package command
 import (
 	"encoding/base64"
 	"fmt"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


tgross (commented):
LGTM code-wise. 

It's a little odd from a semantics perspective though, because devices aren't fungible between each other in the way memory and CPU is. Consider a region limit with 2 devices A (limit 3) and B (limit 4). All the following combinations of usage get written as `4 / 7`: A0 + B4, A1 + B3, A2 + B2, A3 + B1, A4 + B0. It's not debuggable here what device usage is causing the limit.

Maybe we should have a separate listing of per-device quotas?

pkazmierczak:
> Maybe we should have a separate listing of per-device quotas?

I hear this. Refactored in 22cf5a7. Example output:
```
Name        = default-quota
Description = Limit the shared default namespace
Limits      = 2

Quota Limits
Region  CPU Usage  Core Usage  Memory Usage  Memory Max Usage  Variables Usage
eu      - / 3500   - / 2       - / 2000      - / 2000          - / inf
global  0 / 2500   0 / inf     0 / 1000      0 / 1000          0 / 1000

Quota Device Limits
Region  Device Name        Device Usage
eu      nvidia/gpu/1080ti  - / 2
global  nvidia/gpu/1080ti  1 / 1
```

tgross (approved):
LGTM!

PR #23902, opened 2024-09-03
WI: use parent job ID for subject claim

tgross:
When we use the job ID in creating the subject claim (`sub`) for workload identities, we forgot to use the parent job ID when that's available. Child job IDs have a random component that makes them unsuitable for the subject field.

Ref: https://github.com/hashicorp/nomad/pull/23817#discussion_r1717490323
Ref: https://hashicorp.atlassian.net/browse/NET-10714


gulducat (approved):
lgtm!

PR #23907, opened 2024-09-04
Differentiate the error message for different rpc calls

Juanadelacuesta:
When making ControllerPublishVolume, NodePublishVolume, and NodeStageVolume calls, in case of a `FailedPrecondition` error, Nomad returns the same error message, this PR makes a distinction to help users differentiate between the 3 calls.

### DIFF
--- a/plugins/csi/client.go
+++ b/plugins/csi/client.go
@@ -815,7 +815,7 @@ func (c *client) NodeStageVolume(ctx context.Context, req *NodeStageVolumeReques
 				"volume %q is already staged to %q but with incompatible capabilities for this request: %v",
 				req.ExternalID, req.StagingTargetPath, err)
 		case codes.FailedPrecondition:
-			err = fmt.Errorf("volume %q is already published on another node and does not have MULTI_NODE volume capability: %v",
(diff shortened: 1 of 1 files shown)


plugins/csi/client.go
```diff
@@ -881,7 +881,7 @@ func (c *client) NodePublishVolume(ctx context.Context, req *NodePublishVolumeRe
 				"volume %q is already published at target path %q but with capabilities or a read_only setting incompatible with this request: %v",
 				req.ExternalID, req.TargetPath, err)
 		case codes.FailedPrecondition:
-			err = fmt.Errorf("volume %q is already published on another node and does not have MULTI_NODE volume capability: %v",
+			err = fmt.Errorf("volume %q does not have MULTI_NODE volume capability: %v",
```
gulducat, line 884: these two are different from `ControllerPublishVolume` now, but they're still the same as each other?

gulducat: that makes sense. I suppose I figured while we're differentiating error messages that these could be made different too, but it does seem less important than distinguishing between controller plugin activity vs node plugin activity. :+1:

gulducat (approved):
lgtm!

PR #23910, opened 2024-09-04
services: reject node secret for Read/List RPC

tgross:
As of Nomad 1.5.0, Nomad clients never make RPC requests to the ServiceRegistrationList/Read RPC without using a specific Workload Identity rather than the node secret. Nomad servers already have an anti-entropy behavior where terminal allocations / lost nodes get their services cleaned up, so there's no future reason for clients to have this access. Tighten the ACL permissions on these RPCs so that node secrets are no longer valid tokens.

Ref: https://hashicorp.atlassian.net/browse/NET-10009
Ref: https://developer.hashicorp.com/nomad/docs/release-notes/nomad/upcoming#nomad-1-9-0

### DIFF
--- a/.changelog/23910.txt
+++ b/.changelog/23910.txt
@@ -1,0 +1,3 @@
+```release-note:breaking-change
+services: Clients older than 1.5.0 will fail to read Nomad native services via template blocks
+```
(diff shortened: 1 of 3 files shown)


gulducat (approved):
lgtm!

PR #23912, opened 2024-09-04
jobspec: remove HCL1 support

tgross:
This changeset removes support for parsing jobspecs via the long-deprecated HCLv1.

Fixes: https://github.com/hashicorp/nomad/issues/20195
Ref: https://hashicorp.atlassian.net/browse/NET-10220

Issue #20195, opened 2024-03-22
Intent to Deprecate HCLv1


schmichael:
Nomad adopted [HCLv2](https://developer.hashicorp.com/nomad/docs/job-specification/hcl2) for Jobspecs in 2020. HCLv2 is not strictly backward compatible with HCLv1, so we have maintained the HCLv1 jobspec parser behind the [`nomad job run -hcl1 ...` flag](https://developer.hashicorp.com/nomad/docs/commands/job/run#hcl1), the [job parse HCLv1 field](https://developer.hashicorp.com/nomad/api-docs/jobs#hclv1), and the [`jobspec`](https://github.com/hashicorp/nomad/tree/v1.7.6/jobspec) Go package.

## Deprecation Plan

The time has come to deprecate HCLv1.

* After Nomad 1.8.0 no new jobspsec fields will be added to the `jobspec` package.
* In Nomad 1.9.0 all hclv1 support will be removed.

### TODO

* [x] Add deprecation warning to CLI in Nomad 1.8.0 https://github.com/hashicorp/nomad/pull/23913
* [x] Add deprecation comments to `jobspec` package in Nomad 1.8.0 https://github.com/hashicorp/nomad/pull/23913
* [x] Add deprecation notes to docs: https://developer.hashicorp.com/nomad/docs/release-notes/nomad/upcoming#nomad-1-9-0
* [ ] Create a PR for total removal in Nomad 1.9.0 https://github.com/hashicorp/nomad/pull/23912

## Resources

* Original HCLv2 issue: #7220

42wim:
Hi, I'm wondering if there will be a solution for the heredoc issue before hcl1 gets removed ?
- https://github.com/hashicorp/nomad/issues/10777
- https://github.com/hashicorp/nomad/issues/9838

tgross:
Yeah we'll certainly need to do that.

tgross:
Wanted to follow up on this after my investigations in https://github.com/hashicorp/nomad/issues/9838. There's no way for us to do this without direct support in HCL. Adding support for this was previously rejected by the core HCL maintainer team, but even if we got past that (probably possible) it's a lot of work at the very "bottom" of the HCL stack, and we don't have the resources to do this work currently. Given that there's a workaround for this already (either manually escaping in the `template.data` or using `template.source`), we're not going to block HCL1 removal for heredoc support.

PR #23913, opened 2024-09-04
jobspec: add deprecation warnings for removal of HCLv1

tgross:
In Nomad 1.9.0 we're removing support for HCLv1. This has been long deprecated, but this changeset adds warnings to the command line and package documentation to make it that much more obvious.

Ref: https://github.com/hashicorp/nomad/pull/23912
Ref: https://github.com/hashicorp/nomad/issues/20195

Note to reviewers: this branch targets 1.8.x and not `main` because the PR with the removal is already targeting `main`.


schmichael (approved):
Thanks for doing this!

Issue #23875, opened 2024-08-26
1.8.3 crash


sorenisanerd:
Haven't had a chance to dig into it very much, but I got a bunch of these after upgrading from 1.7.7 to 1.8.3. 

```
2024-08-25 20:24:02.951	panic: runtime error: invalid memory address or nil pointer dereference
2024-08-25 20:24:02.951	[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1b6a957]
2024-08-25 20:24:02.952	goroutine 2398 [running]:
2024-08-25 20:24:02.952	github.com/hashicorp/nomad/client/allocrunner/taskrunner.(*DriverHandle).Exec(0x0, 0x2540be400, {0xc000af5780, 0x7}, {0xc0049166e0, 0x2, 0xc004554960?})
2024-08-25 20:24:02.952	github.com/hashicorp/nomad/client/allocrunner/taskrunner/driver_handle.go:70 +0xf7
2024-08-25 20:24:02.953	github.com/hashicorp/nomad/client/allocrunner/taskrunner/template.(*TaskTemplateManager).processScript(0xc0012bf2c0, 0xc0047fc2c0, 0xc001186000?)
2024-08-25 20:24:02.953	github.com/hashicorp/nomad/client/allocrunner/taskrunner/template/template.go:595 +0x99
2024-08-25 20:24:02.953	created by github.com/hashicorp/nomad/client/allocrunner/taskrunner/template.(*TaskTemplateManager).handleChangeModeScript in goroutine 2161
2024-08-25 20:24:02.953	github.com/hashicorp/nomad/client/allocrunner/taskrunner/template/template.go:563 +0x47
```

If I'm reading that stacktrace correctly, it happens here:
https://github.com/hashicorp/nomad/blob/bc90bd7c68fc8b813e624b0d1f89595e7a938abc/client/allocrunner/taskrunner/driver_handle.go#L70

Both memory accesses in that line are relative to `h`, i.e. the `*nomad.client.allocrunner.taskrunner.DriverHandle` the method is bound to. The stack trace indicates that `h` is `nil`.

The call to `.Exec` is here (last line):
https://github.com/hashicorp/nomad/blob/bc90bd7c68fc8b813e624b0d1f89595e7a938abc/client/allocrunner/taskrunner/template/template.go#L586-L595

[...]

PR #23917, opened 2024-09-04
template: lock task handle before trying script check

tgross:
In #23663 we fixed the template hook so that `change_mode="script"` didn't lose track of the task handle during restores. But this revealed a second bug which is that access to the handle is not locked while in use, which can allow it to be removed concurrently.

Fixes: https://github.com/hashicorp/nomad/issues/23875


client/allocrunner/taskrunner/template/template_test.go
```diff
@@ -1219,6 +1220,29 @@ OUTER:
...
+			if strings.Contains(
+				ev.DisplayMessage, "task driver handle is not available") {
+				break OUTER2
+			}
+		case <-harness.mockHooks.SignalCh:
+			t.Fatal(t, "signal not expected")
+		case <-timeout:
+			t.Fatal(t, "should have received an event that task driver handle is unavailable")
+		}
+	}
```
tgross, line 1245: Note for reviewers: unfortunately this test doesn't really exercise the bug that must have happened in https://github.com/hashicorp/nomad/issues/23875, because we'd have to concurrently make the change between the nil check in `processScript` and the subsequent `Exec` call, which is a very short race. Locally this test doesn't add anything meaningful to the time it takes the test to complete (always around ~3.2s on my machine), but I simultaneously hate to have a test that doesn't really test anything and a bugfix without a new test. I could go either way on removing it.

pkazmierczak (approved):
LGTM for the missing lock for sure. I say leave the test, but I'm not a 100% sure I understand how this bug happens.

PR #23922, opened 2024-09-05
[NET-10880] Keep a register of the usable cores to avoid using more than that

Juanadelacuesta:
As reported by support[ here](https://hashicorp.atlassian.net/browse/NET-10880) when a node is downscale to one with less number of cores, Nomad correctly fingerprints it, but the old allocations have the outdated information of the usable cores and might end up assigning more cores to the share and reserve cgroups than the ones present on the node., resulting on an error: 

```"Setup Failure" msg="failed to setup alloc: pre-run hook \"cpuparts_hook\" failed: write /sys/fs/cgroup/nomad.slice/share.slice/cpuset.cpus: numerical result out of range" failed=true```

And blocking the job from starting.

This PR adds a new field to the entity in charge of keeping track of the cgroups cpu distribution, that is populated at client start with the fresh information from the fingerprinting. This new `usableCores` field is used to filter out any cores that are not available anymore whenever an allocation cpu hook wants to update the cpu distribution.

### DIFF
--- a/client/lib/cgroupslib/partition_linux.go
+++ b/client/lib/cgroupslib/partition_linux.go
@@ -106,7 +106,7 @@ func (p *partition) write() error {
 
 	reserveStr := p.reserve.String()
 	if err := os.WriteFile(p.reservePath, []byte(reserveStr), 0644); err != nil {
-		return fmt.Errorf("cgroupslib: unable to update reserve cpuset with \"%s\": %w", reserveStr, err)
+		return fmt.Errorf("cgroupslib: unable to update reserve cpuset with %q: %w", reserveStr, err)
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: client/lib/cgroupslib/partition_linux.go [hunk 1]
```diff
@@ -49,50 +53,60 @@ func NewPartition(cores *idset.Set[hw.CoreID]) Partition {
...
 	share   *idset.Set[hw.CoreID]
 	reserve *idset.Set[hw.CoreID]
 }
 
 func (p *partition) Restore(cores *idset.Set[hw.CoreID]) {
+
 	p.lock.Lock()
 	defer p.lock.Unlock()
 
 	p.share.RemoveSet(cores)
```
tgross, line 71: Noting for other reviewers, because I had to check, that removing from the set is safe even if the cores don't exist in the set: https://go.dev/play/p/3iZeKQThSaV

tgross (approved):
LGTM!

Issue #9838, opened 2021-01-17
Template Stanza single quote and HCL2 cause parser error


idrennanvmware:
Nomad 1.0.2

### Issue
We make pretty extensive usage of file content rendering in our template stanza's but we have noticed that if the file contents have ```''``` in them then the HCL renderer gives us this kind of message

```
Invalid character; Single quotes are not valid. Use double quotes (") to enclose strings.
```

The problem is that if we then replace the ```''``` in the file with ```""``` we then get a different error because we also have ```}``` and that error is

```
Extra characters after interpolation expression; Expected a closing brace to end the interpolation expression, but found extra characters
```

Here's a snippet from the job file
```
template {
        data = <<EOH
        [[ fileContents ( print .service_dir "/journalbeat.yml" ) ]]
        EOH

        destination = "local/journalbeat.yml"
      }
```

and the respective file it's trying loading in (redacted)
```
############################# Journalbeat #####################################

---
journalbeat.inputs:
  - paths: []
    include_matches:
      - "systemd.unit=vault.service"
    seek: cursor
    fields_under_root: true
    fields:
      type: control-plane-vault
      token: ${LOGZIO_TOKEN:''}
      logzio_codec: plain
      ipAddress: ${IP_ADDRESS:''}
      hostName: ${HOSTNAME:''}
      clusterName: ${CLUSTER_NAME:''}
      deploymentType: ${DEPLOYMENT_TYPE:''}
```
and with the single quote replacement
```
############################# Journalbeat #####################################

---
journalbeat.inputs:
  - paths: []
    include_matches:
      - "systemd.unit=vault.service"
    seek: cursor
    fields_under_root: true
    fields:
      type: control-plane-vault
      token: ${LOGZIO_TOKEN:""}
      logzio_codec: plain
      ipAddress: ${IP_ADDRESS:""}
      hostName: ${HOSTNAME:""}
      clusterName: ${CLUSTER_NAME:""}
      deploymentType: ${DEPLOYMENT_TYPE:""}
```

Neither of which are ok in the HCL2 parser.  However, if we run this with the ```-hcl1``` flag for deployment then it all works as expected

tgross:
Thanks for reporting this @idrennanvmware!

notnoop:
Hi @idrennanvmware !  Thanks for raising this issue, and we'll look into how to address this behavior change.

[...]

PR #23923, opened 2024-09-05
hcl2: add fileescaped function

tgross:
When using HCL2 with the `file` function, the contents of the file are included directly in the job specification which means they may be interpolated if they include `${...}` or `%{...}` strings. This is typically a problem if you're using heredoc-style strings for `template.data`.

Add a `fileescaped` function that automatically escapes the contents in a way that's suitable for use in a heredoc. This will reduce the pain of migrating the some of the last remaining HCLv1 backwards incompatibilities.

Fixes: https://github.com/hashicorp/nomad/issues/9838

### DIFF
--- a/jobspec2/functions.go
+++ b/jobspec2/functions.go
@@ -5,6 +5,12 @@ package jobspec2
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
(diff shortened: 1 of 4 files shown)


tgross:
I did a little testing and there's actually no real need for this. See https://github.com/hashicorp/nomad/issues/9838#issuecomment-2332610513

Issue #23911, opened 2024-09-04
registry of default image for pause container is deprecated


tgross:
The [gcr.io container registry is shutting down](https://cloud.google.com/artifact-registry/docs/transition/prepare-gcr-shutdown) in March. we have a default [infra_image](https://developer.hashicorp.com/nomad/docs/drivers/docker#infra_image) in that registry for Docker's "pause containers". Those container images are authored by and will be hosted by the k8s project in their registry.k8s.io going forward. We should update our default to point to that registry.

Because this involves clients reaching out to potentially new domains and IPs, we should announce this as a breaking change in 1.9.0.

PR #23927, opened 2024-09-06
docker: update infra_image to new registry

tgross:
The gcr.io container registry is shutting down in March. Update the default `infra_image` for Docker's "pause" containers to point to the new location hosted by the k8s project.

Fixes: https://github.com/hashicorp/nomad/issues/23911
Ref: https://hashicorp.atlassian.net/browse/NET-10942

---

Question/note for reviewers: I said in #23911 that this is a breaking change so it should go in 1.9.0, but in March when the registry move happens our 1.8 LTS will still be in its support window (as will 1.7 until ~May or so, assuming that's when 1.10 goes out). Anyone on these older versions will have new allocations fail unless they update their pause container. Maybe we should pre-announce backporting this to 1.8 and 1.7 and then do so after 1.9 goes GA to give folks some time to deal with it if they have restrictive firewall environments?

### DIFF
--- a/website/content/docs/upgrade/upgrade-specific.mdx
+++ b/website/content/docs/upgrade/upgrade-specific.mdx
@@ -21,6 +21,13 @@ Nomad 1.9.0 no longer supports the HCLv1 format for job specifications. Using
 the `-hcl1` option for the `job run`, `job plan`, and `job validate` commands
 will no longer work.
 
+#### Default `infra_image` changed
+
+Due to the deprecation of the third-party `gcr.io` registry, the default Docker
+[`infra_image`][] is now `registry.k8s.io/pause-<arch>:3.3`. If you do not
+override the default, clients using the `docker` driver will make outbound
+requests to the new registry.
+
 #### Dropped support for older clients
 
 Nomad 1.9.0 removes support for Nomad client agents older than 1.6.0. Older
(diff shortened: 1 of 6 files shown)


website/content/docs/upgrade/upgrade-specific.mdx
```diff
@@ -21,6 +21,13 @@ Nomad 1.9.0 no longer supports the HCLv1 format for job specifications. Using
 the `-hcl1` option for the `job run`, `job plan`, and `job validate` commands
 will no longer work.
 
+#### Default `infra_image` changed
```
schmichael, line 24: Let's make it really clear this only impacts Docker so people who don't use Docker can breath a sigh of release and not read any further.

```suggestion
#### Default Docker `infra_image` changed
```

schmichael (approved):
LGTM. I don't think preannouncing is necessary, and we should get this into backports ASAP. There's often a weeks-to-months long lag between when a user selects a target version to upgrade their cluster to before they actually roll that version out to every environment. I'd hate for them to select the announcement version only to discover mid-rollout they need to upgrade again or manually update their `infra_image` on every client.

If users upgrade before reading upgrade notes, then preannouncing doesn't help anyway. I vote we rip the bandaid.

PR #23942, opened 2024-09-11
quotas: correct addComputedAllocAttrs

pkazmierczak:
Quota usage calculation depends on `allocation.Resources` field (which will be
deprecated in the future), while device resources are being kept in
`allocation.AllocatedResources` and parsed into a structure (vendor/type/name)
in order for the ranking in the scheduler to find nodes that can satisfy device
requirements. To make device quotas work properly, this has to be temporarily
translated into `allocation.Resources.Devices`.

### DIFF
--- a/nomad/state/state_store_test.go
+++ b/nomad/state/state_store_test.go
@@ -572,9 +572,8 @@ func TestStateStore_UpsertPlanResults_AllocationResources(t *testing.T) {
 
 	must.NoError(t, state.UpsertPlanResults(structs.MsgTypeTestSetup, 1000, &res))
 
-	ws := memdb.NewWatchSet()
-	out, err := state.AllocByID(ws, alloc.ID)
-	must.Nil(t, err)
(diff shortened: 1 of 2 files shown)


nomad/state/state_store_test.go
```diff
@@ -535,6 +535,51 @@ func TestStateStore_UpsertPlanResults_DeploymentUpdates(t *testing.T) {
...
+			Job:   job,
+		},
+		EvalID: eval.ID,
+	}
+
+	must.NoError(t, state.UpsertPlanResults(structs.MsgTypeTestSetup, 1000, &res))
+
+	ws := memdb.NewWatchSet()
+	out, err := state.AllocByID(ws, alloc.ID)
+	must.Nil(t, err)
```
tgross, line 577: For testing this we can skip the watch set:

```suggestion
	out, err := state.AllocByID(nil, alloc.ID)
	must.NoError(t, err)
```

tgross (approved):
LGTM!

Issue #23665, opened 2024-07-22
Store keyring in Raft


schmichael:
### Background

With the introduction of Variables in Nomad 1.4, Nomad has had an internal keyring consisting of:
1. A Data Encryption Key (DEK) for encrypting Variables in Raft and on disk
2. A Key Encryption Key (KEK) for encrypting the DEK
3. A Workload Identity signing key (see #18882 for details)

The KEK is stored on disk distinctly from Raft logs and snapshots so that secret material may not be extracted from leaked snapshots.

As of #23580 Nomad will support using an external KMS for the KEK. However the _wrapped_ (encrypted) DEK and WI keys will still be in the on disk keyring outside of snapshots.

The on disk keyring, even when using a KMS, poses an unfortunate challenge for Nomad operators: they must backup the keyring _like_ snapshots, but distinctly. If a single script backs up both snapshots and keyring to the same location, the only remaining benefit to keeping the two distinct is in the case of sharing snapshots with a third party (eg HashiCorp) for debugging.

Nomad engineering has encountered multiple outages caused by users unaware that they need to backup and restore the keyring distinct but alongside the snapshot.

### Proposal

The downsides of storing the keyring separate from snapshots outweighs the potential security benefits. The security benefits are likely rarely realized due to the difficulty of splitting backups across multiple processes and locations. This user experience should be fixed by doing the following:

1. By default, without a KMS configured, Nomad should store the _unencrypted_ KEK and other wrapped keys in Raft and snapshots. Server agents should log a warning instructing users their snapshots are functionally unencrypted and using a KMS is desirable (or linking to such docs).
2. When using a KMS, _wrapped_ key material should be stored in Raft and snapshots.
3. The existing behavior of storing the KEK on disk distinctly from Raft and snapshots should be available for users who do not want to rely on an external KMS but also do not want to risk exposing their keys in Raft or snapshots.

In order to support storing key material separate from snapshots, the following features should be added:

1. The [Snapshot API](https://developer.hashicorp.com/nomad/api-docs/operator/snapshot#generate-snapshot) and [`operator snapshot save` CLI](https://developer.hashicorp.com/nomad/docs/commands/operator/snapshot/save) should gain a parameter that allows generating snapshots *without* the KEK included (for users _without_ a KMS). Users with a KMS configured will never receive their KEK in a snapshot.
[...]

PR #23943, opened 2024-09-11
refactor keyring state store methods

tgross:
For #23665 I'm about to make add a lot more code to the state store for the keyring, so I'd like to pull these out to their own file. Also updates the test to use `shoenig/test` and changes the name of one method to be a little more accurate.

Note for reviewers:
* I'm backporting this to 1.8.x because it's the LTS but not further than that because with KMS support in 1.8.x the code has changed quite a bit from 1.6.x and 1.7.x.
* I'd also like to refactor `nomad/encrypter.go` into its own package but right now that's got a _lot_ of dependencies on server RPC code because of keyring replication and server lifecycle. So I'm punting on that here. Once I've got #23665 ready to ship, I might revisit that decision but I don't want to push up a refactoring PR only to have to push up another one unwinding half of it once I'm done.


nomad/state/state_store_keyring.go
```diff
@@ -0,0 +1,196 @@
+// Copyright (c) HashiCorp, Inc.
```
pkazmierczak, line 1: bless your heart for moving this to a separate file 🙇‍♂️

nomad/state/state_store_keyring.go
```diff
@@ -0,0 +1,196 @@
...
+					modified = true
+				}
+			case structs.RootKeyStateActive:
+				if rekey {
+					key = key.MakeRekeying()
+				} else {
+					key = key.MakeInactive()
+				}
+				modified = true
+			case structs.RootKeyStateRekeying, structs.RootKeyStateDeprecated:
```
pkazmierczak, line 69: do we have to explicitly case this? why not just skip this line?

pkazmierczak (approved):
LGTM!

nomad/state/state_store_keyring.go
tgross: It absolutely could be `default:` but when you've got a small enum like this I feel like it helps understandability to explicitly list them. If you add a state later it's more obvious that it's missing, rather than simply falling thru the default case and doing nothing.

PR #23946, opened 2024-09-12
Stop referring non existent vault.file attribute

seriousben:
The documentation is referring to a `file` attribute that does not exist on the `vault` block.

This PR changes those references to mention the `disable_file` attribute instead.


schmichael (approved):
Good catch! Thanks!

Issue #15487, opened 2022-12-07
support `SO_REUSEPORT` for safe reuse of static ports


victorstewart:
on [https://developer.hashicorp.com/nomad/docs/job-specification/network#host_network](https://developer.hashicorp.com/nomad/docs/job-specification/network#host_network)

it states... 

> Since multiple services cannot share a port, the port must be open in order to place your task.

but this is false: if using host networking, and SO_REUSEPORT, ports can be shared. docker supports this, i assume the containerd plugin as well.

just wanted to confirm Nomad doesn't refuse to start jobs that all use the same static port.

victorstewart:
okay no worries. between now and then i ended up writing my own machine orchestrator, program scheduler and container runtime... just ended up being the path of least resistance for me for many reasons. so i ended up completely sidestepping this issue.

but i do think this is an important feature.

take an array of QUIC programs that use a load balancer to assist connection establishment, then switch to their unique unicast address. all of these servers must run on the public (read: external facing) 443 port for most client firewalls, at whatever waypoint, to reliably allow the UDP traffic through. but this pattern would be impossible in Nomad today.

(not everyone load balances or proxies all traffic through CNI meshes).

best of luck!

gulducat:
Howdy folks!  I have a draft PR #23956 up for this, and I'd love to hear any feedback about whether it would or would not improve the situation for any particular use cases.

Lots of detail in the PR description, but in short: how would being able to specify a port, but ignore collisions, affect your deployments?

```hcl
group "g" {
    network {
        mode = "host"

        port "http" {
            static = 8000
            to     = 8000

            allow_reuse = true # ignore collisions
        }
    }
    task "t" {
        ...
```

PR #23956, opened 2024-09-13
networking: option to allow static port reuse

gulducat:
Howdy folks, draft PR here to solicit feedback.  Long problem/solution descriptions first, then questions (that I can think of) at the bottom.  (note: the bulk of the changeset are test changes, the actual feature is in 126abd84e5c6f069b897aa09e209a385227475d4)

## Problem

The goal is to address #15487 -- Currently, static ports may be used only once on a single address on a network interface.

This is good, almost all of the time.  Most programs bind exclusively to an addr:port, so other copies of the same program (or other programs) cannot multi-bind to it.  The Nomad scheduler prevents placing allocations that would cause such port collisions to avoid a predictable program error at runtime.

_Some_ programs, however, _do_ allow addr:port reuse, via the [SO_REUSEPORT](https://lwn.net/Articles/542629/) socket option (on linux).  For example, this little toy web server:

<details><summary>reuseport.py</summary>

```python
#!/usr/bin/env python3

import http.server
import socket
import socketserver
from http import HTTPStatus
from os import environ

ADDR = environ.get('ADDR', '')
PORT = int(environ.get('PORT', 8000))
MESSAGE = environ.get('MESSAGE', 'hello')


class Handler(http.server.BaseHTTPRequestHandler):
    def do_HEAD(self):
        self.head()
    def do_GET(self):
        self.head()
        self.wfile.write(bytes(f'{MESSAGE}!\n', 'utf-8'))
    def head(self):
        self.send_response(HTTPStatus.OK)
        self.send_header('Content-Type', 'text/html')
        self.end_headers()


class ReusableServer(socketserver.TCPServer):
    def server_bind(self):
        # socket.SO_REUSEPORT is the magic ingredient.
        self.socket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
        super().server_bind()


with ReusableServer((ADDR, PORT), Handler) as httpd:
    print(f"serving at {ADDR}:{PORT}", flush=True)
    httpd.serve_forever()
```
</details>

A more practical example is [traefik](https://doc.traefik.io/traefik/routing/entrypoints/#reuseport), which also (optionally) supports it, and while I haven't taken a full inventory, I suspect other proxies do, too.

Nomad's `network.port` prevents this use case.  To work around it, one may merely neglect to specify the `port`, and then still bind to it anyway, but this completely removes the guardrails that prevent other allocs from being mistakenly placed with no hope of success.

## Solution

This PR adds a new `allow_reuse` field to `job.group.network.port` (only this scope, i.e. not task networks):

```hcl
job "web" {
    group "g" {
        count = 2
        network {
            mode = "host" # the default, and required for this feature
            port "http" {
                static = 8000
                to     = 8000

                allow_reuse = true
            }
        }
        task "t" {
            driver = "docker"
            config {
                ...
                network_mode = "host" # required for docker or podman drivers
            ...
```

<details><summary>full example: reuseport.nomad.hcl</summary>

(uses `reuseport.py` from above)

```hcl
job "reuseport" {
  group "docker" {
    count = 2
    network {
      mode = "host"
      port "http" {
        static = 8000
        to     = 8000

        allow_reuse = true
      }
    }
    task "t" {
      driver = "docker"
      config {
        image   = "python:slim"
        command = "python3"
        args    = ["/local/reuseport.py"]

        network_mode = "host"
      }
      template {
        destination = "local/reuseport.py"
        data        = file("reuseport.py")
      }
      env {
        MESSAGE = "hello, ${NOMAD_ALLOC_ID}"
      }
    }
  }
}

```
</details>

A job run like that can have multiple copies placed on the same host, then another job that requests the same static port without the new option will be placed elsewhere (or block pending the port becoming available).

This just gets the scheduler out of the way.  It's up to the user (job author and/or cluster operator) to ensure placement on a node that does not already have some other thing listening on the port.

For example, I can imagine a client config like this:

```hcl
client {
    node_pool = "web-proxy"
    reserved {
        reserved_ports = "8000"
    }
    ...
```

[`reserved_ports`](https://developer.hashicorp.com/nomad/docs/configuration/client#reserved_ports) would prevent placement of any group that does not allow port reuse, ahead of the desired job, then `node_pool` (or some other constraint method) ensures the desired job gets placed on such a node.

I do wonder, though, if `port.allow_reuse` actually adds anything beyond merely excluding the port, if `reserved_ports` already prevents collisions from other jobs...  Perhaps a `system` job that runs everywhere, all the time, so could grab the port without needing to reserve it in client config?  I am a little hazy on the real value-add here.

Security-wise, this does not increase risk of some other program trying to squat on ports any more than a host-network'd job currently can, if it simply does not declare that it will use a `port` at all.

## Questions:

1. What specific use cases _are_ solved by this new option?
2. What specific use cases _are not_ solved by it?
3. Any UX or security concerns I haven't considered?
4. Instead of `allow_reuse`, perhaps `ignore_collisions` is more precise? (and feels like a good warning)


tgross:
> I do wonder, though, if port.allow_reuse actually adds anything beyond merely excluding the port, if reserved_ports already prevents collisions from other jobs.

There are a couple of advantages:
* Using `reserved_ports` and no `network.port` doesn't allow the job author to register the port by name as a `service`.
* The `reserved_ports` configuration is owned by the cluster admin and not the job author, and cluster admins will usually want to get out of the way of job authors here (we expect `reserved_ports` is going to be for services the cluster admin is running outside of Nomad, like `ssh`)

nomad/structs/network.go
```diff
@@ -430,15 +431,19 @@ func (idx *NetworkIndex) AddReserved(n *NetworkResource) (collide bool, reasons
...
 		if used.Check(uint(port.Value)) {
-			collide = true
-			reason := fmt.Sprintf("port %d already in use", port.Value)
-			reasons = append(reasons, reason)
+			// TODO: yes. 2.
+			if !port.AllowReuse {
+				collide = true
+				reason := fmt.Sprintf("port %d already in use", port.Value)
+				reasons = append(reasons, reason)
+			}
```
tgross, line 444: I think we might need to call `used.Set` here for when we're ranking a node and adding that node's existing allocations to the `NetworkIndex`. If those allocations have `port.allow_reuse` we want them to collide with the new allocation if that allocation doesn't also have `port.allow_reuse` for that port.

(This will probably pop out once you've got test coverage in the scheduler code.)

gulducat: We only get into this condition `if used.Check()` (line 438 just above here), i.e. if `used.Set()` (line 447) has already been called by some other prior entry here.

the full context: https://github.com/hashicorp/nomad/blob/126abd84e5c6f069b897aa09e209a385227475d4/nomad/structs/network.go#L437-L448

(those "todo: yes. #." comments are the order in which I encountered collisions while stepping through breakpoints.)

In other words, `Set()` always happens if it hasn't already, regardless of `AllowReuse`, then subsequent `Check()`s just may or may not report collisions.

nomad/structs/structs.go
```diff
@@ -2787,18 +2787,20 @@ type NodeNetworkAddress struct {
 }
 
 type AllocatedPortMapping struct {
-	Label  string
-	Value  int
-	To     int
-	HostIP string
+	Label           string
```
tgross, line 2790: This will probably fix the plan normalization test, but being even close to the threshold on that test right now suggests we're missing this on a bunch of other objects too.

```suggestion
	// msgpack omit empty fields during serialization
	_struct bool `codec:",omitempty"` // nolint: structcheck
	Label           string
```

(Similar for the `Port` type definition below)

website/content/docs/job-specification/network.mdx
```diff
@@ -105,6 +105,12 @@ All other operating systems use the `host` networking mode.
 - `host_network` `(string:nil)` - Designates the host network name to use when allocating
   the port. When port mapping the host port will only forward traffic to the matched host
   network address.
+- `ignore_collision` `(bool: false)` - Allows the group to be placed on a node
+  where the port may already be reserved. Intended for programs that support
```
tgross, line 109: Do we want to use the word "reserved" here? There's a specific client configuration option for "reserved ports" that may be confusing. Maybe "allocated" or even just "in use" (maybe the port isn't being used by a Nomad job at all)?

tgross (approved):
LGTM!

nomad/structs/structs.go
gulducat: I swear I tried this and it didn't help much, but now that I try again, it makes a significant impact (on the TestPlanNormalize test).

It'll mean that lots of inline `Port{"label", 0, 0, ""}`s in tests, which I already added `false` to, will need poking again.  I might just make all of those multi-line with explicit fields while I'm at it.

website/content/docs/job-specification/network.mdx
gulducat: I actually imagine that this may be used in conjunction with the [`reserved_ports` client config](https://developer.hashicorp.com/nomad/docs/configuration/client#reserved_ports), so a cluster operator may dedicate ports for applications like these, and run them with Nomad instead of systemd or such.

On the flip side, it also allows job authors to bypass operator intent...  I suppose Sentinel may need to come into play here to retain current guarantees. :thinking:

nomad/structs/structs.go
gulducat: I split this out to #23980 so it can be backported, and should result in this PR being a lot shorter without all the totally tangential test changes.

website/content/docs/job-specification/network.mdx
gulducat: I thought some more about bypassing operator intent around `reserved_ports`: a job author can already do that today, if they merely do not specify a `port{}`, so this does not make that situation any worse.  (I have had to remind myself of this more than once.)

tgross: Excellent point!

PR #23959, opened 2024-09-13
networking: set alloc NetworkStatus.AddressIPv6

gulducat:
#23882 added an IPv6 option for the Nomad `bridge` network mode, but an IPv6 addr created by it was not stored anywhere (aside from ip6tables on the node).

This puts it alongside the IPv4 address in an alloc's NetworkStatus, so it can be found with an API call if needed.

e.g.:

```
$ nomad alloc status -json 3dca | jq '.NetworkStatus'
{
  "Address": "172.26.64.14",
  "AddressIPv6": "fd00:a110:c8::b",
  "DNS": null,
  "InterfaceName": "eth0"
}
```

### DIFF
--- a/api/allocations.go
+++ b/api/allocations.go
@@ -435,6 +435,7 @@ type AllocDeploymentStatus struct {
 type AllocNetworkStatus struct {
 	InterfaceName string
 	Address       string
+	AddressIPv6   string
 	DNS           *DNSConfig
 }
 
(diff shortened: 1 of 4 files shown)


shoenig (approved):
LGTM!

tgross (approved):
:+1: LGTM

Issue #11011, opened 2021-08-06
Add user specified labels to Nomad Allocation Metrics 


kfenech1:
### Proposal
As a developer I would like to be able to add labels to the Nomad Telemetry Metrics.  I would like to add labels in the Job definition which are propagated to the allocations metrics generated by Nomad. This would allow a user to filter better when creating dashboards etc. Currently metrics are not being propagated to the Nomad Metrics telemetry when labels are added to to Task -> Config stanza. Having said that, when cAdvisor is installed as a container on Nomad-client, it manages to pick up those labels and adds them to the metrics generated by cAdvisor. Having the same for Nomad metrics would be very helpful.

### Use-cases
Having this functionality will allow users to perform better filtering when analysing metrics and create dashboards. 

### Attempted Solutions
Install cAdvisor and use metrics generated by cAdvisor instead of Nomad Metrics. This however introduces overhead and more network traffic since a new endpoint for each client (that essentials provides similar metrics to Nomad Metrics) will need to be scraped by your monitoring system.

jrasell:
Hi @kfenech1 and thanks for raising this issue. It is certainly something we would be interested in supporting in the future.

chrismikehogan:
Just wanted to add another use we have that would benefit: We run multiple jobs with different SLA tiers on the same nomad cluster (think bronze, silver, gold), each with different SLAs. 

If gold-tier jobs start queueing due to resource exhaustion in middle of night, we'd like to get alerted. If bronze-tier jobs are queueing however (they have lower priority) we don't want to get woken up. 

Having ability to add custom labels would allow us to do this!

tgross:
Closed by https://github.com/hashicorp/nomad/pull/23964

PR #23964, opened 2024-09-15
metrics: introduce client config to include meta as part of the base label

mvegter:
Took a stab at the following tickets / discussions :
- #11011
- https://discuss.hashicorp.com/t/add-additional-labels-to-nomad-allocs-telemetry/12613

First time opening a pull request using Go , feedback is much appreciated !


client/allocrunner/taskrunner/task_runner_test.go
```diff
@@ -2866,6 +2866,75 @@ func TestTaskRunner_BaseLabels(t *testing.T) {
 	require.Equal(alloc.Namespace, labels["namespace"])
 }
 
+// TestTaskRunner_BaseLabels_IncludesMetadata tests that the base labels include
+// the metadata fields
+func TestTaskRunner_BaseLabels_IncludesMetadata(t *testing.T) {
+	ci.Parallel(t)
+	require := require.New(t)
```
tgross, line 2873: We're slowly migrating all uses of `testify` to `shoenig/test` instead:

```suggestion
```

client/allocrunner/taskrunner/task_runner_test.go
```diff
@@ -2866,6 +2866,75 @@ func TestTaskRunner_BaseLabels(t *testing.T) {
...
+		"command": "whoami",
+	}
+
+	config, cleanup := testTaskRunnerConfig(t, alloc, task.Name, nil)
+	defer cleanup()
+
+	config.ClientConfig.IncludeMetadataInMetrics = true
+
+	tr, err := NewTaskRunner(config)
+	require.NoError(err)
```
tgross, line 2891:
```suggestion
	must.NoError(err)
```

client/allocrunner/taskrunner/task_runner_test.go
```diff
@@ -2866,6 +2866,75 @@ func TestTaskRunner_BaseLabels(t *testing.T) {
...
+
+	labels := map[string]string{}
+	for _, e := range tr.baseLabels {
+		labels[e.Name] = e.Value
+	}
+
+	require.Equal(task.Meta["owner"], labels["owner"])
+	require.Equal(task.Meta["some_dynamic_value"], labels["some_dynamic_value"])
+	require.Equal(task.Meta["my-key"], labels["my_key"])
```
tgross, line 2900: More `shoenig/test` migration. But also, these test assertions could be constants so that you know you're not introducing test dependency on the internals:

```suggestion
	must.Eq(t, "HashiCorp", labels["owner"])
	must.Eq(t, "now()", labels["some_dynamic_value"])
	must.Eq(t, "my-value", labels["my_key"])
```

Alternately, you could skip iterating the `baseLabels` and just `must.Eq` the whole map.

client/allocrunner/taskrunner/task_runner_test.go
```diff
@@ -2866,6 +2866,75 @@ func TestTaskRunner_BaseLabels(t *testing.T) {
...
+}
+
+// TestTaskRunner_BaseLabels_IncludesMetadata tests that the base labels include
+// the metadata fields using the provided allowed list of keys
+func TestTaskRunner_BaseLabels_IncludesMetadata_WithAllowList(t *testing.T) {
```
tgross, line 2905: I appreciate you being fine-grained with tests, but we could probably just squash these two tests together and save some test overhead. They aren't exercising anything all that independent.

client/allocrunner/taskrunner/task_runner_test.go
```diff
@@ -2866,6 +2866,75 @@ func TestTaskRunner_BaseLabels(t *testing.T) {
...
+	require.NoError(err)
+
+	labels := map[string]string{}
+	for _, e := range tr.baseLabels {
+		labels[e.Name] = e.Value
+	}
+
+	require.Equal(task.Meta["owner"], labels["owner"])
+	require.Equal(task.Meta["my-key"], labels["my_key"])
+	require.Empty(labels["some_dynamic_value"])
```
tgross, line 2935: Same here as the previous test.

client/config/config.go
```diff
@@ -203,6 +203,14 @@ type Config struct {
 	// allocation metrics to remote Telemetry sinks
 	PublishAllocationMetrics bool
 
+	// IncludeMetadataInMetrics determines whether nomad should include the
+	// metadata as labels in the metrics to remote Telemetry sinks
```
tgross, line 207: We should be really specific here and call out this is job/allocation metadata, as opposed to client metadata. But that makes me think we should probably make this configuration `IncludeAllocMetadataInMetrics` so that we have the option to add client metadata later if we want.

client/allocrunner/taskrunner/task_runner.go
```diff
@@ -510,6 +510,19 @@ func (tr *TaskRunner) initLabels() {
...
 
+	if tr.clientConfig.IncludeMetadataInMetrics {
+		for meta := range tr.task.Meta {
+			if len(tr.clientConfig.AllowedMetadataKeysInMetrics) > 0 && !slices.Contains(tr.clientConfig.AllowedMetadataKeysInMetrics, meta) {
+				continue
+			}
+
+			tr.baseLabels = append(tr.baseLabels, metrics.Label{
+				Name:  strings.ReplaceAll(meta, "-", "_"),
+				Value: tr.alloc.Job.Meta[meta],
```
tgross, line 521: If we do `for meta, metaValue := range tr.task.Meta` above, we'd have the value already I think?

tgross (changes requested):
Hi @mvegter! This is looking pretty good. I've left some comments. Additionally, can you:
* run `make cl` to add a changelog entry
* add documentation for this new config value to `website/content/docs/configuration/client.mdx`. This should warn about the potential impact to cardinality.

Thanks for this PR!

client/allocrunner/taskrunner/task_runner_test.go
mvegter: I have opted for `must.Eq` as I didn't want to test the base labels like `alloc_id` and others as part of this

mvegter:
Thanks for the feedback @tgross  , I have updated the `telemetry.mdx` instead of the `client.mdx` as there was no mention of metrics nor telemetry on that page.

command/agent/config.go
```diff
@@ -2524,6 +2528,12 @@ func (t *Telemetry) Merge(b *Telemetry) *Telemetry {
 	if b.PublishAllocationMetrics {
 		result.PublishAllocationMetrics = true
 	}
+	if b.IncludeAllocMetadataInMetrics {
+		result.IncludeAllocMetadataInMetrics = true
+	}
+	if b.AllowedMetadataKeysInMetrics != nil {
+		result.AllowedMetadataKeysInMetrics = b.AllowedMetadataKeysInMetrics
+	}
```
tgross, line 2536: Sorry I missed this before, but I think we want to actually merge the two slices for this configuration. There aren't a lot of examples of this in the rest of the config, but where there are this is typically what we do:

```suggestion
	result.AllowedMetadataKeysInMetrics = append(result.AllowedMetadataKeysInMetrics, b.AllowedMetadataKeysInMetrics...)
```

(And yes, dropping the nil check is safe here, see https://go.dev/play/p/-N0RFJRElPC for an example of why)

website/content/docs/configuration/telemetry.mdx
```diff
@@ -58,6 +58,15 @@ The following options are available on all telemetry configurations.
...
   publish runtime metrics of allocations.
 
+- `include_alloc_metadata_in_metrics` `(bool: false)` - This controls whether
+  allocation metadata should be part of the metric labels. Defaults to false
+  as enabling this exposes the risk of high cardinality labels. It is therefore
+  recommended to configure [allowed_metadata_keys_in_metrics](#allowed_metadata_keys_in_metrics).
+
+- `allowed_metadata_keys_in_metrics` `(list: [])` - This filters the metadata
+  keys to be included in the metric publishing. By default it will not filter
+  out any keys and thus include all metadata.
```
tgross, line 68:
```suggestion
- `include_alloc_metadata_in_metrics` `(bool: false)` - This controls whether
  allocation metadata is included in metric labels. Enabling this option may result in 
  high cardinality labels. You should also configure [allowed_metadata_keys_in_metrics](#allowed_metadata_keys_in_metrics).

- `allowed_metadata_keys_in_metrics` `(list: [])` - This filters the metadata
  keys to be included in the metric publishing. By default it will not filter
  out any keys and thus include all metadata.
```

@aimeeu want to weigh in on writing style here beyond my suggestion?

tgross (changes requested):
Looking great @mvegter! Just a few more small things and this should be good-to-go.

website/content/docs/configuration/telemetry.mdx
mvegter: Accepted your suggestion, leaving this comment open pending @aimeeu feedback

website/content/docs/configuration/telemetry.mdx
```diff
@@ -58,6 +58,14 @@ The following options are available on all telemetry configurations.
 - `publish_allocation_metrics` `(bool: false)` - Specifies if Nomad should
   publish runtime metrics of allocations.
 
+- `include_alloc_metadata_in_metrics` `(bool: false)` - This controls whether
+  allocation metadata is included in metric labels. Enabling this option may result in
+  high cardinality labels. You should also configure [allowed_metadata_keys_in_metrics](#allowed_metadata_keys_in_metrics).
+
+- `allowed_metadata_keys_in_metrics` `(list: [])` - This filters the metadata
+  keys to be included in the metric publishing. By default it will not filter
```
aimeeu, line 66:
```suggestion
  keys to be included in the metric publishing. By default it does not filter
```

aimeeu (approved):
Thanks @mvegter for the docs update as part of your code PR! Thanks also Tim for reviewing. I made a minor suggestion so the docs content conforms to our style guide rule to use present tense.

tgross (approved):
LGTM!

website/content/docs/configuration/telemetry.mdx
tgross: Thanks @aimeeu! I've accepted the suggestion on @mvegter's behalf so we can get this merged.

mvegter:
I believe my push with @aimeeu suggestion just arrived before @tgross approval , so I think all should be included now . Thanks both !

tgross:
Ok, this has been merged and will ship in Nomad 1.9.0. Thanks for the contribution @mvegter!

PR #23966, opened 2024-09-16
docker: use official client instead of fsouza/go-dockerclient

pkazmierczak:


drivers/docker/utils.go
```diff
@@ -13,16 +13,25 @@ import (
...
 	"github.com/docker/cli/cli/config/types"
-	"github.com/docker/distribution/reference"
 	registrytypes "github.com/docker/docker/api/types/registry"
 	"github.com/docker/docker/registry"
-	docker "github.com/fsouza/go-dockerclient"
 )
 
 func parseDockerImage(image string) (repo, tag string) {
-	repo, tag = docker.ParseRepositoryTag(image)
+	// deode the image tag
```
tgross, line 24:
```suggestion
	// decode the image tag
```

Also, oof, how is there not a `ParseImage` method with a structured output not already in [`github.com/distribution/reference`](https://pkg.go.dev/github.com/distribution/reference)? :crying_cat_face: I wonder if we could use the regex in that package to do this in a way that catches any edge cases in the surprisingly large grammar?

drivers/docker/driver.go
```diff
@@ -372,29 +387,34 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
...
+	d.logger.Info("created container", "container_id", createdContainer.ID)
 
 	// We don't need to start the container if the container is already running
 	// since we don't create containers which are already present on the host
 	// and are running
+	container, err := dockerClient.ContainerInspect(d.ctx, createdContainer.ID)
+	if err != nil {
+		d.logger.Error("failed to inspect created container", "error", err)
+		return nil, nil, nstructs.WrapRecoverable(fmt.Sprintf("failed to create container: %v", err), err)
+	}
```
tgross, line 411: Don't we already get a container with a valid `ID` and `State` fields (I think those are the only two we query?) from the `createContainer` method? What does re-querying the API get us? And if it _is_ necessary, should we push it down into the `createContainer` method?

drivers/docker/driver.go
```diff
@@ -1424,25 +1431,39 @@ func (d *Driver) createContainerConfig(task *drivers.TaskConfig, driverConfig *T
 
 	if driverConfig.MacAddress != "" {
 		config.MacAddress = driverConfig.MacAddress
+
+		// newer docker versions obsoleve the config.MacAddress field
```
tgross, line 1435: I think?

```suggestion
		// newer docker versions obsolete the config.MacAddress field
```

drivers/docker/driver.go
```diff
@@ -489,24 +501,27 @@ CREATE:
...
-func (d *Driver) createContainer(client createContainerClient, config docker.CreateContainerOptions,
-	image string) (*docker.Container, error) {
+func (d *Driver) createContainer(client createContainerClient, config createContainerOptions, image string) (*types.ContainerJSON, error) {
 	// Create a container
 	var attempted uint64
 	var backoff time.Duration
 
 CREATE:
-	container, createErr := client.CreateContainer(config)
+	_, createErr := client.ContainerCreate(d.ctx, config.Config, config.Host, config.Networking, nil, config.Name)
```
tgross, line 518: Looks like the missing arg here is [`ocispec.Platform`](https://pkg.go.dev/github.com/opencontainers/image-spec/specs-go/v1#Platform). Do we know if that's at all useful for Docker?

drivers/docker/driver_unix_test.go
```diff
@@ -183,22 +186,24 @@ func TestDockerDriver_Sysctl_Ulimit(t *testing.T) {
...
 
 	client, d, handle, cleanup := dockerSetup(t, task, nil)
 	defer cleanup()
-	require.NoError(t, d.WaitUntilStarted(task.ID, 5*time.Second))
+	must.NoError(t, d.WaitUntilStarted(task.ID, 5*time.Second))
 
-	container, err := client.InspectContainer(handle.containerID)
-	assert.Nil(t, err, "unexpected error: %v", err)
+	container, err := client.ContainerInspect(context.Background(), handle.containerID)
+	must.Nil(t, err, must.Sprintf("unexpected error: %v", err))
```
tgross, line 196:
```suggestion
	must.NoError(t, err)
```

drivers/docker/driver_unix_test.go
```diff
@@ -246,16 +253,16 @@ func TestDockerDriver_Sysctl_Ulimit_Errors(t *testing.T) {
...
 		d := dockerDriverHarness(t, nil)
 		cleanup := d.MkAllocDir(task, true)
 		t.Cleanup(cleanup)
 		copyImage(t, task.TaskDir(), "busybox.tar")
 
 		_, _, err := d.StartTask(task)
-		require.NotNil(t, err, "Expected non nil error")
-		require.Contains(t, err.Error(), tc.err.Error())
+		must.NotNil(t, err, must.Sprint("Expected non nil error"))
+		must.StrContains(t, err.Error(), tc.err.Error())
```
tgross, line 265: You can swap this out for:

```suggestion
		must.ErrorContains(t, err, tc.err.Error())
```

drivers/docker/stats.go
```diff
@@ -141,13 +136,20 @@ func (h *taskHandle) collectStats(ctx context.Context, destCh *usageSender, inte
 			retry++
 			continue
 		}
+		defer statsReader.Body.Close()
+
+		var stats containerapi.Stats
+		binary.Read(statsReader.Body, binary.LittleEndian, &stats)
```
tgross, line 142: Should we catch decoding errors here (but not things like `io.EOF`)? I feel like if we collected garbage we probably don't want to send that upstream to throw errors because of empty fields in the decoded object.

tgross (commented):
This is looking great @pkazmierczak. I've left a few suggestions but once we've got those resolved I think this will be good-to-go. Nice work!

drivers/docker/driver.go
pkazmierczak: oof, it's a snafu that happened because initial refactoring had the `createContainer` return the wrong type. I fixed in [55f9670](https://github.com/hashicorp/nomad/pull/23966/commits/55f96706baed8ea17e27f9843a0f6c69ef6bd9d4)

drivers/docker/driver.go
pkazmierczak: We don't use this parameter at the moment in Nomad, but it could be useful in the future. It allows specifying platform of created containers.

drivers/docker/docklog/docker_logger_test.go
```diff
@@ -33,48 +38,38 @@ func testContainerDetails() (image string, imageName string, imageTag string) {
...
 		},
-		Context: context.Background(),
-	}
+		Image: containerImage,
+	}, nil, nil, nil, "")
+	must.NoError(t, err)
 
-	container, err := client.CreateContainer(containerConf)
-	require.NoError(err)
+	defer client.ContainerRemove(ctx, container.ID, containerapi.RemoveOptions{Force: true})
```
shoenig, line 66: I think nowadays we'd use `t.Cleanup` since it runs even if the test fails out

drivers/docker/driver.go
```diff
@@ -375,26 +390,22 @@ CREATE:
...
 		d.logger.Error("failed to create container", "error", err)
-		dockerClient.RemoveContainer(docker.RemoveContainerOptions{
-			ID:    containerCfg.Name,
-			Force: true,
-		})
+		if container != nil {
+			err := dockerClient.ContainerRemove(d.ctx, container.ID, containerapi.RemoveOptions{Force: true})
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to remove container %s: %v", container.ID, err)
+			}
```
shoenig, line 397: try to avoid the shadow'd `err` name

drivers/docker/driver.go
```diff
@@ -1424,25 +1422,39 @@ func (d *Driver) createContainerConfig(task *drivers.TaskConfig, driverConfig *T
 
 	if driverConfig.MacAddress != "" {
 		config.MacAddress = driverConfig.MacAddress
+
+		// newer docker versions obsolete the config.MacAddress field
+		isNewEnough := semver.Compare(fmt.Sprintf("v%s", ver.APIVersion), "v1.44")
```
shoenig, line 1427: seems like this would be `isTooNew` (for mac address field)

drivers/docker/driver.go
```diff
@@ -1630,29 +1638,26 @@ func (d *Driver) DestroyTask(taskID string, force bool) error {
...
 	}
 
-	c, err := dockerClient.InspectContainerWithOptions(docker.InspectContainerOptions{
-		ID: h.containerID,
-	})
+	c, err := dockerClient.ContainerInspect(d.ctx, h.containerID)
 	if err != nil {
-		switch err.(type) {
-		case *docker.NoSuchContainer:
+		if strings.Contains(err.Error(), NoSuchContainerError) {
```
shoenig, line 1643: wouldn't `errors.Is` work here?

drivers/docker/stats.go
pkazmierczak: Yes. I actually ended up revamping this because that implementation was buggy. See [ee27028](https://github.com/hashicorp/nomad/pull/23966/commits/ee2702837362adf4ec9a222e5cf9cfe6d30888d2) for improvements.

drivers/docker/driver_test.go
```diff
@@ -2184,67 +2181,65 @@ func TestDockerDriver_MissingContainer_Cleanup(t *testing.T) {
...
-			t.Fatalf("should err: %v", res)
-		}
-	case <-time.After(time.Duration(tu.TestMultiplier()*10) * time.Second):
-		t.Fatalf("timeout")
-	}
-}
+	must.False(t, ok)
+}
+
+// func TestDockerDriver_Stats(t *testing.T) {
```
shoenig, line 2204: Should this still be commented out?

drivers/docker/stats.go
```diff
@@ -141,13 +136,22 @@ func (h *taskHandle) collectStats(ctx context.Context, destCh *usageSender, inte
 			retry++
 			continue
 		}
+		defer statsReader.Body.Close()
+
+		var stats containerapi.Stats
+		if err := binary.Read(statsReader.Body, binary.LittleEndian, &stats); err != nil {
+			h.logger.Error("error decoding stats data for container", "error", err)
+		}
```
shoenig: is there a reference for where this came from? manually reading little endian seems out of left field

drivers/docker/utils.go
```diff
@@ -13,16 +13,25 @@ import (
...
+	"github.com/distribution/reference"
 	"github.com/docker/cli/cli/config/configfile"
 	"github.com/docker/cli/cli/config/types"
-	"github.com/docker/distribution/reference"
 	registrytypes "github.com/docker/docker/api/types/registry"
 	"github.com/docker/docker/registry"
-	docker "github.com/fsouza/go-dockerclient"
 )
 
 func parseDockerImage(image string) (repo, tag string) {
```
shoenig, line 23: what is the expected input to this now? how come `docker.ParseRepositoryTag` doesn't work anymore?

shoenig (commented):
left some suggestions and questions

drivers/docker/stats.go
```diff
@@ -141,16 +137,40 @@ func (h *taskHandle) collectStats(ctx context.Context, destCh *usageSender, inte
...
+		// sometimes it won't. The reader won't close until the container stops, so it's
+		// up to us to digest this stream carefully.
+		// The scanner below gets just one line and sends it to the channel.
+		for statsStringScanner.Scan() {
+			if err := json.Unmarshal(statsStringScanner.Bytes(), &stats); err != nil {
+				h.logger.Error("error unmarshalling stats data for container", "error", err)
+				break
+			}
+			statsCh <- &stats
+			break
```
tgross, line 155: Seeing this `break` makes me question the outer loop. With each pass through the loop, we're spawning a _new_ `statsCh` and goroutine that receives on that channel to send to the destination channel periodically. But it doesn't look like anyone ever closes that channel. And then we have this inner loop that's pulling of Docker's API but throwing out any backlog. Which means the logic in `dockerStatsCollector` to ensure we're sending only the latest values doesn't work anymore.

drivers/docker/utils.go
pkazmierczak: `docker.ParseRepositoryTag` comes from fsouza client. I just incorporated the logic into our `parseDockerImage` function.

drivers/docker/coordinator.go
```diff
@@ -330,18 +336,18 @@ func (d *dockerCoordinator) removeImageImpl(id string, ctx context.Context) {
...
-		err := d.client.RemoveImageExtended(id, docker.RemoveImageOptions{
+		_, err := d.client.ImageRemove(d.ctx, id, image.RemoveOptions{
 			Force: true, // necessary to GC images referenced by multiple tags
 		})
 		if err == nil {
 			break
 		}
 
-		if err == docker.ErrNoSuchImage {
+		if strings.Contains(err.Error(), "No such image") {
```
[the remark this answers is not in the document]
pkazmierczak: They are... https://pkg.go.dev/github.com/docker/docker@v27.1.1+incompatible/errdefs
but the tricky part is that you can never be quite sure which error type the API returns, as this isn't documented (at least not in the Go SDK). So unless you sit down for a long session with a debugger, there's no way of knowing. 

I'll try to update as many checks of this sort as I can, because of course I agree it's sloppy to rely on string comparisons, but this PR has already been _a lot_ of work, and I may end up creating a follow-up ticket to make these improvements. Then again I'm in the trenches already, might as well do it.

drivers/docker/driver_test.go
pkazmierczak: oh it most definitely shouldn't.

drivers/docker/stats.go
pkazmierczak: _actually_, the more I think about this, the more I'm inclined to indeed remove the outer loop. We could implement interval timer inside the scanner loop, and do the backoff on the API call separately. Hold my coffee, I fix.

pkazmierczak: have a look at [15ef385](https://github.com/hashicorp/nomad/pull/23966/commits/15ef385cf28c62d0e51625270d4b0bdf0ca9bc11)? Doesn't remove the outer loop (it's hard to do backoff retries without it), but improves the code a bit I think.

pkazmierczak:
@tgross @shoenig I addressed most of your comments. I still have 2 unit tests failing, failures I cannot reproduce on my aws linux instance or my macbook:
```
=== Failed
=== FAIL: drivers/docker TestDockerDriver_PidsLimit (23.25s)
    docker.go:34: Successfully connected to docker daemon running version 1.46
    docker.go:34: Successfully connected to docker daemon running version 1.46
2024-09-25T17:35:20.963Z [DEBUG] stdio EOF, exiting copy loop
2024-09-25T17:35:20.963Z [DEBUG] accepting initial connection: addr=/tmp/plugin1611906271
2024-09-25T17:35:20.963Z [DEBUG] stdio EOF, exiting copy loop
2024-09-25T17:35:20.963Z [DEBUG] making new client mux initial connection: addr=/tmp/plugin1611906271
2024-09-25T17:35:20.964Z [DEBUG] client muxer connected: addr=/tmp/plugin1611906271
2024-09-25T17:35:20.964Z [DEBUG] initial server connection accepted: addr=/tmp/plugin1611906271
2024-09-25T17:35:20.964Z [DEBUG] sending conn to default listener
    driver_linux_test.go:101: 
        driver_linux_test.go:101: expected nil error
        ↪ error: Expected "can't fork" in output ""
2024-09-25T17:35:44.212Z [DEBUG] sending conn to default listener
2024-09-25T17:35:44.212Z [DEBUG] stdio: received EOF, stopping recv loop: err="rpc error: code = Unavailable desc = error reading from server: EOF"
=== FAIL: drivers/docker TestDockerDriver_Stats (unknown)
    docker.go:34: Successfully connected to docker daemon running version 1.46
    docker.go:34: Successfully connected to docker daemon running version 1.46
2024-09-25T17:36:34.419Z [DEBUG] stdio EOF, exiting copy loop
2024-09-25T17:36:34.419Z [DEBUG] accepting initial connection: addr=/tmp/plugin1780848558
2024-09-25T17:36:34.419Z [DEBUG] stdio EOF, exiting copy loop
2024-09-25T17:36:34.419Z [DEBUG] making new client mux initial connection: addr=/tmp/plugin1780848558
2024-09-25T17:36:34.420Z [DEBUG] client muxer connected: addr=/tmp/plugin1780848558
2024-09-25T17:36:34.420Z [DEBUG] initial server connection accepted: addr=/tmp/plugin1780848558
2024-09-25T17:36:34.420Z [DEBUG] sending conn to default listener
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x113a826]
goroutine 5892 [running]:
github.com/hashicorp/nomad/drivers/docker.TestDockerDriver_Stats.func1(0xc0019b33b0)
	/home/runner/work/nomad/nomad/drivers/docker/driver_test.go:2228 +0x186
created by github.com/hashicorp/nomad/drivers/docker.TestDockerDriver_Stats in goroutine [577](https://github.com/hashicorp/nomad/actions/runs/11038110752/job/30660607541#step:5:578)4
	/home/runner/work/nomad/nomad/drivers/docker/driver_test.go:2219 +0x265
```

I will look into this again in the morning. I ran an e2e suite against this branch, too: https://github.com/hashicorp/nomad-e2e/actions/runs/11038171001/job/30660853139

Could you have another 👀 at the comment replies and commits referenced?

drivers/docker/driver.go
pkazmierczak: is there a clever way to check against interface types like these?
https://pkg.go.dev/github.com/docker/docker@v27.1.1+incompatible/errdefs#ErrNotFound

drivers/docker/stats.go
```diff
@@ -141,16 +138,36 @@ func (h *taskHandle) collectStats(ctx context.Context, destCh *usageSender, inte
...
+		// during 1 interval window, it will respond with multiple Stats objects,
+		// sometimes it won't. The reader won't close until the container stops, so it's
+		// up to us to digest this stream carefully.
+		s, err := statsStringReader.ReadString('\n')
+		if err != nil {
+			h.logger.Error("error reading stats stream", "error", err)
+		}
+
+		if err := json.Unmarshal([]byte(s), &stats); err != nil {
+			h.logger.Error("error unmarshalling stats data for container", "error", err)
```
tgross, line 155:
```suggestion
		if err != nil {
			h.logger.Error("error reading stats stream", "error", err)
			continue
		}

		if err := json.Unmarshal([]byte(s), &stats); err != nil {
			h.logger.Error("error unmarshalling stats data for container", "error", err)
			continue
```

tgross:
> @tgross @shoenig I addressed most of your comments. I still have 2 unit tests failing, failures I cannot reproduce on my aws linux instance or my macbook:

I haven't been able to reproduce the PIDs limit error either. It seems like a pretty straightforward test. I would recommend updating this to use `shoenig/tests` more robust `wait` package though. It makes the timeouts easier to understand by a lot. That'd be:

```diff
        // Check that data was written to the directory.
        outputFile := filepath.Join(task.TaskDir().LogDir, "redis-demo.stderr.0")
+
        exp := "can't fork"
-       tu.WaitForResult(func() (bool, error) {
+       must.Wait(t, wait.InitialSuccess(wait.ErrorFunc(func() error {
                act, err := os.ReadFile(outputFile)
                if err != nil {
-                       return false, err
+                       return err
                }
                if !strings.Contains(string(act), exp) {
-                       return false, fmt.Errorf("Expected %q in output %q", exp, string(act))
+                       return fmt.Errorf("Expected %q in output %q", exp, string(act))
                }
-               return true, nil
-       }, func(err error) {
-               must.NoError(t, err)
-       })
+               return nil
+       }),
+               wait.Timeout(5*time.Second),
+               wait.Gap(50*time.Millisecond),
+       ))
 }
```

For the `TestDockerDriver_Stats` test, its panicking because it's getting a nil `*structs.TaskResourceUsage` object over the channel. If I make this change:

```diff
                select {
                case ru := <-ch:
+                       must.NotNil(t, ru)
                        must.NotNil(t, ru.ResourceUsage)
                case <-time.After(3 * time.Second):
                        errChan <- struct{}{}
```

Then I get an error like the following instead, but only maybe 1 out of 3 runs, so it's timing dependent.

```
    driver_test.go:2228:
        driver_test.go:2228: expected to not be nil; is nil
    driver_test.go:2247: timeout
```

That means we're getting a nil object (likely because we're reading on a closed channel?) _and_ then sitting there waiting on the `WaitTask` for more than 10s. That's because we only destroy the task if that assertion is successful. We should clean up that goroutine so it cleans up unconditionally.

I noticed that you're missing `continue` lines on your error cases in `collectStats`, so I've added a suggestion for those.

pkazmierczak:
e2e run: https://github.com/hashicorp/nomad-e2e/actions/runs/11055624073/job/30715377429

tgross (approved):

ahjohannessen:
@tgross Seems like this has a breaking change with regards to:

```
plugin "docker" {
    config {
        auth {
            config = "/etc/nomad.d/.docker-creds"

        }
    }
}
```

Nomad 1.8.4 the above works, bumping to 1.9.0 results in:

```
Failed to pull `quay.io/our-repo/app:0.0.1`: Error response from daemon: unauthorized: access to the requested resource is not authorized
```

tgross:
@ahjohannessen please open a new issue

PR #23971, opened 2024-09-16
deps: update to go-set/v3 and refactor to use custom iterators

shoenig:
This PR updates `go-set` to `v3` which includes support for [custom iterators](https://pkg.go.dev/iter#hdr-Iterators).

We can use them to replace cases of `.ForEach` which involved creating a clunky closure with a boolean return value for halting iteration. 

```go
s.ForEach(func(item T) bool {
  // do stuff
  return true
})
```

becomes

```go
for item := range s.Items() {
  // do stuff
}
```

We can also use them to replace cases of `Slice()` when used to range over the elements of a set, avoiding creating a full slice worth of garbage.

```go
for _, v := range s.Slice() { // creates a whole N sized slice
}
```

becomes

```go
for v := range s.Items() { // automagically iterates in-place
}
```

### DIFF
--- a/client/acl.go
+++ b/client/acl.go
@@ -7,7 +7,7 @@ import (
 	"time"
 
 	"github.com/armon/go-metrics"
-	"github.com/hashicorp/go-set/v2"
+	"github.com/hashicorp/go-set/v3"
 	"github.com/hashicorp/nomad/acl"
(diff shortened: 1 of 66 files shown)


tgross (approved):
LGTM! This is cool!

PR #23972, opened 2024-09-16
deps update go1.23.1

shoenig:
- deps: update to go1.23.1 (from go1.22.6)
- deps: update to golangci-lint 1.61.0 (for compatibility)

Wanting to do this in preparation of https://github.com/hashicorp/nomad/pull/23971 which enables us to use custom iterators in `go-set/v3`


tgross (approved):
LGTM

PR #23979, opened 2024-09-17
fingerprint: update landlock test to accept v4+ APIs

tgross:
The landlock fingerprint test assumes there's no version of the landlock API >3. Update the test assertion to allow for the current v4 and any future
versions.


shoenig (approved):
LGTM!

PR #23980, opened 2024-09-17
Port and AllocatedPortMapping msgpack omitempty

gulducat:
to optimize log entries with empty fields, which TestPlanNormalize checks.

split out from #23956 so this can be backported.


nomad/structs/diff_test.go
```diff
@@ -6843,9 +6843,7 @@ func TestTaskDiff(t *testing.T) {
 						"a": 1,
 						"b": 2,
 					},
-					"boom": &Port{
-						Label: "boom_port",
-					},
+					"boom": []string{"boom_port"},
```
gulducat, line 6846: these diff tests are for `task.config`, which doesn't actually know what a `Port` is.  for the docker driver, for example, `ports` is a list of strings (which arguably is already tests with `bam` above here), or a `MapStrInt` `port_map` (`baz`).

maybe `Port` here made sense at some point in the past, but doesn't seem right to me now.

tgross (approved):
LGTM!

PR #23986, opened 2024-09-18
add missing checks for vault binary in unit tests

mismithhisler:
Follows the pattern from `NewTestVault` to check for the vault binary, and skips if it does not exist.

### DIFF
--- a/testutil/vault.go
+++ b/testutil/vault.go
@@ -49,7 +49,7 @@ func NewTestVaultFromPath(t testing.T, binary string) *TestVault {
 	t.Helper()
 
 	if _, err := exec.LookPath(binary); err != nil {
-		t.Skipf("Skipping test %s, Vault binary %q not found in path.", t.Name(), binary)
+		t.Skipf("Skipping test, Vault binary %q not found in path.", binary)
 	}
 
 	// Define which log level to use. Default to the same as Nomad but allow a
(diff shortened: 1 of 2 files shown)


testutil/vault.go
```diff
@@ -140,10 +140,13 @@ func NewTestVault(t testing.T) *TestVault {
...
 
-// NewTestVaultDelayed returns a test Vault server that has not been started.
-// Start must be called and it is the callers responsibility to deal with any
-// port conflicts that may occur and retry accordingly.
-func NewTestVaultDelayed(t testing.T) *TestVault {
+func NewTestVaultDelayedFromPath(t testing.T, binary string) *TestVault {
+	t.Helper()
+
+	if _, err := exec.LookPath(binary); err != nil {
+		t.Skipf("Skipping test %s, Vault binary not %q found in path.", t.Name(), binary)
```
shoenig, line 147: pretty sure `go test` already outputs the test name when you skip it (i think this was just copied from NewTestVaultFromPath anyway)

mismithhisler: I did copy this.  I'm happy to update both `Skipf` lines if you want.

shoenig: It's up to you. For nitpicky stuff like this we leave it up to the PR author. (it can be annoying to wait for a whole CI cycle for unimportant stuff)

mismithhisler: I removed the names of the tests to clean up with output a bit.

PR #24013, opened 2024-09-19
[ui] Version Tags on the job versions page

philrenaud:
Version tag CRUD in the UI. Allows a user to create, modify, and remove version tags.

image

### DIFF
--- a/nomad/job_endpoint.go
+++ b/nomad/job_endpoint.go
@@ -2429,5 +2429,11 @@ func (j *Job) TagVersion(args *structs.JobApplyTagRequest, reply *structs.JobTag
 	}
 
 	reply.Index = index
+	if args.Tag != nil {
+		reply.Name = args.Tag.Name
+		reply.Description = args.Tag.Description
+		reply.TaggedTime = args.Tag.TaggedTime
+	}
+
 	return nil
 }
--- a/ui/app/templates/components/job-version.hbs
+++ b/ui/app/templates/components/job-version.hbs
@@ -4,7 +4,7 @@
 ~}}
 
 <section class="job-version">
-  <div class="boxed-section {{if this.version.taggedVersion "tagged"}}">
+  <div class="boxed-section {{if this.version.taggedVersion "tagged"}}" data-test-tagged-version={{if this.version.taggedVersion "true" "false"}}>
     <header class="boxed-section-head is-light inline-definitions">
       Version #{{this.version.number}}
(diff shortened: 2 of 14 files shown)


ui/app/templates/components/job-version.hbs
```diff
@@ -3,55 +3,110 @@
...
+    </header>
+    {{#if this.isOpen}}
+      <div class="boxed-section-body is-dark">
+        <JobDiff @diff={{this.version.diff}} @verbose={{this.verbose}} />
+      </div>
+    {{/if}}
+    <footer class="boxed-section-foot {{if this.isEditing "editing"}}">
+      {{#if this.isEditing}}
+        <form class="tag-options" autocomplete="off" {{on "submit" (action "saveTag")}}>
+          {{! template-lint-disable no-down-event-binding }}
```
philrenaud, line 42: This is to prevent `{{on "keydown" (action this.handleKeydown)}}` from throwing a linting error. This method watches for an `ESC` keypress to close edit mode.

I have a difference of opinion with our linter, expressed better than I can do it at https://x.com/ID_AA_Carmack/status/1787850053912064005

nomad/job_endpoint.go
```diff
@@ -2429,5 +2429,11 @@ func (j *Job) TagVersion(args *structs.JobApplyTagRequest, reply *structs.JobTag
 	}
 
 	reply.Index = index
+	if args.Tag != nil {
+		reply.Name = args.Tag.Name
+		reply.Description = args.Tag.Description
+		reply.TaggedTime = args.Tag.TaggedTime
+	}
```
philrenaud, line 2436: Side effect: wanted to use the POST/PUT response here and realized I was sending back instantiated-but-not-populated data.

ui/app/templates/components/two-step-button.hbs
```diff
@@ -32,7 +32,7 @@
   />
   <Hds::Button
     data-test-confirm-button
-    @size="medium"
+    @size={{@size}}
```
philrenaud, line 35: Side effect: our two-step-buttons were always medium sized. This gives them all the flexibility of our other Helios buttons.

nomad/job_endpoint.go
gulducat: non-blocking: this reflects back to the user just what they passed in, but so long as we can depend on the raftApply erroring if it failed to enact their intentions, then this seems fine.

PR #24015, opened 2024-09-19
ci: (ent) specify instance types when using self hosted runners

shoenig:
I believe recent internal changes imply these jobs are running on some new default (AWS's smallest) spot instance types. Instead lets run them on something more powerful.

Internal RFC: IPS-056


jrasell (approved):
LGTM, thanks!

PR #24016, opened 2024-09-19
ci: try less hard to cache go modules

shoenig:
This PR drops the `mods` workflow as well as the use of the `hashicorp/setup-go` action. These were added in the past to try and speed up downstream workflows by having go modules already cached. Since then, the `actions/setup-go` action added support for caching which we can now just make use of. But also we've been required to split some workflows into running on self-hosted runners (in ENT). On these runners the caching mechanism implies a slow rsync between the underlying Azure infrastructure and the runner on AWS. Empirically I've found it's much faster to just download the Go modules from the Go module cache hosted by Google. So, disable caching on self hosted runners.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -21,7 +21,7 @@ on:
 jobs:
   checks:
     # largest available self-hosted disk for extra iops because linting is io-intensive
-    runs-on: ${{ endsWith(github.repository, '-enterprise') && fromJSON('["self-hosted", "ondemand", "linux", "disk_gb=255", "type=m7a.2xlarge;m6a.2xlarge"]') || 'ubuntu-22.04' }}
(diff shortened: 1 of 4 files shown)


shoenig:
@tgross the failure is 

```
=== RUN   TestPlanEndpoint_ApplyConcurrent
    plan_endpoint_test.go:194: 
        	Error Trace:	/home/runner/work/nomad/nomad/nomad/plan_endpoint_test.go:194
        	            				/opt/hostedtoolcache/go/1.23.1/x64/src/runtime/asm_amd64.s:1700
        	Error:      	Received unexpected error:
        	            	keyring has not been initialized yet
```

is this just keyring stuff going on?

tgross:
@shoenig yeah, any place we see that we need to swap out `WaitForLeader` to `WaitForKeyring` ([see also this internal thread](https://hashicorp.slack.com/archives/C07TFF3HQ/p1726761609204179))

tgross (approved):
LGTM!

PR #24021, opened 2024-09-20
test: wait for keyring for plan apply tests

tgross:
In #23977 we merged a change to how the keyring was stored. Because keyring initialization takes slightly longer now, this uncovered existing timing bugs in some of our tests where tests that require the keyring (ex. plan applier tests) were waiting for the leader but not the keyring initialization. Fix some of the examples we've seen cause test flakes.


shoenig (approved):
LGTM!

PR #24023, opened 2024-09-20
command line tools for redacting keyring from snapshots

tgross:
In #23977 we moved the keyring into Raft, which can expose key material in Raft snapshots when using the less-secure AEAD keyring instead of KMS. This changeset adds tools for redacting this material from snapshots:

* The `operator snapshot state` command gains the ability to display key metadata (only), which respects the `-filter` option.
* The `operator snapshot save` command gains a `-redact` option that removes key material from the snapshot after it's downloaded.
* A new `operator snapshot redact` command allows removing key material from an existing snapshot.

Full documentation on the new keyring work coming in a separate PR

### DIFF
--- a/command/operator_snapshot_redact.go
+++ b/command/operator_snapshot_redact.go
@@ -1,0 +1,95 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package command
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/hashicorp/nomad/helper/raftutil"
+	"github.com/posener/complete"
+)
+
+type OperatorSnapshotRedactCommand struct {
+	Meta
+}
+
+func (c *OperatorSnapshotRedactCommand) Help() string {
+	helpText := `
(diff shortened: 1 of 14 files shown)


tgross:
Failing test is because of https://github.com/hashicorp/nomad/pull/24024

command/operator_snapshot_redact.go
```diff
@@ -0,0 +1,95 @@
...
+		c.Ui.Error(fmt.Sprintf("Failed to copy snapshot to temporary file: %v", err))
+		return 1
+	}
+
+	err = raftutil.RedactSnapshot(tmpFile)
+	if err != nil {
+		c.Ui.Error(fmt.Sprintf("Failed to redact snapshot: %v", err))
+		return 1
+	}
+
```
shoenig, line 86: might be good to throw in a `tmpFile.Sync()` here, ensuring the following rename is to a fully written file

tgross: Added to the shared function in the `snapshot` package so that any future callers don't have to remember to do it.

shoenig (approved):
LGTM!

PR #24024, opened 2024-09-20
E2E: allow Consul version to omit tags

tgross:
When we start the Consul agent in the `consulcompat` test package, we check that the version matches the version we expect. But Consul agents may omit non-core parts of the version string (ex. `1.20.0-rc1` displays `1.20.0`). Compare only the core portions of the version string.


shoenig (approved):
LGTM

Issue #20335, opened 2024-04-09
raft: upgrade to v1.7.0 which introduces pre-vote extension


jrasell:
In the near future, Raft will release v1.7.0 which introduces the pre-vote extension: https://github.com/hashicorp/raft/pull/530

The feature particularly improves cluster stability when a Nomad server is partition from the cluster, but rejoins at a later date. The PR has a conclusive writeup.

 In order to update Nomad we will want to decide if the new `preVoteDisabled` configuration flag is exposed to operators and perform testing. The testing should include:
 
 - upgrade testing with mixed Raft versions
 - leadership transitions
 - server partitioning and rejoining
 - pre-vote interactions with Autopilot
 - load testing

PR #24029, opened 2024-09-20
deps: upgrade raft from v1.6.1 -> v1.7.1

schmichael:
The major change between Raft v1.6 -> v1.7 was the introduction of the Prevote feature. Before Prevote, when a partitioned node rejoins a cluster it may cause an election even if the cluster was stable. Prevote can avoid this useless election so reintroducing partitioned servers to an otherwise stable cluster becomes seamless.

Full details: https://github.com/hashicorp/raft/pull/530

In #20335 we discussed whether or not to add a configuration option to disable prevote in case bugs were discovered. While bugs have been found (hence the v1.7.1 version as opposed to v1.7.0), I'm choosing to follow Vault's lead of straightfordwardly bumping the raft dependency: hashicorp/vault#27605 and hashicorp/vault#28218

### DIFF
--- a/.changelog/24029.txt
+++ b/.changelog/24029.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+raft: Bump raft to v1.7.1 which includes pre-vote. This should make servers more stable after network partitions
+```
--- a/go.mod
+++ b/go.mod
@@ -82,7 +82,7 @@ require (
 	github.com/hashicorp/memberlist v0.5.1
 	github.com/hashicorp/net-rpc-msgpackrpc/v2 v2.0.0
 	github.com/hashicorp/nomad/api v0.0.0-20230103221135-ce00d683f9be
-	github.com/hashicorp/raft v1.6.1
+	github.com/hashicorp/raft v1.7.1
 	github.com/hashicorp/raft-autopilot v0.1.6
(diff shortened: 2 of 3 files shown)


jrasell (approved):
LGTM 🥳

pkazmierczak (approved):
LGTM and a welcome change for sure, but I'd love to see this tested in nomad-bench against raft 1.6 to see if there aren't any performance regressions.

tgross (approved):
LGTM

As much as a functional change like this is scary to backport, we should do so anyways. There aren't LTS branches for Raft so if there's a security bug in the library in the future we'd end up having to backport anyways.

schmichael:
> As much as a functional change like this is scary to backport, we should do so anyways. There aren't LTS branches for Raft so if there's a security bug in the library in the future we'd end up having to backport anyways.

Vault does not have plans to backport Raft 1.7.1, so I would like to match their behavior. If a Raft CVE or critical-bug fix lands in Raft 1.7.x, we will backport Raft 1.7.x for that fix. Hopefully that will give PreVote a bit more time testing in the wild, although we have no reason to suspect issues.

> LGTM and a welcome change for sure, but I'd love to see this tested in nomad-bench against raft 1.6 to see if there aren't any performance regressions.

Created an internal task for me to do this post-merge: https://hashicorp.atlassian.net/browse/NET-11147

PR #24045, opened 2024-09-23
add changelog entry for workload identity filename feature

mismithhisler:
Adds missing changelog entry for previous [PR](https://github.com/hashicorp/nomad/pull/24038) regarding the new filename parameter in the job spec.


tgross (approved):
LGTM!

PR #24052, opened 2024-09-24
fix: add node pool attribute to interpretable values in task env

mismithhisler:
Added `node.pool` attribute to interpretable node attributes so it could be used as per the [docs](https://developer.hashicorp.com/nomad/docs/runtime/interpolation#node-attributes)


tgross (approved):
LGTM. Don't forget your changelog entry!

shoenig (approved):

PR #24056, opened 2024-09-24
test: add valid identity name for test task

mismithhisler:
Validation of the default task was enabled after adding the filename parameter in [#24038](https://github.com/hashicorp/nomad/pull/24038).  This adds a valid identity name to the test tasks so validation tests will pass.


tgross (approved):
LGTM

Issue #24051, opened 2024-09-24
nil pointer crash due to template change_mode script processing (1.8.3+)


ygersie:
We've just upgraded from Nomad 1.8.2 to 1.8.4 and a lot of our nodes are now in a crash loop due to the following nil pointer:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1c8c1b7]

goroutine 9215 [running]:
github.com/hashicorp/nomad/client/allocrunner/taskrunner.(*DriverHandle).Exec(0x0, 0xdf8475800, {0xc001545240, 0x1d}, {0xc001553f50, 0x1, 0xc0048a8480?})
	github.com/hashicorp/nomad/client/allocrunner/taskrunner/driver_handle.go:70 +0xf7
github.com/hashicorp/nomad/client/allocrunner/taskrunner/template.(*TaskTemplateManager).processScript(0xc002a65950, 0xc0013eb300, 0xc0026dca20?)
	github.com/hashicorp/nomad/client/allocrunner/taskrunner/template/template.go:597 +0x104
created by github.com/hashicorp/nomad/client/allocrunner/taskrunner/template.(*TaskTemplateManager).handleChangeModeScript in goroutine 3975
	github.com/hashicorp/nomad/client/allocrunner/taskrunner/template/template.go:563 +0x47
```

ygersie:
This looks to be related to the triggering of change_mode=script templates on client restarts that was introduced in 1.8.3 here: https://github.com/hashicorp/nomad/issues/24051

jrasell:
Hi @ygersie and thanks for raising this issue and finding a probable change which is associated to this. I'll mark this for roadmapping and get this looked into.

tgross:
Hi @ygersie! This was supposed to have been fixed in Nomad 1.8.4 by https://github.com/hashicorp/nomad/pull/23917 (which is missing from the changelog :facepalm: but I can see it's in the release tag here: https://github.com/hashicorp/nomad/commit/529b3d0d85b34dc9be280b75918f05e4e0464e36).


[...]

PR #15915, opened 2023-01-26
template: restore driver handle on update

lgfa29:
When the template hook Update() method is called it may recreate the template manager if the Nomad or Vault token has been updated.

This caused the new template manager did not have a driver handler because this was only being set on the Poststart hook, which is not called for inplace updates.

Closes #15851

PR #15192, opened 2022-11-09
template: protect use of template manager with a lock

shoenig:
This PR protects access to `templateHook.templateManager` with its lock. So
far we have not been able to reproduce the panic - but it seems either Poststart
is running without a Prestart being run first (should be impossible), or the
Update hook is running concurrently with Poststart, nil-ing out the templateManager
in a race with Poststart.

Fixes #15189

Backport only to 1.4.x and 1.3.x

PR #24057, opened 2024-09-24
template: fix panic in change_mode=script on client restart

tgross:
When we introduced `change_mode=script` to templates, we passed the driver handle down into the template manager so we could call its `Exec` method directly. But the lifecycle of the driver handle is managed by the taskrunner and isn't available when the template manager is first created. This has led to a series of patches trying to fixup the behavior (#15915, #15192, #23663, #23917). Part of the challenge in getting this right is using an interface to avoid the circular import of the driver handle.

But the taskrunner already has a way to deal with this problem using a "lazy handle". The other template change modes already use this indirectly through the `Lifecycle` interface. Change the driver handle `Exec` call in the template manager to a new `Lifecycle.Exec` call that reuses the existing behavior. This eliminates the need for the template manager to know anything at all about the handle state.

Fixes: https://github.com/hashicorp/nomad/issues/24051

### DIFF
--- a/client/allocrunner/taskrunner/driver_handle.go
+++ b/client/allocrunner/taskrunner/driver_handle.go
@@ -66,6 +66,9 @@ func (h *DriverHandle) Signal(s string) error {
 
 // Exec is the handled used by client endpoint handler to invoke the appropriate task driver exec.
 func (h *DriverHandle) Exec(timeout time.Duration, cmd string, args []string) ([]byte, int, error) {
+	if h == nil {
+		return nil, 0, ErrTaskNotRunning
+	}
 	command := append([]string{cmd}, args...)
 	res, err := h.driver.ExecTask(h.taskID, command, timeout)
 	if err != nil {
@@ -80,6 +83,9 @@ func (h *DriverHandle) ExecStreaming(ctx context.Context,
 	command []string,
 	tty bool,
 	stream drivers.ExecTaskStream) error {
+	if h == nil {
+		return ErrTaskNotRunning
+	}
 
 	if impl, ok := h.driver.(drivers.ExecTaskStreamingRawDriver); ok {
 		return impl.ExecTaskStreamingRaw(ctx, h.taskID, command, tty, stream)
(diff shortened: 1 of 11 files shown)


client/allocrunner/taskrunner/template/template_test.go
```diff
@@ -1220,29 +1209,6 @@ OUTER:
 			t.Fatal(t, "should have received an event")
 		}
 	}
-
-	// remove the reference to the task handle and update the template contents
```
tgross, line 1224: Note to reviewers: this is the test we added in https://github.com/hashicorp/nomad/pull/23917 that we knew even at the time didn't really test anything useful. Now that we've fixed the problem by fixing the architecture, this doesn't even have anything left to test so I'm removing it.

client/allocrunner/taskrunner/driver_handle.go
```diff
@@ -80,6 +83,9 @@ func (h *DriverHandle) ExecStreaming(ctx context.Context,
 	command []string,
 	tty bool,
 	stream drivers.ExecTaskStream) error {
+	if h == nil {
+		return ErrTaskNotRunning
+	}
```
tgross, line 88: These changes aren't really necessary now, but they'll provide belt-and-suspenders against future bugs.

client/allocrunner/taskrunner/template/template_test.go
gulducat: makes sense to remove this, but I don't see a test file called `task_runner_hook_test.go`?

gulducat (approved):
LGTM!

client/allocrunner/taskrunner/template/template_test.go
tgross: Sorry, that's a typo, it's this test: https://github.com/hashicorp/nomad/blob/main/client/allocrunner/taskrunner/template_hook_test.go#L277 Note that only an integration test that spins up a real task _and_ can restart a client can actually exercise this bug. So that falls under the general pile of "things we'd really like to do when we have solid E2E upgrade/restart testing".

Issue #23305, opened 2024-06-11
pending allocations stuck in pending state after adoption by a new deployment


lattwood:
### Nomad version
```
Nomad v1.7.7
BuildDate 2024-04-16T19:26:43Z
Revision 0f34c85ee63f6472bd2db1e2487611f4b176c70c
```

### Operating system and Environment details
All Linux

### Issue
* We have a *large* Nomad cluster- > 2k agents.
* We run health checks on each agent and do dynamic metadata updates.
* We have a large job that gets scheduled on 99% of the healthy agents.
* We use a shell script to query nomad for node counts matching a filter on relevant dynamic metadata & run `nomad job scale JOB TG COUNT` when it changes.
* We have a way to identify if the currently running deployment is due to a new Docker Image, or due to a scaling activity.
* If it's due to a scaling activity on another task group, we *cancel* that deployment, and issue a new `nomad job scale` command immediately afterwards.
    * Due to how frequently the dynamic metadata is changing, we are *_constantly_* doing deployments.
    * We kicked off 250 deployments in the last hour.
* Existing allocations _*usually*_ all get adopted by the new deployment.
* The issues:
    * most of the time, but not always, the version on the existing allocs that are still "Pending" at deployment cancelation are updated. but we end up with allocations that don't get the version updated.
    * sometimes the allocation exists, but isn't assigned to a client in the UI. if you inspect the alloc with the Nomad CLI, it tells you which client it has been assigned to.
    * allocations get stuck in the pending state. we've found sending a SIGQUIT to the Nomad Agent fixes this, but that doesn't scale to > 2k nodes.
    * allocations will get *_rescheduled_* before they ever leave the pending state. _YEAH_.
    * possibly related, but might not be- if a node heartbeat is missed and reregistered, we can end up with stuck pending jobs on a node as well. This is also fixed with a SIGQUIT.
        * The stuck jobs aspect is brutal, since that doesn't start the health/deployment timers as far as I can tell.
        * The fact restarting the nomad agent is a little unnerving. I vaguely recall seeing something in the logs about "allocations out of date" or something like that when this issue has manifested itself in the past.

### Reproduction steps
* run a big cluster of nodes with intermittent connectivity issues to the master.
* run a big long running job on said cluster
* constantly be deploying/scaling the jobspec
* wait

#### Expected Result
* nomad schedules and runs jobs

#### Actual Result
* nomad doesn't

#### Screenshots of Absurdity
[...]

PR #24058, opened 2024-09-24
deps: upgrade yamux from v0.1.1 to v0.1.2

schmichael:
The main point of this dependency upgrade is to pull in the fixes in hashicorp/yamux#127 which prevents leaking deadlocked goroutines. It has been observed to improve the issue in hashicorp/nomad#23305 but does not seem sufficient to fix it entirely.

Since touching yamux is a rare and scary event (bugs can have pathological consequences at scale), I do **not** intend to backport this. If we discover the improvements are stable and significant enough, or if further fixes land in yamux, backporting can be done at that time.

### DIFF
--- a/.changelog/24058.txt
+++ b/.changelog/24058.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+deps: Upgraded yamux to v0.1.2 to fix a bug where RPC connections could deadlock
+```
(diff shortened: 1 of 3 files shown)


tgross (approved):
LGTM but it needs a `go mod tidy` first

schmichael:
> LGTM but it needs a `go mod tidy` first

Doh, done, thanks. I really need a script that just runs `go mod tidy` before and after any other `go` command I run.

Issue #23794, opened 2024-08-12
☂️ Golden Versions


philrenaud:
Umbrella issue surrounding the ability to:
- tag/pin a job version
- run `diff` against that job version
- prevent garbage collection of that job version, and of that job.

---

[RFC](https://go.hashi.co/rfc/nmd-197)
[Jira](https://hashicorp.atlassian.net/browse/NET-10380)

---

## Structs:
- [x] #23836
- [x] `TaggedVersion` clone/copy on job versions

## Tagging:
- [x] #23846
  - [x] `nomad job tag` with a `-version` flag
  - [x] `nomad job tag` without a version flag infers latest
  - [x] `nomad job tag unset` to de-tag / lookup by name

## Diffing:
- [x] [Compare versions by name and version number with the nomad history command](https://github.com/hashicorp/nomad/pull/23889)
- [x] Add a `Name` property to objects in `nomad job history` output
- [ ] #23939

## Reverting:
- [x] #24053
- [x] [Prevent copying of job tag when reverting to a version (copying the version)](https://github.com/hashicorp/nomad/pull/24059)

## Config:
- [ ] Add `max_tagged_versions_per_job` agent config guardrail

## UI:
- [x] #24008
  - [x] Add tagnames to the list items in the jobs/job/versions page
  - [x] Add a "Save Version" button on the jobs/job/versions page
  - [x] Editable tagged version description
- [ ] #24063

## GC (TODO pending investigation):
- [x] #23921
- [x] Add conditional to core_sched process for core jobs to prevent a golden version from being reaped

## Sentinel (TODO pending investigation):
- [ ] Expose `job_versions` as a top-level object that accepts a string param
- [ ] ...

PR #24059, opened 2024-09-24
Job revert command and API endpoint can take a string version tag name

### LINKED ISSUE #24053: expand `nomad job revert` to take a tag string

philrenaud:
Allows the second argument on `nomad job revert` to be the string name of a [job version tag](https://github.com/hashicorp/nomad/issues/23794)

Fixes an issue where reverting (either by number or by tag name) to an older job version with a tag, duplicated that tag on the newly-run version.

(This is not desirable for, among other reasons, a same-tag-name situation making `tag unset` unreliable)

Resolves #24053

### DIFF
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -270,6 +270,23 @@ func (j *Jobs) Versions(jobID string, diffs bool, q *QueryOptions) ([]*Job, []*J
 	return j.VersionsOpts(jobID, opts, q)
 }
 
+// VersionByTag is used to retrieve a job version by its TaggedVersion name.
+func (j *Jobs) VersionByTag(jobID, tag string, q *QueryOptions) (*Job, *QueryMeta, error) {
+	versions, _, qm, err := j.Versions(jobID, false, q)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Find the version with the matching tag
+	for _, version := range versions {
+		if version.TaggedVersion != nil && version.TaggedVersion.Name == tag {
+			return version, qm, nil
+		}
+	}
+
+	return nil, nil, fmt.Errorf("version tag %s not found for job %s", tag, jobID)
+}
+
 type VersionsOptions struct {
 	Diffs       bool
 	DiffTag     string
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -531,7 +531,7 @@ func (j *Jobs) Dispatch(jobID string, meta map[string]string,
 // Revert is used to revert the given job to the passed version. If
 // enforceVersion is set, the job is only reverted if the current version is at
 // the passed version.
-func (j *Jobs) Revert(jobID string, version uint64, enforcePriorVersion *uint64,
+func (j *Jobs) Revert(jobID string, version *uint64, versionTag *string, enforcePriorVersion *uint64,
 	q *WriteOptions, consulToken, vaultToken string) (*JobRegisterResponse, *WriteMeta, error) {
 
 	var resp JobRegisterResponse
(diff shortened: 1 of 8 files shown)


api/jobs.go
```diff
@@ -531,7 +531,7 @@ func (j *Jobs) Dispatch(jobID string, meta map[string]string,
 // Revert is used to revert the given job to the passed version. If
 // enforceVersion is set, the job is only reverted if the current version is at
 // the passed version.
-func (j *Jobs) Revert(jobID string, version uint64, enforcePriorVersion *uint64,
+func (j *Jobs) Revert(jobID string, version *uint64, versionTag *string, enforcePriorVersion *uint64,
```
philrenaud, line 534: version becomes optional, and therefore now a pointer.

tgross: We can't change the signature of `api` package methods. We'll need a new `Revert`-like method to do this.

Better yet, we can do what we did on the Tag/Untag Versions PR and query for the version for a given tag in the CLI/UI and then leave this unchanged.

nomad/structs/structs.go
```diff
@@ -943,7 +943,10 @@ type JobRevertRequest struct {
 	JobID string
 
 	// JobVersion the version to revert to.
-	JobVersion uint64
+	JobVersion *uint64
```
tgross, line 946: You can't change the type of fields in RPC structs, because they can be in-flight across upgrades (which can take hours or even days for weirdly cautious users). We should be fixing this in the CLI/UI like we did for untag.

tgross: Hrm, you _might_ be able to get away with this if the encoding works across upgrades actually... but you'd need to test it with production binaries for both older and newer to be sure. It's really not worth the risk, and having pointer fields on these in the RPC handlers is a bit of a red-flag in terms of design anyways.

gulducat: Personally I'm not a fan of pointers to basic types in general.

What if neither were pointers and the contract was ~

```go
if JobVersionTag != "" {
	if JobVersion != 0 {
		return errors.New("can't have it both ways, bud")
	}
	// use JobVersionTag
} else {
	// use JobVersion
}
```

but I suppose if the CLI just does a lookup, then everything else on the backend can just deal in version uints as usual, so this PR could be limited to `command/agent/job_revert*.go`

tgross: Yeah, which is pretty much what we did for untag.

api/jobs.go
```diff
@@ -270,6 +270,23 @@ func (j *Jobs) Versions(jobID string, diffs bool, q *QueryOptions) ([]*Job, []*J
 	return j.VersionsOpts(jobID, opts, q)
 }
 
+// VersionByTag is used to retrieve a job version by its TaggedVersion name.
```
tgross, line 273: Nice little helper! Mirrors what we needed in the state store. :+1:

command/job_revert.go
```diff
@@ -132,15 +136,24 @@ func (c *JobRevertCommand) Run(args []string) int {
...
+
+	parsedVersion, ok, err := parseVersion(args[1])
+	if ok && err == nil {
+		revertVersion = parsedVersion
+	} else {
+		client, err := c.Meta.Client()
+		if err != nil {
+			c.Ui.Error(fmt.Sprintf("Error initializing client: %s", err))
+			return 1
+		}
```
tgross, line 150: You already have a client created above.

```suggestion
```

nomad/job_endpoint.go
```diff
@@ -656,6 +655,10 @@ func (j *Job) Revert(args *structs.JobRevertRequest, reply *structs.JobRegisterR
 	revJob := jobV.Copy()
 	revJob.VaultToken = args.VaultToken   // use vault token from revert to perform (re)registration
 	revJob.ConsulToken = args.ConsulToken // use consul token from revert to perform (re)registration
+
+	// Clear out the TaggedVersion to prevent tag duplication
+	revJob.TaggedVersion = nil
```
tgross, line 660: Oof, good catch!

tgross (approved):
LGTM!

PR #24068, opened 2024-09-26
[ui] Versions page diff dropdown: z-index so you can click on labels

philrenaud:
Noticed that while the radio button itself was clickable, the labels in the dropdown fell beneath the `.timeline` text and so weren't clickable, and in some browsers even showed up behind the timeline itself.

image

### DIFF
--- a/ui/app/styles/components/timeline.scss
+++ b/ui/app/styles/components/timeline.scss
@@ -138,3 +138,12 @@
     }
   }
 }
+
+.versions-page-header {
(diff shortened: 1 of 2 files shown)


pkazmierczak (approved):
LGTM! Percy tests fail though?

philrenaud:
> Percy tests fail though?

Deliberately — it detects a visual diff that was intended (moved the stuff up in the z-index and over to the right)

PR #24073, opened 2024-09-26
[ui] Modify variable access permissions for UI users with write in only certain namespaces

### LINKED ISSUE #23905: Operators with ACL-Limited Namespace Access Can't Easily Create Variable in `nomad/jobs` Path
### Nomad version
```
$ nomad version
Nomad v1.8.3+ent
BuildDate 2024-08-13T07:52:39Z
Revision 82fa712be0e7c1e07d6d630e0583c188347411ee
```

```
$ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.4 LTS"
```

### Issue
Nomad Operators with access limited to a singular namespace cannot click "+ Create Variable" button from Variables UI page when within the `/nomad/jobs/` path (`/ui/variables/nomad/jobs/`) as it is greyed out but the button is able to be clicked from the main Nomad Variables UI page (`/ui/variables/`) but that then requires them to type `nomad/jobs/someJobName` in the "Path" field during creation.

Conversely, one workaround for this is to explicitly add the respective namespace query parameter to the URL (i.e. `/ui/variables/path/nomad/jobs?namespace=<someNamespace>`) and _then_ they are able to use the "+ Create Variable" button from that page view (whereby "Path" field is pre-populated). This is obscure as the page view already filters-out items from the Nomad Variables table which do not correspond to the authenticated user's permitted namespace.

In our case have a internal team ("cpe" ) that acts as Nomad Operators within their own eponymous namespace. They authenticate to Nomad via OIDC and assume a role with the following ACL policy:
```hcl
namespace "cpe" {
  policy = "read"
  capabilities = ["alloc-exec", "read-logs"]

  variables {
    path "*" {
    }
  }
}

quota {
  policy = "read"
}

plugin {
  policy = "read"
}

node {
  policy = "read"
}

agent {
  policy = "read"
}
```

When they go to Nomad Variables UI page (`/ui/variables/`) they are able to create a variable from that view. 
image

When they then click into the `nomad/jobs/` path (`/ui/variables/path/nomad/jobs`), they are unable to create a variable from that view.
image

If they then modify the URL to explicitly include the namespace query param (`/ui/variables/path/nomad/jobs?namespace=cpe`), they are then able to create a variable from that view.
image

In my eyes, this is undesirable behavior (but perhaps I'm just not understanding the underlying design decision behind this).

philrenaud:
Resolves #23905


ui/app/abilities/variable.js
```diff
@@ -116,22 +116,29 @@ export default class Variable extends AbstractAbility {
...
-      // If you're checking if you can write from root, and you don't specify a namespace,
-      // Then if you can write in ANY path in ANY namespace, you can get to /new.
+    if (this.path === WILDCARD_GLOB) {
+      // If checking for write permission on the root path
       return this.policyNamespacesIncludeVariablesCapabilities(
         this.token.selfTokenPolicies,
         ['write'],
-        this._nearestMatchingPath(this.path)
+        WILDCARD_GLOB
       );
```
tgross, line 125: It's not clear to me why we need to special case the root path? Does `_nearestMatchingPath` and the ACL checks not handle this themselves? If not, should we push this logic down into those?

ui/app/abilities/variable.js
```diff
@@ -116,22 +116,29 @@ export default class Variable extends AbstractAbility {
...
-      return this.allPaths
-        .find((path) => path.name === matchingPath)
-        ?.capabilities?.includes('write');
+      if (this.namespace === WILDCARD_GLOB) {
+        // Checking for the * namespace at a specific path
+        return this.policyNamespacesIncludeVariablesCapabilities(
+          this.token.selfTokenPolicies,
+          ['write'],
+          matchingPath
+        );
```
tgross, line 135: For clarity and because I find `policyNamespacesIncludeVariablesCapabilities` a little hard to follow... does this check that the user has write-access to this path in _any_ namespace, or _all_ namespaces?

philrenaud: This will check for the presence of the request capabilities in at least one of the user's namespaces (at the specified path)

philrenaud: I'm going to change up the internal var name `namespacesWithVariableCapabilities` to be something like `variableCapabilitiesAmongNamespaces` or something to make it clearer

ui/app/abilities/variable.js
philrenaud: This ended up being a clue for me that I was barking up the wrong tree here.

I've since made the fix for the customer issue in the template file (instead of the "can I write?" check on the index page being namespace-dropdown dependent, it's a `*`, so if they can write in any namespace they get a "Create New Variable" button)

Then, changed the `can${write|read|destroy}` checks to be bimodal:
- if namespace is `*`, use `policyNamespacesIncludeVariablesCapabilities()` to see if any of the policies' namespaces' rules include the verb at the provided path
- if namespace is specified, run a filter on `allVariablePathRules[]` to check for namespace, path, and capability.

.changelog/24073.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+ui: Fixes an issue where variables paths would not let namespaced users write variables unless they also had * namespace variable write permissions
```
tgross, line 2: I don't know for sure but I would bet that this breaks markdown in the changelog once we stitch all the text files together. Let's avoid the problem by writing it out:

```suggestion
ui: Fixes an issue where variables paths would not let namespaced users write variables unless they also had wildcard namespace variable write permissions
```

tgross (approved):
LGTM

Long-term, I wonder if it wouldn't be a bad idea to model the backend's [`acl`](https://github.com/hashicorp/nomad/tree/main/acl) package directly so that we could run conformance tests between them. That would involve building out the radix trie datastructure we use there, so it's a bit of a lift, but we wouldn't need its transaction safety either.

PR #24075, opened 2024-09-27
docker: use errdefs instead of string comparisons when checking errors

pkazmierczak:


drivers/docker/handle.go
```diff
@@ -192,12 +192,12 @@ func (h *taskHandle) Kill(killTimeout time.Duration, signal string) error {
...
 		if err := h.Signal(ctx, signal); err != nil {
 			// Container has already been removed.
-			if strings.Contains(err.Error(), NoSuchContainerError) {
+			if errdefs.IsNotFound(err) {
 				h.logger.Debug("attempted to signal nonexistent container")
 				return nil
 			}
 			// Container has already been stopped.
-			if strings.Contains(err.Error(), ContainerNotRunningError) {
+			if errdefs.IsNotModified(err) {
```
tgross, line 200: `IsNotModified`?

drivers/docker/coordinator.go
```diff
@@ -9,13 +9,13 @@ import (
...
 	"io"
 	"regexp"
-	"strings"
 	"sync"
 	"time"
 
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/image"
 	"github.com/docker/docker/api/types/registry"
+	"github.com/docker/docker/errdefs"
```
tgross, line 18: This `errdefs` package is so weird if you go read the code. Instead of using the stdlib `errors.Is`, it does a bunch of type assertions. I guess this is ok if it's errors that they emit, so long as they never wrap them before emitting them?

drivers/docker/handle.go
pkazmierczak: yeah 304.

drivers/docker/coordinator.go
pkazmierczak: I checked with dlv, they don't. But yes it's weird and it's bad. Docker is bad.

drivers/docker/handle.go
pkazmierczak: https://docs.docker.com/reference/api/engine/version/v1.47/#tag/Container/operation/ContainerStop

tgross (approved):
LGTM!

PR #24069, opened 2024-09-26
Docs: Terraform prereq clarification

aimeeu:
Clarify Terraform prereq since you don't need to install the Terraform CLI locally.

Fixes: [CE-726](https://hashicorp.atlassian.net/browse/CE-726)

[CE-726]: https://hashicorp.atlassian.net/browse/CE-726?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #24078, opened 2024-09-27
Update to Nomad as an AWS OIDC Provider doc

sofixa:
A few small updates to the recent "Federate access to AWS with Nomad Workload Identity" documentation, most notably that `restart` isn't needed because AWS SDKs handle OIDC reauth gracefully (unlike *any* other type of auth - for all others it's cached statically on startup, so nothing but a full restart works in case your credentials expire).

### DIFF
--- a/website/content/docs/operations/aws-oidc-provider.mdx
+++ b/website/content/docs/operations/aws-oidc-provider.mdx
@@ -25,7 +25,8 @@ The instructions on this page also assume the following:
 
 - Your AWS account has the necessary permissions to create IAM roles, policies, hosted zones,
   and certificates.
-- You are using Terraform to manage your AWS infrastructure.
+- You are using Terraform to manage your AWS infrastructure and you have
+  [configured it to communicate with AWS](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#authentication-and-configuration).
 
 ## Workflow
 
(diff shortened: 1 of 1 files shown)


tgross (commented):
@sofixa the bits about Terraform were covered in #24069 so you'll want to rebase this. Other than that this LGTM

sofixa:
@tgross should I leave the link towards the Terraform AWS provider auth config? It seems relevant to me.

tgross:
> should I leave the link towards the Terraform AWS provider auth config? It seems relevant to me.

We changed this to be generic to backends like TFC, but I don't have a good handle on how provider config differs. Is that relevant still?

aimeeu:
> > should I leave the link towards the Terraform AWS provider auth config? It seems relevant to me.
> 
> We changed this to be generic to backends like TFC, but I don't have a good handle on how provider config differs. Is that relevant still?

Umm... if the user has Nomad running on AWS and is using Terraform to manage the infra, wouldn't the user already have Terraform configured to work with AWS?

tgross:
I've rebased on main and made a formatting adjustment, just so that we're talking about the same edits here.

> Umm... if the user has Nomad running on AWS and is using Terraform to manage the infra, wouldn't the user already have Terraform configured to work with AWS? This is an advanced guide.

Presumably we don't need to tell people they need to have Terraform to make `terraform` commands either. A cross-reference link can't hurt, I'd think. I wouldn't have bothered to open a PR to add one either if there wasn't a technical change. Either way.

aimeeu (approved):

PR #24081, opened 2024-09-27
docker: fix bug in waiting for container to exit

tgross:
In ##23966 when we switched to using the official Docker SDK client, we had more contexts to add because most of the library methods take one. But for some APIs like waiting for a container to exit after we've started it, we never want to close this context, because the container can outlive the Nomad agent itself. If we close this context, the waiter fails and then we stop the container (after only running 5min).

* Ref: [NET-11202 (comment)](https://hashicorp.atlassian.net/browse/NET-11202?focusedCommentId=550009)

### DIFF
--- a/drivers/docker/handle.go
+++ b/drivers/docker/handle.go
@@ -292,12 +292,12 @@ func (h *taskHandle) run() {
 
 	h.startCpusetFixer()
 
-	ctx, cancel := context.WithTimeout(context.Background(), dockerTimeout)
-	defer cancel()
-
 	var werr error
 	var exitCode containerapi.WaitResponse
-	exitCodeC, errC := h.infinityClient.ContainerWait(ctx, h.containerID, containerapi.WaitConditionNotRunning)
+	// this needs to use the background context because the container can
+	// outlive Nomad itself
+	exitCodeC, errC := h.infinityClient.ContainerWait(
+		context.Background(), h.containerID, containerapi.WaitConditionNotRunning)
 
 	select {
 	case exitCode = <-exitCodeC:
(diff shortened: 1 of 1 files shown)


drivers/docker/handle.go
```diff
@@ -308,6 +308,9 @@ func (h *taskHandle) run() {
 		h.logger.Error("failed to wait for container; already terminated")
 	}
 
+	ctx, cancel := context.WithTimeout(context.Background(), dockerTimeout)
```
tgross, line 311: This 5min timeout seems excessive, but we didn't have a context for this or the `ContainerStop` requests before so I'm not sure what we want to do here. Inspect seems like it should return basically immediately?

drivers/docker/handle.go
```diff
@@ -308,6 +308,9 @@ func (h *taskHandle) run() {
 		h.logger.Error("failed to wait for container; already terminated")
 	}
 
+	ctx, inspectCancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer inspectCancel()
```
tgross, line 312: The `dockerTimeout` of 5m seems ridiculous for these two API calls, which should return immediately as far as I can tell. Even 10s seems like a lot, but I don't really know what we can expect out of Docker's API here.

drivers/docker/handle.go
```diff
@@ -332,6 +335,8 @@ func (h *taskHandle) run() {
 
 	// Stop the container just incase the docker daemon's wait returned
 	// incorrectly.
```
schmichael, line 337: Maybe it's just because I haven't touched this code in ages, but it took me some digging to determine why it was ok to ignore kill_timeout here. A bit of a `why` comment could help here, but nbd either way.

```suggestion
	// incorrectly. Container should have exited by now so kill_timeout can be ignored.
```

schmichael (approved):

pkazmierczak (approved):
LGTM!

PR #24092, opened 2024-09-30
docker: reset timer after collecting stats

tgross:
In ##23966 when we switched to using the official Docker SDK client, we had to rework the stats collection loop for the new client. But we missed resetting the timer on the collection loop, which meant that we'd only collect stats once and then never again.

* Ref: [NET-11202 (comment)](https://hashicorp.atlassian.net/browse/NET-11202?focusedCommentId=550814)


pkazmierczak (approved):
LGTM, great catch Tim!

PR #24093, opened 2024-10-01
deps: switch to moby/sys/capability

kolyshkin:
github.com/moby/sys/capability is a fork of the (no longer maintained)
github.com/syndtr/gocapability package.

For changes since the fork took place, see
https://github.com/moby/sys/blob/main/capability/CHANGELOG.md

### DIFF
--- a/drivers/shared/capabilities/defaults.go
+++ b/drivers/shared/capabilities/defaults.go
@@ -7,7 +7,7 @@ import (
 	"fmt"
 	"regexp"
 
-	"github.com/syndtr/gocapability/capability"
+	"github.com/moby/sys/capability"
 )
 
 const (
@@ -40,18 +40,9 @@ func NomadDefaults() *Set {
 func Supported() *Set {
 	s := New(nil)
 
-	last := capability.CAP_LAST_CAP
-
-	// workaround for RHEL6 which has no /proc/sys/kernel/cap_last_cap
-	if last == capability.Cap(63) {
-		last = capability.CAP_BLOCK_SUSPEND
-	}
-
+	list, _ := capability.ListSupported()
 	// accumulate every capability supported by this system
-	for _, c := range capability.List() {
-		if c > last {
-			continue
-		}
+	for _, c := range list {
 		s.Add(c.String())
 	}
 
--- a/drivers/shared/executor/executor.go
+++ b/drivers/shared/executor/executor.go
@@ -27,7 +27,7 @@ import (
 	cstructs "github.com/hashicorp/nomad/client/structs"
 	"github.com/hashicorp/nomad/drivers/shared/executor/procstats"
 	"github.com/hashicorp/nomad/plugins/drivers"
-	"github.com/syndtr/gocapability/capability"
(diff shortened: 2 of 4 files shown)


tgross (changes requested):
@kolyshkin This PR updates a dependency that's worked just fine for us for many years and doesn't feel like it needs much in the way of "maintenance" (some software is simply done). 

In the process, you've removed a bugfix without explaining how the updated library solves it (it doesn't as far as I can tell) or at least reporting in the PR that you've determined which versions of which distros do have the `/proc` file in question so we can decide if that's relevant to our support requirements. I think we'd be willing to accept this PR if it came with a little more in the way of this supporting information.

tgross:
Didn't hear back on this so I'm going to close it out. If you've got more information we'd be happy to revisit.

kolyshkin:
My bad; somehow I missed this. Better late than never, let me address your review comments.

> @kolyshkin This PR updates a dependency that's worked just fine for us for many years and doesn't feel like it needs much in the way of maintenance (some software is simply "done").

I can relate but it is not the case here. In particular, this package, at the very least, requires updates when new capabilities are added into the kernel. It is also unversioned, lacks go.mod/go.sum, and has some real bugs that are fixed in the fork (although they don't affect your usage of the package). We would not fork it otherwise, since no one wants yet another package to maintain.

> In the process, you've removed a bugfix without explaining how the updated library solves it (it doesn't as far as I can tell) or at least reporting in the PR that you've determined which versions of which distros do have the `/proc` file in question so we can decide if that's relevant to our support requirements. I think we'd be willing to accept this PR if it came with a little more in the way of this supporting information.

The updated library does not solve it. The bugfix is buggy by itself, and is no longer needed, as I have explained at great depth in the commit message (my bad; I should have copied this to PR description).

Here's a copy-paste from it:

<blockquote>
Note that the "workaround for RHEL6" is removed for a number of reasons.
Feel free to choose the one you like the most, either is sufficient:

1. /proc/sys/kernel/cap_last_cap is available since RHEL 6.7
   (kernel 2.6.32-573.el6), released 9 years ago (2015-07-22).

2. It incorrectly returns CAP_BLOCK_SUSPEND (36), which was only added
   in kernel v3.5 and was never backported to RHEL6 kernels. The
   correct value for RHEL6 would be CAP_MAC_ADMIN (33).

3. As far as upstream kernels go, /proc/sys/kernel/cap_last_cap was
   added in kernel v3.2, and a correct value depends on the kernel
   version. It could be CAP_WAKE_ALARM (35), added to kernel v3.0, or
   CAP_SYSLOG (34), added to kernel v2.6.38, or possibly a lesser value
   for even older kernels.
</blockquote>

I'm not sure I can add anything to the above explanation, except that a similar change (removal of "workaround for RHEL6") was accepted in other places, e.g. https://github.com/containers/common/pull/2167.

tgross:
Thanks for following up @kolyshkin. I'll take a second look and we'll see about getting this merged in.

tgross (approved):
LGTM

PR #24102, opened 2024-10-01
differently exclude tagged job versions from being pruned

gulducat:
#23983 prevented tagged job versions from being pruned, but sometimes they would still count against `job_tracked_versions`, which was not what we intended.

This puts back the original old-version pruning logic, and instead filters out the tagged versions earlier (in `jobVersionByID()` to avoid additional slice construction to filter the list in `upsertJobVersion()`), so that tagged versions are simply not considered during pruning at all.

The test seems rather convoluted, but I opted to keep it because it exercises the holistic behavior that we want to see, while still also covering the more localized change to `jobVersionByID()`.

### DIFF
--- a/nomad/state/state_store.go
+++ b/nomad/state/state_store.go
@@ -2160,8 +2160,9 @@ func (s *StateStore) upsertJobVersion(index uint64, job *structs.Job, txn *txn) 
 		return fmt.Errorf("index update failed: %v", err)
 	}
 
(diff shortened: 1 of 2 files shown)


mismithhisler (approved):
Looks good!

philrenaud (approved):

tgross (approved):
LGTM!

PR #24104, opened 2024-10-02
fix: fix slice init length

rusttech:
The intention here should be to initialize a slice with a capacity of len(node.Attributes)  rather than initializing the length of this slice.

### DIFF
--- a/command/node_status.go
+++ b/command/node_status.go
@@ -794,7 +794,7 @@ func formatEventDetails(details map[string]string) string {
 
 func (c *NodeStatusCommand) formatAttributes(node *api.Node) {
 	// Print the attributes
-	keys := make([]string, len(node.Attributes))
+	keys := make([]string, 0, len(node.Attributes))
 	for k := range node.Attributes {
 		keys = append(keys, k)
 	}


tgross:
@rusttech the code as it exists here is correct. We're initializing a slice of len and cap equal to the correct size. There's no reason to specify the length as 0 here because we statically know the correct size for both, _and_ this is more tedious to write and read

tgross:
Well unfortunately on a second glance at this, I see that the code change _is_ correct, but that you haven't then followed-up with how the original code is compensating for the empty slice items.

rusttech:
> @rusttech the code as it exists here is correct. We're initializing a slice of len and cap equal to the correct size. There's no reason to specify the length as 0 here because we statically know the correct size for both, _and_ this is more tedious to write and read.
> 
> I'm going to be honest with you @rusttech that I see you've opened this identical PR on a half dozen other repositories. This kind of activity smells like unwelcome automated PRs. Please don't continue to do this.

I submitted them manually one by one, not through an automated program.

rusttech:
> Well unfortunately on a second glance at this, I see that the code change _is_ correct, but that you haven't then followed-up with how the original code is compensating for the empty slice items.

It's okay, as long as the problem is fixed.

Issue #17233, opened 2023-05-18
Remove TimeTable in favor of Object Timestamps


schmichael:
### Background

As of version 1.5 Nomad uses a special TimeTable data structure to track approximate timestamps for Raft indexes. These timestamps are used by Nomad's state garbage collection to determine how old an object is based on its Raft index compared to the wallclock time and GC threshold for that particular object.

For example in the case of Nodes:

- The default `node_gc_threshold` is 24 hours.
- Node A went `down` at Raft index 123
- The TimeTable recorded `2009-11-10 23:00:00 UTC` as the Nearest Time for Raft index 123.

Therefore at the first Node GC internal after `2009-11-11 23:00:00 UTC`, Node A is eligible for garbage collection.

Since all objects record their Raft create and/or modified index, TimeTable has historically provided a generic mechanism for determining when any object is created or modified. TimeTable must make a tradeoff between space used, history kept, and precision:

| | History | Precision | Entries | Size | 
| - | - | - | -| - |
| Current | 72h | 5m | 864 | 27kb |
| 30 days / High Precision | 720h | 5m | 8,640 | 270kb |
| 30 days / Low Precision | 720h | 1h | 720 entries | 23kb |

Thanks to being able to binary search (`O(log N)`) to find its time entry, TimeTable lookups are fairly cheap even for the 8.6k entry case.

However TimeTable is extra accounting (cpu), extra memory, extra state to snapshot. Never very much, but always overhead if there's a timestamp that can be used instead.

### Proposal

Deprecate TimeTable and move all uses to object-based timestamps. The following objects and operations would need to be migrated:

- Allocation - use ModifyTime instead
- Evaluations - use ModifyTime instead
  - pruneUnblockIndexes
  - evalGC
  - expiredACLTokenGC
- Job - use SubmitTime
- Node - use StatusUpdatedAt
- Deployment - **currently lacks a timestamp!**
- CSIVolumes - **currently lacks a timestamp!**
- RootKeyMeta - use CreateTime instead

tgross:
Related: https://github.com/hashicorp/nomad/issues/16359

PR #24112, opened 2024-10-02
state: remove TimeTable and rely on objects' modify times instead

pkazmierczak:
Core scheduler relies on a special table in the state store—the `TimeTable`—to figure out which objects can be GC'd. The `TimeTable` correlates Raft indices with objects insertion time, a solution we used before most of the objects we store in the state contained timestamps. This introduced a bit of a memory overhead and complexity, but most importantly meant that any GC threshold users set greater than `timeTableLimit = 72 * time.Hour` was ignored. This PR removes the `TimeTable` and relies on object timestamps to determine whether they could be GCd or not. 

Fixes https://github.com/hashicorp/nomad/issues/16359
Resolves https://github.com/hashicorp/nomad/issues/17233

internal ref: https://hashicorp.atlassian.net/browse/NET-10269


nomad/core_sched_test.go
```diff
@@ -2160,103 +2032,103 @@ func TestAllocation_GCEligible(t *testing.T) {
...
 		{
 			Desc:             "Don't GC when threshold not met",
 			ClientStatus:     structs.AllocClientStatusComplete,
 			DesiredStatus:    structs.AllocDesiredStatusStop,
-			GCTime:           fail,
-			ModifyIndex:      100,
-			ThresholdIndex:   90,
+			GCTime:           now,
+			ModifyTime:       now.UnixNano(),
+			CutoffTime:       now.Add(-1 * time.Hour),
```
schmichael, line 2101: Oh hey! I think this is susceptible to daylight savings time bugs! Don't change it to half an hour either as that could still break if @philrenaud runs the tests at just the wrong time while visiting Newfoundland: https://en.wikipedia.org/wiki/Newfoundland_Time_Zone (I may enjoy timezone hijinks too much.)

nomad/fsm.go
```diff
@@ -1583,11 +1563,6 @@ func (n *nomadFSM) restoreImpl(old io.ReadCloser, filter *FSMFilter) error {
 		// Decode
 		snapType := SnapshotType(msgType[0])
 		switch snapType {
-		case TimeTableSnapshot:
```
schmichael, line 1586: We actually need to keep this `case` as the `default` case crashes. We don't need to deserialize it, we just need to handle the case in order to prevent an `Unrecognized snapshot error`.

nomad/structs/structs.go
```diff
@@ -10834,6 +10842,9 @@ type DeploymentStatusUpdate struct {
 
 	// StatusDescription is the new status description of the deployment.
 	StatusDescription string
+
+	// UpdatedAt is the time of the update
```
schmichael, line 10846: I love being specific with all of these bare int64 timestamp fields... especially since they're not entirely consistent. :sob: 
```suggestion
	// UpdatedAt is the time of the update, stored as UnixNano
```

schmichael (approved):
Will terminal objects that didn't have a timestamp before be GC'd on their next gc interval? That seems fine.

I think we should make sure to really call this out in Version Specific Upgrade notes. While the vast majority of users shouldn't notice _anything,_ I think there are 2 classes of users who will:

1. Users who either intentionally or unintentionally used gc thresholds > 72h are in for a surprise.
2. Users, especially those with custom high gc thresholds, with lots of terminal objects that suddenly get GC'd "early" post-upgrade (as long as my understanding of the upgrade path is accurate).

This is a pretty astounding PR from an archeological standpoint: lots of 2015 Armon code getting axed! The fact that it Just Worked for most use cases for almost a decade is quite an achievement! :tada:

nomad/fsm.go
pkazmierczak: oh that's a good catch

api/csi.go
```diff
@@ -350,6 +350,8 @@ type CSIVolume struct {
 
 	CreateIndex uint64
 	ModifyIndex uint64
+	CreateTime  int64
+	ModifyTime  int64
```
tgross, line 354: Similar to @schmichael's comment for the structs package, because this is the public API it'd be especially nice for us to document that these are unix nano.

nomad/structs/csi.go
```diff
@@ -372,6 +380,8 @@ func NewCSIVolume(volumeID string, index uint64) *CSIVolume {
 		ID:          volumeID,
 		CreateIndex: index,
 		ModifyIndex: index,
+		CreateTime:  time.Now().UnixNano(),
+		ModifyTime:  time.Now().UnixNano(),
```
tgross, line 384: These values will be different because we're calling it twice (plus now we have to make two vDSO syscalls). Given all the callers for this function are tests, that might make spurious test failures in the future. Maybe this function should take a `now` parameter as we did for `NewDeployment`?

tgross (approved):
LGTM once comments are resolved. 

We should also include both a changelog entry and an upgrade guide note for 1.9.2. One thing in particular that stands out is that because some objects like Deployments are created in the scheduler (which could be on a follower), this requires that servers have at least roughly-sync'd clocks for correct GC. We probably already do implicitly require this _somewhere_ in the code base but it'd be good to have a note about it just in case someone is doing something weird and has been getting away with it for a while.

nomad/fsm.go
schmichael: We probably want to keep the const for a while as well to ensure we don't reuse its value and cause all kinds of fun problems.

schmichael:
> One thing in particular that stands out is that because some objects like Deployments are created in the scheduler (which could be on a follower), this requires that servers have at least roughly-sync'd clocks for correct GC.

Timetable also had the problem of using followers' clocks. I have typed and deleted a lot of conjectures around local skew vs remote skew and what the optimal way to handle time in a distributed system is....... but I don't think any of it is worth worrying about. Our time-based GC can't escape clock skew problems of some kind. Even if we always used the leader's time for insertion _and_ sent the leader's time with gc evals... leadership can change at any time! Clock skew is inescapable, and hopefully our users understand any time-based parameters are only as accurate as their clocks.

PR #24118, opened 2024-10-02
scheduler: non-nil err when no devices match

gulducat:
e.g. before:

```
* Dimension "devices: %!s(<nil>)" exhausted on 1 nodes
```

after:

```
* Dimension "devices: no devices match request" exhausted on 1 nodes
```


tgross (approved):
LGTM :+1:

PR #24124, opened 2024-10-03
update example device readme

gulducat:
While testing our new device quotas (enterprise), I had to (re)discover a couple little things to properly install and configure our `example` device plugin.

### DIFF
--- a/plugins/device/cmd/example/README.md
+++ b/plugins/device/cmd/example/README.md
@@ -50,7 +50,7 @@ touch device01 && chmod 0777 device01
 touch device02
 ```
 
-It should be fingerprinted by the client agent after the ~`list_period`,
+It should be fingerprinted by the client agent after the `list_period`,
 which you can check with:
 
 ```shell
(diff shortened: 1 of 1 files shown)


plugins/device/cmd/example/README.md
```diff
@@ -22,3 +38,84 @@ The valid configuration options are:
...
+Create two instances of the device, one unhealthy:
+
+```shell
+mkdir -p /tmp/nomad-device
+cd /tmp/nomad-device
+touch device01 && chmod 0777 device01
+touch device02
+```
+
+It should be fingerprinted by the client agent after the ~`list_period`,
```
tgross, line 53:
```suggestion
It should be fingerprinted by the client agent after the `list_period`,
```

tgross (approved):
LGTM!

plugins/device/cmd/example/README.md
gulducat: I often use a tilde to indicate "approximately" but I suppose that's not obvious :heavy_check_mark:

PR #24125, opened 2024-10-03
alloc fs: use case-insensitive check for reads of secret/private dir

tgross:
When using the Client FS APIs, we check to ensure that reads don't traverse into the allocation's secret dir and private dir. But this check can be bypassed on case-insensitive file systems (ex. Windows, macOS, and Linux with obscure ext4 options enabled). This allows a user with `read-fs` permissions but not `alloc-exec` permissions to read from the secrets dir.

This changeset updates the check so that it's case-insensitive. This risks false positives for escape (see linked Go issue), but only if a task without filesystem isolation deliberately writes into the task working directory to do so, which is a fail-safe failure mode.

Ref: https://github.com/golang/go/issues/18358
Ref: https://hashicorp.atlassian.net/browse/NET-10664

### DIFF
--- a/client/allocdir/alloc_dir_nonlinux_test.go
+++ b/client/allocdir/alloc_dir_nonlinux_test.go
@@ -1,0 +1,60 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+//go:build !linux
+// +build !linux
+
+package allocdir
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/hashicorp/nomad/ci"
+	"github.com/hashicorp/nomad/helper/testlog"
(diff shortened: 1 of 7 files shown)


dduzgun-security (approved):
LGTM! 👍 Thank you

client/allocdir/alloc_dir_nonlinux_test.go
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+//go:build !linux
+// +build !linux
```
schmichael, line 5: BSDs are gonna be sad, but we don't officially support them anyway. Perhaps:
```suggestion
//go:build darwin || windows
```

(can we drop the old `// +build` way? it was replaced 3+ years ago)

schmichael (approved):

PR #24127, opened 2024-10-03
fix data race in node upsert

tgross:
While testing with agents built with the race-detection option enabled, I encountered a data race while draining a node.

When we upsert a node we copy the `NodeResources` struct and then perform a fixup for backwards compatibility of the topology struct. This fixup was being executed on the original struct and not the copy, which means we're uselessly fixing up the wrong struct and we're corrupting the state store in the process (albeit harmlessly, I suspect).

Fix the data race by calling the method on the correct pointer.

### DIFF
--- a/nomad/structs/structs.go
+++ b/nomad/structs/structs.go
@@ -2251,7 +2251,7 @@ func (n *Node) Canonicalize() {
 		n.SchedulingEligibility = NodeSchedulingEligible
 	}
 
-	// COMPAT remove in 1.9+
+	// COMPAT remove in 1.10+
 	// In v1.7 we introduce Topology into the NodeResources struct which the client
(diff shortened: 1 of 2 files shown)


nomad/structs/structs.go
```diff
@@ -3264,7 +3264,7 @@ func (n *NodeResources) Copy() *NodeResources {
 
 	// COMPAT remove in 1.9+
 	// apply compatibility fixups covering node topology
```
tgross, line 3266: @shoenig: is this instead safe to simply remove in 1.9 as suggested by this comment? It looks like we introduced this in 1.6.x?

gulducat (approved):
:new_moon_with_face: (;new_moon_with_face:)

PR #24133, opened 2024-10-04
[ui] Update the Task Lifecycle Status chart

philrenaud:
- Now using [Helios](https://helios.hashicorp.design/) for component styles
- Added colour/border styles to indicate when a task is failed, running, pending, complete, or dead/never-started

Screenshot 2024-10-04 at 4 16 41 PM
Screenshot 2024-10-04 at 4 16 19 PM

(formerly looked like this:)
image


philrenaud:
A few quick notes for reviewers: This chart shows up in two places: on a Task Group page and on an Allocation page.

You may notice a few conditionals like `this.taskState?.prop` — taskStates are only passed in for the Allocation version of this chart (where these lifecycle hooks can be said to represent an actual running task, and therefore get links, a singular status, etc.) as opposed to on task group pages (where the chart is a map of the configuration of lifecycles, in other words, the map but not the territory)

The pending/shimmer effect is borrowed from the job status/deployments panel

schmichael (approved):
This is fantastic!

ui/app/components/lifecycle-chart-row.js
```diff
@@ -11,13 +11,45 @@ import classic from 'ember-classic-decorator';
...
+      color = 'neutral';
+    }
+    if (this.taskState?.state === 'dead') {
+      if (this.taskState?.failed) {
+        color = 'critical';
+      } else {
+        color = 'neutral';
+      }
     }
+    return color;
```
mismithhisler, line 29: What happens if this returns undefined?  Does it just use some default color?

mismithhisler (approved):

ui/app/components/lifecycle-chart-row.js
philrenaud: That's a good point; in this case it would cause an error because the component requires one of a small set of strings. I'll set a default on it.

Issue #24144, opened 2024-10-07
[ui] A new version run in the background incorrectly displays diffs on the versions page


philrenaud:
[Jira](https://hashicorp.atlassian.net/browse/NET-11205)

> When submitting a newer job version using Diff against previous version in the UI, the original version gets an incorrect diff until the page is refreshed.

PR #24055, opened 2024-09-24
Feature: Golden Versions

philrenaud:
Introduces the ability to tag/pin a version of a job, in order to:
- prevent it from being garbage-collected, and
- allow future comparisons/`diff`s and `revert`s by name (ex: `nomad job revert myJob someVersion` )

Includes new CLI commands (`nomad job history -diff-version`, `nomad job tag apply`, `nomad job tag unset`) and a new endpoint (POST/PUT and DELETE to `v1/job/$jobName/versions/$tagName`), plus new diff and CRUD in the web UI:

image

Resolves https://github.com/hashicorp/nomad/issues/23794

PR #24145, opened 2024-10-07
[ui] Separate Diffs and Versions from the /versions endpoint as far as Ember is concerned

philrenaud:
This separates `Diff` objects from `Version` objects at the route/model layer, preferring to join them at component/view layer.

For context: a diff living at `versions[].diff` has made sense prior to https://github.com/hashicorp/nomad/pull/24055, since a version diff was only ever against its previous version.

Now that we can modify the "version to diff against" on the versions page, that model comes with side-effects (like the bug the Jira ticket uncovers, where a newly-pushed version would unblock the /versions query, but no update the diffs' position until refresh)

Now when you load the /versions page, 3 things happen:
- a request to `/versions` takes place to get job versions and its data is returned on page
- a blocking query against `/versions?index=num` is opened, and resolved any time there is new version information / a new version
- a request to `/versions?diffs=true<&diffVersion=num` takes place _and only the `.Diffs` information is processed_

This is a little inefficient from a fetch perspective, but preserves a lot of conventions in our frontend (like "Watch the versions relationship on this job") and seems like the idiomatic way to do it.

Resolves https://github.com/hashicorp/nomad/issues/24144 ([Jira](https://hashicorp.atlassian.net/browse/NET-11205?focusedCommentId=551956))

### DIFF
--- a/ui/app/components/job-versions-stream.js
+++ b/ui/app/components/job-versions-stream.js
@@ -19,7 +19,9 @@ export default class JobVersionsStream extends Component {
   // Passes through to the job-diff component
   verbose = true;
 
-  @computed('versions.[]')
+  diffs = [];
+
+  @computed('versions.[]', 'diffs.[]')
   get annotatedVersions() {
     const versions = this.versions.sortBy('submitTime').reverse();
     return versions.map((version, index) => {
@@ -38,7 +40,8 @@ export default class JobVersionsStream extends Component {
         }
       }
 
-      return { version, meta };
+      const diff = this.diffs.objectAt(index);
+      return { version, meta, diff };
     });
   }
 }
--- a/ui/app/models/job.js
+++ b/ui/app/models/job.js
@@ -431,7 +431,7 @@ export default class Job extends Model {
 
   @attr('number') version;
 
-  @hasMany('job-versions', { async: true }) versions;
+  @hasMany('job-versions', { async: false }) versions;
   @hasMany('allocations') allocations;
   @hasMany('deployments') deployments;
   @hasMany('evaluations') evaluations;
--- a/ui/app/models/job.js
+++ b/ui/app/models/job.js
@@ -431,7 +431,7 @@ export default class Job extends Model {
 
   @attr('number') version;
 
-  @hasMany('job-versions', { async: false }) versions;
+  @hasMany('job-versions', { async: true }) versions;
   @hasMany('allocations') allocations;
   @hasMany('deployments') deployments;
   @hasMany('evaluations') evaluations;
--- a/ui/app/routes/jobs/job/versions.js
+++ b/ui/app/routes/jobs/job/versions.js
@@ -15,21 +15,8 @@ import { inject as service } from '@ember/service';
 export default class VersionsRoute extends Route.extend(WithWatchers) {
   @service store;
 
-  queryParams = {
-    diffVersion: {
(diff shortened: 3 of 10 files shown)


ui/app/components/job-versions-stream.js
```diff
@@ -19,7 +19,9 @@ export default class JobVersionsStream extends Component {
   // Passes through to the job-diff component
   verbose = true;
 
-  @computed('versions.[]')
+  diffs = [];
+
+  @computed('versions.[]', 'diffs.[]')
```
philrenaud, line 24: the `@computed` prefix to this getter indicates something that should trigger an update.

ui/app/components/job-versions-stream.js
```diff
@@ -38,7 +40,8 @@ export default class JobVersionsStream extends Component {
         }
       }
 
-      return { version, meta };
+      const diff = this.diffs.objectAt(index);
+      return { version, meta, diff };
```
philrenaud, line 44: This is the component layer where we opt to join the Version and its Diff, which are otherwise same-indexed objects in separate arrays.

ui/app/routes/jobs/job/versions.js
```diff
@@ -15,21 +15,8 @@ import { inject as service } from '@ember/service';
 export default class VersionsRoute extends Route.extend(WithWatchers) {
   @service store;
 
-  queryParams = {
-    diffVersion: {
-      refreshModel: true,
-    },
-  };
-
```
philrenaud, line 23: Now that the route doesn't care about Diff data (the controller does), we don't need to make it concerned with queryParams / it doesn't need to re-fetch the versions when the user switches the comparison version via dropdown

ui/app/serializers/job.js
```diff
@@ -210,7 +210,9 @@ export default class JobSerializer extends ApplicationSerializer {
       },
       versions: {
         links: {
-          related: buildURL(`${jobURL}/versions`, { namespace, diffs: true }),
+          related: buildURL(`${jobURL}/versions`, {
+            namespace,
+          }),
```
philrenaud, line 215: We no longer need to get `.Diffs[]` on initial model load

ui/app/models/job.js
```diff
@@ -431,7 +431,7 @@ export default class Job extends Model {
 
   @attr('number') version;
 
-  @hasMany('job-versions', { async: true }) versions;
+  @hasMany('job-versions', { async: false }) versions;
```
tgross, line 434: Am I understanding the docs right that this change makes it so that we're always reading whatever data has already been fetched, rather than returning a promise to await to go get that data? And that's ok because we're fetching the data as-needed elsewhere?

tgross (approved):
LGTM, with one non-blocking question just for my education. :grin:

ui/app/models/job.js
philrenaud: Your read is exactly correct!

I'm reverting this now to make the /versions fetch explicitly happen because of the route model, but the theory was that since my controller is doing the work of hitting /versions anyway, I can assume that I've already got it. The `watchRelationship` hook would fire on query unblock anyway.

However, after testing in shoddy-internet situations, this results in a bad experience where everything-but-versions on the page could load. Moved it back to async: true in order for the watcher to force resolution before page load.

mismithhisler (approved):
Works great!

Issue #23852, opened 2024-08-22
nomad job start feature would be really useful 


mindaugaszilionis:
Hi there,
i know that similar request were already opened [https://github.com/hashicorp/nomad/issues/18558], but maybe it was not fully understood where it could be useful. We had couple incidents, when after some critical database being down, 100+ service were down. After dependent database were fixed, we had manually through GUI to start jobs one by one. 

### Proposal

Would be really nice to have possibility strat hobs through cli (nomad job start job_name) tand massive restart do using, f.e.
_for service in \`nomad job status|grep dead|awk '{print $1}'\`; do nomad job start $service; done_

### Use-cases

Quicker system restore after major incidents. 

### Attempted Solutions

add such feature using nomad cli or in GUI implemnet feature to start all dead jobs through UI
below.

PR #24150, opened 2024-10-08
start: allow users to call job start command to start stopped jobs

martisah:
This PR implements the `job start` command, which allows for users to call "job start [options] <job>" with a valid job that has been previously stopped and starts the most recent running and stable version up. 

Fixes: https://github.com/hashicorp/nomad/issues/23852

After more investigation, we found that batch job versions do not get set as running after a stop, rather they are set to pending, and because of this, it is hard to nail down another marker to use in order for us to establish if the version we’d like to revert to is stable , so for now, this feature has been implemented and tested for service jobs only.

### DIFF
--- a/command/job_start.go
+++ b/command/job_start.go
@@ -1,0 +1,252 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package command
+
+import (
+	"fmt"
+	"github.com/hashicorp/nomad/api"
+	"github.com/hashicorp/nomad/api/contexts"
+	"github.com/posener/complete"
+	"os"
+	"strings"
+	"sync"
+)
+
+type JobStartCommand struct {
+	Meta
+}
+
+func (c *JobStartCommand) Help() string {
+	helpText := `
+Usage: nomad job start [options] <job>
+Alias: nomad start
+
+  Start an existing stopped job. This command is used to start a previously stopped job's
+  most recent running version up. Upon successful start, an interactive
+  monitor session will start to display log lines as the job starts up its
+  allocations based on its most recent running version. It is safe to exit the monitor
+  early using ctrl+c.
+
+  When ACLs are enabled, this command requires a token with the 'submit-job'
+  and 'read-job' capabilities for the job's namespace. The 'list-jobs'
+  capability is required to run the command with job prefixes instead of exact
+  job IDs.
+
+
+General Options:
+
+  ` + generalOptionsUsage(usageOptsDefault) + `
+
+Start Options:
+
+  -detach
+    Return immediately instead of entering monitor mode. After the
+    job start command is submitted, a new evaluation ID is printed to the
+    screen, which can be used to examine the evaluation using the eval-status
+    command.
+
+  -consul-token
+   The Consul token used to verify that the caller has access to the Service
+   Identity policies associated in the targeted version of the job.
+
+  -vault-token
+   The Vault token used to verify that the caller has access to the Vault
+   policies in the targeted version of the job.
+
+  -verbose
+    Display full information.
+`
+	return strings.TrimSpace(helpText)
+}
+
+func (c *JobStartCommand) Synopsis() string {
+	return "Start a stopped job"
+}
+
+func (c *JobStartCommand) AutocompleteFlags() complete.Flags {
+	return mergeAutocompleteFlags(c.Meta.AutocompleteFlags(FlagSetClient),
+		complete.Flags{
+			"-detach":  complete.PredictNothing,
+			"-verbose": complete.PredictNothing,
+		})
+}
+
+func (c *JobStartCommand) AutocompleteArgs() complete.Predictor {
+	return complete.PredictFunc(func(a complete.Args) []string {
+		client, err := c.Meta.Client()
+		if err != nil {
+			return nil
+		}
+
+		resp, _, err := client.Search().PrefixSearch(a.Last, contexts.Jobs, nil)
+		if err != nil {
+			return []string{}
+		}
+		return resp.Matches[contexts.Jobs]
+	})
+}
+func (c *JobStartCommand) Name() string { return "job start" }
+
+func (c *JobStartCommand) Run(args []string) int {
+	var detach, verbose bool
+	var consulToken, vaultToken string
+
+	flags := c.Meta.FlagSet(c.Name(), FlagSetClient)
+	flags.Usage = func() { c.Ui.Output(c.Help()) }
+	flags.BoolVar(&detach, "detach", false, "")
+	flags.BoolVar(&verbose, "verbose", false, "")
+	flags.StringVar(&consulToken, "consul-token", "", "")
+	flags.StringVar(&vaultToken, "vault-token", "", "")
+
+	if err := flags.Parse(args); err != nil {
+		return 1
+	}
+
+	// Check that we got at least one job
+	args = flags.Args()
+	if len(args) < 1 {
+		c.Ui.Error("This command takes at least one argument: <job>")
+		c.Ui.Error(commandErrorText(c))
+		return 1
+	}
+
+	var jobIDs []string
+	for _, jobID := range flags.Args() {
+		jobIDs = append(jobIDs, strings.TrimSpace(jobID))
+	}
+
+	// Get the HTTP client
+	client, err := c.Meta.Client()
+	if err != nil {
+		c.Ui.Error(fmt.Sprintf("Error initializing client: %s", err))
+		return 1
+	}
+
+	statusCh := make(chan int, len(jobIDs))
+	var wg sync.WaitGroup
+
+	for _, jobID := range jobIDs {
+		jobID := jobID
+
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+
+			// Truncate the id unless full length is requested
+			length := shortId
+			if verbose {
+				length = fullId
+			}
+
+			// Check if the job exists and has been stopped
+			jobId, namespace, err := c.JobIDByPrefix(client, jobID, nil)
+			if err != nil {
+				c.Ui.Error(err.Error())
+				statusCh <- 1
+				return
+			}
+			job, err := c.JobByPrefix(client, jobId, nil)
+			if err != nil {
+				c.Ui.Error(err.Error())
+				statusCh <- 1
+				return
+			}
+			if *job.Status != "dead" {
+				c.Ui.Error(fmt.Sprintf("Job  %v has not been stopped and has following status: %v", *job.Name, *job.Status))
+				statusCh <- 1
+				return
+
+			}
+
+			// Get all versions associated to current job
+			q := &api.QueryOptions{Namespace: namespace}
+
+			versions, _, _, err := client.Jobs().Versions(jobID, true, q)
+			if err != nil {
+				c.Ui.Error(fmt.Sprintf("Error retrieving job versions: %s", err))
+				statusCh <- 1
+			}
+
+			// Find the most recent running version for this job
+			var chosenVersion *api.Job
+			var chosenIndex uint64
+			versionAvailable := false
+			for i := len(versions) - 1; i >= 0; i-- {
+				if *versions[i].Status == "running" {
+					chosenVersion = versions[i]
+					chosenIndex = uint64(i)
+					versionAvailable = true
+				}
+
+			}
+			if !versionAvailable {
+				c.Ui.Error(fmt.Sprintf("No previous running versions of job %v,  %s", chosenVersion, err))
+				statusCh <- 1
+				return
+			}
+
+			// Parse the Consul token
+			if consulToken == "" {
+				// Check the environment variable
+				consulToken = os.Getenv("CONSUL_HTTP_TOKEN")
+			}
+
+			// Parse the Vault token
+			if vaultToken == "" {
+				// Check the environment variable
+				vaultToken = os.Getenv("VAULT_TOKEN")
+			}
+
+			// Revert to most recent running version!
+			m := &api.WriteOptions{Namespace: namespace}
+			resp, _, err := client.Jobs().Revert(jobID, chosenIndex, nil, m, consulToken, vaultToken)
+			if err != nil {
+				c.Ui.Error(fmt.Sprintf("Error retrieving job versions: %s, %v", err, chosenIndex))
+				statusCh <- 1
+				return
+			}
+			if *job.Name == "bridge2" {
+				c.Ui.Output(fmt.Sprintf("HERE"))
+
+			}
+
+			// Nothing to do
+			evalCreated := resp.EvalID != ""
+
+			if !evalCreated {
+				statusCh <- 0
+				return
+			}
+
+			if detach {
+				c.Ui.Output("Evaluation ID: " + resp.EvalID)
+				statusCh <- 0
+				return
+			}
+
+			mon := newMonitor(c.Ui, client, length)
+			statusCh <- mon.monitor(resp.EvalID)
+
+		}()
+	}
+	// users will still see
+	// errors if any while we
+	// wait for the goroutines
+	// to finish processing
+	wg.Wait()
+
+	// close the channel to ensure
+	// the range statement below
+	// doesn't go on indefinitely
+	close(statusCh)
+
+	// return a non-zero exit code
+	// if even a single job start fails
+	for status := range statusCh {
+		if status != 0 {
+			return status
+		}
+	}
+	return 0
+}
--- a/command/job_start.go
+++ b/command/job_start.go
@@ -139,7 +139,7 @@ func (c *JobStartCommand) Run(args []string) int {
 				length = fullId
 			}
 
-			// Check if the job exists and has been stopped
+			// Check if the job exists and has been stopped (status is dead)
 			jobId, namespace, err := c.JobIDByPrefix(client, jobID, nil)
 			if err != nil {
 				c.Ui.Error(err.Error())
@@ -153,8 +153,8 @@ func (c *JobStartCommand) Run(args []string) int {
 				return
 			}
 			if *job.Status != "dead" {
-				c.Ui.Error(fmt.Sprintf("Job  %v has not been stopped and has following status: %v", *job.Name, *job.Status))
-				statusCh <- 1
+				c.Ui.Info(fmt.Sprintf("Job  %v has not been stopped and has the following status: %v", *job.Name, *job.Status))
+				statusCh <- 0
 				return
 
 			}
@@ -169,19 +169,18 @@ func (c *JobStartCommand) Run(args []string) int {
 			}
 
 			// Find the most recent running version for this job
-			var chosenVersion *api.Job
 			var chosenIndex uint64
 			versionAvailable := false
 			for i := len(versions) - 1; i >= 0; i-- {
-				if *versions[i].Status == "running" {
-					chosenVersion = versions[i]
+				if *versions[i].Status == "running" && *versions[i].Stop == true {
 					chosenIndex = uint64(i)
 					versionAvailable = true
+					break
 				}
 
 			}
 			if !versionAvailable {
-				c.Ui.Error(fmt.Sprintf("No previous running versions of job %v,  %s", chosenVersion, err))
+				c.Ui.Error(fmt.Sprintf("No previous running versions of job %v,  %s", *job.Name, err))
 				statusCh <- 1
 				return
 			}
@@ -200,9 +199,10 @@ func (c *JobStartCommand) Run(args []string) int {
 
 			// Revert to most recent running version!
 			m := &api.WriteOptions{Namespace: namespace}
+
 			resp, _, err := client.Jobs().Revert(jobID, chosenIndex, nil, m, consulToken, vaultToken)
 			if err != nil {
-				c.Ui.Error(fmt.Sprintf("Error retrieving job versions: %s, %v", err, chosenIndex))
+				c.Ui.Error(fmt.Sprintf("Error retrieving job version %v for job %s: %s,", chosenIndex, jobID, err))
 				statusCh <- 1
 				return
 			}
--- a/command/job_start.go
+++ b/command/job_start.go
@@ -172,7 +172,7 @@ func (c *JobStartCommand) Run(args []string) int {
 			var chosenIndex uint64
 			versionAvailable := false
 			for i := len(versions) - 1; i >= 0; i-- {
-				if *versions[i].Status == "running" && *versions[i].Stop == true {
+				if *versions[i].Status == "running" && *versions[i].Stop {
 					chosenIndex = uint64(i)
 					versionAvailable = true
 					break
--- a/command/job_start.go
+++ b/command/job_start.go
@@ -5,16 +5,18 @@ package command
 
 import (
 	"fmt"
-	"github.com/hashicorp/nomad/api"
-	"github.com/hashicorp/nomad/api/contexts"
-	"github.com/posener/complete"
 	"os"
 	"strings"
 	"sync"
+
+	"github.com/hashicorp/nomad/api"
+	"github.com/hashicorp/nomad/api/contexts"
+	"github.com/posener/complete"
 )
 
 type JobStartCommand struct {
 	Meta
+	versionSelected uint64
 }
 
 func (c *JobStartCommand) Help() string {
@@ -23,8 +25,8 @@ Usage: nomad job start [options] <job>
 Alias: nomad start
 
   Start an existing stopped job. This command is used to start a previously stopped job's
-  most recent running version up. Upon successful start, an interactive
-  monitor session will start to display log lines as the job starts up its
+  most recent running version. Upon successful start, an interactive
+  monitor session will start to display log lines as the job starts its
   allocations based on its most recent running version. It is safe to exit the monitor
   early using ctrl+c.
 
@@ -32,7 +34,6 @@ Alias: nomad start
   and 'read-job' capabilities for the job's namespace. The 'list-jobs'
   capability is required to run the command with job prefixes instead of exact
   job IDs.
-
 
 General Options:
 
@@ -126,8 +127,7 @@ func (c *JobStartCommand) Run(args []string) int {
 	statusCh := make(chan int, len(jobIDs))
 	var wg sync.WaitGroup
 
-	for _, jobID := range jobIDs {
-		jobID := jobID
+	for _, jobIDPrefix := range jobIDs {
 
 		wg.Add(1)
 		go func() {
@@ -139,14 +139,7 @@ func (c *JobStartCommand) Run(args []string) int {
 				length = fullId
 			}
 
-			// Check if the job exists and has been stopped (status is dead)
-			jobId, namespace, err := c.JobIDByPrefix(client, jobID, nil)
-			if err != nil {
-				c.Ui.Error(err.Error())
-				statusCh <- 1
-				return
-			}
-			job, err := c.JobByPrefix(client, jobId, nil)
+			job, err := c.JobByPrefix(client, jobIDPrefix, nil)
 			if err != nil {
 				c.Ui.Error(err.Error())
 				statusCh <- 1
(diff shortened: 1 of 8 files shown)


command/job_start.go
```diff
@@ -0,0 +1,248 @@
...
+				statusCh <- 1
+			}
+
+			// Find the most recent running version for this job
+			var chosenVersion *api.Job
+			var chosenIndex uint64
+			versionAvailable := false
+			for i := len(versions) - 1; i >= 0; i-- {
+				if *versions[i].Status == "running" {
+					chosenVersion = versions[i]
```
martisah, line 177: Working on changing this condition to also check that *versions[i].Stable == true && *versions[i].Stop == false, 
which would guarantee we pick the most recent version that has not been stopped and is stable

command/job_start.go
```diff
@@ -0,0 +1,248 @@
...
+				return
+			}
+			job, err := c.JobByPrefix(client, jobId, nil)
+			if err != nil {
+				c.Ui.Error(err.Error())
+				statusCh <- 1
+				return
+			}
+			if *job.Status != "dead" {
+				c.Ui.Error(fmt.Sprintf("Job  %v has not been stopped and has following status: %v", *job.Name, *job.Status))
```
martisah, line 156: Thinking of changing this to be `c.UI.Info` rather than returning an error when someone tries to start a job that's already running, open for any thoughts!

command/job_start.go
```diff
@@ -0,0 +1,248 @@
...
+				c.Ui.Error(fmt.Sprintf("Error retrieving job versions: %s", err))
+				statusCh <- 1
+			}
+
+			// Find the most recent running version for this job
+			var chosenVersion *api.Job
+			var chosenIndex uint64
+			versionAvailable := false
+			for i := len(versions) - 1; i >= 0; i-- {
+				if *versions[i].Status == "running" {
```
martisah, line 176: Working on changing this condition to also check that `*versions[i].Stable == true && *versions[i].Stop == false`,
which would guarantee we pick the most recent version that has not been stopped and is stable.

command/job_start_test.go
```diff
@@ -0,0 +1,200 @@
...
+	code = runCmd.Run([]string{"-address", addr, "-json", jobFile})
+	must.Zero(t, code,
+		must.Sprintf("job stop stdout: %s", ui.OutputWriter.String()),
+		must.Sprintf("job stop stderr: %s", ui.ErrorWriter.String()),
+	)
+
+	code = cmd.Run([]string{"-address=" + addr, jobID})
+	must.One(t, code)
+	out = ui.ErrorWriter.String()
+	must.StrContains(t, out, "has not been stopped and has following status:")
```
martisah, line 90: If we decide to not return an error when a user attempts to start a running job, this final test for misuse might not be necessary here.

command/job_start.go
```diff
@@ -0,0 +1,248 @@
...
+			if err != nil {
+				c.Ui.Error(fmt.Sprintf("Error retrieving job versions: %s", err))
+				statusCh <- 1
+			}
+
+			// Find the most recent running version for this job
+			var chosenIndex uint64
+			versionAvailable := false
+			for i := len(versions) - 1; i >= 0; i-- {
+				if *versions[i].Status == "running" && *versions[i].Stop == true {
```
martisah, line 175: After some investigation, we found that checking the stable flag would lead to some issues since batch job versions are not set to stable, so the check for *versions[i].Status == "stable" was removed.

tgross (commented):
This is looking great @martisah. I've left some comments on implementation details but the overall design is solid.

* Don't forget to run `make cl` to add a changelog item.
* Given how tricky the reverse iteration and selection of versions is, it'd probably be a good idea to have a test that covers the more complex selection scenarios like having job that's stopped and started multiple times.

command/job_start.go
```diff
@@ -0,0 +1,248 @@
...
+	if err != nil {
+		c.Ui.Error(fmt.Sprintf("Error initializing client: %s", err))
+		return 1
+	}
+
+	statusCh := make(chan int, len(jobIDs))
+	var wg sync.WaitGroup
+
+	for _, jobID := range jobIDs {
+		jobID := jobID
```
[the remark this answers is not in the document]
martisah: Oooh I didn't know that! Thanks for catching this!

command/job_start.go
```diff
@@ -0,0 +1,248 @@
...
+			for i := len(versions) - 1; i >= 0; i-- {
+				if *versions[i].Status == "running" && *versions[i].Stop {
+					chosenIndex = uint64(i)
+					versionAvailable = true
+					break
+				}
+
+			}
+			if !versionAvailable {
+				c.Ui.Error(fmt.Sprintf("No previous running versions of job %v,  %s", *job.Name, err))
```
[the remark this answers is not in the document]
martisah: That's right, not sure why I had that in the first place 😆

command/job_start.go
```diff
@@ -0,0 +1,241 @@
...
+			if err != nil {
+				c.Ui.Error(fmt.Sprintf("Error retrieving job versions: %s", err))
+				statusCh <- 1
+			}
+
+			// Find the most recent version for this job that has not been stopped
+			var chosenVersion uint64
+			versionAvailable := false
+			for i := range versions {
+				if !*versions[i].Stop {
```
martisah, line 169: In light of some bug fixing associated to selecting the correct version, I realized that we should actually be selecting the previously non-stopped version. Also I re-evaluated whether we should be specifically looking for a running version, as often the previously non stopped version tends to be pending (in testing).

command/job_start.go
```diff
@@ -0,0 +1,240 @@
...
+func (c *JobStartCommand) Help() string {
+	helpText := `
+Usage: nomad job start [options] <job>
+Alias: nomad start
+
+  Start an existing stopped job. This command is used to start a previously stopped job's
+  most recent version. Upon successful start, an interactive
+  monitor session will start to display log lines as the job starts its
+  allocations based on its most recent version. It is safe to exit the monitor
+  early using ctrl+c.
```
tgross, line 31: Nitpick: the output of these commands are what get printed literally to the terminal. There's no automatic reflowing in `mitchellh/cli`. So we should try to manually reflow these to 80 cols.

command/job_start.go
```diff
@@ -0,0 +1,240 @@
...
+)
+
+type JobStartCommand struct {
+	Meta
+	versionSelected uint64
+}
+
+func (c *JobStartCommand) Help() string {
+	helpText := `
+Usage: nomad job start [options] <job>
```
tgross, line 24: It looks like we don't have any clue here to the user that they can use more than one job ID, or that if they do they all need to be in the same namespace.

command/job_start.go
```diff
@@ -0,0 +1,240 @@
...
+			versions, _, _, err := client.Jobs().Versions(*job.ID, true, q)
+			if err != nil {
+				c.Ui.Error(fmt.Sprintf("Error retrieving job versions: %s", err))
+				statusCh <- 1
+			}
+
+			// Find the most recent version for this job that has not been stopped
+			var chosenVersion uint64
+			versionAvailable := false
+			for i := range versions {
```
tgross, line 168: You can do `for _, version := range versions` here and then not have to dereference by index.

command/job_start.go
tgross: > as often the previously non stopped version tends to be pending (in testing). Is it the expected behavior to always revert to only running versions?

Oh that's a good call. Yeah we shouldn't revert to pending versions because we don't know that they're stable. That'll complicate testing a bit because you'll need to make sure the job can mark itself running.

command/job_start.go
```diff
@@ -0,0 +1,240 @@
...
+				if !*versions[i].Stop {
+					chosenVersion = *versions[i].Version
+					versionAvailable = true
+					break
+				}
+
+			}
+			c.versionSelected = chosenVersion
```
tgross, line 176: We could potentially have multiple jobs we're going to revert, so we'd be overwriting this field from multiple goroutines, which is a race condition. The production code doesn't ever read this value, only tests, so we probably want to get rid of it if we can. 

I guess there's no way in the API to tell after the fact which version we reverted? 

One way to solve this would be to refactor the body of the goroutine into a method that returns the chosen version, and then have the tests call that method directly. But for this it'd be nice to have the tests exercise the goroutines. Maybe this could be a buffered channel that the goroutine writes into and then the test can pull out the value? That'd be safe from data races.

command/job_start.go
```diff
@@ -0,0 +1,240 @@
...
+				return
+			}
+
+			if consulToken == "" {
+				consulToken = os.Getenv("CONSUL_HTTP_TOKEN")
+			}
+
+			if vaultToken == "" {
+				vaultToken = os.Getenv("VAULT_TOKEN")
+			}
```
tgross, line 190: Nitpick: this is not going to change between jobs, so we can lift these two assignments above the loop.

command/job_start_test.go
```diff
@@ -0,0 +1,245 @@
...
+			must.Sprintf("job stop stdout: %s", ui.OutputWriter.String()),
+			must.Sprintf("job stop stderr: %s", ui.ErrorWriter.String()),
+		)
+
+		code = startCmd.Run(args)
+		must.Zero(t, code,
+			must.Sprintf("job start stdout: %s", ui.OutputWriter.String()),
+			must.Sprintf("job start stderr: %s", ui.ErrorWriter.String()),
+		)
+		must.Eq(t, expectedVersions[i], startCmd.versionSelected)
```
tgross, line 219: Looks like you've still got a test failure here.

command/job_start.go
martisah: Ooh I see, I'll give that a shot!

command/job_start_test.go
martisah: Ah I see, I keep passing this test when I run it locally, but it seems to keep failing when I push, could this be because of race conditions I haven't accounted for (similarly to what you mentioned above)?

command/job_start.go
martisah: Do you know of a method I can use to ensure the jobs are running in my tests? I've been trying to wait for the evaluation to succeed after running each command, but that still hasn't proved to work in terms of getting them to be marked running in tests.

tgross: The job should be marked running as soon as at lease one allocation is placed. So as long as the allocations can start you should be able to poll `api/Jobs.Info` for the `running` status. But if the allocations exit then the job will be dead, so you need to make sure they _stay_ running or they could exit before you can poll and you'll get weird test flakes.

The `testJob` uses the [mock driver](https://github.com/hashicorp/nomad/blob/main/contributing/mock-driver.md), so you may want to look at the configuration of the test job to make sure it'll still be up and running.

command/job_start_test.go
```diff
@@ -0,0 +1,311 @@
...
+			t.Fatalf("status code non zero saw %d", newCode)
+		}
+
+		allocs, _, err = client.Jobs().Allocations("job-start-test", true, nil)
+		must.NoError(t, err)
+		waitForAllocRunning(t, client, allocs[0].ID)
+
+		// check the version selected
+		versionSelected := <-startCmd.versionCh
+		must.Eq(t, 0, versionSelected)
```
martisah, line 285: After more investigation, we found that batch job versions do not get set as running after a stop, rather they are set to pending, and because of this, it is hard to nail down another marker to use in order for us to establish if the version we’d like to revert to is stable , so for now, this feature has been implemented and tested for service jobs only.

A key issue I noticed when testing this feature is that when stopping a job for the first time and starting it again, version 0 is set to be “running” and therefore it is selected as the version to revert to (since version 1 is the version that has stop set to true), however, when a job is stopped, version 2 (the most recent version before calling stop) is marked as “dead” so when that job is started again we will be reverting to version 0 again because it is the only running and non-stopped version available, and this will happen for all the next iterations of stop/start commands. One issue that arises is that when version zero is left behind (as after the size of the versions list reaches a limit we take the oldest version out of the versions list), we have no more running and non-stopped versions to revert to. 

For example, if I run a service job, stop it, and start it again, the job will be reverted to version 0, since that is the most recent running and non-stopped version :
Screenshot 2024-12-05 at 12 41 49 PM 1
When I stop this job and start it again, version 0 will once again be selected to revert to, because version 2 has been marked as dead after the second stop.
Screenshot 2024-12-05 at 12 42 49 PM

website/content/docs/commands/job/start.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+## Usage
+
+```shell-session
+nomad job start [options] <job>
+```
+
+The `job start` command requires one input: the job ID. Nomad will start the
+latest version of the job that is not stopped. If the job has not previously
+been stopped, this command will fail.
```
aimeeu, line 34: "Nomad will start the latest version of the job that is not stopped." -> is this correct or do you mean "of the job that is stopped" ??

Also, suggestion based on assumption that job must have Stopped status:

```suggestion
The `nomad job start` command requires the job ID. Nomad starts the latest
version of the job. This command fails if the job does not have a **Stopped**
or `dead` status.
```

website/content/docs/commands/job/start.mdx
```diff
@@ -0,0 +1,84 @@
...
+first the `-vault-token` flag, then the `$VAULT_TOKEN` environment variable.
+Because the vault token used to [run] the targeted job version was not
+persisted, it must be provided to start if the targeted job includes
+Vault policies and the Nomad servers were configured to require [vault policy]
+authentication.
+
+## Usage
+
+```shell-session
+nomad job start [options] <job>
```
aimeeu, line 29:
```suggestion
nomad job start [options] <job_id>
```

Change to use `<job_id>`, which is the syntax that the [Manage Jobs tutorial](https://developer.hashicorp.com/nomad/tutorials/manage-jobs/jobs-version) uses.

website/content/docs/commands/job/start.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+The `job start` command requires one input: the job ID. Nomad will start the
+latest version of the job that is not stopped. If the job has not previously
+been stopped, this command will fail.
+
+When ACLs are enabled, this command requires a token with the `submit-job`
+capability for the job's namespace. The `list-jobs` capability is required to
+run the command with a job prefix instead of the exact job ID. The `read-job`
+capability is required to monitor the resulting evaluation when `-detach` is
+not used.
```
aimeeu, line 40: For easier reading (ie, scanning), I'm trying to figure out a way to present this info not in a paragraph. Thoughts?

```
When Nomad uses ACLs, the `nomad job start` command requires a token in these
scenarios.

| Scenario | Capability   |
| --------- | -------------- |
| Job uses a namespace  | `submit-job` |
| Run the command with a job prefix instead of an exact job ID  | `list-jobs`  |
| Monitor the resulting evaluation when you do not use this command's `detach` option | `read-job`  |
```

or

```
When Nomad uses ACLs, the `nomad job start` command requires a token with one of
the following capabilities based on the specific scenario:

- The `submit-job`capability for the job's namespace
- The `list-jobs` capability to run the command with a job prefix instead of the
  exact job ID
- The `read-job` capability is required to monitor the resulting evaluation when
  you do not use the `-detach`option.
```

website/content/docs/commands/job/start.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+## General Options
+
+@include 'general_options.mdx'
+
+## Start Options
+
+- `-detach`: Return immediately instead of monitoring. A new evaluation ID
+  will be output, which can be used to examine the evaluation using the
+  [eval status] command.
```
aimeeu, line 50:
```suggestion
- `-detach`: Return immediately instead of monitoring. Nomad outputs a new
  evaluation ID, which you can use to examine the evaluation using the
  [eval status] command.
```

website/content/docs/commands/job/start.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+- `-consul-token`: If set, the passed Consul token is sent along with the Start
+  request to the Nomad servers. This overrides the token found in the
+  `$CONSUL_HTTP_TOKEN` environment variable.
```
aimeeu, line 54:
```suggestion
- `-consul-token`: If set, Nomad sends the Consul token with the
  request. This overrides the token found in the `$CONSUL_HTTP_TOKEN`
  environment variable.
```

website/content/docs/commands/job/start.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+- `-vault-token`: If set, the passed Vault token is sent along with the Start
+  request to the Nomad servers. This overrides the token found in the
+  `$VAULT_TOKEN` environment variable.
```
aimeeu, line 58:
```suggestion
- `-vault-token`: If set, Nomad includes the Vault token with the request. This
  overrides the token found in the `$VAULT_TOKEN` environment variable.
```

website/content/docs/commands/job/start.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+- `-verbose`: Show full information.
+
+## Examples
+
+Start a previously stopped job:
```
aimeeu, line 64:
```suggestion
Start a previously stopped job.
```

aimeeu (changes requested):
Thank you so much for creating the docs content! I left some style guide nits as well as some presentation suggestions. Let me know if you have questions.

website/content/docs/commands/job/start.mdx
```diff
@@ -0,0 +1,84 @@
+---
+layout: docs
+page_title: 'Commands: job start'
+description: |
+  The start command is used to start a stopped job.
```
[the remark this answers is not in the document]
mismithhisler: I'd prefer if we just reference a status of `Stopped`, only because a job can have a CLI status of `dead` but not be stopped.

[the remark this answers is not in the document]
mismithhisler: If you explicitly stop a job, the CLI status will be `dead (stopped)`.  Maybe that update should be make to the job statuses docs?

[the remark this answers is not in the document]
aimeeu: We could remove all mention of CLI status from this page if that would confuse the reader... however,  if I use the CLI to query a job's status, and the result is `dead`... are you saying the job status command returns `dead (stopped)`, `dead (failed)`, or `dead (scaled down)` ? That's not clear from the current [CLI job status command page](https://developer.hashicorp.com/nomad/docs/commands/job/status) - `dead (stopped)` is buried in one of the examples, but what statuses the `nomad job status` command returns are not listed in that page. Do you know offhand where that's defined in the code? (saves me looking) 

The Job concept page has a [Job statuses section](https://developer.hashicorp.com/nomad/docs/concepts/job#job-statuses) that maps CLI status to UI status. Based on your comment, sounds like that mapping is not correct so I need to update.

The job status content has this definition of Stopped: "The Stopped status indicates that a user has manually stopped the job. Nomad's garbage collector has not yet removed or purged the job." [link](https://developer.hashicorp.com/nomad/docs/concepts/job#ui-statuses)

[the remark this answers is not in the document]
mismithhisler: At the moment (although maybe we should change this?), failed and scaled down service jobs, and successfully completed batch jobs, will just show `dead`.  Only when a job has manually been stopped does it show `dead (stopped)`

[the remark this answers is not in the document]
aimeeu: Thanks for the explanation! I grabbed this thread and will make a small change to the job statuses section in the Job page.

website/content/docs/commands/job/start.mdx
mismithhisler: Your suggestion makes more sense here, I've updated it in one of my latest commits.

aimeeu (approved):
LGTM - thank you!

aimeeu (approved):
Thanks for adding that reference!

website/content/docs/commands/job/start.mdx
```diff
@@ -0,0 +1,85 @@
...
+layout: docs
+page_title: 'Commands: job start'
+description: |
+  The nomad job start command starts the latest version of an existing job
+  with a status of Stopped.
+---
+
+# Command: job start
+
+The `job start` command is used to start a stopped job.
```
aimeeu, line 11:
```suggestion
```

Somehow we missed deleting L11, which is replaced by L12-13.

website/content/docs/commands/job/start.mdx
```diff
@@ -0,0 +1,85 @@
+---
+layout: docs
+page_title: 'Commands: job start'
```
aimeeu, line 3:
```suggestion
page_title: 'nomad job start command reference'
```

Changed to match the new title format

website/content/docs/commands/job/start.mdx
```diff
@@ -0,0 +1,85 @@
...
+description: |
+  The nomad job start command starts the latest version of an existing job
+  with a status of Stopped.
+---
+
+# Command: job start
```
aimeeu, line 9:
```suggestion
# nomad job start command reference
```

Changed to match the new title format

website/content/docs/commands/job/start.mdx
```diff
@@ -0,0 +1,85 @@
...
+==> 2025-02-11T15:34:48-05:00: Monitoring evaluation "8b715538"
+    2025-02-11T15:34:48-05:00: Evaluation triggered by job "example"
+    2025-02-11T15:34:49-05:00: Evaluation within deployment: "866ca498"
+    2025-02-11T15:34:49-05:00: Allocation "4d050576" created: node "8e17fa4d", group "echo"
+    2025-02-11T15:34:49-05:00: Evaluation status changed: "pending" -> "complete"
+==> 2025-02-11T15:34:49-05:00: Evaluation "8b715538" finished with status "complete"
+```
+
+[eval status]: /nomad/docs/commands/eval/status
+[consul service identity]: /nomad/docs/configuration/consul#allow_unauthenticated
```
aimeeu, line 82: `/nomad/docs/configuration/consul#allow_unauthenticated` I don't see the `allow_unauthenticated` parameter on this page.

website/content/docs/commands/job/start.mdx
```diff
@@ -0,0 +1,85 @@
...
+If your Nomad servers use [consul service identity] authentication and your job
+uses Consul Connect services, you must supply a Consul token. Nomad does not
+persist the Consul token when you [run] a job. Use the `-consul-token` flag to
+pass the token to the `nomad job start` command. Alternately, you may use the
+`$CONSUL_HTTP_TOKEN` environment variable.
+
+If your Nomad servers require [Vault policy] authentication, you must supply a
+Vault token. Nomad does not persist the Vault token when you [run] a job.
+Use the `-vault-token` flag to pass the token to the `nomad job start` command.
+Alternately, you may use the `$VAULT_TOKEN` environment variable.
```
aimeeu, line 24: I am not able to find the "allow_unauthenticated" parameter on the Consul or Vault configuration pages. 

`[consul service identity]: /nomad/docs/configuration/consul#allow_unauthenticated`

`[vault policy]: /nomad/docs/configuration/vault#allow_unauthenticated`


[March 5 commit 5c4d0e9](https://github.com/hashicorp/nomad/commit/5c4d0e923d2d5ae945aa9c41cd77713131023008#diff-b01a4ae1eca8612aa3a690dd715c64062b07aaa4a68ad17a020bc061bee7830e) removed the `allow_unauthenticated` consul config parameter.

allow_unauthenticated parameter removed from Vault config on Feb 28 [commit 726053](https://github.com/hashicorp/nomad/commit/72680531741bdfc40c584159a54aae7a87389c1a)

website/content/docs/commands/job/start.mdx
```diff
@@ -0,0 +1,85 @@
...
+    2025-02-11T15:34:48-05:00: Evaluation triggered by job "example"
+    2025-02-11T15:34:49-05:00: Evaluation within deployment: "866ca498"
+    2025-02-11T15:34:49-05:00: Allocation "4d050576" created: node "8e17fa4d", group "echo"
+    2025-02-11T15:34:49-05:00: Evaluation status changed: "pending" -> "complete"
+==> 2025-02-11T15:34:49-05:00: Evaluation "8b715538" finished with status "complete"
+```
+
+[eval status]: /nomad/docs/commands/eval/status
+[consul service identity]: /nomad/docs/configuration/consul#allow_unauthenticated
+[vault policy]: /nomad/docs/configuration/vault#allow_unauthenticated
```
aimeeu, line 83: The allow_unauthenticated parameter has been removed. Refer to my L15-24 comment.

aimeeu (changes requested):
Concerned that the start.mdx refers to Consul and Vault `allow_unauthenticated` config parameter that no longer exists.

command/job_start_test.go
```diff
@@ -0,0 +1,151 @@
...
+	t.Run("fails if client request fails", func(t *testing.T) {
+		ui := cli.NewMockUi()
+		cmd := &JobStartCommand{
+			Meta: Meta{
+				Ui: ui,
+			},
+		}
+
+		if code := cmd.Run([]string{"-address=nope", "foo"}); code != 1 {
+			t.Fatalf("expected exit code 1, got: %d", code)
```
mismithhisler, line 92: These should probably use the must package

website/content/docs/commands/job/start.mdx
mismithhisler: Good catch, thanks!

command/job_start.go
```diff
@@ -0,0 +1,174 @@
...
+	helpText := `
+Usage: nomad job start [options] <job>
+Alias: nomad start
+
+
+ Starts a stopped job. The job must be currently registered with Nomad. Upon
+ successful start, an interactive monitor session will start to display log
+ lines as the job starts its allocations based on its most recent version.
+ It is safe to exit the monitor early using ctrl+c.
+
```
tgross, line 30: I want to make explicit what Nomad is doing here in terms of job versions, as that seems to be a source of confusion for users. I've lifted some of this text from `job run`:

```suggestion
Starts a stopped job. The job must be currently registered. Nomad will create a
new version of the job based on its most recent version. Upon successful start,
Nomad will enter an interactive monitor session. This is useful to watch
Nomad's internals make scheduling decisions and place the submitted work onto
nodes. The monitor will end once job placement is done. It is safe to exit the
monitor early using ctrl+c.
```

(We'll want this to be in the docs too.)

tgross (commented):
This is looking good!

command/job_start.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	var jobName, namespace string
+
+	if job.Name != nil {
+		jobName = *job.Name
+	}
+	if job.Namespace != nil {
+		namespace = *job.Namespace
+	}
```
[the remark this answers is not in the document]
mismithhisler: Although now that I say that, it is in the general options, so that would be very misleading.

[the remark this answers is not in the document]
tgross: Oh right. We need to support the flag so we can look up the the job in correct namespace, but you're right that we don't want to allow changing it. I think you're right that it _should_ be impossible for it to be unset (in supported versions of Nomad... I think pre-1.0 versions could have a nil namespace?). Maybe we should default the `namespace` value to the `-namespace` flag then? Or `structs.DefaultNamespace`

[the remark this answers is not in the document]
mismithhisler: Ok I refreshed myself on this a bit. I actually think that adding the namespace option is probably unnecessary logic.

If I ran a job in namespace `dev` originally, I need to stop it via `nomad job stop -namespace dev`.  And then similarly to stop, if I want to start it again, I need to run `nomad job start -namespace dev`.  This is because the command needs to fetch the job from that namespace in order to re-register it.

Let me know if this makes sense or if we want to change this logic up a bit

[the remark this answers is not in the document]
tgross: Yeah, that makes sense to me :+1:

website/content/docs/commands/job/start.mdx
```diff
@@ -0,0 +1,61 @@
...
+with a status of Stopped. Refer to [Job statuses] for status explanations.
+
+## Usage
+
+```shell-session
+nomad job start [options] <job_id>
+```
+
+The `nomad job start` command requires the job ID. Nomad creates new version of
+the job based on it's most recent version. This command fails if the job is not
```
tgross, line 21:
```suggestion
the job based on its most recent version. This command fails if the job is not
```

tgross (approved):
LGTM!

aimeeu (approved):
Thanks for all the docs content updates!

PR #24157, opened 2024-10-09
jobspec: add a chown option to artifact block

shoenig:
This PR adds a boolean `chown` field to the artifact block.

It indicates whether the Nomad client should `chown` the downloaded files
and directories to be owned by the `task.user`. This is useful for drivers
like `raw_exec` and `exec2` which are subject to the host filesystem user
permissions structure. Before, these drivers might not be able to use or
modify the downloaded artifacts since they would be owned by the root
user on a typical Nomad client configuration.

```hcl
artifact {
  source      = "https://example.org/blah/Hello.java"
  destination = "local/java"
  chown       = true
}
```

In this example the `local/java` directory will recursively become owned by the task's user instead of `root`.

Fixes https://github.com/hashicorp/nomad-driver-exec2/issues/51

### DIFF
--- a/api/tasks.go
+++ b/api/tasks.go
@@ -864,7 +864,7 @@ type TaskArtifact struct {
 	GetterMode     *string           `mapstructure:"mode" hcl:"mode,optional"`
 	GetterInsecure *bool             `mapstructure:"insecure" hcl:"insecure,optional"`
 	RelativeDest   *string           `mapstructure:"destination" hcl:"destination,optional"`
-	Chown          *bool             `mapstructure:"chown" hcl:"chown"`
+	Chown          bool              `mapstructure:"chown" hcl:"chown,optional"`
 }
 
 func (a *TaskArtifact) Canonicalize() {
@@ -883,9 +883,6 @@ func (a *TaskArtifact) Canonicalize() {
 	}
 	if len(a.GetterHeaders) == 0 {
 		a.GetterHeaders = nil
-	}
-	if a.Chown == nil {
-		a.Chown = pointerOf(false)
 	}
 	if a.RelativeDest == nil {
 		switch *a.GetterMode {
--- a/api/tasks.go
+++ b/api/tasks.go
@@ -864,6 +864,7 @@ type TaskArtifact struct {
 	GetterMode     *string           `mapstructure:"mode" hcl:"mode,optional"`
 	GetterInsecure *bool             `mapstructure:"insecure" hcl:"insecure,optional"`
 	RelativeDest   *string           `mapstructure:"destination" hcl:"destination,optional"`
+	Chown          *bool             `mapstructure:"chown" hcl:"chown"`
 }
 
 func (a *TaskArtifact) Canonicalize() {
@@ -882,6 +883,9 @@ func (a *TaskArtifact) Canonicalize() {
 	}
 	if len(a.GetterHeaders) == 0 {
 		a.GetterHeaders = nil
+	}
+	if a.Chown == nil {
+		a.Chown = pointerOf(false)
 	}
 	if a.RelativeDest == nil {
 		switch *a.GetterMode {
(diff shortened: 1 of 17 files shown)


shoenig:
not sure what's up with `test-e2e-vault` but i'm sure it's not related

```
failed to download vault 1.17.6: hc-install: will install vault@1.17.6
failed to install vault@1.17.6: unexpected Content-Type: "application/vnd+hashicorp.releases-api.v1+json"
```

tgross:
> not sure what's up with test-e2e-vault but i'm sure it's not related

I've been seeing it fail intermittently today with that `hc-install` issue. Must be something in the artifact server.

command/agent/job_endpoint.go
```diff
@@ -1426,6 +1426,7 @@ func ApiTaskToStructsTask(job *structs.Job, group *structs.TaskGroup,
 					GetterMode:     *ta.GetterMode,
 					GetterInsecure: *ta.GetterInsecure,
 					RelativeDest:   *ta.RelativeDest,
+					Chown:          *ta.Chown,
```
tgross, line 1429: This panics if `ta.Chown` is nil, which will be the case if we receive a HTTP request outside of the Go API because the zero value for the unset field is now nil instead of false. (But see also my note on the API struct.)

api/tasks.go
```diff
@@ -883,6 +884,9 @@ func (a *TaskArtifact) Canonicalize() {
 	if len(a.GetterHeaders) == 0 {
 		a.GetterHeaders = nil
 	}
+	if a.Chown == nil {
+		a.Chown = pointerOf(false)
+	}
```
tgross, line 889: If we're going to canonicalize it to the zero-value (`false`) anyways, we can make it `bool` with the tag `hcl:"chown,optional"` instead of a `*bool` and avoid downstream nil checking.

shoenig: Good idea! Fixed.

tgross (approved):
LGTM!

ahjohannessen:
@tgross @shoenig 

I got this on Flatcar Linux this morning:

```
failed to download artifact "https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/v0.4.35/grpc_health_probe-linux-amd64": getter subprocess failed: exit status 1: failed to download artifact: Get "https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/v0.4.35/grpc_health_probe-linux-amd64": tls: failed to verify certificate: x509: failed to load system roots and no roots provided; open /etc/ssl/certs/ca-certificates.crt: permission denied
```

Seems something changed with regards to artifact download permissions:

```
tls: failed to verify certificate: x509: failed to load system roots and no roots provided; open /etc/ssl/certs/ca-certificates.crt: permission denied
```

After upgrading to 1.9.1 -> 1.9.3. Temporarily solved it by setting `disable_filesystem_isolation = true`, which is probably not a good idea?

On Fedora CoreOS machines I do not have this issue (yet).

Flatcar info:
```
Flatcar Container Linux by Kinvolk stable 4081.2.0 for VMware
core@app03 ~ $ uname -a
Linux app03 6.6.60-flatcar #1 SMP PREEMPT_DYNAMIC Tue Nov 12 16:20:46 -00 2024 x86_64 Intel(R) Xeon(R) Gold 6138 CPU @ 2.00GHz GenuineIntel GNU/Linux
```

Fedora CoreOS info:
```
Fedora CoreOS 41.20241027.3.0
core@app04:~$ uname -a
Linux app04 6.11.5-300.fc41.x86_64 #1 SMP PREEMPT_DYNAMIC Tue Oct 22 20:11:15 UTC 2024 x86_64 GNU/Linux
```

tgross:
@ahjohannessen the sandbox controls weren't really impacted by this change other than chowning the artifact. You're not even getting to that step. Can you open a new issue please?

PR #24165, opened 2024-10-10
scheduler: fix reconnecting allocations getting rescheduled

mismithhisler:
In scenarios where a disconnected client reconnects with failed replacement allocs, the reconcile strategy for picking the correct allocation was not working.  These changes allow evaluation of multiple replacement allocs, and the stopping of allocs that are in a client terminal status.

### DIFF
--- a/scheduler/reconcile.go
+++ b/scheduler/reconcile.go
@@ -10,6 +10,7 @@ package scheduler
 
 import (
 	"fmt"
+	"slices"
 	"sort"
 	"time"
 
@@ -1192,19 +1193,33 @@ func (a *allocReconciler) reconcileReconnecting(reconnecting allocSet, all alloc
 			continue
 		}
 
+		// A replacement allocation could fail and be replaced with another
+		// so follow the replacements in a linked list style
+		replacements := []string{}
+		nextAlloc := reconnectingAlloc.NextAllocation
+		for {
+			val, ok := all[nextAlloc]
+			if !ok {
+				break
+			}
+			replacements = append(replacements, val.ID)
+			nextAlloc = val.NextAllocation
+		}
+
 		// Find replacement allocations and decide which one to stop. A
 		// reconnecting allocation may have multiple replacements.
 		for _, replacementAlloc := range all {
 
-			// Skip allocations that are not a replacement of the one
-			// reconnecting.
-			isReplacement := replacementAlloc.ID == reconnectingAlloc.NextAllocation
+			// Skip the allocation if it is the reconnecting alloc
+			if replacementAlloc == reconnectingAlloc {
+				continue
+			}
 
-			// Skip allocations that are server terminal.
+			// Skip allocations that are server terminal or not replacements.
 			// We don't want to replace a reconnecting allocation with one that
 			// is or will terminate and we don't need to stop them since they
 			// are already marked as terminal by the servers.
-			if !isReplacement || replacementAlloc.ServerTerminalStatus() {
+			if !slices.Contains(replacements, replacementAlloc.ID) || replacementAlloc.ServerTerminalStatus() {
 				continue
 			}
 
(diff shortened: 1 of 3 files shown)


scheduler/reconcile.go
```diff
@@ -1192,19 +1193,33 @@ func (a *allocReconciler) reconcileReconnecting(reconnecting allocSet, all alloc
...
+				continue
+			}
 
-			// Skip allocations that are server terminal.
+			// Skip allocations that are server terminal or not replacements.
 			// We don't want to replace a reconnecting allocation with one that
 			// is or will terminate and we don't need to stop them since they
 			// are already marked as terminal by the servers.
-			if !isReplacement || replacementAlloc.ServerTerminalStatus() {
+			if !slices.Contains(replacements, replacementAlloc.ID) || replacementAlloc.ServerTerminalStatus() {
```
tgross, line 1222: Should we check both client and server terminal status? If the allocation is client-terminal I don't think we can use it as our choice, because it's stopped already.

```suggestion
			if !slices.Contains(replacements, replacementAlloc.ID) || replacementAlloc.TerminalStatus() {
```

scheduler/reconcile.go
```diff
@@ -1221,9 +1236,9 @@ func (a *allocReconciler) reconcileReconnecting(reconnecting allocSet, all alloc
 					})
 				}
 			} else {
-				// The reconnecting allocation is preferred, so stop this
-				// replacement, but avoid re-stopping stopped allocs
-				if replacementAlloc.ClientStatus != structs.AllocClientStatusFailed {
+				// The reconnecting allocation is preferred, so stop any replacements
+				// that are not in server terminal status or stopped already.
+				if _, ok := stop[replacementAlloc.ID]; !ok {
```
tgross, line 1241: I think you're right that we need to check if it's not in `stop` already because otherwise we could append the `allocStopResult` twice. If `replacementAlloc.ClientTerminal()`, will it already be in the `stop` set so the conditional is safe to drop as you've done here?

tgross (commented):
This is looking great @mismithhisler!

I ran a test coverage report via `go test -v -count=1 -cover -coverprofile=$HOME/cover.out ./scheduler && go tool cover -html=$HOME/cover.out` and found that we're missing coverage in two places in this method. Normally we're not sticklers for test coverage but this is clearly a very tricky area of the code. The two areas are:
* The block starting `if stopReconnecting` ([ref](https://github.com/hashicorp/nomad/pull/24165/files#diff-97e6d5511d191b7f15c08c580b602f02f7d22cfc2879f1a7d0a284c206b1c549R1187-R1194))
* The block where `keepAlloc == replacementAlloc`. ([ref](https://github.com/hashicorp/nomad/pull/24165/files#diff-97e6d5511d191b7f15c08c580b602f02f7d22cfc2879f1a7d0a284c206b1c549R1229-R1237)).

Let's try to add a test that covers these two areas, to make sure we're not missing anything. And don't forget to run `make cl` to add a change log entry!

scheduler/reconcile.go
mismithhisler: @Juanadelacuesta and I discussed this today.  We _can_ but then have to do another check to see if it's client terminal, and if so, stop the alloc.  It's a bit redundant because this scenario is caught when we decide which alloc to keep.

scheduler/reconcile.go
mismithhisler: None of the client terminal allocs were in the `stop` set during any of my testing of reconnecting clients.  They all needed to be stopped here.

scheduler/reconcile_test.go
```diff
@@ -5464,6 +5464,9 @@ func TestReconciler_Disconnected_Client(t *testing.T) {
 					},
 				},
 			},
+			reconcileStrategy: structs.ReconcileOptionBestScore,
+			callPicker:        true,
+			pickResult:        "original",
```
mismithhisler, line 5469: Actually going to remove this, as the name of the test is `stop-original-alloc....`

tgross (approved):
LGTM!

PR #24166, opened 2024-10-10
Consul: improve reliability of deregistration

tgross:
When the local Consul agent receives a deregister request, it performs a pre-flight check using the locally cached ACL token. The agent then sends the request upstream to the Consul servers as part of anti-entropy, using its own token. This requires that the token we use for deregistration is valid even though that's not the token used to write to the Consul server.

There are several cases where the service identity token might no longer exist at the time of deregistration:
* A race condition between the sync and destroying the allocation.
* Misconfiguration of the Consul auth method with a TTL.
* Out-of-band destruction of the token.

Additionally, Nomad's sync with Consul returns early if there are any errors, which means that a single broken token can prevent any other service on the Nomad agent from being registered or deregistered.

Update Nomad's sync with Consul to use the Nomad agent's own Consul token for deregistration, regardless of which token the service was registered with. Accumulate errors from the sync so that they no longer block deregistration of other services.

Fixes: https://github.com/hashicorp/nomad/issues/20159
Ref: https://hashicorp.atlassian.net/browse/NET-10286

---

I've got some documentation improvements coming as well: https://github.com/hashicorp/nomad/pull/24167. In addition to the usual tests, I've run the following scenario locally. Run the tproxy example job and verify that services are registered in the server's catalog:

```
$ nomad job run ./tproxy.nomad.hcl
...

$ consul catalog services
consul
count-api
count-api-sidecar-proxy
count-dashboard
count-dashboard-sidecar-proxy
nomad
nomad-client
```

Delete all the tokens associated with our workload identity login method:

```
for token in $(consul acl token list -format=json | jq -r '.[] | select(.AuthMethod == "nomad-workloads") | .AccessorID'); do consul acl token delete -accessor-id "$token"; done
Token "7900e53b-a649-4686-b258-3dde39bdd07d" deleted successfully
Token "ab02193b-e3f9-9f03-bbf6-77589db9a294" deleted successfully
Token "280719dc-1b55-9dcc-91ea-164d8dfcdf17" deleted successfully
```

Stop the job, and verify that all services are deregistered successfully from the server's catalog:

```
$ nomad job stop countdash

$ consul catalog services
consul
nomad
nomad-client
```

No unexpected errors show up on the Consul or Nomad side now. We do see the following in the logs once the allocation is GC'd, which we'd expect to see because we deleted the token so logging that token out will always fail:

```
    2024-10-10T13:47:16.696-0400 [WARN]  client.alloc_runner: error running destroy hooks: alloc_id=d4f61b00-d7e1-ac47-6ab6-9b5e6cc125ac
  error=
  | 1 error occurred:
  | \t* destroy hook "consul" failed: 1 error occurred:
  | \t* 1 error occurred:
  | \t* Unexpected response code: 403 (rpc error making call: ACL not found)
```

### DIFF
--- a/command/agent/consul/service_client.go
+++ b/command/agent/consul/service_client.go
@@ -22,6 +22,7 @@ import (
 	"github.com/armon/go-metrics"
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/go-set/v3"
 	"github.com/hashicorp/nomad/client/serviceregistration"
 	"github.com/hashicorp/nomad/helper"
@@ -973,8 +974,7 @@ func (c *ServiceClient) merge(ops *operations) {
 func (c *ServiceClient) sync(reason syncReason) error {
 	c.logger.Trace("execute sync", "reason", reason)
 
-	sreg, creg, sdereg, cdereg := 0, 0, 0, 0
-	var err error
+	sreg, creg, sdereg, cdereg, fails := 0, 0, 0, 0, 0
 
 	// Get the list of all namespaces created so we can iterate them.
 	namespaces, err := c.namespacesClient.List()
@@ -1001,8 +1001,10 @@ func (c *ServiceClient) sync(reason syncReason) error {
 	// de-registering services.
 	inProbation := time.Now().Before(c.deregisterProbationExpiry)
 
+	var mErr *multierror.Error // collect errors for individual services/checks
+
 	// Remove Nomad services in Consul but unknown to Nomad.
-	for id := range servicesInConsul {
+	for id, service := range servicesInConsul {
 		if _, ok := c.services[id]; ok {
 			// Known service, skip
 			continue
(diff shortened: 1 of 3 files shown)


command/agent/consul/service_client.go
```diff
@@ -2052,18 +2056,6 @@ func isNomadCheck(id string) bool {
 	return strings.HasPrefix(id, nomadCheckPrefix)
 }
 
-// isOldNomadService returns true if the ID matches an old pattern managed by
-// Nomad.
-//
-// Pre-0.7.1 task service IDs are of the form:
```
tgross, line 2058: Note to reviewers: I was touching all the callers of this anyway... I think it's probably safe to remove pre-0.7.1 backwards compatibility code for service IDs at this point.

shoenig (approved):
LGTM!

command/agent/consul/service_client.go
```diff
@@ -1161,7 +1149,28 @@ func (c *ServiceClient) sync(reason syncReason) error {
 	// Only log if something was actually synced
 	if sreg > 0 || sdereg > 0 || creg > 0 || cdereg > 0 {
```
rboyer, line 1150: Should `fails > 0` be present here as well, or are we just letting the multierror handle that case?

tgross: Good catch. Will fix.

Issue #20185, opened 2024-03-21
Consul ACL token is not recreated if deleted


lgfa29:
### Nomad version

```
Nomad v1.7.6
BuildDate 2024-03-12T07:27:36Z
Revision 594fedbfbc4f0e532b65e8a69b28ff9403eb822e
```

### Issue

When using Consul with workload identities, if the ACL token is deleted from Consul it is never recreated and causes the sync loop to fail and exit early, skipping other updates.

### Reproduction steps

1. Start a Consul agent with ACL enabled.
    ```hcl
    # consul.hcl
    
    acl = {
      enabled                  = true
      default_policy           = "deny"
      enable_token_persistence = true
    }
    ```
    ```
    consul agent -dev -config ./consul.hcl
    ```
1. Bootstrap Consul ACL system.
    ```
    consul acl bootstrap
    ```
1. Start a Nomad dev agent with the following configuration.
    ```hcl
    # nomad.hcl

    consul {
      enabled = true
    
      service_identity {
        aud = ["consul.io"]
        ttl = "1h"
      }
    
      task_identity {
        aud = ["consul.io"]
        ttl = "1h"
      }
    }
    ```
    ```
    CONSUL_HTTP_TOKEN=... nomad agent -dev -config ./nomad.hcl
    ```
1. Configure Consul JWT auth method for Nomad.
    ```
    CONSUL_HTTP_TOKEN=... nomad setup consul -y
    ```
1. Register job with Consul service.
    ```hcl
    # example.nomad.hcl
    
    job "example" {
      group "cache" {
        network {
          port "db" {
            to = 6379
          }
        }
      
        service {
          name = "redis"
          port = "db"
        }
      
        task "redis" {
          driver = "docker"
      
          config {
            image = "redis:7"
            ports = ["db"]
          }
        }
      }
    }
    ```
    ```
    nomad run example.nomad.hcl
    ```
1. Delete the Consul ACL token created for the service.
1. Modify the job service with a non-destructive update and run the job again.
    ```diff
    job "example" {
      group "cache" {
        network {
          port "db" {
            to = 6379
          }
        }

        service {
          name = "redis"
          port = "db"

    +     meta {
    +       test = "1"
    +    }
        }

        task "redis" {
          driver = "docker"

          config {
            image = "redis:7"
            ports = ["db"]
          }
        }
      }
    }
    ```
    ```
    nomad run example.nomad.hcl
    ```
#### Expected Result

A new token is created and the service meta is updated.

[...]

PR #24167, opened 2024-10-10
docs: clarify requirements for Consul token policies and TTLs

tgross:
As of #24166, Nomad agents will use their own token to deregister services and checks from Consul. This returns the deregistration path to the pre-Workload Identity workflow. Expand the documentation to make clear why certain ACL policies are required for clients.

Additionally, we did not explicitly call out that auth methods should not set an expiration on Consul tokens. Nomad does not have a facility to refresh these tokens if they expire. Even if Nomad could, there's no way to re-inject them into Envoy sidecars for Consul Service Mesh without recreating the task anyways, which is what happens today. Warn users that they should not set an expiration.

Closes: https://github.com/hashicorp/nomad/issues/20185 (wontfix)
Ref: https://hashicorp.atlassian.net/browse/NET-10262

### DIFF
--- a/website/content/docs/configuration/consul.mdx
+++ b/website/content/docs/configuration/consul.mdx
@@ -116,9 +116,10 @@ agents.
 
 - `token` `(string: "")` - Specifies the token used to provide a per-request ACL
   token. This option overrides the Consul Agent's default token. If the token is
-  not set here or on the Consul agent, it will default to Consul's anonymous policy,
-  which may or may not allow writes. Will default to the `CONSUL_HTTP_TOKEN`
-  environment variable if set.
+  not set here or on the Consul agent, it defaults to Consul's anonymous policy,
+  which may or may not allow writes. Defaults to the `CONSUL_HTTP_TOKEN`
+  environment variable if set. Nomad cannot refresh this token; if the token is
+  deleted, Nomad is not able to communicate with Consul.
 
 - `verify_ssl` `(bool: true)`- Specifies if SSL peer verification should be used
   when communicating to the Consul API client over HTTPS. Will default to the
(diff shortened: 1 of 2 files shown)


website/content/docs/configuration/consul.mdx
```diff
@@ -116,9 +116,11 @@ agents.
 
 - `token` `(string: "")` - Specifies the token used to provide a per-request ACL
   token. This option overrides the Consul Agent's default token. If the token is
-  not set here or on the Consul agent, it will default to Consul's anonymous policy,
-  which may or may not allow writes. Will default to the `CONSUL_HTTP_TOKEN`
-  environment variable if set.
+  not set here or on the Consul agent, it will default to Consul's anonymous
+  policy, which may or may not allow writes. Will default to the
+  `CONSUL_HTTP_TOKEN` environment variable if set. Nomad cannot refresh this
+  token; if the token is deleted Nomad will not be able to communicate with
```
aimeeu, line 122:
```suggestion
  token; if the token is deleted, Nomad is not able to communicate with
```

website/content/docs/configuration/consul.mdx
```diff
@@ -334,6 +336,58 @@ consul {
...
 ```
 
+### Consul ACL policy for Nomad
+
+Nomad agents need access to Consul in order to register themselves in the
+service catalog and discover other Nomad agents via service discovery for
+automatic clustering. Nomad clients use Consul tokens from Workload Identity to
+register services and checks but need permissions on their own token to
+deregister. Nomad servers also create [configuration
+entries][consul_config_entry] for Consul Service Mesh, so the specific
```
aimeeu, line 346: [configuration entries][consul_config_entry] -> link is broken - need to add the URL entry  

```[consul_config_entry]: /consul/docs/connect/config-entries```

**Review by aimeeu — APPROVED** [review]
Thanks for the update!
a few style guide nits and a broken link

File: website/content/docs/configuration/consul.mdx [hunk 3]
```diff
@@ -334,6 +335,58 @@ consul {
...
+permissions vary slightly between Nomad servers and clients. The following
+Consul ACL policies represent the minimal permissions Nomad servers and clients
+need.
```
shoenig, line 348: should we include the (optional) `kv` permissions on the client?

shoenig (approved):

website/content/docs/configuration/consul.mdx
tgross: The client doesn't read from KV anymore; that's only done via Workload Identity.

PR #24168, opened 2024-10-10
[ui] "Edit from Version" function added to Revert

### LINKED ISSUE #24156: UI - Versions - Revert Version improvements
### Proposal

I propose changing the way the 'Revert Version' function works in the Nomad UI:

image

In it's current form, when I click 'Revert Version' it asks for confirmation and submits the changes.

I would like a new button 'Edit this version in the plan UI' where it would take the version selected and drop it into the 'Plan Job' UI. From here, I could make changes to the job before submitting, these changes could include changing the name of the job(thus cloning from an old version ).

### Use-cases

1. I have an older version of a job I want to return to without affecting the running allocations, I press the button give the job a new name and possibly tweak some parameters before submitting
2. I want to run an old version of the job on a different cluster(maybe a testing cluster), if I click this button when I land on the 'Plan Job' page, I can copy the entire content out to the different cluster without having to reconstruct the job manually from the diff and the current version.

### Attempted Solutions

1. Revert to the version I want to investigate
2. Go copy the definition from the definition page
3. Revert the job back to the pre-reverted state in step 1
4. Modify the job in an external editor
5. Submit the job

The issue with this solution is the active allocations might be destroyed as a result of the 'Revert version' and so the operator has to break services to return to this older version.

philrenaud:
Resolves #24156


aimeeu:
@philrenaud when this is merged, we also need to update the screenshots, and possibly text, in the job versions guide.

ui/app/components/job-version.js
```diff
@@ -108,6 +115,46 @@ export default class JobVersion extends Component {
...
+    } catch (e) {
+      this.args.handleError({
+        level: 'danger',
+        title: 'Could Not Edit from Version',
+      });
+    }
+  }
+
+  @action async cloneAsNewJob() {
+    console.log('cloneAsNewJob');
```
tgross, line 139: Debris from debugging, I think?

```suggestion
```

ui/app/routes/jobs/job/definition.js
```diff
@@ -5,33 +5,67 @@
...
     let variableFlags;
     let variableLiteral;
     try {
-      const specificationResponse = await job.fetchRawSpecification();
+      const specificationResponse = await job.fetchRawSpecification(version);
       specification = specificationResponse?.Source ?? null;
       variableFlags = specificationResponse?.VariableFlags ?? null;
       variableLiteral = specificationResponse?.Variables ?? null;
       format = specificationResponse?.Format ?? 'json';
+      console.log({ specification, variableFlags, variableLiteral, format });
```
tgross, line 68: More debugging?

ui/app/templates/jobs/run/index.hbs
```diff
@@ -6,5 +6,11 @@
 <Breadcrumb @crumb={{hash label="Run" args=(array "jobs.run")}} />
 {{page-title "Run a job"}}
 <section class="section">
+  {{#if (and this.sourceString (not this.disregardNameWarning))}}
+    <Hds::Alert @type="inline" @color="warning" data-test-job-name-warning as |A|>
+      <A.Title>Don't forget to change the job name!</A.Title>
+      <A.Description>Since you're cloning a job version's source as a new job, you'll want to change the job name. Otherwise, this will appear as a new version of the original job, rather than a new job.</A.Description>
+    </Hds::Alert>
+  {{/if}}
```
tgross, line 14: It might be nice to swap out the job name with an empty string to prevent accidental overwrites. Unfortunately we don't a JS parser for HCL2. My first pass at a regexp was something like `^job\s+"(.*)"\s+\{` but some quick testing finds some corner cases, so maybe that's not such a good idea after all. If you've got an obvious way to do this, I'd go for it, otherwise we can skip.

ui/tests/unit/adapters/job-test.js
```diff
@@ -501,6 +501,8 @@ module('Unit | Adapter | Job', function (hooks) {
     assert.equal(request.method, 'GET');
   });
 
+  // TODO: test that version requests work for fetchRawDefinition
```
tgross, line 504: Do we want to resolve this TODO in this PR?

tgross (approved):
LGTM!

ui/app/components/job-version.js
```diff
@@ -108,6 +115,45 @@ export default class JobVersion extends Component {
...
+        queryParams: {
+          sourceString: specificationSourceString,
+        },
+      });
+    } catch (e) {
+      this.args.handleError({
+        level: 'danger',
+        title: 'Could Not Clone as New Job',
+        description: messageForError(e),
+      });
```
philrenaud, line 153: To check: how does this handle for specification-doesnt-exist-but-definition-does?

mismithhisler (approved):
Looks good!

PR #24199, opened 2024-10-14
build: update hc-install to current

tgross:
Installing Vault and Consul from releases.hashicorp.com via `hc-install` has been failing intermittently. Update the `hc-install` binaries to be current and add one retry to downloads for our compat tests so that we can get builds more reliably green while the underlying issue is being debugged.


shoenig (approved):
LGTM

Issue #23388, opened 2024-06-19
Add a preferred address family option for network-interface with more than 1 IP address


guifran001:
### Proposal

Some network-interfaces have an IPv4 and an IPv6 address. When using the  `-network-interface` option on these interfaces, Nomad bound itself to the first address. It does not allow to select which one. At least, On Windows, it is indifferent of which which family is prioritised. For example, in dev mode, it is always the IPv6 address that is bound to `NOMAD_IP_<port>`.

As there are service that does not behave correctly under IPv6 and as in dev mode on Windows, IPv6 is selected even when IPv6 is not activated for other interface, an option to prefer IPv4 over IPv6 address family would be great.

I've submitted a pr where I add an option to prefer an adress family (IPv4 or IPv6)
I added also to nomad interpolated env var `NOMAD_IPV4_<port>` and `NOMAD_IPV6_<port>`to allow to chose at run time.

### Use-cases

1. Really practical in Dev-mode: 

On Windows, in dev-mode, when exposing service on localhost, they are exposed on [::1] even though IPv6 is not activated on private interface. What happen then is that Consul cannot resolve DNS request as it ignores the IPv6 ones.

2. Multi-address interface with services that mis-behave or does not support IPv6

3. #6412 Docker service registration for IPv6 addresses fails

### Attempted Solutions

- Prioritise the IP via `netsh interface ipv6 set prefixpolicy ::ffff:0:0/96 51 0`
- Otherwise, I saw in the code that index 0 is used everywhere.

PR #24203, opened 2024-10-14
Consul Connect over IPv6 (except tproxy)

### LINKED ISSUE #7905: Consul Connect doesn't work with IPv6
### Nomad version
Nomad v0.10.5 (4eb2ca3f8d8786c897bb47878f1c12577011ddd3)

CentOS 7.8
CNI plugins v0.8.4

### Issue
Nomad doesn't add ip6tables rules in case of using CNI plugins and Consul Connect, so forwarding works only for IPv4

### Reproduction steps
- Follow this guide: https://www.nomadproject.io/docs/integrations/consul-connect/
- Check ip6tables rules

gulducat:
Mostly resolves #7905 -- #23882 introduced IPv6 support to Nomad's "bridge" network mode, and this extends that to Consul Connect (which also requires "bridge" mode).  I say "mostly" because Transparent Proxy still does not work (the Consul CNI plugin does not do any `ip6tables` at the moment for its extra functionality).

Along the way, I found that since we were always setting Connect/Envoy's `bind_address` to `"0.0.0.0"`, the user couldn't pick anything else (like I had tried `"::"`).  In particular, even with this PR auto-detecting IPv6, I imagine a user might like to set it to `""` (empty), so that Consul [`proxy-defaults` config](https://developer.hashicorp.com/consul/docs/connect/config-entries/proxy-defaults#examples) can come into play.  I did not add a config option for the client along these lines, but an individual job could set it like so:

```hcl
connect {
  sidecar_service {
    proxy {
      config {
        bind_address = "" # let consul proxy-defaults handle it
      }
```

Or set it to whatever they may like.

---

My preferred way to replicate the behavior is (on a host/network with ipv6 support), enable ipv6 on the Nomad bridge and prefer ipv6 for services on a client (per #23388):

```hcl
client {
  enabled = true

  bridge_network_subnet_ipv6 = "fd00:a110:c8::/120"
  preferred_address_family   = "ipv6"
}
```

Consul can run in dev mode. `consul agent -dev`

Then use the basic countdash example:

```shell-session
$ nomad job init -connect -short
Example job file written to example.nomad.hcl
```

<details><summary><code>example.nomad.hcl</code></summary>

```hcl
job "countdash" {

  group "api" {
    network {
      mode = "bridge"
    }

    service {
      name = "count-api"
      port = "9001"

      connect {
        sidecar_service {}
      }
    }

    task "web" {
      driver = "docker"

      config {
        image          = "hashicorpdev/counter-api:v3"
        auth_soft_fail = true
      }
    }
  }

  group "dashboard" {
    network {
      mode = "bridge"

      port "http" {
        static = 9002
        to     = 9002
      }
    }

    service {
      name = "count-dashboard"
      port = "9002"

      connect {
        sidecar_service {
          proxy {
            upstreams {
              destination_name = "count-api"
              local_bind_port  = 8080
            }
          }
        }
      }
    }

    task "dashboard" {
      driver = "docker"

      env {
        COUNTING_SERVICE_URL = "http://${NOMAD_UPSTREAM_ADDR_count_api}"
      }

      config {
        image          = "hashicorpdev/counter-dashboard:v3"
        auth_soft_fail = true
      }
    }
  }
}
```

</details>

The alloc and service addresses will be ipv6, health checks pass, and the counter counts.

### DIFF
--- a/command/agent/consul/connect_test.go
+++ b/command/agent/consul/connect_test.go
@@ -426,6 +426,16 @@ func TestConnect_connectProxyConfig(t *testing.T) {
 			"foo": "bar",
 		}, 42, structs.AllocInfo{AllocID: "test_alloc2"}))
 	})
+
+	t.Run("bind_address override", func(t *testing.T) {
+		require.Equal(t, map[string]interface{}{
+			"bind_address":     "::",
+			"bind_port":        42,
+			"envoy_stats_tags": []string{"nomad.alloc_id=test_alloc1b"},
+		}, connectProxyConfig(map[string]interface{}{
+			"bind_address": "::",
+		}, 42, structs.AllocInfo{AllocID: "test_alloc1b"}))
+	})
 }
 
 func TestConnect_getConnectPort(t *testing.T) {
--- a/command/agent/consul/connect_test.go
+++ b/command/agent/consul/connect_test.go
@@ -409,7 +409,7 @@ func TestConnect_connectProxyConfig(t *testing.T) {
 	ci.Parallel(t)
 
 	t.Run("nil map", func(t *testing.T) {
-		require.Equal(t, map[string]interface{}{
+		must.Eq(t, map[string]any{
 			"bind_address":     "0.0.0.0",
 			"bind_port":        42,
 			"envoy_stats_tags": []string{"nomad.alloc_id=test_alloc1"},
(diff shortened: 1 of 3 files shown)


command/agent/consul/connect_test.go
```diff
@@ -424,7 +424,29 @@ func TestConnect_connectProxyConfig(t *testing.T) {
...
+		}, 42, structs.AllocInfo{AllocID: "test_alloc2"}, nil))
+	})
+
+	t.Run("bind_address override", func(t *testing.T) {
+		require.Equal(t, map[string]interface{}{
+			"bind_address":     "anything",
+			"bind_port":        42,
+			"envoy_stats_tags": []string{"nomad.alloc_id=test_alloc1b"},
+		}, connectProxyConfig(map[string]interface{}{
+			"bind_address": "anything",
```
tgross, line 436: This kind of invalid input just blows up downstream, right?

command/agent/consul/connect_test.go
```diff
@@ -424,7 +424,29 @@ func TestConnect_connectProxyConfig(t *testing.T) {
 			"envoy_stats_tags": []string{"nomad.alloc_id=test_alloc2"},
 		}, connectProxyConfig(map[string]interface{}{
 			"foo": "bar",
-		}, 42, structs.AllocInfo{AllocID: "test_alloc2"}))
+		}, 42, structs.AllocInfo{AllocID: "test_alloc2"}, nil))
+	})
+
+	t.Run("bind_address override", func(t *testing.T) {
+		require.Equal(t, map[string]interface{}{
```
tgross, line 431: We should probably try to avoid adding new tests using `testify` instead of `shoenig/test`

tgross (approved):

Issue #23668, opened 2024-07-22
Windows -> Raw exec = Broken with 1.8.2


guifran001:
### Nomad version
Nomad v1.8.2
BuildDate 2024-07-16T08:50:09Z
Revision 7f0822c1e4f25907d9f60e2d595411950dd1bd28

### Operating system and Environment details

Windows
driver: raw_exec

### Issue

Jobs are stopping unexpectedly in nomad, but the processes are still alive in Windows, so when nomad tries to restart them, it fails because the port of application is already bound or a lock file is already lock.

It might not be 100% reproductible as we have an automated test that deploys a nomad instance and starts some jobs that never failed in 1.7.3, but now is failing since our update to 1.8.2. The test has been successful once with 1.8.2.

The automated test deploy consul 1.17.2 as both server and client, then nomad 1.8.2 as both server and client, then starts a postgres job and a keycloak job.

From the logs, we can see that postgres and keycloak start, then postgres stops and keycloak stops with Exit Code 2. Then they are stuck in a restart loop as the orignial postgres and keycloak processes are still executing.

From the log, we can see the Exit Code 2 that is usually reserved to file not found error. Could it be related to #23595 ?

### Reproduction steps

80%. The test is using pulumi to run the jobs. Each job is run sequentially

#### Expected Result

postgres and keycloak run fine

#### Actual Result

postgres and keycloak are stuck in a restart loop while a detached process of postgres and one from keycloak started by nomad are still running while nomad try to start new ones.

### Job file (if appropriate)

Will be provided if needed.

### Nomad Client logs (if appropriate)
```
POSTGRES Jobs is scheduled: 
2024-07-19T17:47:48.093Z [INFO]  client.alloc_runner.task_runner: Task event: alloc_id=79de797f-3e61-54a6-ce48-2affc507d636 task=postgres type=Received msg="Task received by client" failed=false
2024-07-19T17:47:48.110Z [INFO]  client.alloc_runner.task_runner: Task event: alloc_id=79de797f-3e61-54a6-ce48-2affc507d636 task=postgres type="Task Setup" msg="Building Task Directory" failed=false
[...]

PR #24214, opened 2024-10-15
windows: set job object for executor and children

tgross:
On Windows, if the `raw_exec` driver's executor exits, the child processes are not also killed. Create a Windows "job object" (not to be confused with a Nomad job) and add the executor to it. Child processes of the executor will inherit the job automatically. When the handle to the job object is freed (on executor exit), the job itself is destroyed and this causes all processes in that job to exit.

Fixes: https://github.com/hashicorp/nomad/issues/23668
Ref: https://learn.microsoft.com/en-us/windows/win32/procthread/job-objects

~Note: automated testing for this will need to be end-to-end but we're not running Windows on nightly anymore.~ I've verified the behavior works as expected using Process Explorer. Killing the executor kills all child processes as expected. Killing the root child process causes the executor to exit (as usual) and that in turn kills any child processes that didn't exit already.

_Edit: added tests, which requires changing what runs in CI as well_

### DIFF
--- a/drivers/rawexec/driver_test.go
+++ b/drivers/rawexec/driver_test.go
@@ -12,7 +12,6 @@ import (
 	"path/filepath"
 	"runtime"
 	"strconv"
-	"sync"
 	"syscall"
 	"testing"
 	"time"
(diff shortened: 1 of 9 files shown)


tgross:
As noted in sidebar conversation I'm going to get `drivers/shared` running on Windows and add a test for this there.

drivers/rawexec/driver_test.go
```diff
@@ -237,103 +236,6 @@ func TestRawExecDriver_StartWait(t *testing.T) {
 	require.NoError(harness.DestroyTask(task.ID, true))
 }
 
-func TestRawExecDriver_StartWaitRecoverWaitStop(t *testing.T) {
```
tgross, line 240: Note to reviewers: this is moved to `driver_unix_test.go` unchanged, including any modernization to use `shoenig/test`, `t.Cleanup`, etc. I'll take a second pass over these files in a separate refactoring PR to do all that.

drivers/shared/executor/executor_test.go
```diff
@@ -59,15 +59,6 @@ var (
 	compute  = topology.Compute()
 )
 
-type testExecCmd struct {
```
tgross, line 62: Note to reviewers: this whole file is marked `!windows` now, so I moved these helpers into `utils_test.go`

drivers/rawexec/driver_windows_test.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+// TestRawExecDriver_ExecutorKill verifies that killing the executor will stop
+// its child processes
+func TestRawExecDriver_ExecutorKill(t *testing.T) {
+	ci.Parallel(t)
+
+	d := newEnabledRawExecDriver(t)
+	harness := dtestutil.NewDriverHarness(t, d)
+	t.Cleanup(func() { harness.Kill() })
```
schmichael, line 28:
```suggestion
	t.Cleanup(harness.Kill)
```

schmichael (approved):

Issue #24171, opened 2024-10-11
Regression in Nomad 1.9.0: Stdin might not be closed in Docker exec


MrSerth:
Since updating to Nomad 1.9.0, I am observing a weird behavior I cannot explain. Below is a bug report, since I would classify the behavior as a release regression from Nomad 1.9.0. The same code works flawlessly in Nomad 1.8.4 and lower.

Potentially, this issue is caused by switching to the native Docker library as done with #23966.

The issue is that in some circumstances (with a relatively short input buffer), an allocation Exec doesn't close stdin, so that the command within the allocation keeps waiting. The attached job file and the Go code were extracted from our production setup. They demonstrate the erroneous behavior, but might still be too large. For now, I didn't manage to reduce the example further.

I tried many different steps to nail down the problem, but couldn't. Any help would be greatly appreciated!

### Nomad version
```
Nomad v1.9.0
BuildDate 2024-10-10T07:13:43Z
Revision 7ad36851ec02f875e0814775ecf1df0229f0a615
```

### Operating system and Environment details
Ubuntu 24.04.1 LTS and macOS 15.0.1

```
# Ubuntu
Docker version 27.3.1, build ce12230
```
```
# macOS
Docker version 27.2.0, build 3ab4256
```

### Issue
When a command in an allocation that waits for stdin is executed, the stream might not be closed as expected. This prevents the command to complete successfully. 

### Reproduction steps
Download the following [reproduction_example.zip](https://github.com/user-attachments/files/17344057/reproduction_example.zip) and execute it with `go run example.go`.

Please note a few things:
- The Nomad server / agent version is relevant, not the library version for the Go script
- There are two "TODO" notes in the reproduction example:
  - The first one is at line 40 where the actual file size will be controlled. A file with 8100 bytes will not be processed correctly, but a file with 8200 bytes will.
  - The second one is at line 71, where the actual buffer size that is used for the `Exec` command is measured. This length is important and triggering the issue: A buffer size of 10,000 or less will not work, a buffer size of 10,000 or more works.
- The same code works without any changes and with either file / buffer size in Nomad 1.8.4. It doesn't work with Nomad 1.9.0 any longer.

#### Expected Result

The reproduction example succeeds:

```
Tar archive created with 9728 bytes
fixed_filename.txt
Command executed with exit code: 0
```

#### Actual Result

The reproduction example hangs:

```
Tar archive created with 9728 bytes
# hang
```

### Job file (if appropriate)

```
[...]

PR #24202, opened 2024-10-14
docker: close response connection once stdin is exhausted

shoenig:
This PR ensures we close the hijacked writer connection once there is no more data going into the stdin buffer of the exec'd command. Without this, the command (tar in this case) would hang forever awaiting more input.

Fixes #24171

Internal: NET-11378

PR #24221, opened 2024-10-15
wip: close the stdin connection

shoenig:
This PR closes the hijacked connection used for the `stdin` input to a command being run via `alloc exec`. Without this, a process may remain running waiting for additional input via `stdin` instead of exiting.

Fixes [#24171](https://github.com/hashicorp/nomad/issues/24171)


shoenig:
Using #24202 instead

Issue #24224, opened 2024-10-15
Docker task driver CPU metrics broken in Nomad 1.9.0


tristanmorgan:
### Nomad version
```
Nomad v1.9.0
BuildDate 2024-10-10T07:13:43Z
Revision 7ad36851ec02f875e0814775ecf1df0229f0a615
```

### Operating system and Environment details
multiple platforms.
`Linux 6.6.40-v8+ aarch64 GNU/Linux`
`Darwin 23.6.0 Darwin Kernel Version 23.6.0 root:xnu-10063.141.1.700.5~1/RELEASE_ARM64_T6020 arm64`

### Issue
The Guage nomad.client.allocs.cpu_total_ticks doesn't show the latest value but a running average?

### Reproduction steps
Run Docker job with an earlier version of Nomad and examine metrics, then update to Nomad 1.9.0 and compare.

#### Expected Result
Similar metrics to the previous version

#### Actual Result

This graph shows the metrics of Nomad 1.8.4 then upgraded to 1.9.0 at 09:00 then restarting services at 11:30 to see to the metrics were broken because of the upgrade.
![metrics-graphs-docker](https://github.com/user-attachments/assets/7cfb9200-6c87-4fcf-b97c-c7f584bec9ea)

tristanmorgan:
probably related to #23966

tgross:
Hi @tristanmorgan! That superficially looks like the bug we fixed during the beta here https://github.com/hashicorp/nomad/pull/24092 but apparently either there's another bug or we missed something in testing. I'll investigate and report back shortly.

tgross:
Hello again, @tristanmorgan! I don't seem to be able to reproduce this bug _quite_ as you've described, but I think the way your metrics are getting presented is obscuring the underlying problem, whihc is that the metrics output has changed significantly for some reason.

I used the following job spec:

<details><summary>jobspec</summary>

```hcl
job "httpd" {

  group "web" {
    network {
      mode = "bridge"
      port "www" {
        to = 8001
      }
    }

    task "http" {

      driver = "docker"

      config {
        image   = "busybox:1"
        command = "httpd"
        args    = ["-vv", "-f", "-p", "8001", "-h", "/local"]
        ports   = ["www"]
      }

      template {
        data        = "<html><div>hello, world</div></html>"
        destination = "${NOMAD_TASK_DIR}/index.html"
      }

      resources {
        cpu    = 100
        memory = 100
      }

    }
  }
}
```

</details>


[...]

PR #24229, opened 2024-10-16
docker: use streaming stats collection to correct CPU stats

tgross:
In #23966 we switched to the official Docker SDK for the `docker` driver. In the process we refactored code around stats collection to use the "one shot" version of stats. Unfortunately this "one shot" stats collection does not include the `PreCPU` stats, which are the stats from the previous read. This breaks the calculation we use to determine CPU ticks, because now we're subtracting 0 from the current value to get the delta.

Switch back to using the streaming stats collection. Add a test that fully exercises the `TaskStats` API.

Fixes: https://github.com/hashicorp/nomad/issues/24224
Ref: https://hashicorp.atlassian.net/browse/NET-11348

---

In addition to new tests here, I've tested manually with the setup described here https://github.com/hashicorp/nomad/issues/24224#issuecomment-2417222664 and we get the expected results now.

### DIFF
--- a/.changelog/24229.txt
+++ b/.changelog/24229.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+docker: Fixed a bug where task CPU stats were reported incorrectly
+```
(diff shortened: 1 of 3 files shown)


pkazmierczak (approved):
LGTM!

Issue #24233, opened 2024-10-17
e2e: create a local private registry for docker and podman tests


pkazmierczak:
24215 tried introducing new docker e2e tests to make sure we don't cause any auth regressions, but sadly this approach was flawed—we can't change Nomad client configs _before_ creating the registry or auth configs. This has to be done via terraform during the provisioning of the cluster (same for podman), not in the tests themselves.

PR #24234, opened 2024-10-17
e2e: fix failing tests due to docker plugin settings

pkazmierczak:
[24215](https://github.com/hashicorp/nomad/pull/24215) introduced a fix for a docker auth bug and added e2e tests to cover it. Sadly, by modifying nomad client config on all clients *and* not setting up the registry *before* we run any tests causes, as you might expect, all tests involving docker to fail. PicardRikerFacepalm.gif

This PR disables the new docker e2e tests and we'll revisit the topic in https://github.com/hashicorp/nomad/issues/24233


jrasell (approved):
LGTM!

Issue #24212, opened 2024-10-15
nomad 1.9.0 fails to negotiate docker API version if endpoint argument is set


waldemarmeier:
Hi there,

it seems that nomad 1.9.0 fails to negotiate the docker API version with the daemon if the endpoint argument is set in the task driver configuration. It breaks the support for the docker task driver.

It correlates with setting the `endpoint` argument in the plugin stanza:
https://github.com/hashicorp/nomad/blob/61dd1f3f1090f8ba74be58afbfacd586f472b8b2/drivers/docker/driver.go#L1943C27-L1943C44

Further, the `DOCKER_API_VERSION` env variable is ignored as well. If this behavior is intentioned it should documented somewhere.

### Nomad version
```
Nomad v1.9.0
BuildDate 2024-10-10T07:13:43Z
Revision 7ad36851ec02f875e0814775ecf1df0229f0a615
```

### Operating system and Environment details
AWS EC2 running amazon linux 2023

```
NAME="Amazon Linux"
VERSION="2023"
ID="amzn"
ID_LIKE="fedora"
VERSION_ID="2023"
PLATFORM_ID="platform:al2023"
PRETTY_NAME="Amazon Linux 2023.5.20241001"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2023"
HOME_URL="https://aws.amazon.com/linux/amazon-linux-2023/"
DOCUMENTATION_URL="https://docs.aws.amazon.com/linux/"
SUPPORT_URL="https://aws.amazon.com/premiumsupport/"
BUG_REPORT_URL="https://github.com/amazonlinux/amazon-linux-2023"
VENDOR_NAME="AWS"
VENDOR_URL="https://aws.amazon.com/"
SUPPORT_END="2028-03-15"
```
```
[root@test]# docker version
Client:
 Version:           25.0.5
 API version:       1.44
 Go version:        go1.22.5
 Git commit:        5dc9bcc
 Built:             Wed Aug 21 00:00:00 2024
 OS/Arch:           linux/amd64
 Context:           default

Server:
 Engine:
  Version:          25.0.6
  API version:      1.44 (minimum version 1.24)
  Go version:       go1.22.5
  Git commit:       b08a51f
  Built:            Wed Aug 21 00:00:00 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.7.20
[...]

PR #24237, opened 2024-10-17
docker: always use API version negotiation when initializing clients

pkazmierczak:
During a refactoring of the docker driver in https://github.com/hashicorp/nomad/pull/23966 we introduced a bug: API version negotiation option was not passed to every new client call. 

Fixes https://github.com/hashicorp/nomad/issues/24212
internal ref: https://hashicorp.atlassian.net/browse/NET-11337

### DIFF
--- a/.changelog/24237.txt
+++ b/.changelog/24237.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+docker: Always negotiate API version when initializing clients
+```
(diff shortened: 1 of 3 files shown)


tgross (approved):
LGTM

I realize it's not super realistic for us to exercise the version negotiation in the unit tests, but can you post here how you tested this fixes #24212 just so we have a note on that for the future?

PR #24240, opened 2024-10-17
[ui] Adds an ellipsis and max width to profile nav token name

philrenaud:
Maxes out the width of the profile dropdown for token name so that it isn't as likely to bump into the searchbox etc.

image

(Note: a better, and eventual, solution to this is to use css grid or flexbox, but today is not that day)


pkazmierczak (approved):
LGTM!

Issue #24231, opened 2024-10-16
(1.9.0) missed heartbeats don't mark nodes down


dxdc:
After upgrading to Nomad v1.9.0 from v1.8.4, all client nodes report as "ready", regardless of their actual "ready" status.

### Nomad version

Nomad v1.9.0
BuildDate 2024-10-10T07:13:43Z
Revision 7ad36851ec02f875e0814775ecf1df0229f0a615

### Operating system and Environment details

Ubuntu 24.04 LTS

### Issue

After upgrading servers from 1.8.4 to 1.9.0, all clients are reporting as "ready" even when they are down.

### Reproduction steps

Not sure if it's needed, but this is the client ACL policy:

```hcl
agent {
  policy = "read"
}

node {
  policy = "write"
}

namespace "*" {
  policy       = "read"
  capabilities = ["submit-job", "dispatch-job", "read-logs", "read-fs", "alloc-exec", "alloc-node-exec"]
}
```

#### Expected Result

- Clients should report as down

#### Actual Result

- Clients always report as ready

### Nomad Server logs (if appropriate)

```
Oct 16 18:51:53 ip-172-31-35-53 nomad[48126]:     2024-10-16T18:51:53.070Z [WARN]  nomad.heartbeat: node TTL expired: node_id=5de1e8af-e595-6bdf-4221-9f01dea50822
Oct 16 18:51:53 ip-172-31-35-53 nomad[48126]:     2024-10-16T18:51:53.070Z [ERROR] nomad.heartbeat: update node status failed: error="Permission denied"
Oct 16 18:51:53 ip-172-31-35-53 nomad[48126]:     2024-10-16T18:51:53.070Z [WARN]  nomad.heartbeat: node TTL expired: node_id=9cb8c3cf-c47c-1a81-7182-c6e94f8f6ab9
Oct 16 18:51:53 ip-172-31-35-53 nomad[48126]:     2024-10-16T18:51:53.070Z [ERROR] nomad.heartbeat: update node status failed: error="Permission denied"
Oct 16 18:51:53 ip-172-31-35-53 nomad[48126]:     2024-10-16T18:51:53.070Z [WARN]  nomad.heartbeat: node TTL expired: node_id=afdec3e7-dab0-e5ec-adc6-61eb98456bbf
Oct 16 18:51:53 ip-172-31-35-53 nomad[48126]:     2024-10-16T18:51:53.070Z [ERROR] nomad.heartbeat: update node status failed: error="Permission denied"
[...]

PR #24241, opened 2024-10-17
heartbeat: use leader's ACL token when failing heartbeat

tgross:
In #23838 we updated the `Node.Update` RPC handler we use for heartbeats to be more strict about requiring node secrets. But when a node goes down, it's the leader that sends the request to mark the node down via `Node.Update` (to itself), and this request was missing the leader ACL needed to authenticate to itself.

Add the leader ACL to the request and update the RPC handler test for disconnected-clients to use ACLs, which would have detected this bug. Also added a note to the `Authenticate` comment about how that authentication path requires the leader ACL.

Fixes: https://github.com/hashicorp/nomad/issues/24231
Ref: https://hashicorp.atlassian.net/browse/NET-11384

### DIFF
--- a/.changelog/24241.txt
+++ b/.changelog/24241.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+heartbeat: Fixed a bug where failed nodes would not be marked down
+```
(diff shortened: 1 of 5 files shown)


shoenig (approved):
LGTM!

tgross:
Forgot changelog. Added that and will merge when CI is green.

Issue #2224, opened 2017-01-20
[Improvement] Configurable working directory for exec/raw_exec


dradtke:
The `exec` and `raw_exec` drivers should be able to configure the working directory just like the `docker` driver can.

dradtke:
I actually just realized that there may be some security implications with this change, but it could be minimized by disallowing traversal outside of the specified directory. It also wouldn't play the best with the [environment](https://www.nomadproject.io/docs/runtime/environment.html) that's set up for a task, but that may or may not be very important.

Ashald:
IIUC, isn't it's something that can be achieved as

```
config {
        command = "/bin/bash"
        args    = [
                "-c",
                "cd /my/work/dir && exec $@",
                "my-executable"
        ]
      }
```
?

kaskavalci:
This worked for me:

```
"Config": {
    "args": [
        "-c",
        "cd /my/work/dir && my-executable arg1 arg2 ..."
    ],
    "command": "/bin/sh"
}
```

srene2795:
How do we propagate kill signal to the application  when we execute binary like this ?

"Config": {
    "args": [
        "-c",
        "cd /my/work/dir && my-executable arg1 arg2 ..."
    ],
    "command": "/bin/sh"
}

dradtke:
I think you can do that by prefixing your executable with `exec`, e.g.

```json
"Config": {
  "args": ["-c", "cd /my/work/dir && exec my-executable arg1 arg2 ..."],
  "command": "/bin/sh"
}
```

The idea is to have the shell process actually replace itself with your executable, rather than create a separate subprocess.

nictan94:
Is this going to be enhanced in the future? Because certain legacy windows application requires to be run with executing directory = .exe directory

pcbl:
Hi @tgross ,
Me again. :-) 
Out of curiosity, do you know if there are any plan regarding this possibility of setting the Working Directory of the raw_exec driver?

We have a case where we need to set(it is about a legacy application which demands to work on a specific dir)... So far we do alternatives libe `cd "path" && run` , but would be very nice if we would be abble to set it from the driver.

BR
Pedro

tgross:
We don't have it on our "coming soon" roadmap. So beyond that I couldn't really say when to expect this sort of thing. We'd definitely be open to a PR for it though!

pcbl:
All clear. thanks for the heads up!

picatz:
[...]

PR #24249, opened 2024-10-17
F add cwd to exec drivers

mismithhisler:
Adds support for setting the working directory in `exec` and `raw_exec` drivers.

Fixes: [#2224](https://github.com/hashicorp/nomad/issues/2224)


tgross (commented):
This is looking great @mismithhisler!

Two big picture items:
* It looks like the ticket didn't include the `java` driver but it's nearly identical to the `exec` driver and uses all the same underlying bits. We should probably add that as well.
* We're referring to this as the "working directory for the task" but we already document a "task working directory" in the [Filesystem](https://developer.hashicorp.com/nomad/docs/concepts/filesystem) docs, which is the directory where Nomad creates the `local/` dir, etc. That directory is the default value of `work_dir`, isn't it? That leads to a few questions:
  * Should we use the `work_dir` as part of the binary search path we describe in the `command` docs?
  * Is it possible to simply replace the value of the `TaskDir` field with the value of `work_dir` if set to avoid having an extra field in our protobufs, or does the executor need to know about the fine-grained difference?
  * How do we disambiguate between the two in our docs?

drivers/rawexec/driver.go
```diff
@@ -355,6 +359,7 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
 		Env:              cfg.EnvList(),
 		User:             cfg.User,
 		TaskDir:          cfg.TaskDir().Dir,
+		WorkDir:          driverConfig.WorkDir,
```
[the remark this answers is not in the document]
mismithhisler: Good catch

drivers/shared/executor/executor.go
```diff
@@ -482,7 +490,7 @@ func (e *UniversalExecutor) ExecStreaming(ctx context.Context, command []string,
 
 	cmd := exec.CommandContext(ctx, command[0], command[1:]...)
 
-	cmd.Dir = "/"
+	cmd.Dir = e.childCmd.Dir
```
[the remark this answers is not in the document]
mismithhisler: Yeah this is the one opinionated change I added here.  But it makes `work_dir` function the same as setting the working directory in a container.  I'll add something to the docs.

mismithhisler:
@tgross These are all good points.
> Two big picture items:
> 
> * It looks like the ticket didn't include the `java` driver but it's nearly identical to the `exec` driver and uses all the same underlying bits. We should probably add that as well.

I'll get this added.

> * We're referring to this as the "working directory for the task" but we already document a "task working directory" in the [Filesystem](https://developer.hashicorp.com/nomad/docs/concepts/filesystem) docs, which is the directory where Nomad creates the `local/` dir, etc. That directory is the default value of `work_dir`, isn't it? That leads to a few questions:
>   
>   * Should we use the `work_dir` as part of the binary search path we describe in the `command` docs?

That's a good suggestion.

>   * Is it possible to simply replace the value of the `TaskDir` field with the value of `work_dir` if set to avoid having an extra field in our protobufs, or does the executor need to know about the fine-grained difference?

The executor uses `TaskDir` to lookup the executable, so if we overwrite it in the driver, we won't be able to search it for the executable.  Unless we want to make the search exclusive?

>   * How do we disambiguate between the two in our docs?

An idea here is to change `task working directory` to just `task directory`. But then there is the environment variable named `NOMAD_TASK_DIR` which still points to the `/local` directory within the task directory, and that would be confusing.

tgross:
> The executor uses TaskDir to lookup the executable, so if we overwrite it in the driver, we won't be able to search it for the executable. Unless we want to make the search exclusive?

Eh, come to think of it having the `work_dir` overwrite `TaskDir` is only going to make it _more_ confusing, so let's keep that at is and just make sure we've documented all the places where the driver will treat them differently.

drivers/shared/executor/executor_linux_test.go
```diff
@@ -62,6 +62,7 @@ func testExecutorCommandWithChroot(t *testing.T) *testExecCmd {
 		"/lib64":            "/lib64",
 		"/usr/lib":          "/usr/lib",
 		"/bin/ls":           "/bin/ls",
+		"/bin/pwd":          "/bin/pwd",
```
mismithhisler, line 65: Not using this anymore, so it can be removed

Juanadelacuesta (approved):
This looks pretty straight forward, nice work! Remember to add the backport label.

drivers/java/test-resources/Hello.java
```diff
@@ -5,7 +5,7 @@
 
 public class Hello {
     public static void main(String[] args) {
-        System.out.println("Hello");
+        System.out.println("Hello, the current working directory is: " + System.getProperty("user.dir"));
```
tgross, line 8: I'm not much of a Java developer, but do we need to have both the class file and the jar committed as blobs in the repo? Or can we get away with just the jar?

tgross (approved):
LGTM (pending my comments on the binary blobs)

drivers/java/test-resources/Hello.java
mismithhisler: I believe so. You can actually look at the jar contents and its essentially just pointing to the class file, which I believe contains the bytecode.

jrasell (approved):
LGTM, thanks @mismithhisler!

PR #24260, opened 2024-10-19
csi demo: Remove Portworx CSI demo docs, add JuiceFS CSI demo docs

kmott:
## Summary

Portworx is no longer supported on Nomad ([ref](https://docs.portworx.com/portworx-enterprise/platform/install-with-other/nomad)), so remove the PWX CSI demo docs, and add JuiceFS.

Screenshot_2024-10-19_10-10-02


tgross (commented):
@kmott thanks for this PR. As you can probably tell the demos here aren't really being well-maintained, so I wonder if this JuiceFS demo might be better suited for the Nomad Pack community registry? https://github.com/hashicorp/nomad-pack-community-registry

tgross:
Going to close in lieu of a Pack

Issue #23969, opened 2024-09-16
process not managed by Nomad killed on restart


yigit-erkoca:
### Nomad version
Nomad v1.7.7
BuildDate 2024-04-16T19:26:43Z
Revision 0f34c85ee63f6472bd2db1e2487611f4b176c70c

### Operating system and Environment details
Windows 10 Enterprise 2016 LTSB

### Issue
Nomad instance is killing another process on the system that is not deployed by nomad. 

I see in the application event viewer log the following log from source "Process Exit Monitor"
2024-09-16 23:20:06 The process 'C:\simulator\BoardSimulator.exe' was terminated by the process 'C:\Program Files\Infra\Nomad\nomad.exe' with termination code 1. The creation time for the exiting process was 0x01db07ac4767577e.

### Reproduction steps
On system reboot, some applications like simulator executables start by other windows services, as well as nomad instance.
This is a sporadic issue, does not happen at every reboot, but maybe estimated to be once in every ~100.

#### Expected Result
Nomad not killing/interfering any other process on the system that is not deployed by using nomad jobs.

#### Actual Result
Nomad is terminating other processes that are running in the system.

### Job file (if appropriate)
There is only one job using raw_exec. We are just setting it up.
[NomadJob.txt](https://github.com/user-attachments/files/17026164/NomadJob.txt)

### Nomad Client logs (if appropriate)
[nomad.log](https://github.com/user-attachments/files/17026084/nomad.log)
The timestamp of the process termination according to the eventlog: 2024-09-16 23:20:06

jrasell:
Hi @yigit-erkoca and thanks for raising this issue. Nomad should certainly not be terminating processes it does not control, however, having looked at the client logs nothing stands out initially. I'll mark this for further investigation and add it to our backlog. If you do get any additional information that might be useful, please add it to this issue.

yigit-erkoca:
> Hi @yigit-erkoca and thanks for raising this issue. Nomad should certainly not be terminating processes it does not control, however, having looked at the client logs nothing stands out initially. I'll mark this for further investigation and add it to our backlog. If you do get any additional information that might be useful, please add it to this issue.

Hello @jrasell, thank you for your attention! Indeed I'm extending the logs to find out the root cause. I have a setup that I can reproduce this issue quite easily. I'll add more details to the issue and let you know. I appreciate the help.

tgross:
[...]

PR #24265, opened 2024-10-21
deps: update go-plugin to 1.6.2

tgross:
During allocation recovery, the go-plugin reattachment workflow checks to see if the PID we stored in client state is still running. If so, we try to connect to that process. If that fails, we kill the process under the presumption it's not working. But during reattachment we don't know that the PID we have is still valid. Which means that the process we're trying to attach to may have exited and a different process has spawned with the same PID. This results in some unrelated process getting silently killed.

Update go-plugin to 1.6.2, which includes a bug fix so we do not kill the process.

Fixes: https://github.com/hashicorp/nomad/issues/23969
Ref: https://hashicorp.atlassian.net/browse/NET-11233

### DIFF
--- a/.changelog/24265.txt
+++ b/.changelog/24265.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+deps: Fixed a bug where restarting Nomad could cause an unrelated process with the same PID as a failed executor to be killed
+```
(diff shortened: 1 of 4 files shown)


tgross:
Build broken by https://github.com/hashicorp/go-plugin/pull/312. Need to fix that up.

shoenig (approved):

PR #24272, opened 2024-10-22
[cli] Fixed an error in job tag unset help text

philrenaud:


command/job_tag_unset.go
```diff
@@ -16,7 +16,7 @@ type JobTagUnsetCommand struct {
 
 func (c *JobTagUnsetCommand) Help() string {
 	helpText := `
-Usage: nomad job tag unset [options] <job> -name <tag>
+Usage: nomad job tag unset [options] <job>
```
aimeeu, line 19:
```suggestion
Usage: nomad job tag unset [options] -name <tag> <job>
```


```
$ nomad job tag unset -name "latest-version" hello-world
Tag "latest-version" removed from job "hello-world"
```

aimeeu (approved):
`-name <tag>` should still be in the Usage, but before `<job>`

Issue #24266, opened 2024-10-21
Version tag HTTP API doesn't tag latest version (docs+api inaccurate)


maxramqvist:
### Nomad version
`Nomad v1.9.1
BuildDate 2024-10-21T09:00:50Z
Revision d9ec23f0c1035401e9df6c64d6ffb8bffc555a5e`

With ACLs enabled and using Nomad namespaces.

### Operating system and Environment details
Ubuntu 22.04, kernel 6.9.3 on x86_64

### Issue
According to the [docs](https://developer.hashicorp.com/nomad/tutorials/manage-jobs/jobs-version#create-a-version-tag):

> Optionally, you can also supply a tag description and a version number. If you do not supply a version number, Nomad tags the latest version.

If I don't supply the version number the _oldest_ version gets tagged instead of the _latest_ version.

### Reproduction steps

1. Nomad doesn't tag the latest version. The version must be specified in the POST body for the tagging to work.
```
curl -H "x-nomad-token: $NOMAD_TOKEN" http://localhost:4646/v1/job/test-service/versions/vLatest/tag\?namespace\=mynamespace -X POST -d '{}'  
{"Name":"vLatest","Description":"","TaggedTime":1729537471780368140,"Index":232,"LastContact":0,"KnownLeader":false,"NextToken":""}
```
![image](https://github.com/user-attachments/assets/bbdc8668-55d9-4cec-bf36-de278f3d441a)

Adding _Version_ to the body tags the version intended.
`curl -H "x-nomad-token: $NOMAD_TOKEN" http://localhost:4646/v1/job/test-service/versions/vActualLatest/tag\?namespace\=mynamespace -X POST -d '{"Version": 3}'` Results in:
`{"Name":"vActualLatest","Description":"","TaggedTime":1729537656801230518,"Index":270,"LastContact":0,"KnownLeader":false,"NextToken":""}
`
Results in the desired version getting tagged:
![image](https://github.com/user-attachments/assets/2d0fad2d-dc25-446e-bc91-e4987b83143a)

2. An empty body isn't accepted, but a body with an empty JSON object `{}` is accepted. No biggie really, just not obvious that its required.

`curl -X POST -H "x-nomad-token: $NOMAD_TOKEN" http://localhost:4646/v1/job/test-service/versions/v3/tag\?namespace\=mynamespace ` results in the response: `Request body is empty`.
Specifying an empty JSON as request body works:
[...]

PR #24273, opened 2024-10-22
Docs: Update CLI job tag unset

aimeeu:
CLI help order was wrong, so updating the CLI command.
I updated the usage to use `[options]`, moved general options into each subcommand but put that list in  `<details>` so the long general options list doesn't take up space. I didn't add the details tags to the included general_options file because that would affect several pages, but not all since use of that include is not consistent throughout the CLI command pages.

Fixes: [CE-739]
Relates to: [#24272] [#24266]

https://nomad-git-ce739-hashicorp.vercel.app/nomad/docs/commands/job/tag

[CE-739]: https://hashicorp.atlassian.net/browse/CE-739?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/content/docs/commands/job/tag.mdx
+++ b/website/content/docs/commands/job/tag.mdx
@@ -22,12 +22,6 @@ following subcommands:
 nomad job tag <subcommand> [options] [args]
 ```
 
-## General options
-
-Use these optional general options with the subcommands.
-
-@include 'general_options.mdx'
-
 ## Apply
 
 Use `job tag apply` to create or modify a version tag.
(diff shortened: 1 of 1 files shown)


website/content/docs/commands/job/tag.mdx
```diff
@@ -69,7 +69,7 @@ Use `nomad job tag unset` to delete a tag from a version. This command requires
 ### Unset usage
 
 ```shell-session
-nomad job tag unset [options] <job_id> -name <tag>
+nomad job tag unset [options]  -name <tag> <job_id>
```
philrenaud, line 72: I think `-name <tag>` is actually covered under options and so can be removed

Issue #23505, opened 2024-07-04
Mesh secured Nomad jobs no longer possible across Datacenters?


neuroserve:
### Nomad version

Nomad v1.8.1
BuildDate 2024-06-19T06:43:57Z
Revision 5022543e4b7b8dcec9df123f86630ae3fdcffbe6

### Operating system and Environment details

All Linux (Debian 11) environment with federated Nomad regions on top of federated Consul datacenters in two separate cloud providers (not connected via vpn only via connect mesh gateways).

```
 +----------------+               +----------------+
 |                |               |                |
 | Nomad Region 1 |---federated---| Nomad Region 2 |
 |                |               |                |
 +----------------+               +----------------+
         |                                 |
 +------------------+           +------------------+
 |                  |           |                  |
 | Consul Cluster 1 |-federated-| Consul Cluster 2 |
 |                  |           |                  |
 +------------------+           +------------------+

 Cloud Provider 1                   Cloud Provider 2

```

### Issue

When trying to run a connect-enabled job, the job will not start. Instead I get the error
`failed to derive SI token: Unexpected response code: 500 (rpc error making call: Local tokens are disabled)`

### Reproduction steps
1. Set up at least two Consul federated datacenters with mesh gateways (that involves activating ACLs, token replication and enabling connect).
2. Set up at least two federated Nomad regions/clusters using the Consul datacenters created before (s. diagram)
3. Try to run a connect-enabled Job 

#### Expected Result

Job should start while Consul creates an SI token in the name of Nomad.

#### Actual Result

Job does not start. Instead you get the error `failed to derive SI token: Unexpected response code: 500 (rpc error making call: Local tokens are disabled)`. 
This article (https://support.hashicorp.com/hc/en-us/articles/17767098217235-How-Nomad-Manages-ACL-Tokens-Polices-for-Consul-Service-Mesh) tells us, that ACL tokens automatically generated by Nomad on behalf of Connect-enabled services are now created in [Local](https://developer.hashicorp.com/consul/docs/security/acl/acl-tokens#token-attributes) rather than Global scope, and are no longer replicated globally.

### Job file (if appropriate)
```
job "egw" {

  datacenters = ["dc1"]
  region = "de-west"
  node_pool = "static-clients"

  group "egw" {
    count = 2
    network {
      mode = "bridge"
    }

    service {
      name = "egw"
      connect {
        gateway {
          proxy {}
          terminating {
            service {
              name = "postgres"
            }
            service {
[...]

PR #24275, opened 2024-10-22
docs: warn about Consul auth method locality

tgross:
The locality of Consul tokens we mint via Workload Identity is governed by the Consul auth method configuration. By default tokens are local to the Consul datacenter, which typically maps 1:1 with a Nomad region. Cluster administrators who need cross-datacenter tokens can get them by setting the locality to global, at the risk of placement problems if the primary DC isn't available.

Ref: https://github.com/hashicorp/consul/issues/21863
Fixes: https://github.com/hashicorp/nomad/issues/23505

### DIFF
--- a/website/content/docs/integrations/consul/acl.mdx
+++ b/website/content/docs/integrations/consul/acl.mdx
@@ -109,10 +109,10 @@ Nomad uses to sign workload identities. With these keys, Consul is able to
 validate their origin and confirm that they were actually created by Nomad.
 
 Nomad cannot recreate Consul tokens that have been deleted. The auth method
-configuration should never set the `MaxTokenTTL` field. Consul tokens will be
+configuration should never set the `MaxTokenTTL` field. Consul tokens are
 local to the Consul datacenter unless you set `TokenLocality: "global"` in the
-auth method. We recommend using local tokens (the default) because global tokens
-require the primary Consul datacenter is available when allocations start.
+auth method. We recommend using local tokens, which is the default. Global tokens
+require that the primary Consul datacenter is available when allocations start.
 
 <CodeBlockConfig highlight="2" filename="auth-method.json">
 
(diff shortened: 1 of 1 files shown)


rboyer (approved):

website/content/docs/integrations/consul/acl.mdx
```diff
@@ -109,7 +109,10 @@ Nomad uses to sign workload identities. With these keys, Consul is able to
 validate their origin and confirm that they were actually created by Nomad.
 
 Nomad cannot recreate Consul tokens that have been deleted. The auth method
-configuration should never set the `MaxTokenTTL` field.
+configuration should never set the `MaxTokenTTL` field. Consul tokens will be
```
aimeeu, line 112:
```suggestion
configuration should never set the `MaxTokenTTL` field. Consul tokens are
```

website/content/docs/integrations/consul/acl.mdx
```diff
@@ -109,7 +109,10 @@ Nomad uses to sign workload identities. With these keys, Consul is able to
...
+local to the Consul datacenter unless you set `TokenLocality: "global"` in the
+auth method. We recommend using local tokens (the default) because global tokens
+require the primary Consul datacenter is available when allocations start.
```
aimeeu, line 115:
```suggestion
require that the primary Consul datacenter is available when allocations start.
```

"that" for clarity

website/content/docs/integrations/consul/acl.mdx
```diff
@@ -109,7 +109,10 @@ Nomad uses to sign workload identities. With these keys, Consul is able to
 validate their origin and confirm that they were actually created by Nomad.
 
 Nomad cannot recreate Consul tokens that have been deleted. The auth method
-configuration should never set the `MaxTokenTTL` field.
+configuration should never set the `MaxTokenTTL` field. Consul tokens will be
+local to the Consul datacenter unless you set `TokenLocality: "global"` in the
+auth method. We recommend using local tokens (the default) because global tokens
```
aimeeu, line 114:
```suggestion
auth method. We recommend using local tokens (the default) because global tokens
```
```suggestion
auth method. We recommend using local tokens, which is the default. Global tokens
```
From the Google style guide "Unfortunately, some readers ignore anything that appears in parentheses, so don't put important information in parentheses if you can help it."

aimeeu (approved):
a few style nits

PR #24276, opened 2024-10-23
bench: benchmark signAllocIdentities

schmichael:
This borders on a microbenchmark of RSA signing, so I'm not sure it's appropriate to commit. Since it's not run by default in tests, there's little harm outside of potential code maintenance in the future (that's probably not worth it).

Perhaps if we choose to merge it I should add a `// This benchmark is of limited usefulness and may be removed in the future` style comment? This is not worth hours of maintenance over the years.

---

RSA signing takes about 0.6ms per task per alloc (on my machine).

This is the best case scenario as the simplified signer benchmarked lacks some coordination logic necessary in the real signer.

Even assuming "real" servers could sign identities in 0.1ms, this is still a potential issue because:
1. This CPU latency is incurred in the plan applier: the place where all of our concurrent scheduling is linearized.
2. Since 0.1ms is per task per alloc, moderately sized jobs of 4 tasks with 25 instances block for 10ms of CPU time. That's a pipeline stall potentially worse than a Raft apply.

Full results on my laptop:

```
 (main)~/src/nomad/nomad$ go test -run XXXX -v -bench=. -benchtime=1000x
goos: linux
goarch: amd64
pkg: github.com/hashicorp/nomad/nomad
cpu: 12th Gen Intel(R) Core(TM) i9-12900H
BenchmarkPlanApply_signAllocIdentities
BenchmarkPlanApply_signAllocIdentities/Noop
BenchmarkPlanApply_signAllocIdentities/Noop-20              1000              2401 ns/op
BenchmarkPlanApply_signAllocIdentities/RSA
BenchmarkPlanApply_signAllocIdentities/RSA-20               1000            645455 ns/op
PASS
ok      github.com/hashicorp/nomad/nomad        0.786s
```

### DIFF
--- a/nomad/plan_apply_test.go
+++ b/nomad/plan_apply_test.go
@@ -4,11 +4,15 @@
 package nomad
 
 import (
+	"crypto/rsa"
+	"crypto/x509"
 	"errors"
 	"reflect"
 	"testing"
 	"time"
 
+	"github.com/go-jose/go-jose/v3"
+	"github.com/go-jose/go-jose/v3/jwt"
 	memdb "github.com/hashicorp/go-memdb"
 	"github.com/hashicorp/nomad/ci"
 	"github.com/hashicorp/nomad/helper/testlog"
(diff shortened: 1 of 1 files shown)


jrasell:
> This indirectly reinforces the work @jrasell is doing for adding identity signing to `GetAllocs` (maybe that was your context here?), because we could potentially move _all_ the signing to `GetAllocs` instead of having the default identity in the hot path of the plan applier and everything else not.

This is exactly what I am looking into at the moment based on an internal comment from @schmichael. I have some local updates to this benchmark and will show the team later today as well as describe the problem a little more based on my research this morning.

schmichael:
Closing this as @jrasell implemented a better benchmark in an internal repo (internal not because there's anything to hide: the results are the same! internal because the benchmark is only to demonstrate the importance of removing the code being benchmarked! no need to clutter up the core repo with ephemeral micro-benchmarks)

PR #24280, opened 2024-10-23
connect: handle grpc_address as gosockaddr/template string

shoenig:
This PR fixes a bug where the `consul.grpc_address` could not be set using
a `go-sockaddr/template` string. This was inconsistent with how we do accept
such strings for `consul.address` values.

### DIFF
--- a/.changelog/24280.txt
+++ b/.changelog/24280.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+connect: Able to accept go-sockaddr address for consul grpc address
+```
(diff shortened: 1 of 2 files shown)


shoenig:
Spot check:

`agent.hcl`

```hcl
consul {
  grpc_address = "{{ GetInterfaceIP \"lo\" }}:8502"
}
```

And `countdash.hcl` launches fine

```
➜ nomad job status
ID         Type     Priority  Status   Submit Date
countdash  service  50        running  2024-10-28T14:01:01Z
```

jrasell (approved):
LGTM, thanks @shoenig!

I think we will need a changelog and backport label (assuming 1.9.x) before this can be merged.

Is it also worth following up with a docs PR to call out this option? I took a look and didn't find that we mention it.

Issue #23508, opened 2024-07-04
Envoy proxy doesn't work correctly in Consul transparent proxy mode.


ruslan-y:
Hi there!

I'm going to describe my problem in detail so a lot of text, logs and configs are expected)

### Nomad version
```
Nomad v1.8.0
BuildDate 2024-05-28T17:38:17Z
Revision 28b82e4b2259fae5a62e2ed47395334bea5a24c4
```
### Consul version
```
Consul v1.19.0
Revision bf0166d8
Build Date 2024-06-12T13:59:10Z
```
### Operating system and Environment details
```
5.10.0-23-amd64 #1 SMP Debian 5.10.179-2 (2023-07-14) x86_64 GNU/Linux
```
 
<details>
<summary>Nomad client config</summary>

```
name = "host"
region = "global"
datacenter = "dc1"

enable_debug = false
disable_update_check = false

bind_addr = "0.0.0.0"
advertise {
    http = "<ip>:4646"
    rpc = "<ip>:4647"
    serf = "<ip>:4648"
}
ports {
    http = 4646
    rpc = 4647
    serf = 4648
}

consul {
    address = "localhost:8500"
    ssl = false
    ca_file = ""
    grpc_ca_file = ""
    cert_file = ""
    key_file = ""
    token = "<secret_token>"
    server_service_name = "nomad-servers"
    client_service_name = "nomad-clients"
    tags = []
    auto_advertise = true
    server_auto_join = true
    client_auto_join = true
}

data_dir = "/var/nomad"

log_level = "INFO"
enable_syslog = true

leave_on_terminate = true
leave_on_interrupt = false

tls {
    http = true
    rpc = true
    ca_file = "/etc/nomad/ssl/nomad-ca.pem"
    cert_file = "/etc/nomad/ssl/client.pem"
    key_file = "/etc/nomad/ssl/client-key.pem"
    rpc_upgrade_mode = false
    verify_server_hostname = "true"
    verify_https_client = "true"
}

acl {
    enabled = true
    token_ttl = "30s"
    policy_ttl = "30s"
    replication_token = ""
}

vault {
    enabled = true
    address = "https://<vault_address>"
    allow_unauthenticated = true
    create_from_role = "nomad-cluster"
    task_token_ttl = ""
    ca_file = ""
    ca_path = ""
    cert_file = ""
    key_file = ""
    tls_server_name = ""
    tls_skip_verify = false
    namespace = ""
}

telemetry {
[...]

PR #24291, opened 2024-10-23
docs: warn about UID overlap between workload and Envoy tproxy

tgross:
When using transparent proxy mode with the `connect` block, the UID of the workload cannot be the same as the UID of the Envoy sidecar (currently 101 in the default Envoy container image).

Fixes: https://github.com/hashicorp/nomad/issues/23508

~I'll follow-up with a second PR for the tutorial.~ Actually, that doesn't fit in the tutorial format.

### DIFF
--- a/website/content/docs/integrations/consul/service-mesh.mdx
+++ b/website/content/docs/integrations/consul/service-mesh.mdx
@@ -159,6 +159,8 @@ Using transparent proxy has several important requirements:
   `client_addr`).
 * The Consul agent must be configured with [`recursors`][] if you want
   allocations to make DNS queries for applications outside the service mesh.
+* Your workload's task cannot use the same [Unix user ID (UID)][uid] as the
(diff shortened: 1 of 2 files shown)


website/content/docs/integrations/consul/service-mesh.mdx
```diff
@@ -159,6 +159,7 @@ Using transparent proxy has several important requirements:
   `client_addr`).
 * The Consul agent must be configured with [`recursors`][] if you want
   allocations to make DNS queries for applications outside the service mesh.
+* Your workload's task cannot use the same UID as the Envoy sidecar proxy.
```
gulducat, line 162: worth linking to the `uid` parameter on the jobspec page, to connect the dots across documents?  it also specifies the default "101" there, which might be a nice detail to know for someone who hits this situation.

gulducat (approved):
lgtm!

PR #24306, opened 2024-10-28
update ndjson links due to domain expiry/resale

finnigja:
An external party reported that the `ndjson.org` domain has expired / been reassigned, and is serving malware which does appear to be the case. There's more context at https://github.com/ndjson/ndjson.github.io/issues/24.

This change updates two `ndjson.org` references in Nomad comments / docs to point to the GitHub repo for the spec, https://github.com/ndjson/ndjson-spec.


jrasell (approved):
LGTM, thanks @finnigja!

Issue #24315, opened 2024-10-29
UI - Cmd/Ctrl + Click bug on Vars table


ChefAustin:
### Nomad version
```
$ nomad version
Nomad v1.9.1+ent
BuildDate 2024-10-21T14:02:49Z
Revision d6cbc0d24773eeeaf9bd7ce6117856be2c05d4a9
```

#### Browser Client Details
macOS 14.7 (23H124)
Google Chrome v130.0.6723.70

### Operating system and Environment details
```
$ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.4 LTS"
```

### Issue
Cmd/Ctrl + Click on Vars table opens target URL in both a new tab and currently active tab.

Same issue as outlined in #23544 and #24277 ; I'm guessing this is likely fixable with similar changes as those made to the jobs controller in https://github.com/hashicorp/nomad/pull/23832

### Reproduction steps
Navigate to `/ui/variables`. Cmd/Ctrl + Click `nomad/jobs` link. 

#### Expected Result
`/ui/variables/path/nomad/jobs` view only opens in new inactive tab.

#### Actual Result
`/ui/variables/path/nomad/jobs` view opens in currently active tab and a new inactive tab.

PR #24316, opened 2024-10-29
[ui] Prevent double-open for cmd+click on vars index links

ChefAustin:
I pretty much duplicated the fix performed in https://github.com/hashicorp/nomad/pull/23832 but modified it for Vars instead of Jobs.

Not a JS guy so I could be way off-base here.

[Hopefully] Resolves https://github.com/hashicorp/nomad/issues/24315

### DIFF
--- a/.changelog/24316.txt
+++ b/.changelog/24316.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fix an issue where cmd+click or ctrl+click would double-open a var
+```
--- a/ui/app/components/variable-paths.js
+++ b/ui/app/components/variable-paths.js
@@ -26,13 +26,27 @@ export default class VariablePathsComponent extends Component {
   }
 
   @action
-  async handleFolderClick(path) {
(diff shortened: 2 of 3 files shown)


ChefAustin:
@philrenaud Perhaps you could provide some guidance here on the build failures I'm encountering?

Happy to fix up whatever is busted I'm just a _skosh_ out of my element here.


_[**Pictured below:** Me being out of my element]_
Big Lebowski Out of Your Element Scene

ChefAustin:
> Looks good, @ChefAustin ! Thanks for submitting it.
> 
> I've nudged the Percy tests (visual diff tests / tell us if something moved when it oughtn't have) which are set not to run on forks by default. They should come back soon and if there are visual discrepancies I'll manually review them (flakes sometimes happen with the passage of time)
> 
> I'd like to ask that you make a changelog entry for this PR — you can run `make cl` from the nomad directory and it should guide you through it.

Thanks, @philrenaud! Changelog entry has been added in bbb1d9c.

PR #24319, opened 2024-10-29
Add Nomad workload information to CNI args

### LINKED ISSUE #23830: Provide namespace / job / group / alloc_id to CNI plugins
To make it easier to write CNI plugins it would be great if nomad could provide namespace / job / group / alloc_id via CNI_ARGS to CNI plugins. While it is already possible to use the alloc id (but only when setting the namespace to '*') to query the nomad API and request this information, it would reduce the number of moving parts massively.

The namespace, job and group names are rather important if the CNI provides network policies, which could be nicely written using that information.

Kubernetes already provides the pod name and namespace via CNI_ARGS, so it is not like we are inventing something new here :) Also nomad allows to set similar information for docker containers which makes it easier for 3rd party tooling to do something with it -- why not also make it nicer for CNI. The possibilities are endless, for instance I could use it to update a DNS Server with the alloc ips etc… The changes required in nomad are probably minimal, the extra bytes in the CNI_ARGS will most certainly not hurt and it might make things easier for people trying to do something with CNI. 

/cc @tgross

apollo13:
Tihs PR adds nomad workload information to CNI_ARGS when a `cni/*` is used (not for the `bridge` network which also uses CNI). This information can be used by CNI plugins to correlate the assigned IP to nomad workloads/allocations (among other things).

Fixes #23830

### DIFF
--- a/.changelog/24319.txt
+++ b/.changelog/24319.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+cni: Add Nomad specific workload information to CNI_ARGS
+```
--- a/client/allocrunner/networking_cni.go
+++ b/client/allocrunner/networking_cni.go
@@ -28,6 +28,7 @@ import (
 	consulIPTables "github.com/hashicorp/consul/sdk/iptables"
 	log "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-set/v3"
+	"github.com/hashicorp/nomad/client/taskenv"
 	"github.com/hashicorp/nomad/helper"
 	"github.com/hashicorp/nomad/helper/envoy"
 	"github.com/hashicorp/nomad/nomad/structs"
@@ -117,6 +118,26 @@ func addCustomCNIArgs(networks []*structs.NetworkResource, cniArgs map[string]st
 	}
 }
 
+func addNomadWorkloadCNIArgs(logger log.Logger, alloc *structs.Allocation, cniArgs map[string]string) {
+	for key, value := range map[string]string{
+		// these are the very same keys that are used to build task env vars
+		taskenv.Region:    alloc.Job.Region, // NOMAD_REGION
+		taskenv.Namespace: alloc.Namespace,  // NOMAD_NAMESPACE
+		taskenv.JobID:     alloc.Job.ID,     // NOMAD_JOB_ID
+		taskenv.GroupName: alloc.TaskGroup,  // NOMAD_GROUP_NAME
+		taskenv.AllocID:   alloc.ID,         // NOMAD_ALLOC_ID
+	} {
+		// job ID and group name may contain ";" but CNI_ARGS are ";"-separated
+		// per the spec, so they may not be used in arg keys or values.
+		if strings.Contains(value, ";") {
+			logger.Warn("Skipping CNI arg because it contains a semicolon",
+				"key", key, "value", value)
+		} else {
+			cniArgs[key] = value
+		}
+	}
+}
+
 // Setup calls the CNI plugins with the add action
 func (c *cniNetworkConfigurator) Setup(ctx context.Context, alloc *structs.Allocation, spec *drivers.NetworkIsolationSpec) (*structs.AllocNetworkStatus, error) {
 	if err := c.ensureCNIInitialized(); err != nil {
@@ -132,6 +153,9 @@ func (c *cniNetworkConfigurator) Setup(ctx context.Context, alloc *structs.Alloc
 	tg := alloc.Job.LookupTaskGroup(alloc.TaskGroup)
 
 	addCustomCNIArgs(tg.Networks, cniArgs)
+
+	// Add NOMAD_* after custom args so it cannot be overridden.
+	addNomadWorkloadCNIArgs(c.logger, alloc, cniArgs)
 
 	portMaps := getPortMapping(alloc, c.ignorePortMappingHostIP)
 
--- a/client/allocrunner/networking_cni_test.go
+++ b/client/allocrunner/networking_cni_test.go
@@ -137,7 +137,12 @@ func TestSetup(t *testing.T) {
 				Address:       "99.99.99.99",
 			},
 			expectArgs: map[string]string{
-				"IgnoreUnknown": "true",
+				"IgnoreUnknown":    "true",
+				"NOMAD_ALLOC_ID":   "7cd08c6c-86c8-0bfa-f7ca-338466447711",
+				"NOMAD_GROUP_NAME": "web",
+				"NOMAD_JOB_ID":     "mock-service",
+				"NOMAD_NAMESPACE":  "default",
+				"NOMAD_REGION":     "global",
 			},
 		},
 		{
@@ -148,7 +153,12 @@ func TestSetup(t *testing.T) {
 				Address:       "99.99.99.99",
 			},
 			expectArgs: map[string]string{
-				"IgnoreUnknown": "true",
+				"IgnoreUnknown":    "true",
+				"NOMAD_ALLOC_ID":   "7cd08c6c-86c8-0bfa-f7ca-338466447711",
+				"NOMAD_GROUP_NAME": "web",
+				"NOMAD_JOB_ID":     "mock-service",
+				"NOMAD_NAMESPACE":  "default",
+				"NOMAD_REGION":     "global",
 			},
 		},
 		{
@@ -174,9 +184,31 @@ func TestSetup(t *testing.T) {
 				Address:       "99.99.99.99",
 			},
 			expectArgs: map[string]string{
-				"IgnoreUnknown": "true",
-				"first_arg":     "example",
-				"new_arg":       "example_2",
+				"IgnoreUnknown":    "true",
+				"first_arg":        "example",
+				"new_arg":          "example_2",
+				"NOMAD_ALLOC_ID":   "7cd08c6c-86c8-0bfa-f7ca-338466447711",
+				"NOMAD_GROUP_NAME": "web",
+				"NOMAD_JOB_ID":     "mock-service",
+				"NOMAD_NAMESPACE":  "default",
+				"NOMAD_REGION":     "global",
+			},
+		},
+		{
+			name: "cni workload with invalid job id and namespace",
+			modAlloc: func(a *structs.Allocation) {
+				a.Job.ID = "this;does;not;work"
+				a.Namespace = "no;chance"
+			},
+			expectResult: &structs.AllocNetworkStatus{
+				InterfaceName: "eth0",
+				Address:       "99.99.99.99",
+			},
+			expectArgs: map[string]string{
+				"IgnoreUnknown":    "true",
+				"NOMAD_ALLOC_ID":   "7cd08c6c-86c8-0bfa-f7ca-338466447711",
+				"NOMAD_GROUP_NAME": "web",
+				"NOMAD_REGION":     "global",
 			},
 		},
 		{
@@ -215,6 +247,11 @@ func TestSetup(t *testing.T) {
 				"IgnoreUnknown":          "true",
 				"extra_arg":              "example",
 				"CONSUL_IPTABLES_CONFIG": `{"ConsulDNSIP":"192.168.1.117","ConsulDNSPort":8600,"ProxyUserID":"101","ProxyInboundPort":9999,"ProxyOutboundPort":15001,"ExcludeInboundPorts":["9002"],"ExcludeOutboundPorts":null,"ExcludeOutboundCIDRs":null,"ExcludeUIDs":null,"NetNS":"/var/run/docker/netns/nonsense-ns","IptablesProvider":null}`,
+				"NOMAD_ALLOC_ID":         "7cd08c6c-86c8-0bfa-f7ca-338466447711",
+				"NOMAD_GROUP_NAME":       "web",
+				"NOMAD_JOB_ID":           "mock-service",
+				"NOMAD_NAMESPACE":        "default",
+				"NOMAD_REGION":           "global",
 			},
 		},
 	}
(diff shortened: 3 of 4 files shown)


gulducat:
Heyhey, thanks for the PR!

One reason we structured the new(ish) [`cni{}` block](https://developer.hashicorp.com/nomad/docs/job-specification/network#cni-args) the way we did, like

```hcl
cni {
  args = {"key": "val", ...}
}
```

was to leave the door open for something like this alongside the explicit user-provided args:

```hcl
cni {
  nomad_args = true
}
```

which would magically transform into some standard args that CNI plugin authors could depend on (like requested in #23830!)

I'm more inclined to provide an opt-in mechanism like that.  What do you think?

Although I haven't written many CNI plugins myself (a couple), I also would prefer separate args instead of a JSON blob, and I think following the same convention as [environment variables](https://developer.hashicorp.com/nomad/docs/runtime/environment#job-related-variables) would be nice too, for consistency.

Following that ^, this task could be accomplished with jobspec mutation at registration time, which would turn `cni{ nomad_args = true}` into

```hcl
cni {
  args = {
    "NOMAD_NAMESPACE": "${NOMAD_NAMESPACE}",
    "NOMAD_JOB_ID": "${NOMAD_JOB_ID}",
    ... etc ...
  }
}
```

and that would get interpolated just as user-provided args do, so wouldn't need changes all the way down in allocrunner.

How does that strike you @apollo13?

tgross:
I think @gulducat is on the right track here, with one suggestion:

> Following that ^, this task could be accomplished with jobspec mutation at registration time

We probably don't need/want to do that mutation at registration time, because we have the values available on the client as part of the `taskenv` interpolation. So we can transform the args right in the cni hook here.

gulducat:
In that case, the new `nomad_args` (is that a good name?) bool would need to be passed from the jobspec through to the allocrunner cni hook.  The PR that introduced cni-args-in-the-jobspec #23538 shows all the plumbing for reference.

Alternatively / additionally, the opt-in toggle could be in `client` config, which would apply to all allocs that land on the node.  The plumbing for that would be different from the jobspec approach.

tgross:
> Alternatively / additionally, the opt-in toggle could be in client config, which would apply to all allocs that land on the node. The plumbing for that would be different from the jobspec approach.

That's a good point. The cluster admin is nominally the owner of things like what CNI plugins are available. Maybe just make it opt-out on the client?

apollo13:
> Although I haven't written many CNI plugins myself (a couple), I also would prefer separate args instead of a JSON blob, and I think following the same convention as [environment variables](https://developer.hashicorp.com/nomad/docs/runtime/environment#job-related-variables) would be nice too, for consistency.

Works for me, I haven't written any yet I just wanted some PR to get the discussion started. I am fine with separate env variables instead.

> That's a good point. The cluster admin is nominally the owner of things like what CNI plugins are available. Maybe just make it opt-out on the client?

Do we win much by making it possible to opt-out at all? I am asking because the more options the more to test and given that we set `IgnoreUnknown` anyways I cannot imagine any CNI plugin having problems with NOMAD_* vars.

apollo13:
Pushed a commit with env variables. Are there any others we should add? Task imo makes no sense since the networking information is shared by all tasks in a group.

gulducat:
Aside from "plugins may not properly follow the spec" I can think of only one Nomad reason to make this opt-out-able (or even opt-in).  Job IDs (and group labels) can have `;` in them, which would break the [`CNI_ARGS` spec](https://github.com/containernetworking/cni/blob/main/SPEC.md#parameters).  This is a valid job!

```hcl
job "semi;colon" {
  type = "batch"
  group "et;tu;groupname" {
    network {
      mode = "bridge"
    }
    task "t" {
      driver = "docker"
      config {
        image = "hello-world"
      }
    }
  }
}

```

```
$ nomad run semicolon.nomad.hcl
...
$ nomad status semi
ID            = semi;colon
...
Allocations
ID        Node ID   Task Group       Version  Desired  Status
1f795fc0  f11e544f  et;tu;groupname  0        run      complete

```

That job would suddenly break if ID/group were sent in CNI_ARGS (bridge CNI would barf; see comment from cni_args PR: https://github.com/hashicorp/nomad/pull/23538#discussion_r1676196194).

Maybe we shouldn't need to protect users from this tomfoolery, and it may be exceedingly rare (or never happen!), but it could cause a bad day for someone.  I would love to be able to say "just change the job ID" but who knows what wacky conventions people have wrapped up in their IDs, so it would be nice to have some kind of opt-out mechanism, either in jobspec or client config.  In this weird example, the "error" is in the spec, only affects this one job, so opting out would make sense to include there.  But people may have plugins that handle it poorly, which would be a client/node concern...

I'm on the fence.  Do y'all have any more thoughts here?

---

As for other args to include, I think this is a nice set to start with, and folks can request more later if the need arises.

client/allocrunner/networking_cni.go
```diff
@@ -133,6 +133,12 @@ func (c *cniNetworkConfigurator) Setup(ctx context.Context, alloc *structs.Alloc
 
 	addCustomCNIArgs(tg.Networks, cniArgs)
 
+	// Add NOMAD_* after custom args so it cannot be overridden.
+	cniArgs["NOMAD_NAMESPACE"] = alloc.Namespace
+	cniArgs["NOMAD_JOB_NAME"] = alloc.Job.Name
```
gulducat, line 138: I think this one should be ID, since Name is primarily ornamental, and may differ from the ID.

```suggestion
	cniArgs["NOMAD_JOB_ID"] = alloc.Job.ID
```

also just occurred to me - do we want to include `NOMAD_REGION` too? since it, along with Namespace and ID, uniquely identify the job globally.

apollo13:
> Job IDs (and group labels) can have ; in them, which would break the [CNI_ARGS spec](https://github.com/containernetworking/cni/blob/main/SPEC.md#parameters). This is a valid job!

Sigh, and according to the spec there seems to be no way to escape that. We could escape it in json though ;)

> Maybe we shouldn't need to protect users from this tomfoolery, and it may be exceedingly rare (or never happen!), but it could cause a bad day for someone

Agreed. Sadly the non-restrictiveness of jobnames has already caused quite a few problems elsewhere.

> so opting out would make sense to include there.

But then we'd need it on the job-level again and not on the client I fear.

Btw, this is completely ugly but mostly a matter of documentation. We *could* encode the semicolon using `\x…` or similar -- this way we'd have an extra parse step in the plugins but it would be safe for everything.

gulducat:
We kicked this around internally, and maybe this is a nice simple solution that puts the onus on job authors to not use semicolons, instead of plugin authors to parse them:  Only set the CNI arg if the value does not contain a `";"`.  Then the code is a simple if condition in the Nomad client, and nothing extra required on the plugin side.

How does that sound to you @apollo13 ?

apollo13:
And what is the plugin supposed to do then? Throw it's hands up in the air or just panic and not assign an IP to the container :dagger: 

Joking aside, I think that would be an even harder sell. If a job author uses ";" in the name and the plugin does something different because it has no name (ie network policies can't be applied) then who is supposed to figure out what went wrong and how? Sounds like a debugging nightmare.

client/allocrunner/networking_cni.go
apollo13: Thinking about it I am not sure, when doing network policies what would you expect it to match on? Maybe simply add both?

gulducat:
If it could degrade gracefully, depending on the plugin, then maybe "do something different", but I would exit with a non-zero exit code and output a message with whatever required info is missing (job ID and/or group name). We'd have documentation specifying these limits, maybe log WARN in Nomad agent logs that we're skipping the value, and if the CNI plugin makes a fuss and fails then that gets surfaced too.

e.g. the existing `bridge` plugin fails if it encounters a spec-breaking `;` and that shows up in Nomad client agent logs:

> [ERROR] allocrunner/alloc_runner.go:383: client.alloc_runner: prerun failed: alloc_id=e42a7676-31f6-ca3c-2d4d-9945e82b3759 error="pre-run hook \"network\" failed: failed to configure networking for alloc: failed to configure network: plugin type=\"bridge\" failed (add): ARGS: invalid pair \"colon\""

and it also shows up as a task event in alloc status:

> Recent Events:
> Time                       Type           Description
> 2024-11-05T14:57:29-05:00  Setup Failure  failed to setup alloc: pre-run hook "network" failed: failed to configure networking for alloc: failed to configure network: plugin type="bridge" failed (add): ARGS: invalid pair "colon"

I'll grant you the log is rather verbose for a short message from the plugin, but it's pretty discoverable, eh?

apollo13:
Works for me, just wanted to be sure that this really what we want. That said, I personally don't have any use for jobnames containing a semicolon and would probably welcome a stricter validation there (yeah backwards compat is not going to allow for that -- though I wonder if it might be possible to make it stricter over multiple releases).

client/allocrunner/networking_cni.go
gulducat: In Nomad internally, only the job ID really "matters" -- the Name doesn't really affect anything, and it is not guaranteed to be unique like ID is (you can run two totally different jobs with the same name). So I would expect and encourage plugin authors to use job ID.

Users can still provide custom `cni_args` in the jobspec and include the name or anything else they'd like, or we can add it here later if there's demand for it, but personally I'd like to keep this default list pretty tight and tidy.

apollo13: Ah yes, good point!

gulducat:
Yeah I think the safe thing for today that will also work for 99.9% (totally guessing at that number) of jobs is to:

* skip values that contain `";"`
* log that we did so
* document this behavior (bottom of [networking/cni doc](https://github.com/hashicorp/nomad/blob/v1.9.1/website/content/docs/networking/cni.mdx#use-a-cni-network-with-a-job) should be good, I think)
* test it

and otherwise always provide these defaults (except job id instead of name).

Bonus discoverability option would communicate directly to users earlier on, too:
* issue a warning at job submission time if any `network.mode = "cni/*"` and job id/group name(s) contain ";"

That all work for you @apollo13 ? You up for implementing these changes? :grin:

apollo13:
Works for me yeah, should be able to implement it as well. No promises on the time-frame though since I am swamped with other work currently and don't want to commit to anything…

apollo13:
@gulducat Found some time after all (weekends and such ;)). Feel free to adjust the PR as needed, especially if it is less work than playing back & forth with me.

client/allocrunner/networking_cni.go
```diff
@@ -117,6 +117,14 @@ func addCustomCNIArgs(networks []*structs.NetworkResource, cniArgs map[string]st
 	}
 }
 
+func addNomadWorkloadCNIArg(key string, value string, cniArgs map[string]string, logger log.Logger) {
+	if strings.Contains(value, ";") {
+		logger.Warn("Skipping CNI arg because it contains a semicolon", "key", key, "value", value)
+		return
+	}
+	cniArgs[key] = value
+}
```
gulducat, line 126:
```suggestion
func addNomadWorkloadCNIArgs(logger log.Logger, alloc *structs.Allocation, cniArgs map[string]string) {
	for key, value := range map[string]string{
		// these are the very same keys that are used to build task env vars
		taskenv.Region:    alloc.Job.Region, // NOMAD_REGION
		taskenv.Namespace: alloc.Namespace,  // NOMAD_NAMESPACE
		taskenv.JobID:     alloc.Job.ID,     // NOMAD_JOB_ID
		taskenv.GroupName: alloc.TaskGroup,  // NOMAD_GROUP_NAME
		taskenv.AllocID:   alloc.ID,         // NOMAD_ALLOC_ID
	} {
		// job ID and group name may contain ";" but CNI_ARGS are ";"-separated
		// per the spec, so they may not be used in arg keys or values.
		if strings.Contains(value, ";") {
			logger.Warn("Skipping CNI arg because it contains a semicolon",
				"key", key, "value", value)
		} else {
			cniArgs[key] = value
		}
	}
}
```

client/allocrunner/networking_cni.go
```diff
@@ -133,6 +141,13 @@ func (c *cniNetworkConfigurator) Setup(ctx context.Context, alloc *structs.Alloc
 
 	addCustomCNIArgs(tg.Networks, cniArgs)
 
+	// Add NOMAD_* after custom args so it cannot be overridden.
+	addNomadWorkloadCNIArg("NOMAD_REGION", alloc.Job.Region, cniArgs, c.logger)
+	addNomadWorkloadCNIArg("NOMAD_NAMESPACE", alloc.Namespace, cniArgs, c.logger)
+	addNomadWorkloadCNIArg("NOMAD_JOB_ID", alloc.Job.ID, cniArgs, c.logger)
+	addNomadWorkloadCNIArg("NOMAD_GROUP_NAME", alloc.TaskGroup, cniArgs, c.logger)
+	addNomadWorkloadCNIArg("NOMAD_ALLOC_ID", alloc.ID, cniArgs, c.logger)
```
gulducat, line 149:
```suggestion
	addNomadWorkloadCNIArgs(c.logger, alloc, cniArgs)
```

.changelog/24319.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+cni: Add nomad specific workload information to CNI_ARGS
```
gulducat, line 2: nit:

```suggestion
cni: Add Nomad specific workload information to CNI_ARGS
```

gulducat (approved):
hey thanks, this looks great!

I'm happy to ship your version as-is, but I toyed around with it a bit to see how it felt. feel free to take or leave my suggestions as you like.

apollo13:
Your changes look good imo, I especially like that we don't have duplicate the env names now.

schmichael (approved):
Thanks @apollo13 for sticking with this and @gulducat for the thorough review.

LGTM, but please update the PR title and description if you get a chance. We refer back to historic PRs often, and having to reread comment threads and linked issues slows down investigations.

apollo13:
title & description update, let me know if anything else is missing.

schmichael:
Thanks @apollo13!

PR #24340, opened 2024-10-31
drivers: move executor process out of v1 task cgroup after process starts

### LINKED ISSUE #23951: rawexec: Allow only running task process in cgroup v1 overrides
### Use-cases
Nomad v1.8.0 (#20481) added the [`cgroup_v1_override`](https://developer.hashicorp.com/nomad/docs/drivers/raw_exec#cgroup_v1_override) and corresponding v2 parameters to allow placing task processes in specific cgroups.

The goal of this feature is to enable users' with precise cgroup requirements for their tasks absolute control.

cgroup v2 uses the [`clone3(2)` `CLONE_INTO_CGROUP`](https://man7.org/linux/man-pages/man2/clone3.2.html) to spawn only the task process in the custom cgroups.

cgroup v1 is not supported and uses the traditional double-fork approach:

1. The nomad agent process forks (1) an intermediary process called the `executor`
2. The `executor` sets up (potentially custom) cgroups and forks (2) the [task command](https://developer.hashicorp.com/nomad/docs/drivers/raw_exec#command).

This has the unfortunate side effect of leaving the `executor` process in the custom groups with the task process and prevents users from having full control over their custom cgroups.

### Proposal

The cgroup v1 behavior should match the cgroup v2 behavior: the `executor` should not be part of the custom cgroup.

A straightforward, *but imperfect,* approach would be for the `executor` to detach from the custom cgroups after forking the child process. Writing the `executor`'s pid to the root cgroups *after* forking the task process would remove it from the tasks' cgroups. However there would be a window of time in which both the task and `executor` were running in the custom cgroup.

An alternative that avoids the race condition may be possible but would significantly complicate Nomad's executor: we could triple fork where the new intermediary process handles setting up and entering cgroups but *exits* after forking the user process.

The executor treats subprocesses exiting as the task exiting, so significant code changes would be required to support this new flow just for cgroup v1 override support.

It may be possible for the new intermediary process to avoid the third fork and exiting in favor of calling [`Exec`](https://pkg.go.dev/golang.org/x/sys/unix#Exec) directly to replace itself with the task's command.

shoenig:
This PR changes the behavior of the raw exec task driver on old cgroups v1
systems such that the executor process is no longer a member of the cgroups
created for the task. Now, the executor process is placed into those
cgroups and starts the task child process (just as before), but now then
exits those cgroups and exists in the nomad parent cgroup. This change
makes the behavior sort of similar to cgroups v2 systems, where we never
have the executor enter the task cgroup to begin with (because we can
directly clone(3) the task process into it).

Fixes #23951

### DIFF
--- a/drivers/shared/executor/executor_universal_linux.go
+++ b/drivers/shared/executor/executor_universal_linux.go
@@ -201,7 +201,7 @@ func (e *UniversalExecutor) enterCG1(statsCgroup, cpusetCgroup string) (runningF
 
 	move := func() error {
 		// move the executor back out
-		for _, iface := range ifaces {
+		for _, iface := range append(ifaces, "cpuset") {
 			err := cgroupslib.WriteNomadCG1(iface, "cgroup.procs", pid)
 			if err != nil {
 				e.logger.Warn("failed to move executor cgroup", "interface", iface, "error", err)
--- a/drivers/shared/executor/executor_universal_linux.go
+++ b/drivers/shared/executor/executor_universal_linux.go
@@ -114,53 +114,73 @@ func (e *UniversalExecutor) statCG(cgroup string) (int, func(), error) {
 	return fd, cleanup, err
 }
 
+// runningFunc is called after task startup and is running.
+//
+// its use case is for moving the executor process out of the task cgroup once
+// the child task process has been started (cgroups v1 only)
+type runningFunc func() error
+
+// cleanupFunc is called after task shutdown
+//
+// its use case is for removing the cgroup from the system once it is no longer
+// being used for running the task
+type cleanupFunc func()
+
 // configureResourceContainer on Linux configures the cgroups to be used to track
 // pids created by the executor
 //
 // pid: pid of the executor (i.e. ourself)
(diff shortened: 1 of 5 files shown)


drivers/shared/executor/executor_universal_linux.go
```diff
@@ -114,53 +114,73 @@ func (e *UniversalExecutor) statCG(cgroup string) (int, func(), error) {
 	return fd, cleanup, err
 }
 
+// runningFunc is called after task startup and is running.
+//
+// its use case is for moving the executor process out of the task cgroup once
+// the child task process has been started (cgroups v1 only)
+type runningFunc func() error
```
Juanadelacuesta, line 121: Why do you return this functions instead of making them methods of the executor?

shoenig: No strong reason, it's just that the implementions are different between cgroups v1 and v2. Seemed like defining them as closures inside already-version specific functions was a bit cleaner than expanding the number of functions defined on the executor.

PR #24346, opened 2024-11-01
docs: add note to reschedule block for update progress deadline.

jrasell:
Even when set to unlimited, reschedules will stop once the updates progess deadline is reached.


philrenaud (approved):
Oh this is helpful (and I don't think I would have remembered this), thanks!

Juanadelacuesta (approved):
Thanks!

PR #24347, opened 2024-11-01
[ui] Signing in with a token explicitly sets the region dropdown activeRegion

### LINKED ISSUE #24337: ui: blank region identifier when logging into federated cluster
Tested locally on macOS but there isn't anything OS dependent in the problem.

```
Nomad v1.9.2-dev+ent
BuildDate 2024-10-30T04:10:42Z
Revision bfd073ce668d07633388ea9dfa11e4919fc72f7b
```

I bootstrapped three regions (two would be fine) in a federated manner using only the authoritative regions bootstrap token for all authentication requirements. When I open the UI and login using a token, the region identification display is blank and remains so until I refresh the browser page.

- I am able to navigate the side menu without this changing
- The drop down does contain the list of regions in the federated cluster

Screenshot 2024-10-30 at 09 08 39

I would expect that once I log in and reach this page, the region identifier is show in the top left drop down.

philrenaud:
Relates to @shoenig's work in https://github.com/hashicorp/nomad/pull/24320 so branching off of that

As it stands, when you sign in with a token in a multi-region cluster, the `system.regions` array exists, but the `system.activeRegion` doesn't re-compute.

This makes that value recompute automatically during token verification, and refreshes the regions list to trigger that.

(Note: I don't think this is an issue for SSO-based authentication, since there is an explicit page refresh, which would handle region-setting by virtue of initial page load)

Resolves #24337

### DIFF
--- a/ui/app/controllers/settings/tokens.js
+++ b/ui/app/controllers/settings/tokens.js
@@ -23,7 +23,7 @@ export default class Tokens extends Controller {
   @service token;
   @service store;
   @service router;
-
+  @service system;
   queryParams = ['code', 'state', 'jwtAuthMethod'];
 
   @tracked secret = this.token.secret;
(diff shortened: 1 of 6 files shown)


jrasell (approved):
LGTM, thanks @philrenaud!

I think it's also worthy of a changelog and backport.

shoenig (approved):

PR #24352, opened 2024-11-01
test: fix rawexec driver unix test imports

mismithhisler:
The `must` package import was missing from the `drivers/rawexec/driver_unix_test.go` file.  This PR adds the import so these tests can build and run.


tgross (approved):
LGTM

PR #24353, opened 2024-11-01
dynamic host volumes: initial state store implementation

tgross:
This changeset implements the state store schema for Dynamic Host Volumes, and methods used to query the state for RPCs.


nomad/state/state_store_host_volumes.go
```diff
@@ -0,0 +1,197 @@
...
+		obj, err := txn.First(TableHostVolumes, indexID, v.Namespace, v.ID)
+		if err != nil {
+			return err
+		}
+		if obj != nil {
+			// Only some properties of a volume can be updated in place.
+			//
+			// TODO(1.10.0): we need to check this in the RPC handler before
+			// sending the ClientRPC, so figure out if we really need to do this
+			// here.
```
pkazmierczak, line 72: Hmm this looks like something an RPC-level validation should catch before it ever reaches the state store call indeed. But I'm ok with an extra check.

nomad/structs/host_volumes.go
```diff
@@ -0,0 +1,197 @@
...
+	// HostPath is the path on disk where the volume's mount point was
+	// created. We record this to make debugging easier.
+	HostPath string
+
+	// State represents the overall state of the volume. One of pending, ready,
+	// deleted.
+	State HostVolumeState
+
+	CreateIndex uint64
+	CreateTime  int64
```
pkazmierczak, line 72: add a comment that this is unixnano? Just to make life easier for people, we're not consistent about this...

pkazmierczak (approved):
LGTM! Left some minor comments/questions, but by all means :shipit:

nomad/structs/host_volumes.go
tgross: Good call, will do.

nomad/state/schema.go
```diff
@@ -41,6 +42,7 @@ const (
 	indexName          = "name"
 	indexSigningKey    = "signing_key"
 	indexAuthMethod    = "auth_method"
+	indexNodePool      = "node_pool"
```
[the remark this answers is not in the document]
tgross: This just made me realize we almost certainly _do_ have it elsewhere but weren't referring to it by constant. I've fixed up the schema file to use `indexNodePool` and `indexNodeID` where possible.

PR #24356, opened 2024-11-01
dynamic host volumes: ACL policies

tgross:
This changeset implements the ACLs required for dynamic host volumes RPCs:

* `host-volume-write` is a coarse-grained policy that implies all operations.
* `host-volume-register` is the highest fine-grained privilege because it potentially bypasses quotas.
* `host-volume-create` is implicitly granted by `host-volume-register`
* `host-volume-delete` is implicitly granted only by `host-volume-write`
* `host-volume-read` is implicitly granted by `policy = "read"`,

These are namespaced operations, so the testing here is predominantly around parsing and granting of implicit capabilities rather than the well-tested `AllowNamespaceOperation` method.

This changeset does not include any changes to the `host_volumes` policy which we'll need for claiming volumes on job submit (if any). That'll be covered in a later PR.

### DIFF
--- a/acl/policy.go
+++ b/acl/policy.go
@@ -47,6 +47,11 @@ const (
 	NamespaceCapabilityCSIReadVolume        = "csi-read-volume"
 	NamespaceCapabilityCSIListVolume        = "csi-list-volume"
 	NamespaceCapabilityCSIMountVolume       = "csi-mount-volume"
+	NamespaceCapabilityHostVolumeCreate     = "host-volume-create"
+	NamespaceCapabilityHostVolumeRegister   = "host-volume-register"
+	NamespaceCapabilityHostVolumeRead       = "host-volume-read"
+	NamespaceCapabilityHostVolumeWrite      = "host-volume-write"
+	NamespaceCapabilityHostVolumeDelete     = "host-volume-delete"
 	NamespaceCapabilityListScalingPolicies  = "list-scaling-policies"
 	NamespaceCapabilityReadScalingPolicy    = "read-scaling-policy"
 	NamespaceCapabilityReadJobScaling       = "read-job-scaling"
@@ -207,7 +212,7 @@ func isNamespaceCapabilityValid(cap string) bool {
 		NamespaceCapabilityReadFS, NamespaceCapabilityAllocLifecycle,
 		NamespaceCapabilityAllocExec, NamespaceCapabilityAllocNodeExec,
 		NamespaceCapabilityCSIReadVolume, NamespaceCapabilityCSIWriteVolume, NamespaceCapabilityCSIListVolume, NamespaceCapabilityCSIMountVolume, NamespaceCapabilityCSIRegisterPlugin,
-		NamespaceCapabilityListScalingPolicies, NamespaceCapabilityReadScalingPolicy, NamespaceCapabilityReadJobScaling, NamespaceCapabilityScaleJob:
+		NamespaceCapabilityListScalingPolicies, NamespaceCapabilityReadScalingPolicy, NamespaceCapabilityReadJobScaling, NamespaceCapabilityScaleJob, NamespaceCapabilityHostVolumeCreate, NamespaceCapabilityHostVolumeRegister, NamespaceCapabilityHostVolumeWrite, NamespaceCapabilityHostVolumeRead:
 		return true
 	// Separate the enterprise-only capabilities
 	case NamespaceCapabilitySentinelOverride, NamespaceCapabilitySubmitRecommendation:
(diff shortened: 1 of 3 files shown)


pkazmierczak (approved):
LGTM, but wouldn't we also want an ACL for volume removal?

tgross:
> LGTM, but wouldn't we also want an ACL for volume removal?

That's a good question. Only Variables and Node Pools have an explicit separate of write vs delete today. While I've been working on the RPCs handlers I've been assuming that `host-volume-create` gives you delete as well, but if we were going to do that it should probably be called `host-volume-write`. So maybe it's better just to have a separate delete ACL?

tgross:
I've updated the PR and its description as per the discussion between you, me, and @gulducat.

acl/policy.go
```diff
@@ -241,6 +246,7 @@ func expandNamespacePolicy(policy string) []string {
 		NamespaceCapabilityReadJobScaling,
 		NamespaceCapabilityListScalingPolicies,
 		NamespaceCapabilityReadScalingPolicy,
+		NamespaceCapabilityHostVolumeRead,
```
gulducat, line 249: since there's no addition to `write` below, `policy = "write"` does not imply `host-volume-write` (and its descendants) in contrast to CSI?

acl/policy.go
```diff
@@ -278,6 +284,29 @@ func expandNamespacePolicy(policy string) []string {
...
+		switch cap {
+		case NamespaceCapabilityHostVolumeWrite:
+			extraCaps = append(extraCaps,
+				NamespaceCapabilityHostVolumeRegister, NamespaceCapabilityHostVolumeCreate,
+				NamespaceCapabilityHostVolumeDelete, NamespaceCapabilityHostVolumeRead)
+		case NamespaceCapabilityHostVolumeRegister:
+			extraCaps = append(extraCaps,
+				NamespaceCapabilityHostVolumeCreate, NamespaceCapabilityHostVolumeRead)
+		case NamespaceCapabilityHostVolumeCreate:
+			extraCaps = append(extraCaps, NamespaceCapabilityHostVolumeRead)
```
gulducat, line 301: this is super nitpicky, but I find it easier to read one-per-line:

```suggestion
		case NamespaceCapabilityHostVolumeWrite:
			extraCaps = append(extraCaps,
				NamespaceCapabilityHostVolumeRegister,
				NamespaceCapabilityHostVolumeCreate,
				NamespaceCapabilityHostVolumeDelete,
				NamespaceCapabilityHostVolumeRead)
		case NamespaceCapabilityHostVolumeRegister:
			extraCaps = append(extraCaps,
				NamespaceCapabilityHostVolumeCreate,
				NamespaceCapabilityHostVolumeRead)
		case NamespaceCapabilityHostVolumeCreate:
			extraCaps = append(extraCaps, NamespaceCapabilityHostVolumeRead)
```

acl/policy.go
```diff
@@ -278,6 +284,29 @@ func expandNamespacePolicy(policy string) []string {
 	}
 }
 
+// expandNamespaceCapabilities adds extra capabilities implies by fine-grained
```
gulducat, line 287:
```suggestion
// expandNamespaceCapabilities adds extra capabilities implied by fine-grained
```

gulducat (approved):
lgtm! just one policy question, a style nit, and a typo.

acl/policy.go
tgross: Hm, that's how I wrote it in the RFC, but I'm realizing that I missed that `policy = "write"` gives `csi-write-volume`. Once you're using a coarse-grained policy you can't subtract capabilities from it. So we can do that here, but I think we should probably have it only imply `host-volume-create/read` so that the broad policy doesn't imply `delete` and the quota-breaking `register`

gulducat: is that what's meant by the [ACL policies doc](https://developer.hashicorp.com/nomad/docs/v1.9.x/other-specifications/acl-policy#host-volume-rules) stating

> When both the policy short hand and a capabilities list are provided, the capabilities are merged.

I suppose `deny` would win? but aside from that, a merger would be additive, not subtractive.

aside from losing the simplicity of "write means write", "write means create/read" seems like a reasonable trade-off. :+1:

tgross: Yeah `deny` would win but it's coarse-grained so it flattens everything.

> aside from losing the simplicity of "write means write", "write means create/read" seems like a reasonable trade-off.

Yeah, I think it's an example of how some early choices in Nomad's design (like how we did coarse-grained ACLs) were perfectly good choices at the time but make for awkward decisions in a more mature product years later :grin:

PR #24363, opened 2024-11-05
Allow scaling system jobs to 0

Juanadelacuesta:
This PR introduces the possibility to temporarily stop a system job by scaling it to 0 and then restart it again by scaling it back up to 1, without having to resubmit it.

### DIFF
--- a/e2e/scaling/scaling.go
+++ b/e2e/scaling/scaling.go
@@ -225,9 +225,6 @@ func (tc *ScalingE2ETest) TestScalingBasicWithSystemSchedule(f *framework.F) {
 	stopedAllocs, _, err := jobs.Allocations(jobID, false, nil)
 	f.NoError(err)
 
-	stopedAllocs, _, err = jobs.Allocations(jobID, false, nil)
-	f.NoError(err)
-
 	f.Equal(numberOfNodes, len(filterAllocsByDesiredStatus(structs.AllocDesiredStatusStop, stopedAllocs)))
 	f.Equal(numberOfNodes, len(stopedAllocs))
 
--- a/e2e/scaling/scaling.go
+++ b/e2e/scaling/scaling.go
@@ -175,17 +175,24 @@ func (tc *ScalingE2ETest) TestScalingBasicWithSystemSchedule(f *framework.F) {
 	nomadClient := tc.Nomad()
 
 	// Register a system job with a scaling policy without a group count, it should
-	// default to 1.
+	// default to 1 per node.
 
 	jobID := "test-scaling-" + uuid.Generate()[0:8]
 	e2eutil.RegisterAndWaitForAllocs(t, nomadClient, "scaling/input/namespace_default_system.nomad", jobID, "")
 
 	jobs := nomadClient.Jobs()
-	allocs, _, err := jobs.Allocations(jobID, true, nil)
+	initialAllocs, _, err := jobs.Allocations(jobID, true, nil)
 	require.NoError(t, err)
 
-	require.Equal(t, 1, len(allocs))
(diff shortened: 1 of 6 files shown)


jrasell (commented):
Thanks @Juanadelacuesta, the code here LGTM but I have a couple of comments and questions I believe should be resolved before merging. 

The job endpoint test seems to check that we can call the RPC with the given parameters, but we do not have any additional tests to ensure Nomad runs/stops allocations according to what we expect. Should we add some e2e or additional tests to ensure the correct behaviour?

What is our backwards compatibility stance here? Currently operators can submit system jobs without specifying a count and expect Nomad to default the value to 1. This change would mean these job specifications will no longer result in allocations and means upgraders must modify all job specifications which utilise this behaviour which is a breaking change.

I think it would also be useful to document some nuisances around this feature and how it works and expected behaviour. One example I immediately thought of was how does this interact with Nomad GC in the event I leave a job scaled to zero for an extended period of time?

Juanadelacuesta:
The default to 1 for the system jobs was not done there, it is maintained still, this PR does not change that, so there are no backwards compatibility issues thankfully.  As for the garbage colector, how does it behave with any other type of job?

e2e/scaling/scaling.go
```diff
@@ -165,3 +167,80 @@ func (tc *ScalingE2ETest) TestScalingNamespaces(f *framework.F) {
...
+
+	// Register a system job with a scaling policy without a group count, it should
+	// default to 1.
+
+	jobID := "test-scaling-" + uuid.Generate()[0:8]
+	e2eutil.RegisterAndWaitForAllocs(t, nomadClient, "scaling/input/namespace_default_system.nomad", jobID, "")
+
+	jobs := nomadClient.Jobs()
+	allocs, _, err := jobs.Allocations(jobID, true, nil)
+	require.NoError(t, err)
```
mismithhisler, line 185: Do we use the `must` package in E2E tests as well?

Juanadelacuesta: This package is really old that's why it uses `require`, we always use `must` now, but I didn't want to introduce one test that is different and changing all the tests in the package would make this PR a lot bigger.

nomad/job_endpoint.go
```diff
@@ -1027,8 +1027,9 @@ func (j *Job) Scale(args *structs.JobScaleRequest, reply *structs.JobRegisterRes
 	if job == nil {
 		return structs.NewErrRPCCoded(404, fmt.Sprintf("job %q not found", args.JobID))
 	}
-	if job.Type == structs.JobTypeSystem {
-		return structs.NewErrRPCCoded(http.StatusBadRequest, `cannot scale jobs of type "system"`)
+
+	if job.Type == structs.JobTypeSystem && *args.Count > 1 {
```
mismithhisler, line 1031: I know this is where the current JobTypeSystem validation was located.. But should it really be in the `args.Validate()` method?

Juanadelacuesta: It would make more sense, but the args do not include the job type, so we need to look the job up first before we can validate it. We could add the job specs as an argument to the validation, but we might be making calls for args that are straight up invalid. Thinking more about it, Im not sure how expensive is that call, do you think it is a fair trade off? Also it will mean changing the signature of the validation, Im not sure how big of a change that would be.

mismithhisler: I think I mixed up the fields in the `JobScaleRequest` and thought job was included, but that was wrong.  Looking at it again after reading the above response makes we feel like your current way of doing this might be the better than what I suggested.

mismithhisler (approved):

PR #24368, opened 2024-11-05
docs: ipv6: small fixes

gulducat:
while closing old browser tabs, these little tidbits caught my eye.

* escaping newlines is not allowed in go-sockaddr template
* client{} block in client section
* tiny extra clarification that the NOMAD_ADDR is an example


aimeeu (approved):
Thanks Daniel

PR #24371, opened 2024-11-05
testing: RPC-only test client helper

tgross:
In #10193 we introduced a testing helper that spins up a client RPC server without the rest of the client operations so that we can make server-side client RPC tests lighter. But this wasn't actually ever wired up to the intended target. While working on Dynamic Host Volumes I noticed that this would be useful for RPC tests.

This changeset fixes some bugs in the helper that arose from client code drift, and makes it used by the client RPC tests for CSI. This will also get used for the DHV RPC tests.

Ref: https://github.com/hashicorp/nomad/pull/10193

### DIFF
--- a/client/testing.go
+++ b/client/testing.go
@@ -13,6 +13,7 @@ import (
 	"github.com/hashicorp/nomad/client/fingerprint"
 	"github.com/hashicorp/nomad/client/servers"
 	"github.com/hashicorp/nomad/client/serviceregistration/mock"
+	"github.com/hashicorp/nomad/client/state"
 	agentconsul "github.com/hashicorp/nomad/command/agent/consul"
 	"github.com/hashicorp/nomad/helper/pluginutils/catalog"
 	"github.com/hashicorp/nomad/helper/pluginutils/singleton"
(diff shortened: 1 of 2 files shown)


nomad/client_csi_endpoint_test.go
```diff
@@ -548,17 +512,14 @@ func setupLocal(t *testing.T) (rpc.ClientCodec, func()) {
 	mockCSI.NextDeleteSnapshotError = fmt.Errorf("no plugins registered for type")
 	mockCSI.NextListExternalSnapshotsError = fmt.Errorf("no plugins registered for type")
 
-	c1, cleanupC1 := client.TestClientWithRPCs(t,
-		func(c *config.Config) {
-			c.Servers = []string{s1.config.RPCAddr.String()}
-		},
+	c1, cleanupC1, err := client.TestRPCOnlyClient(t, nil, s1.config.RPCAddr,
```
gulducat, line 515: just flagging that this is the signal in the require->must noise (which is also good!)

client/testing.go
```diff
@@ -91,34 +92,41 @@ func TestClientWithRPCs(t testing.T, cb func(c *config.Config), rpcs map[string]
 // with the server and then returns mock RPC responses for those interfaces
 // passed in the `rpcs` parameter. Useful for testing client RPCs from the
 // server. Returns the Client, a shutdown function, and any error.
-func TestRPCOnlyClient(t testing.T, srvAddr net.Addr, rpcs map[string]interface{}) (*Client, func() error, error) {
+func TestRPCOnlyClient(t testing.T, cb func(c *config.Config), srvAddr net.Addr, rpcs map[string]any) (*Client, func() error, error) {
```
gulducat, line 95: the cleanup cancelFunc being a `func() error` seems a little odd for a test helper, which could instead `t.Error()`. also it looks like all the callers in `client_csi_endpoint_test.go` at least are discarding the error anyway?  maybe it doesn't really matter, but that could possibly leak the cleanup goroutine without us knowing.

gulducat (approved):
lgtm! just a small nit about the cleanup func signature for consideration.

client/testing.go
tgross: Yeah good call -- we can make this return `(*Client, func())` and have it `must.NoError(t, err)` internally for its own errors as well.

gulducat: I tend to do `test.NoError(t, err)` instead of `must` in helper cleanup, because
 * other things after the cleanup may still be attempted (such as other cleanup)
 * [`t.FailNow`](https://pkg.go.dev/testing#T.FailNow) (via `must` -> `t.Fatalf`) can have odd behavior when run in goroutines.

tgross: Normally I'd 100% agree with you, but this is setup code... if it doesn't run we shouldn't continue because we're likely to crash the test process.

PR #24373, opened 2024-11-05
dynamic host volumes: RPC handlers

tgross:
This changeset implements the RPC handlers for Dynamic Host Volumes, including the plumbing needed to forward requests to clients. The client-side implementation is stubbed-out and will be done under a separate PR. Note that although there's no test file for the client RPC that file already has high test coverage from the server RPC tests.


nomad/host_volume_endpoint.go
```diff
@@ -0,0 +1,527 @@
...
+				},
+			)
+
+			filters := []paginator.Filter{
+				paginator.GenericFilter{
+					Allow: func(raw any) (bool, error) {
+						vol := raw.(*structs.HostVolume)
+						if !strings.HasPrefix(vol.Name, args.Prefix) &&
+							!strings.HasPrefix(vol.ID, args.Prefix) {
+							return false, nil
```
gulducat, line 142: just noting that this implicitly checks `if args.Prefix != ""`, i.e. an empty `Prefix` does not filter things out.

nomad/host_volume_endpoint.go
```diff
@@ -0,0 +1,527 @@
...
+			// stubs slice.
+			paginatorImpl, err := paginator.NewPaginator(iter, tokenizer, filters, args.QueryOptions,
+				func(raw any) error {
+					vol := raw.(*structs.HostVolume)
+					vols = append(vols, vol.Stub())
+					return nil
+				})
+			if err != nil {
+				return structs.NewErrRPCCodedf(
+					http.StatusBadRequest, "failed to create result paginator: %v", err)
```
gulducat, line 176: I don't think an HTTP API caller can cause this to happen?  so feels more like a server error than a bad request.

```suggestion
			if err != nil {
				return structs.NewErrRPCCodedf(
					http.StatusInternalServerError, "failed to create result paginator: %v", err)
```

nomad/host_volume_endpoint.go
```diff
@@ -0,0 +1,527 @@
...
+			}
+
+			// Calling page populates our output variable stub array as well as
+			// returns the next token.
+			nextToken, err := paginatorImpl.Page()
+			if err != nil {
+				return structs.NewErrRPCCodedf(
+					http.StatusBadRequest, "failed to read result page: %v", err)
```
gulducat, line 184: ditto here, I think

```suggestion
			nextToken, err := paginatorImpl.Page()
			if err != nil {
				return structs.NewErrRPCCodedf(
					http.StatusInternalServerError, "failed to read result page: %v", err)
```

nomad/host_volume_endpoint.go
```diff
@@ -0,0 +1,527 @@
...
+
+	// ensure we only try to create valid volumes or make valid updates to
+	// volumes
+	validVols, err := v.validateVolumeUpdates(args.Volumes)
+	if err != nil {
+		mErr = multierror.Append(mErr, err)
+	}
+	if len(validVols) == 0 {
+		return helper.FlattenMultierror(mErr)
+	}
```
gulducat, line 244: I would think _any_ invalid volumes would be a bad request, and return (multi)error.  why partially execute their request?

nomad/host_volume_endpoint.go
```diff
@@ -0,0 +1,527 @@
...
+
+	// Attempt to create all the validated volumes and write only successfully
+	// created volumes to raft. And we'll report errors for any failed volumes
+	//
+	// NOTE: creating the volume in the external storage provider can't be
```
gulducat, line 249: I suspect copypasta from CSI -- to what does "external storage provider" refer?

gulducat (commented):
lookin good!  only partway through my review, here's what I have so far.

client/structs/host_volumes.go
```diff
@@ -0,0 +1,61 @@
...
+	// client RPC request so that the server can route the request to the
+	// correct node.
+	NodeID string
+
+	// Path is the host path where the volume's mount point was created. We send
+	// this from the server to allow verification by plugins
+	HostPath string
+
+	// Parameters are an opaque map of parameters for the host volume plugin.
+	Parameters map[string]string
```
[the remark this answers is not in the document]
gulducat: in retrospect I suppose it could be all kinds of things, including secrets to reach out to \<some external system> or who knows.

nomad/host_volume_endpoint.go
tgross: That would certainly simplify things. I think we still need to handle multierrors for client RPCs (because those won't be atomic and we'll want to record any volumes that we've successfully created/updated on the client), but we can certainly exit early here.

nomad/host_volume_endpoint.go
tgross: The client-side plugin, but you're right it's copypasta from CSI. Will fix.

nomad/host_volume_endpoint.go
tgross: You'd be very reasonable to think that! But `NewPaginator` only returns an error if the go-bexpr expression is invalid, which is entirely user-controlled. It might be nice to have a preflight check in the HTTP API for this, but that'd be something we'd want to do across the whole HTTP API.

nomad/fsm.go
```diff
@@ -59,6 +59,7 @@ const (
 	NodePoolSnapshot                     SnapshotType = 28
 	JobSubmissionSnapshot                SnapshotType = 29
 	RootKeySnapshot                      SnapshotType = 30
+	HostVolumeSnapshot                   SnapshotType = 31
```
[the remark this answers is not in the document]
tgross: That'd be a `VolumeSnapshotSnapshot` :grin:

client/structs/host_volumes.go
```diff
@@ -0,0 +1,61 @@
...
+	// NodeID is the node where the volume is placed. It's included in the
+	// client RPC request so that the server can route the request to the
+	// correct node.
+	NodeID string
+
+	// Because storage may allow only specific intervals of size, we accept a
+	// min and max and return the actual capacity when the volume is created or
+	// updated on the client
+	RequestedCapacityMin int64 // bytes
+	RequestedCapacityMax int64 // bytes
```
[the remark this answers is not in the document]
tgross: Sure, we can do that!

nomad/host_volume_endpoint.go
tgross: Right. Maybe that's surprising though and deserves a comment

nomad/host_volume_endpoint.go
tgross: Relevant docstring on that method, for context: 

```go
// NewPaginator returns a new Paginator. Any error creating the paginator is
// due to bad user filter input, RPC functions should therefore return a 400
// error code along with an appropriate message.
```

client/structs/host_volumes.go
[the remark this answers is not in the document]
tgross: Yeah I think as you start digging into the client-side work we'll have a better idea of what this looks like. CSI has like 4 different maps for getting different sets of parameters to the plugins. We can consider this a placeholder until you're further into implementation.

nomad/host_volume_endpoint_test.go
```diff
@@ -0,0 +1,571 @@
...
+		var getResp structs.HostVolumeGetResponse
+		err = msgpackrpc.CallWithCodec(codec, "HostVolume.Get", getReq, &getResp)
+		must.NoError(t, err)
+		must.Nil(t, getResp.Volume)
+
+		// update the allocations terminal so the delete works
+		alloc = alloc.Copy()
+		alloc.ClientStatus = structs.AllocClientStatusFailed
+		nArgs := &structs.AllocUpdateRequest{
+			Alloc: []*structs.Allocation{alloc},
```
gulducat, line 245: `Alloc` field is marked as deprecated :shrug:

nomad/host_volume_endpoint_test.go
```diff
@@ -0,0 +1,571 @@
...
+	vol3.NodeID = nodes[2].ID
+	vol3.NodePool = "prod"
+	vol3.Namespace = ns2
+	vol3.Name = "foobar-example"
+	vol4.Namespace = ns2
+	vol4.NodeID = nodes[1].ID
+	vol4.Name = "foobaz-example"
+
+	registerReq.Volumes = []*structs.HostVolume{vol3, vol4}
+	registerReq.Namespace = ns2
```
gulducat, line 338: I was wondering why these aren't registered with vol1&2, and I suppose it's because different `WriteRequest.Namespace`? I thought vols in separate NS could be registered at the same time... or have I got myself confused?

nomad/host_volume_endpoint_test.go
```diff
@@ -0,0 +1,571 @@
...
+			}
+			must.SliceContainsAll(t, tc.expectVolIDs, gotIDs,
+				must.Sprintf("got: %v", gotIDs))
+		})
+	}
+
+	t.Run("blocking query unblocks", func(t *testing.T) {
+
+		getReq := &structs.HostVolumeGetRequest{
+			ID: vol1.ID,
```
gulducat, line 448: I wasn't sure why this extra Get, but a little ways down I see it's to get the `Index` eh? could save a bit of cognitive load here

```suggestion
		// Get response will include the volume's Index to block on
		getReq := &structs.HostVolumeGetRequest{
			ID: vol1.ID,
```

nomad/host_volume_endpoint_test.go
```diff
@@ -0,0 +1,571 @@
...
+				Namespace:     ns1,
+				AuthToken:     token,
+				MinQueryIndex: getResp.Index,
+			},
+		}
+
+		go func() {
+			codec := rpcClient(t, srv)
+			var getResp structs.HostVolumeGetResponse
+			err := msgpackrpc.CallWithCodec(codec, "HostVolume.Get", getReq, &getResp)
```
gulducat, line 482: this should be `List` right? not `Get`

nomad/structs/host_volumes.go
```diff
@@ -180,18 +180,68 @@ const (
...
+	WriteMeta
+}
+
+type HostVolumeDeleteRequest struct {
+	VolumeIDs []string
+	WriteRequest
+}
+
+type HostVolumeDeleteResponse struct {
+	WriteMeta
```
gulducat, line 225: partial deletes are possible, right? so maybe returning deleted volume IDs would be useful

```suggestion
type HostVolumeDeleteResponse struct {
	VolumeIDs []string
	WriteMeta
```

I suspect this will have implications where this struct is used

nomad/host_volume_endpoint_test.go
```diff
@@ -0,0 +1,571 @@
...
+			expectVolIDs: []string{},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			var resp structs.HostVolumeListResponse
+			err := msgpackrpc.CallWithCodec(codec, "HostVolume.List", tc.req, &resp)
+			must.NoError(t, err)
+			gotIDs := []string{}
```
gulducat, line 436: my IDE makes orange squiggle

```suggestion
			var gotIDs []string
```

nomad/host_volume_endpoint.go
```diff
@@ -0,0 +1,515 @@
...
+	reply.Volumes = raftArgs.Volumes
+	reply.Index = index
+	return helper.FlattenMultierror(mErr)
+}
+
+func (v *HostVolume) validateVolumeUpdates(requested []*structs.HostVolume) ([]*structs.HostVolume, error) {
+
+	now := time.Now().UnixNano()
+
+	vols := []*structs.HostVolume{}
```
gulducat, line 340: more IDE complaining

```suggestion
	var vols []*structs.HostVolume
```

nomad/host_volume_endpoint.go
```diff
@@ -0,0 +1,515 @@
...
+	}
+	if !allowVolume(aclObj, args.RequestNamespace()) {
+		return structs.ErrPermissionDenied
+	}
+
+	if len(args.VolumeIDs) == 0 {
+		return fmt.Errorf("missing volumes to delete")
+	}
+
+	volsToDelete := []string{}
```
gulducat, line 451:
```suggestion
	var volsToDelete []string
```

nomad/host_volume_endpoint.go
```diff
@@ -0,0 +1,515 @@
...
+		}
+
+		// if the volume already exists, we'll ensure we're validating the
+		// update
+		current, err := snap.HostVolumeByID(nil, vol.Namespace, vol.ID, false)
+		if err != nil {
+			mErr = multierror.Append(mErr, err)
+			continue
+		}
+		if err = vol.Validate(current); err != nil {
```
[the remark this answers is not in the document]
tgross: Yeah, because we'll need to validate the volume even if it's a new one, so having validation for both new and updated volumes together seems nicer. Keeping in mind that `HostVolume.Validate` is no-op for the moment until we get a little further along, of course. :grin:

nomad/structs/host_volumes.go
tgross: Good call. In practice I wasn't able to make the HTTP API nice to use with multiple volumes, because of the existing URL routes for CSI which we're sharing. So being able to delete multiple volumes is going to be primarily only here to support soft deletes and periodic GC, which will be in later PRs.

nomad/host_volume_endpoint.go
```diff
@@ -0,0 +1,515 @@
...
+			mErr = multierror.Append(mErr,
+				fmt.Errorf("volume %s in use by allocations: %v", id, allocIDs))
+			continue
+		}
+
+		err = v.deleteVolume(vol)
+		if err != nil {
+			mErr = multierror.Append(mErr, err)
+		} else {
+			volsToDelete = append(volsToDelete, id)
```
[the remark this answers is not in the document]
tgross: Would it be better if it were `deletedVols`?

nomad/host_volume_endpoint_test.go
```diff
@@ -0,0 +1,571 @@
...
+			Namespace: "invalid",
+			AuthToken: token},
+	}
+
+	t.Run("invalid create", func(t *testing.T) {
+
+		// TODO(1.10.0): once validation logic for updating an existing volume is in
+		// place, fully test it here
+
+		req.Namespace = ns
```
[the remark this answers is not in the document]
tgross: Yeah, better to leave it empty and then set it here.

nomad/host_volume_endpoint_test.go
```diff
@@ -0,0 +1,571 @@
...
+		must.NoError(t, err)
+
+		getReq.ID = vol2ID
+		err = msgpackrpc.CallWithCodec(codec, "HostVolume.Get", getReq, &getResp)
+		must.NoError(t, err)
+		must.Nil(t, getResp.Volume)
+	})
+}
+
+func TestHostVolumeEndpoint_List(t *testing.T) {
```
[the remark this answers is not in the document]
tgross: Yeah `List` has a lot more dimensions and wants a one-and-done setup with a bunch of volumes to exercise those, so it wants to be its own test. And `TestHostVolumeEndpoint_CreateRegisterGetDelete` seemed verbose.

nomad/host_volume_endpoint_test.go
tgross: > I thought vols in separate NS could be registered at the same time... or have I got myself confused?

No, you're right! That changed from when I wrote this test a few days ago, because of the bug we discovered in CSI. I can clean that up now.

nomad/host_volume_endpoint.go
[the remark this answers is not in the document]
gulducat: personally yeah I think so

nomad/host_volume_endpoint_test.go
tgross: What's the reasoning behind that? We don't lint for it.

The [short declaration](https://go.dev/ref/spec#Short_variable_declarations) is an alias functionally identical _for slices_ (or value types) to a [variable declaration](https://go.dev/ref/spec#Variable_declarations) because the result is `nil` either way and a nil slices doesn't blow up when you run built-in functions on it. But for maps and pointers those nils cause panics when you get it wrong, and being consistent between them avoids mishaps.

But it's also the wrong tool for the job here, because should be using [helper.ConvertSlice](https://github.com/hashicorp/nomad/blob/main/helper/funcs.go#L401) anyways.

nomad/host_volume_endpoint_test.go
tgross: :facepalm: Sure should be.

nomad/host_volume_endpoint_test.go
```diff
@@ -0,0 +1,571 @@
...
+			return false
+		}
+		if node != nil && node.Status == structs.NodeStatusReady {
+			return true
+		}
+		return false
+	}),
+		wait.Timeout(time.Second*5),
+		wait.Gap(time.Millisecond),
+	))
```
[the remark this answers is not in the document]
tgross: It's waiting for a fingerprint, but yeah I'll add that.

nomad/host_volume_endpoint_test.go
```diff
@@ -0,0 +1,571 @@
...
+
+	return mockClientEndpoint, c1.Node()
+}
+
+func (v *mockHostVolumeClient) setCreate(
+	resp *cstructs.ClientHostVolumeCreateResponse, errMsg string) {
+	v.lock.Lock()
+	defer v.lock.Unlock()
+	v.nextCreateResponse = resp
+	v.nextCreateErr = errors.New(errMsg)
```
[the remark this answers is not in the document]
tgross: Bah, good catch.

nomad/host_volume_endpoint.go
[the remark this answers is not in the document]
tgross: Adding a TODO comment in the `Validate` method for this just to make sure it doesn't fall thru the cracks.

nomad/host_volume_endpoint_test.go
tgross: The client does it the deprecated way still, with a compatibility note on the struct for Nomad 0.11 :laughing: So I'm going to leave this as-is so we're testing the behavior that actually exists.

nomad/host_volume_endpoint.go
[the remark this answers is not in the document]
tgross: Done!

nomad/host_volume_endpoint_test.go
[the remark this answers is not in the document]
tgross: Changed the CRUD test name.

gulducat (approved):
huzzah! :rocket:

PR #24374, opened 2024-11-05
[ui] When your token expires and you sign in again, redirect to your original route.

### LINKED ISSUE #23793: [ui] When a token is expired, upon re-authentication, take me back to the route I came from
Jira: https://hashicorp.atlassian.net/browse/NET-10695

philrenaud:
When your token 403s or 500s (when it's deleted behind the scenes, or expires either while you're using the app, or while it's in your localStorage, say, overnight), this PR adds functionality to remember the page (and queryParams) you were on when it happened.

Upon signing back in (with a renewed token, say), you'll be automatically rerouted back to that page, with a little notification saying so.

Resolves #23793

### DIFF
--- a/.changelog/24374.txt
+++ b/.changelog/24374.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: When your token expires, upon signing back in, redirect to your original route
+```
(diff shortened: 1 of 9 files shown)


schmichael (approved):
:tada:

PR #24376, opened 2024-11-06
test: ensure RPC only test client sets enterprise specific config.

jrasell:
Closes: https://github.com/hashicorp/nomad-enterprise/issues/1985
Additional details can be seen in the linked issue and evidence of the tested fix working.


tgross (approved):
Oof, didn't even think to check that. Well that's why we run CI on ENT!

Thanks for picking this up @jrasell!

PR #24381, opened 2024-11-06
api: new dispatch endpoint sends body as Payload

### LINKED ISSUE #24312: Nomad Job Dispatch - Payload via HTTP request body
### Proposal
The ability to send the job payload *as* the http request body, instead of having to base64-encode it. 

That would allow support for something like this:

```
POST /v1/job/my-parameterized-job/dispatch?BodyAsPayload=true HTTP/1.1
Content-Type: text/plain

This is the amazing payload that will be sent to the parameterized job.
```

### Use-cases
Some external systems support 'webhooks', but don't allow for much control over the contents of the request body. These bodies usually contain information on the event or the changed entity, but do not allow us to specify "put that information in a `Payload` object and base64-encode everything". This makes it difficult to use Nomad's dispatch-job as a webhook handler.

It would be nice to be able to tell Nomad (via a query parameter or a HTTP header) to interpret the entire HTTP request body as the raw payload. 

### Attempted Solutions
- Replace the HTTP request body in the load balancer, before sending it off to Nomad. Even though load balancers (like haproxy, traefik) have good support for adding/replacing HTTP headers, they generally do not like modifying HTTP request bodies.

gulducat:
This opens up dispatching parameterized jobs by systems that do not allow modifying what http request body they send (like webhooks).

e.g. these two requests are equal:

```shell
# existing endpoint
POST '{"Payload": "'"$(base64 <<< "hello")"'"}' /v1/job/my-job/dispatch
# new in this PR
POST 'hello' /v1/job/my-job/dispatch/payload
```

Closes #24312

### DIFF
--- a/.changelog/24312.txt
+++ b/.changelog/24312.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+api: new parameterized dispatch endpoint sends raw HTTP request body as Payload
+```
--- a/command/agent/job_endpoint.go
+++ b/command/agent/job_endpoint.go
@@ -5,6 +5,7 @@ package agent
 
 import (
 	"fmt"
+	"io"
 	"maps"
 	"net/http"
 	"slices"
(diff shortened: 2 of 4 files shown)


EtienneBruines (approved):

Juanadelacuesta (approved):
It looks pretty good

Issue #24379, opened 2024-11-06
Nomad 1.9.1: panic: runtime error: slice bounds out of range [:12] with capacity 0


HINT-SJ:
### Nomad version
```
Nomad v1.9.1
BuildDate 2024-10-21T09:00:50Z
Revision d9ec23f0c1035401e9df6c64d6ffb8bffc555a5e
```

### Operating system and Environment details
Amazon Linux 2023 (minimal)
AWS EC2 Gravitron (t4g.small)
> linux ip-#####.eu-central-1.compute.internal 6.1.112-122.189.amzn2023.aarch64 #1 SMP Tue Oct  8 17:01:34 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux

### Issue

We observed two cases of Nomad clusters failing to start after a scheduled system patch window.
After installing system patches (through AWS SSM) and a followed system reboot (1 instance at a time, 5 instances acting as nomad server instances) we experienced the following panic upon service start on all 5 server instances:

```
nomad.service: Deactivated successfully.
Stopped nomad.service - Nomad Agent.
nomad.service: Consumed 13min 5.870s CPU time.
```

-- Instance Restart --

```
Started nomad.service - Nomad Agent.

==> WARNING: mTLS is not configured - Nomad is not secure without mTLS!
WARNING: keyring exists but -encrypt given, using keyring
==> Loaded configuration from /etc/nomad.d/nomad.hcl
==> Starting Nomad agent...
panic: runtime error: slice bounds out of range [:12] with capacity 0
goroutine 1 [running]:
github.com/hashicorp/go-kms-wrapping/v2/aead.(*Wrapper).Decrypt(0x40006cfa40, {0x40006cdc70?, 0x400091bb00?}, 0x400049e160, {0x0?, 0x20?, 0x21?})
        github.com/hashicorp/go-kms-wrapping/v2@v2.0.16/aead/aead.go:272 +0x1a0
github.com/hashicorp/nomad/nomad.(*Encrypter).loadKeyFromStore(0x4000685490, {0x400079c9c0?, 0x400079c9e0?})
        github.com/hashicorp/nomad/nomad/encrypter.go:809 +0x3a0
github.com/hashicorp/nomad/nomad.(*Encrypter).loadKeystore.func1({0x400079c9c0, 0x52}, {0x368c318?, 0x40009580c0?}, {0x0?, 0x0?})
        github.com/hashicorp/nomad/nomad/encrypter.go:172 +0x264
[...]

PR #24383, opened 2024-11-07
keyring: Fix a panic when decrypting aead with empty RSA block.

jrasell:
Closes: https://github.com/hashicorp/nomad/issues/24379

The code path exists in 1.9 and 1.8, so backporting to both even though reports suggest this is only occurring on 1.9.0+.

Clusters that have gone through several upgrades have be found to
include keyring material which has an empty RSA block.

In more recent versions of Nomad, an empty RSA block is omitted
from being written to disk. This results in the panic not being
present. Older versions, however, did not have this struct tag
meaning we wrote an empty JSON block which is not accounted for
in the current version.

### DIFF
--- a/.changelog/24383.txt
+++ b/.changelog/24383.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+keyring: Fixed a panic on server startup when decrypting AEAD key data with empty RSA block
+```


tgross (approved):
Assuming the file itself hasn't been corrupted on-disk, this case is supposed to be protected against by the block above where you've made a change. Older KEK wrapper versions with AEAD-only have the DEK ciphertext in a different field, and we're supposed to be restoring them here.

My hunch would be that there's an issue with those older keys. But unfortunately I tested this out with Nomad 1.4 and did a single-node upgrade to tip and didn't run into any problems.

For the time being, let's get this merged in because it'll prevent a panic.

tgross (approved):
LGTM!

> Older versions, however, did not have this struct tag
:facepalm:

PR #24389, opened 2024-11-07
gc: be consistent with setting create/modify timestamp tz

pkazmierczak:
Whenever setting objects creation/modify time, we should always use UTC. https://github.com/hashicorp/nomad/pull/24112 introduced some inconsistencies in this area, and this PR fixes it.

### DIFF
--- a/nomad/structs/csi.go
+++ b/nomad/structs/csi.go
@@ -376,7 +376,7 @@ type CSIVolListStub struct {
 
 // NewCSIVolume creates the volume struct. No side-effects
 func NewCSIVolume(volumeID string, index uint64) *CSIVolume {
-	now := time.Now().UnixNano()
+	now := time.Now().UTC().UnixNano()
 	out := &CSIVolume{
 		ID:          volumeID,
 		CreateIndex: index,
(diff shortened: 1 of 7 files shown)


nomad/structs/csi.go
```diff
@@ -376,7 +376,7 @@ type CSIVolListStub struct {
 
 // NewCSIVolume creates the volume struct. No side-effects
 func NewCSIVolume(volumeID string, index uint64) *CSIVolume {
-	now := time.Now().UnixNano()
+	now := time.Now().UTC().UnixNano()
```
tgross, line 379: [`UnixNano`](https://pkg.go.dev/time#Time.UnixNano) is already UTC, so we don't need to call `UTC()` in cases where we're immediately converting `time.Now()` to `int64` via `UnixNano()`. (Here and in `NewCSIPlugin` below.)

```go
func TestTimeUnixNano(t *testing.T) {
	now := time.Now()
	utcnow := now.UTC()
	fmt.Printf("%d\n%d\n", now.UnixNano(), utcnow.UnixNano())
}
```

Run on EDT:

```
$ go test -v -count=1 ./nomad -run TestTimeUnixNano
=== RUN   TestTimeUnixNano
1730995200287246224
1730995200287246224
--- PASS: TestTimeUnixNano (0.00s)
PASS
ok      github.com/hashicorp/nomad/nomad        0.027s
```

tgross (approved):
LGTM!

PR #24392, opened 2024-11-07
drivers: provide empty implementations of cgroup helpers for non-root nomad

shoenig:
This PR fixes a regression in #24340 where non-root Nomad agents would panic when using the `raw_exec` driver. No CL because that change hasn't been released yet.


tgross (approved):
LGTM!

PR #24396, opened 2024-11-07
CSI: fix namespace ACL bypass on create/register APIs

tgross:
When creating or registering a CSI volume, the RPC handler uses the volume specification's namespace instead of the request namespace. This works as intended, but the ACL check is only on the request namespace.

This allows a cross-namespace ACL bypass for authenticated users who have `csi-write-volume` capabilities in one namespace but not another namespace. Such a user can set the volume specification to a forbidden namespace while setting the `-namespace` flag in the CLI or API. The ACL check happens against the namespace they do have permission to, but the volume is created in the forbidden namespace.

This changeset fixes the bug by moving the namespace check into the loop over the volumes being written by the RPCs. It also updates the tests to better cover ACL checking in these two RPCs.

Ref: CVE-2024-10975
Ref: https://hashicorp.atlassian.net/browse/SECVULN-15463
Fixes: https://github.com/hashicorp/nomad/issues/24397

### DIFF
--- a/nomad/csi_endpoint.go
+++ b/nomad/csi_endpoint.go
@@ -304,7 +304,8 @@ func (v *CSIVolume) Register(args *structs.CSIVolumeRegisterRequest, reply *stru
 
 	defer metrics.MeasureSince([]string{"nomad", "volume", "register"}, time.Now())
 
-	if !allowVolume(aclObj, args.RequestNamespace()) || !aclObj.AllowPluginRead() {
+	// permission for the volume namespaces will be checked below
+	if !aclObj.AllowPluginRead() {
 		return structs.ErrPermissionDenied
 	}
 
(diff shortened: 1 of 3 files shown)


nomad/csi_endpoint.go
```diff
@@ -304,32 +304,34 @@ func (v *CSIVolume) Register(args *structs.CSIVolumeRegisterRequest, reply *stru
...
 	}
 
-	// This is the only namespace we ACL checked, force all the volumes to use it.
-	// We also validate that the plugin exists for each plugin, and validate the
+	snap, err := v.srv.State().Snapshot()
+	if err != nil {
+		return err
+	}
+
+	// Validate ACLs, that the plugin exists for each plugin, and validate the
```
gulducat, line 321:
```suggestion
	// Validate ACLs, that the plugin exists for each volume, and validate the
```

nomad/csi_endpoint.go
```diff
@@ -1062,13 +1065,20 @@ func (v *CSIVolume) Create(args *structs.CSIVolumeCreateRequest, reply *structs.
...
 	validatedVols := []validated{}
 
-	// This is the only namespace we ACL checked, force all the volumes to use it.
-	// We also validate that the plugin exists for each plugin, and validate the
+	snap, err := v.srv.State().Snapshot()
+	if err != nil {
+		return err
+	}
+
+	// Validate ACLs, that the plugin exists for each plugin, and validate the
```
gulducat, line 1073:
```suggestion
	// Validate ACLs, that the plugin exists for each volume, and validate the
```

gulducat (approved):
assuming tests pass, lgtm!  only a couple code comment typos

dduzgun-security (approved):
LGTM! Would be good to merge when tests are 🟢, thanks for the awesome quick fix

Issue #24140, opened 2024-10-07
Template change_mode intermittently ignored after client restart


ygersie:
### Nomad version
1.8.2+ent

### Issue
Every time we restart a large fleet of clients some allocations end up with rendered dynamic certificates but they haven't been signaled/restarted despite having a proper `change_mode` set in the jobspec. This leads to production incidents where secrets have expired from a workload perspective (didn't get a signal or restart) even though the certificates on the filesystem have been updated.

### Reproduction steps
I haven't been able to reproduce this in an isolated manner. It does happen every time on just a few allocations across our entire cluster after client restarts. On the same node there are many other allocations that get signaled/restarted just fine.

### Logs
When issuing dynamic certificates from Vault through the template stanza a new certificate is created on each client restart. This should always trigger a `change_mode` event but occasionally that does not happen for some allocations which leads to outages.
Example here from a client log snippet:
```
mynodename:~# zgrep '1f514495-e45b-1428-2ab6-b4f2c67e9a0e' /var/log/nomad.log-*.gz
/var/log/nomad.log-20241001.gz:    2024-09-30T10:51:24.428Z [INFO]  client.alloc_runner.task_runner: Task event: alloc_id=1f514495-e45b-1428-2ab6-b4f2c67e9a0e task=mytask type=Received msg="Task received by client" failed=false
/var/log/nomad.log-20241001.gz:    2024-09-30T10:51:24.934Z [WARN]  client.alloc_runner.task_runner.task_hook.api: error creating task api socket: alloc_id=1f514495-e45b-1428-2ab6-b4f2c67e9a0e task=mytask path=/data/podman/root/nomad/data/alloc/1f514495-e45b-1428-2ab6-b4f2c67e9a0e/mytask/secrets/api.sock error="listen unix /data/podman/root/nomad/data/alloc/1f514495-e45b-1428-2ab6-b4f2c67e9a0e/mytask/secrets/api.sock: bind: invalid argument"
/var/log/nomad.log-20241001.gz:    2024-09-30T10:51:25.119Z [INFO]  agent: (runner) rendered "(dynamic)" => "/data/podman/root/nomad/data/alloc/1f514495-e45b-1428-2ab6-b4f2c67e9a0e/mytask/secrets/client.pem"
[...]

PR #24399, opened 2024-11-07
fix: handle template re-renders on client restart

mismithhisler:
When multiple templates with api functions are included in a task, it's possible for consul-template to re-render templates as it creates watchers, overwriting render event data. This change uses event fields that do not get overwritten, and only executes the change mode for templates that were actually written to disk.

Fixes: [#24140](https://github.com/hashicorp/nomad/issues/24140)

### DIFF
--- a/client/allocrunner/taskrunner/template/template.go
+++ b/client/allocrunner/taskrunner/template/template.go
@@ -317,7 +317,10 @@ WAIT:
 				if event.LastWouldRender.IsZero() {
 					continue WAIT
 				}
-				// If the template _actually_ rendered to disk, mark it dirty
+				// If the template _actually_ rendered to disk, mark it
+				// dirty. We track events here so that onTemplateRendered
+				// doesn't go back to the runner's RenderedEvents and process
+				// new events that don't make us dirty.				
 				if !event.LastDidRender.IsZero() {
 					dirtyEvents[event.Template.ID()] = event
 				}
--- a/client/allocrunner/taskrunner/template/template.go
+++ b/client/allocrunner/taskrunner/template/template.go
@@ -277,6 +277,10 @@ func (tm *TaskTemplateManager) handleFirstRender() {
 		<-eventTimer.C
 	}
 
+	// dirtyEvents are events that actually rendered to disk and need to trigger
+	// their respective change_mode operation
+	dirtyEvents := map[string]*manager.RenderEvent{}
+
 	// outstandingEvent tracks whether there is an outstanding event that should
 	// be fired.
 	outstandingEvent := false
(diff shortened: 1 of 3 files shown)


client/allocrunner/taskrunner/template/template.go
```diff
@@ -308,22 +312,22 @@ WAIT:
...
 				// This template hasn't been rendered
 				if event.LastWouldRender.IsZero() {
 					continue WAIT
 				}
-				if event.WouldRender && event.DidRender {
-					dirty = true
+				// If the template _actually_ rendered to disk, mark it dirty
+				if !event.LastDidRender.IsZero() {
+					dirtyEvents[event.Template.ID()] = event
 				}
```
tgross, line 323: If I'm following this logic correctly, it looks like we can expect multiple events on after we get woken up by `TemplateRenderedCh` but we only want to "keep" any events that are actually rendered-to-disk. That way we're not going back to the `RenderedEvents` and in the `onTemplateRendered` handler and getting new events we haven't previously seen. 

If so, that makes sense to me but we could probably use a comment to that effect here as this area of code has clearly been a bug farm historically. Maybe something like:

```suggestion
				// If the template _actually_ rendered to disk, mark it
				// dirty. We track events here so that onTemplateRendered
				// doesn't go back to the runner's RenderedEvents and process
				// new events that don't make us dirty.				
				if !event.LastDidRender.IsZero() {
					dirtyEvents[event.Template.ID()] = event
				}
```

client/allocrunner/taskrunner/template/template_test.go
```diff
@@ -806,6 +806,69 @@ OUTER:
 	}
 }
 
+// Tests an edge case where a task has multiple templates and the client is restarted.
+// In this case, vault may re-render and overwrite some fields in the first render event
```
tgross, line 810: Vault? I think we're saying that the upstream service (for example Vault) might change and force a render, but it reads a little weird.

client/allocrunner/taskrunner/template/template_test.go
```diff
@@ -806,6 +806,69 @@ OUTER:
...
+// but we still want to make sure it causes a restart.
+// We cannot control the order in which these templates are rendered, so this test will
+// exhibit flakiness if this edge case is not properly handled.
```
tgross, line 813: This makes me want a test rig where we can mock CT and step thru its operations deterministically... but that's a ton of work, probably better fixed by pulling the runner in-tree, and _way_ out of scope for this PR. :grin:

tgross (approved):
LGTM! Nice work hunting this down!

client/allocrunner/taskrunner/template/template_test.go
mismithhisler: Whoops I mean't consule-template.  Good catch!

client/allocrunner/taskrunner/template/template.go
mismithhisler: While attempting to replicate this bug I able to "mostly" replicate the CT behavior from the GH issue fairly consistently (the _actual_ replication was a lot less consistent).  By "mostly" I mean on client restart, the task shouldn't have been restarting, because the event was being overwritten.. But it was restarting!  Turns out that the other template's event was triggering the dirty flag (it had `change_mode = "noop"`), and then in `onTemplateRendered` we were executing the `change_mode` of a template that was not actually dirty, and restarting.

mismithhisler: So technically `onTemplateRendered` was not getting _new_ events it hadn't seen, it was performing _different_ logic to determine if an event should trigger a change_mode than `handleFirstRender` was.  Probably the best fix here is to move all this logic into one location, but that's slightly bigger change.

PR #24403, opened 2024-11-08
release: fix typo in the LAST_RELEASE sed regex

pkazmierczak:
In the past the `make changelog` command expected `v`-prefixed `LAST_VERSION`, but now it doesn't. The regular expression we had in the release workflow did not account for this and thus the changelog update didn't work correctly.


tgross (approved):
LGTM

Issue #23859, opened 2024-08-23
[WI] Vault renew self fails


ygersie:
With workload identity I found myself in a scenario where Vault token renewal fails indefinitely. It looks like determining if this is a fatal error doesn't work correctly as the error is different than what is in this list: https://github.com/hashicorp/nomad/blob/36522ec6320b9663eca967ba1d6ebe7dfa856327/client/vaultclient/vaultclient.go#L439

```
    2024-08-22T18:45:35.294+0200 [ERROR] client.vault: error during renewal of lease or token failed due to a non-fatal error; retrying: name=default
  error=
  | failed to renew the vault token: Error making API request.
  |
  | URL: PUT http://localhost:8200/v1/auth/token/renew-self
  | Code: 400. Errors:
  |
  | * lease expired
```

In my setup this will never resolve I assume because of:
```
  template {
    vault_retry {
      attempts    = 0
      backoff     = "250ms"
      max_backoff = "5m"
    }
  }
```

Juanadelacuesta:
Hi @ygersie, can you please give us a little more context on what your problem is? When the vault client runs into a fatal error, something that will keep on returning an error in time, like a lease expiration, it will stop renewing it. If you provide us with more information, we might be able to have a better insight into your particular problem, and find a better way to help.

ygersie:
Hey @Juanadelacuesta, thanks for your response.

> When the vault client runs into a fatal error, something that will keep on returning an error in time, like a lease expiration, it will stop renewing it.

Exactly, in this case Nomad incorrectly determines that this is a `non-fatal` error and keeps retrying. There may be other errors that are boiling up now with the switch to WI that aren't caught in the error checking right now. I wonder if it wouldn't be better to also check the Vault response status code, if it's in the 4XX range it by definition should be fatal, right?

I've also seen something else that could be problematic. I'm running an extensive local lab setup of hashistack that includes multiple regions, mTLS and Vault + Consul integration using Workload Identity. If I leave my laptop to rest overnight in the morning a bunch of my jobs are dead and won't start anymore. Here's an example of such an alloc:
```
ID                     = 97415db2-cbb0-fb20-b907-54bb412376e3
Eval ID                = e9bf10f1
Name                   = teleport-proxy.teleport-proxy[0]
Node ID                = e7cb90f9
Node Name              = agent-1
Job ID                 = teleport-proxy
Job Version            = 0
Client Status          = failed
Client Description     = Failed tasks
Desired Status         = run
[...]

PR #24409, opened 2024-11-08
vault: catch expired lease as fatal error

tgross:
The errors we get aren't wrapped by the Vault SDK, so unfortunately we have to read the error messages and can't easily enumerate non-fatal error messages (which might be bubbling up from the stdlib). I've audited the errors currently used and have documented their source.

Ref https://github.com/hashicorp/vault/blob/52ba156d47da170bf40471fe57d72522030bdc7e/vault/expiration.go#L1327
Fixes: https://github.com/hashicorp/nomad/issues/23859

### DIFF
--- a/.changelog/24409.txt
+++ b/.changelog/24409.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+vault: Fixed a bug where expired secret leases were treated as non-fatal and retried
+```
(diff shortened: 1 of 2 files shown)


jrasell (approved):
LGTM!

PR #24410, opened 2024-11-08
consul: allow non-root Nomad to rewrite token

tgross:
When a task restarts, the Nomad client may need to rewrite the Consul token, but it's created with permissions that prevent a non-root agent from writing to it. While Nomad clients should be run as root (currently), it's harmless to allow whatever user the Nomad agent is running as to be able to write to it, and that's one less barrier to rootless Nomad.

(Note that the Vault token already has `u+rw` permissions.)

Ref: https://github.com/hashicorp/nomad/issues/23859#issuecomment-2465757392


jrasell (approved):
LGTM!

Issue #24411, opened 2024-11-09
1.9.2 stopped accepting existing tokens


hynek:
### Nomad version

1.9.2

### Operating system and Environment details

Ubuntu Jammy

### Issue

After an update of the servers from 1.9.1. to 1.9.2, our existing tokens are rejected as invalid. Both in UI as well as via API. Downgrading to 1.9.1 fixes that.

### Reproduction steps

Upgrade Nomad servers from 1.9.1. to 1.9.2.

#### Expected Result

Tokens still work.

#### Actual Result

Tokens don't work anymore.

### Nomad Server logs (if appropriate)

There is a warning:

> Setting a Vault token in the agent configuration is deprecated and will be removed in Nomad 1.9. Migrate your Vault configuration to use workload identity

The rest is just raft chatter.

### Nomad Client logs (if appropriate)

Clients are not involved.

maxadamo:
This absolutely appalling. 
It's not only about token: **the nomad cluster turned into a blank slate**. 
Blank slate means, everything was gone. All the jobs gone, all variables deleted, namespaces destroyed, and I had to bootstrap and reapply an ACL, create namespaces, re-run all the jobs. 

Unless you call it alpha or beta, **this is not acceptable**. You may still be shocked by the outcome of the elections, but you still need to pay closer attention, even if this is the community edition.

allroundtechie:
Can confirm the issue about the token (same message appeared like @hynek has posted).
The jobs were not all gone, a few "survived" the update to 1.9.2

tgross:
Hey folks, I saw this issue pop up and did some quick testing to confirm and reproduce. It looks like this slipped past upgrade testing by happening as part of the Raft snapshot restore -- I can't reproduce so far unless there's a Raft snapshot in play (which is pretty much always the case for production clusters but isn't typical in short-lived test clusters). Apologies for this... that's a pretty bad miss. Obviously this is top priority.

That being said, let's all keep in mind the [HashCorp Community Guidelines](https://www.hashicorp.com/community-guidelines) here.

maxadamo:
@tgross I apologise, but it's the second time in a few months that something like this is happening, early this morning I found the cluster wiped, and I started my weekend working, and I was at least disoriented and frustrated.

tgross:
Fix is up here: https://github.com/hashicorp/nomad/pull/24412. We'll get that reviewed on Monday and get it shipped out ASAP. 

[...]

PR #24412, opened 2024-11-09
fsm: fix bug in snapshot restore for removed timetable

tgross:
When we removed the time table in #24112 we introduced a bug where if a previous version of Nomad had written a time table entry, we'd return from the restore loop early and never load the rest of the FSM. This will result in a mostly or partially wiped state for that Nomad node, which would then be out of sync with its peers (which would also have the same problem on upgrade).

The bug only occurs when the FSM is being restored from snapshot, which isn't the case if you test with a server that's only written Raft logs and not snapshotted them.

While fixing this bug, we still need to ensure we're reading the time table entries even if we're throwing them away, so that we move the snapshot reader along to the next full entry.

Fixes: https://github.com/hashicorp/nomad/issues/24411

---

To minimally test:
* Run Nomad 1.9.1 or earlier (a single node cluster is ok, but not `-dev` mode).
* Run `nomad operator snapshot save /tmp/snapshot.tar.gz
* Run `nomad operator root keyring list` to see the current keyring (a good standin to see that the snapshot has been restored)
* Stop the agent
* Start the agent with this branch.
* Verify that you do not see the "initializing keyring" log line.
* Run `nomad operator root keyring list` again to see that the same key has been restored.

We should do some more extensive upgrade testing as well before merging this.

### DIFF
--- a/nomad/fsm.go
+++ b/nomad/fsm.go
@@ -1564,9 +1564,13 @@ func (n *nomadFSM) restoreImpl(old io.ReadCloser, filter *FSMFilter) error {
 		snapType := SnapshotType(msgType[0])
 		switch snapType {
 		case TimeTableSnapshot:
-			// COMPAT: Nomad 1.9.2 removed the timetable, this case kept to gracefully handle
-			// tt snapshot requests
-			return nil
+			// COMPAT: Nomad 1.9.2 removed the timetable, this case kept to
+			// gracefully handle tt snapshot requests
+			var table []TimeTableEntry
+			if err := dec.Decode(&table); err != nil {
+				return err
+			}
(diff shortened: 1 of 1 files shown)


pkazmierczak (approved):
I did similar testing to @jrasell, testing upgrades from 1.7.7 and 1.8.4 to https://github.com/hashicorp/nomad/commit/a67fc1febf4ccff65618ed42ee88acb6c445585f. State was successfully restored. 

Reading through #24112, the only place that touches the snapshot restore code path is indeed `restoreImpl` method in the FSM. Core scheduler changes only apply to GC methods, same as `BlockedEvals`.

PR #24418, opened 2024-11-11
fsm: handle EventSinkSnapshot type correctly

pkazmierczak:
As @tgross noticed in #24412, we never handled `EventSinkSnapshot` correctly. We should error on this snapshot type, since it should never be encountered in prod Nomad clusters.

### DIFF
--- a/nomad/fsm.go
+++ b/nomad/fsm.go
@@ -1798,7 +1798,7 @@ func (n *nomadFSM) restoreImpl(old io.ReadCloser, filter *FSMFilter) error {
 
 		// COMPAT(1.0): Allow 1.0-beta clusterers to gracefully handle
 		case EventSinkSnapshot:
-			return nil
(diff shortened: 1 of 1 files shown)


nomad/fsm.go
```diff
@@ -1798,7 +1798,7 @@ func (n *nomadFSM) restoreImpl(old io.ReadCloser, filter *FSMFilter) error {
 
 		// COMPAT(1.0): Allow 1.0-beta clusterers to gracefully handle
```
tgross, line 1799: We should probably adjust this comment (or even remove it, as the error message is pretty descriptive). Maybe add a `// DEPRECATED` to the 2 deprecated snapshot type constants above, as well?

pkazmierczak: 1d42abb

pkazmierczak: or better yet: 8786619

PR #476, opened 2015-11-20
Fixed some issues with expose, port mapping, and environment variables

cbednarski:
- Port mapping now works for reserved ports as well as dynamic ports
- Environment variables were being set twice in two different ways; now they are only set once
- Comprehensive tests for exposed ports, forwarded ports, and environment variables
- Cleaned up / DRYed up a lot of test code
- Flattened `[]map[string]int` to `map[string]int`

PR #24423, opened 2024-11-11
[gh-476] Sanitise HCL variables before storing on job submission

Juanadelacuesta:
### Description
Currently Nomad only handles HCL variables with new lines, any other non alphanumeric character is left untouched and stored unescaped, which can cause errors while re starting a stopped job, particularly from the UI.

it fixes [#476](https://github.com/hashicorp/terraform-provider-nomad/issues/476#issuecomment-2464341824)

### Contributor Checklist

### Reviewer Checklist
  backporting document.
  history should be preserved.
  within the public repository.

### DIFF
--- a/ui/app/utils/json-to-hcl.js
+++ b/ui/app/utils/json-to-hcl.js
@@ -16,7 +16,8 @@ export default function jsonToHcl(obj) {
 
   for (const key in obj) {
     const value = obj[key];
-    const hclValue = typeof value === 'string' ? `"${value}"` : value;
+    const hclValue = typeof value === 'string' ? JSON.stringify(value) : value;
+
     hclLines.push(`${key}=${hclValue}\n`);
   }
 


api/jobs.go
```diff
@@ -1061,9 +1061,7 @@ func (js *JobSubmission) Canonicalize() {
 	// characters to preserve them. This way, when the job gets stopped and
 	// restarted in the UI, variable values will be parsed correctly.
 	for k, v := range js.VariableFlags {
-		if strings.Contains(v, "\n") {
-			js.VariableFlags[k] = strings.ReplaceAll(v, "\n", "\\n")
-		}
+		js.VariableFlags[k] = url.QueryEscape(v)
```
tgross, line 1064: This fixes the Terraform provider and the CLI, because they both use this `api` package, but does the UI already have an equivalent to this fix? (cc @philrenaud )

Should we add a warning to the API documentation for the [`JobSubmission` field](https://developer.hashicorp.com/nomad/api-docs/jobs#submission) that [third-party callers](https://developer.hashicorp.com/nomad/tools#community-tools) of the HTTP API need to perform this operation as well?

tgross (approved):
LGTM

jrasell (approved):
LGTM!

philrenaud:
Testing this out made me remember we have two different variable-related properties in Job submission data: `VariableFlags`, which this PR addresses and are passed via CLI props, and `Variables`, which are what get entered via -var-file, etc.: https://github.com/hashicorp/nomad/blob/e206993d4990937225ef2c185809895ea7a357e9/api/jobs.go#L978-L984

I think that VariableFlags are affected here, but Variables are not. I'm not sure if the bug in question _could_ show up in that format, but wanted to mention it.

=========================

All that aside, I'm also noticing some perhaps erroneous character additions. Here are screenshots showing VariableFlags and output for a job where a cli-provided multi-line variable has a `+` character added:

on `main`:
Screenshot 2024-11-18 at 2 44 53 PM

on `f-NET-11421-vars`:
Screenshot 2024-11-18 at 2 42 37 PM

command to run was:

```
nomad job run -var meta_string="line1 \
\
\
\
line2" ../nomad-job-specs/unescaped_var_test.nomad.hcl
```

Note the `+` added between "line1" and "line2"

api/jobs.go
philrenaud: The UI runs output here through [this jsonToHcl() helper method](https://github.com/hashicorp/nomad/blob/main/ui/app/utils/json-to-hcl.js#L14-L24)

It adds `\n` between multiple VariableFlags / Variables entries, and the [codemirror editor](https://codemirror.net/) we use for rendering generally handles them correctly. It's not perfect, though: Your example in the test, 

```
"test": `"foo": "bar"`
```

renders like this:

image

which isn't perfect but it's probably enough of an edge case that it's been okay since we've had HCL-in-UI.

============

I'm not sure if this is what url.QueryEscape() is doing, but if we're given output like what's in the test below (%22foo%22%3A+%22bar%22), we will show that verbatim
image

philrenaud (approved):

tgross (approved):
LGTM!

PR #24450, opened 2024-11-13
drivers: fix capabilities on non-linux systems

pkazmierczak:
In #24093 we moved from github.com/syndtr/gocapability to github.com/moby/sys/capability due to the former package no longer being maintainer. The new package's capability function works differently: the known/supported functionality is split now, and the `.ListSupported()` call will always return an empty list on non-linux systems. This means Nomad agents won't start on darwin or windows.

### DIFF
--- a/drivers/shared/capabilities/defaults.go
+++ b/drivers/shared/capabilities/defaults.go
@@ -6,6 +6,7 @@ package capabilities
 import (
 	"fmt"
 	"regexp"
+	"runtime"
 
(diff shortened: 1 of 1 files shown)


Juanadelacuesta (approved):
Thanks!

philrenaud (approved):
Thank you!

tgross (approved):
LGTM.

I'd have thought the Windows tests would have picked this up, as we definitely run clients there; at least enough of one to run some driver tests.

PR #24454, opened 2024-11-13
[cli] UI URL hints for common CLI commands

philrenaud:
Adds a "View more in the Web UI: $url" message at the end of common CLI commands to give users a way to dive deeper with a single click (or copy/paste)

image

Modify the `ui` block of your agent config with the following to disable: `cli_url_links = false`

Add a `-ui` flag to several common commands (`nomad job status` for example) for your default browser to open to the suggested page

### DIFF
--- a/command/job_run.go
+++ b/command/job_run.go
@@ -98,7 +98,7 @@ Run Options:
     the job.
 
   -ui
-    Open the job run page in the browser.
+    Open the job page in the browser.
 
   -policy-override
     Sets the flag to force override any soft mandatory Sentinel policies.
--- a/command/job_run.go
+++ b/command/job_run.go
@@ -97,6 +97,9 @@ Run Options:
     Output the JSON that would be submitted to the HTTP API without submitting
     the job.
 
+  -ui
+    Open the job run page in the browser.
+
   -policy-override
     Sets the flag to force override any soft mandatory Sentinel policies.
 
@@ -147,6 +150,7 @@ func (c *JobRunCommand) AutocompleteFlags() complete.Flags {
 			"-var":              complete.PredictAnything,
 			"-var-file":         complete.PredictFiles("*.var"),
 			"-eval-priority":    complete.PredictNothing,
+			"-ui":               complete.PredictNothing,
 		})
 }
 
@@ -161,7 +165,7 @@ func (c *JobRunCommand) AutocompleteArgs() complete.Predictor {
 func (c *JobRunCommand) Name() string { return "job run" }
 
 func (c *JobRunCommand) Run(args []string) int {
-	var detach, verbose, output, override, preserveCounts bool
+	var detach, verbose, output, override, preserveCounts, openURL bool
 	var checkIndexStr, consulNamespace, vaultNamespace string
 	var evalPriority int
 
@@ -180,6 +184,7 @@ func (c *JobRunCommand) Run(args []string) int {
 	flagSet.Var(&c.JobGetter.Vars, "var", "")
 	flagSet.Var(&c.JobGetter.VarFiles, "var-file", "")
 	flagSet.IntVar(&evalPriority, "eval-priority", 0, "")
+	flagSet.BoolVar(&openURL, "ui", false, "")
 
 	if err := flagSet.Parse(args); err != nil {
 		return 1
@@ -254,6 +259,18 @@ func (c *JobRunCommand) Run(args []string) int {
 		}
 
 		c.Ui.Output(string(buf))
+
+		hint, _ := c.Meta.showUIPath(UIHintContext{
+			Command: "job run",
+			PathParams: map[string]string{
+				"jobID": *job.ID,
+			},
+			OpenURL: openURL,
+		})
+		if hint != "" {
+			c.Ui.Output(hint)
+		}
+
 		return 0
 	}
 
@@ -321,10 +338,33 @@ func (c *JobRunCommand) Run(args []string) int {
 			c.Ui.Output("Evaluation ID: " + evalID)
 		}
 
+		hint, _ := c.Meta.showUIPath(UIHintContext{
+			Command: "job run",
+			PathParams: map[string]string{
+				"jobID": *job.ID,
+			},
+			OpenURL: openURL,
+		})
+		if hint != "" {
+			c.Ui.Output(hint)
+		}
+
 		return 0
 	}
 
 	// Detach was not specified, so start monitoring
+	// Set hint to open the browser
+	hint, _ := c.Meta.showUIPath(UIHintContext{
+		Command: "job run",
+		PathParams: map[string]string{
+			"jobID": *job.ID,
+		},
+		OpenURL: openURL,
+	})
+	if hint != "" {
+		c.Ui.Output(hint)
+	}
+
 	mon := newMonitor(c.Ui, client, length)
 	return mon.monitor(evalID)
 
--- a/command/job_run.go
+++ b/command/job_run.go
@@ -260,17 +260,6 @@ func (c *JobRunCommand) Run(args []string) int {
 
 		c.Ui.Output(string(buf))
 
-		hint, _ := c.Meta.showUIPath(UIHintContext{
-			Command: "job run",
-			PathParams: map[string]string{
-				"jobID": *job.ID,
-			},
-			OpenURL: openURL,
-		})
-		if hint != "" {
-			c.Ui.Output(hint)
-		}
-
 		return 0
 	}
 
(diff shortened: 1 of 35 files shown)


website/content/docs/configuration/ui.mdx
```diff
@@ -106,7 +111,7 @@ header sent with the web UI response.
 - `text` `(string: "")` - Specifies the text of the label that will be
   displayed in the header of the Web UI.
 - `background_color` `(string: "")` - The background color of the label to
-  be displayed. The Web UI will default to a black background. HEX values  
+  be displayed. The Web UI will default to a black background. HEX values
```
aimeeu, line 114:
```suggestion
  be displayed. The Web UI defaults to a black background. HEX values
```

present tense (I know this clashes with the copious amounts of future tense already in the docs. Eventually I will address that.)

aimeeu (approved):
a few style guide nits

command/meta.go
```diff
@@ -427,3 +428,173 @@ type funcVar func(s string) error
...
+	client, err := m.Client()
+	if err != nil {
+		return true
+	}
+	agent, err := client.Agent().Self()
+	if err != nil {
+		return true
+	}
+	agentConfig := agent.Config
+	uiConfig, ok := agentConfig["UI"].(map[string]interface{})
```
schmichael, line 594: I think this needs to be locally configurable instead of clusterwide configurable. For example it's probably desirable on operator machines and bastion hosts, but potentially undesirable on actual Nomad Clients where the CLI is only used for maintenance/debugging.

Since the UI is often proxied across many agents, if not the entire cluster, configuring this parameter clusterwide seems difficult.

I think we probably want to implement something like #24706 or at least an env var to let this behavior be user configurable.

We should probably be careful about doing implicit API calls from the CLI as well. Even though `Agent().Self()` should be cheap for the responding Agent, users running scripts over potentially slow links may see a significant slowdown from the extra roundtrips.

command/meta.go
```diff
@@ -427,3 +428,173 @@ type funcVar func(s string) error
...
+	if err != nil {
+		return "", fmt.Errorf("error getting client config: %v", err)
+	}
+
+	path := route.Path
+	for k, v := range params {
+		path = strings.ReplaceAll(path, fmt.Sprintf(":%s", k), v)
+	}
+
+	return fmt.Sprintf("%s/ui%s", client.Address(), path), nil
```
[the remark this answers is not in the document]
philrenaud: Yeah, that's a good point. I am actually worried that might sink this whole feature, the more I think about it.

Do we have any way of knowing at a config level if there is a reverse proxy set up? Or is that more extra-nomad than we can know about?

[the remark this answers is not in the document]
schmichael: > Do we have any way of knowing at a config level if there is a reverse proxy set up? Or is that more extra-nomad than we can know about?

Sadly I don't think we have anything today to do that, but a `nomad.ui_url` agent configuration parameter like [consul and vault](https://developer.hashicorp.com/nomad/docs/configuration/ui#consul-parameters) have would be sufficient. That would still require an `Agent().Self()` call to retrieve unless we did something very clever in the Agent HTTP API like
sniff the User-Agent, and if it's the Nomad CLI: return an HTTP header with the `ui_url`. As hacky as that sounds I think returning extra metadata to the CLI might be a useful pattern?

command/meta.go
philrenaud: I added a `NOMAD_SHOW_CLI_HINTS` check here: https://github.com/hashicorp/nomad/pull/24454/commits/3c8b7b27f94f5590f4708a25a8c6dbbfadc1ad82

command/meta.go
```diff
@@ -427,3 +428,173 @@ type funcVar func(s string) error
...
+}
+
+func (m *Meta) showUIPath(ctx UIHintContext) (string, error) {
```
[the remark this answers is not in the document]
philrenaud: Hm, so far I guess I've only been using it in a Must.NotError() test context.

Would it be better for me to do error handling within this method? It feels extensible as it stands but I don't know how I'd extend it.

command/meta.go
```diff
@@ -427,3 +428,173 @@ type funcVar func(s string) error
...
+	if err != nil {
+		return true
+	}
+	agentConfig := agent.Config
+	uiConfig, ok := agentConfig["UI"].(map[string]interface{})
+	if !ok {
+		return true
+	}
+
+	return !uiConfig["CLIURLLinks"].(bool)
```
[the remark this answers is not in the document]
philrenaud: I've made this a bit longer but hopefully no longer panic-inducing?

```
	agent, err := client.Agent().Self()
	if err != nil {
		return true
	}

	agentConfig := agent.Config
	agentUIConfig, ok := agentConfig["UI"].(map[string]interface{})
	if !ok {
		return false
	}

	agentShowCLIHints, ok := agentUIConfig["ShowCLIHints"].(bool)
	if !ok {
		return false
	}

	if !agentShowCLIHints {
		// check to see if env var is set to true, overriding the agent setting
		if m.showCLIHints != nil && *m.showCLIHints {
			return false
		}
		return true
	}
```

command/meta.go
philrenaud: The implicit API calls from the CLI bit: this has been mentioned by a couple others, too. Is there a non-api-polling way I could do this? We have a `SetupUi()` method in command meta; maybe I could read the agent config once there and save that to a meta value for all subsequent calls?

command/meta.go
```diff
@@ -427,3 +438,197 @@ type funcVar func(s string) error
...
+		return true
+	}
+
+	agent, err := client.Agent().Self()
+	if err != nil {
+		return true
+	}
+
+	agentConfig := agent.Config
+	agentUIConfig, ok := agentConfig["UI"].(map[string]interface{})
```
tgross, line 615:
```suggestion
	agentUIConfig, ok := agentConfig["UI"].(map[string]any)
```

command/meta.go
```diff
@@ -427,3 +438,197 @@ type funcVar func(s string) error
...
+	// Next, check if the agent config is set to false. If it is, return early.
+	client, err := m.Client()
+	if err != nil {
+		return true
+	}
+
+	agent, err := client.Agent().Self()
+	if err != nil {
+		return true
+	}
```
tgross, line 612: Addressing a comment made in our slack discussion: although this is making another API call, _most_ of our APIs end up making multiple API calls anyways so I don't feel like this is an enormous burden. And it'll drive us towards having some kind of local config file to govern the Nomad CLI too, which is something we've discussed a bunch previously.

command/job_run.go
```diff
@@ -372,10 +383,36 @@ func (c *JobRunCommand) Run(args []string) int {
...
+		PathParams: map[string]string{
+			"jobID":     *job.ID,
+			"namespace": jobNamespace,
+		},
+		OpenURL: openURL,
+	})
+	if hint != "" {
+		c.Ui.Output(hint)
+		// Because this is before monitor, newline so we don't scrunch
+		c.Ui.Output("\n")
```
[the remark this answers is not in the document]
philrenaud: I think I forgot that c.ui.output inherently goes on a new line; changing it to `c.Ui.Output("")` now. Thanks!

schmichael (commented):
Sorry for letting this linger so long

command/meta.go
```diff
@@ -427,3 +438,197 @@ type funcVar func(s string) error
...
+
+	return fmt.Sprintf("%[1]s%[2]s%[3]s%[4]s%[5]s %[6]s%[7]s%[8]s",
+		bold,
+		magenta,
+		uiHintDelimiter[1:], // "==> "
+		description,
+		resetter,
+		blue,
+		url,
+		resetter,
```
schmichael, line 552: No need to change but at this point just doing `strings.Join([]string{ /* long list */ }, "")` is probably more clear.

command/meta.go
```diff
@@ -248,6 +251,14 @@ func (m *Meta) SetupUi(args []string) {
...
 		}
 	}
+
+	// Check to see if the user has disabled hints via env var.
+	showCLIHints := os.Getenv(EnvNomadCLIShowHints)
+	if showCLIHints == "false" {
+		m.showCLIHints = pointer.Of(false)
+	} else if showCLIHints == "true" {
+		m.showCLIHints = pointer.Of(true)
+	}
```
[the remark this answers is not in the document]
schmichael: We should also only show hints if `isTerminal`. Not sure if this is the right place to flip the flag or if we should just short-circuit in `formatUIHint()`. Either is fine as long as we ensure `!isTerminal` takes precedence over the API response and explicit configuration here.

schmichael (approved):

aimeeu:
@philrenaud On my list to review in the next week (by 2/28 at the latest)

schmichael (dismissed):

aimeeu (dismissed):
lgtm thanks!

schmichael (approved):

aimeeu (approved):

PR #24456, opened 2024-11-13
scheduler: fix a bug where force GC wasn't respected

### LINKED ISSUE #24455: Nomad system gc not working as expected
### Nomad version
1.9.3

MacOS and Ubuntu

### Issue
Running `nomad system gc` does not garbage collect stopped jobs.

### Reproduction steps
Run a job, stop it, and then force garbage collection via `nomad system gc`.

#### Expected Result
The job is garbage collected.

#### Actual Result
The job is still shown via `nomad job status`.

### Job file (if appropriate)
N/A

N/A

N/A

pkazmierczak:
This PR fixes a bug where `System.GarbageCollect` endpoint didn't work on objects that weren't older than their respective GC thresholds. `System.GarbageCollect` is used to force garbage collection (also used by the `system gc` command) and should ignore any GC threshold settings. 

Fixes #24455
Internal ref: https://hashicorp.atlassian.net/browse/NET-11747

### DIFF
--- a/nomad/core_sched.go
+++ b/nomad/core_sched.go
@@ -12,6 +12,7 @@ import (
 
 	log "github.com/hashicorp/go-hclog"
 	memdb "github.com/hashicorp/go-memdb"
+	"github.com/hashicorp/go-multierror"
 	version "github.com/hashicorp/go-version"
 	"github.com/hashicorp/nomad/helper/uuid"
 	"github.com/hashicorp/nomad/nomad/state"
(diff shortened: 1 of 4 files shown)


nomad/core_sched.go
```diff
@@ -28,29 +28,42 @@ type CoreScheduler struct {
...
+		"batchEval",
+		"deployment",
+		"csiPlugin",
+		"csiVolume",
+		"token",
+		"node",
+	} {
+		c.setCustomThresholdForObject(objectName, threshold)
+	}
+}
```
schmichael, line 65: My main concern with this kind of approach is that we're relying on matching strings to tables/objects to funcs.

What if instead we passed the threshold into the funcs directly from Process? Process already does string parsing of the Eval (which is a different string than this map keys off of!), so I don't think it would make Process too much more complicated to pass the appropriate threshold from config into funcs. It _might_ make testing easier to? You can pass thresholds in directly to gc funcs or you can customize the config. I would hope that would cover every case.

pkazmierczak: > Since we don't add new gc-able objects often, it seems even more likely we'd forget the force gc path.

That is, sadly, very likely. 

> What if instead we passed the threshold into the funcs directly from Process?

I looked at that but it's a major change. Not only would we have to change every call ro `Process` (of which there are _many_), but we'd also need to change the scheduler interface. Not saying we shouldn't do this, just saying it's a big change.

schmichael: > Not only would we have to change every call to Process

Do we have to call Process or just change the code _inside_ [CoreScheduler.Process](https://github.com/hashicorp/nomad/blob/13657e02dae2e536d2b510e62efdd26dc2d8f3df/nomad/core_sched.go#L58-L81)?

For example could the call to `c.jobGC(eval)` be changed to `c.jobGC(eval, c.srv.config.JobGCThreshold)` to avoid the config lookup inside `jobGC` itself?

mismithhisler: The passed in threshold could still be overridden by custom values, so I don't think it would break tests.  Although it does feel weird to pass in a struct field thats already available to the receiver.
Which makes me wonder if (I think your original idea) of just passing in a forceGC boolean might be the best way to go here.

pkazmierczak: Right. Passing a `force` boolean to every GC method will still be very messy. Because I can't think of a better way of forcing than setting a very very small threshold value. So then for every GC method we'd have to check 3 places where a threshold might've been set: config, custom field in the scheduler, or force bool induced minimal value. That, to me, is even less elegant.

pkazmierczak: @schmichael @mismithhisler Gentlemen I heard your complaints, and I present to you 1d149a0. 

We no longer keep stringy hash maps, and we simplify the way force GC happens. And yet we still let people test "time manipulation" by forcing custom threshold values. See if it pleases you.

mismithhisler: I am pleased.

mismithhisler (approved):
LGTM!

Issue #15489, opened 2022-12-07
Dynamic Host Volumes


akamensky:
### Proposal
Currently to create `host` volume one needs to edit agent configuration to add `volume` stanza and restart the agent. This is impractical as Nomad itself may be provisioned using one of numerous tools (Ansible/Salt/etc) and restarting agent that may already have other tasks running is far from good idea.

CSI volumes can be created/allocated on the fly. However CSI volumes are often networked storage (like NFS). Host volumes are extremely useful for stateful workloads that require high performance local storage (think local SSD array or NVMe for databases like rocksdb for example).

I think allowing to create host volumes on agent nodes on the fly using API calls (and perhaps with corresponding CLI commands) is sensible and very practical approach.

### Use-cases
* High performance local DBs like rocksdb/leveldb which can be built-in into application itself
* On disk cache that would speed up application startup (when storage speed matters)
* Maybe other use cases too, but above is our primary use cases for this

### Attempted Solutions
* raw_exec driver with direct access to host filesystems. This however is not an ideal solution as this removes the process isolation.

akamensky:
@tgross thank you for the response. Not implying any rush with this, just wondering what's the possible ETA on this feature landing in stable release (how long would it normally take from proposal until feature available)? We are evaluating Nomad as a replacement homebrewed deployment/orchestration system for mostly legacy stack and this may be a show-stopper for us.

tgross:
For reasons of "we're a public company now and I'll get put in Comms Jail" :grinning: I can't give an ETA on new features but it's almost certainly not going to get worked on until after 1.5.0 which will go out early in the new year.

mikenomitch:
FYI, I changed the name on this feature to make it easier for us to find internally. We tend to refer to it as "Dynamic Host Volumes" so just updating the title to match that.

akamensky:
I will ask this here first as not sure if this should be bundled in with this FR or raised as a separate ticket. If any of you @tgross or @mikenomitch could provide a feedback on this would appreciate it.

I know with `docker` driver it is possible to provide a simple directory mapping in config, such as:

```
    driver = "docker"
    config {
      volumes = ["/data/test_dir:/data"]
    }
```

Which then utilizes Dockers functionality to do the bind mount.

[...]

PR #24479, opened 2024-11-18
Dynamic Host Volumes

tgross:
Feature integration branch for dynamic host volumes.

Closes: https://github.com/hashicorp/nomad/issues/15489
Ref: https://hashicorp.atlassian.net/browse/NET-11259


tgross:
Ready for merge. All commits on this branch have been reviewed under other PRs.

philrenaud (approved):

ygersie:
@tgross great to see the volume alignment here! One feature request which I hope may be added to the default built-in plugin, can it please implement two `Parameters` in the volume request for `mask` and `uid:gid`? This is going to be extremely important to move forward with adoption for us and it feels like it would be a waste to have to write a custom plugin just for this.

gulducat:
Nice suggestion @ygersie ! Makes sense to want a chown/chmod option on the built-in `mkdir` plugin.  What do you think of this as the specification?

```hcl
type      = "host"
name      = "cool-host-vol"
plugin_id = "mkdir"
parameters = {
  mode = "0755" # default: "0700"
  uid  = 1000   # or "1000" -- default: the user running the client agent
  gid  = 1000   # or "1000"
}
```

ygersie:
Yes that’d be great @gulducat !

PR #24480, opened 2024-11-18
nsutil: wrap error that comes from the syscall so caller can do errors.As

gabivlj:
### Description

User of `nsutil` library should be able to do the following and for it to work:

```
   var errno syscall.Errno
   if errors.As(err, &errno) {
       if errno == unix.EBUSY { ... }
   }
```

This commit fixes that issue.

Call the function and try to do errors.As error handling, it won't work.

### Links
...

### Contributor Checklist
- [x] **Changelog Entry** No user facing behavior
- [x] **Testing** Seems like a pretty simple fix, I can add a test for nsutil if reviewer thinks its necessary
- [x] **Documentation** Doesn't apply.


pkazmierczak (approved):
LGTM, thanks for the contribution @gabivlj!

PR #24489, opened 2024-11-19
agent: set content type header explicitly

pkazmierczak:
This PR addresses an XSS vulnerability where Nomad agents wouldn't explicitly set content type headers for error responses. 

Internal ref: https://hashicorp.atlassian.net/browse/SECVULN-14567

### DIFF
--- a/command/agent/http.go
+++ b/command/agent/http.go
@@ -746,9 +746,9 @@ func (s *HTTPServer) wrap(handler func(resp http.ResponseWriter, req *http.Reque
 				}
 			}
 
+			resp.Header().Set(contentTypeHeader, plainContentType)
 			resp.WriteHeader(code)
 			resp.Write([]byte(errMsg))
-			resp.Header().Set(contentTypeHeader, plainContentType)
 			if isAPIClientError(code) {
 				s.logger.Debug("request failed", "method", req.Method, "path", reqURL, "error", err, "code", code)
 			} else {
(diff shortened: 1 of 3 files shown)


command/agent/http.go
```diff
@@ -745,6 +748,7 @@ func (s *HTTPServer) wrap(handler func(resp http.ResponseWriter, req *http.Reque
 
 			resp.WriteHeader(code)
 			resp.Write([]byte(errMsg))
+			resp.Header().Set(contentTypeHeader, plainContentType)
```
schmichael, line 751:
```suggestion
			resp.Header().Set(contentTypeHeader, plainContentType)
			resp.WriteHeader(code)
			resp.Write([]byte(errMsg))
```

I think setting the header needs to come before the other 2 response calls.

pkazmierczak: oh, does it matter? I didn't realize. I'll update the rest of occurrences.

dduzgun-security: setting the headers first ensures that the proper metadata / content-type can be sent back which reduces the risk possible wrong encoding or other response manipulation

pkazmierczak: ah `WriteHeader` writes the header immediately to the response object 🤦‍♂️ 
"it's all coming back to me now"

pkazmierczak: fixed in 7410b7312efcb2eac422615d8be138cd2ca3606d

schmichael (approved):

dduzgun-security (approved):

Issue #11958, opened 2022-01-28
resource leaking when unsupported caps used for exec driver


tantra35:
### Nomad version
Output from `Nomad v1.1.10 (2f08fe230da05e1b179710ebe0e2582249599a4b+CHANGES)`

### Operating system and Environment details
Ubuntu 20.04

### Issue
If we use unhallowed caps for exec driver after faill we got leeaking nomad exec processes

### Reproduction steps
For example if we use `net_raw` caps that doens't allowed by default for exec driver
```
job testnetworknamespace
{
	region = "global"
	datacenters = ["test"]

	update
	{
		stagger = "1m"
		min_healthy_time = "1m"
		max_parallel = 1
		health_check="checks"
		healthy_deadline = "3m"
		progress_deadline = "6m"
		auto_revert = true
	}

	group testservicecheck
	{
		restart {
			attempts = 2
			delay    = "15s"
		}

		task testservicecheck
		{
			driver = "exec"
			leader=true

			config
			{
				cap_add = ["net_raw"]

				command = "sleep"
				args = ["6000"]
			}

			logs
			{
				max_files = 3
				max_file_size = 10
			}

			resources
			{
				memory = 300
				cpu = 100
			}
		}
	}
} 
```

after allocation on node fail with follow task state(which is absolutely expected behavior)
```
Recent Events:
Time                       Type            Description
2022-01-28T20:22:47+03:00  Killing         Sent interrupt. Waiting 5s before force killing
2022-01-28T20:22:47+03:00  Not Restarting  Error was unrecoverable
2022-01-28T20:22:47+03:00  Driver Failure  driver does not allow the following capabilities: net_raw
2022-01-28T20:22:45+03:00  Task Setup      Building Task Directory
2022-01-28T20:22:40+03:00  Received        Task received by client
```

on client node we got leaked `nomad executor` processes (here we demonstrate some output of `ps axuf`)
```
dnsmasq    33659  0.0  0.2  13932  2088 ?        S    19:16   0:00 /usr/sbin/dnsmasq -x /run/dnsmasq/dnsmasq.pid -u dnsmasq -7 /etc/dnsmasq.d,.dpkg-dist,.dpkg-old,.dpkg-new --local-service --trust-anchor=.,20326,8,2,e0
root       33756  0.6  5.6 1363452 56400 ?       Ssl  19:16   0:25 /opt/nomad/nomad agent -config=/etc/nomad/
[...]

PR #24495, opened 2024-11-19
drivers: update ordering of events in StartTask to fix executor leak

mismithhisler:
### Description

If an error occurs before a task is started, but after this executor process is created, the executor must be explicity stopped.  This change updates the logic in StartTask so launching the task happens immediately after creating the executor.

Run a job with a configuration that would fail after the executor process has started, but before the task is launched.  For example, an exec task with `cap_add = ["net_raw"]`.  After the job fails, the executor process will still be running.

### Links

Fixes [GH #11958](https://github.com/hashicorp/nomad/issues/11958)
### Contributor Checklist

### Reviewer Checklist
  backporting document.
  history should be preserved.
  within the public repository.

### DIFF
--- a/.changelog/24495.txt
+++ b/.changelog/24495.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+drivers: fix executor leak when drivers error starting tasks
+```
--- a/drivers/exec/driver.go
+++ b/drivers/exec/driver.go
@@ -492,13 +492,6 @@ func (d *Driver) StartTask(cfg *drivers.TaskConfig) (*drivers.TaskHandle, *drive
 		Compute:     d.compute,
 	}
 
-	exec, pluginClient, err := executor.CreateExecutor(
-		d.logger.With("task_name", handle.Config.Name, "alloc_id", handle.Config.AllocID),
-		d.nomadConfig, executorConfig)
-	if err != nil {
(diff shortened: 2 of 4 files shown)


mismithhisler:
I wonder if using named return values would be a better way to prevent a regression.  With named return values, we can defer a function that checks `if err != nil`, and kill the plugin client.  This way we don't have to worry about accidentally leaving it running in an error scenario.

jrasell:
> I wonder if using named return values would be a better way to prevent a regression. With named return values, we can defer a function that checks `if err != nil`, and kill the plugin client. This way we don't have to worry about accidentally leaving the executor process running in any added codepaths in the future.
> 
> We could also write a single test that forces any error after the executor is created, and that should suffice to catch this bug in most other error scenarios.

If you think it's a solution worth exploring, then I would encourage you to look into it. The testing aspect in particular would be very useful in my opinion. The current code in this PR looks like it would fix the problem, so we can always fallback here if needed.

jrasell (commented):
LGTM and thanks @mismithhisler!

I tested this on a Linux workstation with the details supplied and it was easy to witness the fix.

I noticed the added test takes 12s to run which seems a little long. It might be a follow up to see if this can be lowered. Interestingly, the `NoOrphanedTasks` test is also long at 27s, so it might be something these tests need.

jrasell (approved):

Issue #24512, opened 2024-11-20
Nomad 1.9.+ : crash


lmorel3:
### Nomad version
1.9.3

### Operating system and Environment details

![image](https://github.com/user-attachments/assets/f5d14e6b-4bf8-4048-97e6-36042cd028e0)

### Issue

I've upgraded from Nomad 1.4.3 to 1.9.3 and everything works fine... except that sometimes the server (not clients) crash and I need to restart it (started via a windows service).

Can't understand why.

### Nomad Server logs (if appropriate)
```
==> WARNING: mTLS is not configured - Nomad is not secure without mTLS!
==> WARNING: Bootstrap mode enabled! Potentially unsafe operation.
==> Failed to check for updates: Get "https://checkpoint-api.hashicorp.com/v1/check/nomad?arch=amd64&os=windows&signature=1f63914d-5e2a-c019-ae76-2330089cda17&version=1.9.3": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x8 pc=0x2642f7b]

goroutine 10 [running]:
github.com/hashicorp/nomad/command/agent/consul.(*ServiceClient).sync(0xc0006ebd40, 0x0)
	github.com/hashicorp/nomad/command/agent/consul/service_client.go:1108 +0xf1b
github.com/hashicorp/nomad/command/agent/consul.(*ServiceClient).Run(0xc0006ebd40)
	github.com/hashicorp/nomad/command/agent/consul/service_client.go:891 +0x407
created by github.com/hashicorp/nomad/command/agent/consul.(*ServiceClientWrapper).Run in goroutine 1
	github.com/hashicorp/nomad/command/agent/consul/service_client.go:488 +0xb5
```

lmorel3:
Wow, dude, you're so quick! Thank you :)

Good job.

lmorel3:
@tgross have you planned to release soon? :)

PR #24513, opened 2024-11-20
consul: handle nil multierror pointer correctly

tgross:
When the service client syncs to Consul, we accumulate service sync errors in a multierror before reading all the local checks. If the API call to the local checks fails, we either return that error or append it to the multierror and return the set of errors. But `multierror.Error.Len()` doesn't nil-check, so we need to do this ourselves.

I've also made a quick pass through the rest of the code base looking for multierror `Len` method calls to see if we have this pattern elsewhere.

Fixes: https://github.com/hashicorp/nomad/issues/24512

### DIFF
--- a/.changelog/24513.txt
+++ b/.changelog/24513.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+consul: Fixed a bug where failures when syncing Consul checks could panic the Nomad agent
+```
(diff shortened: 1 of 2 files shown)


pkazmierczak:
I'll let you be the judge if this warrants a changelog entry or not. Also: backport labels?

PR #24521, opened 2024-11-20
dynamic host volumes: update volume from node fingerprint

tgross:
When dynamic host volumes are created, they're written to the state store in a "pending" state. Once the client fingerprints the volume it's eligible for scheduling, so we mark the state as ready at that point.

Because the fingerprint could potentially be returned before the RPC handler has a chance to write to the state store, this changeset adds test coverage to verify that upserts of pending volumes check the node for a previously-fingerprinted volume as well.


nomad/state/state_store_host_volumes.go
```diff
@@ -190,3 +190,35 @@ func (s *StateStore) hostVolumesIter(ws memdb.WatchSet, index string, sort SortO
...
+				vol = vol.Copy()
+				vol.State = structs.HostVolumeStateReady
+				vol.ModifyIndex = index
+				err = txn.Insert(TableHostVolumes, vol)
+				if err != nil {
+					return fmt.Errorf("host volume insert: %w", err)
+				}
+			}
+		default:
+			// don't touch ready or soft-deleted volumes
```
gulducat, line 221: what's meant by "soft-deleted" here?  deleted on the client without a respective ->server->rpc->state call?

gulducat (approved):
lgtm!

nomad/state/state_store_host_volumes.go
tgross: There's a `HostVolumeStateDeleted` in our state enum. That gives us the option to delete the volume from the client via the RPC but leave the record of it on the server until GC'd, with the reasoning being that operators might care about being able to track that sort of thing. The RPC doesn't currently implement that, because I haven't quite reasoned though the consequences of GC'ing them properly.

Issue #24339, opened 2024-10-31
Allocation Metrics no longer emitted


dosera:
### Nomad version
```
Nomad v1.9.1
BuildDate 2024-10-21T09:00:50Z
Revision d9ec23f0c1035401e9df6c64d6ffb8bffc555a5e
```

### Operating system and Environment details
`AlmaLinux release 9.4 (Seafoam Ocelot)`

`/etc/nomad/base.hcl`:

```
telemetry {
..
  prometheus_metrics = "true"
  publish_allocation_metrics = "true"
  publish_node_metrics = "true"

..
}
```

### Issue
After upgrading nomad from `1.8.4` to `1.9.1` allocation metrics like `nomad_client_allocs_memory_usage` appear not to be properly emitted (or only for several seconds - i am using prometheus to scrape them).

### Reproduction steps
1. Enable telemetry
2. Start any allocation
3. Collect metrics

#### Expected Result
Allocation metrics are available.

#### Actual Result
Allocation metrics are available sporadically (at best). See the attached screenshot - nomad was upgraded on Oct 24th.

This is how it looks in prometheus since then:
![nomad no alloc metrics](https://github.com/user-attachments/assets/e17688bf-52e5-45af-a1dc-95f5958760c4)

Himura2la:
We have the same issue with allocation metrics after updating from 1.8 to 1.9:
![image](https://github.com/user-attachments/assets/36ec3798-2456-422e-91d7-61dadbdd7ed9)
query: `avg_over_time(nomad_client_allocs_memory_usage[10m]) / avg_over_time(nomad_client_allocs_memory_allocated[10m])`

They are returned irregularly, and often values are zero.

our configs:
```hcl
telemetry {
  collection_interval = "30s"
  disable_hostname = true
  prometheus_metrics = true
  publish_allocation_metrics = true
  publish_node_metrics = true
}
```

```yml
global:
  scrape_interval:     1m
  scrape_timeout:     20s
  evaluation_interval: 1m

  - job_name: nomad
    static_configs:
      - targets:
        # ...
    scheme: https
    metrics_path: /v1/metrics
    params:
      format: [prometheus]
    tls_config:
      insecure_skip_verify: true
    relabel_configs:
      - source_labels: [__address__]
        target_label: instance

```

jrasell:
[...]

PR #24525, opened 2024-11-21
[gh-24339] Move from streaming stats to polling for docker

Juanadelacuesta:
### Description
It addresses [24339](https://github.com/hashicorp/nomad/issues/24339). 

Currently to get the stats from a container we are using the  call to `ContainerStats` which returns a stream of data that can be reused, but it only works if the sampling interval for the metrics is `1s` exactly. With other values, decoding errors start to show up and the metric values becomes inconsistent:

```
2024-11-21T13:14:17.645+0100 [DEBUG] client.driver_mgr.docker: error decoding stats data from container: container_id=2317edae0f7a0e1fb2e259a2e87f2575cf6114a2c2c47c3eaf7f992de0b7a569 driver=docker error="json: cannot unmarshal string into Go value of type container.Stats"
2024-11-21T13:14:17.651+0100 [DEBUG] client.driver_mgr.docker: error decoding stats data from container: container_id=f062a6097af27095c1e8eec76005d33c670edaadb2e5b74540f83704f0260a60 driver=docker error="invalid character 'a' looking for beginning of value"
```
It might have gone unnoticed because the error messages were displayed only for debug, which rarely happens on production environments.

To avoid the sync problem, this PR introduces a polling call to the stats endpoint, which is not as efficient, but is more versatile and allows for different sampling intervals.

To see the bug, take the latest release and start a nomad agent, it can be in -dev mode, with telemetry enabled and a `collection_interval` different from `1s`.

```
telemetry {
  collection_interval        = "3s"
  prometheus_metrics         = true
  publish_allocation_metrics = true
}
```
Run a job with multiple allocations and after a while the errors will show and the metrics for the allocations will be unstable. 

### Contributor Checklist

### Reviewer Checklist
  backporting document.
  history should be preserved.
  within the public repository.

### DIFF
--- a/drivers/docker/stats.go
+++ b/drivers/docker/stats.go
@@ -16,8 +16,6 @@ import (
 	cstructs "github.com/hashicorp/nomad/client/structs"
 	"github.com/hashicorp/nomad/drivers/docker/util"
 	"github.com/hashicorp/nomad/helper"
-
-	//"github.com/hashicorp/nomad/helper"
 	nstructs "github.com/hashicorp/nomad/nomad/structs"
 )
 


drivers/docker/stats.go
```diff
@@ -16,6 +16,8 @@ import (
 	cstructs "github.com/hashicorp/nomad/client/structs"
 	"github.com/hashicorp/nomad/drivers/docker/util"
 	"github.com/hashicorp/nomad/helper"
+
+	//"github.com/hashicorp/nomad/helper"
```
tgross, line 20: Let's clean this up.

tgross (commented):
LGTM in terms of implementation, but can we describe how this has been tested in this PR? The original fix to use streaming during the 1.9.0-beta was to fix a leaked goroutine, so I really want to make sure we're not regressing on problems like that.

Juanadelacuesta:
> LGTM in terms of implementation, but can we describe how this has been tested in this PR? The original fix to use streaming during the 1.9.0-beta was to fix a leaked goroutine, so I really want to make sure we're not regressing on problems like that.

Do you mean tests in terms of no leaks? Maybe there was a previous PR touching this same code, but the latest was about using a different call to get a more complete stats report from docker.

tgross:
> Do you mean tests in terms of no leaks? Maybe there was a previous PR touching this same code, but the latest was about using a different call to get a more complete stats report from docker.

Right, I'm just asking that you verified this doesn't leak any goroutines from the called functions. You can do it manually, I just want to make sure we've done so, so that we don't have a regression.

Juanadelacuesta:
I did not, but I will non

Juanadelacuesta:
Tests to detect leaking go routines were manually run using [`goroutine-inspect`](https://github.com/tgross/goroutine-inspect) by checking for go routines before, during and after running a docker workload with multiple allocations and verifying the go back to the initial number.

tgross (approved):
LGTM!

Himura2la:
Does it really address #24339 ?

Issue #24049, opened 2024-09-23
make Consul/Vault fingerprints reloadable, not periodic


tgross:
Nomad fingerprinters can be static, periodic, or "reloadable" (update on [SIGHUP](https://developer.hashicorp.com/nomad/docs/operations/nomad-agent#sighup)). The only two periodic fingerprints are for Consul and Vault. One could also argue that these two fingerprints are unusual in that they fingerprint something other than the host's "own" environment, and this can cause flapping fingerprints if the availability of those services isn't 100%.

So in Nomad 1.4.0 we shipped https://github.com/hashicorp/nomad/pull/14673 which changed the fingerprinters so that we wouldn't change the fingerprint if the up/down status of Consul or Vault changed. But this still leaves the rest of the fingerprint like the version. For Vault, this can make Vault cluster upgrades perilous for large Nomad clusters. If your Vault cluster is behind a DNS name, your Nomad clients will fingerprint different Vault IPs behind that name as you upgrade, causing the fingerprint to change each time. Multiply this by many many clients and you have a storm of `Node.Register` requests and the resulting Raft writes and evals to handle.

Instead, let's move the power to decide when to re-fingerprint Vault and Consul in the cluster administrator's hands. Make these fingerprinters reloadable rather than periodic. We can combine this work with https://github.com/hashicorp/nomad/issues/24048 to make reloading configuration easier as well.

(See also https://github.com/hashicorp/nomad/issues/23526 https://github.com/hashicorp/nomad/issues/18327 for other examples of fingerprints that could be reloadable.)

PR #24526, opened 2024-11-21
fingerprint: convert consul and vault fingerprinters to be reloadable

shoenig:
This PR changes the Consul and Vault fingerprint implementations to be
reloadable rather than periodic. Reasons described in the issue.

Closes: https://github.com/hashicorp/nomad/issues/24049


jrasell (approved):
LGTM!

PR #24531, opened 2024-11-21
dynamic host volumes: search endpoint

tgross:
Add support for dynamic host volumes to the search endpoint. Like many other objects with UUID identifiers, we're not supporting fuzzy search here, just prefix search on the fuzzy search endpoint.

Because the search endpoint only returns IDs, we need to seperate CSI volumes and host volumes for it to be useful. The new context is called `"host_volumes"` to disambiguate it from `"volumes"`. In future versions of Nomad we should consider deprecating the `"volumes"` context in lieu of a `"csi_volumes"` context.

---

To test end-to-end, create sufficient volumes to have at least 2 matching prefix characters in the ID (this may take quite a few tries!).

```
$ nomad volume create ./demo/hostvolume/host.volume.hcl
Created host volume test with ID f6a8949a-55dd-f6cc-985c-216fe98989c8

$ nomad volume create ./demo/hostvolume/host.volume.hcl
Created host volume test with ID f6f8ed01-0f4d-c300-a8da-02e0c78e3dde
```

Then you can use `curl` and get the expected results as shown below. I also verified that the queries the UI makes get the correct results, but because the UI doesn't yet have the models built for host volumes, it doesn't know what do to with the host volumes context and shows no results.

<details><summary>Fuzzy search results</summary>

```
$ curl -sd '{"Text": "f6", "Context": "all", "Namespace": "*"}' "http://localhost:4646/v1/search/fuzzy" | jq .
{
  "Index": 18,
  "KnownLeader": true,
  "LastContact": 0,
  "Matches": {
    "host_volumes": [
      {
        "ID": "f6a8949a-55dd-f6cc-985c-216fe98989c8"
      },
      {
        "ID": "f6f8ed01-0f4d-c300-a8da-02e0c78e3dde"
      }
    ],
    "evals": [],
    "deployment": [],
    "volumes": [],
    "scaling_policy": []
  },
  "NextToken": "",
  "Truncations": {
    "volumes": false,
    "scaling_policy": false,
    "host_volumes": false,
    "jobs": false,
    "nodes": false,
    "vars": false,
    "namespaces": false,
    "deployment": false,
    "node_pools": false,
    "plugins": false,
    "allocs": false,
    "evals": false
  }
}
```

</details>

<details><summary>Prefix search results</summary>

```
$ curl -sd '{"Prefix": "f6", "Context": "all", "Namespace": "*"}' "http://localhost:4646/v1/search" | jq .
{
  "Index": 18,
  "KnownLeader": true,
  "LastContact": 0,
  "Matches": {
    "evals": null,
    "vars": null,
    "namespaces": null,
    "volumes": null,
    "allocs": null,
    "node_pools": null,
    "scaling_policy": null,
    "jobs": null,
    "nodes": null,
    "deployment": null,
    "plugins": null,
    "host_volumes": [
      "f6a8949a-55dd-f6cc-985c-216fe98989c8",
      "f6f8ed01-0f4d-c300-a8da-02e0c78e3dde"
    ]
  },
  "NextToken": "",
  "Truncations": {
    "host_volumes": false,
    "volumes": false,
    "allocs": false,
    "nodes": false,
    "node_pools": false,
    "deployment": false,
    "plugins": false,
    "evals": false,
    "vars": false,
    "namespaces": false,
    "scaling_policy": false,
    "jobs": false
  }
}
```

</details>

### DIFF
--- a/nomad/search_endpoint_test.go
+++ b/nomad/search_endpoint_test.go
@@ -1039,6 +1039,53 @@ func TestSearch_PrefixSearch_CSIVolume(t *testing.T) {
 	require.False(t, resp.Truncations[structs.Volumes])
 }
 
+func TestSearch_PrefixSearch_HostVolume(t *testing.T) {
+	ci.Parallel(t)
+
+	srv, cleanup := TestServer(t, func(c *Config) {
+		c.NumSchedulers = 0
+	})
+	defer cleanup()
+	codec := rpcClient(t, srv)
+	testutil.WaitForLeader(t, srv.RPC)
+
+	store := srv.fsm.State()
+	index, _ := store.LatestIndex()
+
+	node := mock.Node()
+	index++
+	must.NoError(t, store.UpsertNode(structs.MsgTypeTestSetup, index, node))
+
+	id := uuid.Generate()
+	index++
+	err := store.UpsertHostVolumes(index, []*structs.HostVolume{{
+		ID:        id,
+		Name:      "example",
+		Namespace: structs.DefaultNamespace,
+		PluginID:  "glade",
+		NodeID:    node.ID,
+		NodePool:  node.NodePool,
+	}})
+	must.NoError(t, err)
+
+	req := &structs.SearchRequest{
+		Prefix:  id[:6],
+		Context: structs.HostVolumes,
+		QueryOptions: structs.QueryOptions{
+			Region:    "global",
+			Namespace: structs.DefaultNamespace,
+		},
+	}
+
+	var resp structs.SearchResponse
+	must.NoError(t, msgpackrpc.CallWithCodec(codec, "Search.PrefixSearch", req, &resp))
+
+	must.Len(t, 1, resp.Matches[structs.HostVolumes])
+	must.Len(t, 0, resp.Matches[structs.Volumes])
+	must.Eq(t, id, resp.Matches[structs.HostVolumes][0])
+	must.False(t, resp.Truncations[structs.HostVolumes])
+}
+
 func TestSearch_PrefixSearch_Namespace(t *testing.T) {
 	ci.Parallel(t)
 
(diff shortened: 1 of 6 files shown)


nomad/state/state_store_host_volumes.go
```diff
@@ -156,6 +157,30 @@ func (s *StateStore) HostVolumes(ws memdb.WatchSet, sort SortOption) (memdb.Resu
...
+	iter, err := s.hostVolumesIter(ws, indexID, sort)
+	if err != nil {
+		return nil, err
+	}
+	wrappedIter := memdb.NewFilterIterator(iter, func(raw any) bool {
+		vol, ok := raw.(*structs.HostVolume)
+		if !ok {
+			return true
+		}
+		return !strings.HasPrefix(vol.ID, prefix)
```
gulducat, line 179: this took me a sec to parse, so I'm typing it in english to make sure I grasp it:

https://pkg.go.dev/github.com/hashicorp/go-memdb#FilterFunc

> FilterFunc is a function that takes the results of an iterator and returns whether the result should be filtered out. 

`true` = filter it out (do not keep it)
`false` = do not filter it (keep it)

so,
`if !ok` - if it is not a `HostVolume` (somehow), `true` = throw it away
`if !strings.HasPrefix` - if it does not have the prefix, `true` = throw it away

nomad/search_endpoint_test.go
```diff
@@ -1932,6 +1979,52 @@ func TestSearch_FuzzySearch_CSIVolume(t *testing.T) {
...
+		Name:      "example",
+		Namespace: structs.DefaultNamespace,
+		PluginID:  "glade",
+		NodeID:    node.ID,
+		NodePool:  node.NodePool,
+	}})
+	must.NoError(t, err)
+
+	req := &structs.FuzzySearchRequest{
+		Text:    id[0:3], // volumes are prefix searched
```
gulducat, line 2012: this doesn't quite line up with my conception of "fuzzy", but looks like others are like this too.  I suppose the "fuzziness" is that it could match different types of things, not just a prefix search on volumes specifically.

nomad/state/state_store_host_volumes_test.go
```diff
@@ -163,6 +163,12 @@ func TestStateStore_HostVolumes_CRUD(t *testing.T) {
 	must.NoError(t, err)
 	got = consumeIter(iter)
 	must.MapLen(t, 3, got, must.Sprint(`expected 3 volumes remain`))
+
+	prefix := vol.ID[:30] // sufficiently long prefix to avoid flakes
+	iter, err = store.HostVolumesByIDPrefix(nil, "*", prefix, SortDefault)
```
gulducat, line 168: only covering wildcard `"*"`, not a specific namespace?

gulducat (approved):
lgtm!

nomad/state/state_store_host_volumes.go
tgross: Right, it's "filter-out", not "filter-for". I feel like most of Nomad's code uses "filter-for" but this is coming from go-memdb where it's the opposite.

nomad/search_endpoint_test.go
tgross: Right for a lot of other objects it really is fuzzy and will match on fields other than ID. But for CSI volumes we did it only by ID so I'm reproducing that behavior here. It might be worth revisiting to see whether we can expand the prefix search for volumes and other objects more widely in general.

nomad/state/state_store_host_volumes_test.go
tgross: The specific namespace case was trivial because it's just calling an already-testing function, but can't hurt to add a case for it. Done.

gulducat: I tend to be paranoid about stuff like this, less about the code as currently written, but more to guard against potential future regressions.

PR #24533, opened 2024-11-21
dynamic host volumes: autocomplete for CLI

tgross:
_WIP: currently includes the commit in https://github.com/hashicorp/nomad/pull/24531 which is needs to work. Will keep in draft until that's merged_

Adds dynamic host volumes to argument autocomplete for the `volume status` and `volume delete` commands. Adds flag autocompletion for those commands plus `volume create`.


pkazmierczak (approved):
LGTM!

PR #24535, opened 2024-11-21
dynamic host volumes: test client RPC and plugins

gulducat:
adding tests for #24497 

also happened to notice we weren't validating that volume ID was uuid-shaped, so add that too.

main issue: #24479

### DIFF
--- a/client/hostvolumemanager/host_volume_plugin_test.go
+++ b/client/hostvolumemanager/host_volume_plugin_test.go
@@ -1,0 +1,216 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package hostvolumemanager
+
+import (
+	"bytes"
+	"context"
+	"io"
+	"path/filepath"
+	"runtime"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-version"
+	cstructs "github.com/hashicorp/nomad/client/structs"
+	"github.com/shoenig/test"
+	"github.com/shoenig/test/must"
+)
+
+func TestHostVolumePluginMkdir(t *testing.T) {
+	volID := "test-vol-id"
+	tmp := t.TempDir()
+	target := filepath.Join(tmp, volID)
+
+	plug := &HostVolumePluginMkdir{
+		ID:         "test-mkdir-plugin",
+		TargetPath: tmp,
+		log:        hclog.Default(),
+	}
+
+	// contexts don't matter here, since they're thrown away by this plugin,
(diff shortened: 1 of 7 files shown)


nomad/structs/host_volumes.go
```diff
@@ -167,7 +171,7 @@ func (hv *HostVolume) Validate() error {
 		}
 	}
 
-	return mErr.ErrorOrNil()
+	return helper.FlattenMultierror(mErr.ErrorOrNil())
```
tgross, line 174: Nice QoL catch.

client/hostvolumemanager/host_volume_plugin_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+func TestHostVolumePluginMkdir(t *testing.T) {
+	volID := "test-vol-id"
+	tmp := t.TempDir()
+	target := filepath.Join(tmp, volID)
+
+	plug := &HostVolumePluginMkdir{
+		ID:         "test-mkdir-plugin",
+		TargetPath: tmp,
+		log:        hclog.Default(),
```
tgross, line 30: There's a `helper/testlog` package that has some nice features like only emitting logs on failure (or `go test -v`) and including the location in the log outputs:

```suggestion
		log:        testlog.Logger(t),
```

Unfortunately it doesn't have a recorder, which would have been handy for your other test below.

client/hostvolumemanager/host_volume_plugin_test.go
```diff
@@ -0,0 +1,216 @@
...
+)
+
+func TestHostVolumePluginMkdir(t *testing.T) {
+	volID := "test-vol-id"
+	tmp := t.TempDir()
+	target := filepath.Join(tmp, volID)
+
+	plug := &HostVolumePluginMkdir{
+		ID:         "test-mkdir-plugin",
+		TargetPath: tmp,
```
tgross, line 29: This makes me realize that calling this field "target path" is maybe a little misleading, as it's the parent of the eventual target path.

tgross (approved):
LGTM!

client/hostvolumemanager/host_volume_plugin_test.go
gulducat: looks like I need a `testlog.HCLogger(t)` here, but I like!

client/hostvolumemanager/host_volume_plugin_test.go
gulducat: yeah, aside from "target", even "path" is ambiguous: could signify either a file or a dir.  I'll be in this vicinity again with fingerprinting, so I'll keep this in mind.

PR #24536, opened 2024-11-22
docs: add missing MinQuorum docs

schmichael:
Also update Autopilot API examples.

### Reviewer Checklist
  backporting document.
  history should be preserved.

### DIFF
--- a/website/content/api-docs/operator/autopilot.mdx
+++ b/website/content/api-docs/operator/autopilot.mdx
@@ -30,23 +30,23 @@ The table below shows this endpoint's support for
 ### Sample Request
 
 ```shell-session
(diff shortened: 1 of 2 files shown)


jrasell (commented):
If we are going to change the curl commands to use the `operator api`, should we do this across all, rather than have have mix?

aimeeu:
@schmichael Is this docs PR still relevant?

aimeeu:
@schmichael This PR is kinda old... The docs content has migrated to the web-unified-docs repo, so this content needs to be recreated there. I can do that if this PR is still relevant.

tgross:
Definitely still relevant.

PR #24538, opened 2024-11-22
executor: fail early on reattach if listener is not executor

mismithhisler:
### Description

Fixes a bug reported by Windows users where Nomad was killing random processes on reboot.  This was due to the reattach config containing a PID and RPC address that was now occupied by other processes/listeners.  When Nomad would attempt to make an RPC, it would fail, and kill the process. With this change, we fail early on reattach, and do not kill the process.

Reproduction involves simulating this scenario via overriding the reattach PID when starting a task to a known PID (another bash session), rebooting Nomad, killing the executor process, and listening on the reattach address.

### Links

Fixes [GH #23969](https://github.com/hashicorp/nomad/issues/23969)

### Contributor Checklist

### Reviewer Checklist
  backporting document.
  history should be preserved.
  within the public repository.

### DIFF
--- a/drivers/shared/executor/utils.go
+++ b/drivers/shared/executor/utils.go
@@ -75,7 +75,10 @@ func CreateExecutor(
 	return newExecutorClient(config, logger)
 }
 
-// ReattachToExecutor launches a plugin with a given plugin config
+// ReattachToExecutor launches a plugin with a given plugin config and validates it can call the executor.
+// On Windows, go-plugin listens on a localhost port. It is possible on a reboot that another process is
+// listening on that port, and a process is running with the previous executors PID, leading Nomad to kill
+// "random" processes. Fail early via the Version RPC if we detect the listener isn't actually an Executor.
 func ReattachToExecutor(reattachConfig *plugin.ReattachConfig, logger hclog.Logger, compute cpustats.Compute) (Executor, *plugin.Client, error) {
 	config := &plugin.ClientConfig{
 		HandshakeConfig:  base.Handshake,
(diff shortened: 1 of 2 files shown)


drivers/shared/executor/utils.go
```diff
@@ -75,7 +75,10 @@ func CreateExecutor(
 	return newExecutorClient(config, logger)
 }
 
-// ReattachToExecutor launches a plugin with a given plugin config
+// ReattachToExecutor launches a plugin with a given plugin config and validates it can call the executor.
+// On Windows, go-plugin listens on a localhost port. It is possible on a reboot that another process is
+// listening on that port, and a process is running with the previous executors PID, leading Nomad to kill
+// "random" processes. Fail early via the Version RPC if we detect the listener isn't actually an Executor.
```
tgross, line 81: Great addition to the comment. Were you able to to track down where the kill was being emitted downstream? It might be nice to mention that here, as I know both of us spent a good bit of time trying to find it.

tgross (approved):
LGTM! Nice work on this!

drivers/shared/executor/utils.go
mismithhisler: In my reproduction I noticed the downstream call happening [here in the taskRunner](https://github.com/hashicorp/nomad/blob/main/client/allocrunner/taskrunner/task_runner.go#L707) after a failed RPC happened.  I will update the comment to add this context.

drivers/shared/executor/utils.go
```diff
@@ -84,7 +88,14 @@ func ReattachToExecutor(reattachConfig *plugin.ReattachConfig, logger hclog.Logg
 		AllowedProtocols: []plugin.Protocol{plugin.ProtocolGRPC},
 		Logger:           logger.Named("executor"),
 	}
-	return newExecutorClient(config, logger)
+	exec, pluginClient, err := newExecutorClient(config, logger)
+	if err != nil {
+		return nil, nil, err
+	}
+	if _, err := exec.Version(); err != nil {
```
peter-harmann-tfs, line 95: Is reattach always done before new executors are launched? Otherwise, is it possible that the process that is listening on the previous port is just a different executor and the Version call succeeds despite the PID not being an executor? Would it make more sense to check process name or creation time to ensure we are re-attaching the correct process?

peter-harmann-tfs: Great. Thank you for clarifying.

peter-harmann-tfs (approved):

jrasell (approved):
LGTM, thanks @mismithhisler!

PR #24542, opened 2024-11-22
[ui] Volumes and plugins navigation fixes, generally

### LINKED ISSUE #23831: CSI volumes cannot be displayed in Nomad 1.8 GUI
### Nomad version
Nomad 1.8.2

### Issue
When loading the "Storage" page, CSI volumes are initially listed, but clicking on one triggers a 404 error, and the page tries to dynamically load more data, which also triggers 404 errors.

It looks like the UI is trying to call URLs with this pattern:

```
$NOMAD_ADDR/v1/plugin/csi%2Fplugin-name
$NOMAD_ADDR/v1/volume/csi%2Fvolume-name
```

The correct URLs appear to be:

```
$NOMAD_ADDR/v1/plugin/csi/plugin-name
$NOMAD_ADDR/v1/volume/csi/volume-name
```

Note the extraneous `/` URL escaping.

### Reproduction steps
Load https://nomad.url/ui/csi/volumes

#### Expected Result
Volume list is displayed, volume details can be viewed by clicking on a volume.

#### Actual Result
Volume list is displayed momentarily, but soon goes blank (or shows "No Volumes").

nomad-volume-list

Clicking on a volume while it's displayed shows a "Not Found" error.

nomad-volume-details

philrenaud:
### Description
Mainly, this fixes a bug where navigating to a volume from the Storage page would error. This was happening because it was escaping the `/` after the `csi/` prefix that Ember Data uses to route things. I made a [similar bugfix a few months ago](https://github.com/hashicorp/nomad/pull/23625) for plugins, but didn't think to check volumes as well

Additionally, found a few more quality-of-life around navigation within storage:
- There was no keyboard nav for the plugins index table
- the `u` keyboard shortcut would pass more info than needed to get from `/volumes/:id` back to `/volumes` (this is because it's one level more nested than most breadcrumb up-a-level nav; /storage/plugins and /storage are the same page, as far as breadcrumbs are concerned)
- Enumerated keyboard nav to load a volume from the index page was not working, even after the above `csi/` slash escaping bug was fixed; this was because the DOM event being passed to our `lazyClick` method (to simulate clicking on a row when you type in a keynav combo, in this case) was coming back null. There's now a defensive Elvis (`?.`) to return a falsy value in lazyClick.

Make some volumes/plugins and navigate around to your heart's content

Resolves #23831


tgross (commented):
I pulled this down locally to try it out and LGTM in terms of functionality.

It looks like we broke some of the unit tests for volumes in the process though, and I think the Mirage fixtures need to be updated because the preview on Vercel won't let you click through to the individual volume pages.

pkazmierczak (approved):
LGMT in terms of functionality, and from what I can tell JS looks good, but my confidence in that area is low.

PR #24545, opened 2024-11-22
dynamic host volumes: Enterprise stubs and refactor API

tgross:
Most Nomad upsert RPCs accept a single object with the notable exception of CSI. But in CSI we don't actually expose this to users except through the Go API. It deeply complicates how we present errors to users, especially once Sentinel policy enforcement enters the mix.

Refactor the `HostVolume.Create` and `HostVolume.Register` RPCs to take a single volume instead of a slice of volumes.

Add a stub function for Enterprise policy enforcement. This requires splitting out placement from the `createVolume` function so that we can ensure we've completed placement before trying to enforce policy.

### DIFF
--- a/nomad/host_volume_endpoint_ce.go
+++ b/nomad/host_volume_endpoint_ce.go
@@ -1,0 +1,23 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+//go:build !ent
+// +build !ent
+
+package nomad
+
+import (
+	"github.com/hashicorp/nomad/nomad/state"
+	"github.com/hashicorp/nomad/nomad/structs"
+)
+
+// enforceEnterprisePolicy is the CE stub for Enterprise governance via
+// Sentinel policy, quotas, and node pools
+func (v *HostVolume) enforceEnterprisePolicy(
+	_ *state.StateSnapshot,
+	_ *structs.HostVolume,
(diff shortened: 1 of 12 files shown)


nomad/host_volume_endpoint_ce.go
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+//go:build !ent
+// +build !ent
```
tgross, line 5: Note for reviewers: https://github.com/hashicorp/nomad-enterprise/commit/5890c50480738bfecb7eca9560218dacea82d690 has my rough prototype for what this will look like in Nomad Enterprise, so the signature here should be what we want.

pkazmierczak (approved):
Code LGTM, but if we now start returning objects instead of arrays on create and register API calls, shouldn't this be considered a breaking change somehow? People will still be able to make the same requests, but whatever consumes their responses will break.

tgross:
> if we now start returning objects instead of arrays on create and register API calls, shouldn't this be considered a breaking change somehow? 

It would be! But this is for a feature we haven't yet shipped (not even merged to `main` yet), so no backwards compatibility concerns.

pkazmierczak:
> But this is for a feature we haven't yet shipped

🤦‍♂️ 
sorry, lost track of what's "static" and what's dynamic host volumes there for a sec.

PR #24546, opened 2024-11-22
release: always use service user for git ops

gulducat:
We are shoring up our user and bot accesses, and this seems to be our only usage of the github-actions bot.

These were the commits that have been flagged as needing adjustment:

```
workflows/release.yml
145:            git commit --message "Generate files for ${{ github.event.inputs.version }} release"
212:            git commit --message 'Prepare for next release'
```

We may later discover other of our repos that need a similar treatment, if they've just not been run during this information-gathering time, but at present this is the only known case.  Our CE->ENT merge workflow in the ENT repo is already set up properly.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,6 +53,8 @@ jobs:
             exit 1
           fi
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
       - name: Retrieve Vault-hosted Secrets
         if: endsWith(github.repository, '-enterprise')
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -215,5 +216,5 @@ jobs:
           fi
 
 permissions:
-  contents: write
+  contents: read
```
tgross, line 219: This wasn't mentioned in the PR description. I know we've had to set this field to `write` in our workflows explicitly previously but that it was unexpected. Why is it safe to change it back now?

tgross (approved):
LGTM :+1:

PR #24556, opened 2024-11-27
client: emit optional telemetry from prerun and prestart hooks.

jrasell:
The Nomad client can now optionally emit telemetry data from the prerun and prestart hooks. This allows operators to monitor and alert on failures and time taken to complete.

The new datapoints are:
  - nomad.client.alloc_hook.prerun.success (counter)
  - nomad.client.alloc_hook.prerun.failed (counter)
  - nomad.client.alloc_hook.prerun.elapsed (sample)

  - nomad.client.task_hook.prestart.success (counter)
  - nomad.client.task_hook.prestart.failed (counter)
  - nomad.client.task_hook.prestart.elapsed (sample)

The hook execution time is useful to Nomad engineering and will help optimize code where possible and understand job specification impacts on hook performance.

Currently only the PreRun and PreStart hooks have telemetry enabled, so we limit the number of new metrics being produced.

### Contributor Checklist

### Reviewer Checklist
  backporting document.
  history should be preserved.
  within the public repository.


client/allocrunner/alloc_runner_hooks.go
```diff
@@ -189,7 +189,17 @@ func (ar *allocRunner) prerun() error {
...
 
-		if err := pre.Prerun(); err != nil {
+		// If the operator has disabled hook metrics, then don't call the time
+		// function to save 30ns per hook.
+		var hookExecutionStart time.Time
+
+		if !ar.clientConfig.DisableAllocationHookMetrics {
+			hookExecutionStart = time.Now()
+		}
+
```
shoenig, line 199: Just a thought, but what if the HookStatsHandler interface also required a `Start()` implementation that set the start time, which would be kept as internal state of the handler (and would be a noop of the noop handler).

shoenig (approved):
LGTM!

client/allocrunner/alloc_runner_hooks.go
jrasell: I did think of that while writing the code.

I decided against it as all the hooks that emit telemetry use a single `HookStatsHandler` within the task or alloc runner. It would need to assume each hook is called without any concurrency with a blind assumption the start time is meant for the hook when calling Emit. This felt wrong to me.

I also considered having a `HookStatsHandler` implementation per hook call, but did not go down this route to avoid overhead in setting this up per hook and the memory overhead with additional label setup and such.

Let me know your thoughts, I am happy to change the approach if we feel the advantages are worth it and our assumptions are correct.

PR #24558, opened 2024-11-27
stateful deployments: find feasible node for sticky host volumes

pkazmierczak:
This changeset implements node feasibility checks for sticky host volumes.


scheduler/feasible.go
```diff
@@ -212,6 +212,7 @@ func (h *HostVolumeChecker) hasVolumes(n *structs.Node) bool {
 					capOk = true
 					break
 				}
+				return false
```
tgross, line 215: We shouldn't return early here -- a volume can have multiple requested capabilities and we only need one of them to match.

scheduler/generic_sched.go
```diff
@@ -910,6 +911,30 @@ func (s *GenericScheduler) findPreferredNode(place placementResult) (*structs.No
...
+		preferredNode, err := s.state.NodeByID(ws, prev.NodeID)
+		if err != nil {
+			return nil, err
+		}
+
+		if preferredNode != nil && preferredNode.Ready() {
+			// if this node has at least one of the allocation volumes, it's a
+			// preferred one
+			for _, vol := range preferredNode.HostVolumes {
+				if slices.Contains(prev.HostVolumeIDs, vol.VolumeID) {
```
tgross, line 931: Daniel's work named this field `.ID`

scheduler/feasible.go
```diff
@@ -216,6 +216,27 @@ func (h *HostVolumeChecker) hasVolumes(n *structs.Node) bool {
...
 				return false
 			}
+			if req.Sticky {
+				// NOTE: surely there is a better way to find the right alloc?
+				// Should we perhaps search for allocs by job? Could there be a
+				// situation in which there are non-terminal allocations
+				// belonging to the job in question that are sticky, have the
+				// volume IDs that match what the node offers and should not
+				// end up in this check?
+				allocs, err := h.ctx.ProposedAllocs(n.ID)
```
tgross, line 226: The only other place we use `ctx.ProposedAllocs` for feasibility checking is in the distinct hosts checker, because in that case we need to know about _all_ the allocations on the node, including those that might already exist. Do we really want to look at the old allocs here?

scheduler/feasible.go
```diff
@@ -216,6 +216,27 @@ func (h *HostVolumeChecker) hasVolumes(n *structs.Node) bool {
...
+				if err != nil {
+					continue
+				}
+
+				for _, a := range allocs {
+					if a.TerminalStatus() || a.NodeID != n.ID {
+						continue
+					}
+
+					return slices.Contains(a.HostVolumeIDs, volCfg.ID)
```
tgross, line 236: If we return early, we only end up checking a single allocation. But we could have multiple allocations that are mounting the same volume. So we need to return false if the slice is non-empty and doesn't contain the `volCfg.ID`, but let it fall through otherwise.

nomad/structs/structs.go
```diff
@@ -11114,6 +11114,13 @@ type Allocation struct {
 	// AllocatedResources is the total resources allocated for the task group.
 	AllocatedResources *AllocatedResources
 
+	// HostVolumeIDs is a list of host volume IDs that this allocation
+	// requires.
+	HostVolumeIDs []string
+
+	// CSIVolumeIDs is a list of CSI volume IDs that this allocation requires.
+	CSIVolumeIDs []string
```
tgross, line 11122: Clarifies that this is the volumes it's actually been given, not the ones it's asking for:

```suggestion
	// HostVolumeIDs is a list of host volume IDs that this allocation
	// has claimed.
	HostVolumeIDs []string

	// CSIVolumeIDs is a list of CSI volume IDs that this allocation has claimed.
	CSIVolumeIDs []string
```

scheduler/feasible_test.go
```diff
@@ -366,6 +366,86 @@ func TestHostVolumeChecker_ReadOnly(t *testing.T) {
...
+
+	alloc := mock.Alloc()
+	alloc.NodeID = nodes[1].ID
+	alloc.HostVolumeIDs = []string{dhv.ID}
+
+	cases := []struct {
+		name   string
+		node   *structs.Node
+		expect bool
+	}{
```
tgross, line 427: We should include a case for a sticky volume request that hasn't been previously claimed.

pkazmierczak:
I think making the allocation ID available to the feasibility checker makes sense. It's a bit of additional plumbing in the reconciler, but not bad, and knowing exactly which allocation wants which volume makes the `hasVolume` code more reliable and cleaner. Let me know what you think about this @tgross.

scheduler/feasible.go
```diff
@@ -200,32 +209,54 @@ func (h *HostVolumeChecker) hasVolumes(n *structs.Node) bool {
...
+				if slices.Contains(allocation.HostVolumeIDs, vol.ID) {
+					return true, ""
+				}
+
+				// if an allocation doesn't have a volume ID associated with
+				// it, update it
+				if len(allocation.HostVolumeIDs) == 0 {
+					allocation.HostVolumeIDs = []string{vol.ID}
+					// TODO: figure out how to update allocation. Should we
+					// have a new RPC endpoint for this?
```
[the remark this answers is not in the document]
tgross: If an `Allocation` here is a new or updated allocation, then it's getting submitted already to the leader via the `SubmitPlan` RPC.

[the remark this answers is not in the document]
pkazmierczak: You mean just append this alloc to Plan?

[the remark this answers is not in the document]
tgross: Ah, you're right... for some reason I was thinking we had the new `Allocation` already handy but we're calling `Select` on the task group, not the allocation. Maybe trying to add it here doesn't make sense -- can we add it _after_ we've selected the node? (ex. could we write it before we submit the plan, or even as part of plan apply?)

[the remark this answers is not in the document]
tgross: Addendum to this: the `computePlacements` function is where we annotate allocations with things like the previous allocation and the reschedule tracker. That suggests a good place to record which volume ID we've given the allocation: we have `Node` and the `Allocation` both handy and we know at that point that they're paired up (because we've _selected_ the node here in `hasVolumes`).

scheduler/generic_sched.go
```diff
@@ -657,6 +658,21 @@ func (s *GenericScheduler) computePlacements(destructive, place []placementResul
...
+					if v.Sticky {
+						if missing.PreviousAllocation() != nil && len(missing.PreviousAllocation().HostVolumeIDs) > 0 {
+							continue
+						}
+						vol, ok := option.Node.HostVolumes[v.Source]
+						if ok {
+							newHostVolumeIDs = append(newHostVolumeIDs, vol.ID)
+						}
+					}
+				}
```
tgross, line 674: I think we can add the host volume IDs unconditionally at this point? We have a selected alloc-node pair and so any dynamic host volumes the alloc wants _must_ be the correct volume IDs.

scheduler/feasible.go
```diff
@@ -137,87 +139,115 @@ func NewRandomIterator(ctx Context, nodes []*structs.Node) *StaticIterator {
...
+				allocation, err := h.ctx.State().AllocByID(ws, req.allocID)
+				if err != nil {
+					return false, FilterConstraintHostVolumesAllocLookupFailed
+				}
+				if slices.Contains(allocation.HostVolumeIDs, vol.ID) || len(allocation.HostVolumeIDs) == 0 {
+					return true, ""
+				}
+
+				return false, FilterConstraintHostVolumes
+			}
```
tgross, line 239: Rather than going back to the state store, let's add the `Allocation.HostVolumeIDs` to the [`SelectOptions`](https://github.com/hashicorp/nomad/blob/v1.9.3/scheduler/stack.go#L37-L42) struct. We can populate that value [here](https://github.com/hashicorp/nomad/blob/v1.9.3/scheduler/generic_sched.go#L827) in `getSelectOptions`, where we have access to the host. Add a `[]string` parameter for the host volume IDs to `SetVolumes` instead of the alloc ID, and then you can just do a `slices.Contains` check here on that value.

That _should_ get us what we want in terms of propagating the value from the previous allocation.

pkazmierczak:
@tgross I addressed your comments. Here's what we do now:
- we store `hostVolumeIDs` instead of `allocIDs` like you suggested—it's much easier indeed
- in `computePlacements` we either add the volume IDs to allocations or carry them over if they had them before. 

I added some tests and I think this is going the right direction. What do you think?

scheduler/stack.go
```diff
@@ -349,7 +351,7 @@ func (s *SystemStack) Select(tg *structs.TaskGroup, options *SelectOptions) *Ran
 	s.taskGroupDrivers.SetDrivers(tgConstr.drivers)
 	s.taskGroupConstraint.SetConstraints(tgConstr.constraints)
 	s.taskGroupDevices.SetTaskGroup(tg)
-	s.taskGroupHostVolumes.SetVolumes(options.AllocName, s.jobNamespace, tg.Volumes)
+	s.taskGroupHostVolumes.SetVolumes(options.AllocName, s.jobNamespace, tg.Volumes, options.AllocationHostVolumeIDs)
```
tgross, line 354: ~This raises a question: is "sticky" even meaningful for system and sysbatch jobs?~ It is because of CSI volumes, nevermind :grin:

scheduler/stack.go
```diff
@@ -35,10 +35,12 @@ type Stack interface {
...
 type SelectOptions struct {
-	PenaltyNodeIDs map[string]struct{}
-	PreferredNodes []*structs.Node
-	Preempt        bool
-	AllocName      string
+	PenaltyNodeIDs          map[string]struct{}
+	PreferredNodes          []*structs.Node
+	Preempt                 bool
+	AllocName               string
+	AllocID                 string
```
tgross, line 42: I don't think we're using the alloc ID anymore, are we?

scheduler/generic_sched_test.go
```diff
@@ -218,6 +218,124 @@ func TestServiceSched_JobRegister_StickyAllocs(t *testing.T) {
...
+	eval = &structs.Evaluation{
+		Namespace:   structs.DefaultNamespace,
+		ID:          uuid.Generate(),
+		Priority:    job.Priority,
+		TriggeredBy: structs.EvalTriggerNodeUpdate,
+		JobID:       job.ID,
+		Status:      structs.EvalStatusPending,
+	}
+	must.NoError(t, h.State.UpsertEvals(structs.MsgTypeTestSetup, h.NextIndex(), []*structs.Evaluation{eval}))
+	h1 := NewHarnessWithState(t, h.State)
```
tgross, line 319: Minor nitpick: we usually reuse the harness and then just expect that plans are appended, because it makes it easier to update the state correctly between `Process` calls if needed (because the harness tracks the index).

tgross (commented):
I think this is starting to come together @pkazmierczak. If you can, it'd be worth taking a moment to building it and doing some bench testing with real deployments, alloc failures, etc. to make sure we're not missing cases where the volumes should be getting handed off.

scheduler/stack.go
pkazmierczak: ah that's some leftovers from the refactor, i'll clean up

scheduler/generic_sched_test.go
pkazmierczak: yeah sorry, this was copy-pasta from some old test. I'll update.

scheduler/feasible.go
```diff
@@ -163,10 +172,10 @@ func (h *HostVolumeChecker) SetVolumes(allocName string, ns string, volumes map[
 			// provide a unique volume source per allocation
 			copied := req.Copy()
 			copied.Source = copied.Source + structs.AllocSuffix(allocName)
-			h.volumeReqs = append(h.volumeReqs, copied)
+			h.volumeReqs = append(h.volumeReqs, &allocVolumeRequest{volumeReq: copied})
 
 		} else {
-			h.volumeReqs = append(h.volumeReqs, req)
+			h.volumeReqs = append(h.volumeReqs, &allocVolumeRequest{hostVolumeIDs: allocHostVolumeIDs, volumeReq: req})
```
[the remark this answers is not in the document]
pkazmierczak: My reasoning was that for every call of `SetVolumes` we could potentially get a different allocation, and thus should store the alloc data per method call. But I see that `SetVolumes` only ever gets called by `Select`, so effectively it's per-task-group. 

I'll adjust.

tgross (approved):
LGTM!

PR #24563, opened 2024-11-28
make example-host-volume plugin run on macOS

pkazmierczak:
This adapts the shell script for darwin, making it easier to test.


demo/hostvolume/example-host-volume
```diff
@@ -48,6 +48,35 @@ done
 [ $# -lt 2 ] && { echo 'path required; seek --help' 1>&2; exit 1; }
 host_path="$2"
 
+# OS detect
+if [[ "$OSTYPE" == "linux-gnu"* ]]; then
```
tgross, line 52: This script should also work just fine for non-GNU Linux (and most non-macOS unixes AFAIK). For example, Alpine has a different OSTYPE:

```sh
$ docker run -it --rm alpine:latest /bin/sh
/ # apk add bash
fetch https://dl-cdn.alpinelinux.org/alpine/v3.18/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.18/community/x86_64/APKINDEX.tar.gz
(1/4) Installing ncurses-terminfo-base (6.4_p20230506-r0)
(2/4) Installing libncursesw (6.4_p20230506-r0)
(3/4) Installing readline (8.2.1-r1)
(4/4) Installing bash (5.2.15-r5)
Executing bash-5.2.15-r5.post-install
Executing busybox-1.36.1-r0.trigger
OK: 10 MiB in 19 packages
/ # bash
7c1b8bf42b27:/# echo $OSTYPE
linux-musl
7c1b8bf42b27:/#
```

tgross: I suspect but can't quickly confirm that all the `OSTYPE` values for bash are in the source repos for the various downstream Unixes, rather than in the bash source or manual. So trying to enumerate them all is probably a fool's errand. But let's fix this for musl at least because we know that (and support it for Enterprise customers). It'll be easy for the handful of FreeBSD or Illumos users out there to open a PR for adding support.

pkazmierczak: 632e080

tgross (approved):
LGTM

PR #24589, opened 2024-12-02
dynamic host volumes: fingerprint client plugins

gulducat:
This finds executable files in a new `client.dhv_plugin_dir` that respond to a Version call (arg $1 = `version` or `OPERATION=version` env) with a valid version string on stdout, and adds them to the client fingerprint. This fingerprint responds to a SIGHUP.

The main thing I'm wishy-washy on are user-facing names, which feel a bit franken-named here.

---

E.g.: given this client config

```hcl
client {
  enabled = true

  dhv_plugin_dir = "/opt/nomad/plugins-dhv" # default would be alongside alloc_mounts
...
```

and with these contents in that directory:

```shell
drwxrwxr-x 4.0K Nov 22 18:42 .
drwxrwxr-x 4.0K Nov 22 17:18 dir-and-not-a-plugin
-rwxrwxr-x 2.9K Nov 19 15:47 example-host-volume      # our proper example plugin
-rwxrwxr-x   64 Nov 22 18:42 executable-not-a-plugin
-rw-rw-r--    0 Nov 22 17:18 file-but-not-a-plugin
-rwxrwxr-x   69 Nov 22 18:16 not-a-plugin-and-errors
```

here are the client agent logs on start or sighup:

```
[INFO]  client/fingerprint_manager.go:122: client.fingerprint_mgr: reloading fingerprinter: fingerprinter=plugins_dhv
[DEBUG] fingerprint/dynamic_host_volumes.go:36: client.fingerprint_mgr.host_volume_plugins: detected plugin built-in: plugin_id=mkdir version=0.0.1
[DEBUG] hostvolumemanager/host_volume_plugin.go:132: client.fingerprint_mgr.host_volume_plugins: error with plugin: plugin_id=not-a-plugin-and-errors operation=version
  stdout=
  | no, stdout

  stderr=
  | no, stderr
   error="exit status 1"
[DEBUG] fingerprint/dynamic_host_volumes.go:109: client.fingerprint_mgr.host_volume_plugins: failed to get version from plugin: plugin_id=not-a-plugin-and-errors error="error getting version from plugin \"not-a-plugin-and-errors\": exit status 1"
[DEBUG] fingerprint/dynamic_host_volumes.go:109: client.fingerprint_mgr.host_volume_plugins: failed to get version from plugin: plugin_id=executable-not-a-plugin error="error with version from plugin: Malformed version: nah, stdout"
[DEBUG] fingerprint/dynamic_host_volumes.go:66: client.fingerprint_mgr.host_volume_plugins: detected plugin: plugin_id=example-host-volume version=0.0.1
[DEBUG] client/client.go:2548: client: state changed, updating node and re-registering
[INFO]  client/client.go:2081: client: node registration complete
```

and this is the resulting fingerprint:

```shell
$ nomad node status -verbose -self | grep dhv
plugins.dhv.version.example-host-volume                  = 0.0.1
plugins.dhv.version.mkdir                                = 0.0.1  # "mkdir" plugin is built-in
```


client/fingerprint/dynamic_host_volumes.go
```diff
@@ -0,0 +1,119 @@
...
+	for k := range request.Node.Attributes {
+		if strings.HasPrefix(k, "plugins.dhv.") {
+			response.RemoveAttribute(k)
+		}
+	}
+
+	// set the attribute(s)
+	for plugin, version := range plugins {
+		h.logger.Debug("detected plugin", "plugin_id", plugin, "version", version)
+		response.AddAttribute("plugins.dhv.version."+plugin, version)
```
tgross, line 66: I could maybe be convinced otherwise on the plugin configuration field I noted below, but here if we have the `dhv` abbreviation it loses all context by the time it gets to be user-visible. We end up with a mysterious acronym in the middle of a very long list of attributes.

```suggestion
		response.AddAttribute("plugins.host_volume.version."+plugin, version)
```

client/fingerprint/dynamic_host_volumes.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	return nil
+}
+
+func (h *DynamicHostVolumePluginFingerprint) Periodic() (bool, time.Duration) {
+	return false, 0 // not periodic;  db TODO(1.10.0): could be?
```
tgross, line 73: Once https://github.com/hashicorp/nomad/pull/24526 lands I think that's the last periodic fingerprint. Experience has demonstrated periodic fingerprints just end up having little value except to occasionally cause giant flapping storms of fingerprint updates. :grin:

client/fingerprint/dynamic_host_volumes.go
```diff
@@ -0,0 +1,119 @@
...
+
+	plugins := make(map[string]string)
+	mut := sync.Mutex{}
+	var wg sync.WaitGroup
+
+	for file, fullPath := range files {
+		wg.Add(1)
+		go func(file, fullPath string) {
+			defer wg.Done()
+			// really should take way less than a second
```
tgross, line 93: Someone is totally going to write a plugin on the JVM that takes 10s to assemble a DI framework on startup, and we'll :facepalm: about it. But this seems like a reasonable assumption. :grin:

client/fingerprint/dynamic_host_volumes.go
```diff
@@ -0,0 +1,119 @@
...
+			ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+			defer cancel()
+
+			log := log.With("plugin_id", file) // db TODO: "name" like CNI, instead of "ID" ?
```
tgross, line 97: What do you mean by this TODO? I think we probably want to force the plugin ID to be the file name, right?

client/fingerprint/dynamic_host_volumes.go
```diff
@@ -0,0 +1,119 @@
...
+
+			p, err := hvm.NewHostVolumePluginExternal(log, file, fullPath, "")
+			if err != nil {
+				log.Debug("error getting plugin", "error", err)
```
tgross, line 101: We might want to be noisier than debug (probably warn), because something has gone fairly wrong in their configuration to hit any of the error cases we could get  from this method. You'd just end up with a missing attribute and no explanation except in the debug logs.

client/fingerprint/dynamic_host_volumes_test.go
```diff
@@ -0,0 +1,89 @@
...
+	// do it again after deleting our one good plugin.
+	// repeat runs should wipe attributes, so nothing should remain.
+	node.Attributes = resp.Attributes
+	must.NoError(t, os.Remove(filepath.Join(tmp, "happy-plugin")))
+
+	resp = FingerprintResponse{}
+	err = fp.Fingerprint(req, &resp)
+	must.NoError(t, err)
+	must.Eq(t, map[string]string{
+		"plugins.dhv.version.happy-plugin": "", // will get cleaned up later
```
tgross, line 85: What's this comment referring to?

tgross (approved):
I've left some bikesheddy comments about the naming of user-facing attributes and some log levels, but otherwise this looks great!

client/fingerprint/dynamic_host_volumes.go
gulducat: yaknow, I wonder if plugin authors might even use the "responds to `version`" aspect as a way to auto-disable plugins that won't be able to fulfill requests, like if they can't reach an NFS server, or who knows.  they might want to try until some timeout, and that might reasonably be longer than 1s...

client/fingerprint/dynamic_host_volumes.go
gulducat: I was considering maybe doing this all over the place

```suggestion
			log := log.With("name", file)
```

similar to CNI networks and plugins, e.g.

```
[DEBUG] fingerprint/cni.go:76: client.fingerprint_mgr: detected CNI network: name=cool
[DEBUG] fingerprint/plugins_cni.go:119: client.fingerprint_mgr.cni_plugins: failed to parse CNI plugin version: name=cool
```

client/fingerprint/dynamic_host_volumes.go
gulducat: :+1: this may be left over from an earlier iteration that would have had duplicate warn/error logs...

client/fingerprint/dynamic_host_volumes_test.go
gulducat: it refers to this docstring

https://github.com/hashicorp/nomad/blob/v1.9.3/client/fingerprint/structs.go#L52-L54
```go
// RemoveAttribute sets the given attribute to empty, which will later remove
// it entirely from the node
func (f *FingerprintResponse) RemoveAttribute(name string) {
```

just to clarify that the map key still existing is expected. my intuition was that the attribute should be _removed_, not just set to empty.

client/fingerprint/dynamic_host_volumes.go
tgross: Up to you. `plugin_id` matches the field in the volume spec though.

client/fingerprint/dynamic_host_volumes_test.go
tgross: Oh, huh. Yeah mine too. :grin:

client/fingerprint/dynamic_host_volumes.go
gulducat: yeah, I had basically talked myself out of it by writing my comment, and you helped tip it over the edge :yum:  I'll keep plugin_id.

command/agent/config.go
```diff
@@ -229,6 +229,10 @@ type ClientConfig struct {
 	// AllocMountsDir is the directory for storing mounts into allocation data
 	AllocMountsDir string `hcl:"alloc_mounts_dir"`
 
+	// DHVPluginDir is the directory containing dynamic host volume plugins
+	// db TODO(1.10.0): document default directory is alongside alloc_mounts
+	DHVPluginDir string `hcl:"dhv_plugin_dir"` // db TODO(1.10.0): is this a good name?
```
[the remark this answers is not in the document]
gulducat: more bike-shedding;

1. where should it go by default? I put it next to `alloc_mounts` because the intention is for the plugins to put stuff in `alloc_mounts`, but it could go next to default `plugin_dir` within `data_dir` instead?
1. I put `plugins-` as a prefix, so it would sort next to other "plugin" dirs, but none are present by default where I've put this, and `host-volume-plugins` would better match the config value...
1. default `-`-joined `plugins-host-volume` or `_` `plugins_host_volume`?
1. changing CLI flag to `-host-volume-plugin-dir` seems good, eh?

so many little fiddly naming bits :sweat_smile:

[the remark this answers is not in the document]
gulducat: nice, I'll go with `<data-dir>/host_volume_plugins` then. it won't be sorted next to `plugins`, but I like the consistent word order in code and UX more than `ls` output of the data dir.  might even be good to type `/opt/nomad/data/hos<tab><tab>` so autocomplete won't fight with `plugins*`

PR #24595, opened 2024-12-02
dynamic host volumes: client state

gulducat:
Store dynamic host volume creations in client state, so they can be "restored" on agent restart. Restore works by repeating the same Create operation as initial creation, and expecting the plugin to be idempotent.

This is (potentially) especially important after _host_ restarts, which may have dropped mount points or such.

One particular thing I'm not sure of is when to error fatally if there are state operation errors... I don't have good intuition about the failure states there, but it does seem Bad for state to drift at all?

There's a decent chunk of code here, but a lot of it is just satisfying interfaces.


client/client.go
```diff
@@ -535,9 +535,13 @@ func NewClient(cfg *config.Config, consulCatalog consul.CatalogAPI, consulProxie
 	c.devicemanager = devManager
 	c.pluginManagers.RegisterAndRun(devManager)
 
-	c.hostVolumeManager = hvm.NewHostVolumeManager(logger,
+	c.hostVolumeManager, err = hvm.NewHostVolumeManager(logger,
+		c.stateDB, hostVolumeRequestTimeout,
 		cfg.HostVolumePluginDir,
 		cfg.AllocMountsDir)
+	if err != nil {
+		return nil, err // db TODO(1.10.0): don't fail the whole client if state restore fails?
```
tgross, line 543: It sort of depends on the error. When we restore allocations it's not fatal to fail restoring a single allocation, because we assume that the allocation may have been destroyed while we were down (ex. the host rebooted), and there's a process to recover (check back in with the server, restart the alloc). But if we can't read the client state _at all_, then we're in a fatal scenario and want to stop the client from starting and making things worse / undebuggable.

client/hostvolumemanager/host_volumes.go
```diff
@@ -20,19 +22,59 @@ var (
...
+	vols, err := state.GetDynamicHostVolumes()
+	if err != nil {
+		return err
+	}
+	if len(vols) == 0 {
+		return nil // nothing to do
 	}
+
+	// re-"create" the volumes - plugins have the best knowledge of their
+	// side effects, and they should be idempotent.
```
tgross, line 61: I think this is going to have to be a requirement, so let's use language here that reinforces that:

```suggestion
	// side effects, and they must be idempotent.
```

client/hostvolumemanager/host_volumes.go
```diff
@@ -20,19 +22,59 @@ var (
...
+	var wg sync.WaitGroup
+	for _, vol := range vols {
+		wg.Add(1)
+		func() {
```
tgross, line 65: It looks like you meant for this to be run in a goroutine. This makes me realize another requirement we need to remember to document: plugins need to be safe to run concurrently.

client/hostvolumemanager/host_volumes.go
```diff
@@ -20,19 +22,59 @@ var (
...
+			defer wg.Done()
+			ctx, cancel := context.WithTimeout(context.Background(), timeout)
+			defer cancel()
+			// note: this will rewrite client state that we just restored
+			if _, err := hvm.Create(ctx, vol.CreateReq); err != nil {
+				hvm.log.Error("failed to restore", "volume_id", vol.ID, "error", err)
+				// db TODO: multierror w/ mutex?
```
tgross, line 72: If you're going to collect errors, you might want to send them back out on a channel and collect them that way. That would also let you get rid of the wait group.

client/hostvolumemanager/host_volumes.go
```diff
@@ -63,14 +105,22 @@ func (hvm *HostVolumeManager) Create(ctx context.Context,
 		return nil, err
 	}
 
+	volState := &HostVolumeState{
+		ID:        req.ID,
+		CreateReq: req,
+	}
+	if err := hvm.stateMgr.PutDynamicHostVolume(volState); err != nil {
+		hvm.log.Error("failed to save volume in state", "volume_id", req.ID, "error", err)
+		// db TODO: bail or nah?
```
tgross, line 114: If we create the volume on-disk but not in client state / server state, we'll end up with stray volumes we have no way of cleaning up. I'd try to implement a transaction-like semantic here -- if we can't fully complete the task, we should destroy the volume when we're done.

client/hostvolumemanager/host_volumes.go
```diff
@@ -89,7 +139,10 @@ func (hvm *HostVolumeManager) Delete(ctx context.Context,
 
 	resp := &cstructs.ClientHostVolumeDeleteResponse{}
 
-	// db TODO(1.10.0): save the client state!
+	if err := hvm.stateMgr.DeleteDynamicHostVolume(req.ID); err != nil {
+		hvm.log.Error("failed to delete volume in state", "volume_id", req.ID, "error", err)
+		// db TODO: bail or nah?
```
tgross, line 144: If we fail to delete the client state but don't return an error, then we'll try to restore the volume when the client restarts, leaving a stray volume that shouldn't exist but also doesn't exist in the server and therefore can't be deleted. So we should probably return an error here so the user can retry.

client/state/db_bolt.go
```diff
@@ -138,6 +139,8 @@ var (
 
 	// nodeRegistrationKey is the key at which node registration data is stored.
 	nodeRegistrationKey = []byte("node_registration")
+
+	hostVolBucket = []byte("volumes_to_create")
```
tgross, line 143: Just in case we decide to add CSI volumes here someday for some ungodly reason :grin: 

```suggestion
	hostVolBucket = []byte("host_volumes_to_create")
```

client/hostvolumemanager/host_volumes_test.go
```diff
@@ -0,0 +1,114 @@
...
+	})
+}
+
+var _ HostVolumeStateManager = &mockState{}
+
+func newMockState() *mockState {
+	return &mockState{
+		vols: make(map[string]*HostVolumeState),
+	}
+}
```
tgross, line 74: The `client/state/db_mem.go` and `client/state/db_error.go` files are intended to be mock client states you can use for testing these scenarios.

client/client.go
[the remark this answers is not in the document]
pkazmierczak: ah I see Tim already commented https://github.com/hashicorp/nomad/pull/24595#discussion_r1867812893
don't mind me

client/hostvolumemanager/host_volumes_test.go
gulducat: I got an import cycle trying to use them, because they import `hostvolumemanager` for its `HostVolumeState`.  would it make sense to move `HostVolumeState` into `client/structs`? I can avoid the import cycle that way.

client/client.go
gulducat: error handling nailed down in 8043871641cb694d9c990d6a6a129508641591d3

client/hostvolumemanager/host_volumes.go
gulducat: :+1: added a comment as part of error handling in 8043871641cb694d9c990d6a6a129508641591d3

client/hostvolumemanager/host_volumes.go
gulducat: I went with a [`multierror.Group{}`](https://pkg.go.dev/github.com/hashicorp/go-multierror#Group)

client/client.go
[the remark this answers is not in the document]
gulducat: I ended up with this https://github.com/hashicorp/nomad/commit/8043871641cb694d9c990d6a6a129508641591d3#diff-bd3a55a72186f59e2e63efb4951573b2f9e4a7cc98086e922b0859f8ccc1dd09R543-R547

let me know if you have other ideas!

tgross (approved):
LGTM!

PR #24600, opened 2024-12-03
helper: sanitize method on ACL token object

tgross:
There are several places where we want to redact the secret ID of an ACL token, some of which are in the Enterprise code base for Sentinel. Add a new method `Sanitize` that mirrors the one we have on `Node`.

This changeset doesn't change any existing behavior, just makes it easier to reuse.

Ref: https://github.com/hashicorp/nomad-enterprise/pull/2087

### DIFF
--- a/nomad/structs/acl.go
+++ b/nomad/structs/acl.go
@@ -474,6 +474,16 @@ func (a *ACLToken) UnmarshalJSON(data []byte) (err error) {
 	return nil
 }
 
+func (a *ACLToken) Sanitize() *ACLToken {
+	if a == nil {
+		return nil
+	}
+
+	out := a.Copy()
+	out.SecretID = ""
+	return out
+}
+
 // ACLRole is an abstraction for the ACL system which allows the grouping of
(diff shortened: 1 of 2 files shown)


nomad/structs/acl.go
```diff
@@ -474,6 +474,16 @@ func (a *ACLToken) UnmarshalJSON(data []byte) (err error) {
 	return nil
 }
 
+func (a *ACLToken) Sanitize() *ACLToken {
+	if a == nil {
+		return nil
+	}
+
+	out := a.Copy()
+	out.SecretID = ""
```
schmichael, line 483: Why not `"<redacted>"`? I think Node only does `""` because there's no reason for humans to even need to know that field exists. It's an internal implementation detail.

I'm fine with ACLToken going either way, and I suppose there's something to be said for consistency.

schmichael (approved):

nomad/structs/acl.go
tgross: I didn't want to change the format of the objects we're emitting in the event stream (at least not without thinking about that).

PR #24601, opened 2024-12-03
sentinel: remove default scope for Sentinel apply command

tgross:
When we add a Sentinel scope for dynamic host volumes, having a default `-scope` value for `sentinel apply` risks accidentally adding policies for volumes to the job scope. This would immediately prevent any job from being submitted. Forcing the administrator to pass a `-scope` will prevent accidental misuse.

Ref: https://github.com/hashicorp/nomad-enterprise/pull/2087

### DIFF
--- a/.changelog/24601.txt
+++ b/.changelog/24601.txt
@@ -1,0 +1,3 @@
+```release-note:breaking-change
+sentinel: The sentinel apply command now requires the -scope option
+```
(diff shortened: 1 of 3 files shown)


schmichael (approved):
Good call! Don't forget the changelog.

A backwards compatibility break, but I don't think we need to worry about it since the failure case is safe (no operation will occur) and will loudly direct users toward the resolution.

PR #24555, opened 2024-11-26
[ui, ci] retain artifacts from test runs including test timing

philrenaud:
### Description
Our `ember-test-audit` action is currently failing, and I'm using that as an excuse to improve our front-end testing story generally.

This will be a two-step approach, and this PR is the first step:
1. Save artifacts, including test timing, from `test-ui` test results on merge to `main`
2. Compare to a rolling window of these test results on subsequent PRs

This PR adds artifact uploads on the back of `test-ui`. Because those are partitioned/run concurrently 4x, part of this is combining those.

The artifact generated looks like this:

```
{
  "timestamp": "2024-12-02T18:20:59.531Z",
  "sha": "7fd10531ef3e12672a19a7a9aa7a6e4f4f874614",
  "summary": {
    "total": 1606,
    "passed": 1606,
    "failed": 0
  },
  "duration": 841403,
  "tests": [
    {
      "name": "Acceptance | allocation fs: it passes an accessibility audit",
      "partition": 1,
      "browser": "Chrome 131.0",
      "passed": true,
      "duration": 1331,
      "error": null,
      "logs": []
    },
    {
      "name": "Acceptance | allocation fs: visiting filesystem root",
      "partition": 1,
      "browser": "Chrome 131.0",
      "passed": true,
      "duration": 505,
      "error": null,
      "logs": []
    },
    {
      "name": "Acceptance | allocation fs: visiting filesystem paths",
      "partition": 1,
      "browser": "Chrome 131.0",
      "passed": true,
      "duration": 542,
      "error": null,
      "logs": []
    },
...
```

In phase 2 of this, we'll be able to do test-by-test duration averages and comparisons, and tease out those tests with the greatest deltas, find flakes, etc.

PR #24604, opened 2024-12-04
[ui, ci] UI test look-back and comparison for PRs

philrenaud:
### Description
Phase 2 of the project to fix and improve our ember test audits. [Phase 1 can be found here](https://github.com/hashicorp/nomad/pull/24555).


philrenaud:
An attempt at "Which of our tests are flakey over time" which is now far enough out of date that it's worth closing.

PR #24605, opened 2024-12-04
dynamic host volumes: add implicit constraints on plugin fingerprint

tgross:
Node fingerprints include attributes for the host volume plugins, including the built-in plugins. Add an implicit constraint on this fingerprint during volume placement to ensure we only place volumes on hosts with the right plugins.


gulducat (approved):
lgtm!

PR #24607, opened 2024-12-04
ci: only cache from github hosted runners to avoid file perm issues

mismithhisler:
### Description

Pushing a cache bundle from a self-hosted runner may be causing file permission issues when unpacking the bundle in later tests. This attempts to fix that issue by only pushing a bundle from Github hosted runners.

### Contributor Checklist

### Reviewer Checklist
  backporting document.
  history should be preserved.
  within the public repository.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -43,7 +43,7 @@ jobs:
         run: git config --global url.'https://${{ env.ELEVATED_GITHUB_TOKEN }}@github.com'.insteadOf 'https://github.com'
       - uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
         with:
-          cache: true
+          cache: ${{ contains(runner.name, 'Github Actions') }}
           go-version-file: .go-version
           cache-dependency-path: '**/go.sum'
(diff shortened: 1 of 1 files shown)


mismithhisler:
An alternative way to do this is to not use the self-hosted runner in the `checks` job.  The job would take a little longer, but then later unit tests should always have a cache hit.

tgross (approved):
LGTM as an experiment... I don't suppose there's any way to test that this works to solve the problem until it's been merged and then we run CI on ENT? If not, let's just go ahead and give it a try.

mismithhisler:
> LGTM as an experiment... I don't suppose there's any way to test that this works to solve the problem until it's been merged and then we run CI on ENT? If not, let's just go ahead and give it a try.

It's possible it was a coincidence, but I did some testing this morning where a cache produced by the `test-e2e-consul` job (Github hosted runner) could successfully be extracted, and another run where the cache produced by the `check` job (self-hosted) resulted in errors extracting.

shoenig:
I dug into this a bit for one of my own projects; the expectation is that you do not using caching on self hosted runners (ephemeral or static) unless you're paying for github enterprise, the agent for which sets up the sandbox correctly.

PR #24612, opened 2024-12-05
dynamic host volumes: refactor HTTP routes for volumes list dispatch

tgross:
The List Volumes API was originally written for CSI but assumed we'd have future volume types, dispatched on a query parameter. Dynamic host volumes uses this, but the resulting code has host volumes concerns comingled in the CSI volumes endpoint. Refactor this so that we have a top-level `GET /v1/volumes` route that's shared between CSI and DHV, and have it dispatch to the appropriate handler in the type-specific endpoints.

---

I did extra testing of this by running the hostpath CSI plugin demo and creating a bunch of host volumes, and verifying the CLI worked as expected for both.


pkazmierczak (approved):
👍

PR #24613, opened 2024-12-05
dynamic host volumes: volume fingerprinting

gulducat:
Draft because it needs some tidying, and probably more tests.

We did *plugin* fingerprinting in #24589, not to be confused with *volume* fingerprinting, which is done here.  They are very different both conceptually and by implementation.  This *volume* fingerprinting behaves more like device or driver plugins, even though there's not really any long-running process to maintain.  This is in keeping with some of the CSI structure.

With this, the whole flow works end-to-end. The `volume create` monitor can run to completion, because the volume gets set to `ready`.  Node API returns them, and jobs can use them like any other host volume.  It also re-fingerprints properly on agent restart.

```
$ nomad volume create host.volume.hcl
==> Created host volume test with ID 1e1d4540-14f9-2515-8502-24d46291717d
  ✓ Host volume "1e1d4540" ready 💚

    2024-12-05T17:06:46-05:00
    ID        = 1e1d4540-14f9-2515-8502-24d46291717d
    Name      = test
    Namespace = default
    Plugin ID = example-host-volume
    Node ID   = ae118273-5416-3ed7-01dd-10fe8bf49e81
    Node Pool = default
    Capacity  = 47 MiB
    State     = ready  💚
    Host Path = /opt/nomad/data/alloc_mounts/1e1d4540-14f9-2515-8502-24d46291717d
```

(hearts manually added to draw the eye in the description here)

The volume ID is also attached to the `ClientHostVolumeConfig`, so:

```
$ nomad node status -self -json | jq '.HostVolumes'
{
  "cloud-img": {
    "ID": "",      <== "classic" static host volume
    "Path": "/opt/nomad/cloud-img",
    "ReadOnly": false
  },
  "test": {
    "ID": "1e1d4540-14f9-2515-8502-24d46291717d",   <== dynamic host volume
    "Path": "/opt/nomad/data/alloc_mounts/1e1d4540-14f9-2515-8502-24d46291717d",
    "ReadOnly": false
  }
}

```

ref: #24479


nomad/host_volume_endpoint.go
```diff
@@ -610,6 +610,7 @@ func (v *HostVolume) deleteVolume(vol *structs.HostVolume) error {
 	method := "ClientHostVolume.Delete"
 	cReq := &cstructs.ClientHostVolumeDeleteRequest{
 		ID:         vol.ID,
+		Name:       vol.Name,
```
tgross, line 613: Maybe I'm missing it, but I don't see what's changed in the client to necessitate adding this field here?

api/nodes.go
```diff
@@ -515,6 +515,8 @@ type DriverInfo struct {
 
 // HostVolumeInfo is used to return metadata about a given HostVolume.
 type HostVolumeInfo struct {
+	//ID       string
+	//Name     string
```
tgross, line 519: I think this is what gets returned by the [Read Node API](https://developer.hashicorp.com/nomad/api-docs/nodes#read-node), but it's a map by name. Adding the ID makes sense but the name would be redundant.

client/node_updater.go
```diff
@@ -123,6 +132,25 @@ func (c *Client) updateNodeFromCSI(name string, info *structs.CSIInfo) {
...
 }
 
+func (c *Client) updateNodeFromHostVol(name string, vol *structs.ClientHostVolumeConfig) {
+	c.configLock.Lock()
+	defer c.configLock.Unlock()
+
+	newConfig := c.config.Copy()
+
+	if newConfig.Node.HostVolumes == nil {
+		newConfig.Node.HostVolumes = make(map[string]*structs.ClientHostVolumeConfig) // TODO: why?
```
tgross, line 142: Generally speaking the various maps in configs are `nil` and not empty if unset (I suppose so that we're not allocating the [internal structs for the map](https://dave.cheney.net/2018/05/29/how-the-go-runtime-implements-maps-efficiently-without-generics), although admittedly that doesn't seem like a lot of memory overall). If there's no static host volume, the first time we access this it would be an assignment-to-nil-map panic. So we need to initialize the map on first use.

client/hostvolumemanager/host_volumes_test.go
```diff
@@ -21,10 +23,12 @@ func TestNewHostVolumeManager_restoreState(t *testing.T) {
 	vol := &cstructs.HostVolumeState{
 		ID: "test-vol-id",
 		CreateReq: &cstructs.ClientHostVolumeCreateRequest{
+			Name:     "test-vol-name",
 			ID:       "test-vol-id",
 			PluginID: "mkdir",
 		},
 	}
+	fNode := newFakeNode()
```
tgross, line 31: If it's at all feasible, I'd considering using a real client in here rather than a mock. There's lock logic that'd be nice to make sure is getting exercised.

client/hostvolumemanager/host_volumes.go
```diff
@@ -7,20 +7,40 @@ import (
...
 	ErrPluginNotExists     = errors.New("no such plugin")
 	ErrPluginNotExecutable = errors.New("plugin not executable")
+	ErrUpdateTimeout       = errors.New("timeout updating client node")
 )
 
-type HostVolumeStateManager interface {
+type VolumeMap map[string]*structs.ClientHostVolumeConfig
+
+// UpdateVolumeMap returns true if it changes the volume map.
+func UpdateVolumeMap(volumes VolumeMap, name string, vol *structs.ClientHostVolumeConfig) (changed bool) {
```
tgross, line 29: It doesn't look like we ever use the return value of this function.

Also, we're already holding the lock here or are working on a copy, right? We should probably document this in the docstring.

nomad/host_volume_endpoint.go
gulducat: this ends up down in `HostVolumeManager.Delete` so it can be passed into `updateNodeVols` here: https://github.com/hashicorp/nomad/pull/24613/files#diff-0ca03dbe7624ffebd44da8e738ac49a768b68a96c4f612f24f4f0bb606b81c65R214

in fact, I was struggling to understand why the addition to the endpoint test was failing, until I realized I needed this:
https://github.com/hashicorp/nomad/pull/24613/files#diff-054b6367d0da40a926f9565f6129bf0fabc62f7c2bb154540a8c69ac5f10b5cdR74
because otherwise `""` was being "deleted" from the map, leaving the actual volume in place.

client/hostvolumemanager/host_volumes_test.go
gulducat: I figure the actual client usage is being exercised in the client RPC endpoint test:
https://github.com/hashicorp/nomad/pull/24613/files#diff-054b6367d0da40a926f9565f6129bf0fabc62f7c2bb154540a8c69ac5f10b5cdR30-R31

I don't think this thing down here needs to know what a client is.  also, I think I'd hit an import cycle if I import `client`, since `client` imports this.

client/hostvolumemanager/host_volumes.go
gulducat: we're working on a copy or a lock'd map. I figured the docstring saying it might change the volume map implied that, but I'll make it more explicit!

here are its usages:
* copy of config, uses `changed` here: https://github.com/hashicorp/nomad/pull/24613/files#diff-cf4c8c728ec5eb37deb3d780d75c3ca5dff1edf4b02fc46551fce35482510598R53
* ditto here: https://github.com/hashicorp/nomad/pull/24613/files#diff-cf4c8c728ec5eb37deb3d780d75c3ca5dff1edf4b02fc46551fce35482510598R145
* and this is a locked map, which doesn't care about `changed` because it's the first run on client start: https://github.com/hashicorp/nomad/pull/24613/files#diff-cf4c8c728ec5eb37deb3d780d75c3ca5dff1edf4b02fc46551fce35482510598R401

part of my goal for having this `UpdateVolumeMap` func was so that this list is a little easier to track down. the other plugin managers all have small variations of the same logic strewn about, so it was really hard to understand how it all connected! with this, I can set a breakpoint in this one func, and easily catch when this operation takes place.

client/hostvolumemanager/host_volumes_test.go
tgross: Fair enough.

client/hostvolumemanager/host_volumes.go
tgross: Ah, that caller in `batchFirstFingerprints` could probably be:

```
hostVolChanged = hvm.UpdateVolumeMap(newConfig.Node.HostVolumes, name, vol)
```

gulducat: true! I like little tidyings.

client/hostvolumemanager/host_volumes.go
```diff
@@ -31,37 +51,54 @@ type HostVolumeManager struct {
...
-	if err := hvm.restoreState(state, restoreTimeout); err != nil {
-		return nil, err
+func genVolConfig(req *cstructs.ClientHostVolumeCreateRequest, resp *HostVolumePluginCreateResponse) *structs.ClientHostVolumeConfig {
+	if req == nil || resp == nil {
+		return nil
+	}
+	return &structs.ClientHostVolumeConfig{
+		Name:     req.Name,
+		Path:     resp.Path,
+		ReadOnly: false, // db TODO: add to request? (??)
```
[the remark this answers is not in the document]
gulducat: I like this, because I don't have to do anything except perhaps add a comment repeating your explanation :yum:

client/hostvolumemanager/volume_fingerprint.go
```diff
@@ -0,0 +1,65 @@
...
+		hvm.log.Error("failed to restore state", "error", err)
+		return ctx.Done()
+	}
+	for name, vol := range volumes {
+		hvm.updateNodeVols(name, vol) // => batchNodeUpdates.updateNodeFromHostVolume()
+	}
+	return ctx.Done()
+}
+func (hvm *HostVolumeManager) Run()      {}
+func (hvm *HostVolumeManager) Shutdown() {}
```
tgross, line 61: The various RPC handlers pass a request context into the calls to the plugins (which is just a timeout on the background context), but we don't have a way of signalling to plugin calls that the client is shutting down to cancel them. There's two options:
* Stick an agent shutdown context on the `HostVolumeManager` that's closed here in `Shutdown`. And use a [joincontext](https://github.com/LK4D4/joincontext) as we do elsewhere in the client to handle this problem.
* Derive the request context in the RPC handler from an agent shutdown context. This is a much cleaner option... but I suspect the reason we don't have that elsewhere is maybe because the agent doesn't _have_ its own shutdown context? (Because the code base predates contexts.)

I don't think this is blocking to fix in this PR (once the agent process exits, those plugin calls are certainly going to stop!), but let's follow-up on that later.

client/hostvolumemanager/host_volumes.go
```diff
@@ -26,42 +27,68 @@ type HostVolumeStateManager interface {
...
+		return nil
+	}
+	return &structs.ClientHostVolumeConfig{
+		Name: req.Name,
+		ID:   req.ID,
+		Path: resp.Path,
+
+		// dynamic volumes, like CSI, have more robust `capabilities`,
+		// so we always set ReadOnly to false, and let the scheduler
+		// decide when to ignore this and check capabilities instead.
```
tgross, line 76: I think the taskrunner's `volume_hook` will need to handle this too, but I'll follow-up with that later.

tgross (approved):
LGTM!

PR #24616, opened 2024-12-06
[ui] Resolutions for code-scanner/dependabot alerts in UI

philrenaud:
Pins cross-spawn to a healthy version. Yarn will complain that it's incongruent, but the major version upgrade appears to cause no issues and reading through their upgrade notes, it's only to disable support for a very old version of node (8)

Resolves https://github.com/hashicorp/nomad/security/code-scanning/758
Resolves https://github.com/hashicorp/nomad/security/code-scanning/759
Resolves https://github.com/hashicorp/nomad/security/code-scanning/760
Resolves https://github.com/hashicorp/nomad/security/code-scanning/761

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -183,6 +183,7 @@
     "ember-auto-import": "^2.4.0",
     "prop-types": "^15.8.1",
     "**/express/**/path-to-regexp": "0.1.10",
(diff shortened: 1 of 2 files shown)


tgross (approved):
LGTM :+1:

PR #24617, opened 2024-12-06
dynamic host volumes: capabilities check during scheduling

tgross:
Static host volumes have a simple readonly toggle, but dynamic host volumes have a more complex set of capabilities similar to CSI volumes. Update the feasibility checker to account for these capabilities and volume readiness.

Also fixes a minor bug in the state store where a soft-delete (not yet implemented) could cause a volume to be marked ready again. This is needed to support testing the readiness checking in the scheduler.


pkazmierczak (approved):
LGTM, thanks for doing this Tim

PR #24641, opened 2024-12-10
stateful deployments: add Sticky property to task group volumes

pkazmierczak:
This changeset adds `Sticky` property to task group Volumes block, which will later be used by stateful deployments.


tgross (approved):
LGTM :+1:

PR #24643, opened 2024-12-10
tests: swap require for test in util_test.go

shoenig:


tgross:
`testify` has been completely removed from the `scheduler` package as of https://github.com/hashicorp/nomad/pull/25968, so safe to close this out. Thanks!

Issue #24635, opened 2024-12-09
`nomad logs` no longer works without `NOMAD_REGION` set


HerbCSO:
### Nomad version
Nomad v1.9.3
BuildDate 2024-11-11T16:35:41Z
Revision d92bf1014886c0ff9f882f4a2691d5ae8ad8131c

### Operating system and Environment details
MacOS CLI, Linux server

### Issue
Up to and including Nomad CLI 1.7.7 I was able to run `nomad alloc logs deadbeef` without `NOMAD_REGION` being set. Since 1.8.0 `NOMAD_REGION` is required, else I just get a `Unexpected response code: 500 (No path to region)`. So I have to either set `NOMAD_REGION` or run `nomad alloc logs -region myregion deadbeef`.

[From the docs](https://developer.hashicorp.com/nomad/docs/commands/alloc/logs#region) it reads to me like this is unintentional behavior. It says `Defaults to the Agent's local region.` so I would assume I wouldn't need to specify this parameter. I didn't find anything in the release notes that specifies a necessary config change for this, so please let me know if I missed something there.

Also, this sounds a bit like the issue described in https://github.com/hashicorp/nomad-pack/issues/374, but I'm not sure if this is really the same thing.

FWIW our Nomad servers and clients are a bit older, running 1.6.3 - maybe that's part of the issue?

### Reproduction steps
Run `nomad alloc logs` using a version prior to 1.8.0 - works
Run `nomad alloc logs` using a version prior >= 1.8.0 - produces 500 error

#### Expected Result
No 500 error

#### Actual Result
500 error

### Job file (if appropriate)
N/A

### Nomad Server logs (if appropriate)
N/A

### Nomad Client logs (if appropriate)
N/A

tgross:
Hi @HerbCSO! It looks like you're running into the same issue described here: https://github.com/hashicorp/nomad/issues/24609#issuecomment-2524048950. The `alloc logs` RPC uses the exact same paths as the `alloc exec` command described there. I'd ask if you can answer some of the same questions I've asked there.

> FWIW our Nomad servers and clients are a bit older, running 1.6.3 - maybe that's part of the issue?

This is an interesting clue! The CLI is 1.8.0+ and the agents are older?

HerbCSO:
> I'd ask if you can answer some of the same questions I've asked there.

[...]

Issue #24609, opened 2024-12-05
Nomad 1.8.4 `failed to exec into task: No path to region`


mcphailtom:
### Nomad version
Nomad v1.8.4
BuildDate 2024-09-17T20:18:34Z
Revision 22ab32e6cff66cf52f5e3f115b93de776bc09583

### Operating system and Environment details

Linux 20.04.1-Ubuntu

### Issue

As part of regular releases we are upgrading our shipped versions of Nomad. In our latest release candidates we have attempted a two step move from 1.15.17 -> 1.7.7 -> 1.8.4.

During testing we have performed this upgrade on a sample set of clusters and in all cases the upgrade is completely successfully. We have also attempted the 1.5.17 -> 1.8.4 upgrade without issue. Jobs are running as expected.

Several tests in our test suite use the `nomad alloc exec` command for test execution and we are seeing intermittent failures of the tests with the error message: `failed to exec into task: No path to region`

After some initial investigation it appears the command fails on all nodes of the cluster except the node where the allocation is actually running.

This behaviors is not consistent with previous versions of nomad.

### Reproduction steps

Using a sample job:

```
admin@app1:~$ nomad job status mysql
ID            = mysql
Name          = mysql
Submit Date   = 2024-12-05T01:13:35Z
Type          = system
Priority      = 50
Datacenters   = primary
Namespace     = default
Node Pool     = default
Status        = running
Periodic      = false
Parameterized = false

Summary
Task Group  Queued  Starting  Running  Failed  Complete  Lost  Unknown
mysql       0       0         3        0       0         0     0

Allocations
ID        Node ID   Task Group  Version  Desired  Status   Created    Modified
37e41f01  520855e6  mysql       0        run      running  10h7m ago  10h7m ago
69195c2c  8c55b7e6  mysql       0        run      running  10h7m ago  23m38s ago
8cf697a1  9da121d6  mysql       0        run      running  10h7m ago  10h7m ago
admin@app1:~$ nomad node status
ID        Node Pool  DC       Name   Class   Drain  Eligibility  Status
8c55b7e6  default    primary  data1  <none>  false  eligible     ready
0abc5c85  default    primary  app1   <none>  false  eligible     ready
520855e6  default    primary  data2  <none>  false  eligible     ready
9da121d6  default    primary  data3  <none>  false  eligible     ready
8f82e4c6  default    primary  app2   <none>  false  eligible     ready
admin@app1:~$
```

[...]

PR #24644, opened 2024-12-10
Allocation API: fix "no path to region" errors for non-global regions

tgross:
In #16872 we added support for unix domain sockets, but this required mutating the `Config` when parsing the address so as to remove the port number. In #23785 we fixed a bug where if the configuration was used across multiple clients (as in the autoscaler) that mutation would happen multiple times and the address would be incorrectly parsed.

When making `alloc log`, `alloc fs`, or `alloc exec` calls where we have line-of-sight to the client, we attempt to make a HTTP API call directly to the client node. So we create a new API client from the same configuration and then set the address. But in this case we copy the private `url` field and that causes the URL parsing to be skipped for the new client.

This results in the region always being set to the string literal `"global"` (because of mTLS handling code introduced all the way back in 4d3b75d867da), unless the user has set the region specifically. This fails with an error "no path to region" when the cluster isn't non-global and requests are sent to a non-leader.

Arguably the "right" way of fixing this would be for `ClientConfig` not to change the API client's region to `"global"` in the first place, but as this is a public API and extremely longstanding behavior, it could potentially be a breaking change for some downstream consumers. Instead, we'll avoid copying the private `url` field so that the new address is re-parsed.

Fixes: https://github.com/hashicorp/nomad/issues/24635
Fixes: https://github.com/hashicorp/nomad/issues/24609
Ref: https://github.com/hashicorp/nomad/pull/16872
Ref: https://github.com/hashicorp/nomad/pull/23785
Ref: https://github.com/hashicorp/nomad/commit/4d3b75d867dae508011c198c84318f54b4aa6684
Ref: https://hashicorp.atlassian.net/browse/NET-11858

To reproduce, stand up a cluster with `region = "example"` with at least one client, in an environment where you have line-of-sight to all nodes (ex. local development environment should work fine). Deploy a job to that client and run `nomad alloc logs :alloc_id`.

### DIFF
--- a/.changelog/24644.txt
+++ b/.changelog/24644.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+api: Fixed a bug where alloc exec/logs/fs APIs would return errors for non-global regions
+```
--- a/api/api.go
+++ b/api/api.go
@@ -234,7 +234,6 @@ func (c *Config) ClientConfig(region, address string, tlsEnabled bool) *Config {
 		HttpAuth:   c.HttpAuth,
 		WaitTime:   c.WaitTime,
 		TLSConfig:  c.TLSConfig.Copy(),
-		url:        copyURL(c.url),
 	}
 
 	// Update the tls server name for connecting to a client
--- a/e2e/terraform/provision-nomad/etc/nomad.d/base.hcl
+++ b/e2e/terraform/provision-nomad/etc/nomad.d/base.hcl
@@ -1,6 +1,7 @@
 # Copyright (c) HashiCorp, Inc.
 # SPDX-License-Identifier: BUSL-1.1
 
+region       = "e2e"
 bind_addr    = "0.0.0.0"
 data_dir     = "${data_dir}"
 enable_debug = true


tgross:
Another way I considered fixing this would be to have the `Address` field get paired with private `scheme/host` fields and add a `(*Config) Copy()` method that only copies the user-configured / public fields and erases the private fields. Callers would only ever use the internal fields. That would prevent this class of bug in general at the cost of making the internals a lot more fiddly.

mismithhisler (dismissed):
LGTM

mismithhisler (commented):
Should we update this comment?
Edit: Whoops didn't link this correctly

tgross:
> Should we update [this](https://github.com/hashicorp/nomad/pull/24644/files#diff-bf98d5fab5bcfabded9769069e7318683f549ec91cff84c5f87a9ae3a3ca3d8dR209) comment?

That comment expressed the correct _intent_ but incorrectly described the _behavior_ before this bug fix. It's correct now.

tgross:
Sorry @mismithhisler I didn't noticed that a file was in conflict because of a move. Can you :+1: again?

mismithhisler (approved):

PR #24645, opened 2024-12-10
add noswap to secretdir tmpfs

chuckyz:
### Description
This adds `noswap` mount option to the `SecretDir` mount.

- Create an allocation that triggers a secret dir creation

(having a hell of a time trying to get tests to run locally so I'm going to rely on GHA to run unit tests (fingers crossed))

### Links
todo issue

### Contributor Checklist

### Reviewer Checklist
  backporting document.
  history should be preserved.

### DIFF
--- a/.changelog/24645.txt
+++ b/.changelog/24645.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-client: Add noswap mount option to secrets directory wheree supported on Linux
+client: Add noswap mount option to secrets directory where supported on Linux
 ```
--- a/.changelog/24645.txt
+++ b/.changelog/24645.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+client: Add noswap mount option to secrets directory wheree supported on Linux
+```
--- a/client/allocdir/fs_linux.go
+++ b/client/allocdir/fs_linux.go
@@ -73,7 +73,7 @@ func createSecretDir(dir string, size int) error {
 		}
 
 		flags := uintptr(syscall.MS_NOEXEC)
-		// Permanantly disable swap for tmpfs for SecretDir.
+		// Permanently disable swap for tmpfs for SecretDir.
 		options := fmt.Sprintf("size=%dm,noswap", size)
 		if err := syscall.Mount("tmpfs", dir, "tmpfs", flags, options); err != nil {
 			return os.NewSyscallError("mount", err)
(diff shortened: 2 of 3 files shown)


tgross (changes requested):
Hi @chuckyz, nice to hear from you!

This is a great idea, but the `noswap` option was only introduced in Linux 6.4 and so we can't apply it blindly (many users have older kernels). For example, on Ubuntu 22.04:

```
$ uname -r
5.15.0-126-generic
$ mkdir /home/tim/secrets
$ sudo mount -t tmpfs -o size=100M,noswap tmpfs /home/tim/secrets
mount: /home/tim/secrets: wrong fs type, bad option, bad superblock on tmpfs, missing codepage or helper program, or other error.

$ sudo mount -t tmpfs -o size=100M tmpfs /home/tim/secrets
$ mount | grep secrets
tmpfs on /home/tim/secrets type tmpfs (rw,relatime,size=102400k,inode64)
```

But I'd love to find a way to get this in because it's a good incremental improvement to security. We just need a way to test for this option at runtime (maybe try with swap and then fallback to not using it?).

Also, once we've got that figured out can you run `make cl` to add a changelog item? Thanks!

chuckyz:
Hey @tgross that sounds great.

Another option would potentially be `ramfs`, however then we'd (for safety/reliability) probably need to setup the bind-mount as read only and it'd probably make the code base a bit tougher as ideally there'd be an option for tmpfs, or ramfs.

I personally strongly prefer the idea of just sticking with tmpfs.

I was going to call `syscall.Uname`, and parse out >= 6.4, sound good?

tgross:
> I was going to call `syscall.Uname`, and parse out >= 6.4, sound good?

If we do that, we won't be applying it on RHEL9, where the kernel version is 5.14 but they've apparently backported the `noswap` option (in a [security patch](https://access.redhat.com/errata/RHSA-2024:4349.html), if you can believe that! This is what I was obliquely referring to as "frankenkernel"). So that's why I was suggesting we try with the `noswap` option and catch the error and retry if that fails.

chuckyz:
Gave this a shot, got tests _mostly_ working locally.  Would you consider maybe trying `syscall.MNT_DETACH` or sprinkling around `time.Sleep()` to make tests more consistent locally?  Not sure why but I'm really having a hard time actually un-mounting during cleanup phases of tests.

Also I got fancy with this solution and then realized KISS and tried to make this way simpler (catch error once, then try again before returning a real error).

tgross (changes requested):
Looks great @chuckyz! One last item: can you run `make cl` to add a changelog entry. Something like "client: add noswap directive to secrets directory where supported on Linux"

chuckyz:
Should be good to go!

tgross (approved):
LGTM! Thanks @chuckyz!

PR #24646, opened 2024-12-10
plan: inform user if a deployment will be created

schmichael:
### Description

Just a proof of concept because the plan output when deployments are involved is a bit non-obvious. When updating a job with count=3, the plan will only show the first step of the deployment (as determined by update.max_parallel):

```
+/- Job: "sleeper"
+/- Task Group: "sleeper" (1 create/destroy update, 2 ignore)
  +/- Task: "sleeper" (forces create/destroy update)
      +/- Config {
```

This hack plumbs through whether a deployment was attached to the plan. Obviously the output and field names need improving. For backward compatibility we probably shouldn't output _anything_ if there isn't a deployment, but this PoC includes a message for either case for ease of testing/demonstration.

1. `nomad job run -var dur=60 sleeper.nomad.hcl`
2. `nomad job plan -var dur=900 sleeper.nomad.hcl`

> sleep.nomad.hcl
```hcl
variable "dur" {
  type = string
}

job "sleeper" {
  group "sleeper" {
    count = 3

    task "sleeper" {
      driver = "raw_exec"

      config {
        command = "sleep"
        args = [var.dur]
      }
    }
  }
}
```

### Contributor Checklist

### Reviewer Checklist
  backporting document.
  history should be preserved.
  within the public repository.

### DIFF
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -1525,6 +1525,7 @@ type JobPlanResponse struct {
 	CreatedEvals       []*Evaluation
 	Diff               *JobDiff
 	Annotations        *PlanAnnotations
+	CauseDeployment    bool
 	FailedTGAllocs     map[string]*AllocationMetric
 	NextPeriodicLaunch time.Time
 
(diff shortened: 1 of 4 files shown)


tgross (commented):
I like this overall idea. I'd probably return some kind of object rather than a boolean here, because once you have this information you pretty much always want to add to it (ex. "this will create a deployment with X and Y attributes").

Issue #24608, opened 2024-12-05
AWS cloud auto-join doesn't work with IPv6 addresses.


sigmaris:
### Nomad version

```
Nomad v1.9.3
BuildDate 2024-11-11T16:35:41Z
Revision d92bf1014886c0ff9f882f4a2691d5ae8ad8131c
```

### Operating system and Environment details

We have Nomad running on Fedora 40 in AWS, in a dual-stack VPC with an AWS-allocated IPv6 block - the Nomad servers and clients are assigned IPv6 addresses and advertise them for http, rpc and serf.

We are not using Consul.

### Issue

We are trying to join clients outside AWS to the server cluster in AWS, using IPv6.

When using the Cloud auto-join feature, and specifying that **go-discover** should use IPv6 addresses, the Nomad clients cannot connect to the servers, as it seems like the format of IPv6 addresses produced by **go-discover** isn't accepted by Nomad.

### Reproduction steps

* Set up a Nomad server cluster in AWS VPC with IPv6 enabled (dual stack)
* Configure the Nomad servers to advertise IPv6 addresses as described in https://developer.hashicorp.com/nomad/docs/operations/ipv6-support.
* Give the Nomad server EC2 instances an identifiable tag e.g. nomad-cluster=something.
* Configure a Nomad client to use Cloud auto-join to find by tag and join the server cluster, using IPv6 addresses, e.g. with this configuration: `retry_join = ["provider=aws region=eu-west-1 addr_type=public_v6 tag_key=nomad-cluster tag_value=something"]`.
* Start the Nomad client.

#### Expected Result

The Nomad client should discover the IPv6 addresses of the servers and connect to them.

#### Actual Result

We see in the client logs (below) that the clients seem to discover the IPv6 addresses OK, but then Nomad does not accept them due to the address format - I believe **go-discover** should be adding square brackets around the addresses before passing them to Nomad:

### Job file (if appropriate)

N/A

### Nomad Client logs (if appropriate)

```
   2024-12-04T18:02:55.473Z [DEBUG] agent.joiner: discover-aws: Found ip addresses: [2a05:d018:1f18:f300:61e0:6696:db81:77b 2a05:d018:1f18:f301:b3f:c4b4:d61f:6a31 2a05:d018:1f18:f301:3631:4790:7b7c:15d1 2a05:d018:1f18:f305:c24d:3416:b527:9cf0 2a05:d018:1f18:f302:8c1b:9a43:3d7a:3d73]
[...]

PR #24649, opened 2024-12-11
discovery: correctly handle IPv6 addresses from go-discover

tgross:
Nomad sets a default port when resolving server addresses that don't have one. When we get a "bare" IPv6 address without a port, we end up with an unexpected error "too many colons in address" when we try to split the address and host, because the standard library function expects IPv6 addresses to be wrapped in brackets as recommended by RFC5952. User-configured addresses avoid this problem by accepting IP address and port as separate configuration values, but go-discover emits "bare" IPv6 addresses without a port in IPv6 environments.

Fix this by adding brackets to IPv6 addresses when we get the "too many colons" error from the stdlib. This will still give erroneous results if the address includes the port but is missing brackets, but there's no way to unambiguously parse that address.

Ref: https://www.rfc-editor.org/rfc/rfc5952
Fixes: https://github.com/hashicorp/nomad/issues/24608
Ref: https://hashicorp.atlassian.net/browse/NET-11857

### Contributor Checklist

### Reviewer Checklist
  backporting document.
  history should be preserved.
  within the public repository.

### DIFF
--- a/.changelog/24649.txt
+++ b/.changelog/24649.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+discovery: Fixed a bug where IPv6 addresses would not be accepted from cloud autojoin
+```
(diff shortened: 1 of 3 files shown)


mismithhisler (approved):
LGTM

schmichael (approved):

Issue #24292, opened 2024-10-24
CNI: Container loses network routes after node reboot, requires job restart


mr-karan:
### Nomad version

```
-> % nomad version
Nomad v1.9.1
BuildDate 2024-10-21T09:00:50Z
Revision d9ec23f0c1035401e9df6c64d6ffb8bffc555a5e
```

### Operating system and Environment details

Ubuntu 24.04, Hetzner VM

### Issue

After a node reboot following system upgrades, containers started by Nomad lose network connectivity due to missing network routes, despite having proper IP allocation. The issue resolves after a job restart. The issue only manifesting after reboot indicates CNI's network state recovery isn't handling route restoration correctly, despite having all other network components in place.

#### Key Findings

1. **Partial Network Setup**: The critical observation is that CNI completes IP allocation (container gets proper IP) and iptables rules are set up correctly, but routes are missing. This suggests CNI's routing setup phase is failing silently after reboot.

2. **Networking State Inconsistency**:
   - IP is allocated from correct IPAM range (172.26.64.0/20)
   - IPTables rules are properly created
   - eth0 interface exists
   - BUT routing table is empty
   
   This specific combination indicates CNI's network setup is partially complete but fails during route installation.

### Reproduction steps

1. Have Nomad jobs running with bridge networking
2. Perform system upgrade requiring reboot
3. After node is back, containers start but have no network connectivity
4. `ip route` shows empty routing table in container namespace
5. Job restart fixes the issue

#### Technical Details

**Non-working state (after reboot):**
```
# Container network namespace has only loopback
1: lo: <LOOPBACK,UP,LOWER_UP>
    inet 127.0.0.1/8

# Empty routing table
/app # ip route
/app #

# Primary container shows no network
"NetworkSettings": {
    "Bridge": "",
    "Networks": {
        "none": {
            "IPAMConfig": null,
```

**Working state (after job restart):**
```
# Container has proper interface and IP
1: lo: <LOOPBACK,UP,LOWER_UP>
2: eth0@if19: <BROADCAST,MULTICAST,UP,LOWER_UP>
    inet 172.26.64.31/20 brd 172.26.79.255 scope global eth0

# IPTables rules exist for both states
-A NOMAD-ADMIN -d 172.26.64.0/20 -o nomad -j ACCEPT
-A POSTROUTING -s 172.26.64.31/32 -j CNI-c1ff8471ac64d933382ca8e1
```

### Job file (if appropriate)

```hcl
job "doggo" {
[...]

PR #24650, opened 2024-12-11
CNI: use tmpfs location for ipam plugin

tgross:
When a Nomad host reboots, the network namespace files in the tmpfs in `/var/run` are wiped out. So when we restore allocations after a host reboot, we need to be able to restore both the network namespace and the network configuration. But because the netns is newly created and we need to run the CNI plugins again, this create potential conflicts with the IPAM plugin which has written state to persistent disk at `/var/lib/cni`. These IPs aren't the ones advertised to Consul, so there's no particular reason to keep them around after a host reboot because all virtual interfaces need to be recreated too.

Reconfigure the CNI bridge configuration to use `/var/run/cni` as its state directory. We already expect this location to be created by CNI because the netns files are hard-coded to be created there too in `libcni`.

Note this does not fix the problem described for Docker in #24292 because that appears to be related to the netns itself being restored unexpectedly from Docker's state.

Ref: https://github.com/hashicorp/nomad/issues/24292#issuecomment-2537078584
Ref: https://www.cni.dev/plugins/current/ipam/host-local/#files

Run a cluster on a set of VMs, with at least one client. This can't be a server+client because we need to reboot the hosts. You should probably set the `server.heartbeat_grace = "5m"` to give yourself time to work.

* Run a non-Docker task with `network.mode = "bridge"`. Wait for it to be healthy.
* Reboot the client host.
* Make sure the alloc is restored, the tasks are restarted, and networking works.

### Contributor Checklist

### Reviewer Checklist
  backporting document.
  history should be preserved.
  within the public repository.


tgross:
I'm going to move this back into draft until we've had a bit more time to look into the CNI check command workflow being discussed in https://github.com/hashicorp/nomad/issues/24292#issuecomment-2539404666

jrasell (approved):
LGTM!

I tested this using an [AWS environment with 1 server and 1 client instance](https://github.com/jrasell/dev-mess/tree/main/nomad/lab/aws/nomad-dev-cluster). The client instance has CNI v1.3.0 plugins installed.

JobSpec:
```hcl
job "example" {

  group "webserver" {
    network {
      mode = "bridge"
      port "http" {
        to = 80
      }
    }

    task "python3" {
      driver = "exec"

      config {
        command = "python3"
        args    = ["-m", "http.server", "80"]
      }

      resources {
        cpu    = 500
        memory = 256
      }
    }
  }
}
```

Running through the steps using a build from main `f4529485563924462dbdccdd1b4cacbd9d68616e` when I rebooted the instance the allocation failed with the error detailed below:
```console
2024-12-13T10:57:44Z  Setup Failure          failed to setup alloc: pre-run hook "network" failed: failed to configure networking for alloc: failed to configure network: plugin type="bridge" failed (add): failed to allocate for range 0: 172.26.64.12 has been allocated to 0dc525f4-c491-7e25-a67d-0121069ad55e, duplicate allocation is not allowed
2024-12-13T10:57:41Z  Failed Restoring Task  failed to restore task; will not run until server is contacted
```

I then tested this patch `9e1a365ae3d22f04c6bb8aa0ce0fee6d1f83ae6`f I performed the same steps as detailed in the PR and performed in the previous test. When the instance was rebooted the following task events are recorded:
```console
2024-12-13T11:08:42Z  Started                Task started by client
2024-12-13T11:08:41Z  Failed Restoring Task  failed to restore task; will not run until server is contacted
```
The allocation HTTP server is accessible and responds as it did before the reboot.

ygersie:
@tgross this change broke upgrades for existing workloads.. The host-local IPAM storage path changed from `/var/lib/cni/networks/nomad/` to `/var/run/cni/nomad/`. This caused our workloads to get duplicate IP assignments leading to all kinds of issues.

tgross:
@ygersie are you on 1.9.4 or 1.9.5? https://github.com/hashicorp/nomad/pull/24658 didn't end up landing in the same release as 1.9.4 but that was supposed to avoid upgrade problems. But it could also just be that #24658 is actually causing the problem, so it might help to know which you're on.

ygersie:
Hey @tgross we upgraded from 1.8.5 to 1.8.9

tgross:
Hi @ygersie I dug into this and apparently #24658 not only didn't land in the same release as 1.9.4 but when we shipped 1.9.5 (which is the equivalent of your 1.8.9+ent), the backport process dropped the ball and so those commits did not get released in 1.8.9+ent (and 1.7.17+ent) either. That's embarrassing! But they're already in the release branch that should be shipping very soon (todayish)

ygersie:
@tgross I may be missing something here but even a CHECK wouldn't help prevent this issue? The state store of the host-local IPAM plugin has moved without copying current state. So next time an allocation is spun up there's no way for the host-local plugin to determine which IPs are still available, meaning you will likely get overlap. AFAICT the only way to prevent this would be to migrate / copy the state dir to the new location?

ygersie:
The host-local CNI plugin writes the handed out ip addresses to a local state dir. Now that the state dir location changed paths the new dir was empty. That means that the host-local ip started handing out ip addresses which are already in use. The only way to prevent this is by copying/moving the contents of the host-local plugin state dir.

tgross:
Ok, yeah I've now confirmed that behavior. My prior test was bogus because the first IP was selected from the beginning of the range. The client is getting CNI to hand out a new IP, but that IP isn't the one that the running network namespace has already been configured with, and can overlap the range. Let me look into what the best way to make that migration work is.

PR #24667, opened 2024-12-13
dynamic host volumes: add -type flag to volume init

tgross:
Adds a `-type` flag to the `volume init` command that generates an example volume specification with only those fields relevant to dynamic host volumes. This changeset also moves the string literals into uses of `go:embed`


pkazmierczak (approved):
LGTM!

PR #24668, opened 2024-12-13
[ui] Bugfix: prevent parent job from showing another job's dispatches when it has none

### LINKED ISSUE #24666: In the UI, one job displays the job launches of another
### Nomad version

Nomad v1.9.3
BuildDate 2024-11-11T16:35:41Z
Revision https://github.com/hashicorp/nomad/commit/d92bf1014886c0ff9f882f4a2691d5ae8ad8131c

Ubuntu 24.04 x86-64

### Issue

(This has been happening since the last UI redesign on version 1.7 or 1.8)

job1 has had some job launches of it running, job2 hasn't, when switching from visiting job1 to job2, I can see the job launches of job1 in the list of job2.

### Reproduction steps

Create both jobs (batch and parameterized), launch one instance of job1, and in the same tab visit job2, the job launch of job1 will appear in job2.

#### Expected Result

Each job should only display its own job launches

#### Actual Result

Usually if a job has had no recent job launches, it displays the job launches of the previously visited job.

image

### Job file (if appropriate)

**#Job1**

```
job "ping1" {
  datacenters = ["*"]
  type        = "batch"
  
  parameterized {
    payload       = "forbidden"
    meta_optional = ["irrelevant"]
  }

  group "ping" {
    task "ping" {
      driver = "raw_exec"

      config {
        command = "/bin/sh"
        args    = ["-c", "ping -c 100 -W 5 192.168.1.1"]
      }
    }
  }
}
```

**#Job2**
```
job "ping2" {
  datacenters = ["*"]
  type        = "batch"
  
  parameterized {
    payload       = "forbidden"
    meta_optional = ["irrelevant"]
  }

  group "ping" {
    task "ping" {
      driver = "raw_exec"

      config {
        command = "/bin/sh"
        args    = ["-c", "ping -c 100 -W 5 192.168.1.1"]
      }
    }
  }
}
```

philrenaud:
Resolves #24666

We had a bug where our `watchChildJobs` ember concurrency task (which kicks off whenever you hit a parameterized/periodic parent job route) would have a memoized list of childjobs from whenever it was last run. These components are not route-bound, and so don't shed data automatically when their route changes, and since we don't return a `[]` on /statuses with an ?index=1 parameter, but instead hold the blocking query open for further updates, we don't get any kind of "Clear what the app has in cache" signal.

This explicitly clears what the app has in cache on parent job route load.

### DIFF
--- a/.changelog/24668.txt
+++ b/.changelog/24668.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fix an issue where 2 parent jobs would see the others dispatches if it were otherwise empty
+```
--- a/ui/app/controllers/jobs/job/index.js
+++ b/ui/app/controllers/jobs/job/index.js
@@ -98,6 +98,7 @@ export default class IndexController extends Controller.extend(
     { id, namespace },
     throttle = Ember.testing ? 0 : 2000
   ) {
+    this.childJobs = [];
     while (true) {
       let params = {
         filter: `ParentID == "${id}"`,


mismithhisler (approved):
LGTM!

PR #24684, opened 2024-12-16
dynamic host volumes: account for other claims in capability check

tgross:
When we feasibility check a dynamic host volume against a volume request, we check the attachment mode and access mode. This only ensures that the capabilities match, but doesn't enforce the semantics of the capabilities against other claims that may be made on the allocation.

Add support for checking the requested capability against other allocations that the volume claimed.

### DIFF
--- a/scheduler/feasible.go
+++ b/scheduler/feasible.go
@@ -290,14 +290,16 @@ func (h *HostVolumeChecker) hostVolumeIsAvailable(
 		// examine all proposed allocs on the node, including those that might
 		// not have yet been persisted. they have nil pointers to their Job, so
 		// we have to go back to the state store to get them
-		seen := map[structs.NamespacedID]struct{}{}
+		seen := map[string]struct{}{}
 		for _, alloc := range proposed {
-			if _, ok := seen[alloc.JobNamespacedID()]; ok {
-				// all allocs for the same job will have the same read-only flag
-				// and capabilities, so we only need to check a given job once
+			uniqueGroup := alloc.JobNamespacedID().String() + alloc.TaskGroup
+			if _, ok := seen[uniqueGroup]; ok {
+				// all allocs for the same group will have the same read-only
+				// flag and capabilities, so we only need to check a given group
+				// once
 				continue
 			}
-			seen[alloc.JobNamespacedID()] = struct{}{}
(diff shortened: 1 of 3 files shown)


tgross:
Moving back to draft as Daniel has pointed out a fairly serious design flaw in how this works with `group.count > 1` on the first deploy.

tgross:
We're now using the scheduler's context's `ProposedAllocs` (requiring a trip back to the state store to get the `Job` structs, but that's no big deal), and that lets us remove the previously-proposed tracking of the previous alloc ID, counting writers, etc.

scheduler/feasible.go
```diff
@@ -242,6 +241,84 @@ func (h *HostVolumeChecker) hasVolumes(n *structs.Node) bool {
...
+	case structs.HostVolumeAccessModeSingleNodeSingleWriter:
+		// examine all proposed allocs on the node, including those that might
+		// not have yet been persisted. they have nil pointers to their Job, so
+		// we have to go back to the state store to get them
+		seen := map[structs.NamespacedID]struct{}{}
+		for _, alloc := range proposed {
+			if _, ok := seen[alloc.JobNamespacedID()]; ok {
+				// all allocs for the same job will have the same read-only flag
+				// and capabilities, so we only need to check a given job once
+				continue
```
gulducat, line 298: I see what this is doing, but I don't follow the reasoning in this comment.  I read it as meaning that different `group`s in the `job` will all have the same volume params, but this is more about _allocs_ per `group`, specifically `count`, right?

```suggestion
				// all allocs for the same task group will have the same read-only
				// flag and capabilities, so we only need to check a given job once
				continue
```

using the `job`'s ns+id as the `seen` key, without any reference to `group`, threw me a bit, I think. not that you should change that, just that the comment is pretty important to clarify this.

gulducat (approved):
lgtm! just one bit of comment clarity

scheduler/feasible.go
tgross: Yeah, very good point. So that's actually a bug here because if a job has multiple allocs for different task groups on the same node, then we'd potentially miss volume requests to check. Ex. a job has alloc A for group A and alloc B for group B on the same node, and only group B has a volume request. If we happen to check alloc A first we'd miss that.

Will fix.

PR #24693, opened 2024-12-17
E2E: use a variable for region

tgross:
In #24644 we set the region to "e2e" but forgot to setup the TLS certificate names appropriately. Swap the region out for a variable instead.


shoenig (dismissed):
LGTM

I think hclfmt wants
```
the following HCL files are out of sync
 M e2e/terraform/provision-nomad/main.tf
```

jrasell (approved):

PR #24700, opened 2024-12-17
stateful deployments: add corrections to API structs and methods

pkazmierczak:
This changeset includes changes accidentally left out from [24641](https://github.com/hashicorp/nomad/pull/24641).

### DIFF
--- a/api/tasks.go
+++ b/api/tasks.go
@@ -475,7 +475,6 @@ type VolumeMount struct {
 	Volume          *string `hcl:"volume,optional"`
 	Destination     *string `hcl:"destination,optional"`
 	ReadOnly        *bool   `mapstructure:"read_only" hcl:"read_only,optional"`
-	Sticky          *bool   `mapstructure:"sticky" hcl:"sticky,optional"`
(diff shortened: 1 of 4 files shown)


api/tasks.go
```diff
@@ -474,6 +475,7 @@ type VolumeMount struct {
 	Volume          *string `hcl:"volume,optional"`
 	Destination     *string `hcl:"destination,optional"`
 	ReadOnly        *bool   `mapstructure:"read_only" hcl:"read_only,optional"`
+	Sticky          *bool   `mapstructure:"sticky" hcl:"sticky,optional"`
```
tgross, line 478: I missed this on the original PR, but the `volume_mount` block is at the task level, not at the allocation level where scheduling is controlled. I don't think it has any meaning here. We should remove it from the `structs` package version as well.

pkazmierczak: oof that's a slip on my part, I fix

pkazmierczak: fixed in ac2bf73

tgross (approved):
LGTM :+1:

PR #24711, opened 2024-12-18
dynamic host volumes: tweak plugin fingerprint

gulducat:
Instead of a plugin `version` subcommand that responds with a string (established in #24497), respond to a `fingerprint` command with a data structure that we may extend in the future (such as plugin capabilities, like size constraints?). In the immediate term, it's still just the version: `{"version": "0.0.1"}`.

In addition to leaving the door open for future expansion, I think it will also avoid false positives detecting executables that just happen to respond to a `version` command.

This also reverses the ordering of the fingerprint string parts
from `plugins.host_volume.version.mkdir` (which aligned with CNI plugins)
to `plugins.host_volume.mkdir.version`.

Since this is a breaking change, any fine folks out there who have copied `example-plugin-mkfs` from the demo dir will need to copy the example again, and/or update any other custom plugins you may have.

relates to: #24479

### DIFF
--- a/client/hostvolumemanager/host_volume_plugin.go
+++ b/client/hostvolumemanager/host_volume_plugin.go
@@ -154,7 +154,7 @@ func (p *HostVolumePluginExternal) Fingerprint(ctx context.Context) (*PluginFing
 	}
 	fprint := &PluginFingerprint{}
 	if err := json.Unmarshal(stdout, fprint); err != nil {
-		return nil, fmt.Errorf("error parson fingerprint output as json: %w", err)
+		return nil, fmt.Errorf("error parsing fingerprint output as json: %w", err)
 	}
 	return fprint, nil
 }
--- a/client/hostvolumemanager/host_volume_plugin.go
+++ b/client/hostvolumemanager/host_volume_plugin.go
@@ -12,7 +12,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strings"
 
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-multierror"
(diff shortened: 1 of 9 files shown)


demo/hostvolume/example-plugin-mkfs
```diff
@@ -6,6 +6,11 @@
 
 set -euo pipefail
 
+version='0.0.1'
+fingerprint() {
+  echo '{"version": "'"$version"'"}'
```
tgross, line 11: I'd consider using `printf` here as you've done for the other outputs below, just to avoid this ugly escaping.

client/hostvolumemanager/host_volume_plugin.go
```diff
@@ -146,11 +152,11 @@ func (p *HostVolumePluginExternal) Version(ctx context.Context) (*version.Versio
 			"error", err)
 		return nil, fmt.Errorf("error getting version from plugin %q: %w", p.ID, err)
 	}
-	v, err := version.NewVersion(strings.TrimSpace(string(stdout)))
-	if err != nil {
-		return nil, fmt.Errorf("error with version from plugin: %w", err)
+	fprint := &PluginFingerprint{}
+	if err := json.Unmarshal(stdout, fprint); err != nil {
+		return nil, fmt.Errorf("error parson fingerprint output as json: %w", err)
```
tgross, line 157:
```suggestion
		return nil, fmt.Errorf("error parsing fingerprint output as json: %w", err)
```

tgross (approved):
LGTM!

demo/hostvolume/example-plugin-mkfs
gulducat: oh yeah! I think you did that `printf`, actually. I always forget that it's useful for this kinda thing.

client/hostvolumemanager/host_volume_plugin.go
gulducat: what, you don't like Jim Parsons of Big Bang Theory fame?

parsons:max_bytes(150000):strip_icc():focal(644x389:646x391)/jim-parsons-0a080c0832234195b66caf50436b56ab.jpg)

gulducat: it would be easy to search for that error message to debug. what if I do an ascii art of this image as a code comment to explain it?

(I am extra silly-headed today, don't mind me)

PR #24713, opened 2024-12-19
e2e: Update e2e tests to match new e2e infra Nomad region name.

jrasell:
### Contributor Checklist

### Reviewer Checklist
  backporting document.
  history should be preserved.
  within the public repository.


tgross (approved):
Thanks @jrasell! 

Something not being accounted for here is that the region is a variable, so if it were changed these tests would fail again. As far as I know you can safely leave `Region` out of `QueryOptions` entirely (the behavior around that was what we fixed in the PR that changed the region). That would leave the CNI test that explicitly is looking for the region but we could query the cluster for its region first or use some other value in the logs for the assertion (the test itself isn't about "regions" as a concept).

I'm going to approve in case you don't feel like dealing with any of that for now. :grin:

tgross (approved):
LGTM!

PR #24714, opened 2024-12-19
stateful deployments: validate there are no sticky per_alloc volume requests

pkazmierczak:
This changeset adds an additional validation that prevents users from setting `per_alloc` and `sticky` flags on volume requests. 


tgross (approved):
LGTM :+1:

Issue #24680, opened 2024-12-16
deps: migrate `github.com/aws/aws-sdk-go` to `v2`


jrasell:
As per https://github.com/aws/aws-sdk-go/issues/5270 the v1 AWS Go SDK will reach end of life on July 31, 2025 and is already in maintenance mode. We should migrate to https://aws.github.io/aws-sdk-go-v2/docs/.

Linked issue text:

> We previously [announced](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025) the upcoming end-of-support for AWS SDK for Go (v1).

> Per that announcement, as of 7/31/2024, the SDK has entered maintenance mode. Going forward, we will limit releases to address critical bug fixes and security issues only. The SDK will not receive API updates for new or existing services, or be updated to support new regions.

> We are no longer accepting external contributions to this SDK.

> Maintenance mode will last for 1 year. After 7/31/2025, the SDK will enter end-of-support, and no updates at all will be made. We recommend that you migrate to [AWS SDK for Go v2](https://aws.github.io/aws-sdk-go-v2/docs/). For additional details as well as information on how to migrate, please refer to the linked announcement.

PR #24701, opened 2024-12-17
sec: suppress osv alert in CRT

dduzgun-security:
### Description
Ref: [Thread](https://hashicorp.slack.com/archives/CUYKT2A73/p1734461997398449) 🧵 

### Testing & Reproduction steps
CRT should pass.

### Contributor Checklist
- [ ] **Changelog Entry** If this PR changes user-facing behavior, please generate and add a
  changelog entry using the `make cl` command.
- [ ] **Testing** Please add tests to cover any new functionality or to demonstrate bug fixes and
  ensure regressions will be caught.
- [ ] **Documentation** If the change impacts user-facing functionality such as the CLI, API, UI,
  and job configuration, please update the  Nomad website documentation to reflect this. Refer to
  the [website README](../website/README.md) for docs guidelines. Please also consider whether the
  change requires notes within the [upgrade guide](../website/content/docs/upgrade/upgrade-specific.mdx).

### Reviewer Checklist
- [ ] **Backport Labels** Please add the correct backport labels as described by the internal
  backporting document.
- [ ] **Commit Type** Ensure the correct merge method is selected which should be "squash and merge"
  in the majority of situations. The main exceptions are long-lived feature branches or merges where
  history should be preserved.
- [ ] **Enterprise PRs** If this is an enterprise only PR, please add any required changelog entry
  within the public repository.

PR #24720, opened 2024-12-19
deps: upgrade aws-sdk-go from v1 to v2

mismithhisler:
### Description

Upgrades aws-sdk-go from v1 to v2 as the v1 library is in maintenance mode.

Closes [GH #24680](https://github.com/hashicorp/nomad/issues/24680)

### Contributor Checklist

### Reviewer Checklist
  backporting document.
  history should be preserved.
  within the public repository.


mismithhisler:
> @mismithhisler can you check the impact of this change on binary size? In #24701 I noted that updating this library doesn't update the dependencies that have it as a transient dependency for v1, and so we'll be carrying around both versions.

I saw that and using `go mod why` I can see it's indirectly needed via the go-getter library.

It looks like it added 2.7MB to the binary size.

tgross:
> It looks like it added 2.7MB to the binary size.

Eh, ok. That stinks but not worth delaying the fix for. It'd be nice if we could nudge along the other HashiCorp libraries that are pulling this in.

client/fingerprint/env_aws.go
```diff
@@ -77,12 +77,16 @@ func (f *EnvAWSFingerprint) Fingerprint(request *FingerprintRequest, response *F
 		timeout = 1 * time.Millisecond
 	}
 
-	ec2meta, err := ec2MetaClient(f.endpoint, timeout)
+	ctx, cancel := context.WithTimeout(context.TODO(), timeout)
+	defer cancel()
+
+	imdsClient, err := imdsClient(ctx, f.endpoint)
```
tgross, line 83: Where does `f.endpoint` actually get set if we're no longer configuring it in the constructor? It looks like there's possibly fallback behavior if it's empty (maybe the client library reads the env var if set?), but are we ever using the `endpoint` field if we're falling back to that?

e2e/remotetasks/remotetasks.go
```diff
@@ -4,14 +4,15 @@
 package remotetasks
```
tgross, line 4: Should we just remove these tests and the related ECS infra? Let's see what @schmichael thinks. (But probably in a separate PR anyways.)

tgross (dismissed):
LGTM!

client/fingerprint/env_aws.go
mismithhisler: I was actually meaning to leave a comment around this.  I took it out here because from what I can tell, AWS makes no mention of `AWS_ENV_URL`, and I don't think you can specify custom IMDS endpoints.  So that really just leaves overriding the endpoint for testing.

jrasell: +1 on some comments here; took a moment of digging around to see what was happening, otherwise it looks a bit suspect on the surface.

client/fingerprint/env_aws.go
```diff
@@ -104,24 +108,25 @@ func (f *EnvAWSFingerprint) Fingerprint(request *FingerprintRequest, response *F
...
+		if resp == nil {
 			continue
-		} else if awsErr, ok := err.(awserr.Error); ok {
-			// if it's a URL error, assume we're not in an AWS environment
-			// TODO: better way to detect AWS? Check xen virtualization?
-			if _, ok := awsErr.OrigErr().(*url.Error); ok {
-				return nil
-			}
+		}
+		defer resp.Content.Close()
```
jrasell, line 120: This defer is inside a loop, so won't get executed until the iteration finishes and we reach the function end; is this OK? If we wanted to change this, we could move the loop body into it's own function.

client/fingerprint/env_aws.go
```diff
@@ -104,24 +108,25 @@ func (f *EnvAWSFingerprint) Fingerprint(request *FingerprintRequest, response *F
...
 
-			// not sure what other errors it would return
+		bytes, err := io.ReadAll(resp.Content)
+		if err != nil {
 			return err
 		}
+		v := strings.TrimSpace(string(bytes))
+		if v == "" {
+			f.logger.Debug("read an empty value", "attribute", k)
+		}
```
jrasell, line 129: Do we want a continue statement after the log line here which will match the existing behaviour and mean we do not have empty string attributes in node data? It would also be nice if we do need this continue to craft a test to catch this in the future, although I don't know if that is possible or feasible.

client/fingerprint/env_aws.go
```diff
@@ -104,24 +108,25 @@ func (f *EnvAWSFingerprint) Fingerprint(request *FingerprintRequest, response *F
...
 
 	for k, unique := range keys {
-		resp, err := ec2meta.GetMetadata(k)
-		v := strings.TrimSpace(resp)
-		if v == "" {
-			f.logger.Debug("read an empty value", "attribute", k)
-			continue
-		} else if awsErr, ok := err.(awserr.RequestFailure); ok {
-			f.logger.Debug("could not read attribute value", "attribute", k, "error", awsErr)
+		resp, err := imdsClient.GetMetadata(ctx, &imds.GetMetadataInput{
```
jrasell, line 111: Outside the scope for this PR but I wonder if there would be any benefit in the future from not passing individual keys to the metadata call, and rather get the whole response and process that. The benefit would be a single API call at the cost of more complex response processing.

jrasell (commented):
Thanks for doing this @mismithhisler! I've added a couple of inline comments/questions which would be good to resolve before merging but looks and works great.

I ran this locally on macOS and everything looked OK, with the correct message showing up in the logs:
```console
client.fingerprint_mgr.env_aws: error querying AWS IDMS URL, skipping
```

I then ran this PR on an AWS cluster and the node attributes looked how I would expect:
```console
$ nomad node status -verbose bd306835 |grep aws
kernel.version                           = 6.8.0-1009-aws
platform.aws.ami-id                      = ami-0474244c88b835731
platform.aws.instance-life-cycle         = on-demand
platform.aws.instance-type               = m5.xlarge
platform.aws.placement.availability-zone = eu-west-2a
unique.platform.aws.hostname             = ip-10-0-1-15.eu-west-2.compute.internal
unique.platform.aws.instance-id          = i-0765dd85a71bc86bb
unique.platform.aws.local-hostname       = ip-10-0-1-15.eu-west-2.compute.internal
unique.platform.aws.local-ipv4           = 10.0.1.15
unique.platform.aws.mac                  = 06:72:45:96:05:15
unique.platform.aws.public-ipv4          = xx.xxx.xxx.xxx
```

The output from the same Nomad client running the official v1.9.4 release:
```console
$ nomad node status -verbose bd306835 |grep aws
kernel.version                           = 6.8.0-1009-aws
platform.aws.ami-id                      = ami-0474244c88b835731
platform.aws.instance-life-cycle         = on-demand
platform.aws.instance-type               = m5.xlarge
platform.aws.placement.availability-zone = eu-west-2a
unique.platform.aws.hostname             = ip-10-0-1-15.eu-west-2.compute.internal
unique.platform.aws.instance-id          = i-0765dd85a71bc86bb
unique.platform.aws.local-hostname       = ip-10-0-1-15.eu-west-2.compute.internal
unique.platform.aws.local-ipv4           = 10.0.1.15
unique.platform.aws.mac                  = 06:72:45:96:05:15
unique.platform.aws.public-ipv4          = xx.xxx.xxx.xxx
```

go.mod
```diff
@@ -175,6 +179,16 @@ require (
 	github.com/apparentlymart/go-textseg/v13 v13.0.0 // indirect
 	github.com/apparentlymart/go-textseg/v15 v15.0.0 // indirect
 	github.com/armon/go-radix v1.0.0 // indirect
+	github.com/aws/aws-sdk-go v1.55.5 // indirect
```
pkazmierczak, line 182: why do we still need this? Is it a `go mod tidy` snafu or something actually pulls it as a dependency?

jrasell: This is `go-discover` et al.

mismithhisler: Also `go-getter`, I opened an issue to see if we can get the AWS SDK updated in that repo.  Would be nice to get this dependency removed.

client/fingerprint/env_aws.go
mismithhisler: Great catch! I'll get this updated.

client/fingerprint/env_aws.go
mismithhisler: This would be nice, but I'm not sure it's possible with the current IMDS.  If you pass in an empty path, you just get a list of available paths, not all the instance metadata.

client/fingerprint/env_aws.go
mismithhisler: In my testing, I haven't seen IMDS return empty data.  If an attribute wasn't available, the actual endpoint is not available and we get a 404 which is handled via `handleImdsError()`.

There is currently a test that returns empty data and mentions catching this in Hetzner cloud, but it's a false positive because it doesn't include the `ami-id`, so the code will actually return earlier in the fingerprint.  Also Hetzner uses a different metadata path (but maybe they didn't 4 years ago when this test was written).

I'd say we keep it as a "just-in-case" and I can update the test/comments accordingly?

e2e/remotetasks/remotetasks.go
schmichael: Please remove. We removed RTD from the docs and archived the repo. Not a bad idea, but incomplete and lacking in demand.

mismithhisler: I'll get this code removed in a followup PR.

tgross (dismissed):
LGTM! Nice refactoring work on the error/response body reading.

tgross (approved):

PR #24721, opened 2024-12-19
event stream: add events for dynamic host volumes

tgross:
Add a new topic to the event stream for host volumes. We'll emit events when a dynamic host volume is registered or deregistered, and whenever a node fingerprints with a changed volume.

CSI will be done in a separate PR.

In addition to the unit tests, you can use `nomad operator api "/v1/event/stream?topic=HostVolume"` to poll for host volume events while running the host volume demo [setup script](https://github.com/hashicorp/nomad/blob/main/demo/hostvolume/setup.sh)

### Contributor Checklist

### Reviewer Checklist
  backporting document.
  history should be preserved.
  within the public repository.

### DIFF
--- a/nomad/state/events.go
+++ b/nomad/state/events.go
@@ -41,6 +41,8 @@ var MsgTypeEvents = map[structs.MessageType]string{
 	structs.ServiceRegistrationUpsertRequestType:         structs.TypeServiceRegistration,
 	structs.ServiceRegistrationDeleteByIDRequestType:     structs.TypeServiceDeregistration,
 	structs.ServiceRegistrationDeleteByNodeIDRequestType: structs.TypeServiceDeregistration,
+	structs.HostVolumeRegisterRequestType:                structs.TypeHostVolumeRegistered,
+	structs.HostVolumeDeleteRequestType:                  structs.TypeHostVolumeDeleted,
 }
 
 func eventsFromChanges(tx ReadTxn, changes Changes) *structs.Events {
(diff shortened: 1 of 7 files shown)


nomad/state/events.go
```diff
@@ -358,6 +378,24 @@ func eventFromChange(change memdb.Change) (structs.Event, bool) {
...
 		}, true
+	case TableHostVolumes:
+		after, ok := change.After.(*structs.HostVolume)
+		if !ok {
+			return structs.Event{}, false
+		}
+		return structs.Event{
+			Topic: structs.TopicHostVolume,
+			Key:   after.ID,
+			FilterKeys: []string{
```
gulducat, line 389: DRY bikeshed: what if `structs.HostVolume` had an `Event()` method that returned this of itself?  then it could be used in both places.

```go
// nomad/structs/host_volumes.go
func (hv *HostVolume) Event() Event {
	if hv == nil {
		return Event{}
	}
	return Event{
		Topic:     TopicHostVolume,
		Key:       hv.ID,
		Namespace: hv.Namespace,
		FilterKeys: []string{
			hv.ID,
			hv.Name,
			hv.PluginID,
		},
		Payload: &HostVolumeEvent{
			Volume: hv,
		},
	}
}
```

extra bonus bikeshed: if all the other types did this too, then this `eventFromChange` method could be real small, akin to

```go
type eventer interface {
	Event() structs.Event
}
if thing, ok := change.After.(eventer); ok {
	return thing.Event(), true
} else {
	return structs.Event{}, false
}
```

so adding something to the event stream could be (almost) as simple as satisfying that interface.

gulducat (approved):
I have a little flight of fancy to consider, but lgtm!

nomad/state/events.go
tgross: I was thinking something along those lines while implementing this (and https://github.com/hashicorp/nomad/pull/24724). It'd improve the type safety too. But seeing as how I've already tried to shrink this PR by pulling out the CSI work to its own PR, any such refactoring should probably also be its own PR.

There's also a minor annoyance here too that the Eventer interface must be exported from `nomad/structs` none of that code will be covered by `nomad/structs` tests. The "test coverage goons" will frown, but there's really no logic there.

PR #24723, opened 2024-12-19
[ui] Show ALL regions' leaders when viewing servers route

### LINKED ISSUE #24309: Parity between `/ui/servers` and `nomad server members`
While I was recently doing an upgrade of our Nomad Servers (in a federated environment) I realized that the `/ui/servers` view will only display a single server as the 'leader' (which was the leader for the currently selected region) _despite the fact_ that the view _does_ show all servers (across all regions).

Conversely, if one runs `nomad server members` via CLI, it will also display all servers (across all regions) _but_ it does designated the leaders (for each returned region).

image
Note how I've got the table sorted by 'Leader' status so as to show that only 1x Leader server is surfaced in this view. Also important to understand is that the two listed datacenters exist in different regions.

### Proposal
IMHO (but others might disagree), since the `/ui/servers` pane has a table column displaying the region for the listed servers, I think that it would be nice to continue showing all servers across all regions but also to display the leader for each region.

### Use-cases
It's nice to be able to see (in the UI) all servers in all regions without switching between regions via drop-down and it would be doubly nice to be able to see all regions with a designation of which servers are leaders (without switching between regions via drop-down menu).

### Attempted Solutions
Possible currently; just adds and extra click.

philrenaud:
Resolves #24309 

In a multi-region setup:
image

In a single-region setup:
image

### DIFF
--- a/ui/app/models/agent.js
+++ b/ui/app/models/agent.js
@@ -7,10 +7,10 @@ import { inject as service } from '@ember/service';
 import { computed } from '@ember/object';
 import Model from '@ember-data/model';
 import { attr } from '@ember-data/model';
-import classic from 'ember-classic-decorator';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
 import formatHost from 'nomad-ui/utils/format-host';
 
-@classic
 export default class Agent extends Model {
   @service system;
 
@@ -29,9 +29,12 @@ export default class Agent extends Model {
     return formatHost(address, rpcPort);
   }
 
-  @computed('rpcAddr', 'system.leader.rpcAddr')
-  get isLeader() {
-    return this.get('system.leader.rpcAddr') === this.rpcAddr;
+  @tracked isLeader = false;
+
+  @action async checkForLeadership() {
+    const leaders = await this.system.leaders;
+    this.isLeader = leaders.includes(this.rpcAddr);
+    return this.isLeader;
   }
 
   @computed('tags.build')
--- a/ui/app/routes/clients.js
+++ b/ui/app/routes/clients.js
@@ -13,11 +13,6 @@ import classic from 'ember-classic-decorator';
 @classic
 export default class ClientsRoute extends Route.extend(WithForbiddenState) {
   @service store;
-  @service system;
-
-  beforeModel() {
-    return this.get('system.leader');
-  }
 
   model() {
     return RSVP.hash({
--- a/ui/app/services/system.js
+++ b/ui/app/services/system.js
@@ -12,27 +12,23 @@ import { namespace } from '../adapters/application';
 import jsonWithDefault from '../utils/json-with-default';
 import classic from 'ember-classic-decorator';
 import { task } from 'ember-concurrency';
-
 @classic
 export default class SystemService extends Service {
   @service token;
   @service store;
 
-  @computed('activeRegion')
-  get leader() {
-    const token = this.token;
-
-    return PromiseObject.create({
-      promise: token
-        .authorizedRequest(`/${namespace}/status/leader`)
-        .then((res) => res.json())
-        .then((rpcAddr) => ({ rpcAddr }))
-        .then((leader) => {
-          // Dirty self so leader can be used as a dependent key
-          this.notifyPropertyChange('leader.rpcAddr');
-          return leader;
-        }),
-    });
+  /**
+   * Iterates over all regions and returns a list of leaders' rpcAddrs
+   */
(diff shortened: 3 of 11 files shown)


ui/app/models/agent.js
```diff
@@ -29,9 +29,12 @@ export default class Agent extends Model {
...
 
-  @computed('rpcAddr', 'system.leader.rpcAddr')
-  get isLeader() {
-    return this.get('system.leader.rpcAddr') === this.rpcAddr;
+  @tracked isLeader = false;
+
+  @action async checkForLeadership() {
+    const leaders = await this.system.leaders;
+    this.isLeader = leaders.includes(this.rpcAddr);
+    return this.isLeader;
```
philrenaud, line 37: using `tracked` is more idiomatic than the isLeader getter in Ember Octane. And since service getters (like `this.get('system.leader....)` was the thing holding it back as a classic model, removed the `@classic` tag as well.

ui/app/services/system.js
```diff
@@ -12,27 +12,23 @@ import { namespace } from '../adapters/application';
...
+   */
+  @computed('regions.[]')
+  get leaders() {
+    return Promise.all(
+      this.regions.map((region) => {
+        return this.token
+          .authorizedRequest(`/${namespace}/status/leader?region=${region}`)
+          .then((res) => res.json());
+      })
+    );
```
philrenaud, line 31: Best as I can tell, the only thing that uses system.leader was the servers page, which is now using .leaders instead. Removed the old one.

ui/app/routes/clients.js
```diff
@@ -13,11 +13,6 @@ import classic from 'ember-classic-decorator';
 @classic
 export default class ClientsRoute extends Route.extend(WithForbiddenState) {
   @service store;
-  @service system;
-
-  beforeModel() {
-    return this.get('system.leader');
-  }
```
philrenaud, line 20: As best as I can tell, this was accidentally left in here by copying the servers route page ~9 years ago.

ui/app/templates/components/server-agent-row.hbs
```diff
@@ -17,13 +17,20 @@
...
+  <Hds::Badge
+    @text={{if
+      this.agent.isLeader
+      (if
+        this.agent.system.shouldShowRegions
+        (concat "True" " (" this.agent.region ")")
+        "True"
+      )
+      "False"
+    }}
```
philrenaud, line 29: Funky formatting, but:
1. If not a leader, just show "False"
2. If a leader, show "True"
3. If there are multiple regions at play, show region name in parentheses.

ui/app/services/system.js
tgross: What's the `${namespace}` referring to here? `v1`?

ui/app/routes/clients.js
DingoEatingFuzz: nomad-bushes

ui/app/services/system.js
philrenaud: Yeah, `v1`! It's my third favourite among the three ways we use the term "namespace" in the UI.

It imports from https://github.com/hashicorp/nomad/blob/0324e781d45e462709ecaeac039e636cdb1720f0/ui/app/adapters/application.js#L15

ui/mirage/scenarios/default.js
```diff
@@ -92,6 +92,9 @@ function smallCluster(server) {
   server.create('feature', { name: 'Dynamic Application Sizing' });
   server.create('feature', { name: 'Sentinel Policies' });
   server.createList('agent', 3, 'withConsulLink', 'withVaultLink');
+  if (withRegions) {
+    server.db.agents[0].member.Tags.region = server.db.regions[0].id;
+  }
```
philrenaud, line 97: This makes our vercel/mirage demo UI show leadership again.

ui/mirage/config.js
```diff
@@ -13,8 +13,14 @@ import { copy } from 'ember-copy';
...
-export function findLeader(schema) {
-  const agent = schema.agents.first();
+export function findLeader(schema, region = null) {
+  let agent;
+  let agents = schema.agents.all().models;
+  if (region) {
+    agent = agents.find((agent) => agent.member?.Tags?.region === region);
+  } else {
+    agent = agents[0];
+  }
```
philrenaud, line 23: Both update this to use an optional region param, and also start using the plain js `[0]` instead of the mirage-specific `.first()`

tgross (approved):
LGTM

PR #24738, opened 2024-12-19
event stream: refactor to use Eventer interface

tgross:
_Experimental_ 

Add an `Eventer` interface to the core structs so they can emit their own events, and use this to refactor the `eventFromChange` function at the core of the event stream.

Ref: https://github.com/hashicorp/nomad/pull/24721#discussion_r1893028481


tgross:
Leaving a note here that I don't _love_ this approach. I suspect (but haven't yet checked) that none of these functions will get inlined across packages. This drags the event construction away from the `nomad/state` package. But maybe that doesn't matter given the `FooEvent` structs are already defined there.

tgross:
Going to close this out. An interesting experiment but not worth the code churn. Maybe if we were generating more methods.

PR #24739, opened 2024-12-20
taskrunner: fix panic when a task that has a dynamic user is recovered

gabivlj:
Before this, the nomad client would crash with:
```
panic: assignment to entry in nil map
```

### Description
We are seeing in our production clusters the following crash when this codepath is being hit.

The unit test I added doesn't pass without this change.

### Contributor Checklist

### DIFF
--- a/client/allocrunner/taskrunner/dynamic_users_hook.go
+++ b/client/allocrunner/taskrunner/dynamic_users_hook.go
@@ -59,6 +59,7 @@ func (h *dynamicUsersHook) Prestart(_ context.Context, request *interfaces.TaskP
 		return nil
 	}
 
+	response.State = make(map[string]string, 1)
 	// if this is the restart case, the UGID will already be acquired and we
 	// just need to read it back out of the hook's state
 	if request.PreviousState != nil {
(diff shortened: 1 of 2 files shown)


client/allocrunner/taskrunner/dynamic_users_hook.go
```diff
@@ -64,7 +64,7 @@ func (h *dynamicUsersHook) Prestart(_ context.Context, request *interfaces.TaskP
 	if request.PreviousState != nil {
 		ugid, exists := request.PreviousState[dynamicUsersStateKey]
 		if exists {
-			response.State[dynamicUsersStateKey] = ugid
+			response.State = map[string]string{dynamicUsersStateKey: ugid}
```
mismithhisler, line 67: @jrasell A total nit, but could we just move the `response.State` map initialization that happens near the end of this function, to the beginning of the function?  My only rationale here is that being explicit about initializing the state upfront may help curb any future regressions.

mismithhisler: @gabivlj let me know if this suggestion makes sense.  Thanks!

gabivlj: Yup, this sounds good to me.

mismithhisler (approved):
Looks good, thank you for the fix!

PR #24747, opened 2024-12-20
dynamic host volumes: client-side tests, comments, tidying

gulducat:
Sprucing things up after the initial push to get it working.

I added test coverage before moving funcs around to ensure no functional changes.

Lots of lines, but hopefully an easy read!

### DIFF
--- a/client/hostvolumemanager/host_volume_plugin.go
+++ b/client/hostvolumemanager/host_volume_plugin.go
@@ -28,7 +28,6 @@ type HostVolumePlugin interface {
 	Fingerprint(ctx context.Context) (*PluginFingerprint, error)
 	Create(ctx context.Context, req *cstructs.ClientHostVolumeCreateRequest) (*HostVolumePluginCreateResponse, error)
 	Delete(ctx context.Context, req *cstructs.ClientHostVolumeDeleteRequest) error
-	// db TODO(1.10.0): update? resize? ??
 }
 
 type HostVolumePluginCreateResponse struct {
@@ -66,7 +65,8 @@ func (p *HostVolumePluginMkdir) Create(_ context.Context,
 	log.Debug("running plugin")
 
 	resp := &HostVolumePluginCreateResponse{
-		Path:      path,
+		Path: path,
+		// "mkdir" volumes, being simple directories, have unrestricted size
 		SizeBytes: 0,
 	}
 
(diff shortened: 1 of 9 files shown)


client/hostvolumemanager/host_volume_plugin.go
```diff
@@ -132,6 +143,10 @@ func NewHostVolumePluginExternal(log hclog.Logger,
 	}, nil
 }
 
+// HostVolumePluginExternal calls an executable on disk. All operations should
+// be idempotent, and safe to be called concurrently per volume ID.
+// For each call, the executable's stdout and stderr may be logged, so plugin
+// authors should not include any sensitive information in their plugin outputs.
```
tgross, line 149: These docstrings are likely to make their way into a spec for plugin authors. We should probably use the term "must" rather than "should" here in that [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119) sense of the terms.

client/hostvolumemanager/host_volumes.go
```diff
@@ -43,97 +45,37 @@ type Config struct {
...
 		pluginDir:      config.PluginDir,
 		sharedMountDir: config.SharedMountDir,
 		stateMgr:       config.StateMgr,
 		updateNodeVols: config.UpdateNodeVols,
-		log:            logger.Named("host_volume_manager"),
+		builtIns: map[string]HostVolumePlugin{
+			HostVolumePluginMkdirID: &HostVolumePluginMkdir{
+				ID:         HostVolumePluginMkdirID,
+				TargetPath: config.SharedMountDir,
+				log:        logger.With("plugin_id", HostVolumePluginMkdirID),
```
tgross, line 70: Nitpick: this logger won't be named `host_volume_manager` but the loggers we instantiate for the external plugins will, because they get cloned from `hvm.log.With`. Maybe we should hoist the `logger.Named` to the top of this method and then have the mkdir plugin call `.With` on that?

tgross (approved):
LGTM! :+1:

client/hostvolumemanager/host_volume_plugin.go
gulducat: >  **MUST**   This word, or the terms "REQUIRED" or "SHALL", mean that the definition is an absolute requirement of the specification.

I don't think this matches, because including secret strings doesn't break anything, and we have no way of enforcing it.

tgross: Just because it's not enforceable doesn't mean it's compliant with the spec. But "should be idempotent" is probably a better candidate for the stronger "must" wording, because that definitely _will_ break things.

tgross (approved):
LGTM! :+1:

PR #24748, opened 2024-12-20
dynamic host volumes: ensure unique name per node

gulducat:
Volume name must be unique per node.

For existing static host vols, the scheduler will prevent creation of duplicate names ("no node meets constraints"), and it will similarly prevent duplicate dynamic vols _after_ fingerprint (node re-registration), but we need to prevent concurrent duplicate creation that may happen within a fingerprint interval, too.

Here a lock is set for name+id during `Create` (and restore), then released when `Delete`d to prevent such very-fast or concurrent creates.

If the a volume spec is submitted again with a valid `id` set, it can still update the volume, and the plugin's `Create` will be run with the new spec.  I'm guessing that folks may want to do this if either A) the plugin has been updated and folks want to regenerate their vols in place? or B) to increase the vol size (in CSI parlance, "expand").  Not sure if perhaps any other reasons, but I wanted to leave this door open.

A last funny caveat is if a _static_ host volume is added to agent config after a dynamic one has been created, and the agent restarted to use it, the dynamic one will take precedence. A warning is logged to inform the user if this happens.

### DIFF
--- a/client/hostvolumemanager/host_volumes.go
+++ b/client/hostvolumemanager/host_volumes.go
@@ -6,6 +6,7 @@ package hostvolumemanager
 import (
 	"context"
 	"errors"
+	"fmt"
 	"path/filepath"
 	"sync"
 
@@ -19,6 +20,7 @@ import (
 var (
 	ErrPluginNotExists     = errors.New("no such plugin")
 	ErrPluginNotExecutable = errors.New("plugin not executable")
+	ErrVolumeNameExists    = errors.New("volume name already exists on this node")
 )
 
 // HostVolumeStateManager manages the lifecycle of volumes in client state.
@@ -53,6 +55,7 @@ type HostVolumeManager struct {
 	stateMgr       HostVolumeStateManager
 	updateNodeVols HostVolumeNodeUpdater
 	builtIns       map[string]HostVolumePlugin
+	names          *nameLocker
 	log            hclog.Logger
 }
 
@@ -71,7 +74,8 @@ func NewHostVolumeManager(logger hclog.Logger, config Config) *HostVolumeManager
 				log:        logger.With("plugin_id", HostVolumePluginMkdirID),
 			},
 		},
-		log: logger,
+		names: &nameLocker{},
+		log:   logger,
 	}
 }
 
@@ -85,8 +89,14 @@ func (hvm *HostVolumeManager) Create(ctx context.Context,
 		return nil, err
 	}
 
+	// can't have two of the same volume name per client node
+	if err := hvm.names.lock(req.Name); err != nil {
+		return nil, err
+	}
+
 	pluginResp, err := plug.Create(ctx, req)
 	if err != nil {
+		hvm.names.release(req.Name)
 		return nil, err
 	}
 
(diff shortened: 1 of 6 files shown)


client/hostvolumemanager/host_volumes.go
```diff
@@ -220,3 +247,29 @@ func genVolConfig(req *cstructs.ClientHostVolumeCreateRequest, resp *HostVolumeP
...
+// so we need to protect against concurrent duplicate creates.
+type nameLocker struct {
+	locks sync.Map
+}
+
+// lock the provided name, error if it was already locked
+func (l *nameLocker) lock(name, id string) error {
+	currentID, exists := l.locks.LoadOrStore(name, id)
+	if exists && id != currentID.(string) {
+		return fmt.Errorf("%w: name=%q id=%q", ErrVolumeNameExists, name, id)
```
tgross, line 262: The caller has the name and ID and already logs, so we should probably return only `ErrVolumeNameExists` here and then have the caller log the error rather than formatting this by hand and having the formatting mismatch for JSON loggers.

tgross (approved):
LGTM!

client/node_updater.go
```diff
@@ -50,7 +51,8 @@ SEND_BATCH:
 	// host volume updates
 	var hostVolChanged bool
 	c.batchNodeUpdates.batchHostVolumeUpdates(func(name string, vol *structs.ClientHostVolumeConfig) {
-		hostVolChanged = hvm.UpdateVolumeMap(newConfig.Node.HostVolumes, name, vol)
+		hostVolChanged = hvm.UpdateVolumeMap(c.logger.Named("node_updater").With("method", "batchFirstFingerprint"),
```
tgross, line 54: Non-blocking: I get where you're going with having the `method` field here but I do wonder if the terminology is weird for a cluster adminstrator looking through their logs. They probably don't care about the Go method name and more what component is emitting the log line.

Also, as far as I can tell there's only ever one possible log line we can emit from the function we're injecting this logger into, and that can only happen on first fingerprint? That would make adding the `method` field less useful, I think?

tgross (approved):
LGTM! I feel like the `sync.Map` is almost never what you want to use but in this case it was the perfect choice. :shipit:

client/node_updater.go
gulducat: theoretically it should only happen in that one situation, yes, but `UpdateVolumeMap` is called from 3 spots, and I had such a heck of a time tracing the various call sites to construct this whole thing, I figured I'd just go extra specific with these. I could add a `Trace` in `UpdateVolumeMap` to shine light on some of the callbacky weirdness in this area of the codebase...

otherwise, while it should only happen one way presently, I tend to do a lot of guarding against potential future regressions -- it's not likely that someone will modify this at all (I'm sure it's perfect ;), much less cause breakage, but I I'm a little paranoid about the future.

I thought of [`log_include_location`](https://developer.hashicorp.com/nomad/docs/configuration#log_include_location), which is great, but doesn't help in this case because the log isn't emitted right here where it would matter to know about it.  I could un-DRY this like the other similar ones, but overall I do like the consistency that `UpdateVolumeMap` provides.

:shrug:

tgross: Aye, no worries. Let's go ahead and ship this.

Issue #19382, opened 2023-12-08
packaging: Add dmidecode as a Suggested / Recommended package to Nomad's Linux packaging manifest.


shoenig:
Add `dmidecode` as a Suggested / Recommended package to Nomad's Linux packaging manifest.

Nomad 1.7 and later relies on the `dmidecode` command being available particularly in virtualized environments like Amazon EC2 for the detection of CPU performance data.

Not having this as part of the Linux packaging creates a bad UX for first starting up a Nomad 1.7 agent in these environments, e.g.

https://github.com/hashicorp/nomad/issues/19343
https://github.com/hashicorp/nomad/issues/19371

apollo13:
I'd recommend (pun intended) to put it into the recommended section and not suggested. Recommended is usually (not always though) installed by default whereas suggested is usually not installed and also not seen by the user.  Then again if this is only required for virtualized systems it might not make sense to install by default but solely error out in the nomad agent and as such communicating the problem clearly to the user.  On the other hand I'd imagine virtual machines are probably more common than bare metal nowadays for nomad, so it still might make sense to install `dmidecode` by default via a dependency.

MorphBonehunter:
I'm not convinced that installing dmidecode should be adviced as this could lead to suboptimal MHz readings (#19468) in virtual machine environments. So that's maybe also a point to rethink reopening #18213.

PR #24765, opened 2024-12-31
build: recommend dmidecode for rpm and deb packages

pkazmierczak:
Add `dmidecode` to "recommended" dependencies to improve UX on Linux machines that install Nomad via official packages. 

Resolves [#19382](https://github.com/hashicorp/nomad/issues/19382)
Internal ref: https://hashicorp.atlassian.net/browse/NET-10747


jrasell (approved):
LGTM, thanks @pkazmierczak!

Issue #24591, opened 2024-12-02
warn if removing a root key that was used for encrypting Variables


tgross:
The `nomad operator root keyring remove` command hits an RPC that checks if the key is currently active, but doesn't prevent the key from being removed if it's inactive but used to encrypt an existing Variable (or sign an existing Allocation). The GC job uses the same RPC, but has already checked that the key is not in use via the `IsRootKeyInUse` state store method.

We should have the `Keyring.Delete` RPC check the same method and return an error if the key is in use, unless a `Force` flag is set in the request. The GC job will always set that flag. Update the CLI to first send the request without `Force` and then ask the user to confirm their intent.

PR #24766, opened 2024-12-31
keyring: warn if removing a key that was used for encrypting variables

pkazmierczak:
This PR adds an additional check in the `Keyring.Delete` RPC to make sure we're not trying to delete a key that's been used to encrypt a variable. It also adds a `-force` flag for the CLI/API to sidestep that check. 

Resolves https://github.com/hashicorp/nomad/issues/24591
Internal ref: https://hashicorp.atlassian.net/browse/NET-11829

### DIFF
--- a/.changelog/24766.txt
+++ b/.changelog/24766.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+keyring: Warn if deleting a key previously used to encrypt a variable
+```
--- a/.changelog/24766.txt
+++ b/.changelog/24766.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-keyring: Warn if deleting a key previously used to encrypt a variable
+keyring: Warn if deleting a key previously used to encrypt an existing variable
 ```
--- a/api/keyring.go
+++ b/api/keyring.go
@@ -69,6 +69,8 @@ func (k *Keyring) Delete(opts *KeyringDeleteOptions, w *WriteOptions) (*WriteMet
 // KeyringDeleteOptions are parameters for the Delete API
 type KeyringDeleteOptions struct {
 	KeyID string // UUID
+	// Force can be used to force deletion of a root keyring that was used to encrypt
+	// an existing variable or to sign a workload identity
 	Force bool
 }
 
--- a/api/keyring.go
+++ b/api/keyring.go
@@ -6,6 +6,7 @@ package api
 import (
 	"fmt"
 	"net/url"
+	"strconv"
 )
 
 // Keyring is used to access the Variables keyring.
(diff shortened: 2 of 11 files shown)


.changelog/24766.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+keyring: Warn if deleting a key previously used to encrypt a variable
```
jrasell, line 2: Is it worth expanding on this to note the variable should still exist? It immediately reads to me that the warning applies to any key used in the past.

api/keyring.go
```diff
@@ -60,14 +61,15 @@ func (k *Keyring) List(q *QueryOptions) ([]*RootKeyMeta, *QueryMeta, error) {
...
-		url.PathEscape(opts.KeyID)), nil, nil, w)
+	wm, err := k.client.delete(fmt.Sprintf("/v1/operator/keyring/key/%v?force=%v",
+		url.PathEscape(opts.KeyID), strconv.FormatBool(opts.Force)), nil, nil, w)
 	return wm, err
 }
 
 // KeyringDeleteOptions are parameters for the Delete API
 type KeyringDeleteOptions struct {
 	KeyID string // UUID
+	Force bool
```
jrasell, line 72: I know not many options have this, but could we add a comment to this explaining what the force boolean does? It means external tool developers have an easy time knowing and understanding.

nomad/keyring_endpoint.go
```diff
@@ -348,6 +348,15 @@ func (k *Keyring) Delete(args *structs.KeyringDeleteRootKeyRequest, reply *struc
 		return fmt.Errorf("active root key cannot be deleted - call rotate first")
 	}
 
+	// make sure the key was used to encrypt an existing variable
+	rootKeyInUse, err := snap.IsRootKeyInUse(args.KeyID)
+	if err != nil {
+		return err
+	}
+	if rootKeyInUse && !args.Force {
+		return fmt.Errorf("root key in use, cannot delete")
```
jrasell, line 357: There is no formatting, so we could use the below (the suggestion won't compile as the file also needs the new import):
```suggestion
		return errors.New("root key in use, cannot delete")
```

jrasell (changes requested):
Thanks for doing this @pkazmierczak!

I tested this locally and couldn't get a delete to work even when using the force flag. I believe we need to populate the RPC request force parameter from the HTTP request [here](https://github.com/hashicorp/nomad/blob/2f4353412d1bcf305898f676a815d342304ffa28/command/agent/keyring_endpoint.go#L190) in order for the HTTP->RPC translation to handle this new parameter.

It would also be good if we could address the following items:
- The CLI help output doesn't include the new flag option.
- It would be nice to have API and CLI tests to cover this addition as the RPC test did not catch the issue detailed above.
- Document the API and CLI changes.

nomad/keyring_endpoint.go
pkazmierczak: yeah force of habit I guess, but `errors.New` is a better choice here

aimeeu (approved):
Thanks for updating the docs!

pkazmierczak:
@jrasell I addressed all your comments, thanks for the thorough review.

jrasell (approved):
LGTM, thanks @pkazmierczak!

I tested this locally on macOS attempting to delete root keys used for both variable encryption and task workload identity.

PR #24789, opened 2025-01-06
docs: dynamic host volumes API

tgross:
Add API documentation for the dynamic host volumes feature.

Ref: https://github.com/hashicorp/nomad/issues/15489
Ref: https://hashicorp.atlassian.net/browse/NET-11482

### Reviewer Checklist
- [ ] **Backport Labels** Please add the correct backport labels as described by the internal
  backporting document.
- [ ] **Commit Type** Ensure the correct merge method is selected which should be "squash and merge"
  in the majority of situations. The main exceptions are long-lived feature branches or merges where
  history should be preserved.
- [ ] **Enterprise PRs** If this is an enterprise only PR, please add any required changelog entry
  within the public repository.

PR #24795, opened 2025-01-07
dynamic host volumes: require node ID on register

tgross:
When registering a host volume created out-of-band, the volume will have been created on a specific node. Require the node ID field to be set.

Ref: https://github.com/hashicorp/nomad/pull/24789#discussion_r1904690799


pkazmierczak (approved):
👍

PR #24804, opened 2025-01-08
client: move 'waiting for previous alloc to terminate' log messages to info

pkazmierczak:
I'm not a 100% sure if this is the right move, but seeing as some users are hit by bugs that cause allocations to be stuck in `pending` state, they are forced to run agents with `debug` log levels which is very noisy. 

On the other hand, having these messages at `info` will significantly increase the noise of that level. I welcome comments and opinions on whether that's the right way to handle this issue.


tgross (approved):
LGTM

PR #24816, opened 2025-01-08
E2E: dynamic host volumes workflows

tgross:
Initial end-to-end tests for dynamic host volumes. This includes tests for two workflows:

* One where a dynamic host volume is created by a plugin and then mounted by a job.
* Another where a dynamic host volume is created out-of-band and registered by a job, then mounted by another job.

This changeset also moves the existing `volumes` E2E test package to the better-named `volume_mounts`.

Ref: https://hashicorp.atlassian.net/browse/NET-11551


pkazmierczak (approved):
LGTM!

PR #24797, opened 2025-01-07
dynamic host volume reference documentation

tgross:
"Feature branch" for dynamic host volume reference documentation PRs, so that we can re-review the reference docs holistically when we're done.

Includes:
- [x] API docs https://github.com/hashicorp/nomad/pull/24789
- [x] ACL policy docs https://github.com/hashicorp/nomad/pull/24790
- [x] Volume spec docs https://github.com/hashicorp/nomad/pull/24810
- [x] Plugin specification https://github.com/hashicorp/nomad/pull/24924
- [x] CLI docs https://github.com/hashicorp/nomad/pull/24814
- [x] Quota spec updates https://github.com/hashicorp/nomad/pull/24894
- [x] Job spec updates for `volume` ~and `volume_mount`~ https://github.com/hashicorp/nomad/pull/24828
- [x] Config updates for `host_volume` https://github.com/hashicorp/nomad/pull/24827
- [x] Sentinel updates https://github.com/hashicorp/nomad/pull/24825
- [x] Node pool governance updates https://github.com/hashicorp/nomad/pull/24826
- [x] "Considerations for stateful workloads" https://github.com/hashicorp/nomad/pull/24930
 
Ref: https://hashicorp.atlassian.net/browse/NET-11482

PR #24827, opened 2025-01-09
dynamic host volumes: client configuration docs

tgross:
Document the client configuration changes needed to support dynamic host volumes. This changeset excludes the plugin specification/concepts, which will be under a separate PR.

Ref: https://github.com/hashicorp/nomad/pull/24797
Ref: https://hashicorp.atlassian.net/browse/NET-11482

### DIFF
--- a/website/content/docs/configuration/client.mdx
+++ b/website/content/docs/configuration/client.mdx
@@ -201,6 +201,13 @@ client {
 - `host_volume` <code>([host_volume](#host_volume-block): nil)</code> - Exposes
   paths from the host as volumes that can be mounted into jobs.
 
+- `host_volume_plugin_dir` `(string: "")` - Specifies the directory to find host
+  volume plugins. When this parameter is empty, Nomad will generate the path
+  using the [top-level `data_dir`][top_level_data_dir] suffixed with
+  `host_volume_plugins`, like `"/opt/nomad/host_volume_plugins"`. This must be
+  an absolute path. Nomad will create the directory on the host, if it does not
+  exist when the agent process starts.
+
 - `host_network` <code>([host_network](#host_network-block): nil)</code> - Registers
   additional host networks with the node that can be selected when port mapping.
 
@@ -613,11 +620,14 @@ see the [drivers documentation](/nomad/docs/drivers).
 
 ### `host_volume` Block
 
-The `host_volume` block is used to make volumes available to jobs.
+The `host_volume` block is used to make volumes available to jobs. You can also
+configure [dynamic host volumes][] via the [`volume create`][] or [`volume
+register`][] commands.
 
 The key of the block corresponds to the name of the volume for use in the
-`source` parameter of a `"host"` type [`volume`](/nomad/docs/job-specification/volume)
-and ACLs.
+`source` parameter of a `"host"` type
+[`volume`](/nomad/docs/job-specification/volume) and ACLs. Host volumes in the
+configuration cannot have the same names as dynamic host volumes.
 
 ```hcl
 client {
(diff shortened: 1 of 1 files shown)


website/content/docs/configuration/client.mdx
```diff
@@ -201,6 +201,13 @@ client {
 - `host_volume` <code>([host_volume](#host_volume-block): nil)</code> - Exposes
   paths from the host as volumes that can be mounted into jobs.
 
+- `host_volume_plugin_dir` `(string: "")` - Specifies the directory to find host
+  volume plugins. When this parameter is empty, Nomad will generate the path
```
aimeeu, line 205:
```suggestion
  volume plugins. When this parameter is empty, Nomad generates the path
```

website/content/docs/configuration/client.mdx
```diff
@@ -201,6 +201,13 @@ client {
...
+  using the [top-level `data_dir`][top_level_data_dir] suffixed with
+  `host_volume_plugins`, like `"/opt/nomad/host_volume_plugins"`. This must be
+  an absolute path. Nomad will create the directory on the host, if it does not
```
aimeeu, line 208:
```suggestion
  an absolute path. Nomad creates the directory on the host, if it does not
```

aimeeu (approved):
Thanks Time! a few style nits

website/content/docs/configuration/client.mdx
```diff
@@ -201,6 +201,13 @@ client {
 - `host_volume` <code>([host_volume](#host_volume-block): nil)</code> - Exposes
   paths from the host as volumes that can be mounted into jobs.
 
+- `host_volume_plugin_dir` `(string: "")` - Specifies the directory to find host
+  volume plugins. When this parameter is empty, Nomad generates the path
+  using the [top-level `data_dir`][top_level_data_dir] suffixed with
+  `host_volume_plugins`, like `"/opt/nomad/host_volume_plugins"`. This must be
+  an absolute path. Nomad creates the directory on the host, if it does not
+  exist when the agent process starts.
```
gulducat, line 209:
```suggestion
  an absolute path.
```

it does not get auto-created.

and fwiw, this is the only mention of it in agent logs, whether this param set to something custom or not:

```
[DEBUG] client.fingerprint_mgr.host_volume_plugins: plugin dir does not exist: dir=/opt/nomad/data/host_volume_plugins
```

website/content/docs/configuration/client.mdx
```diff
@@ -613,11 +620,14 @@ see the [drivers documentation](/nomad/docs/drivers).
...
+The `host_volume` block is used to make volumes available to jobs. You can also
+configure [dynamic host volumes][] via the [`volume create`][] or [`volume
+register`][] commands.
 
 The key of the block corresponds to the name of the volume for use in the
-`source` parameter of a `"host"` type [`volume`](/nomad/docs/job-specification/volume)
-and ACLs.
+`source` parameter of a `"host"` type
+[`volume`](/nomad/docs/job-specification/volume) and ACLs. Host volumes in the
+configuration cannot have the same names as dynamic host volumes.
```
gulducat, line 630:
```suggestion
[`volume`](/nomad/docs/job-specification/volume) and ACLs. A host volume in the
configuration cannot have the same name as a dynamic host volume on the same node.
```

the plurals threw me off for whatever reason.

website/content/docs/configuration/client.mdx
tgross: Ah, good catch. I was thinking of the various datadirs which the agent does create, but we never create plugin dirs.

PR #24848, opened 2025-01-13
dynamic host volumes: plugin spec tweaks

gulducat:
Some plugin spec changes based on recent discussion.

* All env prefixed with `DHV_`
* Add: `DHV_VOLUME_ID`, `DHV_PLUGIN_DIR`
* 5 second timeout for `fingerprint` operation

Notice to anyone using the `example-plugin-mkfs` demo plugin: you'll need to copy this changed version of it.


tgross (approved):
LGTM!

Issue #18186, opened 2023-08-10
NOMAD_TOKEN provided by workload identity is not accepted in the events API endpoint.


bogue1979:
### Nomad version
Nomad v1.6.1
BuildDate 2023-07-21T13:49:42Z
Revision 515895c7690cdc72278018dc5dc58aca41204ccc

### Operating system and Environment details
single node test cluster
`Linux sb-danielr4 5.15.0-1038-gcp #46~20.04.1-Ubuntu SMP Fri Jul 14 09:48:19 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux`

### Issue
I try to use the events api with a NOMAD_TOKEN provided via workload identity.

My job looks something like this:
```
  job "scheduler-events" {
    namespace = "infra"
    ...
    group "scheduler-events-sink" {
      ...
      task "events-sink" {
        ...
        identity {
          env = true
          file = true
        }
      }
      task "events-vector" {
      ...
      }
    }
  }
```
Additionally I created a corresponding policy according to the ACL requirements documented in https://developer.hashicorp.com/nomad/api-docs/events like this:

```
# nomad acl policy info job-scheduler-events-policy
  Name        = job-scheduler-events-policy
  Description = for scheduler events
  CreateIndex = 1247
  ModifyIndex = 1835

  Associated Workload
  Namespace = infra
  JobID     = scheduler-events
  Group     = scheduler-events-sink
  Task      = events-sink

  Rules

  namespace "*" {
      capabilities = ["read-job"]
  }

  node {
      policy = "read"
  }
```

In my test I'm using the NOMAD_TOKEN stored in the allocation directory but I get  a permission denied error and the response status code is 500 Internal Server Error.
```
/var/nomad/alloc/2ea66317-5e6f-f4d4-eee0-e528270e14bb/events-sink/secrets# curl -s -N -H "Authorization: Bearer $(cat nomad_token)" "localhost:4646/v1/event/stream?namespace=infra&index=100&topic=Evaluation"
```
#### Expected Result
I can connect to the event stream the same way I would do with an other NOMAD_TOKEN with a similar policy.

tgross:
Hi @bogue1979! I happen to be elbow-deep in our authentication/authorization code this week and so this issue immediately grabbed my attention. It looks like the event stream has a weird alternate code path for checking ACLs for filtering by subscriber, and that code path doesn't properly account for Workload Identity.

[...]

PR #24849, opened 2025-01-13
Enable workload identity for event stream

mismithhisler:
### Description

Enables Workload Identities to subscribe to event streams.  Previously, the event broker used custom authentication logic that could only accept the secretID from an ACL Token.  This change refactors the event broker to use an authentication callback for handling authentication while subscribing and after various ACL events.  This allows us to inject the standardized `Authenticator` logic into the event broker.

Enable ACL's, create the appropriate policy, associate it with a workload, and use the WI as the Bearer token when subscribing to the events stream.

### Links

Fixes [GH #18186](https://github.com/hashicorp/nomad/issues/18186)

### Contributor Checklist

### Reviewer Checklist
  backporting document.
  history should be preserved.
  within the public repository.

### DIFF
--- a/nomad/event_endpoint.go
+++ b/nomad/event_endpoint.go
@@ -66,11 +66,11 @@ func (e *Event) stream(conn io.ReadWriteCloser) {
 			if err := e.srv.Authenticate(nil, &args); err != nil {
 				return err
 			}
-			acl, err := e.srv.ResolveACL(&args)
+			resolvedACL, err := e.srv.ResolveACL(&args)
 			if err != nil {
 				return err
 			}
-			return validateACL(args.Namespace, args.Topics, acl)
+			return validateACL(args.Namespace, args.Topics, resolvedACL)
 		},
 	}
 
--- a/nomad/event_endpoint.go
+++ b/nomad/event_endpoint.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/hashicorp/go-msgpack/v2/codec"
 
+	"github.com/hashicorp/nomad/acl"
 	"github.com/hashicorp/nomad/helper/pointer"
 	"github.com/hashicorp/nomad/nomad/stream"
 	"github.com/hashicorp/nomad/nomad/structs"
@@ -61,6 +62,16 @@ func (e *Event) stream(conn io.ReadWriteCloser) {
 		Topics:    args.Topics,
 		Index:     uint64(args.Index),
 		Namespace: args.Namespace,
+		Authenticate: func() error {
+			if err := e.srv.Authenticate(nil, &args); err != nil {
+				return err
+			}
+			acl, err := e.srv.ResolveACL(&args)
+			if err != nil {
+				return err
+			}
+			return validateACL(args.Namespace, args.Topics, acl)
+		},
 	}
 
 	// Get the servers broker and subscribe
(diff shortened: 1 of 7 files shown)


nomad/event_endpoint.go
```diff
@@ -213,3 +224,47 @@ func handleJsonResultError(err error, code *int64, encoder *codec.Encoder) {
...
+		switch topic {
+		case structs.TopicDeployment,
+			structs.TopicEvaluation,
+			structs.TopicAllocation,
+			structs.TopicJob,
+			structs.TopicService:
+			if ok := aclObj.AllowNsOp(namespace, acl.NamespaceCapabilityReadJob); !ok {
+				return structs.ErrPermissionDenied
+			}
+		case structs.TopicHostVolume:
```
mismithhisler, line 239: I don't see subscribing to `HostVolume` or `CSI Read Volume` in the [events api docs](https://developer.hashicorp.com/nomad/api-docs/events).  Should we remove this, or update the docs to include these options?

nomad/stream/event_broker_test.go
mismithhisler, line 1: A lot of tests that covered the custom authentication logic have been removed from here, as we are now using the `Authenticator` logic via the authentication callback.

nomad/event_endpoint.go
tgross: These are added on main but not a release build.

nomad/event_endpoint.go
```diff
@@ -61,6 +62,16 @@ func (e *Event) stream(conn io.ReadWriteCloser) {
 		Topics:    args.Topics,
 		Index:     uint64(args.Index),
 		Namespace: args.Namespace,
+		Authenticate: func() error {
+			if err := e.srv.Authenticate(nil, &args); err != nil {
+				return err
+			}
+			acl, err := e.srv.ResolveACL(&args)
```
jrasell, line 69: `acl` shadows an import name, could we change this variable name?

nomad/event_endpoint.go
```diff
@@ -74,21 +85,21 @@ func (e *Event) stream(conn io.ReadWriteCloser) {
...
 	if subErr != nil {
 		handleJsonResultError(subErr, pointer.Of(int64(500)), encoder)
 		return
 	}
 	defer subscription.Unsubscribe()
 
+	// because we have authenticated, the identity will be set, so extract expiration time
+	var exp time.Time
+	if c := args.GetIdentity().GetClaims(); c != nil {
+		exp = c.Expiry.Time()
```
jrasell, line 98: Is there any possibility `Expiry` can be nil?

nomad/event_endpoint.go
```diff
@@ -213,3 +224,47 @@ func handleJsonResultError(err error, code *int64, encoder *codec.Encoder) {
 		Error: structs.NewRpcError(err, code),
 	})
 }
+
+func validateACL(namespace string, topics map[structs.Topic][]string, aclObj *acl.ACL) error {
```
jrasell, line 228: Is it worth adding a table-driven test to cover the possibilities of this function? If think our current tests only cover a portion of the function.

jrasell (commented):
LGTM, thanks @mismithhisler!

I think the PR still needs a changelog entry and I added a couple of inline queries, so I'll comment as my approval will likely get removed with at least the changelog commit.

nomad/event_endpoint.go
mismithhisler: It's probably possible, but the `Time()` method does a nil check on Expiry.  If you want, I can add a nil check here just to be defensive?

nomad/event_endpoint.go
mismithhisler: Yes I think it is.  I originally was confused about why these untested portions were even present until @tgross they are new code added for 1.10.

I'll refactor the existing test to cover this better but just need to be careful to do it in a way that doesn't make backporting too painful.

tgross: Except for topic-to-policy mapping, mostly this function is covered by `acl` package tests. So if we're going to add tests here, the primary value would be in making sure we're testing that mapping against corner cases like wildcard namespaces.

mismithhisler: My thought here was not to test the internals of `AllowNsOp` (although I see the overlap) but make sure that this validation function correctly matches what's listed in the required ACL section of the [event stream](https://developer.hashicorp.com/nomad/api-docs/events#event-stream) docs.

Let me know if you think that's too much overlap with the `acl` package tests

tgross: Makes sense to me!

jrasell (approved):

PR #24850, opened 2025-01-13
dynamic host volumes: capacity_max may be unset during register

tgross:
When using the register workflow, `capacity_max` is ignored so is likely unset. If the volume is then updated later, the check we had for valid updates assumes that the value was previously. Only perform this check if the value is set.


gulducat (approved):
lgtm!

gulducat (approved):
looks even better this time! :laughing:

PR #24810, opened 2025-01-08
docs: dynamic host volume specification

tgross:
Dynamic host volumes use the same specification file as CSI volumes but require a different set of parameters and have different semantics. This changeset splits the volume specification page into separate CSI and dynamic host volumes spec pages.

While migrating the CSI page, I've also edited it to bring it more in line with the style guide: removed passive voice and future tense, inclusive language, alphabetized the (chaotic!) parameters list, etc.

Ref: https://github.com/hashicorp/nomad/pull/24797
Ref: https://hashicorp.atlassian.net/browse/NET-11482

PR #24851, opened 2025-01-13
dynamic host volumes: allow for node pool and plugin ID changes

tgross:
Update dynamic host volume validation and update logic to allow for changes to the node pool and plugin ID. If the client's node pool changes we'll sync up the correct node pool for the volumes already placed on that client. We'll also allow the plugin ID to be changed to allow for new versions of plugins supporting the same volume over time.

Ref: https://github.com/hashicorp/nomad/pull/24810#discussion_r1911366838
Ref: https://github.com/hashicorp/nomad/pull/24810#discussion_r1911361745

(Note this PR will conflict with https://github.com/hashicorp/nomad/pull/24857, so whichever one lands last will need rebasing.)

### DIFF
--- a/nomad/state/state_store_host_volumes.go
+++ b/nomad/state/state_store_host_volumes.go
@@ -228,25 +228,42 @@ func upsertHostVolumeForNode(txn *txn, node *structs.Node, index uint64) error {
 	if err != nil {
 		return err
 	}
+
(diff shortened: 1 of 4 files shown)


gulducat (approved):
lgtm!

PR #24852, opened 2025-01-13
dynamic host volumes: serialize ops per volume

gulducat:
Let only one of any create/register/delete run at a time per volume ID.

* DHV plugins can assume that Nomad will not run concurrent operations for the same volume
* We avoid interleaving client RPCs with raft writes

Although I've tested it manually on a live agent, this PR is a draft because it lacks unit tests. It's a bit tricky to test, so I'll have to think on it.

### DIFF
--- a/nomad/host_volume_endpoint.go
+++ b/nomad/host_volume_endpoint.go
@@ -4,11 +4,13 @@
 package nomad
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net/http"
 	"regexp"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/armon/go-metrics"
@@ -28,6 +30,9 @@ type HostVolume struct {
 	srv    *Server
 	ctx    *RPCContext
 	logger hclog.Logger
+
+	// volOps is used to serialize operations per volume ID
+	volOps sync.Map
 }
 
 func NewHostVolumeEndpoint(srv *Server, ctx *RPCContext) *HostVolume {
@@ -263,25 +268,31 @@ func (v *HostVolume) Create(args *structs.HostVolumeCreateRequest, reply *struct
 		return err
 	}
 
-	// Attempt to create the volume on the client.
-	//
-	// NOTE: creating the volume on the client via the plugin can't be made
-	// atomic with the registration, and creating the volume provides values we
-	// want to write on the Volume in raft anyways.
-	err = v.createVolume(vol)
-	if err != nil {
-		return err
-	}
+	// serialize client RPC and raft write per volume ID
+	index, err := v.serializeCall(vol.ID, func() (uint64, error) {
+		// Attempt to create the volume on the client.
+		//
+		// NOTE: creating the volume on the client via the plugin can't be made
+		// atomic with the registration, and creating the volume provides values
+		// we want to write on the Volume in raft anyways.
+		if err = v.createVolume(vol); err != nil {
+			return 0, err
+		}
 
-	// Write a newly created or modified volume to raft. We create a new request
-	// here because we've likely mutated the volume.
-	_, index, err := v.srv.raftApply(structs.HostVolumeRegisterRequestType,
-		&structs.HostVolumeRegisterRequest{
-			Volume:       vol,
-			WriteRequest: args.WriteRequest,
-		})
+		// Write a newly created or modified volume to raft. We create a new
+		// request here because we've likely mutated the volume.
+		_, idx, err := v.srv.raftApply(structs.HostVolumeRegisterRequestType,
+			&structs.HostVolumeRegisterRequest{
+				Volume:       vol,
+				WriteRequest: args.WriteRequest,
+			})
+		if err != nil {
+			v.logger.Error("raft apply failed", "error", err, "method", "register")
+			return 0, err
+		}
+		return idx, nil
+	})
 	if err != nil {
-		v.logger.Error("raft apply failed", "error", err, "method", "register")
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


tgross (commented):
I was a little worried this was going to be tough to do without either a coarse-grained mutex or a hot loop, but your implementation elegantly avoids both problems. Nice work on this so far.

nomad/host_volume_endpoint_test.go
```diff
@@ -763,6 +766,103 @@ func TestHostVolumeEndpoint_placeVolume(t *testing.T) {
...
+		return int(a.Sub(b))
+	})
+	// make sure they're a second apart (the "sleep" time)
+	test.GreaterEq(t, time.Second, dones[1].Sub(dones[0]), test.Sprint("second call should be a second after first"))
+	test.GreaterEq(t, time.Second, dones[2].Sub(dones[1]), test.Sprint("third call should be a second after second"))
+
+	// in total, it should have taken 3-4 seconds
+	end := time.Now()
+	test.GreaterEq(t, 3*time.Second, end.Sub(start), test.Sprint("all calls should take 3 seconds total"))
+	test.Less(t, 4*time.Second, end.Sub(start), test.Sprint("all calls should take less than 4 seconds"))
```
gulducat, line 863: shortcomings of this test:
* does not _actually_ test concurrency, only indirectly via:
  * ~1 second between each call's end times
  * ~3 total seconds for all 3 requests
* it says nothing about raft at all

gulducat: I changed the test to use channels to unblock specific RPCs, instead of sleeping. it still doesn't ensure that raft writes are coupled with the RPCs inside of the serialized call, but maybe it's good enough for government work (I am a government, today).

nomad/host_volume_endpoint_test.go
```diff
@@ -763,6 +764,113 @@ func TestHostVolumeEndpoint_placeVolume(t *testing.T) {
...
+			return "timeout"
+		}
+	}
+
+	must.Eq(t, "timeout", pullOp(), must.Sprint("nothing should be unblocked yet"))
+
+	close(rCh)
+	must.Eq(t, "register", pullOp(), must.Sprint("closing register channel should unblock Register"))
+
+	must.Eq(t, "timeout", pullOp(), must.Sprint("again blocked RPCs should remain so"))
```
tgross, line 863: At first look I thought this timeout was effectively testing the test mock's blocking and not the RPC blocking behavior. But really this prevents the test from getting stuck in case of a failure of one of the goroutines. So :+1: 

Because we know the following client RPCs are blocked, I'm pretty sure we can instead query the state store to verify the raft write has completed here too if we want? Ex. at this line we know the volume has been registered but not created (maybe we could set `parameters` on the volume object in the request to have something to check quickly).

tgross (approved):
LGTM!

nomad/host_volume_endpoint_test.go
```diff
@@ -294,7 +295,7 @@ func TestHostVolumeEndpoint_CreateRegisterGetDelete(t *testing.T) {
 		// win a race with the get RPC goroutine
 		time.AfterFunc(200*time.Millisecond, func() {
 			codec := rpcClient(t, srv)
-			var registerResp structs.HostVolumeRegisterResponse
+			var registerResp structs.HostVolumeCreateResponse
```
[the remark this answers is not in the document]
gulducat: this is a lil whoopsie from when I was debugging another thing.

nomad/host_volume_endpoint_test.go
gulducat: ah sure, I'll add a bit to check some unique state based on the operation

gulducat: tangent here: is there a reason we exclude `Parameters` from `HostVolumeStub` that gets returned by the `List` RPC?
https://github.com/hashicorp/nomad/blob/46bd0b171664641144ac5bcc59dbd8e8c1f45c8c/nomad/structs/host_volumes.go#L352-L368

I don't need it for this test, but I bumped into it on the way down a misguided road.

tgross: Stubs are minimal representations -- typically only what you minimally need for the CLI/UI list view (and handling blocking queries). Anything like capabilities, parameters, constraints, etc. are good candidates to omit here.

tgross: Ah right `v.volOps` doesn't let us know which. This is a little rough... we can't force the request goroutines to make their calls in order with `sync.Wait`, because we don't return from the server RPC call till we're done. But we _do_ know when a given RPC handler has started the client RPC, because we can read `mockHostVolumeClient.blockChans`. So with that we could do this:

* Start the register goroutine
* Wait until `mockHostVolumeClient.blockChans` is blocked on `register`
* Start the create goroutine (which we know is blocked)
* Start the delete goroutine (which we know is blocked)
* Unblock register
* Wait until `mockHostVolumeClient.blockChans` is blocked on _either_ create or delete
* Unblock whichever channel is blocked

nomad/host_volume_endpoint_test.go
```diff
@@ -822,6 +1009,12 @@ func (v *mockHostVolumeClient) setDelete(errMsg string) {
 func (v *mockHostVolumeClient) Create(
 	req *cstructs.ClientHostVolumeCreateRequest,
 	resp *cstructs.ClientHostVolumeCreateResponse) error {
+
+	// block until something runs unblockCurrent()
+	if bc := v.getBlockChan(); bc != nil {
+		bc <- "create"
+	}
```
gulducat, line 1016: I refactored it (again), and while it's still convoluted, it does actually test the serialization.

instead of these mock client methods blocking on a _receive_, they block on a _send_. then `unblockCurrent()` unblocks them by receiving, and it returns what it got, i.e. whichever client RPC was being called at the time.

then the test code makes sure that matches which server RPC completed, and we can check the state as appropriate for the operation type.

nomad/host_volume_endpoint_test.go
```diff
@@ -763,6 +766,187 @@ func TestHostVolumeEndpoint_placeVolume(t *testing.T) {
...
+		// and that the server RPC matches the unblocked client RPC.
+		select {
+		case serverOp := <-rpcDoneCh:
+			if serverOp != op {
+				t.Errorf("client RPC says %q; server RPC says %q", op, serverOp)
+				continue
+			}
+		case <-time.After(time.Second):
+			t.Error("timeout waiting for an RPC to finish")
+			break LOOP
```
tgross, line 912: This will deadlock the test if we break while there are any pending `funcs` (i.e. if the first or second times out), because we'll never unblock the channel for the remaining functions but call `funcs.Wait()` after the loop breaks. A deadlocked test manifests as the whole test run blowing up in a panic after the 20min timeout, which is pretty awful to debug on GHA because it's not obvious which test is panicking without a lot of rummaging in the logs.

We _could_ fix this by having a shutdown context on the `mockVolumeClient`. Just before we break the loop we can close that context. And then the send looks something like this:

```go
	// block until something runs unblockCurrent()
	if bc := v.getBlockChan(); bc != nil {
		select {
			case bc <- "delete":
			case <-v.shutdownCtx.Done():
				v.setBlockChan(nil)
		}
	}
```

That would allow the `funcs.Wait()` to run to completion and close out the test. We'd still end up failing the test because of the `t.Error` here.

All that being said, I'll :+1: this PR as-is, as it'd be a bit of effort to fix and only matters in the case we break the test. Up to you.

tgross (approved):
LGTM!

We can probably reuse this logic for tightening up the serialization of CSI operations too, so it's great that we're spending the effort to make this solid.

nomad/host_volume_endpoint_test.go
gulducat: I thought I had caught all the deadlocks! I caused enough of them while writing this :yum: but you're totally right. I pushed another commit to cover it.

> so it's great that we're spending the effort to make this solid.

thanks for saying so; I was worried I was deadlocked myself, spending too much time on it :sweat_smile:

PR #24694, opened 2024-12-17
Update e2e infra provision to expect providers

Juanadelacuesta:
### Description
The `e2e/terraform`  module used to provision the infra for the e2e tests uses an older version of terraform that is incompatible with the directive `depends_on`. This PR updates it by moving it into its own module and passing the providers from the new main, making it compatible with the current set up an reusable for the upgrade testing pipeline.

It also does some renaming to prepare for clients with a different architecture.

### Contributor Checklist
- [ ] **Changelog Entry** If this PR changes user-facing behavior, please generate and add a
  changelog entry using the `make cl` command.
- [ ] **Testing** Please add tests to cover any new functionality or to demonstrate bug fixes and
  ensure regressions will be caught.
- [ ] **Documentation** If the change impacts user-facing functionality such as the CLI, API, UI,
  and job configuration, please update the  Nomad website documentation to reflect this. Refer to
  the [website README](../website/README.md) for docs guidelines. Please also consider whether the
  change requires notes within the [upgrade guide](../website/content/docs/upgrade/upgrade-specific.mdx).

### Reviewer Checklist
- [ ] **Backport Labels** Please add the correct backport labels as described by the internal
  backporting document.
- [ ] **Commit Type** Ensure the correct merge method is selected which should be "squash and merge"
  in the majority of situations. The main exceptions are long-lived feature branches or merges where
  history should be preserved.
- [ ] **Enterprise PRs** If this is an enterprise only PR, please add any required changelog entry
  within the public repository.

PR #24855, opened 2025-01-14
E2E: update .gitignore files to avoid committing runtime files

tgross:
In #24694 we did a major refactoring of the E2E Terraform configuration. After deploying a cluster this morning, I noticed a few moved/removed files were not reflected in the .gitignore files. This changeset updates the .gitignore to have no unstaged files after applying.

### DIFF
--- a/e2e/.gitignore
+++ b/e2e/.gitignore
@@ -1,1 +1,0 @@
-provisioning.json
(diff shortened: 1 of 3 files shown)


jrasell (approved):
LGTM, thanks!

PR #24857, opened 2025-01-14
dynamic host volume: set default capability

tgross:
We can reduce the amount of volume specification configuration many users will need by setting a default capability on a dynamic host volume if none is set. The default capability will allow using the volume in read/write mode on its node, with no further restrictions except those that might be set in the jobspec.

(Note this PR will conflict with https://github.com/hashicorp/nomad/pull/24851, so whichever one lands last will need rebasing.)

### DIFF
--- a/nomad/structs/host_volumes.go
+++ b/nomad/structs/host_volumes.go
@@ -148,14 +148,10 @@ func (hv *HostVolume) Validate() error {
 			hv.RequestedCapacityMaxBytes, hv.RequestedCapacityMinBytes))
 	}
 
-	if len(hv.RequestedCapabilities) == 0 {
-		mErr = multierror.Append(mErr, errors.New("must include at least one capability block"))
-	} else {
-		for _, cap := range hv.RequestedCapabilities {
-			err := cap.Validate()
-			if err != nil {
-				mErr = multierror.Append(mErr, err)
-			}
+	for _, cap := range hv.RequestedCapabilities {
+		err := cap.Validate()
+		if err != nil {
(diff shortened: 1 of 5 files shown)


nomad/structs/host_volumes.go
```diff
@@ -221,6 +217,14 @@ func (hv *HostVolume) CanonicalizeForCreate(existing *HostVolume, now time.Time)
 		hv.CapacityBytes = 0 // returned by plugin
 		hv.HostPath = ""     // returned by plugin
 		hv.CreateTime = now.UnixNano()
+
+		if len(hv.RequestedCapabilities) == 0 {
+			hv.RequestedCapabilities = []*HostVolumeCapability{{
+				AttachmentMode: HostVolumeAttachmentModeFilesystem,
+				AccessMode:     HostVolumeAccessModeSingleNodeWriter,
+			}}
```
gulducat, line 225: I am curious about implications of this in the TF provider -- it could either be a computed property or explicit, and I don't recall how easy it is to account for an explicit option overwriting a computed one, if it gets added to the TF resource.

gulducat (approved):
this feels so satisfying

```hcl
# minimal.volume.hcl
name      = "smol-volume"
type      = "host"
plugin_id = "mkdir"
```

nomad/structs/host_volumes.go
tgross: If I understand what I've seen so far correctly (which isn't a given :grin:), you can set a schema field as `Optional: true` _and_ `Computed: true` and so long as the computed value comes back the same if unchanged, that works fine.

PR #24865, opened 2025-01-15
agent: Fix a bug where all syslog lines are notice when using JSON

jrasell:
The agent syslog write handler was unable to handle JSON log lines correctly, meaning all syslog entries when using JSON log format showed as NOTICE level.

This change adds a new handler to the Nomad agent which can parse JSON log lines and correctly understand the expected log level entry.

The change also removes the use of a filter from the default log format handler. This is not needed as the logs are fed into the syslog handler via hclog, which is responsible for level filtering.

Running on a Linux workstation (Ubuntu 24.04) I added the following lines to the `/etc/rsyslog.conf` before restarting rsyslog via `sudo service rsyslog restart`.

```
$template verbose, "%syslogseverity-text%, %syslogfacility%, %timegenerated%, %HOSTNAME%, %syslogtag%, %msg%\n"
*.* /var/log/all-messages.log;verbose
```

I then ran Nomad via `sudo -i nomad agent -dev -log-json -log-level=<LEVEL> -config=/home/jrasell/config.hcl` using the following configuration:

```hcl
enable_syslog = true
```

### Links
Jira: https://hashicorp.atlassian.net/browse/NET-11973

### Contributor Checklist

### Reviewer Checklist
  backporting document.
  history should be preserved.
  within the public repository.

### DIFF
--- a/command/agent/syslog.go
+++ b/command/agent/syslog.go
@@ -5,9 +5,11 @@ package agent
 
 import (
 	"bytes"
+	"io"
+	"regexp"
+	"strings"
 
 	gsyslog "github.com/hashicorp/go-syslog"
-	"github.com/hashicorp/logutils"
 )
 
 // levelPriority is used to map a log level to a syslog priority level. The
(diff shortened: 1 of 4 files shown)


command/agent/syslog.go
```diff
@@ -63,3 +73,48 @@ func (s *SyslogWrapper) Write(p []byte) (int, error) {
...
+	}
+
+	// Pull the log level from the message using the identified indexes and
+	// knowledge of the JSON formatting from go-hclog.
+	level := strings.ToTitle(string(logBytes[indexes[0][0]+10 : indexes[0][1]-2]))
+
+	// Rip the level out of the log message line. The level will be indicated
+	// by syslog, depending on the operators formatting setup.
+	afterLevel := logBytes[:indexes[0][0]]
+	afterLevel = append(afterLevel, logBytes[indexes[0][1]:]...)
```
tgross, line 116: This seems like an expensive bit of extra string mangling to do. Do we really need to remove the original `@level`?

tgross (approved):
LGTM

command/agent/syslog.go
jrasell: Done within: https://github.com/hashicorp/nomad/pull/24865/commits/eb32515129074c6cf843c923009c67fe958425f2

tgross: I can see a case where it might be confusing to have two different levels.

But I _think_ that because we're doing this before we're calling `WriteLevel`, we have to do this work for _all_ logs. Ex. we're doing this work for trace, debug, and info level and then throwing it out if you're logging at warn. (Unless the level gate is earlier than I'm seeing here, in which case maybe it doesn't matter.)

tgross (approved):

command/agent/syslog.go
jrasell: The level gate does come earlier as these logs are fed from hclog which performs that work. Any logs coming in here, are expected to be written.

I think the two levels are easy enough to reason about it shouldn't be too confusing. Depending on the syslog configuration, the priority might not even be included in the syslog entry output.

PR #24867, opened 2025-01-15
dynamic host volumes: move node pool governance to placement filter (CE)

tgross:
Enterprise governance checks happen after dynamic host volumes are placed, so if node pool governance is active and you don't set a node pool or node ID for a volume, it's possible to get a placement that fails node pool governance even though there might be other nodes in the cluster that would be valid placements.

Move the node pool governance for host volumes into the placement path, so that we're checking a specific node pool when node pool or node ID are set, but otherwise filtering out candidate nodes by node pool.

This changset is the CE version of ENT/2200.

Ref: https://github.com/hashicorp/nomad-enterprise/pull/2200


gulducat (approved):
:briefcase:

PR #24868, opened 2025-01-15
A dedicated variable at nomad/ui/defaults

philrenaud:
Adds a variable at `nomad/ui/defaults` to override agent config ui defaults, but can still be overriden by localStorage and queryParams.

A nice middle ground between "Everybody gets this because it's part of the agent configuration" and "Only you, in this browser, get it", because you can turn variable access on/off via ACL policy definitions, with the added benefit of "You can change this without having to restart your server", which is a problem the agent UI config block has.

### DIFF
--- a/ui/app/services/system.js
+++ b/ui/app/services/system.js
@@ -110,17 +110,21 @@ export default class SystemService extends Service {
     });
   }
 
-  @computed('defaults', 'token.selfToken', 'variableDefaults.Region')
+  @computed(
+    'defaults',
+    'token.selfToken',
+    'variableDefaults.Region',
+    'regions.[]'
+  )
   get defaultRegion() {
-    // // let variableDefaultRegion = this.variableDefaults.Region;
-    // let variableDefaultRegion = false;
-    // if (variableDefaultRegion) {
-    //   return { region: variableDefaultRegion };
+    // 1. If there is a defaults.region, and that region is within this.regions, return that region.
+    // 2. Otherwise, fallback to the agent/members' json.ServerRegion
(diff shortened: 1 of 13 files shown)


ui/app/services/system.js
```diff
@@ -109,17 +110,95 @@ export default class SystemService extends Service {
...
+  }
+
+  defaultProperties = [
+    'userDefaultRegion',
+    'userDefaultNamespace',
+    'userDefaultNodePool',
+  ];
+
+  async establishUIDefaults() {
+    // // First, check to see if there are localStorage properties set for each of the defaults.
```
philrenaud, line 156: Probably a premature optimization

ui/app/services/system.js
```diff
@@ -109,17 +110,95 @@ export default class SystemService extends Service {
...
+    this.variableDefaults = variableDefaults || {};
+    return {
+      agentDefaults: this.agentDefaults,
+      variableDefaults: this.variableDefaults,
+    };
+  }
+
+  async fetchVariableDefaults() {
+    try {
+      // if (this.can.can('read variable', 'nomad/ui/defaults', '*')) { // TODO: is wildcard correctly handled by "can"?
```
philrenaud, line 174: Tested without `can` ability checks and this fails gracefully. Can probably remove the condition.

philrenaud:
This PR was a solution in search of a problem, I think. We already accommodate localStorage and queryParam-based  (and ACL policy-based) settings in the UI, this just added an extra-specific layer. Not needed IMO.

PR #24871, opened 2025-01-15
dynamic host volumes quotas (CE)

tgross:
Allow users to configure a host volumes quota in MB. This will be enforced at the time of provisioning via create/register RPCs. This changeset is the CE version of ENT/2114. I'll have a separate PR for docs.

Ref: https://github.com/hashicorp/nomad-enterprise/pull/2114


gulducat (approved):
lgtm!

PR #24881, opened 2025-01-16
dynamic host volumes: enforce exclusive access in plan apply

tgross:
Some dynamic host volumes are claimed by allocations with the capability we borrowed from CSI called `single-node-single-writer`, which says only one allocation can use the volume, and it can use it in read/write mode. We enforce this in the scheduler, but if evaluations for different jobs were to be processed concurrently by the scheduler, it's possible to get plans that would fail to enforce this requirement. Add a check in the plan applier to ensure that non-terminal allocations have exclusive access when requested.

### DIFF
--- a/nomad/structs/funcs.go
+++ b/nomad/structs/funcs.go
@@ -145,6 +145,9 @@ func AllocsFit(node *Node, allocs []*Allocation, netIdx *NetworkIndex, checkDevi
 	reservedCores := map[uint16]struct{}{}
 	var coreOverlap bool
 
+	hostVolumeClaims := map[string]int{}
+	exclusiveHostVolumeClaims := []string{}
+
 	// For each alloc, add the resources
 	for _, alloc := range allocs {
 		// Do not consider the resource impact of terminal allocations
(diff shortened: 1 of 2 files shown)


nomad/structs/funcs.go
```diff
@@ -163,6 +166,18 @@ func AllocsFit(node *Node, allocs []*Allocation, netIdx *NetworkIndex, checkDevi
...
 			}
 		}
+
+		// Job will be nil in the scheduler, where we're not performing this check anyways
+		if checkDevices && alloc.Job != nil {
+			group := alloc.Job.LookupTaskGroup(alloc.TaskGroup)
+			for _, volReq := range group.Volumes {
+				hostVolumeClaims[volReq.Source]++
+				if volReq.AccessMode ==
+					CSIVolumeAccessMode(HostVolumeAccessModeSingleNodeSingleWriter) {
```
gulducat, line 176: I got a mental hiccup seeing `CSIVolume...(HostVolume...)` -- I'd think it should be

```suggestion
					HostVolumeAccessMode(HostVolumeAccessModeSingleNodeSingleWriter) {
```

gulducat (approved):

nomad/structs/funcs.go
tgross: Unfortunately not (ref [`volumes.go#L100-L101`](https://github.com/hashicorp/nomad/blob/v1.9.4/nomad/structs/volumes.go#L100-L101)) because the type for `VolumeRequest` is named for the CSI mode.

```go
// VolumeRequest is a representation of a storage volume that a TaskGroup wishes
// to use.
type VolumeRequest struct {
	Name           string
	Type           string
	Source         string
	ReadOnly       bool
	Sticky         bool
	AccessMode     CSIVolumeAccessMode
	AttachmentMode CSIVolumeAttachmentMode
	MountOptions   *CSIMountOptions
	PerAlloc       bool
}
```

But both are "defined types" on string, which msgpack serializes to the same thing on the wire. So we _could_ swap out for a new shared `VolumeAccessMode` / `VolumeAttachmentMode` without breaking backcompat. Especially because in this case we didn't define types in the `api` package version of these fields. They're just strings there!

I'll do that, but under a separate PR.

tgross: https://github.com/hashicorp/nomad/pull/24896

PR #24882, opened 2025-01-16
docs: clarify workload-associated policy parameters

tgross:
Workload-associated ACL policies can only be set on a specific job within a namespace, not the namespace as a whole. Clarify the documentation for the CLI and API.

Fixes: https://github.com/hashicorp/terraform-provider-nomad/issues/500
Ref: https://github.com/hashicorp/terraform-provider-nomad/pull/504
Ref: https://hashicorp.atlassian.net/browse/NET-11994


jrasell (approved):
LGTM!

aimeeu (approved):

PR #24896, opened 2025-01-17
refactor: volume request modes to be generic between DHV/CSI

tgross:
When we implemented CSI, the types of the fields for access mode and attachment mode on volume requests were defined with a prefix "CSI". This gets confusing now that we have dynamic host volumes using the same fields. Fortunately the original was a typedef on string, and the Go API in the `api` package just uses strings directly, so we can change the name of the type without breaking backwards compatibility for the msgpack wire format or HTTP API.

Update the names to `VolumeAccessMode` and `VolumeAttachmentMode`. Keep the CSI and DHV specific value constant names for these fields (they aren't currently 1:1), so that we can easily differentiate in a given bit of code which values are valid.

Ref: https://github.com/hashicorp/nomad/pull/24881#discussion_r1920702890

### DIFF
--- a/nomad/structs/csi.go
+++ b/nomad/structs/csi.go
@@ -121,19 +121,11 @@ type CSIVolumeCapability struct {
 	AccessMode     VolumeAccessMode
 }
 
-// VolumeAttachmentMode chooses the type of storage api that will be used to
-// interact with the device.
-type VolumeAttachmentMode string
-
 const (
 	CSIVolumeAttachmentModeUnknown     VolumeAttachmentMode = ""
 	CSIVolumeAttachmentModeBlockDevice VolumeAttachmentMode = "block-device"
 	CSIVolumeAttachmentModeFilesystem  VolumeAttachmentMode = "file-system"
 )
-
-// VolumeAccessMode indicates how a volume should be used in a storage topology
-// e.g whether the provider should make the volume available concurrently.
-type VolumeAccessMode string
 
 const (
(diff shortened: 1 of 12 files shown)


nomad/structs/csi.go
```diff
@@ -117,33 +117,33 @@ func (t *TaskCSIPluginConfig) Copy() *TaskCSIPluginConfig {
...
-	AccessMode     CSIVolumeAccessMode
+	AttachmentMode VolumeAttachmentMode
+	AccessMode     VolumeAccessMode
 }
 
-// CSIVolumeAttachmentMode chooses the type of storage api that will be used to
+// VolumeAttachmentMode chooses the type of storage api that will be used to
 // interact with the device.
-type CSIVolumeAttachmentMode string
+type VolumeAttachmentMode string
```
gulducat, line 126: slightly unintuitive that this now-generalized thing still lives in `csi.go`

gulducat (approved):
lgtm!

nomad/structs/csi.go
tgross: Good point. I'll move this into `nomad/structs/volumes.go`. Looking at that file, I see that the `VolumeRequest` validation is also using the "CSI" values for validating host volumes too, so I'll tweak that.

tgross: Done!

gulducat (approved):
:+1:

PR #24906, opened 2025-01-20
consul token and namespace

tommyvicananza:
### Contributor Checklist

### Reviewer Checklist
  backporting document.
  history should be preserved.
  within the public repository.

### DIFF
--- a/command/job_plan.go
+++ b/command/job_plan.go
@@ -159,7 +159,7 @@ func (c *JobPlanCommand) AutocompleteArgs() complete.Predictor {
 func (c *JobPlanCommand) Name() string { return "job plan" }
 func (c *JobPlanCommand) Run(args []string) int {
 	var diff, policyOverride, verbose bool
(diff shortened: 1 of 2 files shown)


tgross (changes requested):
@tommyvicananza you haven't provided a description of what you're trying to do here and you haven't signed the CLA, both of which are necessary before we can review. 

But also, we don't backport changes to the community edition earlier than the latest version.

tgross:
Had a moment to look through this and it looks like we're adding the command line arguments that `job run` has (ref [`job_run.go#L115-L127`](https://github.com/hashicorp/nomad/blob/v1.9.5/command/job_run.go#L115-L127) to support the legacy Consul token workflow. Which would be a fine feature except that we're removing that workflow entirely in 1.10: https://developer.hashicorp.com/nomad/docs/release-notes/nomad/upcoming#nomad-1-10-0-lts So we're not going to be able to accept this PR unfortunately. But thanks for getting involved!

PR #24911, opened 2025-01-21
volume spec: fix `access_mode` field in examples

tgross:
The `volume init` command creates example volume specifications. But one of the values for `capability.access_mode` is not a valid value. Correct the example to match the validation logic.


pkazmierczak (approved):
Good catch!

PR #24919, opened 2025-01-23
reporting: Update server to accommodate new enterprise reporting.

jrasell:
Nomad Enterprise will utilise new reporting metrics and the changes here allow this work to be conducted.

The server specific GetClientNodesCount function has been removed from CE as this is only called within enterprise code. A new heartbeater function allows us to get the number of active timers, which can be used by the heartbeater metrics and any other callers that want this data.

### PR Process Notes
This has been originally reviewed within the enterprise codebase.

In order to juggle the CE and Enterprise specific code portions, I plan on backporting this change to CE only and then performing manual merging into Enterprise along with the specific changes needed there.

Once the merging juggling has been complete, I will raise a followup documentation and changelog PR.

### Links
Jira: https://hashicorp.atlassian.net/browse/NET-9340

### Contributor Checklist

### Reviewer Checklist
  backporting document.
  history should be preserved.
  within the public repository.

### DIFF
--- a/nomad/heartbeat.go
+++ b/nomad/heartbeat.go
@@ -251,19 +251,26 @@ func (h *nodeHeartbeater) clearAllHeartbeatTimers() error {
 	return nil
 }
 
-// heartbeatStats is a long running routine used to capture
-// the number of active heartbeats being tracked
+// heartbeatStats is a long-running routine used to capture the number of
+// active heartbeats being tracked.
 func (h *nodeHeartbeater) heartbeatStats() {
 	for {
 		select {
 		case <-time.After(5 * time.Second):
-			h.heartbeatTimersLock.Lock()
-			num := len(h.heartbeatTimers)
-			h.heartbeatTimersLock.Unlock()
+			num := h.getHeartbeatTimerNum()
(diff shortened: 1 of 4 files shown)


tgross (approved):
LGTM

There's a semgrep rule firing on that `heartbeatStats()` method around the timer. It's not technically a false positive but it is meaningless. We leak one timer goroutine when we shutdown the agent! I wouldn't worry about this for now.

jrasell:
> LGTM
> 
> There's a semgrep rule firing on that `heartbeatStats()` method around the timer. It's not technically a false positive but it is meaningless. We leak one timer goroutine when we shutdown the agent! I wouldn't worry about this for now.

Interestingly Go stdlib time library does have the following function comment, so it might not even be a problem:

> // Before Go 1.23, this documentation warned that the underlying
// [Timer] would not be recovered by the garbage collector until the
// timer fired, and that if efficiency was a concern, code should use
// NewTimer instead and call [Timer.Stop] if the timer is no longer needed.
// As of Go 1.23, the garbage collector can recover unreferenced,
// unstopped timers. There is no reason to prefer NewTimer when After will do.

mismithhisler (approved):

PR #24922, opened 2025-01-23
CSI: don't overwrite context with empty value from request

tgross:
When a volume is updated, we merge the new definition to the old. But the volume's context comes from the plugin and is likely not present in the user's volume specification. Which means that if the user re-submits the volume specification to make an adjustment to the volume, we wipe out the context field which might be required for subsequent operations by the CSI plugin. This was discovered to be a problem with the Terraform provider and fixed there, but it's also a problem for users of the `volume create` and `volume register` commands.

Update the merge so that we only overwrite the value of the context if it's been explictly set by the user. We still need to support user-driven updates to context for the `volume register` workflow.

Ref: https://github.com/hashicorp/terraform-provider-nomad/pull/503
Fixes: https://github.com/democratic-csi/democratic-csi/issues/438

Using the [Democratic CSI](https://github.com/democratic-csi/democratic-csi) plugin.

**Before**

```
$ nomad volume create ./volume.hcl
Created external volume csi-volume-nfs with ID csi-volume-nfs

$ nomad operator api "/v1/volume/csi/csi-volume-nfs?namespace=prod" | jq .Context
{
  "provisioner_driver": "nfs-client",
  "node_attach_driver": "nfs",
  "server": "192.168.1.170",
  "share": "/srv/nfs_data/v/csi-volume-nfs"
}

$ nomad volume create ./volume.hcl
Created external volume csi-volume-nfs with ID csi-volume-nfs

$ nomad operator api "/v1/volume/csi/csi-volume-nfs?namespace=prod" | jq .Context
{}
```

**After**

```
$ nomad volume create ./volume.hcl
Created external volume csi-volume-nfs with ID csi-volume-nfs

$ nomad operator api "/v1/volume/csi/csi-volume-nfs?namespace=prod" | jq .Context
{
  "server": "192.168.1.170",
  "share": "/srv/nfs_data/v/csi-volume-nfs",
  "provisioner_driver": "nfs-client",
  "node_attach_driver": "nfs"
}

$ nomad volume create ./volume.hcl
Created external volume csi-volume-nfs with ID csi-volume-nfs

$ nomad operator api "/v1/volume/csi/csi-volume-nfs?namespace=prod" | jq .Context
{
  "provisioner_driver": "nfs-client",
  "node_attach_driver": "nfs",
  "server": "192.168.1.170",
  "share": "/srv/nfs_data/v/csi-volume-nfs"
}
```

### Contributor Checklist

### Reviewer Checklist
  backporting document.
  history should be preserved.
  within the public repository.

### DIFF
--- a/nomad/structs/csi.go
+++ b/nomad/structs/csi.go
@@ -846,9 +846,13 @@ func (v *CSIVolume) Merge(other *CSIVolume) error {
 			"volume parameters cannot be updated"))
 	}
 
-	// Context is mutable and will be used during controller
-	// validation
-	v.Context = other.Context
+	// Context is mutable and will be used during controller validation, but we
+	// need to ensure we don't remove context that's been previously stored
+	// server-side if the user has submitted an update without adding it to the
+	// spec manually (which we should not require)
+	if len(other.Context) != 0 {
+		v.Context = other.Context
+	}
 	return errs.ErrorOrNil()
 }
 
--- a/.changelog/24922.txt
+++ b/.changelog/24922.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+csi: Fixed a bug where volume context from the plugin would be erased on volume updates
+```
--- a/nomad/structs/csi_test.go
+++ b/nomad/structs/csi_test.go
@@ -711,6 +711,13 @@ func TestCSIVolume_Merge(t *testing.T) {
 						{Segments: map[string]string{"rack": "R2"}},
 					},
 				},
+				Context: map[string]string{
+					// a typical context for democratic-csi
+					"provisioner_driver": "nfs-client",
+					"node_attach_driver": "nfs",
+					"server":             "192.168.1.170",
+					"share":              "/srv/nfs_data/v/csi-volume-nfs",
+				},
 			},
 			update: &CSIVolume{
 				Topologies: []*CSITopology{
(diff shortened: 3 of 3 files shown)


nomad/structs/csi.go
```diff
@@ -846,9 +846,13 @@ func (v *CSIVolume) Merge(other *CSIVolume) error {
...
-	// Context is mutable and will be used during controller
-	// validation
-	v.Context = other.Context
+	// Context is mutable and will be used during controller validation, but we
+	// need to ensure we don't remove context that's been previously stored
+	// server-side if the user has submitted an update without adding it to the
+	// spec manually (which we should not require)
+	if len(other.Context) != 0 {
+		v.Context = other.Context
+	}
```
gulducat, line 855: this makes good sense for `create`, where the controller plugin generates the `context` dynamically, but I wonder about this edge case:

what if a user `register`s a volume with some `context`, but later wants to remove the context entirely? is that even a plausible concern? I suppose one could `deregister` the volume then re-`register` it with no context, since the actual volume out in the world would be unaffected.

gulducat (approved):
lgtm! I have one imaginary edge case to consider.

nomad/structs/csi.go
tgross: The only time that would happen is if they updated the plugin and the new version of the plugin _requires_ no context. In that case, you're almost certainly better off deregistering and re-registering

